PR #5507, opened 2019-03-25
Don't print ROLLING BACK if atomic is not set

Evesy:
Signed-off-by: Mike Eves <meves23@gmail.com>

Helm will only automatically rollback if `--atomic` has been passed as an argument to the upgrade command. Currently though Helm will print `ROLLING BACK` on _all_ failed upgrades which causes some confusion when atomic has not been set.

This PR changes the logging slightly to be less misleading.


thomastaylor312 (approved):
Good catch on this @Evesy!

PR #5520, opened 2019-03-28
ref(*): Uses go modules and removes glide

thomastaylor312:
Bowing to the inevitable.


Makefile
```diff
@@ -157,23 +160,21 @@ clean:
...
 .PHONY: bootstrap
+# Note that because go modules are not user friendly, we can't just do a go get
+# in the current directory because it will add it to the go.mod
 bootstrap:
-ifndef HAS_GLIDE
-	go get -u github.com/Masterminds/glide
-endif
 ifndef HAS_GOX
-	go get -u github.com/mitchellh/gox
+	cd /tmp && GO111MODULE=off go get -u github.com/mitchellh/gox
```
thomastaylor312: If someone has a better idea on how to handle this, let me know

thomastaylor312:
After discussing this in the dev call (and also to make this land in the top 5 shortest PRs ever), we decided that it would be better to land on Go modules in Helm v3

PR #5527, opened 2019-03-29
fix(pkg/chartutil): only include external objects in capabilities

adamreese:


bacongobbler (approved):
After looking through apimachinery, it makes sense that we don't want to expose internal types to the capabilities object. LGTM!

PR #5112, opened 2018-12-29
fix(helm): Wait for CRDs to reach established state for crd_install hook

mortent:
**What this PR does / why we need it**:
There is a race condition in the crd_install hook implementation, where there is a chance a CRD is not yet ready by the time CRs are being created. This is reported in issue #4925. This change makes sure CRDs installed through the crd_install hook reaches the `established` state before the hook is considered complete.

Fixes #4925

**Special notes for your reviewer**:
Unit-testing code in the kubernetes client is difficult, as the builder/infos is tightly coupled with the API server. I will look into how to improve testing for this part of the codebase, but I would like to separate it from this PR.

**If applicable**:
- [x] this PR contains documentation
- [x] this PR contains unit tests
- [x] this PR has been tested for backwards compatibility

PR #5532, opened 2019-03-31
fix(helm): Delete hooks should wait for resource to be removed from etcd before continuing

### LINKED ISSUE #5375: Deletion of CRD with crd-install and before-hook-creation randomly fails (race condition)?
This is a follow-up to #5278 and is possibly related to #4925.

Output of `helm version`:

Client: &version.Version{SemVer:"v2.12.3", GitCommit:"eecf22f77df5f65c823aacd2dbd30ae6c65f186e", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.12.3", GitCommit:"eecf22f77df5f65c823aacd2dbd30ae6c65f186e", GitTreeState:"clean"}

Output of `kubectl version`:
Client Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.3", GitCommit:"721bfa751924da8d1680787490c54b9179b1fed0", GitTreeState:"clean", BuildDate:"2019-02-01T20:08:12Z", GoVersion:"go1.11.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"12", GitVersion:"v1.12.4", GitCommit:"f49fa022dbe63faafd0da106ef7e05a29721d3f1", GitTreeState:"clean", BuildDate:"2018-12-14T06:59:37Z", GoVersion:"go1.10.4", Compiler:"gc", Platform:"linux/amd64"}

Rancher2 on-premise

Issue description:

I'm installing a helm chart with a CRD and the annotations

```
  annotations:
    helm.sh/hook: crd-install
    helm.sh/hook-delete-policy: before-hook-creation
```

When I Install/remove the chart multiple times, sometimes it fails with the following error:

```
Release "hook-test" does not exist. Installing it now.
Error: object is being deleted: customresourcedefinitions.apiextensions.k8s.io "tests.my.test.com" already exists
```

A test to reproduce this follows.

mortent:
There is a race condition when using the `before-hook-creation` deletion policy. The code in Tiller makes the delete call, but this only sets the deletion timestamp on the resource and doesn't actually remove the object from etcd. With finalizers, there might be some time before the object is actually removed. If the object isn't removed by the time the object is re-created, the chart will fail.
This PR adds code fixes the delete call used in the hook to wait for the object to be deleted from etcd before considering the delete operation completed.

Fixes #5375

**If applicable**:


mortent:
@micw I think this should fix the issue you are describing in #5375. I'm having trouble reproducing the issue, so could you try to build Helm with this change and see if it works?

thomastaylor312:
@mortent Please ping me once you are able to reproduce the issue with @micw or any other person experiencing the issue and I'll review for you

pkg/kube/client.go
```diff
@@ -391,16 +392,25 @@ func (c *Client) Update(namespace string, originalReader, targetReader io.Reader
 // Delete deletes Kubernetes resources from an io.reader.
 //
 // Namespace will set the namespace.
-func (c *Client) Delete(namespace string, reader io.Reader) error {
+func (c *Client) Delete(namespace string, reader io.Reader, timeout int64, shouldWait bool) error {
```
thomastaylor312: Just an FYI that this is a breaking API change. Not many people should be using `pkg/kube` directly that much, but if we need to add the timeout, we'll probably need to add another method so as to not break backwards compatibility

mortent:
I have updated the PR to add a new delete method in addition to the existing one to avoid breaking the API. I've also added unit tests.

I'm not able to reliably reproduce the reported problem on GKE, but the way the delete hook works today does have a weakness that should be addressed by this change. 

@thomastaylor312 This should be ready for review. If you have a chance, #5112 has some similarities to this one and needs review from a second maintainer.

micw:
@mortent @thomastaylor312 Sorry, I was a bit busy those days. I try to verify this PR in my test setup.

micw:
tl;dr: Works!

My procedure to reproduce it:

- clone the helm repo
- fetch the PR as new branch: git fetch origin pull/5532/head:pr-5532
- checkout the branch: git  checkout pr-5532
- build+push helm and the docker image: https://gist.github.com/micw/30190083b3e5d7df8dc30bc1526bdb57
- ran my test script from https://github.com/micw/helm-hook-test with official helm release
--> failed: Error: object is being deleted: customresourcedefinitions.apiextensions.k8s.io "tests.my.test.com" already exists

- switched tiller to micwy/helm-test:pr-5532

```
$ helm version
Client: &version.Version{SemVer:"v2.12.3", GitCommit:"eecf22f77df5f65c823aacd2dbd30ae6c65f186e", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.12+unreleased", GitCommit:"7196ba364d46f93685feb486c255b21609703196", GitTreeState:"clean"}
```

- ran my script again

... running since 10 minutes without error ;-)

I let it run this night and send an update tomorrow.

thomastaylor312:
Thanks for verifying @micw!

pkg/tiller/environment/environment.go
```diff
@@ -119,6 +119,16 @@ type KubeClient interface {
 	// by "\n---\n").
 	Delete(namespace string, reader io.Reader) error
 
+	// Delete destroys one or more resources. If shouldWait is true, the function
```
thomastaylor312: Comment should start with `DeleteWithTimeout`

pkg/tiller/release_server.go
```diff
@@ -448,7 +448,7 @@ func (s *ReleaseServer) deleteHookByPolicy(h *release.Hook, policy string, name,
 	b := bytes.NewBufferString(h.Manifest)
 	if hookHasDeletePolicy(h, policy) {
 		s.Log("deleting %s hook %s for release %s due to %q policy", hook, h.Name, name, policy)
-		if errHookDelete := kubeCli.Delete(namespace, b); errHookDelete != nil {
+		if errHookDelete := kubeCli.DeleteWithTimeout(namespace, b, 60, true); errHookDelete != nil {
```
thomastaylor312: Should this timeout value be configurable by annotation on the hook? If we don't want that, then this definitely should be a constant so we can give it a meaningful name

thomastaylor312 (commented):
Just a few points of discussion

pkg/tiller/release_server.go
mortent: I've moved the timeout value into a constant. Making it configurable makes sense, although I think the default should be a timeout larger than 0. Waiting for the resource to be fully deleted is a reasonable default behavior.

So would it make sense to create an additional annotation, something like `helm.sh/hook-delete-timeout` that would override the default value? Or allow appending the timeout value at the end of the current hooks, something like `before-hook-creation:60`? I think I prefer the first alternative.

thomastaylor312: I think I agree with you with what the name should be. Let me check with some of the other maintainers as well.

thomastaylor312: Yeah, let's go with `helm.sh/hook-delete-timeout` and add that as an override for the default. Thanks for the work on this @mortent!

thomastaylor312:
@mortent I am working on testing this now. Once you add hook-delete-timeout, let me know and I'll give it a final review!

thomastaylor312:
Ok, this is working great and I confirmed that it works with the test chart @micw used to replicate the issue. As soon as we have that timeout configuration in place, this should be good to go

mortent:
@thomastaylor312 I've updated the PR. If it looks good, I can squash the commits before we merge.

thomastaylor312:
Sorry! Somehow this totally slipped through my notifications. I'll look at it first thing tomorrow.

thomastaylor312 (approved):
This is looking great! I'll ping one of the other maintainers for a second review

hickeyma:
Review in progress. Testing looks good. Just need to eye ball the code next.

hickeyma (commented):
@mortent Do you mind squashing the commits?

hickeyma (changes requested):
@mortent It's nearly there. Can you update the (docs)[https://github.com/helm/helm/blob/master/docs/charts_hooks.md] with the new hook?

mortent:
I'll squash the commits and update the docs today.

mortent:
@hickeyma @thomastaylor312 I have squashed the commits and updated the docs.

pkg/kube/client.go
```diff
@@ -449,15 +449,34 @@ func (c *Client) cleanup(newlyCreatedResources []*resource.Info) (cleanupErrors
 //
 // Namespace will set the namespace.
 func (c *Client) Delete(namespace string, reader io.Reader) error {
```
hickeyma: Is this function now obsolete?

thomastaylor312: I don't think it is, but even if it is obsolete, we cannot delete it because it would break backwards compatibility (since it is an exported function)

hickeyma: Ok, probably best to leave it then.

hickeyma (approved):
LGTM @mortent. Thanks for all the work and changes.

PR #5553, opened 2019-04-04
ref(create): app version should be 0.1.0

bacongobbler:
When creating a Helm chart for the first time, the assumption should be that the app version is also 0.1.0, implying this is for a new application.

### DIFF
--- a/cmd/helm/create.go
+++ b/cmd/helm/create.go
@@ -80,7 +80,7 @@ func (o *createOptions) run(out io.Writer) error {
 		Description: "A Helm chart for Kubernetes",
 		Type:        "application",
 		Version:     "0.1.0",
-		AppVersion:  "1.0",
+		AppVersion:  "0.1.0",
 		APIVersion:  chart.APIVersionV1,
 	}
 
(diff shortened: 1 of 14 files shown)


thomastaylor312 (approved):

adamreese:
What do you think about setting it to a TODO?  "TODO: Put the application's version here."
I think bundler does this when scaffolding out a gemspec

bacongobbler:
I like that idea. I'll see what I can do.

bacongobbler:
done. Chart.yaml created with `helm create` now adds some commentary. This resulted in changing some behaviour within chartutil to accomodate the change, as `chartutil.Create` no longer takes chart metadata and writes it to a Chart.yaml. Instead, `chartutil.Create` accepts a chart name and creates the chart with the default values.

Let me know if that design looks too invasive and I can back that out.

adamreese (approved):

maver1ck:
Hi Guys,
Why we remove image.tag here ?
App Version cannot be set from values.yaml which mean that you need to modify chart on every deploy.

Issue #4573, opened 2018-08-31
helm lint --strict fails even if default value is defined


vladimirtiukhtin:
Hello

In the deployments.yaml I have this
```
    spec:
      containers:
        - name: test
          image: "{{ .Values.registry_address }}/{{ .Values.image.name }}:{{ .Values.image.tag | default .Values.environment }}"
```
In the values.yaml
```
---

# Default values for the chart set.
# This is a YAML-formatted file.
# Declare variables to be passed into your templates.

registry_address: registry.example.com
revisionHistoryLimit: 3
environment: production
location: local
```
Anyway ```helm lint --strict``` fails with undefined variable
```
==> Linting synthesis
[INFO] Chart.yaml: icon is recommended
[ERROR] templates/: render error in "test/templates/deployments.yaml": template: test/templates/deployments.yaml:24:99: executing "test/templates/deployments.yaml" at <.Values.test...>: map has no entry for key "tag"

Error: 1 chart(s) linted, 1 chart(s) failed
```
In my opinion if default value for variable is set helm should not treat it as undefined
Output of `helm version`:
```
Client: &version.Version{SemVer:"v2.10.0", GitCommit:"9ad53aac42165a5fadc6c87be0dea6b115f93090", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.10.0", GitCommit:"9ad53aac42165a5fadc6c87be0dea6b115f93090", GitTreeState:"clean"}
```
Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"11", GitVersion:"v1.11.2", GitCommit:"bb9ffb1654d4a729bb4cec18ff088eacc153c239", GitTreeState:"clean", BuildDate:"2018-08-07T23:17:28Z", GoVersion:"go1.10.3", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"11", GitVersion:"v1.11.2", GitCommit:"bb9ffb1654d4a729bb4cec18ff088eacc153c239", GitTreeState:"clean", BuildDate:"2018-08-07T23:08:19Z", GoVersion:"go1.10.3", Compiler:"gc", Platform:"linux/amd64"}
```
Cloud Provider/Platform (AKS, GKE, Minikube etc.): 
```
baremetal
```

bacongobbler:
[...]

Issue #1463, opened 2016-10-26
issues with validateNoMissingValues linter (template: unexpected EOF)


prydonius:
This linter rule extracts each template fragment (`{{.*}}`) in a given template file and executes it on it's own to determine if it is empty (`<no value>` or empty).

This doesn't work if the fragment creates a block (`if`, `else`, `range`), since these cannot be executed independently - the parser expends an `end` fragment. Currently, the linter is designed to ignore the `if` and `else` cases, but ignoring `range` can cause the following issue:

```
  {{- range .Values.alertmanager.ingress.hosts }}
    - host: {{ .hostName }}
  {{- end -}}
```

In the above example, `.hostName` is an element within the context of `.Values.alertmanager.ingress.hosts`, however since `{{ .hostName }}` is executed independently, the linter incorrectly reports that this value is empty.

Potential solutions:
1. Update the linter to correctly deal with control flow
   
   This would mean doing the right thing with `if`, `else`, and `range` by ignoring blocks based on conditionals, or changing context within `range` blocks.
2. Render the whole template and look for keys that are empty, or set to `<no value>`.
   
   This would simplify the linter logic, and also correct the conditional flow issues. However, it will be harder for the user to link back a snippet of rendered template to the template fragments that rendered it.
3. Remove the linter rule
   
   It might be simpler to just remove this linter rule, since:
   1. empty values can be intentional (at least empty string, not `<no value>`
   2. some empty values might cause issues that could be picked up by the yaml validator or the kubernetes validator anyway

cc @mgoodness @technosophos @migmartri

prydonius:
Another issue I found related to this linter:

```
  joomla-password: {{ randAlphaNum 10 | b64enc | quote }}
```

```
[WARNING] templates/secrets.yaml: template: tpl:1:28: executing "tpl" at <b64enc>: wrong number of args for b64enc: want 1 got 0
```

migmartri:
Thanks @prydonius 

Some of those linter rules came up based on my personal preference and were based the Chart examples that existed by then. 

As you stated in your examples, supporting advanced goTemplates syntax (present in most of the current Charts) might be difficult to maintain and costly.

Taking into account that from my observation, this specific linter rule has not proved its usefulness yet, I vote for disabling it.

technosophos:
I'm good with disabling it for now, and perhaps investing some more time in it later.

nrvmodi:
I have same issue and issue was indentation

PR #5555, opened 2019-04-05
Remove missing value validation

### LINKED ISSUE #5417: Checking for all the values in a template files are defined fails when using helm lint --strict
It took a bit of digging but there are multiple tickets of users reporting `[ERROR] templates/: render error in ... <.Values.KeyName>: map has no entry for key "KeyName"` when using `helm lint --strict`. It is solved by setting a default value type for empty default values such as `[]`, `{}` or `""`.

It seems that this check was supposed to be disabled based on https://github.com/helm/helm/issues/1463. I think this is confusing because `map has no entry for key` is being treated as a warning and displayed as an error when using `--strict`. I think it should either:
- Be removed completely as desired in https://github.com/helm/helm/issues/1463
- Be treated as an error and cause `helm lint` to fail
- Be treated as a warning and display in `helm lint` with `[WARNING]` and failed in `helm lint --strict`

Helm version: `v2.13.0-rc.2`

walkafwalka:
Signed-off-by: walkafwalka <2865898-walkafwalka@users.noreply.gitlab.com>

This PR prevents `helm lint --strict` from failing values without a default value defined.

There are possible alternatives available which are described in https://github.com/helm/helm/issues/5417#issuecomment-473381728. I chose to disable the missing value validation because 1) it was meant to be disabled based on https://github.com/helm/helm/issues/1463, 2) this is not being tested, and 3) requires minimal change.

**If applicable**:

closes #5417, #4573

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -228,13 +228,9 @@ func (e *Engine) renderWithReferences(tpls map[string]renderable, referenceTpls 
 		}
 	}()
 	t := template.New("gotpl")
-	if e.Strict {
-		t.Option("missingkey=error")
-	} else {
-		// Not that zero will attempt to add default values for types it knows,
-		// but will still emit <no value> for others. We mitigate that later.
-		t.Option("missingkey=zero")
-	}
+	// Not that zero will attempt to add default values for types it knows,
+	// but will still emit <no value> for others. We mitigate that later.
+	t.Option("missingkey=zero")
 
 	funcMap := e.alterFuncMap(t, referenceTpls)
 


walkafwalka:
Is this is the only validation that the `--strict` flag checks for? If so, should we also deprecate `--strict` if we choose to deprecate this?

dgvigil:
As a user, I would much rather there be a warning on a straight lint and then it fail on a `--strict`. If it is fine for a chart to have an empty file then just remove that line.  The line is there for a reason, and thus should warn/error.

bacongobbler:
Hey @walkafwalka, sorry for not getting back to you sooner. The team has been focusing for the past few months on getting Helm 3 out the door, and that has meant it's been harder to find time to review feature additions for Helm 2.

Helm 2.15 is soon to be released, and will be the last feature release for Helm 2. Because of the time commitment required to review and test this, we won't be able to find the time to review this. That being said, if you want to try again by reconsidering this for Helm 3, that would be much appreciated. I'd hold off for a few weeks until 2.15 is out as we are planning to move the dev-v3 branch over to be the new master branch, so there will be some shuffling around for a bit, but once that's done we'd be happy to consider this for the 3.1.0 release.

Thanks again!

Issue #3159, opened 2017-11-16
Feat: Configuration option for tiller probeAddr IP/port.


libesz:
Currently the probeAddr listens on *:44135 and [it is not configurable](https://github.com/kubernetes/helm/blob/e8e6ac5d7783808cc0bd1adad053bec339849647/cmd/tiller/tiller.go#L62). Together with the host networking option for tiller, installation gives unexpected external visibility for this listening point for any potential user.
This might have a configuration option, similarly to the [grpcAddr](https://github.com/kubernetes/helm/blob/e8e6ac5d7783808cc0bd1adad053bec339849647/cmd/tiller/tiller.go#L70), or in this case, I would even expect that is can be disabled entirely if not needed.

libesz:
Up!
We would have interest in implementing this. Simply need a thumbs up that it will be accepted as described above.
Thanks!

bacongobbler:
So long as this is configurable through the CLI and documented, I don't see why not. 👍

libesz:
Hey, thanks @gavinmead ! 👍 Fine for us, we deferred the fix for this so it is very welcomed!

gavinmead:
Been crazy busy with work.  I have some free time now.  Apologies for the delay.

AlexShemeshWix:
Any update on this issue? I really need it.

mumoshu:
#5557 partly address this

mumoshu:
#5561 resolves this

PR #3440, opened 2018-02-02
tiller probe server is configurable #3159

gavinmead:
This PR includes the following changes:
* replaces probeAddr constant with a flag.
* defines a probe flag to support enable/disable of probe server.  Defaults
to true which matches current functionality.
* uses probe flag as a conditional for probe server startup

closes #3159

PR #5557, opened 2019-04-05
feat: make tiller probeAddr configurable

mumoshu:
As many people have requested and discussed in #3159.

The variable name are kept the same as before. Corresponding command-line flag is named, and description are written, after the existing flag for gRPC.

The scope of this change is intentionally limited to the minimum. That is, I have not yet added `--probe=false`, because it shouldn't be a blocker if we can change the port number.

### DIFF
--- a/cmd/tiller/tiller.go
+++ b/cmd/tiller/tiller.go
@@ -67,7 +67,6 @@ const (
 	storageConfigMap = "configmap"
 	storageSecret    = "secret"
 
-	probeAddr = ":44135"
 	traceAddr = ":44136"
 
 	// defaultMaxHistory sets the maximum number of releases to 0: unlimited
@@ -76,6 +75,7 @@ const (
 
 var (
 	grpcAddr             = flag.String("listen", ":44134", "address:port to listen on")
+	probeAddr            = flag.String("probe-listen", ":44135", "address:port to listen on for probes")
 	enableTracing        = flag.Bool("trace", false, "enable rpc tracing")
 	store                = flag.String("storage", storageConfigMap, "storage driver to use. One of 'configmap', 'memory', or 'secret'")
 	remoteReleaseModules = flag.Bool("experimental-release", false, "enable experimental release modules")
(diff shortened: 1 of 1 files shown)


bacongobbler:
Hey @mumoshu! #3440 appears to be a similar PR that addresses the same issue. Would you mind taking a look at that one and see if it addresses this issue?

The only issue I foresee with this is how Helm fetches the correct port number from the pod. [pkg/helm/portforwarder hardcodes the port value to 44134](https://github.com/helm/helm/blob/e14a0378fd156d39ecb4ebaf188641f5bc54d5ed/pkg/helm/portforwarder/portforwarder.go#L42), so the if someone were to change the listening address, the client would be unable to communicate with tiller unless it was specified through HELM_HOST. Is there a way to make that more flexible?

mumoshu:
@bacongobbler Hey! Thanks for the quick reply.

Actually, I don't care about the portforwarder for my specific use-case, which is to concuerrently run multiple [helm-tiller](https://github.com/rimusz/helm-tiller) managed local tillers via  [helmfile](https://github.com/roboll/helmfile/pull/531). It doesn't rely on portforwarding.

#3440 has more coverage than this PR. I'd say it's too much for my use-case. The smaller the better, as long as it doesn't break anything and still serves some use-case?  😃 

So - IMHO this PR is complete as-is.

But I'm also eager to contribute fixes for portforwarder port and the probe toggle in separate PRs!

bacongobbler (approved):
Sounds good to me! This does solve a particular use case even without the changes to portforwarder, so this is good to go. Thanks for the PR!

mumoshu:
@bacongobbler Thanks a lot for your efforts on maintaining helm, as always!

I've submitted #5561 to add `--probe=[true|false]` as a follow-up for this PR.

Would you mind if I created issues for fixing the portforwarder and init, even though I don't have a specific use-case for those myself?

bacongobbler:
Yes, if you wouldn't mind opening new issues for those particular action items that would be fantastic. Thank you as always! :)

mumoshu:
Thanks for confirming! Opened #5563 for the portforwarder improvement.

mumoshu:
See #5564 for helm init.

PR #5576, opened 2019-04-10
fix(helm): Fix manifest validation

mortent:
Manifest validation is done by the builder, but it requires that the schema is set before the `Stream` function is called. Otherwise the StreamVisitor is created without a schema and no validation is done.
This PR moves the call to `Schema` before the call to `Stream`, so the `StreamVisitor` is created with the correct schema.

This is hard to unit test since the `TestFactory` used by the client unit tests has the schema hardcoded to use the `NullSchema`.

**If applicable**:

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -142,8 +142,8 @@ func (c *Client) BuildUnstructured(namespace string, reader io.Reader) (Result, 
 		ContinueOnError().
 		NamespaceParam(namespace).
 		DefaultNamespace().
-		Stream(reader, "").
 		Schema(c.validator()).
+		Stream(reader, "").
 		Flatten().
 		Do().Infos()
 	return result, scrubValidationError(err)


hickeyma (approved):
LGTM, thanks.

hickeyma:
@bacongobbler Are you ok with this?

bacongobbler:
I need to take a closer look at this one. While it's a small change, we need to investigate its behaviour more closely before we merge.

bacongobbler:
After looking at cli-runtime, this [seems to check out](https://github.com/kubernetes/cli-runtime/blob/11d55751678de1d864dcd909ed061925eef48bd5/pkg/resource/builder.go#L352). LGTM!

bacongobbler (approved):

distorhead:
@mortent @bacongobbler There is compatibility problem with this change: https://github.com/helm/helm/issues/5640

PR #5143, opened 2019-01-09
Fixes #3338: Implement "atomic" upgrade/install

cleac:
**What this PR does / why we need it**:

Add possibility to put "--safe" argument to install and upgrade command that restores the state of cluster in case of failed install/upgrade attempt. 

This resolves the problem of non-atomic upgrade process in helm and removes need to write scripts to handle failed deployments. Also, closes #3338.

**Special notes for your reviewer**:

This is a proposal for such functionality, so I haven't any documentation and unit tests yet. However I've tested this functionality by hands on living cluster.

Also, I am not a professional Go developer, so if there are ways to improve implementation with fluent Go cunstructions, I would be glad for suggestions.

**If applicable**:
- [x] this PR contains documentation
- [x] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

PR #5579, opened 2019-04-10
Add --cleanup-on-fail support into --atomic: use CleanupOnFail param in rollback (optionally)

distorhead:
As suggested in https://github.com/helm/helm/pull/5143#issuecomment-469419200.


thomastaylor312 (approved):
Oops. This was my miss. I thought I saw this in your original PR, but maybe it was only in rollback. Thanks for catching this!

PR #5597, opened 2019-04-13
Helm 3: registry login/logout

### LINKED ISSUE #5461: Support Helm login for V3
Currently you can push a helm chart to an OCI registry in V3. 
We need to support a unified way of storing credentials so that the user name and passwords can be used during push to an OCI registry. 

Output of `helm version`:
version.BuildInfo{Version:"v3.0+unreleased", GitCommit:"a32f8ebb377da054d21510671d63729fe6a00f3a", GitTreeState:"clean"}

/cc @jdolitsky @shizhMSFT

jdolitsky:
Resolves #5461.

This leverages a new version of [oras](https://github.com/deislabs/oras) to add new Helm commands for logging into/out of a registry. I decided to keep everything under `helm chart` subcommand for now to keep all the registry work consistent (`helm chart login`, `helm chart logout`). We should have a planning session soon to discuss where all of the registry-related commands should end up.

Upon successful login, a new entry is placed in `$HELM_HOME/registry/config.json`, which is formatted and functions exactly like Docker CLI, including support for Docker credentials helpers such as `osxkeychain` (oras pulls in Docker as a library):
```
$ cat ~/.helm/registry/config.json
{
	"auths": {
		"localhost:5000": {}
	},
	"credsStore": "osxkeychain"
}
```

cc @sajayantony @stevelasker @yuwaMSFT2 @shizhMSFT @jzelinskie @stevvooe

### Example

*Note: `h3` in commands below refers to binary built from this change*

First, start a registry with auth enabled:
```
$ htpasswd -cB -b auth.htpasswd myuser mypass
$ docker run -it --rm -p 5000:5000 \
    -v $(pwd)/auth.htpasswd:/etc/docker/registry/auth.htpasswd \
    -e REGISTRY_AUTH="{htpasswd: {realm: localhost, path: /etc/docker/registry/auth.htpasswd}}" \
    registry
``` 

Next, login using the new command (manually enter password):
```
$ h3 chart login -u myuser localhost:5000
Password:
Login succeeded
```

Save a chart in the local cache:
```
$ helm fetch stable/postgresql --untar
$ h3 chart save postgresql/ localhost:5000/mycharts/postgresql:latest
Name: postgresql
Version: 3.15.0
Meta: sha256:b32d02932fc66e26248c9f589ddb0667d91597db6800b05c0edc0ccb2eab1179
Content: sha256:2bfdce314c6de5fe58dc88d5d8785a2461fcc61ccc5ee640a3e502d59c352a83
latest: saved
```

Push chart to the registry:
```
$ h3 chart push localhost:5000/mycharts/postgresql:latest
The push refers to repository [localhost:5000/mycharts/postgresql]
Name: postgresql
Version: 3.15.0
Meta: sha256:b32d02932fc66e26248c9f589ddb0667d91597db6800b05c0edc0ccb2eab1179
Content: sha256:2bfdce314c6de5fe58dc88d5d8785a2461fcc61ccc5ee640a3e502d59c352a83
latest: pushed to remote (2 layers, 16.4 KiB total)
```

Logout of registry:
```
$ h3 chart logout localhost:5000
Logout succeeded
```

Try pushing again - DENIED!
```
$ h3 chart push  localhost:5000/mycharts/postgresql:latest
The push refers to repository [localhost:5000/mycharts/postgresql]
Name: postgresql
Version: 3.15.0
Meta: sha256:b32d02932fc66e26248c9f589ddb0667d91597db6800b05c0edc0ccb2eab1179
Content: sha256:2bfdce314c6de5fe58dc88d5d8785a2461fcc61ccc5ee640a3e502d59c352a83
Error: unexpected response: 401 Unauthorized
```


Gopkg.toml
```diff
@@ -44,7 +44,7 @@
 
 [[constraint]]
   name = "github.com/deislabs/oras"
-  version = "~0.3.3"
+  revision = "12b336a3ea9d5d048f6726adbfc0a8271e75e2a8"
```
bacongobbler: is there plans to release a v0.4.0? If so should we wait for that to be released before merging?

cmd/helm/root.go
```diff
@@ -68,10 +70,17 @@ func newRootCmd(actionConfig *action.Configuration, out io.Writer, args []string
 
 	// Add the registry client based on settings
 	// TODO: Move this elsewhere (first, settings.Init() must move)
+	// TODO: handle errors
```
bacongobbler: Would you mind refactoring this to capture these errors? Perhaps this is a good time to call `panic()`. @adamreese any thoughts how we could better handle client initialization like we do in `newActionConfig`?

bacongobbler (approved):
Looks good for a first pass!

cmd/helm/chart_login.go
```diff
@@ -0,0 +1,134 @@
...
+	f := cmd.Flags()
+	f.StringVarP(&usernameOpt, "username", "u", "", "registry username")
+	f.StringVarP(&passwordOpt, "password", "p", "", "registry password or identity token")
+	f.BoolVarP(&passwordFromStdinOpt, "password-stdin", "", false, "read password or identity token from stdin")
+
+	return cmd
+}
+
+// Adapted from https://github.com/deislabs/oras
+func getUsernamePassword(usernameOpt string, passwordOpt string, passwordFromStdinOpt bool) (string, string, error) {
```
yuwaMSFT2: just one usability suggestion: Docker cli recently had a change that if user tries to do docker login myregistry.io without user name and password, it first tries to read from the credential store and use the stored one for login. This comes very handy. Will helm login support the same?

cmd/helm/root.go
```diff
@@ -68,10 +70,17 @@ func newRootCmd(actionConfig *action.Configuration, out io.Writer, args []string
...
+	credentialsFile := filepath.Join(settings.Home.Registry(), registry.CredentialsFileBasename)
+	client, _ := auth.NewClient(credentialsFile)
+	resolver, _ := client.Resolver(context.Background())
```
yuwaMSFT2: It seems currently oras resolver only has default options. Do we plan to have a way to configure different resolver options, like PlainHTTP (for example running a test registry on HTTP but not locahost)?

Gopkg.toml
jdolitsky: We can certainly hold off, should be a few days.

cmd/helm/chart_login.go
jdolitsky: im not sure i understand - what is the "stored one", and why log into something that youve already authed against?

cmd/helm/root.go
jdolitsky: Im happy to use panic for now. The method signature just did allow for return of error object

pkg/registry/client_test.go
```diff
@@ -167,6 +206,15 @@ func (suite *RegistryClientTestSuite) Test_5_RemoveChart() {
 	suite.Nil(err)
 }
 
+func (suite *RegistryClientTestSuite) Test_7_Logout() {
+	// TODO: re-enable once fixed in oras (no error thrown on linux)
+	// err := suite.RegistryClient.Logout("this-host-aint-real-son:5000")
+	// suite.NotNil(err, "error logging out of registry that has no entry")
```
jdolitsky: we should be able to re-enable this now. Ill make the change once we move to a 0.4.0

cmd/helm/root.go
jdolitsky: I'd be hesitant to encourage any use of http from the onset. This can of course come later if needed

cmd/helm/chart_login.go
yuwaMSFT2: The first time you do log in, the credential is stored; the next time you do login without username/pwd, the stored credential is retrieved and used for login.

The login in certain systems has certain significance. For example, when you do login, a token may be retrieved which is used in subsequent requests. But the token may expire after some time. Doing login again basically retrieve a new token.

Or server can do some special logic on login, say refreshing cached permission, etc.

Or I simply forgot if I did login already:)

cmd/helm/root.go
yuwaMSFT2: I agree with you. The point is not the PlainHTTP support. Just point out the possibility that user may want to extend the default resolver, which resolves the host name of the server, schema, and auth.

I understand it's not necessary for now; just brought it up so we think about possibility of extension (probably plugin model).

Gopkg.toml
jdolitsky: done

pkg/registry/client_test.go
jdolitsky: done

jdolitsky:
Ok, updated to released oras version (0.4.0). Should be ok to merge.

@yuwaMSFT2 in regards to the other asks - in the interest in getting this merged, can you add something to the issue queue for us to continue this discussion? Then if it makes sense we can work on a follow-up

cmd/helm/root.go
jdolitsky: I've added `panic`s in latest commit, with a TODO to return the error when possible

yuwaMSFT2:
@jdolitsky Sure please go ahead with the merge. I can open separate issues.

SteveLasker:
looks great Josh. 
For the UX, knowing this is a placeholder, I’m wondering about
helm login
helm repo login
helm chart login feels like i’m logging into a chart, whatever that might mean

Great to see all the progress

bacongobbler:
for now, we're moving all of the OCI integration parts into `helm chart *` commands. We'll eventually move them into the mainline CLI (i.e. `helm login`) once we figure out the migration plan from chart repositories to OCI repositories.

SteveLasker:
Makes sense. I thought I remembered we came to a solution where both would work, but I'm sure there's some important details here to call out.

jdolitsky:
I've gone ahead and moved these into a new `helm registry` subcommand (based on some side convo)

adamreese (approved):
LGTM

@jdolitsky please squash before merging

PR #5601, opened 2019-04-15
ref(pkg/chartutil): remove k8s version object dependency

adamreese:
Flattens the `.Capabilities` built-in and removes useless kubernetes
runtime metadata.

### DIFF
--- a/docs/chart_template_guide/builtin_objects.md
+++ b/docs/chart_template_guide/builtin_objects.md
@@ -19,7 +19,10 @@ In the previous section, we use `{{.Release.Name}}` to insert the name of a rele
 - `Capabilities`: This provides information about what capabilities the Kubernetes cluster supports.
   - `Capabilities.APIVersions` is a set of versions.
   - `Capabilities.APIVersions.Has $version` indicates whether a version (`batch/v1`) is enabled on the cluster.
-  - `Capabilities.KubeVersion` provides a way to look up the Kubernetes version. It has the following values: `Major`, `Minor`, `GitVersion`, `GitCommit`, `GitTreeState`, `BuildDate`, `GoVersion`, `Compiler`, and `Platform`.
+  - `Capabilities.Kube.Version` is the Kubernetes version.
+  - `Capabilities.Kube` is a short form for Kubernetes version.
+  - `Capabilities.Kube.Major` is the Kubernetes major version.
+  - `Capabilities.Kube.Minor` is the Kubernetes minor version.
 - `Template`: Contains information about the current template that is being executed
   - `Name`: A namespaced filepath to the current template (e.g. `mychart/templates/mytemplate.yaml`)
   - `BasePath`: The namespaced path to the templates directory of the current chart (e.g. `mychart/templates`).
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -25,6 +25,11 @@ In Helm 2, in order to purge the release ledger, the `--purge` flag had to be pr
 functionality is now enabled by default. To retain the previous behaviour, use
 `helm uninstall --keep-history`.
 
+### Capabilities
+
+Capabilities built-in has been simplified.
+
+[Built-in Objects](chart_template_guide/builtin_objects.md)
 
 ## Installing
 
--- a/pkg/chartutil/capabilities.go
+++ b/pkg/chartutil/capabilities.go
@@ -16,13 +16,6 @@ limitations under the License.
 package chartutil
 
 import (
-	"encoding/json"
-	"fmt"
-	"runtime"
-	"sort"
-
-	"k8s.io/apimachinery/pkg/version"
-
 	"k8s.io/client-go/kubernetes/scheme"
 )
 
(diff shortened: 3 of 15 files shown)


docs/faq.md
```diff
@@ -15,6 +15,11 @@ Here's an exhaustive list of all the major changes introduced in Helm 3.
 In Helm 3, Helm switched the Go import path over from `k8s.io/helm` to `helm.sh/helm`. If you intend
 to upgrade to the Helm 3 Go client libraries, make sure to change your import paths.
 
+### Capabilities
+
+Capabilities built-in has been simplified.
```
bacongobbler: Would you mind documenting what fields changed, and to what? e.g. `.Capabilities.KubeVersion` to `.Capabilities.Version`. This is a major breaking change to existing charts so we should document the changes required to existing charts.

pkg/chartutil/testdata/subpop/charts/subchart1/templates/service.yaml
```diff
@@ -5,9 +5,9 @@ metadata:
   labels:
     chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
     release-name: "{{ .Release.Name }}"
-    kube-version/major: "{{ .Capabilities.KubeVersion.Major }}"
-    kube-version/minor: "{{ .Capabilities.KubeVersion.Minor }}"
-    kube-version/gitversion: "v{{ .Capabilities.KubeVersion.Major }}.{{ .Capabilities.KubeVersion.Minor }}.0"
+    kube-version/major: "{{ .Capabilities.Major }}"
+    kube-version/minor: "{{ .Capabilities.Minor }}"
+    kube-version/gitversion: "v{{ .Capabilities.Major }}.{{ .Capabilities.Minor }}.0"
```
mattfarina: If this isn't really the `gitversion` should we even include the label?

adamreese: this is just testdata but I updated it for you @mattfarina

docs/faq.md
adamreese: updated to be backwards compatible

docs/chart_template_guide/builtin_objects.md
```diff
@@ -19,7 +19,9 @@ In the previous section, we use `{{.Release.Name}}` to insert the name of a rele
 - `Capabilities`: This provides information about what capabilities the Kubernetes cluster supports.
   - `Capabilities.APIVersions` is a set of versions.
   - `Capabilities.APIVersions.Has $version` indicates whether a version (`batch/v1`) is enabled on the cluster.
-  - `Capabilities.KubeVersion` provides a way to look up the Kubernetes version. It has the following values: `Major`, `Minor`, `GitVersion`, `GitCommit`, `GitTreeState`, `BuildDate`, `GoVersion`, `Compiler`, and `Platform`.
+  - `Capabilities.Version` is the Kubernetes version.
+  - `Capabilities.Major` is the Kubernetes major version.
+  - `Capabilities.Minor` is the Kubernetes minor version.
```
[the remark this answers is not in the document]
adamreese: updated to be backwards compatible

docs/faq.md
bacongobbler: Some of the field names did change (for the better) so we should probably still list out what fields changed, and where. For example:

- `.Capabilities.KubeVersion.GitVersion` -> `.Capabilities.KubeVersion.Version`

I don't believe the platform/runtime fields were flattened either. e.g.

- `.Capabilities.KubeVersion.Platform` -> `.Capabilities.Platform`

Are you planning to add those back?

hickeyma: I agree with @bacongobbler that it would be good to show the changes. Easy for user to see then.

adamreese: Platform was a garbage field that was sentenced to collection.

Some fields we never populated with relevant information.  I deleted all the no-op fields.

@bacongobbler `GitVersion` still returns the same value.  `.Capabilities.KubeVersion.GitVersion == .Capabilities.KubeVersion.Version`

pkg/chartutil/capabilities.go
```diff
@@ -16,89 +16,66 @@ limitations under the License.
...
+	Minor   string // Kubernetes minor version
 }
 
+// String implements fmt.Stringer
+func (kv *KubeVersion) String() string { return kv.Version }
+
+// GitVersion returns the Kubernetes version string.
+//
+// Depreciated: use KubeVersion.Version.
+func (kv *KubeVersion) GitVersion() string { return kv.Version }
```
bacongobbler: oh! I see now. Awesome!

pkg/chartutil/capabilities.go
```diff
@@ -16,89 +16,66 @@ limitations under the License.
...
+	Major   string // Kubernetes major version
+	Minor   string // Kubernetes minor version
 }
 
+// String implements fmt.Stringer
+func (kv *KubeVersion) String() string { return kv.Version }
+
+// GitVersion returns the Kubernetes version string.
+//
+// Depreciated: use KubeVersion.Version.
```
bacongobbler: small FYI: the correct term is "[deprecated](https://en.wikipedia.org/wiki/Deprecation)". :)

bacongobbler (approved):

pkg/chartutil/capabilities.go
```diff
@@ -16,89 +16,67 @@ limitations under the License.
...
+	Major   string // Kubernetes major version
+	Minor   string // Kubernetes minor version
 }
 
+// String implements fmt.Stringer
+func (kv *KubeVersion) String() string { return kv.Version }
+
+// GitVersion returns the Kubernetes version string.
+//
+// Deprecated: use KubeVersion.Version.
```
mattfarina: `KubeVersion.Version` isn't documented in the `builtin_objects.md` file. Can that be added alongside `Major` and `Minor`?

pkg/chartutil/capabilities.go
```diff
@@ -16,89 +16,67 @@ limitations under the License.
...
-	vs := make(VersionSet)
-	for _, v := range apiVersions {
-		vs[v] = struct{}{}
-	}
-	return vs
+// KubeVersion is the Kubernetes version.
+type KubeVersion struct {
+	Version string // Kubernetes version
+	Major   string // Kubernetes major version
+	Minor   string // Kubernetes minor version
```
mattfarina: This is where we get into the ugliness of versions and how they are misused.

I have a gke test cluster with a version of `1.12.7-gke.10`. It's the reason you can see things like `if semverCompare "^1.7-0" .Capabilities.KubeVersion.GitVersion` all over the charts repo. The `-0` on the end tells semver to look at pre-releases. gke is misusing the pre-release part of semver.

Anything after minor is no longer available. On one hand, that will simplify a what many chart authors have to do. On the other hand, do we know of anywhere the lost information will have an impact?

I just wanted to ask the question and document this.

mattfarina (changes requested):
I tried to use this branch to install a configmap into a cluster to see the values for `GitVersion` and `Version`. The cluster was 1.12. The values for both were 1.14.0. It looks like it's using the default values rather than the actual values for the cluster. We need the cluster values for this to be useful.

mattfarina (approved):
@adamreese This looks good but needs a rebase. My only outstanding issue with this was fixed via the other PR that caused the need for the rebase.

PR #5603, opened 2019-04-16
Update install.md

dieutth:
*From Script* should be in the same level as other *Installing the Helm Client* options.


bacongobbler (approved):

bacongobbler:
good catch. Thanks!

PR #5608, opened 2019-04-17
Using constants rather than hard coding

SataQiu:
Using constants rather than hard coding.
This is just the beginning.
If you agree with my idea, I will continue this work to reduce the hard coding.

**If applicable**:

### DIFF
--- a/cmd/helm/installer/install.go
+++ b/cmd/helm/installer/install.go
@@ -36,6 +36,7 @@ import (
 	"k8s.io/helm/pkg/version"
 
 	"k8s.io/helm/pkg/chartutil"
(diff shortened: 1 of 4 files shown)


bacongobbler:
I don't like the idea of adding a separate `constants` package. Can you please use `pkg/tiller/environment` for the default tiller port number much like how we consume [the default tiller namespace](https://github.com/helm/helm/blob/870e21d08897159d04c080ecd7f349a4fcfc4ec6/pkg/tiller/environment/environment.go#L41)?

SataQiu:
That's good! Let me try.

bacongobbler (approved):

PR #5611, opened 2019-04-17
Performance: Batch kube API operations

cgetzen:
Signed-off-by: Charlie Getzen <charlie.getzen@procore.com>

This cuts down tiller's time creating resources. In our tests, we create ~180 resources in a release. Before this change it took approximately 120 seconds to create these resources. It now takes around 45 seconds.

**If applicable**:


bacongobbler (commented):
It's unclear to me if we may run into race conditions with multiple resources being updated at the same time. Have you tested this for race conditions, and if so, did you run into any limitations with this change?

Thanks!

cgetzen:
Just ran through a few tests:
- respecting WaitUntilCRDEstablished
- deletes with missing resources
- upgrades with expected errors
- upgrades with successes

I don't know of any specific edge cases that seem likely to break. We have quite a few mutating and admission webhooks that this seems to play nicely with.

cgetzen:
I see that there is an InstallOrder that this doesn't necessarily respect. Instead of creating all resources at once, I will make a modification to batch them by resource type.

cgetzen:
Tests pass and logs look promising:
non-batching (100% parallel): https://gist.github.com/cgetzen/77feaff82d203d787fe660082089db84
batching by kind: https://gist.github.com/cgetzen/9e48cb7a7114b1ad2cbdee5eaffb6f28

Obviously the performance boost isn't as good as 100% parallelization, but there is still a significant speed increase compared to creating resources sequentially.

cgetzen:
@bacongobbler -- was install order the race condition you were thinking of?

bacongobbler:
yes. Batching by resource type should be safe.

cgetzen:
Is there anything else you'd like to see on this to move forward?

pkg/kube/client.go
```diff
@@ -550,14 +550,40 @@ func perform(infos Result, fn ResourceActorFunc) error {
...
+		if err != nil {
 			return err
 		}
 	}
 	return nil
 }
 
+func batchPerform(infos Result, fn ResourceActorFunc, errs chan<- error) {
+	finished := make(chan bool, 10000)
+	kind := infos[0].Object.GetObjectKind().GroupVersionKind().Kind
```
bacongobbler: Can we add a length check prior to this call? If `infos` is an empty list, this will result in a `panic`. We should probably `return` early in the case where there's no `infos` to process.

bacongobbler (changes requested):

bacongobbler:
Hey @cgetzen, sorry for not getting back to you sooner.

pkg/kube/client.go
cgetzen: Good call. Although `func perform` checks before calling batchPerform, this will be appropriate if it is ever reused.

bacongobbler (approved):
Thanks. This LGTM, though I'd like another maintainer to have a look here just in case I missed something.

pkg/kube/client.go
```diff
@@ -550,14 +550,44 @@ func perform(infos Result, fn ResourceActorFunc) error {
...
+	counter := 0
+	for _, info := range infos {
+		currentKind := info.Object.GetObjectKind().GroupVersionKind().Kind
+		if kind != currentKind {
+			// Wait until the previous kind has finished
+			for i := 0; i < counter; i++ {
+				<-finished
+			}
```
thomastaylor312: This logic should be changed to use `sync.WaitGroup` instead of using a channel and the counter

pkg/kube/client.go
```diff
@@ -550,14 +550,44 @@ func perform(infos Result, fn ResourceActorFunc) error {
 		return ErrNoObjectsVisited
 	}
 
-	for _, info := range infos {
-		if err := fn(info); err != nil {
+	errs := make(chan error)
```
thomastaylor312: Nit: You could save the use of an errors channel by making the call to `batchPerform` blocking and then just returning a slice of errors from `batchPerform`. It would probably be less bug prone in the end, although the tradeoff would be that you'd need a mutex on the error slice in the function

thomastaylor312 (changes requested):
This change will be super nice, thanks for the PR! There are a few code things that need to be changed, namely the need to use `sync.WaitGroup`

thomastaylor312 (approved):
Looking great! Thank you for the quick response. This is definitely faster than before

thomastaylor312 (changes requested):
Sorry, just realized there is one failing unit test. This is good to go once that is resolved

cgetzen:
Thanks for the notes!
> You could save the use of an errors channel by making the call to batchPerform.

One quality of the current implementation is that if there is an error the next batch will not begin, because the error channel has no buffer. I'm not sure how to replicate that cleanly using slices.

> failing unit test

I've been unable to replicate this failure locally, merging master may have solved it. Please advise if more work is needed.

thomastaylor312 (approved):
@cgetzen looks like that did fix it. Thanks!

And seeing as the errors is a nit and you have a good point about interrupting on error, there are no worries from my end

pkg/kube/client.go
```diff
@@ -616,14 +617,35 @@ func perform(infos Result, fn ResourceActorFunc) error {
...
 	}
 	return nil
 }
 
+func batchPerform(infos Result, fn ResourceActorFunc, errs chan<- error) {
+	var kind string
+	var wg sync.WaitGroup
+	for _, info := range infos {
+		currentKind := info.Object.GetObjectKind().GroupVersionKind().Kind
+		if kind != currentKind {
```
bacongobbler: won't we hit a blocker right here? At this point,

- `kind` is set to the empty string
- `currentKind` is set to the current object's kind

From what I can tell, there's no way to break out of this. We need a base condition like we did before with the `infos[0].Object.GetObjectKind().GroupVersionKind().Kind`, don't we?

bacongobbler (approved):

PR #5620, opened 2019-04-19
remove snowflake-emoji 'Happy Helming!'

ecv:
Signed-off-by: Evan Vetere <ecv@sine.com>

This PR removes the pushy, chirpy good-mood-enforcement from Helm's standard output. 

Helm has greatly matured as a project and having snowflake emojis in deployment pipelines that fire hundreds of times a week (and are thus the opposite of snowflakes) feels incorrect, as well as potentially unprofessional. Also, I suspect I'm not the only one who doesn't enjoy being told how to feel. 

Just throwing this out here on a Friday at 6pm, after having spent an hour fixing my institution's deployments ahead of the weekend, while the software continually told me to cheer up about it. 

I'll understand if you guys are super attached to the chirpy messages, but please note that this is a culturally variable thing that just doesn't fly everywhere. I will of course keep enjoying Helm and the work y'all do, if you deny this PR. 🤒

### DIFF
--- a/cmd/helm/init.go
+++ b/cmd/helm/init.go
@@ -310,7 +310,6 @@ func (i *initCmd) run() error {
 			"$ helm init --tiller-image gcr.io/kubernetes-helm/tiller:v2.8.2\n\n")
 	}
 
-	fmt.Fprintln(i.out, "Happy Helming!")
 	return nil
 }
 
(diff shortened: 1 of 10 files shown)


ecv:
I may need another commit here, I think some of the bits I left are actually unit tests.

ecv:
Pardon for the forces, I don't know how to run the integration tests without a PR! 😬

hickeyma (commented):
@ecv Not sure I agree on the need for this PR. However, going to throw it out to other maintainers for their opinions.

/cc @bacongobbler @adamreese

bacongobbler (approved):
I can understand why this may be upsetting. Since it isn't informational, let's stay on the cautious side and remove it.

PR #5635, opened 2019-04-23
WIP: integrate `helm chart *` commands into helm

bacongobbler:
This PR is an attempt to integrate `helm chart *` commands into existing helm commands, such as `helm pull` and `helm package`.

TODO:

- [ ] needs migration docs
- [ ] fix integration tests that relied on `srv.CopyCharts()`
- [ ] figure out a migration path from helm 2 chart repositories to OCI registries.


bacongobbler:
after thinking it through, this is not the design we want to go forward with as this removes support for chart repositories. Closing.

PR #5641, opened 2019-04-24
add bootstrap target when running make/make all

tariq1890:
For better DX.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ BINARIES  := helm tiller
 SHELL=/usr/bin/env bash
 
 .PHONY: all
-all: build
+all: bootstrap build
 
 .PHONY: build
 build:
--- a/docs/developers.md
+++ b/docs/developers.md
@@ -16,7 +16,7 @@ Helm and Tiller.
 We use Make to build our programs. The simplest way to get started is:
 
 ```console
-$ make bootstrap build
+$ make
 ```
 
 NOTE: This will fail if not running from the path `$GOPATH/src/k8s.io/helm`. The


tariq1890:
Looks like a CircleCI. Requesting a re-run :).

hickeyma (changes requested):
This is looking good to me, thanks. Just two small nits:
- Need an update on make in [install](https://github.com/helm/helm/blob/master/docs/install.md#from-source-linux-macos)
- In developers doc: "This will build both Helm and Tiller. make bootstrap will attempt to install certain tools if they are missing." Needs an update to something like "This will build both Helm and Tiller, and will attempt to install certain tools if they are missing."

bacongobbler:
We intentionally did not add make bootstrap to the default make target because it's a one-off target. It doesn't need to be called every time you rebuild the client.

In dev-v3 we added it there because we have smarter detection for when make bootstrap was called and dependencies an update. If we had the same thing in place here then I'd be open to consist this.

hickeyma:
@bacongobbler I had the same thinking.

I thought about this a bit more and a there are a number of usability errors caused by not calling `make bootstrap build`. I then thought the user will not want it to be called all the time but for this they could call `make build`. I suppose doing the `bootstrap` each time for a `make` call is inconvenient but does it balance out in favour of  dependency issues?

tariq1890:
I agree with @hickeyma on this. By instinct I would call `make` or `make all` when I run a project for the first time. Any subsequent rebuilds would be done by a `make build`. That's pretty much the convention if I am not mistaken.

Issue #5639, opened 2019-04-24
Rollback on syntax error does not work on the first try, works on retry


distorhead:
Rollback on syntax error does not work on the first try, works on retry. Example to reproduce below.

```
$ git clone https://github.com/distorhead/ex-helm-upgrade-release-storage-corruption.git
$ cd ex-helm-upgrade-release-storage-corruption
```

Installing chart first time:

```
$ git checkout good
$ helm install . --name example --namespace example
NAME:   example
LAST DEPLOYED: Wed Apr 24 14:50:07 2019
NAMESPACE: example
STATUS: DEPLOYED

RESOURCES:
==> v1/Pod(related)
NAME                       READY  STATUS   RESTARTS  AGE
mydeploy-58c55ff786-52dsn  0/1    Pending  0         0s

==> v1beta1/Deployment
NAME      READY  UP-TO-DATE  AVAILABLE  AGE
mydeploy  0/1    0           0          0s
```

Introduce syntax error into template:

```
$ git checkout bad
$ git diff good
diff --git a/templates/01-example.yaml b/templates/01-example.yaml
index 7616296..3c6d715 100644
--- a/templates/01-example.yaml
+++ b/templates/01-example.yaml
@@ -12,6 +12,6 @@ spec:
         service: mydeploy
     spec:
       containers:
-      - name: main
+      - nam: main
         image: ubuntu:18.04
         command: [ "/bin/bash", "-c", "while true; do date ; sleep 1 ; done" ]
```

Upgrade release:

```
$ helm upgrade example .
UPGRADE FAILED
ROLLING BACK
Error: failed to create patch: map: map[command:[/bin/bash -c while true; do date ; sleep 1 ; done] image:ubuntu:18.04 nam:main] does not contain declared merge key: name
Error: UPGRADE FAILED: failed to create patch: map: map[command:[/bin/bash -c while true; do date ; sleep 1 ; done] image:ubuntu:18.04 nam:main] does not contain declared merge key: name
```

Now **important part**, first rollback:

```
$ helm history example
REVISION	UPDATED                 	STATUS  	CHART                                           	DESCRIPTION
1       	Wed Apr 24 15:53:24 2019	DEPLOYED	ex-helm-upgrade-release-storage-corruption-0.1.0	Install complete
2       	Wed Apr 24 15:53:29 2019	FAILED  	ex-helm-upgrade-release-storage-corruption-0.1.0	Upgrade "example" failed: failed to create patch: map: ma...
[...]

PR #5643, opened 2019-04-25
fix(helm): Only validate new manifests

### LINKED ISSUE #5640: Broken compatibility because of mandatory validation on BuildUnstructured
Validation on BuildUnstructured was introduced in the https://github.com/helm/helm/pull/5576.

Firstly, this is very important fix, thanks @mortent .

But we have encountered problems with compatibility trying to run `helm upgrade` on already installed release in the `DEPLOYED` state.

The error was:

```
$ helm get RELEASE
...
Error: getting release status failed: error validating "": error validating data: ValidationError(CronJob.spec): unknown field "activeDeadlineSeconds" in io.k8s.api.batch.v1beta1.CronJobSpec
```

The error occured because there is indeed an error in the chart template: `CronJob` does not have `activeDeadlineSeconds` field, but it was defined in the manifest. The **important thing** is that it is impossible to fix the issue simply by fixing error in the chart template and redeploying, because `helm upgrade` still fails after fix: validation error occurs when fetching release manifests from storage.

# How to reproduce

```
$ git clone https://github.com/distorhead/ex-helm-validation-incompatibility.git
$ cd ex-helm-validation-incompatibility
```

Installing chart first time with helm `2.13.1`. There is an error in chart: unknown `helloWorld: 1234` field defined for the `CronJob`. But helm 2.13.1 install release successfully:

```
$ git checkout bad
$ helm version
Client: &version.Version{SemVer:"v2.13.1", GitCommit:"618447cbf203d147601b4b9bd7f8c37a5d39fbb4", GitTreeState:"clean"}
$ helm install . --name example --namespace example
NAME:   example
LAST DEPLOYED: Wed Apr 24 19:16:09 2019
NAMESPACE: example
STATUS: DEPLOYED

RESOURCES:
==> v1beta1/CronJob
NAME      SCHEDULE    SUSPEND  ACTIVE  LAST SCHEDULE  AGE
cleaning  32 1 * * *  False    0       <none>         0s

$ helm get example
...
# Source: ex-helm-validation-incompatibility/templates/01-cronjob.yaml
apiVersion: batch/v1beta1
kind: CronJob
metadata:
  name: cleaning
spec:
...
  helloWorld: 1234
...
```

Bad resource manifest has been recorded into release revision.

Then let's try to upgrade release using new helm version (master):

```
$ HELM_HOST=localhost:44134 ~/go/src/k8s.io/helm/bin/helm upgrade example .
UPGRADE FAILED
Error: error validating "": error validating data: ValidationError(CronJob.spec): unknown field "helloWorld" in io.k8s.api.batch.v1beta1.CronJobSpec
Error: UPGRADE FAILED: error validating "": error validating data: ValidationError(CronJob.spec): unknown field "helloWorld" in io.k8s.api.batch.v1beta1.CronJobSpec
```

Delete unknown field and retry:

```
$ git checkout good
$ git diff bad
diff --git a/templates/01-cronjob.yaml b/templates/01-cronjob.yaml
index efd8d4d..7f91a65 100644
--- a/templates/01-cronjob.yaml
+++ b/templates/01-cronjob.yaml
@@ -6,7 +6,6 @@ spec:
   schedule: "32 1 * * *"
   successfulJobsHistoryLimit: 3
   failedJobsHistoryLimit: 1
-  helloWorld: 1234
   concurrencyPolicy: Forbid
   jobTemplate:
     spec:
$ HELM_HOST=localhost:44134 ~/go/src/k8s.io/helm/bin/helm upgrade example .
UPGRADE FAILED
Error: failed decoding reader into objects: error validating "": error validating data: ValidationError(CronJob.spec): unknown field "helloWorld" in io.k8s.api.batch.v1beta1.CronJobSpec
Error: UPGRADE FAILED: failed decoding reader into objects: error validating "": error validating data: ValidationError(CronJob.spec): unknown field "helloWorld" in io.k8s.api.batch.v1beta1.CronJobSpec
```

Still got same error.

The only known way to fix the problem is to downgrade tiller, fix manifest, upgrade release, then upgrade tiller.

Seems that validation should not be applied to manifests fetched from release storage.

mortent:
Validation of manifests in charts was fixed in #5576. But unfortunately this led to a situation where existing releases that doesn't validate can't be updated, even when the issue has been fixed in the new version of the chart (#5640).
I think the root of the problem here is that Helm combines validation and turning yaml manifests into infos into one function. There are many scenarios where turning manifests into infos should succeed, even if a manifest doesn't pass validation.
This PR adds a new function `Validate` to the KubeClient. This function does validation, but does not return infos. The existing `BuildUnstructured` function returns infos, but it does not do any validation. The `Validate` function is called from the release server to do validation of new charts while the `BuildUnstructured` function is used internally in the Client to turn manifests into infos for delete, update, install...

Fixes #5640 

@distorhead @bacongobbler 

**If applicable**:

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -141,7 +141,7 @@ func (c *Client) validator() validation.Schema {
 	return schema
 }
 
-// BuildUnstructured validates for Kubernetes objects and returns unstructured infos.
+// BuildUnstructured reads Kubernetes objects and returns unstructured infos.
 func (c *Client) BuildUnstructured(namespace string, reader io.Reader) (Result, error) {
 	var result Result
 
@@ -150,11 +150,24 @@ func (c *Client) BuildUnstructured(namespace string, reader io.Reader) (Result, 
 		ContinueOnError().
 		NamespaceParam(namespace).
 		DefaultNamespace().
-		Schema(c.validator()).
 		Stream(reader, "").
 		Flatten().
 		Do().Infos()
 	return result, scrubValidationError(err)
+}
+
+// Validate reads Kubernetes manifests and validates the content.
+func (c *Client) Validate(namespace string, reader io.Reader) error {
+	_, err := c.NewBuilder().
+		Unstructured().
+		ContinueOnError().
+		NamespaceParam(namespace).
+		DefaultNamespace().
+		Schema(c.validator()).
+		Stream(reader, "").
+		Flatten().
+		Do().Infos()
+	return scrubValidationError(err)
 }
 
 // Build validates for Kubernetes objects and returns resource Infos from a io.Reader.
--- a/pkg/tiller/environment/environment.go
+++ b/pkg/tiller/environment/environment.go
@@ -147,7 +147,19 @@ type KubeClient interface {
 	UpdateWithOptions(namespace string, originalReader, modifiedReader io.Reader, opts kube.UpdateOptions) error
 
 	Build(namespace string, reader io.Reader) (kube.Result, error)
+
+	// BuildUnstructured reads a stream of manifests from a reader and turns them into
+	// info objects. Manifests are not validated against the schema, but it will fail if
+	// any resoures types are not known by the apiserver.
+	//
+	// reader must contain a YAML stream (one or more YAML documents separated by "\n---\n").
 	BuildUnstructured(namespace string, reader io.Reader) (kube.Result, error)
+
+	// Validate reads a stream of manifests from a reader and validates them against
+	// the schema from the apiserver. It returns an error if any of the manifests does not validate.
+	//
+	// reader must contain a YAML stream (one or more YAML documents separated by "\n---\n").
+	Validate(namespace string, reader io.Reader) error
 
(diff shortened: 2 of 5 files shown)


hickeyma:
@mortent Should this fix #5639 also?

mortent:
@hickeyma I think so, but let me verify.

mortent:
After looking at this, I don't think this will completely fix the issue in #5639. This fix combined with #5576 will make it harder to get into the situation described in #5639 since more chart errors will be caught in validation rather than during the process of applying the update. But #5639 suggest there is an issue somewhere in the rollback logic.

bacongobbler (approved):
LGTM.

thomastaylor312:
After the refactor of KubeClient, we no longer use structured types and allow the k8s API to validate resources when sent, so marking this as port complete for v3

RoyRimjhim:
Hello Team, We are facing this issue in the newly released v3 as well but not in the beta version v3.0.0-beta.4. Kindly help with the resolution.

thomastaylor312:
@RoyRimjhim Could you please open a new issue with all the details of what you are seeing?

PR #5646, opened 2019-04-25
ref(pkg/kube): extract wait logic from install/update

adamreese:
This change adds a new method for waiting for kubernetes resources.

Also snuck in: use Duration flag types

### DIFF
--- a/pkg/action/rollback.go
+++ b/pkg/action/rollback.go
@@ -140,7 +140,7 @@ func (r *Rollback) performRollback(currentRelease, targetRelease *release.Releas
 
 	// pre-rollback hooks
 	if !r.DisableHooks {
-		if err := r.execHook(targetRelease.Hooks, targetRelease.Namespace, hooks.PreRollback); err != nil {
+		if err := r.execHook(targetRelease.Hooks, hooks.PreRollback); err != nil {
 			return targetRelease, err
 		}
 	} else {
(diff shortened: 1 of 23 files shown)


pkg/action/uninstall.go
```diff
@@ -220,15 +219,15 @@ func (u *Uninstall) deleteRelease(rel *release.Release) (kept string, errs []err
...
+		kept = summarizeKeptManifests(filesToKeep, u.cfg.KubeClient)
 	}
 
 	for _, file := range filesToDelete {
 		b := bytes.NewBufferString(strings.TrimSpace(file.Content))
 		if b.Len() == 0 {
 			continue
 		}
-		if err := u.cfg.KubeClient.Delete(rel.Namespace, b); err != nil {
+		if err := u.cfg.KubeClient.Delete(b); err != nil {
```
bacongobbler: is there a way we can refactor `.Wait` to accept a condition? That way we could wait for these resources to reach the deleted state. I've heard people ask about that.

Doesn't have to be done in this PR, just thought if you wondered if that's possible

bacongobbler (commented):
looks great! I think this is missing `--wait` functionality on `helm upgrade` and `helm rollback`, but once that's addressed we should be good to merge. I really love the refactor work done in `pkg/kube` here. :+1:

pkg/action/rollback.go
```diff
@@ -149,7 +149,8 @@ func (r *Rollback) performRollback(currentRelease, targetRelease *release.Releas
 
 	cr := bytes.NewBufferString(currentRelease.Manifest)
 	tr := bytes.NewBufferString(targetRelease.Manifest)
-	if err := r.cfg.KubeClient.Update(targetRelease.Namespace, cr, tr, r.Force, r.Recreate, r.Timeout, r.Wait); err != nil {
+	// TODO add wait
```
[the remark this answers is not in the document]
adamreese: I was going to add the wait logic back in a followup PR.  This just pulls out the wait param from the interface.

pkg/action/uninstall.go
adamreese: That is my plan!

bacongobbler (approved):
All right! The refactor here looks great. Let's get this in.

PR #5650, opened 2019-04-26
Add docs for configuring SSL using Terraform.

jbussdieker:

This adds documentation on using Terraform to generate Tiller certificates.

**If applicable**:


bacongobbler:
Hi @jbussdieker! Generally speaking we try to avoid adding third-party documentation on "using Helm with X". In practice, we provide links to external sites under the "Related Projects" section of the docs. This is to keep the documentation focused on Helm.

For example, in [the section about blogs, how-tos and extra documentation](https://helm.sh/docs/related/#article-blogs-how-tos-and-extra-documentation), there's multiple how-tos on using Helm with other projects like Wercker and GitLab.

Would you be willing to host this elsewhere and link it in the docs?

docs/tiller_ssl.md
```diff
@@ -42,6 +42,8 @@ on getting ready within a small amount of time. For production configurations,
 we urge readers to read [the official documentation](https://www.openssl.org) and
 consult other resources.
 
+Alternatively you can use Terraform to quickly create all the necessary certificates: [Generating Certificate Authorities and Certificates using Terraform](tiller_ssl_terraform.md).
```
hickeyma: Do you mind changing this line to: There are other alternative ways to generating SSL CAs in addition to `openssl`, for example Terraform.  They are not document here but you can find links to these alternative means in [Related Projects and Documentation](https://helm.sh/docs/related/).

hickeyma (changes requested):
We discussed this PR yesterday at the weekly dev meeting. We think the documentation is very good and we are very thankful for your efforts.  Similar to what @bacongobbler said in https://github.com/helm/helm/pull/5650#issuecomment-488768468, we avoid adding third-party documentation  to the Helm docs as we are unable to maintain them. 

If you are willing to host the documentation and you add the reference link to the [blogs, how-tos and extra documentation](https://helm.sh/docs/related/#article-blogs-how-tos-and-extra-documentation) section then we can merge the PR.

Let me know what you think.

jbussdieker:
Thanks for the feedback! I've updated the PR with the requested changes.

docs/tiller_ssl.md
```diff
@@ -42,6 +42,8 @@ on getting ready within a small amount of time. For production configurations,
 we urge readers to read [the official documentation](https://www.openssl.org) and
 consult other resources.
 
+There are other alternative ways to generating SSL CAs in addition to `openssl`, for example Terraform. They are not documented here but you can find links to these alternative means in Related Projects and Documentation.
```
hickeyma: Can you update these words as shown:
Related Projects and Documentation. --> [Related Projects and Documentation](https://helm.sh/docs/related/).

This is to add the link for where the doc is.

hickeyma (changes requested):
Thanks for making the changes. One small update to add the link to the related doc.

docs/tiller_ssl.md
jbussdieker: Good catch

hickeyma (approved):

PR #4913, opened 2018-11-13
chore(deps): update cobra to latest

glb:
Fixes #3669, #4891.

PR #5655, opened 2019-04-29
fix(helm): update Cobra to latest version

marckhouzam:
The version of Cobra used in dev-v3 is older than the one used
for the master branch (helm v2).  Although the dev-v3 branch
was based on master and therefore had the same Cobra version
originally, it was changed a couple of times to choose Cobra
tagged versions instead.

However, the latest Cobra tagged version is older than the version
of Cobra used on the master branch.  Therefore, some of the
improvements made to Cobra and used by helm v2 are not available
to helm v3 currently.

This commit brings Cobra to its latest available commit for
helm v3.  Originally, I thought of selecting the same Cobra version
in v3 as for helm v2, but after reviewing the 5 extra commits
added to Cobra, I feel it is safe and valuable to bring them in
also.

Bringing Cobra up-to-date is essential for upcoming work being
prepared for dynamic bash-completion; there are bug fixes in
Cobra that are necessary for dynamic bash-completion to work
properly.  Specifically,
    https://github.com/spf13/cobra/pull/730
which fixes
    https://github.com/spf13/cobra/issues/694
is essential to avoid the risk of colliding and possibly
breaking kubectl dynamic bash-completion once helm v3 has its
own dynamic completion.

Updates Cobra to tip of master.
This is needed because the version used by helm v3 is missing multiple bash-completion fixes available in Cobra, which are necessary to implement dynamic bash completion (the requested and upcoming v3 version of PR #5562). 

Since Cobra was not part of the `Gopkg.toml` file, I updated `Gopkg.lock` manually.  I didn't use `dep` because it kept updating everything.  Is a manual update the way to go?

On dev-v3, `make test` uses the bash builtin `mapfile`.  This does not work with zsh on my Mac.  Any suggestion on how to run the tests on dev-v3 while on Mac and using zsh?

**If applicable**:


technosophos:
I'm not the right person to ping on this.

marckhouzam:
Looking at the last Cobra update (PR #4913), probably goes to @bacongobbler

marckhouzam:
I've been reading up on `dep` and I'm starting to feel that manually updating the `Gopkg.lock` file is not the way to go.  I was just confused because cobra is not part of the `Gopkg.toml` file.  But now I think adding it to `Gopkg.toml` is the way to go.  I'll post an updated patch to better upgrade Cobra.

marckhouzam:
I'm going to close this PR and make a new one.  I did a revert and the DCO bot does no seem to like that.

PR #5656, opened 2019-04-29
fix(docs): 'From Source' section in install doc

hickeyma:
Update the section for Helm 3 changes as follows:
- Helm path: $GOPATH/src/helm.sh
- make target

**If applicable**:

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -82,18 +82,16 @@ You must have a working Go environment with
 
 ```console
 $ cd $GOPATH
-$ mkdir -p src/k8s.io
-$ cd src/k8s.io
+$ mkdir -p src/helm.sh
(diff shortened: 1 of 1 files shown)


mattfarina (approved):
Note, for those who don't follow, `make` being run without any specifics runs the first goal. In this case "all". This runs build and built runs vendor (which is all bootstrap does now). So, this effectively does the same thing as the existing command which has redundancy built in.

So, this makes sense. For those who aren't familiar with make targets this may be confusing.

@bacongobbler does this look ok for a change? You're the "make guy".

bacongobbler (approved):

hickeyma:
Thanks @mattfarina @bacongobbler

PR #5657, opened 2019-04-29
Update Masterminds/semver to allow >32bit version numbers

### LINKED ISSUE #5648: Helm fails to parse SemVer versions with parts that are larger than 32 bits
Output of `helm version`: 

```
Client: &version.Version{SemVer:"v2.13.1", GitCommit:"618447cbf203d147601b4b9bd7f8c37a5d39fbb4", GitTreeState:"clean"}
```

Output of `kubectl version`:

```
Client Version: version.Info{Major:"1", Minor:"10", GitVersion:"v1.10.11", GitCommit:"637c7e288581ee40ab4ca210618a89a555b6e7e9", GitTreeState:"clean", BuildDate:"2018-11-26T14:38:32Z", GoVersion:"go1.9.3", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"10", GitVersion:"v1.10.11", GitCommit:"637c7e288581ee40ab4ca210618a89a555b6e7e9", GitTreeState:"clean", BuildDate:"2018-11-26T14:25:46Z", GoVersion:"go1.9.3", Compiler:"gc", Platform:"linux/amd64"}
```

Cloud Provider/Platform (AKS, GKE, Minikube etc.): Docker for Desktop

## Issue

The version of `github.com/Masterminds/semver` used by Helm (1.3.1) supports a max size of 32bits for each part of the SemVer version.

e.g. trying to use the version number `1.0.15562920261` will result in the following error when running `helm lint`:

```
[ERROR] Chart.yaml: version '1.0.15562920261' is not a valid SemVer
```

Or when tested directly with the `semver` package produces the following error:

```
Error parsing version segment: strconv.ParseInt: parsing "15562920261": value out of range
```

The semver package has already merged a fix for this issue (https://github.com/Masterminds/semver/pull/61). I've recompiled helm locally and can confirm it fixes the issue for me.

Would it be possible to have this fix released in a future version of helm?

paddycarey:
This PR updates the dependency on https://github.com/Masterminds/semver to provide support for parsing version numbers which contain parts which are larger than 32bits. Closes #5648 

When I ran `glide up` after modifying `glide.yaml` I noticed that the SHA for the Kubernetes dependency also changed. I think this is because that dependency is only pinned to a release branch and not a tag.

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -24,7 +24,7 @@ import:
     version: ^2.19.0
   - package: github.com/ghodss/yaml
   - package: github.com/Masterminds/semver
-    version: ~1.3.1
+    version: ~1.4.2
   - package: github.com/technosophos/moniker
     version: ~0.2
   - package: github.com/golang/protobuf
--- a/glide.lock
+++ b/glide.lock
@@ -1,5 +1,5 @@
-hash: d92d7faee5c7ecbb241dadcd993e5dd8dfba226739d20d97fabf23168613f3ac
-updated: 2019-04-16T15:32:58.609105-07:00
+hash: 8a007d8993bdffd14a1a2d674848bd085a27b09d7f177fab1dc55783059c4dce
+updated: 2019-04-29T12:23:33.902435+01:00
 imports:
 - name: cloud.google.com/go
   version: 3b1ae45394a234c385be014e9a488f2bb6eef821
@@ -188,7 +188,7 @@ imports:
 - name: github.com/Masterminds/goutils
   version: 41ac8693c5c10a92ea1ff5ac3a7f95646f6123b0
 - name: github.com/Masterminds/semver
-  version: 517734cc7d6470c0d07130e40fd40bdeb9bcd3fd
+  version: c7af12943936e8c39859482e61f0574c2fd7fc75
 - name: github.com/Masterminds/sprig
   version: 9f8fceff796fb9f4e992cd2bece016be0121ab74
 - name: github.com/Masterminds/vcs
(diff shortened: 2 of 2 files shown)


glide.yaml
```diff
@@ -24,7 +24,7 @@ import:
     version: ^2.19.0
   - package: github.com/ghodss/yaml
   - package: github.com/Masterminds/semver
-    version: ~1.3.1
+    version: ~1.4
```
hickeyma: Could you set this to the latest: 1.4.2 ?

hickeyma (changes requested):
Thanks for this.

glide.yaml
paddycarey: Done 👍

hickeyma (approved):

hickeyma:
LGTM, thanks. I would like to get another opinion on the `glide.lock` file as unsure for the SHA change for the Kubernetes dependency.

mattfarina (approved):

mattfarina:
The change for the Kubernetes dependency is because the `glide.yaml` is tracking the `kubernetes-1.14` branch rather than pinning to a release. So, it updated to a version from April 26th ([see it here](https://github.com/kubernetes/kubernetes/commit/b8f2b772e38a15165a6247256d650e8b04178318))

We should either keep following the branch as it's specified now or track to a pinned release. I will follow up with an additional PR to change the k8s/k8s dependency to a pinned release and we can discuss it there.

mattfarina:
Quick note, the dev-v3 branch has already been updated to the newer version.

PR #5560, opened 2019-04-05
ref(pkg/chart): add validation method to chart

adamreese:
Consolidate validation of Chart.yaml

PR #5660, opened 2019-04-29
Make errors regarding chart.yaml metadata clear

### LINKED ISSUE #5653: Helm 3: errors regarding Chart.yaml metadata are not clear
Output of `helm version`:

```
version.BuildInfo{Version:"v3.0+unreleased", GitCommit:"1b77848c89e8df414aea055cca5d1a1997a62f0c", GitTreeState:"clean"}
```

Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.0", GitCommit:"641856db18352033a0d96dbc99153fa3b27298e5", GitTreeState:"clean", BuildDate:"2019-03-26T00:05:06Z", GoVersion:"go1.12.1", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"11+", GitVersion:"v1.11.8-gke.6", GitCommit:"394ee507d00f15a63cef577a14026096c310698e", GitTreeState:"clean", BuildDate:"2019-03-30T19:31:43Z", GoVersion:"go1.10.8b4", Compiler:"gc", Platform:"linux/amd64"}
```

I tried to do `helm template` and this is what I got:

```
$ helm template redis                                  
Error: metadata apiVersion is required
```

On checking I noticed that the error was regarding my `Chart.yaml` file not having `apiVersion` field. But the error was not clear about the origin of the issue, due to which I started thinking if my template yamls missed the `apiVersion` field which is part of the k8s resource yamls. I think the error message could tell about the origin, for example

```
$ helm template redis                                  
Error: Chart.yaml: metadata apiVersion is required
```

karuppiah7890:
closes #5653 

Make errors regarding chart.yaml metadata clear 

**If applicable**:

### DIFF
--- a/pkg/chart/metadata.go
+++ b/pkg/chart/metadata.go
@@ -64,21 +64,22 @@ type Metadata struct {
 	Type string `json:"type,omitempty"`
 }
 
+// Validate validates the Chart.yaml contents
 func (md *Metadata) Validate() error {
 	if md == nil {
-		return ValidationError("chart.metadata is required")
+		return ErrMissingMetadata
 	}
 	if md.APIVersion == "" {
-		return ValidationError("chart.metadata.apiVersion is required")
+		return ErrMissingAPIVersion
 	}
 	if md.Name == "" {
-		return ValidationError("chart.metadata.name is required")
+		return ErrMissingName
 	}
 	if md.Version == "" {
-		return ValidationError("chart.metadata.version is required")
+		return ErrMissingVersion
 	}
 	if !isValidChartType(md.Type) {
-		return ValidationError("chart.metadata.type must be application or library")
+		return ErrInvalidType
 	}
 	// TODO validate valid semver here?
 	return nil
--- a/cmd/helm/testdata/output/install-chart-bad-type.txt
+++ b/cmd/helm/testdata/output/install-chart-bad-type.txt
@@ -1,1 +1,1 @@
-Error: validation: chart.metadata.type must be application or library
+Error: validation: chart.metadata.type must be 'application' or 'library' in Chart.yaml
--- a/cmd/helm/testdata/output/template-lib-chart.txt
+++ b/cmd/helm/testdata/output/template-lib-chart.txt
@@ -1,1 +1,1 @@
-Error: validation: chart.metadata.type must be application or library
+Error: validation: chart.metadata.type must be 'application' or 'library' in Chart.yaml
--- a/pkg/chart/errors.go
+++ b/pkg/chart/errors.go
@@ -15,9 +15,19 @@ limitations under the License.
 
 package chart
 
-// ValidationError represents a data validation error.
-type ValidationError string
+import "github.com/pkg/errors"
 
-func (v ValidationError) Error() string {
-	return "validation: " + string(v)
-}
+// ErrMissingMetadata indicates that Chart.yaml is missing
+var ErrMissingMetadata = errors.New("validation: chart.metadata (Chart.yaml) is required")
+
+// ErrMissingAPIVersion indicates that chart apiVersion is missing in Chart.yaml
+var ErrMissingAPIVersion = errors.New("validation: chart.metadata.apiVersion is required in Chart.yaml")
+
+// ErrMissingName indicates that chart name is missing in Chart.yaml
+var ErrMissingName = errors.New("validation: chart.metadata.name is required in Chart.yaml")
+
+// ErrMissingVersion indicates that chart version is missing in Chart.yaml
+var ErrMissingVersion = errors.New("validation: chart.metadata.version is required in Chart.yaml")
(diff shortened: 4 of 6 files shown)


pkg/chart/metadata.go
```diff
@@ -68,18 +68,19 @@ type Metadata struct {
 	Type string `json:"type,omitempty"`
 }
 
+// Validate validates the Chart.yaml contents
 func (md *Metadata) Validate() error {
 	if md == nil {
-		return errors.New("metadata is required")
+		return errors.New("chart metadata is required in Chart.yaml")
```
bacongobbler: It's unclear to me if there's a case where pkg/chart may be validating chart metadata that did NOT come from a Chart.yaml. I think it's safer to remove the "in Chart.yaml" part from the error message, leaving that to downstream consumers. In other words, I'd suggest `pkg/action` should probably report that this error occurred when validating Chart.yaml.

What do you think about that approach?

karuppiah7890: That makes sense. I will make the necessary changes 👍

karuppiah7890: On checking the code, I see that in `pkg/action`, the `Run` gets the chart as a struct. It's already loaded in the `main` package using this line 

https://github.com/helm/helm/blob/895e9192d440e144ead0796fd8c1db81b47ef1c6/cmd/helm/install.go#L177

The `Load` can have errors anywhere. Not sure how we can detect it was in `Chart.yaml`. I was thinking of adding the info about `Chart.yaml` here

https://github.com/helm/helm/blob/ffff0e8c33fde7547ace267c5c91b8c60fc5262c/pkg/chart/loader/load.go#L111

as this function clearly knows that it's loading from files. But then, the `Validate` function is defined on that chart level 

https://github.com/helm/helm/blob/ffff0e8c33fde7547ace267c5c91b8c60fc5262c/pkg/chart/chart.go#L102

and if we add other validations along with metadata validation later in the future, then adding the info that error is in `Chart.yaml` would be wrong. And we can't add the info inside the chart Validate function too, as the consumers need not have used files as the source to construct the chart struct. 

Considering these, I think for now a good option is we can add it here

https://github.com/helm/helm/blob/ffff0e8c33fde7547ace267c5c91b8c60fc5262c/pkg/chart/loader/load.go#L111

But like I said, if we add more validations in `Validate`, then the error message would be misleading. 

What do you think? Did you have any ideas in mind for this? And I need to think about how to append or prepend to an existing error and tell it was from `Chart.yaml` in a sensible way

hickeyma: As far as I can see, I don't think you will know what file the metadata has been loaded from.  @bacongobbler maybe you know something more?

FrenchBen:
Good start, could we also get an error message for when there is simply no `Chart.yaml` definition?
```bash
$ mkdir /tmp/mychart
$ helm template /tmp/mychart
Error: metadata is required
```

pkg/chart/metadata.go
bacongobbler: After looking at different parts of the codebase, this seems like a safe enough assumption to make so my original thinking is a little too defensive. The error change looks good to me.

However, can we make these errors of a typed value such as `ErrChartMetadataMissing`? Have a look at [ErrMissingMetadata](https://github.com/helm/helm/blob/cf79b90867740f58859882b980f3ccc0d84cb44c/pkg/plugin/installer/installer.go#L30-L31) for an example, as well as how that error type is thrown in other areas of the codebase.

By implementing errors in this fashion, error checks like the one on line 121 in pkg/chart/loader/loader_test.go can simply read `if err != ErrChartMetadataMissing`, rather than an exact string match check.

karuppiah7890:
Makes sense, I'll make the necessary changes!

karuppiah7890:
Looks like there have been quite some changes in the files I coded in in this PR https://github.com/helm/helm/pull/5781 . With these changes, now we could consider that it's explicit that it's related to the chart and not related to the kubernetes as now it clearly says `chart.metadata.apiVersion` and not just `metadata apiVersion` which could be confused with k8s template yaml apiVersion field.

I think I could add the place where the error occurred - `in Chart.yaml` and some error types for the different error and also write some tests for those

What are your thoughts @bacongobbler and @hickeyma

hickeyma:
@karuppiah7890 Go for it.

karuppiah7890:
@hickeyma I have made the requested changes 😄

hickeyma (commented):
@karuppiah7890 Thank you for your effort and patience on this PR. The validation was however added in #5560. Closing as already implemented.

hickeyma:
Closing as implemented in #5560.

karuppiah7890:
Oh. Didn't notice that PR. 👍

PR #5662, opened 2019-04-30
fix(chartutil): Scaffold chart fails to deploy pod

hickeyma:
When you try to install a scaffold chart OOTB after a `helm create`, a subsequent install fails on the pod deployment with `InvalidImageName` as follows:

```console
default       pod/mychart-1556529386-576b45f6fb-w8vm4     0/1     InvalidImageName   0          31h
``` 

This is because the image name is missing the version.  The following line is used to set the image in `deployment.yaml`: `image: "{{ .Values.image.repository }}:{{ .Release.AppVersion }}"`. It seems that the `Release` object `AppVersion` property does not exist. Therefore, it gets rendered as follows:
```console
[....]
spec:
      containers:
        - name: mychart
          image: "nginx:"
[...]
```
This PR is to fix this issue.

Using `AppVersion` from the chart object and also setting the version to `stable` as the stable version of nginx is used. Not sure if this different to the approach that was envisaged for #5553. 

**If applicable**:

### DIFF
--- a/pkg/chartutil/create.go
+++ b/pkg/chartutil/create.go
@@ -72,8 +72,9 @@ type: application
 version: 0.1.0
 
 # This is the version number of the application being deployed. This version number should be
-# incremented each time you make changes to the application.
-appVersion: 0.1.0
+# incremented each time you make changes to the application. Depending on the application,
+# 'stable' can signify the latest stable version of the application.
+appVersion: stable
 `
 
 const defaultValues = `# Default values for %s.
@@ -211,7 +212,7 @@ spec:
     spec:
       containers:
         - name: {{ .Chart.Name }}
-          image: "{{ .Values.image.repository }}:{{ .Release.AppVersion }}"
+          image: "{{ .Values.image.repository }}:{{ .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           ports:
             - name: http
--- a/pkg/chartutil/create.go
+++ b/pkg/chartutil/create.go
@@ -72,9 +72,8 @@ type: application
 version: 0.1.0
 
 # This is the version number of the application being deployed. This version number should be
-# incremented each time you make changes to the application. Depending on the application,
-# 'stable' can signify the latest stable version of the application.
-appVersion: stable
+# incremented each time you make changes to the application.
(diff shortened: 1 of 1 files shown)


pkg/chartutil/create.go
```diff
@@ -72,8 +72,9 @@ type: application
 version: 0.1.0
 
 # This is the version number of the application being deployed. This version number should be
-# incremented each time you make changes to the application.
-appVersion: 0.1.0
+# incremented each time you make changes to the application. Depending on the application,
+# 'stable' can signify the latest stable version of the application.
+appVersion: stable
```
bacongobbler: Is there a reason this was changed?

hickeyma: The app being deployed is nginx and we want the stable version.

mattfarina: This get's into something we should talk about more. "stable" is an [nginx tag](https://hub.docker.com/_/nginx/?tab=tags) for the latest stable release. Since the chart uses the `appVersion` as the tag this will work for nginx.

But, this is scaffolding out a chart and guiding end users. This isn't about doing nginx well. Where do we want to guide people and do we want to guide them in a direction like this?

On the charts repo we have a policy against using moving tags. This is not an allowed behavior because expectations can change out from under users. For example, a user could install a chart into QA with one version. It checks out. During the QA period a new stable release comes out. When the user goes to deploy to production a different version is installed. This causes an issue with reproducibility.

In the case of nginx the changes have also been minor versions that should not break compatibility. But, this doesn't always need to be the case. What if nginx 2.0.0 comes out? For these occasions the above problem is worse.

This makes me not want to use moving tags for releases.

hickeyma: @mattfarina Good point, thanks.
I was trying to mirror previous capability with using the 'stable' tag.

I would propose then to set it as follows:
`appVersion: 1.16.0` to define the version of the nginx app that is being deployed.

How does this sound?

bacongobbler: SGTM

hickeyma: Thanks @bacongobbler @mattfarina for the feedback. Update and ready for review again.

mattfarina (approved):

PR #5406, opened 2019-03-05
Fix #5046 compatible with MacOS

marckhouzam:
Signed-off-by: Marc Khouzam <marc.khouzam@ville.montreal.qc.ca>

**What this PR does / why we need it**:

Fix issue #5046 for MacOS.

The sed tool on MacOS does not support the use of `\w`.  This PR uses a more portable syntax to convert bash completion to zsh completions.

Different regex could have been used, but I tried to mimic what the original `\w\+` aimed to do, which was to match one or more characters.  Granted, `.` matches more things than `\w`, but I don't believe it is necessary to be that strict in the matching rule.

I've tested the syntax on my mac and on alpine linux in docker.
I've also built helm with the change on my mac and confirmed that it fixes the zsh completion.

PR #5072, opened 2018-12-18
Fixes #5046, zsh completion

Sarke:
Signed-off-by: Peter Stalman <sarkedev@gmail.com>

**What this PR does / why we need it**:
Closes #5046

**Special notes for your reviewer**:
Small fix.  Removes quotes in aliashash keys to work with ZSH.

**If applicable**:
- [ ] this PR contains documentation
- [ ] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

Issue #5046, opened 2018-12-11
helm completion fails in zsh (prezto) with "bad math expression: operand expected"


ezekills:
I noticed helm completion in zsh is not working.
One part of the problem is that _helm is not executable per default.
Second when I type helm and hit tab I get:
```
helm _helm_root_command:12: bad math expression: operand expected at `"del"'
```
I found that removing the doublequotes in _helm at every aliashash key value fixes it. 
(/usr/local/Cellar/kubernetes-helm/2.12.0/share/zsh/site-functions/_helm)

**not working:**
```
aliashash["del"]="delete" 
```
**fixed:**
```
aliashash[del]="delete"
```

I am using prezto and I installed helm with homebrew.

philoserf:
I have confirmed the issue does not appear in helm 2.11.0

This may have been introduced by https://github.com/helm/helm/pull/4851. It is the only mention of _completion_ in release 2.12.0.

ref: https://github.com/helm/helm/releases/tag/v2.12.0

Sarke:
Same with oh-my-zsh, so the issue seems to be the quotes in the hash maps not being allowed by ZSH.

Here's a diff of the two versions' `helm completion zsh`.

```diff
--- v2.11.0.zsh
+++ v2.12.0.zsh
@@ -1,4 +1,3 @@
-
 #compdef helm
 
 __helm_bash_source() {
@@ -240,7 +239,13 @@
     fi
 
     if [[ ${#COMPREPLY[@]} -eq 0 ]]; then
-        declare -F __custom_func >/dev/null && __custom_func
+		if declare -F __helm_custom_func >/dev/null; then
+			# try command name qualified custom func
+			__helm_custom_func
+		else
+			# otherwise fall back to unqualified for compatibility
+			declare -F ___custom_func >/dev/null && __custom_func
+		fi
     fi
 
     # available in bash-completion >= 2, not always present on macOS
@@ -362,6 +367,14 @@
         __helm_handle_command
     elif [[ $c -eq 0 ]]; then
         __helm_handle_command
+    elif __helm_contains_word "${words[c]}" "${command_aliases[@]}"; then
+        # aliashash variable is an associative array which is only supported in bash > 3.
+        if [[ -z "${BASH_VERSION}" || "${BASH_VERSINFO[0]}" -gt 3 ]]; then
+            words[c]=${aliashash[${words[c]}]}
+            __helm_handle_command
+        else
+            __helm_handle_noun
+        fi
     else
[...]

PR #5680, opened 2019-05-07
v2: fix(completion): --flag=val breaks zsh completion

marckhouzam:

zsh completion cannot handle the flag format --flag=val.

This is a bug I ran into when working on Helm completion.
I was surprised that it didn't happen when I was using
kubectl, so I investigated and found a PR that fixed this
bug in kubectl:
https://github.com/kubernetes/kubernetes/pull/48553

I've marked the original author, @superbrothers , as a co-author to this PR.

I duplicated the code in this commit which:

Removes __helm_declare, which is safe to do since
`declare -F` is already replaced to `whence -w` by
__helm_convert_bash_to_zsh().

The problem was that calling "declare" from inside a function
scopes the declaration to that function only.  So "declare"
should not be called through __helm_declare() but instead
directly.

To reproduce:
1- setup helm completion in zsh
2- helm --kubeconfig=$HOME/.kube/config statu\<TAB\>

you will get the error:
__helm_handle_flag:27: bad math expression: operand expected at end of string

Co-authored-by: Kazuki Suda <kazuki.suda@gmail.com>

I realized that this bug was also the root of the problem of #5046.  With this PR
fixing the issue in the proper way,  I've added a commit to remove
the previous fix for #5046 (PR #5072 and #5406). 

However, for completely other reasons, we still need to remove the quotes from the aliashash associative array in zsh.  I've therefore put it back.


eddiezane (approved):
LGTM. Should be the actual fix. Same thing I did in https://github.com/digitalocean/doctl/pull/403 albeit I didn't quite understand it at the time.

marckhouzam:
@eddiezane you may be interested in applying a similar commit as https://github.com/helm/helm/pull/5680/commits/48f0e3101d49a3de1466ff75fc786ce012f369cc for doctl if you use command aliases.

hickeyma:
@marckhouzam Try doing a forced push to do a rebuild.

allamand:
I run on this problem, would this fix be merged ?

marckhouzam:
The https://github.com/helm/acceptance-testing repo can help maintainers get a feel that this change is safe.

If you run the completion tests on the master branch now they will pass.
If you run them again after applying this PR they will fail with an UNEXPECTED SUCCESS.
This is because this PR fixes a specific test that was marked as a known failure.
The PR to mark the test as expected to pass is ready here https://github.com/helm/acceptance-testing/pull/35

(I know this PASS -> FAIL after a PR is merged is not ideal, but with the tests being in a different repo, it is difficult to handle corresponding PRs).

For helm v2, to run the completion tests you can do the following.
(there is a PR to make it easier #6388 )

```
make && make build-cross && cd $GOPATH/src/helm.sh/acceptance-testing && \
ROBOT_RUN_TESTS=shells.robot ROBOT_HELM_PATH=$GOPATH/src/k8s.io/helm/bin make; cd -
```

marckhouzam:
@mattfarina As discussed, this PR is important to eventually get v2 auto-complete to play well with v3 auto-complete during the migration period.
The change of this PR is the same as #6389 which was merged in v3

bacongobbler (approved):
Nice catch! LGTM

PR #5681, opened 2019-05-07
fix(pkg/action): load clients after flags have been parsed

adamreese:


bacongobbler (approved):

mattfarina (approved):
I manually tested with the bug test I found around this problem and it's fixed with this.

PR #5693, opened 2019-05-08
update golang/protobuf and add Go version to version command

tariq1890:
Why add go version?

Golang goes through a lot of security patch fixes in its development lifecycle and it would be useful to know which version helm is compiled in to detect potential vulnerabilities. Also kubectl prints out go version, so why not helm :) ?


tariq1890:
Re-triggered the test. CircleCI :/

tariq1890:
@bacongobbler Would you be able to take a look at this PR?

bacongobbler:
I'm not sure we can accept this as it changes the output of `helm version`.

tariq1890:
It's an additive change. Would it still break semver contracts?

bacongobbler:
According to our [semver contract](https://github.com/helm/helm/blob/master/CONTRIBUTING.md#semantic-versioning), we don't have a contract for making breaking changes to command output, so I think this is safe.

@mattfarina do you have any thoughts/opinions on this one?

mattfarina:
The Go version for this is additive which should be fine for SemVer. It's useful information to have given the Go version support model and for tracing to bugs with versions. This would need to be ported to the dev-v3 branch if merged.

Updating protobuf is another. What version was this updated to? If the version has changed the [developers.md](https://github.com/helm/helm/blob/master/docs/developers.md) file should be updated, too. I'd like to poke at the differences between versions. The gRPC/protobuf updates have caused problems in the past so I'd like to take updates with care.

tariq1890:
I used protobuf 3.7.0

tariq1890:
Hi @mattfarina. Do you agree that the `developers.md` does not need changing? There are no breaking `.proto` changes and updating `golang/protobuf` is also safe since `helm` already used `protoc-gen-go` for proto generation.

mattfarina:
@tariq1890 The developers.md should be updated to reflect the version we are using. This way it's documented and someone doesn't go back to the older version.

tariq1890:
Apologies for the confusion @mattfarina . I have actually used protoc `3.2.0` and I will have to amend my previous comment. Generating proto files with `3.7.0` is actually a breaking change.
I have only updated the `golang/protobuf` dependency (not `protobuf`) which I believe is not documented in the `developers.md` file.

tariq1890:
I have now updated the libprotoc to 3.7.1, updated protoPackage to version 3, as well as the `developers.md`.

tariq1890:
Can we also add the `Need V3 Fix` label to this one?

mattfarina (approved):

mattfarina:
Note, due to the size of the change it needs to approvals before it can be merged.

tariq1890:
I've created #5859 for helm 3

tariq1890:
Can this get another review please? It's been more than a month since I created this PR.

cmd/helm/version_test.go
```diff
@@ -30,8 +30,8 @@ import (
 func TestVersion(t *testing.T) {
 	lver := regexp.QuoteMeta(version.GetVersionProto().SemVer)
 	sver := regexp.QuoteMeta("1.2.3-fakeclient+testonly")
-	clientVersion := fmt.Sprintf("Client: &version\\.Version{SemVer:\"%s\", GitCommit:\"\", GitTreeState:\"\"}\n", lver)
-	serverVersion := fmt.Sprintf("Server: &version\\.Version{SemVer:\"%s\", GitCommit:\"\", GitTreeState:\"\"}\n", sver)
+	clientVersion := fmt.Sprintf("Client: &version\\.Version{SemVer:\"%s\", GitCommit:\"\", GitTreeState:\"\", GoVersion:\"go1.12.5\"}\n", lver)
```
bacongobbler: embedding runtime/os configuration in testdata is generally considered an anti-pattern. The tests would nee to be updated each time a new version of Go is released. Can you strip this information out? Have a look at https://github.com/helm/helm/pull/5869 for an example.

bacongobbler (changes requested):
As far as I understand, the gRPC version bump is independent of the feature provided in this PR (exposing `runtime.Version()` in `helm version`). A bump to gRPC has bitten us with breaking changes in the past, which is why we generally avoid bumping it unless absolutely necessary, such as when there were several big fixes to the TLS handshake protocol that were worth bringing in.

Helm 2 PR reviews are focused primarily on maintenance, ensuring there are *no* breaking changes between releases. In general, that means we tend to avoid reviewing PRs that

1. come with a certain level of risk associated (like a gRPC version bump)
2. Are new features (as those can break old assumptions or break backwards compatibility)

For new features, we are asking people to consider writing a PR against Helm 3.

Because the change to `helm version` appears to be a minor feature-add, it seems like a harmless feature to consider for Helm 2, however there's a significant risk to backwards compatibility associated with bumping gRPC. If you want to get a quicker review, I suggest splitting out the gRPC bump from this PR entirely and submit that as a separate PR. That way each part can be reviewed and evaluated separately without holding up this new feature.

Thanks.

cmd/helm/version_test.go
tariq1890: Sure.

tariq1890:
@bacongobbler I have addressed your review comment.

Sorry for the delay. This would be super useful for us since we have found ourselves in more situations where knowing compiled go version would help.

tariq1890:
Hmmm. Odd, the tests pass locally for me.

tariq1890:
CircleCI passed :0. A miracle!

bacongobbler:
Sorry for not getting back to you sooner. The team has been focusing for the past few months on getting Helm 3 out the door, and that has meant it's been harder to find time to review feature additions for Helm 2.

Helm 2.15 is soon to be released, and will be the last feature release for Helm 2. Because of the time commitment required to review and test this, we won't be able to find the time to review this. The good news is that this has already been merged in for Helm 3.

Thanks again!

PR #5705, opened 2019-05-09
Add the ability to verify template output

smcavallo:
Validating and unit testing all infrastructure-as-code is requirement, especially for production infrastructure.  There are some third party libraries out there to unit test helm template output.  It makes the most sense for this to live closest to the source and integrated into helm itself.

Why? 
This feature provides the ability for consumers to build tests against charts:
- as part of local development processes.  (Ex. don't check in the chart if validation fails)
- and integrate them into their build/CI pipelines (Ex. run all template tests as part of jenkins build)
- and integrate them into their CD pipelines (Ex. don't run this against production if the command fails)

When running the `--test-outfile FILE` command the output of the template command is captured and compared against known output in `FILE`
It is a superset of the output which makes it extremely flexibly and simple to validate an entire release.
Validation failures result in a message written to Stderr.

Additionally, the design allows us to write better tests against template_test.go as the template tests leverage this functionality.  This results in better unit tests for helm.

It would be nice to provide better output, as a side-by-side comparison.
Ex.  https://github.com/go-test/deep
I did not want to add any additional 3rd party libraries to this project.  I would prefer to use a 3rd party library for better user feedback for the output if the maintainers are open to it.

**If applicable**:
- [X ] this PR contains unit tests


bacongobbler:
Hey @smcavallo, after reviewing this PR, this seems out of scope for `helm template`. However, this would be a fantastic plugin. For example, [helm-unittest](https://github.com/lrills/helm-unittest) verifies the output of `helm template` [snapshot testing](using https://github.com/lrills/helm-unittest#snapshot-testing). I highly suggest looking into this tool and seeing if it solves the same use cases you are trying to accomplish here.

Thanks so much!

PR #5706, opened 2019-05-09
fix(pkg/storage/driver): use shallowReleaseEqual()

bacongobbler:
as noted by @mortent in #5699, using `reflect.DeepEqual` no longer works with the updated go types as protobuf adds some extra fields. This PR replaces the DeepEqual with a shallow equal for release objects that only checks some of the fields.

Addresses the comments made in #5371, fixing the failing tests in master.

### DIFF
--- a/pkg/storage/driver/sql_test.go
+++ b/pkg/storage/driver/sql_test.go
@@ -18,7 +18,6 @@ package driver
 
 import (
 	"fmt"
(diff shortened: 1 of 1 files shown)


SlickNik (approved):
👍 Looks good to me. Thanks @bacongobbler!

Issue #5663, opened 2019-04-30
Move Helm Downloads


mattfarina:
Helm downloads are currently served out of a bucket in GKE owned by the Kubernetes people at Google. This is not part of the Kubernetes infra in the CNCF, where Kubernetes is migrating the things to. The bucket is not in the control of the Helm project.

We should migrate downloads/assets to a new location run by the Helm project. Two possibilities immediately come to mind:

1. GitHub Releases
2. An object storage bucket owned by the Helm project and behind custom DNS (e.g., downloads.helm.sh). The custom DNS would allow us to move the backend location later without impacting end users.

The pros to using GitHub Releases is that a) they are free to use and b) someone viewing the releases can find all the files. The cons to GitHub Releases are the extra complication in automation uploading the files and in the scripts that need to find the right file to download and install.

The pros to a custom storage bucket + DNS is that a) it's in more of our control and b) the scripts we have will be less complicated. The con is cost. But, we could do this on the existing Helm Azure account where we have the room.

We've been asked when we can move off the Google managed infrastructure so this is something we need to do.

Any thoughts on the alternative path to start using?

mattfarina:
cc @helm/helm-core-maintainers

bacongobbler:
Thanks for bringing this up, @mattfarina.

I'm personally leaning towards object storage served with a CDN. Having control of the distribution of Helm releases would be great to have, as it allows us to control the distribution (like in China) and bandwidth (not hogging Github's CDN during spike loads).

Additionally, we would be able to receive reliable download metrics of our content, and from what regions it's being downloaded from. As far as I know, GitHub does not provide those download metrics.

Using the existing Helm Azure account would be great to burn some of that unused budget. :)

Should we decide to go down this route, we should see what we can do about documenting the architecture of the Helm CDN. That way we can quickly rebuild it in case of emergency (or perform maintenance/updates to the system via PR).

technosophos:
I'll just chime in as "Another Matt who agrees". CDN-fronted object storage is a good solution for this, and we do have Azure File as a storage option under our Helm account.


I think we should work on making the metrics public, if possible, even if that means taking a report and exporting it back to object storage as a data file. What level of metrics is appropriate for distributing?


Matt

________________________________

viglesiasce:
I'll break up the Matt-fest a bit :)

I think custom DNS is the most important part and should be prioritized as a first step. From there we can move things around as needed. CDN+Object storage sounds good.

technosophos:
[...]

PR #5715, opened 2019-05-10
feat(ci): push release assets to Azure and GCS

bacongobbler:
This PR installs the Azure CLI and pushes release assets to Azure, making Helm 2 releases available at https://get.helm.sh.

related to #5663 

This PR differs from #5694 because Helm 2 relies on `golang:1.12.5`, which runs as root so no use of `sudo` is required. We also continue publishing to GCS to continue supporting backwards compatibility until that bucket is shut down. At that point we can remove the GCS steps.

### DIFF
--- a/.circleci/deploy.sh
+++ b/.circleci/deploy.sh
@@ -22,6 +22,8 @@ fi
 
 : ${GCLOUD_SERVICE_KEY:?"GCLOUD_SERVICE_KEY environment variable is not set"}
 : ${PROJECT_NAME:?"PROJECT_NAME environment variable is not set"}
+: ${AZURE_STORAGE_CONNECTION_STRING:?"AZURE_STORAGE_CONNECTION_STRING environment variable is not set"}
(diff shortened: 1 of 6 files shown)


jdolitsky (approved):

hickeyma (changes requested):
Looks good @bacongobbler. One small thing [release notes](https://github.com/helm/helm/blob/master/docs/release_checklist.md#8-write-the-release-notes) need updating in release checklist.

bacongobbler:
Thanks! Fixed those up. I also found that `make fetch-dist` still pointed at GCS so I fixed that up as well.

Once the Helm 2.14.0 release assets are available on get.helm.sh, this should be good to merge. I'll give it a quick manual test to make sure.

bacongobbler:
Also applied to https://github.com/helm/helm/pull/5841 to this PR.

mattfarina (approved):

PR #5716, opened 2019-05-10
update links to get.helm.sh

bacongobbler:
related to #5663 

### DIFF
--- a/docs/release_checklist.md
+++ b/docs/release_checklist.md
@@ -94,31 +94,6 @@ index 2109a0a..6f5a1a4 100644
         BuildMetadata = "unreleased"
 ```
 
-The README stores links to the latest release for helm. We want to change the version to the first release candidate which we are releasing (more on that in step 5).
-
-```shell
-$ git diff README.md
-diff --git a/README.md b/README.md
-index 022afd79..547839e2 100644
---- a/README.md
-+++ b/README.md
-@@ -34,10 +34,10 @@ Think of it like apt/yum/homebrew for Kubernetes.
-
- Binary downloads of the Helm client can be found at the following links:
-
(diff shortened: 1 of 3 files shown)


michelleN (approved):

hickeyma (changes requested):
It looks good but for one small point. Should [release checklist](https://github.com/helm/helm/blob/dev-v3/docs/release_checklist.md) also be updated?

bacongobbler:
Good point @hickeyma. I somehow missed that. I'll go and update those links as well. Thanks!

bacongobbler:
In the process of updating links, I found an old section of the docs that wasn't updated after the fork. I removed that section. Thanks for the catch @hickeyma

hickeyma (changes requested):
Updates good. One small nit. The example release note in the release checklist doc points to google storage.

bacongobbler:
Good catch. Fixed!

docs/release_checklist.md
```diff
@@ -149,24 +124,24 @@ git push upstream $RELEASE_CANDIDATE_NAME
 
 CircleCI will automatically create a tagged release image and client binary to test with.
 
-For testers, the process to start testing after CircleCI finishes building the artifacts involves the following steps to grab the client from Google Cloud Storage:
+After CircleCI finishes building the artifacts, the use the following commands to fetch the client for testing:
```
hickeyma: small nit here if you want to change it: :s/the use/use

hickeyma (approved):
LGTM, thanks.

docs/release_checklist.md
bacongobbler: fixed. Thanks again

PR #5717, opened 2019-05-10
Helm 3: pass debug option to registry client

jdolitsky:
This is a small change that allows us to silence the verbose logs coming from ORAS/Containerd, configurable with the global `--debug` flag


bacongobbler (approved):
LGTM. Would be nice to get in for alpha.1.

Issue #5286, opened 2019-02-09
Standardize helm media type for OCI Config Descriptor


sajayantony:
Output of `helm version`:

```json
{  "Version":"v3.0+unreleased", 
    "GitCommit":"a32f8ebb377da054d21510671d63729fe6a00f3a", 
    "GitTreeState" :"clean"
}
```
Helm chart push will enable pushing an OCI artifact to the helm registry. This needs to have a standard artifact type that could potentially be used by oci/distribution. 

###  Current Format 
```json
{
  "schemaVersion": 2,
  "config": {
    "mediaType": "application/vnd.oci.image.config.v1+json",
    "digest": "sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a",
    "size": 2
  },
  "layers": [
    {
      "mediaType": "application/vnd.cncf.helm.chart.meta.v1+json",
      "digest": "sha256:5e7c22341d8c8bfaab4dd1acb230a796dbd2d4a57cfb1bc665dbc50671511583",
      "size": 420,
      "annotations": {
        "org.opencontainers.image.title": "chart-meta.json"
      }
    },
    {
      "mediaType": "application/vnd.cncf.helm.chart.content.v1+tar",
      "digest": "sha256:cefbfee797727bffece4f77b9c4c14b041d4f0787fc3edfeedca8e8753b8fd66",
      "size": 18502,
      "annotations": {
        "org.opencontainers.image.title": "chart-content.tgz",
        "sh.helm.chart.name": "wordpress",
        "sh.helm.chart.version": "5.1.2"
      }
    }
  ]
}
```

### Proposal  

OCI manifest allows specifying a config descriptor and the current discussion is to have the proper media type in the config descriptor. 

```json
{
  "schemaVersion": 2,
  "config": {
    "mediaType": "application/vnd.cncf.helm.config.v1+json",
...
```

| Type | MediaType | 
|------|------------|
| Config      | application/vnd.cncf.helm.chart.config.v1+json | 
| Metadata | application/vnd.cncf.helm.chart.meta.v1+json |
| Content    | application/vnd.cncf.helm.chart.content.v1+tar | 

/cc @jdolitsky @SteveLasker since this is related to the PR deislabs/oras#52

sajayantony:
Once - https://github.com/deislabs/oras/pull/56 is merged we should enable helm to push a config descriptor with the appropriate mime type.

bacongobbler:
closed via #5719

PR #5719, opened 2019-05-10
Helm 3:  set custom manifest config media type on chart push

jdolitsky:
This finalizes the custom mediatype for Helm charts stored in an OCI registry:
```
application/vnd.cncf.helm.chart.v1+json
```

A current manifest looks like the following:
```
{
  "schemaVersion": 2,
  "config": {
    "mediaType": "application/vnd.cncf.helm.chart.v1+json",
    "digest": "sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a",
    "size": 2
  },
  "layers": [
    {
      "mediaType": "application/vnd.cncf.helm.chart.meta.layer.v1+json",
      "digest": "sha256:ca9588a9340fb83a62777cd177dae4ba5ab52061a1618ce2e21930b86c412d9e",
      "size": 19,
      "annotations": {
        "org.opencontainers.image.title": "chart-meta.json"
      }
    },
    {
      "mediaType": "application/vnd.cncf.helm.chart.content.layer.v1+tar",
      "digest": "sha256:a66666c6b35ee25aa8ecd7d0e871389b5a2a0576295d6c366aefe836001cb90d",
      "size": 459,
      "annotations": {
        "org.opencontainers.image.title": "chart-content.tgz",
        "sh.helm.chart.name": "mychart",
        "sh.helm.chart.version": "2.7.0"
      }
    }
  ]
}
```

More modifications coming soon to the "layers" section, but this top-level media type is important to decide on first. Before this, it was using the ORAS default (`application/vnd.oci.image.config.v1+json`).

This is important for registries that wish to whitelist package types that they wish to store.

Also in this PR: upgrade to ORAS 0.5.0 which gives us the `oras.WithConfigMediaType` method.

This PR based on top of #5717 (lets merge that first)


pkg/registry/constants.go
```diff
@@ -20,11 +20,14 @@ const (
 	// HelmChartDefaultTag is the default tag used when storing a chart reference with no tag
 	HelmChartDefaultTag = "latest"
 
-	// HelmChartMetaMediaType is the reserved media type for Helm chart metadata
-	HelmChartMetaMediaType = "application/vnd.cncf.helm.chart.meta.v1+json"
+	// HelmChartConfigMediaType is the reserved media type for the Helm chart manifest config
+	HelmChartConfigMediaType = "application/vnd.cncf.helm.chart.config.v3+json"
```
bacongobbler: shouldn't it be v1 since this is the first iteration of the media type schema?

jdolitsky: That was my original thinking as well. There was convo in #oras about this

@stevelasker @sajayantony Can you please comment on this?

@bacongobbler please also see this list which ACR is currently using as whitelist (obviously subject to change based on Helm maintainers' decision): https://github.com/Azure/acr/blob/master/docs/artifact-media-types.json

Gopkg.toml
```diff
@@ -44,7 +44,11 @@
 
 [[constraint]]
   name = "github.com/deislabs/oras"
-  version = "0.4.0"
+  version = "0.5.0"
+
+[[constraint]]
```
tariq1890: The latest version is `1.4.1`. Also, how about using `klog` instead given that's the de-facto standard for logging?

jdolitsky: unfortunately this has to do with dependency hell of the supporting libs (oras+containerd)

If you want to help clean this up, I think we should open up tickets on ORAS for support for something like `orascontext.SetLevel(oraslog.DebugLevel)` so that we do not need to import logrus here at all
https://github.com/deislabs/oras

SteveLasker: on the version aspect. Just trying to keep too many version mappings. Since this is all accrues to helm 3, it seemed to align better. Here’s the discussion:
https://cloud-native.slack.com/archives/CJ1KHJM5Z/p1557442141035000

SteveLasker: on the version aspect. Just trying to keep too many version mappings. Since this is all accrues to helm 3, it seemed to align better. Here’s the discussion:
https://cloud-native.slack.com/archives/CJ1KHJM5Z/p1557442141035000

jdolitsky: guessing this was meant for the other thread above

pkg/registry/constants.go
sajayantony: This is up to the helm maintainers. From a registry, we would follow what ever the helm maintainers decide is the acceptable media type. 
I think v1 also makes sense since this is the first version, but its more about aligning with helm client which might not be needed and I can see that helm4 might still have v1 or v3 which ever is currently being used.

jdolitsky: > helm4 might still have v1

I think this is good explanation for using v1. The schema can be versioned separately from Helm client. I've gone ahead and switched it to v1.

SteveLasker:
I just noticed the Helm Chart version annotation is placed on the layer. Wouldn't this be more appropriate on the overall object/manifest? 
Either an annotation in the config, or actually use the config object to store this. 
Having at the layer complicates parsing at the registry, and allows different layers to have different versions. 
From the registry perspective, we'd like to basically ignore details on the layer. By parsing information at the manifest, we can surface information in listings (portal or CLI), and we can surface annotations on search results. I would like to avoid the perf and complications of parsing metadata at the layer level.

sajayantony:
That not fully accurate @stevelasker. Layer annotations as in the manifest. So that is already being parsed. It’s not a concern for perf. 
I would discuss the design aspect.

SteveLasker:
I'm still a bit concerned about the possible duplication. It just feels wrong to have annotations of a thing, at layers that can be reused. Although, if the chart is a reusable layer, I may be convincing myself the other way. 
I'd like to start some depth conversations on when information should be in annotations, vs. a config object.

jdolitsky:
Hi, the final annotations and layers are still up for debate. This PR aims
to nail down the manifest media type only.

pkg/registry/client_test.go
```diff
@@ -71,7 +71,8 @@ func (suite *RegistryClientTestSuite) SetupSuite() {
 
 	// Init test client
 	suite.RegistryClient = NewClient(&ClientOptions{
-		Out: suite.Out,
+		Debug: false,
```
bacongobbler: this can be removed as the default value for booleans is `false`.

pkg/registry/constants.go
```diff
@@ -42,7 +45,7 @@ const (
 // KnownMediaTypes returns a list of layer mediaTypes that the Helm client knows about
 func KnownMediaTypes() []string {
```
bacongobbler: Should `HelmChartConfigMediaType` be a known media type?

pkg/registry/constants.go
```diff
@@ -20,11 +20,14 @@ const (
 	// HelmChartDefaultTag is the default tag used when storing a chart reference with no tag
 	HelmChartDefaultTag = "latest"
 
-	// HelmChartMetaMediaType is the reserved media type for Helm chart metadata
-	HelmChartMetaMediaType = "application/vnd.cncf.helm.chart.meta.v1+json"
+	// HelmChartConfigMediaType is the reserved media type for the Helm chart manifest config
+	HelmChartConfigMediaType = "application/vnd.cncf.helm.chart.config.v1+json"
```
bacongobbler: is there a difference in intent between `meta` and `config` in terms of OCI usage here? Why is the name being changed?

SteveLasker: config is the object used to differentiate the `artifactType`. 
In lieu of the pending distribution PR, here's a quick guide for types: https://stevelasker.blog/2019/05/11/authoring-oci-registry-artifacts-quick-guide/
Artifact Type Format
Artifacts are defined by `manifest.config.mediaType` using the following format:

`application/vnd.`[org/company]`.`[objectType]`.`[optional sub type]`.`config.`[version]`+json

bacongobbler: I don't understand *why* we're re-naming it to `config`, and from what I read in that blog post I could not figure out the reason. Can you point out *why* we need to change it to `config`? `meta` is the standard descriptor we've been using for chart metadata so I'm not sure why we need to change the convention to follow.

The [image-spec documentation](https://github.com/opencontainers/image-spec/blob/master/media-types.md) doesn't appear to follow that convention laid out in the blog post either. `manifest`, `descriptor`, `index`, `header` and others all don't follow that convention, so I'm led to believe that this isn't a convention we need to follow.

According to the image-spec, the [OCI image config specification](https://github.com/opencontainers/image-spec/blob/master/config.md) uses the term `config` because the metadata stored in that object is the image configuration - that is, it describes how the layers should be used by a container runtime and execution tool and its relationship to filesystem changesets, and therefore it's "config". In other words, the content and how it's being used defines the schema, not the other way around.

SteveLasker:
I need @jdolitsky to comment on the intent here. The `manifest.config.mediaType` is used to determine the type of artifact. In this case, a Helm Chart. It *could* contain information about the object, just as the image-spec defines how a container image is run. But a config file is optional. 

You're correct, the image-spec covers multiple mediaTypes. The [manifest](https://github.com/opencontainers/image-spec/blob/master/manifest.md#image-manifest-property-descriptions) has a subset, which we're extending for new artifacts. 

I also saw a layer that was a meta layer. If the thing @jdolitsky is trying to persist is a seperate file, used by helm, then, yes, this would be a `meta.layer` in the layers collection. However, we still need a way for registries and registry tools, like docker, helm, singularity, opa, to know what object types they're requesting. The `manifest.config.mediaType` is the piece of information used. If you have no artifact configuration information, just pass in a null pointer, or an empty config.json file. We should support null config files, so if this doesn't currently work, it's a bug.

SteveLasker:
The other key difference is the OCI Image Spec is for container images (docker). The OCI Artifact Registry design is to support additional artifacts. The distribution spec will outline the differences. Although, we're working hard to make them minimal as we expect a lot of common tooling. 
Here's the beginning of the distribution spec, talking about the formatting of the `manifest.config.mediaType`: https://github.com/AzureCR/distribution-spec/blob/artifact-registry/artifacts.md

jdolitsky:
Sorry I am away and sending from email so pardon any lack of detail/clarity
here-

The diff is misleading. We are not renaming from meta to config. We simply
were not using a custom manifest config media type prior to this and were
using the ORAS default.

In terms of the other renames, they just seemed more readable. Keep in mind
this PR is not focused on the later media types, I still want to iterate on
this prior to 3.0. I’m fine with reverting this if necessary.

In terms of the KnownMediaTypes - there is essentially no change here. The
manifest config media type need not be here for ORAS-based pull.

I’ll have computer access later today if you want to request any changes
given this explanation.

SteveLasker:
Ok, so this isn't renaming the `meta` layer, it's really just providing the info for registries and registry tooling to understand the artifact type. The ORAS default is simply OCI, which isn't really a type. It's like saying I have a file on disk, with no type. 
So, we do need a type. The ORAS library leaves it to the artifact author to define their type. 
Here's an example of how ACR knows the pushed artifact is a Helm Chart.
image
Notice the `manifest.config.mediaType` is set to helm. (this was the original mediaType). We've since changed it to: `application/vnd.cncf.helm.chart.config.v1+json`, based on the PR Josh sent.
We also have OPA as well:  https://github.com/Azure/acr/blob/master/docs/artifact-media-types.json

pkg/registry/client.go
```diff
@@ -97,7 +97,8 @@ func (c *Client) PushChart(ref *Reference) error {
 	if err != nil {
 		return err
 	}
-	_, err = oras.Push(c.newContext(), c.resolver, ref.String(), c.cache.store, layers)
+	_, err = oras.Push(c.newContext(), c.resolver, ref.String(), c.cache.store, layers,
+		oras.WithConfigMediaType(HelmChartConfigMediaType))
```
jdolitsky: this is really the core of the change here

pkg/registry/client_test.go
jdolitsky: done

pkg/registry/constants.go
jdolitsky: Just for the layer types. `HelmChartMetaMediaType` not removed, the var name was just changed

pkg/registry/constants.go
sajayantony: The initial design was to actually have an artifact type property as a sibling to .config. This was pushed out since landing V1 was priority. 
If we did get artifactType in V2 then the config for helm would actually truly be config metadata. Either way works.

bacongobbler (approved):
LGTM

jdolitsky:
fixes https://github.com/helm/helm/issues/5286

jdolitsky:
Updated this PR to use config mediatype `application/vnd.cncf.helm.chart.v1+json` instead

cc @bacongobbler @SteveLasker

sajayantony (approved):

bacongobbler (approved):

PR #5725, opened 2019-05-14
Remove uselss indentation of lists

naseemkullah:
Cleans the sample chart files. No need to indent lists.

I guess this is more of a preference. But for a while I thought indenting lists was necessary due to stuff like this, but it actually is not and would like to prevent any confusion.

**If applicable**:

### DIFF
--- a/pkg/chartutil/create.go
+++ b/pkg/chartutil/create.go
@@ -77,13 +77,12 @@ ingress:
     # kubernetes.io/ingress.class: nginx
     # kubernetes.io/tls-acme: "true"
   hosts:
-    - host: chart-example.local
-      paths: []
-
+  - host: chart-example.local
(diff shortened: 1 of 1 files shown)


hickeyma (commented):
Thanks for the PR @naseemkullah. I am not sure however that this is an improvement. IMO, the indentation brings better clarity to the user.

naseemkullah:
Fair enough. Thought I would try... I guess it's a question of preference then.
Feel free to close @hickeyma

unguiculus:
I agree with @hickeyma. It's definitely not useless but rather a matter of taste. Personally, I prefer indentation as well.

hickeyma:
@naseemkullah Thanks for understanding. Closing.

PR #5728, opened 2019-05-14
fix(loader): assume apiVersion is v1 when loading charts

bacongobbler:
In Helm 2, no chart validation was performed on v1 charts, so there are
a few charts out there that never added an apiVersion to the Chart.yaml.
To ensure those charts continue to work for Helm 3, we should assume
that charts loaded without an apiVersion set should be assumed as v1.


mattfarina:
For reference... PR to add the check to helm v2 linting is at #5729.

mattfarina (approved):

hickeyma:
@ahilsend It shouldn't need to be ported as Helm 2.x should work as before. Do you mind raising an issue with the problem you are seeing?

ahilsend:
Of course! Opened the issue https://github.com/helm/helm/issues/6791

PR #5729, opened 2019-05-14
Adding lint check for apiVersion which is a required field

mattfarina:
Fixes #5727

Signed-off-by: Matt Farina <matt@mattfarina.com>

**What this PR does / why we need it**:

**Special notes for your reviewer**:

**If applicable**:
- [ ] this PR contains documentation
- [x] this PR contains unit tests
- [x] this PR has been tested for backwards compatibility

PR #5730, opened 2019-05-14
Fix lint apiversion v3

mattfarina:
Porting #5729 to Helm v3

### DIFF
--- a/pkg/lint/lint_test.go
+++ b/pkg/lint/lint_test.go
@@ -35,12 +35,12 @@ const goodChartDir = "rules/testdata/goodone"
 
 func TestBadChart(t *testing.T) {
 	m := All(badChartDir, values, namespace, strict).Messages
-	if len(m) != 5 {
+	if len(m) != 6 {
 		t.Errorf("Number of errors %v", len(m))
 		t.Errorf("All didn't fail with expected errors, got %#v", m)
 	}
 	// There should be one INFO, 2 WARNINGs and one ERROR messages, check for them
-	var i, w, e, e2, e3 bool
+	var i, w, e, e2, e3, e4 bool
 	for _, msg := range m {
 		if msg.Severity == support.InfoSev {
 			if strings.Contains(msg.Err.Error(), "icon is recommended") {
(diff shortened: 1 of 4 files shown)


pkg/lint/rules/chartfile.go
```diff
@@ -85,6 +87,18 @@ func validateChartNameDirMatch(chartDir string, cf *chart.Metadata) error {
 	return nil
 }
 
+func validateChartAPIVersion(cf *chart.Metadata) error {
+	if cf.APIVersion == "" {
```
bacongobbler: Could we somehow tie into [`cf.Validate()`](https://github.com/helm/helm/blob/55ae0c2efb48751cefe51cd66567608c0f7a650e/pkg/chart/metadata.go#L71) instead? That way we could keep the chart metadata validation DRY and in one package. 

Perhaps we could change the error output in `.Validate` to return typed errors like `ErrNoAPIVersionFound` or `ErrInvalidAPIVersion`. What do you think about that?

pkg/lint/lint_test.go
```diff
@@ -62,9 +62,13 @@ func TestBadChart(t *testing.T) {
 			if strings.Contains(msg.Err.Error(), "directory name (badchartfile) and chart name () must be the same") {
 				e3 = true
 			}
+
+			if strings.Contains(msg.Err.Error(), "apiVersion is required") {
```
bacongobbler: thought: the comment below about typed error values from `.Validate` could be a nice refactor here.

pkg/lint/rules/chartfile.go
mattfarina: I like the idea of typing into the `Validate` function in the chart package. From the looks of things, the current function is simple compared to what linting is doing on the `Chart.yaml` file. Other parts of linting have not been moved over to `Validate` either.

Should we do that migration in a separate PR? One that handles the change wholesale.

bacongobbler: Yeah, we can do that in a separate PR. 👍

bacongobbler (approved):

pkg/lint/lint_test.go
mattfarina: I added the guidance as you suggested.

Issue #5750, opened 2019-05-16
Helm install v2.14.0 "validation failed" error when using a template variable with value ""


daniv-msft:
Hello,

After upgrading from v2.13.1 to v2.14.0, my chart now throws an error on helm install:

**Error: validation failed: error validating "": error validating data: unknown object type "nil" in Deployment.spec.template.metadata.annotations.buildID**

This seems to be due to the use in deployment.yaml file of a template variable "buildID" that is actually never declared in values.yaml.
Extract from deployment.yaml:
```
template:
  metadata:
    labels:
      app: {{ template "gateway.name" . }}
      draft: {{ default "draft-app" .Values.draft }}
      release: {{ .Release.Name }}
    annotations:
      buildID: {{ .Values.buildID }}
```

If I set the buildID variable in values.yaml file to "", I get the same error.
If I set the buildID variable in values.yaml file to any other string, such as "a", then my install works.
If I set "" to buildID in deployment.yaml (buildID: {{ "" }}), I get the same error.
If I set "" directly to buildID in deployment.yaml (buildID: ""), then my install works.

Could you please let me know if this is a known issue, or if I am missing anything here?

Thanks!

------------

**Output of `helm version`:**
Client: &version.Version{SemVer:"v2.14.0", GitCommit:"05811b84a3f93603dd6c2fcfe57944dfa7ab7fd0", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.14.0", GitCommit:"05811b84a3f93603dd6c2fcfe57944dfa7ab7fd0", GitTreeState:"clean"}

**Output of `kubectl version`:**
Client Version: version.Info{Major:"1", Minor:"10", GitVersion:"v1.10.11", GitCommit:"637c7e288581ee40ab4ca210618a89a555b6e7e9", GitTreeState:"clean", BuildDate:"2018-11-26T14:38:32Z", GoVersion:"go1.9.3", Compiler:"gc", Platform:"windows/amd64"}
Server Version: version.Info{Major:"1", Minor:"12", GitVersion:"v1.12.7", GitCommit:"6f482974b76db3f1e0f5d24605a9d1d38fad9a2b", GitTreeState:"clean", BuildDate:"2019-03-25T02:41:57Z", GoVersion:"go1.10.8", Compiler:"gc", Platform:"linux/amd64"}

**Cloud Provider/Platform:** 
AKS

StephanX:
This bit me today as well.

fooka03:
[...]

PR #5760, opened 2019-05-19
fix(helm): Disable schema validation for manifests

mortent:
Signed-off-by: Morten Torkildsen <mortent@google.com>

The change in #5576 and #5643 added schema validation for new manifests. This seems to cause problems for existing Helm setups, so this change removes schema validation of charts in all situations.

Ref discussion in #5750 

**If applicable**:

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -157,13 +157,16 @@ func (c *Client) BuildUnstructured(namespace string, reader io.Reader) (Result, 
 }
 
 // Validate reads Kubernetes manifests and validates the content.
+//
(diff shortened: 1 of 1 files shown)


fooka03:
How painful would it be to have the functionality here (debug) log validation issues instead of disabling it completely?  With the discussion of moving this feature to helm 3, and it being near-ish on the horizon, I could see the value in knowing which of my manifests are going to be problematic in the new version.  This doesn't necessarily have to be part of this PR as I'd rather have a working helm ecosystem sooner than later but wanted to pose the question.

bacongobbler (approved):

PR #5762, opened 2019-05-20
support --output-dir option for helm3 template

torstenwalter:

helm 2 has an option to write the executed templates to files in output-dir instead of stdout.
This PR adds the same to helm3.

**If applicable**:


jdolitsky (approved):
LGTM

pkg/action/install.go
```diff
@@ -363,12 +366,56 @@ func (c *Configuration) renderResources(ch *chart.Chart, values chartutil.Values
 
 	// Aggregate all valid manifests into one big doc.
 	for _, m := range manifests {
-		fmt.Fprintf(b, "---\n# Source: %s\n%s\n", m.Name, m.Content)
+		if outputDir == "" {
```
hickeyma: Should it be checked here for a blank template? This was similar to how it was checked in v2:
```console
if whitespaceRegex.MatchString(m.Content) {
        continue
}
```

pkg/action/install.go
```diff
@@ -363,12 +366,56 @@ func (c *Configuration) renderResources(ch *chart.Chart, values chartutil.Values
...
+			fmt.Fprintf(b, "---\n# Source: %s\n%s\n", m.Name, m.Content)
+		} else {
+			err = writeToFile("build", m.Name, m.Content)
```
hickeyma: :s/"build"/outputDir

hickeyma (changes requested):
@torstenwalter Thanks for submitting the PR. I have added some comments for change.
I think it would also be helpful if you could add a test case.

torstenwalter:
@hickeyma Thanks for your feedback. The suggested changes have been incorporated.

hickeyma:
@torstenwalter Would it be possible to add a unit test as mentioned in https://github.com/helm/helm/pull/5762#pullrequestreview-242671847 ?

torstenwalter:
@hickeyma Do have anything I can use as example? Haven't found any unit test in helm 2 for this.

torstenwalter:
@hickeyma While writing the test I fugured out that the check for empty manifests had no effect. They are never empty.

That's already implemented in
https://github.com/helm/helm/blob/dev-v3/pkg/releaseutil/manifest.go#L49-L51 and in https://github.com/helm/helm/blob/dev-v3/pkg/releaseutil/manifest_sorter.go#L88-L90

torstenwalter:
If you want I can also rebase and squash my changes on `dev-v3` branch.

hickeyma (approved):
LGTM, thanks for all your effort.

PR #5781, opened 2019-05-22
ref(pkg/chartutil): break up chartutil into logical files

adamreese:


technosophos (approved):
LGTM. That's a big refactor, but it sure makes the code a lot easier to read.

bacongobbler (approved):
nice refactor!

PR #5789, opened 2019-05-23
Skip waiting for paused deployments

perher:
In my chart I have multiple Deployments for different micro-services in my application. Some of the services have upgrade dependencies or requires special logic to perform rolling upgrade. To facilitate this I set those deployments to be in `paused` state before the chart is upgraded.

When running helm upgrade with paused deployments and when using the `--wait` flag, the upgrade processed is stalled forever (or until timeout) since Kubernetes will not perform rolling upgrade on paused deployments. I propose to have Tiller ignore paused deployments altogether when waiting for a upgrade to finish since these will otherwise block the upgrade process.

### DIFF
--- a/pkg/kube/wait.go
+++ b/pkg/kube/wait.go
@@ -71,6 +71,10 @@ func (c *Client) waitForResources(timeout time.Duration, created Result) error {
 				if err != nil {
 					return false, err
 				}
+				// If paused deployment will never be ready
+				if currentDeployment.Spec.Paused {
+					continue
+				}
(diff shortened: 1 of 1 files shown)


bacongobbler:
Hey @perher! Thanks for the bugfix. I've restarted the CI tests. This looks like a good fix. If you have the time, would you mind porting this over to the `dev-v3` branch as well? That way Helm 3 also has the same fix. Much appreciated!

bacongobbler (approved):

PR #5672, opened 2019-05-03
add errcheck for Digest method in sign.go

tariq1890:
Doing this as it is a publicly used method and potentially swallows errors. This should be a safe change as the method signature remains unchanged.

PR #5790, opened 2019-05-23
return error when io.Copy fails in Digest

tariq1890:
This PR corrects an oversight in #5672. My apologies.

Requesting your review @bacongobbler @technosophos @hickeyma

### DIFF
--- a/pkg/provenance/sign.go
+++ b/pkg/provenance/sign.go
@@ -405,7 +405,7 @@ func DigestFile(filename string) (string, error) {
 func Digest(in io.Reader) (string, error) {
 	hash := crypto.SHA256.New()
 	if _, err := io.Copy(hash, in); err != nil {
-		return "", nil
+		return "", err
 	}
 	return hex.EncodeToString(hash.Sum(nil)), nil
 }


pkg/repo/index.go
```diff
@@ -251,7 +251,7 @@ func IndexDirectory(dir, baseURL string) (*IndexFile, error) {
 	for _, arch := range archives {
 		fname, err := filepath.Rel(dir, arch)
 		if err != nil {
-			return index, err
+			return nil, err
```
bacongobbler: This was intended as we want to return the current state of the index during the parsing stage. If it fails halfway through, the user may want to know what was indexed prior to failure.

pkg/repo/index.go
```diff
@@ -270,7 +270,7 @@ func IndexDirectory(dir, baseURL string) (*IndexFile, error) {
 		}
 		hash, err := provenance.DigestFile(arch)
 		if err != nil {
-			return index, err
+			return nil, err
```
bacongobbler: Same here.

bacongobbler (changes requested):

tariq1890:
Thanks for the review comments. Fixed.

bacongobbler (approved):

PR #5802, opened 2019-05-27
replace ReplaceAll to Replace

wutongjie23hao:
When I build helm in mac machine (go version go1.10.1 darwin/amd64) . There is something wrong like this: "undefined: strings.ReplaceAll". It's because ReplaceAll is not in latest go version after 1.10. So, it's necessary to change ReplaceAll to Replace. And it worked.

**If applicable**:


bacongobbler:
`strings.ReplaceAll` has been there for quite some time: https://golang.org/pkg/strings/#ReplaceAll

I'd try upgrading to Go 1.12. Helm 3 is being actively developed on the latest version of Go.

PR #5813, opened 2019-05-29
ref(*): expose klog flags

adamreese:


bacongobbler (approved):
LGTM! Had to do a quick check on why the short flags were removed; it's because klog reserves `-v` for verbosity, which makes sense here.

https://github.com/kubernetes/klog/blob/78315d914a8af2453db4864e69230b647b1ff711/klog.go#L444

adamreese:
flag naming conflicts end in panic

Issue #5779, opened 2019-05-22
helm 3: allow adding custom template functions to engine


JeanMertz:
At our company, we're using a custom utility to build/deploy our helm charts:

https://github.com/blendle/kubecrt

I've been experimenting with upgrading this utility to the Helm 3 alpha.

Unfortunately, it seems like v3 no longer provides access to the map of functions that will be used when rendering the charts.

Specifically, we [add (back) the `env` and `expandenv`](https://github.com/blendle/kubecrt/blob/d8602296ef2561632fe58cf45fb96dfbc18265cc/chartsconfig/parser.go#L36-L45) functions, as they provide us with the flexibility needed for our (private) deployment pipeline.

Whereas in 2.x, we could access those functions using `renderer.FuncMap`, it seems like [they are now made private](https://github.com/helm/helm/blob/b9a54967f838723fe241172a6b94d18caf8bcdca/pkg/engine/engine.go#L79-L117), and we're unable to modify them.

Any thoughts on making this use-case possible again?

trevex:
I would second this request, we have a similar use case, where we reference secrets from Hashicorp Vault and retrieve them with custom logic.
It would be great to extend the `renderer.FuncMap` via https://github.com/helm/community/blob/master/helm-v3/005-plugins.md

bacongobbler:
dupe of #5826

PR #5815, opened 2019-05-29
helm3: Use environment variables in templates

dvob:
This PR removes the deletion of the `env` and `expandenv` template functions of sprig. It would be nice if one could use environment variables within templates without passing all values by `--set`. 

According to https://helm.sh/docs/charts_tips_and_tricks/ they were removed to not expose the environment of tiller, so it should be ok to make them available again in helm3.

This PR should also fix the problem of the reporter of #5779 .

**If applicable**:

### DIFF
--- a/pkg/engine/engine_test.go
+++ b/pkg/engine/engine_test.go
@@ -62,13 +62,6 @@ func TestSortTemplates(t *testing.T) {
 
 func TestFuncMap(t *testing.T) {
 	fns := funcMap()
-	forbidden := []string{"env", "expandenv"}
-	for _, f := range forbidden {
-		if _, ok := fns[f]; ok {
-			t.Errorf("Forbidden function %s exists in FuncMap.", f)
-		}
-	}
-
 	// Test for Engine-specific template functions.
 	expect := []string{"include", "required", "tpl", "toYaml", "fromYaml", "toToml", "toJson", "fromJson"}
 	for _, f := range expect {
--- a/pkg/engine/funcs.go
+++ b/pkg/engine/funcs.go
@@ -44,8 +44,6 @@ import (
 //
 func funcMap() template.FuncMap {
 	f := sprig.TxtFuncMap()
-	delete(f, "env")
-	delete(f, "expandenv")
 
 	// Add some extra functionality
 	extra := template.FuncMap{


bacongobbler:
This may still be considered a big security risk. By providing charts access to the environment, we could be exposing details about the running environment the user may not want the chart to have. The fact that the rendering engine was exposing Tiller's environment is a much larger security risk, however the same risk still applies even if Tiller was being run locally.

For example, a common use case for Helm is to be [embedded inside an Operator](https://github.com/operator-framework/operator-sdk/blob/master/doc/helm/user-guide.md), running inside the cluster as a pod. Exposing the environment would be just as dangerous in that case as it was exposing Tiller's environment, and I'm not sure we want that.

@joelanford any thoughts on this one?

joelanford:
Yeah I generally agree. I think the issue is slightly mitigated for the helm-operator use case because the operator (and its embedded tiller with Helm v2) only has access to the charts that the operator developer explicitly included. Also, the operator is generally given only the minimal set of permissions required to manage releases for those charts.

But it sounds like the fundamental issue is still the same. The trust just moves from the chart to the operator. And I don't think this changes at all for the operator use case in Helm 3.

dsbrng25b:
I would like to run helm in a CI/CD process where secrets are passed over environment variables. So it would be convenient to have access to them directly from the chart. Otherwise i need additional logic to write environment variables to a values.yaml and then use them from there. Passing them with `--set` is not really an option, because then the secrets are visible from `ps -ef`.

How do these operators use helm? Do they import the helm code and use helm programmatically or do they call the helm binary?

bacongobbler:
> Passing them with `--set` is not really an option, because then the secrets are visible from `ps -ef`.

You could write those environment variables into a YAML file and pass that in via `-f/--values`. I know we did that often in our CI pipelines with some assistive tooling.

Given the security risks involved with allowing access to the runtime, and the use cases brought up here can be worked around with alternative means, I don't think we should go ahead with this. If there's a larger demand for shell or filesystem access during template execution we can look into this further, as we will likely need to come up with a larger overall design proposal. Thanks for bringing this up!

bacongobbler:
> How do these operators use helm? Do they import the helm code and use helm programmatically or do they call the helm binary?

They import the helm code and use helm programmatically. Have a look at https://github.com/operator-framework/operator-sdk/search?utf8=%E2%9C%93&q=helm

ruruiye:
+1 for this feature, it will be convenient. we can disable it by default, but add an flag to enable it.

devopsmica:
You can replace the env variable on the CI in the values.yaml file with a one line bash script (giving that the env variables exists on the CI) : 
`envsubst < values.yaml |& tee values.yaml`

jeffmccune:
For the use case of `helm install -f values.yaml stable/fooapp`, Helm should support reading environment values directly.  Writing secret values (e.g. credentials) to an intermediate values file is less secure than reading the secrets directly from the environment.  This avoids the increased risk and exposure of the filesystem.

For the use case of CI, there is additional risk of exposing the secret values via build artifacts which are often automatically copied off the the filesystem.

joelanford:
I think there's still a legitimate security concern with the possibility of unknowingly leaking environment details to a chart. A malicious chart could start up a pod that exfils all environment variables.

Perhaps a middle ground would be for `helm install` (and other subcommands) to only pass through environment variables that have a prefix that can be reasonably assumed to be set explicitly for the purpose of passing values to a chart for rendering.

johnnyhuy:
I agree with having just the `helm install` `helm upgrade --install` method with perhaps an additional flag `--env-passthrough` to read environment variables with a prefix as @joelanford mentioned.

treksler:
> Perhaps a middle ground would be for `helm install` (and other subcommands) to only pass through environment variables that have a prefix that can be reasonably assumed to be set explicitly for the purpose of passing values to a chart for rendering.

That's how terraform does it

PR #5185, opened 2019-01-18
Fix nested null value overrides

aeijdenberg:
Closes #5184 

**What this PR does / why we need it**:

This PR adds tests for setting nested values to `null` and ensure such values are actually deleted.

**Special notes for your reviewer**:

I've split this into 4 commits. They do the following:

1. Adds some code to the tests to make it easier to add nested null values tests.
2. Adds tests that fails.
3. Does a fairly minimal fix - this is enough to make the tests pass.
4. The last commit touches a bit more code, and attempts to simplify the coalescence more broadly by consolidating on to a single codepath, and being a bit more careful about which maps are mutated. Tests continue to pass - but I'm a little less confident about the full impact of this commit.

**If applicable**:

- [x] this PR contains unit tests

PR #5821, opened 2019-05-31
Refactor to use common codepath for table coalescing

aeijdenberg:
This PR is split out from #5185 and adds a 2nd commit to refactor the table coalescing to a single code path.

### DIFF
--- a/pkg/chartutil/values.go
+++ b/pkg/chartutil/values.go
@@ -172,9 +172,7 @@ func CoalesceValues(chrt *chart.Chart, vals *chart.Config) (Values, error) {
 		}
 	}
 
-	var err error
-	cvals, err = coalesceDeps(chrt, cvals)
-	return cvals, err
+	return coalesceDeps(chrt, cvals)
 }
 
 // coalesce coalesces the dest values and the chart values, giving priority to the dest values.
@@ -186,8 +184,7 @@ func coalesce(ch *chart.Chart, dest map[string]interface{}) (map[string]interfac
 	if err != nil {
 		return dest, err
 	}
-	coalesceDeps(ch, dest)
-	return dest, nil
+	return coalesceDeps(ch, dest)
 }
 
 // coalesceDeps coalesces the dependencies of the given chart.
@@ -203,7 +200,7 @@ func coalesceDeps(chrt *chart.Chart, dest map[string]interface{}) (map[string]in
 			dvmap := dv.(map[string]interface{})
 
 			// Get globals out of dest and merge them into dvmap.
-			coalesceGlobals(dvmap, dest, chrt.Metadata.Name)
+			dvmap = coalesceGlobals(dvmap, dest, chrt.Metadata.Name)
 
 			var err error
 			// Now coalesce the rest of the values.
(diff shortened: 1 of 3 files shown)


aeijdenberg:
Rebased on `master` since #5185 is now merged.

Now the line counts more accurately reflect this as a net code removal, which was the intention.

pkg/chartutil/values.go
```diff
@@ -326,36 +275,50 @@ func coalesceValues(c *chart.Chart, v map[string]interface{}) (map[string]interf
...
+		}
+
+		srcTable, srcIsTable := val.(map[string]interface{})
+		dstTable, dstIsTable := dv.(map[string]interface{})
+		switch {
+		case srcIsTable && dstIsTable: // both tables, we coalesce
+			rv[key] = coalesceTables(dstTable, srcTable, chartName)
+		case srcIsTable && !dstIsTable:
+			log.Printf("Warning: Merging destination map for chart '%s'. Cannot overwrite table item '%s', with non table value: %v", chartName, key, val)
+			// despite message in warning, we appear to do exactly that, and do take the dst value
```
thomastaylor312: @scottrigby and I spent some time together going over the original algorithm and I am not sure how this comment is true. Based on the original logic there are 4 conditions (using `map` instead of "table" in all cases):

1. If the key does not exist in the destination map, set the incoming value
2. If the incoming value is a `map` and the existing key value is a `map`, merge
3. If the incoming value is not a `map` and the existing key value is a `map`, leave the existing value alone
4. If the existing key is not a `map` and the incoming value is a `map`, leave the existing value alone

Essentially 3 and 4 are the same and could be reworded as "if only one of the destination or incoming value is a `map`, leave the existing value alone"

The logic here reads much nicer (thanks!) but I think has different behavior. Were you observing something different in your testing? I also know that the original code is spaghetti code, so we could have read something wrong as well

thomastaylor312: If we are correct (and please let us know if we are not!) I would clarify the log messages to be something of lines of "cannot override a non-map value with a map value" and remove the `rv[key] = dv` lines in those blocks

thomastaylor312 (changes requested):
I really like how this simplifies things. Just some questions around the merge algorithm refactor to address. It would also be helpful to add unit tests to catch the edge cases where value types don't match.

pkg/chartutil/values.go
aeijdenberg: @thomastaylor312 @scottrigby 

I added the following to `pkg/chartutil/values_test.go`:

```golang
func TestCannotOverriteTableItemWithNonTableValue(t *testing.T) {
	// src has a table value for "foo"
	src := map[string]interface{}{
		"foo": map[string]interface{}{
			"baz": "boz",
		},
	}

	// dst has a non-table value for "foo"
	dst := map[string]interface{}{
		"foo": "bar",
	}

	// warning claims that "foo" will not be overwritten, yet:
	log.Println("Result: ", coalesceTables(dst, src, ""))
}
```

Then:

```bash
$ go test ./pkg/chartutil -run TestCannotOverriteTableItemWithNonTableValue -v
=== RUN   TestCannotOverriteTableItemWithNonTableValue
2019/06/07 15:05:35 Warning: Merging destination map for chart ''. Cannot overwrite table item 'foo', with non table value: map[baz:boz]
2019/06/07 15:05:35 Result:  map[foo:bar]
--- PASS: TestCannotOverriteTableItemWithNonTableValue (0.00s)
PASS
ok      k8s.io/helm/pkg/chartutil       0.006s
```

The warning is wrong because:

1. The value that it prints is actually the `src` value, which is being overwritten *by* the `dst` value, although as written it implies that's the value that you are trying merge on top.
2. It's a no-op because it's the `dst` map (which has the `dst` value) that is actually returned, the `src` value is ignored in the result as shown in the output above.

The refactor preserves the above behavior and warning, even though I think the warning is inaccurate.

In terms of correct behavior - well I'm not entirely sure - and I haven't used Helm enough to get a good sense for it. We only got down this rabbit hole in trying to solve a bug that we encountered. My first impression is that the current behaviour, where the destination value totally overwrites the source in the event of a type mismatch is reasonable, however it shouldn't print the inaccurate warning. Alternatively one could take the view that a type mismatch should result in an error and fail processing. Either way, I think a warning is wrong. Either allow it or fail fast.

thomastaylor312: Ok, that makes sense then, thanks for the clarification! And I agree with you about warning vs failing, but if we begin failing, it will break backwards compatibility (things will start breaking where they didn't before). Can you swap the warning message to be accurate and more clear? After that I think this should be good to go

scottrigby (dismissed):

scottrigby (changes requested):
Just message change is needed to match behavior.

aeijdenberg:
Have added the test above into a [separate commit](https://github.com/helm/helm/pull/5821/commits/29a295eb84ffdfd00e3174818123f37033a02a3e) which I added first, so that we can see the [tests pass](https://circleci.com/gh/helm/helm/8490) before the refactor. 

Then I re-based the refactor, and the warning message change commit on top.

I think that's all that's needed?

scottrigby (approved):

thomastaylor312 (approved):
Thanks for all the great work on this @aeijdenberg! We really appreciate it

PR #5824, opened 2019-05-31
fix issue that values are not applied while running upgrade cmd

### LINKED ISSUE #5792: Helm v3 - upgrade fails to apply value changes
Output of `helm version`:
```
version.BuildInfo{Version:"v3.0.0-alpha.1", GitCommit:"b9a54967f838723fe241172a6b94d18caf8bcdca", GitTreeState:"clean"}
```

## What happens
When running `helm upgrade --set foo=bar RELEASE CHART` the upgrade process does not take into account the newly set values of `foo: bar`.

## How to reproduce
1. Install a chart with `helm install RELEASE CHART` with any values you want.
2. Run an upgrade with a new values.yaml with `-f new-values.yaml` or by passing values in the command line `--set foo=bar`.
3. Check the values with `helm get values RELEASE` and see `null` as the return value or with `helm get values RELEASE -a` and see the previous values from the install without the new `foo` value. You can also verify the chart output with `--dry-run --debug` when modifying a value that is used within the chart as you'll see that the value isn't updated.

Interestingly enough, if you were to modify the values in the original chart repository (for local charts) and run the upgrade with `helm upgrade RELEASE CHART -f CHART/values.yaml` it will work as intended.

rokii:
closes #5792
It should not define additional values as ValueOptions will consolidate from various value sources to rawValues. 
This PR also fixes an UT assert bug.

### DIFF
--- a/cmd/helm/testdata/output/upgrade-with-missing-dependencies.txt
+++ b/cmd/helm/testdata/output/upgrade-with-missing-dependencies.txt
@@ -1,3 +1,1 @@
-Error: "helm upgrade" requires 2 arguments
-
-Usage:  helm upgrade [RELEASE] [CHART] [flags]
+Error: found in Chart.yaml, but missing in charts/ directory: reqsubchart2
--- a/cmd/helm/testdata/testcharts/upgradetest/templates/configmap.yaml
+++ b/cmd/helm/testdata/testcharts/upgradetest/templates/configmap.yaml
@@ -1,0 +1,7 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: "{{ .Release.Name }}-configmap"
+data:
+  myvalue: "Hello World"
+  drink: {{ .Values.favoriteDrink }}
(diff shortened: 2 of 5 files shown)


karuppiah7890:
@rokii Thanks for this PR! 😄 🎉 

I just tried this and it worked when I used `--set`. But weirdly it didn't work for me when I just used `--values`. 

On checking, looks like we need to change this code too - https://github.com/helm/helm/blob/871b092f32ea5920dbec881f75e3f9caa3db2748/pkg/action/upgrade.go#L270

It should check if the values in the upgrade command are empty. But it's checking only the `--set` but like you mentioned, there are other sources from where values come from, so it should use `u.rawValues` and not `u.Values`. I think we need to write test for that too so that later it doesn't get missed 😄

rokii:
@karuppiah7890 thx for pointing out that. And I also add more test cases including other two source values.

hickeyma (approved):
LGTM, thanks. Requires another approval.

bacongobbler:
thanks @rokii. Taking a closer look at this today.

morsok:
I had the same issue and used your fork for this PR to build from source and it worked for me, thank you !

bacongobbler (approved):
LGTM. Nice catch.

PR #5828, opened 2019-06-01
[documentation] make `install` command doc more explicit

paulczar:
The `helm install` command can have unexpected behavior when running `helm install stable/foo` which could
be both a local chart directory or a chart reference.  Providing some notes to this effect seems prudent.

Signed-off-by: Paul Czarkowski <username.taken@gmail.com>

**If applicable**:

### DIFF
--- a/docs/helm/helm_install.md
+++ b/docs/helm/helm_install.md
@@ -57,6 +57,8 @@ There are five different ways you can express the chart you want to install:
 4. By absolute URL: helm install https://example.com/charts/nginx-1.2.3.tgz
 5. By chart reference and repo url: helm install --repo https://example.com/charts/ nginx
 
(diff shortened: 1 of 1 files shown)


hickeyma (changes requested):
Thanks for the PR @paulczar. This document is a generated doc. The document you need to update is: https://github.com/helm/helm/blob/master/cmd/helm/install.go#L99 . When you update this you will need to run `make docs` to regenerate the doc from it. Refer to [here](https://helm.sh/docs/developers/#man-pages) for more details.

bridgetkromhout:
Hi, @paulczar - good news! It's much much less work to edit Helm docs moving forward, as we're moving them into the www repo. Would you mind putting this edit here:
https://github.com/helm/helm-www/blame/dev-v3-docs/content/docs/helm/helm_install.md#L59

That's the `dev-v3-docs` branch for now. Later of course the master branch in `helm/helm-www` would be the right place. 

We'll close out this PR but would appreciate that edit over in the right place. Thanks!

PR #5829, opened 2019-06-03
helm v3: dynamic completion of releases

marckhouzam:
Add dynamic completion after
```
helm status <TAB>
helm uninstall <TAB>
helm history <TAB>
helm test run <TAB>
helm upgrade <TAB>
helm get <subcommand> <TAB>
helm rollback <TAB>
```
by fetching the list of releases from the cluster.

This PR is the adapted v3 version of #5562 as requested by @technosophos.
It puts in the foundation for dynamic completion support in Helm v3.
It is inspired by the dynamic completion provided by kubectl.
Both bash and zsh are handled.

As dynamic completion actually sends queries to the cluster, this PR handles the use of Helm flags that affect which cluster should be contacted:
```
--kubeconfig 
--kube-context 
--home 
--namespace
-n
```

This PR requires the use of Cobra v0.0.4 which allows dynamic completion to co-exist between helm and kubectl (while Cobra v0.0.3 had a bug that would cause the two to clash).

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -879,15 +879,15 @@
   version = "v1.4.1"
 
 [[projects]]
-  digest = "1:e01b05ba901239c783dfe56450bcde607fc858908529868259c9a8765dc176d0"
+  digest = "1:2e72f9cdc8b6f94a145fa1c97e305e1654d40507d04d2fbb0c37bf461a4b85f7"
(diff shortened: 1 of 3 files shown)


bacongobbler (approved):
LGTM. Thanks for working on this!

PR #5840, opened 2019-06-04
doc: tuple replaced by list

badeadan:
All list functions also work on tuple (implemented in Sprig as list, see https://github.com/Masterminds/sprig/blob/2625cd487a689ca112e1301a58b89b347ba2c994/functions.go#L237 ) and for all purposes a 'tuple' is a 'list'. Replace documentation ref. to 'tuple' with 'list' so there is no need to explain what a tuple is.

Behaviour-wise 'list' and associated functions return tuples because values are immutable:
```
  {{ $t := tuple 1 2 -}}
  {{- $t_plus := append $t 3 -}}
  {{ $l := list 4 5 -}}
  {{- $l_plus := append $l 6 -}}
  t: {{ $t }}
  t_plus: {{ $t_plus }}
  l: {{ $l }}
  l_plus: {{ $l_plus }}
```
renders to (note append on a list should update and return source list):
```
  t: [1 2]
  t_plus: [1 2 3]
  l: [4 5]
  l_plus: [4 5 6]
```
where _$l_ is not updated after append as expected. However I think just using 'list' is clearer.

**If applicable**:

### DIFF
--- a/docs/chart_template_guide/accessing_files.md
+++ b/docs/chart_template_guide/accessing_files.md
@@ -54,13 +54,13 @@ metadata:
   name: {{ .Release.Name }}-configmap
 data:
   {{- $files := .Files }}
-  {{- range tuple "config1.toml" "config2.toml" "config3.toml" }}
+  {{- range list "config1.toml" "config2.toml" "config3.toml" }}
   {{ . }}: |-
     {{ $files.Get . }}
   {{- end }}
 ```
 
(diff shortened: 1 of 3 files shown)


bacongobbler (approved):
This function seems to have been part of sprig for quite some time, which means this documentation LGTM as it applies to more than just the latest version of Helm.

PR #5841, opened 2019-06-04
fix(circle): only upload packages and checksums, not cross-builds

bacongobbler:
I noticed that the assets from `make build-cross` was being uploaded along with the release assets. Restricting `az storage blob upload-batch` to just the release assets prevents that from happening going forward.


mattfarina (approved):
I see the problem this fixes. Nice catch.

PR #5842, opened 2019-06-04
Expose the resource types in addition to the api group/version in templates

mattfarina:

Detecting the resources installed is becoming more important with the increasing use of CRDs. Is a particular CRD installed? API Groups may exist by additional CRDs may be installed in that group later. Currently, in templates the way to detect resources is to use `Capabilities.APIVersions.Has`. This only has the group/version and not the resources within it.

The change this PR makes is to add each of the resource types to the list used by that template function. Now you can do things like `Capabilities.APIVersions.Has "apps/v1/Deployment"` and it can detect if that is available. The same applies to CRDs.

cc @adamreese since doing it this way was your idea.

**If applicable**:


bacongobbler:
I just answered a question on Slack that this very PR would solve, so I think there's definitely some value in this feature.

@mattfarina are you interested in porting this over to Helm 3 as well? This seems like it'd be useful for both Helm 2 and Helm 3.

bacongobbler (approved):
LGTM.

mattfarina:
@bacongobbler my plan was to do a Helm 3 port if this got in.

xor22h:
When v2.15 is going to be out?

PR #5848, opened 2019-06-05
chore(testdata): remove stale test output

bacongobbler:
These tests were part of `helm test` before it was refactored as part of the `pkk/action` refactor, which no longer exist.

```
rm -rf cmd/helm/testdata/output/*
go test ./... -update
```


hickeyma:
> rm -rf cmd/helm/testdata/output/*

@bacongobbler Are you saying all the test data in above directory is now obsolete or just the files you are removing?

bacongobbler:
Those were the commands used to regenerate the testdata golden files. These files were never regenerated after `go test -update` because no test output generates them

hickeyma (approved):
LGTM, thanks.

PR #5849, opened 2019-06-06
Use the latest version of Cobra

marckhouzam:

As suggested in PR #5562 this upgrade of Cobra is to keep helm v2 in sync with helm v3.  Looking at the changes in Cobra that will be brought in, I believe this change is safe.


marckhouzam:
As suggested by @bacongobbler

hickeyma (approved):
LGTM, thanks @marckhouzam

PR #5854, opened 2019-06-07
fix inline code formatting in helm install

Scott8440:
**What this PR does / why we need it**: Fixes inline code formatting in helm_install.md. Previously this code block was plain text when it should have been an indented block.

**If applicable**:


bacongobbler:
You'll need to make the change in cmd/helm/helm_install.go then call `make docs`. Let me know if you need help with that.

PR #5859, opened 2019-06-07
add go version to version cmd output of helm

tariq1890:
This is the helm v3 equivalent of #5693. Life sure is simple without tiller :).

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -20,8 +20,10 @@ package version // import "helm.sh/helm/pkg/version"
 type BuildInfo struct {
 	// Version is the current semver.
 	Version string `json:"version,omitempty"`
-	// GitCommit is the git sha1
+	// GitCommit is the git sha1.
 	GitCommit string `json:"git_commit,omitempty"`
-	// GitTreeState is the state of the git tree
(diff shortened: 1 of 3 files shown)


pkg/version/version.go
```diff
@@ -20,8 +20,10 @@ package version // import "helm.sh/helm/pkg/version"
...
 	// Version is the current semver.
 	Version string `json:"version,omitempty"`
-	// GitCommit is the git sha1
+	// GitCommit is the git sha1.
 	GitCommit string `json:"git_commit,omitempty"`
-	// GitTreeState is the state of the git tree
+	// GitTreeState is the state of the git tree.
 	GitTreeState string `json:"git_tree_state,omitempty"`
+	// GoVersion is the version of the Go compiler used.
+	GoVersion string `json:"git_tree_state,omitempty"`
```
bacongobbler: json tag is wrong

bacongobbler (changes requested):

pkg/version/version.go
tariq1890: Thanks. Will fix.

bacongobbler (approved):

mattfarina (approved):

PR #3454, opened 2018-02-05
ADD support for updating TILLER_HISTORY_MAX when upgrading

mattclegg:
> When doing a helm init --max-history=N --upgrade --debug, the returned deployment manifest shows TILLER_HISTORY_MAX env var set on the deployment with the correct value.
> But when doing a kubectl log <tiller_pod>, it shows Max history per release is 0. Doing a describe on the deployment also shows that the TILLER_HISTORY_MAX is missing.

PR #5867, opened 2019-06-10
 Add support for setting --history-max on existing clusters

### LINKED ISSUE #3414: Set --history-max on a already deployed cluster
for the sake of cleanlyness, I'm really interested on adding this option to my cluster.

Reading various thread didn't lead me to a clear path.
I tried various things without success. Would be glad to get some help.

```
helm init --upgrade --history-max=10
```

I think this did work, but I'm still seeing my old ConfigMaps (even after an upgrade of the chart)

How can I confirm that Tiller is configured now. And how can I delete old revisions?
(running v2.8.0)

Thanks!

rossf7:

This PR sets `--history-max` on existing clusters. It is a replacement for #3454 and includes the unit tests added in af316d3. I've tested locally and is working OK.

closes #3414

I changed the unit tests in af316d3 quite a lot to try and sync them with the other tests. The code change is basically unchanged from the original PR.

**If applicable**:

### DIFF
--- a/cmd/helm/installer/install.go
+++ b/cmd/helm/installer/install.go
@@ -30,6 +30,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/kubernetes"
 	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	extensionsclient "k8s.io/client-go/kubernetes/typed/extensions/v1beta1"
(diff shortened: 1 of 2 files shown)


cmd/helm/installer/install.go
```diff
@@ -71,6 +72,11 @@ func Upgrade(client kubernetes.Interface, opts *Options) error {
 	}
 	obj.Spec.Template.Spec.Containers[0].Image = opts.SelectImage()
 	obj.Spec.Template.Spec.Containers[0].ImagePullPolicy = opts.pullPolicy()
+	if opts.MaxHistory > 0 {
```
bacongobbler: how would a user unset this value?

rossf7: Good point. If `opts.MaxHistory` is not set should we revert to the default of `0`?

@bacongobbler I'd appreciate hearing what you think is the best approach? Thanks!

cmd/helm/installer/install.go
```diff
@@ -407,3 +423,32 @@ func generateSecret(opts *Options) (*v1.Secret, error) {
 }
 
 func read(path string) (b []byte, err error) { return ioutil.ReadFile(path) }
+
+// updateEnv updates the existing env vars and allows env vars to be removed
+// by name.
+func updateEnv(existing []v1.EnvVar, env []v1.EnvVar, remove []string) []v1.EnvVar {
```
rossf7: Also, this came from the previous PR and I have a doubt now. Do we need to be able to remove env vars?

As doesn't `TILLER_HISTORY_MAX` always need to be set.

alexcurtin:
Any update on this?

bacongobbler:
Hey @alexcurtin, thanks for the ping.

Unfortunately, we won't be able to get this feature into 2.15, which will be the last feature release for Helm 2. Because Tiller is no longer present in Helm 3, it is hard to justify major changes (which could incur support overhead) to that subsystem.

The good news is that [workarounds exist](https://github.com/helm/helm/issues/3414#issuecomment-361497289), and because Tiller is not present in Helm 3, when you've upgraded you should no longer need this workaround.

Thanks!

PR #5869, opened 2019-06-10
ref(version): catch some edge cases

bacongobbler:
- strip GoVersion from `helm version` test output
- catch some edge cases when GitCommit is unset or less than 7 characters (e.g. during integration tests)
- add a test case for `helm version --short`


hickeyma (approved):
LGTM, thanks

Issue #5866, opened 2019-06-10
Dependency charts with multiple versions being overridden in tar balls


jpninanjohn:
[issue.zip](https://github.com/helm/helm/files/3271903/issue.zip)

Output of `helm version --client`:
```
Client: &version.Version{SemVer:"v2.14.0", GitCommit:"05811b84a3f93603dd6c2fcfe57944dfa7ab7fd0", GitTreeState:"clean"}
```

**Actual:**
When trying to install a chart with multiple versions of the same dependency, we see that the tarball created has only one version of the dependency. (The chart used is attached as a zip file).

If we try to install this release only the one dependency is created. 

`helm install --name issue dummy-chart-0.1.0.tgz` -> This only installs the one dependency that is in the tarball. 

**Expected:** 
All the dependencies should be installed with proper versions.

**Additional Info:**
We also tried with the same chart to install the release without creating the tarball and it creates it correctly.
`helm install --name issue <chart-directory>` ->This works and gives expected output.

**Possible Issue:**

https://github.com/helm/helm/blob/2ca025d48222d6fa188653e2ca5eda6ed799145c/pkg/chartutil/save.go#L167-L209

We are thinking that the filename is getting overwritten when creating the tarball. As an experiment, we tried changing the name of the file in the tarball during creation(we appended the chart version) and it worked.
https://github.com/helm/helm/blob/2ca025d48222d6fa188653e2ca5eda6ed799145c/pkg/chartutil/save.go#L168

karuppiah7890:
cc @hickeyma

PsyonixBarrett:
Any movement on this issue? The bug appears in helm v3 beta 3 as well. We have several "service/application" charts that rely on a single "common" chart to build up their components.

The problem we see is that we can't pin the service charts to utilize specific versions of the common chart. This could cause issue for us down the line if we need/want to introduce breaking changes in the common chart.

bacongobbler:
closing as per discussion in #5872.

PR #5872, opened 2019-06-11
Fix packaging of multiple versions of same dependency

jpninanjohn:
Signed-off-by: Ninan John <jninan@thoughtworks.com>

Closes this issue -> https://github.com/helm/helm/issues/5866
Fix:
Added a flag that appends the version to the filename when packaging into a tarball for dependencies.

I have added a test to check if the tarball has all the versions of the same dependencies. 

**If applicable**:

### DIFF
--- a/pkg/chartutil/save.go
+++ b/pkg/chartutil/save.go
@@ -158,14 +158,19 @@ func Save(c *chart.Chart, outDir string) (string, error) {
 		}
 	}()
 
-	if err := writeTarContents(twriter, c, ""); err != nil {
+	if err := writeTarContents(twriter, c, "", false); err != nil {
 		rollback = true
 	}
 	return filename, err
 }
 
-func writeTarContents(out *tar.Writer, c *chart.Chart, prefix string) error {
-	base := filepath.Join(prefix, c.Metadata.Name)
+func writeTarContents(out *tar.Writer, c *chart.Chart, prefix string, isDependency bool) error {
+	var base string
+	if isDependency {
+		base = filepath.Join(prefix, fmt.Sprintf("%s-%s", c.Metadata.Name, c.Metadata.Version))
+	} else {
+		base = filepath.Join(prefix, c.Metadata.Name)
+	}
 
 	// Save Chart.yaml
 	cdata, err := yaml.Marshal(c.Metadata)
@@ -201,7 +206,7 @@ func writeTarContents(out *tar.Writer, c *chart.Chart, prefix string) error {
 
 	// Save dependencies
 	for _, dep := range c.Dependencies {
-		if err := writeTarContents(out, dep, base+"/charts"); err != nil {
+		if err := writeTarContents(out, dep, base+"/charts", true); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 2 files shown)


tamalsaha:
@jpninanjohn , please see the relevant discussion here: https://kubernetes.slack.com/archives/C51E88VDG/p1561059858120100

It seems that you need to target this for Helm 3 and handle various cases that @bacongobbler  mentioned in slack.

bacongobbler:
With this fix, does it allow you to install both versions in one `helm install`? What's the use case for having two versions of the same dependency managed in a single release object? Won't the objects from both charts collide with each other?

karuppiah7890:
@bacongobbler about the use case for two versions - our team works with lot of charts to deploy components needed for a service. So, for example, for service A, there will be a chart with no templates but multiple dependencies in requirements yaml, like redis chart, postgres chart. And if the service needs more than one redis or postgres, we will have multiple redis and postgres charts, with different aliases. It also usual where a team creating the service wants to add another new redis for some new feature and when we add it to requirements, if the latest redis chart has changed with some new features, possibly breaking too, we still have to add it. The reason for not upgrading the chart version of existing redis dependencies when adding a latest version redis chart - well, it depends, based on the effects of the upgrade. For example
1. If the upgrade is a breaking change, it's quite some work and we might want to put it off for later to upgrade all the old chart versions in all the service components charts and make necessary changes for the breaking change, and possibly try to automate it
2. If the change causes changes in the K8s resources that it will cause a restart of pods. We might want to not do this - as it will cause some downtime which we may want to postpone it for the night based on need.

With the current issue of having only one version of a chart in all the instances in the dependencies, we are forced to use only one version (either old or new). I feel the users should not be forced to do so

bacongobbler:
> And if the service needs more than one redis or postgres, we will have multiple redis and postgres charts, with different aliases

Aliases only change the name of the package on-disk but doesn't affect the templates themself when being rendered as far as I remember. redis-1 and redis-2 will still collide as far as I can remember.

I need to do some testing first to bring myself back up to speed; it's been a while since I've played with chart aliases.

karuppiah7890:
And that's the issue. Currently the different versions of the same chart
are put together in a single folder inside the charts dir and like you said
they collide. Instead we want them to be separate

jpninanjohn:
@bacongobbler I tried a `helm install` with the fix and it does install 2 different versions of the specified chart. The usecase is as @karuppiah7890 mentioned.

hickeyma:
@jpninanjohn Thanks for your work on this PR. 

I am unsure that the changes are in alignment with how dependencies are designed to work. I understand that it fixes a use case you have but not sure of the side effects of such a change. 

Taking into account that Helm 2.15 is near to release, it might be better considering this for Helm 3 instead? That way there would be more time to review the change and any consequences.

@bacongobbler @thomastaylor312 What do you think?

bacongobbler:
I agree, @hickeyma. After closer review, I noticed that this is actually introducing a new use case that originally was not accounted for, hence the change from `bug` to `feature`. The dependency resolver has not before had to resolve two identical dependencies with differing versions before, and I'm concerned that the use case described here is not something we've carved out a good story for.

I am +1 on pushing this out to Helm 3. That way we can give this use case a more properly thought-out review and determine if there are edge cases we may be missing.

bacongobbler:
Closing as we've cut Helm 2.15 and will no longer be accepting feature requests for Helm 2. Thanks!

PR #5873, opened 2019-06-11
ref(chartutil): introduce chartutil.Save

bacongobbler:
chartutil.Save has been renamed to chartutil.SaveArchive.

chartutil.Save accepts a chart and an io.Writer, wrapping the logic around serializing a chart to an io.Writer.

This allows other packages to write higher-level abstractions above chart.Save. For example, streaming a chart to a *bytes.Buffer.

### DIFF
--- a/pkg/chartutil/save.go
+++ b/pkg/chartutil/save.go
@@ -20,6 +20,7 @@ import (
 	"archive/tar"
 	"compress/gzip"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
@@ -33,6 +34,11 @@ import (
 var headerBytes = []byte("+aHR0cHM6Ly95b3V0dS5iZS96OVV6MWljandyTQo=")
 
 // SaveDir saves a chart as files in a directory.
+//
+// This takes an existing chart and a destination directory.
+//
+// If the directory is /foo, and the chart is named bar, with version 1.0.0, this
+// will generate /foo/bar/Chart.yaml, /foo/bar/values.yaml, etc.
 func SaveDir(c *chart.Chart, dest string) error {
 	// Create the chart directory
 	outdir := filepath.Join(dest, c.Name())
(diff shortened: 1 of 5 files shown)


mattfarina (commented):
It may be overkill but should there be another test for just `Save`?

adamreese:
Do the verbs import and export help to clarify what each function is doing?

bacongobbler:
> Do the verbs import and export help to clarify what each function is doing?

Perhaps? `.Save` seems fairly clear to me the intent of the function so it's not the verbs that are tripping me up, but the suffix after the verb. Take `SaveDir` for example: does that archive a chart as a .tgz and write that to a directory, or does it write the contents of a chart in a directory? The naming is slightly ambiguous, causing a little bit of confusion based on the name, so I have to look at the code to inspect what is the function's intended use.

Looking through the code, we have several functions that all deal with charts that are archived. I'll give this package a bit more thought on the design and how we want to address charts coming from multiple different formats (.zip, .bzip, .xz, for example).

bacongobbler:
> It may be overkill but should there be another test for just `Save`?

If history has taught us anything, more tests are always better :)

pkg/chartutil/save.go
```diff
@@ -33,6 +34,11 @@ import (
 var headerBytes = []byte("+aHR0cHM6Ly95b3V0dS5iZS96OVV6MWljandyTQo=")
 
 // SaveDir saves a chart as files in a directory.
+//
+// This takes an existing chart and a destination directory.
+//
+// If the directory is /foo, and the chart is named bar, with version 1.0.0, this
```
thomastaylor312: Nit: The fact that this chart is version 1.0.0 seems to be irrelevant to the rest of the comment and can probably be dropped

thomastaylor312 (approved):
Super tiny, non-blocking nit. Otherwise this is a great addition

bacongobbler:
closing as irrelevant for now

PR #5874, opened 2019-06-11
fix(resolver): compare hash of lockfile against resolved dependencies

### LINKED ISSUE #2731: requirements.lock does not get updated on helm dep up
if there is no change in `requirements.yaml` file but SemVar results into different lock file, `helm dep up` does not replace and existing `requirements.lock` because `requirements.yaml` digest has not changed.

How to reproduce:

We have the following `requirements.yaml` file for a chart `parent1`.

```
dependencies:
- name: child1
  version: ">=1.0.0"
  repository: "http://127.0.0.1:8879/"
- name: child2
  version: ">=1.0.0"
  repository: "http://127.0.0.1:8879/"
```

When we run dep up for the first time it generates following lock file (assuming that we have `child1-1.0.0` and `child2-1.0.0` charts at that point)

```
dependencies:
- name: child1
  repository: http://127.0.0.1:8879/
  version: 1.0.0
- name: child2
  repository: http://127.0.0.1:8879/
  version: 1.0.0
digest: sha256:19b7f91a0fdfc44e14a86e7b3bbfbce2014e063bdc5949c2ad6a5edb113c3b77
generated: 2017-07-27T12:00:08.036838191+05:30

```

Now if we release a new version of `child1` (say 1.1.0) we don't need to update our `requirements.yaml` file but we expect `helm dep up` to update `requirements.lock` file accordingly. 

Instead, `helm dep up` updates `/charts` folder with `child1-1.1.0` but does not update `requirements.lock` file on disk.

When we looked into code, we figured that `requirements.lock` only get updated if `requirements.yaml` gets changed. 

```
// Hash requirements.yaml
hash, err := resolver.HashReq(req)
if err != nil {
         return err
}

// Now we need to find out which version of a chart best satisfies the
// requirements the requirements.yaml
lock, err := m.resolve(req, repoNames, hash)
if err != nil {
	return err
}

// If the lock file hasn't changed, don't write a new one.
oldLock, err := chartutil.LoadRequirementsLock(c)
if err == nil && oldLock.Digest == lock.Digest {
	return nil
}
```

This results into inconsistent state of having different charts in `/charts` directory and `requirements.lock` file. Only work around, currently, is to delete `requirements.lock` file when we do `helm dep up`

bacongobbler:
closes #2731 

### DIFF
--- a/pkg/downloader/manager.go
+++ b/pkg/downloader/manager.go
@@ -121,13 +121,6 @@ func (m *Manager) Update() error {
 		return nil
 	}
 
-	// Hash dependencies
-	// FIXME should this hash all of Chart.yaml
-	hash, err := resolver.HashReq(req)
-	if err != nil {
-		return err
-	}
-
 	// Check that all of the repos we're dependent on actually exist and
 	// the repo index names.
 	repoNames, err := m.getRepoNames(req)
@@ -144,7 +137,7 @@ func (m *Manager) Update() error {
 
 	// Now we need to find out which version of a chart best satisfies the
 	// dependencies in the Chart.yaml
-	lock, err := m.resolve(req, repoNames, hash)
+	lock, err := m.resolve(req, repoNames)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


mattfarina (approved):
Good call making the hash for change be the lock rather than the requirements. I tested the before and after of this change and I saw it work.

hickeyma (approved):
Tested this and LGTM. Thanks @bacongobbler, good to merge.

PR #5881, opened 2019-06-12
helm 3: support writing multiple resources to the same file

torstenwalter:

Templates can contain multiple manifests. As I made a mistake while implementing `--output-dir` option in #5762  for `helm template` only the last resource is written to that file.

This PR fixes that.

**If applicable**:

### DIFF
--- a/internal/test/test.go
+++ b/internal/test/test.go
@@ -58,6 +58,17 @@ func AssertGoldenString(t TestingT, actual, filename string) {
 	}
 }
 
+// AssertGoldenFile assers that the content of the actual file matches the contents of the expected file
+func AssertGoldenFile(t TestingT, actualFileName string, expectedFilename string) {
+	t.Helper()
+
+	actual, err := ioutil.ReadFile(actualFileName)
+	if err != nil {
+		t.Fatalf("%v", err)
+	}
+	AssertGoldenBytes(t, actual, expectedFilename)
+}
+
 func path(filename string) string {
 	if filepath.IsAbs(filename) {
 		return filename
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -365,22 +365,24 @@ func (c *Configuration) renderResources(ch *chart.Chart, values chartutil.Values
 	}
 
 	// Aggregate all valid manifests into one big doc.
+	fileWritten := make(map[string]bool)
 	for _, m := range manifests {
 		if outputDir == "" {
 			fmt.Fprintf(b, "---\n# Source: %s\n%s\n", m.Name, m.Content)
 		} else {
-			err = writeToFile(outputDir, m.Name, m.Content)
+			err = writeToFile(outputDir, m.Name, m.Content, fileWritten[m.Name])
 			if err != nil {
 				return hs, b, "", err
 			}
+			fileWritten[m.Name] = true
 		}
 	}
 
 	return hs, b, notes, nil
 }
 
-// write the <data> to <output-dir>/<name>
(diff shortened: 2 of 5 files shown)


bacongobbler:
Can you write some unit tests to cover this case?

torstenwalter:
Where should these test be added?
`go test -timeout 30s helm.sh/helm/cmd/helm` is broken on dev-v3 branch. Are those executed?
Or should the unit tests be placed in `pkg/action/...`

torstenwalter:
@bacongobbler Should be ready for review.

hickeyma:
@torstenwalter To clarify here, this is to fix the issue whereby you have more than one resource in the same template file and the output just writes the last resource to the output manifest file? I am  asking because you mention "Templates can contain multiple manifests." in the description and it works for this scenario. I would think that this would also need to be fixed for v2.

pkg/action/install.go
```diff
@@ -405,6 +407,14 @@ func writeToFile(outputDir string, name string, data string) error {
 	return nil
 }
 
+func createOrOpenFile(filename string, append bool) (*os.File, error) {
+	if append {
+		return os.OpenFile(filename, os.O_APPEND|os.O_WRONLY, 0600)
+	} else {
```
hickeyma: You can drop the "else". Small 'linter' nit.

hickeyma (changes requested):
Good catch @torstenwalter. Thanks for the fix. It is looking good and tests as expected on manual tests that I ran. Small fix needed.

pkg/action/install.go
torstenwalter: @hickeyma Can you change this directly? Otherwise I'll do it in a week when I am back at my desk.

torstenwalter:
> @torstenwalter To clarify here, is this is to fix the issue whereby you have more than one resource in the same template file and the output just writes the last resource to the output manifest file? I am asking because you mention "Templates can contain multiple manifests." in the description and it does already work for this scenario.
> 
> I think that this would also need to be fixed for v2.

I think it works in v2.

pkg/action/install.go
hickeyma: Yiu can do it when you return @torstenwalter when you return. You have done all the hard work.

torstenwalter: @hickeyma Change is implemented.

hickeyma (approved):
Thanks @torstenwalter, LGTM. Can you squash the changes?

This will need another review before merge.

torstenwalter:
@hickeyma squashed and rebased on dev-v3

hickeyma:
> I think it works in v2.

Yes, it does.

internal/test/test.go
```diff
@@ -58,6 +58,15 @@ func AssertGoldenString(t TestingT, actual, filename string) {
 	}
 }
 
+// AssertGoldenFile assers that the content of the actual file matches the contents of the expected file
+func AssertGoldenFile(t TestingT, actualFileName string, expectedFilename string) {
+	actual, err := ioutil.ReadFile(actualFileName)
```
bacongobbler: can you add a `t.Helper()` here? See above functions^

bacongobbler (changes requested):

hickeyma (commented):
@torstenwalter do you mind squashing the changes?

bacongobbler (approved):

torstenwalter:
@hickeyma, @bacongobbler Thank you for considering this PR and reviewing it.

PR #5890, opened 2019-06-13
[helm create] Add service account to templates

### LINKED ISSUE #5879: helm create - Make it easier to do the right thing, part i: serviceaccount
Hello,

Helm create should add options to create `serviceaccount`, enabled by default imho, as it is a known best practice to run every workload with its own service account, by including the option to create a service account, we are making it easier to do the right thing.

cc @bacongobbler @unguiculus @scottrigby @rimusz

naseemkullah:
closes #5879 

**If applicable**:

### DIFF
--- a/pkg/chartutil/create.go
+++ b/pkg/chartutil/create.go
@@ -254,7 +254,7 @@ kind: ServiceAccount
 metadata:
   name: {{ template "<CHARTNAME>.serviceAccountName" . }}
   labels:
-  {{ include "<CHARTNAME>.labels" . | indent 4 }}
+{{ include "<CHARTNAME>.labels" . | indent 4 }}
 {{- end -}}
 `
 
--- a/pkg/chartutil/create.go
+++ b/pkg/chartutil/create.go
@@ -42,6 +42,8 @@ const (
 	DeploymentName = "deployment.yaml"
 	// ServiceName is the name of the example service file.
 	ServiceName = "service.yaml"
+	// ServiceAccountName is the name of the example serviceaccount file.
+	ServiceAccountName = "serviceaccount.yaml"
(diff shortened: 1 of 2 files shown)


pkg/chartutil/create.go
```diff
@@ -238,6 +248,15 @@ spec:
     app.kubernetes.io/name: {{ include "<CHARTNAME>.name" . }}
     app.kubernetes.io/instance: {{ .Release.Name }}
 `
+const defaultServiceAccount = `{{- if .Values.serviceAccount.create -}}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ template "<CHARTNAME>.serviceAccountName" . }}
+  labels:
+  {{ include "<CHARTNAME>.labels" . | indent 4 }}
```
bacongobbler: this should probably use `nindent` instead.

mattfarina: `nindent` adds a new line at the start. But `{{` is used at the start of this to the prior whitespace is kept including a new line. I think `indent` works but to be consistent with the other files created the two spaces before the `{{` need to be removed.

naseemkullah: Thanks!

mattfarina (approved):
For reference, this pattern is common in the charts repo.

naseemkullah:
Thanks @mattfarina , whats the Needs V3 fix label about? Does this feature have to be duplicated elsewhere?

bacongobbler:
this PR isn't relevant for Helm 3 so I'm marking this as "complete".

bacongobbler:
ported to Helm 3 in https://github.com/helm/helm/pull/6490

PR #5898, opened 2019-06-15
Change port forward command order in default note

Aisuko:
Signed-off-by: Aisuko <urakiny@gmail.com>

**What this PR does / why we need it**:

We should take to get `POD_NAME` at first and forward pod's port to the local port, at last, print it to the screen.

**Special notes for your reviewer**:

**If applicable**:
- [x] this PR contains documentation
- [ ] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

PR #5901, opened 2019-06-17
Revert "Change port forward command order in default note"

bacongobbler:
`kubectl port-forward` is a perpetually running command. Once invoked, it waits for user input to exit (CTRL + C), which is why the `echo` was on the earlier line. Now when a user copies those commands and attempts to execute them, the `echo` will not be shown.

Reverts helm/helm#5898

### DIFF
--- a/pkg/chartutil/create.go
+++ b/pkg/chartutil/create.go
@@ -276,8 +276,8 @@ const defaultNotes = `1. Get the application URL by running these commands:
   echo http://$SERVICE_IP:{{ .Values.service.port }}
 {{- else if contains "ClusterIP" .Values.service.type }}
   export POD_NAME=$(kubectl get pods --namespace {{ .Release.Namespace }} -l "app.kubernetes.io/name={{ include "<CHARTNAME>.name" . }},app.kubernetes.io/instance={{ .Release.Name }}" -o jsonpath="{.items[0].metadata.name}")
-  kubectl port-forward $POD_NAME 8080:80
   echo "Visit http://127.0.0.1:8080 to use your application"
+  kubectl port-forward $POD_NAME 8080:80
 {{- end }}
 `
 


hickeyma:
@bacongobbler You make a point to revert #5898. What is the need though for the echo really? Does it matter if it is run or not?

Aisuko:
The situation you said is right is when the chart was deployed into default namespace. In fact, we deploy every chart into the custom namespace,  If you copy those commands and attempt to execute them all, you would get an error. There may be adding `namespace` for port-forward command.  In my opinion, those commands just information or a tip, tell the user to execute them one by one. But now I think mine understand may in a wrong way.

mattfarina:
@Aisuko changing the order will not help with the namespaces. Instead, the `-n` or `--namespace` flag needs to be used on the `kubectl` command.

Once `kubectl port-forward` is called you cannot paste any more commands into that terminal window until port forwarding is ended. This is why other calls need to go first.

mattfarina (approved):

hickeyma:
@mattfarina Is there any need for the `echo`?

mattfarina:
@hickeyma If something isn't listed there how will someone using the chart know where to connect to their app over the port-forward?

`kubectl port-forward` is going to give you a response like:

```
Forwarding from 127.0.0.1:8080 -> 80
Forwarding from [::1]:8080 -> 80
```

That's not a great UX if someone is going to know where to connect curl or their browser up to (as is the case in this example).

hickeyma:
@mattfarina I agree, it is needed. I think because I use it so often I never looked at the echo. I just used it as is.

Sorry for merging PR but I think it makes most sense to reverse. Thanks @bacongobbler.

hickeyma (approved):
LGTM, thanks.

PR #5902, opened 2019-06-17
Add more capabilities to the detault set used by helm template

mattfarina:
- These are the capabilities as of k8s 1.14
- They are generated because getting them at runtime uses so much
  memory it causes CI errors

**If applicable**:


pkg/chartutil/capabilities.go
```diff
@@ -69,3 +69,55 @@ func (v VersionSet) Has(apiVersion string) bool {
 	_, ok := v[apiVersion]
 	return ok
 }
+
+// Default versions as of 1.14, excluding deprecated versions, focused on workloads
+func defaultVersions() []string {
```
bacongobbler: Is there a lib in kubernetes that lists all of the default API groups? That way we wouldn't have this large list of default API versions laying around.

mattfarina: I'll poke around and see

mattfarina: Looks like there is an API for that and it provides 554 items to stick in the default set.

bacongobbler (approved):

bacongobbler:
looks good to merge once CI passes! Thanks for finding `scheme.Scheme.AllKnownTypes()`.

mattfarina:
@bacongobbler I re-wrote it to get around the runtime memory issues. Uses generation now.

pkg/chartutil/generator/capabilities_default_versions_generate.go
```diff
@@ -0,0 +1,102 @@
...
+			check[gv] = struct{}{}
+			s = append(s, gv)
+		}
+		if _, ok := check[gvk]; !ok {
+			check[gvk] = struct{}{}
+			s = append(s, gvk)
+		}
+	}
+
+	return s
```
bacongobbler: small nit, but would it be worth it to sort the string slice before returning here? That way the string slice is consistent between `go generate` invocations.

bacongobbler (approved):
Code LGTM, just one comment about a potential enhancement to the generated code output

pkg/chartutil/generator/capabilities_default_versions_generate.go
mattfarina: Great idea. sort added.

bacongobbler (approved):

hickeyma (approved):
LGTM, thanks @mattfarina. Good to merge.

PR #5903, opened 2019-06-17
Updating to latest ingress API

mattfarina:
Ingress is now part of a group. A goal is to eventually remove the extensions group.


bacongobbler (approved):

bacongobbler:
Will this work on older kubernetes clusters? e.g. 2.13 and 2.12?

mattfarina:
@bacongobbler it was introduced in 1.14. Should we hold off a bit? Maybe just put this in the helm v3 version?

bacongobbler:
Yeah, we should probably hold off. In general we don't try to support features that won't work on older supported versions of Kubernetes. If this works on just 2.14 clusters, then perhaps we should hold off a bit and port this over to Helm 3.

mattfarina:
For some reason I'd thought this had been around longer.

Issue #5907, opened 2019-06-18
apiVersion still v1 in Helm v3


mattfarina:
The `apiVersion` in charts for Helm v3 is `v1`. Yet, there have been some chart changes including:

- dependencies moved to the `Chart.yaml` file
- `requirements.lock` is now `Chart.lock`
- annotations added to the `Chart.yaml` file

I understand that Helm v3 works for both the older and newer setups. But, external tools that work with charts don't know about the changes.

Should we move the `apiVersion` to `v2`?

bacongobbler:
slating to get this done before we release the first beta.

hickeyma:
@mattfarina @bacongobbler I can work on this if it is open to be done?

bacongobbler:
Sure!

hickeyma:
Here are the tasks required:

- [ ] Update scaffold chart to be "v2"
- [ ] Update linting to handle new dependencies and annotations
- [ ] Update documentation to specify "v2" as default version and "v1" for specific scenarios

bacongobbler:
Additional TODO:

- [ ] Helm 2 should fail to load a chart with an `apiVersion` field of anything other than `v1` or the empty string (Helm 2 doesn't validate the `apiVersion` field, so those charts are assumed to be compiled against `apiVersion` v1

hickeyma:
Thanks @bacongobbler for the heads up in https://github.com/helm/helm/issues/5907#issuecomment-516963275. I have added it to the task list in https://github.com/helm/helm/issues/5907#issuecomment-516063552.

bacongobbler:
all of the PRs for the tasks have been merged so this can now be closed :)

hickeyma:
Thanks @bacongobbler for the doc PR.

PR #5911, opened 2019-06-18
helm test to support new test annotation

michelleN:
This is the first PR to implement the helm test [proposal](https://github.com/helm/community/pull/88/files) for v3.
This PR makes the following changes:
- ~tests are now fetched from `templates/tests` directory in a chart~
- tests are now identified with the following annotation: "helm.sh/test-expect-success"
- tests are no longer implemented as hooks
- removes current test hooks and add a deprecation notice. Manifests with test hooks will simply not be parsed and saved on for a release. 

- Test hooks should be removed entirely by the first beta release.
- A follow up PR/commit will implement `helm test cleanup` which will add the labels and annotations necessary to figure out which pods to clean up as indicated in the [proposal](https://github.com/helm/community/pull/88/files).
- The final PR for helm test will include updates to documentation and cleanup/refactor. Trying not to conflate refactor with change in functionality which is why this is the last bit.
- @bacongobbler suggested that we should update the proposal with why we're going from helm 2 test system to helm 3 test system so I will do that as well (possible blog post)

### DIFF
--- a/docs/examples/nginx/templates/tests/service-test.yaml
+++ b/docs/examples/nginx/templates/tests/service-test.yaml
@@ -8,7 +8,7 @@ metadata:
     helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version }}
     app.kubernetes.io/name: {{ template "nginx.name" . }}
   annotations:
-    "helm.sh/hook": test-success
+    "helm.sh/test-expect-success": "true"
 spec:
   containers:
     - name: curl
--- a/pkg/releaseutil/manifest_sorter.go
+++ b/pkg/releaseutil/manifest_sorter.go
@@ -48,6 +48,7 @@ type manifestFile struct {
 type result struct {
 	hooks   []*release.Hook
 	generic []Manifest
+	tests   []*release.Test
 }
 
 // TODO: Refactor this out. It's here because naming conventions were not followed through.
@@ -66,7 +67,8 @@ var events = map[string]release.HookEvent{
 }
 
 // SortManifests takes a map of filename/YAML contents, splits the file
-// by manifest entries, and sorts the entries into hook types.
+// by manifest entries, skips partials, skips empty manifests, sorts test
+// and hook entries.
 //
 // The resulting hooks struct will be populated with all of the generated hooks.
 // Any file that does not declare one of the hook types will be placed in the
@@ -74,16 +76,16 @@ var events = map[string]release.HookEvent{
 //
 // Files that do not parse into the expected format are simply placed into a map and
 // returned.
-func SortManifests(files map[string]string, apis chartutil.VersionSet, sort KindSortOrder) ([]*release.Hook, []Manifest, error) {
+func SortManifests(files map[string]string, apis chartutil.VersionSet, sort KindSortOrder) ([]*release.Hook, []Manifest, []*release.Test, error) {
 	result := &result{}
 
-	for filePath, c := range files {
-
+	for filepath, c := range files {
 		// Skip partials. We could return these as a separate map, but there doesn't
 		// seem to be any need for that at this time.
-		if strings.HasPrefix(path.Base(filePath), "_") {
+		if strings.HasPrefix(path.Base(filepath), "_") {
 			continue
 		}
+
 		// Skip empty files and log this.
 		if strings.TrimSpace(c) == "" {
 			continue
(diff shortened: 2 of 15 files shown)


mattfarina:
Two things catch my attention...

> tests are no longer implemented as hooks

and 

> A follow up PR/commit will remove current test hooks and add a deprecation notice. Manifests with test hooks will simply not be parsed and saved on for a release. Test hooks should be removed entirely. I'll suggest they be removed by the first beta release.

We have been striving for backwards compatibility with v1 charts. It sounds like this breaks that. Should we wait until Helm v4 to remove the old tests while having a deprecation notice?

michelleN:
That's a good question. @adamreese and I discussed this a bit too.

Your resource manifests for the release would still work as expected. The chart is backwards compatible in that manner. I think we should take this opportunity to make a breaking change like this rather than pushing off to v4. `helm test` was a more of something we wanted to try out and experiment with in a backwards compatible way. That's why it was implemented as a hook. 

Now that we've gotten feedback from the community, I'd like to take this opportunity to move test out into its own abstraction and create a more robust and solid foundation for the `test` feature. We can technically support both the old hook annotation, the new annotation, and still make tests its own thing. It will just be _really_ ugly code. I'm afraid also we'll have some unintended consequences where there is hook logic. If the community feels, we should support the old hook annotation then I can look into that some more.

Note that the migration for tests from helm v2 -> v3 is rather simple though. There are two steps:
1. Move your test manifests to the `templates/tests/` dir. 
2. Change the annotations to `helm.sh/expect-test-success: <true/false>`

Perhaps @jlegrone @atheiman could chime in since they have been helpful in vocalizing needs/requests around `helm test`.

mattfarina:
Several questions come to mind:

* `helm test` isn't documented as an experiment. I don't think end users took it that way. The ones I've spoken with didn't. I know the charts maintainers didn't. I think there is an expectations mismatch. How can we handle communications and experience around that?
* Helm v3 still calls the `apiVersion` for charts `v1`. This breaks semver for that more than anything else. Time to increment the chart version? Should Helm v2 check for charts v2 and note they can't run them?
* `templates/tests` is where many existing tests are at. For example, see the tests in the charts repo. Will those just be skipped if they have the old annotation? What if they have the old and new annotations?

It's worth noting that some use the current test system today. It's used 34 times in the charts repo. Our documents didn't specify that it

michelleN:
> How can we handle communications and experience around that?
At the time we were first developing `helm test`, I asked the question of how we could communicate that very fact. I suggested we have features in `alpha`. That conversation didn't go anywhere and it happened ~2 years ago. I suggest with the next new feature that we figure out how to communicate a feature is an experiment. I suggest opening an issue for this.

> Time to increment the chart version? 
We've had conversations around incrementing the chart version and I think it's time to do that. Would be good if you could bring this up in the call today @mattfarina 

>  Should Helm v2 check for charts v2 and note they can't run them?
That makes sense to me. I would suggest opening an issue for this.

> Will those just be skipped if they have the old annotation?
Yes, that's correct

>What if they have the old and new annotations?
Old annotation will be ignored and only the new annotation will be parsed

mattfarina:
>At the time we were first developing helm test, I asked the question of how we could communicate that very fact. I suggested we have features in alpha. That conversation didn't go anywhere and it happened ~2 years ago. I suggest with the next new feature that we figure out how to communicate a feature is an experiment. I suggest opening an issue for this.

Yeah, in the future we need to communicate alpha or experimental features better. Should we document a way to do that?

I was more thinking about the current `helm test`. I wasn't communicated as being alpha or experimental and people built on it. So, somewhere we need to do a good job documenting the change and this situation for all those folks.

cc @helm/charts-maintainers 

> We've had conversations around incrementing the chart version and I think it's time to do that. Would be good if you could bring this up in the call today @mattfarina

I missed this until after the call and I won't be able to make the next one. There's an open issue on this via #5907. I'll add it to the agenda for the next call even if I'm not there.

> > Should Helm v2 check for charts v2 and note they can't run them?
>
> That makes sense to me. I would suggest opening an issue for this.

Done. #5924

hickeyma:
Thanks @michelleN  for the PR. Have one concern before I review this. It relates to breaking the original test UX as mentioned above. There are potentially a lot of users out there who are using this capability in 1..m charts, used in production. They depend on it for validity of their charts. How do we handle this scenario? If I understand correctly, the idea is to make the current test framework disable silently and a deprecation message which happens immediately? If so, not sure that this is practical approach for users.

docs/examples/nginx/templates/tests/service-test.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
     helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version }}
     app.kubernetes.io/name: {{ template "nginx.name" . }}
   annotations:
-    "helm.sh/hook": test-success
+    "helm.sh/test-expect-success": "true"
```
jlegrone: If we are to make a breaking change, I'd like to eliminate failure test cases. In practice they are hardly used -- on github it appears that there are about [20 usages of test-failure](https://github.com/search?l=YAML&q=helm.sh%2Fhook+test-failure&type=Code) and [8000 usages of test-success](https://github.com/search?l=YAML&q=helm.sh%2Fhook+test-success&type=Code) (0 and 34 respectively in helm/charts). Where `test-failure` _is_ used it can be replaced by a small amount of logic in the test itself (and this is a better fit since there are a variety of ways a pod can fail aside from what runs in its entrypoint).

pkg/releaseutil/manifest_sorter.go
```diff
@@ -142,6 +152,25 @@ func (file *manifestFile) sort(result *result) error {
 			continue
 		}
 
+		expectation, ok := entry.Metadata.Annotations["helm.sh/test-expect-success"]
```
jlegrone: Is there currently a requirement that tests be defined in the `templates/tests` directory? (that seems to be indicated by the PR/commit comments but I'm missing where it's done in the code).

I think I'd rather only use the annotation. See also https://github.com/helm/community/pull/40/files#r297251345

michelleN:
closing in favor of #6054 and @jlegrone will take the test work from here

PR #5917, opened 2019-06-19
ref(getter): introduce Options for passing in getter parameters

bacongobbler:
instead of hard-coding the parameters being passed in the constructor, we should pass in an Options struct that can be used to pass in those parameters.

Additionally, this PR allows one to set the username, password and user agent used for the HTTPGetter when being initialized. Previously, a call had to be made to SetBasicAuth and SetUserAgent.

### DIFF
--- a/pkg/getter/getter.go
+++ b/pkg/getter/getter.go
@@ -24,6 +24,55 @@ import (
 	"helm.sh/helm/pkg/cli"
 )
 
+// options are generic parameters to be provided to the getter during instantiation.
+//
+// Getters may or may not ignore these parameters as they are passed in.
+type options struct {
+	url       string
+	certFile  string
+	keyFile   string
+	caFile    string
+	username  string
+	password  string
+	userAgent string
+}
+
+// Option allows specifying various settings configurable by the user for overriding the defaults
+// used when performing Get operations with the Getter.
+type Option func(*options)
+
+// WithURL informs the getter the server name that will be used when fetching objects. Used in conjunction with
+// WithTLSClientConfig to set the TLSClientConfig's server name.
+func WithURL(url string) Option {
+	return func(opts *options) {
+		opts.url = url
+	}
+}
+
+// WithBasicAuth sets the request's Authorization header to use the provided credentials
+func WithBasicAuth(username, password string) Option {
+	return func(opts *options) {
+		opts.username = username
+		opts.password = password
+	}
+}
+
+// WithUserAgent sets the request's User-Agent header to use the provided agent name.
+func WithUserAgent(userAgent string) Option {
+	return func(opts *options) {
+		opts.userAgent = userAgent
+	}
+}
+
+// WithTLSClientConfig sets the client client auth with the provided credentials.
+func WithTLSClientConfig(certFile, keyFile, caFile string) Option {
+	return func(opts *options) {
+		opts.certFile = certFile
+		opts.keyFile = keyFile
+		opts.caFile = caFile
+	}
+}
+
 // Getter is an interface to support GET to the specified URL.
 type Getter interface {
 	//Get file content by url string
@@ -32,7 +81,7 @@ type Getter interface {
 
 // Constructor is the function for every getter which creates a specific instance
 // according to the configuration
-type Constructor func(URL, CertFile, KeyFile, CAFile string) (Getter, error)
+type Constructor func(options ...Option) (Getter, error)
 
 // Provider represents any getter and the schemes that it supports.
 //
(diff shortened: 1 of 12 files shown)


mattfarina (approved):

pkg/getter/httpgetter_test.go
```diff
@@ -37,12 +41,47 @@ func TestHTTPGetter(t *testing.T) {
...
+
+	transport, ok := hg.client.Transport.(*http.Transport)
+	if !ok {
+		t.Errorf("Expected newHTTPGetter to set up an HTTP transport")
+	}
+
+	test.AssertGoldenString(t, transport.TLSClientConfig.ServerName, "output/httpgetter-servername.txt")
+
+	// Test other options
+	hg, err = NewHTTPGetter(Options{
```
hickeyma: Can you use different values for the uname, passwd and ua to be sure they are set correctly and therefore the check below can prove it?

hickeyma (changes requested):
Thanks for the PR @bacongobbler. Looking good. Just a small change in the test.

pkg/getter/getter.go
```diff
@@ -24,6 +24,19 @@ import (
 	"helm.sh/helm/pkg/cli"
 )
 
+// Options are generic parameters to be provided to the getter during instantiation.
+//
+// Getters may or may not ignore these parameters as they are passed in.
+type Options struct {
```
thomastaylor312: Is there a reason why we aren't using functional options instead for this? I always worry about using "option" or "config" structs because they can balloon fairly quickly and become unwieldy

pkg/getter/httpgetter.go
```diff
@@ -82,21 +82,21 @@ func (g *HTTPGetter) get(href string) (*bytes.Buffer, error) {
 }
 
 // newHTTPGetter constructs a valid http/https client as Getter
-func newHTTPGetter(url, certFile, keyFile, caFile string) (Getter, error) {
-	return NewHTTPGetter(url, certFile, keyFile, caFile)
+func newHTTPGetter(options Options) (Getter, error) {
```
thomastaylor312: This may be out of scope for this PR, but is there a reason the exported `NewHTTPGetter` doesn't just return a `Getter`? Based on what I am seeing in these files, we no longer need to export the `HTTPGetter` and just hide it behind the `Getter` interface. The `HTTPGetter` specific functions are now covered by the `Options` config

pkg/getter/plugingetter.go
```diff
@@ -82,12 +82,12 @@ func (p *pluginGetter) Get(href string) (*bytes.Buffer, error) {
 
 // newPluginGetter constructs a valid plugin getter
 func newPluginGetter(command string, settings cli.EnvSettings, name, base string) Constructor {
```
thomastaylor312: Like my other comment, this may be out of scope, but this seems a little funky. I think if we have a `NewHTTPGetter`, we should also have an exported `NewPluginGetter` instead of this one-off `Constructor` bit that doesn't seem to provide much abstraction value

thomastaylor312 (commented):
Nothing that would necessarily block from merging this PR, but some interesting cruft and discussion from looking at this code I haven't seen in a while

pkg/getter/httpgetter.go
bacongobbler: Some parts of the codebase still rely on NewHTTPGetter returning an HTTPGetter so it can call SetBasicAuth (and SetUserAgent, though that isn't used in the codebase). We can certainly do that in a follow-up now that these options are being passed in and set when the getter is being instantiated.

pkg/getter/httpgetter_test.go
bacongobbler: sure.

pkg/getter/plugingetter.go
bacongobbler: Agreed. Can we do that in a follow-up PR? #5922 would be a good opportunity to do that since it makes changes to the chart downloader's behaviour and how it instantiates the getter.

pkg/getter/getter.go
bacongobbler: Do you have an example? Open to opinions here (or... options... as it were :wink:)

bacongobbler: Ah nvm I found an article. Haven't heard that term before. I believe we did this before with the Helm gRPC client as well. Will take a look at that approach.

bacongobbler: okay, implemented. Let me know what you think.

pkg/getter/plugingetter.go
thomastaylor312: SGTM

pkg/getter/getter.go
thomastaylor312: I am still a little bit unsure of having a backing `options` struct, but I can't think of an easy way to make it work for both types of getters without having duplicate options for each one. It is abstracted away from the user though because only the options funcs are exported, so this should work just fine. We can always change the backing later because it is unexported and we won't break backwards compatibility

thomastaylor312 (approved):
Looking great!

PR #5918, opened 2019-06-19
ref(downloader): remove username/password fields from ChartDownloader

bacongobbler:
Looks like this was dead code from the beginning, as the username and password is read during `getRepoCredentials`. HTTPGetter code was being leaked into other structs that should not have been passed along. Instead, the caller should've been calling the HTTPGetter's `SetBasicAuth` function to pass these parameters in. #5917 provides a way to pass these parameters from the CLI when the getters are being initialized.

The tests passed, so... :man_shrugging: 

### DIFF
--- a/pkg/downloader/chart_downloader.go
+++ b/pkg/downloader/chart_downloader.go
@@ -68,10 +68,6 @@ type ChartDownloader struct {
 	HelmHome helmpath.Home
 	// Getter collection for the operation
 	Getters getter.Providers
-	// Chart repository username
-	Username string
-	// Chart repository password
-	Password string
 }
 
 // DownloadTo retrieves a chart. Depending on the settings, it may also download a provenance file.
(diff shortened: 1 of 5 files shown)


pkg/downloader/chart_downloader.go
```diff
@@ -255,15 +251,9 @@ func (c *ChartDownloader) ResolveChartVersion(ref, version string) (*url.URL, ge
...
-	password = c.Password
 	if r != nil && r.Config != nil {
-		if username == "" {
-			username = r.Config.Username
-		}
-		if password == "" {
-			password = r.Config.Password
-		}
+		username = r.Config.Username
+		password = r.Config.Password
```
mattfarina: The calling code for `getRepoCredentials` creates the passed in repository object with a username or password. See https://github.com/helm/helm/blob/c35dbb7aabf88f7e0641c574b073e7ab904281cd/pkg/downloader/chart_downloader.go#L176-L178. 

Calls to `ResolveChartVersion` will no longer be able to pass in creds with this change. The username and password are set in install.go to use in here. It looks like this removes the ability to pass a username and password.

If you want to trace the use of the removed code you might start at the `addChartPathOptionsFlags` function in the install cmd code.

bacongobbler: Thanks for calling this out. Looks like this needs to be covered by a unit test. I'll see if I can add one to cover this case.

bacongobbler: Ah, I remember now what I was going to do with this...

With the introduction of #5917, it's possible to introduce an `Options` field to the `ChartDownloader` which can be passed along to the Getter being initialized. I'm working on a PR right now to implement just that. The end result is that with #5917, #5918 and that PR merged, we should retain the same functionality as before, but in a more maintainable fashion.

It'd stillbe worthwhile to capture this in a test case to ensure that parameters passed in via `--username` and `--password` are passed along to the Getter.

bacongobbler: See #5922 for the fix

bacongobbler:
closing in favour of #5922, as that PR removes the Username/Password fields from the ChartDownloader in the fashion that I originally wanted to go about.

PR #5919, opened 2019-06-20
[helm create] Add security context to deployment template

### LINKED ISSUE #5880: helm create - Make it easier to do the right thing, part ii: securityPolicy
Hello,

Helm create should add options to create security policies, as it is a known best practice to enable security policies, notably, to run as non root, to have a read only file system , drop linux capabilities and so on.
By adding the option to create security policies we are making it easier to do the right thing.

cc @bacongobbler @unguiculus @scottrigby @rimusz

naseemkullah:
closes #5880 

**If applicable**:

### DIFF
--- a/pkg/chartutil/create.go
+++ b/pkg/chartutil/create.go
@@ -76,6 +76,17 @@ serviceAccount:
   # If not set and create is true, a name is generated using the fullname template
   name:
 
+podSecurityContext: {}
+  # fsGroup: 2000
+
+securityContext: {}
+  # capabilities:
+  #   drop:
+  #   - ALL
+  # readOnlyRootFilesystem: true
+  # runAsNonRoot: true
+  # runAsUser: 1000
+
 service:
   type: ClusterIP
   port: 80
@@ -199,8 +210,12 @@ spec:
         {{- toYaml . | nindent 8 }}
     {{- end }}
       serviceAccountName: {{ template "<CHARTNAME>.serviceAccountName" . }}
+      securityContext:
+        {{- toYaml .Values.podSecurityContext | nindent 8 }}
       containers:
         - name: {{ .Chart.Name }}
+          securityContext:
+            {{- toYaml .Values.securityContext | nindent 12 }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           ports:


naseemkullah:
cc @mattfarina @bacongobbler

naseemkullah:
I could not find a reused pattern in stable repo charts, i'm suggesting this one...

pkg/chartutil/create.go
```diff
@@ -76,6 +76,17 @@ serviceAccount:
   # If not set and create is true, a name is generated using the fullname template
   name:
 
+podSecurityContext:
+  fsGroup: 2000
+
+securityContext:
```
unguiculus: Everything below should be indented.

pkg/chartutil/create.go
```diff
@@ -199,8 +210,12 @@ spec:
         {{- toYaml . | nindent 8 }}
     {{- end }}
       serviceAccountName: {{ template "<CHARTNAME>.serviceAccountName" . }}
+      securityContext:
+        {{- toYaml .Values.podSecurityContext | indent 8 }}
```
unguiculus:
```suggestion
        {{- toYaml .Values.podSecurityContext | nindent 8 }}
```

unguiculus (commented):
Thanks for the PR. I'm not sure we should do this, though. `helm create` gives you an initial setup to start from. We can't add all the things one could possibly want. Also, what defaults to set here? The initial chart uses Nginx, which probably won't work any more out of the box with these defaults. See https://hub.docker.com/_/nginx/.

pkg/chartutil/create.go
naseemkullah: Thanks

naseemkullah:
> Thanks for the PR. I'm not sure we should do this, though. `helm create` gives you an initial setup to start from. We can't add all the things one could possibly want. Also, what defaults to set here? The initial chart uses Nginx, which probably won't work any more out of the box with these defaults. See https://hub.docker.com/_/nginx/.

I just want it to be easy to do the RIGHT thing, and that is to not run as root, and to set securityContext. I think truly having to run as root is a very rare thing, now we should steer the community from doing that if it can be avoided. 

I do agree we can't make the `helm create` initial setup do all the things we want, but jus by adding this, so many future charts will be more secure because it will be easy to add security.

naseemkullah:
At the very least, we can ship this, but defaults will be `{}` ... with possible values commented. Just like we do with `resources` ... On second thought this might be better. Let me comment out default values.

naseemkullah:
But by having that there, it makes it much easier for someone getting started to add security easily, and having it in the values.yaml is a good reminder.

paulczar:
I like the idea of having the boilerplate for these, but agree with @unguiculus that defaults could be dangerous for the "I just want the thing to work" when first setting up a new chart.  I think settings default value to be the empty `{}` and having some *good* defaults as comments below makes sense.

naseemkullah:
Thanks, I've applied the change to leave them empty by default.
Suggested commented runAsUser is `1000`, but kubesec.io suggest something `> 10000` , any preference for default value?

Also not sure why CI is failing?
```
FAIL	k8s.io/helm/cmd/tiller [build failed]
FAIL	k8s.io/helm/pkg/tiller/environment [build failed]
FAIL	k8s.io/helm/pkg/releasetesting [build failed]
```

hickeyma:
@naseemkullah Nothing obvious causing the build issue. Maybe rebase and commit again to do a rebuild?

naseemkullah:
Thanks @hickeyma, that worked.

unguiculus (approved):

hickeyma (approved):
LGTM, thanks.

naseemkullah:
My pleasure.

bacongobbler:
this PR isn't relevant for Helm 3 so I'm marking this as "complete".

bacongobbler:
ported to Helm 3 in https://github.com/helm/helm/pull/6490

PR #5922, opened 2019-06-20
 ref(downloader): pass in options to ChartDownloader

bacongobbler:
Built on top of #5917 and #5918.

This restores the ability to pass in parameters at runtime to the ChartDownloader, enabling users to pass in parameters like the `--username` and `--password` flags. This also adds a test case to cover that scenario, as it was previously untested.

I also had to introduce a `WithMiddleware` function to the test server. That way the tests could inject HTTP basic auth headers into the server to test this PR.

cc'ing @mattfarina, as he raised this concern in https://github.com/helm/helm/pull/5918#discussion_r295825504 which this PR addresses (and adds a test case to help confirm).

### DIFF
--- a/pkg/downloader/chart_downloader.go
+++ b/pkg/downloader/chart_downloader.go
@@ -68,10 +68,8 @@ type ChartDownloader struct {
 	HelmHome helmpath.Home
 	// Getter collection for the operation
 	Getters getter.Providers
-	// Chart repository username
-	Username string
-	// Chart repository password
-	Password string
+	// Options provide parameters to be passed along to the Getter being initialized.
+	Options []getter.Option
 }
 
 // DownloadTo retrieves a chart. Depending on the settings, it may also download a provenance file.
@@ -86,7 +84,17 @@ type ChartDownloader struct {
 // Returns a string path to the location where the file was downloaded and a verification
 // (if provenance was verified), or an error if something bad happened.
 func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *provenance.Verification, error) {
-	u, g, err := c.ResolveChartVersion(ref, version)
+	u, err := c.ResolveChartVersion(ref, version)
+	if err != nil {
+		return "", nil, err
+	}
+
+	constructor, err := c.Getters.ByScheme(u.Scheme)
+	if err != nil {
+		return "", nil, err
+	}
+
+	g, err := constructor(c.Options...)
 	if err != nil {
 		return "", nil, err
 	}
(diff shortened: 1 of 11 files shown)


pkg/downloader/chart_downloader.go
```diff
@@ -173,90 +176,100 @@ func (c *ChartDownloader) ResolveChartVersion(ref, version string) (*url.URL, ge
...
+			return repoURL, err
 		}
 		q := repoURL.Query()
 		// We need a trailing slash for ResolveReference to work, but make sure there isn't already one
 		repoURL.Path = strings.TrimSuffix(repoURL.Path, "/") + "/"
 		u = repoURL.ResolveReference(u)
 		u.RawQuery = q.Encode()
 		// TODO add user-agent
-		g, err := getter.NewHTTPGetter(rc.URL, "", "", "")
+		g, err := getter.NewHTTPGetter(getter.Options{
```
thomastaylor312: Do we need `g` here or the `SetBasicAuth` on line 259? This isn't returning the getter anymore, so it seems like this is just for an error check

pkg/repo/repotest/server.go
```diff
@@ -79,8 +79,15 @@ func NewServer(docroot string) *Server {
...
-	srv     *httptest.Server
+	docroot    string
+	srv        *httptest.Server
+	middleware http.HandlerFunc
+}
+
+// WithMiddleware injects middleware in front of the server. This can be used to inject
+// additional functionality like layering in an authentication frontend.
+func (s *Server) WithMiddleware(middleware http.HandlerFunc) {
+	s.middleware = middleware
```
thomastaylor312: If we are enabling middleware, should we handle chaining multiple middlewares or leave it up to the user to pass a `HandlerFunc` that already has all the middleware composed?

thomastaylor312 (commented):
A few small questions/discussion items. I'll come back and give a green checkmark once the dependent PR is merged

pkg/repo/repotest/server.go
bacongobbler: I can look further into this. I started with the idea of allowing a way to append middleware to the chain but landed on this design as a first pass.

pkg/downloader/chart_downloader.go
bacongobbler: Good catch. No, it's not needed any more. I'll remove those two.

pkg/repo/repotest/server.go
thomastaylor312: Because it is a test server, I think it is ok. The question just came up as I was looking at the code

pkg/getter/httpgetter.go
```diff
@@ -79,13 +68,8 @@ func (g *HTTPGetter) get(href string) (*bytes.Buffer, error) {
 	return buf, err
 }
 
-// newHTTPGetter constructs a valid http/https client as Getter
-func newHTTPGetter(options ...Option) (Getter, error) {
-	return NewHTTPGetter(options...)
-}
-
 // NewHTTPGetter constructs a valid http/https client as HTTPGetter
```
thomastaylor312: This comment should be updated to reflect that it just returns a `Getter` now

thomastaylor312 (commented):
This is looking awesome! Just one exported function comment change and a rebase on `dev-v3` and it should be good to go

bacongobbler:
done

thomastaylor312 (approved):

PR #5928, opened 2019-06-23
modify values-test for big-int value

jan25:
Helm render uses `yaml.Unmarshal` to read yaml files, and then we manually traverse the yaml structure to convert to `chartutil.Values` which is of type `map[string]interface{}`. So any float64 values in yaml will be stringified as is without looking at their types

This PR tries to break `TestCoalesceValues` to reproduce a issue 

I made this PR just to reference in the issue https://github.com/helm/helm/issues/5929. Will close this immediately

**If applicable**:

### DIFF
--- a/pkg/chartutil/values_test.go
+++ b/pkg/chartutil/values_test.go
@@ -285,6 +285,7 @@ right: Null
 left: NULL
 front: ~
 back: ""
+tag: 20190612073634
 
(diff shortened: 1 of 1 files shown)


jan25:
This was from CI before closing the issue
Screen Shot 2019-06-23 at 10 50 23

Issue #5941, opened 2019-06-25
Helm init timeout


ds-ms:
When trying to run `helm init` I get the following error: 

```
$HELM_HOME has been configured at /home/<user>/.helm. 
Warning: Tiller is already installed in the cluster. 
(Use --client-only to suppress this message, or --upgrade to upgrade Tiller to the current version.) 
Error: tiller was not found. polling deadline exceeded 
```

Although, it succeeded afterward when I tried `helm init --force-upgrade`.

It is kind of ironical, the warning says that `Tiller is already installed` but the error says `Tiller was not found`.

I was going through the code and noticed that the error which was thrown by the `installer.Install()` call was getting [masked](https://github.com/helm/helm/blob/master/cmd/helm/init.go#L287). 

Perhaps, this error could be printed in the debug logs, or when `--debug` is enabled?

ds-ms:
I raised a PR to print some extra `[INFO]` logs when `--debug` is enabled.

PR #5942, opened 2019-06-25
Adding extra debug logs

thesattiraju:
**What this PR does / why we need it**: A possible solution for #5941

### DIFF
--- a/cmd/helm/init.go
+++ b/cmd/helm/init.go
@@ -284,6 +284,7 @@ func (i *initCmd) run() error {
 				}
 				fmt.Fprintln(i.out, "\nTiller (the Helm server-side component) has been upgraded to the current version.")
 			} else {
+				debug("The error received while trying to init: %s", err)
 				fmt.Fprintln(i.out, "Warning: Tiller is already installed in the cluster.\n"+
 					"(Use --client-only to suppress this message, or --upgrade to upgrade Tiller to the current version.)")
 			}


bansalaseem (approved):

cmd/helm/init.go
```diff
@@ -284,6 +284,9 @@ func (i *initCmd) run() error {
 				}
 				fmt.Fprintln(i.out, "\nTiller (the Helm server-side component) has been upgraded to the current version.")
 			} else {
+				if settings.Debug {
+					fmt.Fprintln(i.out, fmt.Sprintf("[INFO] The error received while trying to init: %s", err))
+				}
```
mattfarina: I like the desire to display the debug information. But, can you use the `debug` function to do this instead? It's part of the `main` package here in `printer.go`. It has the if statement wrapper, etc.

mattfarina (changes requested):

cmd/helm/init.go
ds-ms: @mattfarina Updated the logic to use `debug` function instead

cmd/helm/init.go
```diff
@@ -284,6 +284,7 @@ func (i *initCmd) run() error {
 				}
 				fmt.Fprintln(i.out, "\nTiller (the Helm server-side component) has been upgraded to the current version.")
 			} else {
+				debug("[INFO] The error received while trying to init: %s", err)
```
hickeyma: Remove the '[INFO]' label as debug function uses '[debug]' label.

hickeyma (changes requested):

cmd/helm/init.go
ds-ms: Updated

hickeyma (approved):
LGTM, thanks.

mattfarina (approved):

thomastaylor312:
Init is being removed in v3, so marking as port complete

PR #5947, opened 2019-06-26
Add `helm prune` command

### LINKED ISSUE #5944: `helm prune`: purge old deleted releases
It might be nice to have a `helm prune` command to `purge` all your old releases that you've deleted and no longer need. It would be similar to the `docker prune` command, which removes old stopped containers and other resources.

It would effectively do this:
```
helm ls --deleted --short | xargs -L1 helm delete --purge
```

alexkreidler:
Adds a `helm prune` command to the CLI, along with a `PruneReleases` function in the library to purge old deleted releases. See #5944 for more background about the use-case/workflow.

Currently, I have tested the command on my own cluster, and it works great!

Closes #5944


mattfarina:
@alexkreidler Can you run "make docs" to regenerate docs from the changes? That was the reason for the CI failure.

alexkreidler:
@mattfarina Just added docs. Thanks for filling me in!

bridgetkromhout:
Hi, @alexkreidler! Thanks for showing this proof of concept! As @bacongobbler mentioned in https://github.com/helm/helm/issues/5944, we'd like to see this as a plugin. (Bonus: that means it will work for Helm 2 and Helm 3!) We'll close this PR, and let's keep working through it over on the issue. Thanks.

PR #5961, opened 2019-07-01
fix(pkg/kube): Fixes wait functionality

thomastaylor312:
This reenables wait functionality and fixes some small bugs in the logic. Please
note that there are still some naive assumptions made about pods belonging to
DaemonSets and StatefulSets, but that is how the logic was before, so it was not
in scope to modify it for this PR. I will improve this logic in a follow up PR

### DIFF
--- a/pkg/kube/wait.go
+++ b/pkg/kube/wait.go
@@ -47,9 +47,12 @@ func (w *waiter) waitForResources(created Result) error {
 	w.log("beginning wait for %d resources with timeout of %v", len(created), w.timeout)
 
 	return wait.Poll(2*time.Second, w.timeout, func() (bool, error) {
-		for _, v := range created[:0] {
+		for _, v := range created {
 			var (
-				ok  bool
+				// This defaults to true, otherwise we get to a point where
+				// things will always return false unless one of the objects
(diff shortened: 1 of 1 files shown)


pkg/kube/wait.go
```diff
@@ -146,6 +153,9 @@ func (w *waiter) waitForResources(created Result) error {
 				ok, err = w.podsReadyForObject(value.Namespace, value)
 			case *appsv1.ReplicaSet:
 				ok, err = w.podsReadyForObject(value.Namespace, value)
+			default:
+				// If the given type is not one we check for, continue in the loop
```
bacongobbler: is this needed? If I'm reading lines 55 and 160 correctly, `ok` should be `true` and `err` should be `nil` if we reach here, meaning that it'll continue iterating through the loop.

bacongobbler (approved):
LGTM, just one comment about the design but the approach makes sense! Thanks for fixing this.

pkg/kube/wait.go
thomastaylor312: I think you are right. I fixed this first and then moved to the next bug. Let me remove it and double check it works

PR #5967, opened 2019-07-03
Documentation upgrade

nehagup:
According to documentation helm install [CHART] [flags] should be the usage pattern. Seems a new update.


nehagup:
I'm not sure why this is failing.

bacongobbler (approved):

PR #5970, opened 2019-07-03
fix(pkg/action): Adds back in missing wait functionality

thomastaylor312:
In my previous PR, I did not notice that the wait functionality had been completely removed
from the actions. This restores wait functionality to upgrade and rollback

### DIFF
--- a/pkg/action/rollback.go
+++ b/pkg/action/rollback.go
@@ -149,7 +149,7 @@ func (r *Rollback) performRollback(currentRelease, targetRelease *release.Releas
 
 	cr := bytes.NewBufferString(currentRelease.Manifest)
 	tr := bytes.NewBufferString(targetRelease.Manifest)
-	// TODO add wait
+
 	if err := r.cfg.KubeClient.Update(cr, tr, r.Force, r.Recreate); err != nil {
 		msg := fmt.Sprintf("Rollback %q failed: %s", targetRelease.Name, err)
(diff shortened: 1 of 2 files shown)


bacongobbler (commented):
Is there a way we can add tests to this? Perhaps with the [PrintingKubeClient](https://github.com/helm/helm/blob/e55a25bfebdf1d8758c2107e0af2859f525b3f23/pkg/kube/printer.go#L27-L29) and perhaps a `FlakyKubeClient` that reports back errors? That way we can at least capture this in tests.

thomastaylor312:
I think for this, the amount of code added + provided benefit doesn't add up to a lot. @jdolitsky is working on stuff to test the wait functionality that I think can eventually capture this. I can add it though if really desired

jdolitsky:
@thomastaylor312 @bacongobbler Yep, I'm halfway done with an acceptance test suite w/ robot that tests Helm's general functionality across Kube API versions (with the --wait being the first test case).

Spins up a test cluster for each supported version with [`kind`](https://kind.sigs.k8s.io/), then runs the tests
Screen Shot 2019-07-03 at 2 52 44 PM

Unit tests are always good, but I think this level of testing gives us better confidence things will work in the wild.

Let me know if there are any other things we can add here that have historically been tested by hand

bacongobbler (approved):

PR #6008, opened 2019-07-10
Add constructor for setting ValueOptions.rawValues from map

rudoi:

This PR adds a constructor function that allows a ValueOptions object to be created from a map.

**If applicable**:

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -693,6 +693,12 @@ func (v *ValueOptions) MergeValues(settings cli.EnvSettings) error {
 	return nil
 }
 
+func NewValueOptions(values map[string]interface{}) ValueOptions {
+	return ValueOptions{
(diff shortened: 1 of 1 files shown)


thomastaylor312 (approved):
LGTM

hickeyma (commented):
@rudio what is the use case for the function?

rudoi:
@hickeyma I believe this provides a way to build ValueOptions that is more friendly for those consuming Helm packages as libraries. I would like to deploy charts from within a CRD controller. Right now, the interfaces provided allow me to pass in a slice of file paths or a slice of key=value strings.

It would be preferable if I could supply an unstructured map[string]interface{} and build a ValueOptions from that.

hickeyma (approved):
@rudoi LGTM, thanks.

Issue #1707, opened 2016-12-17
--set parses number only values as float64


databus23:
I just noticed that something like `helm install --set tag=20161216` ends up in scientific notation in the template and thats because `{{ typeOf .Value.tag }}` yields `float64`.

Thats even true for small integers like `helm install --set tag=1` --> `float64`.

This is happening with helm 2.1.0

technosophos:
Does it work differently if you do `--set tag=\"1\"`?

technosophos:
Thanks to @chancez we know the exact problem: During the conversions to and from JSON done by `ghodss/yaml`, integers are being cast into float64s to represent Javascript's numeric type. This is causing all integers above a certain value to be represented in scientific notation.

ipedrazas:
I've been bitten by the same problem. The way of getting over the hump (or the bug) is to do this:

`--set bignumber=\\"a185576882739235744\\"`

technosophos:
Another workaround we have used is to do stuff like this:

```
--set port=":1234567"
```

Then in the template:

```
{{ .Values.port | replace ":" "" }}
```

databus23:
Yuck! 😷

jemc:
This is a heck of a pain, and @ipedrazas' workaround doesn't seem to work for me, or any other style of quoting I've tried.

I'm not quite willing yet to swallow my pride and try @technosophos' ugly hack...

Smana:
I'm facing the same issue from time to time for the image tag.
I'll try to use one of these workaround. I hope it will be fixed soon :)

holybit:
In my case, seeing this below on `--set image.tag=5997578`:

```bash
$ kubetctl describe po msj-treasure-map-msj-treasure-map-192172122-dnb80
...
Events:
  Type     Reason         Age               From                                    Message
  ----     ------         ----              ----                                    -------
  Normal   Scheduled      1m                default-scheduler                       Successfully assigned msj-treasure-map-msj-treasure-map-192172122-dnb80 to ip-10-253-13-113.ec2.internal
  Warning  InspectFailed  15s (x9 over 1m)  kubelet, ip-10-253-13-113.ec2.internal  Failed to apply default image tag "596297932419.dkr.ecr.us-east-1.amazonaws.com/msj-trmap-uwsgi:5.997578e+06": couldn't parse image reference "596297932419.dkr.ecr.us-east-1.amazonaws.com/msj-trmap-uwsgi:5.997578e+06": invalid reference format
[...]

PR #6010, opened 2019-07-11
chartutil.ReadValues is forced to unmarshal numbers into json.Number refs #1707

osdrv:

This change is an attempt to address the common problem of json number
unmarshalling where any number is converted into a float64 and
represented in a scientific notation on a marshall call. This behavior
breaks things like: chart versions and image tags if not converted to
yaml strings explicitly.

An example of this behavior: k8s failure to fetch an image tagged with a
big number like: $IMAGE:20190612073634 after a few steps of yaml
re-rendering turns into: $IMAGE:2.0190612073634e+13.

Example issue: https://github.com/helm/helm/issues/1707

This commit forces yaml parser to use JSON modifiers and explicitly
enables interface{} unmarshalling instead of float64. The change
introduced might be breaking so should be processed with an extra care.

Due to the fact helm mostly dals with human-produced data (charts), we
have a decent level of confidence this change looses no functionality
helm users rely upon (the scientific notation).

**If applicable**:

### DIFF
--- a/cmd/helm/installer/install_test.go
+++ b/cmd/helm/installer/install_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package installer // import "k8s.io/helm/cmd/helm/installer"
 
 import (
+	"encoding/json"
 	"os"
 	"path/filepath"
 	"reflect"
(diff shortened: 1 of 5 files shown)


bacongobbler:
Wow... That's a surprisingly simple patch to fix this issue. Nice catch.

Would you be willing to port this over to the `dev-v3` branch as well? That way we can get this fix in for Helm 3. Thanks!

bacongobbler:
It turns out that this breaks certain unit tests from passing, however. Would you mind looking into the failing tests?

icanhazbroccoli:
@bacongobbler I really appreciate your instant reply! I'm working on the failing tests atm, will push a fix once I've figured out what's up. Will port the change to dev-v3 with pleasure if the change makes sense to you folks.

icanhazbroccoli:
A quick update on this PR: the recent fix seems to look quite promising for master and 2.x branches.
dev-v3 seems to be a more complicated case: the package dependency graph resolves ghodss/yaml library to version 1.0.0, which is missing the essential commit: https://github.com/ghodss/yaml/commit/8aa042675dd6a7951644be0acb134430ed0dd58b#diff-e51f89b39d976550831afb510a926be0. This library has never issued any other releases since 1.0.0, so it's a bit clumsy. We might consider getting in touch with the folks responsible for ghodss/yaml and asking them to create a meaningful tag (e.g. 1.1.0) so the dependency resolution is no longer tied to a revision.

Dependency declarators I've identified so far: 
* https://github.com/kubernetes/apiextensions-apiserver/blob/master/Godeps/Godeps.json#L126
* https://github.com/kubernetes/apiserver/blob/master/Godeps/Godeps.json#L102

In the current state of dev-v3 one of these components resolves yaml library to 1.0.0, not to c7ce16629ff4. Once it's fixed, the fix porting should be pretty trivial. I'll try to resolve it some time soon and update this thread.

icanhazbroccoli:
Hey folks, as https://github.com/helm/helm/pull/6016 and https://github.com/helm/helm/pull/6032 got merged into dev-v3, I believe the porting job is done. Please let me know if there are any other action points on my side in regards to this PR or if I can provide any other help. Thanks!

bacongobbler (approved):

icanhazbroccoli:
@bacongobbler <3

hickeyma (commented):
LGTM, thanks. Tested before and after fix and show number handled as expected.

PR #6011, opened 2019-07-11
feat(*) Adds atomic flag to v3

thomastaylor312:
This adds back in atomic functionality to v3. Please note this does not contain the `--cleanup-on-fail` flag yet as that will be done in a separate PR.

As part of adding unit tests, I ended up refactoring our fake `KubeClient`s into their own fake package and adding a new failing client so that we didn't end up creating a separate client for each type of failure we wanted. If this is too much for the PR, I can try breaking it out on its own

This is the v3 fix for #5143

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -113,11 +113,11 @@ $(GOIMPORTS):
 
 # install vendored dependencies
 vendor: Gopkg.lock
-	$(DEP) ensure -v --vendor-only
+	$(DEP) ensure --vendor-only
 
 # update vendored dependencies
 Gopkg.lock: Gopkg.toml
-	$(DEP) ensure -v --no-vendor
+	$(DEP) ensure --no-vendor
 
 Gopkg.toml: $(DEP)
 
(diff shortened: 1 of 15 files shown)


thomastaylor312:
Forgot to mention this is a port to v3 for #5143

bacongobbler (approved):
LGTM

Makefile
```diff
@@ -113,11 +113,11 @@ $(GOIMPORTS):
 
 # install vendored dependencies
 vendor: Gopkg.lock
-	$(DEP) ensure -v --vendor-only
+	$(DEP) ensure --vendor-only
```
adamreese: I'm telling @bacongobbler on you

adamreese (approved):

Makefile
bacongobbler: :scream:

thomastaylor312: I couldn't take the verbosity part of this, it was a bit too...verbose 😏

PR #6014, opened 2019-07-12
Use watcher with retries to wait for resources

riconnon:
When waiting for resources use the `ListWatchUntil` instead of
`UntilWithoutRetry` so that if the connection drops between tiller and
the API while waiting the operation can still succeed.

**If applicable**:

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -41,6 +41,7 @@ import (
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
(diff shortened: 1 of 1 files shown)


pkg/kube/client.go
```diff
@@ -809,7 +809,7 @@ func (c *Client) watchUntilReady(timeout time.Duration, info *resource.Info) err
 
 	ctx, cancel := watchtools.ContextWithOptionalTimeout(context.Background(), timeout)
 	defer cancel()
-	_, err = watchtools.UntilWithoutRetry(ctx, w, func(e watch.Event) (bool, error) {
+	_, err = watchtools.ListWatchUntil(ctx, w, func(e watch.Event) (bool, error) {
```
thomastaylor312: Based on the test failure output, it looks like the watcher we create with the resource helper isn't the right type to use with this function. You'll probably need to do something to convert the watch interface to the `cache.ListerWatcher` it expects

riconnon:
That was the conclusion I had also come to but I haven't had an opportunity to look into how we can pass a ListWatcher through.

pkg/kube/client.go
riconnon: @thomastaylor312
After a bit of fiddling I think I've got this right. CI run is now green.

thomastaylor312 (changes requested):
Sorry for the delay on this! Got caught up with the 3.0 beta. I think this is almost ready to go! Just one bug fix that needs to be handled

thomastaylor312 (approved):
This is looking good! Thanks for the changes!

Issue #5950, opened 2019-06-27
Helm 3 does not see to respect the --kubeconfig flag


limscoder:
Output of `helm version`:

version.BuildInfo{Version:"v3.0.0-alpha.1", GitCommit:"b9a54967f838723fe241172a6b94d18caf8bcdca", GitTreeState:"clean"}

Output of `kubectl version`:

Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.1", GitCommit:"b7394102d6ef778017f2ca4046abbaa23b88c290", GitTreeState:"clean", BuildDate:"2019-04-08T17:02:58Z", GoVersion:"go1.12.1", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.1", GitCommit:"b7394102d6ef778017f2ca4046abbaa23b88c290", GitTreeState:"clean", BuildDate:"2019-04-08T17:02:58Z", GoVersion:"go1.12.1", Compiler:"gc", Platform:"linux/amd64"}

V3 does not seem to respect the --kubeconfig flag.

Example:

```
[root@host kubernetes]# helm list
NAME        	NAMESPACE	REVISION	UPDATED                                	STATUS  	CHART
my-chart	default  	1       	2019-06-27 17:20:22.438150461 +0000 UTC	deployed	my-chart-0.1.0
[root@host kubernetes]# mv ~/.kube/config ~/.kube/config.bak
[root@host kubernetes]# helm list
NAME	NAMESPACE	REVISION	UPDATED	STATUS	CHART
[root@host kubernetes]# helm --kubeconfig=~/.kube/config.bak list
NAME	NAMESPACE	REVISION	UPDATED	STATUS	CHART
[root@host kubernetes]#
```

limscoder:
it appears that setting KUBECONFIG works as expected.

hickeyma:
Thanks @limscoder. This seems to be a bug alright.

I tested it as follows:

```console
$ cat ~/.kube/config
apiVersion: v1
clusters:
- cluster:
    server: http://127.0.0.1:32768
  name: dind
- cluster:
    server: ""
  name: test
contexts:
- context:
    cluster: dind
    namespace: default
    user: ""
  name: dind
- context:
    cluster: dind
    namespace: test
    user: ""
  name: test
current-context: test
kind: Config
preferences: {}
users: []

$ helm ls
[...]

PR #6025, opened 2019-07-15
fix(client): Fixes a timing issue with reading client flags

ian-howell:
This fixes an issue where the kubernetes client was being created using
command line flags before those flags had been parsed, causing the
client to always use the default values.

Addresses https://github.com/helm/helm/issues/5950

**If applicable**:


thomastaylor312 (approved):
Thanks for the fix!

PR #5448, opened 2019-03-14
WIP: ref(cmd): migrate to pkg/cli

bacongobbler:
This commit migrates all of the content over to pkg/cli, making it consumable by third parties.

closes #5311

Signed-off-by: Matthew Fisher <matt.fisher@microsoft.com>

PR #6026, opened 2019-07-15
ref(rollback): use `helm upgrade` logic for `helm rollback`

bacongobbler:
A rollback is an "upgrade" of an older revision. This change removes all of the one-off logic from `helm rollback`, relying on the business logic backing `helm upgrade`.

This comes with several advantages:

- uniform behaviour with regards to feature flags like `--force`, `--recreate-pods` and `--no-hooks`
- features implemented for `helm upgrade` can be enabled for `helm rollback` with no significant development cost (for example #6011 could be ported to `helm rollback` by simply adding it to rollback's flagset).

I also caught an unused flag. `--version` was unused. The mandatory argument `[REVISION]` specifies the version to rollback to. Because of this, I removed it.

BREAKING CHANGES:

This change removes the `pre-rollback` and `post-rollback` hooks, opting to use the existing `pre-upgrade` and `post-upgrade` flags. These flags are not implemented by any of the charts in the incubator or stable repositories, so it appears safe to remove. However, this is a compatibility breaker from what's available in Helm 2. Any opinions here on how to proceed here?


cmd/helm/rollback.go
```diff
@@ -36,15 +37,30 @@ second is a revision (version) number. To see revision numbers, run
...
 
 	cmd := &cobra.Command{
 		Use:   "rollback [RELEASE] [REVISION]",
 		Short: "roll back a release to a previous revision",
 		Long:  rollbackDesc,
 		Args:  require.ExactArgs(2),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			_, err := client.Run(args[0])
+
+			version, err := strconv.Atoi(args[1])
```
thomastaylor312: My one concern about removing the `rollback` action is that it makes use as a library more difficult because anyone who wants to rollback will have to duplicate this same block of added logic. If we are going to do this, we should create a helper function that grabs all the right data, merges values, and returns back the chart

thomastaylor312 (commented):
Code looks good to me. I am not sure on how to handle the breaking change though, I think we may want to directly translate `pre/post-rollback` hooks to `pre/post-upgrade` and then print a deprecation warning. I think we give it until something like 3.4-ish timeframe before we fully remove it. That should make it so no one has to change charts immediately.

However, even with this, we end up with a weird division of terminology. If you are using Helm as a library, there will be no such thing as a `rollback` action, so having the `rollback` hooks doesn't make much sense, but if we keep having a `helm rollback` command, then we should keep the `rollback` hooks because that command exists

bacongobbler:
> I think we may want to directly translate `pre/post-rollback` hooks to `pre/post-upgrade` and then print a deprecation warning.

That would cause pre- and post-rollback hooks to run during an upgrade. That seems like a dangerous/unintended side-effect that I wanted to avoid in this case, so it seemed safer to ignore the hook.

After searching for the term `helm.sh/hook: pre-rollback`, most of the hooks appear to be registered as a `helm.sh/hook: pre-install,pre-upgrade,pre-rollback` hook, meaning they are already running as a pre- or post-upgrade hook.

The only interesting case was in [a stale project](https://github.com/javierrivassyapse/versions-service/blob/17a56fec9e1138d590623804858d87fee2db88f8/helm/versions_service/hooks/reset-db-job.yaml) where the pre-rollback hook was used to roll back to an older migration table. Most sane schema updates are non-destructive by design because the software running in production must not break when the updates are applied. Even if you have to rollback production code, you can often keep the updated schema, then simply apply a few additional migrations, if necessary, rather than rolling back and re-applying a modified version of the original migration.

In other words, I believe that this corner case isn't something where a rollback-specific hook makes sense. The application should be designed to handle table inputs from previous versions of the API.

Given this, perhaps the better approach may be to accept pre- and post-rollback hooks, but ignore them. In the case where they're registered as a `pre-upgrade,pre-rollback` hook (like in the example above), they'll continue to be triggered during a `helm rollback`.

cmd/helm/rollback.go
bacongobbler: The same argument could be applied to `helm template`. There was an initiative to expose the CLI-specific logic in #5448, though that stalled out because of certain design issues with the `settings` object that made it difficult to expose as a consumable package.

bacongobbler:
closing in favour of https://github.com/helm/helm/pull/6133

PR #6016, opened 2019-07-12
Replaced ghodss/yaml with sigs.k8s.io/yaml [dev-v3]

osdrv:
**What this PR does / why we need it**:

This PR proposes a replacement of github.com/ghodss/yaml with it's forked version maintained by SIG community. The replaceable library has low-to-none support activity unlike the latter. The SIG version is marked as a permanent fork. We believe the new Helm branch (v3) could benefit from using the Kubernetes community-maintained version on a long-term run as yaml parser is a key component of Helm chart rendering engine.

This commit locks sigs.k8s.io/yaml dependency version on 1.1.0 which is backwards compatible with ghodss/yaml 1.0.0 (this statement needs a remark: strictly speaking, the change **is** breaking: see the relevant comments: https://github.com/ghodss/yaml/issues/50#issuecomment-465777941 and https://github.com/ghodss/yaml/issues/50#issuecomment-511004721, but after a deep dive, Helm and it's dependencies are believed to be backwards compatible with the change).

This change also resolves the outdated dependency version lock for ghodss/yaml (currently 1.0.0) and makes it possible to port changes from https://github.com/helm/helm/pull/6010 to dev-v3.

Signed-off-by: Oleg Sidorov <oleg.sidorov@booking.com>

**Special notes for your reviewer**:
I was asked to port changes from https://github.com/helm/helm/pull/6010 to dev-v3. If this change makes sense, porting the fix would be a pretty trivial follow-up step.

**If applicable**:
- [ ] this PR contains documentation
- [ ] this PR contains unit tests
- [x] this PR has been tested for backwards compatibility

PR #6032, opened 2019-07-17
chartutil.ReadValues is forced to unmarshal numbers into json.Number refs #1707 [dev-v3]

osdrv:

Backport of https://github.com/helm/helm/pull/6010 to dev-v3 (the description below is a copy-paste from the original v2 branch PR). As https://github.com/helm/helm/pull/6016 is now merged to dev-v3, the change is reasonably trivial.

This change is an attempt to address the common problem of json number unmarshalling where any number is converted into a float64 and represented in a scientific notation on a marshall call. This behavior breaks things like: chart versions and image tags if not converted to yaml strings explicitly.

An example of this behavior: k8s failure to fetch an image tagged with a big number like: $IMAGE:20190612073634 after a few steps of yaml re-rendering turns into: $IMAGE:2.0190612073634e+13.

Example issue: #1707

This commit forces yaml parser to use JSON modifiers and explicitly enables interface{} unmarshalling instead of float64. The change introduced might be breaking so should be processed with an extra care.

Due to the fact helm mostly dals with human-produced data (charts), we have a decent level of confidence this change looses no functionality helm users rely upon (the scientific notation).

This PR is coupled with https://github.com/helm/helm/pull/6010, which is still relevant (imo) for branch 2.  

**If applicable**:

### DIFF
--- a/pkg/chartutil/testdata/coleridge.yaml
+++ b/pkg/chartutil/testdata/coleridge.yaml
@@ -10,3 +10,4 @@ water:
   water:
     where: "everywhere"
     nor: "any drop to drink"
+    temperature: 1234567890
--- a/pkg/chartutil/values.go
+++ b/pkg/chartutil/values.go
@@ -17,6 +17,7 @@ limitations under the License.
 package chartutil
 
 import (
+	"encoding/json"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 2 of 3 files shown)


bacongobbler (approved):

icanhazbroccoli:
Thanks a lot for a super fast merge @bacongobbler !

hickeyma (approved):
LGTM, thanks. Tested before and after fix and show number handled as expected.

Issue #6037, opened 2019-07-18
helm3 value.yaml ingress path is a string?


liuliqiang:
### Output of `helm version`:

```
helm version
version.BuildInfo{Version:"v3.0.0-alpha.1", GitCommit:"b9a54967f838723fe241172a6b94d18caf8bcdca", GitTreeState:"clean"}
```
### Output of `kubectl version`:

```
kubectl version
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.1", GitCommit:"b7394102d6ef778017f2ca4046abbaa23b88c290", GitTreeState:"clean", BuildDate:"2019-04-08T17:11:31Z", GoVersion:"go1.12.1", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.2", GitCommit:"66049e3b21efe110454d67df4fa62b08ea79a19b", GitTreeState:"clean", BuildDate:"2019-05-16T16:14:56Z", GoVersion:"go1.12.5", Compiler:"gc", Platform:"linux/amd64"}
```

### Cloud Provider/Platform (AKS, GKE, Minikube etc.): 

- Cluster build with kubeadm

### Problem

Hello, In helm2，i found a pr https://github.com/helm/helm/pull/4703/files which change `ingress.path` in **value.yaml**  from a stirng to list，but in helm3-alpha，https://github.com/helm/helm/blob/dev-v3/pkg/chartutil/create.go#L101， it's also string，should it be changed?

bacongobbler:
Yes, that would be a good patch to port over. Feel like patching that?

liuliqiang:
@bacongobbler  i would like to do that, i will take a patch now.

PR #6038, opened 2019-07-18
Fix paths in the ingress template and values file written by helm create

liuliqiang:
Signed-off-by: Liu Liqiang <liuliqiang@smartx.com>

Fix issue #6037 

**If applicable**:

### DIFF
--- a/pkg/chartutil/create.go
+++ b/pkg/chartutil/create.go
@@ -98,9 +98,10 @@ ingress:
   annotations: {}
     # kubernetes.io/ingress.class: nginx
     # kubernetes.io/tls-acme: "true"
-  path: /
   hosts:
-    - chart-example.local
+    - host: chart-example.local
+      paths: []
+
   tls: []
   #  - secretName: chart-example-tls
   #    hosts:
@@ -150,7 +151,7 @@ const defaultIgnore = `# Patterns to ignore when building packages.
 
 const defaultIngress = `{{- if .Values.ingress.enabled -}}
 {{- $fullName := include "<CHARTNAME>.fullname" . -}}
-{{- $ingressPath := .Values.ingress.path -}}
+{{- $svcPort := .Values.service.port -}}
 apiVersion: networking.k8s.io/v1beta1
 kind: Ingress
 metadata:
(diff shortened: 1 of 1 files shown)


pkg/chartutil/create.go
```diff
@@ -98,7 +98,7 @@ ingress:
   annotations: {}
     # kubernetes.io/ingress.class: nginx
     # kubernetes.io/tls-acme: "true"
-  path: /
+  paths: []
   hosts:
     - chart-example.local
```
hickeyma: `paths: []`  should be part of `host`

pkg/chartutil/create.go
```diff
@@ -150,7 +150,7 @@ const defaultIgnore = `# Patterns to ignore when building packages.
 
 const defaultIngress = `{{- if .Values.ingress.enabled -}}
 {{- $fullName := include "<CHARTNAME>.fullname" . -}}
-{{- $ingressPath := .Values.ingress.path -}}
+{{- $ingressPaths := .Values.ingress.paths -}}
```
hickeyma: Should instead be: `{{- $svcPort := .Values.service.port -}}`

hickeyma (changes requested):
@liuliqiang Thanks for pushing this PR. However, a lot has changed with ingress since PR #4703. I would suggest looking at the current version of ingress in Helm v2 and update this PR accordingly.

liuliqiang:
@hickeyma thanks for your comment，i learn a lot from these comments, i will modify my patch and make it better. 😀

liuliqiang:
hi @hickeyma  , i have check the master branch and update this patch，i hope i doesn't miss anything.

pkg/chartutil/create.go
```diff
@@ -159,28 +160,30 @@ metadata:
 {{ include "<CHARTNAME>.labels" . | indent 4 }}
   {{- with .Values.ingress.annotations }}
   annotations:
-{{ toYaml . | indent 4 }}
-{{- end }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
```
hickeyma: Decrease by 2 indents so `{{- end }}` is inline with `{{- with .Values.ingress.annotations }}`

hickeyma (changes requested):
@liuliqiang Looking good.Just one small indentation nit. When you make the change, can you also squash the changes into one commit?

pkg/chartutil/create.go
liuliqiang: Ok，i will do it now，i am sorry for that。

liuliqiang:
hi @hickeyma ，i have merge commits into one.
And also align the `{{- with .Values.ingress.annotations }}` and `{{end}}`

hickeyma (approved):
LGTM, thanks @liuliqiang.

Issue #5628, opened 2019-04-22
helm 3: --namespace not respected


ericvmw:
It seems the latest helm3 from dev-v3 branch doesn't take the --namespace parameter. Basically I try to deploy an example chart onto a different namespace, but it doesn't seem to take effect. Below are the commands and outputs. How is helm3 planning to support namespace and how should namespaces be specified in helm charts for helm3? Thanks for any help / pointers!

Output of `helm version`:
version.BuildInfo{Version:"v3.0+unreleased", GitCommit:"658c66dc66684a85be787a066c8434ac5212648c", GitTreeState:"clean"}

Output of `kubectl version`:
Client Version: version.Info{Major:"1", Minor:"11", GitVersion:"v1.11.1", GitCommit:"b1b29978270dc22fecc592ac55d903350454310a", GitTreeState:"archive", BuildDate:"2018-08-14T19:36:17Z", GoVersion:"go1.10.3", Compiler:"gc", Platform:"linux/amd64"}

# /usr/local/helm3/linux-amd64/helm install alpine  /usr/local/alpine/ --namespace=us1-dev-web-1 --debug
printer.go:84: [debug] Original chart version: ""
printer.go:84: [debug] CHART PATH: /usr/local/alpineclient.go:94: [debug] building resources from manifest
client.go:99: [debug] creating 1 resource(s)
NAME: alpine
LAST DEPLOYED: 2019-04-19 17:07:30.391894983 +0000 UTC m=+0.063175512
NAMESPACE: default
STATUS: deployed

# kubectl get pod -n default
NAME            READY     STATUS    RESTARTS   AGE
alpine-alpine   1/1       Running   0          6m

bacongobbler:
There was a change in Helm 3 where it now takes the current namespace from your local kube config. If it's not present, the default namespace is used.

To change the current namespace for the alpha, you can use

```
kubectl set-context NAME --namespace=foo
```

Helm will pick that up and use that namespace the next time you invoke it.

There's a few design refactors going on within Helm 3 to address prioritization issues with loading settings from CLI flags and the kube config. As I understand it, the --namespace flag was temporarily removed but it's planned to allow some way to change the namespace on-the-fly.

@adamreese do you have any other context I may be missing here on why we removed the --namespace flag?

bacongobbler:
For reference: https://github.com/helm/helm/blob/658c66dc66684a85be787a066c8434ac5212648c/cmd/helm/helm.go#L108-L113

[...]

PR #6046, opened 2019-07-19
fix(kube): remove namespace enforcement

### LINKED ISSUE #5953: Helm 3: Does not allow charts with resources in multiple namespaces
Output of `helm version`:

v3.0.0-alpha.1

Helm errors when the chart contains a resource in a different namespace than the current kube-context, even when the user has cluster-admin access. 

```
the namespace from the provided object \"target\" does not match the namespace \"default\". You must pass '--namespace=target' to perform this operation.
```

This prevents charts with resources in multiple namespaces from being installed.

bacongobbler:
This fixes an issue where resources that hardcode the `metadata.namespace` parameter cannot be installed.

closes #5953

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -114,7 +114,6 @@ func (c *Client) newBuilder() *resource.Builder {
 		ContinueOnError().
 		NamespaceParam(c.namespace()).
 		DefaultNamespace().
-		RequireNamespace().
 		Flatten()
 }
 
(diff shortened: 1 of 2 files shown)


bacongobbler:
Given that this is a regression from Helm 2's behaviour, this should probably go into alpha.2 as a relatively minor bugfix.

bacongobbler:
Big thanks to @kri5 for identifying the fix in https://github.com/helm/helm/issues/5628#issuecomment-505453318.

kri5:
@bacongobbler You're welcome, thanks for fixing the issues (Wasn't sure if that was the fix or just a workaround :D)

thomastaylor312 (approved):
LGTM

PR #6052, opened 2019-07-20
feat(helm): lint dependent charts with 'helm lint'

### LINKED ISSUE #5767: helm lint to lint also dependent charts
Output of `helm version`:
Not applicable.

Output of `kubectl version`:
Not applicable.

Not applicable.

Currently `helm lint` does not fail when there is an error in dependent chart. `helm lint --strict` fails which is surprising for as the helm of the --strict flag is `fail on lint warnings` - which means that helm lint should detect warnings and --strict just to transform the warnings to errors.

It will be useful if `helm lint` is able to lint also dependent charts (in `charts/`).

nlee737:
With this PR, `helm lint` will also lint a chart's dependent charts. Closes #5767.

**If applicable**:


bridgetkromhout:
Hi, @mnkyl! Thanks for your work on this! We definitely want this, but we don't want to surprise people with a breaking change. Can you implement this using `--with-subcharts`? And as Helm 3 is almost upon us, the place you'll want to target this PR is the https://github.com/helm/helm/tree/dev-v3 branch. Thanks!

mnkyl:
> 
> 
> Hi, @mnkyl! Thanks for your work on this! We definitely want this, but we don't want to surprise people with a breaking change. Can you implement this using `--with-subcharts`? And as Helm 3 is almost upon us, the place you'll want to target this PR is the https://github.com/helm/helm/tree/dev-v3 branch. Thanks!

Thank you for letting me know - I've created a new PR at #7288.

PR #5110, opened 2018-12-29
Add various to helm create

naseemkullah:
**What this PR does / why we need it**:
Modifies the `helm create` command as follows:
 * ~Moves all deployment related values to an `.Values.deployment` umbrella as this makes more sense. (The same way ingress and service values are defined.)~  <- vetoed
* Adds an easy way to create an hpa for the deployment if the user so chooses to do so
* Adds an easy way to add environment variables for the deployment if the user so chooses to do so
* ~Makes deployment's probe paths/ports configurable~ no need really in retrospect, to have them as a value or directly done in the template, no difference for starter chart
* ~Adds an easy way to add security contexts for deployments, be it for pod or container~ Already tackled in seperate PR
* Adds an easy way to add volumes for deployments

**Special notes for your reviewer**:
Feedback is most welcome.

**If applicable**:
- [x] this PR contains documentation
- [x] this PR contains unit tests
- [x] this PR has been tested for backwards compatibility

PR #6053, opened 2019-07-20
[helm create] Annotations

naseemkullah:
Eases the addition of annotations (e.g. `prometheus.io/scrape: "true"` or  `"cluster-autoscaler.kubernetes.io/safe-to-evict": "true"`) for end user creating a chart.


hickeyma:
cc @unguiculus @scottrigby you might be best placed to review this.

unguiculus (approved):
I think it makes sense to add these.

scottrigby:
I agree with @unguiculus these make sense. However what about syntax… in some places we wrap these keys `{{- with .Values.KEY }}` (`ingress.annotations`, `imagePullSecrets`, `nodeSelector`, `affinity`, `tolerations`), but in other places we don't (`resources`). Should we be consistent about this while we're at it?

scottrigby:
PS slightly related to some other keys in #5110

naseemkullah:
Good point @scottrigby , although wrapping with `{{- with .Values.KEY }}` is more popular, I like the approach used in `resources` because its 2 lines less. 
But I'm open to whichever, please advise.

scottrigby:
I'd suggest use `with` for consistency. Additional benefit is the key won't be part of the rendered template unless a user passes custom values.

naseemkullah:
Sounds good, should this be done for `resources` as well?

scottrigby:
I'm not asking you to increase the scope of this small PR, but if you want to do that I'm a fan of consistency, especially when it's easy.

naseemkullah:
Sure @scottrigby not a problem, but while on the topic of consistency:
```
      {{- with .Values.nodeSelector }}
      nodeSelector:
        {{- toYaml . | nindent 8 }}
      {{- end }}
    {{- with .Values.affinity }}
      affinity:
        {{- toYaml . | nindent 8 }}
    {{- end }}
```
Should the `{{- with` align with the key or 2 spaces before it?

Also is everyone else on board with wrapping any ` toYaml` type thing with `{{- with` ? This will include `resources`,  `securityContext`s as well as the hereby added `annotations`?

@scottrigby just  playing devil's advocate: even if the rendered templates contain empty keys if values not et for the non `{{-with`-wrapped, when posted to the k8s api server the resulting objects will be the same. And going with the non-wrapped approach (as resources is e.g.) will result in way less lines and slightly more readable template imho.

naseemkullah:
bump

naseemkullah:
bump and OK I've wrapped all `toYaml` type stuff with `with` ...

scottrigby (approved):
Nice lgtm

scottrigby:
This failed test seems unrelated

naseemkullah:
Merged master into this branch in hopes to pass the failed test.

naseemkullah:
Feel free to merge. :)

bacongobbler:
docs?

naseemkullah:
@bacongobbler Could we get these into helm 3 as well?

bacongobbler:
We're only porting changes that were merged into Helm 2 into Helm 3 at this time. Feature additions that weren't merged into Helm 2 will have to wait until Helm 3 is released; we're trying to avoid delaying the release of helm 3 by getting bogged down in features.

bridgetkromhout:
Hi, @naseemkullah - Thanks again for your work on this! To clarify, We're going to be cutting the last feature release for v2 soon, and this PR won't be in v2, so this one should be targeted to the dev v3 branch: https://github.com/helm/helm/tree/dev-v3. You probably also want to keep working with the charts maintainers. Thanks!

naseemkullah:
Thanks for clarifying @bridgetkromhout 
Oh @bacongobbler was it not possible to change base branch  to `dev-v3`?

hickeyma:
@naseemkullah Unfortunately no. You have to restart from the dev-v3 branch.

naseemkullah:
Ok no problem, thanks

Issue #6020, opened 2019-07-12
Helm 3: Test Enhancement Proposal


jlegrone:
### Features

- [x] Support defining tests as `Jobs` (#6054)
- [x] Enable the management of resources (eg. service accounts, network policy, configmaps) to be made available during test execution (#6054)
- [ ] Stream execution logs & events with a flag (eg. `--verbose`) -- see issues #2298 and #3481
- [x] Enable control over resource deletion behavior on a per-resource basis (may be overridden globally with a flag, eg. `--skip-hook-cleanup`) (#6054)

### Breaking Changes

- [x] Remove `test-failure` hook ([only 20 usages public on Github](https://github.com/search?l=YAML&q=helm.sh%2Fhook+test-failure&type=Code)) (#6054)
- [x] Remove the existing `--cleanup` flag (#6363)
- [ ] Make resource cleanup the default (on success OR failure), and add a flag to skip cleanup for all resources (`--skip-hook-cleanup`)

### Deprecations

- [x] Rename the `test-success` annotation value to `test`. Keep supporting `test-success` but log a warning to the user. (#6054)

### 3.x and Beyond

1. Surface relevant logs and k8s events to users during hook execution or on failure
1. Support structured output with `--output json`, eg:
   ```json
   {
     "logs": [
       {
         "pod": "foo",
         "output": "helloworld"
       }
     ],
     "events": [
       {
         "apiVersion": "v1",
         "involvedObject": {
           "apiVersion": "v1",
           "kind": "Pod",
           "name": "foo"
         },
         "message": "Created container bar",
         "reason": "Created"
       }
     ]
   }
   ```

## Implementation

1. Specify tests and supporting resources with the `helm.sh/hook: test` annotation
1. Converge the business logic for test execution, log output, deletion policy, etc with existing hook implementations
1. Add a `--skip-hook-cleanup` flag to `install`, `upgrade`, `delete`, and `test` commands which will override per-resource `helm.sh/hook-delete-policy` behavior
1. Eventually support log streaming and printing of related events on failure for all Helm hooks

## Other Considerations

Some of the desired features for `helm test` (1 and 2) are already supported by the hook system, while the others would likely be welcome enhancements to existing hooks.

Continuing to share the hooks annotation-based API with `test` is advantageous in that users have only one API to learn, and if desired some tests may also be executed as hooks. The latter is a use case for charts that define "smoke" tests which can do things like prevent upgrades when the target release is in a bad state.

[...]

PR #6054, opened 2019-07-20
Support defining tests as Job resources

jlegrone:

This is a partial implementation of #6020

### New Features
- Allow defining tests as `Job` resources
- Enable creating arbitrary supporting resources (other than Pods) during `helm test`
- Watch `Pod` resources for success during hook execution (required to avoid a breaking change to `helm test`)

### Breaking Changes
- Remove `test-failure` hook
- Long-running Pods created by any hook will cause hook execution to time out
    - Workaround: switch to a `Replicaset` or `Deployment`

### Refactors
- Converge on a single implementation for execution of hooks during `helm install`, `helm upgrade`, `helm delete`, and `helm test`
- Rename `test-success` to `test` (but continue supporting `test-success` as an alias for `test`)

**If applicable**:


pkg/releaseutil/manifest_sorter.go
```diff
@@ -53,16 +52,16 @@ type result struct {
...
+	release.HookPreInstall.String():   release.HookPreInstall,
+	release.HookPostInstall.String():  release.HookPostInstall,
+	release.HookPreDelete.String():    release.HookPreDelete,
+	release.HookPostDelete.String():   release.HookPostDelete,
+	release.HookPreUpgrade.String():   release.HookPreUpgrade,
+	release.HookPostUpgrade.String():  release.HookPostUpgrade,
+	release.HookPreRollback.String():  release.HookPreRollback,
+	release.HookPostRollback.String(): release.HookPostRollback,
+	release.HookTest.String():         release.HookTest,
+	"test-success":                    release.HookTest,
```
michelleN: Might be good to put his string in a variable or constant and also leave a comment here explaining this is a for legacy support and should be removed later.

cmd/helm/release_testing_run.go
```diff
@@ -44,27 +42,7 @@ func newReleaseTestRunCmd(cfg *action.Configuration, out io.Writer) *cobra.Comma
...
-						break
-					}
-
-					if res.Status == release.TestRunFailure {
-						testErr.failed++
-					}
-					fmt.Fprintf(out, res.Msg+"\n")
-				}
-			}
+			return client.Run(args[0])
```
michelleN: Looks like we're not supporting running tests in parallel at the moment in helm 3 so I like this simplification.

docs/chart_tests.md
```diff
@@ -73,11 +71,13 @@ STATUS: DEPLOYED
...
 
 2. `$ helm test quirky-walrus`
+
 ```
 RUNNING: quirky-walrus-credentials-test
 SUCCESS: quirky-walrus-credentials-test
 ```
 
 ## Notes
+
```
michelleN: Maybe we could add in a line about how we still support the test-success annotation temporarily as we transition over to the test annotation under notes for clarity?

pkg/action/hooks.go
```diff
@@ -0,0 +1,90 @@
+package action
```
michelleN: `pkg/action` is the package that maps client verbs to logic so that folks who want to import the client logic can do so. In `helm 2.x`, this all lived under `cmd/helm` and was hard to import/reuse. I'm not sure that the hook execution logic should live here. Did you have some thoughts around you wanted to have this logic in `pkg/action` and not `pkg/release`?

pkg/action/release_testing.go
```diff
@@ -43,52 +42,16 @@ func NewReleaseTesting(cfg *Configuration) *ReleaseTesting {
...
-	go func() {
-		defer close(errc)
-		defer close(ch)
-
-		if err := tSuite.Run(testEnv); err != nil {
-			errc <- errors.Wrapf(err, "error running test suite for %s", rel.Name)
-			return
-		}
-
-		rel.Info.LastTestSuiteRun = &release.TestSuite{
```
michelleN: Could you talk us through why you want to delete this part? Do you plan on re-implementing?

michelleN (changes requested):
I manually tested both success and failure scenarios and it works great. 👍  I do have some questions below around getting some clarity on code changes. Thank you!

pkg/action/hooks.go
jlegrone: I ran into some issues with circular dependencies which is why I ended up leaving this here. All of these methods are private to the actions package however so it'll be possible to refactor them out into a public API in the future... this felt like the most minimal change possible at the moment though. :)

pkg/action/release_testing.go
jlegrone: Replaced with generic hook execution tracking in 593b6a04b9e88ca490816991cb7e2723c63ef088.

docs/chart_tests.md
jlegrone: Docs no longer live here... will have to do a follow up for helm/helm-www later. 😄

michelleN (approved):

pkg/releaseutil/manifest_sorter.go
jlegrone: Leaving as a string literal for now -- I don't want to encourage anyone to reference this since at the moment this is the only place we need to use the value.

bacongobbler (changes requested):
I hit a few compiler errors when trying to run `make test` against this PR. Would you mind looking into these?

```
><> make test
go build  -tags '' -ldflags '-w -s -X helm.sh/helm/internal/version.gitCommit=52de47e02ae4b41d6b3bb4a790d7098dfdb1cad4 -X helm.sh/helm/internal/version.gitTreeState=dirty' -o /home/bacongobbler/code/go/src/helm.sh/helm/bin/helm helm.sh/helm/cmd/helm
/home/bacongobbler/code/go/bin/golangci-lint run
pkg/action/hooks.go:8: File is not `goimports`-ed (goimports)
        "github.com/pkg/errors"
cmd/helm/release_testing_run.go:56:6: U1000: type `testErr` is unused (unused)
type testErr struct {
     ^
Makefile:81: recipe for target 'test-style' failed
```

If you rebase against dev-v3 these errors should pop up now that #6119 has been merged. Thanks!

bacongobbler:
Would you mind providing test steps and documentation on the changes as well? I know users will be asking how to use Jobs for tests as well as the changes involved here.

The Helm 3 documentation is hosted in the [helm-www repo, under the dev-v3-docs branch](https://github.com/helm/helm-www/tree/dev-v3-docs). Let me know how I can help.

jlegrone:
Thanks @bacongobbler, I've rebased from dev-v3 and fixed broken lints.

> Would you mind providing test steps and documentation on the changes as well?

For sure, I did update some documentation but then it got moved to helm-www. 😄Happy to take care of that in a follow up PR.

jlegrone:
Hmm... https://github.com/helm/helm/pull/6085 made a very similar hook execution refactor. Going to take a while to fix conflicts.

jlegrone:
Merge conflicts are resolved, but this is currently untested by me. @thomastaylor312 your input would be appreciated on https://github.com/helm/helm/pull/6054/commits/5ab42504f18365522d2b495b61a6e499c724a096.

thomastaylor312:
Sorry @jlegrone I forgot this PR was in flight when I refactored that. I'll take a look at the commit now

thomastaylor312:
@jlegrone It looks like you merged it just fine. I'll take another look once you get things merged

bacongobbler:
When testing `helm test run` against this PR with an existing chart that implements `helm test`, I noticed that it immediately exited with an exit code zero.

```
><> ./bin/helm test run vault
><> echo $?
0
```

The pod was spawned, but the CLI would not wait for the test pod to run to completion.

This differs from dev-v3, which spawns the test pod, waits for it to complete, then returns its output.

```
><> ./bin/helm test run vault
RUNNING: vault-vault-status-test
FAILED: vault-vault-status-test, run `kubectl logs vault-vault-status-test --namespace default` for more info
```

Can you please look into this and make sure existing test hooks continue to work as before? If not, can you please provide steps to properly test these changes? Thanks.

bacongobbler:
Same thing occurs when I change the test hook from `test-success` to `test`, so it appears to be a regression at first glance.

jlegrone:
@bacongobbler thanks for double checking, I didn't have time to validate the merge earlier.

The chart you mentioned is a good test case; helm should now wait for its tests to pass. `helm status vault` should also report information about the last test run.

jlegrone:
Looks like there may have been another regression with job hooks -- I ran into issues with `stable/spinnaker`. Please hold for now.

bacongobbler:
Hey, just getting back from a long weekend. Will test this out today. If you want to address the merge conflict from merging #5443 in the meantime, that'd be great :) Thanks!

jlegrone:
Conflicts are fixed in https://github.com/jlegrone/helm/tree/test-as-hook; not sure why those commits aren't showing up on the PR atm though.

bacongobbler:
`helm test run --cleanup` doesn't seem to work either. Would you please write some tests to ensure these cases continue to work?

```
><> ./bin/helm test run vault --cleanup
Pod vault-vault-status-test pending
Pod vault-vault-status-test pending
Pod vault-vault-status-test succeeded
><> k get po
NAME                          READY   STATUS      RESTARTS   AGE
vault-vault-cdb859c4c-8m4jk   1/1     Running     0          119s
vault-vault-cdb859c4c-pnc7f   1/1     Running     0          119s
vault-vault-cdb859c4c-v7vmm   1/1     Running     0          119s
vault-vault-status-test       0/1     Completed   0          4s
```

on dev-v3:

```
><> ./bin/helm test run vault --cleanup
RUNNING: vault-vault-status-test
PASSED: vault-vault-status-test
><> k get po
NAME                          READY   STATUS    RESTARTS   AGE
vault-vault-cdb859c4c-mqkgc   1/1     Running   0          31s
vault-vault-cdb859c4c-nmf2l   1/1     Running   0          31s
vault-vault-cdb859c4c-qkcfl   1/1     Running   0          31s
```

pkg/action/release_testing.go
```diff
@@ -43,52 +42,21 @@ func NewReleaseTesting(cfg *Configuration) *ReleaseTesting {
...
-			StartedAt:   tSuite.StartedAt,
-			CompletedAt: tSuite.CompletedAt,
-			Results:     tSuite.Results,
-		}
-
-		if r.Cleanup {
-			testEnv.DeleteTestPods(tSuite.TestManifests)
```
bacongobbler: I think this is the part that needs to be re-implemented for `--cleanup` to work.

pkg/action/upgrade.go
```diff
@@ -22,11 +22,11 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+
```
bacongobbler: this newline can be removed

pkg/kube/client.go
```diff
@@ -422,6 +427,30 @@ func (c *Client) waitForJob(obj runtime.Object, name string) (bool, error) {
...
+// waitForPodSuccess is a helper that waits for a pod to complete.
+//
+// This operates on an event returned from a watcher.
+func (c *Client) waitForPodSuccess(obj runtime.Object, name string) (bool, error) {
+	o, ok := obj.(*v1.Pod)
+	if !ok {
+		return true, errors.Errorf("expected %s to be a *v1.Pod, got %T", name, obj)
+	}
+
+	switch o.Status.Phase {
```
bacongobbler: Would it make sense to add a `default` switch case here to capture other pod statuses and report back? For example, how should the user handle pod phases like `PodUnknown`? What about other pod phases that may be introduced in the future?

pkg/release/hook.go
```diff
@@ -41,13 +43,22 @@ type HookDeletePolicy string
 
 // Hook delete policy types
 const (
-	HookSucceeded          HookDeletePolicy = "succeeded"
-	HookFailed             HookDeletePolicy = "failed"
+	HookSucceeded          HookDeletePolicy = "hook-succeeded"
```
bacongobbler: would you mind explaining the reason behind this change? It looks like this may break backwards compatibility with charts that defined their hook deletion policies, which encompasses way more than `helm test`

pkg/release/hook.go
```diff
@@ -41,13 +43,22 @@ type HookDeletePolicy string
...
+	HookFailed             HookDeletePolicy = "hook-failed"
```
bacongobbler: same here

pkg/release/hook.go
```diff
@@ -60,9 +71,19 @@ type Hook struct {
...
 }
+
+// A HookExecution records the result for the last execution of a hook for a given release.
+type HookExecution struct {
+	// StartedAt indicates the date/time this hook was started
+	StartedAt time.Time `json:"started_at,omitempty"`
+	// CompletedAt indicates the date/time this hook was completed
+	CompletedAt time.Time `json:"completed_at,omitempty"`
+	// Successful indicates whether the hook completed successfully
+	Successful bool `json:"successful"`
```
bacongobbler: Should this be a hook phase instead, similar to pod phases? If I'm checking on the status of a hook execution event, how can I infer that it's in an unknown state instead of a pending state?

Perhaps this is related to my earlier comment on `waitForPodSuccess`.

There is also a TestRun struct in `pkg/release` that includes an `Info` and a `Status` object. It appears similar to this object's intended use. Status can be one of "success", "failed", "running" or "unknown". Perhaps that struct and this one could be consolidated.

bacongobbler (changes requested):

pkg/kube/client.go
jlegrone: Unless we want to print a message to the user to let them know that an unknown pod status was encountered, then no. Currently if an unknown pod status is encountered then we just continue watching until `PodSucceeded` or `PodFailed` is observed.

pkg/release/hook.go
jlegrone: This is actually to ensure backwards compatibility. 😄 

Confusingly, `pkg/hooks` had its own definitions for deletion policies, which are refactored out in this PR:

https://github.com/helm/helm/blob/539bafa58b6ecb68f5a387a6fda0708d31723ab5/pkg/hooks/hooks.go#L46-L51

pkg/release/hook.go
jlegrone: Thanks, I missed `TestRun` earlier. Refactored in fada0c92f8ed47964dced8a639f841a8afaff931.

pkg/action/hooks.go
```diff
@@ -0,0 +1,122 @@
...
+			return errors.Wrapf(err, "unable to build kubernetes object for %s hook %s", hook, h.Path)
+		}
+		if _, err := cfg.KubeClient.Create(resources); err != nil {
+			return errors.Wrapf(err, "warning: Hook %s %s failed", hook, h.Path)
+		}
+
+		// Get the time at which the hook was applied to the cluster
+		h.LastRun = release.HookExecution{
+			StartedAt: time.Now(),
+			Phase:     release.HookPhaseUnknown,
```
bacongobbler: can we introduce a new phase: `release.HookPhaseRunning` and use that here instead? We know it's in a running state at this time. In Kubernetes, a resource in an unknown state is used to inform the user that the resource has entered an erroneous state (i.e. Helm was unable to retrieve the hook's completion state), and the user should intervene to determine what should be done to reconcile its state.

bacongobbler (changes requested):

bacongobbler:
ping. Looks like there's one comment left to address: https://github.com/helm/helm/pull/6054#discussion_r311789258

We can address `--cleanup` in a followup.

pkg/action/hooks.go
jlegrone: `HookPhaseUnknown` was used here rather than `HookPhaseRunning` because
1. No intermediate state about the release is currently persisted during execution.
1. As long as the implementation of `WatchUntilReady` does not panic `HookPhaseFailed` or `HookPhaseSucceeded` should always be set (and if there is a panic, then `HookPhaseUnknown` is the most appropriate value to surface).

0a11574b adds `HookPhaseRunning` and persists intermediate state. Is the extra API call acceptable?

bacongobbler (approved):

PR #6056, opened 2019-07-21
chore(deps): bump kubernetes to v1.15

adamreese:


bacongobbler (approved):

Gopkg.toml
```diff
@@ -28,19 +28,19 @@
...
 
 [[constraint]]
   name = "k8s.io/apimachinery"
-  branch = "release-1.14"
+  branch = "release-1.15"
 
 [[constraint]]
   name = "k8s.io/client-go"
-  version = "kubernetes-1.14.1"
+  version = "kubernetes-1.15.1"
```
hickeyma: Should this be `kubernetes-1.15.1` or `kubernetes-1.15.0`? If 1.15.1 then should all other k8s references be .1 as well? Should it show the patch version for all k8s references?
Also, should `version` be changed to `branch` like the rest of the k8s references?

hickeyma (changes requested):
@adamreese Just have a few questions before I OK.

Gopkg.toml
bacongobbler: it should always be the latest patch release, so 1.15.1. The `release-1.15` branch tracks the latest patch release for other projects. client-go is the odd one out here that doesn't have a `release-1.15` branch, so the `kubernetes-1.15.1` tag is used instead.

hickeyma: ok, thanks for the calrification.

hickeyma (commented):
LGTM, thanks.

PR #6059, opened 2019-07-22
Helm 3: Use chart version as default tag when saving

### LINKED ISSUE #5752: Helm 3-alpha1: pushing chart without tag specified results in untagged chart instead of latest
Output of `helm version`:
`version.BuildInfo{Version:"v3.0.0-alpha.1", GitCommit:"b9a54967f838723fe241172a6b94d18caf8bcdca", GitTreeState:"clean"}`

Output of `kubectl version`:
`Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.1", GitCommit:"b7394102d6ef778017f2ca4046abbaa23b88c290", GitTreeState:"clean", BuildDate:"2019-04-08T17:11:31Z", GoVersion:"go1.12.1", Compiler:"gc", Platform:"windows/amd64"}`

Azure Container Registry (but should be reproducable on any OCI compatible registry)

Repro:
- `helm create test`
- `helm chart save test myreg.azurecr.io/test` (replace registry url)
- `helm chart push myreg.azurecr.io/test`

Output `helm chart push` command:
```
Using default tag: latest
The push refers to repository [myreg.azurecr.io/test]
Name: test
Version: 0.1.0
Meta: sha256:xxx
Content: sha256:xxx
latest: pushed to remote (2 layers, 2.3 KiB total)
```

Expected: 
The output indicates the the default tag (`latest`) is used, so it's expected that the remote registry now contains a `test` repository with a manifest tagged `latest`.

Actual:
The remote registry now contains a `test` repository with a single manifest, but its untagged (or at least not tagged latest).

jdolitsky:
Fixes #5752 

This change essentially always enforces a tag, by using a chart's metadata version

For example:
```
$ cat harbor/Chart.yaml | grep version:
version: 1.1.1
```

```
$ bin/helm chart save harbor/ mysite.io/mylib/harbor
Name: harbor
Version: 1.1.1
Meta: sha256:5666f9e9f89e109a28af426e8947b00fbf22592b0ba79d61d460784744da6f19
Content: sha256:d75af1601a514fe2168b59069b5aec6ea4c73b7792c029adebd79407544eb13a
1.1.1: saved
```

```
$ bin/helm chart list
REF                         	NAME  	VERSION	DIGEST 	SIZE    	CREATED
mysite.io/mylib/harbor:1.1.1	harbor	1.1.1  	d75af16	28.2 KiB	3 seconds
```

cc @nphmuller @muehlburger @bacongobbler @hickeyma

### DIFF
--- a/pkg/registry/reference.go
+++ b/pkg/registry/reference.go
@@ -22,6 +22,8 @@ import (
 	"strings"
 
 	"github.com/containerd/containerd/reference"
+
+	"helm.sh/helm/pkg/chart"
 )
 
 var (
@@ -57,6 +59,22 @@ func ParseReference(s string) (*Reference, error) {
 	}
 
 	return &ref, nil
+}
+
+// ParseReferenceWithChartDefaults converts a string to a Reference,
+// using values from a given chart as defaults
+func ParseReferenceWithChartDefaults(s string, ch *chart.Chart) (*Reference, error) {
+	ref, err := ParseReference(s)
+	if err != nil {
+		return nil, err
+	}
+
+	// If no tag is present, use the chart version
+	if ref.Tag == "" {
+		ref.Tag = ch.Metadata.Version
+	}
+
+	return ref, nil
 }
 
 // setExtraFields adds the Repo and Tag fields to a Reference
--- a/pkg/action/chart_save.go
+++ b/pkg/action/chart_save.go
@@ -48,10 +48,9 @@ func (a *ChartSave) Run(out io.Writer, path, ref string) error {
 		return err
 	}
 
-	r, err := registry.ParseReference(ref)
+	r, err := registry.ParseReferenceWithChartDefaults(ref, ch)
 	if err != nil {
 		return err
 	}
-
 	return a.cfg.RegistryClient.SaveChart(ch, r)
 }
--- a/pkg/registry/cache.go
+++ b/pkg/registry/cache.go
@@ -142,7 +142,7 @@ func (cache *filesystemCache) ChartToLayers(ch *chart.Chart) ([]ocispec.Descript
 }
 
 func (cache *filesystemCache) LoadReference(ref *Reference) ([]ocispec.Descriptor, error) {
-	tagDir := filepath.Join(cache.rootDir, "refs", escape(ref.Repo), "tags", tagOrDefault(ref.Tag))
+	tagDir := filepath.Join(cache.rootDir, "refs", escape(ref.Repo), "tags", ref.Tag)
 
 	// add meta layer
 	metaJSONRaw, err := getSymlinkDestContent(filepath.Join(tagDir, "meta"))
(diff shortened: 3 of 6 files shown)


pkg/registry/reference.go
```diff
@@ -59,6 +61,22 @@ func ParseReference(s string) (*Reference, error) {
...
 
+// ParseReferenceWithChartDefaults converts a string to a Reference,
+// using values from a given chart as defaults
+func ParseReferenceWithChartDefaults(s string, ch *chart.Chart) (*Reference, error) {
+	ref, err := ParseReference(s)
+	if err != nil {
+		return nil, err
+	}
+
+	// If no tag is present, use the chart version
```
bacongobbler: instead of introducing a new function, can we just add this conditional logic to pkg/action/chart_save.go? e.g.

```
r, err := registry.ParseReference(ref)
	r, err := registry.ParseReferenceWithChartDefaults(ref, ch)
	if err != nil {
		return err
	}

	if r.Tag == "" {
		r.Tag = ch.Metadata.Version
	}

	...
```

bacongobbler (changes requested):

pkg/registry/reference.go
jdolitsky: I originally had done this, but decided against it for a few reasons:

a.) it wasn't a change that was easily tested via action package as the method had no returns besides "err" - unless we run a test against the stdout, which seems a little bleh. This new method is easy to verify via tests

b.) it occurred to me that we may want to add more transform features based on the chart, for example `myreg.io/mylib/@name` might replace "@name" with chart name? Something along these lines

c.) we may want to get this functionality somewhere outside of the save method

@bacongobbler what do you think? I'm happy to revert - but need some ideas on how to correctly unit test this if so

bacongobbler: taking a closer deep dive into the code now :eyes:

bacongobbler:
@jdolitsky here's a patch built on top of this PR that implements the changes suggested. Let me know if you'd like me to PR this.


[0001-refactoring-stuff.patch.txt](https://github.com/helm/helm/files/3424092/0001-refactoring-stuff.patch.txt)

jdolitsky:
@bacongobbler yes that looks good - go ahead and submit it.

I would say though that this technically does not prove the functionality:
```
+	// now let's check if `helm chart save` can use the chart version when the tag is not present
+	if err := action.Run(ioutil.Discard, input, "localhost:5000/test"); err != nil {
+		t.Error(err)
+	}
```

This would pass even without the `if r.Tag == ""` change (hence the bug)

bacongobbler:
That's what the call to `.LoadChart` checks for - if that patch wasn't present then the chart would be saved with the reference `localhost:5000/test:latest`. Commenting out the tag check in chart_save.go results in a failing test:

```
Running tool: /usr/local/bin/go test -timeout 30s helm.sh/helm/pkg/action -run ^(TestChartSave)$

--- FAIL: TestChartSave (0.00s)
    /home/bacongobbler/code/go/src/helm.sh/helm/pkg/action/chart_save_test.go:61: readlink /tmp/helm-action-test411517322/refs/localhost_5000/test/tags/0.1.0/meta: no such file or directory
FAIL
FAIL	helm.sh/helm/pkg/action	0.026s
Error: Tests failed.
```

Restoring the tag check in chart_save.go makes the tests pass again because the chart was saved under `localhost:5000/test:0.1.0`.

```
Running tool: /usr/local/bin/go test -timeout 30s helm.sh/helm/pkg/action -run ^(TestChartSave)$

ok  	helm.sh/helm/pkg/action	0.026s
Success: Tests passed.
```

bacongobbler:
closing in favour of #6070. Thanks @jdolitsky!

PR #6060, opened 2019-07-22
fix repo url being decoded

### LINKED ISSUE #6048: helm should not urldecode urls
helm --debug repo add t7 "http://localhost/some%2Fpath/test"
Error: Looks like "http://localhost/some%2Fpath/test" is not a valid chart repository or cannot be reached: Failed to fetch http://localhost/some/path/test/index.yaml : 404 Not Found

notice how  "some%2Fpath" gets translated into "some/path"

This is not expected. It is expected that helm leaves the url alone and that "some%2Fpath" gets transferred to the server.

Output of `helm version`:
Client: &version.Version{SemVer:"v2.11.0", GitCommit:"2e55dbe1fdb5fdb96b75ff144a339489417b146b", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.11.0", GitCommit:"2e55dbe1fdb5fdb96b75ff144a339489417b146b", GitTreeState:"clean"}

karuppiah7890:
Fixes #6048

To test better - I added mocks, and a memory based filesystem so that nothing gets written to the actual file system

**If applicable**:

### DIFF
--- a/pkg/repo/chartrepo.go
+++ b/pkg/repo/chartrepo.go
@@ -21,6 +21,7 @@ import (
 	"io/ioutil"
 	"net/url"
 	"os"
+	"path"
 	"path/filepath"
 	"strings"
 
@@ -111,14 +112,13 @@ func (r *ChartRepository) Load() error {
 // cachePath is prepended to any index that does not have an absolute path. This
 // is for pre-2.2.0 repo files.
 func (r *ChartRepository) DownloadIndexFile(cachePath string) error {
-	var indexURL string
 	parsedURL, err := url.Parse(r.Config.URL)
 	if err != nil {
 		return err
 	}
-	parsedURL.Path = strings.TrimSuffix(parsedURL.Path, "/") + "/index.yaml"
-
-	indexURL = parsedURL.String()
+	parsedURL.RawPath = path.Join(parsedURL.RawPath, "index.yaml")
+	parsedURL.Path = path.Join(parsedURL.Path, "index.yaml")
+	indexURL := parsedURL.String()
 
 	r.setCredentials()
 	resp, err := r.Client.Get(indexURL)
--- a/pkg/repo/index_test.go
+++ b/pkg/repo/index_test.go
@@ -18,6 +18,7 @@ package repo
 
 import (
 	"io/ioutil"
+	"net/http"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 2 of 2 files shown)


karuppiah7890:
If this looks good, I can port the code to dev-v3 branch too!

karuppiah7890:
Not sure why it fails in the CI. `releasetesting` package tests pass in my local. Weird

bacongobbler:
resource limitations in Circle sometimes causes CI to fail.

pkg/repo/chartrepo.go
```diff
@@ -110,15 +112,14 @@ func (r *ChartRepository) Load() error {
 //
 // cachePath is prepended to any index that does not have an absolute path. This
 // is for pre-2.2.0 repo files.
-func (r *ChartRepository) DownloadIndexFile(cachePath string) error {
-	var indexURL string
+func (r *ChartRepository) DownloadIndexFile(cachePath string, fs afero.Fs) error {
```
mattfarina: This changes the public function signature in a non-backwards compatible manner. A breaking change, like this, would need to be on the v3 branch or the change would need to happen in a manner than doesn't break function backwards compatibility.

mattfarina (changes requested):
On Helm we follow Semantic Versioning including for the Go functions. There is a breaking change in this PR which means it cannot go into v2 as it stands.

pkg/repo/chartrepo.go
karuppiah7890: @mattfarina Ah. I overlooked that! It's a breaking change, yes. What do you think about having fs a field in ChartRepository? Like the `Client` Would it make sense? And I'll also test it for backwards compatibility by making sure that if fs is not present, it will be initialized. 

And the reason for putting fs is - for dependency injection to make testing easier. This way, it's possible to inject a in-memory fs for testing and os fs for actual execution

karuppiah7890: @mattfarina I have made changes in such a way that it doesn't break the API and is backwards compatible. Could you check now?

karuppiah7890:
@mattfarina Made the requested changes. Used an `fs` variable in the field of `ChartRepository` struct. Could you review?

karuppiah7890:
Just pinging in case anyone has time to review this cc @mattfarina @bacongobbler

karuppiah7890:
cc @thomastaylor312

karuppiah7890:
Still looking for comments on the PR 😅  cc @mattfarina @bacongobbler

karuppiah7890:
cc @hickeyma . Waiting for PR review 😅

karuppiah7890:
Also updated `pflag` version to v1.0.5 https://github.com/spf13/pflag/tree/v1.0.5

karuppiah7890:
@mattfarina I have rebased with master 😄

pkg/repo/chartrepo_test.go
```diff
@@ -319,3 +324,50 @@ func TestResolveReferenceURL(t *testing.T) {
...
+
+		fs := afero.NewMemMapFs()
+		r.Fs = fs
+
+		err = fs.MkdirAll("/repository/cache/", 0644)
+		assert.NoError(t, err, "problem creating test repository cache")
+
+		mockGetter := mocks.NewMockGetter(ctrl)
+		r.Client = mockGetter
+		indexYamlContent := `apiVersion: v1
```
mattfarina: While I appreciate the testing the style of this code and the use of `github.com/spf13/afero` is different from the way the rest of the Helm code base handles things. The typical situation is to use a temp directory (search for `ensure.TempDir` to find examples) and the repo package has a method for creating and working with a test server.

Would it be possible to use the other setups found in Helm for consistency? Or, is there a reason to break from that.

karuppiah7890: I did notice the temp dir style testing recently in some places. Will try to do the same 👍 and remove tmp for once test is over

I do want to mention some pros for afero though - it makes testing very easy - no need to delete files, you can use in-memory fs and write to anywhere and not just tmp directories

bacongobbler: I would suggest posing that as a follow-up PR to enhance all unit tests. Same with golang/mock... Up to now, we've been using `net/http/httptest` for most tests with chart repositories. For example: https://github.com/helm/helm/blob/7cad59091a9451b2aa4f95aa882ea27e6b195f98/pkg/repo/chartrepo_test.go#L241

karuppiah7890: Sure 👍 I will raise issue to discuss the methods / strategies to test and then raise separate PRs for them

bacongobbler:
Hi @karuppiah7890. Me again. :)

Given the low volume of PRs left on the queue, we are planning to cut a 2.15 within the next few business days. Any chance you can resolve the comments by EOD tomorrow? If not, we may have to punt this over to the next release. Thanks!

karuppiah7890:
I'll try my best! Will start looking into this today! :)

karuppiah7890:
@bacongobbler I made the changes. But I noticed there's a data race condition failure in the tests. It's present in master too. Not sure if it was caught in circle ci or not. I found it in my local. Here are the logs for all the tests in master and the particular test that revealed race condition. It's this test

[test.log](https://github.com/helm/helm/files/3700669/test.log)
[master-branch-tests.log](https://github.com/helm/helm/files/3700673/master.log)

https://github.com/helm/helm/blob/276bb9b1c8ee3c51064f768e8c0e2fb64c3fd5f2/pkg/kube/client_test.go#L460

karuppiah7890:
@bacongobbler I just noticed that circleci runs the tests without race flag enabled. Is that intentional? In any case, we'll be missing race condition issues if we don't run it with `-race` enabled in CI

bacongobbler (approved):

bacongobbler:
> But I noticed there's a data race condition failure in the tests.

would you mind opening a new issue for that?

karuppiah7890:
Sure!

hickeyma (approved):
LGTM, thanks @karuppiah7890.

hickeyma:
Merging this as the race condition in the tests are happening in master at this moment, so prior to this PR.

hickeyma:
Opened issue https://github.com/helm/helm/issues/6617 for the data race condition failure in the tests.

hickeyma:
v3 port: https://github.com/helm/helm/pull/6602

PR #6066, opened 2019-07-23
fix(engine): Fix eating too many colons during template execution

### LINKED ISSUE #6044: `required` function swallows message until first colon (`:`)
Output of `helm version`:

```
version.BuildInfo{Version:"v3.0.0-alpha.1", GitCommit:"b9a54967f838723fe241172a6b94d18caf8bcdca", GitTreeState:"clean"}
```

Example issue (note escaping of `$` is for the shell):

```
export PATH="$HOME/bin/helm-v3.0.0-alpha.1-darwin-amd64:$PATH"
cd /tmp
rm -rf colon-missing
helm create colon-missing
cat >colon-missing/templates/deployment.yaml <<EOF
{{ \$someEmptyValue := "" }}
{{ \$myvar := "abc" }}
{{- required (printf "%s: something is missing" \$myvar) \$someEmptyValue | repeat 0 }}
EOF
helm template myrelease colon-missing/
```

This will fail with a truncated error message:

```
Error: render error at (colon-missing/templates/deployment.yaml:3:4): something is missing
```

Once you remove the `:` from the custom error message (`printf "%s <colon removed here> something is missing" $myvar`), the error output is as expected:

```
Error: render error at (colon-missing/templates/deployment.yaml:3:4): abc <colon removed here> something is missing
```

ian-howell:
This fixes #6044, in which error parsing is greedily eating too many
colons, preventing users from using colons in their warning messages to
the `required` function

Signed-off-by: Ian Howell <ian.howell0@gmail.com>

Closes #6044 

I've arbitrarily decided to use the regex `HELM\[.*\]HELM`. This should probably be changed to something more unique, but works for a POC.

It should also be noted that the `parseTemplateError` has been split into two functions: one for parse errors, the other for execution errors. This lines up with the current usage of the `text/template` package.

**If applicable**:

### DIFF
--- a/pkg/engine/engine_test.go
+++ b/pkg/engine/engine_test.go
@@ -187,7 +187,23 @@ func TestParallelRenderInternals(t *testing.T) {
 	wg.Wait()
 }
 
-func TestRenderErrors(t *testing.T) {
+func TestParseErrors(t *testing.T) {
+	vals := chartutil.Values{"Values": map[string]interface{}{}}
+
+	tplsUndefinedFunction := map[string]renderable{
+		"undefined_function": {tpl: `{{foo}}`, vals: vals},
+	}
+	_, err := new(Engine).render(tplsUndefinedFunction)
+	if err == nil {
+		t.Fatalf("Expected failures while rendering: %s", err)
+	}
+	expected := `parse error at (undefined_function:1): function "foo" not defined`
+	if err.Error() != expected {
+		t.Errorf("Expected '%s', got %q", expected, err.Error())
+	}
+}
+
+func TestExecErrors(t *testing.T) {
 	vals := chartutil.Values{"Values": map[string]interface{}{}}
 
 	tplsMissingRequired := map[string]renderable{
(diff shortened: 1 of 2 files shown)


pkg/engine/engine_test.go
```diff
@@ -187,7 +187,23 @@ func TestParallelRenderInternals(t *testing.T) {
...
-func TestRenderErrors(t *testing.T) {
+func TestParseErrors(t *testing.T) {
+	vals := chartutil.Values{"Values": map[string]interface{}{}}
+
+	tplsUndefinedFunction := map[string]renderable{
+		"undefined_function": {tpl: `{{foo}}`, vals: vals},
+	}
+	_, err := new(Engine).render(tplsUndefinedFunction)
+	if err == nil {
+		t.Fatalf("Expected failures while rendering: %s", err)
```
thomastaylor312: `err` will always be `nil` in this case, so I would just change this to `t.Fatal`

thomastaylor312: `err` will always be nil here, so I would just use `t.Fatal`

pkg/engine/engine_test.go
```diff
@@ -197,23 +213,39 @@ func TestRenderErrors(t *testing.T) {
...
 	}
 
-	tplsUndefinedFunction := map[string]renderable{
-		"undefined_function": {tpl: `{{foo}}`, vals: vals},
+	tplsMissingRequired = map[string]renderable{
+		"missing_required_with_colons": {tpl: `{{required ":this: message: has many: colons:" .Values.foo}}`, vals: vals},
+	}
+	_, err = new(Engine).render(tplsMissingRequired)
+	if err == nil {
+		t.Fatalf("Expected failures while rendering: %s", err)
```
thomastaylor312: Same here with `t.Fatal`

thomastaylor312 (approved):
Two small nits, but otherwise good to go. Thanks for the great test coverage on this!

pkg/engine/engine_test.go
ian-howell: I'm not sure I understand. If `err` is nil, that means the `render` function didn't work as expected.

thomastaylor312:
One more thing, I think the arbitrary string wrapping here will work (although it isn't really pretty, but I can't think of something better either 😆), but I think there might be a _slight_ chance that someone might use the string `HELM` in their error message. Maybe `HELMERR` instead? I'll hold off on merging until you get a chance to respond

ian-howell:
I went ahead and chose a more unique delimiter for the wrapper - it's also `const`-ed now to make it easier to change if it's ever needed.

pkg/engine/engine_test.go
thomastaylor312: This was just noting that you don't need to use `Fatalf` and pass an `err` because it doesn't provide any additional info. e.g. `t.Fatal("Expected failures while rendering")`. But it is just a nit anyway 🙂

thomastaylor312:
Awesome great! Thanks again for the fix!

Issue #5184, opened 2019-01-18
Nested `null` values don't remove keys as expected


aeijdenberg:
<https://github.com/helm/helm/pull/2648> allows for deletion of keys by setting `null` values in a `values.yaml` file. However it doesn't work for nested values, e.g.:

```yaml
web:
  livenessProbe:
    httpGet: null
    exec:
      command:
      - curl
      - -f
      - http://localhost:8080/api/v1/info
```

Will not remove `web.livenessProbe.httpGet` from the original values, rather it just overrides the value with `null` and prints the warning:

```
2019/01/18 11:30:07 Warning: Merging destination map for chart 'concourse'. Cannot overwrite table item 'httpGet', with non table value: map[path:/api/v1/info port:atc]
```

Ironically, above is a small variant on the example given in the docs, which I'm pretty sure does not actually do what it claims to: <https://docs.helm.sh/chart_template_guide/#deleting-a-default-key>

I suspect that since the template rendering doesn't seem to differentiate much between a null value and a value not being specified, that this warnings are being ignored and not having much effect.

Output of `helm version`:

```
Client: &version.Version{SemVer:"v2.12.1", GitCommit:"02a47c7249b1fc6d8fd3b94e6b4babf9d818144e", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.12.1", GitCommit:"02a47c7249b1fc6d8fd3b94e6b4babf9d818144e", GitTreeState:"clean"}
```

Output of `kubectl version`:

```
Client Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.1", GitCommit:"eec55b9ba98609a46fee712359c7b5b365bdd920", GitTreeState:"clean", BuildDate:"2018-12-13T10:39:04Z", GoVersion:"go1.11.2", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"11+", GitVersion:"v1.11.5-eks-6bad6d", GitCommit:"6bad6d9c768dc0864dab48a11653aa53b5a47043", GitTreeState:"clean", BuildDate:"2018-12-06T23:13:14Z", GoVersion:"go1.10.3", Compiler:"gc", Platform:"linux/amd64"}
```

Cloud Provider/Platform (AKS, GKE, Minikube etc.): 

EKS

bacongobbler:
@scottrigby, do you by any chance want to take a gander at this one?

bacongobbler:
[...]

PR #6080, opened 2019-07-25
Fix subchart value deletion

### LINKED ISSUE #3804: Delete sub-chart value key from parent-chart
It seems that it is not possible to delete value key of sub-chart from parent-chart.

```
./values.yaml
name: parent
test.resources: null
```

```
./charts/test/value.yaml
name: child
resurces:
  requests:
    cpu: 500m
    memory: 2Gi
```

I would expect that after the merge the result will be:
```
./charts/test/value.yaml
name: child
```

Client: &version.Version{SemVer:"v2.8.2", GitCommit:"a80231648a1473929271764b920a8e346f6de844", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.8.2", GitCommit:"a80231648a1473929271764b920a8e346f6de844", GitTreeState:"clean"}

Client Version: version.Info{Major:"1", Minor:"9", GitVersion:"v1.9.4", GitCommit:"bee2d1505c4fe820744d26d41ecd3fdd4a3d6546", GitTreeState:"clean", BuildDate:"2018-03-12T16:29:47Z", GoVersion:"go1.9.3", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"8+", GitVersion:"v1.8.8-gke.0", GitCommit:"6e5b33a290a99c067003632e0fd6be0ead48b233", GitTreeState:"clean", BuildDate:"2018-02-16T18:26:58Z", GoVersion:"go1.8.3b4", Compiler:"gc", Platform:"linux/amd64"}

GKE

deitmerit:
Overwriting a subchart's default value with nil in the parent chart's values.yaml should delete the value (see https://github.com/helm/helm/blob/master/docs/chart_template_guide/values_files.md#deleting-a-default-key).

closes #3804

This is my first PR for the helm project, so please let me know if it is missing anything.

The problem seemed to be that when called with a non-nil chart.Config argument, the chartutil.CoalesceValues function called coalesceDeps twice (once via coalesce and once directly). 
The first call removed the subchart's value as expected, but the second call then re-introduced the value, because the nil value from the parent's values is gone.
Roughly like this:
coalesceDeps(values: {subchart.value=nil}, subchart values: {value=foo}) => result: {}
coalesceDeps(values: {}, subchart values: {value=foo}) => result: {subchart.value=foo})

I added a very minimalistic unit test and re-used the moby chart and one of its subcharts for this. Not sure if this should be expanded to include more use-cases.

**If applicable**:

### DIFF
--- a/pkg/chartutil/testdata/moby/charts/spouter/values.yaml
+++ b/pkg/chartutil/testdata/moby/charts/spouter/values.yaml
@@ -1,1 +1,2 @@
 scope: spouter
+foo: bar
--- a/pkg/chartutil/testdata/moby/values.yaml
+++ b/pkg/chartutil/testdata/moby/values.yaml
@@ -22,3 +22,7 @@ web:
     httpGet:
       path: /api/v1/info
       port: atc
+
+# for testing deleting default values in sub charts
+spouter:
+  foo: null
--- a/pkg/chartutil/values.go
+++ b/pkg/chartutil/values.go
@@ -170,10 +170,7 @@ func CoalesceValues(chrt *chart.Chart, vals *chart.Config) (Values, error) {
 		if err != nil {
 			return cvals, err
 		}
-		cvals, err = coalesce(chrt, evals)
-		if err != nil {
-			return cvals, err
-		}
+		return coalesce(chrt, evals)
 	}
 
 	return coalesceDeps(chrt, cvals)
--- a/pkg/chartutil/values_test.go
+++ b/pkg/chartutil/values_test.go
@@ -475,6 +475,33 @@ func TestCoalesceTables(t *testing.T) {
 		t.Errorf("Expected boat string, got %v", dst["boat"])
 	}
 }
+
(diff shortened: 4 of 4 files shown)


deitmerit:
The unit test that fails in circleci is green on my machine :(

bacongobbler:
Circle has been having some performance degradation issues as of late. Rebooting the job should let the tests pass. I've rebooted the job.

thomastaylor312 (approved):
LGTM. I did some manual validation as well to make sure the normal value deletion still worked as well for non-subchart values

bacongobbler:
Removing the v3 port label. This was fixed in Helm 3 during a larger refactor via https://github.com/helm/helm/pull/5781.

arnaud-deprez:
Dear,

According to my comment https://github.com/helm/helm/issues/5184#issuecomment-544525773, it still does not work with `v2.15.0`.

PR #6085, opened 2019-07-25
ref(*): Refactors kube client to be a bit more friendly

thomastaylor312:
This changes most of the KubeClient interface to only ever build objects once and then pass in everything as lists of resources. As a consequence, we needed to refactor several of the actions. I took the opportunity to refactor out some duplicated code while I was in the same area

### DIFF
--- a/pkg/action/uninstall.go
+++ b/pkg/action/uninstall.go
@@ -175,7 +175,7 @@ func (u *Uninstall) deleteRelease(rel *release.Release) (string, []error) {
 
 	var builder strings.Builder
 	for _, file := range filesToDelete {
-		builder.WriteString(file.Content)
+		builder.WriteString("\n---\n" + file.Content)
 	}
 	resources, err := u.cfg.KubeClient.Build(strings.NewReader(builder.String()))
 	if err != nil {
--- a/pkg/action/uninstall.go
+++ b/pkg/action/uninstall.go
@@ -17,15 +17,12 @@ limitations under the License.
 package action
 
 import (
-	"bytes"
-	"sort"
 	"strings"
 	"time"
 
 	"github.com/pkg/errors"
 
 	"helm.sh/helm/pkg/hooks"
-	"helm.sh/helm/pkg/kube"
 	"helm.sh/helm/pkg/release"
 	"helm.sh/helm/pkg/releaseutil"
 )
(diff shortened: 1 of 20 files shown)


thomastaylor312:
Please note I am still manually validating, but all tests pass right now

pkg/kube/interface.go
```diff
@@ -21,46 +21,44 @@ import (
...
-	BuildUnstructured(reader io.Reader) (Result, error)
+	// by "\n---\n")
+	Build(reader io.Reader) (ResourceList, error)
 
 	// WaitAndGetCompletedPodPhase waits up to a timeout until a pod enters a completed phase
 	// and returns said phase (PodSucceeded or PodFailed qualify).
 	WaitAndGetCompletedPodPhase(name string, timeout time.Duration) (v1.PodPhase, error)
+
+	// KubernetesClientSet returns the underlying kubernetes clientset
+	KubernetesClientSet() (kubernetes.Interface, error)
```
thomastaylor312: Wanted to call this out specifically. We need a way to fetch out the underlying client so the actions can perform additional logic (like recreate) without needing to reconstruct the client, but this may be too much to have on the interface

pkg/action/uninstall.go
```diff
@@ -227,23 +168,20 @@ func (u *Uninstall) deleteRelease(rel *release.Release) (kept string, errs []err
...
-		}
-		if err := u.cfg.KubeClient.Delete(b); err != nil {
-			u.cfg.Log("uninstall: Failed deletion of %q: %s", rel.Name, err)
-			if err == kube.ErrNoObjectsVisited {
-				// Rewrite the message from "no objects visited"
-				err = errors.New("object not found, skipping delete")
-			}
-			errs = append(errs, err)
-		}
+		builder.WriteString("\n---\n" + file.Content)
```
thomastaylor312: One more note here. If we do not like glueing manifests back together like this, there are 2 alternatives:
1. Use `KubeClient.Build` for each manifest and then put those objects together into one slice
2. Refactor the filtering logic to filter a `ResourceList` instead of a raw manifest

bacongobbler:
Do you know why Gopkg.lock was updated in this PR? Seems like all of the digests were removed for calculating the lockfile... Which version of dep are you using?

thomastaylor312:
I had updated a dependency to a different pin and then found out I didn't need it, I assumed it had just recalculated some of the deps. My current version is 0.5.4

pkg/kube/interface.go
thomastaylor312: We decided to move this to the `Configuration` object

thomastaylor312: We decided to move this to the `Configuration` object

thomastaylor312:
This has been tested and ready for a full review

bacongobbler (approved):
this may need one more maintainer to make sure this looks good, but after pairing with @thomastaylor312 on this, LGTM!

thomastaylor312:
Based on my conversations with @bacongobbler and the rules around merging PRs, we have tested this enough to feel comfortable merging.

hickeyma (commented):
Sorry for being late on the review. Very nice refactor @thomastaylor312. LGTM.

PR #6092, opened 2019-07-26
fix: upgrade charts with CRD changes

### LINKED ISSUE #5853: can't upgrade charts with crd changes
I have a custom chart for an operator we're writing.

The operator is under development, so it receives several changes... now, we needed to add a new non-required boolean field to the CRD.

When trying to upgrade it, I got this error:

```
UPGRADE FAILED
Error: failed to create patch: merging an object in json but data type is not struct, instead is: map
Error: UPGRADE FAILED: failed to create patch: merging an object in json but data type is not struct, instead is: map
```

The only things changed are the new field on the CRD and the image tag.

If I remove the new field from the CRD and upgrade it works.

What can I do to avoid this?

---
Output of `helm version`:

```
Client: &version.Version{SemVer:"v2.14.0", GitCommit:"05811b84a3f93603dd6c2fcfe57944dfa7ab7fd0", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.14.0", GitCommit:"05811b84a3f93603dd6c2fcfe57944dfa7ab7fd0", GitTreeState:"clean"}
```

Output of `kubectl version`:

```
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.2", GitCommit:"66049e3b21efe110454d67df4fa62b08ea79a19b", GitTreeState:"clean", BuildDate:"2019-05-16T18:55:03Z", GoVersion:"go1.12.5", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"13+", GitVersion:"v1.13.6-gke.6", GitCommit:"fcbc1d20b6bca1936c0317743055ac75aef608ce", GitTreeState:"clean", BuildDate:"2019-05-23T11:42:40Z", GoVersion:"go1.11.5b4", Compiler:"gc", Platform:"linux/amd64"}
```

GKE

mumoshu:

Fixes #5853

In other words, this fixes (I believe) a critial bug that causing Helm to be unable to update CRDs

See the discussion from https://github.com/helm/helm/issues/5853#issuecomment-500187718 for how this happened and why this fixes it.

To test it yourself, grab and extract [two chart archives for reproduction](https://github.com/helm/helm/issues/5853#issuecomment-500187718), and follow the steps below:

```
$ kind create cluster
$ make docker-image
$ kind load docker-image gcr.io/kubernetes-helm/tiller:git-<COMMIT_ID>
$ helm init --tiller-image gcr.io/kubernetes-helm/tiller:git-<COMMIT_ID> --upgrade
$ helm upgrade --install foo ./issue-5853-chart-old
$ helm upgrade --install foo ./issue-5853-chart-old

# Or

$ kind create cluster
$ go run ./cmd/tiller
# In another terminal:
$ helm --host localhost:44134 upgrade --install foo ./issue-5853-chart-old
$ helm --host localhost:44134 upgrade --install foo ./issue-5853-chart-new
```

**If applicable**:

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -656,7 +656,7 @@ func createPatch(target *resource.Info, current runtime.Object) ([]byte, types.P
 	}
 
 	// Get a versioned object
-	versionedObject := asVersioned(target)
+	versionedObject, err := asVersioned(target)
 
 	// Unstructured objects, such as CRDs, may not have an not registered error
 	// returned from ConvertToVersion. Anything that's unstructured should
@@ -664,16 +664,25 @@ func createPatch(target *resource.Info, current runtime.Object) ([]byte, types.P
 	// on objects like CRDs.
 	_, isUnstructured := versionedObject.(runtime.Unstructured)
 
+	// On newer K8s versions, CRDs aren't unstructured but has this dedicated type
+	_, isCRD := versionedObject.(*apiextv1beta1.CustomResourceDefinition)
+
 	switch {
-	case runtime.IsNotRegisteredError(err), isUnstructured:
+	case runtime.IsNotRegisteredError(err), isUnstructured, isCRD:
 		// fall back to generic JSON merge patch
 		patch, err := jsonpatch.CreateMergePatch(oldData, newData)
-		return patch, types.MergePatchType, err
+		if err != nil {
+			return nil, types.MergePatchType, fmt.Errorf("failed to create merge patch: %v", err)
+		}
+		return patch, types.MergePatchType, nil
 	case err != nil:
 		return nil, types.StrategicMergePatchType, fmt.Errorf("failed to get versionedObject: %s", err)
 	default:
 		patch, err := strategicpatch.CreateTwoWayMergePatch(oldData, newData, versionedObject)
-		return patch, types.StrategicMergePatchType, err
+		if err != nil {
+			return nil, types.StrategicMergePatchType, fmt.Errorf("failed to create two-way merge patch: %v", err)
+		}
+		return patch, types.StrategicMergePatchType, nil
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/kube/client.go
```diff
@@ -977,17 +986,23 @@ func isFoundPod(podItem []v1.Pod, pod v1.Pod) bool {
 	return false
 }
 
-func asVersioned(info *resource.Info) runtime.Object {
+func asVersionedOrUnstructured(info *resource.Info) runtime.Object {
+	obj, _ := asVersioned(info)
```
thomastaylor312: Would it be better to just have the caller ignore the error instead of creating a separate function?

thomastaylor312 (commented):
The code looks pretty good, just one minor discussion topic. I'll need to validate this some more in action and I'll get that done on Monday

pkg/kube/client.go
mumoshu: @thomastaylor312 Hey! Thanks a lot for your review 😄 

I thought so at first but ended up with the separate functions for readability and reliability reasons.

The way `asVersioned` works is that it returns an unstructured object rather than a versioned object, which imho can't be inferred from its name. Theoretically I could have given `asVersioned` an another name or make the implementation simpler so that naming it can be also easier. But I was unable to come up with a great idea and had no confidence to not break it along the way.

With these separate functions, anyway, they are easy to read and I'm sure it doesn't break.

`asVersioned` just returns an error when the object isn't versioned(actually it returns an unstructured object with an error object if it wasn't registered as a versioned object but you don't usually use the object in that case).

`asVersionedOrUnstructured` does what its name tells - returns either a versioned or an unstructured object.

That said, I just wanted to clarify my intention here. I'm still open to suggestions. 

So, If you still think the separate function isn't necessary, I'm absolutely fine to merge the functions into one and ignore errors on caller-side when necessary.

WDYT?

thomastaylor312: I trust your opinion on this, so I am fine with having the separate functions. I'll be porting these over to Helm 3 as well (where they are actually exported functions), so now is as good of time as any to change up the function signatures anyway

thomastaylor312 (approved):
Based on my testing, it looks like this fixes the issue and doesn't unintentionally break other types of chart installs, so this should be good to go. Thanks again for the fix @mumoshu

bacongobbler (approved):
LGTM

Issue #6015, opened 2019-07-12
Have a PGP KEYS file for release verification


mattfarina:
Releases of Helm are PGP signed. But, anyone who tries to verify the signatures will have to hunt down the right keys. 

We should follow a model, like Apache, where the keys are readily available. [On the apache project there is a KEYS file with directions on using it ](https://www.apache.org/dev/release-signing.html#keys-policy)and the keys for those who have signed releases. It is easy to import and use. There is an [example file here](https://www.apache.org/dist/ant/KEYS).

We should adopt this method or a similar one to make verification easier.

/cc @hickeyma @bacongobbler @adamreese @technosophos

technosophos:
Agree. I would like to just include the keys file in Git if there isn't a concern about some sort of Git-based attack. Option B might be to host it off of a Keybase web page.

jkroepke:
From: https://www.apache.org/dev/release-signing.html#keys-policy
> Note: this system will be replaced by a better process in the near future. In preparation, please ensure that public keys are connected as strongly as possible to the Apache web of trust and are available from the major public key servers.


If you are using option A your keys should also on major pgp servers.

mattfarina:
@technosophos might I suggest the KEYS file be at helm.sh/KEYS and get.helm.sh/KEYS for someone to fetch. Also, the keys themselves should be in a 3rd party system like keybase and the keyserver network

@jkroepke keybase is different from the keyservers. What are your thoughts on keybase?

nce:
Well in the end we'd like a way to verify the releases. And in some way there is a manual step required in the verification process. 

If we get the keys manually from either git, keybase or helm.sh, we don't care. But i'd strongly recommend to put the keys on a public available keyserver, as it would be possible to automate the keysearch via gpg. And then it would be compliant (for us).
With Keys on 3rd Party Sites, we would be secure from a compromised git repo.

Thanks for your efforts!

mattfarina:
Here's an outline of tasks that need to be accomplished:

- [ ] Create a KEYS file
- [ ] Put the KEYS file in get.helm.sh alongside downloads
- [ ] Have everyone who has signed a release put their public keys in the KEYS file and in the keyserver network
- [ ] In the maintainers guide document how to add a key to the KEYS file and put the key in the keyserver network
- [ ] Document, for consumers, how to verify a helm download with gpg (make it practical)
[...]

PR #6105, opened 2019-07-29
chore(*): Add public signing key

hickeyma:
Add public key for signed releases

Part of: #6015

Public key needed for signed releases.

**If applicable**:


KEYS
```diff
@@ -149,4 +149,62 @@ qdCDQHoUOgtZ4v6+QaxI3WV1KPBsPb7SAjuphubIQVK/6qHse9OoWVwWAABXHFqX
 +YMbPW0/IQ==
 =g11H
 -----END PGP PUBLIC KEY BLOCK-----
+pub   rsa4096 2019-05-15 [SC]
```
mattfarina: Small nit... Can you add a blank line between `-----END PGP PUBLIC KEY BLOCK-----` and the start of your content

hickeyma: Can do and thought about this. One thing though - would this blank line be added if you used the instruction from KEYS file per verbatim?

hickeyma: Added the blank line

mattfarina (approved):

KEYS
mattfarina: > would this blank line be added if you used the instruction from KEYS file per verbatim?

@hickeyma I'm not entirely sure. Would need to dig more into this.

hickeyma: That's ok. just asking.

PR #6110, opened 2019-07-30
pass username and password to pluginGetter

### LINKED ISSUE #6103: Feature Request: Pass username and password to plugin
Output of `helm version`:
```
Client: &version.Version{SemVer:"v2.14.2", GitCommit:"a8b13cc5ab6a7dbef0a58f5061bcc7c0c61598e7", GitTreeState:"clean"}
```

Currently helm does not pass username and password to plugins, only certFile, keyFile, caFile are passed.
See https://github.com/helm/helm/blob/master/pkg/getter/plugingetter.go#L67

I'm thinking about adding it. I can adding more args here, or I can pass username and password via environment variables in https://github.com/helm/helm/blob/master/pkg/getter/plugingetter.go#L69

How do you think? Thanks

xh4n3:
When we're adding a new repo, we should pass username and password to pluginGetter, then our plugin can get the repo index. If we do not pass, there is no way we can add a repo by plugin.

close #6103 
I haven't attached testcase and docs yet, for reviewing only.

### DIFF
--- a/pkg/getter/plugingetter.go
+++ b/pkg/getter/plugingetter.go
@@ -51,22 +51,33 @@ func collectPlugins(settings environment.EnvSettings) (Providers, error) {
 	return result, nil
 }
 
-// pluginGetter is a generic type to invoke custom downloaders,
+// PluginGetter is a generic type to invoke custom downloaders,
 // implemented in plugins.
-type pluginGetter struct {
+type PluginGetter struct {
 	command                   string
+	username, password        string
 	certFile, keyFile, cAFile string
 	settings                  environment.EnvSettings
 	name                      string
 	base                      string
 }
 
+//SetCredentials sets the credentials for the getter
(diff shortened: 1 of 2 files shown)


bacongobbler:
Sorry for not getting back to you sooner. The team has been focusing for the past few months on getting Helm 3 out the door, and that has meant it's been harder to find time to review feature additions for Helm 2.

Helm 2.15 is soon to be released, and will be the last feature release for Helm 2. Because of the time commitment required to review test this, we won't be able to find the time to review.

That being said, if you want to try again by reconsidering this for Helm 3, that would be much appreciated. I'd hold off for a few weeks until 2.15 is out as we are planning to move the dev-v3 branch over to be the new master branch, so there will be some shuffling around for a bit, but once that's done we'd be happy to consider this for the 3.1.0 release. I'd highly suggest taking a look at Helm 3 first to make sure this hasn't already been addressed.

Thanks again!

PR #6111, opened 2019-07-30
Update example to set selector labels

virtualdom:
Updates the `nginx` example to be functional. Without the `selector` block configured correctly, the following error is produced

```
Error: release fuzzy-horse failed: Deployment.apps "fuzzy-horse-nginx" is invalid: [spec.selector: Required value, spec.template.metadata.labels: Invalid value: map[string]string{"app.kubernetes.io/instance":"fuzzy-horse", "app.kubernetes.io/name":"nginx"}: `selector` does not match template `labels`]
```


bacongobbler (approved):
LGTM!

PR #6117, opened 2019-07-30
feat(chart): Update linting and checking for apiVersion v2

hickeyma:
Update linting and checking for apiVersion v1 and v2.

Some work previously done like in PR #5730.

Partial close for: #5907

Helm v3 has introduced new constructs in the charts for apiVersion v2. Checks needed to handle the changes between apiVersion v1 and v2.

What changes to annotations were introduced in v3?

**If applicable**:


pkg/chart/loader/load.go
```diff
@@ -102,7 +102,11 @@ func LoadFiles(files []*BufferedFile) (*chart.Chart, error) {
 
 		// Deprecated: requirements.yaml is deprecated use Chart.yaml.
 		// We will handle it for you because we are nice people
+		// Only supported for apiVersion v1
 		case f.Name == "requirements.yaml":
+			if c.Metadata.APIVersion != chart.APIVersionV1 {
+				return c, errors.Errorf("dependencies in requirements.yaml are not supported in apiVersion '%s'", c.Metadata.APIVersion)
```
bacongobbler: requirements.yaml is deprecated, but we will continue to support it in v2. We could probably inform the user that they should migrate to using Chart.yaml instead but we still should support this if they bump versions. See line 104 :)

bacongobbler (changes requested):

pkg/chart/loader/load.go
hickeyma: Good point @bacongobbler. Someone may bump it but maintain the requirements file. Updated accordingly.

hickeyma:
Ready for review again.

bacongobbler (approved):

pkg/chart/loader/load.go
```diff
@@ -103,6 +104,9 @@ func LoadFiles(files []*BufferedFile) (*chart.Chart, error) {
 		// Deprecated: requirements.yaml is deprecated use Chart.yaml.
 		// We will handle it for you because we are nice people
 		case f.Name == "requirements.yaml":
+			if c.Metadata.APIVersion != chart.APIVersionV1 {
+				log.Printf("Warning: Dependencies are handled in Chart.yaml since apiVersion \"v2\". We recommend migrating dependencies to Chart.yaml.")
+			}
```
mattfarina: This is a warning if the `apiVersion` is `v2` and there is a `requirements.yaml` file. Would it be too much to remove the `requirements.yaml` from the `v2` chart and make this an error?

pkg/chart/loader/testdata/frobnitz.v2.reqs/Chart.lock
```diff
@@ -0,0 +1,8 @@
+dependencies:
```
mattfarina: If there are dependencies in the lock file shouldn't the appropriate piece be in the `Chart.yaml` file, too?

pkg/chart/loader/load.go
hickeyma: @mattfarina There was feed back from @bacongobbler in https://github.com/helm/helm/pull/6117#discussion_r308900043 about handling the requirements file and giving a warning.

pkg/chart/loader/testdata/frobnitz.v2.reqs/Chart.lock
hickeyma: That could only be described as a human error mistake! :) Apologies, will remove.

pkg/chart/loader/load.go
bacongobbler: Because the translation from requirements.yaml to Chart.yaml are 1:1 there's no technical reason why we can't support both in apiVersion v2.

mattfarina: @bacongobbler @adamreese @hickeyma Are we thinking of these as specs we can independently document and have multiple tools work on things that conform to the specs? With the `v1` we didn't do that even though there was a task for it. If we do think of it as a spec... than what does the spec say about the location of the requirements? If one of us writes the spec on this how do we describe the differences?

bacongobbler: I'll defer to @adamreese, but you do raise good points RE: third-party tools looking to support apiVersion v2.

bacongobbler (changes requested):
rescinding LGTM due to ongoing conversations

pkg/chart/loader/load.go
bacongobbler: @mattfarina would it be fair to move forward with this PR as-is and carry on the discussion in a ticket? This retains the same functionality as before. We can discuss whether or not we want to not support requirements.yaml in apiVersion v2 as a separate PR. How does that sound?

mattfarina: That sounds fine. I started #6151 to have that conversation.

Note, the whole reason I thought of a spec being written out is a) I've been asked about one and b) it would cause us to think some things through that we may be missing.

pkg/chart/loader/testdata/frobnitz.v2.reqs/Chart.lock
mattfarina: If you add the dependencies to the `Chart.yaml` I'll come through and approve this one given the other conversation is moving to a follow-up.

pkg/chart/loader/load.go
hickeyma: @mattfarina Good idea. Putting it out in the open will help us and the users afterwards.

pkg/chart/loader/testdata/frobnitz.v2.reqs/Chart.lock
hickeyma: The lock file should not have been there as the dependencies are in the requirements. The test is to show that v3 will handle chart specified as apiVersion "v2" but using dependencies in requirements file.

hickeyma:
@bacongobbler @mattfarina ready for review again.

bacongobbler (approved):

bacongobbler:
feel free to merge when ready

hickeyma:
Thanks @bacongobbler. I'll wait on @mattfarina before I merge.

mattfarina (approved):

hickeyma:
Thanks @mattfarina.

PR #6120, opened 2019-07-30
Add tests for helm completion (#6115)

### LINKED ISSUE #6115: Tests for helm completion
Unless I'm wrong, there is currently no way to test command-line completion for helm.

Although completion is handled quasi-automatically by Cobra, there is code residing in the Helm project directly.  It has happened that completion was broken for helm such as #5072 #5406 .  And there are feature requests as well as bug reports in the completion area such as #3693 #3170 #3171 .  As more completion features are added to the Helm project, such as dynamic completion, tests will be more and more important.

Such tests should allow to cover:
- bash 4
- bash 3
- zsh
- Linux
- MacOS (as much as possible)
- static completion
- dynamic completion
 
I will be proposing a solution soon.

marckhouzam:

This commit adds a framework and tests for helm command-line
completion.

To cover different shells (bash 4, bash 3, zsh) we use docker.

For MacOS, the tests are run locally if the host is MacOS and has
the right setup (bash completion installed, or zsh installed).

Dynamic-completion (e.g., `helm status <TAB>`) is not yet tested, but has
been considered in the design of the framework.

To run the tests:
  `make test-completion`

Although the framework supports marking tests as known failures,
this commit commented out the failing tests to allow for a clean
output for this proposal.  They can be uncommented later on.

A new directory is added: `scripts/completion-tests`.
It contains the required scripts for the completion tests.

`scripts/completion-tests/completionTests.sh`:
- helm-specific completion tests.
- This script should evolve as new tests are added and new completion features implemented.

`scripts/completion-tests/test-completion.sh`:
- This script repeatedly runs the completion-tests in different environments using docker.
- This script would need changes if a new environment needs to be added.

`scripts/completion-tests/lib/completionTests-base.sh`:
- The magic. It is this script that allows completion testing.
- This script is not expected to change in day-to-day helm evolution.

closes #6115

To see a test fail, you can uncomment one of the failing tests of 
`scripts/completion-tests/completionTests.sh`
or you can modify the expected result of one of those tests to something bogus such as
`_completionTests_verifyCompletion "helm stat" "statusBAD"`

Interesting example is that the test
`#_completionTests_verifyCompletion ZFAIL "helm --kubeconfig=/tmp/config lis" "list"`
can be seen to start working if we apply PR #5680.  Should allow to review that PR with more confidence.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,6 +81,12 @@ test-style: vendor $(GOLANGCI_LINT)
 	$(GOLANGCI_LINT) run
 	@scripts/validate-license.sh
 
+.PHONY: test-completion
+test-completion: TARGETS = linux/amd64
+test-completion: build build-cross
+test-completion:
+	scripts/completion-tests/test-completion.sh
+
 .PHONY: verify-docs
 verify-docs: build
 	@scripts/verify-docs.sh
--- a/scripts/completion-tests/completionTests.sh
+++ b/scripts/completion-tests/completionTests.sh
@@ -1,0 +1,64 @@
+#!bash
+#
+# Copyright (C) 2019 Ville de Montreal
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This script tests different scenarios of completion.  The tests can be
+# run by sourcing this file from a bash shell or a zsh shell.
+
+source /tmp/completion-tests/lib/completionTests-base.sh
+
+# Don't use the new source <() form as it does not work with bash v3
+source /dev/stdin <<- EOF
+   $(helm completion $SHELL_TYPE)
+EOF
+
+# No need to test every command, as completion is handled
+# automatically by Cobra.
+# We focus on some smoke tests for the Cobra-handled completion
+# and also on code specific to this project.
+
+# Basic first level commands (static completion)
+_completionTests_verifyCompletion "helm stat" "status"
+_completionTests_verifyCompletion "helm status" "status"
+_completionTests_verifyCompletion "helm lis" "list"
+_completionTests_verifyCompletion "helm r" "registry repo rollback"
+_completionTests_verifyCompletion "helm re" "registry repo"
+
+# Basic second level commands (static completion)
+_completionTests_verifyCompletion "helm get " "hooks manifest values"
+_completionTests_verifyCompletion "helm get h" "hooks"
+_completionTests_verifyCompletion "helm completion " "bash zsh"
+_completionTests_verifyCompletion "helm completion z" "zsh"
+
+# Completion of flags
+#_completionTests_verifyCompletion ZFAIL "helm --kube-con" "--kube-context= --kube-context"
+#_completionTests_verifyCompletion ZFAIL "helm --kubecon" "--kubeconfig= --kubeconfig"
+#_completionTests_verifyCompletion ZFAIL "helm --name" "--namespace= --namespace"
+_completionTests_verifyCompletion "helm -v" "-v"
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -81,6 +81,12 @@ test-style: vendor $(GOLANGCI_LINT)
 	$(GOLANGCI_LINT) run
 	@scripts/validate-license.sh
 
+.PHONY: test-completion
+test-completion: TARGETS = linux/amd64
```
mattfarina: Any chance these tests could be made to work on multiple platforms?

scripts/completion-tests/completionTests.sh
```diff
@@ -0,0 +1,64 @@
+#!bash
+#
+# Copyright (C) 2019 Ville de Montreal
```
mattfarina: The copyright should read:

> Copyright The Helm Authors.

Can you please update it throughout the PR.

mattfarina (commented):
Thanks for the PR.

@jdolitsky recently started an [acceptance testing repo](https://github.com/helm/acceptance-testing). I wonder if this is the type of thing that belongs in there or here.

scripts/completion-tests/completionTests.sh
marckhouzam: I was hoping you'd catch this since I wasn't sure how to handle it.

IIUC the DCO is a way for me to contribute this code to the project under the licence used by the project; the DCO does not transfer the copyright ownership of the code however. 

So, since I'm not a Helm Author and don't have authorization to give away the copyright, I didn't think it was ok to put the "The Helm Authors".  In other projects I've seen the form "<original copyright holder> and others" to more appropriately reflect the copyright ownership.

Do you know of guidelines from the CNCF on this?

I don't mean to be a pain, but I also don't want to overstep what I'm allowed to do.

Makefile
marckhouzam: I made the assumption that the person running these tests would run docker under linux/amd64.  I can try to figure out where docker is being run and choose the correct platform.

Is that what you mean?

marckhouzam:
@mattfarina Good point about the acceptance-testing repo.  I hesitated after seeing the announcement about this new repo.  However, I liked the fact that these tests could be run while doing development as it would allow to handle regressions of the completion logic more quickly.

When/if we move forward and get to the point of adding testing of dynamic completion, I was going to propose to avoid using a real kubernetes cluster and instead use a fake store as is done with the go tests.  So the completion tests could still be run easily by the developer.

I'm just voicing my opinion but of course it is your call.

scripts/completion-tests/completionTests.sh
bacongobbler: The CNCF wants all projects to be Apache 2 licensed. One of the reasons is because of point 5 under the terms of use, which calls this out explicitly: https://github.com/helm/helm/blob/master/LICENSE#L131-L137

jdolitsky:
@marckhouzam originally I had PR'd the acceptance tests to helm/helm under the same premise that I would like to run them during development. However, it is easy enough to just run a "make build" and set PATH appropriately prior to running the acceptance test suite.

On that basis, I'd be in favor of moving this stuff over to [helm/acceptance-testing](https://github.com/helm/acceptance-testing), under some new file `completion.robot`. Let me know if you want to chat about this.

Criteria for what to put in acceptance test repo should probably be "does a fully-built version of Helm work in its target environment(s)?" (different Kubernetes clusters, different shells, etc.)

scripts/completion-tests/completionTests.sh
marckhouzam: This contribution is made under the Apache 2 license, so it can be part of the project directly.

It's just that the Copyright ownership is not transferred by the DCO, so I thought it should be written properly who the copyright owner is, in this case Ville de Montreal.  But the license remains Apache 2.

marckhouzam:
@jdolitsky Thanks for the context.  I'll have a deeper look at the acceptance-testing repo to see how things could work there.  I'll post back once I make progress or if I have questions :-)

marckhouzam:
I'm closing this PR as it was moved to the acceptance-testing repo:
https://github.com/helm/acceptance-testing/pull/12

The question of Copyright is still under discussion.

scripts/completion-tests/completionTests.sh
marckhouzam: To complete on this discussion, I've contacted the CNCF about the copyright question and they explained that by making a contribution to the Helm project, my employer (through me) becomes a "Helm Author". Therefore, it is ok to put "Copyright The Helm Authors" as part of my contribution even though the copyright is not transferred. Sorry for me not realizing this.

PR #6124, opened 2019-07-30
ref(client): use three-way merge patch strategy

### LINKED ISSUE #1873: Helm doesn't updated resource though Chart's template changed
Client version "v2.1.3"

I'm using customized `nginx-lego` chart from a folder. 

First I installed it as it is, then adde arguments here:
https://github.com/kubernetes/charts/blob/master/stable/nginx-lego/templates/nginx-deployment.yaml#L38

So in charts templates I have changed arguments:
image

Then I do `helm upgrade` release, the deployment is not updated there is still two arguments:
image

Is there a way to imperatively upgrade the deployment?

### LINKED ISSUE #3805: Perform three way merge patch during `helm upgrade` call as `kubectl apply` does
I've checked https://github.com/kubernetes/helm/blob/master/pkg/kube/client.go#L432 and found out that Helm v2.8.x uses a two way strategic merge patch between the old and the new configuration of a resource to modify it during `helm upgrade` call.
However, `kubectl apply` uses a three way merge patch between the old configuration, the new configuration and the current state of the resource: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/apply.go#L670-L679 (https://kubernetes.io/docs/concepts/overview/object-management-kubectl/declarative-config/#merge-patch-calculation explains the algorithm in a bit more details)
Is there any specific reason why Helm does not use `strategicpatch.CreateThreeWayMergePatch`? Would the project be ok with using the function if I were to prepare a patch?

Thank you.

bacongobbler:
closes #3805 
closes #1873
rebased on top of #6085

See 489430a for the changes in this PR.

Co-Signed-by: Taylor Thomas <taylor.thomas@microsoft.com>

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -29,7 +29,6 @@ import (
 	"github.com/pkg/errors"
 	batch "k8s.io/api/batch/v1"
 	v1 "k8s.io/api/core/v1"
-	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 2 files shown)


hickeyma:
@bacongobbler @thomastaylor312 Will review when I get a chance.

pkg/kube/client.go
```diff
@@ -301,7 +305,13 @@ func createPatch(target *resource.Info, current runtime.Object) ([]byte, types.P
...
 		return patch, types.MergePatchType, err
 	}
-	patch, err := strategicpatch.CreateTwoWayMergePatch(oldData, newData, versionedObject)
+
+	patchMeta, err := strategicpatch.NewPatchMetaFromStruct(versionedObject)
+	if err != nil {
+		return nil, types.StrategicMergePatchType, errors.Wrap(err, "unable to create patch metadata from object")
+	}
+
+	patch, err := strategicpatch.CreateThreeWayMergePatch(oldData, newData, currentData, patchMeta, true)
```
thomastaylor312: Question for anyone who comes to check this PR out: The last boolean is a flag that forces an overwrite of the incoming data. That is what we consider to be a good default behavior (the new chart always wins). However, an alternative to this would be to have the `--force` flag control this boolean, so the default would be if there are any conflicts, it would error out. Is there any reason to have this second option?

Personally, I think it is better to leave this how it is, but want to point it out in case there are any good use cases we haven't considered

adamreese:
Has this been manually tested?  I'll try breaking it today.  The code LGTM though

adamreese (approved):

thomastaylor312:
@adamreese We tested this fairly extensively before opening the PR, walking through various upgrade and rollback scenarios

hickeyma (commented):
LGTM, thanks. Tested manually with some basic changes in the cluster and then on upgrade.

PR #6135, opened 2019-07-31
Uploading KEYS file to azure along with dist files

mattfarina:


bacongobbler (approved):

bacongobbler:
Nice! Yeah this should land under get.helm.sh/KEYS

PR #6142, opened 2019-08-01
cmd/*,pkg/*: move ValueOptions and decouple from SDK

### LINKED ISSUE #5827: Helm 3: action.Install and action.Upgrade have different mechanisms to set values
Using the Helm 3 libraries, I'm looking for a consistent way to populate an action struct's chart values from an existing `map[string]interface{}`.

### Install 

`install.Run` uses `i.rawValues` (a `map[string]interface{}`) when calling `ToRenderValues`

https://github.com/helm/helm/blob/2a3b7b5395bb7d71919ed3592c8c63c425353851/pkg/action/install.go#L128

`action.Install` exposes `Values` as `[]string` from its embedded `ValuesOptions` struct. It's `rawValues` is unexported.

https://github.com/helm/helm/blob/2a3b7b5395bb7d71919ed3592c8c63c425353851/pkg/action/install.go#L65-L89

So to populate values for an install, it looks like I have to:
1. Marshal my `map[string]interface{}` value out to a temporary YAML file
2. Set `install.ValueFiles`
3. Call `install.ValueOptions.MergeValues`
4. Call `install.Run`
5. Cleanup the temporary file.

### Upgrade

`upgrade.Run` uses `u.Values` when calling `ToRenderValues` (and seems to ignore `u.rawValues`?)

https://github.com/helm/helm/blob/2a3b7b5395bb7d71919ed3592c8c63c425353851/pkg/action/upgrade.go#L157

So since `Values` is exported publicly as a `map[string]interface{}` (shadowing `ValuesOptions.Values`),  it looks like I can set `upgrade.Value` directly, and everything seems to work.

https://github.com/helm/helm/blob/2a3b7b5395bb7d71919ed3592c8c63c425353851/pkg/action/upgrade.go#L37-L59

Just want to make sure I'm understanding things correctly and to see if this discrepancy is intended :slightly_smiling_face:

Thanks!

joelanford:
Closes #5827 

This PR decouples the CLI value parsing from the various Helm SDK `action` clients that require values. It:
- Updates `ValueOptions.MergeValues()` to return merged values as a `map[string]interface{}`
- Removes `ValueOptions.rawValues`
- Removes `ValueOptions` from `action.Install`, `action.Lint`, `action.Package`, and `action.Upgrade`
- Moves `ValueOptions` to `pkg/cli/values` since it is no longer used in the `action` package
- Changes `Run()` signature for `action.Install`, `action.Lint`, `action.Package`, and `action.Upgrade` to include `vals map[string]interface{}` as an input parameter.
- Removes `action.mergeValues` function, since it seems to be unused.
- Removes `action.NewValueOptions()` constructor, since `rawValues` no longer exists.
- Updates unit tests to account for changes.

This PR is a breaking change for existing users of the SDK (e.g. #6008)

/cc @bacongobbler 

Signed-off-by: Joe Lanford <joe.lanford@gmail.com>

### DIFF
--- a/cmd/helm/install.go
+++ b/cmd/helm/install.go
@@ -18,19 +18,26 @@ package main
 
 import (
 	"io"
+	"io/ioutil"
+	"net/url"
+	"os"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
+	"sigs.k8s.io/yaml"
 
 	"helm.sh/helm/cmd/helm/require"
 	"helm.sh/helm/pkg/action"
 	"helm.sh/helm/pkg/chart"
 	"helm.sh/helm/pkg/chart/loader"
+	"helm.sh/helm/pkg/cli"
 	"helm.sh/helm/pkg/downloader"
 	"helm.sh/helm/pkg/getter"
 	"helm.sh/helm/pkg/release"
+	"helm.sh/helm/pkg/strvals"
 )
 
 const installDesc = `
@@ -97,6 +104,7 @@ charts in a repository, use 'helm search'.
 
 func newInstallCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 	client := action.NewInstall(cfg)
+	valueOpts := &ValueOptions{}
 
 	cmd := &cobra.Command{
 		Use:   "install [NAME] [CHART]",
@@ -104,7 +112,7 @@ func newInstallCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 		Long:  installDesc,
 		Args:  require.MinimumNArgs(1),
 		RunE: func(_ *cobra.Command, args []string) error {
-			rel, err := runInstall(args, client, out)
+			rel, err := runInstall(args, client, valueOpts, out)
 			if err != nil {
 				return err
 			}
@@ -113,12 +121,12 @@ func newInstallCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 		},
 	}
 
-	addInstallFlags(cmd.Flags(), client)
+	addInstallFlags(cmd.Flags(), client, valueOpts)
 
 	return cmd
 }
 
-func addInstallFlags(f *pflag.FlagSet, client *action.Install) {
+func addInstallFlags(f *pflag.FlagSet, client *action.Install, valueOpts *ValueOptions) {
 	f.BoolVar(&client.DryRun, "dry-run", false, "simulate an install")
 	f.BoolVar(&client.DisableHooks, "no-hooks", false, "prevent hooks from running during install")
 	f.BoolVar(&client.Replace, "replace", false, "re-use the given name, even if that name is already used. This is unsafe in production")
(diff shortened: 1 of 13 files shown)


bacongobbler (approved):

cmd/helm/install.go
```diff
@@ -223,3 +232,89 @@ func isChartInstallable(ch *chart.Chart) (bool, error) {
 	}
 	return false, errors.Errorf("%s charts are not installable", ch.Metadata.Type)
 }
+
+type ValueOptions struct {
```
bacongobbler: Is there a way we can add this back to the action package? I imagine other users would want to merge values together similar to the CLI, so it'd be great to make this functionality public but optional.

bacongobbler:
This is looking great. Thanks @joelanford! Just one comment on a change otherwise this looks good to go.

cmd/helm/install.go
joelanford: Hmm, moving it to `pkg/cli` make sense to me, but it causes an import cycle.

```
import cycle not allowed
package helm.sh/helm/cmd/helm
	imports helm.sh/helm/cmd/helm/search
	imports helm.sh/helm/pkg/repo
	imports helm.sh/helm/pkg/getter
	imports helm.sh/helm/pkg/cli
	imports helm.sh/helm/pkg/getter
```

The `readFile` function used by `MergeValues`  requires importing `pkg/getter`, but `getter` imports `pkg/cli` to use `cli.EnvSettings`.

Not sure if you have any ideas for good ways to resolve that, but for now I'll move `cmd/helm.ValueOptions` to `pkg/cli/values.Options`

bacongobbler:
LGTM! Just missing a DCO sign-off. Once that's done I can ping another maintainer to review this and get this in for the next release.

joelanford:
Done!

pkg/cli/values/options.go
```diff
@@ -0,0 +1,117 @@
...
+
+type Options struct {
+	ValueFiles   []string
+	StringValues []string
+	Values       []string
+}
+
+// MergeValues merges values from files specified via -f/--values and
+// directly via --set or --set-string, marshaling them to YAML
+func (opts *Options) MergeValues(settings cli.EnvSettings) (map[string]interface{}, error) {
```
mattfarina: Did the tests for `MergeValues` get dropped when moving from one package to another?

joelanford: Possibly. In my initial run through the changes, I noticed that only `mergeValues` (which is unused) was actually being tested in `TestMergeValues`.

Looking closer, I see that `mergeMaps`, which _is_ used in `MergeValues` has the same function signature, and appears to do the same thing, though in a slightly different way?

@adamreese Here's the commit where both `mergeMaps` and `mergeValues` were added: https://github.com/helm/helm/commit/0338576fc5202d003e346f41069d50e4f006a285

Is there any reason we should keep both or choose one over the other?

Since `mergeMaps` is what's currently being used in `MergeValues`, I'll plan to re-add `TestMergeValues` and update it to use `mergeMaps` unless I'm told otherwise.

joelanford: Done. Re-added `TestMergeValues`, testing `mergeMaps`.

bacongobbler (approved):

bacongobbler:
needs one more LGTM before merging, so I'll attach the label.

adamreese (approved):
Looks great thanks!

PR #6171, opened 2019-08-07
fix mis-spelling in  manager.go

AllenZMC:
make the comment more readable


hickeyma (approved):
LGTM, thanks @AllenZMC. circleci build is failing with non connected issues. I have kicked off another build and will merge when it passes.

Issue #5311, opened 2019-02-14
Proposal: make cobra.Command functions public


melan:
Currently all functions generating cobra.Command for CLI are private and I want to make them public. Motivation for such change is simple: this will allow to reuse pieces of helm functionality in other tools.

Story: I'm building a CLI for a project that mostly runs on Kubernetes. The CLI is composed from 2 pieces: commands unique for the project and commands imported from from `kubectl`. I would like to add there few commands from `helm` also - first of all `helm template`. Such aggregated CLI allows to use one command to interact with the project instead of several different tools each with own configuration.

melan:
This change seems minor and unless there are reasons why not to implement it - I'd like to contribute the changes.

bacongobbler:
re-posting my comment in #5353 for posterity:

IIRC Go considers the `cmd` package as an "internal" package, so even if we made them public, third parties won't be able to take advantage of these commands. In the past, the internals of `cmd` changed often due to breaking changes in `k8s.io/kubernetes`, which is why we have avoided exposing these commands in the past. I'm not sure if we're still comfortable with the idea of exposing these endpoints.

That being said, @adamreese and I are refactoring these packages for Helm 3 so that they can be consumed by other third parties in a more meaningful way. We should have more to report in a few weeks. Will that help?

bacongobbler:
So, #5365 is a major internal refactoring of most of the Helm codebase in preparation for Helm 3. Large parts of the functionality in `cmd` has been moved into the `action` package (new for Helm 3), but the commands themselves are still in `cmd`... For now.

After we suss out the details on that PR, my line of thinking to address this ask is to move all the `new*Cmd` functions over to `pkg/cli`. That should address this proposal for Helm 3. I don't want to do that in that PR though as we're already approaching 13k lines of code changed, and we can address this further refactoring in another PR.

Given the amount of code that would need to move over or make public from `cmd/helm`, I'm concerned about the amount of refactoring (and the potential for compatibility issues involved) to support this for Helm 2.

melan:
@bacongobbler do you need help to move this proposal forward?

bacongobbler:
Yes, actually if you'd like to take this work on that'd be great. I started going down the path in #5448 but got sidetracked with higher priority work so I was unable to finish it.

If you'd like to work on this, we'd need:

- the actual work of exposing the cobra commands in a consumable package (like `pkg/cli`)
[...]

PR #2938, opened 2017-09-15
Remove facebookgo dependencies

bacongobbler:
These dependencies were introduced in both #2010 and #2449. Facebook's BSD+patent license is incompatible with the Apache v2 license. Reverting these two PRs for now so we are OK for compliance purposes, but we will need to re-implement these fixes with our own packages/alternative packages that are Apache v2 compatible.

closes #2933 
closes #2934

re-opens #2449 (we'll have to file a new ticket for this)
re-opens #1639

PR #6176, opened 2019-08-07
Remove unneeded public APIs before going beta

thomastaylor312:
After talking with some of the other maintainers, we decided to move any libraries in `pkg` that were unlikely to be used by those consuming the SDK. This allows us more flexibility to refactor and expose those APIs later as needs arise without breaking backwards compatibility. This is a good first pass at moving things to `internal`. Let me know if there are any that shouldn't be there or if there are any I could move over.

The following packages were moved to internal:
- ~chartutil (consumers of the SDK should be using the `chart` package to load charts)~ (reverted)
- ignore
- resolver
- ~strvals~ (reverted)
- sympath
- tlsutil
- urlutil

In addition `pkg/version` has been removed. This package was a weird hodgepodge of old Helm 2 logic and a function that had been co-opted for chart use. The only used function has been moved to chartutil and the version build info has been moved to the internal version package.

Please note that most of these are renames, with a few small "campfire rule" changes I found along the way


thomastaylor312:
Other possibilities for packages in internal:
- The storage drivers
- Releaseutil (if we think people won't be managing the actual release objects themselves)
- helmpath (since it is mostly used internally)

bacongobbler:
From searching on github:

`pkg/chartutil` is [used fairly commonly across multiple use cases](https://github.com/search?q=k8s.io%2Fhelm%2Fpkg%2Fchartutil&type=Code), so it should probably be restored.

`pkg/ignore` is used mostly during custom chart loading, which I agree that users should use `pkg/chart` for that. I think we can get away with making it private. It does appear to be a useful package for others to use for writing their own ignore files, though, so I'm kinda torn about making this private.

`pkg/strvals` is used on occasion by several projects.

sympath was designed to be used by others as an more open alternative to facebookgo/symwalk so it should be made public. See https://github.com/helm/helm/pull/2938 for more context.

`tlsutil` and `urlutil` are useful packages so I don't see why we need to make them private.

I guess the question begs why do we need to make these packages private? They're being used by third parties in production so I don't see why they need to be made private. If anything users have been asking us to make packages MORE public (like in https://github.com/helm/helm/issues/5311).

thomastaylor312:
So for sympath, {tls,url}util, and strvals, my question would be do we want to support APIs that are not directly related to Helm? That was my justification for many of those that I moved to internal. I guess that would also apply to these packages.

However, I am not in anyway attached to these ideas. We had just mentioned moving some things in to internal to keep the API surface more manageable

thomastaylor312:
Please note I will squash all my commits before merge

bacongobbler (approved):

bacongobbler:
nice refactors!

mattfarina (approved):
If we document charts v2 as a spec (something we have been asked to do) than the `ignore`  and  `resolver` packages may be useful for those who decide to develop other tools to interact with charts.  We can always move things out of internal after the v3 release if this becomes requested.

PR #6185, opened 2019-08-08
feat(cmd): add feature gates

bacongobbler:
This adds a new `gates` package used for interacting with feature gates. It also marks the OCI registry work as experimental, signalling to users that it is not a stable feature of Helm. By default, these commands are hidden from the user unless they set the correct feature gate.

This also moves the `registry` package into the `internal/experimental` package so that users do not start to rely on this package for production use as it continues to iterate over time.

To use this:

```
$ ./bin/helm chart pull foo
Error: this feature has been marked as experimental and is not enabled by default. Please set $HELM_EXPERIMENTAL_OCI in your environment to use this feature
$ HELM_EXPERIMENTAL_OCI=1 ./bin/helm chart pull foo
: Pulling from foo
Error: object required
```


bacongobbler:
cc @jdolitsky

mattfarina:
This generally looks good to me. Not going to approve as the tests are failing and @jdolitsky should look at it first.

bacongobbler:
oops. Thanks @mattfarina. Will take a look at the failing tests.

cmd/helm/helm.go
```diff
@@ -33,11 +33,15 @@ import (
...
 	"helm.sh/helm/pkg/action"
 	"helm.sh/helm/pkg/cli"
+	"helm.sh/helm/pkg/gates"
 	"helm.sh/helm/pkg/kube"
 	"helm.sh/helm/pkg/storage"
 	"helm.sh/helm/pkg/storage/driver"
 )
 
+// FeatureGateOCI is the feature gate for checking if `helm chart` and `helm registry` commands should work
+const FeatureGateOCI = gates.Gate("HELM_EXPERIMENTAL_OCI")
```
jdolitsky: should this be public?

jdolitsky:
lgtm!

cmd/helm/helm.go
bacongobbler: this is helm-specific code so I'd prefer to keep this in `cmd` as an internal variable for the CLI.

This should be exposed as part of #5311 though.

jzelinskie:
Looks great! Even integrated with Cobra!

jdolitsky (approved):

Issue #6186, opened 2019-08-08
Make `helm search` work against monocular search


mattfarina:
When the stable repository was removed from Helm v3, people lost the ability to search out of the box. The idea is to enable default search to use the Helm Hub (an instance of monocular). Options would allow search to be pointed at another instance of monocular or to fall back to repositories stored locally. OCI based storage search isn't known yet as OCI distributions don't provide a standard search API.

The tasks as I think I see them are:

1. Create a monocular client that Helm can use to search
2. Wire up the CLI to search monocular by default or the local store with a flag
3. Figure out OCI repository search

bacongobbler:
slating this for the beta, though we could let it slip into beta.2 if required.

waveywaves:
@mattfarina Can I give my hand at <2> or <3> is that is fine with you. Once #6187 gets merged ?

bacongobbler:
closed via #6211. Let's address OCI search once that's actually a part of the OCI spec. Right now that doesn't exist.

PR #6187, opened 2019-08-08
Adding a monocular client as a package

mattfarina:
Includes client creation and search

Relates to #6186


pkg/monocular/search.go
```diff
@@ -0,0 +1,132 @@
...
+	"time"
+
+	"helm.sh/helm/pkg/chart"
+)
+
+// The structs below represent the structure of the response from the monocular
+// search API.
+
+// SearchResult represents an individual chart result
+type SearchResult struct {
```
thomastaylor312: Do these structs exist in monocular and could we use them if so?

pkg/monocular/search.go
```diff
@@ -0,0 +1,132 @@
...
+	// Create the URL to the search endpoint
+	// Note, this is currently an internal API for the Hub. This should be
+	// formatted without showing how monocular operates.
+	p, err := url.Parse(c.BaseURL)
+	if err != nil {
+		return nil, err
+	}
+
+	// Set the path to the monocular API endpoint for search
+	p.Path = path.Join(p.Path, "api/chartsvc/v1/charts/search")
```
thomastaylor312: Consider making this part of the path a constant

thomastaylor312 (commented):
This is looking pretty good.

Just wanted to have one discussion around whether or not this should be in `internal` or not. I could see this code being useful to others, but worry about exporting it and supporting it still. A full monocular client feels like something that should belong in monocular and be imported

pkg/monocular/search.go
mattfarina: They do. But, the package imports `k8s.io/helm/pkg/proto/hapi/chart`. I'd like to avoid that package import from helm v2 to helm v3. Any ideas other than copying?

bacongobbler:
I noticved a lot of code in here shares similarities with `pkg/getter/httpgetter.go`. Would it be worth conslidating the HTTP client creation code? For example, the `tlsutil` setup as well as the proxy setup available in httpgetter may be worthwhile to consider.

bacongobbler:
https://github.com/helm/helm/blob/affd77558ffd343716036c01a075ed16540c5aaa/pkg/getter/httpgetter.go#L79-L100

bacongobbler:
I even notice that the same HTTP user agent introduce here is commented out in httpgetter.go... Perhaps we can just uncomment that. Not sure why it's unavailable :thinking: 

https://github.com/helm/helm/blob/affd77558ffd343716036c01a075ed16540c5aaa/pkg/getter/httpgetter.go#L49

mattfarina:
@bacongobbler The user agent used to be set for the httpgetter. When we used to pull chart stats we could see which versions of Helm were calling for it. On master this is still happening. Somewhere in the v3 refactor something changed.

The httpgetter is coupled to the getter system. Is it worth trying to make the two work using the same code? Should the getter system (or just the httpgetter) be used for the basic http request?

bacongobbler:
It might be worth re-using the getter system for monocular. While monocular does not interface directly with a basic HTTP filestore similar to the chart repository API (s3, azure, chartmuseum, gcs), the function signatures for `.Search(term)` and `.Get(url)` seem similar enough in design WRT HTTP client setup. I'll experiment with this today but I wouldn't block this from going through.

The user agent string on the httpgetter should probably be restored. Consumers can overwrite the user agent string if they choose. I'll file a separate PR for that.

mattfarina:
@thomastaylor312 I made the tweaks you suggested. In addition to that I added a user agent function to the version package since we will need that in a couple other places. This is after chatting with @bacongobbler who is specifying the user agent in another PR.

bacongobbler:
#6197 has already been merged, so if you want to update that part of the codebase to use the `.UserAgent` function, that would be dandy. :)

mattfarina:
@bacongobbler done!

bacongobbler (approved):

pkg/monocular/search.go
thomastaylor312: Nope, that makes complete sense

thomastaylor312 (approved):

waveywaves:
There seems to be conflict merges on `internal/version/version.go`

mattfarina:
@waveywaves I knew this was coming. @thomastaylor312 warned me before merging his change that caused this. A rebase with the most minor of changes (i.e.i removing an import) was the only thing needed to correct this.

PR #6197, opened 2019-08-09
feat(getter): set default User-Agent

bacongobbler:


mattfarina (approved):
This looks good. If no user agent is set it will default to this.

I wonder if the code path was meant to have this as a CLI option (with a default) that was passed to an action that was passed here. That could still be done as a follow-up. It would be useful for an application that exec's out to Helm. I think we considered this at one time for chart-testing.

PR #6199, opened 2019-08-09
ref(cmd/helm): remove helm home command

adamreese:
This command only displays documentation.  I think we should remove it
in favor of documentation and add a command later if required.


bacongobbler (approved):
Yeah... This makes sense. LGTM. The command is a little wonky now that Helm's distributed across multiple directories.

We should update the FAQ on `helm home` as well: https://github.com/helm/helm-www/blob/dev-v3-docs/content/docs/faq.md#xdg-base-directory-support

PR #6208, opened 2019-08-12
fix(pkg/kube): only wait for events from Jobs and Pods

adamreese:
Fixes issue of waiting for events from hook objects that are not Jobs or
Pods.


thomastaylor312 (approved):
Good catch on this one

PR #6211, opened 2019-08-12
Exposing Helm Hub search via the search command

mattfarina:
This retains the ability to search added repositories

Part of #6186

### DIFF
--- a/cmd/helm/search.go
+++ b/cmd/helm/search.go
@@ -17,211 +17,27 @@ limitations under the License.
 package main
 
 import (
-	"fmt"
 	"io"
-	"strings"
 
-	"github.com/Masterminds/semver"
-	"github.com/gosuri/uitable"
-	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
-
-	"helm.sh/helm/cmd/helm/search"
-	"helm.sh/helm/internal/monocular"
-	"helm.sh/helm/pkg/helmpath"
-	"helm.sh/helm/pkg/repo"
 )
 
 const searchDesc = `
-Search reads through all of the repositories configured on the system, and
-looks for matches.
-
-Repositories are managed with 'helm repo' commands.
+Search provides the ability to search for Helm charts in the various places
+they can be stored including the Helm Hub and repositories you have added. Use
+search subcommands to search different locations for charts.
 `
 
-// searchMaxScore suggests that any score higher than this is not considered a match.
-const searchMaxScore = 25
-
-type searchOptions struct {
-	versions       bool
-	regexp         bool
-	version        string
-	searchEndpoint string
-	repositories   bool
-	maxColWidth    uint
-}
-
 func newSearchCmd(out io.Writer) *cobra.Command {
-	o := &searchOptions{}
 
 	cmd := &cobra.Command{
 		Use:   "search [keyword]",
 		Short: "search for a keyword in charts",
 		Long:  searchDesc,
-		RunE: func(cmd *cobra.Command, args []string) error {
-			return o.run(out, args)
-		},
 	}
 
-	f := cmd.Flags()
-	f.StringVar(&o.searchEndpoint, "endpoint", "https://hub.helm.sh", "monocular instance to query for charts")
-	f.BoolVarP(&o.repositories, "repositories", "r", false, "search repositories you have added instead of monocular")
-	f.BoolVarP(&o.regexp, "regexp", "", false, "use regular expressions for searching repositories you have added")
-	f.BoolVarP(&o.versions, "versions", "l", false, "show the long listing, with each version of each chart on its own line, for repositories you have added")
-	f.StringVar(&o.version, "version", "", "search using semantic versioning constraints on repositories you have added")
-	f.UintVar(&o.maxColWidth, "maxColumnWidth", 50, "maximum column width for output table")
+	cmd.AddCommand(newSearchHubCmd(out))
+	cmd.AddCommand(newSearchRepoCmd(out))
 
 	return cmd
-}
-
-func (o *searchOptions) run(out io.Writer, args []string) error {
-
-	// If searching the repositories added via helm repo add follow that path
-	if o.repositories {
-		debug("searching repositories")
-		// If an endpoint was passed in but searching repositories the user should
-		// know the option is being skipped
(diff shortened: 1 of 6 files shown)


cmd/helm/search.go
```diff
@@ -60,14 +64,62 @@ func newSearchCmd(out io.Writer) *cobra.Command {
 	}
 
 	f := cmd.Flags()
-	f.BoolVarP(&o.regexp, "regexp", "r", false, "use regular expressions for searching")
-	f.BoolVarP(&o.versions, "versions", "l", false, "show the long listing, with each version of each chart on its own line")
-	f.StringVar(&o.version, "version", "", "search using semantic versioning constraints")
+	f.StringVar(&o.searchEndpoint, "endpoint", "https://hub.helm.sh", "monocular instance to query for charts")
```
thomastaylor312: Looking at the flag documentation, do you think it would be better to have a `helm search repositories|repos` and `helm search hub` for each of the search types with `helm search` being an alias for `helm search hub`? I think it might be less confusing than using a flag to switch it

cmd/helm/search.go
```diff
@@ -60,14 +64,62 @@ func newSearchCmd(out io.Writer) *cobra.Command {
...
+	if o.version != "" {
+		fmt.Fprintln(out, "Notice: Setting the \"version\" flag has no effect when searching monocular (e.g., Helm Hub)")
+	}
+
+	c, err := monocular.New(o.searchEndpoint)
+	if err != nil {
+		return errors.Wrap(err, fmt.Sprintf("unable to create connection to %q", o.searchEndpoint))
+	}
+
+	q := strings.Join(args, " ")
```
thomastaylor312: Do we need to perform any sanitation here? I don't think so, but want to bring it up for discussion

thomastaylor312 (commented):
Code is looking good. Just 2 discussion items before I give the LGTM

cmd/helm/search.go
bacongobbler: If we're following convention from `helm get`, it's probably best to make `helm search` display usage help for the subcommands `helm search hub` and `helm search repositories`. That way the two commands can accept different flags, different arguments, etc and users won't be confused about using, say, why a flag like `--index-file` won't work with `helm search hub` but works with `helm search repositories`.

bacongobbler: That would also help in the future when we support OCI registries; it'd just be another subcommand: `helm search registry`

cmd/helm/search.go
mattfarina: Good question. The escaping happens inside the monocular package [here](https://github.com/helm/helm/blob/29d6853659e212aebe9a8e4ebcb10f4bf66989eb/internal/monocular/search.go#L108).

cmd/helm/search.go
technosophos: Yeah, I agree with this.

bacongobbler:
Working on a follow-up to this PR. Should have something available in the next few hours

cmd/helm/search.go
mattfarina: @bacongobbler it's worth noting that `helm get` does something by default. It just doesn't display information. :)

In this case I will have it display the subcommand information.

cmd/helm/search_hub.go
```diff
@@ -0,0 +1,102 @@
...
+	searchEndpoint string
+	maxColWidth    uint
+}
+
+func newSearchHubCmd(out io.Writer) *cobra.Command {
+	o := &searchHubOptions{}
+
+	cmd := &cobra.Command{
+		Use:   "hub [keyword]",
+		Short: "search for a keyword in charts",
```
bacongobbler: should this get updated to differentiate from `helm search repo`?

cmd/helm/search_hub.go
```diff
@@ -0,0 +1,102 @@
...
+		Long:  searchHubDesc,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return o.run(out, args)
+		},
+	}
+
+	f := cmd.Flags()
+	f.StringVar(&o.searchEndpoint, "endpoint", "https://hub.helm.sh", "monocular instance to query for charts")
+	f.UintVar(&o.maxColWidth, "maxColumnWidth", 50, "maximum column width for output table")
```
bacongobbler: can we change this to `--max-col-width` to stay consistent with other commands?

Also would it be better to bump this to a higher default? Basic search commands like `helm search hub wordpress` all seen to clip due to the length of the URL. `https://hub.helm.sh/charts/stable/` is an extra 34 characters, leaving 16 characters for the chart name.

cmd/helm/search_repo.go
```diff
@@ -0,0 +1,163 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return o.run(out, args)
+		},
+	}
+
+	f := cmd.Flags()
+	f.BoolVarP(&o.regexp, "regexp", "r", false, "use regular expressions for searching repositories you have added")
+	f.BoolVarP(&o.versions, "versions", "l", false, "show the long listing, with each version of each chart on its own line, for repositories you have added")
+	f.StringVar(&o.version, "version", "", "search using semantic versioning constraints on repositories you have added")
+	f.UintVar(&o.maxColWidth, "maxColumnWidth", 50, "maximum column width for output table")
```
bacongobbler: can we change this to `--max-col-width` to stay consistent with other commands?

bacongobbler (changes requested):
LGTM! Just some minor vanity changes then looks good to ship.

cmd/helm/search_hub.go
mattfarina: I meant to change those and forgot. Thanks for catching that.

cmd/helm/search_hub.go
mattfarina: I toyed with different sizes for the columns. On a big monitor it was fine. But, on a laptop screen, like my 13 inch display, it was a problem if I went much bigger. That is why I ended up making it configurable.

Maybe as a follow-up we should look at sizes and see if we can find a good optimum. I can do that.

mattfarina:
@bacongobbler updated based on your feedback. Thanks.

bacongobbler (approved):

bacongobbler:
:shipit:

Issue #5958, opened 2019-06-30
[feature] helm install --wait does not wait for storageclass


staticdev:
Output of `helm version`: 2.14.1

Output of `kubectl version`: Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.0", GitCommit:"e8462b5b5dc2584fdcd18e6bcfe9f1e4d970a529", GitTreeState:"clean", BuildDate:"2019-06-19T16:40:16Z", GoVersion:"go1.12.5", Compiler:"gc", Platform:"linux/amd64"}

Cloud Provider/Platform (AKS, GKE, Minikube etc.): Kubeadm

To reproduce just try the script:
```sh
helm repo add storageos https://charts.storageos.com
helm install storageos/storageos-operator --namespace storageos-operator --wait
kubectl patch storageclass fast -p '{"metadata": {"annotations":{"storageclass.kubernetes.io/is-default-class":"true"}}}'
```

It installs a apiVersion: apps/v1 Deployment.
https://github.com/storageos/charts/blob/master/stable/storageos-operator/templates/operator.yaml

According to the documentation, the --wait has a limited scope:

>     Waits until all Pods are in a ready state, PVCs are bound, Deployments have minimum (Desired minus maxUnavailable) Pods in ready state and Services have an IP address (and Ingress if a LoadBalancer) before marking the release as successful

bacongobbler:
I'm missing the context here. Is this an ask for the `--wait` flag to wait for storage classes to become available?

@thomastaylor312 what do you think about this?

staticdev:
@bacongobbler Exactly! I am running into problems for automating helm installations that create storageclasses. Because I set --wait and can't patch the class for example to be default after this scripted installation.

thomastaylor312:
@staticdev to clarify, is this needed because your chart _also_ has a PVC that is using the newly created storage class or is it creating the storage class as a pre-req before running other steps/releases down the road?

staticdev:
@thomastaylor312 second option: it creates a storage class that is needed for next steps.

thomastaylor312:
Hmmm...that is interesting. I'll take a look at possibly adding it. To be honest, I might add it just to Helm 3 because the logic has been completely cleaned up there and is still a bit of a mess in Helm 2

staticdev:
@thomastaylor312 I see no problem at all. I am pretty enthusiastic about Helm 3 and looking forward for stable version ASAP =)

[...]

Issue #5871, opened 2019-06-11
Proposal: Manage CRDs


mattfarina:
I have to give credit to @prydonius for this idea. I'm just writing it up here. Can do a proposal to community if this needs to move forward. Not entirely sure what is feasible yet.

**The idea:**

Have a `crd` directory alongside the `templates` directory. Helm would then check for missing CRDs and attempt to install them. If it cannot install them and they are missing it would stop, explain why it can't install them (e.g., permissions), and provide some details (e.g., someone with permissions needs to install the CRDs).

This is because there is common wrapper logic needed around CRDs for each of them that is about the same for this. First, check if it's installed and if it's not then have it there with the crd install hook.

One unknown is handing changes in the CRD content at the same `apiVersion`. This happens in practice and has many potential problems. Can we detect a change and throw an error to require human intervention?

Once this is in place the Helm dependency system can be used to handle CRDs that are dependent upon one another. Each can be in their own chart with their own controllers.

mattfarina:
cc @technosophos @adamreese @bacongobbler @hickeyma @michelleN @viglesiasce @thomastaylor312 @SlickNik

bacongobbler:
I agree that there is a common wrapper logic being used for all CRDs in Helm 2 recently.

I would love to see a practical example! It seems like it's an idea worth investigating. I'm having a hard time conceptualizing this, so having a practical example to play with and understand the solution might be a good way forward to see if this is feasible or not (as well as provide feedback on the design).

technosophos:
As I see it, we have four major problems to sort out with CRDs:

- Install order: CRDs have to be installed in a separate operation from any _usage of_ (instances of) that CRD. 
- Compatibility detection: If a CRD exists, and a request to install the CRD comes in, but the api versions are different, what do we do?
- Update process: CRDs are possibly global, while most other kinds that appear in Helm charts are not. So under what conditions can a chart that contains a CRD upgrade the CRD?
- Deletion of a CRD: When a CRD is deleted, all of its instances are deleted. And CRDs are global. Under what conditions can Helm delete a CRD (clearly, it's not by default)

As far as I can tell, the above makes it easier to solve the first point. But is it in any way helping or hindering with the others? I don't want to take a shot at solving just one prong, only to discover that that solution prevented us from solving the others.

[...]

PR #6212, opened 2019-08-12
refs #5958 : Added Wait on Storage Class

waveywaves:

wrt https://github.com/helm/helm/issues/5958 
Allows the user to wait on availability of storage classes.

Tested it out with

```
helm install efs kubernetes-charts/efs-provisioner --wait
```

**If applicable**:

### DIFF
--- a/pkg/kube/wait.go
+++ b/pkg/kube/wait.go
@@ -27,6 +27,7 @@ import (
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	extensionsv1beta1 "k8s.io/api/extensions/v1beta1"
+	storagev1 "k8s.io/api/storage/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -114,6 +115,14 @@ func (w *waiter) waitForResources(created ResourceList) error {
 				if !w.deploymentReady(newReplicaSet, currentDeployment) {
 					return false, nil
 				}
+			case *storagev1.StorageClass:
+				sc, err := w.c.StorageV1().StorageClasses().Get(value.Name, metav1.GetOptions{})
+				if err != nil {
+					return false, err
+				}
+				if !w.storageClassReady(sc) {
+					return false, nil
+				}
 			case *corev1.PersistentVolumeClaim:
 				claim, err := w.c.CoreV1().PersistentVolumeClaims(value.Namespace).Get(value.Name, metav1.GetOptions{})
 				if err != nil {
@@ -155,6 +164,7 @@ func (w *waiter) waitForResources(created ResourceList) error {
 				ok, err = w.podsReadyForObject(value.Namespace, value)
 			case *appsv1.ReplicaSet:
 				ok, err = w.podsReadyForObject(value.Namespace, value)
+
 			}
 			if !ok || err != nil {
 				return false, err
@@ -216,6 +226,14 @@ func (w *waiter) serviceReady(s *corev1.Service) bool {
 func (w *waiter) volumeReady(v *corev1.PersistentVolumeClaim) bool {
 	if v.Status.Phase != corev1.ClaimBound {
 		w.log("PersistentVolumeClaim is not bound: %s/%s", v.GetNamespace(), v.GetName())
+		return false
+	}
+	return true
+}
+
+func (w *waiter) storageClassReady(sc *storagev1.StorageClass) bool {
+	if len(sc.Provisioner) <= 0 {
+		w.log("Provisioner is not allocated: %s/%s", sc.GetNamespace(), sc.GetName())
 		return false
 	}
 	return true


waveywaves:
cc @bacongobbler @thomastaylor312 @staticdev

pkg/kube/wait.go
```diff
@@ -221,6 +231,14 @@ func (w *waiter) volumeReady(v *corev1.PersistentVolumeClaim) bool {
 	return true
 }
 
+func (w *waiter) storageClassReady(sc *storagev1beta1.StorageClass) bool {
+	if len(sc.Provisioner) <= 0 {
```
thomastaylor312: I was looking at the reference docs and documentation for storage classes and I am not sure this is the best field to be keying off of. It looks like `provisioner` is a required field anyway and so this is more of a validation issue. However, I can't seem to figure out what field would need to be used. Is there something better we can key off of to solve the problem as described in #5958?

pkg/kube/wait.go
```diff
@@ -114,6 +115,14 @@ func (w *waiter) waitForResources(created ResourceList) error {
 				if !w.deploymentReady(newReplicaSet, currentDeployment) {
 					return false, nil
 				}
+			case *storagev1beta1.StorageClass:
```
thomastaylor312: The storage APIs are now at `v1`, so we should make sure to have a case for that api version as well and also just use the v1 storage client to fetch both of them (like what we do with deployments)

thomastaylor312 (changes requested):
A few changes needed and a discussion on how to tell if a storage class is "ready." Also, can you make sure to add storage class to the `--wait` flag documentation as well?

pkg/kube/wait.go
waveywaves: I have updated it to stroragev1

pkg/kube/wait.go
waveywaves: In the given issue the problem is taking place with CRDs not being instantiated properly. And is the issue with the APIs which the storageos operator is trying to expose. The storageclasses defined over there are provided by the storageos apis and not something supported here. Here we are mainly trying to listen on StorageClasses defined in Kubernetes.

waveywaves: I have discussed with a few folks working in storage, more specifically Ceph. It seems that the check the health of the storage class is based on if the provisioner is up. In #5958 @staticdev is trying to install an operator which upon installing exposes the storageos.io/v1

waveywaves:
Added storageclass to the `--wait` documentation(s) https://github.com/helm/helm-www/pull/260

pkg/kube/wait.go
thomastaylor312: If I am understanding it correctly, this is trying to watch for storage classes that are created by the operator, which are completely out of scope from Helm itself. This feels like something that should be done in a `post-install` hook rather than within the wait logic itself

waveywaves:
Build is failing at
```
127.0.0.1 - - [14/Aug/2019:22:49:51 +0000] "HEAD /v2/testrepo/whodis/manifests/9.9.9 HTTP/1.1" 404 95 "" "Go-http-client/1.1"
time="2019-08-14T22:49:51.735955824Z" level=info msg="authorized request" go.version=go1.12.8 http.request.host="localhost:33475" http.request.id=52cd629d-c88a-4cf4-a809-17c90d114ce1 http.request.method=HEAD http.request.remoteaddr="127.0.0.1:36264" http.request.uri=/v2/testrepo/testchart/manifests/1.2.3 http.request.useragent=Go-http-client/1.1 vars.name=testrepo/testchart vars.reference=1.2.3
time="2019-08-14T22:49:51.73632909Z" level=info msg="response completed" go.version=go1.12.8 http.request.host="localhost:33475" http.request.id=52cd629d-c88a-4cf4-a809-17c90d114ce1 http.request.method=HEAD http.request.remoteaddr="127.0.0.1:36264" http.request.uri=/v2/testrepo/testchart/manifests/1.2.3 http.request.useragent=Go-http-client/1.1 http.response.contenttype=application/vnd.oci.image.manifest.v1+json http.response.duration=66.486704ms http.response.status=200 http.response.written=352
127.0.0.1 - - [14/Aug/2019:22:49:51 +0000] "HEAD /v2/testrepo/testchart/manifests/1.2.3 HTTP/1.1" 200 352 "" "Go-http-client/1.1"
--- FAIL: TestParseReference (0.00s)
    reference_test.go:51: 
        	Error Trace:	reference_test.go:51
        	Error:      	Received unexpected error:
        	            	parse dummy://mychart:1.5.0: invalid port ":1.5.0" after host
        	Test:       	TestParseReference
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x120f1e7]
```

thomastaylor312:
@waveywaves I have an open PR to fix that. It is related to the go 1.12.8 security release that broke some things

pkg/kube/wait.go
waveywaves: This PR watches StorageClasses defined via Kubernetes by default. 

The listening on CustomResources such as StorageOS StorageClasses is something that would be part of a larger CRD integration Story as discussed here https://github.com/helm/helm/issues/5871 is what I feel. This PR would not be catering to that, but the Users should be able to listen on successful creation of StorageClasses nevertheless.

waveywaves:
> @waveywaves I have an open PR to fix that. It is related to the go 1.12.8 security release that broke some things

Can you point me to it. I want to see what is wrong :1234:

thomastaylor312:
See #6228

pkg/kube/wait.go
thomastaylor312: So I don't think it will work as defined in this PR though. The wait logic only waits for resources that were defined as part of the release and not the default resources that are already there. I am also still not sure what constitutes a storage class being ready based on all of my past experience and current reading of the docs. As I stated before, according to the spec, `provisioner` is a required field, so basically all we are doing is retrieving the object from the k8s API to see if we can do it. Is there anything that Helm itself can "wait" on that isn't a specific provider of a storage class?

It may be valuable for us to sync up over a call so I make sure I am not missing anything

waveywaves:
Closed this based on a discussion with @thomastaylor312

PR #6217, opened 2019-08-13
fix(test): restore --cleanup

bacongobbler:
#6054 removed `--cleanup`. This restores that functionality.

One thing to test would be `test-success` hooks that define a hook deletion policy, since those will now be cleaned up prior to `--cleanup`. We want to make sure Helm doesn't error out in those cases. Giving that a spin now; figured I'd PR the fix first though.

### DIFF
--- a/pkg/action/release_testing.go
+++ b/pkg/action/release_testing.go
@@ -19,6 +19,7 @@ package action
 import (
 	"bytes"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
@@ -61,18 +62,20 @@ func (r *ReleaseTesting) Run(name string) error {
 	}
 
 	if r.Cleanup {
+		var manifestsToDelete strings.Builder
 		for _, h := range rel.Hooks {
 			for _, e := range h.Events {
 				if e == release.HookTest {
-					hookResource, err := r.cfg.KubeClient.Build(bytes.NewBufferString(h.Manifest))
-					if err != nil {
-						return errors.Wrapf(err, "unable to build kubernetes object for %v hook %s", h, h.Path)
-					}
-					if _, errs := r.cfg.KubeClient.Delete(hookResource); errs != nil {
-						return fmt.Errorf("unable to delete kubernetes object for %v hook %s: %s", h, h.Path, joinErrors(errs))
-					}
+					fmt.Fprintf(&manifestsToDelete, "\n---\n%s", h.Manifest)
 				}
 			}
+		}
+		hooks, err := r.cfg.KubeClient.Build(bytes.NewBufferString(manifestsToDelete.String()))
+		if err != nil {
+			return fmt.Errorf("unable to build test hooks: %v", err)
+		}
+		if _, errs := r.cfg.KubeClient.Delete(hooks); errs != nil {
+			return fmt.Errorf("unable to delete test hooks: %v", joinErrors(errs))
 		}
 	}
 
--- a/pkg/action/release_testing.go
+++ b/pkg/action/release_testing.go
@@ -66,10 +66,10 @@ func (r *ReleaseTesting) Run(name string) error {
 				if e == release.HookTest {
 					hookResource, err := r.cfg.KubeClient.Build(bytes.NewBufferString(h.Manifest))
 					if err != nil {
-						return errors.Wrapf(err, "unable to build kubernetes object for %s hook %s", h, h.Path)
+						return errors.Wrapf(err, "unable to build kubernetes object for %v hook %s", h, h.Path)
 					}
 					if _, errs := r.cfg.KubeClient.Delete(hookResource); errs != nil {
-						return fmt.Errorf("unable to delete kubernetes object for %s hook %s: %s", h, h.Path, joinErrors(errs))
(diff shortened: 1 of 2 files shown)


bacongobbler:
Seems to work when dealing with legacy test cases:

```
><> ./bin/helm install vault incubator/vault
NAME: vault
LAST DEPLOYED: 2019-08-13 12:00:21.270768616 -0700 PDT m=+0.699809736
NAMESPACE: default
STATUS: deployed

NOTES:
  export POD_NAME=$(kubectl get pods --namespace default -l "app=vault" -o jsonpath="{.items[0].metadata.name}")
  echo "Use http://127.0.0.1:8200 as the Vault address after forwarding."
  kubectl port-forward --namespace default  $POD_NAME 8200:8200

><> ./bin/helm test run vault --cleanup
Pod vault-vault-status-test pending
Pod vault-vault-status-test pending
Pod vault-vault-status-test succeeded

><> k get po
NAME                          READY   STATUS    RESTARTS   AGE
vault-vault-cdb859c4c-6pg4z   1/1     Running   0          20s
vault-vault-cdb859c4c-bh4k5   0/1     Running   0          20s
vault-vault-cdb859c4c-wtcwl   0/1     Running   0          20s
```

bacongobbler:
Now if I add a hook delete policy, upgrade the chart and test the same thing:

```
><> ./bin/helm fetch incubator/vault --untar
><> cat vault/templates/tests/test-vault-status.yaml 
apiVersion: v1
kind: Pod
metadata:
  name: "{{ .Release.Name }}-vault-status-test"
  annotations:
    "helm.sh/hook": test-success
    "helm.sh/hook-delete-policy": hook-succeeded
spec:
  containers:
  - name: {{ .Release.Name }}-vault-status-test
    image: "{{ .Values.image.repository }}:{{ default .Chart.AppVersion .Values.image.tag }}"
    env:
      - name: VAULT_ADDR
        value: http://{{ template "vault.fullname" . }}.{{ .Release.Namespace }}:8200
    command: ["sh", "-c", "vault status"]
  restartPolicy: Never
```

I seem to hit an error during cleanup

```
><> ./bin/helm upgrade vault ./vault/
><> ./bin/helm test run vault
Pod vault-vault-status-test pending
Pod vault-vault-status-test pending
Pod vault-vault-status-test succeeded
Error:
```

Looking into that now. That looks like an issue with the hook delete policy than with the PR here.

bacongobbler:
Ah... Same issue as in #6191. I did a quick codebase search and found no additional places where this behaviour may crop up so we should be good now.

bacongobbler:
Things appear to work as expected though with a deletion policy set and `--cleanup` set! No errors are produced.

```
><> ./bin/helm fetch incubator/vault --untar

><> cat vault/templates/tests/test-vault-status.yaml 
apiVersion: v1
kind: Pod
metadata:
  name: "{{ .Release.Name }}-vault-status-test"
  annotations:
    "helm.sh/hook": test-success
    "helm.sh/hook-delete-policy": hook-succeeded
spec:
  containers:
  - name: {{ .Release.Name }}-vault-status-test
    image: "{{ .Values.image.repository }}:{{ default .Chart.AppVersion .Values.image.tag }}"
    env:
      - name: VAULT_ADDR
        value: http://{{ template "vault.fullname" . }}.{{ .Release.Namespace }}:8200
    command: ["sh", "-c", "vault status"]
  restartPolicy: Never

><> ./bin/helm upgrade vault ./vault/

><> ./bin/helm test run vault
Pod vault-vault-status-test pending
Pod vault-vault-status-test pending
Pod vault-vault-status-test succeeded

><> ./bin/helm test run vault --cleanup
Pod vault-vault-status-test pending
Pod vault-vault-status-test pending
Pod vault-vault-status-test succeeded
```

pkg/action/release_testing.go
```diff
@@ -58,5 +60,21 @@ func (r *ReleaseTesting) Run(name string) error {
 		return err
 	}
 
+	if r.Cleanup {
+		for _, h := range rel.Hooks {
+			for _, e := range h.Events {
+				if e == release.HookTest {
+					hookResource, err := r.cfg.KubeClient.Build(bytes.NewBufferString(h.Manifest))
```
thomastaylor312: Non-blocking optimization nit: You could glue all the manifests together and then have a single `client.Delete` call. Totally non blocking though

thomastaylor312 (approved):
LGTM

pkg/action/release_testing.go
bacongobbler: is there a good example on how to do that? I was afraid that injecting a newline between each manifest may break in certain cases but I'd love to lessen the number of calls to k8s if possible.

thomastaylor312: We do it here: https://github.com/helm/helm/blob/63b751ded75c2a48045e896ae3e6418c370038a1/pkg/action/uninstall.go#L183-L192

thomastaylor312: Technically speaking though, I think it does end up being the same amount of delete calls, but it does lessen the number of `Build` calls so that you are only deserializing once

bacongobbler: thank you for the example! I'll go do that now

PR #6228, opened 2019-08-14
fix(registry): Updates registry to handle go 1.12.8 changes

thomastaylor312:
Go 1.12.8 introduced some breaking fixes (see [here](https://github.com/golang/go/commit/3226f2d492963d361af9dfc6714ef141ba606713)) for a CVE. This broke the way we were doing registry reference parsing. This removes the call to the containerd libraries in favor of our own parsing and adds additional unit tests

### DIFF
--- a/internal/experimental/registry/reference.go
+++ b/internal/experimental/registry/reference.go
@@ -19,22 +19,25 @@ package registry // import "helm.sh/helm/internal/experimental/registry"
 import (
 	"errors"
 	"fmt"
+	"net/url"
 	"regexp"
 	"strings"
-
-	"github.com/containerd/containerd/reference"
 )
 
 var (
-	validPortRegEx   = regexp.MustCompile(`^([1-9]\d{0,3}|0|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5])$`) // adapted from https://stackoverflow.com/a/12968117
-	errEmptyRepo     = errors.New("parsed repo was empty")
-	errTooManyColons = errors.New("ref may only contain a single colon character (:) unless specifying a port number")
+	validPortRegEx = regexp.MustCompile(`^([1-9]\d{0,3}|0|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5])$`) // adapted from https://stackoverflow.com/a/12968117
+	// TODO: Currently we don't support digests, so we are only splitting on the
+	// colon. However, when we add support for digests, we'll need to use the
+	// regexp anyway to split on both colons and @, so leaving it like this for
+	// now
+	referenceDelimiter = regexp.MustCompile(`[:]`)
+	errEmptyRepo       = errors.New("parsed repo was empty")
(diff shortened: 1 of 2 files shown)


internal/experimental/registry/reference.go
```diff
@@ -19,45 +19,56 @@ package registry // import "helm.sh/helm/internal/experimental/registry"
...
-
-	"github.com/containerd/containerd/reference"
 )
 
 var (
-	validPortRegEx   = regexp.MustCompile(`^([1-9]\d{0,3}|0|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5])$`) // adapted from https://stackoverflow.com/a/12968117
-	errEmptyRepo     = errors.New("parsed repo was empty")
-	errTooManyColons = errors.New("ref may only contain a single colon character (:) unless specifying a port number")
+	validPortRegEx     = regexp.MustCompile(`^([1-9]\d{0,3}|0|[1-5][0-9]{4}|6[0-4][0-9]{3}|65[0-4][0-9]{2}|655[0-2][0-9]|6553[0-5])$`) // adapted from https://stackoverflow.com/a/12968117
+	referenceDelimiter = regexp.MustCompile(`[:@]`)
```
thomastaylor312: @jdolitsky So the containerd spec supports `@` and I added it in without checking. If you don't think we need it, I will remove and only split on the colon

jdolitsky: So after trying out your change I think we should remove it. It essentially is replacing the "@" with a ":":
```
$ bin/helm chart save minio/ localhost@5000/myrepo/minio@2.5.11
ref:     localhost:5000/myrepo/minio:2.5.11
digest:  f0f2c8eebfb95700b77e435810106bf780a49d2a07db5858097b4d6ecaaef152
size:    12.8 KiB
name:    minio
version: 2.5.11
2.5.11: saved
```
We may as well only support the ":" unless somebody comes along w a usecase for it

jdolitsky (dismissed):
overall looks great, thanks for addressing this. please see one comment above

internal/experimental/registry/reference.go
mattfarina: If I use docker/containerd I can use `:` as the delimiter for tags and `@` as the delimiter for digests (e.g., sha:abc123...). This is documented at https://docs.docker.com/engine/reference/commandline/pull/.

In the parsing docs it notes:
> In practice, we will see this broken down into the following formats:
> // 1. \<tag>
> // 2. \<tag>@\<digest spec>
> // 3. @\<digest spec>

It doesn't look like this code is handling references yet. If that is the case we can skip the `@` but we should file an issue against v3 to add digest support. If digests should be handled I would suggest some tests for that as a followup.

The @ and : have expected user experience for container users. I would suggest we follow the expected experience.

thomastaylor312: Switched back to just parsing the `:`, but I left the regex logic in there so it will be easier to add in the `@` later

thomastaylor312:
Ok @jdolitsky, comments addressed and ready for re-review

jdolitsky:
This lgtm!

mattfarina (approved):

PR #6229, opened 2019-08-15
omit deleted for release info when not set

### LINKED ISSUE #6163: helm3: deleted in release info has strange default
I just did a `helm install wordpress stable/wordpress` with helm3-alpha2.
Then I looked at the release secret `k get secret wordpress.v1 -o jsonpath="{.data.release}" | base64 -D | base64 -D | gzip -d | jq`

`
{
  "name": "wordpress",
  "info": {
    "first_deployed": "2019-08-05T13:52:08.317608+02:00",
    "last_deployed": "2019-08-05T13:52:08.317608+02:00",
    "deleted": "0001-01-01T00:00:00Z",
    "Description": "Install complete",
    "status": "deployed", ...
`

As you can see the deleted timestamp has some strange default value. I think it should be an empty string or something like that.

Best regards,
Tobi

nbrink91:
This bug fix omits deleted from release info when it is not set. This pull request closes #6163.

N/A

**If applicable**:

Signed-off-by: Nicholas Brink <nbrink7@gmail.com>

### DIFF
--- a/cmd/helm/testdata/output/status.json
+++ b/cmd/helm/testdata/output/status.json
@@ -1,1 +1,1 @@
-{"name":"flummoxed-chickadee","info":{"first_deployed":"0001-01-01T00:00:00Z","last_deployed":"2016-01-16T00:00:00Z","deleted":"0001-01-01T00:00:00Z","status":"deployed","notes":"release notes"},"namespace":"default"}
+{"name":"flummoxed-chickadee","info":{"first_deployed":"0001-01-01T00:00:00Z","last_deployed":"2016-01-16T00:00:00Z","status":"deployed","notes":"release notes"},"namespace":"default"}
--- a/cmd/helm/testdata/output/status.yaml
+++ b/cmd/helm/testdata/output/status.yaml
@@ -1,5 +1,4 @@
 info:
-  deleted: "0001-01-01T00:00:00Z"
   first_deployed: "0001-01-01T00:00:00Z"
   last_deployed: "2016-01-16T00:00:00Z"
   resources: |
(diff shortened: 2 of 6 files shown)


mattfarina:
@nbrink91 When Go 1.12.8 came out it included a security fix that broke some tests. Numerous projects have experienced this problem. This has now been fixed. Can you please rebase off the `dev-v3` branch to pick up the fix.

bacongobbler:
After taking a closer look, this appears to be a bug where the `omitempty` annotation is not picking up on the zero value for `time.Time`, indicating a bug in `encoding/json`.

https://github.com/golang/go/issues/11939

Using a `*time.Time` is one workaround suggested in that thread, but it sounds like this is a lower level issue, and I'd rather avoid working around this issue so we don't have to deal with null time values.

What do you think, @nbrink91?

thomastaylor312:
@bacongobbler So I came up with [this option using custom marshaling](https://github.com/helm/helm/compare/dev-v3...thomastaylor312:feat/timestamp_hell?expand=1) and avoids a time pointer. The problem with it is I couldn't get it to read current releases and that it only handles deleted. There are other empty times in the release object (like test run times) that also would need to be handled.

The other option is to create our own custom time wrapper type with Marshaling and use that everywhere, but I also had problems marshaling existing objects when I tried it.

Both have their disadvantages, but it really does result in some weird zero value JSON output that I think is probably Not Good™ as output we return. I'm interested in your thoughts about whether I should try our own custom time type instead

bacongobbler:
That could work @thomastaylor312. `pkg/action` has a `TimeStamper` object which we could pull out as a shareable object and add `MarshalJSON`, `MarshalText` etc. to it.

https://github.com/helm/helm/blob/3edad39e08ac514b6c6b7386c03fc44464cb159e/pkg/action/action.go#L43

thomastaylor312:
@bacongobbler it would have to be `time.Time` but I would pull the existing `Timestamper` out into the new `time` package I'd create

Issue #6079, opened 2019-07-25
Helm v3 - linting multiple charts depends on the order of the charts


andytom:
Output of `helm version`:
```
version.BuildInfo{Version:"v3.0.0-alpha.2", GitCommit:"97e7461e41455e58d89b4d7d192fed5352001d44", GitTreeState:"clean", GoVersion:"go1.12.7"}
```

Output of `kubectl version`:

N/A

Cloud Provider/Platform (AKS, GKE, Minikube etc.):  

N/A

----
When linting multiple charts using helm3 the order that the charts are passed in matters. The following was done using the, at time of writing, latest commit on the master branch of the charts repo https://github.com/helm/charts/commit/8fb57b472e7a9d5117734566146f272ebc071aa2.

```
$ helm3 lint stable/atlantis stable/apm-server
2 chart(s) linted, 0 chart(s) failed
```

```
$ helm3 lint stable/apm-server stable/atlantis
Error: 2 chart(s) linted, 1 chart(s) failed
	%!s(<nil>)
	[ERROR] templates/: render error at (atlantis/templates/configmap-config.yaml:13:27): wrong type for value; expected string; got map[string]interface {}
```

Using the latest version of helm 2 (v2.14.2) this works as expected
```
$ helm lint stable/apm-server stable/atlantis
==> Linting stable/apm-server
Lint OK

==> Linting stable/atlantis
Lint OK

2 chart(s) linted, no failures
```

```
$ helm lint stable/atlantis stable/apm-server
==> Linting stable/atlantis
Lint OK

==> Linting stable/apm-server
Lint OK

2 chart(s) linted, no failures
```

waveywaves:
Hello Team ! I would like to start working on this bug. 
Anything to keep in mind before I get started ?

karuppiah7890:
@waveywaves you can start off and comment here if you have any questions while implementing 😊

waveywaves:
Cool @karuppiah7890 Thanks for that :)

waveywaves:
After reproducing the issue I have been trying to debug this for the past few days.
```
➜  helm3test helm lint ./apm-server ./atlantis
Error: 2 chart(s) linted, 1 chart(s) failed
        %!s(<nil>)
        [ERROR] templates/: template: atlantis/templates/configmap-config.yaml:13:27: executing "atlantis/templates/configmap-config.yaml" at <4>: wrong type for value; expected string; got map[string]interface
{}

[...]

PR #6241, opened 2019-08-16
Clone the vals map for every path to avoid mutation

waveywaves:
This PR fixes https://github.com/helm/helm/issues/6079 
`vals` was getting mutated when multiple charts were provided for linting.
Cloning `vals` for every `path` sees to it that every `path` has it's own unmutated set of `vals`.

**If applicable**:

### DIFF
--- a/pkg/action/lint.go
+++ b/pkg/action/lint.go
@@ -59,15 +59,22 @@ func (l *Lint) Run(paths []string, vals map[string]interface{}) *LintResult {
 
 	result := &LintResult{}
 	for _, path := range paths {
-		if linter, err := lintChart(path, vals, l.Namespace, l.Strict); err != nil {
+		linter, err := lintChart(path, vals, l.Namespace, l.Strict)
+		if err != nil {
 			if err == errLintNoChart {
+				result.Errors = append(result.Errors, err)
+			}
+			if linter.HighestSeverity >= lowestTolerance {
 				result.Errors = append(result.Errors, err)
 			}
 		} else {
 			result.Messages = append(result.Messages, linter.Messages...)
 			result.TotalChartsLinted++
-			if linter.HighestSeverity >= lowestTolerance {
-				result.Errors = append(result.Errors, err)
+			for _, msg := range linter.Messages {
+				if msg.Severity == support.ErrorSev {
+					result.Errors = append(result.Errors, msg.Err)
+					result.Messages = append(result.Messages, msg)
+				}
 			}
 		}
 	}
--- a/pkg/action/lint.go
+++ b/pkg/action/lint.go
@@ -77,6 +77,10 @@ func (l *Lint) Run(paths []string, vals map[string]interface{}) *LintResult {
 func lintChart(path string, vals map[string]interface{}, namespace string, strict bool) (support.Linter, error) {
 	var chartPath string
 	linter := support.Linter{}
+	currentVals := make(map[string]interface{}, len(vals))
+	for key, value := range vals {
+		currentVals[key] = value
+	}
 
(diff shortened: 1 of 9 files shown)


waveywaves:
cc @karuppiah7890 @adamreese @mattfarina

karuppiah7890:
@waveywaves could you add a failing test for the bug and then pass it with the current code? Tests would give confidence on the robustness of the features

waveywaves:
@karuppiah7890 Have updated the PR. Do take a look and let me know if I need to make any changes.

karuppiah7890:
@waveywaves I ran the test without the fix and it passed, may be use existing chart / add some chart that when linted after another one breaks the linting due to the issue? 

For example, in first chart, use a chart with a `config` value which is yaml and has `toYaml` used in the templates to use the `config`, in second chart, use `config` value which is a string and used directly with `.Values.config` with no `toYaml` or anything. The example is a subset of the of the charts mentioned in the issue which caused the bug.

karuppiah7890:
Also, do sign off the commit https://github.com/helm/helm/pull/6241/checks?check_run_id=195950295 for the Developer Certificate of Origin check

waveywaves:
@karuppiah7890 I have added the testcharts against which I was able to have the test fail before the changes. And I have updated the test as well. Let me know if I need to make any changes to this.

karuppiah7890:
I just glanced it. It's awesome 😁. I see you have fixed a Fixme comment too, I have no idea about that, will have to read and understand that. Other than that LGTM. And I can't merge PRs, so cc'ing people. cc @bacongobbler

pkg/action/lint.go
```diff
@@ -59,15 +59,22 @@ func (l *Lint) Run(paths []string, vals map[string]interface{}) *LintResult {
 
 	result := &LintResult{}
 	for _, path := range paths {
-		if linter, err := lintChart(path, vals, l.Namespace, l.Strict); err != nil {
+		linter, err := lintChart(path, vals, l.Namespace, l.Strict)
+		if err != nil {
 			if err == errLintNoChart {
 				result.Errors = append(result.Errors, err)
 			}
+			if linter.HighestSeverity >= lowestTolerance {
```
thomastaylor312: These changes feel out of scope for this PR because it doesn't seem to be related to the values map mutation. Please let me know if I am wrong

thomastaylor312 (commented):
This is looking good. Thanks for adding a good unit test for this! Just one comment around whether a block of changes is needed for this PR

pkg/action/lint.go
waveywaves: None of the `Messages` which are errors are being appended to `Result.Errors` which shouldn't be the case.

waveywaves: While writing the test, I noticed that there was <nil> getting appended to Result.Errors. This was creating an issue because, Result.Errors was supposed tohave a length of zero, instead it had a length of 1, hence failing my test even on the right code.

When running lintChart on a chart which passes the lint, it would return <nil> for err which is expected. But this lint was being appended to result.Errors which isn't the expected behaviour. Do correct me if I'm wrong.

thomastaylor312: OK, that make sense then. So this was a different bug you found along the way and fixed also

thomastaylor312: @waveywaves could you write a unit test to make sure we can validate that in the future?

waveywaves: @thomastaylor312 I have updated the PR, can you check and let me know.

thomastaylor312 (approved):
Thanks for the great work on this!

PR #6243, opened 2019-08-16
feat: Add support for a crds/ directory

technosophos:
**What this PR does / why we need it**: This PR adds support for a `crds/` directory into which you may put your CRDs without fear.

Closes #587

This PR adds support for a `crds/` directory in a chart. During chart _install_ (and ONLY during install), it reads the contents of that directory and attempts to install those files before it renders the rest of the chart.

In a sense, it is a pre-install. It provides a way to add CRDs before the rest of the chart so that things that need CRDs can install them before instances of those CRDs can be created by the chart.

If a CRD is already found in the cluster, Helm will skip installing that CRD and will log this to the debug log. It will neither error out nor see if there are any differences between its version and the existing version.

This behavior can be turned off with `--skip-crds`.

Things this feature does NOT do:

- CRDs are NOT deleted with the chart. Automatic deletion of CRDs is a serious no-no, as deleting CRDs is a cluster-wide operation that deletes every single instance of that CRD. Operators must delete CRDs themselves if they want them to go away.
- CRDs are not upgraded during `helm install` or `helm upgrade`. There is no accepted practice for managing CRD upgrades (again, because they are a global resources, and an upgrade can break things in interesting ways), so we decided not to handle this case. It is up to the operator.
- It does not output CRDs as part of `helm template` or `helm OP --dry-run`. We could add this in the future, or add a separate command for printing CRDs. For example, we could create a `helm crd some_chart` may print the CRDs for that chart. *Feedback welcome.*
- CRDs are currently NOT templated. The reason for this is that we would have to run the template engine twice, but the template context would be different for each. So for our first pass, we are seeing if we can avoid using templates in CRDs. In evaluating public usage of CRDs, we believe this should be just fine.
- The `crd-install` hook is still NOT added back to Helm 3. This is an alternative that we believe is much more robust and safe. (I was the author of `crd-install`, and I am very dissatisfied with it.)

## To Get Started

1. Create a new chart
2. Add a directory in adjacent to `templates/` and `charts/` named `crds/`.
3. Create a YAML or JSON file there whose contents are a single CRD.
4. Add resources in `templates/` that use those CRDs
5. `helm install test-crds $YOURNEWCHART`
6. Profit

You should see your CRDs with `kubectl get crds`. The chart should succeed even if it references those CRDs.

**If applicable**:

### DIFF
--- a/cmd/helm/install.go
+++ b/cmd/helm/install.go
@@ -131,6 +131,7 @@ func addInstallFlags(f *pflag.FlagSet, client *action.Install, valueOpts *values
 	f.BoolVar(&client.Devel, "devel", false, "use development versions, too. Equivalent to version '>0.0.0-0'. If --version is set, this is ignored.")
 	f.BoolVar(&client.DependencyUpdate, "dependency-update", false, "run helm dependency update before installing the chart")
 	f.BoolVar(&client.Atomic, "atomic", false, "if set, installation process purges chart on fail. The --wait flag will be set automatically if --atomic is used")
+	f.BoolVar(&client.SkipCRDs, "skip-crds", false, "if set, no CRDs will be installed. By default, CRDs are installed if not already present.")
 	addValueOptionsFlags(f, valueOpts)
 	addChartPathOptionsFlags(f, &client.ChartPathOptions)
 }
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -29,6 +29,7 @@ import (
 
 	"github.com/Masterminds/sprig"
 	"github.com/pkg/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 
 	"helm.sh/helm/pkg/chart"
 	"helm.sh/helm/pkg/chartutil"
@@ -80,8 +81,10 @@ type Install struct {
 	NameTemplate     string
 	OutputDir        string
 	Atomic           bool
+	SkipCRDs         bool
 }
 
+// ChartPathOptions captures common options used for controlling chart paths
 type ChartPathOptions struct {
 	CaFile   string // --ca-file
 	CertFile string // --cert-file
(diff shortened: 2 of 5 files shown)


mattfarina (approved):
As follow-ups...

- We need to finish creating the legacy plugin
- We should experiment with a crd plugin to diff and update crds. A plugin or many of them is a good way to experiment future capabilities

thomastaylor312 (approved):
I paired with @technosophos on this and we ran it through its paces, so this is an LGTM from me

mattfarina:
Merging as @technosophos is currently unavailable.

hickeyma:
@technosophos @mattfarina @thomastaylor312 Should we add a page to the documentation for CRD support and strategy in Helm 3? Pretty much pull the description from here.

thomastaylor312:
@hickeyma Yep, that would be a good idea

munnerz:
I've opened https://github.com/helm/helm/issues/7735 to discuss some of the assumptions made around upgrading/patching CRDs in this PR - I'd appreciate feedback 😄

Notably:

> There is no accepted practice for managing CRD upgrades (again, because they are a global resources, and an upgrade can break things in interesting ways), so we decided not to handle this case. It is up to the operator.

I disagree with - there _is_ an accepted practice, and it is up to CRD authors to follow this closely. (although it's probably best to centralise discussion in #7735) 😄

pranavmarla:
Calling out this particular section of the original post for my future reference:
> This PR adds support for a `crds/` directory in a chart. **During chart install (and ONLY during install)**, Helm reads the contents of that directory and attempts to install those files before it renders the rest of the chart.

This behaviour is still true today. i.e. If a new Helm chart version adds/introduces a new CRD and you then attempt to upgrade your existing chart to that new version via `helm upgrade`, Helm will NOT add/install that new CRD (and the chart upgrade will fail)! Apparently Helm will only add/install the new CRD if you fully uninstall the existing chart and do a fresh install via `helm install`

Issue #6226, opened 2019-08-14
Add `helm env` command to get details of the environment


mattfarina:
With the move to XDG paths for file locations and the removal (coming soon) of `helm init` there is no method to discover the directories helm is storing files in or other environment details. We should add a command to see these details.

waveywaves:
Please assign this to me @mattfarina I will work on this.

adamreese:
I found this plugin too https://github.com/adamreese/helm-env

akashshinde:
> I found this plugin too https://github.com/adamreese/helm-env

@adamreese Is this compatible with `helm-3` ? since we are using `XDG` paths ?

karuppiah7890:
Actually, currently it shows envs with "HELM" , "TILLER" word in it. Doesn't show XDG ones yet. But it shouldn't be a tough change to make. Just add an extra thing to the grep if you see the code

akashshinde:
@karuppiah7890 @adamreese But I would prefer having `helm env` as internal command over  installing any external plugins.

waveywaves:
@mattfarina can we close this as the PR is merged ? Or do we keep it open for tracking just in case.

PR #6244, opened 2019-08-17
(feature/helm-env) `helm env`

waveywaves:

Wrt https://github.com/helm/helm/issues/6226
Implementation for `helm env`

**TODO**
- [x] Formatting
- [ ] Unit Tests
- [ ] Documentation

cc @mattfarina

### DIFF
--- a/cmd/helm/env.go
+++ b/cmd/helm/env.go
@@ -19,11 +19,8 @@ package main
 import (
 	"fmt"
 	"io"
-	"os"
-	"strings"
 
-	"helm.sh/helm/pkg/helmpath"
-	"helm.sh/helm/pkg/helmpath/xdg"
+	"helm.sh/helm/pkg/cli"
 
 	"github.com/spf13/cobra"
 
@@ -38,6 +35,7 @@ Env prints out all the environment information in use by Helm.
 
 func newEnvCmd(out io.Writer) *cobra.Command {
 	o := &envOptions{}
+	o.settings = cli.New()
 
 	cmd := &cobra.Command{
 		Use:   "env",
@@ -53,48 +51,12 @@ func newEnvCmd(out io.Writer) *cobra.Command {
 }
 
 type envOptions struct {
+	settings *cli.EnvSettings
 }
 
 func (o *envOptions) run(out io.Writer) error {
-	o.setHelmPaths()
-	o.setXdgPaths()
-	for _, prefix := range []string{
-		"HELM_",
-		"XDG_",
-	} {
-		for _, e := range os.Environ() {
-			if strings.HasPrefix(e, prefix) {
-				fmt.Println(e)
-			}
-		}
+	for _, e := range o.settings.EnvironmentVariables {
+		fmt.Printf("%s=\"%s\" \n", e.Name, e.Value)
 	}
 	return nil
-}
-
-func (o *envOptions) setHelmPaths() {
-	for key, val := range map[string]string{
-		"HELM_HOME":                  helmpath.DataPath(),
-		"HELM_PATH_STARTER":          helmpath.DataPath("starters"),
-		"HELM_DEBUG":                 fmt.Sprint(settings.Debug),
-		"HELM_REGISTRY_CONFIG":       settings.RegistryConfig,
-		"HELM_PATH_REPOSITORY_FILE":  settings.RepositoryConfig,
-		"HELM_PATH_REPOSITORY_CACHE": settings.RepositoryCache,
-		"HELM_PLUGIN":                settings.PluginsDirectory,
-	} {
-		if eVal := os.Getenv(key); len(eVal) <= 0 {
-			os.Setenv(key, val)
-		}
-	}
-}
-
-func (o *envOptions) setXdgPaths() {
-	for key, val := range map[string]string{
-		xdg.CacheHomeEnvVar:  helmpath.CachePath(),
-		xdg.ConfigHomeEnvVar: helmpath.ConfigPath(),
-		xdg.DataHomeEnvVar:   helmpath.DataPath(),
-	} {
-		if eVal := os.Getenv(key); len(eVal) <= 0 {
-			os.Setenv(key, val)
-		}
-	}
 }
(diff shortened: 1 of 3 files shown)


waveywaves:
HELM Environment Variables

HELM_DEBUG               :

         XDG Environment Variables

XDG_DATA_DIRS            : /home/vbobade/.local/share/flatpak/exports/share/:/var/lib/flatpak/exports/share/:/usr/local/share/:/usr/share/                                                                        
XDG_RUNTIME_DIR          : /run/user/1000
XDG_SESSION_ID           : 2
XDG_SESSION_TYPE         : wayland
XDG_CURRENT_DESKTOP      : GNOME
XDG_SEAT                 : seat0
XDG_SESSION_DESKTOP      : gnome
XDG_VTNR                 : 2
XDG_MENU_PREFIX          : gnome-
Happy Helming!

akashshinde:
@waveywaves 
you might need to consider making it working for os x and windows as well. 
I did quick round of testing on mac machine(`os x Catalina 10.15 Beta (19A526h)`), following it the output.

> helm init
```
akash@Akashs-MacBook-Pro ~/go/src/helm.sh/helm/bin (dev-v3●)$ ./helm init                                                                     
Helm is now configured to use the following directories:
Cache: /Users/akash/Library/Caches/helm
Configuration: /Users/akash/Library/Preferences/helm
Data: /Users/akash/Library/helm
Happy Helming!

```

> helm env
```
akash@Akashs-MacBook-Pro ~/go/src/helm.sh/helm/bin (dev-v3●)$ ./helm env                                                                      
         HELM Environment Variables

         XDG Environment Variables
Happy Helming!
```

cmd/helm/env.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// run .
+func (o *envOptions) run(out io.Writer) error {
+
+	helmEnvs := make(map[string]string)
+	xdgEnvs := make(map[string]string)
+	o.maxEnvKeyLength = 0
+
+	for _, e := range os.Environ() {
```
mattfarina: Try to think of this like `go env`. It prints environment information including empty environment variables and defaults. It can't just look at the current environment variables. For example, in Go if I don't have a `GOPATH` set it defaults to `~/go` and that shows up if `go env` is run.

Can you please update it work in a similar manner.

cmd/helm/env.go
```diff
@@ -0,0 +1,97 @@
...
+	"regexp"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	"helm.sh/helm/cmd/helm/require"
+)
+
+var envHelp = `
+Env prints out all the environment variables in use by Helm.
```
mattfarina: Can you update this to be environment information instead of environment variables. Go provides a good example. See the help text for that for ideas.

mattfarina (changes requested):
Thanks for taking a shot at this. I have a little feedback.

The `helmpath` package will provide some of the defaults that should be displayed. They are different per OS. This should provide useful information for Mac and Windows as well as Linux.

cmd/helm/env.go
waveywaves: Thank you for checking this out. I will update the PR accordingly.

waveywaves:
@mattfarina @adamreese Does any of this need to be moved to the action package ?

mattfarina (changes requested):
Thanks for your continued work on this.

When I run this I see...

```
± ./bin/helm env
HELM_PLUGIN_DIR=
HELM_PATH_REPOSITORY_FILE=
HELM_HOME=/Users/mfarina/Library/helm
HELM_DEBUG=false
HELM_PLUGIN_NAME=
HELM_BIN=./bin/helm
HELM_PLUGIN=
HELM_PATH_REPOSITORY_CACHE=
HELM_PATH_STARTER=/Users/mfarina/Library/helm/starters
XDG_CACHE_HOME=/Users/mfarina/Library/Caches/helm
XDG_CONFIG_HOME=/Users/mfarina/Library/Preferences/helm
XDG_DATA_HOME=/Users/mfarina/Library/helm
GOPATH=/Users/mfarina/Code/go
GOBIN=~/go/bin
```

Some of these are empty and could use some good defaults. For the CLI environment setup see: https://github.com/helm/helm/blob/f76b5f21adb53a85de8925f4a9d4f9bd99f185b5/pkg/cli/environment.go#L54-L61

The following code is to setup an environment for a plugin to be run. I don't think we need that used here. Some of the information is relevant but should be part of the CLI environment. https://github.com/helm/helm/blob/f76b5f21adb53a85de8925f4a9d4f9bd99f185b5/pkg/plugin/plugin.go#L219-L235

waveywaves:
@mattfarina I have removed the function which creates the environment for helm plugins and have added the funcitonality in a newer function. Do check and let me know.

cmd/helm/env.go
```diff
@@ -0,0 +1,100 @@
...
+		"HELM_HOME":                  helmpath.DataPath(),
+		"HELM_PATH_STARTER":          helmpath.DataPath("starters"),
+		"HELM_DEBUG":                 fmt.Sprint(settings.Debug),
+		"HELM_REGISTRY_CONFIG":       settings.RegistryConfig,
+		"HELM_PATH_REPOSITORY_FILE":  settings.RepositoryConfig,
+		"HELM_PATH_REPOSITORY_CACHE": settings.RepositoryCache,
+		"HELM_PLUGIN":                settings.PluginsDirectory,
+	} {
+		if eVal := os.Getenv(key); len(eVal) <= 0 {
+			os.Setenv(key, val)
```
mattfarina: As I looked at this and looked at how Go implemented it I realized we can do this without setting any environment variables. Can we build an internal model inside Helm of the environment and then display that?

You can see the Go code at https://github.com/golang/go/blob/master/src/cmd/go/internal/envcmd/env.go

mattfarina (changes requested):
Thanks for continuing to work on this and being patient with the feedback.

cmd/helm/env.go
waveywaves: Thank you for reviewing this.
I misread your comment on `understanding what go has done` in terms of understanding the output and not the implementation detalis. I will implement this and update it ASAP.

waveywaves: I have moved the environment variables under EnvSettings where they are queried from.

mattfarina (approved):
Going into the helm v3 branch this will need a second maintainer to approve.

thomastaylor312 (approved):
This LGTM! Thanks for all of the work!

Issue #6232, opened 2019-08-16
feat(cmd/helm): Lazy create config and cache directories


adamreese:
Running the command `helm init` in v3 is only creating the directory structures that helm expects.  This can be done lazily when running commands like `helm repo add`.

PR #6255, opened 2019-08-19
(fix/issue-6237) Add check to see if connected to the kubernetes cluster

### LINKED ISSUE #6237: Helm 3: helm list shows no error when disconnected from internet
No internet:
```
$ bin/helm list --debug
NAME	NAMESPACE	REVISION	UPDATED	STATUS	CHART
$
```

Internet:
```
$ bin/helm list --debug
NAME     	NAMESPACE	REVISION	UPDATED                             	STATUS  	CHART
myrelease	default  	1       	2019-08-15 10:25:39.538926 -0400 EDT	deployed	nginx-4.3.3
$
```

waveywaves:

Fixes #6237 

**If applicable**:

cc @mattfarina @thomastaylor312 @adamreese

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -63,6 +63,16 @@ func New(getter genericclioptions.RESTClientGetter) *Client {
 
 var nopLogger = func(_ string, _ ...interface{}) {}
 
+// Test connectivity to the Client
+func (c *Client) IsReachable() error {
+	client, _ := c.Factory.KubernetesClientSet()
+	_, err := client.ServerVersion()
+	if err != nil {
+		return errors.New("Kubernetes cluster unreachable")
+	}
+	return nil
+}
+
 // Create creates Kubernetes resources specified in the resource list.
 func (c *Client) Create(resources ResourceList) (*Result, error) {
 	c.Log("creating %d resource(s)", len(resources))
--- a/pkg/action/get.go
+++ b/pkg/action/get.go
@@ -38,5 +38,9 @@ func NewGet(cfg *Configuration) *Get {
 
 // Run executes 'helm get' against the given release.
 func (g *Get) Run(name string) (*release.Release, error) {
+	if err := g.cfg.KubeClient.IsReachable(); err != nil {
+		return nil, err
+	}
+
 	return g.cfg.releaseContent(name, g.Version)
 }
--- a/pkg/action/history.go
+++ b/pkg/action/history.go
@@ -41,6 +41,10 @@ func NewHistory(cfg *Configuration) *History {
 
 // Run executes 'helm history' against the given release.
 func (h *History) Run(name string) ([]*release.Release, error) {
+	if err := h.cfg.KubeClient.IsReachable(); err != nil {
+		return nil, err
+	}
+
 	if err := validateReleaseName(name); err != nil {
 		return nil, errors.Errorf("release name is invalid: %s", name)
 	}
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -107,6 +107,10 @@ func NewInstall(cfg *Configuration) *Install {
 //
 // If DryRun is set to true, this will prepare the release, but not install it
 func (i *Install) Run(chrt *chart.Chart, vals map[string]interface{}) (*release.Release, error) {
+	if err := i.cfg.KubeClient.IsReachable(); err != nil {
+		return nil, err
+	}
+
 	if err := i.availableName(); err != nil {
 		return nil, err
 	}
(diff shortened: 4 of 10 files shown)


waveywaves:
Let me know if I need to make any changes.
Output of the command looks like the following when not connected to the internet.
```
➜  helm git:(fix/6237) ✗ helm list --debug
helm.go:132: [debug] Please connect to the Internet.
NAME    NAMESPACE       REVISION   
```

waveywaves:
cc @jdolitsky

jdolitsky:
Hi thanks for looking into this. I think that there is maybe a more elegant
way to check connection via the kube client. I think connected to the
“internet” is misstated - should be connected to “cluster” (think minikube
for example, no internet connection required)

waveywaves:
@jdolitsky That makes more sense. I will update the PR accordingly.

jdolitsky:
Cool. My guess is there is an ignored error somewhere? (Sry away from
computer)

waveywaves:
@jdolitsky I have updated the PR and this time I am checking if the Kubernetes client is reachable. DO let me know if I need ot make any changes.

cmd/helm/helm.go
```diff
@@ -92,6 +92,11 @@ func initActionConfig(actionConfig *action.Configuration, allNamespaces bool) {
 		// TODO return error
 		log.Fatal(err)
 	}
+	err = kc.IsReachable()
```
mattfarina: Two things come to mind here...

1. #6232 is in flight to remove the `init` command. Right now it only setups up missing files locally as tiller is gone. The idea is to lazy create the directories if not present.
2. The init action no longer requires communicating with the server. In this function the goal is to get the client set which is locally available and the namespace which is also locally available. If I'm on a plane working on something and need to init helm why shouldn't it work? This is a real work situation for me a couple times a year. Maybe others.

Does it matter if the Kubernetes cluster is available at this step?

marckhouzam: I second point #2:
- there are times where I don't have a network and still use some parts of helm for development.
- there are times when I have internet but can't reach my cluster (VPN not activated).  

I guess verifying the cluster connection should be done for commands that actually need the cluster.

For example, `helm --help` should always work (it does not with this patch).
`helm repo add ...`, `helm show ...` are other examples where the cluster is not needed.

waveywaves: It only makes sense to have the check run for certain commands which would use kubernetes. Thank you for helping me out. I have updated the PR accordingly.

akashshinde: @mattfarina 
we should have way to check cluster connection just to make sure helm commands going to work. How about having explicit helm command to check such as `helm cluster status` ? 
cc: @marckhouzam

waveywaves: @mattfarina @jdolitsky I have made further changes to the PR. Have added the check only in places which seem necessary in terms of connecting to the Kubernetes cluster.

waveywaves:
Test coverage is failing here, on my local system it works. Any idea why this might be happening @jdolitsky

cmd/helm/get.go
```diff
@@ -47,6 +48,9 @@ func newGetCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 		Long:  getHelp,
 		Args:  require.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
+			kc := kube.New(kubeConfig())
+			kc.IsReachable()
```
mattfarina: Can you please move this to the [`action` package](https://github.com/helm/helm/tree/dev-v3/pkg/action)? In this code flow you'll find it inside `client.Run()`. That way if someone imports the action package to leverage this functionality they will have checks for for connectivity.

pkg/kube/client.go
```diff
@@ -63,6 +64,16 @@ func New(getter genericclioptions.RESTClientGetter) *Client {
 
 var nopLogger = func(_ string, _ ...interface{}) {}
 
+// Test connectivity to the Client
+func (c *Client) IsReachable() {
+	client, _ := c.Factory.KubernetesClientSet()
+	_, err := client.ServerVersion()
+	if err != nil {
+		c.Log("--- Kubernetes Cluster Unreachable ---")
+		os.Exit(1)
```
mattfarina: Instead of performing an `os.Exit` can you return an error. Then where this code is used can you check for an error and return it up the stack.

waveywaves:
@jdolitsky  @mattfarina I have made the necessary changes, do check it out

jdolitsky (approved):
This looks good

mattfarina (approved):

marckhouzam:
@waveywaves This is really nice!  I think `helm test run` would also benefit from this check:

```
> helm test run hello
Error: query: failed to query with labels: 
Get https://mycluster:6443/api/v1/namespaces/default/secrets?labelSelector=name%3Dhello%2Cowner%3Dhelm: 
Get https://myopenid-configuration: dial tcp: lookup mykubehost: no such host
```

PR #6256, opened 2019-08-20
Poposal: Hook to run acceptance tests

marckhouzam:

To run acceptance tests during development, it would be easier to run
```
make test-acceptance
```
directly from the Helm repo, instead of having to do:
```
make build
make build-cross
cd ../acceptance-testing
export ROBOT_HELM_PATH=../helm/bin
make acceptance
cd ../helm
```

Granted that running all the acceptance tests during development may not be a very common case, but it opens the door to easily running a subset of these tests such as the tests for completion, as also included in this PR:
```
make test-completion
```
instead of:
```
make build
make build-cross
cd ../acceptance-testing
export ROBOT_HELM_PATH=../helm/bin
export ROBOT_RUN_TESTS=shells.robot
make acceptance
cd ../helm
```

Beyond convenience, this would also help guide people to finding and using the `acceptance-testing` repo without having to learn much about it: it should just work.

I've used the make targets `test-acceptance` and `test-completion` to keep in line with the other make targets for tests, which are all prefixed with `test-`.


marckhouzam:
Heads up to @jdolitsky who maintains the `acceptance-testing` repo, in case he disagrees.

jdolitsky (changes requested):
This lgtm! 

Do you think we can rename `test-completion` to `test-acceptance-completion`? We can then follow this pattern and add things like `test-acceptance-kubernetes-versions`

marckhouzam:
> Do you think we can rename `test-completion` to `test-acceptance-completion`?
> We can then follow this pattern and add things like `test-acceptance-kubernetes-versions`

Done.

While we're talking about naming, technically, this target should be `test-acceptance-shells`, but that is not what I'm aiming for as what I really want is to be able to run the completion tests.  So I preferred calling it `test-acceptance-completion`.  This will need revisiting the day there are tests added to `shells.robot` that are not related to completion.

jdolitsky (approved):
i think this works for now

marckhouzam:
Is this small improvement to the Makefile ok to merge? It's been approved already.
```
make test-acceptance
```
It would help the verification of PRs such as #6327 and #6340

hickeyma (approved):
The targets look good and I can run them. LGTM, thanks @marckhouzam.

FYI @marckhouzam I am getting some fails when running the tests. Might be my env. Here is the output: https://gist.github.com/hickeyma/8beaa068402c474f176e41087d6c2b24

marckhouzam:
Thanks @hickeyma!
The failure you reported was a problem with the acceptance-testing repo and has just been fixed in https://github.com/helm/acceptance-testing/pull/29.

PR #6258, opened 2019-08-21
fix issue when dependency  not in cache

### LINKED ISSUE #6005: proposal: `helm dependency update` might not require `helm repo add`
I understand the reason a `requirements.yaml` uses the repo URL as the canonical location identifier. That seems like a good idea. I was surprised however when packaging a chart on a CI server that `helm repo add <name> <url>` must also be done before a subchart on a public repo can be used.

Steps to reproduce:

### requirements.yaml
```yaml
dependencies:
  - name: cert-manager
    version: v0.9.0-alpha.0
    repository: https://charts.jetstack.io
```
### actual
```
$ helm package charts/my-chart --dependency-update
Error: no repository definition for https://charts.jetstack.io. Please add them via 'helm repo add'
```
### expected
behavior of `--dependency-update` and `helm dependency update` to be more like `helm fetch`, which does not require `helm repo add` or `helm repo update` to be done.
```
$ helm fetch cert-manager --repo https://charts.jetstack.io --version v0.9.0-alpha.0 --destination charts/my-chart/charts
$ ls charts/my-chart/charts
cert-manager-v0.9.0-alpha.0.tgz
```

daixiang0:
Signed-off-by: Xiang Dai <764524258@qq.com>

Fix #6005

### DIFF
--- a/pkg/resolver/resolver.go
+++ b/pkg/resolver/resolver.go
@@ -71,7 +71,18 @@ func (r *Resolver) Resolve(reqs *chartutil.Requirements, repoNames map[string]st
 			return nil, fmt.Errorf("dependency %q has an invalid version/constraint format: %s", d.Name, err)
 		}
 
-		repoIndex, err := repo.LoadIndexFile(r.helmhome.CacheIndex(repoNames[d.Name]))
+		// repo does not exist in cache but has url info
+		cacheRepoName := repoNames[d.Name]
+		if cacheRepoName == "" && d.Repository != "" {
+			locked[i] = &chartutil.Dependency{
+				Name:       d.Name,
+				Repository: d.Repository,
+				Version:    d.Version,
+			}
+			continue
+		}
+
+		repoIndex, err := repo.LoadIndexFile(r.helmhome.CacheIndex(cacheRepoName))
 		if err != nil {
 			return nil, fmt.Errorf("no cached repo found. (try 'helm repo update'). %s", err)
 		}
--- a/pkg/downloader/chart_downloader.go
+++ b/pkg/downloader/chart_downloader.go
@@ -205,6 +205,11 @@ func (c *ChartDownloader) ResolveChartVersion(ref, version string) (*url.URL, ge
 	}
 	c.setCredentials(r)
 
+	// Skip if dependency not contain name
+	if len(r.Config.Name) == 0 {
+		return u, r.Client, nil
+	}
+
 	// Next, we need to load the index, and actually look up the chart.
 	i, err := repo.LoadIndexFile(c.HelmHome.CacheIndex(r.Config.Name))
 	if err != nil {
--- a/pkg/downloader/manager.go
+++ b/pkg/downloader/manager.go
@@ -233,7 +233,7 @@ func (m *Manager) downloadAll(deps []*chartutil.Dependency) error {
 
 		// Any failure to resolve/download a chart should fail:
 		// https://github.com/kubernetes/helm/issues/1439
-		churl, username, password, err := findChartURL(dep.Name, dep.Version, dep.Repository, repos)
+		churl, username, password, err := m.findChartURL(dep.Name, dep.Version, dep.Repository, repos)
 		if err != nil {
 			saveError = fmt.Errorf("could not find %s: %s", churl, err)
 			break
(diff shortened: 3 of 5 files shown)


thomastaylor312 (changes requested):
I don't think this solves the problem as described in #6005. With this change if I have a dependency on a repository called `foo` and `foo` doesn't exist in my repos, this will not error out like it should. I would suggest adding a section that checks if it is a URL and then add that to the dependency maps

daixiang0:
@thomastaylor312 please review again.

daixiang0:
@thomastaylor312 any update?

daixiang0:
@thomastaylor312 never mind, happy to wait for review:)

thomastaylor312 (changes requested):
When I was testing this, I got the following error when trying to update:
```
Error: no cached repo found. (try 'helm repo update'). open /Users/taylor/.helm/repository/cache/-index.yaml: no such file or directory
```

I believe this is due to the fact that there is no repo name when there is a URL, and so it is pointing to a non-existent cache file. It will need to bypass this when there is no name and fetch the chart directly

daixiang0:
@thomastaylor312 hi, i tried to bypass this, but i am confused that when i run `make build` then check version, its version is ok, but the output still old not my code, is there anything i miss?

daixiang0:
@thomastaylor312 hi, mind have a look?

thomastaylor312:
@daixiang0 Sorry! I must have misread your comment. Are you ready for me to take a look at this or are you having problems getting it to build?

daixiang0:
@thomastaylor312 yep, i met problems when build since what i build do not expected what i code. Do not find detail docs about it, please help me.

thomastaylor312:
I pulled down your branch and was able to run `make bootstrap build` just fine. If things don't seem to be building, you can always do `rm -rf bin/` and then re-run a `make build` to make sure you don't have anything stale around

daixiang0:
@thomastaylor312 i am really confused, after build, check version that client commit is correct, but what i code do not work, and try to add something in debug while not work too.
```
git checkout fix-stable
rm -rf bin
make build
cd /path/to/test
/path/to/bin/helm version <== git commit is correct
/path/to/bin/helm dependency update --debug <=== no custom debug info
```

thomastaylor312:
Assuming you are trying to print some debug logic that you added, it is most likely because you aren't hitting that branch of logic because of some edge case you are running into. I'd need to see exactly what you had added to the code for debugging to be sure though

daixiang0:
@thomastaylor312 i get the key: i should clone to  `k8s.io/helm` not `github.com/helm`.
I want to add docs about build stuff, which place do you think is good for PR?

karuppiah7890:
@daixoang0 adding to this, you could also add docs for helm v3 (dev-v3) branch, for which code has to be present at $GOPATH/src/helm.sh/helm

daixiang0:
@thomastaylor312 please review

daixiang0:
@thomastaylor312 mind have a look?

woodcockjosh:
@thomastaylor312 I have tested @daixiang0 branch. I run `helm dep up` I still get this message `Error: no repository definition for https://helm.elastic.co, https://helm.elastic.co. Please add them via 'helm repo add'` In its current state I don't see how this fixes #6005 but am very interested in helping if any help is needed.

woodcockjosh:
@thomastaylor312 I just checked @daixiang0 branch. This appears to be working.

woodcockjosh:
relates to #6454

pkg/resolver/resolver.go
```diff
@@ -71,7 +71,18 @@ func (r *Resolver) Resolve(reqs *chartutil.Requirements, repoNames map[string]st
 			return nil, fmt.Errorf("dependency %q has an invalid version/constraint format: %s", d.Name, err)
 		}
 
-		repoIndex, err := repo.LoadIndexFile(r.helmhome.CacheIndex(repoNames[d.Name]))
+		// repo does not exist in cache but has url info
+		cacheRepoName := repoNames[d.Name]
```
thomastaylor312: Nit: The if statement could be simplified to
```suggestion
		cacheRepoName, ok := repoNames[d.Name]
		if !ok && d.Repository != "" {
```

thomastaylor312 (approved):
This is looking good! Thanks for all the hard work!

thomastaylor312:
@woodcockjosh Thanks for taking a look at this and validating

pkg/resolver/resolver.go
thomastaylor312: Since this is just a minor, non-blocking nit. I am going to merge

PR #6263, opened 2019-08-22
helm-v3: Dynamic completion for "helm repo" and "helm plugin"

marckhouzam:
**What this PR does / why we need it**:

This PR adds dynamic shell completion for the commands:
- helm repo remove \<TAB\> (which will select from the names of configured repos)
- helm plugin remove \<TAB\> (which will select from the names of installed plugins)
- helm plugin update \<TAB\> (which will select from the names of installed plugins)

This PR has been tested for backwards compatibility through the acceptance-testing repo.
Tests have been prepared and submitted in PR https://github.com/helm/acceptance-testing/pull/17

PR #6265, opened 2019-08-22
v2: Dynamic completion for "helm repo" and "helm plugin"

marckhouzam:

This PR adds dynamic shell completion for the commands:
- helm repo remove \<TAB\> (which will select from the names of configured repos)
- helm repo update \<TAB\> (which will select from the names of configured repos)
- helm plugin remove \<TAB\> (which will select from the names of installed plugins)
- helm plugin update \<TAB\> (which will select from the names of installed plugins)

### DIFF
--- a/cmd/helm/helm.go
+++ b/cmd/helm/helm.go
@@ -111,7 +111,7 @@ __helm_custom_func()
             __helm_list_releases
             return
             ;;
-        helm_repo_remove)
+        helm_repo_remove | helm_repo_update)
             __helm_list_repos
             return
             ;;
--- a/cmd/helm/helm.go
+++ b/cmd/helm/helm.go
@@ -80,7 +80,6 @@ __helm_list_releases()
     fi
 }
 
-
 __helm_list_repos()
 {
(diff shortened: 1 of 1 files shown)


marckhouzam:
@hickeyma @jdolitsky This PR for helm v2 is very similar to #6263 which was just merged.  The difference is that for v2:

1. The `--home` flag can be used, which affects the completion
1. The command `helm repo update` takes a repo name as a param, so also requires completion

marckhouzam:
@hickeyma To facilitate approval of improvements such as this one, we can now run the acceptance **completion tests** with Helm v2 (https://github.com/helm/acceptance-testing/pull/30).

Therefore, if you run the completion tests with the current `master` you will see failures; then if you run them with this PR applied, they should all pass. To run them from `master`:
```
cd $GOPATH/src/helm.sh/acceptance-testing
ROBOT_HELM_PATH=$GOPATH/src/k8s.io/helm/bin ROBOT_RUN_TESTS=shells.robot make
```

I hope this can provide confidence that the proposed completion changes are safe.

hickeyma (approved):
LGTM, thanks @marckhouzam

PR #6266, opened 2019-08-22
Networkpolicy kind sort

### LINKED ISSUE #4199: Install NetworkPolicies after Namespaces
Helm should install NetworkPolicy resources after Namespace resources.

Currently the InstallOrder at https://github.com/kubernetes/helm/blob/master/pkg/tiller/kind_sorter.go#L29 does not list the NetworkPolicy resource. If a chart creates a Namespace resource and a NetworkPolicy resource in this Namespace, the deployment of the chart might fail because helm tries to create the NetworkPolicy before the Namespace has been created.

rekcah78:
closes helm#4199

This PR add NetworkPolicy in tiller kind sorter, without that, NetworkPolicy is created before Namespace and the helm install fail.

**If applicable**:

### DIFF
--- a/pkg/tiller/kind_sorter.go
+++ b/pkg/tiller/kind_sorter.go
@@ -28,6 +28,7 @@ type SortOrder []string
 // Those occurring earlier in the list get installed before those occurring later in the list.
 var InstallOrder SortOrder = []string{
 	"Namespace",
+	"NetworkPolicy",
 	"ResourceQuota",
 	"LimitRange",
 	"PodSecurityPolicy",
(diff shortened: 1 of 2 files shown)


mattfarina:
@rekcah78 When you see an error like:

```
/usr/local/go/pkg/tool/linux_amd64/link: signal: killed
```

it is due to our tests running out of memory. This is a problem we solved in Helm v3 but still have sometimes for Helm v2. We can try to re-run the test until that error goes away.

rekcah78:
@mattfarina thank, Its seem all good now.

hickeyma (approved):
LGTM, thanks @rekcah78

PR #6285, opened 2019-08-26
chore(registry): upgrade to oras v0.7.0

jdolitsky:
Also, adding --insecure flag to "registry login" now exposed by oras

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -44,11 +44,15 @@
 
 [[constraint]]
   name = "github.com/deislabs/oras"
-  version = "0.6.0"
+  version = "0.7.0"
+
+[[constraint]]
+  name = "github.com/containerd/containerd"
+  version = "1.3.0-beta.2"
 
 [[constraint]]
   name = "github.com/sirupsen/logrus"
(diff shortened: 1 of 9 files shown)


Gopkg.toml
```diff
@@ -44,11 +44,15 @@
 
 [[constraint]]
   name = "github.com/deislabs/oras"
-  version = "0.6.0"
+  version = "0.7.0"
+
+[[constraint]]
+  name = "github.com/containerd/containerd"
+  version = "1.3.0-beta.2"
```
hickeyma: Do you think containerd 1.3.0 will be released before Helm 3?

hickeyma (commented):
Looks good. Question about containerd version. Also, can we test for `insecure` mode?

thomastaylor312 (approved):
Minor non-blocking nit. Otherwise good to go

hickeyma (approved):
LGTM, thanks

Gopkg.toml
hickeyma: Sure, being experimental it should be ok.

PR #6298, opened 2019-08-27
fix(cmd/helm): Remove mention of init from help

marckhouzam:

Remove the mention of the `init` command from the help text for v3.

Maybe another introductory sentence should be put in its place instead of simply removing it like I did?


hickeyma (approved):
LGTM, thanks @marckhouzam

PR #6308, opened 2019-08-28
fix(cmd/helm): user friendly error message when repos are not configured

adamreese:


hickeyma (changes requested):
Looking good @adamreese, thanks. Could you add similar error handling in `cmd`, for  where `LoadFile` is called in `pkg/downloader/manager.go` and `./pkg/downloader/chart_downloader.go` also?

hickeyma:
@adamreese When installing directly from repo (for example, `helm install nginx nginx-ingress --repo https://kubernetes-charts.storage.googleapis.com`). More details in https://github.com/helm/helm/pull/6293#discussion_r318443444.

hickeyma (approved):
LGTM, thanks @adamreese

PR #6327, opened 2019-08-30
fix BusyBox sed

tipok:
BusyBox sed works the same way as GNU sed

Code completion for zsh under alpine is not working currently because BusyBox sed is not identified as a GNU sed.
This leads to
LWORD='[[:<:]]'
RWORD='[[:>:]]'
which is not working in BusyBox sed

**If applicable**:

### DIFF
--- a/cmd/helm/completion.go
+++ b/cmd/helm/completion.go
@@ -194,7 +194,7 @@ autoload -U +X bashcompinit && bashcompinit
 # use word boundary patterns for BSD or GNU sed
 LWORD='[[:<:]]'
 RWORD='[[:>:]]'
-if sed --help 2>&1 | grep -q GNU; then
+if sed --help 2>&1 | grep -q 'GNU\|BusyBox'; then
 	LWORD='\<'
 	RWORD='\>'
 fi


bacongobbler:
cc @marckhouzam

marckhouzam:
@tipok Nice find!  The problem also applies to helm v3.  Would you like to make a PR for the dev-v3 branch also?

I've just added some tests for zsh completion on alpine in the [acceptance-testing](https://github.com/helm/acceptance-testing) repo (which currently only works with Helm v3). They fail currently, and start passing when I apply your fix.

I was about to mention that the problem also applies to kubectl, but I see you have already submitted a fix that has been approved (https://github.com/kubernetes/kubernetes/pull/82181) 👍 

@bacongobbler LGTM

tipok:
@marckhouzam sounds good. I'll create an additional PR for the dev-v3 branch

hickeyma (approved):
Approved as per @marckhouzam comment https://github.com/helm/helm/pull/6327#issuecomment-527179787. Thanks @tipok for the fix.

PR #6332, opened 2019-08-30
fix: clear the discovery cache after CRDs are installed

### LINKED ISSUE #6316: Helm v3 beta -- Wait for CRDs to be created
When deploying CRDs with Helm v3, the rest of the chart deployment occurs _immediately_ after a CRD is installed.  This can cause the chart deployment to fail if the CRDs are not yet available.  Helm should wait until the CRD is deployed before continuing with the chart.

technosophos:
This fixes a bug in which CRD instances could not be created in the same chart as the CRDs themselves. This invalidates the local discovery cache and forces a round-trip discovery operation with the Kubernetes API server.

Closes #6316 

**If applicable**:

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -169,6 +169,15 @@ func (i *Install) Run(chrt *chart.Chart, vals map[string]interface{}) (*release.
 				}
 				return i.failRelease(rel, err)
 			}
+
+			// Invalidate the local cache.
+			if discoveryClient, err := i.cfg.RESTClientGetter.ToDiscoveryClient(); err != nil {
+				// On error, we don't want to bail out, since this is unlikely
+				// to impact the majority of charts.
+				i.cfg.Log("Could not clear the discovery cache: %s", err)
+			} else {
+				discoveryClient.Invalidate()
+			}
 		}
 	}
 
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -111,6 +111,49 @@ func (i *Install) Run(chrt *chart.Chart, vals map[string]interface{}) (*release.
 		return nil, err
 	}
 
+	// Pre-install anything in the crd/ directory. We do this before Helm
+	// contacts the upstream server and builds the capabilities object.
+	if crds := chrt.CRDs(); !i.ClientOnly && !i.SkipCRDs && len(crds) > 0 {
+		// We do these one at a time in the order they were read.
+		for _, obj := range crds {
+			// Read in the resources
+			res, err := i.cfg.KubeClient.Build(bytes.NewBuffer(obj.Data))
+			if err != nil {
+				// We bail out immediately
+				return nil, errors.Wrapf(err, "failed to install CRD %s", obj.Name)
+			}
+			// On dry run, bail here
+			if i.DryRun {
+				i.cfg.Log("WARNING: This chart or one of its subcharts contains CRDs. Rendering may fail or contain inaccuracies.")
+				continue
+			}
+			// Send them to Kube
+			if _, err := i.cfg.KubeClient.Create(res); err != nil {
+				// If the error is CRD already exists, continue.
+				if apierrors.IsAlreadyExists(err) {
+					crdName := res[0].Name
+					i.cfg.Log("CRD %s is already present. Skipping.", crdName)
+					continue
+				}
+				return nil, err
+			}
+
+			// Invalidate the local cache.
+			if discoveryClient, err := i.cfg.RESTClientGetter.ToDiscoveryClient(); err != nil {
+				// On error, we don't want to bail out, since this is unlikely
+				// to impact the majority of charts.
+				i.cfg.Log("Could not clear the discovery cache: %s", err)
+			} else {
+				i.cfg.Log("Clearing discovery cache")
+				discoveryClient.Invalidate()
+				// Give time for the CRD to be recognized.
+				time.Sleep(5 * time.Second)
+				// Make sure to force a rebuild of the cache.
+				discoveryClient.ServerGroups()
+			}
+		}
+	}
+
 	if i.ClientOnly {
 		// Add mock objects in here so it doesn't use Kube API server
 		// NOTE(bacongobbler): used for `helm template`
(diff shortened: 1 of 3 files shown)


hickeyma:
@thequailman can you test if this fixes the issue?

karuppiah7890:
I'm wondering if it's possible to write a failing test for the issue and then pass it with this PR

thomastaylor312 (changes requested):
I can't get this to work for me and I think I know why, but am not sure how to fix it. We should pair and see what we can come up with

technosophos:
Looks like were two issues with the earlier draft of this patch:

- The capabilities cache gets generated before the CRDs are installed. I reversed the order of this.
- Some clusters take longer to register CRDs. I was testing locally with Kind, which is fast. But some take a second or two.

The following script is what I have been using for testing:

```bash
BASPATH=/Users/technosophos/Code/Go/src/helm.sh/helm
HELM=$BASPATH/bin/helm
CHART=$BASPATH/_scratch/test-0.1.0.tgz

# Build a fresh Helm
echo "Building Helm"
cd $BASPATH && make build

echo "Cleaning environment"
# Clear all CRDs
kubectl delete crds --all
# Clear cache
rm -r /Users/technosophos/.kube/cache/discovery/127.0.0.1_*

echo "Running test"
# Run a command that forces discovery client to regen
# Install Helm chart
$HELM install --debug test1 $CHART

echo "Deleting helm install"
$HELM delete test1
```

pkg/action/install.go
```diff
@@ -111,6 +111,49 @@ func (i *Install) Run(chrt *chart.Chart, vals map[string]interface{}) (*release.
...
+			// Invalidate the local cache.
+			if discoveryClient, err := i.cfg.RESTClientGetter.ToDiscoveryClient(); err != nil {
+				// On error, we don't want to bail out, since this is unlikely
+				// to impact the majority of charts.
+				i.cfg.Log("Could not clear the discovery cache: %s", err)
+			} else {
+				i.cfg.Log("Clearing discovery cache")
+				discoveryClient.Invalidate()
+				// Give time for the CRD to be recognized.
+				time.Sleep(5 * time.Second)
```
thomastaylor312: We should probably clean up and reuse the code found here in Helm 2 so we aren't just sleeping and hoping it works: https://github.com/helm/helm/blob/3ed073b2e0c37dd5c994562d6fe8e04b5f3609eb/pkg/kube/client.go#L584-L612

thomastaylor312 (commented):
I think we already have some code we can copy paste instead of sleeping to wait for CRDs to reach an established state

pkg/action/install.go
technosophos: I am adding support to Wait() for CRDs, and then just using that. It has some hidden advantages for if people try to sneak non-CRD things into the CRDs() directory. But _sssssh... don't tell them that they can do this_.

thomastaylor312: tenor-138232708

technosophos: @thomastaylor312 Okay, this is now done. We're using an updated version of the Wait logic to watch for the CRD to become available.

pkg/kube/wait.go
```diff
@@ -46,6 +48,9 @@ type waiter struct {
 // until all are ready or a timeout is reached
 func (w *waiter) waitForResources(created ResourceList) error {
 	w.log("beginning wait for %d resources with timeout of %v", len(created), w.timeout)
+	if err := apiextv1beta1.AddToScheme(scheme.Scheme); err != nil {
```
technosophos: Adding CRDs to the globally shared Scheme (Yay, Singletons!) is probably best done elsewhere. Any idea where?

pkg/action/install.go
```diff
@@ -103,6 +103,50 @@ func NewInstall(cfg *Configuration) *Install {
...
+				continue
+			}
+			return errors.Wrapf(err, "failed to instal CRD %s", obj.Name)
+		}
+
+		// Invalidate the local cache.
+		if discoveryClient, err := i.cfg.RESTClientGetter.ToDiscoveryClient(); err != nil {
+			// On error, we don't want to bail out, since this is unlikely
+			// to impact the majority of charts.
+			i.cfg.Log("Could not clear the discovery cache: %s", err)
```
thomastaylor312: So I was thinking about this some more. I think we actually _do_ want to bail out here. If we got into this block, the user was meaning to install CRDs and we don't want their release to just fail on them. What do you think?

pkg/kube/client.go
```diff
@@ -55,6 +58,10 @@ func New(getter genericclioptions.RESTClientGetter) *Client {
 	if getter == nil {
 		getter = genericclioptions.NewConfigFlags(true)
 	}
+	if err := apiextv1beta1.AddToScheme(scheme.Scheme); err != nil {
+		// This should never happen.
```
thomastaylor312: Famous last words... 😉

pkg/kube/client.go
```diff
@@ -55,6 +58,10 @@ func New(getter genericclioptions.RESTClientGetter) *Client {
 	if getter == nil {
 		getter = genericclioptions.NewConfigFlags(true)
 	}
+	if err := apiextv1beta1.AddToScheme(scheme.Scheme); err != nil {
```
thomastaylor312: We should add a comment here so people don't have to do the same research you did

pkg/kube/wait.go
```diff
@@ -257,6 +270,26 @@ func (w *waiter) daemonSetReady(ds *appsv1.DaemonSet) bool {
...
+	for _, cond := range crd.Status.Conditions {
+		switch cond.Type {
+		case apiextv1beta1.Established:
+			if cond.Status == apiextv1beta1.ConditionTrue {
+				return true
+			}
+		case apiextv1beta1.NamesAccepted:
+			if cond.Status == apiextv1beta1.ConditionFalse {
+				// This indicates a naming conflict, but it's probably not the
+				// job of this function to faile because of that. Instead,
```
thomastaylor312: Spelling nit: s/faile/fail

thomastaylor312 (changes requested):
Just a few small changes that I found on this second pass

pkg/action/install.go
thomastaylor312: Oh and also bail out on a failed wait

pkg/action/install.go
```diff
@@ -103,6 +103,50 @@ func NewInstall(cfg *Configuration) *Install {
...
+		} else {
+			i.cfg.Log("Clearing discovery cache")
+			discoveryClient.Invalidate()
```
thomastaylor312: I was looking through this one more time and I think we should do the cache invalidation at the very end, outside of the for loop. When I looked into the code, I think the discovery client calls the api multiple times for each api group, meaning that if someone installs 5-10 CRDs, they will be calling their API hundreds of times

thomastaylor312: Source: https://github.com/kubernetes/cli-runtime/blob/9d670f2021f41f32ed762a6ddbc1c3ac7c441126/pkg/genericclioptions/config_flags.go#L222-L225

technosophos: So originally I was doing this so that the crds folder could have things that depended on each other. And this operates _per file_, not per CRD. But I'll switch it.

pkg/action/install.go
technosophos: I don't like the idea of failing on transient errors that are unlikely to impact the user. It seems wrong to me to take something that is not fatal (a cleared cache, a wait that is likely not going to take long anyway) and then require the user to take action.

However, I don't care enough to drag on this PR longer, so I will change it.

technosophos:
Please re-review.

thomastaylor312 (approved):
LGTM

PR #6335, opened 2019-09-01
doc(cli): restore help text for helm configuration

marckhouzam:

With the removal of the `init` command, the nice help message from @bacongobbler about XDG configuration disappeared.  This commit adds it back in the main help message.


thomastaylor312 (approved):
Thanks for adding this back in!

Issue #6336, opened 2019-09-01
helm3 should show timestamps in debug output


freeseacher:
Deploying helm release to low resources machine i found that option `--debug` is much more useful then in helm2 to get info where deploy is. 
but my deploy lasts about 14 minutes and it will be much more useful if i can see in debug info also timestamp. 
for now my deploy log looks like this
```
$ ${DR} helm3 upgrade --wait --timeout 1200s --install --debug --atomic noc noc --set image.tag=master-dev --set image.pullPolicy=Always
history.go:48: [debug] getting history for release noc
Release "noc" does not exist. Installing it now.
install.go:158: [debug] Original chart version: ""
install.go:175: [debug] CHART PATH: /apps/noc

client.go:68: [debug] creating 100 resource(s)
wait.go:48: [debug] beginning wait for 100 resources with timeout of 20m0s
wait.go:218: [debug] PersistentVolumeClaim is not bound: default/noc-clickhouse
wait.go:218: [debug] PersistentVolumeClaim is not bound: default/noc-clickhouse
wait.go:218: [debug] PersistentVolumeClaim is not bound: default/noc-clickhouse
wait.go:218: [debug] PersistentVolumeClaim is not bound: default/noc-clickhouse
wait.go:218: [debug] PersistentVolumeClaim is not bound: default/noc-clickhouse
wait.go:218: [debug] PersistentVolumeClaim is not bound: default/noc-clickhouse
wait.go:218: [debug] PersistentVolumeClaim is not bound: default/noc-clickhouse
wait.go:218: [debug] PersistentVolumeClaim is not bound: default/noc-clickhouse
wait.go:218: [debug] PersistentVolumeClaim is not bound: default/noc-clickhouse
wait.go:218: [debug] PersistentVolumeClaim is not bound: default/noc-clickhouse
wait.go:218: [debug] PersistentVolumeClaim is not bound: default/noc-clickhouse
wait.go:218: [debug] PersistentVolumeClaim is not bound: default/noc-clickhouse
wait.go:218: [debug] PersistentVolumeClaim is not bound: default/noc-clickhouse
wait.go:218: [debug] PersistentVolumeClaim is not bound: default/noc-clickhouse
wait.go:218: [debug] PersistentVolumeClaim is not bound: default/noc-clickhouse
wait.go:218: [debug] PersistentVolumeClaim is not bound: default/noc-clickhouse
wait.go:218: [debug] PersistentVolumeClaim is not bound: default/noc-clickhouse
[...]

PR #6337, opened 2019-09-02
Update debug to have timestamps

waveywaves:
This PR adds timestamp to debug statements
fixes https://github.com/helm/helm/issues/6336

**If applicable**:

### DIFF
--- a/cmd/helm/helm.go
+++ b/cmd/helm/helm.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
@@ -55,7 +56,8 @@ func init() {
 
 func debug(format string, v ...interface{}) {
 	if settings.Debug {
-		format = fmt.Sprintf("[debug] %s\n", format)
+		timeNow := time.Now().UTC().String()
+		format = fmt.Sprintf("%s [debug] %s\n", timeNow, format)
 		log.Output(2, fmt.Sprintf(format, v...))
 	}
 }


waveywaves:
cc @bacongobbler Can you take a look at this.
@freeseacher What do you think about this. Should I make any more changes ?

cmd/helm/helm.go
```diff
@@ -55,7 +56,8 @@ func init() {
 
 func debug(format string, v ...interface{}) {
 	if settings.Debug {
-		format = fmt.Sprintf("[debug] %s\n", format)
+		timeNow := time.Now().UTC().String()
```
thomastaylor312: This should probably use something like RFC3339 formatted timestamps instead of just the raw string implementation

thomastaylor312 (changes requested):
I wanted to have a discussion around whether or not timestamps would be valuable here or not. Because this is purely for the CLI, I am not entirely sure they will be useful because each command is relatively short-lived. I can see the argument for having it in the logging that comes with `--wait` though, so I am completely open for discussion on this

cmd/helm/helm.go
thomastaylor312: But let me check on the time format before doing anything here

thomastaylor312: I think I'd go with RFC3339, unless you have inclinations otherwise

thomastaylor312:
@waveywaves As I mentioned in my comment, I don't think adding timestamps is really necessary seeing as this is a function used purely by the CLI. Anyone who uses it as a library can attach their own logger with timestamps and if you are using it in CI or other automated usage, timestamps are automatically attached. Please let me know if there is a good argument for them I am missing

freeseacher:
>timestamps are automatically attached
sorry, but they are not. gitlab does not provide such functionality :( 
probably it is corner case and timestpamping should be added with flags

bacongobbler:
Thanks @waveywaves for the contribution. Right now the core maintainers are trying to focus on bug fixes so we can release Helm 3.0 in a timely manner. Can we re-discuss this feature after Helm 3.0 has been released?

waveywaves:
@bacongobbler @thomastaylor312 From my point of view as well, this does not seem to be something that is extremely critical. I have only updated this wrt the feature proposal by @freesearcher.

bacongobbler:

freeseacher:
please reopen it. still usefull in my opinion

bacongobbler:
@freeseacher once the base branch has been closed, the PR cannot be re-opened. That's why I'm asking @waveywaves to open a new PR.

This was a clerical error on our end as part of the migration. Sorry about that.

PR #6341, opened 2019-09-02
Moving initActionConfig to pkg

### LINKED ISSUE #6333: Add function to create configuration for actions in pkg
I'm playing around with the v3 beta, and trying to refactor the terraform provider. There doesn't seem to be an easy way to create a configuration, and I feel like I am duplicating a bunch of logic from initActionConfig in cmd/helm. It would be super helpful if there some methods added to build the configuration object in pkg

aaronmell:
Moved initActionConfig to pkg to make it easier to instantiate the configuration for projects consuming helm

closes #6333 

Signed-off-by: Aaron Mell <amell@lumindigital.com>

I was trying migrate the terraform helm provider to v3 beta. Realized that I was basically copying most of the logic in cmd/helm/helm.go to instantiate the configuration. Figured other people would have to do the same, and it should be easier to create the config without knowing about the internals of helm. 

I'm pretty new to go, so I'd appreciate candid feedback.

**If applicable**:

### DIFF
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -19,12 +19,10 @@ package action
 import (
 	"path"
 	"regexp"
-	"sync"
 	"time"
 
 	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
-	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
@@ -53,9 +51,6 @@ var (
 	errInvalidRevision = errors.New("invalid release revision")
 	// errInvalidName indicates that an invalid release name was provided
 	errInvalidName = errors.New("invalid release name, must match regex ^(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])+$ and the length must not longer than 53")
-
-	config     genericclioptions.RESTClientGetter
-	configOnce sync.Once
 )
 
 // ValidName is a regular expression for names.
@@ -218,7 +213,7 @@ func (c *Configuration) recordRelease(r *release.Release) {
 func InitActionConfig(envSettings *cli.EnvSettings, allNamespaces bool, helmDriver string, log DebugLog) (*Configuration, error) {
 
 	var actionConfig Configuration
-	kubeconfig := kubeConfig(envSettings)
+	kubeconfig := envSettings.KubeConfig()
 
 	kc := kube.New(kubeconfig)
 	kc.Log = log
@@ -229,7 +224,7 @@ func InitActionConfig(envSettings *cli.EnvSettings, allNamespaces bool, helmDriv
 	}
 	var namespace string
 	if !allNamespaces {
-		namespace = GetNamespace(envSettings)
+		namespace = envSettings.Namespace()
 	}
 
 	var store *storage.Storage
@@ -256,23 +251,4 @@ func InitActionConfig(envSettings *cli.EnvSettings, allNamespaces bool, helmDriv
 	actionConfig.Log = log
 
 	return &actionConfig, nil
-}
-
-func kubeConfig(envSettings *cli.EnvSettings) genericclioptions.RESTClientGetter {
-	configOnce.Do(func() {
-		config = kube.GetConfig(envSettings.KubeConfig, envSettings.KubeContext, envSettings.Namespace)
-	})
-	return config
-}
-
-//GetNamespace gets the namespace from the configuration
-func GetNamespace(envSettings *cli.EnvSettings) string {
-	if envSettings.Namespace != "" {
-		return envSettings.Namespace
-	}
-
-	if ns, _, err := kubeConfig(envSettings).ToRawKubeConfigLoader().Namespace(); err == nil {
-		return ns
-	}
-	return "default"
 }
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -19,19 +19,23 @@ package action
 import (
 	"path"
 	"regexp"
+	"sync"
 	"time"
 
 	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 
 	"helm.sh/helm/v3/internal/experimental/registry"
 	"helm.sh/helm/v3/pkg/chartutil"
+	"helm.sh/helm/v3/pkg/cli"
 	"helm.sh/helm/v3/pkg/kube"
 	"helm.sh/helm/v3/pkg/release"
 	"helm.sh/helm/v3/pkg/storage"
+	"helm.sh/helm/pkg/storage/driver"
 )
 
 // Timestamper is a function capable of producing a timestamp.Timestamper.
@@ -49,6 +53,10 @@ var (
 	errInvalidRevision = errors.New("invalid release revision")
 	// errInvalidName indicates that an invalid release name was provided
 	errInvalidName = errors.New("invalid release name, must match regex ^(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])+$ and the length must not longer than 53")
+
+	config     genericclioptions.RESTClientGetter
+	configOnce sync.Once
+	settings   *cli.EnvSettings
 )
 
 // ValidName is a regular expression for names.
@@ -81,6 +89,15 @@ type Configuration struct {
 
 	Log func(string, ...interface{})
 }
+
+// RESTClientGetter gets the rest client
+type RESTClientGetter interface {
+	ToRESTConfig() (*rest.Config, error)
+	ToDiscoveryClient() (discovery.CachedDiscoveryInterface, error)
+	ToRESTMapper() (meta.RESTMapper, error)
+}
+
+type debug func(format string, v ...interface{})
 
 // capabilities builds a Capabilities from discovery information.
 func (c *Configuration) getCapabilities() (*chartutil.Capabilities, error) {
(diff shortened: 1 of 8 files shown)


aaronmell:
Hmm, I get a different result when I run the tests locally.

aaronmell:
My second commit seems a bit much. I'm not sure if there was a better way to handle the change that I made.

thomastaylor312:
Hey @aaronmell thanks for the PR! So I haven't done a full review yet, but the refactor generally looks good and I'm super happy to see some feedback around using Helm as a library. My concern with this PR is around exporting a global data struct as part of the API, which as far as I understand is bad practice. I know that is how it was before, but I was _ok_ with it at best when we had it unexported. 

Based on what I see in the code however, we just need to pass the parameters for the KubeConfig function in (or maybe even just pass in the `RESTClientGetter`, it will be created by default if `nil` is passed) and `Debug` can just be put back in `cmd` and be unexported. You could still have the global settings object unexported like before in the `cmd` package. Any thoughts? I'd be willing to pair with you if needed as well

aaronmell:
@thomastaylor312 I think I understand what your asking. The second commit I made felt wrong, so I'll back it out, and give it another go.

thomastaylor312:
Thanks @aaronmell. Like I said, if you need a second set of eyes, let me know

aaronmell:
@thomastaylor312 how does it look now?

pkg/action/action.go
```diff
@@ -82,6 +89,15 @@ type Configuration struct {
...
 }
 
+// RESTClientGetter gets the rest client
+type RESTClientGetter interface {
+	ToRESTConfig() (*rest.Config, error)
+	ToDiscoveryClient() (discovery.CachedDiscoveryInterface, error)
+	ToRESTMapper() (meta.RESTMapper, error)
+}
+
+type debug func(format string, v ...interface{})
```
thomastaylor312: Because this type is used in the exported `InitActionConfig`, it should be exported as well. We also might want a more specific name like `DebugLog`

thomastaylor312 (changes requested):
OK, just one needed change on this from what I can see. I am going to test it in the meantime to make sure everything is working well!

pkg/action/action.go
aaronmell: Changes Made

thomastaylor312 (approved):
Thanks for all of the hard work on this @aaronmell! I'll get one of the other maintainers to review as well due to this being marked as a large PR

aaronmell:
@thomastaylor312  Any update?

bacongobbler:
All of us were away for Helm Summit EU this week. Should be back by next week to take another look

aaronmell:
@bacongobbler Awesome, Thanks!

aaronmell:
Any chance of getting a review this week?

bacongobbler:
throwing this on the 3.0.0-beta.4 milestone so we don't miss this.

Sorry for the delay on a review @aaronmell. Right now we're busy reviewing the last few PRs for Helm 2.15. Once that is complete, we can circle back to this PR, merge it, then start the release process for 3.0.0-beta.4.

pkg/action/action.go
```diff
@@ -19,19 +19,23 @@ package action
...
 	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 
 	"helm.sh/helm/internal/experimental/registry"
 	"helm.sh/helm/pkg/chartutil"
+	"helm.sh/helm/pkg/cli"
```
mattfarina: Helm moved to Go modules. That means the major version needs to be in the import path. This applies when something has a major version of 2 or higher. So, the helm import paths need to have a `/v3` in them. For example,

`"helm.sh/helm/pkg/cli"` -> `"helm.sh/helm/v3pkg/cli"`

aaronmell: I rebased with the latest. Should be good now.

pkg/action/action.go
```diff
@@ -197,8 +214,65 @@ func (c *Configuration) recordRelease(r *release.Release) {
...
+
+func kubeConfig(envSettings *cli.EnvSettings) genericclioptions.RESTClientGetter {
+	configOnce.Do(func() {
+		config = kube.GetConfig(envSettings.KubeConfig, envSettings.KubeContext, envSettings.Namespace)
+	})
+	return config
+}
+
+//GetNamespace gets the namespace from the configuration
+func GetNamespace(envSettings *cli.EnvSettings) string {
```
bacongobbler: Can you make this part of the `*cli.EnvSettings` object? e.g. in `pkg/cli`:

```
func (e *EnvSettings) Namespace() string {
    if e.Namespace != "" {
        return e.Namespace
    }

    if ns, _, err := e.kubeConfig().ToRawKubeConfigLoader().Namespace(); err != nil {
        return ns
    }

    return "default"
}
```

pkg/action/action.go
```diff
@@ -197,8 +214,65 @@ func (c *Configuration) recordRelease(r *release.Release) {
...
+
+	actionConfig.RESTClientGetter = kubeconfig
+	actionConfig.KubeClient = kc
+	actionConfig.Releases = store
+	actionConfig.Log = log
+
+	return &actionConfig, nil
+}
+
+func kubeConfig(envSettings *cli.EnvSettings) genericclioptions.RESTClientGetter {
```
bacongobbler: Can you make this part of the `*cli.EnvSettings` object? e.g. in `pkg/cli`:

```
func (e *EnvSettings) KubeConfig() genericclioptions.RESTClientGetter {
    configOnce.Do(func() {
        config = kube.GetConfig(envSettings.KubeConfig, envSettings.KubeContext, envSettings.Namespace)
    })
    return config
}
```

Then `InitActionConfig` can call `envSettings.KubeConfig()`.

bacongobbler (changes requested):

pkg/action/action.go
aaronmell: I think I captured the intent here.

pkg/action/action.go
aaronmell: I think I captured the intent here

go.mod
```diff
@@ -123,13 +126,15 @@ require (
 	google.golang.org/grpc v1.23.0
 	gopkg.in/gorp.v1 v1.7.2 // indirect
 	gopkg.in/yaml.v2 v2.2.2
+	helm.sh/helm v2.14.3+incompatible
```
bacongobbler: looks like the go.mod file sneaked in a few accidental adds. Would you mind reverting these?

pkg/cli/environment.go
```diff
@@ -26,18 +26,20 @@ import (
 	"fmt"
 	"os"
 	"strconv"
+	"sync"
 
 	"github.com/spf13/pflag"
 
 	"helm.sh/helm/v3/pkg/helmpath"
+
+	"k8s.io/cli-runtime/pkg/genericclioptions"
```
bacongobbler: can you fix up these package imports?

pkg/cli/environment.go
```diff
@@ -53,9 +55,20 @@ type EnvSettings struct {
 	PluginsDirectory string
 }
 
+var (
+	config     genericclioptions.RESTClientGetter
+	configOnce sync.Once
+
+	// Namespace is the namespace scope.
+	namespace string
```
bacongobbler: This change makes `namespace` a shared global variable between all EnvSettings objects. Is there a reason this was refactored out of the EnvSettings object?

This should be an internal variable of the EnvSettings object. e.g.

```
// EnvSettings describes all of the environment settings.
type EnvSettings struct {
	// namespace is the namespace scope.
	namespace string
```

pkg/cli/environment.go
```diff
@@ -53,9 +55,20 @@ type EnvSettings struct {
...
+	// KubeConfig is the path to the kubeconfig file.
+	kubeConfig string
```
bacongobbler: Same here RE: making this a shared global variable. This should be part of the EnvSettings object as it was before, but as an internal variable.

bacongobbler (changes requested):

pkg/cli/environment.go
aaronmell: Done

bacongobbler (approved):
looking muuuuuuuuch better. Thank you!

I noticed that @thomastaylor312 self-requested another review so I'll hold off merging for now and let him merge after re-reviewing this again.

bacongobbler:
Once last CI error:

> GO111MODULE=on /go/bin/golangci-lint run
> pkg/cli/environment.go:32: File is not `goimports`-ed with -local helm.sh/helm (goimports)

> make: *** [Makefile:85: test-style] Error 1

pkg/cli/environment.go
```diff
@@ -26,23 +26,23 @@ import (
 	"fmt"
 	"os"
 	"strconv"
+	"sync"
 
 	"github.com/spf13/pflag"
 
 	"helm.sh/helm/v3/pkg/helmpath"
+	"helm.sh/helm/v3/pkg/kube"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
```
bacongobbler: here's the line failing CI. This should be moved up beside the pflag import.

General practice for Go import statements is

```
import (
    // standard libraries

    // third-party packages

    // local packages
)
```

bacongobbler (changes requested):

pkg/cli/environment.go
aaronmell: Yeah, It took me a bit to figure that out. Running make format fixed it up for me.

pkg/cli/environment.go
```diff
@@ -93,3 +100,23 @@ func (s *EnvSettings) EnvVars() map[string]string {
...
+
+	if ns, _, err := s.KubeConfig().ToRawKubeConfigLoader().Namespace(); err == nil {
+		return ns
+	}
+	return "default"
+}
+
+//KubeConfig gets the kubeconfig from EnvSettings
+func (s *EnvSettings) KubeConfig() genericclioptions.RESTClientGetter {
+	configOnce.Do(func() {
```
bacongobbler: Okay, one last thing I'm noticing as I'm thinking through this design.

Given the following use case:

```
s := New()
s.KubeContext = "foo"
s2 := New()
s2.KubeContext = "bar"
...
// insert other setup code required here
```

With this in mind, I'd expect that the following would result in two unique kubeconfigs:

```
s.KubeConfig()
s2.KubeConfig()
```

Instead, what we'll end up with is the same kube config. This is because they are both using shared global flags.

Can we make `configOnce` and `config` part of the EnvSettings object as well? Thanks!

bacongobbler (changes requested):

thomastaylor312 (approved):

bacongobbler (approved):

PR #6353, opened 2019-09-04
Environment variable cleanup

mattfarina:
* Renames environment variables for consistent use
* Exposes registry to plugins as env var like the rest
* Removes duplicates from env vars in internal settings
* Display of env vars is alphanum sort ordered


thomastaylor312 (approved):
LGTM!

bacongobbler (approved):

PR #6177, opened 2019-08-08
fix(chartutil): Ensure nested template dir on save

jdolitsky:
If a templates/ dir of a chart contained a subdirectory, for example "templates/tests/test-db.yaml", an error was being thrown on export due to missing the "templates/test" directory prior to saving the template file itself.

Example chart with nested template, stable/wordpress: https://github.com/helm/charts/blob/master/stable/wordpress/templates/tests/test-mariadb-connection.yaml

This was apparently being done for chart.Files but not chart.Templates - I've combined the logic for both fields.

Fixes #5757

cc @bacongobbler

PR #6360, opened 2019-09-05
Fix chartutil.Save returning empty path when passing a non-existent directory (Resolves #6344)

### LINKED ISSUE #6344: Helm package --destination needs to run twice
When running `helm package` with `--destination ./helm/tarballs` and this directory did not exist before, then the command needs to be run twice to succeed:

```
$ rm -rf helm/tarballs/
$ helm package .  --destination ./helm/tarballs/
Successfully packaged chart and saved it to: 
$ helm package .  --destination ./helm/tarballs/
Successfully packaged chart and saved it to: helm/tarballs/mytest-0.1.0.tgz
```
The chart tarball is only written on the second execution. 

Output of `helm version`: v3.0.0.beta.2

Output of `kubectl version`: v1.15.3

wxdao:

Fix chartutil.Save returning empty path when passing a non-existent directory, which causes `helm package --destination dir` has to be run twice when `dir` doesn't exist (#6344).

**If applicable**:

### DIFF
--- a/pkg/chartutil/save.go
+++ b/pkg/chartutil/save.go
@@ -93,7 +93,7 @@ func Save(c *chart.Chart, outDir string) (string, error) {
 	filename := fmt.Sprintf("%s-%s.tgz", c.Name(), c.Metadata.Version)
 	filename = filepath.Join(outDir, filename)
 	if stat, err := os.Stat(filepath.Dir(filename)); os.IsNotExist(err) {
-		if err := os.MkdirAll(filepath.Dir(filename), 0755); !os.IsExist(err) {
+		if err := os.MkdirAll(filepath.Dir(filename), 0755); err != nil {
 			return "", err
 		}
 	} else if !stat.IsDir() {
--- a/pkg/chartutil/save_test.go
+++ b/pkg/chartutil/save_test.go
@@ -19,6 +19,7 @@ package chartutil
 import (
 	"io/ioutil"
 	"os"
+	"path"
 	"path/filepath"
 	"strings"
 	"testing"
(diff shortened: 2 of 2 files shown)


pkg/chartutil/save_test.go
```diff
@@ -33,37 +34,39 @@ func TestSave(t *testing.T) {
...
-	c2, err := loader.LoadFile(where)
-	if err != nil {
-		t.Fatal(err)
-	}
-	if c2.Name() != c.Name() {
-		t.Fatalf("Expected chart archive to have %q, got %q", c.Name(), c2.Name())
-	}
-	if len(c2.Files) != 1 || c2.Files[0].Name != "scheherazade/shahryar.txt" {
-		t.Fatal("Files data did not match")
+	for _, dest := range []string{tmp, path.Join(tmp, "newdir")} {
```
bacongobbler: Would you mind changing this to a `t.Run`? That way, these two tests cases can run in parallel. https://golang.org/pkg/testing/#hdr-Subtests_and_Sub_benchmarks

bacongobbler (changes requested):

pkg/chartutil/save_test.go
wxdao: @bacongobbler Changed it. Thank you!

wxdao:
Hi @bacongobbler. When will you be available to review my PR? I've just resolved a conflict to the upstream. It seems that another PR #6177 has applied the same fix with different motivation. However I think my PR still makes sense since it adds an essential test to avoid future mistakes.

bacongobbler:
Hi @wxdao, sorry for not getting back to you sooner. Right now we are trying to arrange PR reviews for the 3.0.0-beta.4 release, making sure that things are merged smoothly. I'll put this on the beta.4 milestone so we don't lose track of this bugfix. Thanks again!

pkg/chartutil/save.go
```diff
@@ -93,7 +93,7 @@ func Save(c *chart.Chart, outDir string) (string, error) {
 	filename := fmt.Sprintf("%s-%s.tgz", c.Name(), c.Metadata.Version)
 	filename = filepath.Join(outDir, filename)
 	if stat, err := os.Stat(filepath.Dir(filename)); os.IsNotExist(err) {
-		if err := os.MkdirAll(filepath.Dir(filename), 0755); !os.IsExist(err) {
+		if err := os.MkdirAll(filepath.Dir(filename), 0755); err != nil {
```
hickeyma: Will this fail if the directory already exists? Is this expected  behaviour?

If not then maybe the check should be something like this:
`if err != nil && !os.IsExist(err)`

hickeyma (changes requested):

pkg/chartutil/save.go
wxdao: Thanks for reviewing this. It seems that os.MkdirAll will only be called when the outer if fails, i.e. os.IsNotExist == true. So it wouldn't try to mkdir an existing directory, isn't it?

hickeyma: That is correct. From API: "If path is already a directory, MkdirAll does nothing and returns nil."

hickeyma (approved):
lGTM, thanks @wxdao.

Issue #3211, opened 2017-11-29
Test pod not deleted with helm.sh/hook-delete-policy


unguiculus:
I have a chart with tests. The test pod is annotated as follows:

```yaml
annotations:
    helm.sh/hook: test-success
    helm.sh/hook-delete-policy: hook-succeeded
```

I would expect the pod to be deleted once the test has run successfully. This is, however, not the case.

```
$ helm version
Client: &version.Version{SemVer:"v2.7.2", GitCommit:"8478fb4fc723885b155c924d1c8c410b7a9444e6", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.7.2", GitCommit:"8478fb4fc723885b155c924d1c8c410b7a9444e6", GitTreeState:"clean"}
```

jroucheton:
Hi,

I have the same kind of issue.
I defined the following job:

`apiVersion: batch/v1
kind: Job
metadata:
  name: {{ .Chart.Name }}
{{ include "{{ .Chart.Name }}.labels" . | indent 2 }}
  annotations:
    "helm.sh/hook": pre-install,pre-upgrade
    "helm.sh/hook-weight": "1"
    "helm.sh/hook-delete-policy": hook-succeeded
.......`

But even if this job has been executed with success, this one is still created....

Regards,

unguiculus:
@jroucheton My issue is specifically about tests. For other hooks `helm.sh/hook-delete-policy` works as expected for me.

ericleu:
Any update on this issue?  I am experiencing the exact same issue.  Test pod not deleted after success result.

natcho:
I have also observed this issue.

bacongobbler:
re-labeling as a feature if someone wants to tackle this!

rrpolanco:
FWIW I've been able to work around this by telling `helm test` to clean up after itself via `--cleanup` flag.

jlegrone:
This should be closed by https://github.com/helm/helm/pull/6054.

PR #6363, opened 2019-09-05
Remove run test subcommand and --cleanup flag, print test status

jlegrone:

Related to #6020. This PR removes the `helm test run` subcommand and `--cleanup` flag. The output of `helm test --help` is now:

```
The test command runs the tests for a release.

The argument this command takes is the name of a deployed release.
The tests to be run are defined in the chart that was installed.

Usage:
  helm test [RELEASE] [flags]

Flags:
  -h, --help               help for test
      --timeout duration   time to wait for any individual Kubernetes operation (like Jobs for hooks) (default 5m0s)
```

More output has also been added during execution of `helm test`. The output of `helm test <release>` now looks like:

```
Pod nginx-test-bar pending
Pod nginx-test-bar pending
Pod nginx-test-bar succeeded
Pod nginx-test-foo pending
Pod nginx-test-foo pending
Pod nginx-test-foo succeeded
TEST SUITE:     nginx-test-foo
Last Started:   2019-09-05 15:04:12.33368 -0400 EDT m=+5.107841576
Last Completed: 2019-09-05 15:04:14.147617 -0400 EDT m=+6.921723894
Phase:          Succeeded
TEST SUITE:     nginx-test-bar
Last Started:   2019-09-05 15:04:08.31585 -0400 EDT m=+1.090131819
Last Completed: 2019-09-05 15:04:12.314482 -0400 EDT m=+5.088644403
Phase:          Succeeded
```

For now in order to perform test resource cleanup, users will need to set `helm.sh/hook-delete-policy`, manually delete resources with kubectl, or set [TTL for Job resources](https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/). The hook labels proposal in https://github.com/helm/helm/issues/6365 should improve on the ergonomics of using kubectl for cleanup if we don't implement something directly in helm first.

**If applicable**:


cmd/helm/release_testing.go
```diff
@@ -18,28 +18,39 @@ package main
...
 
+	"helm.sh/helm/cmd/helm/require"
 	"helm.sh/helm/pkg/action"
 )
 
 const releaseTestHelp = `
-The test command consists of multiple subcommands around running tests on a release.
-
-Example usage:
-    $ helm test run [RELEASE]
```
marckhouzam: Currently `helm test run <TAB>` will provide dynamic completion by listing available releases.

If you remove the `run` subcommand, could you update the dynamic completion code by changing the string `helm_test_run` to `helm_test` in `cmd/helm/root.go`?

Thank you.

jlegrone: Nice catch @marckhouzam, updated.

technosophos (approved):
Code LGTM

scottrigby (approved):
haven't tested, but look good. also yay for simplifying 💯

bacongobbler:
If at all possible, can you (the next core maintainer to review this PR) please hold on merging this until I get back next week? I want to take a closer look at the ramifications of removing the `--cleanup` flag when it comes to the user experience. I want to make sure we have those user stories captured and well-understood before we decide to remove a fairly significant feature of `helm test` in favour of an alternative approach (hook annotations).

unguiculus:
If I understand it correctly, #3211 is fixed in v3?

jlegrone:
> If I understand it correctly, #3211 is fixed in v3?

Yes, I wasn't aware of that issue but it should be closed by https://github.com/helm/helm/pull/6054.

cmd/helm/release_testing.go
```diff
@@ -18,28 +18,39 @@ package main
 
 import (
 	"io"
+	"time"
 
 	"github.com/spf13/cobra"
 
+	"helm.sh/helm/cmd/helm/require"
```
mattfarina: Helm moved to Go modules. That means the major version needs to be in the import path. This applies when something has a major version of 2 or higher. So, the helm import paths need to have a /v3 in them. For example,

`"helm.sh/helm/cmd/helm/require"` -> `"helm.sh/helm/v3/cmd/helm/require"`

pkg/action/release_testing.go
```diff
@@ -45,39 +40,21 @@ func NewReleaseTesting(cfg *Configuration) *ReleaseTesting {
...
-		return err
+		return rel, err
 	}
 
 	if err := r.cfg.execHook(rel, release.HookTest, r.Timeout); err != nil {
 		r.cfg.Releases.Update(rel)
-		return err
-	}
-
-	if r.Cleanup {
```
bacongobbler: First off, thank you for taking the time to refactor `helm test`. We really appreciate the work you've put into this.

That being said, we cannot remove this feature and replace it with hook annotations. This feature serves a very different purpose than what the hook annotations provide, which is why `helm test` was not originally designed around the hook system.

A very simple test where this flag is required:

1. I have a chart that someone else has built (e.g. `stable/wordpress`)
1. I want to run the tests
1. I want to have the test hooks cleaned up in preparation for the next test run

Prior to this change, the use case was done using the following commands:

```
$ helm install wordpress stable/wordpress
$ helm test run wordpress --cleanup
```

Now with this functionality removed, a user must now perform the following:

```
$ helm fetch stable/wordpress --untar
$ $EDITOR wordpress/templates/tests/
$ helm install wordpress ./wordpress
$ helm test wordpress
```

So now, the user must MODIFY the chart before installing to handle the use case. If they don't want to clean up test hooks after the test run, then they have to modify the chart AGAIN, run `helm upgrade` and finally run `helm test`. This is a massive barrier of entry for users who just want their test hooks to be cleaned up after running `helm test`. It's for that reason this functionality must remain as-is. I don't see a way around this with the way hook annotations work today, sadly.

bacongobbler (changes requested):

bacongobbler:
All that being said, once I rebased this PR on master, everything seems to work as described. Thank you!

pkg/action/release_testing.go
jlegrone: Perhaps we should schedule some time to discuss this; I think the workflow concerns here are addressed by https://github.com/helm/helm/issues/6365.

When those changes are combined with https://github.com/helm/helm/issues/6364 and the proposal to amend the default `hook-delete-policy` from https://github.com/helm/helm/issues/6020 I think we should be able to deliver a much nicer experience than Helm 2 out of the box. This PR is not representative of the whole picture.

bacongobbler (commented):
After discussion on the dev call, we have a path forward and this is a step in that direction. Once this has been rebased and @mattfarina's comments have been addressed we can get this one in. Thanks!

pkg/action/release_testing.go
bacongobbler: Discussed offline. Workflow concerns are indeed addressed through other proposed paths.

bacongobbler (approved):
LGTM! Will need another maintainer to sign off since this is a `size/L`

bacongobbler:
Oh! I see @technosophos LGTM'd this change a while back. Merging!

PR #6368, opened 2019-09-06
Support the legacy client flag for better backward compatibility

### LINKED ISSUE #6322: Feature request: a `helm version` option that works consistently across v2 and v3

RyanSiu1995:
close #6322 

**If applicable**:

### DIFF
--- a/cmd/helm/testdata/output/version-client.txt
+++ b/cmd/helm/testdata/output/version-client.txt
@@ -1,0 +1,1 @@
+Client: version.BuildInfo{Version:"v3.0+unreleased", GitCommit:"", GitTreeState:"", GoVersion:""}
--- a/cmd/helm/testdata/output/version-short-client.txt
+++ b/cmd/helm/testdata/output/version-short-client.txt
@@ -1,0 +1,1 @@
+Client: v3.0+unreleased
--- a/cmd/helm/version.go
+++ b/cmd/helm/version.go
@@ -43,6 +43,7 @@ version.BuildInfo{Version:"v2.0.0", GitCommit:"ff52399e51bb880526e9cd0ed8386f643
 
 type versionOptions struct {
 	short    bool
+	client   bool
 	template string
 }
 
@@ -60,6 +61,7 @@ func newVersionCmd(out io.Writer) *cobra.Command {
 	}
 	f := cmd.Flags()
 	f.BoolVar(&o.short, "short", false, "print the version number")
+	f.BoolVarP(&o.client, "client", "c", false, "print the version number")
 	f.StringVar(&o.template, "template", "", "template for version string format")
 
 	return cmd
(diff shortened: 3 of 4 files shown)


marckhouzam:
As an example, in the acceptance-testing repo we've used
```
if helm version -c &> /dev/null; then
    // v2
else
    //v3
fi
```
which is pretty straightforward.

Adding the legacy client flag to helm v3 would actually make such a check slightly more complex as we'd need to look for a version within the output of `helm version -c`.

But I agree that for consumers it may be more intuitive to parse the result than to look for an error as a way to differentiate between v2 and v3.

Let's see what the maintainers think.

mattfarina:
This is a tough one. What is the best UX, what will people discover for handling this problem, and will people need to detect sub-versions (i.e., 2.14 or 3.1)? I'm not sure of the best path forward. I'll send an email to the helm mailing list asking folks about it. I'm not opposed to either path. I'm just not sure of the better path.

mattfarina:
For reference, that email went out and is at https://lists.cncf.io/g/cncf-helm/topic/detecting_helm_v2_and_v3_in/34288354?p=,,,20,0,0,0::recentpostdate%2Fsticky,,,20,2,0,34288354

paulczar:
would it makes sense to have `-c` return a deprecated message to stderr ?    that way a human should see its not useful anymore, but would still allow anything expecting a straightforward version in stdout work.

paulczar:
otherwise I'm kind of in the camp of using `-c` to detect between helm 2 and helm 3.  If somebody needs to go deeper in version numbers they can further use `helm version --short`  in each case.

on that note, it would be nice if `helm version` responded with a `jq` friendly output for easier parsing.

```
$ helm version         
$ helm version | jq
parse error: Invalid numeric literal at line 1, column 18

```

marckhouzam:
I was thinking that if the `-c` flag is added to v3, a version check that doesn't require parsing could then be:
```
if helm version -c --tls &> /dev/null; then
    // v2
else
    //v3
fi
```
since the `tls` flag is only in v2.

That being said, we're falling into really non-intuitive ways of doing things.

itowlson:
I'm the one who originally requested this feature.  I am not very keen on the "if `-c` works then it must be v2 else it must be v3" strategy - I want to ask Helm what version it is, and I _want_ to parse the answer out of the output (heck, that's what the output's _for_), rather than going "let me pass flags that I know happen to work only in such-and-such version and infer the version from success or failure."  It feels more intention-revealing - otherwise the maintainer has to know that `-c` is a magic flag that only works in v2.

I do get where @marckhouzam is coming from though - I'm in JavaScript so it's easy for me to parse values out of an output, but for a bash script it may be easier examine an exit code, so I can see why being able to just plain fail on `-c` would make it easier for bash to distinguish.  But, as Marc notes in his first comment, parsing the output feels a lot more intuitive to me!  And come Helm 4, the acceptance tests may have to parse the output anyway...! *evil grin*

marckhouzam:
I'm convinced.
For the majority of users, parsing the version output will be the expected way of doing things, and they they'll be looking for a format of `helm version` that will work on both v2 and v3. So they will need the `-c` flag.

So I vote for adding the `-c` flag.
Should it be hidden or should it be described in the help and work with completion^

P.S.:
Parsing the version with the `-c` flag using bash turns out to be pretty simple:
```
if helm version -c |grep -q :\"v3; then
  // v3
else
  // v2
fi
```

bacongobbler:
I think we should go ahead and accept this one given there is user demand for it, even if it's a bit of a kludge. However, I do not think the `Client:` output should have been added. We should just accept the `-c|--client-only` only flag and do nothing with it. `helm version` already shows the client output.

I'll file a PR to follow-up on this. I'll also see if we can mark this as a hidden flag.

bacongobbler:
closing as this has been superseded by #6670. Thanks @RyanSiu1995!

PR #6385, opened 2019-09-07
fix(helm3): Detailed exit code for helm plugins

### LINKED ISSUE #6384: helm-3.0.0-beta.3: unable to return non-zero exit codes other than 1 from helm plugins
While testing [helm-diff](https://github.com/databus23/helm-diff) with helm-3.0.0-beta.3, I found that there's a regression - it is missing the fix for #4170.

`helm-diff` needs to return 1 on errors and 2 on changes detected. helm 2.x had also been returning 1 for all the non-zero exit codes from plugins but it was fixed via #4367.

I've confirmed that it isn't fixed for helm v3 and reading code from the `dev-v3` branch I've also confirmed that it misses changes from #4367.

Can we just cherry-pick the fix into the helm3 dev branch?

---

Output of `helm version`:

```
```

mumoshu:
I've verified this working as expected in combination with helmfile v0.83.0, and helm-diff build from https://github.com/databus23/helm-diff/pull/149.

Fixes #6384

### DIFF
--- a/cmd/helm/helm.go
+++ b/cmd/helm/helm.go
@@ -79,7 +79,12 @@ func main() {
 
 	if err := cmd.Execute(); err != nil {
 		debug("%+v", err)
-		os.Exit(1)
+		switch e := err.(type) {
+		case pluginError:
+			os.Exit(e.code)
+		default:
+			os.Exit(1)
+		}
 	}
 }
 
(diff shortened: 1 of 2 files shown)


bacongobbler (approved):
Thanks for porting this fix over!

PR #6388, opened 2019-09-08
v2: Allow to run acceptance tests from main Helm repo

marckhouzam:
This is a cherry-pick from v3 #6256 

This allows to run the acceptance tests directly from the v2 repo as we have for v3.
As more acceptance tests are developed, it is necessary to test them with both v3 and v2, therefore it would be helpful to have these make targets in v2 also.

To run the acceptance tests, one can now do:
```
    make test-acceptance
```
To run the completion tests, one can now do:
```
```

Currently, only the acceptance completion tests actually run with v2.  The others are skipped and marked as passed.

Also, https://github.com/helm/acceptance-testing/pull/34 is required to have all the acceptance tests pass with the current v2.


marckhouzam:
This change isn't very important anymore as no new features will be include in v2.
The acceptance tests can still be run against v2 manually instead.

Issue #1904, opened 2017-02-02
Optimize scanReposForURL function


technosophos:
Per #1855, there is probably a way to optimize the traversal algorithm (maybe with a cache) that is used by scanReposForURL()

thomastaylor312:
@technosophos This is probably a good one to try and do as part of 2.4 given its focus 😄, should we bring it in?

irajdeep:
@technosophos I am planning to work on this if no one else is currently working on it :)

thomastaylor312:
@irajdeep Sounds great! I would target Helm v3 as we are mostly focusing on bug fixes for v2

irajdeep:
sounds cool, I will start working on it then

PR #6405, opened 2019-09-10
[Proposal] Implemented secondary repository indexing refs #1904

osdrv:

This commit is an attempt to address concerns scoped in #1904.

`ChartDownloader.scanReposForURL()` is called when a fetching chart is
identified by an absolute URL and no trivial reference between the chart
and parental repository could be established. The operation requires a
full linear repository scan. The runtime performance degenerates in a
case of a non-existing URL as a full scan goes over the full range of
existing repos, all their ChartVersion entries and each version of an entry.

The motivation for this change is to speed up this operation, ideally
make the aforementioned operation to run in nearly constant amortised time (a
detailed performance analysis is presented below).

This PR is a consolidation of my and @irajdeep 's thoughts on the way to address #1904. I mark the PR as draft yet as the main intention is to hear an early feedback from the core team on the general course of the proposed change. We do understand we increase the code complexity and introduce a new tiny but still spinning wheel for the sake of optimisation of a reasonably rare use case. On the other hand, this counter-intuitive ratio of the change size to the improvement impact might be the reason why this TODO made it to v3.

In general, it would be great to hear some thoughts and feedback to get a brief understanding if we're on the right way. Again, the changes might look clumsy and incomplete (no unit tests for the proposed change yet): we would be happy to work on it if the proposal has any perspectives. Thanks in advance!

**If applicable**:

There is a constraint that should be taken into account before a conclusion
about overall sense can be made: the function call exists in a vacuum: there is
no continuous runtime between consecutive runs of the function. I.e. we can't
use an in-memory structure that's created once and should consider using
a persistent cache approach (temporary cache files). Dealing with a FS might
come with a significant performance penalty but in the light of the
aforementioned situation that seems decent.

This commit introduces a concept of a secondary repository index: an
extended cached index structure based on the primary repo index.
Effectively it builds an inverted index based on the data read from
index.yaml. In the current implementation it only builds inverted URL
index.

The proposed structure of a secondary file looks like:

	indexes:
		byURL:
			chart_url_1:
				name: chart_name_1
				version: chart_version_1
			chart_url_2:
				name: chart_name_2
				version: chart_version_2

Ideally, the structure should provide a nearly constant lookup time in order to
provide any sensible result. On the other hand, repository cache is a
self-contained substance which exists independently from other repositories.
Therefore it was decided to create a secondary index file for every repository.
In this case we're still dealing with a linear traversal of repositories but
relying on an assumption that an average # of repositories is smaller than an
average # of entries in them. The assumed invariant is irrelevant to the
existing algorithm as it's runtime doesn't change if the ratio changes.

In the current implementation, `scanReposForURL` performs cached index.yaml
loading on linear repo traversal. In the proposed implementation this operation
is substituted by reading index-secondary.yaml instead. We believe it has no
impact on the operation runtime complexity: the files are comparable in size
and nesting complexity.

This commit contains a new benchmark suite to test the performance of the new
implementation. For the sake of comparability we preserved the current
implementation of `scanReposForURL` but renamed it to `scanReposForURLBak`.

The results of a benchmark run:

	goos: darwin
	goarch: amd64
	pkg: helm.sh/helm/pkg/downloader
	BenchmarkScanReposForURL/with_secondary_index-4                    50000             24649 ns/op
	BenchmarkScanReposForURL/with_secondary_index-4                   100000             22762 ns/op
	BenchmarkScanReposForURL/with_secondary_index-4                    50000             23170 ns/op
	BenchmarkScanReposForURL/with_secondary_index-4                   100000             22712 ns/op
	BenchmarkScanReposForURL/with_secondary_index-4                    50000             22346 ns/op
	BenchmarkScanReposForURL/with_secondary_index-4                   100000             22658 ns/op
	BenchmarkScanReposForURL/with_secondary_index-4                   100000             23544 ns/op
	BenchmarkScanReposForURL/with_secondary_index-4                   100000             23062 ns/op
	BenchmarkScanReposForURL/with_secondary_index-4                   100000             22476 ns/op
	BenchmarkScanReposForURL/with_secondary_index-4                   100000             22621 ns/op
	BenchmarkScanReposForURL/no_secondary_index-4                         20         144900320 ns/op
	BenchmarkScanReposForURL/no_secondary_index-4                         10         178732070 ns/op
	BenchmarkScanReposForURL/no_secondary_index-4                         10         130792721 ns/op
	BenchmarkScanReposForURL/no_secondary_index-4                         10         170880342 ns/op
	BenchmarkScanReposForURL/no_secondary_index-4                         10         144174110 ns/op
	BenchmarkScanReposForURL/no_secondary_index-4                         20         147588994 ns/op
	BenchmarkScanReposForURL/no_secondary_index-4                         10         155225613 ns/op
	BenchmarkScanReposForURL/no_secondary_index-4                         10         205336598 ns/op
	BenchmarkScanReposForURL/no_secondary_index-4                         10         168166931 ns/op
	BenchmarkScanReposForURL/no_secondary_index-4                         10         152456294 ns/op

### DIFF
--- a/pkg/repo/secondary.go
+++ b/pkg/repo/secondary.go
@@ -1,0 +1,102 @@
+/*
+Copyright The Helm Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package repo
+
+import (
+	"io/ioutil"
+	"os"
+	"time"
+
+	"sigs.k8s.io/yaml"
+
+	"helm.sh/helm/v3/internal/urlutil"
+)
+
+type ChartVerEntry struct {
+	Name    string `json:"name"`
+	Version string `json:"version"`
+}
+
+type SecondaryIndexes struct {
+	ByURL map[string]ChartVerEntry `json:"byURL,omitempty"`
+}
+
+func NewSecondaryIndexes() *SecondaryIndexes {
+	return &SecondaryIndexes{}
+}
+
+type SecondaryIndexFile struct {
+	APIVersion string            `json:"apiVersion"`
+	Generated  time.Time         `json:"generated"`
+	Digest     string            `json:"digest"`
+	Indexes    *SecondaryIndexes `json:"indexes"`
+}
+
+func NewSecondaryIndexFile() *SecondaryIndexFile {
+	return &SecondaryIndexFile{
+		APIVersion: APIVersionV1,
+		Generated:  time.Now(),
+		Indexes:    NewSecondaryIndexes(),
+	}
+}
+
+//TODO: this function should ensure the loaded secondary is consistent with the
+//existing primary index by comparing digests.
+func LoadSecondaryIndexFile(path string) (*SecondaryIndexFile, error) {
+	data, err := ioutil.ReadFile(path)
+	if err != nil {
+		return nil, err
+	}
+	s := &SecondaryIndexFile{}
+	if err := yaml.Unmarshal(data, s); err != nil {
+		return nil, err
+	}
+	return s, nil
+}
+
+func (s *SecondaryIndexFile) WriteFile(dest string, mode os.FileMode) error {
+	b, err := yaml.Marshal(s)
+	if err != nil {
+		return err
+	}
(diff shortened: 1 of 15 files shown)


pkg/repo/secondary.go
```diff
@@ -0,0 +1,102 @@
...
+package repo
+
+import (
+	"io/ioutil"
+	"os"
+	"time"
+
+	"sigs.k8s.io/yaml"
+
+	"helm.sh/helm/internal/urlutil"
```
mattfarina: Helm moved to Go modules. That means the major version needs to be in the import path. This applies when something has a major version of 2 or higher. So, the helm import paths need to have a /v3 in them. For example,

`"helm.sh/helm/internal/urlutil"` -> `"helm.sh/helm/v3/internal/urlutil"`

icanhazbroccoli: @mattfarina thanks a lot for reviewing this PR! Fixed by now.

bacongobbler:

PR #6408, opened 2019-09-11
feat-v3(cli): Dynamic completion for global flags

marckhouzam:

New dynamic completion of global flags:
    --kube-context - where the available contexts are listed
    --namespace    - where the namespaces of the cluster are listed

For example:
```
   helm --namespace <TAB> # will provide the list of available namespaces on the cluster
```

Inspired greatly from kubectl code.

This PR, to work properly, is dependant and therefore built on top of #6389.

For this new completion to work, the user must have `kubectl` available.

The corresponding acceptance tests are posted in https://github.com/helm/acceptance-testing/pull/38

### DIFF
--- a/cmd/helm/completion.go
+++ b/cmd/helm/completion.go
@@ -126,13 +126,6 @@ __helm_compgen() {
 __helm_compopt() {
 	true # don't do anything. Not supported by bashcompinit in zsh
 }
-__helm_declare() {
-	if [ "$1" == "-F" ]; then
-		whence -w "$@"
-	else
-		builtin declare "$@"
-	fi
-}
 __helm_ltrim_colon_completions()
 {
 	if [[ "$1" == *:* && "$COMP_WORDBREAKS" == *:* ]]; then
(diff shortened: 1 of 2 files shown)


mattfarina:
This is pretty slick. I tested and it works. But, it adds a dependency on kubectl being installed. This is something we should talk about as maintainers. I added this to the helm dev call agenda so we can talk about it.

marckhouzam:
@mattfarina 
> But, it adds a dependency on kubectl being installed.

Yes and no.  If `kubectl` is not available, the new dynamic completion will silently fail and fallback to how things are handled today (no special completion for e.g., --kube-context), so no harm done. 

But I fully understand the maintainers should discuss if i's ok that not having `kubectl` would affect `helm`'s behaviour. Thanks for planning the discussion.

mattfarina (approved):
We talked about this in the developer call and as long as this is a soft additive dependency than kubectl is fine. I tested that it works with and without kubectl being present which was the requirement.

PR #6429, opened 2019-09-14
Updated README to link to install from script instructions.

### LINKED ISSUE #6423: No documentation for `install from script` on main README.
**FWIW**
Output of `helm version`: 2.14.2
Output of `kubectl version`: 1.14

It's not intuitive as to how to install Helm on the GCP console from the main `README` file. Seems that all other methods are documented and it took some googling to solve the problem.

Can I add a simple link to the docs folder that has the info required?

ConnorBrereton:
Closes #6423 

Needed because the documentation to install on a fresh instance that does have **Helm** installed by default is really well hidden. For example, if you wanted to install **Helm** on a new image somewhere it takes a bit of Googling to figure out how to do so. I just think this installation should be clearer on the main documentation page rather than hidden deep in a doc file.

I found this issue when I was taking @StephenGrider's course on Kubernetes. It was not easy to figure out how to get Helm installed. There are a lot of students that take this course and come to this repository so I think this would be **REALLY** helpful.

**If applicable**:
✅this PR contains documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,8 @@ If you want to use a package manager:
 
 If you want to install helm from script:
 
-- [Script](https://github.com/helm/helm/blob/master/docs/install.md#from-script) instructions to install from script on a new instance without Helm. For example, you want to install on a container, cluster, etc. Note: GCP console comes with `Helm` installed by default.
+- [Script](https://helm.sh/docs/using_helm/#from-script)
+instructions to install from script on a new instance without Helm. For example, you want to install on a container, cluster, etc. Note: GCP console comes with `helm` installed by default.
 
 To rapidly get Helm up and running, start with the [Quick Start Guide](https://docs.helm.sh/using_helm/#quickstart-guide).
 
--- a/README.md
+++ b/README.md
@@ -46,6 +46,10 @@ If you want to use a package manager:
 - [Snap](https://snapcraft.io/) users can use `sudo snap install helm
   --classic`.
 
+If you want to install helm from script:
+
+- [Script](https://github.com/helm/helm/blob/master/docs/install.md#from-script) instructions to install from script on a new instance without Helm. For example, you want to install on a container, cluster, etc. Note: GCP console comes with `Helm` installed by default.
+
 To rapidly get Helm up and running, start with the [Quick Start Guide](https://docs.helm.sh/using_helm/#quickstart-guide).
 
 See the [installation guide](https://docs.helm.sh/using_helm/#installing-helm) for more options,


karuppiah7890:
Thanks for the PR. I just have a doubt, what does it mean to install Helm in GCP Console? Like, I thought Console refers to the GCP Web UI and the terminal may be I guess?

ConnorBrereton:
Thanks for the feedback. Here is a link to the GCP console I am talking about. I guess `GCP shell` may be the more appropriate description. 

https://imgur.com/a/zWLL1W4

karuppiah7890:
Ah yes, the cloud shell. Looking at the image, I'm not sure where the shell runs, but looks like the helm connected to a cluster and contacted a tiller in the cluster and got the tiller version and showed the output. May be it's something more than just mentioning about GCP Shell. Not sure if it's needed now. Also, if we start putting notes like this, specific to cloud providers, we'll have to put notes for Azure, Digital Ocean etc etc. Considering we have given them a way to install from script too, I think that's enough 😄 It's up to the user to figure out about what their cloud provider provides using the cloud provider's docs. Tomorrow GCP may not install helm by default, then this doc would become outdated. 

What do you think? 😄

ConnorBrereton:
My additions to the README are *not* specific to a cloud provider. I just put in documentation to show people how to install Helm on a fresh instance, completely agnostic to whatever provider they're working with. 

Users shouldn't have to dig through a docs folder that seems designated for high granularity documentation to find a solution to something that is trivial in scope but an action that is very often needed.

If you want me to remove the mentioning of GCP I can/will.

ConnorBrereton:
Also I'm not sure why ***circleci*** is failing since I didn't touch the code.

karuppiah7890:
I mean the note mentioning GCP is not necessary. May be the core maintainers can tell better. Other than that LGTM 👍 

And yes, circleci shouldn't be failing

cc @thomastaylor312

ConnorBrereton:
Agreed, I'll remove the GCP specification

README.md
```diff
@@ -46,6 +46,10 @@ If you want to use a package manager:
 - [Snap](https://snapcraft.io/) users can use `sudo snap install helm
   --classic`.
 
+If you want to install helm from script:
```
thomastaylor312: I would just combine the item in the bullet point up to here and say

```
If you want to install the Helm using a script, follow [the instructions](link)...
```

Also, make sure to capitalize Helm here unless you mean the binary, in which case write it as `helm`. Same goes for the use of `Helm` below. It likely should be `helm`

README.md
```diff
@@ -46,6 +46,10 @@ If you want to use a package manager:
...
+
+- [Script](https://github.com/helm/helm/blob/master/docs/install.md#from-script) instructions to install from script on a new instance without Helm. For example, you want to install on a container, cluster, etc. Note: GCP console comes with `Helm` installed by default.
```
thomastaylor312: To make it more generic, I would drop the specific mention to GCP and say something like

> Note: Many development environments available in public clouds may already have `helm` installed. Make sure to check before installing

thomastaylor312 (changes requested):
Thanks for adding this in here! I think it will be useful for first time users. Just a few clarifications needed and it should be good to go

thomastaylor312:
@ConnorBrereton the circle CI fails are almost always due to resource constraints. We'll kick off another build if it get stuck again

ConnorBrereton:
Sounds good to me. Can we get this commit finalized?

README.md
bacongobbler: can we reference the official docs instead, which is built from the docs folder? https://helm.sh/docs/using_helm/#from-script

bacongobbler: Also, under "GCP console comes with `Helm` installed", `Helm` should be lowercase. Since it's wrapped in backticks, it implies it's referencing the name used in the console which is lowercase.

bacongobbler (changes requested):

README.md
bacongobbler: Marking as unresolved as this has not been addressed.

bacongobbler:
ping.

ConnorBrereton:
Will finish this off today.

bridgetkromhout:
Thanks, @ConnorBrereton! It will be good to get this in.

ConnorBrereton:
Ready to merge?

bacongobbler:
@ConnorBrereton see https://github.com/helm/helm/pull/6429#discussion_r327670224 for additional feedback missed.

Also looks like you need to fix up the commits such that they comply with the DCO signoff. Instructions on how to fix that up are provided here: https://github.com/helm/helm/pull/6429/checks?check_run_id=248466236

ConnorBrereton:
I made appropriate changes. Looks like it's still having the same issue regarding `make build`

README.md
```diff
@@ -46,6 +46,12 @@ If you want to use a package manager:
 - [Snap](https://snapcraft.io/) users can use `sudo snap install helm
   --classic`.
 
+If you want to install helm from script:
+
+- [Script]
+(https://helm.sh/docs/using_helm/#from-script)
+instructions to install from script on a new instance without Helm. For example, you want to install on a container, cluster, etc. Note: GCP console comes with `Helm` installed by default.
```
bacongobbler:
```suggestion
instructions to install from script on a new instance without Helm. For example, you want to install on a container, cluster, etc. Note: GCP console comes with `helm` installed by default.
```

bacongobbler (changes requested):

ConnorBrereton:
Woo hoo!

README.md
```diff
@@ -46,6 +46,12 @@ If you want to use a package manager:
 - [Snap](https://snapcraft.io/) users can use `sudo snap install helm
   --classic`.
 
+If you want to install helm from script:
```
hickeyma: helm --> Helm

README.md
```diff
@@ -46,6 +46,12 @@ If you want to use a package manager:
...
+
+- [Script]
+(https://helm.sh/docs/using_helm/#from-script)
```
hickeyma: Need to have [Script](https://helm.sh/docs/using_helm/#from-script) on same line beside each other without any space for the link to work as expected.

hickeyma (changes requested):
Nearly there @ConnorBrereton. Few small nits to fix up.

bacongobbler:
I adopted this PR and moved it forward in #6638, [removing the note about GCP](https://github.com/helm/helm/pull/6429#discussion_r325811069) (vendor information belongs in the [Kubernetes Distribution Guide](https://helm.sh/docs/kubernetes_distros/#kubernetes-distribution-guide)) and cleaned up some wording/grammar. Thanks @ConnorBrereton!

PR #6444, opened 2019-09-17
fix(v3): `helm template` output should include hooks by default

### LINKED ISSUE #6443: helm 3.0.0-beta.3: `helm template` missing hook resources
Output of `helm version`: `3.0.0-beta.3`

While we were testing helm 3.0.0-beta.3 in combination with `helm-diff`, @jonasrutishauser found that `helm template` in helm 3 misses hook outputs. That is, `helm template` misses hook resources and `helm template --no-hooks true/false` doesn't change the behavior.

Context: https://github.com/databus23/helm-diff/pull/151

To be extra clear, `helm template` in helm 2 had no `--no-hooks` and hard-coded to output hook-related k8s manifests. In helm 3, we have `--no-hooks` so to (1)to make it backward-compatible I'd like to `helm template` to include hooks in the output and (2)to make it consistent with other helm 3 commands like `install` and `upgrade`, I'd like `helm template --no-hooks` to turn output to include no hooks.

mumoshu:
This fixes `helm template [--no-hooks]` to work like proposed in #6443

Manually tested with running `helm template` against `stable/mysql` chart with helm 2, helm 3.0.0-beta.3, and helm 3 after this fix.

The manual test report follows. It assumes `helmv3` is 3.0.0-beta.3 and `helmv3-nohooksfix` is the binary build from this PR.

```
$ helmv3 fetch stable/mysql

$ helmv3 template mysql-1.3.1.tgz > helm-template-mysql-helm-3.yaml

$ helmv3-nohooksfix template mysql-1.3.1.tgz > helm-template-mysql-helm-3-with-fix.yaml

$ helmv3-nohooksfix template --no-hooks mysql-1.3.1.tgz > helm-template-mysql-helm-3-with-fix-nohooks-enabled.yaml
```

The example below shows that this fix changes `helm template` to output hooks by default:

```
$ diff --unified helm-template-mysql-helm-3{,-with-fix}.yaml
--- helm-template-mysql-helm-3.yaml	2019-09-17 22:21:38.000000000 +0900
+++ helm-template-mysql-helm-3-with-fix.yaml	2019-09-17 22:21:53.000000000 +0900
@@ -13,10 +13,10 @@
 type: Opaque
 data:

-  mysql-root-password: "VGtybWh5N3JnWA=="
+  mysql-root-password: "aGpHN2VEbnhvVA=="

-  mysql-password: "OTNQSXdNVURBYw=="
+  mysql-password: "UmpwQkVuMHpoQQ=="
 ---
 # Source: mysql/templates/tests/test-configmap.yaml
 apiVersion: v1
@@ -167,3 +167,48 @@
           claimName: RELEASE-NAME-mysql
       # - name: extras
       #   emptyDir: {}
+---
+# Source: mysql/templates/tests/test.yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: RELEASE-NAME-mysql-test
+  namespace: default
+  labels:
+    app: RELEASE-NAME-mysql
+    chart: "mysql-1.3.1"
+    heritage: "Helm"
+    release: "RELEASE-NAME"
+  annotations:
+    "helm.sh/hook": test-success
+spec:
+  initContainers:
+    - name: test-framework
+      image: "dduportal/bats:0.4.0"
+      command:
+      - "bash"
+      - "-c"
+      - |
+        set -ex
+        # copy bats to tools dir
+        cp -R /usr/local/libexec/ /tools/bats/
+      volumeMounts:
+      - mountPath: /tools
+        name: tools
+  containers:
+    - name: RELEASE-NAME-test
+      image: "mysql:5.7.14"
+      command: ["/tools/bats/bats", "-t", "/tests/run.sh"]
+      volumeMounts:
+      - mountPath: /tests
+        name: tests
+        readOnly: true
+      - mountPath: /tools
+        name: tools
+  volumes:
+  - name: tests
+    configMap:
+      name: RELEASE-NAME-mysql-test
+  - name: tools
+    emptyDir: {}
+  restartPolicy: Never
```

The example below shows that `helm template --no-hooks` can be used for excluding hooks:

```
$ diff --unified helm-template-mysql-helm-3{,-with-fix-nohooks-enabled}.yaml
--- helm-template-mysql-helm-3.yaml	2019-09-17 22:21:38.000000000 +0900
+++ helm-template-mysql-helm-3-with-fix-nohooks-enabled.yaml	2019-09-17 22:22:03.000000000 +0900
@@ -13,10 +13,10 @@
 type: Opaque
 data:

-  mysql-root-password: "VGtybWh5N3JnWA=="
+  mysql-root-password: "Zk1LYUd6OWgzaQ=="

-  mysql-password: "OTNQSXdNVURBYw=="
+  mysql-password: "OTZPZU9hdlFORg=="
 ---
 # Source: mysql/templates/tests/test-configmap.yaml
 apiVersion: v1
```

Fixes #6443

### DIFF
--- a/cmd/helm/template.go
+++ b/cmd/helm/template.go
@@ -56,6 +56,12 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 				return err
 			}
 			fmt.Fprintln(out, strings.TrimSpace(rel.Manifest))
+			if !client.DisableHooks {
+				for _, m := range rel.Hooks {
+					fmt.Fprintf(out, "---\n# Source: %s\n%s\n", m.Path, m.Manifest)
+				}
+			}
+
 			return nil
 		},
 	}


pkg/action/install.go
```diff
@@ -456,6 +457,20 @@ func (c *Configuration) renderResources(ch *chart.Chart, values chartutil.Values
 		}
 	}
 
+	if outputHooks {
```
thomastaylor312: While I totally see why this is here (install and template use the same code), this modification to the function feels like it doesn’t belong here and should be a `helm template` only implementation. I think someone could accidentally inject the hooks into a manifest they are installing (if using Helm as a library). So I think this code should live in `cmd/helm/template.go` and then we can export the WriteToFile as a helper. Any thoughts on this idea?

thomastaylor312 (commented):
Code works great! One point of discussion here that I'd love your thoughts on

pkg/action/install.go
mumoshu: Hey! Thanks for your feedback.

> So I think this code should live in cmd/helm/template.go

Yeah that makes sense! However, I'm not entirely sure how it can be implemented without refactoring/duplicating existing code.

What we need to achieve is that do it without modifying `func (i *Install) Run`, `renderResources`, and `runInstall`, right?

To me that basically mean we need to implement something like `runTemplate` which calls the existing `renderResources` and this block of code specific to `template`(for rendering hooks). Most of `runTemplate` code will be duplicated from `runInstall`.

Is that okay? Or would you like to me to factor common parts of `runInstall` and `runTemplate`?

thomastaylor312: You are totally right. What if we just did this in `cmd/helm/template.go` with 

```go
// Preexisting line for reference
fmt.Fprintln(out, strings.TrimSpace(rel.Manifest))
if !client.DisableHooks {
	for _, h := range rel.Hooks {
		fmt.Fprintln(out, h.Manifest)
	}
}
```

It would avoid adding an extra flag and no need to refactor the underlying client

mumoshu: That sounds great! I now realize that the release object returned from `runInstall` has what I needed. I'll try and report back soon.

mumoshu:
@thomastaylor312 Pushed 5c18e1b. It's now a 6 lines change. I re-verified it works as expected by running `helm template` with and without `--no-hooks`.

Thanks a lot for your great feedback!

thomastaylor312 (approved):
LGTM. Thanks for the fix @mumoshu!

PR #6442, opened 2019-09-17
Refactor switch statement in pkg/kube/wait.go #6440

u5surf:
**What this PR does / why we need it**:
 #6440 
@fengxsong solved the type switch build failing problems in #6449 
So, I'll deal with the same issue having another cases.

https://github.com/helm/helm/blob/9a592e537970f0140cd295111c74a28ee7cc855d/pkg/kube/wait.go#L163-L170

PR #6449, opened 2019-09-18
Refactor switch statement in 'pkg/kube/wait.go'  closes #6440

### LINKED ISSUE #6440: [Helm 3] I have a question about 'switch ... case ' syntax
- https://github.com/helm/helm/blob/5cb923eecbe80d1ad76399aee234717c11931d9a/pkg/kube/wait.go#L67
- https://github.com/helm/helm/blob/5cb923eecbe80d1ad76399aee234717c11931d9a/pkg/kube/wait.go#L80
- https://github.com/helm/helm/blob/5cb923eecbe80d1ad76399aee234717c11931d9a/pkg/kube/wait.go#L93
- https://github.com/helm/helm/blob/5cb923eecbe80d1ad76399aee234717c11931d9a/pkg/kube/wait.go#L106

Wondering why not use syntax as follow

```go
...
            switch value := AsVersioned(v).(type) {
            case *corev1.Pod:
              ...
            case *appsv1.Deployment, *appsv1beta1.Deployment, *appsv1beta2.Deployment, *extensionsv1beta1.Deployment:
              ...
              ...
            }
```

fengxsong:
FIX #6440 #6442

### DIFF
--- a/pkg/kube/wait.go
+++ b/pkg/kube/wait.go
@@ -121,14 +121,8 @@ func (w *waiter) waitForResources(created ResourceList) error {
 					return false, nil
 				}
 
-			case *corev1.ReplicationController:
-				ok, err = w.podsReadyForObject(value.Namespace, value)
-			case *extensionsv1beta1.ReplicaSet:
-				ok, err = w.podsReadyForObject(value.Namespace, value)
-			case *appsv1beta2.ReplicaSet:
-				ok, err = w.podsReadyForObject(value.Namespace, value)
-			case *appsv1.ReplicaSet:
-				ok, err = w.podsReadyForObject(value.Namespace, value)
+			case *corev1.ReplicationController, *extensionsv1beta1.ReplicaSet, *appsv1beta2.ReplicaSet, *appsv1.ReplicaSet:
+				ok, err = w.podsReadyForObject(v.Namespace, value)
 			}
 			if !ok || err != nil {
 				return false, err
--- a/pkg/kube/wait.go
+++ b/pkg/kube/wait.go
@@ -65,7 +65,7 @@ func (w *waiter) waitForResources(created ResourceList) error {
 					return false, err
 				}
 			case *appsv1.Deployment, *appsv1beta1.Deployment, *appsv1beta2.Deployment, *extensionsv1beta1.Deployment:
-				currentDeployment, err := w.c.AppsV1().Deployments(value.Namespace).Get(value.Name, metav1.GetOptions{})
+				currentDeployment, err := w.c.AppsV1().Deployments(v.Namespace).Get(v.Name, metav1.GetOptions{})
 				if err != nil {
 					return false, err
 				}
--- a/pkg/kube/wait.go
+++ b/pkg/kube/wait.go
@@ -64,46 +64,7 @@ func (w *waiter) waitForResources(created ResourceList) error {
 				if err != nil || !w.isPodReady(pod) {
 					return false, err
 				}
-			case *appsv1.Deployment:
-				currentDeployment, err := w.c.AppsV1().Deployments(value.Namespace).Get(value.Name, metav1.GetOptions{})
-				if err != nil {
-					return false, err
-				}
-				// Find RS associated with deployment
-				newReplicaSet, err := deploymentutil.GetNewReplicaSet(currentDeployment, w.c.AppsV1())
-				if err != nil || newReplicaSet == nil {
-					return false, err
-				}
-				if !w.deploymentReady(newReplicaSet, currentDeployment) {
-					return false, nil
-				}
-			case *appsv1beta1.Deployment:
-				currentDeployment, err := w.c.AppsV1().Deployments(value.Namespace).Get(value.Name, metav1.GetOptions{})
-				if err != nil {
-					return false, err
-				}
-				// Find RS associated with deployment
-				newReplicaSet, err := deploymentutil.GetNewReplicaSet(currentDeployment, w.c.AppsV1())
-				if err != nil || newReplicaSet == nil {
-					return false, err
-				}
-				if !w.deploymentReady(newReplicaSet, currentDeployment) {
-					return false, nil
-				}
-			case *appsv1beta2.Deployment:
-				currentDeployment, err := w.c.AppsV1().Deployments(value.Namespace).Get(value.Name, metav1.GetOptions{})
-				if err != nil {
-					return false, err
-				}
-				// Find RS associated with deployment
-				newReplicaSet, err := deploymentutil.GetNewReplicaSet(currentDeployment, w.c.AppsV1())
-				if err != nil || newReplicaSet == nil {
-					return false, err
-				}
-				if !w.deploymentReady(newReplicaSet, currentDeployment) {
-					return false, nil
-				}
-			case *extensionsv1beta1.Deployment:
+			case *appsv1.Deployment, *appsv1beta1.Deployment, *appsv1beta2.Deployment, *extensionsv1beta1.Deployment:
 				currentDeployment, err := w.c.AppsV1().Deployments(value.Namespace).Get(value.Name, metav1.GetOptions{})
 				if err != nil {
 					return false, err


hickeyma:
@fengxsong and @u5surf you have submitted PRs in v2 (#6442) and v3 (#6449 ) branches, to solve issue (#6440 ). However, you both added the same fix which generates the same build error. Can you fix the build errors please?

fengxsong:
@hickeyma I have submitted a small patch.

hickeyma (changes requested):
Thanks @fengxsong for updating this. One other thins. Can you also refactor the duplication in this switch statement: https://github.com/helm/helm/blob/dev-v3/pkg/kube/wait.go#L343 ?

fengxsong:
@hickeyma Can you see if this method is feasible?

```go
import "k8s.io/apimachinery/pkg/apis/meta/unstructured"

func SelectorsForObject(object runtime.Object) (selector labels.Selector, err error) {
	unstructuredObj, err := runtime.DefaultUnstructuredConverter.ToUnstructured(object)
	if err != nil {
		return nil, errors.Wrapf(err, "invalid type %T", object)
	}
	strMap, found, _ := unstructured.NestedStringMap(unstructuredObj, "spec", "selector")
	if !found || len(strMap) == 0 {
		return nil, errors.New("Object is defined without a selector")
	}
	selector = labels.SelectorFromSet(strMap)
	return
}
```

hickeyma:
@fengxsong What are you trying to achieve with the changes in https://github.com/helm/helm/pull/6449#issuecomment-533374622 ?

fengxsong:
> @fengxsong What are you trying to achieve with the changes in [#6449 (comment)](https://github.com/helm/helm/pull/6449#issuecomment-533374622) ?

Because I have no clear way to refactor it for now. 😂

hickeyma (changes requested):
@fengxsong Changes are good. There is just another change you can make in https://github.com/helm/helm/blob/dev-v3/pkg/kube/wait.go#L344, as it has repetition on the case statements with this line:
`selector, err = metav1.LabelSelectorAsSelector(t.Spec.Selector)`

pkg/kube/wait.go
```diff
@@ -337,49 +292,33 @@ func getPods(client kubernetes.Interface, namespace, selector string) ([]corev1.
...
 
-	return selector, errors.Wrap(err, "invalid label selector")
+	labelset, ok, err := unstructured.NestedStringMap(obj, nestedFields...)
+	if err != nil {
+		return nil, errors.Wrap(err, "invalid label selector")
+	}
+	if !ok || len(labelset) == 0 {
+		return nil, fmt.Errorf("invalid object: %s is defined without a selector", kind)
+	}
+	return labels.SelectorFromSet(labelset), nil
```
bacongobbler: May I ask why the object is being converted to an unstructured object, then trying to fetch its metadata with internal fields? This seems very dangerous, as different versions of a particular type can use different fields for label selectors. This new code does not differentiate between the Group, Version and Kind (GVK), whereas the previous (though highly redundant) code above did.

I'd suggest leaving reverting this and leaving it as-is for the time being. This does not seem to gain us any long term benefit (in fact, it could harm us in the future).

bacongobbler (changes requested):

pkg/kube/wait.go
bacongobbler: A very simple use case where this code could fail:

1. I registered a new CRD with GVK `myappv1.Service`
2. This new `Service` type acts like a completely different object with a completely different API than a `corev1.Service`

In this case, the old code would've identified that this object is not part of the `corev1.Service` GVK and should be ignored. In this new code, it would not, and treat it similar to the `corev1.Service` object, causing indeterminate behaviour.

fengxsong: I agree with you, just leave it as it is.

pkg/kube/wait.go
```diff
@@ -64,47 +64,8 @@ func (w *waiter) waitForResources(created ResourceList) error {
...
-				if err != nil || newReplicaSet == nil {
-					return false, err
-				}
-				if !w.deploymentReady(newReplicaSet, currentDeployment) {
-					return false, nil
-				}
-			case *extensionsv1beta1.Deployment:
-				currentDeployment, err := w.c.AppsV1().Deployments(value.Namespace).Get(value.Name, metav1.GetOptions{})
+			case *appsv1.Deployment, *appsv1beta1.Deployment, *appsv1beta2.Deployment, *extensionsv1beta1.Deployment:
+				currentDeployment, err := w.c.AppsV1().Deployments(v.Namespace).Get(v.Name, metav1.GetOptions{})
```
hickeyma: Why have you changed from `value` to `v`? You are losing the type here.

hickeyma (changes requested):

pkg/kube/wait.go
fengxsong: we don't need the actual type here.

hickeyma: why was it changed?

bacongobbler: Agreeing with @hickeyma here. There may be unintended side-effects where the Kubernetes API treats unversioned types differently than versioned types.

Please revert back to the previous behaviour of performing a lookup using `value` instead of `v`. The goal of a refactor is to retain previous behaviour wherever possible. Switching input types to a function can cause new and unintended behaviour.

If you want to make this change, please submit a *new* PR with the change, along with supporting tests. Thanks!

bacongobbler (changes requested):

pkg/kube/wait.go
```diff
@@ -160,14 +121,8 @@ func (w *waiter) waitForResources(created ResourceList) error {
...
-			case *corev1.ReplicationController:
-				ok, err = w.podsReadyForObject(value.Namespace, value)
-			case *extensionsv1beta1.ReplicaSet:
-				ok, err = w.podsReadyForObject(value.Namespace, value)
-			case *appsv1beta2.ReplicaSet:
-				ok, err = w.podsReadyForObject(value.Namespace, value)
-			case *appsv1.ReplicaSet:
-				ok, err = w.podsReadyForObject(value.Namespace, value)
+			case *corev1.ReplicationController, *extensionsv1beta1.ReplicaSet, *appsv1beta2.ReplicaSet, *appsv1.ReplicaSet:
+				ok, err = w.podsReadyForObject(v.Namespace, value)
```
bacongobbler: same here.

bacongobbler (changes requested):

hickeyma:
@fengxsong Thanks for your efforts so far. To try and get this PR merged:
- Rebase and resolve conflict
- Refactor the duplication in the switch statements and avoid changing code login in this PR. You can do this in another PR as suggested by @bacongobbler.

pkg/kube/wait.go
fengxsong: I don't quite understand, I made these changes based on another part of the code, such as [line 152](https://github.com/helm/helm/blob/cc450563835088945aa79dbfa0bce0bc031fb390/pkg/kube/wait.go#L152) and [line 171](https://github.com/helm/helm/blob/cc450563835088945aa79dbfa0bce0bc031fb390/pkg/kube/wait.go#L171)

bacongobbler: Referring to this discussion here: https://github.com/helm/helm/pull/6449#discussion_r331755560

This code also uses the unversioned type `v` instead of the versioned type `value`. Please change this to use the versioned type `value`

bacongobbler:
So as it turns out, multiple case statements on the same line triggers a case where Go will use its interface rather than the actual type itself. In other words, `value` changes from an `*appsv1.Deployment` to a `runtime.Object`.

I found this out from [Go's documentation on type cases](https://golang.org/ref/spec#Type_switches):

> In clauses with a case listing exactly one type, the variable has that type; otherwise, the variable has the type of the expression in the TypeSwitchGuard. 

> Given an expression x of type interface{}, the following type switch:

```
switch i := x.(type) {
case nil:
	printString("x is nil")                // type of i is type of x (interface{})
case int:
	printInt(i)                            // type of i is int
case float64:
	printFloat64(i)                        // type of i is float64
case func(int) float64:
	printFunction(i)                       // type of i is func(int) float64
case bool, string:
	printString("type is bool or string")  // type of i is type of x (interface{})
default:
	printString("don't know the type")     // type of i is type of x (interface{})
}
```

> could be rewritten:

```
v := x  // x is evaluated exactly once
if v == nil {
	i := v                                 // type of i is type of x (interface{})
	printString("x is nil")
} else if i, isInt := v.(int); isInt {
	printInt(i)                            // type of i is int
} else if i, isFloat64 := v.(float64); isFloat64 {
	printFloat64(i)                        // type of i is float64
} else if i, isFunc := v.(func(int) float64); isFunc {
	printFunction(i)                       // type of i is func(int) float64
} else {
	_, isBool := v.(bool)
	_, isString := v.(string)
	if isBool || isString {
		i := v                         // type of i is type of x (interface{})
		printString("type is bool or string")
	} else {
		i := v                         // type of i is type of x (interface{})
		printString("don't know the type")
	}
}
```

I had a deeper look at how Helm and Kubernetes consume the object being watched, and it looks safe to use the unversioned type. In cases where the versioned type needs to be used, a [safe `.Get` conversion is called](https://github.com/helm/helm/blob/62ed7b3764bae112a6a8838975a0ee777a83ec36/pkg/kube/wait.go#L206), so it doesn't matter whether we pass along an unversioned or a versioned type to these functions.

Because of this, I rebased this PR and created #6608. It takes the work @fengxsong put in here and cleans up the code a little more to no longer rely on `value`. I'm going to close this PR in favour of that one and get that one merged in.

Thanks!

PR #6457, opened 2019-09-18
Updating to newer versions of dependencies

mattfarina:
* This was initiated by new versions from Masterminds
* Linting was set to check for the same chart version as the other
  checks of this were. >0.0.0-0. The -0 is important to pull in
  prereleases. The latest version of semver Validate now finds
  errors here properly where some were missed before
* 0.0.0 is a valid semantic version. Instead of expecting it to be
  invalid in the tests to check validation now using an invalid
  semantic version

**If applicable**:


hickeyma (approved):
LGTM, thanks @mattfarina

PR #6470, opened 2019-09-20
v3: fix(completion): Redirect stderr must be before pipes

marckhouzam:

Redirecting stderr from Helm must be done before piping.
Without this, if there are no repos defined completion will show the error message.
```
> h3 repo remove <TAB>
h3 repo remove ......Error: no repositories to show
```

Tests added for this bug in https://github.com/helm/acceptance-testing/pull/44


mattfarina (approved):
I tested, was able to validate the error, and the fix. This makes sense.

PR #4871, opened 2018-11-01
Fix `no RESOURCE with the name NAME found`

distorhead:
This is the fix for only one particular, but important case.

The case when a new resource has been added to the chart and there is an error in the chart, which leads to release failure. In this case after first failed release upgrade new resource will be created in the cluster. On the next release upgrade there will be the error: `no RESOURCE with the name NAME found` for this newly created resource from the previous release upgrade.

The root of this problem is in the side effect of the first release process. Release invariant says: if resouce exists in the kubernetes cluster, then it should exist in the release storage. But this invariant has been broken by helm itself -- because helm created new resources as side effect and not adopted them into release storage.

To maintain release invariant for such case during release upgrade operation all newly *successfully* created resources will be deleted in the case of an error in the subsequent resources update.

PR #6480, opened 2019-09-23
feat(*): Ports `--cleanup-on-fail` to v3

thomastaylor312:
This ports the functionality of cleanup on fail to v3 as introduced in #4871. This has been tested manually and would be a good candidate for a new acceptance test.

### DIFF
--- a/pkg/action/rollback.go
+++ b/pkg/action/rollback.go
@@ -19,6 +19,7 @@ package action
 import (
 	"bytes"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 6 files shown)


pkg/action/rollback.go
```diff
@@ -165,6 +168,18 @@ func (r *Rollback) performRollback(currentRelease, targetRelease *release.Releas
...
 		r.cfg.recordRelease(targetRelease)
+		if r.CleanupOnFail {
+			r.cfg.Log("Cleanup on fail set, cleaning up %d resources", len(results.Created))
+			_, errs := r.cfg.KubeClient.Delete(results.Created)
+			if errs != nil {
+				var errorList []string
+				for _, e := range errs {
+					errorList = append(errorList, e.Error())
+				}
+				return targetRelease, errors.Wrapf(fmt.Errorf("unable to cleanup resources: %s", strings.Join(errorList, ", ")), "an error occurred while cleaning up resources. original rollback error: %s", err)
```
thomastaylor312: If we don't want this to return an error, we can just log it and continue on

pkg/kube/client.go
```diff
@@ -180,18 +183,17 @@ func (c *Client) Update(original, target ResourceList, force bool) (*Result, err
...
-		return nil, err
+		return res, err
 	case len(updateErrors) != 0:
-		return nil, errors.Errorf(strings.Join(updateErrors, " && "))
+		return res, errors.Errorf(strings.Join(updateErrors, " && "))
 	}
 
 	for _, info := range original.Difference(target) {
 		c.Log("Deleting %q in %s...", info.Name, info.Namespace)
+		res.Deleted = append(res.Deleted, info)
```
bacongobbler: reading the comment in line 193, how come we are appending here before the resource was deleted?

thomastaylor312: This was to match behavior with `Created` and `Updated`. Basically the `Result` should contain everything it _tried_ to create, update, or delete

bacongobbler (approved):

PR #6486, opened 2019-09-24
fix(action): invalidate discovery client cache on startup

### LINKED ISSUE #6452: Helm 3 beta3: go-client caching issue when installing chart resource which is based on a CRD
Output of `helm version`:
```
```

Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.3", GitCommit:"2d3c76f9091b6bec110a5e63777c332469e0cba2", GitTreeState:"clean", BuildDate:"2019-08-19T12:36:28Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.2", GitCommit:"f6278300bebbb750328ac16ee6dd3aa7d3549568", GitTreeState:"clean", BuildDate:"2019-08-05T09:15:22Z", GoVersion:"go1.12.5", Compiler:"gc", Platform:"linux/amd64"}
```

**Issue**
When installing a Helm chart which includes a resource from a previously registered CRD (not included in helm chart), Helm throws an error like the following one.

```
Error: apiVersion "test.com/v1alpha1" in reproducer/templates/demo.yaml is not available
```

This problem was already handled in these issues & PR's:
https://github.com/helm/helm/issues/5654
https://github.com/helm/helm/pull/6049

The actual problem still exists. The reason is, that the client-go discovery cache doesn't gets invalidated before installing the chart.

**Reproducer**

Register a CRD out of the scope of Helm.

```
kubectl apply -f - << EOF 
apiVersion: apiextensions.k8s.io/v1beta1
metadata:
  name: demos.test.com
spec:
  group: test.com
  names:
    kind: Demo
    listKind: DemoList
    plural: demos
    singular: demo
  scope: Namespaced
  subresources:
    status: {}
  version: v1alpha1
  versions:
  - name: v1alpha1
    served: true
    storage: true
EOF
```

Try to install or template (with --validate) a chart with the following resource.

```
apiVersion: test.com/v1alpha1
kind: Demo
metadata:
  name: default
spec: {}
```

This results in the error.

```
Error: apiVersion "test.com/v1alpha1" in reproducer/templates/demo.yaml is not available
```

After invalidating the client-go discovery cache it works as expected. As a workaround it's possible to invalidate the cache by executing `kubectl api-versions`, which internally invalidates the discovery cache (https://github.com/kubernetes/kubernetes/pull/48016/files), before installing the chart.

```
kubectl api-versions
...
helm template testchart --validate
---
# Source: reproducer/templates/demo.yaml
apiVersion: test.com/v1alpha1
kind: Demo
metadata:
  name: default
spec: {}
```

**Possible fix**

Currently Helm internally invalidates the discovery cache only if there are CRD's applied in the context of the Helm chart. (https://github.com/helm/helm/blob/v3.0.0-beta.3/pkg/action/install.go#L131)

A possible solution would be to always invalidate the discovery cache when installing a chart if the mode isn't `clientOnly`.

https://github.com/helm/helm/blob/v3.0.0-beta.3/pkg/action/install.go#L175

```
	if i.ClientOnly {
		// Add mock objects in here so it doesn't use Kube API server
		// NOTE(bacongobbler): used for `helm template`
		i.cfg.Capabilities = chartutil.DefaultCapabilities
		i.cfg.KubeClient = &kubefake.PrintingKubeClient{Out: ioutil.Discard}
		i.cfg.Releases = storage.Init(driver.NewMemory())
	} else {
		// Invalidate the discovery cache here
    }
```

**Related issues and pull requests**
https://github.com/helm/helm/issues/5654
https://github.com/helm/helm/pull/6049

bacongobbler:
We want to force a cache invalidation to ensure that the Capabilities object always has the latest information from the server (Kubernetes server version, available API versions, etc). `kubectl version` forces a cache invalidation every time it's invoked, so this seems like a safe change that is identical to kubectl's behaviour.

closes #6452

### DIFF
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -91,6 +91,8 @@ func (c *Configuration) getCapabilities() (*chartutil.Capabilities, error) {
 	if err != nil {
 		return nil, errors.Wrap(err, "could not get Kubernetes discovery client")
 	}
+	// force a discovery cache invalidation to always fetch the latest server version/capabilities.
+	dc.Invalidate()
 	kubeVersion, err := dc.ServerVersion()
 	if err != nil {
 		return nil, errors.Wrap(err, "could not get server version from Kubernetes")


bacongobbler:
cc'ing @adamreese to see if he knows if there's a more optimal way of selectively invalidating the cache when needed.

thomastaylor312:
@bacongobbler Based on comments for `Fresh`, I think they want to refactor that code anyway because it isn't clear on when the cache is stale.

thomastaylor312 (approved):
Given that this is the same functionality as is in kubectl, I feel like this is a completely safe change

PR #6488, opened 2019-09-24
fix(downloader): bypass index cache when repository URL defined

bacongobbler:
port of #6258 

### DIFF
--- a/internal/resolver/resolver.go
+++ b/internal/resolver/resolver.go
@@ -71,11 +71,20 @@ func (r *Resolver) Resolve(reqs []*chart.Dependency, repoNames map[string]string
 			return nil, errors.Wrapf(err, "dependency %q has an invalid version/constraint format", d.Name)
 		}
 
-		idx := filepath.Join(r.cachepath, helmpath.CacheIndexFile(repoNames[d.Name]))
+		repoName := repoNames[d.Name]
+		// if the repository was not defined, but the dependency defines a repository url, bypass the cache
+		if repoName == "" && d.Repository != "" {
+			locked[i] = &chart.Dependency{
+				Name:       d.Name,
+				Repository: d.Repository,
+				Version:    d.Version,
+			}
+			continue
+		}
 
-		repoIndex, err := repo.LoadIndexFile(idx)
+		repoIndex, err := repo.LoadIndexFile(filepath.Join(r.cachepath, helmpath.CacheIndexFile(repoName)))
 		if err != nil {
-			return nil, errors.Wrapf(err, "no cached repo found. (try 'helm repo update') %s", idx)
+			return nil, errors.Wrapf(err, "no cached repository for %s found. (try 'helm repo update')", repoName)
 		}
 
 		vs, ok := repoIndex.Entries[d.Name]
--- a/cmd/helm/dependency_update_test.go
+++ b/cmd/helm/dependency_update_test.go
@@ -117,33 +117,6 @@ func TestDependencyUpdateCmd(t *testing.T) {
 	}
 }
 
-func TestDependencyUpdateCmd_SkipRefresh(t *testing.T) {
-	defer resetEnv()()
-	defer ensure.HelmHome(t)()
-
-	srv := repotest.NewServer(helmpath.ConfigPath())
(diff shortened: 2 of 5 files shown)


internal/resolver/resolver.go
```diff
@@ -71,11 +72,22 @@ func (r *Resolver) Resolve(reqs []*chart.Dependency, repoNames map[string]string
 			return nil, errors.Wrapf(err, "dependency %q has an invalid version/constraint format", d.Name)
 		}
 
-		idx := filepath.Join(r.cachepath, helmpath.CacheIndexFile(repoNames[d.Name]))
+		repoName := repoNames[d.Name]
+		// if the repository was not defined, but the dependency defines a repository url, bypass the cache
+		if repoName == "" && d.Repository != "" {
+			fmt.Println(d.Name)
+			fmt.Println(d.Repository)
```
mattfarina: Were these print statements debugging that was accidentally left in?

mattfarina (commented):
This may be for a later refactor but, if the local copy in the cache has the same digest/sha it would have the same content. Shouldn't we use the cache? Though, if we don't know the digest that would mean we can't use the cache.

internal/resolver/resolver.go
bacongobbler: Oops, yes they were. Good catch.

bacongobbler:
Yeah, in the current architecture, in order to do a cache lookup we need to know the name of the repository. That's probably why we needed `helm repo add`... That way, `helm dep up` could do a cache lookup.

I haven't put any thought into improving upon what's already in Helm 2. That does seem like a nice enhancement, but I agree that it could be done in a later refactor. Should we move this ahead as-is or do we need to block this until that functionality is complete?

hickeyma:
@bacongobbler I noticed something with cache during migration. The [2to3](https://github.com/helm/helm-2to3) plugin does not copy over the cache. This is a conscious decision as cache should be recreated by Helm 3 rather than copied. The repository file when copied initially, will contain references to Helm v2 cache. I would have expected when `helm repo update` command is run that this cache would be regenerated for the Helm v3 configuration path.  Should it regenerate on an update?

bacongobbler:
Hmm, good point. Yes, the cache should be regenerated on an update. Is that not occurring?

hickeyma:
Sorry for delay in getting back to you @bacongobbler .

On a test system, I have done a migration of configuration and release data from Helm 2 to Helm 3. I also cleaned up the Helm 2 config, release data and Tiller deployment.

I then update the repo and  would expect the repo file to be updated to:
- Refresh any stale cache references
- Remove any local repositories that no longer exist (i.e. that were not copied over to Helm 3) 

```console
$ helm3 repo update
...Unable to get an update from the "local" chart repository (http://127.0.0.1:8879/charts):
	Get http://127.0.0.1:8879/charts/index.yaml: dial tcp 127.0.0.1:8879: connect: connection refused

$ cat ~/.config/helm/repositories.yaml 
apiVersion: v1
generated: "2019-10-02T16:45:23.006506702+01:00"
repositories:
- caFile: ""
  cache: /home/usr1/.helm/repository/cache/stable-index.yaml
  certFile: ""
  keyFile: ""
  name: stable
  password: ""
  url: https://kubernetes-charts.storage.googleapis.com
  username: ""
- caFile: ""
  cache: /home/usr1/.helm/repository/cache/local-index.yaml
  certFile: ""
  keyFile: ""
  name: local
  password: ""
  url: http://127.0.0.1:8879/charts
  username: ""
```

How do you think this should work?

bacongobbler:
I think we discussed that topic during the dev call.

@hickeyma do you have any feedback on this PR? I'd like to bring the discussion back to the PR itself. If there's further discussions about that particular field, can we start that discussion in a new ticket perhaps? I'm just not sure how the discussion topic is relevant to this PR.

hickeyma:
@bacongobbler Sure, agree. I have raised #6644 for the issue with the cache.

I will get back to reviewing this PR shortly.

hickeyma (approved):
LGTM, thanks @bacongobbler

thomastaylor312 (approved):

PR #6490, opened 2019-09-24
fix(chartutil): port over enhancements to `helm create` from Helm 2

bacongobbler:

### DIFF
--- a/cmd/helm/create.go
+++ b/cmd/helm/create.go
@@ -42,6 +42,7 @@ something like this:
 	├── values.yaml   # The default values for your templates
 	├── charts/       # Charts that this chart depends on
 	└── templates/    # The template files
+		└── tests/    # The test files
 
 'helm create' takes a path for an argument. If directories in the given path
 do not exist, Helm will attempt to create them as it goes. If the given
--- a/cmd/helm/create.go
+++ b/cmd/helm/create.go
@@ -36,13 +36,13 @@ directories used in a chart.
 For example, 'helm create foo' will create a directory structure that looks
 something like this:
 
-	foo/
-	├── .helmignore   # Contains patterns to ignore when packaging Helm charts.
-	├── Chart.yaml    # Information about your chart
-	├── values.yaml   # The default values for your templates
-	├── charts/       # Charts that this chart depends on
-	└── templates/    # The template files
-		└── tests/    # The test files
+    foo/
+    ├── .helmignore   # Contains patterns to ignore when packaging Helm charts.
+    ├── Chart.yaml    # Information about your chart
+    ├── values.yaml   # The default values for your templates
+    ├── charts/       # Charts that this chart depends on
+    └── templates/    # The template files
+        └── tests/    # The test files
 
 'helm create' takes a path for an argument. If directories in the given path
 do not exist, Helm will attempt to create them as it goes. If the given
--- a/cmd/helm/create_test.go
+++ b/cmd/helm/create_test.go
@@ -106,8 +106,9 @@ func TestCreateStarterCmd(t *testing.T) {
 		t.Errorf("Wrong API version: %q", c.Metadata.APIVersion)
 	}
 
-	if l := len(c.Templates); l != 6 {
-		t.Errorf("Expected 5 templates, got %d", l)
+	expectedNumberOfTemplates := 8
+	if l := len(c.Templates); l != expectedNumberOfTemplates {
+		t.Errorf("Expected %d templates, got %d", expectedNumberOfTemplates, l)
 	}
 
 	found := false
@@ -173,8 +174,9 @@ func TestCreateStarterAbsoluteCmd(t *testing.T) {
 		t.Errorf("Wrong API version: %q", c.Metadata.APIVersion)
 	}
 
-	if l := len(c.Templates); l != 6 {
-		t.Errorf("Expected 5 templates, got %d", l)
+	expectedNumberOfTemplates := 8
+	if l := len(c.Templates); l != expectedNumberOfTemplates {
+		t.Errorf("Expected %d templates, got %d", expectedNumberOfTemplates, l)
 	}
 
 	found := false
--- a/pkg/chartutil/create.go
+++ b/pkg/chartutil/create.go
@@ -39,6 +39,8 @@ const (
 	TemplatesDir = "templates"
 	// ChartsDir is the relative directory name for charts dependencies.
 	ChartsDir = "charts"
+	// TemplatesTestsDir is the relative directory name for tests.
+	TemplatesTestsDir = TemplatesDir + sep + "tests"
 	// IgnorefileName is the name of the Helm ignore file.
 	IgnorefileName = ".helmignore"
 	// IngressFileName is the name of the example ingress file.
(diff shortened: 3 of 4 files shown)


pkg/chartutil/create.go
```diff
@@ -47,10 +47,14 @@ const (
 	DeploymentName = TemplatesDir + sep + "deployment.yaml"
 	// ServiceName is the name of the example service file.
 	ServiceName = TemplatesDir + sep + "service.yaml"
+	// ServiceAccountName is the name of the example serviceaccount file.
+	ServiceAccountName = TemplatesDir + sep + "serviceaccount.yaml"
 	// NotesName is the name of the example NOTES.txt file.
 	NotesName = TemplatesDir + sep + "NOTES.txt"
 	// HelpersName is the name of the example NOTES.txt file.
```
hickeyma: :s/HelpersName is the name of the example NOTES.txt file./HelpersName is the name of the example NOTES.txt file./HelpersName is the name of the example helpers file.

hickeyma (changes requested):
Thanks @bacongobbler for doing the port. Some comments inline and also the following 2:
- Update this with test dir added: https://github.com/helm/helm/blob/dev-v3/cmd/helm/create.go#L39
- Add ServiceAccountName, TestConnectionName, TemplatesTestsDir to: https://github.com/helm/helm/blob/dev-v3/pkg/chartutil/create_test.go#L52

cmd/helm/create_test.go
```diff
@@ -106,8 +106,8 @@ func TestCreateStarterCmd(t *testing.T) {
 		t.Errorf("Wrong API version: %q", c.Metadata.APIVersion)
 	}
 
-	if l := len(c.Templates); l != 6 {
-		t.Errorf("Expected 5 templates, got %d", l)
+	if l := len(c.Templates); l != 8 {
```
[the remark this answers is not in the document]
bacongobbler: Good call. Yes, that is a better way to write this.

pkg/chartutil/create.go
```diff
@@ -47,10 +47,14 @@ const (
...
 	HelpersName = TemplatesDir + sep + "_helpers.tpl"
+	// TestConnectionName is the name of the example test hook.
```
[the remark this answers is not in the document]
bacongobbler: Yes, good call. That way if there are future tests we want to add in there they can re-use `TemplatesTestsDir`.

bacongobbler:
Thank you for catching all those, @hickeyma. All comments addressed!

cmd/helm/create.go
```diff
@@ -42,6 +42,7 @@ something like this:
 	├── values.yaml   # The default values for your templates
 	├── charts/       # Charts that this chart depends on
 	└── templates/    # The template files
+	└── tests/        # The test files
```
hickeyma: Change 'tests/' --> 'templates/tests'

pkg/chartutil/create_test.go
```diff
@@ -57,6 +57,7 @@ func TestCreate(t *testing.T) {
 		NotesName,
 		ServiceName,
 		TemplatesDir,
+		TemplatesTestsDir,
```
hickeyma: Can you also add: ServiceAccountName & TestConnectionName ?

hickeyma (changes requested):
Thanks for update @bacongobbler. Two small things.

pkg/chartutil/create.go
```diff
@@ -210,8 +231,17 @@ spec:
       labels:
         {{- include "<CHARTNAME>.selectorLabels" . | nindent 8 }}
     spec:
+    {{- with .Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+    {{- end }}
+      serviceAccountName: {{ template "<CHARTNAME>.serviceAccountName" . }}
```
unguiculus: `include` is used in all other places. Let's be consistent.

```suggestion
      serviceAccountName: {{ include "<CHARTNAME>.serviceAccountName" . }}
```

pkg/chartutil/create.go
```diff
@@ -259,6 +289,16 @@ spec:
     {{- include "<CHARTNAME>.selectorLabels" . | nindent 4 }}
 `
 
+const defaultServiceAccount = `{{- if .Values.serviceAccount.create -}}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ template "<CHARTNAME>.serviceAccountName" . }}
```
unguiculus:
```suggestion
  name: {{ include "<CHARTNAME>.serviceAccountName" . }}
```

pkg/chartutil/create.go
```diff
@@ -259,6 +289,16 @@ spec:
...
+  labels:
+{{ include "<CHARTNAME>.labels" . | indent 4 }}
```
unguiculus:
```suggestion
    {{ include "<CHARTNAME>.labels" . | nindent 4 }}
```

pkg/chartutil/create.go
```diff
@@ -334,6 +374,34 @@ Selector labels
...
+{{- end -}}
+{{- end -}}
+`
+
+const defaultTestConnection = `apiVersion: v1
+kind: Pod
+metadata:
+  name: "{{ include "<CHARTNAME>.fullname" . }}-test-connection"
+  labels:
+{{ include "<CHARTNAME>.labels" . | indent 4 }}
```
unguiculus:
```suggestion
    {{ include "<CHARTNAME>.labels" . | nindent 4 }}
```

unguiculus (changes requested):
A few remarks for consistency.

pkg/chartutil/create_test.go
bacongobbler: done

cmd/helm/create.go
bacongobbler: done!

unguiculus (approved):

hickeyma (approved):
LGTM, thanks @bacongobbler

pkg/chartutil/create.go
```diff
@@ -149,16 +173,13 @@ const defaultIgnore = `# Patterns to ignore when building packages.
...
+.vscode/
 `
 
 const defaultIngress = `{{- if .Values.ingress.enabled -}}
 {{- $fullName := include "<CHARTNAME>.fullname" . -}}
 {{- $svcPort := .Values.service.port -}}
-{{- if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion -}}
 apiVersion: networking.k8s.io/v1
-{{- else -}}
-apiVersion: extensions/v1beta1
```
marckhouzam: I'm curious how come this was removed?  Won't v3 need to work with clusters earlier than 1.14?

unguiculus: I would say the change is in line with the official Kubernetes version support policy (https://kubernetes.io/docs/setup/release/version-skew-policy/). Personally, I never liked having conditional logic for API versions in the starter chart. After all, this is not about Helm supporting older Kubernetes versions. It's about creating a starter chart which you will have to adjust to your needs anyways.

marckhouzam: Thanks for the clarification.  I guess I was surprised that this conditional API version logic was added 2 weeks ago and now removed.  But I had missed that this was the charter start code when I skimmed the change.

PR #6492, opened 2019-09-24
fix(cmd): lock repository file during `helm repo add`

bacongobbler:
port of #5678

### DIFF
--- a/cmd/helm/repo_add_test.go
+++ b/cmd/helm/repo_add_test.go
@@ -20,7 +20,10 @@ import (
 	"fmt"
 	"io/ioutil"
 	"path/filepath"
+	"sync"
 	"testing"
+
+	"gopkg.in/yaml.v2"
 
 	"helm.sh/helm/internal/test/ensure"
 	"helm.sh/helm/pkg/repo"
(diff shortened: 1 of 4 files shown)


cmd/helm/repo_add_test.go
```diff
@@ -86,3 +88,51 @@ func TestRepoAdd(t *testing.T) {
...
+	}
+
+	var name string
+	for i := 0; i < 3; i++ {
+		name = fmt.Sprintf("%s-%d", testName, i)
+		if !f.Has(name) {
+			t.Errorf("%s was not successfully inserted into %s: %s", name, repoFile, f.Repositories[0])
+		}
+	}
+}
```
thomastaylor312: Did we want to pull over the [subprocess test](https://github.com/helm/helm/pull/5678/files#diff-5537931905cc99a35fc6d296d4e3feadR154) as well? I figure this concurrency locking will most likely occur with multiple processes

thomastaylor312 (commented):
One discussion item before LGTM

cmd/helm/repo_add_test.go
bacongobbler: The subprocess test may likely be the cause of the circleci failures in helm 2. It's excessive, flaky, and unnecessary IMO. The parallel waitgroup test should be more than enough to cover the case

mattfarina:
@bacongobbler the failed CI is for a linting oops. Should be simple to fix here.

cmd/helm/repo_add_test.go
thomastaylor312: 👌

thomastaylor312 (approved):

PR #6498, opened 2019-09-25
Update glide and client code to use Kubernetes 1.16

jbrette:
Rebuild Helm using Kubernetes 1.16 go module.
- Changed glide.yaml
- Change couple of import path in cmd/client.go and cmd/client_test.go

**If applicable**:


bacongobbler:
This is something we cannot merge at this time unfortunately. By upgrading to 1.16 client libraries, we lose the ability to interact with APIs that work in 1.15. Helm tries to maintain backwards compatibility of n-1 Kubernetes releases, so we need to wait for 1.17 before we can consider bumping client-go.

PR #6500, opened 2019-09-25
Change release storage name to use helm storage type as prefix

### LINKED ISSUE #6435: Helm 2 and Helm 3 release version storage can clash
Helm 2 and Helm 3 use the same naming convention for release versions. It is as follows:

`RELEASE-NAME.v1 ..... RELEASE-NAME.vN, where 'N' is the incremental number of versions of the release RELEASE-NAME`

This storage for Helm 2 is as follows:
```console
$ kubectl get configmap/secret -n kube-system -l "OWNER=TILLER"
NAME         DATA   AGE
chrt-v2.v1   1      42s
chrt-v2.v2   1      12s
```

and for Helm 3:
```console
$ kubectl  get secrets -n default
NAME                  TYPE                                  DATA   AGE
chrt-v2.v1            helm.sh/release                       1      15s
chrt-v2.v2            helm.sh/release                       1      15s
```

Helm 2 and Helm 3 can be installed concurrently to manage the same cluster. This works when Helm 2 uses `ConfigMaps` for storage as Helm 3 uses `Secrets` for storage. **There is however a conflict when Helm 2 uses `Secrets` for storage and stores the release in the same namespace as the release.** The conflict occurs because Helm 3 uses different tags and ownership for the secret objects that Helm 2 does. It can therefore try to create a release that it thinks does not exist but will fail then because Helm 2 already has a secret with that name in that namespace. Here is an example of where this can occur: https://github.com/helm/helm-2to3/issues/30#issuecomment-531335572

Additionally, Helm 2 can be migrated to enable Helm 3 to manage releases previously handled by Helm 2: https://github.com/helm/helm-2to3.  This also works when Helm 2 uses `ConfigMaps` for storage as Helm 3 uses `Secrets` for storage. There is a conflict again however when using secrets because of the same naming convention. Refer to following issue for more details: https://github.com/helm/helm-2to3/issues/30

A possible solution around this would be for Helm 3 to use a different naming convention for release versions.

hickeyma:
Updating the release version name in storage to use the storage type as prefix:
` sh.helm.release.v1.<release_name>.v<version>`

Example:
```console
$ kubectl get secret --all-namespaces -l "owner=helm"
NAMESPACE   NAME                              TYPE              DATA   AGE
default     sh.helm.release.v1.chrt-6435.v1   helm.sh/release   1      6m34s
default     sh.helm.release.v1.chrt-6435.v2   helm.sh/release   1      5m58s
default     sh.helm.release.v1.chrt-6435.v3   helm.sh/release   1      5m53s
default     sh.helm.release.v1.chrt-6435.v4   helm.sh/release   1      40s
default     sh.helm.release.v1.chrt-next.v1   helm.sh/release   1      9s
```

Closes #6435 

Helm v2 stores release versions both as ConfigMaps (default) and Secrets under the Tiller namespace and `TILLER` ownership (type is Opaque). The naming convention used is: `<release_name>.v<version>`. This naming convention ensues uniqueness for Helm v2.

Helm v3  stores release versions as Secrets under the user namespace, `helm` ownership and type `helm.sh/release`. 

Helm v3 therefore cannot read Helm v2 releases because of different ownership and type. This is expected behaviour because Helm v2 and Helm v3 release objects are incompatible.

However, if:
- Helm v3 uses the same release version naming conventions as Helm v2 AND
- Helm v2 stores release object(s) as secrets under the namespace of the release (using the Tiller namespace for release) 

Then there will be a conflict:
- If Helm v3 tries to create a release with name already used by Helm v2 even though Helm v3 can't read it but the secret already exists OR
- Migrate Helm v3 releases over to Helm v3 because the 2 versions exist at same time until removed

This PR is to ensue uniqueness across Helm v2 and Helm v3 release versions in storage.

This is a breaking change to the current pre Helm 3 GA releases. Is this considered breaking as Helm 3 is not GA yet?  

**If applicable**:

### DIFF
--- a/pkg/storage/storage.go
+++ b/pkg/storage/storage.go
@@ -27,6 +27,11 @@ import (
 	"helm.sh/helm/pkg/storage/driver"
 )
 
+// The type field of the Kubernetes storage object which stores the Helm release
+// version. It is modified slightly replacing the '/': sh.helm/release.v1
+// Note: The version 'v1' is incremented if the release object metadata is
+// modified between major releases.
+// This constant is used as a prefix for the Kubernetes storage object name.
 const HelmStorageType = "sh.helm.release.v1"
 
 // Storage represents a storage engine for a Release.
@@ -207,9 +212,9 @@ func (s *Storage) Last(name string) (*rspb.Release, error) {
 	return h[0], nil
 }
 
-// makeKey concatenates the helm type, a release name and version into a
-// string with format:```<helm_storage_type>.<release_name>.v<version>```.
-// The Helm type is prepended to keep name uniqueness between different
+// makeKey concatenates the Kubernetes storage object type, a release name and version
+// into a string with format:```<helm_storage_type>.<release_name>.v<release_version>```.
+// The storage type is prepended to keep name uniqueness between different
 // release storage types. An example of clash when not using the type:
 // https://github.com/helm/helm/issues/6435.
 // This key is used to uniquely identify storage objects.
--- a/pkg/storage/storage.go
+++ b/pkg/storage/storage.go
@@ -27,6 +27,8 @@ import (
 	"helm.sh/helm/pkg/storage/driver"
 )
 
+const HelmStorageType = "sh.helm.release.v1"
+
 // Storage represents a storage engine for a Release.
 type Storage struct {
 	driver.Driver
@@ -205,11 +207,14 @@ func (s *Storage) Last(name string) (*rspb.Release, error) {
 	return h[0], nil
 }
 
-// makeKey concatenates a release name and version into
-// a string with format ```<release_name>#v<version>```.
+// makeKey concatenates the helm type, a release name and version into a
+// string with format:```<helm_storage_type>.<release_name>.v<version>```.
+// The Helm type is prepended to keep name uniqueness between different
+// release storage types. An example of clash when not using the type:
+// https://github.com/helm/helm/issues/6435.
 // This key is used to uniquely identify storage objects.
 func makeKey(rlsname string, version int) string {
-	return fmt.Sprintf("%s.v%d", rlsname, version)
+	return fmt.Sprintf("%s.%s.v%d", HelmStorageType, rlsname, version)
 }
 
 // Init initializes a new storage backend with the driver d.
--- a/pkg/storage/driver/memory.go
+++ b/pkg/storage/driver/memory.go
@@ -49,7 +49,8 @@ func (mem *Memory) Name() string {
 func (mem *Memory) Get(key string) (*rspb.Release, error) {
 	defer unlock(mem.rlock())
 
-	switch elems := strings.Split(key, ".v"); len(elems) {
+	keyWithoutPrefix := strings.TrimPrefix(key, "sh.helm.release.v1.")
+	switch elems := strings.Split(keyWithoutPrefix, ".v"); len(elems) {
 	case 2:
 		name, ver := elems[0], elems[1]
 		if _, err := strconv.Atoi(ver); err != nil {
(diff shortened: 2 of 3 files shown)


rimusz:
@hickeyma it should also change secret type to `helm.sh/release.v2` in https://github.com/helm/helm/blob/dev-v3/pkg/storage/driver/secrets.go#L237

rimusz:
LGTM

hickeyma:
> it should also change secret type to helm.sh/release.v1 

Good catch @rimusz, thanks. Updated now.

rimusz:
[2to3 plugin](https://github.com/helm/helm-2to3) built agains this PR showing it works properly:
```
$ helm3 2to3 convert mysql -t mysql --tiller-out-cluster --dry-run
NOTE: This is in dry-run mode, the following actions will not be executed.
Run without --dry-run to take the actions described below:

Release "mysql" will be converted from Helm 2 to Helm 3.
[Helm 3] Release "mysql" will be created.
[Helm 3] ReleaseVersion "mysql.v1" will be created.

$ kubectl -n mysql get secret
NAME                          TYPE                                  DATA   AGE
default-token-wt2ms           kubernetes.io/service-account-token   3      2m50s
mysql                         Opaque                                2      2m18s
mysql.v1                      Opaque                                1      2m18s

$ helm3 2to3 convert mysql -t mysql --tiller-out-cluster
Release "mysql" will be converted from Helm 2 to Helm 3.
[Helm 3] Release "mysql" will be created.
[Helm 3] ReleaseVersion "mysql.v1" will be created.
[Helm 3] ReleaseVersion "mysql.v1" created.
[Helm 3] Release "mysql" created.
Release "mysql" was converted successfully from Helm 2 to Helm 3. Note: the v2 releases still remain and should be removed to avoid conflicts with the migrated v3 releases.

$ kubectl -n mysql get secret
NAME                          TYPE                                  DATA   AGE
default-token-wt2ms           kubernetes.io/service-account-token   3      2m50s
mysql                         Opaque                                2      2m18s
mysql.v1                      Opaque                                1      2m18s
sh.helm.release.v1.mysql.v1   helm.sh/release.v1                    1      66s

$ helm3 ls -n mysql
NAME     NAMESPACE    REVISION    UPDATED                                 STATUS      CHART
mysql    mysql        1           2019-09-25 11:26:38.131325 +0000 UTC    deployed    mysql-1.3.1
```

mattfarina (commented):
This change generally looks good to me. Secret names can be long enough well run into other problems long before the secret name is one of them due to length.

Because this change will break current beta users I'd like one more person to approve and have their eyes on it. This is something that should be in the announcement about the next beta as well.

pkg/storage/storage.go
```diff
@@ -27,6 +27,8 @@ import (
 	"helm.sh/helm/pkg/storage/driver"
 )
 
+const HelmStorageType = "sh.helm.release.v1"
```
bacongobbler: what does the `v1` mean in this case?

rimusz: it was implemented as per @mattfarina comments in https://github.com/helm/helm/issues/6435#issuecomment-533591950

bacongobbler: I'm asking because it's not immediately apparent what the significance of that number is. Why does it symbolize? When is it incremented?

If it's intended as a versioning scheme for the release name, then there should be a comment explaining the significance of that storage type.

rimusz: @hickeyma can you please update PR with the comments, thanks

hickeyma: @bacongobbler This is a good point especially from a maintenance perspective. I have added comments to the code. Let me know if you need any more clarification.

bacongobbler (approved):
LGTM

pkg/storage/storage.go
bacongobbler: Thank you for the clarification, @mattfarina, and thank you for writing it down, @hickeyma!

hickeyma: Not a bother. FYI, I didn't bother responding in the PR as I hoped the description in the code comments would provide the relevant information.

mattfarina (approved):

hickeyma:
Thanks for reviews and feedback @bacongobbler @mattfarina @rimusz

Issue #5892, opened 2019-06-14
Migration: Helm 2 to Helm 3


hickeyma:
This feature is to cover migration from Helm 2 to Helm 3.

The key changes in Helm 3 which will directly affect Helm 2 are as follows:
- Release state storage. The changes includes the Kubernetes resource for storage and the release object metadata contained in the resource. Releases are now on a per user namespace instead of default namespace like `kube-system`.
- Charts deployed per user namespace and user no longer specifies the namespace explicitly. Namespaces need to be pre-created.
- Commands removed:
  - `helm reset`
  - `helm serve`
- The [Helm official Charts repository](https://github.com/helm/charts) is deprecated. It will be replaced by [Helm Hub](https://hub.helm.sh/) which is an  aggregator for different chart hosts.
- Release binaries are now hosted on `get.helm.sh`

The requirements for Helm 3 migrations are as follows:
- Helm 2 charts `install/template/--dry-run` as is
- Helm 2 and Helm 3 can work concurrently in the same cluster
- Helm 2 release and its history can be converted to Helm 3 release
- Helm 2 release and its history can be removed with and without removing the underlying Kubernetes resources
- Moving Helm 2 to 3 configuration 
- Ability to list releases by `--all-namespaces`
- Helm 2 data cleanup

Closes: https://github.com/helm/community/issues/67  

Tasks:

- [x] Migration documentation (https://github.com/helm/helm-www/pull/287 and #5582)
- [x] Tool for migrating Helm v2 releases in-place to v3 (#6154 and [2to3 plugin](https://github.com/helm/helm-2to3))
- [x] Tool for moving Helm 2 to 3 configuration ([2to3 plugin](https://github.com/helm/helm-2to3))
- [x] Tool for Helm 2 data cleanup ([2to3 plugin](https://github.com/helm/helm-2to3))

hickeyma:
Investigations currently underway and progress shown in WIP doc PR #5582. In conjunction with the doc, there will be tools needed for the migration process.

One of these tools will be for migrating Helm 2 release objects to Helm 3. The tool was previously mentioned in https://github.com/helm/community/issues/67#issuecomment-448045222. 

From yesterdays dev weekly meeting, it’s suggested to write a proposal on the differences between Helm 2 and Helm 3’s object metadata and a how to migrate this data.

hickeyma:
I will attach the proposal when ready.

bacongobbler:
[...]

PR #6502, opened 2019-09-25
v3: Support auto-complete of a renamed helm binary

marckhouzam:

This is in the context of the v2 to v3 migration where a user is likely
to rename the helm binary to helm3 (see #5892). But it is a valid improvement even
outside this migration.

If a user renames the helm binary, the shell needs to be told that this
new name corresponds to the helm completion function. For example if
the user decides to rename the helm v3 binary to `helm3`, then the
following extra command would need be run by the user:
```
complete -o default -F __start_helm helm3
```
This PR automates this by adding this extra command to the generated
shell completion script by looking at what binary was used to call the
helm completion command.  For example, if the user renames the binary
to `helm3` and then calls
```
helm3 completion bash
```
the completion script will include a hook for the binary 'helm3' to the completion script.

This PR also has the zsh completion command call the logic of the bash completion command so as to easily get any improvements to bash completion, such as this one.

The acceptance-testing tests continue to all pass with this change.

**If applicable**:

### DIFF
--- a/cmd/helm/completion.go
+++ b/cmd/helm/completion.go
@@ -16,6 +16,7 @@ limitations under the License.
 package main
 
 import (
+	"fmt"
 	"io"
 	"os"
 	"path/filepath"
@@ -88,12 +89,12 @@ func runCompletionBash(out io.Writer, cmd *cobra.Command) error {
 # to the helm completion function to handle the case where
 # the user renamed the helm binary
 if [[ $(type -t compopt) = "builtin" ]]; then
-    complete -o default -F __start_helm ` + binary + `
+    complete -o default -F __start_helm %[1]s
 else
-    complete -o default -o nospace -F __start_helm ` + binary + `
+    complete -o default -o nospace -F __start_helm %[1]s
 fi
 `
-		out.Write([]byte(renamedBinaryHook))
+		fmt.Fprintf(out, renamedBinaryHook, binary)
 	}
 
 	return err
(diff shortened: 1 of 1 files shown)


marckhouzam:
As a reminder, to run the completion tests from the helm repo, one can do:
```
```

cmd/helm/completion.go
```diff
@@ -77,7 +78,25 @@ func runCompletion(out io.Writer, cmd *cobra.Command, args []string) error {
...
+
+	// In case the user renamed the helm binary (e.g., to be able to run
+	// both helm2 and helm3), we hook the new binary name to the completion function
+	if binary := filepath.Base(os.Args[0]); binary != "helm" {
+		renamedBinaryHook := `
+# Hook the command used to generate the completion script
+# to the helm completion function to handle the case where
+# the user renamed the helm binary
+if [[ $(type -t compopt) = "builtin" ]]; then
+    complete -o default -F __start_helm ` + binary + `
```
thomastaylor312: Consider changing this to use format notation (`%s`) and then doing a `fmt.Fprintf(out, renamedBinaryHook, binary, binary)`. It is a bit safer than trying to glue the various string components together

thomastaylor312 (commented):
Just one comment on code structure. I'll be testing this manually soon

thomastaylor312:
Marking this as a bug after our discussion in the Helm dev call. This is technically a bug because autocompletion does not work when you have helm 2 and 3 installed together

cmd/helm/completion.go
marckhouzam: Better.  Thanks.

thomastaylor312 (approved):
This is looking great! Thanks for the fix

PR #6504, opened 2019-09-25
feat(*): Ports all output functionality from v2

### LINKED ISSUE #6437: Parameter --output in helm3 list function not available
Hi, 
We plan to use Helm in version 3 in the future and have already tested the current beta version (Beta 3). Unfortunately the parameter --output is not available in the list function to get the data either as json or yaml. Is this function completely removed for the future or will it be available again in the release?

Output of `helm version`:
```
Version.BuildInfo{Version:"v3.0.0-beta.3", GitCommit:"5cb923eecbe80d1ad76399aee234717c11931d9a", GitTreeState:"clean", GoVersion:"go1.12.9"}
```

Greetings niesel

thomastaylor312:
As part of this port, I removed some now superfluous code from the `action` package.
This is technically a breaking change, but since the package was introduced in v3, it
is highly unlikely anyone is using it and we are still within the beta window. If we merge, we'll need to add the breaking change note to the next beta

Also closes #6437


cmd/helm/list.go
```diff
@@ -95,8 +104,60 @@ func newListCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
...
-	f.IntVarP(&client.Offset, "offset", "o", 0, "next release name in the list, used to offset from start value")
+	f.IntVar(&client.Offset, "offset", 0, "next release name in the list, used to offset from start value")
 	f.StringVarP(&client.Filter, "filter", "f", "", "a regular expression (Perl compatible). Any releases that match the expression will be included in the results")
+	bindOutputFlag(cmd, &client.OutputFormat)
 
 	return cmd
 }
+
+type releaseElement struct {
+	Name      string
```
thomastaylor312: There is some variance among the commands with key name capitalization that I kept from Helm 2. If we'd like to add some JSON tags to lowercase when marshaling, let me know and we can change it

cmd/helm/list.go
```diff
@@ -95,8 +104,60 @@ func newListCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 	f.BoolVar(&client.Pending, "pending", false, "show pending releases")
 	f.BoolVar(&client.AllNamespaces, "all-namespaces", false, "list releases across all namespaces")
 	f.IntVarP(&client.Limit, "max", "m", 256, "maximum number of releases to fetch")
-	f.IntVarP(&client.Offset, "offset", "o", 0, "next release name in the list, used to offset from start value")
```
thomastaylor312: `o` was already being used on all other commands as the short hand for output, if we think that removing this shorthand is going to be too big of a change, let me know

adamreese (approved):
Looks sane to me

cmd/helm/search_repo.go
```diff
@@ -166,3 +161,60 @@ func (o *searchRepoOptions) buildIndex(out io.Writer) (*search.Index, error) {
 	}
 	return i, nil
 }
+
+type repoChartElement struct {
```
adamreese: many of these could be converted to anonymous structs

pkg/action/output.go
```diff
@@ -87,3 +72,54 @@ func ParseOutputFormat(s string) (out OutputFormat, err error) {
 	}
 	return
 }
+
+// Writer is an interface that any type can implement to write supported formats
+type Writer interface {
```
adamreese: 2 cents:  you could have the abstraction be `Write()` and types for table,json,yaml.  This would protect from having to add a new method to this interface to add a formatter.

thomastaylor312: My thinking with this interface is that it will cause a compilation error if someone adds a new type and doesn't implement it for everything, but if you think this would be cleaner, I can give it a whirl

bacongobbler (approved):
LGTM

pkg/action/output.go
thomastaylor312: Going to merge for now. If you think this would help, let me know and I can do a follow up PR

PR #6508, opened 2019-09-26
ref(*): refactor flag and environment settings

adamreese:


pkg/plugin/plugin.go
```diff
@@ -108,15 +108,22 @@ type Plugin struct {
 // - If both OS and Arch match the current platform, search will stop and the command will be prepared for execution
 // - If OS matches and there is no more specific match, the command will be prepared for execution
 // - If no OS/Arch match is found, return nil
-func getPlatformCommand(cmds []PlatformCommand) []string {
+func getPlatformCommand(cmds []PlatformCommand, env map[string]string) []string {
+	getenv := func(name string) string {
```
thomastaylor312: I am not entirely understanding what value the environment map adds to these functions if we are already setting them with `os.Setenv` in `SetupPluginEnv`. I like that this allows for the ability to pass an arbitrary map of overrides that fallback on the environment variables, but am not sure if that is how they need to be used. Am I missing something here?

thomastaylor312 (commented):
A few points of discussion. Otherwise a great looking refactor

pkg/cli/environment.go
```diff
@@ -51,29 +45,29 @@ type EnvSettings struct {
...
 // AddFlags binds flags to the given flagset.
 func (s *EnvSettings) AddFlags(fs *pflag.FlagSet) {
-	fs.StringVarP(&s.Namespace, "namespace", "n", s.Namespace, "namespace scope for this request")
-	fs.StringVar(&s.KubeConfig, "kubeconfig", "", "path to the kubeconfig file")
-	fs.StringVar(&s.KubeContext, "kube-context", "", "name of the kubeconfig context to use")
 	fs.BoolVar(&s.Debug, "debug", s.Debug, "enable verbose output")
 	fs.StringVar(&s.RegistryConfig, "registry-config", s.RegistryConfig, "path to the registry config file")
 	fs.StringVar(&s.RepositoryConfig, "repository-config", s.RepositoryConfig, "path to the file containing repository names and URLs")
 	fs.StringVar(&s.RepositoryCache, "repository-cache", s.RepositoryCache, "path to the file containing cached repository indexes")
+	s.KubeConfig.AddFlags(fs)
```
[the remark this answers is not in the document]
adamreese: Same to me, I would like to at least hide some of the flags here and for the logging setup

pkg/plugin/plugin.go
adamreese: It is to move away from depending on the users environment.  Depending on os envvars makes testing difficult by polluting the environment with overrides.

thomastaylor312: OK, that sounds like a good enough reason to me 🙂

pkg/cli/environment.go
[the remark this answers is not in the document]
thomastaylor312: I think we should not add the flags now and if the need comes up later, we can add this back in. I'm not married to the idea, so if you like it as is now, that should be fine!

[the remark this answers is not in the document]
thomastaylor312: After looking at this again, I think this is ok to have

pkg/cli/environment.go
```diff
@@ -91,5 +101,6 @@ func (s *EnvSettings) EnvVars() map[string]string {
 		"HELM_REGISTRY_CONFIG":   s.RegistryConfig,
 		"HELM_REPOSITORY_CACHE":  s.RepositoryCache,
 		"HELM_REPOSITORY_CONFIG": s.RepositoryConfig,
+		"HELM_NAMESPACE":         s.Namespace(),
```
databus23: Trying to adopt the helm-diff plugin for helm3 it came up that plugins are unaware of  `--kube-context` and `--namespace` passed to the main binary. This makes it impossible to talk to the k8s cluster used by helm:
https://github.com/databus23/helm-diff/pull/151#issuecomment-537679821

Not sure if this is the right place but how about passing the values from `--kube0context` as `HELM_KUBECONTEXT`?

thomastaylor312: Also of note from that issue in helm-diff is that @databus23 pointed out that we still have `--kube-context` as a top level flag in the helm CLI, and now we will also have `--context` due to the flags we attach with the `AddFlags` method

mumoshu: A dedicated issue opened for adding those envvars: https://github.com/helm/helm/issues/6631

bacongobbler:
Hey @adamreese! How are we on this PR? Do you need another set of eyes to review this one?

PR #6518, opened 2019-09-27
v2: feat(comp) Dynamic completion to use same binary as main call

marckhouzam:
The binary of Helm to use for dynamic completion should be the same
as the actual Helm binary being used.  For example, if PATH points
to a version of helm v3, but the user calls a binary named helm2 to
use a renamed v2 version, then dynamic completion should also use helm2.
If not, in this example, the dynamic completion will use the
information returned by helm v3.

This improvement is particularly useful for users that will run both
helm v2 and helm v3 at the same time.

**If applicable**:

### DIFF
--- a/cmd/helm/helm.go
+++ b/cmd/helm/helm.go
@@ -69,13 +69,22 @@ __helm_override_flags()
     done
 }
 
+__helm_binary_name()
+{
+    local helm_binary
+    helm_binary="${words[0]}"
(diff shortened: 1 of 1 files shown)


bacongobbler (approved):
LGTM!

marckhouzam:
This one is approved, it just needs to be merged.
Thanks!

PR #5498, opened 2019-03-22
Migrating dependency management to go modules

mattfarina:
This PR migrates from dep to go modules for dependency management.

Some notes:

* Helm is now outside the GOPATH for the working directory. This causes go modules to automatically be in use for all go commands
* The cache has been switched to go modules. When the GOPATH is deprecated where will these things be cached?
* `go mod` manages the `go.mod` file and sorts things in ascii alpha order. If you change that and use a `go mod edit` command it will re-sort

**What this PR does / why we need it**:

**Special notes for your reviewer**:

**If applicable**:
- [ ] this PR contains documentation
- [ ] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

PR #6519, opened 2019-09-27
fix(pkg/chartutil): include values.schema.json in packaged chart

SimonAlling:
When playing around with new Helm 3 features, we noticed that `values.yaml` is not validated when running `helm install`, even though [the docs say it should be](https://v3.helm.sh/docs/topics/charts/#schema-files):

> Validation occurs when any of the following commands are invoked:
> 
> * `helm install`
> * `helm upgrade`
> * `helm lint`
> * `helm template`

The reason is that `values.schema.json` is not included in the tarball created by `helm package`, so `helm lint` cannot lint the tarball properly, even if it can lint the chart directory:

    $ helm lint my-chart            # Finds errors in values.yaml
    $ helm package my-chart
    $ helm lint my-chart-1.0.0.tgz  # Does not find errors in values.yaml

### DIFF
--- a/pkg/chartutil/save.go
+++ b/pkg/chartutil/save.go
@@ -19,6 +19,7 @@ package chartutil
 import (
 	"archive/tar"
 	"compress/gzip"
+	"encoding/json"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -52,6 +53,14 @@ func SaveDir(c *chart.Chart, dest string) error {
 		vf := filepath.Join(outdir, ValuesfileName)
 		b, _ := yaml.Marshal(c.Values)
 		if err := writeFile(vf, b); err != nil {
+			return err
+		}
+	}
+
+	// Save values.schema.json if it exists
+	if c.Schema != nil {
+		filename := filepath.Join(outdir, SchemafileName)
+		if err := writeFile(filename, c.Schema); err != nil {
 			return err
 		}
 	}
@@ -147,6 +156,16 @@ func writeTarContents(out *tar.Writer, c *chart.Chart, prefix string) error {
 	}
 	if err := writeToTar(out, filepath.Join(base, ValuesfileName), ydata); err != nil {
 		return err
+	}
+
+	// Save values.schema.json if it exists
+	if c.Schema != nil {
+		if !json.Valid(c.Schema) {
+			return errors.New("Invalid JSON in " + SchemafileName)
+		}
+		if err := writeToTar(out, filepath.Join(base, SchemafileName), c.Schema); err != nil {
+			return err
+		}
 	}
 
 	// Save templates
--- a/pkg/chartutil/save.go
+++ b/pkg/chartutil/save.go
@@ -133,6 +133,7 @@ func Save(c *chart.Chart, outDir string) (string, error) {
 
 	if err := writeTarContents(twriter, c, ""); err != nil {
 		rollback = true
+		return filename, err
 	}
 	return filename, err
 }
--- a/pkg/chartutil/create.go
+++ b/pkg/chartutil/create.go
@@ -35,6 +35,8 @@ const (
 	ChartfileName = "Chart.yaml"
 	// ValuesfileName is the default values file name.
 	ValuesfileName = "values.yaml"
+	// SchemafileName is the default values schema file name.
+	SchemafileName = "values.schema.json"
 	// TemplatesDir is the relative directory name for templates.
 	TemplatesDir = "templates"
 	// ChartsDir is the relative directory name for charts dependencies.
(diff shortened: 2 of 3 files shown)


bacongobbler:
Good catch. Thank you for the PR! Would you mind writing some unit tests under save_test.go to ensure this file is indeed packaged? Thanks!

lindhe:
We would actually mind a little bit. Unless we have missed it, there seems to be no unit test for any of the other required files that are packaged (e.g., Chart.yaml), it would require some effort to implement it. Since we are new to the project, it would be a lot of trial-and error before we could get the tests in place. Maybe it's better to create a new PR after this one, that implements the entire suite of tests needed for the files?

bacongobbler:
It shouldn't be too hard to add a `Schema` object [here](https://github.com/helm/helm/blob/15d277360e1406cbf708e5aef599bc7e0e685e9b/pkg/chartutil/save_test.go#L77-L89) and then check if `.Schema` hasn't changed, similar to how [we check the `.Files` object here](https://github.com/helm/helm/blob/15d277360e1406cbf708e5aef599bc7e0e685e9b/pkg/chartutil/save_test.go#L108-L110). That should be enough to cover the code introduced in this change.

bacongobbler:
A test case for saving invalid JSON would also help capture the validation check.

SimonAlling:
> It shouldn't be too hard to add a `Schema` object [here](https://github.com/helm/helm/blob/15d277360e1406cbf708e5aef599bc7e0e685e9b/pkg/chartutil/save_test.go#L77-L89) and then check if `.Schema` hasn't changed after it's been loaded, similar to how [we check the `.Files` object](https://github.com/helm/helm/blob/15d277360e1406cbf708e5aef599bc7e0e685e9b/pkg/chartutil/save_test.go#L108-L110). That should be enough to cover the code introduced in this change.

I absolutely agree.

> A test case for saving invalid JSON would also help capture the validation check.
> 
> https://github.com/helm/helm/pull/6519/files#diff-9ea0c476e7eb107c6b0f198af1a49caaR163

I'm not entirely sure about this, because as far as we can tell, an error returned from `writeTarContents` doesn't "propagate" to the user. Ergo, if I run `helm package` and something goes wrong such that any of the `return err` statements in `writeTarContents` is reached, then no tarball is created, but `helm package` still prints a success message:

    $ helm package my-chart
    Successfully packaged chart and saved it to: /[...]/my-chart-1.0.0.tgz
    $ ls my-chart-1.0.0.tgz
    ls: cannot access 'my-chart-1.0.0.tgz': No such file or directory

But that seems to us like a separate issue.

(We may not have time to work on this during the weekend, but hopefully on Monday.)

SimonAlling:
I've added the requested test cases now, @bacongobbler. There was a particular confusing aspect of the code – namely, how the name `err` was used – that took me several hours to figure out, but I think that's gonna be a separate PR.

pkg/chartutil/save.go
```diff
@@ -124,6 +133,7 @@ func Save(c *chart.Chart, outDir string) (string, error) {
 
 	if err := writeTarContents(twriter, c, ""); err != nil {
 		rollback = true
+		return filename, err
```
SimonAlling: Seems like I fixed an unrelated (as in "present before this PR") bug here.

SimonAlling:
Hi again, @bacongobbler. I rebased and force-pushed my commits because conflicts had arisen due to #5498.

pkg/chartutil/save.go
```diff
@@ -56,6 +57,14 @@ func SaveDir(c *chart.Chart, dest string) error {
 		}
 	}
 
+	// Save values.schema.json if it exists
+	if c.Schema != nil {
```
bacongobbler: Do we want to check if the schema is valid here as well?

```
if !json.Valid(c.Schema) {
	return errors.New("Invalid JSON in " + SchemafileName)
}
```

SimonAlling: I don't know. But on [line 54](https://github.com/helm/helm/blob/3c977a905518ad54f92df823ed5b4ca488fd6aeb/pkg/chartutil/save.go#L54), `values.yaml` is marshaled without any error being considered. Compare this to [line 154](https://github.com/helm/helm/blob/3c977a905518ad54f92df823ed5b4ca488fd6aeb/pkg/chartutil/save.go#L154), where any marshaling error is caught and returned. I think that's why I did the validation in `writeTarContents`, but not in `SaveDir`.

bacongobbler (approved):
LGTM as-is. We can address schema validation in a follow-up. Thanks!

bacongobbler:
Would you mind rebasing this PR? Once that's done I can merge this. Thanks.

Issue #6140, opened 2019-08-01
EOF when loading the same chart concurrently


aaronmell:
We have a bunch of microservices that use the same helm chart. We are using the terraform provider for helm, which deploys all of our microservices concurrently. We randomly see EOF errors returning from the provider. I've traced the issue into the LoadFile function inside the chartutil by adding logging to the terraform provider.

Failed to load chart from path /var/lib/jenkins/.helm/cache/archive/microservice-0.8.4.tgz, EOF

Output of `helm version`:
Client: &version.Version{SemVer:"v2.14.2", GitCommit:"a8b13cc5ab6a7dbef0a58f5061bcc7c0c61598e7", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.14.2", GitCommit:"a8b13cc5ab6a7dbef0a58f5061bcc7c0c61598e7", GitTreeState:"clean"}

Output of `kubectl version`:
Client Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.1", GitCommit:"eec55b9ba98609a46fee712359c7b5b365bdd920", GitTreeState:"clean", BuildDate:"2018-12-13T19:44:19Z", GoVersion:"go1.11.2", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"12+", GitVersion:"v1.12.6-eks-d69f1b", GitCommit:"d69f1bf3669bf00b7f4a758e978e0e7a1e3a68f7", GitTreeState:"clean", BuildDate:"2019-02-28T20:26:10Z", GoVersion:"go1.10.8", Compiler:"gc", Platform:"linux/amd64"}

Cloud Provider/Platform (AKS, GKE, Minikube etc.): 
EKS

karuppiah7890:
I have actually run into this too in our pipeline. Never thought of submitting the issue. Triggering pipeline solved the problem. There was one more regarding some "pipe broken" issue. I assumed it was some port forward issue or something and again pipeline retrigger solved it. 

I think I should start checking all such issues and submit it.

Regarding this issue, I'll try to find a proper way to reproduce this and try to raise a PR to fix this :)

karuppiah7890:
I have actually run into this too in our pipeline. Never thought of submitting the issue. Triggering pipeline solved the problem. There was one more regarding some "pipe broken" issue. I assumed it was some port forward issue or something and again pipeline retrigger solved it. 

I think I should start checking all such issues and submit it.

[...]

PR #6536, opened 2019-09-30
fix silent lint ignore for non existing packaged charts

### LINKED ISSUE #6428: Helm v3: lint silently ignores non existing packaged charts
Output of `helm version`:
```
$ helm3 version

$ helm version --client
Client: &version.Version{SemVer:"v2.14.3", GitCommit:"0e7f3b6637f7af8fcfddb3d2941fcc7cbebb0085", GitTreeState:"clean"}
```
Output of `kubectl version`:
N/A

N/A

----

If you try to lint a packaged chart that doesn't exist, helm3 silently ignores it.
 
```
$ helm3 lint aaa-1.0.0.tgz
==> Linting aaa-1.0.0.tgz

```

```
$ helm lint aaa-1.0.0.tgz
==> Skipping aaa-1.0.0.tgz
open aaa-1.0.0.tgz: no such file or directory

0 chart(s) linted, no failures
```

karuppiah7890:
Fixes #6428 

**If applicable**:

### DIFF
--- a/pkg/action/lint.go
+++ b/pkg/action/lint.go
@@ -29,8 +29,6 @@ import (
 	"helm.sh/helm/pkg/lint/support"
 )
 
-var errLintNoChart = errors.New("no chart found for linting (missing Chart.yaml)")
-
 // Lint is the action for checking that the semantics of a chart are well-formed.
 //
 // It provides the implementation of 'helm lint'.
@@ -56,24 +54,19 @@ func (l *Lint) Run(paths []string, vals map[string]interface{}) *LintResult {
 	if l.Strict {
 		lowestTolerance = support.WarningSev
 	}
-
 	result := &LintResult{}
 	for _, path := range paths {
 		linter, err := lintChart(path, vals, l.Namespace, l.Strict)
 		if err != nil {
-			if err == errLintNoChart {
-				result.Errors = append(result.Errors, err)
-			}
-			if linter.HighestSeverity >= lowestTolerance {
-				result.Errors = append(result.Errors, err)
-			}
-		} else {
-			result.Messages = append(result.Messages, linter.Messages...)
-			result.TotalChartsLinted++
-			for _, msg := range linter.Messages {
-				if msg.Severity == support.ErrorSev {
-					result.Errors = append(result.Errors, msg.Err)
-				}
+			result.Errors = append(result.Errors, err)
+			continue
+		}
+
+		result.Messages = append(result.Messages, linter.Messages...)
+		result.TotalChartsLinted++
+		for _, msg := range linter.Messages {
+			if msg.Severity >= lowestTolerance {
+				result.Errors = append(result.Errors, msg.Err)
 			}
 		}
 	}
@@ -87,18 +80,18 @@ func lintChart(path string, vals map[string]interface{}, namespace string, stric
 	if strings.HasSuffix(path, ".tgz") {
 		tempDir, err := ioutil.TempDir("", "helm-lint")
 		if err != nil {
-			return linter, err
+			return linter, errors.Wrap(err, "unable to create temp dir to extract tarball")
 		}
 		defer os.RemoveAll(tempDir)
 
 		file, err := os.Open(path)
 		if err != nil {
-			return linter, err
+			return linter, errors.Wrap(err, "unable to open tarball")
 		}
 		defer file.Close()
 
 		if err = chartutil.Expand(tempDir, file); err != nil {
-			return linter, err
+			return linter, errors.Wrap(err, "unable to extract tarball")
 		}
 
 		lastHyphenIndex := strings.LastIndex(filepath.Base(path), "-")
@@ -113,7 +106,7 @@ func lintChart(path string, vals map[string]interface{}, namespace string, stric
 
 	// Guard: Error out if this is not a chart.
 	if _, err := os.Stat(filepath.Join(chartPath, "Chart.yaml")); err != nil {
-		return linter, errLintNoChart
+		return linter, errors.Wrap(err, "unable to check Chart.yaml file in chart")
 	}
 
 	return lint.All(chartPath, vals, namespace, strict), nil
--- a/cmd/helm/testdata/testcharts/chart-with-no-templates-dir/Chart.yaml
+++ b/cmd/helm/testdata/testcharts/chart-with-no-templates-dir/Chart.yaml
@@ -1,0 +1,5 @@
+apiVersion: v1
+name: chart-with-no-templates-dir
+description: an example chart
+version: 199.44.12345-Alpha.1+cafe009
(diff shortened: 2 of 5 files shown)


thomastaylor312 (changes requested):
Thanks for the PR! There are a few changes needed here to handle the errors a bit better. I am also wondering if the `lintError` function should set `ErrorSev` as the highest severity [here](https://github.com/helm/helm/pull/6536/files#diff-8254eeb06bb9f2a4e69f4176be9ef319R85) to make sure that any returned error is added to the `result.Errors` slice in `Run`. That would also solve the problem without needed to return `errLintNoChart` for everything because right now, pretty much any error is swallowed

pkg/action/lint.go
```diff
@@ -93,12 +93,12 @@ func lintChart(path string, vals map[string]interface{}, namespace string, stric
...
 		file, err := os.Open(path)
 		if err != nil {
-			return linter, err
+			return linter, errLintNoChart
 		}
 		defer file.Close()
 
 		if err = chartutil.Expand(tempDir, file); err != nil {
-			return linter, err
+			return linter, errLintNoChart
```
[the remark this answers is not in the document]
karuppiah7890: Yeah, makes sense. I think like you said, let's not swallow these errors. They could be permission issues, corrupted charts and I don't know what other errors. I would say let's not label it too. Let's first bubble it up and fail the lint?

pkg/action/lint.go
```diff
@@ -93,12 +93,12 @@ func lintChart(path string, vals map[string]interface{}, namespace string, stric
 
 		file, err := os.Open(path)
 		if err != nil {
-			return linter, err
+			return linter, errLintNoChart
```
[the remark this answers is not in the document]
karuppiah7890: You are right. Errors like permission issues could occur. True. So, if other errors could occur, why not just send the whole error itself and then fail the lint? "Not exist" is just one of the errors. Or do we want to handle that separately? However I think the error will show up in the console

karuppiah7890:
@thomastaylor312 So, what do you feel? Do you want to specifically handle errors that a chart is corrupted / chart is missing and then just bubble up other errors as is and fail lint?

karuppiah7890:
If we bubble out errors, it's not so bad for some of them. Some examples of permission denied and EOF errors. I'm not sure in what cases EOF occurs. One I can think of is, a `.tgz` not really being a tgz, which is what I did. I just did `touch meh.tgz`. Another is a corrupted one - like only half the content is downloaded from Internet or some place? It's a guess from this error - https://github.com/helm/helm/issues/6140

```
$ helm3 lint blah-0.1.0.tgz
==> Linting blah-0.1.0.tgz
Error open blah-0.1.0.tgz: permission denied

$ helm3 lint dummy-chart/
==> Linting dummy-chart/
Error stat dummy-chart/Chart.yaml: permission denied

$ helm3 lint meh.tgz
==> Linting meh.tgz
Error EOF

```

karuppiah7890:
If we just bubble up, I think we don't need a `errLintNoChart` variable as it'll not be used. May be instead of blindly bubbling up errors, we can say what we tried to do and what the error was. Like `opening tgz file: open blah-0.1.0.tgz: permission denied`, `extracting tgz file: EOF`

karuppiah7890:
I'll also make these changes to #6537 PR

karuppiah7890:
There are actually so many kinds of errors that occur in functions like [`chartutil.Expand`](https://github.com/helm/helm/blob/4f26b658d867569ba69db9324be64d3dfdee2440/pkg/chartutil/expand.go#L28)

karuppiah7890:
I have removed the `lowestTolerance` as it was only being used when there were errors returned by `lintChart` and every error `lintChart` returned was actually a valid and high severity. There was no possibility for warnings or infos over there, as we are clearly returning an error. If it was some warning or info or some unknown, we would put it in linter messages and returned nil error.

So, got rid of the extra code for when error occurs and wrapped the errors with some meaningful messages to understand the error. Let me know if it makes sense. I'll also see if I need to add more tests!

pkg/action/lint.go
```diff
@@ -87,18 +76,18 @@ func lintChart(path string, vals map[string]interface{}, namespace string, stric
 	if strings.HasSuffix(path, ".tgz") {
 		tempDir, err := ioutil.TempDir("", "helm-lint")
 		if err != nil {
-			return linter, err
+			return linter, errors.Errorf("creating temp dir to extract tarball: %s", err.Error())
```
thomastaylor312: All of these should just be `errors.Wrap(err, "message")`

thomastaylor312: Nit: The errors should probably be more clear with a message like "unable to create temp dir..."

thomastaylor312 (changes requested):
So I think this is looking good, just a few small changes to the errors. I would like another maintainer who has a bit more experience with the lint code double check that this change won't have a larger impact

pkg/action/lint.go
```diff
@@ -52,28 +50,19 @@ func NewLint() *Lint {
...
-				}
+			result.Errors = append(result.Errors, err)
+			continue
+		}
+
+		result.Messages = append(result.Messages, linter.Messages...)
+		result.TotalChartsLinted++
+		for _, msg := range linter.Messages {
+			if msg.Severity == support.ErrorSev {
+				result.Errors = append(result.Errors, msg.Err)
```
mattfarina: I think we might have a bug that was in the old code that was carried over to the new logic. When `l.Strict` is false `support.ErrorSev` is the error level. When `l.Strict` is true `support.WarningSev` should be used for appending to `result.Errors`.

This is important for the `helm lint` command as the different levels tell it when to return a non-zero exit code to the OS.

In the old code `lowestTolerance` was doing this when an err was returned from `lintChart`.  Could you fix this logic?

For example...
```
lowestTolerance := support.ErrorSev
if l.Strict {
	lowestTolerance = support.WarningSev
}
```

and then later...
```
if msg.Severity >= lowestTolerance {
```

mattfarina (changes requested):
There appears to be one small bug in the code around strict handling

pkg/action/lint.go
karuppiah7890: Ah sure. I'll add some failing tests and do it. And I did doubt this, but I really didn't know how lint was entirely working. I had read some code inside lint.All and sorts, to understand things like highest severity, not much at the higher level. I better read more code 😅 

So this way we will also be getting warnings along with errors in strict case but only errors in a normal (non strict) case

karuppiah7890:
Thanks for the reviews @thomastaylor312 and @mattfarina , I have incorporated them in my PR 😄 please do check it out!

thomastaylor312 (approved):
This looks good! Thank you for the unit tests here!

PR #3477, opened 2018-02-09
helm install when resources already exist will fail

xchapter7x:
This is a PR to fix the issue described here: https://github.com/kubernetes/helm/issues/3319

- the intent here is to make the client smart enough to fail if a `Create` call is not able to complete successfully, due to a conflicting pre-existing resource.

I decided to implement this pre-existing resource check within the kubeclient. Reason being, if a `Create` call can not actually create the resource b/c the resource already exists, then one would expect the call to return an error.

This seemed more appropriate then implementing a check in the tiller package as part of the `InstallRelease` call, as tiller knows only about installing a release, and that it has a client interface upon which it can call `Create`. It shouldn't need to know anything about the expected behavior of the client Create call, and its handling of pre-existing resources.

The kubeclient Create implementation borrows the approach taken in the Update call on the kubeclient. Obvious difference being, update will skip pre-existing resources and create will now fail on pre-existing resources.

PR #6542, opened 2019-09-30
fix(action): Protects against current resource conflicts

### LINKED ISSUE #6407: [Helm 3] helm install --atomic deletes existing resources on failure
When `helm install --atomic` encounters an error, any resources that already existed in the cluster and conflict with resources which would be created by `helm install` are deleted during rollback. The `--atomic` argument should allow existing resources to be left intact instead of deleting them.

Output of `helm version`:

`version.BuildInfo{Version:"v3.0+unreleased", GitCommit:"d285097f58818533bb20eb8dff1cfb0a88e1878f", GitTreeState:"clean", GoVersion:"go1.12.9"}`

thomastaylor312:
Currently, if using the --atomic flag or deleting a release that failed due to an already existing
resource, Helm will delete those resources that aren't managed by it. This PR fixes the issue
by checking for pre-existing resources during install and upgrade. This is done as a validation
step so the release will not even be started if resources currently exist. This PR is inspired by
@xchapter7x's work in #3477.

This also fixes a small bug in upgrade where deletes fail if the resource was already deleted

Fixes #6407

### DIFF
--- a/pkg/action/upgrade.go
+++ b/pkg/action/upgrade.go
@@ -199,12 +199,6 @@ func (u *Upgrade) performUpgrade(originalRelease, upgradedRelease *release.Relea
 		return upgradedRelease, errors.Wrap(err, "unable to build kubernetes objects from new release manifest")
 	}
 
-	if u.DryRun {
-		u.cfg.Log("dry run for %s", upgradedRelease.Name)
-		upgradedRelease.Info.Description = "Dry run complete"
-		return upgradedRelease, nil
-	}
-
 	// Do a basic diff using gvk + name to figure out what new resources are being created so we can validate they don't already exist
 	existingResources := make(map[string]bool)
 	for _, r := range current {
@@ -220,6 +214,12 @@ func (u *Upgrade) performUpgrade(originalRelease, upgradedRelease *release.Relea
 
 	if err := existingResourceConflict(toBeCreated); err != nil {
 		return nil, errors.Wrap(err, "rendered manifests contain a new resource that already exists. Unable to continue with update")
+	}
+
+	if u.DryRun {
+		u.cfg.Log("dry run for %s", upgradedRelease.Name)
+		upgradedRelease.Info.Description = "Dry run complete"
+		return upgradedRelease, nil
 	}
 
 	u.cfg.Log("creating upgraded release for %s", upgradedRelease.Name)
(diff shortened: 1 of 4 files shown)


jlegrone (approved):
Thanks @thomastaylor312, I haven't tested locally but the code lgtm.

bacongobbler (approved):

pkg/kube/client.go
```diff
@@ -191,7 +191,7 @@ func (c *Client) Update(original, target ResourceList, force bool) (*Result, err
 	for _, info := range original.Difference(target) {
 		c.Log("Deleting %q in %s...", info.Name, info.Namespace)
 		res.Deleted = append(res.Deleted, info)
-		if err := deleteResource(info); err != nil {
+		if err := deleteResource(info); err != nil && !apierrors.IsNotFound(err) {
```
mattfarina: I might be missing something but if Helm expects a resource to be present so it tries to delete it but the resource isn't present so an error is thrown... should that be noted?

pkg/action/upgrade.go
```diff
@@ -211,6 +199,34 @@ func (u *Upgrade) performUpgrade(originalRelease, upgradedRelease *release.Relea
...
+	var toBeCreated kube.ResourceList
+	for _, r := range target {
+		if !existingResources[objectKey(r)] {
+			toBeCreated = append(toBeCreated, r)
+		}
+	}
+
+	if err := existingResourceConflict(toBeCreated); err != nil {
+		return nil, errors.Wrap(err, "rendered manifests contain a new resource that already exists. Unable to continue with update")
+	}
```
mattfarina: Should this check happen before returning on dry runs?

pkg/kube/client.go
thomastaylor312: If you are referring to logging it, I think so. I'll make the modification that it will always log the error, but only return an error if it isn't a not found error

pkg/action/upgrade.go
thomastaylor312: Actually yes. I had it that way at first, but then confused myself because I was working on `helm template` stuff yesterday, which _shouldn't_ reach out to k8s. I'll fix it

pkg/kube/client.go
thomastaylor312: Does my update address what you were looking for?

mattfarina (approved):

PR #6547, opened 2019-10-01
ref(pkg/chartutil): Simplify processDependencyConditions

SimonAlling:

It simplifies the code by removing a complicated, unnecessarily stateful, piece of control flow.

The tests seem to cover the affected parts well; the test suite failed when I had `return` instead of `break`. As far as I can tell, the commit doesn't change the semantics.


thomastaylor312 (approved):
@SimonAlling This is amazing! We've needed to clean up this code for a while and this was an extremely helpful refactor

PR #6550, opened 2019-10-01
fix(cmd): Fix all the outputs

### LINKED ISSUE #6238: Helm 3: Missing values from debug output
As discussed in Developer Call on 2019-08-15: I think a refactor of how releases are printed may have removed the `COMPUTED VALUES:` and `USER-SUPPLIED VALUES:` sections of the output. 

Helm 2: https://github.com/helm/helm/blob/f4a55985b6d101409583067ca69018cd014cbf28/cmd/helm/printer.go#L33-L35 (Is used [by template](https://github.com/helm/helm/blob/master/cmd/helm/template.go#L185) and [by install](https://github.com/helm/helm/blob/master/cmd/helm/install.go#L449))

Helm 3: 
Refactors printing responsibility to [pkg/action](https://github.com/helm/helm/blob/f73e3235c2ef8931274898588d65b7802b56d265/pkg/action/printer.go#L30), which doesn't print values anymore.

Observations:
`pkg/action PrintRelease()` is only passed a `Release` struct, which doesn't have the the context of `USER-SUPPLIED VALUES:` anymore. This is due to the coalesce which  merges the Chart values during the install run (see [pkg/action/install.go](https://github.com/helm/helm/blob/f73e3235c2ef8931274898588d65b7802b56d265/pkg/action/install.go#L138)). By the time the `Run` method of `Install` struct has run, it loses the original values supplied.

Suggested change:
Adding a [`Release`](https://github.com/helm/helm/blob/f73e3235c2ef8931274898588d65b7802b56d265/pkg/release/release.go#L22) struct field for `UserSuppliedValues` or potentially `OriginalConfig` to match the existing `Config` naming. This new field would house the original map rendered from [`values.Options{}.MergeValues()`](https://github.com/helm/helm/blob/f73e3235c2ef8931274898588d65b7802b56d265/pkg/cli/values/options.go#L41).

Open to more suggestions and discussion though!

* * *
## Issue Template Info
Output of `helm version`: 
```
version.BuildInfo{Version:"v3.0.0-alpha.2", GitCommit:"97e7461e41455e58d89b4d7d192fed5352001d44", GitTreeState:"clean", GoVersion:"go1.12.7"}
```

Output of `kubectl version`: 
```
Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.1", GitCommit:"4485c6f18cee9a5d3c3b4e523bd27972b1b53892", GitTreeState:"clean", BuildDate:"2019-07-18T09:18:22Z", GoVersion:"go1.12.5", Compiler:"gc", Platform:"linux/amd64"}
```

N/A

thomastaylor312:
There were two different methods and varying ways to output the status of a release.
This standardizes all of the outputs, but requires a breaking change. Output will
not perfectly match previous v3 output, and we had to break the printing function
in the `action` package, but now things are much more standardized.

Fixes #6238


pkg/action/printer.go
```diff
@@ -45,20 +49,48 @@ func PrintRelease(out io.Writer, rel *release.Release) {
...
-				fmt.Sprintf("Last Completed: %s", h.LastRun.CompletedAt),
+				fmt.Sprintf("Last Started:   %s", h.LastRun.StartedAt.Format(time.ANSIC)),
+				fmt.Sprintf("Last Completed: %s", h.LastRun.CompletedAt.Format(time.ANSIC)),
 				fmt.Sprintf("Phase:          %s", h.LastRun.Phase),
 			)
 		}
 	}
 
-	if strings.EqualFold(rel.Info.Description, "Dry run complete") {
+	if fullInfo {
```
adamreese: What kinds of information should be in `fullInfo` verses regular info.  If a value is added how would I make the decision?

pkg/action/printer.go
```diff
@@ -20,21 +20,25 @@ import (
...
 
+	"helm.sh/helm/pkg/chartutil"
 	"helm.sh/helm/pkg/release"
 )
 
-// PrintRelease prints info about a release
-func PrintRelease(out io.Writer, rel *release.Release) {
+// PrintRelease prints info about a release. If fullInfo is true, it will print
+// the user supplied values and the computed values used to render the chart
+func PrintRelease(out io.Writer, rel *release.Release, fullInfo bool) error {
```
adamreese: Breaking out the fullInfo logic into its own function would allow you to refactor out the boolean flag

pkg/action/printer.go
```diff
@@ -20,21 +20,25 @@ import (
...
+		return nil
 	}
 	fmt.Fprintf(out, "NAME: %s\n", rel.Name)
 	if !rel.Info.LastDeployed.IsZero() {
-		fmt.Fprintf(out, "LAST DEPLOYED: %s\n", rel.Info.LastDeployed)
+		fmt.Fprintf(out, "LAST DEPLOYED: %s\n", rel.Info.LastDeployed.Format(time.ANSIC))
 	}
 	fmt.Fprintf(out, "NAMESPACE: %s\n", rel.Namespace)
 	fmt.Fprintf(out, "STATUS: %s\n", rel.Info.Status.String())
+	fmt.Fprintf(out, "REVISION: %d\n", rel.Version)
```
adamreese: Sneaking this in too I see. :trollface:

cmd/helm/status.go
```diff
@@ -74,6 +74,7 @@ func newStatusCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 
 type statusPrinter struct {
 	release *release.Release
+	debug   bool
```
adamreese: is this used?

adamreese (changes requested):
Thanks for this.  It has bugged me for some time.

cmd/helm/status.go
thomastaylor312: Yep, it is passed into the print command for `fullInfo`

pkg/action/printer.go
thomastaylor312: The reason it is in here is for ordering purposes. I tried to think of a way that would preserve the current output formatting order with a separate function, but if we want to maintain the input the same, it needs to be in the middle. Basically, I could have this function take a function parameter that can insert things (like `func PrintRelease(out io.Writer, rel *release.Release, extras ExtraFunc...)`). But as this is specifically for printing a whole release, I feel like it makes sense here. Any ideas how I could make it cleaner?

pkg/action/printer.go
thomastaylor312: I _personally_ know what it is, but I think the best way around this is a comment. I'm thinking something like

> Anything printed in this block should be additional information about the release that is not necessarily needed for normal output (like the full manifest or all of the computed values). Basic information such as the name, revision, and release date should always be printed

pkg/action/printer.go
thomastaylor312: We can also break the current output ordering, which would probably be fine given how this is already breaking other things. That would also solve this issue

adamreese (approved):

thomastaylor312:
@adamreese and I talked and decided to do a bit heavier of a refactor here. I'll get to that at the beginning of next week

thomastaylor312:
Ok, I finished an additional refactor. Several breaking changes of note:

- `PrintRelease` has been entirely removed from the `action` package and now is a CLI tool only detail in the `cmd` package
- Output formats and the like have now been moved to `pkg/cli/output` as they didn't really belong in `action` anymore

I also added a cobra/pflag implementation for the output format in `cmd` so that we don't have to validate the format everywhere we use it

bacongobbler (approved):
LGTM! I like the change to the time output formatting as well. Much more human-readable.

adamreese (approved):
🎵 Time is on my side, yes it is 🎵

PR #6554, opened 2019-10-02
ref(pkg/chartutil): Dry up file and path names

SimonAlling:

It makes the code more robust and maintainable by reusing file and directory names from `create.go` instead of hardcoding them as string literals, and by using `filepath.Join` instead of literal slashes.

Incidentally, it also fixes an actual bug caused by the [WET](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself#DRY_vs_WET_solutions)ness it removes. Prior to this commit, the user would be presented with this error if `Chart.yaml` could not be read:

    cannot read Chart.Yaml in directory "my-chart"

(Note the capital `Y`.)

### DIFF
--- a/pkg/chartutil/chartfile.go
+++ b/pkg/chartutil/chartfile.go
@@ -59,14 +59,14 @@ func IsChartDir(dirName string) (bool, error) {
 		return false, errors.Errorf("%q is not a directory", dirName)
 	}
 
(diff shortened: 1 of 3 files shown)


hickeyma (approved):
LGTM, thanks @SimonAlling

PR #6555, opened 2019-10-02
Create file locking directory if it does not exist

### LINKED ISSUE #6521: v3: Can't add very first repo
The acceptance-tests started failing after #6492 (had to do a git bisect to figure that out).

The reason is that it is no longer possible to add the very first repo when starting from an empty environment as shown below:
```
> export XDG_CONFIG_HOME=/tmp/test/config
> mkdir -p ${XDG_CONFIG_HOME}

> $GOPATH/src/helm.sh/helm/bin/helm repo add gitlab https://charts.gitlab.io
Error: open /tmp/test/config/helm/repositories.yaml: no such file or directory

> /bin/rm -rf ${XDG_CONFIG_HOME}
```

The reason is that the locking of repository file expects the directory of the `repositories.yaml` file to exist.

I first thought of ignoring the error if it was a missing file error, but then I thought that the race the locking logic is fixing could also happen if two `helm repo add` try to create the `repositories.yaml` file concurrently.  I don't know how to fix that...

hickeyma:
As Helm v3 uses lazy creation for configuration, directories and files are not created until required. File locking when doing repo add was introduced in v2 and ported to v3 in #6492. It locks on the config directory where the repo file resides and therefore needs the directory to exist. This fix adds the directory if it doesn't exist.

Closes #6521 

This fixes the issue where user cannot add the first repository. Bug was introduced in #6492.

**If applicable**:

### DIFF
--- a/cmd/helm/repo_add.go
+++ b/cmd/helm/repo_add.go
@@ -22,6 +22,7 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"path/filepath"
 	"time"
 
 	"github.com/gofrs/flock"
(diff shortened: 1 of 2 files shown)


karuppiah7890:
Is it possible to add a test for this fix?

thomastaylor312:
I think I agree with @karuppiah7890 here. The change looks sane, but let's add it to a unit test for safety sake

hickeyma:
Sorry got distracted, meant to reply. I agree with @karuppiah7890. I will add a test case.

hickeyma:
Unit test added. Ready for review again.

thomastaylor312 (approved):
LGTM

PR #6563, opened 2019-10-03
v3: fix(cmd) have history use global output flag

marckhouzam:

PR #6504 add the `--output` flag to multiple commands in a centralized fashion.
The `helm history` command already had an `--output` flag.  This PR moves it to use the new global handling.

Also, it seems that for helm v3, the description of flags all start with a lowercase, so this PR changes the first letter of the description of the `--output` flag to a lowercase.


marckhouzam:
What do you think @thomastaylor312 ?

thomastaylor312 (approved):
Good catch on this one. I thought I had moved all of these over

PR #6571, opened 2019-10-03
v3: Performance: Batch kube API operations

cgetzen:
Signed-off-by: Charlie Getzen <charlie.getzen@procore.com>

This cuts down tiller's time creating resources. In our tests, we create ~180 resources in a release. Before this change it took approximately 120 seconds to create these resources. It now takes around 45 seconds.

v3 port of https://github.com/helm/helm/pull/5611

**If applicable**:

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -274,11 +274,7 @@ func perform(infos ResourceList, fn func(*resource.Info) error) error {
 	}
 
 	errs := make(chan error)
-	for _, info := range infos {
-		go func(i *resource.Info) {
-			errs <- fn(i)
-		}(info)
-	}
+	go batchPerform(infos, fn, errs)
 
 	for range infos {
 		err := <-errs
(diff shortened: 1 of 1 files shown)


bacongobbler:
Would you mind looking into the failing test run?

pkg/kube/client.go
```diff
@@ -273,14 +273,44 @@ func perform(infos ResourceList, fn func(*resource.Info) error) error {
...
+	for range infos {
+		err := <-errs
+		if err != nil {
 			return err
 		}
 	}
 	return nil
 }
 
+func batchPerform(infos Result, fn ResourceActorFunc, errs chan<- error) {
```
thomastaylor312: @cgetzen As an FYI, `Result` has been renamed to `ResourceList`, which is what is causing your error

thomastaylor312:
@bacongobbler I assume we can wait on this until #5611 is finished so @cgetzen has the change to address all the comments there and port them here.

@cgetzen Thanks for opening this port!

pkg/kube/client.go
cgetzen: better name! thanks

thomastaylor312 (approved):
LGTM. Thanks for the port!

PR #6573, opened 2019-10-03
fix(cmd): Removes subprocess test causing flakes

thomastaylor312:
We are pretty sure this test is causing some of the flakes, so I removed the test as it is fairly well covered with the concurrent process test


bacongobbler:
Nope... Looks like a more extensive inspection may be in order.

thomastaylor312:
Sigh...yep

PR #6334, opened 2019-09-01
Upgrade golang to 1.13

ialidzhikov:
**What this PR does / why we need it**:
Upgrade golang to 1.12.9.

**Special notes for your reviewer**:

**If applicable**:
- [ ] this PR contains documentation
- [ ] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

PR #6577, opened 2019-10-03
upgrade Go to 1.13.1

bacongobbler:
Upgrades Go to 1.13.1. Also fixes up the generated package comment error mentioned in https://github.com/helm/helm/pull/6334#issuecomment-531805500.

supersedes #6334


thomastaylor312 (approved):

karuppiah7890:
@bacongobbler If we use `golang:1.13`, we'll always get the latest version in the `1.13` series right? I did that for Helm 3 [here](https://github.com/helm/helm/blob/dev-v3/.circleci/config.yml#L8) but there we use `circleci/golang` image and they also give latest in `1.13` series. I just checked

karuppiah7890:
If we intentionally don't want the latest always and want to know what version is being used to build, then we need to change it in Helm 3 too. As currently it always gets latest

PR #3422, opened 2018-01-30
fix(helm): Accept dependency in requirements.yaml from charts directory

splisson:
This change allow the use of charts from the charts sub-directory of the current chart in requirements.yaml. It checks the version. It leaves the chart in place.

Closes #3221

PR #6578, opened 2019-10-03
fix(helm): Accept dependency in requirements.yaml from charts directory

### LINKED ISSUE #3221: Helm should allow easier "management" of child charts placed manually in `charts`
I have a PR (kubernetes/charts#1719) that uses two subcharts, with a `requirements.yaml` to make it easy to choose whether one or the other is not deployed.  Currently the acceptance test is failing because it does `helm dep up` before trying to install the chart, and with no `repository` line in requirements.yaml `helm dep up` fails.  Declaring the "local" repository causes `helm dep up` to fail a la #2887.  I could eliminate the use of `requirements.yaml` but then I'd have to wrap every template in an `{{- if foo.enabled }}` block, which seems inelegant at best.

It may seem nonsensical to have the source repo be the `charts` directory but in my case it's the direct result of these constraints:

* using subcharts
* wanting to allow subcharts to be turned on or off independently
* wanting that to be easy to maintain

I propose one of the following two fixes:

* `helm dep up` ignores dependencies that have no repo declared instead of throwing an error
* if `helm dep up` sees a dependency source URL that resolves to the `charts` directory of the installing chart, it does not try to download the dependency but checks that it is already present and meets any declared constraints like version requirements, erroring only if those checks fail.

bacongobbler:
supersedes #3422
closes #3221

### DIFF
--- a/pkg/downloader/manager.go
+++ b/pkg/downloader/manager.go
@@ -218,10 +218,11 @@ func (m *Manager) downloadAll(deps []*chartutil.Dependency) error {
 	for _, dep := range deps {
 		// No repository means the chart is in charts directory
 		if dep.Repository == "" {
+			fmt.Fprintf(m.Out, "Dependency %s did not declare a repository. Assuming it exists in the charts directory\n", dep.Name)
 			chartPath := filepath.Join(tmpPath, dep.Name)
 			ch, err := chartutil.LoadDir(chartPath)
 			if err != nil {
-				return fmt.Errorf("Unable to load local chart: %v", err)
+				return fmt.Errorf("Unable to load chart: %v", err)
 			}
 
 			constraint, err := semver.NewConstraint(dep.Version)
@@ -231,11 +232,11 @@ func (m *Manager) downloadAll(deps []*chartutil.Dependency) error {
 
 			v, err := semver.NewVersion(ch.Metadata.Version)
 			if err != nil {
-				return fmt.Errorf("Chart %s has an invalid version format: %s", dep.Name, err)
+				return fmt.Errorf("Invalid version %s for dependency %s: %s", dep.Version, dep.Name, err)
 			}
 
 			if !constraint.Check(v) {
-				saveError = fmt.Errorf("Can't get a valid version for dependency %s", dep.Name)
+				saveError = fmt.Errorf("Dependency %s at version %s does not satisfy the constraint %s", dep.Name, ch.Metadata.Version, dep.Version)
 				break
 			}
 			continue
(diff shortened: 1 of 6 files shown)


pkg/resolver/resolver_test.go
```diff
@@ -90,6 +90,33 @@ func TestResolve(t *testing.T) {
...
+				Dependencies: []*chartutil.Dependency{
+					{Name: "localdependency", Repository: "", Version: "0.1.0"},
+				},
+			},
+		},
+		{
+			name: "repo from valid path under charts path",
+			req: &chartutil.Requirements{
+				Dependencies: []*chartutil.Dependency{
+					{Name: "inexistentdependency", Repository: "", Version: "0.1.0"},
```
thomastaylor312: Nit: `nonexistentdependency` instead of `inexistentdependency`

pkg/downloader/manager.go
```diff
@@ -216,6 +216,30 @@ func (m *Manager) downloadAll(deps []*chartutil.Dependency) error {
 	fmt.Fprintf(m.Out, "Saving %d charts\n", len(deps))
 	var saveError error
 	for _, dep := range deps {
+		// No repository means the chart is in charts directory
+		if dep.Repository == "" {
```
thomastaylor312: What are your thoughts on using a keyword like `local` (which I know is already used in v2, but technically by doing `@local`, plus very few people use the local repo). Assuming an empty value means local could lead to weird error messages for people who may have just forgot and `local` is more expressive

thomastaylor312: Also, we should add documentation of this new feature to the help text for `helm dep`

thomastaylor312 (changes requested):
Code and tests look great, wanted to have a chat about possibly using a specific keyword to indicate a repo is local.

Also one small ask for some added documentation to the help text for `helm dep`

pkg/downloader/manager.go
bacongobbler: You raise a good point here. I'll take a deeper dive into the code and see if there's an alternative approach we should take instead.

bacongobbler: After further review, I think it's probably fine to accept an empty repository field as on-disk. However, I think we should inform the user that Helm is attempting to load the chart from disk. I added this in the latest commit. Does that seem fair to you?

I'm afraid of adding additional "one-off" syntax to an already unruly function that desperately needs a rework

thomastaylor312: Sounds like a good compromise

thomastaylor312 (approved):

hickeyma:
Working on the port for this

hickeyma:
Port to v3: #6611

PR #6580, opened 2019-10-04
feat(comp): Dynamic completion for --output flag

marckhouzam:

This PR adds shell completion for the `--output` flag.  For example a user could now do:
```
helm repo list --output <TAB>
```
and will then be given the choices:
```
json   table  yaml
```
This works for the `--output` flag of any command that supports it.

I don't like that the three values `json table yaml` are hard-coded in the completion logic.  If a new format is ever added, it is likely that the completion code will be forgotten and not updated.  I would have preferred to create the list of choices by looping through the "enum" of formats.  However, I don't believe this is feasible with the current implementation of the output formats which are unrelated string contants.  Doing something with `iota` would be possible, but would require multiple changes in the output format code.

What I did instead is that the acceptance test will generate the expected completion (`json table yaml`  for now) output based on the help text of the `--output` flag.  I feel this help text has a better chance of being updated than the completion code.  So, if an extra format is added, the acceptance tests will fail if the completion code is not matching the help text for the `--output` flag.

**If applicable**:
- this PR has been tested for backwards compatibility through the acceptance-testing repo
- new tests for this feature have been prepared in https://github.com/helm/acceptance-testing/pull/51

### DIFF
--- a/cmd/helm/flags.go
+++ b/cmd/helm/flags.go
@@ -52,6 +52,8 @@ func addChartPathOptionsFlags(f *pflag.FlagSet, c *action.ChartPathOptions) {
 // value to the given format pointer
 func bindOutputFlag(cmd *cobra.Command, varRef *output.Format) {
 	cmd.Flags().VarP(newOutputValue(output.Table, varRef), outputFlag, "o", fmt.Sprintf("prints the output in the specified format. Allowed values: %s, %s, %s", output.Table, output.JSON, output.YAML))
+	// Setup shell completion for the flag
+	cmd.MarkFlagCustom(outputFlag, "__helm_output_options")
 }
 
 type outputValue output.Format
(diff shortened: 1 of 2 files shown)


hickeyma (commented):
Thanks for this @marckhouzam. Can you squash your commits as it would help reviewing?

bacongobbler (approved):

hickeyma:
Can we wait to merge until commits are squashed?

bacongobbler:
Yes absolutely. Alternatively, github allows you to "squash and merge" pull requests if you wish.

image

hickeyma:
sure do, just wanted to test against the the branch manually and having problems with the commits

marckhouzam:
Squashed.

bacongobbler:
thanks @marckhouzam! I'll let @hickeyma merge this one. :)

hickeyma (approved):
LGTM, thanks @marckhouzam

PR #6582, opened 2019-10-04
Use same output format for hooks as for manifest

jonasrutishauser:
Use the same output format in `get hooks` as in `get manifest`.
Additionally it adds a missing line break before the yaml separator.

I didn't fix the hooks in the `get` output as this will be done in #6550


thomastaylor312 (approved):
Thanks for fixing this!

Issue #6585, opened 2019-10-05
Symlinks to directories get duplicated in chart package


mauvilsa:
Output of `helm version`:
```
Client: &version.Version{SemVer:"v2.14.3", GitCommit:"0e7f3b6637f7af8fcfddb3d2941fcc7cbebb0085", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.14.3", GitCommit:"0e7f3b6637f7af8fcfddb3d2941fcc7cbebb0085", GitTreeState:"clean"}
```

When creating a packaged chart, symlinks to directories get included in the package twice, once in the proper location of the symlink and again as the full path of the resolved symlink.

I have prepared an example to reproduce the issue. It is the same `docs/examples/alpine` in the helm repo, but moving out the templates as if these were intended to be shared across multiple charts. Steps to reproduce:

1. Download [helm_symlinks_issue.tar.gz](https://github.com/helm/helm/files/3693776/helm_symlinks_issue.tar.gz) to your /tmp directory and extract it there.
```
$ cd /tmp
$ tar xzf helm_symlinks_issue.tgz
```

2. Go into the extracted directory and build the helm package.
```
$ cd /tmp/helm_symlinks_issue
$ helm package alpine
Successfully packaged chart and saved it to: /tmp/helm_symlinks_issue/alpine-0.1.0.tgz
```

3. List the files in the created package.
```
$ tar tzf alpine-0.1.0.tgz
alpine/Chart.yaml
alpine/values.yaml
alpine/templates/_helpers.tpl
alpine/templates/alpine-pod.yaml
alpine/README.md
alpine/tmp/helm_symlinks_issue/shared_templates/_helpers.tpl
alpine/tmp/helm_symlinks_issue/shared_templates/alpine-pod.yaml
```

The issue can be seen in the last two listed files.

mauvilsa:
Created pull request https://github.com/helm/helm/pull/6586 that I hoped resolved the issue. Unfortunately the circleci tests failed.

PR #6586, opened 2019-10-05
symwalk process symlinks only with its un-resolved path

mauvilsa:
Pull request intended to resolve issue https://github.com/helm/helm/issues/6585

Fixes file `pkg/sympath/walk.go` so that when a symlink is encountered, only the info of the path is replaced so that it gets walked in the right location, but don't walk the resolved symlink.


bacongobbler:
Hi @mauvilsa, it looks like the unit tests failed fantastically with a panic dumptrace. Would you mind looking into this? Additionally, would you mind adding a test case that demonstrates the issue you are trying to resolve? Thank you.

mauvilsa:
Hi @bacongobbler.

Regarding the test case, yes I can try to create one. Though I would appreciate some ideas on how it could be done for this issue. It is the first time I look at the helm source code and even the first time I look at go source code.

Regarding the fail in circleci I have looked into it a bit and I am tempted to say that it does not relate to the code changed in this pull request. It failed in `pkg/kube` which does use `pkg/sympath`. I tried to run the tests locally just like it is done in circleci. Initially it didn't fail, but by enabling race detection it does fail in the same place. I also ran it with the current helm master and it failed the same way. So it might be because of a problem with a race condition in `kube` and it was only a coincidence that it failed for this pull request.

More specifically what I did was:

1. Pull the `golang:1.13.1` docker image and start a container mounting the go directory to /go and changing to the /go/src/k8s.io/helm directory.
2. Run `.circleci/bootstrap.sh`.
3. Run `make build`.
4. Run `HELM_HOME=/no/such/dir go test  -run . ./cmd/... ./pkg/... -race`

bacongobbler:
https://golang.org/pkg/testing/
https://helm.sh/docs/developers/
https://github.com/helm/helm/blob/master/pkg/sympath/walk_test.go

it's possible the failure might've occurred due to https://github.com/helm/helm/issues/6617. I restarted the tests to verify.

bacongobbler:
ping, any progress on this one?

mauvilsa:
> ping. Any progress on this one?

Unfortunately not. I have been busy with other stuff. But I will get to it as soon as possible.

bacongobbler:
We just switched branches over. Helm 2 development is on the `dev-v2` branch, and Helm 3 development is on the `master` branch. Would you mind re-opening this PR against the correct branch once this is ready for review again? Thanks.

PR #6590, opened 2019-10-07
Update Ingress API group in scaffold chart

hickeyma:
Change ingress manifest generated by scaffold chart to have the newest API group.

`networking.k8s.io/v1beta1` API, serving Ingress since v1.14.  `extensions/v1beta1` API group will be deprecated in K8s v1.20. (https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/)

**If applicable**:

### DIFF
--- a/pkg/chartutil/create.go
+++ b/pkg/chartutil/create.go
@@ -151,7 +151,11 @@ const defaultIgnore = `# Patterns to ignore when building packages.
 const defaultIngress = `{{- if .Values.ingress.enabled -}}
 {{- $fullName := include "<CHARTNAME>.fullname" . -}}
 {{- $svcPort := .Values.service.port -}}
+{{- if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion -}}
+apiVersion: networking.k8s.io/v1beta1
+{{- else -}}
 apiVersion: extensions/v1beta1
+{{- end }}
 kind: Ingress
 metadata:
   name: {{ $fullName }}


pkg/chartutil/create.go
```diff
@@ -151,7 +151,7 @@ const defaultIgnore = `# Patterns to ignore when building packages.
 const defaultIngress = `{{- if .Values.ingress.enabled -}}
 {{- $fullName := include "<CHARTNAME>.fullname" . -}}
 {{- $svcPort := .Values.service.port -}}
-apiVersion: extensions/v1beta1
+apiVersion: networking.k8s.io/v1beta1
```
mattfarina: Could we have some logic to handle both the old and new `apiVersion`? This way create will work across k8s versions. #6448 added it to dev-v3 but I don't see it on v2 or even in v3 anymore.

hickeyma: I would have been of the same thinking but there was a good [response](https://github.com/helm/helm/pull/6490#discussion_r331752270) from @unguiculus on this in #6490. I am easy either way but why are we putting this in place when we support k8s n-1 anyways?

mattfarina: The charts repository has had an n-1 support philosophy. In practice it hasn't really followed that lately. Does helm itself have something? I ask because many of the public clouds have Kubernetes n-2 as their latest stable. That includes EKS, Google Cloud, and Azure. The Google Cloud current default is 1.13. `helm create` not being able to create charts that run in the latest stable releases from the largest public clouds seems like a problem. Is there a good case for n-1 support?

hickeyma: Ok, I put back the check in place for k8s < 1.14.

hickeyma:
@mattfarina @bacongobbler ready to review again.

bacongobbler (approved):

bacongobbler:
Hey @hickeyma! I know you're probably sleeping at this time. Given how small and concise this PR's intent is, it seems fairly safe to merge, so I am going to merge this one on your behalf in order to kick off the 2.15.0-rc.1 release dance. If there are things we still need to address further, please let us know.

Thank you for your contribution!

hickeyma:
Thanks @bacongobbler for getting this in. It was good to go.

PR #5392, opened 2019-03-04
Allow customization of API Versions when using Helm Template

bnookala:
Signed-off-by: Bhargav Nookala <nooknb@gmail.com>

**What this PR does / why we need it**:

Resolves https://github.com/helm/helm/issues/3377 for `helm template`. 
This PR adds `--api-version` (or `-a`) to add custom API versions to Capabilites.APIVersion, allowing a user to correctly render templates that depend on certain API versions.

**Special notes for your reviewer**:

**If applicable**:
- [x] this PR contains unit tests
- [x] this PR contains documentation

PR #6594, opened 2019-10-07
feat(template): Ports api-versions flag to v3

thomastaylor312:
This is a port of #5392. It also takes care of the small chore to update the default k8s
version to 1.16, which is the latest supported version.


thomastaylor312:
Please note that the number of files changed was due to updating the golden output to match the new k8s version

hickeyma (approved):
LGTM, thanks @thomastaylor312

bacongobbler (approved):

Issue #6606, opened 2019-10-08
Drop golang dependency on k8s.io/kubernetes


howardjohn:
Helm currently imports k8s.io/kubernetes in a few places. At time of writing:
```
pkg/kube/wait.go
32:     deploymentutil "k8s.io/kubernetes/pkg/controller/deployment/util"

pkg/kube/client.go
58:     "k8s.io/kubernetes/pkg/api/legacyscheme"
59:     "k8s.io/kubernetes/pkg/apis/core"
60:     "k8s.io/kubernetes/pkg/kubectl/cmd/get"
61:     cmdutil "k8s.io/kubernetes/pkg/kubectl/cmd/util"
62:     "k8s.io/kubernetes/pkg/kubectl/validation"

pkg/kube/client_test.go
37:     cmdtesting "k8s.io/kubernetes/pkg/kubectl/cmd/testing"
38:     kubectlscheme "k8s.io/kubernetes/pkg/kubectl/scheme
```

Kubernetes intentionally ships a broken `go.mod` file to discourage people from depending on k8s.io/kubernetes (see https://github.com/kubernetes/kubernetes/issues/79384). As a result, in order to depend on helm, we must go through major hacks to get go modules working.

Is it feasible to drop this dependency?

https://github.com/kubernetes/kubernetes/issues/79384#issuecomment-538740756 indicates it may be possible to move the controller library out of k8s.io/kubernetes, and I think the `api` and `kubectl` ones can already use `k8s.io/api` and `k8s.io/kubectl`

bacongobbler:
Looking at those three files in the `dev-v3` branch, Helm 3 only has one reference to `k8s.io/kubernetes` at this time, and it's in wait.go:

https://github.com/helm/helm/blob/c00890cae1cf756ba0bc6eac82385c6cffeb5001/pkg/kube/wait.go#L39

I'd suggest starting there, as it looks much more feasible to drop the last dependency on k8s.io/kubernetes for Helm 3 than it is for Helm 2.

scbizu:
Hi there , I will have my dependency on helm with `helm.sh/helm/v3/pkg/chart` , but when I go get this package, it says 
```
go get: helm.sh/helm/v3@v3.0.0-beta.5 requires
        k8s.io/api@v0.0.0: reading k8s.io/api/go.mod at revision v0.0.0: unknown revision v0.0.0
```

my Go version is go1.13, I understand that I can do some go mod replacement at my `go.mod`,but are there any other solutions ?

[...]

PR #5038, opened 2018-12-10
Fix rename for helm dep upgrade. For issue #2998

ghost:
helm dependency upgrade wasn't working on certain file system because it assumes that os.rename is available. Since rename isn't available for subfolders in docker containers, I ripped a fallback rename strategy from [dep](https://github.com/golang/dep/blob/5b1fe9e6d89372487d0aac78d9c5362a517857e7/internal/fs/fs.go#L103-L118) that works in docker. I wasn't able to reference this package as a dependency because it is an internal package in dep.

This is my first submission to the open source community, so please let me know if committing this fs package from golang/dep is the wrong approach, or if something in their license prohibits this.

https://github.com/golang/dep

PR #6609, opened 2019-10-08
Drop dependency on k8s.io/kubernetes

howardjohn:
https://github.com/helm/helm/issues/6606

Depending on k8s.io/kubernetes is not recommended by Kubernetes, and
forces dependencies of Helm to also depend on them. We are only using
this dependency in one relatively isolated occurance, which can be
easily copied over rather than depending on the entire Kubernetes.

Copying this code is not very desirable, so if we don't want to have
this duplication we can at least use this PR as a PoC and see if we can
get Kubernetes to publish the controller package as a separate Go module
(see
https://github.com/kubernetes/kubernetes/issues/79384#issuecomment-538740756)

Closes https://github.com/helm/helm/issues/6606

**If applicable**:


bacongobbler:
This doesn't seem legally sound... Unless I'm mistaken, deploymentutil.go is copied from Kubernetes code, so we cannot claim copyright ownership of the file. This seems very similar to the same situation described in #5038.

Similar to https://github.com/helm/helm/pull/5038#issuecomment-474133510, can you please

1. move `/pkg/kube/deploymentutil.go` to a new directory: `/third_party/k8s.io/kubernetes/deployment/util/deployment_util.go` or something indicative that the code came from Kubernetes
2. Please place the full Apache license header comments from the source file. The end result should look identical to [the code in Kubernetes](https://github.com/kubernetes/kubernetes/blob/e856613dd5bb00bcfaca6974431151b5c06cbed5/pkg/controller/deployment/util/deployment_util.go)
3. Add a comment stating where [the original source code exists](https://github.com/kubernetes/kubernetes/blob/e856613dd5bb00bcfaca6974431151b5c06cbed5/pkg/controller/deployment/util/deployment_util.go)
4. If there are changes made to the original deployment_util.go file, please also add comments stating what changed, and why.

This way we have all our legal bases covered as advised from the CNCF legal team.

Thanks!

pkg/chartutil/expand.go
```diff
@@ -45,9 +45,6 @@ func Expand(dir string, r io.Reader) error {
 			if err := yaml.Unmarshal(file.Data, ch); err != nil {
 				return errors.Wrap(err, "cannot load Chart.yaml")
 			}
-			if err != nil {
```
bacongobbler: would you mind reverting this? Seems like this isn't relevant to the change at hand.

pkg/kube/deploymentutil.go
```diff
@@ -0,0 +1,176 @@
+/*
+Copyright The Helm Authors.
```
bacongobbler: Here's the legal issue I'm referring to.

bacongobbler (changes requested):

howardjohn:
> This doesn't seem legally sound... Unless I'm mistaken, deploymentutil.go is copied from Kubernetes code, so we cannot claim copyright ownership of the file. This seems very similar to the same situation described in #5038.

I was wondering about this. Thanks for the guidance on how to make this work, I'll send a fixup soon

pkg/chartutil/expand.go
howardjohn: Hmm looks like it was a lint error, I must have a different version locally. I'll revert

jlegrone (commented):
We should probably avoid making this part of Helm's public API surface. Can we move `pkg/kube` to `internal/kube`? Ref: https://github.com/golang-standards/project-layout/tree/master/internal

Thanks for taking the initiative on this, I think this change will render some hacks I've had to do in the past obsolete!

bacongobbler:
`pkg/kube` has received several enhancements that make it useful to others trying to create/update/delete/watch objects similar to Helm, especially after #6085. Is there a particular reason it should it be made internal?

jlegrone:
Oh right, I forgot that `pkg/kube` already existed. I do think we should consider keeping the methods in `pkg/kube/deploymentutil.go` private, or move the file to an internal package.

The rationale is that we can choose to make these new functions public at any time, but while they're internal we can always refactor to depend on `kubernetes/kubernetes` if this functionality becomes consumable in the future, without introducing a breaking change.

bacongobbler:
Good point.

@howardjohn, instead of `/third_party/k8s.io/kubernetes/deployment/util/deployment_util.go`, can you please move it to `/internal/third_party/k8s.io/kubernetes/deployment/util/deployment_util.go`? Much appreciated.

howardjohn:
@bacongobbler I replaced the Helm license with the Kubernetes license and the license check fails. I added an exception to filter out `third_party`. Let me know if I should do something else instead, like adding both licenses

bacongobbler (approved):
LGTM! Thanks for doing this. Holding off on merging until another maintainer approves this as well, given the size and legal issues surrounding this.

jlegrone (approved):

howardjohn:
Is there anything else needed to get this merged?

bacongobbler:
@jlegrone is not a maintainer. Needs another LGTM from a project maintainer of Helm. I'll ping someone today to take a look at this.

howardjohn:
Sounds good, thanks!

hickeyma (approved):
I am not happy with the way this has to be done because it is quite dirty to have to copy over a file in this way. However, if it is the only way to work around it in the circumstance then it is what it is.

Thanks for the work on this @howardjohn. Good tracking down.

sslavic:
Please consider applying equivalent change to `release-2.15` too. That would make migration easier, decouple steps - adopting Go modules and switch to helm v3 once available.

bacongobbler:
it won't be possible without breaking backwards compatibility, unfortunately. Many contributors have tried, which is why we had to target this for Helm 3.

sslavic:
I guess answer is same for release-2.16, correct?
Curious about backward compatibility being broken, this change internalizes a testing support package and removes no longer used transitive dependency on k8s.io/kubernets; if someone using helm as library used also directly k8s.io/kubernetes without specifying it as direct dependency, depending on helm to bring it in transitively, they should be able to easily fix their build tool dependency configuration and declare direct dependency to k8s.io/kubernetes - what am I missing here?

bacongobbler:
You're right that this change only makes one change to the public API for Helm 3. In Helm 2, the changes are a little [more involved](https://github.com/helm/helm/blob/0ca95341c19fcaf9cecdf50caef1693568835a84/pkg/kube/client.go#L60-L61), which makes it incredibly difficult to remove that dependency without introducing breaking changes.

Hope this helps!

PR #6622, opened 2019-10-09
fix(pkg/kube): validate with OpenAPI on install

adamreese:


thomastaylor312 (approved):

bacongobbler:
the only other spot I was wondering about is whether test hooks should be validated as well, since they are not part of the execution hooks or the rendered manifest. What do you think, @adamreese?

bacongobbler:
nvm, after looking at the code, that's just test hooks to be deleted. LGTM!

bacongobbler (approved):

PR #6632, opened 2019-10-10
v3: Propagate --kube-context, --kubeconfig and --namespace values to plugins

### LINKED ISSUE #6631: v3: --kube-context, --kubeconfig and --namespace values are not passed to plugins
## Problem

While upgrading [helm-diff](https://github.com/databus23/helm-diff/pull/151#issuecomment-537679821) to work with helm v3, we noticed that a few settings that formerly passed via the local proxy to Tiller now need to be passed explicitly to plugins:

- --kubeconfig
- --kube-context
- --namespace

// Additional discussion in helm/helm: https://github.com/helm/helm/pull/6508#discussion_r330893957

This is due to that some plugin like helm-diff does need to access release objects normally managed by helm.

More concretely, as of today when a plugin like helm-diff is invoked via `helmv3 --kubeconfig mykubeconfig --kube-context mycontext --namespace myns diff upgrade myrelease mychart`, the plugin needs all the three settings in order to read the release object for `myrelease`, but there's no way to access those from within the plugin.

## Proposed solution

I believe these 3 settings should somehow be propagated to plugins. I'd suggest using envvars as it seems like the primary way to pass contextual settings to plugins today: https://github.com/helm/helm/blob/9bc7934f350233fa72a11d2d29065aa78ab62792/pkg/plugin/plugin.go#L218-L225

---
Output of `helm version`: helm 3.0.0-beta.1~4
Cloud Provider/Platform (AKS, GKE, Minikube etc.): Any

mumoshu:

Closes #6631

**If applicable**:

### DIFF
--- a/pkg/cli/environment.go
+++ b/pkg/cli/environment.go
@@ -84,12 +84,20 @@ func envOr(name, def string) string {
 }
 
 func (s *EnvSettings) EnvVars() map[string]string {
-	return map[string]string{
+	envvars := map[string]string{
 		"HELM_BIN":               os.Args[0],
 		"HELM_DEBUG":             fmt.Sprint(s.Debug),
 		"HELM_PLUGINS":           s.PluginsDirectory,
 		"HELM_REGISTRY_CONFIG":   s.RegistryConfig,
 		"HELM_REPOSITORY_CACHE":  s.RepositoryCache,
 		"HELM_REPOSITORY_CONFIG": s.RepositoryConfig,
+		"HELM_NAMESPACE":         s.Namespace,
+		"HELM_KUBECONTEXT":       s.KubeContext,
 	}
+
+	if s.KubeConfig != "" {
+		envvars["KUBECONFIG"] = s.KubeConfig
+	}
+
+	return envvars
 }


pkg/cli/environment.go
```diff
@@ -91,5 +91,8 @@ func (s *EnvSettings) EnvVars() map[string]string {
 		"HELM_REGISTRY_CONFIG":   s.RegistryConfig,
 		"HELM_REPOSITORY_CACHE":  s.RepositoryCache,
 		"HELM_REPOSITORY_CONFIG": s.RepositoryConfig,
+		"HELM_NAMESPACE":         s.Namespace,
```
mumoshu: The name `HELM_NAMESPACE` is borrowed form an another PR from @adamreese https://github.com/helm/helm/pull/6508/files#r330893957

Other two are named such so that they become consistent with all the other envvars.

pkg/cli/environment.go
```diff
@@ -91,5 +91,8 @@ func (s *EnvSettings) EnvVars() map[string]string {
...
+		"HELM_KUBECONTEXT":       s.KubeContext,
+		"HELM_KUBECONFIG":        s.KubeConfig,
```
databus23: How about exposing this as `KUBECONFIG` for plugins? That way invocations of`helm` (or `kubectl`) from a plug-in would automatically pick in the correct config. This would have to be conditional so that if `—kubeconfig` is not set we don‘t clobber a value set in the `KUBECONFIG`  env var

mumoshu: @databus23 Sounds good! Addressed in the last commit.

bacongobbler (approved):

PR #6634, opened 2019-10-10
fix(v3): Bring back the missing `helm template [-x|--execute] PATH/TO/SINGLE/TEMPLATE`

### LINKED ISSUE #6633: v3: `helm template` is missing the flag to render only selected templates
In helm v2 we had `helm template mychart-1.2.3.tgz -x templates/your.yaml` that can be used to render only selected template files in the chart.

If you run `helm template --help`, this is featured in the beginning of the output like:

```
To render just one template in a chart, use '-x':

	$ helm template mychart -x templates/deployment.yaml

Usage:
  helm template [flags] CHART

Flags:
  -x, --execute stringArray      only execute the given templates
```

This feature isn't announced for removal so I'm assuming this is a regression in v3. Can we just reimplement it for v3?

If we are to reimplement this for helm v3, a concern is that `helm template` now supports templating the remote chart directly, without firstly `helm fetch`ing the chart. The v3 version of the feature should just work regardless of the chart is local or remote.

---

Output of `helm version`: helm-3.0.0-beta.1~4

Cloud Provider/Platform (AKS, GKE, Minikube etc.): Any

mumoshu:

Closes #6633

**If applicable**:

### DIFF
--- a/cmd/helm/template.go
+++ b/cmd/helm/template.go
@@ -17,9 +17,14 @@ limitations under the License.
 package main
 
 import (
+	"bytes"
 	"fmt"
 	"io"
+	"path/filepath"
+	"regexp"
 	"strings"
+
+	"helm.sh/helm/v3/pkg/releaseutil"
 
 	"github.com/spf13/cobra"
 
@@ -42,6 +47,7 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 	client := action.NewInstall(cfg)
 	valueOpts := &values.Options{}
 	var extraAPIs []string
+	var renderFiles []string
 
 	cmd := &cobra.Command{
 		Use:   "template [NAME] [CHART]",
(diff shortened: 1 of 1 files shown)


mumoshu:
I've manually verified this to work as expected by running it like `helm template foo stable/mysql -x templates/tests/test.yaml`.

Perhaps this worth being extracted out into a separate function and be unit-tested? Or can I do it in another PR so that this regression is fixed soon anyway?

cmd/helm/template.go
```diff
@@ -58,11 +64,51 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
...
+					}
+					if missing {
+						return fmt.Errorf("could not find template %s in chart", f)
+					}
+					for _, m := range manifestsToRender {
+						fmt.Fprintf(out, "---\n%s\n", m)
+					}
 				}
+			} else {
+				fmt.Fprintf(out, "%s", manifests.String())
```
mumoshu: For anyone familiar with the helm-v2 code, this work is very similar to the implementation in v2. I just adapted it to work well with v3.

https://github.com/helm/helm/blob/66e0c0b44da1a7b604c592c87de88f545ce737bb/cmd/helm/template.go#L197-L231

cmd/helm/template.go
```diff
@@ -71,6 +117,7 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 
 	f := cmd.Flags()
 	addInstallFlags(f, client, valueOpts)
+	f.StringArrayVarP(&renderFiles, "execute", "x", []string{}, "only execute the given templates")
```
bacongobbler: Can we change the name of the flag to something more representative of the actual behaviour? In Helm 2, I believe we were choosing which templates to pass through the rendering engine. In Helm 3, we are rendering all manifests, but we're selectively choosing which files to display on stdout.

What about `--show-only`?

bacongobbler (commented):
This looks like the right path forward. I have one comment about the flag name but the code here looks good!

bacongobbler:
throwing this on the beta.5 milestone as it restores a feature that once existed in Helm 2.

cmd/helm/template.go
mumoshu: After reading your comment at https://github.com/helm/helm/issues/6633#issuecomment-540632323 - it makes sense!

Added an additional commit for the change: 74a2adf

bacongobbler (approved):

drewwells:
Did this get removed again?
```
helm template -x                                                                   
Error: unknown shorthand flag: 'x' in -x
```

bacongobbler:
See https://github.com/helm/helm/pull/6634#discussion_r333647125

PR #5177, opened 2019-01-16
Allow lint command to parse pre-release charts

adshmh:
**What this PR does / why we need it**:
Allow lint command to parse charts with pre-release version. Closes #5150

**Special notes for your reviewer**:

**If applicable**:
- [ ] this PR contains documentation
- [x] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

PR #6640, opened 2019-10-10
feat(lint): Ports v2 functionality for linting pre-release charts

thomastaylor312:
This is a port of #5177


bacongobbler (approved):

pkg/action/lint.go
```diff
@@ -94,12 +94,15 @@ func lintChart(path string, vals map[string]interface{}, namespace string, stric
...
 		}
 
-		lastHyphenIndex := strings.LastIndex(filepath.Base(path), "-")
-		if lastHyphenIndex <= 0 {
-			return linter, errors.Errorf("unable to parse chart archive %q, missing '-'", filepath.Base(path))
+		files, err := ioutil.ReadDir(tempDir)
+		if err != nil {
+			return linter, errors.Wrapf(err, "unable to read temporary output directory %s", tempDir)
+		}
+		if !files[0].IsDir() {
```
jdolitsky: Should length  be checked here?

jdolitsky (changes requested):

pkg/action/lint.go
thomastaylor312: It was not originally, but it probably should. Fixing now

thomastaylor312: So actually, it turns out that the tarball logic errors if no files exist in the chart archive, so no need to check it here

jdolitsky (approved):

PR #6641, opened 2019-10-11
ref(cmd): Use method to list formats

marckhouzam:

There are two instances where the list of `output.format` is hard-coded outside the `output.go` file.  This breaks the rules of encapsulation and increases the risk that when adding a new format, the hard-coded lists could be forgotten.

This PR provides a method for listing the string representation of the different formats from within the `output.go` file.

### DIFF
--- a/cmd/helm/flags.go
+++ b/cmd/helm/flags.go
@@ -52,14 +52,8 @@ func addChartPathOptionsFlags(f *pflag.FlagSet, c *action.ChartPathOptions) {
 // bindOutputFlag will add the output flag to the given command and bind the
 // value to the given format pointer
 func bindOutputFlag(cmd *cobra.Command, varRef *output.Format) {
-	var formats strings.Builder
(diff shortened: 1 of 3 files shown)


marckhouzam:
@thomastaylor312 has recently worked on this area and may have an opinion on this possible change.

marckhouzam:
I've thought things over and I've pushed a second commit that greatly simplifies this change.  It is not as generic because it no longer provides a method to list all `output.Format` but instead a method that only lists the **string representation** of those formats.

Since it is the string representation that is actually needed, the original complexity of the more generic approach was not justified in my eyes.

marckhouzam:
I kept the two commits in case a maintainer wanted to look at the more generic approach to get a better understanding.  But I can squash the two commits if requested.

thomastaylor312 (approved):
I think the current implementation is the best compromise. It encapsulates things properly and serves the most common need for the formats

PR #6653, opened 2019-10-11
fix(action): Fixes ordering of variable binding

thomastaylor312:
The recent init action config switched the order of how variables get bound
and where. This led to the namespace variable not being propagated down into
the calls to kubernetes. To manage this, we have changed the behavior of the `InitActionConfig` to be a method on the config struct called `Init`


adamreese (approved):
Manually tested

Issue #5814, opened 2019-05-29
Helm 3: --wait flag does not seem to wait.


jeremyje:
I ran the following command which you can [find in the open-match Makefile](https://github.com/jeremyje/open-match/blob/helmet/Makefile#L272). It does not seem that `--wait` waits for the resources to be ready. This is from a new GKE cluster and can also repro by deleting the chart via `helm delete open-match`.

The output that comes out is:

```bash
helm upgrade open-match --install --wait --debug install/helm/open-match \
        --timeout=10m \
        --namespace=open-match \
        --set openmatch.image.registry=gcr.io/jeremyedwards-gaming-dev \
        --set openmatch.image.tag=0.0.0-dev-67e02bd \
        --set grafana.enabled=true \
        --set jaeger.enabled=true \
        --set prometheus.enabled=true \
        --set redis.enabled=true \
        --set openmatch.monitoring.stackdriver.enabled=true \
        --set openmatch.monitoring.stackdriver.gcpProjectId=jeremyedwards-gaming-dev \
        --debug
history.go:108: [debug] getting history for release open-match
Release "open-match" does not exist. Installing it now.
printer.go:84: [debug] Original chart version: ""
printer.go:84: [debug] CHART PATH: /usr/local/google/home/jeremyedwards/workspace/helm/open-match/install/helm/open-match

client.go:78: [debug] building resources from manifest
client.go:83: [debug] creating 79 resource(s)
wait.go:47: [debug] beginning wait for 79 resources with timeout of 10m0s
```
Output of `helm version`:
```
version.BuildInfo{Version:"v3.0.0-alpha.1", GitCommit:"b9a54967f838723fe241172a6b94d18caf8bcdca", GitTreeState:"clean"}
```

Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.2", GitCommit:"66049e3b21efe110454d67df4fa62b08ea79a19b", GitTreeState:"clean", BuildDate:"2019-05-16T16:23:09Z", GoVersion:"go1.12.5", Compiler:"gc", Platform:"linux/amd64"}
[...]

PR #6655, opened 2019-10-12
fix(pkg/downloader): Remove comparing hashed digests on build cmd

### LINKED ISSUE #6416: "Error: Chart.lock is out of sync with Chart.yaml" If Chart.yaml dependency has an alias
I've started seeing this issue after updating to beta 3 of Helm 3 from 2.14.3. If I remove the alias from my dependency, Chart.lock and Chart.yaml stay in sync. 

Output of `helm version`: 

Output of `kubectl version`: 

Docker Desktop

hangpark:
Signed-off-by: Hang Park <hangpark@kaist.ac.kr>

This PR includes the removal of checking sync status of `Chart.yaml` and `Chart.lock` dependencies before executing the main functionality of `helm dependency build`.

This resolves #6416. Moreover, similar to #6416, this resolves abnormal failures on `build` when `Chart.yaml` contains some subchart versions in the form of **SemVer range**, too.

Fixes #6416 

- #6416 reports that `helm dependency build` command (abbr. `build`) fails if `Chart.yaml` contains subchart's alias.
- I found that a same failure happens if `Chart.yaml` contains *SemVer range* on some of its subcharts.

These two bugs are originated from a mismatch of hashed digests of deps in `Chart.yaml` and `Chart.lock`.

---

Before #5874 merged, hashed digest in `Chart.lock` is calculated from `Chart.yaml` deps.
This makes `Chart.lock` not be updated properly when all of below happened:

1. subchart's version is written in SemVer range
2. that subchart is updated to a new version (still within the range)
3. `helm dep up` is executed and so the new version is installed

because it is true that the contents of `Chart.yaml` are not changed!

So, #5874 made digest calculation to use `Chart.lock` contents, not of `Chart.yaml`.

---

But notice that (1) lock file does not contain alias information and (2) lock file has specific *resolved* versions while yaml file may have version ranges.

Checking sync on yaml & lock (actually comparing hashed digests of deps in each file) in `build` command worked well before #5874.
However, after that PR merged, this comparing process is not valid with ranged versions (and #6416 was a bug even before #5874.)

More technically, `build` command (1) loads a chart and stores it in `var c *chart.Chart`, (2) calculates the hash value of `c.Meta.Dependencies`, and then (3) compares it with `c.Lock.Digest` value, which was already written in the `Chart.lock` file.

That `c.Lock.Digest` is originated from previous `dep update` command, which (1) loads a chart, (2) resolves deps, (3) calculates the hash value of **resolved deps**, and then (4) stores resolved deps along with its hash value as a digest into `Chart.lock` file.

Note that `c.Meta.Dependencies[i].Alias` may exist but `c.Lock.Dependencies[i].Alias` is always empty, and that `c.Meta.Dependencies[i].Version` may be SemVer range but `c.Lock.Dependencies[i].Version` is just a single resolved version. (see internal/resolver/resolver.go)

---

**Thus, I removed comparing hashed digests of two files in `build` command.**

My PR is the simplest way to fix bugs I think, but there are many other approaches **without removal of comparing two files**.  For example:

1. record digests of both original deps declaration and resolved lock file contents, so use one of it in each desired situation properly (we can concatenate two digests into one digest, maybe)
2. instead of comparing digests, determine out-of-sync manually like: `dep_yaml.name == dep_lock.name` && `dep_yaml.repo == dep_lock.repo` && `dep_yaml.version CONTAINS dep_lock.version` && each pairs of `(name, repo)` in `Chart.yaml` APPEARS in `Chart.lock` also (and vice-versa.)

If you guys think these approaches are better than this PR, I'd like to work further on this issue.

**If applicable**:


karuppiah7890:
Thanks a lot for this PR @hangpark . Looks like you fixed a bug, could you write a failing test that reproduces the bug / root cause of the bug? So that your code can pass the failing test 😄 I'll try to understand the code in the meantime

hangpark:
Hi, @karuppiah7890. I just pushed a commit for failing tests. These tests are all passed in mine but failed in upstream `dev-v3`. Following is a result of reproduced failures:

```
=== RUN   TestBuild_WithoutOptionalFields
--- PASS: TestBuild_WithoutOptionalFields (0.01s)
=== RUN   TestBuild_WithSemVerRange
--- FAIL: TestBuild_WithSemVerRange (0.01s)
    manager_test.go:250: Chart.lock is out of sync with Chart.yaml
=== RUN   TestBuild_WithAlias
--- FAIL: TestBuild_WithAlias (0.01s)
    manager_test.go:250: Chart.lock is out of sync with Chart.yaml
=== RUN   TestBuild_WithCondition
--- FAIL: TestBuild_WithCondition (0.01s)
    manager_test.go:250: Chart.lock is out of sync with Chart.yaml
=== RUN   TestBuild_WithTags
--- FAIL: TestBuild_WithTags (0.01s)
    manager_test.go:250: Chart.lock is out of sync with Chart.yaml
FAIL
```

You can check that optional fields make a failure.

thomastaylor312:
@hangpark Thanks for digging into this! It seems like the logic is pretty sound, but I'm going to get @mattfarina and @bacongobbler to take a look at this as well because the dependency locking code is not something I've messed with too much

pkg/downloader/manager.go
```diff
@@ -78,11 +78,6 @@ func (m *Manager) Build() error {
 		return m.Update()
 	}
 
-	req := c.Metadata.Dependencies
-	if sum, err := resolver.HashReq(req); err != nil || sum != lock.Digest {
-		return errors.New("Chart.lock is out of sync with Chart.yaml")
-	}
```
mattfarina: Consider the case where a dependency has a range set on it of `^1.2.3` in the Chart.yaml file. Now, someone changes that range to `^1.3.4`. But, the lock file has not been updated since this change happened. The current code should catch that. Removing this section would stop the end user from knowing that something is wrong.

We need to fix the broken cases but I would want to warn the user if something is wrong.

mattfarina (changes requested):

hangpark:
@mattfarina Thanks for your review. Instead of comparing digests, I tried to implement comparing the contents of the chart and the lock file manually. Is this good enough to inform unsynchronized status to the end-users?

mattfarina:
@hangpark The idea of the digest is to make sure the lock file was generated from the content of the requirements. If the requirements changed then the hash would change and you would know the lock file was not generated from those requirements. This is a simple check. Why does this not work?

hangpark:
@mattfarina I agree that digest comparison is a simple idea to compare dependencies and the lock file. But by #5814, the digest in a lock file is generated from lock contents. These lock contents don't have any information about *aliases, conditions, tags, and version ranges*. The digest comparison in `dep build` command compares that digest from lock contents, to a newly calculated digest from `Chart.yaml` dependencies. It fails if `Chart.yaml` has any of aliases, conditions, tags, or version ranges. Unless digest calculation is changed to use both `Chart.yaml` and `Chart.lock`, we cannot know whether the lock file is out-of-sync or not from the digest.

Since I'm new to here, I couldn't determine which solution is the best one, as I introduced several approaches in the `Special notes for your reviewer` section above. I picked the second one, and I know that one doesn't use the digest anymore.

To reflect your point in this PR, I think we should change the digest calculation function (`HashReq()`) to accept two arguments, `Chart.yaml` dependencies and `Chart.lock` dependencies. It can make the digest be aware of both `Chart.yaml` and `Chart.lock` changes. How do you think about this approach? It would work with both `dep build` and `dep up`.

In this approach, I worry about an edge case:
1. `Chart.yaml` has `>= 1.0.0` dep
1. Lock file created, `1.0.1` targetted.
1. Change `Chart.yaml` from `>= 1.0.0` to `>= 1.0.1`
1. Lock file still valid in range, but `dep build` will fail with `out of sync`.

bacongobbler:

hangpark:
Re-open in #6736

PR #6656, opened 2019-10-12
v3: 'helm show' -> 'helm show all' and 'helm get' -> 'helm get all'

### LINKED ISSUE #6552: v3: Commands should not have both args and subcommands
I don't know if this was ever discussed and it is probably too late to bring it up but here goes, just in case.

`helm get` and `helm show` accept both an argument (RELEASE_NAME or CHART respectively) or a subcommand.  This can cause conflicts where helm does not know if the argument is a subcommand or a real argument.  For example:

If I have a directory named `chart` (which I do), the command `helm show chart` does not work as it thinks the `chart` argument is the subcommand.  So I need to run `helm show ./chart`; if the directory is named something else, such as `mychart`, I do not need to do this.

As for `helm get`, if I have the bad idea of naming a release `values` (or `manifest` or `hooks`), then `helm get values` will not work, although I have respected the `helm get [RELEASE_NAME]` format.  There is probably no workaround for this except running all three `helm get` subcommands sequentially, or renaming my release.

Finally, shell-completion (surprise! :wink:) of a RELEASE_NAME is not possible for `helm get <TAB>` since it is the subcommands that will automatically be suggested.  On the other hand `helm get values <TAB>` currently does provide dynamic suggestions of release names.

I don't know if these minor hiccups can justify a breaking change at this time.  If so, I think the subcommand `helm get all [RELEASE_NAME]` and `helm show all [CHART]` could be a way to replace the existing format that does not use subcommands.

I've looked at `kubectl`, and I believe it is careful not to allow such situations.

marckhouzam:
Closes #6552

Replace `helm show` with `helm show all`
Replace `helm get` with `helm get all`

Thid is a break in compatibility because `helm show` and `helm get` are no longer valid commands on their own; what they used to do is now achieved with `helm show all` ans `helm get all` respectively.

This change avoids confusion between chart reference/release name and subcommands.
It also allows dynamic shell completion to work properly.

See #6552 for more details.

This is an approved break in command compatibility.

There are also a couple of small changes in help messages to standardize across subcommands (e.g. for `helm get notes`)

**If applicable**:
- [x] this PR contains documentation (assuming help messages count as documentation)
- [x] this PR contains unit tests (modification to naming of existing tests)

### DIFF
--- a/cmd/helm/get.go
+++ b/cmd/helm/get.go
@@ -23,51 +23,27 @@ import (
 
 	"helm.sh/helm/v3/cmd/helm/require"
 	"helm.sh/helm/v3/pkg/action"
-	"helm.sh/helm/v3/pkg/cli/output"
 )
 
 var getHelp = `
-This command shows the details of a named release.
-
-It can be used to get extended information about the release, including:
+This command consists of multiple subcommands which can be used to
+get extended information about the release, including:
 
   - The values used to generate the release
-  - The chart used to generate the release
   - The generated manifest file
-
-By default, this prints a human readable collection of information about the
-chart, the supplied values, and the generated manifest file.
+  - The notes provided by the chart of the release
+  - The hooks associated with the release
 `
 
 func newGetCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
-	var template string
-	client := action.NewGet(cfg)
-
 	cmd := &cobra.Command{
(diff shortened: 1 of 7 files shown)


thomastaylor312 (approved):
Looking great! Thanks for handling this. Going to request @bacongobbler for the second review since he discussed the original issue with you

bacongobbler (approved):

PR #6657, opened 2019-10-12
Remove all known arguments in plugin invocations

jonasrutishauser:
Consistenly removes all arguments in plugin invocations which are passed as environment variables:
- `--context` is renamed to `--kube-context`
- added missing known arguments including short option of `--namespace`

Additional fix: Get all arguments from environment variables passed to plugins:
- `HELM_KUBECONTEXT` was the only environment variable not processed

### DIFF
--- a/cmd/helm/load_plugins.go
+++ b/cmd/helm/load_plugins.go
@@ -127,7 +127,7 @@ func loadPlugins(baseCmd *cobra.Command, out io.Writer) {
 func manuallyProcessArgs(args []string) ([]string, []string) {
 	known := []string{}
 	unknown := []string{}
-	kvargs := []string{"--context", "--namespace"}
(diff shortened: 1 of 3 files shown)


thomastaylor312:
@jonasrutishauser Just wanted to clarify after reading the thread in https://github.com/databus23/helm-diff/pull/151. The idea here is to _only_ pass arguments via environment variables for known flags, correct?

jonasrutishauser:
Exactly.

This way all known flags are parsed by helm and passed to plugins via environment variables.
If a plugin later calls helm again, the configuration will be picked from the environment variables passed previously to the plugin.
If a plugin needs one of this values it can read it from the environment.

This way theses flags will always have the meaning defined by helm and can't accidentally be overridden by a plugin.

adamreese (approved):

thomastaylor312 (approved):

PR #6661, opened 2019-10-13
fix install storing computed values in release

### LINKED ISSUE #6659: [HELM 3] Get Values return all default values of parent object
When we try to get values from a release deployed with one value set, we get the parent object with all this default value.

Steps to reproduce:
```
$ helm install ingress-controller -n ingress stable/nginx-ingress --set controller.replicaCount=5
```

with helm 2
```
$ helm get values
controller:
  replicaCount: 5
```
with helm 3
```
$ helm get values -n ingress ingress-controller
controller:
  addHeaders: {}
  admissionWebhooks:
    enabled: false
    failurePolicy: Fail
    patch:
      enabled: true
      image:
        pullPolicy: IfNotPresent
        repository: jettech/kube-webhook-certgen
        tag: v1.0.0
...
  replicaCount: 5
...
 ```

The issue seems to appear only when the value set has a parent object.
With this, there is no issue:
```
$ helm install ingress-controller -n ingress stable/nginx-ingress --set revisionHistoryLimit=15
```

Output of `helm version`:
```
```

Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.1", GitCommit:"d647ddbd755faf07169599a625faf302ffc34458", GitTreeState:"clean", BuildDate:"2019-10-02T23:49:07Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"darwin/amd64"}
```

GKE, Docker-for-desktop

### PR DESCRIPTION
**Pull request by karuppiah7890**
this was partially fixed in #6430 but the fix only
worked for values without nesting. this PR fixes it.
this is done by doing a deep copy of values rather
than a top level keys copy. deep copy ensures
values are not mutated during coalesce()
execution which leads to bugs like #6659

the deep copy code has been copied from:

https://gist.github.com/soroushjp/0ec92102641ddfc3ad5515ca76405f4d

which is in turn inspired by this stackoverflow answer:
http://stackoverflow.com/a/28579297/1366283

Signed-off-by: Karuppiah Natarajan <karuppiah7890@gmail.com>

Fixes #6659 

**If applicable**:

### DIFF
--- a/internal/third_party/deepcopy/util.go
+++ b/internal/third_party/deepcopy/util.go
@@ -1,0 +1,39 @@
+package deepcopy
+
+// DeepCopyMap a function for deep copying map[string]interface{}
+// values. Inspired by:
+// https://gist.github.com/soroushjp/0ec92102641ddfc3ad5515ca76405f4d by Soroush Pour
+// The gist code is MIT licensed
+// which in turn has been inspired by the StackOverflow answer at:
+// http://stackoverflow.com/a/28579297/1366283
+//
+// Uses the golang.org/pkg/encoding/gob package to do this and therefore has the
+// same caveats.
+// See: https://blog.golang.org/gobs-of-data
+// See: https://golang.org/pkg/encoding/gob/
+
+import (
+	"bytes"
+	"encoding/gob"
+)
+
+func init() {
+	gob.Register(map[string]interface{}{})
+}
+
+// Map performs a deep copy of the given map m.
+func Map(m map[string]interface{}) (map[string]interface{}, error) {
+	var buf bytes.Buffer
+	enc := gob.NewEncoder(&buf)
+	dec := gob.NewDecoder(&buf)
+	err := enc.Encode(m)
+	if err != nil {
+		return nil, err
+	}
+	var mapCopy map[string]interface{}
+	err = dec.Decode(&mapCopy)
+	if err != nil {
+		return nil, err
+	}
+	return mapCopy, nil
+}
--- a/pkg/action/action_test.go
+++ b/pkg/action/action_test.go
@@ -173,7 +173,17 @@ func withName(name string) chartOption {
 }
 
 func withSampleValues() chartOption {
-	values := map[string]interface{}{"someKey": "someValue"}
+	values := map[string]interface{}{
+		"someKey": "someValue",
+		"nestedKey": map[string]interface{}{
+			"simpleKey": "simpleValue",
+			"anotherNestedKey": map[string]interface{}{
+				"yetAnotherNestedKey": map[string]interface{}{
+					"youReadyForAnotherNestedKey": "No",
+				},
+			},
+		},
+	}
 	return func(opts *chartOptions) {
 		opts.Values = values
 	}
--- a/pkg/action/upgrade_test.go
+++ b/pkg/action/upgrade_test.go
@@ -223,7 +223,6 @@ func TestUpgradeRelease_ReuseValues(t *testing.T) {
 		expectedValues := map[string]interface{}{
 			"subchart": map[string]interface{}{
 				"enabled": false,
-				"global":  map[string]interface{}{},
 			},
 		}
 		is.Equal(expectedValues, updatedRes.Config)
--- a/internal/third_party/deepcopy/util_test.go
+++ b/internal/third_party/deepcopy/util_test.go
@@ -1,0 +1,130 @@
+package deepcopy_test
+
+import (
+	"helm.sh/helm/v3/internal/third_party/deepcopy"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCopyMap(t *testing.T) {
+	testCases := []struct {
+		// original and expectedOriginal are the same value in each test case. We do
+		// this to avoid unintentionally asserting against a mutated
+		// expectedOriginal and having the test pass erroneously. We also do not
+		// want to rely on the deep copy function we are testing to ensure this does
+		// not happen.
+		original         map[string]interface{}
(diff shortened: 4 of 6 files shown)

### CODE REVIEW

**Comment by karuppiah7890** [comment]
Also, a question, I don't know how coalesce stuff works exactly. The code looks pretty complex 🙈 , but I see that `copyMap` has been used here : https://github.com/helm/helm/blob/0dae8bbd3fb398ec9e760c93bfd73852f57ce4fc/pkg/chartutil/coalesce.go#L109 

What kind of copying is expected here? One level of keys copy? or a deep copy? As #6430 previously used `copyMap` and now uses DeepCopyMap for fixing the bug. Just curious

**Comment by karuppiah7890** [comment]
legit failure in tests. I will fix the PR!

File: pkg/action/action_test.go [hunk 1]
```diff
@@ -173,7 +173,17 @@ func withName(name string) chartOption {
...
 
 func withSampleValues() chartOption {
-	values := map[string]interface{}{"someKey": "someValue"}
+	values := map[string]interface{}{
+		"someKey": "someValue",
+		"nestedKey": map[string]interface{}{
+			"simpleKey": "simpleValue",
+			"anotherNestedKey": map[string]interface{}{
+				"yetAnotherNestedKey": map[string]interface{}{
+					"youReadyForAnotherNestedKey?": "No",
```
thomastaylor312: Might want to remove the `?` just in case here. If it goes through a yaml marshaler in a future test, I am not sure what will happen

thomastaylor312 (changes requested):
I tested everything manually and this looks good. Just one small nit to avoid potential headaches in the future

pkg/action/action_test.go
karuppiah7890: Sure! I'll do that. Also, about the third part package, is it okay to put the code at that level or should there be something like a "util" package inside the third party directory?

karuppiah7890: I have made the changes for the `?`

thomastaylor312:
@karuppiah7890 You brought up a good point about the third party thing. I thought we had other code at that top level but we don't. Let's just put it in its own package called `deepcopy` inside of the third_party directory

karuppiah7890:
@thomastaylor312 Cool, I will do that 👍

karuppiah7890 (commented):
@thomastaylor312 Please check this too

karuppiah7890:
@thomastaylor312 I have done the appropriate changes. Please do check it out. Also do check this comment - https://github.com/helm/helm/pull/6661#pullrequestreview-301615697

internal/third_party/deepcopy/util.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"bytes"
+	"encoding/gob"
+)
+
+func init() {
+	gob.Register(map[string]interface{}{})
+}
+
+// DeepCopyMap performs a deep copy of the given map m.
```
thomastaylor312: Comment needs to be updated to match the new function signature

pkg/action/upgrade_test.go
```diff
@@ -223,7 +223,6 @@ func TestUpgradeRelease_ReuseValues(t *testing.T) {
 		expectedValues := map[string]interface{}{
 			"subchart": map[string]interface{}{
 				"enabled": false,
-				"global":  map[string]interface{}{},
```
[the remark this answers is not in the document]
thomastaylor312: In this case, whatever Helm 2 does, should be what this does. Global is generally a top level field though. Just to make sure I understand, the DeepCopy fix caused this test to start failing, correct? Or is this a separate issue?

thomastaylor312 (commented):
Just a few more small things/continued discussion on the unit test question you brought up

internal/third_party/deepcopy/util.go
karuppiah7890: Oops, yes, will do 🙈

karuppiah7890: Weird that ci passed :thin

pkg/action/upgrade_test.go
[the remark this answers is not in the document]
karuppiah7890: Correct, I missed to mention that. The above test failed due to this PR's code

internal/third_party/deepcopy/util.go
karuppiah7890: Created an issue for that #6683

pkg/action/upgrade_test.go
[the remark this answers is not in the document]
karuppiah7890: I'll reproduce such an install in helm2 and see what values are stored. My guess is - it will not have global. Will check it. Will also figure out why it came up in the first place and how it got removed with this fix 😄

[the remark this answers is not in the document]
karuppiah7890: In latest helm2, it doesn't store `global` field. And how it came up - not sure. My guess is, since `subchart` seems to be a key with nested value, and overriding value is nested, similar to issue #6659 , during some coalescing, the global key came up due to the absence of deep copy, so references were there. And now with this fix, it goes away

internal/third_party/deepcopy/util.go
karuppiah7890: Fixed this! 😄

pkg/action/upgrade_test.go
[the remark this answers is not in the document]
karuppiah7890: I think we are good with this change. What do you think?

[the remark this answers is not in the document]
thomastaylor312: Yep, I'm good with it!

thomastaylor312 (approved):
Thanks for working through all the iterations on this!

bacongobbler (approved):

PR #6662, opened 2019-10-13
Fix list not showing multiple releases with same name in different namespaces

### LINKED ISSUE #6658: Helm3: list all namespaces not showing multiple releases with same name in different namespaces
I created two releases with same name in two different namespaces, and then tried to list releases in all namespaces, but only one of them showed up instead of two showing up

Output of `helm version`:
```
```

karuppiah7890:
Fixes #6658 

**If applicable**:

### DIFF
--- a/pkg/action/list.go
+++ b/pkg/action/list.go
@@ -17,6 +17,7 @@ limitations under the License.
 package action
 
 import (
+	"path"
 	"regexp"
 
 	"helm.sh/helm/v3/pkg/release"
@@ -218,27 +219,23 @@ func (l *List) sort(rels []*release.Release) {
 }
 
 // filterList returns a list scrubbed of old releases.
-func filterList(rels []*release.Release) []*release.Release {
-	idx := map[string]int{}
+func filterList(releases []*release.Release) []*release.Release {
+	latestReleases := make(map[string]*release.Release)
 
-	for _, r := range rels {
-		name, version := r.Name, r.Version
-		if max, ok := idx[name]; ok {
-			// check if we have a greater version already
-			if max > version {
-				continue
-			}
+	for _, rls := range releases {
+		name, namespace := rls.Name, rls.Namespace
+		key := path.Join(namespace, name)
+		if latestRelease, exists := latestReleases[key]; exists && latestRelease.Version > rls.Version {
+			continue
 		}
-		idx[name] = version
+		latestReleases[key] = rls
 	}
 
-	uniq := make([]*release.Release, 0, len(idx))
-	for _, r := range rels {
-		if idx[r.Name] == r.Version {
-			uniq = append(uniq, r)
-		}
+	var list = make([]*release.Release, 0, len(latestReleases))
+	for _, rls := range latestReleases {
+		list = append(list, rls)
 	}
-	return uniq
+	return list
 }
 
 // setStateMask calculates the state mask based on parameters.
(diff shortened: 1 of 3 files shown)


karuppiah7890:
I tried to write a integration test in this file `cmd/helm/list_test.go` but couldn't write as the driver storage I used got overridden by the new driver storage that got created over here (for all namespaces alone) https://github.com/helm/helm/blob/01e593fbcd7f4df7b3db75c0db79c067f3372df4/cmd/helm/list.go#L73 
So I had no way to inject the dummy driver storage I created in tests

thomastaylor312:
Waiting on follow up to this before doing review: https://github.com/helm/helm/issues/6658#issuecomment-541747888

karuppiah7890:
@thomastaylor312 The issue exists. You can check it out 👍

thomastaylor312:
@karuppiah7890 So I am wondering if we actually need to do something with the memory driver (it actually makes me wonder if we even want the memory driver anymore). Would it be a more accurate test to just use the fake client set or export the mock implementation [here](https://github.com/helm/helm/blob/34b930cb9db8dba90aaba6299200c34664219a57/pkg/storage/driver/mock_test.go#L70)? It would also make this PR much simpler than modifying the in memory store (which is really only meant for testing)

hickeyma:
@thomastaylor312 This looks like a bug alright. It seems to be as per the scenario shown in https://github.com/helm/helm/issues/6658#issuecomment-541932629

karuppiah7890:
@thomaataylor312 sure. I tried to keep it consistent with other tests which also use memory driver. But yeah, mocking kube client could be done. I'll check it out! Are we sure about the memory driver part? I do see that it's only for testing. And it's true that we have to maintain and write code for that too to keep it working for the tests.

karuppiah7890:
Like, are we planning to get rid of it in the future?

karuppiah7890:
Cool, I'll use the other mocks. What about the support for namespaces in memory driver? Let that be there? Or do you want to get rid of it?

karuppiah7890:
Also, I want to know what we can do to make testing easier at the command level. Check this comment - https://github.com/helm/helm/pull/6662#issuecomment-541437946

thomastaylor312:
Let's get rid of the namespace support there for now (or introduce it in a separate PR)

thomastaylor312:
Oh, and honestly I am not sure about how to make the testing easier there, but I think we can revisit that later

karuppiah7890:
Makes sense, I'll just add unit tests then. Will create separate PR for namespace support for memory storage driver - review it and if it's needed we can merge it

karuppiah7890:
Created separate PR for adding namespace support to memory storage driver - #6684

thomastaylor312:
@karuppiah7890 Are you going to pull those changes out from here or are you waiting on that other PR?

karuppiah7890:
I'm not waiting on the other PR :)

karuppiah7890:
@thomastaylor312 Even the `MockSecretsInterface` doesn't support secrets in multiple namespaces. Just one namespace is supported. At least that's how it looks. I just tried. Could you check?

I'm thinking of other ways to test this. Probably by mocking methods, using github.com/golang/mock may be? But I have to see how much mocking has to be done 😅

karuppiah7890:
I added tests for just the `filterList` function

bacongobbler:
Apologies for the automatic closure. We just switched the `dev-v3` branch to `master`, and the old `master` branch is now `dev-v2`. Would you mind re-opening a new PR with this change? Thanks again and sorry for the hassle!

PR #6669, opened 2019-10-15
ref(pkg/chartutil): Improve error handling code in Save

SimonAlling:
This PR is intended to make the code more understandable, to prevent future mistakes and confusion. See the commit message for details.

### DIFF
--- a/pkg/chartutil/save.go
+++ b/pkg/chartutil/save.go
@@ -135,7 +135,7 @@ func Save(c *chart.Chart, outDir string) (string, error) {
 		rollback = true
 		return filename, err
 	}
-	return filename, err
+	return filename, nil
 }
 
 func writeTarContents(out *tar.Writer, c *chart.Chart, prefix string) error {


bacongobbler:
[Effective Go convention](https://github.com/golang/go/wiki/CodeReviewComments#variable-names) has described using short variable names: `err` instead of `errCreatingFile`. If the variable's original intent is unclear then it should be made more descriptive. However, since there is no difference in handling between `errCreatingFile`, `errInChartYaml` and `errInValuesYaml`, `err` is more than descriptive enough.

My other issue with this PR is that some of these variable names are misleading. `errCreatingFile` doesn't always mean `os.Create` failed to create the file. It can also occur because `os.Create` failed to truncate the file that already existed. By giving a variable a "descriptive" error name can exclude certain cases that the developer may miss because they inferred the intent based on the varaible name.

I'm 👎 on this change because it goes against the current convention of Effective Go, and it doesn't actually change the behaviour of anything other than a few variable names.

SimonAlling:
@bacongobbler Aha. I must admit I didn't know about the Effective Go convention. As for `errCreatingFile`, I picked that name based on the name of `os.Create` without further ado. I definitely see your point there.

Variable names aside, please do note that the current code did actually cause a pretty serious bug where most errors were silently ignored (which I semi-accidentally [fixed](https://github.com/helm/helm/pull/6519/files#diff-9ea0c476e7eb107c6b0f198af1a49caaR136) as part of #6519) – a bug that would never have occurred had each error variable either been more locally scoped (not possible, AFAICT) or named differently. If you take a look at [lines 125–128](https://github.com/helm/helm/blob/38d4f33a0fc1829765ee720327ed29c370d68a79/pkg/chartutil/save.go#L125:L128) before #6519, you can probably see how said bug was caused by how the `err` variable is named and used. The purpose of this PR is to prevent similar mistakes in the future.

Even if you don't want to change any names, I absolutely think the last line of `Save` should be changed from `return filename, err` to `return filename, nil`. That would certainly decrease the risk of introducing similar bugs again. Anyway, it doesn't really make sense at that point to return the `err` variable declared and assigned only once back in line 112, and already known to be `nil`.

What do you think?

SimonAlling:
Done, @bacongobbler!

bacongobbler (approved):

PR #6670, opened 2019-10-15
fix(version): implement `helm version -c`, mark as hidden

### LINKED ISSUE #6322: Feature request: a `helm version` option that works consistently across v2 and v3

bacongobbler:

supersedes #6368
closes #6322

### DIFF
--- a/cmd/helm/version.go
+++ b/cmd/helm/version.go
@@ -59,8 +59,10 @@ func newVersionCmd(out io.Writer) *cobra.Command {
 		},
 	}
 	f := cmd.Flags()
+	f.BoolP("client", "c", true, "display client version information")
 	f.BoolVar(&o.short, "short", false, "print the version number")
 	f.StringVar(&o.template, "template", "", "template for version string format")
+	f.MarkHidden("client")
 
 	return cmd
 }
--- a/cmd/helm/testdata/output/version-client-shorthand.txt
+++ b/cmd/helm/testdata/output/version-client-shorthand.txt
@@ -1,0 +1,1 @@
+version.BuildInfo{Version:"v3.0+unreleased", GitCommit:"", GitTreeState:"", GoVersion:""}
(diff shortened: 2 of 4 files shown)


cmd/helm/version.go
```diff
@@ -59,8 +59,10 @@ func newVersionCmd(out io.Writer) *cobra.Command {
 		},
 	}
 	f := cmd.Flags()
+	f.BoolP("client-only", "c", true, "display client version information")
```
marckhouzam: Isn't the flag in helm v2 named `client`?

marckhouzam: `--client-only` did feel familiar but I couldn't quite figure out why.
Then I searched the code and found that it is `helm init --client-only` but `helm version --client` 😉

cmd/helm/version.go
```diff
@@ -59,8 +59,10 @@ func newVersionCmd(out io.Writer) *cobra.Command {
...
 	f.BoolVar(&o.short, "short", false, "print the version number")
 	f.StringVar(&o.template, "template", "", "template for version string format")
+	f.MarkHidden("client-only")
```
marckhouzam: 👍 for hidden

cmd/helm/version.go
bacongobbler: Gah. I'm mixing flags again -_-

Thanks for catching this.

hickeyma (dismissed):
@bacongobbler Initially I was ok with adding this capability in but during the review I am thinking this is not such a good idea.

My reasoning is as follows:
We are adding a flag which is artificial and contrived in Helm 3 as there is no client and server anymore. There is just helm.  Helm 3 removes Tiller and the client/server architecture and therefore all UX around this. When we add this `--c` flag, we are reintroducing a workaround for a niche case that some users used in v2.  I know that `helm version` has roundtrip issues with Tiller in v2 but it provides the capability in a standard UX fashion.
I am therefore not sure that this justifies putting this in place.

thomastaylor312:
I think I agree with @hickeyma on this one. I feel like adding a no-op flag for something now non-existent like Tiller that is essentially The Thing™ people have been wanting Helm 3 is probably not the best. This would be helping one specific use case that can still be easily worked around, albeit in a slightly verbose/clunky fashion

bacongobbler:
Okay. Closing as there is consensus to not go ahead with this feature.

PR #6679, opened 2019-10-16
feat(*): Adds custom time package for better marshalling

### LINKED ISSUE #6163: helm3: deleted in release info has strange default

thomastaylor312:
This package mainly exists to workaround an issue in Go where the serializer doesn't omit an empty value for time: https://github.com/golang/go/issues/11939. This replaces all release and hook object time references with the new time package so things actually marshal correctly.

I also have a second commit that wraps most of the stdlib `Time` object methods so we can use our time wrapper without any weird conventions. This can be popped off and discarded if not desired.

I fully admit this adds some code and mental overhead, but in service of good output for users of the CLI and SDK. If we consider this trade off too much, I have no problem just closing this PR. See the thread starting [here](https://github.com/helm/helm/pull/6229#issuecomment-541936523) for additional context.

Supersedes: #6229 
Fixes #6163

### DIFF
--- a/cmd/helm/history.go
+++ b/cmd/helm/history.go
@@ -19,7 +19,7 @@ package main
 import (
 	"fmt"
 	"io"
-	"time"
+	stdtime "time"
 
 	"github.com/gosuri/uitable"
 	"github.com/spf13/cobra"
@@ -30,6 +30,7 @@ import (
 	"helm.sh/helm/v3/pkg/cli/output"
 	"helm.sh/helm/v3/pkg/release"
 	"helm.sh/helm/v3/pkg/releaseutil"
+	"helm.sh/helm/v3/pkg/time"
 )
 
 var historyHelp = `
@@ -98,7 +99,7 @@ func (r releaseHistory) WriteTable(out io.Writer) error {
 	tbl := uitable.New()
 	tbl.AddRow("REVISION", "UPDATED", "STATUS", "CHART", "APP VERSION", "DESCRIPTION")
 	for _, item := range r {
-		tbl.AddRow(item.Revision, item.Updated.Format(time.ANSIC), item.Status, item.Chart, item.AppVersion, item.Description)
+		tbl.AddRow(item.Revision, item.Updated.Format(stdtime.ANSIC), item.Status, item.Chart, item.AppVersion, item.Description)
 	}
 	return output.EncodeTable(out, tbl)
 }
--- a/cmd/helm/history.go
+++ b/cmd/helm/history.go
@@ -19,7 +19,7 @@ package main
 import (
 	"fmt"
 	"io"
-	stdtime "time"
+	"time"
 
 	"github.com/gosuri/uitable"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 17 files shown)


cmd/helm/history.go
```diff
@@ -19,7 +19,7 @@ package main
 import (
 	"fmt"
 	"io"
-	"time"
+	stdtime "time"
```
bacongobbler: Would it make sense to call helm's time package `helmtime` and keep `time` as `time`? I feel that may be less confusing to the reader that `stdtime` isn't some other third party package. Also, it opens up the opportunity for us to say

> it's `helmtime`

:trollface:

bacongobbler:
ewww, but it works so :man_shrugging: 

What if we made `pkg/time` internal? `internal/time`

pkg/release/info.go
```diff
@@ -24,9 +26,9 @@ type Info struct {
 	// LastDeployed is when the release was last deployed.
 	LastDeployed time.Time `json:"last_deployed,omitempty"`
 	// Deleted tracks when this object was deleted.
-	Deleted time.Time `json:"deleted,omitempty"`
+	Deleted time.Time `json:"deleted"`
 	// Description is human-friendly "log entry" about this release.
-	Description string `json:"Description,omitempty"`
+	Description string `json:"description,omitempty"`
```
bacongobbler: :+1:

thomastaylor312:
@bacongobbler I was going to do the same thing, but because this time type is used in external packages, it needs to be exported so consumers of the SDK can use it. However, I am in favor of adding a very clear warning to the documentation for the package that it is not subject to the same backwards compatibility guarantees. What do you think?

cmd/helm/history.go
thomastaylor312: I thought about that. Since you brought it up, I'll change it

hickeyma (commented):
Thanks for this @thomastaylor312, the changes look good.

I am however, not so sure the gains of adding wrapper code like this based on a PR pending in the other package. I am just not sure do we want to inherit this. It would be worth getting some feedback from the other maintainers also.

thomastaylor312:
@hickeyma There isn't a PR pending for this issue in Go. The issue has been open for 4 years with multiple closed PRs attached to it, and the proposal seems to be stuck based on the labels (`Proposal-Hold`), so that isn't my worry.

My main worry is around having this code to maintain with the trade off in value in provides. I _really_ don't like marshaling out random timestamp strings that represent a zero value in Go, but are nonsense to anyone else. We could add custom marshaling to each object with a timestamp, but I think that would be even more brittle than this option (it was what I did on my first try). 

tl;dr: Is the user experience here (sane empty values) worth the code overhead?

hickeyma (approved):
I appreciate your feedback on my reservation.  It seems like good value to add then. LGTM, thanks @thomastaylor312

cmd/helm/history.go
thomastaylor312: Done

bacongobbler (approved):

Issue #6642, opened 2019-10-11
helm3 3.0.0-beta.4 error upgrading PersistentVolumeClaim PVC with no changes


pmahoney:
I have a helm chart with a single Deployment and single PersistentVolumeClaim. The Deployment mounts the PVC. My cluster is configured with a default volume provisioner, which does create the PV and adds annotations to the PVC.

With helm 3.0.0-beta.4, I run `upgrade --install` which installs as expected. But if I run the command again, with no changes, I get the following error:

```
$ helm upgrade myrelease /tmp/data/chart --namespace default --logtostderr --install --history-max 3

client.go:385: Cannot patch PersistentVolumeClaim: "mypvc" (PersistentVolumeClaim "mypvc" is invalid: spec: Forbidden: is immutable after creation except resources.requests for bound claims)
client.go:404: Use --force to force recreation of the resource
Error: UPGRADE FAILED: PersistentVolumeClaim "mypvc" is invalid: spec: Forbidden: is immutable after creation except resources.requests for bound claims
```

Additionally, if I supply `--force`, helm does delete the PVC (which is **not** desirable) but then fails to re-create it because the resource is still in the process of being deleted (in other words, helm issued the delete command but did not wait for deletion).

Note with helm 3.0.0-alpha.2, I **can** run `upgrade --install` twice in a row with no changes and no errors (as expected).

The PVC after being provisioned
```
kubectl get pvc mypvc -o yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  annotations:
    pv.kubernetes.io/bind-completed: "yes"
    pv.kubernetes.io/bound-by-controller: "yes"
    volume.beta.kubernetes.io/storage-provisioner: rancher.io/local-path
    volume.kubernetes.io/selected-node: docker-desktop
  creationTimestamp: "2019-10-11T03:29:37Z"
  finalizers:
  - kubernetes.io/pvc-protection
  labels:
    app.kubernetes.io/name: mypvc
  name: mypvc
  namespace: default
  resourceVersion: "7770"
  selfLink: /api/v1/namespaces/default/persistentvolumeclaims/mypvc
  uid: 59fe0916-ebd7-11e9-b99d-0242ac110002
spec:
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 10Mi
  storageClassName: local-path
  volumeMode: Filesystem
  volumeName: pvc-59fe0916-ebd7-11e9-b99d-0242ac110002
status:
[...]

PR #6681, opened 2019-10-16
fix(kube): replace rather than delete/create on `helm install --force`

bacongobbler:
fixes a sub-issue in #6642 by replacing the object with another, forcing an overwrite. This ends in the same result, just with less API calls (and with no need to wait for a resource to be deleted before creating another object).


jenting (approved):
/LGTM

hickeyma (approved):
LGTM, thanks @bacongobbler.

One small nit before you merge. This probably needs updating: https://github.com/helm/helm/blob/dev-v3/cmd/helm/upgrade.go#L149 ?

PR #6684, opened 2019-10-17
feat(test): add ability to store releases in multiple namespaces in the memory storage driver

karuppiah7890:
This will be useful to test things that involves multiple namespaces. Like listing releases in all namespaces. I have done this as a lot of unit tests currently depend on the memory storage driver to test things. 

This was initially part of PR https://github.com/helm/helm/pull/6662 , later we thought we would use configmap / secrets mock instead of memory storage to test #6662 PR. Some parts of the implementation don't look very neat - I can see that. Let me know what you folks think. We can discuss more on that while reviewing but this is the best I could come up with what's currently available. There was some struggle around "how to set namespace" and also store records of all namespaces. In a real world scenario for secrets and configmaps - k8s has state, and it stores everything and for communication with k8s, we set namespace just once, or don't set one for denoting all namespaces, at least that's how it looks. Doesn't seem straight forward to do the same for memory storage and write tests too

**If applicable**:

### DIFF
--- a/cmd/helm/helm_test.go
+++ b/cmd/helm/helm_test.go
@@ -88,7 +88,7 @@ func runTestActionCmd(t *testing.T, tests []cmdTestCase) {
 }
 
 func storageFixture() *storage.Storage {
-	return storage.Init(driver.NewMemory())
+	return storage.Init(driver.NewMemory("default"))
 }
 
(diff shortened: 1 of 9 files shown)


thomastaylor312:
Marked as breaking as it changes a publicly exported function

PR #6687, opened 2019-10-17
feat(lint): Adds v3 chart checks to linter

### LINKED ISSUE #5733: Add lint feature to detect charts that will no longer work in v3
The idea is to add a lint feature (maybe even to v2) to detect elements in charts that will no longer work. For example `.Release.Revision` was removed in https://github.com/helm/helm/commit/aa859e3f88291c94220d26e9d1127c1557c2af98. It would be nice to provide some direction, possibly at the other end of a link, for these as well.

thomastaylor312:
Fixes #5733

This adds two specific checks. A warning if a chart has a `crd-install` hook
and an error if the chart contains `.Release.Time`. Further checks can be added
down the road as needed using the same pattern I use here

**If applicable**:

### DIFF
--- a/pkg/lint/rules/template.go
+++ b/pkg/lint/rules/template.go
@@ -19,6 +19,7 @@ package rules
 import (
 	"os"
 	"path/filepath"
+	"regexp"
 
 	"github.com/pkg/errors"
 	"sigs.k8s.io/yaml"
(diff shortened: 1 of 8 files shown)


bacongobbler (approved):
nice tests. LGTM!

mattfarina (approved):

PR #6691, opened 2019-10-17
fix(chartutil): Uses copystructure for deep copy to avoid using gob

thomastaylor312:
We already had the copystructure library in our dependencies transitively through sprig. This solves a gob encoding bug that was causing issues with chart testing


bacongobbler (approved):
red is good

thomastaylor312:
Because this is large purely for a removal of unneeded code, I am going to merge.

PR #6612, opened 2019-10-08
feat(test): add `--logs` to `helm test`

bacongobbler:
The original submitter hasn't responded in over a week, so I rebased #6167 off master and [addressed an earlier comment](https://github.com/helm/helm/pull/6167#discussion_r329748511).

supersedes #6167
closes #2923

PR #6693, opened 2019-10-17
feat(test): Adds --logs flag

thomastaylor312:
This is a v3 port of #6612. There have been significant changes due to the way
Helm 3 refactored things. I chose to add the method for getting logs to the
testing client because it seemed like something that someone using Helm
as an SDK might want. It takes a writer because it is more efficient (less
copying) and can write to any sort of buffer desired

Pod logs are not very easy to unit test, so I manually tested this with a succeeding test and a failing test to make sure it works as intended

### DIFF
--- a/cmd/helm/release_testing.go
+++ b/cmd/helm/release_testing.go
@@ -17,6 +17,7 @@ limitations under the License.
 package main
 
 import (
+	"fmt"
 	"io"
 	"time"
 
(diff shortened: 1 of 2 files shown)


cmd/helm/release_testing.go
```diff
@@ -36,25 +37,43 @@ The tests to be run are defined in the chart that was installed.
 
 func newReleaseTestCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 	client := action.NewReleaseTesting(cfg)
-	var outfmt output.Format
+	client.Namespace = settings.Namespace()
```
bacongobbler: Would you mind moving this to run inside `RunE`, right before we call `client.Run`? I fear there may be a risk of flag parsing race conditions occurring if we set the namespace when the commands are being assembled. We may want to set the namespace parameter when the command is actually executed like we do for other commands:

https://github.com/helm/helm/blob/c3e356faba0c23626a4eb84c3bb3656279468585/cmd/helm/upgrade.go#L74-L79

thomastaylor312: Good point. Done!

bacongobbler (approved):

PR #6695, opened 2019-10-17
docs(CONTRIBUTING): add Helm 2's support contract

### LINKED ISSUE #6144: Proposal: Helm v2 support plan once v3 is released
In the Helm developer call we discussed the Helm v2 support plan once v3 is released. Given the time table we are looking at there are migration issues for uses due to the holiday shopping season and tax season. We want to try to take that into account.

With that in mind here is the proposal once Helm v3.0.0 is released:

* Helm v2 is supported for bugs for the next 6 months. All new features will be in Helm v3.0.0
* Helm v2 is supported for security fixes only for the 6 months after that
* After the year support for Helm v2 will formally end. The Google Cloud buckets for the charts repositories, Helm client download, and tiller may go away. I would suggest this is likely. They are managed by Google (created prior to the CNCF). We will make tiller available in an alternative location (it can be used with the `--tiller-image` flag on `init`). Helm v2 downloads are already available at https://get.helm.sh as an alternative.

Note, if there was a method to transfer Google Cloud buckets from one project to another we would consider that but it is not possible. From our statistics we know that there have been numerous people using older Helm versions so we are sensitive to not everyone using the latest releases.

If you have questions, feedback, or thoughts on this please fee free to share.

bacongobbler:
closes #6144

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,7 +12,7 @@ us a chance to try to fix the issue before it is exploited in the wild.
 
 ## Sign Your Work
 
-The sign-off is a simple line at the end of the explanation for a commit. All 
+The sign-off is a simple line at the end of the explanation for a commit. All
 commits needs to be signed. Your signature certifies that you wrote the patch or
 otherwise have the right to contribute the material. The rules are pretty simple,
 if you can certify the below (from [developercertificate.org](https://developercertificate.org/)):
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -109,9 +109,9 @@ An issue that we are not sure we will be doing will not be added to any mileston
 
 A milestone (and hence release) is considered done when all outstanding issues/PRs have been closed or moved to another milestone.
 
-## Semantic Versioning
+## Backwards Compatibility
```
mattfarina: I think it's a good idea to call out semantic versioning as our scheme somewhere. This week I came across a project that changed an API in a patch release and it wasn't for security. Not everyone has the same idea of backwards compatibility that we do. It would be good to still have that listed somewhere.

bacongobbler: good call. I've reverted this change.

mattfarina (approved):

PR #6697, opened 2019-10-18
fix(cmd): Add release name to install examples

marckhouzam:

The `helm install` help text gives examples of an install command with different chart references.  Those commands don't actually work with v3 because v3 now requires a release name parameter.

This PR adds the release name to those examples.


hickeyma (approved):
LGTM, thanks @marckhouzam

PR #6698, opened 2019-10-18
Upgrade to go 1.13.3 (CVE-2019-17596)

tao12345666333:

full diff https://github.com/golang/go/compare/go1.13.1...go1.13.3

<p>
go1.13.2 (released 2019/10/17) includes security fixes to the
<code>crypto/dsa</code> package and the compiler.
See the <a href="https://github.com/golang/go/issues?q=milestone%3AGo1.13.2">Go
1.13.2 milestone</a> on our issue tracker for details.
</p>

<p>
go1.13.3 (released 2019/10/17) includes fixes to the go command,
the toolchain, the runtime, <code>syscall</code>, <code>net</code>,
<code>net/http</code>, and <code>crypto/ecdsa</code> packages.
See the <a href="https://github.com/golang/go/issues?q=milestone%3AGo1.13.3">Go
1.13.3 milestone</a> on our issue tracker for details.
</p>

**If applicable**:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@ jobs:
     working_directory: /go/src/k8s.io/helm
     parallelism: 3
     docker:
-      - image: golang:1.13.1
+      - image: golang:1.13.3
     environment:
       PROJECT_NAME: "kubernetes-helm"
     steps:
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 DOCKER_REGISTRY   ?= gcr.io
 IMAGE_PREFIX      ?= kubernetes-helm
-DEV_IMAGE         ?= golang:1.13.1
+DEV_IMAGE         ?= golang:1.13.3
 SHORT_NAME        ?= tiller
 SHORT_NAME_RUDDER ?= rudder
 TARGETS           ?= darwin/amd64 linux/amd64 linux/386 linux/arm linux/arm64 linux/ppc64le linux/s390x windows/amd64


mattfarina:
Note, the CI run was failing due to the image being pushed about 5 hours after this PR was created.

karuppiah7890:
We could use golang:1.13 and we would get the latest minor version always
when the build happens. What do you all think?

tao12345666333:
1) If we don't specify a minor version, it may not be easy to troubleshoot sporadic issues, although this may be rare.

2) If we don't specify a minor version, we don't need to pay attention to this change throughout the lifecycle of the go version.

3) Specify minor version, which is in line with best practices.

karuppiah7890:
@tao12345666333 

For this
> If we don't specify a minor version, it may not be easy to troubleshoot sporadic issues, although this may be rare.

We are actually specifying a patch version in this case, not minor. And I agree that not knowing golang version could make debugging tougher. `helm` currently shows what's the golang version it was built with:

```
$ helm version
version.BuildInfo{Version:"v3.0+unreleased", GitCommit:"3ee8f78c79db3025f991d4f399d444d516dff52e", GitTreeState:"dirty", GoVersion:"go1.13.1"}
```
so it's easy to troubleshoot

For
> If we don't specify a minor version, we don't need to pay attention to this change throughout the lifecycle of the go version.

Again, it's patch version here. I would recommend being always having the latest patches. I actually even recommend having the latest minor versions - why say no to good features and bug fixes. Only major versions are a pain as they contain breaking changes most times. Instead of always paying attention to the lifecycle of go version, and also manually changing version, it would be easier to just use `golang:1.13`. In reality a lot of people keep an eye on the golang versions and blogs, in our case, we just don't have to make changes continuously for the releases

For
> Specify minor version, which is in line with best practices

Again, it's patch version. And I haven't seen anyone tell that specifying patch / minor version is a best practice. In many package management systems, people mention packages with the spec that they want the latest in the given minor version, or even the latest in the given major version. And the versions are locked in a lock file and sometimes updated when installing packages. I think it's completely fine to have latest and greatest. Just my opinion. Core maintainers might feel differently

Issue #6505, opened 2019-09-25
helm3 template fails when there are custom resource templates


paulczar:
Output of `helm version`: v3.0.0-beta.3

When I have a custom resource in a helm template, if I try and use `helm template` it fails with:

```
$ helm template .
Error: apiVersion "certmanager.k8s.io/v1alpha1" in test/templates/cert-pks.yaml is not available
```

but works with `helm install`

```
$ AME: test
LAST DEPLOYED: 2019-09-25 17:27:46.218812309 -0500 CDT m=+1.092955819
NAMESPACE: test
STATUS: deployed
NOTES:
1. Get the application URL by running these commands:
  export POD_NAME=$(kubectl get pods -l "app=test,release=test" -o jsonpath="{.items[0].metadata.name}")
  echo "Visit http://127.0.0.1:8080 to use your application"
  kubectl port-forward $POD_NAME 8080:80
```

for reference sake ... I confirmed this behavior with a fresh `helm create test` and then added the following into templates dir

```yaml
apiVersion: certmanager.k8s.io/v1alpha1
kind: Certificate
metadata:
  name: {{ include "test.fullname" . }}
  labels:
    app.kubernetes.io/name: {{ include "test.name" . }}
    helm.sh/chart: {{ include "test.chart" . }}
    app.kubernetes.io/instance: {{ .Release.Name }}
    app.kubernetes.io/managed-by: {{ .Release.Service }}
spec:
  secretName: {{ include "test.fullname" . }}-cert
  dnsNames: ["bacon.potato.org"]
  acme:
    config:
    - dns01:
        provider: prod-clouddns
      domains: ["bacon.potato.org"]
  issuerRef:
    name: letsencrypt-prod
    kind: ClusterIssuer

```

bacongobbler:
Does `helm template --validate` work?

paulczar:
SMDH!  it sure does! Thanks!

do you think it would make sense to include a message telling folks they might want to try `--validate` on this kind of error ?

sidharthsurana:
I also ran into this issue. even though `--validate` works. but that means I need a working kubernetes cluster to connect to in order to just render the template. This does not fit with the use case I was planning to use the template command for. 
In my case, I was planning on using the template command with the provided var overrides to render the template in isolation (without k8s access) and then use the output to discover the objects that would be created by this chart if deployed. This info in our case could be used to for instance to show the end-user the blueprint of objects that would be created without actually deploying the chart.

[...]

Issue #6680, opened 2019-10-16
feature: register CRDs present in crds/ directory during `helm template`


bacongobbler:
A common use case in Helm 3 is to render all its templates to stdout. If your templates contain a custom resource that relies on a CRD present in the `crds/` directory, `helm template` will fail with a message similar to the following:

```
Error: apiVersion "core.oam.dev/v1alpha1" in rudr/templates/traits.yaml is not available
```

Two workarounds exist today:

- install the CRDs into your cluster and run with the `--validate` flag
- register the CRDs using the `--api-versions` flag

One usability fix for this problem may be to register these API versions present in the `crds/` directory into the Capabilities object similar to the `--api-versions` flag.

PR #6701, opened 2019-10-18
Skip api version checks when install action is configured in client mode only

lminaudier:

Hi there,

With helm 3, when an install action is run in client-mode only (`install` without `--validate`) and if your chart contains custom resources you end up hitting error that look like that.

```
apiVersion "example.com/v1" in my-app/templates/cutoms.yaml is not available
```

This PR attempt to workaround the issue by skipping entirely the check if we are in client mode as we can't be aware of any custom resource definition present on the remote cluster in advance as we don't interact with it in such mode.

Thoughts?

The error hit is similar to the one described in #6680 

**If applicable**:

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -211,7 +211,7 @@ func (i *Install) Run(chrt *chart.Chart, vals map[string]interface{}) (*release.
 	rel := i.createRelease(chrt, vals)
 
 	var manifestDoc *bytes.Buffer
-	rel.Hooks, manifestDoc, rel.Info.Notes, err = i.cfg.renderResources(chrt, valuesToRender, i.OutputDir, i.SubNotes)
+	rel.Hooks, manifestDoc, rel.Info.Notes, err = i.cfg.renderResources(chrt, valuesToRender, i.OutputDir, i.SubNotes, i.ClientOnly)
 	// Even for errors, attach this if available
 	if manifestDoc != nil {
 		rel.Manifest = manifestDoc.String()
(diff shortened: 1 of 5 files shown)


jlegrone:
I believe this should close https://github.com/helm/helm/issues/6505.

bacongobbler:
Hmm. I don't think passing a boolean across several API layers just to disable a check is a great design... The validation is in place to ensure that there aren't CRDs being installed which will not render correctly given the available APIVersions.

Have you considered the proposal in #6680? What's your current use case, and would that approach solve it for you?

jlegrone:
I believe the issue and solution described in #6680 is a little different. That proposal would fix template rendering when a chart also creates each CRD that it uses, however that won't work for charts that simply create custom resources and not their definitions (this is probably the more common of the two use cases).

> We still want to catch validation errors with malformed templates to ensure the user won't run into validation errors when they invoke `helm install`.

This PR appears to only skip returning the specific error for a missing apiVersion:

https://github.com/helm/helm/blob/b57c8967f33df85330c9e7d0007f8b3663178885/pkg/releaseutil/manifest_sorter.go#L132-L134

If clientOnly is explicitly set to false via the `--validate` flag, then we get the full error checking. This seems like the best of both worlds from an end-user perspective (and gives us the same behavior as `helm template` in v2).

jlegrone:
That said #6680 is also valid and would require some additional work to allow template to succeed when `--validate` is passed.

bacongobbler:

lminaudier:
Reopened. No problem Matthew!

jlegrone:
@bacongobbler FYI it is possible to switch the target branch of an existing pull request and retain the existing review comments/discussion.

bacongobbler:
Thank you for the heads up. I found out about that new feature after we deleted the branch. :disappointed: 

Once the branch has been deleted, there is no way to switch the target branch and re-open the PR.

Issue #6708, opened 2019-10-19
Regression for numeric Value type in Helm v2.15 breaking templates


armab:
The regression found in CI/CD nightly builds when relying on latest Helm which is now `v2.15` (since Friday? :trollface:)

In new Helm `v2.15` there seems to be some changes in values types.

### Helm Values:
```yaml
replicas: 3
```

### Helm Template:
```go
$replicas := int (.Values.replicas)
```
### Will produce:
`0` in `v2.15`
`3` in `v2.14`

---

After closer look, `.Values.replicas` is now of type `json.Number` in `v2.15`, previously was `float64` in `v2.14`. This means some previously worked cast/conversions in many user's templates and charts would fail.
For reference:
https://github.com/helm/charts/search?q=int+.Values.&unscoped_q=int+.Values
https://github.com/helm/charts/search?q=until&unscoped_q=until

### Workaround
Workaround for Helm `v2.15` templates is to cast to string first before converting to int:
```diff go
- $replicas := int (.Values.replicas)
+ $replicas := int (toString (.Values.replicas))
```

### Output of `helm version`:
`v2.15.0`
### Cloud Provider/Platform (AKS, GKE, Minikube etc.): 
`minikube`

Fixes https://github.com/StackStorm/stackstorm-ha/issues/89
Closes https://github.com/StackStorm/stackstorm-ha/pull/91

bacongobbler:
cc'ing @icanhazbroccoli as it looks like https://github.com/helm/helm/pull/6010 caused a regression. Perhaps we need to update `int` and other functions to accept `json.Number`?

icanhazbroccoli:
@bacongobbler indeed, it's a miss from my side. Mind if I start working on a fix?

bacongobbler:
Yes please.

Stono:
Was just going to report this too!

```
❯ helm template . --execute templates/test.yaml --set=replicas=1
---
# Source: platform-helm-at-service/templates/test.yaml

original: 1
int: 0
intOnNumber: 1
```

```
original: {{ .Values.replicas }}
int: {{ int .Values.replicas }}
intOnNumber: {{ int 1 }}
```

bacongobbler:
fixed in #6749.

armab:
:tada: 
Thanks for a quick fix!

Hopefully we'll see `v2.15.1` release soonish.

hickeyma:
`v2.15.1` released.

rgl:
[...]

PR #6709, opened 2019-10-19
Fixing regression for numeric values refs #6708

osdrv:

This commit is a fix for the regression introduced in 2.15
with an attempt to carry all numeric values in json.Number container
(see https://github.com/helm/helm/issues/6708 for more details).

The fix overloads a series of Helm template functions:
  * int64
  * int
  * float64
  * add1
  * add
  * sub
  * div
  * mod
  * mul
  * max
  * min
  * ceil
  * floor
  * round
  * until
  * untilStep
  * splitn
  * abbrev
  * abbrevboth
  * trunc
  * substr
  * repeat
  * randAlphaNum
  * randAlpha
  * randAscii
  * randNumeric
  * wrap
  * wrapWith
  * indent
  * nindent
  * plural
  * slice
  * eq
  * gt
  * ge
  * lt
  * le
  * ne

The PR might need some polishing and refactoring: this is very much a first working fix.

**If applicable**:

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -96,7 +96,7 @@ func FuncMap() template.FuncMap {
 		f[k] = v
 	}
 
-	return f
+	return OverloadJsonNumberFuncs(f)
 }
 
 // Render takes a chart, optional values, and value overrides, and attempts to render the Go templates.
--- a/pkg/engine/engine_test.go
+++ b/pkg/engine/engine_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package engine
 
 import (
+	"encoding/json"
 	"fmt"
 	"sync"
 	"testing"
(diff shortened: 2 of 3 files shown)


icanhazbroccoli:
Hey folks, my progress on this fix is pretty humble yet: the majority of numeric functions seems to have been fixed, but I'm still struggling implementing a workaround for comparison functions without a full re-definition of these monsters (https://golang.org/src/text/template/funcs.go#L441). I need to think over how these functions could be overloaded. In the meantime, if there are any hints on how it could be wrapped nicely, I'd be super happy to hear these. Thanks in advance!

icanhazbroccoli:
My progress so far:
* numeric conversion functionality has been fixed: functions like int, int64, float64 etc are working (fixes issues like: --set=replicas=3 which is known to set the value to 0 in 2.15.0)
* comparison functions are still broken for numeric arguments (only fixed for float64): the original builtins from text/template are context-dependant, I need to make overloaded functions to be smarter (TODO).

I am planning to work on fixing comparison functions and extending pkg/engine test suite. An ETA for this operation: a couple of days.

In the meantime, I'd really appreciate an early code review and feedback: it's super important for me to realise the core team has no strict objections against this approach. If there are objections, another rational solution would be to fully revert my earlier Number-converting patch so 2.15 can be unscrewed.

icanhazbroccoli:
Recent update on the fix: known reported issues seem to be addressed and fixed by this PR. Waiting for a review from the core team to figure out the next action points.

icanhazbroccoli:
https://github.com/helm/helm/issues/6721 might be related to the same regression and it has to be tested carefully before this PR could be counted safe to merge.

bacongobbler (commented):
thank you for the changes as well as the tests. My concern lies with having to overload every function that involves a numerical value, including any new functions that are introduced by sprig. Can we cast the `json.Number` to its native type before passing it as a value? That might prevent the need to wrap every function with a call to cast a `json.Number` to the expected type.

Given the changes involved in this PR, I'd be more inclined to revert the original commit than to introduce this code, and carry forward with a better fix in Helm 3. What do you think about that approach?

icanhazbroccoli:
@bacongobbler I totally agree with you. Indeed, this looks like an overload madness to me as well: a small change that looked good turned an absolute evil. I'd like to ask for 1 more evening of hacking to see if I can play with values the way you suggested above. If it doesn't work, I'm in favour of reverting the original commit. I'll update this PR shortly.

icanhazbroccoli:
Closed in favor of https://github.com/helm/helm/pull/6749

PR #6717, opened 2019-10-21
feat: (6715) Add --environment argument to helm template

oforero:

This is an initial implementation for 'refs#6715', I am sure it will require some improvements since I have only patched the `helm template` command to take the parameter, but I tested this and it generates the templates correctly for both normal and umbrella charts.

I think the change is minimal and the flexibility will be useful in many situations, adopting extra tools is not always feasible in many institutions.

This patch adds a parameter to the template command, the parameter is used only if the chart contains a file with that name, in which case the contents of it will use to override those in the `values.yaml`

This is WIP, if it is deemed as a good addition I am will write the missing pieces and bring any rough edge up to the necessary level for inclusion. It will close 'closes#6715'


bacongobbler:
Hi @oforero, unfortunately we cannot accept new features into Helm 2. As per [the Helm 2.15.0 release announcement](https://github.com/helm/helm/releases/tag/v2.15.0), 2.15.0 was the final feature release for Helm 2. We are asking users to please consider PR'ing them against Helm 3 instead at this time. Thank you!

PR #6726, opened 2019-10-21
Add support for a git url as a repository

rally25rs:

This PR adds support for using a git URL as a `repository` in `requirements.yaml`.

I thought this would be a nice-to-have feature for organizations that house their helm charts in a git repository and would like to maintain private access to those charts.

The format supported is:

```
dependencies:
- name: "{dependency name}"
  repository: "git:{git repo url}"
  version: "{git ref name}"
```
where:
* `{git repo url}` is any url type that a `git clone ...` command would accept (an ssh location, an https url, etc.).
* {git ref name} is an existing tag or branch name on the repo (a commit sha cannot be used at this time)

For example:

```
dependencies:
- name: "a"
  repository: "git:https://github.com/rally25rs/helm-test-chart.git"
  version: "master"
```

(the above `rally25rs/helm-test-chart` does exist. feel free to use it for testing if you want.)

This is my first time coding in Go, so please feel free to comment on the code and provide any refactoring suggestions or ways to do things better.

I wasn't sure if there was a way to add more detailed docs outside what is auto generated by `make docs`. If there is, then I could add more details for reading on the web; but I didn't want to add too much to the command output.

Since other package managers like `npm/yarn`, `ruby bundler`, `glide`, etc... all support github URLs, I thought this might be a nice to have feature. If this isn't the direction that the `helm` community wants to take for dependency management, we can reject this PR. I won't be offended :) It was still a fun learning experience to do something with Go, and this only took a day and a half to whip up. I actually made this PR instead of trying to build my own private custom registry in the traditional helm way, because that seemed more difficult and would likely require some process to monitor our github organization for changes, check them out to a central location, run the helm command to build the index.yaml, and push it all to a central server that only our employees have access to. Git already has SSH to secure it, so including our private charts in this way seemed a lot easier.

Thanks for your consideration and for taking the time to review this!

**Implementation Details**

The implementation is a simplified take on what `yarn` does for git dependencies https://github.com/yarnpkg/yarn/blob/master/src/util/git.js (I'm also a `yarn` contributor so was familiar with their code base) 

What it basically does is, if it find a `registry` that starts with `git:` then:
* run a `git ls-remote` sub process to get the tag and branch names for the repo.
* verify that the `version` is a valid tag or branch name.
* create a temp directory.
* run a (shallow) `git clone` sub process to fetch the repo at the specified branch/tag into the temp dir
* Treat the cloned git repo similar to a `file:///path/to/temp/dir` style requirement; use `chartutils` to load that directory (which in turn applied the logic for filtering the files through `.helmignore`) and archives it to `charts/`
* Delete the temp dir.

**Caveats / Known Limitations**

* Singe this spawns `git` as a child process, `git` has to be available on the system and in the path.
* There is currently no handling for forwarding `stdin` to the git child process, so if git asks for input, helm will likely appear to hang as git pries to prompt for input. It will cause an issue if you are trying to reference a private repo with an SSL cert that needs a password, or with an https url that requires username/password authentication. (`yarn` package manager, `ansible`, and other tools have this same issue). I couldn't think of a great way to handle this, since you don't really want to put all of the git command's `stdout` to the screen; just any prompts it might display, but there isn't really an easy way to differentiate.

**If applicable**:
- [X] this PR contains unit tests (I added a few where I could)
- [X] this PR has been tested for backwards compatibility (since no existing registry urls would have started with `git:` there shouldn't be any problem with existing registries)

**Future Possibilities**

I intentionally left this functionality simple for now to see how well the helm community would receive it. If this PR is accepted, I can see some room for improvements in future PRs:

* Handle having a chart in a sub-directory of a git repo (currently Chart.yaml must be in the git repo root)
* Better handle git command stdin/out - Simply adding `GIT_TERMINAL_PROMPT=0` to the env when running the sub commands may help https://github.blog/2015-02-06-git-2-3-has-been-released/#the-credential-subsystem-is-now-friendlier-to-scripting
* I originally wanted to write the current commit sha for the branch/tag to the `requirements.lock` file, so that if branches or tags move forward, you would still be locked to the last commit. However  the `git clone --branch {name}` command has no way to shallow checkout a single commit, only a branch or tag (`git archive` can, but github doesn't support that command). It'd be nice to lock to a specific commit. This could be done with a full `git clone` (non shallow) then a `git checkout` but I wanted to avoid the overhead of the full clone.


rally25rs:
Closing because this made a gigantic diff unexpectedly... will reopen with the actual correct changes. (oops!)

bacongobbler:
We just switched over the branch. Helm 3 is now the new `master` branch, so any development on Helm 2 should go towards the `dev-v2` branch going forward.

That being said, I will note that we are no longer accepting feature requests for Helm 2. Helm 2.15.0 was the last feature release. I would encourage you to implement this as a Helm 3 feature. Thanks! :)

rally25rs:
Yeah that must be what happened. I was so confused... I just forked like 3 days ago, and even github was telling me that my `master` was like 1,000 commits "ahead of helm/master" which made no sense... I ended up having to re-fork the whole repo and am reapplying my changes.

Issue #6665, opened 2019-10-14
Helm v3 can not template CRD


anikin-aa:
Hi there !

Been playing around with helmv3 and found next problem.

i've got a CRD template that locates in `templates` folder and named as `crd.yaml` with next content
```
{{ if .Values.crd.create }}
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
metadata:
  name: testservices.test.com
spec:
  scope: Namespaced
  group: test.com
  names:
    kind: testService
    listKind: testServiceList
    plural: testservices
    shortNames:
    - test
    - tests
    singular: testservice
  version: v1
  versions:
  - name: v1
    served: true
    storage: true
{{ end }}
```

Output of `helm version`:
```
$ helm version
version.BuildInfo{Version:"v3.0.0-beta.4", GitCommit:"7ffc879f137bd3a69eea53349b01f05e3d1d2385", GitTreeState:"dirty", GoVersion:"go1.13.1"}
```

Output of `kubectl version`:
```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"11", GitVersion:"v1.11.0", GitCommit:"91e7b4fd31fcd3d5f436da26c980becec37ceefe", GitTreeState:"clean", BuildDate:"2018-06-27T20:17:28Z", GoVersion:"go1.10.2", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"11+", GitVersion:"v1.11.0+d4cacc0", GitCommit:"d4cacc0", GitTreeState:"clean", BuildDate:"2019-01-29T15:19:24Z", GoVersion:"go1.10.3", Compiler:"gc", Platform:"linux/amd64"}
```

`helm template . --kubeconfig /home/me/.kube/config` fails with next error:
```
Error: apiVersion "apiextensions.k8s.io/v1beta1" in crd.yaml is not available
``` 

`kubectl` says, that i can create CRD on the cluster:
```
$ kubectl --kubeconfig /home/me/.kube/config auth can-i create crd
yes
```
 
templating of non cluster wide resources are working fine

Cloud Provider/Platform (AKS, GKE, Minikube etc.): 
Openshift 3.11

anikin-aa:
https://github.com/helm/helm/pull/6243

Need to place all of crds under `crds/` folder

hickeyma:
[...]

Issue #6663, opened 2019-10-14
`helm template` command with `--validate` option fails when the resource exists


yoichiwo7:
Output of `helm version`: v3.0.0-beta.4

Output of `kubectl version`: Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.1", GitCommit:"d647ddbd755faf07169599a625faf302ffc34458", GitTreeState:"clean", BuildDate:"2019-10-02T17:01:15Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"14+", GitVersion:"v1.14.6-gke.2", GitCommit:"c9de33b5439df6e206d7ba646787c6ace92d737b", GitTreeState:"clean", BuildDate:"2019-09-06T18:30:33Z", GoVersion:"go1.12.9b4", Compiler:"gc", Platform:"linux/amd64"}

Cloud Provider/Platform (AKS, GKE, Minikube etc.): GKE

#### Problem

helm v3.0.0-beta.4 fails to generate template output when executing `helm template` command with `--validate` option if the same kubernetes resources already exits. 

The error message is like following:

`Error: rendered manifests contain a resource that already exists. Unable to continue with install: existing resource conflict: kind: ServiceAccount, namespace: default, name: hello-world`

#### Expected behavior

`helm template` command with `--validate` must succeed to generate the template output even if the same kubernetes resources already exists.

#### Reproduce the problem

You can run following scripts to reproduce the problem. `helm v3.0.0-beta.4` fails and `helm v3.0.0-beta.3` succeeds.

```bash
#!/bin/bash
set -x

# Check version
helm version

# helm init
helm create hello

# 1st
helm template --validate hello-world hello  > hello.yaml
kubectl apply -f hello.yaml
sleep 10
kubectl get all

# 2nd (Succeeds on beta3, fails on beta4)
helm template --validate hello-world hello  > hello2.yaml

# cleanup
kubectl delete -f hello.yaml
rm -rf hello hello*.yaml
```

The result of the `helm v3.0.0-beta.4`:

```
+ helm version
version.BuildInfo{Version:"v3.0.0-beta.4", GitCommit:"7ffc879f137bd3a69eea53349b01f05e3d1d2385", GitTreeState:"dirty", GoVersion:"go1.13.1"}
+ helm create hello
Creating hello
+ helm template --validate hello-world hello
+ kubectl apply -f hello.yaml
[...]

PR #6729, opened 2019-10-21
Skip api version checks when install action is configured in client mode only

### LINKED ISSUE #6505: helm3 template fails when there are custom resource templates

lminaudier:
closes #6505

Note: Reopening #6701 after switch from `dev-v3` to `master` base branch. Discussion started in that PR.

--- 

Original PR content

Hi there,

With helm 3, when an install action is run in client-mode only (`install` without `--validate`) and if your chart contains custom resources you end up hitting error that look like that.

```
apiVersion "example.com/v1" in my-app/templates/cutoms.yaml is not available
```

This PR attempt to workaround the issue by skipping entirely the check if we are in client mode as we can't be aware of any custom resource definition present on the remote cluster in advance as we don't interact with it in such mode.

Thoughts?

The error hit is similar to the one described in #6680 

**If applicable**:

### DIFF
--- a/pkg/releaseutil/manifest_sorter.go
+++ b/pkg/releaseutil/manifest_sorter.go
@@ -74,7 +74,7 @@ var events = map[string]release.HookEvent{
 //
 // Files that do not parse into the expected format are simply placed into a map and
 // returned.
-func SortManifests(files map[string]string, apis chartutil.VersionSet, sort KindSortOrder) ([]*release.Hook, []Manifest, error) {
+func SortManifests(files map[string]string, apis chartutil.VersionSet, sort KindSortOrder, clientOnly bool) ([]*release.Hook, []Manifest, error) {
 	result := &result{}
 
 	for filePath, c := range files {
@@ -95,7 +95,7 @@ func SortManifests(files map[string]string, apis chartutil.VersionSet, sort Kind
 			apis:    apis,
 		}
 
-		if err := manifestFile.sort(result); err != nil {
+		if err := manifestFile.sort(result, clientOnly); err != nil {
 			return result.hooks, result.generic, err
 		}
 	}
@@ -122,14 +122,14 @@ func SortManifests(files map[string]string, apis chartutil.VersionSet, sort Kind
 //  metadata:
 // 		annotations:
 // 			helm.sh/hook-delete-policy: hook-succeeded
-func (file *manifestFile) sort(result *result) error {
+func (file *manifestFile) sort(result *result, clientOnly bool) error {
 	for _, m := range file.entries {
 		var entry SimpleHead
 		if err := yaml.Unmarshal([]byte(m), &entry); err != nil {
 			return errors.Wrapf(err, "YAML parse error on %s", file.path)
 		}
 
-		if entry.Version != "" && !file.apis.Has(entry.Version) {
+		if !clientOnly && (entry.Version != "" && !file.apis.Has(entry.Version)) {
 			return errors.Errorf("apiVersion %q in %s is not available", entry.Version, file.path)
 		}
 
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -211,7 +211,7 @@ func (i *Install) Run(chrt *chart.Chart, vals map[string]interface{}) (*release.
 	rel := i.createRelease(chrt, vals)
 
 	var manifestDoc *bytes.Buffer
-	rel.Hooks, manifestDoc, rel.Info.Notes, err = i.cfg.renderResources(chrt, valuesToRender, i.OutputDir, i.SubNotes)
+	rel.Hooks, manifestDoc, rel.Info.Notes, err = i.cfg.renderResources(chrt, valuesToRender, i.OutputDir, i.SubNotes, i.ClientOnly)
 	// Even for errors, attach this if available
 	if manifestDoc != nil {
 		rel.Manifest = manifestDoc.String()
@@ -409,7 +409,7 @@ func (i *Install) replaceRelease(rel *release.Release) error {
 }
 
 // renderResources renders the templates in a chart
-func (c *Configuration) renderResources(ch *chart.Chart, values chartutil.Values, outputDir string, subNotes bool) ([]*release.Hook, *bytes.Buffer, string, error) {
+func (c *Configuration) renderResources(ch *chart.Chart, values chartutil.Values, outputDir string, subNotes bool, clientOnly bool) ([]*release.Hook, *bytes.Buffer, string, error) {
 	hs := []*release.Hook{}
 	b := bytes.NewBuffer(nil)
 
@@ -452,7 +452,7 @@ func (c *Configuration) renderResources(ch *chart.Chart, values chartutil.Values
 	// Sort hooks, manifests, and partials. Only hooks and manifests are returned,
 	// as partials are not used after renderer.Render. Empty manifests are also
 	// removed here.
-	hs, manifests, err := releaseutil.SortManifests(files, caps.APIVersions, releaseutil.InstallOrder)
+	hs, manifests, err := releaseutil.SortManifests(files, caps.APIVersions, releaseutil.InstallOrder, clientOnly)
 	if err != nil {
 		// By catching parse errors here, we can prevent bogus releases from going
 		// to Kubernetes.
--- a/pkg/action/uninstall.go
+++ b/pkg/action/uninstall.go
@@ -170,7 +170,8 @@ func (u *Uninstall) deleteRelease(rel *release.Release) (string, []error) {
 	}
 
 	manifests := releaseutil.SplitManifests(rel.Manifest)
-	_, files, err := releaseutil.SortManifests(manifests, caps.APIVersions, releaseutil.UninstallOrder)
+
+	_, files, err := releaseutil.SortManifests(manifests, caps.APIVersions, releaseutil.UninstallOrder, false)
 	if err != nil {
 		// We could instead just delete everything in no particular order.
 		// FIXME: One way to delete at this point would be to try a label-based
--- a/pkg/action/upgrade.go
+++ b/pkg/action/upgrade.go
@@ -160,7 +160,7 @@ func (u *Upgrade) prepareUpgrade(name string, chart *chart.Chart, vals map[strin
 		return nil, nil, err
 	}
 
-	hooks, manifestDoc, notesTxt, err := u.cfg.renderResources(chart, valuesToRender, "", u.SubNotes)
+	hooks, manifestDoc, notesTxt, err := u.cfg.renderResources(chart, valuesToRender, "", u.SubNotes, false)
 	if err != nil {
 		return nil, nil, err
 	}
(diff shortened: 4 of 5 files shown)


jlegrone:
@lminaudier please update PR description to note this closes https://github.com/helm/helm/issues/6505. Thanks!

mumoshu:
@bacongobbler @jlegrone Hey! Excuse me but can we reconsider this to be merged before 3.0.0 GA? This is breaking a helm plugin badly: https://github.com/databus23/helm-diff/issues/158

pkg/releaseutil/manifest_sorter.go
```diff
@@ -74,7 +74,7 @@ var events = map[string]release.HookEvent{
 //
 // Files that do not parse into the expected format are simply placed into a map and
 // returned.
-func SortManifests(files map[string]string, apis chartutil.VersionSet, sort KindSortOrder) ([]*release.Hook, []Manifest, error) {
+func SortManifests(files map[string]string, apis chartutil.VersionSet, sort KindSortOrder, clientOnly bool) ([]*release.Hook, []Manifest, error) {
```
thomastaylor312: This parameter should probably be called `ignoreVersions` or `validate` to more clearly indicate what it is doing

thomastaylor312:
My one worry here is that the `--api-versions` flag will now have no value because it was only used for cases when we are in client only mode. So we should remove that if we go this direction

thomastaylor312:
Oh, and as a note, I am in favor of pulling this into Helm 3.0.0 and have pinged other maintainers to weigh in

bacongobbler:
Recalling a bit of context from the previous issue in https://github.com/helm/helm/pull/6701#issuecomment-543875223...

@bacongobbler said:

> Hmm. I don't think passing a boolean across several API layers just to disable a check is a great design. The validation is in place to ensure that CRDs with malformed/unavailable APIVersions will be still be reported to the user. We still want to catch validation errors with malformed templates to ensure the user won't run into validation errors when they invoke helm install.

> Have you considered the proposal in #6680? What's your current use case, and would that approach solve it for you?

@jlegrone said:

> I believe the issue and solution described in #6680 is a little different. That proposal would fix template rendering when a chart also creates each CRD that it uses, however that won't work for charts that simply create custom resources and not their definitions (this is probably the more common of the two use cases).

> We still want to catch validation errors with malformed templates to ensure the user won't run into validation errors when they invoke `helm install`.

> This PR appears to only skip returning the specific error for a missing apiVersion:

https://github.com/helm/helm/blob/b57c8967f33df85330c9e7d0007f8b3663178885/pkg/releaseutil/manifest_sorter.go#L132-L134

> If clientOnly is explicitly set to false via the `--validate` flag, then we get the full error checking. This seems like the best of both worlds from an end-user perspective (and gives us the same behavior as `helm template` in v2).

> That said #6680 is also valid and would require some additional work to allow template to succeed when --validate is passed.

bacongobbler:
I'd like to return back to the use case this proposal is trying to solve before coming up with a solution.

Here's the use case I'm considering:

1. I have a chart
1. I intend to register a CRD with apiVersion `bacongobbler.com/v1`
1. The chart will consume this new API by creating custom resources

There are two situations we need to handle in `helm template`:

1. I have templates in my chart that are valid. Please render me to stdout.
1. I have templates in my chart that are invalid. Please return an error when the chart is malformed.

With this approach, we are actively disabling a check. If I create a chart that has a custom resource consuming apiVersion `bacongobbler.com/v2` (`v2` instead of `v1`),  `helm template` will succeed and will render the templates, signalling to the user that the chart can be installed, which it can't, because there is no apiVersion that supports `bacongobbler.com/v2`.

This is why I asked about considering the proposal in #6680. As the `crds/` feature was described in #6243:

> If a CRD is already found in the cluster, Helm will skip installing that CRD and will log this to the debug log. It will neither error out nor see if there are any differences between its version and the existing version.

If that's the case, then it should be considered OK for each chart to declare the CRDs they rely upon. If those CRDs are present in `crds/`, then we can add those CRDs to the registered APIs and check the templates for validity.

In other words, the use case described by @jlegrone earlier... 

> however that won't work for charts that simply create custom resources and not their definitions

...could be considered an error condition: if a chart declares that they intend to use a particular CRD, but the chart itself does not declare that CRD, then `helm template` *should* fail unless the user provides the `--validate` flag to check if the Kubernetes API server has that CRD registered.

From reading #6243, I am under the impression that

1. charts that rely on installing a custom resource should *always* have their CRDs declared in the `crds/` directory
1. Charts that do NOT have that custom resource in the `crds/` directory should be considered a special case, and users should use the `--validate` flag instead.

bacongobbler:
The other thought with this proposal is that we can remove the `--api-versions` flag from Helm 3 entirely. It was only necessary in Helm 2 because of the same use case: `helm template` has no way of determining what CRDs have been registered, so we needed a way to register them for charts with CRDs to render correctly.

mumoshu:
@bacongobbler Thanks for summarizing it!

I now believe you're absolutely correct on that we shouldn't disable the validation.

#6680 seems the right approach. But can we register all the CRDs present in the `templates/` rather than only those in `templates/crds/`? That seems to fix #6505 for existing "valid" charts that worked in v2.

mumoshu:
Also, if we don't go the way disabling the validation at all, #6663 needs to be fixed as well.

That is, even after #6680 `helm template --validate=true` can fail on a chart that was valid in v2 due to #6663.

mumoshu:
Lastly, even though considering ClientOnly = `--validate`, like this PR does, turned wrong, we should at least allow disabling the validation and the api-versions check at all when required.

It doesn't work today:

```
$ helm3b5 template --validate=false ./mychart
Error: apiVersion "apiextensions.k8s.io/v1beta1" in mychart/templates/crd.yaml is not available
```

To me this seems to, possibly, imply that we need to separate the flags for toggling ClientOnly and the validation.

mumoshu:
After all these comments addressed, I think `helm template` should be documented as:

1. `helm template [--validate=true]` should be used when you want it to fail when the result has any reference to CRDs not in `crds/` or the cluster.
   -  I believe #6663 needs to be fixed for this use-case. Otherwise running `helm template --validate=true | kubectl apply -f` twice always fail on the second run.
   - I believe https://github.com/helm/helm/pull/6729#issuecomment-547231609 should be addressed as well for this use-case. Otherwise charts that passed `helm template` in v2 can't pass `helm template` in v3. An example of such chart is one that has both the definition and the instance for a custom resource.
2. `helm template --validate=false` should be used when you want to disable the validation
   - This seems to not work today as said in https://github.com/helm/helm/pull/6729#issuecomment-547248883. `--validate=false` doesn't actually disable all the validations.
3. `helm template --client-only --api-versions API_VERSION [--validate=true]`  should be used when you want it to fail when the result has any reference to CRDs not in `crds/`.
   - If we don't need this, as @bacongobbler suggested in https://github.com/helm/helm/pull/6729#issuecomment-547199504, `--api-versions` should be removed as unnecessary. But then `CilentOnly bool` in the codebase should be changed to something like `Validate bool`. It seems like the source of confusion here was we don't really need/have a client-only validation.

mumoshu:
> My one worry here is that the --api-versions flag will now have no value because it was only used for cases when we are in client only mode. So we should remove that if we go this direction

@thomastaylor312 Could be! I've mentioned about that in the third point in https://github.com/helm/helm/pull/6729#issuecomment-547249311.

But please note that, if we were to remove `--api-versions` and don't have any better alternatives, (I think) we would still need a way to disable api versions check at all instead. `--validate=false` as of today doesn't work for that. I've mentioned about that in the second point in my prev comment.

mumoshu:
> > however that won't work for charts that simply create custom resources and not their definitions
> 
> ...could be considered an error condition: if a chart declares that they intend to use a particular CRD, but the chart itself does not declare that CRD, then helm template should fail.

@bacongobbler I've reread your comment and I think it does break existing charts like [official charts that creates `ServiceMonitor`s](https://github.com/helm/charts/search?q=ServiceMonitor&unscoped_q=ServiceMonitor) and community charts like [podinfo](https://github.com/weaveworks/flagger/tree/master/charts/podinfo) and [monochart](https://github.com/cloudposse/charts/tree/master/incubator/monochart).

databus23:
> If every chart declares CRDs along with their custom resources, then the need for the --api-versions flag disappears, as the CRD will be registered with the kube client and we can validate the templates against it.
I think this is only feasible in certain cases where a CRD is really local to a specific chart. But as CRD are be now the standard way to extend Kubernetes there is an ever growing list of CRDs that  are used more like general extensions to kubernetes. Something that is just expected the be there. As mentioned by @mumoshu we are using the prometheus-operator in our Organization and every chart is expected to ship its on alert rules, service monitors etc. It would be very cumbersome if every of our 50+ chart would need to ship 3 crds just so that `helm template` can be used.
I don't pretend to know anything about Istio or Knative but it is my understanding that those extensions  could be other popular examples. If I'm not mistaken those frameworks  rely heavily on a plethora of custom resource definitions that will be used by other charts then the "istio" chart.

mumoshu:
I ended up submitting https://github.com/helm/helm/pull/6811 as an alternative to this PR.

mumoshu:
Also, `--validate` with the client-only mode seems broken(?) as it doesn't even see `apiextensions.k8s.io/v1beta1` as a known api type. See https://github.com/helm/helm/issues/6665 for info.

The reporter closed it as resolved as the error disappears with the current `--validate`(do validate in the non-client-only mode), but that doesn't make sense in this context.

So probably we should fix #6665 as well if we do provide the option to run client-only validations.

bacongobbler:
Charts that create ServiceMonitors rely on the server (Kubernetes) having prometheus-operator installed. In that case, they should be using the `--validate` flag to check if the CRD they are attempting to install is present. That is one of those "special cases" I was referring to earlier. Same goes for `podinfo` and `monochart`; they rely on the presence of flagger being installed server-side, and should have to check with the server to ensure they will install correctly.

Issue #4618, opened 2018-09-10
Helm seems to not use its local charts cache


mboret:
Hi, I'm surprise to discover that helm is not using his local cache folder(.helm/cache/archives/) during an install/upgrade process. 

We faced this issue after our chartmuseum repository was unavailable generating an I/O timeout error when we tried to perform an helm upgrade. 

The weird thing is the chart was already in .helm/cache/archives/

I was thinking that in a upgrade/install process, helm looks inside this archives folder, check if the package exists, if yes, use it, otherwise try to download it. 

It seems that the reality is different. Can someone confirme me this behavior and if it's normal?

Thanks.

Output of `helm version`:

Client: &version.Version{SemVer:"v2.8.0", GitCommit:"14af25f1de6832228539259b821949d20069a222", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.8.1", GitCommit:"6af75a8fd72e2aa18a2b278cfe5c7a1c5feca7f2", GitTreeState:"clean"}

Output of `kubectl version`:

Client Version: version.Info{Major:"1", Minor:"9", GitVersion:"v1.9.3", GitCommit:"d2835416544f298c919e2ead3be3d0864b52323b", GitTreeState:"clean", BuildDate:"2018-02-07T12:22:21Z", GoVersion:"go1.9.2", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"8", GitVersion:"v1.8.6", GitCommit:"6260bb08c46c31eea6cb538b34a9ceb3e406689c", GitTreeState:"clean", BuildDate:"2017-12-21T06:23:29Z", GoVersion:"go1.8.3", Compiler:"gc", Platform:"linux/amd64"}

bacongobbler:
Yes, that is the current behaviour, but I wouldn't call it normal. :)

As it stands right now, `helm install stable/wordpress` will look up the chart based on what's in `$HELM_HOME/repository/cache/stable-index.yaml`, find the URL of the chart referenced in the repository index (`wordpress` in this example) and fetches it. however, we should probably check if that chart has already been fetched before and it exists in the cache before trying to re-fetch it.

Filing as a good first feature for someone to tackle!

mattfarina:
I was just talking with @technosophos about this. The way the cache works doesn't work with repo/chart uniqueness. So, if I have two charts with the same name and version but different content which one will be in the cache.

Helm charts have a digest. We could use that in a few interesting ways...

[...]

PR #6733, opened 2019-10-22
Chart cache

razor-x:
This is a first step to implementing https://github.com/helm/helm/issues/4618

Please note I have not done any real Go development, so my questions may seem very basic. Ideally I can get some validation on if I'm putting things logically in the right place and can iterate from there.

Anyone is welcome to contribute code or help to this PR. I will gladly give contrib access.

Basic plan (does this make sense?):

0. Ensure there is a `helmpath` method for `$helmChartCache` that creates the cache directory
1. Add a `WithCache` function to `getter.go` that sets `cachePath = $helmChartCache/$digest`
2. Add some kind of option in Helm that would be passed as an option to `ChartDownloader` to toggle the cache: `type CacheStrategy int` (where to put this option?)
3. If this option is set, have the `ChartDownloader` function `DownloadTo` invoke `WithCache`, passing the `digest` from the chart manifest (if available, otherwise do not invoke).
4. Implement (somewhere) the actual cache wrapping behaviour when `opts.cachePath` has been set (save to `cachePath` whenever it downloads something, check `cachePath` before it downloads something)
5. Ensure this does not interfere with the verification step: it should happen every time if requested, even on files returned from the cache.
6. What path should be returned by the Getter? Should it always be the cached file or will the cached file be copied to wherever it would have normally downloaded it?

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+indent_style = tab
+indent_size = 2
--- a/pkg/getter/getter.go
+++ b/pkg/getter/getter.go
@@ -35,6 +35,7 @@ type options struct {
 	username  string
 	password  string
 	userAgent string
+	cachePath string
 }
 
 // Option allows specifying various settings configurable by the user for overriding the defaults
@@ -70,6 +71,14 @@ func WithTLSClientConfig(certFile, keyFile, caFile string) Option {
 		opts.certFile = certFile
 		opts.keyFile = keyFile
 		opts.caFile = caFile
+	}
+}
+
+// WithCache will wrap all HTTP requests through the local file system cache.
+// Charts are cached in a flat structure with name equal to the chart digest.
+func WithCache(digest string) Option {
+	return func(opts *options) {
+		opts.cachePath = path.Join(helmpath.ChartCachePath(), digest) // new?
 	}
 }
 


bacongobbler:
Thanks for the PR. Helm 3 is in the home stretch. We are in bug fix mode now. We can add features in minor releases so I am adding it to the 3.1.0 milestone so we can hopefully start reviewing it after the initial release. I just wanted to let you know it might be a little bit before we get to this.

To answer your questions:

1. There is a `helmpath` package which provides you with convenient functions to create and access common directories. Have a look at the [CachePath](https://github.com/helm/helm/blob/484d43913f97292648c867b56768775a55e4bba6/pkg/helmpath/home.go#L23) function, and have a search for `helmpath.CachePath` in the codebase for examples.
1. I don't think we should support multiple caching strategies. One that works is good enough. :)
1. See above
1. See above
1. yep. Files can be modified out-of-band, so verification should occur regardless if it was fetched or cached locally.
1. Not sure. What does it currently return?

hickeyma (commented):
@razor-x Helm 3 is now released, so it is a good time to restart this PR.  Can you rebase with latest code and push on with feedback from @bacongobbler  in https://github.com/helm/helm/pull/6733#issuecomment-545028846 ?

bacongobbler:
ping

razor-x:
Thanks for the update. Personally I haven't had much priority to work on this, and I'm not sure when that will change. I don't want to block this feature for anyone else looking for it, so the best I can say at the moment is that I will pick this up again when I find some time, but anyone with more urgency on this should feel free to continue from here.

bacongobbler:
Closing for now due to inactivity. Anyone willing to push this forward, please feel free to open a new PR. Thanks!

juniorz:
I can happily push this forward! I have opened a new PR #8831 with the proposal and current progress.

PR #6736, opened 2019-10-22
fix(pkg/downloader): Calculate digest from both Chart.lock & Chart.yaml deps

### LINKED ISSUE #6416: "Error: Chart.lock is out of sync with Chart.yaml" If Chart.yaml dependency has an alias

hangpark:
Re-open of #6655 
Fixes #6416 
Caused by #5874

To make digests include information about Chart.yaml dependencies, not only the lock file, digest calculation is changed to accept both contents. This terminates the `dep build` command if Chart.yaml dependencies have been updated so that `dep up` should be executed properly, to prevent downloading wrong versions or mismatched subcharts.

Note that previous Helm cannot know whether Chart.yaml dependencies were changed or not since the Chart.lock's digest is calculated by only Chart.lock contents, which don't include information about SemVer ranges and extra dependency fields such as aliases, conditions, and tags. Especially, SemVer can be written as a version range in Chart.yaml, but Chart.lock has the specific, resolved version of that range.

This PR is the re-open version of #6655. I rebased commits from the previous PR and tried to apply @mattfarina's arguments. Let me know if I'm going the wrong way.

**If applicable:**


internal/resolver/resolver_test.go
```diff
@@ -156,20 +156,36 @@ func TestResolve(t *testing.T) {
 }
 
 func TestHashReq(t *testing.T) {
-	expect := "sha256:d661820b01ed7bcf26eed8f01cf16380e0a76326ba33058d3150f919d9b15bc0"
+	expect := "sha256:96045f8dfd8769a3a1af6202449fa620c900992277803e799fe31c3897f179e8"
 	req := []*chart.Dependency{
-		{Name: "alpine", Version: "0.1.0", Repository: "http://localhost:8879/charts"},
+		{Name: "alpine", Version: "^0.1.0", Repository: "http://localhost:8879/charts"},
```
bacongobbler: Can we please keep this as-is? I'd rather we keep the existing tests in place and add new tests to the suite rather than replacing existing test cases.

bacongobbler (changes requested):

bacongobbler:
Just tacking on a fly-by comment here. I haven't dove deep into the change yet. In general we try to keep existing test cases in place instead of replacing them with others. Would you mind amending this PR to keep the existing test cases for `TestHashReq` in place? Thanks!

hangpark:
@bacongobbler I restored the original test with my new tests and used `t.Run()`! But note that the expected digest is changed as `HasReq()` implementation changed.

bacongobbler:
That's fine. Thanks!

internal/resolver/resolver.go
```diff
@@ -148,8 +148,8 @@ func (r *Resolver) Resolve(reqs []*chart.Dependency, repoNames map[string]string
 //
 // This should be used only to compare against another hash generated by this
 // function.
-func HashReq(req []*chart.Dependency) (string, error) {
-	data, err := json.Marshal(req)
+func HashReq(req []*chart.Dependency, lock []*chart.Dependency) (string, error) {
```
adamreese: nit: could be replaced with `func(req, lock []*chart.Dependency)`

adamreese (approved):
Spent some time manually testing and I think this looks good.

internal/resolver/resolver.go
hangpark: Fixed it!

bacongobbler (approved):

PR #6738, opened 2019-10-22
Support cross device rename

ysh7:
This PR includes changes mentioned on #5038 required for merging it.

closes #2998

PR #6750, opened 2019-10-22
Support cross device rename for helm v3

ysh7:
This PR includes changes on #6738 for Helm v3

/cc @thomastaylor312

### DIFF
--- a/pkg/downloader/manager.go
+++ b/pkg/downloader/manager.go
@@ -30,6 +30,7 @@ import (
 	"github.com/pkg/errors"
 	"sigs.k8s.io/yaml"
 
+	"helm.sh/helm/v3/internal/dep/fs"
 	"helm.sh/helm/v3/internal/resolver"
 	"helm.sh/helm/v3/internal/urlutil"
 	"helm.sh/helm/v3/pkg/chart"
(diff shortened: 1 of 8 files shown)


pkg/downloader/manager.go
```diff
@@ -30,6 +30,7 @@ import (
 	"github.com/pkg/errors"
 	"sigs.k8s.io/yaml"
 
+	"helm.sh/helm/v3/internal/dep/fs"
```
thomastaylor312: This should be 
```suggestion
	"helm.sh/helm/v3/internal/third_party/dep/fs"
```

thomastaylor312 (changes requested):
Thanks for opening the port @ysh7! One error in here that is causing the build failure

pkg/downloader/manager.go
ysh7: Change you asked for is done. Sorry this was very lame mistake on my part.

thomastaylor312:
@ysh7 Thanks for fixing that! Looks like you pulled in some commits that were already in master. Anyway you could rebase to remove those?

thomastaylor312 (approved):
This LGTM from my end. Pinging @mattfarina for a second review on this

thomastaylor312:
This slipped by our review of the queue for some reason. Pulling in to 3.0 as we are also pulling #6738 into 2.16

mattfarina (approved):

PR #6528, opened 2019-09-29
fix lint not failing in windows for bad template yamls

karuppiah7890:
Fixes #6527 
Port of #6523 (v2 fix)

**What this PR does / why we need it**:
It fixes lint not failing in windows for bad template yamls

**If applicable**:
- [ ] this PR contains documentation
- [ ] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

PR #6523, opened 2019-09-28
fix lint not failing in windows for bad template yamls

karuppiah7890:
**What this PR does / why we need it**:
Fixes #6418 

**Special notes for your reviewer**:
There is no test in this PR. One way this can be tested is - if we run tests on windows. As the issue is involved with calling the functions in `filepath` package which works differently in Linux/Unix and in Windows. 

I have manually tested this in Windows in a VM. Let me know if you find any issues. And I must say my confidence level on the PR is not great because of the lack of tests. But I was told we can address that separately in a different issue 😄 

**If applicable**:
- [ ] this PR contains documentation
- [ ] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

PR #6755, opened 2019-10-23
[WIP] Fix lint in windows

karuppiah7890:
**What this PR does / why we need it**:
Fixes #6418 
Twin of #6523 which got closed due to branch renaming

**If applicable**:
- [ ] this PR contains documentation
- [ ] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

Issue #6155, opened 2019-08-02
Support index.json in addition to index.yaml in repositories


mattfarina:
`index.yaml` files that grow large in size can use significant resources to parse. Some time ago I produced some benchmarks at https://github.com/mattfarina/yamlbench. This project was originally showcased by some at jFrog at the first helm summit.

As a feature addition, we could add an `index.json` file alongside the YAML file. JSON uses significantly fewer resources when being parsed. For older helm clients the `index.yaml` would still exist. Helm could check for the json file. If not present fall back to the `index.yaml` file. This would enable helm to work with older repositories.

This could go into both Helm v3 and v2.

bacongobbler:
SGTM. This definitely sounds like a feature addition where we could fallback to the original request.

karuppiah7890:
I would love to give this a try! Are there any special requirements other than what's mentioned above? 

In terms of code, I'm thinking - I'll check every place where index yaml file is used/parsed, and try to add support for json too if json file is present, or fallback to yaml. Write index.json related tests in all places which has tests regarding `index.yaml` files

karuppiah7890:
@bacongobbler @mattfarina Should we consider more options? Since this is an index and it will only be generated by programs and read by programs and not humans, is it possible to checkout other formats to store the index? To make the storage and processing more efficient

marckhouzam:
The completion code would greatly benefit from much faster parsing, although all it needs is the list of chart names. The completion currently uses `helm search repo` which is much too slow (when the stable repo is provisioned) in the context of completion.

I've proposed the creation of another cache file which would only list the charts as part of #6809.

It would be nice to consider this requirement when addressing the current issue, if my proposed new cache file is not approved.

karuppiah7890:
I'm starting off with implementing the creation of a `index.json` file along with `index.yaml` file when we run

`helm repo index .`

karuppiah7890:
what about --merge option?

if index.yaml is given, it will be converted to index object and then used to create both index.json and index.yaml - okay?

what about supporting merging with index.json ?

should we add another option called --merge-json ? or --merge-json-index ?

and how will this option work? as in, do we allow both --merge and --merge-json-index at a time?
or just one at a time?

if we allow both, then we will have to convert both to index object, and then merge
all of them.

thoughts? @bacongobbler @mattfarina

karuppiah7890:
[...]

PR #6757, opened 2019-10-23
[WIP] fix lint not failing in windows for bad template yamls

### LINKED ISSUE #6527: Helm3: Lint does not catch bad template yamls
Output of `helm version`:
```
```

This is related to the issue #6418 occurring in helm v2. Both have the same root cause.

[chart.zip](https://github.com/helm/helm/files/3666518/chart.zip) . The deployment yaml's `kind` key doesn't have `:` after it

```
$ helm lint chart/
==> Linting chart/

```

I'll raise a PR for this too!

karuppiah7890:
Fixes #6527 
Port of #6523 (v2 fix)
Twin of #6528 which was closed due to branch renaming

It fixes lint not failing in windows for bad template yamls

**If applicable**:

### DIFF
--- a/cmd/helm/helm.go
+++ b/cmd/helm/helm.go
@@ -56,6 +56,11 @@ func debug(format string, v ...interface{}) {
 	}
 }
 
+func warning(format string, v ...interface{}) {
+	format = fmt.Sprintf("WARNING: %s\n", format)
+	fmt.Fprintf(os.Stderr, format, v...)
+}
+
 func initKubeLogs() {
 	pflag.CommandLine.SetNormalizeFunc(wordSepNormalizeFunc)
 	gofs := flag.NewFlagSet("klog", flag.ExitOnError)
--- a/cmd/helm/install.go
+++ b/cmd/helm/install.go
@@ -17,7 +17,6 @@ limitations under the License.
 package main
 
 import (
-	"fmt"
 	"io"
 	"log"
 	"time"
@@ -207,7 +206,7 @@ func runInstall(args []string, client *action.Install, valueOpts *values.Options
 	}
 
 	if chartRequested.Metadata.Deprecated {
-		fmt.Fprintln(out, "WARNING: This chart is deprecated")
+		warning("This chart is deprecated")
 	}
 
 	if req := chartRequested.Metadata.Dependencies; req != nil {
--- a/cmd/helm/testdata/output/deprecated-chart.txt
+++ b/cmd/helm/testdata/output/deprecated-chart.txt
@@ -1,4 +1,3 @@
-WARNING: This chart is deprecated
 NAME: aeneas
 LAST DEPLOYED: Fri Sep  2 22:04:05 1977
 NAMESPACE: default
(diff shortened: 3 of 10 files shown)


thomastaylor312:
@karuppiah7890 Just to make sure, this is marked as WIP because you are working on a unit test for it correct?

karuppiah7890:
Correct! :)

karuppiah7890:
Similar to https://github.com/helm/helm/pull/6755#issuecomment-557781293 , I'm not able to build the code due to the presence of syscall.Umask in the code, which is not available in windows.

```
pkg/plugin/installer/http_installer_test.go
196:    oldmask := syscall.Umask(0000)
198:            syscall.Umask(oldmask)
```

thomastaylor312:
@karuppiah7890 so that is only in a unit test. Is there any way we can replace that with something else or rewrite the test to not use it?

karuppiah7890:
Yeah, I was checking about it. It's just that idk how permission masks make sense in Windows as it looks like it completely differently in Windows unlike Linux and Unix. I digged deeper but couldn't find much on a similar function for Windows. Also this has been quite tough and I just kept putting it away with every new hurdle 😅🙈 but this hurdle looks a bit harder. I'll see what can be done

thomastaylor312:
Honestly I would shoot for rewriting the test so it tests the same functionality. Last resort we could consider putting this test in its own file with a build tag that excludes it on windows

karuppiah7890:
I actually didn't know about the tags functionality. And yeah, I'll see about rewriting. 👍

bacongobbler:
ping. What's the current status on this PR?

karuppiah7890:
I've paused work on this, since it has been quite a bit of work, with windows VMs. I'll get back to this soon, and when I do, I'll post here, but if someone else is picking it up, that's fine too 😄

Currently, I'm working on #6155

karuppiah7890:
I'm closing this PR, because I don't have the bandwidth to attend to it.

Some info on this PR so that someone else can pick it up
* The fix is a very small fix - https://github.com/helm/helm/pull/6757/commits/851e7fc3145970163ede3cb0e1972b03f517a8f8
* The PR needed tests so that the bug doesn't pop up again. With a test the bug will be caught if it pops up.
* The testing can be done only in Windows OS as the thing we are testing is a bit OS specific - Windows OS using backslash `\` vs *nix OSes using forward slash `/`
* There is no CI/CD currently for running tests in Windows. Maintainers asked to still add tests that will catch the bug in Windows OS in case it pops up, and that running tests in Windows can be done later
* The codebase has lots of failing tests when running in Windows, so first those have to be fixed I think. And then we gotta check if the existing tests can help with verifying this issue's behavior or we can add an extra test. The failing tests are lot but the reasons for failure might be similar - like new line differences between OS and *nix systems that could be affecting assertions where we assert with test data (files) called [golden files](https://softwareengineering.stackexchange.com/questions/358786/what-is-golden-files)
* Even if the tests are fixed to run on Windows too, there's no guarantee that no one will break it again, given the CI/CD is present only for Linux, so not sure if it will be worth it to spend all the time in fixing the tests to make it work for Windows, and it can also be created as a separate issue and I would recommend adding Windows environment for testing for the CI/CD part if this is being done

PR #6771, opened 2019-10-24
Removed inaccurate statement re library

mikery:
This PR is a self-admitted exercise in pedantry. The previous wording suggested a distinction between the Helm client and Helm library, but from a user perspective the usage of the library is hidden. As a user, I use the Helm client to render my templates and communicate with k8s, so mentioning the library in this section reduces clarity.


hickeyma (approved):
@mikery LGTM, thanks. I think it brings better clarity.

PR #6779, opened 2019-10-24
fix(registry): change mediatype for chart content layer

jdolitsky:
Mediatype changed to application/tar+gzip. Please see the
following OCI mailing list item for more info:
https://groups.google.com/a/opencontainers.org/forum/#!topic/dev/pdc1lucm_Ak

Also, improved check for invalid manifests, a nil reference error was
occurring when upgrading from existing cache with old mediatype.


bacongobbler (approved):

sajayantony:
LGTM.

Issue #6712, opened 2019-10-20
Inconsistency between helm search and helm install


mszostok:
In the Service Catalog, one of our users reported inconsistency between the `helm search` and `helm install`: https://github.com/kubernetes-sigs/service-catalog/issues/2725

**TL;DR**

`helm search` shows version 0.3.0-beta.1 (the newest one), but `helm install` installs the 0.2.1 version (the old one).

**How to reproduce**
```
helm repo add svc-cat https://svc-catalog-charts.storage.googleapis.com

helm repo add svc-cat https://svc-catalog-charts.storage.googleapis.com
"svc-cat" has been added to your repositories

helm search service-catalog
NAME                     	CHART VERSION	APP VERSION	DESCRIPTION
svc-cat/catalog          	0.3.0-beta.1 	           	service-catalog webhook server and controller-manager hel...

helm install svc-cat/catalog \
>     --name catalog --namespace catalog

helm list
NAME                	REVISION	UPDATED                 	STATUS  	CHART                         	APP VERSION	NAMESPACE           
catalog             	1       	Mon Oct  7 09:55:27 2019	DEPLOYED	**catalog-0.2.1**                 	           	catalog             
```

**Investigation**

The `helm search` strategy just grabs the 0 entry from the index file:
https://github.com/helm/helm/blob/9668ad4d90c5e95bd520e58e7387607be6b63bb6/cmd/helm/search/search.go#L71-L73

but the `helm install` uses the `"github.com/Masterminds/semver"` lib to check which version should be installed:
https://github.com/helm/helm/blob/9668ad4d90c5e95bd520e58e7387607be6b63bb6/pkg/repo/index.go#L180-L188

the constraint is defined as `semver.NewConstraint("*")` so it will skip prereleases when looking at the list of releases while `>=0.0.0-0` will evaluate and find prereleases.

In `helm install` documentation there is this statement:
> (..)It will install the latest version of that chart unless you also supply a version number with the '--version' flag.

_source: https://github.com/helm/helm/blob/release-2.15/docs/helm/helm_install.md_

**Expected behavior**
The `helm search` and `helm install` should be consistent. Additionally, documentation should be adjusted to clearly say that pre-releases are skipped or alternatively the constraint in source code should be changed to match also the pre-release versions.
 
---

Output of `helm version`:

```
[...]

PR #6780, opened 2019-10-24
fix(repo/search): fix helm repo search command to display proper versions

mszostok:
**Description**
Fix the helm repo search command to display proper versions. 

Changes proposed in this pr:

- Adjust documentation to clearly state about installed/displayed versions.
- Introduce the `--devel` flag for `helm repo search` command.

**Current behavior:** 
```shell
# Searches only for stable releases, prerelease versions will be skipped
$ helm search 
 
# Searches for releases and prereleases (alpha, beta, and release candidate releases)
$ helm search --devel

# searches only for release in version 1.0.0
$ helm search --version 1.0.0
```
Signed-off-by: Mateusz Szostok <szostok.mateusz@gmail.com>

Fixes: https://github.com/helm/helm/issues/6712


mszostok:
Looks like the `TestSqlCreate` test is flaky. The time.Now() is not mocked in tests so it's by default not deterministic: https://circleci.com/gh/helm/helm/10202?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

bacongobbler (approved):

cmd/helm/search.go
```diff
@@ -34,6 +34,19 @@ const searchDesc = `
 Search reads through all of the repositories configured on the system, and
 looks for matches.
 
+It will display the latest stable versions of that chart until you specify '--devel' flag
```
bacongobbler: Would you mind updating the help string to look identical to the one in https://github.com/helm/helm/pull/6786? Thanks!

bacongobbler (changes requested):

mszostok:
Adjusted:) now it looks much better

I have only one question: based on SemVer the numbers are MAJOR.MINOR.PATCH, so shouldn’t this
```
    # Search for the latest patch release for nginx-ingress 1.x
    helm search repo nginx-ingress --version ^1.0.0
```

Be changed to
```
    # Search for the latest patch release for nginx-ingress 1.0.x
    helm search repo nginx-ingress --version ~1.0.x
```
?

bacongobbler:
Thanks for your thoughts @mszostok. Maybe the docstring should remove the word "patch". I am trying to illustrate finding the latest version of nginx-ingress in the 1.x timeline, from 1.0.0 to 2.0.0. Does that help clarify?

mszostok:
I also asked my teammates and for us `the latest patch release` is connected with the contract on SemVer. Sth like what's described here: https://documentation.spryker.com/guidelines/major-minor-patch-release.htm#what-is-a--patch-release--

Even in semver spec they are using the term "patch release" in this context:
> **What should I do if the bug that is being fixed returns the code to being compliant with the public API?**

> (..) Use your best judgment. If you have a huge audience that will be drastically impacted by changing the behavior back to what the public API intended, then it may be best to perform a major version release, even though the fix could strictly be considered a patch release.

source: https://semver.org/spec/v1.0.0.html#what-should-i-do-if-the-bug-that-is-being-fixed-returns-the-code-to-being-compliant-with-the-public-api-ie-the-code-was-incorrectly-out-of-sync-with-the-public-api-documentation

So IMO if you want to refer to the 1.x timeline then just say:
> Search for the latest minor release for nginx-ingress 1.x

This term is also used in semver docs.

mattfarina:
@mszostok With regard to the usage of "patch release". The example is for `^1.0.0`. This is going to find the latest minor and latest patch release on that minor. It would just find the latest patch release on `1.0.0`. I think this is where there can be some confusion.

It might be clearer to say something like:
> Search for the latest stable release for nginx-ingress with a major version of 1

mattfarina (approved):

Issue #6788, opened 2019-10-25
helm-3.0.0-beta.5: unable to return any non-zero exit status other than 1 from helm plugins


mumoshu:
helm-3.0.0-beta.5 swallows and rounds any non-zero exit status greater than 1 from a helm plugin to `1`.

This, for example, breaks `helm-diff` which relies on helm able to return `2` when `helm-diff` returned `2`.

This is a regression since helm-3.0.0-beta.5. The problem is exactly the same as #6384, which had been fixed in helm-3.0.0-beta.4 via #6385, but was reintroduced to helm-3.0.0-beta.5 via #6341.

---

Output of `helm version`:

```
version.BuildInfo{Version:"v3.0.0-beta.5", GitCommit:"509a8819b7e8dba86b7982c006fce22628aeea28", GitTreeState:"clean", GoVersion:"go1.13.3"}
```

karuppiah7890:
Oops. I guess this could have been avoided if we had a test for the PR which would have been a safety net and would have failed if someone changed behavior

mumoshu:
@karuppiah7890 Definitely.

I'd like to, but am unsure where to add one.

To avoid such regression, what we need to verify is "what `func main()` in cmd/helm/helm.go returns to the os`. Do you have any good idea?

Would you like to change helm's main func to be testable, like allowing to mock/stub any funcs called from it...?

karuppiah7890:
Sorry, no idea. I've to check the code. I haven't read this part of the code. And currently I've lot of other issues / PRs to work on including helm repo's 🙈 , so I would look for help from the community. Probably first timers. But we can't wait for long. v3 stable is going to be out soon in November (probably in the mid). I forgot the date, but it was mentioned in the dev call yesterday. And we'll be having release candidates

mumoshu:
@karuppiah7890 Okay! Anyway, I can submit a fix tonight and keep fixing it until v3 stable is out if you're okay with that for now :)

karuppiah7890:
> keep fixing it until v3 stable is out

? You mean if we can't write tests - keep fixing it when it breaks? We need to know if it breaks first, manually checking it is not feasible. I wouldn't advise it. 

I'll try to help with the tests for the PR. We can also get help from the maintainers. And in previous PRs also, I have refrained from submitting PRs without tests and maintainers also recommend it, an example - https://github.com/helm/helm/pull/6523#issuecomment-536603044

[...]

PR #6790, opened 2019-10-25
fix(v3): fix regression on non-zero plugin exist status

mumoshu:
Fixes the regression in helm-3.0.0-beta.5 that swallows and rounds any non-zero exit status greater than 1 from a helm plugin to `1`.

This, for example, breaks `helm-diff` which relies on helm able to return `2` when `helm-diff` returned `2`.

Closese #6788

### DIFF
--- a/cmd/helm/helm_test.go
+++ b/cmd/helm/helm_test.go
@@ -20,6 +20,8 @@ import (
 	"bytes"
 	"io/ioutil"
 	"os"
+	"os/exec"
+	"runtime"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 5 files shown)


cmd/helm/helm_test.go
```diff
@@ -151,3 +154,52 @@ func testChdir(t *testing.T, dir string) func() {
 	}
 	return func() { os.Chdir(old) }
 }
+
+func TestPluginExitCode(t *testing.T) {
```
mumoshu: This verifies that the `helm` command can return a non-zero exit status other than 1 from a plugin, by "actually running a helm process".

cmd/helm/plugin_test.go
```diff
@@ -86,11 +86,13 @@ func TestLoadPlugins(t *testing.T) {
...
+		code   int
 	}{
-		{"args", "echo args", "This echos args", "-a -b -c\n", []string{"-a", "-b", "-c"}},
-		{"echo", "echo stuff", "This echos stuff", "hello\n", []string{}},
-		{"env", "env stuff", "show the env", "env\n", []string{}},
-		{"fullenv", "show env vars", "show all env vars", envs + "\n", []string{}},
+		{"args", "echo args", "This echos args", "-a -b -c\n", []string{"-a", "-b", "-c"}, 0},
+		{"echo", "echo stuff", "This echos stuff", "hello\n", []string{}, 0},
+		{"env", "env stuff", "show the env", "env\n", []string{}, 0},
+		{"exitwith", "exitwith code", "This exits with the specified exit code", "", []string{"2"}, 2},
```
mumoshu: This doesn't actually verify which exit code is passed to `os.Exit`. But this does ensure that we don't break the helm plugin system to not swallow non-zero exit codes other than 1.

mumoshu: I mean this compliments `TestPluginExitCode` rather than duplicating it.

cmd/helm/testdata/helmhome/helm/plugins/exitwith/plugin.yaml
```diff
@@ -0,0 +1,4 @@
+name: exitwith
```
mumoshu: This is a helm plugin for testing purpose, that exits immediately with the given arg as the exit code. So `helm exitwith 2` should exit with 2 with means:

```
$ helm extiwith 1
$ echo $?
1
$ helm exitwith 2
$ echo $?
2
```

mumoshu:
Although this is now a size/M PR, the body is just addition of 5 lines and removal of 5 lines. Everything else is for tests.

Also - I don't usually write tests for `main()` but as we've already seen a regression on this feature and I think we've discussed enough the necessity of it at https://github.com/helm/helm/issues/6788#issuecomment-546300454, I believe this should be a good addition to helm's test suite.

mumoshu:
cc/ @karuppiah7890

karuppiah7890:
Will review this soon! :)

thomastaylor312 (approved):
Probably the most interesting unit test I have seen, but I think it makes sense. Going to let one more core maintainer look at this one to double check me

adamreese (approved):
interesting unit test!

PR #6798, opened 2019-10-27
Allow namespace to be set by programs consuming helm.

aaronmell:
Signed-off-by: Aaron Mell <amell@lumindigital.com>
When consuming the helm client, there is no way to correctly set the namespace except via an env variable, and that doesn't work with terraform possibly running the client concurrently in multiple namespaces. 

I'm not sure if you want to expose Namespace. If thats the case, maybe a SetNamespace function?

### DIFF
--- a/cmd/helm/helm.go
+++ b/cmd/helm/helm.go
@@ -67,7 +67,7 @@ func main() {
 	actionConfig := new(action.Configuration)
 	cmd := newRootCmd(actionConfig, os.Stdout, os.Args[1:])
 
-	if err := actionConfig.Init(settings, false, os.Getenv("HELM_DRIVER"), debug); err != nil {
+	if err := actionConfig.Init(settings.RESTClientGetter(), settings.Namespace(), os.Getenv("HELM_DRIVER"), debug); err != nil {
 		debug("%+v", err)
 		switch e := err.(type) {
 		case pluginError:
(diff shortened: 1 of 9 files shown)


bacongobbler (approved):

bacongobbler:
Would you mind writing a unit test to cover this?

aaronmell:
Sure, were you thinking a test to verify that the namespace is correct? or something else?

thomastaylor312:
Labeled as breaking because it is changing a publicly exported function signature

aaronmell:
@bacongobbler @thomastaylor312 I reworked my change to reduce the scope, I think its a better way to fix it than changing the Namespace scope.

cmd/helm/helm.go
```diff
@@ -67,7 +67,12 @@ func main() {
 	actionConfig := new(action.Configuration)
 	cmd := newRootCmd(actionConfig, os.Stdout, os.Args[1:])
 
-	if err := actionConfig.Init(settings, false, os.Getenv("HELM_DRIVER"), debug); err != nil {
+	// var namespace string
```
thomastaylor312: I assume this commented code doesn't need to be here?

thomastaylor312 (changes requested):
This does look much cleaner. Just one thing to remove

thomastaylor312 (approved):
I'll let @bacongobbler give this one more look because he started review

bacongobbler (approved):

PR #5644, opened 2019-04-25
Fix environment list in helm doc

hickeyma:
**What this PR does / why we need it**:
Closes #5344

**Special notes for your reviewer**:

**If applicable**:
- [x] this PR contains documentation
- [ ] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

PR #6806, opened 2019-10-28
fix(cmd): clean up helpstring formatting

bacongobbler:
- use tabs instead of spaces for `--help` output
- fixed a place in the bash autocompletion script that partially indents using tabs instead of spaces
- remove redundant usage examples, as `cobra` displays that information to the user
- place environment variable information into a table for clearer formatting

I decided to make the environment variables displayed in table format instead of the list format used in Helm 2 via #5644. This way, the help output is more consistent with the XDG default directory table shown lower in the help text.

### DIFF
--- a/cmd/helm/root.go
+++ b/cmd/helm/root.go
@@ -282,14 +282,14 @@ __helm_custom_func()
             __helm_list_releases
             return
             ;;
-		helm_repo_remove)
-			__helm_list_repos
-			return
-			;;
-		helm_plugin_remove | helm_plugin_update)
-			__helm_list_plugins
-			return
-			;;
+        helm_repo_remove)
+            __helm_list_repos
+            return
+            ;;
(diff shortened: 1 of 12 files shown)


cmd/helm/root.go
```diff
@@ -315,13 +315,18 @@ Common actions for Helm:
 - helm install:   upload the chart to Kubernetes
```
marckhouzam: Since you're cleaning up the help message, how about replacing helm fetch with helm pull right above here? Must be left over from helm v2.

bacongobbler: good catch. Will fix

marckhouzam:
Thanks for fixing the completion indent! It's been bugging me for weeks but I didn't want to make a PR just for that 😬

mattfarina (approved):

PR #6809, opened 2019-10-29
feat(comp): Speed up completion of charts

marckhouzam:
I feel that this improvement in performance when listing charts makes it much more useful to use shell completion when wanting to find a chart.

The completion of charts was using `helm search repo` which can be quite slow as it must parse the entire yaml of every repo cache file.

EDIT: please see https://github.com/helm/helm/pull/6809#issuecomment-560977848 below for a description of the reworked PR.

~~This commit makes the completion logic access the correct repo cache file directly and only extract the chart names from it.~~

~~With only the stable repo configured, this optimization makes the completion of charts about 10 times faster, going from 1.1 seconds (which is quite slow for a user doing completion) to 0.14 seconds; such a difference give a much better user experience.~~

If other repos are configured, `helm search repo` only become slower, while the new completion logic will not be affected as it only looks for the relevant single repo file.

**Notes for maintainers**:

~~Can a maintainer tell me their opinion on the approach of using the cache file by:
1- extracting the cache directory by looking for HELM_REPOSITORY_CACHE in the output of `helm env`
2- greping for chart names (using `\egrep -e '^  [a-zA-Z]'` on the `$repo-index.yaml` cache file~~

~~I would have liked something a little more straightforward, but nothing better jumped out at me.~~

This change was tested for backwards-compatibility using the acceptance-testing repo

### DIFF
--- a/pkg/repo/chartrepo.go
+++ b/pkg/repo/chartrepo.go
@@ -133,10 +133,21 @@ func (r *ChartRepository) DownloadIndexFile() (string, error) {
 		return "", err
 	}
 
-	if _, err := loadIndex(index); err != nil {
+	indexFile, err := loadIndex(index)
+	if err != nil {
 		return "", err
 	}
 
+	// Create the chart list file in the cache directory
+	var charts strings.Builder
+	for name := range indexFile.Entries {
+		fmt.Fprintln(&charts, name)
+	}
+	chartsFile := filepath.Join(r.CachePath, helmpath.CacheChartsFile(r.Config.Name))
+	os.MkdirAll(filepath.Dir(chartsFile), 0755)
+	ioutil.WriteFile(chartsFile, []byte(charts.String()), 0644)
+
+	// Create the index file in the cache directory
 	fname := filepath.Join(r.CachePath, helmpath.CacheIndexFile(r.Config.Name))
 	os.MkdirAll(filepath.Dir(fname), 0755)
 	return fname, ioutil.WriteFile(fname, index, 0644)
--- a/cmd/helm/repo_add_test.go
+++ b/cmd/helm/repo_add_test.go
@@ -19,6 +19,7 @@ package main
 import (
 	"fmt"
 	"io/ioutil"
+	"os"
 	"path/filepath"
 	"sync"
 	"testing"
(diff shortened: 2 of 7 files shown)


marckhouzam:
Although technically not a bug fix, this is only a 2-line change that makes the completion of charts much nicer.  The change is safe as it can only affect completion.

I wonder if this could be considered for the 3.0 release?

The user experience would be much nicer because, when completing the name of a chart, (e.g. `helm install myrelease stable/ngi<TAB>` a user may press `<TAB>` multiple times; without this fix, each `<TAB>` will take around 2 seconds or more, which defeats the purpose of asking the computer to complete the word faster than if the user had typed it.

marckhouzam:
I reworked this PR in a much less hacky fashion.

This new attempt has helm create a new cache file for each repo which contains the list of charts for that repo.  Helm will do this whenever it downloads the repo index file.

The completion logic then uses this new cache file directly and obtains the chart names very quickly.
With only the stable repo configured, this optimization makes the completion of charts about 85 times faster, going from 1.2 seconds to 0.014 seconds; such a difference gives a much better user experience when completing chart names (e.g., `helm install nginx stable/ng<TAB>`).

On the other hand, adding the creation of the new chart-list cache file to 'helm update' or 'helm add' is pretty much negligible compared to the time spent parsing the index files.

marckhouzam:
I've just added tests for this change and also fixed the fact that I had forgotten to remove the new cache file when doing a `helm repo remove`

pkg/repo/chartrepo.go
```diff
@@ -133,10 +133,21 @@ func (r *ChartRepository) DownloadIndexFile() (string, error) {
...
-	if _, err := loadIndex(index); err != nil {
+	indexFile, err := loadIndex(index)
+	if err != nil {
 		return "", err
 	}
 
+	// Create the chart list file in the cache directory
+	var charts strings.Builder
+	for name := range indexFile.Entries {
+		fmt.Fprintln(&charts, name)
```
bacongobbler: would you mind writing a unit test for this if possible?

bacongobbler (approved):
LGTM! I would suggest adding more test coverage for some of the new added functionality, but overall this looks like it solves the issue

pkg/repo/chartrepo.go
marckhouzam: Will do. Thanks for the review!

marckhouzam:
This last forced push was just to rebase and to correct the commit message.  New tests still pending.

pkg/repo/chartrepo.go
marckhouzam: The simplest way to test this was to extend the Index file test of `index_test.go`.
The latest version of the PR does just that.

marckhouzam:
@bacongobbler ready for review again.

bacongobbler (approved):
manually tested. LGTM!

PR #6811, opened 2019-10-29
fix(v3): Add `--client-only` and fix `--validate` for `helm template`

### LINKED ISSUE #6505: helm3 template fails when there are custom resource templates

mumoshu:
A.k.a "allow `helm template` to optionally skip api versions check while keeping existing use-cases and features as-is"

This introduces a new flag `--client-only` to `helm template`, so that (1)we can optionally disable the validation at all, (2)while keeping the ability to switch what are validated.

(1) is achieved by changing the semantics of the existing flag `--validate=[true|false]` new in v3 and (2) is achieved by adding `--client-only=[true|false]` that mostly the same as the former `--validate`.

This should address all the three use-cases listed in https://github.com/helm/helm/pull/6729#issuecomment-547249311 after fixing other related bugs.

Closes #6505
Replaces #6729

---

Even though the semantics of the existing flag `--validate` has changed, the existing and default behavior of `helm template` is maintained. That is, `helm template` without any flags means `do client-only validation` before/after this change.

Regarding changes in the `--validate`, as you can see from the existing code-base and the description of the flag, unlike it's name, it had been used for toggling the client-only mode of the validation. For instance, `helm template [--validate=false]` had been left api versions check running even though, `--validate=false` or omitting the flag would had disabled the whole validation according to it's name.

This change adds `--client-only` so that we can toggle the validation and the client-only mode separately, which allows more granular control, and more specifically, support the `do not validate at all` scenario.

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -68,6 +68,7 @@ type Install struct {
 	ChartPathOptions
 
 	ClientOnly       bool
+	NoValidate       bool
 	DryRun           bool
 	DisableHooks     bool
 	Replace          bool
@@ -99,6 +100,10 @@ type ChartPathOptions struct {
 	Username string // --username
 	Verify   bool   // --verify
 	Version  string // --version
+}
+
+type RenderOptions struct {
+	NoValidate bool
 }
 
 // NewInstall creates a new Install object with the given configuration.
@@ -151,8 +156,10 @@ func (i *Install) installCRDs(crds []*chart.File) error {
 //
 // If DryRun is set to true, this will prepare the release, but not install it
 func (i *Install) Run(chrt *chart.Chart, vals map[string]interface{}) (*release.Release, error) {
+	validate := !i.NoValidate
+
 	// Check reachability of cluster unless in client-only mode (e.g. `helm template` without `--validate`)
-	if !i.ClientOnly {
+	if validate && !i.ClientOnly {
 		if err := i.cfg.KubeClient.IsReachable(); err != nil {
 			return nil, err
 		}
@@ -164,7 +171,7 @@ func (i *Install) Run(chrt *chart.Chart, vals map[string]interface{}) (*release.
 
 	// Pre-install anything in the crd/ directory. We do this before Helm
 	// contacts the upstream server and builds the capabilities object.
-	if crds := chrt.CRDs(); !i.ClientOnly && !i.SkipCRDs && len(crds) > 0 {
+	if crds := chrt.CRDs(); validate && !i.SkipCRDs && len(crds) > 0 {
 		// On dry run, bail here
 		if i.DryRun {
 			i.cfg.Log("WARNING: This chart or one of its subcharts contains CRDs. Rendering may fail or contain inaccuracies.")
@@ -211,7 +218,7 @@ func (i *Install) Run(chrt *chart.Chart, vals map[string]interface{}) (*release.
 	rel := i.createRelease(chrt, vals)
 
 	var manifestDoc *bytes.Buffer
-	rel.Hooks, manifestDoc, rel.Info.Notes, err = i.cfg.renderResources(chrt, valuesToRender, i.OutputDir, i.SubNotes)
+	rel.Hooks, manifestDoc, rel.Info.Notes, err = i.cfg.renderResources(chrt, valuesToRender, i.OutputDir, i.SubNotes, RenderOptions{NoValidate: i.NoValidate})
 	// Even for errors, attach this if available
 	if manifestDoc != nil {
 		rel.Manifest = manifestDoc.String()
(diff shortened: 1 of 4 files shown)


pkg/action/install.go
```diff
@@ -68,6 +68,7 @@ type Install struct {
 	ChartPathOptions
 
 	ClientOnly       bool
+	NoValidate       bool
```
mumoshu: This should be `Validate `for readability that that seemed to require me to update many places that initializes `Install` objects.

pkg/releaseutil/manifest_sorter.go
```diff
@@ -74,7 +79,12 @@ var events = map[string]release.HookEvent{
 //
 // Files that do not parse into the expected format are simply placed into a map and
 // returned.
-func SortManifests(files map[string]string, apis chartutil.VersionSet, sort KindSortOrder) ([]*release.Hook, []Manifest, error) {
+func SortManifests(files map[string]string, apis chartutil.VersionSet, sort KindSortOrder, opts ...SortOptions) ([]*release.Hook, []Manifest, error) {
```
mumoshu: Should we use the functional option pattern here instead?

pkg/releaseutil/manifest_sorter.go
```diff
@@ -129,7 +140,7 @@ func (file *manifestFile) sort(result *result) error {
 			return errors.Wrapf(err, "YAML parse error on %s", file.path)
 		}
 
-		if entry.Version != "" && !file.apis.Has(entry.Version) {
+		if file.validate && entry.Version != "" && !file.apis.Has(entry.Version) {
 			return errors.Errorf("apiVersion %q in %s is not available", entry.Version, file.path)
```
thomastaylor312: I feel like a lot of these problems (and a few others we have run into) could be solved more simply by removing this check in the sorter. I am not sure why the sorter cares if it has an unsupported API version (trying to track down why that is). An unknown type should be sorted to the bottom of the list and then an actual validator should catch the missing API type. Going to do some more digging here

thomastaylor312: I actually am leaning towards removing this check, which would solve the issue (I think). Let me whip up (yet another) PR and link it to this

thomastaylor312:
Hey @mumoshu! We merged #6817 to fix the main issue, but we think there is still some value for being able to turn off all API validation during an install or upgrade (for the "YOLO cases"), so we are keeping this one open in case you want to continue forward with the `NoValidate` option

bacongobbler:
Hey @mumoshu. https://github.com/helm/helm/pull/6819 should be identical to the feature proposed in this PR now that #6817 has been merged. I feel that a more descriptive flag than `--client-only` is the right approach here, as that flag has implications with Helm 2's `helm version --client-only` flag, implying that TIller is still somehow involved.

Let me know what you think

mumoshu:
@thomastaylor312 Thanks! It all makes sense now.

One of my purposes for this PR was to be able to disable the api-versions check, and your change did resolve that. So I have no use for this PR at least in that regard.

mumoshu:
Hey @bacongobbler, thanks a lot for all your hard work and releasing rc.2! I tested it today and it worked like a charm with helmfile and helm-diff 🎉 

`--disable-openapi-validation` looks much better. But I don't have a specific use-case for that myself.

I think I just happened to cover some of `--disable-openapi-validation` use-case in this PR due to the way I tried to make the api-versions check togglable. I thought it was necessary, but turned out we could just remove the api-versions check in the sorter instead.

That said, anyways, I'm fine with closing this PR as not necessary anymore.

Re `--disable-openapi-validation`, I believe there's no specific and concrete use-case for `--disable-openapi-validation` yet, as you've also mentioned in the PR.

Perhaps we can close it for now to reduce the maintenance work and focus on GA of v3.0.0, and wait for anyone to raise an issue with real use-case(s) that can possibly be addressed with `--disable-openapi-validation`?

Also, I understand your concerns on the naming of `--validation`. I have no better idea either.

But I believe that the name works for most people now. Now that the vanilla `helm template` doesn't berk on api-versions anymore, it would LOOK LIKE to have no validations by default. Implementation-wise that might be wrong, but U/X wise that's fine as long as they encounter any real problem.

mumoshu:
Closing this as unnecessary now. Thanks a lot for all your works and thoughtful responses! You made my day 😄

PR #6817, opened 2019-10-29
fix(releaseutil): Removes API version checks from kind sorter

thomastaylor312:
The sorting method for manifests contained a check to see if the API
version existed. This violates separation of concerns as the sorter
should just sort and leave validation to other parts of the code.

I validated that things work as intended, I had a chart with this manifest in it:
```yaml
kind: Foo
apiVersion: foo.bar/v1
spec:
  foo: bar
```

When I run a `helm template` without `--validate`, it outputs templates successfully.

When I run it with `--validate`, I get the following output:

```shell
$ ./bin/helm template failme --validate
Error: unable to build kubernetes objects from release manifest: unable to recognize "": no matches for kind "Foo" in version "foo.bar/v1"
```

This will supersede #6811 if accepted

### DIFF
--- a/pkg/releaseutil/manifest_sorter.go
+++ b/pkg/releaseutil/manifest_sorter.go
@@ -129,10 +129,6 @@ func (file *manifestFile) sort(result *result) error {
 			return errors.Wrapf(err, "YAML parse error on %s", file.path)
 		}
 
-		if entry.Version != "" && !file.apis.Has(entry.Version) {
-			return errors.Errorf("apiVersion %q in %s is not available", entry.Version, file.path)
-		}
-
 		if !hasAnyAnnotation(entry) {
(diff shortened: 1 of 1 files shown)


thomastaylor312:
As a note, we should be able to remove `--api-versions` from `helm template` if people are ok with this PR. I would do that in a quick follow up PR

adamreese (approved):

bacongobbler (approved):

thomastaylor312:
After talking more with @adamreese and @bacongobbler we decided there is definitely still use for the `--api-versions` flag.

pasumarthivijaykumar:
i have installed kubernetes-helm v3.0.2, facing same issue. Can you please let me know what is latest version with this fix?

PR #6819, opened 2019-10-29
feat(install): introduce --disable-openapi-validation

bacongobbler:
When enabled, during the rendering process, this feature flag will not validate rendered templates against the Kubernetes OpenAPI Schema.

This proposes an alternative approach to #6811 now that #6817 has been merged, completely disabling the OpenAPI schema validation.

This change is backwards-compatible.

### DIFF
--- a/cmd/helm/install.go
+++ b/cmd/helm/install.go
@@ -138,6 +138,7 @@ func addInstallFlags(f *pflag.FlagSet, client *action.Install, valueOpts *values
 	f.StringVar(&client.Description, "description", "", "add a custom description")
 	f.BoolVar(&client.Devel, "devel", false, "use development versions, too. Equivalent to version '>0.0.0-0'. If --version is set, this is ignored")
 	f.BoolVar(&client.DependencyUpdate, "dependency-update", false, "run helm dependency update before installing the chart")
+	f.BoolVar(&client.DisableOpenAPIValidation, "disable-openapi-validation", false, "if set, the installation process will not validate rendered templates against the Kubernetes OpenAPI Schema")
 	f.BoolVar(&client.Atomic, "atomic", false, "if set, installation process purges chart on fail. The --wait flag will be set automatically if --atomic is used")
 	f.BoolVar(&client.SkipCRDs, "skip-crds", false, "if set, no CRDs will be installed. By default, CRDs are installed if not already present")
 	f.BoolVar(&client.SubNotes, "render-subchart-notes", false, "if set, render subchart notes along with the parent")
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -67,23 +67,24 @@ type Install struct {
 
 	ChartPathOptions
 
-	ClientOnly       bool
-	DryRun           bool
-	DisableHooks     bool
-	Replace          bool
-	Wait             bool
-	Devel            bool
-	DependencyUpdate bool
-	Timeout          time.Duration
-	Namespace        string
-	ReleaseName      string
-	GenerateName     bool
-	NameTemplate     string
-	Description      string
-	OutputDir        string
-	Atomic           bool
-	SkipCRDs         bool
-	SubNotes         bool
+	ClientOnly               bool
+	DryRun                   bool
+	DisableHooks             bool
+	Replace                  bool
+	Wait                     bool
+	Devel                    bool
+	DependencyUpdate         bool
+	Timeout                  time.Duration
+	Namespace                string
+	ReleaseName              string
+	GenerateName             bool
+	NameTemplate             string
+	Description              string
+	OutputDir                string
+	Atomic                   bool
+	SkipCRDs                 bool
+	SubNotes                 bool
+	DisableOpenAPIValidation bool
 	// APIVersions allows a manual set of supported API Versions to be passed
 	// (for things like templating). These are ignored if ClientOnly is false
 	APIVersions chartutil.VersionSet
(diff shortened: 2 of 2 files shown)


thomastaylor312:
Should this be added to the upgrade command as well?

bacongobbler:
Maybe... There's two ways we can consider this feature for Helm 3.1.0:

1. Allow users to run every command without OpenAPI validation (HIGHLY discouraged, but maybe there's a valid use case we're not seeing?)
2. Allow users to run `helm template` without OpenAPI validation

I'm not sold on either approach personally. I don't think this should be configurable at all, but I'm open to hearing from others on whether this would be useful or not.

hickeyma:
@bacongobbler Do you want to rebase to resolve the conflict before review?

bacongobbler:
rebased

muvaf:
We had to tell our users to use helm2 in case their Kubernetes is pre-1.15 https://github.com/crossplaneio/crossplane/pull/1208/files#diff-1cbebe06013098b1262a48984d51b379R28

It'd be great to merge this to fix https://github.com/helm/helm/issues/7404 in the next release.

vinzent:
@bacongobbler  

> Allow users to run every command without OpenAPI validation (HIGHLY discouraged, but maybe 
> there's a valid use case we're not seeing... YOLO?)

unfortunately for example OpenShift 3.11 extentions (probably still in latest OpenShift 4 too) have marked many optional properties as required leading to false validation errors.

So my usecase it to workaround schema definition issues that I can't influence/change myself.

See https://github.com/helm/helm/issues/6830 and https://github.com/openshift/origin/issues/24060

bacongobbler:
Yep. I’ve heard that from others too. This is necessary to work around issues with openshift’s OpenAPI schema. I’m hoping they can fix this upstream, but this can hopefully get people hobbling forward with openshift.

muvaf:
It's not only openshift but any CRD that has some field types that did not exist in the prior versions of Kubernetes. For example https://github.com/kubernetes-sigs/controller-tools/issues/392 . `IntOrString` exists in `Deployment` struct only in post-1.15 but if your controller embeds that type (or `Deployment` struct ) and you want to deploy it to a pre-1.15 cluster, you can't do it via helm 3 because validation would fail.

K-DOT:
Any updates on this?

bacongobbler:
waiting on a review. It's slated for Helm 3.1.

mattfarina (approved):

xrow:
Should this exist for upgrade too? If so it needs fixing.

```
[root@False helm-ezplatform]# helm upgrade --install --version 0.10.1 cert-manager jetstack/cert-manager -n cert-manager --disable-openapi-validation --set webhook.enabled=false --set global.isOpenshift=true 
Error: unknown flag: --disable-openapi-validation
```

mherwig:
I see that this new flag is working for ``helm install`` but not for ``helm upgrade``.

Since I use helmfile in combination with the 'apply' command which will run ``helm upgrade --install`` underneath I cannot use this flag.

Can you please also add this flag to upgrade command?

hhellbusch:
+1 for the request on adding this flag for `helm upgrade`.  We currently use `helm upgrade --install` in our CD.  If this is added for for the upgrade command, then we can use this as a work around until we're able to upgrade to OCP4.4 (see https://bugzilla.redhat.com/show_bug.cgi?id=1773682).

bacongobbler:
see #7570. Someone from the community (thanks @Maronee!) has already gone ahead and started work on this.

bacongobbler:
For the record, I am [incredibly uncomfortable disabling API validation on helm install and helm upgrade](https://github.com/helm/helm/pull/6819#issuecomment-547645422). It introduces cases where Kubernetes will accept manifests that it shouldn't normally accept, and can cause indeterminate behaviour during upgrades. I'd highly suggest pushing Openshift to fix their OpenAPI schema and moving off this flag ASAP.

If you are going to opt into this feature flag, please make sure you're doing it for the right reasons.

bacongobbler:
@hhellbusch looking at that bug, it looks like the last comment confirms the bug is still present in v4.4.0. I'm not sure upgrading to Openshift 4.4.0 will fix this. Some users have found workarounds in https://github.com/openshift/origin/issues/24060 if you want to look into those options.

hhellbusch:
@bacongobbler I agree 100% - disabling the OpenAPI spec should be avoided.  

And yeah, the fix is slated for 4.4 which is still a WIP. (4.3 was just GA'd a week or two ago). (Also note that targeting to release the fix with 4.4 is not a promise, but a target).
The last comment on the bugzilla verifies the bug as a problem in the 4.4 nightly build.  Thanks for sharing the workaround in https://github.com/openshift/origin/issues/24060

In the case of OpenShift, anyone who is reading that is also a RH customer, please contact support / your account team and let them know you would like the fix to be backported.  This helps the product team see the need/demand for the backport.

jkroepke:
> In the case of OpenShift, anyone who is reading that is also a RH customer
Wrong. OpenShift is also available as opensource distribution calling OKD without any support.

bacongobbler:
I read that his comment as "if you happen to be a customer of RedHat running OpenShift", not "if you are running OpenShift, you are a RedHat customer. Therefore..."

Either way, urging RedHat to fix this in their product is a good call-to-action here.

jkroepke:
and start petitions to backport the fixes to the logterm 3.x branch ....

PR #6831, opened 2019-10-30
[Makefile] Support reproducible builds

Foxboron:
Circleci is used to build the release artifacts and embeds build paths
into the binary release. To reproduce the release binaries we then need
to also build in the same path as a result.

    $ strings linux-amd64/helm  | grep "home/circleci" | wc -l
    174

Go 1.13 introduces `-trimpath` which strips the build path from all
compiled binaries. This should enable people to reproduce the
distributed helm binaries.

https://reproducible-builds.org/docs/source-date-epoch/
https://golang.org/doc/go1.13#go-command

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ TAGS       :=
 TESTS      := .
 TESTFLAGS  :=
 LDFLAGS    := -w -s
-GOFLAGS    :=
+GOFLAGS    := -trimpath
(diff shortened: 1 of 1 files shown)


thomastaylor312 (approved):
This looks great! Thanks for doing this. As we are currently in the RC phase for 3.0, this will be going in to 3.1

PR #6837, opened 2019-10-30
fix(action): strip file extensions from name

### LINKED ISSUE #6600: Helm v3 generates invalid release name
Output of `helm version`:

Output of `kubectl version`:
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.6", GitCommit:"96fac5cd13a5dc064f7d9f4f23030a6aeface6cc", GitTreeState:"clean", BuildDate:"2019-08-19T11:13:49Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"darwin/amd64"}

Docker Desktop 

Running `helm install mongodb-7.3.0.tgz -g`
Error: Service "mongodb-7.3.0.tgz-1570513937" is invalid: metadata.name: Invalid value: "mongodb-7.3.0.tgz-1570513937": a DNS-1035 label must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?')

I think the issue arise from 
https://github.com/helm/helm/blob/cc450563835088945aa79dbfa0bce0bc031fb390/pkg/action/install.go#L556-L562

As the base = "mongodb-7.3.0.tgz". 

Should we add validation like Helm v2
https://github.com/helm/helm/blob/fb2cbb001986b28dec5303f5acad65390a1a8118/cmd/helm/install.go#L260-L263

bacongobbler:
This PR fixes an issue where `helm install mychart.tgz --generate-name` generates an invalid name with the file extension present. This also adds a ton of unit test coverage around `NameAndChart` that wasn't there before.

closes #6600

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -564,6 +564,10 @@ func (i *Install) NameAndChart(args []string) (string, string, error) {
 	if base == "." || base == "" {
 		base = "chart"
 	}
+	// if present, strip out the file extension from the name
+	if idx := strings.Index(base, "."); idx != -1 {
+		base = base[0:idx]
+	}
 
 	return fmt.Sprintf("%s-%d", base, time.Now().Unix()), args[0], nil
 }
--- a/pkg/action/install_test.go
+++ b/pkg/action/install_test.go
@@ -33,6 +33,7 @@ import (
 	kubefake "helm.sh/helm/v3/pkg/kube/fake"
 	"helm.sh/helm/v3/pkg/release"
 	"helm.sh/helm/v3/pkg/storage/driver"
+	"helm.sh/helm/v3/pkg/time"
 )
 
 type nameTemplateTestCase struct {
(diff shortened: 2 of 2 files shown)


pkg/action/install.go
```diff
@@ -564,6 +564,10 @@ func (i *Install) NameAndChart(args []string) (string, string, error) {
 	if base == "." || base == "" {
 		base = "chart"
 	}
+	// if present, strip out the file extension from the name
+	if idx := strings.Index(base, "."); idx != -1 {
```
thomastaylor312: Are there any valid chart names that could have `.` in the name? I feel like there possibly could be and it would be better to use `strings.ReplaceAll(base, file path.Ext(base), "")`

thomastaylor312 (commented):
One blocking question/discussion here. Otherwise looks great

pkg/action/install.go
bacongobbler: `path.Ext` only catches the first file extenion, but won't capture the `.tar.gz` case. I'm not sure if there's a way we can detect that without dropping down to `os.Stat` to check if the file handler is for a directory.

thomastaylor312: Ok, sounds good.

thomastaylor312 (approved):
Good to go pending file fix

PR #6839, opened 2019-10-30
(nit) Add namespace option to example kubectl command

jeffb4:
Helm emits sample commands for accessing a release. Unfortunately it does not have the `--namespace` option in the example `kubectl port-forward` command which trips one up if the deployment was not to the default ns.


bacongobbler (approved):

bacongobbler:
thanks for catching this.

PR #6840, opened 2019-10-30
Revert "[Makefile] Support reproducible builds"

thomastaylor312:
Reverts helm/helm#6831. Currently `gox` [doesn't support this new flag](https://github.com/mitchellh/gox/pull/138). It looks like there is a work around in the gox issue or we can wait until support is merged.

/cc @Foxboron

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ TAGS       :=
 TESTS      := .
 TESTFLAGS  :=
 LDFLAGS    := -w -s
-GOFLAGS    := -trimpath
+GOFLAGS    :=
 SRC        := $(shell find . -type f -name '*.go' -print)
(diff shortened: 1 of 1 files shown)


bacongobbler (approved):

Foxboron:
Right. Supporting `GOFLAGS` the env variable would need some changes to the current Makefile as the syntax differ between the CLI and `GOFLAGS` flags.

Also, having the CI test with `gox` instead of `go` would have caught this. But `gox` seems abandoned with little to no activity the past year.

thomastaylor312:
@Foxboron Yeah, we might want to do some larger refactoring with the Makefile and possibly remove the dependency on gox

PR #6841, opened 2019-10-30
fix(show): restore comments from raw values

### LINKED ISSUE #6354: Helm 3: helm inspect values does not show comments
Output of `helm version`:
version.BuildInfo{Version:"v3.0.0-beta.2", GitCommit:"26c7338408f8db593f93cd7c963ad56f67f662d4", GitTreeState:"clean", GoVersion:"go1.12.9"}

The new `helm show values` (was `helm inspect`) outputs the parsed values.yaml of the chart with comments stripped out. This is a problem for helm charts that rely heavily on comments to document the exposed options (instead of repeating that information in the README). 

I think default behavior should be the same as `helm inspect` and dump the values.yaml file verbatim. There's definitely value in presenting the parsed/stripped values as it does now, but this should probably be exposed with a flag (e.g. `--stripped`, `--parsed`, or `--no-comments`).

bacongobbler:
`helm show values` was reading the values as marshaled from `loader.Load`, which stripped the original comments from the values file. This PR restores the `RawValues` field to `chart.Chart` to restore `helm show values`'s previous behaviour to render the raw values file, comments and all.

closes #6354

### DIFF
--- a/pkg/chart/loader/load.go
+++ b/pkg/chart/loader/load.go
@@ -74,6 +74,7 @@ func LoadFiles(files []*BufferedFile) (*chart.Chart, error) {
 	subcharts := make(map[string][]*BufferedFile)
 
 	for _, f := range files {
+		c.Raw = append(c.Raw, &chart.File{Name: f.Name, Data: f.Data})
 		switch {
 		case f.Name == "Chart.yaml":
 			if c.Metadata == nil {
--- a/pkg/action/show.go
+++ b/pkg/action/show.go
@@ -24,6 +24,7 @@ import (
 
 	"helm.sh/helm/v3/pkg/chart"
 	"helm.sh/helm/v3/pkg/chart/loader"
+	"helm.sh/helm/v3/pkg/chartutil"
 )
 
 type ShowOutputFormat string
@@ -76,11 +77,11 @@ func (s *Show) Run(chartpath string) (string, error) {
 		if s.OutputFormat == ShowAll {
 			fmt.Fprintln(&out, "---")
 		}
-		b, err := yaml.Marshal(chrt.Values)
-		if err != nil {
-			return "", err
+		for _, f := range chrt.Raw {
+			if f.Name == chartutil.ValuesfileName {
+				fmt.Fprintln(&out, string(f.Data))
+			}
 		}
-		fmt.Fprintf(&out, "%s\n", b)
 	}
 
 	if s.OutputFormat == ShowReadme || s.OutputFormat == ShowAll {
--- a/pkg/chart/chart.go
+++ b/pkg/chart/chart.go
@@ -26,13 +26,18 @@ const APIVersionV2 = "v2"
 // Chart is a helm package that contains metadata, a default config, zero or more
 // optionally parameterizable templates, and zero or more charts (dependencies).
 type Chart struct {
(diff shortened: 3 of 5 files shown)


bacongobbler:
Slating this for 3.0.0 as this was a code regression introduced in Helm 3 that users had come to expect.

bacongobbler:
Gonna pull out 7e43089 into a separate PR; it doesn't actually affect the behaviour and should be slated as a 3.1 feature.

bacongobbler:
done; moved to #6845

bacongobbler:
@adamreese and I paired on this one in the afternoon. While this does restore prior functionality, there has to be a better way to avoid re-introducing the `RawValues` object back to the chart. Will sleep on this and see if we can tackle this feature in a different way, likely to do with refactoring how the `Files` field works instead.

bacongobbler:
Okay, after thinking it over last night, we're trying a new approach by loading the raw files in memory.

The problem with this new approach is that the raw files will be stored in the secret once we create the release. We could scrub it from memory during a save, though that seems a bit of a hack.

Additionally, because the `Raw` field is made public, the field can be modified after we call `LoadDir`. That doesn't seem like a good design from the SDK's point of view. It should be read-only after loading the chart.

@mattfarina suggested we could use an alternative YAML parser that preserves comments, though those libraries are currently unstable and could introduce other issues down the road.

If anyone has alternatives, I'm all ears at this point.

bacongobbler:
hack uploaded to demonstrate a possible workaround.

bacongobbler:
manually testing this seems to work as expected: `helm show values` display the values as expected, and the `.Raw` field is not saved during `helm install` as part of the release metadata.

```
helm create foo
helm install foo ./foo
kubectl get secret sh.helm.release.v1.foo.v1 -o jsonpath='{.data.release}' | base64 -d | base64 -d | gunzip
```

hickeyma (approved):
LGTM, thanks @bacongobbler

hickeyma:
@bacongobbler It might be worth other maintainers to take a look before merging?

bacongobbler:
Yeah. I'd like to get @adamreese's review on this one before merging as we discussed possible paths forward here.

mattfarina:
@bacongobbler Can you resolve the conflict that's now come up here.

pkg/chart/loader/load.go
```diff
@@ -74,6 +74,7 @@ func LoadFiles(files []*BufferedFile) (*chart.Chart, error) {
 	subcharts := make(map[string][]*BufferedFile)
 
 	for _, f := range files {
+		c.Raw = append(c.Raw, &chart.File{Name: f.Name, Data: f.Data})
```
mattfarina: It looks like all of the `Raw` files will be stored on the release object (by default a secret). This is going to cause them to balloon in size and we will run into storage space problems.

What happened to the change that deleted `Raw` from the release object?

mattfarina (changes requested):

pkg/chart/loader/load.go
hickeyma: When I tested this manually, I did not see `Raw` field stored.

bacongobbler:
conflicts resolved.

mattfarina (approved):

PR #6842, opened 2019-10-30
fix(template): process manifests in file path order, then in order found in each file (before sorting manifests)

AndiDog:
Previously, `helm template` would go through e.g. `templates/*.yml` in unspecified order, and the same issue applied when looping through the manifests within each file. That's because Go does not guarantee ordering of a map ranges, but Helm loops over maps. The unfortunate result: each call to `helm template` might produce a different order of manifest objects. While this surely went unnoticed for so long because most people use Helm as deployment operator, it's pretty bad for people who either commit the `helm template` output somewhere, or even auto-commit in a CI/CD pipeline – producing random results would make the diffs unreadable because objects keep moving to somewhere in the generated output.

This PR adds a fix and test (+ 100 test repeats because with the buggy implementation, the test could have randomly produced the expected result if run only once).

### DIFF
--- a/cmd/helm/template_test.go
+++ b/cmd/helm/template_test.go
@@ -84,6 +84,14 @@ func TestTemplateCmd(t *testing.T) {
 			cmd:    fmt.Sprintf("template '%s' --include-crds", chartPath),
 			golden: "output/template-with-crds.txt",
 		},
+		{
+			name:   "sorted output of manifests (order of filenames, then order of objects within each YAML file)",
+			cmd:    fmt.Sprintf("template '%s'", "testdata/testcharts/object-order"),
+			golden: "output/object-order.txt",
+			// Helm previously used random file order. Repeat the test so we
+			// don't accidentally get the expected result.
+			repeat: 10,
+		},
 	}
 	runTestCmd(t, tests)
 }
--- a/pkg/releaseutil/manifest.go
+++ b/pkg/releaseutil/manifest.go
@@ -19,6 +19,7 @@ package releaseutil
 import (
 	"fmt"
 	"regexp"
+	"strconv"
 	"strings"
 )
 
@@ -37,8 +38,9 @@ var sep = regexp.MustCompile("(?:^|\\s*\n)---\\s*")
 // SplitManifests takes a string of manifest and returns a map contains individual manifests
 func SplitManifests(bigFile string) map[string]string {
 	// Basically, we're quickly splitting a stream of YAML documents into an
-	// array of YAML docs. In the current implementation, the file name is just
-	// a place holder, and doesn't have any further meaning.
+	// array of YAML docs. The file name is just a place holder, but should be
+	// integer-sortable so that manifests get output in the same order as the
+	// input (see `BySplitManifestsOrder`).
 	tpl := "manifest-%d"
 	res := map[string]string{}
 	// Making sure that any extra whitespace in YAML stream doesn't interfere in splitting documents correctly.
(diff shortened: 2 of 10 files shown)


bacongobbler:
is there a bug in teh queue this PR should reference?

AndiDog:
I created #6843 for completeness' sake 😉

thomastaylor312:
See https://github.com/helm/helm/issues/6843#issuecomment-548143055 for additional discussion, but we are labeling this as a feature and adding to the 3.1 milestone

AndiDog:
Renamed the commit to call it a "feature", not a "fix"

pkg/releaseutil/manifest.go
```diff
@@ -39,7 +39,7 @@ func SplitManifests(bigFile string) map[string]string {
 	// Basically, we're quickly splitting a stream of YAML documents into an
 	// array of YAML docs. In the current implementation, the file name is just
 	// a place holder, and doesn't have any further meaning.
-	tpl := "manifest-%d"
+	tpl := "manifest-%010d"
```
bacongobbler: This would be considered a breaking change as the expected output has been modified. The zero-index used to be `manifest-0`, but now it'll be `manifest-0000000000`. Users may have calling code expecting the former.

Additionally, that means this function has been limited to sort a maximum of 10 billion manifests, whereas before it could sort `math.MaxInt`. Not a big deal as I don't think anyone was sorting that amount of manifests, but regardless it's still a behavioural change.

A non-breaking change would be to replace `sort.Strings` with `sort.Sort` in `SortManifests`, and implement a sorter for the output of `SplitManifests`. That way, `sort.Sort` sorts the manifests in the expected order during SortManifests.

bacongobbler (changes requested):

cmd/helm/template_test.go
```diff
@@ -79,6 +79,12 @@ func TestTemplateCmd(t *testing.T) {
 			cmd:    fmt.Sprintf("template --api-versions helm.k8s.io/test '%s'", chartPath),
 			golden: "output/template-with-api-version.txt",
 		},
+		{
+			name:   "sorted output of manifests (order of filenames, then order of objects within each YAML file)",
+			cmd:    fmt.Sprintf("template '%s'", "testdata/testcharts/object-order"),
+			golden: "output/object-order.txt",
+			repeat: 100,
```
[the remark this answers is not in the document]
AndiDog: Before the fix, Helm took files in random map order and could therefore produce indeterministic results and even give the expected, sorted result by chance. I've reduced the repeats to 10 and explained the rationale. Once flaky, always flaky, so not much we can do here if we want to ensure previous behavior doesn't happen anymore.

pkg/releaseutil/manifest.go
AndiDog: Applied as suggested, sorting by the part after `manifest-`.

In addition, I changed the kind sorter to not sort by name (if other attributes are equal), since that would be against the idea of keeping the original file order as suggested in the PR.

bacongobbler (approved):
LGTM! Thanks for working on this. Needs another maintainer's review before we can merge this.

thomastaylor312 (approved):
This is looking great! Thank you for the well written test coverage

tmshort:
Will this be in 3.0.3?

hickeyma:
@tmshort As it is a feature, it will probably be in 3.1.0

tmshort:
I would argue that it's a bug; without this fix ("fix" is in the title), it prevents reasonably diff'ing the output of different runs of `helm template`. I'm seen cases where identical input to `helm template` (specifically for istio) have differences of several thousand lines due to the mis-ordering.

AndiDog:
I'm not seeing consistent output in 3.1.0-rc.3.

@bacongobbler @thomastaylor312 Was this feature actually merged into the release candidate?

AndiDog:
Seems to work in v3.1.0

hickeyma:
@AndiDog It looks like it was added as per the change log: https://github.com/helm/helm/releases/tag/v3.1.0

PR #6845, opened 2019-10-30
ref(chart): replace loader.BufferedFile with chart.File

bacongobbler:
Using Go's gradual code migration feature to deprecate BufferedFile. load_test.go continues to work even with this refactor, and was left alone to test backwards compatibility.

### DIFF
--- a/pkg/chart/loader/load.go
+++ b/pkg/chart/loader/load.go
@@ -29,6 +29,12 @@ import (
 	"helm.sh/helm/v3/pkg/chart"
 )
 
+// BufferedFile represents a file as a name/value pair.
+//
+// By convention, name is a relative path within the scope of the chart's
+// base directory.
+type BufferedFile = chart.File
+
 // ChartLoader loads a chart.
 type ChartLoader interface {
 	Load() (*chart.Chart, error)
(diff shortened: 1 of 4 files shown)


thomastaylor312 (approved):
Yay for refactors!

pkg/chart/loader/load.go
```diff
@@ -29,6 +29,14 @@ import (
 	"helm.sh/helm/v3/pkg/chart"
 )
 
+// BufferedFile represents a file as a name/value pair.
+//
+// By convention, name is a relative path within the scope of the chart's
+// base directory.
+//
+// DEPRECATED: use chart.File instead
+type BufferedFile = chart.File
```
adamreese: Why have a deprecation in a x.x.0 release?

bacongobbler: Not technically deprecated. The user can still choose to use either field.

Would a better comment message be

```
// TODO: remove this in Helm 4
```

?

adamreese: Why not remove all together?

thomastaylor312: My guess is that now that we are RC, the time for breaking changes is past

bacongobbler: SGTM. Will remove.

bacongobbler:
There's no real gain from this refactor. Closing.

PR #6860, opened 2019-11-01
v3: fix(comp): Protect against user's aliases

marckhouzam:
If a user has aliases commands that we use in the completion script
such as grep, cut, tail, it may cause the script to misbehave.

By escaping the commands, we tell the shell not to use any aliases.


bacongobbler (approved):

marckhouzam:
Thanks @bacongobbler

PR #6866, opened 2019-11-01
fix(chart): add JSON tags to chart object

bacongobbler:
Go capitalizes field names by default. Release metadata is JSON-encoded. This wasn't an issue in Helm 2 because everything was encoded in gRPC protocol buffers over the wire. In Helm 3, the client creates the release object directly from the Release struct.

To replicate:

```
helm create foo
helm install foo ./foo
kubectl get secret sh.helm.release.v1.foo.v1 -o jsonpath='{.data.release}' | base64 -d | base64 -d | gunzip
```

Marking as a last-minute 3.0 fix as changing this after 3.0 would break compatibility.

### DIFF
--- a/pkg/chart/chart.go
+++ b/pkg/chart/chart.go
@@ -27,18 +27,18 @@ const APIVersionV2 = "v2"
 // optionally parameterizable templates, and zero or more charts (dependencies).
 type Chart struct {
 	// Metadata is the contents of the Chartfile.
-	Metadata *Metadata
+	Metadata *Metadata `json:"metadata"`
 	// LocK is the contents of Chart.lock.
-	Lock *Lock
+	Lock *Lock `json:"lock"`
 	// Templates for this chart.
-	Templates []*File
+	Templates []*File `json:"templates"`
 	// Values are default config for this template.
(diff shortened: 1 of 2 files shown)


hickeyma (approved):
LGTM, thanks @bacongobbler

hickeyma:
@bacongobbler Maybe worth another maintainer also taking a look?

karuppiah7890:
Just curious, why is this called a fix? Does capital case break anything?

karuppiah7890:
And in any case, I would recommend adding tests for this, or things might be changed by mistake later

hickeyma:
@karuppiah7890 Naming consistency for json object field names.

hickeyma:
But a good point on what does it break?

bacongobbler:
serializing and de-serializing release object prior to this change. This is a backwards compatibility breaker.

bacongobbler:
> Maybe worth another maintainer also taking a look?

Yeah, I can get @mattfarina or @thomastaylor312 to review this.

thomastaylor312 (approved):
Been wanting to do this change, so should be good to go

PR #6868, opened 2019-11-01
fix(version): implement `helm version -c`, mark as hidden

### LINKED ISSUE #6322: Feature request: a `helm version` option that works consistently across v2 and v3

bacongobbler:
This has been requested by multiple users, and incurs no additional maintenance on our end, so I'm re-opening this for further discussion/debate.

Users are struggling to find a solution to this problem. No alternative solutions have been reached. This is the best solution as discussed at this time. IMO I'd rather have a half-baked solution that works for community members than no solution. Having this in for Helm 3 will reduce friction for adopting Helm 3 in third party tools like the [VSCode Kubernetes Extension](https://github.com/Azure/vscode-kubernetes-tools) and in [helmfile](https://github.com/roboll/helmfile).

this is identical to #6670, just re-opened against `master` after the big branch switch.
closes #6322

### DIFF
--- a/cmd/helm/testdata/output/version-client-shorthand.txt
+++ b/cmd/helm/testdata/output/version-client-shorthand.txt
@@ -1,0 +1,1 @@
+version.BuildInfo{Version:"v3.0+unreleased", GitCommit:"", GitTreeState:"", GoVersion:""}
--- a/cmd/helm/testdata/output/version-client.txt
+++ b/cmd/helm/testdata/output/version-client.txt
@@ -1,0 +1,1 @@
+version.BuildInfo{Version:"v3.0+unreleased", GitCommit:"", GitTreeState:"", GoVersion:""}
--- a/cmd/helm/version.go
+++ b/cmd/helm/version.go
@@ -61,6 +61,8 @@ func newVersionCmd(out io.Writer) *cobra.Command {
 	f := cmd.Flags()
 	f.BoolVar(&o.short, "short", false, "print the version number")
 	f.StringVar(&o.template, "template", "", "template for version string format")
+	f.BoolP("client", "c", true, "display client version information")
+	f.MarkHidden("client")
 
 	return cmd
 }
--- a/cmd/helm/version_test.go
+++ b/cmd/helm/version_test.go
@@ -32,6 +32,14 @@ func TestVersion(t *testing.T) {
 		name:   "template",
 		cmd:    "version --template='Version: {{.Version}}'",
 		golden: "output/version-template.txt",
+	}, {
+		name:   "client",
+		cmd:    "version --client",
+		golden: "output/version-client.txt",
+	}, {
+		name:   "client shorthand",
+		cmd:    "version -c",
+		golden: "output/version-client-shorthand.txt",
 	}}
 	runTestCmd(t, tests)
 }


marckhouzam:
I just want to emphasize that if the `--client` flag is added to v3, I think it should be in 3.0.0 and no later.  Adding it in a future release would cause backwards-compatibility issues for people who will have written code that determines the v3 version based on the absence of this flag.

hickeyma:
> Users are struggling to find a solution to this problem. No alternative solutions have been reached. This is the best solution as discussed at this time. IMO I'd rather have a half-baked solution that works for community members than no solution.

@bacongobbler Sorry but I disagree with this.

`helm version` works in both v2 and v3. The difference in v3 is that the `--client` flag does not exist.  This flag is described as follows:
` -c, --client                Client version only`.
It does not make sense in v3 as there is no client/server architecture.

As described by @thomastaylor312 in https://github.com/helm/helm/pull/6670#issuecomment-542706306:
This PR would be adding a feature to help one specific use case that can still be easily worked around, albeit in a slightly verbose/clunky fashion.

scottrigby:
@bacongobbler def agree we want to make v2 -> v3 adoption as easy for users as possible, as long as it's easy/maintainable.

I'm curious to hear replies to this question https://github.com/helm/helm/issues/6322#issuecomment-549263504, but we at least have the OP use case in #6322. If I understand that issue correctly, the issue with `version` isn't just that there's no longer a `--client, -c` option, but also that the output is different from v2, so any automation that relies on parsing the output format will fail. The PR currently still outputs a different format:

```console
$ helm version -c
Client: &version.Version{SemVer:"v2.15.2", GitCommit:"8dce272473e5f2a7bf58ce79bb5c3691db54c96b", GitTreeState:"clean"}
$ bin/helm version -c
version.BuildInfo{Version:"v3.0+unreleased", GitCommit:"444d006fe27bdea5ade0333fdb78c2141de2f175", GitTreeState:"clean", GoVersion:"go1.13.3"}
```

```console
$ helm version -c --short
Client: v2.15.2+g8dce272
$ bin/helm version -c --short
v3.0+unreleased+g444d006
```

Question: Is it easy/maintainable to reformat this to exactly match v2 format when the hidden `--client, -c` option is used? I'd guess we'd have to add some cruft to make that happen, but I haven't looked closely at the output format code changes yet, so I don't want to assume.

If making the output match exactly isn't not easy to do/maintain, then I'm not sure this is worth doing this part-way, since we can't know how users custom automation is parsing the output.

If that is easy/maintainable, then cool 😅. My only thought in this case is maybe note in the helm text that this is for backwards compatibility to ease v3 adoption?

bacongobbler:
> If I understand that issue correctly, the issue with `version` isn't just that there's no longer a `--client, -c` option, but also that the output is different from v2 [...]

The `--template` function allows them to reformat the output. It is not our goal to change the output to match Helm 2, nor would it be easy to do so without re-introducing a ton of code that'd be DOA.

> This PR would be adding a feature to help one specific use case that can still be easily worked around, albeit in a slightly verbose/clunky fashion.

That's the point. The slightly verbose/clunky fashion is a big friction point that users will require working around. The goal here is to ease the transition from Helm 2 for those users.

I see you and @thomastaylor312's point about Helm 3 removing the client/server architecture so this flag really has no place in Helm 3.

BUT,

- the code introduced here is 2 lines of code, and 10 lines of test output
- it doesn't affect the behaviour of Helm in any shape or form
- it's hidden from the CLI output so new users will not be confused by this flag
- it has been requested by several community projects to ease the migration path.

The benefits here seem like a much bigger win for users (ease of transition without the use of a hacky error code check) than the trade-off for us (we continue to "maintain" two lines of code to support the `--client-only` flag in Helm 3), which is why I want to bring back the discussion topic.

We are setting a new precedent here: is it the responsibility of Helm to contain hacky workarounds in the code base to help users with a use case, or do we put that burden on our users?

mattfarina:
Here's my 2  cents...

Helm, and most software, is not a mathematical solution seeking the perfectly balanced equation. It's a software solution used by people and their other tools. What makes software successful is the way it is able to help people. We need to keep that in mind.

Look at how the PHP community has handled their upgrade paths and compare that to the way the python community has handled the change from 2 -> 3. There is a valuable lesson there in working with the community and making their work simple.

With this in mind, I'm for this change going in. I would go so far as to call this a user experience bugfix.

mattfarina (approved):

hickeyma:
It's not ideal but look if it helps migration for some users then lets go with it.

hickeyma (approved):

Issue #2492, opened 2017-05-24
Proposal: Allow templating in values.yaml


benjigoldberg:
Hello Helm Team,

I would like to propose adding limited template rendering capabilities to the `values.yaml` file. I believe this would be a valuable feature for interacting with subcharts.

For example, lets say I am defining a Chart for a web application deployment and it has a simple values file as such:

```
image:
    repository: xyz
    tag: latest
    pullPolicy: Always
port: 80
```

And then I wanted to bundle this service alongside the `nginx-ingress` controller, so I add the dependency and a few settings, including one which tells the [kubernetes-incubator/external-dns](https://github.com/kubernetes-incubator/external-dns) controller how to route traffic to the nginx service:

```
image:
    repository: xyz
    tag: latest
    pullPolicy: Always
port: 80

nginx-ingress:
    enabled: true
    controller:
        service:
            annotations:
                external-dns.alpha.kubernetes.io/hostname: {{ .Release.Name }}.example.com
```

As it stands today, this is impossible to accomplish automatically. A user is required to `helm upgrade <release-name> chart-name --set nginx-ingress.controller.service.annotations.external-dns.alpha.kubernetes.io/hostname=<release-name>.example.com` _after_ the initial installation because the release name is not accessible at installation as a top-level variable.

In my mind the variables available at this level would be extremely limited, but include some of the common variables in `.Chart` and `.Release`

I would be happy to participate in adding this functionality to helm should the proposal be accepted.

seh:
Are you proposing interpolating the literal values from this hypothetical _values.yaml_ file into the chart's templates, and then interpolating those whole templates again to expand the release name (to use your example), or do you imagine interpolating the _values.yaml_ file first, and then treating its expanded form as the input for interpolating the rest of the chart's templates?

technosophos:
This has been discussed at length before, and for a variety of technical reasons it is problematic. But a far less problematic solution would be to support expansion of things that look and behave like environment variables.

So, for example, you'd be able to do `myval: ${RELEASE_NAME}.foo.bar` in your values.yaml.

On the Golang side, it should be as simple as calling `os.Exand()` with a custom list of variables that should be expanded.

To @seh's point, it is a little tricky to decide _when_ you want to run the expansion. I would think you'd want to do it after the values are coalesced, but before the values are sent into the template engine. That would bring with it the caveat that variables would have to be represented as well-formed YAML. But I think that is workable

benjigoldberg:
[...]

Issue #2133, opened 2017-03-17
Support for using {{ values }} within values.yaml


DrEsteban:
For example:

**values.yaml**
```
myApp:
  fooServiceName: {{ .Release.Name }}-foo-service
```

It seems like it would be a common pattern to incorporate {{ .Chart.Name }} or {{ .Release.Name }} into some variables.  I got the following when I tried this:

```
error converting YAML to JSON: yaml: line 15: did not find expected key
```

### Edit
To better illustrate my issue, consider a Chart with the following structure:

**serviceA Chart directory structure:**
```
charts/
    serviceB/
        templates/
            serviceB.yaml
        Chart.yaml
templates/
    serviceA.yaml
Chart.yaml
<etc...>
```

**serviceB.yaml**
<pre>
apiVersion: apps/v1beta1
kind: Deployment
metadata:
  name: <b><i>{{ .Release.Name }}-service-b</i></b>
spec:
  <...spec...>
</pre>

--------------
Now, currently if I want `serviceA` to make calls to `serviceB`, I would have to do something like this:
**serviceA.yaml**
<pre>
apiVersion: apps/v1beta1
kind: Deployment
metadata:
  name: {{ .Release.Name }}-service-a
spec:
  containers:
    - env:
        - name: SERVICE_B_HOSTNAME
          value: <b><i>{{ .Release.Name }}-{{ .Values.serviceB }}</i></b>
</pre>
**values.yaml**
<pre>
serviceB: <b><i>service-b</i></b>
</pre>
--------------
When I'd _like_ to be able to do something like this:
**serviceA.yaml**
<pre>
apiVersion: apps/v1beta1
kind: Deployment
metadata:
  name: {{ .Release.Name }}-service-a
spec:
  containers:
    - env:
        - name: SERVICE_B_HOSTNAME
          value: <b><i>{{ .Values.serviceB }}</i></b>
</pre>
**values.yaml**
<pre>
serviceB: <b><i>{{ .Release.Name }}-service-b</i></b>
</pre>
--------------

technosophos:
We have talked about supporting these as environment variables (e.g. `$RELEASE_NAME`, `$CHART_NAME`, etc). 

However, the file that is passed into the template engine as a source of data for templates is itself not passed through the template engine. We almost definitely will not do that because it gets very confusing for users. It also breaks both standard YAML compatibility and backward compatibility with all existing Helm versions.

thomastaylor312:
[...]

Issue #4535, opened 2018-08-26
RFC: an apparatus to execute a template in a subchart's scope


intelfx:
There seems to be a popular demand ([SO#47791971](https://stackoverflow.com/questions/47791971/how-can-you-call-a-helm-helper-template-from-a-subchart-with-the-correct-conte), #4314, #2133 if you de-X-Y it, and some other issues I forgot IDs of) to pass computed values from subcharts to parent charts. Indeed, the use-case is obvious: if we have a chart that depends on a subchart that adheres to the current best practice of scoping its resource names with release name (e. g. the community-maintained [`redis`](https://github.com/helm/charts/tree/master/stable/redis) chart), we quickly run into a problem of referring to the subchart's resources (e. g. services) by their (dynamically generated) names.

Continuing with the redis example, the [redis master service](https://github.com/helm/charts/blob/master/stable/redis/templates/redis-master-svc.yaml#L4) has a name of `{{ template "redis.fullname" . }}-master`, where the [`redis.fullname`](https://github.com/helm/charts/blob/master/stable/redis/templates/_helpers.tpl#L21) template is 11 lines of logic.

* We could merely copy that template into the main chart, adjusting it to run from the parent scope, but this is a glaring violation of DRY and is prone to breakage so we won't do that.
* We could set up a `redis.fullnameOverride` in our `values.yaml`, but if we do this, we won't be able to scope the fullname with the release name (because `values.yaml` is not templateable and will never be), thus breaking the best practice.
* Neither we can simply call it from the parent chart because the `.` of the parent chart is different (e. g. `Chart.Name` is not `redis` anymore).

Therefore, I propose a middle-ground solution: the dot object (`.`) will gain a new key, e. g. `Subcharts`, whose value will be a dictionary containing subcharts' dot objects, keyed by the subcharts' names relative to this chart.

With this change, we will be able to compute a subchart's service name from the parent chart by invoking as follows:
```
{{ template "redis.fullname" .Subcharts.redis }}-master
```

intelfx:
For anyone looking an interim solution, I wrote a meta-template that calls another template in a subchart's ersatz-scope by synthesizing a dot-object. It is **incomplete** (e. g. `.Chart` lacks any fields except `Name`, there is no `.Files` because doing it properly means rescoping to the subchart's directory and I never bothered to do it) but it works for common helper templates:

```
{{- define "call-nested" -}}                                                                                                                                                                                                     
[...]

Issue #3920, opened 2018-04-18
Usable sub-charts templates from parent chart


thomas-riccardi:
~~Templates are global are recursively namespaced~~ Named templates are global, so the guideline is to namespace them with the chart name to avoid collision.

This leads to the natural desire to use sub-charts templates from parent charts. But it doesn't work because `$.Charts` refers to the current chart (maybe `$.Templates` too? and others?), and I haven't found a way to reconstruct a context equal to the one used in the sub-chart.

# Use-case
My use-case is to get the service name of a sub-chart.

- chart `A` has sub-chart `postgresql`
- would like to use `{{ template "postgresql.fullname <some_context> }}` to know its Service name, to use postgresql.

`<some_context>` could be something like `(subchartContext "postgresql")` (or `(subchartContext "A.postgresql")`) (and support recursion)

# Workaround
The current workaround is to ~duplicate the template definition in the parent chart, which is not future-proof: the sub-chart internal implementation becomes a public API and cannot be easily changed, and if it happens the error is unclear.

I found workaround in official charts (https://github.com/kubernetes/charts), but not in the helm documentation; it could be documented. 
**UPDATE**:
Example: 
`stable/gitlab-ce` depends on `postgresql` and redefines `gitlab-ce.postgresql.fullname` by copying `postgresql.fullname` and adapting it (replacing `.Chart` by hardcoded values)
https://github.com/kubernetes/charts/blob/78c4b0163065699f4be8de1e795f4cbae74a98db/stable/gitlab-ce/templates/_helpers.tpl#L18-L24
It has already diverged from its source:  
https://github.com/kubernetes/charts/blob/78c4b0163065699f4be8de1e795f4cbae74a98db/stable/postgresql/templates/_helpers.tpl#L9-L25

Some existing discussion: 
- https://github.com/kubernetes/charts/issues/44#issuecomment-284420598
- https://stackoverflow.com/questions/47757968/how-to-reference-a-value-defined-in-a-template-in-a-sub-chart-in-helm-for-kubern

The larger use-case is to easily use sub-charts: 
instantiate sub-charts, use them from parent chart, make 2 sub-charts talk to each other.

- instantiate sub-chart: OK
- get resource names of the sub-chart from the parent chart template (Service, ConfigMap, Secret): 
  - workaround somewhat OK by duplicating sub-chart templates
  - longer term solution would require sub-chart explicitly declaring its API: 
    - expose rendered values for resources names
[...]

PR #6876, opened 2019-11-04
feat(helm): Adding values templates in order to customize values with go-template, for the chart and its dependencies

aure-olli:
There have been many requests for a way to use templates for chart `values.yaml` (#2492, #2133, ...). The main reason being that it's currently impossible to derive the values of the subcharts from templates. However, having templates in `values.yaml` make them unparsable and create a "chicken or the egg" problem when rendering them.

This merge request creates an intuitive solution without such a problem: an optional `values/` directory which templates are rendered using `values.yaml`, and then merge them with it. Those `values/`  templates are only required for specific cases, work the same way as `templates/` templates, and `values.yaml` will remain the main parsable source of value.

The rendering order has also been changed to allow those new values to enable or disable dependencies, and to avoid rendering values templates of disabled dependencies.

## New possibilities:

- Can now customize dependencies values, which was previously totally impossible
    ```
    ### values/subchart.yaml ###
    subchart:
        fullnameOverride: subchart-{{ .Relese.Name }}
        debug: {{ default "false" .Values.debug }}
    ```
- Can derive dependencies conditions from values
    ```
    ### Chart.yaml ###
    dependencies:
    - name: subchart
      condition: subchart.enabled
    
    ### values/subchart.yaml ###
    subchart:
    {{- if eq .Values.environment "production" -}}
      enabled: true
    {{- else -}}
      enabled: false
    {{- end -}}
    ```
- Can handle more smoothly default arguments
    ```
    ### values.yaml ###
    environment: development
    replicas: auto

    ### values/default.yaml ###
    {{- if eq .Values.replicas "auto" -}}
    {{- if eq .Values.environment "production" -}}
    replicas: 3
    {{- else -}}
    replicas: 1
    {{- end -}}
    {{- end -}}
    ```
- Can avoid using templates and variables in `template/` files
    ```
    ### values/helper.yaml ###
    {{- $name := default .Chart.Name .Values.nameOverride -}}
    fullname: {{ printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" }}

    ### templates/service.yaml ###
    apiVersion: v1
    kind: Service
    metadata:
    {{- if .Values.service.name }}
    name: {{ .Values.service.name }}
    {{- else }}
    name: {{ .Values.fullname }}
    {{- end }}
    ```
## Compatibility:

The behavior of helm is unchanged.
However one side effect has disappeared: the values of disabled dependencies are non longer read. This is a side effect which has no practical usage, and should be unnoticed.

## Behavior:

Values evaluation is now made in 2 step. First the `values.yaml` file is read. Then the `values/***.yaml` files are rendered using those values (`values/***.tpl` files ara still allowed as helpers), and are used to update the values.

For each chart and subchart, the following steps are performed:
1. If the chart has a parent, parent sub-values and global values are retrieved.
2. The chart values are coalesced with the retrieved value, the chart values are not authoritative.
3. The chart `values/` templates are rendered, the same way `templates/` templates are. The values are not updated during this step.
4. The values are updated with the rendered `values/` templates, sorted by name. The `values/` templates are authoritative and can replace values from the chart `values.yaml` or the previous `values/` templates.
5. The dependencies conditions and tags are evaluated over the updated values, and disabled dependencies are removed.
6. The same process is performed on enabled dependencies.

Once the all render values are recursively updated, they are used for templates rendering.

## Other code modifications:

- Conversion between `map[string]interface{}` and `chart.Values` works usually, but fails when converted to  `interface{}` first (for instance if stored as a map value). Thus, Most usages of `chart.Values` have been removed to avoid ever storing a `chart.Values` in map values. In particular, most functions now return a `map[string]interface{}`. Fixes trying to handle this problem have been removed too.

- `chartutil.ToRenderValues` no longer coalesces the content of the chart, it just creates a render values object with the options passed.

- Abusive usages of `chartutil.CoalesceValues` in actions have been replaced by `chartutil.CoalesceRoot` (`actions/package.go`, `actions/upgrade.go`), which only coalesces with the root chart.

- `chartutil.ProcessDependencyEnabled` is no longer recursive and `chartutil.ProcessDependencies` has been removed.

- `engine.Engine.updateRenderValues` now take cares of recursively both coalescing the values and filtering dependencies (see Behavior).

## Tests

All the concerned tests has been updated, many tests now have a `values/` templates, and new tests have been created for checking that the intended behavior is respected.

## Conclusion

I hope this merge will give a satisfying solution to this long lasting problem.


corlettb:
This sounds amazing! +1

aure-olli:
Another usage: generating random passwords that can be used across several templates

Before:
```
### templates/password.tpl ###
{{- define "password" -}}
{{ randAlphaNum 32) }} # every usage of this template will generate a new password
{{- end -}}
```
After:
```
### values/password.taml ###
password: {{ randAlphaNum 32) }} # this password will be the same every time the value is read
```

bacongobbler:
Hi @aure-olli. Can you please address the issues raised with this design as explained in https://github.com/helm/helm/issues/2492#issuecomment-413635332? By the looks of it, those concerns were not addressed in this PR, so I'm not sure what is the expected outcome here. We've shared our concerns about the design of a templated values.yaml in great detail. In particular, backwards compatibility and circular dependency issues are top of mind.

bacongobbler (changes requested):
See comment above^

aure-olli:
Hi. Here is a clearer answer to those issues.

**The tl;dr**  
There is a value file (as regular YAML), and complementary values templates. The value file (`values.yaml`) is not rendered in any way. The values templates (`values/*.yaml`) are rendered using the value file only (and the value files and templates of the parent chart), and then update the overhaul values, potentially overriding them. At each step, it is clear what are the values.

**Rendering order and logic**  
There is no circular logic, only a 2 step generation. Step 1, generate `values/*.yaml` using `values.yaml` and updates the values. Step 2, generate `templates/*.yaml` using the updated values. The solution is currently to use a lot of `{{ define "xxx" }}`, which does involve circular logic (those definitions could potentially be circular, and are spread around several `*.tpl` files), poorer readability, and inconsistency when using random. These are the 68 lines of code that try to solve this `define` problem: https://github.com/helm/helm/blob/master/pkg/engine/engine.go#L171-L238

**Backwards compatibility**
Don't use any `values/*.yaml` and the behavior is identical. My code is passing all the previous tests. The only "difference" is on a side effect: the values of all subcharts are currently read, even when disabled. `values/*.yaml` provide a way to deal with problems you cannot normally solve, but are not the main source of values.

**Language complexity**
`values/*.yaml` are in all points identical to `templates/*.yaml`. Nothing to learn, it's still YAML generation using go-templates.

**Obsolete by Helm 3**
Well, Lua scripts are not yet implemented. And they do add language complexity. I'm not saying they will be useless, they will probably give solutions to problems that require a real programming language. But they will add another layer of complexity and unreadability in the templates. My solution only relies on what everybody is currently using, by adding a step before template generation.

Hope this satisfies your concerns :)

aure-olli:
Please let me know what other elements you need to consider this update.

dudicoco:
What is the status of this PR? This is a great feature!

bacongobbler:
The PR has over 1,400 lines of code being changed. It's going to take some time to review.

mstavrev:
Any update on this one, or another alternative that can be used meanwhile?

bacongobbler:
nope. feel free to test this Pr and provide feedback!

mstavrev:
@aure-olli, can you please update the PR, so it is compatible with latest changes on the master? I am planning to do some tests and report back.

aure-olli:
Updated
I have several failing tests, but none that seem directly related to my PR, so I suppose it only happens on my computer.

aure-olli:
merged again

@mstavrev please let me know if you need any information.

TBBle:
It's a big PR, but a lot (a majority by line-count, I suspect) is tests and test-data.

Is there any chance this could be broken up into separate changes? If possible, perhaps the fix for `TestLintCmdWithSubchartsFlag`, the fix to not process values from disabled subcharts, and the changes about `map[string]interface{}` and `chartutil.Values` could each be separated out, reviewed, and landed more-easily.

The `map[string]interface{}` change seems like it might be independently-motivated, as I recently bounced off a problem where I wanted to pass a dictionary of parameters to a named template, one of which would be `.Values` so that I could pass that on to a `tpl` call in that named template. However, by passing it through the dict, my `.Values` lost its type and so the `tpl` call rejected it.

In the end, the workaround was to make the dict passed to the named template be `.` merged with the named template's parameters. See [the template](https://github.com/googleforgames/open-match/blob/94d2105809a351315e7dc9c3ab411e5de8c53c19/install/helm/open-match/templates/_helpers.tpl#L76-L82) and [a call-site](https://github.com/googleforgames/open-match/blob/94d2105809a351315e7dc9c3ab411e5de8c53c19/install/helm/open-match/templates/backend.yaml#L85).

It's been a while, but I recall that I wanted the call-site to be
```
include "openmatch.volumes.configs" (dict "configs" .Values.configs "dollar" .)
```
but there was some reason that didn't work so my fallback was
```
include "openmatch.volumes.configs" (dict "configs" .Values.configs "Values" .Values "Chart" .Chart "Release" .Release ...)
```
which is where I recall this problem occurring, because I couldn't rebuild an acceptable `dict` to pass to `tpl` as `.Values` had become `interface{}` when passing through the generated `map[string]interface{}`.

Either way is closer to how it was [before I touched it](https://github.com/googleforgames/open-match/commit/79e9afeca781457cbcf952ee520c4aafed0ff69f#diff-9aef976c5da058ebc51e9717416080dfL85), and I'm mildly unhappy with the final result:
```
include "openmatch.volumes.configs" (. | merge (dict "configs" .Values.evaluatorConfigs))
```

intelfx:
@aure-olli, this is a great feature and I've long wished for something like this. I'd like to clarify a few things to make sure this feature will be applicable for my use-cases.

1. As I understand, all `values/*.yaml` are rendered strictly in context of `values.yaml`. Would it be feasible instead to e. g. sort all `values/*.yaml` by name, then render them one by one and merge each yaml into `.Values` _before_ rendering the next one?

2. I'd like to confirm the global order of evaluation for classic values and "computed values" in presence of chart dependencies. IMO, computed values should be rendered in reverse order, so charts would be able to use computed values to generate values for their dependencies.  
For example, ff we have a chart dependency graph of `A -> B -> C` (i. e. A depends on B depends on C), I suggest that the values should be parsed and merged in the following order:
   - C/values.yaml
   - B/values.yaml
   - A/values.yaml
   - A/values/*.yaml
   - B/values/*.yaml
   - C/values/*.yaml

I'd appreciate if you could comment on the above.

aure-olli:
@TBBle Thanks for recent update, I was despairing that this PR would ever be considered (it's closed to one year old already).

About breaking that into pieces, I guess I can come with 3 pieces:
- As mentioned in the first commit, abandoning `chartutil.Values` was motivated by the fact that golang cannot convert `map[string]interface{}` => `interface{}` => `chartutil.Values`, despite `chartutil.Values` being `map[string]interface{}`. It was responsible for many confusing explicit conversions across the code (including a `chartutil.Values.asMap` function), whereas `chartutil.Values` had no other interest than being a shortcut.
- One big part of the code was breaking the workflow to compile each sub chart one by one, instead of as one big batch that even ignored if those should actually be activated in the first place.
- The rest is just interceding the `values/*.yaml` inside the new workflow.

However, I've been maintaining mergeability over 6 months for people claiming they would review my code. I feel discouraged to split my whole code if the result will be the same.

@intelfx Honestly the order of the templates doesn't matter much, as much use case should not imply conflicting `values/*.yaml`. The current sorting function is 
```golang
// sortValuesTemplates sorts the rendered yaml values files from lowest to highest priority
func sortValuesTemplates(tpls map[string]string) []string {
	keys := make(sort.StringSlice, len(tpls))
	i := 0
	for key := range tpls {
		keys[i] = key
		i++
	}
	sort.Sort(keys)
	// sort.Sort(sort.Reverse(keys))
	return keys
}
```
As you can see, this is a stupid alphabetic sorting, so the latest name will replace anything declared before. I don't event know how golang sorts `A.yaml` compared to `A/A.yaml`. If you can come with a more clever sorting, you're welcome.

I however am skeptic about compiling each `values/*.yaml` with the knowledge of the previous ones. I think it's much safer that all you need to care about is `values.yaml` when writting a `values/*.yaml`. The goal isn't to become turing complete ^^ `values/*.tpl` are still available for trickier (and probably rare) use cases.

intelfx:
@aure-olli Okay, let's ignore the first point about rendering each `values/*.yaml` with the knowledge of previous ones. It's not really a dealbreaker. I care a lot more about (2) — would you mind clarifying that part?

aure-olli:
@intelfx Oh right I misread the question ! The order is:
- read parent chart `values.yaml`
- render parent chart `values/*.yaml`
- merge parent chart `values/*.yaml` into `values.yaml`
- determine if subcharts should be rendered (`condition`, `tags`, ...)
- repeat on subcharts, on the the `.Values.{subchart}` object

So, to clarify:
- subcharts gets both `values.yaml` and `values/*.yaml` from the parent chart (that was the main purpose of this PR)
- subcharts only receive and update the `.Values.{subchart}` part of the values. So it doesn't matter which subchart is evaluated first, they won't receive and update the same parts of the values anyway.
- templates will receive the final values (updated by `values.yaml`, `values/*.yaml` and the subcharts).

intelfx:
@aure-olli That sounds encouraging. My main wish is that a subchart should be able to see the parent's rendered values (the relevant part of them, anyway) by the time it's rendering its own values.

aure-olli:
> @aure-olli That sounds encouraging. My main wish is that a subchart should be able to see the parent's rendered values (the relevant part of them, anyway) by the time it's rendering its own values.

It will be the case. That was the main purpose of this PR: finally being able to dynamically change the values of the subcharts.

TBBle:
I appreciate that breaking this up into multiple pieces is a hassle. My observation is that this is a 1400 line patch over 75 files (as was mentioned earlier by one of the Helm developers), and when I had a skim of the code, that was a misleading result of having several changes in one PR, one is type refactoring, one fixes a bug, and one is new behaviour, and I think the new behaviour is being seen as "hard to review" due to the large overall line count, even though it's probably the smaller of the three by line-count.

Particularly if the two parts that aren't a new behaviour are able to be reviewed and ideally pulled in sooner, then that also makes the behaviour-change PR easier to comprehend for reviewers, _and_ easier to carry if the behaviour itself is considered controversial.

If the commits were structured for review, i.e. rebased into individual changes rather than a "change + fix + fix + fix + fix" as they are now, that might help as well, but I suspect most reviewers are going to look at the PR as a whole rather than the individual commits anyway, until they want to know extra context for why something was done, so if it was me, I'd be looking at a set of stacked PRs, given the breadth of things touched here.

Also, I agree that the _values/*.yaml_ should be evaluated only with the input values.yaml and user parameters, not with the results of previous _values/*.yaml_.

I suspect merging order is already going to be slightly surprising (I would probably name them with numeric prefixes if I was going to rely on the merging order, and test it carefully if I was using subdirectories or other order-confounding things), so having the merging order _also_ affect value visibility seems doubly-risky, and would prevent parallelisation of the rendering, if that was ever desired.

Also, I didn't look at the implementation, but shouldn't it be _values/\_*.tpl_? From memory, the leading `_` is the clue to Helm that it's not generating an output.

aure-olli:
@TBBle OK I'll start working on splitting this into PRs.

For "fix + fix + fix + fix", I was kinda lazy to each time recreate a commit from the new master after each merge (which requires creating a diff and stuffs). It always can be done at any point anyway, clearly would have done it if a merge was planned.

Checking the code again, `_` is the only clue to not render a file:
https://github.com/helm/helm/blob/master/pkg/engine/engine.go#L243

I've kept all the parsing/rendering process anyway, just reordered its usage.

TBBle:
Discussing #3920, I've come across something that should be clarified here.

Using something like [`call-nested`](https://github.com/helm/helm/issues/4535#issuecomment-477778391), you are able to have circular interactions: If I use `call-nested` in my parent chart's _values/*.yaml_ to call a named template in a subchart, to set a value which is passed to the same subchart, then I am able to change an input to that named template based on its own output, which means evaluating the same template during _templates/*.yaml_ rendering will give a different result.

I'm also wondering (I haven't checked), if `call-nested` would even _work_ in a parent's _values/*.yaml_. Have the child values been populated _at all_ at that point? I suspect they're not, based on being able to disable sub-charts before their values are read...

For reference, call-nested is a pretty simple named template, which simulates calling a named template in the context of a subchart, by walking `.Values` for the given chart list.
```
{{ /* usage: include "call-nested" (list . "redis" "redis.fullname") */ }}
{{- define "call-nested" }}
{{- $dot := index . 0 }}
{{- $subchart := index . 1 | splitList "." }}
{{- $template := index . 2 }}
{{- $values := $dot.Values }}
{{- range $subchart }}
{{- $values = index $values . }}
{{- end }}
{{- include $template (dict "Chart" (dict "Name" (last $subchart)) "Values" $values "Release" $dot.Release "Capabilities" $dot.Capabilities) }}
{{- end }}
```

This is actually my major use-case for this work, I want to be able to include, e.g. an instance of MongoDB in the parent chart, and pass its Service name (extracted using `call-nested`) down to the various services that want a mongoDB URL, when the "batteries-included" configuration is enabled; and have a single mongoDB URL input for users running an external mongoDB, where they can pass the URL to the parent chart once, and it populates all the subcharts with appropriate mongoDB details.

The other option is that I have to always override things like the Service name for the MongoDB in the parent chart, and then pass that same value down to all the children that need a MongoDB URL. That's the fallback path, and it's not awful. I think it makes the parent chart codebase slightly more verbose, and I need to generate names for all such shared services, rather than just being able to ask the subchart what it's going to pick.

intelfx:
> I'm also wondering (I haven't checked), if call-nested would even work in a parent's values/*.yaml. Have the child values been populated at all at that point? I suspect they're not, based on being able to disable sub-charts before their values are read...

That's precisely why I wrote the above wall of text to confirm values loading/evaluation order. This will work, if and only if the values are loaded and merged in a "hairpin" order:
   - load and merge `C/values.yaml`
   - load and merge `B/values.yaml`
   - load and merge `A/values.yaml`
   - render and merge `A/values/*.{tpl,yaml}`
   - render and merge `B/values/*.{tpl,yaml}`
   - render and merge `C/values/*.{tpl,yaml}`

aure-olli:
Hi @TBBle @intelfx 

Wow what you want to do is another level of complexity. With subcharts A and B, you want a computed value of B to feed the values of A. treating subcharts `values.yaml` before `values/*.yaml` would probably have little interest, because `values.yaml` are not even dynamic.

Currently, it is possible using `import-values`:
```yaml
### values.yaml ###

subchartA: #...

subchartB:
  database: http://placeholder
```

```yaml
### Chart.yaml ###
dependencies:
- name: subchartA
- name: subchartB
  import-values:
  - child: databaseUrl
    parent: subchartA.database
```

```yaml
### charts/subchartB/values/database.yaml ###
databaseUrl: {{ .Values.schema }}://{{ .values.user }}:{{ .values.password }}@{{ .values.host }}:{{ .values.port }}
```

```yaml
### charts/subchartA/tempaltes/deployment.yaml ###
env:
- name: DATABASE_URL
  value: {{ .Values.databaseUrl }}
```
It works if the subchart A doesn't use `databaseUrl` in any of the `values/*.yaml`.

For a more general approach, we would need to:
- treat subcharts in a specific order, one being able to influence the other
- treat `values/*.yaml` in a specific order, after each subchart

Maybe could:
- follow the order defined in `Chart.yaml:dependencies`
- only treat `values/{{subchartName}}/*.yaml` once `{{subchartName}}` has been treated

That's not much change to my PR, basically needs to change the order `Engine.updateRenderValues` loads the values and perform the import values. If we still do `import-values` only at the end of all that, users won't notice a thing if they don't create any `values/{{subchartName}}/*.yaml`.

aure-olli:
OK all 3 PRs have been uploaded:
- #8677 for getting rid of the dangerous `chartutil.Values`
- #8679 for changing the way `values.yaml` and dependencies are treated (ie, to become recursive), this is clearly the heaviest one because it has to refactor some pieces of code
- #8690 for finally adding values templates

intelfx:
> With subcharts A and B, you want a computed value of B to feed the values of A.

Not really. I'd be glad with feeding static values of B to A. That would be enough for general use-cases of executing templates in subchart context. Feeding computed values of B to A will of course contradict everything else that's written before.

aure-olli:
> Not really. I'd be glad with feeding just the _static_ values of B to A. That would probably be enough for general use-cases of executing templates in subchart context (as it works now, cf. `call-nested`). Feeding _computed_ values of B to A will of course contradict everything else that's been discussed before.

The this is exactly what you can already do with `import-values`

intelfx:
Well, right now it works without import-values at all. Would be nice if it continued to be so.

aure-olli:
@intelfx I haven't touched the way templates are generated. If it worked before in templates, it should work now ! If you want to be sure this behavior is preserved along helm updates, write a test about it.

It cannot work in values templates because you need to generate those before ever deciding if if the dependency should be enabled in the first place.

bacongobbler:
Hi @aure-olli.

Yesterday, the core maintainers discussed what should be the next steps here, as it's been nearly 10 months since this PR was first opened. Now that it's being split into multiple PRs - all of which are HUGE changes to the codebase, and each with their own rationale and design concerns - it is quite possible that that these proposals will never be merged. We need to have a proper discussion on the design goals and the motivation for these large changes to the Helm codebase.

As discussed in the call, we decided it would be best if development on these PRs would stop for the time being, and instead, a proposal on the subject matter is written instead. This PR introduces a number of significant features to Helm, and there are a number of parties involved here, each with their own [questions](https://github.com/helm/helm/pull/6876#issuecomment-684794139). [Several](https://github.com/helm/helm/pull/6876#issuecomment-683815078) community members ([including myself](https://github.com/helm/helm/pull/6876#issuecomment-566163629)) have raised [concerns](https://github.com/helm/helm/pull/6876#issuecomment-686579535) about the design, and it would be best to address those concerns in a design document. That way we can have a proper discussion on the design.

Instead of trying to track the discussion over multiple github comments, it would be best to move this proposal forward by [writing out a HIP](https://github.com/helm/community/blob/master/hips/hip-0001.md). We intend HIPs to be the primary mechanisms for proposing major new features, for collecting community input on an issue, and for documenting the design decisions that have gone into the project. That way, all of the information, discussions, rationale, etc. are all contained in one document which we can all reference back to when reviewing the implementation, and we can refer back to the design document in case a regression is introduced.

Moving forward, we will be asking all community members to propose any significant features or suggestions as a HIP. This way, we have a proposal that

1. provides a high-level overview of the feature
1. provides the motivation behind writing this feature
1. contains a technical specification of the feature
1. describes what motivated the design, including discussions over alternative proposals or ideas
1. addresses concerns about backwards compatibility
1. produces a link to a reference implementation (this PR)

By writing out a proposal, we can then look back at the proposal and determine why this feature is necessary, and what parts of the implementation are necessary vs. what can be pulled out and reviewed separately.

Thank you.

bacongobbler:
> If you want to be sure this behavior is preserved along helm updates, write a test about it.

FYI, this is the responsibility of the PR author, not of a community member. It is the author's responsibility to ensure concerns about backwards compatibility are maintained.

bacongobbler (changes requested):
As this PR introduces a number of design concerns, a HIP is required before moving forward with the review.

intelfx:
> it is quite possible that that these proposals will never be merged

> As discussed in the call, we decided it would be best if development on these PRs would stop for the time being

I would just like to ask the Helm maintainers to consider and take into account that this is quite a demanded feature (the demand can be observed by counting instances of `tpl` in helm/charts).

In other words, I really hope this is not a polite form of "we're not interested" (i. e. death by bureaucracy).

bacongobbler:
We're very well aware that this feature has been highly requested, and is why we discussed how to move this proposal forward last Thursday. The old system of reviewing a PR and trying to irk out the design through comments wasn't scaling well, so we're trying something new.

aure-olli:
Hi, ok let me know when you settle on a contributing policy.

rabidscorpio:
@bacongobbler Not sure how this isn't death by bureaucracy. While I agree with the fact that the contributed PRs are large and include breaking changes, this PR has been around for 10 months so to suddenly say that development should stop and that a HIP should be created is disingenuous because it doesn't identify concerns that the core team have other than maybe backwards compatibility until that HIP is created. @aure-olli has created a version of this feature with documented reasons for why he chose the implementation he did that obviously doesn't sit well with the core team so If they were serious about adding this feature, they should be the ones describing how it should be implemented instead of making contributors guess what would be acceptable.

What should be happening is the core team should be creating the HIP (for such an in-demand feature as this) so that contributors can use it as a template for actually implementing the change. @aure-olli has been extremely responsive to requests for changes and he keeps getting shut down. This doesn't really encourage people to contribute and it makes it seem like the core team isn't really serious about at least adding this feature. As an example, in #8677 Helm 4 isn't even on the milestones page yet so who knows when that's even going to get looked at not to mention saying something like "it is quite possible that that these proposals will never be merged".

How about something a little more constructive?

UkonnRa:
Hi, just wondering what the status of this PR. Is it already released?

hickeyma:
@UkonnRa A proposal is requested to continue as described in https://github.com/helm/helm/pull/6876#issuecomment-687233515

bacongobbler:
I'll touch on a few points raised in https://github.com/helm/helm/pull/6876#issuecomment-701599486.

Yes, you are right that we should have stepped in a little sooner to discuss the design. We apologize for that. That is why we updated our [Contributing Guidelines](https://github.com/helm/helm/blob/master/CONTRIBUTING.md) to match the new process. I still need to add the HIP process to that document - I'll take that as an action item for today.

I still stand by the [original statement](https://github.com/helm/helm/pull/6876#issuecomment-687233515). The core maintainers reviewed the PR to decide how we should proceed. We agreed that a Helm Improvement Proposal is the right choice moving forward.

The first PR started out as a 1,400 line change. It changed how charts are loaded, how dependencies are rendered, how values are coalesced, and even how the rendering engine renders templates, breaking many function signatures or re-writing existing business logic. The PR was then split into three PRs (#8677, #8679, and #8690) which now changed over 2,500 lines of code.

It is logical to assume that many existing use cases would break with these proposed changes.

For a project like Helm, a PR as significant as this needs a plan to ensure the correct maintainers and community members are involved, a clear test plan is in place, and sufficient documentation has been written to educate users on its usage and how it's intended to work (or not work). That is what the HIP process is designed to help. It is NOT "death by bureaucracy" - it ensures everyone has the right tools in place to plan, test, and review significant changes to the Helm codebase.

> What should be happening is the core team should be creating the HIP (for such an in-demand feature as this) so that contributors can use it as a template for actually implementing the change.

The goal of a HIP is to raise design issues to the core maintainers and propose a solution that would resolve that issue. The community is more well-equipped to determine what design issues they have with the software and what solutions would work for them. It's the core maintainer's responsibility to determine how and where that proposal fits into Helm's design.

A HIP is not intended to be used as a "wish list" where maintainers propose their ideas and ask the community to write out the solution. That is why the author(s) of a HIP are responsible for presenting a "reference implementation" along with the proposal.

> As an example, in #8677 Helm 4 isn't even on the milestones page yet so who knows when that's even going to get looked at not to mention saying something like "it is quite possible that that these proposals will never be merged".

If you'd like to discuss a topic for Helm 4, please feel free to reach out either on the mailing list or during the weekly developer calls. Links to these communication channels are [in the README](https://github.com/helm/helm#community-discussion-contribution-and-support).

abdennour:
Why this PR took long time & at the end no merge ?
This is highly needed for many and many cases.
I will go with an argocd plugin that i will called helm-template-value-computing . 
I will make some conventions in the team to keep these files with .yaml.tmpl extension and let's see.
Since no owner one can listen here, our teams must afford this burden.
Good luck everyone!

abdennour:
Why this PR took long time & at the end no merge ?
This is highly needed for many and many cases.
I will go with an argocd plugin that i will call it helm-template-value-computing . 
I will make some conventions in the team to keep these files with .yaml.tmpl extension and let's see.
Since no owner one can listen here, our teams must afford this burden.
Good luck everyone!

ztripez:
To bad this feature is being buried, I would helped our setup a lot

Ogguz:
Would be so useful for us. RIP.

bacongobbler:
Locking the thread. I’ve [stated next steps](https://github.com/helm/helm/pull/6876#pullrequestreview-482774751). If someone wants to revive this work, please go and write a HIP so we can discuss the design before proceeding to an implementation. Thanks.

PR #6881, opened 2019-11-04
fix(cmd): Standardizes all output to use lower snake_case names

thomastaylor312:
After discussing similar changes in #6866, we decided to make sure all
JSON and YAML output uses lower snake case names as is generally standard practice

### DIFF
--- a/cmd/helm/list.go
+++ b/cmd/helm/list.go
@@ -110,13 +110,13 @@ func newListCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 }
 
 type releaseElement struct {
-	Name       string
-	Namespace  string
-	Revision   string
-	Updated    string
-	Status     string
-	Chart      string
-	AppVersion string
+	Name       string `json:"name"`
+	Namespace  string `json:"namespace"`
+	Revision   string `json:"revision"`
+	Updated    string `json:"updated"`
+	Status     string `json:"status"`
+	Chart      string `json:"chart"`
+	AppVersion string `json:"app_version"`
 }
 
 type releaseListWriter struct {
--- a/cmd/helm/repo_list.go
+++ b/cmd/helm/repo_list.go
@@ -51,8 +51,8 @@ func newRepoListCmd(out io.Writer) *cobra.Command {
 }
 
 type repositoryElement struct {
-	Name string
-	URL  string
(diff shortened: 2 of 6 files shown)


bacongobbler (approved):

databus23:
@thomastaylor312 Ahm, Am I reading this correct that with the change all releases that are created before will be invalid and can't be read? Breaking all exiting helm3 releases?

bacongobbler:
Yes.

databus23:
That is quite disappointing. We are already using helm3 semi productively . I don't understand how such a change can be justified in a late rc phase of a release.

bacongobbler:
I take full responsibility for the change. Apologies for introducing a breaking change so late into the release cycle. This was caught at the last minute and had to go in before 3.0 was released. We  weren't anticipating any breaking changes to the backend, but here we are.

The main issue had to do with marshaling and unmarshaling releases from Kubernetes. Some fields were capitalized, some were not. This inconsistency would be difficult for other users wishing to re-implement the storage backend in parallel projects. Go projects can use the SDK, but clients written in other languages would have to rely on the spec. Once we went 3.0, there would be no going back, and we'd have to live with this inconsistency, so we had to make the change *now* or never.

Again, we're sorry for not catching this sooner in the release cycle. It's disappointing but necessary to make it easier for third party clients in the future.

bacongobbler:
FYI #6866 is the change that introduced the breaking change on the storage backend. This PR only changes the output of the `-o json` for certain commands.

databus23:
@bacongobbler Thanks for the response. I looked into #6866 and after my initial shock I might think the change is not really backward incompatible:
According to the https://golang.org/pkg/encoding/json/#Unmarshal documentation it says:

> To unmarshal JSON into a struct, Unmarshal matches incoming object keys to the keys used by Marshal (either the struct field name or its tag), preferring an exact match but also accepting a case-insensitive match.

This means it should still be possible to deserialise old releases because no snake casing with introduced, only lowercasing the first letter.

This playground example show that it should indeed accept the literal field name even with a lower case son tag: https://play.golang.org/p/KlfhKHfAqMw

bacongobbler:
Oh! That's super convenient. I wonder if serializing/de-serializing fields in JSON should be considered case-insensitive. 

Phew!

PR #6891, opened 2019-11-06
fix(cli): Sort output of helm env

marckhouzam:
Before this change, running 'helm env' multiple times would generate
a different order for the output each time.  This is because Go
purposely loops over hash maps in a random order.

This commit sorts the environment variables by alphabetical order before
generating the output.


marckhouzam:
This one is a simple one to make the order of the output of `helm env` deterministic.
If a maintainer has some time.

hickeyma (approved):
LGTM, thanks @marckhouzam.  Good to merge.

Issue #6890, opened 2019-11-06
v3: Proposal: helm env to show XDG info


marckhouzam:
With the use of XDG directories for helm configuration, it has become difficult to find and automate where helm data can be found.  See https://github.com/helm/helm-www/issues/336 for a concrete example where I believe the user may find themselves at a loss.

**Point 1:**
I suggest extending `helm env` to provide the location of the current XDG paths in a machine consumable fashion.  Concretely I propose to add a `--show` (or some other name) flag to `helm env` like so:
```
$ helm env --show <cache|config|data>
```
which would print the corresponding path.
On Linux for example:
```
$ helm env --show data
 /home/marckhouzam/.local/share/helm
```
This would allow to tell users to do such things as:
```
cp ./myStarter $(helm env --show data)/starters
```
which to my knowledge has no other good alternative right now.

**Point 2:**
We could also have `helm env` return all 3 XDG paths in its output like so:
```
$ helm3 env
HELM_NAMESPACE="default"
[...]
XDG_CACHE_HOME="/Users/marckhouzam/Library/Caches"
XDG_CONFIG_HOME="/Users/marckhouzam/Library/Preferences"
XDG_DATA_HOME="/Users/marckhouzam/Library"
```
Note that just to be able to write the proper paths above I had to run `helm -h` and look at the default XDG locations.  This is another example that shows the issue that this proposal aims to address.

I can work on a PR if maintainers agree they find value.

marckhouzam:
Thinking about this some more, I believe that plugins must also suffer from this 'limitation'.  How would a plugin know where to find helm data if the XDG variables are not set in the shell? I guess it would need to figure out the default values itself, based on the current OS.  I think we should improve that.

I see that listing the XDG paths in `helm env` was actually part of the original `helm env` PR: https://github.com/helm/helm/pull/6244 (/cc @waveywaves )

However, it was purposely removed by https://github.com/helm/helm/pull/6362.
From the title of the PR, it makes sense to me that we don't want to "override users xdg directories".  However, I think it is because there was a bug in how 'helm env' would set these variables originally, where it would include the '/helm' suffix in the XDG vars, which is not correct.

I think if we strip that suffix, it would be safe to set the XDG vars in `helm env`.
 /cc @adamreese

bacongobbler:
> Thinking about this some more, I believe that plugins must also suffer from this 'limitation'. How would a plugin know where to find helm data if the XDG variables are not set in the shell?

[...]

Issue #6636, opened 2019-10-10
helm lint of packaged charts (tgz) not finding Chart.yaml with some SemVer2 compliant version values.


dvpierce:
If I set the "version" field in my Chart.yaml to a version with a pre-release value (e.g., 1.0.1-development) the helm lint of the directory will succeed, but the helm lint of the packaged tgz file (e.g., davesproject-1.0.1-development.tgz) will fail.

Replacing the hyphen in the version name with another permitted character (e.g., 1.0.1+development) and repacking (davesproject-1.0.1+development.tgz) works fine as a workaround, but I'm assuming this wasn't intentional.

Thanks,
Dave P.

```
[pierdavpa@lin2dv2ap55 _helm]$ helm lint davetestanother-1.0.4-development.tgz
==> Skipping davetestanother-1.0.4-development.tgz
No chart found for linting (missing Chart.yaml)

Error: 0 chart(s) linted, 1 chart(s) failed
[pierdavpa@lin2dv2ap55 _helm]$ helm lint davetestanother-1.0.4+development.tgz
==> Linting davetestanother-1.0.4+development.tgz
[INFO] Chart.yaml: icon is recommended

1 chart(s) linted, no failures
[pierdavpa@lin2dv2ap55 _helm]$
```

Output of `helm version`:
`Client: &version.Version{SemVer:"v2.14.3", GitCommit:"0e7f3b6637f7af8fcfddb3d2941fcc7cbebb0085", GitTreeState:"clean"}`

Output of `kubectl version`:
`Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.3", GitCommit:"2d3c76f9091b6bec110a5e63777c332469e0cba2", GitTreeState:"clean", BuildDate:"2019-08-19T11:13:54Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"linux/amd64"}`

Cloud Provider/Platform (AKS, GKE, Minikube etc.):  AKS/k8s hybrid

kroonprins:
Similar error here for "-rc1" version: https://github.com/kroonprins/helm-lint-error

kroonprins:
Was fixed in helm 2.15 by this commit https://github.com/helm/helm/commit/bfc0d76fff148865aa1e9d5ba28693da2947b4fe

PR #6892, opened 2019-11-06
feat(cli): Have helm env show current XDG paths

marckhouzam:
I believe plugins and users need to know where the XDG helm config paths are located.
To my knowledge there is no current good way of doing this when the XDG variables are unset in the user's environment.

This PR adds them to `helm env` which also makes them available to plugins.

It addresses concerns brought up in #6890

### DIFF
--- a/pkg/cli/environment.go
+++ b/pkg/cli/environment.go
@@ -26,6 +26,7 @@ import (
 	"fmt"
 	"os"
 	"strconv"
+	"strings"
 	"sync"
 
 	"github.com/spf13/pflag"
@@ -33,6 +34,7 @@ import (
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 
 	"helm.sh/helm/v3/pkg/helmpath"
+	"helm.sh/helm/v3/pkg/helmpath/xdg"
 	"helm.sh/helm/v3/pkg/kube"
 )
 
@@ -100,6 +102,9 @@ func (s *EnvSettings) EnvVars() map[string]string {
 		"HELM_REPOSITORY_CONFIG": s.RepositoryConfig,
 		"HELM_NAMESPACE":         s.Namespace(),
 		"HELM_KUBECONTEXT":       s.KubeContext,
+		xdg.CacheHomeEnvVar:      strings.TrimSuffix(helmpath.CachePath(), "/helm"),
+		xdg.ConfigHomeEnvVar:     strings.TrimSuffix(helmpath.ConfigPath(), "/helm"),
+		xdg.DataHomeEnvVar:       strings.TrimSuffix(helmpath.DataPath(), "/helm"),
 	}
 
 	if s.KubeConfig != "" {


mattfarina:
@marckhouzam One of the files in the PR isn't `gofmt`'d. Can you please fix the formatting issue.

marckhouzam:
> @marckhouzam One of the files in the PR isn't `gofmt`'d. Can you please fix the formatting issue.

The comment (// TODO) prevents the go formatter to behave as expected.  I'll remove the comment.

pkg/cli/environment.go
```diff
@@ -100,6 +102,9 @@ func (s *EnvSettings) EnvVars() map[string]string {
 		"HELM_REPOSITORY_CONFIG": s.RepositoryConfig,
 		"HELM_NAMESPACE":         s.Namespace(),
 		"HELM_KUBECONTEXT":       s.KubeContext,
+		xdg.CacheHomeEnvVar:      strings.TrimSuffix(helmpath.CachePath(), "/helm"),
```
hickeyma: Just wondering should this contain the `helm` suffix as its XDG paths related to Helm?

hickeyma (commented):
Thanks for PR @marckhouzam. I have a question around the path suffix.

pkg/cli/environment.go
marckhouzam: Good question.  Originally, the code had been merged including the suffix.  These variables were then removed by #6636 .  

I believe having the suffix creates two problems:

1. the printed variables (e.g. `XDG_CACHE_HOME`) are valid environment variable (not limited to Helm), and they don't contain the suffix, so printing them with the suffix can be confusing

1. more importantly, all the variables listed in `EnvVars()` are actually created with their specified values when calling a plugin.  This implies that if a user had set e.g. `XDG_CACHE_HOME=/tmp`, the var would be overwritten for the plugin call with `XDG_CACHE_HOME=/tmp/helm`, which would modify the user's full cache directory location.

I also struggled to choose the best way to represent this.  I considered not showing `XDG_CACHE_HOME` but instead showing `HELM_XDG_CACHE_HOME` with the `/helm` suffix, but then I decided it would also be confusing.

What do you think?

hickeyma: hmmm .. it probably may cause some confusion either way! :) I think that as you mentioned, it's better to output the variable values that are associated to the  particular variable, for example `XDG_CACHE_HOME`.

hickeyma (approved):
LGTM @marckhouzam, thanks. Good to merge.

marckhouzam:
Thanks @hickeyma
Because I'm semi undoing a previous PR by @adamreese I'd like to get is ok too. I'll assign this to him for review.

adamreese (changes requested):
XDG_* environment variables are OS level and not helm specific.  If this is for scripting, the environment variables should already be available.

bacongobbler:
One concern with this PR, and it aligns with @adamreese's train of thought:

```
XDG_CONFIG_HOME/tmp helm env | grep XDG_CONFIG_HOME
XDG_CONFIG_HOME=/tmp/helm
```

This doesn't seem correct.

One case that sticks out to me is the case where XDG_CONFIG_HOME or others are not set by the user. Plugins have no idea where help references these directories. I believe this becomes useful in the case where a plugin wants to know where Helm stores data when XDG_* is unset. e.g.

```
helm env | grep XDG_CONFIG_HOME
XDG_CONFIG_HOME=/home/bacongobbler/.config/helm
```

How do we handle that use case without clobbering XDG_ environment variables set by the user? Do we prefix these environment variables with `HELM_`; e.g. `HELM_XDG_CONFIG_HOME`?

marckhouzam:
> XDG_* environment variables are OS level and not helm specific.

Agreed.  And I believe that is why you removed them in #6362 (@hickeyma I had quoted the wrong PR when I answered your question, sorry)
 
> If this is for scripting, the environment variables should already be available.

The problem (as mentioned in a later comment by @bacongobbler) is that the XDG_* variable be unset.  In that case, they default to some obscure values.  This PR aims to make those default values easy to find. To illustrate:
```
$ export XDG_CONFIG_HOME=/tmp
$ h3 env | \grep XDG_CONFIG_HOME
XDG_CONFIG_HOME="/tmp"
$ unset XDG_CONFIG_HOME
$ echo $XDG_CONFIG_HOME
# empty.  This is where finding the default value is hard. But with this PR:
$ h3 env | \grep XDG_CONFIG_HOME
XDG_CONFIG_HOME="/Users/marckhouzam/Library/Preferences"
```

marckhouzam:
> One concern with this PR, and it aligns with @adamreese's train of thought:
> 
> ```
> XDG_CONFIG_HOME/tmp helm env | grep XDG_CONFIG_HOME
> XDG_CONFIG_HOME=/tmp/helm
> ```

@bacongobbler I assume you are missing an`=` sign in setting the variable in the first command?

For me (with the `=`):
```
XDG_CONFIG_HOME=/tmp helm env | \grep XDG_CONFIG_HOME
XDG_CONFIG_HOME="/tmp"
```
which is what we would want, I believe.

adamreese:
Also applications that are executed from within a plugin could depend on XDG being set or unset. 

In your script you can use `$XDG_CONFIG_HOME/helm"

If that is not an option I would prefer a name like HELM_CONFIG over HELM_XDG_CONFIG_HOME.  XDG is an OS level concept isn't important here.

marckhouzam:
> Also applications that are executed from within a plugin could depend on XDG being set or unset.

Touché.
Although I wasn't sure if that scenario was that important.

> In your script you can use `$XDG_CONFIG_HOME/helm"

That reminds me of the original reason I made this PR!
 Plugins are likely too assume they can use `$XDG_CONFIG_HOME/helm`, which the docs even refer to doing, please see https://github.com/helm/helm-www/issues/336.
But when `$XDG_CONFIG_HOME` is unset,  this is going to end up being `/helm`.

> If that is not an option I would prefer a name like HELM_CONFIG over HELM_XDG_CONFIG_HOME. XDG is an OS level concept isn't important here.

I considered that but was torn. We'd be introducing 3 new HELM_* variables and I wasn't sure we wanted to go that route

hickeyma:
> XDG_* environment variables are OS level and not helm specific. If this is for scripting, the environment variables should already be available.

I think @adamreese makes a good point here. Do we even need to display these env vars? Do we just make sure docs describe the XDG vars and that we use `helm` as path after the variables? The user can then just get the var setting from the system.

marckhouzam:
It's that normally those variables are not set. See the good explanation from @bacongobbler https://github.com/helm/helm/pull/6892#issuecomment-564713367

#6890  details the issue some more

bacongobbler:
do we have consensus to move ahead with this PR? Looks like my concern was addressed in https://github.com/helm/helm/pull/6892#issuecomment-564744115 so I'd be happy to move forward with this PR. The use case provided here (providing plugins a value to store content even if XDG_* is not set) seems like a valid use case.

Perhaps to address adam's concerns, is there a way to lookup if those variables have been set in the environment, and if so, do not append them to the list in `EnvVars`?

marckhouzam:
Thanks for reviving this.

> Perhaps to address adam and martin's concerns, is there a way to lookup if those variables have been set in the environment, and if so, do not append them to the list in `EnvVars`?

So if the variables are set in the user's environment then we would end up listing them exactly as they are set, which I believe is harmless, and actually gives a little more consistency in my opinion.

Another point we need to agree on is another remark from @adamreese :
> Also applications that are executed from within a plugin could depend on XDG being set or unset.

We must either let that point go, or we need to use different variable names to show what the default value of unset XDG_* variables are. Using different names is a little heavier but avoids force-setting those XDG variables.

marckhouzam:
And if we use different variable names (e.g., HELM_XDG_*) we would probably need to update the docs to refer to those variable names when suggesting paths and such, to avoid the whole unset XDG_ problem.

Personally, I feel that will cause some confusion with users...

marckhouzam:
After discussion and investigation (thanks to @mattfarina and @adamreese ) it was decided in the Helm dev call that there will be a larger effort to wrap the XDG* handling into configurable HELM_* variables.  This will allow the user to modify certain locations used by Helm without affecting the rest of their system.

Consequently, this PR is no longer something we should do.

PR #6904, opened 2019-11-07
print gvk information on existing resource conflict

bakito:
This PR adds additional information to resource conflict error messages.
We were facing some issues due to API version changes and had difficulties on finding the reason of the conflict.
Adding group and version helps a lot, since the id's of resources are compared with all this data.

Signed-off-by: Marc Brugger <github@bakito.ch>

### DIFF
--- a/pkg/action/validate.go
+++ b/pkg/action/validate.go
@@ -33,7 +33,8 @@ func existingResourceConflict(resources kube.ResourceList) error {
 		}
 
 		helper := resource.NewHelper(info.Client, info.Mapping)
-		if _, err := helper.Get(info.Namespace, info.Name, info.Export); err != nil {
+		existing, err := helper.Get(info.Namespace, info.Name, info.Export)
+		if err != nil {
 			if apierrors.IsNotFound(err) {
 				return nil
 			}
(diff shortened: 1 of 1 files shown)


cdrage:
@bakito Trying to replicate the issue, mind providing some steps @bakito ?

bakito:
We die an upgrade of our OpenShift cluster, where the api of some resources got converted from api version alpha to beta.
Since the Helm chart war installed with the api version alpha, we could not do an upgrade of the chart, because the version of the resource in the manifest (alpha) of the old chart did not match the version of the current resource (beta).

The error message that is changed in this PR only shows kind and name of the resource, but not the version.

I've spent hours on finding the problem, since  helm could not upgrade the chart.

This was also why i wrote this helm plugin to patch the chart to be able to do an upgrade after api version change. https://github.com/bakito/helm-patch

pkg/action/validate.go
```diff
@@ -33,15 +33,16 @@ func existingResourceConflict(resources kube.ResourceList) error {
...
+		if err != nil {
 			if apierrors.IsNotFound(err) {
 				return nil
 			}
 
 			return errors.Wrap(err, "could not get information about the resource")
 		}
 
-		return fmt.Errorf("existing resource conflict: kind: %s, namespace: %s, name: %s", info.Mapping.GroupVersionKind.Kind, info.Namespace, info.Name)
+		return fmt.Errorf("existing resource conflict: namespace: %s, name: %s, existing: [gvk: %s, ] / new: [gvk: %s]", info.Namespace, info.Name, existing.GetObjectKind().GroupVersionKind(), info.Mapping.GroupVersionKind)
```
bacongobbler: can you fix the error output? looks like some extra characters were added in here.

should be:

```
"existing resource conflict: namespace: %s, name: %s, existing_kind: %s, new_kind: %s",
```

bacongobbler (changes requested):

bacongobbler (approved):

bacongobbler:
thanks!

PR #6909, opened 2019-11-08
fix(cmd): Fixes logging on action conf init error

### LINKED ISSUE #6863: Invalid kubeconfig causes exit with no output
Output of `helm version`:
(None, that's the error. Issue observed in `v3.0.0-rc.1` and master.)

Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"12", GitVersion:"v1.12.0", GitCommit:"0ed33881dc4355495f623c6f22e7dd0b7632b7c0", GitTreeState:"clean", BuildDate:"2018-09-28T15:18:13Z", GoVersion:"go1.11", Compiler:"gc", Platform:"darwin/amd64"}
Unable to connect to the server: dial tcp 54.164.44.248:6443: i/o timeout
```

An error in the kubeconfig causes helm invocations to exit with code 1 for all commands, and no output is observed. 

Running with `HELM_DEBUG=true` emits the following:
```
helm.go:71: [debug] invalid configuration: [client-key-data or client-key must be specified for admin to use the clientCert authentication method., certificate-authority-data and certificate-authority are both specified for test-clester. certificate-authority-data will override.]
```
This line is actually helpful, but the error causes helm to exit, so it ought to be emitted at FATAL, not DEBUG. 

Separately, I propose that a broken kubeconfig should not result in a fatal error for most helm subcommands like `--help` and `version`.

jorge-gasca:
Errors relating to initializing the action config cause helm to exit silently,
except when in debug mode. This now emits the useful error.

Closes #6863

### DIFF
--- a/cmd/helm/helm.go
+++ b/cmd/helm/helm.go
@@ -68,8 +68,7 @@ func main() {
 	cmd := newRootCmd(actionConfig, os.Stdout, os.Args[1:])
 
 	if err := actionConfig.Init(settings.RESTClientGetter(), settings.Namespace(), os.Getenv("HELM_DRIVER"), debug); err != nil {
-		debug("%+v", err)
-		os.Exit(1)
+		log.Fatalf("%+v", err)
 	}
 
 	if err := cmd.Execute(); err != nil {
--- a/cmd/helm/helm.go
+++ b/cmd/helm/helm.go
@@ -68,7 +68,7 @@ func main() {
 	cmd := newRootCmd(actionConfig, os.Stdout, os.Args[1:])
 
 	if err := actionConfig.Init(settings.RESTClientGetter(), settings.Namespace(), os.Getenv("HELM_DRIVER"), debug); err != nil {
-		log.Fatalf("%+v", err)
+		log.Fatal(err)
 	}
 
 	if err := cmd.Execute(); err != nil {


hickeyma (approved):
LGTM, thanks @jorge-gasca. I am going to hold back on merging as want to get other maintainers opinion on merging as we are in the rc phase.

hickeyma (changes requested):
On revisiting this, the flow has changed as now checking for `IsReachable()` (https://github.com/helm/helm/blob/master/pkg/kube/client.go#L76). 

This outputs with error as follows if config is not valid:
```console
$ helm ls
```

It however behaves as follows with this fix:
```console
$ helm ls
```

@jorge-gasca Do you mind taking a look at the PR again with the current code base?

jorge-gasca:
@hickeyma I've rebased locally and this fix seems to be working ok for me. Here is my manual test:

1. I make an invalid KubeConfig(by copying my regular KubeConfig to a temporary file and make the "certificate authority" for the cluster point to a file that doesn't exist).
2. `KUBECONFIG=<TEMP_FILE_PATH> ./bin/helm ls`

Doing that, I get back the detailed error with the invalid KubeConfig. If I use my valid KubeConfig with a cluster that is not reachable, I correctly  get `Error: Kubernetes cluster unreachable`.

hickeyma:
@jorge-gasca From comments I made in https://github.com/helm/helm/pull/6909#pullrequestreview-330051817, Do we still needs the changes in this PR or is it now being handled correctly else where in the code?

jorge-gasca:
@hickeyma This is still needed. Going through my same test case but without my commit, when the incorrect KubeConfig is used, helm silently exits instead of giving the useful KubeConfig validation error. 

The `IsReachable()` logic may address pointing to a non-existent cluster, but this PR shows the validation errors in the KubeConfig
(e.g. `invalid configuration: [client-key-data or client-key must be specified for admin to use the clientCert authentication method., certificate-authority-data and certificate-authority are both specified for test-clester. certificate-authority-data will override.]` from #6863).

hickeyma:
@jorge-gasca I agree with you assessment. There can still be flow which will reach the exit code. This is shown in #7271

cmd/helm/helm.go
```diff
@@ -68,8 +68,7 @@ func main() {
 	cmd := newRootCmd(actionConfig, os.Stdout, os.Args[1:])
 
 	if err := actionConfig.Init(settings.RESTClientGetter(), settings.Namespace(), os.Getenv("HELM_DRIVER"), debug); err != nil {
-		debug("%+v", err)
-		os.Exit(1)
+		log.Fatalf("%+v", err)
```
hickeyma: Why use `%+v`? Should this be `%s`?

hickeyma (commented):
@jorge-gasca Thanks for the changes. Got a small question around format of message.

cmd/helm/helm.go
jorge-gasca: I kept that from the previous way this error was being handled. The `%+v` is supposed to print out the error struct nicely, but because the error struct just contains a string, it makes no difference using `%+v` vs `%s`. In fact, we could avoid using `log.Fatalf` all together and use `log.Fatal(err)`. Let me know and I can change it to whichever seems best.

bacongobbler: `log.Fatal(err)` is the more idiomatic approach, so let's go ahead with that.

hickeyma: Yep, I agree with `log.Fatal(err)` as well.

hickeyma (approved):
LGTM, thanks for improving this @jorge-gasca

PR #6915, opened 2019-11-08
feat(helm): Add map of cmds that req actionConfig

jorge-gasca:
Most commands unrelated to release do not require actionConfig. This
provides a way to differentiate between commands that use it and those
that do not, so that a user doesn't need to worry about their kubeconfig
if they are running `helm version` for example.

This feature was suggested in #6863 .

Signed-off-by: Jorge Gasca <jorge.ignacio.gasca@gmail.com>

### DIFF
--- a/cmd/helm/helm.go
+++ b/cmd/helm/helm.go
@@ -67,9 +67,15 @@ func main() {
 	actionConfig := new(action.Configuration)
 	cmd := newRootCmd(actionConfig, os.Stdout, os.Args[1:])
 
-	if err := actionConfig.Init(settings.RESTClientGetter(), settings.Namespace(), os.Getenv("HELM_DRIVER"), debug); err != nil {
-		debug("%+v", err)
+	subCmd, _, err := cmd.Find(os.Args[1:])
+	if err != nil {
 		os.Exit(1)
+	}
+	actionConfigRequired := action.ActionConfigRequired(subCmd.Use)
+	if actionConfigRequired {
+		if err := actionConfig.Init(settings.RESTClientGetter(), settings.Namespace(), os.Getenv("HELM_DRIVER"), debug); err != nil {
+			log.Fatalf("%+v", err)
+		}
 	}
 
 	if err := cmd.Execute(); err != nil {
--- a/cmd/helm/helm.go
+++ b/cmd/helm/helm.go
@@ -71,7 +71,7 @@ func main() {
 	if err != nil {
 		os.Exit(1)
 	}
-	actionConfigRequired := action.ActionConfigRequired(subCmd.Use)
+	actionConfigRequired := action.ConfigRequired(subCmd.Use)
 	if actionConfigRequired {
 		if err := actionConfig.Init(settings.RESTClientGetter(), settings.Namespace(), os.Getenv("HELM_DRIVER"), debug); err != nil {
 			log.Fatalf("%+v", err)
--- a/cmd/helm/helm.go
+++ b/cmd/helm/helm.go
@@ -71,7 +71,7 @@ func main() {
 	if err != nil {
 		os.Exit(1)
 	}
-	actionConfigRequired := action.ConfigRequired(subCmd.Use)
+	actionConfigRequired := action.ConfigRequired(subCmd.Name())
 	if actionConfigRequired {
 		if err := actionConfig.Init(settings.RESTClientGetter(), settings.Namespace(), os.Getenv("HELM_DRIVER"), debug); err != nil {
 			log.Fatalf("%+v", err)
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -245,7 +245,7 @@ func (c *Configuration) Init(getter genericclioptions.RESTClientGetter, namespac
 	return nil
 }
 
-func ActionConfigRequired(command string) bool {
+func ConfigRequired(command string) bool {
 	var actionConfigRequired = map[string]bool{
 		// release commands
 		"get":         true,
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -244,3 +244,44 @@ func (c *Configuration) Init(getter genericclioptions.RESTClientGetter, namespac
 
 	return nil
 }
+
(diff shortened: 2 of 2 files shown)


bacongobbler:
```
pkg/action/action.go:248:6: func name will be used as action.ActionConfigRequired by other packages, and that stutters; consider calling this ConfigRequired (golint)
func ActionConfigRequired(command string) bool {
     ^
make: *** [Makefile:85: test-style] Error 1
Exited with code 2
```

pkg/action/action.go
```diff
@@ -244,3 +244,44 @@ func (c *Configuration) Init(getter genericclioptions.RESTClientGetter, namespac
...
+		"env":        false,
+		"plugin":     false,
+		"version":    false,
+		"docs":       false,
+
+		//root
+		"helm": false,
+	}
+
+	return actionConfigRequired[command]
```
marckhouzam: What happens in Go if the command parameter does not exist in the map?

This will be important as it seems highly likely to me that it will be forgotten sometimes to add a new command here.
A compiler failure would be best, which is not what will happen here.

What happens for plugins?

jorge-gasca: Ah, true. I tested nonexistent subcommands, but not a subcommand that exists but hasn't been added to the map(or a plugin).

I will update this to check if the key exists in the map, and if it does not, return `true` by default. That way new commands and plugins will just have the current functionality of failing on invalid kubeconfigs until they are added to the map.

Initially I wanted to check if a command needed the config by some property in the `cobra.Command` struct, but that doesn't seem easy to do since the parameter indicating that it needs the action config is in the function that creates it(e.g. `newListCmd()`). Having something like that would expose this functionality to plugins as well, and remove the need for maintaining a map like this.

marckhouzam:
I tried the PR a bit and hit this:
```
$ h3 history anything
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x2183c11]

goroutine 1 [running]:
helm.sh/helm/v3/pkg/action.(*History).Run(0xc00000e980, 0x7ffeefbff5f8, 0x8, 0x16, 0x16, 0x16, 0x15, 0xc000178000)
	/Users/marckhouzam/go/src/helm.sh/helm/pkg/action/history.go:44 +0x31
main.getHistory(0xc00000e980, 0x7ffeefbff5f8, 0x8, 0x4, 0x257cb5a, 0x4, 0x11d4160, 0xc00068dcf8)
	/Users/marckhouzam/go/src/helm.sh/helm/cmd/helm/history.go:108 +0x49
main.newHistoryCmd.func1(0xc000530c80, 0xc0002e9c50, 0x1, 0x1, 0x0, 0x0)
	/Users/marckhouzam/go/src/helm.sh/helm/cmd/helm/history.go:63 +0x6d
github.com/spf13/cobra.(*Command).execute(0xc000530c80, 0xc0002e9c30, 0x1, 0x1, 0xc000530c80, 0xc0002e9c30)
	/Users/marckhouzam/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:826 +0x465
github.com/spf13/cobra.(*Command).ExecuteC(0xc000326f00, 0x14, 0x0, 0x2)
	/Users/marckhouzam/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:914 +0x2fc
	/Users/marckhouzam/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:864
main.main()
	/Users/marckhouzam/go/src/helm.sh/helm/cmd/helm/helm.go:81 +0x150
```

cmd/helm/helm.go
```diff
@@ -67,10 +67,16 @@ func main() {
 	actionConfig := new(action.Configuration)
 	cmd := newRootCmd(actionConfig, os.Stdout, os.Args[1:])
 
-	if err := actionConfig.Init(settings.RESTClientGetter(), settings.Namespace(), os.Getenv("HELM_DRIVER"), debug); err != nil {
-		debug("%+v", err)
+	subCmd, _, err := cmd.Find(os.Args[1:])
+	if err != nil {
 		os.Exit(1)
 	}
+	actionConfigRequired := action.ConfigRequired(subCmd.Use)
```
marckhouzam: I think you should use `subCmd.Name()` instead of `subCmd.Use`.  I believe it will also fix the crash.

marckhouzam:
> Initially I wanted to check if a command needed the config by some property in the `cobra.Command` struct, but that doesn't seem easy to do since the parameter indicating that it needs the action config is in the function that creates it(e.g. `newListCmd()`). Having something like that would expose this functionality to plugins as well, and remove the need for maintaining a map like this.

I've been thinking about this.  I think that if we want to keep the call to `Configuration.Init(...)` centralized, you have found an imaginative way to do just that.  I personally don't have any better suggestion.  If we are willing to de-centralize the call to `Configuration.Init(...)` I think it could be pushed to the `RunE` methods of each of the relevant cobra commands.  Although the `cfg` parameter is referenced before `RunE` is called, it does not actually need to have been initialized until `RunE` is called.

This same approach was done to check if the cluster is reachable.  For example https://github.com/helm/helm/blob/750b870aedd35b69b9ca1e1517635fa70367a309/pkg/action/install.go#L156
In that case, the code was put in the action package to make it available to the SDK.  In your case, I believe it must be kept in the `main` package.  So, for example, it could be added at https://github.com/helm/helm/blob/750b870aedd35b69b9ca1e1517635fa70367a309/cmd/helm/install.go#L148 for the install command.

Which of the two approaches is better could be argued.  I just wanted to mention this other option, which I like because it keeps the `Init()` code close to the command that needs it.

What do you think?

I haven't actually tried this idea, so I may have missed some limitation with it.

marckhouzam:
Thanks for the effort @jorge-gasca 

The issue this PR is trying to address is being solved as a side-effect of #7831 

A lazy-loading client will be used to delay the use of the kube client until actually necessary.  So any command that does not require the kube cluster will no longer be victim to a bad kubeconfig.

Consequently, I will close this PR

PR #6920, opened 2019-11-09
v3: Make helm upgrade check if the cluster is reachable

marckhouzam:
The 'helm upgrade' command was not checking if the cluster was reachable.

Before the PR:
```
$ h3 upgrade nginx stable/nginx-ingress
Error: UPGRADE FAILED: query: failed to query with labels: Get https://blabla:6443/api/v1/namespaces/default/secrets?labelSelector=name%3Dnginx%2Cowner%3Dhelm%2Cstatus%3Ddeployed: Get https://moreblabla: dial tcp 10.17.10.20:443: i/o timeout
$ h3 upgrade --install nginx stable/nginx-ingress
Error: UPGRADE FAILED: query: failed to query with labels: Get https://blabla:6443/api/v1/namespaces/default/secrets?labelSelector=name%3Dnginx%2Cowner%3Dhelm%2Cstatus%3Ddeployed: Get https://moreblabla: dial tcp 10.17.10.20:443: i/o timeout
```
After the PR:
```
$ h3 upgrade nginx stable/nginx-ingress
Error: UPGRADE FAILED: Kubernetes cluster unreachable
$ h3 upgrade --install nginx stable/nginx-ingress
```
Also, `helm upgrade --install` first checks if the release exists already.  If that check fails because the cluster is not reachable, there is no point in continuing the upgrade.  This optimization avoids
a second timeout of 30 seconds when trying to do the upgrade (one to check the release existence, the other to try the upgrade).

**Notes to maintainers**

~~This change creates a new API: `kube.ErrClusterUnreachable`.
Also, it imports `helm.sh/helm/v3/pkg/kube` from `cmd/helm/upgrade.go`.
This may not be desirable, I don't know.  Maybe the public error should be put in another package?~~

### DIFF
--- a/cmd/helm/upgrade.go
+++ b/cmd/helm/upgrade.go
@@ -91,8 +91,7 @@ func newUpgradeCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 			}
 
 			if client.Install {
-				// If a release does not exist, install it. If another error occurs during
-				// the check, ignore the error and continue with the upgrade.
+				// If a release does not exist, install it.
 				histClient := action.NewHistory(cfg)
 				histClient.Max = 1
 				if _, err := histClient.Run(args[0]); err == driver.ErrReleaseNotFound {
@@ -119,6 +118,8 @@ func newUpgradeCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 						return err
 					}
 					return outfmt.Write(out, &statusPrinter{rel, settings.Debug})
+				} else if err != nil {
+					return err
 				}
 			}
 
--- a/pkg/action/upgrade.go
+++ b/pkg/action/upgrade.go
@@ -75,6 +75,10 @@ func NewUpgrade(cfg *Configuration) *Upgrade {
 
 // Run executes the upgrade on the given release.
 func (u *Upgrade) Run(name string, chart *chart.Chart, vals map[string]interface{}) (*release.Release, error) {
+	if err := u.cfg.KubeClient.IsReachable(); err != nil {
+		return nil, err
+	}
+
 	// Make sure if Atomic is set, that wait is set as well. This makes it so
 	// the user doesn't have to specify both
 	u.Wait = u.Wait || u.Atomic


marckhouzam:
This is a small bugfix.  Anyone have time to review it?

cmd/helm/upgrade.go
```diff
@@ -91,10 +92,16 @@ func newUpgradeCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
...
 			if client.Install {
 				// If a release does not exist, install it. If another error occurs during
-				// the check, ignore the error and continue with the upgrade.
+				// the check, ignore the error (unless it indicates the cluster is not reachable)
+				// and continue with the upgrade.
 				histClient := action.NewHistory(cfg)
 				histClient.Max = 1
-				if _, err := histClient.Run(args[0]); err == driver.ErrReleaseNotFound {
+
+				_, err := histClient.Run(args[0])
```
bacongobbler: Is it possible to capture other non-nil errors as well? e.g.

```
if _, err := histClient.Run(args[0]); err == driver.ErrReleaseNotFound {
    ...
} else if err != nil {
    return err
}
```

That way we don't have to refactor this in the future if we introduce further non-nil errors

pkg/action/upgrade.go
```diff
@@ -70,6 +70,10 @@ func NewUpgrade(cfg *Configuration) *Upgrade {
 
 // Run executes the upgrade on the given release.
 func (u *Upgrade) Run(name string, chart *chart.Chart, vals map[string]interface{}) (*release.Release, error) {
+	if err := u.cfg.KubeClient.IsReachable(); err != nil {
```
bacongobbler: we should check if `u.DryRun` is. `helm upgrade --dry-run` does not rely on a cluster being connected IIRC

bacongobbler (changes requested):

pkg/action/upgrade.go
bacongobbler: https://github.com/helm/helm/blob/0afce196fda3bad3e7c30c4da53d8b18335524da/pkg/action/install.go#L157-L162

bacongobbler: scratch that... `--dry-run` != `--client-only`. It does rely on the cluster being reachable. Disregard

cmd/helm/upgrade.go
marckhouzam: Good point!  Done.

cmd/helm/upgrade.go
```diff
@@ -91,9 +91,11 @@ func newUpgradeCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 
 			if client.Install {
 				// If a release does not exist, install it. If another error occurs during
-				// the check, ignore the error and continue with the upgrade.
```
marckhouzam: @bacongobbler I just saw this comment again.  It explicitly says to ignore other errors.  It was written by @technosophos back in 2016 in 36e6094.  Maybe it was because of some tiller need?  Do you feel it is safe to stop ignoring other errors?

pkg/kube/client.go
```diff
@@ -47,8 +47,13 @@ import (
...
 )
 
-// ErrNoObjectsVisited indicates that during a visit operation, no matching objects were found.
-var ErrNoObjectsVisited = errors.New("no objects visited")
+var (
+	// ErrNoObjectsVisited indicates that during a visit operation, no matching objects were found.
+	ErrNoObjectsVisited = errors.New("no objects visited")
+
+	// ErrClusterUnreachable indicates that the Kubernetes cluster cannot be reached
+	ErrClusterUnreachable = errors.New("Kubernetes cluster unreachable")
```
marckhouzam: If we go ahead and abort the upgrade on all errors (except the releaseNotFound error), then technically I don't need the  modifications in this file.  But I find it is nice to have, in case it is useful in the future.  But it does introduce a new API.  

Keep or remove the new `ErrClusterUnreachable` constant?

cmd/helm/upgrade.go
technosophos: This code was originally executed in Tiller. Back then, the history client returned only a small set of errors, and the only one we cared about was ErrReleaseNotFound. If that happened, we knew we _could_ create a new install. Any other error case, we didn't care about -- we would just press on with the upgrade and see if we could make more progress. At the time, any other errors coming out of histClient.Run did not seem to warrant aborting the entire upgrade effort.

Your change makes a lot of sense now that Tiller is gone and connecting to the API server is more volatile. It is possible that we may catch a spurious error that wouldn't actually stop a successful upgrade, but I'd have to read the current Run code to know if that's the case.

marckhouzam: Great. Thanks for the explanation. So we'll go with @bacongobbler 's suggestion of aborting on all errors.

marckhouzam: I'll push a new version removing the comment.

pkg/kube/client.go
marckhouzam: @bacongobbler do you want to keep this part?  It is not technically needed.

marckhouzam: This error message is no longer a constant on the latest master.  I'll simply remove the fact that I create a variable for it.

marckhouzam:
This little fix is down to 5 lines.  Ok to merge?

marckhouzam:
I had to squash and force-push because there was a conflict in one of the middle commits and git doesn't care that a further commit resolved it.

hickeyma (approved):
LGTM, thanks @marckhouzam

marckhouzam:
Thanks @hickeyma.
I've made the changes requested by @bacongobbler a while back, so I will proceed with the merge.

Issue #6378, opened 2019-09-06
Helm 3 - upgrade nginx - spec.clusterIP: Invalid value: "": field is immutable


vplauzon:
I use the following to install / upgrade a chart:

./helm upgrade --install \
  --set rbac.create=false \
  --set controller.replicaCount=2 \
  --set controller.service.loadBalancerIP=$ip \
  --wait main-ingress stable/nginx-ingress

(Where $ip is an IP, e.g. 10.0.0.1)

That's done in a CI/CD pipeline, so the idea is to install the first time, upgrade next times.

It installs fine.  At the second run, it outputs the following:

_client.go:339: Cannot patch Service: "main-ingress-nginx-ingress-controller" (Service "main-ingress-nginx-ingress-controller" is invalid: spec.clusterIP: Invalid value: "": field is immutable)
client.go:358: Use --force to force recreation of the resource
client.go:339: Cannot patch Service: "main-ingress-nginx-ingress-default-backend" (Service "main-ingress-nginx-ingress-default-backend" is invalid: spec.clusterIP: Invalid value: "": field is immutable)
client.go:358: Use --force to force recreation of the resource
Error: UPGRADE FAILED: Service "main-ingress-nginx-ingress-controller" is invalid: spec.clusterIP: Invalid value: "": field is immutable && Service "main-ingress-nginx-ingress-default-backend" is invalid: spec.clusterIP: Invalid value: "": field is immutable_

I also get this on helm list:

NAME            NAMESPACE       REVISION        UPDATED                                 STATUS          CHART           
main-ingress    default         1               2019-09-06 13:17:33.8463781 -0400 EDT   deployed        nginx-ingress-1.18.0
main-ingress    default         2               2019-09-06 13:21:11.6428945 -0400 EDT   failed          nginx-ingress-1.18.0

So, the release has failed.

I didn't have that problem with Helm 2.  Is it due to a change of behaviour in helm 3 or is it a bug?  If it's the former, how could I change the command not to have that problem?

Output of `helm version`:  version.BuildInfo{Version:"v3.0.0-beta.2", GitCommit:"26c7338408f8db593f93cd7c963ad56f67f662d4", GitTreeState:"clean", GoVersion:"go1.12.9"}

[...]

PR #6933, opened 2019-11-12
fix for #6378

johannges:
only run replace after patch failed

This fixes the Issue #6378, that `helm upgrade --force` doesn’t work anymore


bacongobbler:
Looks like some extra logic from another PR was lodged in here. Can you please review your changes and make sure only the minimal set of changes are in this PR. Thanks.

PR #5151, opened 2019-01-10
Fix: type conversion for zero values

Flask:
**What this PR does / why we need it**:
The current implementation does not allow allow to set a value to `0`. E.g. 
```
helm tiller install -n test stable/some-chart --set securityContext.runAsUser=0
```
fails with the following message:
```
Error: release test failed: Deployment in version "v1beta1" cannot be handled as a Deployment: v1beta1.Deployment.Spec: v1beta1.DeploymentSpec.Template: v1.PodTemplateSpec.Spec: v1.PodSpec.Containers: []v1.Container: v1.Container.SecurityContext: v1.SecurityContext.RunAsUser: readUint64: unexpected character: �, error found in #10 byte of ...|nAsUser":"0"},"volum|..., bigger context ...|text":{"runAsUser":"0"},"volumeMounts":[{"mountPath":"/etc/kubernetes/|...
```
since helm tries to convert 0 to a string. 

**Special notes for your reviewer**:

**If applicable**:
- [ ] this PR contains documentation
- [x] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

PR #4682, opened 2018-09-21
fix(helm): --set for out of order list values

danjwinter:
When a user specifies value overrides for list values out of order,
strvals.listItem panics. Change strvals.listItem to handle this case by
re-initializing nil values to a new map.

Closes #4503

Co-authored-by: Cameron Childress <cameron@cchildress.org>
Co-authored-by: Kevin Collette <hal.collette@gmail.com>
Co-authored-by: Connor McKelvey <connor.mckelvey@gmail.com>
Co-authored-by: Dan Winter <dan.j.winter@gmail.com>
Signed-off-by: Dan Winter <dan.j.winter@gmail.com>

PR #4142, opened 2018-05-29
Make --set-string flag parse booleans and null as strings

ScriptingSquirrel:
This makes the `--set-string` flag behave more intuitively. With this change, the values `true`, `false` and `null` are no longer interpreted as booleans but as strings if `--set-string` is used.

Closes #2848 
Relates to #4006

PR #3912, opened 2018-04-17
 fix(pkg/strvals): evaluate "null" values 

michelleN:
resolves #3891

PR #6941, opened 2019-11-12
fix(strvals): port #3912, #4142, #4682, and #5151 to Helm 3

### LINKED ISSUE #6940: helm 3.0.0-rc.3 --set-string does not work with booleans
It seems that --set-string does not work with boolean as expected under helm 3.0.0-rc.3

According to https://helm.sh/docs/using_helm/#the-format-and-limitations-of-set, 
> --set-string foo=true results in a string value of "true".

Running the following, I would expect it to quote the "true", and deploy the nginx-ingress on the nodes matching the selector.
```
helm upgrade --install nginx-ingress stable/nginx-ingress --namespace nginx-ingress --set-string controller.nodeSelector."boolean\.dotted/slashed-dashed"=true
```
However, I get the following error:
```
Release "nginx-ingress" does not exist. Installing it now.
Error: Deployment in version "v1" cannot be handled as a Deployment: v1.Deployment.Spec: v1.DeploymentSpec.Template: v1.PodTemplateSpec.Spec: v1.PodSpec.NodeSelector: ReadString: expects " or n, but found t, error found in #10 byte of ...|-dashed":true},"serv|..., bigger context ...|,"nodeSelector":{"boolean.dotted/slashed-dashed":true},"serviceAccountName":"nginx-ingress-nginx-|...
```

It seems to be a regression of #2848, the code merged in #4142 has disappeared 

Output of `helm version`:
```helm version
version.BuildInfo{Version:"v3.0.0-rc.3", GitCommit:"2ed206799b451830c68bff30af2a52879b8b937a", GitTreeState:"clean", GoVersion:"go1.13.4"}
```
Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-15T19:18:23Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-15T19:09:08Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}
```
Bare metal nodes deployed with RKE

bacongobbler:
see #3912, #4142. #4682, and #5151
closes #6940

### DIFF
--- a/pkg/strvals/parser.go
+++ b/pkg/strvals/parser.go
@@ -84,7 +84,7 @@ func ParseFile(s string, reader RunesValueReader) (map[string]interface{}, error
 	return vals, err
 }
 
-// ParseIntoString parses a strvals line nad merges the result into dest.
+// ParseIntoString parses a strvals line and merges the result into dest.
 //
 // This method always returns a string as the value.
 func ParseIntoString(s string, dest map[string]interface{}) error {
@@ -108,6 +108,9 @@ type RunesValueReader func([]rune) (interface{}, error)
 
 // parser is a simple parser that takes a strvals line and parses it into a
 // map representation.
+//
+// where sc is the source of the original data being parsed
+// where data is the final parsed data from the parses with correct types
 type parser struct {
 	sc     *bytes.Buffer
 	data   map[string]interface{}
@@ -285,7 +288,13 @@ func (t *parser) listItem(list []interface{}, i int) ([]interface{}, error) {
 		// We have a nested object. Send to t.key
 		inner := map[string]interface{}{}
 		if len(list) > i {
-			inner = list[i].(map[string]interface{})
+			var ok bool
+			inner, ok = list[i].(map[string]interface{})
+			if !ok {
+				// We have indices out of order. Initialize empty value.
+				list[i] = map[string]interface{}{}
+				inner = list[i].(map[string]interface{})
+			}
 		}
 
 		// Recurse
@@ -367,6 +376,11 @@ func inMap(k rune, m map[rune]bool) bool {
 
 func typedVal(v []rune, st bool) interface{} {
 	val := string(v)
+
+	if st {
+		return val
+	}
+
 	if strings.EqualFold(val, "true") {
 		return true
 	}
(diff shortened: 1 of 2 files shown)


bacongobbler:
Looks like golangci-lint is failing to build:

> go/pkg/mod/github.com/golangci/golangci-lint@v1.21.0/pkg/golinters/wsl.go:43:6: unknown field 'AllowCaseTrailingWhitespace' in struct literal of type wsl.Configuration

Will look into this further.

bacongobbler:
https://github.com/helm/helm/pull/6942 should fix the CI issue, but this should be reviewable on its own.

pkg/strvals/parser.go
```diff
@@ -375,8 +380,12 @@ func typedVal(v []rune, st bool) interface{} {
 		return false
 	}
 
-	// If this value does not start with zero, and not returnString, try parsing it to an int
-	if !st && len(val) != 0 && val[0] != '0' {
+	if strings.EqualFold(val, "null") {
+		return nil
+	}
+
```
mattfarina: In the current v2 there is also...

https://github.com/helm/helm/blob/0ca95341c19fcaf9cecdf50caef1693568835a84/pkg/strvals/parser.go#L396-L398

It's between these spaces. Should that be here, too.

bacongobbler: yes, good catch. That was from #5151. I picked up the null fix from #3912.

bacongobbler:
Okay, I ran through parser.go and parser_test.go to see if there were any other missing changes. I managed to find a few fixes that were not ported over from Helm 2. That should be fixed now.

pkg/strvals/parser.go
bacongobbler: fix applied.

mattfarina (approved):

PR #6942, opened 2019-11-12
fix(ci): pin golangci-lint to v1.21.0

bacongobbler:
should fix the issue in master with building golangci-lint during a CI run. See #6941.

Will need to cherry-pick this into the 3.0 branch to ensure CI runs will succeed.

https://github.com/golangci/golangci-lint#go-get

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,6 @@ GOPATH        = $(shell go env GOPATH)
 DEP           = $(GOPATH)/bin/dep
 GOX           = $(GOPATH)/bin/gox
 GOIMPORTS     = $(GOPATH)/bin/goimports
-GOLANGCI_LINT = $(GOPATH)/bin/golangci-lint
 
 ACCEPTANCE_DIR:=$(GOPATH)/src/helm.sh/acceptance-testing
 # To specify the subset of acceptance tests to run. '.' means all tests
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -118,9 +116,6 @@ format: $(GOIMPORTS)
 $(GOX):
 	(cd /; GO111MODULE=on go get -u github.com/mitchellh/gox)
 
-$(GOLANGCI_LINT):
-	(cd /; GO111MODULE=on go get -u github.com/golangci/golangci-lint/cmd/golangci-lint)
```
mattfarina: If you want to specify the version you can do the following...

```
(cd /; GO111MODULE=on go get -u github.com/golangci/golangci-lint/cmd/golangci-lint@v1.21.0)
```

This will fetch it at a version and install it.

mattfarina (approved):
See the note just as an FYI. This follows the docs on installing it so lgtm

Makefile
bacongobbler: That would be a less invasive approach. I like that. Let me try that instead.

bacongobbler: nope... resulted in the same issue. Back to the original proposal for now.

mattfarina:
still lgtm to merge.

PR #6949, opened 2019-11-13
fix(get): install Helm v2.16.1

bacongobbler:
This is a temporary fix. This prevents the get script from installing Helm 3 as soon as it's released, potentially causing disruption to users that were expecting a Helm 2 release.

A `get-helm-3` script is also supplied, which is identical to the previous `get` script. That way, Helm 3 users also have an equivalent way to install Helm 3.

Will update the Helm 3 documentation as a follow-up. This does not affect users for Helm 2.

### DIFF
--- a/scripts/get-helm-3
+++ b/scripts/get-helm-3
@@ -187,7 +187,7 @@ testVersion() {
 help () {
   echo "Accepted cli arguments are:"
   echo -e "\t[--help|-h ] ->> prints this help"
-  echo -e "\t[--version|-v <desired_version>]"
+  echo -e "\t[--version|-v <desired_version>] . When not defined it defaults to latest"
   echo -e "\te.g. --version v2.4.0  or -v latest"
   echo -e "\t[--no-sudo]  ->> install without sudo"
 }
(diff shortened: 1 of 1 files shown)


technosophos (approved):
LGTM

scripts/get
```diff
@@ -213,7 +213,9 @@ while [[ $# -gt 0 ]]; do
     '--version'|-v)
        shift
        if [[ $# -ne 0 ]]; then
-           export DESIRED_VERSION="${1}"
+          # FIXME(bacongobbler): hard code the desired version for the time being.
+          # A better fix would be to filter for Helm 2 release pages.
+           export DESIRED_VERSION="${1:="v2.16.1"}"
```
marckhouzam: This looks suspicious.
The `${1:="v2.16.1"}` form will set the `v2.16.1` string only if `$1` is unset, however, we are in an if statement that checked that `$1` was not null.
Maybe forcing the version should instead be done in `checkDesiredVersion()`?

When testing this code change, things might still have worked since 2.16.1 is actually the latest version :-)

bacongobbler: Thank you for catching this. I'll see what I can do.

scripts/get-helm-3
```diff
@@ -0,0 +1,245 @@
...
+  else
+    TAG=$DESIRED_VERSION
+  fi
+}
+
+# checkHelmInstalledVersion checks which version of helm is installed and
+# if it needs to be changed.
+checkHelmInstalledVersion() {
+  if [[ -f "${HELM_INSTALL_DIR}/${PROJECT_NAME}" ]]; then
+    local version=$("${HELM_INSTALL_DIR}/${PROJECT_NAME}" version -c | grep '^Client' | cut -d'"' -f2)
```
marckhouzam: If this script is used and v3 is already installed, the `grep '^Client'` above will fail since `helm3 version -c` does not print the `Client` prefix anymore.  Removing the `grep` and just keeping `cut` would be sufficient I believe.

bacongobbler: addressed in #6953. Thanks!

PR #6953, opened 2019-11-13
fix(get): hard code DESIRED_VERSION when unset

bacongobbler:
thanks to @marckhouzam for identifying the issue with the earlier approach in #6949. I commented out the previous code; Once Helm 3 has been released I'll un-comment this section and work on the proper fix as recommended in the comments.

You can manually test this by switching the TAG value to something other than 2.16.1 (like, say, 2.16.0) and the get script will fetch that version by default.

### DIFF
--- a/scripts/get-helm-3
+++ b/scripts/get-helm-3
@@ -18,7 +18,6 @@
 # the package manager for Go: https://github.com/Masterminds/glide.sh/blob/master/get
 
 PROJECT_NAME="helm"
-TILLER_NAME="tiller"
 
 : ${USE_SUDO:="true"}
 : ${HELM_INSTALL_DIR:="/usr/local/bin"}
@@ -94,7 +93,7 @@ checkDesiredVersion() {
 # if it needs to be changed.
 checkHelmInstalledVersion() {
   if [[ -f "${HELM_INSTALL_DIR}/${PROJECT_NAME}" ]]; then
-    local version=$("${HELM_INSTALL_DIR}/${PROJECT_NAME}" version -c | grep '^Client' | cut -d'"' -f2)
+    local version=$("${HELM_INSTALL_DIR}/${PROJECT_NAME}" version --template="{{ .Version }}")
     if [[ "$version" == "$TAG" ]]; then
       echo "Helm ${version} is already ${DESIRED_VERSION:-latest}"
       return 0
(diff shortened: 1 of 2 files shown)


marckhouzam:
For what it's worth: LGTM

bacongobbler:
I really appreciate you catching these issues. Thanks Marc.

bacongobbler:
The `get-helm-3` fixes can be tested with `./scripts/get-helm-3 --version canary`.

thomastaylor312 (approved):

scripts/get-helm-3
```diff
@@ -94,7 +93,7 @@ checkDesiredVersion() {
 # if it needs to be changed.
 checkHelmInstalledVersion() {
   if [[ -f "${HELM_INSTALL_DIR}/${PROJECT_NAME}" ]]; then
-    local version=$("${HELM_INSTALL_DIR}/${PROJECT_NAME}" version -c | grep '^Client' | cut -d'"' -f2)
+    local version=$("${HELM_INSTALL_DIR}/${PROJECT_NAME}" version --template="{{ .Version }}")
```
marckhouzam: So now I'm probably pushing it...
The first install of v3 will normally be with a current installation of v2.  Two things:
1. `--template="{{ .Version }}"` returns `<no value>` for v2.  Not a real problem since it will still go ahead with the install of v3 (there will be a printout to the user saying `Helm canary is available. Changing from version <no value>.`)
1. if the cluster is not reachable, `helm2 version` will take a long time.  It would be nice to have the `-c` flag like before.

bacongobbler: Good call. Whatever fix we apply to `get`'s version selection filter should probably be applied here as well. That way this script cannot be used with Helm 2, and vice versa.

bacongobbler: As for

> The first install of v3 will normally be with a current installation of v2

This will be fixed once Helm 3 has been released, and given this script is called `get-helm-3`, hopefully it's clear to users that this should only be used for fetching Helm 3 release binaries. :)

marckhouzam: > This will be fixed once Helm 3 has been released, and given this script is called `get-helm-3`, hopefully it's clear to users that this should only be used for fetching Helm 3 release binaries. :)

Right. So we can assume the user will be downloading v3. But the first time they do so there will probably be a v2 currently installed on their machine (that was my case when I tried it). IIUC `checkHelmInstalledVersion()` checks the *currently* installed version, so this call will run with v2.

bacongobbler: Ah! I see where you're getting at now. Thank you for clarifying.

marckhouzam: That being said, things do work as needed as they stand.

PR #6966, opened 2019-11-14
Update get script after new main release Helm v3

micheleIT:
read latest v2 stable from release page html (no -rc versions)
grep for tag - href path and filter out the plain version number


jdolitsky:
Going to close, I believe this was resolved in https://github.com/helm/helm/pull/7942

PR #6979, opened 2019-11-15
Add comments about release Version variable

lindhe:
I was looking into the `get` command, and got tripped up by the
`Version` variable. It was unclear to me what Version represents, since
it's called REVISION when doing e.g., `helm list`.

But even after knowing this, it was not very clear to me why we
(implicitly) set the Version variable to 0 but never seem to use it.
`mhickey` explained to me on Slack that this gets the latest revision of
the release. Makes sense, but I added a comment about that too, to
clarify.

### DIFF
--- a/pkg/action/get.go
+++ b/pkg/action/get.go
@@ -26,6 +26,7 @@ import (
 type Get struct {
 	cfg *Configuration
 
+	// Initializing Version to 0 will get the latest revision of the release.
 	Version int
 }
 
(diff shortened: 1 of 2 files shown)


bacongobbler:
would you mind signing off on the DCO? More details here: https://github.com/helm/helm/pull/6979/checks?check_run_id=304632347

jdolitsky (approved):

liuming-dev (approved):

bacongobbler (approved):

PR #6987, opened 2019-11-16
ref(client): add overwrite option

### LINKED ISSUE #6980: Using Openshift Imagestream with helm 3 results in unnecessary redeployments
Output of `helm version`:
```
```
Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-15T23:42:50Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"11+", GitVersion:"v1.11.0+d4cacc0", GitCommit:"d4cacc0", GitTreeState:"clean", BuildDate:"2019-10-17T19:23:55Z", GoVersion:"go1.10.8", Compiler:"gc", Platform:"linux/amd64"}
```

Openshift on AWS

On Openshift we are using Kubernetes Deployment and StatefulSets in combination with Openshift ImageStreams. It's possible to define the trigger as annotation of the Deployment Object:
https://docs.openshift.com/container-platform/3.11/dev_guide/managing_images.html#image-stream-kubernetes-resources

> In the following example the trigger fires when the example:latest imagestream tag is updated. Upon firing, the object’s pod template image reference for the web container is updated with a new image value

In our charts, we are defining a dummy value since the image field value of the pod spec is controlled by openshift like:

```
---
# Source: timp/templates/qs/qs-dpl.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  annotations:
    image.openshift.io/triggers: |
      [{
        "from": {
          "kind": "ImageStreamTag",
          "name": "jkr-timp-qs:latest"
        },
        "fieldPath": "spec.template.spec.containers[?(@.name==\"qs\")].image"
      }]
  name: jkr-timp-qs
  labels:
    app: "jkr-timp"
spec:
  replicas: 1
  selector:
    matchLabels:
      release: jkr-timp
      app: "jkr-timp"
      app.kubernetes.io/component: qs
  template:
    metadata:
      annotations:
        checksum/secret: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
      labels:
        release: jkr-timp
        app: "jkr-timp"
        app.kubernetes.io/component: qs
    spec:
      containers:
      - name: qs
        image: " "
        imagePullPolicy: IfNotPresent
        ports:
        - name: http
          containerPort: 8080
          protocol: TCP
        env:
        - name: TZ
          value: "Europe/Berlin"
        resources:
          limits:
            memory: 100Mi
        livenessProbe:
          httpGet:
            path: /healthz
            port: http
          initialDelaySeconds: 120
        readinessProbe:
          httpGet:
            path: /healthz
            port: http
          initialDelaySeconds: 5
        volumeMounts:
          - mountPath: /etc/nginx/default.d/qs.conf
            subPath: qs.conf
            name: qs-config
          - mountPath: /usr/share/nginx/html/
            name: data
      securityContext:
        {}
      volumes:
        - name: qs-config
          configMap:
            name: "jkr-timp-qs"
        - name: data
          persistentVolumeClaim:
            claimName: "data-jkr-timp-qs"
```

After applying the manifest, the image field value is changed to `nginx:lastest` by openshift.

On the next helm upgrade, helm replaces the image field value back to `" "`. We identify this it with `helm upgrade -v10`:

```
I1115 16:45:06.834193   49498 request.go:968] Request Body: {"spec":{"template":{"spec":{"$setElementOrder/containers":[{"name":"qs"}],"containers":[{"image":" ","name":"qs"}]}}}}
I1115 16:45:06.834224   49498 round_trippers.go:423] curl -k -v -XPATCH  -H "Accept: application/json" -H "Content-Type: application/strategic-merge-patch+json" -H "Authorization: Bearer XXX" 'https://openshift.local:443/apis/apps/v1/namespaces/experimental/deployments/jkr-timp-qs'
```

The result forces a redeployment since the openshift trigger will reapply the image field to nginx.

This behaviour is new in helm 3.

Is this an expected behaviour? It will break all setups like this. Using Openshift Imagestreams with Deployments is a common use case inside Openshift.

Demo Chart which covers this issue perfectly: https://github.com/jkroepke/helm3-imagetrigger-test/tree/master

jkroepke:
Fixes: https://github.com/helm/helm/issues/6980

Adds a parameter called `--overwrite`. If set to false it enables the helm v2 upgrade behaviour. 

Sometimes a manifest is modified by an operator. In this cases helm should not overwrite the changes. Since this is very rarely i decide to add an optional parameter to `helm upgrade`. If overwrite is enabled, the old CreateTwoWayMergePatch like in helm 2 is used.

I used the code from https://github.com/helm/helm/pull/6124

This is my first-time contribution to this problem. Let me know if I miss something. Thanks

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -31,6 +31,7 @@ import (
 	batch "k8s.io/api/batch/v1"
 	v1 "k8s.io/api/core/v1"
 	apiextv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -142,7 +143,7 @@ func (c *Client) Build(reader io.Reader, validate bool) (ResourceList, error) {
 // occurs, a Result will still be returned with the error, containing all
 // resource updates, creations, and deletions that were attempted. These can be
 // used for cleanup or other logging purposes.
-func (c *Client) Update(original, target ResourceList, force bool) (*Result, error) {
+func (c *Client) Update(original, target ResourceList, force bool, overwrite bool) (*Result, error) {
 	updateErrors := []string{}
 	res := &Result{}
 
@@ -177,7 +178,7 @@ func (c *Client) Update(original, target ResourceList, force bool) (*Result, err
 			return errors.Errorf("no %s with the name %q found", kind, info.Name)
 		}
 
-		if err := updateResource(c, info, originalInfo.Object, force); err != nil {
+		if err := updateResource(c, info, originalInfo.Object, force, overwrite); err != nil {
 			c.Log("error updating the resource %q:\n\t %v", info.Name, err)
 			updateErrors = append(updateErrors, err.Error())
 		}
@@ -323,7 +324,7 @@ func deleteResource(info *resource.Info) error {
 	return err
 }
 
-func createPatch(target *resource.Info, current runtime.Object) ([]byte, types.PatchType, error) {
+func createPatch(target *resource.Info, current runtime.Object, overwrite bool) ([]byte, types.PatchType, error) {
 	oldData, err := json.Marshal(current)
 	if err != nil {
 		return nil, types.StrategicMergePatchType, errors.Wrap(err, "serializing current configuration")
@@ -359,6 +360,19 @@ func createPatch(target *resource.Info, current runtime.Object) ([]byte, types.P
 		return patch, types.MergePatchType, err
 	}
 
+	if !overwrite {
+		// While different objects need different merge types, the parent function
+		// that calls this does not try to create a patch when the data (first
+		// returned object) is nil. We can skip calculating the merge type as
+		// the returned merge type is ignored.
+		if apiequality.Semantic.DeepEqual(oldData, newData) {
+			return nil, types.StrategicMergePatchType, nil
+		}
+
+		patch, err := strategicpatch.CreateTwoWayMergePatch(oldData, newData, versionedObject)
+		return patch, types.StrategicMergePatchType, err
+	}
+
 	patchMeta, err := strategicpatch.NewPatchMetaFromStruct(versionedObject)
 	if err != nil {
 		return nil, types.StrategicMergePatchType, errors.Wrap(err, "unable to create patch metadata from object")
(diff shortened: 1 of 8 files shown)


bacongobbler:
@hickeyma and I talked about maintaining multiple different merge patch strategies a while back to allow users to determine the patch strategy they desire, so this makes sense from a design point of view.

What do you think about renaming the flag to `--merge-strategy two-way`? That way we can default to the current merge strategy (`three-way`), and potentially enable alternative merge patch strategies as they become useful (like Kubernetes' experimental server-side apply logic).

jkroepke:
@bacongobbler 

This was my first intention. Then I choose `--overwrite` since this is more understandable without read the docs then `--merge-strategy`. But I'm also fine with the `--merge-strategy` option but this needs more documentation.

`kubectl apply` has also a parameter `overwrite` but it could be have a different meaning, e.g. the overwrite option at the three-way patch strategy.

bacongobbler:
Yes. It has different semantics. It overwrites objects when there's a merge conflict

jkroepke:
@bacongobbler Do you want to discuss it in your weekly if exists? Didn't know the next steps here.

bacongobbler:
the next two weeks' meetings are cancelled due to KubeCon NA, so if you can hold off until then, that would be a good place to discuss next steps. You could also ask other maintainers to weigh in on this topic in the #helm-dev channel on Kubernetes' slack channel. :)

jkroepke:
I would like to see this in helm 3.1 but I don't want to push or force someone.

It would be great to get some feedback.

pkg/kube/client.go
```diff
@@ -142,7 +143,7 @@ func (c *Client) Build(reader io.Reader, validate bool) (ResourceList, error) {
 // occurs, a Result will still be returned with the error, containing all
 // resource updates, creations, and deletions that were attempted. These can be
 // used for cleanup or other logging purposes.
-func (c *Client) Update(original, target ResourceList, force bool) (*Result, error) {
+func (c *Client) Update(original, target ResourceList, force bool, overwrite bool) (*Result, error) {
```
jdolitsky, line 146: I think we will want to find a way to do this without modifying the method signature

jdolitsky (changes requested):

bacongobbler:
It's unclear to me whether this case is covered already by `--force`. `--force` forgoes the 3-way merge patch and performs a PUT operation on the manifest. As far as I understand, `--overwrite` is just re implementing the same use case but in a different way.

Have you had a chance to look at the `--merge-strategy` idea proposed earlier? It's been 8 months since we last discussed this so it might not be readily available in my mind.

It's also worth asking whether there is there still a use case for this 8 months later.

jkroepke:
> It's also worth asking whether there is there still a use case for this 8 months later.

I meantime we found an other a solution to resolve our problem.

PR #7005, opened 2019-11-18
Add HelmVersion to Capabilities

### LINKED ISSUE #6689: Add HelmVersion to Capabilities
In Helm v2 the Capabilities included `TillerVersion` which would allow you to target and work with different minor versions of Helm. When Tiller went away so did the ability to detect the minor version of Helm.

The idea is to add this back as `HelmVersion` and available though Capabilities.

Note, I'm filing this but not expecting to add this until after v3.0.0 is released.

EricLemieux:
## Context
Adds the HelmVersion the Capabilities so that templates can query against helm minor versions.

From the linked issue
```
In Helm v2 the Capabilities included TillerVersion which would allow you to target and work with different minor versions of Helm. When Tiller went away so did the ability to detect the minor version of Helm.

The idea is to add this back as HelmVersion and available though Capabilities.
```

## Linked issues
Fixes #6689

## Notes
I am unsure if the test I added is satisfactory, as it seems brittle. Is there another approach that would be preferred?
* This test would need to be updated when 3.1 is released
* All of the other BuildInfo fields are empty strings currently

### DIFF
--- a/pkg/chartutil/capabilities.go
+++ b/pkg/chartutil/capabilities.go
@@ -20,6 +20,8 @@ import (
 
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+
+	helmversion "helm.sh/helm/v3/internal/version"
 )
 
 var (
@@ -34,6 +36,7 @@ var (
 			Minor:   "18",
 		},
 		APIVersions: DefaultVersionSet,
+		HelmVersion: helmversion.Get(),
 	}
 )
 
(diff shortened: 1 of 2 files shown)


jdolitsky (approved):
This lgtm

bacongobbler:
Merging this now will likely cause a CI failure due to the reliance on `v3.0` being the output of HelmVersion. On master we're now at `v3.2`.

Can you please perform the following:

- rebase this against master, changing the version number to v3.2
- reinstate the documentation that was present in Helm 2's [release checklist](https://v2.helm.sh/docs/release_checklist/#2-major-minor-releases-change-the-version-number-in-git) for Helm 3. Source code for Helm 3 docs available here: https://github.com/helm/helm-www/blob/master/content/en/docs/community/release_checklist.md

Thanks

EricLemieux:
@bacongobbler Not sure I understand what is missing from the release checklist, should I just add a note about updating this test?

bacongobbler:
Oh! after re-reviewing, it looks like that section is already in place, oddly. my mistake. No action to take on that front. Sorry!

bacongobbler:
Actually now I see where I was getting with that thought.... Just needed more time to think. And more coffee. ☕ 

Yes, please add a note to update the test as well please. That way the tests won't fail with a version mismatch error. Thanks!

bacongobbler (approved):
LGTM

PR #7009, opened 2019-11-19
fix(*): Helm v3 handling of APIVersion v1 charts dependencies

### LINKED ISSUE #6974: Helm v3 packaging apiVersion v1 Charts packages in apiVersion v2 format
Output of `helm version`:

> version.BuildInfo{Version:"v3.0.0", GitCommit:"e29ce2a54e96cd02ccfce88bee4f58bb6e2a28b6", GitTreeState:"clean", GoVersion:"go1.13.4"}

Output of `kubectl version`:

> Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.7", GitCommit:"8fca2ec50a6133511b771a11559e24191b1aa2b4", GitTreeState:"clean", BuildDate:"2019-09-18T14:47:22Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"windows/amd64"}
> Server Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.4", GitCommit:"c27b913fddd1a6c480c229191a087698aa92f0b1", GitTreeState:"clean", BuildDate:"2019-02-28T13:30:26Z", GoVersion:"go1.11.5", Compiler:"gc", Platform:"linux/amd64"}

Cloud Provider/Platform (AKS, GKE, Minikube etc.): Bare-metal K8s deployment

This would be a duplicate of #6887, but that's been closed as an intended feature affecting downstream users of the dependency system.

The bug *here* is that the Chart's apiVersion is still `v1` after packaging, but it no longer works with tools that support Chart apiVersion v1, such as Helm v2.

Along auto-merging of requirements.yaml into Chart.yaml by `helm package`, the Chart apiVersion should have been auto-updated to `v2` so that Helm v2 would have given a hard failure to install with meaningful output.

My preference would have been to continue including requirements.yaml when packaging an apiVersion v1 Chart.

The problem this introduces for Helm v2 is that aliases in requirements.yaml are no longer visible, so installations that use the same chart multiple times via aliases fails and logs something less-directly useful like

> Error: release XXX failed: configmaps "XXX-chartName-config" already exists

TBBle:
This fixes unexpected changes in APIVersion1 (Helm v2) Charts when they are operated on by Helm v3 using `package` or `dependency update`.

The fixes:
- `requirements.yaml` was being unconditionally merged into `Chart.yaml`, which produced packages which failed `helm lint`.
- `requirements.lock` was being converted into `Chart.lock` by `helm dep up` on an APIVersion1 chart.
- The code that loads `requirements.lock` and `requirements.yaml` in APIVersion1 charts for backwards compatibility, was incorrectly excluding those files from `helm package` and the `Files` collection in templates. As well as `Files` behaviour changing, this also broke sub-chart aliases for Helm v2 as those are stored in `requirements.yaml`.

Helm v3 incorrectly reads aliases from `Chart.yaml`'s `dependencies` map in APIVersion1 charts, but I didn't change that, as `helm lint` warns you that this is an unexpected combination already, and it's *as well as* `requirements.yaml` so doesn't break anything.

The changes in the third commit (ae74c91) are possibly not go-idiomatic, and I specifically welcome feedback or suggested improvements.

Fixes: #6974

### DIFF
--- a/cmd/helm/dependency_update_test.go
+++ b/cmd/helm/dependency_update_test.go
@@ -182,7 +182,7 @@ func TestDependencyUpdateCmd_DontDeleteOldChartsOnError(t *testing.T) {
 func createTestingMetadata(name, baseURL string) *chart.Chart {
 	return &chart.Chart{
 		Metadata: &chart.Metadata{
-			APIVersion: chart.APIVersionV1,
+			APIVersion: chart.APIVersionV2,
 			Name:       name,
 			Version:    "1.2.3",
 			Dependencies: []*chart.Dependency{
--- a/pkg/chartutil/chartfile.go
+++ b/pkg/chartutil/chartfile.go
@@ -42,7 +42,16 @@ func LoadChartfile(filename string) (*chart.Metadata, error) {
 //
 // 'filename' should be the complete path and filename ('foo/Chart.yaml')
 func SaveChartfile(filename string, cf *chart.Metadata) error {
+	// Pull out the dependencies of a v1 Chart, since there's no way
+	// to tell the serialiser to skip a field for just this use case
+	savedDependencies := cf.Dependencies
+	if cf.APIVersion == chart.APIVersionV1 {
+		cf.Dependencies = nil
+	}
 	out, err := yaml.Marshal(cf)
+	if cf.APIVersion == chart.APIVersionV1 {
+		cf.Dependencies = savedDependencies
+	}
 	if err != nil {
 		return err
 	}
--- a/pkg/chartutil/save.go
+++ b/pkg/chartutil/save.go
@@ -141,8 +141,17 @@ func Save(c *chart.Chart, outDir string) (string, error) {
 func writeTarContents(out *tar.Writer, c *chart.Chart, prefix string) error {
 	base := filepath.Join(prefix, c.Name())
 
+	// Pull out the dependencies of a v1 Chart, since there's no way
+	// to tell the serialiser to skip a field for just this use case
+	savedDependencies := c.Metadata.Dependencies
+	if c.Metadata.APIVersion == chart.APIVersionV1 {
+		c.Metadata.Dependencies = nil
+	}
 	// Save Chart.yaml
 	cdata, err := yaml.Marshal(c.Metadata)
+	if c.Metadata.APIVersion == chart.APIVersionV1 {
+		c.Metadata.Dependencies = savedDependencies
+	}
 	if err != nil {
 		return err
 	}
--- a/pkg/chart/loader/load.go
+++ b/pkg/chart/loader/load.go
@@ -114,11 +114,17 @@ func LoadFiles(files []*BufferedFile) (*chart.Chart, error) {
 			if err := yaml.Unmarshal(f.Data, c.Metadata); err != nil {
 				return c, errors.Wrap(err, "cannot load requirements.yaml")
 			}
+			if c.Metadata.APIVersion == chart.APIVersionV1 {
+				c.Files = append(c.Files, &chart.File{Name: f.Name, Data: f.Data})
+			}
 		// Deprecated: requirements.lock is deprecated use Chart.lock.
 		case f.Name == "requirements.lock":
 			c.Lock = new(chart.Lock)
 			if err := yaml.Unmarshal(f.Data, &c.Lock); err != nil {
 				return c, errors.Wrap(err, "cannot load requirements.lock")
+			}
+			if c.Metadata.APIVersion == chart.APIVersionV1 {
+				c.Files = append(c.Files, &chart.File{Name: f.Name, Data: f.Data})
 			}
 
 		case strings.HasPrefix(f.Name, "templates/"):
--- a/pkg/downloader/manager.go
+++ b/pkg/downloader/manager.go
@@ -163,7 +163,7 @@ func (m *Manager) Update() error {
 	}
 
 	// Finally, we need to write the lockfile.
-	return writeLock(m.ChartPath, lock)
+	return writeLock(m.ChartPath, lock, c.Metadata.APIVersion == chart.APIVersionV1)
 }
 
 func (m *Manager) loadChartDir() (*chart.Chart, error) {
@@ -634,12 +634,16 @@ func (m *Manager) loadChartRepositories() (map[string]*repo.ChartRepository, err
 }
 
 // writeLock writes a lockfile to disk
-func writeLock(chartpath string, lock *chart.Lock) error {
+func writeLock(chartpath string, lock *chart.Lock, legacyLockfile bool) error {
 	data, err := yaml.Marshal(lock)
 	if err != nil {
 		return err
 	}
-	dest := filepath.Join(chartpath, "Chart.lock")
+	lockfileName := "Chart.lock"
+	if legacyLockfile {
+		lockfileName = "requirements.lock"
+	}
+	dest := filepath.Join(chartpath, lockfileName)
 	return ioutil.WriteFile(dest, data, 0644)
 }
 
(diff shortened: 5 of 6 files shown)


TBBle:
I intend to fix the code-coverage failure, but not immediately. Hopefully later this week, pending other workload.

TBBle:
I haven't added tests to cover the APIVersion v1 behaviours; the changes were visible in the test runs as they broke the dependency tests which were testing APIVersion v2 behaviours with a generated APIVersion v1 chart. Instead I changed those tests to test the behaviour of APIVersion v2 charts, and they pass again.

TBBle:
Is there any kind of process to go through to have a PR reviewed? This's been sitting idle for a couple of weeks now, and I'd love to know if there's more I need to do.

cmd/helm/dependency_update_test.go
```diff
@@ -182,7 +182,7 @@ func TestDependencyUpdateCmd_DontDeleteOldChartsOnError(t *testing.T) {
 func createTestingMetadata(name, baseURL string) *chart.Chart {
 	return &chart.Chart{
 		Metadata: &chart.Metadata{
-			APIVersion: chart.APIVersionV1,
+			APIVersion: chart.APIVersionV2,
```
hickeyma: Can you make this change also in the following places:
- https://github.com/helm/helm/blob/master/pkg/downloader/manager_test.go#L214
- https://github.com/helm/helm/blob/master/pkg/downloader/manager_test.go#L288

pkg/chartutil/chartfile.go
```diff
@@ -42,7 +42,16 @@ func LoadChartfile(filename string) (*chart.Metadata, error) {
 //
 // 'filename' should be the complete path and filename ('foo/Chart.yaml')
 func SaveChartfile(filename string, cf *chart.Metadata) error {
+	// Pull out the dependencies of a v1 Chart, since there's
```
hickeyma: You can exclude fields in the serializer but it is at definition time and would therefore exclude the field from all serializing (e.g. `Dependencies []*Dependency json:"dependencies,omitempty"` --> `Dependencies []*Dependency`). This seems best option as it is just for the v1 case and not in all cases.

hickeyma: Maybe update the comment

pkg/chartutil/save.go
```diff
@@ -141,8 +141,17 @@ func Save(c *chart.Chart, outDir string) (string, error) {
 func writeTarContents(out *tar.Writer, c *chart.Chart, prefix string) error {
 	base := filepath.Join(prefix, c.Name())
 
+	// Pull out the dependencies of a v1 Chart, since there's
```
hickeyma: Same as comment in `pkg/chartutil/chartfile.go `

hickeyma (changes requested):
@TBBle Thank you for pushing this PR. It is a good fix. I have added some comments for consideration.  
One other change is that the `requirements.lock` file should not be saved when packaging. The check could potentially be added before: https://github.com/helm/helm/blob/master/pkg/chartutil/save.go#L182

hickeyma:
@TBBle Thanks for the details. Some additional feedback below.

> requirements.yaml was being unconditionally merged into Chart.yaml, which produced packages which failed helm lint.

I am unable to reproduce this. Can you give more details about it?

> The changes in the third commit (ae74c91) are possibly not go-idiomatic, and I specifically welcome feedback or suggested improvements.

Added feedback in https://github.com/helm/helm/pull/7009#discussion_r355539730

pkg/chartutil/chartfile.go
TBBle: I'm not sure what you want changed in the comment. Would it be clearer to say "there's no way to tell the serialiser to conditionally skip a field"?

TBBle:
Why should `requirements.lock` be excluded from packaging? It was included in packages generated by Helm v2.

pkg/chartutil/chartfile.go
TBBle: > > requirements.yaml was being unconditionally merged into Chart.yaml, which produced packages which failed helm lint.
> 
> I am unable to reproduce this. Can you give more details about it?

With Helm v2.16.2:
```powershell
helm create testchart
```

Add a requirements.yaml:
```yaml
dependencies:
  - name: magic-ip-address
    version: ">0"
    repository: "@stable"
```

With Helm v2.16.2:
```powershell
helm dependency update testchart
```

With Helm v3.0.0 or v3.0.1
```powershell
helm package testchart
helm lint testchart-0.1.0.tgz
```

This will reproduce the failure:
```
==> Linting .\testchart-0.1.0.tgz
[ERROR] Chart.yaml: dependencies are not valid in the Chart file with apiVersion 'v1'. They are valid in apiVersion 'v2'   
```

TBBle:
> > requirements.yaml was being unconditionally merged into Chart.yaml, which produced packages which failed helm lint.
> 
> I am unable to reproduce this. Can you give more details about it?

With Helm v2.16.2:
```powershell
helm create testchart
```

Add a requirements.yaml:
```yaml
dependencies:
  - name: magic-ip-address
    version: ">0"
    repository: "@stable"
```

With Helm v2.16.2:
```powershell
helm dependency update testchart
```

With Helm v3.0.0 or v3.0.1
```powershell
helm package testchart
helm lint testchart-0.1.0.tgz
```

This will reproduce the failure:
```
==> Linting .\testchart-0.1.0.tgz
[ERROR] Chart.yaml: dependencies are not valid in the Chart file with apiVersion 'v1'. They are valid in apiVersion 'v2'   
```

pkg/chartutil/chartfile.go
hickeyma: Maybe something like this: "no way to tell the serialiser to skip a field" --> "no way to tell the serialiser to skip a field for just this use case"

hickeyma:
> Why should requirements.lock be excluded from packaging of APIVersion1 charts? It was included in packages generated by Helm v2, reconfirmed with Helm v2.16.2 with the same setup as the following repro case

I agree that it is included in Helm 2 but the `lock` file is not included in Helm 3. Is it still needed in the package as you will have the dependencies included? I would probably like to mirror Helm 3 in this regard.

hickeyma:
> requirements.yaml was being unconditionally merged into Chart.yaml, which produced packages which failed helm lint.

Thanks for clarification of this in https://github.com/helm/helm/pull/7009#issuecomment-563989554.  I missed the subtly of doing update first with Helm 2.

TBBle:
> > Why should requirements.lock be excluded from packaging of APIVersion1 charts? It was included in packages generated by Helm v2, reconfirmed with Helm v2.16.2 with the same setup as the following repro case

> I agree that it is included in Helm 2 but the `lock` file is not included in Helm 3. Is it still needed in the package as you will have the dependencies included? I would probably like to mirror Helm 3 in this regard.

Personally, I'd keep it around for API stability reasons, in case someone's scripts are using `.Files` to extract the real subchart versions from `requirements.lock` or similar. It seems a reasonably-arbitrary API break for `v1` to suddenly disappear the file.

hickeyma (approved):
LGTM, thanks for the fix @TBBle

technosophos:
picked into release-3.0

Nitive:
@TBBle generating `requirements.lock` instead of `Chart.lock` was a breaking change for helmfile, you can see details in https://github.com/roboll/helmfile/issues/1052. Not sure if it should be fixed on a helm side

TBBle:
As noted in roboll/helmfile#1052, the change back to requirements.yaml for Helm APIVersion 1 charts was reverting an incorrect API breakage, and has now been fixed in helmfile, which was incorrectly generating APIVersion 1 (by default) charts with the APIVersion 2 `Chart.lock`; so were the unit tests in Helm v3 itself until this fix, so it's not surprising, just a shame this wasn't caught earlier in the Helm v3 release cycle.

PR #7013, opened 2019-11-19
Simplify chart installable check

SimonAlling:
The error conveys at least as much information as the boolean, so the latter seems redundant, especially from a type-driven perspective: How should a return value like `false, nil` be interpreted, for example? (Of course it "should never happen"™, but the typechecker won't help with that.) Cheers! :slightly_smiling_face:

### DIFF
--- a/cmd/helm/install.go
+++ b/cmd/helm/install.go
@@ -176,8 +176,7 @@ func runInstall(args []string, client *action.Install, valueOpts *values.Options
 		return nil, err
 	}
 
-	validInstallableChart, err := isChartInstallable(chartRequested)
-	if !validInstallableChart {
+	if err := checkIfInstallable(chartRequested); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


jdolitsky (approved):
I had trouble understanding this at first, but I see how the boolean is redundant/unneeded. There is never a case where `err` is not nil and the boolean is true. Lgtm

cmd/helm/install.go
```diff
@@ -209,13 +208,13 @@ func runInstall(args []string, client *action.Install, valueOpts *values.Options
 	return client.Run(chartRequested, vals)
 }
 
-// isChartInstallable validates if a chart can be installed
+// checkIfInstallable validates if a chart can be installed
 //
 // Application chart type is only installable
-func isChartInstallable(ch *chart.Chart) (bool, error) {
+func checkIfInstallable(ch *chart.Chart) error {
```
bacongobbler, line 214: shouldn't the proper signature be `func checkIfInstallable(ch *chart.Chart) bool {`, leaving the implementation of the error message to the caller?

bacongobbler (changes requested):

cmd/helm/install.go
SimonAlling: `checkIfInstallable` would then be able to convey very little information compared to the current implementation. I have no strong opinion on the matter, but either way, maybe that's a different PR? This one was only intended to simplify the implementation of the function, not redefine its purpose.

bacongobbler (approved):

PR #7024, opened 2019-11-19
v3: Prettify Updated time in helm list

### LINKED ISSUE #6972: helm3 change the deployed timezone and time
how to change the deployed timezone and time?
image

jan25:
Fixes https://github.com/helm/helm/issues/6972.

We used `time.ANSIC` formatter in v2. This PR makes v3 use same formatter as default for `helm ls` commands.

### Current behavior
```
❯ helm ls
NAME                  	NAMESPACE	REVISION	UPDATED                             	STATUS  	CHART            	APP VERSION
hello-world-1574188666	default  	1       	2019-11-19 19:37:46.993599 +0100 CET	deployed	hello-world-0.1.0	1.16.0
```
I believe the UPDATED time is in the Kubernetes cluster TZ

### Changed behavior
```
❯ helm ls
NAME                  	NAMESPACE	REVISION	UPDATED                 	STATUS  	CHART            	APP VERSION
hello-world-1574188666	default  	1       	Tue Nov 19 19:37:46 2019	deployed	hello-world-0.1.0	1.16.0
```
This is prettified time in user local TZ

### Note
This is a preliminary PR to discuss on the issue.
I think the ANSIC formatter is user friendly than the longer formatter. Accuracy up to seconds would be good enough.
If we think TZ is needed, we could provide additional flag `--date-utc`. Thoughts?

### DIFF
--- a/cmd/helm/list.go
+++ b/cmd/helm/list.go
@@ -29,7 +29,6 @@ import (
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/cli/output"
 	"helm.sh/helm/v3/pkg/release"
-	helmtime "helm.sh/helm/v3/pkg/time"
 )
 
 var listHelp = `
@@ -165,7 +164,7 @@ func newReleaseListWriter(releases []*release.Release, timeFormat string) *relea
 		t := "-"
 		if tspb := r.Info.LastDeployed; !tspb.IsZero() {
 			if timeFormat != "" {
-				t = helmtime.Format(tspb, timeFormat)
+				t = tspb.Format(timeFormat)
 			} else {
 				t = tspb.String()
 			}
--- a/cmd/helm/list.go
+++ b/cmd/helm/list.go
@@ -153,45 +153,28 @@ func newReleaseListWriter(releases []*release.Release, formatTime bool) *release
 	// Initialize the array so no results returns an empty array instead of null
 	elements := make([]releaseElement, 0, len(releases))
 	for _, r := range releases {
-		var element releaseElement
+		element := releaseElement{
+			Name:       r.Name,
+			Namespace:  r.Namespace,
+			Revision:   strconv.Itoa(r.Version),
+			Status:     r.Info.Status.String(),
+			Chart:      fmt.Sprintf("%s-%s", r.Chart.Metadata.Name, r.Chart.Metadata.Version),
+			AppVersion: r.Chart.Metadata.AppVersion,
+		}
 
-		if formatTime {
-			element = timeFormattedElement(r)
-		} else {
-			element = releaseElement{
-				Name:       r.Name,
-				Namespace:  r.Namespace,
-				Revision:   strconv.Itoa(r.Version),
-				Status:     r.Info.Status.String(),
-				Chart:      fmt.Sprintf("%s-%s", r.Chart.Metadata.Name, r.Chart.Metadata.Version),
-				AppVersion: r.Chart.Metadata.AppVersion,
-			}
-			t := "-"
-			if tspb := r.Info.LastDeployed; !tspb.IsZero() {
+		t := "-"
+		if tspb := r.Info.LastDeployed; !tspb.IsZero() {
+			if formatTime {
+				t = helmtime.Format(tspb)
+			} else {
 				t = tspb.String()
 			}
-			element.Updated = t
 		}
+		element.Updated = t
 
 		elements = append(elements, element)
 	}
 	return &releaseListWriter{elements}
-}
-
-func timeFormattedElement(r *release.Release) releaseElement {
-	t := "-"
-	if tspb := r.Info.LastDeployed; !tspb.IsZero() {
-		t = helmtime.Format(tspb)
-	}
-	return releaseElement{
-		Name:       r.Name,
-		Namespace:  r.Namespace,
-		Revision:   strconv.Itoa(r.Version),
-		Updated:    t,
-		Status:     r.Info.Status.String(),
-		Chart:      fmt.Sprintf("%s-%s", r.Chart.Metadata.Name, r.Chart.Metadata.Version),
-		AppVersion: r.Chart.Metadata.AppVersion,
-	}
 }
 
 func (r *releaseListWriter) WriteTable(out io.Writer) error {
--- a/cmd/helm/list.go
+++ b/cmd/helm/list.go
@@ -47,8 +47,8 @@ regular expressions (Perl compatible) that are applied to the list of releases.
 Only items that match the filter will be returned.
 
     $ helm list --filter 'ara[a-z]+'
-    NAME                UPDATED                         CHART
-    maudlin-arachnid    1977-09-02 22:04:05 +0000 UTC   alpine-0.1.0
+    NAME                UPDATED                               CHART
+    maudlin-arachnid    1977-09-02 22:04:05.46104 +0000 UTC   alpine-0.1.0
 
 If no results are found, 'helm list' will exit 0, but with no output (or in
 the case of no '-q' flag, only headers).
--- a/cmd/helm/list.go
+++ b/cmd/helm/list.go
@@ -47,8 +47,8 @@ regular expressions (Perl compatible) that are applied to the list of releases.
 Only items that match the filter will be returned.
 
     $ helm list --filter 'ara[a-z]+'
-    NAME                UPDATED                               CHART
-    maudlin-arachnid    1977-09-02 22:04:05.46104 +0000 UTC   alpine-0.1.0
+    NAME                UPDATED                                  CHART
+    maudlin-arachnid    2020-06-18 14:17:46.125134977 +0000 UTC  alpine-0.1.0
 
 If no results are found, 'helm list' will exit 0, but with no output (or in
 the case of no '-q' flag, only headers).
--- a/cmd/helm/list.go
+++ b/cmd/helm/list.go
@@ -105,17 +105,17 @@ func newListCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 					}
 					return nil
 				default:
-					return outfmt.Write(out, newReleaseListWriter(results, client.FormatTime))
+					return outfmt.Write(out, newReleaseListWriter(results, client.TimeFormat))
 				}
 			}
 
-			return outfmt.Write(out, newReleaseListWriter(results, client.FormatTime))
+			return outfmt.Write(out, newReleaseListWriter(results, client.TimeFormat))
 		},
 	}
 
 	f := cmd.Flags()
 	f.BoolVarP(&client.Short, "short", "q", false, "output short (quiet) listing format")
-	f.BoolVar(&client.FormatTime, "format-time", false, "format time")
+	f.StringVar(&client.TimeFormat, "time-format", "", "format time. Example: --time-format 2009-11-17 20:34:10 +0000 UTC")
 	f.BoolVarP(&client.ByDate, "date", "d", false, "sort by release date")
 	f.BoolVarP(&client.SortReverse, "reverse", "r", false, "reverse the sort order")
 	f.BoolVarP(&client.All, "all", "a", false, "show all releases without any filter applied")
@@ -149,7 +149,7 @@ type releaseListWriter struct {
 	releases []releaseElement
 }
 
-func newReleaseListWriter(releases []*release.Release, formatTime bool) *releaseListWriter {
+func newReleaseListWriter(releases []*release.Release, timeFormat string) *releaseListWriter {
 	// Initialize the array so no results returns an empty array instead of null
 	elements := make([]releaseElement, 0, len(releases))
 	for _, r := range releases {
@@ -164,8 +164,8 @@ func newReleaseListWriter(releases []*release.Release, formatTime bool) *release
 
 		t := "-"
 		if tspb := r.Info.LastDeployed; !tspb.IsZero() {
-			if formatTime {
-				t = helmtime.Format(tspb)
+			if timeFormat != "" {
+				t = helmtime.Format(tspb, timeFormat)
 			} else {
 				t = tspb.String()
 			}
--- a/cmd/helm/list.go
+++ b/cmd/helm/list.go
@@ -29,6 +29,7 @@ import (
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/cli/output"
 	"helm.sh/helm/v3/pkg/release"
+	helmtime "helm.sh/helm/v3/pkg/time"
 )
 
 var listHelp = `
@@ -46,8 +47,8 @@ regular expressions (Perl compatible) that are applied to the list of releases.
 Only items that match the filter will be returned.
 
     $ helm list --filter 'ara[a-z]+'
-    NAME                UPDATED                     CHART
-    maudlin-arachnid    Mon May  9 16:07:08 2016    alpine-0.1.0
+    NAME                UPDATED                         CHART
+    maudlin-arachnid    1977-09-02 22:04:05 +0000 UTC   alpine-0.1.0
 
 If no results are found, 'helm list' will exit 0, but with no output (or in
 the case of no '-q' flag, only headers).
@@ -104,16 +105,17 @@ func newListCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 					}
 					return nil
 				default:
-					return outfmt.Write(out, newReleaseListWriter(results))
+					return outfmt.Write(out, newReleaseListWriter(results, client.FormatTime))
 				}
 			}
 
-			return outfmt.Write(out, newReleaseListWriter(results))
+			return outfmt.Write(out, newReleaseListWriter(results, client.FormatTime))
 		},
 	}
 
 	f := cmd.Flags()
 	f.BoolVarP(&client.Short, "short", "q", false, "output short (quiet) listing format")
+	f.BoolVar(&client.FormatTime, "format-time", false, "format time")
 	f.BoolVarP(&client.ByDate, "date", "d", false, "sort by release date")
 	f.BoolVarP(&client.SortReverse, "reverse", "r", false, "reverse the sort order")
 	f.BoolVarP(&client.All, "all", "a", false, "show all releases without any filter applied")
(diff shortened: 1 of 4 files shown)


karuppiah7890:
I insist that timezone information must be present when showing time. It is very important. Or else the user is just confused or might assume wrong timezone or be ignorant of the timezone factor which can cause issues

karuppiah7890:
I don't want the user to even have a doubt about the timezone. So even if the time is in their timezone, I say show it to them

IRobL:
Interesting, I wouldn't have expected that @karuppiah7890, but I suppose it's because I only have experience working with co-located teams where timezone rarely enters into the conversation.  Additionally, in my case, when timezone does come up, it's not too complicated because we tend to simply add/ subtract whole hours instead of dealing with half hours which I presume makes the situation even more complex for you.  Would having a switch + an env variable `HELM_DATE_UTC` improve the `helm list` usability enough for you?  For a long time I've had `TILLER_NAMESPACE` as a config value for working with helm and would have hated having to type the `--tiller-namespace` switch out every time I just wanted to take a glance at my deployments, for instance.  

Great work moving this issue forward, btw @jan25!

jan25:
Just a FYI - helm v2 used no TZ to show for `helm ls`, and I suppose it used tiller local time. Also its prettified in same format as this PR.
I think we can keep this format and clarify in `helm ls help` saying UPDATED time is in users TZ. To view time in UTC, we could provide `--time-utc` flag as i mentioned.

karuppiah7890:
I actually agree and see the fact that lots of software work this way - showing users the time in their timezone, and without TZ info, that is no `T` or `Z`s and `+0530` `+0100` etc. For example - I have seen it in slack, grafana too. But when I first used grafana, I got this doubt - "Hmm. Wait. Am I looking at the right time? Is it my timezone ? Or UTC? Or the grafana server time zone? or what?" . 

So, I've got these points from my side:
1. Helm deployment time accuracy at the level of split second is not needed. Just seconds or even minutes level accuracy is good enough. If people want more accuracy about the deployment, I think checking the k8s resource - for example pod creation time and other timestamps would help, better compared to the Helm release time. As Helm release time is just more regarding Helm. Better accuracy can be obtained only at k8s resource level timestamp data, which may be different from the Helm release time

2. Showing time in a pretty manner is good to the eye. With helm 3's current output, my brain would do some processing (is it just me? idk) to read it something like `Tue Nov 19 19:37:46 2019`. I feel that directly giving that kind of output is good enough. But it's still subjective, since "pretty", "better UX" are all subjective. Only data talks better, for which we need user's feedback.

3. Considering we are doing all this in kubernetes environment, I checked the pod yaml data in kubernetes, for some inspiration. In all the date time yaml fields, it shows timezone information, and when kubectl cli is used, it shows relative time, for example pod started running 4 hours ago. I would take that as inspiration and say either show Helm release time in relative time, or show timezone information along with date time as shown in kubernetes resource yaml fields. For example creationTimestamp is one such field

karuppiah7890:
Also, I'm not going to look too much into Helm v2. Helm v3 is a breaking release, and is the way forward, so new decisions can come in if they seem good. Also, the fact that you say `I suppose it used tiller local time` shows the confusion a user has. Even if it was user's time, you wouldn't know that. In this case, since you are working on a PR, you would know about the Helm 3 behavior. I agree docs can solve the problem, but intuitiveness is also a good thing. People can't be checking the doc for each and every simple thing. If I just show them the timezone information along with the date time, then that's awesome. Or we can see other ways to show time. Like how I mentioned above - taking inspiration from kubectl

jan25:
@karuppiah7890 Thanks for your insights! I don't have strong opinion toward either case - keeping TZ or no TZ. So i can include TZ for now in the default output. Let me push that fix when i get a chance. `2019-11-19 19:37:46 +0100 CET` this'll be the format.

bacongobbler:
would you mind looking into the failing tests?

jan25:
I'm surprised the tests didn't fail for 'new v3 formatter' introduced before this PR: `2019-11-19 19:37:46.993599 +0100 CET`

hickeyma (approved):
LGTM, thanks @jan25.

hickeyma:
I would like to get feedback from @helm/helm-core-maintainers for their opinions before merging .

hickeyma (commented):
Should the `history` command be aligned also to the same format?

For example, 

```console
$ helm ls
NAME       	NAMESPACE	REVISION	UPDATED                      	STATUS  	CHART             	APP VERSION
chrt-106   	default  	1       	2020-01-16 11:28:02 +0000 UTC	deployed	chrt-106-0.1.0    	1.0        
chrt-7210  	default  	1       	2019-12-16 11:18:03 +0000 UTC	deployed	chrt-7210-0.1.0   	1.16.0     
chrt-7258  	default  	3       	2020-01-23 14:54:38 +0000 UTC	deployed	chrt-7258-0.1.0   	1.16.0

$ helm history chrt-7258
REVISION	UPDATED                 	STATUS    	CHART          	APP VERSION	DESCRIPTION     
1       	Wed Dec 18 12:01:48 2019	superseded	chrt-7258-0.1.0	1.16.0     	Install complete
2       	Wed Dec 18 12:02:26 2019	superseded	chrt-7258-0.1.0	1.16.0     	Upgrade complete
3       	Thu Jan 23 14:54:38 2020	deployed  	chrt-7258-0.1.0	1.16.0     	Upgrade complete
```

thomastaylor312:
I definitely think it should align. One worry I have though is does this output change count as a breaking one? I hope people aren't just parsing through table output instead of JSON or YAML, but it does happen and so I don't want this to break people

thomastaylor312 (changes requested):
Blocking for now until we can talk about whether or not this is a breaking change

hickeyma:
> Blocking for now until we can talk about whether or not this is a breaking change

Agree @thomastaylor312, added it to discussion agenda for this week's dev meeting

cjmaloof:
> Blocking for now until we can talk about whether or not this is a breaking change

Just as a data point, while looking into #7486 I found that this may break Helmsman, which parses the `helm list` output [here](https://github.com/Praqma/helmsman/blob/master/internal/app/helm_release.go#L42) expecting particular formats [here](https://github.com/Praqma/helmsman/blob/master/internal/app/helm_time.go#L9). But in the medium term, the new approach seems more robust.

mattfarina:
While this will break helmsman there is a pattern inside of helmsman to handle multiple time strings already. You can find it [here](https://github.com/Praqma/helmsman/blob/d5a9cf6257da2f33622aa3627f6b1e29777d0f3e/internal/app/helm_time.go#L32-L35). If Helm does make a change we should contribute a change there as well.

For Helm 4 it would be nice if we used a spec RFC time format for display.

This whole thing makes me want the `ls` and `history` commands available as JSON. Sounds like a useful feature.

hickeyma:
Thanks @cjmaloof @mattfarina for the feedback. I have a fix for it if needs be. Just trying to confirm if this functionality is being used first.

hickeyma (changes requested):
@jan25 Thanks for the work so far on this PR. The time format for `helm history` command should be the same as `helm ls` command (see https://github.com/helm/helm/pull/7024#pullrequestreview-347362257). Do you mind updating it ?

jan25:
@hickeyma thanks for the inputs. I'll wrap up the `history` changes this weekend

hickeyma:
@cjmaloof @mattfarina I talked with @luisdavim from the Helmsman community about the changes to the time format (i.e. removing the milliseconds). @luisdavim Is happy with the change and suggested holding back on making a change to Helmsman until we have adding the change to a Helm release.

@luisdavim Let me know if I have misinterpreted anything. :)

jan25:
I noticed the `history` command has `--output json|yaml` support. This PR won't change the time fields in those outputs. We could create a separate issue for those? Additionally we must support those output formats for `list` command

jan25:
@hickeyma could you review this when you get a chance?

hickeyma (changes requested):
@jan25 It is nearly there. Few comments inline. Also could you update the following time formats:
- https://github.com/helm/helm/blob/master/cmd/helm/list.go#L51
https://github.com/helm/helm/blob/master/cmd/helm/history.go#L48

jan25:
Updated as suggested. Because `time.go` is thought of as wrapper around stdlib `Time` i put `func Format` in a different file. But semantically both ways are the same though, so moved it to `helm/time.go`

hickeyma (approved):
LGTM, thanks for the effort and perseverance on this PR.

hickeyma:
> I noticed the history command has --output json|yaml support. This PR won't change the time fields in those outputs. 

The time format used for the JSON AND YAML time output is [ISO_8601](https://en.wikipedia.org/wiki/ISO_8601):

```console
$ helm history mydemo -o yaml
- app_version: 1.16.0
  chart: mychart-0.1.0
  description: Install complete
  revision: 1
  status: deployed
  updated: "2020-03-03T17:52:40.769361223Z"

$ helm history mydemo -o json
[{"revision":1,"updated":"2020-03-03T17:52:40.769361223Z","status":"deployed","chart":"mychart-0.1.0","app_version":"1.16.0","description":"Install complete"}]
```

I think that this is a good format if these outputs are to be processed programatically afterwards. It would be good to get the opinions of other  @helm/helm-core-maintainers on this also.

hickeyma:
@jan25 I am going to hold back on merging as I would like to get other @helm/helm-core-maintainers to take a look again.  I will update the  Helmsman community if/when it goes into a release. I am aiming for 3.2.0 which means it can be tested during the RC phase.

hickeyma:
@technosophos @mattfarina I would appreciate some 👀 before merging.

technosophos (changes requested):
I am confused. This appears to be a cosmetic change that will break anything that reads the output of this command. We have a long history of requiring these changes to be behind flags. This was brought up a few times in the comments, but nobody seems to be sticking with the long-standing view that Helm's output is considered "stable" from release to release.

I hate to be hard-nosed about this stuff. But if we take compatibility seriously, we _cannot_ just change output even if it looks nicer. And notifying a few select people does not eliminate the requirement that we keep the output stable.

We know people use this data in shell scripts and CI pipelines all over the place. Every one of those projects will have to fix a suddenly breaking behavior because we made this change. I do not believe this is the right thing to do.

This MUST be behind a flag.

TheNotary:
I think you're missing the point of this issue.  Helm never intended to change the formatting of it's output, it drifted as an unforeseen situation caused by the removal of tiller.  I agree, when that happened, it should have been behind a switch, but again, that was a defect that no one detected until the originator of this issue spoke up and the investigation began.

mattfarina:
@TheNotary The problem is that this is a breaking change. There are tools that execute commands and parse the output. More than we are aware of. Changing the format, even if there is a "better" method, will be a breaking change to those tools.

Helm follows Semantic Versioning which means we'd have to increment the major version for a breaking change like this. I realize that Kubernetes and some other tools may not follow Semantic Versioning. If I'm being honest, the breaking changes between minor versions of Kubernetes have caused problems for me. Helm aims to follow Semantic Versioning unless a security vulnerability requires a break.

This is the kind of change that can happen behind a flag.

TheNotary:
Ok, you two are absolutely right, my mistake.  `helm list` is an application programming interface, it's not a human interface.  When people write software that consumes the outputs of CLIs, they do so under the assumption that this output will be reliable across all platforms and minor/patch versions of the tool.  People relying on this code would probably not have the ability to avoid upgrades to their version of Helm either and would always be using the latest version released.  Millions of hypothetical people would be adversely impacted if this issue were to be resolved and they would have no hope of recovering from the incident.  We should probably schedule a couple more reviewers on this one, I feel as though 6 months isn't enough time for everyone to prepare their input.  I'm wondering if maybe we should put this behind 2 feature flags, one being "--UNLOCK-DANGEROUS-FEATURES` and another for the actual feature, like `--USE-HUMAN-READABLE-LIST-OUTPUT-THAT-WILL-BREAK-CICD-AS-WE-KNOW-IT-WHY-ARE-YOU-DOING-THIS-WHYYYYYYYYYYYYYYYYYY`.

bacongobbler:
Please be more respectful of other's opinions. I can understand your frustration, but there are more constructive ways to discuss topics.

Both @technosophos and @mattfarina have a point. This pull request introduces a breaking change to the CLI output. Our semver contract [quite clearly states](https://github.com/helm/helm/blob/master/CONTRIBUTING.md#semantic-versioning) we cannot introduce changes to the CLI output that would introduce a breaking change.

For posterity:

> Command line commands, flags, and arguments MUST be backward compatible

Based on previous comments in this thread, this would be considered a breaking change, and must be placed behind a feature flag. Something similar to the `--output` flag would be a good place to start. Perhaps `--time-format`, as then we can introduce other timestamp formats as more use cases evolve.

TheNotary:
I hope you guys don't take what I said personally, I think everyone's doing what that think is best and while you may have interpretations of this command output as different than mine, I think you're probably great people.

technosophos:
I will leave it to @TheNotary to decide whether a switch is good, or whether this PR should be closed.

IRobL:
Please remove me from this list ;)

On April 9, 2020 7:56:06 PM CDT, Matt Butcher <notifications@github.com> wrote:
>I will leave it to @TheNotary to decide whether a switch is good, or
>whether this PR should be closed.
>
>-- 
>You are receiving this because you commented.
>Reply to this email directly or view it on GitHub:
>https://github.com/helm/helm/pull/7024#issuecomment-611822607

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

mattfarina:
@IRobL This is a GitHub PR. To be removed you need to remove yourself. On the PR page there is a button to unsubscribe from notifications.

mattfarina:
@TheNotary When [Ken Thompson](https://en.wikipedia.org/wiki/Ken_Thompson) wrote the Unix philosophy one of the things he noted was

> Expect the output of every program to become the input to another, as yet unknown, program.

This happens a lot in CLI tools. For example, the output of Helm is used by tools. We have used it for that in the CI process for the charts repo. Helmsman uses it. There are two examples of this part of the Unix philosophy at work.

The reason we don't make breaking changes easily is because we consider others and the impact on them.

TheNotary:
I found it hypocritical that the person demanding respect used the downvote emoji.  I guess I'm not the only sarcastic one here, lol.  

@mattfarina I concede to the decision you've made there's no need to GH lawyer this to me any further; the PR must be reworked and then the maintainer's list should probably be pinged again if new commits come it so that it can be further checked off.

hickeyma:
Throughout this PR we have discussed/raised the issue of backward compatibility.  I appreciate @jan25 patience as the PR has taken time because of this and trying to finding the best solution.

One of the key requirement and a great attribute of Helm is to respect compatibility in Helm minor releases. It is something I have heard since joining the project.  That is why when I approved this based on technical and usability grounds, I requested further reviews from maintainers as I was not 100%. Something lingered in my mind. I know we seemed to have addressed it in https://github.com/helm/helm/pull/7024#issuecomment-580822136 but still doubt remained.

I am glad to get the feedback from @technosophos @mattfarina. 👍 

@jan25 I hope this is ok with you and that you might be willing to put it behind a flag.

hickeyma (changes requested):
I agree with @technosophos feedback in https://github.com/helm/helm/pull/7024#pullrequestreview-391085230 that this would be a breaking change and should be behind a flag.

jan25:
Thanks. I'll investigate and add this behind a feature soon

hickeyma:
@jan25 Did you get a chance to look at this again? Just asking as we are thinking of cutting a 3.3.0 release soon.

jan25:
@hickeyma Yes, I have changes for `list` command which i pushed now. Could you review? If it looks good i'll go ahead and update the tests and `history` command too.

Sorry, i almost forgot about pushing my changes to branch.

hickeyma:
@jan25 Its looks good to go. Just 2 things:
- Maybe change flag from `pretty` to `tz-format` and description as "time zone format"
- Can you sign the dco when committing: https://helm.sh/blog/helm-dco/index.html

jan25:
I added `--format-time` flag in list command.

And, I undid changes to `history` command as it needs rework because time is confusingly formatted across different output format. E.g. yaml and json uses different time formatting from table output. Probably a good idea to fix `history` in different issue/PR

hickeyma:
Thanks @jan25 for the effort and time you have put into this PR.  Here is output from manual tested that I did.

Before the PR:
```console
$ helm ls
NAME      	NAMESPACE	REVISION	UPDATED                                	STATUS  	CHART           	APP VERSION
mysql     	default  	1       	2020-06-18 14:17:46.125134977 +0000 UTC	deployed	mysql-1.6.4     	5.7.30     
pr-7613-1 	default  	1       	2020-06-18 11:15:32.298743117 +0000 UTC	deployed	pr-7613-0.1.0   	1.16.0     
tstocichrt	default  	1       	2020-06-18 14:05:16.515958406 +0000 UTC	deployed	tstocichrt-0.1.0	1.16.0     
```

With PR:
```console
$ helm ls --format-time
NAME      	NAMESPACE	REVISION	UPDATED                      	STATUS  	CHART           	APP VERSION
mysql     	default  	1       	2020-06-18 14:17:46 +0000 UTC	deployed	mysql-1.6.4     	5.7.30     
pr-7613-1 	default  	1       	2020-06-18 11:15:32 +0000 UTC	deployed	pr-7613-0.1.0   	1.16.0     
tstocichrt	default  	1       	2020-06-18 14:05:16 +0000 UTC	deployed	tstocichrt-0.1.0	1.16.0
```

I appreciate that the PR shortens the time format by removing the milliseconds but I am at this stage unsure of the overall gain of adding the flag. I think the bigger issue is that there is different formats for date-time in the output of Helm commands. I have raised #8352 to tackle this in Hem 4.

/cc @helm/helm-core-maintainers What is your opinion on the PR. Should it do into Helm 3 or handle the format in Helm 4?

jdolitsky (approved):

jdolitsky:
I think its fine for Helm 3 with a flag, but wondering if other formats will be requested. Instead of `--format-time` as boolean but rather string used for format like `--format-time "2006-01-02 15:04:05 -0700 MST"` ?

bacongobbler:
Usual convention is to accept a format flag that accepts control characters as input for each unit measurement. `%h` for hours, `%m` for millis, etc.

For example: `--format-time “%y-%m-%d %h:%m:%s %z`

cmd/helm/list.go
```diff
@@ -145,28 +147,51 @@ type releaseListWriter struct {
...
-			Chart:      fmt.Sprintf("%s-%s", r.Chart.Metadata.Name, r.Chart.Metadata.Version),
-			AppVersion: r.Chart.Metadata.AppVersion,
-		}
-		t := "-"
-		if tspb := r.Info.LastDeployed; !tspb.IsZero() {
-			t = tspb.String()
+		var element releaseElement
+
+		if formatTime {
+			element = timeFormattedElement(r)
```
mattfarina, line 157: Why is it that one place calls out to a function to create the releaseElement while the next location does it inline in this function? The difference between the two instances of releaseElement is the `Updated` portion. Couldn't the releaseElement be created and then the `Updated` portion be set based on the config?

mattfarina (changes requested):
Thanks for the pull request. In addition to my code flow nit, I agree with the other maintainers on the flag usage accepting time strings.

bacongobbler:
3.3.0-rc.1 is being cut tomorrow, so we're removing this from the milestone.

cmd/helm/list.go
```diff
@@ -46,8 +47,8 @@ regular expressions (Perl compatible) that are applied to the list of releases.
 Only items that match the filter will be returned.
 
     $ helm list --filter 'ara[a-z]+'
-    NAME                UPDATED                     CHART
-    maudlin-arachnid    Mon May  9 16:07:08 2016    alpine-0.1.0
```
jdolitsky, line 50: I think this example needs to be reverted, if the default remains "Mon May  9 16:07:08 2016"

jdolitsky (changes requested):

bacongobbler:
Hi @jan25. It looks like there's consensus among the helm maintainers on next steps here for this PR. There's a few comments left in the review that need to be addressed or resolved. Would you mind going through those comments and addressing those, then let us know when this is ready for another review?

Thanks again.

cmd/helm/list.go
jan25: I guess this example was copied over during migration from helm2. Actual default now is `2020-06-18 14:17:46.125134977 +0000 UTC`.

cmd/helm/list.go
jan25: You're right, i'll update this bit to something simpler and create `releaseElement` only once.

jan25: Done :)

jan25: Could you please review?

jan25:
Sorry for not getting back on this. I pushed required changes. Now it is good to review, let me know if something is missing.

technosophos (approved):

mattfarina (approved):
lgtm

mattfarina:
The issue josh called out has been fixed. Going to merge this for the 3.4 release.

mattfarina:
Also, martins comments were addressed and we discussed as Taylor had requested. Should be good to go.

PR #7026, opened 2019-11-19
fix(lint): Remove requirement that directory name and chart name match

sco11morgan:
The constraint that directory name and chart name must match was removed from `helm package` in https://github.com/helm/helm/pull/4141. This PR removes it from linting.

```
$ helm lint ./helm
==> Linting ./helm
[ERROR] Chart.yaml: directory name (helm) and chart name (example) must be the same

```

### DIFF
--- a/pkg/lint/lint_test.go
+++ b/pkg/lint/lint_test.go
@@ -35,12 +35,12 @@ const goodChartDir = "rules/testdata/goodone"
 
 func TestBadChart(t *testing.T) {
 	m := All(badChartDir, values, namespace, strict).Messages
-	if len(m) != 8 {
+	if len(m) != 7 {
 		t.Errorf("Number of errors %v", len(m))
 		t.Errorf("All didn't fail with expected errors, got %#v", m)
 	}
 	// There should be one INFO, 2 WARNINGs and one ERROR messages, check for them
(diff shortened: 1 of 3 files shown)


bacongobbler:
```
pkg/lint/rules/chartfile_test.go:44:5: var `goodChart` is unused (unused)
var goodChart, _ = chartutil.LoadChartfile(goodChartFilePath)
    ^
make: *** [Makefile:84: test-style] Error 1
Exited with code 2
```

james-r-smith:
@sco11morgan, doesn't it also need to be removed in `rules/chartfile.go`?

I'm experiencing this problem as well. This was a huge crutch in v2, and I'm hoping v3 will be more CI-friendly. 😊

mattfarina (approved):

mattfarina:
@james-r-smith The entire function that does the checking is removed in this PR that includes the function in `rules/chartfile.go`.

thiagowfx:
Pingback to https://stackoverflow.com/questions/79166730/helm-how-to-enforce-that-the-directory-name-must-match-the-chart-name for those of us who would like to re-add this check in 2024.

(Sorry for posting in an old thread, but in this case there is value in linking to workarounds.)

PR #7037, opened 2019-11-21
Add a function to set namespace of envsetting

Mikedu1988:
Add a function to set namespace of envsetting, to provide easier integration for non-cli projects.

We have a micro-service using helm-v3 package.
Recently we upgraded the pkg from dev-v3 branch to v3.0.0.0-rc.2, then we found that "Namespace" of env setting changed to a function and the attribute is not accessible any more other than AddFlags function. But we are not a cli tool, we don't want to construct a flagset when we serving requests.

Add the function could make non-cli projects integration with helm api much more easier.

### DIFF
--- a/cmd/helm/helm.go
+++ b/cmd/helm/helm.go
@@ -67,7 +67,7 @@ func main() {
 	actionConfig := new(action.Configuration)
 	cmd := newRootCmd(actionConfig, os.Stdout, os.Args[1:])
 
-	if err := actionConfig.Init(settings.RESTClientGetter(), settings.Namespace(), os.Getenv("HELM_DRIVER"), debug); err != nil {
+	if err := actionConfig.Init(settings.RESTClientGetter(), settings.Namespace, os.Getenv("HELM_DRIVER"), debug); err != nil {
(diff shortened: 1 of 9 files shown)


Mikedu1988:
related issue #7037

bacongobbler:
We've discussed earlier that it'd be better to expose the .namespace field as public (`.Namespace`). Can you make that change instead and write some tests? Thanks

Mikedu1988:
> We've discussed earlier that it'd be better to expose the .namespace field as public (`.Namespace`). Can you make that change instead and write some tests? Thanks

Ok. Will do. 
But that means I'll have to remove or rename the current function Namespace.
I was trying to keep the change size minimum.
 
I'll make the change accordingly :)

Mikedu1988:
Mem storage seems to return all release regardless what namespace provided to the client,
so it's impossible to test "change the namespace in setting" and then list or upgrade the release only in that namespace :(

namespace flag can't be tested unless a RestClientGetter is initialized.

bacongobbler:
https://github.com/helm/helm/pull/10230

Issue #7045, opened 2019-11-21
Helm ignoring child chart’s values propagation to the parent chart


ferantivero:
When a [dependency](https://v2.helm.sh/docs/developing_charts/#chart-dependencies) (`requirements.yaml`) for a chart attempts to [propagate its values (data) to override parent's chart values](https://v2.helm.sh/docs/developing_charts/#importing-child-values-via-requirements-yaml), it results in being ignored.

EDIT: Although reproducible with Helm v2.16.1, I've been to track it down to v.15.0 in which ~- I'm totally guessing -~ it has been introduced by ~https://github.com/helm/helm/pull/6080~ https://github.com/helm/helm/pull/5821 specifically from https://github.com/helm/helm/pull/5821/commits/dda84976bf75e4f3fca9090f0aabe6a7208d709d.

My use case is relatively simple, parent chart can be installed with/without its dependencies using [tags](https://v2.helm.sh/docs/developing_charts/#tags-and-condition-fields-in-requirements-yaml). Therefore, parent chart will provide with some default values for standalone installations but when getting it fully installed with its dep(s), default parent's chart values could be override. It's been working as expected till v2.14.2.

Please let me know if am I missing something around docs/understanding. 

UPDATE: after reverting https://github.com/helm/helm/pull/5821/commits/dda84976bf75e4f3fca9090f0aabe6a7208d709d everything seems to be back to normal

## Repro steps
This can be easily reprod by downloading [foo-0.1.0.zip](https://github.com/helm/helm/files/3876002/foo-0.1.0.zip) + executing the following command or just create your own a chart using the details below:

`helm install foo-0.1.0.zip --dep-up --dry-run --debug`

Expected COMPUTED VALUES (<= v2.14.2; aligned with [docs](https://v2.helm.sh/docs/developing_charts/#importing-child-values-via-requirements-yaml) based on my understanding)

```yaml
...
COMPUTED VALUES:
bar:
  exports:
    data:
      nonce: 42
  global: {}
nonce: 42
...
```

Actual COMPUTED VALUES (>= v2.15.0)

```yaml
...
COMPUTED VALUES:
bar:
  exports:
    data:
      nonce: 42
  global: {}
...
```

Output of `helm version`:

Client: &version.Version{SemVer:"v2.16.1", GitCommit:"bbdfe5e7803a12bbdf97e94cd847859890cf4050", GitTreeState:"clean"}
[...]

PR #7047, opened 2019-11-21
fix(requirements): refactor to use common codepath for table coalescing

ferantivero:
closes: #7045

### what?
regression has been introduced from dda84976bf75e4f3fca9090f0aabe6a7208d709d around the [exports format](https://v2.helm.sh/docs/developing_charts/#using-the-exports-format) behavior, resulting in child values not being propagated to parent's chart values.

### this PR

  - align both formats behaviors and now they will just differ in how to discover their paths
  - add coverage for exports format and fix expected assertions for [parent-child format](https://v2.helm.sh/docs/developing_charts/#using-the-child-parent-format) to match the logic child values always wins
  - just partially revert dda84976bf75e4f3fca9090f0aabe6a7208d709d, this way parents values could be overridden when coalescing
  - after getting better coverage we were able to refact both formats behaviors by merging their propagation logics into a single code path.

### DIFF
--- a/pkg/chartutil/requirements.go
+++ b/pkg/chartutil/requirements.go
@@ -396,7 +396,7 @@ func processImportValues(c *chart.Chart) error {
 	if err != nil {
 		return err
 	}
-	b := cvals.AsMap()
+	b := make(map[string]interface{}, 0)
 	// import values from each dependency if specified in import-values
 	for _, r := range reqs.Dependencies {
 		// only process raw requirement that is found in chart's dependencies (enabled)
@@ -453,6 +453,7 @@ func processImportValues(c *chart.Chart) error {
 			r.ImportValues = outiv
 		}
 	}
+	b = coalesceTables(b, cvals, c.Metadata.Name)
 	y, err := yaml.Marshal(b)
 	if err != nil {
 		return err
--- a/pkg/chartutil/requirements.go
+++ b/pkg/chartutil/requirements.go
@@ -441,12 +441,15 @@ func processImportValues(c *chart.Chart) error {
 					}
 					outiv = append(outiv, nm)
 					s := name + "." + nm["child"]
-					vm, err := cvals.Table(s)
+					// get child table
+					vv, err := cvals.Table(s)
 					if err != nil {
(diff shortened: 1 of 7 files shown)


bacongobbler:
Hi there! A few things here need to be squared away before we can review.

1. Would you mind writing a test to cover #7045?
1. You need to sign off on the DCO before we can accept the PR. More info in the failed check: https://github.com/helm/helm/pull/7047/checks?check_run_id=314807683
1. A longer explanation about the change and why it has to be reverted might be nice. I don't like to revert changes if at all possible, as it will reintroduce the same issue #5821 was trying to solve. Have you considered alternative approaches?

ferantivero:
thanks for your feedback @bacongobbler 

> Would you mind writing a test to cover #7045?

done 

You need to sign off on the DCO before we can accept the PR. 

done

> A longer explanation about the change and why it has to be reverted might be nice. 

When a dependency (requirements.yaml) for a chart attempts to propagate its values (data) to override parent's chart values, it results in being ignored.

Please let's take a look at the commit messages for further details and also please let's have a look at the #7045 where more information, repro steps, and chart details were provided.

> I don't like to revert changes if at all possible, as it will reintroduce the same issue #5821 was trying to solve. 

we got a partial revert now 😸

> Have you considered alternative approaches?

Yes, so now we aligned both formats behavior with the approach leaf-sub-chart-values-always-win. Therefore, sub chart values override the parent values, what I think is now pretty much aligned with [docs](https://v2.helm.sh/docs/developing_charts/#importing-child-values-via-requirements-yaml) and how `exports format` has been working till v2.15.0.

ferantivero:
ping @bacongobbler, friendly reminder 😃

ferantivero:
@bacongobbler just been testing some of our charts using helm3, and I wanted to confirm that this PR is fully aligned with Helm 3 behavior. In other words, #7045 is not present in Helm 3.

technosophos (approved):
I'm going to beg @adamreese to double check (since he did the work on this portion in Helm 3), but this looks good to me.

I am provisionally putting this on 2.16.4.

adamreese (approved):
Great refactor!

PR #7054, opened 2019-11-22
Reintroduce --is-upgrade to template command

databus23:
This PR improves the correctness of `helm template` when rendering charts for the "upgrade" case.

This is an attempt to fix issues reported by multiple uses of the helm diff plugin: https://github.com/databus23/helm-diff/issues/165

Currently it is not possible to properly diff upgrades when templates make use of `.Capabilites` or `.Release.IsUpgrade`.
Mostly because the plugin can't  use the `--validate` flag at the moment because of this: https://github.com/helm/helm/issues/6663#issuecomment-547526721

Adding this flag would allow the plugin to render charts correctly using `helm template --validate --is-upgrade`.

### DIFF
--- a/cmd/helm/template.go
+++ b/cmd/helm/template.go
@@ -129,6 +129,7 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 	f.StringVar(&client.OutputDir, "output-dir", "", "writes the executed templates to files in output-dir instead of stdout")
 	f.BoolVar(&validate, "validate", false, "validate your manifests against the Kubernetes cluster you are currently pointing at. This is the same validation performed on an install")
 	f.BoolVar(&includeCrds, "include-crds", false, "include CRDs in the templated output")
+	f.BoolVar(&client.IsUpgrade, "is-upgrade", false, "set .Release.IsUpgrade instead of .Release.IsInstall")
 	f.StringArrayVarP(&extraAPIs, "api-versions", "a", []string{}, "Kubernetes api versions used for Capabilities.APIVersions")
 
 	return cmd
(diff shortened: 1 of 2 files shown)


databus23:
@thomastaylor312 Could you be bothered to take a look? `helm diff` is seriously broken with helm v3 as soon as a chart makes use of `.Capabilities` at the moment. Thanks! ❤️

hickeyma (commented):
@databus23 I am trying to understand the need/requirement for introducing the `--is-upgrade` flag. At the moment, you can run `helm template` without the `--validate` flag and it will render regardless if resources already exist. The `--validate` flag specifies that it wants to compare against the cluster which should mean accounting for existing resources.  Does this then imply that `--validate` works as expected and that adding a flag on top of this flag is circumventing its functionality?

databus23:
The problem with `--validate` atm is that it can only covers the `install` case at the moment.
It can't be used for previewing a template for an existing release (e.g. the upgrade case).
The exiting resource check makes sure that this error out. This is what this PR tries to solve. The exiting resource checks makes no sense for an exiting release. 
I think correctly rendering template for an existing release is a valid use case in general.

Not using `--validate` is not an option because than the `.Capabilities` section of the template variables is not properly populated. If the chart contains conditionals that make use of `.Capabilities...` (or `.Release.IsUpgrade`) the rendered template will differ from what is actually applied when running `helm upgrade`.

technosophos (approved):
Yeah, this looks nice and concise. Thanks

bacongobbler (approved):

bacongobbler:
Would you mind rebasing this? Thanks!

databus23:
@bacongobbler rebased

travisghansen:
It should be possible to avoid the use of `--validate` if `--api-versions` is specified right? In short, the 2 options should not need to be coupled (they may not be implemented that way).

PR #7060, opened 2019-11-23
Add --lint and --strict options to helm template

### LINKED ISSUE #6506: Add --lint and --strict options to helm template
It would be nice if helm template had --strict and --lint flags which ran the linter when running a template. (--strict might imply --lint --strict)

My main use case is that I'm creating a template and I would like to have fields that must be overridden by the caller of the template, and the template will fail if they aren't set. I want to fail this when compiling the template (which is very much preferable to failing when applying the manifest with kubectl.)

I'm not sure if there's another way to accomplish that that I'm missing.

sco11morgan:
Resolves https://github.com/helm/helm/issues/6506

When writing tests I ran into testdata still using Charts using `apiVersion: v1`.  To be able to use the existing testdata for linting tests I upgraded charts in `cmd/helm/testdata` and `pkg/chart/loader/testdata` to v2


sco11morgan:
closing in favor of https://github.com/helm/helm/pull/7660

PR #7062, opened 2019-11-24
add go version, compiler and platform to .Capabilities.KubeVersion

### LINKED ISSUE #7004: [helm 3] How to get Capabilities.KubeVersion.Platform ?
Hello,

We currently use Capabilities.KubeVersion.Platform to get OS info.
Now, helm 3 is generating the following error :
executing "arch" at <.Capabilities.KubeVersion.Platform>: can't evaluate field Platform in type chartutil.KubeVersion

Does it exist a replacement ?
Thanks for your help

karuppiah7890:
Closes #7004 

I added golang/mock library and generated some mocks with `mockgen`. I hope that's okay


karuppiah7890:
Can someone please review this? In the mean time I'll work on rebasing this with master

pkg/action/mock_action.go
```diff
@@ -0,0 +1,98 @@
...
+	reflect "reflect"
+
+	gomock "github.com/golang/mock/gomock"
+	meta "k8s.io/apimachinery/pkg/api/meta"
+	discovery "k8s.io/client-go/discovery"
+	rest "k8s.io/client-go/rest"
+)
+
+// MockRESTClientGetter is a mock of RESTClientGetter interface
+type MockRESTClientGetter struct {
```
bacongobbler: Can you move this to pkg `internal/test`? It should not be a part of the public API.

pkg/mocks/mock_discovery.go
```diff
@@ -0,0 +1,215 @@
+/*
```
bacongobbler: same here. Can you please move this to `internal/test`?

pkg/action/action_test.go
```diff
@@ -301,3 +308,63 @@ func TestGetVersionSet(t *testing.T) {
...
+		}
+		mockDiscoveryClient.EXPECT().Invalidate()
+		mockDiscoveryClient.EXPECT().ServerVersion().Return(serverVersion, nil)
+		mockDiscoveryClient.EXPECT().ServerGroupsAndResources().Return(nil, nil, nil)
+
+		mockRESTClientGetter := NewMockRESTClientGetter(ctrl)
+
+		mockRESTClientGetter.EXPECT().ToDiscoveryClient().Return(mockDiscoveryClient, nil)
+
+		configuration := &Configuration{RESTClientGetter: mockRESTClientGetter}
```
bacongobbler: Aren't we just testing the behaviour that the mock discovery client responds with the expected output, rather than testing the actual discovery client? I'm having a hard time seeing how this tests the discovery client.

bacongobbler (changes requested):

pkg/action/mock_action.go
karuppiah7890: okay 👍

pkg/mocks/mock_discovery.go
karuppiah7890: okay 👍

pkg/action/action_test.go
karuppiah7890: yeah, so stub. let me see if we can do better. I can't remember much now why this was done 🙈 I think this was the only way that I could find?

karuppiah7890: I need to be able to stub the server right? so when the code calls the server, the stub is used to get the information. just checking if a call was made is not enough. that would mean just verifying if a call was made. This is more around, when the server has a given information, the same is shown by the code that fetches the information and uses it

karuppiah7890: I'm only making sure the mock discover client returns the dummy information that the server has and then I test how the code interprets the information. I'm not testing the discovery client. The subject under test here is `getCapabilities` method here, and I'm stubbing the external calls. If I was testing discovery client, the subject under test, then mocking it would be wrong 🙈

karuppiah7890: does that make sense?

pkg/mocks/mock_discovery.go
karuppiah7890: I've moved the file now 👍

pkg/action/mock_action.go
karuppiah7890: I've moved the files now 👍

karuppiah7890:
I have pushed some changes now and rebased with master branch too. Could you review now? and do check my comment on the mock for discovery client and let me know if it makes sense :)

karuppiah7890:
cc @bacongobbler

technosophos:
I am going to take this one while @bacongobbler is working on some other stuff. As soon as it is rebased, let me know and I will do one quick final review and then merge

technosophos (approved):
LGTM, pending rebase

adamreese:
Also asked in the issue but adding here... What are you using `Platform` for in your chart?

karuppiah7890:
@adamreese I just raised PR for the issue 🙈 Come to think of it, I don't know why people would need the platform.

karuppiah7890:
@technosophos Thanks! Rebase done! 

Just one thought. Currently the test output checks the golang version, which is based on the docker image used in the build which is `circleci/golang:1.13` and the golang version could keep changing, the patch version that is. And in local too, if people use any different version, it could be a problem. Thoughts?

technosophos:
Probably the test should execute `go version` and compare, since what we are testing is whether the internal var matches the actual version used to compile.

karuppiah7890:
@technosophos That would require templatizing the expected output. I think I'm going to create another chart to just test `go version` alone, and use templatization for it

karuppiah7890:
I have updated the PR! Apologies for the late response!

karuppiah7890:
@technosophos @bacongobbler !

mattfarina:
@karuppiah7890 can you rebase off of master?

@bacongobbler does this look good now?

bacongobbler (approved):

karuppiah7890:
Will do! :)

karuppiah7890:
Done rebasing 👍

go.mod
```diff
@@ -19,6 +19,8 @@ require (
 	github.com/evanphx/json-patch v4.5.0+incompatible
 	github.com/gobwas/glob v0.2.3
 	github.com/gofrs/flock v0.7.1
+	github.com/golang/mock v1.4.1
+	github.com/googleapis/gnostic v0.1.0
```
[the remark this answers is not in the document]
bacongobbler: can this be removed? I don't see where this dependency is being used in the project.

bacongobbler (changes requested):
a few version bumps are required before merging. Code looks good enough to merge after comments are addressed. Thanks

go.mod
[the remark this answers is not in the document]
karuppiah7890: It's used here https://github.com/helm/helm/pull/7062/files?file-filters%5B%5D=.go#diff-fcc6d2f6f361b38313543d9944506dd9R27

internal/test/discovery/mock_discovery.go
```diff
@@ -0,0 +1,215 @@
...
+// Source: k8s.io/client-go/discovery (interfaces: CachedDiscoveryInterface)
+
+// Package discovery is a generated GoMock package.
+package discovery
+
+import (
+	reflect "reflect"
+
+	gomock "github.com/golang/mock/gomock"
+	openapi_v2 "github.com/googleapis/gnostic/OpenAPIv2"
```
karuppiah7890, line 27: `github.com/googleapis/gnostic` is used here

karuppiah7890: @bacongobbler

go.mod
```diff
@@ -19,6 +19,8 @@ require (
 	github.com/evanphx/json-patch v4.5.0+incompatible
 	github.com/gobwas/glob v0.2.3
 	github.com/gofrs/flock v0.7.1
+	github.com/golang/mock v1.4.1
```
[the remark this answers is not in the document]
karuppiah7890: Bumped it 👍

karuppiah7890:
Made the version changes

bacongobbler:
Style tests are failing:

```
 
internal/test/discovery/mock_discovery.go:24: File is not `goimports`-ed with -local helm.sh/helm/v3 (goimports)
 
import (
 
internal/test/action/mock_action.go:24: File is not `goimports`-ed with -local helm.sh/helm/v3 (goimports)
 
import (
 
make: *** [Makefile:92: test-style] Error 1
 
 
Exited with code exit status 2 
CircleCI received exit code 2 
```

karuppiah7890:
Oops. Checking..

karuppiah7890:
@bacongobbler Fixed the CI now. Also added a script for mock generation in case anyone wants to update it

bacongobbler (commented):
I guess this begs the question before we consider merging this... Why does the chart need to know about the compiler and Go version used to compile Kubernetes? These were removed earlier because it was believed they never provide any helpful functionality.

It's also important to keep in mind that the platform information only provides information about the **API server**. It does not inform the user what is the **node** os/architecture it will be scheduled upon, which ([in the original issue](https://github.com/helm/helm/issues/7004#issuecomment-616390380)) appears to be what some users are looking for. This is what node selectors and [multi-arch images](https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/) should be used for.

For reference, this question was brought up in the original issue as well. https://github.com/helm/helm/issues/7004#issuecomment-613050271

The more I think about it the more this starts to make less and less sense.

bacongobbler (changes requested):
I don't believe this should be merged after further assessment. The use cases provided in the original issue do not align with the features provided by this pull request, which was why these capabilities were removed from Helm 3.

I'd ask you try and get other maintainers to weigh in here and see if they think differently.

karuppiah7890:
Makes sense. If that's the case, I can close the PR, and we can close the issue too 👍

PR #7078, opened 2019-11-26
v3: Auto-completion framework for plugins

### LINKED ISSUE #7043: Proposal: Auto-completion framework for plugins
When using a Helm plugin, it would be useful to have shell auto-completion as is available for Helm core commands.  While auto-completion of plugin names is already supported (e.g.`helm 2to<TAB>` will already complete to `helm 2to3`), once the user has typed the plugin name, no further completion is available; for example, `helm 2to3 con<TAB>` and  `helm 2to3 convert <TAB>` currently provide no completion choices.  Using the proposed framework, the `2to3` plugin could allow  `helm 2to3 con<TAB>` to complete to `helm 2to3 convert` while `helm 2to3 convert <TAB>` would complete to possible v2 releases to convert.

In an attempt to describe the proposed framework, I’ll start by giving a proposed user documentation for using the framework.  Following that I provide some design decision explanations.

At the basis of this proposal is the reality that auto-completion choices for a plugin must be provided by the plugin itself: Helm cannot know them if the plugin does not provide them since Helm knows nothing about the plugin beyond how to call it. 

*Plugin developer documentation*

## Providing shell auto-completion

Starting with Helm 3.?, a plugin can optionally provide support for shell auto-completion as part of Helm’s existing auto-completion script.  

### Static auto-completion

If your plugin provides its own flags and/or sub-commands, it can inform Helm of them, through an optional `completion.yaml` file in your plugin’s root directory.  The `completion.yaml` file has the form:
```
name: <pluginName>
flags:
- <flag 1>
- <flag 2>
validArgs:
- <arg value 1>
- <arg value 2>
commands:
  name: <commandName>
  flags:
  - <flag 1>
  - <flag 2>
  validArgs:
  - <arg value 1>
  - <arg value 2>
  commands:
     <and so on, recursively> 
```
Notes:
1. All sections are optional but should be provided if applicable.
1. Flags should not include the `-` or `--` prefix.
1. Both short and long flags can and should be specified. A short flag need not be associated with its corresponding long form; they simply should both be listed.
1. Flags need not be ordered in any way, but need to be listed at the correct point in the sub-command hierarchy.
1. Helm existing global flags are already handled by Helm’s auto-completion script.  So plugins need not specify the following flags `--debug`, `--namespace` or `-n`, `--kube-context`, and `--kubeconfig`.
1. The `validArgs` list provides a list of possible completions for the first parameter following a sub-command.  It is not always possible to provide such a list in advance (see Dynamic Completion section below), in which case the `validArgs` section can be omitted.

The `completion.yaml` file is entirely optional.  If it is not provided, Helm will simply not provide auto-completion for the plugin (unless Dynamic Completion is supported by the plugin; see section below).

For example, for the [fullstatus plugin](https://github.com/marckhouzam/helm-fullstatus) which has no sub-commands but accepts the same flags as the `helm status` command, the `completion.yaml` file should simply be:
```
name: fullstatus
flags:
- o
- output
- revision
```

A more intricate example for the [2to3 plugin](https://github.com/helm/helm-2to3), has a `completion.yaml` file of:
```
name: 2to3
commands:
- name: cleanup
  flags:
  - config-cleanup
  - dry-run
  - l
  - label
  - release-cleanup
  - s
  - release-storage
  - tiller-cleanup
  - t
  - tiller-ns
  - tiller-out-cluster
- name: convert
  flags:
  - delete-v2-releases
  - dry-run
  - l
  - label
  - s
  - release-storage
  - release-versions-max
  - t
  - tiller-ns
  - tiller-out-cluster
- name: move
  commands:
  - name: config
    flags:
    - dry-run
```

### Dynamic completion

Helm 3.? also allows plugins to provide dynamic auto-completion.  Dynamic auto-completion is the completion of parameter values or flag values that cannot be defined in advance.  For example, completion of the names of helm releases currently available on the cluster.  For the plugin to support dynamic auto-completion, it must provide an **executable** file in its root directory called `plugin.complete`. When the Helm completion script requires dynamic completions for the plugin, it will execute the `plugin.complete` file, passing it the command-line that needs to be completed.  The `plugin.complete` script will need to have the logic to determine what the proper completion choices are and output them to be consumed by the completion script. 

The output of the `plugin.complete` script should be a new-line separated list such as:
```
rel1 rel2 rel3
```

When `plugin.complete` is called, the plugin environment is set just like when the plugin's main script is called. Therefore, the $HELM_NAMESPACE, $KUBECONFIG, and all other plugin variables will already be set, and their corresponding flags will be removed.

The `plugin.complete` file can be in any executable format; it can be a shell script, a go program, or any other type of program that Helm can execute.
The `plugin.complete` file ***must*** have executable permissions for the user.

In some cases, dynamic completion will require to obtain information from the Kubernetes cluster.  For example, the `helm fullstatus` plugin requires a release name as input. In the `fullstatus` plugin, for its `plugin.complete` script to provide completion for current release names, it can simply run `helm list -q`.  

If you prefer to use the same executable for plugin execution and for plugin completion, you can choose to have your `plugin.complete` script call your main plugin executable with a special parameter of your choosing, requesting it to do the completion. In our example, `plugin.complete` could be implemented like this:
```
#!/usr/bin/env sh

# "$@" is the entire command-line that requires completion.
# It is important to double-quote the "$@" variable to preserve a possibly empty last parameter.
$HELM_PLUGIN_DIR/status.sh --complete "$@"
```
The `fullstatus` plugin's real script (`status.sh`) must then look for the `--complete` flag and if found, printout the proper completions.

### Tips and tricks

1. The shell will automatically filter out completion choices that don't match user input.  A plugin can therefore return all relevant completions without removing the ones that don't match the user input.  For example, if the command-line is `helm fullstatus ngin<TAB>`, the `fullstatus.complete` script can print *all* release names (of the `default` namespace), not just the ones starting with `ngin`; the shell will only retain the ones starting with `ngin`.
1. To simplify dynamic completion support, especially if you have a complex plugin, you can have your  `plugin.complete` script call your main plugin script and request completion choices.  See the Dynamic Completion section above for an example.

---

## Design details

My first idea for completion was for the completion script to call the plugin directly with a special flag to request completion choices, e.g., `helm 2to3 --complete <command-line to complete>`.  However, this would make supporting completion mandatory for plugins, or else who knows what the plugin would do when being called with the flag.

My second attempt was to only add the `plugin.complete` script concept.  If the script was present, it would indicate the plugin supported completion, and that script could safely be called to generate completion choices.  The `plugin.complete` would handle all completions, meaning that sub-commands and flags would also be returned by this script. However, this makes it difficult on plugin developers, as they would have to handle all flag and sub-command parsing.

This is when I thought to introduce the `completion.yaml` configuration file for static completion.  I find this approach makes it easier for plugin developers to provide static completion in a declarative fashion.

### Static completion solution

When generating the auto-completion script and only then (i.e., when running the `helm completion bash|zsh` command), Helm will look for a `completion.yaml` file in each of the configured plugins' directory.  If the file is found, it will parse its yaml content, and will create *fake* Cobra commands and flags to represent the plugins sub-commands and flags.  Cobra will then generate the completion script including all plugins sub-commands and flags.

### Dynamic completion solution

For dynamic completion, when helm loads a plugin, it will register for it a dynamic completion function, which, when triggered, will set the plugin environment variables and call `plugin.complete`, if it exists, passing it the entire command-line.  That function will then take the output of `plugin.complete` and use it as completion choices. 

Comments on the approach, the naming convention used, the yaml format, or any other topic is very welcomed.

marckhouzam:
Closes #7043

This PR adds a framework allowing plugins to provide both static auto-completion and dynamic auto-completion, as described in #7043

I've prepared the `fullstatus` plugin to make use of this framework to illustrate how it works.
The new `completion.yaml` file providing static completion can be seen here:
https://github.com/marckhouzam/helm-fullstatus/blob/master/completion.yaml

The new `plugin.complete` file providing dynamic completion can be seen here:
https://github.com/marckhouzam/helm-fullstatus/blob/master/plugin.complete

I've also done this for the `2to3` plugin in my own fork.
The new `completion.yaml` file providing static completion can be seen here:
https://github.com/marckhouzam/helm-2to3/blob/master/completion.yaml
while the following commit shows how a simple dynamic completion was done:
https://github.com/marckhouzam/helm-2to3/commit/741e547b5ac3e4917bdf88b0ea016eb9b84d9e74

To test:
```
cd helm; make
bin/helm plugin install https://github.com/marckhouzam/helm-fullstatus
bin/helm plugin install https://github.com/marckhouzam/helm-2to3

source <(bin/helm completion bash)

bin/helm fullstatus --ou<TAB>
bin/helm fullstatus --output <TAB>
bin/helm fullstatus <TAB>

bin/helm 2to3 <TAB>
bin/helm 2to3 con<TAB>
bin/helm 2to3 convert <TAB>
# etc
```

### DIFF
--- a/cmd/helm/load_plugins.go
+++ b/cmd/helm/load_plugins.go
@@ -33,7 +33,11 @@ import (
 	"helm.sh/helm/v3/pkg/plugin"
 )
 
-const pluginStaticCompletionFile = "completion.yaml"
+const (
+	pluginStaticCompletionFile        = "completion.yaml"
+	pluginDynamicCompletionExecutable = "plugin.complete"
+	pluginDynamicCompletionIndicator  = "__complete__"
+)
 
 type pluginError struct {
 	error
@@ -105,6 +109,9 @@ func loadPlugins(baseCmd *cobra.Command, out io.Writer) {
 				for k, v := range settings.EnvVars() {
 					env = append(env, fmt.Sprintf("%s=%s", k, v))
 				}
+
+				// Check if the plugin has been called to provide completions
+				main, argv = checkForCompletionRequest(plug, main, argv)
 
 				prog := exec.Command(main, argv...)
 				prog.Env = env
@@ -304,4 +311,19 @@ func loadFile(path string) (*pluginCommand, error) {
 
 	err = yaml.Unmarshal(b, cmds)
 	return cmds, err
+}
+
+func checkForCompletionRequest(plug *plugin.Plugin, main string, args []string) (string, []string) {
+	if len(args) > 0 && args[len(args)-1] == pluginDynamicCompletionIndicator {
+		// The plugin was called to request completion choices.
+		// Call the dynamic completion script of the plugin instead of the main plugin script
+		newMain := strings.Join([]string{plug.Dir, pluginDynamicCompletionExecutable}, string(filepath.Separator))
+		// Remove the temporary pluginDynamicCompletionIndicator parameter
+		newArgs := args[:len(args)-1]
+		if settings.Debug {
+			log.Output(2, fmt.Sprintf("[info] calling %s with args %v\n", newMain, newArgs))
+		}
+		return newMain, newArgs
+	}
+	return main, args
 }
--- a/cmd/helm/load_plugins.go
+++ b/cmd/helm/load_plugins.go
@@ -18,6 +18,8 @@ package main
 import (
 	"fmt"
 	"io"
+	"io/ioutil"
+	"log"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -26,9 +28,12 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+	"sigs.k8s.io/yaml"
 
 	"helm.sh/helm/v3/pkg/plugin"
 )
+
+const pluginStaticCompletionFile = "completion.yaml"
 
 type pluginError struct {
 	error
@@ -59,6 +64,13 @@ func loadPlugins(baseCmd *cobra.Command, out io.Writer) {
 			return nil, err
 		}
 		return u, nil
+	}
+
+	// If we are dealing with the completion command, we try to load more details about the plugins
+	// if available, so as to allow for command and flag completion
+	if subCmd, _, err := baseCmd.Find(os.Args[1:]); err == nil && subCmd.Name() == "completion" {
+		loadPluginsForCompletion(baseCmd, found)
+		return
 	}
 
 	// Now we create commands for all of these.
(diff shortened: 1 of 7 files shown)


marckhouzam:
This PR still needs some tests to be written, but it is ready to be taken for a spin for anyone interested.
Documentation for a plugin developer can be found in the description of #7043

marckhouzam (commented):
I've purposely implemented this PR as two commits. The first adds static completion for plugins, while the second adds dynamic completion. It should ease the review process and/or can allow to split the feature.

marckhouzam:
I've added dynamic completion for the `2to3` plugin in my fork, to help illustrate how to use this framework in different (and simple) ways.

marckhouzam:
https://github.com/helm/acceptance-testing/pull/63 are completion tests for this PR within the acceptance-testing repo.

marckhouzam:
I've forced push a new version of this PR which now contains go tests for the modifications.

The PR is ready to be reviewed.

marckhouzam:
I'm working on a substantial refactoring of auto-completion for Helm, which will change the implementation details for plugin completion, making this PR obsolete.  I will be posting a new PR in the future.

marckhouzam:
This PR is replaced by #7468

PR #5198, opened 2019-01-22
fix: ignore pax_global_header "file" in chart validation

glb:
**What this PR does / why we need it**:

PR #5165 introduced tar file validation but that validation fails when attempting to install charts that are GitHub archives (eg. `https://github.com/org/repo/master.tar.gz`) due to the presence of a meta-file called `pax_global_header`.

Files of type `x` or `g` can be ignored.

**Special notes for your reviewer**:

I wasn't sure how to write a unit test for this. Thoughts?

**If applicable**:
- [ ] this PR contains documentation
- [ ] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

/cc @technosophos

PR #7085, opened 2019-11-26
fix(pkg): ignore pax header files in chart validation

msfuko:
**What this PR does / why we need it:**

this PR is to carry over the fix from helm2 https://github.com/helm/helm/pull/5198
the issue occurs again in helm3.

the validation fails when installing charts that are GitHub archives (eg. https://github.com/org/repo/master.tar.gz)

```
Error: chart illegally contains content outside the base directory: "pax_global_header"
```

Please refer to original PR #5198 if needed.

### DIFF
--- a/pkg/chart/loader/archive.go
+++ b/pkg/chart/loader/archive.go
@@ -126,6 +126,12 @@ func LoadArchiveFiles(in io.Reader) ([]*BufferedFile, error) {
 			continue
 		}
 
+		switch hd.Typeflag {
+		// We don't want to process these extension header files.
+		case tar.TypeXGlobalHeader, tar.TypeXHeader:
+			continue
+		}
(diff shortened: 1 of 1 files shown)


bacongobbler (approved):
thanks for carrying this patch over. Looks good to me!

bacongobbler:
If you could write a unit test to cover this case in the future, that would be lovely. Not necessary, but it's always helpful :)

glb:
@bacongobbler I'd be happy to try writing a unit test for this in a separate PR, would it be possible to get this merged & released before then? We have some folks eager to get going with Helm 3 and failing because of this issue.

bacongobbler (approved):

glb:
Thank you!!

PR #7107, opened 2019-11-28
fix(tests): mapfile is not available on MacOS

marckhouzam:
The mapfile command is not available on a standard MacOS install. To faciliate doing `make test` which runs `validate-license.sh`, let's use an array instead of mapfile.

### DIFF
--- a/scripts/validate-license.sh
+++ b/scripts/validate-license.sh
@@ -27,14 +27,16 @@ find_files() {
   \( -name '*.go' -o -name '*.sh' \)
 }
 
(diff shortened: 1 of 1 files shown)


marckhouzam:
This change fails on Linux.  I'll investigate.

mattfarina (approved):
Thanks for fixing this.

marckhouzam:
And thanks for merging!  I was stuck commenting out the line `@scripts/validate-license.sh` from the Makefile each time I wanted to do a `make test`.

PR #7114, opened 2019-11-29
fix stack overflow error

### LINKED ISSUE #7111: Helm template throws stack overflow error when trying to use file checksum
Output of `helm version`:

Output of `kubectl version`:
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.1", GitCommit:"b7394102d6ef778017f2ca4046abbaa23b88c290", GitTreeState:"clean", BuildDate:"2019-04-08T17:11:31Z", GoVersion:"go1.12.1", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.3-k3s.2", GitCommit:"e7e6a3c4e9a7d80b87793612730d10a863a25980", GitTreeState:"clean", BuildDate:"2019-11-18T18:31:23Z", GoVersion:"go1.13.4", Compiler:"gc", Platform:"linux/amd64"}

k3s

Helm template blows up when trying to use checksums. To reproduce it, simply create a helm chart with "helm create" then edit the deployment.yaml so the spec template contains:
...
template:
    metadata:
      annotations:
        checksum/config: {{ include (print $.Template.BasePath "/deployment.yaml") . | sha256sum }}
...

Try to generate the manifests with: "helm template --output-dir manifests". You will get something like the log attached.
This of course affects any other commands that rely on generating the manifests like "helm install".

[log.txt](https://github.com/helm/helm/files/3904819/log.txt)

zwwhdls:
fix #7111

I found that when a circular reference existed in template "include" function, an endless loop happened in rendering until stack overflow.

My solution is recording the name which is included in a slice. Check the name when the next render comes, if it has been in the recorded slice, throw an error. If not, execute template and pop it out.

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -93,11 +93,19 @@ func warnWrap(warn string) string {
 // initFunMap creates the Engine's FuncMap and adds context-specific functions.
 func (e Engine) initFunMap(t *template.Template, referenceTpls map[string]renderable) {
 	funcMap := funcMap()
+	includedNames := make([]string, 0)
 
 	// Add the 'include' function here so we can close over t.
 	funcMap["include"] = func(name string, data interface{}) (string, error) {
 		var buf strings.Builder
+		for _, n := range includedNames {
+			if n == name {
+				return "", errors.Wrapf(fmt.Errorf("unable to excute template"), "rendering template in a loop, name: %s", name)
+			}
+		}
+		includedNames = append(includedNames, name)
 		err := t.ExecuteTemplate(&buf, name, data)
+		includedNames = includedNames[:len(includedNames)-1]
 		return buf.String(), err
 	}
 
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -100,7 +100,7 @@ func (e Engine) initFunMap(t *template.Template, referenceTpls map[string]render
 		var buf strings.Builder
 		for _, n := range includedNames {
 			if n == name {
-				return "", errors.Wrapf(fmt.Errorf("unable to excute template"), "rendering template in a loop, name: %s", name)
+				return "", errors.Wrapf(fmt.Errorf("unable to excute template"), "rendering template has a nested reference name: %s", name)
 			}
 		}
 		includedNames = append(includedNames, name)
--- a/pkg/engine/engine_test.go
+++ b/pkg/engine/engine_test.go
@@ -460,6 +460,15 @@ func TestAlterFuncMap_include(t *testing.T) {
 		},
 	}
 
+	// Check nested reference in include FuncMap
+	d := &chart.Chart{
+		Metadata: &chart.Metadata{Name: "nested"},
+		Templates: []*chart.File{
+			{Name: "templates/quote", Data: []byte(`{{include "nested/templates/quote" . | indent 2}} dead.`)},
+			{Name: "templates/_partial", Data: []byte(`{{.Release.Name}} - he`)},
+		},
+	}
+
 	v := chartutil.Values{
 		"Values": "",
 		"Chart":  c.Metadata,
@@ -476,6 +485,12 @@ func TestAlterFuncMap_include(t *testing.T) {
 	expect := "  Mistah Kurtz - he dead."
 	if got := out["conrad/templates/quote"]; got != expect {
 		t.Errorf("Expected %q, got %q (%v)", expect, got, out)
+	}
+
+	out, err = Render(d, v)
+	expectErrName := "nested/templates/quote"
+	if err == nil {
+		t.Errorf("Expected err of nested reference name: %v", expectErrName)
 	}
 }
 
--- a/pkg/engine/engine_test.go
+++ b/pkg/engine/engine_test.go
@@ -487,7 +487,7 @@ func TestAlterFuncMap_include(t *testing.T) {
 		t.Errorf("Expected %q, got %q (%v)", expect, got, out)
 	}
 
-	out, err = Render(d, v)
+	_, err = Render(d, v)
 	expectErrName := "nested/templates/quote"
 	if err == nil {
 		t.Errorf("Expected err of nested reference name: %v", expectErrName)


pkg/engine/engine.go
```diff
@@ -93,11 +93,19 @@ func warnWrap(warn string) string {
...
 func (e Engine) initFunMap(t *template.Template, referenceTpls map[string]renderable) {
 	funcMap := funcMap()
+	includedNames := make([]string, 0)
 
 	// Add the 'include' function here so we can close over t.
 	funcMap["include"] = func(name string, data interface{}) (string, error) {
 		var buf strings.Builder
+		for _, n := range includedNames {
+			if n == name {
+				return "", errors.Wrapf(fmt.Errorf("unable to excute template"), "rendering template in a loop, name: %s", name)
```
grissomsh: How about to use following error message - "rendering template has a nested reference name: %s" ?

hickeyma (approved):
@zwwhdls LGTM, thanks for this check.

alf009:
Hi guys are you aware of the fact that is *back-ward incompatible* change that breaks a lot of valid use-cases (and also reduces expressive power of the templates)?

So charts devoted for 3.0.0 helm that are using recursion in templates(which is permitted)  will not work in recent 3.0.2++ helm releases.
One of many valid use-cases for recursion: processing a dictionary and for all values perform an action (e.g., apply `tpl` function to solve helm inability to have computed values).

Another existing use case is here: https://github.com/IBM/charts/blob/8432f96f510d86b9d3204cff0809414d92f439b2/samples/ibm-sch/templates/_config.tpl#L148-L161
where in the template they validate (that loaded yaml file with) metadata does not contain nil objects.

bacongobbler:
can you demonstrate an example that worked prior to this change that broke with this change? That way we have a case which we can address. Thanks.

@zwwhdls can you please look into this case further?

zwwhdls:
@bacongobbler Hi, I'd like to work on this case. But I haven't understand totally about this case, maybe I need @alf009 to demonstrate this example. Thanks.

alf009:
Sorry for late response.

Note that you may consider reducing max the depth of nested template calls (e.g., 1000 calls may be hopefully enough for everyone).
Better way is to validate that the content of parameter used to call the template is same as well (not just the template name) to handle infinite loop detection correctly.

----
I hit this problem in the linked ibm-sch chart in mine the original post; exactly the problematic recursive template is referenced there.

We are using this `ibm-sch` chart as a sub-chart; the problem pops-up to me when I was trying use helm (ver 3.0.2) instead of helm (ver 2.9.1). But if we use helm 3.0.0 there is no problem to render our chart. The error appears only if helm 3.0.2 is used.

Since our chart is not publicly available, I created 2 smaller samples to show the problem,
  both are in the  https://github.com/alf009/helm3-demo repo.

Both charts can be "installed" in helm 3.0.0 (`helm install --dry-run --debug sample CHART_NAME`), while helm 3.0.2 will fail to render them.

---
## Description of the examples:

Recursion1 is example pointing the problem in the ibm-sch (library) chart.
The problematic template is https://github.com/alf009/helm3-demo/blob/a7f5af1c30c112a6ba1816934828ec0cc34ba199/recursion1/charts/ibm-sch/templates/_config.tpl#L148-L161
```
{{- define "sch.validate.valuesMetadata" -}}
{{- $valuesMetadata := (index . 0) -}}
{{- $prefix := (index . 1) -}}
{{- range $key, $value := $valuesMetadata -}}
  {{- $fullkey:= (list $prefix $key | join "." ) | replace ".valuesMetadata." "" -}}
   {{- if (hasPrefix "map[string]" (typeOf $value)) -}}
     {{- include "sch.validate.valuesMetadata" (list $value $fullkey) -}}
   {{- else -}}
     {{- if eq (typeOf $value) "<nil>" -}}
       {{- fail (cat "Unable to process values-metadata.yaml as the key" $fullkey "has a value of <nil>") -}}
     {{- end -}}
   {{- end -}}
{{- end -}}
{{- end -}}
```
this template essentially validates some input content; in particular if `values-metadate.yaml` file  ( https://github.com/alf009/helm3-demo/blob/master/recursion1/values-metadata.yaml) exists in the chart and that it contains non-empty values.


Recursion2 chart in mine repo, just recursively applies `tpl` function to all string values (in a dictionary);
  if installed the NOTES.txt will print the original `.Values`, and then the "rendered" ones.

The problematic template is here:
https://github.com/alf009/helm3-demo/blob/a7f5af1c30c112a6ba1816934828ec0cc34ba199/recursion2/templates/_helpers.tpl#L15

The template is a bit complicated but essentially it does this:
```
  if  kindIs "map" --> recursively process all dictionary keys and store into output dictionary
  if kindIs "slice" --> recursively process all list elements and store them output list
  if kindIs "sting" --> apply tpl funtion and store to the output
  else store the original value to the output.
```

If chart is installed, the NOTES.txt produces this output:
```
#####################################
# Dumping original values
#####################################
global:
  image:
    pullPolicy: IfNotPresent
    repository: hub.docker.com
image:
  name: postgres
  pullPolicy: '{{ .Values.global.image.pullPolicy }}'
  repository: '{{ .Values.global.image.repository }}'
  tag: latest
#####################################


# Rendering values


#####################################
# Dumping rendered values
#####################################
global:
  image:
    pullPolicy: IfNotPresent
    repository: hub.docker.com
image:
  name: postgres
  pullPolicy: IfNotPresent
  repository: hub.docker.com
  tag: latest
#####################################
```
As you can see both the templates in the original values are rendered and replaced; we us this technique with tpl function heavily.

zwwhdls:
Hi @alf009 , sorry for late response.

I have found that in your `_helper.tpl`, `recursion2.valueTemplateRenderer.impl` include itself in line 26:

https://github.com/alf009/helm3-demo/blob/a7f5af1c30c112a6ba1816934828ec0cc34ba199/recursion2/templates/_helpers.tpl#L26

But when template it, its recursion loop only happened 10 times. So it didn't result in endless recursion loop.

The method you raised above is: consider reducing max the depth of nested template calls (e.g., 1000 calls may be hopefully enough for everyone). I think it will help but maybe there is some better way. 

I will think deeply on it and look for a better solution. Whatever the final solution is, I would like to work on it.

Do you have some suggestions on it? @bacongobbler

alf009:
Yes, there is no real use case for infinite recursion (reaching the same state again and again), but in your PR you denied any recursion, which has some reasonable use-cases; hopefully this is clear now.

The way how to solve original problem, e.g., blocking infinite recursion correctly, it is really completely up you yours decision.

petr-buchin:
@zwwhdls could you please provide an approximate date when this will be fixed?

We use a lot of recursion in our templates, and this PR actually blocks our whole company from upgrading Helm.

bacongobbler:
Given this breaks an existing use case, this has to be reverted unless a fix can be provided before 3.1 is released. @zwwhdls, are you planning on PRing a fix to the issues raised or should we go ahead and revert? Thanks.

zwwhdls:
@bacongobbler Hi, can I raise another PR for this case?

zwwhdls:
@bacongobbler Hi, I have raised another issue https://github.com/helm/helm/pull/7443 for this case.
My solution is count the number of recursion in template, when it's over 1000 then raise an error.

Issue #6843, opened 2019-10-30
`helm template` output of manifests does not follow order of filenames and manifests within each file (random results)


AndiDog:
Output of `helm version`:

(all of these versions are buggy as described)

```
Client: &version.Version{SemVer:"v2.14.3", GitCommit:"0e7f3b6637f7af8fcfddb3d2941fcc7cbebb0085", GitTreeState:"clean"}

version.BuildInfo{Version:"v3.0.0-beta.2", GitCommit:"26c7338408f8db593f93cd7c963ad56f67f662d4", GitTreeState:"clean", GoVersion:"go1.12.9"}

version.BuildInfo{Version:"v3.0.0-rc.1", GitCommit:"ee77ae3d40fd599445ebd99b8fc04e2c86ca366c", GitTreeState:"clean", GoVersion:"go1.13.3"}
```

`helm template` loops over file paths and manifests within a file via a `map[string]string`, which produces random results as Go does not guarantee any order. More details in the proposed fix #6842 (sorry, created the PR first).

Example from the PR, using a buggy helm version. Running the same command twice:

```
$ helm-3.0.0-rc.1 template ../helm/cmd/helm/testdata/testcharts/object-order/ | grep -E '^# \d'
# 1
# 2
# 3
# 7
# 8
# 5
...

$ helm-3.0.0-rc.1 template ../helm/cmd/helm/testdata/testcharts/object-order/ | grep -E '^# \d'
# 2
# 3
# 1
# 7
# 8
# 5
...
```

The example manifests here all have the same kind, and therefore should be sorted in a stable manner. Namely, the order from the template file should be kept. This helps the author match source with generated output, which is especially critical for CI/CD pipelines where `helm template` output gets committed to a Git repo and people would like to have readable diffs instead of manifests changing their position with every change.

thomastaylor312:
This makes complete sense to have, but I think in this case we have never guaranteed that `helm template` output is deterministic (please let me know if I missed somewhere that we did!). So we are going to classify this as a feature and bring into 3.1

vikstrous:
I think it's extremely important for `helm template` output to be deterministic. For my use case the order doesn't matter, but I need to be able to diff the output. Without being able to diff the output, how can I make sure that when I change a chart, I'm changing what I intend to change? I don't want to deploy a kubernetes cluster to validate that a trivial config change is made correctly or a refactor hasn't changed the output.

diist:
[...]

PR #7118, opened 2019-11-29
Ensure stable order with multiple entries in same file

cristiklein:
This is a new take on #6843, inspired by #6842 with the following differences:

- No changes are made to test code, only the test data and golden output were amended. In my experiments, `go test -race` (as invoked by `make test`) is sufficient to discover unstable sorting order.
- The last criterion for sorting is the Name of the entry (`.metadata.name`), which means that the sort order is stable even if entries are reordered within a file.


cristiklein:
Fixed by #6842

PR #7122, opened 2019-12-01
override kube namespace with install/upgrade ns

### LINKED ISSUE #7135: Namespace flag of the Install/Upgrade action not working correctly on the sdk
Helm sdk version: 3.0.0

Currently the `Namespace` https://github.com/helm/helm/blob/master/pkg/action/install.go#L78 flag is not passed down to the kube client. So when specifying it on a `Install` action and then running this action, all the object without specified namesapce will be created either in the `default` namespace or in the namespace retrieved by the `RESTClientGetter`. 

One solution is to always set the right namesapce when using the `RESTClientGetter`; however a new client has to be instanciated for each namespace. 
The other solution is to allow the `Namespace` flag to override the kubernetes namespace

Sh4d1:
The `Namespace` flag of the `Install` and `Upgrade` actions does not fix the kube namespace.
This PR enable the `Namespace` flag to specify the NS in which the object are being installed/upgraded

Fixes #7135 

Signed-off-by: Patrik Cyvoct <patrik@ptrk.io>

### DIFF
--- a/pkg/action/hooks.go
+++ b/pkg/action/hooks.go
@@ -54,7 +54,7 @@ func (cfg *Configuration) execHook(rl *release.Release, hook release.HookEvent, 
 			return err
 		}
 
-		resources, err := cfg.KubeClient.Build(bytes.NewBufferString(h.Manifest), true)
+		resources, err := cfg.KubeClient.Build(bytes.NewBufferString(h.Manifest), true, "")
 		if err != nil {
 			return errors.Wrapf(err, "unable to build kubernetes object for %s hook %s", hook, h.Path)
 		}
@@ -126,7 +126,7 @@ func (cfg *Configuration) deleteHookByPolicy(h *release.Hook, policy release.Hoo
 		return nil
 	}
 	if hookHasDeletePolicy(h, policy) {
-		resources, err := cfg.KubeClient.Build(bytes.NewBufferString(h.Manifest), false)
+		resources, err := cfg.KubeClient.Build(bytes.NewBufferString(h.Manifest), false, "")
 		if err != nil {
 			return errors.Wrapf(err, "unable to build kubernetes object for deleting hook %s", h.Path)
 		}
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -113,7 +113,7 @@ func (i *Install) installCRDs(crds []*chart.File) error {
 	totalItems := []*resource.Info{}
 	for _, obj := range crds {
 		// Read in the resources
-		res, err := i.cfg.KubeClient.Build(bytes.NewBuffer(obj.Data), false)
+		res, err := i.cfg.KubeClient.Build(bytes.NewBuffer(obj.Data), false, i.Namespace)
 		if err != nil {
 			return errors.Wrapf(err, "failed to install CRD %s", obj.Name)
 		}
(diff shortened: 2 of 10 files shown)


Sh4d1:
@hickeyma good for review?

marckhouzam:
I'm new to this code, but what scenario does this fix/improve?

Sh4d1:
It fixes the issue of using helm as a go package with the `Namespace` flag of the install and upgrade actions. Right now it doesnt use this flag when creating objects that doesn't specify a namespace @marckhouzam

hickeyma:
@Sh4d1 Is there an issue for this PR?

Sh4d1:
@hickeyma don't think so. I can create one if necessary though

hickeyma:
@Sh4d1 That would be great, thanks. It would be good to get a detailed understanding of the problem and how to reproduce it.

Sh4d1:
@hickeyma  so it should fix #7135

bacongobbler:
Unfortunately, this is a breaking change to the SDK. We cannot accept breaking changes to the codebase.

However, if there was a new field called `.Namespace` as part of the kube client that overrides the behaviour of `client.namespace()`, that would be accepted.

bacongobbler:
https://github.com/helm/helm/blob/f388224186dd22f402622541c65cf977626722ee/pkg/kube/client.go#L51-L55

https://github.com/helm/helm/blob/f388224186dd22f402622541c65cf977626722ee/pkg/kube/client.go#L108-L113

Sh4d1:
@bacongobbler not sure I see why this is breaking change though :thinking:

bacongobbler:
https://github.com/helm/helm/pull/7122/files#diff-a483d6c0863082c3df21f4aad513afe2R125 changes the function signature for `.Build`.

bacongobbler:
https://github.com/helm/helm/blob/master/CONTRIBUTING.md#semantic-versioning

Sh4d1:
Regarding the kube client, not everyone is using a kubeconfig file, and it's already a mess to have a `RESTClientConfig` working :sweat_smile:

Sh4d1:
Fair enough, forgot the `Build` function :sweat_smile:

Sh4d1:
Would a new `BuildWithNamespace` function be okay?

bacongobbler:
> Regarding the kube client, not everyone is using a kubeconfig file, and it's already a mess to have a `RESTClientConfig` working

That's why I mentioned this:

> However, if there was a new field called .Namespace as part of the kube client that overrides the behaviour of client.namespace(), that would be accepted.

A PR that introduces a new field to change the namespace parameter, bypassing the need for a kubeconfig, that would be accepted.

bacongobbler:
IOW, it would look something like this:

```
func (c *Client) namespace() string { 
	if c.Namespace != "" {
		return c.Namespace
	}
	if ns, _, err := c.Factory.ToRawKubeConfigLoader().Namespace(); err == nil { 
		return ns 
	} 
 	return v1.NamespaceDefault 
 } 
```

Sh4d1:
Yep, kay I see , I'll take a closer look at this approach next week and submit a new PR :)

hickeyma:
@Sh4d1 Ping ^

Sh4d1:
Arf sorry, went on Holliday and completely forgot 😅 I'll get to it tomorrow!

PR #7006, opened 2019-11-18
Friendly error message for non-existent Chart.yaml file

anshulvermapatel:
Fixes #6713

Have updated the error message for the missing Chart.yaml
I created a [PR](https://github.com/helm/helm/pull/6743) for this, but that got messed up, so I am creating a fresh one.
This is almost my first PR so please do let me know if there is anything wrong, I will try to fix it.
I have created test for my change for helm/pkg/chart/metadata.go in helm/pkg/chart/metadata_test.go

cc @karuppiah7890

PR #7127, opened 2019-12-02
Friendly error message for non-existent Chart while packaging

### LINKED ISSUE #6713: Helm3: Package error for non-existent directory could be more user friendly
Output of `helm version`:
```
```

Currently, this is what you see when packaging a non-existent directory

```
$ helm package blah-bloo-blee-blee
Error: validation: chart.metadata is required
```

I think it could be more user friendly. It could be like `blah-bloo-blee-blee: No such file or directory`, similar to `ls` command's error. What do you folks think?

anshulvermapatel:
Fixes [#6713](https://github.com/helm/helm/issues/6713)

Have updated the error message for the missing Chart.
Please do let me know if there is anything wrong, I will try to fix it.
I have created test for my change for helm/pkg/chart/metadata.go in helm/pkg/chart/metadata_test.go

I raised a PR for this earlier, creating a new one as I lost my work.

@karuppiah7890 , As you suggested in the last PR [#7006](https://github.com/helm/helm/pull/7006)

> Thanks for the PR. It looks like it doesn't consider the issue that you have linked to. #6713 issue is about helm not giving an appropriate or accurate error for package sub command when the chart itself doesn't exist. Your test seems to be related to chart.yaml file not existing. Could you change the test according to the issue ? Make sure it's a failing test and then you can pass the test by fixing the code :)

I have made some changes while will give an error as you suggested(talking about the whole chart not only the Chart.yaml file.)

Thank You :)


karuppiah7890:
I'm not sure if this is the right way, as I'm afraid this will also be the same error when there's a chart, but no chart.yaml file in it and error message will say no chart was found. 

Could you check more? I'll also debug on where the error about file not being found needs to be caught

anshulvermapatel:
Thank @karuppiah7890 for the reply.
I think the issue was to display a friendly error message than "validation: chart.metadata is required".
I tried/checked and came to know that when there is no Chart.yaml file is there, this error is thrown, but when there is an Empty Chart.yaml file, "validation: chart.metadata.name is required" error is thrown.
It simply means that when there is no Chart.yaml File the metadata is nil , and when the metadata is nil, we should know that Chart.yaml file is missing, which is the main part of a helm chart(I don't think a helm chart is still a chart without Chart.yaml in the directory).
That is why I previously changed the error to "validation: Chart.yaml file is missing" <- this could have been very friendly as the user would know that even the Chart.yaml file does not exist in the directory he/she wants to package.

Curious to know your views on this.

karuppiah7890:
I want to differentiate the error you just said from the error where the
chart directory itself does not exist. I don't think having the same error
message for both makes sense. I'll let other maintainers tell what they feel

bacongobbler:
Yep. There are other ways to trigger the same error. You could have a Chart.yaml present, but the data could be invalid. That would trigger the same error, which would mislead the user into thinking their Chart.yaml is missing.

I'd try to write a unit test that captures the case where a chart has no Chart.yaml, but keep the existing test cases the same. That way you can write a more descriptive error message, capture that specific error case, and retain the existing error message for cases where the chart metadata is missing or invalid.

karuppiah7890:
Actually https://github.com/helm/helm/issues/6713 is about the chart directory mentioned itself missing. The cases being talked about look like a different problem for which we might have to create a GitHub issue to describe the issue and track it

karuppiah7890:
@anshulvermapatel any update on this PR? Also, please check the comments. I feel that we have deviated from the original issue #6713 talks about. Or I could create a PR for that issue, as this PR seems to be solving a different problem which I can't understand very clearly.

anshulvermapatel:
Hello Team,

I am extremely sorry for this long delay, I am back now and I promise, I won't delay anything henceforth on this PR.

Okay, so from the above comments, I understand that the issue if about giving an error message for the directory itself missing. for example -
$ helm package non-existent-directory
Error: non-existent-directory: : no such file or directory

Please let me know if I am still wrong.
I found the code and I made the changes and checked it, it gives the same issue, please confirm and I will write a test for this and will push the change to reflect in the PR.

Thank You Guys.

anshulvermapatel:
Any update?
@karuppiah7890 @bacongobbler

karuppiah7890:
@anshulvermapatel you are right about issue. That's exactly what's been told in #6713 . You can update the PR for the same 😄

anshulvermapatel:
@karuppiah7890 , done, can you please check and let me know.
Thank You :)

anshulvermapatel:
@karuppiah7890 , any update?

cmd/helm/package.go
```diff
@@ -75,6 +76,10 @@ func newPackageCmd(out io.Writer) *cobra.Command {
 				if err != nil {
 					return err
 				}
+				if _, err := os.Stat(path); os.IsNotExist(err) {
```
karuppiah7890: We can bubble out any `Stat` errors we get. Let's not restrict it to just does-no-exist errors.

I just tried this -

```
$ mkdir yet-another-dummy
$ sudo chmod 000 yet-another-dummy
$ ls -al | rg dummy
d---------    3 karuppiahn  staff       96 Jan 19 12:41 yet-another-dummy
$ helm package yet-another-dummy/this-does-not-exist
Error: validation: chart.metadata is required
```

Which leads us to the same issue as before, so an error message that's easier for the user to debug could be just bubble up what error `Stat` gives, which in this case looks like 

```
$ helm package yet-another-dummy/this-does-not-exist
Error: stat /Users/karuppiahn/yet-another-dummy/this-does-not-exist: permission denied
```

karuppiah7890: @anshulvermapatel ^

karuppiah7890:
The test looks good 😄

cmd/helm/package.go
karuppiah7890: One way to avoid such issues is - write as little code as possible to pass the test. In this case, I think this is enough, what do you think?

```
if _, err := os.Stat(args[i]); err != nil {
    return err
}
```

karuppiah7890: And then if you have other cases in mind, you can write tests for it and write more code. For example, other kind of errors like permission denied and so on. 😄

anshulvermapatel:
@karuppiah7890 
Yes, you are right, I got carried with the absolute path and so checked if the absolute path existed or not.
I tried adding a test for permission denied but then for that, a directory with no permission was required and I was not getting pushed, even if I added it using sudo, it did not get pushed.
But I don't think test for permission denied is required as the error should come from Go's standard library.
Anyways, I have made the changes, please have a look if they can be merged.
Thanks

karuppiah7890:
Looks good to me! 😄 cc @bacongobbler

anshulvermapatel:
Guys any update? @karuppiah7890 @bacongobbler @hickeyma

hickeyma (approved):
Thanks for PR @anshulvermapatel. It works for non existent directory and if permission or other I/O problems. The error message is bit generic but I think it is ok for the moment. Also, the correct metadata message is raised if an issue with Chart.yaml file.

Issue #6930, opened 2019-11-11
[helm3][question] How to render CRDs with helm template?


mtougeron:
How do I have `helm3` render the CRDs in the `mychart/crds` directory when running `helm template`? It would be nice if we were able to render all chart files when running `helm template` (even if they are not templated). That would give closer feature parity to running helm install. Is this possible?

edit: I understand that it won't render them as templates but it would be nice if they could be generated/output at all.

Thanks, Mike

Output of `helm version`:

```
version.BuildInfo{Version:"v3.0.0-rc.3", GitCommit:"2ed206799b451830c68bff30af2a52879b8b937a", GitTreeState:"clean", GoVersion:"go1.13.4"}
```

Output of `kubectl version`:

```
Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.4", GitCommit:"67d2fcf276fcd9cf743ad4be9a9ef5828adc082f", GitTreeState:"clean", BuildDate:"2019-09-18T14:51:13Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.4", GitCommit:"67d2fcf276fcd9cf743ad4be9a9ef5828adc082f", GitTreeState:"clean", BuildDate:"2019-09-18T14:41:55Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"linux/amd64"}
```

amuraru:
@mtougeron Mike, can you post an example with the issue you're seeing?

mtougeron:
An example app could look like:

```
$> tree charts/touge
charts/touge
├── Chart.yaml
├── charts
├── crds
│   └── mycrd.yaml
├── templates
│   ├── NOTES.txt
│   ├── _helpers.tpl
│   ├── deployment.yaml
│   ├── ingress.yaml
│   ├── service.yaml
│   └── serviceaccount.yaml
└── values.yaml

3 directories, 9 files
```

```
$> helm template touge --namespace default --output-dir rendered-charts charts/touge
wrote rendered-charts/touge/templates/serviceaccount.yaml
wrote rendered-charts/touge/templates/service.yaml
wrote rendered-charts/touge/templates/deployment.yaml
```

```
$> tree rendered-charts 
rendered-charts
└── touge
    └── templates
        ├── deployment.yaml
        ├── service.yaml
        └── serviceaccount.yaml

2 directories, 3 files
```

[...]

PR #7138, opened 2019-12-03
Add a flag to allow template to output CRDs

GrahamGoudeau:
Addresses #6930 

Add an optional flag that will cause `helm template` to output CRDs along with the templated manifests.


technosophos (approved):
This looks like a solid and non-breaking way to address the issue. Thank you.

bacongobbler (approved):

hlascelles:
The flag is `--include-crds`. eg:

```
helm template projectcalico/tigera-operator --include-crds
```

PR #7142, opened 2019-12-04
Add helm prefix for release dir

daixiang0:
Signed-off-by: Xiang Dai <764524258@qq.com>

For now, extract release pkg like helm-v3.0.0-linux-arm64.tar.gz would get a dir named as `linux-arm64`, it is hard to know that this dir belongs to helm, add "helm" prefix to optimize it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -126,7 +126,7 @@ $(GOIMPORTS):
 .PHONY: build-cross
 build-cross: LDFLAGS += -extldflags "-static"
 build-cross: $(GOX)
-	GO111MODULE=on CGO_ENABLED=0 $(GOX) -parallel=3 -output="_dist/{{.OS}}-{{.Arch}}/$(BINNAME)" -osarch='$(TARGETS)' $(GOFLAGS) -tags '$(TAGS)' -ldflags '$(LDFLAGS)' ./cmd/helm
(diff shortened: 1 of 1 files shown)


daixiang0:
@hickeyma Hi, mind have a look?

hickeyma (commented):
@daixiang0 I am unsure what is the advantage of adding this. The `_dist` directory is within the helm project. It think you would assume that they are Helm distributions. What do you think?

daixiang0:
The `_dist` if within when `make build`, but for release, there are only tgz pkgs, which only include a dir named as `linux-arm64`, which is not clear and meaningless.

hickeyma (commented):
@daixiang0 The `{{.OS}}-{{.Arch}}` directory is within the archive/package you download or get using `get-helm` script. You will know what you are extracting from and not sure what adding a prefix to the directory within the package adds for clarity.

technosophos:
This potentially breaks automation scripts, and thus is considered a minor breaking change. I would recommend not doing this if we can avoid it, or doing it in a way that preserves the old name too.

mattfarina:
Helm has been using this naming pattern for a long time. Both helm 2 and 3 do so.

Helm is downloaded by many CI systems. The scripts that download and unpack Helm know the existing pattern. If the pattern changes it will cause issues in those CI systems. There are three reasons to not break this experience.

1. The change will cause some confusion for people. We should try to minimize that or eliminate it.
2. We will get support requests about it. Likely, a bunch of them. That's more work for us.
3. When changes in compatibility like this happen users don't tend to like it. We don't want situations like that if we can avoid it.

I like the proposed change for the reason stated. But, I would want to do it in Helm 4 rather than 3.

bacongobbler:
I think there’s enough consensus here that this PR should not be merged. Closing!

PR #7147, opened 2019-12-04
fs_test.go: Assume non-root user instead of failing the test.

pmacik:
When the unit tests are run in a container as a random-like user ID (e.g. in Kubernetes or OpenShift) or the user ID is not mentioned in `/etc/passwd` the tests fail complaining like:
```
--- FAIL: TestCopyDirFail_SrcInaccessible (0.00s)
    fs_test.go:181: Failed to get name of current user: user: unknown userid 1026330000
--- FAIL: TestCopyDirFail_DstInaccessible (0.00s)
    fs_test.go:220: Failed to get name of current user: user: unknown userid 1026330000
--- FAIL: TestCopyDirFailOpen (0.00s)
    fs_test.go:320: Failed to get name of current user: user: unknown userid 1026330000
--- FAIL: TestCopyFileFail (0.00s)
    fs_test.go:489: Failed to get name of current user: user: unknown userid 1026330000
--- FAIL: TestIsDir (0.00s)
    fs_test.go:580: Failed to get name of current user: user: unknown userid 1026330000
--- FAIL: TestIsSymlink (0.00s)
    fs_test.go:637: Failed to get name of current user: user: unknown userid 1026330000
```

The `currentUser` here is really only used to check if it is `root` and the very same function that is used here to get the current user (`user.Current()`) is not used anywhere else in the codebase. It should be sufficient to try to get the current user and if that fails, assume the user is a non-root and go on without actually failing the unit test.

This makes the unit tests less flaky.

### DIFF
--- a/internal/third_party/dep/fs/fs_test.go
+++ b/internal/third_party/dep/fs/fs_test.go
@@ -35,7 +35,6 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
-	"os/user"
 	"path/filepath"
 	"runtime"
 	"sync"
@@ -175,13 +174,9 @@ func TestCopyDirFail_SrcInaccessible(t *testing.T) {
 		t.Skip("skipping on windows")
 	}
 
-	var currentUser, err = user.Current()
+	var currentUID = os.Getuid()
 
-	if err != nil {
-		t.Fatalf("Failed to get name of current user: %s", err)
-	}
-
-	if currentUser.Name == "root" {
+	if currentUID == 0 {
 		// Skipping if root, because all files are accessible
 		t.Skip("Skipping for root user")
 	}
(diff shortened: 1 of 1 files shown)


ldimaggi (approved):

ldimaggi (approved):

technosophos:
What is the process, @bacongobbler or @mattfarina, for patching tests in external (but forked) libraries like this? We normally try to stay as close to upstream, right?

sbose78:
Hi @mattfarina and @adamreese Would you be able to take a look at this ? We are blocked with running these tests in OpenSift :-(

sbose78:
@mattfarina @bacongobbler @waveywaves @adamreese Need help with this :)

technosophos (approved):
Alright, I just trawled through the commit history for this package, and it appears that our fork is a rather permanent feature of Helm at this point (though I wish this were not the case). 

Is there a reason why we are testing against a string name? Wouldn't it be better to test against the UID itself? I am concerned that we might mask a legitimate error with the proposed PR.

technosophos (commented):
Oops... sorry... meant to mark this as "Comment" not "Approved"

pedjak:
@technosophos thanks for the hint - the PR is updated to use `os.Getuid()` instead of `user.Current()`. Please take a look.

technosophos (approved):
Ooo! I very much like this new version!

Alright, given that it is my understanding that we (the Helm Core Maintainers) have made a long-term commitment to support this as a permanent fork of the (now abandoned) upstream, I am LGTMing this.

mattfarina (approved):
The change has the same intent. lgtm

PR #7153, opened 2019-12-04
update gopkg.in/yaml.v2 to v2.2.7

karuppiah7890:
Signed-off-by: Karuppiah Natarajan <karuppiah7890@gmail.com>

This is to get the latest features and bug fixes of gopkg.in/yaml.v2 which is being used by a lot of packages which helm depends on. `replace` directive has also been put so that all of them get the latest version!


bacongobbler:
I found only one location where we reference yaml.v2 in the entire codebase, and that was in a unit test.

Dependencies should maintain their own dependency tree. I've opened https://github.com/helm/helm/pull/7352 which removes this from the dependency tree, leaving underlying packages to update on their own.

PR #7157, opened 2019-12-05
fix this inconsistency in the docs

### LINKED ISSUE #7155: [Helm 3] helm package requires argument
According to the description for `helm package`:

> If no path is given, this will look in the present working directory for a
Chart.yaml file, and (if found) build the current directory into a chart.

But this is not the case, shown by running `helm package`:
```
Error: need at least one argument, the path to the chart
```

Recommended solution, rather than throwing an error if no args are provided, append to `args`:
```
if len(args) == 0 {
  args = append(args, ".")
}
```

Should be an easy enough PR. 😉 Only one change to the tests should be needed. I don't have `golangci-lint`, so I couldn't run the tests locally.

Output of `helm version`:
```
```

zwwhdls:
Fx this inconsistency in the docs

### DIFF
--- a/cmd/helm/package.go
+++ b/cmd/helm/package.go
@@ -36,9 +36,6 @@ This command packages a chart into a versioned chart archive file. If a path
 is given, this will look at that path for a chart (which must contain a
 Chart.yaml file) and then package that directory.
 
-If no path is given, this will look in the present working directory for a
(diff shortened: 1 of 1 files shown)


hickeyma (approved):
@zwwhdls LGMT, thanks.

Issue #7140, opened 2019-12-03
Cannot install chart on private Artifactory registry with self-signed cert


jmcelwain:
We are using the Artifactory Helm plugin on Artifactory version `6.13.1`. When installing a chart in a repository via `helm install`, the following error occurs:

```shell
» helm install --generate-name helm-virtual/foo
Error: failed to download "helm-virtual/foo" (hint: running `helm repo update` may help)
```
This repository was added with a specified CA file:

```shell
» helm repo add helm-virtual $ARTIFACTORY_URL/artifactory/helm-virtual --username $USER --password $PASS --ca-file=/tmp/artifactory.crt
"helm-virtual" has been added to your repositories
```

This is confirmed by checking `~/.helm/repository/repositories.yaml`.

By changing the following to be hard coded, I can get past this error: 
```diff
diff --git a/pkg/getter/httpgetter.go b/pkg/getter/httpgetter.go
index 89abfb1c..f19196dd 100644
--- a/pkg/getter/httpgetter.go
+++ b/pkg/getter/httpgetter.go
@@ -89,8 +89,8 @@ func NewHTTPGetter(options ...Option) (Getter, error) {
 }

 func (g *HTTPGetter) httpClient() (*http.Client, error) {
-       if (g.opts.certFile != "" && g.opts.keyFile != "") || g.opts.caFile != "" {
-               tlsConf, err := tlsutil.NewClientTLS(g.opts.certFile, g.opts.keyFile, g.opts.caFile)
+       if true {
+               tlsConf, err := tlsutil.NewClientTLS(g.opts.certFile, g.opts.keyFile, "/tmp/artifactory.crt")
                if err != nil {
                        return nil, errors.Wrap(err, "can't create TLS config for client")
                }
```

Output of `helm version`:
```
» helm version
version.BuildInfo{Version:"v3.0.0", GitCommit:"e29ce2a54e96cd02ccfce88bee4f58bb6e2a28b6", GitTreeState:"clean", GoVersion:"go1.13.4"}
```

Output of `kubectl version`:

```
» kubectl version
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.3", GitCommit:"5e53fd6bc17c0dec8434817e69b04a25d8ae0ff0", GitTreeState:"clean", BuildDate:"2019-06-06T01:44:30Z", GoVersion:"go1.12.5", Compiler:"gc", Platform:"darwin/amd64"}
```

[...]

PR #7164, opened 2019-12-05
fix(install): use ca file for install (#7140)

tychedelia:
Fixes a few bugs related to tls config when installing charts:

1. When installing via relative path, tls config for the selected
repository was not being set.

2. The `--ca-file` flag was not being passed when constructing the
downloader.

3. Setting tls config was not checking for zero value in repo
config, causing flag to get overwritten with empty string.

There's still a few oddities here. I would expect that the flag
passed in on the command line would override the repo config, but
that's not currently the case. Also, we always set the cert, key
and ca files as a trio, when they should be set individually
depending on combination of flags / repo config.

Signed-off-by: James McElwain <jmcelwain@gmail.com>

### DIFF
--- a/pkg/downloader/chart_downloader.go
+++ b/pkg/downloader/chart_downloader.go
@@ -181,8 +181,10 @@ func (c *ChartDownloader) ResolveChartVersion(ref, version string) (*url.URL, er
 		c.Options = append(
 			c.Options,
 			getter.WithURL(rc.URL),
-			getter.WithTLSClientConfig(rc.CertFile, rc.KeyFile, rc.CAFile),
 		)
+		if rc.CertFile != "" || rc.KeyFile != "" || rc.CAFile != "" {
+			getter.WithTLSClientConfig(rc.CertFile, rc.KeyFile, rc.CAFile)
+		}
 		if rc.Username != "" && rc.Password != "" {
 			c.Options = append(
 				c.Options,
@@ -210,12 +212,14 @@ func (c *ChartDownloader) ResolveChartVersion(ref, version string) (*url.URL, er
 	if err != nil {
 		return u, err
 	}
-	if r != nil && r.Config != nil && r.Config.Username != "" && r.Config.Password != "" {
-		c.Options = append(c.Options, getter.WithBasicAuth(r.Config.Username, r.Config.Password))
-	}
 
-	if r.Config.CertFile != "" || r.Config.KeyFile != "" || r.Config.CAFile != "" {
-		c.Options = append(c.Options, getter.WithTLSClientConfig(r.Config.CertFile, r.Config.KeyFile, r.Config.CAFile))
+	if r != nil && r.Config != nil {
+		if r.Config.CertFile != "" || r.Config.KeyFile != "" || r.Config.CAFile != "" {
+			c.Options = append(c.Options, getter.WithTLSClientConfig(r.Config.CertFile, r.Config.KeyFile, r.Config.CAFile))
+		}
+		if r.Config.Username != "" && r.Config.Password != "" {
+			c.Options = append(c.Options, getter.WithBasicAuth(r.Config.Username, r.Config.Password))
+		}
 	}
 
 	// Next, we need to load the index, and actually look up the chart.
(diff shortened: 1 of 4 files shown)


pkg/repo/repotest/server.go
```diff
@@ -143,6 +145,40 @@ func (s *Server) Start() {
...
+	tlsConf, err := tlsutil.NewClientTLS(pub, priv, ca)
+	if err != nil {
+		panic(err)
+	}
+	tlsConf.BuildNameToCertificate()
+	tlsConf.ServerName = "helm.sh"
+	s.srv.TLS = tlsConf
+	s.srv.StartTLS()
+
+	// Set up repositories config with ca file
```
jmcelwain: Rewriting the repository file here is a little gross, but follows the pattern in other tests in chart_downloader_test.go where the server is stopped then restarted explicitly.

pkg/downloader/chart_downloader.go
```diff
@@ -251,9 +259,6 @@ func (c *ChartDownloader) ResolveChartVersion(ref, version string) (*url.URL, er
 		if _, err := getter.NewHTTPGetter(getter.WithURL(rc.URL)); err != nil {
 			return repoURL, err
 		}
-		if r != nil && r.Config != nil && r.Config.Username != "" && r.Config.Password != "" {
```
jmcelwain: `r` is not reassigned, so this line is redundant and can be removed.

mattfarina:
Thanks for the PR to fix this. There is some duplication with #7071. Can you rebase and cleanup. You should be able to build off that.

jmcelwain:
Huh, rebase from `SvenskaSpel/helm` brought in a bunch of garbage, will work to clean up...

pkg/downloader/chart_downloader.go
```diff
@@ -210,12 +212,14 @@ func (c *ChartDownloader) ResolveChartVersion(ref, version string) (*url.URL, er
 	if err != nil {
 		return u, err
 	}
-	if r != nil && r.Config != nil && r.Config.Username != "" && r.Config.Password != "" {
-		c.Options = append(c.Options, getter.WithBasicAuth(r.Config.Username, r.Config.Password))
-	}
 
-	if r.Config.CertFile != "" || r.Config.KeyFile != "" || r.Config.CAFile != "" {
```
jmcelwain: Was missing null check from previous pr.

jmcelwain:
@mattfarina oops, didn't see the other PR had been merged. Just rebased and should be all clean now. Thanks!

pkg/downloader/chart_downloader.go
kaos: I didn't find that null check meaningful, as `r` could only be null if `&{ ... }` fails (out of memory or some such), and `r.cfg` is the `rc` from the argument passed in to `repo.NewChartRepository()`, which can not be null at that point.

jmcelwain: Makes sense, sounds like this could be cleaned up by removing the check entirely. Just wanted to be consistent with guarding in the case it's necessary.

kaos: Indeed, there were quite a few points that could be cleaned up throughout, but didn't want to introduce to many unrelated changes..

hickeyma:
Closes #7140

jmcelwain:
Sorry to be annoying, but is there anyway this can get reviewed and merged sometime soon? This is currently blocking our orgs internal adoption of Helm due to pervasive use of self-signed certs.

jdolitsky (approved):
This seems straightforward. Lgtm

pkg/downloader/chart_downloader.go
```diff
@@ -181,8 +181,10 @@ func (c *ChartDownloader) ResolveChartVersion(ref, version string) (*url.URL, er
 		c.Options = append(
 			c.Options,
 			getter.WithURL(rc.URL),
-			getter.WithTLSClientConfig(rc.CertFile, rc.KeyFile, rc.CAFile),
 		)
+		if rc.CertFile != "" || rc.KeyFile != "" || rc.CAFile != "" {
+			getter.WithTLSClientConfig(rc.CertFile, rc.KeyFile, rc.CAFile)
```
bacongobbler: should this be appended to `c.Options` like it was before?

jmcelwain: fixed in bf5c0ae

mattfarina (approved):

mattfarina:
@jmcelwain Thanks for the fix

PR #7170, opened 2019-12-05
Restoring fetch-dist and sign Make targets

mattfarina:
These make targets are used as part of the release process. They
had yet to be brought over to the v3 branch from the v2 branch
as they were developed after the branching happened.


bacongobbler (approved):
Lgtm

PR #6908, opened 2019-11-07
fix: stop discovery errors from halting chart rendering.

technosophos:
This blocks a particular error (caused by upstream discovery client),
printing a warning instead of failing. It's not a great solution, but is
a stop-gap until Client-Go gets fixed.

Closes #6361

Signed-off-by: Matt Butcher <matt.butcher@microsoft.com>

Issue #6361, opened 2019-09-05
unable to retrieve the complete list of server APIs


planetf1:
Output of `helm version`:
version.BuildInfo{Version:"v3.0+unreleased", GitCommit:"180db556aaf45f34516f8ddb9ddac28d71736a3e", GitTreeState:"clean", GoVersion:"go1.13"}

Output of `kubectl version`:
lient Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.3", GitCommit:"2d3c76f9091b6bec110a5e63777c332469e0cba2", GitTreeState:"clean", BuildDate:"2019-08-19T12:36:28Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.3+IKS", GitCommit:"66a72e7aa8fd2dbf64af493f50f943d7f7067916", GitTreeState:"clean", BuildDate:"2019-08-23T08:07:38Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"linux/amd64"}

Cloud Provider/Platform (AKS, GKE, Minikube etc.): 
IBM Cloud 

Helm chart deployment fails with:
```
➜  charts git:(h2update2) helm install vdc -f ~/etc/cloud-noes.yaml vdc                                                                                                                                <<<
coalesce.go:155: warning: skipped value for image: Not a table.
Error: could not get apiVersions from Kubernetes: unable to retrieve the complete list of server APIs: custom.metrics.k8s.io/v1beta1: the server is currently unable to handle the request
```
(The first error is in a confluent chart... here I discuss the second issue)

Looking at the error I see a similar problem with
```
➜  charts git:(h2update2) kubectl api-resources
NAME                              SHORTNAMES      APIGROUP                           NAMESPACED   KIND
bindings                                                                             true         Binding
componentstatuses                 cs                                                 false        ComponentStatus
configmaps                        cm                                                 true         ConfigMap
endpoints                         ep                                                 true         Endpoints
events                            ev                                                 true         Event
limitranges                       limits                                             true         LimitRange
namespaces                        ns                                                 false        Namespace
nodes                             no                                                 false        Node
persistentvolumeclaims            pvc                                                true         PersistentVolumeClaim
persistentvolumes                 pv                                                 false        PersistentVolume
pods                              po                                                 true         Pod
podtemplates                                                                         true         PodTemplate
[...]

PR #7176, opened 2019-12-06
fix: catch one additional discovery client warning

technosophos:
In comments on #6908 it was suggested that there is a second place where we should catch the discovery client's error. I have not been able to find or reproduce an error that triggers the aforementioned behavior... however, the fix was easy enough.

Can someone who can reproduce the error either (a) provide me a way to reproduce, or (b) test this out and let me know if it fixes the error? *This should not be merged until we know it solves the problem.*

### DIFF
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -167,8 +167,8 @@ func (c *Configuration) releaseContent(name string, version int) (*release.Relea
 // GetVersionSet retrieves a set of available k8s API versions
 func GetVersionSet(client discovery.ServerResourcesInterface) (chartutil.VersionSet, error) {
 	groups, resources, err := client.ServerGroupsAndResources()
-	if err != nil {
-		return chartutil.DefaultVersionSet, err
+	if err != nil && !discovery.IsGroupDiscoveryFailedError(err) {
+		return chartutil.DefaultVersionSet, errors.Wrap(err, "could not get apiVersions from Kubernetes")
 	}
 
 	// FIXME: The Kubernetes test fixture for cli appears to always return nil


technosophos:
See also https://github.com/helm/helm/issues/6361

That problem should have been solved by #6908 already, but it is still related.

technosophos:
I still have not gotten any confirmation that this fixes a problem. So I am holding, still.

bacongobbler:
@kenperkins and @cofyc, since you both raised concerns in #6908, can you please take a look at this and see if this addresses those concerns? Thanks!

cofyc:
This LGTM. I think we can register an external APIService without backend, then DiscoveryAPI will return `ErrGroupDiscoveryFailed` because it will fail to get supported resources for this API group.

bacongobbler:
@cofyc can you confirm this works with a test case to show how this fixes this particular issue? It's unclear to us how this actually fixes the issue for you, so a set of steps to reproduce the issue and how this PR fixes it would be very helpful. Thanks!

bacongobbler:
ping

technosophos:
This is probably safe enough to merge as-is. It will only catch a very specific class of error, so it is unlikely to introduce a new bug. I just feel a little weird fixing a "theoretical" bug.

cofyc:
hi, sorry for being late

I'm not sure how to add a test in helm yet, but I can provide a basic example here:

1, create a normal CRD

```
apiVersion: apiextensions.k8s.io/v1beta1
metadata:
  name: examples.example.com 
spec:
  group: example.com
  scope: Namespaced
  names:
    plural: examples
    singular: example
    kind: Example
  versions:
  - name: v1alpha1 
    served: true
    storage: true
```

2, create an invalid external `APIService`

```
apiVersion: apiregistration.k8s.io/v1beta1
kind: APIService
metadata:
  name: v1alpha1.admission.example.com
spec:
  insecureSkipTLSVerify: true
  group: admission.example.com
  groupPriorityMinimum: 1000
  versionPriority: 15
  service:
    name: service
    namespace: default
  version: v1alpha1
---
apiVersion: v1
kind: Service
metadata:
  name: service
spec:
  ports:
    - name: https-webhook
      port: 443
      targetPort: 443
  selector:
    app.kubernetes.io/name: unkown
```

3, create a helm chart and use `.Capabilities.APIVersions`

```
$ helm create example
$ cat <<EOF > ./example/templates/crdobj.yaml
{{- if .Capabilities.APIVersions.Has "example.com/v1alpha1" }}
apiVersion: example.com/v1alpha1
kind: Example
metadata:
  name: {{ include "example.fullname" . }}
  labels:
    app.kubernetes.io/name: {{ include "example.name" . }}
    helm.sh/chart: {{ include "example.chart" . }}
spec:
  foo: var 
{{- end }}
EOF
```

### testing

1, without this patch

```
./bin/helm uninstall test
./bin/helm install test ./example --debug
```

`test-example` of `example.example.com/v1alpha1` cannot be created because we discarded partial results on the discovery error. So even if the CRD API service (each CRD has a local API service) is available, but `.Capabilities.APIVersions` does not contain it if some other API service has a problem.

2, with this patch

```
./bin/helm uninstall test
./bin/helm install test ./example --debug
```

`test-example` of `example.example.com/v1alpha1` will be created.

technosophos:
:shipit:

bacongobbler (approved):

Issue #4949, opened 2018-11-20
helm lint is missing `--version`


s-spindler:
`helm package` supports setting a chart's version via the `--version` flag.
This is useful when the version is managed by a different tool such as make, maven/gradle etc. as it allows omitting the version in Chart.yaml and setting it when packaging the chart.

`helm lint` should be consistent and support setting a version on the command line as well.
Instead, it gives the following error when trying to lint a chart without version:
```
$ helm lint .    
==> Linting .
[ERROR] Chart.yaml: version is required
```

Output of `helm version`:
`Client: &version.Version{SemVer:"v2.11.0", GitCommit:"2e55dbe1fdb5fdb96b75ff144a339489417b146b", GitTreeState:"clean"}`

s-spindler:
Apparently this was implemented in #5086 - any chance for that PR to get a review?

technosophos:
Anyone interested in taking this on for Helm 3?

technosophos:
I am going to give this a quick try as the last of my 3.3 Helm lint fixes.

technosophos:
I worked on this for a while, and have come to the conclusion that it is outside of the scope of the linter to be able to rewrite parts of the chart before testing. I just can't find a compelling reason to say that in this case it makes sense to test the linter, but modify the input data of the linter. It doesn't match the philosophy we have taken with values, and there are no other places where we allow the linter to overwrite parts of the chart.

I think it is completely reasonable to say that fields must all be set prior to beginning the linting process.

Furthermore, this is not a "consistency with `helm package`" issue. `helm package` is a write operation that writes as output a chart. The linter is a read-only operation. While I can appreciate that it might make a CI pipeline slightly easier, I don't think that is a good enough reason to make this deep of a modification to the linter.

I'm am closing this with "wont fix"

PR #7183, opened 2019-12-07
Issue-4949 feed version of the charts through flags while linting 

openmohan:
Things done:
1. The chart version can be set through cli flag while linting
ie: helm lint chart_name --version=1.0
2. The validation on charts templates is taken out from LoadFiles (pkg->chart->loader->load.go), so that the templates can be loaded first, then version can be set if provided through flags(params) and then the validation function can be ran. This lead to refactoring in many places.

Changing on the ChartLoader interface 
```
type ChartLoader interface {
	Load(version string) (*chart.Chart, error)
}
``` 
will also lead to changes in many places, so thought of getting the validation function outside of LoadFiles.

 

Signed-off-by: openmohan <openmohan@gmail.com>


openmohan:
Fixed the [bug-4949](https://github.com/helm/helm/issues/4949) for helm v3.
@bacongobbler

bacongobbler:
closing as #4949 was closed as a wontfix, so we will not be accepting any PRs attempting to address this.

PR #7186, opened 2019-12-07
Updating sprig and semver from masterminds to fix issue

mattfarina:
There was an issue with ^0.0 ranges that is fixed by this update.

Since Go modules use MVS which looks for the minimal supported
version there were releases to sprig and now an update here to
set the minimum version higher to avoid the bug.


hickeyma (commented):
@mattfarina thanks for this PR. Do you mind adding the references to these fixes so as to provide more context?

mattfarina:
Closing as we should use #7227 instead. There was a second semver bug fix and a new release. The issues being fixed are noted on that PR.

PR #7187, opened 2019-12-08
stop with an error immediately if a file or directory with that name already exists

### LINKED ISSUE #7182: Helm fetch untar needs conflict management
Fetching and extracting charts to the local system does not currently warn the user when the chart name conflicts with an existing file or directory. Instead it has ambiguous behavior depending on the conflict.

Two ambiguous conflicts:

1. File exists in `$PWD` with same name

   Download fails with error message
   ```
   $ echo test > velero
   $ helm fetch stable/velero --untar
   Error: open velero/Chart.yaml: not a directory
   $ cat velero 
   test
   ``` 

2. Directory exists in `$PWD` with same name

   Download succeeds but gets merged into existing directory

   ```
   $ mkdir velero
   $ echo test > velero/test.txt
   $ helm fetch stable/velero --untar
   $ ls velero/
   Chart.yaml	OWNERS		README.md	ci		crds		templates	test.txt	values.yaml
   $ cat velero/test.txt 
   test
   ```

Output of `helm version`:

```
$ helm version --client
Client: &version.Version{SemVer:"v2.11.0", GitCommit:"2e55dbe1fdb5fdb96b75ff144a339489417b146b", GitTreeState:"clean"}
```

Output of `kubectl version`:

```
$ kubectl version --client
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.8", GitCommit:"211047e9a1922595eaa3a1127ed365e9299a6c23", GitTreeState:"clean", BuildDate:"2019-10-15T12:11:03Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"darwin/amd64"}
```

AWS

zwwhdls:
fix #7182

If the path already exited, no matter it is a file or directory, stop with an error immediately.

### DIFF
--- a/cmd/helm/pull_test.go
+++ b/cmd/helm/pull_test.go
@@ -21,6 +21,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"strings"
 	"testing"
 
 	"helm.sh/helm/v3/pkg/repo/repotest"
@@ -41,7 +42,10 @@ func TestPullCmd(t *testing.T) {
 	tests := []struct {
 		name         string
 		args         string
+		existFile    string
+		existDir     string
 		wantError    bool
+		wantErrorMsg string
 		failExpect   string
 		expectFile   string
 		expectDir    bool
@@ -88,13 +92,25 @@ func TestPullCmd(t *testing.T) {
 			expectDir:  true,
 		},
 		{
+			name:         "Fetch untar when file with same name existed",
+			args:         "test/test1 --untar --untardir test1",
+			existFile:    "test1",
+			wantError:    true,
+			wantErrorMsg: "failed to untar",
+		},
+		{
+			name:         "Fetch untar when dir with same name existed",
+			args:         "test/test2 --untar --untardir test2",
+			existDir:     "test2",
+			wantError:    true,
+			wantErrorMsg: "failed to untar",
+		},
+		{
 			name:         "Fetch, verify, untar",
-			args:         "test/signtest --verify --keyring=testdata/helm-test-key.pub --untar --untardir signtest",
-			expectFile:   "./signtest",
+			args:         "test/signtest --verify --keyring=testdata/helm-test-key.pub --untar --untardir signtest2",
+			expectFile:   "./signtest2",
 			expectDir:    true,
 			expectVerify: true,
-			failExpect:   "Failed to untar signtest",
-			wantError:    true,
 		},
 		{
 			name:       "Chart fetch using repo URL",
@@ -129,9 +145,27 @@ func TestPullCmd(t *testing.T) {
 				filepath.Join(outdir, "repositories.yaml"),
 				outdir,
 			)
+			// Create file or Dir before helm pull --untar, see: https://github.com/helm/helm/issues/7182
+			if tt.existFile != "" {
+				file := filepath.Join(outdir, tt.existFile)
+				_, err := os.Create(file)
+				if err != nil {
+					t.Fatal("err")
+				}
+			}
+			if tt.existDir != "" {
+				file := filepath.Join(outdir, tt.existDir)
+				err := os.Mkdir(file, 0755)
+				if err != nil {
+					t.Fatal(err)
+				}
+			}
 			_, out, err := executeActionCommand(cmd)
 			if err != nil {
 				if tt.wantError {
+					if tt.wantErrorMsg != "" && strings.Contains(tt.wantErrorMsg, err.Error()) {
+						t.Fatalf("%q reported error not equel wantErr, reported: %s, wanted: %s", tt.name, err, tt.wantErrorMsg)
+					}
 					return
 				}
 				t.Fatalf("%q reported error: %s", tt.name, err)
--- a/cmd/helm/pull_test.go
+++ b/cmd/helm/pull_test.go
@@ -21,7 +21,6 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
-	"strings"
 	"testing"
 
 	"helm.sh/helm/v3/pkg/repo/repotest"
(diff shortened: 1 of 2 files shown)


pkg/action/pull.go
```diff
@@ -109,13 +109,13 @@ func (p *Pull) Run(chartRef string) (string, error) {
...
-		if fi, err := os.Stat(ud); err != nil {
+		if _, err := os.Stat(ud); err != nil {
 			if err := os.MkdirAll(ud, 0755); err != nil {
 				return out.String(), errors.Wrap(err, "failed to untar (mkdir)")
 			}
 
-		} else if !fi.IsDir() {
-			return out.String(), errors.Errorf("failed to untar: %s is not a directory", ud)
+		} else {
+			return out.String(), errors.Errorf("failed to untar: %s already exited", ud)
```
karuppiah7890: How does this error message look? `failed to untar: a file or directory with the name "signtest" already exists`

karuppiah7890 (changes requested):

pkg/action/pull.go
zwwhdls: sounds great!

technosophos (approved):
LGTM

hickeyma (changes requested):
@zwwhdls Like the fix, thanks. Just wondering about the error message. This is what I get when doing manual testing:

```console
$ helm pull stable/mysql --untar
Error: failed to untar: a file or directory with the name . already exists
```
It seems to be outputting the current directory (`.`) that the archive will be extracted into and not the clashing directory.  Maybe it would be possible to get the clashing directory?

zwwhdls:
@hickeyma Thank you for your review! I almost made a dangerous mistake. Will fix it soon!

technosophos:
Whoa -- good catch, @hickeyma . Sorry I missed that.

cmd/helm/pull_test.go
```diff
@@ -127,9 +145,27 @@ func TestPullCmd(t *testing.T) {
...
+				file := filepath.Join(outdir, tt.existDir)
+				err := os.Mkdir(file, 0755)
+				if err != nil {
+					t.Fatal(err)
+				}
+			}
 			_, out, err := executeActionCommand(cmd)
 			if err != nil {
 				if tt.wantError {
+					if tt.wantErrorMsg != "" && strings.Contains(tt.wantErrorMsg, err.Error()) {
```
karuppiah7890: Looks like in this case, the error message is something that goes out to the user in the console output and we want it to be informative and correct for better UX. 

And in regards to that, I think we could do an exact check on the error message compared to a `Contains` check. This will also catch issues that @hickeyma pointed out 😄 What do you think @zwwhdls ?

karuppiah7890 (changes requested):

cmd/helm/pull_test.go
zwwhdls: I don't think it's necessary.
But if you insist on this, I will fix it too.

zwwhdls:
@hickeyma Hi, thanks again for your comments. I have add a condition for this and waiting for your review. If you approve this change, I will fix it in v2 too.

pkg/action/pull.go
```diff
@@ -109,13 +109,13 @@ func (p *Pull) Run(chartRef string) (string, error) {
 		if !filepath.IsAbs(ud) {
 			ud = filepath.Join(p.DestDir, ud)
 		}
-		if fi, err := os.Stat(ud); err != nil {
+		if _, err := os.Stat(ud); err != nil {
```
hickeyma: Probably need to refactor the code here if we want to check if that directory exists. This code originally just checked if the `destination/untardir` existed and created if it did not. For this PR we need to also account for what subdirectory will be extracted from the untar. 

For example, running command below:

```console
$ pwd
/home/usr1/test/helm-charts

$ ls -lrt

drwxr-xr-x  4 root root  4096 Nov 27 13:21 foo
[....]
drwxr-xr-x  4 root root  4096 Dec 16 11:17 chrt-7210

$ helm fetch stable/mysql --untar

$ ls -lrt

[...]
drwxr-xr-x  4 root root  4096 Dec 16 11:17 chrt-7210
drwxr-xr-x  3 root root  4096 Dec 16 12:25 mysql
```

I specify no directory for flags `--destination` or `--untardir`, it will use the current directory which is `.` in Linux and append the `<package_directory>`  which is `mysql`. Therefore need to check for the `<directory_path>/<package_directory>` because it is the `<package_directory>` which will cause the conflict within the directory path.

cmd/helm/pull_test.go
```diff
@@ -127,9 +144,27 @@ func TestPullCmd(t *testing.T) {
...
+				err := os.Mkdir(file, 0755)
+				if err != nil {
+					t.Fatal(err)
+				}
+			}
 			_, out, err := executeActionCommand(cmd)
 			if err != nil {
 				if tt.wantError {
+					if tt.wantErrorMsg != "" && tt.wantErrorMsg == err.Error() {
+						t.Fatalf("%q reported error not equel wantErr, reported: %s, wanted: %s", tt.name, err, tt.wantErrorMsg)
```
hickeyma: %q reported error not equel wantErr, reported: %s, wanted: %s --> Actual error \"%s\", not equal to expected error \"%s\"

hickeyma (changes requested):
Thanks for update @zwwhdls. Some comments around checking the package directory in addition to the destination/untar directory.

pkg/action/pull.go
zwwhdls: Got it. I will fix it soon.

zwwhdls:
ping @hickeyma

pkg/action/pull.go
```diff
@@ -109,13 +109,18 @@ func (p *Pull) Run(chartRef string) (string, error) {
 		if !filepath.IsAbs(ud) {
 			ud = filepath.Join(p.DestDir, ud)
 		}
-		if fi, err := os.Stat(ud); err != nil {
-			if err := os.MkdirAll(ud, 0755); err != nil {
+		// Let udCheck to check conflict file/dir without replacing ud when untarDir is the current directory(.).
+		udCheck := ud
+		if udCheck == "." {
+			_, udCheck = filepath.Split(chartRef)
+		}
```
hickeyma: Need an else statement similar to below,  for when path is passed with `--destination` and `--untardir` flags:

```
else {
           _, chartName := filepath.Split(chartRef)
           udCheck = filepath.Join(udCheck, chartName)
}
```

hickeyma (changes requested):
Good improvement on fix. thanks @zwwhdls.

I found a use case that fails unexpectedly though. It is for when you pass destination/untar path. Here is an example of it using dest path:

```console
$ helm pull stable/mysql -d ~/test/helm-charts/test-7182 --untar
Error: failed to untar: a file or directory with the name /usr/test/helm-charts/test-7182 already exists
```

I have added possible solution to fix this inline in the code review. It would also need a test case for using the `--destination` and ` --untardir` flags.

technosophos:
I just re-read the new version, and it looks good other than the thing that @hickeyma pointed out.

pkg/action/pull.go
zwwhdls: Thank you for your review. You are vary careful and I have added this change.

hickeyma (approved):
LGTM, thanks for your patience and work on this @zwwhdls

hickeyma:
@zwwhdls When you get a chance can you update #7188 following this merge?

PR #7188, opened 2019-12-08
[v2]stop with an error immediately if a file or directory with that name already exists

zwwhdls:
fix https://github.com/helm/helm/issues/7182

same as https://github.com/helm/helm/pull/7187

Signed-off-by: zwwhdls <zwwhdls@hotmail.com>

### DIFF
--- a/cmd/helm/fetch.go
+++ b/cmd/helm/fetch.go
@@ -166,13 +166,20 @@ func (f *fetchCmd) run() error {
 		if !filepath.IsAbs(ud) {
 			ud = filepath.Join(f.destdir, ud)
 		}
-		if _, err := os.Stat(ud); err != nil {
-			if err := os.MkdirAll(ud, 0755); err != nil {
+		// Let udCheck to check conflict file/dir without replacing ud when untarDir is the current directory(.).
+		udCheck := ud
+		if udCheck == "." {
+			_, udCheck = filepath.Split(f.chartRef)
+		} else {
+			_, chartName := filepath.Split(f.chartRef)
+			udCheck = filepath.Join(udCheck, chartName)
+		}
+		if _, err := os.Stat(udCheck); err != nil {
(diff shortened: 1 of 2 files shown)


cmd/helm/fetch.go
```diff
@@ -166,13 +166,13 @@ func (f *fetchCmd) run() error {
...
-		if fi, err := os.Stat(ud); err != nil {
+		if _, err := os.Stat(ud); err != nil {
 			if err := os.MkdirAll(ud, 0755); err != nil {
 				return fmt.Errorf("Failed to untar (mkdir): %s", err)
 			}
 
-		} else if !fi.IsDir() {
-			return fmt.Errorf("Failed to untar: %s is not a directory", ud)
+		} else {
+			return fmt.Errorf("ailed to untar: a file or directory with the name %s already exists", ud)
```
karuppiah7890: a small typo `failed`

karuppiah7890 (changes requested):

cmd/helm/fetch.go
zwwhdls: 🤣fixed.

technosophos (approved):
This might be seen as "breaking backward compatibility", but I would argue that it is a bug fix. We shouldn't let the user partially overwrite a chart because it will lead to really hard to debug problems.

Fix LGTM

karuppiah7890:
I just wanted to add this - could we check for the error message in the test? To be sure that it's the expected error message. And also have separate tests for file and directory, both with same name as chat being extracted/untarred

hickeyma (changes requested):
We should check on the clashing directory in the message as per review for Helm 3 https://github.com/helm/helm/pull/7187#pullrequestreview-330454226. Lets hold on this PR until v3 is merged first.

hickeyma (approved):
LGTM, thanks for porting this @zwwhdls

PR #7189, opened 2019-12-09
Repair unit tests failing on Linux caused by using the os.Stat

pmacik:
The `os.Stat` can return different values for directories, based on the OS. For Linux, for example, it always returns the same size of the directory (which is `4096`) regardless of the size of the contents of the directory.

Credits goes to @ldimaggi

### DIFF
--- a/pkg/chartutil/expand_test.go
+++ b/pkg/chartutil/expand_test.go
@@ -39,19 +39,6 @@ func TestExpand(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	files, err := ioutil.ReadDir(dest)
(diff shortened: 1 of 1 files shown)


technosophos (approved):
LGTM

waveywaves:
@technosophos Thanks for checking this out. Any idea when we can expect this to be merged ? 
cc @mattfarina

mattfarina:
Sorry for taking so long to get to this. Surprised it wasn't merged back in December. I've moved it to the 2.2.0 milestone which should come out prior to KubeCon/CloudNativeCon EU.

PR #7196, opened 2019-12-10
fix: backport #6361 fix to Helm 2

technosophos:
This does a similar fix to the one(s) that handle #6361 for Helm 3. This won't necessarily fix Rudder, but it will fix Tiller.

### DIFF
--- a/pkg/tiller/release_server.go
+++ b/pkg/tiller/release_server.go
@@ -270,7 +270,9 @@ func capabilities(disc discovery.DiscoveryInterface) (*chartutil.Capabilities, e
 // TODO(mattfarina): In Helm v3 merge with GetVersionSet
 func GetAllVersionSet(client discovery.ServerResourcesInterface) (chartutil.VersionSet, error) {
 	groups, resources, err := client.ServerGroupsAndResources()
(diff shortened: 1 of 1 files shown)


technosophos:
I would like to propose adding this to 2.16.2 if we can.

jdolitsky (approved):

caarlos0:
hey, sorry to dig here...

is there an ETA for this to be released?

technosophos:
We do not have a release planned for 2.x for the remainder of the year. 3.0.2, which shipped today, has a version of this fix. So for 2.x, we are probably looking at January 2020

arkadijs:
Could we please have 2.16.4 released? This issue happens on GKE every other time. Very annoying.

Alexand:
Question: Since this was labeled as "picked" on 2020-03-23, shouldn't it be present on Helm 2.16.5 released on 2020-03-26? 

I don't see it in the diff from 2.16.3: 🤔

 https://github.com/helm/helm/compare/v2.16.3...v2.16.5

mrajashree:
Backports https://github.com/helm/helm/pull/6908

PR #7201, opened 2019-12-10
fix(package): remove --set, --values, etc. flags

### LINKED ISSUE #6951: `helm package` is stripping comments from `values.yaml`
In Helm 3.0.0 final, `helm package` appears to be stripping the comments out of the values file. This happens when the values file is parsed and then re-marshaled (since the Go YAML parser does not keep comments).

To reproduce:

```
$ helm create foo
$ helm package foo
$ helm inspect values ./foo
$ helm inspect values foo-0.1.0.tgz
```

bacongobbler:
These flags snuck in through a feature that was reverted and removed in Helm 2, but snuck into Helm 3. See https://github.com/helm/helm/issues/6951#issuecomment-564261923 for more information.

This also fixes an issue where comments from values.yaml were being stripped during `helm package`, `helm create --starter` and `helm chart export`.

closes #6951 

To test:

```
./bin/helm create foo
./bin/helm package foo
rm -rf foo
tar xzf foo-0.1.0.tgz
cat foo/values.yaml
```

### DIFF
--- a/cmd/helm/package.go
+++ b/cmd/helm/package.go
@@ -118,7 +118,6 @@ func newPackageCmd(out io.Writer) *cobra.Command {
 	f.StringVar(&client.AppVersion, "app-version", "", "set the appVersion on the chart to this version")
 	f.StringVarP(&client.Destination, "destination", "d", ".", "location to write the chart.")
 	f.BoolVarP(&client.DependencyUpdate, "dependency-update", "u", false, `update dependencies from "Chart.yaml" to dir "charts/" before packaging`)
-	addValueOptionsFlags(f, valueOpts)
 
 	return cmd
 }
--- a/cmd/helm/package_test.go
+++ b/cmd/helm/package_test.go
@@ -17,13 +17,9 @@ package main
 
 import (
 	"bytes"
-	"fmt"
-	"io/ioutil"
 	"os"
 	"path/filepath"
 	"regexp"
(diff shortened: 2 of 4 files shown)


mattfarina (commented):
Why isn't the package handling in actions touched? For example...

https://github.com/helm/helm/blob/f4f5b2e46f8d32670f4cc3344d77c996b593f8fe/pkg/action/package.go#L63-L67

ch.Values is set in the `Run` method on the `Client` but that is completely ignored by `chartutil.Save`. Isn't that going to cause some confusion. The `vals` passed into `Run` is not not used but work is done with it.

bacongobbler:
> Why isn't the package handling in actions touched?

Good point. I didn't think that there was any values being handled in the action package, but you rightly pointed out one place where we happen to do some coalescing.

Looking at the code following that call to `CoalesceValues`, it doesn't look like `.Values` is ever used. We can probably go ahead and remove that. I'll do a bit more investigation later this week to see if there was a reason that call was added.

bacongobbler:
removed.

hickeyma (commented):
@bacongobbler Could you rebase to fix the conflicts?

bacongobbler:
thanks for the ping. Rebased!

technosophos (approved):
LGTM.

mattfarina (approved):

Issue #5159, opened 2019-01-13
Support for bearer token and other options in Helm client


JTG2020:
Output of `helm version`:
2.12
Output of `kubectl version`:
1.12
Cloud Provider/Platform (AKS, GKE, Minikube etc.): On Prem

I am opening this issue as suggested by @jdolitsky -[here] (https://github.com/helm/community/issues/58).
Helm should support all or most options that kubectl supports with respect to command line. Currently we could not use helm in our organization in spite of its great features because it just works on the existence of kubeconfig files for authentication. We are using multiple Kubernetes clusters and deployments happening through centralize deployment tool which has pool of servers with kubectl and helm installed. As it is centralized pool of servers, we don;t want to have any kubeconfig files on servers, instead we are using bearer token provided by OIDC for authentication. In additional we are passing necessary parameters such as master url, namespace etc to kubectl at runtime to execute deployments.
The bearer token is masked from display as shown in below example -
`kubectl --server=https://xyz.net:8443 --namespace=x-team1 --insecure-skip-tls-verify=true --token **** apply --filename=my-deployment.yaml `
Note that "=" between parameter and value works with space as well e.g.
`kubectl --server https://xyz.net:8443 --namespace x-team1 --insecure-skip-tls-verify true --token **** apply --filename my-deployment.yaml `
With Helm we can not use any of these options and hence can not use helm.
I am hoping that as tiler is being removed from architecture, helm cli should also support options same as kubectl to perform deployments by passing parameters at run time as oppose to static kubeconfig file.
If you need any more information, please let me know. I can also help with testing if needed.

Again more details on issue can be found [here in community issues](https://github.com/helm/community/issues/58)

bacongobbler:
closing as inactive.

PR #7230, opened 2019-12-14
feat(cli) Pass token via cli

### LINKED ISSUE #7292: Support passing token and api server to the cli
### Feature Request
It should be possible for us to pass the token and apiserver info to the cli to target execution of commands via tokens and pointing specific clusters or kube masters respectively, as fashioned below
`helm install stable/mariadb --kube-token 1234aasfsadf --kube-server api.openshift.com/1234/qwer:6443`

waveywaves:

- Add the ability to pass the token via a cli

fixes https://github.com/helm/helm/issues/7292

### DIFF
--- a/pkg/cli/environment.go
+++ b/pkg/cli/environment.go
@@ -46,8 +46,10 @@ type EnvSettings struct {
 	KubeConfig string
 	// KubeContext is the name of the kubeconfig context.
 	KubeContext string
-	// Bearer Token used for authentication
-	Token string
+	// Bearer KubeToken used for authentication
+	KubeToken string
+	// Kubernetes API Server Endpoint for authentication
+	KubeAPIServer string
 	// Debug indicates whether or not Helm is running in Debug mode.
 	Debug bool
 	// RegistryConfig is the path to the registry config file.
@@ -65,6 +67,8 @@ func New() *EnvSettings {
 	env := EnvSettings{
 		namespace:        os.Getenv("HELM_NAMESPACE"),
 		KubeContext:      os.Getenv("HELM_KUBECONTEXT"),
+		KubeToken:        os.Getenv("HELM_KUBETOKEN"),
+		KubeAPIServer:    os.Getenv("HELM_KUBEAPISERVER"),
 		PluginsDirectory: envOr("HELM_PLUGINS", helmpath.DataPath("plugins")),
 		RegistryConfig:   envOr("HELM_REGISTRY_CONFIG", helmpath.ConfigPath("registry.json")),
 		RepositoryConfig: envOr("HELM_REPOSITORY_CONFIG", helmpath.ConfigPath("repositories.yaml")),
@@ -79,7 +83,8 @@ func (s *EnvSettings) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVarP(&s.namespace, "namespace", "n", s.namespace, "namespace scope for this request")
 	fs.StringVar(&s.KubeConfig, "kubeconfig", "", "path to the kubeconfig file")
 	fs.StringVar(&s.KubeContext, "kube-context", s.KubeContext, "name of the kubeconfig context to use")
-	fs.StringVar(&s.Token, "token", s.Token, "bearer token used for authentication")
+	fs.StringVar(&s.KubeToken, "kube-token", s.KubeToken, "bearer token used for authentication")
+	fs.StringVar(&s.KubeAPIServer, "kube-apiserver", s.KubeAPIServer, "the address and the port for the Kubernetes API server")
 	fs.BoolVar(&s.Debug, "debug", s.Debug, "enable verbose output")
 	fs.StringVar(&s.RegistryConfig, "registry-config", s.RegistryConfig, "path to the registry config file")
 	fs.StringVar(&s.RepositoryConfig, "repository-config", s.RepositoryConfig, "path to the file containing repository names and URLs")
@@ -103,7 +108,8 @@ func (s *EnvSettings) EnvVars() map[string]string {
 		"HELM_REPOSITORY_CONFIG": s.RepositoryConfig,
 		"HELM_NAMESPACE":         s.Namespace(),
 		"HELM_KUBECONTEXT":       s.KubeContext,
-		"HELM_KUBETOKEN":         s.Token,
+		"HELM_KUBETOKEN":         s.KubeToken,
+		"HELM_KUBEAPISERVER":     s.KubeAPIServer,
 	}
 
 	if s.KubeConfig != "" {
@@ -129,7 +135,13 @@ func (s *EnvSettings) Namespace() string {
 func (s *EnvSettings) RESTClientGetter() genericclioptions.RESTClientGetter {
 	s.configOnce.Do(func() {
 		clientConfig := kube.GetConfig(s.KubeConfig, s.KubeContext, s.namespace)
-		clientConfig.BearerToken = &s.Token
+		if len(s.KubeToken) > 0 {
+			clientConfig.BearerToken = &s.KubeToken
+		}
+		if len(s.KubeAPIServer) > 0 {
+			clientConfig.APIServer = &s.KubeAPIServer
+		}
+
 		s.config = clientConfig
 	})
 	return s.config
--- a/pkg/cli/environment.go
+++ b/pkg/cli/environment.go
@@ -46,6 +46,8 @@ type EnvSettings struct {
 	KubeConfig string
 	// KubeContext is the name of the kubeconfig context.
 	KubeContext string
+	// Bearer Token used for authentication
+	Token string
 	// Debug indicates whether or not Helm is running in Debug mode.
 	Debug bool
 	// RegistryConfig is the path to the registry config file.
@@ -77,6 +79,7 @@ func (s *EnvSettings) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVarP(&s.namespace, "namespace", "n", s.namespace, "namespace scope for this request")
 	fs.StringVar(&s.KubeConfig, "kubeconfig", "", "path to the kubeconfig file")
 	fs.StringVar(&s.KubeContext, "kube-context", s.KubeContext, "name of the kubeconfig context to use")
+	fs.StringVar(&s.Token, "token", s.Token, "bearer token used for authentication")
 	fs.BoolVar(&s.Debug, "debug", s.Debug, "enable verbose output")
 	fs.StringVar(&s.RegistryConfig, "registry-config", s.RegistryConfig, "path to the registry config file")
 	fs.StringVar(&s.RepositoryConfig, "repository-config", s.RepositoryConfig, "path to the file containing repository names and URLs")
@@ -100,6 +103,7 @@ func (s *EnvSettings) EnvVars() map[string]string {
 		"HELM_REPOSITORY_CONFIG": s.RepositoryConfig,
 		"HELM_NAMESPACE":         s.Namespace(),
 		"HELM_KUBECONTEXT":       s.KubeContext,
+		"HELM_KUBETOKEN":         s.Token,
 	}
 
 	if s.KubeConfig != "" {
@@ -124,7 +128,9 @@ func (s *EnvSettings) Namespace() string {
 //RESTClientGetter gets the kubeconfig from EnvSettings
 func (s *EnvSettings) RESTClientGetter() genericclioptions.RESTClientGetter {
 	s.configOnce.Do(func() {
-		s.config = kube.GetConfig(s.KubeConfig, s.KubeContext, s.namespace)
+		clientConfig := kube.GetConfig(s.KubeConfig, s.KubeContext, s.namespace)
+		clientConfig.BearerToken = &s.Token
+		s.config = clientConfig
 	})
 	return s.config
 }
--- a/cmd/helm/load_plugins.go
+++ b/cmd/helm/load_plugins.go
@@ -127,7 +127,7 @@ func loadPlugins(baseCmd *cobra.Command, out io.Writer) {
 func manuallyProcessArgs(args []string) ([]string, []string) {
 	known := []string{}
 	unknown := []string{}
-	kvargs := []string{"--kube-context", "--namespace", "-n", "--kubeconfig", "--registry-config", "--repository-cache", "--repository-config"}
+	kvargs := []string{"--kube-context", "--namespace", "-n", "--kubeconfig", "--kube-apiserver", "--kube-token", "--registry-config", "--repository-cache", "--repository-config"}
 	knownArg := func(a string) bool {
 		for _, pre := range kvargs {
 			if strings.HasPrefix(a, pre+"=") {


marckhouzam:
This is not my area, but I was trying to figure out if this new global flag may need to be handled in the completion logic.

@waveywaves Could you describe in which use case a user would specify this flag? Am I right to assume it is to authenticate with kube cluster?

sbose78:
I would find this useful. Thank you!

How's the kubeconfig context / API Server specification going to work with this? I'm guessing that's picked up from the default kubeconfig context.

In addition, is it possible to support something similar to the following for `helm` commands as well:
```
oc get pods --token=rQ3BfJ7T1dKWGUK7NH-z2Weu4Dcr8PdCqAAFXEFpRZI --server=https://api.gamore.devcluster.openshift.com:6443
```

waveywaves:
@marckhouzam Yes it will be used to authenticate as well as allow `helm` to be used as middleware for executing commands via users tokens, without having to do multiple kubecontext changes (which is not possible in multi-tenant environments).

marckhouzam:
Thanks @waveywaves. So, I'm thinking about two aspects since this is a global flag:
1. completion
1. plugins

For completion since this flag will affect how to connect with the cluster, it needs to be added to the completion override flag list here:
https://github.com/helm/helm/blob/0fd76b2a2bca5b3febf6488c860a7c6502c70086/cmd/helm/root.go#L34

For example, if you type
```
helm --token 12345 status <TAB>
```
the completion script will run a `helm list` to figure out which releases can be listed for `helm status`.  However, to connect to the cluster `helm list` also needs the token.  By adding it to the list mentioned above, all this will be done automatically.

About plugins, should the token be made available for plugins to use?  All the other global flags have a corresponding environment variable which allows to pass the value of the flag to the plugin. See
https://github.com/helm/helm/blob/0fd76b2a2bca5b3febf6488c860a7c6502c70086/pkg/cli/environment.go#L101
https://github.com/helm/helm/blob/0fd76b2a2bca5b3febf6488c860a7c6502c70086/pkg/cli/environment.go#L64
https://github.com/helm/helm/blob/0fd76b2a2bca5b3febf6488c860a7c6502c70086/cmd/helm/load_plugins.go#L130

sbose78:
Tested this out in a pod in an OpenShift cluster. Looks good.

Verification: 
On doing a  `helm install ... --token=xyz --namespace=xyz` , I could successfully create objects which the `xyz` token had privileges to, but the `default` service account didn't.

pkg/kube/config.go
```diff
@@ -19,10 +19,11 @@ package kube // import "helm.sh/helm/v3/pkg/kube"
 import "k8s.io/cli-runtime/pkg/genericclioptions"
 
 // GetConfig returns a Kubernetes client config.
-func GetConfig(kubeconfig, context, namespace string) *genericclioptions.ConfigFlags {
+func GetConfig(kubeconfig, context, token, namespace string) *genericclioptions.ConfigFlags {
```
bacongobbler: Users consuming `kube.GetConfig` would have to change their function signatures once they upgrade. Unfortunately, this would be considered a backwards-incompatible change, so we cannot accept it as it currently stands.

One way to work around this would be to revert this change, and just make `RestClientGetter` set the token parameter directly. Would you mind making that change?

bacongobbler (changes requested):

pkg/kube/config.go
waveywaves: @bacongobbler thanks for checking this out I will make that change then.

waveywaves: @bacongobbler I have made the necessary changes to check it out

waveywaves:
@sbose78 Thank you for that, it does make more sense to have a --server option alongside token as well. Just added that to the PR as well.

waveywaves:
@marckhouzam I have added correspoding Env Vars for the `token` and `server` flags, so they can be consumed by the plugins as well. Let me know if I need to make any further changes

pkg/cli/environment.go
```diff
@@ -77,6 +83,8 @@ func (s *EnvSettings) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVarP(&s.namespace, "namespace", "n", s.namespace, "namespace scope for this request")
 	fs.StringVar(&s.KubeConfig, "kubeconfig", "", "path to the kubeconfig file")
 	fs.StringVar(&s.KubeContext, "kube-context", s.KubeContext, "name of the kubeconfig context to use")
+	fs.StringVar(&s.Token, "token", s.Token, "bearer token used for authentication")
+	fs.StringVar(&s.APIServer, "server", s.APIServer, "the address and the port for the Kubernetes API server")
```
marckhouzam: kubectl has the `-s` shortflag for `--server`.  Do we want to do the same?

bacongobbler: This would prevent other subcommands from using `-s` as a shortflag, right? If so that would block other existing shortflags like `helm create -s/--starter`

marckhouzam: Ouch! Nice save.  Thanks!

marckhouzam:
Nice patch.
No more comments from me.
I'll let another maintainer approve though.  I'm still too green 😄

karuppiah7890:
@waveywaves Could you link the issue that this PR is for? 😄

pkg/cli/environment.go
```diff
@@ -77,6 +83,8 @@ func (s *EnvSettings) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVarP(&s.namespace, "namespace", "n", s.namespace, "namespace scope for this request")
 	fs.StringVar(&s.KubeConfig, "kubeconfig", "", "path to the kubeconfig file")
 	fs.StringVar(&s.KubeContext, "kube-context", s.KubeContext, "name of the kubeconfig context to use")
+	fs.StringVar(&s.Token, "token", s.Token, "bearer token used for authentication")
```
bacongobbler: should be `kube-token` as per the existing kube config feature flag convention.

pkg/cli/environment.go
bacongobbler: Can we rename this to `kube-server` as per standard convention? e.g. `--kube-context`, `--kubeconfig`

This way we can differentiate between what comes from Helm, and what comes from Kubernetes.

bacongobbler (changes requested):

pkg/cli/environment.go
marckhouzam: Valid point.
In that case we'll need to adapt the flag when passing it to kubectl during completion.  We'll need to adapte the following method:
https://github.com/helm/helm/blob/12e0ab7e44eb02738e1128ac01ee0843e47b4e79/cmd/helm/root.go#L62
@waveywaves This may be tricky, so let me know if you need help.

pkg/cli/environment.go
marckhouzam: Valid point.
In that case we'll need to adapt the flag when passing it to kubectl during completion.  We'll need to adapt the following method:
https://github.com/helm/helm/blob/12e0ab7e44eb02738e1128ac01ee0843e47b4e79/cmd/helm/root.go#L62
@waveywaves This may be tricky, so let me know if you need help.

BTW, I'm working on making all this automatic, but it is not quite ready yet.

pkg/cli/environment.go
waveywaves: @bacongobbler @marckhouzam Thanks for reviewing this. I will make the necessary changes.

pkg/cli/environment.go
waveywaves: @bacongobbler @marckhouzam Thanks for reviewing this. I will make the necessary changes.

waveywaves:
@marckhouzam @bacongobbler I have made the necessary changes.

bacongobbler:
Hi @waveywaves, it’s the holiday season right now. Probably won’t get to reviewing this until the second week of January. Thanks for your patience.

waveywaves:
@bacongobbler No worries. Have a good one :)

bacongobbler:
closes #5159

waveywaves:
@bacongobbler can you check this out ?

cmd/helm/root.go
```diff
@@ -63,8 +63,12 @@ __helm_override_flags_to_kubectl_flags()
 {
     # --kubeconfig, -n, --namespace stay the same for kubectl
     # --kube-context becomes --context for kubectl
+	# --kube-token becomes --token for kubectl
+	# --kube-server becomes --server for kubectl
     __helm_debug "${FUNCNAME[0]}: flags to convert: $1"
     echo "$1" | \sed s/kube-context/context/
+	echo "$1" | \sed s/kube-token/token/
+	echo "$1" | \sed s/kube-server/server/
```
marckhouzam: I haven't tested it but I believe this won't work. Say the flag is --kube-server, the first echo will add it unchanged to the helm command used for completion, the second echo will add it unchanged a second time, and the third echo will add it a third time but properly renamed.

Instead you should probably do a single echo with a sed command replacing all three flags.  You can use the form
```
sed -e 's/a/b/' -e 's/c/d' ...
```

cmd/helm/root.go
```diff
@@ -31,13 +31,13 @@ import (
...
 	bashCompletionFunc = `
-__helm_override_flag_list=(--kubeconfig --kube-context --namespace -n)
+__helm_override_flag_list=(--kubeconfig --kube-context --namespace -n --kube-server --kube-token)
 __helm_override_flags()
 {
     local ${__helm_override_flag_list[*]##*-} two_word_of of var
     for w in "${words[@]}"; do
         if [ -n "${two_word_of}" ]; then
-            eval "${two_word_of##*-}=\"${two_word_of}=\${w}\""
+            eval "${two_word_of##*-}=\"$git add {two_word_of}=\${w}\""
```
marckhouzam: What is this git add? 🙂

pkg/cli/environment.go
```diff
@@ -63,6 +67,8 @@ func New() *EnvSettings {
 	env := EnvSettings{
 		namespace:        os.Getenv("HELM_NAMESPACE"),
 		KubeContext:      os.Getenv("HELM_KUBECONTEXT"),
+		KubeToken:        os.Getenv("HELM_KUBETOKEN"),
+		APIServer:        os.Getenv("HELM_APISERVER"),
```
marckhouzam: I wonder if we want the variable to be HELM_KUBEAPISERVER?

I'll let @bacongobbler decide.

marckhouzam:
@waveywaves you should be able to run `make test-acceptance` to run some regression tests which include completion tests. Or to focus on completion `make test-acceptance-completion`

marckhouzam:
Hi @waveywaves .  Good news.  With the merging of #7323 , you no longer need to make modifications for completion in this PR.  Everything should be handled automatically for completion.  Therefore, you just need to remove your completion changes and rebase on master.

Thanks.

marckhouzam:
Hi @waveywaves .  Good news.  With the merging of #7323 , you no longer need to make modifications for completion in this PR.  Everything should be handled automatically for completion.  Therefore, you just need to remove your completion changes and rebase on master.

Thanks.

waveywaves:
@marckhouzam That's good news. I will do the same.

pkg/cli/environment.go
waveywaves: I am updating to KUBEAPISERVER definitely makes more sense

bacongobbler:
This is looking great!

I want to make sure that passing an empty string for `config.KubeToken` and `config.KubeApiServer` won't accidentally pass along some extra headers, tricking certain Kubernetes api server implementations.

Marking this for 3.2 for review

01mcgrady01:
I tried 
 `helm install happy-panda stable/mariadb --kube-token xxx --kube-apiserver xxx` 
but got error
`unknown flag: --kube-apiserver, --kube-token`
Is that right way to specify k8s cluster token and api server address?

bacongobbler:
yes. Are you sure you're using a version of helm built using this PR?

01mcgrady01:
> yes. Are you sure you're using a version of helm built using this PR?

I build helm from source code of https://github.com/waveywaves/helm/tree/update/token

pkg/cli/environment.go
```diff
@@ -124,7 +134,15 @@ func (s *EnvSettings) Namespace() string {
...
 func (s *EnvSettings) RESTClientGetter() genericclioptions.RESTClientGetter {
 	s.configOnce.Do(func() {
-		s.config = kube.GetConfig(s.KubeConfig, s.KubeContext, s.namespace)
+		clientConfig := kube.GetConfig(s.KubeConfig, s.KubeContext, s.namespace)
+		if len(s.KubeToken) > 0 {
+			clientConfig.BearerToken = &s.KubeToken
+		}
+		if len(s.KubeAPIServer) > 0 {
+			clientConfig.APIServer = &s.KubeAPIServer
+		}
```
waveywaves: @bacongobbler I have updated this to have empty string checks. This should avoid any empty string security flaws the API might have.

waveywaves: > This is looking great!
> 
> I want to make sure that passing an empty string (`&""` instead of `nil`) for `clientConfig.BearerToken` and `clientConfig.ApiServer` won't accidentally pass along some extra headers, tricking certain Kubernetes api server implementations.
> 
> Marking this for 3.2 for review

@bacongobbler This should solve the empty string issue.

pkg/cli/environment.go
```diff
@@ -124,7 +134,15 @@ func (s *EnvSettings) Namespace() string {
 //RESTClientGetter gets the kubeconfig from EnvSettings
 func (s *EnvSettings) RESTClientGetter() genericclioptions.RESTClientGetter {
 	s.configOnce.Do(func() {
-		s.config = kube.GetConfig(s.KubeConfig, s.KubeContext, s.namespace)
+		clientConfig := kube.GetConfig(s.KubeConfig, s.KubeContext, s.namespace)
+		if len(s.KubeToken) > 0 {
```
bacongobbler: generally speaking, `if s.KubeToken != ""` is the more idiomatic approach. This looks good though, and we can address it in a follow-up PR.

bacongobbler (approved):

bacongobbler:
thanks!

pkg/cli/environment.go
bacongobbler: https://github.com/helm/helm/pull/7721

kkorada:
I am unable to use server and token with helm cli
e.g.
`helm install happy-panda stable/mariadb --kube-token xxx --kube-apiserver xxx`

chancejin:
Same issue with kkorada.

I tried to use helm command in my local.

1. helm list  --kubeconfig {kubeconfigFilePath} --kube-context {contextName}
>> It works fine.

2. helm list --kube-apiserver https://{API server IP}:{API server Port} --kube-token {Kubetoken}
>> It shows error like below
Error: Kubernetes cluster unreachable: Get https://10.251.205.11:6443/version?timeout=32s: x509: certificate signed by unknown authority"

Is there necessary data do I have to add to use "--kube-apiserver" without "--kubeconfig"?

bacongobbler:
Please open a new ticket.

PR #7234, opened 2019-12-15
more friendly error message when missing values.yaml

### LINKED ISSUE #7222: Helm3 warning in case of missing values.yaml
**Problem**

It's mostly my own stupidity, but maybe it can help the future users of helm. So I renamed my `values.yaml` to `values-old.yaml`, and I created a value.yaml. Then I run `helm template .` and it's throw me issues it can't find the .Values.xxx however I double check the variable many times and they must correct. As a new helm user I thought something with my layout of the value.yaml. Than I read there is a command `helm lint .` and this told me there isn't `values.yaml` at all.

**Solution**
It make sense to me adding a warning in that case, something like _Anyway we are lack of values.yaml_. 

Output of `helm version`:
```
```

Output of `kubectl version`: **not related**
```
```

Cloud Provider/Platform (AKS, GKE, Minikube etc.): **not related**

zwwhdls:
fix #7222

When missing values.yaml and no "--set" input, make the error message more friendly.


PumpkinSeed:
I prefer to not to use reflection package. In that case you check the length of the array. Reflection is always a heavy operation.

zwwhdls:
@PumpkinSeed You are right, I have changed this. Thank you.

bacongobbler:
can you please write a test for this?

zwwhdls:
Hi @bacongobbler ,

Sorry for late conflict fix. I have fetched forked registry and fixed the conflict until now.

bacongobbler:
`helm lint` already captures this case.

```
><> helm lint test-7234/
==> Linting test-7234/
[INFO] values.yaml: file does not exist
```

The values file is optional, which is why this error message is not displayed. There are valid use cases out there where the values.yaml file is not necessary, and displaying an "error message" is misleading in those cases.

`helm lint` captures this case, and that's intentional. So we are not going to accept this. Thank you for your submission.

PR #7235, opened 2019-12-15
Show errors when linting for Chart.yaml version and appVersion fields of type non-string

### LINKED ISSUE #6921: Helm3: when App Version is a commit sha, it's getting converted to scientific notation
Output of `helm version`:
`version.BuildInfo{Version:"v3.0.0-rc.3", GitCommit:"2ed206799b451830c68bff30af2a52879b8b937a", GitTreeState:"clean", GoVersion:"go1.13.4"}
`

When app version is a commit sha, it gets converted to a scientific notation. See below for more details on how to reproduce and about the issue:

```
$ helm create test-dummy

$ # change appVersion to 72225e2
$ vi test-dummy/Chart.yaml

$ cat test-dummy/Chart.yaml
apiVersion: v2
name: test-dummy

# A chart can be either an 'application' or a 'library' chart.
#
# Application charts are a collection of templates that can be packaged into versioned archives
# to be deployed.
#
# Library charts provide useful utilities or functions for the chart developer. They're included as
# a dependency of application charts to inject those utilities and functions into the rendering
# pipeline. Library charts do not define any templates and therefore cannot be deployed.
type: application

# This is the chart version. This version number should be incremented each time you make changes
# to the chart and its templates, including the app version.
version: 0.1.0

# This is the version number of the application being deployed. This version number should be
# incremented each time you make changes to the application.
appVersion: 72225e2

$ helm package test-dummy
Successfully packaged chart and saved it to: /Users/karuppiahoss/oss/helm-issues/issue-6833/dummy-repo-local/test-dummy-0.1.0.tgz

$ # check Chart.yaml contents in tar ball
$ vi test-dummy-0.1.0.tgz
```

`Chart.yaml` contents in packaged tar ball will look like this:
```
apiVersion: v2
appVersion: "7.2225e+06"
name: test-dummy
type: application
version: 0.1.0
```

And then in repository index:

```
$ helm repo index .
$ # a part of the index
$ cat index.yaml
apiVersion: v1
entries:
  test-dummy:
  - apiVersion: v2
    appVersion: "7.2225e+06"
    created: "2019-11-09T13:07:38.472282+05:30"
    digest: 626ec3a02750e500a386d80b053f6f66eef40354a45e8ad18b7ddc8426314037
    name: test-dummy
    type: application
    urls:
    - test-dummy-0.1.0.tgz
    version: 0.1.0
generated: "2019-11-09T13:07:38.471001+05:30"
```

while adding repo and doing search

```
$ # host index.yaml using https://www.npmjs.com/package/serve
$ serve
$ # in another tab
$ helm repo add local http://localhost:5000
$ helm search repo test
helm search repo test
NAME                    CHART VERSION   APP VERSION     DESCRIPTION
local/test-dummy        0.1.0           7.2225e+06      A Helm chart for Kubernetes
```

I expected the search to show `72225e2`, and that happens only when I put double quotes around `72225e2`. Found the issue when checking https://github.com/helm/helm/issues/6849 . I think the lint and package should give error when it detects numbers instead of string in `appVersion` field in the yaml. According to yaml spec, the above is a number. 

In helm, instead of erroring out for the number, the number is converted to a string somehow.

karuppiah7890:
Fixes #6921 . Fixes linting. Package fix - have to discuss

### DIFF
--- a/pkg/action/testdata/charts/multiplecharts-lint-chart-1/Chart.yaml
+++ b/pkg/action/testdata/charts/multiplecharts-lint-chart-1/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v1
 name: multiplecharts-lint-chart-1
-version: 1
+version: "1"
 icon: ""
--- a/pkg/action/testdata/charts/multiplecharts-lint-chart-2/Chart.yaml
+++ b/pkg/action/testdata/charts/multiplecharts-lint-chart-2/Chart.yaml
@@ -1,4 +1,4 @@
 apiVersion: v1
 name: multiplecharts-lint-chart-2
-version: 1
+version: "1"
 icon: ""
--- a/pkg/lint/rules/chartfile.go
+++ b/pkg/lint/rules/chartfile.go
@@ -18,12 +18,14 @@ package rules // import "helm.sh/helm/v3/pkg/lint/rules"
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 
 	"github.com/Masterminds/semver/v3"
 	"github.com/asaskevich/govalidator"
 	"github.com/pkg/errors"
+	"sigs.k8s.io/yaml"
 
 	"helm.sh/helm/v3/pkg/chart"
 	"helm.sh/helm/v3/pkg/chartutil"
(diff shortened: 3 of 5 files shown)


pkg/chartutil/chartfile.go
```diff
@@ -91,3 +91,13 @@ func IsChartDir(dirName string) (bool, error) {
 
 	return true, nil
 }
+
+func LoadChartFileForTypeCheck(filename string) (map[string]interface{}, error) {
```
mattfarina: Why is this a public function? That is, a new function to be part of the public API under the semver guarantee? Would this be better as a private function that is part of the linting of the chartfile?

If this should be a public function and part of the API it needs a comment describing when you would use it and why it would be used instead of LoadChartfile. Since the name refers to type checking someone might wonder how to do that.

On my first pass, I would do this as a private function in pkg/link/rules/chartfile.go.

mattfarina (changes requested):

pkg/chartutil/chartfile.go
karuppiah7890: Makes sense. I'll change it into a private method and also put a comment regarding what it does

karuppiah7890:
I've pushed the changes @mattfarina ! Added a small comment too for the private function, for clarity

bacongobbler (approved):
Works as intended. LGTM.

```
~/code/go/src/helm.sh/helm ><> ./bin/helm create test-7235
Creating test-7235
~/code/go/src/helm.sh/helm ><> ./bin/helm lint ./test-7235/
==> Linting ./test-7235/

~/code/go/src/helm.sh/helm ><> vim ./test-7235/Chart.yaml 
~/code/go/src/helm.sh/helm ><> ./bin/helm lint ./test-7235/
==> Linting ./test-7235/
[ERROR] Chart.yaml: version should be of type string but it's of type float64

```

Issue #6977, opened 2019-11-15
Helm v3 Error when parsing the URL for helm chart save


oneoneonepig:
### Situation

I've downloaded a chart and attempted to upload the chart to registry.
My registry is Harbor, installed in the Kubernetes cluster.
I've logged in the registry successfully with command `helm registry login -u admin --insecure 10.20.131.50:30003`
But when I tried to issue "helm chart save", it shows that the URL is incorrect
```
$ helm chart save harbor-helm/ 10.20.131.50:30003/library/harbor
Error: parse 10.20.131.50:30003/library/harbor: first path segment in URL cannot contain colon
```

while removing the port number and the colon will work

```
$ helm chart save harbor-helm/ 10.20.131.50/library/harbor:latest
ref:     10.20.131.50/library/harbor:latest
digest:  df10ae52f374f547d0f21951f72788acbdb3ae2d21c26c5e4cdb6327a7e40c48
size:    31.0 KiB
name:    harbor
version: 1.2.0-dev
latest: saved
```

but apparently that is not the correct registry endpoint, and that won't work while pushing charts to the registry.

In the [registries documentation](https://v3.helm.sh/docs/topics/registries/#the-chart-subcommand), it works perfectly with port number specified.

```
$ helm chart save mychart/ localhost:5000/myrepo/mychart:2.7.0
ref:     localhost:5000/myrepo/mychart:2.7.0
digest:  1b251d38cfe948dfc0a5745b7af5ca574ecb61e52aed10b19039db39af6e1617
size:    2.4 KiB
name:    mychart
version: 0.1.0
2.7.0: saved
```

### Version info

Output of `helm version`:
```
version.BuildInfo{Version:"v3.0.0", GitCommit:"e29ce2a54e96cd02ccfce88bee4f58bb6e2a28b6", GitTreeState:"clean", GoVersion:"go1.13.4"}
```

Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-15T19:18:23Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}
[...]

PR #7236, opened 2019-12-16
Fix issue #6977 parse url failed when the host is IP

mccxj:
Fix issue #6977 . See https://github.com/golang/go/issues/19297

### DIFF
--- a/internal/experimental/registry/reference.go
+++ b/internal/experimental/registry/reference.go
@@ -100,7 +100,7 @@ func (ref *Reference) validateRepo() error {
 	}
 	// Makes sure the repo results in a parsable URL (similar to what is done
 	// with containerd reference parsing)
-	_, err := url.Parse(ref.Repo)
+	_, err := url.Parse("//" + ref.Repo)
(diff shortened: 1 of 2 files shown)


internal/experimental/registry/reference.go
```diff
@@ -100,7 +100,7 @@ func (ref *Reference) validateRepo() error {
 	}
 	// Makes sure the repo results in a parsable URL (similar to what is done
 	// with containerd reference parsing)
-	_, err := url.Parse(ref.Repo)
+	_, err := url.Parse("//" + ref.Repo)
```
bacongobbler: can you add a comment here explaining the fix? That way the next person reading this can better understand what the `//` is supposed to do. Thanks.

bacongobbler:
ping

internal/experimental/registry/reference.go
mccxj: PTAL

jdolitsky:
apologies, closing as duplicate of #8089

Issue #6850, opened 2019-10-31
Unable to perform helm upgrade due to resource conflict


efernandes-ANDigital:
Output of `helm version`: v3.0.0-rc.1

Output of `kubectl version`: Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.3", GitCommit:"2d3c76f9091b6bec110a5e63777c332469e0cba2", GitTreeState:"clean", BuildDate:"2019-08-19T12:36:28Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"13+", GitVersion:"v1.13.10-eks-5ac0f1", GitCommit:"5ac0f1d9ab2c254ea2b0ce3534fd72932094c6e1", GitTreeState:"clean", BuildDate:"2019-08-20T22:39:46Z", GoVersion:"go1.11.13", Compiler:"gc", Platform:"linux/amd64"}

Cloud Provider/Platform (AKS, GKE, Minikube etc.): AWS

Seem to be experiencing a weird bug when doing helm upgrade. The error states "Error: UPGRADE FAILED: rendered manifests contain a new resource that already exists. Unable to continue with update: existing resource conflict: kind: ServiceMonitor, namespace: dcd, name: bid-management". 

We've tested on the following helm versions:

Helm Version:"v3.0.0-beta.2", "v3.0.0-beta.3"  

We get the following error - "Error: UPGRADE FAILED: no ServiceMonitor with the name "bid-management" found". Though I can confirm it exists.

Helm Version:"v3.0.0-rc.1", "3.0.0-beta.4", "3.0.0-beta.5" 

We get the error above "Error: UPGRADE FAILED: rendered manifests contain a new resource that already exists. Unable to continue with update: existing resource conflict: kind: ServiceMonitor, namespace: dcd, name: bid-management"

bacongobbler:
Can you provide a set of steps to reproduce the issue?

efernandes-ANDigital:
@bacongobbler Apologies for the delay. Realised it's harder to reproduce locally with minikube since we've got everything setup for/with AWS EKS. At the moment I can confirm the apiVersion of the serviceMonitor doesn't change so it doesn't seem to be a relation to [#6583](https://github.com/helm/helm/issues/6583). 

When I run helm template the first time: 

`---
# Source: k8s-service/templates/servicemonitor.yaml
apiVersion: monitoring.coreos.com/v1
[...]

PR #7237, opened 2019-12-16
fix last-deployed-release

jason-liew:


jason-liew:
[#6850 comment](https://github.com/helm/helm/issues/6850#issuecomment-566032294)

PR #7240, opened 2019-12-16
fix(helm): Validate number of arguments in install client

Blokje5:
When passing more then two arguments (e.g. due to a mistake in a flag) the following error was returned `must either provide a name or specify --generate-name`, which can be confusing. Now the following error message will be returned `expected at most two arguments, unexpected arguments: bar`, which would at least point the user towards the issue.


bacongobbler (approved):

bacongobbler:
Nice! Thank you for the fix.

PR #7242, opened 2019-12-16
feature/bugfix:(dep build) Allow dependency build to add/resolve repo

### LINKED ISSUE #7214: Dependency build without added repos
When running `helm dependency build` or `helm install --dep-up`, I need to add a repository using `helm repo add <repo_name> <url>`. As far as I can think of, this makes sense for 2 situations:

- Commands that depend on `<name>` (I.e. `helm install <release_name> <repo_name>/<chart_name>`
- Chart repositories that require specific security requirements.

However, this seems like an unnecessary step when using `helm dependency build` or `helm install --dep-up`. For example, I have a chart with the following `requirements.yaml`:

```yaml
dependencies:
  - name: postgresql
    version: 6.5.9
    repository: https://kubernetes-charts.storage.googleapis.com
    condition: postgresql.enabled
```

All information to download the chart is there. Still I need to run `helm repo add thiscouldbeanythingreally https://kubernetes-charts.storage.googleapis.com` first. On a development machine it’s no problem to add these repositories once, but it’s annoying having to run this step in CI every time.

I suggest to just use default repository settings the repository url isn’t registered. Worst case scenario Helm fails to download the dependency, which is the same as having wrong credentials defined.

Output of `helm version`: 

```go
```

waveywaves:
fixes #7214 #8036 
The main issue was due to dependency build not adding the
repo causing the command to fail. The ideal response should be
to add the repo and then continue with the command,
something that dependency update does.

The functionality has been exported and then shared between
dependency build and update as a method.
cc @hickeyma

### DIFF
--- a/cmd/helm/dependency_build_test.go
+++ b/cmd/helm/dependency_build_test.go
@@ -113,3 +113,15 @@ func TestDependencyBuildCmdWithHelmV2Hash(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+func TestDependencyBuildCmdWithUnavailableRepos(t *testing.T) {
+	chartName := "testdata/testcharts/issue-7214"
+
+	cmd := fmt.Sprintf("dependency build '%s'", chartName)
+	_, out, err := executeActionCommand(cmd)
+
+	if err != nil {
+		t.Logf("Output: %s", out)
+		t.Fatal(err)
+	}
+}
--- a/cmd/helm/testdata/testcharts/issue-7214/.helmignore
+++ b/cmd/helm/testdata/testcharts/issue-7214/.helmignore
@@ -1,0 +1,23 @@
+# Patterns to ignore when building packages.
+# This supports shell glob matching, relative path matching, and
+# negation (prefixed with !). Only one pattern per line.
+.DS_Store
+# Common VCS dirs
+.git/
+.gitignore
+.bzr/
+.bzrignore
+.hg/
+.hgignore
+.svn/
+# Common backup files
+*.swp
+*.bak
+*.tmp
+*.orig
+*~
+# Various IDEs
+.project
+.idea/
+*.tmproj
+.vscode/
--- a/cmd/helm/testdata/testcharts/issue-7214/Chart.lock
+++ b/cmd/helm/testdata/testcharts/issue-7214/Chart.lock
@@ -1,0 +1,6 @@
+dependencies:
+- name: postgresql
+  repository: https://kubernetes-charts.storage.googleapis.com
+  version: 6.5.9
+digest: sha256:bc78b47b5a97f33923d7e3f1582bc2fb246f08783cb2d004142de93b0023ea4c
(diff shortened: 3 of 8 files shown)


karuppiah7890:
There are some failing tests. Could you check them? https://circleci.com/gh/helm/helm/10734?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

hickeyma:
@waveywaves Can you fix the build issues?

waveywaves:
@hickeyma on it sorry for the long delay

waveywaves:
@hickeyma I have resolved the failing tests

waveywaves:
@hickeyma @bacongobbler can you PTAL?

hickeyma:
@waveywaves While testing #7214, it seems like it is working and doesn't need any changes. I am awaiting a reply from the OP for confirmation and will block the PR until then. Thanks for your patience.

waveywaves:
@hickeyma, thanks for taking the time to check this out. It seems like a better implementation also.

hickeyma (changes requested):
Thanks for pushing the PR @waveywaves.  Can you make changes required for `helm dep build` as described in https://github.com/helm/helm/issues/7214#issuecomment-606052189?

waveywaves:
> Thanks for pushing the PR @waveywaves.  Can you make changes required for `helm dep build` as described in https://github.com/helm/helm/issues/7214#issuecomment-606052189?

@hickeyma I will just make the necessary changes.

hickeyma (changes requested):
Following my investigations in https://github.com/helm/helm/issues/7214#issuecomment-613957411, the dependency commands are working without a need to add a repo explicitly. 

@waveywaves Lets wait until we get feedback from the OP. In the meantime, do you mind testing to see if my investigations are correct?

waveywaves:
@hickeyma I wasn't able to reproduce the issue after building with latest code from master.

hickeyma:
Closing this PR as unable to reproduce this issue in #7214. It may have been resolved in another PR.

Thanks for the effort on this @waveywaves

hickeyma:
@waveywaves Re-opened this PR in light of @akanix42 finding pattern to reproduce in https://github.com/helm/helm/issues/8036#issuecomment-627692870. Would you be interested in continuing with it? If so, do you mind rebasing against master as it has been a while since last used.

waveywaves:
@hickeyma let me try that out and rework base d on the new reproducer

waveywaves:
@hickeyma I have rebased the branch and see no issues anymore. i think we can merge this. Can you confirm ?

hickeyma:
@waveywaves We might just have to hold on this because @technosophos has unearthed some more issues around `helm dep` in #8118. You input here is welcome @technosophos also.

pniederlag:
Thx for all of you for putting your efforts into solving this problem. It's really a quite anoying behaviour when working with different repositories and a CI/CD setup. A .lock is a lock and has all information to fetch and verify the required resource. 

Our workaround is now to drop the Chart.lock from git

PR #7246, opened 2019-12-16
[helm create] Include serviceAccount.annotations value

naseemkullah:
In order to ease the use of features such as [GKE Workload Identity](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#enable_workload_identity_on_a_new_cluster), it is beneficial to allow users to annotate service accounts out of the box.

### DIFF
--- a/pkg/chartutil/create.go
+++ b/pkg/chartutil/create.go
@@ -103,6 +103,7 @@ fullnameOverride: ""
 serviceAccount:
   # Specifies whether a service account should be created
   create: true
+  annotations: {}
   # The name of the service account to use.
   # If not set and create is true, a name is generated using the fullname template
   name:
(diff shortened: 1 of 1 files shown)


naseemkullah:
bump

bacongobbler:
cc @helm/charts-maintainers.

davidkarlsen:
I think this looks sensible!

pkg/chartutil/create.go
```diff
@@ -103,6 +103,7 @@ fullnameOverride: ""
 serviceAccount:
   # Specifies whether a service account should be created
   create: true
+  annotations: {}
```
hickeyma: I think it would be helpful to add a comment to the `annotations` value

hickeyma (changes requested):
Thanks for adding this @naseemkullah. A small change needed to add a comment to the value.

pkg/chartutil/create.go
naseemkullah: True, I was not sure if it was self explanatory enough or not. Thanks, added!

naseemkullah:
My pleasure @hickeyma. The change has been applied.

hickeyma (approved):
LGTM, thanks for this @naseemkullah.

Going to merge following feedback from chart maintainer @davidkarlsen.

PR #7249, opened 2019-12-17
fix(kube) only add to scheme.Scheme once

### LINKED ISSUE #6566: consider making apiextv1beta1.AddToScheme something that can only be called once
I'm doing some work on a Kubernetes controller that deploys Helm charts to multiple remote clusters through a CRD. Regularly, I'd have this controller reconcile the Custom Resources concurrently, but there's a race condition caused by concurrent writes to scheme.Scheme.

Taking a look at `kube.New` today: 
https://github.com/helm/helm/blob/5c592dcca685b3e5887e3c241bb2a73a43191f62/pkg/kube/client.go#L62

The call to `AddToScheme`, behind the scenes appends to a slice. This means that attempting to call AddToScheme concurrently can cause a panic. 

The good news is, because that scheme.Scheme is globally available, we only need to add the CRD types once, no matter how many Kube clients we need to create. So, I'd like to propose perhaps wrapping that AddToScheme in a `sync.Once` or something of that nature.

I understand that mine is a pretty specific use case, and I'd be happy to contribute any fixes for making Helm more controller-friendly. Thanks!

jabielecki:
Multiple call to `AddToScheme` is (1) redundant and (2) destroying concurrency of `New`.

Closes #6566

Signed-off-by: Jakub Bielecki <jakub.bielecki@codilime.com>

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -57,20 +57,23 @@ type Client struct {
 	Log     func(string, ...interface{})
 }
 
+var addToScheme sync.Once
+
 // New creates a new Client.
 func New(getter genericclioptions.RESTClientGetter) *Client {
 	if getter == nil {
 		getter = genericclioptions.NewConfigFlags(true)
 	}
 	// Add CRDs to the scheme. They are missing by default.
-	if err := apiextv1.AddToScheme(scheme.Scheme); err != nil {
-		// This should never happen.
-		panic(err)
-	}
-	if err := apiextv1beta1.AddToScheme(scheme.Scheme); err != nil {
-		// This should never happen.
-		panic(err)
-	}
+	addToScheme.Do(func() {
+		if err := apiextv1.AddToScheme(scheme.Scheme); err != nil {
+			// This should never happen.
+			panic(err)
+		}
+		if err := apiextv1beta1.AddToScheme(scheme.Scheme); err != nil {
+			panic(err)
+		}
+	})
 	return &Client{
 		Factory: cmdutil.NewFactory(getter),
 		Log:     nopLogger,


bacongobbler:
Note to other maintainers: this conflicts with #7245. Once one of these have been merged, the other will have to be rebased.

jabielecki:
Whichever PR merges first, the other can be easily rebased. Nothing complicated. @bacongobbler

waveywaves:
@bacongobbler I will rebase if needed.

bacongobbler:
would you mind rebasing this now that #7245 has been merged?

jabielecki:
rebased :heavy_check_mark:

pkg/kube/client.go
```diff
@@ -57,19 +57,25 @@ type Client struct {
...
 	if getter == nil {
 		getter = genericclioptions.NewConfigFlags(true)
 	}
 	// Add CRDs to the scheme. They are missing by default.
-	if err := apiextv1.AddToScheme(scheme.Scheme); err != nil {
-		// This should never happen.
-		panic(err)
+	var errv1, errv1beta1 error
+	addToScheme.Do(func() {
+		errv1 = apiextv1.AddToScheme(scheme.Scheme)
```
bacongobbler: if we're going to `panic()`, it should be called immediately before calling another function. Can you please re-write this to the following:

```
addToScheme.Do(func() {
    if err := apiextv1.AddToScheme(scheme.Scheme); err != nil {
        // This should never happen.
        panic(err)
    }
    if err := apiextv1beta1.AddToScheme(scheme.Scheme); err != nil {
        // This should never happen.
        panic(err)
}
```

bacongobbler (changes requested):

bacongobbler (approved):

pkg/kube/client.go
jabielecki: Done that way :) :heavy_check_mark:

hickeyma:
@bacongobbler Is this a bug and should it be set for 3.0.3 milestone then?

bacongobbler:
I believe this would be considered a feature enhancement as a safety guard around calling AddToScheme multiple times in third party code. I think this should go in 3.1

hickeyma:
Thanks for confirming my thinking.

joelanford:
Thanks @bacongobbler!

PR #7254, opened 2019-12-17
Add --insecure-skip-tls-verify for repositories

### LINKED ISSUE #5434: Need support for --insecure-skip-tls-verify
Output of `helm version`: 2.13

Output of `kubectl version`: 1.13

Need support for `--insecure-skip-tls-verify`

E.g.,

`helm repo update --insecure-skip-tls-verify`

xvzf:
Fixes #5434


pkg/repo/chartrepo.go
```diff
@@ -260,3 +263,8 @@ func ResolveReferenceURL(baseURL, refURL string) (string, error) {
 	parsedBaseURL.Path = strings.TrimSuffix(parsedBaseURL.Path, "/") + "/"
 	return parsedBaseURL.ResolveReference(parsedRefURL).String(), nil
 }
+
+func (e *Entry) String() string {
```
bacongobbler: is this needed? I don't see a reference to this anywhere.

bacongobbler (changes requested):
Thank you for your contribution. Would you mind writing some tests and documentation to cover this new feature? Thanks.

pkg/getter/httpgetter.go
```diff
@@ -111,5 +112,18 @@ func (g *HTTPGetter) httpClient() (*http.Client, error) {
 
 		return client, nil
 	}
+
+	if g.opts.insecureSkipVerifyTLS {
+		client := &http.Client{
+			Transport: &http.Transport{
```
bacongobbler: I'm wondering if this should be consolidated with the `if` block above. The `http.ProxyFromEnvironment` and other settings should probably apply here as well , for example (see line 109)

pkg/repo/chartrepo.go
xvzf: Yes, it seems so - it seems like the struct now including a bool cannot be converted to a string here: https://github.com/xvzf/helm/blob/393fd5fe173540703df320e267b8d706aa9e61b9/cmd/helm/repo_add_test.go#L144

pkg/getter/httpgetter.go
xvzf: You're right - I'll come up with another way on passing the variable.

xvzf: I decided to not consolidate the two if clauses since ca cert check skipping has nothing in common with client auth. Instead I just added `http.ProxyFromEnvironment`.

xvzf:
@bacongobbler Test cases are added and I dropped a line to the docs here: https://github.com/helm/helm-www/pull/440

xvzf:
@bacongobbler Can you have a look? I think the docs for this were already merged (maybe on accident?)

xvzf:
Is there a timeline for this PR? Requested changes are implemented.

hmisbah:
any update on this ?

xvzf:
@hmisbah this is done, awaiting merging.

bacongobbler:
I’ve added this to 3.2 so it’s not lost in the noise. I’m currently away for a few days for an internal meeting. Thanks for your patience.

pkg/repo/chartrepo.go
```diff
@@ -271,3 +274,8 @@ func ResolveReferenceURL(baseURL, refURL string) (string, error) {
...
+	buf, _ := json.Marshal(e)
```
bacongobbler: We should probably `panic` here if an error occurs while marshaling. It's very unlikely to occur, but we shouldn't silently ignore the error either.

bacongobbler (changes requested):
Looks good! One more change then should be good to merge.

pkg/repo/chartrepo.go
xvzf: You are right. In case it fails there it should exit immediately. Fixed that in the latest commit!

bacongobbler (approved):
LGTM. Will need a second maintainer’s review before merging due to the size of this PR.

xvzf:
👍 Great. I can't add anotherone so let's just wait!

jdolitsky (approved):

TPX01:
Nice work! I'm testing this actual solution and the adding to repository seems to work fine but the helm pull still seems to ignore this.

```
bamboo@app-bamboo-agent-5:~$ helm pull fluxcd/flux
Error: Get https://<helm-repository>/repository/helm-fluxcd/flux-1.3.0.tgz: x509: certificate signed by unknown authority
```

```
bamboo@app-bamboo-agent-5:~$ helm repo list
NAME    URL
fluxcd  https://<helm-repository>/repository/helm-fluxcd/
```

Was this insecure feature only meant for adding a repo to helm?

xvzf:
Thanks for the feedback @TPX01
I'll have a look at it.

Which helm version are you using?

xvzf:
I was able to reproduce it; can you open an issue?

I'll address it asap!

TPX01:
> I was able to reproduce it; can you open an issue?
> 
> I'll address it asap!

Thanks alot @xvzf ! I've created the ticket: https://github.com/helm/helm/issues/7875

Le me know if that is enough information to get this going. I'm rather new to the actual code but it looks like the downloader is always using TLS veritification.

xvzf:
Should be sufficient, I’ll have to figure out where it fails since the getters are configured to respect the option..

PR #7261, opened 2019-12-18
fix(chart): lock digest differs when dependency build with Helm 2 and then Helm 3

### LINKED ISSUE #7233: Lock digest differs when performing dependency build on chart in Helm 2 and then Helm 3
When doing `helm dependency build` with the Helm v3 CLI on a chart using the Helm `v1` API, the following error occurs:

```
┌─[/tmp/test-chart]
└─▪ helm dependency build
Error: Chart.lock is out of sync with Chart.yaml
```

#### Expected Behaviour
Based on the [Helm v2 to v3 FAQ](https://v3.helm.sh/docs/faq/#consolidation-of-requirements-yaml-into-chart-yaml), Helm v3 should behave the same as Helm v2 with a chart using the Helm `v1` API.

### Steps To Reproduce

1. Create a new chart using the Helm v2 CLI

```
┌─[/tmp]
└─▪ helm2 create test-chart
Creating test-chart
```

2. Add a `requirements.yaml` to the chart and add one or more dependencies. I used the following:

```yaml
dependencies:
- name: postgresql
  version: 6.5.5
  repository: https://kubernetes-charts.storage.googleapis.com/
  condition: postgresql.enabled
```

3. Run `dependency build` using the Helm v2 and v3 CLI, observing the different outputs.
Helm v2
```
┌─[/tmp/test-chart]
└─▪ helm2 dependency build
...Unable to get an update from the "local" chart repository (http://127.0.0.1:8879/charts):
	Get http://127.0.0.1:8879/charts/index.yaml: dial tcp 127.0.0.1:8879: connect: connection refused
Update Complete.
Saving 1 charts
Downloading postgresql from repo https://kubernetes-charts.storage.googleapis.com/
Deleting outdated charts
```
Helm v3
```
┌─[/tmp/test-chart]
└─▪ helm dependency build
Error: Chart.lock is out of sync with Chart.yaml
```

### CLI Versions

```
└─▪ helm2 version
Client: &version.Version{SemVer:"v2.16.1", GitCommit:"bbdfe5e7803a12bbdf97e94cd847859890cf4050", GitTreeState:"clean"}
└─▪ helm version
```

hickeyma:
When apiVersion v1 chart dependencies are built with Helm 2 and then built with Helm 3, the lock digests differ. To avoid this issue, a dependency update is forced.

Fixes #7233


karuppiah7890:
Do you think it's possible to could add a failing test here for the issue #7233 ?

pkg/downloader/manager.go
```diff
@@ -86,6 +86,11 @@ func (m *Manager) Build() error {
 	}
 
 	if sum, err := resolver.HashReq(req, lock.Dependencies); err != nil || sum != lock.Digest {
+		// If lock digest differs and chart is apiVersion v1, it maybe because the lock was built
+		// with Helm 2 and should therefore be regenerated
+		if c.Metadata.APIVersion == chart.APIVersionV1 {
+			return m.Update()
```
TBBle: I feel like Update is the wrong thing here, because that might pull in *different* semver-matching versions of packages, ignoring the versions *locked* in the Lockfile.

Can we detect an APIVersionV1 Hash and recognise it?

technosophos: I agree that update is a "destructive" operation, and thus should not be used here. 

i am wondering if we have to add a function on the Manager that re-calculates the hashes for v2. But even there, I am a little bit uncertain.

TBBle: How hard would it be to make `resolver.HasReq(req, nil)` generate the same hash as Helm v2? Did `chart.Dependencies` change a lot?

mattfarina: The semver package is a major version change between Helm v2 and v3. Behind semver some rules were mildly modified in how they calculate versions. In order to calculate the rules the same as Helm v2 we would need to pull in semver v1 and perform the calculation using that. You can read about the changes [here](https://github.com/Masterminds/semver/blob/master/CHANGELOG.md#300-2019-09-12).

hickeyma: @mattfarina Thanks for the details about the semver.

@technosophos @TBBle FYI, I went with the "nuclear" approach first as I could be certain about the result. I am not however happy with this approach (or even doing a hashing function to map from v2 to v3) as I feel we are circumventing a valid error being raised.

I think that @technosophos suggestion in https://github.com/helm/helm/issues/7233#issuecomment-579023372 about adding a warning with a specific message might be the best approach.

I am open to ideas on how best to work this.

TBBle: @mattfarina Those changes seem to be about how the semver is matched? The Helm v2 HashReq was only looking at the pre-match expression (as a string), so I don't think it's a major issue if we accept whatever was selected at `helm dep update` time, if the has tells us that the requirements.yaml (and for v3 requirements.lock) are the same as at that time, even if `helm dep up` *now* would pick different versions.

The structure itself (`Dependency`) is unchanged, unless hashing picks up the package of the type.

I don't think this is a valid failure to raise as an error, personally. The chart *is* valid under Helm v2, a newer version of Helm shouldn't suddenly consider the chart *invalid*. We actually have the converse problem that `APIVersion1` charts produced with Helm v3 would fail this check with Helm v2, except I think the check didn't exist in Helm v2?

hickeyma: > The chart is valid under Helm v2, a newer version of Helm shouldn't suddenly consider the chart invalid.

@TBBle I appreciate your justification on why this is needed.

I have pushed changes which will now check against v2 style hash. let me know what you think.

hickeyma:
Updated and ready for review again.

TBBle (approved):

mattfarina (commented):
@hickeyma A large part of this is rather large test charts. Do we want these full charts in the codebase or would something simpler suffice?

hickeyma:
Good point @mattfarina, thanks. Its only after the commit, I realised that it had mushroomed! :)

I have refactored it now and it is ready for review again.

pkg/downloader/manager.go
```diff
@@ -86,7 +87,17 @@ func (m *Manager) Build() error {
...
+		// If lock digest differs and chart is apiVersion v1, it maybe because the lock was built
+		// with Helm 2 and therefore should be checked with Helm v2 hash
+		// Fix for: https://github.com/helm/helm/issues/7233
+		if c.Metadata.APIVersion == chart.APIVersionV1 {
+			log.Println("warning: a valid Helm v3 hash was not found. Checking against Helm v2 hash...")
+			if sum, err := resolver.HashV2Req(req); err != nil || sum != lock.Digest {
+				return errors.New("the lock file is out of sync with the dependencies file. Please update the dependencies")
+			}
+		} else {
+			return errors.New("the lock file is out of sync with the dependencies file. Please update the dependencies")
```
TBBle: The error messages could be more specific, and name the relevant files. We know whether we're an APIVersionV1 or APIVersionV2 chart at this point.

cmd/helm/dependency_build_test.go
```diff
@@ -100,3 +100,22 @@ func TestDependencyBuildCmd(t *testing.T) {
...
+	_, out, err := executeActionCommand(cmd)
+
+	// Want to make sure the build can verify Helm v2 hash
+	if err != nil {
+		t.Logf("Output: %s", out)
+		t.Fatal(err)
+	}
+
+	// Make sure the child chart file got downloaded.
+	expect := filepath.Join(chartName, "charts/alpine-0.1.0.tgz")
```
TBBle: The chart's committed. So if the download doesn't happen, this still passes? Or does this logic rely on `helm dep build` removing the existing file first?

Or is alpine-0.1.0.tgz supposed to be committed outside this Chart, so that `helm dep build` copies it in?

pkg/downloader/manager.go
hickeyma: Good point. Updated the messages to include the filenames.

cmd/helm/dependency_build_test.go
hickeyma: Hmm, good point. Removing the check as it doesn't make lot of sense as including the tarball.

TBBle (approved):

mattfarina (approved):

hickeyma:
Thanks for the reviews @mattfarina @TBBle.

@technosophos Are you ok with this to merge?

technosophos (approved):
I think this is ready to merge. Thanks for rewriting that!

hickeyma:
Thanks @technosophos for reviews.

PR #7269, opened 2019-12-19
Add back fix for CRD patch creation

adriangonz:
It seems that #5853 was re-introduced with the work for the `v3.x` release. This PR adds back the fix submitted by @mumoshu on #6092 for `v2.x`.


thomastaylor312 (approved):
Thanks for catching this and porting the fix

PR #7284, opened 2019-12-20
fix(helm): add .orig as typical backup file in .helmignore

heyleke:
Mercurial VCS (hg) backout's can generate '.orig' files
to avoid these being picked, generate a .helmignore where
also the .orig files are ignored.

Signed-off-by: Jan Heylen <jan.heylen@nokia.com>


thomastaylor312 (approved):
Thanks for the nice improvement!

Issue #4910, opened 2018-11-13
helm lint doesnt verify subcharts are well-formed


antoniogomezalvarado:
Hi,

I have the following chart structure:
```
ls -la ./main_chart

total 16
drwxr-xr-x   7 antoniogomez  staff   238 Nov 13 00:16 .
drwxr-xr-x   5 antoniogomez  staff   170 Jun  6 17:57 ..
-rw-r--r--   1 antoniogomez  staff   333 Apr 24  2018 .helmignore
-rw-r--r--   1 antoniogomez  staff    83 Apr 24  2018 Chart.yaml
drwxr-xr-x  54 antoniogomez  staff  1836 Nov 12 23:49 charts
drwxr-xr-x   6 antoniogomez  staff   204 Oct  3 10:34 roles
drwxr-xr-x   4 antoniogomez  staff   136 Apr 24  2018 templates
```

now if I run ```helm lint ./main_chart -f /tmp/input_values.yaml``` the following is the output:
```
==> Linting ./main_chart/
[INFO] Chart.yaml: icon is recommended
[INFO] values.yaml: file does not exist

1 chart(s) linted, no failures
```
Which indicates the `main_chart` is well formed. However, I put a malformed YAML template in one of the charts. If I check that specific chart ( one of the `charts` under `main_chart` dir ) with `helm lint main_chart/charts/sub_chart -f /tmp/input_values.yaml` I get the following:
```
==> Linting sub_chart
[INFO] Chart.yaml: icon is recommended
[INFO] values.yaml: file does not exist
[ERROR] templates/test.yaml: unable to parse YAML
	error converting YAML to JSON: yaml: line 27: did not find expected key
```
So i'm wondering is this by design? Helm lint only checks top level chart? Or is this a bug?

Thanks in advance dear community,

Antonio

Output of `helm version`:
Client: &version.Version{SemVer:"v2.9.0", GitCommit:"f6025bb9ee7daf9fee0026541c90a6f557a3e0bc", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.11.0", GitCommit:"2e55dbe1fdb5fdb96b75ff144a339489417b146b", GitTreeState:"clean"}
Output of `kubectl version`:
[...]

PR #7288, opened 2019-12-23
feat(cmd/lint): added flag to lint dependent charts

### LINKED ISSUE #5767: helm lint to lint also dependent charts

nlee737:
feat(cmd/lint): added flag to lint dependent charts. Supplying the flag `--with-subcharts` will also lint the dependent charts in `/charts`

Closes #5767.

Signed-off-by: Nick Lee nmkyu.lee@gmail.com

### DIFF
--- a/cmd/helm/lint.go
+++ b/cmd/helm/lint.go
@@ -19,6 +19,8 @@ package main
 import (
 	"fmt"
 	"io"
+	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/pkg/errors"
@@ -51,6 +53,21 @@ func newLintCmd(out io.Writer) *cobra.Command {
 			if len(args) > 0 {
 				paths = args
 			}
+			if client.WithSubcharts {
+				for _, p := range paths {
+					filepath.Walk(filepath.Join(p, "/charts"), func(path string, info os.FileInfo, err error) error {
+						if info != nil {
+							if info.Name() == "Chart.yaml" {
+								paths = append(paths, filepath.Dir(path))
+							} else if strings.HasSuffix(path, ".tgz") || strings.HasSuffix(path, ".tar.gz") {
+								paths = append(paths, path)
+							}
+						}
+						return nil
+					})
+				}
+			}
+
 			client.Namespace = settings.Namespace()
 			vals, err := valueOpts.MergeValues(getter.All(settings))
 			if err != nil {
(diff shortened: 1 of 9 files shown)


cmd/helm/lint.go
```diff
@@ -51,6 +53,21 @@ func newLintCmd(out io.Writer) *cobra.Command {
 			if len(args) > 0 {
 				paths = args
 			}
+			if client.WithSubcharts {
+				for _, p := range paths {
+					filepath.Walk(filepath.Join(p, "/charts"), func(path string, info os.FileInfo, err error) error {
```
bacongobbler: the slash in `/charts` can be removed. `filepath.Join` will add those as needed.

bacongobbler (approved):
found one thing that should be changed. Otherwise the code in here looks like it'll solve the use case. Thanks for writing this.

bacongobbler:
We discussed during today's dev call that we are looking at a Helm 3.1 RC in the next week or so. Are you going to have the time to fix up the issue I mentioned in that timeframe?

nick-infoway:
Thank you for the review - I've pushed a commit with your suggested changes.

mnkyl:
Thank you for the review - I've pushed a commit with your suggested changes.

hickeyma (commented):
Lets take into consideration issue #4910 as well. Is #4910 a duplicate issue of #5767 or5 do we need to extend the capability? See comment https://github.com/helm/helm/issues/4910#issuecomment-580704672

hickeyma (commented):
#7502 now closed as a duplicate of this PR. @mnkyl please consider any points raised in issue #4910 are covered in the PR.

mnkyl:
Unless we're being keen on the flag name, I think this PR should address all points brought up in #4910.

bacongobbler:
@hickeyma if you're comfortable with the change, feel free to merge this one in. I'm okay with this PR but I want to make sure your points are addressed and you're comfortable with it too.

Issue #7279, opened 2019-12-20
Moving CRDs from templates/ to crds/ results in CRDs being deleted


PhilGrayson:
Output of `helm version`:
```
version.BuildInfo{Version:"v3.0.1", GitCommit:"7c22ef9ce89e0ebeb7125ba2ebf7d421f3e82ffa", GitTreeState:"clean", GoVersion:"go1.13.4"}
```

Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"11", GitVersion:"v1.11.3", GitCommit:"a4529464e4629c21224b3d52edfe0ea91b072862", GitTreeState:"clean", BuildDate:"2018-09-09T18:02:47Z", GoVersion:"go1.10.3", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.7", GitCommit:"8fca2ec50a6133511b771a11559e24191b1aa2b4", GitTreeState:"clean", BuildDate:"2019-09-18T14:39:02Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"linux/amd64"}
```

Cloud Provider/Platform (AKS, GKE, Minikube etc.): 
N/A. Tested on Kubernetes in Docker-for-Mac. Also on on-premise Kubernetes deployments.

---

I have a chart with CustomResourceDefinitions defined in the `templates/` directory. I used the 2to3 plugin to migrate my deployments from Helm v2 to v3. I'm now trying to move the CustomResourceDefinitions from `templates/` to `crds/` as per the best practices. 

I'm not able to perform the move without Helm 3 deleting the CustomResourceDefinitions. I think this is a bug since the documentation is quite clear that Helm does not delete CRDs (https://helm.sh/docs/topics/charts/#custom-resource-definitions-crds)
>CRDs are never deleted. Deleting a CRD automatically deletes all of the CRD’s contents across all namespaces in the cluster. Consequently, Helm will not delete CRDs.

### Steps to replicate
Setup and install a minimum chart.
```
mkdir crd-issue
mkdir crd-issue/templates
mkdir crd-issue/crds

cd crd-issue

cat > Chart.yaml <<EOF
name: crd-issue
version: 0.0.1
EOF

cat > templates/crd.yaml <<EOF
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
metadata:
  name: greetings.example.com
spec:
  group: example.com
  version: v1
[...]

PR #7320, opened 2020-01-01
Do not delete templated CRDs

PhilGrayson:
Fixes issue #7279.

Prevent the deletion of CRDs that were defined in the `templates/`
directory. This makes CRD deletion behaviour consistent with Helm
documentation:

 > CRDs are never deleted. Deleting a CRD automatically deletes all of the
 > CRD’s contents across all namespaces in the cluster. Consequently, Helm
 > will not delete CRDs.

Previous the documentation only applied to CRDs that were defined in the
`crds/` directory. It did not consider that Charts could have CRDs in the
`templates/` directory (for example charts that were written before the
`crds/` directory feature or if the Chart author needed templated CRDs).

Signed-off-by: Phil Grayson <phil@philgrayson.com>

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -197,6 +197,11 @@ func (c *Client) Update(original, target ResourceList, force bool) (*Result, err
 	}
 
 	for _, info := range original.Difference(target) {
+		if info.Mapping.GroupVersionKind.Kind == "CustomResourceDefinition" {
+			c.Log("Skipping the deletion of CustomResourceDefinition %q", info.Name)
+			continue
+		}
+
 		c.Log("Deleting %q in %s...", info.Name, info.Namespace)
 		res.Deleted = append(res.Deleted, info)
 		if err := deleteResource(info); err != nil {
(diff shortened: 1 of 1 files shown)


PhilGrayson:
If there is support for this feature I'll have a go at implementing tests.

thomastaylor312:
Good catch on this @PhilGrayson! So my one concern is around people who may use CRDs in the template directory without them being the old Helm 2 style "crd-install" hooks. Deleting those if you delete the chart could be a desired behavior. I am pinging some of the other maintainers to get their opinions on it as well. 

One possible solution would be to only ignore those that have the `crd-install` hook annotation, but I wouldn't necessarily start down that path until we figure out what the best behavior would be

thomastaylor312 (approved):
@PhilGrayson Ok, after talking to a few of the other maintainers I think we are ok with this as is. You are totally correct that we have always said in the docs that we will not delete CRDs and deleting CRDs is very dangerous. We also want to follow the "principle of least surprise" for all Helm users, so that doing something like migrating CRDs doesn't accidentally result in deleted CRDs for someone.

We are going to run this by the other maintainers in the dev call this week before merging,

bacongobbler (approved):

jlegrone (approved):

thomastaylor312:
Everyone in the dev call was good with this, so we are fine with merging. Thanks again for this PR!

PhilGrayson:
That's brilliant news. Thank you for representing the PR in the call.

Issue #7304, opened 2019-12-28
bash completion:  tail: cannot open ‘+2’ for reading: No such file or directory


finkr:
When using bash completion feature, I get the error:

    tail: cannot open ‘+2’ for reading: No such file or directory

The bash completion ( https://github.com/helm/helm/pull/6730 ) has an issue because it uses "tail +2" which isn't supported by coreutils-8.22-24.el7 (RHEL/CentOS) implementation of the command tail. The tail [specification](https://pubs.opengroup.org/onlinepubs/009695399/utilities/tail.html) requires using using `-n` prefix, so use `tail -n +2`

Two lines should be fixed in [cmd/helm/root.go](https://github.com/helm/helm/blob/v3.0.2/cmd/helm/root.go#L72):

    eval $(__helm_binary_name) repo list 2>/dev/null | \tail +2 | \cut -f1
    ..
    if out=$(eval $(__helm_binary_name) plugin list 2>/dev/null | tail +2 | cut -f1); then

As a workaround, one can load bash completion this way to _patch_ the completion code:

     . <(helm completion bash| sed -e 's/tail +2/tail -n +2/g')

(ping @marckhouzam )

marckhouzam:
I was afraid some shell differences would start showing up to cause trouble.

@finkr would you be interested in posting a PR?

finkr:
Yes, i will

marckhouzam:
@finkr To be transparent: I'm working on a proposal that will completely remove the code affected.  However your PR would still be useful for the bug releases leading up to helm 3.1 and, of course, in the case my proposal is not accepted.

marckhouzam:
Thanks @finkr for reporting and fixing this bug!

marckhouzam:
I've created a PR to the https://github.com/helm/acceptance-testing repo to make sure completion continues to work as expected on centos: https://github.com/helm/acceptance-testing/pull/67

PR #7323, opened 2020-01-02
Auto-completion implemented in Go

### LINKED ISSUE #7321: Proposal: Auto-completion in Go instead of bash
The [cobra project](https://github.com/spf13/cobra) provides ways for Helm to perform custom auto-completions (e.g., `helm status <TAB>` or `helm status --namespace <TAB>`) (see Cobra's [bash_completions.md](https://github.com/spf13/cobra/blob/master/bash_completions.md) for details).

The problem is that these methods require Helm to inject shell scripts into the global auto-completion script.  So Helm developers have to code these completions in bash instead of in Go.  Furthermore, debugging these completions is **very** tedious.  Maintaining them is also complex. There are currently over 250 lines of bash script injected by Helm.  See the code starting at https://github.com/helm/helm/blob/6c604a400ec2c31cbc5d2069a34685c938a37ab9/cmd/helm/root.go#L33

This proposal aims for shell auto-completion code to be written in Go instead of bash.
 In summary, Helm would support a new hidden command `helm __complete`, which the Cobra auto-completion script would call to obtain the auto-completion choices.

The current completion flow looks as follows.
*Please open on GitHub to see diagram properly*

```
                     bash completion script
               ____________________________________  helm list -aqm 1000  ______
 ______       |  ___________        _____________  |    ----------->     |      |
|      |      | | generated |      |    helm     | |                     |      |
| User | ---> | |    by     |      |   custom    | |   helm repo list    | helm |
|      |      | |   Cobra   | ---> | completions | |    ----------->     | bin  |
 ------       | |   (bash)  |      |   (bash)    | |                     | (Go) |
              |  -----------        -------------  |  helm plugin list   |      |
               ------------------------------------     ----------->      ------
helm status <TAB>                         ^                                 |
helm repo remove <TAB>                    |                                 |
helm plugin update <TAB>                   ---------------------------------
        etc                                 output must be parsed by script
                                                       (bash)
```

The new prosed completion flow would look like this:
```
                bash completion script
              __________________________     helm __complete status      _____________
 ______      | ___________ ____________ |       --------------->        |   helm bin  |
|      |     || generated |  generic   ||                               | __complete  |
| User | --> ||    by     | use of cmd ||  helm __complete repo remove  |   cmd and   |
|      |     ||   Cobra   | __complete ||       --------------->        |  completion |
 ------      ||   (bash)  |   (bash)   ||                               |   routines  |
             | ----------- ------------ | helm __complete plugin update |     (Go)    |
              --------------------------        --------------->         -------------
helm status <TAB>                                      ^^^
helm repo remove <TAB>                      generic request for helm           
helm plugin update <TAB>                   to complete the command-line
        etc
```

Advantages of this approach:
- Custom completions written in Go instead of bash
- Access to full Helm code base to perform completions instead of having to rely on calling helm externally.  For example, calling `helm search repo` is more limited than coding the logic directly in helm in Go.
- Debugging and development **much** easier because helm developers can directly call e.g., `helm __complete status j<ENTER>` to test the corresponding completion code
- The Go debugger can even be used
- Go tests can be written for the custom completion code
- Flag parsing done automatically by Cobra (instead of having to be coded in bash by helm)
- Global flags handled automatically instead of being coded explicitly in bash (see changes required by #7230)
- No need to make sure the custom completion code works for both bash and zsh
- No more issues with the use of shell tools (e.g., `echo`, `cat`, `tail`) that work differently on different OSes (e.g., #7304)
- Use of `client-go` library instead of `kubectl`

marckhouzam:
Closes #7321 

This one aims to help the Helm  maintainers and developers.

This PR converts most of Helm's auto-completion custom bash code into Go implementations.
It is fully backwards-compatible with previous auto-completion support.  It has been tested successfully with the acceptance-testing tests for completion.

To illustrate how much easier it is to provide completion code in Go, this PR also augments the existing completion support (in Go) (the last two commits of the PR):
1. to consider the `--generate-name` flag for proper completion of `helm install` and `helm template`
1. to provide completion for the `--revision` flag (`helm status` and `helm get`).

The PR **introduces a new *hidden* helm sub-command: `__complete`** that is to be called by the auto-completion script.  Please see #7321 for details.
I chose the name with a double underscore prefix because ultimately, I would like to push this functionality into Cobra, which will need to make sure this command does not collide with any existing commands.

The support of the `__complete` command, and everything around it that enables to write auto-completions in Go, is contained within a new `completion` package.  The idea is that if this feature is eventually pushed down to Cobra, the new `completion` package can be removed almost transparently.

From a Helm user's point-of-view, nothing changes (except the new completion support for the `--generate-name` and `--revision` flags).  The real value of this PR is for Helm maintainers and developers who could now deal with shell auto-completion through Go code.

To try out the feature, set up completion exactly as before:
- check out the PR branch
- build helm
- source the script output of `helm completion bash` or `helm completion zsh`
- perform shell completion just like before:
    e.g.,
    - `helm status <TAB>`
    - `helm --namespace <TAB>`
    - `helm status -o <TAB>`
    - etc

Advantages of this approach:
- Custom completions written in Go instead of bash
- Access to full Helm code base to perform completions instead of having to rely on calling helm externally.  For example, calling `helm search repo` is more limited than coding the logic directly in helm in Go.
- Debugging and development **much** easier because helm developers can directly call e.g., `helm __complete status ngin<ENTER>` to test the corresponding completion code
- The Go debugger can even be used
- Go tests can be written for the custom completion code
- Flag parsing done automatically by Cobra (instead of having to be coded in bash by helm)
- Global flags handled automatically instead of being coded explicitly in bash (see changes required by #7230)
- No need to make sure the custom completion code works for both bash and zsh
- No more issues with the use of shell tools (e.g., `echo`, `cat`, `tail`) that work differently on different OSes (e.g., #7304)
- Use of `client-go` library instead of `kubectl`

**Questions for reviewers:**

1. No new APIs should be introduced as they are likely to be removed if this feature becomes part of Cobra.  I believe that by putting the new `completion` package under `cmd/helm`, the package is not considered API.  Should the new `completion` package be put under `internal` instead?

1. Is the new sub-command name `__complete` acceptable?


marckhouzam:
I've posted an update to the acceptance-testing repo to pass with this PR.
The change is at https://github.com/helm/acceptance-testing/pull/70

mattfarina:
Thanks for putting this together. It'll take me a few days to get through something this large. From the get to I'll respond to your two questions with my two cents...

> No new APIs should be introduced as they are likely to be removed if this feature becomes part of Cobra. I believe that by putting the new completion package under cmd/helm, the package is not considered API. Should the new completion package be put under internal instead?

I would suggest `internal` for two reasons:

1. The API can change. If in trying to move the package to cobra something needs to be changed, such as a Go API name, that should be OK to change here if it is in `internal` as it's not part of the public API.
2. Even if it's in the `cmd/helm/completion` package it becomes part of the public API we need to support through the v3 lifecycle. Even if it is merged into cobra and we can reference it there instead. We cannot get rid of it until v4. If it is in `internal` we can remove it to use the cobra one, if it lands there, because it's not part of the public API.

I think we have an easier support part if this is in `internal` for now.

> Is the new sub-command name __complete acceptable?

This seems fine to me. Are there examples of cobra itself of anyone else using the `__` to prefix an internal command?

marckhouzam:
@mattfarina Thanks for already getting involved on this.  I've moved the new `completion` package under `internal`.

>> Is the new sub-command name __complete acceptable?

> This seems fine to me. Are there examples of cobra itself of anyone else using the `__` to prefix an internal command?

Not that I'm aware of.  I'm not sure a similar situation has arisen before.  Another approach would be to name the command in whatever way helm wants and then create an alias for that name if Cobra decides to use a different name.  Personally, I think we can stick with `__complete`; after all, this command is meant for developers, not users.

internal/completion/complete.go
```diff
@@ -0,0 +1,316 @@
...
+
+	flag := findFlag(finalCmd, flagName)
+	if flag == nil {
+		// Flag not supported by this command, nothing to complete
+		CompDebugln(fmt.Sprintf("Subcommand '%s' does not support flag '%s'", finalCmd.Name(), flagName))
+		os.Exit(int(BashCompDirectiveNoFileComp))
+	}
+
+	if !flagWithEqual {
+		if _, err = finalCmd.Flags().GetBool(flag.Name); err == nil {
```
marckhouzam: This code is wrong.  I won't work for inherited flags.  So the following completion will not trigger the dynamic completion:
helm __complete status --debug ""

marckhouzam: I have forced pushed a fix for this.

marckhouzam:
I've started looking at adding go tests for this.  I now realize that calling `os.Exit` explicitly in my code will prevent tests from succeeding.

I will update the patch to use Errors instead.

marckhouzam:
This revised series of commits better supports go tests for completion.

It also adds some go tests for completion of the `status` command as well as the `--output` and `--revision` flags.  It helps illustrate one of the benefits of this PR, which is that it becomes possible to write go tests for completion.

Although more tests should eventually be added, considering the already xxl size of this patch, I don't think a review should wait for all the tests.

marckhouzam:
Test failure caused by #7391

marckhouzam:
The last forced push is a rebase to allow the tests to pass.

marckhouzam:
The last forced push is a rebase and corresponding modification to make use of the new charts cache file introduced in #6809

cmd/helm/flags.go
```diff
@@ -52,10 +53,20 @@ func addChartPathOptionsFlags(f *pflag.FlagSet, c *action.ChartPathOptions) {
...
 // value to the given format pointer
 func bindOutputFlag(cmd *cobra.Command, varRef *output.Format) {
-	cmd.Flags().VarP(newOutputValue(output.Table, varRef), outputFlag, "o",
+	f := cmd.Flags()
+	f.VarP(newOutputValue(output.Table, varRef), outputFlag, "o",
 		fmt.Sprintf("prints the output in the specified format. Allowed values: %s", strings.Join(output.Formats(), ", ")))
-	// Setup shell completion for the flag
-	cmd.MarkFlagCustom(outputFlag, "__helm_output_options")
+
+	flag := f.Lookup(outputFlag)
```
bacongobbler: instead of creating with `f.VarP` then performing a lookup, have you looked into constructing the flags before-hand?

```
oFlag := pflag.VarP(newOutputValue(output.Table, varRef), outputFlag, "o", fmt.Sprintf("prints the output in the specified format. Allowed values: %s", strings.Join(output.Formats(), ", ")))

f.AddFlag(oFlag)
completion.RegisterFlagCompletionFunc(oFlag, ...
```

cmd/helm/root.go
```diff
@@ -364,12 +101,35 @@ func newRootCmd(actionConfig *action.Configuration, out io.Writer, args []string
...
+	// Setup shell completion for the namespace flag
+	completion.RegisterFlagCompletionFunc(flag, func(cmd *cobra.Command, args []string, toComplete string) ([]string, completion.BashCompDirective) {
+		if client, err := actionConfig.KubernetesClientSet(); err == nil {
+			// Choose a long enough timeout that the user notices somethings is not working
+			// but short enough that the user is not made to wait very long
+			to := int64(3)
+			completion.CompDebugln(fmt.Sprintf("About to call kube client for namespaces with timeout of: %d", to))
+
+			nsNames := []string{}
+			// TODO can we request only the namespace with request.prefix?
```
bacongobbler: unfortunately not. You can filter using label and field selectors, but it cannot be filtered with a given prefix.

https://github.com/kubernetes/apimachinery/blob/4147c925140e4e3aa51d6b46b41650821712d457/pkg/apis/meta/v1/types.go#L320-L321

bacongobbler (approved):
The code in here looks great. So happy to have the auto-completion written in Go so we can take advantage of existing packages!

I had one comment regarding how the calling code could be improved to save a lookup call. Otherwise this all looks good to me.

cmd/helm/flags.go
marckhouzam: I assume you mean `pflag.VarPF(...)`.
That would be more efficient. I'll make the change.

Background:
This particular flag is special because it calls `pflag.VarP(...)` directly so could call `pflag.VarPF(...)` instead.

I had hesitantly chosen against it because that technique is not applicable to any of the other helm flags which call more specific methods such as `pflag.IntVar(..)` or `pflag.StringVarP(..)`, and those methods don't have a version that returns the flag.  And since the `--output` flag was the first one I coded in Go, I wanted to have it be an example of how other flags should be coded.

But that now that completion for the other flags are coded, that decision is not really as valid.

marckhouzam:
Thanks @bacongobbler for reviewing this!  I realize it was really big.

@mattfarina are you good with the single approval? I don't mind waiting if you want to have a look as well.

mattfarina (approved):

marckhouzam:
Thanks a lot @mattfarina and @bacongobbler !

Issue #7154, opened 2019-12-04
An image referencing convention


jhvhs:
We, as well as other members of the community, are facing the following issues: 
1. It is difficult or even impossible to automatically extract all the image references from a given helm chart. This is important for security, scanning and license checking purposes.
2. It is difficult for the end user to easily install an existing helm chart using images that have been re-located to a private registry. There is a large number of users who must run their workloads behind a firewall.
3. The problems above escalate with the use of sub-charts.

See below conversations within the community for more context:
- https://kubernetes.slack.com/archives/C0NH30761/p1559589151166400
- https://kubernetes.slack.com/archives/C0NH30761/p1571850216043700
- https://kubernetes.slack.com/archives/C0NH30761/p1571156640168300

## Short term solution
Since helm 3 has just gotten out, we shouldn't be pushing for a major structural change,
so our short-term proposal is to introduce a convention for referencing images via the `values.yaml`.
To simplify the problem with re-location, it would be nice if the helm charts adopted a convention for enabling the end user to set the registry and other default image settings in the `global` section. For example:
```
global:
  imageRegistry: my.registry.internal
  imageNamespace: project-xyz
  imagePullPolicy: IfNotPresent
  imagePullSecrets: []
```
where the image namespace is the part of the image path between the registry hostname and the image name, e.g. a docker user id on dockerhub.

In addition to that, the end user should be able to override the global setting on a chart or the image level by either setting the registry and namespace setting on a chart or even the image level. They should also be able to skip the defined overrides on both levels, which becomes tricky, as the templating engine is unable to restore the value was overwritten by the end user. In order to overcome the issue, we propose that the chart authors define the registry and the namespace using the `originalRegistry` and `originalNamespace` properties on the image itself, or the `originalImageRegistry` and `originalImageNamespace` properties on the chart itself.
The user will then be able to revert to the original values for certain images or even entire charts by setting the `useOriginalRegistry`/`useOriginalNamespace` boolean properties to `true` on the image or the whole chart.

This functionality can then be utilised when defining the images to be used in the chart. Unfortunately, the current template for `values.yaml` assumes that there is a one-to-one correspondence between the chart and an image, which is often not true (see https://hub.helm.sh/charts/choerodon/istio, https://hub.helm.sh/charts/aerospike/aerospike, https://hub.helm.sh/charts/bitnami/postgresql). It would be nicer if, instead of the `image` section, the `values.yaml` would have had an `images` section, that could be easily referred to in the templates, for example:
```
[...]

PR #7332, opened 2020-01-03
Support the convention for multiple images, BOM and re-location in the chart template

jhvhs:
See https://github.com/helm/helm/issues/7154 for more context

Signed-off-by: Konstantin Semenov <ksemenov@pivotal.io>

### DIFF
--- a/pkg/chartutil/create.go
+++ b/pkg/chartutil/create.go
@@ -93,20 +93,19 @@ const defaultValues = `# Default values for %s.
 # Default values for images used in the chart; all these values can be overridden in the 
 # 'images' section for individual images
 global:
-  imageRegistry: docker.io # Allows for easy re-location of all the images to a private registry
-  imagePullPolicy: IfNotPresent
+  # imageRegistry: docker.io # Allows for easy re-location of all the images to a private registry
+  # imagePullPolicy: IfNotPresent
   imagePullSecrets: []
 
 replicaCount: 1
-
 images:
   nginx:
-    name: bitnami/nginx
+    name: nginx
+    originalNamespace: bitnami
     tag: 1.16.1-debian-9-r105
     digest: sha256:582aa10676417e79995989c0e06ffa8d48ced0cc2b9883669b0379ec9a7f45fb # Optional, to ensure image authenticity
-  metrics:
-    name: bitnami/nginx-exporter
-    tag: 0.4.2-debian-9-r129
+  busybox:
+    name: busybox
 
 nameOverride: ""
 fullnameOverride: ""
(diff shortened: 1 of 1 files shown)


thomastaylor312:
This looks pretty nice. Pinging @helm/charts-maintainers to take a look at this to make sure it matches up with current best practices

jhvhs:
@thomastaylor312  I'd be more than happy to organise a pull request for the documentation website repo to update the best practices section.

prydonius:
cc @carrodher

jlegrone (commented):
The issue of how to reference images by digest has been brought up in https://github.com/helm/charts/issues/13449. I think I prefer the simplicity of that proposal over this, especially as the default "helm create" experience.

I also think Helm should provide a first-class type for image references, much like CNAB does, but that's beyond the scope of this PR. Perhaps this would make sense to trial as a library chart first?

jhvhs:
> The issue of how to reference images by digest has been brought up in [helm/charts#13449](https://github.com/helm/charts/issues/13449). I think I prefer the simplicity of that proposal over this, especially as the default "helm create" experience.

Are you suggesting that we comment out the digest, leaving it in as an optional possibility?

jlegrone:
> Are you suggesting that we comment out the digest, leaving it in as an optional possibility?

I am against specifying digest as a separate field and then appending it to the tag, eg. `my/image:tag@digest`. If we do this without building image validation directly into Helm, then users will frequently end up confused as to why updating the tag override has no effect (this will be a non-obvious problem to detect as well).

I'm also wary of adding ~100 lines of helper templates to the default Helm starter chart. I'd much rather see this complexity handled directly in Helm rather than user templates which can start from conventions but are also subject to entropy.

jhvhs:
> The more desirable behavior would be for Helm to check that a tag resolves to a digest before deploying, and _then_ update image references in rendered resources to use that digest. This ensures that if either tag OR digest change, users will know (as opposed to digest silently always taking precedence).

It would be great to have helm verify the image, but unfortunately, the fact that the helm cli  can communicate with the k8s cluster doesn't necessarily mean it can also communicate with the registry - for example, it has no knowledge of the pull secrets, or there might be a firewall in place preventing any connections apart from the clusters themselves.

The only reason we have to involve digests is that, unfortunately, tags are mutable. Referencing the image with the same tag does not guarantee that the exact same binary blob will be pulled down. Because of that, I believe that the digest should be used for stricter blob identification, while the tag should be be used as a visual version reference for the operator.

jhvhs:
> I also think Helm should provide a first-class type for image references, much like CNAB does...

We have filed a proposal https://github.com/helm/helm/issues/7754 for the first-class type support

joejulian:
Is this abandoned?

joejulian:
closing this as abaondoned

PR #7348, opened 2020-01-08
Fix a bug in storage/driver/secrets.go Delete()

tiendc:
Signed-off-by: Dao Cong Tien <tiendc@vn.fujitsu.com>

secrets.Delete calls secrets.Get not secrets.impl.Get, and secrets.Get already processes the error type, we don't need to process again.

### DIFF
--- a/pkg/storage/driver/secrets_test.go
+++ b/pkg/storage/driver/secrets_test.go
@@ -194,6 +194,12 @@ func TestSecretDelete(t *testing.T) {
 
 	secrets := newTestFixtureSecrets(t, []*rspb.Release{rel}...)
 
+	// perform the delete on a non-existing release
+	_, err := secrets.Delete("nonexistent")
+	if err != ErrReleaseNotFound {
+		t.Fatalf("Expected ErrReleaseNotFound, got: {%v}", err)
+	}
+
 	// perform the delete
 	rls, err := secrets.Delete(key)
 	if err != nil {
--- a/pkg/storage/driver/secrets.go
+++ b/pkg/storage/driver/secrets.go
@@ -185,11 +185,7 @@ func (secrets *Secrets) Update(key string, rls *rspb.Release) error {
 func (secrets *Secrets) Delete(key string) (rls *rspb.Release, err error) {
 	// fetch the release to check existence
 	if rls, err = secrets.Get(key); err != nil {
-		if apierrors.IsNotFound(err) {
-			return nil, ErrReleaseNotFound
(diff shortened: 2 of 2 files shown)


adamreese (changes requested):
Can you add the same fix to cfgmaps.go?

Cancel: just saw #7367

pkg/storage/driver/secrets_test.go
```diff
@@ -184,3 +185,30 @@ func TestSecretUpdate(t *testing.T) {
...
+	name := "smug-pigeon"
+	namespace := "default"
+	key := testKey(name, vers)
+	rel := releaseStub(name, vers, namespace, rspb.StatusDeployed)
+
+	secrets := newTestFixtureSecrets(t, []*rspb.Release{rel}...)
+
+	// perform the deletion on a non-existing release
+	_, err := secrets.Delete("nonexistent")
+	if err == nil || !strings.Contains(err.Error(), "not found") {
```
adamreese: This could be changed to the specific error.

```
if _, err := secrets.Delete("nonexistent"); err == ErrReleaseNotFound { ... }
```

pkg/storage/driver/secrets_test.go
```diff
@@ -184,3 +185,30 @@ func TestSecretUpdate(t *testing.T) {
...
+	}
+
+	// perform the deletion on an existing release
+	if _, err = secrets.Delete(key); err != nil {
+		t.Fatalf("Failed to delete release: %s", err)
+	}
+
+	// try to get the deleted one
+	_, err = secrets.Get(key)
+	if err == nil || !strings.Contains(err.Error(), "not found") {
```
adamreese: Same as above.

```
if _, err := secrets.Get(key); err == ErrReleaseExists { ... }
```

adamreese (changes requested):

bacongobbler:
ping

tiendc:
> ping

Sorry, it was Lunar new year holiday in my country.

pkg/storage/driver/secrets_test.go
tiendc: I've checked your suggestion. I didn't work as the "if" condition apierrors.IsNotFound(err) is False, so the code in the "if" block cannot be reached. The actual error came from the next line of code with value:
delete: failed to get release "nonexistent": release: not found

adamreese (approved):

adamreese (approved):

PR #7368, opened 2020-01-09
test(helm): remove usage of pointers in tests for /cmd/helm

latiif:
Before this commit, test data was modified by test code, affecting subsequent tests. Specifically, when working on a bug in the memory driver, we noticed that, under some circumstances, whatever test case – _no matter which one_ – was placed first in `/cmd/helm/rollback_test.go` would pass while the others would fail.

This was the change we made:
```diff
diff --git a/pkg/storage/driver/memory.go b/pkg/storage/driver/memory.go
index ca8756c0..9bcff55c 100644
--- a/pkg/storage/driver/memory.go
+++ b/pkg/storage/driver/memory.go
@@ -106,6 +106,9 @@ func (mem *Memory) Query(keyvals map[string]string) ([]*rspb.Release, error) {
 			return true
 		})
 	}
+	if len(ls) == 0 {
+		return nil, ErrReleaseNotFound
+	}
 	return ls, nil
 }
 
```
With this change, the order of the tests in `/cmd/helm/rollback_test.go` would matter, exposing the aforementioned problem. 

Pair-effort with @SimonAlling 

Signed-off-by: Latiif <abdullatif.alshriaf@combitech.se>

### DIFF
--- a/cmd/helm/get_all_test.go
+++ b/cmd/helm/get_all_test.go
@@ -27,12 +27,12 @@ func TestGetCmd(t *testing.T) {
 		name:   "get all with a release",
 		cmd:    "get all thomas-guide",
 		golden: "output/get-release.txt",
-		rels:   []*release.Release{release.Mock(&release.MockReleaseOptions{Name: "thomas-guide"})},
+		rels:   []release.Release{release.Mock(&release.MockReleaseOptions{Name: "thomas-guide"})},
 	}, {
 		name:   "get all with a formatted release",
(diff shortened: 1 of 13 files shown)


SimonAlling:
Context: We're working on Helm 3 support in [Kubeapps](https://github.com/kubeapps/kubeapps). We found our way into the Helm code base after discovering that the memory driver worked differently from the configmaps one, causing problems in Kubeapps tests.

This PR doesn't fix the memory driver problem. We're planning a separate PR for that, but it'll depend on this one.

latiif:
Superseded by @absoludity's #7372

PR #7372, opened 2020-01-10
Memory driver Query should return ErrReleaseNotFound for zero results

absoludity:
Alternative to #7368 , see that PR for background.

The [`Storage.History` method](https://github.com/helm/helm/blob/master/pkg/storage/storage.go#L148), which just calls `s.Driver.Query`, claims that it returns `ErrReleaseNotFound` if no such release name exists. This is the case when the driver is the `ConfigMaps` driver, the [`Query` of which explicitly returns `ErrReleaseNotFound`](https://github.com/helm/helm/blob/master/pkg/storage/driver/cfgmaps.go#L115). But the `MemoryDriver` just happily returns an empty list of releases, which is inconsistent.

This PR updates and tests the memory driver to be consistent. As mentioned on #7368, a test error then shows up in `TestRollbackRelease` [1], the cause of which is that all the test cases were using the same (mutable) data, rather than generating test data for each test (which this PR does).

[1] The reason it only shows up with this change is because `TestRollbackRelease` after running the first test successfully (which updated the test data), was then *not* erroring when it failed to find the release.

cc @latiif @SimonAlling

Signed-off-by: Michael Nelson <minelson@vmware.com>

### DIFF
--- a/pkg/action/uninstall.go
+++ b/pkg/action/uninstall.go
@@ -70,9 +70,6 @@ func (u *Uninstall) Run(name string) (*release.UninstallReleaseResponse, error) 
 	if err != nil {
 		return nil, errors.Wrapf(err, "uninstall: Release not loaded: %s", name)
 	}
-	if len(rels) < 1 {
-		return nil, errMissingRelease
-	}
 
 	releaseutil.SortByRevision(rels)
 	rel := rels[len(rels)-1]
--- a/cmd/helm/rollback_test.go
+++ b/cmd/helm/rollback_test.go
@@ -23,7 +23,7 @@ import (
 	"helm.sh/helm/v3/pkg/release"
 )
 
-func TestRollbackCmd(t *testing.T) {
+func makeTestReleases() []*release.Release {
 	rels := []*release.Release{
 		{
 			Name:    "funny-honey",
@@ -38,32 +38,35 @@ func TestRollbackCmd(t *testing.T) {
 			Version: 2,
 		},
 	}
+	return rels
+}
 
+func TestRollbackCmd(t *testing.T) {
 	tests := []cmdTestCase{{
 		name:   "rollback a release",
 		cmd:    "rollback funny-honey 1",
 		golden: "output/rollback.txt",
-		rels:   rels,
+		rels:   makeTestReleases(),
 	}, {
 		name:   "rollback a release with timeout",
 		cmd:    "rollback funny-honey 1 --timeout 120s",
 		golden: "output/rollback-timeout.txt",
-		rels:   rels,
+		rels:   makeTestReleases(),
 	}, {
 		name:   "rollback a release with wait",
 		cmd:    "rollback funny-honey 1 --wait",
 		golden: "output/rollback-wait.txt",
-		rels:   rels,
+		rels:   makeTestReleases(),
 	}, {
 		name:   "rollback a release without revision",
 		cmd:    "rollback funny-honey",
 		golden: "output/rollback-no-revision.txt",
-		rels:   rels,
+		rels:   makeTestReleases(),
 	}, {
 		name:      "rollback a release without release name",
 		cmd:       "rollback",
 		golden:    "output/rollback-no-args.txt",
-		rels:      rels,
+		rels:      makeTestReleases(),
 		wantError: true,
 	}}
 	runTestCmd(t, tests)
--- a/pkg/storage/driver/memory.go
+++ b/pkg/storage/driver/memory.go
@@ -106,6 +106,9 @@ func (mem *Memory) Query(keyvals map[string]string) ([]*rspb.Release, error) {
 			return true
 		})
 	}
+	if len(ls) == 0 {
+		return nil, ErrReleaseNotFound
+	}
 	return ls, nil
 }
 
--- a/pkg/storage/driver/memory_test.go
+++ b/pkg/storage/driver/memory_test.go
@@ -83,22 +83,30 @@ func TestMemoryGet(t *testing.T) {
 
 func TestMemoryQuery(t *testing.T) {
 	var tests = []struct {
-		desc string
-		xlen int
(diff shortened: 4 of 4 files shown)


SimonAlling:
Nice job! We think this check can also be removed, since it's made obsolete by your changes to the `Memory` driver (and it returns the wrong error anyway):

```diff
diff --git a/pkg/action/uninstall.go b/pkg/action/uninstall.go
index dfaa9847..87098c5b 100644
--- a/pkg/action/uninstall.go
+++ b/pkg/action/uninstall.go
@@ -70,9 +70,6 @@ func (u *Uninstall) Run(name string) (*release.UninstallReleaseResponse, error)
        if err != nil {
                return nil, errors.Wrapf(err, "uninstall: Release not loaded: %s", name)
        }
-       if len(rels) < 1 {
-               return nil, errMissingRelease
-       }
 
        releaseutil.SortByRevision(rels)
        rel := rels[len(rels)-1]
```

absoludity:
> Nice job! We think this check can also be removed, since it's made obsolete by your changes to the `Memory` driver (and it returns the wrong error anyway):
> 
> ```diff
> diff --git a/pkg/action/uninstall.go b/pkg/action/uninstall.go
> index dfaa9847..87098c5b 100644
> --- a/pkg/action/uninstall.go
> +++ b/pkg/action/uninstall.go
> @@ -70,9 +70,6 @@ func (u *Uninstall) Run(name string) (*release.UninstallReleaseResponse, error)
>         if err != nil {
>                 return nil, errors.Wrapf(err, "uninstall: Release not loaded: %s", name)
>         }
> -       if len(rels) < 1 {
> -               return nil, errMissingRelease
> -       }
>  
>         releaseutil.SortByRevision(rels)
>         rel := rels[len(rels)-1]
> ```

Yes - indeed, that was where we started on this yak-shaving experience, wasn't it :) Updating... Thanks!

latiif (approved):
Nice work.
Applying the changes in this PR, resulted in the same expected behaviour of both drivers `MemoryDriver` and `ConfigMapsDriver`

pkg/action/uninstall.go
```diff
@@ -70,9 +70,6 @@ func (u *Uninstall) Run(name string) (*release.UninstallReleaseResponse, error)
 	if err != nil {
 		return nil, errors.Wrapf(err, "uninstall: Release not loaded: %s", name)
 	}
-	if len(rels) < 1 {
-		return nil, errMissingRelease
```
bacongobbler: how does this removal affect the other storage drivers (secrets and configmaps)?

I'm concerned removing this will cause us to rely on the storage layer for this functionality, and there's no standard in the interface to capture the case when no releases were found.

absoludity: > how does this removal affect the other storage drivers (secrets and configmaps)?

Both the secrets and configmap drivers' implementation of `Query` already return `ErrReleaseNotFound` (see [secret driver](https://github.com/helm/helm/blob/release-3.1/pkg/storage/driver/secrets.go#L124) and [configmap driver](https://github.com/helm/helm/blob/release-3.1/pkg/storage/driver/cfgmaps.go#L133) respectively for the returned error) as is documented one level above in the [`History` storage](https://github.com/helm/helm/blob/release-3.1/pkg/storage/storage.go#L149) method, and so will already have returned two lines above.

> I'm concerned removing this will cause us to rely on the storage layer for this functionality, and there's no standard in the interface to capture the case when no releases were found.

Yep, I agree - but in that case, the places mentioned above in ConfigMap and Secret drivers which already return the error should be removed, `storage.History` updated to check and return the error explicitly, and this error here should be updated to have the correct type for consistency (`ErrReleaseNotFound` rather than `errMissingRelease`), I think? Our issue was simply that our test code using the `Memory` had different expectations of which error was returned compared to the production `Secret` driver.

mikeshng:
Hi @absoludity I agree with the memory driver change. But the sql driver deserves a second look.

Maybe the check for sql driver should change from
```
	if len(releases) == 0 {
		return nil, ErrReleaseNotFound
```
to
```
	if len(records) == 0 {
		return nil, ErrReleaseNotFound
```
so that it's consistent with secret and cfgmap.

I think if the release exists, but all the decoding fails and its going to return an empty array then that should be a new error (not ErrReleaseNotFound) but something that tells the user, release exists but it fails to decode it. Maybe the comment for History can mention the fact that empty return doesn't mean the release doesn't exists. What do you think?

absoludity:
> Hi @absoludity I agree with the memory driver change. But the sql driver deserves a second look.
> 
> Maybe the check for sql driver should change from
> 
> ```
> 	if len(releases) == 0 {
> 		return nil, ErrReleaseNotFound
> ```
> 
> to
> 
> ```
> 	if len(records) == 0 {
> 		return nil, ErrReleaseNotFound
> ```
> 
> so that it's consistent with secret and cfgmap.

+1, I agree with your #8470 :)

> 
> I think if the release exists, but all the decoding fails and its going to return an empty array then that should be a new error (not ErrReleaseNotFound) but something that tells the user, release exists but it fails to decode it. Maybe the comment for History can mention the fact that empty return doesn't mean the release doesn't exists. What do you think?

Makes sense, yes. I'm not sure this PR is going to get any more attention. Perhaps it's worth doing the two smaller changes in isolation that I've got here (the memory driver, and the update action, as the update action change is the one which worried Matt, I think).

hickeyma (commented):
@absoludity Feedback in https://github.com/helm/helm/issues/8458#issuecomment-664969792 to help progress this.

absoludity:
Closing this, since the main change to update the memory driver landed as part of https://github.com/helm/helm/commit/1ab52fa79c100332bc8014095cf7aed6937cae8a , as mentioned by @mikeshng .

hickeyma:
Thanks for your effort on this @absoludity. This will feed into improving consistency in the driver area.

PR #7373, opened 2020-01-10
add option to bypass kubeconfig namespace

### LINKED ISSUE #7135: Namespace flag of the Install/Upgrade action not working correctly on the sdk

### LINKED ISSUE #7377: Over engineering when using Helm SDK without a kubeconfig
Currently when using helm SDK, a `RESTClientGetter` is needed. This is trivial when a kubeconfig file is available. However when the program only have access to the CA, token/certs and url in memory, it becomes a bit more complicated. 

In fact the program needs to implement the following interfaces:
- `ToRESTConfig() (*rest.Config, error)`
- `ToDiscoveryClient() (discovery.CachedDiscoveryInterface, error)`
- `ToRESTMapper() (meta.RESTMapper, error)`

Of course it can still be done with some magic, but then it is also needed to create a new abstraction over the namespace with the `ToRawKubeConfigLoader()` method. 

A fix would be to add a `Namespace` field to the kube client in order to lighten a bit the custom implementation written to use Helm3 without a kubeconfig

Sh4d1:
Fixes #7377  (kind of)
This PR allows to bypass the kubeconfig namespace (or even not implement some RESTClientGetter methods)

Signed-off-by: Patrik Cyvoct <patrik@ptrk.io>

cc @hickeyma @bacongobbler

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -55,6 +55,8 @@ var ErrNoObjectsVisited = errors.New("no objects visited")
 type Client struct {
 	Factory Factory
 	Log     func(string, ...interface{})
+	// Namespace allows to bypass the kubeconfig file for the choice of the namespace
+	Namespace string
 }
 
 // New creates a new Client.
@@ -113,6 +115,9 @@ func (c *Client) Wait(resources ResourceList, timeout time.Duration) error {
 }
 
 func (c *Client) namespace() string {
+	if c.Namespace != "" {
+		return c.Namespace
+	}
 	if ns, _, err := c.Factory.ToRawKubeConfigLoader().Namespace(); err == nil {
 		return ns
 	}


hickeyma:
@Sh4d1 If this supersedes #7122, can you close #7122? If this doesn't fully fix #7135, can you update the issue to better explain what is being fixed and what are the limitations in the fix?

Sh4d1:
@hickeyma maybe I'm better off creating a new issue? This PR allows a fix to bypass #7135 but doesnt really fixes it :thinking:

hickeyma:
@Sh4d1 It would be good to better understand for users, what this PR fixs.

Sh4d1:
@hickeyma #7377 :)

Sh4d1:
@hickeyma any news on this one?

bacongobbler (approved):

PR #7374, opened 2020-01-10
ref(pkg/storage): Improve error handling

SimonAlling:
* The error returned from `DeployedAll` [will never contain `"not found"`](https://github.com/helm/helm/blob/f35719a62a35bb0673c5dda2d6c6ba231b685dfe/pkg/storage/storage.go#L141:L143).
* The error returned at the end of `Deployed` is already known to be `nil`, and we never want to return `ls[0]` together with a non-`nil` error anyway.


hickeyma (approved):
LGTM, thanks @SimonAlling

PR #4858, opened 2018-10-30
ref: require name by default on 'helm install'

technosophos:
This is described in the official Helm 3 proposal: https://github.com/helm/community/blob/master/helm-v3/000-helm-v3.md

The new version of `helm install` is now:

```
$ helm install NAME CHART
```

Moniker (a.k.a. "the cute namer") is gone 😢 🐼 , and the default naming scheme for `--generate-name` is now `basename(CHART)-unix_timestamp()`

PR #7380, opened 2020-01-11
feat(helm): reintroduce name flag

### LINKED ISSUE #7345: install: reintroduce name long flag (--name) for v2 compatibility
## Summary 

To support backwards compatibility, helm should honor `name` passed in as both an argument or a long option.  

### What happened

I ran the following command:
```
helm install stable/kubernetes-dashboard \
  --name helm-dashboard \
  --set ingress.enabled="false" \
  --set enableSkipLogin="true
```

And I got the error 

`Error: unknown flag: --name`

### What I expected to happen

I expected helm to recognize that a name _has_ been provided. This command (`--name`) worked in Helm v2. 

I understand and agree with the change from v2 --> v3 such that helm name's should be explicitly set or generated but this feels like an unnecessary breaking change.

Helm is getting name explicitly passed to it.

Output of `helm version`:
```
```

Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.8", GitCommit:"211047e9a1922595eaa3a1127ed365e9299a6c23", GitTreeState:"clean", BuildDate:"2019-10-15T12:11:03Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.8", GitCommit:"211047e9a1922595eaa3a1127ed365e9299a6c23", GitTreeState:"clean", BuildDate:"2019-10-15T12:02:12Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}```
```

I am willing to help contribute a fix to this issue.

desimone:
These changes reintroduce `--name` as an accepted long flag. 

Fixes https://github.com/helm/helm/issues/7345

### DIFF
--- a/cmd/helm/install.go
+++ b/cmd/helm/install.go
@@ -142,6 +142,7 @@ func addInstallFlags(f *pflag.FlagSet, client *action.Install, valueOpts *values
 	f.BoolVar(&client.Atomic, "atomic", false, "if set, installation process purges chart on fail. The --wait flag will be set automatically if --atomic is used")
 	f.BoolVar(&client.SkipCRDs, "skip-crds", false, "if set, no CRDs will be installed. By default, CRDs are installed if not already present")
 	f.BoolVar(&client.SubNotes, "render-subchart-notes", false, "if set, render subchart notes along with the parent")
+	f.StringVarP(&client.ReleaseName, "name", "", "", "The release name (and omit NAME parameter)")
 	addValueOptionsFlags(f, valueOpts)
 	addChartPathOptionsFlags(f, &client.ChartPathOptions)
 }
--- a/cmd/helm/install.go
+++ b/cmd/helm/install.go
@@ -142,7 +142,7 @@ func addInstallFlags(f *pflag.FlagSet, client *action.Install, valueOpts *values
 	f.BoolVar(&client.Atomic, "atomic", false, "if set, installation process purges chart on fail. The --wait flag will be set automatically if --atomic is used")
 	f.BoolVar(&client.SkipCRDs, "skip-crds", false, "if set, no CRDs will be installed. By default, CRDs are installed if not already present")
 	f.BoolVar(&client.SubNotes, "render-subchart-notes", false, "if set, render subchart notes along with the parent")
-	f.StringVarP(&client.ReleaseName, "name", "", "", "The release name (and omit NAME parameter)")
+	f.StringVarP(&client.ReleaseName, "name", "", "", "the release name (and omit NAME parameter)")
 	addValueOptionsFlags(f, valueOpts)
 	addChartPathOptionsFlags(f, &client.ChartPathOptions)
 }
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -553,7 +553,7 @@ func (i *Install) NameAndChart(args []string) (string, string, error) {
 			return errors.New("cannot set --name-template and also specify a name")
 		}
 		if i.ReleaseName != "" {
-			return errors.New("cannot set --name and also specify specify a name argument")
+			return errors.New("cannot set --name and also specify a name argument")
 		}
 		return nil
 	}
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -552,6 +552,9 @@ func (i *Install) NameAndChart(args []string) (string, string, error) {
 		if i.NameTemplate != "" {
 			return errors.New("cannot set --name-template and also specify a name")
 		}
+		if i.ReleaseName != "" {
+			return errors.New("cannot set --name and also specify specify a name argument")
+		}
 		return nil
 	}
 
--- a/cmd/helm/install_test.go
+++ b/cmd/helm/install_test.go
@@ -189,7 +189,17 @@ func TestInstall(t *testing.T) {
 			cmd:    "install aeneas testdata/testcharts/deprecated --namespace default",
 			golden: "output/deprecated-chart.txt",
 		},
+		// Install using v2 name flags (--name)
+		{
+			name:   "basic install using name flag no arguments",
+			cmd:    "install --name aeneas --namespace default testdata/testcharts/empty",
+			golden: "output/install.txt",
+		},
+		{
+			name:      "basic install using name flag and name argument no arguments",
+			cmd:       "install --name aeneas --namespace default aeneas testdata/testcharts/empty",
+			wantError: true,
+		},
 	}
-
 	runTestActionCmd(t, tests)
 }


cmd/helm/install.go
```diff
@@ -142,6 +142,7 @@ func addInstallFlags(f *pflag.FlagSet, client *action.Install, valueOpts *values
 	f.BoolVar(&client.Atomic, "atomic", false, "if set, installation process purges chart on fail. The --wait flag will be set automatically if --atomic is used")
 	f.BoolVar(&client.SkipCRDs, "skip-crds", false, "if set, no CRDs will be installed. By default, CRDs are installed if not already present")
 	f.BoolVar(&client.SubNotes, "render-subchart-notes", false, "if set, render subchart notes along with the parent")
+	f.StringVar(&client.NameFlag, "name", "", "The release name (and omit NAME parameter)")
```
desimone: I believe v2 also had a short flag `-n` set. so this should probably be:

```go
	f.StringVarP(&client.NameFlag, "name", "n", "", "The release name (and omit NAME parameter)")
```

```log
panic: unable to redefine 'n' shorthand in "install" flagset: it's already used for "name" flag [recovered]
	panic: unable to redefine 'n' shorthand in "install" flagset: it's already used for "name" flag
....
....
	/Users/bdd/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:895 +0x60c
helm.sh/helm/v3/cmd/helm.executeActionCommandC(0xc0001070a0, 0x2652554, 0x3c, 0x1062fa0, 0xc0004786f8, 0x56df0838, 0xc000461000, 0xc000478708)
	/Users/bdd/helm/cmd/helm/helm_test.go:116 +0x1f8
helm.sh/helm/v3/cmd/helm.runTestActionCmd.func1(0xc000738300)
	/Users/bdd/helm/cmd/helm/helm_test.go:83 +0x147
testing.tRunner(0xc000738300, 0xc0002e4400)
```
Unfortunately, this results in a panic in the test setup.

pkg/action/install.go
```diff
@@ -552,6 +552,9 @@ func (i *Install) NameAndChart(args []string) (string, string, error) {
 		if i.NameTemplate != "" {
 			return errors.New("cannot set --name-template and also specify a name")
 		}
+		if i.ReleaseName != "" {
+			return errors.New("cannot set --name and also specify specify a name argument")
```
marckhouzam: Double 'specify'

bacongobbler:
LGTM pending @marckhouzam's comment

cmd/helm/install.go
```diff
@@ -142,6 +142,7 @@ func addInstallFlags(f *pflag.FlagSet, client *action.Install, valueOpts *values
 	f.BoolVar(&client.Atomic, "atomic", false, "if set, installation process purges chart on fail. The --wait flag will be set automatically if --atomic is used")
 	f.BoolVar(&client.SkipCRDs, "skip-crds", false, "if set, no CRDs will be installed. By default, CRDs are installed if not already present")
 	f.BoolVar(&client.SubNotes, "render-subchart-notes", false, "if set, render subchart notes along with the parent")
+	f.StringVarP(&client.ReleaseName, "name", "", "", "The release name (and omit NAME parameter)")
```
bacongobbler: can you lowercase the capital T at the beginning of the flag description? That way it's consistent with the rest of the flag descriptions.

bacongobbler (changes requested):

cmd/helm/install.go
desimone: Yes, good catch.

pkg/action/install.go
desimone: Thanks, good catch :)

desimone:
@bacongobbler / @marckhouzam 
Thanks for the review. Let me know if you'd like me to rebase.

marckhouzam:
I expressed a concern in the corresponding issue.

mattfarina (changes requested):
We talked about this at length in the Helm developer call this week. Several of the Helm maintainers shared a concern with bringing this in as it sets a precedent and did not want it merged. Note, this is around setting precedents and justifications. It has nothing to do with the quality of this change.

Helm v3 is a major version change and numerous CLI names, arguments, and flags changed. This is a change for everyone (I'm still getting used to a lack of automatic name generation). How much work do we put into adding more compatibility to v2 and why?

The original issue gave the reason for the change as, "this feels like an unnecessary breaking change". Can we get something more as a justification for this? Something we can use in the decision making process.

For example, it feels more difficult that the automatic generation of names no longer happens. Does this mean we should detect when no name is passed and auto-generate it? Or, what about the case with purging and keeping history on delete. That change affects scripts that call it. Should we let it work both ways and try to infer meaning based on the flags passed in? I would suggest not.

@bacongobbler since you approved, do you have any comments on this?

Before merging (if merged) we should have more discussion.

bacongobbler:
After further discussion and consideration, I agree with the other maintainers' point of view about this setting a new precedent. This would also cause confusion amongst newer community members... "do I use `<NAME>` - the positional argument - or `--name`? How do they relate?".

I too would like to like to hear a more persuasive argument. We removed `--name` in Helm 3 because it was changed from being an optional argument to a mandatory argument.

desimone:
> We talked about this at length in the Helm developer call this week. Several of the Helm maintainers shared a concern with bringing this in as it sets a precedent and did not want it merged. Note, this is around setting precedents and justifications. It has nothing to do with the quality of this change.

That seems reasonable to me. 

However, I guess to me there are two things going on here:

1) If and when (not necessarily this issue) Helm decides that they made a mistake in breaking backwards compatibility what will the approach be? Is the precedent to stick to breaking changes no matter what because that's what a major version means? If that's not the case, what does the bar for that look like for the reintroduction of backwards compatibility? 
2) Does this particular compatibility patch hit that water mark?

> How much work do we put into adding more compatibility to v2 and why?

I think as with most things, the answer is it depends. Reasonable minds can disagree about what compatibility issues are best avoided based on some sliding scale of broad categories like existing use, difficulty in implementation, technical debt, and user confusion.

> Can we get something more as a justification for this? Something we can use in the decision making process.

For this particular change, "this feels like an unnecessary breaking change" because on the above matrix of tradeoffs, look _to me_ like:

- How prevalent was usage? 5/5 (of course there were many other breaking changes in the release, but I'm hard pressed to find an argument like name that breaks almost every single existing tutorial) 
- How painful is lack of backward compatibility? 1/5 (It's a google search away, and an easy fix. If this PR is deemed not appropriate, I would still prefer a deprecation message for at least a release)
- How much additional technical debt is there to maintain? 1/5 (Assuming my code is correct which is a big if, it's a very small patch)
- User confusion? 1/5 (I get why other folks may feel differently, but I personally don't think having both is confusing, especially when documented and when the code path asserts the correct behavior). 

For continuing to support `--name` seemed like a good tradeoff and a net win to me. Again, I respect that reasonable minds can disagree here, and elsewhere. 

> For example, it feels more difficult that the automatic generation of names no longer happens. Does this mean we should detect when no name is passed and auto-generate it? Or, what about the case with purging and keeping history on delete. That change affects scripts that call it. Should we let it work both ways and try to infer meaning based on the flags passed in? I would suggest not.

As to whether all those features have an equally attractive set of tradeoffs, I cannot say, but I would say each decision should be treated individually and judged according to some common set of criteria.

mattfarina:
@desimone This is the kind of justification I like to see. Thank you for that.

I started to look at the charts listed in the stable repo and in the Helm Hub. In the charts repo we maintainer on the Helm project `--name` shows up hundreds of times. It's very common to use that to specify a name in the examples install command.

This isn't just in the charts from the Helm charts repo. As I look over charts provided by other repositories to the Helm Hub I find many other use `--name` as part of their install commands.

To avoid confusion in the installation process for people copying and pasting commands We should either:

1. Restore the `--name` flag
2. Update the charts in the charts repo to stop using it and reach out to the other repo maintainers we can find and ask them to change the usage there as well

@bacongobbler @marckhouzam @hickeyma @bridgetkromhout Any thoughts?

marckhouzam:
> In the charts repo we maintain on the Helm project --name shows up hundreds of times

That's a strong motivator.

I assume fixing all the charts ASAP is a large endeavor.

So let's see how to best try to ease the pain points.  If adding the flag back, having it deprecated seems right, as suggested by @mattfarina.

We probably want to do our best to avoid people using it, so how about also making it a hidden flag? 

I assume that (if approved) since the argument to add the flag back is to avoid too much pain on our users, it implies it should be added in the the next patch release?

hickeyma:
Firstly, thanks @desimone for the great feedback in https://github.com/helm/helm/pull/7380#issuecomment-577825261. It helps to get a better understanding of the issue.

I looked back into why `--name` was removed and it goes back to the design proposal: https://github.com/helm/community/blob/master/helm-v3/000-helm-v3.md#commandflag-differences-from-helm-2 and was changed in PR #4858.

The decision makes sense as it would be expected that most if not all users would specify a name (especially in production) rather than the default name moniker. 

The decision to have name as a positional argument rather than a specified flag argument, can be argued both ways. CLIs use both approaches. However, if the Helm CLI is striding to be more like `kubectl` then positional is the standard, for example:
` kubectl create job NAME ......`

Even though Helm's policy is to allow breaking changes between major releases, maybe we still need to review these breaking changes on a per change basis to understand the impact it can have to the users of Helm. This is probably one such case where reverting the change is feasible without major impact,  and as specified by @mattfarina in  https://github.com/helm/helm/pull/7380#issuecomment-578287521 would be beneficial to maintain usability with charts.

I would agree with @marckhouzam about adding it back in as a hidden flag with a deprecation warning. Should the `--name` flag then override the positional field (i.e. if flag used then NAME not required)?

bacongobbler:
> Should the `--name` flag then override the positional field (i.e. if flag used then NAME not required)?

Yeah. This PR contains code to check for that case. It will return with an error saying "cannot set --name and also specify a name argument".

Marking it as a hidden flag would solve the ambiguity. Really we should be pushing people to use the positional argument than the flag, but I'm okay with both.

My final concern about this PR is in regards to the SDK. `NameAndChart` is commonly used to parse the command line flags as well as the parameters to `action.Install` to determine the release name.

Before this change, this code was OK:

```
cfg := action.Configuration{...}
client := action.NewInstall(cfg)
client.ReleaseName = "myrelease"
args = []string{"mychart", "./mychart"}
name, chart, err := client.NameAndChart(args)
if err != nil {
    log.Fatal(err)
}
```

With this change, however, `err` will be set to "cannot set --name and also specify a name argument".

Is that considered a breaking change to the API, or is that a bug fix? I don't have enough evidence to suggest that this may break someone's code, but it is a behavioral difference from v3.0.

mattfarina:
@desimone I think we can move forward with restoring `--name` with a few changes needed to this PR:

1. Make `--name` a hidden flag
2. Fix the code issue @bacongobbler highlighted
3. Add a deprecation notice about the `--name` flag

Thank you for making your case on this.

desimone:
Thank for your reconsideration and feedback.

1. By hidden flag, is the intention to hide`--name` from the cli help message? I understand the intent of wanting to push users to not use the flag, but I would argue against it for the following reasons :
  a) `--name` becomes undefined behavior that's incredibly confusing if all you have at your disposal is the command line interface.  I think a some form of a giant `(DEPRECATED DO NOT USE)` gets the same message across without potentially confusing users.
  b)  To me, it feels a little paternalistic to _hide_ the flag from the user. I'm all for informing, warning, and making code paths that encourage users to use the preferred code path...  but _hiding_ feels like we don't trust users. 
 c) Are there other places in the helm codebase a similar approach was taken either to hide, or deprecate fields and arguments? 
2. I can look into that, but it makes sense. I'm rather new to the codebase to it will take me some time to implement. 
3. 👍

marckhouzam:
In my mind, a hidden flag will basically mean that unless the user is copy/pasting something that uses `--name` they'll never see this flag.  This specifically addresses the situation of many README's still using the `--name` flag.

But if others also feel it may cause more confusion than value, I'm willing to let that go.

bacongobbler:
Agreed. @desimone can you please mark the flag as hidden? Once that's done this should be good to go.

bacongobbler:
A deprecation notice when the --name flag is used would be great as well. Good point @mattfarina.

marchmallow:
I just started to move to helm3 and stumbled upon this.. my feedback on it: it is a easy change to accommodate in scripts and even if the flag had been maintained in many cases I would have to update automation scripts to use new uninstall option to delete a release.
This was in a way useful as I was following the install guide for a new chart which was updated for helm3 which I didn't notice and running it with helm2 the command failed so I realized should use helm3.
On my shell PATH at least I am currently keeping the helm binary at 2.16.x level since have a bunch of envs on helm2 yet to upgrade, and call helm 3 just `helm3` while I work on migrating charts.. the different install syntax in a way prevents me from calling helm 2 instead of helm3 by mistake.

bacongobbler:
@desimone any progress on this issue? We were planning to cut the first release candidate of Helm 3.2.0 within the next week or so (before KubeCon EU). Any idea if you'll have the time to work on this? If not, we should remove this from the 3.2.0 milestone. Thanks.

desimone:
Unfortunately, I wont have time to work on this in the next week or two.

bacongobbler:
thansk for the update. Removing from the 3.2 milestone.

bacongobbler:
closing for now due to inactivity over the past 7 months. When you have the time to work on this again, please open a new PR and link back to this one for background history. Thanks!

PR #7381, opened 2020-01-11
[helm create] Add hpa boilerplate

### LINKED ISSUE #7369: [helm create] Add autoscaling(hpa) boilerplate
Elasticity is an important feature for cloud native stateless apps such as the one deployed out of the box with helm create.  Allowing users to easily autoscale newly created apps would be beneficial.

I propose we add a disabled by default hpa, with default values of 1 to 100 replicas and 80% cpu utilization threshold.

cc @helm/chart-maintainers

naseemkullah:
Adds hpa boilerplate with following properties which imho are most sane:

* disabled by default
* if enabled:
  * 1 to 100 replicas by default
  * 80% cpu average threshold by default
  * commented out and thus disabled 80% memory average threshold

Fixes #7369

### DIFF
--- a/pkg/chartutil/create.go
+++ b/pkg/chartutil/create.go
@@ -53,6 +53,8 @@ const (
 	ServiceName = TemplatesDir + sep + "service.yaml"
 	// ServiceAccountName is the name of the example serviceaccount file.
 	ServiceAccountName = TemplatesDir + sep + "serviceaccount.yaml"
+	// HorizontalPodAutoscalerName is the name of the example hpa file.
+	HorizontalPodAutoscalerName = TemplatesDir + sep + "hpa.yaml"
 	// NotesName is the name of the example NOTES.txt file.
 	NotesName = TemplatesDir + sep + "NOTES.txt"
 	// HelpersName is the name of the example helpers file.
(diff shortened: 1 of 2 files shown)


naseemkullah:
bump

hickeyma:
CC @helm/charts-maintainers

naseemkullah:
bump

naseemkullah:
bump

scottrigby (approved):
This lgtm in general. It would be good to have a boilerplate best practice for this, especially avoiding conflict with replicas when a HPA is enabled.

pkg/chartutil/create.go
```diff
@@ -231,7 +240,9 @@ metadata:
   labels:
     {{- include "<CHARTNAME>.labels" . | nindent 4 }}
 spec:
+{{- if not .Values.autoscaling.enabled }}
```
scottrigby: PS this is an especially important change that got rolled out to the stable charts repo. Noted in #7369

naseemkullah: 👍

pkg/chartutil/create.go
```diff
@@ -149,6 +151,13 @@ resources: {}
   #   cpu: 100m
   #   memory: 128Mi
 
+autoscaling:
+  enabled: false
```
scottrigby: This line is also important. I think optional - but good to have templated - components like this should be `false` by default.

naseemkullah:
Hi @hickeyma, as per @scottrigby's approval I think this is good to merge now. 🙏

hickeyma (approved):
LGTM, thanks @naseemkullah for adding this.

naseemkullah:
My pleasure @hickeyma!

PR #7383, opened 2020-01-12
feat(chore): Remove unused code

marckhouzam:
This code was not used and was marked for removal a long time ago.
From my understanding of APIs, this is safe to do since neither entities started with an upper-case letter.
As for the test, I assume tests are not considered APIs.
 

mattfarina (approved):
If the test is removed the deadcode checker we run finds the unused code. The test is what stopped the deadcode checker from finding it.

Being unexported this is not part of the API.

marckhouzam:
@mattfarina thanks. Our of curiosity, what dead code checker are you referring to?

PR #7391, opened 2020-01-13
fix(test): Make resetEnv() properly reset settings

marckhouzam:
I got test failures in #7323 when I started using the `--debug` flag in one of my tests commands.

It turns out that because the 'settings' variable is a pointer, it cannot be used to store the original envSettings and then restore them.

Instead, this commit creates an entirely new envSettings, after having set the environment variables, which may affect it.


mattfarina (approved):

marckhouzam:
Thanks @mattfarina

PR #7401, opened 2020-01-14
fix(kube): use non global Scheme to convert

### LINKED ISSUE #7400: Usage of global Scheme can cause Helm to wrongly choose a 3-way-merge patch
While using Helm as a package instead of just using the binary the problem exists that due to the [usage of the global `scheme.Scheme` in `pkg/kube/converter.go` to create versioned objects](https://github.com/helm/helm/blob/master/pkg/kube/converter.go#L40), custom resources added to `scheme.Scheme` by e.g. versioned clients [will make Helm attempt to 3-way-merge patch those resources as they are no longer of type `runtime.Unstructured`](https://github.com/helm/helm/blob/master/pkg/kube/client.go#L373).

My proposal is to use a separate `scheme.Scheme`, with only native Kubernetes added to it, to break free from (generated) versioned Kubernetes clients automagically registering their resources to the global `scheme.Scheme`.

hiddeco:
But instead use a newly initialized Scheme with only Kubernetes native
resources added. This ensures the 3-way-merge patch strategy is not
accidentally chosen for custom resources due to them being added
to the global Scheme by e.g. versioned clients while using Helm as a
package, and not a self-contained binary.

Signed-off-by: Hidde Beydals <hello@hidde.co>

Closes #7400

### DIFF
--- a/pkg/kube/converter.go
+++ b/pkg/kube/converter.go
@@ -17,9 +17,12 @@ limitations under the License.
 package kube // import "helm.sh/helm/v3/pkg/kube"
 
 import (
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/cli-runtime/pkg/resource"
 	"k8s.io/client-go/kubernetes/scheme"
 )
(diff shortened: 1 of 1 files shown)


pkg/kube/converter.go
```diff
@@ -33,12 +36,28 @@ func AsVersioned(info *resource.Info) runtime.Object {
...
+// native resources added to it. This is required to break free of custom resources
+// that may have been added to scheme.Scheme due to Helm being used as a package in
+// combination with e.g. a versioned kube client. If we would not do this, the client
+// may attempt to perform e.g. a 3-way-merge strategy patch for custom resources.
+func kubernetesNativeScheme() *runtime.Scheme {
+	s := runtime.NewScheme()
+	utilruntime.Must(scheme.AddToScheme(s))
+	// API extensions are not in the above scheme set,
+	// and must thus be added separately.
+	utilruntime.Must(apiextensionsv1beta1.AddToScheme(s))
```
hiddeco: Depending on other internals I may not be aware of, not registering this would make it possible to drop the `isCRD` patch recently added here: https://github.com/helm/helm/blob/master/pkg/kube/client.go#L376

pkg/kube/converter.go
```diff
@@ -33,12 +36,28 @@ func AsVersioned(info *resource.Info) runtime.Object {
...
 	return obj
 }
+
+// kubernetesNativeScheme returns a clean *runtime.Scheme with _only_ Kubernetes
+// native resources added to it. This is required to break free of custom resources
+// that may have been added to scheme.Scheme due to Helm being used as a package in
+// combination with e.g. a versioned kube client. If we would not do this, the client
+// may attempt to perform e.g. a 3-way-merge strategy patch for custom resources.
+func kubernetesNativeScheme() *runtime.Scheme {
+	s := runtime.NewScheme()
```
thomastaylor312: We should probably use `sync.Once` to set this up and then only generate it once instead of every time. `convertWithMapper` is used a lot (basically for every resource), so it might be good to do this once. It could be stored in a non-exported global variable local to this package

pkg/kube/converter.go
thomastaylor312: I think we register them for now. We've had a lot of issues with not registering API extensions, so let's play it safe

thomastaylor312 (changes requested):
Good fix on this. Just one point of feedback around constructing the scheme

thomastaylor312 (approved):
LGTM. Thanks for this fix!

PR #7409, opened 2020-01-16
log.Printf replaced to fix the helm upgrade log issue

akazemis:
this is to resolve the log issue

related issue which is actually the consequence of this bug was seen here:
https://github.com/microsoft/azure-pipelines-tasks/issues/11414

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -21,7 +21,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"log"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


akazemis:
the main problem is when you run helm upgrade --force in say azure devops. the log message is considered to be an error so the pipeline fails. Having a look at the code, noticed the log method called is not consistent with the others and fixed it.

thomastaylor312 (approved):
Good catch on this. Thank you for the fix

akazemis:
> Good catch on this. Thank you for the fix

No problem, thanks for the approval

PR #7410, opened 2020-01-16
Add unit test for Reverse() in pkg/releaseutil/sorter.go

tiendc:
Signed-off-by: Dao Cong Tien <tiendc@vn.fujitsu.com>

### DIFF
--- a/pkg/releaseutil/sorter_test.go
+++ b/pkg/releaseutil/sorter_test.go
@@ -79,3 +79,30 @@ func TestSortByRevision(t *testing.T) {
 		return vi < vj
 	})
 }
+
+func TestReverseSortByName(t *testing.T) {
+	Reverse(releases, SortByName)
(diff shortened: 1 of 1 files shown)


pkg/releaseutil/sorter_test.go
```diff
@@ -79,3 +79,29 @@ func TestSortByRevision(t *testing.T) {
 		return vi < vj
 	})
 }
+
+func TestReverse(t *testing.T) {
```
hickeyma: @tiendc Thanks for adding the tests. The checks for each test will however be cached and only fail if the last test fails. Can you separate them into  different test functions?

hickeyma (changes requested):

hickeyma (approved):
LGTM. Thanks for adding the tests @tiendc. Always good to have more tests.

PR #7417, opened 2020-01-16
Update values.yaml linting to respect value options flags

### LINKED ISSUE #7273: `lint` reports error when required value is not set in `values.yaml`
## Description

If a chart has required settings (defined in `values.schema.json`), but those settings don't have a default provided in `values.yaml`, `helm lint` will always report an error, even when the values are set “externally” to the chart. 

For instance, with the following chart:

```
$ tree -F .
.
├── Chart.yaml
├── templates/
├── values.schema.json
├── values.test.yaml
└── values.yaml

$ tail -n +1 Chart.yaml values.schema.json values.yaml values.test.yaml
==> Chart.yaml <==
---
apiVersion: v2
name: foobar
version: 0.1.0
appVersion: 0.1.0

==> values.schema.json <==
{
  "$schema": "https://json-schema.org/schema#",
  "type": "object",
  "properties": {
    "foo": {
      "type": "string"
    },
    "bar": {
      "type": "string"
    }
  },
  "additionalProperties": false,
  "required": [
    "foo",
    "bar"
  ]
}

==> values.yaml <==

==> values.test.yaml <==
---
foo: foo
bar: bar
```

This is an expected behavior:

```
$ helm lint .; echo $?
==> Linting .
[ERROR] values.yaml: - (root): foo is required
- (root): bar is required

[ERROR] templates/: values don't meet the specifications of the schema(s) in the following chart(s):
foobar:
- (root): foo is required
- (root): bar is required


1
```

But this should not report any errors:

```
$ helm lint --set foo=foo --set bar=bar .; echo $?
==> Linting .
[ERROR] values.yaml: - (root): foo is required
- (root): bar is required


1

$ helm lint --set foo=foo --set bar=bar .; echo $?
==> Linting .
[ERROR] values.yaml: - (root): foo is required
- (root): bar is required


1

$ helm lint -f values.test.yaml .; echo $? 
==> Linting .
[ERROR] values.yaml: - (root): foo is required
- (root): bar is required


1
```

## Expected behavior

`helm lint` should not report an error when required values are set “externally”.

## Installation details

Output of `helm version`: `version.BuildInfo{Version:"v3.0.2", GitCommit:"19e47ee3283ae98139d98460de796c1be1e3975f", GitTreeState:"clean", GoVersion:"go1.13.5"}`
Output of `kubectl version`: **not relevant**
Cloud Provider/Platform (AKS, GKE, Minikube etc.): **not relevant**

robbie-demuth:
Previously, the lint command ignored values passed via value options
flags when linting a chart's `values.yaml` file. This was problematic
because such values are often required when validating against JSON
schema

Closes #7273

Signed-off-by: Robbie deMuth <robbie.demuth@appian.com>

### DIFF
--- a/cmd/helm/lint_test.go
+++ b/cmd/helm/lint_test.go
@@ -14,23 +14,25 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package lint // import "helm.sh/helm/v3/pkg/lint"
+package main
 
 import (
-	"path/filepath"
-
-	"helm.sh/helm/v3/pkg/lint/rules"
-	"helm.sh/helm/v3/pkg/lint/support"
+	"fmt"
+	"testing"
 )
 
-// All runs all of the available linters on the given base directory.
-func All(basedir string, values map[string]interface{}, namespace string, strict bool) support.Linter {
-	// Using abs path to get directory context
-	chartDir, _ := filepath.Abs(basedir)
-
-	linter := support.Linter{ChartDir: chartDir}
-	rules.Chartfile(&linter)
-	rules.Values(&linter, values)
-	rules.Templates(&linter, values, namespace, strict)
-	return linter
+func TestLintCmd(t *testing.T) {
+	chart := "testdata/testcharts/chart-with-schema-negative"
+	tests := []cmdTestCase{
+		{
+			name:      "check fails without value option flags",
+			cmd:       fmt.Sprintf("lint %s", chart),
+			wantError: true,
+		},
+		{
+			name: "check passes with value option flags",
+			cmd:  fmt.Sprintf("lint %s --set age=25,employmentInfo.salary=100000", chart),
+		},
+	}
+	runTestCmd(t, tests)
 }
--- a/pkg/lint/rules/values.go
+++ b/pkg/lint/rules/values.go
@@ -28,7 +28,7 @@ import (
 )
 
 // Values lints a chart's values.yaml file.
-func Values(linter *support.Linter) {
+func Values(linter *support.Linter, values map[string]interface{}) {
 	file := "values.yaml"
 	vf := filepath.Join(linter.ChartDir, file)
 	fileExists := linter.RunLinterRule(support.InfoSev, file, validateValuesFileExistence(vf))
@@ -37,7 +37,7 @@ func Values(linter *support.Linter) {
 		return
 	}
 
-	linter.RunLinterRule(support.ErrorSev, file, validateValuesFile(vf))
+	linter.RunLinterRule(support.ErrorSev, file, validateValues(vf, values))
 }
 
 func validateValuesFileExistence(valuesPath string) error {
@@ -48,8 +48,8 @@ func validateValuesFileExistence(valuesPath string) error {
 	return nil
 }
 
-func validateValuesFile(valuesPath string) error {
-	values, err := chartutil.ReadValuesFile(valuesPath)
+func validateValues(valuesPath string, values map[string]interface{}) error {
+	fv, err := chartutil.ReadValuesFile(valuesPath)
 	if err != nil {
 		return errors.Wrap(err, "unable to parse YAML")
 	}
@@ -63,5 +63,5 @@ func validateValuesFile(valuesPath string) error {
 	if err != nil {
 		return err
 	}
-	return chartutil.ValidateAgainstSingleSchema(values, schema)
+	return chartutil.ValidateAgainstSingleSchema(chartutil.CoalesceTables(values, fv), schema)
 }


pkg/lint/rules/values.go
```diff
@@ -63,5 +63,5 @@ func validateValuesFile(valuesPath string) error {
 	if err != nil {
 		return err
 	}
-	return chartutil.ValidateAgainstSingleSchema(values, schema)
+	return chartutil.ValidateAgainstSingleSchema(chartutil.CoalesceTables(values, fv), schema)
```
robbie-demuth: I took a bit of inspiration from https://github.com/helm/helm/blob/v3.0.2/pkg/lint/rules/template.go#L64. That being said, I noticed some potential oddities when poking around https://github.com/helm/helm/blob/v3.0.2/pkg/chartutil/coalesce.go. 

First, as per the first link, template linting invokes `chartutil.CoalesceValues(chrt *chart.Chart, vals map[string]interface{}) (Values, error)`. Somewhere along the line, this in turn calls `coalesceValues(c *chart.Chart, v map[string]interface{})` which handles special cases like removing keys for null values. The function I'm invoking here, on the other hand, doesn't contain any code paths that invoke that same logic. Is that going to be a problem?

Second, `chartutil.CoalesceValues(chrt *chart.Chart, vals map[string]interface{}) (Values, error)` operates on a `*chart.Chart` object. This means that it properly handles values from chart dependencies, etc. Per https://helm.sh/docs/topics/charts/#schema-files, the final `.Values` object is checked against all subchart schemas. Does this mean that `values.yaml` linting needs to respect more than just the parent chart's `values.yaml` file and JSON schema? If so, I think the changes will likely be much more involved than what I've done here

robbie-demuth: Also, I wasn't sure what to do in terms of testing. Currently, there isn't really any testing around `values.yaml` linting. I could certainly add some unit tests for this function, but they wouldn't have caught this issue. Catching this issue would've likely required some higher-level integration tests that actually execute the CLI. Do you all think that this change requires additional test code? If so, what do you all recommend?

robbie-demuth:
Also, given that this is a bug fix, is making the PR against the master branch the right thing to do? Should it be released in the next `v3.0.x` patch?

cmd/helm/lint_test.go
```diff
@@ -0,0 +1,38 @@
...
+func TestLintCmd(t *testing.T) {
+	chart := "testdata/testcharts/chart-with-schema-negative"
+	tests := []cmdTestCase{
+		{
+			name:      "check fails without value option flags",
+			cmd:       fmt.Sprintf("lint %s", chart),
+			wantError: true,
+		},
+		{
+			name: "check passes with value option flags",
```
robbie-demuth: As expected, this test case fails without the other changes

bacongobbler:
> Also, given that this is a bug fix, is making the PR against the master branch the right thing to do?

yes

robbie-demuth:
Is there anything else I need to do here, @hickeyma CC @bacongobbler @adamreese

robbie-demuth:
Quick ping, @hickeyma @bacongobbler @adamreese

pkg/lint/rules/values.go
```diff
@@ -28,7 +28,7 @@ import (
 )
 
 // Values lints a chart's values.yaml file.
-func Values(linter *support.Linter) {
+func Values(linter *support.Linter, values map[string]interface{}) {
```
bacongobbler: As per our semver contract, we cannot change function signatures. Users consuming this function in their codebases would break when they upgrade from 3.0 to 3.1. General rule is "code that used to work before will continue to work in future releases".

Can you try to find another way to resolve this without breaking backwards compatibility?

bacongobbler (changes requested):

pkg/lint/rules/values.go
robbie-demuth: That makes sense. Is changing the internals of exposed functions okay as I've done [here](https://github.com/helm/helm/pull/7417/files#diff-432bdc8e6c573117062cb071cdd1aaa6R33) okay? If so, I think exposing a new function that I call there while leaving the existing function intact should work. Also, what do you think about [my thoughts below](https://github.com/helm/helm/pull/7417/files#r367946663)

bacongobbler: As long as the expected behaviour is the same (that is, if you provide the same input, you'll receive the same output), then yes.

robbie-demuth: Hmmm, there's definitely room for interpretation around the work 'expected'. Given that this is a bug, the current behavior isn't really the expected behavior. Today, if someone were to call `pkg/lint/All(...)` with values that satisfy the requirements of JSON schema that are not satisfied by the chart's builtin values, the linter would return errors. That being said, _this is unexpected behavior_ as described in the pull request and related issue. With the change I've made to the internals of the function, the linter would no longer return any errors - which is a different output with the same input, but yields the desired behavior

technosophos:
As was pointed out in Feb., this has an API breaking change that needs to be fixed before we can re-review. Is this PR dead or still undergoing work?

sean-krail:
Is this a duplicate of https://github.com/helm/helm/pull/7984?

robbie-demuth:
I mostly stopped working on this because I wasn't sure what I needed to do to address the concerns around the API change. To me, [this comment](https://github.com/helm/helm/pull/7417#discussion_r375961183) makes it seem like adding a new function to `pkg/lint/rules/values.go` still isn't sufficient because the change in `pkg/lint/lint.go` would still cause the `All` function to return a different output when given the same input. I suppose I could introduce a new function there as well, but then that would just bubble up to invocations of the CLI itself

As for @sean-krail's question, I believe the answer is yes. I'm curious to see if there are any concerns there around the change to the internals of the `All` function in `pkg/lint/lint.go`

technosophos:
I explained things in more detail on #7984, but the tl;dr is that the only thing we're really concerned with on this PR is the breaking change to `Values()`.

robbie-demuth:
Closing in favor of https://github.com/helm/helm/pull/7984

PR #7427, opened 2020-01-17
Use /usr/bin/env for bash

poelzi:
After this change, make works on nixos.


lindhe:
I've seen this solution many times, and used it too, so I know that it works. But just out of curiosity: why can we assume to know the location of `env` but not of `bash`?

poelzi:
On nixos this is a convinence wrapper and the only file in /usr/bin . NixOS does not follow the Linux Standard Base.
Each profile has a different path and each executed program runs in the same environment it was compiled in.

bacongobbler (approved):

PR #7431, opened 2020-01-18
feat(helm): add recreate upgrade (rollback) strategy

### LINKED ISSUE #7082: proposal: handle immutable resources on update
Steps to reproduce the issue:
```
helm create tmp
echo '
kind: RoleBinding
metadata:
roleRef:
    kind: ClusterRole
    name: {{ .Values.role }}
subjects: []
```
This fails with 
```
```

Output of `helm version`:
Output of `kubectl version`:

kind (and Openshift)

dastrobu:
An additional optional flag `--recreate` can be passed on upgrade (or rollback) of a release.
In combination with the `--force` flag the following strategies are employed when updating a resource (which can be directly compared to kubectl):

```
helm                        kubectl          action                         on 'invalid' or 'conflict'
--------------------------------------------------------------------------------------------------------------

upgrade                     apply            PATCH                          fail

upgrade --force             replace          PUT                            fail

upgrade --recreate          apply --force    PATCH                          DELETE -> GET (poll) -> POST

upgrade --recreate --force  replace --force  DELETE -> GET (poll) ->  POST  fail
```

The 'on error' column should be interpreted as follows. The server responds with 'invalid' e.g. if a certain resource contains an immutable field, which cannot be patched or updated by a PUT. In theses cases it can be helpful to delete the resource and recreate it. Examples for theses cases are:

 * roleRef in RoleBinding
 * spec.clusterIP in Service
 * parameters in StorageClass

Closes #7082.


bacongobbler:
Hi @dastrobu. Would you mind looking into the failing tests?

pkg/kube/client.go
```diff
@@ -149,8 +150,8 @@ func (c *Client) Build(reader io.Reader, validate bool) (ResourceList, error) {
 // occurs, a Result will still be returned with the error, containing all
 // resource updates, creations, and deletions that were attempted. These can be
 // used for cleanup or other logging purposes.
-func (c *Client) Update(original, target ResourceList, force bool) (*Result, error) {
-	updateErrors := []string{}
+func (c *Client) Update(original, target ResourceList, force bool, recreate bool, timeout time.Duration) (*Result, error) {
```
bacongobbler: This would be considered an API change to the SDK. In general we try to avoid these. Is there a way we can refactor this such that this function signature remains the same?

bacongobbler (changes requested):

dastrobu:
@bacongobbler thanks for looking at the code already. As mentioned in the [pr description](https://github.com/helm/helm/pull/7431#issue-364409976) this pr is still WIP. I just wanted to implement the feature quickly to play around with it and get feedback if we should implement it like that. 
I interpret your comments that you would encourage implementing the feature that way. 

I am aware of the API change, see my TODO comment in [interface.go#L48](https://github.com/helm/helm/blob/5e3bf420c7becc7f7cbed3bfe2352f9751bbf572/pkg/kube/interface.go#L48). However I don't have a good idea to do it non-breaking. Hints would be welcome. 
Ideas: 
 1. add parameters `recreate bool, timeout time.Duration` -> breaking
 2. add new interface method `UpdateRecreeate` -> breaking (and ugly)
 3. add a new extended Interface `kube.InterfaceV2`, add field to [Configuration](https://github.com/helm/helm/blob/6cc039ea79a435c9f335d677b78d9be574b8e770/pkg/action/action.go#L68) and use `KubeClientV2` if present and `KubeClient` otherwise. I guess this would be non breaking but also not very nice and complicated. 

Any better ideas? 

from: [Semantic Versioning](https://github.com/helm/helm/blob/master/CONTRIBUTING.md#semantic-versioning)
> Go libraries inside of pkg/ SHOULD remain backward compatible
I would appreciate some advice how to do it non breaking without cluttering the code base too much or if we should consider it an acceptable API change. 

If we would change method signatures from 
```
Update(original, target ResourceList, force bool) (*Result, error)
```
to 
```
Update(original, target ResourceList, options UpdateOptions) (*Result, error)
```
it would have been easier to add more optional flags. But this is also not possible without a breaking API change.

Of course I will look into failing tests and add new tests where appropriate (maybe next week).

bacongobbler:
> Hints would be welcome.

You could try adding `RecreateOnUpdate` and `UpdateTimeout` to the [`Client` object](https://github.com/helm/helm/blob/5e3bf420c7becc7f7cbed3bfe2352f9751bbf572/pkg/kube/client.go#L55-L59), updating those values during `prepareUpgrade` and `prepareRollback`.

pkg/kube/client.go
```diff
@@ -412,25 +413,61 @@ func updateResource(c *Client, target *resource.Info, currentObj runtime.Object,
...
+			if apierrors.IsConflict(err) || apierrors.IsInvalid(err) {
+				err = c.deleteAndCreate(helper, target, timeout)
+			}
+			return errors.Wrapf(err, "failed to recreate %q with kind %s", target.Name, kind)
+		}
+	case force:
 		obj, err = helper.Replace(target.Namespace, target.Name, true, target.Object)
 		if err != nil {
-			return errors.Wrap(err, "failed to replace object")
+			return errors.Wrapf(err, "failed to replace %q with kind $s", target.Name, kind)
```
bacongobbler: typo: should be `%s` rather than `$s`

bacongobbler (changes requested):

bacongobbler:
this is looking really close, by the way. Nice work!

bacongobbler:
hey @dastrobu. We discussed during today's dev call that we are looking at a Helm 3.1 RC in the next week or so. Are you going to have the time to work on this in that timeframe?

dastrobu:
@bacongobbler I plan to work on this today, so PR should be ready for review on Monday hopefully. Would be great to see this feature in 3.1.

dastrobu:
> > Hints would be welcome.
> 
> You could try adding `RecreateOnUpdate` and `UpdateTimeout` to the [`Client` object](https://github.com/helm/helm/blob/5e3bf420c7becc7f7cbed3bfe2352f9751bbf572/pkg/kube/client.go#L55-L59), updating those values during `prepareUpgrade` and `prepareRollback`.

The problem I see is that `prepareUpgrade` and `prepareRollback` do not have access to the `Client` object, these methods can only use the `kube.Interface` which is passed around in the `action.Configuration` object. So I am not sure how this could work
I will try to create an extended `kube.InterfaceV2` defining the extended API in a portable way. I wouldn't call this solution beautiful, but it will be non breaking and will also provide access to the new flag via the go API. 
Any better ideas are still welcome.

pkg/kube/client.go
dastrobu: Changes should be non breaking now, although the solution is not very nice. PR is ready for final review.

bacongobbler:
Thanks for looking into possible solutions. This is going to take a bit more time to review, so we'll have to reschedule this for 3.2.

jkroepke:
Do we see this feature in 3.2?

bacongobbler:
I'm going to re-review this today after reviewing #7649.

dastrobu:
@bacongobbler any updates on the review? It would be great to see this in 3.2 and I am still open for feedback on the PR or suggestions on how to improve the implementation.

antonakv:
Do we see this "Important for users but not important at all for software architect dreaming in a deep theories only" feature in 3.2? ;)

technosophos:
Checking in on this one.

- It needs a quick rebase
- I need to test it against some other PRs that came through that could have an impact
- @bacongobbler if I don't hear back from you, I'm going to go ahead and re-do the review myself. I know you're super busy, and I don't want to distract you. I just want to make sure you have a chance to weigh in if you want.

technosophos:
@antonakv I do not understand your question. Can you please ask it more clearly? Also, have you tested this and confirmed it works? I am not sure what you are trying to get at.

jkroepke:
@technosophos This issue is a blocker for some of us which try to migrate from helm 2 to helm 3.

For example: While changes on hostnames at ingress objects works well in helm by pass a `--force` to helm2 upgrade in helm 3 the ingress object needs to be manually deleted before run helm upgrade.

Or: In helm 3 there is no drop-in replacement for helm 2 `helm upgrade --force` (or `kubectl apply --force`) which is mandatory in some certain situations, e.g developers doesn't have permissions to modify objects in production but deployments via `CI/CD` are allowed.

Since the milestone of this issue / pull-request is set to [3.2.0](https://github.com/helm/helm/milestone/83) our big hope is to see this feature / pull-request in the next version otherwise we have to wait months again e.g. 3.3.0 ...

technosophos:
I have removed this off of 3.2.0. There are several things here that are concerning to us--things that are likely to break existing deployments. So we are going to do a more thorough review after 3.2 is released.

During our review, the main thing that came up was how well the main issue in #7082 was addressed here. It was unclear how come a combo of `helm delete` and `helm upgrade --install` was not sufficient to accomplish the desired result.

voron:
>a combo of helm delete and helm upgrade --install was not sufficient

is it possible to make a combo atomic ?

dastrobu:
> It was unclear how come a combo of helm delete and helm upgrade --install was not sufficient to accomplish the desired result.

The major issue for me is that this causes as unnecessary downtime. Hence, it is not acceptable in production. Especially when you have a large umbrella chart with many resources and potentially long running pre and post install hooks deleting all resources and installing everything from scratch because one immutable property in one resource changed does not seem to be the answer.

I think the expectations on Helm boil down to two aspects:

1. bring the cluster to the desired state, whatever it takes to reach that state
2. keep as many resources as possible to avoid unnecessary downtimes

I am not saying that this PR is the only way of handling issues with updates of immutable fields, but a solution is definitely required before we can employ Helm 3 in production. 

Apart from the arguments above I think requiring people to do an imperative `delete` and `install` afterwards is a Kubernetes anti pattern where everything is done declarative for good reasons. What if my CI pipeline crashes after the `delete` statement... I don't even want to think about that in production.

antonakv:
> > It was unclear how come a combo of helm delete and helm upgrade --install was not sufficient to accomplish the desired result.
> 
> The major issue for me is that this causes as unnecessary downtime. Hence, it is not acceptable in production. Especially when you have a large umbrella chart with many resources and potentially long running pre and post install hooks deleting all resources and installing everything from scratch because one immutable property in one resource changed does not seem to be the answer.
> 
> I think the expectations on Helm boil down to two aspects:
> 
> 1. bring the cluster to the desired state, whatever it takes to reach that state
> 2. keep as many resources as possible to avoid unnecessary downtimes
> 
> I am not saying that this PR is the only way of handling issues with updates of immutable fields, but a solution is definitely required before we can employ Helm 3 in production.
> 
> Apart from the arguments above I think requiring people to do an imperative `delete` and `install` afterwards is a Kubernetes anti pattern where everything is done declarative for good reasons. What if my CI pipeline crashes after the `delete` statement... I don't even want to think about that in production.

Exactly! +100500

blodan:
Note sure if it's possible to se how many is subscribed to this PR but as I can't seem to see it anywhere, I'll just add a note that we are also waiting for this PR to get merged as we currently can't use helm 3 in our pipelines so we are still stuck on helm 2

bacongobbler:
@antonakv friendly reminder that this project abides by the CNCF Code of Conduct. Harassment is not OK. For a quick refresher: https://github.com/cncf/foundation/blob/master/code-of-conduct.md

Upon further review of this PR and https://github.com/helm/helm/issues/7082#issuecomment-575612042, it looks like there does appear to be value in this PR. Apologies for the confusion here.

@technosophos have you had a chance to read https://github.com/helm/helm/issues/7082#issuecomment-575612042 yet? After re-reviewing that comment, this PR starts to make a lot more sense. This PR seeks to implement `kubectl apply --force` and `kubectl replace --force`, so the use cases between this flag and `helm delete && helm install --replace` see different.

The one concern we had in discussions over this feature was how recreating resources may impact by downstream resources. We may still want to do a bit more digging, but I believe the kind sorter protects us from most issues when resources are deleted and re-created. Did you have any sample charts or examples that may run into resource collision issues we may want to investigate further?

jkroepke:
Hello!

thanks for the 3.2 release of helm. I wish we could move forward on this topic.
If you look below at https://github.com/helm/helm/issues/7082#issuecomment-590943722 the list of linked issues grow.

About the SemVer concerns comments ago. SemVer allows such breaking changes by bumping the major version. By using SemVer you should not spring back to use this.

bacongobbler:
I’m not sure what you’re suggesting here. Are you suggesting that we wait until Helm 4 before releasing this? That would not be ideal if we can move this forwards in a backwards-compatible fashion (and I’m sure we can).

jkroepke:
Helm 4 could be the next after Helm 3.2. 

It‘s not a suggestion. Just a idea or alternative.

At the end it would be great to have a working force switch. kubectl apply --force might be a good reference as mention above.

An other idea would be to mark this as experimental flag and then merge it. The benefit is a large userbase wo could test this flag and receive some feedback from them. A good example is the API (alpha, beta, stable) level a Kubernetes.

Whatever. It‘s not worth to discuss it.

Follow the bc fashion might be the best. It would be also great to see such features similar to this pr in the next helm version.

bacongobbler:
Adding to the 3.3.0 milestone so we don't lose track of this feature in the mix

bacongobbler:
@jkroepke it's a good suggestion. For better or worse, the industry believes that a major milestone warrants more than a small breaking change in the API. It's unlikely we'd release a Helm in place of a Helm 3.3 just for one feature... @technosophos is already starting to look at other features we'd consider for Helm 4 like #8051.

In other words, it would be highly unlikely that this feature would be shipped any time soon if we shot for Helm 4, which is why I'd prefer to look at other solutions. :)

Hope that helps clarify our current train of thought here.

bacongobbler:
A feature gate would be a neat idea, however the breaking change in this case was further down the stack in `pkg/kube`. A feature gate wouldn't help in this scenario.

antonakv:
I'm curios what was the idea to create Helm 3 without upgrade(rollback) strategy ? Are the some user cases where it's not applicable ? Is it usable somewhere excepting of sandbox environments ?

wasker:
Isn't this feature naturally gated by the `--recreate` command line option?

squillace:
> I'm curios what was the idea to create Helm 3 without upgrade(rollback) strategy ? Are there some user cases where it's not applicable ? Is Helm 3 usable somewhere on Production and not only on sandbox environments ?

I personally love snark, but when it's wrong and incorrectly aimed in serious environments without context, it tends to convey the idea that one person correctly understands the world and no one else. Is that the idea you intended to convey, @antonakv ?

bacongobbler:
> Isn't this feature naturally gated by the `--recreate` command line option?

Yes, but not entirely, @wasker. While it is indeed behind a feature flag, there's a difference between some features that are considered stable or unstable in Helm. For example, the [OCI registry work](https://helm.sh/docs/topics/registries/) (`helm chart` subcommands) are all gated behind a feature flag `$HELM_EXPERIMENTAL_OCI`. They are considered unsafe for production use at this time, but users are welcome to experiment with it to find bugs and report them before it's considered a stable feature.

We're getting off-topic here though. There's no intention to put this flag behind a feature gate or to punt this work off until Helm 4. We're just trying to figure out how to move this forward without breaking our SDK compatibility guarantees for others.

If you have further questions, I'd be more than happy to discuss Helm's SemVer contract in another issue or on Slack. :)

> I'm curios what was the idea to create Helm 3 without upgrade(rollback) strategy ?

`helm rollback` and `--atomic` exist for rollbacks. This is a new upgrade strategy to to re-create resources on an upgrade. I'm not sure what you're getting at here.

antonakv:
> > I'm curios what was the idea to create Helm 3 without upgrade(rollback) strategy ? Are there some user cases where it's not applicable ? Is Helm 3 usable somewhere on Production and not only on sandbox environments ?
> 
> I personally love snark, but when it's wrong and incorrectly aimed in serious environments without context, it tends to convey the idea that one person correctly understands the world and no one else. Is that the idea you intended to convey, @antonakv ?

I'll try to be maximally correct in my answer to don't violate good manners agreement used here. It's not wrong and truly reflects the situation If you check all the user complains about Helm 3 production usage in this ticket plus [#7082](https://github.com/helm/helm/issues/7082). It's reason why a lot of projects are not considering migration from Helm 2 to Helm 3. It's simply proved by user experience.

squillace:
I appreciate your attempt to asymptotically approach the boundaries of proper communication, and I accept your statements at face value. The positions you assert here are not wrong, _per se_ -- everyone has a real data set for them! -- but this doesn't make yours persuasive; certainly not to me, though I'm only a voice here. 

My response would be that the explosion of helm usage since helm 3 was released above and beyond the previous usage in helm 2 stand in the way of your assertions. Whilst you might argue otherwise, I've personally helped some of the world's largest organizations (both non- and for-profit) use Helm in dev, staging, and prod. You are completely within your right to disbelieve this should you wish. 

But let's remain civil, always, given that we're all working on code that we do share and each of us is an individual with the best of intentions.

mcavoyk:
@bacongobbler Is there any update on this for the 3.3.0 milestone? This feature would be very useful for some of our CI pipelines.

allen-servedio:
Following on https://github.com/helm/helm/pull/7431#issuecomment-637852116, we too would really like to see this feature in 3.3.0 of Helm. I see that there are conflicts now with the Kube client. Do you all need any assistance clearing those so that this can roll out in 3.3.0? Thank you all for this excellent tool and for this added feature to it.

allen-servedio:
@dastrobu - do you need any assistance resolving these conflicts? 

Is it possible to have this released with your 3.3.0 Milestone when it is ready @bacongobbler? 

Apologies for the second comment on this, but we are trying to make use of this feature with our automation to minimize how much we have to change when we move to Helm 3. Again, really appreciate all of your efforts here and we are willing to help - just let us know what you need. Thanks!

bacongobbler:
> Is it possible to have this released with your 3.3.0 Milestone when it is ready @bacongobbler?

Assuming @dastrobu finds the time to rebase this PR, yes, we have discussed reviewing this for 3.3.0 assuming it's ready to merge again.

We'll have to get a few other core maintainers to have a look just to make sure it's safe to merge. @technosophos and @adamreese [raised some concerns earlier](https://github.com/helm/helm/pull/7431#issuecomment-614181196) which I believe [I addressed in another comment](https://github.com/helm/helm/pull/7431#issuecomment-618084939), but it would be good to have them check over this again and weigh in on what those concerns might've been.

allen-servedio:
That seems more than reasonable @bacongobbler. We appreciate both the huge amounts of time that you invest in this project and for taking the time to respond to my question. 

@dastrobu please let us know if we can be of any help. If you do not have any further time to invest in this solution, please let us know - maybe we can take it over for you (though that would certainly bump it out of 3.3.0)? Just let us know.

jkroepke:
@allen-servedio Cloud you create a new pull request based on this one ... as fallback if @dastrobu isn't available. 

git author and git commiter are 2 fields in git. You could preserve the original author in your PR, if you want.

dastrobu:
@allen-servedio I started rebasing yesterday, but there are some changes that require some non trivial manual adjustments of this PR (e.g. the refactored shell completion). I hope I can get this done tomorrow or at least next week.

marckhouzam:
> @allen-servedio I started rebasing yesterday, but there are some changes that require some non trivial manual adjustments of this PR (e.g. the refactored shell completion). I hope I can get this done tomorrow or at least next week.

@dastrobu Looking at the conflicts, I believe the only one related to completion is `scripts/completions.bash`.  This file has been removed from the `master` branch, so you can simply remove it when you do your rebase.

dastrobu:
I rebased the PR, did some adjustments and quick testing of the compiled helm client. 
@bacongobbler, @technosophos and @adamreese I would appreciate your feedback. 
If there are still some concerns on the concept or problem this PR addresses feel free to comment.

allen-servedio:
Thank you @dastrobu - you rock!!! Really appreciate your efforts to try to keep this going.

bacongobbler:
Given v3.3.0-rc.1 is scheduled for release tomorrow, I'm going to bump this to 3.4 as we're still waiting on other maintainers to review this. Sorry @dastrobu.

ewiger:
This is highly anticipated feature and fix. This strategy will help to resolve a lot of confusion spread over multiple issues, including the recent once.

An alternative approach to recreating resources would be, to optionally filter out immutable fields directly from the YAML manifest of the patch. But I must admit, that implementation by @dastrobu is superior to that. It is not hard to imagine that more fields can be marked as immutable on the kubernetes api side in the future. So this strategy will be able to work with it.

What are still the missing steps to get this strategy merged into master? How can I help speeding up the process?

pkg/kube/client.go
ewiger: this has been fixed as:
`return errors.Wrapf(err, "cannot patch %q with kind %s", target.Name, kind)`

jkroepke:
And with this proposal how I'm able to change immutable fields like I could do it on helm 2?

allen-servedio:
@jkroepke - see the [original PR comment](https://github.com/helm/helm/pull/7431#issue-364409976). One scenario (just `--recreate` without `--force`) will try to do a PATCH first and, if that is invalid or conflict, it will do a DELETE, wait for the delete to complete, then POST. This is similar to what we had with Helm 2.0's  `--force` parameter. See https://github.com/helm/helm/issues/7082#issuecomment-559558318 for an explanation of what Helm 2.0 used to do and what Helm 3.0 does now.

In any event, several of us would like to use the behavior in this PR for workflows that depended on that behavior from Helm 2.0. We have been holding off upgrading specifically to get this feature and reduce the numbers of changes in our workflows or helm charts due to the change in behavior between Helm 2 and Helm 3.

In any event, we totally understand how busy the maintainers of Helm are and that they want to do what they feel is right in reviewing this kind of change (because if it is not right, who is going to get all of the new issues :-) ). We hope to see this in Helm 3.4.

dastrobu:
one of the early remarks of @bacongobbler in https://github.com/helm/helm/issues/7082#issuecomment-559558318 was: 

> Helm 2 used to delete and re-install the resource on a PATCH failure, which led to a lot of cases where things like spurious client timeouts causing the patch to fail would delete the resource... Not exactly what you'd call a great experience.

So it would help if an experienced Helm 2 user (which I am not) could explain in more detail what went wrong in Helm 2. Given that the strategy I am proposing is basically copied over from kubectl which is available for quite some time now. I was assuming that this works quite well. Getting a better understanding of might go wrong with this approach could help. 

I also thought about approach to  
> optionally filter out immutable fields directly from the YAML 

but I assume there is no way to detect those fields. While one could hard code that for standard Kubernetes resources it would definitely not work for CRDs. One case where changing an immutable field is actually what we want in an helm upgrade is the change of the host in an OpenShift route object (see [1426666](https://bugzilla.redhat.com/show_bug.cgi?id=1426666)). There may be many other cases where changing an immutable field via helm upgrade is exactly what we want and currently there is no other way than either uninstalling everything before (not acceptable) or deleting the resources that should be changes manually before the helm upgrade. This is also a dangerous workaround that we currently need to do from time to time.

allen-servedio:
Hi @bacongobbler, @technosophos and @adamreese - since we do not have visibility into your (and your colleagues') no doubt hectic schedules, I am looking for a little guidance on your ability to spend time reviewing this PR for the upcoming [3.4.0 milestone](https://github.com/helm/helm/milestone/94). If you feel that it is unrealistic to properly review this for that milestone and that you strongly feel it will continue to slide out, please let us know. I don't know if you all work against a backlog where we would have a sense of where this fits in, but even a rough estimate of when it would be realistic for you all to review this would be super helpful (if that is even practical...).

This is clearly an itch several of us are trying to scratch, but we totally understand that you have MANY important issues that you are working on and that you are trying to do the right thing by the tool and the community (which we all appreciate). Any guidance that you can give us will help us decide on what next steps make the most sense for us to take.

Thank you for all you do and for your continued extraordinary efforts on behalf of everyone to make Helm better and better. Also thanks to @dastrobu for creating this PR and giving us the opportunity to use this workflow - very much appreciated.

technosophos:
The three of us you pinged are currently prioritizing other work over this particular feature addition. This PR is definitely difficult to review because it has the potential to break releases in unanticipated ways. I have become increasingly reluctant to introduce features in minor releases that have the potential to break things. Adam and I already did a deep review of the code, and I already left my own feeling that delete/recreate is a sufficient solution. While the trade-off there is possibly more downtime, it's a _known_ trade-off without side-effects.

You'll have to get some of the other core maintainers to spend some time reviewing this if you really want it to get in.

dastrobu:
@technosophos thank you for sharing your thoughts about this PR. 

> This PR is definitely difficult to review because it has the potential to break releases in unanticipated ways.

Could you explain this in more detail? If it would help, we could try to break down the PR in smaller parts or try to change parts that you consider "dangerous". I am not sure how the feature (completely guarded behind a new flag) would break anything in existing use cases. 
I would be interested in particular if you have doubts about deleting and recreating a resource in general or the way it is implemented or controlled by the flag. Since there could be alternative approaches to control the recreation of resources (e.g. annotation based, like for helm hooks) I don't see any other option to recreating a resource if an immutable field should be updated. This more detailed feedback would help the community to think about alternative approaches the immutable field update problem that may not have "the potential to break releases in unanticipated ways".

> I already left my own feeling that delete/recreate is a sufficient solution. While the trade-off there is possibly more downtime, it's a known trade-off without side-effects.

I think this is the key issue where I and many watchers of this PR disagree with your position. If downtime is acceptable depends on the business needs. And it is quite hard to understand that one needs a downtime to change e.g. a roleRef in a RoleBinding. 
If it is a common position of the core maintainers that Helm 3 favours uninstall/install including downtime over an upgrade strategy that can handle immutable fields, this has some implications that should clearly be reflected in the docs: 

1. If downtime is not acceptable, users should consider to stay at Helm 2, since it can handle the upgrades with --force. 
2. Umbrella charts become kind of an antipattern, since the number of resources can become huge with an uninstall/install of the entire umbrella chart. (see [Complex Charts with Many Dependencies](https://helm.sh/docs/howto/charts_tips_and_tricks/#complex-charts-with-many-dependencies) "The current best practice for composing a complex application from discrete parts is to create a top-level umbrella chart..."). Re-installing an entire umbrella chart just because one immutable property in one resource is not what most users want. Instead one would prefer to install all sub-charts individually, to be able to re-install only the sub chart that requires a re-install to restrict downtime to the sub chart.
3. Chart authors should consider to split their Helm charts into several charts, e.g. one with long running hooks, deployments and so on and one with resources containing immutable properties such as Service (clusterIp), RoleBinding (roleRef) and possible others. 

From my point of view none of the three points above are any good recommendation. I just wanted to point out what consequences the decision to favour re-installs with downtimes over an upgrade strategy that can handle immutable fields might have. I personally love umbrella charts and we usually setup one umbrella chart to handle all resources in an entire namespace, but we would need to reconsider this practice.

dudicoco:
@technosophos for many of us downtime is NOT a viable solution in a production environment.

bacongobbler:
@dudicoco then why are you asking for this feature? This feature flag WILL incur downtime. That is its intended design.

> If downtime is not acceptable, users should consider to stay at Helm 2, since it can handle the upgrades with --force.

I don’t think this is the right approach here. Again, this feature flag will delete the objects from the cluster, then re-create them. That is, by definition, incurring downtime.

@technosophos’s assessment is correct - this feature can be replicated today with `helm uninstall && helm install -replace`. If `helm uninstall -wait` was implemented, then we’d be able to replicate the entire feature set of this proposal without introducing much logic to the client, and with minimal downtime.

dastrobu:
@bacongobbler and @technosophos I have the impression there is still some misunderstanding about the behaviours of `helm upgrade --recreate` and `helm uninstall && helm install --replace`.  

Let me try to illustrate by examples. 

**Umbrella chart**
Given the following umbrella charts in pseudo code
```
umbrealla-v1: 
 - business-critical-app-v1
 - less-critical-monitoring-component-v1
```
This should be updated to 
```
umbrealla-v2: 
 - business-critical-app-v1
 - less-critical-monitoring-component-v2
```
Which may involve updating a resource `r` with an immutable field part of `less-critical-monitoring-component-v1`.

With `helm upgrade --recreate` all the resources of `business-critical-app-v1` remain untouched und just `r` gets deleted and re-created (possibly creating a short downtime on `less-critical-monitoring-component-v2`). Business is happy, since there was no downtime on `business-critical-app-v1`. 

With `helm uninstall && helm install --replace` all resources of `business-critical-app-v1` get deleted. Afterwards potentially long running pre-install hooks get executed. After they did run `business-critical-app-v1` gets created again. Additionally there may be long start up times of the replicas of `business-critical-app-v1`, which is usually not an issue if doing a rolling deployment to a new version. This adds up to a downtime comprising the pre-install hook time plus the container start-up time of the first replica. This is a major drawback of the `helm uninstall && helm install --replace` approach. As pointed out in https://github.com/helm/helm/pull/7431#issuecomment-665787742, this could be mitigated by discouraging umbrella charts, so one could do `helm uninstall && helm install --replace` on sub charts alone.

**Non-Downtime causing resources**
Given the following chart
```
umbrealla-v1: 
 - business-critical-app-deployment-v1
 - cronjob-v1 (with serviceaccount and other stuff)
```
This should be updated to
```
umbrealla-v2: 
 - business-critical-app-deployment-v1
 - cronjob-v2 (with serviceaccount and other stuff)
```
Assuming there is a permanently running deployment and a cronjob doing some batch work (generating a report once a day), where updating it requires the re-creation of one resource `r` part of the `cronjob-v1`.

The result is the same as for the umbrella chart case. `helm upgrade --recreate` causes no downtime at all, while  `helm uninstall && helm install --replace` causes a downtime due to container start up times and hooks. 

Here, there is no easy mitigation by splitting up an umbrella chart. As also pointed out in https://github.com/helm/helm/pull/7431#issuecomment-665787742, this would require chart authors to split the cronjob part into its own helm chart although its defining part of the system.

In essence recreating any resource that does not cause pod restarts of an entire replica set (or stateful set) should be more or less safe, while deleting something that causes simultaneous pod restarts of an entire replica set should be avoided whenever possible since it causes downtime. 

So I am still hoping for an explanation of how the proposed recreate approach would break a deployment. Is it that you believe that the recreate flag could cause an accidental DELETE request on `business-critical-app-deployment-v1`?
If so, one could think of the direction of letting chart developers add an annotation to resources which are save to be recreated (something like `on-conflict: recreate` which would be added to `cronjob-v2` but not to `business-critical-app-deployment-v1`. This could be an alternative approach to controlling the recreate behaviour on individual resources without a `--recreate` flag. With the current feedback I am a bit stuck. Thinking of alternative approaches is hard since I did not fully understand the concerns with the proposed approach. And with the simplified real world examples above it feels hard to believe either that `helm uninstall && helm install --replace` should be the right way of handling theses problems (regardless of having a `uninstall --wait` implemented, which would basically result in the same behaviour though in a more elegant way). 

Last but not least there is the issue of CI/CD pipelines (not illustrated in the examples) which should always run the same Helm command. For a pipeline it would be hard to decide if `helm upgrade` or `helm uninstall && helm install --replace` needs to be run. Always running `helm uninstall && helm install --replace` would destroy the release history and would cause a downtime on everything even if there is just a change in a comment of the helm chart. 

I hope I can convince you that my primary goal is not getting this PR merged but rather find a good solution to the problem described above so we can all run Helm 3 on our business critical production environments without fearing `helm uninstall` needs to be run on production ever.

voron:
> This feature flag WILL incur downtime. That is its intended design.

Replacement of single resource may not lead to downtime. Or this downtime will be much shorter compared to `uninstall +install`. Take into account all dependent resources such as generated passwords (it survives over upgrade using helm lookup to itself) or created by StatefulSet PVCs & PVs with ` persistentVolumeReclaimPolicy: Delete`. It works fine with `helm2 upgrade --force` and StatefulSet recreation, while there is no simple option to get the data back into auto-created PVC & PV after `helm3 uninstall`. Similar with other immutable fields, for example with Service, a couple seconds downtime to recreate one object compared to `uninstall + reinstall` of complete chart.  What about umbrella charts with tens of minutes rollout time from scratch compared to 1-2 objects replacement? 
People do not want to delete everything in production just to change a small part, which doesn't affect traffic flow frequently. 

I don't know by the way why `install --replace` is marked as `unsafe in production` in helm's help.

>I don’t think this is the right approach here.

That's what we do actually. I cannot migrate to helm3 due to missing real "force". I have to stay on helm2 or cover all the cases where helm3 may stuck with immutable fields(I usually hit StatefulSet or Service's ClusterIP ) with workarounds like "get kubectl and remove some resource manually". I cannot use `helm3 upgrade` inside CI/CD, as pipeline should be durable as much as possible.

bacongobbler:
> How to implement `atomic` with `uninstall + install` ?

The `--atomic` flag is a `helm upgrade` with a fallback to `helm rollback` in case of failure. There's not much else going on. Have a look at the source code if you're more interested in the behaviour.

voron:
>Have a look at the source code if you're more interested in the behaviour.

How could I get `atomic` behaviour with `uninstall + install`? What's a simple way to rollback, when `install` fails after `uninstall`, let's say due to invalid chart or unavailable remote chart repo.

bacongobbler:
@voron if you have a question about things that are unrelated to the PR review, please either ask your questions in #7082 or open a new ticket. That way we can keep the discussion here focused on reviewing this PR.

Okay. Let's rewind and recap some of the use cases here. Let's go back to the **original** use cases as defined in #7082, and explain why this PR will not resolve those cases.

#### [Use case 1](https://github.com/helm/helm/issues/7082#issue-528769808): an upgrade failed because Kubernetes rejected the change, like changing a `roleRef` in a RoleBinding. I want Helm to forcefully delete and re-create this resource.

For background context, https://github.com/kubernetes/kubernetes/issues/66353#issuecomment-408559096 describes why some of these objects in Kubernetes cannot be changed after they have been created.

RoleBindings are relatively safe objects to delete and re-create. But, what is NOT listed in that comment are the potential side-effects of deleting/recreating resources other than a RoleBinding. There are many cases where deleting and re-creating one resource may cause an application to enter an unknown state which the user cannot recover from.

@voron's comment also reinforces this argument: 

> People do not want to delete everything in production just to change a small part, which doesn't affect traffic flow frequently.

Let's take your example, @voron. You're talking about selectively deleting/re-creating volume-backed resources (Secrets, Services, ConfigMaps, PVCs) without touching the Deployment. If a Persistent Volume Claim is deleted and re-created, there is a risk that an application (Deployment) consuming that Persistent Volume Claim will not receive the new volume's mount point. The same can be said for Secrets, Configmaps, Services, and other resources which are deployed and relied upon by another resource. Environment variables populated in a Deployment from a Service won't be updated until the Pods are destroyed.

These are the risks that are unknown, and the behaviour between these resources are not clarified in the Kubernetes documentation. This is why there is a major risk involved here that could result in applications going into an unknown state, and why @technosophos and @adamreese were concerned about the behaviour introduced here.

This is also why @technosophos explained in https://github.com/helm/helm/pull/7431#issuecomment-658440433 that while the trade-off of a `helm uninstall --keep-history && helm install --replace` results in possibly more downtime, the expected downtime is a known trade-off without these unknown and undocumented side-effects.

#### Use case 2: Helm 2's `--force` flag differs from Helm 3's `--force` flag. I want the same behaviour from `helm upgrade --force` in Helm 2.

That one's easy. As I explained earlier in https://github.com/helm/helm/issues/7082#issuecomment-559558318, Helm 2's `--force` flag would delete and re-install EVERYTHING on a PATCH failure. There was no selective delete and re-create like what is being proposed here (and why we have reservations against this flag - as mentioned earlier). Helm 2's `helm upgrade --force` is identical to `helm uninstall --keep-history && helm install --replace`. The only difference is that there's  one extra step: perform a `helm upgrade` and run `helm uninstall --keep-history && helm install --replace` if the upgrade fails.

If you don't believe me, have a look at the code. I am the author of the `--force` flag. I [even documented the behaviour in the comments](https://github.com/helm/helm/blob/73d88af8661828743f08d9d43bbcf9c2f3d7c433/pkg/tiller/release_update.go#L155).

There are other concerns we had with the PR as well (mostly to do with KubeClientV2), but those can all be fixed with further code reviews, feedback and iterations. But the two use cases provided above prevent us from approving this PR.

Hope that helps clarify our current stance on this issue.

jkroepke:
Hi @bacongobbler! 

**Use case 1:**

You are right, using `helm upgrade --force` brings some known risks and some unknown risks. But I would say the benefits of `helm upgrade --force` are heavier then the risks. As you mentions above, most of the object can be re-create without any lost expect the PVC one. But if you are using StatefulSets instead Deployments for applications with data volumes, the PVC will created outside helm and are not effected from `helm upgrade --force`. I guess deleting and recreate Secrets, Configmaps, Services are safe for most of our deployments.

An other use case would be the Route object from RH Openshift since the hostname field is immutable. If I change a hostname in a chart deployed by helm, in helm 3 I have to delete the route objects before the get a successful deployment. Helm 2 could success it with the `--force` Flag.

**Use case 2:**

> That one's easy. As I explained earlier in #7082 (comment), Helm 2's --force flag would delete and re-install EVERYTHING on a PATCH failure. There was no selective delete and re-create like what is being proposed here (and why we have reservations against this flag - as mentioned earlier).

I disagree. I have check this with 2.16 right now.

```
Client: &version.Version{SemVer:"v2.16.9", GitCommit:"8ad7037828e5a0fca1009dabe290130da6368e39", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.16.7", GitCommit:"5f2584fd3d35552c4af26036f0c464191287986b", GitTreeState:"clean"}
```

My test and one of my production charts a has 12 Deployments, 4 StatefulSets, 20 Services, no sub charts. `helm2 upgrade --force` would only remove the affected resource and re-create it. 

No pods where destroyed or re-created while running `helm2 upgrade --force`. 

Based on this statement above, the discussion about the downtimes should be more clear now. A whole uninstall and re-install would destroy all pods while a helm upgrade --force won't (delete and re-create a services should takes less then 2 seconds).

From CI/CD and automation perspective, the `helm2 upgrade --force` was perfect. As a helm user I should not care about immutable fields because helm just resolve such situation if an immutable field needs to be upgrade.

To replicate the behavior in helm3 this would be:

* Run helm3 upgrade
* Read the error messages (find the affected resources)
* Delete them each manually by hand.
* Run helm3 upgrade

It's a no go since this is hard to automate.

> If you don't believe me, have a look at the code.

I don't check and read the code. I wouldn't understand it. I'm not a go developer. But I just run the helm command and look into my cluster what happens.

> Hope that helps clarify our current stance on this issue.
Sorry, It wont help me or can't get the message. But my problem with helm 3 is still there.

voron:
@bacongobbler 
>Helm 2's --force flag would delete and re-install EVERYTHING on a PATCH failure. There was no selective delete and re-create like what is being proposed here (and why we have reservations against this flag - as mentioned earlier).

Sorry, I've just tested it with helm 2.16.9/ k8s 1.16.9 and I see the only re-created resource is the resource with immutable field. Here is [my log](https://gist.github.com/voron/6954a6af3a193ad2ae7ba1d34bbf9b00) with k8s resources' UIDs and timestamps. 
I see that within single `helm2 upgrade --force`:
* immutable resources are deleted & re-created with `uid`,`creationTimestamp` and `resourceVersion` update
* mutable resources are patched with `resourceVersion` update
* unaffected by upgrade but managed by chart resources are not touched, `resourceVersion` is the same

It looks exactly like selective delete & recreate for me.

bacongobbler:
You two are trying to argue about the semantics of `helm upgrade --force` without actually having looked at the code. And trying to tell me how it works.

You two realize I wrote that code, right?

> I don't check and read the code. I wouldn't understand it. I'm not a go developer.

Reading and writing code are two very different things. I’ve never written a line of Scala in my life, but I can try to understand what the author was trying to write, provided that the code is legible.

Here’s where it deletes the old record, along with every resource the chart manages.

https://github.com/helm/helm/blob/73d88af8661828743f08d9d43bbcf9c2f3d7c433/pkg/tiller/release_update.go#L212-L217

And here’s where the new release is created.

https://github.com/helm/helm/blob/73d88af8661828743f08d9d43bbcf9c2f3d7c433/pkg/tiller/release_update.go#L235-L270

jkroepke:
@bacongobbler

Did you run the helm the binary agist a real cluster to verify this or did you just read the helm code?

bacongobbler:
My apologies. I was conflating two separate features of `--force`:

- calling `helm upgrade --force` on a FAILED state causes a `helm uninstall && helm install --replace` (the code I was referring to)
- Helm 2's `helm upgrade --force` will selectively delete and re-create objects if the object cannot be patched (the code @voron referenced)

Regardless, there are unknown risks from selectively deleting and recreating objects, exasperated by the fact that Kubernetes does not define this behaviour. Deleting and re-creating every object is a known state, which is why we're recommending it for those relying on this behaviuour.

Again, in your case as well as the original case described in OP's first comment in #7082 are safe to delete/recreate. But there are other cases like the ones mentioned above that are known to be problematic, and those are not address in the documentation, so we do not know what is the expected outcome of these scenarios.

I'm going to step out here though. I am prioritizing other work than this PR. You'll have to get some of the other core maintainers to spend some time reviewing this if you really want it to get in.

wasker:
I've read the whole thread and while I understand the criticism of --replace behavior, it's better to have this flag in its current incarnation than not to have anything at all. I'd love to see this shipped with the behavior of potential downtime being documented (and maybe a warning message in CLI's output wouldn't hurt), and the "ideal" replacement of the --replace being worked as a separate work item.

cdunford:
Is this going to make it for 3.4?

allen-servedio:
Hi all... We carefully read the concerns of the Helm maintainers about this feature and thought about what we could do to allow our automation to work consistently with Helm's current behavior. As a result, we implemented the following within our automation pipelines:

1. We do a client side `helm lint` of the chart, then we render the helm chart manifest with all the values provided and then pass it to and a `kubeclt apply --validate --dry-run=client -f manifest.yaml` to check that the Helm chart is at least syntactically okay and all the configurations are correctly structured (we have always done this).
2. We run the command `kubectl apply --dry-run=server -f manifest.yaml` (or `kubectl replace ...` if it is a force upgrade) and specifically look for `/immutable|updates .+ forbidden/` pattern as a failure.
3. Our users are able to describe to us what resources (kinds and names) they are okay to be deleted (or should NOT be deleted) when this happens.
4. We then delete the resource if it only has the failure from step 2 above and is allowed to be deleted per step 3.
5. We then do the helm upgrade. Obviously does not matter for an install as there is nothing to compare against - that’s why steps 2-4 are skipped for initial helm release install.

Sadly, this is embedded within a much larger set of custom automation that we have and so all we can offer the community are the steps that we are taking within our automation vs. something you can take and run.

dastrobu:
@cdunford  with the [comment](https://github.com/helm/helm/pull/7431#issuecomment-667161830) of @bacongobbler it seems very unlikely that this will make it into 3.4. I had some thoughts on the problem in the meantime, how one could address the replacement of resources in a better way than just implementing a feature flag and I am trying to propose a [modified solution](https://github.com/helm/helm/issues/7082#issuecomment-687752364). So let's step back and discuss some alternative approaches in the scope of the original issue #7982. Looking forward to your feedback.

wasker:
@dastrobu While these are great call outs, there's also another category of users who are OK with limitations of the proposed approach. E.g., test scenarios, or environments where unsafe recreate is tolerable, etc.

I'd like to suggest pushing this feature in as is, again. If there're concerns about non-safety of the operation, let's rename the verb to "unsafe-recreate" and print a warning in the console. The feature _is useful_ even with its known limitations.

dastrobu:
@wasker a missing fix for this issue is blocking us all. However, it is not my decision if this PR gets merged or not. If you want that feature merged, feel free to convince core maintainers to get this merged. Since @technosophos and @bacongobbler did already share their opinion I guess you need to find some other maintainers who support this PR. If there is a chance to get this merged and there are PR remarks that would need to be implemented, I'll be happy to help finalising this PR.

mattfarina:
Just reviewed this as part of the 3.5.0 release process and some thoughts come to mind...

- A `deleteAndCreate` code path is introduced. Why is this not a manual step? Or a step that automation does when the author of that knows full well they want this? `helm uninstall` followed by a `helm install` would handle that path. The one difference is the retained release history which is useful for rollbacks. But, in this case rolling back to before the uninstall is dangerous. Right now there are no test for all these situations which make me nervous. I need to not be nervous about these code paths.
- I get the impression we're trying to make Helm commands be able to do the same thing as kubectl. They have different target features and we don't need Helm to things with the same number of commands. It's an assumption I picked up through this.
- When people think of using flags they often think of it applying a feature. If you apply 2 flags it applies 2 features. But, in this case applying `--force` changes the first feature. That doesn't feel like a great UX to me.

Issues like this are why we are having trouble taking this to completion. If you could help us with this it would go a long way in moving it along.

dastrobu:
@mattfarina great to see this issue on the 3.5.0 milestone. I am still happy to help on developing this further, although I had to focus on different topics during the last months. 

Some comments on your remarks: 

> A deleteAndCreate code path is introduced. Why is this not a manual step? Or a step that automation does when the author of that knows full well they want this? helm uninstall followed by a helm install would handle that path. [...]

The major difference is that only those resources would be re-created that really need to, which makes a big difference in a huge umbrella chart causing down time as mentioned in https://github.com/helm/helm/pull/7431#issuecomment-665787742 and https://github.com/helm/helm/pull/7431#issuecomment-666572304. 

Many of us expected the return of the Helm 2 `--force` behaviour. However, @bacongobbler convinced me that this might not the best approach: https://github.com/helm/helm/pull/7431#issuecomment-666735299. I am still not convinced, though, that we shouldn't do something about the issue to [handle immutable resources on update](https://github.com/helm/helm/issues/7082). I already proposed an alternative way in https://github.com/helm/helm/issues/7082#issuecomment-687752364 but did not receive much feedback on that proposal which is why I did not take the time to create a PR on that. 

> I get the impression we're trying to make Helm commands be able to do the same thing as kubectl. 

I think the idea was more to adopt the way kubectl "applies" changes to the cluster and adopt what seems to be working well.  

> When people think of using flags they often think of it applying a feature. If you apply 2 flags it applies 2 features. But, in this case applying --force changes the first feature. That doesn't feel like a great UX to me.

Totally agree, if users don't need to think about which flags to apply in standard use cases, that would be great UX from my point of view. This is just another aspect favouring an annotation based API, which would shift the decision what is save to re-create to chart developers, as proposed in https://github.com/helm/helm/issues/7082#issuecomment-687752364.

dastrobu:
Install times in huge charts (and hence downtime) is of concern as als noted in https://github.com/helm/helm/pull/8946#issuecomment-750318991. #7763 is about speeding up install times. It would not be a solution to the issue of this PR but could mitigate the situation for some users.

knrt10:
We have rebased this PR solving merge conflicts on top of `v3.5.1`. You can find the reference here https://github.com/kinvolk/helm/tree/knrt10/v3.5.1-rollback

jkroepke:
@knrt10 do you want to create a new PR here?

invidian:
>@knrt10 do you want to create a new PR here?

@jkroepke created #9836.

jkroepke:
@invidian thats nice. One question here. The helm maintainer spents already a lot of time to review the code here. Did you do just a rebase of the existing code or does your new pull request already gain some benefits from the existing reviews here?

invidian:
> @invidian thats nice. One question here. The helm maintainer spents already a lot of time to review the code here. Did you do just a rebase of the existing code or does your new pull request already gain some benefits from the existing reviews here?

I only did a rebase. I don't see any review comments here and I don't really have time to go trough all the 70+ comments here and addressing them. I only published the rebase efforts, so they don't get lost. Perhaps resetting number of comments on the PR will also help to get it merged? Or maybe maintainers would have time to look and push some fixes to it if it's not significant?

bacongobbler:
> I don't see any review comments here

Four of the current Helm core maintainers have reviewed the code and weighed in here.

> and I don't really have time to go trough all the 70+ comments here and addressing them.

Part of the review process is to discuss a feature's motivation, explain its technical details, and describe how to test, maintain, and teach this new feature. It is important to read and consider those comments in future work. If you're going to ask maintainers to review your code, please be respectful of other's time and make sure to read prior feedback - core maintainers and community members alike.

I'm going to close your PR because a rebase is not going to fix the issues laid out in earlier discussions.

jkroepke:
@bacongobbler 

I would recommend to close this pull request, too. This is PR is staled for almost one year.

bacongobbler:
@dastrobu has expressed continued interest in this matter. For that reason I'm happy to keep it open for further discussion. Should he decide he's no longer interested in pursuing this feature, he can close this PR.

jkroepke:
@bacongobbler I see, but there is no open review anymore, I saw that @mattfarina (https://github.com/helm/helm/pull/7431#issuecomment-747667696) and you (https://github.com/helm/helm/pull/7431#issuecomment-666735299) do a general non in-code review and @dastrobu wrote answers to the review points here. 

By going through the comments, I would agree. I do not see any opened/unanswered reviews here or I could not see them, too. Maybe I do not understand them.

A summary the opened key point would help some motivated people, too.

dastrobu:
I am still motivated to work on a feature, solving #7082. As @jkroepke pointed out, code review is not the issue with this PR. There are conceptual concerns (https://github.com/helm/helm/pull/7431#issuecomment-666735299), which is why I proposed an alternative approach in #9174. Unfortunately, I did not receive much feedback on this alternative yet, which is why I paused working on this as long as there is no clear path, to get one of the proposals merged.

joejulian:
Are you still interested, @dastrobu, another year later?  I think I agree that an annotation seems more "correct" (for some definition of correct). Perhaps the very idea of what should be an annotation and what should be a flag should be decided in a hip.

joejulian:
Closing this as abandoned.

PR #7432, opened 2020-01-18
fix(helm): sort hooks by kind for equal weight

### LINKED ISSUE #7416: respect sort order for equal weight hooks
Helm is currently sorting hooks by weight and then by name. This is nice, however, for hooks with equal weight I would expect that the same sorting by kind is done before executing the hooks with equal weight. 
Consider e.g. the following example. A hook should run a job, but the job requires a certain service account to do that. So create a template with the service account and the jobs annotated as helm hook with some weight. Then it would make sense to sort the hooks by kind, like done for normal resources, to make sure that the service account is created before the job. 
I think this would resource handling make more consistent, if it is done the same way for hooks and non-hooks.

To workaround this, one can manage the order by carefully choosing weights. For the example above that would mean that the service account must get a smaller weight than the job.

I am not sure if it is considered a feature or a bug, but I can provide a pull request to fix this, if the proposed solution is accepted.

Output of `helm version`:

dastrobu:
Use the same install order for hooks as for normal resources (non-hooks) for hooks with equal weight.
This makes resource handling more consistent and helps, when there are hook consisting of several resources like e.g. a service account and a job using this service account.

The sort functions are changed from an in place search to an out of place sort to avoid inout parameters.

Closes #7416.

### DIFF
--- a/pkg/action/hooks.go
+++ b/pkg/action/hooks.go
@@ -38,7 +38,8 @@ func (cfg *Configuration) execHook(rl *release.Release, hook release.HookEvent, 
 		}
 	}
 
-	sort.Sort(hookByWeight(executingHooks))
+	// hooke are pre-ordered by kind, so keep order stable
+	sort.Stable(hookByWeight(executingHooks))
 
 	for _, h := range executingHooks {
 		// Set default delete policy to before-hook-creation
--- a/pkg/releaseutil/kind_sorter.go
+++ b/pkg/releaseutil/kind_sorter.go
@@ -16,7 +16,11 @@ limitations under the License.
 
 package releaseutil
 
-import "sort"
+import (
+	"sort"
+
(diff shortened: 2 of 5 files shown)


mattfarina:
@dastrobu Thanks for the PR. CI found issues with the formatting of the code change. Can you please fix it.

dastrobu:
@mattfarina fixed it (I will look at the build myself next time).

bacongobbler (approved):

bacongobbler:
Thanks for working on this @dastrobu. After reviewing this with @mattfarina, we found some cases where we could improve the code to prevent extra memory allocations. Specifically, `sortManifestsByKind` did not need to be an out-of-place sort. Since Go passed `manifests` by value, the original order was preserved. I also took the time to DRY up some of the code.

Have a look at #7499 for the successor to this PR.

Thanks!

PR #7435, opened 2020-01-19
fix(test): Remove invalid subcommand in test

marckhouzam:
The 'home' command was removed for v3, so this commit adapts the tests.
Having an invalid subcommand does not cause problems currently, but
will start causing failures on newer versions of Cobra, which complain
on invalid subcommands.


bacongobbler (approved):
nice find! Thanks for fixing this.

PR #7440, opened 2020-01-20
Render the CRDs to output-dir files

mtougeron:
This expands on #7138 (solving  #6930) by writing the files in the `crds` directory to the `--output-dir` or display as appropriate when running `helm template`. 

Signed-off-by: Mike Tougeron <tougeron@adobe.com>


mtougeron:
@technosophos / @bacongobbler Hey folks. Just wanted to check in with you on this PR. It's my first attempt at something for the core helm project and I wanted to make sure I was on the right path for the way I did it. Thanks, Mike

pkg/chart/chart.go
```diff
@@ -117,18 +129,19 @@ func (ch *Chart) AppVersion() string {
 	return ch.Metadata.AppVersion
 }
 
-// CRDs returns a list of File objects in the 'crds/' directory of a Helm chart.
-func (ch *Chart) CRDs() []*File {
-	files := []*File{}
+// CRDs returns a list of CRD objects in the 'crds/' directory of a Helm chart & subcharts
+func (ch *Chart) CRDs() []CRD {
```
technosophos: So technically speakign we cannot break the API for this, and changing the function's return type would break the API.

I would recommend just creating a function named something like `CRDObjects()` and returning the `[]CRD` from that, and figuring another way to keep this function basically as it was.

Then flag this function as DEPRECATED and we can remove it in Helm 4.

mtougeron: Thanks, I didn't realize these APIs were set for other usage. I'll adjust accordingly shortly.

mtougeron: @technosophos how's this update for you?

mtougeron:
@technosophos wanted to follow-up with you on this PR. Any other feedback? Thanks.

technosophos (approved):
This revision looks great! Thanks!

mtougeron:
ah bugger. looks like conflicts showed up since I pinged you. I'll look this afternoon. :(

mtougeron:
@technosophos okay, I resolved the conflicts and the checks are green again.

technosophos:
I think this just needs @bacongobbler 's approval and then we can merge.

bacongobbler (approved):

bacongobbler:
thanks for the refactor as well! nice cleanup.

mtougeron:
Awesome! Thanks for the help. Now we'll be able to upgrade as soon as the next release drops. I'm excited! :)

PR #7466, opened 2020-01-25
fix(comp): Allow zsh completion to handle -n flag

marckhouzam:
I ran into this while adding more tests to the acceptance-testing repo.

When doing zsh completion, the `-n` flag would not be handled properly.
Doing
```
helm -n<TAB>
```
would not add the space after the `-n`.
This was caused by the fact that the `-n` flag was being swallowed by the `echo` command.

To fix this, we use `printf` instead.

This has been regression-tested with the acceptance-testing repo

### DIFF
--- a/cmd/helm/completion.go
+++ b/cmd/helm/completion.go
@@ -139,7 +139,10 @@ __helm_compgen() {
 	fi
 	for w in "${completions[@]}"; do
 		if [[ "${w}" = "$1"* ]]; then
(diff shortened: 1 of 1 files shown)


mattfarina (approved):
@marckhouzam Merge at your convenience.

I tested myself.

marckhouzam:
Thanks @mattfarina!

PR #7472, opened 2020-01-27
Add TLS renegotiation support 

### LINKED ISSUE #7344: Consider supporting TLS re-negotiation
See related Gobug: https://github.com/golang/go/issues/5742

Error: looks like "https:/helm.example.com/pulp/Default_Organization/Library/custom/sde/sde-helm/" is not a valid chart repository or cannot be reached: Get https://helm.example.com/pulp/Default_Organization/Library/custom/sde/sde-helm/index.yaml: local error: tls: no renegotiation

Output of `helm version`:

antevens:
Add flag to support TLS renegotiation for chart repos and refactor chartrepo find code.

Fixes https://github.com/helm/helm/issues/7344

add a chart repository

Usage:
  helm repo add [NAME] [URL] [flags]

Flags:
      --ca-file string       verify certificates of HTTPS-enabled servers using this CA bundle
      --cert-file string     identify HTTPS client using this SSL certificate file
  -h, --help                 help for add
      --key-file string      identify HTTPS client using this SSL key file
      --no-update            raise error if repo is already registered
      --password string      chart repository password
      --renegotiate string   TLS renegotiation strategy, valid options include Never, OnceAsClient and FreelyAsClient (default "Never")
      --username string      chart repository username

### DIFF
--- a/cmd/helm/flags.go
+++ b/cmd/helm/flags.go
@@ -47,6 +47,7 @@ func addChartPathOptionsFlags(f *pflag.FlagSet, c *action.ChartPathOptions) {
 	f.StringVar(&c.CertFile, "cert-file", "", "identify HTTPS client using this SSL certificate file")
 	f.StringVar(&c.KeyFile, "key-file", "", "identify HTTPS client using this SSL key file")
 	f.StringVar(&c.CaFile, "ca-file", "", "verify certificates of HTTPS-enabled servers using this CA bundle")
+	f.StringVar(&c.Renegotiate, "renegotiate", "never", "TLS renegotiation strategy, valid options include 'never', 'once' and 'freely'")
 }
 
 // bindOutputFlag will add the output flag to the given command and bind the
--- a/cmd/helm/repo_add.go
+++ b/cmd/helm/repo_add.go
@@ -49,6 +49,8 @@ type repoAddOptions struct {
 
 	repoFile  string
 	repoCache string
+
+	renegotiate string
 }
 
 func newRepoAddCmd(out io.Writer) *cobra.Command {
@@ -75,6 +77,7 @@ func newRepoAddCmd(out io.Writer) *cobra.Command {
 	f.StringVar(&o.certFile, "cert-file", "", "identify HTTPS client using this SSL certificate file")
 	f.StringVar(&o.keyFile, "key-file", "", "identify HTTPS client using this SSL key file")
 	f.StringVar(&o.caFile, "ca-file", "", "verify certificates of HTTPS-enabled servers using this CA bundle")
+	f.StringVar(&o.renegotiate, "renegotiate", "never", "TLS renegotiation strategy, valid options include 'never', 'once' and 'freely'")
 
 	return cmd
 }
@@ -113,13 +116,14 @@ func (o *repoAddOptions) run(out io.Writer) error {
 	}
 
 	c := repo.Entry{
-		Name:     o.name,
-		URL:      o.url,
-		Username: o.username,
-		Password: o.password,
-		CertFile: o.certFile,
-		KeyFile:  o.keyFile,
-		CAFile:   o.caFile,
+		Name:        o.name,
+		URL:         o.url,
+		Username:    o.username,
+		Password:    o.password,
+		CertFile:    o.certFile,
+		KeyFile:     o.keyFile,
+		CAFile:      o.caFile,
+		Renegotiate: o.renegotiate,
 	}
 
 	r, err := repo.NewChartRepository(&c, getter.All(settings))
--- a/internal/tlsutil/tls.go
+++ b/internal/tlsutil/tls.go
@@ -20,12 +20,47 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"io/ioutil"
+	"strings"
 
 	"github.com/pkg/errors"
 )
 
-// NewClientTLS returns tls.Config appropriate for client auth.
+type TLSRenegotiationStrategy string
+
+const (
+	RenegotiateNever          string = "never"
+	RenegotiateOnceAsClient   string = "once"
+	RenegotiateFreelyAsClient string = "freely"
+)
+
+// Validate and return TLS renegotiation settings
+func GetRenegotiation(option string) (strategy tls.RenegotiationSupport, err error) {
+	option = strings.ToLower(option)
+	switch option {
+	// RenegotiateNever disables renegotiation.
+	case RenegotiateNever:
+		return tls.RenegotiateNever, nil
+	// RenegotiateOnceAsClient allows a remote server to request
+	// renegotiation once per connection.
+	case RenegotiateOnceAsClient:
+		return tls.RenegotiateOnceAsClient, nil
+	// RenegotiateFreelyAsClient allows a remote server to repeatedly
+	// request renegotiation.
+	case RenegotiateFreelyAsClient:
+		return tls.RenegotiateFreelyAsClient, nil
+	}
+	return tls.RenegotiateNever, errors.New("invalid TLS Renegotiation strategy selected")
+}
+
+// NewClientTLS returns tls.Config compatible with client auth but without
+// support for TLS renegotiation
 func NewClientTLS(certFile, keyFile, caFile string) (*tls.Config, error) {
+	return NewClientTLSWithRenegotiate(certFile, keyFile, caFile, RenegotiateNever)
+}
+
+// NewClientTLS returns tls.Config compatible with both client auth and TLS
+// Renegotiation
+func NewClientTLSWithRenegotiate(certFile, keyFile, caFile string, renegotiate string) (*tls.Config, error) {
 	config := tls.Config{}
 
 	if certFile != "" && keyFile != "" {
@@ -43,6 +78,12 @@ func NewClientTLS(certFile, keyFile, caFile string) (*tls.Config, error) {
 		}
 		config.RootCAs = cp
 	}
+
+	renegotiation, err := GetRenegotiation(renegotiate)
+	if err != nil {
+		return nil, errors.Wrapf(err, "valid options include '%s', '%s' and '%s'", RenegotiateNever, RenegotiateOnceAsClient, RenegotiateFreelyAsClient)
+	}
+	config.Renegotiation = renegotiation
 
 	return &config, nil
 }
--- a/internal/tlsutil/tlsutil_test.go
+++ b/internal/tlsutil/tlsutil_test.go
@@ -61,12 +61,12 @@ func testfile(t *testing.T, file string) (path string) {
 	return path
 }
 
-func TestNewClientTLS(t *testing.T) {
+func TestNewClientTLSWithRenegotiate(t *testing.T) {
 	certFile := testfile(t, testCertFile)
 	keyFile := testfile(t, testKeyFile)
 	caCertFile := testfile(t, testCaCertFile)
 
-	cfg, err := NewClientTLS(certFile, keyFile, caCertFile)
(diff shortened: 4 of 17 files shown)


cmd/helm/repo_add.go
```diff
@@ -75,6 +77,7 @@ func newRepoAddCmd(out io.Writer) *cobra.Command {
 	f.StringVar(&o.certFile, "cert-file", "", "identify HTTPS client using this SSL certificate file")
 	f.StringVar(&o.keyFile, "key-file", "", "identify HTTPS client using this SSL key file")
 	f.StringVar(&o.caFile, "ca-file", "", "verify certificates of HTTPS-enabled servers using this CA bundle")
+	f.StringVar(&o.renegotiate, "renegotiate", "Never", "TLS renegotiation strategy, valid options include Never, OnceAsClient and FreelyAsClient")
```
bacongobbler: can we make these values case-insensitive?

internal/tlsutil/tls.go
```diff
@@ -24,8 +24,27 @@ import (
 	"github.com/pkg/errors"
 )
 
-// NewClientTLS returns tls.Config appropriate for client auth.
-func NewClientTLS(certFile, keyFile, caFile string) (*tls.Config, error) {
+// Validate and return TLS renegotiation settings
+func GetRenegotiation(option string) (strategy tls.RenegotiationSupport, err error) {
+	switch option {
```
bacongobbler: same here RE: making these fields case-insensitive. In general, if a flag accepts an enumerated string, it's a good idea to cast user flag input to lowercase so that `NEVER`, `never` and `Never` are all valid input.

cmd/helm/flags.go
```diff
@@ -47,6 +47,7 @@ func addChartPathOptionsFlags(f *pflag.FlagSet, c *action.ChartPathOptions) {
 	f.StringVar(&c.CertFile, "cert-file", "", "identify HTTPS client using this SSL certificate file")
 	f.StringVar(&c.KeyFile, "key-file", "", "identify HTTPS client using this SSL key file")
 	f.StringVar(&c.CaFile, "ca-file", "", "verify certificates of HTTPS-enabled servers using this CA bundle")
+	f.StringVar(&c.Renegotiate, "renegotiate", "", "TLS renegotiation strategy, valid options include Never(default), OnceAsClient and FreelyAsClient")
```
bacongobbler: the default here appears to be the empty string. should it be "never" instead, similar to repo_add.go?

internal/tlsutil/tls.go
```diff
@@ -24,8 +24,27 @@ import (
...
+	// RenegotiateNever disables renegotiation.
+	case "Never":
```
bacongobbler: a good idea might be to turn these cases into constant values. That way, we avoid a case where a typo may be used in the codebase. It also keeps the code more DRY.

`type ShowOutputFormat` is a good example to follow here. See https://github.com/helm/helm/blob/83847cc89980185ccbbd98747d7dc725e1e4910d/pkg/action/show.go#L30-L37

pkg/repo/chartrepo.go
```diff
@@ -182,28 +184,29 @@ func (r *ChartRepository) generateIndex() error {
 
 // FindChartInRepoURL finds chart in chart repository pointed by repoURL
 // without adding repo to repositories
-func FindChartInRepoURL(repoURL, chartName, chartVersion, certFile, keyFile, caFile string, getters getter.Providers) (string, error) {
-	return FindChartInAuthRepoURL(repoURL, "", "", chartName, chartVersion, certFile, keyFile, caFile, getters)
+func FindChartInRepoURL(repoURL, chartName, chartVersion, certFile, keyFile, caFile string, renegotiate string, getters getter.Providers) (string, error) {
```
bacongobbler: Because this change modifies the function signature, this would be considered a backwards compatibility breaker.

A more in-depth refactor may be necessary here for this area of the code base. Perhaps something like the [builder pattern](https://en.wikipedia.org/wiki/Builder_pattern) may be useful here. This is how `getter`s can introduce new features like `WithTLSRenegotiate` without breaking existing use cases.

What do you think?

pkg/repo/chartrepo.go
```diff
@@ -182,28 +184,29 @@ func (r *ChartRepository) generateIndex() error {
...
+	return FindChartInAuthRepoURL(repoURL, "", "", chartName, chartVersion, certFile, keyFile, caFile, renegotiate, getters)
 }
 
 // FindChartInAuthRepoURL finds chart in chart repository pointed by repoURL
 // without adding repo to repositories, like FindChartInRepoURL,
 // but it also receives credentials for the chart repository.
-func FindChartInAuthRepoURL(repoURL, username, password, chartName, chartVersion, certFile, keyFile, caFile string, getters getter.Providers) (string, error) {
+func FindChartInAuthRepoURL(repoURL, username, password, chartName, chartVersion, certFile, keyFile, caFile string, renegotiate string, getters getter.Providers) (string, error) {
```
bacongobbler: same here

bacongobbler (changes requested):

cmd/helm/repo_add.go
antevens: Changed these to being case-insensitive, simplified and used lower case as default and for examples.

internal/tlsutil/tls.go
antevens: see above

cmd/helm/flags.go
antevens: Yes, done

internal/tlsutil/tls.go
antevens: Sorry if my code isn't that great, I'm very new to Golang and I'm still catching up on the best practices, patterns and style.

pkg/repo/chartrepo.go
antevens: I refactored the whole code in a way that I think makes sense but again I'm very new to Golang so this might not be what you meant with your comment, I'm happy to take director or get feedback.

pkg/repo/chartrepo.go
antevens: see above

antevens:
Hey @bacongobbler, sorry to trouble you but I was wondering if there was any chance you have the bandwidth to comment on the changes I made and if they address your concerns.

bacongobbler:
Thanks for the ping. Given the size of this PR it may be a while before I can find the time to review and test this PR. Just letting you know ahead of time. Thanks!

jdolitsky (changes requested):
I think we probably do not want to change the signature of the public method. Maybe instead add a new method "NewClientTLSWithRenegotiateSetting"

internal/tlsutil/tls.go
```diff
@@ -20,12 +20,41 @@ import (
...
+	// request renegotiation.
+	case RenegotiateFreelyAsClient:
+		return tls.RenegotiateFreelyAsClient, nil
+	}
+	return tls.RenegotiateNever, errors.New("invalid TLS Renegotiation strategy selected")
+}
+
+// NewClientTLS returns tls.Config compatible with both client auth and TLS
+// Renegotiation
+func NewClientTLS(certFile, keyFile, caFile string, renegotiate string) (*tls.Config, error) {
```
jdolitsky: ^ referring to this

antevens: Thanks @jdolitsky, I've made changes to that section as suggested and the method signature is now unchanged with a new method called NewClientTLSWithRenegotiate.

If you have more suggestions, tips or advice please don't hesitate to comment.

internal/tlsutil/tlsutil_test.go
```diff
@@ -61,12 +61,12 @@ func testfile(t *testing.T, file string) (path string) {
 	return path
 }
 
-func TestNewClientTLS(t *testing.T) {
```
bacongobbler: Would it be worth restoring the tests for `NewClientTLS` as well? That way we can capture that we are setting renegotiation to `RenegotiateNever` when `NewClientTLS` is called.

bacongobbler (commented):
This is shaping up really well. I think we're really close here!

bacongobbler:
A couple of quick notes here as well to expedite the review:

- looks like a few commits are missing the DCO sign-off. Would you mind amending your commits and sign off on them? More info here: https://github.com/helm/helm/pull/7472/checks?check_run_id=485556727
- The tests are failing because the code has not been formatted. You should be able to run `go fmt` on the code and it should reformat itself.

Let us know if you run into any issues.

internal/tlsutil/tls.go
```diff
@@ -20,12 +20,47 @@ import (
...
+	// RenegotiateOnceAsClient allows a remote server to request
+	// renegotiation once per connection.
+	case RenegotiateOnceAsClient:
+		return tls.RenegotiateOnceAsClient, nil
+	// RenegotiateFreelyAsClient allows a remote server to repeatedly
+	// request renegotiation.
+	case RenegotiateFreelyAsClient:
+		return tls.RenegotiateFreelyAsClient, nil
+	}
+	return tls.RenegotiateNever, errors.New("invalid TLS Renegotiation strategy selected")
```
bacongobbler: Can we add test cases to check for these case statements? e.g. if a user passes an invalid string like `GetRenegotiation("blah")`, an error should be returned.

internal/tlsutil/tls.go
```diff
@@ -20,12 +20,47 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"io/ioutil"
+	"strings"
 
 	"github.com/pkg/errors"
 )
 
-// NewClientTLS returns tls.Config appropriate for client auth.
+type TLSRenegotiationStrategy string
```
bacongobbler: Why are we defining a type, then never using it? Can we make `GetRenegotiation` accept a type `TLSRenegotiationStrategy`, and use this type for the renegotiation strategies `RenegotiateNever`, `RenegotiateOnceAsClient` and `RenegotiateFreelyAsClient`?

Here's a good reference on applying enum types: https://www.ribice.ba/golang-enums/

bacongobbler (changes requested):

internal/tlsutil/tlsutil_test.go
antevens: I would really have liked to actually test the re-negotiation but I don't know of any public place or test (badssl.com) that has re-negotiation enabled. Given that I think the latest change is the best we can do for now.

`

internal/tlsutil/tls.go
antevens: I think that was my original thought but that would mean GetRenegotiation wouldn't accept a case insensitive string for the strategy. Should I implement that and move the  ToLower and type cast to somewhere else?

internal/tlsutil/tlsutil_test.go
antevens: Done

internal/tlsutil/tls.go
antevens: The implementation depends on the conversation below, will add the test once the preferred argument type has been determined.

internal/tlsutil/tls.go
bacongobbler: yeah. The calls to `ToLower` and the type cast can be moved over to `cmd`, after parsing the flags.

In fact, you could introduce a function that parses a string and returns a `TLSRenegotiationStrategy`. Many logging packages use this to parse strings like "debug" or "info" for log levels.

logrus is a good example that demonstrates this pattern. Have a look at https://github.com/sirupsen/logrus/blob/53000c4c0f981521df1bada12a9e9ea20728fb62/logrus.go#L12-L112

joejulian:
Is this abandoned?

antevens:
> Is this abandoned?

Yes, I don't think there was enough interest to ever merge this.

PR #7474, opened 2020-01-27
fix(lookup_func): do not return error when object is not found

### LINKED ISSUE #7473: a failed `lookup` removes all rendered output
After testing the `lookup` function introduced in #6752, I found calling `lookup` in a template caused the rendered output to be empty.

Here's what I used to test. After creating a new chart with `helm create foo`, I modified `foo/templates/deployment.yaml` to add a new label that performs a lookup:

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: {{ include "foo.fullname" . }}
  labels:
    foo: {{ (lookup "v1" "Deployment" .Release.Namespace "nginx").metadata.name }}
...
```

And then created a deployment to match that lookup:

```
><> kubectl run nginx --image=nginx --replicas=5
deployment.apps/nginx created
```

No output is rendered.

```
><> ./bin/helm template --validate ./foo/

><> ./bin/helm version
version.BuildInfo{Version:"v3.0+unreleased", GitCommit:"9c680e604be8d516960eecbbc3718ace91643cea", GitTreeState:"dirty", GoVersion:"go1.13.6"}
```

Calling `helm install` fails to render, as no objects were rendered.

```
><> ./bin/helm install foo ./foo/
Error: no objects visited
```

I will look further into this, but @raffaelespazzoli, can you please look further into this and see where the issue may be stemming from?

bacongobbler:
closes #7473.

### DIFF
--- a/pkg/engine/lookup_func.go
+++ b/pkg/engine/lookup_func.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 
 	"github.com/pkg/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/discovery"
@@ -30,6 +31,8 @@ import (
 
 type lookupFunc = func(apiversion string, resource string, namespace string, name string) (map[string]interface{}, error)
 
+// NewLookupFunction returns a function for looking up objects in the cluster. If the resource does not exist, no error
+// is raised.
 func NewLookupFunction(config *rest.Config) lookupFunc {
 	return func(apiversion string, resource string, namespace string, name string) (map[string]interface{}, error) {
 		var client dynamic.ResourceInterface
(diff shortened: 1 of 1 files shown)


raffaelespazzoli:
if that is the logic we want to apply, then we should apply it also at line 62.

bacongobbler:
good call. Fixed.

mattfarina (approved):

mattfarina:
Would be nice to have tests on this but not blocking on them.

ichekrygin:
@bacongobbler - continue from https://github.com/helm/helm/issues/7473#issuecomment-580418891

```
⋊> ~/g/s/g/h/helm on 1b1d6bba ⨯ helm version
version.BuildInfo{Version:"v3.0+unreleased", GitCommit:"1b1d6bba9cec81a6bbc77dd948779a45bc8a63c8", GitTreeState:"dirty", GoVersion:"go1.12.12"}
⋊> ~/g/s/g/h/helm on 1b1d6bba ⨯ helm install foo ./foo -n bar
Error: create: failed to create: namespaces "bar" not found
```

PR #7491, opened 2020-01-30
Add namespace support for memory driver

marckhouzam:
This PR adds namespace support for the memory driver.  This allows to write Go tests that use different namespaces.

The change is a bit tricky and I need the opinion of other maintainers:
1. The `template` command uses the memory driver, so I made an adjustment in the `pkg/action/install.go` file to support namespace, but I'm not very sure about that change
1. I've kept the API syntax backwards-compatible (unless I missed something)
1. I've believe I've kept backwards-compatibility for tests success.  The idea is that if someone wrote some tests out there, this change should not break them.
1. is this a feature or a bug fix?

### DIFF
--- a/cmd/helm/helm_test.go
+++ b/cmd/helm/helm_test.go
@@ -143,6 +143,9 @@ func executeActionCommandC(store *storage.Storage, cmd string) (*cobra.Command, 
 	root.SetOutput(buf)
 	root.SetArgs(args)
 
+	if mem, ok := store.Driver.(*driver.Memory); ok {
+		mem.SetNamespace(settings.Namespace())
+	}
 	c, err := root.ExecuteC()
 
 	return c, buf.String(), err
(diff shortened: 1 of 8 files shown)


marckhouzam:
> do we even want this fix, since it may cause backwards-compatibility issues?

My motivation for this PR is that I aim to write acceptance tests using the memory driver and that I'll need it to support namespaces for the tests to be complete.  So I'm hoping we can find a satisfactory way forward for this PR.

marckhouzam:
I figured out a very simple way to make this change backwards-compatible.
This latest version should be completely backwards-compatible both in syntax and in not making existing external tests fail.  In fact, I did not need to adapt any of Helm's test, they all kept passing.  
The test modifications of this PR are new tests.

bacongobbler (approved):
LGTM!

marckhouzam:
Thanks @bacongobbler

PR #7499, opened 2020-01-30
ref(helm): sort hooks by kind

### LINKED ISSUE #7416: respect sort order for equal weight hooks

bacongobbler:
supercedes #7432, DRYing up a ton of code and reverting to an in-place sort for `sortManifestsByKind`, preventing a ton of extra memory allocation.

`sortManifestsByKind` passes `manifests` by value, so it receives a copy of the input. Order is preserved. I added a test to confirm this.

closes #7416

### DIFF
--- a/pkg/releaseutil/kind_sorter.go
+++ b/pkg/releaseutil/kind_sorter.go
@@ -16,7 +16,11 @@ limitations under the License.
 
 package releaseutil
 
-import "sort"
+import (
+	"sort"
+
+	"helm.sh/helm/v3/pkg/release"
+)
 
 // KindSortOrder is an ordering of Kinds.
 type KindSortOrder []string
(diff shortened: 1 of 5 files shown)


dastrobu:
Nice cleanup @bacongobbler! 
I would suggest to rename the functions
```
manifestsSortedByKind -> sortManifestsByKind
hooksSortedByKind -> sortHooksByKind
```
since `sorted` usually refers to an out of place sort and `sort` to in place sort. This might make the behavior of the functions clearer from the function names.

pkg/releaseutil/kind_sorter.go
```diff
@@ -99,46 +103,42 @@ var UninstallOrder KindSortOrder = []string{
...
-// sortByKind does an in-place sort of manifests by Kind.
+// sort manifests by kind.
 //
 // Results are sorted by 'ordering', keeping order of items with equal kind/priority
-func sortByKind(manifests []Manifest, ordering KindSortOrder) []Manifest {
-	ks := newKindSorter(manifests, ordering)
-	sort.Stable(ks)
-	return ks.manifests
-}
+func manifestsSortedByKind(manifests []Manifest, ordering KindSortOrder) []Manifest {
```
mattfarina: The public function in `manifest_sorter.go` is `SortManifests`. This raises two questions for me...

1. Should the manifest sorting be in that file due to the name?
2. Should the function name be `sortManifestsByKind` to follow a similar naming pattern?

bacongobbler: done

mattfarina (approved):
+1

PR #7503, opened 2020-01-31
feat(template): Allow template output to use release name

### LINKED ISSUE #5664: `helm template --output-dir x --name release` renders to colliding path
Output of `helm version`: `Client: &version.Version{SemVer:"v2.13.1", GitCommit:"618447cbf203d147601b4b9bd7f8c37a5d39fbb4", GitTreeState:"clean"}`

For example, say I want to render two releases of nginx, mine1 and mine2:
```
% helm template -n mine1 --output-dir out docs/examples/nginx
wrote out/nginx/templates/pre-install-secret.yaml
wrote out/nginx/templates/configmap.yaml
wrote out/nginx/templates/service.yaml
wrote out/nginx/templates/service-test.yaml
wrote out/nginx/templates/deployment.yaml
wrote out/nginx/templates/post-install-job.yaml
% helm template -n mine2 --output-dir out docs/examples/nginx
wrote out/nginx/templates/pre-install-secret.yaml
wrote out/nginx/templates/configmap.yaml
wrote out/nginx/templates/service.yaml
wrote out/nginx/templates/service-test.yaml
wrote out/nginx/templates/deployment.yaml
wrote out/nginx/templates/post-install-job.yaml
```
The second invocation clobbers the first. I would have expected `out/nginx/templates` to include the release name to avoid such a collision.

hickeyma:
`helm template --output-dir` command uses the chart name only when writing templates to disk. This causes collision in the path when  running the command for different releases of the same chart. This feature allows the release name to be used to avoid second invocation clobbering the previous output.

New path format: `outputDir/<releaseName>/<chart>/templates`

Closes #5664

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -424,7 +424,7 @@ func (i *Install) replaceRelease(rel *release.Release) error {
 }
 
 // renderResources renders the templates in a chart
-func (c *Configuration) renderResources(ch *chart.Chart, values chartutil.Values, releaseName string, outputDir string, subNotes bool, useReleaseName bool) ([]*release.Hook, *bytes.Buffer, string, error) {
+func (c *Configuration) renderResources(ch *chart.Chart, values chartutil.Values, releaseName string, outputDir string, subNotes, useReleaseName bool) ([]*release.Hook, *bytes.Buffer, string, error) {
 	hs := []*release.Hook{}
 	b := bytes.NewBuffer(nil)
 
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -89,6 +89,9 @@ type Install struct {
 	APIVersions chartutil.VersionSet
 	// Used by helm template to render charts with .Release.IsUpgrade. Ignored if Dry-Run is false
 	IsUpgrade bool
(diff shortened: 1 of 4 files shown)


pkg/action/install.go
```diff
@@ -421,7 +424,7 @@ func (i *Install) replaceRelease(rel *release.Release) error {
 }
 
 // renderResources renders the templates in a chart
-func (c *Configuration) renderResources(ch *chart.Chart, values chartutil.Values, outputDir string, subNotes bool) ([]*release.Hook, *bytes.Buffer, string, error) {
+func (c *Configuration) renderResources(ch *chart.Chart, values chartutil.Values, releaseName string, outputDir string, subNotes bool, useReleaseName bool) ([]*release.Hook, *bytes.Buffer, string, error) {
```
thomastaylor312: Nit: consider short handing the last 2 args to `subNotes, useReleaseName bool`

thomastaylor312 (approved):

pkg/action/install.go
hickeyma: Updated

hickeyma:
Thanks for review @thomastaylor312

PR #7504, opened 2020-01-31
Add EnvSettings.SetNamespace method

diversario:
I would like to be able to set the namespace for install/upgrade/etc actions programmatically (that is, not via CLI options) in my code that uses Helm's Go API (including the `cli` package for some CLI options and for action configuration).

At the moment, the workaround I found is to pass the desired namespace to `action.Configuration.Init` [like so](https://github.com/helm/helm/blob/a6b2c9e2126753f6f94df231e89b2153c2862764/cmd/helm/list.go#L74), but I believe being able to set namespace on `EnvSettings` would be the proper way.

### DIFF
--- a/pkg/cli/environment.go
+++ b/pkg/cli/environment.go
@@ -121,6 +121,11 @@ func (s *EnvSettings) Namespace() string {
 	return "default"
 }
 
+//SetNamespace sets the namespace
+func (s *EnvSettings) SetNamespace(ns string) {
+	s.namespace = ns
+}
+
 //RESTClientGetter gets the kubeconfig from EnvSettings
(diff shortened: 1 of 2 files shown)


diversario:
I was wrong, using the workaround I mentioned does not work – if actual templates do not set `metadata.namespace` they're deployed into `default` namespace, while the release record is put into the custom namespace.

bacongobbler:
have you had a chance to look at https://github.com/helm/helm/pull/7373?

diversario:
Hmm. So if I set it on the client and _not_ set it on the `EnvSettings` it would work?.. 🤔

PR #7508, opened 2020-02-01
fix(test) use newly created index instead of ignoring it

karuppiah7890:
This fixes a possible issue in the test for repo index

I think the test code ignores the index in one place and uses the old index values to do wrong assertions

Please do check and provide comments. Looks like it was intentionally ignored, but I don't get why. So, felt it's an issue

### DIFF
--- a/cmd/helm/repo_index_test.go
+++ b/cmd/helm/repo_index_test.go
@@ -119,7 +119,7 @@ func TestRepoIndexCmd(t *testing.T) {
 		t.Error(err)
 	}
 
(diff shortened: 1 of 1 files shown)


marckhouzam:
Nice find!  It looks to me like this was a copy/paste error.

I'd like a maintainer with more experience to confirm, but I think this looks good.

bacongobbler (approved):

bacongobbler:
probably just an oversight. LGTM

karuppiah7890:
Makes sense 😄

PR #7510, opened 2020-02-02
[WIP] Support index.json in addition to index.yaml in repositories

### LINKED ISSUE #6155: Support index.json in addition to index.yaml in repositories

karuppiah7890:
Fixes #6155 

- [x] feat(repo_index): generate index.json too for repo index command

### DIFF
--- a/cmd/helm/repo_add_test.go
+++ b/cmd/helm/repo_add_test.go
@@ -104,6 +104,67 @@ func TestRepoAdd(t *testing.T) {
 	}
 }
 
+func TestRepoAdd_IndexJSON(t *testing.T) {
+	ts, err := repotest.NewTempServer("testdata/testserver/*.*")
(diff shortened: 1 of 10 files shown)


karuppiah7890:
Since it's been months now, and I haven't worked much on the PR, it's fair to open it for others.

Issue #6844, opened 2019-10-30
refactor: consolidate the --debug and -v flags together


bacongobbler:
A use case was described in https://github.com/helm/helm/issues/6618.

"`helm --debug` is too chatty for me. I'd like to see something more human-readable."

-v is currently constrained to the Kubernetes client's chattiness, where `helm -v 10` will dump raw HTTP request methods, payloads, responses, etc. There may be a way where we can tie into this flag such that `helm --debug` is equivalent to a `helm -v 5`. Users could then use `helm -v 1` for slightly more chatty messages than the current `helm --debug` flag.

shubhambhattar:
How to use `-v` flag? Is it documented anywhere? I am just trying with values from `1` to `10`. Or I can't even figure out if higher numbers are more verbose of the range of numbers which it'll accept?

marckhouzam:
@shubhambhattar I believe level 10 will print the most output.

The `-v` flag is currently only used to increase debug printout when helm talks to kubernetes.  So if you want to see it in action you can try something like `helm list --all-namespaces -v 10`

PR #7523, opened 2020-02-04
fix(sympath): add option to silence symlink log warning

griffin:
~This log line is printed if you have a system link in your chart dir. We have some tooling that sandboxs chart packaging with symlinks, leading to this being printed frequently on engineer's machines. The message isn't clear and is confusing if it is an error or not. Symlinks seem to be supported in helm, so there is no reason to print this debugging message.~

Edit:

Took a look at the work required to support #6844 and seems like to much of a refactor and interface change to take on within a reasonable timeframe. Since the goal seems to be to use the `klog` logging library I switched the logging package in this module using a verbosity of 0, which should line up with the [logging standards](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md) for the kubernetes project to always output. If a user wants to disable the logging they can use `helm -v -1`.

Signed-off-by: Griffin Dunn <griffin.dunn@datadoghq.com>

### DIFF
--- a/internal/sympath/walk.go
+++ b/internal/sympath/walk.go
@@ -26,6 +26,7 @@ import (
 	"sort"
 
 	"github.com/pkg/errors"
+	"k8s.io/klog"
 )
 
 // Walk walks the file tree rooted at root, calling walkFn for each file or directory
@@ -70,6 +71,7 @@ func symwalk(path string, info os.FileInfo, walkFn filepath.WalkFunc) error {
 		if err != nil {
 			return errors.Wrapf(err, "error evaluating symlink %s", path)
 		}
+		klog.V(0).Infof("found symbolic link in path: %s resolves to %s", path, resolved)
 		if info, err = os.Lstat(resolved); err != nil {
 			return err
 		}
--- a/internal/sympath/walk.go
+++ b/internal/sympath/walk.go
@@ -21,7 +21,6 @@ limitations under the License.
 package sympath
 
 import (
-	"log"
 	"os"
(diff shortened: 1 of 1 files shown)


bacongobbler:
Hi @griffin. To give some background context on this change, it was introduced as part of a CVE during the Helm security audit. More information can be found here: https://helm.sh/blog/2019-10-30-helm-symlink-security-notice/

Based on the feedback provided to us by the security team, this warning message should not be removed. Perhaps a better resolution to this may be to work on https://github.com/helm/helm/issues/6844, which then users could strip this log line by using `helm package --log-level error`.

What do you think?

griffin:
@bacongobbler Updated the description, I think this is a solid middleground.

mattfarina (changes requested):
Reporting on symlinks is for security. The alternative is that everyone needs to opt-in to the use of symlinks when present. That would mean detecting symlinks and then performing user interaction or a flag at runtime.

Someone could use a symlink for nefarious purposes while could end up being a vulnerability.

We can't hide the current symlink messages behind debug messages because security.

rubensayshi:
@bacongobbler @mattfarina 

how about some way to set a whitelist of paths that are accepted to be a symlink within the chart?

it will require adding that to the chart config file and then passing it down into the `Walk`.  
do you think that's too much to deal with this for the few people (like me and @griffin) who are making use of symlinks in their charts?

I'd be willing to put in the work if you at least think that such a change would have chance of being accepted.

PR #7530, opened 2020-02-05
fix(helm): Don't wait for service to be ready when external IP are set

### LINKED ISSUE #7513: Timeout 'helm upgrade --wait'
Output of `helm version`:
```
version.BuildInfo{Version:"v3.0.3", GitCommit:"ac925eb7279f4a6955df663a0128044a8a6b7593", GitTreeState:"clean", GoVersion:"go1.13.6"}
```
Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.2", GitCommit:"59603c6e503c87169aea6106f57b9f242f64df89", GitTreeState:"clean", BuildDate:"2020-01-18T23:30:10Z", GoVersion:"go1.13.5", Compiler:"gc", Platform:"windows/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.2", GitCommit:"59603c6e503c87169aea6106f57b9f242f64df89", GitTreeState:"clean", BuildDate:"2020-01-18T23:22:30Z", GoVersion:"go1.13.5", Compiler:"gc", Platform:"linux/amd64"}
```

I created a helm chart to run in a release pipeline in devops with the flag --wait. The build task times out and with --debug flag it continuously shows:
```
2020-02-03T07:46:29.3015941Z wait.go:182: [debug] Service does not have IP address: ******
```
But when I check the service in k8s it is running and functioning properly.

This is the output of --dry-run:
```
# Source: identityserver/templates/service.yaml
apiVersion: v1
kind: Service
metadata:
  name: identityserver
  labels:
    app: identityserver
    chart: identityserver-0.1.0
    release: identityserver-1580717645
    heritage: Helm
spec:
  type: LoadBalancer
  ports:
    - port: 80
      targetPort: http
      protocol: TCP
      name: http
  externalIPs:
    - 172.16.253.90
  selector:
    app: identityserver
    release: identityserver-1580717645
```

FedeBev:
Resolves #7513
As, [according to the doc](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#servicespec-v1-core), the externalIPs are not managed by k8s; helm should not wait for services  which set at least one externalIPs.

### DIFF
--- a/pkg/kube/wait.go
+++ b/pkg/kube/wait.go
@@ -188,12 +188,24 @@ func (w *waiter) serviceReady(s *corev1.Service) bool {
 	}
 
 	// Make sure the service is not explicitly set to "None" before checking the IP
-	if (s.Spec.ClusterIP != corev1.ClusterIPNone && s.Spec.ClusterIP == "") ||
-		// This checks if the service has a LoadBalancer and that balancer has an Ingress defined
-		(s.Spec.Type == corev1.ServiceTypeLoadBalancer && s.Status.LoadBalancer.Ingress == nil) {
-		w.log("Service does not have IP address: %s/%s", s.GetNamespace(), s.GetName())
+	if s.Spec.ClusterIP != corev1.ClusterIPNone && s.Spec.ClusterIP == "" {
 		return false
 	}
+
+	// This checks if the service has a LoadBalancer and that balancer has an Ingress defined
+	if s.Spec.Type == corev1.ServiceTypeLoadBalancer {
+		// do not wait when at least 1 external IP is set
+		if len(s.Spec.ExternalIPs) > 0 {
+			w.log("Service has externaIPs addresses: %s/%s (%v)", s.GetNamespace(), s.GetName(), s.Spec.ExternalIPs)
+			return true
+		}
+
+		if s.Status.LoadBalancer.Ingress == nil {
+			w.log("Service does not have IP address: %s/%s", s.GetNamespace(), s.GetName())
+			return false
+		}
+	}
+
 	return true
 }
 
--- a/pkg/kube/wait.go
+++ b/pkg/kube/wait.go
@@ -189,6 +189,7 @@ func (w *waiter) serviceReady(s *corev1.Service) bool {
 
 	// Make sure the service is not explicitly set to "None" before checking the IP
 	if s.Spec.ClusterIP != corev1.ClusterIPNone && s.Spec.ClusterIP == "" {
+		w.log("Service does not have IP address: %s/%s", s.GetNamespace(), s.GetName())
 		return false
 	}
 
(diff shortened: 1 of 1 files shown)


bacongobbler:
cc'ing @thomastaylor312 as the resident `--wait` master

pkg/kube/wait.go
```diff
@@ -188,12 +188,24 @@ func (w *waiter) serviceReady(s *corev1.Service) bool {
 	}
 
 	// Make sure the service is not explicitly set to "None" before checking the IP
-	if (s.Spec.ClusterIP != corev1.ClusterIPNone && s.Spec.ClusterIP == "") ||
-		// This checks if the service has a LoadBalancer and that balancer has an Ingress defined
-		(s.Spec.Type == corev1.ServiceTypeLoadBalancer && s.Status.LoadBalancer.Ingress == nil) {
-		w.log("Service does not have IP address: %s/%s", s.GetNamespace(), s.GetName())
+	if s.Spec.ClusterIP != corev1.ClusterIPNone && s.Spec.ClusterIP == "" {
```
thomastaylor312: Can you add back some logging here in this `if` block to explain why it isn't ready?

pkg/kube/wait.go
```diff
@@ -188,12 +188,24 @@ func (w *waiter) serviceReady(s *corev1.Service) bool {
...
 		return false
 	}
+
+	// This checks if the service has a LoadBalancer and that balancer has an Ingress defined
+	if s.Spec.Type == corev1.ServiceTypeLoadBalancer {
+		// do not wait when at least 1 external IP is set
+		if len(s.Spec.ExternalIPs) > 0 {
+			w.log("Service has externaIPs addresses: %s/%s (%v)", s.GetNamespace(), s.GetName(), s.Spec.ExternalIPs)
```
thomastaylor312: This message should be reworded to something like "Service has external IP addresses, marking service as ready" to be more clear that it is being marked as ready because we don't have a way to ascertain if it is

thomastaylor312 (changes requested):
This logic looks great. Thank you for tracking this down and fixing. Just a few changes needed around logging messages and clarity

pkg/kube/wait.go
FedeBev: Sorry, my bad. I'm adding back the original message. 
However I think it can be made a little more specific, for example: `w.log("Service does not have cluster IP address: %s/%s", s.GetNamespace(), s.GetName())`

thomastaylor312: Feel free to make it more specific now that the logic is separated out!

FedeBev: I've Just pushed a commit with a little improvement of the logs.
If for your is fine, for me it's done.

thomastaylor312 (approved):
Looking good. Thanks for the work on this!

PR #7531, opened 2020-02-05
fix(tests): Ignores tarball that will change on dep update

thomastaylor312:
The unit test added to cover #7233 was causing changes to show up in git
when tests were ran. This was due to the dependency build creating a new
tarball. These changes would cause a dirty build when we build our major
versions, so I removed the subchart tarball from git and added the charts
folder for that test chart to the gitignore to avoid any future problems.
Based on all I can see, this shouldn't have any impact on the test itself


thomastaylor312:
I pulled this into 3.1 so we don't end up with an accidentally dirty build

adamreese (approved):

PR #7546, opened 2020-02-07
IsReachable() needs to give detailed error message.

ylvmw:
The IsReachable() function in pkg/kube/client.go should give detailed error message, since the users need to know what the error is when the cluster is unreachable.

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -86,7 +86,7 @@ func (c *Client) IsReachable() error {
 	client, _ := c.Factory.KubernetesClientSet()
 	_, err := client.ServerVersion()
 	if err != nil {
-		return errors.New("Kubernetes cluster unreachable")
+		return fmt.Errorf("Kubernetes cluster unreachable: %s", err.Error())
 	}
 	return nil
 }


pkg/kube/client.go
```diff
@@ -86,7 +86,7 @@ func (c *Client) IsReachable() error {
 	client, _ := c.Factory.KubernetesClientSet()
 	_, err := client.ServerVersion()
 	if err != nil {
-		return errors.New("Kubernetes cluster unreachable")
+		return errors.New("Kubernetes cluster unreachable: " + err.Error())
```
bacongobbler: can we change this to

```
return fmt.Errorf("Kubernetes cluster unreachable: %s", err.Error())
```

bacongobbler (changes requested):

pkg/kube/client.go
ylvmw: Sure. I will change it.

ylvmw: I have changed it and amend the original commit.

ylvmw:
Seems this PR is outdated, just open another PR #7591 to track this.

PR #7094, opened 2019-11-27
fixes #7088 template command bug

vazul:
The template command (and install/upgrade) will dump the content
of the problematic yaml content if it cannot be unmarshalled.

Signed-off-by: Laszlo Varadi <laszlo.varadi@gmail.com>

PR #7556, opened 2020-02-07
Fix 'helm template' to also print invalid yaml

unguiculus:
Signed-off-by: Reinhard Naegele <unguiculus@gmail.com>

### DIFF
--- a/cmd/helm/template.go
+++ b/cmd/helm/template.go
@@ -65,8 +65,15 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 			client.IncludeCRDs = includeCrds
 			rel, err := runInstall(args, client, valueOpts, out)
 
-			// We ignore a potential error here because we always want to print the YAML,
-			// even if it is not valid. The error is still returned afterwards.
+			if err != nil && !settings.Debug {
+				if rel != nil {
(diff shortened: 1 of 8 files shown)


unguiculus:
As mentioned in https://github.com/helm/helm/pull/7094#issuecomment-583562707, this is an alternative solution for #7094.

mattfarina:
Over on #7094 it was suggested by @bacongobbler that this should be behind the `--debug` flag. Any thoughts on that?

bacongobbler:
Yeah, same feedback as @mattfarina provided. The code looks good, but I'm interested in the CLI design. Would users want the full template output displayed when a rendering error occurs, or should it only be displayed when `--debug` is provided?

My concern is that if we display the full template output, the error message will be lost in the noise. For example, CI/CD pipelines probably don't want their templates rendered out in the log. By placing this under the `--debug` flag, users can opt into that functionality to gain a better understanding why their templates failed.

unguiculus:
The way I implemented it basically follows what Helm 2 does. Was YAML parsing introduced on purpose here or is this just an unintentional side effect of the refactoring? Anyways I'm open to placing it under the `--debug` flag. Might be the better option but I don't really have a strong opinion on that.

rimusz:
+1 for the `--debug` flag

unguiculus:
So I went ahead and placed it under the `--debug` flag.

bacongobbler (approved):

PR #7471, opened 2020-01-27
fix(engine): allow limited recursion in templates

bacongobbler:
This is a backport for PR #7443.

Signed-off-by: Matthew Fisher <matt.fisher@microsoft.com>

PR #7443, opened 2020-01-21
allow limited recursion in templates

zwwhdls:
fix https://github.com/helm/helm/issues/7439

My solution is count the number of recursion in template, when it's over 1000 then raise an error.

PR #7558, opened 2020-02-07
fix recursion count in templates

### LINKED ISSUE #7559: template recursion bug
Recursion in template functions is not properly detected. The count is never decremented after the template executes which results in a false positive. Link to the PR which fixes this: https://github.com/helm/helm/pull/7558

Output of `helm version`:
```
version.BuildInfo{Version:"v3.0.3", GitCommit:"ac925eb7279f4a6955df663a0128044a8a6b7593", GitTreeState:"clean", GoVersion:"go1.13.7"}
```
Output of `kubectl version`:

With a function defined as follows:
```
{{- define "test.nonRecursive" -}}
{{- printf "this is fine" -}}
{{- end -}}
```
And a template file with
```
{{- $ok := include "test.nonRecursive" . -}}
```
repeated several thousand times, the recursion detection kicks in and returns the error
```
Error: template: ibm-dlaas/templates/test.yaml:1002:11: executing "ibm-dlaas/templates/test.yaml" at <include "test.nonRecursive" .>: error calling include: rendering template has a nested reference name: test.nonRecursive: unable to execute template
```
even though there is no recursion involved.

ghost:
The count in `includedNames` is incremented every time a name is seen, but is not decremented after executing the template. In a large helm chart, a function could be included many times, which exceeds the `1000` limit, without necessarily including any recursion.

Recursive functions such as
```
{{- define "test.recursionTest" -}}
{{- printf "%s" (include "test.recursionTest" .) -}}
{{- end -}}
```
are still detected and result in an error.

Closes https://github.com/helm/helm/issues/7559

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -120,6 +120,7 @@ func (e Engine) initFunMap(t *template.Template, referenceTpls map[string]render
 			includedNames[name] = 1
 		}
 		err := t.ExecuteTemplate(&buf, name, data)
+		includedNames[name]--
 		return buf.String(), err
 	}
 
--- a/pkg/engine/engine_test.go
+++ b/pkg/engine/engine_test.go
@@ -643,3 +643,58 @@ func TestAlterFuncMap_tplinclude(t *testing.T) {
 	}
 
 }
+
+func TestRenderRecursionLimit(t *testing.T) {
+	// endless recursion should produce an error
+	c := &chart.Chart{
+		Metadata: &chart.Metadata{Name: "bad"},
+		Templates: []*chart.File{
+			{Name: "templates/base", Data: []byte(`{{include "recursion" . }}`)},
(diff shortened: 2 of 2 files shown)


bacongobbler:
Duplicate of #7471?

danielchengibm:
No, this is an issue with https://github.com/helm/helm/pull/7471. Just opened an issue here https://github.com/helm/helm/issues/7559.

bacongobbler:
Sorry. After looking more closely at the issue I realized this was after #7443 was merged. I thought I’d wasn’t in a release yet. My bad!

Thanks for the PR.

bacongobbler (commented):
Would you mind writing a unit test to cover this case? That way we can prevent this from occurring in the future. Thanks.

danielchengibm:
Sure, I can add one, probably early next week.

mattfarina:
Thanks for catching this and writing a great test.

mattfarina (approved):
This looks great. Really like the testing.

PR #7560, opened 2020-02-08
feat(tests): Allow to provision memory driver

marckhouzam:
The memory driver is used for go tests. It can also be used from the
command-line by setting the environment variable HELM_DRIVER=memory.
In the latter case however, there was no way to pre-provision some
releases.

This commit introduces the HELM_MEMORY_DRIVER_DATA variable which
can be used to provide a colon-separated list of yaml files specifying
releases to provision automatically.

An example to try this out:
   HELM_DRIVER=memory \
   HELM_MEMORY_DRIVER_DATA=./testdata/releases.yaml \
   helm list --all-namespaces

Note:

- This is meant to allow writing acceptance tests.  To illustrate new tests have been implemented in the following acceptance-testing PR: https://github.com/helm/acceptance-testing/pull/75

- This PR only allows to *pre-load* releases in the memory driver, it does not persist any changes made by the helm command itself.  It would be possible to simply dump the resulting memory driver content into the same yaml file to create an off-cluster storage mechanism.  However, I purposefully did not do this because it may give users the impression it is a valid way to store releases and since each helm command would parse the entire yaml file of releases and would also dump it, a growing set of releases would cause more and more slowness.  I think the right solution for an off-cluster storage is to use a database (like helm v2 had), not a file (this is beyond the scope of this PR).

### DIFF
--- a/cmd/helm/helm.go
+++ b/cmd/helm/helm.go
@@ -19,6 +19,7 @@ package main // import "helm.sh/helm/v3/cmd/helm"
 import (
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"log"
 	"os"
 	"strings"
@@ -26,6 +27,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"k8s.io/klog"
+	"sigs.k8s.io/yaml"
 
 	// Import to initialize client auth plugins.
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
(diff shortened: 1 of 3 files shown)


bacongobbler (approved):

hickeyma (changes requested):
LGTM @marckhouzam thanks. Do you mind rebasing and fixing the conflict?

marckhouzam:
Thanks @hickeyma. I rebased and forced pushed.  In a situation like this, does the previous approval stand (assuming the ci build passes) or do the maintainers have to re-approve?

hickeyma (approved):
Good to merge @marckhouzam, thanks.

PR #7561, opened 2020-02-08
fix(comp): Fix broken completion for --output flag

marckhouzam:
This was detected because the completion tests in the acceptance-testing repo started to fail. 

For commands using the new post-render flag, the completion for the --output flag was broken.

For example:
```
$   helm install -o <TAB>
__helm_handle_reply:47: command not found: __helm_output_options
```

The bash __helm_output_options function is no longer used but was referred to by mistake.
This commit removes the offending code.


mattfarina (approved):
lgtm. tested and it fixes the problem. I noticed this issue on Friday but failed to file an issue. Glad it was caught.

marckhouzam:
> I noticed this issue on Friday

I love that you saw it.  It means you use completion 😉 

Thanks for the review.

PR #7277, opened 2019-12-20
feat(checksum): Generate shasum/sha256sum -c compatible sha256 file

t2sa:
Commands shasum -a 256 -c (or) sha256sum -c can read the SHA sum and validate the TAR/ZIP archive
Example:
  Download helm-v3.0.2-darwin-amd64.tar.gz.sha256 and helm-v3.0.2-darwin-amd64.tar.gz and running below will result in
    shasum -a 256 -c helm-v3.0.2-darwin-amd64.tar.gz.sha256
    helm-v3.0.2-darwin-amd64.tar.gz: OK

Closes #4968

PR #7568, opened 2020-02-10
Fix shasums to be usable by shasum and sha256sum applications

### LINKED ISSUE #7567: New shasums.txt file is not accessible to download and does not handle multiple version
A new shasums file was created (`shasums.txt`) that uses the format `shasum` and `sha256sum` can validate. But, there are two problems with it:

1. It is not available for download to verify against.
2. Does not handle multiple releases. The file is named `shasums.txt` for 3.1.0. When 3.1.1 comes out the file name for that one is also `shasums.txt`.

There there is the usability issue. If someone does have the `shasums.txt` file and tries to use it after downloading a single file they get something like...

```
$ shasum -a 256 -c shasums.txt
helm-3.1.0-rc.1-darwin-amd64.tar.gz: OK
shasum: helm-3.1.0-rc.1-linux-386.tar.gz:
helm-3.1.0-rc.1-linux-386.tar.gz: FAILED open or read
shasum: helm-3.1.0-rc.1-linux-amd64.tar.gz: No such file or directory
helm-3.1.0-rc.1-linux-amd64.tar.gz: FAILED open or read
shasum: helm-3.1.0-rc.1-linux-arm.tar.gz: No such file or directory
helm-3.1.0-rc.1-linux-arm.tar.gz: FAILED open or read
shasum: helm-3.1.0-rc.1-linux-arm64.tar.gz: No such file or directory
helm-3.1.0-rc.1-linux-arm64.tar.gz: FAILED open or read
shasum: helm-3.1.0-rc.1-linux-ppc64le.tar.gz: No such file or directory
helm-3.1.0-rc.1-linux-ppc64le.tar.gz: FAILED open or read
shasum: helm-3.1.0-rc.1-linux-s390x.tar.gz: No such file or directory
helm-3.1.0-rc.1-linux-s390x.tar.gz: FAILED open or read
shasum: helm-3.1.0-rc.1-windows-amd64.tar.gz: No such file or directory
helm-3.1.0-rc.1-windows-amd64.tar.gz: FAILED open or read
shasum: helm-3.1.0-rc.1-darwin-amd64.zip: No such file or directory
helm-3.1.0-rc.1-darwin-amd64.zip: FAILED open or read
shasum: helm-3.1.0-rc.1-linux-386.zip: No such file or directory
helm-3.1.0-rc.1-linux-386.zip: FAILED open or read
shasum: helm-3.1.0-rc.1-linux-amd64.zip: No such file or directory
helm-3.1.0-rc.1-linux-amd64.zip: FAILED open or read
shasum: helm-3.1.0-rc.1-linux-arm.zip: No such file or directory
helm-3.1.0-rc.1-linux-arm.zip: FAILED open or read
shasum: helm-3.1.0-rc.1-linux-arm64.zip: No such file or directory
helm-3.1.0-rc.1-linux-arm64.zip: FAILED open or read
shasum: helm-3.1.0-rc.1-linux-ppc64le.zip: No such file or directory
helm-3.1.0-rc.1-linux-ppc64le.zip: FAILED open or read
shasum: helm-3.1.0-rc.1-linux-s390x.zip: No such file or directory
helm-3.1.0-rc.1-linux-s390x.zip: FAILED open or read
shasum: helm-3.1.0-rc.1-windows-amd64.zip: No such file or directory
helm-3.1.0-rc.1-windows-amd64.zip: FAILED open or read
shasum: WARNING: 15 listed files could not be read

$ echo $?
1
```

Because of all the missing files the exit code IS NOT 0.

We should fix this prior to releasing 3.1.0.

I would suggest a 2 step process:

1. We break the `shasums.txt` file into one file for each download. Name them after the file + `.sha256sum` extension. When downloading a file they only need to download the corresponding checksum file and can use it with `shasum` out of the box. Being named after the file it will contain the version. This solves for 2 of the 3 problems.
2. Upload the file to get.helm.sh. The fetch-dist and signing work should also handle this new file. This will solve the other problem.

How does this plan sound?

mattfarina:
When #7277 was merged is was intended to create shasums accessible
in a way shasum -c or sha256sum could use to verify the files the
Helm project ships. The solution created a new file named
shasums.txt. This setup contained a few problems:

1. The new file file was not uploaded to get.helm.sh for someone
   to download and use.
2. The file had not version in the naming or path. This means that
   each new release of Helm will overwrite it. Downloading and
   validating an old file is impossible.
3. If one downloads a single file, the shasums.txt file, and uses
   shasum -c it will return an exit code that is 1. This is because
   of missing files as it is looking for all the files from the
   release.
4. The shasums.txt file is not signed for verification like the
   other files.

This change fixes these problems with the following changes:

* Instead of a shasums.txt file there is a .sha256sum file for
  each package. For example, helm-3.1.0-linux-amd64.zip.sha256sum.
  This file will can be used with `shasum -a 256 -c` to verify
  the single file helm-3.1.0-linux-amd64.zip. The exit code of
  checking a single file is 0 if the file passes.
* This new .sha256sum file is signed just like the .tar.gz, .zip,
  and .sha256 files. The provenance can be verified.
* The file name starts with `helm-` meaning the existing upload
  script in the deploy.sh file will move it to get.helm.sh.

Note, the existing .sha256 file can be deprecated and removed
in Helm v4 with the new .sha256sum file taking over. But,
for backwards compatibility with scripts it needs to be kept
during v3.

Closes #7567

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -157,15 +157,22 @@ fetch-dist:
 
 .PHONY: sign
 sign:
-	for f in _dist/*.{gz,zip,sha256} ; do \
+	for f in _dist/*.{gz,zip,sha256,sha256sum} ; do \
 		gpg --armor --detach-sign $${f} ; \
 	done
 
+# The contents of the .sha256sum file are compatible with tools like
+# shasum. For example, using the following command will verify
+# the file helm-3.1.0-rc.1-darwin-amd64.tar.gz:
+#   shasum -a 256 -c helm-3.1.0-rc.1-darwin-amd64.tar.gz.sha256sum
+# The .sha256 files hold only the hash and are not compatible with
+# verification tools like shasum or sha256sum. This method and file can be
+# removed in Helm v4.
 .PHONY: checksum
 checksum:
-	@if [ -f "_dist/shasums.txt" ]; then >_dist/shasums.txt; fi
 	for f in _dist/*.{gz,zip} ; do \
-		shasum -a 256 "$${f}"  | sed 's/_dist\///' | tee -a "_dist/shasums.txt" | awk '{print $$1}' > "$${f}.sha256" ; \
+		shasum -a 256 "$${f}" | sed 's/_dist\///' > "$${f}.sha256sum" ; \
+		shasum -a 256 "$${f}" | awk '{print $$1}' > "$${f}.sha256" ; \
 	done
 
 # ------------------------------------------------------------------------------


bacongobbler (approved):

bacongobbler:
It was argued that `--ignore-missing` could be used with shasums.txt, but the upload/override issue with each subsequent Helm release is real. LGTM

PR #7570, opened 2020-02-11
feat(upgrade): introduce --disable-openapi-validation:

Maronee:
This is a copy of the --disable-openapi-validation flag from the install command as introduced by Matthew Fisher.
See commit 67e57a5fbb7b210e534157b8f67c15ffc3445453

It allows upgrading releases without the need to validate the Kubernetes OpenAPI Schema.

### DIFF
--- a/cmd/helm/upgrade.go
+++ b/cmd/helm/upgrade.go
@@ -109,6 +109,7 @@ func newUpgradeCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 					instClient.Namespace = client.Namespace
 					instClient.Atomic = client.Atomic
 					instClient.PostRenderer = client.PostRenderer
+					instClient.DisableOpenAPIValidation = client.DisableOpenAPIValidation
 
 					rel, err := runInstall(args, instClient, valueOpts, out)
 					if err != nil {
@@ -165,6 +166,7 @@ func newUpgradeCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 	f.MarkDeprecated("recreate-pods", "functionality will no longer be updated. Consult the documentation for other methods to recreate pods")
 	f.BoolVar(&client.Force, "force", false, "force resource updates through a replacement strategy")
 	f.BoolVar(&client.DisableHooks, "no-hooks", false, "disable pre/post upgrade hooks")
+	f.BoolVar(&client.DisableOpenAPIValidation, "disable-openapi-validation", false, "if set, the upgrade process will not validate rendered templates against the Kubernetes OpenAPI Schema")
 	f.DurationVar(&client.Timeout, "timeout", 300*time.Second, "time to wait for any individual Kubernetes operation (like Jobs for hooks)")
 	f.BoolVar(&client.ResetValues, "reset-values", false, "when upgrading, reset the values to the ones built into the chart")
 	f.BoolVar(&client.ReuseValues, "reuse-values", false, "when upgrading, reuse the last release's values and merge in any overrides from the command line via --set and -f. If '--reset-values' is specified, this is ignored")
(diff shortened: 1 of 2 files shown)


hickeyma (commented):
@Maronee Thanks for pushing the PR. Can you fix the go formatting/style issue in the build? (https://circleci.com/gh/helm/helm/11180?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)

Maronee:
@hickeyma Thanks, fixed the formating and adjusted the code to take the flags as well when an initial install is needed for the upgrade command

bacongobbler (approved):
CC'ing @mattfarina; we may need to get this into the 3.1.0 RC before it's released so users can disable validation during `helm upgrade`.

For the record, I am [incredibly uncomfortable disabling API validation on `helm install` and `helm upgrade`](https://github.com/helm/helm/pull/6819#issuecomment-547645422). It introduces cases where Kubernetes will accept manifests that it shouldn't normally accept, and can cause indeterminate behaviour during upgrades. If you are going to opt into this feature, please make sure you're doing it for the right reasons.

Okay, rant over. Code LGTM.

cmd/helm/upgrade.go
```diff
@@ -165,6 +166,7 @@ func newUpgradeCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 	f.MarkDeprecated("recreate-pods", "functionality will no longer be updated. Consult the documentation for other methods to recreate pods")
 	f.BoolVar(&client.Force, "force", false, "force resource updates through a replacement strategy")
 	f.BoolVar(&client.DisableHooks, "no-hooks", false, "disable pre/post upgrade hooks")
+	f.BoolVar(&client.DisableOpenAPIValidation, "disable-openapi-validation", false, "if set, the installation process will not validate rendered templates against the Kubernetes OpenAPI Schema")
```
hickeyma: This is a nit but it might read better for users as:
"the installation process" --> "the upgrade process"

hickeyma (changes requested):
It looks good to me apart from one nit that is probably better be updated from a usability point of view. I will hold on approval but good ahead and merge @bacongobbler if you think it is ok to go with.

hickeyma:
> For the record, I am incredibly uncomfortable disabling API validation on helm install and helm upgrade. It introduces cases where Kubernetes will accept manifests that it shouldn't normally accept, and can cause indeterminate behaviour during upgrades. If you are going to opt into this feature, please make sure you're doing it for the right reasons.

I agree with your point fully. We are seeing a lots of issues with Helm releases going into an indeterminate state with the cluster when people perform operations that circumvent the recommended approach. As long as users are aware that doing things like this may mean having to remove the release to recover then they at least understand the risk.

cmd/helm/upgrade.go
Maronee: adjusted the message

Maronee:
@hickeyma Hi, I adjusted the message with the latest commit but remvoed the message by accident when "resolving" it

hickeyma (approved):
LGTM, thanks @Maronee

placaze:
Any way this fix could be included in v3.1.2 instead of v3.2.0, please?

bacongobbler:
We follow the convention of [semantic versioning](https://semver.org/spec/v2.0.0.html) as best as we can. We follow this convention religiously to clearly state what should and should not go into a patch release.

Semantic versioning tells us that new features should be introduced in a minor release (3.x.0), and release patches (3.1.x) should only contain fixes for bugs or regressions introduced since the release of 3.x.0.

From https://semver.org/spec/v2.0.0.html:

> Bug fixes not affecting the API increment the patch version, backwards compatible API additions/changes increment the minor version, and backwards incompatible API changes increment the major version.

Because of that, this will have to go into 3.2.

PR #7575, opened 2020-02-12
Workaround obsolete API versions

pho-enix:
This is a workaround for changes in API versions like
extensions/v1beta1 => apps/v1

From Helm's perspective those are new Resources as the API version
cannot be changed. However there is a compat layer in k8s that will
return the old resources for the new API version. Helm will reject the
resource 'creation' as there is already a resource of the same kind and
version.

Do we want to maintain compat code in this place? Probably not.

### DIFF
--- a/pkg/action/upgrade.go
+++ b/pkg/action/upgrade.go
@@ -452,12 +452,11 @@ func isCRD(groupVersion string) bool {
 		return false
 	}
 	group := s[0]
-	s = strings.Split(group, ".")
-	if len(s) == 1 {
+	if strings.Count(group, ".") == 0 {
 		// "apps/v1"
 		return false
 	}
-	if s[len(s)-2] == "k8s" && s[len(s)-1] == "io" {
+	if strings.HasSuffix(group, "k8s.io") {
 		// "networking.k8s.io/v1"
 		return false
 	}
--- a/pkg/action/upgrade.go
+++ b/pkg/action/upgrade.go
@@ -202,19 +202,7 @@ func (u *Upgrade) performUpgrade(originalRelease, upgradedRelease *release.Relea
 		return upgradedRelease, errors.Wrap(err, "unable to build kubernetes objects from new release manifest")
 	}
 
-	// Do a basic diff using gvk + name to figure out what new resources are being created so we can validate they don't already exist
-	existingResources := make(map[string]bool)
-	for _, r := range current {
-		existingResources[objectKey(r)] = true
-	}
-
-	var toBeCreated kube.ResourceList
-	for _, r := range target {
-		if !existingResources[objectKey(r)] {
-			toBeCreated = append(toBeCreated, r)
-		}
-	}
-
+	toBeCreated := detectNewResources(&current, &target)
 	if err := existingResourceConflict(toBeCreated); err != nil {
 		return nil, errors.Wrap(err, "rendered manifests contain a new resource that already exists. Unable to continue with update")
 	}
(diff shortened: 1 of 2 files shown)


mattfarina:
@pho-enix thanks for the PR. Helm uses a[ developers certificate of origin (DCO)](developercertificate.org/) to show you are allowed to contribute said material. The checks tab at the top will show you how to correct the commit. In new commits you can use the `--signoff` flag.

I'm unsure if Helm should transform api versions. For one, they don't always translate from one to another as some subtle changes happened along the way. The error response coming back may not make sense.

I've added this as a discussion topic to the next Helm dev call so we can talk about it. You are welcome to attend if you like. If not, I'm prepared to talk about it.

pkg/action/upgrade.go
```diff
@@ -426,5 +426,26 @@ func recreate(cfg *Configuration, resources kube.ResourceList) error {
 
 func objectKey(r *resource.Info) string {
 	gvk := r.Object.GetObjectKind().GroupVersionKind()
-	return fmt.Sprintf("%s/%s/%s/%s", gvk.GroupVersion().String(), gvk.Kind, r.Namespace, r.Name)
+
+	kind := gvk.Kind
+	groupVersion := gvk.GroupVersion().String()
+
+	if groupVersion == "apps/v1beta1" || groupVersion == "apps/v1beta2" {
+		groupVersion = "apps/v1"
```
DavidZisky: Why don't you just check `if old_api_version != new_api_version` instead?

pho-enix: Good idea. Thanks for the hint. However we have to be a bit cautious about the Group. There might be the same Kind in different Groups.

For example:
NetworkPolicy - networking.k8s.io/v1
NetworkPolicy - crd.projectcalico.org/v1

In principle anyone could create a `Deplyoment` Kind in his CRD.

DavidZisky: Fair enough, but then you could check instead if only bit after "/" has been changed (for the same api type). That would be a more future proof solution - if there will be "apps/v2" version released in the future you'll have to rewrite this

pho-enix: No, this will not be sufficient. As for example `Deployment` moved
`extensions/v1beta1` -> `apps/v1`

Please check the updated code :)

DavidZisky: Yeah, I see. Updated code looks cool tho +1

pho-enix:
Issue 1:
Clashes of Kubernetes Resources and CRD with same Kind might produce strange errors (if a Resource with the same name already exists)
Issue 2:
Clashes of CRD and CRD with same Kind might produce strange errors (if a Resource with the same name already exists)

I general we'd have to decide how many corner cases we want to take into account:

Option 1) Simply ignore Version and Group during upgrade.
+ Code change will be super small
- [ ] Issue 1 fixed
- [ ] Issue 2 fixed

Option 2) Current PR (ed739d29beacc700bddef50ca9a6eefd7c68de62)
+ Quite a bit of code
- [x] Issue 1 fixed
- [ ] Issue 2 fixed

pkg/action/upgrade.go
```diff
@@ -425,6 +413,54 @@ func recreate(cfg *Configuration, resources kube.ResourceList) error {
...
+		// "v1"
+		return false
+	}
+	group := s[0]
+	s = strings.Split(group, ".")
+	if len(s) == 1 {
+		// "apps/v1"
+		return false
+	}
+	if s[len(s)-2] == "k8s" && s[len(s)-1] == "io" {
```
dschniepp: Excellent code 👍 For readability I would support to use `strings.HasSuffix(s, "k8s.io")`.

jlegrone:
I think there might be a simpler solution to this. Opening a PR for discussion in a little bit and I'll link it here. Thanks for your work on this so far @pho-enix!

pho-enix:
@jlegrone Agreed that my solution is a bit complex. As this issue is really a blocker for a lot of users we should aim for a fast solution (#7627).

adnan-kamili:
Maybe this behavior can be enabled through a command-line option so that it doesn't break anything and users who know what they are doing use this command-line option.

avielb:
Waiting for this as well
could be very useful

bacongobbler:
closed in favour of https://github.com/helm/helm/pull/7649.

PR #7601, opened 2020-02-15
feat(pkg): Install / Template Charts without downloading to filesystem

### LINKED ISSUE #7545: Use in-memory cache instead of filesystem HELM_REPOSITORY_CACHE
Hi,
As part of `helm install ...` ( and other flows ), the chart tgz gets written to HELM_REPOSITORY_CACHE. 
Any thoughts on doing that in-memory ?

I'm using the Helm Golang APIs from a service running in a container and I would want to avoid writing to the filesystem altogether.

waveywaves:
fixes https://github.com/helm/helm/issues/7545

- This will allow users to use the charts without them being downloaded to the HELM_REPOSITORY_CACHE by setting the value for the same to "off"

### DIFF
--- a/cmd/helm/install.go
+++ b/cmd/helm/install.go
@@ -19,6 +19,7 @@ package main
 import (
 	"fmt"
 	"io"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
@@ -163,13 +164,13 @@ func runInstall(args []string, client *action.Install, valueOpts *values.Options
 		client.Version = ">0.0.0-0"
 	}
 
-	name, chart, err := client.NameAndChart(args)
+	name, schart, err := client.NameAndChart(args)
 	if err != nil {
 		return nil, err
 	}
 	client.ReleaseName = name
 
-	cp, err := client.ChartPathOptions.LocateChart(chart, settings)
+	cp, err := client.ChartPathOptions.LocateChart(schart, settings)
 	if err != nil {
 		return nil, err
 	}
@@ -183,7 +184,12 @@ func runInstall(args []string, client *action.Install, valueOpts *values.Options
 	}
 
 	// Check chart dependencies to make sure all are present in /charts
-	chartRequested, err := loader.Load(cp)
+	var chartRequested *chart.Chart
+	if settings.RepositoryCache == "off" {
+		chartRequested, err = downloader.GetChart(schart, strings.TrimSpace(client.ChartPathOptions.Version))
+	} else {
+		chartRequested, err = loader.Load(cp)
+	}
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -620,7 +620,6 @@ OUTER:
 func (c *ChartPathOptions) LocateChart(name string, settings *cli.EnvSettings) (string, error) {
 	name = strings.TrimSpace(name)
 	version := strings.TrimSpace(c.Version)
-
 	if _, err := os.Stat(name); err == nil {
 		abs, err := filepath.Abs(name)
 		if err != nil {
@@ -660,20 +659,24 @@ func (c *ChartPathOptions) LocateChart(name string, settings *cli.EnvSettings) (
 		name = chartURL
 	}
 
-	if err := os.MkdirAll(settings.RepositoryCache, 0755); err != nil {
+	// If HELM_REPOSITORY_CACHE has an "off" value, the chart won't be downloaded
+	if err := os.MkdirAll(settings.RepositoryCache, 0755); err != nil && settings.RepositoryCache != "off" {
 		return "", err
 	}
 
-	filename, _, err := dl.DownloadTo(name, version, settings.RepositoryCache)
+	chartLocation, _, err := dl.DownloadTo(name, version, settings.RepositoryCache)
 	if err == nil {
-		lname, err := filepath.Abs(filename)
+		if settings.RepositoryCache != "off" {
+			return chartLocation, err
+		}
+		lname, err := filepath.Abs(chartLocation)
 		if err != nil {
-			return filename, err
+			return chartLocation, err
 		}
 		return lname, nil
 	} else if settings.Debug {
-		return filename, err
+		return chartLocation, err
 	}
 
-	return filename, errors.Errorf("failed to download %q (hint: running `helm repo update` may help)", name)
+	return chartLocation, errors.Errorf("failed to download %q (hint: running `helm repo update` may help)", name)
 }
--- a/pkg/downloader/chart_downloader.go
+++ b/pkg/downloader/chart_downloader.go
@@ -16,12 +16,18 @@ limitations under the License.
 package downloader
 
 import (
+	"bytes"
 	"fmt"
 	"io"
 	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
+
+	lru "github.com/hashicorp/golang-lru"
+
+	"helm.sh/helm/v3/pkg/chart"
+	"helm.sh/helm/v3/pkg/chart/loader"
 
 	"github.com/pkg/errors"
 
@@ -72,6 +78,16 @@ type ChartDownloader struct {
 	RepositoryCache  string
 }
 
+type inMemoryChart struct {
+	ChartData              *bytes.Buffer
+	ProvenanceVerification *provenance.Verification
+}
+
+// inMemoryCharts maps version to a particular chart
+type inMemoryCharts map[string]inMemoryChart
+
+var inMemoryCache, _ = lru.New(128)
+
 // DownloadTo retrieves a chart. Depending on the settings, it may also download a provenance file.
 //
 // If Verify is set to VerifyNever, the verification will be nil.
@@ -94,43 +110,74 @@ func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *proven
 		return "", nil, err
 	}
 
-	data, err := g.Get(u.String(), c.Options...)
+	chartData, err := g.Get(u.String(), c.Options...)
 	if err != nil {
 		return "", nil, err
 	}
 
-	name := filepath.Base(u.Path)
-	destfile := filepath.Join(dest, name)
-	if err := fileutil.AtomicWriteFile(destfile, data, 0644); err != nil {
-		return destfile, nil, err
-	}
-
 	// If provenance is requested, verify it.
 	ver := &provenance.Verification{}
-	if c.Verify > VerifyNever {
-		body, err := g.Get(u.String() + ".prov")
-		if err != nil {
-			if c.Verify == VerifyAlways {
-				return destfile, ver, errors.Errorf("failed to fetch provenance %q", u.String()+".prov")
-			}
-			fmt.Fprintf(c.Out, "WARNING: Verification not found for %s: %s\n", ref, err)
-			return destfile, ver, nil
-		}
-		provfile := destfile + ".prov"
-		if err := fileutil.AtomicWriteFile(provfile, body, 0644); err != nil {
+	chartRef := ""
+	if dest == "off" {
+		chartRef = ref
+	} else {
+		filename := filepath.Base(u.Path)
+		destfile := filepath.Join(dest, filename)
+		if err := fileutil.AtomicWriteFile(destfile, chartData, 0644); err != nil {
 			return destfile, nil, err
 		}
+		chartRef = destfile
+	}
+	if c.Verify > VerifyNever {
+		provData, err := g.Get(u.String() + ".prov")
+		if err != nil {
+			if c.Verify == VerifyAlways {
+				return chartRef, ver, errors.Errorf("failed to fetch provenance %q", u.String()+".prov")
+			}
+			fmt.Fprintf(c.Out, "WARNING: Verification not found for %s: %s\n", ref, err)
+			return chartRef, ver, nil
+		}
 
-		if c.Verify != VerifyLater {
-			ver, err = VerifyChart(destfile, c.Keyring)
-			if err != nil {
-				// Fail always in this case, since it means the verification step
-				// failed.
-				return destfile, ver, err
+		if dest == "off" {
+			if c.Verify != VerifyLater {
+				ver, err = verifyInMemoryChart(chartRef, c.Keyring, chartData, provData)
+				if err != nil {
+					// Fail always in this case, since it means the verification step
+					// failed.
+					return chartRef, ver, err
+				}
+			}
+		} else {
+			provfile := chartRef + ".prov"
+			if err := fileutil.AtomicWriteFile(provfile, provData, 0644); err != nil {
+				return chartRef, nil, err
+			}
+
+			if c.Verify != VerifyLater {
+				ver, err = VerifyChart(chartRef, c.Keyring)
+				if err != nil {
+					// Fail always in this case, since it means the verification step
+					// failed.
+					return chartRef, ver, err
+				}
 			}
 		}
 	}
-	return destfile, ver, nil
+
+	if dest == "off" {
+		inMemoryCache.ContainsOrAdd(chartRef, make(inMemoryCharts))
+		tempCharts, ok := inMemoryCache.Get(chartRef)
+
+		if !ok {
+			return chartRef, ver, errors.New(fmt.Sprintf("error during retrival of chart list %s from Cache", ref))
+		}
+		tempCharts.(inMemoryCharts)[version] = inMemoryChart{
+			ChartData:              chartData,
+			ProvenanceVerification: ver,
+		}
+		inMemoryCache.Add(chartRef, tempCharts)
+	}
+	return chartRef, ver, nil
 }
 
 // ResolveChartVersion resolves a chart reference to a URL.
@@ -153,6 +200,10 @@ func (c *ChartDownloader) ResolveChartVersion(ref, version string) (*url.URL, er
 		return nil, errors.Errorf("invalid chart URL format: %s", ref)
 	}
 	c.Options = append(c.Options, getter.WithURL(ref))
+
+	if c.RepositoryCache == "off" {
+		return u, nil
+	}
 
 	rf, err := loadRepoConfig(c.RepositoryConfig)
 	if err != nil {
@@ -293,6 +344,19 @@ func VerifyChart(path, keyring string) (*provenance.Verification, error) {
 	return sig.Verify(path, provfile)
 }
 
+// verifyInMemoryChart takes key to a chart in the the inMemoryCache and a keyring,
+// and verifies the chart.
+//
+// It assumes that a chart data is accompanied by a embedded provenance file whose
+// name is the archive file name plus the ".prov" extension.
+func verifyInMemoryChart(ref, keyring string, chartData, provData *bytes.Buffer) (*provenance.Verification, error) {
+	sig, err := provenance.NewFromKeyring(keyring, "")
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to load keyring")
+	}
+	return sig.VerifyFromData(ref, chartData, provData)
+}
+
 // isTar tests whether the given file is a tar file.
 //
 // Currently, this simply checks extension, since a subsequent function will
@@ -366,4 +430,13 @@ func loadRepoConfig(file string) (*repo.File, error) {
 		return nil, err
 	}
 	return r, nil
+}
+
+func GetChart(chart string, version string) (*chart.Chart, error) {
+	chartList, _ := inMemoryCache.Get(chart)
+	if chartList != nil {
+		chartData := chartList.(inMemoryCharts)[version].ChartData
+		return loader.LoadArchive(chartData)
+	}
+	return nil, errors.New("Chart not found in cache")
 }
--- a/pkg/provenance/sign.go
+++ b/pkg/provenance/sign.go
@@ -229,7 +229,7 @@ func (s *Signatory) ClearSign(chartpath string) (string, error) {
 	return out.String(), err
 }
 
-// Verify checks a signature and verifies that it is legit for a chart.
+// Verify checks a signature and verifies that it is legit for a chart (stored in the filesystem)
 func (s *Signatory) Verify(chartpath, sigpath string) (*Verification, error) {
 	ver := &Verification{}
 	for _, fname := range []string{chartpath, sigpath} {
@@ -240,8 +240,31 @@ func (s *Signatory) Verify(chartpath, sigpath string) (*Verification, error) {
 		}
 	}
 
+	// Get sigData
+	sigData, err := ioutil.ReadFile(sigpath)
+	if err != nil {
+		return nil, err
+	}
+
+	// Get chartData
+	chartData, err := ioutil.ReadFile(chartpath)
+	if err != nil {
+		return nil, err
+	}
+
+	ver, nil := s.VerifyFromData(chartpath, bytes.NewBuffer(chartData), bytes.NewBuffer(sigData))
+
+	// TODO: when image signing is added, verify that here.
+
+	return ver, nil
+}
+
+// VerifyFromData checks the signature and verifies that it is legit for a chart (not stored in the filesystem)
+func (s *Signatory) VerifyFromData(chartRef string, chartData, sigData *bytes.Buffer) (*Verification, error) {
+	ver := &Verification{}
+
 	// First verify the signature
-	sig, err := s.decodeSignature(sigpath)
+	sig, err := s.decodeSignatureFromData(sigData.Bytes())
 	if err != nil {
 		return ver, errors.Wrap(err, "failed to decode signature")
 	}
@@ -253,28 +276,37 @@ func (s *Signatory) Verify(chartpath, sigpath string) (*Verification, error) {
 	ver.SignedBy = by
 
 	// Second, verify the hash of the tarball.
-	sum, err := DigestFile(chartpath)
+	sum, err := Digest(chartData)
 	if err != nil {
 		return ver, err
 	}
+
+	verification, err := s.verifySum(sum, chartRef, sig, ver)
+	if err != nil {
+		return verification, err
+	}
+
+	// TODO: when image signing is added, verify that here.
+
+	return ver, nil
+}
+
+func (s *Signatory) verifySum(sum string, chartRef string, sig *clearsign.Block, ver *Verification) (*Verification, error) {
 	_, sums, err := parseMessageBlock(sig.Plaintext)
 	if err != nil {
 		return ver, err
 	}
 
 	sum = "sha256:" + sum
-	basename := filepath.Base(chartpath)
+	basename := filepath.Base(chartRef)
 	if sha, ok := sums.Files[basename]; !ok {
-		return ver, errors.Errorf("provenance does not contain a SHA for a file named %q", basename)
+		return ver, errors.Errorf("provenance does not contain a SHA for reference named %q", basename)
 	} else if sha != sum {
 		return ver, errors.Errorf("sha256 sum does not match for %s: %q != %q", basename, sha, sum)
 	}
 	ver.FileHash = sum
 	ver.FileName = basename
-
-	// TODO: when image signing is added, verify that here.
-
-	return ver, nil
+	return nil, nil
 }
 
 func (s *Signatory) decodeSignature(filename string) (*clearsign.Block, error) {
@@ -283,6 +315,10 @@ func (s *Signatory) decodeSignature(filename string) (*clearsign.Block, error) {
 		return nil, err
 	}
 
+	return s.decodeSignatureFromData(data)
+}
+
+func (s *Signatory) decodeSignatureFromData(data []byte) (*clearsign.Block, error) {
 	block, _ := clearsign.Decode(data)
 	if block == nil {
 		// There was no sig in the file.
--- a/go.mod
+++ b/go.mod
@@ -22,6 +22,7 @@ require (
 	github.com/gosuri/uitable v0.0.4
 	github.com/jmoiron/sqlx v1.2.0
 	github.com/lib/pq v1.3.0
+	github.com/hashicorp/golang-lru v0.5.1
 	github.com/mattn/go-shellwords v1.0.10
 	github.com/mitchellh/copystructure v1.0.0
 	github.com/opencontainers/go-digest v1.0.0-rc1
--- a/pkg/downloader/chart_downloader_test.go
+++ b/pkg/downloader/chart_downloader_test.go
@@ -322,6 +322,84 @@ func TestDownloadTo_VerifyLater(t *testing.T) {
 	}
 }
 
+func TestDownloadToInMemoryCache(t *testing.T) {
+	// Set up a fake repo with basic auth enabled
+	srv, err := repotest.NewTempServer("testdata/*.tgz*")
+	srv.Stop()
+	if err != nil {
+		t.Fatal(err)
+	}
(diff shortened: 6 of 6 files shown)


waveywaves:
cc @sbose78 @mattfarina

sbose78:
Thank you!
Could you add a test for a `helm install` or a `helm template` flow?

marckhouzam:
Thanks @waveywaves for taking a stab at this.  I'm having trouble understanding what the PR does however.  

From what I understood of #7545, the ask is that when `helm install` (called by API) downloads a chart, it not write it to disk.  The modifications of this PR are within the `main` package which I don't believe gets called when a program uses the helm API.

Similarly, since the changes are under `cmd/helm`, I believe they get triggered when helm is run from the command-line.  In that case however, each command is independent, so running a `helm repo add` is a different call than `helm install`, so anything stored in memory will not be persisted between calls.

Finally, the new `InMemoryCache` has repos added to it and removed from it, but the cache is never read.  Is there something missing?

This is not my area of expertise, so forgive me if I completely missed something.

waveywaves:
@marckhouzam Thank you for reviewing this. And as you said, I have only taken an initial "stab" at this hahaha. I was having trouble understanding where exactly the cache would lie. In this case considering that we are dealing with "helm repository cache" the first thing that came to my mind was to enable caching mechanism there. Based on your comment

> Similarly, since the changes are under cmd/helm, I believe they get triggered when helm is run from the command-line. In that case however, each command is independent, so running a `helm repo add` is a different call than `helm install`, so anything stored in memory will not be persisted between calls.

I was confused about this as well. It makes more sense now. I will update the PR based on this and the comment below, just need to evaluate some bits. 

> Finally, the new InMemoryCache has repos added to it and removed from it, but the cache is never read. Is there something missing?

waveywaves:
@marckhouzam I have updated the PR. Now we are able to install charts without saving them to cache. Can you PTAL

cc @sbose78

jdolitsky:
@waveywaves can you take a look at the CI failure? seems related to your change:
```
Downloading testdata/testcharts/empty to InMemoryCache--- FAIL: TestInstall (0.04s)
    --- FAIL: TestInstall/install_with_--use-inmemory-cache (0.00s)
        helm_test.go:86: expected error, got 'repo testdata not found'
        helm_test.go:89: does not match golden file testdata/output/cached-chart.txt
            
            WANT:
            'NAME: cached
            LAST DEPLOYED: Fri Sep  2 22:04:05 1977
            NAMESPACE: default
            STATUS: deployed
            REVISION: 1
            TEST SUITE: None
            '
            
            GOT:
            'Error: repo testdata not found
            '
```

sbose78:
@akashshinde  could you please review this?

waveywaves:
@jdolitsky I have added the required output needed in repo testdata, can you help me out in figuring what I might be doing wrong here ?

pkg/downloader/chart_downloader.go
```diff
@@ -133,6 +137,26 @@ func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *proven
 	return destfile, ver, nil
 }
 
+func (c *ChartDownloader) DownloadToInMemoryCache(ref, version string) (string, *provenance.Verification, error) {
```
akashshinde: @waveywaves any specific reason we are returning `*provenance.Verification`  ? since anyways we return `nil` in all paths.

waveywaves: I have updated to code to return a non null, provenance verification is important to check the source of the chart against the given keyring. I would like to cover this in another PR if that is fine as it would involve some refactoring as well.

waveywaves:
@jdolitsky Can you PTAL ?

marckhouzam:
Thanks for the rework @waveywaves!

Based on the recommendation at https://github.com/helm/helm/issues/7545#issuecomment-589706721 could you limit this PR to the necessary changes to the API (pkg/) and avoid changes to the command-line (cmd/)?

This will help ensure the original request is fulfilled and avoids adding new flags until we get a little more experience with this feature.

marckhouzam:
To simplify testing, how about separating the changes of `pkg/` into one PR (this one) and have a second PR using the first one, where you'd include the `cmd/` changes?  That way we could easily see how the `pkg/` changes are meant to be used, but we wouldn't have to include the cli changes right away.

marckhouzam:
Thanks @waveywaves for splitting the PR, it makes the review easier.  The code looks quite good!

I've finished the review but before I post comments, let's discuss two points:

1. the true request from  #7545 is not necessarily for an in-memory cache (although that is what the title says) but, based on https://github.com/helm/helm/issues/7545#issuecomment-587270220, to not leave modifications to the file system after installing the chart.  Therefore, a possible solution is to still create the necessary files, but to cleanup after ourselves; this is not as elegant as the memory cache, but may allow you to implement the provenance check more easily.  I actually like the in-memory cache better, but I feel the provenance check must be implemented and may be causing too much code duplication.  I'll let you think about it and make a recommendation.
 
1. In this comment https://github.com/helm/helm/issues/7545#issuecomment-589706721 @mattfarina  suggested making the cache 'pluggable', which means it could be an in-memory cache, or some other caching technique (he mentions object-storage) that could be selected by the user (assuming we implement it eventually).  I don't think you need to implement the whole pluggable solution, but we do need to keep the API generic so that the full pluggable solution could still be possible.  So, at a minimum, we should avoid any new API that specifically refers to 'in-memory', but use more generic terms.

waveywaves:
@marckhouzam I have updated the PR with Provenance Verification done without having to download the chart. 
Regarding the naming convention for this solution, considering that "in-memory" can cause some issues, can you suggest some other names ?

marckhouzam:
Thanks for the update! The provenance was important.

As for the naming convention, I'm having trouble coming up with something future proof.  We are introducing a new API but I'm not clear enough on what this cache could evolve into to define the API well.

Maybe @mattfarina can explain a little more what he has imagine as a "pluggable cache"?

karuppiah7890:
@waveywaves Please check the breaking change here - https://github.com/helm/helm/pull/7601/files#diff-4140e2fe89434d8f5a9133898128bcefR233 for the `Verify` method name change. It's an exported method. We can't have breaking changes in the API, also mentioned [here](https://github.com/helm/helm/issues/7545#issuecomment-589706721). I get it that a new method has been added - I would recommend giving a good name to that (if you think the existing name is good, cool 👍 ) but we can't change the name for the existing method till we go to v4

waveywaves:
@karuppiah7890 I am slowly getting the hand of what is backward compatibility in help. Thanks for checking it out. I will update the PR with the exiting name. That seems alright.
Thanks for checking this out :D

pkg/downloader/chart_downloader.go
```diff
@@ -97,6 +98,13 @@ func atomicWriteFile(filename string, body io.Reader, mode os.FileMode) error {
 	return os.Rename(tempName, filename)
 }
 
+type InMemoryChart struct {
```
marckhouzam: Can you move this to before the first function declaration, as well as the associated variable `InMemoryCache`.

pkg/downloader/chart_downloader.go
```diff
@@ -97,6 +98,13 @@ func atomicWriteFile(filename string, body io.Reader, mode os.FileMode) error {
...
+	Chart                  *bytes.Buffer
```
marckhouzam: I don't believe we need to make this type public.  So it should start with a lowercase letter instead.

pkg/downloader/chart_downloader.go
```diff
@@ -97,6 +98,13 @@ func atomicWriteFile(filename string, body io.Reader, mode os.FileMode) error {
...
+	ProvenanceVerification *provenance.Verification
+}
+
+var InMemoryCache = make(map[string]InMemoryChart)
```
marckhouzam: Can we make this private and encapsulate the details? Users should not know the details of how the cache is implemented.

pkg/downloader/chart_downloader.go
```diff
@@ -97,6 +98,13 @@ func atomicWriteFile(filename string, body io.Reader, mode os.FileMode) error {
...
+
```
marckhouzam: Can you explain why we need a map? Are there use-cases where we need to store more than one chart at a time?

pkg/downloader/chart_downloader.go
```diff
@@ -158,6 +166,52 @@ func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *proven
...
+			ver, err = VerifyInMemoryChart(ref, c.Keyring, chartData, provData)
+			if err != nil {
+				// Fail always in this case, since it means the verification step
+				// failed.
+				return ref, ver, err
+			}
+		}
+	}
+
+	InMemoryCache[ref] = InMemoryChart{
```
marckhouzam: If we really need a map, then don't we need to index on both ref and version? Or else a request for the same chart but a different version will overwrite the previous request of the same chart. If this is not a problem, maybe we don't need a map at all?

marckhouzam (changes requested):
Thanks @waveywaves for this effort.  Here are some comments to move this along, although I haven't reviewed the provenance part yet.  But @karuppiah7890's comment about backwards-compatibility is very important.

pkg/downloader/chart_downloader.go
waveywaves: I have added the Version attribute for the chart. I think we don't really need it as of now, but for the sake of semantics we should keep it. Otherwise what is tracking the InMemory Charts ?

pkg/downloader/chart_downloader.go
waveywaves: While installing charts with dependencies multiple charts would be downladded and stored. Let me know if I am missing a fact here.

pkg/downloader/chart_downloader.go
waveywaves: Done.

pkg/downloader/chart_downloader.go
waveywaves: Done

pkg/downloader/chart_downloader.go
waveywaves: Done

pkg/downloader/chart_downloader.go
waveywaves: I think we don't really need it as of now, but for the sake of semantics we should keep it. Otherwise what is tracking the InMemory Charts ?

pkg/action/install.go
```diff
@@ -760,10 +763,15 @@ func (c *ChartPathOptions) LocateChart(name string, settings *cli.EnvSettings) (
 		name = chartURL
 	}
 
+	locationName := ""
 	if err := os.MkdirAll(settings.RepositoryCache, 0755); err != nil {
-		return "", err
+		return locationName, err
+	}
+	if c.UseInMemoryCache {
+		fmt.Print("Downloading ", name, " to InMemoryCache")
```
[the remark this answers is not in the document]
waveywaves: Changed this to a debug print statement.

pkg/action/install.go
```diff
@@ -760,10 +763,15 @@ func (c *ChartPathOptions) LocateChart(name string, settings *cli.EnvSettings) (
 		name = chartURL
 	}
 
+	locationName := ""
```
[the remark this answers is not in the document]
waveywaves: Removed it.

pkg/action/install.go
```diff
@@ -722,20 +723,22 @@ func (c *ChartPathOptions) LocateChart(name string, settings *cli.EnvSettings) (
...
 
-	if _, err := os.Stat(name); err == nil {
-		abs, err := filepath.Abs(name)
-		if err != nil {
-			return abs, err
-		}
-		if c.Verify {
-			if _, err := downloader.VerifyChart(abs, c.Keyring); err != nil {
-				return "", err
+	if !c.UseInMemoryCache {
```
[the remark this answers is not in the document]
waveywaves: Currentlty this PR only supports downloading to in memory cache from a url. I will update this to support downloading from filesystem to inmemory cache.

pkg/action/install.go
```diff
@@ -760,10 +763,15 @@ func (c *ChartPathOptions) LocateChart(name string, settings *cli.EnvSettings) (
 		name = chartURL
 	}
 
+	locationName := ""
 	if err := os.MkdirAll(settings.RepositoryCache, 0755); err != nil {
```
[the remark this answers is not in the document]
waveywaves: I missed that. Have moved the block above.

waveywaves (commented):
I have updated the PR with the necessary changes. 
Testing it now.

pkg/downloader/chart_downloader.go
waveywaves: This is important as this would be used by `cmd` to query the charts.

pkg/downloader/chart_downloader.go
waveywaves: the InMemoryChart would need to be used by the API as this is thestore where the charts are kept and not in the filesystem.

pkg/downloader/chart_downloader.go
marckhouzam: I guess my question was: do we need to put more than one chart at a time in the InMemoryCache, or do we only need one chart.  If we only need one chart, the InMemoryCache could simply be `var InMemoryCache InMemoryChart`

pkg/downloader/chart_downloader.go
marckhouzam: Sorry, I wasn't clear.  

When downloading dependencies, is it possible that we will need to download multiple versions of the same chart name?  I don't know enough about dependencies to answer that.  We need another maintainer.

I'm asking because, in the code
```
InMemoryCache[ref] = InMemoryChart{
```
the variable `ref` is just the chart name (does not include the version I believe).  So, if we need to download multiple versions of the same chart name, then the index of `InMemoryCache` needs to contain both  the name and the version of the chart.

Say we download `stable/grafana` version 1.6 and then `stable/grafana` version 1.2.  The second download will replace the first since the same name is used as `ref`.

pkg/action/install.go
```diff
@@ -736,21 +749,22 @@ OUTER:
...
+				return abs, err
+			}
+			if c.Verify {
+				if _, err := downloader.VerifyChart(abs, c.Keyring); err != nil {
+					return "", err
+				}
 			}
+			return abs, nil
+		}
+		if _, err := os.Stat(name); err == nil && (filepath.IsAbs(name) || strings.HasPrefix(name, ".")) {
```
marckhouzam, line 765: I don't understand this code.  If we get here, we know that the if statement had `os.Stat(name)` return an error.  So why are we checking it in this if statement?

pkg/action/install.go
```diff
@@ -105,15 +106,27 @@ type Install struct {
...
+	CaFile           string // --ca-file
+	CertFile         string // --cert-file
+	KeyFile          string // --key-file
+	Keyring          string // --keyring
+	Password         string // --password
+	RepoURL          string // --repo
+	Username         string // --username
+	Verify           bool   // --verify
+	Version          string // --version
+	UseInMemoryCache bool   // --use-inmemory-cache
```
marckhouzam, line 118: To be more generic, how about naming this `readOnlyFS`?

pkg/downloader/chart_downloader.go
```diff
@@ -318,6 +377,19 @@ func VerifyChart(path, keyring string) (*provenance.Verification, error) {
 	return sig.Verify(path, provfile)
 }
 
+// VerifyInMemoryChart takes key to a chart in the the InMemoryCache and a keyring,
+// and verifies the chart.
+//
+// It assumes that a chart data is accompanied by a embedded provenance file whose
+// name is the archive file name plus the ".prov" extension.
+func VerifyInMemoryChart(ref, keyring string, chartData, provData *bytes.Buffer) (*provenance.Verification, error) {
```
marckhouzam, line 385: Let's make this function private for now (lower case first letter).

marckhouzam (changes requested):

marckhouzam:
Thanks for the update @waveywaves.  We're getting closer 😄 
I've put new comments after reviewing the latest version.

pkg/action/install.go
```diff
@@ -776,6 +790,11 @@ func (c *ChartPathOptions) LocateChart(name string, settings *cli.EnvSettings) (
 		name = chartURL
 	}
 
+	if c.UseInMemoryCache {
+		debug("Downloading ", name, " to InMemoryCache")
```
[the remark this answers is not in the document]
waveywaves: Updated

pkg/downloader/chart_downloader.go
```diff
@@ -158,6 +159,64 @@ func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *proven
 	return destfile, ver, nil
 }
 
+type InMemoryChart struct {
+	ChartData              *bytes.Buffer
+	Version                string
+	ProvenanceVerification *provenance.Verification
+}
+
+var InMemoryCache = make(map[string]InMemoryChart)
```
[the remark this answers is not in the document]
waveywaves: I have made these two private and then have added a getter to call from the client side.

pkg/action/install.go
```diff
@@ -105,15 +106,27 @@ type Install struct {
...
+	RepoURL          string // --repo
+	Username         string // --username
+	Verify           bool   // --verify
+	Version          string // --version
+	UseInMemoryCache bool   // --use-inmemory-cache
+}
+
+var (
+	settings = cli.New()
+)
```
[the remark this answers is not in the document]
waveywaves: removed

pkg/action/install.go
```diff
@@ -105,15 +106,27 @@ type Install struct {
...
+
+func debug(format string, v ...interface{}) {
+	if settings.Debug {
+		format = fmt.Sprintf("[debug] %s\n", format)
+		log.Output(2, fmt.Sprintf(format, v...))
+	}
```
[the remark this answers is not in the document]
waveywaves: removed

pkg/action/install.go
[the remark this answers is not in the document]
waveywaves: Updated

pkg/downloader/chart_downloader.go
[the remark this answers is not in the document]
waveywaves: Done

pkg/action/install.go
waveywaves: You are correct, this augmentation of if statement is unnecessary as os.Stat is being checked before as well.

pkg/action/install.go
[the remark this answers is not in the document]
waveywaves: I have updated this by removing the check, I had introduced it in the first place as I was facing issues with provenance checking with local charts. Not sure how it is working again.

pkg/action/install.go
waveywaves: Updated, this is a great idea !

pkg/downloader/chart_downloader.go
waveywaves: That makes sense, the ref should be the name and the version of the chart. But considering that in the code the chart is read solely based on the name without the version so I guess installing multiple versions of the same chart or having them as dependencies is not a scenario considered here.

pkg/downloader/chart_downloader.go
waveywaves: Done

pkg/downloader/chart_downloader.go
```diff
@@ -158,6 +159,66 @@ func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *proven
 	return destfile, ver, nil
 }
 
+type inMemoryChart struct {
+	ChartData              *bytes.Buffer
+	Version                string
+	ProvenanceVerification *provenance.Verification
+}
+
+type InMemoryChartList map[string]inMemoryChart
```
marckhouzam, line 168: Please remove.  I don't want to expose any detail about the cache.  Please see comment on `GetInMemoryCache()` for an explanation

pkg/downloader/chart_downloader.go
```diff
@@ -392,3 +466,7 @@ func loadRepoConfig(file string) (*repo.File, error) {
 	}
 	return r, nil
 }
+
+func GetInMemoryCache() InMemoryChartList {
```
marckhouzam, line 470: This still exposes the details of the cache.  All the client needs is a `*chart.Chart` to be returned.
Please replace this method with:
```
func GetChart(chart string) *chart.Chart, error {
     chartData := inMemoryCache[chart].ChartData
     return loader.LoadArchive(chartData)
}
```

But I would still like this to be part of the `loader` package somehow, if possible.

pkg/provenance/sign.go
```diff
@@ -277,12 +277,57 @@ func (s *Signatory) Verify(chartpath, sigpath string) (*Verification, error) {
 	return ver, nil
 }
 
+// VerifyFromData checks the signature and verifies that it is legit for a chart (not stored in the filesystem)
+func (s *Signatory) VerifyFromData(chartRef string, chartData, sigData *bytes.Buffer) (*Verification, error) {
```
marckhouzam, line 281: There is a lot of code duplication between `Verify()` and `VerifyFromData()`.

What you could do is move the first part of the code from `decodeSignature()` into `Verify()` so as to extract the chartData and the copy the first part of `Digest()` in `Verify(), so that you can then have `Verify()` call `VerifyFromData()` passing it the chartData and sigData.

marckhouzam (changes requested):

pkg/downloader/chart_downloader.go
```diff
@@ -158,6 +159,66 @@ func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *proven
...
+			ver, err = verifyInMemoryChart(ref, c.Keyring, chartData, provData)
+			if err != nil {
+				// Fail always in this case, since it means the verification step
+				// failed.
+				return ref, ver, err
+			}
+		}
+	}
+
+	inMemoryCache[ref] = inMemoryChart{
```
marckhouzam, line 213: Getting back to the concern that `ref` ignores the chart version.

I'm under the impression that when a chart is downloaded, the version is taken into account.  See here
https://github.com/helm/helm/blob/master/pkg/downloader/chart_downloader.go#L257

If this is true, then we need to know if multiple versions of the same chart can be used as dependencies.  If so, the index of the cache must consider the version also.

marckhouzam, line 213: Getting back to the concern that `ref` ignores the chart version.

I'm under the impression that when a chart is downloaded, the version is taken into account.  See here
https://github.com/helm/helm/blob/master/pkg/downloader/chart_downloader.go#L257

If this is true, then we need to know if multiple versions of the same chart can be used as dependencies.  If so, the index of the cache must consider the version also.

pkg/action/install.go
```diff
@@ -776,6 +777,13 @@ func (c *ChartPathOptions) LocateChart(name string, settings *cli.EnvSettings) (
 		name = chartURL
 	}
 
+	if c.ReadOnlyFS {
+		if settings.Debug {
+			log.Output(2, fmt.Sprintf("[debug] Downloading %s to inMemoryCache\n", name))
+		}
+		name, _, err := dl.DownloadToInMemoryCache(name, version)
```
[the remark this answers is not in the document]
waveywaves: FIxed this one

pkg/downloader/chart_downloader.go
waveywaves: Had to add the new type because of the go-lint, but have made it private now

pkg/downloader/chart_downloader.go
waveywaves: I have addred the function to Get the Chart from the Cache only. I tried moving this function to loader. Considering that the downloader downloads and stores the Chart Data locally in the package, we would have to use a GetChart in Loader which gets it from the Downloader and serves it to the Client Side. The issue happening here is that it is creating an import cycle and helm won't build after.

pkg/provenance/sign.go
waveywaves: I have reduecd the footprint of duplicated code as of now. We would be able to solve the problem with sigData, and chartData but there is an issue of reading this stuff as bytes.Buffer for which we would have to ass some more code.

pkg/downloader/chart_downloader.go
waveywaves: I have updated this so that the charts are also saved by revisions. Can you PTAL ?

waveywaves: I have updated this so that the chart would be queried based on the name AND version. Can you PTAL ?

pkg/action/install.go
```diff
@@ -776,6 +777,16 @@ func (c *ChartPathOptions) LocateChart(name string, settings *cli.EnvSettings) (
 		name = chartURL
 	}
 
+	if c.ReadOnlyFS {
+		if settings.Debug {
+			log.Output(2, fmt.Sprintf("[debug] Downloading %s to inMemoryCache\n", name))
```
pedjak, line 782: Could we refactor and use https://github.com/helm/helm/blob/7f3339cb4eec9800cb09a46512050fb509480f7e/cmd/helm/helm.go#L55  instead of duplicating? or add the same function in `action` package?

pkg/downloader/chart_downloader.go
```diff
@@ -158,6 +162,71 @@ func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *proven
...
+
+// inMemoryChartList maps a chart name to a particular set of charts
+type inMemoryChartList map[string]inMemoryCharts
+
+var inMemoryCache = make(inMemoryChartList)
+
+//DownloadToInMemoryCache takes the chart reference and version and downloads it to the inMemoryCache for usage.
+//
+//The chart downloaded to the inMemoryCache can either be from the filesystem or from a URL
+func (c *ChartDownloader) DownloadToInMemoryCache(name, version string) (string, *provenance.Verification, error) {
```
pedjak, line 181: a general question: instead adding the new function and coping the most of the code from `DownloadTo` function, could we not instead modify the later so that does not write to the filesystem, e.g when `dest` arg is empty or `nil`? That would also allow CLI support immediately, by `HELM_REPOSITORY_CACHE` env cache to `""`

pkg/downloader/chart_downloader.go
```diff
@@ -158,6 +162,71 @@ func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *proven
...
+	ProvenanceVerification *provenance.Verification
+}
+
+// inMemoryCharts maps version to a particular chart
+type inMemoryCharts map[string]inMemoryChart
+
+// inMemoryChartList maps a chart name to a particular set of charts
+type inMemoryChartList map[string]inMemoryCharts
+
+var inMemoryCache = make(inMemoryChartList)
```
pedjak, line 176: I would be cautious when naming something as cache, and not providing any invalidation/replacement strategy. Currently, this is just a map that will grow indefinitely (certainly bad thing for somebody who wants to run a long-living service)

pkg/downloader/chart_downloader.go
pedjak: Perhaps using a library like https://github.com/spf13/afero (or similar) could make things easier?

pkg/action/install.go
```diff
@@ -105,15 +106,16 @@ type Install struct {
...
+	CaFile     string // --ca-file
+	CertFile   string // --cert-file
+	KeyFile    string // --key-file
+	Keyring    string // --keyring
+	Password   string // --password
+	RepoURL    string // --repo
+	Username   string // --username
+	Verify     bool   // --verify
+	Version    string // --version
+	ReadOnlyFS bool   // --use-inmemory-cache
```
pedjak, line 118: We could implement this even without introducing the new field, if we establish the convention that `settings.RepositoryCache` is `nil` or `empty`. IMHO, it makes more sense.

pedjak (commented):
IMHO, `LocalChart` function cannot return anymore the chart path on the repository cache on the local file system - it should be generalized a bit, for example it could return an URL/URI so that one can distinguish between charts available on the filesystem/in-memory/etc.

Also, the proposed datastructure grows indefinitely over time - not important for CLI invocations, but very important for someone implementing long-lived service on top of Helm.

pkg/action/install.go
waveywaves: I don't know if settings.RepositoryCache can be empty because lazypath sets this by default AFAIK.

pkg/downloader/chart_downloader.go
waveywaves: I dont know about this, currently in Helm we follow the XDG MIME convention, this might interfere with that. @mattfarina Can you comment on this if possible ?

pkg/action/install.go
waveywaves: This is used only in one place so defining a function would increase the symbol footprint unnecessarilty I feel.

pkg/downloader/chart_downloader.go
waveywaves: The current implementation works in a content addressable way. Stuff is not added on top of the existing continuously. But I agree that considering that variable here is called cache there needs to be a invalidation/replacement strategy.

pedjak: If I have a service that install various charts over the time this map will just grow in size. I do not see any place in this PR that removes entries from it.

pkg/action/install.go
pedjak: If I read it correctly here: https://github.com/helm/helm/blob/14f6d1ea97eeef158adf7db0e9b42206905930bf/pkg/cli/environment.go#L95 , an env var could be defined in empty, and this value will used.

waveywaves: With this though the behavior will be implicit and not explicit right ?

waveywaves: I think for now let's try to keep it explicit functionality. @marckhouzam WDYT ?

pedjak: Well, it depends :) `HELM_REPOSITORY_CACHE` env var is there so that user can customize the cache location. Now, we can set it to empty and with that turn off cache fully, or think about some other special value (e.g. `off`, or `none`) and use it later in the code.

yinzara: I think having a specialty value like "off" or "none" is the best solution. Allows us to to reuse the current flag and configuration.

marckhouzam: I really like the idea of setting HELM_REPOSITORY_CACHE to "" and avoiding the extra field.  Thanks @pedjak for the idea!

I don't know if there would be side-effects in other areas using the cache, but ultimately, the goal is to not write anything to the file-system, so the cache should never be used.

@waveywaves can you try this approach and see how it goes?

marckhouzam:
> Also, the proposed datastructure grows indefinitely over time - not important for CLI invocations, but very important for someone implementing long-lived service on top of Helm.

Good point!  We should probably use a  Least Recently Used (LRU) cache.

pkg/action/install.go
waveywaves: @marckhouzam I will try this out.

waveywaves:
@marckhouzam @pedjak Do check it out. setting HELM_REPOSITORY_CACHE to "off" yeilds the desired result. And the cache has been updated to a LRU Cache.

cmd/helm/install.go
```diff
@@ -182,7 +184,12 @@ func runInstall(args []string, client *action.Install, valueOpts *values.Options
 	}
 
 	// Check chart dependencies to make sure all are present in /charts
-	chartRequested, err := loader.Load(cp)
+	var chartRequested *chart.Chart
+	if os.Getenv("HELM_REPOSITORY_CACHE") == "off" {
```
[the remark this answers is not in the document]
waveywaves: I agree. Based on our last discussion I will try working this out.

PR #7602, opened 2020-02-15
feat(comp): Move kube-context completion to Go

marckhouzam:
This moves that last custom completion done in Bash to Go.
This actually removes the last dependency to `kubectl`.  

I hadn't done that before because the call to `action.Init()` would fail before doing the completion, when the `--kube-context` value was partial (because it was being completed).  But I found a way around that by looking for that specific scenario, as implemented in this PR.


marckhouzam:
Corresponding acceptance tests here: https://github.com/helm/acceptance-testing/pull/76

bacongobbler (approved):
LGTM!

PR #7607, opened 2020-02-17
Remove log in sympath

rkt2spc:
Remove this log so `helm template` can work properly.

### DIFF
--- a/internal/sympath/walk.go
+++ b/internal/sympath/walk.go
@@ -21,7 +21,6 @@ limitations under the License.
 package sympath
 
 import (
-	"log"
 	"os"
 	"path/filepath"
 	"sort"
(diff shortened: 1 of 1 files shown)


hickeyma (commented):
@rocketspacer Can you explain what this PR is supposed to fix or do?

rocketspacer:
@hickeyma this line always write text to stdout whenever I do `helm template`
making the output un-usable for further operations like `helm template | k apply -k`

bacongobbler:
To give some background context on this change, it was introduced as part of a CVE during the Helm security audit. More information can be found here: https://helm.sh/blog/2019-10-30-helm-symlink-security-notice/

Based on the feedback provided to us by the security team, this warning message should not be removed. Perhaps a better resolution to this may be to work on #6844, which then users could strip this log line by using a lower verbosity level such as helm package -v.

What do you think?

bacongobbler:
FYI, pkg `log` prints these log messages on stderr, not stdout. Piping stderr to /dev/null will remove the output.

https://golang.org/pkg/log/

rocketspacer:
@bacongobbler Your method works :)
`helm template release-123 . 2> /dev/null | grep walk.go`

PR #7613, opened 2020-02-17
Add support for using OCI references as the dependency repository

### LINKED ISSUE #6593: Helm3: Use registry as dependency URL
Helm 3 has a new chart command to push/pull charts to a OCI registry. This seems to be an alternative to repositories to maintain dependencies. But ATM you have to retrieve each of the charts manually:

```
helm3 chart pull registry.local/application/helm-chart:latest
helm3 chart export registry.local/application/helm-chart:latest
```

I would expect that helm could manage the charts retrieved from OCI registries in the same way it pulls dependencies from chart repositories. Therefore chart dependencies in OCI registries should be declared in Charts.yaml in the same way:

```
dependencies: 
  - name: my-application
    version: 1.0.0
    condition: my-application.enabled
    repository: registry.local/application/helm-chart:latest
```

However this currently just causes an error.

Do I misunderstand the concept of OCI image/registry support for charts? Or is the additional support for OCI registries as dependency just a missing feature?

phroggyy:
## Summary

This PR introduces the ability to specify OCI references as the `repository` for chart dependencies. This means it will be possible to specify:
```yaml
# Chart.yaml
dependencies:
- name: nginx
  version: "1.2.3"
  repository: "oci://nginx:latest"
```

This MR *does not* (yet anyway) add support for directly installing OCI-compatible images via `helm install`, i.e `helm install --name nginx oci://nginx`, although that addition _should_ be fairly easy.

### Versions

Currently, the helm versioning system relies on SemVer. This is great, and something I think should definitely be kept. *However*, with OCI registries, we have no way of determining a version or any way of indexing. Thus, we have some options:

1. Ignore the version: this is what the current code does, and I don't like it. This will rely solely on the image name and tag for the download.
2. Enforce SemVer tagging and disallow image tags in the image reference. This would mean that the download would just use the version as the image tag. e.g `version: 1.10` and `repository: oci://nginx` would download `nginx:1.10`. I'm not a fan of enforcing this SemVer tagging, as it makes Helm's system less flexible than what image tags allow, and may disrupt workflows for application charts that developers might be used to (e.g using the commit ref and `latest` as tags)
3. Require _either_ version _or_ a tag. This would mean that if no tag is specified, users are required to specify the `version`, which will then be used as the image tag. If a tag is provided, we'll use that, and not require the version to be specified (as it will be whatever image was specified). This is my personal favourite. One downside to this is that the "either or" might be confusing, and I can see it confusing developers if using both version and tag. Or if a tag is not specified, knowing that the version is what's being used.

I'd love to see some discussion around the versioning, as that's currently the main problem I can see with an OCI implementation. As mentioned, my preferred option is _some variation_ of the third option. Most importantly, I'd like to allow Helm users to use specific image tags shall they wish to do so.

## Limitations

It is not possible to do provenance verification at download-time (meaning this would be incompatible with `install --verify`) as no provenance file is present in the image.

## Notes

Resolves #6593

### DIFF
--- a/cmd/helm/dependency.go
+++ b/cmd/helm/dependency.go
@@ -71,6 +71,15 @@ the dependency charts stored locally. The path should start with a prefix of
 If the dependency chart is retrieved locally, it is not required to have the
 repository added to helm by "helm add repo". Version matching is also supported
 for this case.
+
+Starting from 3.3.0, if OCI Registry support has been enabled via the HELM_EXPERIMENTAL_OCI
+flag, repository can be defined as an OCI image reference. The path should start with a
+prefix of "oci://". For example,
+
+    # Chart.yaml
+    dependencies:
+    - version: "1.2.3"
+      repository: "oci://localhost:5000/myrepo/mychart:2.7.0"
 `
 
 const dependencyListDesc = `
@@ -82,7 +91,7 @@ the contents of a chart.
 This will produce an error if the chart cannot be loaded.
 `
 
-func newDependencyCmd(out io.Writer) *cobra.Command {
+func newDependencyCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:     "dependency update|build|list",
 		Aliases: []string{"dep", "dependencies"},
@@ -92,8 +101,8 @@ func newDependencyCmd(out io.Writer) *cobra.Command {
 	}
 
 	cmd.AddCommand(newDependencyListCmd(out))
-	cmd.AddCommand(newDependencyUpdateCmd(out))
-	cmd.AddCommand(newDependencyBuildCmd(out))
+	cmd.AddCommand(newDependencyUpdateCmd(cfg, out))
+	cmd.AddCommand(newDependencyBuildCmd(cfg, out))
 
 	return cmd
 }
--- a/cmd/helm/dependency_update.go
+++ b/cmd/helm/dependency_update.go
@@ -22,6 +22,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"helm.sh/helm/v3/cmd/helm/require"
+	"helm.sh/helm/v3/internal/experimental/registry"
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/downloader"
 	"helm.sh/helm/v3/pkg/getter"
@@ -43,7 +44,7 @@ in the Chart.yaml file, but (b) at the wrong version.
 `
 
 // newDependencyUpdateCmd creates a new dependency update command.
-func newDependencyUpdateCmd(out io.Writer) *cobra.Command {
+func newDependencyUpdateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 	client := action.NewDependency()
 
 	cmd := &cobra.Command{
@@ -57,12 +58,16 @@ func newDependencyUpdateCmd(out io.Writer) *cobra.Command {
 			if len(args) > 0 {
 				chartpath = filepath.Clean(args[0])
 			}
+			getters := getter.All(settings)
+			if FeatureGateOCI.IsEnabled() {
+				getters = append(getters, registry.NewRegistryGetterProvider(cfg.RegistryClient))
+			}
 			man := &downloader.Manager{
 				Out:              out,
 				ChartPath:        chartpath,
 				Keyring:          client.Keyring,
 				SkipUpdate:       client.SkipRefresh,
-				Getters:          getter.All(settings),
+				Getters:          getters,
 				RepositoryConfig: settings.RepositoryConfig,
 				RepositoryCache:  settings.RepositoryCache,
 				Debug:            settings.Debug,
--- a/pkg/chartutil/save.go
+++ b/pkg/chartutil/save.go
@@ -21,6 +21,7 @@ import (
 	"compress/gzip"
 	"encoding/json"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"time"
@@ -124,28 +125,50 @@ func Save(c *chart.Chart, outDir string) (string, error) {
 		return "", err
 	}
 
-	// Wrap in gzip writer
-	zipper := gzip.NewWriter(f)
-	zipper.Header.Extra = headerBytes
-	zipper.Header.Comment = "Helm"
-
-	// Wrap in tar writer
-	twriter := tar.NewWriter(zipper)
 	rollback := false
+	// save the chart to the file
+	err = Write(c, f)
+
+	if err != nil {
+		rollback = true
+		return filename, err
+	}
+
 	defer func() {
-		twriter.Close()
-		zipper.Close()
 		f.Close()
 		if rollback {
 			os.Remove(filename)
 		}
 	}()
 
-	if err := writeTarContents(twriter, c, ""); err != nil {
-		rollback = true
-		return filename, err
-	}
 	return filename, nil
+}
+
+// Write streams an archived chart to an io.writer interface.
+//
+// This takes an existing chart and a destination writer.
+func Write(c *chart.Chart, w io.Writer) error {
+	if err := c.Validate(); err != nil {
+		return errors.Wrap(err, "chart validation")
+	}
+
+	// Wrap in gzip writer
+	zipper := gzip.NewWriter(w)
+	zipper.Header.Extra = headerBytes
+	zipper.Header.Comment = "Helm"
+
+	// Wrap in tar writer
+	twriter := tar.NewWriter(zipper)
+	defer func() {
+		twriter.Close()
+		zipper.Close()
+	}()
+
+	if err := writeTarContents(twriter, c, ""); err != nil {
+		return err
+	}
+
+	return nil
 }
 
 func writeTarContents(out *tar.Writer, c *chart.Chart, prefix string) error {
--- a/pkg/downloader/chart_downloader.go
+++ b/pkg/downloader/chart_downloader.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/pkg/errors"
 
+	"helm.sh/helm/v3/internal/experimental/registry"
 	"helm.sh/helm/v3/internal/fileutil"
 	"helm.sh/helm/v3/internal/urlutil"
 	"helm.sh/helm/v3/pkg/getter"
@@ -94,12 +95,31 @@ func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *proven
 		return "", nil, err
 	}
 
-	data, err := g.Get(u.String(), c.Options...)
+	downloadURL := u.String()
+	name := filepath.Base(u.Path)
+
+	if _, ok := g.(*registry.Getter); ok {
+		parts := strings.Split(filepath.Base(u.Path), ":")
+
+		if len(parts) == 1 && version == "" {
+			return "", nil, errors.New("no version or tag provided")
+		}
+
+		if len(parts) != 2 {
+			parts = append(parts, version)
+			u.Path = fmt.Sprintf("%s:%s", u.Path, version)
+		}
+
+		downloadURL = u.String()
+		name = fmt.Sprintf("%s-%s.tgz", parts[0], parts[1])
+	}
+
+	data, err := g.Get(downloadURL, c.Options...)
+
 	if err != nil {
 		return "", nil, err
 	}
 
-	name := filepath.Base(u.Path)
 	destfile := filepath.Join(dest, name)
 	if err := fileutil.AtomicWriteFile(destfile, data, 0644); err != nil {
 		return destfile, nil, err
--- a/pkg/downloader/manager.go
+++ b/pkg/downloader/manager.go
@@ -31,6 +31,7 @@ import (
 	"github.com/pkg/errors"
 	"sigs.k8s.io/yaml"
 
+	"helm.sh/helm/v3/internal/experimental/registry"
 	"helm.sh/helm/v3/internal/resolver"
 	"helm.sh/helm/v3/internal/third_party/dep/fs"
 	"helm.sh/helm/v3/internal/urlutil"
@@ -306,7 +307,7 @@ func (m *Manager) downloadAll(deps []*chart.Dependency) error {
 			},
 		}
 
-		if _, _, err := dl.DownloadTo(churl, "", destPath); err != nil {
+		if _, _, err := dl.DownloadTo(churl, dep.Version, destPath); err != nil {
 			saveError = errors.Wrapf(err, "could not download %s", churl)
 			break
 		}
@@ -397,6 +398,11 @@ Loop:
 	for _, dd := range deps {
 		// If repo is from local path, continue
 		if strings.HasPrefix(dd.Repository, "file://") {
+			continue
+		}
+
+		// if repo is an OCI registry, continue
+		if strings.HasPrefix(dd.Repository, fmt.Sprintf("%s://", registry.OCIProtocol)) {
 			continue
 		}
 
@@ -545,7 +551,7 @@ func (m *Manager) parallelRepoUpdate(repos []*repo.Entry) error {
 // repoURL is the repository to search
 //
 // If it finds a URL that is "relative", it will prepend the repoURL.
-func (m *Manager) findChartURL(name, version, repoURL string, repos map[string]*repo.ChartRepository) (url, username, password string, err error) {
+func (m *Manager) findChartURL(name, version, repoURL string, repos map[string]*repo.ChartRepository) (chartURL, username, password string, err error) {
 	for _, cr := range repos {
 		if urlutil.Equal(repoURL, cr.Config.URL) {
 			var entry repo.ChartVersions
@@ -558,7 +564,7 @@ func (m *Manager) findChartURL(name, version, repoURL string, repos map[string]*
 			if err != nil {
 				return
 			}
-			url, err = normalizeURL(repoURL, ve.URLs[0])
+			chartURL, err = normalizeURL(repoURL, ve.URLs[0])
 			if err != nil {
 				return
 			}
@@ -567,12 +573,23 @@ func (m *Manager) findChartURL(name, version, repoURL string, repos map[string]*
 			return
 		}
 	}
-	url, err = repo.FindChartInRepoURL(repoURL, name, version, "", "", "", m.Getters)
+
+	u, err := url.ParseRequestURI(repoURL)
+	if err != nil {
+		return
+	}
+
+	if u.Scheme == registry.OCIProtocol {
+		chartURL = repoURL
+		return
+	}
+
+	chartURL, err = repo.FindChartInRepoURL(repoURL, name, version, "", "", "", m.Getters)
 	if err == nil {
 		return
 	}
 	err = errors.Errorf("chart %s not found in %s", name, repoURL)
-	return
+	return chartURL, username, password, err
 }
 
 // findEntryByName finds an entry in the chart repository whose name matches the given name.
--- a/cmd/helm/dependency_build.go
+++ b/cmd/helm/dependency_build.go
@@ -24,6 +24,7 @@ import (
 	"k8s.io/client-go/util/homedir"
 
 	"helm.sh/helm/v3/cmd/helm/require"
+	"helm.sh/helm/v3/internal/experimental/registry"
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/downloader"
 	"helm.sh/helm/v3/pkg/getter"
@@ -40,7 +41,7 @@ If no lock file is found, 'helm dependency build' will mirror the behavior
 of 'helm dependency update'.
 `
 
-func newDependencyBuildCmd(out io.Writer) *cobra.Command {
+func newDependencyBuildCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 	client := action.NewDependency()
 
 	cmd := &cobra.Command{
@@ -53,11 +54,15 @@ func newDependencyBuildCmd(out io.Writer) *cobra.Command {
 			if len(args) > 0 {
 				chartpath = filepath.Clean(args[0])
 			}
+			getters := getter.All(settings)
+			if FeatureGateOCI.IsEnabled() {
+				getters = append(getters, registry.NewRegistryGetterProvider(cfg.RegistryClient))
+			}
 			man := &downloader.Manager{
 				Out:              out,
 				ChartPath:        chartpath,
 				Keyring:          client.Keyring,
-				Getters:          getter.All(settings),
+				Getters:          getters,
 				RepositoryConfig: settings.RepositoryConfig,
 				RepositoryCache:  settings.RepositoryCache,
 				Debug:            settings.Debug,
--- a/cmd/helm/root.go
+++ b/cmd/helm/root.go
@@ -141,7 +141,7 @@ func newRootCmd(actionConfig *action.Configuration, out io.Writer, args []string
 	cmd.AddCommand(
 		// chart commands
 		newCreateCmd(out),
-		newDependencyCmd(out),
+		newDependencyCmd(actionConfig, out),
 		newPullCmd(out),
 		newShowCmd(out),
 		newLintCmd(out),
--- a/go.mod
+++ b/go.mod
@@ -26,6 +26,7 @@ require (
 	github.com/mitchellh/copystructure v1.0.0
 	github.com/opencontainers/go-digest v1.0.0-rc1
 	github.com/opencontainers/image-spec v1.0.1
+	github.com/phayes/freeport v0.0.0-20180830031419-95f893ade6f2
 	github.com/pkg/errors v0.9.1
 	github.com/rubenv/sql-migrate v0.0.0-20200212082348-64f95ea68aa3
 	github.com/sirupsen/logrus v1.4.2
--- a/internal/experimental/registry/client_test.go
+++ b/internal/experimental/registry/client_test.go
@@ -22,7 +22,6 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
-	"net"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
@@ -33,12 +32,12 @@ import (
 	"time"
 
 	"github.com/containerd/containerd/errdefs"
-
 	auth "github.com/deislabs/oras/pkg/auth/docker"
 	"github.com/docker/distribution/configuration"
 	"github.com/docker/distribution/registry"
 	_ "github.com/docker/distribution/registry/auth/htpasswd"
 	_ "github.com/docker/distribution/registry/storage/driver/inmemory"
+	"github.com/phayes/freeport"
 	"github.com/stretchr/testify/suite"
 	"golang.org/x/crypto/bcrypt"
 
@@ -107,7 +106,7 @@ func (suite *RegistryClientTestSuite) SetupSuite() {
 
 	// Registry config
 	config := &configuration.Configuration{}
-	port, err := getFreePort()
+	port, err := freeport.GetFreePort()
 	suite.Nil(err, "no error finding free port for test registry")
 	suite.DockerRegistryHost = fmt.Sprintf("localhost:%d", port)
 	config.HTTP.Addr = fmt.Sprintf(":%d", port)
@@ -252,21 +251,6 @@ func (suite *RegistryClientTestSuite) Test_8_ManInTheMiddle() {
 
 func TestRegistryClientTestSuite(t *testing.T) {
 	suite.Run(t, new(RegistryClientTestSuite))
-}
-
-// borrowed from https://github.com/phayes/freeport
-func getFreePort() (int, error) {
-	addr, err := net.ResolveTCPAddr("tcp", "localhost:0")
-	if err != nil {
-		return 0, err
-	}
-
-	l, err := net.ListenTCP("tcp", addr)
-	if err != nil {
-		return 0, err
-	}
-	defer l.Close()
-	return l.Addr().(*net.TCPAddr).Port, nil
 }
 
 func initCompromisedRegistryTestServer() string {
--- a/internal/experimental/registry/constants.go
+++ b/internal/experimental/registry/constants.go
@@ -22,6 +22,9 @@ const (
 
 	// HelmChartContentLayerMediaType is the reserved media type for Helm chart package content
 	HelmChartContentLayerMediaType = "application/tar+gzip"
+
+	// OCIProtocol is the protocol used for OCI registry URLs
+	OCIProtocol = "oci"
 )
 
 // KnownMediaTypes returns a list of layer mediaTypes that the Helm client knows about
--- a/internal/experimental/registry/getter.go
+++ b/internal/experimental/registry/getter.go
@@ -1,0 +1,78 @@
+/*
+Copyright The Helm Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package registry // import "helm.sh/helm/v3/internal/experimental/registry"
+
+import (
+	"bytes"
+	"net/url"
+
+	"helm.sh/helm/v3/pkg/chartutil"
+	"helm.sh/helm/v3/pkg/getter"
+)
+
+// Getter is the HTTP(/S) backend handler for OCI image registries.
+type Getter struct {
+	Client *Client
+}
+
+func NewRegistryGetter(c *Client) *Getter {
+	return &Getter{Client: c}
+}
+
+func NewRegistryGetterProvider(c *Client) getter.Provider {
+	return getter.Provider{
+		Schemes: []string{OCIProtocol},
+		New: func(options ...getter.Option) (g getter.Getter, e error) {
+			return NewRegistryGetter(c), nil
+		},
+	}
+}
+
+func (g *Getter) Get(href string, options ...getter.Option) (*bytes.Buffer, error) {
+	u, err := url.Parse(href)
+
+	if err != nil {
+		return nil, err
+	}
+
+	ref, err := ParseReference(u.Host + u.Path)
+
+	if err != nil {
+		return nil, err
+	}
+
+	// first we'll pull the chart
+	err = g.Client.PullChart(ref)
+
+	if err != nil {
+		return nil, err
+	}
+
+	// once we know we have the chart, we'll load up the chart
+	c, err := g.Client.LoadChart(ref)
+
+	if err != nil {
+		return nil, err
+	}
+
+	buf := bytes.NewBuffer(nil)
+
+	// lastly, we'll write the tarred and gzipped contents of the chart to our output buffer
+	err = chartutil.Write(c, buf)
+
+	return buf, err
+}
--- a/internal/experimental/registry/getter_test.go
+++ b/internal/experimental/registry/getter_test.go
@@ -1,0 +1,129 @@
+/*
+Copyright The Helm Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package registry
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+
+	auth "github.com/deislabs/oras/pkg/auth/docker"
+	"github.com/docker/distribution/configuration"
+	"github.com/docker/distribution/registry"
+	"github.com/phayes/freeport"
+	"github.com/stretchr/testify/assert"
+	"golang.org/x/crypto/bcrypt"
+
+	"helm.sh/helm/v3/pkg/chart"
+	"helm.sh/helm/v3/pkg/chart/loader"
+)
+
+func TestValidRegistryUrlWithImageTag(t *testing.T) {
+	os.RemoveAll(testCacheRootDir)
+	os.Mkdir(testCacheRootDir, 0700)
+
+	var out bytes.Buffer
+	credentialsFile := filepath.Join(testCacheRootDir, CredentialsFileBasename)
+
+	client, err := auth.NewClient(credentialsFile)
+	assert.Nil(t, err, "no error creating auth client")
+
+	resolver, err := client.Resolver(context.Background(), http.DefaultClient, false)
+	assert.Nil(t, err, "no error creating resolver")
+
+	// create cache
+	cache, err := NewCache(
+		CacheOptDebug(true),
+		CacheOptWriter(&out),
+		CacheOptRoot(filepath.Join(testCacheRootDir, CacheRootDir)),
+	)
+	assert.Nil(t, err, "no error creating cache")
+
+	// init test client
+	registryClient, err := NewClient(
+		ClientOptDebug(true),
+		ClientOptWriter(&out),
+		ClientOptAuthorizer(&Authorizer{
+			Client: client,
+		}),
+		ClientOptResolver(&Resolver{
+			Resolver: resolver,
+		}),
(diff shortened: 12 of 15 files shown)


cmd/helm/dependency_update.go
```diff
@@ -16,6 +16,7 @@ limitations under the License.
 package main
 
 import (
+	"helm.sh/helm/v3/pkg/getter"
```
jdolitsky: this should go back down at bottom of imports

pkg/getter/getter.go
```diff
@@ -18,10 +18,9 @@ package getter
 
 import (
 	"bytes"
-
 	"github.com/pkg/errors"
-
 	"helm.sh/helm/v3/pkg/cli"
+	"net/url"
```
jdolitsky: stdlib imports at the top

pkg/getter/registrygetter.go
```diff
@@ -0,0 +1,49 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package getter
+
+import (
+	"bytes"
+	"helm.sh/helm/v3/internal/experimental/registry"
```
jdolitsky: put the helm.sh at bottom of imports, separated by newline

pkg/getter/registrygetter.go
```diff
@@ -0,0 +1,49 @@
+/*
```
jdolitsky: perhaps this package can be moved into internal with the rest of the registry stuff

jdolitsky (changes requested):
Few minor comments, but overall looks good!

jdolitsky:
@phroggyy looks like CI is failing - can you try running tests locally?

pkg/getter/registrygetter.go
phroggyy: see the TODO for this – I'm intending to just drop this file entirely and make the internal getter.go hold all this. The reason it was done this way originally was because of an import loop, but that's since been resolved

jdolitsky:
Im personally a fan of #3 suggestion, "require either version or a tag. This would mean that if no tag is specified, users are required to specify the version"

that sounds reasonable to me. this is actually how it works now with `helm chart save`

internal/test/net.go
```diff
@@ -0,0 +1,34 @@
+/*
```
jdolitsky: I'm not sure we should add an entire package just for this method. I dont think there are other places in the codebase that have subpackages just for test utils. I'd suggest either copy paste the freeport code whereever its needed or simply add phayes/freeport as a new dependency

jdolitsky (changes requested):

internal/test/net.go
phroggyy: I only added this because the `internal/test` package already exists – it currently contains the `HelmHome` and `TempDir` functions, so I don't see a reason not  to add this to the package too.

jdolitsky: i see it there now, ok - plz ignore

pkg/downloader/manager.go
```diff
@@ -379,6 +379,11 @@ Loop:
 			continue
 		}
 
+		// if repo is an OCI registry, continue
+		if strings.HasPrefix(dd.Repository, "oci://") {
```
jdolitsky: nit: perhaps `oci` as a protocol becomes a constant somewhere

jdolitsky (approved):
This is pretty great. Looking forward to this. LGTM

pkg/downloader/manager.go
phroggyy: fixed :)

zendril:
@phroggyy What happens when the registry is airgapped?

Today when using a chart itself, you have a default value for registry/repo:tag, and most charts allow you to override <registry>, <repo (aka image)> and <tag>. That way you can if your images need to come from an internally hosted registry (instead of public internet like docker.io), then you have that flexibility.

From what I currently see, it is nice that the dependent chart can be specified, but it is then "hard-coded" to pull from a specific registry at `helm install/pull` time.

With current charts, the tgz is completely self contained. Is that to continue to be the case?
If so, then this question is moot.

If we are expecting that the dependent charts can be dynamically resolved at install/pull time, then I think we need to account for this override scenario.

zendril:
After chat on slack, confirmed that the tgz will still remain as it is today. All transitive chart dependencies are still packed inside. This means that the OCI references are only at build time, not runtime.

All good. :)

jdolitsky:
This affects mostly just experimental stuff, so I think we are safe to merge.

Other maintainers: biggest change is here is adding methods to the `Getter` interface:
```
	Filename(url *url.URL, version string) string
	// Allow the getter to alter the download URL before retrieval
	URL(u *url.URL, version string) (string, error)
```

pkg/getter/getter.go
```diff
@@ -85,6 +86,10 @@ func WithTLSClientConfig(certFile, keyFile, caFile string) Option {
 type Getter interface {
 	// Get file content by url string
 	Get(url string, options ...Option) (*bytes.Buffer, error)
+	// Retrieve the filename given a source URL
+	Filename(url *url.URL, version string) string
```
bacongobbler: `Get` takes a string as an argument: `href`. Why do we need to discern the difference between the URL and the version here? Shouldn't this just accept an `href` as well?

pkg/getter/getter.go
```diff
@@ -85,6 +86,10 @@ func WithTLSClientConfig(certFile, keyFile, caFile string) Option {
...
+	// Allow the getter to alter the download URL before retrieval
+	URL(u *url.URL, version string) (string, error)
```
bacongobbler: why should the getter need to alter the download URL prior to a call to `Get`? Shouldn't that be done during `Get`?

bacongobbler (changes requested):
left a few questions about the design changes to the `Getter` interface in this PR

pkg/downloader/manager.go
```diff
@@ -559,12 +565,23 @@ func (m *Manager) findChartURL(name, version, repoURL string, repos map[string]*
...
+		return
+	}
+
+	chartURL, err = repo.FindChartInRepoURL(repoURL, name, version, "", "", "", m.Getters)
 	if err == nil {
 		return
 	}
 	err = errors.Errorf("chart %s not found in %s", name, repoURL)
-	return
+	return chartURL, username, password, err
```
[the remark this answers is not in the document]
phroggyy: if memory serves golang was complaining about this for whatever reason. I'll check this again, good spot

pkg/getter/getter.go
phroggyy: That would require a breaking change to the `Get` function, as we need to pass the version in to automatically use that as the image tag if it was passed (https://github.com/helm/helm/pull/7613/files#diff-4e8b0b35b9905621c4d605f9681d9366R97). Happy to discuss alternatives here

pkg/getter/getter.go
phroggyy: This is for the downloaded tgz to maintain the same naming as currently. Even if your image tag is `charts/nginx:commithash`, it will download to `nginx-{chartVersion}.tgz`. See the comment at https://github.com/helm/helm/pull/7613/files#diff-9069fe52469166bbcbe989af4b5ae3f5R85. The core problem is that we can't determine the filename to download to from the URL alone, as the URL may not contain the chart version. Both the `URL` and `Filename` functions were added to avoid having to do a bunch of case-based code in the downloader (don't want the downloader to care if it's using a registry getter or some other getter).

Happy to consider alternatives here as well.

pkg/getter/getter.go
jdolitsky: Do we need to consider adding a `func GetV2` or `type GetterV2`?

jdolitsky: or something like `func GetWithDetails` ?

phoenix-bjoern:
@bacongobbler can you please continue with the code review and merge this PR before the 3.2.0 release? This is the most important feature to get the OCI charts working. 
Thanks.

pkg/getter/getter.go
phroggyy: hm, we could do a GetterV2 or something 🤔 The problem we run into then is we need to introduce a bunch of case-based code elsewhere to check if you're using a v1 or v2 getter. We'd need to e.g pass something like a `GetterRef` that holds `Getter` and `GetterVersion` or something...

bacongobbler:
I am busy with other work at the moment and cannot continue with the review. Those comments were merely observations about what the `getter` package was originally written for, and this PR appears to change it’s originally intended behaviour. I would suggest reaching out to another maintainer for review.

pkg/getter/getter.go
jdolitsky: Consider something like
```
type (
  Details struct {
      Content  *bytes.Buffer
      Filename string
      URL      string
  }

  Getter interface {
      Get(url string, options ...Option) (*bytes.Buffer, error)
      GetDetails(url string, options ...Option) (*Details, error)
  }
)
```

Then we just change anywhere using `Get` to use `GetWithDetails` - and use `details.Content` to replace the old bytes.Buffer used

phroggyy: ah yeah we could do that. `GetWithDetails` would need to pass in the version string there, but that should be alright.

pkg/downloader/manager.go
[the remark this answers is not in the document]
jdolitsky: we're we able to determine if we can use just `return` ?

pkg/getter/getter.go
jdolitsky: I'm wondering if we can leave the same input args, and add version as an option?
```
func WithVersion(version string) Option {
	return func(opts *options) {
		opts.version = version
	}
}
```

phroggyy: I did consider that, that would possibly allow using just the Get function too... I initially decided against it since the version didn't really feel like it was inline with other options; felt like purely an implementation detail.

I'll have a thorough look at how the Getter is implemented atm (and others too) to see if there's a better way. I do agree it doesn't feel very clean the way it's done now.

DracoBlue:
Hello,

as the oci registries are called "registries" on the official docs and as sub command (https://helm.sh/docs/topics/registries/), wouldn't it be better to call it like this:

```
# Chart.yaml
dependencies:
- name: nginx
  version: "1.2.3"
  registry: "oci://nginx:latest"
```

phroggyy:
I had a look into refactoring the getter bit as had been discussed. My initial thoughts were to let the downloader set the filename, but this would be a breaking change and does indeed break existing tests. Going down the `GetWithDetails` route is feasible, but still doesn't _feel_ right to me; we're making _every other getter_ re-implement identical behaviour, just to be able to treat the registry getter differently. If we could require a tgz extension, one option would be to have the downloader substitute `:` for `-` and ensure a tgz extension. That seems like a less intrusive way of doing it, and not requiring re-implementing identical behaviour across all-but-one getters.

sajayantony:
Singularity has a similar model of `{scheme}://{name}` [reference](https://github.com/sylabs/singularity-userdocs/blob/master/singularity_and_docker.rst#building-containers-for-singularity-from-oci-images) 
There are differences in SymVer 2.0 [grammar](https://semver.org/#backusnaur-form-grammar-for-valid-semver-versions) and the [Reference spec](https://github.com/docker/distribution/blob/master/reference/reference.go) which disables certain characters like `+` so trying to represent the tag as a projection of the version will need some canonicalization which could lead to more issues. I would try to just reference the tag directly and avoid inference to avoid the similar issues like those that come due to`latest`.

jdolitsky:
@phroggyy apologies for delay here. I think the `GetWithDetails` is an acceptable solution. Could you rebase and address the DCO and Circle errors?

anschoewe:
I would love to see this merged. It would make working with OCI sub-charts/dependencies in an umbrella chart so much easier.  I'll try to learn go in the near future so I can actually help improve the tool myself.

phroggyy:
@jdolitsky sorry, been away for a couple days. I'll address those issues today.

pkg/downloader/manager.go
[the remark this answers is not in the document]
phroggyy:
```
pkg/downloader/manager.go:593:2: naked return in func `findChartURL` with 40 lines of code (nakedret)
	return
	^
```
this is what I'm getting with a naked return

phroggyy:
just to keep everyone in the loop: I have tests running and passing on local, but not on CI. Tried to sort it with @jdolitsky yesterday but didn't make progress. I'm hoping to have another look at it today. I'm currently trying to reproduce the CI results on my local.

jdolitsky:
@phroggyy - pulled down your changes and I'm unable to reproduce the failure. I re-ran the tests via Circle and failed again which leads me to believe its either a Circle problem or Go 1.13 problem - Are you using Go 1.14+ locally?

phroggyy:
I'm on 1.13.4 on local

jdolitsky:
@phroggyy I was hoping that was not the case.. I'm really not sure. If you could try rebasing, we've just updated CI to use 1.14. Thanks for your patience with this, we will get it sorted

phroggyy:
@jdolitsky good news! I have successfully reproduced the error on my local. Strangely, tests pass with `make test`, but I noticed CI runs `make test-coverage`. Running that instead of `test` **will** reliably reproduce the CI issue.

phroggyy:
Something is seriously whack. I threw some `fmt.Println` into the tests, and now they pass. My mind is blown.

TBBle:
@DracoBlue The registry is the host, but the line in the config file you're referencing is the full OCI repository (which may start with a registry, but contains also non-registry parts, and a tag).

TBBle:
What was the decision on the Versioning question? Was it resolved? (I didn't check the PR code...)

I don't like the first option. ^_^

I would be fine with the second, except that I think OCI registry tags cannot represent all of SemVer ([`+` is not allowed](https://github.com/moby/moby/issues/16304), per the [spec](https://github.com/opencontainers/distribution-spec/blob/master/spec.md#overview), which is the only reason we use `appVersion` in our local charts). We could have a consistent mangle from SemVer to tag, which _probably_ doesn't need to be reversible?

If we need tag-mangling anyway, the third option seems like it'd be easy-enough and provide user-controlled override if necessary. I equate the third version to something like `appVersion`, allowing a free-text piece of version information when appropriate.

In that case, would it make sense for a tag on the OCI to _override_ a version, rather than barfing if both are provided? That's how we are moving to use `appVersion` locally, as mentioned above.

phroggyy:
@TBBle it currently follows either or. You may specify both, in which case the version is ignored. The chart is downloaded in completeness and the tgz named according to the declared chart version, rather than the tag. So version and tag can differ, allowing you to follow semver completely in your chart versioning, just not in the image tagging (due to the problem you mentioned).

TBBle:
Is there any chance we can catch the occurrence that a version is specified, and a tag, and the tagged chart doesn't match the version?

Also, I've just realised that we have a large problem here, that we lose the SemVer match because version and tag now need to be exact. This might be a general problem with the OCI registry approach? I just realised I didn't look at how the OCI support integrates with that at the individual chart `helm chart ...` level.

phroggyy:
As there's no chart repository that can supply a list of what's available, that is indeed a limitation of using OCI registries. The only way to get around that is tagging the same chart with multiple tags, which is very common with docker images. e.g tagging the same image as `3`, `3.4`, `3.4.8`, and optionally `latest`, this is fairly common practice at least among major images.

Regarding version + tag specified, we could catch that, yes. I'd leave that as a further improvement though.

phroggyy:
Aaand we're all green! Big thanks to @jdolitsky for pointing out that the culprit might be the CI server's slowness. That turned out to be the issue, with the fix being https://github.com/helm/helm/pull/7613/files#diff-4b0e41e9f7e363114877669c8f5c38e8R161-R171

pkg/getter/getter.go
```diff
@@ -38,6 +39,11 @@ type options struct {
 	userAgent             string
 }
 
+type ChartResponse struct {
+	ChartContent *bytes.Buffer
```
jdolitsky, line 43: can we rename this to just "Content" since the type has chart in the name already?

jdolitsky (changes requested):

pkg/chartutil/save.go
```diff
@@ -116,28 +117,50 @@ func Save(c *chart.Chart, outDir string) (string, error) {
...
+		}
+	}()
+
+	return filename, nil
+}
+
+// Write streams an archived chart to an io.writer interface.
+//
+// This takes an existing chart and a destination writer.
+func Write(c *chart.Chart, w io.Writer) error {
```
phroggyy, line 142: this diff looks a bit wonky. What's happened here is that a portion of `Save` was extracted into its own function `Write`. That is then used by both `Save` (https://github.com/helm/helm/pull/7613/files#diff-9ea0c476e7eb107c6b0f198af1a49caaR122), and in https://github.com/helm/helm/pull/7613/files#diff-9069fe52469166bbcbe989af4b5ae3f5R80

pkg/downloader/manager.go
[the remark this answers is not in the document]
jdolitsky: Does it fail `make test-style` ?

jdolitsky (approved):
Looks good. Glad we were able to get the tests in a reliable state.

pkg/downloader/manager.go
```diff
@@ -400,6 +401,11 @@ Loop:
 			continue
 		}
 
+		// if repo is an OCI registry, continue
+		if strings.HasPrefix(dd.Repository, fmt.Sprintf("%s://", registry.OciProtocol)) {
```
bacongobbler, line 405: there are now 3 `if` clauses here. Can we optimize this to prevent 3 separate evaluations?

bacongobbler (changes requested):
Most of the code here looks great. Thank you for taking the time to look into this.

I'm still concerned about the design of GetWithDetails. pkg `getter` was designed to not care about the content being fetched. It's a package to fetch arbitrary content based on the URL scheme. Prior to this PR, it had no notion of the content being fetched.

In this PR, that clear separation of concerns is removed. `GetWithDetails` returns a `ChartResponse`. That goes against the original design of the package. It should not have any knowledge of the content being fetched.

Take an example where I want to fetch a plugin, like with `helm plugin install https://github.com/adamreese/helm-nuke --version 0.1.0`. The call to `GetWithDetails` would look something like this:

```
data, err := g.GetWithDetails(u, version, c.Options...)
```

Where `u` is `https://github.com/adamreese/helm-nuke`, and `version` is `0.1.0`.

If I switch the protocol to `oci://`, using the OCI protocol, this will fail because it assumes the content I am attempting to fetch is a chart, and it mangles the URL to match to the OCI protocol.

This is important to note because we have talked about fetching content other than charts using the OCI protocol, like @jdolitsky and I discussed long ago about how the OCI protocol could be used to fetch and push helm plugins.

I would remove `GetWithDetails` out of the `getter` interface entirely. The only location where `GetWithDetails` is called is when a chart is being fetched under `chart_downloader.go`. I would move the chart-specific URL logic there instead of trying to move it into pkg `getter`.

Hope this helps.

phroggyy:
@bacongobbler I was a bit reluctant to put getter specific details into a common place (like the downloader); having the downloader aware of which specific getters are used feels off. However, if that's fine with the Helm maintainers I'm happy to do that instead!

pkg/downloader/chart_downloader.go
```diff
@@ -94,14 +94,13 @@ func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *proven
 		return "", nil, err
 	}
 
-	data, err := g.Get(u.String(), c.Options...)
+	data, err := g.GetWithDetails(u, version, c.Options...)
```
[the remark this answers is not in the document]
jdolitsky: this is probably fine. makes this block a bit messier but prevents modifying the Getter interface. maybe also a comment with whats really going on

jdolitsky:
+1 to Matt's suggestion

> having the downloader aware of which specific getters are used feels off

yes, but realistically there will only be a limited number of Getter implementations in the Helm codebase

phroggyy:
I'll make these changes in the coming week. Hopefully moving those changes out of the getter interface will have the side effect of somewhat reducing the size of this diff as well 👍

phroggyy:
@jdolitsky @bacongobbler one thought that _did_ just cross my mind is if we'd introduce some `Formatter` or `ChartFormatter` interface to complement the `Getter` interface, placed next to the chart downloader. Basically allowing you to register a formatter for specific getters if they need specific chart retrieval. That _feels_ "more correct", but I'm not sure if it is, or if it's worthwhile. But it would reduce complexity of the downloader, basically doing

```
formatter := &DefaultFormatter{}
for _, f := range downloader.formatters {
    if f.handles(g) {
        formatter = f
    }
}

g.Get(formatter.DownloadUrl(u, version), ...opts)
```

It might be overcomplicating things though, but figured I'd share the idea

pkg/downloader/manager.go
phroggyy: I don't think we'll get around 3 evaluations code-wise (and the compiler should be able to optimise what can be optimised there), but we can perhaps clean it up... I'm not sure what'd be the best approach. You could do something like

```golang
specialPrefixes := map[string]string{"oci", "file"}

parts := strings.Split(dd.Repository, "://")
if _, ok := specialPrefixes[parts[0]]; ok || dd.Repository == "" {
    continue
}
```

but to me, that doesn't feel any cleaner.

We can also chunk it into a single if statement, but that won't reduce the number of evaluations.

phroggyy:
I've now implemented the concept of a download `Formatter`. After @bacongobbler's explanation of the purpose of the getter package, it felt appropriate to put inside the downloader package.

However, if you feel it's adding too much complexity for little benefit, I'm happy to cut it out and just put it right in the downloader. The main benefit it brings is allowing us to keep everything experimental completely separate in the code; it avoids cluttering the downloader with specifics of experimental features, instead allowing us to make those changes independently.

jdolitsky:
@phroggyy I think for the sake of simplicity, we should approach the problem by checking for registry getter in DownloadTo as @bacongobbler described. I'm not totally opposed to the Formatter, but it adds an additional public field. We should get the simplest thing merged, then we can follow up later with a PR if necessary

phroggyy:
dropped the formatter as requested, it's all been moved into the chart downloader now.

pkg/downloader/chart_downloader.go
```diff
@@ -17,6 +17,7 @@ package downloader
 
 import (
 	"fmt"
+	"helm.sh/helm/v3/internal/experimental/registry"
```
jdolitsky, line 20: can you place this import with the other "helm.sh" ones below?

jdolitsky (changes requested):
one small comment on imports order but otherwise lgtm

jdolitsky (approved):

bacongobbler:
Thank you.

As this is a significant feature being introduced, would you mind writing documentation covering this topic?

https://github.com/helm/helm-www/blob/master/content/en/docs/topics/registries.md
https://github.com/helm/helm-www/blob/master/content/en/docs/chart_best_practices/dependencies.md

atrauzzi:
Guys, this is so exciting.  Will definitely be using this as soon as it's ready, but I wanted to say thank you for all your efforts here.

OCI is a game changer!

guy-frontegg:
Thank you all for your hard work, I have a question please. When will this feature will be available in your opinion ?

jdolitsky:
This is in good shape, added to the 3.3.0 milestone. Hopefully available in that version @guy-frontegg

guy-frontegg:
@jdolitsky, thank you

hustclf:
I am also looking forward to this good feature,  thanks for your contribution. @guy-frontegg .

There is only a file conflict before merged, any progress?

phroggyy:
I try to regularly resolve conflicts @hustclf. Many things get merged to upstream in between so conflicts reoccur quite regularly.

bridgetkromhout:
Hi, @phroggyy! Thank you for your work on this; I would be happy to see this go forward soon. Can you take a look at the current build failure?

phroggyy:
From what I can tell there have been some changes to the linting rules, which is what's causing the current issues. I've rebased down to a single commit for ease of managing this. I'll be looking at these linting issues shortly.

phroggyy:
@jdolitsky could you check out the current linting issue? It's complaining about a `goimports` issue, but even when running the exact `goimports` command, it outputs the exact same file.

phroggyy:
@bacongobbler I think this should be ready to go

bridgetkromhout:
@phroggyy Thanks for your work on this! I think we're planning on this for 3.3.0 - unfortunately we've got another merge conflict in `internal/experimental/registry/client_test.go` - would it be possible for you to correct that? Thanks!

phroggyy:
Thanks for keeping an eye on this when I haven't @bridgetkromhout! Fixed the conflict in github but looks like the style check's failing now due to a build problem, so I'll check it properly later today.

phroggyy:
Wasn't able to get to it until the weekend, but the conflict has now been resolved and the pipeline should be passing (does on my local anyway).

hickeyma:
Tested this manually and it is looking good. Great work @phroggyy. I tested as follows:

```console
$ helm3 create newocichrt
Creating newocichrt

$ helm3 chart save newocichrt/ localhost:5000/myrepo/newocichrt
ref:     localhost:5000/myrepo/newocichrt:0.2.2
digest:  f3b22bfb323f62a46cf61f3c27da1d33d1e2ec2d21a37cea7c7957e6b6bc76b9
size:    3.5 KiB
name:    newocichrt
version: 0.2.2
0.2.2: saved

$ helm3 chart list
REF                                   	NAME      	VERSION	DIGEST 	SIZE   	CREATED   
localhost:5000/myrepo/newocichrt:0.2.2	newocichrt	0.2.2  	f3b22bf	3.5 KiB	11 seconds

$ helm3 chart push localhost:5000/myrepo/newocichrt:0.2.2
The push refers to repository [localhost:5000/myrepo/newocichrt]
ref:     localhost:5000/myrepo/newocichrt:0.2.2
digest:  f3b22bfb323f62a46cf61f3c27da1d33d1e2ec2d21a37cea7c7957e6b6bc76b9
size:    3.5 KiB
name:    newocichrt
version: 0.2.2
0.2.2: pushed to remote (1 layer, 3.5 KiB total)

$ helm3 create tstocichrt
Creating tstocichrt

$ rm -rf tstocichrt/templates/*
$ rm tstocichrt/values.yaml

$ cat tstocichrt/Chart.yaml 
apiVersion: v2
name: tstocichrt

[...]

dependencies:
    - name: newocichrt
      version: "0.2.2"
      repository: "oci://localhost:5000/myrepo/newocichrt:0.2.2"


$ helm3 dep update tstocichrt/
Saving 1 charts
Downloading newocichrt from repo oci://localhost:5000/myrepo/newocichrt:0.2.2
0.2.2: Pulling from localhost:5000/myrepo/newocichrt
ref:     localhost:5000/myrepo/newocichrt:0.2.2
digest:  f3b22bfb323f62a46cf61f3c27da1d33d1e2ec2d21a37cea7c7957e6b6bc76b9
size:    3.5 KiB
name:    newocichrt
version: 0.2.2
Status: Chart is up to date for localhost:5000/myrepo/newocichrt:0.2.2
ref:     localhost:5000/myrepo/newocichrt:0.2.2
digest:  f3b22bfb323f62a46cf61f3c27da1d33d1e2ec2d21a37cea7c7957e6b6bc76b9
size:    3.5 KiB
name:    newocichrt
version: 0.2.2
Deleting outdated charts

$ helm3 install tstocichrt tstocichrt/
NAME: tstocichrt
LAST DEPLOYED: Thu Jun 18 14:05:16 2020
NAMESPACE: default
STATUS: deployed
REVISION: 1

$ helm3 ls
NAME      	NAMESPACE	REVISION	UPDATED                                	STATUS  	CHART           	APP VERSION
tstocichrt	default  	1       	2020-06-18 14:05:16.515958406 +0000 UTC	deployed	tstocichrt-0.1.0	1.16.0 
```

hickeyma:
I was able to pull the chart from docker registry as well and it error on content as probably to be expected:

```console
$ docker pull localhost:5000/myrepo/newocichrt:0.2.2
0.2.2: Pulling from myrepo/newocichrt
f3b22bfb323f: Extracting [==================================================>]  3.564kB/3.564kB
invalid rootfs in image configuration
```

I also tested that can install charts as usual. Here is one example:

```console
$ helm3 install mysql stable/mysql
NAME: mysql
LAST DEPLOYED: Thu Jun 18 14:17:46 2020
NAMESPACE: default
STATUS: deployed
REVISION: 1
NOTES:
MySQL can be accessed via port 3306 on the following DNS name from within your cluster:
mysql.default.svc.cluster.local
[....]
```

hickeyma (approved):
This has been a great effort @phroggyy with a lot of patience. Thanks for keeping going.

LGTM with a suggestion for some additional PRs:
- Improvements in docs to provide a better end to end example (refer to https://github.com/helm/helm-www/issues/637)
- If you pull an image (e.g. `repository: "oci://localhost:5000/nginx:latest"`) using `helm dep update` it seems to corrupt the cache. It can cause an error to `helm chart` commands,  similar to the following:
```console
$ helm3 chart list
Error: blob sha256:e7f0e207a03719c72ac76264984cd607f2bbaf196b208b77dc4ccbcb44e420f5 expected at /root/.cache/helm/registry/cache/blobs/sha256/e7f0e207a03719c72ac76264984cd607f2bbaf196b208b77dc4ccbcb44e420f5: not found
```

hickeyma:
@jdolitsky  @bacongobbler Would you be happy to merge this PR?

bacongobbler:
I want to take a quick look at something first to verify before merging, but LGTM

internal/test/net.go
```diff
@@ -0,0 +1,34 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package test
+
+import "net"
+
+// borrowed from https://github.com/phayes/freeport
```
bacongobbler, line 21: This project is BSD-3 licensed. Please review the license terms.

https://github.com/phayes/freeport/blob/master/LICENSE.md

Because of the terms set out there, we cannot copy this code verbatim and claim copyright.

bacongobbler (changes requested):
Some of the code in here violates licensing terms. We cannot accept code that we did not originally write and claim that copyright for ourselves.

If you can provide a workaround that replaces this code with your own original work (or, include the project as a dependency in go.mod), then we can work around those terms. But merging this PR as-is would open us up to legal issues.

cmd/helm/dependency.go
```diff
@@ -71,6 +71,16 @@ the dependency charts stored locally. The path should start with a prefix of
...
+
+Starting from 3.3.0, if OCI Registry support has been enabled via the HELM_EXPERIMENTAL_OCI
+flag, repository can be defined as an OCI image reference. The path should start with a
+prefix of "oci://". For example,
+
+    # Chart.yaml
+    dependencies:
+    - name: nginx
+      version: "1.2.3"
+      repository: "oci://nginx:latest"
```
bacongobbler, line 83: can you update this to use a proper helm chart URL. `nginx:latest` is a docker container, not a chart. Thanks

bacongobbler (changes requested):

internal/test/net.go
hickeyma: Nice catch matt, missed that.

cmd/helm/dependency.go
hickeyma: Add a comment about this in the review also: https://github.com/helm/helm/pull/7613#pullrequestreview-433406051

internal/test/net.go
phroggyy: @bacongobbler I copied this code from another test, so I'd imagine I should be able to use the same workaround as has been used there. I'll try to find that piece of code and use that instead, good catch!

cmd/helm/dependency.go
phroggyy: Yep I brought it up in https://github.com/helm/helm-www/issues/637 after chatting with @hickeyma, I'll adjust that in this PR too

internal/test/net.go
phroggyy: For reference: it's in the current codebase https://github.com/helm/helm/pull/7613/files#diff-80ccd988dcde8b4c1b8553b6061327fdL257-L271 I'll try pulling it in as a dependency

phroggyy:
@bacongobbler fixed for your requests!

pkg/downloader/chart_downloader.go
```diff
@@ -94,12 +95,31 @@ func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *proven
...
 
-	data, err := g.Get(u.String(), c.Options...)
+	downloadURL := u.String()
+	name := filepath.Base(u.Path)
+
+	if _, ok := g.(*registry.Getter); ok {
+		parts := strings.Split(filepath.Base(u.Path), ":")
+
+		if len(parts) == 1 && version == "" {
+			return "", nil, errors.New("no version or tag provided")
```
bacongobbler, line 105: So, let me get this straight... The `repository` field is required to pass a tag in the URL. Right?

Doesn't this go against the discussion in https://github.com/helm/helm/issues/8332#issuecomment-648287823?

pkg/downloader/chart_downloader.go
```diff
@@ -94,12 +95,31 @@ func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *proven
...
+		}
+
+		if len(parts) != 2 {
```
bacongobbler, line 108: If I'm reading this right, the following will append a version to the path if there are multiple colons in the URL path, like so:

```
repository: oci://localhost:5000/testrepo/testchart:1.2.3:1.2.3
version: 1.0.0
```

Which will result in the following URL path: `/testrepo/testchart:1.2.3:1.2.3:1.0.0`

Is that right? That doesn't seem like valid input.

pkg/downloader/chart_downloader.go
```diff
@@ -94,12 +95,31 @@ func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *proven
...
+			parts = append(parts, version)
+			u.Path = fmt.Sprintf("%s:%s", u.Path, version)
+		}
+
+		downloadURL = u.String()
+		name = fmt.Sprintf("%s-%s.tgz", parts[0], parts[1])
```
bacongobbler, line 114: Giving the same example:

```
repository: oci://localhost:5000/testrepo/testchart:1.0.1:1.2.3
version: 1.0.0
```

How is this considered valid input? How is the chart downloader able to determine that `parts[1]` is the right version?

This makes me go back to the original question: why are we accepting tags in the URL path in the first place?

bacongobbler (changes requested):

pkg/downloader/chart_downloader.go
phroggyy: Multiple colons in the path isn't allowed; you can't tag an OCI image with `localhost:5000/testrepo/testchart:1.2.3:1.2.3`;

```sh
➜ docker tag nginx:1.16.0 localhost:5000/nginx:1.16.0:1.2.3
Error parsing reference: "localhost:5000/nginx:1.16.0:1.2.3" is not a valid repository/tag: invalid reference format
```

pkg/downloader/chart_downloader.go
phroggyy: That is valid input. We only error if **neither** version or tag is provided. 

`if len(parts) == 1` => no tag specified `&& version == ""` => no version specified

pkg/downloader/chart_downloader.go
phroggyy: Yep that shouldn't be valid, as that `repository` tag is invalid. As mentioned in the other comment – I can add validation for that if desired.

As to why we're accepting them: supporting the same format as `helm chart save` (as you mentioned in https://github.com/helm/helm/issues/8332#issuecomment-648305142). I think it's perfectly fine to drop support for arbitrary tags, but then we should do that for `helm chart` operations as well.

pmalic:
I've been following this PR for months and have now finally tried storing charts in a container registry. Want to share my experience, could be useful. It's an enterprise app suite that consists of eight charts, one parent chart that has no templates and seven subcharts that represent different apps/components that comprise this app suite. Each of these subcharts comes from a different team, so the parent chart has all of them listed as dependencies in Chart.yaml. 

In short, it works! :) Thanks @phroggyy for all the hard work and patience :+1:

The reason why we were interested in this PR in the first place is, of course, the ability to store the charts together with container images. The limitations with chart versioning & tags are not a problem in our case. We basically have no need for anything beyond limited version matching that already works. Yes, retrieving the list of tags via `/v2/<name>/tags/list` and doing version matching (e.g. `^1.2.3`, `'>= 1.2.3 < 1.5.0'`, …) would be cool, but I think you can live without that unless you want to run a public repo with charts from different unrelated parties that are not in sync (i.e. communicating with each other) or you’re a big company with a large number of teams. I mean, who’s most likely to start using container registries to store charts, I think people/teams who want to simplify their infrastructure, and I’d say these tend to be smaller teams. 

Having said that, if you do have plans to later add support for retrieving tags and doing version matching on them, then IMHO you should not allow tags to be specified in the `repository` key, because a tag, although mutable, always points to a single image, and that makes the `version` key somewhat useless. For example:

```
repository: oci://localhost:5000/testrepo/testchart:1.0.1
version: 1.0.0
```

This will always retrieve version 1.0.1, the `version` field in this case could only serve as some “sanity check”, i.e. verify that the chart “behind” the 1.0.1 tag is actually of version 1.0.0. This doesn’t make sense in this particular example, but would make sense here:

```
repository: oci://localhost:5000/testrepo/testchart:latest
version: '^1.0.0'
```

This should fail if the chart “behind” the `latest` tag is of e.g. version 2.0.0.

I’ve read the arguments that disabling tags would make everything less flexible, but you wouldn’t be restricting anything because no one is using this yet :) The flexibility of tags is of great importance with container images, but we’re not talking about container images here. Is there a use case for tagging Helm charts with something other than the version, e.g. `stable-alpine`? :))) I agree, `latest` is useful, but you can support it without allowing tags: if you don’t specify `version` key, then use `latest` tag (I mean, this is how image referencing generally works, if you don’t specify tag, `latest` is implied).

So, that's my two cents regarding the version/tag discussion. 

For the end, I have a specific question for @phroggyy: When doing `helm dep update` I’ve noticed that the local cache isn’t really used. By that I mean that if you have a chart in the local cache, but not yet pushed to the container registry, `dep update` will fail. Also, `dep update` will always pull the chart from the registry and always overwrite the version in the local cache. This is analogous to the `--pull` switch of docker client, but this switch is not turned on by default. This is problematic if you’re changing something, you have to push it first before you can use it. This means you have to temporarily use a different tag not used by anyone if you don’t want to accidentally break something for someone. Is there a way to control this behavior? Thanks!

pkg/downloader/chart_downloader.go
pmalic: Sorry for barging in, but isn't this shorter & more correct:
```
		if len(parts) == 1 {
			if version == "" {
				return "", nil, errors.New("no version or tag provided")
			}

			parts = append(parts, version)
			u.Path = fmt.Sprintf("%s:%s", u.Path, version)
		}
```

You wouldn't be doing `len()` twice and would only be appending version if `:` is not present in the path at all.

bacongobbler:
Given v3.3.0-rc.1 is scheduled for release tomorrow, I'm going to bump this to 3.4 as there is still some open questions with regards to the overall design of this feature.

phroggyy:
@bacongobbler do you have a summary of the current concerns? The ones I've seen are

- arbitrary tags: happy to just drop this from the PR (but we'll need some kind of cleanup task to remove it from helm chart save)
- SemVer matching: imo this should be done as a separate improvement. This feature is still experimental which allows us to iteratively improve it rather than have one monster PR
- using local cache: I can look at this as part of the PR.

It'd be good to have a rough idea of what's needed to get it through (or otherwise to close the PR if it's not close enough to what it "should be"), as I feel I personally have been too deep in the code of this to reasonably determine what makes sense anymore.

pmalic:
This SemVer matching on tags is "challenging". When matching, are you always going to retrieve all the tags for each image from the registry? There could be thousands of them for each dependency, you have to using paging if the registry has some limit of returned tags for one request (e.g. Azure CR has 999 as limit). If you aren't going to retrieve them every time, you have to cache them, but how and where? For "classic" repositories you have a local cache that you can update with `repo update`, but with OCI, each image would have to be treated like a repository and have its own cache.

TBBle:
I agree that SemVer matching is a different challenge. We'll need to resolve it (even if that resolution is "you can't do that...") before this feature leaves experimental, but if we limit the supported tags to SemVer now, then we aren't blocking off a route to SemVer Matching via tag listing. If a different approach is found, the restriction on tag format can be relaxed less disruptively than it can be introduced if tag listing _is_ the best way forward.

yurrriq:
any update on plans/timeline here? we would love to see #3 (and `helm install` support). can I help somehow?

bacongobbler:
see https://github.com/helm/community/pull/136

bsuchorowski:
Any news when this is going to be released? We are using an umbrella chart with Helm2 and we would love to switch onto Helm3 as soon as possible.

mattfarina:
An alternative implementation has come in via #8843. Removing this from the milestone as we look at the options.

jdolitsky:
Closing as #8443 was merged

Thanks @phroggyy for all your efforts here, which helped push the conversation forward and resulted in [HIP 6](https://github.com/helm/community/blob/master/hips/hip-0006.md)

phroggyy:
Just to clarify for future readers, that should be #8843, not #8443 🙂  (or so I assume)

PR #7625, opened 2020-02-18
Handle resource apiVersion upgrades

jlegrone:
Related to https://github.com/helm/helm/issues/6850, and an alternative implementation to https://github.com/helm/helm/pull/7575.

### Steps to Reproduce

Create a chart with

```yaml
apiVersion: apps/v1beta2
kind: Deployment
```

Deploy a release of that chart, then switch to apps/v1:

```diff
- apiVersion: apps/v1beta2
+ apiVersion: apps/v1
kind: Deployment
```

Run `upgrade`, and you should get the following error before this patch:

```
rendered manifests contain a new resource that already exists. Unable to continue with update: existing resource conflict: namespace: default, name: version-migrate-test, existing_kind: apps/v1, Kind=Deployment, new_kind: apps/v1, Kind=Deployment
```

PR #7627, opened 2020-02-19
Workaround obsolete API versions

pho-enix:
This is a workaround for changes in API versions like
extensions/v1beta1 => apps/v1

Ignore group and version.

Signed-off-by: Sebastian Poehn <sebastian.poehn@gmail.com>

### DIFF
--- a/pkg/action/upgrade.go
+++ b/pkg/action/upgrade.go
@@ -427,5 +427,5 @@ func recreate(cfg *Configuration, resources kube.ResourceList) error {
 
 func objectKey(r *resource.Info) string {
 	gvk := r.Object.GetObjectKind().GroupVersionKind()
-	return fmt.Sprintf("%s/%s/%s/%s", gvk.GroupVersion().String(), gvk.Kind, r.Namespace, r.Name)
+	return fmt.Sprintf("%s/%s/%s", gvk.Kind, r.Namespace, r.Name)
 }


pho-enix:
#6850 
#7575

pho-enix:
Drawback of this solution is that it will not catch resource conflicts of pre-existing resources with the same Kind but different GroupVersion. This can only happen for CRDs. I would expect that this case is rather rare. Impact might be that applying the change could fail.

bacongobbler:
Can you explain the difference between this PR, #7575, and the one @jlegrone proposed in #7625? It's getting really unclear what is the goal with three PRs trying to tackle the same problem, and what the goals of each PR accomplishes. There's not a whole lot of information here to describe what's the intention behind these PRs.

pho-enix:
Sorry for raising confusion. I wanted to provide different implementation options so we can decide for the best one.

#7627 Ignores `groupVersion` when comparing existing vs. planned resources. Thus a change of the API version will no longer block an upgrade. However `Kind` on it's own is not unique. Example:
```yaml
kind: Deployment
apiVersion: apps/v1
metadata:
  name: new
```

```yaml
kind: Deployment
apiVersion: naughty.inc/v1
metadata:
  name: existing
```
In the Kubernetes maintained resource types there is no such conflicts. But CRDs could do weird things. Thus this implementation would not detect a pre-existing resource that does not belong to the Helm Release, for the case of clashing Kind + Name + Namespace.

**I would favour this solution as it is simple and the described corner-case sounds rather exotic to me.**
---
#7575 Same as #7627 but it partially handles the described corner-case. It will detect the clash, if the pre-existing or the created resource is a CRD.

pkg/action/upgrade.go
```diff
@@ -427,5 +427,5 @@ func recreate(cfg *Configuration, resources kube.ResourceList) error {
 
 func objectKey(r *resource.Info) string {
 	gvk := r.Object.GetObjectKind().GroupVersionKind()
-	return fmt.Sprintf("%s/%s/%s/%s", gvk.GroupVersion().String(), gvk.Kind, r.Namespace, r.Name)
+	return fmt.Sprintf("%s/%s/%s", gvk.Kind, r.Namespace, r.Name)
```
jlegrone: I considered this as well. Unfortunately ignoring group and version would allow collisions with custom resources from different API groups. Perhaps this is something we might be able to tolerate and call out in docs, but I'm not sure.

@bacongobbler @hickeyma any thoughts?

jlegrone: If we could exclude `gvk.GroupVersion().String()` for native k8s resource types but _not_ for custom resources, then I think this might be fine.

bacongobbler: This would be a change in Helm's existing behaviour, right? I don't think we can remove this behaviour if we made the assumption that we would not allow collisions in previous versions of Helm 3.

technosophos: We also do not have any assurance from Upstream Kubernetes that they won't at some point decide to have multiple kinds with the same names that have different group/version.

bacongobbler:
closed in favour of https://github.com/helm/helm/pull/7649.

PR #7635, opened 2020-02-19
Migrate SQL storage driver to Helm 3

emaincourt:
As stated in #5371 the SQL backend driver that was implemented for Tiller is still missing in Helm 3.

This PR migrates it, in respect with the changes that have been introduced in the aforementioned version for the other drivers. So far, the noticeable changes are:
* rename table columns to match the new labels that are now in camel case
* add the `type` column to match the [new standard](https://github.com/emaincourt/helm/blob/fix/migrate-sql-storage-driver-to-helm-3/pkg/storage/driver/secrets.go#L236)

Unit tests are passing and my local experiments are working so far.

@elafarge, @hickeyma comments are welcome :)


pkg/storage/driver/sql.go
```diff
@@ -0,0 +1,339 @@
...
+import (
+	"fmt"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/jmoiron/sqlx"
+	migrate "github.com/rubenv/sql-migrate"
+
+	// Import pq for potgres dialect
```
pho-enix:
```suggestion
	// Import pq for postgres dialect
```

emaincourt: Thanks 🙏

hickeyma (commented):
Thanks @emaincourt for porting the driver to Helm 3.

The PR looks good. I have a question around the release table.

In case you haven't heard, Helm 3 introduced the concept of storing release data in the namespace of the release. Helm 2 used the Tiller namespace.
Therefore in Helm 3, release names are now only unique per namespace and not across the whole cluster. From a usability perspective this means that:
- A release name can be used more than once 
- User rights are based on the user rbac in the cluster i.e. limited access to namespaces etc.

With regard to SQL driver, is there only going to be one release table? If so, how can you segregate on a namespace basis and control access per user?

bacongobbler (changes requested):
I agree with @hickeyma. Looking at the code, it looks like the primary key is still based on release name and version. It should probably be based from the name, version and the namespace (which I don't see a field for). Can you please address this?

emaincourt:
@hickeyma @bacongobbler of course, I totally forgot to integrate this change when migrating the code. Will fix it as soon as possible.

emaincourt:
@hickeyma sorry for being late. I've been thinking a bit about what you said. It is actually pretty interesting.
 
Concerning the fact that releases are being scoped to one specific namespace, the solution looks pretty straightforward and as @bacongobbler stated we only need to add a `namespace` field to the `releases` table and update the unicity constraints.  However your second point about authorisation (which I believe is a very nice opportunity offered with the removal of tiller) can barely be addressed. Indeed one would probably need to handle it with the privileges the identified database users have been granted with. 

I see different options here.

The first and most obvious one would be to keep things as they are now and advertise that using SQL as a store  bypasses the k8s RBAC policies.

The second option I can think of is using one table for each namespace. The issue with this approach is that, as far as I can think, tables would need to be created prior to any usage manually by an authorised pg user, who could then setup the rights on it. It would mean users to be aware of the table schema and run a migration script which does not really make sense to me. It would also allow us to remove the migration from the `Init` method though.

@hickeyma @elafarge do you think of any other alternative that might help us here ?

hickeyma:
Thanks for the feedback @emaincourt . Some feedback below.

> Concerning the fact that releases are being scoped to one specific namespace, the solution looks pretty straightforward and as @bacongobbler stated we only need to add a namespace field to the releases table and update the unicity constraints.

I agree that should suffice when adding namespace to key.

WRT to authorisation, after thinking about this again it might not be an issue. I think it can be handled by Kubernetes when the resources are being created. in the cluster. In other words, if the user does not has RBAC rights to create the said resources in a namespace, then it will be error. @bacongobbler Does this seem reasonable?

hickeyma:
One other thing.  Helm v2 and Helm v3 might both be running concurrently. In this case would they be using the same DB instance? If so, the DB table (`releases`) would cause a conflict?  Is it worth having different tables names between Helm v2 and v3?

emaincourt:
@hickeyma thank you for your feedback. 

Regarding conflicts between Helm v2 and Helm v3, we could definitely use a new table name for the latter one. I was not really expecting people running both versions concurrently. So in our case any existing deployment of Tiller would *exclusively* read/write from/to a specific table and all Helm v3 clients would *exclusively* read/write from/to another table. 

What about migrating releases from v2 to v3 ? Do users need to create a new release and delete the old one managed by Tiller or should we provide them with a way to migrate their releases ?

You're actually right that Kubernetes will handle it properly with RBAC permissions for creation and deletion. What I’m worried about is the user being able to get/list releases and potentially secret values from a namespace he or she is not supposed to have rights on. What do you think about it ? Is it acceptable ?

hickeyma:
>  I was not really expecting people running both versions concurrently.

That might be the case but I think we need to be able to handle versions running concurrently as per [migration doc](https://helm.sh/docs/topics/v2_v3_migration/) because SQL storage backend is supported.

> What about migrating releases from v2 to v3 ? Do users need to create a new release and delete the old one managed by Tiller or should we provide them with a way to migrate their releases ?

I think a util/lib to return the v2 releases could be plugged into the Helm "2to3" plugin. I opened an issue to cover this: https://github.com/helm/helm-2to3/issues/126

> You're actually right that Kubernetes will handle it properly with RBAC permissions for creation and deletion. What I’m worried about is the user being able to get/list releases and potentially secret values from a namespace he or she is not supposed to have rights on. What do you think about it ? Is it acceptable ?

What are the options for providing authorization per user in the DB?

emaincourt:
@hickeyma Thank you for your answer.

> I think a util/lib to return the v2 releases could be plugged into the Helm "2to3" plugin. I opened an issue to cover this: [helm/helm-2to3#126](https://github.com/helm/helm-2to3/issues/126)

I would be glad to take care of it, it sounds great and pretty straightforward to implement.

> What are the options for providing authorization per user in the DB?

Well if we want to handle the authentication on the db side, we could leverage the [RLS](https://www.postgresql.org/docs/9.5/ddl-rowsecurity.html) feature of PostgreSQL. Therefore, admin roles would be able to read/write on/to all the entries of the `releases_v3` table, while other users would require an admin to enable their rights on the table for specific values of the `namespace` column (or any other column of the table) with something like:
```sql
CREATE POLICY name ON releases_v3 FOR ALL TO user USING (namespace = 'default');
```

It would prevent us from the need to create multiple tables, and would also be transparent to the users which would only get the releases belonging to the namespaces their are explicitly allowed to list/update/delete. Otherwise, they will only get an empty list.

I've already implemented and tested it on my machine and it works well.

What do you think ? Does it make sense to you ?

hickeyma:
> Well if we want to handle the authentication on the db side, we could leverage the RLS feature of PostgreSQL. 

Sounds good @emaincourt . Does this require changes/hooks on the Helm side or is it more documentation for the db admin?

emaincourt:
@hickeyma as far as I know it does not require any changes on the Helm side. But it definitely needs to be properly documented. I will take care of it and will then ask you to review this PR once again if you don't mind :)

pkg/storage/driver/sql.go
```diff
@@ -0,0 +1,361 @@
...
+
+func (s *SQL) ensureDBSetup() error {
+	// Populate the database with the relations we need if they don't exist yet
+	migrations := &migrate.MemoryMigrationSource{
+		Migrations: []*migrate.Migration{
+			{
+				Id: "init",
+				Up: []string{
+					`
+						CREATE TABLE releases_v3 (
```
hickeyma: Can we call the table `release.v1` ?
This would correspond to the type used for Secret/Configmap storage: https://github.com/helm/helm/blob/master/pkg/storage/driver/secrets.go#L236

hickeyma (changes requested):
@emaincourt Do you mind doing the following:
- Rebase and fix conflicts
- Fix the failing tests
- Sign the [DCO:](https://helm.sh/blog/helm-dco/)
- Inline comment about table name

pkg/storage/driver/sql.go
emaincourt: @hickeyma postgres uses dot notation to access databases, schemes and table names. Therefore it is not allowed to use them for table/column names. There are workarounds but would you mind using `releases_v1` instead of `releases.v1` ?

emaincourt:
@hickeyma I’ve modified the code, fixed the tests and signed the DCO.

go.mod
```diff
@@ -34,6 +38,7 @@ require (
 	k8s.io/apimachinery v0.17.3
 	k8s.io/cli-runtime v0.17.3
 	k8s.io/client-go v0.17.3
+	k8s.io/helm v2.16.3+incompatible
```
hickeyma: What is the need to reference Helm 2?

pkg/storage/driver/sql.go
```diff
@@ -0,0 +1,463 @@
...
+						CREATE INDEX ON releases_v1 (name, namespace);
+						CREATE INDEX ON releases_v1 (version);
+						CREATE INDEX ON releases_v1 (status);
+						CREATE INDEX ON releases_v1 (owner);
+						CREATE INDEX ON releases_v1 (createdAt);
+						CREATE INDEX ON releases_v1 (modifiedAt);
+						
+						GRANT ALL ON releases_v1 TO PUBLIC;
+
+						ALTER TABLE releases_v1 ENABLE ROW LEVEL SECURITY;
```
hickeyma: Would it be better to create this string first (e.g. `fmt.Sprintf`) before assigning to `Up`  and you could use the `sqlReleaseTableName` definition?

hickeyma (changes requested):
Thanks for the updates @emaincourt. Few comments to address.

pkg/storage/driver/sql.go
hickeyma: That's ok.

hickeyma:
> where do you think I should write down the documentation for the db permissions ?

Hmmm.  I think it should be a new topic in https://helm.sh/docs/.  Have a look in `https://github.com/helm/helm-www/tree/master/content/en/docs/topics` for a pattern.

Would you also be able to add the SQl storage backend as per [v2 docs](https://v2.helm.sh/docs/using_helm/#sql-storage-backend)? https://github.com/helm/helm-www/issues/422 is an open issue for it.

emaincourt:
@hickeyma thank you for your feedback.

I've opened a PR on the `helm-www` repository: https://github.com/helm/helm-www/pull/538. I've created a new topic for roles management. 

Regarding your second point, it looks like the whole `Storage backends` section was removed from the [Using Helm](https://helm.sh/docs/intro/using_helm/) documentation. Am I right ? And, if yes, would you like me to also add documentation for setting up the other type of backends (cm, secrets, memory) ? Otherwise where do you want me to write it down ?

Thank you in advance

pkg/storage/driver/sql.go
emaincourt: You're absolutely right. I've updated the migration to make use of all the constants declared at the top of the file.

go.mod
emaincourt: My mistake, I removed it.

hickeyma:
I opened PR https://github.com/helm/helm-www/pull/542 to add "Storage backend" section to the current docs.  Do you mind reviewing this PR to see if the SQL section is correct?

hickeyma:
@emaincourt Thanks for adding role management doc. 

It looks like the code updates from review comments are not pushed yet?

emaincourt:
@hickeyma you were right, I forgot to push the changes regarding the Helm2 dependency. My bad, it's done now.

I've been reviewing the "Storage backend" section you've added to the docs and it looks perfect on my side. I only added a minor comment. Thanks a lot for taking the time to write this part !

pkg/storage/driver/sql.go
```diff
@@ -0,0 +1,484 @@
...
+							%s INTEGER NOT NULL,
+							%s INTEGER NOT NULL DEFAULT 0,
+							PRIMARY KEY(%s, %s)
+						);
+						CREATE INDEX ON releases_v1 (%s, %s);
+						CREATE INDEX ON releases_v1 (%s);
+						CREATE INDEX ON releases_v1 (%s);
+						CREATE INDEX ON releases_v1 (%s);
+						CREATE INDEX ON releases_v1 (%s);
+						CREATE INDEX ON releases_v1 (%s);
```
hickeyma: Is it possible to substitute the table name instead?

pkg/storage/driver/sql.go
hickeyma: Indices need updating

hickeyma (changes requested):
Thanks for update @emaincourt. Nearly there, just one more small thing. Also, you need to re-sign the DCO.

go.mod
```diff
@@ -18,11 +19,14 @@ require (
 	github.com/gobwas/glob v0.2.3
 	github.com/gofrs/flock v0.7.1
 	github.com/gosuri/uitable v0.0.4
+	github.com/jmoiron/sqlx v1.2.0
+	github.com/lib/pq v1.3.0
```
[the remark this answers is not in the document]
emaincourt: @hickeyma I'm definitely not an expert in licenses but as far as I can see the only limitation here is:
```
The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
```
Which we'll be in respect with as the license is included in the dependency that will be pulled. Moreover this library is basically the core psql driver, I don't think there should be any constraints of use.

What do you think ?

emaincourt:
@hickeyma done. I let you double check the license. As I said, I won't be very helpful there...

go.mod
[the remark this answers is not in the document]
mattfarina: This is the [MIT license](https://opensource.org/licenses/MIT). My guess is that something about the first line is causing the automated parsing to not realize it. The [CNCF whitelist policy](https://github.com/cncf/foundation/blob/master/whitelist-policy.md) has the list of licenses and includes the MIT license. This should be ok to use from a license standpoint.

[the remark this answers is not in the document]
hickeyma: @mattfarina How can you tell it is part of MIT license? I can only see this: https://github.com/lib/pq/blob/master/LICENSE.md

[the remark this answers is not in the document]
mattfarina: @hickeyma That license is the MIT license. Compare it to https://opensource.org/licenses/MIT. There is something odd about the file so that GitHub doesn't auto-recognize it.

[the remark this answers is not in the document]
hickeyma: I see what you mean now. Thanks for the clarification.

hickeyma (approved):
LGTM, thanks @emaincourt for your patience and driving on the changes to port this over to Helm 3.

hickeyma:
@emaincourt I will need a +1 from another @helm/helm-core-maintainers before merging.

pkg/storage/driver/sql.go
```diff
@@ -0,0 +1,490 @@
...
+	Name       string `db:"name"`
+	Namespace  string `db:"namespace"`
+	Version    int    `db:"version"`
+	Status     string `db:"status"`
+	Owner      string `db:"owner"`
+	CreatedAt  int    `db:"createdAt"`
+	ModifiedAt int    `db:"modifiedAt"`
+}
+
+// NewSQL initializes a new memory driver.
```
technosophos: nit: I think this comment needs updating.

pkg/storage/driver/sql.go
```diff
@@ -0,0 +1,490 @@
...
+		if filter(release) {
+			releases = append(releases, release)
+		}
+	}
+
+	return releases, nil
+}
+
+// Query returns the set of releases that match the provided set of labels.
+func (s *SQL) Query(labels map[string]string) ([]*rspb.Release, error) {
```
technosophos: Here I am also concerned about string-building queries where the values are not sanitized. At a glance, it seems like there are a number of places where someone could intentionally pass in malformed data and perform a SQL injection attack.

technosophos (changes requested):
I think I need someone to sanity check my complaints about SQL injection attacks here. I'll ask @mattfarina to double check me on this.

*UPDATE:* I did talk to @mattfarina and I did do some testing. I have confirmed one place where a malformed namespace argument could result in a SQL injection attack. However, the fix is pretty easy: Just switch to a prepared statement as used elsewhere in this PR.

pkg/storage/driver/sql.go
technosophos: Okay, I think that I have ruled out SQL injections in this function.

pkg/storage/driver/sql.go
```diff
@@ -0,0 +1,490 @@
...
+	query := fmt.Sprintf(
+		"SELECT %s FROM %s WHERE %s = '%s'",
+		sqlReleaseTableBodyColumn,
+		sqlReleaseTableName,
+		sqlReleaseTableOwnerColumn,
+		sqlReleaseDefaultOwner,
+	)
+
+	// If a namespace was specified, we only list releases from that namespace
+	if s.namespace != "" {
```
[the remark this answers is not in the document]
technosophos: Update: I believe I have _CONFIRMED_ that this is unsafe code. It should use a prepared statement with a placeholder for `s.namespace`.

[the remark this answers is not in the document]
hickeyma: Nice catch @technosophos. Thanks for finding this.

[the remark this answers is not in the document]
emaincourt: @technosophos Thanks a lot for catching this one. This is now a prepared statement.

emaincourt:
@hickeyma thank you once again for reviewing the code.

@technosophos @mattfarina Thank you for taking the time to review this PR, especially the SQL injection part. I addressed all your comments and we're now making use of prepared statements everywhere. Moreover I eventually thought it my be of great value to use an SQL builder here for the following reasons:

* First, make sure that we're using prepared queries everywhere.
* Make the driver more extendable. It is now far easier to add new dialects only with changing the `StatementBuilder` of the `SQL` driver.
* Finally, the readability of the code gets improved a lot. The code needs to be not too messy to be properly maintainable.

In respect with your concerns about SQL injections, I could also take care of adding security tests in the CI to ensure we're safe (which would probably require some more time).

technosophos (approved):
LGTM. I did _not_ expect to see Squirrel in here! Hah! But in re-reviewing, this looks good, and looks like clean code.

technosophos:
Update: As soon as this is rebased, I will merge it. I think we can get it included in time for 3.2!

pkg/action/action.go
```diff
@@ -255,6 +257,17 @@ func (c *Configuration) Init(getter genericclioptions.RESTClientGetter, namespac
 		}
 		d.SetNamespace(namespace)
 		store = storage.Init(d)
+	case "sql":
```
technosophos, line 260: I know this is the way it worked on Helm 2... but maybe we could just set this (`HELM_DRIVER`) to "postgres" and then get rid of `HELM_DRIVER_SQL_DIALECT`.

bacongobbler (approved):
Original concern about the release name/version primary key has since been addressed. have not looked at the rest of the code, but I'll let other maintainers weigh in here. LGTM

technosophos:
Alright, if this gets rebased by tomorrow, I'll add it to 3.2. Otherwise I will push this onto the 3.3 tag.

hickeyma (approved):
LGTM, thanks @emaincourt. Can you do a rebase so the PR can be merged?

emaincourt:
@technosophos @bacongobbler @hickeyma Thanks a lot for taking the time to review this PR. I rebased it against master.

As stated by @technosophos, I also removed references to `HELM_DRIVER_SQL_DIALECT` to use `postgres` all the time. Hence we need to update the [docs](https://github.com/helm/helm-www/pull/558). @hickeyma do you think you can take care of ? Otherwise I would be more than glad to help you do so.

hickeyma:
Thanks @emaincourt. The doc PR is now updated.

PR #7640, opened 2020-02-20
fix(helm): stdin values for helm upgrade --install

### LINKED ISSUE #7002: Support reading values from STDIN for `upgrade` command
#2709 introduced support for using STDIN with the "-" character for `install`. However, I don't think this works with the `upgrade` command. 

My example usecase is decrypting a sops file and using the output in the `helm upgrade` command:

```sh
# I can do this and values in secrets.sops.yaml are available to the release:
sops -d secrets.sops.yaml | helm install -f - -f values.yaml my-release release/my-chart 

# but this doesn't work:
sops -d secrets.sops.yaml | helm upgrade -f - -f values.yaml my-release release/my-chart 
```

yinzara:
Fixes #7002

The cause of this issue is the helm upgrade cmd parses the command line options (which reads from std in) before determining if it needs to call the helm install command (because of the missing release).  Then when helm install is called, it rereads the command line args and this time reads nothing from stdin.

There were never any test cases that tested using the "--values -"/"-f -" syntax to pass values via stdin. Because of this there was no test harness or way to achieve this. I updated helm_test.go to support passing stdin as a *os.File.

The actual fix was in cmd/helm/upgrade.go and involved changing the order of operations of the parsing of arguments and calling install if necessary.

I added test cases to actually test "helm upgrade --install" (there were none) in combination with using the stdin values and another for just "helm upgrade" with stdin values.

Signed-off-by: Matt Morrissette <yinzara@gmail.com>

### DIFF
--- a/cmd/helm/helm_test.go
+++ b/cmd/helm/helm_test.go
@@ -97,10 +97,15 @@ func storageFixture() *storage.Storage {
 }
 
 func executeActionCommandC(store *storage.Storage, cmd string) (*cobra.Command, string, error) {
+	return executeActionCommandStdinC(store, nil, cmd)
+}
+
+func executeActionCommandStdinC(store *storage.Storage, in *os.File, cmd string) (*cobra.Command, string, error) {
 	args, err := shellwords.Parse(cmd)
 	if err != nil {
 		return nil, "", err
 	}
+
 	buf := new(bytes.Buffer)
 
 	actionConfig := &action.Configuration{
@@ -111,15 +116,26 @@ func executeActionCommandC(store *storage.Storage, cmd string) (*cobra.Command, 
 	}
 
 	root := newRootCmd(actionConfig, buf, args)
-	root.SetOutput(buf)
+	root.SetOut(buf)
+	root.SetErr(buf)
 	root.SetArgs(args)
+
+	oldStdin := os.Stdin
+	if in != nil {
+		root.SetIn(in)
+		os.Stdin = in
+	}
 
 	if mem, ok := store.Driver.(*driver.Memory); ok {
 		mem.SetNamespace(settings.Namespace())
 	}
 	c, err := root.ExecuteC()
 
-	return c, buf.String(), err
+	result := buf.String()
+
+	os.Stdin = oldStdin
+
+	return c, result, err
 }
 
 // cmdTestCase describes a test case that works with releases.
--- a/cmd/helm/upgrade.go
+++ b/cmd/helm/upgrade.go
@@ -75,21 +75,8 @@ func newUpgradeCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			client.Namespace = settings.Namespace()
 
-			if client.Version == "" && client.Devel {
-				debug("setting version to >0.0.0-0")
-				client.Version = ">0.0.0-0"
-			}
-
-			vals, err := valueOpts.MergeValues(getter.All(settings))
-			if err != nil {
-				return err
-			}
-
-			chartPath, err := client.ChartPathOptions.LocateChart(args[1], settings)
-			if err != nil {
-				return err
-			}
-
+			// Fixes #7002 - Support reading values from STDIN for `upgrade` command
+			// Must load values AFTER determining if we have to call install so that values loaded from stdin are are not read twice
 			if client.Install {
 				// If a release does not exist, install it. If another error occurs during
 				// the check, ignore the error and continue with the upgrade.
@@ -119,6 +106,21 @@ func newUpgradeCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 					}
 					return outfmt.Write(out, &statusPrinter{rel, settings.Debug})
 				}
+			}
+
+			if client.Version == "" && client.Devel {
+				debug("setting version to >0.0.0-0")
+				client.Version = ">0.0.0-0"
+			}
+
+			chartPath, err := client.ChartPathOptions.LocateChart(args[1], settings)
+			if err != nil {
+				return err
+			}
+
+			vals, err := valueOpts.MergeValues(getter.All(settings))
+			if err != nil {
+				return err
 			}
 
 			// Check chart dependencies to make sure all are present in /charts
--- a/pkg/action/rollback.go
+++ b/pkg/action/rollback.go
@@ -211,7 +211,7 @@ func (r *Rollback) performRollback(currentRelease, targetRelease *release.Releas
 	}
 
 	deployed, err := r.cfg.Releases.DeployedAll(currentRelease.Name)
-	if err != nil {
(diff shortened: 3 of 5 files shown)


cmd/helm/helm_test.go
```diff
@@ -97,6 +97,8 @@ func storageFixture() *storage.Storage {
 	return storage.Init(driver.NewMemory())
 }
 
+// As of Git commit 598dc504d634332f955f5e984d1d53696ba07abc (Web Feb 19, 2020), go-shellwords seems to have been fixed
```
bacongobbler: can you remove this comment as well as the function that was commented out?

cmd/helm/upgrade.go
```diff
@@ -74,27 +74,14 @@ func newUpgradeCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
...
-
+			// Fixes #7002 - Support reading values from STDIN for `upgrade` command
+			// Must load values AFTER determining if we have to call install so that values loaded from stdin are are not read twice
 			if client.Install {
 				// If a release does not exist, install it. If another error occurs during
 				// the check, ignore the error and continue with the upgrade.
 				histClient := action.NewHistory(cfg)
 				histClient.Max = 1
-				if _, err := histClient.Run(args[0]); err == driver.ErrReleaseNotFound {
+				if release, err := histClient.Run(args[0]); err == driver.ErrReleaseNotFound || release == nil {
```
bacongobbler: why is the check for `release == nil` necessary?

pkg/cli/values/options.go
```diff
@@ -38,14 +38,12 @@ type Options struct {
 
 // MergeValues merges values from files specified via -f/--values and directly
 // via --set, --set-string, or --set-file, marshaling them to YAML
-func (opts *Options) MergeValues(p getter.Providers) (map[string]interface{}, error) {
+func (opts *Options) MergeValues(p getter.Providers, in io.Reader) (map[string]interface{}, error) {
```
bacongobbler: why do we need to pass in a reader? `readFile` should only be able to read from `os.Stdin` if the filepath is set to `-`.

bacongobbler (changes requested):

bacongobbler:
Thanks for the PR. Can you break this PR down to the bare minimum required to fix #7002? It looks like there's a lot of refactoring done in this PR, most of which isn't relevant to fixing the issue. While we appreciate the effort to refactor the codebase, can you split that out into a separate PR? That way we can discuss the refactors separate from the actual fix. Thanks!

cmd/helm/upgrade.go
yinzara: The "memory" driver used in test cases seems not not behave identically to the regular driver.  It's returning nil when there is no history instead of throwing the driver.ErrReleaseNotFound error.  I thought it was lower risk to just check for nil here then change the behavior of it.  Thoughts?

yinzara:
Apologies for the horribly and unnecessarily huge PR. Go is relatively new to me and I thought to mock os.Stdin I had to do a lot more work.  I have now fixed the PR to only include changes related to my fix.  It's much simpler and smaller now.

cmd/helm/helm_test.go
yinzara: done

cmd/helm/helm_test.go
```diff
@@ -140,15 +120,26 @@ func executeActionCommandC(store *storage.Storage, cmd string) (*cobra.Command,
 	}
 
 	root := newRootCmd(actionConfig, buf, args)
-	root.SetOutput(buf)
+	root.SetOut(buf)
```
yinzara: root.SetOutput was deprecated in favor of calling SetOut and SetErr

cmd/helm/upgrade.go
yinzara: I actually decided to go and fix the memory driver so it behaved like the other drivers and ended up discovering a defect in the rollback functionality.  When a rollback occurred and it rolled back to a release in which there were no prior deployed releases, it would cause a "... has no deployed releases" error.  I fixed that defect with my latest commit.

yinzara:
Rebased against master into a single commit.

@bacongobbler 

Can we get this merged?

yinzara:
@bacongobbler Any chance you'll have time this week? This is keeping my org from upgrading to helm 3 (yes I know I can run a custom version but our build scripts are based around using released versions).

yinzara:
:) Still waiting. I know the world is going crazy but we'd really love to see this merged.

bacongobbler:
Thank you for the ping.

At this time, I'm focused on other work. I do not have enough time in my work schedule to review your code in the near future. That being said, there are plenty of other Helm maintainers who may have the time to review your work. I'd reach out and see if someone else is available to review this.

Sorry for the inconvenience.

yinzara:
Thanks man! I appreciate it anyway.

How about @mattfarina ?

yinzara:
There have now been 3 pull requests opened that duplicate this one.  This is obviously a big bug that multiple people have experienced. Can we please get this merged? Someone?

yinzara:
@technosophos Can we get this merged?

lev-kuznetsov:
fwiw I built a docker image from this fork at levk/helm feel free to use it until hopefully this will get merged

yinzara:
@technosophos I understand my other PR about "--replace" is controversial and adding a new feature takes time, but this is a very simple bug fix that multiple duplicate PRs have been opened about.  There are automated tests and the code change is minimal. 

Can we please get this merged?

pkg/action/rollback.go
```diff
@@ -211,7 +211,7 @@ func (r *Rollback) performRollback(currentRelease, targetRelease *release.Releas
 	}
 
 	deployed, err := r.cfg.Releases.DeployedAll(currentRelease.Name)
-	if err != nil {
+	if err != nil && !strings.Contains(err.Error(), "has no deployed releases") {
```
technosophos, line 214: What is this change for? It seems unrelated to the rest of this PR.

technosophos:
I left one comment on something I don't understand, but otherwise I am good with it. I'm adding to the 3.2 milestone. I am optimistic that we can get this merged for the 3.2 release.

pkg/action/rollback.go
yinzara: It was actually related to a bug I found in the memory driver when I added the two test cases (see my change to memory.go). It was behaving differently and returning nil instead of throwing the "no deployed releases". When I corrected that I discovered another defect in the rollback where it was attempting to set previous releases as superceded but when there were no previous releases it was throwing an error instead of ignoring.  If you review the comments above, I originally had added a nil check in the upgrade.go but decided to fix the defect in the memory driver which made this other issue visible.

technosophos: Alright. That makes sense.

technosophos (approved):
On re-reviewing all of the recent changes, I am marking this LGTM and merging for 3.2.

Thanks!

mattfarina (approved):
lgtm

Issue #2730, opened 2017-07-27
Proposal: helm commandeer


thomastaylor312:
## Problem
Many users want to be able to "adopt" currently running pods, services, etc. into a new or existing chart. One example of this is self-hosted Kubernetes control plane components (apiserver, DNS, etc.) after they have been bootstrapped by something like bootkube.

## Proposed solution
I propose we add a new `commandeer` (to stick with a naval theme, we can go with `adopt` if people find it too confusing) subcommand to Helm. The reason behind this is we don't want to add additional complexity and flags to an already large flag set on `install`, `update`, and `rollback`. The command would take a defined chart (e.g. `helm commandeer <chart_name>`) with matching manifests and use these to "commandeer" the objects into a chart manifest. Future iterations of the feature could add the ability to pass specific names of resources you want to commandeer.

### Behavior
The passed chart would be rendered like a normal `install` or `upgrade` before moving on to a "capture" step. The "capture" step would use the object `GroupVersionKind`, namespace, and name to find a match in the already existing objects. For safety sake, if any of the rendered manifests do not exactly match a resource, the "capture" step will fail. There will also be an `--override` flag and an `--ignore-match` flag to help configure this behavior. The `--override` flag will make it so as long as the Kind and name match, it will accept the match. The `--ignore-match` flag will drop any non-matching manifest from the final chart.

The last step will be a merge patch of the new manifest with the existing one. Once these are merged and sent to the API, the new chart will be saved into storage. So the tl;dr is below

1. Render chart
2. Capture resources using `GroupVersionKind`, namespace, and name
3. Merge manifests
4. Save new chart

#### The `--into` flag
We also want to have the ability to adopt resources into a currently installed chart (e.g. bringing resources into an umbrella chart). This will be done with the `--into` flag, which will take a release name as an argument. All of the same steps will be followed to adopt resources with the exception that existing resource will remain the same. This is a purely additive operation

## Open questions
- Should the `--override` flag force resources in the chart that don't have a match to be created anyway?
- Is `--ignore-match` something people will actually use?
- Do we need something like a `--replace` flag where the new manifest completely replaces the existing one?
- Is this too complex for a first pass and should we just stick with specifying resource names to commandeer?

## Other notes
I will update this proposal with suggestions from the comments so as to not have multiple proposal versions clogging up the issue

nebril:
[...]

Issue #2388, opened 2017-05-04
Resource ownership between charts


helgi:
I have a scenario where 2 charts installed via helm both think they own identically named resource (f.ex. secret), which generally isn't a problem until you want to `helm delete` one of them.

Lets take a simple example: Chart `dingo` has a `Secret` named `db-access` and then later a new cool chart named `baby` is created, which also happens to have a `Secret` named `db-access`. Both are in the same `namespace`.

`helm status` on both of the releases tells me they have `db-access` however when I run `helm delete --purge dingo` I do not see the `db-access` after running `helm status baby`, it is as if the `dingo` ate the `baby` resource... so to speak.

What is the expected behaviour and potential fixes?

Few things come to mind:
* Only remove resources that are not in other releases
* Installation fails if there are identically named resources (of the same resource type) in a different release
* A flag (or flags) to support either or both of the above scenarios, or the inverse

helgi:
```
helm version
Client: &version.Version{SemVer:"v2.4.1", GitCommit:"46d9ea82e2c925186e1fc620a8320ce1314cbb02", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.4.1", GitCommit:"46d9ea82e2c925186e1fc620a8320ce1314cbb02", GitTreeState:"clean"}
```

krancour:
This seems to me like a glaring deficiency in the charts themselves-- both of them. Ideally, all resources in a chart are prefixed with the chart name _and_ release name specifically to avoid these sort of naming conflicts in the event that two releases share a namespace.

krancour:
In fact, when you `helm new <name>`, the scaffolding for your new chart even includes a template partial that's intended to be used... and re-used... and re-used ad-nauseum throughout the chart to make prefixing resource names with chart name and release name quasi-easy.

jchauncey:
The problem with that thought though is that there is no real way to enforce that model against the chart. So if we allow developers to create charts that are used to deploy their applications there are bound to be naming collisions at some point. I think having a better linting system to enforce rules would be a good first step.

see #1495

helgi:
@krancour when you use words like *ideally* and *intended* and that using chart name and release is a recommendation at the best of times, not a requirement or enforced, then this is not a deficiency in the chart :)

Now, my question was around expected behaviour - Is the answer here that unless I use the recommended partials and naming conventions then Helm will not able to understand it has more than 1 package that owns the same resource within the same namespace?

[...]

Issue #7219, opened 2019-12-12
no matches for kind "Deployment" in version "apps/v1beta1"


curtisr7:
Output of `helm version`:
```
$ helm version
Client: &version.Version{SemVer:"v2.12.1", GitCommit:"02a47c7249b1fc6d8fd3b94e6b4babf9d818144e", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.15.2", GitCommit:"8dce272473e5f2a7bf58ce79bb5c3691db54c96b", GitTreeState:"dirty"}
```
Output of `kubectl version`:
```
$ ./kubectl version
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.1", GitCommit:"d647ddbd755faf07169599a625faf302ffc34458", GitTreeState:"clean", BuildDate:"2019-10-02T17:01:15Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-15T19:09:08Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}
```
Cloud Provider/Platform (AKS, GKE, Minikube etc.): 
minikube

--------------

Over the past few months there have been numerous issues opened for this very issue[1] and in the end they were mostly closed as "working as designed" / "this is a k8s problem". When looking at the k8s issues linked, k8s also seems to be "working as designed". 

[1]https://github.com/helm/helm/issues/6583
https://github.com/helm/helm/issues/6969

--------------

I've been digging through the helm code the past couple days to better understand the problem / try to hack a solution together. My understanding of the problem is (simplified):

1. Install k8s 1.15
2. helm install a chart w/ deployment @ `apps/v1beta1`
* helm stores these templates in a config map for later operations
3. upgrade k8s to 1.16
* k8s automagically "upgrades" the deployment to `apps/v1`
4. Update chart to use `apps/v1`
5. helm upgrade

The `helm upgrade` fails as it attempts to create a diff patch from steps 2 to 4. The problem is that helm doesn't know that it's source is no longer valid for the current version of k8s.

[...]

Issue #6561, opened 2019-10-02
helm3: How to adopt existing resources


juborto:
Hi,
Helm raises an error if we try to install with a chart some resources which already exist in our cluster.
According to @bacongobbler in https://github.com/helm/helm/issues/6031, _"Helm does not "adopt" existing resources; if a resource exists before it was introduced in the chart, it's considered an error condition, hence the error message"_.
Some post suggested a workaround with Helm version 2. During the first installation, the release was in a state FAILED, but a second installation of the same chart made the release in 
a DEPLOYED state.

With Helm3, it does not work anymore:
```
$ helm3 install test ./helm-test --debug
install.go:158: [debug] Original chart version: ""
install.go:175: [debug] CHART PATH: [...]
client.go:86: [debug] creating 3 resource(s)
Error: services "service" already exists
helm.go:81: [debug] services "service" already exists
```

```
$ helm3 install test ./helm-test --replace --debug 
install.go:158: [debug] Original chart version: ""
install.go:175: [debug] CHART PATH: [...]
client.go:86: [debug] creating 3 resource(s)
Error: services "service" already exists
helm.go:81: [debug] services "service" already exists
```

```
$ helm3 ls
NAME	NAMESPACE	REVISION	UPDATED                                	STATUS	CHART          
test	default  	1       	2019-10-02 17:29:13.735401532 -0400 EDT	failed	helm-test-0.1.0
test	default  	2       	2019-10-02 17:31:05.690696039 -0400 EDT	failed	helm-test-0.1.0
```

So this behavior is probably intended, but in this case, how can I merge the k8s resources of an active cluster to a Helm chart? It is suggested to delete the objects and recreate them with Helm, but it is not a realistic option with a running cluster. What is your advice to "adopt" running resources?

Thanks for your support,
JB

```
$ helm3 version
version.BuildInfo{Version:"v3.0.0-beta.3", GitCommit:"5cb923eecbe80d1ad76399aee234717c11931d9a", GitTreeState:"clean", GoVersion:"go1.12.9"}
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"12+", GitVersion:"v1.12.9-gke.7" [...]}
Server Version: version.Info{Major:"1", Minor:"12+", GitVersion:"v1.12.8-gke.10" [...]} 
```

Cloud Provider/Platform (AKS, GKE, Minikube etc.): GKE

juborto:
Hi,
[...]

PR #7649, opened 2020-02-21
Adopt resources into release with correct instance and managed-by labels

### LINKED ISSUE #2947: `helm upgrade --install` doesn't actually perform install if there are duplicates
If you have the same resources (_ie secrets_) in a chart, _(or more likely, multiple subcharts create the same secret)_ `helm upgrade --install` fails with 

`Error: release releaseName failed: secrets "secret-name" already exists`

-----
**To reproduce:** 
* `helm create test`
* add secrets file to `test/templates`
* make a copy of the secrets file in the same path
* `helm upgrade --install releaseName test/`

---- 
**Work-around:**
* run the same cmd again: `helm upgrade --install releaseName test/`

It seems that after the namespace exists, `upgrade` allows for the duplicate resources. It switches from `creating 3 resource(s)` to `checking 3 resources for changes`.

This workaround defeats the whole purpose of having an `--install` option to upgrade. But it's worth noting that this error also appears when running `helm install test/  --name releaseName`

----
Logs in the tiller:
```
[tiller] 2017/09/18 15:28:13 getting history for release t2
[storage] 2017/09/18 15:28:13 getting release history for "t2"
[tiller] 2017/09/18 15:28:13 preparing install for t2
[storage] 2017/09/18 15:28:13 getting release history for "t2"
[tiller] 2017/09/18 15:28:13 rendering test2 chart using values
2017/09/18 15:28:13 info: manifest "test2/templates/ingress.yaml" is empty. Skipping.
[tiller] 2017/09/18 15:28:13 performing install for t2
[tiller] 2017/09/18 15:28:13 executing 0 pre-install hooks for t2
[tiller] 2017/09/18 15:28:13 hooks complete for pre-install t2
[storage] 2017/09/18 15:28:13 getting release history for "t2"
[storage] 2017/09/18 15:28:13 creating release "t2.v1"
[kube] 2017/09/18 15:28:13 building resources from manifest
[kube] 2017/09/18 15:28:13 creating 3 resource(s)
[tiller] 2017/09/18 15:28:13 warning: Release "t2" failed: secrets "secret-name" already exists
[storage] 2017/09/18 15:28:13 updating release "t2.v1"
[tiller] 2017/09/18 15:28:13 failed install perform step: release t2 failed: secrets "secret-name" already exists
```
and if you run the same command again (ie: workaround):
```
[tiller] 2017/09/18 15:23:27 getting history for release t2
[storage] 2017/09/18 15:23:27 getting release history for "t2"
[tiller] 2017/09/18 15:23:28 preparing update for t2
[storage] 2017/09/18 15:23:28 getting last revision of "t2"
[storage] 2017/09/18 15:23:28 getting release history for "t2"
[tiller] 2017/09/18 15:23:28 rendering test2 chart using values
2017/09/18 15:23:28 info: manifest "test2/templates/ingress.yaml" is empty. Skipping.
[tiller] 2017/09/18 15:23:28 creating updated release for t2
[storage] 2017/09/18 15:23:28 creating release "t2.v2"
[tiller] 2017/09/18 15:23:28 performing update for t2
[tiller] 2017/09/18 15:23:28 executing 0 pre-upgrade hooks for t2
[tiller] 2017/09/18 15:23:28 hooks complete for pre-upgrade t2
[kube] 2017/09/18 15:23:28 building resources from updated manifest
[kube] 2017/09/18 15:23:28 checking 3 resources for changes
[kube] 2017/09/18 15:23:28 Looks like there are no changes for Secret "secret-name"
[kube] 2017/09/18 15:23:28 Looks like there are no changes for Secret "secret-name"
[kube] 2017/09/18 15:23:28 Looks like there are no changes for Secret "secret-name"
[tiller] 2017/09/18 15:23:28 executing 0 post-upgrade hooks for t2
[tiller] 2017/09/18 15:23:28 hooks complete for post-upgrade t2
[storage] 2017/09/18 15:23:28 updating release "t2.v1"
[tiller] 2017/09/18 15:23:28 updating status for updated release for t2
[storage] 2017/09/18 15:23:28 updating release "t2.v2"
[storage] 2017/09/18 15:23:29 getting last revision of "t2"
[storage] 2017/09/18 15:23:29 getting release history for "t2"
[kube] 2017/09/18 15:23:29 Doing get for Secret: "secret-name"
[kube] 2017/09/18 15:23:29 Doing get for Secret: "secret-name"
[kube] 2017/09/18 15:23:29 Doing get for Secret: "secret-name"
```

### LINKED ISSUE #4824: Ignore existing resource on install if resource-policy is keep
I wanted to know if there is possibility for Helm install command to ignore existing resource (if already created) if the **helm/resource-policy** is **keep**.

**Scenario:**
1. I created release with Secret containing random generated password. This Secret is marked as resource-policy: keep
2. I deleted release
3. I do helm install with same release name
4. Install ignores existing secret so that all other services are connected to Secret with old password.

The same thing is needed for **PVC**.

### LINKED ISSUE #6850: Unable to perform helm upgrade due to resource conflict

### LINKED ISSUE #7418: `Error: rendered manifests contain a resource that already exists` but nothing shows up on `helm list --all`
Output of `helm version`:

Output of `kubectl version`:
Server Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.1", GitCommit:"eec55b9ba98609a46fee712359c7b5b365bdd920", GitTreeState:"clean", BuildDate:"2018-12-13T10:31:33Z", GoVersion:"go1.11.2", Compiler:"gc", Platform:"linux/amd64"}

Minikube

I'm trying to helm install but helm comes back saying with following error
`Error: rendered manifests contain a resource that already exists. Unable to continue with install: existing resource conflict: kind: CustomResourceDefinition, namespace: , name: scheduledsparkapplications.sparkoperator.k8s.io`

but when I do `helm list --all`, nothing comes up and therefore I unable to remove any resource to proceed with installation.

Any help appreciated! Thanks!

See the attached screenshot:
Screen Shot 2020-01-16 at 3 36 00 PM

jlegrone:
Alternative implementation to https://github.com/helm/helm/pull/7627, https://github.com/helm/helm/pull/7625, and https://github.com/helm/helm/pull/7575.

Help with validation would be appreciated, but in theory this closes https://github.com/helm/helm/issues/6850, closes https://github.com/helm/helm/issues/4824, closes https://github.com/helm/helm/issues/2947, and closes https://github.com/helm/helm/issues/7418. This implementation would also help make https://github.com/helm/helm/issues/2730 very approachable.

### ~Option A: Standard Labels~ https://github.com/helm/helm/pull/7649/commits/a29365b3c663f1c182ea78461825ae28b8573915

Helm will assume that it has the right to "adopt" any existing Kubernetes resource containing the following labels:

* `app.kubernetes.io/managed-by: Helm`
* `app.kubernetes.io/instance: <RELEASE_NAME>`

The major downside is that in order to get this functionality, setting these labels is required. Despite this having been a [documented best practice](https://helm.sh/docs/topics/chart_best_practices/labels/#standard-labels) since well before Helm 3, that might be hard to accomplish for chart authors who have historically been using non-standard values for these labels. A workaround for this would be to enable the same behavior via custom annotations.

### Option B: New Helm Annotations https://github.com/helm/helm/pull/7649/commits/271cb2268bfd94012639f4866e8d52189e12a97b

Helm will assume that it has the right to "adopt" any existing Kubernetes resource containing the following annotations:

* `meta.helm.sh/release-name: <RELEASE_NAME>`
* `meta.helm.sh/release-namespace: <RELEASE_NAMESPACE>`

For consistency and to make things easier for chart authors, we will automatically inject or override existing values for these annotations. I think this ends up being the simplest solution, and probably something we can safely expose without a feature flag.

## To Do

1. Validate that 3-way diffs work correctly. See PR comment below.
1. Decide if this should be opt-in/behind an experimental feature flag
1. Error out if annotations list a different release as owner of resource (fix for https://github.com/helm/helm/issues/7699 / #2388)
    - This will be enabled in a followup PR, but is implemented in https://github.com/helm/helm/pull/7649/commits/8d1de39fe3234c8925dac6d3efca6aba687ebf87
1. ~Add tests~

## Release Note

Helm will no longer error when attempting to create a resource that already exists in the target cluster if the existing resource has the correct `meta.helm.sh/release-name` and `meta.helm.sh/release-namespace` annotations, and matches the label selector `app.kubernetes.io/managed-by=Helm`. This facilitates zero-downtime migrations to Helm 3 for managing existing deployments, and allows Helm to "adopt" existing resources that it previously created.

In order to allow an existing resource to be adopted by Helm, add release metadata and the managed-by label:

```
KIND=deployment
NAME=my-app-staging
RELEASE=staging
NAMESPACE=default
kubectl annotate $KIND $NAME meta.helm.sh/release-name=$RELEASE
kubectl annotate $KIND $NAME meta.helm.sh/release-namespace=$NAMESPACE
kubectl label $KIND $NAME app.kubernetes.io/managed-by=Helm
```

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -38,6 +38,7 @@ import (
 	"helm.sh/helm/v3/pkg/downloader"
 	"helm.sh/helm/v3/pkg/engine"
 	"helm.sh/helm/v3/pkg/getter"
+	"helm.sh/helm/v3/pkg/kube"
 	kubefake "helm.sh/helm/v3/pkg/kube/fake"
 	"helm.sh/helm/v3/pkg/postrender"
 	"helm.sh/helm/v3/pkg/release"
@@ -242,6 +243,7 @@ func (i *Install) Run(chrt *chart.Chart, vals map[string]interface{}) (*release.
 	// Mark this release as in-progress
 	rel.SetStatus(release.StatusPendingInstall, "Initial install underway")
 
+	var adoptedResources kube.ResourceList
 	resources, err := i.cfg.KubeClient.Build(bytes.NewBufferString(rel.Manifest), !i.DisableOpenAPIValidation)
 	if err != nil {
 		return nil, errors.Wrap(err, "unable to build kubernetes objects from release manifest")
@@ -254,9 +256,11 @@ func (i *Install) Run(chrt *chart.Chart, vals map[string]interface{}) (*release.
 	// deleting the release because the manifest will be pointing at that
 	// resource
 	if !i.ClientOnly && !isUpgrade {
-		if err := existingResourceConflict(resources); err != nil {
+		toBeUpdated, err := existingResourceConflict(resources, rel.Name)
+		if err != nil {
 			return nil, errors.Wrap(err, "rendered manifests contain a resource that already exists. Unable to continue with install")
 		}
+		adoptedResources = toBeUpdated
 	}
 
 	// Bail out here if it is a dry run
@@ -291,7 +295,7 @@ func (i *Install) Run(chrt *chart.Chart, vals map[string]interface{}) (*release.
 	// At this point, we can do the install. Note that before we were detecting whether to
 	// do an update, but it's not clear whether we WANT to do an update if the re-use is set
 	// to true, since that is basically an upgrade operation.
-	if _, err := i.cfg.KubeClient.Create(resources); err != nil {
+	if _, err := i.cfg.KubeClient.Update(adoptedResources, resources, false); err != nil {
 		return i.failRelease(rel, err)
 	}
 
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -249,7 +249,8 @@ func (i *Install) Run(chrt *chart.Chart, vals map[string]interface{}) (*release.
 		return nil, errors.Wrap(err, "unable to build kubernetes objects from release manifest")
 	}
 
-	err = resources.Visit(setMetadataVisitor(rel.Name, rel.Namespace))
+	// It is safe to use "force" here because these are resources currently rendered by the chart.
+	err = resources.Visit(setMetadataVisitor(rel.Name, rel.Namespace, true))
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -249,6 +249,11 @@ func (i *Install) Run(chrt *chart.Chart, vals map[string]interface{}) (*release.
 		return nil, errors.Wrap(err, "unable to build kubernetes objects from release manifest")
 	}
 
+	err = resources.Visit(setMetadataVisitor(rel.Name, rel.Namespace))
+	if err != nil {
+		return nil, err
+	}
+
 	// Install requires an extra validation step of checking that resources
 	// don't already exist before we actually create resources. If we continue
 	// forward and create the release object with resources that already exist,
@@ -256,7 +261,7 @@ func (i *Install) Run(chrt *chart.Chart, vals map[string]interface{}) (*release.
 	// deleting the release because the manifest will be pointing at that
 	// resource
 	if !i.ClientOnly && !isUpgrade {
-		toBeUpdated, err := existingResourceConflict(resources, rel.Name)
+		toBeUpdated, err := existingResourceConflict(resources, rel.Name, rel.Namespace)
 		if err != nil {
 			return nil, errors.Wrap(err, "rendered manifests contain a resource that already exists. Unable to continue with install")
 		}
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -243,7 +243,7 @@ func (i *Install) Run(chrt *chart.Chart, vals map[string]interface{}) (*release.
 	// Mark this release as in-progress
 	rel.SetStatus(release.StatusPendingInstall, "Initial install underway")
 
-	var adoptedResources kube.ResourceList
+	var toBeAdopted kube.ResourceList
 	resources, err := i.cfg.KubeClient.Build(bytes.NewBufferString(rel.Manifest), !i.DisableOpenAPIValidation)
 	if err != nil {
 		return nil, errors.Wrap(err, "unable to build kubernetes objects from release manifest")
@@ -261,11 +261,10 @@ func (i *Install) Run(chrt *chart.Chart, vals map[string]interface{}) (*release.
 	// deleting the release because the manifest will be pointing at that
 	// resource
 	if !i.ClientOnly && !isUpgrade {
-		toBeUpdated, err := existingResourceConflict(resources, rel.Name, rel.Namespace)
+		toBeAdopted, err = existingResourceConflict(resources, rel.Name, rel.Namespace)
 		if err != nil {
 			return nil, errors.Wrap(err, "rendered manifests contain a resource that already exists. Unable to continue with install")
 		}
-		adoptedResources = toBeUpdated
 	}
 
 	// Bail out here if it is a dry run
@@ -300,8 +299,14 @@ func (i *Install) Run(chrt *chart.Chart, vals map[string]interface{}) (*release.
 	// At this point, we can do the install. Note that before we were detecting whether to
 	// do an update, but it's not clear whether we WANT to do an update if the re-use is set
 	// to true, since that is basically an upgrade operation.
-	if _, err := i.cfg.KubeClient.Update(adoptedResources, resources, false); err != nil {
-		return i.failRelease(rel, err)
+	if len(toBeAdopted) == 0 {
+		if _, err := i.cfg.KubeClient.Create(resources); err != nil {
+			return i.failRelease(rel, err)
+		}
+	} else {
+		if _, err := i.cfg.KubeClient.Update(toBeAdopted, resources, false); err != nil {
+			return i.failRelease(rel, err)
+		}
 	}
 
 	if i.Wait {
--- a/pkg/action/validate.go
+++ b/pkg/action/validate.go
@@ -22,36 +22,22 @@ import (
 	"github.com/pkg/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
-	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/cli-runtime/pkg/resource"
 
 	"helm.sh/helm/v3/pkg/kube"
 )
 
-var (
-	managedByReq, _ = labels.NewRequirement("app.kubernetes.io/managed-by", selection.Equals, []string{"Helm"})
-	accessor        = meta.NewAccessor()
+var accessor = meta.NewAccessor()
+
+const (
+	helmReleaseNameAnnotation      = "meta.helm.sh/release-name"
+	helmReleaseNamespaceAnnotation = "meta.helm.sh/release-namespace"
 )
 
-func newReleaseSelector(release string) (labels.Selector, error) {
-	releaseReq, err := labels.NewRequirement("app.kubernetes.io/instance", selection.Equals, []string{release})
-	if err != nil {
-		return nil, err
-	}
-
-	return labels.Parse(fmt.Sprintf("%s,%s", releaseReq, managedByReq))
-}
-
-func existingResourceConflict(resources kube.ResourceList, release string) (kube.ResourceList, error) {
-	sel, err := newReleaseSelector(release)
-	if err != nil {
-		return nil, err
-	}
-
+func existingResourceConflict(resources kube.ResourceList, releaseName, releaseNamespace string) (kube.ResourceList, error) {
 	requireUpdate := kube.ResourceList{}
 
-	err = resources.Visit(func(info *resource.Info, err error) error {
+	err := resources.Visit(func(info *resource.Info, err error) error {
 		if err != nil {
 			return err
 		}
@@ -67,13 +53,10 @@ func existingResourceConflict(resources kube.ResourceList, release string) (kube
 		}
 
 		// Allow adoption of the resource if it already has app.kubernetes.io/instance and app.kubernetes.io/managed-by labels.
-		lbls, err := accessor.Labels(existing)
-		if err == nil {
-			set := labels.Set(lbls)
-			if sel.Matches(set) {
-				requireUpdate = append(requireUpdate, info)
-				return nil
-			}
+		anno, err := accessor.Annotations(existing)
+		if err == nil && anno != nil && anno[helmReleaseNameAnnotation] == releaseName && anno[helmReleaseNamespaceAnnotation] == releaseNamespace {
+			requireUpdate = append(requireUpdate, info)
+			return nil
 		}
 
 		return fmt.Errorf(
@@ -82,4 +65,22 @@ func existingResourceConflict(resources kube.ResourceList, release string) (kube
 	})
 
 	return requireUpdate, err
+}
+
+func setMetadataVisitor(releaseName, releaseNamespace string) resource.VisitorFunc {
+	return func(info *resource.Info, err error) error {
+		if err != nil {
+			return err
+		}
+		anno, err := accessor.Annotations(info.Object)
+		if err != nil {
+			return err
+		}
+		if anno == nil {
+			anno = make(map[string]string)
+		}
+		anno[helmReleaseNameAnnotation] = releaseName
+		anno[helmReleaseNamespaceAnnotation] = releaseNamespace
+		return accessor.SetAnnotations(info.Object, anno)
+	}
 }
--- a/pkg/action/validate.go
+++ b/pkg/action/validate.go
@@ -21,13 +21,37 @@ import (
 
 	"github.com/pkg/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/cli-runtime/pkg/resource"
 
 	"helm.sh/helm/v3/pkg/kube"
 )
 
-func existingResourceConflict(resources kube.ResourceList) error {
-	err := resources.Visit(func(info *resource.Info, err error) error {
+var (
+	managedByReq, _ = labels.NewRequirement("app.kubernetes.io/managed-by", selection.Equals, []string{"Helm"})
+	accessor        = meta.NewAccessor()
+)
+
+func newReleaseSelector(release string) (labels.Selector, error) {
+	releaseReq, err := labels.NewRequirement("app.kubernetes.io/instance", selection.Equals, []string{release})
+	if err != nil {
+		return nil, err
+	}
+
+	return labels.Parse(fmt.Sprintf("%s,%s", releaseReq, managedByReq))
+}
+
+func existingResourceConflict(resources kube.ResourceList, release string) (kube.ResourceList, error) {
+	sel, err := newReleaseSelector(release)
+	if err != nil {
+		return nil, err
+	}
+
+	requireUpdate := kube.ResourceList{}
+
+	err = resources.Visit(func(info *resource.Info, err error) error {
 		if err != nil {
 			return err
 		}
@@ -42,7 +66,20 @@ func existingResourceConflict(resources kube.ResourceList) error {
 			return errors.Wrap(err, "could not get information about the resource")
 		}
 
-		return fmt.Errorf("existing resource conflict: namespace: %s, name: %s, existing_kind: %s, new_kind: %s", info.Namespace, info.Name, existing.GetObjectKind().GroupVersionKind(), info.Mapping.GroupVersionKind)
+		// Allow adoption of the resource if it already has app.kubernetes.io/instance and app.kubernetes.io/managed-by labels.
+		lbls, err := accessor.Labels(existing)
+		if err == nil {
+			set := labels.Set(lbls)
+			if sel.Matches(set) {
+				requireUpdate = append(requireUpdate, info)
+				return nil
+			}
+		}
+
+		return fmt.Errorf(
+			"existing resource conflict: namespace: %s, name: %s, existing_kind: %s, new_kind: %s",
+			info.Namespace, info.Name, existing.GetObjectKind().GroupVersionKind(), info.Mapping.GroupVersionKind)
 	})
-	return err
+
+	return requireUpdate, err
 }
--- a/pkg/action/validate.go
+++ b/pkg/action/validate.go
@@ -22,6 +22,7 @@ import (
 	"github.com/pkg/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/cli-runtime/pkg/resource"
 
 	"helm.sh/helm/v3/pkg/kube"
@@ -30,12 +31,14 @@ import (
 var accessor = meta.NewAccessor()
 
 const (
+	appManagedByLabel              = "app.kubernetes.io/managed-by"
+	appManagedByHelm               = "Helm"
 	helmReleaseNameAnnotation      = "meta.helm.sh/release-name"
 	helmReleaseNamespaceAnnotation = "meta.helm.sh/release-namespace"
 )
 
 func existingResourceConflict(resources kube.ResourceList, releaseName, releaseNamespace string) (kube.ResourceList, error) {
-	requireUpdate := kube.ResourceList{}
+	var requireUpdate kube.ResourceList
 
 	err := resources.Visit(func(info *resource.Info, err error) error {
 		if err != nil {
@@ -48,39 +51,134 @@ func existingResourceConflict(resources kube.ResourceList, releaseName, releaseN
 			if apierrors.IsNotFound(err) {
 				return nil
 			}
-
 			return errors.Wrap(err, "could not get information about the resource")
 		}
 
-		// Allow adoption of the resource if it already has app.kubernetes.io/instance and app.kubernetes.io/managed-by labels.
-		anno, err := accessor.Annotations(existing)
-		if err == nil && anno != nil && anno[helmReleaseNameAnnotation] == releaseName && anno[helmReleaseNamespaceAnnotation] == releaseNamespace {
-			requireUpdate = append(requireUpdate, info)
-			return nil
+		// Allow adoption of the resource if it is managed by Helm and is annotated with correct release name and namespace.
+		if err := checkOwnership(existing, releaseName, releaseNamespace); err != nil {
+			return fmt.Errorf("%s exists and cannot be imported into the current release: %s", resourceString(info), err)
 		}
 
-		return fmt.Errorf(
-			"existing resource conflict: namespace: %s, name: %s, existing_kind: %s, new_kind: %s",
-			info.Namespace, info.Name, existing.GetObjectKind().GroupVersionKind(), info.Mapping.GroupVersionKind)
+		requireUpdate.Append(info)
+		return nil
 	})
 
 	return requireUpdate, err
 }
 
-func setMetadataVisitor(releaseName, releaseNamespace string) resource.VisitorFunc {
+func checkOwnership(obj runtime.Object, releaseName, releaseNamespace string) error {
+	lbls, err := accessor.Labels(obj)
+	if err != nil {
+		return err
+	}
+	annos, err := accessor.Annotations(obj)
+	if err != nil {
+		return err
+	}
+
+	var errs []error
+	if err := requireValue(lbls, appManagedByLabel, appManagedByHelm); err != nil {
+		errs = append(errs, fmt.Errorf("label validation error: %s", err))
+	}
+	if err := requireValue(annos, helmReleaseNameAnnotation, releaseName); err != nil {
+		errs = append(errs, fmt.Errorf("annotation validation error: %s", err))
+	}
+	if err := requireValue(annos, helmReleaseNamespaceAnnotation, releaseNamespace); err != nil {
+		errs = append(errs, fmt.Errorf("annotation validation error: %s", err))
+	}
+
+	if len(errs) > 0 {
+		err := errors.New("invalid ownership metadata")
+		for _, e := range errs {
+			err = fmt.Errorf("%w; %s", err, e)
+		}
+		return err
+	}
+
+	return nil
+}
+
+func requireValue(meta map[string]string, k, v string) error {
+	actual, ok := meta[k]
+	if !ok {
+		return fmt.Errorf("missing key %q: must be set to %q", k, v)
+	}
+	if actual != v {
+		return fmt.Errorf("key %q must equal %q: current value is %q", k, v, actual)
+	}
+	return nil
+}
+
+// setMetadataVisitor adds release tracking metadata to all resources. If force is enabled, existing
+// ownership metadata will be overwritten. Otherwise an error will be returned if any resource has an
+// existing and conflicting value for the managed by label or Helm release/namespace annotations.
+func setMetadataVisitor(releaseName, releaseNamespace string, force bool) resource.VisitorFunc {
 	return func(info *resource.Info, err error) error {
 		if err != nil {
 			return err
 		}
-		anno, err := accessor.Annotations(info.Object)
-		if err != nil {
-			return err
+
+		if !force {
+			if err := checkOwnership(info.Object, releaseName, releaseNamespace); err != nil {
+				return fmt.Errorf("%s cannot be owned: %s", resourceString(info), err)
+			}
 		}
-		if anno == nil {
-			anno = make(map[string]string)
+
+		if err := mergeLabels(info.Object, map[string]string{
+			appManagedByLabel: appManagedByHelm,
+		}); err != nil {
+			return fmt.Errorf(
+				"%s labels could not be updated: %s",
+				resourceString(info), err,
+			)
 		}
-		anno[helmReleaseNameAnnotation] = releaseName
-		anno[helmReleaseNamespaceAnnotation] = releaseNamespace
-		return accessor.SetAnnotations(info.Object, anno)
+
+		if err := mergeAnnotations(info.Object, map[string]string{
+			helmReleaseNameAnnotation:      releaseName,
+			helmReleaseNamespaceAnnotation: releaseNamespace,
+		}); err != nil {
+			return fmt.Errorf(
+				"%s annotations could not be updated: %s",
+				resourceString(info), err,
+			)
+		}
+
+		return nil
 	}
+}
+
+func resourceString(info *resource.Info) string {
+	_, k := info.Mapping.GroupVersionKind.ToAPIVersionAndKind()
+	return fmt.Sprintf(
+		"%s %q in namespace %q",
+		k, info.Name, info.Namespace,
+	)
+}
+
+func mergeLabels(obj runtime.Object, labels map[string]string) error {
+	current, err := accessor.Labels(obj)
+	if err != nil {
+		return err
+	}
+	return accessor.SetLabels(obj, mergeStrStrMaps(current, labels))
+}
+
+func mergeAnnotations(obj runtime.Object, annotations map[string]string) error {
+	current, err := accessor.Annotations(obj)
+	if err != nil {
+		return err
+	}
+	return accessor.SetAnnotations(obj, mergeStrStrMaps(current, annotations))
+}
+
+// merge two maps, always taking the value on the right
+func mergeStrStrMaps(current, desired map[string]string) map[string]string {
+	result := make(map[string]string)
+	for k, v := range current {
+		result[k] = v
+	}
+	for k, desiredVal := range desired {
+		result[k] = desiredVal
+	}
+	return result
 }
--- a/pkg/action/upgrade.go
+++ b/pkg/action/upgrade.go
@@ -203,6 +203,11 @@ func (u *Upgrade) performUpgrade(originalRelease, upgradedRelease *release.Relea
 		return upgradedRelease, errors.Wrap(err, "unable to build kubernetes objects from new release manifest")
 	}
 
+	err = target.Visit(setMetadataVisitor(upgradedRelease.Name, upgradedRelease.Namespace))
+	if err != nil {
+		return upgradedRelease, err
+	}
+
 	// Do a basic diff using gvk + name to figure out what new resources are being created so we can validate they don't already exist
 	existingResources := make(map[string]bool)
 	for _, r := range current {
@@ -216,7 +221,7 @@ func (u *Upgrade) performUpgrade(originalRelease, upgradedRelease *release.Relea
 		}
 	}
 
-	toBeUpdated, err := existingResourceConflict(toBeCreated, upgradedRelease.Name)
+	toBeUpdated, err := existingResourceConflict(toBeCreated, upgradedRelease.Name, upgradedRelease.Namespace)
 	if err != nil {
 		return nil, errors.Wrap(err, "rendered manifests contain a resource that already exists. Unable to continue with update")
 	}
--- a/pkg/action/upgrade.go
+++ b/pkg/action/upgrade.go
@@ -203,7 +203,8 @@ func (u *Upgrade) performUpgrade(originalRelease, upgradedRelease *release.Relea
 		return upgradedRelease, errors.Wrap(err, "unable to build kubernetes objects from new release manifest")
 	}
 
-	err = target.Visit(setMetadataVisitor(upgradedRelease.Name, upgradedRelease.Namespace))
+	// It is safe to use force only on target because these are resources currently rendered by the chart.
+	err = target.Visit(setMetadataVisitor(upgradedRelease.Name, upgradedRelease.Namespace, true))
 	if err != nil {
 		return upgradedRelease, err
 	}
--- a/pkg/action/upgrade.go
+++ b/pkg/action/upgrade.go
@@ -216,9 +216,18 @@ func (u *Upgrade) performUpgrade(originalRelease, upgradedRelease *release.Relea
 		}
 	}
 
-	if err := existingResourceConflict(toBeCreated); err != nil {
-		return nil, errors.Wrap(err, "rendered manifests contain a new resource that already exists. Unable to continue with update")
+	toBeUpdated, err := existingResourceConflict(toBeCreated, upgradedRelease.Name)
+	if err != nil {
+		return nil, errors.Wrap(err, "rendered manifests contain a resource that already exists. Unable to continue with update")
 	}
+
+	toBeUpdated.Visit(func(r *resource.Info, err error) error {
+		if err != nil {
+			return err
+		}
+		current.Append(r)
+		return nil
+	})
 
 	if u.DryRun {
 		u.cfg.Log("dry run for %s", upgradedRelease.Name)
--- a/pkg/action/validate_test.go
+++ b/pkg/action/validate_test.go
@@ -1,0 +1,123 @@
+/*
+Copyright The Helm Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package action
+
+import (
+	"testing"
+
+	"helm.sh/helm/v3/pkg/kube"
+
+	appsv1 "k8s.io/api/apps/v1"
+
+	"github.com/stretchr/testify/assert"
+	"k8s.io/apimachinery/pkg/api/meta"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/cli-runtime/pkg/resource"
+)
+
+func newDeploymentResource(name, namespace string) *resource.Info {
+	return &resource.Info{
+		Name: name,
+		Mapping: &meta.RESTMapping{
+			Resource:         schema.GroupVersionResource{Group: "apps", Version: "v1", Resource: "deployment"},
+			GroupVersionKind: schema.GroupVersionKind{Group: "apps", Version: "v1", Kind: "Deployment"},
+		},
+		Object: &appsv1.Deployment{
+			ObjectMeta: v1.ObjectMeta{
+				Name:      name,
+				Namespace: namespace,
+			},
+		},
+	}
+}
+
+func TestCheckOwnership(t *testing.T) {
+	deployFoo := newDeploymentResource("foo", "ns-a")
+
+	// Verify that a resource that lacks labels/annotations is not owned
+	err := checkOwnership(deployFoo.Object, "rel-a", "ns-a")
+	assert.EqualError(t, err, `invalid ownership metadata; label validation error: missing key "app.kubernetes.io/managed-by": must be set to "Helm"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "rel-a"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "ns-a"`)
+
+	// Set managed by label and verify annotation error message
+	_ = accessor.SetLabels(deployFoo.Object, map[string]string{
+		appManagedByLabel: appManagedByHelm,
+	})
+	err = checkOwnership(deployFoo.Object, "rel-a", "ns-a")
+	assert.EqualError(t, err, `invalid ownership metadata; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "rel-a"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "ns-a"`)
+
+	// Set only the release name annotation and verify missing release namespace error message
+	_ = accessor.SetAnnotations(deployFoo.Object, map[string]string{
+		helmReleaseNameAnnotation: "rel-a",
+	})
+	err = checkOwnership(deployFoo.Object, "rel-a", "ns-a")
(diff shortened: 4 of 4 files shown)


pkg/action/validate.go
```diff
@@ -42,7 +66,20 @@ func existingResourceConflict(resources kube.ResourceList) error {
 			return errors.Wrap(err, "could not get information about the resource")
 		}
 
-		return fmt.Errorf("existing resource conflict: namespace: %s, name: %s, existing_kind: %s, new_kind: %s", info.Namespace, info.Name, existing.GetObjectKind().GroupVersionKind(), info.Mapping.GroupVersionKind)
+		// Allow adoption of the resource if it already has app.kubernetes.io/instance and app.kubernetes.io/managed-by labels.
+		lbls, err := accessor.Labels(existing)
+		if err == nil {
+			set := labels.Set(lbls)
+			if sel.Matches(set) {
+				requireUpdate = append(requireUpdate, info)
```
jlegrone: Here we return the `resource.Info` for the _latest_ version of the chart, not the resource that we've retrieved from the cluster. This might cause changes to not be applied when computing the 3-way diff; this still needs to be verified.

pkg/action/validate.go
```diff
@@ -42,7 +49,34 @@ func existingResourceConflict(resources kube.ResourceList) error {
 			return errors.Wrap(err, "could not get information about the resource")
 		}
 
-		return fmt.Errorf("existing resource conflict: namespace: %s, name: %s, existing_kind: %s, new_kind: %s", info.Namespace, info.Name, existing.GetObjectKind().GroupVersionKind(), info.Mapping.GroupVersionKind)
+		// Allow adoption of the resource if it already has app.kubernetes.io/instance and app.kubernetes.io/managed-by labels.
+		anno, err := accessor.Annotations(existing)
+		if err == nil && anno != nil && anno[helmReleaseAnnotation] == release {
+			requireUpdate = append(requireUpdate, info)
```
jlegrone: Here we return the `resource.Info` for the _latest_ version of the chart, not the resource that we've retrieved from the cluster. This might cause changes to not be applied when computing the 3-way diff; this still needs to be verified.

pho-enix (commented):
Nice solution!

Personally I would prefer to re-use the
`app.kubernetes.io/managed-by`
`app.kubernetes.io/instance`
annotations and auto-inject those.

A lot of people already use them in their Charts or at least know then. Meaning less confusion because of double labelling. Additionally this would push the adoption of a best-practice.

jlegrone:
See also discussion on labels in https://github.com/helm/helm/issues/6365#issuecomment-547151315

bacongobbler:
thanks @jlegrone. I'm going to give this a spin today. Looks like there are some significant behavioural changes under the hood to `helm install` with the change to call `KubeClient.Update`, so I want to make sure we're not introducing any regressions here.

bacongobbler:
is this still a work-in-progress, or can we remove that label?

jlegrone:
> is this still a work-in-progress, or can we remove that label?

@bacongobbler yes, see TODOs in the description. I think the main outstanding issue is to validate 3-way diff behavior, which is why I've kept the wip label for now.

pkg/action/validate.go
```diff
@@ -42,7 +52,35 @@ func existingResourceConflict(resources kube.ResourceList) error {
 			return errors.Wrap(err, "could not get information about the resource")
 		}
 
-		return fmt.Errorf("existing resource conflict: namespace: %s, name: %s, existing_kind: %s, new_kind: %s", info.Namespace, info.Name, existing.GetObjectKind().GroupVersionKind(), info.Mapping.GroupVersionKind)
+		// Allow adoption of the resource if it already has app.kubernetes.io/instance and app.kubernetes.io/managed-by labels.
```
bacongobbler: the annotations listed above don't match up to the comment provided here. Does the comment here need to be updated?

pkg/action/install.go
```diff
@@ -291,7 +300,7 @@ func (i *Install) Run(chrt *chart.Chart, vals map[string]interface{}) (*release.
 	// At this point, we can do the install. Note that before we were detecting whether to
 	// do an update, but it's not clear whether we WANT to do an update if the re-use is set
 	// to true, since that is basically an upgrade operation.
-	if _, err := i.cfg.KubeClient.Create(resources); err != nil {
+	if _, err := i.cfg.KubeClient.Update(adoptedResources, resources, false); err != nil {
```
bacongobbler: if `adoptedResources` is empty, should we revert back to a `Create` operation here?

pkg/action/validate.go
```diff
@@ -21,12 +21,22 @@ import (
...
 
 	"helm.sh/helm/v3/pkg/kube"
 )
 
-func existingResourceConflict(resources kube.ResourceList) error {
+var accessor = meta.NewAccessor()
+
+const (
+	helmReleaseNameAnnotation      = "meta.helm.sh/release-name"
+	helmReleaseNamespaceAnnotation = "meta.helm.sh/release-namespace"
```
bacongobbler: Interesting thought... How do we want to handle global objects like ClusterRoles? I know they're kind of an edge case, but what is the expected behaviour if they have these annotations? As it stands, it looks like you could add an annotation like

```
meta.helm.sh/release-name: foo
meta.helm.sh/release-namespace: foo
```

To an existing ClusterRole, call `helm install --namespace foo foo` with a chart that installs this ClusterRole and it'll adopt the ClusterRole. Is that intentional behaviour here? If not, how do we want to handle these resources?

pkg/action/install.go
```diff
@@ -242,21 +243,29 @@ func (i *Install) Run(chrt *chart.Chart, vals map[string]interface{}) (*release.
...
+
 	// Install requires an extra validation step of checking that resources
 	// don't already exist before we actually create resources. If we continue
 	// forward and create the release object with resources that already exist,
 	// we'll end up in a state where we will delete those resources upon
 	// deleting the release because the manifest will be pointing at that
 	// resource
 	if !i.ClientOnly && !isUpgrade {
-		if err := existingResourceConflict(resources); err != nil {
+		toBeUpdated, err := existingResourceConflict(resources, rel.Name, rel.Namespace)
```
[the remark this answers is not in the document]
bacongobbler: they haven't been adopted at this point.

bacongobbler (commented):
A few quick comments as I'm reading the code. Testing this out now to verify behaviour

pkg/action/validate.go
jlegrone: ah, yep that's a holdover from the initial implementation

pkg/action/validate.go
jlegrone: This is intentional behavior. Please correct me if I'm wrong, but my understanding is that Helm requires every release to be associated with a namespace, even if that release only acts on cluster scoped resources.

bacongobbler:
A basic test looks promising!

I created two pods: `foo` and `bar`. One had the annotations, the other didn't.

Here's what I used for testing: [testchart.zip](https://github.com/helm/helm/files/4294581/testchart.zip)

To create the pods, I used:

```
><> unzip testchart
><> helm template foo testchart | kubectl create -f -
><> helm template bar testchart --set addAnnotations=true | kubectl create -f -
```

Then I tested with the following:

```
><> helm install car testchart
NAME: car
LAST DEPLOYED: Thu Mar  5 10:44:28 2020
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
><> helm install foo testchart
Error: rendered manifests contain a resource that already exists. Unable to continue with install: existing resource conflict: namespace: default, name: foo, existing_kind: /v1, Kind=Pod, new_kind: /v1, Kind=Pod
><> helm install bar testchart
NAME: bar
LAST DEPLOYED: Thu Mar  5 10:44:34 2020
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
```

It looks like the `existing_kind` and `new_kind` error messages when there's a resource conflict aren't populated correctly. That looks unrelated to this PR as I can see this with helm 3.1.1 as well. I'll file a new ticket for that.

bacongobbler:
3-way merge seems to check out.

I modified the template a little to add junk labels to the manifest when requested: [testchart.zip](https://github.com/helm/helm/files/4294781/testchart.zip)

```
$ helm template foo . --set addAnnotations=true | kubectl create -f -
$ helm template bar . --set addAnnotations=true,addLabels=true | kubectl create -f -
$ helm install foo . --set addLabels=true
$ helm install bar .
```

Both `foo` and `bar` had the labels and annotations attached as expected.

Another test was to attempt to take ownership over an object residing in the cluster during an upgrade:

```
$ helm template foo . --set addAnnotations=true | kubectl create -f -
$ helm template bar . --set addAnnotations=true | kubectl create -f -
$ helm install foo .
$ kubectl edit pod bar # change meta.helm.sh/release-name to "foo"
```

I added `bar` to the template, intending to take ownership of the object in the next upgrade:

```
apiVersion: v1
kind: Pod
metadata:
  name: {{ .Release.Name }}
  {{ if .Values.addLabels -}}
  labels:
    meta.helm.sh/release-name: "{{ .Release.Name }}"
    meta.helm.sh/release-namespace: "{{ .Release.Namespace }}"
  {{- end }}
  {{ if .Values.addAnnotations -}}
  annotations:
    meta.helm.sh/release-name: "{{ .Release.Name }}"
    meta.helm.sh/release-namespace: "{{ .Release.Namespace }}"
  {{- end }}
spec:
  containers:
  - name: {{ .Release.Name }}
    image: busybox
    command: ['sh', '-c', 'echo Hello Kubernetes! && sleep 3600']
---
apiVersion: v1
kind: Pod
metadata:
  name: bar
  {{ if .Values.addLabels -}}
  labels:
    meta.helm.sh/release-name: "{{ .Release.Name }}"
    meta.helm.sh/release-namespace: "{{ .Release.Namespace }}"
  {{- end }}
  {{ if .Values.addAnnotations -}}
  annotations:
    meta.helm.sh/release-name: "{{ .Release.Name }}"
    meta.helm.sh/release-namespace: "{{ .Release.Namespace }}"
  {{- end }}
spec:
  containers:
  - name: bar
    image: busybox
    command: ['sh', '-c', 'echo Hello Kubernetes! && sleep 3600']
```

```
$ helm upgrade foo .
```

Finally, doing that same workflow, but I did not add any ownership annotations to the resource.

```
$ helm template foo . --set addAnnotations=true | kubectl create -f -
$ helm template bar . --set addAnnotations=true | kubectl create -f -
$ helm install foo .
$ # edit the template as described above to take ownership of "bar"
$ helm upgrade foo .
Error: UPGRADE FAILED: rendered manifests contain a resource that already exists. Unable to continue with update: existing resource conflict: namespace: default, name: bar, existing_kind: /v1, Kind=Pod, new_kind: /v1, Kind=Pod
```

Looks like everything checks out here!

pkg/action/validate.go
bacongobbler: Yep! Seems like it's in alignment here. So these annotations relate to the name/namespace this object is "owned" by. :+1:

pkg/action/install.go
```diff
@@ -242,21 +243,29 @@ func (i *Install) Run(chrt *chart.Chart, vals map[string]interface{}) (*release.
 	// Mark this release as in-progress
 	rel.SetStatus(release.StatusPendingInstall, "Initial install underway")
 
+	var adoptedResources kube.ResourceList
 	resources, err := i.cfg.KubeClient.Build(bytes.NewBufferString(rel.Manifest), !i.DisableOpenAPIValidation)
 	if err != nil {
 		return nil, errors.Wrap(err, "unable to build kubernetes objects from release manifest")
 	}
 
+	err = resources.Visit(setMetadataVisitor(rel.Name, rel.Namespace))
```
bacongobbler: So we're attaching this annotation to any objects going forward, both during an install AND an upgrade. NICE!

bacongobbler (approved):
Code LGTM. Given the usefulness of this feature, would you mind writing up some documentation for this? I know a lot of users would love to know how to adopt existing resources a helm release. This enables a lot of use cases that previously weren't possible.

jlegrone:
@bacongobbler I'd be happy to add a release note. Should I just add a comment to this PR or would that be a PR against the community repo?

I'm removing the WIP label; test cases have been added and error messages should now be much more descriptive.

My last question is whether we should go with labels instead of annotations. It could be very useful to have consistent labels for querying resources, both for end users and for implementing things like #6364. At the same time I'm not a fan of having duplicate values for `app.kubernetes.io/instance` and `meta.helm.sh/release` labels everywhere. Perhaps this is moot if we think it's feasible to support both these new annotations and standard k8s labels in the future.

dudicoco:
@jlegrone thank you for this amazing contribution! A much needed feature.
Regarding the labels/annotations - after adding these if needed and having helm adopt the resources can the labels/annotations be removed in case I do not want my resources to include them?

jlegrone:
> after adding these if needed and having helm adopt the resources can the labels/annotations be removed in case I do not want my resources to include them?

The labels and annotations used to track ownership will be added to all resources managed by your chart each time you do a deployment; this behavior is required in order to tackle #7699. So removing ownership metadata would usually not make sense because it will just be added back on the next `upgrade`.

technosophos (approved):
I am in favor of this feature. Code LGTM.

I am slightly concerned that Option A could under some circumstances have unforeseen side effects. But the only example I can think of is this:

- I have a release named `foo` that has a deployment
- I decide that I no longer want that release to be managed by Helm, so I delete the Helm release
- Sometime later, someone comes along and creates a new release named `foo`. because the release is gone for that, Helm allows this

In that case, it seems  that Option A could lead to the surprising adoption of the deployment. Given that the workflow described above is "unsupported" from Helm's standpoint, though, I don't know that I would consider it a blocker.

So I'm gonna go ahead and LGTM, leaving that decision up to you.

jlegrone:
> * I decide that I no longer want that release to be managed by Helm, so I delete the Helm release
> * Sometime later, someone comes along and creates a new release named foo. because the release is gone for that, Helm allows this

Yes, this a potential problem. I tried to hedge against it by requiring the `app.kubernetes.io/managed-by` label to be set to `Helm` in addition to requiring release annotations to match.

In practice I don't know of any deployment tools that actually set this label, but if this becomes a more standard practice in the future that will improve the safety of Helm's resource adoption model.  It might not be a bad idea for us to bring this up in sig-apps and open issues against Spinnaker, kapp, etc. to get this done.

technosophos:
That makes sense. So I'll re-iterate that I am good to go on this one.

hickeyma (commented):
Thanks for pushing this functionality @jlegrone. Sorry but only getting to provide feedback now and missed the boat before merging.

Its looks good and manual testing worked well. One small thing came up with the scaffold chart not containing the adopt annotations. It therefore doesn't work from a chart created using `helm create` without modifications. I have opened PR  #7786 to handle this.

marko-espi:
In which version this feature will be available? I don't see them in release notes of helm 3

jlegrone:
> In which version this feature will be available? I don't see them in release notes of helm 3

[3.2.0](https://github.com/helm/helm/milestone/83), which is not yet released. @bacongobbler should we consider a release candidate for this change to get some early feedback?

zifeo:
@jlegrone I think this would be a great idea. I already tested on master and looks good so far. However do we agree that we have to annotate tag the release before being able to adopt it? Maybe this should be added to the changelog.

jlegrone:
> However do we agree that we have to annotate tag the release before being able to adopt it? Maybe this should be added to the changelog.

@zifeo does the release note in this PR look adequate to you? Applying these annotations will also be implemented as a Helm subcommand (possibly as early as the 3.2.0 release). See #2730

zifeo:
@jlegrone sounds good to me!

bacongobbler:
RE: cutting a 3.2.0 release candidate, there are a few tickets in the milestone yet to be resolved before we go ahead with a release candidate: https://github.com/helm/helm/milestone/83

For now, users can use the [canary release](https://helm.sh/docs/intro/install/#from-canary-builds) to test this out in a development environment.

Hope this helps

hickeyma:
> I'd be happy to add a release note. Should I just add a comment to this PR or would that be a PR against the community repo?

@jlegrone I would suggest opening a PR in the docs repo, adding a section to https://github.com/helm/helm-www/blob/master/content/en/docs/topics/advanced.md

ento:
Does this also fix #7697?

lucasreed:
> Does this PR also fix #7697?

Yes this should fix that issue.

Wobbley:
After digging through numerous issues concerning `existing resource conflict`. I seem to have reached the holy grail. However it seems undocumented so far? Is there documentation anywhere on how to use this new feature?

olivierboudet:
I did not found any documentation either but error messages where trying to install with existing resources are pretty clear with helm 3.2.0 :)

And thanks for this, it helps a lot !

pkg/action/validate.go
```diff
@@ -21,12 +21,25 @@ import (
...
 
 	"helm.sh/helm/v3/pkg/kube"
 )
 
-func existingResourceConflict(resources kube.ResourceList) error {
+var accessor = meta.NewAccessor()
+
+const (
+	appManagedByLabel              = "app.kubernetes.io/managed-by"
+	appManagedByHelm               = "Helm"
```
liuming-dev, line 35:
```yaml
apiVersion: v1
data:
  kubernetesServiceHost: pre.master.thanos.sogou
  kubernetesServicePort: "6443"
kind: ConfigMap
metadata:
  annotations:
    meta.helm.sh/release-name: configmap-test
    meta.helm.sh/release-namespace: bizdev
  creationTimestamp: "2020-05-07T11:33:46Z"
  labels:
    123.a: 456.b
  name: configmap-test
  namespace: bizdev
  resourceVersion: "359608781"
  selfLink: /api/v1/namespaces/bizdev/configmaps/configmap-test
  uid: 6109837e-d753-41fb-be76-ad3145341978
```

I think `app.kubernetes.io/managed-by` value in lower camel case will be better, like `app.kubernetes.io/managed-by: helm`, to make the style consistency.

tomcruise81:
Should this be cleaning up resources that were "adopted" but are not part of the new chart that's being installed? Lets say:

1. there's a ConfigMap that is part of the pre-Helm version of the application
1. the appropriate labels and annotations are applied to that ConfigMap
1. the helm chart that would "adopt" the other resources for the given application is installed

Shouldn't the update/install take care of removing that ConfigMap?

bacongobbler:
Feel free to open a PR/feature request for that functionality, @tomcruise81. I'm not sure that was in scope for this PR, but I could see a use for that functionality.

abraverm:
It doesn't seem to work with hook resources (Helm v3.2.3) such as pre-install

casey-robertson:
Sadly it's not working for me.  Added the 2 annotations outlined in the solution (the label was already there) and it's still throwing the upgrade error.

TBBle:
@thelo-gaultier That's a different problem to this, see #7219 and https://github.com/hickeyma/helm-mapkubeapis. The `helm-mapkubeapis` plugin specifically supports `rbac.authorization.k8s.io` migration (https://github.com/hickeyma/helm-mapkubeapis/issues/8). I'm pretty sure adding the 'adopt' metadata was the wrong fix, as something deployed by Helm 3.2.0 would already _have_ the metadata.

That said, I suspect this working for your use-case was accidental, and possibly a bug.

My guess of the flow is below, assuming that I have your use-case correct:

Installed already is a chart with `rbac.authorization.k8s.io/v1beta1 ClusterRole` ("old version"), and you're upgrading it with a chart that has been updated to `rbac.authorization.k8s.io/v1 ClusterRole` ("new version") with the same name.

Without the annotations: "old version" is deleted, "new version" is created. Because getting the "new version" returns something (the old version in the new format), the upgrade is rejected.

With the annotations: The 'new version' is scheduled to be replaced with itself, as we pretend it always existed in the chart. However, that should have left the "old version" in the delete list, and since [updates](https://github.com/helm/helm/blob/a28d695c43ad5b889923c5cf0e7694503bdb47df/pkg/kube/client.go#L204) happen before [deletes](https://github.com/helm/helm/blob/a28d695c43ad5b889923c5cf0e7694503bdb47df/pkg/kube/client.go#L236), the net output should have been updating the resource (by the new name) and *then deleting the resource* (by the old name).

I don't know why that didn't go badly wrong for you. I don't see anything in this PR that would protect that, unless I'm misunderstanding something.

ThoTischner:
Are the docs updated to reflect this possibility?
Can not find the new annotations in the docs.

TBBle:
Since this PR is merged (and hence "done"), I opened https://github.com/helm/helm-www/issues/755 so that at least we're tracking that the documentation is missing.

rodalli:
> It doesn't seem to work with hook resources (Helm v3.2.3) such as pre-install

I'm observing the same behavior on 3.2.4

Chandrakanth-R:
For ClusterRole which is not a namespaced resource from kubernetes, if helm assumes the right to adopt with annotations containing namespace, there is a problem when the chart has to be re-deployed in the same cluster in a different namespace. For eg, when the chart is deployed for the first time in `namespace1`, ClusterRole is created. In the second attempt, when it is tried to install the chart on a different namespace `namespace2`,  helm will try to adopt the existing ClusterRole and there is a error thrown.

> Error: rendered manifests contain a resource that already exists. Unable to continue with install: ClusterRole "XXXX" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-namespace" must equal "namespace2": current value is "namespace1"

TBBle:
In that case, you will have to manually edit the resource's metadata to indicate that you want it to be adopted. This is the same as if you'd created the `ClusterRole` by some other mechanism than Helm.

The only time Helm can safely adopt a resource without manual intervention if it it is orphaned and then adopted by the _same_ release in the _same_ namespace, since clearly if it can be installed, then there isn't already a release by that name in that namespace. That is why it will only adopt resources that are already marked as belonging to the installation being attempted.

There's no way for Helm to know that it's safe to adopt that resource otherwise, as it cannot guarantee that it can accurately see the status of the release that the metadata points to.

#2730 was a proposal to add tooling to edit the metadata for you, but it hasn't moved in a while.

Chandrakanth-R:
Okay so I understand that I will have to manually update `meta.helm.sh/release-namespace` value to namespace2. What would be the effect of this for the resources in namespace1?
This issue is seen because helm adds the annotations while creating ClusterRole. And while deploying it checks for the annotations. Annotations will not be added if ClusterRole is created directly using kubernetes.

TBBle:
Wait...

I had read 're-deployed' as "uninstalled and reinstalled", I guess perhaps you meant "installed a second copy"?

Just to be clear, you should only change those annotations if the `ClusterRole` has been orphaned, left behind by another release of the chart *after uninstallation*. If you still have the release in namespace1, then don't adopt that `ClusterRole` into the release in `namespace2`, or your release in namespace1 will be in an invalid state, and you're creating the same or worse problems for yourself (or someone else) later.

If this is your situation, then it's a problem in the chart itself: it should not be creating cluster-scoped resources that will conflict with other installations of the same chart in different namespaces. Some charts that do this will have an option to disable installing the `ClusterRole` (or other cluster-wide resources) so that the second parallel installation onwards will not conflict with the first, but things will break if you uninstall the first install before the others.

A better solution is for cluster-wide resources from a chart to contain the name of the namespace they were installed into, so they won't conflict. Or put the resources into a different chart that you install precisely once. The latter case is a bit like the situation with CRDs, which are cluster-wide resources that you can't rename.

Chandrakanth-R:
Sorry for the confusion on "re-deployed" statement. Yes I tried to install a second copy.

I accept your point of having namespace in the ClusterRole name. But from kubernetes point of view ClusterRole is not namespaced resource. Helm is introducing the annotations and restricting the resource to be re-used in a different namespace.

TBBle:
Helm is not 'restricting the resource to be re-used'. Helm is restricting 'the resource from being *owned* by more than one installation'. Helm does not support having a resource owned by two different chart deployments, whether they're in the same or different namespace, or are the same or different chart or not. It's the chart-creator's responsibility to ensure that their chart only produces unique resources, including unique non-namespace resources.

The one option the chart creator does have to 'share' resources is a hook, which will create the resource, but never delete it. It's not really _owned_ by the chart, it's more like a resource-creation script that is run at install-time. As I mentioned, CRDs are a special-case example of non-namespaced resources, with their own hook-like behaviour.

This is now a different topic from the 'adoption' mechanism for this MR. as the adoption mechanism is for taking ownership of _unowned_ resources (that maybe used to be owned by a chart, but that's not the main use-case, I believe).

I'd suggest opening a new GitHub issue if you want to explore your situation and needs further.

dancnfoo:
Seems like this change is backward incompatible with older Helm installations.

We have a situation that this change fails our cluster upgrade process,

1. Old cluster k8s 1.16 had releases installed with helm 3.0.2
2. Cluster is updated with new image which has k8s 1.18, helm 3.4.1 and updated charts
3. Now `helm-3.4.1 upgrade <existing release> <updated chart>` fails with,
`Error: UPGRADE FAILED: rendered manifests contain a resource that already exists. Unable to continue with update: ClusterRoleBinding "<rolebinding name>" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: missing key "app.kubernetes.io/managed-by": must be set to "Helm"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "<release name>"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "<namespace>"`

To work around this error, seems we are left with 2 choices,

1. For each k8s resource deployed from previous Helm charts, run the `k annotate` and `k label` commands to update them.  This would require us to gather all the resources for each chart and update manually.
2. `helm-3.4.1 upgrade<existing release> <old charts>` _BEFORE_ `helm-3.4.1 upgrade <existing release> <updated charts>` to let helm to backfill needed annotations and label.  This requires us to carry old charts when upgrading new charts.

Neither approach above seem ideal.  Is there a better alternative?

TBBle:
Does this happen for _all_ resources or just some resources? The error suggests that the "updated charts" include resources that weren't previously part of the original chart, and those resources already existed outside Helm on your cluster.

Perhaps there's a different bug? A resource that was part of the old install should not be "imported into the current release", the annotations are not how Helm knows if something is part of the current release, that data is held in the release Secret. The annotations are only relevant when trying to adopt non-Helm resources into a Helm release.

Do you have a public chart where your problem manifests, so it can be reproduced?

If `helm-3.4.1 upgrade <existing release> <old charts>` and then `helm-3.4.1 upgrade <existing release> <updated charts>` fixes the problem, the issue might be related to silent resource upgrades between k8s versions, where the chart has also been upgraded. Although {{ClusterRoleBinding}} shouldn't be one of those.

lonnix:
I'm also having a problem switching from 3.0.2. I am trying to move a release to a new namespace, so I intentionally orphaned some resources then tried to adopt them with a new deployment

1. I created a helm deployment with 3.0.2 in the `default` namespace
2. I used `kubectl` to delete the helm release secrets in the `default` namespace
3. I manually changed `meta.helm.sh/release-name` and `meta.helm.sh/release-namespace` to the new, correct namespace and name (via `kubectl edit ...`)
4. The label `app.kubernetes.io/managed-by=Helm` was already present and still is
5. I ran `helm upgrade --install --namespace new-namespace new-release-name` and got this error: (still using 3.0.2)
```
Error: rendered manifests contain a resource that already exists. Unable to continue with install: existing resource conflict: kind: Service, namespace: myNamespace, name: myService
```

I upgraded to helm 3.4.1 and re-ran step #5 and got the exact same error.

Also worth noting, the `addLabels` and `addAnnotations` options do not seem to work for me with `helm template`. If I run `helm template .... --set addAnnotations=true,addLabels=true` I do not the label or either of the annotations on any of the objects

dancnfoo:
In our case, there was an API version upgrade (`certmanager.k8s.io/v1alpha1` to `cert-manager.io/v1` from k8s 1.16 to k8s 1.18 ) in a sub chart of the rel that required us to run `helm mapkubeapis` beforehand.

Not sure if the discontinued API version could be what contributed to the "Helm trying to adopt non-Helm resources but failed" behavior?

To workaround this, we opted to patch all resources from previous release with the annotations and label before upgrade.

TBBle:
The cert-manager documentation seems to [strongly disrecommend](https://cert-manager.io/docs/installation/upgrading/upgrading-0.10-0.11/) upgrading in-place from their v1alpha1 API under `certmanager.k8s.io/v1alpha1` to their new api under `cert-manager.io/v1`, and I suspect the helm-mapkubeapis plugin was the wrong thing to use with that transition.

I don't think helm-mapkubeapis changes the state of the resources on the cluster, just the data Helm holds, so unless you _also_ did the manual uninstall/reinstall of your cert-manager resources per that above link, that should be unrelated.

Built-in kubernetes resources have magic behind them that exposes even removed API resources at their replacement API versions, which is what helm-mapkubeapis fixes the Helm-side view of.

For CRDs, when the old API CRD is removed, all the related data is deleted too.

What I think happened is that you have run helm-mapkubeapis, which upgraded the Helm metadata for `ClusterRoleBinding` from `rbac.authorization.k8s.io/v1alpha1` to `rbac.authorization.k8s.io/v1` because you're on 1.18, but your chart (or some subchart) still uses the old API Version `rbac.authorization.k8s.io/v1alpha1`.

This means that the chart now tries to create `rbac.authorization.k8s.io/v1alpha1` `ClusterRoleBinding` as new, because the metadata says it doesn't exist, but the resource _does_ exist.

So my suspicion is that you shouldn't have used helm-mapkubeapis, or perhaps have used it wth a map file that doesn't include the `ClusterRoleBinding`.

This is one of the reasons it's a plugin, and not a core part of Helm, because the correct changes to make depend on the changes made in the chart that is _going_ to be applied. and helm-mapkubeapis doesn't know that, so you have to create the appropriate map yourself.

helm-mapkubeapis isn't something to just be run because your cluster is upgraded. Things work fine in that case, it's for running when your _charts_ have changed their API versions and you're trying to upgrade across the gap.

Making the appropriate map is hard. I'd suggest using something like the helm-diff plugin to compare your proposed upgraded chart with the current metadata, and identify any proposed delete/adds that are supposed to be updates as potential map entries.

hickeyma:
@TBBle Thanks for the feedback in https://github.com/helm/helm/pull/7649#issuecomment-733622185.

> helm-mapkubeapis isn't something to just be run because your cluster is upgraded. Things work fine in that case, it's for running when your charts have changed their API versions and you're trying to upgrade across the gap.

This is a good point. Maybe the plugin should only update once an API is removed and not when deprecated?

TBBle:
I think the plugin is advertised to fix the wrong problem, and that it's because the problem has been mis-described in the [Helm docs](https://helm.sh/docs/topics/kubernetes_apis/), and points people at the plugin as part of the "what if APIs have been removed by k8s version upgrades?" docs. So I think the question is more complex than 'removed' vs 'deprecated'.

----

The problem is described as an issue of removal of deprecated APIs from kubernetes, but that's not the issue, that's the underlying trigger for chart changes that cause the problem.

The problem that helm-mapkubeapis solves is that if a _chart_ has the APIVersion of a multi-versioned resource changed between last installed version and version being upgraded to, without changing the name. This can happen even if the old API is still present on the cluster, i.e. without a kubernetes update; and doesn't _just_ affect built-in resources, but also affects multi-versioned CRDs with a Conversion WebHook, e.g. cert-manager, as far as I know.

Specifically, the trigger is that if a chart is installed with a resource on one version of an API, e.g. `certmanager.k8s.io/v1alpha1/Certficate`, and then an upgrade of the chart has changed the API version, e.g. `certmanager.k8s.io/v1/Certificate`, then Helm (correctly) sees that as a _delete old GVK resource_ and _create new GVK resource_, as it has no data (currently) that there's a relationship between them. It then fails because when it talks to the API server, the same resource is visible at both versions, and so the create (first) fails. Had the delete been done first, it would have succeeded, albeit sub-optimally, as any local changes to the resource would be lost, where a strategic merge would have succeeded.

helm-mapkubeapis turns this delete+create into a strategic merge, but this should only be done for resources which have changed API version _when comparing the installed release and to-be-installed manifest_. Even applying it blindly to all known-deprecated resources fails if the chart author has, for example, fixed their old Deployment resources, but overlooked an Ingress object was removed in a later k8s release and so wasn't fixed at the time.

This can also happen with non-deprecated resources, e.g., a chart with a `autoscaling/v1/HorizontalPodAutoscaler` releases a new version where that's been changed into an `autoscaling/v2beta2/HorizontalPodAutoscaler` to take advantage of non-CPU-based scaling. `autoscaling/v1/HorizontalPodAutoscaler` is not deprecated, but it's the [same resource by _either_ GVK](https://cloud.google.com/kubernetes-engine/docs/how-to/horizontal-pod-autoscaling#viewing), so the same problem exists.

Which effectively means the mapping file for helm-mapkubeapis depends on the diff between the two chart versions, and lining up the GVK changes correctly. Which is why I mentioned the `helm diff` plugin, because it can show you the changes between on-cluster chart release, and the proposed upgrade, and any unexpected create+delete can be added to the appropriate map file.

This is pretty advanced behaviour for the target Helm user though, which I assume is why helm-mapkubeapis bundles a map file for the k8s APIs known to be deprecated/removed.

I feel like that might be a mistake, as it can lead to situations like this, where blindly running the mapping after a k8s upgrade will make it impossible to upgrade if the chart has not yet itself changed the api versions in its source.

---

Now I've written that, it's possible the adoption logic has made this _worse_, because now when trying create+delete, the create by the new API Version will succeed if it was installed with recent-enough Helm (as the necessary metadata is added by default), and _then_ it will delete the resource by its old API Version, leaving the resource *missing*.

That should be trivial to put a test-case together for, as it doesn't require changing anything on the cluster, merely creating with a deprecated API Version, and the upgrading to a more-recent APIVersion.

That said, this is _way_ off topic for the MR. If I find the time to put together a test-case for the above situations and the problem pans out like I'm guessing, I'll open a new issue or two for it.

hickeyma:
Thanks for the details feedback @TBBle. I would like to try and improve this if we can. I have some feedback below.
It might be worth opening an issue against the plugin to handle this? 

> I think the plugin is advertised to fix the wrong problem, and that it's because the problem has been mis-described in the Helm docs, and points people at the plugin as part of the "what if APIs have been removed by k8s version upgrades?" docs. So I think the question is more complex than 'removed' vs 'deprecated'.

The docs mention the plugin as follows: 

_"The manifest is a property of the Helm release object which is stored in the data field of a Secret (default) or ConfigMap in the cluster. The data field contains a gzipped object which is base 64 encoded (there is an additional base 64 encoding for a Secret). There is a Secret/ConfigMap per release version/revision in the namespace of the release._

_You can use the Helm mapkubeapis plugin to perform the update of a release to supported APIs. Check out the readme for more details."_

Can you describe what is misleading in this?

> The problem is described as an issue of removal of deprecated APIs from kubernetes, but that's not the issue, that's the underlying trigger for chart changes that cause the problem.

This is the concrete issue that breaks Helm when trying to do an upgrade of a release which contains API versions (from previous deployment) which are now removed from the Kubernetes version. 

> helm-mapkubeapis turns this delete+create into a strategic merge, but this should only be done for resources which have changed API version when comparing the installed release and to-be-installed manifest. Even applying it blindly to all known-deprecated resources fails if the chart author has, for example, fixed their old Deployment resources, but overlooked an Ingress object was removed in a later k8s release and so wasn't fixed at the time.

> This can also happen with non-deprecated resources, e.g., a chart with a autoscaling/v1/HorizontalPodAutoscaler releases a new version where that's been changed into an autoscaling/v2beta2/HorizontalPodAutoscaler to take advantage of non-CPU-based scaling. autoscaling/v1/HorizontalPodAutoscaler is not deprecated, but it's the same resource by either GVK, so the same problem exists.

> Which effectively means the mapping file for helm-mapkubeapis depends on the diff between the two chart versions, and lining up the GVK changes correctly. Which is why I mentioned the helm diff plugin, because it can show you the changes between on-cluster chart release, and the proposed upgrade, and any unexpected create+delete can be added to the appropriate map file.

The plugin works in a very simple non opinionated way. It will just update a release metadata with the mappings defined in the mapping file. This needs to be done for a release which contains API versions which are removed in a Kubernetes version. Otherwise, you will not be able to upgrade the release anymore. An example of this is if you had a deployment object in Kubernetes 1.15 with API: `apiVersion: apps/v1beta1`. If you then upgrade the cluster to 1.16, this API version is now removed. In the cluster upgrade, Kubernetes will upgrade the deployment object to supported version: `apiVersion: v1` but the version in the Helm release metadata stays in the unsupported version. There will be an error then when trying to do a Helm upgrade. The plugin updates this release metadata.

From what you describe, I think the issue with the plugin is that the update of the API versions in the metadata should only be once that version is removed and not while still available but in a deprecated state.

TBBle:
I think I'm still not being clear.

> In the cluster upgrade, Kubernetes will upgrade the deployment object to supported version: apiVersion: v1

That's not what actually happens, as far as I understand. The Kubernetes API server was _already_ offering the same object at both `apiVersion v1beta1` and `apiVersion v1` before the cluster upgrade. In fact, per https://github.com/kubernetes/kubernetes/issues/52185, even after the upgrade the old apiversion is what is stored until the resource is next written; [auto storage migration](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/storage-migration-auto-trigger.md) is not yet implemented as [the migrator is not production-ready](https://github.com/kubernetes-sigs/kube-storage-version-migrator/releases). It's just that the old API version is no longer visible to API callers.

(This is also one of the core issues identified in the CRD discussions for Helm, as the same thing applies there.)

After the cluster upgrade, the `apiVersion v1beta1` is not longer available, so the _same_ Helm chart version cannot be used, so the user grabs an updated version, which contains the same resource but with `apiVersion v1`, and we have the delete+create issue.

However, that exact same chart upgrade could be done on a cluster that still supports both APIs, e.g. the new one may be beta like [HorizontalPodAutoscaler](https://cloud.google.com/kubernetes-engine/docs/how-to/horizontal-pod-autoscaling#viewing), and the same problem would happen.

All the API removal on cluster upgrade does is _force_ the user to get a new chart version, because they can't reinstall the existing chart version anymore.

And as seen here, that is leading users (or at least one...) to see it as "After upgrading the cluster, run helm-mapkuebapis on your charts", which is the wrong result.

Or at least, that's my reading of the problem and the docs. As mentioned, I haven't produced a simple repro case yet, but I hope to soon, as I suspect it will also reveal a new data-loss case from the new adoption support.

shubhamc183:
**How to import multi-namespaced and non-namespaced resources into Helm?** 

I have a helm chart that creates various `namespace`, `secret`, `configmap`, `clusterrole`, `clusterrolebinding`.

I have **stupidly** applied it as in various k8s clusters.
```
cat <<EOF | kubectl apply --context $context -f -
$(helm template rbac-chart -f $values)
EOF
```

Now, I want to import these resources inside the helm but I am not sure what value to keep for `meta.helm.sh/release-namespace=$NAMESPACE` as some of the resources are
- namespace
- secret per namespace
- role per namespace
- **Non-namespaced objects** too[which doesn't belong to any namespace]
  - clusterrole
  - clusterrolebinding

@jlegrone

TBBle:
`meta.helm.sh/release-namespace` should be the namespace the _chart_ is deployed into, i.e the place Helm will put its own Secrets to track release status. This is separate from the namespace of the resources created by the chart. If you pass `--namespace` to `helm install` etc, that's the release namespace. Otherwise it's the namespace that's active when you run `helm`.

A resource without a `meta.namespace` will be created in the namespace of the release, so for single-namespace charts, it's a good practice to not use `meta.namespace` in the resources, so the chart is portable. You can also reference the releases's namespace with `{{ .Release.Namespace }}` in your templates.

In your case, your resources have explicit namespaces rather than using `{{ .Release.Namespace}}`, so you can install the chart into whatever namespace you like.

If your chart _does_ reference `{{ .Release.Namespace}}`, then you'll need to work out what value `helm template` used for that (it could be a dummy value?) and use that as the release namespace.

TBBle:
The labels/annotations only function during a `helm install` or `helm upgrade`, because they simply change the behaviour of "Thing you are trying to create already exists" from "fail" to "treat it as an upgrade" during `helm install` or `helm upgrade`.

If you have an existing chart deployment, and you have secrets that you want to become Helm-managed, then that implies you're trying to upgrade from a version or configuration of the chart that did not contain those secrets, to a version or configuration that does contain those secrets. That's what this feature is for.

I don't use helmfile, so I don't know what it's doing underneath, but if it's just driving `helm upgrade`, then it should "just work".

If it's trying to do more, particularly inspecting existing deployed resources, then it will have to implement a similar adoption logic to recognise that if it tries to create a new resource, and finds that it already exists and has the right labels/annotations, it should should that object as if it was part of the previous release with the current content, and the existing resource will be replaced with the one generated by the chart (since a 3-way merge between X and Y applying to X will always produce Y).

mark007:
Thanks for the reply. I have actually narrowed this down to what seems to be a bug in helm itself. Another user mentioned it here. I don't know how this feature can be usable with this particular issue.

https://github.com/helm/helm/issues/6561#issuecomment-687382740

Does anyone know if this is already written as a bug or would be accepted as one.

Basically if we have a very tiny chart with a tiny object like a service with eg

labels:
  keya: valuea

And we install and then annotate that object to be part of another helm release via the appropriate annotations and labels.

If we then try to upgrade the new release, the helm upgrade of course will not fail due to the annotations. If during that upgrade the object has slightly changed to

labels:
  keyb: valueb

The resulting object in the cluster after the install, instead of having only keyb: valueb only which is the intention, it has both labels.

labels:
  keya: valuea
  keyb: valueb

Its like that 3 way merge logic is not behaving very well in this scenario. No yaml file in the test has both labels. The pre-upgrade yaml file has keya, the target / upgraded yaml file has keyb, but the result of the upgrade is both keys. Do you think this is worth writing a bug on?

TBBle:
The comments you linked to sound like a bug (or at least are surprising to me) if the outcome of the `helm upgrade` with adoption is the file _as it was_ instead of either _as it is in the upgraded manifest_ or _some merge of the two_. But that's not what you're seeing, so I think you're seeing something different.

Your case sounds like a reasonable 3-way merge behaviour, *if* I understand your example correctly... You have an object with
```
labels:
 keya: valuea
```

and you have annotated it for adoption into a release, and applied an upgrade to that Helm release where the object _was created_ during the upgrade as
```
labels:
  keyb: valueb
```
then it's reasonable (although I didn't realise it worked this way) that a key-by-key merge will see the "created" step as "add `labels.keyb: valueb`", and instead of creating a new resource with that value, it's applied that change to the existing, ending up with both labels. That implies that I had misunderstood the 3-way merge done here, and in fact the adoption does a merge between 0 and Y applying to X, which would produce X+Y.

That said, I'm _really_ unsure I understood your example, perhaps some command-lines would make that clearer, and also allow reprodution.

And besides all that, since this is a merged PR: Yes, a new Issue is probably a better place to raise this, so it's more-visible.

mark007:
Thanks. I have created it as an issue here. https://github.com/helm/helm/issues/9958
For me, having contents of both an old and a new manifest, in the resulting manifest is definitely not good and not expected. Its leading to objects failing to upgrade as they contain details from both their old manifest AND their new manifest, not just the new one.

Arsen-Uulu:
@bacongobbler @jlegrone 

This works only for Helm not to complain about existence of a resources, but if you try to update that resource with helm - it doesn't work. Resource is not getting updated.

Shiva19908:
Hi Guys,
Does helm adopt resources created by a helm hook, for example, by a helm pre-upgrade hook? We are getting below error during the upgrade:
`upgrade.go:434: [debug] warning: Upgrade "abc" failed: no Service with the name "abc-headless" found
Error: UPGRADE FAILED: no Service with the name "abc-headless" found
helm.go:84: [debug] no Service with the name "abc-headless" found
helm.sh/helm/v3/pkg/kube.(*Client).Update.func1
        helm.sh/helm/v3/pkg/kube/client.go:257
helm.sh/helm/v3/pkg/kube.ResourceList.Visit
        helm.sh/helm/v3/pkg/kube/resource.go:32
helm.sh/helm/v3/pkg/kube.(*Client).Update
        helm.sh/helm/v3/pkg/kube/client.go:230
helm.sh/helm/v3/pkg/action.(*Upgrade).releasingUpgrade
        helm.sh/helm/v3/pkg/action/upgrade.go:376
runtime.goexit
        runtime/asm_amd64.s:1571
UPGRADE FAILED
main.newUpgradeCmd.func2
        helm.sh/helm/v3/cmd/helm/upgrade.go:201
        github.com/spf13/cobra@v1.5.0/command.go:872
        github.com/spf13/cobra@v1.5.0/command.go:990
        github.com/spf13/cobra@v1.5.0/command.go:918
main.main
        helm.sh/helm/v3/cmd/helm/helm.go:83
runtime.main
        runtime/proc.go:250
runtime.goexit
        runtime/asm_amd64.s:1571` 

What we are doing is:
 We are creating a K8S service in the pre-upgrade hook and adding relavant labels and annotations so that Helm would adopt this service. But this is not working as expected. However, if we create the same resource before the upgrade, Helm adopts the resource and works as expected. Furthermore, if we run helm upgrade the second time it also passes! Hence we would like to know if Helm supports adopting of resources creating during the upgrade by a hook?

TBBle:
I'd suggest opening that as a new ticket, with more details, e.g., the chart you're trying to deploy or a minimal replication of the problem. That way the issue/question is likely to get more attention from people who know whether it's a bug, or if it's working-as-intended, e.g., what you're trying to do is not supported.

lkoniecz:
Does it work other way around?
I mean if I remove those 2 annotations and the label will the resource be unpinned from the chart?

PR #7655, opened 2020-02-21
fix: add new static linter and fix issues it found

technosophos:
This PR adds a new static analyzer to the testing target.

In the course of adding, the analyzer found two cases where code was incorrect. While neither case imposes any risk, I have fixed them both.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -17,6 +17,7 @@ linters:
     - structcheck
     - unused
     - varcheck
+    - staticcheck
 
 linters-settings:
   gofmt:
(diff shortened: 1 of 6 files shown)


jdolitsky (changes requested):
@technosophos looks like things are failing in circle - maybe related to the version of linter or versions being used?

technosophos:
Okay, updated for the new checks done by the new static analyzer. Weeee!

jdolitsky (approved):

adamreese (approved):

PR #4106, opened 2018-05-22
Download charts to a temporary file

amir:
Similar to https://github.com/kubernetes/helm/issues/3253 we're using
Helm by directly importing the Go package.

A concurrency issue arises when multiple requests come for the same
package. Currently DownloadTo truncates contents of destination file and
rewrites it.

This patch changes the behaviour so instead of directly writing into the
destination it'll write the content in a temp file and renames it once
download is finished.

This change fixes sporadic issues in
https://github.com/amir/skeg/blob/5c5c6a1c215cf5110843a28f1fa389aa69a3f452/main_test.go

PR #7659, opened 2020-02-22
Use locks to support downloading archives in parallel

yurrriq:
Address https://github.com/helm/helm/issues/6140, and supersede https://github.com/helm/helm/pull/4106, by adding lock files to the chart downloader.

I'm not sure how best to test this, but want to be sure to, so I'd love some feedback.

### DIFF
--- a/pkg/downloader/chart_downloader.go
+++ b/pkg/downloader/chart_downloader.go
@@ -136,29 +136,45 @@ func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *proven
 		return "", nil, err
 	}
 
-	data, err := g.Get(u.String(), c.Options...)
-	if err != nil {
-		return "", nil, err
-	}
+	if fi, err := os.Stat(destfile); os.IsNotExist(err) || fi.Size() == 0 {
+		data, err := g.Get(u.String(), c.Options...)
+		if err != nil {
+			return "", nil, err
+		}
 
-	if err := atomicWriteFile(destfile, data, 0644); err != nil {
-		return destfile, nil, err
+		if err := atomicWriteFile(destfile, data, 0644); err != nil {
+			return destfile, nil, err
+		}
 	}
 
 	// If provenance is requested, verify it.
 	ver := &provenance.Verification{}
 	if c.Verify > VerifyNever {
-		body, err := g.Get(u.String() + ".prov")
-		if err != nil {
-			if c.Verify == VerifyAlways {
-				return destfile, ver, errors.Errorf("failed to fetch provenance %q", u.String()+".prov")
-			}
-			fmt.Fprintf(c.Out, "WARNING: Verification not found for %s: %s\n", ref, err)
-			return destfile, ver, nil
-		}
 		provfile := destfile + ".prov"
-		if err := atomicWriteFile(provfile, body, 0644); err != nil {
-			return destfile, nil, err
+
+		// Acquire a file lock for process synchronization
+		fileLock := flock.New(provfile)
+		lockCtx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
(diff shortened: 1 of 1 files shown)


aaronmell:
> /cc @aaronmell @karuppiah7890

Awesome!

yurrriq:
I've updated this to avoid re-downloading archives and provenance files. Also to lock the provenance file.

yurrriq:
Any feedback? I would love to see this or similar get merged. The concurrency issue is quite a problem for us.

karuppiah7890:
I need to read more to understand the changes. But, it looks like simple and clean changes 😄 Just one suggestion which I always request for - tests. Please do add tests for these. This is to make the software more robust and as much bug free as possible, especially when more features or tweaks are done to this code in the future.

karuppiah7890:
Also, these changes are done for Helm v3, once this is merged, do check if it's possible and easy to make the same changes in v2. Or I could do it. My team uses v2 even now 🙈 Migration will take quite some time 😅

yurrriq:
As I mentioned above, I'm not sure how best to test this and will need some guidance. No need to try to sell me on the merits of testing. :smile: I'm in total agreement.

karuppiah7890:
Ah, cool. My bad I didn't read properly! 🙈  I'll look at the code again and checkout scenarios and comment here what I think and how we can test this in a clean manner 😄

yurrriq:
Rebased.

luryus:
Any chance of getting this reviewed and merged? We constantly run into issues when running multiple Helm upgrades concurrently with helmfile.

bacongobbler:
A couple of quick comments for feedback:

- for testing code paths where two calls running concurrently can run into lock scenarios, consider using [`WaitGroups`](https://gobyexample.com/waitgroups)
- this PR is out of date, and needs another rebase

Given the amount of code being changed and/or modified here, I would not be comfortable merging this PR without writing a few tests to

a) ascertain the base case (#6140)
b) ensure that the fix provided here addresses that case

Thanks!

pkg/downloader/chart_downloader.go
```diff
@@ -114,36 +117,64 @@ func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *proven
 		return "", nil, err
 	}
 
-	g, err := c.Getters.ByScheme(u.Scheme)
+	name := filepath.Base(u.Path)
+	destfile := filepath.Join(dest, name)
+
+	// Acquire a file lock for process synchronization
+	fileLock := flock.New(destfile)
+	lockCtx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
bacongobbler, line 125: How come this is specificaly set as a 30 second timeout? What made you decide to land on that number as a timeout?

yurrriq: The 30 second timeout is arbitrary, copped from https://github.com/helm/helm/blob/v3.3.1/cmd/helm/repo_add.go#L93

yurrriq:
>     * for testing code paths where two calls running concurrently can run into lock scenarios, consider using [`WaitGroups`](https://gobyexample.com/waitgroups).

Thanks for the tip. I'll have a go sometime in the next few days. Holiday weekend here, so I'm optimistic.

>     * this PR is out of date, and needs another rebase

Yep, will do.

> Given the amount of code being changed and/or modified here, I would not be comfortable merging this PR without writing a few tests to
> 
> a) ascertain the base case (#6140), and ensure the test case fails prior to this patch
> b) ensure that the fix provided here addresses that case

I fully agree. I'll attempt to address both concerns.

yurrriq:
I rebased and wrote tests, and confirmed that #7720 addressed the issue since v3.2.0.

PR #7660, opened 2020-02-23
Add --lint and --strict options to helm template (this time without updating old tests)

### LINKED ISSUE #6506: Add --lint and --strict options to helm template

sco11morgan:
Resolves #6506

This is an updated version of https://github.com/helm/helm/pull/7060 that does not update old tests.


bacongobbler:
The case to be made for introducing `helm upgrade --install` made sense: users wanted to create the initial release if it didn't already exist.

The use case for `helm template --lint` provided in #6506 seems much less clear here. `helm template` does not needs to perform a `helm lint` to accomplish its task of templating the manifests. Validation occurs during the rendering process. Linting is performed while developing the chart.

#6506 just wants a way to share `--set` and `--values` flags with each command. Merging the two commands accomplishes that, but it introduces more design issues.

This PR just consolidates one command (`helm lint`) into the other, and it removes that clear separation of concerns each command has.

Given that assessment, I feel like we should not go forward with this PR. Perhaps a plugin would be better suited here.

Another approach might be to implement `helm config` where users can set these CLI flags ahead of time, similar to how git uses `git-config`. More info here: https://git-scm.com/docs/git-config/1.8.2

What do you think?

bacongobbler:
Removing from 3.2.0 as this PR needs more discussion before we consider merging

sco11morgan:
@bacongobbler The reason this PR is useful is that strict validation is not done during the rendering process. If a particular entry is missing from the values map, then at render time an empty string is inserted.  This missing value may or may not be picked up when the manifest is applied to Kubernetes.

```
$ helm template .
# Source: test-helm/templates/service.yaml
apiVersion: v1
kind: Service
metadata:
  name: RELEASE-NAME-test-helm
  labels:
    helm.sh/chart: test-helm-0.1.0
    app.kubernetes.io/name: test-helm
    app.kubernetes.io/instance: RELEASE-NAME
spec:
  type: ClusterIP
  ports:
    - port:
      targetPort: http
      protocol: TCP
      name: http
  selector:
    app.kubernetes.io/name: test-helm
    app.kubernetes.io/instance: RELEASE-NAME

$ echo $?
0

$ helm lint .
==> Linting .

$ helm lint . --strict
==> Linting .
[ERROR] templates/: template: test-helm/templates/service.yaml:10:22: executing "test-helm/templates/service.yaml" at <.Values.service.port>: map has no entry for key "port"

```


Values schema support was a nice addition to Helm 3 but it does not fit out use case.  We have a large number of apps sharing a common library subchart.  Not all of the apps need all of the components of the subchart so we can't have a common schema in the subchart.

PR #7661, opened 2020-02-23
fix(helm): add --skip-crds flag to 'helm upgrade'

### LINKED ISSUE #7452: upgrade --install should allow --skip-crds
It looks like when you use `helm upgrade --install` there is no way to pass `--skip-crds` and thus the CRDs are by default installed.

akash-gautam:
Fixes #7452


hickeyma (approved):
LGTM, thanks @akash-gautam  for this fix.

PR #7678, opened 2020-02-25
fix(cmd/helm): upgrade go-shellwords

adamreese:
Removes workaround introduced in #7323


marckhouzam (approved):
Thanks for doing this.
LGTM

PR #7690, opened 2020-02-27
Full completion support for Fish shell

marckhouzam:
This PR extends the **__complete** command (which currently provides custom completions) by making it handle all completions (command names and flag names as well as custom completions).

Using this, it becomes possible to provide completions for other shells relatively easily.  In this PR, full completion support is added for the Fish shell.

To try it:
```
brew install fish
rehash
fish
helm <TAB> # To auto-load old completion rules so helm can delete them
helm completion fish | source

helm stat<TAB>
helm status --name<TAB>
helm status --namespace <TAB>
```

**Note to maintainers**:

The new flag `--no-descriptions` is added to the `helm completion` command.  This allows the user to turn off completion descriptions for the fish shell (and later on for the zsh shell). For example:

```
fish> helm completion fish | source
fish> helm s<TAB>
search            (search for a keyword in charts) 
show                 (show information of a chart)  
status  (displays the status of the named release)

fish> helm completion --no-descriptions fish | source
fish> helm s<TAB>
search  show  status
```


marckhouzam:
The Fish shell supports adding a description for a completion choice.  I've added a commit which includes the description of the command or flag being completed.  For example:
```
fish> helm s<TAB>
search            (search for a keyword in charts) 
show                 (show information of a chart)  
status  (displays the status of the named release)
```
I've allowed the user to choose to disable this feature by adding a new flag to `helm completion`: `--no-descriptions`.  Adding this new flag should be approved by other maintainers as it becomes API.

marckhouzam:
Any maintainer have time for this one? It would be cool to get Fish completion in 3.2. A big part of the code change is actually just moving things to another function.

internal/completion/complete.go
```diff
@@ -255,23 +209,161 @@ func NewCompleteCmd(settings *cli.EnvSettings, out io.Writer) *cobra.Command {
...
+	}
+	return completions
+}
+
 func isFlag(arg string) bool {
 	return len(arg) > 0 && arg[0] == '-'
 }
 
+func nonCompletableFlag(flag *pflag.Flag) bool {
+	return flag.Hidden || len(flag.Deprecated) > 0
```
bacongobbler: Instead of a length check, can we check `flag.Deprecated` against the empty string here? It's a little clearer to the reader what we're trying to check for

bacongobbler (approved):
Love the refactor done here as well. Nice work!

I found one place where we could improve readability, but I have no qualms with the behaviour of the code itself.

LGTM!

bacongobbler:
In the future, if you'd like to get a PR reviewed for a particular milestone and it's ready to go, feel free to assign it to the milestone!

marckhouzam:
Thanks for the quick response! I'll remember about the milestone for next time.

I'll push a revision with your recommendation soon.

internal/completion/complete.go
marckhouzam: Done

marckhouzam:
I've addressed the review comment and merged master to make sure the tests still pass (just in case).

marckhouzam:
Hmm, plugin completion does not work for Fish.  I'll have to investigate.  Let's hold on merging this.

marckhouzam:
Ok, I realize that the logic for plugin static completion only works when using the traditional cobra completion script.  For Fish we only use the `__complete` command which I hadn't prepared the plugin completion for.

Now that the problem is understood and considering there was no Fish completion at all before, I think it is still worth merging this without the support for plugin completion in Fish.  I'll work on it though.

Ok?

marckhouzam:
The last commit fixes the problem with Fish completion and plugin static completion.
Everything is ready to merge.

bacongobbler:
@marckhouzam looks like CI is reporting some failures that seem related to this PR. Would you mind looking into that?

marckhouzam:
> @marckhouzam looks like CI is reporting some failures that seem related to this PR. Would you mind looking into that?

Sorry about that.  I mistakenly committed new tests I'm working on but which were not working yet.  I've removed them and will post separately if I figure out a way to fix them.

marckhouzam:
@bacongobbler Does this still look good to you?

marckhouzam:
> @bacongobbler Does this still look good to you?

BTW, the only change since your review is the small change to `load_plugins.go`.

marckhouzam:
Sigh...
I'm going to backtrack... again 😖 

While confirming that everything was working properly, I noticed that the fish completion for flags with an `=` (e.g., `helm --namespace=<TAB>`) doesn't work.  Fish behaves a little differently than bash and zsh for this case.  I will fix the script.

I'm sorry about this.  Unlike for bash, I don't have acceptance-tests that work with fish, so it makes it much more difficult to avoid broken corner-cases.  Luckily, the fish completion script should almost never need to be changed, once it works properly (and it will eventually be pushed down to Cobra, hopefully).

marckhouzam:
A little update on this one:

Fundamentally, the Fish completion is correct.  However, it has brought to the forefront that the plugin completion feature currently has a limitation.  

Completion for plugins for bash and zsh works well because it uses the completion script for static completion and uses the `helm __complete` command for dynamic completion.  This separation allows the two types of completion to work well together.  

Fish goes further and uses the `helm __complete` command for both static completion and dynamic completion, and this is where we run into trouble.  By having to provide knowledge of the static completions to the `helm __complete` command, we cause Cobra to parse plugin sub-commands and flags, and to no longer pass them to the `plugin.complete` file in charge of dynamic completion.

I'm still looking for a good solution to solve this, but no big inspiration yet.

In the mean time, I will adjust this PR to provide Fish completion without supporting plugin completion.  This will be an improvement for users of the Fish shell, while giving me time to figure out a way to properly handle plugin completion entirely from the `helm __complete` command.

Code update coming soon.

marckhouzam:
Ok, so this was a doozy but I feel like I got it right now.

- Fish completion is fully functional (manual testing only, but detailed).
- Plugin completion continues to work as expected and confirmed with the acceptance-testing regression tests.
- The go tests all pass, except the ones for dynamic completion which I had to comment out for now.  It is not the feature that is broken, but the test infrastructure that fails in this particular scenario.  I will look into it, but I don't think it should block this feature from the 3.2 release.

**For full transparency**: I had to move things around in `cmd/helm/load_plugins.go`.  However, I was careful not to make changes to the real plugin functionality, but only to the plugin completion stuff.

@bacongobbler Are you comfortable in putting this into 3.2?

marckhouzam:
I took the time to modify the acceptance-testing repo to be able to test the new fish completion feature.
https://github.com/helm/acceptance-testing/pull/77
This will gives us much more confidence for maintaining this code.

But by doing this I took the opportunity to simplify the fish completion script of this PR.
The tests also revealed one bug, which was that I forgot to support completion for a binary using a different name e.g. `helm3`.  The latest commit fixes that.

Ready for review to hopefully merge for 3.2.

marckhouzam:
After working with Cobra, I am getting concerned that the feature of disabling descriptions in Fish completion may not get accepted.  I have not indication of it, but if Cobra chooses to always keep descriptions enabled, Helm would get stuck supporting this feature without being able to use Cobra for Fish completion.

Therefore, I will pull out that option to avoid introducing the new `--no-descriptions` flag, until we know Cobra will support the feature.

cmd/helm/completion.go
```diff
@@ -52,13 +56,14 @@ func newCompletionCmd(out io.Writer) *cobra.Command {
...
 		Use:   "completion SHELL",
-		Short: "Generate autocompletions script for the specified shell (bash or zsh)",
+		Short: "Generate autocompletions script for the specified shell (bash, zsh or fish)",
 		Long:  completionDesc,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return runCompletion(out, cmd, args)
 		},
 		ValidArgs: shells,
 	}
+	cmd.PersistentFlags().BoolVar(&completionNoDesc, "no-descriptions", false, "disable completion description for shells that support it")
```
marckhouzam, line 66: Don't introduce this flag yet, until we know that Cobra will support disabling descriptions.  Or else we'll be stuck supporting this flag without being able to use Cobra for Fish completion.

marckhouzam:
Things are going well with moving all this completion code to the Cobra project.  Based on this comment https://github.com/spf13/cobra/pull/1035#issuecomment-608537029 I expect Fish completion to be supported soon.

Therefore let's not have anyone review the current PR as most of it will be provided by Cobra pretty soon. I'll post a replacement PR using Cobra once it is possible.

weitzj:
Just for reference, if somebody stumbles along:

- Fish shell completion support is merged into Cobra: https://github.com/spf13/cobra/pull/1048

If you want to do **process substitution** as you know from Bash, you have to use the `psub` syntax for fish. 

See: https://github.com/fish-shell/fish-shell/issues/1786#issuecomment-587251893


i.e.

```
# bash
kubectl apply -f <(istioctl kube-inject -f samples/httpbin/httpbin.yaml)

# fish equivalent
kubectl apply -f (istioctl kube-inject -f samples/httpbin/httpbin.yaml | psub)
```

PR #7691, opened 2020-02-27
Add unit test for lint/values.go

longkb:
Signed-off-by: Kim Bao Long <longkb@vn.fujitsu.com>

### DIFF
--- a/pkg/lint/rules/values_test.go
+++ b/pkg/lint/rules/values_test.go
@@ -1,0 +1,37 @@
+/*
+Copyright The Helm Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package rules
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+var (
+	nonExistingValuesFilePath = filepath.Join("/fake/dir", "values.yaml")
+)
+
+func TestValidateValuesYamlNotDirectory(t *testing.T) {
+	_ = os.Mkdir(nonExistingValuesFilePath, os.ModePerm)
+	defer os.Remove(nonExistingValuesFilePath)
+
+	err := validateValuesFileExistence(nonExistingValuesFilePath)
+	if err == nil {
+		t.Errorf("validateValuesFileExistence to return a linter error, got no error")
+	}
+}


jdolitsky (approved):
more tests always good!

hickeyma (changes requested):
@longkb Thanks for pushing the tests. `validateValuesFileExistence` and `validateValuesFile` are being tested quite extensively in https://github.com/helm/helm/blob/master/pkg/lint/lint_test.go where the [All function](https://github.com/helm/helm/blob/master/pkg/lint/lint.go#L27) is being called.

I am therefore unsure what value add is gained by adding these tests. What do you think yourself?

hickeyma:
Ping @longkb

longkb:
> 
> 
> @longkb Thanks for pushing the tests. `validateValuesFileExistence` and `validateValuesFile` are being tested quite extensively in https://github.com/helm/helm/blob/master/pkg/lint/lint_test.go where the [All function](https://github.com/helm/helm/blob/master/pkg/lint/lint.go#L27) is being called.
> 
> I am therefore unsure what value add is gained by adding these tests. What do you think yourself?

Hi @hickeyma ,
I think I was missed out `https://github.com/helm/helm/blob/master/pkg/lint/lint_test.go`. 
I have just checked the test files and found out that `values.go` are missing the test file. So I would like to write the UT for this `values.go` file :)

hickeyma:
@longkb Can you explain what you test that are missing and update the PR accordingly then?

mattfarina:
@hickeyma I'll provide 2 cases for this...

First, for what ever reason the `values.go` functions are being listed as untested... see...

```
helm.sh/helm/v3/pkg/lint/rules/values.go:31:                            Values                          0.0%
helm.sh/helm/v3/pkg/lint/rules/values.go:43:                            validateValuesFileExistence     0.0%
helm.sh/helm/v3/pkg/lint/rules/values.go:51:                            validateValuesFile              0.0%
```

This is bad for our test coverage metrics.

Second, this test does cover something new. First, the lack of existence of a `values.yaml` file. The current test does not check for that.

I'm also curious why the current bad values test has an error with `"cannot unmarshal"` in it (terrible language for end users) does not have the error from the function (`"unable to parse YAML"`).

Thoughts?

hickeyma:
@mattfarina Some feedback to your comments.

> First, for what ever reason the values.go functions are being listed as untested... see..

This is strange, I agree that this should not being happening. 

When I test locally it seems to call the tests:
```console
$ make test
[...]
=== RUN   TestBadChart
--- PASS: TestBadChart (0.00s)
=== RUN   TestInvalidYaml
--- PASS: TestInvalidYaml (0.01s)
=== RUN   TestBadValues
--- PASS: TestBadValues (0.00s)
=== RUN   TestGoodChart
--- PASS: TestGoodChart (0.01s)
PASS
ok      helm.sh/helm/v3/pkg/lint        1.081s
[....]
```

From looking at the code, it seems that the following tests call [All](https://github.com/helm/helm/blob/master/pkg/lint/lint.go#L27) function:
- [TestBadChart](https://github.com/helm/helm/blob/master/pkg/lint/lint_test.go#L36)
- [TestInvalidYaml](https://github.com/helm/helm/blob/master/pkg/lint/lint_test.go#L81)
- [TestBadValues](https://github.com/helm/helm/blob/master/pkg/lint/lint_test.go#L91)
- [TestGoodChart](https://github.com/helm/helm/blob/master/pkg/lint/lint_test.go#L101)

The [All](https://github.com/helm/helm/blob/master/pkg/lint/lint.go#L27) function calls the [values rules](https://github.com/helm/helm/blob/master/pkg/lint/lint.go#L33).  This should then call [validateValuesFileExistence](https://github.com/helm/helm/blob/master/pkg/lint/rules/values.go#L43) and [validateValuesFile](https://github.com/helm/helm/blob/master/pkg/lint/rules/values.go#L51).vim out

> Second, this test does cover something new. First, the lack of existence of a values.yaml file. The current test does not check for that.

It look like it should [here](https://github.com/helm/helm/blob/master/pkg/lint/rules/values.go#L34)

> I'm also curious why the current bad values test has an error with "cannot unmarshal" in it (terrible language for end users) does not have the error from the function ("unable to parse YAML").

Not sure why. This can be cleaned up in a separate PR.

pkg/lint/rules/values_test.go
```diff
@@ -0,0 +1,49 @@
...
+		t.Errorf("validateValuesFileExistence to return a linter error, got no error")
+	}
+}
+
+func TestValidateValuesName(t *testing.T) {
+	err := validateValuesFile(badValuesFilePath)
+	if err == nil {
+		t.Errorf("validateValuesFile to return a linter error, got no error")
+	}
+}
```
hickeyma: This is tested already here: https://github.com/helm/helm/blob/master/pkg/lint/lint_test.go#L91

pkg/lint/rules/values_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+func TestValidateValuesYamlNotDirectory(t *testing.T) {
+	_ = os.Mkdir(nonExistingValuesFilePath, os.ModePerm)
+	defer os.Remove(nonExistingValuesFilePath)
+
+	err := validateValuesFileExistence(nonExistingValuesFilePath)
+	if err == nil {
+		t.Errorf("validateValuesFileExistence to return a linter error, got no error")
+	}
+}
```
hickeyma: This doesn't seem to be tested.  This would be a good test.

hickeyma (changes requested):
Looking again at the tests, I think one of the test cases are not already being tested.  Some feedback inline.

pkg/lint/rules/values_test.go
longkb: I am sorry for interruption. I have removed this test :)

hickeyma (approved):
LGTM, thanks for the test @longkb

hickeyma:
> I'm also curious why the current bad values test has an error with "cannot unmarshal" in it (terrible language for end users) does not have the error from the function ("unable to parse YAML").

@mattfarina I pushed #7779 to improve this

PR #7692, opened 2020-02-27
Update README.md

eyablokov:
Typo fix: Space missed in Markdown header.


hickeyma (approved):
LGTMN, thanks for the fix @eyablokov

PR #7695, opened 2020-02-27
Install golangci-lint if needed

longkb:
Signed-off-by: Kim Bao Long <longkb@vn.fujitsu.com>

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,7 +15,7 @@ jobs:
       - checkout
       - run:
           name: install test dependencies
-          command: .circleci/bootstrap.sh
+          command: make lint
       - run:
           name: test style
           command: make test-style
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,8 @@ GOX           = $(GOPATH)/bin/gox
 GOIMPORTS     = $(GOPATH)/bin/goimports
 ARCH          = $(shell uname -p)
 
+GOLANGCI_LINT_VERSION = "1.21.0"
+
 ACCEPTANCE_DIR:=../acceptance-testing
 # To specify the subset of acceptance tests to run. '.' means all tests
 ACCEPTANCE_RUN_TESTS=.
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -126,6 +126,9 @@ format: $(GOIMPORTS)
 $(GOX):
 	(cd /; GO111MODULE=on go get -u github.com/mitchellh/gox)
 
+$(GOLANGCI_LINT):
+	(cd /; GO111MODULE=on GOLANGCI_LINT_VERSION="1.21.0" ./.circleci/bootstrap.sh)
```
jdolitsky: Would it be possible to simply do the "go get" here? I don't think we should be calling out to scripts specific to circleci (i do realize that it is installing golangcilint)

jdolitsky (changes requested):

Makefile
longkb: > 
> 
> Would it be possible to simply do the "go get" here? I don't think we should be calling out to scripts specific to circleci (i do realize that it is installing golangcilint)

It is not recommended to install via `go get` in the home page: https://github.com/golangci/golangci-lint#go

jdolitsky: @longkb I think that would probably be best, then in circle bootstrap we can call the new make target

longkb: Done. I did it :) Please give me your comments if needed

.circleci/config.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       - checkout
       - run:
           name: install test dependencies
-          command: .circleci/bootstrap.sh
```
mattfarina: Why wasn't the bootstrap file removed if it's no longer used? Can we remove it.

Makefile
```diff
@@ -89,10 +91,19 @@ test-coverage:
...
 	@scripts/validate-license.sh
 
+.PHONY: lint
+lint:
+	set -euo pipefail
+	cd /
+	GO111MODULE=on
+	curl -sSL https://github.com/golangci/golangci-lint/releases/download/v${GOLANGCI_LINT_VERSION}/golangci-lint-${GOLANGCI_LINT_VERSION}-linux-amd64.tar.gz | tar xz
+	sudo mv golangci-lint-${GOLANGCI_LINT_VERSION}-linux-amd64/golangci-lint /usr/local/bin/golangci-lint
+	rm -rf golangci-lint-${GOLANGCI_LINT_VERSION}-linux-amd64
```
mattfarina: Might I suggest an alternative method to install it. See https://github.com/Masterminds/semver/blob/master/Makefile where it doesn't use a `.PHONY` target but detects the presence and installs if missing. If present it does not install.

In CircleCI it made sense to install to `/usr/local/bin` but I wouldn't want to do that on every developers system. Can we instead use the GOPATH/bin instead.

mattfarina (changes requested):
Thanks for the PR. I have a cleanup and make change suggestion to improve the native use of make's abilities.

bacongobbler:
ping. Checking in to see what's the status here. Last activity was 6 months ago

bacongobbler:
closing as inactive.

PR #7698, opened 2020-02-27
Make GetCapabilities() method public

tamalsaha:
Signed-off-by: Tamal Saha <tamal@appscode.com>

### DIFF
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -95,7 +95,7 @@ type RESTClientGetter interface {
 type DebugLog func(format string, v ...interface{})
 
 // capabilities builds a Capabilities from discovery information.
-func (c *Configuration) getCapabilities() (*chartutil.Capabilities, error) {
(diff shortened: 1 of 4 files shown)


tamalsaha:
I am using helm as a library to render charts on my server side code. It would helpful, if this method can be made public.

jdolitsky (approved):
this seems straightforward - anyone opposed?

bacongobbler:
Will take a closer look next week. Generally speaking functions like getCapabilities are made private because the underlying behaviour can change between kubernetes client releases.

mattfarina:
@tamalsaha Can you share your use case for using this method? It's part of the actions package so I am wondering if you're not able to use the actions or your building something different. Can you help us understand more clearly?

tamalsaha:
We have wrote a `check permissions` "action" that will check if the user / client getter has the necessary permissions to install/uninstall/update charts using SubjectReview api. We need access to the capabilities to call
 
`chartutil.ToRenderValues(chrt.Chart, vals, options, caps)`

api.

Issue #7701, opened 2020-02-28
Reference template parsing order is non-deterministic


asenci:
Reference template parsing order is non-deterministic, causing tpl function to render the wrong value in certain situations.

Output of `helm version`:
```
version.BuildInfo{Version:"v3.1.1", GitCommit:"afe70585407b420d0097d07b21c47dc511525ac8", GitTreeState:"clean", GoVersion:"go1.13.8"}
```

Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.5", GitCommit:"20c265fef0741dd71a66480e35bd69f18351daea", GitTreeState:"clean", BuildDate:"2019-10-15T19:16:51Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.2+k3s1", GitCommit:"cdab19b09a84389ffbf57bebd33871c60b1d6b28", GitTreeState:"clean", BuildDate:"2020-01-27T18:09:26Z", GoVersion:"go1.13.6", Compiler:"gc", Platform:"linux/amd64"}
```

Cloud Provider/Platform (AKS, GKE, Minikube etc.): 
```
k3s
```

Minimal code to reproduce the issue:

[chart.zip](https://github.com/helm/helm/files/4264859/chart.zip)

```
$ unzip chart.zip
$ for i in $(seq 0 99); do helm template test01 parent >> results.txt; done
$ grep 'childName: test01-child' results.txt | wc -l
$ grep 'childName: test01-parent' results.txt | wc -l
```

I am seeing an average split of 80/20% between the values.

PR to come.

asenci:
Fixed by PR #7702

PR #7702, opened 2020-02-28
Parse reference templates in predictable order

asenci:
Fix issue #7701


jdolitsky:
@asenci this change looks generally straightforward. Could we just add a test to pkg/engine/engine_test.go to ensure this bug doesnt get re-introduced?

jdolitsky (changes requested):
see comment above

asenci:
> see comment above

Hope this is ok, not sure how to reliably reproduce the issue due to the randomness in the map iteration, so I am repeating the test 100 times hopping to hit the wrong order at least once.

It's not pretty, but from my observations 100 iterations should be more than enough to guarantee at least one hit, and shouldn't slow down the tests too much.

jdolitsky (approved):
👍

technosophos (approved):
LGTM

PR #7714, opened 2020-03-02
fix(helm): respect resource policy on ungrade

### LINKED ISSUE #7677: Annotation "helm.sh/resource-policy: keep" is ignored on helm upgrade
I use the annotation "helm.sh/resource-policy": keep to protect some PVCs from deletion on helm uninstall, which works fine. Alas, on helm upgrade, the PVCs are deleted, taking my data with them.
Observed on Helm 3.0.3 and Helm 3.1.1.

This seems to be a regression, as I found this issue:
https://github.com/helm/helm/issues/3673

donggangcj:
Don't delete a resource on upgrade if it is annotated with
helm.io/resource-policy=keep. This can cause data loss for users
if the annotation is ignored(e.g. for a PVC)

Close #7677

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"k8s.io/apimachinery/pkg/api/meta"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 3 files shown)


bacongobbler (approved):

turkenh:
What is the ETA for a helm3 release including this fix?

jgoeres:
Same question here, we need this ASAP, too.

midestefanis:
Same question here, we need this ASAP, too.

midestefanis:
I was having the same problem with helm v3.1.2. Please I need this solution. Thank you

bacongobbler:
For the time being: https://helm.sh/docs/intro/install/#from-canary-builds

PR #7721, opened 2020-03-02
ref(environment): use string checking instead

bacongobbler:
It is more idiomatic to compare the string against the empty string than to check the string's length.


marckhouzam (approved):
LGTM

I've been flip flopping between the two approaches for my PRs and I appreciate this clear recommendation.

PR #7727, opened 2020-03-04
fix(helm): add BackendConfig to sort order

bergemalm:
If "custom" crd's are allowed to be added here? - Else I'll close this PR.

GKE's BackendConfig needs to be installed before ingress as
part of service configuration (hence before service).
Could save 5+ min on bootstrap if BackendConfig configurations are used.

Signed-off-by: Mikael Bergemalm <mbergemalm@gmail.com>


jdolitsky:
@bergemalm thanks for opening, but as you mentioned, support for CRDs in this sort order wont be accepted since it opens the floodgates for all custom resources to be added by all vendors.

Going to close, but feel free to open an issue that requests some feature to solve your problem in a different way and any suggestion you might have.

PR #7743, opened 2020-03-06
Fix nested null value overrides

### LINKED ISSUE #5184: Nested `null` values don't remove keys as expected

uzxmx:
Closes #5184

I tested with the latest release v3.1.1, but with no luck, the [deleting a default key](https://helm.sh/docs/chart_template_guide/values_files/#deleting-a-default-key) still didn't work.  I found the merged PR #5185 solved that problem, but it only lives on dev-v2 branch, seems it hasn't been picked in v3.x.

So I created this PR. Hopefully it will help others who has met such problem as me.

### DIFF
--- a/pkg/chartutil/coalesce.go
+++ b/pkg/chartutil/coalesce.go
@@ -47,10 +47,7 @@ func CoalesceValues(chrt *chart.Chart, vals map[string]interface{}) (Values, err
 	if valsCopy == nil {
 		valsCopy = make(map[string]interface{})
 	}
-	if _, err := coalesce(chrt, valsCopy); err != nil {
-		return valsCopy, err
-	}
-	return coalesceDeps(chrt, valsCopy)
+	return coalesce(chrt, valsCopy)
 }
 
 // coalesce coalesces the dest values and the chart values, giving priority to the dest values.
--- a/pkg/chartutil/coalesce.go
+++ b/pkg/chartutil/coalesce.go
@@ -186,19 +186,18 @@ func CoalesceTables(dst, src map[string]interface{}) map[string]interface{} {
 	// Because dest has higher precedence than src, dest values override src
 	// values.
 	for key, val := range src {
-		if istable(val) {
-			switch innerdst, ok := dst[key]; {
-			case !ok:
-				dst[key] = val
-			case istable(innerdst):
-				CoalesceTables(innerdst.(map[string]interface{}), val.(map[string]interface{}))
-			default:
+		if dv, ok := dst[key]; ok && dv == nil {
+			delete(dst, key)
+		} else if !ok {
+			dst[key] = val
+		} else if istable(val) {
+			if istable(dv) {
+				CoalesceTables(dv.(map[string]interface{}), val.(map[string]interface{}))
+			} else {
 				log.Printf("warning: cannot overwrite table with non table for %s (%v)", key, val)
 			}
-		} else if dv, ok := dst[key]; ok && istable(dv) {
+		} else if istable(dv) {
 			log.Printf("warning: destination for %s is a table. Ignoring non-table value %v", key, val)
-		} else if !ok { // <- ok is still in scope from preceding conditional.
-			dst[key] = val
 		}
 	}
 	return dst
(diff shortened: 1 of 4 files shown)


bacongobbler:
After #5185 was merged, we also had to address it with #5821 and #6080. Can you please follow the discussion in #5184 and make sure this PR reflects what had to be addressed in order to get this to work?

strider-golang:
Seems to be working as expected. 👍

uzxmx:
@bacongobbler Thanks for pointing out those information. I've read #5821, #6080 and #6146. Below is what I learned:

For #5821, actually this has nothing to do with the issue in #5184, the author just did some refactor work based on dev-v2 branch when he created #5185. But since v3 has been released, a lot of refactor work about `pkg/chartuti/*.go` has been made. So we may not need to port those refactor from #5821.

For #6080, the author solved the problem that multiple calling `coalesceDeps` causes subchart value deletion not working.

For #6146, it's the same as #6080, the author just added another unit test due to repeated fix with #6080.

And for the whole story in #5184, there are two issues that need to be solved. One bug is about value deletion for top level chart. The second is about value deletion for subchart.

My first commit only solved the first bug. Thanks to your information, I've ported #6080 (the same as #6146) to the second commit, and added test cases about subchart value deletion.

Please give it a try. If there is other problem, feel free to tell me.

pkg/chartutil/coalesce.go
```diff
@@ -47,10 +47,7 @@ func CoalesceValues(chrt *chart.Chart, vals map[string]interface{}) (Values, err
 	if valsCopy == nil {
 		valsCopy = make(map[string]interface{})
 	}
-	if _, err := coalesce(chrt, valsCopy); err != nil {
-		return valsCopy, err
-	}
-	return coalesceDeps(chrt, valsCopy)
+	return coalesce(chrt, valsCopy)
```
technosophos, line 50: You did not update the docs that explain why this was called twice (lines 38-39 above).

technosophos (approved):
LGTM. I left a note that one of the comments in the code is wrong. Once that is fixed, I'll merge.

adamreese (approved):

pkg/chartutil/coalesce.go
uzxmx: Sorry, I overlooked the comment at lines 38 - 39. I think we can just remove that comment, or we can change the comment to "create a copy of vals and then pass it to coalesce, as it will mutate the passed values", because `coalesceDeps` was invoked in `colesce`, so that invoke was duplicated.

technosophos: I removed the comment in a followw-up PR because I wanted to get this merged in time for Helm 3.2 RC.1. Thanks again!

PR #7751, opened 2020-03-09
Fixing issue where archives created on windows have broken paths

### LINKED ISSUE #7748: helm v3 packaging charts differently on windows and linux (breaks artifactory)
I'm trying to package a helm chart and deploy it to a private helm repo on artifactory 6.18.0.

This works fine if I package the chart on linux, but fails if the chart is packaged on windows.

It also works fine with charts packaged on windows with a v2 helm client, so it appears to be an issue introduced in helm v3.

I can also see that the resultant tgz files are different by listing their contents with tar (note the backslashes):

Good (linux or windows v2):
```
$ tar -tvf tds-proxy-1.0.0.tgz
-rwxr-xr-x 0/0             129 2020-03-09 11:47 tds-proxy/Chart.yaml
-rwxr-xr-x 0/0             195 2020-03-09 11:47 tds-proxy/values.yaml
-rwxr-xr-x 0/0            1033 2020-03-09 11:47 tds-proxy/templates/deployment.yaml
-rwxr-xr-x 0/0             547 2020-03-09 11:47 tds-proxy/templates/service.yaml
-rwxr-xr-x 0/0             364 2020-03-09 11:47 tds-proxy/.helmignore
```

Bad (helm v3 on windows):
```
$ tar -tvf tds-proxy-1.0.0.tgz
-rw-r--r-- 0/0             129 2020-03-09 11:57 tds-proxy\\Chart.yaml
-rw-r--r-- 0/0             195 2020-03-09 11:57 tds-proxy\\values.yaml
-rw-r--r-- 0/0            1033 2020-03-09 11:57 tds-proxy\\templates\\deployment.yaml
-rw-r--r-- 0/0             547 2020-03-09 11:57 tds-proxy\\templates\\service.yaml
-rw-r--r-- 0/0             364 2020-03-09 11:57 tds-proxy\\.helmignore
```

This has been [raised as an issue with artifactory](https://www.jfrog.com/jira/browse/RTFACT-21145), but I feel it's up to helm to package consistently irrespective of where helm package is run. This looks like a bug in the helm v3 windows client.

Output of `helm version`:
`version.BuildInfo{Version:"v3.1.1", GitCommit:"afe70585407b420d0097d07b21c47dc511525ac8", GitTreeState:"clean", GoVersion:"go1.13.8"}`

Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.2", GitCommit:"59603c6e503c87169aea6106f57b9f242f64df89", GitTreeState:"clean", BuildDate:"2020-01-18T23:30:10Z", GoVersion:"go1.13.5", Compiler:"gc", Platform:"windows/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.2", GitCommit:"59603c6e503c87169aea6106f57b9f242f64df89", GitTreeState:"clean", BuildDate:"2020-01-18T23:22:30Z", GoVersion:"go1.13.5", Compiler:"gc", Platform:"linux/amd64"}
```

not relevant (bare metal ish)

mattfarina:
When archives are created on windows the path spearator in the
archive file is \\. This causes issues when the file is unpacked.
For example, on Linux the files are unpacked in a flat structure
and \ is part of the file name. This causes comp issues. In Helm
v2 the path was set as / when the archive was written. This works
on both Windows and POSIX systems.

The fix being implemented is to use the ToSlash function to ensure
/ is used as the separator.

Fixes #7748

### DIFF
--- a/pkg/chartutil/save.go
+++ b/pkg/chartutil/save.go
@@ -223,7 +223,7 @@ func writeTarContents(out *tar.Writer, c *chart.Chart, prefix string) error {
 func writeToTar(out *tar.Writer, name string, body []byte) error {
 	// TODO: Do we need to create dummy parent directory names if none exist?
 	h := &tar.Header{
-		Name:    name,
+		Name:    filepath.ToSlash(name),
 		Mode:    0644,
 		Size:    int64(len(body)),
 		ModTime: time.Now(),


meurig:
Thanks @mattfarina, sounds great!

I'm having some trouble building helm behind our corporate firewall, but once I've got those issues sorted I'll give this a go.

meurig:
I've tested this and I can confirm that it fixes #7748.

Thanks Matt!

mattfarina:
@meurig Thanks for testing.

This reminds me that we should start using GitHub Actions to perform windows testing.

bacongobbler:
I recall us fixing this in Helm 2. Perhaps it wasnot ported over. I'm pretty sure this was how we fixed it. I'll see if I can pull up the PR

bacongobbler:
In Helm 2, we use `filepath.ToSlash` [on line 215](https://github.com/helm/helm/blob/b7d2947d1010a5abf8f986886b3189d6e55b1d36/pkg/chartutil/save.go#L215). In Helm 3, it looks like this is not reflected there [on line 226](https://github.com/helm/helm/blob/66292e5e240dd9ab3d28415d9b3946ca8efc521e/pkg/chartutil/save.go#L226).

To save the number of calls to `filepath.ToSlash`, would it be easier to place this under `writeToTar`? That would be forwards-compatible as well - any calling code provided in the future won't have to ensure they escape all filepath slashes in the tar header.

mattfarina:
@bacongobbler Great catch. I should have noticed the pattern. Updated to the much simpler fix.

bacongobbler (approved):

PR #7752, opened 2020-03-09
Make get script eaiser for helm versions to live side by side (helm3 etc)

scottrigby:
Extremely tiny convenience PR. Someone on Linux asked and I thought "just use the script", then realized oh… easy enough to fix.

Example: `BINARY_NAME=helm3-1-0 DESIRED_VERSION=v3.1.0 ./get-helm-3`

Signed-off-by: Scott Rigby <scott@r6by.com>

### DIFF
--- a/scripts/get-helm-3
+++ b/scripts/get-helm-3
@@ -17,8 +17,7 @@
 # The install script is based off of the MIT-licensed script from glide,
 # the package manager for Go: https://github.com/Masterminds/glide.sh/blob/master/get
 
-PROJECT_NAME="helm"
-
+: ${PROJECT_NAME:="helm"}
 : ${USE_SUDO:="true"}
 : ${HELM_INSTALL_DIR:="/usr/local/bin"}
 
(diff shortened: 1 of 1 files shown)


scripts/get-helm-3
```diff
@@ -17,8 +17,7 @@
 # The install script is based off of the MIT-licensed script from glide,
 # the package manager for Go: https://github.com/Masterminds/glide.sh/blob/master/get
 
-PROJECT_NAME="helm"
-
+: ${PROJECT_NAME:="helm"}
```
jdolitsky: PROJECT_NAME, maybe better as HELM_EXE? or something?

jdolitsky (changes requested):

scripts/get-helm-3
scottrigby: Good idea. Or BINARY_NAME etc, something that conveys the purpose

jdolitsky (approved):
👍

scripts/get-helm-3
mspinelli: Would be nice to make this a command line option as well:

```bash
    '--name'|-n)
       shift
       if [[ $# -ne 0 ]]; then
           BINARY_NAME="${1}"
       else
           echo -e "Please provide the desired binary name. e.g. --name helm3 or -v helm"
       fi
       ;;
```

and for the help function:

```bash
  echo -e "\t[--name|-n <binary_name>] . When not defined defaults to helm otherwise useful for installing side by side a previous version"
  echo -e "\te.g. --name helm3 or -n helm3"
```

technosophos (approved):

technosophos:
@scottrigby feel free to merge this whenever you are ready.

scottrigby:
Missed the note @technosophos Thanks for merging 🍺

Issue #7749, opened 2020-03-09
Listing releases (helm list) with short listing format (--short) and json output (--output json) gives unexpected results.


git-enzo:
Output of `helm version`:
version.BuildInfo{Version:"v3.1.1", GitCommit:"afe70585407b420d0097d07b21c47dc511525ac8", GitTreeState:"clean", GoVersion:"go1.13.8"}

Output of `kubectl version`:
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.0", GitCommit:"2bd9643cee5b3b3a5ecbd3af49d09018f0773c77", GitTreeState:"clean", BuildDate:"2019-09-18T14:36:53Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.7", GitCommit:"c83d931fb9bece427bc63a02349755e0f8696d3e", GitTreeState:"clean", BuildDate:"2020-01-31T20:09:49Z", GoVersion:"go1.12.12", Compiler:"gc", Platform:"linux/amd64"}

Cloud Provider/Platform (AKS, GKE, Minikube etc.): AKS, Minikube

Hello,

When I try to list all releases for a specified namespace with json output and short (quiet) listing format, then output is no json.

It seems that combination of these two parameters breaks specified output in helm3. If these parameters are used separately, then it works as expected (either json output or short listing). This behavior is correct in helm2.

Below examples, assuming that we have 2 releases named 'test1' and 'test2' in 'test' namespace.

_Helm 3:_

Command:
`helm list --namespace test --short --output json` or `helm list --namespace test -q -o json` or other combinations

Output:
```
test1
test2
```
there is only short listing format (missing json output)

_Helm 2:_

Command:
`helm list --namespace test --short --output json`

Output:
`["test1","test2"]` as expected, short listing format with json output

mattfarina:
Thanks for the report. I have replicated the issue.

anshulvermapatel:
@mattfarina , I too replicated the issue and am working on a PR, give me some more time.

mattfarina:
@anshulvermapatel great. please ping me to review when you have it.

anshulvermapatel:
@mattfarina , Raised a PR, please review it once you get time.
Thank You, do let me know if there is any change required.

mattfarina:
#7753 was merged with a fix for this.

PR #7753, opened 2020-03-09
Solves the issue #7749

anshulvermapatel:
Resolves the [issue](https://github.com/helm/helm/issues/7749) where proper formatting was not being done if short(-q) option was used with other formatting options like json, yaml.
@mattfarina , Can you please review this.

Please do let me know if any change is required.
Thank You very much.


cmd/helm/list.go
```diff
@@ -83,6 +83,27 @@ func newListCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
...
+						output.EncodeYAML(out, names)
+						return nil
+					default:
+						return outfmt.Write(out, newReleaseListWriter(results))
+					}
+				}
+
 				for _, res := range results {
 					fmt.Fprintln(out, res.Name)
 				}
```
mattfarina: If I use `helm list --short` the default value for the output is used of `table`. This shows a different result from the use of `helm list --short --output table` which is currently showing the same result as if it were not short.

I would suggest a few changes to this block...

1. Remove the `if outputFlag.Changed {` block. This section can format for all the types including the default.
2. Add a case for `table` (the default) and have its contents be the for loop currently listed after the if changed block.
3. Change the default case (not a known output format) to return an error of `fmt.Errorf("Unknown output format \"%s\"", outputFlag.Value.String())`. This is based on what's in Helm v2.

This would make the case statement handle all short forms in all cases.

mattfarina (changes requested):
Thanks for working on this. I found an issue in the code block and I think the design can use a little improvement to clean up the issue and duplicate code. See below...

cmd/helm/list.go
anshulvermapatel: >If I use `helm list --short` the default value for the output is used of `table`. This shows a different result from the use of `helm list --short --output table` which is currently showing the same result as if it were not short.

Actually, I deployed Helm 2 and saw that there was no output format as `table` in helm.

>$ ./helm version
Client: &version.Version{SemVer:"v2.16.3", GitCommit:"1ee0254c86d4ed6887327dabed7aa7da29d7eb0d", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.16.3", GitCommit:"1ee0254c86d4ed6887327dabed7aa7da29d7eb0d", GitTreeState:"clean"
 $ ./helm list --short --output table
Error: Unknown output format "table

Hence thought that if the output is table then it should be the table, but ya  you are right, if default format is `table` then `helm list --short --output table` should also be the same as the output of `helm list --short`.

I will make the changes as given in your 1st and 2nd point.
Regarding the 3rd point. What I think is that in the `default` case of the `switch` if we return `outfmt.Write(out, newReleaseListWriter(results))` the if will throw the normal Helm 3 error message of bad format.
>╭─ansverma@localhost ~/contribution/helm/7749/helm2/helm ‹helm/7749/helm_list_short_listing_format*› 
╰─$ bin/helm list --short --output table
mysql
yoursql
╭─ansverma@localhost ~/contribution/helm/7749/helm2/helm ‹helm/7749/helm_list_short_listing_format*› 
╰─$ bin/helm list --short --output json 
["mysql","yoursql"]
╭─ansverma@localhost ~/contribution/helm/7749/helm2/helm ‹helm/7749/helm_list_short_listing_format*› 
╰─$ bin/helm list --short --output yaml
\- mysql
\- yoursql
╭─ansverma@localhost ~/contribution/helm/7749/helm2/helm ‹helm/7749/helm_list_short_listing_format*› 
╰─$ bin/helm list --short --output asdflkj
Error: invalid argument "asdflkj" for "-o, --output" flag: invalid format type

I think, the error message `Error: invalid argument "asdflkj" for "-o, --output" flag: invalid format type` is better than the one  in Helm 2.
I am currently making the changes. Please let me know if you want to have that particular error message which you mentioned anyways, I will do that.

anshulvermapatel:
@mattfarina any update?

anshulvermapatel:
@mattfarina , Hello, Any update on this?

@bacongobbler , Can you please have a look at this?

bacongobbler:
Thank you for the ping. At this time, I do not have enough time in my work schedule to review your code. I am busy with other work. Sorry!

anshulvermapatel:
@bacongobbler , No Problem Matthew.

mattfarina (approved):
lgtm. thanks for working this out.

Issue #7528, opened 2020-02-05
Install Helm plugins globally


mkjpryor-stfc:
It would be great if Helm could have two levels of plugins - global and per-user. Currently, we only have per-user. This makes, for example, building a re-usable Docker image containing Helm tricky.

mattfarina:
You can do this today and it needs better docs. I've filed a feature request which I think would make this easier via #7529.

You can sorta do this today using the `HELM_PLUGINS` environment variable. This can accept a list of directories to look in for plugins using the systems path separator for the different directories. If you run `helm env` you can see where the current home directory for the user is.

Does this help?

bitdancer:
Not really.  IMO the point here is that there should be a *global* directory, AKA a Well Known system directory (such as /var/lib/helm) where global plugins are installed, and which helm would check without needing any alteration to the user's environment.  This is how most tailorable systems work in unix.

technosophos:
This will be fixed via support of XDG_DATA_DIRS.

bitdancer:
What does desktop environments have to do with helm?  I don't use a desktop, and the concern raised here is for installing helm in a *container*, which certainly doesn't have a desktop.

I don't think support XDG_DATA_DIRS fixes this request.

technosophos:
The XDG base dir specification seems to be widely, if not universally, accepted as the new location for what used to be dotfiles in the home directory. https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

Over the last few years, we have observed most other tools supporting this convention. So we switched as well.

One thing that XDG_DATA_DIRS offers us is a convention for where we can place system-wide plugins, which is exactly what this issue is discussing.

AFAIK, the presence or absence of a "desktop environment" has no bearing on usage XDG Base Dir specification, though many desktop environments do have additional support for this specification.

All of this is part of the effort to not have a mess of specialized environment variables that Helm users need to learn and tweak in order to get Helm to work the way they want.

bitdancer:
OK, that doesn't seem crazy :)  I've looked at the XDG spec a bit more (though still not in detail) and the linked PR, and it seems like there are still a couple missing pieces.  I also see that gcloud supports XDG, so it makes sense for helm to do so as well, regardless.

[...]

PR #7755, opened 2020-03-11
Add XDG_DATA_DIRS support for using plugins from multiple locations

### LINKED ISSUE #7529: Support $XDG_DATA_DIRS
The [XDG Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) has an environment variable known as `$XDG_DATA_DIRS`. This variable can have multiple data directories to look for things (e.g., plugins). The spec notes:

> $XDG_DATA_DIRS defines the preference-ordered set of base directories to search for data files in addition to the $XDG_DATA_HOME base directory. The directories in $XDG_DATA_DIRS should be seperated with a colon ':'. 

A docker container or something else could system wide data in here (e.g., plugins).

waveywaves:
fixes https://github.com/helm/helm/issues/7529

This will add support for XDG_DATA_DIRS env var which will
allow for using pugins from multiple locations defined in XDG_DATA_DIRS


waveywaves:
cc @mattfarina 
Need some clarity on a few pointers below 

- We need to Introduce HELM_DATA_DIRS envVar which can leverage XDG_DATA_DIRS for information.

- XDG_DATA_DIRS already exists as an envVar in our PCs, how do we segregate HELM related EnvVars ? OR do we set them directly in HELM_DATA_DIRS if we plan to introduce them ?

- Plugin Download to multiple DATA DIRS can be supported with proper implementation, https://github.com/helm/helm/issues/7528. Does the global installation of plugin need to be facilitated by a `--global` | `-g` flag ? As it makes sense to keep plugin install local to the preferred HELM_PLUGINS directory.

technosophos:
What is the status on this PR?

jdolitsky:
@waveywaves can you check out the CI failure?
```
internal/test/ensure/ensure.go:34: File is not `gofmt`-ed with `-s` (gofmt)
	os.Setenv(xdg.DataDirsEnvVar,base+";")
cmd/helm/root_test.go:34: File is not `gofmt`-ed with `-s` (gofmt)
		envvars                                     map[string]string
```

waveywaves:
@technosophos Can you PTAL ?

technosophos:
This one still needs a rebase, but after that I'll work with @adamreese on us getting it reviewed.

waveywaves:
@technosophos done with rebase. Do check it oit

hickeyma (commented):
I wonder should the PR possibly look to support general case of an overall override, similar to a HELM_HOME in v2? Issue #7919 could be supported by the pr also.

technosophos:
Please hold on this PR until we get clarity on #7529. I don't think this PR solves the issue, but we need to make sure we understand the issue.

pkg/helmpath/lazypath.go
```diff
@@ -47,3 +48,13 @@ func (l lazypath) configPath(elem ...string) string {
 func (l lazypath) dataPath(elem ...string) string {
 	return l.path(xdg.DataHomeEnvVar, dataHome, filepath.Join(elem...))
 }
+
+// dataPath defines the base directory relative to which user specific data files should be stored.
+func (l lazypath) dataDirsPath(elem ...string) string {
```
technosophos, line 53: The spec says that local data dir (XDG_LOCAL_DATA) should be read first, and should be pre-pended to the list of XDG_DATA_DIRS.

technosophos: Sorry... `XDG_DATA_HOME`, not `XDG_LOCAL_DATA`. See https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

pkg/helmpath/lazypath_unix.go
```diff
@@ -43,3 +44,10 @@ func configHome() string {
 func cacheHome() string {
 	return filepath.Join(homedir.HomeDir(), ".cache")
 }
+
+// dataDirs defines the directories relevant directories in use separated with a colon
+//
+// $XDG_DATA_DIRS is usually set by the system
+func dataDirs() string {
```
technosophos, line 51: This does not seem to be what the spec suggests doing.

The spec seems to suggest that the system can provide additional paths on XDG_DATA_DIRS that application code (like Helm) may/should consider as other locations where data may be stored. But cache and config are not also to be considered data dirs. This would imply that stuff in the cache was to be treated as data dirs (and likewise with config). I don't think that is what the spec intends.

technosophos (changes requested):
I think a few different things got put in the same PR.

- the XDG spec provides a convention for naming three different _kinds_ of directories: Cache, Data, and Config. Different data is stored in each of these. They are not interchangeable. So we should never return, say, the config directory when a request is made for a data directory.
- The XDG_DATA_DIRS directory is supplied by the system, and is to be _appended_ to the user-supplied XDG_DATA_HOME directory. Helm should never override or presume to generate a value for XDG_DATA_DIRS. Think of it as communication from the system to the user and to Helm.
- What @mattfarina is asking for in #7529 is that when Helm loads plugins, it also checks the XDG_DATA_DIRS for additional plugins, the way `os/exec.LookPath` looks at each of the subdirs in `PATH` to find the first matching executable.

This seems to have support for #7919. There is some debate over whether we ought to do that. So maybe we should pick one of the two issues and solve it, rather than have both issues in the same PR.

waveywaves:
@technosophos I understand this better now. Let me update the PR accordingly.

waveywaves:
Continuing work on this today and finishing off.

waveywaves:
@technosophos @adamreese can you PTAL ? I have updated to search for plugins in the XDG_DATA_DIRS dirs

waveywaves:
@technosophos @adamreese can you PTAL ? Have updated it based on your comments.

PR #7758, opened 2020-03-11
feat(cmd) Add flag for InMemory Cache for downloading charts

waveywaves:
This feature allows us to use the --use-inmemory-cache flag to
use a chart without downloading it to the helm repository cache,
instead which is downloaded to InMemoryCache

augments and dependent on
 https://github.com/helm/helm/pull/7601

### DIFF
--- a/cmd/helm/install.go
+++ b/cmd/helm/install.go
@@ -19,6 +19,8 @@ package main
 import (
 	"fmt"
 	"io"
+	"os"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
@@ -151,6 +153,7 @@ func addInstallFlags(f *pflag.FlagSet, client *action.Install, valueOpts *values
 	f.BoolVar(&client.Atomic, "atomic", false, "if set, the installation process deletes the installation on failure. The --wait flag will be set automatically if --atomic is used")
 	f.BoolVar(&client.SkipCRDs, "skip-crds", false, "if set, no CRDs will be installed. By default, CRDs are installed if not already present")
 	f.BoolVar(&client.SubNotes, "render-subchart-notes", false, "if set, render subchart notes along with the parent")
+	f.BoolVar(&client.ReadOnlyFS, "readonly-fs", false, "if set, do not download charts to the HELM_REPOSITORY_CACHE")
 	addValueOptionsFlags(f, valueOpts)
 	addChartPathOptionsFlags(f, &client.ChartPathOptions)
 }
@@ -162,13 +165,13 @@ func runInstall(args []string, client *action.Install, valueOpts *values.Options
 		client.Version = ">0.0.0-0"
 	}
 
-	name, chart, err := client.NameAndChart(args)
+	name, schart, err := client.NameAndChart(args)
 	if err != nil {
 		return nil, err
 	}
 	client.ReleaseName = name
 
-	cp, err := client.ChartPathOptions.LocateChart(chart, settings)
+	cp, err := client.ChartPathOptions.LocateChart(schart, settings)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


waveywaves:
cc @marckhouzam

cmd/helm/install.go
```diff
@@ -181,10 +183,19 @@ func runInstall(args []string, client *action.Install, valueOpts *values.Options
 		return nil, err
 	}
 
-	// Check chart dependencies to make sure all are present in /charts
-	chartRequested, err := loader.Load(cp)
-	if err != nil {
-		return nil, err
+	// Check chart dependencies to make sure all are present in the InMemoryCache or in /charts
+	chartRequested := &chart.Chart{}
```
marckhouzam, line 187: I'm not very experienced in Go, so maybe this does not matter, but it sounds more efficient, instead of initializing to a value that will never be used, tojust declaring the variable like so: `var chartRequested chart.Chart`

cmd/helm/install.go
```diff
@@ -181,10 +183,19 @@ func runInstall(args []string, client *action.Install, valueOpts *values.Options
...
+	if client.UseInMemoryCache {
+		if _, err := os.Stat(cp); err == nil {
+			chartRequested, err = loader.Load(cp)
+		} else {
+			chartRequested = downloader.InMemoryCache[schart].Chart
```
marckhouzam, line 192: This does not compile.  There is no `Chart` field in you latest version of #7601

marckhouzam (changes requested):

cmd/helm/install.go
```diff
@@ -181,10 +183,19 @@ func runInstall(args []string, client *action.Install, valueOpts *values.Options
...
 
-	// Check chart dependencies to make sure all are present in /charts
-	chartRequested, err := loader.Load(cp)
-	if err != nil {
-		return nil, err
+	// Check chart dependencies to make sure all are present in the InMemoryCache or in /charts
+	chartRequested := &chart.Chart{}
+	if client.UseInMemoryCache {
+		if _, err := os.Stat(cp); err == nil {
+			chartRequested, err = loader.Load(cp)
```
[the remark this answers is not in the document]
marckhouzam: This suggestion may not be feasible, I haven't tried.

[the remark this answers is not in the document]
waveywaves: This is not feasible in the current scenario, we need to load the chart with LoadArchive from the data loaded from the cache

cmd/helm/install.go
```diff
@@ -151,6 +152,7 @@ func addInstallFlags(f *pflag.FlagSet, client *action.Install, valueOpts *values
 	f.BoolVar(&client.Atomic, "atomic", false, "if set, installation process purges chart on fail. The --wait flag will be set automatically if --atomic is used")
 	f.BoolVar(&client.SkipCRDs, "skip-crds", false, "if set, no CRDs will be installed. By default, CRDs are installed if not already present")
 	f.BoolVar(&client.SubNotes, "render-subchart-notes", false, "if set, render subchart notes along with the parent")
+	f.BoolVar(&client.ReadOnlyFS, "readonly-fs", false, "if set, do not download charts to the HELM_REPOSITORY_CACHE")
```
technosophos, line 155: This flag name doesn't seem to match what it does. I mean, I could run this on a read-write filesystem just as easily, right?

marckhouzam: I'm glad you brought this up. I'm having a lot of trouble coming up with a future-proof name, for the flag but also for the corresponding API in #7601.

waveywaves: @technosophos Yes you should be able to run this on a read write file system just easily.

jdolitsky:
@waveywaves can you checkout the CI failure?
```
# helm.sh/helm/v3/cmd/helm [helm.sh/helm/v3/cmd/helm.test]
cmd/helm/install.go:156: client.ReadOnlyFS undefined (type *action.Install has no field or method ReadOnlyFS)
cmd/helm/install.go:189: client.ReadOnlyFS undefined (type *action.Install has no field or method ReadOnlyFS)
cmd/helm/install.go:193: undefined: downloader.GetChart
FAIL	helm.sh/helm/v3/cmd/helm [build failed]
```

waveywaves:
Closing as #7601 wouldn't need the flag to run anymore.

PR #7785, opened 2020-03-19
Restores display of object status when displaying helm status

### LINKED ISSUE #6896: helm status stopped showing containers
Couldn't find it in the latest changelog, but:

before upgrade to 2.16, it used to show more detailed information about running pods. Now I need to run **kubectl -n <namespace> get pods -l app=\<app-name\>** to find out if my containers started and passed the Liveness probes. Previously I could just do **helm status \<release\>**. Any chances to return to the original behavior?

Output of `helm version`: 
Client: &version.Version{SemVer:"v2.16.0", GitCommit:"e13bc94621d4ef666270cfbe734aaabf342a49bb", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.16.0", GitCommit:"e13bc94621d4ef666270cfbe734aaabf342a49bb", GitTreeState:"clean"}

Output of `kubectl version`:
kubectl version
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-15T23:41:55Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"14+", GitVersion:"v1.14.7-eks-e9b1d0", GitCommit:"e9b1d0551216e1e8ace5ee4ca50161df34325ec2", GitTreeState:"clean", BuildDate:"2019-09-21T08:33:01Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"linux/amd64"}

EKS

mattfarina:
Changes to the Kubernetes API server and kubectl libraries caused
the status to no longer display when helm status was run for a
release. This change restores the status display.

Generation of the tables for display was optionally moved server
side. A request for the data as a table is made and a kubectl
printer for tables can display this data. Kubectl uses this setup and
the structure here closely resembles kubectl. kubectl is still
able to display objects as tables from prior to server side
printing. Server side printing is able to handle non-core
resources like CRDs.

Note, an extra request is made because table responses cannot be
easily transformed into Go objects for Kubernetes types to work
with. There is one request to get the resources for display in
a table and a second request to get the resources to lookup the
related pods. The related pods are now requested as a table as
well for display purposes.

This is likely part of the larger trend to move features like
this server side so that more libraries in more languages can
get to the feature.

The fix only works for Kubernetes 1.15 and newer. This is when
the table functionality was added to the api server. Kubernetes
supports the latest 3 minor versions meaning kubectl does the
same. It only needs to support back to 1.15 now for table
display. In Kubernetes 1.14 kubectl can still display the table
information using different printing methods.

Closes #6896

**If someone knows how to fix this without using server side printing I would appreciate knowing as it would enable support to 1.14 and older versions of Kubernetes.**

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -23,6 +23,7 @@ import (
 	goerrors "errors"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"log"
 	"sort"
 	"strings"
@@ -42,6 +43,8 @@ import (
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	metav1beta1 "k8s.io/apimachinery/pkg/apis/meta/v1beta1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -51,13 +54,14 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"k8s.io/cli-runtime/pkg/printers"
 	"k8s.io/cli-runtime/pkg/resource"
 	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
 	cachetools "k8s.io/client-go/tools/cache"
 	watchtools "k8s.io/client-go/tools/watch"
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
 	"k8s.io/kubectl/pkg/validation"
-	"k8s.io/kubernetes/pkg/apis/core"
 	"k8s.io/kubernetes/pkg/kubectl/cmd/get"
 )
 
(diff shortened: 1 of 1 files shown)


bacongobbler:
Taking this for a spin and trying to sift through the Kubernetes codebase to see if there's a better way.

bacongobbler (approved):
After banging my head against kubectl print flags for a few hours, I could not find any backwards-compatible solutions that worked without server-side printing.

LGTM.

mattfarina:
Note, the latest change also fixes sorting and versions of Kubernetes 1.14 and older.

thomastaylor312 (approved):

pkg/kube/client.go
```diff
@@ -158,6 +162,40 @@ func (c *Client) BuildUnstructured(namespace string, reader io.Reader) (Result,
...
+}
+
+// This is used to retrieve a table view of the data. A table view is how kubectl
+// retrieves the information Helm displays as resources in status. Note, table
+// data is returned as a Table type that does not conform to the runtime.Object
+// interface but is that type. So, you can't transform it into Go objects easily.
+func transformRequests(req *rest.Request) {
+
+	// The request headers are for both the v1 and v1beta1 versions of the table
+	// as Kubernetes 1.14 and older used the beta version.
```
bacongobbler: :tada:

bacongobbler (approved):

mattfarina:
Thanks for the reviews. This was a hairy problem.

PR #7789, opened 2020-03-21
Return dry-run upgrade status

dippynark:
The Helm Terraform provider currently cannot determine whether an upgrade of a particular chart is required: https://github.com/terraform-providers/terraform-provider-helm/issues/444

This PR adds a new status type when a dry-run upgrading is performed when  and resources are missing from Kubernetes. This should allow the provider to dry-run an upgrade when reading to decide whether an actual (non dry-run) upgrade is required.

I'm not sure if this would be the best way to do this, or if this breaks other assumptions in Helm about status, but a suggested implementation just for feedback atm.

I assume the note in the code:

```
// NOTE: Make sure to update cmd/helm/status.go when adding or modifying any of these statuses.
```

is outdated - I can't seem to see any references to the status constants in that file.

### DIFF
--- a/pkg/action/upgrade.go
+++ b/pkg/action/upgrade.go
@@ -242,6 +242,9 @@ func (u *Upgrade) performUpgrade(originalRelease, upgradedRelease *release.Relea
 	})
 
 	if u.DryRun {
+		if len(toBeCreated) > 0 {
+			upgradedRelease.Info.Status = release.StatusRequiresUpgrade
+		}
 		u.cfg.Log("dry run for %s", upgradedRelease.Name)
 		if len(u.Description) > 0 {
 			upgradedRelease.Info.Description = u.Description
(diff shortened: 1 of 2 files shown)


technosophos:
This appears to be a breaking change, since it changes the output to something unexpected. That said, it's unclear to me whether there are any existing uses cases where this would be a surprise to users.

We need to give this some thought before merging.

technosophos:
Any update on this?

dippynark:
@technosophos thanks for taking a look - I thought it would catch deletions as the `toBeCreated` list will be populated when things are missing? It just wouldn't catch changes not applied through Helm.

Potentially there are better ways of doing this though and this would introduce a breaking change, happy for this to be closed if that is the case, it was just an initial idea without knowing the wider implications.

PR #7791, opened 2020-03-21
Implement support for multiple args for repo remove

lindhe:
I found that I sometimes I want to remove multiple repos at once.
Thought I would give it a Go to try and to make that a bit easier to do.

This is my first contribution to the Helm code base, and I am doing this
party to become more familiar with Helm and Go. So please point me in
the right direction if I've missed to change something.

Usually, I'd create an issue before creating a PR. If this change is
unwanted, please feel free to discard it – no hard feelings. I made it
mostly as an exercise anyway, so it was not wasted time for me.

### DIFF
--- a/cmd/helm/repo_list.go
+++ b/cmd/helm/repo_list.go
@@ -108,8 +108,8 @@ func filterRepos(repos []*repo.Entry, ignoredRepoNames []string) []*repo.Entry {
 	filteredRepos := make([]*repo.Entry, 0)
 
 	ignored := make(map[string]bool, len(ignoredRepoNames))
-	for _, repo := range ignoredRepoNames {
-		ignored[repo] = true
+	for _, repoName := range ignoredRepoNames {
+		ignored[repoName] = true
 	}
 
 	for _, repo := range repos {
--- a/cmd/helm/repo_list.go
+++ b/cmd/helm/repo_list.go
@@ -107,7 +107,7 @@ func filterRepos(repos []*repo.Entry, ignoredRepoNames []string) []*repo.Entry {
 
 	filteredRepos := make([]*repo.Entry, 0)
 
-	ignored := make(map[string]bool, 0)
+	ignored := make(map[string]bool, len(ignoredRepoNames))
 	for _, repo := range ignoredRepoNames {
 		ignored[repo] = true
 	}
--- a/cmd/helm/repo_list.go
+++ b/cmd/helm/repo_list.go
@@ -97,13 +97,39 @@ func (r *repoListWriter) encodeByFormat(out io.Writer, format output.Format) err
 	return nil
 }
 
+// Returns all repos from repos, except those with names matching ignoredRepoNames
+// Inspired by https://stackoverflow.com/a/28701031/893211
+func filterRepos(repos []*repo.Entry, ignoredRepoNames []string) []*repo.Entry {
+	// if ignoredRepoNames is nil, just return repo
+	if ignoredRepoNames == nil {
+		return repos
+	}
+
+	filteredRepos := make([]*repo.Entry, 0)
+
+	ignored := make(map[string]bool, 0)
+	for _, repo := range ignoredRepoNames {
+		ignored[repo] = true
+	}
+
+	for _, repo := range repos {
+		_, removed := ignored[repo.Name]
+		if !removed {
+			filteredRepos = append(filteredRepos, repo)
+		}
+	}
+
+	return filteredRepos
+}
+
 // Provide dynamic auto-completion for repo names
-func compListRepos(prefix string) []string {
+func compListRepos(prefix string, ignoredRepoNames []string) []string {
 	var rNames []string
 
 	f, err := repo.LoadFile(settings.RepositoryConfig)
 	if err == nil && len(f.Repositories) > 0 {
-		for _, repo := range f.Repositories {
+		filteredRepos := filterRepos(f.Repositories, ignoredRepoNames)
+		for _, repo := range filteredRepos {
 			if strings.HasPrefix(repo.Name, prefix) {
 				rNames = append(rNames, repo.Name)
 			}
--- a/cmd/helm/repo_list.go
+++ b/cmd/helm/repo_list.go
@@ -113,8 +113,7 @@ func filterRepos(repos []*repo.Entry, ignoredRepoNames []string) []*repo.Entry {
 	}
 
 	for _, repo := range repos {
-		_, removed := ignored[repo.Name]
-		if !removed {
+		if _, removed := ignored[repo.Name]; !removed {
 			filteredRepos = append(filteredRepos, repo)
 		}
 	}
--- a/cmd/helm/repo_remove.go
+++ b/cmd/helm/repo_remove.go
@@ -55,7 +55,7 @@ func newRepoRemoveCmd(out io.Writer) *cobra.Command {
 
 	// Function providing dynamic auto-completion
 	completion.RegisterValidArgsFunc(cmd, func(cmd *cobra.Command, args []string, toComplete string) ([]string, completion.BashCompDirective) {
-		return compListRepos(toComplete), completion.BashCompDirectiveNoFileComp
+		return compListRepos(toComplete, args), completion.BashCompDirectiveNoFileComp
 	})
 
 	return cmd
--- a/cmd/helm/repo_remove.go
+++ b/cmd/helm/repo_remove.go
@@ -55,9 +55,6 @@ func newRepoRemoveCmd(out io.Writer) *cobra.Command {
 
 	// Function providing dynamic auto-completion
 	completion.RegisterValidArgsFunc(cmd, func(cmd *cobra.Command, args []string, toComplete string) ([]string, completion.BashCompDirective) {
-		if len(args) != 0 {
-			return nil, completion.BashCompDirectiveNoFileComp
-		}
 		return compListRepos(toComplete), completion.BashCompDirectiveNoFileComp
 	})
 
--- a/cmd/helm/repo_remove.go
+++ b/cmd/helm/repo_remove.go
@@ -43,7 +43,7 @@ func newRepoRemoveCmd(out io.Writer) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:     "remove [REPO1 [REPO2 ...]]",
 		Aliases: []string{"rm"},
-		Short:   "remove a chart repository",
+		Short:   "remove one or more chart repositories",
 		Args:    require.MinimumNArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			o.repoFile = settings.RepositoryConfig
--- a/cmd/helm/repo_remove.go
+++ b/cmd/helm/repo_remove.go
@@ -32,7 +32,7 @@ import (
 )
 
 type repoRemoveOptions struct {
-	name      string
+	names     []string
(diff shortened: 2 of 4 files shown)


lindhe:
Ah, tests. Cool. I'll try to go through them at some opportune moment.

cmd/helm/repo_remove.go
```diff
@@ -41,14 +41,14 @@ func newRepoRemoveCmd(out io.Writer) *cobra.Command {
 	o := &repoRemoveOptions{}
 
 	cmd := &cobra.Command{
-		Use:     "remove [NAME]",
+		Use:     "remove [REPO1 [REPO2 ...]]",
 		Aliases: []string{"rm"},
 		Short:   "remove a chart repository",
```
marckhouzam, line 46: Please update the Short description for this new behavior

marckhouzam (changes requested):
I like this. It gives users a little more flexibility.

Can you update the completion code a little lower to match? It currently only provides repo names as completions if there are no other arguments. With your change it should provide repo names all the time.  If you want to be really thorough, the completions could ignore any repo already provided by previous arguments.

cmd/helm/repo_remove.go
lindhe: Got it. :) Will look into it.

lindhe:
It wasn't trivial for me to make it ignore already completed repos, but I'll give it one more shot.

marckhouzam:
> It wasn't trivial for me to make it ignore already completed repos, but I'll give it one more shot.

The list of repos passed in is in the `args` parameter of the completion function. You can probably just loop over the `args` and remove the existing ones from the received completion list.  It's basically removing all elements of one list from another list. 

Then again, this is just completion so it's not a big deal if we suggest a repo that's already being removed. The user just won't select it.  The cool thing about filtering is that a user could know what repos are left by simply typing TAB as she adds repos to the removal command.

It's not a blocking request.

marckhouzam:
@lindhe your last commit is not signed. You should force-push an updated one please.

lindhe:
What about this solution? Any good? It looks like it has quadratic complexity, but perhaps it's not a big deal in this context?

cmd/helm/repo_list.go
```diff
@@ -97,13 +97,39 @@ func (r *repoListWriter) encodeByFormat(out io.Writer, format output.Format) err
...
+// Inspired by https://stackoverflow.com/a/28701031/893211
+func filterRepos(repos []*repo.Entry, ignoredRepoNames []string) []*repo.Entry {
+	// if ignoredRepoNames is nil, just return repo
+	if ignoredRepoNames == nil {
+		return repos
+	}
+
+	filteredRepos := make([]*repo.Entry, 0)
+
+	ignored := make(map[string]bool, 0)
```
marckhouzam, line 110: Since we know that `ignored` will always be of size `len(ignoredRepoNames)`, how about setting that size in the call to `make`?

cmd/helm/repo_list.go
```diff
@@ -97,13 +97,39 @@ func (r *repoListWriter) encodeByFormat(out io.Writer, format output.Format) err
...
+	for _, repo := range ignoredRepoNames {
```
marckhouzam, line 111: Nit: to avoid using the same variable name with two different types, can you rename `repo` here to `name` or `repoName`?

marckhouzam (changes requested):
Nicely done! Two little nits and we're good to go.

marckhouzam:
Oh and the `repo_remove_test.go` has to be updated to match, as it is failing now.
You can do
```
make test>result
```
You will see it no longer compiles

marckhouzam:
One other request @lindhe, if you don't mind.  Can you add a test that removes more than one repo and checks that they are both removed?

lindhe:
> One other request @lindhe, if you don't mind. Can you add a test that removes more than one repo and checks that they are both removed?

I very much agree that we should test that, but I don't have more time to work on this today. Feel free to write such tests, or we can just wait until a day when I have more time. :) Not sure if that'll be tomorrow or in two weeks though.

lindhe:
BTW, it seems I'm missing at least `golangci-lint` because `make test` fails and tells me that. Any pointers on how to install it (on Ubuntu)?

marckhouzam:
> BTW, it seems I'm missing at least `golangci-lint` because `make test` fails and tells me that. Any pointers on how to install it (on Ubuntu)?
```
curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.24.0
```
taken from https://github.com/golangci/golangci-lint#install

I'll let you finish off writing the test, whenever you have time. You've come far enough to deserve full credit for this improvement.

lindhe:
It'd be best to keep all the old tests, right? Just adding a new test that also tests multiple repos at once?

marckhouzam:
Yes, might as well keep the existing test to make sure removing a single repo works, and adding a very similar test for removing two or three repos.

lindhe:
I had to refactor the code a bit to avoid too much copy-paste programming. But I don't quite grasp the parts that I had to refactor (the use of `cacheIndexFile` and `cacheChartFile`), so would definitely be good with some eyes on that.

cmd/helm/repo_remove_test.go
```diff
@@ -95,4 +81,74 @@ func TestRepoRemove(t *testing.T) {
 	if f.Has(testRepoName) {
 		t.Errorf("%s was not successfully removed from repositories list", testRepoName)
 	}
+
+	// Test removal of multiple repos in one go
+	var testRepoNames = []string{"foo", "bar", "baz"}
+	b2 := bytes.NewBuffer(nil)
```
lindhe, line 87: Is it right to use the same buffer object for all three repos?

cmd/helm/repo_remove_test.go
```diff
@@ -95,4 +81,74 @@ func TestRepoRemove(t *testing.T) {
...
+		f, err := repo.LoadFile(repoFile)
+		if err != nil {
+			t.Error(err)
+		}
+		if f.Has(repoName) {
+			t.Errorf("%s was not successfully removed from repositories list", repoName)
+		}
+		cacheIndex := idxs[repoName][0]
+		cacheChart := idxs[repoName][1]
+		testIndices(t, cacheIndex, cacheChart, repoName)
```
lindhe, line 129: I'm pretty unhappy about this solution. Looks pretty messy, and the definition of `idxs` on line 102 is pretty bad too. Do we even need to check each repo individually, or can we perhaps check them all in one swoop, getting rid of having to store the index files for each repo...?

cmd/helm/repo_remove_test.go
marckhouzam: The buffer is simply used to gather the printouts of the remove command.  You can use the same buffer throughout the function as long as you call `b.Reset()` before calling a new `repo remove` command.

cmd/helm/repo_remove_test.go
marckhouzam: Personally, I would go simpler.  You are creating two repos, and each repo will have an index file and a cache file.  The test just needs to make sure both files are removed after the repos are removed.  I would just have 4 variables. Instead of the old `idx` and `idx2` which were poorly named anyway, I would simply have  `indexFile1`, `cacheFile1` and `indexFile2`, `cacheFile2`.

cmd/helm/repo_remove_test.go
```diff
@@ -95,4 +81,74 @@ func TestRepoRemove(t *testing.T) {
...
+	}
+}
+
+func indexing(buf *bytes.Buffer, rootDir string, repoName string) (cacheIndexFile string, cacheChartsFile string) {
```
marckhouzam, line 133: The buffer is not related to the handling below, so you don't need to pass it in.  The `b.Reset()` should just be done before the call to `o.Run(b)` so that the output of the `repo remove` command will end up in an empty buffer.

cmd/helm/repo_remove_test.go
```diff
@@ -62,15 +62,7 @@ func TestRepoRemove(t *testing.T) {
...
 	}
 
-	idx := filepath.Join(rootDir, helmpath.CacheIndexFile(testRepoName))
-	mf, _ := os.Create(idx)
-	mf.Close()
-
-	idx2 := filepath.Join(rootDir, helmpath.CacheChartsFile(testRepoName))
-	mf, _ = os.Create(idx2)
-	mf.Close()
-
```
marckhouzam, line 72: The above code is cryptic.  What it means to do is to get a reference to the indexFile and to the cacheFile, so that after the `repo remove` command , we have a handle to check that both files no longer exist.

marckhouzam (changes requested):
Yeah, the test is a bit cryptic.  I tried to explain a bit.  Let me know if you want more clarifications.

marckhouzam (approved):

marckhouzam:
LGTM!
Thanks for you perseverance @lindhe 

Because of the tests, the size of the PR is now considered "large", so I need another maintainer to approve also.

cmd/helm/repo_list.go
```diff
@@ -97,13 +97,39 @@ func (r *repoListWriter) encodeByFormat(out io.Writer, format output.Format) err
...
+
+	filteredRepos := make([]*repo.Entry, 0)
+
+	ignored := make(map[string]bool, len(ignoredRepoNames))
+	for _, repoName := range ignoredRepoNames {
+		ignored[repoName] = true
+	}
+
+	for _, repo := range repos {
+		_, removed := ignored[repo.Name]
```
technosophos, line 116: nit: This could be reduced to:

```go
if _, removed := ignored[repo.Name]; !removed {...
```

Not a blocker.

lindhe:
I intended to also clean up the flow of the test code (at least the bit from @marckhouzam's [suggestions](https://github.com/helm/helm/pull/7791#discussion_r404506435)), because I'm having a hard time following the current logic.

But I'm not particularly enthusiastic about rewriting some working but messy code, and I'm not sure when I have the time and motivation for it. So if you are happy with the current state, maybe we call this PR done and leave the tidying up for another PR. Your choice.

cmd/helm/repo_remove_test.go
```diff
@@ -95,4 +84,79 @@ func TestRepoRemove(t *testing.T) {
...
+		if f.Has(repoName) {
+			t.Errorf("%s was not successfully removed from repositories list", repoName)
+		}
+		cacheIndex := idxs[repoName][0]
+		cacheChart := idxs[repoName][1]
+		testIndices(t, cacheIndex, cacheChart, repoName)
+	}
+}
+
+func indexing(rootDir string, repoName string) (cacheIndexFile string, cacheChartsFile string) {
```
technosophos, line 143: This function name does not make sense to me. Can we name it something that better indicates that it is just creating two files?

technosophos (approved):
Finished reviewing the code, and it is good to go. I left one non-blocking nit.

@marckhouzam : Feel free to merge whenever you are ready.

cmd/helm/repo_remove_test.go
marckhouzam: Since it was brought up, I agree, might as well give it a descriptive name.
So, one last change please @lindhe and then we merge.

lindhe:
I took the liberty to improve some other names too.

marckhouzam:
Thanks @lindhe, this looks great.

And thanks @technosophos for the second review.

PR #7800, opened 2020-03-24
chartutil: apply all labels to pods in default deployment template

mig4:
Pods created from deployment template generated with `helm create`
currently only have the selector labels applied (i.e.
`app.kubernetes.io/name` and `app.kubernetes.io/instance`). This
changes the template so that all labels are applied to pods, just like
any other object the app consists of.

Signed-off-by: mig4 <42650719@auril.club>

### DIFF
--- a/pkg/chartutil/create.go
+++ b/pkg/chartutil/create.go
@@ -238,7 +238,7 @@ spec:
   template:
     metadata:
       labels:
-        {{- include "<CHARTNAME>.selectorLabels" . | nindent 8 }}
+        {{- include "<CHARTNAME>.labels" . | nindent 8 }}
     spec:
     {{- with .Values.imagePullSecrets }}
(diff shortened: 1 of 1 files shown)


mig4:
I've been thinking about it some more and I'm actually on the fence on whether that should be the case. After all the deployment template applies to all containers in a pod and therefore it could be confusing to have e.g. `app.kubernetes.io/version=1.0.0` label on a pod that runs `the-app-1.0.0` in one container and `a-sidecar-3.14` in another. On the other hand the labels describe the main app, not sidecars, so... Anyway, it'd be cool to hear some opinions.

technosophos:
I'm not opposed to the PR, but I am also not sure I understand why we would make the change. The current version is designed to satisfy the main requirements of a deployment label selector.

Issue #5595, opened 2019-04-12
app-name has no deployed releases


sta-szek:
**Output of `helm version`:**
```
$ helm version 
Client: &version.Version{SemVer:"v2.13.1", GitCommit:"618447cbf203d147601b4b9bd7f8c37a5d39fbb4", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.13.1", GitCommit:"618447cbf203d147601b4b9bd7f8c37a5d39fbb4", GitTreeState:"clean"}
```
**Output of `kubectl version`:**
```
$ kubectl version 
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.0", GitCommit:"641856db18352033a0d96dbc99153fa3b27298e5", GitTreeState:"clean", BuildDate:"2019-03-25T15:53:57Z", GoVersion:"go1.12.1", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"10", GitVersion:"v1.10.11", GitCommit:"637c7e288581ee40ab4ca210618a89a555b6e7e9", GitTreeState:"clean", BuildDate:"2018-11-26T14:25:46Z", GoVersion:"go1.9.3", Compiler:"gc", Platform:"linux/amd64"}
```

**Cloud Provider/Platform (AKS, GKE, Minikube etc.):**  Amazon

**What is happening:**
After few broken deployments, helm (or tiller) is broken and all subsequent deployments (no matter if fixed or still broken) ends with following error: `app-name has no deployed releases`

**How to reproduce:**
We have 
```
spec:
  revisionHistoryLimit: 1
```
but i think it is not relevant.

Path a:
1. Deploy any service - working
2. Break it, e.g. by exiting containers after startup, so the whole deployment will be broken
3. Repeat it exactly 3 times
4. All next deployments will have error, no matter if fixed or broken

Path b:
1. Deploy broken service - see #2 above
2. All next deployments will have error, no matter if fixed or broken

reschex:
Hi - can you give some more detail on how you're deploying? are you using `helm upgrade --install` by any chance? And if you do, what is the state of the deployment when it's broken (`helm ls`) - presumably it is `Failed`?

If this is the case, a `helm delete --purge <deployment>` should do the trick.

sta-szek:
hi, sorry for missing info.
Yes i am using `helm upgrade --install` 
And yes, the deployment stays in `Failed` forever.
[...]

PR #4978, opened 2018-11-28
fix(storage): when pruning release versions, never delete the last deployed revision

ultimateboy:
When tiller-max-history has been set, let's say to 3, if you attempt 3 releases that fail, you end up with a release with no deployed releases.

This doesn't make sense. By changing the prune function to never delete the most recent successfully deployed revision, we never enter a case where a release has no deployed revisions.

See the test case for a good example of this.

PR #7806, opened 2020-03-25
fix(storage): preserve last deployed revision

yurrriq:
Port #4978 to Helm 3 to prevent (but not solve) the problem described in #5595.


yurrriq:
h/t @ultimateboy. Let me know if there's some better way I can give credit.

bacongobbler (approved):

technosophos (approved):

PR #7814, opened 2020-03-25
Add unit test for pkg/chart/chart.go

hs0210:
Signed-off-by: Hu Shuai <hus.fnst@cn.fujitsu.com>

### DIFF
--- a/pkg/chart/chart_test.go
+++ b/pkg/chart/chart_test.go
@@ -78,3 +78,21 @@ func TestSaveChartNoRawData(t *testing.T) {
 
 	is.Equal([]*File(nil), res.Raw)
 }
+
+func TestMetadata(t *testing.T) {
+	chrt := Chart{
+		Metadata: &Metadata{
+			Name:       "foo.yaml",
+			AppVersion: "1.0.0",
+			APIVersion: "v2",
+			Version:    "1.0.0",
+			Type:       "application",
+		},
+	}
+
+	is := assert.New(t)
+
+	is.Equal("foo.yaml", chrt.Name())
+	is.Equal("1.0.0", chrt.AppVersion())
+	is.Equal(nil, chrt.Validate())
+}


jdolitsky (approved):

pkg/chart/chart_test.go
```diff
@@ -78,3 +78,13 @@ func TestSaveChartNoRawData(t *testing.T) {
 
 	is.Equal([]*File(nil), res.Raw)
 }
+
+func TestName(t *testing.T) {
```
hickeyma: Should this be renamed TestMetadata? Then the test would check the rest of the properties of metadata.
I am not sure of the value of just checking the name only.

hickeyma (changes requested):
Thanks for push the test @hs0210. Got some suggestions on improving the test to include other properties of the metadata.

hs0210:
PTAL

pkg/chart/chart_test.go
```diff
@@ -78,3 +78,21 @@ func TestSaveChartNoRawData(t *testing.T) {
 
 	is.Equal([]*File(nil), res.Raw)
 }
+
+func TestMetadata(t *testing.T) {
+	chrt := Chart{
+		Metadata: &Metadata{
+			Name:       "foo.yaml",
+			AppVersion: "1.0.0",
+			APIVersion: "1.0.0",
```
hickeyma: Can you set the `APIVersion` to v2?

hickeyma (changes requested):
Thanks for update @hs0210. Small nit to fix.

hs0210:
PTAL

pkg/chart/chart_test.go
```diff
@@ -78,3 +78,21 @@ func TestSaveChartNoRawData(t *testing.T) {
 
 	is.Equal([]*File(nil), res.Raw)
 }
+
+func TestMetadata(t *testing.T) {
+	chrt := Chart{
+		Metadata: &Metadata{
+			Name:       "foo.yaml",
+			AppVersion: "1.0.0",
+			APIVersion: "2.0.0",
```
hickeyma, line 87: Do you mind setting it as follows: `APIVersion: "v2"`? It is to signify the chart v2 format

hickeyma (changes requested):
Small nit still on the `APIVersion`.

hs0210:
I've changed it. PTAL, Thanks.

hickeyma (approved):
LGTM, thanks @hs0210

PR #7816, opened 2020-03-25
Improve --show-only flag

### LINKED ISSUE #7809: Render only manifests in given directory
It would be nice if the `-s/--show-only` flag of `helm template` could understand `*` or directories. Currently it just returns `Error: could not find template ... in chart`.

Output of `helm version`:
```
```

Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.4", GitCommit:"8d8aa39598534325ad77120c120a22b3a990b5ea", GitTreeState:"clean", BuildDate:"2020-03-12T21:03:42Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.0", GitCommit:"70132b0f130acc0bed193d9ba59dd186f0e634cf", GitTreeState:"clean", BuildDate:"2020-01-14T00:09:19Z", GoVersion:"go1.13.4", Compiler:"gc", Platform:"linux/amd64"}
```

Cloud Provider/Platform (AKS, GKE, Minikube etc.):  `kind`

avorima:
Allows `--show-only` to handle directories and globs.

Fixes #7809


technosophos:
This looks great! Can you please add a test to `template_test.go`. See lines 87-96 for examples. Basically, you create an file in `output/` that has what you expect from the command on `cmd`.

https://github.com/helm/helm/blob/master/cmd/helm/template_test.go#L88-L96

https://github.com/helm/helm/blob/master/cmd/helm/testdata/output/template-show-only-multiple.txt

avorima:
@technosophos I've added 2 test cases, PTAL.

cmd/helm/template.go
```diff
@@ -104,10 +114,19 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 
 							// if the filepath provided matches a manifest path in the
 							// chart, render that manifest
-							if f == manifestPath {
-								manifestsToRender = append(manifestsToRender, manifest)
-								missing = false
+							if strings.HasSuffix(f, "*") {
```
technosophos, line 117: It looks to me like you could remove the `strings.HasSuffix(...)` guard here, and just let all tests go through `filepath.Match`. Then you also wouldn't need your last`if f != manifestPath` check. This would have the added advantage of also allowing `?` and other match characters. Up to you though. I don't think this is blocking.

avorima: Yes, you're right. I added this suffix check to allow specifying directories without glob, but this approach is way more versatile.

technosophos (approved):

adamreese (approved):

PR #7827, opened 2020-03-26
Snapcraft installation instructions added

idvoretskyi:
Helm is available as a snap - https://snapcraft.io/helm; added this to the installation options.

Signed-off-by: Ihor Dvoretskyi <ihor@linux.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ If you want to use a package manager:
 - [Chocolatey](https://chocolatey.org/) users can use `choco install kubernetes-helm`.
 - [Scoop](https://scoop.sh/) users can use `scoop install helm`.
 - [GoFish](https://gofi.sh/) users can use `gofish install helm`.
+- [Snapcraft](https://snapcraft.io/) users can use `snap install helm --classic`
 
 To rapidly get Helm up and running, start with the [Quick Start Guide](https://docs.helm.sh/using_helm/#quickstart-guide).
 


README.md
```diff
@@ -41,6 +41,7 @@ If you want to use a package manager:
 - [Chocolatey](https://chocolatey.org/) users can use `choco install kubernetes-helm`.
 - [Scoop](https://scoop.sh/) users can use `scoop install helm`.
 - [GoFish](https://gofi.sh/) users can use `gofish install helm`.
+- [Snapcraft](https://snapcraft.io/) users can use `snap install helm --classic`
```
hickeyma: This probably should this be: `snap install helm`

hickeyma (changes requested):
Thanks for adding this @idvoretskyi . Small change. Also, PR to update the doc is needed: https://github.com/helm/helm-www/blob/master/content/en/docs/intro/install.md

README.md
bacongobbler: This appears correct according to https://github.com/snapcrafters/helm/blob/master/README.md

I believe the flag is intentional to handle a case in snap’s configuration. Not sure of the reason behind it, but that flag is required to install helm last time I tried.

idvoretskyi: @bacongobbler is correct, this snap requires the `--classic` flag for the "classic" confinement. You may find more details here - https://snapcraft.io/docs/snap-confinement.

PS. It also requires the superuser power to install, but I haven't added `sudo` following the consistency with the installation methods listed above.

bacongobbler (approved):

idvoretskyi:
@hickeyma submitted another PR to the website repo - https://github.com/helm/helm-www/pull/550, PTAL.

@bacongobbler thanks :)

README.md
hickeyma: OK, my bad, thought it was a typo

hickeyma (approved):
LGTM, thanks @idvoretskyi

PR #7831, opened 2020-03-26
ref(*): kubernetes v1.18

adamreese:
Upgrade Kubernetes libraries to v0.18.0


pkg/repo/chartrepo_test.go
```diff
@@ -308,8 +308,8 @@ func TestErrorFindChartInRepoURL(t *testing.T) {
 
 	if _, err := FindChartInRepoURL("http://someserver/something", "nginx", "", "", "", "", g); err == nil {
 		t.Errorf("Expected error for bad chart URL, but did not get any errors")
-	} else if !strings.Contains(err.Error(), `looks like "http://someserver/something" is not a valid chart repository or cannot be reached: Get http://someserver/something/index.yaml`) {
-		t.Errorf("Expected error for bad chart URL, but got a different error (%v)", err)
+	} else if !strings.Contains(err.Error(), `looks like "http://someserver/something" is not a valid chart repository or cannot be reached: Get "http://someserver/something/index.yaml"`) {
```
adamreese: I am not sure why this error changed by adding quotes. ¯\_(ツ)_/¯

adamreese:
I have found an issue with the new kubernetes libraries.  If you run a command without a connection to cluster it will give an error message.

```
helm.go:88: Missing or incomplete configuration info.  Please point to an existing, complete config file:

  1. Via the command-line flag --kubeconfig
  2. Via the KUBECONFIG environment variable
  3. In your home directory as ~/.kube/config

To view or setup config directly use the 'config' command.
```

liggitt:
> If you run a command without a connection to cluster it will give an error message.

That means clients are being created preemptively with missing config before they are needed (e.g. `helm --help` caused construction of a clientset). Something like the last commit on https://github.com/liggitt/helm/commits/kubernetes-1.18 moves the client construction to where it is actually required (at which point an error about the client config being missing would be helpful)

technosophos:
@liggitt We follow SemVer very strictly. We cannot break the API for `pkg/action` until Helm 4. We're doing what is necessary to preserve the API. I am not sure what you're trying to recommend here.

liggitt:
My suggested change does not modify the existing pkg/action API. The existing constructor continues to behave identically.

pkg/action/lazyclient.go
```diff
@@ -0,0 +1,176 @@
...
+
+	// clientFn loads a kubernetes client
+	clientFn func() (*kubernetes.Clientset, error)
+
+	namespace string
+}
+
+func (s *lazyClient) init() error {
+	var err error
+	if s.client == nil {
```
liggitt: this isn't thread-safe… I'd recommend using a sync.Once here

adamreese:
A diff of +295 −56 for a minor version bump ಠ_ಠ

pkg/repo/chartrepo_test.go
adamreese: Fixed in #7835

cmd/helm/root.go
```diff
@@ -92,7 +93,7 @@ func newRootCmd(actionConfig *action.Configuration, out io.Writer, args []string
 			completion.CompDebugln(fmt.Sprintf("About to call kube client for namespaces with timeout of: %d", to))
 
 			nsNames := []string{}
-			if namespaces, err := client.CoreV1().Namespaces().List(metav1.ListOptions{TimeoutSeconds: &to}); err == nil {
+			if namespaces, err := client.CoreV1().Namespaces().List(context.Background(), metav1.ListOptions{TimeoutSeconds: &to}); err == nil {
```
joelanford: Should this use `context.WithTimeout`?

```suggestion
			ctx, cancel := context.WithTimeout(context.Background(), time.Second * time.Duration(to))
			defer cancel()
			if namespaces, err := client.CoreV1().Namespaces().List(ctx, metav1.ListOptions{TimeoutSeconds: &to}); err == nil {
```

And for other places that `context.Background()` is used, is it truly the intention for those functions to be allowed to wait forever? Or should we use `context.TODO()` as a reminder that we plan to add cancellation?

pkg/kube/client.go
```diff
@@ -563,9 +569,12 @@ func scrubValidationError(err error) error {
...
 // and returns said phase (PodSucceeded or PodFailed qualify).
 func (c *Client) WaitAndGetCompletedPodPhase(name string, timeout time.Duration) (v1.PodPhase, error) {
-	client, _ := c.Factory.KubernetesClientSet()
+	client, err := c.Factory.KubernetesClientSet()
+	if err != nil {
+		return v1.PodUnknown, err
+	}
 	to := int64(timeout)
-	watcher, err := client.CoreV1().Pods(c.namespace()).Watch(metav1.ListOptions{
+	watcher, err := client.CoreV1().Pods(c.namespace()).Watch(context.Background(), metav1.ListOptions{
```
joelanford: Ditto: `context.WithTimeout()`?

joelanford (commented):
Looking forward to this! 🎉  

With the client-go changes in 1.18, just a few questions about how to add contexts to all the function calls.

cmd/helm/root.go
bacongobbler: to retain backwards compatibility, yes.

adamreese: We use semver which is why all these workarounds are necessary.

pkg/kube/client.go
adamreese: same as above

technosophos (approved):
LGTM

pkg/action/lazyclient.go
```diff
@@ -0,0 +1,180 @@
...
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/kubernetes"
+	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+)
+
+type lazyClient struct {
```
jdolitsky, line 31: why is it called lazyClient? because it only runs init() once used?

technosophos: It only initializes the underlying client when it must contact the API server.

Kubernetes 1.18 introduced a breaking change in the way they load the kubeconfig. It now contacts the API server before any user-requested operations are executed.

Helm has traditionally loaded the Kubeconfig very early, and passed the client around from outside via the normal dependency injection model. With this change, we had to figure out a way to not change or break the way the Helm client is created, but defer the underlying Kubernetes initialization until as late as possible. We did this by "lazy loading" the client in the `lazyClient`.

This was all done in the name of preserving the same API for Helm API users (as opposed to the other suggestion earlier in the thread, which would have required downstream users to rewrite their code as well).

bacongobbler (approved):

technosophos:
I am manually testing against 1.17 for backward compat

technosophos:
Tested on 1.17 and it works fine

pkg/action/lazyclient.go
liggitt: > Kubernetes 1.18 introduced a breaking change in the way they load the kubeconfig. It now contacts the API server before any user-requested operations are executed.

It actually doesn't contact the server, but it does stop silently defaulting to an insecure configuration when asked to load a non-existent configuration.

technosophos: You are right. Net result: Same thing.

adamreese: Changing behavior is a breaking change.  Unlike Kubernetes, Helm follows semantic versioning.  We need to figure out work arounds when Kubernetes breaks backwards compatibility.

pkg/action/lazyclient.go
```diff
@@ -0,0 +1,180 @@
...
+	clientFn func() (*kubernetes.Clientset, error)
+
+	namespace string
+}
+
+func (s *lazyClient) init() error {
+	var err error
+	s.initClient.Do(func() {
+		if s.client == nil {
+			s.client, err = s.clientFn()
```
liggitt, line 46: lazyClient should probably cache any client initialization error as well… if init() errors on first call, subsequent calls will return nil, but s.client will still be nil, leading to a NPE

adamreese:
Added some comments about why all these workarounds were needed

marckhouzam:
This is awesome!  
For people using kind, the early loading of the kubeconfig has been an annoyance for a while.
Furthermore, the early loading also forced me to include special handling for completion, which can now be removed (I'll do it in a follow-up PR).   Specifically, this code can soon be removed:
https://github.com/helm/helm/blob/c12a9aee02ec07b78dce07274e4816d9863d765e/cmd/helm/helm.go#L76-L84

technosophos (approved):
Re-reviewed with comments and fix to client loader. LGTM

tamalsaha:
FYI, https://github.com/kubernetes/kubernetes/issues/90074#issuecomment-614191423

Issue #7807, opened 2020-03-25
Install flag of the Upgrade action is not working on the sdk


Nitive:
### Not working code example

**Code**
```go
upgrade := action.NewUpgrade(actionConfig)
upgrade.Install = true

upgrade.Run("release", chart, values)
```

**Expected result:** chart deployed
**Actual result:** `Error: release has no deployed releases`

It's a rough example, but it seems to be enough for understanding the problem.
Happy to provide repository with minimal but complete code needed for reproduction if there is need for it.

### Cause of bug

This happens because install flag logic implemented in “main” package and tightly coupled with cli:
https://github.com/helm/helm/blob/193850a9e2c509acf1a499d98e8d23c12c134f11/cmd/helm/upgrade.go#L93-L97

---

Helm sdk version: 3.1.2

liuming-tech:
> Actual result: `Error: release has no deployed releases`
this message tells us that the release exists, but status is not deployed.

`helm upgrade --install` install flag will work when no release exists.

Nitive:
@liuming-tech cli is working fine, the issue is with go sdk usage

technosophos:
The `Install` flag on `action/upgrade.go` does not do an install if the release is not present. See `cmd/upgrade.go` to see how `helm upgrade --install` works.

The `action`'s `Upgrade.Install` flag could definitely use some documentation. That is confusing.

Nitive:
@technosophos, I can see that it works that way but there is two problems with that

1. To make `upgrade --install` from go sdk I have to repeat `cmd/upgrade.go` logic in my application
2. Like you said, current behavior is confusing and do not align with other actions

I could change the code to respect Install flag and create a PR, but it would be a breaking change. Minor one, but still. If I do create a PR, would it be considered to be a part of helm v3 or it have to be postponed until helm v4?

bridgetkromhout:
Hi, @Nitive - as you may have seen, we added an inline explanation in https://github.com/helm/helm/pull/7834. 

The Helm project does follow strict SemVer so if we are going to make a breaking change, it will be in Helm v4. I'll resolve this issue at this time, and if you decide that working on v4 is something you'd like to do, we'd welcome your participation. Thanks!

Issue #7811, opened 2020-03-25
Documentation request about --cleanup-on-fail, --atomic, --force


consideRatio:
I'd love to gain clarity on the flags `--cleanup-on-fail`, `--atomic`, and `--force` in `helm3 upgrade`. What is it that they do if used by themselves, and will this change if they are used together with another flag?

```
      --atomic                   if set, upgrade process rolls back changes made in case of failed upgrade. The --wait flag will be set automatically if --atomic is used
      --cleanup-on-fail          allow deletion of new resources created in this upgrade when upgrade fails
      --force                    force resource updates through a replacement strategy
```

## Questions
1. Will `--force` use a replacement strategy for all resources no matter what?
2. Will `--atomic` imply `--cleanup-on-fail`?
3. In `--cleanup-on-fail`, what does it mean to `allow deletion` of new resources on failure? Will these resources be marked as deletable for the next upgrade attempt? Will this flag imply `--wait` and then automatically delete new resources not existing before on an upgrade failure?
4. In `--cleanup-on-fail`, what are considered `new resources`?
5. Does `--atomic` imply `--cleanup-on-fail`? What does `--cleanup-on-fail` add to `--atomic`?

technosophos:
In all cases I could find, if a flag implies another flag, it says so in the documentation.

1. Force will use a replacement strategy on all resources that it finds existing already. Replacement means: If resource X exists, and the inbound version of X is different than the version in the cluster, and the two conflict, force the new X to overwrite the old X. This is the same as `kubectl apply --force-conflicts`, IIRC.
2. Atomic does not imply cleanup-on-fail
3. There are cases where an upgrade creates a resource that was not present in the last release. Setting this flag allows Helm to remove those new resources if the release fails. The default is to not remove them (Helm tends to avoid destruction-as-default, and give users explicit control over this)
4. If X is created afresh (as opposed to upgrading an existing resource), it is considered new.
5. Cleanup-on-fail deletes items that were marked as new. In some rare cases, it would be possible for a resource to be successfully created and then be orphaned by a rollback (if --atomic is set). If you are worried about those cases, then you could set --atomic and --cleanup-on-fail both.

consideRatio:
Thank you for putting your mind into this @technosophos !

## Ideas on docs improvements
1. Add *all* to `--force` description: force *all* resource updates through a replacement strategy

---

> There are cases where an upgrade creates a resource that was not present in the last release. Setting this flag allows Helm to remove those new resources if the release fails.

[...]

PR #7834, opened 2020-03-28
docs: Update inline docs on action/upgrade.go

technosophos:
See #7807 for an example of how the absence of docs confused a user.

### DIFF
--- a/pkg/action/upgrade.go
+++ b/pkg/action/upgrade.go
@@ -79,9 +79,9 @@ type Upgrade struct {
 	Recreate bool
 	// MaxHistory limits the maximum number of revisions saved per release
 	MaxHistory int
-	// Atomic, if true, will set the upgrade mode to an atomic upgrade.
+	// Atomic, if true, will roll back on failure.
 	Atomic bool
-	// CleanupOnFail will, if true, cause the upgrade to delete resources on a failed update.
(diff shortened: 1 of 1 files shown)


technosophos:
Also tried to clarify vis a vis #7811

pkg/action/upgrade.go
```diff
@@ -42,27 +42,53 @@ type Upgrade struct {
...
+	MaxHistory int
+	// Atomic, if true, will roll back on failure.
+	Atomic bool
+	// CleanupOnFail will, if true, cause the upgrade to delete newly-created resources on a failed update.
+	CleanupOnFail bool
+	// SubNotes determines whether sub-notes are rendered in the chart.
+	SubNotes bool
+	// Description is the description of this operation
+	Description  string
+	PostRenderer postrender.PostRenderer
```
hickeyma: Might be worth adding a description for the post renderer also?

hickeyma (changes requested):
Thanks for adding the descriptions @technosophos. Looking good, just missing a description.

pkg/action/upgrade.go
technosophos: Can I please get a re-review on this?

jdolitsky (approved):

hickeyma (approved):
LGTM, thanks @technosophos

PR #7835, opened 2020-03-28
fix: update unit test for go 1.14 error string change

technosophos:
I accidentally updated to Go 1.14, and upon doing so my unit tests started failing. I traced this down to a change in the error coming back from the `http.Client`'s `Do()` method. I updated the error message in a way that is backward compatible to previous Go versions, but also works with Go 1.14 and beyond.

### DIFF
--- a/pkg/getter/httpgetter.go
+++ b/pkg/getter/httpgetter.go
@@ -18,7 +18,6 @@ package getter
 import (
 	"bytes"
 	"crypto/tls"
-	"fmt"
 	"io"
 	"net/http"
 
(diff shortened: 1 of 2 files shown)


hickeyma (approved):
LGTM, thanks @technosophos

technosophos:
I discussed with @adamreese, and we decided that changing tests to test the string value of Go's built-in was unnecessary. This fix merely tests that we return the right error message from Helm code.

technosophos:
Note -- also discovered this impacts Helm 2, so I am fixing there as well.

adamreese (approved):

Issue #6324, opened 2019-08-30
How can I push chart to an insecure OCI registry with helm v3


ywk253100:
How can I push chart to an insecure registry(HTTP or HTTPS with self-signed certificate) with helm v3?

Should we provide an option to skip the security checking? It's useful in developing/testing environments

jdolitsky:
@ywk253100 Hi, we need to document this, but try this:
```
export HELM_EXPERIMENTAL_OCI=1
helm registry login mysite.io --insecure
```
Please let me know if works for you

ywk253100:
It works for `helm registry login`, but not for `helm chart push`

jdolitsky:
@ywk253100 interesting. Can you please share the output of push command with `--debug` ?

ywk253100:
```
helm chart push --insecure --debug 192.168.0.1/library/harbor:1.1.2
Error: unknown flag: --insecure
helm.go:81: [debug] unknown flag: --insecure
```

ywk253100:
```
yinw-a01:helm yinw$ helm3 version
version.BuildInfo{Version:"v3.0.0-beta.2", GitCommit:"26c7338408f8db593f93cd7c963ad56f67f662d4", GitTreeState:"clean", GoVersion:"go1.12.9"}
```

jdolitsky:
@ywk253100 The `--insecure` flag is only for login. So you login once, then on push that registry is marked as insecure. Can you try
```
export HELM_EXPERIMENTAL_OCI=1
helm registry login 192.168.0.1 --insecure
helm chart push --debug 192.168.0.1/library/harbor:1.1.2
```
?

ywk253100:
The registry serves on HTTP, does this work for HTTP?
```
yinw-a01:helm yinw$ helm3 registry login 10.193.11.7 --insecure
Username: admin
Password:
INFO[0006] Error logging in to v2 endpoint, trying next endpoint: Get https://10.193.11.7/v2/: dial tcp 10.193.11.7:443: connect: connection refused
Login succeeded
yinw-a01:helm yinw$ helm3 chart push --debug 10.193.11.7/library/harbor:1.1.2
The push refers to repository [10.193.11.7/library/harbor]
ref:     10.193.11.7/library/harbor:1.1.2
digest:  0f8c0650d55f5e00d11d7462381c340454a3b9e517e15a0187011dc305690541
size:    28.1 KiB
name:    harbor
version: 1.1.2
[...]

PR #7844, opened 2020-03-31
Implement chart registry TLS configuration

Foxboron:
* Implements --insecure-skip-tls-verify
* Implements custom cert, key and ca file

Solves https://github.com/helm/helm/issues/6324

I have not added documentation nor tests as I reckon someone should look over the implementation first.

### DIFF
--- a/cmd/helm/chart.go
+++ b/cmd/helm/chart.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/spf13/cobra"
 
+	"helm.sh/helm/v3/internal/experimental/registry"
 	"helm.sh/helm/v3/pkg/action"
 )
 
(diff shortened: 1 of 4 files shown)


jdolitsky (changes requested):
This looks great!

Would you be able to modify the tests to use https plus these options? https://github.com/helm/helm/blob/master/internal/experimental/registry/client_test.go

You would have to construct the Docker registry with extra config (check for `registry.NewRegistry` in that file above). Here is where Docker parses the config to listen w TLS: https://github.com/docker/distribution/blob/master/registry/registry.go#L139

Here is the docker docs, but the Go fields may be slightly different than the yaml config: https://docs.docker.com/registry/configuration/#http

Foxboron:
I started working a bit on the tests and it's a bit weird. The resolver is initialized in the setup of the suite, thus omits the added code that deals with initializing the http client.

I'm unsure what we would be accomplishing by just adding the TLS enabled client in the setup of the suite, as this should be covered by oras and containerd test suites already? 


https://github.com/helm/helm/blob/master/internal/experimental/registry/client_test.go#L70

Foxboron:
OCI support is being changed which deprecated this pull-request.

PR #7848, opened 2020-03-31
fix: fixes for Go 1.14

technosophos:
This makes a few backward-compatible (works with < Go 1.14) fixes to testing.

Go 1.14 changed some error messages, and this caused some unit tests to break. In addition, the formatting rules changed in Go 1.14, which caused the proto.go files to fail style checks. Since there is not really any reason to check the style of generated files, I removed them from `make test-style`.


adamreese (approved):

adamreese:
LGTM

PR #7850, opened 2020-04-01
Post Render args example

### LINKED ISSUE #7849: Proposal: Post-Render Args or Env Vars
Post render functionality implemented in #7260 is awesome (thanks @thomastaylor312).  I (want to) manage my helm charts with Helmfile and Kustomize and don't want to rely on having to necessarily parse manifests from `stdin` or duplicate the render script for each app, instead favouring something simple and generic like so (thanks again @thomastaylor312 😆 ):
```
#!/bin/bash
set -eu
cat <&0 > $1/all.yaml
kustomize build $1 && rm $1/all.yaml
```

With an example file structure like:
```
helmfile.yaml
redis/kustomization.yaml
redis/patch.yaml
app2/...
app3/...
```

example `kustomization.yaml`:
```
resources:
- all.yaml
patches:
- patch.yaml
```

I therefore propose either being able to pass arguments or environment variables to post render execution, thus avoiding [some others kinds of hackiness](https://github.com/roboll/helmfile#helmfile--kustomize)

acaire:
WIP - Proof of concept to fix #7849

### DIFF
--- a/pkg/postrender/exec.go
+++ b/pkg/postrender/exec.go
@@ -21,29 +21,34 @@ import (
 	"io"
 	"os/exec"
 	"path/filepath"
+	"strings"
 
 	"github.com/pkg/errors"
 )
 
 type execRender struct {
 	binaryPath string
+	args       []string
 }
 
 // NewExec returns a PostRenderer implementation that calls the provided binary.
 // It returns an error if the binary cannot be found. If the path does not
 // contain any separators, it will search in $PATH, otherwise it will resolve
 // any relative paths to a fully qualified path
-func NewExec(binaryPath string) (PostRenderer, error) {
-	fullPath, err := getFullPath(binaryPath)
+// Any passed arguments are joined by commas, and sent as one argument
+func NewExec(binaryPathAndArgs string) (PostRenderer, error) {
+	binaryPathAndArgsCsv := strings.Split(binaryPathAndArgs, ",")
+	args := []string{strings.Join(binaryPathAndArgsCsv[1:], ",")}
+	fullPath, err := getFullPath(binaryPathAndArgsCsv[0])
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


technosophos:
What is the status of this? It has been marked WIP for almost two months with no activity. Is the intention to carry this on to completion?

acaire:
It was as complete as it needed to be when submitted for the related issue which I waited almost two months for an answer on, so yes the intention is to complete it

technosophos:
When it is marked WIP (Work In Progress), we do not put it in the review queue. I will remove the label and modify the title.

Please respect the fact that there are few of us and many, many issues/PRs. We work as fast as we can, but it is quite simply more work than we can handle.

acaire:
I only got a response 7 days ago, and I haven’t had a chance to implement what has been requested, so for all intents and purposes this PR is still WIP.  I have no qualms about the timeliness of this process, but I’d appreciate if you were less curt.

technosophos (changes requested):
The `,` character is a legal character in paths. So, for example, I can have an executable named `foo,bar.sh`. The current PR breaks with names like this:

```
$ helm install --dry-run redis stable/redis --post-renderer ./foo,bar.sh,prod,redis 
Error: invalid argument "./foo,bar.sh,prod,redis" for "--post-renderer" flag: unable to find binary at ./foo: exec: "./foo": stat ./foo: no such file or directory
```

Also, unit tests will be required for final acceptance.

acaire:
A different implementation was requested as visible in the linked issue, so comma splitting won’t be necessary. Thanks for the review, I’ll let you know when it’s ready to review.

acaire:
Unfortunately I no longer have the requirement or bandwidth to implement this, closing in favour of [improved helmfile support](https://github.com/roboll/helmfile/pull/1172)

PR #7851, opened 2020-04-01
Read stdin only one time (fix for #7002).

### LINKED ISSUE #7002: Support reading values from STDIN for `upgrade` command

Yannig:
Read stdin only if not already been readed.
Avoid a bug where we read multiple times stdin when using `upgrade` with `--install` option. Indeed, the second time, the standard entry is empty.

Fix: https://github.com/helm/helm/issues/7002

### DIFF
--- a/pkg/cli/values/options.go
+++ b/pkg/cli/values/options.go
@@ -104,10 +104,20 @@ func mergeMaps(a, b map[string]interface{}) map[string]interface{} {
 	return out
 }
 
+var stdin []byte = nil
+
 // readFile load a file from stdin, the local directory, or a remote file with a url.
 func readFile(filePath string, p getter.Providers) ([]byte, error) {
 	if strings.TrimSpace(filePath) == "-" {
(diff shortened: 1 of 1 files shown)


hickeyma:
@Yannig Is there a reason why you pushed this PR when PR #7640 is in progress to fix #7002 ?

Yannig:
@hickeyma no sorry. I didn't realised that there was already a fix. You can close it as you wish.

hickeyma:
@Yannig I'll close it then. You can add your comments to #7640

PR #7863, opened 2020-04-03
fix(template): fix the bug when helm template with flag --output-dir

### LINKED ISSUE #7836: [helm3] helm template with --output-dir doesn't write template with a hook to file
Output of `helm version`:
3.1.2

Output of `kubectl version`:
1.18.0

CentOS Linux release 7.7

The same problem still persists in Helm 3.1.2.
[https://github.com/helm/helm/issues/7001](url)

If a template contains hook annotation `helm template --output-dir mydir` works incorrectly.
It does not write generated text for template with a hook to the file.
Instead it prints the output to the console.

**Example**
```
cat init-job.yaml

apiVersion: batch/v1
kind: Job
metadata:
  annotations:
    "helm.sh/hook": "post-install"
    "helm.sh/hook-weight": "0"
    "helm.sh/hook-delete-policy": "before-hook-creation"
  name: init
  ...

helm template --debug --output-dir /test/out -f values-my.yaml

install.go:158: [debug] Original chart version: ""
install.go:175: [debug] CHART PATH: /test/mychart

wrote /test/out/secret.yaml
wrote /test/out/service.yaml
...

---
# Source: init-job.yaml
apiVersion: batch/v1
kind: Job
metadata:
  annotations:
    "helm.sh/hook": "post-install"
    "helm.sh/hook-weight": "0"
    "helm.sh/hook-delete-policy": "before-hook-creation"
  name: init
  ...
```

donggangcj:
Issue:helm template with --output-dir doesn't write template with a hook to
file

Code Detail:
1. move the `writeToFile` function from `action/install` to `helm/template`
2. add function `parseManifest` to parse the `Release` field Manifest.

Close #7836

### DIFF
--- a/cmd/helm/template.go
+++ b/cmd/helm/template.go
@@ -20,6 +20,8 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"os"
+	"path"
 	"path/filepath"
 	"regexp"
 	"sort"
@@ -34,6 +36,8 @@ import (
 	"helm.sh/helm/v3/pkg/cli/values"
 	"helm.sh/helm/v3/pkg/releaseutil"
 )
+
+const defaultDirectoryPermission = 0755
 
 const templateDesc = `
 Render chart templates locally and display the output.
@@ -77,10 +81,50 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 			if rel != nil {
 				var manifests bytes.Buffer
 				fmt.Fprintln(&manifests, strings.TrimSpace(rel.Manifest))
-
-				if !client.DisableHooks {
-					for _, m := range rel.Hooks {
-						fmt.Fprintf(&manifests, "---\n# Source: %s\n%s\n", m.Path, m.Manifest)
+				newDir := client.OutputDir
+				if newDir != "" {
+					// Aggregate all valid manifests into one big doc.
+					fileWritten := make(map[string]bool)
+					if client.UseReleaseName {
+						newDir = filepath.Join(client.OutputDir, client.ReleaseName)
+					}
+					if client.IncludeCRDs {
+						for _, crd := range rel.Chart.CRDObjects() {
+							err = writeToFile(newDir, crd.Filename, string(crd.File.Data[:]), fileWritten[crd.Name])
+							if err != nil {
+								return err
+							}
+							fileWritten[crd.Name] = true
+						}
+					}
+					if rel.Manifest != "" {
+						splitManifests := releaseutil.SplitManifests(rel.Manifest)
+						for _, v := range splitManifests {
+							name, content, err := parseManifest(v)
+							if err != nil {
+								return err
+							}
+							err = writeToFile(newDir, name, content, fileWritten[name])
+							if err != nil {
+								return err
+							}
+							fileWritten[name] = true
+						}
+					}
+					if !client.DisableHooks {
+						for _, m := range rel.Hooks {
+							err := writeToFile(newDir, m.Path, m.Manifest, false)
+							if err != nil {
+								return err
+							}
+						}
+					}
+					manifests.Reset()
+				} else {
+					if !client.DisableHooks {
+						for _, m := range rel.Hooks {
+							fmt.Fprintf(&manifests, "---\n# Source: %s\n%s\n", m.Path, m.Manifest)
+						}
 					}
 				}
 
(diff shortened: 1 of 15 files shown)


cmd/helm/template.go
```diff
@@ -77,10 +81,50 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 			if rel != nil {
 				var manifests bytes.Buffer
 				fmt.Fprintln(&manifests, strings.TrimSpace(rel.Manifest))
-
-				if !client.DisableHooks {
-					for _, m := range rel.Hooks {
-						fmt.Fprintf(&manifests, "---\n# Source: %s\n%s\n", m.Path, m.Manifest)
+				newDir := client.OutputDir
```
technosophos, line 84: This section should probably be put in a new function or several new functions. The cyclomatic complexity of newTemplateCmd is already too high, and this makes it _really_ hard to read.

cmd/helm/template.go
```diff
@@ -154,3 +198,58 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
...
+
+	defer f.Close()
+
+	_, err = f.WriteString(fmt.Sprintf("---\n# Source: %s\n%s\n", name, data))
+
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("wrote %s\n", outfileName)
```
technosophos, line 224: Is this left-over debugging output, or is this supposed to be here? If it is intended to be here, it should use `fmt.Fprintf()` and send the results to whatever the command declares as `out`. This allows other things to override STDOUT with another ouput.

technosophos (changes requested):
This appears to remove substantial functionality from the `pkg/action` package. We can't do this, because that package is an externally-consumed package that other applications use. If we just remove the functionality from there, then existing implementations of Helm tools will break.

So I think that the implementation must continue to use `pkg/action` to write the files until we can change the behavior in Helm 4.

cmd/helm/template.go
donggangcj: Thanks! This line is used to debug.I will polish it with `fmt.Fprintf()` and forgive me for  my bad english.😄

donggangcj: Thanks! This line is used to debug.Is it better to use [`debug()`](https://github.com/helm/helm/blob/7f3339cb4eec9800cb09a46512050fb509480f7e/cmd/helm/helm.go#L55) instead of `fmt.FprintF()`? Forgive me for  my bad english.😄

donggangcj:
> This appears to remove substantial functionality from the `pkg/action` package. We can't do this, because that package is an externally-consumed package that other applications use. If we just remove the functionality from there, then existing implementations of Helm tools will break.
> 
> So I think that the implementation must continue to use `pkg/action` to write the files until we can change the behavior in Helm 4.

Yes, Other applications which depend on `pkg/action` may be break and i will keep the behavior of `pkg/action` not

cmd/helm/template.go
donggangcj: All right! I will refactor this function to reduce its complexity .

donggangcj:
> This appears to remove substantial functionality from the `pkg/action` package. We can't do this, because that package is an externally-consumed package that other applications use. If we just remove the functionality from there, then existing implementations of Helm tools will break.
> 
> So I think that the implementation must continue to use `pkg/action` to write the files until we can change the behavior in Helm 4.

If a appliaction use `pkg/action install` to realize the feature like template outputing,it will break.I think it is
very important to decouple template and install in helm4 。I will try to fix this issue under the premise that not change the behavior
of the package `pkg/action`.

bacongobbler:
> I think it is very important to decouple template and install in helm4

Why do you think that? The functionally is nearly identical.

donggangcj:
@bacongobbler  @technosophos  I reopen a PR to keep on because these commits have been a while.

cmd/helm/template.go
donggangcj: This line is aimed to print some help information  
```bash
➜  myapp helm template  --output-dir tes . --set ingress.enabled=true --no-hooks
wrote tes/myapp/templates/serviceaccount.yaml
wrote tes/myapp/templates/service.yaml
wrote tes/myapp/templates/deployment.yaml
wrote tes/myapp/templates/ingress.yaml
```

donggangcj:
@bacongobbler  The `template` is separated from  `install` ，and the `output` feature  belong to `template` command. will make their responsibilities clearer. 😀

bacongobbler:
#7836 was fixed with #8156. Closing this.

PR #7767, opened 2020-03-13
fix: rollback from uninstalled release with no error message

liuming-dev:
**helm version**
```bash
~/.kube » helm version
version.BuildInfo{Version:"v3.1.2", GitCommit:"d878d4d45863e42fd5cff6743294a11d28a9abce", GitTreeState:"clean", GoVersion:"go1.13.8"}
```

**situation**
I have a deployed release
```
~/.kube » helm list -n testzy
NAME           	NAMESPACE	REVISION	UPDATED                             	STATUS  	CHART                        	APP VERSION
helm-app-test-a	testzy   	3       	2020-03-27 17:45:02.329406 +0800 CST	deployed	helm-app-sample-1.0.0        	1.0.3
```

uninstall release
```
~/.kube » helm uninstall helm-app-test-a -n testzy --keep-history
release "helm-app-test-a" uninstalled

~/.kube » helm list -n testzy -a
NAME           	NAMESPACE	REVISION	UPDATED                             	STATUS     	CHART                        	APP VERSION
helm-app-test-a	testzy   	3       	2020-03-27 17:45:02.329406 +0800 CST	uninstalled	helm-app-sample-1.0.0        	1.0.3
```

rollback to former revision, get a error message, and the current release status is set to `pending-rollback`
```
~/.kube » helm rollback helm-app-test-a 1 -n testzy
Error: "helm-app-test-a" has no deployed releases

~/.kube » helm list -n testzy -a
NAME           	NAMESPACE	REVISION	UPDATED                             	STATUS          	CHART                        	APP VERSION
helm-app-test-a	testzy   	4       	2020-03-27 18:52:45.08371 +0800 CST 	pending-rollback	helm-app-sample-1.0.0        	1
```

but the resources of the release has been created, and their's status are ok
```
~/.kube » kubectl -n testzy get deployment helm-app-test-a
NAME              READY   UP-TO-DATE   AVAILABLE   AGE
helm-app-test-a   1/1     1            1           5m1s
```

therefore, i think helm allows user to rollback from uninstalled release, this is a bug

this PR fixs the bug, and it will rollback successfully without error message.

fixs #6978, #7366 

Please help to review the code, thanks.

PR #7867, opened 2020-04-03
fix(storage): Deployed, DeployedAll, and History methods should return correct errors.

jmontroy90:
The `Deployed`, `DeployedAll`, and `History` methods all claim in their docstrings to return `ErrReleaseNotFound` when there are no matching releases, but their actual behavior is not correct:

1. `History` is dependent on the underlying implementation, which as #7372 points out below, is inconsistent between `ConfigMap` / `Secret` vs. `Memory`.
2. `Deployed` and `DeployedAll` return entirely custom errors.

This PR makes the `Storage` client consistent in behavior, regardless of the underlying `Driver` implementations:

1. All three methods now return `ErrReleaseNotFound` when `len(ls) == 0`. 
2. The test `TestStorageDeployed` in `storage_test.go` was calling `Last`, not `Deployed`. That's fixed.

I also added a tiny error test. I'd love to move these storage tests to table tests in a subsequent PR.

This is a breaking change for anyone who uses the `Storage` client with any driver, but it's breaking in that now at least the code and docstrings match.

I searched and found that this is adjacent to https://github.com/helm/helm/pull/7372, but gets at the problem at the client level, rather than the implementations. I also took the testing fix from that, so that change should be changed first, in my opinion.

cc @absoludity @latiff @SimonAlling

### DIFF
--- a/cmd/helm/rollback_test.go
+++ b/cmd/helm/rollback_test.go
@@ -23,7 +23,7 @@ import (
 	"helm.sh/helm/v3/pkg/release"
 )
 
-func TestRollbackCmd(t *testing.T) {
+func makeTestReleases() []*release.Release {
 	rels := []*release.Release{
 		{
 			Name:    "funny-honey",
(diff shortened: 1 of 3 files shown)


pkg/action/rollback.go
```diff
@@ -211,7 +213,7 @@ func (r *Rollback) performRollback(currentRelease, targetRelease *release.Releas
 	}
 
 	deployed, err := r.cfg.Releases.DeployedAll(currentRelease.Name)
-	if err != nil {
+	if err != nil && err != driver.ErrReleaseNotFound {
```
luisdavim, line 216: are you sure you want to ignore the not found errors here?

pkg/action/upgrade.go
```diff
@@ -130,7 +132,7 @@ func (u *Upgrade) prepareUpgrade(name string, chart *chart.Chart, vals map[strin
 
 	// finds the deployed release with the given name
 	currentRelease, err := u.cfg.Releases.Deployed(name)
-	if err != nil {
+	if err != nil && err != driver.ErrReleaseNotFound {
```
luisdavim, line 135: are you sure you want to ignore the not found errors here?

pkg/action/rollback.go
jmontroy90: Yea, the tests need to be fixed for this.

pkg/action/upgrade.go
jmontroy90: Yea, the tests need to be fixed for this.

jmontroy90: Fixed per #7372's handling of mutable data.

pkg/action/rollback.go
jmontroy90: Fixed per #7372's handling of mutable data.

liuming-tech:
cc #7767

technosophos:
I am not sure this constitutes a non-breaking change. It would be less disruptive to change the docstrings, I am afraid.

technosophos:
@mattfarina what is your opinion on whether changing error types/messages constitutes breaking changes?

PR #7874, opened 2020-04-04
fix: Ignore local path when `--repo` is provided

### LINKED ISSUE #7862: Bug: `--repo` should takes precedence over local folder
Using local chart when a `--repo` flag is provided doesn't make sense, the assumption in [LocateChart](https://github.com/helm/helm/blob/c20d06d930aad1862b73aebdfbf36e1f00e75d28/pkg/action/install.go#L726-L795) function is not correct in this case,

Is it possible to ignore the local folders when `--repo` is provided?

Output of `helm version`:
```
```

mforutan:
When `--repo` flag is provided, local path should be ignored to locate a chart.

Closes #7862


pkg/action/install.go
```diff
@@ -737,20 +737,22 @@ func (c *ChartPathOptions) LocateChart(name string, settings *cli.EnvSettings) (
...
+				return abs, err
 			}
+			if c.Verify {
+				if _, err := downloader.VerifyChart(abs, c.Keyring); err != nil {
+					return "", err
+				}
+			}
+			return abs, nil
+		}
+		if filepath.IsAbs(name) || strings.HasPrefix(name, ".") {
```
marckhouzam, line 753: Should we still check for an absolute or relative path and error out when `--repo` is specified?

mforutan: It doesn't make sense to me to check the path when RepoURL is provided as `--repo` only accepts something remote with a protocol (http, https, s3, ...), the only thing I can see will be different is that with this change, it may allow a chart name in a remote repository to start with `.`, but not sure how much an issue that could be.

marckhouzam: You're right. I was just thinking about a user error in specifying the arguments, but then the error message we currently have wouldn't even be correct.  Let's just leave it as you have it.

mforutan:
@marckhouzam What is the next step to get this PR merged?

marckhouzam:
@mforutan Would you be able to write a test for this case?

mforutan:
> @mforutan Would you be able to write a test for this case?

I tried to find an existing test for `LocateChart` function to create a new one for this case, but I wasn't able to find any, not sure how can I create a new unit test for this function due to its dependencies, can you please let me know where is the best place to test this case?

mforutan:
@marckhouzam, I spent some time to get a better understanding of the test cases and their coverage, now I think I know a little better. At the first glance, looks like the best place to add a test for this case is inside [pull_test.go](https://github.com/helm/helm/blob/master/cmd/helm/pull_test.go#L44-L142), then I realized that `pull` logic doesn't use the same function to [locate a chart](https://github.com/helm/helm/blob/master/pkg/action/pull.go#L90-L96) which is to me inconsistent behavior and probably that is why this issue hadn't been discovered before. Also, all of the existing tests in [install_test.go](https://github.com/helm/helm/blob/master/cmd/helm/install_test.go) and [upgrade_test.go](https://github.com/helm/helm/blob/master/cmd/helm/upgrade_test.go) are written based on a local chart so not an easy change to add remote repository for tests.

I believe the best approach should be changing the `pull.go` logic to use the same function and then write some tests for `LocateChart` function itself or even moving the `LocateChart` to a new package with some tests. I will try to propose a change for it, but really appreciate some inputs from everyone else.

technosophos:
Possibly an easy way to test it would be to break out the logic into its own function, then write a test against that function.

mforutan:
@marckhouzam I added a new test to cover this issue.

mforutan:
@marckhouzam @technosophos do I need anything else for this?

mforutan:
@mattfarina @bacongobbler can you guys please have a look into this PR?

technosophos:
Unless @mattfarina can show otherwise, this is a SemVer breaking change, and is deferred until Helm 4. For the record, I do agree that this is better UX.... but it still breaks our SemVer policy.

technosophos:
Oh! I did think of another option. We could mark `--repo` as deprecated, and add another flag that instituted this new behavior. That would get us around the SemVer constraints, and allow us to do what (I still agree) is the better UX. /cc @mattfarina

mattfarina:
I have to concur with @technosophos on this.

We want to be super careful about breaking changes.

joejulian:
Is this abandoned?

joejulian:
closing this as abandoned

mforutan:
@joejulian From my understanding, the initial issue still exists: https://github.com/helm/helm/issues/7862 but there is an argument about treating it as a bug or UX improvement!

joejulian:
Ok, happy to reopen it. I asked a few months ago but got no response.

mforutan:
> Ok, happy to reopen it. I asked a few months ago but got no response.

Sorry, I didn't notice it.

joejulian:
I'll offer the same observation as #12252 that's covering the same issue. This changes the default behavior. Changes of behavior are precluded without opting in by [hip-0004](https://github.com/helm/community/blob/main/hips/hip-0004.md).

I think a more correct solution would be to split the function into two parts. A local part and a remote part. The `--repo` flag would still call `LocateChart` which would do the local function first just like it does now. If it gets past that, it would call a remote version, ie. `LocateRemoteChart`.

Then I'd add a new flag, `--remote-repo`, that calls `LocateRemoteChart` instead of `LocateChart`.

`LocateChart` would look something like:
```
func (c *ChartPathOptions) LocateChart(name string, settings *cli.EnvSettings) (string, error) {
	name = strings.TrimSpace(name)
	version := strings.TrimSpace(c.Version)

	if !c.ForceRemote {
		if _, err := os.Stat(name); err == nil {
			abs, err := filepath.Abs(name)
			if err != nil {
				return abs, err
			}
			if c.Verify {
				if _, err := downloader.VerifyChart(abs, c.Keyring); err != nil {
					return "", err
				}
			}
			return abs, nil
		}
		if filepath.IsAbs(name) || strings.HasPrefix(name, ".") {
			return name, errors.Errorf("path %q not found", name)
		}
	}
	return c.LocateRemoteChart(name, settings)
}
```

mforutan:
@joejulian I still consider this as a bug, as I can't imagine a scenario that someone intentionally use `--repo` flag but they use a local folder to pull the chart, and if they do, it is wrong! I do understand however that this is a change in behavior but fixing any other bugs would be a change in behavior! 

The issue with introducing a new flag is the assumption that people are aware of this behavior and just want to have a workaround, which is not the case, at least it wasn't in my case! The workaround is very easy! just remove the empty local folder or rename it, that would fix the issue, but it will take at least few hours to reach that point! and it will be the same for using the new flag!

Having said that, I can't spend more time to update this code so please feel free to take it as it is or ignore it.

joejulian:
Thanks anyway.

PR #7878, opened 2020-04-05
[create] feat: allow image tag override

naseemkullah:
While using the chart's appVersion as image tag is the sanest default, it is not uncommon to want to override this if using a custom image, or using helm to manage  an in-house app running different tags across different environments.

cc @scottrigby @unguiculus @rimusz @mattfarina etc.

### DIFF
--- a/pkg/chartutil/create.go
+++ b/pkg/chartutil/create.go
@@ -97,6 +97,8 @@ replicaCount: 1
 image:
   repository: nginx
   pullPolicy: IfNotPresent
+  # Overrides the image tag whose default is the chart version.
+  tag: ""
 
 imagePullSecrets: []
 nameOverride: ""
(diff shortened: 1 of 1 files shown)


rimusz:
thanks @naseemkullah it looks as a good practices example to have in the `create` of the chart

hickeyma (approved):
LGTM thanks @naseemkullah. I would like to get some feedback from @helm/charts-maintainers before merging.

unguiculus (approved):
I like it. I was wondering anyways why `image.tag` had been dropped. Using `.Chart.AppVersion` makes only sense for using public charts when it should always match `image.tag`. For inhouse charts I always remove `AppVersion` completely because you pass in `image.tag` during deployment.

So, yes, let's merge this. 🎉

Issue #7389, opened 2020-01-13
helm 3.0.2 doesn't respect dependencies conditions


zzarkouna:
Output of `helm version`:
helmenv version
3.0.2
Output of `kubectl version`:
k version
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-15T23:42:50Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"14+", GitVersion:"v1.14.7-gke.23", GitCommit:"81c87c699557fed991e292cd328b2129c2f242a2", GitTreeState:"clean", BuildDate:"2019-11-07T19:23:23Z", GoVersion:"go1.12.11b4", Compiler:"gc", Platform:"linux/amd64"}
Cloud Provider/Platform (AKS, GKE, Minikube etc.):  GKE

parent chart.yaml
```
apiVersion: v2
description: A Kong chart for Kubernetes
name: kong
version: 0.1.0

dependencies:
- name: postgres
  version: "0.1.0"
  repository: "file://../postgres"
  condition: postgres.enabled, kong.postgres.enabled
- name: gcloud-sqlproxy
  version: "0.1.0"
  repository: "file://../gcloud-sqlproxy"
  condition: gcloud-sqlproxy.enabled, kong.gcloud-sqlproxy.enabled
```
and parent values.yaml
```
gcloud-sqlproxy:
  enabled: false
postgres:
  enabled: false
```
subchart chart.yaml
```
apiVersion: v2
description: A Kong chart for Kubernetes
name: kong
version: 0.1.0

dependencies:
- name: kong
  version: "0.1.0"
  repository: "file://../../../my_chart/kong"
```
and subchart values.yaml
```
kong:
  postgres:
    enabled: true

  gcloud-sqlproxy:
    enabled: false
```

`helm packages` asks me to run a dependency update
and when it packages it doesn't remove deps
and when it upgrades neither

alghazanth:
Same in 3.1.0 rc 3.

bacongobbler:
closing as stale.

Issue #5780, opened 2019-05-22
Helm 3 - legacy dependency conditions not respected 


planetf1:
Output of `helm version`:
```
version.BuildInfo{Version:"v3.0.0-alpha.1", GitCommit:"b9a54967f838723fe241172a6b94d18caf8bcdca", GitTreeState:"clean"}
```

Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.1", GitCommit:"b7394102d6ef778017f2ca4046abbaa23b88c290", GitTreeState:"clean", BuildDate:"2019-04-08T17:11:31Z", GoVersion:"go1.12.1", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.1+IKS", GitCommit:"7212ba91727da63ce9cfb2b5eba7f988f0f125d0", GitTreeState:"clean", BuildDate:"2019-05-09T14:28:29Z", GoVersion:"go1.12.1", Compiler:"gc", Platform:"linux/amd64"}
```
Cloud Provider/Platform (AKS, GKE, Minikube etc.): 
IBM Cloud (IKS)

I have a helm 2 chart that includes another chart by this content in requirements.yaml
```
dependencies:
  - name: cp-helm-charts
    version: 0.1.0
    repository: https://confluentinc.github.io/cp-helm-charts/
    condition: kafka.internal.enabled
  - name: openldap
    version: 0.3.0
    repository: https://kubernetes-charts.storage.googleapis.com/
```

However if kafka.internal.enabled is set to false (in my case values.yaml sets it to true, but I have content in an additional file specified by '-f ~/cloud.yaml which sets it to false), the cp-helm-chart (confluent kafka) IS deployed regardless

jdolitsky:
@planetf1 thanks for opening

Perhaps somehow related to https://github.com/helm/helm/issues/5757

hd-rk:
This issue happens even for dependencies specified in `Chart.yaml`. I have set up an example at https://github.com/hd-rk/helm-examples/tree/develop/v3-condition.

I took a look at the code, it might be because the function for processing sub-chart conditions/tags is not called in `action.Install.Run`
https://github.com/helm/helm/blob/ff818fceca20a4a5414fa26bcf63a305c6ef9343/pkg/chartutil/dependencies.go#L27-L32

Howerver, it's actually called in `action.Upgrade.Run`
[...]

PR #7887, opened 2020-04-07
fix: Change early exit condition to accommodate charts with sub dependencies

smckend:
My team has been having trouble using an umbrella chart type structure with Helm 3.

One of our sub charts has a dependency that Helm no longer honours so I decided to take a look into it. Turns out that the early exit condition in the `processDependencyEnabled` function seemed to be causing the problems.

I didn't want to remove the early exit check completely so I just added a small check to see whether the chart has any dependencies, as well, which might also have dependency metadata. I didn't think it was worth looping through all the potential dependencies looking for chart metadata. 

This solution is from someone who only just looked at the codebase today so I'm open to feedback and suggestions. Let me know.

Seems to cover #7389 and #5780.

Signed-off-by: Sam McKendrick <sam.macca9@gmail.com>

### DIFF
--- a/pkg/chartutil/dependencies.go
+++ b/pkg/chartutil/dependencies.go
@@ -116,7 +116,7 @@ func getAliasDependency(charts []*chart.Chart, dep *chart.Dependency) *chart.Cha
 
 // processDependencyEnabled removes disabled charts from dependencies
 func processDependencyEnabled(c *chart.Chart, v map[string]interface{}, path string) error {
-	if c.Metadata.Dependencies == nil {
+	if c.Metadata.Dependencies == nil && len(c.Dependencies()) == 0 {
 		return nil
 	}
 
--- a/pkg/chartutil/dependencies_test.go
+++ b/pkg/chartutil/dependencies_test.go
@@ -269,6 +269,63 @@ func TestProcessDependencyImportValuesForEnabledCharts(t *testing.T) {
 	}
 }
 
+func TestProcessDependencyEnabledRespectsChildConditionsWhenParentLacksDependenciesBlock(t *testing.T) {
+	c := loadChart(t, "testdata/respect-conditions-without-dependencies-block/parent-chart")
+
+	if err := processDependencyImportValues(c); err != nil {
+		t.Fatalf("processing import values dependencies %v", err)
+	}
+
+	if len(c.Dependencies()) != 1 {
+		t.Fatalf("expected 1 dependency for this chart, but got %d", len(c.Dependencies()))
+	}
+
+	if len(c.Metadata.Dependencies) != 0 {
+		t.Fatalf("expected 0 dependencies specified in parent chart's Chart.yaml, got %d", len(c.Metadata.Dependencies))
+	}
+
+	if len(c.Dependencies()[0].Dependencies()) != 3 {
+		t.Fatalf("expected 3 dependencies for the subchart, but got %d", len(c.Dependencies()))
+	}
+
+	if len(c.Dependencies()[0].Metadata.Dependencies) != 3 {
+		t.Fatalf("expected 3 dependencies specified in subchart's Chart.yaml, got %d", len(c.Dependencies()[0].Metadata.Dependencies))
+	}
+
+	if err := processDependencyEnabled(c, c.Values, ""); err != nil {
+		t.Fatalf("expected no errors but got %q", err)
+	}
+
+	if len(c.Dependencies()) != 1 {
+		t.Fatalf("expected 1 dependency for this chart, but got %d", len(c.Dependencies()))
+	}
+
+	if len(c.Dependencies()[0].Dependencies()) != 2 {
+		t.Fatalf("expected 2 dependencies for this chart, but got %d", len(c.Dependencies()[0].Dependencies()))
+	}
+
+	if len(c.Metadata.Dependencies) != 0 {
+		t.Fatalf("expected 0 dependencies specified in Chart.yaml, got %d", len(c.Metadata.Dependencies))
+	}
+
+	if len(c.Dependencies()[0].Metadata.Dependencies) != 2 {
+		t.Fatalf("expected 2 dependencies specified in subchart's Chart.yaml, got %d", len(c.Dependencies()[0].Metadata.Dependencies))
+	}
+
+	subChartDependencyChart0Name := c.Dependencies()[0].Dependencies()[0].Name()
+	subChartDependencyChart1Name := c.Dependencies()[0].Dependencies()[1].Name()
+	devChartName := "dev"
+	prodChartName := "prod"
+
+	if subChartDependencyChart0Name != devChartName {
+		t.Fatalf("subchart's dependency chart name should be %s but got %s", devChartName, subChartDependencyChart1Name)
+	}
+
+	if subChartDependencyChart1Name != prodChartName {
+		t.Fatalf("subchart's dependency chart name should be %s but got %s", prodChartName, subChartDependencyChart1Name)
+	}
+}
+
 func TestGetAliasDependency(t *testing.T) {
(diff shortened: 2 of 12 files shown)


technosophos:
Oh! That's interesting! I wonder if there is a bug where `Metadata.Dependencies` is not getting set when it is supposed to. Perhaps the enabled/disabled checks are being performed after the metadata object is built.

Let me take a look into this and make sure that there is not a deeper bug lurking. Otherwise, your fix and test look great.

pkg/chartutil/testdata/respect-conditions-from-sub-subchart/parent-chart/Chart.yaml
```diff
@@ -0,0 +1,4 @@
+apiVersion: v2
+name: parent-chart
+version: v0.1.0
+appVersion: v0.1.0
```
technosophos, line 4: This chart has a dependency (daughter-chart) that is not declared here.

technosophos:
Okay, I found some interesting things while testing this.

First, if you add the `dependencies:` section to `parent-chart/Chart.yaml`, it will come closer to working as expected. To be clear, in Helm 3 it is required that umbrella charts use a `dependencies` declaration. In Helm 2, we had a failover condition that did not require this because the original Chart v1 spec did not have a `requirements.yaml`. But in Helm 3, we got strict on this point.

Second, I started with your test charts and added a `templates/secret.yaml` to the `dev` dependency. I found a surprising bug.

This worked fine: `helm template ./parent-chart --set=dev.enabled=true`:

```yaml
---
# Source: parent-chart/charts/daughter-chart/charts/dev/templates/secret.yaml
apiVersion: v1
kind: Secret
metadata:
  name: issue7887
stringData:
  key: value
```

But this did not work at all: `helm template ./parent-chart --set=global.dev.enabled=true`

So on a whim, I went and changed one thing in the Chart.yaml for daughter-chart:

```yaml
apiVersion: v2
name: daughter-chart
version: v0.1.0
appVersion: v0.1.0
dependencies:
  - name: dev
    repository: "file://envs/dev"
    version: ">= 0.0.1"
    condition: global.dev.enabled,dev.enabled    # <-- reversed
  - name: prod
    repository: "file://envs/prod"
    version: ">= 0.0.1"
    condition: prod.enabled,global.prod.enabled
```

And now, `helm template ./parent-chart --set=global.dev.enabled=true` works fine, but `helm template ./parent-chart --set=dev.enabled=true ` outputs nothing.

So there is _definitely_ a bug in whatever is supposed to process the `condition` line.

I'll look into this more on Monday, but I thought I'd leave an update here in case you can make more progress in my absence.

smckend:
Thanks for your comments and clarifications. 

So with every chart that is added to the `charts` folder there should be a dependency entry? That's definitely going to mess with our automation as we're just committing sub-charts to the `charts` directory. Seems that for the most part this issue was due to our misunderstandings but I can definitely help look into the uncovered issue.

In regards to your last comment, shouldn't the second command's value key be located under the daughter chart? so `helm template ./parent-chart --set=daughter-chart.dev.enabled=true`.

technosophos:
Yeah, every subchart should have a `dependencies` entry. Thinking about this over the weekend, I think we should probably add a linter rule to validate this for users of `helm lint`.

On your second question, yes -- you are right. I'm going to re-test, but something is definitely wrong in this code. Things that should not work are working, while things that should work are not.

technosophos:
Added the proposal at #7910

smckend:
I think if this is the way forward adding linter rules is a must. Especially because it isn't very well documented that you must have a dependency entry in the `Chart.yaml` file (at least at the moment from my experience). After looking at the issues here in Github it seems a lot of people are running into the same problem with conditional dependencies and I'd bet it's either to do with a lack of the `dependencies` block or because of this second issue you've uncovered. 

Is this fix for the first issue/misunderstanding not harmless enough that we could include it and maybe remove it later when the restrictions around metadata dependencies are more mature? (as in, charts in the `charts/` folder are ignored if there is no dependency entry). The problem is that helm picks up these charts but doesn't process them properly which is very misleading.

And regarding your second issue found, I'll most likely be able to find some time mid week to have a dive into the code to see if I can help.

technosophos:
Right now, I am trying to get 3.2 released. The next thing I will do after I get that rolling is start on all of the dependency-related issues. There are a bunch of very closely related issues that I think we can get fixed. This one makes a lot of sense to start with because it is straightforward.

smckend:
Okay slightly delayed but I managed to take a look at it today. 

I'd argue that what you found is not a bug but rather expected behaviour as per some assumptions I've made. I extended my test to replicate the issue and shall use the test chart to explain my thoughts:

```yaml
# Daughter chart's Chart.yaml dependency block
  - name: dev
    repository: "file://charts/dev"
    version: ">= 0.0.1"
    condition: global.dev.enabled,dev.enabled // global first
  - name: staging
    repository: "file://charts/staging"
    version: ">= 0.0.1"
    condition: staging.enabled,global.staging.enabled  // local first
  - name: prod
    repository: "file://charts/prod"
    version: ">= 0.0.1"
    condition: global.prod.enabled,prod.enabled  // global first
```
I added a third chart `staging` which has the order of conditions reversed compared to the other 2 charts. Therefore `dev` and `prod` will prioritise global values but `staging` will prioritise the local one.

```yaml
# Values for daughter-chart.
dev:
  enabled: false
staging:
  enabled: false
prod:
  enabled: true
```
In my test, I disable both `dev` and `staging` in the daughter chart's `values.yaml`. Only looking at this file we expect prod to be deployed.

```yaml

# Values for parent-chart.
global:
  dev:
    enabled: true
  staging:
    enabled: true
```
In the `values.yaml` of the parent chart, I then try to enable both `dev` and `staging`.

So what would we expect to be the output of a `helm template` on the parent chart?

Without my change, **it would be all 3.** 
With my change, **it would be just `dev` and `prod`.**

This table helps shows the logic:
|         | First dependency condition | Local value | Global value | Processed Value |
|---------|----------------------------|-------------|--------------|-----------------|
| dev     | global                     | false       | true         | true            |
| staging | local                      | false       | true         | false           |
| prod    | global                     | true        |              | true            |

`prod` doesn't have a global value so it uses the second dependency condition but for the other 2 dependencies they (with my change) correctly take the value they find from their first depdency condition. So things actually look more correct to me.

Thoughts?

smckend:
Just to clarify, this is only to fix the behaviour when there is a missing `dependencies` block. With the `dependencies` block, everything works the way I described above.

I'd just like to either ignore charts not defined in the `dependencies` block or acknowledge them and treat them as you would a correctly configured chart. This change is for the latter so that we don't get any unexpected behaviour.

technosophos:
The issue we are chasing is this: With Helm 3, we _assumed_ (during development) that dependencies would match between Chart.yaml and the directory. But the top level checks were never consistently added. Relaxing the checks will very likely lead to inconsistent behavior. So now we have to figure out what to do.

The two options on the table are to either rewrite the Helm 3 internals to NOT require that the two be in sync, or to make everything stricter.

I am STRONGLY in favor of just make the checks stricter across the board. I don't think there should ever be a case where a chart appears in the `charts/` directory but not in the `Chart.yaml`. I think Helm should produce an error any time it encounters that situation.

However, I'm waiting for other core maintainers to weigh in on a variety of related issues. Then I hope we will come to a uniform view, even if it's not my preferred view.

/cc @mattfarina since we literally talked about this a few hours ago. Also, someone signed up last Helm call to create a meta-issue for this and all related issues. Does anyone remember who signed up? If not, I will start working on it.

smckend:
Makes sense to me. As long as we protect against that unexpected behaviour, I'm happy.

Once a decision has been made let me know so we can make a call on whether to close this issue or not.

dobharweim:
Has a decision been made on this one way or the other?

PR #7892, opened 2020-04-08
Adding notes on semver to create Chart.yaml

mattfarina:
The version field in the Chart.yaml has a comment describing it
but it did not note the version needs to follow SemVer. There
have been numerous questions, over time, about this format. Add
note here so it's exposed in more places.


scottrigby (approved):
This seems clear 👌

hickeyma (approved):
LGTM, thanks @mattfarina

PR #7897, opened 2020-04-09
fix: rebuild chart after dependency update on install

### LINKED ISSUE #7857: Helm 3 is not possible to install dependent charts by executing `helm install` with `--dependent-update` on a local chart file
Output of `helm version`: 
```
```

Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.0", GitCommit:"70132b0f130acc0bed193d9ba59dd186f0e634cf", GitTreeState:"clean", BuildDate:"2019-12-07T21:20:10Z", GoVersion:"go1.13.4", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.0", GitCommit:"70132b0f130acc0bed193d9ba59dd186f0e634cf", GitTreeState:"clean", BuildDate:"2020-01-14T00:09:19Z", GoVersion:"go1.13.4", Compiler:"gc", Platform:"linux/amd64"}
```

Cloud Provider/Platform (AKS, GKE, Minikube etc.): kind

----

# What happens

With helm 3.1.2, when run `helm install` with `--dependent-update` on a local chart file, it install target chart but **does not install dependent charts**.

If you run `helm install` after running `helm dependency update` it will install all the resources which include dependent charts. So, I think this is a bug of `--dependent-update` option.

# How to reproduce

1. Create chart files locally
    * ex. run `helm create sample`
    * It doesn't matter chart's apiVersion (v1 or v2 either is fine)
2. Set at least one dependency
    * When using helm v2, add dependencies into requirements.yaml
    * When using helm v3, add dependencies into Chart.yaml
3. Run `helm install` with `--dependency-update` using helm v3
    * ex. run `helm install --dependency-update sample ./sample`
    * When run `helm install`, it is needed to be in a state of unresolved dependency
    * The point is that the `requirements.lock` and `Chart.lock` does not exist

# Workaround

Update dependency before install chart.

1. Run `helm dependency update`
2. Run `helm install`

# log

Using [this](https://github.com/ryu-sato/helm-charts-sample) sample chart.

This chart is created by `helm create` and add requirement.txt manually.

[log.txt](https://github.com/helm/helm/files/4422665/log.txt)

technosophos:
This reloads the chart after dependencies have been updated on a `helm install --dependency-update ...` operation.

This fixes the bug in which a dependency would be fetched locally, but not deployed. The issue was that even though the chart was fetched, the dependencies were loaded before the dependency list was refreshed. This simply reloads the chart.

Added a unit test of `--dependency-update`

Closes #7857 


cmd/helm/install.go
```diff
@@ -210,10 +210,15 @@ func runInstall(args []string, client *action.Install, valueOpts *values.Options
 					Getters:          p,
 					RepositoryConfig: settings.RepositoryConfig,
 					RepositoryCache:  settings.RepositoryCache,
+					Debug:            true,
```
mattfarina, line 213: Should the value of `Debug` be `settings.Debug` instead of `true`?

technosophos: Oh, I guess it should actually be omitted. Good catch

technosophos: I think I will add this as settings.Debug, since that is probably what was originally intended. There was no Debug setting on this before, and I set it to `true` to see if it generated any output (which it doesn't)

technosophos: Fixed. Please re-review when you have a moment.

adamreese (approved):

jdolitsky (approved):

PR #7906, opened 2020-04-12
fix(comp): Prepare plugin completion for Cobra 1.0

marckhouzam:

Prepares Helm plugin completion for Cobra 1.0.

See [Cobra v1.0.0 changelog](https://go.libhunt.com/cobra-changelog/1.0.0) for more details, but the changes we aim for are:
- Custom completions coded in Go (instead of Bash) https://github.com/spf13/cobra/pull/1035
- Fish completion (including support for Go custom completion) https://github.com/spf13/cobra/pull/1048

However, supporting these changes will break helm's current plugin completion.  This PR fixes that (see below for details).

This PR seems larger than it really is as it is mostly a refactoring to extract the logic to call a plugin's executable script.  The behavioural change is small.

The completion part was verified with the acceptance-testing repo.

***

Cobra 1.0.0 introduces support for Custom Go Completions.  However, there is a small difference compared to helm's existing solution.  The difference is that Cobra will not complete sub-commands if there are any dynamic completion function registered.

The way helm implemented plugin completions had the dynamic completion function associated with the root plugin command (e.g. 2to3); with Cobra, this prevents any of the sub-commands of the plugin to be completed (i.e., 2to3 <TAB> will not show the static sub-commands).

The solution is to only register the dynamic completion function for a plugin command that doesn't have any further sub-commands.  This allows all sub-commands to be completed properly, and when there are no more levels of sub-commands, the dynamic completion is called.

This commit had to make two changes to achieve this:
1. refactor the load_plugins.go file to extract the logic to call the plugin.complete executable (for dynamic completions), so that it could be added to each final sub-command.
1. load the static completion not only for the `completion` command like before, but also for the `__complete` command, so that when the ` __complete` command is called for dynamic completion of a sub-command, it is aware of the sub-command in question.

## If applicable

### DIFF
--- a/cmd/helm/load_plugins.go
+++ b/cmd/helm/load_plugins.go
@@ -64,54 +64,12 @@ func loadPlugins(baseCmd *cobra.Command, out io.Writer) {
 		return
 	}
 
-	processParent := func(cmd *cobra.Command, args []string) ([]string, error) {
-		k, u := manuallyProcessArgs(args)
-		if err := cmd.Parent().ParseFlags(k); err != nil {
-			return nil, err
-		}
-		return u, nil
-	}
-
-	// If we are dealing with the completion command, we try to load more details about the plugins
-	// if available, so as to allow for command and flag completion
-	if subCmd, _, err := baseCmd.Find(os.Args[1:]); err == nil && subCmd.Name() == "completion" {
-		loadPluginsForCompletion(baseCmd, found)
-		return
-	}
-
 	// Now we create commands for all of these.
 	for _, plug := range found {
 		plug := plug
 		md := plug.Metadata
 		if md.Usage == "" {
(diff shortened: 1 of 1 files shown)


marckhouzam:
This PR is a blocker for contribution #8111, but also for 3 or 4 improvements I have tucked away. Basically the following stuff is blocked:
- Fish completion
- Dropping the internal package “completion”
- Improved custom completion (adding descriptions)
- Upcoming improved Zsh completion (with descriptions)

scottrigby (approved):
Thanks for the detailed chat. Updated description a bit per our convo. No change to end users or plugin maintainers. Also congrats on those changes to spf13/cobra. Looks good! :+1:

bacongobbler (approved):
code LGTM

marckhouzam:
Thanks @scottrigby and @bacongobbler for the review!

PR #7653, opened 2020-02-21
fix(helm): allow a previously failed release to be upgraded

yinzara:
This PR changes the behavior of "helm upgrade" command when there have been no successful and at least one failed release.  This allows the user to recover from a failed install (or a partially failed install) without uninstalling the release.   The user can now decide to decide to ignore the "no deployed releases" errors and allow it to upgrade a failed release.

There were no test cases that tested the failure scenarios of upgrading a failed/missing release so I added them and also added additional test cases that test the new functionality.

Fixes issue #5595

Signed-off-by: Matt Morrissete <yinzara@gmail.com>

Issue #3353, opened 2018-01-17
`helm upgrade --install` doesn't perform an install/upgrade if the first ever install fails


jstrachan:
Using `helm upgrade --install` is a nice way to install or upgrade depending on if the release exists. But it looks like there's a bug in the logic; its not handling failed installs. In my case the first install failed; then a subsequent attempt wasn't even made as it crashes out immediately. 

Maybe if the last release failed then `helm upgrade --install` should delete it and install again?

```
$ helm list
NAME           	REVISION	UPDATED                 	STATUS  	CHART                         	NAMESPACE
foo            	2       	Wed Jan 17 11:48:08 2018	FAILED	something-0.0.1                     	default

$ helm upgrade "foo" . --install 
Error: UPGRADE FAILED: "foo" has no deployed releases
```

bacongobbler:
This was intentional by design of https://github.com/kubernetes/helm/pull/3097. Basically, diffing against a failed deployment caused undesirable behaviour, most notably this long list of bugs:

- https://github.com/kubernetes/helm/issues/2257
- https://github.com/kubernetes/helm/issues/2341
- https://github.com/kubernetes/helm/issues/2346
- https://github.com/kubernetes/helm/issues/2437
- https://github.com/kubernetes/helm/issues/3068

If your initial release ends up in a failed state, we recommend purging the release via `helm delete --purge foo` and trying again. After a successful initial release, any subsequent failed releases will be ignored, and helm will do a diff against the last known successful release.

Now that being said, it might be valuable to *not* perform a diff when no successful releases have been deployed. I'd be a little concerned about certain use cases though. @adamreese do you have any opinions on this one?

chancez:
The suggested fix seems completely untenable in an automated system. I definitely don't want everything invoking helm to have to know about "if first release fails, delete and retry". For one, most of my tooling isn't aware if it's an install or upgrade, or if it's the first time or 100th time, it's almost always just running `helm upgrade --install`.

chancez:
I'd also like to call out that I commented on the original PR https://github.com/kubernetes/helm/pull/3097#discussion_r151808599 asking specifically about this case.

stealthybox:
The old behavior was better for this case.
I agree with @chancez. This makes `upgrade --install` non-idempotent for a common occurrence.

@bacongobbler
If we're worried about releases failing and leaving shrapnel due to failed hooks, I'd say that's a design issue the chart. ( Hooks work better when they are idempotent )
[...]

PR #3097, opened 2017-11-03
fix(tiller): upgrade last deployed release

adamreese:
Fixes #2437

Two bugs were causing this behavior

- Tiller was marking the previous release superseded when an upgrade
failed.
- Upgrade was diffing against failed releases

Issue #3068, opened 2017-10-26
Helm upgrade -i will "succeed" after failing on duplicate resource.


tcolgate:
## What happens

create a service named myservice in a cluster.

create a chart that attempts to create/manage a service named myservice.

`helm upgrade -i ./chart mychart`

First run will fail. The release status will be "FAILED". No pods or other resources will be created.

Now run `helm upgrade -i ./chart mychart` again. The deploy now succeeds, but myservice is not updated and does not match the requested service in the release. release status is now DEPLOYED.

## What should happen.

The second upgrade -i should also fail. It should be clear to the user that the api objects are not being successfully managed.

bacongobbler:
fixed via #3097

PR #7913, opened 2020-04-14
fix: allow upgrading when history has nothing but failures

technosophos:
**This is a Work In Progress. DO NOT MERGE IT.**

It has long been possible for people to get their releases into a state where they could no longer successfully upgrade because they had failed enough times that no deployed releases are left in their history.

This can happen in the following ways:

- They can install and fail (without `--atomic`), which leaves them with only one failed release
- They can install successfully, but have supplied `--wait`, and failed when a pod did not come up (which will mark the deployment failed)
- They can set their history to some low-ish number, and then perform enough failed releases in a row that no `deployed` ones exist

And there are probably others. All of these cases are exacerbated by CI/CD systems that naively install and then don't clean up after failures.

See #5595 for context

## The Approach

The approach I took here was to attempt to leverage Helm 3's 3-way diff to recover from a broken release. (Note: This approach is not possible in Helm 2.)

In the usual `prepareUpdate`, Helm attempts to load the last _deployed_ release from history. If none is found, the deployment fails. The proposed fix here extends that logic:

- If the deployed release exists, use that release
- If a deployed release does NOT exist, try to load the last release and use that, regardless of whether it failed
- If both loads fail, the release fails

The wager here is that even if there is a failed release, the 3-way diff will be able to detect the changes in cluster and recover from the potentially inaccurate release manifest.

I am proposing this as a bug fix, as it fixes a case that cannot currently be accounted for at all (when history gets truncated, and only failed releases are left). It seems to me that if this route does not introduce any anomalous behavior, it is right to make it the default action taken. I think it would be a patently bad idea to add yet another flag to the already misleading jumble of flags hanging off of `helm update`.

_We need to look very carefully at this approach._
* Does it solve the problem of history + failed releases?
* Does it break expected behavior?
* Does it introduce new possible bugs by diffing against a failed manifest?

## A few misc notes:

- This should work fine paired with `--atomic`, since it will not trigger the atomic rollback condition unless there is a new failure.
- This should work with `--cleanup-on-fail`, which will remove broken resources, since this continues to use the existing diff logic
- This will work with any version of --history-max greater than or equal to 1.
- When `--reuse-values` is set, this will use values that were set on the last failed update
- How do we verify its behavior?

## Example

I have a chart called `alwaysfails`, which is basically `helm create alwaysfails` with the `values.yaml` setting `image` to `nosuchimage9999`.

```console
# Run this command as often as you like
$ helm3 upgrade --install alwaysfails ./alwaysfails --debug --wait --timeout=5s
# Then remove the `--wait --timeout=5s`
$ helm3 upgrade --install alwaysfails ./alwaysfails --debug
```
I ran three failed releases, and a successful 4th.

Since removing the `wait` will mark the deployment successful, you will see this in history:

```console
$ helm3 history alwaysfails
REVISION	UPDATED                 	STATUS    	CHART            	APP VERSION	DESCRIPTION
1       	Mon Apr 13 19:33:24 2020	failed    	alwaysfails-0.1.0	1.16.0     	Release "alwaysfails" failed: timed out waiting for the condition
2       	Mon Apr 13 19:33:32 2020	failed    	alwaysfails-0.1.0	1.16.0     	Upgrade "alwaysfails" failed: timed out waiting for the condition
3       	Mon Apr 13 19:33:39 2020	superseded	alwaysfails-0.1.0	1.16.0     	Upgrade "alwaysfails" failed: timed out waiting for the condition
4       	Mon Apr 13 19:35:27 2020	deployed  	alwaysfails-0.1.0	1.16.0     	Upgrade complete
```

(Note that it does currently mark the recovered release as `superseded` instead of `failed`. We could change that. Not sure what the utility is either way.)

### DIFF
--- a/pkg/action/upgrade.go
+++ b/pkg/action/upgrade.go
@@ -33,6 +33,7 @@ import (
 	"helm.sh/helm/v3/pkg/postrender"
 	"helm.sh/helm/v3/pkg/release"
 	"helm.sh/helm/v3/pkg/releaseutil"
+	"helm.sh/helm/v3/pkg/storage"
 )
 
 // Upgrade is the action for upgrading releases.
@@ -132,7 +133,23 @@ func (u *Upgrade) prepareUpgrade(name string, chart *chart.Chart, vals map[strin
 	// finds the deployed release with the given name
 	currentRelease, err := u.cfg.Releases.Deployed(name)
 	if err != nil {
-		return nil, nil, err
+		switch err.(type) {
+		case storage.NoDeployedReleases:
+			// Try lookup again. Note that what we are doing here is attempting to get the
+			// _last release_, which might have been a failure. Our assumption is that this
+			// release best represents the current state of a rendered Helm chart. We are then
+			// relying upon the 3-way diff to help us repair the current in-cluster version.
+			u.cfg.Log("Could not find a deployed release for %q. Trying to get an older (possibly failed) release", name)
+			currentRelease, err = u.cfg.Releases.Last(name)
+			if err != nil {
+				// Regardless of what the error is here, it is likely that we cannot proceed
+				// any more with the upgrade. So we bail.
+				return nil, nil, errors.Wrapf(err, "there are no recorded releases for %q", name)
+			}
+			u.cfg.Log("loaded release %s.%d", currentRelease.Name, currentRelease.Version)
+		default:
+			return nil, nil, err
+		}
 	}
 
 	// determine if values will be reused
@@ -196,6 +213,7 @@ func (u *Upgrade) prepareUpgrade(name string, chart *chart.Chart, vals map[strin
 	if len(notesTxt) > 0 {
 		upgradedRelease.Info.Notes = notesTxt
 	}
+	u.cfg.Log("Manifest: %s", manifestDoc)
 	err = validateManifest(u.cfg.KubeClient, manifestDoc.Bytes(), !u.DisableOpenAPIValidation)
 	return currentRelease, upgradedRelease, err
 }
--- a/pkg/storage/storage.go
+++ b/pkg/storage/storage.go
@@ -107,6 +107,14 @@ func (s *Storage) ListDeployed() ([]*rspb.Release, error) {
 	})
 }
 
+// NoDeployedReleases describes the situation where none of the releases in the history are deployed.
+//
+// Aggressive pruning can mask the fact that there is a deployed release, but that it
+// was cleaned up by the history pruner.
+type NoDeployedReleases string
+
+func (e NoDeployedReleases) Error() string { return string(e) }
+
 // Deployed returns the last deployed release with the provided release name, or
 // returns ErrReleaseNotFound if not found.
 func (s *Storage) Deployed(name string) (*rspb.Release, error) {
@@ -116,7 +124,7 @@ func (s *Storage) Deployed(name string) (*rspb.Release, error) {
 	}
 
 	if len(ls) == 0 {
-		return nil, errors.Errorf("%q has no deployed releases", name)
+		return nil, NoDeployedReleases(fmt.Sprintf("%q has no deployed releases", name))
 	}
 
 	// If executed concurrently, Helm's database gets corrupted
@@ -140,7 +148,7 @@ func (s *Storage) DeployedAll(name string) ([]*rspb.Release, error) {
 		return ls, nil
 	}
 	if strings.Contains(err.Error(), "not found") {
-		return nil, errors.Errorf("%q has no deployed releases", name)
+		return nil, NoDeployedReleases(fmt.Sprintf("%q has no deployed releases", name))
 	}
 	return nil, err
 }


marckhouzam:
I just learned about Draft PRs today, if you're interested: https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/changing-the-stage-of-a-pull-request#converting-a-pull-request-to-a-draft

pkg/storage/storage.go
```diff
@@ -107,6 +107,14 @@ func (s *Storage) ListDeployed() ([]*rspb.Release, error) {
 	})
 }
 
+// NoDeployedReleases describes the situation where none of the releases in the history are deployed.
+//
+// Aggressive pruning can mask the fact that there is a deployed release, but that it
+// was cleaned up by the history pruner.
+type NoDeployedReleases string
+
+func (e NoDeployedReleases) Error() string { return string(e) }
```
jlegrone, line 116: Assuming we're ok with targeting go 1.13, we could use wrapped errors: https://blog.golang.org/go1.13-errors

https://play.golang.org/p/4ryS6CCiPCa

liuming-dev:
similar PR #7767 about rollback, need to get reviewer

bacongobbler:
With Helm 2, @adamreese and I paired on this particular case. It's a doozy.

Together, we introduced https://github.com/helm/helm/pull/3097 to prevent cases where a failed deployment could not be reconciled simply because Helm 2 could not reconcile against the live state of the cluster.

This is likely the most controversial pull request we ever merged into Helm as seen with https://github.com/helm/helm/issues/3353#issuecomment-358367529. For posterity, those issues were #2257, #2341, #2346, #2437, and #3068, but there are likely others.

Now that we can reconcile against the live state of the cluster in Helm 3, I think it's fine to relax this constraint as we can at least perform _some form_ of reconciliation with the last failed release.

If you wouldn't mind looking through those past issues and see if there may be any cases that might not be addressed by this proposal, that'd be wonderful. I'm all for this but perhaps there may be some case I'm forgetting back when Helm 2 used to diff against the last failed release that might be worth looking into.

yinzara:
This is solved by #7653 as well.

technosophos:
Yeah, it looks like this takes a pretty similar approach to #7653, but without introducing a new flag. When I looked at that PR, I got about as far as the help text, and the description made it sound  highly dangerous. The help text in that PR suggest that that version is not production-safe... and I am wondering why @yinzara.

I'm getting concerned that the sheer number of combinatorics on the `helm upgrade` command are making it so complicated that we are never going to be able to solve this problem. (And, let's be honest, a huge part of this problem is just the design of Kubernetes.)

For example, depending on which combination of `--install` `--atomic`, `--wait`, `--force`, `--reuse-values`, `--recreate`, `--timeout`, you can get the exact same chart + values to end up in any number of different states over time. We somehow have to get the default Helm behavior to be sane enough that this is not the responsibility of the user to untangle all of the flags.

Issue #5595 is a good example: People naively think that because they all got a similar error message, they are all hitting the same "bug." But if you read through the thread, you will see there there are dozens of minuscule differences in flags + history depth + expectations that are leading to a variety of actual runtime conditions that are "not desired". In the end, releases they believe should be running or recoverable are not -- but the causes are variegated, massively compounded by the complexity of all of the flags.

I don't particularly have strong feelings about whether it is better to go ahead with a slightly modified version of #7653 or with this. But I am going to spend some time analyzing how dangerous this approach actually is -- and whether the new diffing strategy in Helm 3 can actually do an adequate job. If it can't then perhaps we do use #7653 as-is and just let people live on the edge. (Which, no doubt, will just produce new acrimonious threads with people accusing us of neglect, ignorance, and malice like they are in #5595.) But if we can show this solution to be reasonably effective, I think we should fix it as part of the core.

Finally, I am still on the fence about `--atomic`. Maybe we have to start telling people, "look, there is _no recovery story_ for failing a `helm install` unless you use `--atomic`." That actually sounds reasonable to me. But I have reservations that `--atomic` is all that reliable in all cases. Rollbacks are not time machines (as @bridgetkromhout likes to say). They come with their own set of difficulties. Are they acceptable difficulties? Perhaps. Might we be able to just skip this particular problem and rely on `--atomic`? I am not sure.

technosophos:
The solution in #7767 is to just treat the current release as the last release. That is an interesting proposal as well, though I think that in some cases it will orphan objects that were created by a previous release. Might be good to investigate that approach as well, though.

pkg/action/upgrade.go
```diff
@@ -196,6 +213,7 @@ func (u *Upgrade) prepareUpgrade(name string, chart *chart.Chart, vals map[strin
 	if len(notesTxt) > 0 {
 		upgradedRelease.Info.Notes = notesTxt
 	}
+	u.cfg.Log("Manifest: %s", manifestDoc)
```
adamreese, line 216: Heavy log entry

technosophos: I will remove.

technosophos:
Okay @adamreese and I paired on this for a few hours today, and looked deeply into the patching code. We did find a few minor things that needed fixing. Specifically, the diffing logic was too likely to leave the release in an unstable state because it was too conservative in deletions. So we are looking into a fix for that, but will do that in a separate PR.

We also looked at the case where the new chart added an object that the old (possibly failed) release did not have, but that exists in the cluster. We decided to retain the current behavior, which is that this case _should_ fail. Otherwise we could accidentally adopt a resource that Helm was not intended to manage. But this could present an edge case if we go ahead with this proposal.

@yinzara are you interested in updating your PR to remove the flag? And maybe to use a typed error? After pair-reviewing this with @adamreese, we think this approach will work well enough that we don't need a flag. And since your patch landed before mine, I'm good with continuing with yours. We left a couple of comments on your PR, but they were all easy to fix -- especially if the u.Replace flag is removed.

We think only the following release statuses are acceptable to recover from (so both of our proposed fixes are wrong as-is):

- Failed
- Superseded

But it is unclear whether you can actually ever get a `Superseded` without first finding a `Deployed`. So probably _just_ `Failed`. (`Deleted` and all of the `Pending*` ones will introduce additional issues in the diff/merge logic.)

yinzara:
I'm happy to update the PR to whatever you'd like.  I'm confused what you mean by using a typed error instead of a flag?

yinzara:
Oh wait. I think you're saying to just make this the default behavior and to not introduce the flag but limit it only to the Failed and Superseded statuses.  No issues. I'll handle that.  And what you mean a typed error? I'm not 100% sure what you're referring to. Which error? You mean if there is only one in another status? or?

technosophos:
Closing in favor of #7653

PR #7925, opened 2020-04-14
fix(upgrade): Add an improved user error message for removed Kubernetes APIs

hickeyma:
The error message returned from Kubernetes when APIs are removed is not very informative. This PR adds additional information to the user. It covers the current release manifest APIs validation.

Partial #7219

### DIFF
--- a/pkg/action/upgrade.go
+++ b/pkg/action/upgrade.go
@@ -233,6 +233,13 @@ func (u *Upgrade) prepareUpgrade(name string, chart *chart.Chart, vals map[strin
 func (u *Upgrade) performUpgrade(originalRelease, upgradedRelease *release.Release) (*release.Release, error) {
 	current, err := u.cfg.KubeClient.Build(bytes.NewBufferString(originalRelease.Manifest), false)
 	if err != nil {
+		// Checking for removed Kubernetes API error so can provide a more informative error message to the user
+		// Ref: https://github.com/helm/helm/issues/7219
+		if strings.Contains(err.Error(), "unable to recognize \"\": no matches for kind") {
+			return upgradedRelease, errors.Wrap(err, "current release manifest contains removed kubernetes api(s) for this "+
+				"kubernetes version and it is therefore unable to build the kubernetes "+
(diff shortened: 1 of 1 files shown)


Nastradamus (approved):

pkg/action/upgrade.go
```diff
@@ -233,6 +233,13 @@ func (u *Upgrade) prepareUpgrade(name string, chart *chart.Chart, vals map[strin
 func (u *Upgrade) performUpgrade(originalRelease, upgradedRelease *release.Release) (*release.Release, error) {
 	current, err := u.cfg.KubeClient.Build(bytes.NewBufferString(originalRelease.Manifest), false)
 	if err != nil {
+		// Checking for Kubernetes deprecated APIs error so can provide a more informative error message to the user
+		// Ref: https://github.com/helm/helm/issues/7219
+		if strings.Contains(err.Error(), "unable to recognize \"\": no matches for kind") {
+			return upgradedRelease, errors.Wrap(err, "current release manifest contains deprecated kubernetes api(s) for this "+
```
marckhouzam, line 239: I don't mean to be a stickler on terminology but IIUC a "deprecated" API won't cause a problem but a "removed" API (which was deprecated in the previous versions of kube) will cause the problem.

So the work message should probably replace "deprecated" with "removed".

hickeyma: Good point. Updated.

marckhouzam:
I don't have the setup to test this, but the code looks good.  Nice improvement for this annoying situation.  Thanks!

LGTM

marckhouzam (approved):

hickeyma:
Thanks for the reviews @Nastradamus and @marckhouzam. I will ask some of the other @helm/helm-core-maintainers to take a look at it also, as it is a bit of a workaround.

technosophos (approved):
Since it's a small bug, you technically didn't need a second review. However, I'm happy to review and say LGTM

hickeyma:
Thanks @technosophos. Just wanted to make sure I wasn't violating any set standard in the community.

PR #7926, opened 2020-04-15
fix: lint with strict passes for helm create

technosophos:
A change in the YAML parser between Helm 2 and Helm 3 led to a subtle bug in `helm lint`

Closes #7923

PR #7927, opened 2020-04-15
test: forward-port regression test from Helm 2

technosophos:
Add a regression test from #7926 (Helm v2) that makes sure `helm create` produces charts that will pass `helm lint --strict`.


hickeyma (approved):
LGTM. Good idea to put this test here after bug in v2.

PR #7929, opened 2020-04-15
fix(pkg/kube): continue deleting objects when one fails

adamreese:
* Continue deleting objects when one fails to minimize the risk of an
  upgrade ending in an unrecoverable state
* Exclude failed deleted object from the returned result set

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -223,6 +223,7 @@ func (c *Client) Update(original, target ResourceList, force bool) (*Result, err
 
 		if err := info.Get(); err != nil {
 			c.Log("Unable to get obj %q, err: %s", info.Name, err)
+			continue
 		}
 		annotations, err := metadataAccessor.Annotations(info.Object)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/kube/client.go
```diff
@@ -232,16 +233,11 @@ func (c *Client) Update(original, target ResourceList, force bool) (*Result, err
 			c.Log("Skipping delete of %q due to annotation [%s=%s]", info.Name, ResourcePolicyAnno, KeepPolicy)
 			continue
 		}
-
-		res.Deleted = append(res.Deleted, info)
 		if err := deleteResource(info); err != nil {
-			if apierrors.IsNotFound(err) {
```
bacongobbler, line 238: do we still want to check for this error type, or is the error message returned from `err` more than sufficient to handle this case?

I ask this because I do recall certain kubernetes API errors being quite cryptic in the past, which is why we were doing the error type check in the first place.

pkg/kube/client.go
```diff
@@ -232,16 +233,11 @@ func (c *Client) Update(original, target ResourceList, force bool) (*Result, err
...
-				c.Log("Attempted to delete %q, but the resource was missing", info.Name)
-			} else {
-				c.Log("Failed to delete %q, err: %s", info.Name, err)
-				return res, errors.Wrapf(err, "Failed to delete %q", info.Name)
-			}
+			c.Log("Failed to delete %q, err: %s", info.ObjectName(), err)
+			continue
 		}
+		res.Deleted = append(res.Deleted, info)
```
bacongobbler, line 240: Nice cleanup!

For context, I think we added the `.Deleted` object beforehand because we wanted to make sure the sum of `Added + Updated + Deleted` equaled the sum of objects, but I can't be totally sure. @thomastaylor312 do you remember if there was any particular reason we pre-emptively added the objects to the `Deleted` list? If so does this look good to you?

bacongobbler (approved):
I had two small comments but they are non-blocking. I'd let others weigh in here first before merging but the code here looks great.

LGTM!

PR #7932, opened 2020-04-15
Add checking of length of resourceList before install of uninstall

### LINKED ISSUE #7537: Error/panic when installing/deleting chart with only CRDs
A chart with only CRDs fails to install, and if running with `--debug`, also panics.

Output of `helm version`:
```
version.BuildInfo{Version:"v3.0.3", GitCommit:"ac925eb7279f4a6955df663a0128044a8a6b7593", GitTreeState:"clean", GoVersion:"go1.13.6"}
```

Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.2", GitCommit:"66049e3b21efe110454d67df4fa62b08ea79a19b", GitTreeState:"clean", BuildDate:"2019-05-16T16:23:09Z", GoVersion:"go1.12.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"14+", GitVersion:"v1.14.10-gke.17", GitCommit:"bdceba0734835c6cb1acbd1c447caf17d8613b44", GitTreeState:"clean", BuildDate:"2020-01-17T23:10:13Z", GoVersion:"go1.12.12b4", Compiler:"gc", Platform:"linux/amd64"}
```

GKE

Installation/deletion:
```
$ helm install crd-test . --debug
install.go:149: [debug] Original chart version: ""
install.go:166: [debug] CHART PATH: <redacted>

client.go:88: [debug] creating 1 resource(s)
client.go:88: [debug] creating 1 resource(s)
client.go:88: [debug] creating 1 resource(s)
client.go:88: [debug] creating 1 resource(s)
client.go:88: [debug] creating 1 resource(s)
install.go:139: [debug] Clearing discovery cache
wait.go:51: [debug] beginning wait for 5 resources with timeout of 1m0s
client.go:88: [debug] creating 0 resource(s)
Error: no objects visited
helm.go:75: [debug] no objects visited
helm.sh/helm/v3/pkg/kube.init
        /home/circleci/helm.sh/helm/pkg/kube/client.go:50
runtime.doInit
        /usr/local/go/src/runtime/proc.go:5222
runtime.doInit
        /usr/local/go/src/runtime/proc.go:5217
runtime.doInit
        /usr/local/go/src/runtime/proc.go:5217
runtime.main
        /usr/local/go/src/runtime/proc.go:190
runtime.goexit
```

If I then delete the release, same:
```
$ helm delete crd-test --debug
uninstall.go:91: [debug] uninstall: Deleting crd-test
uninstall.go:124: [debug] purge requested for crd-test
Error: uninstallation completed with 1 error(s): object not found, skipping delete
helm.go:75: [debug] uninstallation completed with 1 error(s): object not found, skipping delete
helm.sh/helm/v3/pkg/action.(*Uninstall).Run
        /home/circleci/helm.sh/helm/pkg/action/uninstall.go:132
main.newUninstallCmd.func1
        /home/circleci/helm.sh/helm/cmd/helm/uninstall.go:53
        /go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:826
        /go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:914
        /go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:864
main.main
        /home/circleci/helm.sh/helm/cmd/helm/helm.go:74
runtime.main
runtime.goexit
```

waveywaves:
fixes https://github.com/helm/helm/issues/7537

A chart being installed which only contains CRDs and not
any templates tries to install the resources by default.
The resourceList which is used in this case does not check
if there are resources present in it or not. This commit
adds checks to those particular places where we need to check
if the size of resourceList > 0 during installation and deletion.

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -145,20 +145,24 @@ func (i *Install) installCRDs(crds []chart.CRD) error {
 		}
 		totalItems = append(totalItems, res...)
 	}
-	// Invalidate the local cache, since it will not have the new CRDs
-	// present.
-	discoveryClient, err := i.cfg.RESTClientGetter.ToDiscoveryClient()
-	if err != nil {
-		return err
+	if len(totalItems) > 0 {
+		// Invalidate the local cache, since it will not have the new CRDs
+		// present.
+		discoveryClient, err := i.cfg.RESTClientGetter.ToDiscoveryClient()
+		if err != nil {
+			return err
+		}
+		i.cfg.Log("Clearing discovery cache")
+		discoveryClient.Invalidate()
+		// Give time for the CRD to be recognized.
+
+		if err := i.cfg.KubeClient.Wait(totalItems, 60*time.Second); err != nil {
+			return err
+		}
+
+		// Make sure to force a rebuild of the cache.
+		discoveryClient.ServerGroups()
 	}
-	i.cfg.Log("Clearing discovery cache")
-	discoveryClient.Invalidate()
-	// Give time for the CRD to be recognized.
-	if err := i.cfg.KubeClient.Wait(totalItems, 60*time.Second); err != nil {
-		return err
-	}
-	// Make sure to force a rebuild of the cache.
-	discoveryClient.ServerGroups()
 	return nil
 }
 
@@ -265,7 +269,7 @@ func (i *Install) Run(chrt *chart.Chart, vals map[string]interface{}) (*release.
 	// we'll end up in a state where we will delete those resources upon
 	// deleting the release because the manifest will be pointing at that
 	// resource
-	if !i.ClientOnly && !isUpgrade {
+	if !i.ClientOnly && !isUpgrade && len(resources) > 0 {
 		toBeAdopted, err = existingResourceConflict(resources, rel.Name, rel.Namespace)
 		if err != nil {
 			return nil, errors.Wrap(err, "rendered manifests contain a resource that already exists. Unable to continue with install")
@@ -330,11 +334,11 @@ func (i *Install) Run(chrt *chart.Chart, vals map[string]interface{}) (*release.
 	// At this point, we can do the install. Note that before we were detecting whether to
 	// do an update, but it's not clear whether we WANT to do an update if the re-use is set
 	// to true, since that is basically an upgrade operation.
-	if len(toBeAdopted) == 0 {
+	if len(toBeAdopted) == 0 && len(resources) > 0 {
 		if _, err := i.cfg.KubeClient.Create(resources); err != nil {
 			return i.failRelease(rel, err)
 		}
-	} else {
+	} else if len(resources) > 0 {
 		if _, err := i.cfg.KubeClient.Update(toBeAdopted, resources, false); err != nil {
 			return i.failRelease(rel, err)
 		}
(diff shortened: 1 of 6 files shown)


pkg/action/install.go
```diff
@@ -154,9 +154,10 @@ func (i *Install) installCRDs(crds []chart.CRD) error {
 	i.cfg.Log("Clearing discovery cache")
 	discoveryClient.Invalidate()
 	// Give time for the CRD to be recognized.
+	if len(totalItems) > 0 {
 	if err := i.cfg.KubeClient.Wait(totalItems, 60*time.Second); err != nil {
 		return err
-	}
+	}}
```
jdolitsky, line 160: Can we fix the tabs/spacing here?

pkg/kube/client.go
```diff
@@ -252,22 +252,25 @@ func (c *Client) Update(original, target ResourceList, force bool) (*Result, err
...
-			res.Deleted = append(res.Deleted, info)
-		}
-		return nil
-	})
+	if len(resources) > 0 {
+		err = perform(resources, func(info *resource.Info) error {
+			c.Log("Starting delete for %q %s", info.Name, info.Mapping.GroupVersionKind.Kind)
+			if err := c.skipIfNotFound(deleteResource(info)); err != nil {
+				mtx.Lock()
+				defer mtx.Unlock()
```
jdolitsky, line 263: appears the mutex lock can happen above the `if err` statement so its not repeated

jdolitsky (changes requested):

pkg/action/install.go
waveywaves: fixed via gofmt

pkg/kube/client.go
waveywaves: I have moved the check to uninstall instead of the client.

technosophos:
This needs some regression tests, provided we can reproduce the panic without this patch.

waveywaves:
@technosophos I have updated it with a test for regression. Can you PTAL ?

pkg/action/install.go
```diff
@@ -145,20 +145,24 @@ func (i *Install) installCRDs(crds []chart.CRD) error {
...
-	discoveryClient, err := i.cfg.RESTClientGetter.ToDiscoveryClient()
-	if err != nil {
-		return err
-	}
-	i.cfg.Log("Clearing discovery cache")
-	discoveryClient.Invalidate()
-	// Give time for the CRD to be recognized.
-	if err := i.cfg.KubeClient.Wait(totalItems, 60*time.Second); err != nil {
-		return err
+	if len(totalItems) > 0 {
```
technosophos, line 148: I do not understand this change. Is this for performance? I am worried that if another process deletes a CRD, the local cache could be outdated here.

technosophos (approved):
I think this is okay, though I don't understand the condition around the discovery cache, and would object if that was _merely_ for performance.

pkg/action/install.go
waveywaves: When testing, the `res` returned by `i.cfg.KubeClient.Build(bytes.NewBuffer(obj.File.Data), false)` is coming as `nil`, and the `err` is nil as well which means that there is no issue with building the CRDs and installing them. When `res` returns as `nil` it means that the builder used was not able to get the kube cache here. This is even reflected in this change I have made here. If the length of totalItems is 0 and still the CRD is applied then the Cache is not present on the machine under $HOME/.kube/cache/discovery.

technosophos: Sounds fine. I'll restate my LGTM for this PR.

waveywaves: Thanks Matt !

hickeyma (approved):
LGTM, thanks for this fix @waveywaves

PR #7938, opened 2020-04-16
fix: Fixed a regression that was introduced with changed nil handling

technosophos:
This fixes a test case that is causing `master` to fail.

The issue is caused by the changed handling of empty field values:

```yaml
name:
```

This used to be interpreted as `name: ""`, but with a fix to some removal code is now treated as `name: null`, which in turn is treated as `delete the field name`.

The fix was to make the `values.yaml` file read:

```yaml
name: ""
```

This is the same as the fix in Helm 2. Hooray for regression tests!

### DIFF
--- a/pkg/chartutil/create.go
+++ b/pkg/chartutil/create.go
@@ -113,7 +113,7 @@ serviceAccount:
   annotations: {}
   # The name of the service account to use.
   # If not set and create is true, a name is generated using the fullname template
-  name:
+  name: ""
 
 podAnnotations: {}
 


bacongobbler (approved):

adamreese (approved):

bacongobbler:
hold up, should the log message be removed?

pkg/lint/lint_test.go
```diff
@@ -130,6 +130,9 @@ func TestHelmCreateChart(t *testing.T) {
 	m := All(createdChart, values, namespace, true).Messages
 	if ll := len(m); ll != 1 {
 		t.Errorf("All should have had exactly 1 error. Got %d", ll)
+		for _, msg := range m {
+			t.Logf("Lint message: %s", msg)
+		}
```
mattfarina, line 135: The debug code should be removed.

mattfarina (approved):
lgtm

Issue #6711, opened 2019-10-20
Helm 3.0.0-beta4: Generation of Helm secrets is failing


HoaFox:
Hi,

I am using for the moment Helm 3.0.0-beta.4 for deploying OpenFaaS to my K8s cluster. Unfortunately I get the following error when I try to install OpenFaaS directly from the chart source:

Error: create: failed to create: Secret "sh.helm.release.v1.openfaas.v1" is invalid: data: Too long: must have at most 1048576 characters

Steps to reproduce:
1. git clone https://github.com/openfaas/faas-netes.git
2. kubectl apply -f https://raw.githubusercontent.com/openfaas/faas-netes/master/namespaces.yml
3. kubectl -n openfaas create secret generic basic-auth --from-literal=basic-auth-user=admin --from-literal=basic-auth-password=admin
4. helm install openfaas faas-netes/chart/openfaas/ --namespace openfaas --set basic_auth=true --set functionNamespace=openfaas-fn

marcopaga:
I'm having the same issue with a reasonable sized Chart consisting of multiuple services. Here the debug output:

```
install.go:148: [debug] Original chart version: ""
install.go:165: [debug] CHART PATH: /Users/marco/Projects/helm-3-test

Error: create: failed to create: Secret "sh.helm.release.v1.name.v1" is invalid: data: Too long: must have at most 1048576 characters
helm.go:76: [debug] Secret "sh.helm.release.v1.name.v1" is invalid: data: Too long: must have at most 1048576 characters
create: failed to create
helm.sh/helm/v3/pkg/storage/driver.(*Secrets).Create
	/home/circleci/helm.sh/helm/pkg/storage/driver/secrets.go:160
helm.sh/helm/v3/pkg/storage.(*Storage).Create
	/home/circleci/helm.sh/helm/pkg/storage/storage.go:66
helm.sh/helm/v3/pkg/action.(*Install).Run
	/home/circleci/helm.sh/helm/pkg/action/install.go:261
main.runInstall
	/home/circleci/helm.sh/helm/cmd/helm/install.go:209
main.newInstallCmd.func1
	/home/circleci/helm.sh/helm/cmd/helm/install.go:115
github.com/spf13/cobra.(*Command).execute
	/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:826
github.com/spf13/cobra.(*Command).ExecuteC
	/go/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:914
[...]

PR #7940, opened 2020-04-17
Add checking of length of secret before pushing it.

khrd:
This PR is to fix a bug that has been occurring in #6711 .

If there is a large file in the chart directory, it is now prevented from pushing past the maximum number of characters in Secret.
And, returns an error message.

Note: If you have set the object limit in etcd to 1 MB or more, this change may cause problems.
I don't think it happens very often...

### DIFF
--- a/pkg/storage/driver/secrets.go
+++ b/pkg/storage/driver/secrets.go
@@ -38,6 +38,9 @@ var _ Driver = (*Secrets)(nil)
 // SecretsDriverName is the string name of the driver.
 const SecretsDriverName = "Secret"
 
+// SecretSizeLimit is the maximum number of characters in Secrets.
+const SecretSizeLimit = 1048576
+
 // Secrets is a wrapper around an implementation of a kubernetes
 // SecretsInterface.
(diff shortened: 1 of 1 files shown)


technosophos (changes requested):
Thanks for the PR. There are a few things that need to be addressed.

First, this appears to just replace one error with another error... I am not sure I understand why we would do that.

Second, this hard-codes the limit into Helm, even though the 1M limit is imposed by etcd. Not all Kubernetes control planes use `etcd`. Does anyone else (@ultimateboy perhaps?) know if non-etcd backends allow >1M objects? I don't want to impose a limit in Helm if it doesn't apply universally to all Kubernetes instances.

Third, this needs unit tests.

Fourth, the ConfigMap backend would need a parallel fix and tests, which I think should go into this PR so that the two drivers do not get out of sync.

If we can get confirmation that _all_ backends impose a 1M limit, I guess I would be okay with a slight improvement to the error messages.

khrd:
I see, you're right.
I'm sorry for not knowing this.

PR #7942, opened 2020-04-17
Updating get scripts to skip pre-releases

### LINKED ISSUE #7941: get-helm-3 script download by default the release-candidate
Hello,

The get-helm-3 script download the latest v3. tags even if they are in release candidate state.
I would expect it to download only the stable releases.

curl -s https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3 | bash

Downloading https://get.helm.sh/helm-v3.2.0-rc.1-linux-amd64.tar.gz

What do you guys think?
Alban

mattfarina:
A recent change to the get scripts causes them to pickup
pre-releases in addition to stable releases. This update causes
only stable releases to be fetched by the get scripts.

Fixed #7941

### DIFF
--- a/scripts/get
+++ b/scripts/get
@@ -82,9 +82,9 @@ checkDesiredVersion() {
     local release_url="https://github.com/helm/helm/releases"
     if type "curl" > /dev/null; then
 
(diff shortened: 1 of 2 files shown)


bacongobbler (approved):
nice! Thanks for fixing this. LGTM!

PR #7945, opened 2020-04-17
feat: add dependency tests

### LINKED ISSUE #7910: Proposal: Add Linter rule for making sure subcharts have `dependencies` in Chart.yaml
This is a proposal to add a new linting rule to `helm lint`. This rule would warn the user on cases where a subchart exists in `charts/` but is not declared in `Chart.yaml`.

Background:

In Helm 2, we added `requirements.yaml` after the major release, and thus had to keep things backward compatible until Helm 3. For that reason, it was possible to add a chart to the `charts/` folder without adding it in `requirements.yaml`.

In Helm 3, we removed most of the backward compatibility code, making it _required_ to add all subcharts to the `dependencies` section of Chart.yaml.

However, this was not enforced in code, and it is still possible to put subcharts into the chart without declaring them as dependencies, and get _some_ of the desired behavior. While we should find and fix those cases, in the meantime we should add a linter rule warning that this is not expected behavior.

Proposed Rule:

The rule should check all of the charts in the `charts/` directory and make sure they are present in the `dependencies` section of Chart.yaml.

technosophos:
This adds linters that check to make sure that dependencies are in the chart, and that the Chart.yaml points to the dependencies.

Closes #7910


cmd/helm/lint_test.go
```diff
@@ -27,7 +27,7 @@ func TestLintCmdWithSubchartsFlag(t *testing.T) {
 		name:      "lint good chart with bad subcharts",
 		cmd:       fmt.Sprintf("lint %s", testChart),
 		golden:    "output/lint-chart-with-bad-subcharts.txt",
-		wantError: false,
```
technosophos, line 30: The desired behavior for this changed in Helm 3, but this test was never updated. In Helm 2, we had to retain backward compatibility to Helm 2.0.0, which did not have a requirements.yaml, and which did not impose restrictions on the formatting of subcharts. But in Helm 3, we added the requirement that a subchart be present in the Chart.yaml, and that it be valid and well-formed. (The purpose of this issue is to update the linter to reflect Helm's actual requirements/assumptions in code)

technosophos:
I am going to rewrite the tests on this. So I just added the `wip` label.

PR #7948, opened 2020-04-18
[WIP] fix: send hook yaml into post-renderer (#7891)

### LINKED ISSUE #7891: Post-renderer not receiving chart hook manifests?
Output of `helm version`: 

```
```

Output of `kubectl version`:

```
```

```
apiVersion: v1
kind: Pod
metadata:
  name: post-render-test
  labels:
    # release.
    values: {{.Values.Name}}
spec:
  containers:
  - name: waiter
    image: "alpine:3.4"
```

and one pre-install hook:

```
apiVersion: batch/v1
kind: Job
metadata:
  name: "{{ .Release.Name }}"
  labels:
  annotations:
    "helm.sh/hook-weight": "-5"
spec:
  template:
    metadata:
      name: "{{ .Release.Name }}"
      labels:
    spec:
      restartPolicy: Never
      containers:
      - name: pre-deploy-job
        image: "alpine:3.4"
        command: ["/bin/false"]
{{- else }}
{{- end }}
```

```
#!/bin/bash

tee --append  <&0 all.yaml
```

kabakaev:
Fixes #7891.

Todo:
- [x] Rewrite `runPostRenderer()` to preserve YAML comments.
- [ ] Update unit tests.

### DIFF
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -25,12 +25,14 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/agext/levenshtein"
 	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
+	"sigs.k8s.io/yaml"
 
 	"helm.sh/helm/v3/internal/experimental/registry"
 	"helm.sh/helm/v3/pkg/chart"
@@ -44,6 +46,9 @@ import (
 	"helm.sh/helm/v3/pkg/storage/driver"
 	"helm.sh/helm/v3/pkg/time"
 )
+
+// Filename string for objects added by a post-renderer.
+const postRendererNewObjectFileName = "added-by-post-renderer.yaml"
 
 // Timestamper is a function capable of producing a timestamp.Timestamper.
 //
@@ -159,6 +164,13 @@ func (cfg *Configuration) renderResources(ch *chart.Chart, values chartutil.Valu
 	}
 	notes := notesBuffer.String()
 
+	if pr != nil {
+		files, err = runPostRenderer(files, pr)
+		if err != nil {
+			return hs, b, notes, err
+		}
+	}
+
 	// Sort hooks, manifests, and partials. Only hooks and manifests are returned,
 	// as partials are not used after renderer.Render. Empty manifests are also
 	// removed here.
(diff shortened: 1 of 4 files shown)


kabakaev:
@technosophos, @thomastaylor312, could you please take a look. 
I tried to implement [your suggestion from the corresponding issue](https://github.com/helm/helm/issues/7891#issuecomment-611780768).

In case you read this via email, i've also asked some questions [in the PR](https://github.com/helm/helm/pull/7948#issue-405519416).

pkg/action/install.go
```diff
@@ -584,14 +592,107 @@ func (c *Configuration) renderResources(ch *chart.Chart, values chartutil.Values
...
+			postProcessedFiles[fileName] = string(document)
 		}
 	}
 
-	return hs, b, notes, nil
+	return postProcessedFiles, nil
+}
+
+// yamlSetTemporaryAnnotation adds or cuts out an annotation text in yaml kubernetess object
+func yamlSetTemporaryAnnotation(document, annotationName, annotationValue string) ([]byte, string, error) {
```
thomastaylor312, line 655: This function name is a bit misleading as it both sets and removes annotations. I would split this functionality into 2 separate functions

pkg/action/install.go
```diff
@@ -584,14 +592,107 @@ func (c *Configuration) renderResources(ch *chart.Chart, values chartutil.Values
...
+// postProcessFiles sends files for post-processing.
+// Some post-processing may need all objects at once.
+// Therefore we concatenate all rendered files, post-process and split them back.
+// A temporary annotation is added to preserve template filename.
+func (c *Configuration) postProcessFiles(files map[string]string, pr postrender.PostRenderer) (map[string]string, error) {
+	if pr == nil {
+		return files, nil
+	}
+
+	annotationName := "helmPostProcessFileName"
```
thomastaylor312, line 607: This should be a `const` and not defined in the function. It should probably also be prefaced with a `helm.sh/`

pkg/action/install.go
```diff
@@ -584,14 +592,107 @@ func (c *Configuration) renderResources(ch *chart.Chart, values chartutil.Values
...
+
+	// Split the rendered files into documents and add the temp annotation.
+	reYamlDocumentSeparator := regexp.MustCompile(`\n---\s*\n`)
+	for fileName, fileContent := range files {
+		for _, document := range reYamlDocumentSeparator.Split(fileContent, -1) {
+			if strings.TrimSpace(document) == "" {
+				continue
+			}
+			b.WriteString("---\n")
+			documentYaml, _, err := yamlSetTemporaryAnnotation(document, annotationName, fileName)
```
thomastaylor312, line 619: This is an interesting approach. My one concern about this is what happens if the post-render step mucks with the annotations (either removing unknown annotations or setting their own). Are there any other ways we could do this without an annotation?

pkg/action/install.go
```diff
@@ -584,14 +592,107 @@ func (c *Configuration) renderResources(ch *chart.Chart, values chartutil.Values
...
+
+	// Run post-processing.
+	b, err := pr.Run(b)
+	if err != nil {
+		return nil, errors.Wrap(err, "error while running post render on files")
+	}
+
+	postProcessedFiles := make(map[string]string)
+	// Split the post-processed stream into files.
+	for _, document := range strings.Split(b.String(), "---\n") {
```
thomastaylor312, line 635: Should this be using the regex from above to split?

pkg/action/install.go
```diff
@@ -584,14 +592,107 @@ func (c *Configuration) renderResources(ch *chart.Chart, values chartutil.Values
...
+	var documentMap map[string]interface{}
+	err := yaml.Unmarshal([]byte(document), &documentMap)
+	if err != nil {
+		return nil, "", err
+	}
+
+	// Take the annotations map.
+	metadata, ok := documentMap["metadata"].(map[string]interface{})
+	if !ok {
+		return nil, "", fmt.Errorf("document has no metadata: %s", document)
```
thomastaylor312, line 666: If there is no metadata for some weird reason, we shouldn't need to fail here. Instead I'd initialize an empty `map[string]interface{}` for the metadata

pkg/action/install.go
```diff
@@ -584,14 +592,107 @@ func (c *Configuration) renderResources(ch *chart.Chart, values chartutil.Values
...
+			return nil, "", fmt.Errorf("annotation %s value is not a string: %v", annotationName, annotations[annotationName])
+		}
+		delete(annotations, annotationName)
+	} else if annotationValue != "" { // Add
+		if !ok {
+			annotations = make(map[string]interface{})
+		}
+		annotations[annotationName] = annotationValue
+	} else { // Nothing to remove.
+		return nil, "", fmt.Errorf("annotation %s is not set and therefore cannot be deleted", annotationName)
```
thomastaylor312, line 684: When splitting this out to its own function for removing, you probably don't need to return an error here. I don't think it indicates any real problem if there is no annotation to remove

thomastaylor312 (changes requested):
Thanks for working on this! There are a few changes requested here. Also, if you could add a simple post render test that makes sure it gets hooks from the chart, that would be great

thomastaylor312:
Also, both of these points:
> func (c *Configuration) postProcessFiles( or func postProcessFiles(? Since this function is used only once, it's not really a separate piece of code. It just makes the main function more readable (smaller). Thus the c may be used in future refactoring, such as with c.Log(). As of now, c is not used.

> Variable and function names are quite verbose. Should they be shortened? Suggestions are welcome :)

I think everything is fine as is.

The documentation will need to be updated when this is merged. It can be found at: https://github.com/helm/helm-www/blob/master/content/en/docs/topics/advanced.md

pkg/action/install.go
kabakaev: The options are:
1. Run the post-renderer separately for non-hook templates (`[]byte`) and for each hook.

2. Remember template filename string for each object that is being marshalled for post-processing. For example, before postprocessing, fill the `map[string]string`, where `key:=fmt.Sprintf("%s;%s;$s", apiVersion, kind, metaName)` and value is the filename. Upon unmarshalling, use that map to find the filename. If not found, assign some random string.

3. Use [any other field from metadata](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#objectmeta-v1-meta), such as name or label. Unfortunately, other fields have more strict limitations on length and a set of allowed characters. IMHO, `annotations` is the best option there.

4. Use annotations, but if our `helm.sh/separator` annotation is not found, then "adopt" such object with a random file name. Are the original template filenames important after all? Don't know how they usedapart from a comment in the final yaml.

5. Mix options 2 and 4 by running post-renderer twice. Use first run with extra annotation to obtain extra mapping. In second run, post-render without extra annotations and assign a filename which best matches both mappings.

Option 4 is implemented now.

Option 5 may be dangerous if a user's postrenderer script is not idempotent.

I personally like the second option too. `kind` almost always stays the same. `apiVersion` often gets changed to a similar value. `metadata.name` is usually prefixed or suffixed with a string. So it may worth a try. I just don't know (yet) how to implemement a string similarity algorithm which works good for `metadata.name` and especially for `apiVersion` comparison.

pkg/action/install.go
kabakaev: I though initially that these document separators are guaranteed to be `"---\n"`, because they are added in the same function a few lines above. So I thought `strings.Split()` would save a few CPU cycles compared to `regexp.Split()`.
But now when you asked, it is indeed possible that post-renderer adds own separators that may have whitespaces, which may indent (and invalidate) the line after separator in case a simple `strings.Split()` is used. You're right, `regexp.Split()` must be used there.

pkg/action/install.go
kabakaev: I already shoot myself in the foot because of that error, when Kustomize post-renderer added a new networkPolicy object  :)

kabakaev:
@thomastaylor312, PTAL! The post-renderer PR is updated according to your suggestions.

The test was also added:
```bash
# git checkout hook-post-renderer
# go test ./pkg/action/...
ok      helm.sh/helm/v3/pkg/action      0.090s
## revert changes to current master
# wget https://github.com/helm/helm/raw/master/pkg/action/action.go -O ./pkg/action/action.go
# go test ./pkg/action/...
---
# Source: hello/templates/hello
hello: world

--- FAIL: TestInstallRelease_PostRendererReceivesHooks (0.00s)
    install_test.go:319: Failed install: error while running post render on files: Hook was not post-processed
```

It seems the temporary annotation way works fine for a couple of "beta-testers" on the original issue #7891. But if you find it useful to add the failback mechanism described above in option `2. Remember template filename string...`, then I'd suggest to add it as a separate PR. Otherwise let's postpone it till someone claims a real-world usage scenario which gets broken by the temp annotation way.

pkg/action/action.go
```diff
@@ -210,14 +222,144 @@ func (c *Configuration) renderResources(ch *chart.Chart, values chartutil.Values
...
+	// Unmarshal the document.
+	var documentMap map[string]interface{}
+	err := yaml.Unmarshal([]byte(document), &documentMap)
+	if err != nil {
+		return nil, "", err
+	}
+
+	// Take the annotations map.
+	metadata, ok := documentMap["metadata"].(map[string]interface{})
+	if !ok {
```
thomastaylor312, line 327: Hopefully this never happens, but I wouldn't error here. If the yaml got so messed up that this doesn't have a metadata any more, we can let the kubernetes serialization catch it

pkg/action/action.go
```diff
@@ -210,14 +222,144 @@ func (c *Configuration) renderResources(ch *chart.Chart, values chartutil.Values
...
+		return nil, "", fmt.Errorf("document has no metadata: %s", document)
+	}
+
+	annotations, ok := metadata["annotations"].(map[string]interface{})
+
+	annotationValue := ""
```
thomastaylor312, line 333: Nit: This variable would be more descriptive for future readers if it was called `fileName`

pkg/action/action.go
```diff
@@ -210,14 +222,144 @@ func (c *Configuration) renderResources(ch *chart.Chart, values chartutil.Values
...
+		annotationValue, ok = annotations[tempAnnotationName].(string)
+		if !ok {
+			return nil, "", fmt.Errorf("annotation %s value is not a string: %v", tempAnnotationName, annotations[tempAnnotationName])
+		}
+		delete(annotations, tempAnnotationName)
+	} else {
+		// Nothing to remove. Either post-renderer removed the annotation
+		// or a new k8s object was added without such annotation.
+		// Let's add a fake file name.
+		annotationValue = fmt.Sprintf("non-existent-%d.yaml", rand.Intn(9999))
```
thomastaylor312, line 345: This should probably use a hash of the data or a UID of some kind. It is unlikely, but you can still have a chance of a number in a random set repeating and causing a collision when trying to write out files later on

thomastaylor312 (changes requested):
Almost there! Just a few more small things, and then I'll need another maintainer to review due to the size

mattfarina:
Since the bug is in the 3.3.0 milestone I moved the PR there as well.

mattfarina:
@kabakaev Can you make the changes that @thomastaylor312 suggested?

bacongobbler:
3.3.0-rc.1 is being cut tomorrow, so we're removing this from the milestone.

pkg/action/action.go
kabakaev: Good point. Let's return the document as is.

pkg/action/action.go
kabakaev: sha1 is used instead.

thomastaylor312 (approved):
This is looking good to go! Thanks for addressing all the changes. This will need 1 more maintainer review before it can be merged

technosophos (changes requested):
It appears that the method proposed here strips all comments out of YAML before it is sent to the renderer. If a template author inserts something meaningful in the comments (like YTT templates) that are meant to be used by the post-render, it appears that these are lost before the post-render step. The ability to do this was one of the selling points of the post-renderer step.

If comments are stripped, this is a breaking change.

Any solution that can be merged in the Helm 3 cannot have breaking changes. So this PR would need to be refactored to avoid that behavior.

kabakaev:
@technosophos, thanks for review. 

I was not aware `ytt` exists let alone [it uses YAML comments for templating](https://get-ytt.io/#example:example-k8s-helm-ish).

I'll try to refactor the PR accordingly.

technosophos:
If you don't mind delaying for a few days, @kabakaev , I think that the real underlying problem is that the `renderResources` function is just doing too many things. I would like to look to see if there is a sensible way to refactor it that would make it much easier for changes like yours.

kabakaev:
That sounds great. It's interesting to see what would be a refactored `renderResources`.

technosophos:
A quick update: I am about 50% of the way through a refactor of renderResources.

will-beta:
any progress?

bhiravabhatla:
Any update on this?

ron1:
Still waiting for an update on this.

renier:
@technosophos Is this held up waiting on [the renderResources refactor](https://github.com/helm/helm/pull/7948#issuecomment-664608458)? or is it waiting on @kabakaev to ensure [comments are not stripped](https://github.com/helm/helm/pull/7948#issuecomment-664606643).

technosophos:
I have given up on the render refactor between now and Helm 4 because the underlying code keeps changing too quickly. But we definitely cannot merge the PR if it strips comments.

master4044:
@technosophos @kabakaev What is the plan for this change? @kabakaev are you still working on finding a way to not strip the comments?

I'm working with helm and Kustomize where I'm attempting to insert a value into jobs created by helm. I'm doing this to enforce the cleanup of those jobs if the creator of the job didn't set a ttlSecondsAfterComplete themselves. Without the post-render hook working, helm simply ignores the modification to the yaml.

ThomasSteinbach:
So disappointing to me. I rebuilt my complete workflow using the Helm post-renderer just to recognize 10% of the K8s objects are not processed - and then seing this potential solution on the horizon like a rainbow. The closer you come, the rainbow keeps its distance 😞

Chili-Man:
@technosophos @kabakaev anything I can do to help get this change through the finish line?

Fantaztig:
Also interested in this, are the missing unit tests the only thing missing atm?

kuskoman:
I also need this change in helm, is there anything I can do to speed up merging this PR?

gaborho:
We are hindered by this limitation as well. Would be great if hook manifests would be passed over to post renderer. Would appreciate if this PR would be soon merged.

MatthewCane:
This limitation is still causing issues, is there likely to be any progress on this?

michelesr:
I'm not sure this work is complete:

```
$ helm version                                                                 
version.BuildInfo{Version:"v3.7+unreleased", GitCommit:"726556c6ad9dc774c5a5d1bdfe513017f2d727c1", GitTreeState:"clean", GoVersion:"go1.19.2"}

$ helm template --debug chart --post-renderer cat                                  
install.go:178: [debug] Original chart version: ""
install.go:199: [debug] CHART PATH: /home/michele/Projects/foo/chart

panic: runtime error: index out of range [1] with length 1

goroutine 1 [running]:
helm.sh/helm/v3/pkg/action.runPostRenderer(0xc00056b658?, {0x2930d60, 0xc00060a880})
        helm.sh/helm/v3/pkg/action/action.go:278 +0x60c
helm.sh/helm/v3/pkg/action.(*Configuration).renderResources(0xc000318880, 0xc0000e3900, 0x7?, {0x2657651, 0xc}, {0x0, 0x0}, 0x0, 0x0, 0x0, ...)
        helm.sh/helm/v3/pkg/action/action.go:168 +0x5e5
helm.sh/helm/v3/pkg/action.(*Install).RunWithContext(0xc00033c580, {0x29487e0?, 0xc000485800}, 0x2?, 0x1?)
        helm.sh/helm/v3/pkg/action/install.go:259 +0x77d
main.runInstall({0xc000319bc0?, 0x1, 0x4}, 0xc00033c580, 0xc00068fa80?, {0x2931ba0?, 0xc000014018})
        helm.sh/helm/v3/cmd/helm/install.go:265 +0x8d9
main.newTemplateCmd.func2(0xc000004500?, {0xc000319bc0?, 0x4?, 0x4?})
        helm.sh/helm/v3/cmd/helm/template.go:82 +0x22a
github.com/spf13/cobra.(*Command).execute(0xc000004500, {0xc000319b80, 0x4, 0x4})
        github.com/spf13/cobra@v1.2.1/command.go:856 +0x67c
github.com/spf13/cobra.(*Command).ExecuteC(0xc000673900)
        github.com/spf13/cobra@v1.2.1/command.go:974 +0x3bd
main.main()
        helm.sh/helm/v3/cmd/helm/helm.go:87 +0x235
```

You can use the chart at this repo to reproduce: https://github.com/michelesr/skaffold-test-repo

joejulian:
Doesn't this completely change the api for post-renderer? Prior to this change, post-renderer could be used without it providing any output. If I'm reading this right, this change will require the post-renderer to output yaml which helm will then reinterpret. Am I reading this wrong?

fuog:
I am also interested and would be happy if the change is adopted. Thanks a lot for all who have contributed here

kabakaev:
Closing ins favor of https://github.com/helm/helm/issues/7891#issuecomment-1507430101

PR #7950, opened 2020-04-19
feat(getter): add timeout option

hiddeco:
To allow finer grain control over the request when utilizing `getter` as a package.

Addresses item 1 of #7958.

### DIFF
--- a/pkg/getter/httpgetter.go
+++ b/pkg/getter/httpgetter.go
@@ -119,6 +119,7 @@ func (g *HTTPGetter) httpClient() (*http.Client, error) {
 
 	client := &http.Client{
 		Transport: transport,
+		Timeout:   g.opts.timeout,
 	}
 
 	return client, nil
--- a/pkg/getter/getter.go
+++ b/pkg/getter/getter.go
@@ -18,6 +18,7 @@ package getter
 
 import (
 	"bytes"
+	"time"
 
 	"github.com/pkg/errors"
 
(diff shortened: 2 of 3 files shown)


technosophos:
Is there an issue or proposal attached to this PR?

hiddeco:
@technosophos there now is!

technosophos (approved):
I'm good with the proposed PR,but just want to make sure that @bacongobbler is also happy with this one as it is now.

pkg/getter/httpgetter.go
```diff
@@ -90,7 +90,9 @@ func NewHTTPGetter(options ...Option) (Getter, error) {
 }
 
 func (g *HTTPGetter) httpClient() (*http.Client, error) {
-	if (g.opts.certFile != "" && g.opts.keyFile != "") || g.opts.caFile != "" {
+	client := http.DefaultClient
+	switch {
+	case (g.opts.certFile != "" && g.opts.keyFile != "") || g.opts.caFile != "":
```
technosophos, line 95: I am not entirely sure this was correct in the old version or the new version. I was under the impression that InsecureSkipVerify did not mean that the certificates should be completely omitted -- just that they would not be verified against the CA file (or against the host-oriented fields). But are they completely unused altogether?

This shouldn't block this PR, since this PR is a more idiomatic version of the old code... but if anyone knows more than me about this, it would be good to get an opinion over whether certs vs InsecureSkipVerify is correct.

bacongobbler: This was also pointed out in #8031 if you want to have a look there :)

bacongobbler (approved):
I'm happy with this change. I just wanted to make sure we weren't missing something else. I'm more than happy to introduce this to the SDK as a toggle-able feature. LGTM!

jdolitsky (approved):

PR #7720, opened 2020-03-02
Make the charts cache safe in presence of several Helm instances

misha-ridge:
If several instances of Helm are run at the same moment and try to download the
same chart, some of them might see an empty or incomplete file in cache. Prevent
that by saving the dowloaded file atomically.

Closes #7600

---
This change is [Reviewable](https://reviewable.io/reviews/helm/helm/7720)

Issue #2998, opened 2017-10-03
helm dependency update fails after helm upgrade


nsitbon:
Hi

I'm using Helm from Concourse pipeline I just upgraded Helm to 2.6.1 (from 2.4.2) and my pipeline doesn't work anymore.
The culprit line seems to be https://github.com/kubernetes/helm/blob/v2.6.1/pkg/downloader/manager.go#L220
```console
$ helm dependency update
Hang tight while we grab the latest from your chart repositories...
...Successfully got an update from the "ci" chart repository
...Successfully got an update from the "local" chart repository
...Successfully got an update from the "stable" chart repository
Update Complete. ⎈Happy Helming!⎈
Error: Unable to move current charts to tmp dir: rename /tmp/build/put/staging-helm/cvhub-api/charts /tmp/build/put/staging-helm/cvhub-api/tmpcharts: invalid cross-device link
```
I cannot change the path used by Concourse  but if I do the same command manually on my mac from my home directory (i.e not from /tmp) it works:
```console
% ~/Downloads/helm dependency update
Hang tight while we grab the latest from your chart repositories...
...Successfully got an update from the "ci" chart repository
...Successfully got an update from the "local" chart repository
...Successfully got an update from the "stable" chart repository
Update Complete. ⎈Happy Helming!⎈
Saving 1 charts
Downloading elasticsearch-singlenode from repo http://127.0.0.1:8879
±|develop ✘| cvhub-api
% pwd
/Users/nsitbon/github/cvhub-api/scripts/helm/cvhub-api
```
As far as I understand the error and the related code it seems that you should not use Rename rather you should use Move.

Any help would be appreciated.

Best regards.

redbaron:
It seems that either source or destination is mount point, therefore rename doesn’t work. If you change your concourse pipeline so that both source and destination are on the same mount point it should work

nsitbon:
I know and like I said I'm just a Concourse user I cannot change how it works and how it mounts directory. However I expect Helm working whatever the configuration like before.

bacongobbler:
This should be a relatively quick fix. I know what’s causing this. I’ll make a PR this morning.

bacongobbler:
For now you should downgrade to v2.5.x.

nsitbon:
This is exactly what I did. Thx @bacongobbler

thomastaylor312:
@bacongobbler Did you end up opening a PR for this? Is it a bug that should be pulled into 2.7?

[...]

PR #7954, opened 2020-04-20
fix: write index.yaml file atomically

### LINKED ISSUE #7988: `index.yaml` are not written atomically
`index.yaml` files, as generated by `helm repo index`, are not written atomically. As a result files can be half-written when helm or another client reads them.

This is an example of a custom YAML scanner breaking because of a half-written file:
```
/<redacted>/path/to/scanner.py", line 319, in scan_index
"could not find expected ':'", self.get_mark())
yaml.scanner.ScannerError: while scanning a simple key
in "<unicode string>", line 3497, column 5:
appVersi
^
could not find expected ':'
in "<unicode string>", line 3497, column 13:
appVersi
^
```

rabadin:
This refactors the already-existing `AtomicWriteFile` utility
to a central location and uses it to write index files
atomically.
This is done to avoid having half-written index files break
client requests.

Drive-bys:
  - Add test for AtomicWriteFile.
  - Add test IndexFile.WriteFile.

Closes #7988

### DIFF
--- a/internal/fileutil/fileutil.go
+++ b/internal/fileutil/fileutil.go
@@ -21,6 +21,8 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+
+	"helm.sh/helm/v3/internal/third_party/dep/fs"
 )
 
 // AtomicWriteFile atomically (as atomic as os.Rename allows) writes a file to a
@@ -45,5 +47,5 @@ func AtomicWriteFile(filename string, reader io.Reader, mode os.FileMode) error 
 		return err
 	}
 
-	return os.Rename(tempName, filename)
+	return fs.RenameWithFallback(tempName, filename)
 }
--- a/internal/fileutil/fileutil.go
+++ b/internal/fileutil/fileutil.go
@@ -1,0 +1,49 @@
+/*
+Copyright The Helm Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


technosophos:
Is there an issue or existing proposal attached to this?

rabadin:
@technosophos Thanks for looking at this! No, I didn't file an issue because I thought the problem was fairly transparent: it's the usual problem when clients access a half-written file and break. I'm happy to file an issue if that helps. Please let me know if you need me to do this or if this PR and its description is enough.

technosophos:
Yeah -- please file an issue and link to it. We're trying to get a little better at our bookkeeping.

rabadin:
@technosophos Sure; that's done now (see #7954).

internal/fileutil/fileutil.go
```diff
@@ -0,0 +1,49 @@
...
+
+	if err := tempFile.Close(); err != nil {
+		return err
+	}
+
+	if err := os.Chmod(tempName, mode); err != nil {
+		return err
+	}
+
+	return os.Rename(tempName, filename)
```
bacongobbler, line 48: Can you use `RenameWithFallback` in pkg `internal/third_party/dep/fs` instead? Otherwise this will result in the same cross-device errors seen in #2998 and #7987.

bacongobbler (changes requested):

bacongobbler:
You may also want to remove `atomicWriteFile` in favour of this package, which was introduced in #7720.

internal/fileutil/fileutil.go
rabadin: Good point. Done.

bacongobbler (approved):
needs another maintainer's approval before merging, but this looks good to me!

fixes #7988 
fixes #7987

technosophos (approved):
Looks great.

technosophos:
I believe it was #7988 that this was supposed to correct, so I am taking the liberty of updating the original issue post, and then I am merging this for 3.3.0

technosophos:
Based on #7987, I have marked this for inclusion on 3.2.1 instead of 3.3.0, as it is a bug fix by that definition.

Issue #3141, opened 2017-11-15
add --values and --set flag to helm package


sslavic:
Just like install and upgrade support --values flag, please support the same for helm package command.

Resulting package archive should contain parent chart's values file merged with any value files provided via --values flag.

bacongobbler:
related request: #2566

adshmh:
I have started work on this.
Considering the comments on #2566, are we looking for --values or --set (or both?)

ngeor:
A nice use case is described in the issue I opened ( #3198 ), in which I'd like to be able to do:

`helm package --version 1.2.3 --set image.tag 1.2.3`

This way the chart version and the docker image version are one and the same.

adshmh:
I have just completed the first draft and posted a PR for this.

adshmh:
@ngeor , @sslavic the PR I submitted adds both --set and --values options, could you please take a look?

ngeor:
Hi @adshmh , looks good; unfortunately I don't know Go but I see you covered it with the test cases and all, so it probably does what it says :) Thanks!

cabrinoob:
When will this feature be available? In 2.9.0 ?

technosophos:
Re-opening this, as #3471 needs to be backed out.

GKTheOne:
What does this do to the comments/etc in the archive package? eg, are they stripped or left alone or ??

Currently I'm trying to do the same thing with [`yq`](https://github.com/mikefarah/yq) but it just dumps the final values after processing. ie, all the comments, blank lines for organisation, etc are gone.

bacongobbler:
#3471 stripped all the comments out, which is why we needed to back that PR out. See https://github.com/kubernetes/helm/pull/3471#issuecomment-381779783 for more info

volkodavs:
Reverted from 2.10 "feat: add --set and --values options to 'helm package'" https://github.com/helm/helm/commit/7b8aae466761448522acbd3beb2a16ad2283013a

FedeBev:
Some news about this?
Thanks

bacongobbler:
Same as above; nobody's worked on a follow-up to fix the bugs in #3471 so this has not been implemented. You're more than welcome to fork Helm and compile against #3471 if you're okay with comments being stripped from the values file.

ahsannaseem:
Is this done ? I tried on versions `2.11.0` and `2.12.0` none of them seems to have `helm package --set...`

iftachsc:
[...]

PR #7963, opened 2020-04-21
 Use go-yaml v3 directly

silasdavis:
This code removes dependency on sigs.k8s.io/yaml which itself is a
mirror of github.com/ghodss/yaml.

All but one usages of ghodss/yaml were no-ops. That is they passed directly
to go-yaml v2 (gopkg.in/yaml.v2). Upgradring them to use
gopkg.in/yaml.v3 was trivial.

The one non-trivial use of ghodss/yaml was a single use of the `YAMLToJSON` function. I
have copied a minimal tree of dependency functions to chartutil/yaml to
support just that function hosted with helm.

Upgrading to yaml.v3 supports retaining comments on round-trip
serialisation which is blocking the ability for helm package to support
the --set flag see #3141.


silasdavis:
I am happy to fix up the test styles (they were copied more-or-less verbatim), but let's see if there is any chance of this being accepted first.

I'm also happy to be a code owner for this package within helm if you have such a concept.

If somehow you'd find this easier to accept packaged as it's own go module in separate repo, then... I can do that, I guess.

bacongobbler:
@silasdavis, I'm incredibly upset that you proposed this PR after I told you that we would not accept PRs that vendored code straight from ghodss/yaml. Have you tried proposing a PR to ghodss/yaml first?

bacongobbler:
I'm closing this as this goes entirely against what we discussed As discussed earlier: propose a PR upstream that would update ghodss/yaml to yaml.v3.

silasdavis:
On a point of order, this didn't go against what we had discussed, as I said in my previous comment on the issue I made this PR before reading your comment, so it merely goes against the last thing you said, there was no discussion to be had. Furthermore I do not consider this to be 'vendoring code'. This is a fork of a small subset of the code which I would argue reduces your liability. Unfortunately you don't seem willing to have that discussion. I'll move along.

PR #7970, opened 2020-04-22
Lint dependencies

### LINKED ISSUE #7910: Proposal: Add Linter rule for making sure subcharts have `dependencies` in Chart.yaml

technosophos:
This adds linter rules for dependencies.

Closes #7910


technosophos:
Might wait on this while waiting for a stronger decision about how required this behavior is. It appears that we might have mis-documented this feature for Helm 3 by carrying over too many docs from Helm 2.

pkg/lint/rules/dependencies.go
```diff
@@ -0,0 +1,82 @@
...
+//
+// See https://github.com/helm/helm/issues/7910
+func Dependencies(linter *support.Linter) {
+	c, err := loader.LoadDir(linter.ChartDir)
+	if !linter.RunLinterRule(support.ErrorSev, "", validateChartFormat(err)) {
+		return
+	}
+
+	linter.RunLinterRule(support.ErrorSev, linter.ChartDir, validateDependencyInMetadata(c))
+	linter.RunLinterRule(support.WarningSev, linter.ChartDir, validateDependencyInChartsDir(c))
```
mattfarina, line 40: There are many cases where someone would have a dependency listed in a chart but not have it in the `charts` directory. Should this be a warning?

technosophos: The issue is that there are never _supposed_ to be cases like this. Helm 3 assumes (for proper behavior) that the contents of `Chart.Metadata.Dependencies` is the same as `Chart.Dependencies()`. When these are out of sync, weird things happen.

So we have a case where the code assumes something, but never explicitly documented this assumption. However, many features of Helm need this to be the case.

My suspicion is that a lot of the issues filed about problems with deeply nested charts will simply vanish when people are forced to keep their stated dependencies in sync with the implicit dependencies they drop in `charts/`.

mattfarina (approved):
lgtm

pkg/lint/rules/dependencies_test.go
```diff
@@ -0,0 +1,103 @@
...
+	"helm.sh/helm/v3/pkg/chart"
+	"helm.sh/helm/v3/pkg/chartutil"
+	"helm.sh/helm/v3/pkg/lint/support"
+)
+
+const (
+	badChartDepsDir = "testdata/lint-deps"
+)
+
+func chartWithBadDependencies() chart.Chart {
```
adamreese, line 33: nice in-memory fixtures!!!

pkg/lint/rules/dependencies_test.go
```diff
@@ -0,0 +1,103 @@
...
+	"testing"
+
+	"helm.sh/helm/v3/internal/test/ensure"
+	"helm.sh/helm/v3/pkg/chart"
+	"helm.sh/helm/v3/pkg/chartutil"
+	"helm.sh/helm/v3/pkg/lint/support"
+)
+
+const (
+	badChartDepsDir = "testdata/lint-deps"
```
adamreese, line 30: Can this be deleted?

adamreese (changes requested):

technosophos:
Fixed and ready for re-review

pkg/lint/rules/dependencies_test.go
technosophos: Done

hickeyma (approved):
LGTM, thanks @technosophos. I am wondering should a doc PR be pushed to update https://helm.sh/docs/topics/charts/#chart-dependencies ?

adamreese (approved):

mattfarina:
@technosophos this is ready for you to merge if you're still ready for it.

technosophos:
I would like to hold it for 3.4. It is a little contentious, so I figure we let the other linter changes go through, then do this one separately. As soon as the 3.3 branch is cut, I can merge this.

pkg/lint/rules/dependencies_test.go
```diff
@@ -0,0 +1,99 @@
...
+
+	c := chartWithBadDependencies()
+	err := chartutil.SaveDir(&c, tmp)
+	if err != nil {
+		t.Fatal(err)
+	}
+	linter := support.Linter{ChartDir: filepath.Join(tmp, c.Metadata.Name)}
+
+	Dependencies(&linter)
+	if l := len(linter.Messages); l != 2 {
```
renan, line 93: Instead of just doing a count it would easier to understand what is happening by asserting on the message content as well.

bridgetkromhout:
@technosophos would you like to merge this now that the rc is cut? https://github.com/helm/helm/releases/tag/v3.3.0-rc.1

PR #7972, opened 2020-04-23
Remove utf-8 BOM when loading chart files and helmignore

### LINKED ISSUE #6081: UTF-8 BOM in a resource template fails deployment
When installing or upgrading a helm chart with a file that is encoded as UTF-8 BOM (Byte Order Mark) the deployment fails with the following message:
Error: release <name> failed: the server could not find the requested resource

You can find an example chart here: https://www.dropbox.com/s/y9qxzd1m73v6fak/test-bom-0.1.0.tgz?dl=0
Note the service.yaml is encoded in UTF-8 with BOM 

A UTF-8 BOM is usually added to the template file when the file is created in Windows

This issue is also not detected by a `helm lint`
It would help allot if either helm automatically removes the invisible BOM from the resource files and/or `helm lint` will detect the issue. 

Output of `helm version`:
Client: &version.Version{SemVer:"v2.14.2", GitCommit:"a8b13cc5ab6a7dbef0a58f5061bcc7c0c61598e7", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.14.2", GitCommit:"a8b13cc5ab6a7dbef0a58f5061bcc7c0c61598e7", GitTreeState:"clean"}

Output of `kubectl version`:
Server Version: version.Info{Major:"1", Minor:"13+", GitVersion:"v1.13.7-eks-c57ff8", GitCommit:"c57ff8e35590932c652433fab07988da79265d5b", GitTreeState:"clean", BuildDate:"2019-06-07T20:43:03Z", GoVersion:"go1.11.5", Compiler:"gc", Platform:"linux/amd64"}

EKS

IppX:
Closes #6081 

Removes the BOM prefix if present, in read files before processing the data.
Affects the following pkg:
- pkg/chart/loader: directory and archive loader
- internal/ignore: when loading .helmignore file

It's my first contribution and I'm no golang expert please be gentle :).

I initially tried another solution with [BOMOverride](https://pkg.go.dev/golang.org/x/text@v0.3.2/encoding/unicode?tab=doc#BOMOverride) to also handle UTF-16 BOM but it broke other parts that I did not understand.

Also let me know if I need to backport this to v2.x.


technosophos:
Asking out of ignorance: Is there any case where a bare BOM might appear in the middle of a file, and legitimately so? For example, if a configmap has a field that contains a complete file, and that file for some reason needs the BOM.

I looked around a little, and I can't find information on this. My guess is that a BOM in the middle of a YAML doc is against the YAML spec. But I am not sure.

mattfarina (commented):
Can we get a test to verify the BOM is in the test chart? This is to make sure a change from someone does not accidentally remove it. We want to make sure the invisible characters are not accidentally dropped.

IppX:
@technosophos 
BOMs are not allowed in the middle unless they are in quotes, according to the spec (https://yaml.org/spec/1.2/spec.html#id2771184):
>  To make it easier to concatenate streams, byte order marks may appear at the start of any document. However all documents in the same stream must use the same character encoding.
To allow for JSON compatibility, byte order marks are also allowed inside quoted scalars. For readability, such content byte order marks should be escaped on output.

IppX:
> Can we get a test to verify the BOM is in the test chart? This is to make sure a change from someone does not accidentally remove it. We want to make sure the invisible characters are not accidentally dropped.

I added a test to check the presence of BOM in the test data beforehand. Is this what you were looking for ?

technosophos (approved):

technosophos:
Based on @IppX 's last note from the YAML spec, I think this is good as-is. @mattfarina can you weigh in on the test you asked for? I was not totally clear what you were asking for there, but I _think_ the new test addresses your concern.

mattfarina (approved):
This is great. Thanks for adding the test. This will help us to not accidentally delete the BOMs in the test files.

IppX:
Thanks for taking the time to review this !

PR #7984, opened 2020-04-23
feat: make the linter coalesce the passed-in values before running values tests

### LINKED ISSUE #7756: Helm lint (v3) doesn't take values files into account during schema validation
Edit: Steps to reproduce: https://github.com/helm/helm/issues/7756#issuecomment-617889123

I have a bunch of charts with `values.schema.json` files that won't pass schema validation without passing in necessary, environment-specific values. I don't want these values in the base `values.yaml` file because they are environment-specific, but should fail at install/upgrade time.

When running `helm lint ${chart} --values ./lint-values.yaml` I'm getting errors like below **even though these values are provided in `./lint-values.yaml`**:
```
==> Linting ${chart}
[ERROR] values.yaml:
- ingress.hosts: Array must have at least 1 items
...

```

Important to note that `helm template ${chart} --values ./line-values.yaml` does not throw any schema validation errors. When I run `helm template ${chart}` I see the same errors as above:
```
Error: values don't meet the specifications of the schema(s) in the following chart(s):
agent-services:
- ingress.hosts: Array must have at least 1 items
...
```

I think this is a bug. `helm lint` should run values schema validation against the final `.Values` object like the docs say:
> Note that the schema is applied to the final `.Values` object, and not just to the `values.yaml` file. This means that the following `yaml` file is valid, given that the chart is installed with the appropriate `--set` option shown below.

Source: https://helm.sh/docs/topics/charts/#schema-files

---

Output of `helm version`:
```
version.BuildInfo{Version:"v3.1.1", GitCommit:"afe70585407b420d0097d07b21c47dc511525ac8", GitTreeState:"clean", GoVersion:"go1.13.8"}
```

Output of `kubectl version`: not relevant

Cloud Provider/Platform (AKS, GKE, Minikube etc.): not relevant

technosophos:
When support for `helm lint --values` was added, this support did not extend to the values file tests. This PR extends the support to values files, which lets the user schema-test the combination of the chart and the passed-in values.

Also added tests for other values lints.

Closes #7756 

### DIFF
--- a/pkg/lint/lint.go
+++ b/pkg/lint/lint.go
@@ -30,7 +30,7 @@ func All(basedir string, values map[string]interface{}, namespace string, strict
 
 	linter := support.Linter{ChartDir: chartDir}
 	rules.Chartfile(&linter)
-	rules.Values(&linter)
+	rules.ValuesWithOverrides(&linter, values)
 	rules.Templates(&linter, values, namespace, strict)
 	return linter
 }
--- a/pkg/lint/rules/values.go
+++ b/pkg/lint/rules/values.go
@@ -36,7 +36,7 @@ func Values(linter *support.Linter) {
 
 // ValuesWithOverrides tests the values.yaml file.
 //
-// If a schema iis present in the chart, values are tested against that. Otherwise,
+// If a schema is present in the chart, values are tested against that. Otherwise,
 // they are only tested for well-formedness.
 //
 // If additional values are supplied, they are coalesced into the values in values.yaml.
--- a/pkg/lint/rules/values.go
+++ b/pkg/lint/rules/values.go
@@ -28,7 +28,19 @@ import (
 )
 
 // Values lints a chart's values.yaml file.
+//
+// This function is deprecated and will be removed in Helm 4.
 func Values(linter *support.Linter) {
+	ValuesWithOverrides(linter, map[string]interface{}{})
+}
+
+// ValuesWithOverrides tests the values.yaml file.
+//
+// If a schema iis present in the chart, values are tested against that. Otherwise,
+// they are only tested for well-formedness.
+//
+// If additional values are supplied, they are coalesced into the values in values.yaml.
+func ValuesWithOverrides(linter *support.Linter, values map[string]interface{}) {
 	file := "values.yaml"
 	vf := filepath.Join(linter.ChartDir, file)
 	fileExists := linter.RunLinterRule(support.InfoSev, file, validateValuesFileExistence(vf))
@@ -37,7 +49,7 @@ func Values(linter *support.Linter) {
 		return
 	}
 
-	linter.RunLinterRule(support.ErrorSev, file, validateValuesFile(vf))
+	linter.RunLinterRule(support.ErrorSev, file, validateValuesFile(vf, values))
 }
 
 func validateValuesFileExistence(valuesPath string) error {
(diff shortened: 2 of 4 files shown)


pkg/lint/rules/values.go
```diff
@@ -28,7 +28,19 @@ import (
...
 // Values lints a chart's values.yaml file.
+//
+// This function is deprecated and will be removed in Helm 4.
 func Values(linter *support.Linter) {
+	ValuesWithOverrides(linter, map[string]interface{}{})
+}
+
+// ValuesWithOverrides tests the values.yaml file.
+//
+// If a schema iis present in the chart, values are tested against that. Otherwise,
```
sean-krail, line 39: `iis` -> `is`

sean-krail (changes requested):
Looks good and thanks for pushing a fix so quickly. Just found a small spelling mistake.

pkg/lint/rules/values.go
technosophos: Good catch. Thanks.

sean-krail (approved):

mattfarina (approved):
lgtm

technosophos:
I believe this needs one more maintainer review for merge.

bacongobbler (approved):

pkg/lint/lint.go
```diff
@@ -30,7 +30,7 @@ func All(basedir string, values map[string]interface{}, namespace string, strict
 
 	linter := support.Linter{ChartDir: chartDir}
 	rules.Chartfile(&linter)
-	rules.Values(&linter)
```
robbie-demuth, line 33: Are there no concerns about changing the internals of this function per https://github.com/helm/helm/pull/7417#discussion_r375961183 where I tried to make a similar change? While this function's signature isn't changing, it's output certainly might in cases where it was passed a non-empty values map, no?

robbie-demuth: (Don't get me wrong, I'm all for this change. I was just curious)

technosophos: The issue that was raised with your change (which I hadn't seen until yesterday) was that you changed the public signature of `Values()`. That is an API breaking change, and is not allowed until Helm 4. @bacongobbler was not blocking your PR based on an abstract notion of "sameness." He was blocking it based on our policy to not introduce breaking changes to the public API. It's true that we do want to make sure things continue to function similarly, but I think you may have gotten sidetracked by that. Fixing a bug is, of course, a proper behavior change. He was just making a sidelong comment that if no values were provided, it should still function the same.

FWIW, I don't have a preference over which our our PRs gets merged. So if you'd rather update yours, I'll review the changes. If I had noticed it before writing this one, I would have offered feedback then instead of doing what equates to duplicate work.

robbie-demuth: Thanks for the explanation! I'm fine with this moving forward since I'd have to resolve conflicts and you already have approvals. Thanks for making the fix!

robbie-demuth: Can you also mark that this closes https://github.com/helm/helm/issues/7273?

mattfarina: @robbie-demuth that is a good questions. Notice that the `Values` function is still present with the same API definition. There is now a new public function named `ValuesWithOverrides` with the new signature. The Go public API change was additive in the form of a new function with the new signature.

mattfarina:
@technosophos this is ready for you to merge when ready.

soad666p:
this is still a problem in helm v3.4.1

PR #7995, opened 2020-04-25
Fix read file for 'helm lint -f some-file'

### LINKED ISSUE #7985: nil pointer dereference on invalid filepath in `helm lint -f`
By running

```bash
helm lint -f valu%s.yaml
# OR
helm lint -f %
# OR
helm lint -f :
```

helm throws an SIGSEGV.
```
% helm lint -f valu%s.yaml
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x226ad1c]

goroutine 1 [running]:
helm.sh/helm/v3/pkg/cli/values.readFile(0xc0006f6e70, 0xb, 0xc000244700, 0x2, 0x2, 0x3b54008, 0x0, 0x1, 0xc000665710, 0xc0002e7b60)
        /private/tmp/helm-20200423-41927-615sa8/src/helm.sh/helm/pkg/cli/values/options.go:115 +0x6c
helm.sh/helm/v3/pkg/cli/values.(*Options).MergeValues(0xc000681bc0, 0xc000244700, 0x2, 0x2, 0xc00043abc0, 0x1, 0x1)
        /private/tmp/helm-20200423-41927-615sa8/src/helm.sh/helm/pkg/cli/values/options.go:48 +0x114
main.newLintCmd.func1(0xc00042a580, 0xc00000eea0, 0x0, 0x2, 0x0, 0x0)
        /private/tmp/helm-20200423-41927-615sa8/src/helm.sh/helm/cmd/helm/lint.go:72 +0x35f
github.com/spf13/cobra.(*Command).execute(0xc00042a580, 0xc00000ee80, 0x2, 0x2, 0xc00042a580, 0xc00000ee80)
        /private/tmp/helm-20200423-41927-615sa8/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:842 +0x460
github.com/spf13/cobra.(*Command).ExecuteC(0xc0000f1340, 0x2973300, 0xc0001049a0, 0xc0005ed964)
        /private/tmp/helm-20200423-41927-615sa8/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:950 +0x349
        /private/tmp/helm-20200423-41927-615sa8/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:887
main.main()
        /private/tmp/helm-20200423-41927-615sa8/src/helm.sh/helm/cmd/helm/helm.go:83 +0x206
```
It doesn't matter if `%` exists.

Output of `helm version`:
```
version.BuildInfo{Version:"v3.2.0", GitCommit:"e11b7ce3b12db2941e90399e874513fbd24bcb71", GitTreeState:"clean", GoVersion:"go1.13.10"}
```

OS: macOS 10.15.4

lou-lan:
Fix #7985, the description for 'helm lint -f some-file'  seems wrong, and this PR is going to fix this. @bacongobbler 

Signed-off-by: lou-lan loulan@loulan.me

### DIFF
--- a/pkg/cli/values/options.go
+++ b/pkg/cli/values/options.go
@@ -109,12 +109,21 @@ func readFile(filePath string, p getter.Providers) ([]byte, error) {
 	if strings.TrimSpace(filePath) == "-" {
 		return ioutil.ReadAll(os.Stdin)
 	}
-	u, _ := url.Parse(filePath)
+	u, err := url.Parse(filePath)
+	if err != nil {
+		if _, err := os.Stat(filePath); err == nil {
+			return ioutil.ReadFile(filePath)
+		}
+		return nil, err
+	}
 
 	// FIXME: maybe someone handle other protocols like ftp.
 	g, err := p.ByScheme(u.Scheme)
 	if err != nil {
-		return ioutil.ReadFile(filePath)
+		if _, err := os.Stat(filePath); err == nil {
+			return ioutil.ReadFile(filePath)
+		}
+		return nil, err
 	}
 	data, err := g.Get(filePath, getter.WithURL(filePath))
 	return data.Bytes(), err
--- a/pkg/cli/values/options_test.go
+++ b/pkg/cli/values/options_test.go
@@ -17,8 +17,15 @@ limitations under the License.
 package values
 
 import (
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"os"
 	"reflect"
 	"testing"
+
+	"helm.sh/helm/v3/pkg/cli"
+	"helm.sh/helm/v3/pkg/getter"
 )
 
 func TestMergeValues(t *testing.T) {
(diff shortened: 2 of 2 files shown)


technosophos (approved):
I do not believe it is correct behavior to trap a parse error and then attempt to open the result as a file handle. Please explain why proper behavior would be to try to open the file when the file name was unparsable.

Also, there is no test.

technosophos (changes requested):
Meant to click "request changes"

lou-lan:
> I do not believe it is correct behavior to trap a parse error and then attempt to open the result as a file handle. Please explain why proper behavior would be to try to open the file when the file name was unparsable.
> 
> Also, there is no test.


If `helm lint -f` specifies a local file, such as `values.yaml`, calling `url.Parse` parsing will get an error, `u` is a nil pointer. calling `p.ByScheme(u.Scheme)` will panic.
https://github.com/helm/helm/blob/2139e82c68e160a713cf19eb3ac55e91aef21cde/pkg/cli/values/options.go#L107-L121

lou-lan:
First, check whether `filePath` equals `-`, then check if it's a network file, in other cases, we should regard it as a local path.

bacongobbler:
Can you provide a test case? Perhaps that would better explain the justification for this change. I'm still not understanding why a URL parse error should result in an attempt to open the file.

lou-lan:
I added the test case, and I'm not sure if that solves it better, or want a better way to get schmea.

technosophos:
I am not suggesting that the error should panic. I am suggesting that on error, you should not just blindly ignore the error and attempt to read the file. That is a dangerous behavior. The best behavior is probably to just return the error. A second object would be to Stat the file and see if it exists, and then create an appropriate URL object.

Consider this code:

```
package main

import (
	"fmt"
	"net/url"
)

func main() {

	urls := []string{
		"/var/run/foo",
		"file:///var/run",
		"http://example.com/foo/bar",
	
	}
	for _, u := range urls {
		p, err := url.Parse(u)
		if err != nil {
			fmt.Printf("failed parse: %s\n", err)
		} else {
			fmt.Printf("Scheme of %q is %s\n", u, p.Scheme)
		}
	}
}
```

Note that even the regular file path correctly parses.

We use the URL parser for two things:

1. To sanity check the data that is passed in
2. To allow plugins to override protocol handlers

Your patch circumvents the regular scheme handling code, attempting a raw read on the file system after the URL parser has indicated that there was an error parsing the path. That's not what we intended to do.

Returning an error is fine here. Perhaps, though, there are cases where a character is illegal in a URL, but legal on the local filesystem. If you really want to run a secondary validation, you can os.Stat instead of reading... but please don't just do a raw file read and return. That's not how the code is supposed to work.

The test looks fine, but note that it passes even without your patch. What is the case that you believe your patch is fixing?

lou-lan:
> I am not suggesting that the error should panic. I am suggesting that on error, you should not just blindly ignore the error and attempt to read the file. That is a dangerous behavior. The best behavior is probably to just return the error. A second object would be to Stat the file and see if it exists, and then create an appropriate URL object.
> 
> Consider this code:
> 
> ```
> package main
> 
> import (
> 	"fmt"
> 	"net/url"
> )
> 
> func main() {
> 
> 	urls := []string{
> 		"/var/run/foo",
> 		"file:///var/run",
> 		"http://example.com/foo/bar",
> 	
> 	}
> 	for _, u := range urls {
> 		p, err := url.Parse(u)
> 		if err != nil {
> 			fmt.Printf("failed parse: %s\n", err)
> 		} else {
> 			fmt.Printf("Scheme of %q is %s\n", u, p.Scheme)
> 		}
> 	}
> }
> ```
> 
> Note that even the regular file path correctly parses.
> 
> We use the URL parser for two things:
> 
> 1. To sanity check the data that is passed in
> 2. To allow plugins to override protocol handlers
> 
> Your patch circumvents the regular scheme handling code, attempting a raw read on the file system after the URL parser has indicated that there was an error parsing the path. That's not what we intended to do.
> 
> Returning an error is fine here. Perhaps, though, there are cases where a character is illegal in a URL, but legal on the local filesystem. If you really want to run a secondary validation, you can os.Stat instead of reading... but please don't just do a raw file read and return. That's not how the code is supposed to work.
> 
> The test looks fine, but note that it passes even without your patch. What is the case that you believe your patch is fixing?

Thank you for your patience in viewing the code, it was my carelessness. I have updated the code, please check again.

cmd/demo/main.go
```diff
@@ -0,0 +1 @@
+package demo
```
jdolitsky, line 1: Can we delete this file?

jdolitsky (changes requested):

cmd/demo/main.go
lou-lan: Thanks, I deleted it.

PR #7996, opened 2020-04-25
Fix watcher may have nil

lou-lan:


technosophos (approved):

technosophos:
This needs to be tied to an issue before it can be merged.

mattfarina:
@lou-lan can you please create an issue for this that describes the reason for the change.

technosophos:
Having thought about his, I am a little worried that this may result in an unintended behavior change. Are we sure we want to short-circuit the watch in this case?

bacongobbler:
closing. Multiple maintainers have asked for changes and longer explanations on the reason for this change.

When you are ready, please open a new PR following the guidelines suggested by the other maintainers, and link back to this PR for reference. Thanks.

Issue #6705, opened 2019-10-19
lint undefined keys and "and"


dbainbri-ciena:
defaults:
    image_tag: ~

tag: '{{- if and (hasKey .Values.defaults "image_tag") (.Values.defaults.image_tag){{.Values.defaults.image_tag

The above line, with definition in values, does not pass lint strict because if complains that that the key ".Values.defaults.image_tag" is not defined.

1. It is defined as nil
2. There is a check in the first and term to make sure it is defined before it is referenced

Output of `helm version`:

2.14.3

Output of `kubectl version`:

Cloud Provider/Platform (AKS, GKE, Minikube etc.):

mattfarina:
This is now fixed in Helm v3 but not v2. Does someone want to fix it in Helm v2?

bacongobbler:
Gentle reminder that the support window for bugfixes closes in a month. If nobody works on this, we'll have to close this out and consider this as Helm 2's intended behaviour.

hickeyma:
#8516 now merged, closing.

Issue #6057, opened 2019-07-22
helm lint documentation misleading


howardjohn:
Output of `helm version`:
```
$ helm version --client
Client: &version.Version{SemVer:"v2.12.3", GitCommit:"eecf22f77df5f65c823aacd2dbd30ae6c65f186e", GitTreeState:"clean"}
```

`helm lint --help` shows this documentation:
```
      --strict                   fail on lint warnings
```

However, this seems wrong. When I run `helm lint ...`, I get no output. When I run with `helm lint --strict ...` I get error messages.

My expectation is that I get warnings, an that if I enable `--strict` those result in an exit code of 1. But the warnings should always show up!.

Demo with https://github.com/istio/installer:

```
$ helm lint istio-control/istio-discovery -f global.yaml
==> Linting istio-control/istio-discovery
Lint OK

1 chart(s) linted, no failures
$ helm lint istio-control/istio-discovery -f global.yaml --strict
==> Linting istio-control/istio-discovery
[ERROR] templates/: render error in "istio-discovery/templates/deployment.yaml": template: istio-discovery/templates/deployment.yaml:15:32: executing "istio-discovery/templates/deployment.yaml" at <include (print $.Tem...>: error calling include: template: istio-discovery/templates/configmap.yaml:52:18: executing "istio-discovery/templates/configmap.yaml" at <.Values.pilot.authPo...>: map has no entry for key "authPolicy"

Error: 1 chart(s) linted, 1 chart(s) failed
```

howardjohn:
Same with v2.14.2

hickeyma:
I think this is related to #5992 and overall issues with linting.

Aisuko:
I hit this issue on the helm v3 which I make a build from source code.

technosophos:
I have found the issue that causes this. `--strict` _also_ turns on extra flags in the template engine which will report errors if a template value evaluates to `nil`. 

There is a PR here to change that behavior. Please weigh in if you have thoughts about whether you would like to see this behavior changed: #8017

technosophos:
Closed by #8017

Issue #5992, opened 2019-07-08
lint vs lint --strict confusing


schollii:
Output of `helm version`:

2.11.0

I created a chart and wanted to find out if any of my templates were referring to non-existing values (due to typos). So after I created the chart, I purposely created a typo reference via `{{ .Values.someMap.doesNotExist` and ran helm lint: 
```
helm create test-chart
edit values.yaml etc to create someMap but not someMap.doesNotExist
helm lint 
```
This showed no error: 
```
==> Linting test-application/
[INFO] Chart.yaml: icon is recommended

1 chart(s) linted, no failures
```

Then I tried 
```
helm lint --strict  # 
```
and this failed because nameOverride and fullnameOverride are not defined in the values.yaml, because they are used in the `_helpers.tpl`:
```
==> Linting test-application/
[INFO] Chart.yaml: icon is recommended
[ERROR] templates/: render error in "test-application/templates/service.yaml": template: test-application/templates/service.yaml:6:31: executing "test-application/templates/service.yaml" at <include "chart.name"...>: error calling include: template: test-application/templates/_helpers.tpl:6:31: executing "chart.name" at <.Values.nameOverride>: map has no entry for key "nameOverride"

Error: 1 chart(s) linted, 1 chart(s) failed
```
Then once I define nameOverride and fullnameOverride the above error disappears and finally the missing key that I purposely referenced in the template causes error: 
```
==> Linting test-application/
[INFO] Chart.yaml: icon is recommended
[ERROR] templates/: render error in "test-application/templates/deployment.yaml": template: test-application/templates/deployment.yaml:43:29: executing "test-application/templates/deployment.yaml" at <.Values.http.apiUrl2>: map has no entry for key "apiUrl2"
```
Frankly this is a bit of a mess: 
- firstly there should at least be a warning if keys are not defined, even without strict
- secondly, it doesn't make sense that strict causes failure on the nameOverride and fullnameOverride because there was no warning on those without strict mode
- nameOverride and fullnameOverride should only be needed if the chart name is not defined, therefore they should not cause error even in strict mode since the chart name is defined

I think this behavior is confusing. When I lint source code, I expect that potential problems are warned, and errors show up as errors, and that strict mode would cause warnings to be interpreted as errors. New errors can appear, but not of the sort that should have been warnings (missing keys).

hickeyma:
I agree that there are issues around the 'linting'.  Do you feel like writing a PR to fix this?

technosophos:
[...]

PR #8017, opened 2020-04-28
fix: removed strict template errors in linter

### LINKED ISSUE #7483: Helm3: lint --strict broken for "default" on maps
Output of `helm version`:

Given the following values:
```
myMap:
  key1: value1
```

And the following expressions:
```
{{ default "default1" .Values.myMap.key1 }}
{{ default "default2" .Values.myMap.key2 }}
```

Renders to
```
value1
default2
```

`helm lint` is fine with that expressions. Nevertheless, `helm lint --strict` fails with

```
[ERROR] templates/: template: [templatename].yaml:5:29: executing "[templatename].yaml" at <.Values.myMap.key2>: map has no entry for key "key2"
```

technosophos:

This PR closes #7483 by no longer setting the `Engine.Strict` flag on the template engine when `helm lint --strict` is executed.

According to the documentation for `--strict`, when it is specified, it converts warnings to errors. However, in the code, `--strict=true` also sets the template engine into `Strict`. When the template engine is in strict mode, if any value is missing from the `.Values` object, it produces an error. However, templates frequently use things like `{{ default "foo" .Values.optionalThing }}` to allow users to override a default. In this case, it is likely that a value will not be present in the `Values.yaml`.

Issue #7483 reports this as a bug. I think it is open to interpretation. So I have opened a PR that changes the behavior, but we could certainly decide not to change this particular behavior. 

If we merge this PR, some charts that did _not_ pass `helm lint --strict` will now pass.

**If applicable**:

Closes #7483 

### DIFF
--- a/pkg/lint/rules/template.go
+++ b/pkg/lint/rules/template.go
@@ -81,7 +81,6 @@ func Templates(linter *support.Linter, values map[string]interface{}, namespace 
 		return
 	}
 	var e engine.Engine
-	e.Strict = strict
 	e.LintMode = true
 	renderedContentMap, err := e.Render(chart, valuesToRender)
 
--- a/pkg/lint/rules/template_test.go
+++ b/pkg/lint/rules/template_test.go
@@ -174,3 +174,55 @@ func TestDeprecatedAPIFails(t *testing.T) {
 		t.Errorf("Surprised to learn that %q is deprecated", err.Deprecated)
 	}
 }
+
+const manifest = `apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: foo
+data:
+  myval1: {{default "val" .Values.mymap.key1 }}
+  myval2: {{default "val" .Values.mymap.key2 }}
+`
+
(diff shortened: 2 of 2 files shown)


technosophos:
Closes #6705

mattfarina (approved):
I'd argue this is a bug because it doesn't follow the documentation. But, it's also a bit of a behavior change. Someone may have been expecting the behavior instead of the documented expectation.

technosophos:
I'll leave this for a bit longer in case anyone wants to weigh in on whether this is a breaking change. I agree with @mattfarina's interpretation: The docs are clear, but people _could_ be expecting the undocumented behavior, and we should be sensitive to that.

technosophos:
Closes #6057
Closes #5992

technosophos:
Closes #5417 for Helm 3 (without going the opposite route suggested in that PR)

technosophos:
Given that this PR resolves five bugs, I am going to merge it and assume that the expected behavior is NOT to change the template rendering engine.

PR #8025, opened 2020-04-29
Fix markdown table in helm command doc

dol:
The tables aren't rendered correctly because the syntax was wrong. Fixed it by applying the https://www.markdownguide.org/extended-syntax/#tables table syntax.

See https://helm.sh/docs/helm/helm/#synopsis for the wrong rendering.

Broken table in html

Fix for helm/helm-www#575

### DIFF
--- a/cmd/helm/root.go
+++ b/cmd/helm/root.go
@@ -43,17 +43,15 @@ Common actions for Helm:
 
 Environment variables:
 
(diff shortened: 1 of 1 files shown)


hickeyma (approved):
LGTM as specified in https://www.markdownguide.org/extended-syntax/#tables  for rendering into table for docs. Thanks for fixing this @dol

Issue #7710, opened 2020-03-01
Dependencies alias require the repo to be added


rocketspacer:
Output of `helm version`:            
```
version.BuildInfo{Version:"v3.1.1", GitCommit:"afe70585407b420d0097d07b21c47dc511525ac8", GitTreeState:"clean", GoVersion:"go1.13.8"}
```

Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.3", GitCommit:"06ad960bfd03b39c8310aaf92d1e7c12ce618213", GitTreeState:"clean", BuildDate:"2020-02-13T18:08:14Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"14+", GitVersion:"v1.14.9-eks-502bfb", GitCommit:"502bfb383169b124d87848f89e17a04b9fc1f6f0", GitTreeState:"clean", BuildDate:"2020-02-07T01:31:02Z", GoVersion:"go1.12.12", Compiler:"gc", Platform:"linux/amd64"}
```

Cloud Provider/Platform (AKS, GKE, Minikube etc.):  **AWS EKS**

I use fully qualified url for my Chart dependencies.

This works fine when I do `helm dep build`
```yaml
# Chart.yaml
dependencies:
  - name: argo-cd
    version: 1.8.0
    repository: https://argoproj.github.io/argo-helm
```
![image](https://user-images.githubusercontent.com/17026876/75619480-3421e200-5bb7-11ea-8b25-795fd2ebb18c.png)

However, whenever I use **alias**, like:
```yaml
# Chart.yaml
dependencies:
  - name: argo-cd
    version: 1.8.0
    repository: https://argoproj.github.io/argo-helm
    alias: argocd
```
`helm dep build` then require the repo to be added.
![image](https://user-images.githubusercontent.com/17026876/75619386-1d2ec000-5bb6-11ea-804c-c03b789b6f6f.png)

mattfarina:
I just looked at the docs and this can be a bit confusing. We either need to clean up the docs or the feature.

You can add a repo using the `helm repo add` command with a name. The `alias` here enables you to refer to that name.

mattfarina:
@rocketspacer from your understanding, how did you expect the feature to work? I'm trying to make sure I know so we can help close any gaps.

rocketspacer:
@mattfarina The behaviour I'm expecting is:
[...]

PR #8028, opened 2020-04-29
Remove hasAllRepos function

kristinnardal2:
refs #7710 

Helm doesn't download from local cache so it's redundant to check
whether they exist in local cache before downloading. This leads to bugs
where `helm dep build` doesn't work when it should.

**If applicable**:

Signed-off-by: Kristinn Björgvin Árdal <kristinnardalsecondary@gmail.com>

### DIFF
--- a/pkg/downloader/manager.go
+++ b/pkg/downloader/manager.go
@@ -113,11 +113,6 @@ func (m *Manager) Build() error {
 		}
 	}
 
-	// Check that all of the repos we're dependent on actually exist.
-	if err := m.hasAllRepos(lock.Dependencies); err != nil {
-		return err
-	}
-
 	if !m.SkipUpdate {
 		// For each repo in the file, update the cached copy of that repo
 		if err := m.UpdateRepositories(); err != nil {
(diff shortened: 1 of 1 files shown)


technosophos:
At one point, it was required that all repositories had to be locally pulled before a dependency could be correctly managed. I think this removes one area where that behavior was enforced. I need to take a closer look at the code to see if it is indeed globally optional now (in which case we can merge this PR as-is), or whether there is still a case where this is required.

I suspect that when we added the OCI registry support, we removed the last of the hard dependencies on a local cached copy. But again -- I will verify. It just may take me a while.

technosophos:
So far, this issue has helped us turn up a longstanding bug in `helm dep update`. It may be the case that `helm dep update` should require a valid local repo just like `helm dep build` does. Ironically, `helm dep build` may not need the local repo if the lock file is present.

technosophos (changes requested):
Marking this as blocked until #8118 is resolved. Currently, there are no changes necessary to this PR, but if it is merged, it will make #8118 worse by introducing the same problem in `helm dep build` when a lockfile is present.

hickeyma:
@technosophos FYI Issue #8036 . It is relevant to the investigations that you are doing.

PR #8042, opened 2020-05-01
bump version to v3.3

marckhouzam:

The master branch still had version 3.1 for the helm binary.

I assume that since 3.2 is released, master should show 3.3.


marckhouzam:
Where is the doc describing how to perform a release?
I would like to see if changing the version is mentioned before creating the release branch.

mattfarina:
The release checklist can be found at https://helm.sh/docs/community/release_checklist/. The change to v3.2 should be pulled to master. See [this](https://helm.sh/docs/community/release_checklist/#2-change-the-version-number-in-git). Looks like we missed a step.

@bacongobbler do you know why master isn't on v3.3 instead of needing to be v3.2 as the docs say?

bacongobbler:
It's all for issue triage. Way back in early Helm days, we had clients running varying versions of Helm with their own custom patches on top of stable Helm releases. It was much quicker at a glance to determine that if the version output was `3.2+unreleased`, they were using Helm 3.2 as their base branch, and we could try to reproduce their issue using Helm 3.2. If the issue could not be reproduced, then it was likely something to do with their patchset.

Either 3.2 or 3.3 works. If we're following the same pattern as Helm 2, 3.2 should be the correct version number.

marckhouzam:
If it is more helpful to have the same version on master as the latest release, then I'll update my patch.

With that in mind, would it be a reasonable idea to update the release process to first change the version and then create the release branch?  This would avoid having to cherry-pick the version change (and sometimes forgetting 😄 ).

marckhouzam:
This new commit is a cherry-pick of from the release branch

technosophos (approved):
With the change to 3.2, this looks good.

PR #8048, opened 2020-05-04
scripts: do not use optional 'which' command in get-helm installation

nixpanic:
When installing Helm, the following warning gets printed on the console:

    Downloading https://get.helm.sh/helm-v2.16.6-linux-amd64.tar.gz
    Preparing to install helm and tiller into /usr/local/bin
    helm installed into /usr/local/bin/helm
    tiller installed into /usr/local/bin/tiller
    main: line 178: which: command not found
    Run 'helm init' to configure helm.

The 'which' command is optional, and not always installed on all
environments (like a Fedora container). Instead, use 'command -v' to
detect if the executable is in the $PATH.

### DIFF
--- a/scripts/get
+++ b/scripts/get
@@ -175,7 +175,7 @@ fail_trap() {
 # testVersion tests the installed client to make sure it is working.
 testVersion() {
   set +e
-  HELM="$(which $PROJECT_NAME)"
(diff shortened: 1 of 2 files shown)


technosophos (approved):
LGTM

PR #8077, opened 2020-05-08
feature(cmd): Add --output option to `helm dep list`

### LINKED ISSUE #8064: helm options  doesn't all have the --output flag
the option to output to json is not available for helm dep

$ helm dep list
NAME VERSION REPOSITORY STATUS
tomcat 6.3.6 https://charts.bitnami.com/bitnami ok

$ helm dep list --output json
Error: unknown flag: --output

is it possible to add the --output option to all commands ?

yonatankahana:
This PR resolves issue #8064. 
some helm commands, like `helm repo list` support `--output` flag that can change the output format to table, json or yaml. There is even package and interface just for that (`output.Writer`). The `helm dep list` doesn't follow this convention. I changed the `helm dep list` command implementation to implement `output.Writer` interface and by that allowing to change the format with the `--output` flag. The tests extended to check every output option, and of course for backward comparability when `--output` not specified.

**If applicable**:

close #8064


pkg/action/dependency.go
```diff
@@ -43,25 +44,66 @@ func NewDependency() *Dependency {
...
+	Version    string `json:"version"`
+	Repository string `json:"repository"`
+	Status     string `json:"status"`
+}
+
+type DependencyListWriter struct {
+	Chartpath string
+}
+
+func (w *DependencyListWriter) WriteTable(out io.Writer) error {
```
mattfarina, line 58: If you look at the existing places that write to JSON (e.g., history, repo listing, etc) it is in the `cmd/helm` package. It's part of the output of the CLI. If there isn't a reason to break from that pattern, can this be refactored to use the existing pattern.

mattfarina (changes requested):
Thanks for creating this PR. I look forward to having this. One structural issue I noticed.

pkg/action/dependency.go
yonatankahana: As I wrote in the PR description, I tried to keep making as many minor changes as possible to ease the code review process. I saw the need for these changes myself, but I was afraid it would cause a lot of changes to it which would increase the size of the PR. 
I made the changes you requested, changed the tests to fit and added some informative comments where it was missing.

yonatankahana:
the commit 44a2225035d93ab6e8dc40af7c3ee4ff13452fae that merged 3 days ago caused conflicts to my changes. i pushed new code after merge and sorting out the conflicts.

yonatankahana:
i pushed another commit that merges latest `master` and fix the conflict from the change https://github.com/helm/helm/commit/d6708667da5c79ab4a093ea7473e96b93be5f1f7#diff-652b5654e858ccc21ad086dc6fe01c09R55-R57.

PR #8082, opened 2020-05-08
fix: init `Storage` by the Log function used by `Configuration`, concrete `Driver`, etc.

### LINKED ISSUE #8108: fix: `Storage`should be init with Log function having concrete logic rather than a empty body
https://github.com/helm/helm/blob/bf12ae39344aef012927ab391eb26ddd3a30ae30/pkg/storage/storage.go#L254-L263

`Storage` is init with Log function with empty body that do nothing, and the caused result is that helm will never record log messages of  `Storage` functions, we will lost these helped messages.

Therefore, `Storage` should be init with a concrete log function to record there messages.

liuming-dev:
commit message:

1. init `Storage` by the Log function used by `Configuration`, concrete `Driver`, etc.
2. fix some log output format for log style consistency

close #8108 

### DIFF
--- a/pkg/storage/storage.go
+++ b/pkg/storage/storage.go
@@ -252,12 +252,17 @@ func makeKey(rlsname string, version int) string {
 // Init initializes a new storage backend with the driver d.
 // If d is nil, the default in-memory driver is used.
 func Init(d driver.Driver) *Storage {
+	return NewStorage(d, func(_ string, _ ...interface{}) {})
+}
+
+// NewStorage creates a new Storage object with the given Driver and log function.
+func NewStorage(d driver.Driver, debug func(format string, args ...interface{})) *Storage {
 	// default driver is in memory
 	if d == nil {
 		d = driver.NewMemory()
 	}
 	return &Storage{
 		Driver: d,
-		Log:    func(_ string, _ ...interface{}) {},
+		Log:    debug,
 	}
 }
--- a/cmd/helm/install.go
+++ b/cmd/helm/install.go
@@ -186,7 +186,7 @@ func runInstall(args []string, client *action.Install, valueOpts *values.Options
 		return nil, err
 	}
 
-	debug("CHART PATH: %s\n", cp)
+	debug("CHART PATH: %s", cp)
 
 	p := getter.All(settings)
 	vals, err := valueOpts.MergeValues(p)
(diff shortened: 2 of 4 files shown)


pkg/storage/storage.go
```diff
@@ -251,13 +251,13 @@ func makeKey(rlsname string, version int) string {
 
 // Init initializes a new storage backend with the driver d.
 // If d is nil, the default in-memory driver is used.
-func Init(d driver.Driver) *Storage {
+func Init(d driver.Driver, log func(format string, args ...interface{})) *Storage {
```
bacongobbler, line 254: As a note, this would be considered a breaking change to the function signature. As per our [semver contract](https://github.com/helm/helm/blob/master/CONTRIBUTING.md#semantic-versioning), we have committed to not making any changes to existing function signatures until Helm 4 as much as possible, though we have made exceptions in the past to fix regressions or security flaws. Given this is a small quality-of-life change I'm not sure this warrants breaking backwards compatibility.

Have you considered alternative solutions to address this issue?

bacongobbler (changes requested):

cmd/helm/helm_test.go
```diff
@@ -93,7 +93,7 @@ func runTestActionCmd(t *testing.T, tests []cmdTestCase) {
 }
 
 func storageFixture() *storage.Storage {
-	return storage.Init(driver.NewMemory())
+	return storage.Init(driver.NewMemory(), debug)
```
jdolitsky, line 96: Should not break the public API

jdolitsky (changes requested):

pkg/storage/storage.go
liuming-dev: Yep, this change might go to be a break change. I have change this to add a `NewStorage` function, this will be  backwards compatibility.

cmd/helm/helm_test.go
liuming-dev: Yes, I have change this, please review.

liuming-dev:
ping

bacongobbler (changes requested):
passing in a log function to the constructor is not the right design choice for this problem. Instead, the store can set the `Log` field after calling `Init`, as this is a public field. That can be done without having to introduce any code to the storage driver interface.

Because of that, I'm closing this as we do not intend to move forward with this design.

PR #8111, opened 2020-05-13
Support fish completion

### LINKED ISSUE #8109: Completion for shells supported by cobra
I took a quick look at the code handling completion in the helm command, and there seems to be a slight issue with renamed binaries, where the generated completion would not have a proper name.

Nowadays fish completion is derivated from the bash one, and there is a bash snippet in case the binary has been renamed from the default.

A Better way of handling it might be to pass the binary name as the name passed when creating the Root cobra command, and so it will propagate everywhere else.

Feel free to comment on this, and i'll try to make a PR to implement this

kri5:

This PR delegates support for completion entirely to cobra.

This allows to support more shells without the need to maintain specific code on helm side.

closes #8109

### DIFF
--- a/cmd/helm/completion.go
+++ b/cmd/helm/completion.go
@@ -26,7 +26,7 @@ import (
 )
 
 const completionDesc = `
-Generate autocompletions script for Helm for the specified shell (bash or zsh).
+Generate autocompletions script for Helm for the specified shell (bash, zsh or fish).
 
 This command can generate shell autocompletions. e.g.
 
@@ -41,6 +41,7 @@ var (
 	completionShells = map[string]func(out io.Writer, cmd *cobra.Command) error{
 		"bash": runCompletionBash,
 		"zsh":  runCompletionZsh,
+		"fish": runCompletionFish,
 	}
 )
 
@@ -52,7 +53,7 @@ func newCompletionCmd(out io.Writer) *cobra.Command {
 
 	cmd := &cobra.Command{
 		Use:   "completion SHELL",
-		Short: "generate autocompletions script for the specified shell (bash or zsh)",
+		Short: "generate autocompletions script for the specified shell (bash, zsh or fish)",
 		Long:  completionDesc,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return runCompletion(out, cmd, args)
@@ -243,4 +244,8 @@ __helm_bash_source <(__helm_convert_bash_to_zsh)
 `
 	out.Write([]byte(zshTail))
 	return nil
+}
+
+func runCompletionFish(out io.Writer, cmd *cobra.Command) error {
+	return cmd.Root().GenFishCompletion(out, true)
 }
--- a/cmd/helm/completion.go
+++ b/cmd/helm/completion.go
@@ -43,6 +43,7 @@ var (
 		"zsh":  runCompletionZsh,
 		"fish": runCompletionFish,
 	}
+	completionNoDesc bool
 )
 
 func newCompletionCmd(out io.Writer) *cobra.Command {
@@ -60,6 +61,7 @@ func newCompletionCmd(out io.Writer) *cobra.Command {
 		},
 		ValidArgs: shells,
 	}
+	cmd.Flags().BoolVar(&completionNoDesc, "no-descriptions", false, "disable completion descriptions for shells that support it")
 
 	return cmd
 }
@@ -247,5 +249,5 @@ __helm_bash_source <(__helm_convert_bash_to_zsh)
 }
 
 func runCompletionFish(out io.Writer, cmd *cobra.Command) error {
-	return cmd.Root().GenFishCompletion(out, true)
+	return cmd.Root().GenFishCompletion(out, !completionNoDesc)
 }


kri5:
Sorry for the noise on this PR.

I first tried to change the name of the binary directly when the command is created, but it leads to tests not passing because the name is also used for usage, and tests hardcode "helm" binary name.

So i went for a method were i override "just-in-time" the name of the command, in order to generate completion with cobra only and get rid of any "completion" code specific to a shell in helm.

Then i added the support for the shells cobra already supports.

Feel free to comment

cmd/helm/completion.go
```diff
@@ -74,173 +74,23 @@ func runCompletion(out io.Writer, cmd *cobra.Command, args []string) error {
...
-			# Use printf instead of echo beause it is possible that
-			# the value to print is -n, which would be interpreted
-			# as a flag to echo
-			printf "%s\n" "${w}"
-		fi
-	done
-}
-__helm_compopt() {
-	true # don't do anything. Not supported by bashcompinit in zsh
+	return cmd.GenZshCompletion(out)
```
marckhouzam, line 91: Cobra 1.0's support for Zsh is a lesser solution than what helm currently provides.
The biggest issue is that it does not support custom completions, so
`helm status <TAB>` would no longer work, for example.

We first need https://github.com/spf13/cobra/pull/1070 to be merged and released before we can start using Cobra's Zsh completion.

cmd/helm/completion.go
```diff
@@ -74,173 +74,23 @@ func runCompletion(out io.Writer, cmd *cobra.Command, args []string) error {
...
 
-	zshTail := `
-BASH_COMPLETION_EOF
-}
-__helm_bash_source <(__helm_convert_bash_to_zsh)
-`
-	out.Write([]byte(zshTail))
-	return nil
+func runCompletionFish(out io.Writer, cmd *cobra.Command) error {
+	return cmd.GenFishCompletion(out, true)
```
marckhouzam, line 95: What do you think of making the second parameter configurable to helm's users?
I suggest adding a `--no-descriptions` flag to the `completion` command.
You can get inspiration from the work started here: https://github.com/helm/helm/pull/7690

cmd/helm/completion.go
```diff
@@ -52,7 +52,7 @@ func newCompletionCmd(out io.Writer) *cobra.Command {
 
 	cmd := &cobra.Command{
 		Use:   "completion SHELL",
-		Short: "generate autocompletions script for the specified shell (bash or zsh)",
+		Short: "generate autocompletions script for the specified shell (bash, zsh or fish)",
```
marckhouzam, line 55: I think the long description also needs to include 'fish'

cmd/helm/completion.go
```diff
@@ -74,173 +74,23 @@ func runCompletion(out io.Writer, cmd *cobra.Command, args []string) error {
 	if !found {
 		return errors.Errorf("unsupported shell type %q", args[0])
 	}
+	// In case the user renamed the helm binary (e.g., to be able to run
+	// both helm2 and helm3), we overwrite the new binary name to the command
+	// in order to generate proper completion
+	rootCmd := cmd.Root()
+	rootCmd.Use = filepath.Base(os.Args[0])
```
marckhouzam, line 81: I'm a little concerned about unknown side-effects of doing this.
But it may be ok.  I'm just leaving a note to keep thinking about it.

marckhouzam (changes requested):
I'm really excited about this contribution!  It is great to see others getting involved with Helm completion code.

Note that for the Fish completion code to also work with Cobra 1.0, we need PR #7906 to get merged.

cmd/helm/completion.go
kri5: Agreed

cmd/helm/completion.go
kri5: Will try to do that, i was not aware of the work you previously done on this.

cmd/helm/completion.go
kri5: Agreed, tottaly missed that one

cmd/helm/completion.go
kri5: We might want to create a deep copy of the command?
I haven't investigated much as if the command was still used afterwards.

cmd/helm/completion.go
kri5: 76c63b819325c72bfb4f6f7460ff52a83bfb9c10 should fix it

cmd/helm/completion.go
marckhouzam: Simpler than a deep-copy could be to
- set Use field
- generate completion script
- reset Use field to old value.

But all this feels hacky to me.

cmd/helm/completion.go
marckhouzam: No problem.  Contributions are always welcomed and help spread the knowledge.

marckhouzam:
I just remembered that as it stands, custom completions (`helm status <TAB>`) will not work for Fish.  We have to move all our custom completion code to Cobra 1.0.0.

I have a patch ready for that but it first needs #7906 to get merged.

cmd/helm/completion.go
kri5: Seems hacky to me too, but i think that's a job for cobra, it definitly should be configurable.
Meanwhile we might want something like that.

I changed the way it was done in order to save/restore the Use field in fe31b1db37f1f1ee7452234a852aa286e2aea581

cmd/helm/completion.go
kri5: should be fixed in c6849c4014c3524bca4b2dd6698594767fa8c386

marckhouzam:
> I just remembered that as it stands, custom completions (`helm status <TAB>`) will not work for Fish. We have to move all our custom completion code to Cobra 1.0.0.
> 
> I have a patch ready for that but it first needs #7906 to get merged.

#7906 was merged 🎉 and I have posted the follow-up PR #8131 which is required to allow Fish completions to have custom completions.

marckhouzam:
@kri5, if you want to rework this PR to only focus on Fish completion (no changes to Bash or Zsh), we could bring it in right away, even before #8131. That would at least provide fish completion for commands and flags.

marckhouzam:
Hi @kri5 .  PR #8131 which moved all dynamic completions to Cobra's new framework has been merged.  This means we can get full Fish completion support very easily.

Will you have time to adapt your PR to limit it to Fish completion?  If not, I can take it over if it helps.  Thanks.

kri5:
Sorry for the late response, there should be only fish completion related changes.
Feel free to comment :)

marckhouzam (changes requested):

cmd/helm/completion.go
```diff
@@ -244,3 +247,7 @@ __helm_bash_source <(__helm_convert_bash_to_zsh)
 	out.Write([]byte(zshTail))
 	return nil
 }
+
+func runCompletionFish(out io.Writer, cmd *cobra.Command) error {
+	return cmd.GenFishCompletion(out, !completionNoDesc)
```
marckhouzam, line 252: Actually, this does not work.  You need to call `cmd.Root().GenFishCompletion(...)`.

marckhouzam (changes requested):
Almost there.

cmd/helm/completion.go
```diff
@@ -52,13 +54,14 @@ func newCompletionCmd(out io.Writer) *cobra.Command {
...
 		Use:   "completion SHELL",
-		Short: "generate autocompletions script for the specified shell (bash or zsh)",
+		Short: "generate autocompletions script for the specified shell (bash, zsh or fish)",
 		Long:  completionDesc,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return runCompletion(out, cmd, args)
 		},
 		ValidArgs: shells,
 	}
+	cmd.Flags().BoolVar(&completionNoDesc, "no-descriptions", false, "disable completion descriptions for shells that support it")
```
marckhouzam, line 64: It was bothering me that this flag would apply to all shells instead of just the fish shell.
I've opened #8314 to allow to improve on this.  Let's see what the other maintainers think of this.

kri5: Agreed! :+1:

marckhouzam:
Because we're waiting for #8134 let's target the 3.4 release for this improvement.

marckhouzam:
@kri5 could you rebase this PR to adapt to #8314 which has just been merged? You will then be able to add the `--no-descrptions` flag to the `fish` command only.

marckhouzam:
Hi @kri5. Just checking to see if you will have some time to work on the fish completion rebase.  There's no real urgency since it's targeted for the 3.4 release but I just wanted to ask.  Thanks.

marckhouzam:
Hi @kri5. With 3.3 released, work has begun for helm 3.4.  Will you be able to rebase this PR so we can provide fish completion?

If you are unable to, I can take it over if you prefer?

marckhouzam:
I have pushed a new PR, #8664, to replace this one, to make sure it makes it on time for the 3.4 release, with some soak time.

Thank you @kri5 for getting the ball rolling on this change.  Any future contribution is welcomed.

PR #8131, opened 2020-05-15
Move go custom completions to Cobra 1.0

marckhouzam:

Cobra 1.0 introduces custom Go completions.  This commit replaces Helm's own solution to use Cobra's solution. This allows to completely remove Helm's internal "completion" package and reduce our maintenance burden.

It is also a requirement to benefit from other Cobra completion improvements like Fish completions (#8111), an upcoming native zsh completion, etc.

Cobra's solution is actually a port of Helm's solution so it is extremely similar.  The main difference is that Cobra's solution is slightly more elegant because it introduces a new field to the Cobra command structure: `ValidArgsFunction` which replaces helm's `completion.RegisterValidArgsFunc()`.

This PR basically does the following:
1. remove the `internal/completion/complete.go` file (the `completion` package)
1. move each call to `completion.RegisterValidArgsFunc()` to the new `ValidArgsFunction` field
1. replace `completion.RegisterFlagCompletionFunc()` with `cmd.RegisterFlagCompletionFunc()`
1. replace constants from helm's `completion` package, with the corresponding one from the `cobra` package 

No functionality is changed and no new functionality is introduced.
This was mostly a `chore` commit.

**If applicable**:
- [x] this PR has been tested for backwards compatibility using the acceptance-testing repo


marckhouzam:
The force push was a rebase to master.

If anyone has time to have a look, a quick way to feel warm and fuzzy about this change is to run the completion tests of the acceptance-testing repo.  Which can be done from the helm repo with: 

`make test-acceptance-completion`

marckhouzam:
Force-push done to resolve conflicts with latest master.

cmd/helm/flags.go
```diff
@@ -55,19 +55,22 @@ func addChartPathOptionsFlags(f *pflag.FlagSet, c *action.ChartPathOptions) {
...
 			if strings.HasPrefix(format, toComplete) {
 				formatNames = append(formatNames, format)
 			}
 		}
-		return formatNames, completion.BashCompDirectiveDefault
+		return formatNames, cobra.ShellCompDirectiveDefault
 	})
+
+	if err != nil {
+		log.Fatal(err)
```
adamreese, line 72: There are a number of `log.Fatal` calls added to the completion functions.  Could this cause Helm to exit when completion is not critical?  Meaning should we print an error and continue?

go.sum
```diff
@@ -725,6 +725,7 @@ gopkg.in/yaml.v2 v2.2.8 h1:obN1ZagJSUGI0Ek/LBmuj4SNLPfIny3KsKFopxRdj10=
 gopkg.in/yaml.v2 v2.2.8/go.mod h1:hI93XBmqTisBFMUTm0b8Fm+jr3Dg1NNxqwp+5A1VGuI=
 gotest.tools v2.2.0+incompatible h1:VsBPFP1AI068pPrMxtb/S8Zkgf9xEmTLJjfM+P5UIEo=
 gotest.tools v2.2.0+incompatible/go.mod h1:DsYFclhRJ6vuDpmuTbkuFWG+y2sxOXAzmJt81HFBacw=
+helm.sh/helm v2.16.7+incompatible h1:3Hp8GLmr6uuBNYGmTQ4p3J4N6xKqxvFDswxHAOposlM=
```
adamreese, line 728: helm-v3 should not import helm-v2

adamreese (changes requested):

go.sum
marckhouzam: Sorry about that.  I don't know how that got there.  But I should have noticed it before pushing it.  I have removed it in the last push and also rebased to master.

cmd/helm/flags.go
marckhouzam: I wasn't sure if it followed helm's correct practices when I added the `log.Fatal()` calls.

You'll notice that all of them are on the initialization code path, and not on the completion generation path.  Such a `log.Fatal()` call would be triggered when receiving a Cobra error when setting up completion during initialization, and would be caused by a coding error.  So I thought it was ok to abort helm since this would be seen immediately during development.

But I'll follow whichever best practice you recommend.

adamreese (approved):

marckhouzam:
Thanks @adamreese!

PR #8142, opened 2020-05-15
feat: Detect missing selector during lint

### LINKED ISSUE #1990: Add lint rule for spec.matchLabels on Deployment
Resources like Deployment, ReplicaSet, and DaemonSet should, as a best practice, declare `selector.matchLabels` rules in their pod template:

https://kubernetes.io/docs/concepts/tools/kubectl/object-management-using-declarative-config/#defining-controller-selectors-and-podtemplate-labels

Doing so prevents a number of odd-ball cases where pods are not appropriately updated.

We should write a linting rule to catch this case.

technosophos:

This issue closes #1990 by adding a quick test for `matchLabels` or `matchExpressions` in specific Kubernetes kinds. This was a requested feature because so many Helm issues were being filed because of malformed charts. I don't know if this is any longer necessary. If not, we can probably just close-won't-fix #1990.

Because a chart with no valid selector is _illegal_, this is an error. It catches `matchLabels` and `matchSelectors` (instead of just `selector`) because the requirement is that one of these two most be present. These are the fields frequently left off by chart authors.

Like all other chart lints, this does not enforce anything other than that the strings are present.

This change will definitely cause some charts that did not pass before to fail a lint now. However, those charts would have been broken had they been installed.

**If applicable**:

Closes #1990

### DIFF
--- a/pkg/lint/rules/template.go
+++ b/pkg/lint/rules/template.go
@@ -131,6 +131,7 @@ func Templates(linter *support.Linter, values map[string]interface{}, namespace 
 			linter.RunLinterRule(support.ErrorSev, path, validateYamlContent(err))
 			linter.RunLinterRule(support.ErrorSev, path, validateMetadataName(&yamlStruct))
 			linter.RunLinterRule(support.ErrorSev, path, validateNoDeprecations(&yamlStruct))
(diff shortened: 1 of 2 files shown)


bacongobbler (approved):

bacongobbler:
this is a good change. LGTM.

Given that we are planning to release 3.3.0-rc.1 soon, I'll go ahead and merge this one in. Thanks @technosophos!

PR #8151, opened 2020-05-16
omit "no matches for kind" errors in uninstall as resources that has …

### LINKED ISSUE #8413: Inconsistent state after uninstallation if a custom resource definition is not available anymore
Given a chart that installs or works with custom resources can leave an inconsistent state after deletion.

Example chart: stable/prometheus 11.6.1

To reproduce:
```
helm install prometheus-operator stable/prometheus-operator
kubectl delete crd prometheuses.monitoring.coreos.com
helm delete prometheus-operator
```

Then the output of the final command is:
```
Error: uninstallation completed with 1 error(s): unable to build kubernetes objects for delete: unable to recognize "": no matches for kind "Prometheus" in version "monitoring.coreos.com/v1"
```

The problem is that none of the actual resources in the release have been removed, although the release itself is gone.

Running the last command once again:
```
helm delete prometheus-operator
Error: uninstall: Release not loaded: prometheus-operator: release: not found
```

Although resources are still available:
```
kubectl get po
NAME                                                      READY   STATUS    RESTARTS   AGE
alertmanager-prometheus-operator-alertmanager-0           2/2     Running   0          105s
prometheus-operator-grafana-77d5b9b8b8-sxwf6              2/2     Running   0          118s
prometheus-operator-kube-state-metrics-69fcc8d48c-5mjkb   1/1     Running   0          118s
prometheus-operator-operator-6ff4d9f5c-5fjgw              2/2     Running   0          118s
prometheus-operator-prometheus-node-exporter-mbg88        1/1     Running   0          118s
```

I am opinionated about this and would suggest that Helm should not fail, but remove resources that does not have a matching kind available, since it could then conclude the resource is already gone. The following PR is about to fix that:
https://github.com/helm/helm/pull/8151

With helm2 I beleive this worked differently in that resources had been built and removed one-by-one instead in two seperate phases. 

If there is an issue in the [first phase](https://github.com/helm/helm/blob/3c9099290b75ee2dd74f1751dc5b454b7b60af7e/pkg/action/uninstall.go#L199) the [second phase](https://github.com/helm/helm/blob/3c9099290b75ee2dd74f1751dc5b454b7b60af7e/pkg/action/uninstall.go#L204) will never get executed, but the release will still be [removed](https://github.com/helm/helm/blob/3c9099290b75ee2dd74f1751dc5b454b7b60af7e/pkg/action/uninstall.go#L130) regardless of the errors.

I'm not sure how you would want to address this, but I guess it's nontrivial if you consider existing clients may already build/depend on this behaviour. However I would consider [omitting "no matches for kind"](https://github.com/helm/helm/pull/8151) errors much lower risk.

Output of `helm version`:
```
```

Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.3", GitCommit:"2e7996e3e2712684bc73f0dec0200d64eec7fe40", GitTreeState:"clean", BuildDate:"2020-05-21T14:51:23Z", GoVersion:"go1.14.3", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.2", GitCommit:"52c56ce7a8272c798dbc29846288d7cd9fbae032", GitTreeState:"clean", BuildDate:"2020-04-30T20:19:45Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
```

```
kind v0.8.1 go1.14.2 darwin/amd64
```

pepov:
Check for "no matches for kind" type of errors during kubernetes resource deserialization and skip them to successfully remove a release even with resources that had no matching kinds. These resources should already be garbage collected, so they can be considered deleted.

With helm v2 this seemed to behave differently as manifests had been deserialized and removed one-by-one instead in a single batch so even if errors accumulated all the other resources have been removed. 

Fixes #8413

### DIFF
--- a/pkg/action/uninstall.go
+++ b/pkg/action/uninstall.go
@@ -21,6 +21,7 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+	k8serrors "k8s.io/apimachinery/pkg/util/errors"
 
 	"helm.sh/helm/v3/pkg/release"
 	"helm.sh/helm/v3/pkg/releaseutil"
@@ -198,7 +199,12 @@ func (u *Uninstall) deleteRelease(rel *release.Release) (string, []error) {
 
 	resources, err := u.cfg.KubeClient.Build(strings.NewReader(builder.String()), false)
 	if err != nil {
-		return "", []error{errors.Wrap(err, "unable to build kubernetes objects for delete")}
+		remainingErr := k8serrors.FilterOut(err, func(err error) bool {
+			return strings.Contains(err.Error(), "no matches for kind")
+		})
+		if remainingErr != nil {
+			return "", []error{errors.Wrap(remainingErr, "unable to build kubernetes objects for delete")}
+		}
 	}
 	if len(resources) > 0 {
 		_, errs = u.cfg.KubeClient.Delete(resources)


sagikazarmark:
Is there anything else we can do here?

pkg/action/uninstall.go
```diff
@@ -198,7 +199,12 @@ func (u *Uninstall) deleteRelease(rel *release.Release) (string, []error) {
 
 	resources, err := u.cfg.KubeClient.Build(strings.NewReader(builder.String()), false)
 	if err != nil {
-		return "", []error{errors.Wrap(err, "unable to build kubernetes objects for delete")}
+		remainingErr := k8serrors.FilterOut(err, func(err error) bool {
```
liuming-dev, line 202: https://github.com/helm/helm/blob/25aa9d0e659679a3e3c2cda303b886c1f5148a20/pkg/action/upgrade.go#L249-L260

Maybe we should keep the same implementation as `upgrade` dose.

pepov: I have to disagree. The goal of the upgrade is that resources should be created/upgraded and if that cannot be achieved, the process should fail.

In the uninstall case the goal is that resources should be gone, and if certain resources are already gone - because their API types are gone - then the goal is achieved, and those errors should not result in the process to fail. 

This is why I used the `FilterOut` helper to remove only these types of errors, bot not others.

liuming-dev: Yes, I mean the error returned by `u.cfg.KubeClient.Build` to process.

pepov:
@bacongobbler could you please help us either by telling why this couldn't be merged or what needs to be done to get this in a better shape?

pkg/action/uninstall.go
pepov: @liuming-dev could you please clarify how the patch should be changed exactly?

pepov:
@bacongobbler @liuming-dev anything we can do here to make this happen or understand if it should be done differently?

sagikazarmark:
Is there anything to do with this PR? Would be nice to have this upstream and not having to use a fork.

hickeyma (commented):
@pepov What is this a fix for? Is there an issue raised for detailing the issue?

pepov:
There is no issue raised, I tried to make this self descriptive, but will rephrase and extend with more details if that helps or I can even create an issue if that's blocking the review.

hickeyma:
I think an issue would be helpful

bacongobbler:
^^ bingo. I don't have enough context here to understand what this PR is attempting to fix. That makes it very hard to determine whether we need to address this or not.

bacongobbler:
tests would also be great.

pepov:
@hickeyma created #8413 
@bacongobbler where can I find existing tests covering this code path, that I could extend with this scenario?

pepov:
@hickeyma @bacongobbler can you please take a look at this again? Is the explanation and/or the approach that should be different or you just want a test case for this? If yes, is there an integration test environment that has a running API server I can interact with?

sagikazarmark:
Can someone take a look at this please? As far as I understand the patch itself is acceptable, just needs some tests for which we need a maintainer to comment.

bacongobbler:
Have a look at install_test.go in the same directory for inspiration. You can use that to provide a fake kube client.

Here is where the fake kube client is set up for the install test case: https://github.com/helm/helm/blob/fc9b46067f8f24a90b52eba31e09b31e69011e93/pkg/action/action_test.go#L85

piotrmsc:
I have run into the same issue. @pepov is there any update on this? Your issue got staled

pepov:
No, we used this patch from a fork, but ended up dropping it and removing CRDs explicitly after helm delete in those cases where we needed it. Unfortunately I didn't have the time to dig into and implement the required testing around it.

pkg/action/uninstall.go
bacongobbler: No action necessary here.

bacongobbler:
> Unfortunately I didn't have the time to dig into and implement the required testing around it.

Okay. I'm going to close this PR as it doesn't look like it's going to be in a mergable state any time soon. If someone wishes to take the work @pepov started, please feel free.

PR #8159, opened 2020-05-18
ref(pkg/chartutil): use minimal in-memory fixtures

adamreese:


technosophos (approved):
Yay! The beginning of what I hope will be the best trend of 2020!

PR #8167, opened 2020-05-20
Add --file-only flag to helm dep up to update only file:// dependencies

AndyKilpatrick:
Signed-off-by: Andy Kilpatrick <Andy.Kilpatrick@metaswitch.com>

This change adds a --file-only flag to helm dep up, so that only file:// dependencies are updated.  

Why:  Significant speed up test-fix cycle on helm charts when working from home (common at the moment...) 
on an umbrella chart with both local (file://) and third-party subcharts.  helm dep up is slow in this case as it downloads large subcharts from internet repos each time (with no actual changes)

Change: "helm dep up --file-only <my-chart>" skips these downloads, only updating file:// dependencies.  It prints out any skipped dependencies.  Sample output from `helm dep up mychart  --file-only`

```Hang tight while we grab the latest from your chart repositories...
Saving 12 charts
Updating file:// dependencies only - skip xxx from repo https://xxx.com/helm-virtual
Updating file:// dependencies only - skip elasticsearch from repo https://kubernetes-charts.storage.googleapis.com
Deleting outdated charts
```

I'm unclear if www-helm docs would be automatically updated by this change, or if a separate pull request is required for the docs.
This is my first github / helm change, apologies if I've made some mistakes in process.

**If applicable**:

### DIFF
--- a/cmd/helm/dependency_update.go
+++ b/cmd/helm/dependency_update.go
@@ -62,6 +62,7 @@ func newDependencyUpdateCmd(out io.Writer) *cobra.Command {
 				ChartPath:        chartpath,
 				Keyring:          client.Keyring,
 				SkipUpdate:       client.SkipRefresh,
+        FileOnly:         client.FileOnly,
 				Getters:          getter.All(settings),
 				RepositoryConfig: settings.RepositoryConfig,
 				RepositoryCache:  settings.RepositoryCache,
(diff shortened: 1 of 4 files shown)


technosophos:
This PR needs both tests and an upstream issue that explains the feature.

That said, the reason we don't do this is that it creates a potential case in which SemVer negotiation does not work as expected. Two users may build the same chart at the same time, but if one uses `--file-only` they can get a different, potentially incompatible chart. You can even get a broken build if a --file-only prevents a fetch that would provide versions taht satisfy semver constraints that the local cache does not have.

Anymore, I'm not sure if we care about these cases. We used to try to prevent users from hitting edge cases like this, but so many PRs have been merged ignoring this guideline that I suppose we might as well do it here, too. Just be aware that this actually introduces new surface area for edge-case breakages.

PR #8183, opened 2020-05-22
Using flags instead of persistent flags on status

### LINKED ISSUE #8149: Revision flag missing in help for helm status
With the changes of #7917 , the help text for `helm status` no longer shows the `revision` flag.

My guess is that this is caused because helm status creates its revision flag as a PersistentFlag:
https://github.com/helm/helm/blob/3779d959667095e03308a2e49480748f7911b228/cmd/helm/status.go#L77

However, the kubectl package used to group commands, I'm guessing shows persistent flags in a `helm <subcommand> options`.

I think there is no need to use a persistent flag for `helm status --revision` (a persistent flag is a flag that is inherited by sub-commands).  Note that the `--revision` flags of the different `helm get` commands are not persistent.

Output of `helm version`:
version.BuildInfo{Version:"v3.2+unreleased", GitCommit:"3779d959667095e03308a2e49480748f7911b228", GitTreeState:"clean", GoVersion:"go1.13.7"}

/cc @zgfh

mattfarina:
Persistent flags are available on sub-commands. Status does not
need the persistent nature.

Closes #8149

### DIFF
--- a/cmd/helm/status.go
+++ b/cmd/helm/status.go
@@ -74,7 +74,7 @@ func newStatusCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 		return compListReleases(toComplete, cfg)
 	})
 
-	f := cmd.PersistentFlags()
+	f := cmd.Flags()
 
 	f.IntVar(&client.Version, "revision", 0, "if set, display the status of the named release with revision")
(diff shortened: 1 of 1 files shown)


marckhouzam (approved):
Thanks for doing this.
LGTM

PR #8185, opened 2020-05-22
Adding a counter to the dependencies cache

### LINKED ISSUE #8184: Helm v2 CI failing on caching issue
When `make bootstrap` is run in environments that already have a Glide cache there's now the error:

```
[ERROR]	Update failed for google.golang.org/genproto: The Remote does not match the VCS endpoint
[ERROR]	Failed to install: The Remote does not match the VCS endpoint
```

This is affecting caching in the CI system as well (example [here](https://circleci.com/gh/helm/helm/12220)).

This is because the source for google.golang.org/genproto moved from https://github.com/google/go-genproto to https://github.com/googleapis/go-genproto. It it now hosted on a different GitHub org.

mattfarina:
The cache needed to be cleared due to an issue presented when the
source for one dependency moved from one place to another. To
accomplish this we add a counter. This is similar to the recommendation
in the CircleCI documentation on caching

Closes #8184

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,13 +13,13 @@ jobs:
           version: 18.06.0-ce
       - restore_cache:
           keys:
-            - glide-{{ checksum "glide.yaml" }}-{{ checksum "glide.lock" }}
-            - glide- # used as a fall through if the checksum fails to find exact entry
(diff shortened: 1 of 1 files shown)


hickeyma (approved):
LGTM, thanks for the fix @mattfarina

PR #8189, opened 2020-05-22
ref(tests): localize unit test fixtures to package

adamreese:
Use test fixtures that are in the same package as test.


pkg/action/show_test.go
```diff
@@ -17,55 +17,50 @@ limitations under the License.
...
+	"helm.sh/helm/v3/pkg/chart"
 )
 
 func TestShow(t *testing.T) {
 	client := NewShow(ShowAll)
-
-	output, err := client.Run("../../cmd/helm/testdata/testcharts/alpine")
-	if err != nil {
-		t.Fatal(err)
+	client.chart = &chart.Chart{
```
technosophos, line 27: Ah! Yes! This is much cleaner for testing.

technosophos (approved):
LGTM.

This is exclusively a re-organization of tests. The one non-test API that changed is changed in a non-breaking way to expose a private variable to the testing fixtures. Many unnecessary files are remoed. And the more disk IO we can ditch during testing, the better our tests will perform.

Issue #8081, opened 2020-05-08
proposal: helm get dependencies


survivant:
One problem with helm list  is that it show the main chart that was deployed but we don't know if there are dependencies deployed too.

helm list
`wordpress       default         1               2020-05-06 10:59:24.0373954 -0400 EDT   deployed        wordpress-9.0.3         5.3.2`

kubectl get pods
```
NAME                                            READY   STATUS    RESTARTS   AGE
wordpress-7b7b8d7cf6-52chl                      0/1     Pending   0          45h
wordpress-mariadb-0                             0/1     Pending   0          45h
```

we should be able to get that kind of information

mattfarina:
@survivant Are you asking for a list of the dependent charts or all the resources a chart deployed?

survivant:
all resources deployed for the chart.

like if I deploy this chart : stable/elastic-stack   under name "k"

helm list
`k               default         1               2020-05-08 09:22:08.9069397 -0400 EDT   deployed        elastic-stack-2.0.0     6
`

but I want to know if the was others charts( dependencies that were deployed too)

like

kubectl get pods --show-labels

```
k-elasticsearch-client-7796d6569c-pfqr8         0/1     Running   0          126m   app=elasticsearch,component=client,pod-template-hash=7796d6569c,release=k
k-elasticsearch-client-7796d6569c-shh4j         0/1     Running   0          126m   app=elasticsearch,component=client,pod-template-hash=7796d6569c,release=k
k-elasticsearch-data-0                          0/1     Pending   0          126m   app=elasticsearch,component=data,controller-revision-hash=k-elasticsearch-data-9cdb89448,release=k,role=data,statefulset.kubernetes.io/pod-name=k-elasticsearch-data-0
k-elasticsearch-master-0                        0/1     Pending   0          126m   app=elasticsearch,component=master,controller-revision-hash=k-elasticsearch-master-787985d74f,release=k,role=master,statefulset.kubernetes.io/pod-name=k-elasticsearch-master-0
k-kibana-797f75695d-842xm                       1/1     Running   0          126m   app=kibana,pod-template-hash=797f75695d,release=k
k-logstash-0                                    0/1     Pending   0          126m   app=logstash,controller-revision-hash=k-logstash-85c7d6dcb6,release=k,statefulset.kubernetes.io/pod-name=k-logstash-0
```

Prims47:
Hello :)

I can take this features 😊 .
Can you assign me ? Thanks

[...]

PR #8190, opened 2020-05-22
feat(get-dependencies): Add get dependencies command (#8081)

Prims47:

Add subcommand `dependencies` to show dependencies for a given release.
See #8081 

**If applicable**:


cmd/helm/get_dependencies.go
```diff
@@ -0,0 +1,129 @@
...
+
+var getDependenciesHelp = `
+This command list all dependencies for a given release.
+`
+
+type dependencyElement struct {
+	Name       string `json:"name"`
+	Version    string `json:"version,omitempty"`
+	Repository string `json:"repository"`
+	Enabled    bool   `json:"enabled,omitempty"`
```
bacongobbler, line 40: Can you elaborate why this flag should be shown in the output? the definition of a dependency that was installed is “enabled”, so wouldn’t this always return true? If so, then is there any point relaying this information back to the user?

Prims47: Hi 😄 ,

That's a very good question. I read in the [documentation](https://helm.sh/docs/topics/charts/#the-chartyaml-file) that an addiction can be disabled. I don't know in what context this can be done.

Prims47: @bacongobbler do you want that i delete it ?

Prims47: @bacongobbler do you want that i delete it ?

Prims47: small up ? :)

Prims47: Can anyone tell me if my answer is okay or is there something I'm expected to do?

bacongobbler: Hi @Prims47. I cannot answer that question for you. I am merely pointing out a question on the design. It's up to the author of the pull request (that's you!) to determine if this makes sense from your point of view. It is up to you to hold these opinions, as I noted in the comment:

> Can you elaborate why this flag should be shown in the output?

A good first step would be to use this pull request and test it against a number of charts in the stable repository, and determine if the enabled flag makes sense from a usability standpoint. That way you'd have a better idea whether a user would need this flag or not.

Hope this helps.

side note: I'm spending most of my time on other projects, and I generally don't respond to pings over the weekend. Sorry for the delay on a response.

Prims47: Thanks for the tips. 

Indeed no stable repository uses this notion. So I removed it.

No worries about the deadline. We all have plans on the side, I completely understand. At worst (after several days) I ping on Slack :)

Prims47: @bacongobbler small ping :)

cmd/helm/get_dependencies.go
```diff
@@ -0,0 +1,127 @@
...
+)
+
+var getDependenciesHelp = `
+This command list all dependencies for a given release.
+`
+
+type dependencyElement struct {
+	Name       string `json:"name"`
+	Version    string `json:"version,omitempty"`
+	Repository string `json:"repository"`
```
hickeyma, line 39: `Alias` might be worth adding: https://github.com/helm/helm/blob/master/pkg/chart/dependency.go#L49

hickeyma (changes requested):
Thanks for the PR and your patience. Some comments inline to address.

Issue #3612, opened 2018-03-05
`helm package` should produce bitwise-deterministic tar files


cmaher:
We have a CI process for publishing charts where we try to determine if pushed repository needs to have it's charts built and published. We would like to do this with a bitwise comparison of the generated tgz files. This works for charts without dependencies, but it is inconsistent for charts with dependencies.

For the setup (Using v2.8.1):
```
charts
├── child1
│   └── Chart.yaml
├── child2
│   └── Chart.yaml
├── child3
│   └── Chart.yaml
└── parent
    ├── Chart.yaml
    └── requirements.yaml
```

With charts:
```
# child{1..3}/Chart.yaml
name: child{1..3}
version: 1.0.0

# parent/Chart.yaml
name: parent
version: 1.0.0

# parent/requirements.yaml
dependencies:
- name: child1
  repository: file://../child1
  version: "*"
- name: child2
  repository: file://../child2
  version: "*"
- name: child3
  repository: file://../child3
  version: "*"

```

Running:
```
# working dir: parent
$ helm dep update
$ for i in {1..1000}; do helm package . && sha256sum parent-1.0.0.tgz && rm parent-1.0.0.tgz; done
```
produces several (3, after a significant number of iterations) different shas. The shas also appear to be non-uniformly distributed (e.g. 1 sha appears 70% of the time).

My guess is that https://github.com/kubernetes/helm/blob/master/pkg/chartutil/save.go#L160 is iterating over dependencies in a non-deterministic order, thus producing different tar files.

bacongobbler:
I think this is actually due to gzip. Part of the gzip header has a mod time for whatever is compressed in the file, so by calling `helm package` multiple times, each one creates their own tarball which is then packaged by gzip and given a unique shasum because the modtime differed.

If you `gunzip` the .tgz, is the shasum identical between tarballs?

cmaher:
`gunzip`ing the tars still produces different tarballs. There still appear to be 3 different shasums.

bacongobbler:
Following up on my previous comment, you can even reproduce this yourself at home:

```
><> helm fetch stable/traefik
><> gunzip traefik-1.24.1.tgz 
><> helm fetch stable/traefik
><> shasum -a 256 *
a032f1eba2c54b76d8de38d3763bab5ca5a6383af1156f30caf4595adaec67a0  traefik-1.24.1.tar
[...]

PR #4161, opened 2018-06-02
Fix incorrect timestamp when helm package

mattjmcnaughton:
Fix #4158 

The files contained in the `tar` file from `helm package` did not have
the correct timestamp. This lack of timestamp occurred because we were
not writing any time to the tar header file.

Issue #4158, opened 2018-06-02
wrong timestamp when using 'helm package'


samwelkey:
!-- If you need help or think you have found a bug, please help us with your issue by entering the following information (otherwise you can delete this text): -->

Output of `helm version`: v2.9.2
Output of `kubectl version`: 1.7.6

Cloud Provider/Platform (AKS, GKE, Minikube etc.): 

When using 'helm package my-chart' to pack my chart. The archive seems to be okey. But when I try to unzip that archive, the files in the folder all have the wrong time stamp which is: 1969-12-31 19:00:00.

mattjmcnaughton:
Great catch! You are totally right. I'll work on a pr to fix this :)

PR #8216, opened 2020-05-28
Cache chart layers without timestamp

### LINKED ISSUE #8212: Same chart saved at different times has different digest
After investigating this issue, I have tracked it down to a change that was made in release v.3.0.3.  Here is a demonstration and a way you can reproduce:

```
export HELM_EXPERIMENTAL_OCI=1
curl -s https://get.helm.sh/helm-v3.0.3-linux-amd64.tar.gz | tar xzvf - -C /tmp/
rm -rf "${HOME}/.cache/helm/registry"
/tmp/linux-amd64/helm create /tmp/mychart
/tmp/linux-amd64/helm chart save /tmp/mychart first:latest
sleep 1
/tmp/linux-amd64/helm chart save /tmp/mychart second:latest
```

The output is like so:
```
$ /tmp/linux-amd64/helm chart save /tmp/mychart first:latest
ref:     first:latest
digest:  470a1e836bbef03577c15cf44c80f6d17fd3f9a953beba4789ff4fe081a9e3fa
...

$ /tmp/linux-amd64/helm chart save /tmp/mychart second:latest
ref:     second:latest
digest:  9b0f939702799c192221d8bb1baffbfca0ec006a4144fbae27d828746a804717
...
```

This is a result of the timestamp being included in `pkg/chartutil/save.go`:
```go
func writeToTar(out *tar.Writer, name string, body []byte) error {
	// TODO: Do we need to create dummy parent directory names if none exist?
	h := &tar.Header{
		Name:    name,
		Mode:    0644,
		Size:    int64(len(body)),
		ModTime: time.Now(),   // <-------- this line was added in release 3.0.3
	}
	if err := out.WriteHeader(h); err != nil {
		return err
	}
	_, err := out.Write(body)
	return err
}
```

The timestamp was added as a fix to #4158, so that archives created with `helm package` would bear the correct timestamp.  However, this breaks the correct functioning of `helm chart save`, which should produce identical digests for identical content. I am preparing a PR to resolve this issue, which will leave the functionality for `helm package` in tact while fixing it for `helm chart save`. 

```
helm version
version.BuildInfo{Version:"v3.1.1", GitCommit:"afe70585407b420d0097d07b21c47dc511525ac8", GitTreeState:"clean", GoVersion:"go1.13.8"}
```

```
kubectl version
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.3", GitCommit:"b3cbbae08ec52a7fc73d334838e18d17e8512749", GitTreeState:"clean", BuildDate:"2019-11-13T11:23:11Z", GoVersion:"go1.12.12", Compiler:"gc", Platform:"linux/amd64"}
error: You must be logged in to the server (the server has asked for the client to provide credentials)
```

Cloud provider is irrelevant in this case

pmengelbert:
This commit allows for the creation of a tar.gz archive of a chart
without a timestamp. Doing so ensures that charts with identical content
will have identical digests, which is important for storage in a
registry. Effort was made to ensure the functionality of `helm package`
was not affected.

Additionally, a unit test for the registry cache's StoreReference
function, to help track this bug.

closes #8212

**If applicable**:

### DIFF
--- a/internal/experimental/registry/cache.go
+++ b/internal/experimental/registry/cache.go
@@ -288,7 +288,7 @@ func (cache *Cache) saveChartConfig(ch *chart.Chart) (*ocispec.Descriptor, bool,
 func (cache *Cache) saveChartContentLayer(ch *chart.Chart) (*ocispec.Descriptor, bool, error) {
 	destDir := filepath.Join(cache.rootDir, ".build")
 	os.MkdirAll(destDir, 0755)
-	tmpFile, err := chartutil.Save(ch, destDir, chartutil.WithoutTimestamp())
+	tmpFile, err := chartutil.SaveWithOpts(ch, destDir, chartutil.WithoutTimestamp())
 	defer os.Remove(tmpFile)
 	if err != nil {
 		return nil, false, errors.Wrap(err, "failed to save")
--- a/internal/experimental/registry/cache.go
+++ b/internal/experimental/registry/cache.go
@@ -288,7 +288,7 @@ func (cache *Cache) saveChartConfig(ch *chart.Chart) (*ocispec.Descriptor, bool,
 func (cache *Cache) saveChartContentLayer(ch *chart.Chart) (*ocispec.Descriptor, bool, error) {
 	destDir := filepath.Join(cache.rootDir, ".build")
 	os.MkdirAll(destDir, 0755)
-	tmpFile, err := chartutil.Save(ch, destDir)
+	tmpFile, err := chartutil.Save(ch, destDir, chartutil.WithoutTimestamp())
 	defer os.Remove(tmpFile)
 	if err != nil {
 		return nil, false, errors.Wrap(err, "failed to save")
--- a/pkg/action/chart_save_test.go
+++ b/pkg/action/chart_save_test.go
@@ -19,6 +19,7 @@ package action
 import (
 	"io/ioutil"
 	"testing"
+	"time"
 
 	"helm.sh/helm/v3/internal/experimental/registry"
 )
@@ -66,5 +67,50 @@ func TestChartSave(t *testing.T) {
 	ref.Tag = "0.1.0"
 	if _, err := action.cfg.RegistryClient.LoadChart(ref); err != nil {
 		t.Error(err)
+	}
+}
+
+func TestIdenticalDigests(t *testing.T) {
+	ref1String := "localhost:5000/test1:0.2.0"
+	ref2String := "localhost:5000/test2:0.2.0"
+
+	tmpDir, err := ioutil.TempDir("", "helm-digest-test")
+	if err != nil {
+		t.Error(err)
+	}
+
+	cache, err := registry.NewCache(
+		registry.CacheOptRoot(tmpDir),
+	)
+	if err != nil {
+		t.Error(err)
+	}
+
+	input := buildChart()
+
+	ref1, err := registry.ParseReference(ref1String)
+	if err != nil {
+		t.Error(err)
+	}
+
+	ref2, err := registry.ParseReference(ref2String)
+	if err != nil {
+		t.Error(err)
+	}
+
+	sum1, err := cache.StoreReference(ref1, input)
+	if err != nil {
+		t.Error(err)
+	}
+
+	time.Sleep(1 * time.Second)
+
+	sum2, err := cache.StoreReference(ref2, input)
+	if err != nil {
+		t.Error(err)
+	}
+
+	if a, b := string(sum1.Digest), string(sum2.Digest); a != b {
+		t.Fatalf("digest %s and digest %s do not match", a, b)
 	}
 }
--- a/pkg/chartutil/save.go
+++ b/pkg/chartutil/save.go
@@ -31,6 +31,22 @@ import (
 	"helm.sh/helm/v3/pkg/chart"
 )
 
+type (
+	saveConfig struct {
+		noTimestamp bool
+	}
+
+	// SaveOpt is an option for modifying the behavior of Save
+	SaveOpt func(*saveConfig)
+)
+
+// WithoutTimestamp provides the ability to save tar archives without a timestamp
+func WithoutTimestamp() SaveOpt {
+	return func(cfg *saveConfig) {
+		cfg.noTimestamp = true
+	}
+}
+
 var headerBytes = []byte("+aHR0cHM6Ly95b3V0dS5iZS96OVV6MWljandyTQo=")
 
 // SaveDir saves a chart as files in a directory.
@@ -99,7 +115,12 @@ func SaveDir(c *chart.Chart, dest string) error {
 // will generate /foo/bar-1.0.0.tgz.
 //
 // This returns the absolute path to the chart archive file.
-func Save(c *chart.Chart, outDir string) (string, error) {
+func Save(c *chart.Chart, outDir string, saveOpts ...SaveOpt) (string, error) {
+	config := new(saveConfig)
+	for _, fn := range saveOpts {
+		fn(config)
+	}
+
 	if err := c.Validate(); err != nil {
 		return "", errors.Wrap(err, "chart validation")
 	}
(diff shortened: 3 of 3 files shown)


jdolitsky (approved):

pmengelbert:
@bacongobbler This PR has been updated to reflect the changes you requested in #8249 . Please let me know if anything further is needed here.

pmengelbert:
Once again, updated based on comments made on PR #8249

internal/experimental/registry/cache.go
```diff
@@ -288,7 +288,7 @@ func (cache *Cache) saveChartConfig(ch *chart.Chart) (*ocispec.Descriptor, bool,
 func (cache *Cache) saveChartContentLayer(ch *chart.Chart) (*ocispec.Descriptor, bool, error) {
 	destDir := filepath.Join(cache.rootDir, ".build")
 	os.MkdirAll(destDir, 0755)
-	tmpFile, err := chartutil.Save(ch, destDir)
+	tmpFile, err := chartutil.SaveWithOpts(ch, destDir, chartutil.WithoutTimestamp())
```
bacongobbler, line 291: should this be changed for both `helm package` and `helm dep up` as well? Why should this be a special case just for `helm chart save`?

bacongobbler: https://github.com/helm/helm/blob/65ee1e0bc14907bb59daa4ac938a7ba145d5f2e5/pkg/action/package.go#L92
https://github.com/helm/helm/blob/65ee1e0bc14907bb59daa4ac938a7ba145d5f2e5/pkg/downloader/manager.go#L711

pmengelbert: @bacongobbler I can't speak for `helm dep up`, but as for `helm package`, the timestamp was deliberately introduced in Helm v3.0.3 (specifically PR [#4161](https://github.com/helm/helm/pull/4161) to resolve issue [#4158](https://github.com/helm/helm/issues/4158).

Let me know your thoughts on this -- I did not want to break someone else's fix, but am happy to do as you ask.

bacongobbler: okay, so coming back to this... Why should we make a special case for `helm chart save`? If users expect that their files are timestamped correctly within the package, why should `helm chart save` be any different?

bacongobbler: The reason I ask this is because we have been talking about using OCI as the standard protocol for distributing charts. If #4158 was opened for `helm package` and `helm dep up`, why wouldn't someone from the community open a similar issue against `helm chart save`?

pmengelbert: > The reason I ask this is because we have been talking about using OCI as the standard protocol for distributing charts. The changes we make here are likely going to be used by community members in future Helm releases.
> 
> If #4158 was opened for `helm package` and `helm dep up`, why wouldn't someone from the community open a similar issue against `helm chart save`?

@bacongobbler The OCI specification is such that the digest of each layer (or blob) is determined by its content. The digests of the individual layers "bubble up" to affect the digest of the manifest, as well.  This is used for registry-side deduplication of both blobs and manifests, as well as identification. If two charts contain the exact same content but a different digest (due to the timestamp), it causes OCI deduplication and identification efforts to fail.

bacongobbler: Isn't that an issue if you're rebuilding the chart from source via `helm chart save`? If you're fetching and pushing the same content across multiple registries via `helm chart pull` and `helm chart push`, the digest remains the same.

jdolitsky: The UX of `helm chart save` is designed to mimic that of `docker build`.

Here's an example `Dockerfile` and `index.html`:
```
# Dockerfile
FROM scratch
ADD index.html /index.html
```
```
<!-- index.html -->
<html><h1>Hello World</h1></html>
```

Now we run `docker build`:
```
$ docker build .
Sending build context to Docker daemon  3.072kB
Step 1/2 : FROM scratch
 --->
Step 2/2 : ADD index.html /index.html
 ---> 49c8907c89fa
Successfully built 49c8907c89fa
```

Now wait 1 minute, and run the command again without modifying either file:
```
$ docker build .
Sending build context to Docker daemon  3.072kB
Step 1/2 : FROM scratch
 --->
Step 2/2 : ADD index.html /index.html
 ---> Using cache
 ---> 49c8907c89fa
Successfully built 49c8907c89fa
```

Notice that the resulting image built has the same digest (`49c8907c89fa`).

Now in the context of Helm - why, if we have not made any modification to any files within a chart directory, should new digests be generated dependent on the time which they are saved?

> If you're fetching and pushing the same content across multiple registries via helm chart pull and helm chart push, the digest remains the same

This is correct, however it fails if I decide to republish/mirror these charts into my own private registry. There needs to be some indication that 2 charts across separate registries are indeed identical, and the digest is what we are using to determine that.

There may be a way around this if we decide to decouple `helm chart save` from the tar-ing process (i.e. first run `helm package mychart/` then `helm chart save mychart-1.0.0.tgz`, using the raw .tgz for the layer), but I do not think this is necessarily a user-friendly approach.

I dont know the correct answer, but I do see this PR as fixing a registry-related regression introduced in 3.0.3

bacongobbler: > why, if we have not made any modification to any files within a chart directory, should new digests be generated dependent on the time which they are saved?

I'm not arguing whether the digests are generated based on the time that `helm package` is called. I am arguing whether the digests are generated based on the *time the files have been modified*. There's a slight nuance in behaviour here.

#4158 was opened because the user expected the timestamps of their files to be preserved during a `helm package` similar to `tar -c`.

In regards to `docker build`'s behaviour... That's another matter entirely. They cache the underlying build context. If the files did not change in your local directory since the last build, `docker build` re-uses the same build context.

Taking your same example Dockerfile here, but instead let's add the `--no-cache` flag to demonstrate:

```
><> cat Dockerfile 
# Dockerfile
FROM scratch
ADD index.html /index.html
><> cat index.html 
<!-- index.html -->
<html><h1>Hello World</h1></html>
><> docker build .
Sending build context to Docker daemon  3.072kB
Step 1/2 : FROM scratch
 ---> 
Step 2/2 : ADD index.html /index.html
 ---> Using cache
 ---> 1eb4e4f55090
Successfully built 1eb4e4f55090
><> docker build --no-cache .
Sending build context to Docker daemon  3.072kB
Step 1/2 : FROM scratch
 ---> 
Step 2/2 : ADD index.html /index.html
 ---> 6b92fac2d0a8
Successfully built 6b92fac2d0a8
><>
```

Notice that the resulting digest differed from the first call to `docker build`. This is because `docker build` checks the mtime of the build context to compute the build context, and Docker re-computed the package digest based on that timestamp.

In other words, if you invalidate the cache (the build context which `docker build` runs from) and call `docker build` using the *exact same files*, it will result in a new digest.

In other words, if you run `docker build` on the **exact same files** as someone else on another computer, you will not end up with the exact same digest.

bacongobbler: If you want to have `helm package` produce bitwise-deterministic packages, then I'd suggest attempting to fix https://github.com/helm/helm/issues/3612. Stripping timedata from the underlying files isn't what's causing this issue.

As reported in that issue's first comment, `helm package` will produce the same SHA 70% of the time. The other 30% are due to files being in another order within the tar package.

This was also clearly pointed out earlier in https://github.com/helm/helm/issues/3612#issuecomment-370786781.

bacongobbler (changes requested):
I just tested this with the test case provided in #3612: https://github.com/cmaher/nondeterministic-charts

```
$ export HELM_EXPERIMENTAL_OCI=1
$ ./helm version
version.BuildInfo{Version:"v3.2+unreleased", GitCommit:"d862cd9843663fc9cd621fcbb2ad8cdef90e2b1e", GitTreeState:"clean", GoVersion:"go1.14.4"}
$ cd parent
$ helm dep up
$ for i in {1..1000}; do ../helm chart save . helm-test-8216:$i; done
```

Here's the result.

```
$ helm chart list | awk '{ print $4 }' | sort | uniq
28be032
71c690a
95b90db
DIGEST
```

Please have a look at the assertions being made in #3612. Stripping the timestamp alone will not produce bitwise-deterministic package checksums.

This PR does not solve the problem this PR is attempting to fix. Please try out the suggestions made in #3612 to see if it will result in bitwise-deterministic packages. Based on some of the comments, I believe that fixing it in this way will get you the results you're trying to achieve.

bacongobbler:
removing from the milestone.

bacongobbler:
closing as we're planning to remove the OCI cache, working around the need for this as users will be pushing packages built with `helm package`. If there are further issues to address, please bring those up in #3612. Thanks.

PR #8221, opened 2020-05-28
Report full template result if it is invalid YAML

karolinepauls:
Signed-off-by: Karoline Pauls <code@karolinepauls.com>

This PR contains a simple change that could have saved me cumulative hours of debugging the dreaded "UPGRADE FAILED: YAML parse error on <FILE>.yaml: error converting YAML to JSON: yaml: line <NUMBER>: <YAML ERROR>". It adds full contents of the incorrectly rendered file that is the culprit for the failure.

I am not aware of another way to retrieve the result of template rendering when the resulting file contains a YAML syntax error.

When this error occurs, the only helpful bits of information are the file and the line where it contains the error. The user won't be bothered by the volume of output. It is impossible to know which lines are relevant in a general case, however it may be beneficial to display them with line numbers.

I am not a Go programmer. I've never programmed in Go before.

**If applicable**:

### DIFF
--- a/pkg/releaseutil/manifest_sorter.go
+++ b/pkg/releaseutil/manifest_sorter.go
@@ -143,7 +143,7 @@ func (file *manifestFile) sort(result *result) error {
 
 		var entry SimpleHead
 		if err := yaml.Unmarshal([]byte(m), &entry); err != nil {
-			return errors.Wrapf(err, "YAML parse error on %s", file.path)
+			return errors.Wrapf(err, "Error in template %s:\n### RESULT START\n%s\n### RESULT END\nYAML parse error on %s", file.path, m, file.path)
 		}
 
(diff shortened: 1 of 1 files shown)


karolinepauls:
I noticed somewhere between the version of Helm 3 that I was using and the current one, `helm template` was made render invalid YAMLs.

While it would be really nice to have a the failing file printed out when Helm fails to upgrade or install, my change doesn't play well with the change in helm `template` and I don't think I will be fixing it.

You can close this.

bacongobbler:
Is the YAML available with `helm template -debug`? In general -debug is your friend when it comes to template errors.

If that does not solve your issue, please make sure to provide a test case along with the PR to test and validate this case. Thanks.

bacongobbler:
> You can close this.

Okay.

PR #8225, opened 2020-05-29
fix(doc): generic description for --version/verify

marckhouzam:
The flags added by `addChartPathOptionsFlags()` are used for 
`helm install` and `helm upgrade` 
but also for 
`helm template`, `helm show` and `helm pull`.

Because of the latter three, we should not use the word 'install' in the description of the `--version` and `--verify` flags.


bacongobbler (approved):
Good call. Thanks for clearing this confusion up. 👍

bacongobbler:
Feel free to merge when ready

marckhouzam:
Thank you @bacongobbler !

PR #8226, opened 2020-05-29
chore(helm): Avoid confusion in command usage

marckhouzam:
This is just a suggested tweak for maintainability.

Having both the `showCmd` and the `subCmd` passed to `addShowFlags()` can easily lead to mistakes in using the wrong command (well, I say 'easily' to make myself feel better for having made such a mistake on a change I'm working on 😄 ).  Instead, `addShowFlags()` should only focus on adding the flags to the `subCmd` and not handle the `addCommand()` call, which is independent.

What do you think @hickeyma?


hickeyma (approved):
LGTM, thanks @marckhouzam

PR #8244, opened 2020-06-02
Report what cause finding chart to fail

carlpett:
Currently, if finding a chart fails, the underlying error is discarded. That error can be non-obvious from the message produced (in my case a non-writeable cache directory)

-

**If applicable**:

### DIFF
--- a/pkg/downloader/manager.go
+++ b/pkg/downloader/manager.go
@@ -571,7 +571,7 @@ func (m *Manager) findChartURL(name, version, repoURL string, repos map[string]*
 	if err == nil {
 		return
 	}
-	err = errors.Errorf("chart %s not found in %s", name, repoURL)
+	err = errors.Errorf("chart %s not found in %s: %s", name, repoURL, err)
 	return
 }
 


pkg/downloader/manager.go
```diff
@@ -571,7 +571,7 @@ func (m *Manager) findChartURL(name, version, repoURL string, repos map[string]*
 	if err == nil {
 		return
 	}
-	err = errors.Errorf("chart %s not found in %s", name, repoURL)
+	err = errors.Errorf("chart %s not found in %s: %v", name, repoURL, err)
```
mattfarina, line 574: Instead of `%v` is should be `%s` as we want the error as a string.

mattfarina (changes requested):
Thanks for bringing out the error. One nit on how the string is constructed though...

pkg/downloader/manager.go
carlpett: :+1: Actually I had changed to `errors.Wrapf` but apparently didn't save before committing :facepalm: Any preference there?

bacongobbler: let's keep the change as minimal as possible. We can address the change to a `WrapF` in a separate PR.

carlpett:
@mattfarina @bacongobbler Is it okay now?

bacongobbler:
the code looks good to me. Would you mind writing a unit test to cover a few of the expected error messages? That way we can see what information you're trying to relay back to the user.

See https://github.com/helm/helm/blob/master/pkg/downloader/manager_test.go for more info.

bacongobbler:
ping.

bacongobbler:
ping.

bacongobbler (approved):

PR #8249, opened 2020-06-03
Manifest digest

### LINKED ISSUE #8248: OCI: Use manifest digest vs. tarball digest
For OCI artifacts, the top-level "envelope" of the artifact is the manifest. This is equally as unique as the chart tarball layer. Currently, on "helm chart save", "helm chart list", and others, the digest displayed is that of the chart tarball layer.

Here is some example output from `helm chart save`:
```
$ helm chart save mychart/ localhost:5000/myrepo/mychart:2.7.0
ref:     localhost:5000/myrepo/mychart:2.7.0
digest:  1b251d38cfe948dfc0a5745b7af5ca574ecb61e52aed10b19039db39af6e1617
size:    2.4 KiB
name:    mychart
version: 0.1.0
2.7.0: saved
```

The digest displayed here is `1b251d38cfe948dfc0a5745b7af5ca574ecb61e52aed10b19039db39af6e1617`. This is that of the chart tarball, represented by `layers[0].digest` in the manifest:

```
$ cat ~/Library/Caches/helm/registry/cache/blobs/sha256/31fb454efb3c69fafe53672598006790122269a1b3b458607dbe106aba7059ef | jq
{
  "schemaVersion": 2,
  "config": {
    "mediaType": "application/vnd.cncf.helm.config.v1+json",
    "digest": "sha256:8ec7c0f2f6860037c19b54c3cfbab48d9b4b21b485a93d87b64690fdb68c2111",
    "size": 117
  },
  "layers": [
    {
      "mediaType": "application/tar+gzip",
      "digest": "sha256:1b251d38cfe948dfc0a5745b7af5ca574ecb61e52aed10b19039db39af6e1617",
      "size": 2487
    }
  ]
}
```

Instead, the digest displayed should be `31fb454efb3c69fafe53672598006790122269a1b3b458607dbe106aba7059ef`, the digest of the manifest itself.

This also affects the output of `helm chart list` (note the DIGEST column, displaying first 7 characters of the digest):

```
$ helm chart list
REF                                                     NAME                    VERSION DIGEST  SIZE            CREATED
localhost:5000/myrepo/mychart:2.7.0                     mychart                 2.7.0   84059d7 454 B           27 seconds
localhost:5000/stable/acs-engine-autoscaler:2.2.2       acs-engine-autoscaler   2.2.2   d8d6762 4.3 KiB         2 hours
localhost:5000/stable/aerospike:0.2.1                   aerospike               0.2.1   4aff638 3.7 KiB         2 hours
localhost:5000/stable/airflow:0.13.0                    airflow                 0.13.0  c46cc43 28.1 KiB        2 hours
localhost:5000/stable/anchore-engine:0.10.0             anchore-engine          0.10.0  3f3dcd7 34.3 KiB 
```

From a registry's perspective, manifests are a "first-class citizen" which is used to further determine metadata and any included layers (such as the chart tarball layer).

The proposal here is to change this digest to use the manifest digest instead. This would normally be considered a breaking change, but the OCI feature set is still in experimental mode.

pmengelbert:
Determine chart digest by manifest

Currently, whenever the chart is printed, the digest of the .tar.gz
content layer is printed as the digest. The manifest digest is important
for OCI purposes, particularly in pushing to a registry.

Resolves #8248

**If applicable**:

### DIFF
--- a/internal/experimental/registry/client.go
+++ b/internal/experimental/registry/client.go
@@ -240,7 +240,7 @@ func (c *Client) PrintChartTable() error {
 // printCacheRefSummary prints out chart ref summary
 func (c *Client) printCacheRefSummary(r *CacheRefSummary) {
 	fmt.Fprintf(c.out, "ref:     %s\n", r.Name)
-	fmt.Fprintf(c.out, "digest:  %s\n", r.Digest.Hex())
+	fmt.Fprintf(c.out, "digest:  %s\n", r.Manifest.Digest.Hex())
 	fmt.Fprintf(c.out, "size:    %s\n", byteCountBinary(r.Size))
 	fmt.Fprintf(c.out, "name:    %s\n", r.Chart.Metadata.Name)
 	fmt.Fprintf(c.out, "version: %s\n", r.Chart.Metadata.Version)
@@ -257,7 +257,7 @@ func (c *Client) getChartTableRows() ([][]interface{}, error) {
 		refsMap[r.Name] = map[string]string{
 			"name":    r.Chart.Metadata.Name,
 			"version": r.Chart.Metadata.Version,
-			"digest":  shortDigest(r.Digest.Hex()),
+			"digest":  shortDigest(r.Manifest.Digest.Hex()),
 			"size":    byteCountBinary(r.Size),
 			"created": timeAgo(r.CreatedAt),
 		}


jdolitsky (approved):

pkg/chartutil/save.go
```diff
@@ -99,7 +115,12 @@ func SaveDir(c *chart.Chart, dest string) error {
 // will generate /foo/bar-1.0.0.tgz.
 //
 // This returns the absolute path to the chart archive file.
-func Save(c *chart.Chart, outDir string) (string, error) {
+func Save(c *chart.Chart, outDir string, saveOpts ...SaveOpt) (string, error) {
```
bacongobbler, line 118: this breaks backwards compatibility.

My suggestion here would be to introduce a new function: `SaveWithOpts`.

bacongobbler (changes requested):

pkg/chartutil/save.go
pmengelbert: This has been addressed in the underlying PR (#8216 ). I will rebase this now to reflect those changes

pkg/chartutil/save.go
```diff
@@ -100,8 +128,64 @@ func SaveDir(c *chart.Chart, dest string) error {
 //
 // This returns the absolute path to the chart archive file.
 func Save(c *chart.Chart, outDir string) (string, error) {
+	tgzWriter, err := createTarGzipWriter(c, outDir)
```
bacongobbler, line 131: Can't we replace all of this in `Save` with a call to `SaveWithOpts(c, outDir)`? The code in `Save` and `SaveWithOpts` are nearly identical. We try to prevent adhering to the WET (Write Everything Twice) principle wherever possible.

bacongobbler (changes requested):

pkg/chartutil/save.go
pmengelbert: @bacongobbler Sure thing. I'll revise here shortly

pmengelbert: @bacongobbler this has been resolved in PR #8216 (on which this PR is based). This PR has been rebased to include the changes there

pmengelbert:
@bacongobbler This PR has been rebased onto master, and no longer depends on PR #8216. We are looking into that PR separately, but this should be good to merge.

bacongobbler (approved):
LGTM.

PR #4305, opened 2018-07-04
fix `helm template -x` pathing issues on Windows

bacongobbler:
tiller's rendering engine converts os filepath separators into unix-style filepath separators, so we need to split manifest names with a forward slash.

closes #4304

PR #8270, opened 2020-06-09
fix template command use --show-only flags error in windows environment

ShenXinKang:
Signed-off-by: ShenXinkang <610716076@qq.com>

when I use "helm template” command with "--show-only" flags, command error in windows envrionment.
image
change path style and command success:
image

I run template_test.go and print manifests and f, the output is:
> manifestPath: templates\subdir\serviceaccount.yaml, f: templates/service.yaml

**filepath.Join and user's input path style** cause this problem and I fix it.

Now everything is ok.
image
image

### DIFF
--- a/cmd/helm/template.go
+++ b/cmd/helm/template.go
@@ -102,6 +102,8 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 					var manifestsToRender []string
 					for _, f := range showFiles {
 						missing := true
+						// Use linux-style filepath separators to unify user's input path
+						f = filepath.ToSlash(f)
 						for _, manifestKey := range manifestsKeys {
 							manifest := splitManifests[manifestKey]
 							submatch := manifestNameRegex.FindStringSubmatch(manifest)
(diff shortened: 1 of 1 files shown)


bacongobbler:
looks like https://github.com/helm/helm/pull/4305 didn't make it into the Helm 3 branch. Can you add a comment similar to the one I wrote in #4305 so this behaviour is documented?

ShenXinKang:
OK. I have added the comment to the code.

cmd/helm/template.go
```diff
@@ -100,6 +100,8 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 					var manifestsToRender []string
 					for _, f := range showFiles {
 						missing := true
+						// Use linux-style filepath separators to unify user's input path
+						f = filepath.ToSlash(f)
```
bacongobbler, line 104: why is this necessary? This wasn't part of the original code...

https://github.com/helm/helm/blob/df3332e61acf9fea88376102ae3372bf7e6d23b6/cmd/helm/template.go#L195-L231

The assertion on https://github.com/helm/helm/blob/df3332e61acf9fea88376102ae3372bf7e6d23b6/cmd/helm/template.go#L209-L210 means that there should be no need to convert the filepath separator.

Can you please either add a unit test that demonstrates what this is attempting to fix or remove this code. This wasn't part of the original fix.

bacongobbler (changes requested):

cmd/helm/template.go
ShenXinKang: On windows, file path can use ```templates/service.yaml``` or ```templates\service.yaml```

I test two format file path and the result is:
- helm/master version run on windows:
    - use templates/service.yaml ------------------[ fail ]
    - use templates\service.yaml ------------------[ success ]
- My first fix version run on windows:
    - use templates/service.yaml ------------------[ success ]
    - use templates\service.yaml ------------------[ fail ]
- My second fix version run on windows:
    - use templates/service.yaml ------------------[ success ]
    - use templates\service.yaml ------------------[ success ]

And I try to write a unit test to demonstrate this.It works on Windows well,but error on Linux.

bacongobbler (approved):
LGTM

PR #8272, opened 2020-06-09
feat: Encode chart information into metadata

mrueg:
As a helm user we want to retrieve information about state of
our helm deployments, so that we can have a cluster-level dashboard for it.

Our approach was employing https://github.com/sstarcher/helm-exporter to
expose helm metadata via prometheus. Unfortunately with the chosen secret
storage backend, read access is required to get the information about deployed
chart and application versions.
Kubernetes does not allow to differentiate between different types of secrets
on a namespace, so providing read access to secrets could leak other secret
information (API Tokens, Certificate Keys, Credentials) through such an
exporter.
This change exposes chart and application versions in the metadata, so we
don't have to provide read access to secrets for helm-exporter anymore, but
helm-exporter could simply use a "list" RBAC permission to read the
metadata.

**If applicable**:

### DIFF
--- a/pkg/chart/chart.go
+++ b/pkg/chart/chart.go
@@ -129,6 +129,14 @@ func (ch *Chart) AppVersion() string {
 	return ch.Metadata.AppVersion
 }
 
+// Version returns the chartversion of the chart.
+func (ch *Chart) Version() string {
+	if ch.Metadata == nil {
+		return ""
+	}
+	return ch.Metadata.Version
+}
+
 // CRDs returns a list of File objects in the 'crds/' directory of a Helm chart.
 // Deprecated: use CRDObjects()
 func (ch *Chart) CRDs() []*File {
--- a/pkg/chart/chart_test.go
+++ b/pkg/chart/chart_test.go
@@ -94,6 +94,7 @@ func TestMetadata(t *testing.T) {
 
 	is.Equal("foo.yaml", chrt.Name())
 	is.Equal("1.0.0", chrt.AppVersion())
+	is.Equal("1.0.0", chrt.Version())
 	is.Equal(nil, chrt.Validate())
 }
 
--- a/pkg/storage/driver/mock_test.go
+++ b/pkg/storage/driver/mock_test.go
@@ -31,6 +31,7 @@ import (
 	kblabels "k8s.io/apimachinery/pkg/labels"
 	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 
+	chart "helm.sh/helm/v3/pkg/chart"
 	rspb "helm.sh/helm/v3/pkg/release"
 )
 
@@ -40,6 +41,14 @@ func releaseStub(name string, vers int, namespace string, status rspb.Status) *r
 		Version:   vers,
 		Namespace: namespace,
 		Info:      &rspb.Info{Status: status},
+		Chart: &chart.Chart{
+			Metadata: &chart.Metadata{
+				APIVersion: "v1",
+				Name:       "hello",
+				Version:    "0.1.0",
+				AppVersion: "0.2.0",
+			},
+		},
 	}
 }
 
--- a/pkg/storage/driver/secrets.go
+++ b/pkg/storage/driver/secrets.go
@@ -224,6 +224,8 @@ func newSecretsObject(key string, rls *rspb.Release, lbs labels) (*v1.Secret, er
 	lbs.set("owner", owner)
 	lbs.set("status", rls.Info.Status.String())
 	lbs.set("version", strconv.Itoa(rls.Version))
+	lbs.set("appVersion", rls.Chart.AppVersion())
+	lbs.set("chartVersion", rls.Chart.Version())
 
 	// create and return secret object.
 	// Helm 3 introduced setting the 'Type' field


liuming-dev (commented):
It looks good, and will solve the problem you faced. 

And my personal opinions about this PR:

- I think  most users will not benefit from this feature
- Helm set labels in secret for management purpose
- If we set these special labels, there remains possible labels to be set with the same reason in many cases

bacongobbler:
> I think most users will not benefit from this feature

As a general rule, we try not to discourage others by weighing in on whether their contribution will benefit the larger community. Instead we try to talk about whether a feature makes sense from a design standpoint. In other words, does a feature fit into the use case a package manager would solve? If so, let's talk about it.

As an example, we thought [Rudders](https://github.com/helm/helm/pull/2079) were going to be the next big thing when it was released in Helm 2. We spent WEEKS on the design, and it was a large part of the code base. It was rarely used and eventually removed in Helm 3. `helm serve` is another similar example of a feature we thought others would benefit from.

This seems dangerous from a design standpoint, though. Releases should not expose any chart information in plain sight. If you can determine the chart name and the chart version used, as an attacker you can look up what exploits are available for the app/version. It exposes information an attacker can use to determine an exploit. These are known as [version disclosure attacks](https://www.netsparker.com/web-vulnerability-scanner/vulnerabilities/version-disclosure-apache/). Apache and nginx are well-known targets for this particular exploit.

mrueg:
> This seems dangerous from a design standpoint, though. Releases should not expose any chart information in plain sight. If you can determine the chart name and the chart version used, as an attacker you can look up what exploits are available for the app/version. It exposes information an attacker can use to determine an exploit. These are known as [version disclosure attacks](https://www.netsparker.com/web-vulnerability-scanner/vulnerabilities/version-disclosure-apache/). Apache and nginx are well-known targets for this particular exploit.

First of all, thanks @bacongobbler for your quick and extensive feedback and sharing your concerns about potential version disclosure attacks. I think Kubernetes still provides enough prevention from that to happen.

What would an attacker need to gather the same information from helm?
* On an RBAC-enabled Cluster: "list" permissions to secrets vs. "read" permissions to secrets. It still would require authorized access to gather the information.

What would an attacker know already from "list" permissions? 
* Chart Name, Modification/Creation Date to make guesses if the release has been fixed (e.g. a deployment prior to a fix being published).

What would users do currently to get the same level of information? 
* Provide read permission to secrets or use a different storage provider. Neither of it is ideal, since it would disclose more information (e.g. potentially hardcoded creds in the chart or rendered yaml).

Back to the version disclosure risk: For the helm-exporter, I definitely see that risk, since it would expose the data on an endpoint, that might be unsecured. I see that outside of helm's scope though and eventually helm-exporter needs to be secured in order to only allow authorized access.

pkg/chart/chart.go
```diff
@@ -129,6 +129,14 @@ func (ch *Chart) AppVersion() string {
 	return ch.Metadata.AppVersion
 }
 
+// ChartVersion returns the chartversion of the chart.
+func (ch *Chart) ChartVersion() string {
```
mattfarina, line 133: This should be `Version` instead of `ChartVersion`.

mattfarina (commented):
I'm still thinking about this whole idea.

Kubernetes has a [recommended set of labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/) that should include things like version. For Helm, there is the label `helm.sh/chart`. (Most?) charts have these labels. The ones that are not Helm specific are for all apps. Even those installed a different way.

Would that provide you the information you need? If not, is there an issue with these labels?

pkg/chart/chart.go
mrueg: Thanks, I fixed the name!

mrueg:
@mattfarina are you suggesting to add those labels (i.e. rename the labels added in this PR) to the helm secret? If so, that definitely would solve it.

As of right now, if defined within the chart, the labels would get added to the objects the chart deploys. To access them we would require cluster-level list access to all objects vs. with this approach only to secrets. Some charts only might deploy a specific CRD so it would be really hard to catch all of them.

mattfarina:
@mrueg I was not thinking of adding them to the secret. Your information around access helps to understand this. Thanks.

mattfarina:
We talked about this on the Helm developer call. An angle I had not thought about was the security angle. There is a difference between list and get on secrets for RBAC. Moving this information to labels moves this information to be available on list as well as get (I think).

This left me wondering if there was another way to make this information available. One in which we (the Helm project) could get out of the way and not be a blocker on.

aauren:
I may be missing something, but I'm not sure necessarily how the security angle applies here. Listing secrets is a discrete permission that can be either granted or denied via RBAC. If someone has a fear of version disclosure attacks, then all they have to do is ensure that they don't allow users/roles to list secrets within their cluster.

This PR gives user's options and a gradient with which to disclose chart specifics. In general, I think that it solves a security problem by giving the cluster administrator more options for disclosing chart specifics. If they just want to expose versions of chart deployments they can now do so by granting list instead of granting access to the entire secret via read. If they want to give access to the entire secret then they can do so by granting list and read. And if they don't want to give away any information about helm secrets they can choose to do so by not granting either list or read.

I think that this PR also increases consistency within Helm itself. For the same reason that these labels are valuable on helm managed objects (like deployments, pods, services, etc.), these labels are also valuable on the secrets that helm keeps for itself. It seems that the security argument that is being made here could equally be applied to the labels helm puts on all of the Kubernetes objects that Helm manages. And yet these labels are considered a standard in the way that Helm works within the cluster.

mrueg:
@mattfarina apologies for getting back to so late, I concur with @aauren here on the consistency point, most helm charts add those labels to other objects as part of the chart as well. 
Is this something that could get you convinced to accept this PR?

mrueg:
@mattfarina @bacongobbler still interested in this PR, anything that I could change to convince you to get it in (e.g. a feature-flag or similar)?

tuananhnguyen-ct:
Almost one year later and I'm still interested in this PR too :)

mrueg:
Closing this PR as there seems to chance to integrate it. If anyone else is interested, feel free to reopen a new PR.

PR #8262, opened 2020-06-07
Add Warning when installing or upgrading a deprecated chart

jiefeng-xu:
**What this PR does / why we need it**:
Close #3724 

Add warning when installing or upgrading a deprecated chart, for release-2.16

**Special notes for your reviewer**:

Signed-off-by: Stanton Xu <xjiefeng@gmail.com>

**If applicable**:
- [ ] this PR contains documentation
- [ ] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

PR #8273, opened 2020-06-09
fix: helm template discard warnings

FabianKramm:
Signed-off-by: fabiankramm <kramm@covexo.com>

When executing `helm template` warnings are still printed, which essentially makes the output of `helm template` useless for something like `helm template | kubectl apply -f -`. An example of such a case is `helm template redis redis --repo https://kubernetes-charts.storage.googleapis.com`, which prints `WARNING: This chart is deprecated` alongside the rendered manifests.  

This pull request solves this issue by discarding the output of the underlying install command (errors remain untouched). An alternative solution would be to save the output in a buffer and print it in an error case.

### DIFF
--- a/cmd/helm/template.go
+++ b/cmd/helm/template.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"path/filepath"
 	"regexp"
 	"sort"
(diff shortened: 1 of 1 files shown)


liuming-dev:
It's necessary to show the users  the possible warning messages. And a common different case, it `--debug` flag is set, what the output should be?

bacongobbler:
#8262 hasn’t even been merged yet. I added feedback to that PR to display errors on standard error rather than standard out as per convention. That should allow you to strip the error message out with a pipe.

Are there other error messages preventing you from piling the output to kubectl, or was that the only one?

bacongobbler:
I noticed that you've commented on that PR. Let's keep the conversation going over there. This is not a PR we want to accept. Instead the output should be displayed on standard error as per convention. Thanks.

FabianKramm:
@bacongobbler I see and understand the reasoning behind it. My main issue is the `WARNING: This chart is deprecated` that is printed to os.Stdout in helm v3 (#8262 is targeting helm v2) that prevents us from using `helm template` anymore. I can change this PR or make a new PR that will essentially do the same change as (#8262) and print this warning to os.Stderr

bacongobbler:
Oh! That would be great. Thank you.

liuming-dev:
I agree with @bacongobbler , warnings should be output to standard error instead of discarding the warnings.

PR #8277, opened 2020-06-09
fix(cmd): display warnings on stderr

bacongobbler:
The warnings introduced when a chart has been deprecated is displayed on standard out. This is a regression for users piping the output of `helm template` from a deprecated chart to `kubectl`. This changes the error message to display on standard error instead.

identified in #8262.

### DIFF
--- a/cmd/helm/helm.go
+++ b/cmd/helm/helm.go
@@ -56,6 +56,11 @@ func debug(format string, v ...interface{}) {
 	}
 }
 
+func warning(format string, v ...interface{}) {
+	format = fmt.Sprintf("WARNING: %s\n", format)
+	fmt.Fprintf(os.Stderr, format, v...)
+}
+
 func initKubeLogs() {
 	pflag.CommandLine.SetNormalizeFunc(wordSepNormalizeFunc)
 	gofs := flag.NewFlagSet("klog", flag.ExitOnError)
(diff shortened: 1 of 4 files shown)


liuming-dev (approved):

jdolitsky (changes requested):
I think the test is capturing stdout only, so either need to remove the `[warning]` line from `deprecated-chart.txt` or modify the test to capture both

bacongobbler:
thanks for catching that. Fixed.

hickeyma (commented):
Thanks for the PR @bacongobbler. Got a question about the output. With this PR, the output is as follows:

```helm.go:61: WARNING: This chart is deprecated```

Do we want it in this style with the helm go file details?

File: cmd/helm/helm.go [hunk 1]
```diff
@@ -56,6 +56,11 @@ func debug(format string, v ...interface{}) {
 	}
 }
 
+func warning(format string, v ...interface{}) {
+	format = fmt.Sprintf("WARNING: %s\n", format)
+	log.Printf(format, v...)
```
mattfarina, line 61: To echo what @hickeyma had to say, I don't think the internal details of the Go code are relevant to the user here. Do we need them?

jdolitsky (approved):

cmd/helm/helm.go
bacongobbler: good call. Will fix.

PR #8282, opened 2020-06-10
Make helm ls return only current releases if providing state filter

### LINKED ISSUE #7495: Helm ls --failed returns releases which are "deployed" now, but has a failure in the history .
```
helm ls --namespace=reform-scan
NAME                   	NAMESPACE  	REVISION	UPDATED                                	STATUS  	CHART                         	APP VERSION
reform-scan-blob-router	reform-scan	50      	2020-01-30 10:16:28.851853992 +0000 UTC	deployed	reform-scan-blob-router-0.0.22	1.0
```

```
helm ls --namespace=reform-scan  --failed
NAME                   	NAMESPACE  	REVISION	UPDATED                                	STATUS	CHART                         	APP VERSION
reform-scan-blob-router	reform-scan	44      	2020-01-29 14:19:37.643962942 +0000 UTC	failed	reform-scan-blob-router-0.0.20	1.0
```
I wouldn't expect --failed to return a helm release which is currently deployed. At least, that was how helm 2 was behaving. Is this a change or bug in helm 3?

Output of `helm version`: 

Output of `kubectl version`: 
`Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-15T23:41:55Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.7", GitCommit:"6c143d35bb11d74970e7bc0b6c45b6bfdffc0bd4", GitTreeState:"clean", BuildDate:"2019-12-13T18:46:24Z", GoVersion:"go1.12.12", Compiler:"gc", Platform:"linux/amd64"}`

andrewmelis:

Previously, the `helm ls --$state` operation would display outdated
releases under certain conditions.

Given the following set of releases:
```
NAME  REVISION  UPDATED                   STATUS    CHART      APP VERSION  NAMESPACE
bar   1         Wed Apr  8 16:54:39 2020  DEPLOYED  bar-4.0.0  1.0          default
foo   1         Fri Feb  7 06:16:56 2020  DEPLOYED  foo-0.1.0  1.0          default
foo   2         Mon May  4 07:16:56 2020  FAILED    foo-0.1.0  1.0          default
foo   3         Mon May  4 07:20:00 2020  FAILED    foo-0.1.0  1.0          default
foo   4         Tue May  5 08:16:56 2020  DEPLOYED  foo-0.2.0  1.0          default
qux   1         Tue Jun  9 10:32:00 2020  DEPLOYED  qux-4.0.3  1.0          default
qux   2         Tue Jun  9 10:57:00 2020  FAILED    qux-4.0.3  1.0          default
```

`helm ls --failed` produced the following output:
```
NAME  REVISION  UPDATED                   STATUS    CHART      APP VERSION  NAMESPACE
foo   3         Mon May  4 07:20:00 2020  FAILED    foo-0.1.0  1.0          default
qux   2         Tue Jun  9 10:57:00 2020  FAILED    qux-4.0.0  1.0          default
```

Including the `qux` release in that `helm ls --failed` output is not
controversial; the most recent revision of `qux` was not successful
and an operator should investigate.

Including the `foo` release in the output, however, is
questionable. Revision 3 of `foo` is _not_ the most recent release of
`foo`, and that FAILED release was fixed in a susubsequent upgrade. A
user may see that FAILED deploy and start taking inappropriate
action. Further, that issue was fixed months ago in this example --
troubleshooting an old deploy may not be safe if significant changes
have occurred. Concern over this behavior was raised in
https://github.com/helm/helm/issues/7495.

This behavior applied to all the state filter flags (--deployed,
--failed, --pending, etc.), and a user could pass multiple state
filter flags to a single command. The previous behavior can be
summarized as follows:

For each release name, all release revisions having any of the
supplied state flags were retrieved, and the most recent revision
among these was returned (regardless of whether a newer revision of an
unspecified state exists).

This change request alters the helm list action to match user
expectations such that only "current" releases are shown when
filtering on release state. After this change, the following output
would be produced by `helm ls --failed`:

```
NAME  REVISION  UPDATED                   STATUS    CHART      APP VERSION  NAMESPACE
qux   2         Tue Jun  9 10:57:00 2020  FAILED    qux-4.0.0  1.0          default
```
The command now returns only `qux` because it is the only "current" FAILED release.

This behavior change applies to all the state filters _except_
`superseded`, which now becomes a special case. By definition, at
least one newer release exists ahead of each superseded release. A
conditional is included in this change request to maintain the
preexisting behavior (return "most recent" superseded revison for
each release name) if the superseded state filter is requested.

---

Note that there is an alternate perspective that a state filter flag
should return all releases of a given state rather than only the
"current" releases. In the above example, `helm ls --failed` with this
alternate approach would return the following:
```
NAME  REVISION  UPDATED                   STATUS    CHART      APP VERSION  NAMESPACE
foo   2         Mon May  4 07:16:56 2020  FAILED    foo-0.1.0  1.0          default
foo   3         Mon May  4 07:20:00 2020  FAILED    foo-0.1.0  1.0          default
qux   2         Tue Jun  9 10:57:00 2020  FAILED    qux-4.0.0  1.0          default
```
Multiple FAILED `foo` revisions are included in the output, unlike the current behavior.

This approach is logical and achievable. It allows a user to find
exactly what is requested: all historical releases of a given
state. In order to achieve continuity with helm behavior, however, a
new filter (something like "current") would probably need to be
implemented and made the new default.

Given current helm behavior as well as the comments in the #7495, I
did not pursue this approach.

---

Technical details:

 - Moved list action state mask filter after latest release filter

Previously, the list operation in helm/pkg/action/list.go skipped
releases that were not covered by the state mask on _retrieval_ from
the Releases store:

```
results, err := l.cfg.Releases.List(func(rel *release.Release) bool {
		// Skip anything that the mask doesn't cover
		currentStatus := l.StateMask.FromName(rel.Info.Status.String())
		if l.StateMask&currentStatus == 0 {
			return false
		}
                ...
```
https://github.com/helm/helm/blob/8ea6b970ecd02365a230420692350057d48278e5/pkg/action/list.go#L154-L159

While filtering on retrieval in this manner avoided an extra iteration
through the entire list to check on the supplied condition later, it
introduced the possibility of returning an outdated release to the
user because newer releases (that would have otherwise squashed
outdated releases in the `filterList` function) are simply not
included in the set of working records.

This change moves the state mask filtering process to _after_ the set
of current releases is built. Outdated, potentially misleading
releases are scrubbed out prior to the application of the state mask
filter.

As written, this state mask filtration (in the new `filterStateMask`
method on `*List`) incurs an additional, potentially expensive
iteration over the set of releases to return to the user. An
alternative approach could avoid that extra iteration and fit this
logic into the existing `filterList` function at the cost of making
`filterList` function a little harder to understand.

 - Rename filterList to filterLatestReleases for clarity

Another function that filters the list is added, so update
to the more descriptive name here.

 - List superseded releases without filtering for latest

This change makes superseded releases a special case, as they would
_never_ be displayed otherwise (by definition, as superseded releases have been
replaced by a newer release), so a conditional maintains current
behavior ("return newest superseded revision for each release name")

Fixes #7495.

Signed-off-by: Andrew Melis <andrewmelis@gmail.com>

Fixes #7495.

**If applicable**:

### DIFF
--- a/pkg/action/list.go
+++ b/pkg/action/list.go
@@ -250,7 +250,8 @@ func (l *List) filterStateMask(releases []*release.Release) []*release.Release {
 
 	for _, rls := range releases {
 		currentStatus := l.StateMask.FromName(rls.Info.Status.String())
-		if l.StateMask&currentStatus == 0 {
+		mask := l.StateMask & currentStatus
+		if mask == 0 {
 			continue
 		}
 		desiredStateReleases = append(desiredStateReleases, rls)
--- a/pkg/action/list.go
+++ b/pkg/action/list.go
@@ -152,12 +152,6 @@ func (l *List) Run() ([]*release.Release, error) {
 	}
 
 	results, err := l.cfg.Releases.List(func(rel *release.Release) bool {
-		// Skip anything that the mask doesn't cover
-		currentStatus := l.StateMask.FromName(rel.Info.Status.String())
-		if l.StateMask&currentStatus == 0 {
-			return false
-		}
-
 		// Skip anything that doesn't match the filter.
 		if filter != nil && !filter.MatchString(rel.Name) {
 			return false
@@ -173,7 +167,16 @@ func (l *List) Run() ([]*release.Release, error) {
 		return results, nil
 	}
 
-	results = filterList(results)
+	// by definition, superseded releases are never shown if
+	// only the latest releases are returned. so if requested statemask
+	// is _only_ ListSuperseded, skip the latest release filter
+	if l.StateMask != ListSuperseded {
+		results = filterLatestReleases(results)
+	}
+
+	// State mask application must occur after filtering to
+	// latest releases, otherwise outdated entries can be returned
+	results = l.filterStateMask(results)
 
 	// Unfortunately, we have to sort before truncating, which can incur substantial overhead
 	l.sort(results)
@@ -222,8 +225,8 @@ func (l *List) sort(rels []*release.Release) {
 	}
 }
 
-// filterList returns a list scrubbed of old releases.
-func filterList(releases []*release.Release) []*release.Release {
+// filterLatestReleases returns a list scrubbed of old releases.
+func filterLatestReleases(releases []*release.Release) []*release.Release {
 	latestReleases := make(map[string]*release.Release)
 
 	for _, rls := range releases {
@@ -240,6 +243,20 @@ func filterList(releases []*release.Release) []*release.Release {
 		list = append(list, rls)
 	}
 	return list
+}
+
+func (l *List) filterStateMask(releases []*release.Release) []*release.Release {
+	desiredStateReleases := make([]*release.Release, 0)
+
+	for _, rls := range releases {
+		currentStatus := l.StateMask.FromName(rls.Info.Status.String())
+		if l.StateMask&currentStatus == 0 {
+			continue
+		}
+		desiredStateReleases = append(desiredStateReleases, rls)
+	}
+
+	return desiredStateReleases
 }
 
 // SetStateMask calculates the state mask based on parameters.
--- a/pkg/action/list_test.go
+++ b/pkg/action/list_test.go
@@ -188,6 +188,56 @@ func TestList_StateMask(t *testing.T) {
 	is.Len(res, 3)
 }
 
+func TestList_StateMaskWithStaleRevisions(t *testing.T) {
+	is := assert.New(t)
+	lister := newListFixture(t)
+	lister.StateMask = ListFailed
+
+	makeMeSomeReleasesWithStaleFailure(lister.cfg.Releases, t)
+
+	res, err := lister.Run()
+
+	is.NoError(err)
+	is.Len(res, 1)
+
+	// "dirty" release should _not_ be present as most recent
+	// release is deployed despite failed release in past
+	is.Equal("failed", res[0].Name)
+}
+
+func makeMeSomeReleasesWithStaleFailure(store *storage.Storage, t *testing.T) {
+	t.Helper()
+	one := namedReleaseStub("clean", release.StatusDeployed)
+	one.Namespace = "default"
+	one.Version = 1
+
+	two := namedReleaseStub("dirty", release.StatusDeployed)
+	two.Namespace = "default"
+	two.Version = 1
+
+	three := namedReleaseStub("dirty", release.StatusFailed)
+	three.Namespace = "default"
+	three.Version = 2
+
+	four := namedReleaseStub("dirty", release.StatusDeployed)
+	four.Namespace = "default"
+	four.Version = 3
+
+	five := namedReleaseStub("failed", release.StatusFailed)
+	five.Namespace = "default"
+	five.Version = 1
+
+	for _, rel := range []*release.Release{one, two, three, four, five} {
+		if err := store.Create(rel); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	all, err := store.ListReleases()
+	assert.NoError(t, err)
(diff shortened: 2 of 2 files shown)


andrewmelis:
Is there anything I can do to help move this forward? I'm a first-time contributor and would be happy to provide any additional inputs if I haven't followed the proper protocol.

It looks like a second issue (https://github.com/helm/helm/issues/8319) was filed for this behavior, so I'd love to solve some real-world pain with this patch.

eddycharly:
Any chance to get a release soon with the `helm ls` fix ?

bacongobbler:
We're swamped with current work for the 3.3.0 milestone, so we haven't had a chance to review this quite yet. We will likely be able to have a look at this for 3.4.0. Thanks for your patience.

andrewmelis:
Appreciate the heads up @bacongobbler 👍

pkg/action/list.go
```diff
@@ -242,6 +245,20 @@ func filterList(releases []*release.Release) []*release.Release {
 	return list
 }
 
+func (l *List) filterStateMask(releases []*release.Release) []*release.Release {
+	desiredStateReleases := make([]*release.Release, 0)
+
+	for _, rls := range releases {
+		currentStatus := l.StateMask.FromName(rls.Info.Status.String())
+		if l.StateMask&currentStatus == 0 {
```
jdolitsky, line 253: Can we put a space between statemask, "&", and currentStatus for readability?

jdolitsky (changes requested):

pkg/action/list.go
andrewmelis: Hi @jdolitsky -- thanks for the review.

It looks like gofmt wants to remove the spaces when the `&` operator is used in this if statement ([here's](https://github.com/helm/helm/blob/70f89e5f26015a92432bdeb94ddb6d6e52532b19/pkg/action/list_test.go#L51) a similar formatting elsewhere in the codebase, presumably from gofmt). I'm reluctant to fight gofmt here, but I'll defer to your judgement.

It looks like gofmt will keep the spaces around `&` when used in an assignment, however.
```
// https://play.golang.org/p/XmQYHSKQT5B
package main

import (
	"fmt"
)

func main() {
	a := 100
	b := 1

	if a&b == 0 {
		fmt.Println("foo")
	}

	fmt.Printf("%b\n", a&b)

	out := a & b
	fmt.Printf("%b\n", out)
}
```


We could break the condition out into a variable assignment on the line above to improve readability, would you prefer that?
```suggestion
		shouldInclude := l.StateMask & currentStatus
		if shouldInclude {
```

andrewmelis: my proposed change above isn't quite right -- should have taken my time.

gofmt removes the spaces around `&` if the assignment includes the equality operator
```
shouldNotInclude := l.StateMask&currentStatus == 0
```
in order to maintain the space around `&`, it would need to be something like
```

mask := l.StateMask & currentStatus
if mask == 0 {
...
```

jdolitsky: I agree, lets not fight with gofmt :)

I personally think the `mask := l.StateMask & currentStatus` is more readable. The change overall look fine to me

andrewmelis: addressed in f9d3524 @jdolitsky 👍

jdolitsky (approved):

hickeyma (approved):
LGTM, thanks for the fix @andrewmelis

PR #8292, opened 2020-06-11
fix: unknown kind resources install/uninstall order

### LINKED ISSUE #8291: install/uninstall order for unknown kind resources not in InstallOrder/UninstallOrder
Output of `helm version`: version.BuildInfo{Version:"v3.2.3", GitCommit:"8f832046e258e2cb800894579b1b3b50c2d83492", GitTreeState:"clean", GoVersion:"go1.13.12"}

Output of `kubectl version`: Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.3", GitCommit:"2e7996e3e2712684bc73f0dec0200d64eec7fe40", GitTreeState:"clean", BuildDate:"2020-05-21T14:51:23Z", GoVersion:"go1.14.3", Compiler:"gc", Platform:"darwin/amd64"}

**Description**

We have two CRDs `NginxVhost` and `Registration`, and the usage limits that install nginxvhost first then registration, registration first then nginxvhost when uninstall.

Helm's logic about install/uninstall CRs is goging as the alphabetical order, and the caused result is that nginxvhost is uninstalled first, that's not we expect.

For such case, it's possible that helm dose not order the CRs by the kind with alphabetical order, we just respect the raw manifest order when install and reversed order when uninstall.

For Example:

two CRs in manifest:
```
Kind: NginxVhost
...
---
Kind: Registration
....
```

it will install NginxVhost then Registration, and uninstall Registration then NginxVhost.

 I will have a try to do this.

liuming-dev:
close #8291

The order rule of install/uninstall unknown kind resources faces some problems, see #8291, and the PR is going to fix it.

The modified function `lessByKind` is also used by `sortHooksByKind`, I'm not sure this PR is suitable for hooks install/uninstall.

**If applicable**:

### DIFF
--- a/pkg/releaseutil/kind_sorter.go
+++ b/pkg/releaseutil/kind_sorter.go
@@ -17,6 +17,7 @@ limitations under the License.
 package releaseutil
 
 import (
+	"reflect"
 	"sort"
 
 	"helm.sh/helm/v3/pkg/release"
@@ -137,12 +138,11 @@ func lessByKind(a interface{}, b interface{}, kindA string, kindB string, o Kind
 	first, aok := ordering[kindA]
 	second, bok := ordering[kindB]
 
+	// both kinds are unknown
+	// the installing order of the unknown kinds is as the order that they are in original manifests/hooks,
+	// and the uninstalling order is the reverse order that they are in original manifests/hooks.
 	if !aok && !bok {
-		// if both are unknown then sort alphabetically by kind, keep original order if same kind
-		if kindA != kindB {
-			return kindA < kindB
-		}
-		return first < second
+		return reflect.DeepEqual(o, UninstallOrder)
 	}
 	// unknown kind is last
 	if !aok {
--- a/pkg/releaseutil/kind_sorter_test.go
+++ b/pkg/releaseutil/kind_sorter_test.go
@@ -18,13 +18,14 @@ package releaseutil
 
 import (
 	"bytes"
+	"sort"
 	"testing"
 
 	"helm.sh/helm/v3/pkg/release"
 )
 
-func TestKindSorter(t *testing.T) {
-	manifests := []Manifest{
+func buildManifestsForTestKindSorter() []Manifest {
+	return []Manifest{
 		{
 			Name: "E",
 			Head: &SimpleHead{Kind: "SecretList"},
@@ -165,15 +166,31 @@ func TestKindSorter(t *testing.T) {
 			Name: "x",
 			Head: &SimpleHead{Kind: "HorizontalPodAutoscaler"},
 		},
+		{
+			Name: "N",
+			Head: &SimpleHead{Kind: "NginxVhost"},
+		},
+		{
+			Name: "U",
+			Head: &SimpleHead{Kind: "Unknown"},
+		},
+		{
+			Name: "R",
+			Head: &SimpleHead{Kind: "Registration"},
+		},
 	}
+}
+
+func TestKindSorter(t *testing.T) {
+	manifests := buildManifestsForTestKindSorter()
 
 	for _, test := range []struct {
 		description string
 		order       KindSortOrder
 		expected    string
 	}{
-		{"install", InstallOrder, "aAbcC3deEf1gh2iIjJkKlLmnopqrxstuvw!"},
-		{"uninstall", UninstallOrder, "wvmutsxrqponLlKkJjIi2hg1fEed3CcbAa!"},
+		{"install", InstallOrder, "aAbcC3deEf1gh2iIjJkKlLmnopqrxstuvw!NUR"},
+		{"uninstall", UninstallOrder, "wvmutsxrqponLlKkJjIi2hg1fEed3CcbAaRUN!"},
 	} {
 		var buf bytes.Buffer
 		t.Run(test.description, func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


mattfarina (changes requested):
Can you please write or update a test that fails without the change and passes with the change to show the issue is fixed.

liuming-dev:
Test cases has been provided, please review, thanks @mattfarina .

liuming-dev:
Any update or feedback on this PR? @mattfarina

pkg/releaseutil/kind_sorter_test.go
```diff
@@ -329,3 +330,144 @@ func TestKindSorterForHooks(t *testing.T) {
 		})
 	}
 }
+
+func TestKindSorterFailedForCRsUsingAlphabeticalOrder(t *testing.T) {
```
mattfarina, line 334: Is there a reason that `TestKindSorter` wasn't updated? It already has [one custom king](https://github.com/helm/helm/blob/561cc4280830af4a40f2e60c5f17fb3ee69b4e39/pkg/releaseutil/kind_sorter_test.go#L68-L71) in it. I would like to see and have tests of the sorting in the broader context rather than solely focused on CRs. This is a touchy area.

liuming-dev: Test cases have been updated, please review, thanks.

pkg/releaseutil/kind_sorter.go
```diff
@@ -138,9 +139,10 @@ func lessByKind(a interface{}, b interface{}, kindA string, kindB string, o Kind
 	second, bok := ordering[kindB]
 
 	if !aok && !bok {
-		// if both are unknown then sort alphabetically by kind, keep original order if same kind
-		if kindA != kindB {
-			return kindA < kindB
+		// if install, keep CRs installing order as the order of CRs in the raw manifests
+		// if uninstall, keep CRs uninstalling order as the reversed order of CRs in the raw manifests
```
mattfarina, line 143: The ordering doesn't just apply to CRs. There are numerous resources Helm isn't explicitly tracking (e.g., Endpoint) that are treated the same. Can the comment read that Kinds Helm is not tracking, that are not typically used in charts, are sorted alpha or reverse alpha after the ones helm is tracking.

pkg/releaseutil/kind_sorter.go
```diff
@@ -138,9 +139,10 @@ func lessByKind(a interface{}, b interface{}, kindA string, kindB string, o Kind
...
+		if reflect.DeepEqual(o, UninstallOrder) {
+			return true
```
mattfarina, line 145: This change doesn't work exactly as expected. In the tests change the order of NginxVhost and Registration and they will fail. The sorting isn't really working when both kinds are unknown. You'll need to check if it's installing or uninstalling and do the comparison appropriately in each case.

mattfarina (changes requested):
This PR doesn't currently work as expected. More details in the comments.

pkg/releaseutil/kind_sorter.go
liuming-dev: Yes, the ordering actually applies to all unknown kinds that are not listed in `InstallOrder` or `UninstallOrder`, and I have updated the comment.

Now, the whole order rules should be:

1. install or uninstall known kinds listed in `InstallOrder` or `UninstallOrder` firstly, then unknown kinds
2. install or uninstall known kinds as the order in `InstallOrder` or `UninstallOrder`
3. install unknown kinds as the order in original manifests/hooks, uninstall unknown kinds as the *reversed* order in original manifests/hooks

pkg/releaseutil/kind_sorter.go
liuming-dev: I have updated this PR, please review, thanks.

liuming-dev:
ping

liuming-dev:
Could this PR be accepted? Or could someone give me some suggestions to fix this issue?

hickeyma (commented):
@liuming-dev Is this solution to fix the same use case as the solution in #8448?

liuming-dev:
> @liuming-dev Is this solution to fix the same use case as the solution in #8448?

Yep, I think this PR works for that case.

liuming-dev:
Ask again, could this PR be accepted or any feedback? :smiley:

liuming-dev:
ping

liuming-dev:
ping +1

liuming-dev:
Hi Helm core maintainers,

I'm trying to attract your attentions on this PR.

Here I give my reason trying to convince you to accept this PR:

**Known kind resource**: list in `InstallOrder ` `UninstallOrder`
**Unknown kind resource**: not list in `InstallOrder ` `UninstallOrder`

Helm current default behavior of installing or uninstalling unknown kind resources is that sorting the unknown kind resources alphabetically by kind, this logic supposes that there is no dependence relationship among these unknown kind resources. It seems that we don't know how to process the installing or uninstalling order of unknown kind resources, so we chose a simple and rough way to do this.

However, as both Helm and CRD are more and more widely used, more and more custom resources will appear in release manifests, the cases of dependence relationships among CRs will happen,  the Helm default behavior for these cases seems unsuitable any more.

A common and simple practice is installing the resource dependent on by other resources first, and uninstall the resource if no resources depend on it first. This is like Helm present behavior for known kind resources(`UninstallOrder` is the reverse order of  `InstallOrder`). 

**Uninstalling order is the reversed order of installing order.**

Giving a example:

**B** depend on **A**

- installing order: **A** --> **B**
- uninstalling order: **B** --> **A**

This PR dose like this.  The installing order is as the unknown kind resources in the release raw manifests, and the uninstalling order is the reversed order of installing order.

And in my background, this PR will not break the backend compatibility.

Please help to review this PR and give me some feedback, thanks. @mattfarina @hickeyma @technosophos @bacongobbler

technosophos:
@liuming-dev We have a lot to do. It currently takes a long time to review PRs. Please be patient and stop pinging us.

OrangeFlag:
ping +1

delanni:
All discussions on the topic seem to have dried up, neither this nor the other (https://github.com/helm/helm/pull/8448) ordering PR doesn't seem to move forward. Did you all end up finding another way to work around the custom resource creation ordering, or just forked off to have your own copies of Helm? :)

liuming-dev:
> All discussions on the topic seem to have dried up, neither this nor the other (https://github.com/helm/helm/pull/8448) ordering PR doesn't seem to move forward. Did you all end up finding another way to work around the custom resource creation ordering, or just forked off to have your own copies of Helm? :)

The PR has been merged in a copy of Helm, and has been released since june, last year.

PR #1461, opened 2016-10-26
Add deployment, service and NOTES to create

viglesiasce:
This brings the `helm create` command templates up to date with what we expect to see in the Charts repo.

---

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/helm/1461)

PR #8306, opened 2020-06-12
fix(chartutil): do not set helpers.tpl filetype for vim

adamreese:
Helpers filetype is not mustache


bacongobbler (approved):
Wow... This has been here since v2.0.0-rc.1. https://github.com/helm/helm/pull/1461

Nice catch.

PR #8314, opened 2020-06-15
feat(cmd): Subcommands for the completion command

### LINKED ISSUE #8313: completion command applies flags to every shell
The helm `completion` command supports the `bash` shell and `zsh` shell.  With #8111, it will also support the fish shell.  The fish shell introduces a *description* for a completion.  For example:
```
$ helm s<TAB>
search  (search for a keyword in charts) 
show  (show information of a chart) 
status  (display the status of the named release)
``` 
We want to allow the user to disable descriptions, in case they don't like having them.  So #8111 introduces the `--no-descriptions` to the `completion` command.  However, this flag does not make sense at this time for neither `bash` nor `zsh`, but the flag still applies to them.  It is not a big deal, but we could do better.

marckhouzam:

Fixes #8313.

This PR converts the `bash` and `zsh` arguments of the `completion` command to sub-commands.

Making each shell a subcommand of the `completion` command has multiple advantages:
- simplifies the code,
- allows to have different flags for each shell,
  for example, a future `--no-descriptions` flag for `fish` only,
- allows to tailor the help text per shell,
- allows Cobra to handle error cases (specifying an invalid shell).

This should be done before #8111 is released to users, or else it would technically become a breaking change.  This is because #8111 will add a new `--no-descriptions` flag to the `completion` command, and without this PR, we cannot restrict it to the `fish` shell.

The PR also improves on the help text for each shell.

**If applicable**:
- [x] this PR has been tested for backwards compatibility using the completion tests of the acceptance-testing repo

### DIFF
--- a/cmd/helm/completion.go
+++ b/cmd/helm/completion.go
@@ -21,61 +21,70 @@ import (
 	"os"
 	"path/filepath"
 
-	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+
+	"helm.sh/helm/v3/cmd/helm/require"
 )
 
 const completionDesc = `
-Generate autocompletions script for Helm for the specified shell (bash or zsh).
+Generate autocompletions script for Helm for the specified shell.
+`
+const bashCompDesc = `
+Generate the autocompletion script for Helm for the bash shell.
 
-This command can generate shell autocompletions. e.g.
+To load completions in your current shell session:
+$ source <(helm completion bash)
 
-    $ helm completion bash
-
-Can be sourced as such
-
-    $ source <(helm completion bash)
+To load completions for every new session, execute once:
+Linux:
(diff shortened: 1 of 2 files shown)


marckhouzam:
For the sake of transparency, my plan is that the proposed `--no-descriptions` flag eventually be applicable to all shells.  But that may take a while.  So I like that this PR allows to control flags on a per-shell basis.

marckhouzam:
If release 3.3 still has some room (time-wise), I wouldn't mind a review on this small change to unblock #8111 which brings support for Fish completion.

bacongobbler (approved):
code LGTM. Given that 3.3.0-rc.1 is about to be released, I'll go ahead and merge this one in. Thanks @marckhouzam!

marckhouzam:
Thanks for getting this into 3.3 @bacongobbler!

PR #8330, opened 2020-06-19
Implementation of Install API

devdinu:
* Exposed install, upgrade, list via HTTP
* added basic service HTTP test for install


hickeyma (changes requested):
@devdinu Do you mind creating a proposal issue for this PR, explaining:
- Use case(s) for this functionality
- How you are going to implement the different Helm APIs including timelines etc.

This would allow the maintainers discuss a substantial feature like this before code is pushed.

devdinu:
Sorry for the PR, it was meant to be on our fork.

@hickeyma Our usecase needs deployment via HTTP api calls. we deploy helm charts everyday for users. After helm3 we could have built a service using this library. since, this seems like widely needed feature [helm-web-api](https://github.com/microsoft/helm-web-api) and seen other blogs.

We could create a proposal, and if accepted we can contribute. I'm thinking of working of fork and if accepted will work on merging it to helm. 

Please share your suggestions/thoughts.

hickeyma:
@devdinu I think if you have the project already then that's ok. My question was more related to whether you wanted to push this PR into core.

PR #8331, opened 2020-06-20
add template file size to 3m and throw  clearly error

Cweiping:
It is to solve a problem that the error cannot be understood completely after the template creation fails.
**How to reproduce**：
Use a template package over 3m：
```shell
ll test-0.1.0.tgz
```
```
-rw-r--r--  1 xxx  staff    13M  6 20 22:17 test-0.1.0.tgz
```
```shell
helm install test test-0.1.0.tgz --debug
```
```
install.go:159: [debug] Original chart version: ""
install.go:176: [debug] CHART PATH: /Users/xxx/Desktop/test-0.1.0.tgz

Error: create: failed to create: Request entity too large: limit is 3145728
helm.go:84: [debug] Request entity too large: limit is 3145728
create: failed to create
helm.sh/helm/v3/pkg/storage/driver.(*Secrets).Create
	/home/circleci/helm.sh/helm/pkg/storage/driver/secrets.go:161
helm.sh/helm/v3/pkg/storage.(*Storage).Create
	/home/circleci/helm.sh/helm/pkg/storage/storage.go:66
helm.sh/helm/v3/pkg/action.(*Install).Run
	/home/circleci/helm.sh/helm/pkg/action/install.go:315
main.runInstall
	/home/circleci/helm.sh/helm/cmd/helm/install.go:229
main.newInstallCmd.func1
	/home/circleci/helm.sh/helm/cmd/helm/install.go:117
	/go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:842
	/go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:950
	/go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:887
main.main
	/home/circleci/helm.sh/helm/cmd/helm/helm.go:83
runtime.main
runtime.goexit
```

This is because the api-server [limits the size of the requested body](https://github.com/kubernetes/kubernetes/blob/release-1.18/staging/src/k8s.io/apiserver/pkg/server/config.go#L323) to 3m when creating secrets. Therefore, we can verify the size of the template problem before installation, avoid unnecessary requests, and give user-friendly prompts.


wangkai1994:
good!

cmd/helm/install.go
```diff
@@ -192,6 +193,10 @@ func runInstall(args []string, client *action.Install, valueOpts *values.Options
 		return nil, err
 	}
 
+	if _, err := isLimitSize(chartRequested.Raw); err != nil {
+		return nil, err
+	}
```
mattfarina, line 198: This size issue is not with the requested chart but rather the release object as it is stored. The release object contains some information in addition to the requested chart and the chart is not stored raw. This is checking the wrong object.

cmd/helm/install.go
```diff
@@ -240,6 +245,20 @@ func isChartInstallable(ch *chart.Chart) (bool, error) {
...
+//
+// This is because the server requires the requestbody to be less than 3m when creating secrets.For details, please refer to: https://github.com/kubernetes/kubernetes/blob/release-1.18/staging/src/k8s.io/apiserver/pkg/server/config.go#L323
+func isLimitSize(raw []*chart.File) (bool, error) {
+	totalSize := int64(0)
+	for _, r := range raw {
+		totalSize += int64(bytes.Count(r.Data, []byte("")))
+	}
+	if totalSize < int64(3 * 1024 * 1024) {
+		return true, nil
+	}
```
mattfarina, line 258: 3mb is an issue for secrets and only when using etcd as the backing storage for those. Helm can use a database and that has different size constraints. Hard coding one size may not be the best way to check for this.

mattfarina (changes requested):
I like the idea of providing a better error message here. I think there is a better way to approach this.

First, this should be built into both install and upgrade.

Second, it should happen in the `action` package so that anything using the action will get it.

Third, it should be able to support different storage mechanisms size constraints.

I wonder if writing this secret and catching the failure is a good way to do it.

cmd/helm/install.go
Cweiping: > I like the idea of providing a better error message here. I think there is a better way to approach this.
> 
> First, this should be built into both install and upgrade.
> 
> Second, it should happen in the `action` package so that anything using the action will get it.
> 
> Third, it should be able to support different storage mechanisms size constraints.
> 
> I wonder if writing this secret and catching the failure is a good way to do it.



> I like the idea of providing a better error message here. I think there is a better way to approach this.
> 
> First, this should be built into both install and upgrade.
> 
> Second, it should happen in the `action` package so that anything using the action will get it.
> 
> Third, it should be able to support different storage mechanisms size constraints.
> 
> I wonder if writing this secret and catching the failure is a good way to do it.

Thank you for your patient review. I didn't think about storing more when I submitted the code. I tried to add the error prompt where I think it is appropriate. Please give some suggestions. I have tested all four supported storage formats, and only ConfigMaps and Secrets have this limitation. Should this problem be treated as a feature fix in v3.3 ?

cmd/helm/install.go
Cweiping: Hard coding one size the worst-case scenario
.And I submitted an update and look forward to your suggestion.

bridgetkromhout:
Hi, @Cweiping - this is a problem: "Cweiping wants to merge 2 commits into helm:release-3.2 from Cweiping:add_limit_template_size"

That is the 3.2 release branch. You will want to retarget this PR to helm:master. Here's how: https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/changing-the-base-branch-of-a-pull-request

pkg/storage/driver/driver.go
```diff
@@ -29,6 +29,8 @@ var (
 	ErrReleaseNotFound = errors.New("release: not found")
 	// ErrReleaseExists indicates that a release already exists.
 	ErrReleaseExists = errors.New("release: already exists")
+	// ErrPackageTooLarge indicates that package too large,because of storage limit,reference: https://github.com/kubernetes/kubernetes/blob/release-1.18/staging/src/k8s.io/apiserver/pkg/server/config.go#L323
+	ErrPackageTooLarge = errors.New("Chart package too large,storage limit is 3145728")
```
bacongobbler, line 33: typo. Please add a space after the comma.

Also, the number 3145728 is not very descriptive. Taking this from a user's perspective: is this 3145728 bits? 3145728 bytes? 3145728 kilobytes?

I'd re-word this as "release: storage limit is 3MB" to remain consistent with the other error messages. The issue may not be related to the chart package being too large. You can, in theory, request for a package that's 3145727 bytes in size (just under the 3MB limit) and the rest of the secret metadata required brings it over the limit.

bacongobbler (changes requested):

Cweiping:
0.0. This PR merge is broken. It is related to the new PR: https://github.com/helm/helm/pull/8357

pkg/storage/driver/driver.go
Cweiping: done. move to PR:https://github.com/helm/helm/pull/8357. 
At the same time, I have an [issue](https://github.com/helm/helm/issues/8359) that has nothing to do with the PR.

PR #8340, opened 2020-06-23
version bump

bacongobbler:


mattfarina (approved):
lgtm

The number of transitive dependencies added is amazing. See the go.sum file.

Note, go does dependencies at the module level rather than the package level. So, if there are packages in your modules that have dependencies but you don't use those package... the unused packages dependencies are still listed.

This does not affect the compiled binary... just your local cache in any environment.

PR #8357, opened 2020-06-28
add template file size to 3m and throw clearly error

Cweiping:
It is to solve a problem that the error cannot be understood completely after the template creation fails.
**How to reproduce**：
Use a template package over 3m：
```shell
ll test-0.1.0.tgz
```
```
-rw-r--r--  1 xxx  staff    13M  6 20 22:17 test-0.1.0.tgz
```
```shell
helm install test test-0.1.0.tgz --debug
```
```
install.go:159: [debug] Original chart version: ""
install.go:176: [debug] CHART PATH: /Users/xxx/Desktop/test-0.1.0.tgz

Error: create: failed to create: Request entity too large: limit is 3145728
helm.go:84: [debug] Request entity too large: limit is 3145728
create: failed to create
helm.sh/helm/v3/pkg/storage/driver.(*Secrets).Create
	/home/circleci/helm.sh/helm/pkg/storage/driver/secrets.go:161
helm.sh/helm/v3/pkg/storage.(*Storage).Create
	/home/circleci/helm.sh/helm/pkg/storage/storage.go:66
helm.sh/helm/v3/pkg/action.(*Install).Run
	/home/circleci/helm.sh/helm/pkg/action/install.go:315
main.runInstall
	/home/circleci/helm.sh/helm/cmd/helm/install.go:229
main.newInstallCmd.func1
	/home/circleci/helm.sh/helm/cmd/helm/install.go:117
	/go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:842
	/go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:950
	/go/pkg/mod/github.com/spf13/cobra@v1.0.0/command.go:887
main.main
	/home/circleci/helm.sh/helm/cmd/helm/helm.go:83
runtime.main
runtime.goexit
```

This is because the api-server [limits the size of the requested body](https://github.com/kubernetes/kubernetes/blob/release-1.18/staging/src/k8s.io/apiserver/pkg/server/config.go#L323) to 3m when creating secrets. Therefore, we can verify the size of the template problem before installation, avoid unnecessary requests, and give user-friendly prompts.

### DIFF
--- a/pkg/storage/driver/cfgmaps.go
+++ b/pkg/storage/driver/cfgmaps.go
@@ -166,6 +166,8 @@ func (cfgmaps *ConfigMaps) Create(key string, rls *rspb.Release) error {
 	if _, err := cfgmaps.impl.Create(context.Background(), obj, metav1.CreateOptions{}); err != nil {
 		if apierrors.IsAlreadyExists(err) {
 			return ErrReleaseExists
+		} else if apierrors.IsRequestEntityTooLargeError(err) {
+			return errors.Wrapf(err, "release: storage limit")
 		}
 
 		cfgmaps.Log("create: failed to create: %s", err)
--- a/pkg/storage/driver/cfgmaps_test.go
+++ b/pkg/storage/driver/cfgmaps_test.go
@@ -17,11 +17,11 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"reflect"
+	"strings"
 	"testing"
 
-	v1 "k8s.io/api/core/v1"
-
 	rspb "helm.sh/helm/v3/pkg/release"
+	v1 "k8s.io/api/core/v1"
 )
 
 func TestConfigMapName(t *testing.T) {
@@ -178,6 +178,12 @@ func TestConfigMapCreate(t *testing.T) {
 	if !reflect.DeepEqual(rel, got) {
 		t.Errorf("Expected {%v}, got {%v}", rel, got)
 	}
+
+	//  configmap overhead entity limit
+	if err := cfgmaps.Create(testKey("large-chart", 1), releaseStub("large-chart", 1, "", rspb.StatusDeployed)); !strings.Contains(err.Error(), "storage limit: Request entity too large") {
+		t.Errorf("Expected {%v}, got {%v}", "release: storage limit: Request entity too large: Request entity too large: limit is 3145728", err)
+	}
+
 }
 
 func TestConfigMapUpdate(t *testing.T) {
--- a/pkg/storage/driver/mock_test.go
+++ b/pkg/storage/driver/mock_test.go
@@ -139,6 +139,9 @@ func (mock *MockConfigMapsInterface) Create(_ context.Context, cfgmap *v1.Config
 	if object, ok := mock.objects[name]; ok {
 		return object, apierrors.NewAlreadyExists(v1.Resource("tests"), name)
 	}
+	if name == "large-chart.v1" {
+		return nil, apierrors.NewRequestEntityTooLargeError("Request entity too large: limit is 3145728")
+	}
 	mock.objects[name] = cfgmap
 	return cfgmap, nil
 }
--- a/pkg/storage/driver/secrets.go
+++ b/pkg/storage/driver/secrets.go
@@ -156,6 +156,8 @@ func (secrets *Secrets) Create(key string, rls *rspb.Release) error {
 	if _, err := secrets.impl.Create(context.Background(), obj, metav1.CreateOptions{}); err != nil {
 		if apierrors.IsAlreadyExists(err) {
 			return ErrReleaseExists
+		} else if apierrors.IsRequestEntityTooLargeError(err) {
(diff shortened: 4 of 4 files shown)


bridgetkromhout:
Hi, @Cweiping - for us to review this, we will first need the tests to pass. Can you take a look? Thanks!
```
pkg/storage/driver/cfgmaps.go:169: File is not `gofmt`-ed with `-s` (gofmt)
 
		}else if apierrors.IsRequestEntityTooLargeError(err){
 
pkg/storage/driver/secrets.go:159: File is not `gofmt`-ed with `-s` (gofmt)
 
		}else if apierrors.IsRequestEntityTooLargeError(err){
 ```

**Comment by Cweiping** [comment]
> Hi, @Cweiping - for us to review this, we will first need the tests to pass. Can you take a look? Thanks!
> 
> ```
> pkg/storage/driver/cfgmaps.go:169: File is not `gofmt`-ed with `-s` (gofmt)
>  
> 		}else if apierrors.IsRequestEntityTooLargeError(err){
>  
> pkg/storage/driver/secrets.go:159: File is not `gofmt`-ed with `-s` (gofmt)
>  
> 		}else if apierrors.IsRequestEntityTooLargeError(err){
> ```

Done.

File: pkg/storage/driver/driver.go [hunk 1]
```diff
@@ -29,6 +29,8 @@ var (
 	ErrReleaseNotFound = errors.New("release: not found")
 	// ErrReleaseExists indicates that a release already exists.
 	ErrReleaseExists = errors.New("release: already exists")
+	// ErrReleaseStorageLimit indicates that a release storage limit exceeded, reference: https://github.com/kubernetes/kubernetes/blob/release-1.18/staging/src/k8s.io/apiserver/pkg/server/config.go#L323
+	ErrReleaseStorageLimit = errors.New("release: storage limit is 3MB")
```
mattfarina, line 33: This is an exported error so we won't be changing the content of it during Helm v3. That would be a breaking change for package consumers via SemVer. With that in mind, I do not like hard coding a value for the storage limit. If (or when) that happens we could run into issues.

Also, the storage size is configurable in etcd to some extent. There are docs at https://etcd.io/docs/v3.4.0/dev-guide/limit/

I'm by no means an etcd expert.

Can we have the error respond in a way that doesn't hard code the storage limit in a Helm public variable?

pkg/storage/driver/cfgmaps.go
```diff
@@ -166,6 +166,8 @@ func (cfgmaps *ConfigMaps) Create(key string, rls *rspb.Release) error {
 	if _, err := cfgmaps.impl.Create(context.Background(), obj, metav1.CreateOptions{}); err != nil {
 		if apierrors.IsAlreadyExists(err) {
 			return ErrReleaseExists
+		} else if apierrors.IsRequestEntityTooLargeError(err) {
+			return ErrReleaseStorageLimit
```
mattfarina, line 170: I see how you followed the pattern for returning an error just like the one above. The API is returning an error with the limit amount. Wouldn't it be better to pass on the limit k8s is returning?

mattfarina (changes requested):
Thanks for the pull request. I'm wondering if there is a better way to pass the max value around.

pkg/storage/driver/driver.go
Cweiping: This is a hard code defined by kubernetes: [limits the size of the requested body](https://github.com/kubernetes/kubernetes/blob/release-1.18/staging/src/k8s.io/apiserver/pkg/server/config.go#L323).
Our only dependency is api-server. We can't sense the limit size of ETCD. We should not consider the limit of ETCD. If you don't consider hard coding, you can consider trying to extract the API's error information and parse its limit size.

pkg/storage/driver/cfgmaps.go
Cweiping: Try to wrap kubernetes' error return information directly?
Example:
```
release: storage limit, Request entity too large: limit is 3145728
```

pkg/storage/driver/driver.go
Cweiping: I use `errors.Wrapf(err, "release: storage limit")` to package it, output:
```
Error: release: storage limit: Request entity too large: limit is 3145728
```
The problem of hard coding is solved.There is no explicit prompt to the user, kubernetes does not add units.

Cweiping:
> Thanks for the pull request. I'm wondering if there is a better way to pass the max value around.

Changes have been made according to your suggestions, waiting for your review again.

bacongobbler:
Any chance we can get test coverage for these changes? Possibly by having a test fixture return `apierrors.RequestEntityTooLargeError`. Thanks.

Cweiping:
> Any chance we can get test coverage for these changes? Possibly by having a test fixture return `apierrors.RequestEntityTooLarge`. Thanks.

Oh, my God. I'm waiting for you to review.

Done.
The test case has been completed and will be reviewed again

PR #8363, opened 2020-06-28
helm upgrade with --wait support jobs in manifest to be completed

### LINKED ISSUE #5948: Helm upgrade doesn't wait for Job to complete
Output of `helm version`:
```
Client: &version.Version{SemVer:"v2.14.1", GitCommit:"5270352a09c7e8b6e8c9593002a73535276507c0", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.14.1", GitCommit:"5270352a09c7e8b6e8c9593002a73535276507c0", GitTreeState:"clean"}
```

Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.6", GitCommit:"abdda3f9fefa29172298a2e42f5102e777a8ec25", GitTreeState:"clean", BuildDate:"2019-05-08T13:53:53Z", GoVersion:"go1.11.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"13+", GitVersion:"v1.13.6-gke.13", GitCommit:"fcbc1d20b6bca1936c0317743055ac75aef608ce", GitTreeState:"clean", BuildDate:"2019-06-19T20:50:07Z", GoVersion:"go1.11.5b4", Compiler:"gc", Platform:"linux/amd64"}
```

I'm installing a chart that consists of a single Kubernetes `Job`. Since I want it to run on every `upgrade`, I've given it a name that changes each time:
`name: keycloak-migrations-{{ now | date "20060102150405" }}`

The problem is that Helm doesn't seem to wait for the `Job` to finish properly:
```
% helm upgrade --wait -i keycloak-post-install .                                                         Release "keycloak-post-install" has been upgraded.
LAST DEPLOYED: Wed Jun 26 20:08:22 2019
NAMESPACE: default
STATUS: DEPLOYED

RESOURCES:
==> v1/ConfigMap
NAME                 DATA  AGE
keycloak-migrations  1     95s

==> v1/Job
NAME                                COMPLETIONS  DURATION  AGE
keycloak-migrations-20190626180822  0/1          7s        7s

==> v1/Pod(related)
NAME                                      READY  STATUS   RESTARTS  AGE
keycloak-migrations-20190626180822-bc7zl  1/1    Running  0         7s
```

I guess this is a bug, since Helm waited for all of the other Jobs I've run using it.
Additional question: is there a simple way to run a `Job` on each `upgrade`?

zh168654:
Resolves #5948 

Helm --wait should wait for all manifest resources to be ready
However helm doesn't wait for jobs to be completed when job appears in the manifest, not in the hooks

I noticed that many are using jobs in the manifest, not just in hooks. They need all the jobs (in the manifest) to be completed to mean that helm release is successfully installed/upgraded

For example, a job performs an initialization task. Although when deployment and related pods are all in the ready state, the job is incomplete and the service is still unavailable, then this release should be considered as not ready

Why jobs not using hooks? we want jobs managed together with the release lifecycle

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -127,7 +127,7 @@ func (c *Client) Create(resources ResourceList) (*Result, error) {
 }
 
 // Wait up to the given timeout for the specified resources to be ready
-func (c *Client) Wait(resources ResourceList, timeout time.Duration) error {
+func (c *Client) Wait(resources ResourceList, timeout time.Duration, waitForJobsEnabled bool) error {
 	cs, err := c.getKubeClient()
 	if err != nil {
 		return err
@@ -137,7 +137,7 @@ func (c *Client) Wait(resources ResourceList, timeout time.Duration) error {
 		log:     c.Log,
 		timeout: timeout,
 	}
-	return w.waitForResources(resources)
+	return w.waitForResources(resources, waitForJobsEnabled)
 }
 
 func (c *Client) namespace() string {
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -127,7 +127,7 @@ func (c *Client) Create(resources ResourceList) (*Result, error) {
 }
 
 // Wait up to the given timeout for the specified resources to be ready
-func (c *Client) Wait(resources ResourceList, timeout time.Duration, waitForJobsEnabled bool) error {
+func (c *Client) Wait(resources ResourceList, timeout time.Duration) error {
 	cs, err := c.getKubeClient()
 	if err != nil {
 		return err
@@ -137,7 +137,21 @@ func (c *Client) Wait(resources ResourceList, timeout time.Duration, waitForJobs
 		log:     c.Log,
 		timeout: timeout,
 	}
-	return w.waitForResources(resources, waitForJobsEnabled)
+	return w.waitForResources(resources, false)
+}
+
+// WaitWithJobs wait up to the given timeout for the specified resources to be ready, including jobs.
+func (c *Client) WaitWithJobs(resources ResourceList, timeout time.Duration) error {
+	cs, err := c.getKubeClient()
+	if err != nil {
+		return err
+	}
+	w := waiter{
+		c:       cs,
+		log:     c.Log,
+		timeout: timeout,
+	}
+	return w.waitForResources(resources, true)
 }
 
 func (c *Client) namespace() string {
--- a/cmd/helm/install.go
+++ b/cmd/helm/install.go
@@ -140,6 +140,7 @@ func addInstallFlags(cmd *cobra.Command, f *pflag.FlagSet, client *action.Instal
 	f.BoolVar(&client.Replace, "replace", false, "re-use the given name, only if that name is a deleted release which remains in the history. This is unsafe in production")
 	f.DurationVar(&client.Timeout, "timeout", 300*time.Second, "time to wait for any individual Kubernetes operation (like Jobs for hooks)")
 	f.BoolVar(&client.Wait, "wait", false, "if set, will wait until all Pods, PVCs, Services, and minimum number of Pods of a Deployment, StatefulSet, or ReplicaSet are in a ready state before marking the release as successful. It will wait for as long as --timeout")
+	f.BoolVar(&client.WaitForJobs, "wait-for-jobs", false, "if set and --wait enabled, will wait until all Jobs have been completed before marking the release as successful. It will wait for as long as --timeout")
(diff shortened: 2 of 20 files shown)


bacongobbler:
can you write a unit test to cover this case. Thanks.

zh168654:
> please write a unit test to cover this case. Thanks.

@bacongobbler Thanks for your attention! I have completed the unit test for this case and any other cases. Please take another look?

mark007:
@zh168654 do you know what the plan is for this one. Is it going to be merged. Itll be a huge addition to helm when this goes in. It might make use of ordered hooks that all run post-install, less attractive than jobs that can run in parallel to the deployment of the pods.

bacongobbler:
Adding to 3.4.0 for review.

zh168654:
@bacongobbler I'm ready for a review on this. Would be great to see this feature in 3.4.0

mattfarina:
@zh168654 Sorry for the slow response. We talked about this because of the impact.

People today depend on jobs not being behind the `--wait` flag. It would be a behavior change to put it behind the flag in Helm v3. In order to wait for jobs we would need a second flat to opt-in. Can you make that change?

zh168654:
@mattfarina This sounds reasonable and I'd like to have a try. Is it okay to use --wait-jobs as the second flag, or do you have any suggestions?

zh168654:
@mattfarina @bacongobbler I've made the changes and introduced `--wait-for-jobs`. So CLI examples:

`$ helm install --wait --wait-for-jobs mysql stable/mysql`

Looking forward to your review. Thanks very much

pkg/kube/client.go
```diff
@@ -127,7 +127,7 @@ func (c *Client) Create(resources ResourceList) (*Result, error) {
 }
 
 // Wait up to the given timeout for the specified resources to be ready
-func (c *Client) Wait(resources ResourceList, timeout time.Duration) error {
+func (c *Client) Wait(resources ResourceList, timeout time.Duration, waitForJobsEnabled bool) error {
```
mattfarina, line 130: `Wait` is part of the public API for the Go package. The signature cannot be changed until Helm v4. Can you please create a new function with a new name (e.g., `WaitWithJobs`). The old function can call the new function.

This has to do with Helm providing a Semantically Versioned API. It applies everywhere `Wait` was changed.

mattfarina (changes requested):

pkg/kube/client.go
zh168654: Thanks for your detailed instructions. Done

zh168654:
@mattfarina Already updated and ready for review again

seaneagan:
Another way to get this functionality would be #8661, which would avoid needing to expose a `--wait-for-jobs` flag.

seaneagan:
Could also consider replacing `--wait-for-jobs` with e.g. `HELM_EXPERIMENTAL_WAIT` in this PR, and then if/when #8661 is implemented, this env var will simply provide expanded functionality beyond just waiting for jobs.

novas0x2a:
In our setup, we've actually run into both situations, some where we wanted jobs to be waited for, and some we did not. We'd been discussing potentially submitting a patch to have helm honor an annotation on the job to determine whether it's waited for; would that approach be more palatable than adding a new flag? (the default could be helm's current behavior)

seaneagan:
@novas0x2a I agree something like a `helm.sh/ignore-wait: true` or `helm.sh/ignore-status: true` could be useful, for all resources not just Jobs, and could potentially fit under the same HELM_EXPERIMENTAL_WAIT flag and HIP, although that too is indeed orthogonal if there were a desire to separate them. It's possible kstatus could implement a similar annotation based functionality, in which case that could potentially be gotten for free in Helm, so a feature request there could be relevant.

technosophos (approved):
The new round of changes look great! Thanks for making the changes.

mattfarina:
@seaneagan waiting for jobs can't be an experiment in Helm v3 because we cannot make a change in Helm v3 around this. It would break semver. The earliest the change could be made in Helm v4. So, a flag makes it something supported in v3. Using `HELM_EXPERIMENTAL_WAIT` would not really make that the case.

There is also a separation in control. Things in the chart like an annotation are for chart creators. Those who use the Helm cli to add flags are Helm users and chart consumers. Those are different roles that are often fulfilled by people at different companies. So a flag or an annotation is about who we give control to in the situation.

mattfarina (approved):

PR #8364, opened 2020-06-29
Complete revision parameter for rollback command

marckhouzam:

This PR adds completion for the optional second parameter of the `rollback` command.
```
$ helm rollback nginx <TAB>
12  13  14  15  16  17  18  19  20  21
$ helm rollback nginx 2<TAB>
20  21
```

Although the revision parameter is optional for the `rollback` command, if the user presses `<TAB>` the completion will always suggest a revision.  I don't have a good way to indicate that the parameter is optional using completion, but I think it is still better to show the completion options in this case.

**If applicable**:
- [x] this PR has been tested for backwards compatibility with go tests and the acceptance-testing repo


bacongobbler (approved):
LGTM! I am going to go ahead and merge this in given that we are about to release the first 3.3.0 release candidate soon. Nicely done!

marckhouzam:
Thanks for getting this into 3.3 @bacongobbler!

PR #8369, opened 2020-06-29
Environment variable for setting the max history for an environment

### LINKED ISSUE #8367: Environment variable for max release history
Environment variable for setting the max history for an environment. This would remove need to set it on each upgrade with the `--history-max` flag. Similar to `TILLER_HISTORY_MAX` variable in Helm 2.

Output of `helm version`:
Helm 3

wawa0210:
Environment variable for max release history

fix #8367

**If applicable**:

### DIFF
--- a/pkg/cli/environment.go
+++ b/pkg/cli/environment.go
@@ -33,6 +33,9 @@ import (
 	"helm.sh/helm/v3/pkg/helmpath"
 )
 
+// defaultMaxHistory sets the maximum number of releases to 0: unlimited
+const defaultMaxHistory = 10
+
 // EnvSettings describes all of the environment settings.
 type EnvSettings struct {
 	namespace string
@@ -56,11 +59,14 @@ type EnvSettings struct {
 	RepositoryCache string
 	// PluginsDirectory is the path to the plugins directory.
 	PluginsDirectory string
+	// MaxHistory is the max release history maintained.
+	MaxHistory int
 }
 
 func New() *EnvSettings {
 	env := &EnvSettings{
 		namespace:        os.Getenv("HELM_NAMESPACE"),
+		MaxHistory:       envIntOr("HELM_MAX_HISTORY", defaultMaxHistory),
 		KubeContext:      os.Getenv("HELM_KUBECONTEXT"),
 		KubeToken:        os.Getenv("HELM_KUBETOKEN"),
 		KubeAPIServer:    os.Getenv("HELM_KUBEAPISERVER"),
@@ -102,6 +108,18 @@ func envOr(name, def string) string {
 	return def
 }
 
+func envIntOr(name string, def int) int {
+	if name == "" {
+		return def
+	}
+	envVal := envOr(name, strconv.Itoa(def))
+	ret, err := strconv.Atoi(envVal)
+	if err != nil {
+		return def
+	}
+	return ret
+}
+
 func (s *EnvSettings) EnvVars() map[string]string {
 	envvars := map[string]string{
 		"HELM_BIN":               os.Args[0],
@@ -114,6 +132,7 @@ func (s *EnvSettings) EnvVars() map[string]string {
 		"HELM_REPOSITORY_CACHE":  s.RepositoryCache,
 		"HELM_REPOSITORY_CONFIG": s.RepositoryConfig,
 		"HELM_NAMESPACE":         s.Namespace(),
+		"HELM_MAX_HISTORY":       strconv.Itoa(s.MaxHistory),
 
 		// broken, these are populated from helm flags and not kubeconfig.
 		"HELM_KUBECONTEXT":   s.KubeContext,
--- a/cmd/helm/root.go
+++ b/cmd/helm/root.go
@@ -50,6 +50,7 @@ Environment variables:
 | $HELM_DATA_HOME                    | set an alternative location for storing Helm data.                                |
 | $HELM_DRIVER                       | set the backend storage driver. Values are: configmap, secret, memory, postgres   |
 | $HELM_DRIVER_SQL_CONNECTION_STRING | set the connection string the SQL storage driver should use.                      |
+| $HELM_MAX_HISTORY                  | set the maximum number of helm release history.                                   |
 | $HELM_NO_PLUGINS                   | disable plugins. Set HELM_NO_PLUGINS=1 to disable plugins.                        |
 | $KUBECONFIG                        | set an alternative Kubernetes configuration file (default "~/.kube/config")       |
 
--- a/cmd/helm/upgrade.go
+++ b/cmd/helm/upgrade.go
@@ -180,7 +180,7 @@ func newUpgradeCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 	f.BoolVar(&client.ReuseValues, "reuse-values", false, "when upgrading, reuse the last release's values and merge in any overrides from the command line via --set and -f. If '--reset-values' is specified, this is ignored")
 	f.BoolVar(&client.Wait, "wait", false, "if set, will wait until all Pods, PVCs, Services, and minimum number of Pods of a Deployment, StatefulSet, or ReplicaSet are in a ready state before marking the release as successful. It will wait for as long as --timeout")
 	f.BoolVar(&client.Atomic, "atomic", false, "if set, upgrade process rolls back changes made in case of failed upgrade. The --wait flag will be set automatically if --atomic is used")
-	f.IntVar(&client.MaxHistory, "history-max", 10, "limit the maximum number of revisions saved per release. Use 0 for no limit")
+	f.IntVar(&client.MaxHistory, "history-max", settings.MaxHistory, "limit the maximum number of revisions saved per release. Use 0 for no limit")
 	f.BoolVar(&client.CleanupOnFail, "cleanup-on-fail", false, "allow deletion of new resources created in this upgrade when upgrade fails")
 	f.BoolVar(&client.SubNotes, "render-subchart-notes", false, "if set, render subchart notes along with the parent")
 	f.StringVar(&client.Description, "description", "", "add a custom description")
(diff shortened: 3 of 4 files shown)


wawa0210:
@hickeyma @marckhouzam 

I'm a little confused, if other modules (such as rollback) want to use this method, `historyMaxFromEnv()`, where is the best place to put it. Do you have any better ideas

bacongobbler:
`pkg/cli` is used to read and set other environment variables. That might be a better place to put this code logic.

https://github.com/helm/helm/blob/9b90e848ca659e147cae66dde9c16dcda578f113/pkg/cli/environment.go#L105

wawa0210:
> `pkg/cli` is used to read and set other environment variables. That might be a better place to put this code logic.
> 
> https://github.com/helm/helm/blob/9b90e848ca659e147cae66dde9c16dcda578f113/pkg/cli/environment.go#L105

Thanks for reply, done.

pkg/cli/environment.go
```diff
@@ -56,6 +59,8 @@ type EnvSettings struct {
 	RepositoryCache string
 	// PluginsDirectory is the path to the plugins directory.
 	PluginsDirectory string
+	// HistoryMax is the name of the env var for setting max history.
+	HistoryMax int
```
hickeyma, line 63: Observation: I think for consistency with the arg `--history-max`, `HistoryMax` should remain but really `MaxHistory` would read better.

pkg/cli/environment.go
```diff
@@ -68,6 +73,7 @@ func New() *EnvSettings {
 		RegistryConfig:   envOr("HELM_REGISTRY_CONFIG", helmpath.ConfigPath("registry.json")),
 		RepositoryConfig: envOr("HELM_REPOSITORY_CONFIG", helmpath.ConfigPath("repositories.yaml")),
 		RepositoryCache:  envOr("HELM_REPOSITORY_CACHE", helmpath.CachePath("repository")),
+		HistoryMax:       envIntOr("HISTORY_MAX", defaultMaxHistory),
```
hickeyma, line 76: HISTORY_MAX --> HELM_HISTORY_MAX

pkg/cli/environment.go
```diff
@@ -102,6 +108,18 @@ func envOr(name, def string) string {
 	return def
 }
 
+func envIntOr(name string, def int) int {
+	if name == "" {
+		return def
+	}
+	envVal := envOr(name, "")
```
hickeyma, line 115: Might be safer to also pass the default value instead of ""

hickeyma (changes requested):
Thanks for pushing the PR @wawa0210. Some small things to address.

pkg/cli/environment.go
wawa0210: done.

pkg/cli/environment.go
wawa0210: done.

pkg/cli/environment.go
wawa0210: I am also hesitant about this, if ok, I can update to "MaxHistory"

😄😄😄

liuming-dev: I agree with @hickeyma, `MaxHistory` reads better.

pkg/cli/environment.go
```diff
@@ -68,6 +73,7 @@ func New() *EnvSettings {
 		RegistryConfig:   envOr("HELM_REGISTRY_CONFIG", helmpath.ConfigPath("registry.json")),
 		RepositoryConfig: envOr("HELM_REPOSITORY_CONFIG", helmpath.ConfigPath("repositories.yaml")),
 		RepositoryCache:  envOr("HELM_REPOSITORY_CACHE", helmpath.CachePath("repository")),
+		HistoryMax:       envIntOr("HELM_HISTORY_MAX", defaultMaxHistory),
```
liuming-dev, line 76: `HistoryMax` → `MaxHistory`

pkg/cli/environment.go
```diff
@@ -119,6 +137,7 @@ func (s *EnvSettings) EnvVars() map[string]string {
 		"HELM_KUBECONTEXT":   s.KubeContext,
 		"HELM_KUBETOKEN":     s.KubeToken,
 		"HELM_KUBEAPISERVER": s.KubeAPIServer,
+		"HELM_HISTORYMAX":    strconv.Itoa(s.HistoryMax),
```
liuming-dev, line 140: `HELM_HISTORYMAX ` → `HELM_MAXHISTORY`

pkg/cli/environment.go
wawa0210: Deal, modified.  😄

hickeyma (changes requested):
Sorry for the confusion here. My comment in https://github.com/helm/helm/pull/8369#discussion_r447449985 is an observation. I would prefer to change it to `MaxHistory` but I think because of the flag which cannot be changed until Helm 4, we should follow the naming for consistency.

@wawa0210 Do you mind changing it back?

wawa0210:
> I would prefer to change it to `MaxHistory` but I think because of the flag which cannot be changed until Helm 4, we should follow the naming for consistency.

What we change is only the name of the environment variable, not the name of the command line flag~ Will this affect? If there is a negative impact, I can try to roll back

hickeyma (changes requested):
Looking good @wawa0210. Can you fix the CI build as style issue in `pkg/cli/environment.go`.

hickeyma:
I am ok with the naming now.

wawa0210:
> I am ok with the naming now.

It seems that all problems have been fixed, can you help approve?

marckhouzam:
@hickeyma do you think the new environment variable, since it is user-facing, should use the same name as the flag?  So HELM_HISTORY_MAX? I'm not convinced it is essential but I wanted to ask.

wawa0210:
@hickeyma  @marckhouzam @liuming-dev 

All questions have been answered, can you help approve?

pkg/cli/environment.go
```diff
@@ -119,6 +137,7 @@ func (s *EnvSettings) EnvVars() map[string]string {
 		"HELM_KUBECONTEXT":   s.KubeContext,
 		"HELM_KUBETOKEN":     s.KubeToken,
 		"HELM_KUBEAPISERVER": s.KubeAPIServer,
+		"HELM_MAX_HISTORY":   strconv.Itoa(s.MaxHistory),
```
hickeyma, line 140: Move just after `HELM_NAMESPACE` because comment applies to the kube vars

pkg/cli/environment.go
```diff
@@ -56,6 +59,8 @@ type EnvSettings struct {
 	RepositoryCache string
 	// PluginsDirectory is the path to the plugins directory.
 	PluginsDirectory string
+	// MaxHistory is the name of the env var for setting max history.
```
hickeyma, line 62: MaxHistory is the name of the env var for setting max history. --> MaxHistory is the max release history maintained.

hickeyma (changes requested):
Nearly there @wawa0210, thanks. A few small nits and need to add the variable and description to https://github.com/helm/helm/blob/master/cmd/helm/root.go#L52

pkg/cli/environment.go
hickeyma: no, `HELM_MAX_HISTORY` instead

hickeyma:
> do you think the new environment variable, since it is user-facing, should use the same name as the flag? So HELM_HISTORY_MAX? I'm not convinced it is essential but I wanted to ask.

Thought about it alright @marckhouzam and its something I deliberated over. I think it can remain as its a new variable and I would hope the flag would be changed in Helm 4 to be more descriptive.

hickeyma:
I am still a little unconvinced that the env variable needs to be global in the code (i.e. in `pkg/cli`) as I think it will probably only be used by `upgrade` as it makes the most sense. I am however, open to the feedback from @bacongobbler and @marckhouzam in https://github.com/helm/helm/pull/8369#issuecomment-651420879 and https://github.com/helm/helm/issues/8367#issuecomment-651041514.

hickeyma:
@wawa0210 If you can updated this asap, we can try and get it into the 3.3.0 release. It is dues the RC today I think.
/cc @bacongobbler

marckhouzam:
> I am still a little unconvinced that the env variable needs to be global in the code (i.e. in pkg/cli) as I think it will probably only be used by upgrade as it makes the most sense.

@hickeyma I guess it depends on if we want to apply the Max history logic to `helm rollback` too?  My impression is yes, but I can't evaluate if there are ramifications to doing that.

bacongobbler:
we're cutting rc.1 later this morning, so we'll have to move this over to the 3.4.0 milestone.

bacongobbler:
>  I guess it depends on if we want to apply the Max history logic to `helm rollback` too?

Yes, this should also be made available for `helm rollback` as it was for Helm 2.

pkg/cli/environment.go
wawa0210: Feeling your feedback, done.

wawa0210:
> @wawa0210 If you can updated this asap, we can try and get it into the 3.3.0 release. It is dues the RC today I think.

Sorry, I didn't have time to update yesterday, maybe the time is too late to 3.3.0 release

hickeyma (changes requested):
It is nearly there @wawa0210. Do you mind adding a unit test in https://github.com/helm/helm/blob/master/pkg/cli/environment_test.go to test for the variable?

hickeyma:
> Sorry, I didn't have time to update yesterday, maybe the time is too late to 3.3.0 release

That's ok. It can go into 3.4.

wawa0210:
> It is nearly there @wawa0210. Do you mind adding a unit test in https://github.com/helm/helm/blob/master/pkg/cli/environment_test.go to test for the variable?

Added, troublesome review again

hickeyma (approved):
LGTM, thanks @wawa0210

pchico83:
is this feature included in 3.3.0? Will be included on the next release?

marckhouzam:
@pchico83 this feature is not in the 3.3 release but will be part of 3.4

Issue #8279, opened 2020-06-10
Upgrade with --reuse-values breaks values set with --set-file when using a binary file


srajkovic:
Output of `helm version`: `version.BuildInfo{Version:"v3.2.3", GitCommit:"8f832046e258e2cb800894579b1b3b50c2d83492", GitTreeState:"clean", GoVersion:"go1.13.12"}`

Output of `kubectl version`: `Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.2", GitCommit:"52c56ce7a8272c798dbc29846288d7cd9fbae032", GitTreeState:"clean", BuildDate:"2020-04-16T11:56:40Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"windows/amd64"}
Server Version: version.Info{Major:"1", Minor:"16+", GitVersion:"v1.16.8-eks-e16311", GitCommit:"e163110a04dcb2f39c3325af96d019b4925419eb", GitTreeState:"clean", BuildDate:"2020-03-27T22:37:12Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}`

Cloud Provider/Platform (AKS, GKE, Minikube etc.): EKS

Scenario: I want to use --set-file to pass in a binary file, like a certificate, as a value, which I then use in a secrets.yml template, which base64 encodes the file into a secret. In my deployment I mount that secret as a volume into my container, and my service reads it.

This works great for install, and great if I pass the file in via --set-file again on upgrade, but if I don't pass the file in again and rely on --reuse-values the file is mangled, with ton's of random garbage in the resulting base64, like `+/ve` and `+977` all over the place. If I pass in a file via --set-file that's just ASCII text (even if the text is very large, much larger than the certificate), the value is reused just fine, so it does appear to be an issue with the binary file and how it's reused by --reuse-values.

https://github.com/helm/helm/issues/4131#issuecomment-592018120 mentions this, but that issue is closed so I figured it might not receive any more responses.

srajkovic:
Thanks Matt! I don't have a ton of time or any Go/Helm internals knowledge right now, but if I do have some time free up I'll try to take a crack at it.

Cweiping:
@srajkovic  I can't repeat the question. Can you give me an example?

srajkovic:
Sorry, I haven't had a chance to make a small repro project. I did notice on upgrades I get this error:
```
USER-SUPPLIED VALUES:
[...]

PR #8373, opened 2020-06-29
Add TestUpgradeWithValueFile test

lann:
This test (currently _failing_) demonstrates issue #8279


bacongobbler:
thank you for the test case. Are you just asserting/demonstrating incorrect behaviour for #8279, or are you looking to try and merge this code? If the former, we should probably go ahead and close this. Thanks.

lann:
Primarily demonstrating the behavior, but this would be a regression test for it as well.

If you don't want the open PR floating around pending a fix, closing is fine.

PR #8384, opened 2020-07-01
change the DNS name limit

### LINKED ISSUE #6006: Why Helm release name is a DNS label with max length of 53 characters?
Helm documentation states the following:
-1) (k8s) metadata.name is restricted to a maximum length of 63 characters because of limitations to the DNS system
-2) For that reasons, release names are (DNS labels that are) limited to 53 characters

Statement 1) is not correct. 
k8s does not impose a max length of 63 characters on resource names.
The actual max length for a resource name is 253 characters. 
The k8s resource name is in fact a DNS subdomain (that can be made of one or more DNS labels).

Should we make the release name a DNS subdomain?

renatosuero:
Signed-off-by: Renato Suero <suerorenato@gmail.com>

closes #6006 

I've increased the limit for resource naming. The project was using 63(the value for Kubernetes version 1.4) now the new value is 253.  

**If applicable**:

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -52,10 +52,10 @@ import (
 
 // releaseNameMaxLen is the maximum length of a release name.
 //
-// As of Kubernetes 1.4, the max limit on a name is 63 chars. We reserve 10 for
-// charts to add data. Effectively, that gives us 53 chars.
-// See https://github.com/helm/helm/issues/1528
-const releaseNameMaxLen = 53
+// As of Kubernetes 1.18, the max limit on a name is 253 chars. We reserve 10 for
+// charts to add data. Effectively, that gives us 231 chars.
+// See https://github.com/helm/helm/issues/6006#issuecomment-652291971
+const releaseNameMaxLen = 231
 
 // NOTESFILE_SUFFIX that we want to treat special. It goes through the templating engine
 // but it's not a yaml file (resource) hence can't have hooks, etc. And the user actually


mattfarina:
According to the [Kubernetes docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names), some resource names have a max length of 63 characters while others have a max length of 253 characters. I'm trying to figure out which objects have the 63 character length right now. It appears that Containers, Namespaces, Service Ports, Volumes, CSI Drivers and EndpointPorts are all limited to 63 character names.

Charts will occasionally create namespaces. How should this be handled?

bacongobbler:
ping.

renatosuero:
Sorry  busy days. I checked in the kubernetes code, I found label and job has the limitation. Regarding how to should this be handled, maybe we could check the resource and apply the correct limit, what do you think @mattfarina ?

liuming-dev:
The fixed max length releaseName may face a special case:

secret name max length limitation is 253.

the release name length is 231, the secret name length is 253 equally, and then upgrade the release to version 10, the computed secret name length could be 254, which is not satisfied with kubernetes object name spec.

renatosuero:
@liuming-dev, I didn't get your point. the 231 value was calculated [here](https://github.com/helm/helm/issues/6006#issuecomment-652291971). Could you explain your point

liuming-dev:
> @liuming-dev, I didn't get your point. the 231 value was calculated [here](https://github.com/helm/helm/issues/6006#issuecomment-652291971). Could you explain your point

For example: 
release name is `aaaaaaaa…`(231 characters of `a`), and the latest version is 9, that means sh.helm.release.v1.`aaaaaaaa…`.v9, length up to 253 already.  when upgrade the release to version10, sh.helm.release.v1.`aaaaaaaa…`.v10, length up to 254 and has exceeded the maximum limit. 

Another point need to be noticed, release name is used in release secret label, that means its length should not be more than 63 characters.

renatosuero:
thanks @liuming-dev I got your point now. what do you suggest to fix that?

renatosuero:
ping

hickeyma:
@renatosuero This potential issue here is that not all Kubernetes resources are max character length. Some charts name resources using the release name. Have you investigated all Kubernetes resource/object name sizes?

renatosuero:
@hickeyma I'll double check again, but If I'm not wrong the labels are the only one which still have 63 chars

bacongobbler:
@renatosuero as a suggestion, you could write unit tests to generate kubernetes resources using `releaseNameMaxLen` to verify. That would help ensure future updates won't accidentally break something as well.

Have a look at the kubernetes test infrastructure for ideas.

gejunqiang:
Any updated?

bacongobbler:
closing as inactive. If someone wants to carry on the work and address some of the other maintainer's comments, feel free.

XenoAmess:
I don't think it be necessary for helm to make the package totally usable by k8s.
'cause k8s use helm, not versa.
It is the user's duty to make sure their package usable by k8s, or elsewhere.
I really think we should only print a warn message at console when people wanna make such helm, but no hard liminitions.

Issue #6783, opened 2019-10-24
crd-install apiextensions.k8s.io/v1 does not work


strainovic:
https://github.com/helm/helm/blob/0a9bf4f966726f0bb0faa6a82db29909abe16126/pkg/kube/client.go#L597

Helm version: 2.15.1

```
Client: &version.Version{SemVer:"v2.15.1", GitCommit:"cf1de4f8ba70eded310918a8af3a96bfe8e7683b", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.15.1", GitCommit:"cf1de4f8ba70eded310918a8af3a96bfe8e7683b", GitTreeState:"clean"}
```

Kubernetes version: 1.16.2

```
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-15T19:18:23Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-15T19:09:08Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}
```

Kubernetes distribution: Canonical Microk8s

I've tried to deploy CRD with new apiextensions.k8s.io/v1 API but once I add "helm.sh/hook": crd-install installation failed with error "unable to convert to CRD type...".

```
Error: unable to convert to CRD type: no kind "CustomResourceDefinition" is registered for version "apiextensions.k8s.io/v1" in scheme "k8s.io/client-go/kubernetes/scheme/register.go:65"
```

apiextensions.k8s.io/v1beta1 is depricated in 1.16 and will be removed in 1.19.

How much is it likely for this bug to be fixed in 2.x version?

Thanks

thomastaylor312:
@strainovic Thanks for reporting this. So just to double check, apiextensions v1 is new in 1.16 right? If so, this would be a bug we should fix and we would have to update our version of client-go.

thomastaylor312:
Actually I just found it in the release notes. Look like we will need to update client-go for both v2 and v3: https://kubernetes.io/docs/setup/release/notes/#api-changes

bacongobbler:
[...]

PR #8405, opened 2020-07-06
Backporting fix from Helm v3

bridgetkromhout:
We had input from @caseydavenport in https://github.com/helm/helm/issues/6783 that the dev-v2 branch was missing a scheme addition from v3. @bacongobbler and @thomastaylor312 said we should backport it, so I'm giving that a try in this PR.

I'm not very familiar with this codebase, so input is welcome. I tried just copying over the relevant v3 code. I don't see any pitfalls, but that doesn't mean they aren't lurking. 

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -79,17 +79,23 @@ type Client struct {
 	Log func(string, ...interface{})
 }
 
+var addToScheme sync.Once
+
 // New creates a new Client.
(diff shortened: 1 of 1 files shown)


bridgetkromhout:
Okay, this is more complex than I imagined; closing until I can test more thoroughly, so nobody else wastes time on it!

PR #8416, opened 2020-07-08
test: improve repo_update test fatal message

lemonli:
Signed-off-by: lemonli <liwenjun0323@gmail.com>

I got error message when do `make test`: 
```
--- FAIL: TestUpdateCustomCacheCmd (76.01s)
repo_update_test.go:70: error finding created index file in custom cache: &os.PathError{Op:"stat", Path:"/var/folders/1m/pst1v4fd2t320wn5hx78vrdw6kv787/T/helm433448320/updcustomcache/charts-index.yaml", Err:0x2}
```
I don't known what is the reason until I read whole codes in repo_update. I think detail error message may be helpful.

`TestUpdateCustomCacheCmd` in `cmd/helm/repo_update_test.go` download index file from `https://kubernetes-charts.storage.googleapis.com` to cache dir. But in China, the url can not be reached without a VPN proxy. Detail error message may help to found this reason.

**If applicable**:

### DIFF
--- a/cmd/helm/repo_update_test.go
+++ b/cmd/helm/repo_update_test.go
@@ -67,7 +67,7 @@ func TestUpdateCustomCacheCmd(t *testing.T) {
 		t.Fatal(err)
 	}
 	if _, err := os.Stat(filepath.Join(cachePath, "charts-index.yaml")); err != nil {
-		t.Fatalf("error finding created index file in custom cache: %#v", err)
+		t.Fatalf("error finding created index file in custom cache: %#v \ngot: %s", err, out.String())
 	}
 }
 


lemonli:
thx, @hickeyma @mattfarina

hickeyma (commented):
Thanks for pushing the PR @lemonli. I would have thought if there is an issue like resolving URL then it would have failed during the repo update command in https://github.com/helm/helm/blob/master/cmd/helm/repo_update_test.go#L66? I am therefore not sure that the output would be of use here.

lemonli:
> Thanks for pushing the PR @lemonli. I would have thought if there is an issue like resolving URL then it would have failed during the repo update command in https://github.com/helm/helm/blob/master/cmd/helm/repo_update_test.go#L66? I am therefore not sure that the output would be of use here.

It is not failed when call https://github.com/helm/helm/blob/master/cmd/helm/repo_update.go#L80 and then https://github.com/helm/helm/blob/master/cmd/helm/repo_update.go#L84 write error msg into `out io.Writer`

hickeyma:
@lemonli Sure, I realised that later but didn't get to respond.

I am unsure about adding it here as these are units tests and not sure of the overall value. Maybe other @helm/all-maintainers might have an opinion on this.

jdolitsky (approved):
I think that if this helps debug an issue, its no harm to merge since its just in unit tests

bacongobbler (changes requested):
I don't think the output needs to be displayed to fix this issue. You need to adjust the format output so that the error message is displayed instead of its type. In other words, change `%#v` to `%v` and you should see the error message. There's no need to display the command output.

lemonli:
> I don't think the output needs to be displayed to fix this issue. You need to adjust the format output so that the error message is displayed instead of its type. In other words, change `%#v` to `%v` and you should see the error message. There's no need to display the command output.

https://github.com/helm/helm/blob/master/cmd/helm/repo_update.go#L93 write error msg into `out io.Writer`

The origin message is
`
--- FAIL: TestUpdateCustomCacheCmd (76.44s)
    repo_update_test.go:70: error finding created index file in custom cache: &os.PathError{Op:"stat", Path:"/var/folders/1m/pst1v4fd2t320wn5hx78vrdw6kv787/T/helm380963295/updcustomcache/charts-index.yaml", Err:0x2}
`

In my commit, it is
`
--- FAIL: TestUpdateCustomCacheCmd (76.44s)
    repo_update_test.go:70: error finding created index file in custom cache: &os.PathError{Op:"stat", Path:"/var/folders/1m/pst1v4fd2t320wn5hx78vrdw6kv787/T/helm380963295/updcustomcache/charts-index.yaml", Err:0x2}
        got: Hang tight while we grab the latest from your chart repositories...
        ...Unable to get an update from the "charts" chart repository (https://kubernetes-charts.storage.googleapis.com):
        	Get "https://kubernetes-charts.storage.googleapis.com/index.yaml": dial tcp 34.64.4.16:443: connect: operation timed out
`

bacongobbler:
Closing as something we do not plan to merge. We only require the error message to debug the test failure; not the full CLI output.

As mentioned previously, changing `%#v` to `%v` in the formatter would display the error message. I will address this in another PR.

bacongobbler:
opened #8929 to address this issue.

PR #8426, opened 2020-07-09
fix lint error if parent chart doesn't have template directory

### LINKED ISSUE #8418: Subcharts templates not linted if parent chart doesn't have templates directory
I have seen a linting issue in sub chart templates that only is only detected by lint when the parent chart has a templates directory.
Looking at the Helm source it seems (I am not familiar with Go) like linting is only performed on templates if the parent directory has a templates directory.

Previous to the this point the parent chart did not need any templates so our issue is only discovered now.

Have I understood the Helm source correctly?
Is this intentional behaviour for the lint of templates? (It was not the _expected_ behaviour)

https://github.com/helm/helm/blob/bf9c64f48ee8f799c89de6cfbd44302f798b3531/pkg/lint/rules/template.go#L56

wawa0210:
Signed-off-by: wawa0210 <xiaozhang0210@hotmail.com>

At the time of helm lint, if the template directory does not exist, no error will be reported, just a warning prompt, which does not meet expectations

fix https://github.com/helm/helm/issues/8418

### DIFF
--- a/pkg/lint/rules/template.go
+++ b/pkg/lint/rules/template.go
@@ -50,12 +50,10 @@ func Templates(linter *support.Linter, values map[string]interface{}, namespace 
 	path := "templates/"
 	templatesPath := filepath.Join(linter.ChartDir, path)
 
-	templatesDirExist := linter.RunLinterRule(support.WarningSev, path, validateTemplatesDir(templatesPath))
-
-	// Templates directory is optional for now
-	if !templatesDirExist {
-		return
-	}
+	// Templates directory is optional
+	// if a warning level alarm occurs, just output the result,
+	// and then continue to execute the subsequent logic
+	linter.RunLinterRule(support.WarningSev, path, validateTemplatesDir(templatesPath))
 
 	// Load chart and parse templates
 	chart, err := loader.Load(linter.ChartDir)
--- a/cmd/helm/testdata/output/lint-chart-with-bad-subcharts-with-subcharts.txt
+++ b/cmd/helm/testdata/output/lint-chart-with-bad-subcharts-with-subcharts.txt
@@ -1,6 +1,7 @@
 ==> Linting testdata/testcharts/chart-with-bad-subcharts
 [INFO] Chart.yaml: icon is recommended
 [WARNING] templates/: directory not found
+[ERROR] templates/: error unpacking bad-subchart in chart-with-bad-subcharts: validation: chart.metadata.name is required
 [ERROR] : unable to load chart
 	error unpacking bad-subchart in chart-with-bad-subcharts: validation: chart.metadata.name is required
 
@@ -10,6 +11,7 @@
 [ERROR] Chart.yaml: version is required
 [INFO] Chart.yaml: icon is recommended
 [WARNING] templates/: directory not found
+[ERROR] templates/: validation: chart.metadata.name is required
 [ERROR] : unable to load chart
 	validation: chart.metadata.name is required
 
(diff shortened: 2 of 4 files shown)


mattfarina:
@wawa0210 The change you made affects tests. Those will also need to be updated.

wawa0210:
> @wawa0210 The change you made affects tests. Those will also need to be updated.

Thanks for reminding, I read the error and have some confusion

The error message as the follows 
```
 helm_test.go:67: does not match golden file testdata/output/lint-chart-with-bad-subcharts.txt
```

Then I targeted the relevant code

https://github.com/helm/helm/blob/e31818f2efd3e1e0867d9c189507c25c8102264a/cmd/helm/lint_test.go#L24-L37

It should be that the direct result of the command is inconsistent with the content of the golden corresponding file, but I did not find the file template corresponding to the golden. Can you give some help and tips, thank you very much

wawa0210:
The ci error has been fixed, please help me review it  again

@mattfarina

wawa0210:
@bacongobbler @mattfarina 

Can you help me review this pr again ?

pkg/lint/rules/template.go
```diff
@@ -50,7 +50,7 @@ func Templates(linter *support.Linter, values map[string]interface{}, namespace
 	path := "templates/"
 	templatesPath := filepath.Join(linter.ChartDir, path)
 
-	templatesDirExist := linter.RunLinterRule(support.WarningSev, path, validateTemplatesDir(templatesPath))
+	templatesDirExist := linter.RunLinterRule(support.ErrorSev, path, validateTemplatesDir(templatesPath))
```
hickeyma, line 53: I don't think the change should be here. An absent `templates` directory should only be a warning as it does not affect a chart being deployed. See below for removing the check on template directory exists.

pkg/lint/rules/template.go
```diff
@@ -50,7 +50,7 @@ func Templates(linter *support.Linter, values map[string]interface{}, namespace
...
 
 	// Templates directory is optional for now
 	if !templatesDirExist {
```
hickeyma, line 56: This check should be removed instead. This then would then remove most if not all updates to the unit tests. It might be worth though adding a unit test for an error in a subchart if it does not exist already.

hickeyma (changes requested):
Thanks for pushing the PR @wawa0210. Some comments provided inline.

pkg/lint/rules/template.go
wawa0210: > This check should be removed instead. This then would then remove most if not all updates to the unit tests. It might be worth though adding a unit test for an error in a subchart if it does not exist already.

Do you mean that you don't verify the `template`, just remove it?

hickeyma (changes requested):
@wawa0210 Thanks for pushing updates. Some small nits inline.

While testing this manually, it  seems that without the change (i.e. off main branch) it is now catching the error. I cannot explain how this has happened unless changes in some other rule is now catching it prior to checking for the `templates` directory. Do you mind testing this?

I any case, it might still be worth making the change as the `templates` directory is not mandatory for a chart.

wawa0210:
> I any case, it might still be worth making the change as the `templates` directory is not mandatory for a chart.

Do you mean we need to remove the `validateTemplatesDir` validation, even the warning level?

hickeyma:
@wawa0210 Thanks for the work on this PR. As you mentioned that changes in PR #7970 also fixes the issue #8418. 

This unfortunately makes the PR now obsolete. There is therefore no reason to push these changes. If anything it add an extra unnecessary error line similar to the following: `[ERROR] templates/: error unpacking bad-subchart in chart-with-bad-subcharts: validation: chart.metadata.name is required`.
 What do you think?

wawa0210:
> What do you think?

Agree with you, i will close this pr

PR #8437, opened 2020-07-13
fix(kube): use logger instead of fmt.Printf

hiddeco:
This seemed to be the only place where `fmt.Printf` was used instead of `Log` or `fmt.Fprintf(out, ...)`. Given `Log` is available within the client, accepting this PR would give consumers better control over how things are printed during test runs.


stefanprodan:
This will help us with https://github.com/fluxcd/helm-controller/issues/11

donggangcj:
👍 Nice catch!

hickeyma (approved):
LGTM, thanks @hiddeco

PR #8438, opened 2020-07-13
feature(show): support jsonpath output for `helm show value`

### LINKED ISSUE #7787: `helm show values` should have a way to specify what values to show
`Helm show values` doesn't have a way to specify which values you want to output. It would be nice to have this because  tools like bash don't have a yaml parser natively so to get a specific value from a chart I have to download a yaml parser. I believe helm should have this integrated because helm seems to be using something like this for `helm install --set` already.

donggangcj:

Closes #7787

1. This is a preliminary realization. I will improve the implementation after listening to everyone's advice.
2. Only `helm show values` support the jsonpath format.
3. This is  implemented through Library `k8s.io/cli-runtime` and related document is [here](https://kubernetes.io/docs/reference/kubectl/jsonpath/).

**If applicable**:

### DIFF
--- a/cmd/helm/show.go
+++ b/cmd/helm/show.go
@@ -152,7 +152,7 @@ func addShowFlags(subCmd *cobra.Command, client *action.Show) {
 
 	f.BoolVar(&client.Devel, "devel", false, "use development versions, too. Equivalent to version '>0.0.0-0'. If --version is set, this is ignored")
 	if subCmd.Name() == "values" {
-		f.StringVar(&client.JsonPathTemplate, "jsonpath", "", "use jsonpath output format")
+		f.StringVar(&client.JSONPathTemplate, "jsonpath", "", "use jsonpath output format")
 	}
 	addChartPathOptionsFlags(f, &client.ChartPathOptions)
 
--- a/cmd/helm/show.go
+++ b/cmd/helm/show.go
@@ -152,7 +152,7 @@ func addShowFlags(subCmd *cobra.Command, client *action.Show) {
 
 	f.BoolVar(&client.Devel, "devel", false, "use development versions, too. Equivalent to version '>0.0.0-0'. If --version is set, this is ignored")
 	if subCmd.Name() == "values" {
-		f.StringVar(&client.JSONPathTemplate, "jsonpath", "", "use jsonpath output format")
+		f.StringVar(&client.JSONPathTemplate, "jsonpath", "", "supply a JSONPath expression to filter the output")
 	}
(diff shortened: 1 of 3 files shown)


cmd/helm/show.go
```diff
@@ -151,6 +151,9 @@ func addShowFlags(subCmd *cobra.Command, client *action.Show) {
 	f := subCmd.Flags()
 
 	f.BoolVar(&client.Devel, "devel", false, "use development versions, too. Equivalent to version '>0.0.0-0'. If --version is set, this is ignored")
+	if subCmd.Name() == "values" {
+		f.StringVar(&client.JSONPathTemplate, "jsonpath", "", "use jsonpath output format")
```
mattfarina, line 155: The help text here isn't very descriptive. A user unfamiliar with JSONPath might think this is a binary flag. Could the help text read something like:

> supply a JSONPath expression to filter the output

mattfarina (commented):
This mostly looks good. I can imagine using this flag and pattern elsewhere. I have one nit about the language of the flag. Could you please update it.

cmd/helm/show.go
donggangcj: ok! done;

jdolitsky (approved):
LGTM

pkg/action/show.go
```diff
@@ -87,9 +89,17 @@ func (s *Show) Run(chartpath string) (string, error) {
 		if s.OutputFormat == ShowAll {
 			fmt.Fprintln(&out, "---")
 		}
-		for _, f := range s.chart.Raw {
-			if f.Name == chartutil.ValuesfileName {
-				fmt.Fprintln(&out, string(f.Data))
+		if s.JSONPathTemplate != "" {
+			printer, err := printers.NewJSONPathPrinter(s.JSONPathTemplate)
+			if err != nil {
+				return "", fmt.Errorf("error parsing jsonpath %s, %v", s.JSONPathTemplate, err)
```
hickeyma, line 95: Better to use `errors.Wrapf`. Pattern used throughout codebase including https://github.com/helm/helm/blob/master/pkg/action/install.go#L144.

hickeyma (changes requested):
Thanks for pushing this functionality @donggangcj. I think it will be useful to users.

Small change inline.

pkg/action/show.go
donggangcj: Thank you for your detailed instructions！ Done;

hickeyma (approved):
LGTM, thanks @donggangcj

Issue #3635, opened 2018-03-08
Proposal: sortable manifests (templates)


phemmer:
This is a proposal to solve issues such as #2994. I wanted to open this in a separate issue so that this specific implementation could be discussed and potentially declined separately from the problem report.

The proposal is to:
1. Add a `helm.sh/order-weight` annotation to the first object in a manifest (a.k.a template) file. The annotation would sort the files within a chart.
  An alternative here would be a comment header at the top of the file (e.g. `# weight: 123`)
2. Add a `weight` attribute to each dependency in `requirements.yaml`. The attribute would sort the charts as whole units.
  An alternative here would be a `sort: true` in the top level of `requirements.yaml`, which would enable sorting the dependencies in the listed order.

In the event that a subchart is required by multiple charts, the highest level chart's requirements sets the weight of that chart.

Sorting precedence would then become:
(kind sorting) -> (annotation sorting) -> (subchart sorting) -> (chart sorting)

Weight would be a signed integer, with a default of 0. Thus all existing charts would be sorted the exact same way since kind sorting would not be overridden.

&nbsp;

I'd be willing to work on this as this is currently a major blocker for us since we're using object types that helm is not aware of, and thus it applies them in the wrong order.  
However I wanted to get approval that this would be accepted before working on it because the amount of work involved would be moderately sized. Right now manifests are sorted without regard to the chart they came from, and that context is not available at sort time. So the code would have to be restructured a bit.

thomastaylor312:
This is definitely an interesting proposal. I'd be interested in @technosophos' or @adamreese's thoughts on this. We could discuss this in one of the dev calls as well. This next week's call is focused on Helm 3, but we could slot it in after that as well

thomastaylor312:
One more clarification question here, is this purely to solve the CRD use case or is there another use case it would solve for you?

phemmer:
> One more clarification question here, is this purely to solve the CRD use case or is there another use case it would solve for you?

Our use case is actually around openshift. Since helm doesn't recognize any of the openshift resource types, it treats them all as unknown and puts them last. Same thing as the CRD though.
I think there's also another more advanced use case that I ran into the other day with ordering between the charts (a resource in one chart depended on a resource in another chart), but I can't recall what that was.

darren-bell-nanthealth:
Hello. Has there been any progress with this proposal or is there an alternative approach. We are particularly interested in the weighting/ordering of the charts in the requirements.yaml.

[...]

PR #8441, opened 2020-07-13
Add support for weights for non-hook artifacts

### LINKED ISSUE #8439: Openshift SCC should be installed before any resource which creates a pod
When creating a [SCC](https://docs.openshift.com/enterprise/3.0/admin_guide/manage_scc.html) and service account as part of a helm chart, since it is not ordered to be created before deployment and other resources, the scc does not have an effect on the pods. 

The simplest fix might be to just add kind "SecurityContextConstraints" just after "ServiceAccount" in the kind_sorter.go file.

https://github.com/helm/helm/blob/release-3.0/pkg/releaseutil/kind_sorter.go

Output of `helm version`:

Output of `kubectl version`:
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.0", GitCommit:"9e991415386e4cf155a24b1da15becaa390438d8", GitTreeState:"clean", BuildDate:"2020-03-26T06:16:15Z", GoVersion:"go1.14", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"16+", GitVersion:"v1.16.2+18cfcc9", GitCommit:"18cfcc9", GitTreeState:"clean", BuildDate:"2020-05-25T09:09:52Z", GoVersion:"go1.12.12", Compiler:"gc", Platform:"linux/amd64"}

ROKS (RedHat Openshift Kubernetes service - IBM Cloud)

ashokponkumar:
closes #8439 #3635

It adds the capability to order manifests. It uses a functionality very similar to helm.io/hook-weight. It adds helm.io/order-weight.

Ueses the following logic to support any CR.

Have an annotation by the name "helm.sh/weight", which takes an Integer value (-ve or +ve). This a concept that Helm anyway uses for hooks ("helm.sh/hook-weight").
We retain the InstallOrder. We install CRs which have a negative weight in ascending order before those in InstallOrder. Then those with InstallOrder, and then Install those with +ve weights in ascending order.
If anyone specifies a weight for a resource in Install order, which is +ve or -ve, it goes with that batch.
If the weight is 0, then it happens just before those in the install order. So if there is a resource which in InstallOrder, but has a weight of 0, it will be applied before any other resource.

**If applicable**:


liuming-dev:
I like this PR, it gives users more flexibilities to define resources install/uninstall order. 


> If the weight is 0, then it happens just before those in the install order. So if there is a resource which in InstallOrder, but has a weight of 0, it will be applied before any other resource.

 How can I differ setting the weight=0 from not setting weight hook(`Weight` default value is 0)?

ashokponkumar:
@liuming-dev Thanks! Good point. While implementing I realised that hook-weight does not differentiate between weight 0 and no weight. So modelled order-weight too same way. Will modify the proposal above.

hickeyma (changes requested):
@ashokponkumar Do you mind rebasing with master? You seem to be out of sync and pulling in changes to this PR which are already merged.

ashokponkumar:
> @ashokponkumar Do you mind rebasing with master? You seem to be out of sync and pulling in changes to this PR which are already merged.

sure @hickeyma. Am new to this process. Guess I am messing things up a bit. I tried following the suggestions by DCO tool. Now am trying to follow [this](https://github.com/kubernetes/community/blob/master/contributors/guide/github-workflow.md). Do you have any document on how to pull changes after a PR is raised without adding the new commits in master being shown as part of the pull request. Am trying myself too and will try to create a new pull request.

hickeyma:
@ashokponkumar Have a look at GitHub docs on git rebase: https://docs.github.com/en/github/using-git/about-git-rebase

ashokponkumar:
> @ashokponkumar Have a look at GitHub docs on git rebase: https://docs.github.com/en/github/using-git/about-git-rebase

Thanks! Will look into it.

ashokponkumar:
Have created a new pull request #8446 . Hence closing this. @hickeyma Would much appreciate if you can have a look at #8446 and let me know if more changes are required.

PR #8443, opened 2020-07-13
Clarify comments to match practice

bridgetkromhout:
I'm updating the release documentation, and it seems there are some inaccurate comments in this file.

Based on the revision history in https://github.com/helm/helm/commits/master/internal/version/version.go it does not appear we update that file for patch releases, and it does not appear we currently use any format other than vX.y, so I am removing misleading comments.

### DIFF
--- a/internal/version/version.go
+++ b/internal/version/version.go
@@ -25,11 +25,9 @@ import (
 var (
 	// version is the current version of Helm.
 	// Update this whenever making a new release.
-	// The version is of the format Major.Minor.Patch[-Prerelease][+BuildMetadata]
 	//
 	// Increment major number for new feature additions and behavioral changes.
 	// Increment minor number for bug fixes and performance enhancements.
-	// Increment patch number for critical fixes to existing releases.
 	version = "v3.3"
 
 	// metadata is extra build time data


hickeyma (approved):
LGTM, thanks @bridgetkromhout

mattfarina (approved):
Another good catch

internal/version/version.go
```diff
@@ -25,11 +25,9 @@ import (
 var (
 	// version is the current version of Helm.
 	// Update this whenever making a new release.
-	// The version is of the format Major.Minor.Patch[-Prerelease][+BuildMetadata]
```
bacongobbler, line 28: This version does apply once a tag is built. This statement holds true.

https://github.com/helm/helm/blob/a2d11c92883509f283b9edd84a203b6e49a9580a/Makefile#L38-L41

Can you please reinstate this?

bacongobbler (changes requested):

internal/version/version.go
bacongobbler:
```
 ><> helm version
```

bridgetkromhout:
Thanks, @bacongobbler. Reinstated in https://github.com/helm/helm/pull/8452.

PR #8445, opened 2020-07-13
Correct make target in Makefile

bridgetkromhout:
When updating documentation I noticed that a typo was preventing this make target from working. This does not affect any automation around the build; it is only used in an optional step in the release checklist.


bacongobbler (approved):

mattfarina (approved):

mattfarina:
@bridgetkromhout nice catch. I've gotten that error and just typed the right thing before.

Issue #8392, opened 2020-07-03
 Bearer token auth when pull chart from private repo


lemonli:
Private repos  use `Bearer Token` rather than `Basic Auth` (eg. chartmuseum support auth with bearer token).
Like this:
```
helm repo add sample-repo --token <token>
helm install/upgrade xxx --token <token>
helm pull xxx --token <token>
```
I can give it a try if it make sense

hobbytp:
@lemonli thanks for this commit, it is very useful. 
BTW, have you considered to support **helm push xxx --token <token>**  :)

lemonli:
> @lemonli thanks for this commit, it is very useful.
> BTW, have you considered to support **helm push xxx --token **, and we still need to use "helm push" plugin to do bearer auth?

`helm push` is a plugin and the code is `out-of-tree`. I think https://github.com/chartmuseum/helm-push#token maybe useful for you.

PR #8447, opened 2020-07-14
feat: Support Bearer token auth when get chart

lemonli:
Signed-off-by: lemonli <liwenjun0323@gmail.com>

PR for #8392 

**If applicable**:

### DIFF
--- a/pkg/repo/chartrepo.go
+++ b/pkg/repo/chartrepo.go
@@ -48,6 +48,7 @@ type Entry struct {
 	KeyFile               string `json:"keyFile"`
 	CAFile                string `json:"caFile"`
 	InsecureSkipTLSverify bool   `json:"insecure_skip_tls_verify"`
+	Token                 string `json:"token"`
 }
 
 // ChartRepository represents a chart repository
@@ -57,6 +58,70 @@ type ChartRepository struct {
 	IndexFile  *IndexFile
 	Client     getter.Getter
 	CachePath  string
+}
+
+// options are generic parameters to be provided to find chart.
+type options struct {
+	chartName             string
+	chartVersion          string
+	repoURL               string
+	certFile              string
+	keyFile               string
+	caFile                string
+	username              string
+	password              string
+	token                 string
+	insecureSkipTLSverify bool
+}
+
+// Option allows specifying various settings configurable by the user for overriding the defaults
+// used when performing find chart.
+type Option func(*options)
+
+// WithChartInfo informs the chart info.
+func WithChartInfo(chartName, chartVersion string) Option {
+	return func(opts *options) {
+		opts.chartName = chartName
+		opts.chartVersion = chartVersion
+	}
+}
+
+// WithRepoURL informs the repo url that will be used when find chart.
+func WithRepoURL(url string) Option {
+	return func(opts *options) {
+		opts.repoURL = url
+	}
+}
+
+// WithBasicAuth sets the request's Authorization header to use the provided credentials
+func WithBasicAuth(username, password string) Option {
+	return func(opts *options) {
+		opts.username = username
+		opts.password = password
+	}
+}
+
+// WithTLSClientConfig sets the client auth with the provided credentials.
+func WithTLSClientConfig(certFile, keyFile, caFile string) Option {
+	return func(opts *options) {
+		opts.certFile = certFile
+		opts.keyFile = keyFile
+		opts.caFile = caFile
+	}
+}
+
+// WithBearerToken sets the request's Authorization header to use the provided token
+func WithBearerToken(token string) Option {
+	return func(opts *options) {
+		opts.token = token
+	}
+}
+
+// WithInsecureSkipVerifyTLS determines if a TLS Certificate will be checked
+func WithInsecureSkipVerifyTLS(insecureSkipTLSverify bool) Option {
+	return func(opts *options) {
+		opts.insecureSkipTLSverify = insecureSkipTLSverify
+	}
 }
 
 // NewChartRepository constructs ChartRepository
@@ -127,6 +192,7 @@ func (r *ChartRepository) DownloadIndexFile() (string, error) {
 		getter.WithInsecureSkipVerifyTLS(r.Config.InsecureSkipTLSverify),
 		getter.WithTLSClientConfig(r.Config.CertFile, r.Config.KeyFile, r.Config.CAFile),
 		getter.WithBasicAuth(r.Config.Username, r.Config.Password),
+		getter.WithBearerToken(r.Config.Token),
 	)
 	if err != nil {
 		return "", err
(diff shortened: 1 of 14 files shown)


pkg/repo/chartrepo.go
```diff
@@ -198,13 +200,13 @@ func (r *ChartRepository) generateIndex() error {
...
 func FindChartInRepoURL(repoURL, chartName, chartVersion, certFile, keyFile, caFile string, getters getter.Providers) (string, error) {
-	return FindChartInAuthRepoURL(repoURL, "", "", chartName, chartVersion, certFile, keyFile, caFile, getters)
+	return FindChartInAuthRepoURL(repoURL, "", "", chartName, chartVersion, certFile, keyFile, caFile, "", getters)
 }
 
 // FindChartInAuthRepoURL finds chart in chart repository pointed by repoURL
 // without adding repo to repositories, like FindChartInRepoURL,
 // but it also receives credentials for the chart repository.
-func FindChartInAuthRepoURL(repoURL, username, password, chartName, chartVersion, certFile, keyFile, caFile string, getters getter.Providers) (string, error) {
+func FindChartInAuthRepoURL(repoURL, username, password, chartName, chartVersion, certFile, keyFile, caFile, token string, getters getter.Providers) (string, error) {
```
mattfarina, line 209: Helm uses [semantic versioning](https://semver.org). This function is part of the public API. Changing the signature is considered a breaking change that would require incrementing the major version to Helm v4. We aren't going to do that right now.

Instead, you can have a new function for this change while preserving the old one. The old one can be marked as deprecated. #7984 is a PR that does this, as an example.

mattfarina (changes requested):
Thanks for the PR linked to the issue. Helm follows SemVer so the function signature cannot change.

pkg/repo/chartrepo.go
lemonli: Thanks, I have update my commit that follow your advise

lemonli:
@mattfarina  Would you please review again? Thanks.

pkg/repo/chartrepo.go
```diff
@@ -198,12 +200,13 @@ func (r *ChartRepository) generateIndex() error {
...
 // without adding repo to repositories
 func FindChartInRepoURL(repoURL, chartName, chartVersion, certFile, keyFile, caFile string, getters getter.Providers) (string, error) {
-	return FindChartInAuthRepoURL(repoURL, "", "", chartName, chartVersion, certFile, keyFile, caFile, getters)
+	return FindChartInRepoURLWithAuth(repoURL, "", "", chartName, chartVersion, certFile, keyFile, caFile, "", getters)
 }
 
 // FindChartInAuthRepoURL finds chart in chart repository pointed by repoURL
 // without adding repo to repositories, like FindChartInRepoURL,
 // but it also receives credentials for the chart repository.
+// Deprecated: this function is deprecated and will be removed in Helm 4, please use FindChartInRepoURLWithAuth instead.
```
jdolitsky, line 209: @mattfarina - do you think this is the best way to avoid modifying the public API? I'm not coming up with a good alternative given that this function takes other positional args such as "username" "password".

Should we take this moment to redesign the function to be more future proof? In the case there are more args? We can maybe use options to do things like `FindChartInRepoURLWithAuth(WithCaFile(...), WithToken(...) ...` instead?

jdolitsky (changes requested):

pkg/repo/chartrepo.go
lemonli: @jdolitsky I think this is a good way as you said. 
@mattfarina I can update my commit if you agree with this.

lemonli:
@jdolitsky I have updated my commit, would you please review again? Thanks.

pkg/repo/chartrepo.go
```diff
@@ -258,6 +322,68 @@ func FindChartInAuthRepoURL(repoURL, username, password, chartName, chartVersion
...
+		URL:      repoOpts.repoURL,
+		Username: repoOpts.username,
+		Password: repoOpts.password,
+		CertFile: repoOpts.certFile,
+		KeyFile:  repoOpts.keyFile,
+		CAFile:   repoOpts.caFile,
+		Name:     name,
+		Token:    repoOpts.token,
+	}
+	r, err := NewChartRepository(&c, getters)
```
jdolitsky, line 349: Since most of this (besides relating to token) is reused from FindChartInAuthRepoURL, can we move it into a private helper function such as `findChartInAuthRepoURL(token)` so that the file has overall less lines of code?

jdolitsky (changes requested):
Sorry for delay. Looks pretty good! just one comment

pkg/repo/chartrepo.go
lemonli: I have updated my commit, would you please review again? Thanks.

jdolitsky (dismissed):
LGTM

bacongobbler:
needs a rebase before it can be merged.

lemonli:
> needs a rebase before it can be merged.

done

mattfarina (changes requested):
I wanted to share the hold up I'm having with this PR.

Someone should not pass the bearer token in via the command because it will get caught up in the shell history. This is a security issue. Avoiding this with passwords is why Helm has the ability to read it from stdin.

There is also the case that the dearer token changes. Storing the bearer token so that it's used 6 months or a year from now doesn't seem like a good idea. It should expire. How does someone update that?

Does anyone have thoughts on these issues?

marcosnils:
> Someone should not pass the bearer token in via the command because it will get caught up in the shell history. This is a security issue. Avoiding this with passwords is why Helm has the ability to read it from stdin.

You can always set the token in an env var and pass it with --token $MYTOKEN. Mayabe eventually having a --token-stdin like `docker login` has could be a good idea, but I IMHO doesn't seem to be such a big of a concern that needs this PR to be held.

In addition to my previous comment, this also happens with `username` and `password` correct? Doesn't seem to be any difference there.


> There is also the case that the dearer token changes. Storing the bearer token so that it's used 6 months or a year from now doesn't seem like a good idea. It should expire. How does someone update that?

Not sure I follow. Updating and rotating the token should be the user's  concern? I guess the same happens with username/passowrd also right?

mattfarina:
> > Someone should not pass the bearer token in via the command because it will get caught up in the shell history. This is a security issue. Avoiding this with passwords is why Helm has the ability to read it from stdin.
> 
> You can always set the token in an env var and pass it with --token $MYTOKEN. Mayabe eventually having a --token-stdin like docker login has could be a good idea, but I IMHO doesn't seem to be such a big of a concern that needs this PR to be held.

The Helm project is rounding out a second security review that included things like a threat analysis. The details will be released soon. These types of things were a topic of conversation in that process. This is the kind of security some ask of Helm.

When you specify a username and no password, Helm will prompt you enter it via stdin. This way it's not caught in the shell history.

The usage of stdin matters. For this example, I'll use CIrcleCI. CircleCI has something called [secrets masking](https://circleci.com/docs/2.0/env-vars/#secrets-masking) where the value of an environment variable is masked when used with print and echo. That means it won't, for example, be masked if you use it with the `--password` flag on the `docker login` command. CircleCI isn't the only CI system that does things like this.

A password could be inadvertently exposed through a CI system. This is something we want to take seriously on Helm.

In the name of security, we need a stdin based method to accept a token. That way you can do something like...

```
$ echo $TOKEN | helm repo add foo https://foo.example.com --token-stdin
```

This security issue is my primary hangup.

marcosnils:
> In the name of security, we need a stdin based method to accept a token. That way you can do something like...

I'm not going to go against something that improves the security of any project. If we can add this faeture also it'd be awesome. Not sure what's @lemonli's bandwidth to do it.

hobbytp:
Any update here? it is useful feature, can we separate the submit to more steps, first with --token $MYTOKEN, and 2nd step to add "--token-stdin"?

josedev-union:
I hope @jdolitsky can have a chance to make a workaround for --token-stdin.

lemonli:
Sorry for delay. I will update my commit to support `--token-stdin` soon.

lemonli:
> > > Someone should not pass the bearer token in via the command because it will get caught up in the shell history. This is a security issue. Avoiding this with passwords is why Helm has the ability to read it from stdin.
> > 
> > 
> > You can always set the token in an env var and pass it with --token $MYTOKEN. Mayabe eventually having a --token-stdin like docker login has could be a good idea, but I IMHO doesn't seem to be such a big of a concern that needs this PR to be held.
> 
> The Helm project is rounding out a second security review that included things like a threat analysis. The details will be released soon. These types of things were a topic of conversation in that process. This is the kind of security some ask of Helm.
> 
> When you specify a username and no password, Helm will prompt you enter it via stdin. This way it's not caught in the shell history.
> 
> The usage of stdin matters. For this example, I'll use CIrcleCI. CircleCI has something called [secrets masking](https://circleci.com/docs/2.0/env-vars/#secrets-masking) where the value of an environment variable is masked when used with print and echo. That means it won't, for example, be masked if you use it with the `--password` flag on the `docker login` command. CircleCI isn't the only CI system that does things like this.
> 
> A password could be inadvertently exposed through a CI system. This is something we want to take seriously on Helm.
> 
> In the name of security, we need a stdin based method to accept a token. That way you can do something like...
> 
> ```
> $ echo $TOKEN | helm repo add foo https://foo.example.com --token-stdin
> ```
> 
> This security issue is my primary hangup.

@mattfarina I have updated my commit, looking forward to your reply.

jrgwv:
Hi folks, any update on when this might become available?

ztbrown:
Any update on this?

cuttingedge1109:
@lemonli seems need to rebase again to merge

ztbrown:
@lemonli I have rebased your branch locally. Let me know if you'd like me to PR that to your repo.

lemonli:
> @lemonli I have rebased your branch locally. Let me know if you'd like me to PR that to your repo.

Sure, thanks @ztbrown

ztbrown:
https://github.com/lemonli/helm/pull/1

gimler:
what is the state of this pr?

almed4:
Can we get an update? Really looking forward to this feature.

koalalorenzo:
Do we have any update? I know it might be redundant but I would love to get this merged! what can we do to help getting this merged? @mattfarina @jdolitsky any idea?

jverce:
@mattfarina I think this is pending the resolution of a change you requested that's already been addressed, no?
If there's no more changes needed, can we have this merged?

PS: @lemonli this needs a rebase (it's been a while 😞)

ztbrown:
https://github.com/helm/helm/pull/10848

Someone check to see if I did the rebase correctly?

marcosnils:
dog-reading

juldou:
can someone check the rebase? https://github.com/helm/helm/pull/10848

bmiag:
any news on this?

edijsdrezovs:
Any news?

PR #8448, opened 2020-07-14
add support for weights for resources

### LINKED ISSUE #8439: Openshift SCC should be installed before any resource which creates a pod

### LINKED ISSUE #8446: Add support for ordering of resources within a chart for Custom Resources
Similar to how we have an ordering requirement for known resources in the [InstallOrder](https://github.com/helm/helm/blob/release-3.0/pkg/releaseutil/kind_sorter.go), users want the capability to order custom resources too. For example a custom resource like SCC need to be applied before creation of any pods as discussed in #8439 

To enable this capability we can follow the same approach we use for hooks. Essentially add a annotation "helm.sh/order-weight" which can take an integer value (+ve or -ve). This will behave the same way as "helm.sh/hook-weight", but for resources not covered by hooks. 

Here is how it will work: 
1. The first resources to be installed will be all resources that have negative weights, and then those with zero or no weights, and then those with positive weights. All in ascending order.
1. For resources which have same weight, Install order is preserved for known resources.

ashokponkumar:
closes #8439 

Signed-off-by: Ashok Pon Kumar <ashokponkumar@in.ibm.com>

**If applicable**:

### DIFF
--- a/pkg/action/resource_policy.go
+++ b/pkg/action/resource_policy.go
@@ -31,7 +31,7 @@ const resourcePolicyAnno = "helm.sh/resource-policy"
 //   during an uninstallRelease action.
 const keepPolicy = "keep"
 
-func filterManifestsToKeep(manifests []releaseutil.Manifest) (keep, remaining []releaseutil.Manifest) {
+func filterManifestsToKeep(manifests []*releaseutil.Manifest) (keep, remaining []*releaseutil.Manifest) {
 	for _, m := range manifests {
 		if m.Head.Metadata == nil || m.Head.Metadata.Annotations == nil || len(m.Head.Metadata.Annotations) == 0 {
 			remaining = append(remaining, m)
--- a/pkg/releaseutil/kind_sorter.go
+++ b/pkg/releaseutil/kind_sorter.go
@@ -108,12 +108,13 @@ var UninstallOrder KindSortOrder = []string{
 // sort manifests by kind.
 //
 // Results are sorted by 'ordering', keeping order of items with equal kind/priority
-func sortManifestsByKind(manifests []Manifest, ordering KindSortOrder) []Manifest {
-	sort.SliceStable(manifests, func(i, j int) bool {
-		return lessByKind(manifests[i], manifests[j], manifests[i].Head.Kind, manifests[j].Head.Kind, ordering)
+func sortManifestsByKind(manifests []*Manifest, ordering KindSortOrder) []*Manifest {
+	m := manifests
+	sort.SliceStable(m, func(i, j int) bool {
+		return lessByKind(m[i], m[j], m[i].Head.Kind, m[j].Head.Kind, ordering)
 	})
 
-	return manifests
+	return m
 }
 
 // sort hooks by kind, using an out-of-place sort to preserve the input parameters.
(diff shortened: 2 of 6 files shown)


ashokponkumar:
**Implemented solution:** Similar to how we have an ordering requirement for known resources in the InstallOrder, users want the capability to order custom resources too. For example a custom resource like SCC need to be applied before creation of any pods.

To enable this capability we can follow the same approach we use for hooks. Essentially add a annotation "helm.sh/order-weight" which can take an integer value (+ve or -ve). This will behave the same way as "helm.sh/hook-weight", but for resources not covered by hooks.

Here is how it will work:

The first resources to be installed will be all resources that have negative weights, and then those with zero or no weights, and then those with positive weights. All in ascending order.
For resources which have same weight, Install order is preserved for known resources.

----

**Chart weights:**
To address the question raised in today's developer call around dependent charts, one way to handle it could be to have a chart weight specifiable along with the dependency specification in chart.yaml. The value specified in it will get added as a "helm.io/chart-weight" annotation in each of those charts resources, when the resources are templated. In case of dependency within a dependency, the "helm.io/chart-weight" will be the sum of the all chart-weights encountered till now.

And while each resource order-weight is computed, the chart-weight gets added to its value. For the base chart, the value is 0.

hickeyma:
@ashokponkumar It might be worth setting this to draft until you iron/flesh out the solution more?

ashokponkumar:
> @ashokponkumar It might be worth setting this to draft until you iron/flesh out the solution more?

sure. Done.

bacongobbler:
Checking in here to see what’s the status on this. This was opened up in July, and it looks like there is some interest from other community members for this feature.

ashokponkumar:
> Checking in here to see what’s the status on this. This was opened up in July, and it looks like there is some interest from other community members for this feature.

@bacongobbler If we can settle on the right technique, I can try to take a stab at implementing it. [Here](https://github.com/helm/helm/issues/8439#issuecomment-659552690) was the last discussion based on the discussion in the developer call.

priyavj08:
hi, what is the status on this?

gonzaloserrano:
👀

jvossler-cogility:
Any update on this effort. I have this issue with deploying gatekeeper via helm. The custom helm chart includes a number of constraint templates and constraints. The templates deploy correctly, but the constraints will not deploy as the templates are not available to all three manager nodes. I need to get the templates fully installed before attempting to reference them with security constraints.

PR #8460, opened 2020-07-15
Support apiextensions.k8s.io/v1

### LINKED ISSUE #6783: crd-install apiextensions.k8s.io/v1 does not work

jimmidyson:

Adds `apiextensions.k8s.io/v1` support. Fixes #6783.

**If applicable**:

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -39,6 +39,7 @@ import (
 	batch "k8s.io/api/batch/v1"
 	v1 "k8s.io/api/core/v1"
 	extv1beta1 "k8s.io/api/extensions/v1beta1"
+	apiextv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apiextv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -81,16 +82,24 @@ type Client struct {
 	Log func(string, ...interface{})
 }
 
+var addToScheme sync.Once
+
 // New creates a new Client.
 func New(getter genericclioptions.RESTClientGetter) *Client {
 	if getter == nil {
 		getter = genericclioptions.NewConfigFlags(true)
 	}
 
-	err := apiextv1beta1.AddToScheme(scheme.Scheme)
-	if err != nil {
-		panic(err)
-	}
+	// Add CRDs to the scheme. They are missing by default.
+	addToScheme.Do(func() {
+		if err := apiextv1.AddToScheme(scheme.Scheme); err != nil {
+			// This should never happen.
+			panic(err)
+		}
+		if err := apiextv1beta1.AddToScheme(scheme.Scheme); err != nil {
+			panic(err)
+		}
+	})
 
 	return &Client{
 		Factory: cmdutil.NewFactory(getter),
@@ -385,7 +394,7 @@ func (c *Client) Get(namespace string, reader io.Reader) (string, error) {
 			return nil
 		}
 
-		//Get the relation pods
+		// Get the relation pods
 		objs, err = c.getSelectRelationPod(info, objs)
 		if err != nil {
 			c.Log("Warning: get the relation pod is failed, err:%s", err.Error())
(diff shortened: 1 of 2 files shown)


pkg/kube/client.go
```diff
@@ -739,20 +748,20 @@ func (c *Client) pollCRDUntilEstablished(timeout time.Duration, info *resource.I
...
-		crd := &apiextv1beta1.CustomResourceDefinition{}
+		crd := &apiextv1.CustomResourceDefinition{}
 		err = scheme.Scheme.Convert(info.Object, crd, nil)
 		if err != nil {
 			return false, fmt.Errorf("unable to convert to CRD type: %v", err)
 		}
 
 		for _, cond := range crd.Status.Conditions {
 			switch cond.Type {
-			case apiextv1beta1.Established:
```
bacongobbler, line 750: Won't this break users who are still using the apiextensions/v1beta1 API? As far as I can remember, the business logic behind the two are slightly different, which is why we have `crdReady` and `crdBetaReady` in Helm 3. If so, this could be considered a breaking change for those users, as Helm will no longer wait for apiextensions/v1beta1 to become established.

Helm 3 handles both APIs. See here: https://github.com/helm/helm/blob/e31818f2efd3e1e0867d9c189507c25c8102264a/pkg/kube/wait.go#L111-L132

jimmidyson: Interesting, yeah I think you're right. We might be able to convert to `v1` from `v1beta1` via the internal type (there are conversions to/from `k8s.io/apiextensions-apiserver/pkg/apis/apiextensions` for both `v1` and `v1beta1`)

jimmidyson: I've refactored this to use unstructured directly instead. This seems to work nicely and supports both v1beta1 and v1.

jimmidyson: @bacongobbler Any thoughts on this approach?

jdolitsky (approved):

pkg/kube/client.go
jimmidyson: @bacongobbler Sorry to pester you but would you mind re-reviewing please?

pkg/kube/client.go
```diff
@@ -739,20 +748,38 @@ func (c *Client) pollCRDUntilEstablished(timeout time.Duration, info *resource.I
...
-		crd := &apiextv1beta1.CustomResourceDefinition{}
-		err = scheme.Scheme.Convert(info.Object, crd, nil)
-		if err != nil {
-			return false, fmt.Errorf("unable to convert to CRD type: %v", err)
+		crd, ok := info.Object.(*unstructured.Unstructured)
+		if !ok {
+			return false, fmt.Errorf("unable to cast CRD to unstructured.Unstructured type: %v", err)
+		}
+
+		conditions, ok, err := unstructured.NestedSlice(crd.UnstructuredContent(), "status", "conditions")
```
bacongobbler, line 756: Why are we casting to an unstructured object, then trying to determine if `.status.conditions` is part of the object? Why can't we do the same conversion logic as is available in Helm 3?

In other words..


```
switch value := AsVersioned(info.Object).(type) {
case *apiextv1beta1.CustomResourceDefinition:
    ...
case *apiextv1.CustomResourceDefinition:
    ...
}
```

There may be other object types that also report their status via the `.status.conditions` field. As I understand it, converting this object to an unstructured type means we're no longer checking the object's type, and could potentially trigger wait logic on a non-CRD type (like a Deployment).

bacongobbler (changes requested):

bridgetkromhout:
Hi, @jimmidyson - can you take a look at the feedback and revise this PR? We are nearly at the end of the change window for v2, so we really need to get this in in the next few days if at all. Thanks.

jimmidyson:
@bridgetkromhout @bacongobbler I'm not going to have time in the next few days to fix this up so unless someone else picks it up, I guess it will have to miss the v2 change window. I'm moving to helm v3 ASAP - which is a good thing :)

bacongobbler:
We are no longer accepting bug fixes for Helm 2. See https://helm.sh/blog/helm-v2-deprecation-timeline/ for more details.

PR #8462, opened 2020-07-16
fix(create): update the hook name of test-connection pod

### LINKED ISSUE #8457: Update test hook for created chart
When `helm create` is used to create a new chart it creates a test, by default. The test hook is set to `test-success`. This is for legacy purposes as this was the hook in Helm v2. The chart created by Helm v3 is set to `apiVersion: v2` but uses the legacy hook. Internally to the Helm code the legacy hook name is an alias to the new name.

At some point we should update the hook name to `test` in the scaffolded chart.

donggangcj:
close #8457


mattfarina (approved):
lgtm

Issue #8458, opened 2020-07-15
storage History function has inconsistent behaviours depending on the driver


mikeshng:
The storage `History` function states `returns ErrReleaseNotFound if no such release name exists.`
https://github.com/helm/helm/blob/master/pkg/storage/storage.go#L150

For secret and cfgmap, it's possible to have an empty return because there is no length check.
https://github.com/helm/helm/blob/master/pkg/storage/driver/secrets.go#L129-L138
https://github.com/helm/helm/blob/master/pkg/storage/driver/cfgmaps.go#L138-L147

For memory and sql, it's not possible to return an empty due to a length check.
https://github.com/helm/helm/blob/master/pkg/storage/driver/memory.go#L145-L146
https://github.com/helm/helm/blob/master/pkg/storage/driver/sql.go#L313-L325

This seems to be inconsistent behaviour. Is this intentional? Thanks.

Output of `helm version`:
```
version.BuildInfo{Version:"v3.1.2", GitCommit:"d878d4d45863e42fd5cff6743294a11d28a9abce", GitTreeState:"clean", GoVersion:"go1.13.8"}
```
Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.2", GitCommit:"52c56ce7a8272c798dbc29846288d7cd9fbae032", GitTreeState:"clean", BuildDate:"2020-04-16T11:56:40Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.4", GitCommit:"224be7bdce5a9dd0c2fd0d46b83865648e2fe0ba", GitTreeState:"clean", BuildDate:"2020-01-14T01:25:43Z", GoVersion:"go1.12.12", Compiler:"gc", Platform:"linux/amd64"}
```

Cloud Provider/Platform (AKS, GKE, Minikube etc.): 
`kind`

bacongobbler:
Hi @mikeshng. Please have a look at the PR queue. https://github.com/helm/helm/pull/7372 appears to address this case for the memory driver, but perhaps there's some feedback the author needs here with regards to the postgresql driver. Please check and confirm if this PR will fix this issue for you. Thanks.

bacongobbler:
[...]

PR #8470, opened 2020-07-16
fix(sql storage): Query() should return ErrReleaseNotFound immediately when no records are found

mikeshng:
Signed-off-by: Mike Ng <ming@redhat.com>

There is an inconsistency (and maybe wrong) of how `ErrReleaseNotFound` is being returned in the sql storage driver Query() function.
In both secrets and cfgmap drivers,
https://github.com/helm/helm/blob/master/pkg/storage/driver/secrets.go#L125-L138
https://github.com/helm/helm/blob/master/pkg/storage/driver/cfgmaps.go#L134-L147
the number of items are checked first and if it's zero return `ErrReleaseNotFound`. But that's not the case for sql driver.
It's possible for the sql driver to failed to decode all the releases and return the error  `ErrReleaseNotFound` but that's misleading and not true because the release exists just failed to decode.

This change is to sql driver Query() consistent with the other drivers.

related https://github.com/helm/helm/issues/8458 and https://github.com/helm/helm/pull/7372

**If applicable**:
- [NA] this PR contains documentation

### DIFF
--- a/pkg/storage/driver/sql.go
+++ b/pkg/storage/driver/sql.go
@@ -310,6 +310,10 @@ func (s *SQL) Query(labels map[string]string) ([]*rspb.Release, error) {
 		return nil, err
 	}
 
+	if len(records) == 0 {
+		return nil, ErrReleaseNotFound
+	}
+
 	var releases []*rspb.Release
 	for _, record := range records {
 		release, err := decodeRelease(record.Body)
@@ -318,10 +322,6 @@ func (s *SQL) Query(labels map[string]string) ([]*rspb.Release, error) {
 			continue
 		}
 		releases = append(releases, release)
-	}
-
-	if len(releases) == 0 {
-		return nil, ErrReleaseNotFound
 	}
 
 	return releases, nil
--- a/pkg/storage/driver/sql.go
+++ b/pkg/storage/driver/sql.go
@@ -324,6 +324,10 @@ func (s *SQL) Query(labels map[string]string) ([]*rspb.Release, error) {
 		releases = append(releases, release)
 	}
 
+	if len(releases) == 0 {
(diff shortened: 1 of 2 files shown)


hickeyma (commented):
@mikeshng Feedback in https://github.com/helm/helm/issues/8458#issuecomment-664969792 to help progress this.

pkg/storage/driver/sql.go
```diff
@@ -320,10 +324,6 @@ func (s *SQL) Query(labels map[string]string) ([]*rspb.Release, error) {
 		releases = append(releases, release)
 	}
 
-	if len(releases) == 0 {
-		return nil, ErrReleaseNotFound
-	}
-
```
absoludity, line 326: It might be better to leave this for now, so as not to change the behaviour of an error being raised when there were releases but they couldn't be parsed (it might make it easier / clearer to get a subsequent PR approved which updates this to a different error).

absoludity (approved):

pkg/storage/driver/sql.go
mikeshng: I agree. I shouldn't have break the current behaviour. @hickeyma what do you think? Thanks.

bacongobbler: Yes. We cannot change existing behavior unless it's unintended behavior (bug) or a security issue.

mikeshng: @bacongobbler I agree. I have reverted that now. This PR now only contains the fix of return ErrReleaseNotFound immediately when no records are found so that it's consistent with the other drivers.

bacongobbler (approved):

hickeyma (approved):
LGTM, thanks @mikeshng

Issue #8464, opened 2020-07-16
helm chart pull/push should support --plain-http and --insecure option


youyongsong:
Output of `helm version`: 3.0.0-rc1

helm only support `--insecure` option in the `helm registry login` command. In some cases we want to push or pull a chart to a insecure registry **anonymously**, currently when we use `helm chart pull` or `helm chart push` to a insecure registry it will return the following error:

```
$ helm chart push 192.168.3.6:5000/ysyou/cert-manager:v2.13
The push refers to repository [192.168.3.6:5000/ysyou/cert-manager]
ref:     192.168.3.6:5000/ysyou/cert-manager:v2.13
digest:  c9d19b41e1e28d59d1da8ccc0dc9078341f46afb300b24f1c9709cb6d790bcf0
size:    57.9 KiB
name:    cert-manager
version: v2.0-b.1
Error: failed to do request: Head "https://192.168.3.6:5000/v2/ysyou/cert-manager/blobs/sha256:6a9498d039faadfaa2e6bec408865a1a83d1f952f85fe431407964c10e9cad18": http: server gave HTTP response to HTTPS clien
```

Hope helm can add `--plain-http` and `--insecure` options in `helm chart push` and `helm chart pull` commands.

hickeyma:
@jdolitsky Is this something that would be considered?

bacongobbler:
Duplicate of https://github.com/helm/helm/issues/6324.

PR #8477, opened 2020-07-17
support insecure and plain-http options in helm chart pull and push c…

### LINKED ISSUE #6324: How can I push chart to an insecure OCI registry with helm v3

youyongsong:
This pr tries to resolve issue #8464 and #6324.
helm only support `--insecure` option in the helm registry login command. In some cases we want to push or pull a chart to a insecure or http registry anonymously without login, so I add `--insecure` and `--plain-http` options to meet these cases.

close #6324

Signed-off-by: youyongsong <youyongsong@gmail.com>

### DIFF
--- a/cmd/helm/chart_pull.go
+++ b/cmd/helm/chart_pull.go
@@ -32,7 +32,8 @@ This will store the chart in the local registry cache to be used later.
 `
 
 func newChartPullCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
-	return &cobra.Command{
+	var insecureOpt, plainHTTPOpt bool
+	cmd := &cobra.Command{
 		Use:    "pull [ref]",
 		Short:  "pull a chart from remote",
 		Long:   chartPullDesc,
@@ -40,7 +41,13 @@ func newChartPullCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 		Hidden: !FeatureGateOCI.IsEnabled(),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			ref := args[0]
-			return action.NewChartPull(cfg).Run(out, ref)
+			return action.NewChartPull(cfg).Run(out, ref, insecureOpt, plainHTTPOpt)
 		},
 	}
+
+	f := cmd.Flags()
+	f.BoolVarP(&insecureOpt, "insecure", "", false, "allow connections to TLS registry without certs")
+	f.BoolVarP(&plainHTTPOpt, "plain-http", "", false, "use plain http and not https")
+
+	return cmd
 }
--- a/internal/experimental/registry/client.go
+++ b/internal/experimental/registry/client.go
@@ -18,12 +18,14 @@ package registry // import "helm.sh/helm/v3/internal/experimental/registry"
 
 import (
 	"context"
+	"crypto/tls"
 	"fmt"
 	"io"
 	"io/ioutil"
 	"net/http"
 	"sort"
 
+	"github.com/containerd/containerd/remotes"
 	auth "github.com/deislabs/oras/pkg/auth/docker"
 	"github.com/deislabs/oras/pkg/oras"
 	"github.com/gosuri/uitable"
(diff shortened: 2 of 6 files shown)


youyongsong:
hi @jdolitsky could you help review this pr ?

internal/experimental/registry/client.go
```diff
@@ -117,7 +119,17 @@ func (c *Client) Logout(hostname string) error {
 }
 
 // PushChart uploads a chart to a registry
-func (c *Client) PushChart(ref *Reference) error {
+func (c *Client) PushChart(ref *Reference, insecure bool, plainHTTP bool) error {
+	if insecure || plainHTTP {
```
pytimer, line 123: The if condition maybe not necessary? Because `c.newResolver` have check `insecure` true or false.

youyongsong: In line 79: `resolver, err := client.newResolver(false, false)` set with default resolver with `insecure=false` and `plainHTTP=false`. So i just want to replace the default resolver when `insecure = true` or `plainHTTP = true`, and still use the default resolver in other cases.

cmd/helm/chart_pull.go
```diff
@@ -32,15 +32,22 @@ This will store the chart in the local registry cache to be used later.
...
+	cmd := &cobra.Command{
 		Use:    "pull [ref]",
 		Short:  "pull a chart from remote",
 		Long:   chartPullDesc,
 		Args:   require.MinimumNArgs(1),
 		Hidden: !FeatureGateOCI.IsEnabled(),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			ref := args[0]
-			return action.NewChartPull(cfg).Run(out, ref)
+			return action.NewChartPull(cfg).Run(out, ref, insecureOpt, plainHTTPOpt)
```
mattfarina, line 44: The `Run` function is part of the public API and its signature cannot change. A new function needs to be created. The old one should be listed as deprecated and there should be a note on the new one to clean this up in Helm v4.

This does not apply to internal functions as they are not part of the public API.

cmd/helm/chart_pull.go
```diff
@@ -32,15 +32,22 @@ This will store the chart in the local registry cache to be used later.
...
 		},
 	}
+
+	f := cmd.Flags()
+	f.BoolVarP(&insecureOpt, "insecure", "", false, "allow connections to TLS registry without certs")
```
mattfarina, line 49: Should this use the same insecure flag name and description used elsewhere for skipping the verification?

mattfarina (changes requested):
Thanks for the pull request. This section is going to be going through numerous changes in the near future. @jdolitsky should have a look at it, too.

Foxboron:
Reminder that https://github.com/helm/helm/pull/7844 still exists. The problem has been that the test suite doesn't validate correctly when testing the new feature and oras has changed their APIs for TLS validation which isn't exposed by helm.

jdolitsky:
The ORAS APIs should be redesigned to use options such as `WithPlainHTTP` instead of positional arguments.. that will make this a bit cleaner. Just opened https://github.com/deislabs/oras/issues/176

bhabermaas:
Is this issue going to get resolved? I am running the latest CLI and still unable to push/pull to a HTTP registry.
version.BuildInfo{Version:"v3.3.4", GitCommit:"a61ce5633af99708171414353ed49547cf05013d", GitTreeState:"clean", GoVersion:"go1.14.9"}

bacongobbler:
hi @bhabermaas. It does not look like the original submitter is responding to feedback. If you would like to continue moving this initiative forward, please feel free to contribute.

chenwng:
Hi @bacongobbler. What is the proper way to continue this initiative? Should another PR be created with fixes to the review comments and commit from the original author? Please shed some light.

bacongobbler:
Generally yes.

bacongobbler:
I'm going to close this ticket due to inactivity, but please re-open if this still needs to be addressed. Thanks!

PR #8478, opened 2020-07-17
Lockfiles without version ranges

### LINKED ISSUE #8449: Dependencies: Resolve version when only URL is known
Currently, when is resolving dependencies and a dependency is specified as a URL but it has not been added to Helm via `helm repo add` the dependency is locked to the exact value specified for the requested dependency. If the specified value is a range (e.g., `^1.2.3) the range is the version it is locked to.

The current situation creates a situation where locking isn't really locked to a version.

With more distributed repos that cross link each other this problem has the potential to get worse.

Instead of locking to the range, Helm should work out the version and lock to the version.

I would suggest the steps:

1. If not cached locally via `helm repo add`, create an automated and not-visible to the user cache.
2. Resolve the version from this invisible cache
3. Lock to the version
4. Clean up any invisible caches (periodically?)

mattfarina:
There were two situations where a lockfile could end up with a version range as a version instead of a pinned version.

1. When `file://` was used. This would point to the file location on the filesystem but, instead of using the version in that chart at that file location the version range would be used.
2. If a repository was used that was not added with `helm repo add`. Instead of causing an error that the repository was not available the version range was used in the lockfile. This is not locking.

This change looks to correct these problems. The `file://` handling fix was fairly straight forward. In that case update gets the current version of the dependency on disk and locks to that.

Handling unknown repos went through 4 different major versions. This 4th one is the one that settled on because all the changes are in the download manager which handles network issues. Other cases had added network operations to the resolver which is meant to iterate over existing versions. I chose this as the network and downloading stay in the same struct and it's methods are largely reusable within the new logic.

Closes #8449 

**If applicable**:

### DIFF
--- a/pkg/downloader/manager.go
+++ b/pkg/downloader/manager.go
@@ -171,6 +171,10 @@ func (m *Manager) Update() error {
 	// For the repositories Helm is not configured to know about, ensure Helm
 	// has some information about them and, when possible, the index files
 	// locally.
+	// TODO(mattfarina): Repositories should be explicitly added by end users
+	// rather than automattic. In Helm v4 require users to add repositories. They
+	// should have to add them in order to make sure they are aware of the
+	// respoitories and opt-in to any locations. For security.
 	repoNames, err = m.ensureMissingRepos(repoNames, req)
 	if err != nil {
 		return err
--- a/pkg/downloader/manager.go
+++ b/pkg/downloader/manager.go
@@ -108,13 +108,7 @@ func (m *Manager) Build() error {
 		}
 	}
 
-	repoNames, err := m.resolveRepoNames(req)
-	if err != nil {
-		return err
-	}
-
-	_, err = m.ensureMissingRepos(repoNames, req)
-	if err != nil {
+	if _, err := m.resolveRepoNames(req); err != nil {
 		return err
 	}
 
@@ -130,6 +124,11 @@ func (m *Manager) Build() error {
 		} else {
 			return errors.New("the lock file (Chart.lock) is out of sync with the dependencies file (Chart.yaml). Please update the dependencies")
 		}
+	}
+
+	// Check that all of the repos we're dependent on actually exist.
+	if err := m.hasAllRepos(lock.Dependencies); err != nil {
+		return err
 	}
 
 	if !m.SkipUpdate {
@@ -174,7 +173,7 @@ func (m *Manager) Update() error {
 	// TODO(mattfarina): Repositories should be explicitly added by end users
 	// rather than automattic. In Helm v4 require users to add repositories. They
 	// should have to add them in order to make sure they are aware of the
-	// respoitories and opt-in to any locations. For security.
+	// respoitories and opt-in to any locations, for security.
 	repoNames, err = m.ensureMissingRepos(repoNames, req)
 	if err != nil {
 		return err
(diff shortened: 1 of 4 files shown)


mattfarina:
While this is technically a bug fix it does change some behavior when `helm dep build` is used. So, I'm putting it into v3.4.0.

mattfarina:
Note, this doesn't change the download mechanism so it will be compatible with lockfiles that had a version range slip in.

technosophos:
Say a chart dependency references repo `https://foo.example.com/stable`. My understanding is that in the current code, it will add a repository named something like `helm-manager-a576fe4...`. So if a user does a `helm repo list`, they'll see a really weirdly named repo.

Could we go for a more clear name, like naming it `foo.example.com/stable` or `foo-example-com-stable`? That might make it a little easier on the user.

mattfarina:
@technosophos the repo being added is just to the cache and the internals during update/build. The repository configuration isn't updated so `helm repo list` won't have the unmanaged one. The reason for the hash as a key name is so that the cache is highly unlikely to have collisions.

technosophos (approved):
LGTM. Also paired with @mattfarina to ask questions, and I believe this is the best path forward for solving several related problems without breaking compatibility.

pkg/downloader/manager.go
```diff
@@ -147,14 +150,27 @@ func (m *Manager) Update() error {
...
 	if err != nil {
 		return err
 	}
 
-	// For each repo in the file, update the cached copy of that repo
+	// For the repositories Helm is not configured to know about, ensure Helm
+	// has some information about them and, when possible, the index files
+	// locally.
+	// TODO(mattfarina): Repositories should be explicitly added by end users
+	// rather than automattic. In Helm v4 require users to add repositories. They
```
jdolitsky, line 164: typos in comment (automattic), "For security" vs. ", for securirty"

pkg/downloader/manager.go
```diff
@@ -525,15 +565,27 @@ func (m *Manager) parallelRepoUpdate(repos []*repo.Entry) error {
 		wg.Add(1)
 		go func(r *repo.ChartRepository) {
 			if _, err := r.DownloadIndexFile(); err != nil {
-				fmt.Fprintf(m.Out, "...Unable to get an update from the %q chart repository (%s):\n\t%s\n", r.Config.Name, r.Config.URL, err)
+				// For those dependencies that are not known to helm and using a
+				// generated key name we display the repo url.
+				if strings.HasPrefix(r.Config.Name, "helm-manager-") {
```
jdolitsky, line 570: Should we make "helm-manager-" a constant, as its used here 3+ times?

jdolitsky (changes requested):

pkg/downloader/manager.go
mattfarina: Fixed... did you notice you misspelled "securirty" ;-). I find it amusing that a typo fix includes a typo.

pkg/downloader/manager.go
mattfarina: done

mattfarina:
@technosophos @jdolitsky I updated the PR per Josh's feedback and to undo a change I made to Build. Build keeps its behavior.

mattfarina:
Also, rebased off master due to a change landing there that impacted the same code this does.

pkg/downloader/manager.go
jdolitsky: hahaha

jdolitsky (approved):

technosophos (approved):

Issue #8450, opened 2020-07-14
helm uninstall: a flag for retain the release history when uninstall error(s) has occurred


mikeshng:
It's hard to reproduce this issue but looking at the code:
https://github.com/helm/helm/blob/v3.2.4/pkg/action/uninstall.go#L112-L137

It seems like even if `deleteRelease()` had errors, then unless I use `--keep-history ` flag the release will always get purge.
So sometimes there are resources left behind but when I do `helm ls` the release is not there anymore.

This makes scripting `helm uninstall` difficult because the desire behaviour is to purge the history, but at the same time I want to know when some resources are not cleaned up properly, so I can issue another `helm uninstall`.

Is it possible to change the default behaviour to keep the release if there is an error or have another flag like `--keep-history-if-error`?

Output of `helm version`:
```
version.BuildInfo{Version:"v3.1.2", GitCommit:"d878d4d45863e42fd5cff6743294a11d28a9abce", GitTreeState:"clean", GoVersion:"go1.13.8"}
```
Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.2", GitCommit:"52c56ce7a8272c798dbc29846288d7cd9fbae032", GitTreeState:"clean", BuildDate:"2020-04-16T11:56:40Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.4", GitCommit:"224be7bdce5a9dd0c2fd0d46b83865648e2fe0ba", GitTreeState:"clean", BuildDate:"2020-01-14T01:25:43Z", GoVersion:"go1.12.12", Compiler:"gc", Platform:"linux/amd64"}
```

Cloud Provider/Platform (AKS, GKE, Minikube etc.): 
`kind`

Cweiping:
Without changing the existing logic, I split the error types: 

1. there is an error in generating the list of resources to be deleted, and the operation fails; 
2. in the process of deleting resources, an error occurs in the upstream server, and the operation is successful, but it will prompt the user that some errors occur in the process of deleting, which will not affect the deletion of release information;

If we consider the deletion of release and resource deletion as atomic operations, it will be rolled back if one of the commands fails. We can't recover the service when half of the resource is deleted.
To sum up, I don't think that using the flag to keep history at this time will not improve the user experience.

PR #8481, opened 2020-07-18
fix uninstall cmd error, retain the release history when uninstall error(s) has occurred

Cweiping:
Fix Issue : https://github.com/helm/helm/issues/8450

fix uninstall cmd error:
1. build delete resources error should be uninstall failed. For Example, when `BuildDeleteResources()` throw error ，output：
```
Error: uninstall: Failed to build delete resources for release: unable to build kubernetes objects for delete

```
2. clarify when delete resources error from upstream not affect the success of the operation.For Example, when `deleteResources()` throw error ，output：
```
WARNING: delete resources with 1 error(s): resource not found. This will not affect the completion of the operation.
release "dao-2048" uninstalled

```

### DIFF
--- a/pkg/action/uninstall.go
+++ b/pkg/action/uninstall.go
@@ -17,12 +17,15 @@ limitations under the License.
 package action
 
 import (
+	"fmt"
+	"os"
 	"strings"
 	"time"
 
 	"github.com/pkg/errors"
 
 	"helm.sh/helm/v3/pkg/chartutil"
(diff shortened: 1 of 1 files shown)


Cweiping:
Without changing the existing logic, I split the error types:

there is an error in generating the list of resources to be deleted, and the operation fails;
in the process of deleting resources, an error occurs in the upstream server, and the operation is successful, but it will prompt the user that some errors occur in the process of deleting, which will not affect the deletion of release information;

Issue #6410, opened 2019-09-11
[Feature Request]: running a single test & ability to pass values specific to the test


Leekao:
I'd like the ability to run something like "helm test [deployment name] [test name]" and helm will run the test with the corresponding name, also, I'd like the ability to override the values that were initially defined in the test with new values, something like "helm test [deployment name] --set key=value" would be awesome.

chriswells0:
In addition to being able to run a single test, it would be useful for an an application operator/distributor to be able to specify one or more tests to be skipped when executing the test hook. This is helpful when performing QA on a chart that has a large number of tests with one or more failing due to a defect in the test. It allows QA to proceed without being blocked in order to create a complete list of failing tests.

PR #8484, opened 2020-07-19
feat: Allow helm test to run a subset of tests

chriswells0:

refs #6410

This is very useful when developing Helm tests in a chart that has many tests and/or tests that take a long time to execute. Similarly, it allows QA testers to validate that issues with a specific test have been resolved.

I will also submit a PR to skip specific tests. If it is accepted, this PR will need a 1-line addition.


hickeyma (changes requested):
@chriswells0 Thanks for pushing the PR. When I compare this PR with #8485, it makes me think that there should be only this PR. I think the one flag would suffice to cover both use cases. You list the test you want run which as a result skips the non specified. My reason is to simplify code and user flags.

chriswells0:
That would definitely work, but it could be painful for charts having many tests. In our case, we ran into opposite problems while working on charts with dozens of tests.

1. With only `--skip`, a dev debugging a single test would need to run them all and copy/paste the names of all tests except one.
2. With only `--subset`, a QA engineer who's blocked from completing test executions due to a failure would need to copy/paste all the successful test names and then... refer to documentation to find what else to list?

I agree that they could be seen as redundant, but they do meet different needs. If you disagree, feel free to close the other PR. Thanks!

paulczar:
If we bring support for the `--post-renderer` flag into the `helm test` command then we likely use a post renderer to provide this functionality.  Whether or not that's a user friendly idea is another question, but that could also allow for tests that are out-of-band to the chart itself.

If we want to allow test filtering like this ... I would agree with providing both the `skip` and `subset` options. as I've certainly had frustrations with tools that only provide one or the other.  Classic example k8s RBAC  is allow-list only, not supporting a deny-list is a UX nightmare.

paulczar:
An alternative solution would be to allow for a `--filter` flag that would cover both use cases `--filter (name=testSecret)`  or `--filter (! name=testSecret).

hickeyma:
> An alternative solution would be to allow for a --filter flag that would cover both use cases --filter (name=testSecret) or `--filter (! name=testSecret).

I would be more in favour of this as I feel the less flags the better sometimes.

chriswells0:
To clarify the ask here, are you wanting that literal syntax or was it an example? I like the idea of a single parameter with a NOT value, but it seems like an awkward syntax with the parentheses, name=value, and optional `!` with (or without?) space. I could code it that way, but I want to be sure before putting the effort in. Just the name and an optional `!` like `--filter test-one` and `--filter !test-two` seems much simpler, so is that OK?

chriswells0:
I've reimplemented this as `--filter` accepting `name=test-name` and `!name=test-name`/`! name=test-name`. I have to admit it's a bit awkward, though. If you provide a `!`, then you have to wrap it in quotes to keep the shell from choking. They also have to be single quotes, because the shell will still choke if you use double quotes. Thus, the user experience is a bit error-prone, but it does work.

I didn't bother to add the parentheses since they made the syntax more awkward. If you want them, please let me know.

bacongobbler:
Adding to 3.4.0 for further review.

mattfarina:
@chriswells0 ping. We are looking to put together the 3.4.0 release and I was wondering how this was going.

chriswells0:
@mattfarina I was waiting for feedback. I made the changes described in my last comment based on the previous feedback. Please let me know if the current behavior is what you would like or if further changes are needed.

pkg/action/release_testing.go
```diff
@@ -46,7 +46,7 @@ func NewReleaseTesting(cfg *Configuration) *ReleaseTesting {
 }
 
 // Run executes 'helm test' against the given release.
-func (r *ReleaseTesting) Run(name string) (*release.Release, error) {
+func (r *ReleaseTesting) Run(name string, filters map[string][]string) (*release.Release, error) {
```
seaneagan, line 49: isn't this a breaking change to add a required argument to this method? why not just add it to the `ReleaseTesting` struct?

chriswells0: It didn't appear to be a breaking change as I didn't find any other place calling the method. As for why I didn't add it to `ReleaseTesting`, the simple answer is that this is the only Go code I've ever written. I wasn't even aware of the meaning of that method syntax until just now.

mattfarina: `Run` is a part of the public API and Helm uses SemVer. That means function signature can't change. It can be documented as deprecated and a new function created and used with the new signature.

I realize this is in a test and it's highly unlikely that this function is in use by anyone. But, there is no way to check that.

This function is also listed on pkg.go.dev.

So, can we do this in a way that doesn't change any signature.

chriswells0: Makes sense. I should be able to revisit it this weekend.

Are there any other changes needed?

mattfarina:
Let me first apologize for not getting to this sooner. We have had some other things we had to jump on and didn't get to this as quickly as I would have liked. I need to move this out to 3.5 as I don't have time to review for 3.4. We need to get 3.4 out the door.

Starting with 3.5, we have a schedule for releasing minor releases (where there are new features). This one will be on January 13th. You can read about this process at https://github.com/helm/community/blob/master/hips/hip-0002.md

pkg/action/release_testing.go
chriswells0: I've made this requested change. My apologies for the delay. Busy times.

chriswells0: @seaneagan: For some reason, it won't allow me to request a re-review from you. I'm adding the mention just to be sure you're aware that the change has been made.

mattfarina:
There are two things I'm working through with regard to this PR...

1. Who are the actors in this system and what are their needs? We have a set of [user profiles (roles)](https://github.com/helm/community/blob/master/user-profiles.md) that are intentionally ordered.

`helm test` is a great way for an application operator to know if the application was stood up and is working correctly. An application operator isn't expected to know much about the business logic of the application or specific elements of Kubernetes needed for that. Think about when someone uses apt to install PostgreSQL. They don't need to know the business logic of PostgreSQL to run it and they aren't expected to know all the detail on how debian based OSes work.

For these folks, what will the `name` mean to them? How would it be useful and why?

Then there are the application distributors. They know about the chart, Kubernetes, and the applications. They can control the chart (something an application operator isn't expected to be able to do). So, they could bake in business logic to skip or control tests that's not documented as public via values. Flags that could be passed in and used by a QA person in their organization. This could even be paired with something like [chart-testing](https://github.com/helm/chart-testing) to test different variants.

What will the `name` mean to someone in QA? Is this just a name that shows up in a failed test?

2. Tests can include other assets, like a ConfigMap or Secret. So, test hooks could deploy a secret and a pod that uses it. How does that work in this scenario?

chriswells0:
There is no role listed for someone performing QA on charts that will be deployed. I don't know if you consider a QA engineer to be an application operator or application distributor when performing deployments. In the case of QA as well as developers, the `name` always refers to the name of the test.

When you have a large deployment consisting of charts that have subcharts that have their own subcharts, controlling the tests is essentially impossible today. QA may need to determine what tests are failing, which can't be done when `helm test` stops after 1 failure. They could get the `name` from the failed test and specify to skip it for the next round of tests. Application developers can run only the test they've modified so it doesn't take 10 or 20 minutes to verify small changes.

I didn't test a ConfigMap/Secret scenario. I would expect them to be skipped since the test itself is skipped.

chriswells0:
Just FYI: When I read #6410 again just now, I remembered that I did not implement the part about using `--set` to provide values to tests. Someone more familiar may need to implement that if it makes sense or is even possible.

technosophos (approved):
I am okay with this change, but want to be clear that this pushes more burden to test developers. Because a filter could filter out resources that are dependencies for a test case, it would be possible for tests to return both false positives and false negatives if the tests are not written to execute independently.

This, of course, is true of basically _all_ testing platforms. (You can write a unit test that assumes that another unit test was run.) So I am comfortable accepting this.

Again, while I am fine with it for now, it should be noted that filtering only on `name` might not be sufficient. We can add more in the future.

mattfarina (approved):
The UX around this feature is a little funny to me. If you have no tests it doesn't tell you that anywhere. But, I think it's ok for now as this is an advanced feature we can provide more with later. Following along with technosophos thoughts.

hickeyma (approved):
I agree with @technosophos @mattfarina that it is good for now and can build on it as we go along.

hickeyma:
Thanks for working on this @chriswells0

PR #8491, opened 2020-07-21
fix(sdk): Polish the downloader/manager package error return

### LINKED ISSUE #8471: The sdk dependency manager error refers to Helm CLI
Output of `helm version`: v3.4.0 development and previous v3 releases

The download manager within the SDK notes... https://github.com/helm/helm/blob/6a536661fcf1dc0315798ece70bbf193e725945c/pkg/downloader/manager.go#L414

The SDK should have an error that's general and the CLI code in the cmd directory should detect it and decorate with any CLI related elements. Separation of concerns and all.

donggangcj:
Close #8471

fix #8471

**If applicable**:

### DIFF
--- a/cmd/helm/dependency_build.go
+++ b/cmd/helm/dependency_build.go
@@ -16,12 +16,11 @@ limitations under the License.
 package main
 
 import (
+	"fmt"
 	"io"
 	"os"
 	"path/filepath"
-	"strings"
 
-	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"k8s.io/client-go/util/homedir"
 
@@ -69,7 +68,7 @@ func newDependencyBuildCmd(out io.Writer) *cobra.Command {
 			}
 			err := man.Build()
 			if e, ok := err.(downloader.ErrRepoNotFound); ok {
-				return errors.Errorf("no repository definition for %s. Please add the missing repos via 'helm repo add'", strings.Join(e.Repos, ", "))
+				return fmt.Errorf("%s. Please add the missing repos via 'helm repo add'", e.Error())
 			}
 			return err
 		},
--- a/cmd/helm/dependency_build.go
+++ b/cmd/helm/dependency_build.go
@@ -19,7 +19,9 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strings"
 
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"k8s.io/client-go/util/homedir"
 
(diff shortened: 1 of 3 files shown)


donggangcj:
@mattfarina  I'm not sure if these modifications are consistent with your ideas.

jdolitsky (approved):

cmd/helm/dependency_build.go
```diff
@@ -65,7 +67,11 @@ func newDependencyBuildCmd(out io.Writer) *cobra.Command {
 			if client.Verify {
 				man.Verify = downloader.VerifyIfPossible
 			}
-			return man.Build()
+			err := man.Build()
+			if e, ok := err.(downloader.ErrRepoNotFound); ok {
+				return errors.Errorf("no repository definition for %s. Please add the missing repos via 'helm repo add'", strings.Join(e.Repos, ", "))
```
bacongobbler, line 72: couldn't this display repositories that might've succeeded? If I'm reading the code correctly, `e.Repos` and `missing` might not evaluate to the same result. There may only be a few repositories missing from the list.

I'd suggest appending the helpful error message to the output of `e.Error()`, like so:

```
return fmt.Errorf("%s. Please add the missing repos via 'helm repo add'", e.Error())
```

That way we leave the output of the error message up to the download manager.

bacongobbler (changes requested):
Please write a unit test to ensure the output of `Error()` works as expected. Thanks.

cmd/helm/dependency_build.go
donggangcj: > couldn't this display repositories that might've succeeded? 

`Manager` method `hasAllRepos` ensures that all of the referenced deps are in the local repo cache and it doesn't return succeeded repositories. 

> If I'm reading the code correctly, `e.Repos` and `missing` might not evaluate to the same result. There may only be a few repositories missing from the list.

My bad! I did not process erros in the function `resolveRepoNames`，but i didn't find the reason why `e.Repos` and `missing` is different. Do you mean to merge `resolveRepoNames` with `hasAllRepos`?🤣

> 
> I'd suggest appending the helpful error message to the output of `e.Error()`, like so:
> 
> ```
> return fmt.Errorf("%s. Please add the missing repos via 'helm repo add'", e.Error())
> ```

Done; I removed some tips of `resolveRepoName` and is it necessary to show it here by a error handling function.

https://github.com/helm/helm/blob/241785c70fb38b2c074d7b3ddf0925812fb3fc69/pkg/downloader/manager.go#L469-L478

There seems to be an obvious error in this code

pkg/downloader/manager_test.go
```diff
@@ -360,3 +360,32 @@ func TestBuild_WithRepositoryAlias(t *testing.T) {
...
+		name   string
+		fields fields
+		want   string
+	}{
+		{
+			name: "OK",
+			fields: fields{
+				Repos: []string{"https://charts1.example.com", "https://charts2.example.com"},
+			},
+			want: "no repository definition for https://charts1.example.com, https://charts2.example.com",
```
bacongobbler, line 378: Can you please also add a test case for the issue we discussed earlier, where only a few repositories in the list are missing.

bacongobbler (approved):

Issue #8467, opened 2020-07-16
[helm-3.3.0-rc.1] Helm Lint fails due to blank resource name


rblaine95:
I've got a helm chart which uses internal Helm Libraries to initialize a Database and User in a DB Cluster as a pre-install hook.
```yaml
# templates/pg-job.yaml
{{- if .Values.pgJob.enabled }}
  {{- include "zenlibs.pgJob" (list . .Values.pgJob) }}
{{- end }}

# values.yaml
[...]
pgJob:
  enabled: true
  init: true
  clean: true
  admin:
    secret:
      name: dev-aurora
  host: *pgHost
  port: 5432
  version: 10.7
```
Brief output of `helm template example .`
```yaml
# Source: example/templates/pg-job.yaml
kind: Role
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: example-secret-role
[...]
# Source: example/templates/pg-job.yaml
kind: RoleBinding
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: example-secret-rolebinding
[...]
# Source: example/templates/pg-job.yaml
apiVersion: batch/v1
kind: Job
metadata:
  name: example-initpg
[...]
# Source: example/templates/pg-job.yaml
apiVersion: batch/v1
kind: Job
metadata:
  name: example-cleanpg
[...]
```
Output of `helm lint`
```sh
==> Linting .
[ERROR] templates/pg-job.yaml: object name does not conform to Kubernetes naming requirements: ""

Error: 1 chart(s) linted, 1 chart(s) failed
```

`helm lint` succeeds if I use `helm lint --set pgJob.enabled=false`

### Extra Info

Output of `helm version`:
```sh
» helm version
version.BuildInfo{Version:"v3.3.0-rc.1", GitCommit:"5c2dfaad847df2ac8f289d278186d048f446c70c", GitTreeState:"dirty", GoVersion:"go1.14.4"}
```

Output of `kubectl version`:
```sh
» kubectl version
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.5", GitCommit:"e6503f8d8f769ace2f338794c914a96fc335df0f", GitTreeState:"archive", BuildDate:"2020-07-01T16:28:46Z", GoVersion:"go1.14.4", Compiler:"gc", Platform:"linux/amd64"}
[...]

PR #8496, opened 2020-07-21
Fix/8467 linter failing

technosophos:

This PR addresses the problem identified in #8467. It adds a new linter rule that issues a warning when a YAML document starts out with extra spaces before the first YAML attribute.

For example, this YAML would produce a warning:

```yaml
  apiVersion: v1
kind: Pod
```

It warns that there are extra spaces before `apiVersion`.

The reason for warning is that the YAML parser we use calibrates its spaces to the first line. Thus, it is a formatting error to start with spaces.

**If applicable**:

### DIFF
--- a/pkg/lint/rules/template.go
+++ b/pkg/lint/rules/template.go
@@ -17,6 +17,8 @@ limitations under the License.
 package rules
 
 import (
+	"bufio"
+	"bytes"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -121,7 +123,7 @@ func Templates(linter *support.Linter, values map[string]interface{}, namespace 
 
 		renderedContent := renderedContentMap[filepath.Join(chart.Name(), fileName)]
 		if strings.TrimSpace(renderedContent) != "" {
-			fmt.Printf("Template %s: \n %s", path, renderedContent)
+			linter.RunLinterRule(support.WarningSev, path, validateTopIndentLevel(renderedContent))
 			var yamlStruct K8sYamlStruct
 			// Even though K8sYamlStruct only defines a few fields, an error in any other
 			// key will be raised as well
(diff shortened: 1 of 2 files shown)


jdolitsky (approved):

donggangcj (commented):
Everything is ok!

pkg/lint/rules/template.go
```diff
@@ -136,6 +139,32 @@ func Templates(linter *support.Linter, values map[string]interface{}, namespace
...
+	// Read lines until we get to a non-empty one
+	scanner := bufio.NewScanner(bytes.NewBufferString(content))
+	for scanner.Scan() {
+		line := scanner.Text()
+		// If line is empty, skip
+		if strings.TrimSpace(line) == "" {
+			continue
+		}
+		// If it starts with one or more spaces, this is an error
+		if strings.HasPrefix(line, " ") || strings.HasPrefix(line, "\t") {
```
mattfarina, line 159: I've been digging through YAML parsers, validators, and the specification. I've learned a few things...

If everything is indented the same level, it's valid. So, if the first line is indented two spaces but the others aren't it's invalid. But, if each line has that two space indentation it's valid. The YAML parser we use handles the all the things are indented case. So the two following cases are valid:

```
foo: bar
baz: quz
```
and
```
  foo: bar
  baz: quz
```
but, the following is invalid...
```
  foo: bar
baz: quz
```
Different tools handle the failure case differently. I don't see, in the spec, a correct way to handle this. Some tools take the lines with the indentation and parse up to the point where the outdent happens and then ignore the rest. The YAML parser we have appears to do that.

I'm not entirely sure of the right way to check this prefix situation.

technosophos: Linter's job is mainly to be prescriptive. Indenting "at the top level" is a bad practice, so linter should warn about it.

mattfarina:
I've slated this for 3.4.0 because it's technically a new feature. The linter catching something new. 3.3.0 is already in the RC stage so we are only looking to include certain fixes there.

PR #8516, opened 2020-07-26
fix: removed strict template errors from v2 linter

### LINKED ISSUE #6705: lint undefined keys and "and"

jeff-knurek:
Signed-off-by: Jeff Knurek <knurek.stuff@gmail.com>

closes #6705 

followed similar solution as in v3 (#8017) ~~but also cleared out the need for passing the `strict` bool around to different functions~~
* also added a test to the lintCmd to confirm the `--strict` flag functions as expected

**If applicable**:
- [x] this PR has been tested for backwards compatibility _(some charts that did not pass `helm lint --strict` will now pass.)_


bacongobbler (approved):
LGTM. Will need another maintainer to sign off on this before we can merge.

pkg/lint/rules/template.go
```diff
@@ -32,7 +32,7 @@ import (
 )
 
 // Templates lints the templates in the Linter.
-func Templates(linter *support.Linter, values []byte, namespace string, strict bool) {
+func Templates(linter *support.Linter, values []byte, namespace string) {
```
hickeyma, line 35: Is this a backwards compatibility change? If so then the change should only be the removal of line as you did below:

```yaml
if strict {	
    e.Strict = true	
}
```

Maybe, just make changes similar to #8017?

hickeyma (changes requested):
Thanks for pushing the PR @jeff-knurek. Comment inline about backwards compatibility.

pkg/lint/rules/template.go
technosophos: I agree. It is okay to leave a no-op param in the name of preserving API compatibility.

jeff-knurek: oh, you're right, I wasn't thinking about any external usage. I agree that it does make more sense to keep the param. 

however, is the change really still "backwards compatible" if someone using the library function with the current logic, and is still expecting `e.Strict = true` to be set? I mean to say that such a change _should_ break the api. but the benefits for doing so in `v2.x` don't outweigh the negative impact of breaking the api, so i'll edit the PR to keep the param.

jeff-knurek:
test failure unrelated to change:
```
FAIL	k8s.io/helm/cmd/tiller [build failed]
```

hickeyma (changes requested):
Thanks for the updates @jeff-knurek. It's nearly there. There is just some cleanup of unnecessary test files as follows:

- `cmd/helm/testdata/testcharts/bad.name/.helmignore`
- `cmd/helm/testdata/testcharts/bad.name/templates/*` (you can leave the `templates` dir)

hickeyma (approved):
LGTM, thanks @jeff-knurek.

technosophos (approved):
This looks great. Thanks for getting this done before the code freeze deadline

PR #8521, opened 2020-07-27
introduce stale issue bot

bacongobbler:
running once per day, this action will check to see if any issues have been stale for 90 days. After an additional 30 days, the issue will be automatically closed.

This is in response to the growing number of issues in the queue. Threads can be marked as "keep open" if they want to be kept open.

### DIFF
--- a/.github/workflows/stale-issue-bot.yaml
+++ b/.github/workflows/stale-issue-bot.yaml
@@ -1,0 +1,15 @@
+name: "Close stale issues"
+on:
+  schedule:
+  - cron: "0 0 * * *"
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/stale@v3
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        stale-issue-message: 'This issue has been marked as stale because it has been open for 90 days with no activity. This thread will be automatically closed in 30 days if no further activity occurs.'
+        exempt-issue-labels: 'keep open'
+        days-before-stale: 90
+        days-before-close: 30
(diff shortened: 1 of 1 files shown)


mattfarina:
This has been a hot topic over time. Not using the stale issue bot was an intentional choice the last time we had it.

cc @helm/helm-core-maintainers

thomastaylor312:
I feel like the rule set here is much more flexible. It is a 3 month period before marking as stale, and then another month before it actually closes. This would allow us to make sure we know exactly what we are keeping open and why as it requires labeling from the maintainers. I think it would be acceptable, but we definitely should have a bigger discussion in the community call

bacongobbler:
Right. I've been the biggest proponent against using a stale issue bot in the past. With the amount of issues kept open over time, my stance has changed.

A few months ago we discussed manually closing out issues based on these parameters. Three weeks ago I gave it a trial run, and we cleared out a significant number of issues without much backlash, so I'm inclined to believe this approach of 3 months = notice and 1 additional month before closing seems to work out.

mattfarina (approved):
This looks good to me and we've given time for others to weigh in

.github/workflows/stale-issue-bot.yaml
```diff
@@ -0,0 +1,15 @@
...
+  - cron: "0 0 * * *"
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/stale@v3
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        stale-issue-message: 'This issue has been marked as stale because it has been open for 90 days with no activity. This thread will be automatically closed in 30 days if no further activity occurs.'
+        exempt-issue-labels: 'keep open'
```
mattfarina, line 13: Can we add the label for helm v4 issues to this?

bacongobbler: good call.

bacongobbler: I think I found an open issue with multiple labels using spaces via https://github.com/actions/stale/issues/98. I'll see if I can write a test case to see if that may be at play here as well, and if we need to submit a patch upstream before we can add the v4 label to the exemption list.

bacongobbler: https://github.com/actions/stale/issues/98#issuecomment-676649307

PR #8522, opened 2020-07-27
feat(cmd): add Chart Version to helm list output

### LINKED ISSUE #8503: proposal: add Chart Version to helm list output
We have a use case for determining the Chart Version of a chart installed in to a cluster, as distinct from the App Version, which can already be inferred / discovered.

I'd like to request / propose that Chart Version be added as a column to the output of `helm list`.

For example, here's the current output:

```
NAME                    NAMESPACE       REVISION        UPDATED                                 STATUS          CHART                           APP VERSION
...
prometheus-operator     prometheus      1               2020-07-21 23:32:48.402769335 +0000 UTC deployed        prometheus-operator-8.5.12      0.34.0
```

Interestingly, the Chart Version is already encoded into the Chart field. Under the hood, this is generated by concatenating the `Metadata.Name` and `Metadata.Version` fields:

https://github.com/helm/helm/blob/0ad800ef43d3b826f31a5ad8dfbb4fe05d143688/cmd/helm/list.go#L158

Perhaps these two individual fields could be printed, separately, in addition to the concatenated form.

nicktrav:

This PR implements the proposal in #8503 to add the "Chart Version" to the `helm list` output.

Closes #8503.

### DIFF
--- a/cmd/helm/testdata/output/list-all.txt
+++ b/cmd/helm/testdata/output/list-all.txt
@@ -1,9 +1,9 @@
-NAME       	NAMESPACE	REVISION	UPDATED                      	STATUS         	CHART          	APP VERSION
-drax       	default  	1       	2016-01-16 00:00:01 +0000 UTC	uninstalling   	chickadee-1.0.0	0.0.1      
-gamora     	default  	1       	2016-01-16 00:00:01 +0000 UTC	superseded     	chickadee-1.0.0	0.0.1      
-groot      	default  	1       	2016-01-16 00:00:01 +0000 UTC	uninstalled    	chickadee-1.0.0	0.0.1      
-hummingbird	default  	1       	2016-01-16 00:00:03 +0000 UTC	deployed       	chickadee-1.0.0	0.0.1      
-iguana     	default  	2       	2016-01-16 00:00:04 +0000 UTC	deployed       	chickadee-1.0.0	0.0.1      
-rocket     	default  	1       	2016-01-16 00:00:02 +0000 UTC	failed         	chickadee-1.0.0	0.0.1      
-starlord   	default  	2       	2016-01-16 00:00:01 +0000 UTC	deployed       	chickadee-1.0.0	0.0.1      
-thanos     	default  	1       	2016-01-16 00:00:01 +0000 UTC	pending-install	chickadee-1.0.0	0.0.1      
+NAME       	NAMESPACE	REVISION	UPDATED                      	STATUS         	CHART          	APP VERSION	CHART VERSION
+drax       	default  	1       	2016-01-16 00:00:01 +0000 UTC	uninstalling   	chickadee-1.0.0	0.0.1      	1.0.0        
+gamora     	default  	1       	2016-01-16 00:00:01 +0000 UTC	superseded     	chickadee-1.0.0	0.0.1      	1.0.0        
+groot      	default  	1       	2016-01-16 00:00:01 +0000 UTC	uninstalled    	chickadee-1.0.0	0.0.1      	1.0.0        
+hummingbird	default  	1       	2016-01-16 00:00:03 +0000 UTC	deployed       	chickadee-1.0.0	0.0.1      	1.0.0        
+iguana     	default  	2       	2016-01-16 00:00:04 +0000 UTC	deployed       	chickadee-1.0.0	0.0.1      	1.0.0        
+rocket     	default  	1       	2016-01-16 00:00:02 +0000 UTC	failed         	chickadee-1.0.0	0.0.1      	1.0.0        
+starlord   	default  	2       	2016-01-16 00:00:01 +0000 UTC	deployed       	chickadee-1.0.0	0.0.1      	1.0.0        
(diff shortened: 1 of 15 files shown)


nicktrav:
cc: @bacongobbler

cmd/helm/testdata/output/list-all.txt
```diff
@@ -1,9 +1,9 @@
-NAME       	NAMESPACE	REVISION	UPDATED                      	STATUS         	CHART          	APP VERSION
-drax       	default  	1       	2016-01-16 00:00:01 +0000 UTC	uninstalling   	chickadee-1.0.0	0.0.1      
-gamora     	default  	1       	2016-01-16 00:00:01 +0000 UTC	superseded     	chickadee-1.0.0	0.0.1      
-groot      	default  	1       	2016-01-16 00:00:01 +0000 UTC	uninstalled    	chickadee-1.0.0	0.0.1      
-hummingbird	default  	1       	2016-01-16 00:00:03 +0000 UTC	deployed       	chickadee-1.0.0	0.0.1      
-iguana     	default  	2       	2016-01-16 00:00:04 +0000 UTC	deployed       	chickadee-1.0.0	0.0.1      
-rocket     	default  	1       	2016-01-16 00:00:02 +0000 UTC	failed         	chickadee-1.0.0	0.0.1      
-starlord   	default  	2       	2016-01-16 00:00:01 +0000 UTC	deployed       	chickadee-1.0.0	0.0.1      
-thanos     	default  	1       	2016-01-16 00:00:01 +0000 UTC	pending-install	chickadee-1.0.0	0.0.1      
+NAME       	NAMESPACE	REVISION	UPDATED                      	STATUS         	CHART          	APP VERSION	CHART VERSION
```
mattfarina, line 1: I notice that the character count for the columns goes from 104 to 118 in width in this example. I wonder what impact this would have on users with smaller screens. I ran a couple tests on a Surface Go (yeah, i do dev on it sometimes) and managed to get some screen wrap.

mattfarina (commented):
The chart field has the chart name and version in it. I understand that the one that calls this will need to parse the chart version from the returned data but it's not all that hard. You split on the `-` and the second part is the version.

I'm curious for the reason of adding it as a second field. We're already running into terminal space issues.

cmd/helm/testdata/output/list-all.txt
hickeyma: I agree with @mattfarina re the need for the field especially as it is already part of the `Chart` field.

nicktrav: Here's my perspective, as the motivation for the original issue, and the change here.

Currently, to get the _chart_ version, we have to strip the release name from the Chart field, which leaves the version as the remainder (`strings.TrimPrefix` type of thing). However, if the chart name is different to the release name (the thing being stripped), then this quickly becomes error prone, and you have to resort to something like a regular expression to pluck the semantic version out of the string. Definitely doable, but it's smelly.

My question for the both of you would be, why does a user have to do that when the "primitive" information is contained in struct? Flipping it around the other way, if I want "Chart", I can just combine the chart name and the version in whatever tooling I have (e.g. `Chart.Metadata.Name + Chart.Metadata.Version`), provided both are made available to the user.

Put another way, I think the utility of having the two fields exposed separately and requiring the user to join the two (if they even need that) is greater than having one field and  requiring the user to reverse engineer how the strings were joined to get the constituent elements. 

Respectfully, I don't really buy the argument that we shouldn't add fields because we might screen wrap on small screens. If that's the case, I'd suggest _removing_ the "Chart" field altogether, and instead give the user access to both the primitive fields `Chart.Metadata.Name` and `Chart.Metadata.Version` individually.

mattfarina: There are a few things here....

First, it would be easier of the chart name and version were separate. But, we can't remove the chart column until Helm v4. As part of our SemVer policy we don't remove the output as it will break tools that read the output strings and parse them. These tools do exist and we are aware of some of them. Helm takes stability seriously and this is part of that.

For Helm v4, where a major version change can break APIs, I would like to break these into two columns. It would make things easier.

Second, we take into account ergonomics for people. If the screen is going to wrap for a bunch of users it makes the output difficult to read. Would you want a tool that did that? We would be many complaints if we did that.

Third, people parse the chart column in languages other than Go. If we think in general terms it's possible to come up with a simple solution. Split the chart string based on `-`. Parse the last part as semver. If it doesn't parse take the last two sections and parse those. You'll have the version.

This is because the last part is either the semver or the pre-release on a semver. If it's a pre-release than the last 2 parts are the full semver. You don't need to know the chart name compared to the release name for this.

I hope this helps you understand why we have the position we take and have some practical steps in moving forward without requiring Helm to change. You can be unblocked in the short term.

nicktrav: Great, thanks for the context. I'll go ahead and close this out.

> For Helm v4, where a major version change can break APIs, I would like to break these into two columns. It would make things easier.

Is this being tracked anywhere? Is it possible to re-purpose #8503 for this?

PR #8527, opened 2020-07-28
Fix issue with install and upgrade running all hooks

mattfarina:
When #8156 was merged it had the side effect that all hooks were
run all the time. All the hooks were put in the flow of the
content rendered and sent to Kubernetes on every command.

For example, if you ran the following 2 commands the test hooks
would run:

    helm create foo
    helm install foo ./foo

This should not run any hooks. But, the generated test hook is run.

The change in this commit moves the writing of the hooks to output
or disk back into the template command rather than in a private
function within the actions. This is where it was for v3.2.

One side effect is that post renderers will not work on hooks. This
was the case in v3.2. Since this bug is blocking the release of v3.3.0
it is being rolled back. A refactor effort is underway for this section
of code. post renderer for hooks should be added back as part of that
work. Since post renderer hooks did not make it into a release it
is ok to roll it back for now.

There is code in the cmd/helm package that has been duplicated from
pkg/action. This is a temporary measure to fix the immediate bug
with plans to correct the situation as part of a refactor
of renderResources.

It fixes a release blocking bug to v3.3.0. v3.3.0-rc.1 currently runs all hooks on all commands (e.g., `helm install`).

@technosophos and I discussed this. He's volunteered to refactor this section of code for v3.4. If he doesn't, I will. There are mixes of concerns and some other bugs (e.g., post renderers don't work when helm template writes to an output directory). The other bugs are not regressions and were there in v3.2 so they are not release blockers.

### DIFF
--- a/cmd/helm/template.go
+++ b/cmd/helm/template.go
@@ -20,6 +20,8 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"os"
+	"path"
 	"path/filepath"
 	"regexp"
 	"sort"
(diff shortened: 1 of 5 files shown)


technosophos (approved):
This is the fix we had discussed, and it looks good to me.

technosophos:
I have started on a refactor of the `renderResources` function, but it is probably weeks out at least (given our PR velocity).

bacongobbler (approved):
LGTM. Nice catch on this one.

bacongobbler:
> I have started on a refactor of the `renderResources` function, but it is probably weeks out at least (given our PR velocity).

Let me know when you have something ready. `renderResources` is an absolute mess and I'd love to see any effort to clean up that part of the codebase.

PR #8529, opened 2020-07-29
feat: add flag `--show-desc` to display description for `helm status`

### LINKED ISSUE #8525: Question/Feature: Display description in `status` command?
The `status` command won't output description but `history` command does. Is there any reason?

I think it's useful when debugging failed releases, and web dashboard using helm SDK like `kubeapps` can show the failed reason easily.

I would like to create a PR to resolve this if it's OK.

lingsamuel:
Signed-off-by: Ling Samuel <lingsamuelgrace@gmail.com>

Closes #8525 

Display description in `helm status` command.

**If applicable**:

### DIFF
--- a/cmd/helm/status.go
+++ b/cmd/helm/status.go
@@ -39,6 +39,8 @@ The status consists of:
 - last deployment time
 - k8s namespace in which the release lives
 - state of the release (can be: unknown, deployed, uninstalled, superseded, failed, uninstalling, pending-install, pending-upgrade or pending-rollback)
+- revision of the release
+- description of the release (can be completion message or error message, need to enable --show-desc)
 - list of resources that this release consists of, sorted by kind
 - details on last test suite run, if applicable
 - additional notes provided by the chart
@@ -68,7 +70,7 @@ func newStatusCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 			// strip chart metadata from the output
 			rel.Chart = nil
 
-			return outfmt.Write(out, &statusPrinter{rel, false})
+			return outfmt.Write(out, &statusPrinter{rel, false, client.ShowDescription})
 		},
 	}
 
@@ -88,13 +90,15 @@ func newStatusCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 	}
 
 	bindOutputFlag(cmd, &outfmt)
+	f.BoolVar(&client.ShowDescription, "show-desc", false, "if set, display the description message of the named release")
 
 	return cmd
 }
 
 type statusPrinter struct {
-	release *release.Release
-	debug   bool
+	release         *release.Release
+	debug           bool
+	showDescription bool
 }
 
 func (s statusPrinter) WriteJSON(out io.Writer) error {
@@ -116,6 +120,9 @@ func (s statusPrinter) WriteTable(out io.Writer) error {
 	fmt.Fprintf(out, "NAMESPACE: %s\n", s.release.Namespace)
 	fmt.Fprintf(out, "STATUS: %s\n", s.release.Info.Status.String())
 	fmt.Fprintf(out, "REVISION: %d\n", s.release.Version)
+	if s.showDescription {
+		fmt.Fprintf(out, "DESCRIPTION: %s\n", s.release.Info.Description)
+	}
 
 	executions := executionsByHookEvent(s.release)
 	if tests, ok := executions[release.HookTest]; !ok || len(tests) == 0 {
--- a/pkg/action/status.go
+++ b/pkg/action/status.go
@@ -27,6 +27,11 @@ type Status struct {
 	cfg *Configuration
 
 	Version int
+
+	// If true, display description to output format,
+	// only affect print type table.
+	// TODO Helm 4: Remove this flag and output the description by default.
+	ShowDescription bool
 }
 
 // NewStatus creates a new Status object with the given configuration.
--- a/cmd/helm/get_all.go
+++ b/cmd/helm/get_all.go
@@ -59,7 +59,7 @@ func newGetAllCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 				return tpl(template, data, out)
 			}
 
-			return output.Table.Write(out, &statusPrinter{res, true})
+			return output.Table.Write(out, &statusPrinter{res, true, false})
 		},
 	}
 
--- a/cmd/helm/install.go
+++ b/cmd/helm/install.go
@@ -122,7 +122,7 @@ func newInstallCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 				return err
 			}
 
-			return outfmt.Write(out, &statusPrinter{rel, settings.Debug})
+			return outfmt.Write(out, &statusPrinter{rel, settings.Debug, false})
 		},
 	}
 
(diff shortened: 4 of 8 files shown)


cmd/helm/status.go
```diff
@@ -116,6 +116,9 @@ func (s statusPrinter) WriteTable(out io.Writer) error {
 	fmt.Fprintf(out, "NAMESPACE: %s\n", s.release.Namespace)
 	fmt.Fprintf(out, "STATUS: %s\n", s.release.Info.Status.String())
 	fmt.Fprintf(out, "REVISION: %d\n", s.release.Version)
+	if len(s.release.Info.Description) > 0 {
```
hickeyma, line 119: we usually use following to test for empty string: `if s.release.Info.Description != "" {`

hickeyma (changes requested):
Thanks for pushing the PR @lingsamuel.

hickeyma:
@lingsamuel  I would like to confirm with @helm/helm-core-maintainers that this PR does not break backwards compatibility.

hickeyma:
> I would like to confirm with @helm/helm-core-maintainers that this PR does not break backwards compatibility.

On the dev weekly meeting today, we discussed this. The general consensus is that adding the description field breaks backwards compatibility  as it adds a new row of data that could potentially break a user parsing code.

@lingsamuel Can you therefore, add this behind a flag and add a comment that says that the flag should be removed in Hem 4 and the field added automatically to the output.

cmd/helm/status.go
```diff
@@ -39,7 +39,7 @@ The status consists of:
 - last deployment time
 - k8s namespace in which the release lives
 - state of the release (can be: unknown, deployed, uninstalled, superseded, failed, uninstalling, pending-install, pending-upgrade or pending-rollback)
-- list of resources that this release consists of, sorted by kind
```
hickeyma, line 42: Add this back in as it should still be output and there is/are bug(s) currently opened to fix it.

lingsamuel:
I add a flag `--show-desc` for:
- status
- get all
- install
- upgrade

These commands call `statusPrinter`.

cmd/helm/status.go
```diff
@@ -116,6 +119,9 @@ func (s statusPrinter) WriteTable(out io.Writer) error {
 	fmt.Fprintf(out, "NAMESPACE: %s\n", s.release.Namespace)
 	fmt.Fprintf(out, "STATUS: %s\n", s.release.Info.Status.String())
 	fmt.Fprintf(out, "REVISION: %d\n", s.release.Version)
+	if s.statusPrinterOptions.ShowDescription && s.release.Info.Description != "" {
```
liuming-dev, line 122: If `s.release.Info.Description` is empty string `''`, could we show the DESCRIPTION as empty string to tell users the DESCRIPTION is empty rather than don't show the DESCRIPTION?

pkg/action/get_all.go
```diff
@@ -0,0 +1,50 @@
...
+
+	// Initializing Version to 0 will get the latest revision of the release.
+	Version int
+
+	//
+	StatusPrinterOptions
+}
+
+// NewGetAll creates a new GetAll object with the given configuration.
+func NewGetAll(cfg *Configuration) *GetAll {
```
liuming-dev, line 37: Why should we almostly make a copy of `pkg/action/get.go`, duplicate codes? could we try to move `GetAll ` to `pkg/action/get_all.go` ？

```yaml
type GetAll struct {
    Get
    StatusPrinterOptions
}
```

lingsamuel: Because only `get all` prints all status messages including description. Other sub-commands such as `get notes` only prints NOTES so `--show-desc` is meaningless.

cmd/helm/status.go
lingsamuel: Why do we need to tell users there is no description by prints empty string? Are we tell users there is not NOTES by prints empty string?

hickeyma (changes requested):
@lingsamuel The scope of this PR is for adding the description field to `helm status`.  I don't think that it should be extended to other operations as you have done in the last commit. Do you mind  rolling back those changes?

lingsamuel:
@hickeyma  Done

liuming-dev:
Would you mind rebasing the commits?

lingsamuel:
Is this repo enabled `squash/rebase merge`? https://docs.github.com/en/github/administering-a-repository/configuring-commit-squashing-for-pull-requests

cmd/helm/status.go
```diff
@@ -116,6 +119,9 @@ func (s statusPrinter) WriteTable(out io.Writer) error {
 	fmt.Fprintf(out, "NAMESPACE: %s\n", s.release.Namespace)
 	fmt.Fprintf(out, "STATUS: %s\n", s.release.Info.Status.String())
 	fmt.Fprintf(out, "REVISION: %d\n", s.release.Version)
+	if s.showDescription && s.release.Info.Description != "" {
```
hickeyma, line 122: I think the description should be shown regardless if empty or not. The reasoning is that someone might parse the output and expect it to be shown.

pkg/action/status.go
```diff
@@ -27,6 +27,11 @@ type Status struct {
 	cfg *Configuration
 
 	Version int
+
+	// If true, display description to output format,
+	// only affect print type table.
+	// This flag will be true by default in Helm 4.
```
hickeyma, line 33: This flag will be true by default in Helm 4. --> This flag will removed in Helm 4 and description will be output by default.

cmd/helm/status.go
```diff
@@ -40,6 +40,7 @@ The status consists of:
 - k8s namespace in which the release lives
 - state of the release (can be: unknown, deployed, uninstalled, superseded, failed, uninstalling, pending-install, pending-upgrade or pending-rollback)
 - list of resources that this release consists of, sorted by kind
+- description of the release (can be completion message or error message, need to enable --show-desc)
```
hickeyma, line 43: Ordering is slightly incorrect for output:
Move this above `list of resources that this release consists of, sorted by kind`.

Also, can you add `release revision` after `state of release`. It is missing from the list.

hickeyma (changes requested):
Thanks for the update @lingsamuel. Some small changes inline.

cmd/helm/status.go
lingsamuel: I think the user should parse the `DESCRIPTION` like `NOTES`, it's optional. Well, maybe it should be optional in Helm 4? Currently it's a explicit flag...

hickeyma: I think if `s.showDescription ` flag set then show the description even if empty

hickeyma (changes requested):
Still outstanding comment.

cmd/helm/status.go
lingsamuel: Should it be optional in Helm 4? As you say, this flag will be removed in Helm 4. Do I need to add a TODO here?

lingsamuel:
Any updates?

liuming-dev:
It will be better if you could squash these commits into one.

liuming-dev (approved):

hickeyma:
Edited title and description of the context for the PR. This is to correct where the context altered in https://github.com/helm/helm/pull/8529#issuecomment-666887998.

hickeyma (approved):
LGTM, thanks @lingsamuel for working on this.

PR #8543, opened 2020-07-31
report name length errors in linter

technosophos:

This is a bug fix.

When I rewrote linter for Helm 3.3, I missed the length check on resource names. This now produces an error if a name is longer than the prescribed 253 characters.

**If applicable**:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.13
 require (
 	github.com/BurntSushi/toml v0.3.1
 	github.com/DATA-DOG/go-sqlmock v1.4.1
+	github.com/Masterminds/goutils v1.1.0
 	github.com/Masterminds/semver/v3 v3.1.0
 	github.com/Masterminds/sprig/v3 v3.1.0
 	github.com/Masterminds/squirrel v1.4.0
--- a/pkg/action/upgrade.go
+++ b/pkg/action/upgrade.go
@@ -147,7 +147,8 @@ func validateReleaseName(releaseName string) error {
 		return errMissingRelease
 	}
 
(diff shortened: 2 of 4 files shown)


cmd/helm/testdata/helmhome/helm/repository/test-name-index.yaml
```diff
@@ -1,3 +1,3 @@
 apiVersion: v1
 entries: {}
-generated: "2020-06-23T10:01:59.2530763-07:00"
+generated: "2020-07-31T12:50:37.129802-06:00"
```
bacongobbler, line 3: this looks like a test fixture that keeps re-generating any time you run `make test`. While it's unrelated to your PR, we should fix this. Would you mind reverting this for the time being?

bacongobbler (changes requested):
code LGTM barring one change

cmd/helm/testdata/helmhome/helm/repository/test-name-index.yaml
mattfarina: This got me the other day, too. It's a test fixture that regenerates that file each time.

technosophos: That must be a new bug in that test. I will revert the file in this PR.

technosophos: Fixed. I think I will go hunt down that test fixture error, too. Tests are not supposed to overwrite fixtures.

bacongobbler (approved):

mattfarina (approved):

Issue #7430, opened 2020-01-18
Helm 3.0.2 - generating an action.Configuration for in-cluster runtime


mmailhos:
It seems to me that Helm 3 does not provide a way to create an `action.Configuration` structure if the code is running from within the cluster.

Here is a code sample that I have tried, building my own generic flags:

```
config, err := rest.InClusterConfig()
if err != nil {
    panic(err)
}
insecure := false

genericConfigFlag := &genericclioptions.ConfigFlags{
    Timeout: stringptr("0"),
    Insecure: &insecure,
    APIServer: stringptr(config.Host),
    CAFile: stringptr(config.CAFile),
    BearerToken: stringptr(config.BearerToken),
    ImpersonateGroup: &[]string{},
    Namespace: stringptr(namespace),
}

actionConfig := &action.Configuration{
    RESTClientGetter: genericConfigFlag,
    KubeClient: kube.New(genericConfigFlag),
    Log: log.Infof,
}

```

Unfortunately, this result in a `SIGSEGV` error later when running `action.NewList(actionConfig).Run()`.

Is it the right way to define an action config for Helm 3 from within a Kubernetes cluster, without altering the current library of course?

hickeyma:
@mmailhos Thanks for raising this. I am trying to understand your use case to get an understanding what you are trying to do. Do you mind providing more details about it?

I suppose you have already looked at examples of using the Helm Go library to get the `action.Configuration`, for example: https://github.com/helm/helm-2to3/blob/master/pkg/v3/connect.go

mmailhos:
@hickeyma Thank your for your reply. I'm building a simple reporting tool to describe my charts and monitore them.

The problem with the `cli` package is that it integrates with the environment and the KubeConfig while I was looking for a more _native_ solution (maybe using the k8s api, not sure). Anyway that's better as I got rid of my segfault error and now getting `context deadline exceeded`. Still troubleshooting that integration.

hickeyma:
@mmailhos This snippet of code might be useful then:

```code
kubeConfigPath := "/path/to/kubeconfig"
namespace := "<namespace>"
actionConfig := new(action.Configuration)
if err := actionConfig.Init(kube.GetConfig(kubeConfigPath, "", namespace), namespace, os.Getenv("HELM_DRIVER"), func(format string, v ...interface{}) {
		fmt.Sprintf(format, v)
}); err != nil {
		panic(err)
}
```

hickeyma:
[...]

PR #8551, opened 2020-08-03
Extend Getter Options with context.Context

easyCZ:
Extend Getter Options with context.Context and have it propagate into downstream calls. This follows recommended Go practice
where IO exposes context for cancellation and metadata propagation.

Relates to #7430

This PR is needed to be able to use Helm as a library, and be able to set context for HTTP requests.

Could also force each invocation of `Get` to supply a context (`Get(ctx, ...)`), but that would be a breaking change.

**If applicable**:

### DIFF
--- a/pkg/getter/getter.go
+++ b/pkg/getter/getter.go
@@ -18,6 +18,7 @@ package getter
 
 import (
 	"bytes"
+	"context"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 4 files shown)


easyCZ:
PTAL

bacongobbler:
Hi @easyCZ. We're working on getting Helm 2.17 out the door, working on messaging for when Helm 2 support is dropped in November, and helping out with some other efforts (like #8843) at this time. I don't think we have the bandwidth to take a look at this proposal for a while. Thank you for being patient.

easyCZ:
@bacongobbler Thanks for the information! If you think there are areas where I can contribute, let me know.

For context, at @improbable we use Helm to deploy Agones games. As such, we have a vested interest in improving Helm and helping you push forward. With that, we are happy to contribute to the project and help out in what way we can.

PR #8558, opened 2020-08-05
Update Debian repo signing key location

mattfox:
A small update to the location of the public key for the Debian repo. Now it's served from the same CDN as the indexes and packages, for speed and reliability.

The key is still available at the old location.

Signed-off-by: Matthew Fox <matt@getbalto.com>


bridgetkromhout:
This is the same change as in https://github.com/helm/helm-www/pull/762 but for the dev-v2 branch. LGTM.

bacongobbler (approved):

PR #8561, opened 2020-08-05
Detect kubeconfig as known argument in plugin invocations 

jayme-github:
With Helm 2 --kubeconfig is not detected as a known argument on plugin invocations. This means it depends on the order or arguments if helm picks it up or not which might not be under users control (for me in case of helmfile calling "helm diff").

Helm 3 has this fixed with https://github.com/helm/helm/commit/0650d6953de6267c0f20ca9014af25c5abb508f5

### DIFF
--- a/cmd/helm/load_plugins.go
+++ b/cmd/helm/load_plugins.go
@@ -137,11 +137,21 @@ func manuallyProcessArgs(args []string) ([]string, []string) {
 		}
 		return false
 	}
+
+	isKnown := func(v string) string {
+		for _, i := range kvargs {
+			if i == v {
+				return v
+			}
+		}
+		return ""
+	}
+
 	for i := 0; i < len(args); i++ {
 		switch a := args[i]; a {
 		case "--debug":
 			known = append(known, a)
-		case "--host", "--kube-context", "--home", "--tiller-namespace":
+		case isKnown(a):
 			known = append(known, a, args[i+1])
 			i++
 		default:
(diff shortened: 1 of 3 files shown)


paulczar:
@mattfarina 

would we need to add the following to https://github.com/helm/helm/blob/dev-v2/pkg/plugin/plugin.go#L196  
  
```
  if settings.KubeConfig != "" {
		envvars["KUBECONFIG"] = settings.KubeConfig
	}
```

mattfarina (changes requested):
I ran some tests. The commit referenced in the original PR adds the value from the kubeconfig to the environment. This PR does not.

There is a priority to the passed in config...
1. The flag passing in a value
2. An environment variable for it

If there is an environment variable and a flag is passed in it takes precedent. This PR removes the possibility of a flag value overriding the environment variable. Currently the plugin would have to take care of this (in Helm v3 it's handled for you) but the option is there.

I ran some tests on a plugin and I can regularly get access to the --kubeconfig flag. I understand there are sometimes it doesn't work.

We need the environment variable set to the value from the kubeconfig flag along with this.

jayme-github:
Thanks for the review. I've added setting/overriding the environment variable for plugins in case of --kubeconfig.

paulczar (approved):

bacongobbler:
We are no longer accepting bug fixes for Helm 2. See https://helm.sh/blog/helm-v2-deprecation-timeline/ for more details.

PR #8562, opened 2020-08-05
Update Common Lables template in starter chart

andytom:
Have update the Common Labels template in the starter chart so that the value for the `app.kubernetes.io/version` is set to the same value as the image tag used in the deployment.

**If applicable**:

### DIFF
--- a/pkg/chartutil/create.go
+++ b/pkg/chartutil/create.go
@@ -424,9 +424,7 @@ Common labels
 {{- define "<CHARTNAME>.labels" -}}
 helm.sh/chart: {{ include "<CHARTNAME>.chart" . }}
 {{ include "<CHARTNAME>.selectorLabels" . }}
-{{- if .Chart.AppVersion }}
-app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
-{{- end }}
+app.kubernetes.io/version: {{ .Values.image.tag | default .Chart.AppVersion | quote }}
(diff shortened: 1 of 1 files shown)


hickeyma (approved):
LGTM, thanks @andytom

mattfarina:
I moved this to 3.4.0 as it's not a bug fix.

mattfarina:
I'm not sure we should do this. The container version may not represent the application version. For example, consider WordPress using the image tag `php7.3-fpm-alpine`. This doesn't tell you the WordPress version in use which is what `AppVersion` should be. Image tag names often don't translate into application versions.

What pattern do we want to use here, why do we want it, and how should we implement it? I'd like to see a case for why.

I've added this to the next meeting and may open an issue on it after that meeting.

hickeyma:
This was discussed at todays dev weekly meeting. It was agreed that the app version and image version should be separate  because of use case like @mattfarina mentioned in https://github.com/helm/helm/pull/8562#issuecomment-683937270. The changes will therefore be reverted in #8695.

@andytom Thanks for effort you did on this PR.

andytom:
> This was discussed at todays dev weekly meeting. It was agreed that the app version and image version should be separate because of use case like @mattfarina mentioned in [#8562 (comment)](https://github.com/helm/helm/pull/8562#issuecomment-683937270). The changes will therefore be reverted in #8695.
> 
> @andytom Thanks for effort you did on this PR.

No problem, @hickeyma.

PR #8566, opened 2020-08-06
[Trival]  change lint chart name from test-release to true chart name

wawa0210:

### DIFF
--- a/pkg/lint/rules/template.go
+++ b/pkg/lint/rules/template.go
@@ -67,7 +67,7 @@ func Templates(linter *support.Linter, values map[string]interface{}, namespace 
 	}
 
 	options := chartutil.ReleaseOptions{
-		Name:      "test-release",
+		Name:      chart.Name(),
 		Namespace: namespace,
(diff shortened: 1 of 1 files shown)


hickeyma (changes requested):
@wawa0210 Thanks for pushing the PR. 

As this is the release name, `test-release` is a valid name to use. Do you mind explaining why this needs to be changed.?

wawa0210:
> @wawa0210 Thanks for pushing the PR.
> 
> As this is the release name, `test-release` is a valid name to use. Do you minf explaining why this needs to be changed.?

Thanks your reminding,  i think `test-release` is not a well know name for this, the  real chart name is a better choice

bacongobbler:
Unless there’s a specific reason that the test is failing or testing incorrect behaviour, we try not to change test cases. It’s entirely possible this name was chosen for a reason.

I’d advocate to close this PR. There’s no real reason to change this behaviour, as “test-release” is a valid release name.

PR #8580, opened 2020-08-11
Added support to pass values to sub-charts via `map.yaml`

### LINKED ISSUE #8576: Backward-compatible proposal to pass values to sub-charts by supporting value constraints
This is a proposal to extend Helm 3 in a backward compatible manner to support passing arbitrary values to subcharts without requiring developers to expose them on `values.yaml` or as globals. This is resolved as a consequence of the broader problem that this proposal addresses: add support to pass transformed/mapped values to `.Values`.

This proposal allows:

* chart developers to hint to chart users that some values are not intended to be overwriten, by removing them from `values.yaml`.
* pass values to sub-charts that are _derived_ from `values.yaml` (example below)
* improve chart's API by not asking users to use anchors or copy-pasting values passed to `helm create ...`.

In summary, this allows chart developers to have full discretion over the API that they want to expose in `values.yaml`, by hiding implementation detail currently on `values.yaml` to a different place.

# Specification

Introduce a new optional reserved file in a chart, `map.yaml`, on the root of the chart (same level as `Chart.yaml`) that:

1. is (in-memory) templated using all values passed via `values.yaml`, `--set`, etc.
2. is (in-memory) merged to `values.yaml` before all values are passed to `.Values`

when users call any `helm` commands that require rendering.

`map.yaml` uses the same syntax as any other template, and is expected to be rendered to a normal `yaml` file, like `values.yaml`. It's primary goal is to offer developers the opportunity to map/transform any value in `values.yaml` to a set of new values that are required by the chart.

For example, if `map.yaml` contains `sub.a: {{ print "%s-%s" .Values.b .Values.c }}`, `.Values.sub.a` equals `{b}-{c}` where `{b}` and `{c}` are the values passed by either `values.yaml` or `--set b= --set c=`.

When templating `map.yaml`, `.Values` is populated according to the same precedence rules of other templates (priority is given to values passed by the user over the ones in `values.yaml`).

A chart's `.Values.valueX` is populated according to the following precedence (higher takes priority):

1. `valueX` provided by the user (e.g. via `--set valueX`)
2. `valueX` rendered on `map.yaml` when that key is defined in `map.yaml` and `map.yaml` exists in the chart
3. `valueX` on `value.yaml`

This precedence applies equally to both a chart's main templates with subchart's templates. Specifically, subchart's that have a `map.yaml` use their `map.yaml` to populate their own `.Values` (together with whatever was passed by the parent chart).

# Motivation

Chart developers that use a subchart often specialize the subchart by requiring specific values to be passed to the subchart in order for the whole deployment to work as intended (a value constraint). Currently, chart developers have two ways to achieve this:

* add an entry `subchart: ...` to the chart's `values.yaml`
* refactor the subchart to use `globals` and use `globals` on `values.yaml`

These two approaches suffer from two distinct but fundamental limitations:

* `globals` are prone to naming collision across charts, both vertically (a chart with 5 sub-charts) and in depth (sub-chart of subchart of subchart).
* `subchart` requires the chart to expose all options to its user, even if some of them are not intended to be changed or break if changed. I.e. it does not allow the `subchart` to be an implementation detail of the parent chart.

The problem of passing values to sub-charts can be summarized as value constraints: a value passed to a sub-chart is often not independent, but derived from (constrained by) other values.

For simple charts, chart developers address value derivations through templates: create a named template that constructs a new value `Y` based on `.Values.X` and use the template throughout the chart. For subcharts, they do not have a mechanism to address this, and instead have to ask the chart's userto address it themselves (or use globals).

When a developer exposes a set of keys in `values.yaml` from subcharts that are related to each other, they are not doing so because the values are configurable, but because they have no mechanism to pass values to subcharts that are inter-dependent.

However, chart users rely on `values.yaml` to know how they can parameterise a chart. I.e. users only need to worry about those values and  appreciate when a chart's developer already though through which variables users need to worry about. This implies that keys in `values.yaml` that need some inter-dependency are an hassle to users, as they need to be educated (by chart developers) on which values to use and which constraints those values need to have.

Example of usage of this proposal: given a chart with

```yaml
# values.yaml
foo: ""

auth:
  app: "myapp"
```

```yaml
# map.yaml

# this chart expects that `bar` to be `test-{%foo}` (a constraint)
bar: {{ printf "test-%s" .Values.foo  }}

subchart:
  # same here: `some_app` is expected to be constructed in this specific way (a constraint)
  some_app: {{ printf "%s-%s" .Values.auth.app .Values.foo  }}
```

that a user installs with 

```bash
helm install ... --set foo="aa" --set auth.app="otherapp"
```

, the resulting `.Values` is

```yaml
foo: "aa"
bar: "test-aa"

auth:
  app: "otherapp"

subchart:
  some_app: "otherapp-aa"
```

This would be equivalent to have the chart written as

```
# values.yaml

foo: ""
bar: "test-"

auth:
  app: "myapp"

subchart:
   some_app: ""
```

and the user run

```
helm install ... --set foo="aa" --set bar="test-aa" --set auth.app="otherapp" --set "subchart.some_app=otherapp-aa"
```

Note how, in this call, the user needs to set `subchart.some_app` to be "%{auth.app}-%{foo}" for the chart to work as expected (a constraint). This is the core problem that this proposal addresses.

# Rationale

This design was selected primarily based on the following factors: 

* backward compatibility
* recursive support for sub-charts
* Clarity of chart's public API
* simplicity of usage

### backward compatibility

This design was considered due to its backward compatibility, and how it has a backward compatible migration path for chart developers. This change is backward compatible because:

1. `map.yaml` is optional and this functionality is only activated by its existence
2. ``--set subchart.valueX`` takes precedence over the values derived in `map.yaml` and thus charts can add `map.yaml` without breaking backward compatibility
3. `map.yaml` does not collide with existing templates (because it is not placed in `templates/`)

Furthermore, developers have a clear migration path from charts without `map.yaml` to charts with `map.yaml`. Specifically:

1. add `map.yaml` and to it any value that needs to be written in `.Values` that depend on any value from `values.yaml`
1. warn users in INFO that the chart plans to move to `map.yaml`, so that users no longer need to populate certain values.
1. remove entries from `values.yaml` that are now populated by `map.yaml`

They can decide when to move from one step to the next as part of their own chart's release life-cycle.

### Recursive support for sub-charts

Another reason for this design is that a sub-chart can itself have a `map.yaml`, and the precedence described in the specification works recursively. This allows charts to be self contained, as the developer of `subchart1` can decide which values to expose in `values.yaml` and which ones to derive using `map.yaml`, irrespectively of whether `subchart1` has `subchart2` or not. A chart that has subchart `subchart1` only needs to populate the values of `subchart1`'s `value.yaml`, and leave any value of `subchart2` to be populated by `subchart1`'s `map.yaml`.

### Clarity of chart's public API

So far, `values.yaml` has been a place to allow users to pass values to `.Values` and subcharts. A major reason for the choice of design in this proposal is that it effectively allows chart developers to make `values.yaml` _the public interface_ of their helm chart.

Even if a subchart would require a value `complicated_name_oauth_client_id` to be set, developers can expose that value as `oauth_client_id` and map it to `subchart.complicated_name_oauth_client_id` on `map.yaml`. Before, developers had to expose in `values.yaml` an implementation detail - that the chart uses a subchart, and users had to set `subchart.complicated_name_oauth_client_id` themselves.

### Simplicity of usage

The usage of the functionality hereby proposed is unchanged for users of helm charts: they continue to use helm exactly as before, and require no extra calls. This was a major reason for consideration of this design, as it ensures that no training is required for chart users.

For chart developers, the functionality here proposed is easy to explain: `values.yaml` contains all values that a user can change independently. All dependent (constrained, derived) values are placed in `map.yaml` so that they are available in `.Values`. For backward compatibility, keep placing those values also in `values.yaml`, so that they can continue to be passed by users using an Helm version that does not support `map.yaml`.

Another simplification of this design is that developers no longer need to write named templates to construct names derived from `.Values`. They can just write them in `map.yaml`. An example would be `componentName: {{ .Release.name }}-component` on `map.yaml`, thus allowing them to use `.Values.componentName` on the chart (instead of declaring the named template to the same effect).

This proposal considers `map.yaml` a template for two reasons:

1. because chart developers are already familiar with templates;
2. because templates provide full flexibility on the transformations they support

# Backwards Compatibility

This proposal is backward compatible.

# Reference Implementation

A proof of concept is presented in [#8580](https://github.com/helm/helm/pull/8580).

jorgecarleitao:

This PR allows developers to declare a `map.yaml` file on their chart, which can be used to map values in `values.yaml` to derived values that are used in templating, including sub-charts (see #8576 for the long explanation). This allows developers to write e.g.

```
apiVersion: v1
description: Chart with map and subcharts
name: chart-with-map
version: 0.0.1
dependencies:
  - name: backend
    version: 0.0.1
  - name: frontend
    version: 0.0.1
```

```
# values.yaml
domain: example.com
```

```
# map.yaml
backend:
  uri: {{ printf "https://api.%s" .Values.domain }}

frontend:
  uri: {{ printf "https://app.%s" .Values.domain }}

other_uri: {{ printf "https://blabla.%s" .Values.domain }}
```

thereby not having to expose `backend: uri`, `frontend: uri` in `values.yaml` (for the subchart), nor ask chart users to have to pass the same value in multiple keys for consistency (or use global names that lead to naming collisions).

I.e. it allows subcharts to be populated with derived (or `map`ed) values without exposing these values to `values.yaml` (the public interface of the chart).

This solves a long standing issue of being very painful to pass values to subcharts that derive values in `values.yaml` (more details in #8576). Closes #8576.

**Special notes for your reviewer**: This PR is not complete yet: this is part of the proposal, and is intended to demonstrate how the proposal is implemented in practice. There are at least three issues to address, two documented as `# todo`, plus   associated documentation.

To run the example,

```
make build
./bin/helm install test --debug --dry-run cmd/helm/testdata/testcharts/chart-with-map/
```

which leads to the manifest

```
---
# Source: chart-with-map/charts/subchart/templates/test1.yaml
apiVersion: v1
kind: Pod
metadata:
    name: "aaa"
    labels:
        derived: aa-a
        default: aa
---
# Source: chart-with-map/templates/test1.yaml
apiVersion: v1
kind: Pod
metadata:
    name: "aaa"
    labels:
        derived: a-a
        public1: a
```

**If applicable**:

### DIFF
--- a/cmd/helm/install_test.go
+++ b/cmd/helm/install_test.go
@@ -201,6 +201,12 @@ func TestInstall(t *testing.T) {
 			name: "install chart with only crds",
 			cmd:  "install crd-test testdata/testcharts/chart-with-only-crds --namespace default",
 		},
+		// Install, base case with map and subcharts
+		{
+			name:   "basic install with a map and subcharts",
+			cmd:    "install aeneas testdata/testcharts/chart-with-map --namespace default",
+			golden: "output/install.txt",
+		},
 	}
 
 	runTestActionCmd(t, tests)
--- a/cmd/helm/testdata/testcharts/chart-with-map/Chart.yaml
+++ b/cmd/helm/testdata/testcharts/chart-with-map/Chart.yaml
@@ -1,0 +1,7 @@
+apiVersion: v1
+description: Chart with map and subcharts
+name: chart-with-map
+version: 0.0.1
+dependencies:
+  - name: subchart
+    version: 0.0.1
--- a/cmd/helm/testdata/testcharts/chart-with-map/charts/subchart/Chart.yaml
+++ b/cmd/helm/testdata/testcharts/chart-with-map/charts/subchart/Chart.yaml
@@ -1,0 +1,4 @@
+apiVersion: v1
+description: subchart
+name: subchart
+version: 0.0.1
--- a/cmd/helm/testdata/testcharts/chart-with-map/charts/subchart/map.yaml
+++ b/cmd/helm/testdata/testcharts/chart-with-map/charts/subchart/map.yaml
@@ -1,0 +1,1 @@
+derived: {{ printf "%s-%s" .Values.default "a" }}
--- a/cmd/helm/testdata/testcharts/chart-with-map/charts/subchart/templates/test1.yaml
+++ b/cmd/helm/testdata/testcharts/chart-with-map/charts/subchart/templates/test1.yaml
@@ -1,0 +1,7 @@
+apiVersion: v1
+kind: Pod
+metadata:
+    name: "aaa"
+    labels:
+        derived: {{ .Values.derived }}
+        default: {{ .Values.default }}
--- a/cmd/helm/testdata/testcharts/chart-with-map/charts/subchart/values.yaml
+++ b/cmd/helm/testdata/testcharts/chart-with-map/charts/subchart/values.yaml
@@ -1,0 +1,1 @@
+default: "subchart-default"
--- a/cmd/helm/testdata/testcharts/chart-with-map/map.yaml
+++ b/cmd/helm/testdata/testcharts/chart-with-map/map.yaml
@@ -1,0 +1,4 @@
+derived: {{ printf "%s-%s" .Values.public1 "a" }}
+
+subchart:
+  default: "aa"
--- a/cmd/helm/testdata/testcharts/chart-with-map/templates/test1.yaml
+++ b/cmd/helm/testdata/testcharts/chart-with-map/templates/test1.yaml
@@ -1,0 +1,7 @@
+apiVersion: v1
+kind: Pod
+metadata:
+    name: "aaa"
+    labels:
+        derived: {{ .Values.derived }}
+        public1: {{ .Values.public1 }}
--- a/cmd/helm/testdata/testcharts/chart-with-map/values.yaml
+++ b/cmd/helm/testdata/testcharts/chart-with-map/values.yaml
@@ -1,0 +1,1 @@
+public1: a
--- a/pkg/chart/chart.go
+++ b/pkg/chart/chart.go
@@ -42,6 +42,8 @@ type Chart struct {
 	Templates []*File `json:"templates"`
 	// Values are default config for this chart.
 	Values map[string]interface{} `json:"values"`
+	// Map are default derivations of values of this chart.
+	Map *File `json:"map"`
 	// Schema is an optional JSON schema for imposing structure on Values
 	Schema []byte `json:"schema"`
 	// Files are miscellaneous files in a chart archive,
--- a/pkg/chart/loader/load.go
+++ b/pkg/chart/loader/load.go
@@ -99,6 +99,8 @@ func LoadFiles(files []*BufferedFile) (*chart.Chart, error) {
 			if err := yaml.Unmarshal(f.Data, &c.Values); err != nil {
 				return c, errors.Wrap(err, "cannot load values.yaml")
 			}
+		case f.Name == "map.yaml":
+			c.Map = &chart.File{Name: f.Name, Data: f.Data}
 		case f.Name == "values.schema.json":
 			c.Schema = f.Data
 
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -28,6 +28,7 @@ import (
 
 	"github.com/pkg/errors"
 	"k8s.io/client-go/rest"
+	"sigs.k8s.io/yaml"
 
 	"helm.sh/helm/v3/pkg/chart"
 	"helm.sh/helm/v3/pkg/chartutil"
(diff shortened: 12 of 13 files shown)


TBBle:
As a note, by doing this work in `recAllTpls`, `import-values` won't be able to see the data from _map.yaml_ as `chartutil.ProcessDependencies` calls `processImportValues` much earlier in the flow.

bacongobbler:
Hi @jorgecarleitao. We're asking new proposals to first write up a Helm Improvement Proposal, also known as a HIP. This gives the maintainers an opportunity to consider the design of the proposal and iron out any design kinks before a reference implementation is provided. This saves both you and us time in the long run.

Please have a read through https://github.com/helm/community/blob/master/hips/hip-0001.md and let me know if you have any questions.

bacongobbler (changes requested):
Hi @jorgecarleitao. We're asking new proposals to first write up a Helm Improvement Proposal, also known as a HIP. This gives the maintainers an opportunity to consider the design of the proposal and iron out any design kinks before a reference implementation is provided. This saves both you and us time in the long run.

Please have a read through https://github.com/helm/community/blob/master/hips/hip-0001.md and let me know if you have any questions.

m-yosefpor:
any progress on this?

bacongobbler:
No further progress has been made since chnages were requested 12 months ago. Closing as stale.

agates4:
@jorgecarleitao please don't give up. :)

treksler:
Expecting anyone who is capable and willing to write code (for free) to write an HIP document instead is nuts. Like 100% nuts. If you have a problem where people are writing code that seems like it is coming out of left field, then you have lost touch with your user base, not the other way around.

PR #8590, opened 2020-08-12
release/mock: Ensure conversion from int to string yields a string

Foxboron:
With the release of go 1.15, the test-suite doesn't pass as `go test` got
a new warning for improper `string(x)` usage.

https://golang.org/doc/go1.15#vet

    $ make test-unit
    # helm.sh/helm/v3/pkg/release
    pkg/release/mock.go:56:27: conversion from int to string yields a string of one rune, not a string of digits (did you mean fmt.Sprint(x)?)
    [snip]
    make: *** [Makefile:82: test-unit] Error 2

This patch changes ensures we are utilizing `fmt.Sprint` instead as
recommended.

### DIFF
--- a/pkg/release/mock.go
+++ b/pkg/release/mock.go
@@ -17,6 +17,7 @@ limitations under the License.
 package release
 
 import (
+	"fmt"
 	"math/rand"
 
 	"helm.sh/helm/v3/pkg/chart"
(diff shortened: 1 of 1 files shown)


mattfarina (commented):
Do you have an example where the conversion to a string does not occur so that this change is needed?

bacongobbler (approved):
>  For example, string(9786) does not evaluate to the string "9786"; it evaluates to the string "\xe2\x98\xba", or "☺".

I wrote a test program to confirm this behaviour:

```
><> cat main.go
package main

import (
        "fmt"
)

func main() {
        fmt.Println(string(9786))
}
><> go version
go version go1.15 linux/amd64
><> go run main.go
☺
```

Changing `string(9786)` to `fmt.Sprint(9786)` returns the expected value:

```
><> go run main.go
9786
```

PR #8608, opened 2020-08-18
Reduce linting severity for users of out-of-date kubernetes

### LINKED ISSUE #8596: Helm lint throws an error where a warning should be thrown instead
As a result of these 2 PRs:

1. **https://github.com/helm/helm/pull/7986 - feat: implement deprecation warnings in helm lint**
1. **https://github.com/helm/helm/pull/8011 - feat: lint the names of templated resources**

now `helm lint` validates  whether resource `metadata.name` complies Kubernetes naming rules and whether the `apiVersion` is up to date. And these are great features, thank you for implementing them! 

However, these rules are having ERROR severity instead of WARNING. These should be warnings because these are NOT errors from the Kubernetes perspective - Kubernetes API still accepts resources with deprecated API versions as well as resources with metadata names that, for example, contain `:` in them. If these are warnings instead of errors, then it will be up to a client to decide whether he wants to fail lint (via using `--strict` flag for `helm lint` command) or not.

Setting the error severity level for these issues also complicates the migration of automated deployments 
 to helm 3.3, especially form Helm 2.x.

_How to fix?_ - I believe [these lines](https://github.com/technosophos/k8s-helm/blob/2d1ce6d7866e1edc2431eb637864f0bdc7e6a0a7/pkg/lint/rules/template.go#L133-L134) of code should be rewritten to:

```go
linter.RunLinterRule(support.WarningSev, path, validateMetadataName(&yamlStruct))
linter.RunLinterRule(support.WarningSev, path, validateNoDeprecations(&yamlStruct))
```

joejulian:
Reduce linting severity for users of out-of-date kubernetes, closes #8596

**If applicable**:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,8 +56,9 @@ LDFLAGS += -X helm.sh/helm/v3/internal/version.gitTreeState=${GIT_DIRTY}
 LDFLAGS += $(EXT_LDFLAGS)
 
 # Define constants based on the client-go version
-K8S_MODULES_MAJOR_VER=$(shell awk '/client-go/{s=$$2;sub("^v","",s);split(s,v,".");print v[1]+1}' go.mod)
-K8S_MODULES_MINOR_VER=$(shell awk '/client-go/{s=$$2;sub("^v","",s);split(s,v,".");print v[2]}' go.mod)
+K8S_MODULES_VER=$(subst ., ,$(subst v,,$(shell go list -f '{{.Version}}' -m k8s.io/client-go)))
+K8S_MODULES_MAJOR_VER=$(shell echo $$(($(firstword $(K8S_MODULES_VER)) + 1)))
+K8S_MODULES_MINOR_VER=$(word 2,$(K8S_MODULES_VER))
 
 LDFLAGS += -X helm.sh/helm/v3/pkg/lint/rules.k8sVersionMajor=$(K8S_MODULES_MAJOR_VER)
 LDFLAGS += -X helm.sh/helm/v3/pkg/lint/rules.k8sVersionMinor=$(K8S_MODULES_MINOR_VER)
--- a/Makefile
+++ b/Makefile
@@ -55,6 +55,10 @@ LDFLAGS += -X helm.sh/helm/v3/internal/version.gitCommit=${GIT_COMMIT}
 LDFLAGS += -X helm.sh/helm/v3/internal/version.gitTreeState=${GIT_DIRTY}
 LDFLAGS += $(EXT_LDFLAGS)
 
+# Define constants for the client-go version
+LDFLAGS += -X helm.sh/helm/v3/pkg/lint/rules.k8sVersionMajor=$(shell awk '/client-go/{s=$$2;sub("^v","",s);split(s,v,".");print v[1]+1}' go.mod)
+LDFLAGS += -X helm.sh/helm/v3/pkg/lint/rules.k8sVersionMinor=$(shell awk '/client-go/{s=$$2;sub("^v","",s);split(s,v,".");print v[2]}' go.mod)
+
 .PHONY: all
 all: build
 
--- a/Makefile
+++ b/Makefile
@@ -55,9 +55,14 @@ LDFLAGS += -X helm.sh/helm/v3/internal/version.gitCommit=${GIT_COMMIT}
 LDFLAGS += -X helm.sh/helm/v3/internal/version.gitTreeState=${GIT_DIRTY}
 LDFLAGS += $(EXT_LDFLAGS)
 
-# Define constants for the client-go version
-LDFLAGS += -X helm.sh/helm/v3/pkg/lint/rules.k8sVersionMajor=$(shell awk '/client-go/{s=$$2;sub("^v","",s);split(s,v,".");print v[1]+1}' go.mod)
-LDFLAGS += -X helm.sh/helm/v3/pkg/lint/rules.k8sVersionMinor=$(shell awk '/client-go/{s=$$2;sub("^v","",s);split(s,v,".");print v[2]}' go.mod)
+# Define constants based on the client-go version
+K8S_MODULES_MAJOR_VER=$(shell awk '/client-go/{s=$$2;sub("^v","",s);split(s,v,".");print v[1]+1}' go.mod)
+K8S_MODULES_MINOR_VER=$(shell awk '/client-go/{s=$$2;sub("^v","",s);split(s,v,".");print v[2]}' go.mod)
+
+LDFLAGS += -X helm.sh/helm/v3/pkg/lint/rules.k8sVersionMajor=$(K8S_MODULES_MAJOR_VER)
+LDFLAGS += -X helm.sh/helm/v3/pkg/lint/rules.k8sVersionMinor=$(K8S_MODULES_MINOR_VER)
+LDFLAGS += -X helm.sh/helm/v3/pkg/chartutil.k8sVersionMajor=$(K8S_MODULES_MAJOR_VER)
+LDFLAGS += -X helm.sh/helm/v3/pkg/chartutil.k8sVersionMinor=$(K8S_MODULES_MINOR_VER)
 
 .PHONY: all
 all: build
--- a/pkg/chartutil/capabilities.go
+++ b/pkg/chartutil/capabilities.go
@@ -29,7 +29,7 @@ import (
 
 const (
 	k8sVersionMajor = 1
-	k8sVersionMinor = 19
+	k8sVersionMinor = 20
 )
 
 var (
--- a/pkg/chartutil/capabilities.go
+++ b/pkg/chartutil/capabilities.go
@@ -16,12 +16,20 @@ limitations under the License.
 package chartutil
 
 import (
+	"fmt"
+	"strconv"
+
 	"k8s.io/client-go/kubernetes/scheme"
 
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apiextensionsv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 
 	helmversion "helm.sh/helm/v3/internal/version"
+)
+
+const (
+	k8sMajorVer = 1
+	k8sMinorVer = 19
 )
 
 var (
@@ -31,9 +39,9 @@ var (
 	// DefaultCapabilities is the default set of capabilities.
 	DefaultCapabilities = &Capabilities{
 		KubeVersion: KubeVersion{
-			Version: "v1.18.0",
-			Major:   "1",
-			Minor:   "18",
+			Version: fmt.Sprintf("v%d.%d.0", k8sMajorVer, k8sMinorVer),
+			Major:   strconv.Itoa(k8sMajorVer),
+			Minor:   strconv.Itoa(k8sMinorVer),
 		},
 		APIVersions: DefaultVersionSet,
 		HelmVersion: helmversion.Get(),
--- a/pkg/chartutil/capabilities.go
+++ b/pkg/chartutil/capabilities.go
@@ -28,8 +28,8 @@ import (
 )
 
 const (
-	k8sMajorVer = 1
-	k8sMinorVer = 19
+	k8sVersionMajor = 1
+	k8sVersionMinor = 19
 )
 
 var (
@@ -39,9 +39,9 @@ var (
 	// DefaultCapabilities is the default set of capabilities.
 	DefaultCapabilities = &Capabilities{
 		KubeVersion: KubeVersion{
-			Version: fmt.Sprintf("v%d.%d.0", k8sMajorVer, k8sMinorVer),
-			Major:   strconv.Itoa(k8sMajorVer),
-			Minor:   strconv.Itoa(k8sMinorVer),
+			Version: fmt.Sprintf("v%d.%d.0", k8sVersionMajor, k8sVersionMinor),
+			Major:   strconv.Itoa(k8sVersionMajor),
+			Minor:   strconv.Itoa(k8sVersionMinor),
 		},
 		APIVersions: DefaultVersionSet,
 		HelmVersion: helmversion.Get(),
--- a/pkg/lint/rules/deprecations.go
+++ b/pkg/lint/rules/deprecations.go
@@ -29,7 +29,7 @@ const (
 	// This should be set in the Makefile based on the version of client-go being imported.
 	// These constants will be overwritten with LDFLAGS
 	k8sVersionMajor = 1
-	k8sVersionMinor = 19
+	k8sVersionMinor = 20
 )
 
 // deprecatedAPIError indicates than an API is deprecated in Kubernetes
--- a/pkg/lint/rules/deprecations.go
+++ b/pkg/lint/rules/deprecations.go
@@ -18,50 +18,176 @@ package rules // import "helm.sh/helm/v3/pkg/lint/rules"
 
 import "fmt"
 
-// deprecatedAPIs lists APIs that are deprecated (left) with suggested alternatives (right).
-//
-// An empty rvalue indicates that the API is completely deprecated.
-var deprecatedAPIs = map[string]string{
-	"extensions/v1beta1 Deployment":                             "apps/v1 Deployment",
-	"extensions/v1beta1 DaemonSet":                              "apps/v1 DaemonSet",
-	"extensions/v1beta1 ReplicaSet":                             "apps/v1 ReplicaSet",
-	"extensions/v1beta1 PodSecurityPolicy":                      "policy/v1beta1 PodSecurityPolicy",
-	"extensions/v1beta1 NetworkPolicy":                          "networking.k8s.io/v1beta1 NetworkPolicy",
-	"extensions/v1beta1 Ingress":                                "networking.k8s.io/v1beta1 Ingress",
-	"apps/v1beta1 Deployment":                                   "apps/v1 Deployment",
-	"apps/v1beta1 StatefulSet":                                  "apps/v1 StatefulSet",
-	"apps/v1beta1 ReplicaSet":                                   "apps/v1 ReplicaSet",
-	"apps/v1beta2 Deployment":                                   "apps/v1 Deployment",
-	"apps/v1beta2 StatefulSet":                                  "apps/v1 StatefulSet",
-	"apps/v1beta2 DaemonSet":                                    "apps/v1 DaemonSet",
-	"apps/v1beta2 ReplicaSet":                                   "apps/v1 ReplicaSet",
-	"apiextensions.k8s.io/v1beta1 CustomResourceDefinition":     "apiextensions.k8s.io/v1 CustomResourceDefinition",
-	"rbac.authorization.k8s.io/v1alpha1 ClusterRole":            "rbac.authorization.k8s.io/v1 ClusterRole",
-	"rbac.authorization.k8s.io/v1alpha1 ClusterRoleList":        "rbac.authorization.k8s.io/v1 ClusterRoleList",
-	"rbac.authorization.k8s.io/v1alpha1 ClusterRoleBinding":     "rbac.authorization.k8s.io/v1 ClusterRoleBinding",
-	"rbac.authorization.k8s.io/v1alpha1 ClusterRoleBindingList": "rbac.authorization.k8s.io/v1 ClusterRoleBindingList",
-	"rbac.authorization.k8s.io/v1alpha1 Role":                   "rbac.authorization.k8s.io/v1 Role",
-	"rbac.authorization.k8s.io/v1alpha1 RoleList":               "rbac.authorization.k8s.io/v1 RoleList",
-	"rbac.authorization.k8s.io/v1alpha1 RoleBinding":            "rbac.authorization.k8s.io/v1 RoleBinding",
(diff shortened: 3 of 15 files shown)


alejandroEsc (approved):
thank you, this should not be a blocking error, or should be optional somehow, but a warning is probably the safest approach.

kofan:
If a client decides, he can always use `--strict` flag to convert warnings to errors

bacongobbler:
Given that @technosophos wrote these PRs, it might be best if he can weigh in here. I know he's out for today though. Otherwise I'll defer the decision to the other core maintainers who reviewed these PRs (@mattfarina and @hickeyma). :)

mattfarina:
I want to talk with @technosophos about this first.

technosophos:
I mean, the obvious point is that the intention was to PREVENT charts from using deprecated APIs that were removed. This downgrades to only warning them (by default). If we feel like that's the best service to our users, then it's fine with me if we merge it.

I don't think the "if they want errors they can use strict" argument is valid, because that turns ALL warnings to errors. But again, if we really feel like a warning is better than an error here, that's fine.

joejulian:
If that were the result, I think many of us would be ok with that. Unfortunately the result that brought this to our attention is that  `rbac.authorization.k8s.io/v1beta1` has not been removed and won't be until 1.20, but is still throwing errors.

Additionally, starting with 1.19, support is extending to cover the release branches for one year following the initial release of a new minor version instead of N-2 so this may get even more controversial. :/

technosophos:
If that's the _only_ reason, I would be totally fine with removing that particular resource type. And maybe it makes sense to have warnings for those with future deprecations, and errors for ones that have been removed altogether.

Lint has always been a bit of a wildcard because it is a combination of opinions and facts. "You shouldn't do X" versus "You cannot do X." In this particular case, it appears that we have accidentally conflated the two into the same linter. It makes sense to me to produce an error when a chart contains a kind that Kubernetes does not support anymore. But you are right -- a warning is a better experience if a thing _will be removed in the future_, but is still usable in today's Kubernetes.

hickeyma (changes requested):
Thanks for pushing the PR @joejulian.

I think it would be very useful to the user to show additional information like what Kubernetes version the API is deprecated in and more importantly what version the API is removed in. This would require refactoring the map in https://github.com/helm/helm/blob/master/pkg/lint/rules/deprecations.go#L24 and the subsequent function: https://github.com/helm/helm/blob/master/pkg/lint/rules/deprecations.go#L71. The information on the APIs can be garnered from: https://github.com/hickeyma/helm-mapkubeapis/blob/master/config/Map.yaml.

This would enable the user to then better understand when updates are needed to the chart for the relevant Kubernetes version.

joejulian:
```
WARNING] templates/deployment.yaml: the kind "apps/v1beta2 Deployment" is deprecated as of "v1.9" and removed in "v1.16" in favor of "apps/v1 Deployment"
```

joejulian:
Don't merge yet. A [similar feature](https://github.com/kubernetes/enhancements/tree/master/keps/sig-api-machinery/1693-warnings#graduation-criteria) was added to kubernetes. I'd like to see about using their definitions instead of hard-coding them.

hickeyma:
Set it to draft while you investigate this further @joejulian. Nice catch on the new feature. Just wonder if this is available only from 1.19?

pkg/lint/rules/deprecations.go
```diff
@@ -18,50 +18,176 @@ package rules // import "helm.sh/helm/v3/pkg/lint/rules"
...
+		NewAPI:       "rbac.authorization.k8s.io/v1 RoleBinding",
+		DeprecatedIn: "v1.17",
+		RemovedIn:    "v1.22",
+	},
+	"rbac.authorization.k8s.io/v1alpha1 RoleBindingList": {
+		NewAPI:       "rbac.authorization.k8s.io/v1 RoleBindingList",
+		DeprecatedIn: "v1.17",
+		RemovedIn:    "v1.22",
+	},
+	"rbac.authorization.k8s.io/v1beta ClusterRole": {
```
hickeyma, line 139: v1beta --> v1beta1

hickeyma (changes requested):
Thanks for pushing the PR @joejulian. It is looking good bar a fix on one of the apis mentioned inline.

joejulian:
I don't know how I did that. I didn't intend to convert it from draft.

joejulian:
```
helm lint .
==> Linting .
[WARNING] templates/deployment.yaml: extensions/v1beta1 Deployment is deprecated in v1.8+, unavailable in v1.18+; use apps/v1 Deployment

```

A future improvement could be to allow a flag to set the version of k8s you wish to test against, and set the severity based on that flag. That's outside the scope of what I have time to contribute with this iteration.

go.mod
```diff
@@ -35,13 +35,14 @@ require (
 	github.com/stretchr/testify v1.6.1
 	github.com/xeipuuv/gojsonschema v1.2.0
 	golang.org/x/crypto v0.0.0-20200622213623-75b288015ac9
-	k8s.io/api v0.18.8
+	k8s.io/api v0.19.0
 	k8s.io/apiextensions-apiserver v0.18.8
```
hickeyma, line 39: why not use v0.19.0?

pkg/lint/rules/deprecations.go
```diff
@@ -16,65 +16,64 @@ limitations under the License.
...
-	msg := fmt.Sprintf("the kind %q is deprecated", e.Deprecated)
-	if e.Alternative != "" {
-		msg += fmt.Sprintf(" in favor of %q", e.Alternative)
-	}
+	msg := e.Message
 	return msg
 }
 
 func validateNoDeprecations(resource *K8sYamlStruct) error {
+	// resource is not a resource
```
hickeyma, line 46: What do you mean by this and why need to check?

go.mod
hickeyma, line 39: Why not up to 0.19?

hickeyma (commented):
Thanks for the effort on this @joejulian, especially around researching the new deprecation messaging from k8s. It is looking good bar some small things commented inline.

pkg/lint/rules/deprecations.go
joejulian: https://www.github.com/helm/helm/tree/master/pkg%2Flint%2Frules%2Ftestdata%2Fgoodone%2Ftemplates%2Fgoodone.yaml

This was failing tests because it has no gvk. This check prevents that failure based on the assumption that that's actually a "goodone" that should not fail.

go.mod
joejulian: Missed it. Done.

pkg/lint/rules/deprecations.go
joejulian: cc @technosophos

joejulian:
Tested this against helm/charts/stable and everything passes. Plenty of warnings, but no errors.

pkg/lint/rules/deprecations.go
technosophos: Is "resource is not a resource" just "resource is not properly formatted"?

technosophos (commented):
My only concern is that this appears to require developers to manually update a constant field each time the Kubernetes library iterates. I would put money on that field being neglected frequently. Do we have any ideas for automating that?

This PR also appears to go from k8s 1.18 to 1.19. Has anyone run "the usual manual tests" to make sure that this works with our usual backwards compatibility assurances?

Otherwise, I am :100: on board with this PR.

pkg/lint/rules/deprecations.go
```diff
@@ -16,65 +16,68 @@ limitations under the License.
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apiserver/pkg/endpoints/deprecation"
+	kscheme "k8s.io/client-go/kubernetes/scheme"
+)
+
+const (
+	// This should be set based on the version of client-go being imported
+	k8sVersionMajor = 1
+	k8sVersionMinor = 19
```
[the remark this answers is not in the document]
bacongobbler: The same functionality can be found under https://github.com/helm/helm/blob/6898ad14576463ea1df857bb17f2f0ee47653756/pkg/chartutil/capabilities.go#L34-L36, and will need to be updated as well.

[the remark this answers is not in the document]
joejulian: My _wish_ is that this could be gleaned from the client-go library. Short of that it could be defined in the Makefile in LDFLAGS, maybe even by parsing go.mod. Is that the preference?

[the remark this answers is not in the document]
bacongobbler: Less maintenance hassle is always preferred, yes. :)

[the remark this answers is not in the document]
joejulian: Using LDFLAGS per slack conversation

pkg/lint/rules/deprecations.go
joejulian: Ok, if that yaml is supposed to be a valid resource, it's not. This should probably be determined in a _different_ validation rule than the deprecation validation. I'd like to keep this here to prevent a deprecation warning for an invalid gkv.

If there needs to be a gkv validation, that should be a separate issue.

joejulian: I updated the comment to make it clearer

Makefile
```diff
@@ -50,6 +50,10 @@ LDFLAGS += -X helm.sh/helm/v3/internal/version.metadata=${VERSION_METADATA}
 LDFLAGS += -X helm.sh/helm/v3/internal/version.gitCommit=${GIT_COMMIT}
 LDFLAGS += -X helm.sh/helm/v3/internal/version.gitTreeState=${GIT_DIRTY}
 
+# Define constants for the client-go version
+LDFLAGS += -X helm.sh/helm/v3/pkg/lint/rules.k8sVersionMajor=$(shell awk '/client-go/{s=$$2;sub("^v","",s);split(s,v,".");print v[1]+1}' go.mod)
+LDFLAGS += -X helm.sh/helm/v3/pkg/lint/rules.k8sVersionMinor=$(shell awk '/client-go/{s=$$2;sub("^v","",s);split(s,v,".");print v[2]}' go.mod)
```
technosophos, line 55: Ah! This is clever! I like it!

technosophos (approved):

technosophos:
When @adamreese is back from his time off, I will pair with him to run through the migration checklist for using a new version of Kubernetes libraries. Since @hickeyma has made a change request, he also needs to sign off on this. But as it stands now, I think this PR is in good shape.

Maintainers: *Don't merge until @adamreese has reviewed this*

Makefile
bacongobbler: Can we do the same for [capabilities,go](https://github.com/helm/helm/blob/6898ad14576463ea1df857bb17f2f0ee47653756/pkg/chartutil/capabilities.go#L34-L36) too? It would be wonderful to no longer need to update these variables.

joejulian: done(ish). There's still a bunch of fixed-strings in yaml that is used for validating results. I'm hesitant to do anything about that since the goal is to have a fixed target for the tests.

technosophos:
Uh-oh... the branch seems to be broken now.

pkg/lint/rules/deprecations.go
```diff
@@ -16,65 +16,69 @@ limitations under the License.
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apiserver/pkg/endpoints/deprecation"
+	kscheme "k8s.io/client-go/kubernetes/scheme"
+)
+
+const (
+	// This should be set in the Makefile based on the version of client-go being imported.
+	// These constants will be overwritten with LDFLAGS
+	k8sVersionMajor = 1
+	k8sVersionMinor = 19
```
hickeyma, line 32: Should this be picked up by the LDFLAGS  now?

hickeyma (changes requested):
@joejulian Looks like still using version instead of LDFLAGS .

pkg/lint/rules/deprecations.go
joejulian: It's overridden by LDFLAGS.

Leaving in constants that will be overridden allows VSCode debugger to work (albeit using the constant values instead of the overrides) so I always leave them in out of convenience.

hickeyma: ok

hickeyma (approved):
LGTM, thanks @joejulian

adamreese (approved):
We should remove the dependency on awk in the Makefile but I won't hold this up. LGTM

technosophos:
Helm maintainers: I had to artificially delay merging this in order to get this week's security release out. But with that done, this is clear to merge as soon as you see fit.

hickeyma (commented):
@joejulian Thanks for all the effort and perseverance on this PR. While doing some more manual testing on this before merging, I saw this output:

```console
$ helm3 lint pr-7986
==> Linting pr-7986
[WARNING] templates/deployment.yaml: extensions/v1beta1 Deployment is deprecated in v1.8+, unavailable in v1.18+; use apps/v1 Deployment

$ more pr-7986/templates/deployment.yaml
kind: Deployment
metadata:
  name: {{ include "pr-7986.fullname" . }}
  labels:
    {{- include "pr-7986.labels" . | nindent 4 }}
spec:
[...] 
```

While it correctly identifies the removed API and suggests the supported version, the Kubernetes version that it is removed is incorrect. It was removed in 1.16. See https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/ for more details.

hickeyma (changes requested):
Can we hold back on merging until we resolve the issue as described in https://github.com/helm/helm/pull/8608#pullrequestreview-492521367.

joejulian:
> @joejulian Thanks for all the effort and perseverance on this PR. While doing some more manual testing on this before merging, I saw this output:
> 
> ```
> $ helm3 lint pr-7986
> ==> Linting pr-7986
> [INFO] Chart.yaml: icon is recommended
> [WARNING] templates/deployment.yaml: extensions/v1beta1 Deployment is deprecated in v1.8+, unavailable in v1.18+; use apps/v1 Deployment
> 
> 1 chart(s) linted, 0 chart(s) failed
> 
> $ more pr-7986/templates/deployment.yaml
> apiVersion: extensions/v1beta1
> kind: Deployment
> metadata:
>   name: {{ include "pr-7986.fullname" . }}
>   labels:
>     {{- include "pr-7986.labels" . | nindent 4 }}
> spec:
> [...] 
> ```
> 
> While it correctly identifies the removed API and suggests the supported version, the Kubernetes version that it is removed is incorrect. It was removed in 1.16. See https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/ for more details.

This looks to be an upstream bug. https://github.com/kubernetes/api/blob/52bf62410745c20f2be50855c3e8c88d08292f85/extensions/v1beta1/types.go#L79
```
// +k8s:prerelease-lifecycle-gen:removed=1.18
```
This code is working correctly. Does this need to be blocked until upstream gets fixed?

hickeyma:
It is upstream alright. I have raised following issue for it: https://github.com/kubernetes/kubernetes/issues/94933

caarlos0:
FWIW, kubebuilder still generates CRD with the deprecated api version, so it is often impractical to migrate from v1beta1 to v1.

https://github.com/kubernetes-sigs/kubebuilder/issues/1065

hickeyma:
@joejulian Even though the code is working and the issue is with the versions defined upstream, it will ultimately give incorrect information to the user. I don't think that this is best practices and a good thing to do. @helm/helm-core-maintainers what do you think?

hickeyma:
Kubernetes upstream issue kubernetes/kubernetes#94933now resolved. Waiting on cherry pick but it could be the 1.20 release.

joejulian:
cherry-pick https://github.com/kubernetes/kubernetes/pull/94973
1.19.3 release slated for 2020-10-14. Will bump k8s libraries when that happens.

joejulian:
@hickeyma Our upstream changes finally got published. This is ready to go.

hickeyma:
Thanks @joejulian. I will review this when I get a chance.

hickeyma (approved):
LGTM, thanks for the work on this @joejulian

hickeyma:
Need approval from another maintainer before merging. @adamreese @technosophos do you mind doing a review again not that the new k8s libs have been added?

pkg/chartutil/capabilities.go
```diff
@@ -24,16 +27,21 @@ import (
 	helmversion "helm.sh/helm/v3/internal/version"
 )
 
+const (
+	k8sMajorVer = 1
+	k8sMinorVer = 19
```
adamreese, line 32: LDFLAGS uses the variable name `k8sVersionMajor` and `k8sVersionMinor`

pkg/lint/rules/deprecations_test.go
```diff
@@ -27,10 +27,9 @@ func TestValidateNoDeprecations(t *testing.T) {
 	if err == nil {
 		t.Fatal("Expected deprecated extension to be flagged")
 	}
-
 	depErr := err.(deprecatedAPIError)
-	if depErr.Alternative != "apps/v1 Deployment" {
-		t.Errorf("Expected %q to be replaced by %q", depErr.Deprecated, depErr.Alternative)
+	if depErr.Message == "" {
+		t.Fatal("Expected error message to be non-blank")
```
adamreese, line 32: Including the error returned would be useful

adamreese (changes requested):

pkg/chartutil/capabilities.go
joejulian: done

pkg/lint/rules/deprecations_test.go
joejulian: Thanks, @adamreese . Done.

allen-servedio:
So it looks like this got some conflicts after not being merged for the last release. Is there any way we can get the conflicts resolved and this released? Thank you for all of your efforts on this @joejulian  to address the reviewers enhancement suggestions.

joejulian:
Sure, I can rebase. I've done that a few times now. Hopefully this can get merged this time.

mattfarina:
@joejulian Two quick things about this...

1. Can you update the Kubernetes version to 1.20 instead of 1.19? That's all I'd ask before approving.
2. As a follow-up for Helm 3.6 we could look at passing the kubernetes version to `helm lint`. Just as an idea.

joejulian:
Thanks. Bumped those constants (as an aside, those constants are placeholders where the values are set in the Makefile). I've also added a validation script to ensure that future k8s version bumps include changes to those constants.

joejulian:
>     2. As a follow-up for Helm 3.6 we could look at passing the kubernetes version to `helm lint`. Just as an idea

It was on my mind to add this already. I'll create an issue.

mattfarina (approved):

joejulian:
> Thanks. Bumped those constants (as an aside, those constants are placeholders where the values are set in the Makefile). I've also added a validation script to ensure that future k8s version bumps include changes to those constants.

In talking with @mattfarina on slack, we decided to pull out the validation script and do it in a separate PR to simplify review.

mattfarina:
Note, I asked @joejulian to resolve the conflicts in the mod/sum files.

bacongobbler:
My one concern with this PR has to do with the Kubernetes version update from 1.18 to 1.20.

To clarify, the only reason this PR bumps the kubernetes version is to take advantage of a new client library to detect deprecated APIs. Is that correct?

The reason I'm concerned is that bumping from one minor release to the next has caused many, MANY compatibility issues in the past. If we're to bump not one but two Kubernetes versions in one PR, it could be difficult to test and determine whether there are compatibility issues stemming from this PR or from the underlying Kubernetes version bump.

Is there a way we could refrain from bumping Kubernetes in this PR and either use the existing APIs to detect deprecated APIs, retrofit `k8s.io/apiserver/pkg/endpoints/deprecation` for 1.18, or hold off until we've reviewed and updated to a newer version of Kubernetes? I understand the latter is less desirable since this is attempting to fix a bug. But I don't want to introduce a slew of compatibility issues just to fix one bug.

joejulian:
This doesn't bump the version. Those are coming from master. This only bumps the representation of the version in constants and their related test files.

bacongobbler:
Ah yes. So our hardcoded kubernetes version numbers are out of date, and this PR is bringing those back to up match what's in go.mod. Thank you for the clarification.

bacongobbler:
Looks like the golden test files need to be updated to use v1.20.1 instead of v1.20.0. After taht the tests should pass :)

bacongobbler (approved):

bacongobbler:
I'll let @mattfarina or @technosophos merge this one in as they're coordinating efforts towards cutting the 3.5 RC.

joejulian:
> Looks like the golden test files need to be updated to match v1.20.1. After that the tests should pass :)

Yeah, @bacongobbler, it's a bit of a whack-a-mole here. I have a validation PR I'll open soon that will enumerate the files that need updated to make it easier for future k8s library updates.

bacongobbler:
In the future, I've used `cd cmd/helm && go test ./... --update` in the past to update golden files. Saves a ton of time having to manually update each file.

bacongobbler:
https://github.com/helm/helm/blob/6297c021cbda1483d8c08a8ec6f4a99e38be7302/internal/test/test.go#L28-L29

technosophos (approved):
Given the most recent round of changes, I'm good with this.

PR #8595, opened 2020-08-14
Bump Kubernetes to v0.18.8

meyskens:
**What this PR does / why we need it**:
This updates the Kubernetes dependencies to the latest patch release v0.18.8, this includes fixes for a bug in the JSON patching of deeply nested manifests (especially CRDs).

fixes #8594 

**Special notes for your reviewer**:

**If applicable**:
- [ ] this PR contains documentation
- [ ] this PR contains unit tests
- [x] this PR has been tested for backwards compatibility

PR #8612, opened 2020-08-19
bump json-patch lib to v5.1.0

tariq1890:
Updating json-patch to go mod compliant version


go.mod
```diff
@@ -16,7 +16,7 @@ require (
 	github.com/docker/distribution v2.7.1+incompatible
 	github.com/docker/docker v1.4.2-0.20200203170920-46ec8731fbce
 	github.com/docker/go-units v0.4.0
-	github.com/evanphx/json-patch v0.0.0-20200808040245-162e5629780b
+	github.com/evanphx/json-patch/v5 v5.0.0
```
mattfarina, line 19: The older hash based version references v5.1.0. See https://github.com/evanphx/json-patch/commit/162e5629780b. Can you please set the minimum version to v5.1.0 so that the same code is used.

mattfarina (changes requested):

tariq1890:
Thanks @mattfarina . I've amended the commit

bacongobbler:
json-patch was bumped to its current version due to a Kubernetes version compatibility issue in https://github.com/helm/helm/pull/8595. Is there a compatibility concern here by bumping json-patch higher than what Kubernetes is using?

Here is what Kubernetes is using in `master`: https://github.com/kubernetes/kubernetes/blob/dd3d66b27d60459a719197202819bcfcf16de03e/go.mod#L44

tariq1890:
I believe the json-patch bugfixes were merged upstream.

tariq1890:
Looks like the fixes haven't yet had a release

mattfarina:
@bacongobbler Good catch on the Kubernetes usage. The version Kubernetes uses (i.e., 162e5629780b) is the same version as v5.1.0. Same commit.

I would suggest that this change be made in Kubernetes. Maybe even before making it here in Helm.

PR #8613, opened 2020-08-19
optimise if condition in service ready logic

tariq1890:
If we are checking for an empty string, (pre)ANDing it with another value check is redundant

### DIFF
--- a/pkg/kube/wait.go
+++ b/pkg/kube/wait.go
@@ -188,8 +188,8 @@ func (w *waiter) serviceReady(s *corev1.Service) bool {
 		return true
 	}
 
-	// Make sure the service is not explicitly set to "None" before checking the IP
-	if s.Spec.ClusterIP != corev1.ClusterIPNone && s.Spec.ClusterIP == "" {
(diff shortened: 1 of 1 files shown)


pkg/kube/wait.go
```diff
@@ -188,8 +188,8 @@ func (w *waiter) serviceReady(s *corev1.Service) bool {
 		return true
 	}
 
-	// Make sure the service is not explicitly set to "None" before checking the IP
-	if s.Spec.ClusterIP != corev1.ClusterIPNone && s.Spec.ClusterIP == "" {
+	// Ensure that the service cluster IP is not empty
+	if s.Spec.ClusterIP == "" {
```
tariq1890, line 192: @thomastaylor312 Can you review this?

bacongobbler (approved):
After looking thoroughly at the logic behind this line of code (it's seen a number of changes since Kubernetes 1.6), this looks good to me.

tariq1890:
Thanks for the review!

PR #8626, opened 2020-08-20
Replace Helm Hub with Artifact Hub

scottrigby:

Helm Hub is going away in favor of Artifact Hub (pending lazy consensus, EOD Today).

There is a plan in place to safely migrate so end users will not experience any difference when using `helm search hub`, even without this change to the helm client. See https://github.com/helm/hub/issues/439

This PR updates:
- The `helm search hub` `--endpoint` flag default to https://artifacthub.io/
- This ensures user output will match the browsable page, so search URLs correctly display the final destination
- this also allows the `[KEYWORD]` arg to optionally accept quoted rich queries. But single keyword continues to work
- tests
- cli help
- other docs and references

For backwards compatibility, previous versions of helm client will still work:
- helm search hub output URLs for hub.helm.sh will DNS redirect to artifacthub.io 👉 
- cli help and documentation won't be correct, but there's nothing we can do about that for previous versions of the client 🤷 

**If applicable**:

### DIFF
--- a/cmd/helm/search_hub.go
+++ b/cmd/helm/search_hub.go
@@ -68,7 +68,7 @@ func newSearchHubCmd(out io.Writer) *cobra.Command {
 	}
 
 	f := cmd.Flags()
-	f.StringVar(&o.searchEndpoint, "endpoint", "https://artifacthub.io", "Hub instance to query for charts")
+	f.StringVar(&o.searchEndpoint, "endpoint", "https://hub.helm.sh", "Hub instance to query for charts")
 	f.UintVar(&o.maxColWidth, "max-col-width", 50, "maximum column width for output table")
 	bindOutputFlag(cmd, &o.outputFormat)
 
--- a/cmd/helm/search_hub.go
+++ b/cmd/helm/search_hub.go
@@ -39,7 +39,7 @@ Foundation sandbox project. You can browse the hub at https://artifacthub.io/
 
 The [KEYWORD] argument accepts either a keyword string, or quoted string of rich
 query options. For rich query options documentation, see
-https://artifacthub.github.io/hub/api/#/Packages/get_packages_search
+https://artifacthub.github.io/hub/api/?urls.primaryName=Monocular%20compatible%20search%20API#/Monocular/get_api_chartsvc_v1_charts_search
 
 Previous versions of Helm used an instance of Monocular as the default
 'endpoint', so for backwards compatibility Artifact Hub is compatible with the
--- a/cmd/helm/search_hub.go
+++ b/cmd/helm/search_hub.go
@@ -106,7 +106,14 @@ type hubSearchWriter struct {
 func newHubSearchWriter(results []monocular.SearchResult, endpoint string, columnWidth uint) *hubSearchWriter {
 	var elements []hubChartElement
 	for _, r := range results {
+		// Backwards compatibility for Monocular
 		url := endpoint + "/charts/" + r.ID
+
+		// Check for artifactHub compatibility
+		if r.ArtifactHub.PackageURL != "" {
+			url = r.ArtifactHub.PackageURL
+		}
+
 		elements = append(elements, hubChartElement{url, r.Relationships.LatestChartVersion.Data.Version, r.Relationships.LatestChartVersion.Data.AppVersion, r.Attributes.Description})
 	}
 	return &hubSearchWriter{elements, columnWidth}
--- a/cmd/helm/search_hub.go
+++ b/cmd/helm/search_hub.go
@@ -30,15 +30,23 @@ import (
 )
 
 const searchHubDesc = `
-Search the Helm Hub or an instance of Monocular for Helm charts.
+Search for Helm charts in the Artifact Hub or your own hub instance.
 
-The Helm Hub provides a centralized search for publicly available distributed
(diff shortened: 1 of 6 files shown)


paulczar:
thinking about this more ... we should keep the default search url to hub.helm.sh ...  that way we're tied to our own URL ...  eventually this should be just a cname to artifacthub ... but allows us to retain control of the default search location.

scottrigby:
@paulczar The only downside I see with that is users of the client would output a different URL than the user would wind up on. That may make users wonder if they've been hijacked?

paulczar:
I'm not sure its a problem ... right now I do 

```
helm search hub discourse
URL                                         	CHART VERSION	APP VERSION	DESCRIPTION                                       
https://hub.helm.sh/charts/bitnami/discourse	0.3.3        	2.5.0      	A Helm chart for deploying Discourse to Kubernetes
https://hub.helm.sh/charts/halkeye/discourse	0.2.1        	v2.2.1     	A platform for community discussion. Free, open...
```

if we continue to use hub.helm.sh as the search url, but its actually being served by artifacthub I'd get:

```
helm search hub discourse
URL                                         	CHART VERSION	APP VERSION	DESCRIPTION                                       
https://artifacthub.io/charts/bitnami/discourse	0.3.3        	2.5.0      	A Helm chart for deploying Discourse to Kubernetes
https://artifacthub.io/charts/halkeye/discourse	0.2.1        	v2.2.1     	A platform for community discussion. Free, open...
```

either way ... it doesn't appear as if I've been hijacked ...   I guess if I did `helm search hub --endpoint hub.helm.sh` then the output wouldn't match ... but I think that's a edge case.

tegioz:
> if we continue to use hub.helm.sh as the search url, but its actually being served by artifacthub I'd get:
> 
> ```
> helm search hub discourse
> URL                                         	CHART VERSION	APP VERSION	DESCRIPTION                                       
> https://artifacthub.io/charts/bitnami/discourse	0.3.3        	2.5.0      	A Helm chart for deploying Discourse to Kubernetes
> https://artifacthub.io/charts/halkeye/discourse	0.2.1        	v2.2.1     	A platform for community discussion. Free, open...
> ```

I'm not sure if that's what you'd get @paulczar. As it's in the code right now, the endpoint provided or the default one will be used in all urls, as it's hardcoded.

https://github.com/helm/helm/blob/d6708667da5c79ab4a093ea7473e96b93be5f1f7/cmd/helm/search_hub.go#L101

If you prefer leaving `hub.helm.sh` as the default search url, which makes sense to keep control as you mentioned, I think the piece of the code referenced above should be changed to populate the charts url dynamically from the server response (maybe doing something similar to the example we discussed earlier today in Slack). I could provide it in the new Monocular compatible search API endpoint, as a new field or reusing an existing one.

scottrigby:
hey gang, I had to miss the Helm Dev meeting Thursday (had a busy week!), but just caught up on the meeting discussion notes under my [Open question about Helm Hub-> Artifact Hub move: should we keep hub.helm.sh as the default endpoint even after pointing that to artifacthub.io?](https://docs.google.com/document/d/1d-6xJEx0C78csIYSPKJzRPeWaHG_8W1Hjl72OJggwdc/edit#heading=h.853cd0jffl7w) SGTM 👍  just reverted the default endpoint back to hub.helm.sh.

bridgetkromhout:
@mattfarina oops, looks like this didn't make it into https://github.com/helm/helm/releases/tag/v3.4.0-rc.1 - would you like to add it?

mattfarina (approved):
lgtm

cmd/helm/search_hub.go
```diff
@@ -51,16 +59,16 @@ func newSearchHubCmd(out io.Writer) *cobra.Command {
...
+		Short: "search for charts in the Artifact Hub or your own hub instance",
 		Long:  searchHubDesc,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			return o.run(out, args)
 		},
 	}
 
 	f := cmd.Flags()
-	f.StringVar(&o.searchEndpoint, "endpoint", "https://hub.helm.sh", "monocular instance to query for charts")
+	f.StringVar(&o.searchEndpoint, "endpoint", "https://hub.helm.sh", "Hub instance to query for charts")
```
jdolitsky, line 71: Is this still the correct endpoint?

jdolitsky (changes requested):

cmd/helm/search_hub.go
scottrigby: Yes, but only because in the helm dev meeting where I brought this up the group decided to keep the helm.hub.sh domain. The stated reason was if we ever wanted to change the default hub again, we could retroactively do this for previous client versions as well since we control the domain. fe41873 was made as a concession to this decision. It works because for the foreseeable future it's a passthrough to artifacthub.io, and artifact hub has a redirect in place for the monocular paths to the artifact hub chart paths.

mattfarina: @jdolitsky To add to what @scottrigby had to say... Artifact Hub is a sandbox project. If something happens to it and it does not continue on long term we need a method to not break users. By controlling the DNS we could take it back and users would not be affected. This is a contingency just in case. Currently, Artifact Hub manages hub.helm.sh so that it can point to their things.

technosophos (approved):
LGTM.

PR #8637, opened 2020-08-21
Fixing failing CI for windows

mattfarina:
A fix introduced in #8631 caused a bug in Windows builds due to
a type difference between POSIX and Windows environments. This
change corrects that problem and provides a code comment to warn
others.


bacongobbler (approved):
LGTM.

PR #8646, opened 2020-08-24
Test the version number for semver compatibility

### LINKED ISSUE #8645: "helm version" does not report patch number when compiled outside regular process
Helm version 3.3 breaks semver compatibility because of missing patch version. Tools which rely on a correct version synthax are failing.

(Ex: https://github.com/GoogleContainerTools/skaffold/issues/4711)

The next patch version will fix this problem but it should be considered for the next minor version.

Output of `helm version`:
```
 helm version --client                                                                                                                                                                                                           
version.BuildInfo{Version:"v3.3", GitCommit:"", GitTreeState:"", GoVersion:"go1.15"}
```

bdudelsack:
Adds a test for semver compatibility

closes: #8645 

**If applicable**:


internal/version/version_test.go
```diff
@@ -0,0 +1,31 @@
...
+package version
+
+import (
+	"testing"
+
+	"github.com/blang/semver"
+)
+
+func TestVersionSemver(t *testing.T) {
+	_, err := semver.Parse(version)
```
mattfarina, line 26: Some context... while a version of Helm is being developed the version is set to the minor version only. You end up with versions output by the `version` command of `3.3+unreleased`.

When Helm is compiled `helm.sh/helm/v3/internal/version.version` is set as part of the processes with a semantic version. The `Makefile` does this for us.

mattfarina (changes requested):
The test here fails due to the internal version not being SemVer. That would need to be fixed, so the version passes, as part of the PR. We can't merge failing tests in.

mattfarina:
@bdudelsack Helm uses a [DCO](https://developercertificate.org/). Can you please add a Git signoff. The _checks_ tab on the PR has directions on fixing the commit.

bdudelsack:
Is it intented to use a version without the patch number in development? It's not clear for me if i'm on the right way doing this PR here or is it a problem of package creators?

I've corrected the Version to v3.3.0, added a test for semver with Masterminds/semver packages and corrected other tests to comply with the new version.

pkg/lint/rules/chartfile.go
```diff
@@ -22,11 +22,12 @@ import (
 	"os"
 	"path/filepath"
 
-	"github.com/Masterminds/semver/v3"
 	"github.com/asaskevich/govalidator"
 	"github.com/pkg/errors"
 	"sigs.k8s.io/yaml"
 
+	"github.com/Masterminds/semver/v3"
```
bacongobbler, line 29: What's with the move here? Can you move this back?

bacongobbler (changes requested):
Can you please move the import statements back to their original location?

The reason we never added the patch version number was because we did not bump the internal version during patch numbers.

Unless we update the internal version every patch release, this is going to cause odd scenarios where the development version is `v3.3.0+unreleased`, but the latest release is `v3.3.2`, which is why we chose to use `v3.3+unreleased` in the first place.

@adamreese and others may have more opinions here on this change.

pkg/lint/rules/chartfile.go
bdudelsack: It was moved by the `gitimports` command. I reverted the changes.

PR #8651, opened 2020-08-25
fix for linting on templates with only comments

### LINKED ISSUE #8621: Linting fails in 3.3.0 when a template file results in only a comment.
Output of `helm version`:
version.BuildInfo{Version:"v3.3.0", GitCommit:"8a4aeec08d67a7b84472007529e8097ec3742105", GitTreeState:"dirty", GoVersion:"go1.14.7"}

An error occurs when linting a chart that results in a template file containing only a comment:
```` 
# comment on first line
{{- if false }}
apiVersion: v1
kind: ConfigMap
metadata:
  name: mychart-configmap
data:
  myvalue: "Hello World"
{{- end }}
````

the resulting lint throws an error:
````
[ERROR] templates/configmap-false.yaml: object name does not conform to Kubernetes naming requirements: ""
````

Essentially an empty file with only a comment triggers this issue.

```` 
# comment on first line
{{- if true }}
apiVersion: v1
kind: ConfigMap
metadata:
  name: mychart-configmap
data:
  myvalue: "Hello World"
{{- end }}
````

results in a successful lint:
````
````

### PR DESCRIPTION
**Pull request by fidato13**
Signed-off-by: Tarun Kumar <fidato.july13@gmail.com>

Closes #8621 

This PR handles the error scenario wherein the linting fails when the template files contain only comments.

**If applicable**:

### DIFF
--- a/pkg/lint/rules/template.go
+++ b/pkg/lint/rules/template.go
@@ -123,6 +123,7 @@ func Templates(linter *support.Linter, values map[string]interface{}, namespace 
 		// linter.RunLinterRule(support.WarningSev, fpath, validateQuotes(string(preExecutedTemplate)))
 
 		renderedContent := renderedContentMap[path.Join(chart.Name(), fileName)]
+		renderedContent = cleanseRenderedContent(renderedContent)
 		if strings.TrimSpace(renderedContent) != "" {
 			linter.RunLinterRule(support.WarningSev, fpath, validateTopIndentLevel(renderedContent))
 			var yamlStruct K8sYamlStruct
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Review by bacongobbler — CHANGES_REQUESTED** [review]
this potentially conflicts with https://github.com/helm/helm/pull/8608 which reduces the severity to a warning. Might be worth checking that PR out and seeing if that'll fix the issue for you.

I would be concerned about arbitrarily stripping lines from the output based on there being a certain character in there. For example, the following would fail:

```
longstring: -
This is a multi-line message, which includes
# commentary and
--- more commentary, which is valid YAML
and is part of the original string
```

Also, I believe content prefaced with whitespace will still be rendered here, so it doesn't look like this will cover all cases.

PR #8650, opened 2020-08-25
feat: add charts versions in output of dependencies downloader

aleksey-hariton:
PR making output of `helm dependency update` prettier and more clear, especially in CI output.

**Why**

When you running `helm` as part of CI pipeline, its not clear which versions of charts are used during deployment.
This PR improving it, example of output:

```
> helm dependency update

...

Saving 3 charts
Downloading nginx 1.2.9 from repo https://...
Downloading posgresql 3.2.5 from repo https://...
Downloading gitlab-runners 1.0.1 from repo https://...
```

PR #8662, opened 2020-08-27
Add requested version to install error output

Mikulas:

When installing a helm chart with incorrect version, the resulting error message is ambiguous:

```console
$ helm repo add bitnami https://charts.bitnami.com/bitnami
$ helm repo update
$ helm3 install test bitnami/nginx --version 1.19.1
Error: failed to download "bitnami/nginx" (hint: running `helm repo update` may help)
```

```console
$ helm3 search repo bitnami/nginx
NAME                            	CHART VERSION	APP VERSION	DESCRIPTION
bitnami/nginx                   	6.1.0        	1.19.1     	Chart for the nginx server
```

This is patch improves the error message to read:
```console
$ helm3 install test bitnami/nginx --version 1.19.1
Error: failed to download "bitnami/nginx" at version "1.19.1" (hint: running `helm repo update` may help)
```

**If applicable**:

In this simplified example it seem obvious. In a CI context when the user may specify the version indirectly, it is a tremendous help when the version is in the actual error message. It may help with obvious typos as well as it hints that the error may not be in the repository url but rather the version itself.

### DIFF
--- a/cmd/helm/show_test.go
+++ b/cmd/helm/show_test.go
@@ -50,6 +50,13 @@ func TestShowPreReleaseChart(t *testing.T) {
 			expectedErr: "failed to download \"test/pre-release-chart\"",
 		},
 		{
+			name:        "show pre-release chart",
+			args:        "test/pre-release-chart",
+			fail:        true,
+			flags:       "--version 1.0.0",
+			expectedErr: "failed to download \"test/pre-release-chart\" at version \"1.0.0\"",
+		},
+		{
 			name:  "show pre-release chart with 'devel' flag",
 			args:  "test/pre-release-chart",
 			flags: "--devel",
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -677,5 +677,9 @@ func (c *ChartPathOptions) LocateChart(name string, settings *cli.EnvSettings) (
 		return filename, err
 	}
 
-	return filename, errors.Errorf("failed to download %q (hint: running `helm repo update` may help)", name)
+	atVersion := ""
+	if version != "" {
+		atVersion = fmt.Sprintf(" at version %q", version)
+	}
+	return filename, errors.Errorf("failed to download %q%s (hint: running `helm repo update` may help)", name, atVersion)
 }


marckhouzam:
I think this is a nice improvement, thank you.

I'll need input from an other maintainer to know if changing an error output is ok for backwards-compatibility.

How come the PR mentions an improvement with the `install` command but the test modified is for the `show` command?

That being said, I think this improvement would be nice for the `show` command (maybe `update` and `pull` too?)

aleksey-hariton:
Hey @marckhouzam, mb you can take a loot into very similar PR (versions in `dependency update`) - https://github.com/helm/helm/pull/8650

Mikulas:
@marckhouzam Good point on the backwards compatibility. I figured it was a human friendly error message which should not be parsed, but I can see how it could break some CI flows.

With regards to the naming: I updated the format (with unconditional change at first) and the only failing test case was the one in `show_test.go`. Afterwards I proceeded with adding the condition as to not output `at version ""` for unset version, which would be worse for backwards compatibility.

I agree the unit test should be in a different test case / file. The current test (before this MR) is more of a side effect of the overall pre-release package test. Is this in scope of this MR or is it preferable to merge this as is and refactor the tests in a separate MR?

marckhouzam:
> I agree the unit test should be in a different test case / file. The current test (before this MR) is more of a side effect of the overall pre-release package test. Is this in scope of this MR or is it preferable to merge this as is and refactor the tests in a separate MR?

@Mikulas I'm not sure I understand which test you are referring to here.  However I think that as long as we have a test covering the new code it should be enough. No need to refactor in this PR.  So if it's too complicated to add a test on the `install` and that the one on `show` covers the change, then it's ok for me.

bacongobbler:
Changing error strings is fine. Changing concrete error types is not okay.

Since this is just adding more information to an error message, this is fine.

bacongobbler (approved):

marckhouzam:
Sorry for the delay @Mikulas and thank you for the contribution!  This will be better for users.

PR #8664, opened 2020-08-27
feat(comp): Add support for fish completion

marckhouzam:
Part of #8109
Replaces #8111 
  
Provides support for Fish shell completion through Cobra

This PR replaces #8111 because the contributor has not come back to his PR, which needs a rebase, since July 7th, 2020.
I've reached out three times but no luck. 

Also, a new flag `--no-descriptions` was added just for the command `helm completion fish`, to allow users to disable descriptions if they don't like to have them.  Descriptions are a feature of fish completion (and native zsh completion) that looks like this:
```
$ bin/helm s<TAB>
search  (search for a keyword in charts)  
show    (show information of a chart)  
status  (display the status of the named release)
```

### DIFF
--- a/cmd/helm/completion.go
+++ b/cmd/helm/completion.go
@@ -52,6 +52,25 @@ To load completions for every new session, execute once:
 $ helm completion zsh > "${fpath[1]}/_helm"
 `
 
+const fishCompDesc = `
+Generate the autocompletion script for Helm for the fish shell.
+
+To load completions in your current shell session:
+$ helm completion fish | source
+
+To load completions for every new session, execute once:
+$ helm completion fish > ~/.config/fish/completions/helm.fish
+
+You will need to start a new shell for this setup to take effect.
+`
+
+const (
+	noDescFlagName = "no-descriptions"
+	noDescFlagText = "disable completion descriptions"
+)
+
+var disableCompDescriptions bool
+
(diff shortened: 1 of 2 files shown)


mattfarina (approved):
lgtm.

Manually testing this out was fun. There are some things I like about fish that I forget about.

marckhouzam:
Thanks @mattfarina.
I also liked fish after trying out.  The use of colors is nice, the scripting language is more intuitive, but the feature that I enjoy the most was the grayed-out command-line that recalls my previous similar command.  I have installed a plugin in zsh to mimic that last feature, I can't live without it anymore.

thernstig:
@marckhouzam and @mattfarina - I just recently tested this, but assume it has not made it in to any release yet, and that it can be expected in Helm 3.4.0.

But I also noticed that this PR forgot to update the documentation, more specifically found here:
https://helm.sh/docs/helm/helm_completion/

Also note that there should be a documentation entry at this location: https://helm.sh/docs/helm/helm_completion_fish/

marckhouzam:
Thank you @thernstig for trying this out.  You are right, it will be part of Helm 3.4.
From what I know, the docs you mention are automatically generated and should therefore be updated once 3.4 has been released.

If you notice anything unexpected in your usage of Fish completion, I encourage you to continue reporting it, it will help others.

PR #8671, opened 2020-08-28

### LINKED ISSUE #8660: Wait logic doesn't wait for all deployment and daemonset pods to be updated/ready

seaneagan:
The wait logic for Deployments and Daemonsets subtracts their maxUnavailable
value from the amount of replicas they should have ready. This can lead to
waiting for 0 pods to be ready. This change ensures at least 1 replica is ready,
unless there are 0 total desired replicas.

Fixes #8660.

### DIFF
--- a/pkg/kube/wait.go
+++ b/pkg/kube/wait.go
@@ -220,7 +220,17 @@ func (w *waiter) volumeReady(v *corev1.PersistentVolumeClaim) bool {
 }
 
 func (w *waiter) deploymentReady(rs *appsv1.ReplicaSet, dep *appsv1.Deployment) bool {
+	// If there are no desired replicas, there will be nothing to wait for.
+	if *dep.Spec.Replicas == 0 {
+		return true
+	}
+
+	// Ensure the expected available replicas are ready.
 	expectedReady := *dep.Spec.Replicas - deploymentutil.MaxUnavailable(*dep)
+	// Also ensure at least 1 replica is ready.
+	if expectedReady < 1 {
+		expectedReady = 1
+	}
 	if !(rs.Status.ReadyReplicas >= expectedReady) {
 		w.log("Deployment is not ready: %s/%s. %d out of %d expected pods are ready", dep.Namespace, dep.Name, rs.Status.ReadyReplicas, expectedReady)
 		return false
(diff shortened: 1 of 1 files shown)


bridgetkromhout:
Hi, @seaneagan! Thanks for working on this PR. I realize it is a work in progress, but please sign the DCO as explained in https://github.com/helm/helm/pull/8671/checks?check_run_id=1043117851 - thanks!

bacongobbler:
Hi @seaneagan. We ask that contributors please either write a unit test to cover this scneario, or write up a set of steps maintainers can use to verify and test the behaviour is working as intended. This saves both us time having to write test cases for each incoming PR, and it saves you time waiting for us to write said test case.

seaneagan:
@bacongobbler I don't see any automated tests for the wait logic, if I missed them let me know.

Below is a test case where the wait should fail due to none of the pods becoming ready due to `faults.unready=true`, but currently it doesn't as the [maxUnavailable is set to 1](https://github.com/stefanprodan/podinfo/blob/master/charts/podinfo/templates/deployment.yaml#L17) and replicaCount is set to 1, so it waits for 1-1=0 pods to be ready. With this PR the wait fails as it waits for 1 pod to become ready.

```
kubectl create ns test
helm repo add podinfo https://stefanprodan.github.io/podinfo
helm upgrade --install \
--wait \
--timeout 10s \
--namespace test \
--set replicaCount=1 \
--set faults.unready=true \
--debug \
podinfo \
podinfo/podinfo
```

jon-depop:
Bump for this PR. @bacongobbler if you're satisfied with the test steps mentioned, it would be great to see this PR get merged. Thanks both!

jon-depop:
@seaneagan do you have a PR for the observedGeneration race condition?

seaneagan:
@bacongobbler Is there anything else I can do to move this PR along? Thanks!

cbf123:
Just wondering if anyone was planning on ever merging this?  Last update was from a year ago.

mattfarina:
@seaneagan I'm sorry it's taken so long for us to really look at this. Yikes. Is there any way you can rebase? There's a merge conflict.

cbf123:
> @seaneagan I'm sorry it's taken so long for us to really look at this. Yikes. Is there any way you can rebase? There's a merge conflict.

Hi, I'm the guy that joined the weekly call.  I've rebased this change under https://github.com/helm/helm/issues/8660 after checking whether Sean was interested in reviving it or not.

mattfarina:
Closing this PR as https://github.com/helm/helm/pull/10831 is where current work is at.

PR #8673, opened 2020-08-29
fix: remove print statement when writing data to an output directory

### LINKED ISSUE #8672: Add option to supress output for helm template
Output of `helm version`: 3.3.0

Output of `kubectl version`: 1.16.0

Executing `helm template <chart-directory> --output-dir=<random-location>` prints endless lines of this form:
```
wrote ...
wrote ...
wrote ...
.
.
```

Seems like these lines are coming from here: 
https://github.com/helm/helm/blob/master/pkg/action/install.go#L504

Would be nice to have an option to disable printing these lines. They make debugging hard for projects with many charts.

I can make a PR, if it makes sense.

ankitm123:
Signed-off-by: ankitm123 <ankitmohapatra123@gmail.com>

This PR removes the print line when writing the data to an output directory. This line printed lot of statements, which made it hard to debug issues with helm template.

fixes #8672

### DIFF
--- a/cmd/helm/template.go
+++ b/cmd/helm/template.go
@@ -197,7 +197,6 @@ func writeToFile(outputDir string, name string, data string, append bool) error 
 		return err
 	}
 
-	fmt.Printf("wrote %s\n", outfileName)
 	return nil
 }
 
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -500,8 +500,6 @@ func writeToFile(outputDir string, name string, data string, append bool) error 
 	if err != nil {
 		return err
 	}
-
-	fmt.Printf("wrote %s\n", outfileName)
 	return nil
 }
 


marckhouzam:
Thanks @ankitm123.
I've dug a little deeper into this and I was wrong.  The `printf` seems to be intentional, as it comes from helm v2:
https://github.com/helm/helm/blob/dev-v2/cmd/helm/template.go#L282

Also, removing it may be considered a breaking change as existing scripts could be expecting this output.

Finally, I wonder if the line should not be printing to `out` as in v2.  Meaning:
```
	fmt.Fprintf(out, "wrote %s\n", outfileName)
```

I'll have to defer to someone with more experience.

Maybe @bacongobbler or @hickeyma could weigh in?

ankitm123:
I think it could be shown when the debug flag is set to true 🤔 . Printing it all the time, seems like noise to me.

ankitm123:
> Finally, I wonder if the line should not be printing to `out` as in v2. Meaning:
> 
> ```
> 	fmt.Fprintf(out, "wrote %s\n", outfileName)
> ```

Ah, that explains why I did not see those lines when doing helm template with helm v2.

ankitm123:
@bacongobbler @hickeyma thoughts? I can make it the same as helm v2, if that is the intended behavior 🤔 .

hickeyma (commented):
Thanks for pushing the PR @ankitm123.

Making this change would break backwards compatibility in Helm v3. It could be changed in Helm v4 alright.  

I am wondering if the output should be changed though?

At the moment the output is as follows:

```console
$ helm template mychart --output-dir ./out
wrote ./out/mychart/templates/serviceaccount.yaml
wrote ./out/mychart/templates/service.yaml
wrote ./out/mychart/templates/deployment.yaml
wrote ./out/mychart/templates/tests/test-connection.yaml
```

IMO, it is informative to the user on the files written to the directory.  You could change it to say "file written to <output_dir>" and then make the filenames output be debug. I am however unconvinced of the need.

ankitm123:
I think it's fine when it's only one chart, but imagine the scenario where we have many charts, it just becomes noise (informational noise - if that's even a thing :smile: ).
Also, if a `--supress-ouput` flag can be added to helm template, then it wont be a breaking change if not specifying that flag would set that option to false, right?
wdyt?

marckhouzam:
> Also, if a `--supress-ouput` flag can be added to helm template, then it wont be a breaking change if not specifying the flag would set that option to false, right?

How about just doing something like:
```
$ helm template mychart --output-dir ./out | grep -v "wrote "
```

ankitm123:
Yeah, that is what I do now (I edited my comment some time before to reflect that idea). But I honestly prefer it to be a flag, but I am ok with this for now ...

hickeyma:
> lso, if a --supress-ouput flag can be added to helm template, then it wont be a breaking change if not specifying the flag would set that option to false, right?

I don't think we should add another flag just to suppress output.

ankitm123:
I guess this PR can be closed then, I am fine with using grep for this :+1:

marckhouzam:
Thank you anyway @ankitm123

PR #8676, opened 2020-08-31
feat(helm): add support for a `pre-ready` hook

### LINKED ISSUE #202: Use of --wait by default breaks some charts
**Is this a request for help?**:
Yes.

---

**Is this a BUG REPORT or FEATURE REQUEST?** (choose one):
BUG REPORT

**Version of Helm and Kubernetes**:
Helm:
```
Client: &version.Version{SemVer:"v2.16.1", GitCommit:"bbdfe5e7803a12bbdf97e94cd847859890cf4050", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.16.1", GitCommit:"bbdfe5e7803a12bbdf97e94cd847859890cf4050", GitTreeState:"clean"}
```
Kubernetes:
```
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.2", GitCommit:"59603c6e503c87169aea6106f57b9f242f64df89", GitTreeState:"clean", BuildDate:"2020-01-18T23:30:10Z", GoVersion:"go1.13.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.0", GitCommit:"70132b0f130acc0bed193d9ba59dd186f0e634cf", GitTreeState:"clean", BuildDate:"2019-12-07T21:12:17Z", GoVersion:"go1.13.4", Compiler:"gc", Platform:"linux/amd64"}
```
chart-testing:  v2.4.0, using the [quay.io Docker image](https://quay.io/repository/helmpack/chart-testing?tag=latest&tab=tags)

Basic testing with Helm3 and review of the chart-testing 3.0.0 beta code suggests the same issue should occur with them.

**What happened**:
A chart that installs successfully with `helm install -f example.yaml /path/to/chart` fails to pass `ct install`, as a critical post-install hook is not created, and some pods never come online as a result.

**What you expected to happen**:
Given the same options `ct install` results in the same installation success/failure as `helm install`, and creates the post-install hook at the same time.

**How to reproduce it** (as minimally and precisely as possible):
Generally speaking, install a chart with a post-install hook job, where one or more pods depend on that job completing successfully. Use `helm install` normally, and do not pass `--wait`. Attempt the same with `ct install`, without adding `--helm-extra-args "--wait"`.

For a more concrete example, see https://github.com/Kong/charts/pull/36 and its associated branch/CI result. That branch installs normally with Helm itself, but fails due to lack of readiness when given to chart-testing.

**Anything else we need to know**:
`--wait` is hard-coded into the Helm invocations used in [2.4](https://github.com/helm/chart-testing/blob/v2.4.0/pkg/tool/helm.go#L62-L63) and [3.0](https://github.com/helm/chart-testing/blob/v3.0.0-beta.1/pkg/tool/helm.go#L58-L59), and has been present since the initial Go conversion. It was present in the [initial shell script import/split from "charts"](https://github.com/helm/chart-testing/blob/96d2ce7c0ebb573fc119175789be5cb4acfe72b9/lib/chartlib.sh#L286) but not in what appears to be the [equivalent script in "charts"](https://github.com/helm/charts/blob/efa000e429d2169c1c7142c68b8da678898100ff/test/changed.sh#L104) before that split.

After initial review, it's not clear to me whether it's possible to handle this case by simply removing the flag without making changes to Helm also. At least for Helm3, `--wait` appears to pull double duty, handling both "fail the release if pods do not become ready" and "require pods become ready before creating post-install resources": https://github.com/helm/helm/blob/v3.0.3/pkg/action/install.go#L282-L293

To remove that and handle this without changes to Helm, I'd expect that chart-testing would need some other means of checking pod readiness afterwards.

### LINKED ISSUE #7740: Create a new "pre-ready" hook
This is similar to #3282, but from the opposite direction: I'd like a hook that always uses the behavior of the `post-install` hook when `--wait` is not used, i.e. hook resources are always created immediately after installation, without waiting to determine if installation succeed.

My main use case for this is handling a database initialization task: [our chart](https://github.com/Kong/charts/tree/master/charts/kong) spawns both the main application and database (via a subchart) during install along with a `post-install` job that initializes the database schema. The main application fails to start until schema initialization completes. This works fine without `--wait`, since both the application and job are effectively created at the same time, but with `--wait`, the job isn't created until the application starts, creating a circular dependency.

I don't believe there's any way to use `pre-install` here, as I'm not aware of a way to spawn a subchart in `pre-install`. Doing so would create several other complications, as it'd probably necessitate creating and then deleting the database pod during `pre-install` to then re-create it during install, as it wouldn't be managed as part of the release otherwise--this may work, but seems a bit cumbersome.

Unlike `post-ready`, this shouldn't introduce additional complications related to returning the prompt; it should be possible with a new hook handler at https://github.com/helm/helm/blob/v3.1.1/pkg/action/install.go#L297

lackstein:

Adds a new `pre-ready` hook type that runs after resources have been
created or updated, but before they're necessarily in a ready state.

This allows for charts that provision a resources using a subchart,
then use a hook to configure those resources (for example, migrating a
database), and require that subchart resources be configured before
the pods in their chart can be in a ready state.

Using the `--wait` flag with Helm will currently fail with these charts,
making them incompatible with the chart-testing tool.

Fixes #7740, fixes helm/chart-testing#202

Signed-off-by: Noah Lackstein <noah@lackstein.com>

**If applicable**:

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -344,6 +344,12 @@ func (i *Install) Run(chrt *chart.Chart, vals map[string]interface{}) (*release.
 		}
 	}
 
+	if !i.DisableHooks {
+		if err := i.cfg.execHook(rel, release.HookPreReady, i.Timeout); err != nil {
+			return i.failRelease(rel, fmt.Errorf("failed pre-ready: %s", err))
+		}
+	}
+
 	if i.Wait {
 		if err := i.cfg.KubeClient.Wait(resources, i.Timeout); err != nil {
 			return i.failRelease(rel, err)
--- a/pkg/action/upgrade.go
+++ b/pkg/action/upgrade.go
@@ -340,6 +340,12 @@ func (u *Upgrade) performUpgrade(originalRelease, upgradedRelease *release.Relea
 		}
 	}
 
+	if !u.DisableHooks {
+		if err := u.cfg.execHook(upgradedRelease, release.HookPreReady, u.Timeout); err != nil {
+			return u.failRelease(upgradedRelease, results.Created, fmt.Errorf("failed pre-ready: %s", err))
+		}
+	}
+
 	if u.Wait {
 		if err := u.cfg.KubeClient.Wait(target, u.Timeout); err != nil {
 			u.cfg.recordRelease(originalRelease)
--- a/pkg/release/hook.go
+++ b/pkg/release/hook.go
@@ -33,6 +33,7 @@ const (
 	HookPostUpgrade  HookEvent = "post-upgrade"
 	HookPreRollback  HookEvent = "pre-rollback"
 	HookPostRollback HookEvent = "post-rollback"
+	HookPreReady     HookEvent = "pre-ready"
 	HookTest         HookEvent = "test"
 )
 
--- a/pkg/releaseutil/manifest_sorter.go
+++ b/pkg/releaseutil/manifest_sorter.go
@@ -61,6 +61,7 @@ var events = map[string]release.HookEvent{
 	release.HookPostUpgrade.String():  release.HookPostUpgrade,
 	release.HookPreRollback.String():  release.HookPreRollback,
 	release.HookPostRollback.String(): release.HookPostRollback,
+	release.HookPreReady.String():     release.HookPreReady,
 	release.HookTest.String():         release.HookTest,
 	// Support test-success for backward compatibility with Helm 2 tests
 	"test-success": release.HookTest,
--- a/pkg/releaseutil/manifest_sorter_test.go
+++ b/pkg/releaseutil/manifest_sorter_test.go
@@ -62,7 +62,8 @@ metadata:
   annotations:
     "helm.sh/hook": post-install
 `,
-		}, {
+		},
+		{
 			name:  []string{"third"},
 			path:  "three",
 			kind:  []string{"ReplicaSet"},
@@ -74,7 +75,8 @@ metadata:
   annotations:
     "helm.sh/hook": no-such-hook
 `,
-		}, {
+		},
+		{
 			name:  []string{"fourth"},
 			path:  "four",
 			kind:  []string{"Pod"},
@@ -85,7 +87,8 @@ metadata:
   name: fourth
   annotations:
     nothing: here`,
-		}, {
+		},
+		{
 			name:  []string{"fifth"},
 			path:  "five",
 			kind:  []string{"ReplicaSet"},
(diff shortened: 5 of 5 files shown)


sc-chad:
Would this solve the use case I'm trying to accomplish here? https://github.com/helm/helm/issues/8694  ( obviously it won't fix the actual issue, but looks like a good work around? )

bacongobbler (changes requested):
Hi @lackstein. We're asking new proposals to first write up a Helm Improvement Proposal, also known as a HIP. This gives the maintainers an opportunity to consider the design of the proposal and iron out any design kinks before a reference implementation is provided. This saves both you and us time in the long run.

Please have a read through https://github.com/helm/community/blob/master/hips/hip-0001.md and let me know if you have any questions.

bacongobbler:
For example, I would consider reading up on hook weights, and why that does not solve the issue for your use case.

https://helm.sh/docs/topics/charts_hooks/#writing-a-hook

lackstein:
@bacongobbler We're already leveraging hook weights within our chart, but they don't provide enough flexibility to allow us to deploy using Helm with `--wait`.

Specifically, we have a chart that installs our app and has a dependency on the PostgreSQL chart. Our app requires the database to be migrated before the pods become ready. There's currently no available hook that will allow us to run migrations after PostgreSQL is installed but before our pods become ready, assuming Helm is run with `--wait`. We're currently using `post-install, post-upgrade`, but those will only run after the pods are ready and as such are incompatible with `--wait` and chart-testing.

The root cause of this pain is that dependencies are installed at the same time as top-level chart, and there's no way to run a hook after a dependency is ready but before the rest of the chart. Any fix to this ordering issue would likely have to wait till Helm 4 to maintain backwards compatibility.

bacongobbler:
Removing myself from review. I [provided a set of next steps](https://github.com/helm/helm/pull/8676#pullrequestreview-489061557) in my last review. Please consider following the process described earlier. Thanks.

iamtakingiteasy:
> Hi @lackstein. We're asking new proposals to first write up a Helm Improvement Proposal, also known as a HIP. This gives the maintainers an opportunity to consider the design of the proposal and iron out any design kinks before a reference implementation is provided. This saves both you and us time in the long run.
> 
> Please have a read through https://github.com/helm/community/blob/master/hips/hip-0001.md and let me know if you have any questions.

This bureaucracy also ensures that some limitations would not be addressed despite having ready PR for many months, if not years.

PR #8677, opened 2020-09-01
fix: limit the usage of chartutil.Values to avoid conversion bugs

aure-olli:
This is the 1st PR splitted from #6876

Golang allows implicit conversion between `map[string]interface{}` and `chartutil.Values`, but not when converted to `interface{}` first. The usage of `chartutil.Values` as values of a root `chartutil.Values` leads to various bugs, only avoided by the usage of `.AsMap` and some random checks. This was previously believed to be the consequence of `nil` values.

This PR finally gets rid of those random bugs by limitating the usage of `chartutil.Values` and preventing its conversion to `iterface{}`.

- All functions now return `map[string]interface{}`. Implicit conversion to `chartutil.Values` makes those changes quite transparent, and avoid accidentally turning `chartutil.Values` to `interface{}`.
- All creations of a `chartutil.Values{...}` have been replaced by a `map[string]interface{}{...}`.
- In general, `chartutil.Values` are only used in one of those 2 cases:
  - when `chartutil.Values` methods are used (often as argument of a function, implicitly converted on function call)
  - when `chartutil.Values` represents the root value for templates (as in `chart.Values` expressions)
- `chartutil.Values.AsMap` has been removed to discorage the usage of `chartutil.Values` (implicit conversion works anyway).
- Tries to convert `interface{}` to `chartutil.Values` have also been removed to avoid hiding bad usages of `chartutil.Values`.

Signed-off-by: Aurélien Lambert <aure@olli-ai.com>

We should wait for the next PRs to merge this one, I may include few additional changes in the future (but it should globally stay the same)

**If applicable**:

### DIFF
--- a/pkg/chartutil/coalesce.go
+++ b/pkg/chartutil/coalesce.go
@@ -34,7 +34,7 @@ import (
 //	- Scalar values and arrays are replaced, maps are merged
 //	- A chart has access to all of the variables for it, as well as all of
 //		the values destined for its dependencies.
-func CoalesceValues(chrt *chart.Chart, vals map[string]interface{}) (Values, error) {
+func CoalesceValues(chrt *chart.Chart, vals map[string]interface{}) (map[string]interface{}, error) {
 	v, err := copystructure.Copy(vals)
 	if err != nil {
 		return vals, err
--- a/pkg/chartutil/values.go
+++ b/pkg/chartutil/values.go
@@ -32,6 +32,16 @@ import (
 const GlobalKey = "global"
 
 // Values represents a collection of chart values.
+//
+// It should never be stored as an interface{} as conversion to
+// map[string]interface{} will then fail (in particular, as value in Values
+// or map[string]interface{} structures).
+// To avoid this to happen, all function should return map[string]interface{},
+// since implicit conversion to Values is allowed anyway.
+//
+// Values should only be used in one of these cases:
+// - when Values methods are used (return type must be map[string]interface{})
+// - to store the values for template rendering (as in chart.Values)
 type Values map[string]interface{}
 
 // YAML encodes the Values into a YAML string.
@@ -52,7 +62,7 @@ func (v Values) YAML() (string, error) {
 // foo".
 //
 // An ErrNoTable is returned if the table does not exist.
-func (v Values) Table(name string) (Values, error) {
+func (v Values) Table(name string) (map[string]interface{}, error) {
 	table := v
 	var err error
 
@@ -62,16 +72,6 @@ func (v Values) Table(name string) (Values, error) {
 		}
 	}
 	return table, err
-}
-
-// AsMap is a utility function for converting Values to a map[string]interface{}.
-//
-// It protects against nil map panics.
-func (v Values) AsMap() map[string]interface{} {
-	if v == nil || len(v) == 0 {
-		return map[string]interface{}{}
-	}
-	return v
 }
 
 // Encode writes serialized Values information to the given io.Writer.
@@ -84,7 +84,7 @@ func (v Values) Encode(w io.Writer) error {
 	return err
 }
 
-func tableLookup(v Values, simple string) (Values, error) {
+func tableLookup(v map[string]interface{}, simple string) (map[string]interface{}, error) {
 	v2, ok := v[simple]
 	if !ok {
 		return v, ErrNoTable{simple}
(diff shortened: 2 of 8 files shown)


pkg/chartutil/coalesce.go
```diff
@@ -34,7 +34,7 @@ import (
 //	- Scalar values and arrays are replaced, maps are merged
 //	- A chart has access to all of the variables for it, as well as all of
 //		the values destined for its dependencies.
-func CoalesceValues(chrt *chart.Chart, vals map[string]interface{}) (Values, error) {
+func CoalesceValues(chrt *chart.Chart, vals map[string]interface{}) (map[string]interface{}, error) {
```
bacongobbler, line 37: As this is a public-facing function signature, this cannot be changed until Helm 4.

pkg/chartutil/values.go
```diff
@@ -52,7 +62,7 @@ func (v Values) YAML() (string, error) {
 // foo".
 //
 // An ErrNoTable is returned if the table does not exist.
-func (v Values) Table(name string) (Values, error) {
+func (v Values) Table(name string) (map[string]interface{}, error) {
```
bacongobbler, line 65: As this is a public-facing function, this cannot be changed until Helm 4.

pkg/chartutil/values.go
```diff
@@ -64,16 +74,6 @@ func (v Values) Table(name string) (Values, error) {
 	return table, err
 }
 
-// AsMap is a utility function for converting Values to a map[string]interface{}.
-//
-// It protects against nil map panics.
-func (v Values) AsMap() map[string]interface{} {
```
bacongobbler, line 70: Functions cannot be removed, as third parties may rely on this functionality. Instead, a note that this function is deprecated and will be removed in Helm 4 is preferred.

pkg/chartutil/values.go
```diff
@@ -93,27 +93,20 @@ func tableLookup(v Values, simple string) (Values, error) {
...
-		return vv, nil
-	}
-
-	return Values{}, ErrNoTable{simple}
+	return map[string]interface{}{}, ErrNoTable{simple}
 }
 
 // ReadValues will parse YAML byte data into a Values.
-func ReadValues(data []byte) (vals Values, err error) {
+func ReadValues(data []byte) (vals map[string]interface{}, err error) {
```
bacongobbler, line 100: See above.

pkg/chartutil/values.go
```diff
@@ -93,27 +93,20 @@ func tableLookup(v Values, simple string) (Values, error) {
...
 	if len(vals) == 0 {
-		vals = Values{}
+		vals = map[string]interface{}{}
 	}
 	return vals, err
 }
 
 // ReadValuesFile will parse a YAML file into a map of values.
-func ReadValuesFile(filename string) (Values, error) {
+func ReadValuesFile(filename string) (map[string]interface{}, error) {
```
bacongobbler, line 109: See above.

pkg/chartutil/values.go
```diff
@@ -134,7 +127,7 @@ type ReleaseOptions struct {
 // ToRenderValues composes the struct from the data coming from the Releases, Charts and Values files
 //
 // This takes both ReleaseOptions and Capabilities to merge into the render values.
-func ToRenderValues(chrt *chart.Chart, chrtVals map[string]interface{}, options ReleaseOptions, caps *Capabilities) (Values, error) {
+func ToRenderValues(chrt *chart.Chart, chrtVals map[string]interface{}, options ReleaseOptions, caps *Capabilities) (map[string]interface{}, error) {
```
bacongobbler, line 130: See above.

bacongobbler (changes requested):
This PR introduces a number of breaking changes to the API. For that reason, we cannot accept this PR in its current form. Please address the concerns over backwards compatibility before moving forward.

Please also write a unit test that demonstrates the issue. That way we have something to reference as we test this PR and discuss the rationale behind the design changes made to the package. Thank you.

aure-olli:
I tried use deprecation mechanism instead, it can work with this commit. clearly won't with the next ones, as we change the global mechanism of parsing values.

If we consider that any public function can be used by an external library, and must be preserved identical, there is no way we can change the way values are treated, as most of the used functions are public.

So maybe for a helm 4

bacongobbler:
Okay. Let's hold off until Helm 4 is underway before considering this PR.

In the future, I'd also ask that you provide some unit tests to cover this case. I only saw changes in the unit tests to change a variable's type, but there weren't any tests to confirm whether this PR actually fixed a "conversion bug", so we have no way to tell whether this actually fixes anything.

Thank you for your submission. Closing.

PR #8678, opened 2020-09-01
Makefile: Fix LDFLAGS overriding

### LINKED ISSUE #8645: "helm version" does not report patch number when compiled outside regular process

Foxboron:
When distributions build software it's desriable to have the ability to
define own linker flags, or Go flags. As `-ldflags` defined in `go
build` overrides `-ldflags` defined in the env variable `GOFLAGS`, there
is a distinct need to be able to define new replace the default values
with new ones.

This patch introduces the override directive to the Makefile so we are
able to supply own LDFLAGS from the `make` invocation along with
appending any needed LDFLAGS from the Makefile.

https://ftp.gnu.org/old-gnu/Manuals/make-3.79.1/html_chapter/make_6.html#SEC66

Fixes #8645

```bash
λ helm » make LDFLAGS="-s -w -linkmode external" GOFLAGS="-buildmode=pie -trimpath"
GO111MODULE=on go build -buildmode=pie -trimpath -tags '' -ldflags '-s -w -linkmode external -X helm.sh/helm/v3/internal/version.version=v3.3.0 -X helm.sh/helm/v3/internal/version.metadata= -X helm.sh/helm/v3/internal/version.gitCommit=8a4aeec08d67a7b84472007529e8097ec3742105 -X helm.sh/helm/v3/internal/version.gitTreeState=dirty' -o /home/fox/Git/prosjekter/Go/helm/bin/helm ./cmd/helm
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,6 +49,7 @@ endif
 LDFLAGS += -X helm.sh/helm/v3/internal/version.metadata=${VERSION_METADATA}
 LDFLAGS += -X helm.sh/helm/v3/internal/version.gitCommit=${GIT_COMMIT}
 LDFLAGS += -X helm.sh/helm/v3/internal/version.gitTreeState=${GIT_DIRTY}
+LDFLAGS += $(EXT_LDFLAGS)
 
 .PHONY: all
 all: build


Foxboron:
After a little bit of thinking I realized this might be overkill and having an own variable for external `LDFLAGS` addition might be better.

`LDFLAGS    := -w -s $(EXT_LDFLAGS)` or something. Thoughts?

bacongobbler:
If `-w -s` is always required, then perhaps an EXT_LDFLAGS may make more sense. Or perhaps just change line 21 to an `override` as well. Up to you.

bacongobbler:
ping.

Foxboron:
I'm terribly sorry you had to ping me! I opted for `LDFLAGS += $(EXT_LDFLAGS)`. It's less intrusive and doesn't introduce new semantics on people that might assume `LDFLAGS=` overrides the `Makefile` ones.

I hope this seems fine. I also edited the commit message.

bacongobbler (approved):
LGTM.

PR #8683, opened 2020-09-02
fix: rollback to release revision containing resource with `helm.sh/resource-policy=keep` anntation

### LINKED ISSUE #8228: Error when performing helm rollback to a release containing a resource with the `helm.sh/resource-policy: keep` annotation
Hi,
We are creating configmaps with a hash in the name, generated from the contents of the configmap data. This is done as a workaround to the k8s issue with rolling updates and configmaps.
In addition, we add the `helm.sh/resource-policy: keep` annotation to make sure that the previous configmap is not deleted during the rolling update, so we end up with several "versioned" configmaps such as:
`config-common-223ee6a5d95cc68fd4af39af3b059b595f54222acf4d47ab0b53134ce89c2cfe`

When performing `helm rollback` to a previous revision which contains a configmap name hash different than the current release revision we get the following error:
`Error: no ConfigMap with the name "config-common-223ee6a5d95cc68fd4af39af3b059b595f54222acf4d47ab0b53134ce89c2cfe" found`

Performing another helm rollback resolves this issue.

I assume the addition of https://github.com/helm/helm/pull/7649 allowed adopting the configmap back into the release, however this should happen on the first rollback attempt.

Output of `helm version`:
```
```
Output of `kubectl version`:
```
\Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.6", GitCommit:"96fac5cd13a5dc064f7d9f4f23030a6aeface6cc", GitTreeState:"clean", BuildDate:"2019-08-19T11:13:49Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"14+", GitVersion:"v1.14.9-eks-502bfb", GitCommit:"502bfb383169b124d87848f89e17a04b9fc1f6f0", GitTreeState:"clean", BuildDate:"2020-02-07T01:31:02Z", GoVersion:"go1.12.12", Compiler:"gc", Platform:"linux/amd64"}
```
Cloud Provider/Platform (AKS, GKE, Minikube etc.): AWS

liuming-dev:
When rollback to a previous release revision that contains a resource having the
annotation `helm.sh/resource-policy=keep`, while the current release revision dosen't
have this resource, rollback option will give `Error: no XXX with the name "XXX" found`
error message. A case was reported in #8228.

We should not care about weather the resource with annotaion `helm.sh/resource-policy=keep`
exists in the current release revison or not.

fix #8228

**If applicable**:

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -177,8 +177,11 @@ func (c *Client) Update(original, target ResourceList, force bool) (*Result, err
 			return err
 		}
 
+		kind := info.Mapping.GroupVersionKind.Kind
+
 		helper := resource.NewHelper(info.Client, info.Mapping)
-		if _, err := helper.Get(info.Namespace, info.Name, info.Export); err != nil {
+		currentObject, err := helper.Get(info.Namespace, info.Name, info.Export)
+		if err != nil {
 			if !apierrors.IsNotFound(err) {
 				return errors.Wrap(err, "could not get information about the resource")
 			}
@@ -191,14 +194,25 @@ func (c *Client) Update(original, target ResourceList, force bool) (*Result, err
 				return errors.Wrap(err, "failed to create resource")
 			}
 
-			kind := info.Mapping.GroupVersionKind.Kind
 			c.Log("Created a new %s called %q in %s\n", kind, info.Name, info.Namespace)
 			return nil
 		}
 
+		// Figure out whether this resource in the target release revision has
+		// annotation "helm.sh/resource-policy=keep", if yes, Helm should not
+		// manage this resource anymore.
+		annotations, err := metadataAccessor.Annotations(currentObject)
+		if err != nil {
+			return errors.Errorf("failed to get annotations of kind %s with the name %q", kind, info.Name)
+		}
+		if annotations != nil {
+			if val, exist := annotations[ResourcePolicyAnno]; exist && val == KeepPolicy {
+				return nil
+			}
+		}
+
 		originalInfo := original.Get(info)
 		if originalInfo == nil {
-			kind := info.Mapping.GroupVersionKind.Kind
 			return errors.Errorf("no %s with the name %q found", kind, info.Name)
 		}
 


pkg/kube/client.go
```diff
@@ -191,14 +193,25 @@ func (c *Client) Update(original, target ResourceList, force bool) (*Result, err
 				return errors.Wrap(err, "failed to create resource")
 			}
 
-			kind := info.Mapping.GroupVersionKind.Kind
 			c.Log("Created a new %s called %q in %s\n", kind, info.Name, info.Namespace)
 			return nil
 		}
 
+		// Figure out weather this resource in the target release revision has
```
bacongobbler, line 200: typo: whether

bacongobbler (changes requested):
Hi @liuming-dev. Once again, I must ask that you provide either a test case to test these bug fixes, or provide an example when the code cannot be easily tested. I wasted a few hours trying to come up with a  test case to test your PR, which is why it took so long to review. Please be respectful of other's time.

Regardless, the proposed PR seems like a workaround to the core issue here. Helm cannot recognize the presence of the first resource being present in the cluster. The error report mentions the following message:

```
no ConfigMap with the name "config-common-223ee6a5d95cc68fd4af39af3b059b595f54222acf4d47ab0b53134ce89c2cfe" found
```

Which is untrue. The resource DOES exist in the cluster. It's because the annotation was removed from the object in the upgrade, preventing the kube client from finding the original.

I think there's a deeper issue here that needs to be resolved. This is not as simple to fix at first glance. Notice that further down in the code, once the original info is retrieved, Helm will ignore the resource due to the keep annotation, so this is just duplicating that check a little further up. I feel that the check was put in place after fetching the original target to prevent certain cases, so I’m hesitant to consider this PR as acceptable to merge.

liuming-dev:
Hi @bacongobbler , thanks for reminder, I will improve in the next PR if I could.

As for this PR, it's not easily to make a test case, but I provide some typical examples here, all examples have passed, but some details need to be discussed, see the following:

- case 1: 
  1. revision 1:  manifest contains resource ConfigMap `cfg-0` with annotation `helm.sh/resource-policy: keep`
  2. revision 2: ConfigMap `cfg-0` was removed from manifest
  `cfg-0` was not managed by Helm and was removed from manifest, `cfg-0` existed in kubernetes
  3. revision 3: rollback to revision 1
  **without this PR**: rollback failed with `Error: no ConfigMap with the name "cfg-0" found` returned
  **with this PR**: rollback successfully, `cfg-0` was kept as revision 1
- case 2:
  1. revision 1:  manifest contains resource ConfigMap `cfg-0` with annotation `helm.sh/resource-policy: keep`
  2. revision 2: annotation `helm.sh/resource-policy: keep` of ConfigMap `cfg-0` was removed from manifest
  **without this PR**: `cfg-0` was still managed by Helm, annotation `helm.sh/resource-policy: keep` of ConfigMap `cfg-0` was removed both in release manifest and kubernetes object (***Is this expected?***)
  **with this PR**: `cfg-0` was not managed by Helm, annotation `helm.sh/resource-policy: keep` of ConfigMap `cfg-0` was removed from manifest, `cfg-0` in kubernets was kept as revision 1
  3. revision 3: rollback to revision 1
  rollback successfully, `cfg-0` was kept as revision 1
- case 3:
  1. revision 1:  manifest contains resource ConfigMap `cfg-0` WITHOUT annotation `helm.sh/resource-policy: keep`
  2. revision 2: ConfigMap `cfg-0` add annotation `helm.sh/resource-policy: keep` in manifest
  3. revision 3: rollback to revision 1
  **without this PR**: `cfg-0` was still managed by Helm, annotation `helm.sh/resource-policy: keep` of ConfigMap `cfg-0` was removed from both release manifest and kubernetes object (***Should we respect keep policy first then rollback option?***)
  **with this PR**: `cfg-0` was not managed by Helm, `cfg-0` in release manifest is like revision 1, `cfg-0` in kubernets was kept as revision 2

pkg/kube/client.go
liuming-dev: fixed

bacongobbler:
> 2\. `cfg-0` was still managed by Helm, annotation `helm.sh/resource-policy: keep` of ConfigMap `cfg-0` was removed both in release manifest and kubernetes object (_**Is this expected?**_)

Yes. That is expected behaviour.

bacongobbler:
Thank you for providing both the test cases you tested as well as the behaviour. Based on the test results provided here, not only does this introduce new code regressions, but it also does not solve the underlying issue as I explained earlier. It only circumvents the issue.

Because of this, I'm going ahead and closing this.

PR #8684, opened 2020-09-02
Add support to make install helm

MaXinjian:
install the binary that was compiled by make build

Signed-off-by: Ma Xinjian <maxj.fnst@cn.fujitsu.com>

**If applicable**:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,5 @@
 BINDIR      := $(CURDIR)/bin
+INSTALL_PATH ?= /usr/local/bin
 DIST_DIRS   := find * -type d -exec
 TARGETS     := darwin/amd64 linux/amd64 linux/386 linux/arm linux/arm64 linux/ppc64le linux/s390x windows/amd64
 TARGET_OBJS ?= darwin-amd64.tar.gz darwin-amd64.tar.gz.sha256 darwin-amd64.tar.gz.sha256sum linux-amd64.tar.gz linux-amd64.tar.gz.sha256 linux-amd64.tar.gz.sha256sum linux-386.tar.gz linux-386.tar.gz.sha256 linux-386.tar.gz.sha256sum linux-arm.tar.gz linux-arm.tar.gz.sha256 linux-arm.tar.gz.sha256sum linux-arm64.tar.gz linux-arm64.tar.gz.sha256 linux-arm64.tar.gz.sha256sum linux-ppc64le.tar.gz linux-ppc64le.tar.gz.sha256 linux-ppc64le.tar.gz.sha256sum linux-s390x.tar.gz linux-s390x.tar.gz.sha256 linux-s390x.tar.gz.sha256sum windows-amd64.zip windows-amd64.zip.sha256 windows-amd64.zip.sha256sum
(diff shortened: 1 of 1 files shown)


bacongobbler (changes requested):
Typically, `make install` refers to installing the binary that was compiled by `make`, or `make build`. `get-helm-3` fetches the latest release from github.

Please modify this new target to install what was compiled with `make`. For those who want to use the `get-helm-3` script, they can call `./scripts/get-helm-3` directly.

Thanks.

Makefile
```diff
@@ -1,4 +1,5 @@
 BINDIR      := $(CURDIR)/bin
+HELM_INSTALL_DIR := /usr/local/bin
```
mattfarina, line 2: If we want this variable to be able to be set in the environment it should use `?=` to set the value. This way it could be set or changed in the outside environment. Like the `BINNAME` variable in this file.

mattfarina (changes requested):
This also makes me wonder about a `link` target that links the binary in `bin/helm` to `/usr/local/bin/helm`. Because I'm repeatedly compiling new and varying versions this could be useful. This is me thinking aloud rather than something I want to see in the PR.

The change has to do with Makefile syntax...

mattfarina:
@MaXinjian what is your reason for proposing this change? What workflow are you attempting to achieve or what problem does it help you solve? I am curious to know the "why" behind this proposed change.

MaXinjian:
> @MaXinjian what is your reason for proposing this change? What workflow are you attempting to achieve or what problem does it help you solve? I am curious to know the "why" behind this proposed change.
I want to install as new as possible helm through a simple "make install" cmd.

Makefile
```diff
@@ -62,6 +63,20 @@ build: $(BINDIR)/$(BINNAME)
...
+
+.PHONY: install
+install:
+	@if [ ! -f "$(BINDIR)/$(BINNAME)" ]; then \
+		echo "please run 'make build' first" && \
+		exit 1; \
+	fi
+
+	@echo "Preparing to install $(BINNAME) into ${HELM_INSTALL_DIR}"
+	@ln -sf "$(BINDIR)/$(BINNAME)" "$(HELM_INSTALL_DIR)/$(BINNAME)"
```
bacongobbler, line 77: A symbolic link should not be used here. If you `rm -rf ..` as most users do after they've compiled the program, helm will no longer be found on the system.

Most `install` targets use `install` instead. Have a look at the [kernel documentation on kbuild](https://www.kernel.org/doc/html/v5.4/kbuild/makefiles.html) on how they configure their `install` target. `INSTALL_PATH` is the common variable for referencing where binaries should be installed, for example.

See `man install` for more information.

bacongobbler (changes requested):

Makefile
```diff
@@ -1,4 +1,5 @@
 BINDIR      := $(CURDIR)/bin
+HELM_INSTALL_DIR ?= /usr/local/bin
```
bacongobbler, line 2: Looking at the [kernel documentation on kbuild](https://www.kernel.org/doc/html/v5.4/kbuild/makefiles.html), `INSTALL_PATH` is the common variable for referencing where binaries should be installed. Can you please change this from HELM_INSTALL_DIR to INSTALL_PATH?

bacongobbler (changes requested):

bacongobbler (approved):
LGTM.

```
~/code/go/src/helm.sh/helm ><> make clean
~/code/go/src/helm.sh/helm ><> make
GO111MODULE=on go build  -tags '' -ldflags '-w -s -X helm.sh/helm/v3/internal/version.metadata=unreleased -X helm.sh/helm/v3/internal/version.gitCommit=8b2cf17648889270ac5c5985f5bb9ef5e43d12d6 -X helm.sh/helm/v3/internal/version.gitTreeState=clean' -o '/home/bacongobbler/code/go/src/helm.sh/helm/bin'/helm ./cmd/helm
~/code/go/src/helm.sh/helm ><> make install
~/code/go/src/helm.sh/helm ><> helm version
version.BuildInfo{Version:"v3.3+unreleased", GitCommit:"8b2cf17648889270ac5c5985f5bb9ef5e43d12d6", GitTreeState:"clean", GoVersion:"go1.15.2"}
```

PR #8696, opened 2020-09-03
fix: boilerplate template indentation

f0rmiga:

Just an indentation fix. We need this so picky people like me can stop freaking out every time we create a new chart using the template. :joy: 

**If applicable**:

### DIFF
--- a/pkg/chartutil/create.go
+++ b/pkg/chartutil/create.go
@@ -254,10 +254,10 @@ spec:
       {{- include "<CHARTNAME>.selectorLabels" . | nindent 6 }}
   template:
     metadata:
-    {{- with .Values.podAnnotations }}
+      {{- with .Values.podAnnotations }}
       annotations:
         {{- toYaml . | nindent 8 }}
-    {{- end }}
+      {{- end }}
       labels:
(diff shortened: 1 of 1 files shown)


bacongobbler:
Hi @f0rmiga. Have you tested this PR to make sure this won't break compatibility? We've introduced small indentation changes like this in the past causing templates to not render correctly.

bacongobbler:
ping.

f0rmiga:
@bacongobbler I cannot imagine how this could ever break. I'd love to see the `small indentation changes` that broke - perhaps I can explain that.

In my case, I'm indenting 2 spaces for the opening and closing sections. They all use `{{-`, which basically does a strip of all the white spaces before it. A grotesque example:
https://play.golang.org/p/2fuvxJ-MoUJ
```go
package main

import (
	"log"
	"os"
	"text/template"
)

const tmplStr = `
foo:
                         



                       {{- if true }}
  bar: baz
  {{- end }}
`

func main() {
	if err := template.Must(template.New("").Parse(tmplStr)).Execute(os.Stdout, nil); err != nil {
		log.Fatal(err)
	}
}

```

f0rmiga:
Ping @bacongobbler.

bacongobbler (approved):

Issue #8654, opened 2020-08-26
Show line numbers when rendering YAML


max-rocket-internet:
Many of the templating errors that Helm produces mention line numbers but these line numbers are from the rendered YAML. For example:

```
Error: YAML parse error on test-chart/templates/deployment.yaml: error converting YAML to JSON: yaml: line 44: did not find expected '-' indicator
```

But then when we run helm with `--debug` to get the rendered YAML, there are no line numbers shown in the rendering. Some of our charts render out to over 3000 lines of YAML it would be helpful to have the line numbers shown and perhaps some colours 😃

bacongobbler:
PRs from the community to improve this functionality would be most welcome.

roy-work:
> But then when we run helm with `--debug` to get the rendered YAML

Running `helm template` with `--debug`, for me, doesn't output the YAML? If it did, it'd probably get me most of the way there, but every time I hit this issue, I'm essentially fumbling in the dark for what is actually wrong w/ the input, since the line numbers are completely divorced from anything in my editor…

ebuildy:
You can use grep for that!

```
helm template -s templates/deployment.yaml . --debug | grep -n '^'
```

Will print:

```
install.go:192: [debug] Original chart version: ""
install.go:209: [debug] CHART PATH: /Users/tdecaux/dev/charts/xxxxx

Error: YAML parse error on xxxxx/templates/deployment.yaml: error converting YAML to JSON: yaml: line 34: mapping values are not allowed in this context
1:---
2:# Source: templates/deployment.yaml
3:apiVersion: apps/v1
4:kind: Deployment
5:metadata:
6:  name: release-name
7:  labels:
8:    helm.sh/chart: xxxxx0.1.0
9:    app.kubernetes.io/name: xxxx
10:    app.kubernetes.io/instance: release-name
11:    app.kubernetes.io/version: "0.9.3"
12:    app.kubernetes.io/managed-by: Helm
...
```

max-rocket-internet:
> You can use grep for that!

Indeed, your suggestion is very helpful 💖 But this would still be a total pain with complex charts with many templates and involves running a second command. This feature should be in Helm itself IMO.

max-rocket-internet:
Possible fix in https://github.com/helm/helm/pull/8699 but there is some dependencies 🙏

PR #8699, opened 2020-09-06
feat: include line numbers on yaml output when yaml fails validation

artemmikhalitsin:
Addresses #8654 
Currently when the yaml contains an error, the error message will state the line of the rendered yaml, not the template. This feature helps with debugging issues in large resources.

The line numbers are only shown on yaml that fails unmarshalling. 
It's currently implemented as a flag on install or template. This could also be just part of the default behaviour when outputting yaml that contains errors, but was implemented as flag to give users the option (not sure whether users currently rely on being able to copy/paste failed yaml from the output). Please provide feedback in terms of best implementation and any other feedback. Thank you.

**If applicable**:

### DIFF
--- a/cmd/helm/testdata/output/template-with-invalid-yaml-debug.txt
+++ b/cmd/helm/testdata/output/template-with-invalid-yaml-debug.txt
@@ -1,13 +1,13 @@
 ---
 # Source: chart-with-template-with-invalid-yaml/templates/alpine-pod.yaml
-apiVersion: v1
-kind: Pod
-metadata:
-  name: "release-name-my-alpine"
-spec:
-  containers:
-  - name: waiter
-    image: "alpine:3.9"
-    command: ["/bin/sleep","9000"]
-invalid
+ 1 apiVersion: v1
+ 2 kind: Pod
+ 3 metadata:
+ 4   name: "release-name-my-alpine"
+ 5 spec:
+ 6   containers:
+ 7   - name: waiter
+ 8     image: "alpine:3.9"
+ 9     command: ["/bin/sleep","9000"]
+10 invalid
 Error: YAML parse error on chart-with-template-with-invalid-yaml/templates/alpine-pod.yaml: error converting YAML to JSON: yaml: line 11: could not find expected ':'
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -17,12 +17,14 @@ limitations under the License.
 package action
 
 import (
+	"bufio"
 	"bytes"
 	"fmt"
 	"os"
 	"path"
 	"path/filepath"
 	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/pkg/errors"
@@ -173,6 +175,7 @@ func (cfg *Configuration) renderResources(ch *chart.Chart, values chartutil.Valu
 			if strings.TrimSpace(content) == "" {
 				continue
 			}
+			content = addLineNumbers(content)
 			fmt.Fprintf(b, "---\n# Source: %s\n%s\n", name, content)
 		}
 		return hs, b, "", err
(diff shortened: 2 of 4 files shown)


pkg/action/action.go
```diff
@@ -224,6 +228,34 @@ func (c *Configuration) renderResources(ch *chart.Chart, values chartutil.Values
...
+	lines := strings.Split(resource, "\n")
+	resultLines := []string{}
+	lineFormat := getNumberedLineFormat(len(lines))
+	for index, line := range lines {
+		lineNumber := index + 1
+		lineWithNum := fmt.Sprintf(lineFormat, lineNumber, line)
+		resultLines = append(resultLines, lineWithNum)
+	}
+
+	return strings.Join(resultLines, "\n")
```
bacongobbler, line 242: Will this work on Windows?

bacongobbler (changes requested):

cmd/helm/install.go
```diff
@@ -149,6 +149,7 @@ func addInstallFlags(cmd *cobra.Command, f *pflag.FlagSet, client *action.Instal
 	f.BoolVar(&client.Atomic, "atomic", false, "if set, the installation process deletes the installation on failure. The --wait flag will be set automatically if --atomic is used")
 	f.BoolVar(&client.SkipCRDs, "skip-crds", false, "if set, no CRDs will be installed. By default, CRDs are installed if not already present")
 	f.BoolVar(&client.SubNotes, "render-subchart-notes", false, "if set, render subchart notes along with the parent")
+	f.BoolVar(&client.WithLineNumbers, "with-line-numbers", false, "if set, if the templating encouters errors, add line numbers to the rendered resources for debugging (should be used with --debug)")
```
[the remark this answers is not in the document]
artemmikhalitsin: Thought the same thing. Flag removed.

pkg/action/action.go
```diff
@@ -224,6 +228,34 @@ func (c *Configuration) renderResources(ch *chart.Chart, values chartutil.Values
 	return hs, b, notes, nil
 }
 
+// addLineNumbers numbers the lines of a resource
+func addLineNumbers(resource string) string {
+	lines := strings.Split(resource, "\n")
```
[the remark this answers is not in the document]
artemmikhalitsin: Thanks for the heads up on this. Actually this result is even better as previously it was adding an empty newline.

cmd/helm/testdata/output/template-with-invalid-yaml-debug.txt
```diff
@@ -1,13 +1,13 @@
...
+ 2 kind: Pod
+ 3 metadata:
+ 4   name: "RELEASE-NAME-my-alpine"
+ 5 spec:
+ 6   containers:
+ 7   - name: waiter
+ 8     image: "alpine:3.9"
+ 9     command: ["/bin/sleep","9000"]
+10 invalid
 Error: YAML parse error on chart-with-template-with-invalid-yaml/templates/alpine-pod.yaml: error converting YAML to JSON: yaml: line 11: could not find expected ':'
```
bacongobbler, line 13: The error is on line 10, yet the error message states it's on line 11. Is that to be expected here?

artemmikhalitsin: Looks like this is related to the yaml package. I assume, when parsing in Unmarshaler, it adds "---" to the start of yaml resource (If I add "---" to the template, it points to the correct line. 

I'll look further into how to fix this.

artemmikhalitsin: Looks like this is an issue in go-yaml fixed in v3 https://github.com/go-yaml/yaml/issues/542

artemmikhalitsin:
Hi @bacongobbler, given that this is an issue in a dependency, how should I proceed with this PR?

bacongobbler:
I'd suggest looking into fixing the issue upstream. We're not going to merge code that doesn't work. :)

max-rocket-internet:
This would be so good. Thanks @artemmikhalitsin 💖

ricardogaspar2:
yeah this would be very helpful!

max-rocket-internet:
Any update @bacongobbler @artemmikhalitsin?

PR #8701, opened 2020-09-07
helm-8700-package: removing the unused paramter 

### LINKED ISSUE #8700: Helm package code
Hello there

Actually this ticket is not related to a specific issue but it concerns the source code of the helm package command. 
I am currently working on a helm plugin related to the `helm package` command and when I looked at the code of that command I realized that on the method `Run` the formal parameter `vals map[string]interface{}` is declared but never used; 

so my question is, have you already spotted this or it has been done in purpose just to satisfy an interface ? here is the code below

````golang
// Run executes 'helm package' against the given chart and returns the path to the packaged chart.
func (p *Package) Run(path string, vals map[string]interface{}) (string, error) {
	ch, err := loader.LoadDir(path)
	if err != nil {
		return "", err
	}

	// If version is set, modify the version.
	if p.Version != "" {
		if err := setVersion(ch, p.Version); err != nil {
			return "", err
		}
	}

	if p.AppVersion != "" {
		ch.Metadata.AppVersion = p.AppVersion
	}

	if reqs := ch.Metadata.Dependencies; reqs != nil {
		if err := CheckDependencies(ch, reqs); err != nil {
			return "", err
		}
	}

	var dest string
	if p.Destination == "." {
		// Save to the current working directory.
		dest, err = os.Getwd()
		if err != nil {
			return "", err
		}
	} else {
		// Otherwise save to set destination
		dest = p.Destination
	}

	name, err := chartutil.Save(ch, dest)
	if err != nil {
		return "", errors.Wrap(err, "failed to save")
	}

	if p.Sign {
		err = p.Clearsign(name)
	}

	return name, err
}

````
thank you for your help.

thynquest:
this PR removes the formal parameter `val[string]interface{}` of the run method of the package command because it is unused
closes #8700

### DIFF
--- a/cmd/helm/package.go
+++ b/cmd/helm/package.go
@@ -27,7 +27,6 @@ import (
 	"github.com/spf13/cobra"
 
 	"helm.sh/helm/v3/pkg/action"
-	"helm.sh/helm/v3/pkg/cli/values"
 	"helm.sh/helm/v3/pkg/downloader"
 	"helm.sh/helm/v3/pkg/getter"
 )
@@ -50,7 +49,6 @@ unless your environment is otherwise configured.
 
 func newPackageCmd(out io.Writer) *cobra.Command {
 	client := action.NewPackage()
-	valueOpts := &values.Options{}
 
 	cmd := &cobra.Command{
 		Use:   "package [CHART_PATH] [...]",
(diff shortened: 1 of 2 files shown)


hickeyma (commented):
Thanks for pushing the PR @thynquest. Unfortunately I wasn't thinking clearly. This breaks the API and in Helm we support backwards compatibility between minor/path releases. Hence, this cannot be  changed until Helm 4. I will therefore have to close this PR and will need to wait until Helm 4 dev window opens. Sorry for this.

hickeyma:
Closing as breaks backward compatibility.

PR #8703, opened 2020-09-08
Add option to skip rendering helm tests

### LINKED ISSUE #8691: Add an option to helm template not render resources in tests directory
This issue was already brought up in #6483 and proposed to be incorporated as a feature in helm 3.

#6444 implements an option to not render hooks, but that's not the same.

For example https://github.com/helm/charts/tree/master/stable/jenkins/templates/tests contains two manifests:

```
apiVersion: v1
kind: Pod
metadata:
  name: "{{ .Release.Name }}-ui-test-{{ randAlphaNum 5 | lower }}"
  namespace: {{ template "jenkins.namespace" . }}
  annotations:
    "helm.sh/hook": test-success
```

and

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: {{ template "jenkins.fullname" . }}-tests
  namespace: {{ template "jenkins.namespace" . }}:
```

As you can see only the pod has the hook annotation. I've seen the same in other charts some are also rendering `Roles` or `RoleBindings` which are needed for testing e.g. https://github.com/helm/charts/tree/master/stable/grafana/templates/tests

torstenwalter:

This helps to avoid rendering manifests which are only used for test execution.

    
When helm template is called with `--skip-tests` no manifests with helm test hook annotations are rendered.

Closes #8691

NOTE: I updated description to reflect revised implementation. It now only ignores resources with helm test annotation and not everything within tests directory.

- `skipTests` defaults to `false` to keep backwards compatibility
- I used the existing `testcharts/subchart` for testing and fixed Kubernetes warnings along the way when I was trying to install the chart locally. It's now valid and can be installed. 
  By using an existing test I did not have to write add and additional test case which check rendering if `skip-tests` is not used. f you prefer to create a different chart for testing just let me know and I will adapt.
- How should I update the documentation? Is there a make target or something else which can be used to generate cobra docs?

**If applicable**:

### DIFF
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -101,7 +101,7 @@ type Configuration struct {
 // TODO: This function is badly in need of a refactor.
 // TODO: As part of the refactor the duplicate code in cmd/helm/template.go should be removed
 //       This code has to do with writing files to disk.
-func (c *Configuration) renderResources(ch *chart.Chart, values chartutil.Values, releaseName, outputDir string, subNotes, useReleaseName, includeCrds bool, skipTests bool, pr postrender.PostRenderer, dryRun bool) ([]*release.Hook, *bytes.Buffer, string, error) {
+func (c *Configuration) renderResources(ch *chart.Chart, values chartutil.Values, releaseName, outputDir string, subNotes, useReleaseName, includeCrds bool, pr postrender.PostRenderer, dryRun bool) ([]*release.Hook, *bytes.Buffer, string, error) {
 	hs := []*release.Hook{}
 	b := bytes.NewBuffer(nil)
 
@@ -194,7 +194,7 @@ func (c *Configuration) renderResources(ch *chart.Chart, values chartutil.Values
 		}
 	}
 
-	for _, m := range filterManifests(manifests, skipTests) {
+	for _, m := range manifests {
 		if outputDir == "" {
 			fmt.Fprintf(b, "---\n# Source: %s\n%s\n", m.Name, m.Content)
 		} else {
@@ -222,19 +222,6 @@ func (c *Configuration) renderResources(ch *chart.Chart, values chartutil.Values
 	}
 
 	return hs, b, notes, nil
-}
-
-func filterManifests(manifests []releaseutil.Manifest, skipTests bool) []releaseutil.Manifest {
-	if skipTests {
-		var manifestsWithoutTests []releaseutil.Manifest
-		for _, m := range manifests {
-			if !strings.Contains(m.Name, "tests/") {
-				manifestsWithoutTests = append(manifestsWithoutTests, m)
-			}
-		}
-		return manifestsWithoutTests
-	}
-	return manifests
 }
 
 // RESTClientGetter gets the rest client
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -101,7 +101,7 @@ type Configuration struct {
 // TODO: This function is badly in need of a refactor.
 // TODO: As part of the refactor the duplicate code in cmd/helm/template.go should be removed
 //       This code has to do with writing files to disk.
-func (c *Configuration) renderResources(ch *chart.Chart, values chartutil.Values, releaseName, outputDir string, subNotes, useReleaseName, includeCrds bool, pr postrender.PostRenderer, dryRun bool) ([]*release.Hook, *bytes.Buffer, string, error) {
+func (c *Configuration) renderResources(ch *chart.Chart, values chartutil.Values, releaseName, outputDir string, subNotes, useReleaseName, includeCrds bool, skipTests bool, pr postrender.PostRenderer, dryRun bool) ([]*release.Hook, *bytes.Buffer, string, error) {
 	hs := []*release.Hook{}
 	b := bytes.NewBuffer(nil)
 
@@ -194,7 +194,7 @@ func (c *Configuration) renderResources(ch *chart.Chart, values chartutil.Values
 		}
 	}
 
-	for _, m := range manifests {
+	for _, m := range filterManifests(manifests, skipTests) {
 		if outputDir == "" {
 			fmt.Fprintf(b, "---\n# Source: %s\n%s\n", m.Name, m.Content)
 		} else {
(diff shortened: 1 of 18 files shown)


torstenwalter:
Sure. But if would cover at least the convention. In addition we could also filter out all resources with the hook.
Are you sure that the `"helm.sh/hook": test` can be added to other resources? I thought that helm expects that these are only added to pods, but I never tested it.

unguiculus:
It's not documented but it works. Found that out recently myself.

bacongobbler:
Jobs are another accepted type as test hooks, yes.

danylohetmantsev:
unfortunately when you running helm template with --no-hooks flag - it crashes on helm charts with hook: crd-install (such as velero, prometheus-operator and etc).
basically it doesn't do anything, tests with hook annotations are still templated as well as crd's.

bacongobbler:
@danylohetmantsev would you mind opening a new issue for that, please?

pkg/action/action.go
```diff
@@ -224,6 +224,19 @@ func (c *Configuration) renderResources(ch *chart.Chart, values chartutil.Values
 	return hs, b, notes, nil
 }
 
+func filterManifests(manifests []releaseutil.Manifest, skipTests bool) []releaseutil.Manifest {
+	if skipTests {
+		var manifestsWithoutTests []releaseutil.Manifest
+		for _, m := range manifests {
+			if !strings.Contains(m.Name, "tests/") {
```
bacongobbler, line 231: I would suggest looking for test hooks rather than searching within a test directory. Here's one way you could do this:

https://github.com/helm/helm/blob/c2da4fd53dc57189a46797a87b6b30dcd8d44879/pkg/action/release_testing.go#L81-L83

bacongobbler (changes requested):

bacongobbler:
> No matter if they have a `"helm.sh/hook": test` annotation or not.

Why this design decision? Doesn't that go against what is considered a test hook and what is not?

torstenwalter:
You are right that this is mainly a convention, but it's used by quite some charts and not to render these resources helps to get rid of ConfigMaps / Resources which are otherwise not needed.

Is there a better way to mark those resources as belonging to tests only?
If yes then it would be great if that could be documented somewhere. Then one could raise PRs for those helm charts to add the annotations and point to those docs.

It would still be helpful to exclude test resources from rendering and only tests, not all hooks.

pkg/action/action.go
torstenwalter: Thanks for the pointer. I will look into it.

bacongobbler:
What about charts that do not use a `tests` directory at all? If I write a test hook and place it under `templates/test-connection.yaml` then call `helm template --skip-tests`, I'd expect the test hook to be skipped.

> Is there a better way to mark those resources as belonging to tests only?

I do not think there is a way to exclude those from being rendered even in the usual manner. Resources that are not marked with the test hook annotation are considered normal resources, not tests.

I'm curious what issues are you running into here, and what problems are you trying to solve for.

rocketspacer:
@bacongobbler We're trying to apply gitops at our company. The flow expects raw k8s manifests rendered from `helm template`. Something like `helm template <release> . | kubectl apply -f -`
Test resources are not cleaned up in the same manner as when you use helm install/upgrade.

bacongobbler:
@rocketspacer I'm aware of the use case. I'm asking why is the PR designed in such a way that it fails to acknowledge common use cases like test hooks outside of the `tests/` directory, or why it considers templates inside the `tests/` directory as test hooks, when Helm itself does not consider that case.

mattfarina (changes requested):
First, let me say that I understand the need to make sure all test resources are handled and marked as tests.

If a chart has resources in it that are tests and do not have the test annotation, I would argue that chart has a bug.

Tests in the `tests` directory isn't a convention (conventions are agreed to patterns). It's something `helm create` does that many have copied and many have not.

This PR takes something that's not a convention and codifies it. It has impact on other tools that perform analysis and testing on charts (e.g., chart-testing).

Why can't the test hook be added to all the test resources?

torstenwalter:
@bacongobbler @mattfarina I was simply not aware of the fact that it's possible to have a `helm.sh/hook: test` annotation on resources which support tests such as ConfigMaps, Roles etc.

I just noticed that quite some charts have these kinds of resources and I have never seen those outside of tests directories. It's not my intention to codify something of there's a better solution.


https://helm.sh/docs/topics/chart_tests/ is not very clear - at least for me. My understanding from reading it was that the annotation should only be used for job definitions.

> A test in a helm chart lives under the templates/ directory and is a job definition that specifies a container with a given command to run. 

I read job definition as Kubernetes `Job` or a `Pod` as that is used as an example.

Are these annotation on other objects simply ignored or what's the behavior there? Was that always the case or only introduced with a specific helm version?
I am asking whether one could rely on that. If that's the case that it would be cleaner to just exclude resources with those annotations from rendering.

Together with that it would make sense to point it out in the documentation that annotations could also be used for resources supporting in tests.
That would help when creating PRs for existing charts which do not have these annotations as one could point to the docs.

I am happy to update this PR and also to help with the docs. Just want to be sure if that's goes in the right direction.

mattfarina:
@torstenwalter sorry for my slow response. I'll try to respond to your questions...

> https://helm.sh/docs/topics/chart_tests/ is not very clear - at least for me.

The docs in this section could use some major improvements. They are not clear enough. You are right.

> Are these annotation on other objects simply ignored or what's the behavior there?

When something has the test annotation it is ignored during a normal installation and only installed when tests are run. A deployment as a test isn't typically going to be a good idea so we point people to Jobs and Pods. When Helm [sorts the manifests as part of rendering](https://github.com/helm/helm/blob/6297c021cbda1483d8c08a8ec6f4a99e38be7302/pkg/action/action.go#L165) it [uses the hook annotation](https://github.com/helm/helm/blob/6297c021cbda1483d8c08a8ec6f4a99e38be7302/pkg/releaseutil/manifest_sorter.go#L133).

> Was that always the case or only introduced with a specific helm version?

This isn't new. It's this way for all of Helm v3. I'd need to dig into Helm v2 but I think it's that way there, too.

> I am asking whether one could rely on that. If that's the case that it would be cleaner to just exclude resources with those annotations from rendering.

You can rely on it. The docs need work. We'd love help with that.

torstenwalter:
@mattfarina @bacongobbler I changed it now to skip tests based on hook annotations.

unguiculus:
I like it now. 👍 

LGTM

torstenwalter:
@mattfarina @bacongobbler Any chance to get feedback on this PR?

immanuelfodor:
I really hope this feature will be merged and released soon, I'd love to skip rendering tests with the template command.

technosophos (approved):
I have reviewed the code and am happy with the state of this. It does not appear to break backward compatibility, and the solution looks correct to me.

mattfarina (approved):

torstenwalter:
@technosophos @mattfarina Thanks a lot.

PR #8707, opened 2020-09-09
Cleanup tempfiles

zhijianli88:
we introuduce a new testing feature T.cleanup to remove temp directory
T.cleanup was introduced since go-1.14: https://golang.org/pkg/testing/#T.Cleanup

**If applicable**:

### DIFF
--- a/pkg/repo/repotest/server.go
+++ b/pkg/repo/repotest/server.go
@@ -50,6 +50,8 @@ func NewTempServerWithCleanup(t *testing.T, glob string) (*Server, error) {
 //
 // The caller is responsible for destroying the temp directory as well as stopping
 // the server.
+//
+// Deprecated: use NewTempServerWithCleanup
 func NewTempServer(glob string) (*Server, error) {
 	tdir, err := ioutil.TempDir("", "helm-repotest-")
 	if err != nil {
--- a/pkg/repo/repotest/server.go
+++ b/pkg/repo/repotest/server.go
@@ -21,6 +21,7 @@ import (
 	"net/http/httptest"
 	"os"
 	"path/filepath"
+	"testing"
 
 	"helm.sh/helm/v3/internal/tlsutil"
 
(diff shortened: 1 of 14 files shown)


pkg/repo/repotest/server.go
```diff
@@ -36,11 +37,13 @@ import (
 //
 // The caller is responsible for destroying the temp directory as well as stopping
 // the server.
-func NewTempServer(glob string) (*Server, error) {
+func NewTempServer(t *testing.T, glob string) (*Server, error) {
```
bacongobbler, line 40: As this is a public function signature, we cannot change it due to backwards compatibility concerns as per our contributing guide (see CONTRIBUTING.md).

I would suggest instead introducing a new function called `NewTempServerWithCleanup(t *testing.T, glob string) (*Server, error) {}`, and add a comment above `NewTempServer` that this function has been deprecated in favour of `NewTempServerWithCleanup`. I'd also add an additional comment to replace `NewTempServer` with `NewTempServerWithCleanup` in Helm 4.

Thanks!

bacongobbler (changes requested):

pkg/repo/repotest/server.go
zhijianli88: Yeap, great thanks for your suggestion.

pkg/repo/repotest/server.go
```diff
@@ -29,13 +30,41 @@ import (
...
 
+// NewTempServerWithCleanup creates a server inside of a temp dir.
+//
+// If the passed in string is not "", it will be treated as a shell glob, and files
+// will be copied from that path to the server's docroot.
+//
+// The caller is responsible for stopping the server.
+// The temp dir will be removed by testing package automatically when test finished.
+func NewTempServerWithCleanup(t *testing.T, glob string) (*Server, error) {
+	tdir, err := ioutil.TempDir("", "helm-repotest-")
```
bacongobbler, line 41: To make this code more DRY, why not replace this all with the following:

```
func NewTempServerWithCleanup(t *testing.T, glob string) (*Server, error) {
    s, err := NewTempServer(glob)
    t.Cleanup(func() { os.RemoveAll(s.docroot) })
}
```

bacongobbler (changes requested):

bacongobbler (approved):
LGTM.

Issue #5304, opened 2019-02-13
User impersonation at resource level


hickeyma:
Install a chart with the ability to assign different persona per Kubernetes resources being created by a single chart.

This is similar to [Kubernetes user impersonation](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#user-impersonation) .

hickeyma:
Related to #5303

hickeyma:
@adamreese As we discussed.

bacongobbler:
Similar to [@adamreese's comment in #5303](https://github.com/helm/helm/issues/5303#issuecomment-463309170), Helm 3 now creates Kubernetes API objects as the user defined in their kube config. I am not sure how a chart would map to a specific user. Can you please describe your use case in a little more detail?

hickeyma:
This issue was brought to my attention by an Istio community member. I have unfortunately been unable to get more concrete details on the use case.

mtougeron:
I can help with a use-case that we have.

By default our users don't have permission to interact with `Secret` and similar resources. In order to do so they must be part of a group that is allowed to impersonate an "admin" user for their namespace. The RBAC rules define who is allowed to impersonate that admin user. e.g., `kubectl get secret -n ns1 --as ns1-admin --as-group ns1_admins`

Currently helm3, as far as I can tell, doesn't support the similar ability to pass along the `--as` and `--as-group` params to the Kubernetes API.

hickeyma:
Thanks @mtougeron, interesting use case. Am I correct in saying you only want the impersonation on that resource at not at chart level for all resources?

mtougeron:
I think at chart level when the command is executed is fine. e.g., `helm upgrade --install foo --as ns1-admin --as-group ns1_admins` (or whatever the appropriate cli args would be)

bacongobbler:
From looking at kubectl's implementation, `--as` and `--as-group` are override flags for fields set in your kubeconfig. In other words, Helm already has the ability to perform user impersonation; just modify your `$HOME/.kube/config` and away you go.



https://github.com/kubernetes/kubernetes/blob/30c7df5cd822067016640aa267714204ac089172/staging/src/k8s.io/cli-runtime/pkg/genericclioptions/config_flags.go#L146-L151

[...]

PR #8710, opened 2020-09-09
Implement Impersonation via flags/env

### LINKED ISSUE #5303: implement `helm --as` | `helm --as-group`
**Feature:**
Ability to install a chart for a different persona to the user calling the Helm CLI or Helm library.

**Use Case:**
A user logs into a portal that provides the ability to install charts.  The user is assigned to a team, which grants the user access to specific resources on the system:  namespaces, repositories, and charts within the repository.  The user is also assigned a role on the specific team which controls what actions the user may perform, such as a viewer or an editor.

On the UI, there is a catalog of charts to which the user has been given access.  The user selects one of the charts, updates the configuration values, and clicks `install`.  The request is sent to a backend processor for installing the chart.  The backend retrieves the user identify from an access token the user was assigned when logging into the portal, and the backend maps the teams to which the user is assigned to a set of groups for the user.

The backend service has been given the right to perform [user impersonation](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#user-impersonation), and includes the user identity and groups in impersonation headers that are sent to Kubernetes.  Kubernetes validates the backend service and since it is able to impersonate users, accepts the user identity and user groups passed in the impersonation headers.

The key is the user does not use the helm library directly but is used instead by the backend service on behalf of the user.

stealthybox:
This patch implements similar global flags + env to `kubectl --as=user --as-group=admins --as-group=operators` for helm.
The equivalent flags look like `helm --kube-as-user=user --kube-as-group=admins --kube-as-group=operators`.

Fixes #5303

The flags were prefixed with `kube-` -- this follows the existing flags and was mentioned in this commment:
> We will probably want to preface the flags with --kube- to match with the current convention (like how kubectl's --context flag matches with Helm's --kube-context flag).
> https://github.com/helm/helm/issues/5304#issuecomment-551306428

I also prepended an extra commit which updates the complete list of global ENV vars for the root CLI help-text.

**If applicable**:
- [*] this PR contains documentation
- [*] this PR contains unit tests
- [*] this PR has been tested for backwards compatibility

Changes:
- Support impersonation via flags similar to kubectl --as="user"
- Document all env vars for CLI help


bacongobbler (approved):
Thank you for this! I know many users have been asking for this feature. And thank you for writing out a few tests to go along with it. We really appreciate it.

LGTM!

PR #8720, opened 2020-09-11
Parse numbers as Float64 instead of Int64

### LINKED ISSUE #8719: Number values set by "--set" parsed in wrong type
In [crossplane provider-helm](https://github.com/crossplane-contrib/provider-helm), we're using helm 3 Go SDK and have implemented a k8s controller which observes actual state and installs/upgrades helm release's based on desired spec.

[Problem that we are having](https://github.com/crossplane-contrib/provider-helm/issues/19) right now is, when release spec has numbers set via "--set" option, we see continuous upgrades in each reconcile loop since observed and desired config does not match. We are using the same parsing function that helm cli uses, i.e. [strvals.ParseInto](https://github.com/crossplane-contrib/provider-helm/blob/master/pkg/controller/values.go#L83)

Putting our code aside, and debugging helm with `helm upgrade --install wp bitnami/wordpress --set port=1000` (i.e. running same upgrade couple of times), I observed the following:

Screen Shot 2020-09-11 at 14 33 04

Even I did use exact same configuration (`--set port=1000`), currentReleaes.Config["port"] has a different type than upgradedReleaes.Config["port"] (`float64` vs `int64`). And this is what is breaking our case (reflect.deepequal returns not equal and we run upgrade).

Digging further, I found the root cause of this issue is the way helm parsing numbers differs from golangs json package parses. This is only valid for `--set` case, because in case of `values` parsing, same json.Unmarshal is used and result is consistent.

[Helm parses numbers for --set](https://github.com/helm/helm/blob/9429af8b39c6888a41ffa2945d7c73676afb577e/pkg/strvals/parser.go#L440) as Int64:

```
func typedVal(v []rune, st bool) interface{} {
         ...
	// If this value does not start with zero, try parsing it to an int
	if len(val) != 0 && val[0] != '0' {
		if iv, err := strconv.ParseInt(val, 10, 64); err == nil {
			return iv
		}
	}
	return val
}
```

However, [json.Unmarshal parses as float64](https://golang.org/pkg/encoding/json/#Unmarshal):

> To unmarshal JSON into an interface value, Unmarshal stores one of these in the interface value:
> 
> bool, for JSON booleans
> float64, for JSON numbers <---
> string, for JSON strings
> []interface{}, for JSON arrays
> map[string]interface{}, for JSON objects
> nil for JSON null
> 

During deployment (install or upgrade) what happens is as follows (AFAIU):
1. Helm composes value overrides, using values and set's
1. In our specific example, there is no additional values, just one set, `--set port=1000`
1. Helm creates a Release struct where Config is a map[string]interface{} with int64 value as below:

Screen Shot 2020-09-11 at 16 43 43

1. Helm marshals this Release struct to json (encodes/compresses?) and stores it in a secret on k8s api
1. When helm wants to get that release object (e.g. `currentRelease` above, or get last release observation in our case), gets the secret content, (uncompress/decodes?) and unmarshals back to `Release` struct. At this exact point we can `Release.Config["port"]` variable as `float64` which was originally stored as `int64`.

Output of `helm version`: 
```
```

Output of `kubectl version`: 
```
Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.3", GitCommit:"2d3c76f9091b6bec110a5e63777c332469e0cba2", GitTreeState:"clean", BuildDate:"2019-08-19T11:13:54Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.9", GitCommit:"a17149e1a189050796ced469dbd78d380f2ed5ef", GitTreeState:"clean", BuildDate:"2020-05-01T02:22:44Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
```

Cloud Provider/Platform (AKS, GKE, Minikube etc.): Kind

turkenh:
Fixes #8719

Signed-off-by: Hasan Turken <turkenh@gmail.com>

**If applicable**:

### DIFF
--- a/pkg/strvals/parser.go
+++ b/pkg/strvals/parser.go
@@ -432,12 +432,12 @@ func typedVal(v []rune, st bool) interface{} {
 	}
 
 	if strings.EqualFold(val, "0") {
-		return int64(0)
+		return float64(0)
 	}
 
 	// If this value does not start with zero, try parsing it to an int
 	if len(val) != 0 && val[0] != '0' {
-		if iv, err := strconv.ParseInt(val, 10, 64); err == nil {
+		if iv, err := strconv.ParseFloat(val, 64); err == nil {
 			return iv
 		}
 	}


bacongobbler:
Hi @turkenh, we ask users to please provide unit tests along with their changes to verify their PRs fix the original issue. Can you please write a unit test to cover the case described in #8719? Thanks.,

bacongobbler (changes requested):
needs tests before approval.

bacongobbler:
ping.

bacongobbler:
closing as inactive.

PR #8727, opened 2020-09-14
fix(comp): Disable file comp for output formats

marckhouzam:

It does not make sense to suggest files to the user as output formats.
This PR fixed that.

Currently:
```
bash-5.0$ bin/helm status --output <TAB>
json   table  yaml
bash-5.0$ bin/helm status --output go.<TAB>
go.mod  go.sum
```
Notice that the `go.*` files are mistakenly suggested when the real completions don't match.
With this PR:
```
bash-5.0$ bin/helm status --output <TAB>
json   table  yaml
bash-5.0$ bin/helm status --output go.<TAB>
# No suggestions
```

Unit test has been corrected.

### DIFF
--- a/cmd/helm/flags.go
+++ b/cmd/helm/flags.go
@@ -69,7 +69,7 @@ func bindOutputFlag(cmd *cobra.Command, varRef *output.Format) {
 				formatNames = append(formatNames, format)
 			}
 		}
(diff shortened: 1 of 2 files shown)


bacongobbler (approved):
LGTM.

PR #8730, opened 2020-09-14
Add engine.NewWithClient func

### LINKED ISSUE #8725: Golang API: how to use `engine.RenderWithClient` in with `Strict` enabled
The `Strict bool` flag is public in the `engine.Engine` struct, but `client *rest.Client` is not.  The `RenderWithClient` func sets the `client`, but does not provide an opportunity (as far as I can see) to enable `Strict`.

Perhaps I am missing something?  Or can we change the struct to make the client reference public?

Thanks!

object88:
closes #8725 

Signed-off-by: Paul Brousseau <object88@gmail.com>

**If applicable**:

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -44,6 +44,13 @@ type Engine struct {
 	config *rest.Config
 }
 
+// NewWithClient creates a new instance of client-aware Engine
+func NewWithClient(config *rest.Config) *Engine {
+	return &Engine{
+		config: config,
+	}
+}
+
 // Render takes a chart, optional values, and value overrides, and attempts to render the Go templates.
 //
 // Render can be called repeatedly on the same engine.
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -44,11 +44,15 @@ type Engine struct {
 	config *rest.Config
 }
 
-// NewWithClient creates a new instance of client-aware Engine
-func NewWithClient(config *rest.Config) *Engine {
-	return &Engine{
-		config: config,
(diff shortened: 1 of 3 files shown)


bacongobbler:
hi @object88. Generally speaking we ask contributors to provide tests along with their PRs to ensure things are tested.

Have you played around with this PR yet? What do you think about the design? As it stands, right now you have to instantiate the Engine before setting the Strict flag:

```
e := engine.NewWithClient(client)
e.Strict = true
```

Perhaps there's a better way to improve the design here. What do you think?

bacongobbler:
I'm wondering maybe instead of a `NewWithClient`, perhaps a new function called `func (e *Engine) UseConfig(config *rest.Config)` might be slightly better here.

```
e := Engine{Strict: true, LintMode: true}
e.UseConfig(config)
```

Happy to hear your thoughts on this one.

object88:
Personally, I love using [`Option` funcs](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) pattern, and I see that's in use in the `getter` package.  The `engine` package doesn't have as many properties to contend with, but I find that it's a good way to ensure that one can expand later.  And it allows properties to be kept private (and concurrency-safe, which `UseConfig` may not, on an assumption).

It wasn't terribly obvious to be what unit tests would be meaningful for the initial PR.  But rewriting this with a `New(opts ...OptionFunc)` to set config, strict, etc., properties would certainly call for it.

object88:
Sorry for the delays in this; setting to draft while I get to reimplementing.

Issue #7747, opened 2020-03-09
Support for .yml extension for Chart.yaml/value.yaml


Tom-Br:
Output of `helm version`: version.BuildInfo{Version:"v3.1.0", GitCommit:"b29d20baf09943e134c2fa5e1e1cab3bf93315fa", GitTreeState:"clean", GoVersion:"go1.13.7"}

Output of `kubectl version`: Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.5", GitCommit:"20c265fef0741dd71a66480e35bd69f18351daea", GitTreeState:"clean", BuildDate:"2019-10-15T19:16:51Z", GoVe
rsion:"go1.12.10", Compiler:"gc", Platform:"windows/amd64"}
Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.6", GitCommit:"72c30166b2105cd7d3350f2c28a219e6abcd79eb", GitTreeState:"clean", BuildDate:"2020-01-18T23:23:21Z", GoVe
rsion:"go1.13.5", Compiler:"gc", Platform:"linux/amd64"}

I am trying to lint a local Helm chart. 

```
helm lint ./zookeeper
==> Linting ./zookeeper
Error unable to check Chart.yaml file in chart: CreateFile zookeeper\Chart.yml: The system cannot find the path specified.

Error: 1 chart(s) linted, 1 chart(s) failed

```

However if I run the same command with .yaml it works. Thus Helm is not handling .yml the same as .yaml

```
Y:\brecourt\infra\zookeeper\chart>helm lint ./zookeeper
==> Linting ./zookeeper
[INFO] Chart.yaml: icon is recommended
[INFO] values.yaml: file does not exist

1 chart(s) linted, 0 chart(s) failed
```

anshulvermapatel:
Hello @Tom-Br 

Actually, as far as I know it has to be 'Chart.yaml' , you cannot have 'Chart.yml'.
You can check this [Doc page](https://helm.sh/docs/topics/charts/) .
@bacongobbler must agree with me.

jdolitsky:
Certain files such as values.yaml, Chart.yaml are required by name (as noted above). Files in templates/ dir should be able to use the .yml extension if necessary.

@Tom-Br Going to close, renaming file to Chart.yaml should be the fix you need.

astropuffin:
[...]

PR #8731, opened 2020-09-14
improved user-facing error messages to explain the underlying problem

technosophos:

The issue #7747 is a little bit meandering, but at root, the solution is to fix our error messages to be clearer on why a chart load failed. This replaces a terse error message with one that is more easily understood.

I added some tests, and also modified some existing tests.

**If applicable**:

### DIFF
--- a/pkg/chart/loader/load.go
+++ b/pkg/chart/loader/load.go
@@ -143,6 +143,10 @@ func LoadFiles(files []*BufferedFile) (*chart.Chart, error) {
 		}
 	}
 
+	if c.Metadata == nil {
+		return c, errors.New("Chart.yaml file is missing")
+	}
+
 	if err := c.Validate(); err != nil {
 		return c, err
 	}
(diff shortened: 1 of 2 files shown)


bridgetkromhout:
This seems clearer, in that you are explicit about the name of the file Helm is expecting to find here. If someone has named their file with another name, that could provide a pointer as to what to fix.

bacongobbler:
This certainly clarifies the error message better for users of the CLI, and I appreciate the effort to improve the error message for these users.

However, at the point where `c.Validate()` is called, we've already loaded the chart in memory. Users who consume the SDK may be "loading" a chart from a JSON stream or constructing the chart in some manner that would not utilize a Chart.yaml. Some might even construct a chart by hand like we do in the unit tests. This error message may confuse those users as no Chart.yaml was consumed while loading their chart.

Is there a way we can instead improve the error message during [`LoadFiles`](https://github.com/helm/helm/blob/a96ceeac706951a46e604c94fcdbfb40c5fa7e26/pkg/chart/loader/load.go#L72) instead, [before `Validate` is called](https://github.com/helm/helm/blob/a96ceeac706951a46e604c94fcdbfb40c5fa7e26/pkg/chart/loader/load.go#L146-L148)? That way we can enforce the validation when a Chart.yaml is not present in an archive/directory.

technosophos:
@bacongobbler Done.

bacongobbler (approved):

bacongobbler:
LGTM!

mattfarina (approved):

PR #8742, opened 2020-09-16
engine: make template function map public

bashims:
Signed-off-by: Benjamin Ash <bash@intelerad.com>

**What this PR does / why we need it**: We have some template rendering needs that are satisfied by the template funcMap from pkg/engine.go.  Currently engine.funcMap() is private so we cannot use it in our code. This PR makes the function public.

**If applicable**:

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -105,7 +105,7 @@ func warnWrap(warn string) string {
 
 // initFunMap creates the Engine's FuncMap and adds context-specific functions.
 func (e Engine) initFunMap(t *template.Template, referenceTpls map[string]renderable) {
-	funcMap := funcMap()
+	funcMap := FuncMap()
 	includedNames := make(map[string]int)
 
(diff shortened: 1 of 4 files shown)


bacongobbler:
`FuncMap` was intentionally made private to ensure that we could make changes to the internal function mapping over time due to our promises on backwards compatibility. For example, if FuncMap was made public prior to this change, introducing the `lookup` function would've broken backwards compatibility as templates consuming this function might've broken with that change.

Because of that design constraint, I'm hesitant to merge this PR.

bashims:
Okay, I understand, although it is a bit of a bummer. Perhaps there is an alternative solution for our use case. Currently we want to render HelmRelease templates that are then handled by Flux+Helm-Operator. Would you have any recommendations for doing this aside from what I have done in this PR?

bacongobbler:
You could use the post renderer. More info here: https://helm.sh/docs/topics/advanced/#post-rendering

PR #8746, opened 2020-09-17
Tweak LDFLAGS in Makefile

jimmidyson:

This PR tweaks the use of LDFLAGS in the Makefile to allow overriding LDFLAGS at the time of running `make`, while still retaining the necessary version linker flags. This was a problem in helm package in Archlinux, which overrides linker flags with:

```
make LDFLAGS="-s -w -linkmode external" GOFLAGS="-buildmode=pie -trimpath"
```

Due to the nature of make variables, this meant that the necessary version linker flags do not include the version linker flags, leading to `helm version` being incorrect:

```
$ helm version
version.BuildInfo{Version:"v3.3", GitCommit:"", GitTreeState:"", GoVersion:"go1.15.2"}
```

After these changes, the helm binary is correctly built and reports version correctly:

```
$ helm version
version.BuildInfo{Version:"v3.3.1", GitCommit:"249e5215cde0c3fa72e27eb7a30e8d55c9696144", GitTreeState:"clean", GoVersion:"go1.15.2"}
```

This PR also consistently uses parentheses for make variables (parentheses and braces are interchangeable in a Makefile so this is just for tidiness, not functionality).

**If applicable**:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,24 +33,23 @@ GIT_DIRTY  = $(shell test -n "`git status --porcelain`" && echo "dirty" || echo 
 
 ifdef VERSION
 	BINARY_VERSION = $(VERSION)
+else
+	BINARY_VERSION = $(GIT_TAG)
 endif
-BINARY_VERSION ?= ${GIT_TAG}
 
 # Only set Version if building a tag or VERSION is set
 ifneq ($(BINARY_VERSION),)
-	LDFLAGS += -X helm.sh/helm/v3/internal/version.version=${BINARY_VERSION}
+	VERSION_LDFLAGS += -X helm.sh/helm/v3/internal/version.version=$(BINARY_VERSION)
 endif
 
-VERSION_METADATA = unreleased
-# Clear the "unreleased" string in BuildMetadata
-ifneq ($(GIT_TAG),)
-	VERSION_METADATA =
+# Set the "unreleased" string in BuildMetadata
+ifeq ($(GIT_TAG),)
+	VERSION_METADATA = unreleased
 endif
 
-LDFLAGS += -X helm.sh/helm/v3/internal/version.metadata=${VERSION_METADATA}
-LDFLAGS += -X helm.sh/helm/v3/internal/version.gitCommit=${GIT_COMMIT}
-LDFLAGS += -X helm.sh/helm/v3/internal/version.gitTreeState=${GIT_DIRTY}
-LDFLAGS += $(EXT_LDFLAGS)
+VERSION_LDFLAGS += -X helm.sh/helm/v3/internal/version.metadata=${VERSION_METADATA}
+VERSION_LDFLAGS += -X helm.sh/helm/v3/internal/version.gitCommit=${GIT_COMMIT}
+VERSION_LDFLAGS += -X helm.sh/helm/v3/internal/version.gitTreeState=${GIT_DIRTY}
 
 .PHONY: all
 all: build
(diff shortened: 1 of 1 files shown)


jimmidyson:
@Foxboron FYI - I've seen your recent PR about this and wonder if you could review this please?

Foxboron:
Hmm, I think this is cleaner in the long run as you get more control in the build system if you need it without sacrificing current usage . Admittedly the `$EXT_LDFLAGS` patch was the easier option then reworking the `$LDFLAGS` handling.

jimmidyson:
@bacongobbler I saw that you reviewed the previous PR that touched linker flags - would you mind taking a look at this one too please?

Makefile
```diff
@@ -32,25 +32,24 @@ GIT_TAG    = $(shell git describe --tags --abbrev=0 --exact-match 2>/dev/null)
...
+BINARY_VERSION = $(VERSION)
+else
+BINARY_VERSION = $(GIT_TAG)
 endif
-BINARY_VERSION ?= ${GIT_TAG}
 
 # Only set Version if building a tag or VERSION is set
 ifneq ($(BINARY_VERSION),)
-	LDFLAGS += -X helm.sh/helm/v3/internal/version.version=${BINARY_VERSION}
+VERSION_LDFLAGS += -X helm.sh/helm/v3/internal/version.version=$(BINARY_VERSION)
```
mattfarina, line 42: Out of curiosity, why did you remove the indentation of the body of the conditionals?

mattfarina (changes requested):

mattfarina (changes requested):
Thanks for the PR and for making this easier to build in other environments.

Aside from my one question, this PR changes the `build-cross` make target which accepts the `LDFLAGS`. For example, the EXT_LDFLAGS is no longer part of it. The `build-cross` target should be updated in a similar manner to the way the `build` target was updated.

Makefile
jimmidyson: I believe it's more idiomatic to not indent conditionals - see https://www.gnu.org/software/make/manual/html_node/Conditional-Syntax.html for examples that show no indentation.

jimmidyson:
@mattfarina Good spot about the `build-cross` target - thanks! Updated and pushed.

jimmidyson:
@mattfarina Is there anything else in this PR that needs addressing?

Makefile
mattfarina: The conditional examples in the manual have an indent. See https://www.gnu.org/software/make/manual/html_node/Conditional-Example.html#Conditional-Example. I was looking at those two pages before I made my comment because I wanted to find out what was idiomatic. I couldn't tell because those two pages in the manual aren't consistent.

Do you have another piece of information to show that it's idiomatic to not indent? If not, to be consistent with what we have had I would ask to keep the indent. It's a little easier to read.

jimmidyson: @mattfarina Fair point. I've reverted the indentation.

mattfarina:
I'm curious why you couldn't use the existing `EXT_LDFLAGS` to pass in `-linkmode external`. The `-s -w` was already set by default.

Is there something I'm missing? It could entirely be the case.

jimmidyson:
Hi @mattfarina sorry for late reply. My reasoning is in https://github.com/helm/helm/pull/8746#issuecomment-694154281. If you don't feel that is sufficient reason, then let's close this pr, no worries.

PR #8773, opened 2020-09-20
Log config permission warning to stderr

Lucaber:

This PR changes the writer of the warning
```
WARNING: Kubernetes configuration file is group-readable. This is insecure. Location: /home/lucaber/.kube/config
WARNING: Kubernetes configuration file is world-readable. This is insecure. Location: /home/lucaber/.kube/config
```
to the writer of the default logger, which is `stderr`.

This is needed to always show the warning to the user instead of it being parsed by other tools.

Im running `source <(helm completion zsh)` in my `.zshrc` and noticed this error on every start of a new terminal:
```/proc/self/fd/11:1: command not found: WARNING:```
In this case it is not even visible to the user that helm caused the issue.
With this PR the full warning is displayed to the user instead.

We could also use the logger `log.Warn`? in `checkPerms` directly.

**If applicable**:

### DIFF
--- a/cmd/helm/root.go
+++ b/cmd/helm/root.go
@@ -205,7 +205,7 @@ func newRootCmd(actionConfig *action.Configuration, out io.Writer, args []string
 	loadPlugins(cmd, out)
 
 	// Check permissions on critical files
-	checkPerms(out)
+	checkPerms(log.Writer())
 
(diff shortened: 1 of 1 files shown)


bacongobbler:
fixed in helm 3.3.4.

PR #8785, opened 2020-09-22
fix: explicitly set namespace on KubeClient

### LINKED ISSUE #8685: Go SDK upgrade performing actions in the wrong namespace
Output of `helm version`: 
```
version.BuildInfo{Version:"v3.3.1", GitCommit:"249e5215cde0c3fa72e27eb7a30e8d55c9696144", GitTreeState:"clean", GoVersion:"go1.14.7"}
```

Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.0", GitCommit:"9e991415386e4cf155a24b1da15becaa390438d8", GitTreeState:"clean", BuildDate:"2020-03-25T14:58:59Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.6", GitCommit:"dff82dc0de47299ab66c83c626e08b245ab19037", GitTreeState:"clean", BuildDate:"2020-07-15T16:51:04Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
```

Cloud Provider/Platform (AKS, GKE, Minikube etc.): Rancher on bare-metal

I am writing an application using the Helm Go SDK which is supposed to upgrade an already deployed helm chart on my cluster. The chart is `bitnami/kafka` and is deployed in namespace `kafka`. All works well apart from the fact that for some reason when I actually perform the upgrade, helm starts creating StatefulSets and ConfigMaps in my default namespace and deletes things from the `kafka` namespace. I already spent 1 day investigating this issue and I am nowhere so some help would be appreciated.

```golang
package main

import (
	"log"
	"os"

	"helm.sh/helm/v3/pkg/action"
	"helm.sh/helm/v3/pkg/chart/loader"
	"helm.sh/helm/v3/pkg/cli"
)

func main() {
	settings := cli.New()

	// You can pass an empty string instead of settings.Namespace() to list
	// all namespaces
	if err := actionConfig.Init(settings.RESTClientGetter(), "kafka", os.Getenv("HELM_DRIVER"), log.Printf); err != nil {
		log.Printf("%+v", err)
		os.Exit(1)
	}

	// Get current state
	getClient := action.NewGetValues(actionConfig)
	release, err := getClient.Run("kafka")
	if err != nil {
		log.Printf("%+v", err)
		os.Exit(1)
	}

	currentReplicaCount, ok := release["replicaCount"].(float64)
	if !ok {
		log.Printf("The current replica count %s is not a number.", release["replicaCount"])
		os.Exit(1)
	}
	log.Printf("Replica count: %+v", currentReplicaCount)

	// Upgrade
	upgradeClient := action.NewUpgrade(actionConfig)

	upgradeClient.Namespace = "kafka"
	upgradeClient.Atomic = true
	upgradeClient.ReuseValues = true

	//upgradeClient.DryRun = true
	updatedValues := map[string]interface{}{
		"replicaCount": currentReplicaCount + 1,
	}

	chartPath, err := upgradeClient.ChartPathOptions.LocateChart("bitnami/kafka", settings)
	if err != nil {
		log.Printf("%+v", err)
	}

	ch, err := loader.Load(chartPath)
	if err != nil {
		log.Printf("%+v", err)
	}

	upgrade, err := upgradeClient.Run("kafka", ch, updatedValues)
	if err != nil {
		log.Printf("%+v", err)
		os.Exit(1)
	}

	log.Printf("%+v", upgrade)
}
```

 - Retrieving the current config works fine
 - Dry run of the upgrade seems to return correct values (all namespaces are `kafka`)
 - Actual execution output:
```
2020/09/02 09:18:52 &{cfg:0xc000057d80 ChartPathOptions:{CaFile: CertFile: KeyFile: InsecureSkipTLSverify:false Keyring: Password: RepoURL: Username: Verify:false Version:} Install:false Devel:false Namespace:kafka SkipCRDs:false Timeout:0s Wait:false DisableHooks:false DryRun:false Force:false ResetValues:false ReuseValues:true Recreate:false MaxHistory:0 Atomic:true CleanupOnFail:false SubNotes:false Description: PostRenderer:<nil> DisableOpenAPIValidation:false}
2020/09/02 09:18:53 preparing upgrade for kafka
2020/09/02 09:18:53 reusing the old release's values
2020/09/02 09:18:54 performing update for kafka
2020/09/02 09:18:54 creating upgraded release for kafka
2020/09/02 09:18:54 checking 10 resources for changes
2020/09/02 09:18:54 Looks like there are no changes for PodDisruptionBudget "kafka-zookeeper"
2020/09/02 09:18:54 Looks like there are no changes for PodDisruptionBudget "kafka"
2020/09/02 09:18:54 Looks like there are no changes for ServiceAccount "kafka"
2020/09/02 09:18:54 Created a new ConfigMap called "kafka-scripts" in default
2020/09/02 09:18:54 Looks like there are no changes for Service "kafka-zookeeper-headless"
2020/09/02 09:18:55 Looks like there are no changes for Service "kafka-zookeeper"
2020/09/02 09:18:55 Looks like there are no changes for Service "kafka-headless"
2020/09/02 09:18:55 Created a new StatefulSet called "kafka" in default
2020/09/02 09:18:55 beginning wait for 10 resources with timeout of 0s
2020/09/02 09:18:57 StatefulSet is not ready: default/kafka. 0 out of 3 expected pods are ready
```

P.S.: The output of ` helm upgrade -n kafka --reuse-values --set replicaCount=4 kafka bitnami/kafka --debug`

```
upgrade.go:121: [debug] preparing upgrade for kafka
upgrade.go:440: [debug] reusing the old release's values
upgrade.go:129: [debug] performing update for kafka
upgrade.go:308: [debug] creating upgraded release for kafka
client.go:173: [debug] checking 10 resources for changes
client.go:436: [debug] Looks like there are no changes for PodDisruptionBudget "kafka-zookeeper"
client.go:436: [debug] Looks like there are no changes for PodDisruptionBudget "kafka"
client.go:436: [debug] Looks like there are no changes for ServiceAccount "kafka"
client.go:436: [debug] Looks like there are no changes for ConfigMap "kafka-scripts"
client.go:436: [debug] Looks like there are no changes for Service "kafka-zookeeper-headless"
client.go:436: [debug] Looks like there are no changes for Service "kafka-zookeeper"
client.go:436: [debug] Looks like there are no changes for Service "kafka-headless"
upgrade.go:136: [debug] updating status for upgraded release for kafka
Release "kafka" has been upgraded. Happy Helming!
```

### LINKED ISSUE #8780: Go SDK installing Kubernetes resources to the wrong namespaces
I am working on a standalone Go binary that uses the `helm` Go APIs to deploy a chart.  The tricky part is that the binary is a long-running service running in a particular namespace, and it's deploying arbitrary helm charts to other, arbitrary namespaces.  The perhaps more tricky part is that this binary is expected to be concurrency-safe, so more than one deployment may happen at once.

The problem that I am experiencing is when that the service (running in namespace `nhi`) deploys a chart, the helm chart _itself_ lands in the desired namespaces ("default"), but the kubernetes resources described by the chart land in the `nhi` namespaces.

This situation has been discussed in a number of other issues, but I am not clear at this point if this is the expected behavior (and if so, I don't understand why), or if it's an acknowledged bug.  I _think_ that I have read (and I may be wrong) that the target namespace get derived from the kube context, but I am not sure that works for me given the constraints mentioned above.

I have a demo project that illustrates the problem: https://github.com/object88/namespaced-helm-install. Relevant code snippet:

``` go
	c := cli.New()

	a := &action.Configuration{
		Capabilities: chartutil.DefaultCapabilities,
	}
	debug := func(msg string, args ...interface{}) {
		fmt.Printf(fmt.Sprintf("** %s\n", msg), args...)
	}
	a.Init(c.RESTClientGetter(), "default", "", debug)

	act := action.NewInstall(a)
	act.Atomic = true
	act.Namespace = "default"
	act.ReleaseName = "foo"

	ch, _ := loader.Load("/opt/charts/foo")
	act.Run(ch, nil)
```

The `test.sh` command will build and push the "server" docker image to the `nhi` namespace, and deploy the server.  (if you run this, you may need to manually create that namespace.  Also, you will need to comment this out code to build and publish the docker image, or change to a different dockerhub repo)  The server then attempts to deploy a different chart, "foo" to the "default" namespace.  After running `test.sh`, you can run the following commands:

```
$ helm list --all-namespaces
NAME              	NAMESPACE	REVISION	UPDATED                                	STATUS  	CHART                        	APP VERSION
foo               	default  	1       	2020-09-21 17:46:08.694358671 +0000 UTC	deployed	foo-0.0.0
nhi               	nhi      	1       	2020-09-21 10:46:06.468626 -0700 PDT   	deployed	namespaced-helm-install-0.1.0	1.16.0

$ kubectl get all --namespace default
NAME                 TYPE        CLUSTER-IP   EXTERNAL-IP   PORT(S)   AGE
service/kubernetes   ClusterIP   10.96.0.1    <none>        443/TCP   34d

$ kubectl get all --namespace nhi
NAME                                              READY   STATUS    RESTARTS   AGE
pod/foo-nginx-deployment-6696cf858-2cqrc          1/1     Running   0          5m8s
pod/nhi-namespaced-helm-install-9fc497b7d-svqlw   1/1     Running   0          5m11s

NAME                                  TYPE        CLUSTER-IP     EXTERNAL-IP   PORT(S)   AGE
service/nhi-namespaced-helm-install   ClusterIP   10.104.221.1   <none>        80/TCP    5m11s

NAME                                          READY   UP-TO-DATE   AVAILABLE   AGE
deployment.apps/foo-nginx-deployment          1/1     1            1           5m9s
deployment.apps/nhi-namespaced-helm-install   1/1     1            1           5m11s

NAME                                                    DESIRED   CURRENT   READY   AGE
replicaset.apps/foo-nginx-deployment-6696cf858          1         1         1       5m8s
replicaset.apps/nhi-namespaced-helm-install-9fc497b7d   1         1         1       5m11s
```

Contents of `go.mod`:

```
module ...

go 1.15

require (
	[SNIP]
	helm.sh/helm/v3 v3.3.3
)
```

Output of `kubectl version`:

```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.2", GitCommit:"f5743093fd1c663cb0cbc89748f730662345d44d", GitTreeState:"clean", BuildDate:"2020-09-16T21:51:21Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.3", GitCommit:"2e7996e3e2712684bc73f0dec0200d64eec7fe40", GitTreeState:"clean", BuildDate:"2020-05-20T12:43:34Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
```

`minikube`

Thanks for any feedback, corrections for my code, or otherwise.

object88:

There seems to be a problem in the Go SDK that deploys k8s resources to the namespace of the kube config, instead of a namespace specified by the action.  See the linked issues below.

This requires testing, etc.,  The PR is submitted to get a quick feedback, i.e., "this is not the right idea at all", or "yeah, seems like it could be legit".  Will write tests, etc.

**If applicable**:

closes #8780
closes #8685

### DIFF
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -424,7 +424,6 @@ func (c *Configuration) Init(getter genericclioptions.RESTClientGetter, namespac
 		client.Namespace = namespace
 		return client
 	}
-	c.KubeClient = nil
 	c.Releases = store
 	c.Log = log
 
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -77,10 +77,14 @@ var (
 // https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
 var ValidName = regexp.MustCompile(`^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`)
 
+type DoMake func(namespace string) kube.Interface
+
 // Configuration injects the dependencies that all actions share.
 type Configuration struct {
 	// RESTClientGetter is an interface that loads Kubernetes clients.
-	RESTClientGetter RESTClientGetter
+	RESTClientGetter genericclioptions.RESTClientGetter
+
+	Do DoMake
 
 	// Releases stores records of releases.
 	Releases *storage.Storage
@@ -365,7 +369,6 @@ func (c *Configuration) recordRelease(r *release.Release) {
 func (c *Configuration) Init(getter genericclioptions.RESTClientGetter, namespace, helmDriver string, log DebugLog) error {
 	kc := kube.New(getter)
 	kc.Log = log
-	kc.Namespace = namespace
 
 	lazyClient := &lazyClient{
 		namespace: namespace,
@@ -413,6 +416,12 @@ func (c *Configuration) Init(getter genericclioptions.RESTClientGetter, namespac
 	}
 
 	c.RESTClientGetter = getter
+	c.Do = func(namespace string) kube.Interface {
+		client := kube.New(c.RESTClientGetter)
+		client.Log = c.Log
+		client.Namespace = namespace
+		return client
+	}
 	c.KubeClient = kc
 	c.Releases = store
 	c.Log = log
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -77,19 +77,21 @@ var (
 // https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
 var ValidName = regexp.MustCompile(`^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`)
 
-type DoMake func(namespace string) kube.Interface
+// GetKubeClientFunc returns a Kube client with the specified namespace
+type GetKubeClientFunc func(namespace string) kube.Interface
 
 // Configuration injects the dependencies that all actions share.
 type Configuration struct {
 	// RESTClientGetter is an interface that loads Kubernetes clients.
 	RESTClientGetter genericclioptions.RESTClientGetter
 
-	Do DoMake
+	// GetKubeClient returns an instance of a Kubernetes API client.
+	GetKubeClient GetKubeClientFunc
 
 	// Releases stores records of releases.
 	Releases *storage.Storage
 
-	// KubeClient is a Kubernetes API client.
+	// Deprecated: KubeClient should not be used; use GetKubeClient instead
 	KubeClient kube.Interface
 
 	// RegistryClient is a client for working with registries
@@ -416,13 +418,13 @@ func (c *Configuration) Init(getter genericclioptions.RESTClientGetter, namespac
 	}
 
 	c.RESTClientGetter = getter
-	c.Do = func(namespace string) kube.Interface {
+	c.GetKubeClient = func(namespace string) kube.Interface {
 		client := kube.New(c.RESTClientGetter)
 		client.Log = c.Log
 		client.Namespace = namespace
 		return client
 	}
-	c.KubeClient = kc
+	c.KubeClient = nil
 	c.Releases = store
 	c.Log = log
 
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -365,6 +365,7 @@ func (c *Configuration) recordRelease(r *release.Release) {
 func (c *Configuration) Init(getter genericclioptions.RESTClientGetter, namespace, helmDriver string, log DebugLog) error {
 	kc := kube.New(getter)
 	kc.Log = log
+	kc.Namespace = namespace
 
 	lazyClient := &lazyClient{
 		namespace: namespace,
--- a/cmd/helm/helm.go
+++ b/cmd/helm/helm.go
@@ -35,6 +35,7 @@ import (
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/cli"
 	"helm.sh/helm/v3/pkg/gates"
+	"helm.sh/helm/v3/pkg/kube"
 	kubefake "helm.sh/helm/v3/pkg/kube/fake"
 	"helm.sh/helm/v3/pkg/release"
 	"helm.sh/helm/v3/pkg/storage/driver"
@@ -130,7 +131,9 @@ func loadReleasesInMemory(actionConfig *action.Configuration) {
 		return
 	}
 
-	actionConfig.KubeClient = &kubefake.PrintingKubeClient{Out: ioutil.Discard}
+	actionConfig.GetKubeClient = func(namespace string) kube.Interface {
+		return &kubefake.PrintingKubeClient{Out: ioutil.Discard}
+	}
 
 	for _, path := range filePaths {
 		b, err := ioutil.ReadFile(path)
--- a/cmd/helm/helm_test.go
+++ b/cmd/helm/helm_test.go
@@ -112,11 +112,11 @@ func executeActionCommandStdinC(store *storage.Storage, in *os.File, cmd string)
 	fake := &kubefake.PrintingKubeClient{Out: ioutil.Discard}
 
 	actionConfig := &action.Configuration{
-		Do: func(namespace string) kube.Interface {
+		GetKubeClient: func(namespace string) kube.Interface {
 			return fake
 		},
 		Releases:     store,
-		KubeClient:   fake,
+		KubeClient:   nil,
 		Capabilities: chartutil.DefaultCapabilities,
 		Log:          func(format string, v ...interface{}) {},
 	}
--- a/cmd/helm/helm_test.go
+++ b/cmd/helm/helm_test.go
@@ -32,6 +32,7 @@ import (
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/chartutil"
 	"helm.sh/helm/v3/pkg/cli"
+	"helm.sh/helm/v3/pkg/kube"
 	kubefake "helm.sh/helm/v3/pkg/kube/fake"
 	"helm.sh/helm/v3/pkg/release"
 	"helm.sh/helm/v3/pkg/storage"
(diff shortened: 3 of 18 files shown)


bacongobbler:
Hi @object88. We ask for contributors to provide tests along with their changes to ensure the fix actually ensures it fixes the issue. It also self-documents what you are trying to fix, and ensures regressions won’t be introduced in the future. Thanks!

bridgetkromhout:
Hi, @object88 - in order for us to merge if/when the time comes, we do need a DCO signature; thanks! https://github.com/helm/helm/pull/8785/checks?check_run_id=1147227884

pkg/action/action.go
```diff
@@ -365,6 +365,7 @@ func (c *Configuration) recordRelease(r *release.Release) {
 func (c *Configuration) Init(getter genericclioptions.RESTClientGetter, namespace, helmDriver string, log DebugLog) error {
 	kc := kube.New(getter)
 	kc.Log = log
+	kc.Namespace = namespace
```
mattfarina, line 368: This is a place to set the namespace but it causes a problem.

The Release secrets can be installed into a different namespace from the release manifests. The Helm client doesn't do that. But, because the namespace is passed in one place for the action config and a separate place for the actions this is an option for users of the API.

If it is set here than the namespace for the two things will be the same and the namespace passed to the actions themselves will be ignored. So, we should set the namespace elsewhere.

We need to find a different place to set it and, while doing so, make it concurrency safe.

mattfarina (changes requested):

pkg/action/action.go
object88: Ah, thank you!  I had a feeling that there were some wrinkles I was not aware of.  We have some discussion going both here and in the related issue -- where is the best place to continue?  I have an _idea_ of how to address it, but I don't know if it would feel right to the `helm` team.

mattfarina: @object88 I have an idea how to fix it but I dunno if the rest of the client team would agree. Gimmie a little bit to confer with some of them.

"I think" a new KubeClient needs to be created inside of each action (i.e., the `Run` or `New` function). There is can be created in a similar manner to actionconfig `Init` but have the correct namespace set. This would apply to each of the actions.

This way the namespace setting is separate and it's concurrency safe.

mattfarina: @object88 The thought on KubeClients in each action for that action and kept private to it looks like a good way to go. Did you want to take a shot at writing that? If not, I can.

object88: I would like to.  I'm enjoying digging into this, and it's relevant to a project I'm working on, so I'm motivated.  If I find that I am dropping my focus, I will let you know.

pkg/action/action.go
```diff
@@ -77,10 +77,14 @@ var (
 // https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
 var ValidName = regexp.MustCompile(`^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`)
 
+type DoMake func(namespace string) kube.Interface
```
object88, line 80: TODO: come up with a better name.

pkg/action/action.go
object88: I took a shot at implementing the idea.  Can't say if I landed on target, so any feedback is appreciated.  PR currently only changes the `Install` action, and there's documentation & testing to do, and, well, maybe `Do` isn't a very good name, etc., etc.

The thought is that the `Configuration` can provide a `kube.Interface` to use within the scope of a `Run`, with the [namespace established](https://github.com/helm/helm/pull/8785/files#diff-f1c3ff8851f4e702c61ff00f3d76583bR419-R424) at that time.  The original `KubeConfig` is still there unchanged, but it's not used... so, why is it there (aside from the fact that removing it would be a breaking API change)?

If this looks like it's the right direction, I will make the same change across other actions, fix up any unit tests (unsure if any need to be added, aside from a concurrency test?), and document the changes.  If not, I am more than happy to take it in another direction.

object88: Hello @mattfarina , I took a shot at implementing the rest of the changes.  Please let me know if there are dangerous edges or anything that I've missed.  Thanks!

bacongobbler:
does https://github.com/helm/helm/pull/10230 solve your issue?

benoittgt:
I think it can be close, and the major change will come with Helm 4 ?

PR #8795, opened 2020-09-23
Adding annotation to index.yaml file

### LINKED ISSUE #8767: Feature: Add annotations field to the index.yaml file
Systems may want to add metadata to an index.yaml file. This feature would be to add an `annotations` top level field, similar to the one in `Chart.yaml` that can be used to add extra information.

mattfarina:
Chart.yaml files have an annotation field that allow a chart to
have custom information similar to the way Kubernetes annotations
work.

In an index.yaml file each chart version can have annotations in
a similar manner to the Chart.yaml file. It is derived from the
same underlying struct.

These enable extension points where people can add their own info.

One thing missing is the ability to extend the top level of an
index file. This change adds annotations to the top level of an
index.yaml file. This would provide top level support for vendors
to extent index.yaml files.

Closes #8767 

Vendors have started to extend index.yaml files with their own metadata. This provides top level support in index.yaml files

**If applicable**:

### DIFF
--- a/pkg/repo/index.go
+++ b/pkg/repo/index.go
@@ -83,6 +83,10 @@ type IndexFile struct {
 	Generated  time.Time                `json:"generated"`
 	Entries    map[string]ChartVersions `json:"entries"`
 	PublicKeys []string                 `json:"publicKeys,omitempty"`
+
+	// Annotations are additional mappings uninterpreted by Helm. They are made available for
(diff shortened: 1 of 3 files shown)


bacongobbler (approved):

jdolitsky (approved):
👍

PR #8829, opened 2020-09-30
Cleanup tempfiles

zhijianli88:
After the PR, we can cleanup all remaining tempdir

lizj@FNSTPC:~/workspace/k8s/helm$ export TMPDIR=/tmp/xxxxxx ; make clean; make test >/dev/null && echo PASS; ls -l $TMPDIR
PASS
total 0

### DIFF
--- a/internal/test/ensure/ensure.go
+++ b/internal/test/ensure/ensure.go
@@ -27,7 +27,7 @@ import (
 )
 
 // HelmHome sets up a Helm Home in a temp dir.
-func HelmHome(t *testing.T) func() {
+func HelmHome(t *testing.T) {
 	t.Helper()
 	base := TempDir(t)
 	os.Setenv(xdg.CacheHomeEnvVar, base)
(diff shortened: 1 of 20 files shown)


internal/test/ensure/ensure.go
```diff
@@ -36,9 +36,6 @@ func HelmHome(t *testing.T) func() {
 	os.Setenv(helmpath.CacheHomeEnvVar, "")
 	os.Setenv(helmpath.ConfigHomeEnvVar, "")
 	os.Setenv(helmpath.DataHomeEnvVar, "")
-	return func() {
-		os.RemoveAll(base)
```
provPaulBrousseau, line 40: It might be useful to leave the directory in place if the test fails.

zhijianli88: indeed, but currently, the HelmHome() caller always call it as 'defer HelmHome(t)() ' which has the same behaviors with T.cleanup() i guess.

provPaulBrousseau: Apologies; I wasn't clear with my comment.  I don't think that `helm` should keep the returned func (I like how it's cleaned up), but rather, in the func called by `t.Cleanup`, it might be good to check `t.Failed` before calling `os.RemoveAll`, because if a test fails, it could be useful to look in those directories for the state of config, cache, etc.  Maybe that's a different PR & discussion.

zhijianli88: thanks for your explanation, i just append a commit to do this

zhijianli88:
Kindly ping

zhijianli88:
i just rebased the PR, anyone could help to have a review ?

zhijianli88:
Ping

PR #8831, opened 2020-09-30
Caching downloaded charts

juniorz:
## Caching downloaded charts

Introduces a caching mechanism for chart downloads (#4618).

### Current design

Charts are mainly downloaded via `downloader.ChartDownloader.DownloadTo()` which:
- Resolve a chart URL from a "chart reference" - either a repo ref. (i.e. `repo/chart`) or a URL ref (i.e. `https://charts.io/chart-0.0.1.tgz`)
- Downloads the chart archive and its provenance file (if needed)
- Stores the downloaded files at a specific path
- Depending on the "verification strategy", verify the chart provenance

This feature is used in:

- `helm install` (locate chart & download dependencies)
- `helm pull` (download chart)
- `helm dependency build` (download dependencies)
- `helm dependency update` (download dependencies)
- `helm package` (download dependnecies)
- `helm show` (locate chart)
- `helm upgrade` (locate chart)

Some of these usages already use the concept of a charts cache by storing the chart files either in the repository cache dir or a temporary directory.

The reason is either the (a) need of an intermediate validation step between fetching and storing the chart archive on the destination or (b) the convenience of passing around a file path between parts of the process.

### Proposed design

The proposed design introduces a `HELM_CHART_CACHE` which (a) serves as permanent workinig directory for working with downloaded chart files during a single operation, and (b) can be safely persisted and reused among subsequent operations.

The API `downloader.ChartDownloader.Fetch()` is introduced. This API is responsible for iimplementing caching semantics for chart downloads, and works as follow:

- Resolves a chart URL.
- Downloads chart archive and provenance file (if needed) - only if they are not cached.
- Stores the downloaded files at `HELM_CHART_CACHE`.
- Depending on the "verification strategy", verify the chart provenance.

Chart caching is enabled by default and required (i.e. can't be disabled). Cache invalidation is out of scope.

### Implementation details

Downloaded chart archives and provenance files are stored in `HELM_CHART_CACHE` under a path that is intended to be human readable and uniquely identify a chart URL in a given repo.

Charts from the same repo have a parent dir in commmon, and so do charts from the same base URL. Consequently, cache is automatically invalidated when the repo name or the chart URL change. Preserving the cache in those cases breaks the user's expectation of using charts from the repo's configured base URL - and even represent a security vulnerability.

As an example, consider a repo named `foo` is added with a base URL `https://charts.io/`.

- `https://charts.io/chart-0.0.1.tgz` is cached at `HELM_CHART_CACHE/foo/4bb376a04a641ae55541c36d073705c6d2bb7e35/chart-0.0.1.tgz`
- `https://charts.io/chart-0.1.0.tgz` is cached at `HELM_CHART_CACHE/foo/4bb376a04a641ae55541c36d073705c6d2bb7e35/chart-0.1.0.tgz`
- `https://charts.io/another-0.1.0.tgz` is cached at `HELM_CHART_CACHE/foo/4bb376a04a641ae55541c36d073705c6d2bb7e35/another-0.1.0.tgz`

Charts from an unknown repo are stored under `HELM_CHART_CACHE/-/`.

If the user changes that repo name to `bar`, preserving the base URL `https://charts.io/`, the cache path changes:

- `https://charts.io/chart-0.0.1.tgz` is cached at `HELM_CHART_CACHE/bar/4bb376a04a641ae55541c36d073705c6d2bb7e35/chart-0.0.1.tgz`
- `https://charts.io/chart-0.1.0.tgz` is cached at `HELM_CHART_CACHE/bar/4bb376a04a641ae55541c36d073705c6d2bb7e35/chart-0.1.0.tgz`
- `https://charts.io/another-0.1.0.tgz` is cached at `HELM_CHART_CACHE/bar/4bb376a04a641ae55541c36d073705c6d2bb7e35/another-0.1.0.tgz`

If the `bar` repo later changes its base URL to `https://new.charts.io/`, the cache path changes:

- `https://charts.io/chart-0.0.1.tgz` is cached at `HELM_CHART_CACHE/bar/3262d0eae5c54b51ee1d5389d4d7567afd385ed3/chart-0.0.1.tgz`
- `https://charts.io/chart-0.1.0.tgz` is cached at `HELM_CHART_CACHE/bar/3262d0eae5c54b51ee1d5389d4d7567afd385ed3/chart-0.1.0.tgz`
- `https://charts.io/another-0.1.0.tgz` is cached at `HELM_CHART_CACHE/bar/3262d0eae5c54b51ee1d5389d4d7567afd385ed3/another-0.1.0.tgz`

### Future improvements

- Concurrently fetch charts from a lockfile.
- Implement a `--no-cache` feature to disable chart caching for an operation.

### Open questions

- Does the current proposal breaks any of the user's expectations?
   - For example, is `helm pull` semantics to always download a chart? Does the user expect it to fail if there's no connectivity but the chart is already cached?


juniorz:
cc/ @bacongobbler

juniorz:
Unsure why the naked return is now a problem, as it was already there.

object88:
I have a couple of questions about the design.

Is this concurrency safe, either from the CLI or API?  What happens if two processes or go routines fetch the same chart at the same time?

What happens if a chart in a museum is overwritten with a new version?  The [chart museum options](https://chartmuseum.com/docs/#other-cli-options) explicitly allow this behavior, and I don't see how this proposal would support that.  Is there a mechanism where helm requests the digest before assuming that its cached version is still valid?

Is there a cache purging mechanism?  While it's true that charts are _generally_ small, my particular use-case works with umbrella charts, which compose many charts into one.  These are currently around 250kb compressed, and we generate several a day (3-4).  I don't want to pretend that over the course of a year, 250MB is a lot of used space in a CI/CD system, but it also appears that it's unrecoverable space.

You state that the cache path should be human-readable, but also share example paths that have GUIDs.  What is that GUID, and how do I understand what I'm looking at if I'm manually going through the directory structure?

I _like_ caching (enough that I've implemented a similar mechanism for chart tarballs in a personal project), but along with off-by-one errors, it's one of the three most difficult CS problems.  I don't want to dissuade you from this effort (especially as someone who's not a part of the `helm` team), it's just tricky.

juniorz:
Hi @object88, thanks for bringing up these points. I'll answer them inline.

> Is this concurrency safe, either from the CLI or API? What happens if two processes or go routines fetch the same chart at the same time?

The caching mechanism does not provide fewer guarantees than `helm` currently does for concurrent writes: files are atomically stored on disk via `fileutil.AtomicWriteFile` and last writer wins.

Note that this guarantee is not even consistently provided across all file downloads in `helm` - AFAICT the repos index files are not safe for concurrent write: https://github.com/helm/helm/blob/82f739072cb3c94f26e3103e4524a1f74526342d/pkg/repo/chartrepo.go#L152

If your concern is regarding concurrent reads and writes, this does not seem to be a guarantee currently provided by `helm` either. `helm install` for example, does not seem to be safe in those cases since concurrent routines share the same temporary destination of archives: https://github.com/helm/helm/blob/82f739072cb3c94f26e3103e4524a1f74526342d/pkg/action/install.go#L669

The same seems to be true for concurrently running `helm update` and any other operation that makes use of the cached repo index. 

For this reason, I believe this concern is out of scope for an initial take on chart caching.

> What happens if a chart in a museum is overwritten with a new version?

That's is a good point. This is not supported in the current implementation because it seems to be the exception, not the rule: overwriting the content of a versioned artifact without changing its version is usually not a good practice. For simplicity, I would be more inclined to solve this by having some mechanism to disallow caching for these repos over relying on using response headers and additional roundtrips to determine freshness. 

> What is that GUID, and how do I understand what I'm looking at if I'm manually going through the directory structure?

That is a digest of the archive base URL and is used as the cache key. See the examples I provided for cases of why it is useful. Regarding your concern with storage, something similar already happens with `helm install`: chart archives are already stored in the repository cache dir and there is no cache eviction strategy AFAICT.

I acknowledge that caching is hard, and having a comprehensive complex implementation makes it even harder. For this reason, my goal was to come with a minimal implementation that might work under the 80/20 rule.

Hopefully, we can land on an initial implementation that is simple enough and can reach beta status (disabled by default) to solve 80% of the use cases.

bacongobbler (changes requested):
Hi @juniorz. With a PR as large as this one, it would seem prudent to first take a step back and write up a Helm Improvement Proposal to discuss the design criteria as well as to understand the issues/edge cases this proposal intends to address. It also gives us a clear indication on what would be considered the acceptable criteria for this change, what still needs addressing, and what the community can do to help.

Please read through [HIP 1](https://github.com/helm/community/blob/master/hips/hip-0001.md) and [HIP 2](https://github.com/helm/community/blob/master/hips/hip-0002.md). Let me know if you have any questions.

Thanks.

sgandon:
Any news on this PR ?

Ornias1993:
> Any news on this PR ?

Most likely the submitter doesn't want to spend even more time going through all sorts of discussions/design-doc-phases before a PR can be reviewed.

bacongobbler:
https://github.com/helm/community/pull/185

bacongobbler:
Closing this; there’s an open improvement proposal where the design is being reviewed closely before looking at physical implementations. I’d suggest taking a look at helm/community#185 and provide feedback there before we look at implementations. We’re trying to solve this for more than just the chart cache.

PR #8840, opened 2020-10-02
Add include-file flag for external files

### LINKED ISSUE #3276: Allow helm to add external files to chart deployment

hunsche:

Thank you!

**Parsing**

**Single files**
```

# in chart/templates/cmap.yml
```

**Dirs**

```

# in chart/templates/cmap.yml
```

**Globs**

```

# in chart/templates/cmap.yml
```

**Closing**

closes #3276
closes #8227

* currently, files are parsed first, then dirs; if a file overlaps somehow, the dir will overwrite it. What if we parse files last, so you can potentially overwrite an unwanted file in a dir? This one is easy to do
* if a path is not found, is unreadable, or if a glob yields no files, the helm command currently fails. Should we continue the install with a warning?
* all external files are added to the root of the chart. This means you can overwrite any chart file, including any template. Should we add these files to a special `_external` dir? Chart authors will need to append it to every lookup: `{{ (.Files.Glob "_external/certs/") }}`

This PR is a continuation of the PR #8227 created by @vladfr

**Still todo:**

- [x] implement --include-dir to get all files in a dir; this could be enhanced with glob
- [x] add tests
- [ ] add docs


donatelloOo:
What is the status of this PR ? Why it's not yet included in master branch ?

PR #8862, opened 2020-10-08
lint: lint all documents in a multi-doc yaml file

### LINKED ISSUE #8621: Linting fails in 3.3.0 when a template file results in only a comment.

bonifaido:
Signed-off-by: Nandor Kracser <bonifaido@gmail.com>

The current linting code handled YAML files incorrectly, so I tried my best to fix multiple issues around that:
- multiple documents, only the first document got parsed and linted
- YAML files without actual documents, just containing a `---` marker
- empty YAML files with comments

There might be other issues that I have fixed accidentally.

Related: #8621

**If applicable**:

### DIFF
--- a/pkg/lint/rules/template.go
+++ b/pkg/lint/rules/template.go
@@ -20,6 +20,7 @@ import (
 	"bufio"
 	"bytes"
 	"fmt"
+	"io"
 	"os"
 	"path"
 	"path/filepath"
@@ -27,7 +28,7 @@ import (
 	"strings"
 
 	"github.com/pkg/errors"
-	"sigs.k8s.io/yaml"
+	"k8s.io/apimachinery/pkg/util/yaml"
 
 	"helm.sh/helm/v3/pkg/chart/loader"
 	"helm.sh/helm/v3/pkg/chartutil"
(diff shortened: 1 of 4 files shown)


pkg/lint/rules/template.go
```diff
@@ -20,14 +20,15 @@ import (
...
+	"io"
 	"os"
 	"path"
 	"path/filepath"
 	"regexp"
 	"strings"
 
 	"github.com/pkg/errors"
-	"sigs.k8s.io/yaml"
+	"k8s.io/apimachinery/pkg/util/yaml"
```
bacongobbler, line 31: this might be considered a breaking change as it's unclear whether these two yaml parsers are compatible with each other. One YAML parser's output can differ from another.

Can you see if this can be reverted?

bonifaido: The main feature why I have changed to `apimachinery` is that it supports multi-document YAML files correctly (the `---` separator).

Just checked which YAML parser is used underneath, and `apimachinery` uses the same: `sigs.k8s.io/yaml`, it just wraps it to add multi-document support, so I think it is safe to say that it is the same YAML parser.

https://github.com/kubernetes/apimachinery/blob/d10b9c3a6d2ee7cd58bec9931482a0d604534231/pkg/util/yaml/decoder.go#L32

mattfarina (approved):
lgtm

pkg/lint/rules/testdata/multi-template-fail/templates/_helpers.tpl
```diff
@@ -0,0 +1,63 @@
+{{/* vim: set filetype=mustache: */}}
```
hickeyma, line 1: You shouldn't need the `_helpers.tpl` file

pkg/lint/rules/testdata/multi-template-fail/values.yaml
```diff
@@ -0,0 +1,66 @@
+# Default values for multi-template-fail.
```
hickeyma, line 1: You shouldn't need the `values.yaml` file

pkg/lint/rules/template.go
hickeyma: Looked at the code in `k8s.io/apimachinery/pkg/util/yaml` and also discussed with @mattfarina. The code wraps `sigs.k8s.io/yaml` with an IO reader to read multi-doc yaml file and then calls the yaml unmarshall for each doc contained. It is using the same parser as previous and good to use.

hickeyma (changes requested):
Thanks for fixing this @bonifaido. Its nearly there just some small cleanup of test files to address.

hickeyma:
@bonifaido You will need to rebase as well to resolve conflict.

bonifaido:
I hopefully managed to fix those, thanks!

hickeyma (approved):
LGTM, thanks for the fix @bonifaido.

PR #8881, opened 2020-10-13
Prune the last release history in case no deployed succeed before

tomahawk28:

In case of a history of release has no `failed` release, they could grow over the number of `--max-history` configuration. 
`giantswarm` found out in some cases, helm secrets are counting almost ~1000 secrets for this reason.

This PR is to delete the last `failed` release history in case of no successful deployed yet, so we can respect `--max-history` for the edge cases.

**If applicable**:

### DIFF
--- a/pkg/storage/storage.go
+++ b/pkg/storage/storage.go
@@ -174,7 +174,9 @@ func (s *Storage) removeLeastRecent(name string, max int) error {
 	relutil.SortByRevision(h)
 
 	lastDeployed, err := s.Deployed(name)
-	if err != nil {
+	if errors.Is(err, driver.ErrNoDeployedReleases) {
+		// no-op
+	} else if err != nil {
 		return err
 	}
 
--- a/pkg/storage/storage_test.go
+++ b/pkg/storage/storage_test.go
@@ -384,6 +384,55 @@ func TestStorageDontDeleteDeployed(t *testing.T) {
 	}
 }
 
+func TestStorageDeleteLastFailed(t *testing.T) {
+	storage := Init(driver.NewMemory())
+	storage.Log = t.Logf
+	storage.MaxHistory = 3
+
+	const name = "angry-bird"
+
+	// Set up storage with test releases.
+	setup := func() {
+		// release records
+		rls0 := ReleaseTestData{Name: name, Version: 1, Status: rspb.StatusFailed}.ToRelease()
+		rls1 := ReleaseTestData{Name: name, Version: 2, Status: rspb.StatusFailed}.ToRelease()
+		rls2 := ReleaseTestData{Name: name, Version: 3, Status: rspb.StatusFailed}.ToRelease()
+
+		// create the release records in the storage
+		assertErrNil(t.Fatal, storage.Create(rls0), "Storing release 'angry-bird' (v1)")
+		assertErrNil(t.Fatal, storage.Create(rls1), "Storing release 'angry-bird' (v2)")
+		assertErrNil(t.Fatal, storage.Create(rls2), "Storing release 'angry-bird' (v3)")
+	}
+
+	setup()
+
+	rls3 := ReleaseTestData{Name: name, Version: 4, Status: rspb.StatusFailed}.ToRelease()
+	assertErrNil(t.Fatal, storage.Create(rls3), "Storing release 'angry-bird' (v4)")
+
+	// On inserting the 4th record, we expect the oldest failed release get deleted
+	hist, err := storage.History(name)
+	if err != nil {
+		t.Fatal(err)
+	} else if len(hist) != storage.MaxHistory {
+		for _, item := range hist {
+			t.Logf("%s %v", item.Name, item.Version)
+		}
+		t.Fatalf("expected %d items in history, got %d", storage.MaxHistory, len(hist))
+	}
+
+	expectedVersions := map[int]bool{
+		2: true,
+		3: true,
+		4: true,
+	}
+
+	for _, item := range hist {
+		if !expectedVersions[item.Version] {
+			t.Errorf("Release version %d, found when not expected", item.Version)
+		}
+	}
+}
+
 func TestStorageLast(t *testing.T) {
 	storage := Init(driver.NewMemory())
 


bacongobbler:
Hi @tomahawk28. Can you please provide a practical example we can use to reproduce the original issue? That way we can better understand how this scenario can occur.

For example, I cannot see how this might solve the usual scenario where you have a release ledger with 1 DEPLOYED release and then 1000 FAILED releases thereafter. This particular fix only appears to solve cases where you never reached an initial DEPLOYED state in the first place, or that state is somehow missing from the ledger, neither of which seem like practical scenarios.

Thanks!

tomahawk28:
@bacongobbler  Thanks for your response. 
It's hard to give a concrete example you can run, but I can explain our situation. 
We are automating helm install/upgrade by [our component](https://github.com/giantswarm/chart-operator) and it's failed to run every helm install/upgrade since we now have an `admission-webhook` running in the background. 

```
helm -n giantswarm history aws-operator-9.1.2-dev
REVISION	UPDATED                 	STATUS	CHART                                                      	APP VERSION     DESCRIPTION
22      	Mon Oct 12 19:47:38 2020	failed	aws-operator-9.1.1-604634a7c4b900b4746b41197f2a8afd665387e6	9.1.2-dev Upgrade "aws-operator-9.1.2-dev" failed: failed to create resource: admission webhook "validation.gatekeeper.sh" denied the request: [denied by unique-operator-version] label app.kubernetes.io/version=9.1.2-dev is not unique in resources of kind "Deployment"
23      	Mon Oct 12 19:52:36 2020	failed	aws-operator-9.1.1-604634a7c4b900b4746b41197f2a8afd665387e6	9.1.2-dev Upgrade "aws-operator-9.1.2-dev" failed: failed to create resource: admission webhook "validation.gatekeeper.sh" denied the request: [denied by unique-operator-version] label app.kubernetes.io/version=9.1.2-dev is not unique in resources of kind "Deployment"
24      	Mon Oct 12 19:57:41 2020	failed	aws-operator-9.1.1-604634a7c4b900b4746b41197f2a8afd665387e6	9.1.2-dev Upgrade "aws-operator-9.1.2-dev" failed: failed to create resource: admission webhook "validation.gatekeeper.sh" denied the request: [denied by unique-operator-version] label app.kubernetes.io/version=9.1.2-dev is not unique in resources of kind "Deployment"
25      	Mon Oct 12 20:02:52 2020	failed	aws-operator-9.1.1-604634a7c4b900b4746b41197f2a8afd665387e6	9.1.2-dev Upgrade "aws-operator-9.1.2-dev" failed: failed to create resource: admission webhook "validation.gatekeeper.sh" denied the request: [denied by unique-operator-version] label app.kubernetes.io/version=9.1.2-dev-is not unique in resources of kind "Deployment"
26      	Mon Oct 12 20:08:10 2020	failed	aws-operator-9.1.1-604634a7c4b900b4746b41197f2a8afd665387e6	9.1.2-dev Upgrade "aws-operator-9.1.2-dev" failed: failed to create resource: admission webhook "validation.gatekeeper.sh" denied the request: [denied by unique-operator-version] label app.kubernetes.io/version=9.1.2-dev is not unique in resources of kind "Deployment"
... up to 500...
```

In the above case, it makes sense that you could reach the maximum number of helm histories as soon, and given the fact that admission-controller have been used widely these days, it's not the edge case anymore you have no `deployed` helm histories in a release ledger, (it could fail from the install process).

hope that it works for you!

tomahawk28:
@bacongobbler Sorry to ping again, but these are happening a lot from our clusters these days. We would like to see this patch from upstream. Thanks.

tomahawk28:
To give you more background regarding this issue; When using Helm in an automation platform like from within CICD, with a new Helm chart or a new cluster that has settings that lead to a failed release, it can happen that the release fails already on the first try. 
When in this case automation is set to retry on failure it can lead to the case that the history/ledger gets way longer than the set value in max-history.
This could be the edge case but others who failed on helm release would benefit from this patch since the number of helm secrets would not be flooded by initial failure.

mattfarina:
The documentation for this flag says...

> limit the maximum number of revisions saved per release. Use 0 for no limit (default 10)

I've not really thought about how failed deployments play into that. If Helm is keeping around many failed deployments this could be thought of as a bug. Marketing that way and going to bring this up in meeting.

bacongobbler:
I'm still confused how one can get into this initial state. I thought we captured the case where an initial FAILED deployment will be overridden by the next `helm install`. Is that not the case?

I think it'd be best if we can get a base case going here so we can better understand the issue and the scenarios surrounding it before jumping to conclusions on a fix.

bacongobbler:
> If Helm is keeping around many failed deployments this could be thought of as a bug.

I'm partially aware of this behaviour. The `--release-max` flag does not kick in until the first DEPLOYED release. That way the flag ensures the last DEPLOYED release is not purged from the release history.

For example, if you have 10 releases in the following order:

DEPLOYED
FAILED
FAILED
FAILED
FAILED
FAILED
FAILED
FAILED
FAILED
FAILED

If you deploy for the 11th time and it fails, `--release-max` will retain that initial DEPLOYED state until another DEPLOYED state is reached.

On the other side, this also works like the following:

FAILED
FAILED
FAILED
FAILED
DEPLOYED
FAILED
FAILED
FAILED
FAILED
FAILED

If you deploy another FAILED release, the release ledger should bump the last release off the list since it isn't the last DEPLOYED state.

But that goes back to the original question: how do you get into a state with no DEPLOYED releases and 1000 FAILED ones?

We can certainly try to fix this, but I'd like to better understand how one enters this state and see if there's a better way to address this issue.

tomahawk28:
> But that goes back to the original question: how do you get into a state with no DEPLOYED releases and 1000 FAILED ones?

For example, if I try to apply a deployment object with a different `spec.selector.matchLabels` and `spec.template.metadata.labels` with `helm install` command and this would result in a failed release. 

Later that, If I try to `helm upgrade` a failed release by a reasonable amount of retries, all helm secrets goes over the number of maximum helm records. 

```fish
# try to upgrade a failed release 20 times
for i in (seq 1 20); helm upgrade test1 -n giantswarm ./; end;
Error: UPGRADE FAILED: failed to create resource: Deployment.apps "test1" is invalid: spec.template.metadata.labels:...

# check the number of helm secrets belongs to `test1`
k -n giantswarm get secret -lowner=helm,name=test1
NAME                           TYPE                 DATA   AGE
sh.helm.release.v1.test1.v1    helm.sh/release.v1   1      18m
sh.helm.release.v1.test1.v10   helm.sh/release.v1   1      13m
sh.helm.release.v1.test1.v11   helm.sh/release.v1   1      13m
sh.helm.release.v1.test1.v12   helm.sh/release.v1   1      13m
sh.helm.release.v1.test1.v13   helm.sh/release.v1   1      13m
sh.helm.release.v1.test1.v14   helm.sh/release.v1   1      13m
sh.helm.release.v1.test1.v15   helm.sh/release.v1   1      13m
sh.helm.release.v1.test1.v16   helm.sh/release.v1   1      13m
sh.helm.release.v1.test1.v17   helm.sh/release.v1   1      13m
sh.helm.release.v1.test1.v18   helm.sh/release.v1   1      13m
sh.helm.release.v1.test1.v19   helm.sh/release.v1   1      13m
sh.helm.release.v1.test1.v2    helm.sh/release.v1   1      15m
sh.helm.release.v1.test1.v20   helm.sh/release.v1   1      13m
sh.helm.release.v1.test1.v21   helm.sh/release.v1   1      13m
sh.helm.release.v1.test1.v22   helm.sh/release.v1   1      13m
sh.helm.release.v1.test1.v3    helm.sh/release.v1   1      14m
sh.helm.release.v1.test1.v4    helm.sh/release.v1   1      14m
sh.helm.release.v1.test1.v5    helm.sh/release.v1   1      14m
sh.helm.release.v1.test1.v6    helm.sh/release.v1   1      14m
sh.helm.release.v1.test1.v7    helm.sh/release.v1   1      13m
sh.helm.release.v1.test1.v8    helm.sh/release.v1   1      13m
sh.helm.release.v1.test1.v9    helm.sh/release.v1   1      13m
```

Someone can manually check whether the helm upgrade process is going well and stop retries but that means that person also needs to take care of helm secrets that left by the failed helm upgrades.

hickeyma:
This is technically a bug. Thinking about this however, I would have to agree with @bacongobbler. Elaborating on a few points:

- You want the `--release-max` flag to be based off the first/last `deployed` release so you contain at least one successful release in your history. Otherwise you have no successful release to rollback.
- If you have a `deployed` release then the release history maintains the last successful release and purges the failed releases which maintains rollback functionality.
- I wonder rather than "how one enters the state of all failed releases", should the question be why? Why if you failed the first install, would you continue upgrading unsuccessfully over and over again to the point of having 10s/100s of release revisions?

hickeyma:
Lets push this out till 3.5.0 to get more feedback as 3.4.1 is being cut today.

tomahawk28:
"how one enters the state of all failed releases" is a plausible scenario if we use helm install/update operation with the automated method, likely operators. I admit I exaggerate the situation by saying there are 1000 failed releases only, but this is the bug we actually need to fix since not everyone using helm knows about helm didn't count the number of release ledger if releases are full of the failed one. 

1. What if `--max-history` is 5, and history is full of 20 failed releases? I think it makes sense for helm intervene to delete the last 15 failed records from the helm ledger, even if that number is small.

tomahawk28:
@hickeyma @bacongobbler Let me know if it's still not convincing; Using helm in one of our [component](https://github.com/giantswarm/chart-operator) caught this problem and we notice it only after Prometheus alert reported the sudden increase of the number of secrets in k8s clusters. I think users who might have the same problem didn't catch this yet.

PR #8884, opened 2020-10-13
Moving Tiller to new location

### LINKED ISSUE #8346: Make Tiller Image Available on Docker Hub
When [Helm v2 hits it's end of life](https://github.com/helm/helm/issues/8343) the container repository hosted in gcr is likely going to be garbage collected. So that people can continue to get Tiller as an image since some will not migrate yet, I propose we:

1. Make the Tiller images available in Docker Hub in addition to GCR
2. Make Helm use the image in Docker Hub starting v2.17.0

This way people can get their images and things continue to keep working.

On Docker Hub we can use the helmpack arg (we already have that, I think).

mattfarina:
The tiller primary location is now GitHub Container Registry.
It is also replicatated to GCR, Docker Hub, and Quay.

GCR is going away. We want to preserve tiller beyond the end of life for Helm v2. This moves Tiller to GHCR along with making it available on Quay and Docker Hub.

The environment variables are already setup in CI.

**If applicable**:

### DIFF
--- a/.circleci/deploy.sh
+++ b/.circleci/deploy.sh
@@ -29,8 +29,9 @@ VERSION=
 if [[ -n "${CIRCLE_TAG:-}" ]]; then
   VERSION="${CIRCLE_TAG}"
 else
-  echo "Skipping deploy step; this is not a tag"
-  exit
+  # Canary version is used with helm init --canary-image flag.
+  # Does not push canary binary which is Helm v3.
+  VERSION="canary"
 fi
 
 echo "Install docker client"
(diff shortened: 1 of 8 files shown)


technosophos (approved):
LGTM

.circleci/deploy.sh
```diff
@@ -44,35 +45,62 @@ export CLOUDSDK_CORE_DISABLE_PROMPTS=1
...
-apt install -y apt-transport-https
-echo "deb [arch=amd64] https://packages.microsoft.com/repos/azure-cli/ stretch main" | tee /etc/apt/sources.list.d/azure-cli.list
-curl -L https://packages.microsoft.com/keys/microsoft.asc | apt-key add
-apt update
-apt install -y azure-cli
+echo "Configuring Docker Hub configuration"
+echo ${DOCKER_PASS} | docker login -u ${DOCKER_USER} --password-stdin
+
+echo "Configuring Quay configuration"
+echo ${QUAY_PASS} | docker login ghcr.io -u ${QUAY_USER} --password-stdin
```
bacongobbler, line 61: Here we're calling it "quay configuration", but we're logging into ghcr.io. This may be a mismatch, as on line 49 we've already logged into ghcr.io with another account.

bacongobbler (changes requested):
Saw one tiny gotcha in the CI script. Everything else looks good otherwise. Thanks for working on this!

.circleci/deploy.sh
```diff
@@ -44,35 +45,62 @@ export CLOUDSDK_CORE_DISABLE_PROMPTS=1
...
-echo "Pushing binaries to Azure"
-az storage blob upload-batch -s _dist/ -d "$AZURE_STORAGE_CONTAINER_NAME" --pattern 'helm-*' --connection-string "$AZURE_STORAGE_CONNECTION_STRING"
+  echo "Pushing binaries to Azure"
+  az storage blob upload-batch -s _dist/ -d "$AZURE_STORAGE_CONTAINER_NAME" --pattern 'helm-*' --connection-string "$AZURE_STORAGE_CONNECTION_STRING"
 
-echo "Pushing KEYS file to Azure"
-az storage blob upload -f "KEYS" -n "KEYS" -c "$AZURE_STORAGE_CONTAINER_NAME" --connection-string "$AZURE_STORAGE_CONNECTION_STRING"
+  echo "Pushing KEYS file to Azure"
+  az storage blob upload -f "KEYS" -n "KEYS" -c "$AZURE_STORAGE_CONTAINER_NAME" --connection-string "$AZURE_STORAGE_CONNECTION_STRING"
+endif
```
scottrigby, line 106: shellify. also adding newline at EOF

```suggestion
fi

```

.circleci/deploy.sh
```diff
@@ -44,35 +45,62 @@ export CLOUDSDK_CORE_DISABLE_PROMPTS=1
...
+docker tag "ghcr.io/helm/tiller:${VERSION}" "helmpack/tiller:${VERSION}"
+docker push "helmpack/tiller:${VERSION}"
+
+echo "Pushing image to Quay"
+docker tag "ghcr.io/helm/tiller:${VERSION}" "quay.io/helmpack/tiller:${VERSION}"
+docker push "quay.io/helmpack/tiller:${VERSION}"
+
+# Canary version is used with helm init --canary-image flag.
+# Does not push canary binary which is Helm v3.
+ifneq ($(VERSION),"canary")
```
scottrigby, line 85: looks like some make crept in. shellify-ing:
```suggestion
if [ "$VERSION" != "canary" ]; then
```

scottrigby (commented):
otherwise this lgtm 👍 

I was going to suggest perhaps a GitHub action could simplify auth now that we're omving to ghcr, but I see the GITHUB_TOKEN [doesn't have the right perms](https://docs.github.com/en/free-pro-team@latest/packages/getting-started-with-github-container-registry/migrating-to-github-container-registry-for-docker-images#authenticating-with-the-container-registry) for that yet during the beta. Plus we want to auth for the other registries anyway. So could be a follow-up if we want to do that.

.circleci/deploy.sh
mattfarina: Fixed, along with some of the other issues @scottrigby found.

.circleci/deploy.sh
mattfarina: Fixed. thanks

.circleci/deploy.sh
mattfarina: Good catch. fixed. thanks.

scottrigby (approved):

bacongobbler (approved):

PR #8901, opened 2020-10-15
Change default repositories for Helm v2

technosophos:
Per the discussion in the Helm call, this PR is part of the final set of changes to Helm 2. It does the following:

- Init now initializes stable to `charts.helm.sh/stable`
- There is a flag to _disable_ this behavior and use the old repo
- Every command checks to see if `stable` or `incubator` is pointed at the Google buckets. If so, Helm issues a warning.

**Obviously this is a breaking change.** All of the Helm core maintainers, chart maintainers, and org maintainers have been notified of this, and we agree that it is necessary because Google is going to decommission the existing charts repo.

Note that I have NOT automatically converted existing repository URLs to the new one. I will do that in a separate PR because the core maintainers did not reach a consensus as to whether or not that was appropriate.

*NOTE:* Race detection on unit tests has been disabled because there is a known issue with current versions of Go and macOS Catalina.

### DIFF
--- a/cmd/helm/helm.go
+++ b/cmd/helm/helm.go
@@ -36,6 +36,7 @@ import (
 	helm_env "k8s.io/helm/pkg/helm/environment"
 	"k8s.io/helm/pkg/helm/portforwarder"
 	"k8s.io/helm/pkg/kube"
+	"k8s.io/helm/pkg/repo"
 	"k8s.io/helm/pkg/tlsutil"
 )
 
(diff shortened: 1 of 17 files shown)


mattfarina:
@technosophos the error is the one that always catches me... gotta run `make docs`.

technosophos:
Oh yeah... Helm 2... :-|

cmd/helm/helm.go
```diff
@@ -175,13 +176,60 @@ Environment:
...
+	for _, exp := range expiredRepos {
+		r, ok := repoFile.Get(exp.name)
+		if !ok {
+			return
+		}
+
+		if url := r.URL; strings.Contains(url, exp.old) {
+			fmt.Fprintf(
+				os.Stderr,
+				"WARNING: %q is deprecated for %q and will be deleted Nov. 13, 2020.\nWARNING: You must switch to %q\n",
```
hickeyma, line 215: Just wondering if this would read better as:
You must switch to %q\n --> You should switch to %q asap.\n

hickeyma (commented):
I think this will be helpful for users, thanks @technosophos. Just small comment inline and few things as follows:

Do you mind checking in the following files as I think there is still usage of the old storage that might need to be  updated: ./docs/using_helm.md, ./cmd/helm/installer/init_test.go and ./pkg/repo/index_test.go 

Is this working as expected or should the add also check for the old storage?

```console
$ ./bin/helm repo add old https://kubernetes-charts.storage.googleapis.com
"old" has been added to your repositories

$ ./bin/helm repo list
NAME    	URL                                             
stable  	https://charts.helm.sh/stable                   
local   	http://127.0.0.1:8879/charts                    
jetstack	https://charts.jetstack.io                      
old     	https://kubernetes-charts.storage.googleapis.com

$ ./bin/helm repo remove stable
"stable" has been removed from your repositories

$ ./bin/helm repo add stable https://kubernetes-charts.storage.googleapis.com
"stable" has been added to your repositories

$ ./bin/helm repo update
WARNING: "kubernetes-charts.storage.googleapis.com" is deprecated for "stable" and will be deleted Nov. 13, 2020.
WARNING: You must switch to "https://charts.helm.sh/stable"
...Skip local chart repository
...Successfully got an update from the "jetstack" chart repository
...Successfully got an update from the "old" chart repository
Update Complete.

$ ./bin/helm install --name mysql-1 stable/mysql
WARNING: "kubernetes-charts.storage.googleapis.com" is deprecated for "stable" and will be deleted Nov. 13, 2020.
WARNING: You must switch to "https://charts.helm.sh/stable"
NAME:   mysql-1
LAST DEPLOYED: Mon Oct 19 11:30:35 2020
NAMESPACE: default
STATUS: DEPLOYED
[...]
```

technosophos:
@hickeyma In last Thursday's call, someone remarked that we _should_ be able to add the old repos under a different name, so people could work with the new and the old repos for a bit (I don't know why... but it was one of the goals I wrote down in my notes from that meeting). I supported the behavior by warning on stable/incubator, but allowing for other names like "old"

technosophos:
I can also backport the error-generating code to Helm 2, but my understanding from the last meeting was that we didn't want to do that for Helm 2. Just Helm 3.

mattfarina (approved):
lgtm

PR #8902, opened 2020-10-15
Rewrite a whole bunch of old repo URLs to the new repo URL

technosophos:

This rewrites a good chunk of the old Google URLs to the new charts.helm.sh URLs. For the most part, these rewrites are cosmetic, and impact exclusively testing data. Consequently, this is a low-risk PR.

A second PR will make substantive changes to the Helm command.

**If applicable**:


mattfarina (approved):
lgtm.

I find it to be sneaky that a `go.sum` file change showed up in there. On a change like this that does not change packages it's a surprise to see this. But, that's Go.

technosophos:
Ugh... I tried to get rid of that... but the Go modules system is just... you know. Here's to writing Helm 4 in Rust!

go.sum
```diff
@@ -339,6 +339,7 @@ github.com/golang/groupcache v0.0.0-20190129154638-5b532d6fd5ef h1:veQD95Isof8w9
 github.com/golang/groupcache v0.0.0-20190129154638-5b532d6fd5ef/go.mod h1:cIg4eruTrX1D+g88fzRXU5OdNfaM+9IcxsU14FzY7Hc=
 github.com/golang/groupcache v0.0.0-20190702054246-869f871628b6 h1:ZgQEtGgCBiWRM39fZuwSd1LwSqqSW0hOdXCYYDX0R3I=
 github.com/golang/groupcache v0.0.0-20190702054246-869f871628b6/go.mod h1:cIg4eruTrX1D+g88fzRXU5OdNfaM+9IcxsU14FzY7Hc=
+github.com/golang/groupcache v0.0.0-20191227052852-215e87163ea7 h1:5ZkaAPbicIKTF2I64qf5Fh8Aa83Q/dnOafMYV0OMwjA=
```
hickeyma, line 342: Why is there a change in the sum file?

hickeyma (changes requested):
Nice tidy up @technosophos, thanks. Tests passing as expected.

Small comment inline. Also, just wondering should these files be updated: `./pkg/action/testdata/output/list-uncompressed-deps-tgz.txt` and 
`./pkg/action/testdata/output/list-compressed-deps-tgz.txt`

go.sum
technosophos: Go auto-updated that dependency and I can't figure out why. I no longer understand go modules.

technosophos:
@hickeyma those tests seem to contain regression tests for data in tar files generated by an older version of Helm. I am wary of trying to manually reconstruct that tar ball. So I left the old references.

go.sum
technosophos: Okay, so in tracing things, it appears that this came from a rebase, not from Go itself. I'll try removing it.

technosophos: I manually removed the line. But it looks like there was a problem in the go modules file (notice a missing entry) that the go module tool chain tried to fix during a rebase. It seems that the local compile will work without this line. But my guess is that the next time somebody updates or rebuilds the module cache, this exact same thing will happen again. That said... this PR can do without it, and I am tired of trying to deal with Go modules. So removing it is fine for this PR.

mattfarina (approved):
lgtm

PR #8903, opened 2020-10-15
warn and block old repo URLs

technosophos:

This is the least heavy-handed way I could come up with to stop people from using the old charts repo.

- It warns if `stable` or `incubator` are mapped to one of the old URLs (but allows someone to, say, map them to `oldstable`)
- It refuses to add one of the old URLs using `helm repo add` unless you supply `--allow-deprecated-repos`

**This is a breaking change**, but one that the core maintainers, org maintainers, and chart maintainers have agreed is necessary. Google will turn off the old chart repository on Nov. 13, 2020. This provides warnings to users who are pointing to the old repo, and prevents users from adding the deprecated repos. It stops short of rewriting the repos on the user's behalf.

**If applicable**:

### DIFF
--- a/cmd/helm/repo_add.go
+++ b/cmd/helm/repo_add.go
@@ -37,12 +37,19 @@ import (
 	"helm.sh/helm/v3/pkg/repo"
 )
 
+// Repositories that have been permanently deleted and no longer work
+var deprecatedRepos = map[string]string{
+	"//kubernetes-charts.storage.googleapis.com":           "https://charts.helm.sh/stable",
+	"//kubernetes-charts-incubator.storage.googleapis.com": "https://charts.helm.sh/incubator",
+}
+
 type repoAddOptions struct {
-	name        string
-	url         string
-	username    string
-	password    string
-	forceUpdate bool
+	name                 string
+	url                  string
+	username             string
+	password             string
+	forceUpdate          bool
+	allowDeprecatedRepos bool
 
 	certFile              string
 	keyFile               string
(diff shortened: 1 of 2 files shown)


cmd/helm/repo_add.go
```diff
@@ -83,11 +90,21 @@ func newRepoAddCmd(out io.Writer) *cobra.Command {
...
 
 	return cmd
 }
 
 func (o *repoAddOptions) run(out io.Writer) error {
+	// Block deprecated repos
+	if !o.allowDeprecatedRepos {
+		for _, url := range deprecatedRepos {
+			if strings.Contains(o.url, url) {
+				return fmt.Errorf("Repo %q is no longer available", o.url)
```
mattfarina, line 103: Should we tell them about the new location here?

cmd/helm/root.go
```diff
@@ -208,5 +210,54 @@ func newRootCmd(actionConfig *action.Configuration, out io.Writer, args []string
...
+			name: "incubator",
+			old:  "kubernetes-charts-incubator.storage.googleapis.com",
+			new:  "https://charts.helm.sh/incubator",
+		},
+	}
+
+	// parse repo file.
+	// Ignore the error because it is okay for a repo file to be unparseable at this
+	// stage. Later checks will trap the error and respond accordingly.
+	repoFile, err := repo.LoadFile(repofile)
```
mattfarina, line 241: Do we really want to hit the disk and read this file on every command every time? I'm just thinking out loud. It is a small file but on many commands it will be read more than once.

cmd/helm/repo_add.go
technosophos: So if you run it, you'll see that the line immediately above it says what the new repo is. That message comes out of the root function as a warning. This one is just the error that stops the repo add.

cmd/helm/root.go
technosophos: I'm open to alternatives...

cmd/helm/root.go
```diff
@@ -208,5 +210,54 @@ func newRootCmd(actionConfig *action.Configuration, out io.Writer, args []string
...
+	for _, exp := range expiredRepos {
+		r := repoFile.Get(exp.name)
+		if r == nil {
+			return
+		}
+
+		if url := r.URL; strings.Contains(url, exp.old) {
+			fmt.Fprintf(
+				os.Stderr,
+				"WARNING: %q is deprecated for %q and will be deleted Nov. 13, 2020.\nWARNING: You must switch to %q\n",
```
hickeyma, line 255: It might read better as:

You must switch to %q\n --> You should switch to %q asap.\n

cmd/helm/root.go
hickeyma: Valid observation from @mattfarina. Can't think of an alternative at moment.

hickeyma (changes requested):
Thanks @technosophos for adding this. Some comments inline.

cmd/helm/repo_add.go
technosophos: Oh -- you're right... that warning comes up AFTER you add the repo

cmd/helm/root.go
technosophos: Currently, this is blocking the PR merge, but nobody has any alternatives. Not sure what you want me to do.

mattfarina (approved):
lgtm

technosophos:
Based on a conversation with Matt Farina, we'll merge this one with the pre-check on the repo file. If anyone comes up with a better idea, we can replace.

PR #8906, opened 2020-10-16
Disbling pushing Tiller to quay in CI

mattfarina:
CircleCI has been unable to automatically push images to Quay. An
error of "denied: requested access to the resource is denied" is
returned when pushing to Quay.

When SSHing into the CI runner and manually running the commands
which uses the CircleCI stored environment variables the commands
succeed. The Quay credentials work.

Disabling Quay as it is blocking CI but the problem is not
immediately obvious.


mattfarina:
Closes #8907

technosophos (approved):
LGTM

PR #8915, opened 2020-10-19
Bugfix: panic when chart contains requirements.lock

zheng1:

**If applicable**:

### DIFF
--- a/pkg/chart/loader/load.go
+++ b/pkg/chart/loader/load.go
@@ -123,6 +123,9 @@ func LoadFiles(files []*BufferedFile) (*chart.Chart, error) {
 			if err := yaml.Unmarshal(f.Data, &c.Lock); err != nil {
 				return c, errors.Wrap(err, "cannot load requirements.lock")
 			}
+			if c.Metadata == nil {
+				c.Metadata = new(chart.Metadata)
+			}
 			if c.Metadata.APIVersion == chart.APIVersionV1 {
 				c.Files = append(c.Files, &chart.File{Name: f.Name, Data: f.Data})
 			}
--- a/pkg/chart/loader/load_test.go
+++ b/pkg/chart/loader/load_test.go
@@ -206,6 +206,32 @@ func TestLoadFile(t *testing.T) {
 	verifyDependencies(t, c)
 }
 
+func TestLoadFiles_BadCases(t *testing.T) {
+	for _, tt := range []struct {
+		name          string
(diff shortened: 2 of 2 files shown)


bacongobbler:
Hi @zheng1. Can you please write a unit test that covers this case. Thanks.

zheng1:
> Hi @zheng1. Can you please write a unit test that covers this case. Thanks.

Hi, @bacongobbler, I had added a test for this case, please take a look again.

bacongobbler (approved):

bacongobbler:
needs a second maintainer to review before merging

mattfarina (approved):
lgtm

mattfarina:
For some reason, the CircleCI hook failed with an HTTP response status code of 400 (Bad Request). This is why we don't have the CI info for it. I think it's for this entire PR rather than any particular commits on it.

mattfarina (approved):

zheng1:
Hi, @mattfarina, what should I do next to get this PR merged?

bridgetkromhout:
> For some reason, the CircleCI hook failed with an HTTP response status code of 400 (Bad Request). This is why we don't have the CI info for it. I think it's for this entire PR rather than any particular commits on it.

Hi, @mattfarina - Every once in a while I see that kind of timeout on various projects, and when I push a new change to that PR, it re-runs successfully. I suggest that @zheng1 may want to try that, as this may be an intermittent failure with CircleCI?

zheng1:
It still doesn't look like circleCI has been re-run, I'll try to close and reopen the issue.

zheng1:
Done, I think I've found a way to re-run circleCI.

gy-lehel:
There is a logical error in this patch, as it assumes API version to be V1. The crash is caused by the fact, that during the unpacking of the tgz the requirements.lock and yaml files are unpacked before the chart metadata is taken from Chart.yaml. Since the original code expresses a difference in the processing of for API versions higher than V1 this commit will introduce an inconsistency in the accepted file list, based purely on the order of file processing!

mattfarina:
@gy-lehel Thanks for pointing out this issue. If there is no `Chart.yaml` file in the chart a panic would still occur if this is not merged. I have looked at your PR and put it into the 3.4.1 milestone, too. The ordering issue is an issue, too.

mattfarina:
@zheng1 Thanks for fixing the tests.

PR #8929, opened 2020-10-21
fix(test): display error message

bacongobbler:
This fixes the error output to display the error's default value (the error message) rather than
Go's internal representation of its value.

supersedes #8416.


hickeyma (approved):
LGTM, thanks @bacongobbler

PR #8932, opened 2020-10-22
helm search supports semver pre version numbers starting with 0

wawa0210:
fix #8924

**What this PR does / why we need it**:
If the current helm repo version number contains illegal semver rules (pre version starts with 0), helm repo search will not work properly

**Special notes for your reviewer**:

**If applicable**:
- [ ] this PR contains documentation
- [x] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

Issue #8864, opened 2020-10-08
helm package: No error printed on invalid semantic version


ammario:
This leads to confusing errors when we aren't setting versions correctly. E.g:

```
$ helm search repo --debug -l coder-nightly
No results found
```

See 

https://github.com/helm/helm/blob/fc9b46067f8f24a90b52eba31e09b31e69011e93/pkg/repo/index.go#L187-L196

hickeyma:
Thanks for raising the issue @ammario. Can you describe in more detail what you mean by "confusing errors"? What do you expect to happen?

I am asking to get a better understanding of your issue as search command works as expected for me.

mattfarina:
@ammario Do you use `helm lint` on your chart before packaging it?

bacongobbler:
> Do you use `helm lint` on your chart before packaging it?

Interesting... It appears that it is OK to package a chart with an invalid semver. That does not appear to be expected behaviour.

```
/tmp ><> helm create test-8864
Creating test-8864
/tmp ><> cd test-8864/
/tmp/test-8864 ><> vim Chart.yaml
/tmp/test-8864 ><> cat Chart.yaml
apiVersion: v2
name: test-8864
description: A Helm chart for Kubernetes
type: application
version: foo
appVersion: 1.16.0
/tmp/test-8864 ><> helm package .
Successfully packaged chart and saved it to: /tmp/test-8864/test-8864-foo.tgz
```

`helm lint` catches it though.

```
/tmp/test-8864 ><> helm lint .
==> Linting .
[ERROR] Chart.yaml: version 'foo' is not a valid SemVer
[INFO] Chart.yaml: icon is recommended

Error: 1 chart(s) linted, 1 chart(s) failed
```

I thought that was one of the scenarios we were supposed to catch on `helm package`, as the rest of Helm relies on the package being a valid semver to perform things like automatic upgrades to the latest version by default.

We cannot capture and return error output on `helm search`. It is intentional for Helm to silently ignore packages that are not valid semver. If one package in the repository index is failing a semver check, the entire chart userbase cannot pull any charts from that repository.

bacongobbler:
So.... Looking at this further, I'd propose we do not look at addressing the `helm search` case, but instead look at the ways where a user can package a Helm chart with an invalid semver. That should prevent users from packaging and uploading invalid charts in the first place.

ammario:
[...]

PR #8933, opened 2020-10-22
Fix  invalid semver version number of the helm package command will escape

wawa0210:

Create a chart package and use helm package to package it. If we enter a version number that does not conform to the semver specification, an error will be reported, as expected

command is as follows

```
helm create nginx 
helm package --version 1.17.0-0565674
Error: Version segment starts with 0
```

But if we write the non-compliant version directly in Cahrts.yaml, and then use the helm package command to package, this verification is escaped. A package with a version number that does not conform to the semver specification is generated
```
➜  myrepo cat nginx/Chart.yaml | grep "version:"
version: 0.1.0-0234257
➜  myrepo helm package nginx
Successfully packaged chart and saved it to: /Users/zhangxiao/myrepo/nginx-0.1.0-0234257.tgz
```

Prevent charts with illegal semver version numbers from being packaged

**If applicable**:

ref 
#8932
#8864

### DIFF
--- a/pkg/action/package_test.go
+++ b/pkg/action/package_test.go
@@ -22,30 +22,10 @@ import (
 	"path"
 	"testing"
 
+	"github.com/Masterminds/semver/v3"
+
 	"helm.sh/helm/v3/internal/test/ensure"
-	"helm.sh/helm/v3/pkg/chart"
 )
-
-func TestSetVersion(t *testing.T) {
-	c := &chart.Chart{
-		Metadata: &chart.Metadata{
-			Name:    "prow",
-			Version: "0.0.1",
-		},
-	}
-	expect := "1.2.3-beta.5"
-	if err := setVersion(c, expect); err != nil {
-		t.Fatal(err)
-	}
-
-	if c.Metadata.Version != expect {
-		t.Errorf("Expected %q, got %q", expect, c.Metadata.Version)
-	}
-
-	if err := setVersion(c, "monkeyface"); err == nil {
-		t.Error("Expected bogus version to return an error.")
-	}
-}
 
 func TestPassphraseFileFetcher(t *testing.T) {
 	secret := "secret"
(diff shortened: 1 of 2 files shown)


pkg/action/package_test.go
```diff
@@ -23,30 +23,8 @@ import (
 	"testing"
 
 	"helm.sh/helm/v3/internal/test/ensure"
-	"helm.sh/helm/v3/pkg/chart"
 )
 
-func TestSetVersion(t *testing.T) {
```
bacongobbler, line 29: can you please add this test back. test cases should not be removed. If it needs to be modified as part of a bug fix, then let's fix up the test case, but we should not be removing test coverage unless absolutely necessary.

bacongobbler (changes requested):

pkg/action/package_test.go
wawa0210: In the latest revision, the `setVersion` method has been moved out and a new verification method has been added, so I removed this exported `TestSetVersion`, is it necessary to add a unit test for `validateVersion`

My personal opinion is not necessary

bacongobbler: Yes, a test case is necessary to ensure regressions are not introduced in the future. Especially something that can be easily tested like validateVersion.

I’d write test cases for the most common use cases to ensure the behaviour works as expected.

bacongobbler (changes requested):

pkg/action/package_test.go
wawa0210: It makes sense, but I have a strange feeling that the unit test here feels like testing the semver library, like the responsibility of the library itself

mattfarina: @wawa0210 In this case we are testing the Helm experience. What if something breaks in the SemVer library? We want to catch that before it bubbles up to Helm users. On a widely used and stable project like this it's important to catch issues early.

pkg/action/package_test.go
```diff
@@ -100,3 +80,46 @@ func TestPassphraseFileFetcher_WithInvalidStdin(t *testing.T) {
...
+				ver: "1.1.3.sd.023658",
+			},
+			semver.ErrInvalidSemVer,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if err := validateVersion(tt.args.ver); err != nil {
+				if err != tt.wantErr {
+					t.Errorf("validateVersion() error = %v, wantErr %v", err, tt.wantErr)
```
mattfarina, line 119: The format of the error string is not what we typically use. Can you please update it to a format like...

```
t.Errorf("Expected %s got %s", tt.wantErr, err)
```

mattfarina (changes requested):

pkg/action/package_test.go
wawa0210: done~

mattfarina (approved):
lgtm

hickeyma (approved):
LGTM, thanks for fixing this @wawa0210

i540608:
@hickeyma @mattfarina  Hi, sorry for bothering; Is the change still available in Helm 3.5.2? 

I just noticed that one of my CI/CD pipeline failed due the exact same error.
I am wondering if there has been changes and if it is permanent? 
I can easily update my versioning scheme, but i would like to give head up to other colleagues who use same versioning scheme;


```
helm version
# version.BuildInfo{Version:"v3.5.2", GitCommit:"167aac70832d3a384f65f9745335e9fb40169dc2", GitTreeState:"dirty", GoVersion:"go1.15.7"}

helm package helm/example1 --version 1.0-117.0269326
Error: Version segment starts with 0
```

PR #8942, opened 2020-10-27
Fix helm-lint print only one error

### LINKED ISSUE #8934: Helm Lint prints one error and then ends
This may or may not be a feature request, but it would be really nice if running the command `helm lint` would print more than one error at a time. The behavior I'm seeing currently is that the lint ends upon first error detected. Is that by design, and/or am I missing something?

Output of `helm version`: version.BuildInfo{Version:"v3.3.4", GitCommit:"a61ce5633af99708171414353ed49547cf05013d", GitTreeState:"dirty", GoVersion:"go1.15.2"}

Output of `kubectl version`: Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.3", GitCommit:"1e11e4a2108024935ecfcb2912226cedeafd99df", GitTreeState:"clean", BuildDate:"2020-10-14T18:49:28Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"darwin/amd64"}

Cloud Provider/Platform (AKS, GKE, Minikube etc.): Mac OS

donggangcj:

Close #8934

**If applicable**:

### DIFF
--- a/pkg/lint/rules/template.go
+++ b/pkg/lint/rules/template.go
@@ -55,11 +55,7 @@ func Templates(linter *support.Linter, values map[string]interface{}, namespace 
 	// Load chart and parse templates
 	chart, err := loader.Load(linter.ChartDir)
 
-	chartLoaded := linter.RunLinterRule(support.ErrorSev, fpath, err)
-
-	if !chartLoaded {
-		return
-	}
+	linter.RunLinterRule(support.ErrorSev, fpath, err)
 
 	options := chartutil.ReleaseOptions{
 		Name:      "test-release",


pkg/lint/rules/template.go
```diff
@@ -55,11 +55,7 @@ func Templates(linter *support.Linter, values map[string]interface{}, namespace
 	// Load chart and parse templates
 	chart, err := loader.Load(linter.ChartDir)
 
-	chartLoaded := linter.RunLinterRule(support.ErrorSev, fpath, err)
-
-	if !chartLoaded {
-		return
-	}
```
mattfarina, line 62: Unfortunately, this is not going to work. If the chart did not load a message is added to the linters messages. Since the chart could not be loaded the other linting tests will also fail. There is no chart to operate on. So, it needs to return.

Looking through the times this function returns, it only happens when something that blocks future steps from happening is encountered. This return should not be removed.

mattfarina (changes requested):

bacongobbler:
hi @donggangcj. Please also make sure to provide test cases to ensure you're fixing the issue you set out to fix. If you don't have a test case, please ask the original author of the ticket for one. Thanks.

pkg/lint/rules/template.go
donggangcj: Yes, I also found this problem. if there is an error in `Chart.yaml`, this check function should return.

> Looking through the times this function returns, it only happens when something that blocks future steps from happening is encountered. This return should not be removed.

The return time of the function is fine. Maybe this is in line with the design.

donggangcj:
> hi @donggangcj. Please also make sure to provide test cases to ensure you're fixing the issue you set out to fix. If you don't have a test case, please ask the original author of the ticket for one. Thanks.

According to the discussion in the related issue, I think there is no obvious bug. Thanks!

bacongobbler:
I'm going to close this as something we do not intend to merge. Without a test case, we cannot be certain this PR addresses the issue the OP raised in the ticket. Changing code for the sake of changing code is risky for a project that is so widely used to se tend to stick to a more conservative approach when reviewing PRs.

Thanks.

Issue #7696, opened 2020-02-27
Helm v3 inspect nil pointer stacktrace


zkutasi:
Output of `helm version`:
`version.BuildInfo{Version:"v3.1.0", GitCommit:"b29d20baf09943e134c2fa5e1e1cab3bf93315fa", GitTreeState:"clean", GoVersion:"go1.13.7"}`

Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-15T19:18:23Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-16T08:00:38Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}
```

Cloud Provider/Platform (AKS, GKE, Minikube etc.): 
Company-internal distribution based on SLES15

Issue:
I work on a private project, so I will not disclose any info about it, but...
I have an umbrella helm-chart in the packaged format (tgz file), where I get the following stacktrace:

```
helm3 inspect CHARTNAME
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x98 pc=0x1296a9c]

goroutine 1 [running]:
helm.sh/helm/v3/pkg/chart/loader.LoadFiles(0xc00018a000, 0x9b, 0x100, 0x9b, 0x100, 0x0)
	/home/circleci/helm.sh/helm/pkg/chart/loader/load.go:126 +0xa9c
helm.sh/helm/v3/pkg/chart/loader.LoadArchive(0x1be8a60, 0xc00000f650, 0xc00000f650, 0x0, 0x0)
	/home/circleci/helm.sh/helm/pkg/chart/loader/archive.go:193 +0x87
helm.sh/helm/v3/pkg/chart/loader.LoadFile(0xc000380140, 0x40, 0x0, 0x0, 0x0)
	/home/circleci/helm.sh/helm/pkg/chart/loader/archive.go:65 +0x221
helm.sh/helm/v3/pkg/chart/loader.FileLoader.Load(...)
[...]

PR #8948, opened 2020-10-27
[#7696] Avoid crash in chart loader on unexpected file sequence

gy-lehel:
Make sure, that chart metadata is initialized by the time the processing
of the chart is started.

Helm inspect crashes in 3.2, 3.3, and 3.4.0 if the files in the chart do not arrive in a presumed order. If for example requirements.yaml will precede Chart.yaml in the file list, the helm process will sigsegv due to uninitialized metadata.

Tested with charts generated with 2.16.11, 2.17.0, 3.2.4, and 3.4.0 in a project where charts generated with these versions provoked this crash in 3.2.4, 3.3.x, and 3.4.0.

refs #7696

Signed-off-by: Lehel Gyuro <lehel@freemail.hu>

**What this PR does / why we need it**: helm inspect crashes, if file sequence of the chart does not include Chart.yaml as first.

**If applicable**:

### DIFF
--- a/pkg/chart/loader/load.go
+++ b/pkg/chart/loader/load.go
@@ -73,10 +73,11 @@ func LoadFiles(files []*BufferedFile) (*chart.Chart, error) {
 	c := new(chart.Chart)
 	subcharts := make(map[string][]*BufferedFile)
 
+	// do not rely on assumed ordering of files in the chart and crash
+	// if Chart.yaml was not coming early enough to initialize metadata
 	for _, f := range files {
 		c.Raw = append(c.Raw, &chart.File{Name: f.Name, Data: f.Data})
-		switch {
-		case f.Name == "Chart.yaml":
+		if f.Name == "Chart.yaml" {
 			if c.Metadata == nil {
 				c.Metadata = new(chart.Metadata)
 			}
@@ -89,6 +90,14 @@ func LoadFiles(files []*BufferedFile) (*chart.Chart, error) {
 			if c.Metadata.APIVersion == "" {
 				c.Metadata.APIVersion = chart.APIVersionV1
 			}
+		}
+	}
+	for _, f := range files {
+		c.Raw = append(c.Raw, &chart.File{Name: f.Name, Data: f.Data})
+		switch {
+		case f.Name == "Chart.yaml":
+			// already processed
+			continue
 		case f.Name == "Chart.lock":
 			c.Lock = new(chart.Lock)
 			if err := yaml.Unmarshal(f.Data, &c.Lock); err != nil {
--- a/pkg/chart/loader/load.go
+++ b/pkg/chart/loader/load.go
@@ -93,7 +93,6 @@ func LoadFiles(files []*BufferedFile) (*chart.Chart, error) {
 		}
 	}
 	for _, f := range files {
-		c.Raw = append(c.Raw, &chart.File{Name: f.Name, Data: f.Data})
 		switch {
 		case f.Name == "Chart.yaml":
 			// already processed


Ferenc-:
@mattfarina could you please take a look at this?

mattfarina (approved):
lgtm

Note to other reviewers... when looking at other files there are checks like `c.Metadata.APIVersion == chart.APIVersionV1`. If the `Chart.yaml` file was not loaded first these checks could guess the wrong API version.

hickeyma:
I tested the issue as follows:

- Created a scaffold chart : `helm2 create issue-7696-v2`
- Added a dependency: `issue-7696-v2/requirements.yaml`
- Updated the dependency: `helm2 dep update issue-7696-v2`
- Remove the `Chart.yaml`:  `mv issue-7696-v2/Chart.yaml` .
- Archive the chart: `tar -cvzf issue-7696-v2-0.1.0.tgz issue-7696-v2/`
- Return `Chart.yaml`: 
   - `rm -rf issue-7696-v2`
   - `mkdir issue-7696-v2`
   - `cp Chart.yaml issue-7696-v2/`
- Uncompressed the zip: `gunzip issue-7696-v2-0.1.0.tgz`
- Appended the `Chart.yaml` to the archive: `tar -rf issue-7696-v2-0.1.0.tar issue-7696-v2/Chart.yaml`
- Compress the archive:
   - `gzip issue-7696-v2-0.1.0.tar`
   - `mv issue-7696-v2-0.1.0.tar.gz issue-7696-v2-0.1.0.tgz`
- Inspect the archive: `helm3 inspect all issue-7696-v2-0.1.0.tgz`

The dependency is as follows:

```console
 cat issue-7696-v2/requirements.yaml
dependencies:
- name: mysql
  version: "^1.6.2"
  repository: "https://charts.helm.sh/stable/"
```

When I run `inspect` before using the PR (i.e. current code from tip of main), it works as expected:

```console
$ helm3 inspect all issue-7696-v2-0.1.0.tgz
load.go:112: Warning: Dependencies are handled in Chart.yaml since apiVersion "v2". We recommend migrating dependencies to Chart.yaml.
apiVersion: v1
appVersion: "1.0"
dependencies:
- name: mysql
  repository: https://charts.helm.sh/stable/
  version: ^1.6.2
name: issue-7696-v2
version: 0.1.0

---
# Default values for issue-7696-v2.
[...]
nodeSelector: {}

tolerations: []

affinity: {}

---
```

I would have expected this to fail as per #7696.  @mattfarina @gy-lehel I would appreciate if you could let me know how best to reproduce the issue.

hickeyma:
If I looks at the archive it shows file order as:

```console
$ tar -tvzf issue-7696-v2-0.1.0.tgz
drwxr-xr-x usr/usr       0 2020-11-04 17:01 issue-7696-v2/
drwxr-xr-x usr/usr       0 2020-11-04 15:30 issue-7696-v2/templates/
-rw-r--r-- usr/usr    1036 2020-11-04 15:30 issue-7696-v2/templates/ingress.yaml
drwxr-xr-x usr/usr       0 2020-11-04 15:30 issue-7696-v2/templates/tests/
-rw-r--r-- usr/usr     400 2020-11-04 15:30 issue-7696-v2/templates/tests/test-connection.yaml
-rw-r--r-- usr/usr    1775 2020-11-04 15:30 issue-7696-v2/templates/_helpers.tpl
-rw-r--r-- usr/usr    1763 2020-11-04 15:30 issue-7696-v2/templates/deployment.yaml
-rw-r--r-- usr/usr     426 2020-11-04 15:30 issue-7696-v2/templates/service.yaml
-rw-r--r-- usr/usr     219 2020-11-04 15:30 issue-7696-v2/templates/serviceaccount.yaml
-rw-r--r-- usr/usr    1568 2020-11-04 15:30 issue-7696-v2/templates/NOTES.txt
-rw-rw-r-- usr/usr      95 2020-11-04 16:59 issue-7696-v2/requirements.yaml
-rw-r--r-- usr/usr    1514 2020-11-04 15:30 issue-7696-v2/values.yaml
-rw-r--r-- usr/usr     214 2020-11-04 17:00 issue-7696-v2/requirements.lock
drwxr-xr-x usr/usr       0 2020-11-04 17:00 issue-7696-v2/charts/
-rw-r--r-- usr/usr   11424 2020-11-04 17:00 issue-7696-v2/charts/mysql-1.6.7.tgz
-rw-r--r-- usr/usr     342 2020-11-04 15:30 issue-7696-v2/.helmignore
-rw-r--r-- usr/usr     109 2020-11-04 17:04 issue-7696-v2/Chart.yaml
```

bacongobbler:
code LGTM, though would it be possible to write a test case to capture this case in the future?

mattfarina:
@hickeyma The panic from the issue was fixed when #8915 was merged. There is a second issue, through. It is an order issue.

Look at your output and you see...

```
load.go:112: Warning: Dependencies are handled in Chart.yaml since apiVersion "v2". We recommend migrating dependencies to Chart.yaml.
```

You should only see that if the `apiVersion` is `v2`. In your case it's `v1`. You should not see that message. You can [see the code here](https://github.com/helm/helm/blob/5436aeb0e5ead09db298ff782ff287045cecf410/pkg/chart/loader/load.go#L111-L113).

The reason this message is coming up is that when the `requirements.yaml` file is loaded there is a check for the `apiVersion` value. Since the `Chart.yaml` file has not been loaded the metadata is initialized to nil. This message is showing because Helm doesn't know a value it's looking for because it hasn't loaded it yet.

hickeyma:
>  The panic from the issue was fixed when #8915 was merged. There is a second issue, through. It is an order issue.

Ok, makes sense then. Thanks @mattfarina/

LGTM also but agree with @bacongobbler that a unit test would be helpful.

technosophos (approved):
This is a provisional LGTM assuming that #9009 (the test cases) is also merged.

PR #8950, opened 2020-10-28
completion: move to native zshCompletion

### LINKED ISSUE #8893: Convert zsh completion to GenZshCompletion
Cobra released their new version yesterday https://github.com/spf13/cobra/releases/tag/v1.1.1 which included `ValidArgsFunction` for zsh completion. This means all the code needed for zsh here https://github.com/helm/helm/blob/master/cmd/helm/completion.go#L147 can be removed and converted to `GenZshCompletion`

Output of `helm version`: `version.BuildInfo{Version:"v3.3.4", GitCommit:"a61ce5633af99708171414353ed49547cf05013d", GitTreeState:"dirty", GoVersion:"go1.15.2"}`

Output of `kubectl version`: `Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.2", GitCommit:"f5743093fd1c663cb0cbc89748f730662345d44d", GitTreeState:"clean", BuildDate:"2020-09-16T21:51:49Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"darwin/amd64"}`

knrt10:
Cobra https://github.com/spf13/cobra/releases/tag/v1.1.1 fixed the issue
which did not need zshCompletion to be changed to bash.

Also https://github.com/helm/helm/pull/8950#issuecomment-721109633

closes: #8893
Signed-off-by: knrt10 <tripathi.kautilya@gmail.com>

Remove unnecessary code for zsh completion.

### DIFF
--- a/cmd/helm/completion.go
+++ b/cmd/helm/completion.go
@@ -103,6 +103,7 @@ func newCompletionCmd(out io.Writer) *cobra.Command {
 			return runCompletionZsh(out, cmd)
 		},
 	}
+	zsh.Flags().BoolVar(&disableCompDescriptions, noDescFlagName, false, noDescFlagText)
 
 	fish := &cobra.Command{
 		Use:                   "fish",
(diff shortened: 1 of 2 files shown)


cmd/helm/completion.go
```diff
@@ -145,148 +145,25 @@ fi
...
+	// both helm2 and helm3), we hook the new binary name to the completion function
+	if binary := filepath.Base(os.Args[0]); binary != "helm" {
+		renamedBinaryHook := `
+# Hook the command used to generate the completion script
+# to the helm completion function to handle the case where
+# the user renamed the helm binary
+if [[ $(type -t compopt) = "builtin" ]]; then
+    complete -o default -F __start_helm %[1]s
+else
+    complete -o default -o nospace -F __start_helm %[1]s
```
marckhouzam, line 160: This whole if statement can be removed since it's bash syntax.

It should be replaced by the zsh version which is simply: `compdef _helm helm`

Cobra does not add that line at all because it favors installing the script on `fpath` instead of sourcing it.  To be backwards compatible though, helm should support sourcing the script. Therefore you'll need to add that line at the very end of the script generated by Cobra all the time.

You would also need to add it in case the binary name is not `helm`.  We may also need to change the first line of the generated script to the different binary name, to support putting the script on `fpath`; we'd have to test it to be sure.

marckhouzam (changes requested):
Thanks for this!
A couple of tweaks to make and we'll be ready.

knrt10:
@marckhouzam did I made the changes correctly?

cmd/helm/completion.go
knrt10: I tested for sourcing for both cases using helm binary and different name binary it works. Also `fpath` for helm binary works. But I am not sure how to check for changed name for `fpath`

cmd/helm/completion.go
```diff
@@ -145,148 +145,27 @@ fi
...
-	-e "s/${LWORD}compgen${RWORD}/__helm_compgen/g" \
-	-e "s/${LWORD}compopt${RWORD}/__helm_compopt/g" \
-	-e "s/${LWORD}declare${RWORD}/builtin declare/g" \
-	-e "s/\\\$(type${RWORD}/\$(__helm_type/g" \
-	-e 's/aliashash\["\(.\{1,\}\)"\]/aliashash[\1]/g' \
-	-e 's/FUNCNAME/funcstack/g' \
-	<<'BASH_COMPLETION_EOF'
-`
-	out.Write([]byte(zshInitialization))
+	err := cmd.Root().GenZshCompletion(out)
```
marckhouzam, line 148: Could you also support disabling completion descriptions like we do for fish shell?
This means 
- adding the `--no-descriptions` flag for zsh as we do higher up for fish
- then here, calling `cmd.Root().GenZshCompletion(out)` if descriptions are enabled or `cmd.Root().GenZshCompletionNoDesc(out)` if descriptions are disabled by the user.

cmd/helm/completion.go
```diff
@@ -145,148 +145,27 @@ fi
...
+	}
 
 	zshTail := `
-BASH_COMPLETION_EOF
-}
-__helm_bash_source <(__helm_convert_bash_to_zsh)
+compdef _helm helm
 `
+
 	out.Write([]byte(zshTail))
```
marckhouzam, line 166: To be consistent with the call above, and to simplify, instead of using the `zshTail` pattern, how about just using `fprintf` here?
And maybe add a comment to explain this slightly bizarre thing we do.  Something like:
```
	// Allow to source the zsh completion file, although Cobra doesn't
	fmt.Fprintf(out, "compdef _helm helm")
```

marckhouzam (changes requested):
> I tested for sourcing for both cases using helm binary and different name binary it works. Also fpath for helm binary works. But I am not sure how to check for changed name for fpath

I also can't figure out how to make it work for a renamed binary using $fpath.
Let's forget that case, it was not supported before your PR anyway.

So, a couple more things to add and were done.

marckhouzam:
@knrt10 ping

knrt10:
Sorry was little busy, will update it today

marckhouzam (approved):
Brilliant!  Thanks @knrt10.

LGTM
We will need another maintainer to approve to get this merged.

marckhouzam:
An important point to mention is that this PR goes beyond removing hacky scripting code.  By moving to native zsh completions, helm will now automatically provide completion descriptions (like we have for Fish shell completions).  Descriptions will be enabled by default, but the user can disable them by using the `--no-descriptions` flag on the `completion zsh` command.  Note that completion behaviour need-not be "backwards-compatible" as it is not used in scripts.

Here is an example:
```
$ source <(bin/helm completion zsh)
$ h3 s<TAB>
search  -- search for a keyword in charts
show    -- show information of a chart
status  -- display the status of the named release

$ source <(bin/helm completion zsh --no-descriptions)
$ h3 s<TAB>
search  show    status
```

mattfarina (approved):
lgtm

PR #8954, opened 2020-10-29
Updates to helm create generated _helpers.tpl and Chart.yaml for readability

consideRatio:
It improves readability of `helm create` generated files.

- I added a link to the reference documentation of Chart.yaml inside Chart.yaml
- I removed trunc 63 on the defined template .name, because it will only be used in the helm.sh/chart label and be combined in the defined template .fullname which will get truncated to 63 where it actually matters.
- I made `default "some-default-value" "some-potentially-falsy-valueg"` be `"some-potentially-falsy-valueg" | default "some-default-value"` like in examples where default is used in the documentation. I find it to be easier to read.
- I updates notes about the defined templates in _helpers.tpl, which I have found to be lacking in their explanation on how they are used and at the same time quite hard to parse logically, which make them more mysterical than they should be. I also think it is good to use the terminology of "defined templates" to help developers learn to speak about them more clearly.
- 

**If applicable**:

### DIFF
--- a/pkg/chartutil/create.go
+++ b/pkg/chartutil/create.go
@@ -77,7 +77,9 @@ const maxChartNameLength = 250
 
 const sep = string(filepath.Separator)
 
-const defaultChartfile = `apiVersion: v2
+const defaultChartfile = `# Chart.yaml describes the Helm chart, for more information see:
+# https://helm.sh/docs/topics/charts/#the-chartyaml-file
+apiVersion: v2
 name: %s
 description: A Helm chart for Kubernetes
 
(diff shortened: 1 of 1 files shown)


bacongobbler:
Hi @consideRatio. Have you reached out to the chart maintainers to determine whether this is considered the "correct" convention they would prefer to use going forward? I'm afraid that without the buy-in of the chart maintainers I don't see a way we can confidently merge this into Helm. There are a number of opinionated changes here that need to be vetted by another group of community members. For example the removal of `trunc 63` could introduce a regression in behaviour when developing new charts.

consideRatio:
> Have you reached out to the chart maintainers to determine whether this is considered the "correct" convention they would prefer to use going forward? I'm afraid that without the buy-in of the chart maintainers I don't see a way we can confidently merge this into Helm. There are a number of opinionated changes here that need to be vetted by another group of community members.

Ah I understand, no I haven't reached out, but would like to do so regarding convention preferences.

> For example the removal of trunc 63 could introduce a regression in behaviour when developing new charts.

I wanted to learn a bit more, and did.
- Label's values max length is 63, so, you are double right that we shouldn't remove this trunc!
- metadata.name length is limited to 253 characters generally, except for k8s Service resource names due to the DNS name they will be provided makes it need to be truncated to 63.

### Close PR
@bacongobbler let's close this PR, but perhaps you could provide input if you think it could be a good idea to create a formatting convention poll to reach out with?

If you think so, I can open an issue to suggest an action plan with regards to that. Here are some ideas of conventions to poll right away that I have.

### Convention poll questions
Here are some ideas of what to poll.

#### Piping into default?

```diff
-{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
+{{- .Values.nameOverride | default .Chart.Name | trunc 63 | trimSuffix "-" }}
```

#### Piping into formatters?

```diff
   {{- with .Values.ingress.annotations }}
   annotations:
-    {{- toYaml . | nindent 4 }}
+    {{- . | toYaml | nindent 4 }}
   {{- end }}
```

#### Explicit root scope or dedicated variables?

```diff
 {{- if .Values.ingress.enabled -}}
-{{- $fullName := include "<CHARTNAME>.fullname" . -}}
-{{- $svcPort := .Values.service.port -}}
 # ...
 metadata:
-  name: {{ $fullName }}
+  name: {{ include "<CHARTNAME>.fullname" . }}
   # ...
   rules:
     {{- range .Values.ingress.hosts }}
     - host: {{ .host | quote }}
       http:
         paths:
           {{- range .paths }}
           - path: {{ . }}
             backend:
-              serviceName: {{ $fullName }}
-              servicePort: {{ $svcPort }}
+              serviceName: {{ include "<CHARTNAME>.fullname" $. }}
+              servicePort: {{ $.Values.service.port }}
           {{- end }}
     {{- end }}
   {{- end }}
```

bacongobbler:
I do not have enough knowledge in this area of the codebase to formulate an opinion. Sorry.

I'd reach out to the chart maintainers (#charts on Slack) and ask them what they think.

PR #8961, opened 2020-10-30
feat(helm): Allow generating markdown docs headers

marckhouzam:

Generating the command docs requires adding the title of each page by hand.
At least that is what I understood from this documentation: https://github.com/helm/helm-www#updating-the-helm-cli-reference-docs

Cobra provides a way to add the header automatically, which is what this PR does.

The flag `--generate-headers` is added to the `docs` command to keep this feature backwards-compatible.

### DIFF
--- a/cmd/helm/docs.go
+++ b/cmd/helm/docs.go
@@ -16,8 +16,11 @@ limitations under the License.
 package main
 
 import (
+	"fmt"
 	"io"
+	"path"
 	"path/filepath"
+	"strings"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
@@ -38,9 +41,10 @@ It can also generate bash autocompletions.
 `
 
 type docsOptions struct {
-	dest          string
-	docTypeString string
-	topCmd        *cobra.Command
+	dest            string
+	docTypeString   string
+	topCmd          *cobra.Command
+	generateHeaders bool
 }
 
 func newDocsCmd(out io.Writer) *cobra.Command {
@@ -62,6 +66,7 @@ func newDocsCmd(out io.Writer) *cobra.Command {
 	f := cmd.Flags()
 	f.StringVar(&o.dest, "dir", "./", "directory to which documentation is written")
 	f.StringVar(&o.docTypeString, "type", "markdown", "the type of documentation to generate (markdown, man, bash)")
+	f.BoolVar(&o.generateHeaders, "generate-headers", false, "generate standard headers for markdown files")
 
 	return cmd
 }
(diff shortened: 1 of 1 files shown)


bridgetkromhout:
I tested this by building helm locally from your fork's https://github.com/VilledeMontreal/helm/tree/feat/docHeader branch:

```
$ /Users/bridget/github/helm/helm/bin/helm version
version.BuildInfo{Version:"v3.4+unreleased", GitCommit:"200a8e30b38976b1732208e45eb82877f5cf854e", GitTreeState:"clean", GoVersion:"go1.14.3"}
$ helm version
version.BuildInfo{Version:"v3.4.0", GitCommit:"7090a89efc8a18f3d8178bf47d2462450349a004", GitTreeState:"dirty", GoVersion:"go1.15.3"}
```

I then ran through the steps at https://github.com/helm/helm-www#updating-the-helm-cli-reference-docs and can verify that this largely preserves the headers as-is. (Minor variance occurs but I do not think it's significant, like `title: "Helm Completion - bash"` changing to `title: "Helm Completion Bash"` and the header "Synopsis" being removed from several files.)

I recommend this for merging ASAP as it will make docs updates much easier.

> This change is not backwards compatible, however I have assumed that the hidden docs command need not remain backwards-compatible. Is that correct?

While this command is explained on the www repo, it does not show up in the `helm --help` output, so I doubt we are making backwards-compat guarantees. In short, I think this can go out asap. Adding to the next milestone for consideration.

bridgetkromhout (approved):

marckhouzam:
OMG instead of quoting @bridgetkromhout reply, I edited it and replaced it with mine!  I'll try to fix it back.
Here is my actually reply:

Thanks for testing @bridgetkromhout 

> Minor variance occurs but I do not think it's significant, like `title: "Helm Completion - bash"` changing to `title: "Helm Completion Bash"` 

I believe the extra `-` was added by hand by the first person that generated those completion sub-commands as they were first introduced in helm 3.3.  So I think removing the `-` is ok to allow for automation.

> and the header "Synopsis" being removed from several files.)

Good eye!  I missed that change.  After investigating, it turns out it is an improvement in Cobra, unrelated to this PR.  Basically, it removes the Synopsis for commands that don't have a long description to avoid duplicating the same text twice.  Please see https://github.com/spf13/cobra/pull/1207

marckhouzam:
@bridgetkromhout I've fixed your comment https://github.com/helm/helm/pull/8961#issuecomment-720716814 which I erased by mistake, and put back what I had received by email.  If you had done manual modifications after that, I lost them.  Sorry!

mattfarina:
As this is a feature, I think it needs to go into v3.5

marckhouzam:
After discussing backwards-compatibility at the dev call today, it was decided that even hidden commands should remain backwards-compatible; for the case of `helm docs`, it is possible others are using the command to generate the docs themselves.

I therefore have to update this PR to make the change backwards-compatible.
I wasn't sure how to make sure this PR was blocked, so I converted it to a Draft until it is ready.

marckhouzam:
I've added the `--generate-headers` flag to the `docs` command to remain backwards-compatible.
The PR is ready for review.

bacongobbler (approved):
This is wonderful. Thank you so much @marckhouzam for implementing this!

PR #8963, opened 2020-10-30
Updating descriptions for HELM_KUBEASGROUPS & HELM_KUBEASUSER

bridgetkromhout:
It appears that these descriptions were transposed from the order that seems to have been intended when they were introduced in https://github.com/helm/helm/pull/8710; I'm correcting this here (and fixing a small typo).


marckhouzam (approved):
LGTM

marckhouzam:
Since this is XS, it does not need a second approval.
Thank you @bridgetkromhout

bridgetkromhout:
Added this to the 3.4.1 milestone for consideration. As it appears to be a typo fix that won't change functionality but may confuse end users, perhaps this is the sort of thing that can and should go out in that release.

stealthybox:
🤦 
Thanks for catching this @bridgetkromhout

PR #8969, opened 2020-11-02
Clarifies action needed to list new stable repo

bridgetkromhout:
I'm seeing that https://github.com/helm/helm/pull/8903 adds warnings around the stable repo URL change (thanks, @technosophos) but I'm wondering if we want to go farther and explicitly tell people what they need to do. This may help avert some user confusion.

### DIFF
--- a/cmd/helm/root.go
+++ b/cmd/helm/root.go
@@ -252,8 +252,10 @@ func checkForExpiredRepos(repofile string) {
 		if url := r.URL; strings.Contains(url, exp.old) {
 			fmt.Fprintf(
 				os.Stderr,
-				"WARNING: %q is deprecated for %q and will be deleted Nov. 13, 2020.\nWARNING: You should switch to %q\nWARNING: Switch via: helm repo add stable https://charts.helm.sh/stable --force-update",
+				"WARNING: %q is deprecated for %q and will be deleted Nov. 13, 2020.\nWARNING: You should switch to %q via:\nWARNING: helm repo add %q %q --force-update\n",
 				exp.old,
+				exp.name,
+				exp.new,
 				exp.name,
 				exp.new,
 			)
--- a/cmd/helm/root.go
+++ b/cmd/helm/root.go
@@ -252,7 +252,7 @@ func checkForExpiredRepos(repofile string) {
 		if url := r.URL; strings.Contains(url, exp.old) {
 			fmt.Fprintf(
 				os.Stderr,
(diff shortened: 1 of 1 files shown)


bridgetkromhout:
I would like us to consider this for release as soon as is practical, so I am adding it to the nearest milestone (which is soon enough that this might help users with short-term confusion). If it is not suitable for 3.4.1, please move it to an appropriate later milestone. Thanks!

cmd/helm/root.go
```diff
@@ -252,7 +252,7 @@ func checkForExpiredRepos(repofile string) {
 		if url := r.URL; strings.Contains(url, exp.old) {
 			fmt.Fprintf(
 				os.Stderr,
-				"WARNING: %q is deprecated for %q and will be deleted Nov. 13, 2020.\nWARNING: You should switch to %q\n",
+				"WARNING: %q is deprecated for %q and will be deleted Nov. 13, 2020.\nWARNING: You should switch to %q\nWARNING: Switch via: helm repo add stable https://charts.helm.sh/stable --force-update",
```
hickeyma, line 255: So this message would potentially also be used for `incubator`. It might be best then to update it as follows:

helm repo add stable https://charts.helm.sh/stable --force-update --> helm repo add %q %q --force-update

Then add the addition params to the `Fprintf` function of:
exp.name, exp.new

hickeyma (changes requested):
Thanks for adding more info for the user @bridgetkromhout. Update required inline.

bridgetkromhout:
I've adjusted it so now this message shows for either case:

$ bin/helm repo list
WARNING: "kubernetes-charts.storage.googleapis.com" is deprecated for "stable" and will be deleted Nov. 13, 2020.
WARNING: You should switch to "https://charts.helm.sh/stable" via:
WARNING: helm repo add "stable" "https://charts.helm.sh/stable" --force-update
WARNING: "kubernetes-charts-incubator.storage.googleapis.com" is deprecated for "incubator" and will be deleted Nov. 13, 2020.
WARNING: You should switch to "https://charts.helm.sh/incubator" via:
WARNING: helm repo add "incubator" "https://charts.helm.sh/incubator" --force-update
NAME        	URL                                                       
stable      	https://kubernetes-charts.storage.googleapis.com                                    
incubator   	https://kubernetes-charts-incubator.storage.googleapis.com

bacongobbler (approved):

hickeyma (approved):
LGTM, thanks @bridgetkromhout

PR #8992, opened 2020-11-06
Remove spurious debug output

srueg:
It removes spurious debug output which pollutes stdout.

**If applicable**:

### DIFF
--- a/cmd/helm/template.go
+++ b/cmd/helm/template.go
@@ -197,7 +197,7 @@ func writeToFile(outputDir string, name string, data string, append bool) error 
 		return err
 	}
 
-	fmt.Printf("wrote %s\n", outfileName)
+	debug("wrote %s\n", outfileName)
 	return nil
 }
 
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -186,7 +186,7 @@ func (c *Configuration) renderResources(ch *chart.Chart, values chartutil.Values
 			if outputDir == "" {
 				fmt.Fprintf(b, "---\n# Source: %s\n%s\n", crd.Name, string(crd.File.Data[:]))
 			} else {
-				err = writeToFile(outputDir, crd.Filename, string(crd.File.Data[:]), fileWritten[crd.Name])
+				err = writeToFile(outputDir, crd.Filename, string(crd.File.Data[:]), fileWritten[crd.Name], c.Log)
 				if err != nil {
 					return hs, b, "", err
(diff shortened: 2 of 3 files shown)


bacongobbler:
On a side note, I think it may be prudent to consolidate these two functions somewhere to keep the code more DRY. That way we don't have to update the code in two locations. The function signature, name, and behaviour is identical.

bacongobbler (commented):
Perhaps it'd be better to demote this informational message to `debug`: https://github.com/helm/helm/blob/eefd939c6cd302c963c89ea0fed695aac62f4cdf/cmd/helm/helm.go#L49

That way users can still retrieve this information with `helm install --debug`.

What do you think?

srueg:
> On a side note, I think it may be prudent to consolidate these two functions somewhere to keep the code more DRY. That way we don't have to update the code in two locations. The function signature, name, and behaviour is identical.

That's what I thought too but then saw the comment of the function:

> // The following functions (writeToFile, createOrOpenFile, and ensureDirectoryForFile)
// are coppied from the actions package. This is part of a change to correct a
// bug introduced by #8156. As part of the todo to refactor renderResources
// this duplicate code should be removed. It is added here so that the API
// surface area is as minimally impacted as possible in fixing the issue.


> Perhaps it'd be better to demote this informational message to debug

Good point, I'll take a look.

bacongobbler (approved):
code LGTM. Needs another maintainer's approval before we can merge.

marckhouzam:
This looks like a duplicate of #8673 which was not merged due to backwards-compatibility concerns and the ease of working around it.  But @bacongobbler  would know better than I if this should go ahead or not.

bacongobbler:
In this particular case I think it can be argued that this was intended to be debug output and therefore a bug. Users can still fetch this information with `helm install --debug`. I do not believe this could be a breaking change for some users, but I have been wrong before.

The more conservative approach may be to hold off on merging this until Helm 4.

srueg:
Is the decision final to wait for Helm 4?
I'd also argue that it's a bug and the previous behavior can be re-enabled with `--debug`.

joejulian:
As much as I'd love to see this change, reading the discussion on the referenced PR does seem like this is pretty clearly considered a breaking change so the finial decision is, indeed, to wait for Helm 4.

joejulian (changes requested):
blocking as this seems to be considered a breaking change

PR #8995, opened 2020-11-07
Adds support to print rendered charts on `helm lint --debug`

### LINKED ISSUE #8982: Feature Request: `helm lint --debug` similar to `helm template --debug`
Currently when using `helm lint` this will fail if for example an invalid default value is set in a template. This then prints out an error similar to below:

```
[ERROR] templates/deployment.yaml: unable to parse YAML: error converting YAML to JSON: yaml: line 37: did not find expected key
```

However to debug this issue I then need to run `helm template --debug` to print out the invalid yaml to identify this line. 

It would be useful if we could pass the `--debug` flag to `helm lint` to render the invalid yaml, similar to the template command, rather than adding additional logic to only run `helm template --debug` if linting fails.

JasonWhall:
closes #8982 (hopefully)

This PR is to add support to render out the YAML from `helm lint` by using the `--debug` flag. It should support both valid and invalid YAML. This has similar behaviour to `helm template --debug`

**If applicable**:

### DIFF
--- a/cmd/helm/lint.go
+++ b/cmd/helm/lint.go
@@ -100,6 +100,12 @@ func newLintCmd(out io.Writer) *cobra.Command {
 					failed++
 				}
 
+				if settings.Debug {
+					for _, contents := range result.RenderedContents {
+						fmt.Fprintf(&message, "---\n# Source: %s\n%s\n", path, contents)
+					}
+				}
+
 				// Adding extra new line here to break up the
 				// results, stops this from being a big wall of
 				// text and makes it easier to follow.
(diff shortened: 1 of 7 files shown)


bacongobbler (changes requested):
Hi @JasonWhall. It appears that CI is failing on this build. Before we can review this, please take a look into fixing the failing tests. Thanks!

JasonWhall:
Hey @bacongobbler, looks like a transient issue which seems unrelated as I'm not near that area of code. I've kicked off another ci build to run and this has passed if you could take another look? Thanks!

kuisathaverat:
It would be really helpful to have this merged, sometimes the line number is not enough to know what is the problem, in my case I have an indentation issue and the error does not help much

```
Error: YAML parse error on my-chart/templates/deployment.yaml: error converting YAML to JSON: yaml: line 61: did not find expected '-' indicator
```

the actual error was here, the indent must be 12, not 8
```
          resources:
            {{- toYaml .Values.resources | nindent 8 }}
```

PR #8998, opened 2020-11-08
[PoC] Implement an azblob storage driver

paulgmiller:
Not sure we need this but playing around with if we can just dump releases into blobs

**If applicable**:


bacongobbler:
Updated the title to better track the proposal.

Issue #6581, opened 2019-10-04
add new CRDs as part of upgrade


rudoi:
the `crds` directory is applied on install but not on upgrade because of potential weirdness with upgrading CRDs. however, this means that new CRDs are also not added as part of `helm upgrade`. there should be no conflict in doing this, I believe. 

Output of `helm version`: `version.BuildInfo{Version:"v3.0.0-beta.3", GitCommit:"5cb923eecbe80d1ad76399aee234717c11931d9a", GitTreeState:"clean", GoVersion:"go1.12.9"}`

bacongobbler:
This was discussed in #6243:

> CRDs are not upgraded during helm install or helm upgrade. There is no accepted practice for managing CRD upgrades (again, because they are a global resources, and an upgrade can break things in interesting ways), so we decided not to handle this case. It is up to the operator.

See #5871 for more background on this discussion.

rudoi:
@bacongobbler yeah i saw the discussion. I totally understand the reasoning for not upgrading existing CRDs.

But if I add a new CRD to a chart, why wouldn't we want `upgrade` to install it? As it sits today:
1. I have a chart with n CRDs, `helm install` installs them
2. if i modify the CRDs and update the chart, `helm upgrade` does nothing (I get this part)
3. if I add a new CRD and update the chart, `helm upgrade` still does nothing <- this is the case I'm concerned about

`helm install` already has logic for looping through the CRDs. would it be unreasonable to use that same logic in `upgrade` and only install CRDs that don't yet exist?

bacongobbler:
fair enough. I think it's safe enough to handle the install case, just not the upgrade case. Marking as a feature enhancement

munnerz:
What concerns are there with upgrading CRDs? A proper multi version CRD that undergoes proper management (ie no backward incompatible schema changes) should allow updates just fine.

If upgrades should be left to the operator, IMO installation should be too. This causes a big disconnect and will mean that all helm end-users installing any operator will one day land on this issue, scratching their heads.

Is there a place to gather, discuss and resolve these concerns? If upgrading CRDs is truly this unworkable, we should look at paths to make it easier upstream. But for the most part a user will just “kubectl apply” new CRDs anyway, which makes me think the “updating CRD resources is hard” train of thought may be moot 🙂

munnerz:
Sorry, I see there’s quite a lot of discussion on the linked issues above already 🙈 I’ll try and catch up there and add my feedback!

[...]

PR #9007, opened 2020-11-10
fix(pkg/chart): do not install CRDs for disabled dependencies 

### LINKED ISSUE #8508: CRD's from subchart being deployed even when subchart dependencies.condition is false
Using nginx ingress as a subchart. Even when I have the deployment condition resolve to false, it still tries to deploy the CRDs.

Output of `helm version`:  
```
```
Output of `kubectl version`:  
```
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.0", GitCommit:"9e991415386e4cf155a24b1da15becaa390438d8", GitTreeState:"clean", BuildDate:"2020-03-25T14:58:59Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"windows/amd64"}
Server Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.5", GitCommit:"20c265fef0741dd71a66480e35bd69f18351daea", GitTreeState:"clean", BuildDate:"2019-10-15T19:07:57Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}
```
AKS

martonw:
closes #8508 
The original implementation installs all CRDs from dependencies without checking whether the dependency will actually be installed or not. This fix moves the CRD collection and installation process (`CRDObjects()`) a few lines down so the dependency check happens first, marking dependencies to be installed or not (`Metadata.Dependencies[].Enabled`). Furthermore, the `CRDObjects()` implementation is modified to actually check for this flag and only collect CRDs from enabled dependencies.

Unit tests are added for 2 cases: `CRDObjects()` call with a dependency that is enabled and has a CRD definition; and a variant where the dependency is disabled (and in this case we expect the dependent CRD to NOT get installed). The latter test would fail without this patch being applied. 

**If applicable**:

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -181,17 +181,6 @@ func (i *Install) Run(chrt *chart.Chart, vals map[string]interface{}) (*release.
 		return nil, err
 	}
 
-	// Pre-install anything in the crd/ directory. We do this before Helm
-	// contacts the upstream server and builds the capabilities object.
-	if crds := chrt.CRDObjects(); !i.ClientOnly && !i.SkipCRDs && len(crds) > 0 {
-		// On dry run, bail here
-		if i.DryRun {
-			i.cfg.Log("WARNING: This chart or one of its subcharts contains CRDs. Rendering may fail or contain inaccuracies.")
-		} else if err := i.installCRDs(crds); err != nil {
-			return nil, err
-		}
-	}
-
 	if i.ClientOnly {
 		// Add mock objects in here so it doesn't use Kube API server
 		// NOTE(bacongobbler): used for `helm template`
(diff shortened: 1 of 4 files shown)


bacongobbler:
Given the case where CRDs will not be installed during an update (#6581), this would introduce an issue when a dependency is initially disabled, but then eventually enabled as part of a `helm upgrade`.

How does this intend to address that particular case?

iAnomaly:
@bacongobbler I feel like the case you raise would fall under the concern of #6581 if/once it's picked up no? As you point out [there](https://github.com/helm/helm/issues/6581#issuecomment-538515612) the behavior that exists today (until #6581 is completed) is that the effect of re-enabling dependency CRDs later is materially the same as adding new CRDs to an already deployed chart later.

Am I misunderstanding? I don't think this PR can handle the case you mention without generally being the PR that resolves the feature describe in #6581.

bacongobbler:
Exactly. Which is why I'm saying that we cannot merge this PR as it will introduce cases where a release will enter an unrecoverable state.

In other words, it is better to retain the current behaviour *until* #6581 is addressed.

bacongobbler:
Closing for now. We can re-open this for discussion once we resolve these edge cases.

Issue #9028, opened 2020-11-16
action.List client is unable to return a list of all releases


dlipovetsky:
The `action.List` client is unable to return a list of all releases.

Because it [applies this filter]((https://github.com/helm/helm/blob/c5e9732a9fdda5c1bb74a1d29827cf60158d218c/pkg/action/list.go#L178)), it is only able to return the following:
(a) A list of the recent releases
(b) A list of _all_  releases marked Superseded

I believe this is due to the change introduced in https://github.com/helm/helm/pull/8282.

If you use the helm SDK, then I think this behavior is a bug. However, if you use the helm CLI `list` command, then this behavior may conform to the requirements of the CLI.

I have written a unit test (#9029) that asserts that `action.List` returns a list of all releases. If there's consensus on a fix, it can be added to that PR.

dlipovetsky:
Responding to @bacongobbler  from #9029:

> Thank you for demonstrating the issue. However, it is difficult to determine whether the proposed change to the StateMask would introduce a breaking change to the SDK's intended output or to the output of `helm list`. Right now this seems more of a theoretical issue, and I'd like to see a more practical example on what this intends to fix.

I can confirm that this is not a theoretical issue. I consume the Helm SDK. I expect the following code, which sets `StateMask = action.ListAll` before it calls `listClient.Run()`, to return _all_ releases matching the name. It does not, because of this bug.

```
func listAllReleasesForName(cfg *action.Configuration, name string) ([]*helm3release.Release, error) {
	listClient := action.NewList(cfg)
	listClient.Limit = 0
	listClient.Filter = name
	listClient.StateMask = action.ListAll

	return listClient.Run()
}
```

> I would also suggest writing a unit test under `cmd/helm/list_test.go` to verify this does not unintentionally break backwards compatibility for the CLI.

I can do that. I'll open a separate PR. Once that merges, I can rebase, fix the bug in the  `action` package, and if the CLI unit test fails, I'll make an additional fix to the CLI package.

dlipovetsky:
I thought that action.List could be (and should be) used to return a list of all release _versions_. This is incorrect; the right API to use is action.History. I filed this issue by mistake, so I'm closing it. I've added some comments (#9139) to clarify the difference in the two APIs.

PR #9029, opened 2020-11-16
Add test to verify that lister returns all releases matched by StateMask

dlipovetsky:
Failing unit test to demonstrate #9028 

The test is intended to fail.  Once there's a consensus that there is a bug, this PR can be amended with a fix commit, or the unit test commit can be cherry-picked to another PR.
  
**If applicable**:

### DIFF
--- a/pkg/action/list_test.go
+++ b/pkg/action/list_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package action
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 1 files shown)


bacongobbler (changes requested):
Thank you for demonstrating the issue. However, it is difficult to determine whether the proposed change to the StateMask would introduce a breaking change to the SDK's intended output or to the output of `helm list`. Right now this seems more of a theoretical issue, and I'd like to see a more practical example on what this intends to fix.

I would also suggest writing a unit test under `cmd/helm/list_test.go` to verify this does not unintentionally break backwards compatibility for the CLI.

Please feel free to work on a fix to better demonstrate the issue. Thanks!

dlipovetsky:
I opened #9135, which includes a fix, and this unit test.

PR #9037, opened 2020-11-19
[actions/stale] increase number of operations per run to 100

bacongobbler:
The default number of operations the stale issue bot will run is 30. This is a good size for a small-to-medium sized project and it prevents rate limits. However, for a project as large as Helm, the number of operations need to be increased so that it can find and close stale issues.


marckhouzam:
Good idea.
LGTM

PR #9038, opened 2020-11-20
fix(helm): Don't discard pre-releases upon index merge

joestringer:
Previously, when attempting to merge a new version into an existing
index.yaml which includes overlapping release and pre-release versions
(such as 0.1.0 and 0.1-dev), the merge would silently discard the
pre-release version from the newly merged index.

Fix this by performing an exact semantic comparison of the versions from
the provided index file.

### DIFF
--- a/pkg/repo/index.go
+++ b/pkg/repo/index.go
@@ -135,8 +135,13 @@ func (i IndexFile) Add(md *chart.Metadata, filename, baseURL, digest string) {
 
 // Has returns true if the index has an entry for a chart with the given name and exact version.
 func (i IndexFile) Has(name, version string) bool {
-	_, err := i.Get(name, version)
-	return err == nil
+	vs := i.Entries[name]
+	for _, v := range vs {
+		if v.Version == version {
+			return true
+		}
+	}
+	return false
 }
 
 // SortEntries sorts the entries by version in descending order.
--- a/pkg/repo/index_test.go
+++ b/pkg/repo/index_test.go
@@ -214,6 +214,81 @@ func TestMerge(t *testing.T) {
 
 }
 
+func TestMergePrerelease(t *testing.T) {
+	ind1 := NewIndexFile()
+	ind1.Add(&chart.Metadata{
+		Name:    "dreadnought",
+		Version: "0.2.0",
+	}, "dreadnought-0.2.0.tgz", "http://example.com", "aaaa")
+
+	ind2 := NewIndexFile()
+	ind2.Add(&chart.Metadata{
+		Name:    "dreadnought",
+		Version: "0.1.0",
+	}, "dreadnought-0.1.0.tgz", "http://example.com", "aaaabbbb")
+	ind2.Add(&chart.Metadata{
+		Name:    "dreadnought",
+		Version: "0.1-dev",
+	}, "dreadnought-0.1-dev.tgz", "http://example.com", "ccccbbbb")
+
+	ind1.Merge(ind2)
+
+	if len(ind1.Entries) != 1 {
+		t.Errorf("Expected 1 entries, got %d", len(ind1.Entries))
+	}
+	vs := ind1.Entries["dreadnought"]
+	if len(vs) != 3 {
+		t.Errorf("Expected 3 versions, got %d", len(vs))
+		for i := 0; i < len(vs); i++ {
+			t.Errorf("* %+v", ind1.Entries["dreadnought"][i].Metadata.Version)
+		}
+	}
+	v := vs[0]
+	if v.Version != "0.2.0" {
+		t.Errorf("Expected %q version to be 0.2.0, got %s", v.Name, v.Version)
+	}
+
+}
+
+func TestMergeSameVersion(t *testing.T) {
+	ind1 := NewIndexFile()
+	ind1.Add(&chart.Metadata{
+		Name:    "dreadnought",
+		Version: "0.2.0",
+	}, "dreadnought-0.2.0.tgz", "http://example.com", "aaaa")
+
+	ind2 := NewIndexFile()
+	ind2.Add(&chart.Metadata{
+		Name:    "dreadnought",
+		Version: "0.2.0",
+	}, "dreadnought-0.2.0.tgz", "http://example.com", "aaaabbbb")
+	ind2.Add(&chart.Metadata{
+		Name:    "dreadnought",
+		Version: "0.1-dev",
+	}, "dreadnought-0.1-dev.tgz", "http://example.com", "ccccbbbb")
+
+	ind1.Merge(ind2)
+
+	if len(ind1.Entries) != 1 {
+		t.Errorf("Expected 1 entries, got %d", len(ind1.Entries))
+	}
+	vs := ind1.Entries["dreadnought"]
+	if len(vs) != 2 {
+		t.Errorf("Expected 2 versions, got %d", len(vs))
+		for i := 0; i < len(vs); i++ {
+			t.Errorf("* %+v", ind1.Entries["dreadnought"][i].Metadata.Version)
+		}
+	}
+	v := vs[0]
+	if v.Version != "0.2.0" {
+		t.Errorf("Expected %q version to be 0.2.0, got %s", v.Name, v.Version)
+	}
+	expectedDigest := "aaaa"
+	if v.Digest != expectedDigest {
+		t.Errorf("Expected merged %q digest to be %q, got %s", v.Name, expectedDigest, v.Digest)
+	}
+}
+
 func TestDownloadIndexFile(t *testing.T) {
 	t.Run("should  download index file", func(t *testing.T) {
 		srv, err := startLocalServerForTests(nil)


joestringer:
I'm not sure if the current behaviour was specifically intended. In some sense, having a merge of {1.2.0} and {1.1-dev, 1.1.0, ...} should in theory end with the union of all input versions so that's the main argument point for the current approach. This also has the implication that over time the index will collect potentially a large number of pre-release versions. Without some garbage collection functionality I'm not sure whether this is ideal. Definitely open to discussion on this.

I can probably work around this issue on my side, but at the same time if upstream Helm worked a bit more predictably in this regard in future, then other projects could potentially benefit as well and avoid having their own variations of workarounds.

mattfarina:
@joestringer We usually want an issue describing the problem and how you got it. Would you mind writing up an issue with all kinds of details.

I've not replicated it yet but this does look like a bug from the description.

pkg/repo/index.go
```diff
@@ -219,7 +219,7 @@ func (i IndexFile) WriteFile(dest string, mode os.FileMode) error {
 func (i *IndexFile) Merge(f *IndexFile) {
 	for _, cvs := range f.Entries {
 		for _, cv := range cvs {
-			if !i.Has(cv.Name, cv.Version) {
+			if !i.Has(cv.Name, "=="+cv.Version) {
```
mattfarina, line 222: The `==` is not working here. `Has` is calling `Get` which is returning a `improper constraint` constraint error. It happens to be passing the test but not for the right reason.

The `Has` function needs to be rewritten so that it does a comparison without using constraint checking.

pkg/repo/index_test.go
```diff
@@ -214,6 +214,42 @@ func TestMerge(t *testing.T) {
...
+	}, "dreadnought-0.1-dev.tgz", "http://example.com", "ccccbbbb")
+
+	ind1.Merge(ind2)
+
+	if len(ind1.Entries) != 1 {
+		t.Errorf("Expected 1 entries, got %d", len(ind1.Entries))
+	}
+	vs := ind1.Entries["dreadnought"]
+	if len(vs) != 3 {
+		t.Errorf("Expected 3 versions, got %d", len(vs))
```
mattfarina, line 241: Can you also check for the case where the exact same version is in both and it is overwritten via the merge. This would have caught the the error with the use of `Has` and be good for future proofing.

mattfarina (changes requested):
Thanks for finding this bug. This is happening due to constraints and ranges in `Get` that is used by `Has`. `Get` can retrieve ranges, it seems, and that means `Has` has some edge case bugs.

pkg/repo/index.go
joestringer: I had wondered about that approach, but I was uneasy about the possible impact of making such a change. In particular, I note the explicit comment above that function around ignoring pre-release versions:

https://github.com/helm/helm/blob/0c7a1a5af5be49f78a61598ea449f4b50902d32b/pkg/repo/index.go#L154-L157

Looking around further, this seems like the core function used for finding versions in a Chart. I'm not familiar enough with the existing test cases to know whether modifications there could end up leading to some particularly adverse behaviour like unintentionally installing a prerelease version when the user didn't intend so.

pkg/repo/index_test.go
joestringer: Do you mean an exact same version but with a different digest for instance? Can do.

pkg/repo/index.go
mattfarina: @joestringer Dug into this further. `Get` isn't entirely clear on the range vs exact version. The current form of the function has been around for years. Due to API compat via SemVer we can't really change that function.

`Has` notes that it looks for the exact chart version. Yet, it is not doing that. So, `Has` should look be changed to look for the exact version instead of calling `Get`.

The note on `Get` is when the version is empty and in that case this is the default behavior of the semver package. Since our case is when a version is supplied the comment does not relate.

Would you be up for making this change to `Has`?

pkg/repo/index_test.go
mattfarina: I mean where index 1 and index 2 have the same version... the output index should have only one of them and it should be the right one.

joestringer: The part that confused me is if they're the same version, how do you know which one is the "right" one?

joestringer: Looks like there's some existing logic that assumes merges will not overwrite existing contents, so I made the test follow that same approach.

pkg/repo/index.go
joestringer: Done. I also realized that the `==` was invalid semver, it should have been `=` so it was previously only working due to `semver.NewConstraint()` always generating an error. That part is fixed.

Additionally, semver would determine that `1.9-dev` is semantically equivalent to `1.9.0` so to properly satisfy the documentation for `Has`, I added an extra direct string comparison of the versions to determine whether such versions should be merged.

mattfarina:
@joestringer This is failing tests. Can you please correct the problem.

joestringer:
CircleCI doesn't seem to provide a way to see your test results without signing up for their service, so I tried running the tests locally. I found unit test failure with the search and fixed that in the commit. I also hit a bunch of errors like this which I'm guessing I can ignore as they rely on some CI credentials:

```
--- FAIL: TestDependencyUpdateCmd (0.22s)
    dependency_update_test.go:41: Listening on directory /tmp/helm-repotest-463555852
    server.go:164: error logging into registry with good credentials
```

joestringer:
Hmm, looks like circleci is still not happy. Is it possible to run the full testsuite locally? I have just been trying `go test ./...`.

bacongobbler:
`make test` will run the full test suite. In this case, `make test-style` is failing due to linting errors.

joestringer:
Looks like it passed this time, thanks for the pointers.

pkg/repo/index.go
```diff
@@ -135,8 +135,13 @@ func (i IndexFile) Add(md *chart.Metadata, filename, baseURL, digest string) {
 
 // Has returns true if the index has an entry for a chart with the given name and exact version.
 func (i IndexFile) Has(name, version string) bool {
-	_, err := i.Get(name, version)
-	return err == nil
+	vs := i.Entries[name]
```
jdolitsky, line 138: Should there be a check here to ensure `i.Entries` has key `name`?

cmd/helm/search_repo_test.go
```diff
@@ -70,7 +70,7 @@ func TestSearchRepositoriesCmd(t *testing.T) {
 		golden: "output/search-output-json.txt",
 	}, {
 		name:   "search for 'maria', expect one match with semver begin with zero development version",
-		cmd:    "search repo maria --devel",
+		cmd:    "search repo maria --devel -l",
```
jdolitsky, line 73: why is this needed?

jdolitsky (changes requested):

pkg/repo/index.go
joestringer: If it's not present, the iteration should iterate zero times and return false which is the behaviour we want:

https://play.golang.org/p/UrJS0jt108P

cmd/helm/search_repo_test.go
joestringer: Hrm. I seem to recall the output ordering to be unreliable, but I can't seem to reproduce it now. I'll drop this section.

jdolitsky (approved):

joestringer:
@mattfarina I believe I've addressed your feedback, let me know if you want any further changes.

joestringer:
ping @mattfarina

pkg/repo/index_test.go
```diff
@@ -214,6 +214,81 @@ func TestMerge(t *testing.T) {
...
+	}, "dreadnought-0.2.0.tgz", "http://example.com", "aaaa")
+
+	ind2 := NewIndexFile()
+	ind2.Add(&chart.Metadata{
+		Name:    "dreadnought",
+		Version: "0.1.0",
+	}, "dreadnought-0.1.0.tgz", "http://example.com", "aaaabbbb")
+	ind2.Add(&chart.Metadata{
+		Name:    "dreadnought",
+		Version: "0.1-dev",
```
bacongobbler, line 231: This is an interesting case here... This tag does not comply to the [Semantic Versioning 2.0.0 specification](https://semver.org/spec/v2.0.0.html). Previously these versions would be rejected to protect us from potential connection string injection attacks.

The old behaviour was to reject this version from being included to protect from those attacks. I wonder if `0.1.0-dev` would work with the old code instead.

joestringer: Good catch, it doesn't match semver v2. That's worth exploring whether my original use case is resolved based on that.

> Previously these versions would be rejected

IIRC, this was not the behaviour I observed. It was more like "Depending on whether you create an index with the `v0.1-dev` style first and then merge in other indexes like `v0.1`, or if you first create the index with `v0.1` and then try to merge `v0.1-dev`, the result is different."

joestringer: I took a brief look again and indeed, if we simply use `0.1.0-dev` style dev versions then Helm does the right thing.

If however we use `0.1-dev` style dev versions, then depending on the merge ordering, Helm will inconsistently either keep or discard the `0.1-dev` versions from the index.

There is no indication to the user that the `v0.1-dev` version is invalid, it's just silently discarded in some cases depending on how you invoke helm.

bacongobbler:
Hi there. Checking in to see what's the status here and how you want to proceed. Should we close this, or is there something that we should still address here?

joestringer:
It sounds like there's no appetite upstream for making the merging behaviour consistent for versions that do not match semver v2 (ie merging `v0.1-dev` into an index with `v0.1.0` results in the same output as merging `v0.1.0` into an index with `v0.1-dev`), so we can just close this out.

If anyone discovers this PR in future, the simplest way to resolve this would be to ensure that your project uses [Semantic Versioning 2.0.0](https://semver.org/spec/v2.0.0.html) correctly.

PR #9040, opened 2020-11-20
feat(helm): Support setting --kube-version

### LINKED ISSUE #7326: (feat): helm 3 template is missing --kube-version flag
helm 2 template has --kube-version flag. But helm 3 seems to be missing it. This flag is necessary to optionally remove features from old version of Kubernetes. 

Can this flag be added back?

thatsmydoing:
Closes https://github.com/helm/helm/issues/7326 although it's already closed due to staleness

Helm 2 supported specifying `--kube-version` for `helm template` to set the kubernetes version used. This ports it to helm 3. `--api-versions` is a similar flag that was already ported.

**If applicable**:

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -198,7 +198,7 @@ func (i *Install) Run(chrt *chart.Chart, vals map[string]interface{}) (*release.
 	if i.ClientOnly {
 		// Add mock objects in here so it doesn't use Kube API server
 		// NOTE(bacongobbler): used for `helm template`
-		i.cfg.Capabilities = chartutil.DefaultCapabilities
+		i.cfg.Capabilities = chartutil.DefaultCapabilities.Copy()
 		if i.KubeVersion != nil {
 			i.cfg.Capabilities.KubeVersion = *i.KubeVersion
 		}
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -92,8 +92,10 @@ type Install struct {
 	SubNotes                 bool
 	DisableOpenAPIValidation bool
 	IncludeCRDs              bool
+	// KubeVersion allows specifying a custom kubernetes version to use and
 	// APIVersions allows a manual set of supported API Versions to be passed
 	// (for things like templating). These are ignored if ClientOnly is false
+	KubeVersion *chartutil.KubeVersion
 	APIVersions chartutil.VersionSet
 	// Used by helm template to render charts with .Release.IsUpgrade. Ignored if Dry-Run is false
 	IsUpgrade bool
@@ -197,6 +199,9 @@ func (i *Install) Run(chrt *chart.Chart, vals map[string]interface{}) (*release.
 		// Add mock objects in here so it doesn't use Kube API server
 		// NOTE(bacongobbler): used for `helm template`
 		i.cfg.Capabilities = chartutil.DefaultCapabilities
+		if i.KubeVersion != nil {
+			i.cfg.Capabilities.KubeVersion = *i.KubeVersion
+		}
 		i.cfg.Capabilities.APIVersions = append(i.cfg.Capabilities.APIVersions, i.APIVersions...)
 		i.cfg.KubeClient = &kubefake.PrintingKubeClient{Out: ioutil.Discard}
 
--- a/pkg/chartutil/capabilities.go
+++ b/pkg/chartutil/capabilities.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"strconv"
 
+	"github.com/Masterminds/semver/v3"
 	"k8s.io/client-go/kubernetes/scheme"
 
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
@@ -72,6 +73,19 @@ func (kv *KubeVersion) String() string { return kv.Version }
 //
 // Deprecated: use KubeVersion.Version.
 func (kv *KubeVersion) GitVersion() string { return kv.Version }
+
+// ParseKubeVersion parses kubernetes version from string
+func ParseKubeVersion(version string) (*KubeVersion, error) {
+	sv, err := semver.NewVersion(version)
+	if err != nil {
+		return nil, err
+	}
+	return &KubeVersion{
+		Version: sv.Original(),
+		Major:   strconv.FormatUint(sv.Major(), 10),
+		Minor:   strconv.FormatUint(sv.Minor(), 10),
+	}, nil
+}
 
 // VersionSet is a set of Kubernetes API versions.
 type VersionSet []string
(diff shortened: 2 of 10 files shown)


pkg/action/install.go
```diff
@@ -195,7 +197,10 @@ func (i *Install) Run(chrt *chart.Chart, vals map[string]interface{}) (*release.
 	if i.ClientOnly {
 		// Add mock objects in here so it doesn't use Kube API server
 		// NOTE(bacongobbler): used for `helm template`
-		i.cfg.Capabilities = chartutil.DefaultCapabilities
+		i.cfg.Capabilities = chartutil.DefaultCapabilities.Copy()
```
thatsmydoing, line 200: In actual use, this doesn't really matter since the command exits immediately, but tests seem to reuse the same global instance so any changes here were affecting downstream tests.

pkg/chartutil/capabilities.go
```diff
@@ -65,6 +76,18 @@ func (kv *KubeVersion) String() string { return kv.Version }
 // Deprecated: use KubeVersion.Version.
 func (kv *KubeVersion) GitVersion() string { return kv.Version }
 
+func ParseKubeVersion(version string) (*KubeVersion, error) {
+	sv, err := semver.NewVersion(version)
+	if err != nil {
+		return nil, err
+	}
+	return &KubeVersion{
+		Version: "v" + sv.String(),
```
thatsmydoing, line 85: semver does not have a leading `v` but I'm just matching the default KubeVersion which does have the leading `v`.

pkg/chartutil/capabilities.go
```diff
@@ -65,6 +76,18 @@ func (kv *KubeVersion) String() string { return kv.Version }
 // Deprecated: use KubeVersion.Version.
 func (kv *KubeVersion) GitVersion() string { return kv.Version }
 
+func ParseKubeVersion(version string) (*KubeVersion, error) {
+	sv, err := semver.NewVersion(version)
```
bacongobbler, line 80: This will break against versions provided by some cloud providers like GKE. They use a non-compliant semver in their kubeversion range.

Because of this, we cannot assume that the kubeversion returned by Kubernetes is semver compliant. This change will break backwards compatibility for some users.

bacongobbler (changes requested):

pkg/action/install.go
bacongobbler: This could potentially affect users who are consuming the Go SDK and it could be argued that this could be considered a behavioural change.

What's the impact here?

thatsmydoing: In my opinion, the existing behavior is a bug as it was leaking changes for `APIVersions` to unrelated calls. So instead of setting `APIVersions` only for the current call of `Run`, it affects all further calls to `Run` even with different sets of `APIVersions`. It would effectively just keep appending `APIVersions` every time without "resetting" it to the default set.

The code path itself is only for `ClientOnly` which rules out most helm usage. The two main use cases left are `helm template` which shouldn't see any change in behavior and anything that calls the Go SDK to generate manifests. For the latter case, there would only be a change of behavior if they call `Run` with differing `APIVersions` and the charts they're using use those `APIVersions` somehow. I believe the only users like this would be something like `argocd` that generates manifests to be applied separately against a running Kubernetes instance. But even for those, the set of `APIVersions` should be static unless you manage multiple clusters. And `argocd` actually just calls `helm template` instead of using the SDK so that's not even an actual example of use.

pkg/chartutil/capabilities.go
thatsmydoing: This is only used for parsing `--kube-version` in `helm template` so existing users should not be affected as the option doesn't even exist before this. I can inline the logic in the call site if this is a cause for confusion.

[semver.NewVersion](https://godoc.org/github.com/Masterminds/semver#NewVersion) also parses loosely so non-compliant semver should still be parsed correctly. In fact, `semverCompare` which is what's suggested to be used for comparison uses the same `semver.NewVersion` and that already works with GKE versions as when using `helm` it just provides the kubernetes version as returned by the client as-is.

This feature in [helm2](https://github.com/helm/helm/blob/a690bad98af45b015bd3da1a41f6218b1a451dbe/pkg/renderutil/render.go#L74-L80) is also implemented similarly. The main difference is just that the `Version` string is generated based on just the major/minor versions but here I build it from the parsed `String()`. I can change that to match or at least just return the original version.

afirth:
@bacongobbler I believe the review comments were addressed, and this is (still) stopping me from moving to helm3 in workflows which use template.

Using templates allows me to patch upstream outputs without adding infinite values that nobody else needs. It also allows me to `git diff` the changes that will be applied to the cluster in PRs for review

afirth:
workaround with kustomize to patch the apiVersion: 

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
resources:
[...]
patches:
- patch: |-
    - op: replace
      path: "/apiVersion"
      value: rbac.authorization.k8s.io/v1beta1
  target:
    group: rbac.authorization.k8s.io
    version: v1
    labelSelector: "app.kubernetes.io/managed-by=Helm"
```

Shini31:
I'm starting to study the migration to HELM 3 and the delete of `--kube-version` parameter is a problem for me with the use of the template command

jannfis:
We'd really love to see this PR being merged, too. Many OTS charts these days make use of `.Capabilities.KubeVersion` to render the right constructs from their templates. Without the possibility to specify a given Kubernetes version for `helm template`, this pretty much disables its usage in a disconnected environment. 

Also, the default value assumed for `.Capabilities.KubeVersion` seems to change from version to version. For example in Helm v3.4.1 it was `v1.18.0` while in Helm 3.5.1 it is `v1.20.0`. This may result in the very same Chart being rendered completely differently by `helm template`, using the same parameters if the chart makes use of `semverCompare` on `.Capabilities.KubeVersion`.

pkg/action/install.go
devzx: I have an issue and PR raised that specifically tries to address this issue. I've run in data races  consuming the SDK concurrently https://github.com/helm/helm/issues/9618 https://github.com/helm/helm/pull/9619

jannfis:
Thanks for chiming in, @bacongobbler!

No worries, I do fully understand that you're busy, and that there's lots of other work on the table. As per your suggestion, I will try to get some more attention to this PR from other people.

sbose78:
Not a maintainer, but I'll be happy to review/test this PR to make it easier to get to :)

GolubevV:
@bridgetkromhout, @bacongobbler   - Can you please clarify if this issue is going to be included into 3.6.0 version of Helm as marked? 
It seems like this PR has all checks passed and this feature is very useful for scenario where helm is used as templating engine - in tools like Spinnaker, ArgoCD, etc.

technosophos (approved):

pkg/chartutil/capabilities.go
technosophos: In discussing this during issue triage, we don't think this should be a problem for the use cases presented here.

mattfarina (approved):

PR #9047, opened 2020-11-22
Support arguments in post render

### LINKED ISSUE #9045: cmd arguments for post-renderer
Hi,

currently its not possible to pass arguments to post-render.

```
jkr@joe-nb charts % helm upgrade -i -n kube-system calico "calico" --post-renderer kubectl kustomize calico/kustomize 
Error: "helm upgrade" requires 2 arguments

Usage:  helm upgrade [RELEASE] [CHART] [flags]
```
```
jkr@joe-nb charts % helm upgrade -i -n kube-system calico "calico" --post-renderer "kubectl kustomize calico/kustomize"
Error: invalid argument "kubectl kustomize calico/kustomize" for "--post-renderer" flag: unable to find binary at kubectl kustomize calico/kustomize: exec: "kubectl kustomize calico/kustomize": stat kubectl kustomize calico/kustomize: no such file or directory
```

A shell file could be a workaround, but it isn't flexible.

Output of `helm version`:

```
version.BuildInfo{Version:"v3.4.1", GitCommit:"c4e74854886b2efe3321e185578e6db9be0a6e29", GitTreeState:"dirty", GoVersion:"go1.15.4"}
```

jkroepke:
closes #9045 

Supports arguments for post render command (see linked issue for context)

Second time go contribution. 

**If applicable**:

### DIFF
--- a/pkg/postrender/exec_test.go
+++ b/pkg/postrender/exec_test.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -37,7 +38,7 @@ sed s/FOOTEST/${1:-BARTEST}/g <&0
 func TestGetFullPath(t *testing.T) {
 	is := assert.New(t)
 	t.Run("full path resolves correctly", func(t *testing.T) {
-		testpath, cleanup := setupTestingScript(t)
+		testpath, cleanup := setupTestingScript(t, "post-render-test.sh")
 		defer cleanup()
 
 		fullPath, err := getFullPath(testpath)
@@ -46,7 +47,7 @@ func TestGetFullPath(t *testing.T) {
 	})
 
 	t.Run("relative path resolves correctly", func(t *testing.T) {
-		testpath, cleanup := setupTestingScript(t)
+		testpath, cleanup := setupTestingScript(t, "post-render-test.sh")
 		defer cleanup()
 
 		currentDir, err := os.Getwd()
@@ -59,7 +60,7 @@ func TestGetFullPath(t *testing.T) {
 	})
 
 	t.Run("binary in PATH resolves correctly", func(t *testing.T) {
-		testpath, cleanup := setupTestingScript(t)
+		testpath, cleanup := setupTestingScript(t, "post-render-test.sh")
 		defer cleanup()
 
 		realPath := os.Getenv("PATH")
@@ -113,7 +114,7 @@ func TestExecRun(t *testing.T) {
 		t.Skip("skipping on windows")
 	}
 	is := assert.New(t)
-	testpath, cleanup := setupTestingScript(t)
+	testpath, cleanup := setupTestingScript(t, "post-render-test.sh")
 	defer cleanup()
 
 	renderer, err := NewExec(testpath)
@@ -130,7 +131,7 @@ func TestExecRunWithArgs(t *testing.T) {
 		t.Skip("skipping on windows")
 	}
 	is := assert.New(t)
-	testpath, cleanup := setupTestingScript(t)
+	testpath, cleanup := setupTestingScript(t, "post-render-test.sh")
 	defer cleanup()
 
 	renderer, err := NewExec(testpath + " ARGUMENT")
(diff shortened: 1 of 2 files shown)


pkg/postrender/exec_test.go
```diff
@@ -116,7 +116,7 @@ func TestExecRun(t *testing.T) {
 	testpath, cleanup := setupTestingScript(t)
 	defer cleanup()
 
-	renderer, err := NewExec(testpath)
+	renderer, err := NewExec(testpath + " BARTEST")
```
marckhouzam, line 119: Could you make a separate test for arguments instead of changing the existing one?  I think it is important to keep a test that makes sure the "no-argument" case still works.

marckhouzam (changes requested):
This looks great!  Just an ajustement to the tests please.

pkg/postrender/exec_test.go
jkroepke: Done 👍

marckhouzam (commented):
Nice job.

I was about to approve but then I got to thinking about backwards-compatibility.
This PR slightly changes the behaviour of NewExec() compared to before.  What I mean is that it will no longer work with a binary that has a space in it.  Like a binary called "my binary.exe" (with the space as part of the name).  Maybe this is a valid case, especially on Windows.

I need another maintainer to weigh in on this.

We may at least want to support escaping a space.  Something like
`helm template rel bitnami/nginx --post-renderer "./my\ binary.exe"`

jkroepke:
@marckhouzam Alright. Take a look on the new commit please as alternative.

I used https://github.com/mattn/go-shellwords/ . The dependency was ready there.

Additionally, this support `helm template rel bitnami/nginx --post-renderer "'./my binary.exe'"` and `helm template rel bitnami/nginx --post-renderer "./my\ binary.exe"`

I didn't knew the side effects here. Please recheck.

marckhouzam:
I will be discussing the backwards-compatibility aspect of this at the developper call today.

marckhouzam:
After discussion in the developer call, it was agreed that there is a backwards-compatibility issue here.
Although the case of a binary having a space in its name (e.g., "my binary.exe") can be considered rare, having a space in the path of the binary is more common (e.g., "C:\Program Files (x86)\Internet Explorer\iexplore.exe"

Therefore, requiring the user to escape a space when they currently do not need to would break backwards-compatibility.

An alternative solution could be to add a separate flag for the arguments, maybe something like `--post-renderer-args`.  So we would have:
```
helm upgrade -i -n kube-system calico "calico" --post-renderer "kubectl"  --post-renderer-args "kustomize calico/kustomize"
```

jkroepke:
That means, the requirement to quote binaries with spaces won't fit as solution?

Like:
```
helm upgrade -i -n kube-system calico "calico" --post-renderer "'C:\Program Files (x86)\Internet Explorer\iexplore.exe' -args"
```

marckhouzam:
> That means, the requirement to quote binaries with spaces won't fit as solution?
> 
> Like:
> 
> ```
> helm upgrade -i -n kube-system calico "calico" --post-renderer "'C:\Program Files (x86)\Internet Explorer\iexplore.exe' -args"
> ```

That would not be sufficient because today someone can do:
```
helm upgrade -i -n kube-system calico "calico" --post-renderer "C:\Program Files (x86)\Internet Explorer\iexplore.exe -args"
```
without the single quotes, but if we applied your PR, this would no longer work.

Helm is very careful not to break anything that works currently.

marckhouzam:
Oh, for the sake of completeness, another option that was suggested in the call was to say that everything following `--` would be considered as arguments, like so:
```
helm upgrade -i -n kube-system calico "calico" --post-renderer "kubectl -- kustomize calico/kustomize"
```
After thinking about it, although the `--` idea is simpler to use and type, I personally think it feels like a workaround and may be forgotten by the user more often than not.  On the other hand a new `--post-renderer-args` flag has the benefit of making the new feature easier to discover.

jkroepke:
Thats breaking, too. A user could have a folder named `kubectl -- kustomize calico` that contains a binary `kustomize`.
If helm strictly avoid to be breaky, this should not allowed.

I suggest `--post-renderer-args`, too. But I can't provide this.

PR #9057, opened 2020-11-24
fix: ingress path issue

ssalaues:
Fixes a templating error with the default chart when the ingress is enabled:

Error:
```
failed to create resource: Ingress.extensions "test-chart" is invalid: [spec.rules[0].http.paths[0].path: Invalid value: "map[backend:map[serviceName:chart-example.local servicePort:80] path:/]": must be an absolute path, spec.rules[0].http.paths[0].path: Invalid value: "map[backend:map[serviceName:chart-example.local servicePort:80] path:/]": must be a valid regex]
```
Reproducible with the following values:
```
ingress:
  enabled: true
  hosts:
    - host: chart-example.local
      paths:
        - path: /
          backend:
            serviceName: chart-example.local
            servicePort: 80
```

**If applicable**:

### DIFF
--- a/pkg/chartutil/create.go
+++ b/pkg/chartutil/create.go
@@ -242,7 +242,7 @@ spec:
       http:
         paths:
           {{- range .paths }}
-          - path: {{ . }}
+          - path: {{ .path }}
             backend:
               serviceName: {{ $fullName }}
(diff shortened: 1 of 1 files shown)


bacongobbler (approved):
thanks.

PR #9060, opened 2020-11-25
[FIX]repository config file generated always '0001-01-01T00:00:00Z'

scaat:
Signed-off-by: Scaat Feng <scaat.feng@gmail.com>

Fix repository config file generated always '0001-01-01T00:00:00Z'

**If applicable**:

### DIFF
--- a/cmd/helm/repo_add.go
+++ b/cmd/helm/repo_add.go
@@ -186,6 +186,8 @@ func (o *repoAddOptions) run(out io.Writer) error {
 
 	f.Update(&c)
 
+	f.Generated = time.Now()
+
 	if err := f.WriteFile(o.repoFile, 0644); err != nil {
 		return err
 	}
--- a/pkg/repo/repo.go
+++ b/pkg/repo/repo.go
@@ -112,7 +112,6 @@ func (r *File) Remove(name string) bool {
 
 // WriteFile writes a repositories file to the given path.
 func (r *File) WriteFile(path string, perm os.FileMode) error {
-	r.Generated = time.Now()
 	data, err := yaml.Marshal(r)
 	if err != nil {
 		return err
--- a/pkg/repo/repo.go
+++ b/pkg/repo/repo.go
@@ -112,6 +112,7 @@ func (r *File) Remove(name string) bool {
 
 // WriteFile writes a repositories file to the given path.
 func (r *File) WriteFile(path string, perm os.FileMode) error {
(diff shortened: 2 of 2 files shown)


pkg/repo/repo.go
```diff
@@ -112,6 +112,7 @@ func (r *File) Remove(name string) bool {
 
 // WriteFile writes a repositories file to the given path.
 func (r *File) WriteFile(path string, perm os.FileMode) error {
+	r.Generated = time.Now()
```
mattfarina, line 115: If you use `NewFile` to create the `File` than `Generated` is set at that time. This is why using `helm repo index [directory]` has a generated time. To get `Generated` with a value of `'0001-01-01T00:00:00Z'` is means you are creating your own `File` instance without setting `Generated.

In this case, the new `Generated` value will overwrite the existing value if `NewFile` was used. We do not want that.

Can you share your use case and how your are encountering the problem.

mattfarina (commented):
In addition to the issue I noticed, a change like this should have tests.

pkg/repo/repo.go
scaat: Thank you for your reply.
I just think that when adding a repo, there should be a default generation time. But based on what you said, I also think that the time should not be generated here, it should be before calling the `WriteFile` method. What's your point of view?

cmd/helm/repo_add.go
```diff
@@ -186,6 +186,8 @@ func (o *repoAddOptions) run(out io.Writer) error {
 
 	f.Update(&c)
 
+	f.Generated = time.Now()
```
bacongobbler, line 189: This introduces potentially dangerous behaviour. This timestamp is intended to demonstrate when the index file was originally generated by `helm repo index` on the server-side, not when `helm repo add` was called. Instead this change will mask that generation timestamp.

I don't think we can accept this change.

bacongobbler (changes requested):

bacongobbler:
As @mattfarina mentioned, please use the `NewFile` function in your code. If you're creating a `File` struct directly, then the `Generated` timestamp will not be generated until you assign to it.

But at this time, there have been two changes to the SDK, both of which do not fix the original problem. At this point the recommendation going forward is to please use `NewFile` to generate new index files. That code is working as intended.

https://github.com/helm/helm/blob/037a43c98bc6331ffe7027d89e57fb1334cc1189/pkg/repo/repo.go#L42

PR #9066, opened 2020-11-26
[FIX]Unified go specification

scaat:
Unified go specification.

**If applicable**:

### DIFF
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -102,11 +102,11 @@ type Configuration struct {
 // TODO: This function is badly in need of a refactor.
 // TODO: As part of the refactor the duplicate code in cmd/helm/template.go should be removed
 //       This code has to do with writing files to disk.
-func (c *Configuration) renderResources(ch *chart.Chart, values chartutil.Values, releaseName, outputDir string, subNotes, useReleaseName, includeCrds bool, pr postrender.PostRenderer, dryRun bool) ([]*release.Hook, *bytes.Buffer, string, error) {
+func (cfg *Configuration) renderResources(ch *chart.Chart, values chartutil.Values, releaseName, outputDir string, subNotes, useReleaseName, includeCrds bool, pr postrender.PostRenderer, dryRun bool) ([]*release.Hook, *bytes.Buffer, string, error) {
 	hs := []*release.Hook{}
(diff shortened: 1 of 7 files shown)


mattfarina:
Can you explain what you mean by "Unified go specification". We are looking for a reason for this change.

scaat:
@mattfarina Thank you for your reply.
For example, such as "error string should not be capitalized", you can refer to [here](https://github.com/golang/go/wiki/CodeReviewComments#error-strings).

> Error strings should not be capitalized (unless beginning with proper nouns or acronyms) or end with punctuation, since they are usually printed following other context. That is, use fmt.Errorf("something bad") not fmt.Errorf("Something bad"), so that log.Printf("Reading %s: %v", filename, err) formats without a spurious capital letter mid-message. This does not apply to logging, which is implicitly line-oriented and not combined inside other messages.

bacongobbler:
What's the justification from changing `c` to `cfg`? I see that other actions use a `cfg` field, but the "specification" from the interface's point of view is already unified (`(c *Configuration)`). The two (function signatures vs private struct fields) are mutually exclusive.

I'm wondering the same thing @mattfarina mentions. What is the reason for that particular change?

scaat:
@bacongobbler  Thank you for your reply.
In other files, such as `hooks.go`, the method of `*Configuration` is also defined, and the receiver here is `cfg *Configuration`. This pr is just making some suggested changes, and does not actually affect the code.

bacongobbler:
Right, but as I said, those are private struct fields and do not affect the "public" API in any way. So I'm having a tough time determining what "spec" is being followed here.

When I hear someone refer to the "go specification" they're usually referring to [the public Go API](https://pkg.go.dev/helm.sh/helm/v3@v3.4.1/pkg/action). I can see there are a few spots where `c *Configuration` and `cfg *Configuration` are used interchangeably, but it's far fewer instances than what's being changed here.

bacongobbler (approved):
I'm okay with these changes. In the future, please separate each unit of work into a separate PR so they can be reviewed and approved in isolation of each other.

scaat:
@bacongobbler Thank you, I will pay attention to the distinction next time.

hickeyma (approved):
LGTM, thanks @scaat

PR #9074, opened 2020-12-01
fix(pkg/kube/client.go): cannot upgrade release after upgrade release failed in a certain case

skydai:
In a certain case, if charts updates with add a new k8s resource, and upgrade release failed. The failed reason is upgrade other res- ources failed excepet the n ew k8s resource. In this case, results cannot upgrade success anymore, it would throw an error like that " no %s with the name %q found"(release not found)

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -211,8 +211,21 @@ func (c *Client) Update(original, target ResourceList, force bool) (*Result, err
 
 		originalInfo := original.Get(info)
 		if originalInfo == nil {
-			kind := info.Mapping.GroupVersionKind.Kind
-			return errors.Errorf("no %s with the name %q found", kind, info.Name)
+			c.Log("Warning: %s/%s is found in k8s, but not found in previous release info", info.Namespace, info.Name)
+			c.Log("Deleting %q in %s...", info.Name, info.Namespace)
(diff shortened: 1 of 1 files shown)


bacongobbler:
Hi @skydai. We ask that contributors please start by first filing an issue before filing a PR. This saves both you and us some time determining the correct fix for the problem. It also provides us with some context on how to test and verify the PR fixes the issue.

In addition, we also ask for unit tests along with any changes to the codebase. This ensures that

a) we are testing the fix, and
b) we can ensure no further regressions are introduced

Please point to an issue. If none exists, please file one. Thanks.

bacongobbler (changes requested):
Hi @skydai. We ask that contributors please start by first filing an issue before filing a PR. This saves both you and us some time determining the correct fix for the problem. It also provides us with some context on how to test and verify the PR fixes the issue.

In addition, we also ask for unit tests along with any changes to the codebase. This ensures that

a) we are testing the fix, and
b) we can ensure no further regressions are introduced

Please point to an issue. If none exists, please file one. Thanks.

skydai:
ok ， i would recommit it later with requested rules

bacongobbler:
closing as inactive.

PR #9095, opened 2020-12-04
Make helm repo add insensitive to trailing slashes

### LINKED ISSUE #9092: helm repo add is sensitive to trailing slash in repo URL
Running the same `helm repo add` command twice, with and without a trailing slash in the repo URL, gives an error:

```sh
$ helm repo add owkin https://owkin.github.io/charts
"owkin" has been added to your repositories
```

```sh
$ helm repo add owkin https://owkin.github.io/charts/
Error: repository name (owkin) already exists, please specify a different name
```

This can be problematic especially since the second `helm repo add` returns a non-0 exit code. As a result, some of our automated scripts fail. This could easily be avoided if the URLs with and without trailing slash were treated as identical (see also: https://github.com/helm/helm/issues/1588 https://github.com/helm/helm/issues/1664)

Another way to address this problem would be to provide guidance as to whether or not trailing slashes SHOULD be used in URLs. I couldn't find such guidance in the current helm documentation.

Output of `helm version`:

```
```

Output of `kubectl version`:

Not applicable

Not applicable

dominikbraun:
Closes #9092. 

`helm repo add` currently is sensitive to trailing slashes. This means that adding the same repository twice - one time with and the other time without a trailing slash, as explained in issue #9092 - will result in an unnecessary error.

I've tested multiple approaches, but I found adding the `URLWithTrailingSlash` method and always comparing URLs with a trailing slash to be most elegant solution as it avoids many nested `if`s.

**If applicable**:

### DIFF
--- a/cmd/helm/repo_add.go
+++ b/cmd/helm/repo_add.go
@@ -160,10 +160,8 @@ func (o *repoAddOptions) run(out io.Writer) error {
 	// 2. When the config is different require --force-update
 	if !o.forceUpdate && f.Has(o.name) {
 		existing := f.Get(o.name)
-		existing.URL = existing.URLWithTrailingSlash()
-		c.URL = c.URLWithTrailingSlash()
 		// Only fail with an error if the configs are different.
-		if c != *existing {
+		if c.URLWithTrailingSlash() != existing.URLWithTrailingSlash() {
 			return errors.Errorf("repository name (%s) already exists, please specify a different name", o.name)
 		}
 
--- a/cmd/helm/repo_add.go
+++ b/cmd/helm/repo_add.go
@@ -160,11 +160,10 @@ func (o *repoAddOptions) run(out io.Writer) error {
 	// 2. When the config is different require --force-update
 	if !o.forceUpdate && f.Has(o.name) {
 		existing := f.Get(o.name)
-		// In case of of the repo URLs ends with a trailing slash and the other
-		// one doesn't, the config is considered to be different even though they
-		// refer to the same repository.
-		// Only fail with an error if the configs actually are different.
-		if c != *existing && c.URLWithTrailingSlash() != existing.URLWithTrailingSlash() {
+		existing.URL = existing.URLWithTrailingSlash()
+		c.URL = c.URLWithTrailingSlash()
+		// Only fail with an error if the configs are different.
+		if c != *existing {
 			return errors.Errorf("repository name (%s) already exists, please specify a different name", o.name)
 		}
 
--- a/cmd/helm/repo_add.go
+++ b/cmd/helm/repo_add.go
@@ -161,10 +161,13 @@ func (o *repoAddOptions) run(out io.Writer) error {
 	if !o.forceUpdate && f.Has(o.name) {
 		existing := f.Get(o.name)
 		if c != *existing {
-
-			// The input coming in for the name is different from what is already
-			// configured. Return an error.
-			return errors.Errorf("repository name (%s) already exists, please specify a different name", o.name)
+			// In case of of the repo URLs ends with a trailing slash and the other
+			// one doesn't, the config is considered to be different even though they
+			// refer to the same repository.
+			// Only fail with an error if the configs actually are different.
+			if c.URLWithTrailingSlash() != existing.URLWithTrailingSlash() {
+				return errors.Errorf("repository name (%s) already exists, please specify a different name", o.name)
+			}
 		}
 
 		// The add is idempotent so do nothing
--- a/cmd/helm/repo_add.go
+++ b/cmd/helm/repo_add.go
@@ -160,14 +160,12 @@ func (o *repoAddOptions) run(out io.Writer) error {
 	// 2. When the config is different require --force-update
 	if !o.forceUpdate && f.Has(o.name) {
 		existing := f.Get(o.name)
-		if c != *existing {
-			// In case of of the repo URLs ends with a trailing slash and the other
-			// one doesn't, the config is considered to be different even though they
-			// refer to the same repository.
-			// Only fail with an error if the configs actually are different.
-			if c.URLWithTrailingSlash() != existing.URLWithTrailingSlash() {
-				return errors.Errorf("repository name (%s) already exists, please specify a different name", o.name)
-			}
+		// In case of of the repo URLs ends with a trailing slash and the other
+		// one doesn't, the config is considered to be different even though they
+		// refer to the same repository.
+		// Only fail with an error if the configs actually are different.
+		if c != *existing && c.URLWithTrailingSlash() != existing.URLWithTrailingSlash() {
+			return errors.Errorf("repository name (%s) already exists, please specify a different name", o.name)
 		}
 
 		// The add is idempotent so do nothing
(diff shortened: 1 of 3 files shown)


pkg/repo/chartrepo.go
```diff
@@ -292,3 +292,13 @@ func (e *Entry) String() string {
 	}
 	return string(buf)
 }
+
+// URLWithTrailingSlash returns the repository URL with a trailing slash.
+// If the URL already ends with a slash, it will be returned unchanged.
+func (e *Entry) URLWithTrailingSlash() string {
+	if e.URL[len(e.URL)-1:] == "/" {
```
bacongobbler, line 299: This can be replaced with a call to `strings.HasSuffix(e.URL, "/")`. Less prone to array length errors and way more readable.

This could fail if `e.URL`'s length is 0.

https://golang.org/pkg/strings/#HasSuffix

You could also just replace this with `strings.TrimSuffix(e.URL, "/") + "/"`. It'll strip any trailing slash then append one to the end of the string.

bacongobbler (changes requested):

cmd/helm/repo_add.go
```diff
@@ -161,10 +161,13 @@ func (o *repoAddOptions) run(out io.Writer) error {
...
 		if c != *existing {
-
-			// The input coming in for the name is different from what is already
-			// configured. Return an error.
-			return errors.Errorf("repository name (%s) already exists, please specify a different name", o.name)
+			// In case of of the repo URLs ends with a trailing slash and the other
+			// one doesn't, the config is considered to be different even though they
+			// refer to the same repository.
+			// Only fail with an error if the configs actually are different.
+			if c.URLWithTrailingSlash() != existing.URLWithTrailingSlash() {
```
bacongobbler, line 168: Why are we performing a nested `if` check here? Why can't we perform this on line 163?

pkg/repo/chartrepo.go
dominikbraun: Right, I didn't have the edge case of `len(e.URL) == 0` in mind and went with your second approach now.

> It won't cover cases such as empty strings [...]

So the `TrimSuffix`-based solution will simply return `"/"` for empty strings. We could explicitly check if the passed string is empty and return an empty string if it is, but that wouldn't make a difference at least in this scenario.

cmd/helm/repo_add.go
dominikbraun: Wasn't sure if that rather long `if` condition would be ok 😄

cmd/helm/repo_add.go
```diff
@@ -160,10 +160,11 @@ func (o *repoAddOptions) run(out io.Writer) error {
...
 		existing := f.Get(o.name)
-		if c != *existing {
-
-			// The input coming in for the name is different from what is already
-			// configured. Return an error.
+		// In case of of the repo URLs ends with a trailing slash and the other
+		// one doesn't, the config is considered to be different even though they
+		// refer to the same repository.
+		// Only fail with an error if the configs actually are different.
+		if c != *existing && c.URLWithTrailingSlash() != existing.URLWithTrailingSlash() {
```
bacongobbler, line 167: If we're already comparing the same URLs with a trailing slash, why do we need the first comparator? Effectively you're comparing the same URL twice. The first check is no longer required.

bacongobbler (changes requested):

pkg/repo/chartrepo_test.go
```diff
@@ -400,3 +400,19 @@ func TestResolveReferenceURL(t *testing.T) {
...
+
+	if e.URLWithTrailingSlash() != urlWithTrailingSlash {
+		t.Errorf("Expected unchanged repository URL")
+	}
+
+	urlWithoutTrailingSlash := strings.TrimSuffix(urlWithTrailingSlash, "/")
+	e = Entry{URL: urlWithoutTrailingSlash}
+
+	if e.URLWithTrailingSlash() != urlWithTrailingSlash {
+		t.Errorf("Expected repository URL without trailing slash")
```
bacongobbler, line 416: Given the comparator above, shouldn't the error message report that it expected the repository URL _with_ the trailing slash?

bacongobbler (changes requested):

cmd/helm/repo_add.go
sagar-jadhav: Instead of adding `c.URLWithTrailingSlash() != existing.URLWithTrailingSlash()` check, You can set URL's of both the objects  before line `162` as below

```
c.url = c.URLWithTrailingSlash()
existing.url = existing.URLWithTrailingSlash()
```

In this manner single check `c != *existing` is sufficient.

dominikbraun: > If we're already comparing the same URLs with a trailing slash, why do we need the first comparator?

If the URL is the only thing that musn't exist already, I'll just remove the `c != existing` condition. I didn't know the exact business rules at this point.

> You can set URL's of both the objects before line 162 as below

This implies that all repo URLs will be processed with a trailing slash even if no trailing slash has been specified by the user. Would this be ok or should we just use the `c.URLWithTrailingSlash() != existing.URLWithTrailingSlash()` comparison as it is, @bacongobbler?

sagar-jadhav: @dominikbraun Although all repo URLs will be processed with a trailing slash even if no trailing slash has been specified by the user but these changes won't be persisted. See line no. 172.

bacongobbler: > This implies that all repo URLs will be processed with a trailing slash even if no trailing slash has been specified by the user. Would this be ok or should we just use the `c.URLWithTrailingSlash() != existing.URLWithTrailingSlash()` comparison as it is, @bacongobbler?

Yes, this is fine. Consider the test cases we are attempting to catch and their expected outcomes:

1. repository URLs match, neither having a trailing slash - no-op
1. repository URLs match, one having a trailing slash - no-op
1. repository URLs match, both having a trailing slash - no-op
1. repository URLs do not match, neither having a trailing slash - fail
1. repository URLs do not match, one having a trailing slash - fail
1. repository URLs do not match, both having a trailing slash - fail

Consider the question: does removing the check result in the same expected behaviour? If yes, then it's fine to remove. In this case, yes it is fine to remove and only process these URLs with a trailing slash - the end result is the same.

cmd/helm/repo_add.go
```diff
@@ -160,10 +160,10 @@ func (o *repoAddOptions) run(out io.Writer) error {
 	// 2. When the config is different require --force-update
 	if !o.forceUpdate && f.Has(o.name) {
 		existing := f.Get(o.name)
+		existing.URL = existing.URLWithTrailingSlash()
```
bacongobbler, line 163: Just remove these two lines and change line 166 to

```
if c.URLWithTrailingSlash() != *existing.URLWithTrailingSlash() {
```

Like I mentioned back in December with https://github.com/helm/helm/pull/9095#discussion_r536440127. Please ignore [this comment suggestion](https://github.com/helm/helm/pull/9095#discussion_r536517277) - by doing what he suggested, you are overwriting a variable which can affect downstream behaviour. He even called that out in another comment further down: https://github.com/helm/helm/pull/9095#discussion_r536802134

bacongobbler (changes requested):

Issue #9100, opened 2020-12-08
Helm3 filter isn't working properly; large deployments timeout before filter is in place. 


hunkeelin:
# Problem
As title said and anyone who uses helm, large deployment `helm ls` will result in timeout. I understand it's trying to read and parse all the secrets and revision. In order to actually fix it we will have to change how data is stored. Helm's filter function filter the results after it got query. 

# Solutions
**Long term solution (helm 4)**: Change how data is stored so large deployments won't get bottlenecked.

**Short term solution (helm 3)**: Current helm make use of `k8s.io/apimachinery/pkg/apis/meta/v1` library to retrieve secrets. We should use `labelselector` when calling `List` to filter results instead of querying all secrets and apply custom regex filter.  

```
 82 func (secrets *Secrets) List(filter func(*rspb.Release) bool) ([]*rspb.Relea    se, error) {
 83     lsel := kblabels.Set{"owner": "helm"}.AsSelector()
 84     if os.Getenv("helm-serviceName") != "" {
 85         lsel = kblabels.Set{
 86             "owner": "helm",
 87             "name":  os.Getenv("helm-serviceName"),
 88         }.AsSelector()
 89     }
 90     opts := metav1.ListOptions{LabelSelector: lsel.String()}
 91
 92     list, err := secrets.impl.List(context.Background(), opts)
 93     if err != nil {
 94         return nil, errors.Wrap(err, "list: failed to list")
 95     }
```
The above code is a workaround by using environment variables. In line 87 one can specify labels so the query results will be more specific. As of now, implementation of this functionality is difficult. That is because there is no way to change the behavior of the driver's output base on input.  More info in [this pull request](https://github.com/helm/helm/pull/9101)

mattfarina:
Can you provide a range for what you mean by large? Helm has used ConfigMaps (previously the default - before Secrets were possible) and now Secrets and rarely has this been an issue. Some data on what large means would be helpful in understanding the problem. Also, how that size relates to namespaces. Is it all in one namespace or spread among many.

The size will also relate to the history of releases you can store. Are you using the default history length for a release of a chart or are you extending that to be longer?

For these more complicated storage needs, Helm has [an SQL storage back-end](https://helm.sh/docs/topics/advanced/#sql-storage-backend) that can be used instead. This is an option for the large deployments.

[...]

PR #9101, opened 2020-12-08
ISSUE-9100 POC on having better filter

hunkeelin:
Relates to [ISSUE-9100](https://github.com/helm/helm/issues/9100)

This is just a hacky POC. The limitation is mainly due to how the software is architect. It architect in a way that it's doing some rpc-like kind of function calls via `driver` and there's no way to change the behavior of driver's output base on input.  I have to get around that by using `env` variables; this PR is not meant to be merged. What's the best way to implement what I am trying to do?

### DIFF
--- a/cmd/helm/list.go
+++ b/cmd/helm/list.go
@@ -124,6 +124,7 @@ func newListCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 	f.BoolVar(&client.Deployed, "deployed", false, "show deployed releases. If no other is specified, this will be automatically enabled")
 	f.BoolVar(&client.Failed, "failed", false, "show failed releases")
 	f.BoolVar(&client.Pending, "pending", false, "show pending releases")
+	f.StringVarP(&client.ServiceName, "servicename", "s", "", "Name of the helm deployment")
 	f.BoolVarP(&client.AllNamespaces, "all-namespaces", "A", false, "list releases across all namespaces")
 	f.IntVarP(&client.Limit, "max", "m", 256, "maximum number of releases to fetch")
 	f.IntVar(&client.Offset, "offset", 0, "next release name in the list, used to offset from start value")
--- a/pkg/action/list.go
+++ b/pkg/action/list.go
@@ -17,6 +17,7 @@ limitations under the License.
 package action
 
 import (
+	"os"
 	"path"
 	"regexp"
 
@@ -126,6 +127,7 @@ type List struct {
 	Uninstalled  bool
 	Superseded   bool
 	Uninstalling bool
+	ServiceName  string
 	Deployed     bool
 	Failed       bool
 	Pending      bool
@@ -154,7 +156,7 @@ func (l *List) Run() ([]*release.Release, error) {
 			return nil, err
 		}
 	}
-
+	os.Setenv("helm-serviceName", l.ServiceName)
 	results, err := l.cfg.Releases.List(func(rel *release.Release) bool {
 		// Skip anything that doesn't match the filter.
 		if filter != nil && !filter.MatchString(rel.Name) {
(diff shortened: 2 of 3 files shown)


pkg/storage/driver/secrets.go
```diff
@@ -80,6 +81,12 @@ func (secrets *Secrets) Get(key string) (*rspb.Release, error) {
 // secret fails to retrieve the releases.
 func (secrets *Secrets) List(filter func(*rspb.Release) bool) ([]*rspb.Release, error) {
 	lsel := kblabels.Set{"owner": "helm"}.AsSelector()
+	if os.Getenv("helm-serviceName") != "" {
```
mattfarina, line 84: The `pkg` directory in the Helm code is something other applications can import as a library and use in their own code. The `cmd` top level directory is for the Helm client application. Information passed to the client isn't read directly by the package. It needs to be passed in through Go. Environment variables are only used for system level information like knowing the Kuberentes context.

Also, environment variables must be in all uppercase letters and use _ as a separator.

pkg/action/list.go
```diff
@@ -154,7 +156,7 @@ func (l *List) Run() ([]*release.Release, error) {
 			return nil, err
 		}
 	}
-
+	os.Setenv("helm-serviceName", l.ServiceName)
```
mattfarina, line 159: Setting an environment variable to be picked up later is a bit of a hack. We don't pass information that way in Helm. It would need to be handled through passing through function signatures.

cmd/helm/list.go
```diff
@@ -124,6 +124,7 @@ func newListCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 	f.BoolVar(&client.Deployed, "deployed", false, "show deployed releases. If no other is specified, this will be automatically enabled")
 	f.BoolVar(&client.Failed, "failed", false, "show failed releases")
 	f.BoolVar(&client.Pending, "pending", false, "show pending releases")
+	f.StringVarP(&client.ServiceName, "servicename", "s", "", "Name of the helm deployment")
```
mattfarina, line 127: Service and Deployment are names used by Kubernetes. In this case I think you mean Release. The client property, flag name, and description should reflect that.

mattfarina (changes requested):
I'm not entirely sure what you are trying to do with `helm list` here. If you pass in a release you will have 1 item in the returned response. Or, are you looking to pair this with `-A`? Can you please provide the use case so we have more context.

cmd/helm/list.go
bacongobbler: I'd also ask how this would interact with other similar flags like `--filter` and `--selector`.

hunkeelin: The current `--filter` is a custom regex filter after helm list all the secrets. For large deployments helm usually timeout before the filter is applied.

pkg/action/list.go
hunkeelin: Yes, this is a hack and it's not meant to merge. I was hoping for any suggestion on implementing what I want to do in a right way. As of now, there's no way to change the behavior of driver's output.

bacongobbler:
I'm going ahead and closing this as something we do not plan on merging into Helm at this time.

Before opening a PR, please consider waiting for a response on next steps. This saves all of us time - it saves you from spending time writing code, and it saves us time reviewing code.

In the future, please also consider writing a [Helm Improvement Proposal](https://github.com/helm/community/tree/master/hips) first. This HIP process is what we are suggesting to new contributors looking to re-architect sections of the Helm codebase. It helps answer many of the cases maintainers need to address when considering a new feature. Backwards compatibility, documentation, security issues, etc. are all aspects that need to be discussed and addressed during the design process.

Thanks.

PR #9103, opened 2020-12-08
Updating to Kubernetes 1.19.4 package versions

mattfarina:
Note, klog is now set to v2. This is because k8s 1.19 uses klog v2.
v1.0.0, which was previously used, also had a bug in the flag
initialization which we were using. Helm was pulling klog v2.2.0
prior to this for use with k8s. Helm was using the wrong version
of the library to initialize the flags. Updating that fixes the
issues that could arise there.


bacongobbler (approved):
Good catch on the klog issue. Gotta love Go's versioning system :-)

LGTM.

PR #9116, opened 2020-12-10
Builds with go 1.15

mattfarina:
Note, our current Go 1.14 will move out of support while we are still maintaining Helm 2.5.x.


bridgetkromhout:
> while we are still maintaining Helm 2.5.x.

And of course Mr. Farina means Helm 3.5.x (because, happily, the v2 support is ended).

jdolitsky (approved):

Issue #7419, opened 2020-01-16
helm search hub should optionally return the repo URL as part of the results


omkensey:
It would be nice if `helm search hub` could give the URL of the repo (not necessarily by default, but as an optional column/JSON key), so as to avoid a roundtrip to the Hub description page for the actual repo URL if I can tell from the search listings which chart repo I want.

bacongobbler:
This is a great suggestion, @omkensey. I think adding another field to the table output displaying the repository URL would be a welcome change.

A little bit of context here: The idea with `helm search hub` returning that URL was that the chart repository maintainers were looking to add support to use that URL as input to `helm install`, like so:

```
><> helm search hub wordpress --max-col-width 0 | tail -n 1
https://hub.helm.sh/charts/bitnami/wordpress           	9.1.1        	5.3.2      	Web publishing platform for building blogs and websites.
```

then:

```
><> helm install wordpress https://hub.helm.sh/charts/bitnami/wordpress
```

cc'ing @mattfarina as he may have more context on where we are with that particular feature.

SimonHeimberg:
> A little bit of context here: ...
Thank you. I did not see this in the doc. Now I found it in `content/docs/intro/using_helm.md` since I knew what to look for.

tharun208:
I like to work on this

mattfarina:
I like the idea of adding the column. I would also like to see `><> helm install wordpress https://hub.helm.sh/charts/bitnami/wordpress` work. This might be something to add into the Artifact Hub as well.

PR #9119, opened 2020-12-12
Added flag --list-repo

ErmakovDmitriy:
The flag makes helm print the chart repo URL to use with helm repo add.
Issue: #7419

Signed-off-by: Dmitriy Ermakov <demonihin@gmail.com>

This PR allows printing of charts repo URLs. The printed URLs could be used in helm repo add <name> <URL>.
The PR is a possible solution to --list-repo flag which mentioned in #7419.

**If applicable**:


bacongobbler:
A new flag is not required to introduce a new field to the table. Just add it to the end of the table instead of the middle. That way users who use `awk` to parse information by its column number will continue to work.

ErmakovDmitriy:
Hi @bacongobbler,
I did it, please, take a look.

cmd/helm/search_hub.go
```diff
@@ -91,11 +92,17 @@ func (o *searchHubOptions) run(out io.Writer, args []string) error {
 	return o.outputFormat.Write(out, newHubSearchWriter(results, o.searchEndpoint, o.maxColWidth))
 }
 
+type hubChartRepo struct {
+	URL  string `json:"url"`
+	Name string `json:"name"`
```
bacongobbler, line 97: I don't see this being used anywhere. Can we remove this field?

Perhaps this struct should be collapsed into `hubChartElement` as one field called `Repository` of type `string` rather than of type `hubChartRepo`.

bacongobbler (changes requested):

cmd/helm/search_hub.go
ErmakovDmitriy: If you mean replace both `URL` and `Name` in `hubChartElement` with just one `Repository` which contains the URL (as string), I can do it.
On the other hand,
I used the structure to have "nicer" output in case of JSON, YAML, see below (output is cut):

```json
  {
    "url": "https://artifacthub.io/packages/helm/bitnami/phpmyadmin",
    "version": "8.1.0",
    "app_version": "5.0.4",
    "description": "phpMyAdmin is an mysql administration frontend",
    "repository": {
      "url": "https://charts.bitnami.com/bitnami",
      "name": "bitnami"
    }
  }
```

Here, I think, having repository: {url: "path", name: "name"} could have some meaning.
The same is for YAML:

```yaml
- app_version: 5.0.4
  description: phpMyAdmin is an mysql administration frontend
  repository:
    name: bitnami
    url: https://charts.bitnami.com/bitnami
  url: https://artifacthub.io/packages/helm/bitnami/phpmyadmin
  version: 8.1.0

```

bacongobbler: Good point. I hadn't thought of the JSON/YAML output formatters.

bacongobbler (approved):

technosophos:
Technically, the backward compatibility rules state that a flag is necessary to introduce new data into the output. Unfortunately this was not well-documented until HIP-0004. We will let this PR through, but in the future we will definitely require feature flags to display new output.

technosophos (approved):

mattfarina (changes requested):
I'm sorry to say that this introduces a usability problem. The output is too wide. Here's an example...

Screen Shot 2021-05-17 at 6 23 51 PM

The wrapping does not happen today at this size. On smaller displays or ever terminal windows that aren't wide on large displays there's a layout problem.

One potential fix is to put this column behind a flag again. I feel bad suggesting this but the layout issue is something we need to sort out.

ErmakovDmitriy:
Hi everyone!
At one of the stages of this PR there was a flag --list-repo.
I can restore it and make a new commit to this PR.
What do you think?

bacongobbler:
Yes. I believe that is the best path forward here for this feature request based on other's feedback.

ErmakovDmitriy:
Hi everyone,
Thanks for your patience!

I reintroduced the --list-repo flag but for some reason the circleci test fails with timeout:

```shell
#!/bin/bash -eo pipefail
make test-style
ERRO Timeout exceeded: try increasing it by passing --timeout option 
make: *** [Makefile:116: test-style] Error 4

Exited with code exit status 2
CircleCI received exit code 2
```

bacongobbler:
Can you try rebasing on top of main to see if that fixes things?

cmd/helm/search_hub.go
```diff
@@ -70,6 +71,8 @@ func newSearchHubCmd(out io.Writer) *cobra.Command {
 	f := cmd.Flags()
 	f.StringVar(&o.searchEndpoint, "endpoint", "https://hub.helm.sh", "Hub instance to query for charts")
 	f.UintVar(&o.maxColWidth, "max-col-width", 50, "maximum column width for output table")
+	f.BoolVar(&o.listRepoURL, "list-repo", false, "print charts repository URL")
```
yxxhero, line 74: why not use `list-repo-url`?

yxxhero (changes requested):

cmd/helm/search_hub.go
ErmakovDmitriy: Thank you for your suggestion. I changed the flag, please take a look.

mattfarina:
I'm sorry to say, this needs to move to 3.8.0. We need 2 maintainer approves on the latest code and we've run out of time for 3.7.0.

mattfarina (approved):

technosophos (approved):
I am confident that this is ready to ship

hickeyma (approved):
LGTM, works as expected.

PR #2636, opened 2017-06-29
feat(tiller): limit number of versions stored per release

technosophos:
This adds a new configuration option to Tiller to limit the number of
records stored per release.

Tiller stores historical release information (helm history, helm
rollback). This makes it possible to set a maximum number of versions
per release.

To enable this feature, use `helm init --history-max NNN`. Note that
because of the restrictions on Deployment objects, you will have to
re-install Tiller to add a limit.

Along the way, I found an unreported bug in the Memory storage driver.
This fixes that bug and adds substantially more tests to catch
regressions.

For local testing, run `bin/tiller --history-max 10` and then set `export HELM_HOST=localhost:44134`

Closes #2332

PR #9131, opened 2020-12-15
Fix: If `storage.Create` fails to clean up least-recent releases, return an error

### LINKED ISSUE #9145: action.Upgrade and action.Rollback create a new release, even if cleanup of least-recent release versions fails
`action.Upgrade` and `action.Rollback` allow the caller to define a maximum number of releases versions to keep, and attempt to cleanup the least-recent release versions, so that only the maximum number remain.

The cleanup can fail, e.g., because the helm client is unable to delete a ConfigMap or Secret for some release version. In that case, an error is returned:
https://github.com/helm/helm/blob/c6a00e63ef638c14be5560534c339484e35eb7ba/pkg%2Fstorage%2Fstorage.go#L161

However, the error is _ignored_:
https://github.com/helm/helm/blob/c6a00e63ef638c14be5560534c339484e35eb7ba/pkg%2Fstorage%2Fstorage.go#L62-L65

The `action.Upgrade` or `action.Rollback` caller has no way of knowing (short of parsing the log to which the helm SDK writes) that an error occurred. If the upgrade or rollback fails, and the caller retries, a new release version will be created, and releases versions will grow without bound.

dlipovetsky:
`storage.Create` cleans up the least-recent release versions. Prior to this PR, when clean up failed, it would ignore the error. This meant that a user that relied on the least-recent release version cleanup would not be notified if that cleanup failed, and release versions could grow without bound.

Users have no alternative in helm itself to clean up least-recent release versions. There was a proposal for a `prune` command (#5944), but it was not implemented.

Closes #9145.

The first commit is a unit test. It fails, because release cleanup fails, but `storage.Create` does not return an error. The second commit is the fix. With the fix, the unit test passes.

**If applicable**:

### DIFF
--- a/pkg/storage/storage.go
+++ b/pkg/storage/storage.go
@@ -61,7 +61,10 @@ func (s *Storage) Create(rls *rspb.Release) error {
 	s.Log("creating release %q", makeKey(rls.Name, rls.Version))
 	if s.MaxHistory > 0 {
 		// Want to make space for one more release.
-		s.removeLeastRecent(rls.Name, s.MaxHistory-1)
+		if err := s.removeLeastRecent(rls.Name, s.MaxHistory-1); err != nil &&
+			!errors.Is(err, driver.ErrReleaseNotFound) {
+			return err
+		}
 	}
 	return s.Driver.Create(makeKey(rls.Name, rls.Version), rls)
 }
--- a/pkg/storage/storage_test.go
+++ b/pkg/storage/storage_test.go
@@ -21,6 +21,8 @@ import (
 	"reflect"
 	"testing"
 
+	"github.com/pkg/errors"
+
 	rspb "helm.sh/helm/v3/pkg/release"
 	"helm.sh/helm/v3/pkg/storage/driver"
 )
@@ -273,6 +275,32 @@ func TestStorageHistory(t *testing.T) {
 	}
 	if len(h) != 4 {
 		t.Fatalf("Release history (%q) is empty\n", name)
+	}
+}
+
+func TestStorageRemoveLeastRecentWithError(t *testing.T) {
+	storage := Init(driver.NewMemory())
+	storage.Log = t.Logf
+
+	storage.MaxHistory = 1
+
+	const name = "angry-bird"
+
+	// setup storage with test releases
+	setup := func() {
+		// release records
+		rls1 := ReleaseTestData{Name: name, Version: 1, Status: rspb.StatusSuperseded}.ToRelease()
+
+		// create the release records in the storage
+		assertErrNil(t.Fatal, storage.Driver.Create(makeKey(rls1.Name, rls1.Version), rls1), "Storing release 'angry-bird' (v1)")
+	}
+	setup()
+
+	rls2 := ReleaseTestData{Name: name, Version: 2, Status: rspb.StatusSuperseded}.ToRelease()
+	wantErr := driver.ErrNoDeployedReleases
+	gotErr := storage.Create(rls2)
+	if !errors.Is(gotErr, wantErr) {
+		t.Fatalf("Storing release 'angry-bird' (v2) should return the error %#v, but returned %#v", wantErr, gotErr)
 	}
 }
 


joejulian (approved):
This was discovered in multiple clusters where CD was happily trying to ensure the releases reached the desired state, was failing, rolling back, and being attempted again. At the same time, a misbehaving webhook was preventing the deletion of secrets causing the clean up to fail but continue anyway. Eventually there were so many secrets that it caused a performance problem.

Allowing unbounded creation of release secrets can be a serious issue so failing as indicated in the PR description is better than succeeding in this case.

mattfarina:
@dlipovetsky Thanks for the pull request. Something about your pull request or fork is causing CircleCI to fail to accept the webhook. I'm not entirely sure what is doing on. CircleCI is giving a 400 error.

dlipovetsky:
@mattfarina Thanks for the heads-up. I thought maybe someone had to "approve" the PR before the build job would start. I'm not sure what the issue could be. Not sure if it's helpful, but this is [the relevant circleci doc](https://support.circleci.com/hc/en-us/articles/360044395953-GitHub-Updates-Results-in-400-Response-for-Webhook) I found.

sbose78:
Interesting find, thank you for the fix.

Do we know if there was a reason why errors were being ignored in the past?

dlipovetsky:
> Do we know if there was a reason why errors were being ignored in the past?

That's an excellent question. I did look at the history of the code, and it's mercifully short; `removeLeastRecent` has been returning an error, and `Create` has been ignoring that error, since the code was introduced in #2636.

In the PR conversation, I find no evidence that the error is _supposed to be_ ignored. I think it might have just been missed in the review.

dlipovetsky:
I signed up with circleci, authorizing it to access all my public repos, force-pushed, and now the job triggered.

sbose78:
Thank you for the confirmation @dlipovetsky !

dlipovetsky:
@mattfarina Is there some reason this fix got moved to the 3.5.1 milestone? I want to help get this merged. Does this need additional work, or additional reviewers?

jdolitsky (approved):
lgtm

dlipovetsky:
@jdolitsky Thanks for the review!

mattfarina:
@dlipovetsky Thank you for bringing this up. I started looking at the code and found an inconsistency in Helm. On install we display a message rather than bubbling an error all the way up. See..

https://github.com/helm/helm/blob/35da7a808d81770d197cc21f93f17f65162ac6c9/pkg/action/install.go#L372-L381

But, on upgrade we return an error for the same type of situation...

https://github.com/helm/helm/blob/35da7a808d81770d197cc21f93f17f65162ac6c9/pkg/action/upgrade.go#L139-L141

If we merge this as-is there is another case of an error bubbling all the way up.

I've added this to the Helm weekly developer meeting on Thursday to discuss. I think we should discuss this handling with more maintainers before we move on this fix (and maybe some others with it).

dlipovetsky:
> If we merge this as-is there is another case of an error bubbling all the way up.

Thanks for catching that!

> I think we should discuss this handling with more maintainers before we move on this fix (and maybe some others with it).

I'll try to join the discussion this Thursday.

I think that bubbling up the error is an important change because it empowers the user (human or machine) to decide whether to retry or abort. Right now, errors return by `removeLeastRecent` are ignored completely, so the user has no idea when the cleanup has failed. Cleanup that fails repeatedly can impact the health of the cluster.

If we are concerned about preserving existing UX, perhaps we can guard this new UX with a flag.

mattfarina (approved):
In the Helm developer call we discussed this situation. A couple things came out of it...

1. Helm should try to be consistent and return a non-zero exit code if there is an error. The current situation of returning a non-zero exit code in some commands can be duplicated elsewhere to be consistent.
2. In this case it's ok to add another error to the list. It should be noted that this error happens after the elements needed for a dry run occur but before any kubernetes manifests are sent to a cluster. This is the first thing to be done if not in a dry run. An error would occur without breaking anything for the workload in the cluster.

PR #9135, opened 2020-12-16
Fix "list all" behavior in pkg/action and pkg/cli

### LINKED ISSUE #9028: action.List client is unable to return a list of all releases

dlipovetsky:
- Fixes an error in the `pkg/action/List.Run` logic that would return only the latest releases, even when all releases were requested. Adds a unit test to verify the logic. Also fixes a "list all" CLI test, which was passing because the "golden" output was an incomplete list of releases. 
- Fixes a in the "list" shell completion that would output duplicates; this bug was masked by the logic error above.

 Fixes #9028.

**If applicable**:

### DIFF
--- a/cmd/helm/testdata/output/list-all.txt
+++ b/cmd/helm/testdata/output/list-all.txt
@@ -5,5 +5,6 @@ groot      	default  	1       	2016-01-16 00:00:01 +0000 UTC	uninstalled    	chi
 hummingbird	default  	1       	2016-01-16 00:00:03 +0000 UTC	deployed       	chickadee-1.0.0	0.0.1      
 iguana     	default  	2       	2016-01-16 00:00:04 +0000 UTC	deployed       	chickadee-1.0.0	0.0.1      
 rocket     	default  	1       	2016-01-16 00:00:02 +0000 UTC	failed         	chickadee-1.0.0	0.0.1      
+starlord   	default  	1       	2016-01-16 00:00:01 +0000 UTC	superseded     	chickadee-1.0.0	0.0.1      
 starlord   	default  	2       	2016-01-16 00:00:01 +0000 UTC	deployed       	chickadee-1.0.0	0.0.1      
 thanos     	default  	1       	2016-01-16 00:00:01 +0000 UTC	pending-install	chickadee-1.0.0	0.0.1      
--- a/cmd/helm/list.go
+++ b/cmd/helm/list.go
@@ -209,8 +209,12 @@ func compListReleases(toComplete string, cfg *action.Configuration) ([]string, c
 	}
 
 	var choices []string
+	uniqueNames := map[string]bool{}
 	for _, res := range results {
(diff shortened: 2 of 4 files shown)


cmd/helm/testdata/output/list-all.txt
```diff
@@ -5,5 +5,6 @@ groot      	default  	1       	2016-01-16 00:00:01 +0000 UTC	uninstalled    	chi
 hummingbird	default  	1       	2016-01-16 00:00:03 +0000 UTC	deployed       	chickadee-1.0.0	0.0.1      
 iguana     	default  	2       	2016-01-16 00:00:04 +0000 UTC	deployed       	chickadee-1.0.0	0.0.1      
 rocket     	default  	1       	2016-01-16 00:00:02 +0000 UTC	failed         	chickadee-1.0.0	0.0.1      
+starlord   	default  	1       	2016-01-16 00:00:01 +0000 UTC	superseded     	chickadee-1.0.0	0.0.1
```
bacongobbler, line 8: The goal with `helm list --all` is to show the latest revision of every release, regardless of what state the latest release is in. So this does appear like it would break backwards compatibility, as it not only shows the latest revision but EVERY revision in the ledger.

Can you look into refactoring the state mask while retaining prior behaviour?

bacongobbler (changes requested):

cmd/helm/testdata/output/list-all.txt
dlipovetsky: Thanks for clarifying, @bacongobbler! The expected behavior of `helm list --all` wasn't clear to me.

dlipovetsky: Getting a historical list of a particular release is precisely what I thought `pkg/action.List` was used for. I can see now that I misunderstood.

I think that part of my confusion comes from learning helm semantics from the SDK (i.e. `pkg/action`), rather than the CLI. I will close this PR, and the original issue, and instead open a PR adding some comments to `pkg/action.List`, so that the purpose of `action.List` and `action.History` is clear.

dlipovetsky: Filed #9139 with comments. Closing this PR.

PR #9141, opened 2020-12-17
Switch from BurntSushi/toml to pelletier/go-toml

hakman:
Signed-off-by: Ciprian Hacman <ciprian.hacman@sematext.com>

Helm uses https://github.com/BurntSushi/toml for marshaling to TOML. The lib is no longer supported, as stated here https://github.com/BurntSushi/toml#this-project-is-unmaintained.
github.com/pelletier/go-toml seems to be a good alternative.

n/a

**If applicable**:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module helm.sh/helm/v3
 go 1.14
 
 require (
-	github.com/BurntSushi/toml v0.3.1
+	github.com/BurntSushi/toml v0.3.1 // indirect
 	github.com/DATA-DOG/go-sqlmock v1.5.0
 	github.com/Masterminds/goutils v1.1.0
 	github.com/Masterminds/semver/v3 v3.1.1
@@ -27,6 +27,7 @@ require (
 	github.com/mitchellh/copystructure v1.0.0
 	github.com/opencontainers/go-digest v1.0.0
 	github.com/opencontainers/image-spec v1.0.1
+	github.com/pelletier/go-toml v1.8.1
 	github.com/pkg/errors v0.9.1
 	github.com/rubenv/sql-migrate v0.0.0-20200616145509-8d140a17f351
 	github.com/sirupsen/logrus v1.7.0
--- a/pkg/engine/funcs_test.go
+++ b/pkg/engine/funcs_test.go
@@ -56,7 +56,7 @@ func TestFuncs(t *testing.T) {
 	}, {
 		// Regression for https://github.com/helm/helm/issues/2271
 		tpl:    `{{ toToml . }}`,
-		expect: "[mast]\n  sail = \"white\"\n",
+		expect: "\n[mast]\n  sail = \"white\"\n",
 		vars:   map[string]map[string]string{"mast": {"sail": "white"}},
 	}, {
 		tpl:    `{{ fromYaml . }}`,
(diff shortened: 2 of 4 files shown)


pkg/engine/funcs_test.go
```diff
@@ -56,7 +56,7 @@ func TestFuncs(t *testing.T) {
 	}, {
 		// Regression for https://github.com/helm/helm/issues/2271
 		tpl:    `{{ toToml . }}`,
-		expect: "[mast]\n  sail = \"white\"\n",
+		expect: "\n[mast]\n  sail = \"white\"\n",
```
mattfarina, line 59: Why is the expected response from `toToml` changing? If someone is today using `toToml` then they may have issues in their template if they moved over to this due to a new line being added at the start.

go.mod
```diff
@@ -3,7 +3,7 @@ module helm.sh/helm/v3
 go 1.14
 
 require (
-	github.com/BurntSushi/toml v0.3.1
+	github.com/BurntSushi/toml v0.3.1 // indirect
```
mattfarina, line 6: It looks like the legacy library is still in use somewhere in the dependency tree. Do you happen to know what's using it?

mattfarina (commented):
Thanks for pointing out the issue with the package we're using. A maintained path forward would be great. Two things about this change did catch my attention...

go.mod
hakman: There are still many projects using `BurntSushi/toml`, like containerd that was also thinking of changing https://github.com/containerd/containerd/issues/4370.

pkg/engine/funcs_test.go
hakman: I agree that if you expect exactly the same output, things may break. Though, as long as the TOML that is exported is valid, any parser should ignore new lines, so don't see this as a big issue.

bacongobbler:
While I sympathize with the need to update to a maintained client library, what's wrong with the current library? As I understand it, the current parser works fine with no security issues. Switching it to pelletier/go-toml does appear to introduce backwards-incompatible changes, and I'm concerned that may break certain use cases relying on burntsushi's TOML parser (such as the case in `toToml`.

If there's a fork of burntsushi's TOML library that is considered a successor, that may be the more tenable option as that would mitigate the compatibility risk.

bacongobbler:
ping^

hakman:
Hi @bacongobbler. Sorry for the very long delay in answering, but had some pretty busy months.
To answer your questions:
1. There's nothing wrong with the current lib, except the TOML version support which is `v0.4.0` for [BurntSushi/toml](https://github.com/BurntSushi/toml) and `v1.0.0-rc.3` for [pelletier/go-toml](https://github.com/pelletier/go-toml). Not sure how much difference this makes in practice.
2. The output of the library should be compatible with the TOML specs and so, anything parsing back the output generated by `toToml` should still work.
3. I am not aware of any better TOML lib. It doesn't seem to be that popular as a config format lately.

To conclude, I noticed this while looking for such a lib for a project of mine and ended up noticing that many projects still pull BurntSushi by vendoring Helm. This made me think that would be a good idea to update to something more maintained.

I don't mind if this PR is considered too dangerous and it is not accepted. I still think the risks are small due to the reduced popularity of TOML.

hakman:
Seems that containerd just switched to https://github.com/pelletier/go-toml: https://github.com/containerd/containerd/pull/5278.

bacongobbler:
Closing out this topic, it appears they’re planning to revive the project.

https://github.com/BurntSushi/toml/issues/272

pelletier:
Hi! Just saw this, if you are interest go-toml v2 is nearing completion. It aims to be fast for marshal/unmarshal operations (https://github.com/pelletier/go-toml/tree/v2).

PR #9142, opened 2020-12-17
Prefix USE_SUDO with HELM_INSTALL

metal3d:

Because we probably want to set HELM_INSTALL_DIR and USE_SUDO in our`.bashrc` file, the variable name `USE_SUDO` is too wide. It's safier to use `HELM_INSTALL_USE_SUDO`.

Maybe we can leave `USE_SUDO` that reads `HELM_INSTALL_USE_SUDO` for backward compatibility

Signed-off-by: Patrice Ferlet <metal3d@gmail.com>

### DIFF
--- a/scripts/get-helm-3
+++ b/scripts/get-helm-3
@@ -18,7 +18,7 @@
 # the package manager for Go: https://github.com/Masterminds/glide.sh/blob/master/get
 
 : ${BINARY_NAME:="helm"}
-: ${USE_SUDO:="true"}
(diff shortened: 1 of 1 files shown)


bacongobbler:
This breaks backwards compatibility for those currently using the script. For that reason we cannot accept this PR. Just unset USE_SUDO before invoking the script.

metal3d:
I can change this to have backward compatibility so

PR #9143, opened 2020-12-17
Make lint error message more specific

sergueidmitriev:
the original error message does not specify where 'name' object belongs too, it's confusing to troubleshoot the charts relying only on lint console output. Current output example:

`````
==> Linting ./helm
[INFO] values.yaml: file does not exist
[ERROR] templates/deployment.yaml: object name must be between 0 and 253 characters: ""
[ERROR] templates/hpa.yaml: object name must be between 0 and 253 characters: ""
`````

Signed-off-by: Serguei Dmitriev <serguei.dmitriev@ssense.com>


bacongobbler:
Looks like this has been fixed in a more recent release of Helm.

PR #9152, opened 2020-12-18
fix: Don't write to output-dir when doing a "helm install"

sbose78:
Even though --output-dir isn't a valid flag for `install`, I noticed there was code lying around for that ? We already have `helm template`-specific code that writes to the filesystem. 

```
$ helm install mychart --output-dir=/tmp/dump
Error: unknown flag: --output-dir
```

Removed the unit tests since they were no longer needed ( or should I keep them and check if they *aren't* created anymore :) )

### DIFF
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -21,7 +21,6 @@ import (
 	"fmt"
 	"os"
 	"path"
-	"path/filepath"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


sbose78:
I do feel uncomfortable removing tests, are acceptance tests being run on every PR on the CI?

bacongobbler:
`helm template` uses the install action. Currently `helm template` uses `OutputDir` directly, but it is intended for this code to me ported over to `pkg/action`. I think the better course of action here would be to see if we can port over the code in `cmd/helm/template.go` and move that over to `pkg/action/install.go`. That way users who invoke the install action with the OutputDir flag enabled will have the output written to a file.

There's also the risk here from the SDK point of view that this would break said SDK users relying on this functionality.

bacongobbler:
checking to see what's the status on this PR

sbose78:
Let me get this done this week @bacongobbler

PR #9155, opened 2020-12-20
Print an error message when helm fails to access `~/.config`

### LINKED ISSUE #8606: Helm Fails silently when `/.config` isn't owned appropriately
Output of `helm version`:  version.BuildInfo{Version:"v3.3.0", GitCommit:"8a4aeec08d67a7b84472007529e8097ec3742105", GitTreeState:"dirty", GoVersion:"go1.14.7"}

Output of `kubectl version`:
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.6", GitCommit:"dff82dc0de47299ab66c83c626e08b245ab19037", GitTreeState:"clean", BuildDate:"2020-07-16T14:19:25Z", GoVersion:"go1.13.13", Compiler:"gc", Platform:"linux/amd64"}

Cloud Provider/Platform (AKS, GKE, Minikube etc.):  ???

-------------------------------
```bash
helm        # produces no output
echo $?   # produces 1
```
At least the rc status was 1.  

It turns out the issue is that my ~/.config/ directory was owned by root:root rather than myself.  

As a work around:

```bash
sudo chown -R $USER:$USER ~/.config/
helm  # produces help
```

hanyucui:
This is done by calling log.Fatal() instead of debug(). I think an
error causing helm to exit can be considered fatal.

This closes #8606

Signed-off-by: Hanyu Cui <hanyuc.gml@gmail.com>

**What this PR does / why we need it**: This PR closes #8606

**If applicable**:

### DIFF
--- a/cmd/helm/helm.go
+++ b/cmd/helm/helm.go
@@ -62,8 +62,7 @@ func main() {
 	actionConfig := new(action.Configuration)
 	cmd, err := newRootCmd(actionConfig, os.Stdout, os.Args[1:])
 	if err != nil {
-		debug("%+v", err)
-		os.Exit(1)
+		log.Fatal(err)
 	}
 
 	// run when each command's execute method is called


bacongobbler:
There are many commands that do not rely on reading from ~/.config. For example `helm help` or `helm docs`. Switching from a debug to a fatal error would prevent those commands from executing.

Because of that we cannot accept this PR.

hanyucui:
@bacongobbler Thanks for taking a look. Quick question, I was just replacing `debug()` and `os.Exit()` with `log.Fatal()`. Are you saying there are situations where `os.Exit()` does not exit? In my testing, when `~/.config` is inaccessible, all commands, including `help`, `version`, and `docs, exit silently. I agree maybe a more comprehensive fix is needed to not fail the commands that do not need to access `~/.config`, but this PR doesn't hurt.

PR #9170, opened 2020-12-24
Force input to start at templates/ directory so path matching works as expected.

### LINKED ISSUE #9033: helm template --show-only option path matching is confusing
Output of `helm version`:
```
```

Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.5", GitCommit:"e6503f8d8f769ace2f338794c914a96fc335df0f", GitTreeState:"clean", BuildDate:"2020-06-26T03:47:41Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
```

N/A

The helm template flag ```--show-only, -s``` seems to have a confusing pattern for selecting the path of the template you want to render when using a local chart. 

How to reproduce:
```
helm create example

# render template
helm template ./example
```

This will render the template as expected producing output like the following (taking the last rendered template):
```
...

---
# Source: example/templates/tests/test-connection.yaml
apiVersion: v1
kind: Pod
metadata:
  name: "RELEASE-NAME-example-test-connection"
  labels:
    helm.sh/chart: example-0.1.0
    app.kubernetes.io/name: example
    app.kubernetes.io/instance: RELEASE-NAME
  annotations:
    "helm.sh/hook": test
spec:
  containers:
    - name: wget
      image: busybox
      command: ['wget']
      args: ['RELEASE-NAME-example:80']
  restartPolicy: Never
```

Based on the ```# Source: example/templates/tests/test-connection.yaml``` line, I would expect that I could use the following to render just that file:
```
helm template ./example -s example/templates/tests/test-connection.yaml
Error: could not find template example/templates/tests/test-connection.yaml in chart
```

However, that does not work. Instead, after a lot of trial and error, I found that the following works:
```
helm template ./example -s templates/tests/test-connection.yaml
```

To me it does not seem intuitive that the path displayed on that output is not what you would put for the path. I looked at the code and found where this was happening, and I would like to propose a small change that might make this more straight forward
[https://github.com/aro5000/helm/commit/821296f0113dec44ec6bc2f50374797f65b27982](https://github.com/aro5000/helm/commit/821296f0113dec44ec6bc2f50374797f65b27982)

This is the first time I have ever submitted an issue for anything like this so please be patient with me and let me know if there is anything else I need to do or information I can provide. I'm a heavy user of this project and I think all the work happening in this ecosystem is amazing! If the code suggestion does not seem reasonable, please let me know if there is a place where I could maybe make the docs a bit more obvious or if I should submit a PR. I would like to get involved and contribute any way I can!

aro5000:
Signed-off-by: Aaron Stults <aaron710@gmail.com>

closes #9033 
When running `helm template` with the `--show-only` option, it assumes the path starts at the `templates/` directory which isn't intuitive if you are running the command from outside of the helm chart directory on a local chart. This PR trims the path provided to match starting at the assumed templates/ directory. An example is provided in #9033 

As mentioned in the issue, if this is better resolved by updating documentation, I'm happy to do that as well if someone can point me in the right direction!

### DIFF
--- a/cmd/helm/template.go
+++ b/cmd/helm/template.go
@@ -119,6 +119,17 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 						missing := true
 						// Use linux-style filepath separators to unify user's input path
 						f = filepath.ToSlash(f)
+						// the following path matching assumes starting at the "templates/" directory
+						// so we will split the input path on "/" and reslice starting at "templates" if it exists
+						fileSplit := strings.Split(f, "/")
+						for i,v := range fileSplit {
+							if v == "templates" {
+								fileSplit = fileSplit[i:]
+								break;
+							}
+						}
+						f = strings.Join(fileSplit, "/")
+
 						for _, manifestKey := range manifestsKeys {
 							manifest := splitManifests[manifestKey]
 							submatch := manifestNameRegex.FindStringSubmatch(manifest)


bacongobbler:
I'd prefer a docs update rather than hardcoding this behaviour. Thanks!

PR #9174, opened 2020-12-28
feat(helm): add new update-policy annotation to recreate resources

### LINKED ISSUE #7082: proposal: handle immutable resources on update

dastrobu:
Add annotation handling of new annotation helm.sh/update-policy with values recreate-on-conflict and recreate-on-invalid values to better deal with the following cases.
The server responds with 'invalid' e.g. if a certain resource contains an immutable field, which cannot be patched or updated by a PUT. In theses cases it can be helpful to delete the resource and recreate it. Examples for theses cases are:

 * roleRef in RoleBinding
 * spec.clusterIP in Service
 * parameters in StorageClass

If a chart developer has made sure that recreating a resource is safe, the annotation can be added to a resource template. Alternatively a user may add the annotation manually to an existing resource to allow Helm to recreate a certain resource.

Adding the new annotation in general by default or changing the default behaviour of Helm is not a good idea, since deletion and recreation of certain resources can lead to an unwanted or undefined state as described in detail in https://github.com/helm/helm/pull/7431#issuecomment-666735299:

"[...] selectively deleting/re-creating volume-backed resources (Secrets, Services, ConfigMaps, PVCs) without touching the Deployment. If a Persistent Volume Claim is deleted and re-created, there is a risk that an application (Deployment) consuming that Persistent Volume Claim will not receive the new volume's mount point. The same can be said for Secrets, Configmaps, Services, and other resources which are deployed and relied upon by another resource. Environment variables populated in a Deployment from a Service won't be updated until the Pods are destroyed."

Closes #7082.

This is an alternative approach to https://github.com/helm/helm/pull/7431 as proposed in https://github.com/helm/helm/issues/7082#issuecomment-687752364. 

There is a new timeout parameter introduced to `Update` in a backwards compatible way as `UpdateWithTimeout` as suggested in https://github.com/helm/helm/pull/9174#discussion_r1186547143. In Helm 4 this should be cleaned up as stated in the comments.  

The `update-policy` and its possible values is a proposal related to the `delete-policy` of hooks and could be changed to something different. 

**If applicable**:

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -25,6 +25,7 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
+	"sort"
 	"strings"
 	"sync"
 	"time"
@@ -47,6 +48,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/strategicpatch"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/cli-runtime/pkg/resource"
@@ -56,6 +58,17 @@ import (
 	cachetools "k8s.io/client-go/tools/cache"
 	watchtools "k8s.io/client-go/tools/watch"
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
+)
+
+var accessor = meta.NewAccessor()
+
+// Annotation for resource recreation on update. The two cases "conflict" and "invalid" are the two cases where
+// kubectl apply --force tries a recreate, so those seem to be the most relevant two cases users usually want
+// to control.
+const (
+	updatePolicyAnnotation         = "helm.sh/update-policy"
+	updatePolicyRecreateOnConflict = "recreate-on-conflict"
+	updatePolicyRecreateOnInvalid  = "recreate-on-invalid"
 )
 
 // ErrNoObjectsVisited indicates that during a visit operation, no matching objects were found.
@@ -379,14 +392,20 @@ func (c *Client) BuildTable(reader io.Reader, validate bool) (ResourceList, erro
 	return result, scrubValidationError(err)
 }
 
-// Update takes the current list of objects and target list of objects and
+// Update is a wrapper for UpdateWithTimeout to avoid a breaking API change.
+// Deprecated: prefer UpdateWithTimeout, cannot be removed until Helm 4
+func (c *Client) Update(original, target ResourceList, force bool) (*Result, error) {
+	return c.UpdateWithTimeout(original, target, force, time.Duration(0))
+}
+
+// UpdateWithTimeout takes the current list of objects and target list of objects and
 // creates resources that don't already exist, updates resources that have been
 // modified in the target configuration, and deletes resources from the current
 // configuration that are not present in the target configuration. If an error
 // occurs, a Result will still be returned with the error, containing all
 // resource updates, creations, and deletions that were attempted. These can be
 // used for cleanup or other logging purposes.
-func (c *Client) Update(original, target ResourceList, force bool) (*Result, error) {
+func (c *Client) UpdateWithTimeout(original, target ResourceList, force bool, timeout time.Duration) (*Result, error) {
 	updateErrors := []string{}
 	res := &Result{}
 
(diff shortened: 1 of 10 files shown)


z4ce:
Did you ever get any feedback on this? It seems like a pretty good idea to me.  I wonder why the DCO check hasn't passed on this.  

IMHO this PR seems preferable to #7431 should that one be closed in favor of this?

pkg/kube/client.go
```diff
@@ -194,7 +207,7 @@ func (c *Client) Build(reader io.Reader, validate bool) (ResourceList, error) {
 // occurs, a Result will still be returned with the error, containing all
 // resource updates, creations, and deletions that were attempted. These can be
 // used for cleanup or other logging purposes.
-func (c *Client) Update(original, target ResourceList, force bool) (*Result, error) {
+func (c *Client) Update(original, target ResourceList, force bool, timeout time.Duration) (*Result, error) {
```
joejulian, line 210: As you said, this is a breaking change as written. You can work around that by doing something like this:

```suggestion
func (c *Client) Update(original, target ResourceList, force bool) (*Result, error) {
	return UpdateWithTimeout(original, target, force, time.Duration(0))
}	

func (c *Client) UpdateWithTimeout(original, target ResourceList, force bool, timeout time.Duration) (*Result, error) {
```

Where if timeout is 0 you revert to the original behavior.

joejulian:
Same question:
https://github.com/helm/helm/pull/7431#issuecomment-1536877604

pkg/kube/client.go
dastrobu: @joejulian thanks for the suggestion. I noticed that there was a similar issue in the meantime which lead to: 
```
type InterfaceExt interface {
	// WaitForDelete wait up to the given timeout for the specified resources to be deleted.
	WaitForDelete(resources ResourceList, timeout time.Duration) error
}
```
So I will follow that pattern and remove the breaking change by introducing a trait like additional interface.

dastrobu: @joejulian PR is cleaned up and all breaking changes have been removed as suggested. Looking forward to your feedback.

dastrobu:
@joejulian could you unlock #7082 to be able to comment on in again, since I am not able to comment on this issue anymore?

scottrigby:
> @joejulian could you unlock #7082 to be able to comment on in again, since I am not able to comment on this issue anymore?

I unlocked that issue

dastrobu:
remove stale

f1603849451-arch:
I’d like to ask about the current status of this feature and when it is expected to be merged.

silence-coding:
Are there any further updates on this feature? I’ve noticed that the current `forceReplace` configuration is too coarse-grained and does not meet production-level requirements. In a production environment, we do not want to trigger a full delete-and-recreate process, as this could result in prolonged service downtime.

TerryHowe:
There are a lot of issues with this PR and then there are conflicts as well.

dastrobu:
This PR is pretty old, unless the maintainers show interest in the proposal  https://github.com/helm/helm/issues/7082 I dont think it makes sense to spend time on this PR (rebasing, ...).

silence-coding:
@TerryHowe Is there enough interest from the Helm maintainers/community to accept this feature?

silence-coding:
Are there any other features that can address this issue? We can't expect users to uninstall the entire chart before proceeding, can we?

joejulian:
Thanks for this, and for keeping it alive as long as you did. The need is real: some resources have immutable fields, and a normal update or apply cannot get past them.

Helm 4 now covers both cases this PR set out to handle, as operator-invoked flags rather than a chart annotation:

- `--force-replace` recreates resources that cannot be updated in place, which is the `recreate-on-invalid` / immutable-field case. (`--force` is now deprecated in favor of `--force-replace`.)
- `--force-conflicts` forces server-side-apply changes against field-ownership conflicts, which covers the `recreate-on-conflict` case.

To be precise about server-side apply: no apply strategy can mutate a truly immutable field, so this is not SSA quietly solving it. It is that the delete-and-recreate path now has a first-class flag (`--force-replace`) alongside the SSA conflict override.

The one real difference from this PR is that those are flags the person running the upgrade sets, not an annotation the chart author sets, and that is deliberate. Recreating a resource is destructive and cluster-state-dependent, and whether it is safe to delete and recreate a given object (a PVC, or anything holding state) is something the operator knows at upgrade time and the chart author usually does not. Keeping that decision with the operator rather than baking it into the chart is the separation we hold elsewhere, so the flag form is the shape we want.

Given that, plus the conflicts against the Helm 4 update path (rewritten for server-side apply) and the related #7431 already being closed, I am going to close this as superseded. I genuinely appreciate the work, and it surfaced a real need that Helm 4 ended up filling.

If you think there is still a case for a targeted, per-resource, only-on-invalid recreate beyond these flags, that would be a Helm Improvement Proposal conversation rather than a reopen, since it changes the chart contract.

Issue #9150, opened 2020-12-18
sub-sub-charts of aliased sub-charts not evaluated correctly


dastrobu:
# Summary

Given a chart with a subchart `child` and a sub-sub-chart `grandchild`
```
charts/
 + child
   + grandchild
```
where the sub chart is referenced twice with an alias from the `Chart.yaml`:
```
 dependencies:
 - name: child
   alias: foo
   version: 1.0.0
 - name: child
   alias: bar
   version: 1.0.0
```

It is expected that the resources of the `grandchild` chart are rendered twice with different values specified for the two aliases `foo` and `bar`. 

# [SSCCE](http://www.sscce.org)
To reproduce it fully, run: 

```
mkdir -p parent/charts/child/charts/grandchild/templates
mkdir -p parent/charts/child/templates

echo '
apiVersion: v2
appVersion: 1.0.0
name: grandchild
type: application
version: 1.0.0
' > parent/charts/child/charts/grandchild/Chart.yaml

echo '
apiVersion: v2
appVersion: 1.0.0
name: child
type: application
version: 1.0.0
' > parent/charts/child/Chart.yaml

echo '
apiVersion: v2
appVersion: 1.0.0
name: parent
type: application
version: 1.0.0

dependencies:
- name: child
  alias: foo
  version: 1.0.0
- name: child
  alias: bar
  version: 1.0.0
' > parent/Chart.yaml

echo '
apiVersion: v1
kind: ConfigMap
metadata:
  name: {{ .Chart.Name }}
data:
{{ toYaml .Values | indent 2}}
' > parent/charts/child/templates/dummy.yaml

echo '
apiVersion: v1
kind: ConfigMap
metadata:
  name: {{ .Chart.Name }}-{{ .Values.from }}
data:
{{ toYaml .Values | indent 2}}
' > parent/charts/child/charts/grandchild/templates/dummy.yaml


echo '
foo:
  grandchild:
    from: foo
bar:
  grandchild:
    from: bar
' > parent/values.yaml

helm template parent
```

which prints out: 
```
---
# Source: parent/charts/bar/templates/dummy.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: bar
data:
  global: {}
  grandchild:
    from: bar
    global: {}
---
# Source: parent/charts/child/charts/grandchild/templates/dummy.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: grandchild-bar
data:
  from: bar
  global: {}
---
[...]

PR #9175, opened 2020-12-28
fix: copy dependencies on aliasing to avoid sharing chart references on multiply aliased dependencies

dastrobu:
Dependencies keep a reference on their parent chart, which breaks if a chart reference is shared among multiple aliases.
By copying the dependencies, parent information can be set correctly to render the templates as expected later on.

Note that this change will make ChartFullPath return a different path for sub-subcharts. It will contain the alias names instead of the path to the chart files which makes it consistent with paths to templates on the subchart level.

Closes #9150

Signed-off-by: Daniel Strobusch <1847260+dastrobu@users.noreply.github.com>

**If applicable**:
- [ ] this PR contains documentation
- [x] this PR contains unit tests
- [x] this PR has been tested for backwards compatibility

PR #9176, opened 2020-12-28
lint and validate dependency metadata to reference dependencies with …

### LINKED ISSUE #9169: dependency name (and alias) clashes should be reported as errors
While trying to fix #9150 I found another class of issues with aliases and name clashes of dependencies. 

I am proposing to report the following situation as errors on linting (and template/installing/upgrading) a chart: 

```
dependencies:
- name: foo
  alias: baz # ← baz used twice
  version: 1.0.0
- name: bar
  alias: baz # ← baz used twice
  version: 1.0.0
```
```
dependencies:
- name: foo
  alias: bar # ← shadows chart below
  version: 1.0.0
- name: bar
  version: 1.0.0
```
```
dependencies:
- name: foo
  version: 1.0.0
- name: foo # ← chart with same name as above (although version or repo will be different, this will not work currently)
  version: 1.2.3
```
While the last example may be working in the future (see #7413), I can't see any valid use case for the first two examples. 
Currently Helm renders templates, resulting in more or less undefined behaviour. Some templates will override each other, others won't, depending on file names in the actual charts and on the chart names itself. 

Output of `helm version`:
`version.BuildInfo{Version:"v3.4.2", GitCommit:"23dd3af5e19a02d4f4baa5b2f242645a1a3af629", GitTreeState:"clean", GoVersion:"go1.14.13"}`

dastrobu:
…a unique key (name or alias)

Report charts with the following bad dependency specifications as bad charts:

    dependencies:
    - name: foo
      alias: baz # ← baz used twice
      version: 1.0.0
    - name: bar
      alias: baz # ← baz used twice
      version: 1.0.0

    dependencies:
    - name: foo
      alias: bar # ← shadows chart below
      version: 1.0.0
    - name: bar
      version: 1.0.0

    dependencies:
    - name: foo
      version: 1.0.0
    - name: foo # ← chart with same name as above (although version or repo will be different, this will not work currently)
      version: 1.2.3

Closes #9169

**If applicable**:


pkg/chart/metadata_test.go
```diff
@@ -72,6 +72,32 @@ func TestValidate(t *testing.T) {
...
 		},
+		{
+			&Metadata{
+				Name:       "test",
+				APIVersion: "v2",
+				Version:    "1.0",
+				Type:       "application",
+				Dependencies: []*Dependency{
+					{Name: "foo", Alias: ""},
+					{Name: "foo", Alias: ""},
```
sbose78, line 83: Do you want to add different versions to these dependencies ( third use case here https://github.com/helm/helm/issues/9169 ) ? That way, in future, when supported, we would have more explicit usecases.

dastrobu: @sbose78 thanks for reviewing, not sure if I got your comment. I added two tests for first and second case from #9169. I omitted the third case since it would not add any coverage w.r.t to the current algorithm and if Helm 4 will support the third case (same name different version/repo) the code will need adjustment anyway (plus positive test). 
So do you want me to add another test for third case which might get removed in Helm 4? I can easily do that if that was your intention, please just clarify.

sbose78: Correct.
In fact, you could modify the existing test case a little to highlight the third use case of yours just so that your findings are documented in the form of a test.

dastrobu: Thanks for clarifying. I added two cases (different versions, different repos) and extended test cases with a name for better documentation. Also pointed out that the latter two cases could make sense in the future in a comment. Please take another look at the PR.

sbose78: Thank you!

sbose78 (approved):
LGTM

dastrobu:
Rebased onto current main.

jburnitz:
why isn't this merged

dastrobu:
Rebased onto current main.

pkg/chart/metadata_test.go
```diff
@@ -97,6 +168,7 @@ func TestValidate(t *testing.T) {
 			ValidationError("maintainer cannot be an empty list"),
 		},
 		{
+			"version valid",
```
MaximevanderSmissen, line 171: Suggestion: "version invalid"

Maxine-van-der-Smissen (approved):
Note: I am not a maintainer.

Thanks for making the PR

pkg/chart/metadata_test.go
dastrobu: good catch! Thanks for the review.

joejulian (approved):
I tried being creative but I can't break this. Nice work!

sabre1041 (approved):
LGTM

mattfarina (approved):

dbluxo:
I'm now (v3.14.0) getting error messages, is this a non valid use case?:

```
dependencies:
  - name: kube-rbac-proxy
    version: "0.9.1"
    repository: "<...>"
    alias: loki-query-proxy
    condition: loki-query-proxy.enabled
  - name: kube-rbac-proxy
    version: "0.9.1"
    repository: "<...>"
    alias: loki-ingestion-proxy
    condition: loki-ingestion-proxy.enabled
   ```

`index.go:366: skipping loading invalid entry for chart "<...>" "1.3.0" from <...>: validation: more than one dependency with name or alias "kube-rbac-proxy"`

bt909:
According to the [documentation for the Chart.yaml](https://helm.sh/docs/topics/charts/#the-chartyaml-file) using name twice with separate aliases seems to be okay.

Quote: alias: (optional) Alias to be used for the chart. Useful when you have to add the same chart multiple times

But now this seemed to be treated as error!

dastrobu:
> `index.go:366: skipping loading invalid entry for chart "<...>" "1.3.0" from <...>: validation: more than one dependency with name or alias "kube-rbac-proxy"`
> 
> Or am I missing something?

@dbluxo I guess your case should work as long as the repo and version is identical for both dependencies. Otherwise you will hit the case described in https://github.com/helm/helm/issues/9169 (undefined behaviour). 
As this PR is closed, I'd suggest to create a new issue for this. 
I can try to create a fix for this. Note that I am not a maintainer, though.

dbluxo:
@dastrobu Thank you for your response. I have created an [issue](https://github.com/helm/helm/issues/12748), would be great if our use case (same repository and version) would work again.

dastrobu:
While looking for a fix I found that aliasing as you wish to use it will actually not work in many cases, see #9150 and https://github.com/helm/helm/pull/9175/files#r1043661769. 
As I reported this three years ago, I did not remember all the details. 

My suggestion would be to lift the validation constraint on your case, when #9150 is fixed. Otherwise, there is a high risk of getting undefined behavior, but maintainers may have other opinions.

ziqianggeoffreychen:
> > `index.go:366: skipping loading invalid entry for chart "<...>" "1.3.0" from <...>: validation: more than one dependency with name or alias "kube-rbac-proxy"`
> > Or am I missing something?
> 
> @dbluxo I guess your case should work as long as the repo and version is identical for both dependencies. Otherwise you will hit the case described in #9169 (undefined behaviour). As this PR is closed, I'd suggest to create a new issue for this. I can try to create a fix for this. Note that I am not a maintainer, though.

Hi @dastrobu , I opened an issue related to this change. In my case, the two dependencies are exactly the same with different aliases. But the alias is removed by JFrog in its index.yaml metadata. If all versions of the helm charts are with the same kind (aliases duplicate), it seems they will show up in search result. But that's not reliable. Refer to my issue for more details. Thanks.
https://github.com/helm/helm/issues/12744

PR #9177, opened 2020-12-29
Fix a func name in pkg/cli/values/options_test.go

hs0210:
Signed-off-by: Hu Shuai <hus.fnst@cn.fujitsu.com>

**If applicable**:


bacongobbler:
The current name is correct as it's testing to check whether merging two values works. `mergeMaps` is an internal function and could be replace by another function call.

Issue #4558, opened 2018-08-29
helm upgrade > timeout on pre-upgrade hook > revision stuck in `PENDING_UPGRADE` and multiple `DEPLOYED` revisions arise soon


consideRatio:
## Reproduction and symptom
1. `helm upgrade` with a helm pre-upgrade hook that times out.
2. `Error: UPGRADE FAILED: timed out waiting for the condition`.
3. `helm history my-release-name`
   ```bash
   # the last line...
   22      	Wed Aug 29 17:59:48 2018	PENDING_UPGRADE	jupyterhub-0.7-04ccf1a 	Preparing upgrade
   ```

## Expected outcome
The revision should end up in `FAILED` rather than `PENDING_UPGRADE` right?

consideRatio:
By running the same `helm upgrade` over, I ended up with multiple revisions being `DEPLOYED`.

My helm hooks are mostly a DaemonSet pulling images with init-containers with a pause image used by the main container. I also have a Job that awaits the DaemonSet to have the desired ready pods. All hooks have the following annotations.

```
  annotations:
    "helm.sh/hook": pre-install,pre-upgrade
    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
    "helm.sh/hook-weight": "0"
``` 

I think that `tiller` is fooled by the delete-policy of `before-hook-creation`. Something like this...

1. An upgrade is done, getting stuck in the pre-upgrade hook for some reason. This upgrade will remain in `PENDING_UPGRADE` no matter what if the helm client times out or not.
2. A new upgrade is done, the resources are removed and tiller believes the new hooks actually finished when they were deleted and progresses towards the actual upgrade.

An indication of this to be somewhat correct, is that the upgrade were considered to be succeeded in the end when it really should or could not be! I had asked it to pull images never to be found in the pre-upgrade hooks, but the actual upgrade happened even though those images did not exist. Somehow, tiller was fooled to believe the hooks completed successfully!

andytom:
I have notice the same issue but we are not using the `before-hook-creation`. Our releases still get stuck but it doesn't seem to create duplicate deployed releases.

rodriguez-facundo:
Same: 

```bash
➜ helm history jupyterhub
REVISION	UPDATED                 	STATUS         	CHART                 	DESCRIPTION
1       	Mon Aug 12 21:29:28 2019	DEPLOYED       	jupyterhub-0.8-ff69a77	Install complete
[...]

Issue #8987, opened 2020-11-05
Helm v3.4 Error: UPGRADE FAILED: another operation (install/upgrade/rollback) is in progress


jlcrow:
Upgraded from Helm 3.3 to Helm 3.4, existing charts started failing the upgrade with the message:
Error: UPGRADE FAILED: another operation (install/upgrade/rollback) is in progress

At the same time a helm list -n myns the chart disappeared and didn't show up in the list at all

This is a chart that's been upgraded over 800 times successfully, only change was the helm version bump, chart failed twice in an attempt to deploy with the command:

```
helm upgrade --install --namespace myns --timeout 1800s --atomic mychart charts/app/standalone --values values-override.yaml
```

Once I rolled back to 3.3 I was able to upgrade the chart successfully.

Output of `helm version`: 

```
version.BuildInfo{Version:"v3.4.0", GitCommit:"7090a89efc8a18f3d8178bf47d2462450349a004", GitTreeState:"dirty", GoVersion:"go1.15.3"}
```
Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.13", GitCommit:"39a145ca3413079bcb9c80846488786fed5fe1cb", GitTreeState:"clean", BuildDate:"2020-07-15T16:18:19Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"16+", GitVersion:"v1.16.13-gke.401", GitCommit:"eb94c181eea5290e9da1238db02cfef263542f5f", GitTreeState:"clean", BuildDate:"2020-09-09T00:57:35Z", GoVersion:"go1.13.9b4", Compiler:"gc", Platform:"linux/amd64"}
```
Cloud Provider/Platform (AKS, GKE, Minikube etc.): GKE

bacongobbler:
Hi @jlcrow.

We ask to please provide a test case or some way for the maintainers to test and reproduce the error you are experiencing. This way we can better help understand the issue and determine if there's a fix we can apply to the next release.

Thanks.

mattfarina:
@jlcrow Yikes. While I wouldn't ask you to do this in production could you do an upgrade while using the `--debug` and `-v 6` flags set. This will provide more diverse output that can help us find the problem.

jlcrow:
[...]

PR #9180, opened 2020-12-30
Handle SIGTERMs during helm upgrade and helm install

Moser-ss:
This PR is open to try to fix issue #4558 and issue #8987. By handling the SIGTERM and throw an error that can be handled internally my intent is to mark the release as `FAILED` when the helm was interrupted or terminated 
I still have some TODOS:
- [X] Implement the same logic to the install action
- [X] In case the flag atomic is passed do a rollback
- [x] Implement/modify tests 
I would like to have an opinion from the maintainers to understand if this approach makes sense from your point of view
I tested in my machine ( a MacOS) and when the Ctrl +C sequence is hit the release is marked as failed

**If applicable**:


jdolitsky (changes requested):
Hello. This appears to be a useful change, but definitely introduces lots of new code with channels etc. that might cause unexpected issues.

If there any way to test these with unit tests? By inserting SIGTERM on a timer, for example?

Moser-ss:
Thanks for the suggestion.
I was thinking about how to add unit tests to this flow, I was hoping that could get some suggestions from the mantainers. I will look for the approach by using timer and SIGTERM

bacongobbler:
How do you suggest we handle cases where an upgrade is midway through completion and CTRL+C is invoked? If I understand this correctly, any hooks currently being executed (e.g. pre-upgrade) are not interrupted. There are a few scenarios I could perceive where an upgrade is cancelled halfway but the pre-upgrade hooks install/upgrade certain resources, and a second call to `helm upgrade` would fail as the previous upgrade did not complete. By introducing this behaviour, you could effectively enter a deadlock. It is also possible that providing the `--atomic` flag with this behaviour may enter a deadlock as well.

Have you tested for these scenarios? What was the outcome?

PhilThurston:
@bacongobbler from my understanding CTRL+C sends the SIGINT signal so simply having different logic for SIGTERM vs SIGINT would work here.

bacongobbler:
Fair enough. Point still stands, though just replace a CTRL+C with a hook that runs past the given timeout threshold. The same problem cases and issues are present in a similar case... Just a different event triggering the stoppage.

bacongobbler:
Here's the case I'm trying to clarify here:

1. I have a Helm chart.
2. I introduce a hook that accidentally runs for 30 days instead of 30 minutes.
3. I call `helm upgrade --timeout 31m --wait ...`
4. (as per this PR) My release enters the FAILED state.
5. I correct my mistake, changing the hook's execution time to 30 minutes.
6. I call `helm upgrade --timeout 31m --wait`

Isn't this going to cause a deadlock? We haven't "cleaned up" or rolled back any of the changes that occurred during the failed release. We just marked it as FAILED.

Moser-ss:
> How do you suggest we handle cases where an upgrade is midway through completion and CTRL+C is invoked?

The flows I have in my mind are the following:
In the case of an upgrade **without** `--atomic` flag just mark the release as failed with a message that it was marked as failed because helm received a SIGTERM or SIGINT, in this way it will not leave the release as pending 
```
 ➜  helm git:(feature-handle-SIGINT)  ./bin/helm upgrade -i issue /Users/stephanemoser/tmp/issue-4558 --wait
^CError: UPGRADE FAILED: SIGTERM or SIGINT received, release failed
```
```
REVISION        UPDATED                         STATUS          CHART                   APP VERSION     DESCRIPTION

1               Mon Jan 25 23:11:09 2021        deployed        helm-to-fail-0.2.0                      Install complete

2               Mon Jan 25 23:13:19 2021        failed          helm-to-fail-0.2.0                      Upgrade "issue" failed: SIGTERM or SIGINT received, release failed
```
In case if an upgrade with the flag `--atomic`, it will rollback to the previous release using the same logic in case a release normally fail 
```
 ➜  helm git:(feature-handle-SIGINT)  ./bin/helm upgrade -i issue /Users/stephanemoser/tmp/issue-4558 --atomic
^CError: UPGRADE FAILED: release issue failed, and has been rolled back due to atomic being set: SIGTERM or SIGINT received, release failed
```
```
3               Mon Jan 25 23:17:20 2021        superseded      helm-to-fail-0.2.0                      Upgrade complete
4               Mon Jan 25 23:20:42 2021        failed          helm-to-fail-0.2.0                      Upgrade "issue" failed: SIGTERM or SIGINT received, release failed
5               Mon Jan 25 23:20:52 2021        deployed        helm-to-fail-0.2.0                      Rollback to 3
```

Moser-ss:
> Here's the case I'm trying to clarify here:
> 
> 1. I have a Helm chart.
> 2. I introduce a hook that accidentally runs for 30 days instead of 30 minutes.
> 3. I call `helm upgrade --timeout 31m --wait ...`
> 4. (as per this PR) My release enters the FAILED state.
> 5. I correct my mistake, changing the hook's execution time to 30 minutes.
> 6. I call `helm upgrade --timeout 31m --wait`

I tried to reproduce this behavior using this example https://github.com/helm/helm/issues/4558#issuecomment-417126172
The first time, it fails with the timeout, then I fix the job to be able to exit and it was capable to finish the release with success

bacongobbler:
The reason I'm asking is because this appears to be the only case where Helm can enter the PENDING_UPGRADE state during an upgrade.

If the intention is to return to the FAILED state, instead of introducing all this timeout/capture logic (which can be prone to deadlocks or unintended breaking behaviour), why not just enter the FAILED state until the upgrade succeeds?

Moser-ss:
> why not just enter the FAILED state until the upgrade succeeds?

I think I am not the right person to answer this question; I was waiting for some guidance from the maintainers about this topic.

From Helm User's experience, I notice that when an upgrade starts, helm defines the state of the release as `PENDING_UPGRADE` until it receives a timeout or a success. 

In version 3.4.0 looks that helm starts to check if there is any release in `PENDING_UPGRADE`  before starting the upgrade. I don't know if that was introduced to avoid concurrent upgrades, but in reality, if some external factor kills the helm execution, we are lock-in that `PENDING_UPGRADE state.

 So my initial idea was to handle the SIGINT and SIGTERM to mark the release a `FAILED`, but then I found if I set the flag atomic, it will not revert to the previous successfully  release, for that reason, I introduced the mutex to lock to flow, and the release is rollbacked
 
 But I am open to suggestions of which will be the best approach to tackle the problem of releases been lock when helm is killed or timeout in a pre-hook

Moser-ss:
@bacongobbler / @jdolitsky Anything I should improve in this PR? Or should we have a different approach for this issue?

bacongobbler:
Sorry for not being totally clear. I'm asking you to research whether it would be more feasible to replace any reference to the PENDING_UPGRADE state with the FAILED state instead. I do not know which is the right approach, but if my intuition is right, moving forward with that behavior would be much simpler to maintain as you would not have to introduce all this control/timing logic in the first place.

Moser-ss:
Well, we have this PENDING_UPGRADE state to avoid corruption of the storage, or at least is what I understand from this [PR](https://github.com/helm/helm/pull/7322) that created the lock around the PENDIND_UPGRADE state.
If we accept that the only solution for releases that are blocked in the PENDING_X state is to do a rollback I can open PR to improve the documentation.

hiddeco (commented):
With an eye on projects that make use of `pkg/action` as a "Helm SDK" (without CLI bits), and that may have their own way of dealing with locks.

I have some concerns about the introduction of the `sync.Mutex`, channels, etc. directly into the package, instead of them e.g. being injected from higher up.

kzap:
if UPGRADE_PENDING is used as a lock so 2 upgrades don't happen at the same time and corrupt storage. Why not have a new state WAITING_STATUS or something that happens after the upgrade is complete but we are `--wait`

Then a new deployment can still proceed as long as its not possible for the corruption of storage during the actual upgrade process of helm applying all the manifests.

Moser-ss:
> I have some concerns about the introduction of the `sync.Mutex`, channels, etc. directly into the package, instead of them e.g. being injected from higher up.

Any suggestion when the Helm CLI should start listing and handling OS signals?

hiddeco:
> Any suggestion when the Helm CLI should start listing and handling OS signals?

I am not an expert on Helm's architecture, but have a pretty good idea on how things are wired together due to having read most of the code.

My first concern about the `sync.Mutex` could be solved by moving it from the global `var` to the `Install` structure. By doing this, the lock only applies to a single release, and no longer gets in the way of SDK consumers that e.g. perform concurrent operations on different releases.

The entry point for starting to listen to OS signals is in my opinion https://github.com/helm/helm/tree/master/cmd/helm.

Moser-ss:
Update in this PR:

- The branch was rebased with the most recent main branch
- The `sync.Mutex` were moved to inside of `action/install` and `action/upgrade` struct so every action will have its own lock.
- Tests were added.

So I ask you (@hiddeco / @jdolitsky / @bacongobbler ) if you could have a second look at this PR.

If this solution still brings problems in your POV, just tell me; I will have more time to work on this PR in the next weeks.

In the worst scenario we could have a disclaimer in the documentations that helm doesn't helm interruptions and to recover the user or machine needs to execute an rollback command

pkg/kube/fake/printer.go
```diff
@@ -47,7 +48,11 @@ func (p *PrintingKubeClient) Create(resources kube.ResourceList) (*kube.Result,
 	return &kube.Result{Created: resources}, nil
 }
 
-func (p *PrintingKubeClient) Wait(resources kube.ResourceList, _ time.Duration) error {
+func (p *PrintingKubeClient) Wait(resources kube.ResourceList, d time.Duration) error {
+	if p.WaitDuration != 0 {
+		time.Sleep(p.WaitDuration)
```
bacongobbler, line 53: I think if I recall correctly, the time.Duration was meant more as a deadline (wait for X seconds) instead of calling `.Sleep`. So you'd set a `context.WithDeadline` and return an error if the deadline was reached.

https://golang.org/pkg/context/#WithDeadline

By introducing this, users who previously introduced a wait time duration may now have their libraries/projects sleep for the described duration, and I don't think that's how this field was intended to be used.

pkg/kube/fake/printer.go
```diff
@@ -30,7 +30,8 @@ import (
 // PrintingKubeClient implements KubeClient, but simply prints the reader to
 // the given output.
 type PrintingKubeClient struct {
-	Out io.Writer
+	Out          io.Writer
+	WaitDuration time.Duration
```
bacongobbler, line 34: Why was this introduced instead of using the time.Duration in the function signature for `Wait`? What's the expected behaviour of the following code:

```
p := PrintingKubeClient{ WaitDuration: 10*time.Second, Out: ioutil.Discard }
p.Wait(resources, 10*time.Second)
```

pkg/action/upgrade_test.go
```diff
@@ -18,7 +18,11 @@ package action
 
 import (
 	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
 	"testing"
+	gotime "time"
```
bacongobbler, line 25: generally speaking standard libraries are referenced by their usual name, so `time` should not be referenced as `gotime` - kinda redundant when the code's already written in Go anyways.

If there's a third-party library attempting to munging with a similar namespace as a stdlib, then we provide an alias for the third-party library, but the standard lib remains the same.

For example `time` and `helmtime` as referenced here: https://github.com/helm/helm/blob/21251378b432fb629cb1a49af3892e20c1ec9624/pkg/action/hooks.go#L21-L26

bacongobbler (changes requested):

pkg/kube/fake/printer.go
Moser-ss: This was introduced to fake a waiting period. In my case, it is to make the process of installing or upgrading wait, so the other part of the test sends the SIGTERM.

```
p := PrintingKubeClient{ WaitDuration: 10*time.Second, Out: ioutil.Discard }
p.Wait(resources, 10*time.Second)
```
The expected behavior is to wait 10 secs to received the response.

pkg/kube/fake/printer.go
Moser-ss: From what I understood of the PrintingKubeClient's code, the time.Duration` doesn't do anything it is just a placeholder to make this function have the same signature that the real function of KubeClient

pkg/kube/fake/printer.go
Moser-ss: After triple-checking the code, I decided to retract my changes in the PrintingKubeClient.
Instead the logic to simulate a delay in the wait for resources will be in the FakeKubeClient

pkg/action/upgrade_test.go
Moser-ss: Thanks for pointing out this, I already change all the entries I introduced

Moser-ss:
@bacongobbler I finish my changes according to your comments

pkg/kube/fake/fake.go
```diff
@@ -50,8 +51,11 @@ func (f *FailingKubeClient) Create(resources kube.ResourceList) (*kube.Result, e
 	return f.PrintingKubeClient.Create(resources)
 }
 
-// Wait returns the configured error if set or prints
+// Waits the amount of time defined on f.WaitDuration, then returns the configured error if set or prints.
 func (f *FailingKubeClient) Wait(resources kube.ResourceList, d time.Duration) error {
+	if f.WaitDuration != 0 {
```
bacongobbler, line 56: `time.Sleep(time.Duration(0 * time.Second))` is a no-op so this `if` check is redundant.

From https://pkg.go.dev/time#Sleep:

> A negative or zero duration causes Sleep to return immediately.

bacongobbler (changes requested):

pkg/action/upgrade.go
```diff
@@ -306,11 +317,44 @@ func (u *Upgrade) performUpgrade(originalRelease, upgradedRelease *release.Relea
...
+	u.Lock.Lock()
+	if err != nil {
+		rel, err = u.failRelease(rel, created, err)
+	}
+	c <- resultMessage{r: rel, e: err}
+	u.Lock.Unlock()
+}
+
+// Setup listener for SIGINT and SIGTERM
+func (u *Upgrade) handleSignals(c chan<- resultMessage, upgradedRelease *release.Release) {
```
bacongobbler, line 341: Because pkg/action is a public API, this could be construed as a breaking change as this introduces new behaviour to interact with the API.

There could be cases where a program uses Helm's go libraries to perform upgrades in a CI/CD pipeline, and they do not want the upgrade to be interrupted when a SIGTERM/SIGINT are captured.

This should either be part of `cmd/helm` or added as an opt-in flag.

pkg/action/install_test.go
```diff
@@ -361,7 +363,62 @@ func TestInstallRelease_Wait(t *testing.T) {
...
+			res, err := instAction.Run(buildChart(), vals)
+			is.Error(err)
+			is.Contains(res.Info.Description, "SIGTERM or SIGINT received, release failed")
+			is.Equal(res.Info.Status, release.StatusFailed)
+		})
+		return
+
+	}
+	t.Run("Setup TestInstallRelease_Wait_Interrupted", func(t *testing.T) {
+		cmd := exec.Command(os.Args[0], "-test.run=TestInstallRelease_Wait_Interrupted")
```
bacongobbler, line 387: Why is the command re-executing itself? Why can't we run a function in the background using a `context` to test certain cancellation triggers?

https://pkg.go.dev/context

You can have `handleSignals()` return a `CancelFunc`, which can be provided to `Upgrade` as a `context.Context`. Then in test code you can either test `handleSignals` or provide an alternative context, like `context.WithDeadline()`.

pkg/action/install_test.go
```diff
@@ -361,7 +363,62 @@ func TestInstallRelease_Wait(t *testing.T) {
 	is.Contains(res.Info.Description, "I timed out")
 	is.Equal(res.Info.Status, release.StatusFailed)
 }
+func TestInstallRelease_Wait_Interrupted(t *testing.T) {
+	if os.Getenv("HANDLE_SIGINT") == "1" {
```
bacongobbler, line 367: What if the user had already set HANDLE_SIGINT for some other program's execution?

bacongobbler (changes requested):

bacongobbler:
looking good! provided some feedback to help push things along. Let me know when you're finished with those changes and I will schedule another round of reviews.

pkg/action/upgrade.go
bacongobbler: One point of note: contexts should not be stored in a structure. They should be passed in the function signature.

Unfortunately that means we may have to introduce a new `UpgradeWithContext` function to accommodate contexts.

From https://pkg.go.dev/context

> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx:

hiddeco:
Sorry about the late reply after getting pinged, I was on vacation and very busy the weeks prior to that.

The mutex on the action in combination with the latest comments from Matthew around the signal handler would address all my previous concerns :+1:

pkg/action/upgrade.go
Moser-ss: Thanks for the tip, because I am not well knowledgeable about the context package I will start to dig in some docs and explanations, and in all the flow I try to add the context in the helm/cmd and see what we can do

pkg/action/install_test.go
Moser-ss: With the usage of context.Context this issue was removed

pkg/action/install_test.go
Moser-ss: With the usage of context.Context this issue was removed

Moser-ss:
The logic to handle signals was moved from pkg/action to cmd/helm
A new function was created to Install and Upgrade types `RunWithContext`

I think I am duplicating code in  [install](https://github.com/Moser-ss/helm/blob/c62ce12bed39ded7ccaa2a09de3b3b351bf2d4b7/cmd/helm/install.go#L248-L259) and [upgrade](https://github.com/Moser-ss/helm/blob/c62ce12bed39ded7ccaa2a09de3b3b351bf2d4b7/cmd/helm/upgrade.go#L181-L192) but I didn't find a good place to add a function to set up the context and handle the signals

The tests were rewritten and are cleaner now that we are using context

I don't add any customization to the chart description, basically,  it will receive the error message from the context

@bacongobbler / @jdolitsky the PR is ready for a new review

LuckySB:
And what about cases when the deployment process is interrupted not by a SIGTERM, but due to network loss

cmd/helm/install.go
```diff
@@ -239,7 +244,21 @@ func runInstall(args []string, client *action.Install, valueOpts *values.Options
...
+	// Create context and prepare the handle of SIGTERM
+	ctx := context.Background()
+	ctx, cancel := context.WithCancel(ctx)
+
+	// Handle SIGTERM
+	cSignal := make(chan os.Signal)
+	signal.Notify(cSignal, os.Interrupt, syscall.SIGTERM)
+	go func() {
+		<-cSignal
+		fmt.Fprintf(out, "Release %s has been cancel.\n", args[0])
```
bacongobbler, line 257: typo: cancelled*

cmd/helm/upgrade.go
```diff
@@ -174,7 +178,20 @@ func newUpgradeCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
...
+			// Create context and prepare the handle of SIGTERM
+			ctx := context.Background()
+			ctx, cancel := context.WithCancel(ctx)
+
+			// Handle SIGTERM
+			cSignal := make(chan os.Signal)
+			signal.Notify(cSignal, os.Interrupt, syscall.SIGTERM)
+			go func() {
+				<-cSignal
+				fmt.Fprintf(out, "Release %s has been cancel.\n", args[0])
```
bacongobbler, line 190: typo: cancelled*

bacongobbler (changes requested):

bacongobbler:
This looks much better. Thanks for taking the time to refactor

bacongobbler:
> And what about cases when the deployment process is interrupted not by a SIGTERM, but due to network loss

We could potentially tie into the `context.Context` by adding a `WithTimeout`. But that's out of this PR's current scope.

cmd/helm/upgrade.go
Moser-ss: Resolved in commit https://github.com/helm/helm/pull/9180/commits/4bc901c95f3589b6d6b4b3bc0843347334cd0f68

cmd/helm/install.go
Moser-ss: Resolved in commit https://github.com/helm/helm/pull/9180/commits/4bc901c95f3589b6d6b4b3bc0843347334cd0f68

pkg/kube/fake/fake.go
Moser-ss: Resolved in commit https://github.com/helm/helm/pull/9180/commits/4bc901c95f3589b6d6b4b3bc0843347334cd0f68

bridgetkromhout:
Discussed in today's community call; an additional reviewer is needed due to the scope of this PR. Thanks!

hickeyma (commented):
Thanks for the effort @Moser-ss. The code looks good. Just a few things:

- Due to the changes involved, I would like a few more maintainers to review this in addition to the required 2.
- The PR now sets `failed` state when you interrupt (CTRL-C) an `install` and `upgrade`
- While testing manually I saw output as follows:

```console
$ helm install foo mychart/ --debug

install.go:178: [debug] Original chart version: ""
install.go:195: [debug] CHART PATH: /Users/mhickey/tmp/helm-charts/mychart

^CRelease foo has been cancelled.
client.go:122: [debug] creating 3 resource(s)
Error: context canceled
helm.go:81: [debug] context canceled

$ helm upgrade issue-10058 issue-10058 --debug 

upgrade.go:139: [debug] preparing upgrade for issue-10058
upgrade.go:147: [debug] performing update for issue-10058
upgrade.go:319: [debug] creating upgraded release for issue-10058
client.go:203: [debug] checking 3 resources for changes
client.go:466: [debug] Looks like there are no changes for ServiceAccount "issue-10058"
client.go:466: [debug] Looks like there are no changes for Service "issue-10058"
^CRelease issue-10058 has been cancelled.
upgrade.go:420: [debug] warning: Upgrade "issue-10058" failed: context canceled
client.go:466: [debug] Looks like there are no changes for Deployment "issue-10058"
Error: UPGRADE FAILED: context canceled
helm.go:81: [debug] context canceled
UPGRADE FAILED
main.newUpgradeCmd.func2
	helm.sh/helm/v3/cmd/helm/upgrade.go:196
	github.com/spf13/cobra@v1.1.3/command.go:852
	github.com/spf13/cobra@v1.1.3/command.go:960
	github.com/spf13/cobra@v1.1.3/command.go:897
main.main
	helm.sh/helm/v3/cmd/helm/helm.go:80
runtime.main
	runtime/proc.go:225
runtime.goexit
	runtime/asm_amd64.s:1371
```

Should I have got a stack trace on the `upgrade`?

hickeyma:
@Moser-ss Testing again, it would seem that stack trace will happen depending on when you you send the signal rather than always happening.

hickeyma (approved):
LGTM, thanks for working on this @Moser-ss

mattfarina:
Note, a follow-up pull request could add this to the uninstall command.

mattfarina (approved):
It looks like I get the easy rubber stamp after @bacongobbler and @hickeyma did all the real reviewer work.

Moser-ss:
> Note, a follow-up pull request could add this to the uninstall command.

@mattfarina Do you mean to add the capability to handle SIGTERM to uninstall command? And what will be the behavior?

In the case of install and upgrade commands, we have a wait flag that gives us a bigger opportunity window to press the Ctrl+C and with that interrupt the installation and left the chart in the upgrading state.

I am not sure if I am correct, but in the case of uninstalling command we are not changing the state of the chart, we are just deleting all resources and history without using an intermediate state.

But if you clarify the behaviour you have in mind I can work in that PR

czy006:
Helm Version : 3.7.2
K8S Version: 1.22.5
Hi, if i using --atomic args to upgrade/install , and use command ctrl c (on mac or linux) to interrupt(SIGTERM) and exit it, next time use helm status to check the status can be see status is holding on pending-upgrade/pending-rollback,but it was failed. I think the right status is failed, Because it's not normal anymore.  Is this discussion about solving the problem ？

PR #9181, opened 2021-01-01
fix chart name verification does not meet the specification

### LINKED ISSUE #9179: Helm create accepts uppercase characters
> helm create MyChart

Creates a chart called MyChart. The expected behavior would be failing with an "Uppercase characters in chart name not supported" message.

wawa0210:
fix #9179

from the doc ,https://helm.sh/docs/chart_best_practices/conventions/#chart-names

We know that chartsname needs to meet the following verification rules

```
Chart names must be lower case letters and numbers. Words may be separated with dashes (-):

Examples:

drupal
nginx-lego
aws-cluster-autoscaler
Neither uppercase letters nor underscores can be used in chart names. Dots should not be used in chart names.

The directory that contains a chart MUST have the same name as the chart. Thus, the chart nginx-lego MUST be created in a directory called nginx-lego/. This is not merely a stylistic detail, but a requirement of the Helm Chart format.
```

The current regular match has a bug

**If applicable**:

### DIFF
--- a/pkg/chartutil/create.go
+++ b/pkg/chartutil/create.go
@@ -36,7 +36,7 @@ import (
 // This regular expression is probably stricter than it needs to be. We can relax it
 // somewhat. Newline characters, as well as $, quotes, +, parens, and % are known to be
 // problematic.
-var chartName = regexp.MustCompile("^[a-zA-Z0-9._-]+$")
+var chartName = regexp.MustCompile("^[a-z0-9\\-]+$")
 
 const (
 	// ChartfileName is the default Chart file name.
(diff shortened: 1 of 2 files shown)


marckhouzam:
Thanks @wawa0210!

What happens when trying to install a chart that has uppercase letters in its name?  If it works then I worry that preventing it would break backwards compatibility.

BTW I've been using directories named `chart` (which doesn't match the chart name) for years and it works. But I never package them.  Maybe we should clarify the docs about that.

sbose78 (approved):
Thank you for the fix, LGTM!

sbose78:
Ah I see your comment now @marckhouzam , I agree this does have a potential to break backwards compatibility.

wawa0210:
@bacongobbler 

Can you help, take some time to help reivew this pr? Looking forward to your comments

bacongobbler:
We do not have enough consensus to determine if this will break compatibility for some users. Let's hold off on merging this until Helm 4.

In the future, please check to make sure CircleCI tests pass before asking for a review. Thanks.

PR #9184, opened 2021-01-03
bug: Fix Aliased dependencies condition not respecting specified default value

### LINKED ISSUE #9052: Aliased dependencies condition not respecting specified default value
Output of `helm version`:
```
version.BuildInfo{Version:"v3.4.1", GitCommit:"c4e74854886b2efe3321e185578e6db9be0a6e29", GitTreeState:"clean", GoVersion:"go1.15.4"}
```
Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.4", GitCommit:"d360454c9bcd1634cf4cc52d1867af5491dc9c5f", GitTreeState:"archive", BuildDate:"2020-11-15T17:26:49Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"17+", GitVersion:"v1.17.9-eks-4c6976", GitCommit:"4c6976793196d70bc5cd29d56ce5440c9473648e", GitTreeState:"clean", BuildDate:"2020-07-17T18:46:04Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
```
**Issue**
The dependencies of a helm chart fail to respect their `condition` in case the dependency has an `alias`. I've prepared a repo that reproduces the error here: https://github.com/oliviermichaelis/helm-bug-reproduction.
The charts have the following structure:
```
umbrella1:
  parent1:
    dep
  parent1:
    dep
umbrella2:
  parent2:
    dep
  parent2:
    dep

```
The `parent2` chart has the following dependency in the `Chart.yaml` with `condition: depAlias.enabled` :
```
dependencies:
  - name: dep
    alias: depAlias
    version: "0.1.0"
    repository: "file://../dep"
    condition: depAlias.enabled
```
The `values.yaml` disables the dependency by default:
```
depAlias:
  enabled: false
```
However, the `umbrella2` charts subdependency `dep` is rendered, even though it should be disabled:
```
$ helm template test ./umbrella2   
---
# Source: umbrella2/charts/parent2b/charts/dep/templates/service.yaml
apiVersion: v1
kind: Service
metadata:
  name: test-dep
  labels:
    helm.sh/chart: dep-0.1.0
    app.kubernetes.io/name: dep
    app.kubernetes.io/instance: test
spec:
  type: ClusterIP
  ports:
    - port: 80
      targetPort: http
      protocol: TCP
      name: http
  selector:
    app.kubernetes.io/name: dep
    app.kubernetes.io/instance: test
```
Note that only a single resource is rendered, not both. I guess it has something to do with how the hierarchical values are merged.

---
In order to help to better understand what's going on, I have included the `umbrella1` chart without an `alias`. 
The `parent1` chart has the following dependency in the `Chart.yaml` with `condition: dep.enabled` :
```
dependencies:
  - name: dep
    alias: dep
    version: "0.1.0"
    repository: "file://../dep"
    condition: dep.enabled
```
The `values.yaml` disables the dependency by default:
```
dep:
  enabled: false
```
As expected, the `dep` subcharts are not rendered:
```
$ helm template test ./umbrella1

```

Thanks for having a look, I'm happy to help where I can! :)

oliviermichaelis:
This PR fixes #9052.

In case a chart is referenced multiple times as a dependency with a condition, a `false` condition is only respected for the dependency that is processed first. The remaining dependencies whose condition is `false` will get rendered incorrectly.

I found out that the chart dependency`Metadata` is initialized once the dependency`Chart` is loaded (e.g. from the filesystem). This `Metadata` object is then referenced by all aliased depency `Chart`.

Once the first dependency is processed, the `Metadata`s `Name` is changed to the `Alias`:
https://github.com/helm/helm/blob/960b3b53c6c30d72c68738fbed3ccf3724b30130/pkg/chartutil/dependencies.go#L143-L145

For the next dependency, the check in Line 132 will fail since `req.Name` has now changed to the alias:
https://github.com/helm/helm/blob/960b3b53c6c30d72c68738fbed3ccf3724b30130/pkg/chartutil/dependencies.go#L130-L135

The remaining disabled dependencies will therefore not get removed from the list of charts that will be rendered.

I decided to deep-copy the`Metadata`s dependencies in this PR to prevent metadata corruption.

This PR contains a test case for a chart that renders incorrectly with the current `master`, but renders correctly with this PR.
The structure of the chart is:
```
chart-with-dependency-aliased-twice-with-condition:
  childFoo:
    grandchildAlias
  childBar:
    grandchildAlias
```

**If applicable**:


oliviermichaelis:
The CircleCI build did not run, is it possible to retrigger that?

bacongobbler:
We've seen this occur before when other community members fork Helm and point their fork at another CircleCI account IIRC. I ca't fully recall what's the correct course of action here, but looking at the webhook system from GIthub's end it looks like the webhook payloads are being sent to CircleCI just fine on our end.

Can  you please look into this on your end?

oliviermichaelis:
Thanks for the pointer, I have deleted the CircleCI integration on my fork. Can we somehow trigger the build again?

msvechla:
@bacongobbler we also experience this bug, can we get this merged?

FireDrunk:
I've just stumbled across this bug, and this seems fixed in the latest version if this is the issue you were getting at:

```yaml
dependencies:
  # Databases
  - name: postgresql
    version: 12.8.0
    repository: oci://registry-1.docker.io/bitnamicharts
    alias: db-a
    condition: db-a.enabled

  - name: postgresql
    version: 12.8.0
    repository: oci://registry-1.docker.io/bitnamicharts
    alias: db-b
    condition: db-b.enabled
```

This works.

joejulian:
Tested your repro with helm 3.12.2. Still a problem.

joejulian (dismissed):

pderrierw:
Please can we have this merged ? This is a real issue for us

we have a helm chart A with a conditional dependency on redis:
`dependencies:
  - name: redis
    version: 17.3.8
    condition: redis.enabled
`

in the application we reference multiple time the chart A as dependency:
`dependencies:
  - name: A
    alias: application1
  - name: A
    alias: application2
`

with following values.yaml:
`application1:
  redis:
    enabled: false
    
application2:
  redis:
     enabled: true`
     
     
Impossible to get a stable rendering of the final chart, sometimes the redis is rendered, sometimes not, I have the impression it depends on the name of application1 and 2 (so of the ordering of the processing by helm)

That matches the content of this MR, thanks @oliviermichaelis for the fix
Why isn't merged yet?

scottrigby:
The code does work, but not sure that `TestChartWithDependencyAliasedTwiceWithCondition()` is testing this fully. Here's what a manual test gives:

1. first, we need to update the test chart included in this PR to:
    ```diff
    $ git diff pkg/chartutil/testdata/chart-with-dependency-aliased-twice-with-condition/charts/child/charts/grandchild/templates/service.yaml
    ...
    @@ -1,7 +1,7 @@
     apiVersion: v1
     kind: Service
     metadata:
    -  name: {{ include "grandchild.fullname" . }}
    +  name: {{ include "grandchild.fullname" . | lower }}
       labels:
         {{- include "grandchild.labels" . | nindent 4 }}
     spec:
    ```

    otherwise it will fail with:
    ```console
    $ helm install test pkg/chartutil/testdata/chart-with-dependency-aliased-twice-with-condition \
      --set childFoo.grandchildAlias.enabled=true
    Error: INSTALLATION FAILED: 1 error occurred:
	    * Service "test-grandchildAlias" is invalid: metadata.name: Invalid value: "test-grandchildAlias": a DNS-1035 label must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?')
    ```
    because while Helm has `ValidateReleaseName()` in pkg/chartutil/validate_name.go`, the dependency alias when passed to `.Chart.Name` doesn't validate on install/upgrade, so (for now) we can only test by converting to lower in the template before attempting to create a service with an uppercase character ("grandchildAlias").

2. now we can reproduce the error:

    ```console
    $ helm install test pkg/chartutil/testdata/chart-with-dependency-aliased-twice-with-condition \
      --set childFoo.grandchildAlias.enabled=true
      NAME: test
      STATUS: deployed
      ...
    
    $ k get all
    NAME                            TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)   AGE
    service/test-grandchild        ClusterIP   10.96.7.47      <none>        80/TCP    12s
    service/test-grandchildalias   ClusterIP   10.96.162.130   <none>        80/TCP    12s
    ```
    I see the error is there is one dependency chart's template installed—using the alias. And another installed using the dependency name—not the alias. Only the first one should have been created.
    
    If you change the order, and only enable the 2nd dependency in the list `--set childBar.grandchildAlias.enabled=true` instead, you get only the dependency chart's template installed using the dependency name—not the alias:
    ```console
    $ helm uninstall test
    $ helm install test pkg/chartutil/testdata/chart-with-dependency-aliased-twice-with-condition \
      --set childBar.grandchildAlias.enabled=true
    NAME                      TYPE        CLUSTER-IP     EXTERNAL-IP   PORT(S)   AGE
    service/test-grandchild   ClusterIP   10.96.190.22   <none>        80/TCP    3s
    ```
    
3. Test this PR. Seems to fix it:
    ```console
    $ helm uninstall test
    $ gh pr checkout 9184
    $ make build
    # test first case above
    $ ./bin/helm install test pkg/chartutil/testdata/chart-with-dependency-aliased-twice-with-condition \
      --set childFoo.grandchildAlias.enabled=true
      ...
      
    $ k get all
    NAME                           TYPE        CLUSTER-IP     EXTERNAL-IP   PORT(S)   AGE
    service/test-grandchildalias   ClusterIP   10.96.73.213   <none>        80/TCP    2s
    
    # test second case above
    $ ./bin/helm install test pkg/chartutil/testdata/chart-with-dependency-aliased-twice-with-condition \
      --set childBar.grandchildAlias.enabled=true
    $ k get all
    NAME                           TYPE        CLUSTER-IP     EXTERNAL-IP   PORT(S)   AGE
    service/test-grandchildalias   ClusterIP   10.96.57.100   <none>        80/TCP    3s
    ```
    
4. Fun note, if you want to test enabling both dependencies, you'll need to do a name override on one of the grandchildren charts, otherwise the service will fail creation because the same name will be used and `service/test-grandchildalias` already exists. Like this:
    ```console
    $ helm install pkg/chartutil/testdata/chart-with-dependency-aliased-twice-with-condition \
      --set childFoo.grandchildAlias.enabled=true \
      --set childBar.grandchildAlias.enabled=true \
      --set childBar.grandchildAlias.nameOverride=bar-grandchild
    $ k get all
                  TYPE        CLUSTER-IP    EXTERNAL-IP   PORT(S)   AGE
    service/test-bar-grandchild    ClusterIP   10.96.23.11   <none>        80/TCP    2s
    service/test-grandchildalias   ClusterIP   10.96.86.83   <none>        80/TCP    2s
    ```

So good work on the code change. Correct me if I'm wrong about the automated test—not sure it covers this fully.

oliviermichaelis:
Thanks for your effort @scottrigby, much appreciated! I incorporated your feedback about the invalid resource name in a1b5d6f0222247589dfdfa6c31257a9ab9b856e2.

>Correct me if I'm wrong about the automated test—not sure it covers this fully.

Can you be more specific about this statement, please? What exactly is missing? This is the first time I'm looking at my code from a couple of years ago and I'm certainly missing something. Thanks!

scottrigby:
> Thanks for your effort @scottrigby, much appreciated! I incorporated your feedback about the invalid resource name in [ffd5d61](https://github.com/helm/helm/commit/ffd5d6106f88aa93dbe0705be6ef6f18550532f8).
> 
> > Correct me if I'm wrong about the automated test—not sure it covers this fully.
> 
> Can you be more specific about this statement, please? What exactly is missing? This is the first time I'm looking at my code from a couple of years ago and I'm certainly missing something. Thanks!

What I meant was that if the unit test didn't catch the template bug I found, then why are we adding templating in the test chart.
You're right that for this single use case/bugfix, the test covers it ok. We are processing only the dependencies.
Given that, can you please simplify the testdata chart to have only the bare minimum needed for the test itself? It doesn't need boilerplate comments, values, or even templating. That will make it more clear what the test chart is for when paired with the unit test. And then I think this will be good to go.

oliviermichaelis:
@scottrigby I rebased on the latest `main` and addressed your comments by:
- Renamed/shortened testdata chart name
- Removed boilerplate comments, values, and templating to the absolute minimum. This should make it easier to understand what's actually going on in the test

scottrigby (approved):
@oliviermichaelis hi again—I was able to review your changes to this PR since you made them in February, and they look good. Are you able to fix the conflicts since changes were made in main? LGTM once that's complete

hobti01:
Can confirm that this change fixes the issue when applied to the 3.17 codebase.  We're very much looking forward to seeing this long-standing issue squashed.

oliviermichaelis:
In the meantime #9175 for v4 got merged and there is also #30802 to backport the fix to v3. As far as I can tell, they both address the same problem as this PR.
Since a definitive solution seems close,I would consider to close my PR in favor of the previously mentioned PRs.

PR #9189, opened 2021-01-05
Bumping kubernetes to 1.20.1

mattfarina:


davidkarlsen (approved):

technosophos (approved):

go.sum
```diff
@@ -1177,19 +1177,34 @@ honnef.co/go/tools v0.0.1-2019.2.3/go.mod h1:a3bituU0lyd329TUQxRnasdCoJDkEUEAqEt
 honnef.co/go/tools v0.0.1-2020.1.3/go.mod h1:X/FiERA/W4tHapMX5mGpAtMSVEeEUOyHaw9vFzvIQ3k=
 k8s.io/api v0.20.0 h1:WwrYoZNM1W1aQEbyl8HNG+oWGzLpZQBlcerS9BQw9yI=
 k8s.io/api v0.20.0/go.mod h1:HyLC5l5eoS/ygQYl1BXBgFzWNlkHiAuyNAbevIn+FKg=
```
adamreese, line 1179: why does it add new dependancies rather than replace them for a version bump?

adamreese (approved):

go.sum
mattfarina: This is the sum file. This file was added when Russ realized that git tags were mutable and they needed a way to make sure the tag used didn't change. So, they record a hash for each version. It was originally meant to be append only so that all old hashes would be in it. Forever. I discussed the size it would grow to with example large repos and he was initially ok with that. At some point some level of pruning came in to play. The intent is still the same, though.

PR #9192, opened 2021-01-05
fix(Makefile): rebuild the binary if go.mod has changed

adamreese:
Rebuild the binary when go.mod or go.sum has changed


technosophos (approved):
❤️

PR #9193, opened 2021-01-06
Adding apiserver to mod/sum

mattfarina:
This is a follow-up to #8608. k8s.io/apiserver was added but
not added to the go.mod file. Go handled the situation cleanly
but left a dirty git tree.


technosophos (approved):
LGTM

PR #9208, opened 2021-01-08
skip if object from older release exists

### LINKED ISSUE #9207: --atomic --cleanup-on-fail with leftover objects from original release leads to rollback failure
Output of `helm version`: 
`
version.BuildInfo{Version:"v3.2.4", GitCommit:"0ad800ef43d3b826f31a5ad8dfbb4fe05d143688", GitTreeState:"clean", GoVersion:"go1.13.12"}`

Output of `kubectl version`:
`
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.2", GitCommit:"59603c6e503c87169aea6106f57b9f242f64df89", GitTreeState:"clean", BuildDate:"2020-01-18T23:30:10Z", GoVersion:"go1.13.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"16+", GitVersion:"v1.16.15-gke.4300", GitCommit:"7ed5ddc0e67cb68296994f0b754cec45450d6a64", GitTreeState:"clean", BuildDate:"2020-10-28T09:23:22Z", GoVersion:"go1.13.15b4", Compiler:"gc", Platform:"linux/amd64"}`

Take any helm chart (I tried it on [helm guestbook](https://github.com/argoproj/argocd-example-apps/tree/master/helm-guestbook)). Follow these steps
1. `helm install guestbook helm-guestbook`
2. Add `nameOverride: foo_bar` in the values.yaml
3. Run `helm upgrade --atomic --cleanup-on-fail guestbook helm-guestbook`
4. Upgrade fails and throws the following error
`Error: UPGRADE FAILED: an error occurred while rolling back the release. original upgrade error: failed to create resource: Service "guestbook-foo_bar" is invalid: metadata.name: Invalid value: "guestbook-foo_bar": a DNS-1035 label must consist of lower case alphanumeric characters or '-', start with an alphabetic character, and end with an alphanumeric character (e.g. 'my-name',  or 'abc-123', regex used for validation is '[a-z]([-a-z0-9]*[a-z0-9])?'): no Service with the name "guestbook-helm-guestbook" found`

To be specific, `no Service with the name "guestbook-helm-guestbook" found` is the error thrown when rollback is attempted. This is because, when the upgrade fails, resources from the older release were still present and when helm attempts to rollback, it comes to this piece of code
https://github.com/helm/helm/blob/5d08a0d00e4f0d31d6db55e71bccbcd665366c9a/pkg/kube/client.go#L215

Helm expects that the objects from the older release would be present in the failed release but they aren't which leads to the above error. We could modify the code slightly to `continue` if it comes across a scenario where the objects from the older release are present but are not a part of the failed release. Helm can safely ignore them. However, since this change would be in a generic piece of code which is also used for upgrade, I don't have the complete idea of its impact on other pieces of code. 

Another thing we could do is, add some documentation around this scenario so the user is aware of what's happening in case they run into the above issue because the error is not very helpful (since the object from the old release exists)

vadasambar:
- this is to fix #9207
Closes #9207

**If applicable**:

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -225,8 +225,8 @@ func (c *Client) Update(original, target ResourceList, force bool) (*Result, err
 
 		originalInfo := original.Get(info)
 		if originalInfo == nil {
-			kind := info.Mapping.GroupVersionKind.Kind
-			return errors.Errorf("no %s with the name %q found", kind, info.Name)
+			c.Log("obj %q already exists; skipping", info.Name)
+			return nil
 		}
 
 		if err := updateResource(c, info, originalInfo.Object, force); err != nil {


bacongobbler:
Hi @vadasambar. We ask that any pull requests provide unit tests to ensure that backwards behaviour is not unintentionally introduced.

As I understand it, this change could cause unintentional changes. The Kubernetes API could return a `nil` `originalInfo` object when the object does not already exist in the cluster, hence why the original error message was "no Secret with the name foo found". You are now assuming that a `nil` originalInfo object means that the resource DOES exist, which is in direct conflict with the original behaviour.

I also asked in https://github.com/helm/helm/issues/9207#issuecomment-762987234 that you please test with a more recent version of Helm before submitting a patch to verify this is still an issue with more recent versions of Helm. Have you done that yet?

vadasambar:
Hey @bacongobbler , sorry for the late response. Let me see if I can add unit tests for the change. 
>You have not clarified why this PR would fix the issue you originally described either. A bit more clarification/reasoning behind the change would be welcome.

I had explained why we needed this change in the issue description (https://github.com/helm/helm/issues/9207). Please let me know if you need any further clarifications. Would be happy to comply! 

>The Kubernetes API returns a nil originalInfo object when the object does not already exist in the cluster, hence why the original error message was "no Secret with the name foo found". You are now assuming that a nil originalInfo object means that the resource DOES exist, which is in direct conflict with the original behaviour. Can you clarify this behaviour?

I have explained this in detail in the issue description (please let me know if I should clarify more) but in a nutshell, 
When we go from (Release A works, Release B doesn't)
Release A -> Release B, upgrade fails (but the resources from Release A are still present) and helm attempts a rollback Release B -> Release A
The way our code for `Update` works right now when rolling back from Release B -> Release A (as per my understanding)
1. Check if the resource in Release A is present in the cluster
  1.1. If it is, check if it is from the Release B
     1.1.1 If it is from Release B, update it
     1.1.2. If it is not from Release B, throw an error and exit
  1.2. If it is not there in the cluster, create the resource
  
 1.1.2 is the problem here because in this particular case, resources from Release A are still present and hence 1.1 succeeds and goes to 1.1.2 which leads to this issue.

bacongobbler:
Closing.

PR #9211, opened 2021-01-08
Allow uninstall to process ALL manifests (including hooks)

### LINKED ISSUE #9206: Manifests with hooks not getting deleted during helm uninstall
Output of `helm version`: 
```
version.BuildInfo{Version:"v3.4.2", GitCommit:"23dd3af5e19a02d4f4baa5b2f242645a1a3af629", GitTreeState:"clean", GoVersion:"go1.14.13"}
```

Output of `kubectl version`: 
```
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.6", GitCommit:"96fac5cd13a5dc064f7d9f4f23030a6aeface6cc", GitTreeState:"clean", BuildDate:"2019-08-19T11:13:49Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"windows/amd64"}
Server Version: version.Info{Major:"1", Minor:"11+", GitVersion:"v1.11.0+d4cacc0", GitCommit:"d4cacc0", GitTreeState:"clean", BuildDate:"2020-04-23T19:49:32Z", GoVersion:"go1.10.8", Compiler:"gc", Platform:"linux/amd64"}
```

Cloud Provider/Platform (AKS, GKE, Minikube etc.): Openshift

I'm trying to facilitate the ability to work with `PodDisruptionBudget` in my chart, and have run into a few things.

1. A `PodDisruptionBudget` in older versions of Kubernetes/Openshift is immutable, and results in the need to do something like https://github.com/kubernetes/kubernetes/issues/45398#issuecomment-478619253:
```yaml
apiVersion: policy/v1beta1
kind: PodDisruptionBudget
metadata:
  name: {{ template "app.name" . }}
  release: {{ .Release.Name }}
  annotations:
    "helm.sh/hook": "post-upgrade, post-install"
    "helm.sh/hook-delete-policy": "before-hook-creation"
spec:
```
2. I've made sure that `"helm.sh/resource-policy": keep` isn't being set
3. Not sure if it's strictly tied to the hooks, but when running `helm uninstall`, the `PodDisruptionBudget` is **NOT** deleted

I feel like I'm missing something obvious about the hooks and how they should be used that's preventing this from deleting successfully, but I can't seem to find any documentation to this effect.

tomcruise81:
Signed-off-by: Sean Dukehart <tomcruise81@users.noreply.github.com>

closes #9206

**What this PR does / why we need it**: This PR handles cleanup of hooks alongside other manifests as part of calls to `helm uninstall`. 

_Caveat_ - It does **NOT** prevent `pre-delete` or `post-delete` hooks from running, and if a `post-delete` hook exists for a given manifest, what gets created during the `post-delete` hook has the potential to be orphaned.

   - New global methods have been created in lieu of manipulating existing global methods to ensure backwards compatibility
   - It seeks to complete the Helm 3.x work alluded to at https://helm.sh/docs/topics/charts_hooks/#hook-resources-are-not-managed-with-corresponding-releases

### DIFF
--- a/pkg/action/uninstall.go
+++ b/pkg/action/uninstall.go
@@ -174,8 +174,8 @@ func (u *Uninstall) deleteRelease(rel *release.Release) (string, []error) {
 		return rel.Manifest, []error{errors.Wrap(err, "could not get apiVersions from Kubernetes")}
 	}
 
-	manifests := releaseutil.SplitManifests(rel.Manifest)
-	_, files, err := releaseutil.SortManifests(manifests, caps.APIVersions, releaseutil.UninstallOrder)
+	manifests := releaseutil.SplitAllManifests(rel)
+	files, err := releaseutil.SortAllManifests(manifests, caps.APIVersions, releaseutil.UninstallOrder)
 	if err != nil {
 		// We could instead just delete everything in no particular order.
 		// FIXME: One way to delete at this point would be to try a label-based
--- a/pkg/releaseutil/manifest.go
+++ b/pkg/releaseutil/manifest.go
@@ -21,6 +21,8 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+
+	rspb "helm.sh/helm/v3/pkg/release"
 )
 
 // SimpleHead defines what the structure of the head of a manifest file
@@ -35,13 +37,14 @@ type SimpleHead struct {
 
 var sep = regexp.MustCompile("(?:^|\\s*\n)---\\s*")
 
+const manifestTpl = "manifest-%d"
+
 // SplitManifests takes a string of manifest and returns a map contains individual manifests
 func SplitManifests(bigFile string) map[string]string {
 	// Basically, we're quickly splitting a stream of YAML documents into an
 	// array of YAML docs. The file name is just a place holder, but should be
 	// integer-sortable so that manifests get output in the same order as the
 	// input (see `BySplitManifestsOrder`).
-	tpl := "manifest-%d"
 	res := map[string]string{}
 	// Making sure that any extra whitespace in YAML stream doesn't interfere in splitting documents correctly.
 	bigFileTmp := strings.TrimSpace(bigFile)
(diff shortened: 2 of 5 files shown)


pkg/action/uninstall.go
```diff
@@ -174,8 +174,8 @@ func (u *Uninstall) deleteRelease(rel *release.Release) (string, []error) {
 		return rel.Manifest, []error{errors.Wrap(err, "could not get apiVersions from Kubernetes")}
 	}
 
-	manifests := releaseutil.SplitManifests(rel.Manifest)
-	_, files, err := releaseutil.SortManifests(manifests, caps.APIVersions, releaseutil.UninstallOrder)
+	manifests := releaseutil.SplitAllManifests(rel)
```
tomcruise81, line 177: Here's a rather naïve solution.

bacongobbler:
It is intentional behavior for hooks to NOT be removed when `helm uninstall` is invoked. From the [documentation](https://helm.sh/docs/topics/charts_hooks/#hook-resources-are-not-managed-with-corresponding-releases):

> Practically speaking, this means that if you create resources in a hook, you cannot rely upon helm uninstall to remove the resources. To destroy such resources, you need to either add a custom helm.sh/hook-delete-policy annotation to the hook template file, or set the time to live (TTL) field of a Job resource.

Because this is documented, intended behavior, I'm closing this as it directly contradicts with the hook system's described behavior.

tomcruise81:
What about the line that states:

> [Garbage collection of hook resources when the corresponding release is deleted may be added to Helm 3 in the future.](https://helm.sh/docs/topics/charts_hooks/#hook-resources-are-not-managed-with-corresponding-releases:~:text=Garbage%20collection%20of%20hook%20resources%20when,to%20Helm%203%20in%20the%20future) 

That is also documented...

mancdaz:
I agree. The documentation just describes current functionality. It shouldn't be used to reject changes to functionality where useful. And as @tomcruise81 states, it's even documented that this functionality might change in the future (something I've been waiting for someone to do, like @tomcruise81 tried to do!).
I second the need for this cleanup of orphaned hook resources. Maybe with a separate flag (uninstall-all ?) so as not to surprise those who expect 'uninstall' to behave in the current way

bai3shuo4:
@bacongobbler I think it's necessary to re-open this PR. Think about this case, if I use helm client API to design a deployment system. If I set some timeout to redeploy certain package, when some package prehook hang because of some system reasons. I just use helm uninstall method from client package and remove all prehook things. I can't do it

Issue #9203, opened 2021-01-07
Lint: helm/pkg/lint validation rule mismatch k8s validation rule


popozy:
Output of `helm version`:
version.BuildInfo{Version:"v3.4.2", GitCommit:"23dd3af5e19a02d4f4baa5b2f242645a1a3af629", GitTreeState:"clean", GoVersion:"go1.14.13"}

Output of `kubectl version`:
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.0", GitCommit:"e19964183377d0ec2052d1f1fa930c4d7575bd50", GitTreeState:"clean", BuildDate:"2020-08-26T14:30:33Z", GoVersion:"go1.15", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.0", GitCommit:"e19964183377d0ec2052d1f1fa930c4d7575bd50", GitTreeState:"clean", BuildDate:"2020-08-26T14:23:04Z", GoVersion:"go1.15", Compiler:"gc", Platform:"linux/amd64"}

Cloud Provider/Platform (AKS, GKE, Minikube etc.): 
self-building Kubernetes cluster

Reproduce step:
+ get tidb-opeartor.tgz helm package here: http://charts.pingcap.org/tidb-operator-v1.1.5.tgz
+ decompress the archive package
+ helm lint tidb-operato
it shows:
``` shell
==> Linting tidb-operator
[INFO] Chart.yaml: icon is recommended
[ERROR] templates/controller-manager-rbac.yaml: object name does not conform to Kubernetes naming requirements: "test-release:tidb-controller-manager": invalid metadata name, must match regex ^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$ and the length must not be longer than 253
[ERROR] templates/controller-manager-rbac.yaml: the kind "rbac.authorization.k8s.io/v1beta1 ClusterRole" is deprecated in favor of "rbac.authorization.k8s.io/v1 ClusterRole"
[ERROR] templates/controller-manager-rbac.yaml: object name does not conform to Kubernetes naming requirements: "test-release:tidb-controller-manager": invalid metadata name, must match regex ^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$ and the length must not be longer than 253
[...]

PR #9118, opened 2020-12-12
Allow colon in RBAC resource names

jksmth:
Signed-off-by: Jake Smith <jakemgsmith@gmail.com>

**What this PR does / why we need it**:

closes #8835

Allows colons in name of RBAC resources
 - Role
 - ClusterRole
 - RoleBinding
 - ClusterRoleBinding

**Special notes for your reviewer**:

**If applicable**:
- [ ] this PR contains documentation
- [x] this PR contains unit tests
- [x] this PR has been tested for backwards compatibility

PR #9218, opened 2021-01-11
Fix RBAC resource name verification

### LINKED ISSUE #9197: Helm lint fails due to ":" character in .metadata.name
Many default clusterroles and clusterrolebindings contain the character `:` in .metadata.name, for instance `system:coredns`, `system:controller:deployment-controller` etc.  

Helm validation is throwing the following error for object names containing `:` :
```
[ERROR] templates/clusterrolebinding.yaml: object name does not conform to Kubernetes naming requirements: "eks:kube-proxy": invalid metadata name, must match regex ^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$ and the length must not be longer than 253
```
I believe this is a bug as these objects are accepted by the kube-apiserver.
This bug exists from Helm version `3.3.0` on.  
In Helm version 3.2.4 linting for the objects with `:` works fine.

Output of `helm version`:
version.BuildInfo{Version:"v3.4.2", GitCommit:"23dd3af5e19a02d4f4baa5b2f242645a1a3af629", GitTreeState:"dirty", GoVersion:"go1.15.5"}

Output of `kubectl version`:
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.4", GitCommit:"d360454c9bcd1634cf4cc52d1867af5491dc9c5f", GitTreeState:"clean", BuildDate:"2020-11-12T01:09:16Z", GoVersion:"go1.15.4", Compiler:"gc", Platform:"darwin/amd64"}

AWS

wawa0210:
Signed-off-by: wawa0210 <xiaozhang0210@hotmail.com>

fix #9197 and #9203

ref
https://kubernetes.io/docs/reference/access-authn-authz/rbac/#default-roles-and-role-bindings
https://github.com/kubernetes/kubernetes/blob/585fc57cc7aaf73baf6aef84bf6dab65bde0ef9e/pkg/apis/rbac/validation/validation.go#L32

The RBAC resource name verification in the kubernetes rules is different from other resources. kubernetes does not require all the object's metadata name obey the RFC 1123, but helm does not do some processing on this. All are processed in accordance with the RFC 1123 rules. Cause RBAC creation to be unsuccessful.
Therefore, helm needs to match the kubernetes verification rule

**If applicable**:

### DIFF
--- a/pkg/lint/rules/template.go
+++ b/pkg/lint/rules/template.go
@@ -206,11 +206,18 @@ func validateMetadataName(obj *K8sYamlStruct) error {
 	if len(obj.Metadata.Name) == 0 || len(obj.Metadata.Name) > 253 {
 		return fmt.Errorf("object name must be between 0 and 253 characters: %q", obj.Metadata.Name)
 	}
-	// This will return an error if the characters do not abide by the standard OR if the
-	// name is left empty.
-	if err := chartutil.ValidateMetadataName(obj.Metadata.Name); err != nil {
-		return errors.Wrapf(err, "object name does not conform to Kubernetes naming requirements: %q", obj.Metadata.Name)
+	if obj.APIVersion == "rbac.authorization.k8s.io/v1" {
+		if err := chartutil.ValidateRBACMetadataName(obj.Metadata.Name); err != nil {
+			return errors.Wrapf(err, "object name does not conform to Kubernetes naming requirements: %q", obj.Metadata.Name)
+		}
+	} else {
+		// This will return an error if the characters do not abide by the standard OR if the
+		// name is left empty.
+		if err := chartutil.ValidateMetadataName(obj.Metadata.Name); err != nil {
+			return errors.Wrapf(err, "object name does not conform to Kubernetes naming requirements: %q", obj.Metadata.Name)
+		}
 	}
+
 	return nil
 }
 
--- a/pkg/chartutil/validate_name.go
+++ b/pkg/chartutil/validate_name.go
@@ -19,6 +19,7 @@ package chartutil
 import (
 	"fmt"
 	"regexp"
+	"strings"
 
 	"github.com/pkg/errors"
 )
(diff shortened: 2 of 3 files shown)


pkg/lint/rules/template.go
```diff
@@ -206,11 +206,18 @@ func validateMetadataName(obj *K8sYamlStruct) error {
 	if len(obj.Metadata.Name) == 0 || len(obj.Metadata.Name) > 253 {
 		return fmt.Errorf("object name must be between 0 and 253 characters: %q", obj.Metadata.Name)
 	}
-	// This will return an error if the characters do not abide by the standard OR if the
-	// name is left empty.
-	if err := chartutil.ValidateMetadataName(obj.Metadata.Name); err != nil {
-		return errors.Wrapf(err, "object name does not conform to Kubernetes naming requirements: %q", obj.Metadata.Name)
+	if obj.APIVersion == "rbac.authorization.k8s.io/v1" {
```
popozy, line 209: I think the solution to solve the bug by hard coding just rbac kind is not a good idea.
All the object under Kubernetes should be checked and helm [version_skew](https://helm.sh/docs/topics/version_skew/#supported-version-skew) is also another concern which should be token into consideration.

wawa0210: Do you have better suggestions to handle this kind of verification? RBAC is indeed inconsistent with other checks

wawa0210:
@bacongobbler 

Can you help, take some time to help reivew this pr? Looking forward to your comments

pkg/lint/rules/template.go
bacongobbler: Do not check the object type. Instead call the object's validation methods against the Kubernetes object schema. Kubernetes exposes these schema validation libraries as part of [apimachinery](https://pkg.go.dev/k8s.io/apimachinery). [ValidateObjectMeta](https://pkg.go.dev/k8s.io/apimachinery@v0.20.2/pkg/api/validation#ValidateObjectMeta) might be usefu for examplel.

I haven't looked much further, but I would suggest going that route rather than hardcoding APIVersion checks.

bacongobbler (changes requested):

bacongobbler:
conflicts with #9118.

croomes:
Apologies for submitting another PR for this, I tried to implement the suggestions from this PR.  I couldn't find a clean way, but it's perhaps a bit more complete.  https://github.com/helm/helm/pull/9416

PR #9222, opened 2021-01-11
chore(Makefile): add target to generate golden files

adamreese:
Add target to generate golden files used in unit tests


technosophos (approved):
LGTM

PR #9240, opened 2021-01-15
print warning message instead of debug message when ~/.config exists but is not accessible

### LINKED ISSUE #9216: helm exits without comment, when ~/.config exists but is not accessible
If helm does not have appropriate right to access `~/.config` it exits without comment. (I had a previous npm installation, which generated the .config, but only with root access).

Proposal:
`helm.go:65: [debug]` should output in any case, that permission is not granted (not only when debug flag is set). `helm` can obviously not start, if the permission is not granted.

wawa0210:

If helm does not have appropriate right to access helm config file it exits without comment.
In order not to affect compatibility, changing it to a warning message might be a feasible solution.

fix #9216

**If applicable**:

### DIFF
--- a/cmd/helm/helm.go
+++ b/cmd/helm/helm.go
@@ -62,7 +62,7 @@ func main() {
 	actionConfig := new(action.Configuration)
 	cmd, err := newRootCmd(actionConfig, os.Stdout, os.Args[1:])
 	if err != nil {
-		debug("%+v", err)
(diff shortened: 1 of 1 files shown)


wawa0210:
@bacongobbler  Looking forward to your comments

bacongobbler (approved):
LGTM.

PR #9261, opened 2021-01-21
Use kube libraries v0.20.2

sbose78:
* Updated all kube dependencies to v0.20.2
* Azure/go-autorest pinning removed.


sbose78:
Hi! @mattfarina Here's the follow-up to today's developer call.

jdolitsky (approved):
The go.mod/go.sum changes look fine and do not introduce any new modules

bacongobbler (approved):
Thanks.

PR #9318, opened 2021-02-04
Fix issue with helm pull failing if pulling from a repository that redirects to another domain

wahabmk:
This PR fixes issue #9317

**If applicable**:

### DIFF
--- a/pkg/getter/httpgetter_test.go
+++ b/pkg/getter/httpgetter_test.go
@@ -370,7 +370,7 @@ func TestDownloadTLSWithRedirect(t *testing.T) {
 		t.Fatal(errors.Wrap(err, "can't create TLS config for client"))
 	}
 
-	tlsSrv2 := httptest.NewUnstartedServer(http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {
+	tlsSrv2 := httptest.NewUnstartedServer(http.HandlerFunc(func(rw http.ResponseWriter, _ *http.Request) {
 		rw.Header().Set("Content-Type", "text/plain")
 		rw.Write([]byte(srv2Resp))
 	}))
--- a/pkg/getter/httpgetter_test.go
+++ b/pkg/getter/httpgetter_test.go
@@ -358,6 +358,121 @@ func TestDownloadTLS(t *testing.T) {
 	}
 }
 
+func TestDownloadTLSWithRedirect(t *testing.T) {
+	cd := "../../testdata"
+	srv2Resp := "hello"
+	insecureSkipTLSverify := false
+
+	// Server 2 that will actually fulfil the request.
+	ca, pub, priv := filepath.Join(cd, "rootca.crt"), filepath.Join(cd, "localhost-crt.pem"), filepath.Join(cd, "key.pem")
+	tlsConf, err := tlsutil.NewClientTLS(pub, priv, ca, insecureSkipTLSverify)
+	if err != nil {
+		t.Fatal(errors.Wrap(err, "can't create TLS config for client"))
+	}
+
+	tlsSrv2 := httptest.NewUnstartedServer(http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {
+		rw.Header().Set("Content-Type", "text/plain")
+		rw.Write([]byte(srv2Resp))
+	}))
+
+	tlsSrv2.TLS = tlsConf
+	tlsSrv2.StartTLS()
+	defer tlsSrv2.Close()
+
+	// Server 1 responds with a redirect to Server 2.
+	ca, pub, priv = filepath.Join(cd, "rootca.crt"), filepath.Join(cd, "crt.pem"), filepath.Join(cd, "key.pem")
+	tlsConf, err = tlsutil.NewClientTLS(pub, priv, ca, insecureSkipTLSverify)
+	if err != nil {
+		t.Fatal(errors.Wrap(err, "can't create TLS config for client"))
+	}
+
+	tlsSrv1 := httptest.NewUnstartedServer(http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {
+		u, _ := url.ParseRequestURI(tlsSrv2.URL)
+
+		// Make the request using the hostname 'localhost' (to which 'localhost-crt.pem' is issued)
+		// to verify that a successful TLS connection is made even if the client doesn't specify
+		// the hostname (SNI) in `tls.Config.ServerName`. By default the hostname is derived from the
+		// request URL for every request (including redirects). Setting `tls.Config.ServerName` on the
+		// client just overrides the remote endpoint's hostname.
+		// See https://github.com/golang/go/blob/3979fb9/src/net/http/transport.go#L1505-L1513.
+		u.Host = fmt.Sprintf("localhost:%s", u.Port())
+
+		http.Redirect(rw, r, u.String(), http.StatusTemporaryRedirect)
+	}))
+
+	tlsSrv1.TLS = tlsConf
+	tlsSrv1.StartTLS()
+	defer tlsSrv1.Close()
+
+	u, _ := url.ParseRequestURI(tlsSrv1.URL)
+
+	t.Run("Test with TLS", func(t *testing.T) {
+		g, err := NewHTTPGetter(
+			WithURL(u.String()),
+			WithTLSClientConfig(pub, priv, ca),
+		)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		buf, err := g.Get(u.String())
+		if err != nil {
+			t.Error(err)
+		}
+
+		b, err := io.ReadAll(buf)
+		if err != nil {
+			t.Error(err)
+		}
+
+		if string(b) != srv2Resp {
+			t.Errorf("expected response from Server2 to be '%s', instead got: %s", srv2Resp, string(b))
+		}
+	})
+
+	t.Run("Test with TLS config being passed along in .Get (see #6635)", func(t *testing.T) {
+		g, err := NewHTTPGetter()
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		buf, err := g.Get(u.String(), WithURL(u.String()), WithTLSClientConfig(pub, priv, ca))
+		if err != nil {
+			t.Error(err)
+		}
+
+		b, err := io.ReadAll(buf)
+		if err != nil {
+			t.Error(err)
+		}
+
+		if string(b) != srv2Resp {
+			t.Errorf("expected response from Server2 to be '%s', instead got: %s", srv2Resp, string(b))
+		}
+	})
+
+	t.Run("Test with only the CA file (see also #6635)", func(t *testing.T) {
+		g, err := NewHTTPGetter()
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		buf, err := g.Get(u.String(), WithURL(u.String()), WithTLSClientConfig("", "", ca))
+		if err != nil {
+			t.Error(err)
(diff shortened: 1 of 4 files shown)


bridgetkromhout:
I don't have permission to re-trigger the build in https://app.circleci.com/pipelines/github/helm/helm/2128/workflows/9899ccf8-58b5-4862-8a8a-23e3b3c4b953/jobs/13676 but I suspect it may be an intermittent failure; I can't reproduce it.

wahabmk:
> I don't have permission to re-trigger the build in https://app.circleci.com/pipelines/github/helm/helm/2128/workflows/9899ccf8-58b5-4862-8a8a-23e3b3c4b953/jobs/13676 but I suspect it may be an intermittent failure; I can't reproduce it.

The builds pass now. Added to the existing tests for new changes, and refactored the tests a bit to make it easier to read.

technosophos:
NOTE: This PR _MUST_ got through a security audit before merge, as it alters the SSL code.

wahabmk:
> NOTE: This PR _MUST_ got through a security audit before merge, as it alters the SSL code.

@technosophos what is the process for that? Anything I can do to contribute?

technosophos:
At least one LGTM must come from a security team member.

wahabmk:
> At least one LGTM must come from a security team member.

@technosophos could you (or somebody else) add a security team for review whenever schedule allows. Thanks for clarifying the process!

internal/tlsutil/tls.go
```diff
@@ -49,14 +50,26 @@ func NewClientTLS(certFile, keyFile, caFile string) (*tls.Config, error) {
 
 // CertPoolFromFile returns an x509.CertPool containing the certificates
 // in the given PEM-encoded file.
+// For non-windows OS, it loads the system cert pool and adds the certificates
```
bacongobbler, line 53: Wouldn't this be construed as a backwards-incompatible change, and at worst a security issue? There are many valid cases where you would not want Helm to load your system certificate pool.

The use case I'm thinking here... A potential attacker could add a malicious certificate to your root cert chain and redirect you to their chart server. That chart server contains malicious charts, and your Helm client just verified that chart server as a trusted endpoint.

bacongobbler (changes requested):
The changes suggested here open us up to several root cert chain attacks. Ones which we explicitly avoided by only allowing Helm to load a local certificate pool and add ONLY the certificates which were explicitly added by the end user.

I'm also highly concerned about the complete removal of the hostname verification that was done in httpgetter.go. While it is true that `BuildNameToCertificate` was deprecated, that doesn't mean we should remove all of the server name validation guards we put in.

If you can provide a set of unit tests which test for root cert chain attacks and hostname verification attacks, then I'd be more comfortable with the changes involved here. But at this time this PR could open us up to several significant security flaws. All of which I'm VERY uncomfortable with.

internal/tlsutil/tls.go
wahabmk: I reverted this change back to not loading system certs by default. Maybe we can optionally use system certs. However I will create another proposal/PR for that.

wahabmk:
> The changes suggested here open us up to several root cert chain attacks. Ones which we explicitly avoided by only allowing Helm to load a local certificate pool and add ONLY the certificates which were explicitly added by the end user.
> 
> I'm also highly concerned about the complete removal of the hostname verification that was done in httpgetter.go. While it is true that `BuildNameToCertificate` was deprecated, that doesn't mean we should remove all of the server name validation guards we put in.
> 
> If you can provide a set of unit tests which test for root cert chain attacks and hostname verification attacks, then I'd be more comfortable with the changes involved here. But at this time this PR could open us up to several significant security flaws. All of which I'm VERY uncomfortable with.

@bacongobbler thanks for the review! You pointed out 2 concerns:
1. Loading system cert pool by default causes vulnerability to root cert chain attacks.
2. Not specifying the SNI through `tls.Config.ServerName` causes vulnerability to hostname verification attacks.

I reverted the changes made for 1 so it is no longer applicable to this PR. As for 2, Go by default, already [validates](https://github.com/golang/go/blob/3979fb9af9ccfc0b7ccb613dcf256b18c2c295f0/src/net/http/transport.go#L1505-L1513) the remote hostname against certificate for every request. So if there is a redirect, Go will do the validation for the hostname in the original request and then for the hostname in the subsequent redirect as well. Specifying the `ServerName` manually just overwrites the hostname of the request. So, removing the `tls.Config.ServerName` overwrite would not introduce any new vulnerability.

wahabmk:
@bridgetkromhout the build failed due to timeout, could you re-run. Tanks!

bacongobbler:
> As for 2, Go by default already validates the remote hostname against certificate for every request. So if there is a redirect, Go will do the validation for the hostname in the original request and then for the hostname in the subsequent redirect as well. Specifying the ServerName manually just overwrites the hostname of the request. So, removing the tls.Config.ServerName overwrite would not introduce any new vulnerability.

Thank you for clarifying. However I'd still feel more comfortable if there were unit tests prepared along with this PR to validate that assumption. As at the end of the day we're liable for millions of users on a daily basis. I'd hate for there to be an unintentional security flaw introduced in a PR because we did not add a few test fixtures to ensure this is the case. At the end of the day it's up to the maintainers to respond to security issues, so taking a few extra steps to assert those assumptions would be wonderful.

wahabmk:
> > As for 2, Go by default already validates the remote hostname against certificate for every request. So if there is a redirect, Go will do the validation for the hostname in the original request and then for the hostname in the subsequent redirect as well. Specifying the ServerName manually just overwrites the hostname of the request. So, removing the tls.Config.ServerName overwrite would not introduce any new vulnerability.
> 
> Thank you for clarifying. However I'd still feel more comfortable if there were unit tests prepared along with this PR to validate that assumption. As at the end of the day we're liable for millions of users on a daily basis. I'd hate for there to be an unintentional security flaw introduced in a PR because we did not add a few test fixtures to ensure this is the case. At the end of the day it's up to the maintainers to respond to security issues, so taking a few extra steps to assert those assumptions before removing that code would really help us out (and make us more comfortable with merging these changes).
> 
> While we're at it... Tests to validate the first assumption (these changes allow `helm pull` to fetch with a redirect) would be great to see as well. It's important to write tests along with any fixes to ensure future iterations do not break backwards compatibility. There have been many times where a fix _looks_ correct but doesn't actually fix the problem, so it's worth writing that down.
> 
> I restarted the CI build a few times. Try rebasing on HEAD to see if there was a fix applied to resolve this issue. It's also possible that CircleCI is seeing a performance degradation in its service so that could be causing a slowdown/timeout.

@bacongobbler updated the PR with a unit test to verify TLS works with redirects and the client receives a response from redirected server.

wahabmk:
@bacongobbler could you do a re-review, thanks.

bacongobbler (dismissed):
code LGTM. Will need another maintainer's approval before we can merge as per the contributing policy.

shahzadlone:
Is there any progress on this? seems like it's been marinating with a single approval for over a year now.

shahzadlone (approved):
LGTM

joejulian:
@wahabmk can you fix the conflict?

wahabmk:
Working on it...

joejulian:
@wahabmk I merged in main to kick off CI which, unfortunately, failed. Could you take a look please?

pkprzekwas:
What's the status of this issue. We are affected and wonder when it could possibly be merged.

SimonTheLeg:
I have pushed a PR that builds on this one and fixes all the CI issues (at least when running the tests locally). Additionally it includes some documentation on how to generate the certificates:

https://github.com/helm/helm/pull/12545

wahabmk:
@SimonTheLeg Thanks for the mention. I fixed the unit tests but I still have the DCO failing. @bacongobbler does this PR still need 1 more approval?

wahabmk:
@bacongobbler @joejulian @technosophos @mattfarina I have fixed the CI and rebased from `main` branch in the latest commit I pushed. There was a previous commit by me that was not signed which caused the DCO to fail. So I rebased my commits into a single one and signed it. This PR needs 1 more approval. Could somebody kindly review it as it has been sitting since 2021. Thanks!

pkg/getter/httpgetter_test.go
```diff
@@ -331,6 +331,121 @@ func TestDownloadTLS(t *testing.T) {
...
+
+	tlsSrv1 := httptest.NewUnstartedServer(http.HandlerFunc(func(rw http.ResponseWriter, r *http.Request) {
+		u, _ := url.ParseRequestURI(tlsSrv2.URL)
+
+		// Make the request using the hostname 'localhost' (to which 'localhost-crt.pem' is issued)
+		// to verify that a successful TLS connection is made even if the client doesn't specify
+		// the hostname (SNI) in `tls.Config.ServerName`. By default the hostname is derived from the
+		// request URL for every request (including redirects). Setting `tls.Config.ServerName` on the
+		// client just overrides the remote endpoint's hostname.
+		// See https://github.com/golang/go/blob/3979fb9/src/net/http/transport.go#L1505-L1513.
```
wahabmk, line 370: In the latest force-push, I just updated this URL in the comment. The original one was outdated. No code changes were done.

shahzadlone (approved):
LGTM

pkg/getter/httpgetter_test.go
gjenkins8: to confirm the above:

https://github.com/helm/helm/compare/95d986b7ca68438e2d16f43a1b72d3c0371969f9..4c4cd8ecd1a3bac047c4662eaca3faaa8659eaee

```
- 		// See https://golang.org/src/net/http/transport.go#L1505.
+		// See https://github.com/golang/go/blob/3979fb9/src/net/http/transport.go#L1505-L1513.
```

shahzadlone:
> @joejulian -- want to rerun CI here? (not sure why the github checks are referring to circle CI here, this branch is up to date / on top of recent helm main branch)

Branch seems to be 13 commits behind at the moment: https://github.com/helm/helm/compare/2745909d...main

image

wahabmk:
Rebased from upstream `main` branch

shahzadlone (approved):
> Rebased from `helm:main` branch

LGTM, lets have this merged without further delay.

@bacongobbler @joejulian @technosophos @mattfarina @gjenkins8

wahabmk:
@gjenkins8 just wanted to inquire if there is any progress for getting this PR reviewed/merged. Thanks!

sabre1041 (commented):
It makes sense to not enforce SNI based on the hostname specifically when a CA is provided. Interesting that this type of logic was implemented in this fashion.

However, would we also want to implement the same changes for other getters (like [oci](https://github.com/helm/helm/blob/main/pkg/getter/ocigetter.go#L132-L136))

gjenkins8 (dismissed):

wahabmk:
@gjenkins8 Rebased from `main`. Looks like the CI needs approval to run.

wahabmk:
@gjenkins8 Fixed the reported lint error in 2nd commit.

shahzadlone (approved):
LGTM

wahabmk:
@gjenkins8 could you run the CI please, thanks.

shahzadlone:
The lack of priority on this small PR, which resolves noteable issues for many, is mind-boggling.

wahabmk:
@gjenkins8 rebased

wahabmk:
@gjenkins8 could you run the CI please so we could know whether anything else needs fixing. Thanks!

gjenkins8 (approved):

scottrigby (approved):
@wahabmk nice work 👏 Also thanks for the tests 💯 LGTM

PR #8582, opened 2020-08-11
Wait: turn the debug logs into stdout print

desaintmartin:
**What this PR does / why we need it**:

if `--wait` flag for helm `install/upgrade` is specified, Helm will log to standard output (`stdout`) the wait informations.

Before: it was logged only with `--debug` flag to `stderr`.

Solve https://github.com/helm/helm/issues/7760

**Example output:**

using a sample `values.yaml` containing:
```yaml
persistence:
    enabled: false
replication:
    enabled: false
    slaveReplicas: 3
```

```bash
bin/helm upgrade --install test stable/postgresql --wait --values values.yaml
Release "test" does not exist. Installing it now.
WARNING: This chart is deprecated
beginning wait for 6 resources with timeout of 5m0s
StatefulSet is not ready: default/test-postgresql-slave. 1 out of 3 expected pods have been scheduled
StatefulSet is not ready: default/test-postgresql-slave. 1 out of 3 expected pods have been scheduled
StatefulSet is not ready: default/test-postgresql-slave. 1 out of 3 expected pods have been scheduled
StatefulSet is not ready: default/test-postgresql-slave. 1 out of 3 expected pods have been scheduled
StatefulSet is not ready: default/test-postgresql-slave. 1 out of 3 expected pods have been scheduled
StatefulSet is not ready: default/test-postgresql-slave. 1 out of 3 expected pods have been scheduled
StatefulSet is not ready: default/test-postgresql-slave. 1 out of 3 expected pods have been scheduled
StatefulSet is not ready: default/test-postgresql-slave. 1 out of 3 expected pods have been scheduled
StatefulSet is not ready: default/test-postgresql-slave. 1 out of 3 expected pods have been scheduled
StatefulSet is not ready: default/test-postgresql-slave. 1 out of 3 expected pods have been scheduled
StatefulSet is not ready: default/test-postgresql-slave. 1 out of 3 expected pods have been scheduled
StatefulSet is not ready: default/test-postgresql-slave. 1 out of 3 expected pods have been scheduled
StatefulSet is not ready: default/test-postgresql-slave. 1 out of 3 expected pods have been scheduled
StatefulSet is not ready: default/test-postgresql-slave. 2 out of 3 expected pods have been scheduled
StatefulSet is not ready: default/test-postgresql-slave. 2 out of 3 expected pods have been scheduled
StatefulSet is not ready: default/test-postgresql-slave. 2 out of 3 expected pods have been scheduled
[...]

Issue #7760, opened 2020-03-12
How to achieve a good level of verbosity with `helm ... --wait` ?


pikapika2020:
I want to achieve a good level of verbosity with `helm --wait`. Its output is completely silent. I want to have same strings as with `--debug` flag, but without YAMLs being printed

`helm upgrade --wait --install --set test=guest  --debug test .`:
```
upgrade.go:90: [debug] performing update for test
upgrade.go:232: [debug] creating upgraded release for test
client.go:163: [debug] checking 2 resources for changes
client.go:403: [debug] Looks like there are no changes for ConfigMap "test1"
client.go:403: [debug] Looks like there are no changes for ConfigMap "test2"

...computed values, YAMLs and so on...
```

`-v 6` and so on does not give me what would be preferrable

It would be great if with `--wait` flag it printed out something like:
```
[wait] checking these resources for changes:
[wait] <blah blah blah>
[wait] patching resource ConfigMap test1
[wait] ...
[wait] all resources ready
[wait] ... etc.
```

seandilda:
I would like to second this.  Having an option to give the current debug output minus the rendered yaml files would be very helpful.

smueller18:
As a workaround I currently use grep to filter only release summary output and ignore rendered template output (but it is also possible, that some lines in the rendered template output could match). The debug output before the release summary is printed on stderr so it is not filtered by grep at all.

```bash
helm upgrade -i --debug --wait release-name . \
  | grep -E "(Happy\ Helming|NAME\: |LAST DEPLOYED\: |NAMESPACE\: |STATUS\: |REVISION\: | TEST SUITE\: )" \
    --color=never || true
```

An example output would look like

```
history.go:52: [debug] getting history for release kured
upgrade.go:121: [debug] preparing upgrade for kured
upgrade.go:129: [debug] performing update for kured
upgrade.go:308: [debug] creating upgraded release for kured
client.go:173: [debug] checking 10 resources for changes
client.go:431: [debug] Looks like there are no changes for PodSecurityPolicy "kured"
client.go:431: [debug] Looks like there are no changes for ServiceAccount "kured"
client.go:431: [debug] Looks like there are no changes for ConfigMap "grafana-dashboards-kured"
client.go:431: [debug] Looks like there are no changes for ClusterRole "kured"
client.go:431: [debug] Looks like there are no changes for ClusterRoleBinding "kured"
client.go:431: [debug] Looks like there are no changes for Role "kured"
[...]

PR #2040, opened 2017-02-28
install: Return more user-friendly response on eexists

dmonakhov:
Most newcomers hit helm's 'feature' there deleted instalations are not really deleted.
Use-case
# helm install  --name gitlab-ce-1 ./stable/gitlab-ce
# helm delete gitlab-ce-1
# helm list
# helm install  --name gitlab-ce-1 ./stable/gitlab-ce
Error: a release named "gitlab-ce-1" already exists

There is a lot duplicated bugs in bugzilla which simply explains that packages
should be deleted with --pure flag. But such bugs appeared again and again
because this behavior is not obvious.

Let's help user to figure out what happens ASAP.

PR #9323, opened 2021-02-04
Adding new verbose flag

### LINKED ISSUE #8809: Introduce verbose output
Helm currently has it's normal output and output with `--debug` messages. This is a proposal to introduce another level between normal and debug output.

kubectl, using klog, has the ability to set a verbosity level. This level is not clearly defined so users have to work it out. But, there are multiple levels. I wouldn't recommend we have that experience.

Homebrew, the mac package manager, has two levels. `--verbose` and `--debug` that can be layered on to the normal output. Verbose prints more information about what's happening while debug prints details you'd need if you want to debug something.

This feature request suggests the same idea for Helm. For example, if you used `--verbose` on `helm install` it would tell you what's happening as it locates the chart, downloads it, renders the templates, pushes them, etc. I wouldn't need to give lots of internal values and details that debug should supply.

Thoughts?

sorenmat:
This allow us to output the progress of the installation / upgrade
with a bit more detail.
This is particular useful in combination with the wait flag, as it
outputs the progress into the console.

Fixes: #8809 #7760 #2040 #8582 
Example output of `bin/helm upgrade --install my-kubeview kubeview/kubeview --version 0.1.19 --wait --verbose`                                                                                                         
```
Getting history for release my-kubeview
Release "my-kubeview" does not exist. Installing it now.
creating 5 resource(s)
Beginning wait for 5 resources with timeout of 5m0s
Service does not have load balancer ingress IP address: default/my-kubeview
Deployment is not ready: default/my-kubeview. 0 out of 1 expected pods are ready
Deployment is not ready: default/my-kubeview. 0 out of 1 expected pods are ready
Deployment is not ready: default/my-kubeview. 0 out of 1 expected pods are ready
Deployment is not ready: default/my-kubeview. 0 out of 1 expected pods are ready
NAME: my-kubeview
LAST DEPLOYED: Thu Feb  4 09:39:14 2021
NAMESPACE: default
STATUS: deployed
REVISION: 1
NOTES:
=====================================
==== KubeView has been deployed! ====
=====================================
  To get the external IP of your application, run the following:

  export SERVICE_IP=$(kubectl get svc --namespace default my-kubeview -o jsonpath='{.status.loadBalancer.ingress[0].ip}')
  echo http://$SERVICE_IP

  NOTE: It may take a few minutes for the LoadBalancer IP to be available.
        You can watch the status of by running 'kubectl get --namespace default svc -w my-kubeview'
```
**If applicable**:


sorenmat:
@bacongobbler you seems to be involved in some of the discussions on the issues this addresses, any thoughts?

bacongobbler:
In a recent Helm security audit, it was recommended by the auditors that a global `--log-level` flag should be introduced to the project.

I would be interested in PRs that introduce that behaviour at a higher level. The logging capabilities provided by languages like RUST_LOG seem like a more ambitious goal to achieve, but it will be worth it in the long run.

Reading through this PR it looks as though each package has to pass a `log.Logger` and consume that logger. There are also no way to separate between different levels of verbosity, which has been highly requested by many community members.

In other words, I'd aim to write a log interface that sits at a much higher level than what you've proposed here. Something akin to Rust's logging system would be much more pluggable and easier to maintain going forward.

bacongobbler:
It is also worth keeping in mind that this clashes with the existing `-v` flag inherited from kubectl. `-v 10 --verbose` introduces two separate levels of logging. So the design here is really poor for the end user.

sorenmat:
seems like this was a bad idea, according to the comments

PR #9335, opened 2021-02-06
WIP: [Install]: Change --dependency-update Behavior

### LINKED ISSUE #7162: Add --dependency-update flag to the helm upgrade command
Output of `helm version`:v3.0.0

Output of `kubectl version`:v1.15.0

oliverbaehler:

This Pull Request changes the functionality on how dependencies are handled when using the install subcommand and adds the same functionality to the `upgrade` subcommand. We noticed that when using the `--dependency-update` dependencies are only updated, if they are not existent in the **charts/**. If they are already existent no update will be made. In our opinion that's not the correct behavior. The update should always be made when adding the `--dependency-update`. That's what has changed on the `install` subcommand (We did not create a dedicated issue for this). We are not sure if that's the behavior you expected from this flag. But looking at the documentation it states:

```
--dependency-update            run helm dependency update before installing the chart
```

And the change actually enforces this behavior. For us this makes a lot more sense, especially when you have version ranges for dependencies. Let us know what you think about it. This change should be classified as bugfix but changes functionality, but is in our opinion backwards compatible.

The same functionality was implemented for the `upgrade` subcommand. It adds the `--dependency-update` flag dependencies are updated the same as on the install subcommand. If the `--install`flag is present, the parameter ist just given to the InstallClient. Issues regarding this change: 
  - fixes #7162 

This change should be classified as feature.

This is our first commit on this project. If you have any feedback please let us know :)

**If applicable**:

### DIFF
--- a/cmd/helm/install.go
+++ b/cmd/helm/install.go
@@ -209,32 +209,39 @@ func runInstall(args []string, client *action.Install, valueOpts *values.Options
 		warning("This chart is deprecated")
 	}
 
+	// Only check dependencies if there are any
 	if req := chartRequested.Metadata.Dependencies; req != nil {
 		// If CheckDependencies returns an error, we have unfulfilled dependencies.
 		// As of Helm 2.4.0, this is treated as a stopping condition:
 		// https://github.com/helm/helm/issues/2209
-		if err := action.CheckDependencies(chartRequested, req); err != nil {
-			if client.DependencyUpdate {
-				man := &downloader.Manager{
-					Out:              out,
-					ChartPath:        cp,
-					Keyring:          client.ChartPathOptions.Keyring,
-					SkipUpdate:       false,
-					Getters:          p,
-					RepositoryConfig: settings.RepositoryConfig,
-					RepositoryCache:  settings.RepositoryCache,
-					Debug:            settings.Debug,
-				}
-				if err := man.Update(); err != nil {
-					return nil, err
-				}
-				// Reload the chart with the updated Chart.lock file.
-				if chartRequested, err = loader.Load(cp); err != nil {
-					return nil, errors.Wrap(err, "failed reloading chart after repo update")
-				}
-			} else {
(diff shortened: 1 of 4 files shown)


bacongobbler:
Can you please check CI? The unit tests are failing due to this change.

bacongobbler:
Hi @oliverbaehler. CI is still failing. In addition one of your commits has not been signed off. Please review the failing checks.

bacongobbler:
> This change should be classified as bugfix but changes functionality, but is in our opinion backwards compatible.

If it changes the behaviour of the feature, it is backwards-incompatible. Because of that reason we cannot consider merging it until Helm 4.

If you can determine that the original behaviour of `--dependency-update` matches the behaviour described above, then we can make the case that a regression was introduced in a future update. But if this has always been `--dependency-update`'s behaviour, then to change it in the name of "improving cohesion" could potentially break the use cases it sought out to address, and therefore breaking compatibility.

There are two ways to address this:

1. If you can discuss this issue with the original author of `--dependency-update` and get a response whether this would be construed as a bug, or
2. we wait until Helm 4 to refactor this feature's existing behaviour to match `helm dependency update`'s behaviour.

bacongobbler:
If you just want `helm upgrade --dependency-update`, I'd suggest breaking this out into two separate PRs:

1. A PR that introduces the feature to `helm upgrade`
2. A PR that proposes changing the behaviour of `--dependency-update` to match `helm dependency update`

The former we can probably review and accept without any issue. But the latter will require more discussion as described earlier.

oliverbaehler:
Hi @bacongobbler, Thanks for your Feedback. Sorry I have kinda lost focus on working o this, still was trying to figure out why the CI was failing since the change did break compatibility with that that Unit Test. Anyway I fully agree with you with splitting this into two PRs. If that's okay i am going to let this one be the one which changes the `--dependency-update` on the install cmd, since your thoughts are already present as comments. I will create a new PR for the upgrade cmd and reference it when created.

PR #9347, opened 2021-02-09
fix(helm): get/get-helm-3 whitespace support in runAsRoot

### LINKED ISSUE #9346: [get/get-helm-3] runAsRoot is broken if arguments contains spaces
Currently the `get-helm-3`-script fails if `HELM_INSTALL_DIR` contains spaces, this is due to an issue in the `runAsRoot` function, not wrapping the command properly.

Reproduce: Run `get-helm-3` with a `HELM_INSTALL_DIR` containing at least one space

PS: yes I know, spaces in filenames are a pita ;)

mmusenbr:
When `get`/`get-helm-3` is run with a HELM_INSTALL_DIR containing spaces, the installation fails.

This PR fixes the issues, by wrapping the parameters correctly.

Closes #9346

### DIFF
--- a/scripts/get-helm-3
+++ b/scripts/get-helm-3
@@ -57,13 +57,11 @@ initOS() {
 
 # runs the given command as root (detects if we are root already)
 runAsRoot() {
-  local CMD="$*"
-
-  if [ $EUID -ne 0 -a $USE_SUDO = "true" ]; then
-    CMD="sudo $CMD"
+  if [ $EUID -ne 0 -a "$USE_SUDO" = "true" ]; then
+    sudo "${@}"
+  else
+    "${@}"
   fi
-
-  $CMD
 }
 
 # verifySupported checks that the os/arch combination is supported for
--- a/scripts/get
+++ b/scripts/get
@@ -50,13 +50,11 @@ initOS() {
 
 # runs the given command as root (detects if we are root already)
 runAsRoot() {
-  local CMD="$*"
-
-  if [ $EUID -ne 0 -a $USE_SUDO = "true" ]; then
-    CMD="sudo $CMD"
+  if [ $EUID -ne 0 -a "$USE_SUDO" = "true" ]; then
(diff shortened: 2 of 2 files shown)


scripts/get-helm-3
```diff
@@ -57,13 +57,11 @@ initOS() {
 
 # runs the given command as root (detects if we are root already)
 runAsRoot() {
-  local CMD="$*"
-
-  if [ $EUID -ne 0 -a $USE_SUDO = "true" ]; then
-    CMD="sudo $CMD"
+  if [ $EUID -ne 0 -a "$USE_SUDO" = "true" ]; then
+    exec sudo "${@}"
```
jdolitsky, line 61: Could this be done without using `exec`? For example `"sudo \"$CMD\""`

jdolitsky (changes requested):

scripts/get-helm-3
redeamer: Sorry, yes of course we need to find another way, I was wrongly thinking, exec is not suitable here. I will recheck in the evening.

redeamer: I was changing it, and rebased to current master, but the changes do not show up, but are visible in my repo.. let's see. I ping you again, once I see the changes in the PR as well.

redeamer: @jdolitsky I had to change the destination branch forth and back, and now the latest changes show up. But now the label was not updated back and the DCO checks fails now for an old commit already in master; where the check was green after intially opening the PR.
Should I simply open a new PR to have a cleaner state?

jdolitsky: yes, its probably the easiest approach to close and re-open

redeamer:
Close due to issues (DCO checks) after changing the merge target to overcome update issue of github. Will open a new PR.

redeamer:
@jdolitsky reopened in #9349

PR #9357, opened 2021-02-12
improve tests on windows

dragonchaser:
We have seen that the tests running on windows fail when the fixtures in `cmd/helm/testdata/output` have been generated in linux (and vica versa) due to the different ways the two OS handle line breaks (`\r\n` vs. `\n`). 
The changes in this PR us the normalize function from `internal/test/test.go` to replace `\r\n` by `\n` in the `actual` and the `expected` strings to achieve compatibility of the fixtures.

### DIFF
--- a/internal/test/test.go
+++ b/internal/test/test.go
@@ -77,6 +77,7 @@ func path(filename string) string {
 }
 
 func compare(actual []byte, filename string) error {
+	actual = normalize(actual)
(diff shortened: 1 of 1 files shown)


mattfarina:
@dragonchaser Could you rebase off of master. There is a lint CI fix there that will help your tests pass.

dragonchaser:
done

mattfarina (approved):

PR #9363, opened 2021-02-15
feat(comp): Improve completion for plugin commands

marckhouzam:

The `plugin update` and `plugin uninstall` commands can accept more than one plugin name as arguments; this commit teaches the completion logic to respect that.

The logic to not suggest a plugin that is already on the command-line is the same as the logic for completion of the `helm repo rm` command which also takes multiple arguments. See https://github.com/helm/helm/blob/64c8df187af892c2a9f872f91e892a717df2e7ff/cmd/helm/repo_list.go#L103

**If applicable**:


bacongobbler (approved):
LGTM

PR #9367, opened 2021-02-16
Increase golangci-lint timeout

marckhouzam:

I'm not sure we need this change.
Recently, CircleCI has been timing out on `make test-style` for all PRs (4 PRs and over 10 build attempts).  I've restarted the build multiple times without success.  It only worked when I increased the timeout for golangci-lint. This all works locally without problem.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -113,7 +113,7 @@ test-coverage:
 
 .PHONY: test-style
 test-style:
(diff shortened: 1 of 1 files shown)


mattfarina:
I've seen the same problem and I've also seen it in projects importing Helm to use as a library who are using GH Actions for CI.

PR #9380, opened 2021-02-18
fix reuse values

bshh:
fix reuse-values
If we add some new key values which readed from Values.yaml to the new chart ,upgrade release by reuse-values will fail to upgrade.

### DIFF
--- a/pkg/action/upgrade.go
+++ b/pkg/action/upgrade.go
@@ -442,15 +442,12 @@ func (u *Upgrade) reuseValues(chart *chart.Chart, current *release.Release, newV
 	if u.ReuseValues {
 		u.cfg.Log("reusing the old release's values")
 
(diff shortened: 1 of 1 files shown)


bacongobbler:
hi @bshh. Please first submit a ticket describing the issue you are seeing before submitting a PR. That way we can discuss what might be causing your issue before determining a fix. It is quite possible that there are other tickets that are similar to the issue you are describing, or that this could be its intended behaviour.

In the future, we also ask that you provide tests along with any proposed fixes. That way no regressions are introduced in the future, and we have some way to verify that the "issue" has been resolved.

Once you have completed those steps, please proceed with submitting a PR. But I'm going ahead and closing this for now. We do not plan on reviewing any PRs before discussing the issue at-hand.

Thanks.

Issue #7987, opened 2020-04-24
"helm fetch/pull" failing in Helm 3.2.0


carrodher:
I am trying to fetch a chart using the new Helm 3.2.0 version but I obtain the following error:
```console
$ helm repo update
Hang tight while we grab the latest from your chart repositories...
...Successfully got an update from the "bitnami" chart repository
Update Complete. ⎈ Happy Helming!⎈

$ helm fetch bitnami/redis-cluster
Error: rename /tmp/redis-cluster-2.0.8.tgz770913527 redis-cluster-2.0.8.tgz: invalid cross-device link
```
the same issue happens using `pull` instead of `fetch`:
```console
$ helm pull bitnami/redis-cluster
Error: rename /tmp/redis-cluster-2.0.8.tgz205065242 redis-cluster-2.0.8.tgz: invalid cross-device link
```

But, using the latest Helm 3.1.X, the command works as expected:
```console
$ helm313 repo update
Hang tight while we grab the latest from your chart repositories...
...Successfully got an update from the "bitnami" chart repository
Update Complete. ⎈ Happy Helming!⎈

$ helm313 fetch bitnami/redis-cluster

$ ls | grep redis
redis-cluster-2.0.8.tgz
```

----------------------------------
Output of `helm version`:
```
$ helm version
version.BuildInfo{Version:"v3.2.0", GitCommit:"e11b7ce3b12db2941e90399e874513fbd24bcb71", GitTreeState:"clean", GoVersion:"go1.13.10"}
```

Output of `kubectl version`:
```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.9", GitCommit:"2e808b7cb054ee242b68e62455323aa783991f03", GitTreeState:"clean", BuildDate:"2020-01-18T23:33:14Z", GoVersion:"go1.12.12", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"15+", GitVersion:"v1.15.11-gke.1", GitCommit:"aa93664bafe74f3c051e021b27a64e28ac34ab40", GitTreeState:"clean", BuildDate:"2020-03-18T22:49:35Z", GoVersion:"go1.12.17b4", Compiler:"gc", Platform:"linux/amd64"}
```

Cloud Provider/Platform (AKS, GKE, Minikube etc.): 
GKE

carrodher:
Not 100% sure, but this PR can be related #7720

[...]

PR #9393, opened 2021-02-22
Use generated temp directory for helm dependencies

### LINKED ISSUE #5567: when killing `helm dep up` the next `dep up` fails
when killing the process running `helm dep up` it sometimes leave a temporary folder named `tmpcharts` and this makes the next `helm dep up` fail with the following error.
```
Error: Unable to move current charts to tmp dir: rename /mychart/charts /mychart/tmpcharts: file exists
```
This is easily reproducable by creating a folder named `tmpcharts` inside an helm chart root folder and try to launch `helm dep up`

Of course a workaround is to remove the folder manually before running the `helm dep up` but that would be great that helm did that automatically somehow.

Output of `helm version`:
```
$helm version
Client: &version.Version{SemVer:"v2.12.2", GitCommit:"7d2b0c73d734f6586ed222a567c5d103fed435be", GitTreeState:"clean"}
```
Output of `kubectl version`:
```
kubectl version
Client Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.3", GitCommit:"721bfa751924da8d1680787490c54b9179b1fed0", GitTreeState:"clean", BuildDate:"2019-02-04T04:48:03Z", GoVersion:"go1.11.5", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.0", GitCommit:"ddf47ac13c1a9483ea035a79cd7c10005ff21a6d", GitTreeState:"clean", BuildDate:"2018-12-03T20:56:12Z", GoVersion:"go1.11.2", Compiler:"gc", Platform:"linux/amd64"}
```
Cloud Provider/Platform (AKS, GKE, Minikube etc.): docker for mac

The code handling this is here : https://github.com/helm/helm/blob/53d432fa58748412ff3dc10bc27cbf996d96c3ed/pkg/downloader/manager.go#L190

jsoref:
* Use [ioutil.TempDir()](https://golang.org/pkg/io/ioutil/#TempDir) to pick temporary chart directory name
* Add test

Closes #5567, #8846 

* @gzur suggested this approach https://github.com/helm/helm/pull/8846#issuecomment-768479847
* @deathname wrote the test -- I tested by running w/o this patch and the test fails -- I tested with this patch and the test passes

I'm sorry for pushing this forward, we're hitting this and it's a small enough thing that I'd rather get it in than wait for someone else. Similarly, if any work that I propose anywhere gets stuck because I'm unavailable (which happens periodically), I'm more than happy for someone else to take my work and run with it.

We = consumers of argoproj/argo-cd#5107

**If applicable**:

### DIFF
--- a/pkg/downloader/manager.go
+++ b/pkg/downloader/manager.go
@@ -247,7 +247,7 @@ func (m *Manager) downloadAll(deps []*chart.Dependency) error {
 	}
 
 	destPath := filepath.Join(m.ChartPath, "charts")
-	tmpPath, err := ioutil.TempDir("", "tmpcharts-*")
+	tmpPath, err := ioutil.TempDir(m.ChartPath, "tmpcharts-*")
 	if err != nil {
 		return errors.Wrap(err, "failed to create tmpcharts dir")
 	}
--- a/pkg/downloader/manager.go
+++ b/pkg/downloader/manager.go
@@ -247,7 +247,11 @@ func (m *Manager) downloadAll(deps []*chart.Dependency) error {
 	}
 
 	destPath := filepath.Join(m.ChartPath, "charts")
-	tmpPath := filepath.Join(m.ChartPath, "tmpcharts")
+	tmpPath, err := ioutil.TempDir("", "tmpcharts-*")
+	if err != nil {
+		return errors.Wrap(err, "failed to create tmpcharts dir")
+	}
+	os.Remove(tmpPath)
 
 	// Create 'charts' directory if it doesn't already exist.
 	if fi, err := os.Stat(destPath); err != nil {
--- a/pkg/downloader/manager_test.go
+++ b/pkg/downloader/manager_test.go
@@ -17,6 +17,7 @@ package downloader
 
 import (
 	"bytes"
+	"os"
 	"path/filepath"
 	"reflect"
 	"testing"
@@ -178,6 +179,42 @@ func TestGetRepoNames(t *testing.T) {
 		if !eq {
 			t.Errorf("%s: expected map %v, got %v", tt.name, l, tt.name)
 		}
+	}
+}
+
+func TestDownloadAll(t *testing.T) {
+	// If we run downloadAll without defer than this test will fail
+	// 1st downloadAll creates charts folder
+	// 2nd downloadAll copies existing charts to tmpcharts and fails but tmpcharts folder will be removed
+	// 3rd downloadAll again runs with another dependency
+	b := bytes.NewBuffer(nil)
+	m := &Manager{
+		Out:              b,
+		RepositoryConfig: repoConfig,
+		RepositoryCache:  repoCache,
+	}
+	// remove charts folder at the end of test
+	defer os.RemoveAll("charts")
+
+	// call downloadAll so that it will create charts directory
+	err := m.downloadAll([]*chart.Dependency{
+		{Name: "local-dep", Repository: "file://./testdata/signtest", Version: "0.1.0"},
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	// this will copy charts directory content to tmp directory and recreates new charts
+	err = m.downloadAll([]*chart.Dependency{
+		{Name: "local-subchart", Repository: ""},
+	})
+	if err == nil {
+		t.Fatal(err)
+	}
+	err = m.downloadAll([]*chart.Dependency{
+		{Name: "local-dep", Repository: "file://./testdata/signtest", Version: "0.1.0"},
+	})
+	if err != nil {
+		t.Fatal(err)
 	}
 }
 
--- a/pkg/downloader/manager_test.go
+++ b/pkg/downloader/manager_test.go
@@ -17,10 +17,13 @@ package downloader
 
 import (
 	"bytes"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"reflect"
 	"testing"
+
+	"helm.sh/helm/v3/internal/third_party/dep/fs"
 
 	"helm.sh/helm/v3/pkg/chart"
 	"helm.sh/helm/v3/pkg/chartutil"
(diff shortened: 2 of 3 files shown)


bacongobbler:
This was the tactic we used in the past. Unfortunately it lead to issues like https://github.com/helm/helm/issues/7987.

Can you please check and verify this change will not re-introduce that regression?

jsoref:
Oh, this is probably not precisely related to tempdir, it's related to any time there's a `rename` or `move` operation.

But, sure, I can see about making a version that doesn't trip on this.

The easiest place to trip over this is indeed `/tmp` because it's _often_ a distinct file system, but any two directories can be different file systems :-)

I'll try to have something today.

bacongobbler:
We have a package dedicated to rename or move operations that work around these cross-link device errors. See [`RenameWithFallback`](https://github.com/helm/helm/blob/8ad90b25e55090eb1fe8f68f0a751474c99a5d41/internal/third_party/dep/fs/fs.go#L49-L52).

jsoref:
@bacongobbler, so in 3.2.0, this code in `downloader.atomicWriteFile()` used `os.Rename()`:
https://github.com/helm/helm/blob/853ba2de16a04d0715c44937162e0b58752a99d6/pkg/downloader/chart_downloader.go#L97

conveniently, it's exactly where something would be upset about renames across paths.

That code doesn't exist in master.

jsoref:
There are 3 instances of `os.Rename()` today as of fd5788ef0a488b3e78e18c6cf365ee868190dc7a:

https://github.com/helm/helm/search?q=os.rename
1. `AtomicWriteFile` -- this is a lie, it's not anywhere near as atomic, because it isn't calling `os.Rename` directly, it's calling `fs.RenameWithFallback`:
https://github.com/helm/helm/blob/6297c021cbda1483d8c08a8ec6f4a99e38be7302/internal/fileutil/fileutil.go#L28-L50
2. `TestTemplateIntegrationHappyPath` (I'm happy to change):
https://github.com/helm/helm/blob/0ba4fcba8782ee226183b405af6fa30337f94ff7/pkg/lint/rules/template_test.go#L76-L79
3. `fs.RenameWithFallback()` itself: https://github.com/helm/helm/blob/6297c021cbda1483d8c08a8ec6f4a99e38be7302/internal/third_party/dep/fs/fs.go#L52-L63

iow, I _believe_ that other than the tests, you've essentially already fixed the underlying buggy code.

If you can check my logic, I'd appreciate it, as I'm not claiming to be an expert on any of this, although I do have past experience w/ rename/move hell on other platforms/environments.

pkg/downloader/manager.go
```diff
@@ -247,7 +247,11 @@ func (m *Manager) downloadAll(deps []*chart.Dependency) error {
 	}
 
 	destPath := filepath.Join(m.ChartPath, "charts")
-	tmpPath := filepath.Join(m.ChartPath, "tmpcharts")
+	tmpPath, err := ioutil.TempDir("", "tmpcharts-*")
+	if err != nil {
+		return errors.Wrap(err, "failed to create tmpcharts dir")
+	}
+	os.Remove(tmpPath)
```
gzur, line 254: If I understand this correctly, we create a `tmpcharts-*`directory via `ioutil.TempDir` and then delete it, so that we can rename `charts` to it later.

Isn't this unnecessarily brittle? 
`os.Remove` [can return an error](https://golang.org/src/os/file_unix.go?s=9160:9190#L324) in which case we might (or might not) fail a bit further down.

Would it not be cleaner to just generate a name (something like what is being done inside `ioutil.TempDir`), make sure it doesn't exist via `os.Stat` and move on from there.

gzur: Having said that, I have no authority here 🙃

jsoref: I personally prefer to do as little work as possible and rely on libraries rather than copy+pasting code.

That said, I'm happy to do a certain amount of extra work as requested by an org in order to get a PR merged.

I could also use `ioutil.Tempfile` and just get the filename (minus the path) and then use that as the local name instead.

That seems like a reasonable middle ground. Yes, crossing file systems is something one should generally try to avoid.

gzur: > I personally prefer to do as little work as possible and rely on libraries rather than copy+pasting code.

Amen to that, and on second thought, I think the touch/delete on the dir is not that big a deal.

How about something like this then? (https://github.com/gzur/helm/commit/ca0fe09bbccb567273563243cf07c14e80f656d3)
```
tmpPath, err := ioutil.TempDir(m.ChartPath, "tmpcharts-*")
```

jsoref: @gzur I've pulled in your changes and fixed them up. If you don't care about authorship, I can squash it into one of the other commits. And/or I can add a `co-authored-by` field

gzur: I was gonna be all blasé and tell you just squash and ship, but on second thought a mention in the `co-authored-by` field would be nice.

gzur (approved):
lgtm (though I have no authority here)

pkg/downloader/manager.go
jsoref: Thinking about it, I think I want to squash and use co-authored-by, because even the initial suggestion was yours.

jsoref: Sigh, your changes are also to the test by @deathname. So, either I fold everything together, or leave things as is.

I'm leaving them alone unless someone directs otherwise.

gzur: How do we move this forward?

jsoref:
@bacongobbler ?

bacongobbler:
As I stated earlier, this will [re-introduce a bug](https://github.com/helm/helm/issues/7987) that was introduced by a similar change made in Helm 3.2.0. I highly suggest reading through that ticket as well as a few others related to this proposal, most notably https://github.com/helm/helm/issues/2998 and the current discussion in #8846. But as it stands, we cannot accept this knowing it will re-introduce a bug.

In light of that, I'm going ahead and closing this in favor of #8846. @mattfarina provided actionable feedback in that PR how we could potentially move forward with that change. We are just waiting for someone to implement that change, and we'd be happy to review it.

Thanks.

PR #9399, opened 2021-02-23
[upgrade]: Introduce --dependency-update argument

oliverbaehler:

This Pull Request introduces the `--dependency-update ` to the upgrade cmd. With this addition it will be possible to upgrade dependencies before upgrading a helm release. The argument will also be passed to the install cmd when using the `--install` argument in addition, where the flag is already present. 

The PR was split into two seperat PRs. For further information see:

  * #9335 
  
**If applicable**:

### DIFF
--- a/cmd/helm/upgrade.go
+++ b/cmd/helm/upgrade.go
@@ -148,7 +148,7 @@ func newUpgradeCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 			// Only check dependencies if there are any
 			if req := ch.Metadata.Dependencies; req != nil {
 				if client.DependencyUpdate {
-					// Update all dependencies if DependencyUpdate is true
+					// Update all dependencies if DependencyUpdate flag is set (Implemented with PR#9399)
 					if ch, err = action.UpdateDependencies(chartPath, client.ChartPathOptions.Keyring, settings, out, p); err != nil {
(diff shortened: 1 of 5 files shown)


cmd/helm/upgrade.go
```diff
@@ -142,7 +145,37 @@ func newUpgradeCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 			if err != nil {
 				return err
 			}
+
+			// Only check dependencies if there are any
 			if req := ch.Metadata.Dependencies; req != nil {
+				// Update all dependencies if DependencyUpdate is true
+				if client.DependencyUpdate {
```
bacongobbler, line 152: Since this code is identical to the code in `helm install`, can you please refactor this out into a separate function? That way the code can remain [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).

bacongobbler (changes requested):

cmd/helm/upgrade.go
oliverbaehler: @bacongobbler Thanks for the Review. I tried my best with my newbie go knowledge.. :-). Note: I did not include the client in the function, since the Struct differs between the Install and Upgrade Class and that would have required some changes I wasn't confident doing.

PR #8011, opened 2020-04-27
feat: lint the names of templated resources

technosophos:
**What this PR does / why we need it**:

This adds a linter rule to check that resource names conform to the Kubernetes requirements. 
See #7672 for details.

**Special notes for your reviewer**:

This PR does make the decision that the `name` field is _required_. There are a few cases where Kubernetes does not strictly require the name field, but in those cases we would risk breaking Helm if we did not supply names. So I made names required.

**If applicable**:
- [ ] this PR contains documentation
- [x] this PR contains unit tests
- [x] this PR has been tested for backwards compatibility

Closes #7672 

Signed-off-by: Matt Butcher <matt.butcher@microsoft.com>

PR #9416, opened 2021-02-25
Add name validation rules for different object kinds

### LINKED ISSUE #9197: Helm lint fails due to ":" character in .metadata.name

croomes:
fix #9197 and #9203

ref
https://kubernetes.io/docs/reference/access-authn-authz/rbac/#default-roles-and-role-bindings
https://github.com/kubernetes/kubernetes/blob/585fc57cc7aaf73baf6aef84bf6dab65bde0ef9e/pkg/apis/rbac/validation/validation.go#L32

This is an alternative to https://github.com/helm/helm/pull/9218 which has some feedback but appears to have stalled.  Please feel free to reject this if you prefer to pursue that instead.

The name validation change that went into https://github.com/helm/helm/releases/tag/v3.3.0 in https://github.com/helm/helm/pull/8011 doesn't take into account the different validation rules used by the different apis.

For us, the linter is failing on the RBAC rules but I can see other Kinds are also affected.

I had a good look and I can't find any way to use a higher-level api-machinery or equivalent function to validate in a generic way - even [ValidateObjectMeta](https://pkg.go.dev/k8s.io/apimachinery@v0.20.2/pkg/api/validation#ValidateObjectMeta) needs to be passed the validation function to use.  I can't find any way to set these without comparing the object's Kind. I've not used ValidateObjectMeta directly as it checks other things besides the name.

Personally, I feel this approach is brittle but I'd like to do my part to push this closer to a resolution.  I'm not aware of the motivation behind the original PR but my vote would be to revert it.  This is the best I could come up with that matches the intended behaviour.

I've left the old validation function in the utils package as it is public.  I can remove it if requested - no Helm code is using it.

**If applicable**:

### DIFF
--- a/pkg/lint/rules/template.go
+++ b/pkg/lint/rules/template.go
@@ -215,7 +215,7 @@ func validateMetadataName(obj *K8sYamlStruct) error {
 		allErrs = append(allErrs, field.Invalid(field.NewPath("metadata").Child("name"), obj.Metadata.Name, msg))
 	}
 	if len(allErrs) > 0 {
-		return fmt.Errorf("object name does not conform to Kubernetes naming requirements: %q (%s)", obj.Metadata.Name, allErrs.ToAggregate().Error())
+		return errors.Wrapf(allErrs.ToAggregate(), "object name does not conform to Kubernetes naming requirements: %q", obj.Metadata.Name)
 	}
 	return nil
 }
--- a/pkg/lint/rules/template.go
+++ b/pkg/lint/rules/template.go
@@ -28,6 +28,9 @@ import (
 	"strings"
 
 	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/api/validation"
+	apipath "k8s.io/apimachinery/pkg/api/validation/path"
+	"k8s.io/apimachinery/pkg/util/validation/field"
 	"k8s.io/apimachinery/pkg/util/yaml"
 
 	"helm.sh/helm/v3/pkg/chart/loader"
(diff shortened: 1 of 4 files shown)


croomes:
@bacongobbler is this closer to what you're looking for?

pkg/lint/rules/template.go
```diff
@@ -202,18 +205,70 @@ func validateYamlContent(err error) error {
...
-	// name is left empty.
-	if err := chartutil.ValidateMetadataName(obj.Metadata.Name); err != nil {
-		return errors.Wrapf(err, "object name does not conform to Kubernetes naming requirements: %q", obj.Metadata.Name)
+
+	allErrs := field.ErrorList{}
+	for _, msg := range fn(obj.Metadata.Name, false) {
+		allErrs = append(allErrs, field.Invalid(field.NewPath("metadata").Child("name"), obj.Metadata.Name, msg))
+	}
+	if len(allErrs) > 0 {
+		return fmt.Errorf("object name does not conform to Kubernetes naming requirements: %q (%s)", obj.Metadata.Name, allErrs.ToAggregate().Error())
```
adamreese, line 222: This could be changed to use `pkg/errors`

```
return errors.Wrapf(allErrs.ToAggregate(), "object name does not conform to Kubernetes naming requirements: %q", obj.Metadata.Name)
```

pkg/lint/rules/template.go
```diff
@@ -202,18 +205,70 @@ func validateYamlContent(err error) error {
...
+		// No validation.
+		// https://github.com/kubernetes/kubernetes/blob/v1.20.0/pkg/apis/certificates/validation/validation.go#L137-L140
+		return func(name string, prefix bool) []string { return nil }
+	case "role", "clusterrole", "rolebinding", "clusterrolebinding":
+		// https://github.com/kubernetes/kubernetes/blob/v1.20.0/pkg/apis/rbac/validation/validation.go#L32-L34
+		return func(name string, prefix bool) []string {
+			return apipath.IsValidPathSegmentName(name)
+		}
+	default:
+		return nil
```
adamreese, line 268: Should this return `validation.NameIsDNSSubdomain` to remove the nil check in the calling function?

adamreese (changes requested):
Thanks for doing all this research!

pkg/lint/rules/template.go
jdolitsky: or `func(name string, prefix bool) []string { return nil }` ?

jdolitsky (commented):
Does this mean `chartutil.ValidateMetadataName` is to be deprecated?

pkg/lint/rules/template.go
croomes: I've gone with Adam's suggestion as any new object kinds are likely to use `validation.NameIsDNSSubdomain`.  Failures would be noticed immediately vs silently not validating.  Happy to take guidance on this.

pkg/lint/rules/template.go
croomes: Good catch, fixed

croomes:
> Does this mean `chartutil.ValidateMetadataName` is to be deprecated?

Yes, I think it can be.  I've added a deprecation notice and updated the previous notice.

adamreese (approved):

croomes:
Thanks @adamreese!

Issue #8768, opened 2020-09-18
Feature: ready command to report the readiness of a release


joejulian:
The code used in the wait loop could be moved out of the loop and exported to be used as a `ready` command:

```bash
$ helm ready --namespace foo myapp
false
```

Additional options _could_ be verbose and machine readable output that reports the readiness details for each of the resources.

As users of the helm API, we have a project that installs sets of charts that are all configured to work together. Currently, we have to use wait to know the state of the installation or upgrade. If we didn't have to wait, we could expect eventual consistency and poll for readiness while doing other work.

joejulian:
@bacongobbler I'm not sure how it can be a plugin without changes to `helm.sh/helm/v3/pkg/kube`. The readiness check code that is in the wait loop would need to be moved out of the loop (probably into a utils file) and enhanced to give a list of results (in addition to bool and error) and made public.

Can you give a little more guidance on this?

bacongobbler:
When we talk about implementing new feautures in Helm, we have to consider whether the community would see value in it Otherwise we end up with a bloated Helm client with features re-implemented in different solutions. We try to follow the UNIX approach of doing one thing well.

I would suggest implementing a `helm ready` plugin first to see if there's utility in de-coupling `--wait` as its own feature.

> I'm not sure how it can be a plugin without changes to `helm.sh/helm/v3/pkg/kube`

A plugin could re-implement the code present in pkg/kube without having to import it straight from Helm. If there are parts of the codebase that can safely be refactored into its own package, that is also an option. This is how experiments like a plugin helps improve the usability overall - we can determine what needs refactoring, and how to slice it up so clients can re-implement core parts of Helm.

Hope this helps provide some guidance.

joejulian:
I'm working on this.

joelanford:
I just mentioned in https://github.com/operator-framework/operator-sdk/issues/4102#issuecomment-717315042 that we need this feature for the Operator SDK Helm operator. Would love to see it in the core Helm SDK.

joejulian:
My idea is to split the readiness check into its own function. In the copy that I forked for our internal project, that function optionally takes a flag and return the readiness of every resource instead of exiting early if any resource is not ready.

I suspected that I should probably do this as two PRs, one just moving the tests into a function without changing any behavior, and one that adds the new behavior.

joejulian:
I should have some time again soon (release push happening now).

joejulian:
/reopen

... doesn't work, does it...

[...]

PR #9425, opened 2021-02-28
Add ReadyChecker to decouple ready check logic from --wait 

joelanford:
Signed-off-by: Joe Lanford <joe.lanford@gmail.com>

This PR helps solve #8768, and it enables users of the SDK to perform the same readiness check that the `--wait` flag does without actually waiting for everything to be ready.

If I made this change correctly, it should not change any functionality with existing `--wait` logic.

**If applicable**:

### DIFF
--- a/pkg/kube/ready.go
+++ b/pkg/kube/ready.go
@@ -1,0 +1,397 @@
+/*
+Copyright The Helm Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package kube // import "helm.sh/helm/v3/pkg/kube"
+
+import (
+	"context"
+
+	appsv1 "k8s.io/api/apps/v1"
+	appsv1beta1 "k8s.io/api/apps/v1beta1"
+	appsv1beta2 "k8s.io/api/apps/v1beta2"
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


pkg/kube/ready.go
```diff
@@ -0,0 +1,397 @@
...
+		}
+	case *appsv1.Deployment, *appsv1beta1.Deployment, *appsv1beta2.Deployment, *extensionsv1beta1.Deployment:
+		currentDeployment, err := c.client.AppsV1().Deployments(v.Namespace).Get(ctx, v.Name, metav1.GetOptions{})
+		if err != nil {
+			return false, err
+		}
+		// If paused deployment will never be ready
+		if currentDeployment.Spec.Paused {
+			return c.pausedAsReady, nil
+		}
```
joelanford, line 120: This is a slight change for increased flexibility. Since the `ReadyChecker` is now a standalone type, it probably makes sense for the default behavior of the `ReadyChecker` to return `false` for paused deployments.

For the purposes of `kube.Client.Wait()`, `c.pausedAsReady` is overridden set to `true` so that existing behavior is preserved.

pkg/kube/ready.go
```diff
@@ -0,0 +1,397 @@
...
+		// things will always return false unless one of the objects
+		// that manages pods has been hit
+		ok  = true
+		err error
+	)
+	switch value := AsVersioned(v).(type) {
+	case *corev1.Pod:
+		pod, err := c.client.CoreV1().Pods(v.Namespace).Get(ctx, v.Name, metav1.GetOptions{})
+		if err != nil || !c.isPodReady(pod) {
+			return false, err
```
joejulian, line 103: This is really close to what I have. One difference is that here I just wrap this error in some context and append it to a list so I can add a verbose flag to the command and return a list of all the reasons a release is not ready.

joejulian (approved):

pkg/kube/ready.go
joelanford: @joejulian and I discussed this offline. We came to the conclusion that looping over the release resources and building a readiness report could be done out-of-tree in a plugin, and that this repo only needs the the lower-level functionality to check readiness for an individual resource, which is what the new `ReadyChecker` struct provides.

technosophos (approved):
LGTM

mattfarina (approved):

joejulian:
:tada:

PR #9514, opened 2021-03-19
fix defaultIngress template

### LINKED ISSUE #9485: Ingress syntax mismatch, helm 3.5.3 and kubernetes 19.x
Output of `helm version`: 

Output of `kubectl version`:
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.2", GitCommit:"f5743093fd1c663cb0cbc89748f730662345d44d", GitTreeState:"clean", BuildDate:"2020-09-16T13:32:58Z", GoVersion:"go1.15", Compiler:"gc", Platform:"linux/amd64"}

minikube version: v1.13.1

Helm documentation, supported kubernetes version:
https://helm.sh/docs/topics/version_skew/

While deploying a helm chart to a local minikube cluster, ingress seems to have an outdated syntax in <CHART-NAME>/templates/ingress.yaml, resulting in following error message:
"Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: [ValidationError(Ingress.spec.rules[0].http.paths[0].backend): unknown field "serviceName" in io.k8s.api.networking.v1.IngressBackend, ValidationError(Ingress.spec.rules[0].http.paths[0].backend): unknown field "servicePort" in io.k8s.api.networking.v1.IngressBackend]"

Beside this i was also required to change the following part in <CHART-NAME>/templates/ingress.yaml
(at row 5)
from:
```yaml
```
to:
```yaml
```

However, changes according to the following tip:
https://stackoverflow.com/a/64126069/517457
Referencing:
https://kubernetes.io/docs/concepts/services-networking/ingress/#the-ingress-resource

And changing in <CHART-NAME>/templates/ingress.yaml, does acctually make the above helm install error go away.
(starting at row 33)
from:
```yaml
paths:
          {{- range .paths }}
          - path: {{ .path }}
            backend:
              serviceName: {{ $fullName }}
              servicePort: {{ $svcPort }}
```

to:
```yaml
paths:
          {{- range .paths }}
          - path: {{ .path }}
            pathType: Prefix
            backend:
              service:
                name: {{ $fullName }}
                port:
                  number: {{ $svcPort }}
```

yxxhero:

fix https://github.com/helm/helm/issues/9485

**If applicable**:

### DIFF
--- a/pkg/chartutil/create.go
+++ b/pkg/chartutil/create.go
@@ -216,8 +216,12 @@ const defaultIgnore = `# Patterns to ignore when building packages.
 const defaultIngress = `{{- if .Values.ingress.enabled -}}
 {{- $fullName := include "<CHARTNAME>.fullname" . -}}
 {{- $svcPort := .Values.service.port -}}
-{{- if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion -}}
+{{- if eq "true" (include "<CHARTNAME>.ingressApiVersionChanged" .) }}
+{{- if semverCompare "<1.19-0" .Capabilities.KubeVersion.GitVersion -}}
 apiVersion: networking.k8s.io/v1beta1
+{{- else -}}
+apiVersion: networking.k8s.io/v1
+{{- end }}
 {{- else -}}
 apiVersion: extensions/v1beta1
 {{- end }}
@@ -242,6 +246,22 @@ spec:
     {{- end }}
   {{- end }}
   rules:
+    {{- if eq "true" (include "<CHARTNAME>.ingressApiVersionChanged" .) }}
+    {{- range .Values.ingress.hosts }}
+    - host: {{ .host | quote }}
+      http:
+        paths:
+          {{- range .paths }}
+          - path: {{ .path }}
+            pathType: Prefix
+            backend:
+              service:
+                name: {{ $fullName }}
+                port:
+                  number: {{ $svcPort }}
+          {{- end }}
+    {{- end }}
+    {{- else }}
     {{- range .Values.ingress.hosts }}
     - host: {{ .host | quote }}
       http:
@@ -253,7 +273,8 @@ spec:
               servicePort: {{ $svcPort }}
           {{- end }}
     {{- end }}
-  {{- end }}
+    {{- end }}
+{{- end }}
 `
 
 const defaultDeployment = `apiVersion: apps/v1
@@ -434,6 +455,17 @@ Create chart name and version as used by the chart label.
 */}}
 {{- define "<CHARTNAME>.chart" -}}
 {{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+According to the kubernetes version, judge whether the content of ingress needs to be updated.
+*/}}
+{{- define "<CHARTNAME>.ingressApiVersionChanged" -}}
+{{- if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion }}
+{{- printf "%s" "true" }}
+{{- else }}
+{{- printf "%s" "false" }}
+{{- end }}
 {{- end }}
 
 {{/*


pkg/chartutil/create.go
```diff
@@ -216,10 +216,13 @@ const defaultIgnore = `# Patterns to ignore when building packages.
 const defaultIngress = `{{- if .Values.ingress.enabled -}}
 {{- $fullName := include "<CHARTNAME>.fullname" . -}}
 {{- $svcPort := .Values.service.port -}}
-{{- if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion -}}
+{{- $capabilities := .Capabilities -}}
+{{- if semverCompare "<1.14-0" $capabilities.KubeVersion.GitVersion -}}
+apiVersion: extensions/v1beta1
+{{- else if and (semverCompare ">=1.14-0" $capabilities.KubeVersion.GitVersion) (semverCompare "<1.19-0" $capabilities.KubeVersion.GitVersion) -}}
```
bacongobbler, line 222: This can be compressed to `else if semverCompare "~1.18-0"`. There's no need for the `semverCompare ">=1.14-0"`: the prior `if` branch will cover all other cases.

pkg/chartutil/create.go
```diff
@@ -216,10 +216,13 @@ const defaultIgnore = `# Patterns to ignore when building packages.
 const defaultIngress = `{{- if .Values.ingress.enabled -}}
 {{- $fullName := include "<CHARTNAME>.fullname" . -}}
 {{- $svcPort := .Values.service.port -}}
-{{- if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion -}}
+{{- $capabilities := .Capabilities -}}
```
bacongobbler, line 219: why is this necessary? I don't see a reason why we need to assign it to a variable.

pkg/chartutil/create.go
```diff
@@ -248,9 +251,18 @@ spec:
         paths:
           {{- range .paths }}
           - path: {{ .path }}
+            {{- if semverCompare "<1.14-0" $capabilities.KubeVersion.GitVersion }}
```
bacongobbler, line 254: If there are multiple locations hardcoding this capability check, we should wrap this in a function call so that the code is more [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).

bacongobbler (changes requested):

pkg/chartutil/create.go
KalleJohansson: I do like the idea of using a function, other than that it all seems ok to me.

yxxhero:
@bacongobbler  I have been thinking for a long time, and have not found a very elegant way to solve this problem. Looking forward to your guidance. Thanks very much.

pkg/chartutil/create.go
```diff
@@ -216,9 +216,13 @@ const defaultIgnore = `# Patterns to ignore when building packages.
...
-{{- if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion -}}
+{{- if eq "true" (include "<CHARTNAME>.ingressApiVersionChanged" .) }}
+{{- if semverCompare "<1.19-0" .Capabilities.KubeVersion.GitVersion -}}
 apiVersion: networking.k8s.io/v1beta1
 {{- else -}}
+apiVersion: networking.k8s.io/v1
+{{- end }}
+{{- else -}}
 apiVersion: extensions/v1beta1
 {{- end }}
```
sagikazarmark, line 227: How about the following?

```
{{- if semverCompare ">=1.19-0" .Capabilities.KubeVersion.GitVersion -}}
{{- else if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion -}}
{{- else -}}
{{- end }}
```

pkg/chartutil/create.go
```diff
@@ -242,6 +246,22 @@ spec:
     {{- end }}
   {{- end }}
   rules:
+    {{- if eq "true" (include "<CHARTNAME>.ingressApiVersionChanged" .) }}
```
sagikazarmark, line 249:
```suggestion
    {{- if semverCompare ">=1.19-0" .Capabilities.KubeVersion.GitVersion -}}
```

pkg/chartutil/create.go
```diff
@@ -436,6 +457,17 @@ Create chart name and version as used by the chart label.
...
+According to the kubernetes version, judge whether the content of ingress needs to be updated.
+*/}}
+{{- define "<CHARTNAME>.ingressApiVersionChanged" -}}
+{{- if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion }}
+{{- printf "%s" "true" }}
+{{- else }}
+{{- printf "%s" "false" }}
+{{- end }}
+{{- end }}
+
```
sagikazarmark, line 470:
```suggestion
```

sagikazarmark (commented):
I think this PR can be simplified quite a bit (unless I'm missing something).

bacongobbler:
fixed in favour of #9621.

Issue #9348, opened 2021-02-09
Helm dep update is not respecting the "version" stipulated in the requirements.


mdebarros:
Helm dep update is not respecting the "version" stipulated in the parent requirements file. 

Parent chart's requirement.yaml example:
```
- name: child-chart
  version: 1.0.1
  repository: "file://../child-chart"
  condition: child-chart.enabled
```

Child chart's chart.yaml example:
```
apiVersion: v1
description: child-chart
name: child-chart
version: 1.0.0
appVersion: "1.0"
```

When running `helm dep up`, the parent chart will include the child chart in the charts folder, i.e.

parent/charts/child-charts-1.0.0.tgz <-- This should not happen, and the `helm dep up` should respect the version stipulated in the requirements.yaml `v1.0.1` and thus should fail. When using helm v2.17.0, and helm v3.3.0, the `helm dep up` would fail in this manner.

When running `helm install parent`, the Installation will excluding the missing child-chart as there is no `parent/charts/child-charts-1.0.1.tgz` available. There is no way to easily identify if there is a missing dependency as the installation is "successful" even though one of the dependencies are missing.

Output of `helm version`:

Experienced this issues originally with v3.4.2:
```
version.BuildInfo{Version:"v3.4.2", GitCommit:"23dd3af5e19a02d4f4baa5b2f242645a1a3af629", GitTreeState:"dirty", GoVersion:"go1.15.5"}
```

Upgraded to latest stable via Brew and experienced same issue:
```
version.BuildInfo{Version:"v3.5.2", GitCommit:"167aac70832d3a384f65f9745335e9fb40169dc2", GitTreeState:"dirty", GoVersion:"go1.15.7"}
```

Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.3", GitCommit:"1e11e4a2108024935ecfcb2912226cedeafd99df", GitTreeState:"clean", BuildDate:"2020-10-14T18:49:28Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"11", GitVersion:"v1.11.6", GitCommit:"b1d75deca493a24a2f87eb1efde1a569e52fc8d9", GitTreeState:"clean", BuildDate:"2018-12-16T04:30:10Z", GoVersion:"go1.10.3", Compiler:"gc", Platform:"linux/amd64"}
```

[...]

PR #9527, opened 2021-03-23
#9348 the metadata is always true

Akiros001:

**If applicable**:

### DIFF
--- a/internal/version/version.go
+++ b/internal/version/version.go
@@ -53,9 +53,6 @@ type BuildInfo struct {
 
 // GetVersion returns the semver string of the version
 func GetVersion() string {
-	if metadata == "" {
-		return version
-	}
 	return version + "+" + metadata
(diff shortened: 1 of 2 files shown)


internal/version/version.go
```diff
@@ -53,9 +53,6 @@ type BuildInfo struct {
 
 // GetVersion returns the semver string of the version
 func GetVersion() string {
-	if metadata == "" {
-		return version
-	}
```
mattfarina, line 58: When we have a stable release of Helm the `metadata` is an empty string. Without this case we would have versions returned such as `v3.5.3+` instead of `v3.5.3`.

Can you provide examples and cases to what you're intending?

mattfarina (changes requested):

bacongobbler:
closing as inactive.

PR #9536, opened 2021-03-25
validate release name during install

### LINKED ISSUE #9339: Helm does not validate release name during install, but validates it during uninstall
Output of `helm version`:
`helm.sh/helm/v3 v3.5.0` (from `go.mod`)

Local

Scenario:

1. While writing a UT to test `helm install` and `helm uninstall` from my Go program, I repeatedly hit the error: `'uninstall: Release name is invalid: testNamespace-chart-name-id1-id2`.

2. The install went fine with no errors.

3. Looking at the source for `helm uninstall`, I see [here](https://github.com/helm/helm/blob/master/pkg/action/uninstall.go#L66-L68) that Helm validates the release name to [conform to the regex](https://github.com/helm/helm/blob/master/pkg/chartutil/validate_name.go#L30) `[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*`.

4. A similar check is present during an `upgrade` and `rollback`, but absent during an `install`.

5. I removed the camel-casing in my release name and things work fine.

Expected behaviour:

If the release name is invalid, an error should be encountered immediately while installing the release and not while uninstalling it.

mengjiao-liu:
Helm validate release name during install

fix #9339

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -54,13 +54,6 @@ import (
 	"helm.sh/helm/v3/pkg/storage/driver"
 )
 
-// releaseNameMaxLen is the maximum length of a release name.
-//
-// As of Kubernetes 1.4, the max limit on a name is 63 chars. We reserve 10 for
-// charts to add data. Effectively, that gives us 53 chars.
-// See https://github.com/helm/helm/issues/1528
-const releaseNameMaxLen = 53
-
 // NOTESFILE_SUFFIX that we want to treat special. It goes through the templating engine
 // but it's not a yaml file (resource) hence can't have hooks, etc. And the user actually
 // wants to see this file after rendering in the status command. However, it must be a suffix
@@ -458,14 +451,10 @@ func (i *Install) failRelease(rel *release.Release, err error) (*release.Release
 //	- used by a deleted release, and i.Replace is false
 func (i *Install) availableName() error {
 	start := i.ReleaseName
-	if start == "" {
-		return errors.New("name is required")
-	}
 
-	if len(start) > releaseNameMaxLen {
-		return errors.Errorf("release name %q exceeds max length of %d", start, releaseNameMaxLen)
+	if err := chartutil.ValidateReleaseName(start); err != nil {
+		return errors.Wrapf(err, "release name %q", start)
 	}
-
 	if i.DryRun {
 		return nil
 	}
--- a/pkg/chartutil/validate_name.go
+++ b/pkg/chartutil/validate_name.go
@@ -52,7 +52,12 @@ var (
 )
 
 const (
-	// maxNameLen is the maximum length Helm allows for a release name
+	// According to the Kubernetes docs (https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#rfc-1035-label-names)
+	// some resource names have a max length of 63 characters while others have a max
+	// length of 253 characters. As we cannot be sure the resources used in a chart, we
+	// therefore need to limit it to 63 chars and reserve 10 chars for additional part to name
+	// of the resource. The reason is that chart maintainers can use release name as part of
+	// the resource name (and some additional chars).
 	maxReleaseNameLen = 53
 	// maxMetadataNameLen is the maximum length Kubernetes allows for any name.
 	maxMetadataNameLen = 253
--- a/cmd/helm/install_test.go
+++ b/cmd/helm/install_test.go
@@ -123,7 +123,7 @@ func TestInstall(t *testing.T) {
 		// Install, using the name-template
 		{
 			name:   "install with name-template",
-			cmd:    "install testdata/testcharts/empty --name-template '{{upper \"foobar\"}}'",
+			cmd:    "install testdata/testcharts/empty --name-template '{{ \"foobar\"}}'",
 			golden: "output/install-name-template.txt",
 		},
 		// Install, perform chart verification along the way.
--- a/cmd/helm/template.go
+++ b/cmd/helm/template.go
@@ -74,7 +74,7 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 			}
 
 			client.DryRun = true
-			client.ReleaseName = "RELEASE-NAME"
+			client.ReleaseName = "release-name"
 			client.Replace = true // Skip the name check
 			client.ClientOnly = !validate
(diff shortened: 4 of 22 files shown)


mengjiao-liu:
This is the first time I submitted PR at helm.
Can you review the code ？ @bacongobbler @adamreese

yxxhero (approved):
LGTM
@hickeyma @bacongobbler WDYT?

pkg/action/install.go
```diff
@@ -410,14 +403,10 @@ func (i *Install) failRelease(rel *release.Release, err error) (*release.Release
...
 func (i *Install) availableName() error {
 	start := i.ReleaseName
-	if start == "" {
-		return errors.New("name is required")
-	}
 
-	if len(start) > releaseNameMaxLen {
-		return errors.Errorf("release name %q exceeds max length of %d", start, releaseNameMaxLen)
+	if err := chartutil.ValidateReleaseName(start); err != nil {
+		return errors.Errorf("install: Release name is invalid: %s, err: %s", start, err)
```
hickeyma, line 408: Can you try changing the error return:
return errors.Errorf("install: Release name is invalid: %s, err: %s", start, err)

to the following: return errors.Wrapf(err, "release name %q is invalid", start)

It would follow how we wrap error strings usually.

pkg/chartutil/validate_name.go
```diff
@@ -37,7 +37,7 @@ var validName = regexp.MustCompile(`^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-
 
 var (
 	// errMissingName indicates that a release (name) was not provided.
-	errMissingName = errors.New("no name provided")
+	errMissingName = errors.New("name is required")
```
hickeyma, line 40: I would not change this as this is not the error but what is required.

pkg/chartutil/validate_name.go
```diff
@@ -53,6 +53,9 @@ var (
 
 const (
 	// maxNameLen is the maximum length Helm allows for a release name
+	// As of Kubernetes 1.4, the max limit on a name is 63 chars. We reserve 10 for
+	// charts to add data. Effectively, that gives us 53 chars.
+	// See https://github.com/helm/helm/issues/1528
```
hickeyma, line 58: I would replace this comment with the following as its more accurate now:

```go
\\ According to the Kubernetes docs (https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#rfc-1035-label-names)
\\ some resource names have a max length of 63 characters while others have a max
\\ length of 253 characters. As we cannot be sure the resources used in a chart, we
\\ therefore need to limit it to 63 chars and reserve 10 chars for additional part to name
\\ of the resource. The reason is that chart maintainers can use release name as part of
\\ the resource name (and some additional chars).

```

hickeyma (changes requested):
Thanks for pushing PR @mengjiao-liu. Some small comments to look at.

pkg/action/install.go
mengjiao-liu: Thanks for your advice, it has been updated.

pkg/chartutil/validate_name.go
mengjiao-liu: Yes, then there is no need to change, it has been restored.

pkg/chartutil/validate_name.go
mengjiao-liu: Okay, updated comments.

mengjiao-liu:
Request review again.  Thanks! @yxxhero @hickeyma

pkg/action/install.go
hickeyma: Looking at the error returned again during manual testing, do you mind changing to the following as it would read better:

return errors.Wrapf(err, "release name %q", start)

hickeyma (changes requested):
Nearly there. One small change to error message.

Also, one of the unit tests is failing because test was added after the initial push of the PR. It is using this golden file: https://github.com/helm/helm/blob/main/cmd/helm/testdata/output/template-with-kube-version.txt and needs to be updated. You will need to rebase to pick up the latest from main and then change the file.

mengjiao-liu:
> Nearly there. One small change to error message.
> 
> Also, one of the unit tests is failing because test was added after the initial push of the PR. It is using this golden file: https://github.com/helm/helm/blob/main/cmd/helm/testdata/output/template-with-kube-version.txt and needs to be updated. You will need to rebase to pick up the latest from main and then change the file.

Thanks for catch！Rebased and updated.

pkg/action/install.go
hickeyma: @mengjiao-liu  can you remove "is invalid""? ^^

mengjiao-liu: Ok, updated.

hickeyma (approved):
LGTM, thanks @mengjiao-liu.

Needs another core maintainer to approve before merging.

mengjiao-liu:
ping @adamreese @bacongobbler @jdolitsky 
for approval

yxxhero (approved):
LGTM

mengjiao-liu:
ping @mattfarina @bacongobbler

mengjiao-liu:
No changes, just too long ago, sync with upstream main branch.

mattfarina:
I've tested both and can see the difference. For example, see the below output...

```sh
❯ helm-pr install fooTestName ./footestname
Error: INSTALLATION FAILED: release name "fooTestName": invalid release name, must match regex ^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$ and the length must not be longer than 53

❯ helm install fooTestName ./footestname
Error: INSTALLATION FAILED: create: failed to create: Secret "sh.helm.release.v1.fooTestName.v1" is invalid: metadata.name: Invalid value: "sh.helm.release.v1.fooTestName.v1": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')
```

mattfarina (approved):

PR #9540, opened 2021-03-26
WIP: Dependency override in CLI commands

### LINKED ISSUE #2205: Proposal: Allow dynamic dependency override
**Proposal:** Enable overriding one or more 'umbrella' chart dependencies with specific sub-chart version(s) and/or sub-chart repo(s) at chart installation time.

Updated `install` command might look something like:
```
$ helm install umbrella-chart-repo/umbrella-chart \
   --set-requirement sub-chart-name.version=specific-chart-version \
   --set-requirement sub-chart-name.repository=specific-chart-repo
```
(If `sub-chart-name.repository` not provided, would use value currently existing in umbrella chart's `requirements.yaml`)

Alternatively, perhaps one could supply an entirely new `requirements.yaml` file and supply it at install time in a manner similar to `-f values.yaml`:
```
$ helm install umbrella-chart-repo/umbrella-chart \
   -r custom-requirements.yaml
```
To support this latter case, it would most likely warrant a separate ticket adding a way to inspect an umbrella chart's requirements à la `helm inspect values umbrella-chart-repo/umbrella-chart`:
```
$ helm inspect requirements umbrella-chart-repo/umbrella-chart > default-requirements.yaml

$ # modify file as necessary and save as custom-requirements.yaml

$ helm install umbrella-chart-repo/umbrella-chart \
   -r custom-requirements.yaml
```

**Considerations:** 

The details of supporting this may be tricky.  At a high level, I would imagine `helm install` would trigger a `helm dep update` behind the scenes, perhaps staging this updated (and packaged?) chart locally and then installing.  The bulk of the considerations/implications around this proposal may indeed lie in the details therein.

I would assume `helm upgrade` would be updated to mirror the functionality added to `helm install`.

**Similar Issues**:
 * https://github.com/kubernetes/helm/issues/1947 requests the addition of a `helm dependency create/add ...` command but it seems to be centered around chart build/package/publish time and so doesn't satisfy the dynamic needs of this request.
 * https://github.com/kubernetes/helm/issues/2183 requests a "development" mode wherein invocations of `helm install` would use locally staged/updated charts dynamically.  This is closer in spirit but with the fundamental difference that this proposal would enable fetching _remote_ charts as opposed to ones existing locally.  (Support for both would be ideal.)

**Use Cases**:

I'll mention the use case that triggered this feature request.  I work on CI in support of publishing and releasing an umbrella chart dependent on many sub-charts.  

For testing a PR involving a change to a given sub-chart, we first publish the sub-chart artifact and we then publish a correlating umbrella chart artifact after injecting the said sub-chart version and repo into the `requirements.yaml` and running `helm dependency update`.  Lastly, we run our E2E tests against the umbrella chart artifact.

This pipeline could be simplified by removing the need to publish a new umbrella chart artifact for every change in a given sub-chart.  We'd only need to publish the sub-chart artifact and then run our E2E tests by installing the default umbrella chart but with the updated `requirements.yaml` specified at runtime.

This use case could be extrapolated into a general enhancement/UX upgrade for dev/hacking on a given umbrella chart's dependencies.  It also supports the ability to quickly substitute one or more sub-chart versions for trouble-shooting/debugging/etc.

mladedav:
Signed-off-by: David Mládek <david.mladek.cz@gmail.com>

closes #2205
This would allow people to change during install/upgrade commands versions and repositories of their dependencies that were declared in Chart.yaml.

This is not ready at all right now, I have found out some issues with the usage as it is written now. I plan to work on it more.

**If applicable**:

### DIFF
--- a/cmd/helm/install.go
+++ b/cmd/helm/install.go
@@ -19,6 +19,7 @@ package main
 import (
 	"io"
 	"log"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
@@ -150,6 +151,7 @@ func addInstallFlags(cmd *cobra.Command, f *pflag.FlagSet, client *action.Instal
 	f.BoolVar(&client.Atomic, "atomic", false, "if set, the installation process deletes the installation on failure. The --wait flag will be set automatically if --atomic is used")
 	f.BoolVar(&client.SkipCRDs, "skip-crds", false, "if set, no CRDs will be installed. By default, CRDs are installed if not already present")
 	f.BoolVar(&client.SubNotes, "render-subchart-notes", false, "if set, render subchart notes along with the parent")
+	f.StringSliceVar(&client.Dependencies, "set-dependency", []string{}, "override versions's dependencies (can specify multiple or separate values with commas: key1=val1,key2=val2)")
 	addValueOptionsFlags(f, valueOpts)
 	addChartPathOptionsFlags(f, &client.ChartPathOptions)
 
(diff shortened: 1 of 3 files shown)


mladedav:
I will close this for the time being as the feature still needs more clarifications than I originally expected. See the linked issue for details.

PR #9564, opened 2021-04-02
Fix insecure option for chart pull/push

### LINKED ISSUE #6324: How can I push chart to an insecure OCI registry with helm v3

chenwng:

This PR is based on the work in #8477. In that PR, the original author proposed to add `insecure` and `plain-http` options to `helm chart push/pull` commands. 
This PR adds more TLS options to `helm chart pull/push` and addresses some comments from the reviewers by adding more registry client options to keep `Run` signature.
This PR also makes the `insecure-skip-tls-verify`, cert/key file options of `helm pull/install` work with oci registry.

close #6324

### DIFF
--- a/cmd/helm/root.go
+++ b/cmd/helm/root.go
@@ -153,15 +153,26 @@ func newRootCmd(actionConfig *action.Configuration, out io.Writer, args []string
 	flags.ParseErrorsWhitelist.UnknownFlags = true
 	flags.Parse(args)
 
-	registryClient, err := registry.NewClient(
-		registry.ClientOptDebug(settings.Debug),
-		registry.ClientOptWriter(out),
-		registry.ClientOptCredentialsFile(settings.RegistryConfig),
-	)
-	if err != nil {
-		return nil, err
+	cmd.PersistentPreRunE = func(cmd *cobra.Command, args []string) error {
+		registryClient, err := registry.NewClient(
+			registry.ClientOptDebug(settings.Debug),
+			registry.ClientOptWriter(out),
+			registry.ClientOptCredentialsFile(settings.RegistryConfig),
+			registry.ClientOptPlainHTTP(registry.PlainHTTPOpt),
+			registry.ClientOptInsecureSkipVerifyTLS(registry.InsecureOpt),
+			registry.ClientOptCAFile(registry.CAFileOpt),
+			registry.ClientOptCertKeyFiles(registry.CertFileOpt, registry.KeyFile),
+		)
+		if err != nil {
+			return err
+		}
+		// Used by helm install/show/upgrade/template in LocateChart
+		registry.DefaultClient = registryClient
+		// Used by helm pull/registry/chart/...
+		actionConfig.RegistryClient = registryClient
+
+		return nil
 	}
-	actionConfig.RegistryClient = registryClient
 
 	// Add subcommands
 	cmd.AddCommand(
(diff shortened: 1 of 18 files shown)


chenwng:
Hi @jdolitsky, can you please help review this?

pkg/getter/ocigetter.go
```diff
@@ -49,6 +49,28 @@ func (g *OCIGetter) get(href string) (*bytes.Buffer, error) {
 		return nil, err
 	}
 
+	// In case any TLS setting is set, recreate the registry client
+	opts := []registry.ClientOption{}
```
jdolitsky, line 53: this change mostly looks good and thanks for adding lots of tests! My main concern is the registryclient being recreated in several places.. is there some way we create it just once in the CLI configuration?

chenwng: Thanks for the comments. I added another commit to move the client creation into `PersistentPreRunE` to create the registry client just before command run with the tls options. However, I am not able to find a way to make `helm install/upgrade/...` to use the same client without modifying existing function signature. Currently, `helm pull` uses the global registry client, but `helm install/upgrade/...` use registry client created in OCIGetter and there is no way to pass the global registry client to NewOCIGetter without modifying `LocateChart`. With this, there is a bug about registry config file. If user specifies registry config file with global ` --registry-config`, `helm install/upgrade/...` won't honor it.

chenwng:
@jdolitsky , can you please take a look at the latest one? I made some changes so that the same registry client can be used by different subcommands.

cmd/helm/root.go
```diff
@@ -153,15 +153,26 @@ func newRootCmd(actionConfig *action.Configuration, out io.Writer, args []string
...
 	flags.Parse(args)
 
-	registryClient, err := registry.NewClient(
-		registry.ClientOptDebug(settings.Debug),
-		registry.ClientOptWriter(out),
-		registry.ClientOptCredentialsFile(settings.RegistryConfig),
-	)
-	if err != nil {
-		return nil, err
+	cmd.PersistentPreRunE = func(cmd *cobra.Command, args []string) error {
```
jdolitsky, line 156: Other maintainers - this is the biggest part of the change

jdolitsky (approved):
@chenwng
looks good to me. I like what you've done with cobra PersistentPreRunE

Other maintainers - is this is an acceptable approach?

bacongobbler:
@jdolitsky are there pieces here we should keep around after #9782 is merged? Looks like there is some resolver code as well as some ocigetter fixes in here.

cmd/helm/root.go
```diff
@@ -153,15 +153,26 @@ func newRootCmd(actionConfig *action.Configuration, out io.Writer, args []string
...
+		registryClient, err := registry.NewClient(
+			registry.ClientOptDebug(settings.Debug),
+			registry.ClientOptWriter(out),
+			registry.ClientOptCredentialsFile(settings.RegistryConfig),
+			registry.ClientOptPlainHTTP(registry.PlainHTTPOpt),
+			registry.ClientOptInsecureSkipVerifyTLS(registry.InsecureOpt),
```
jdolitsky, line 162: @bacongobbler - specifically I think this option needs to be carried over if is not already

lxs137:
Is there an ETA when this PR be merged?

jdolitsky (changes requested):
Change needs rebase. The entire `helm chart` subcommand has been removed since 3.7.0

PR #9577, opened 2021-04-08
Allow for helm install script to work on Windows

mburst:
This installs to the GOPATH bin folder which will work cross platform and will likely be setup on the user's path. Previously it used a standard unix path which is non-existent on Windows. As currently written it requires the user to have go installed. Since that may not be the case for all Windows users we may want to make this a fallback option. If this path is desired and will be approved I'd be happy to make that update.

To run on windows you then just need to pass the --no-sudo flag.

You're awesome!

**If applicable**:

### DIFF
--- a/scripts/get-helm-3
+++ b/scripts/get-helm-3
@@ -22,7 +22,7 @@
 : ${DEBUG:="false"}
 : ${VERIFY_CHECKSUM:="true"}
 : ${VERIFY_SIGNATURES:="false"}
-: ${HELM_INSTALL_DIR:="/usr/local/bin"}
+: ${HELM_INSTALL_DIR:="${GOPATH}/bin"}
 : ${GPG_PUBRING:="pubring.kbx"}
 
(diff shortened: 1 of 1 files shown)


bacongobbler:
> and will likely be setup on the user's path

I wouldn't assume this is always the case. Many users install Helm on operating systems without Go ever being installed. Production systems, for example.

Because of this I'm concerned this change may break compatibility for those relying on installing via a separate directory.

bacongobbler:
The better option may be to implement a PowerShell script specifically designed for Windows systems rather than attempt to conflate the two.

bacongobbler:
An alternative solution here may be to change the HELM_INSTALL_DIR on your system before calling the script. For example:

```
$ curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3
$ chmod 700 get_helm.sh
$ HELM_INSTALL_DIR=$GOPATH/bin ./get_helm.sh
```

That way no changes to the script have to be made. You just need to change the target directory on your system.

Hope that helps.

Issue #9462, opened 2021-03-11
New Ingress Template (helm new) might be wrong / or at least hard to understand


adrianliechti:
in template/ingress.yaml

the name of the service is generated by the _helper using the release name
{{- $fullName := include "chart.fullname" . -}}
{{- $svcPort := .Values.service.port -}}

```
  rules:
    {{- range .Values.ingress.hosts }}
    - host: {{ .host | quote }}
      http:
        paths:
          {{- range .paths }}
          - path: {{ .path }}
            backend:
              serviceName: {{ $fullName }}
              servicePort: {{ $svcPort }}
          {{- end }}
    {{- end }}
  {{- end }}
```

but in values.yaml, the recommended way seems to overwrite the whole path object including backend:
https://github.com/helm/helm/commit/1e126ff4d1f09b806c1952757516f6fc67506ce0

```yaml
  hosts:
    - host: chart-example.local
      paths:
      - path: /
        backend:
          serviceName: chart-example.local
          servicePort: 80
```

I'm not entirely sure, but I think this backend setting would be ignored anyway using the generated ingress template...
with the ingress v1 warning on helm install, this topic might get more important for customers

what is your recommended way now to specify a simple/minimal ingress with v1 in mind?

Thanks for clarifying,
Adrian

bacongobbler:
> what is your recommended way now to specify a simple/minimal ingress with v1 in mind?

I'm sorry, I don't fully understand your question. What do you mean by "with v1 in mind"?

Helm templates are generated based on the values provided by the user. In this particular case, the ingress rules will read what's in values.yaml and generate an ingress route. Add another route to values.yaml and the ingress routes will be updated.

adrianliechti:
with a change in the last release, the template got changed what "helm create" generates for values.yaml:
https://github.com/helm/helm/commit/1e126ff4d1f09b806c1952757516f6fc67506ce0

I do have the feeling the change was not a good idea - since the actual generated ingress (template/ingress.yaml) will not work with this suggesting in the values.yaml.

furthermore, the Kubernetes CLI throws warnings when installing/upgrading such charts generated with "helm create".
infos here: https://kubernetes.io/blog/2020/09/03/warnings/

Your template currently generates ingress in version v1beta1. In the meantime v1 has been released and changes the section quite a lot. So I think it's worth to refactor the whole helm create template for ingress

bacongobbler:
Thank you for pointing this out. PRs welcome to fix this.

[...]

PR #9585, opened 2021-04-09
Modernize `helm create` Ingress Template

adrianliechti:
> This pull request and idea is open for discussion
> It's based on this discussion: https://github.com/helm/helm/issues/9462

- fix non-working values in `helm create` template
- introduce support for Ingress v1
- support ingressClass support of K8s 1.18 
- remove support for Kubernetes 1.13 and below

TODO (as soon of a community acceptance):
- adjust unit tests & test data
- adjust documentation
- test on older clusters

### DIFF
--- a/pkg/chartutil/create.go
+++ b/pkg/chartutil/create.go
@@ -147,16 +147,13 @@ service:
 
 ingress:
   enabled: false
+  # className: nginx
   annotations: {}
-    # kubernetes.io/ingress.class: nginx
     # kubernetes.io/tls-acme: "true"
-  hosts:
+  rules:
     - host: chart-example.local
-      paths:
-      - path: /
-        backend:
-          serviceName: chart-example.local
-          servicePort: 80
+      path: /
+      pathType: ImplementationSpecific
   tls: []
   #  - secretName: chart-example-tls
   #    hosts:
@@ -214,23 +211,53 @@ const defaultIgnore = `# Patterns to ignore when building packages.
 `
 
 const defaultIngress = `{{- if .Values.ingress.enabled -}}
+{{- if .Values.ingress.enabled -}}
 {{- $fullName := include "<CHARTNAME>.fullname" . -}}
 {{- $svcPort := .Values.service.port -}}
-{{- if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion -}}
-apiVersion: networking.k8s.io/v1beta1
+{{- $kubeVersion := .Capabilities.KubeVersion.GitVersion -}}
+{{- if semverCompare ">=1.19-0" $kubeVersion -}}
+apiVersion: networking.k8s.io/v1
 {{- else -}}
-apiVersion: extensions/v1beta1
+apiVersion: networking.k8s.io/v1beta1
 {{- end }}
 kind: Ingress
 metadata:
   name: {{ $fullName }}
   labels:
     {{- include "<CHARTNAME>.labels" . | nindent 4 }}
+  {{- if and .Values.ingress.className (not (semverCompare ">=1.18-0" $kubeVersion)) }}
+    {{- if not (hasKey .Values.ingress.annotations "kubernetes.io/ingress.class") }}
+    {{- $_ := set .Values.ingress.annotations "kubernetes.io/ingress.class" .Values.ingress.className}}
+    {{- end }}
+  {{- end }}
   {{- with .Values.ingress.annotations }}
   annotations:
     {{- toYaml . | nindent 4 }}
   {{- end }}
 spec:
+  {{- if and .Values.ingress.className (semverCompare ">=1.18-0" $kubeVersion) }}
+  ingressClassName: {{ .Values.ingress.className }}
+  {{- end }}
+  rules:
+    {{- range .Values.ingress.rules }}
+    - host: {{ .host | quote }}
+      http:
+        paths:
+        - path: {{ .path }}
+          {{- if and .pathType (semverCompare ">=1.18-0" $kubeVersion) }}
+          pathType: {{ .pathType }}
+          {{- end }}
+          backend:
+            {{- if semverCompare ">=1.19-0" $kubeVersion }}
+            service:
+              name: {{ $fullName }}
+              port:
+                number: {{ $svcPort }}
+            {{- else -}}
+            serviceName: {{ $fullName }}
+            servicePort: {{ $svcPort }}
+            {{- end }}
+    {{- end }}
   {{- if .Values.ingress.tls }}
   tls:
     {{- range .Values.ingress.tls }}
@@ -241,18 +268,6 @@ spec:
       secretName: {{ .secretName }}
     {{- end }}
   {{- end }}
-  rules:
-    {{- range .Values.ingress.hosts }}
-    - host: {{ .host | quote }}
-      http:
-        paths:
-          {{- range .paths }}
-          - path: {{ .path }}
-            backend:
-              serviceName: {{ $fullName }}
-              servicePort: {{ $svcPort }}
-          {{- end }}
-    {{- end }}
   {{- end }}
 `
 
@@ -382,10 +397,8 @@ spec:
 
 const defaultNotes = `1. Get the application URL by running these commands:
 {{- if .Values.ingress.enabled }}
-{{- range $host := .Values.ingress.hosts }}
-  {{- range .paths }}
-  http{{ if $.Values.ingress.tls }}s{{ end }}://{{ $host.host }}{{ .path }}
-  {{- end }}
+{{- range $rule := .Values.ingress.rules }}
+  http{{ if $.Values.ingress.tls }}s{{ end }}://{{ .host }}{{ .path }}
 {{- end }}
 {{- else if contains "NodePort" .Values.service.type }}
   export NODE_PORT=$(kubectl get --namespace {{ .Release.Namespace }} -o jsonpath="{.spec.ports[0].nodePort}" services {{ include "<CHARTNAME>.fullname" . }})
--- a/pkg/chartutil/create.go
+++ b/pkg/chartutil/create.go
@@ -211,7 +211,6 @@ const defaultIgnore = `# Patterns to ignore when building packages.
 `
 
 const defaultIngress = `{{- if .Values.ingress.enabled -}}
-{{- if .Values.ingress.enabled -}}
 {{- $fullName := include "<CHARTNAME>.fullname" . -}}
 {{- $svcPort := .Values.service.port -}}
 {{- $kubeVersion := .Capabilities.KubeVersion.GitVersion -}}
@@ -268,7 +267,7 @@ spec:
       secretName: {{ .secretName }}
     {{- end }}
   {{- end }}
-  {{- end }}
+{{- end }}
 `
 
 const defaultDeployment = `apiVersion: apps/v1


adrianliechti:
with the release of k8s 1.21, we could consider to remove support for 1.18.
this would ease the template slightly

https://kubernetes.io/docs/setup/release/version-skew-policy/

sagikazarmark:
> with the release of k8s 1.21, we could consider to remove support for 1.18.

Sadly, 1.18 is still widely used in managed Kubernetes distributions. I understand the Kubernetes support policy, but it's simply not practical to follow when more than half of the world (or more) is still on earlier versions.

(I still receive PRs/feature requests to support versions as early as 1.14 which is insane, but the world is apparently not ready for upgrading their systems at this velocity.)

sagikazarmark:
So it looks like there is more than one, little bit hidden change in this PR:

- order of `rules` and `tls` changed (Why?)
- it's no longer possible to have multiple paths for a single rule (Why?)

adrianliechti:
one of the main questions for me is about the target group of "helm create". is it the experienced guy for a complex app with all the things (affinity, anti-affinity, node selectors... etc etc). or for the guy which got in contact with helm / k8s manifests the first time?

currently the default template is pretty crowded. i can imagine people are scared to modify it...
at another side, it's not super flexible too. there are a lot of cases where it's lacking (eg multiple services, pv/pvc...)

with multiple routes it would be still possible to route another path. of course paths could also be an optional option....

at the other side, i could imagine that forward routes to anothet service in an ingress is more common: e.g. /api to a backend, all othetd to the frontend which mostly might be a spa which wants all routes on index.html


maybe it's an idea for a parameter to helm create to
a) select the minimal target k8s version
b) allow some levels or templates (webapp, workerapp etc)

maybe even remove that from the main code and extract it to plugins

sagikazarmark:
IMHO the template serves as an example and as such, it should aim for compatibility and presenting the available options. I don't really see ingress changing much in the forseeable future, so I wouldn't mind a more detailed example that we can simplify over time (ie. remove 1.14 support in a year or something).

> with multiple routes it would be still possible to route another path. of course paths could also be an optional option....

This is actually a breaking change in the sense that it works entirely different from how it worked before. If new charts start to behave differently, we should have a good reason for that and I don't see one at the moment.

This is what I implemented for now: https://github.com/sagikazarmark/helm-charts/pull/57

And I'm already thinking about renaming `rules` to `hosts`, but I actually like the new name there, so I might keep it.

adrianliechti:
I'm more than happy to close this change in favours of yours!

Maybe i was wrong thinking that `helm create` is an important aspect in aligning and helping the helm community for best practises. There might be too many different requirements for a command without any parameters and without clear goals.
It's a huge task anyway. Thanks a ton for fixing these two issues! This will help a lot.

But based on the limited communication about changes on ingress or other aspects of the template, I think we for us have to change to another approach and create a more opinionated "create new template engine" for our customers. maybe as helm plugin or including in our cloud control plane cli. and/or recommend to start from scratch (e.g by creating resources with kubectl (it got better), and then template it for helm

cheers to you!

sagikazarmark:
I don't think you should close it. It's a good start, we should iterate on it.

pkg/chartutil/create.go
```diff
@@ -216,21 +213,50 @@ const defaultIgnore = `# Patterns to ignore when building packages.
...
 kind: Ingress
 metadata:
   name: {{ $fullName }}
   labels:
     {{- include "<CHARTNAME>.labels" . | nindent 4 }}
+  {{- if and .Values.ingress.className (not (semverCompare ">=1.18-0" $kubeVersion)) }}
+    {{- if not (hasKey .Values.ingress.annotations "kubernetes.io/ingress.class") }}
+    {{- $_ := set .Values.ingress.annotations "kubernetes.io/ingress.class" .Values.ingress.className}}
+    {{- end }}
+  {{- end }}
```
sagikazarmark, line 231: I'd move this section up to the variable declarations.

pkg/chartutil/create.go
```diff
@@ -216,21 +213,50 @@ const defaultIgnore = `# Patterns to ignore when building packages.
...
 {{- $fullName := include "<CHARTNAME>.fullname" . -}}
 {{- $svcPort := .Values.service.port -}}
-{{- if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion -}}
-apiVersion: networking.k8s.io/v1beta1
+{{- $kubeVersion := .Capabilities.KubeVersion.GitVersion -}}
+{{- if semverCompare ">=1.19-0" $kubeVersion -}}
+apiVersion: networking.k8s.io/v1
 {{- else -}}
-apiVersion: extensions/v1beta1
+apiVersion: networking.k8s.io/v1beta1
```
sagikazarmark, line 220: I'd preserve `extensions/v1beta1` support for now.

See here: https://github.com/sagikazarmark/helm-charts/blob/master/charts/tusd/templates/ingress.yaml#L14

pkg/chartutil/create.go
```diff
@@ -147,16 +147,13 @@ service:
...
     # kubernetes.io/tls-acme: "true"
-  hosts:
+  rules:
     - host: chart-example.local
-      paths:
-      - path: /
-        backend:
-          serviceName: chart-example.local
-          servicePort: 80
+      path: /
```
sagikazarmark, line 155: I'm not sure removing support for multiple paths is a good idea, although I admit I rarely supply multiple paths. Let's hear other opinions as well.

pkg/chartutil/create.go
```diff
@@ -147,16 +147,13 @@ service:
 
 ingress:
   enabled: false
+  # className: nginx
   annotations: {}
-    # kubernetes.io/ingress.class: nginx
     # kubernetes.io/tls-acme: "true"
-  hosts:
+  rules:
```
sagikazarmark, line 153: This is also a significant change in the template, but I actually like it.

sagikazarmark:
@bacongobbler can we get some eyes on this PR? Would be nice to make it part of the 3.6 release.

sagikazarmark:
@adrianliechti you might want to sign your commits so that all checks are passing.

mab:
I guess the adaption done in `create.go` needs to be done in `pkg/lint/rules/testdata/v3-fail/templates/ingress.yaml` too.

pkg/chartutil/create.go
bacongobbler: Unless this is something that was written up in a proposal and was generally accepted by the broader community, we cannot make significant changes in functionality to `helm create`'s output. Someone out there put multiple path support in here. We can't just remove that without justification.

pkg/chartutil/create.go
bacongobbler: There are multiple people still running older versions of Kubernetes. Please do not drop API extensions until Helm 4.

pkg/chartutil/create.go
bacongobbler: Breaking changes to `helm create` cannot be made as per the backwards compatibility policy. Please revert.

pkg/chartutil/create.go
```diff
@@ -147,16 +147,13 @@ service:
 
 ingress:
   enabled: false
+  # className: nginx
```
bacongobbler, line 150: what's the reason behind this comment?

pkg/chartutil/create.go
mab: So proposal would be: 
```
{{- if semverCompare ">=1.19-0" .Capabilities.KubeVersion.GitVersion -}}
{{- else if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion -}}
{{- else -}}
{{- end }}
```

pkg/chartutil/create.go
adrianliechti: it's not really a comment; it will be emitted to the values.yaml as a hint that this value is configurable

if not set, kubernetes will use the default ingress class - which might be different to nginx (e.g. contour, ambassador etc .etc.)

you see something similar on the resource hints

pkg/chartutil/create.go
adrianliechti: I am aware of the change... it's an alignment to ingress v1

helm create is a template for a chart. I do not see why there is a compatibility needed.
once created, the chart will even if you delete the create command at whole. also there is no "upgrade" functionallity.

as I wrote this change is for asking the community. I'm totally fine to close this pull request and ignore the change.

pkg/chartutil/create.go
adrianliechti: this path value changed not so lang ago and was breaking too...

not it is combined with the rule - especially since a "virtual host" usually is the hostname - and not does not include the path.

I have a limited understanding why to have the create command compatible to something helm specific and not adapt to changes at kubernetes itself... again, helm create is one-time command; In my eyes to ease accessibility for creating helm charts... it should be as easy as possible for 80% of the cases

pkg/chartutil/create.go
adrianliechti: which probably also needs a change at the ingress creation at the bottom since extensions/v1beta1 does not support a lot of the newer config.

I proposed an additional parameter for helm create to choose a minimal (or target) kubernetes version.
this makes the temple much more readable. there are currently too many if and elses blocks in the template.
you are all experienced helm users - but for new ones it looks pretty scary.

helm is currently build against kubernetes api 0.20.x - which would be theoretically not be compatible to old versions anyway... why not use old versions of helm in combination with old versions of kubernetes?

pkg/chartutil/create.go
sagikazarmark: > I do not see why there is a compatibility needed.

It's needed because people tend to follow the helm template and as a result a lot of charts look the same, making it easy for people to start using new charts. It's also very easy to overlook a change like this leading to endless hours of debugging.

pkg/chartutil/create.go
sagikazarmark: The changes in the new ingress object are not related to how Helm charts are structured, they are only semantic changes. I think the idea for keeping compatibility in this case is to keep supporting the same features as before, even on the newer ingress version. It's certainly possible to do that regardless of the breaking changes.

I also think this kind of change would require justification and I don't see any.

pkg/chartutil/create.go
sagikazarmark: > why not use old versions of helm in combination with old versions of kubernetes?

Because the helm version generating the chart is not necessarily the same as installing the chart and older Helm 3.x versions are compatible with newer charts.

pkg/chartutil/create.go
sagikazarmark: This is a new optional field allowing to customize the ingress class name. It's also backward compatible with older ingress versions, so you can use it instead of the annotation: on older Kubernetes versions it will fall back to the annotation.

adrianliechti:
if one looks at skaffolding tools like:

- https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-new
- https://maven.apache.org/guides/introduction/introduction-to-archetypes.html
- angular, react new... etc. etc.

creating a single template to fit it all seems really hard... their solution is a way to select from multiple templates.
maybe this could be a future of the helm create command (helm plugin support is already there).

the current one could be the "default". there might be a "minimal stateless web app" one, and a "stateful app with persistence"

bacongobbler:
Agreed. `helm create` was built long ago when the project hardly had 100 users. These are the tech problems that arise once you reach a certain scale.

If you're interested in proposing a rewrite of `helm create`, I would suggest looking at [The Contributing Guide under "Proposing an idea"](https://github.com/helm/helm/blob/main/CONTRIBUTING.md#proposing-an-idea). A plugin would be a great first step to prove out that idea.

adrianliechti:
thanks for this plan! this absolutely makes sense.

sorry for not doing that in the first place. it was not my intent to fall into the house with the door with breaking changes.
that's why I declared this pull request as draft and more like a pull idea.

I would recommend to close this pull request - and maybe fix the existing igress template with as few as needed changes.
my next steps will be to create some helm starterkit-templates for our customers in our own cli; and while doing that thinking about how to do this scalable in helm.

I could imagine that your library chart type feature (or an extension to it) and to leverage the existing repo download for such templates could be one of the possible rough ideas. I'm will check the contributing guide and follow the rules

wish you the best and thanks for all your time

sagikazarmark:
@bacongobbler I can open another PR with compatible changes. WDYT?

sagikazarmark:
Went ahead and opened a new PR:  #9621

PR #9595, opened 2021-04-13
Upgrade to kubernetes 1.21 / libraries 0.21.0

sbose78:
Use the latest and greatest client libraries for Kubernetes APIs

**If applicable**:


sbose78:
Would this need updates to the doc https://helm.sh/docs/topics/version_skew/ ?

pkg/action/lazyclient.go
```diff
@@ -116,6 +117,13 @@ func (s *secretClient) Patch(ctx context.Context, name string, pt types.PatchTyp
 	return s.client.CoreV1().Secrets(s.namespace).Patch(ctx, name, pt, data, opts, subresources...)
 }
 
+func (s *secretClient) Apply(ctx context.Context, secretConfiguration *applycorev1.SecretApplyConfiguration, opts metav1.ApplyOptions) (*v1.Secret, error) {
```
bacongobbler, line 120: I'm assuming these are new functions introduced to an interface somewhere. Do you have a link handy to the changes so we can investigate further?

bacongobbler:
Looks good. Just want to dive deeper into the changes to see if there's something we're missing here.

pkg/action/lazyclient.go
sbose78: https://github.com/helm/helm/blob/257a427866883a9fb745be40800ffe8e9ceda82b/pkg/action/lazyclient.go#L58

This line caught the changes :)

pkg/action/lazyclient.go
```diff
@@ -60,7 +61,7 @@ func newSecretClient(lc *lazyClient) *secretClient {
 	return &secretClient{lazyClient: lc}
 }
 
-func (s *secretClient) Create(ctx context.Context, secret *v1.Secret, opts metav1.CreateOptions) (*v1.Secret, error) {
+func (s *secretClient) Create(ctx context.Context, secret *v1.Secret, opts metav1.CreateOptions) (result *v1.Secret, err error) {
```
mattfarina, line 64: Why was the return signature changed to `(result *v1.Secret, err error)`?

mattfarina (commented):
Note, k8s 1.21.1 was just released so an update will be coming to the libs soon. They aren't available as of this comment. Just wanted to give a heads up.

adamreese (approved):

mattfarina (approved):

pkg/action/lazyclient.go
sbose78: Hi Matt,
This was changed to align with the updated signature in https://pkg.go.dev/k8s.io/client-go/kubernetes/typed/core/v1#SecretInterface

Issue #9342, opened 2021-02-08
validation: chart.metadata.version is invalid (breaking change in v3.5.2)


arcln:
Changelog for version 3.5.2 (which is a PATCH update) states the following :

> WARNING: Helm is now stricter about sanitizing data in Chart.yaml, index.yaml, and plugin.yaml. In particular, we are stricter about SemVer strings.

This definitely is a breaking change. In our CI/CD, we create deployments using `helm`. For deploying our feature branches environments, we use branch names as the chart version string. Since this update, our CI/CD is failing with the error :

```
Error: validation: chart.metadata.version "master" is invalid
```

I am not sure what measure should be applied now, as the update is already publicly available, but I though it was worth mentioning it.
This commit actually breaks semantic versioning rules while it is supposed to enforce them :^)

Output of `helm version`:

```
version.BuildInfo{Version:"v3.5.2", GitCommit:"167aac70832d3a384f65f9745335e9fb40169dc2", GitTreeState:"dirty", GoVersion:"go1.15.7"}
```

Output of `kubectl version`:

```
Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.2", GitCommit:"faecb196815e248d3ecfb03c680a4507229c2a56", GitTreeState:"clean", BuildDate:"2021-01-13T13:28:09Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.2", GitCommit:"faecb196815e248d3ecfb03c680a4507229c2a56", GitTreeState:"clean", BuildDate:"2021-01-13T13:20:00Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"linux/amd64"}
```

Cloud Provider/Platform (AKS, GKE, Minikube etc.): 

Bare metal ubuntu server, cluster installed using kubeadm.

bacongobbler:
Hi @arcln. We have documented in multiple locations the fact that branch names as chart version strings are an invalid use of the `version` field. Only valid Semver 2 versions are allowed.

As per our policy on backwards compatibility:

> Any chart that worked on a previous version of Helm 3 MUST work on a new version of Helm 3 (barring the cases where (a) Kubernetes itself changed, and (b) the chart worked because it exploited a bug)

This would be construed as the latter case of exploiting a bug.

[...]

PR #9604, opened 2021-04-14
feat(helm): optionally hide validation warnings

### LINKED ISSUE #9407: Overly Verbose Output When Encountering Old Charts With Invalid Semvers
When I do `helm repo update`, it dumps ~5000 lines saying:

```
index.go:339: skipping loading invalid entry for chart <chart name> <version> from <repo>: validation: chart.metadata.version <version> is invalid
```

A bunch of old charts in the repo have a semver with a branch name slapped on. For build reproduction (which would use the at-the-time version of helm) I can't remove those charts.

Currently, I'm neither using the chart nor using the version but am using the repo. I'd like to suppress this error message as it is causing some issues (flooding the terminal with 5000 lines takes some time and makes build logs ugly). I could do `helm repo update | grep -v "skipping loading invalid"` as a workaround but this suppression seems like something that should be `in` the codebase as a flag or the default.

I briefly looked in the docs and could not find a flag or configuration to suppress this message.

I read https://github.com/helm/helm/issues/9342#issuecomment-775269042 to understand the logic of why the thing is invalid. I agree with it. I'm not asking for it to be recognized as valid, only seeing if we can suppress the log message.

I thought about adding this as a comment to https://github.com/helm/helm/issues/9356. It is the same underlying issue but a different use case; I don't care that I can't use the chart because I'm not using the chart.

If one of the maintainers think this is a good idea, I could look into doing the (hopefully short) work myself.

Output of `helm version`:

rcousineau-xandr:
Closes #9407

when doing the following commands:
* `helm repo add`
* `helm repo update`
* `helm search repo`

the new flag `--hide-validation-warnings` will suppress the message
"skipping loading invalid entry for chart"

**If applicable**:
- [ ] this PR contains documentation - https://github.com/helm/helm-www/pull/1069

### DIFF
--- a/cmd/helm/dependency_build_test.go
+++ b/cmd/helm/dependency_build_test.go
@@ -109,7 +109,7 @@ func TestDependencyBuildCmd(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	i, err := repo.LoadIndexFile(filepath.Join(rootDir, "index.yaml"))
+	i, err := repo.LoadIndexFile(filepath.Join(rootDir, "index.yaml"), false)
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/cmd/helm/dependency_build_test.go
+++ b/cmd/helm/dependency_build_test.go
@@ -109,7 +109,7 @@ func TestDependencyBuildCmd(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	i, err := repo.LoadIndexFile(filepath.Join(rootDir, "index.yaml"), false)
+	i, err := repo.LoadIndexFile(filepath.Join(rootDir, "index.yaml"))
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/cmd/helm/dependency_update_test.go
+++ b/cmd/helm/dependency_update_test.go
@@ -96,7 +96,7 @@ func TestDependencyUpdateCmd(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	i, err := repo.LoadIndexFile(dir(helmpath.CacheIndexFile("test")))
+	i, err := repo.LoadIndexFile(dir(helmpath.CacheIndexFile("test")), false)
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/cmd/helm/dependency_update_test.go
+++ b/cmd/helm/dependency_update_test.go
@@ -96,7 +96,7 @@ func TestDependencyUpdateCmd(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	i, err := repo.LoadIndexFile(dir(helmpath.CacheIndexFile("test")), false)
+	i, err := repo.LoadIndexFile(dir(helmpath.CacheIndexFile("test")))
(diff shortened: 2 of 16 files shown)


bacongobbler (changes requested):
We cannot accept changes to public function signatures in Helm 3. To be specific, the changes to `LoadIndexFile` and `DownloadIndexFile` break backwards compatibility. Because of that we cannot accept these changes as-is.

https://github.com/helm/helm/blob/main/CONTRIBUTING.md#semantic-versioning

rcousineau-xandr:
@bacongobbler any ideas for a different way to deal with this?

https://github.com/helm/helm/blob/a499b4b179307c267bdf3ec49b880e3dbd2a5591/pkg/repo/index.go#L339

I could maybe create a copy of `LoadIndexFile` called `LoadIndexFileNoValidate` or something?

bacongobbler:
That log message was explicitly added to bubble up a security issue. We will not be accepting PRs to bypass or otherwise ignore the error message entirely. I explained why in https://github.com/helm/helm/issues/9342#issuecomment-775269042.

rcousineau-xandr:
> The resolution is simple

Unfortunately we can't just *simply* remove all of these packages (nearly ~3k for our repo).

I understand not wanting to completely eliminate the message, and I'm just trying to discuss a backwards-compatible approach to optionally suppressing it.

Proposal: if `--hide-validation-warnings` is passed, `LoadIndexFileNoValidate` is called instead of `LoadIndexFile`. 

No existing interface changes, so it's backwards-compatible.

rcousineau-xandr:
> security issue

I think it's worth mentioning that we're not talking about suppressing validation of charts that we're publishing from this point forward. We have charts in this repository that were created with helm 2.x and were valid at the time.

nathschmidt:
@bacongobbler Now that @rcousineau-xandr has ensured these changes are backwards compatible, could you revisit this PR? The verbosity of this logging is causing pretty significant issues for me - particularly when it comes to reading otherwise very small logs.

I'd also like to echo the comments of @rcousineau-xandr - removing all of the offending charts is not an option, in all or many cases. These charts were valid when published and in many cases are still being used with Helm 2.

bacongobbler:
Since we cannot determine that there is any difference between what are considered “safe” and “unsafe” versions, disabling the warnings means you are now vulnerable to malicious attackers injecting malicious version numbers into any chart repository index. To do so would revert back to Helm’s prior behaviour, and the reason why the auditors raised this issue in the first place.

Because of that, we cannot accept this “feature”.

rcousineau-xandr:
@bacongobbler can you please take a look at my alternative approach: #9611 -- if we could condense the multitude of errors into a summary (instead of hiding it altogether) that would help immensely.

pierluigilenoci:
Still an interesting feature

benoittgt:
I think it might make sense to close this one regardless of https://github.com/helm/helm/pull/9604#issuecomment-849260843, and instead focus our efforts on https://github.com/helm/helm/pull/9611.

PR #9611, opened 2021-04-15
feat(helm): consolidate validation warnings

### LINKED ISSUE #9407: Overly Verbose Output When Encountering Old Charts With Invalid Semvers

rcousineau-xandr:
Closes #9407

when doing the following commands:
* `helm repo add`
* `helm repo update`
* `helm search repo`

during validation of repository, invalid charts are counted and a
single warning is printed:

    skipped loading 2964 invalid chart entries from www.helm.repo

also, new `helm repo update --show-all-warnings` flag that shows a warning for each invalid chart

**If applicable**:

### DIFF
--- a/cmd/helm/repo_update.go
+++ b/cmd/helm/repo_update.go
@@ -41,7 +41,7 @@ To update all the repositories, use 'helm repo update'.
 var errNoRepositories = errors.New("no repositories found. You must add one before updating")
 
 type repoUpdateOptions struct {
-	showAllWarnings      bool
+	showValidationErrors bool
 	update               func([]*repo.ChartRepository, io.Writer, bool, bool) error
 	repoFile             string
 	repoCache            string
@@ -75,7 +75,7 @@ func newRepoUpdateCmd(out io.Writer) *cobra.Command {
 	// This should be deprecated in Helm 4 by update to the behaviour of `helm repo update` command.
 	f.BoolVar(&o.failOnRepoUpdateFail, "fail-on-repo-update-fail", false, "update fails if any of the repository updates fail")
 
-	f.BoolVar(&o.showAllWarnings, "show-all-warnings", false, "show every invalid chart while indexing repository")
+	f.BoolVar(&o.showValidationErrors, "show-repo-validation-errors", false, "show every invalid chart while indexing repository")
 
 	return cmd
 }
@@ -114,10 +114,10 @@ func (o *repoUpdateOptions) run(out io.Writer, args []string) error {
 		}
 	}
 
-	return o.update(repos, out, o.failOnRepoUpdateFail, o.showAllWarnings)
+	return o.update(repos, out, o.failOnRepoUpdateFail, o.showValidationErrors)
 }
 
-func updateCharts(repos []*repo.ChartRepository, out io.Writer, failOnRepoUpdateFail bool, showAllWarnings bool) error {
+func updateCharts(repos []*repo.ChartRepository, out io.Writer, failOnRepoUpdateFail bool, showValidationErrors bool) error {
 	fmt.Fprintln(out, "Hang tight while we grab the latest from your chart repositories...")
 	var wg sync.WaitGroup
 	var repoFailList []string
@@ -126,7 +126,7 @@ func updateCharts(repos []*repo.ChartRepository, out io.Writer, failOnRepoUpdate
 		go func(re *repo.ChartRepository) {
 			defer wg.Done()
 			var err error
-			if showAllWarnings {
+			if showValidationErrors {
 				_, err = re.ValidateAndDownloadIndexFile()
 			} else {
 				_, err = re.DownloadIndexFile()
--- a/cmd/helm/repo_update.go
+++ b/cmd/helm/repo_update.go
@@ -41,7 +41,8 @@ To update all the repositories, use 'helm repo update'.
 var errNoRepositories = errors.New("no repositories found. You must add one before updating")
 
 type repoUpdateOptions struct {
-	update               func([]*repo.ChartRepository, io.Writer, bool) error
+	showAllWarnings      bool
+	update               func([]*repo.ChartRepository, io.Writer, bool, bool) error
 	repoFile             string
 	repoCache            string
 	names                []string
@@ -64,7 +65,7 @@ func newRepoUpdateCmd(out io.Writer) *cobra.Command {
 			o.repoFile = settings.RepositoryConfig
 			o.repoCache = settings.RepositoryCache
 			o.names = args
-			return o.run(out)
+			return o.run(out, args)
 		},
 	}
 
@@ -74,10 +75,12 @@ func newRepoUpdateCmd(out io.Writer) *cobra.Command {
 	// This should be deprecated in Helm 4 by update to the behaviour of `helm repo update` command.
 	f.BoolVar(&o.failOnRepoUpdateFail, "fail-on-repo-update-fail", false, "update fails if any of the repository updates fail")
 
+	f.BoolVar(&o.showAllWarnings, "show-all-warnings", false, "show every invalid chart while indexing repository")
+
 	return cmd
 }
 
-func (o *repoUpdateOptions) run(out io.Writer) error {
+func (o *repoUpdateOptions) run(out io.Writer, args []string) error {
 	f, err := repo.LoadFile(o.repoFile)
 	switch {
 	case isNotExist(err):
@@ -111,10 +114,10 @@ func (o *repoUpdateOptions) run(out io.Writer) error {
 		}
 	}
 
-	return o.update(repos, out, o.failOnRepoUpdateFail)
+	return o.update(repos, out, o.failOnRepoUpdateFail, o.showAllWarnings)
 }
 
-func updateCharts(repos []*repo.ChartRepository, out io.Writer, failOnRepoUpdateFail bool) error {
+func updateCharts(repos []*repo.ChartRepository, out io.Writer, failOnRepoUpdateFail bool, showAllWarnings bool) error {
 	fmt.Fprintln(out, "Hang tight while we grab the latest from your chart repositories...")
 	var wg sync.WaitGroup
 	var repoFailList []string
@@ -122,7 +125,13 @@ func updateCharts(repos []*repo.ChartRepository, out io.Writer, failOnRepoUpdate
 		wg.Add(1)
 		go func(re *repo.ChartRepository) {
 			defer wg.Done()
-			if _, err := re.DownloadIndexFile(); err != nil {
+			var err error
+			if showAllWarnings {
+				_, err = re.ValidateAndDownloadIndexFile()
+			} else {
+				_, err = re.DownloadIndexFile()
+			}
+			if err != nil {
 				fmt.Fprintf(out, "...Unable to get an update from the %q chart repository (%s):\n\t%s\n", re.Config.Name, re.Config.URL, err)
 				repoFailList = append(repoFailList, re.Config.URL)
 			} else {
--- a/cmd/helm/repo_update_test.go
+++ b/cmd/helm/repo_update_test.go
@@ -35,7 +35,7 @@ func TestUpdateCmd(t *testing.T) {
 	var out bytes.Buffer
 	// Instead of using the HTTP updater, we provide our own for this test.
 	// The TestUpdateCharts test verifies the HTTP behavior independently.
-	updater := func(repos []*repo.ChartRepository, out io.Writer, failOnRepoUpdateFail bool) error {
+	updater := func(repos []*repo.ChartRepository, out io.Writer, failOnRepoUpdateFail bool, showAllWarnings bool) error {
 		for _, re := range repos {
 			fmt.Fprintln(out, re.Config.Name)
 		}
@@ -45,7 +45,7 @@ func TestUpdateCmd(t *testing.T) {
 		update:   updater,
 		repoFile: "testdata/repositories.yaml",
 	}
-	if err := o.run(&out); err != nil {
+	if err := o.run(&out, []string{}); err != nil {
 		t.Fatal(err)
 	}
 
@@ -60,7 +60,7 @@ func TestUpdateCmdMultiple(t *testing.T) {
 	var out bytes.Buffer
 	// Instead of using the HTTP updater, we provide our own for this test.
 	// The TestUpdateCharts test verifies the HTTP behavior independently.
-	updater := func(repos []*repo.ChartRepository, out io.Writer, failOnRepoUpdateFail bool) error {
+	updater := func(repos []*repo.ChartRepository, out io.Writer, failOnRepoUpdateFail bool, showAllWarnings bool) error {
 		for _, re := range repos {
 			fmt.Fprintln(out, re.Config.Name)
 		}
@@ -71,7 +71,7 @@ func TestUpdateCmdMultiple(t *testing.T) {
 		repoFile: "testdata/repositories.yaml",
 		names:    []string{"firstexample", "charts"},
 	}
-	if err := o.run(&out); err != nil {
+	if err := o.run(&out, []string{}); err != nil {
 		t.Fatal(err)
 	}
 
@@ -86,7 +86,7 @@ func TestUpdateCmdInvalid(t *testing.T) {
 	var out bytes.Buffer
 	// Instead of using the HTTP updater, we provide our own for this test.
 	// The TestUpdateCharts test verifies the HTTP behavior independently.
-	updater := func(repos []*repo.ChartRepository, out io.Writer, failOnRepoUpdateFail bool) error {
+	updater := func(repos []*repo.ChartRepository, out io.Writer, failOnRepoUpdateFail bool, showAllWarnings bool) error {
 		for _, re := range repos {
 			fmt.Fprintln(out, re.Config.Name)
 		}
@@ -97,7 +97,7 @@ func TestUpdateCmdInvalid(t *testing.T) {
 		repoFile: "testdata/repositories.yaml",
 		names:    []string{"firstexample", "invalid"},
 	}
-	if err := o.run(&out); err == nil {
+	if err := o.run(&out, []string{}); err == nil {
 		t.Fatal("expected error but did not get one")
 	}
 }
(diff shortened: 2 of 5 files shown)


rcousineau-xandr:
@bacongobbler here's an alt approach that might be better than https://github.com/helm/helm/pull/9604

rcousineau-xandr:
@bacongobbler have you had a chance to look at this?

pierluigilenoci:
@rcousineau-xandr maybe you should do a PR rebase. 

@adamreese
@bacongobbler
@hickeyma
@jdolitsky
@marckhouzam
@mattfarina
@prydonius
@SlickNik
@technosophos

would anyone be so kind and helpful to take a look at this PR?

schollii:
Over 2 months since @pierluigilenoci, is there any hope to get this approved, is there anything I can do to help?

pierluigilenoci:
@schollii I'm not a maintainer of this repo so I can't do anything more than tag all the maintainers. 😉

schollii:
@pierluigilenoci I know, that's why I referenced your post, not much more we can do about this PR but flag people.

cc @bacongobbler @mattfarina @technosophos

bacongobbler:
Wayyyyyy too busy with other work to review this right now. I'll throw this onto the 3.7.0 milestone for reiew with the hopes that someone else might find the time to review this.

bacongobbler:
One quick note: I am concerned that this masks _what_ chart entries are masked. The user has no idea whether any chart entries they are attempting to pull in are part of that dataset or not.

> skipped loading 2964 invalid chart entries from www.helm.repo

rcousineau-xandr:
if one needs the individual invalid chart names, the current behavior is preserved with
```
helm repo update --show-all-warnings
```

schollii:
@bacongobbler Good point about the charts being pulled in. How about the warning only applies to charts that are not pulled in? So if in a helm install you are pulling in 3 charts, only those 3 would get warnings, regardless of the `--show-all-warnings` setting. Then if you really want all warnings even for charts that are not being used, use the flag. 

I suggest suggest also that the flag be more specific, like `--warn-all-invalid-charts-in-repo`. This is a bit verbose but it is clear what it is for, and if there are ever more types of warnings that should come to life, this will not interfere. 

@rcousineau-xandr these should be small mods, if you don't have time I could try (based on responses to above suggestions).

pkg/repo/index.go
```diff
@@ -324,23 +324,30 @@ func IndexDirectory(dir, baseURL string) (*IndexFile, error) {
...
 	}
 
+	invalidCharts := 0
 	for name, cvs := range i.Entries {
 		for idx := len(cvs) - 1; idx >= 0; idx-- {
 			if cvs[idx].APIVersion == "" {
 				cvs[idx].APIVersion = chart.APIVersionV1
 			}
 			if err := cvs[idx].Validate(); err != nil {
-				log.Printf("skipping loading invalid entry for chart %q %q from %s: %s", name, cvs[idx].Version, source, err)
```
rcousineau-xandr, line 339: @schollii is there an indicator here of whether these are new? this is just loading everything from `source`...

schollii: Doesn't look like it. This function is a pure index-reader, the args dont' have the charts that will be used. 

I tried a helm install here on a chart that is listed as skipped: 
```$ helm install junk myrepo/bad-chart-2018.20.3-0956887.tgz 
...
index.go:339: skipping loading invalid entry for chart "bad-chart" "2018.20.3-0956887"
 from /home/user/.cache/helm/repository/myrepo-index.yaml: 
validation: chart.metadata.version "2018.20.3-0956887" is invalid
Error: failed to download "myrepo/bad-chart-2018.20.3-0956887.tgz" 
(hint: running `helm repo update` may help)
Done
```
In the "failed to download" error message, the hint is incorrect and misleading. If there is a way to return a list of skipped charts from that index function, and make it available to where that download error is shown, then the error could be something like `hint: chart version is invalid, rerun command with --warn-invalid-charts-in-index for details`. 

The flag `consolidateWarnings` is a little vague, can it be renamed to `warnInvalidChartsInIndex`?

rcousineau-xandr: > The flag `consolidateWarnings` is a little vague, can it be renamed to `warnInvalidChartsInIndex`?

agreed! was waiting for a suggestion of a better name 😀

rcousineau-xandr: > I tried a helm install here on a chart that is listed as skipped:
> 
> ```
> $ helm install junk myrepo/bad-chart-2018.20.3-0956887.tgz 
> ...
> index.go:339: skipping loading invalid entry for chart "bad-chart" "2018.20.3-0956887"
>  from /home/user/.cache/helm/repository/myrepo-index.yaml: 
> validation: chart.metadata.version "2018.20.3-0956887" is invalid
> Error: failed to download "myrepo/bad-chart-2018.20.3-0956887.tgz" 
> (hint: running `helm repo update` may help)
> Done
> ```
> 
> In the "failed to download" error message, the hint is incorrect and misleading. 

Splitting hairs a bit, but it's not necessarily incorrect/misleading in general for this command to say that `helm repo update` ***may*** help. Just not correct in this case. 

The error you posted does include:

```
validation: chart.metadata.version "2018.20.3-0956887" is invalid
```

So it seems pretty clear to me.

schollii: I did not say that the hint is always wrong, I said it is wrong in this situation where the chart is one of the skipped ones (because updating the repo will not solve the issue). 

Does that "validation" line get printed if your warning flag is off? The point of @bacongobbler was that if hiding the warning that validation error, then the user has no way of knowing what's happened. If that validation error does get printed even with warning off, then that should address @bacongobbler concern and the only change left is a good name for that flag.

rcousineau-xandr: oh, I misunderstood! I think you're right that

```
validation: chart.metadata.version "2018.20.3-0956887" is invalid
```

was printed as part of the index update and in my case (and others) it would be accompanied by thousands of similar validation errors. those would indeed get consolidated into

```
skipped loading 2964 invalid chart entries from www.helm.repo
```

which is unfortunately also what you'd see if you followed the suggestion of running `helm repo update` (without a flag to enable all validation errors)

perhaps the best (also easiest) way to solve this is for the consolidated message to be something like

```
skipped loading 2964 invalid chart entries from www.helm.repo, run `helm repo update --show-all-warnings` to see each invalid entry
```

schollii: Yeah that sounds good, if the user sees that summary message and they see the "not downloaded" message right after, they have the info they need I think to dig deeper. 

Are you going to rename the flag to something like `--warn-invalid-charts-in-index`?

rcousineau-xandr: > Are you going to rename the flag to something like `--warn-invalid-charts-in-index`?

To reiterate, the current default behavior is that (while indexing) the following message is printed for *each* chart/version that fails validation

```
validation: chart.metadata.version "2018.20.3-0956887" is invalid
```

The new default behavior will be to instead show

```
skipped loading 2964 invalid chart entries...
```

And the flag is meant to expand the default summary message into the full set of validation failure messages.

I'm not sure what to call these: "chart validation errors"?

I'm considering something like `--enumerate-validation-errors`.

schollii:
Well it can't be "enumerate" unless each error has a number. Also I think it needs to be specific to index (although technically it is not the index that is incorrect, it is the chart version, but it's a gray area). So how about `--show-index-validation-errors, or just `--show-index-errors`?

rcousineau-xandr:
> Well it can't be "enumerate" unless each error has a number.

> **e·nu·mer·ate**
> /əˈn(y)o͞oməˌrāt/
> mention (a number of things) one by one.
> "there is not space to enumerate all his works"

> Also I think it needs to be specific to index (although technically it is not the index that is incorrect, it is the chart version, but it's a gray area). So how about `--show-index-validation-errors, or just `--show-index-errors`?

It happens while indexing the repo, it is not validation of the index that fails.

`--show-repo-validation-errors` ?

schollii:
I've been using Python for too long I guess ;) 

I vote for `--show-repo-validation-errors`

schollii:
@bacongobbler so IINM based on what we found the consolidated warning will make it clear that the final error about chart not downloaded is due to bad chart. If the chart downloads, then it was not one of the bad ones, and most users will not care about the details. Those that do, can re-run with the flag `--show-repo-validation-errors`.

pierluigilenoci:
@schollii maybe make sense to resolve conflicts. I still have the secret hope that this PR will come to an end. 😜

pierluigilenoci:
@schollii could you please take a look?

schollii (commented):
Can you point me to where the new functionality is tested? I'm seeing that tests were updated with the extra parameter false meaning test without consolidating warning or showing them. Sorry if they are there.

rcousineau-xandr:
@schollii okay I've added a test

schollii:
@rcousineau-xandr i will build and run. But I'm trying to remember how to reproduce this, do you remember how the invalid charts ended up in the helm registry? was it the public registry or a private registry in S3?

rcousineau-xandr:
@schollii in my company's private helm repository on an instance of artifactory we have thousands of charts that were published prior to the helm 3.5.2 update that added the requirement that the `version` be semver compliant. so using a helm 3.5.2+ client and doing `helm repo update`, we get thousands of messages printed to the console, each one stating that an individual chart's version is invalid.

jdmaguire:
To add some more context @schollii, before 3.5.2, helm could package and push helm charts with invalid semvers. [This was considered an exploitation of a bug](https://github.com/helm/helm/issues/9342#issuecomment-775269042) that was fixed in 3.5.2.

To get an invalid helm chart packaged/pushed, you either need to use an old helm version & helm push or say curl to push directory to a helm repository that doesn't do validation on incoming pushes.

It quite common to accidentally not have a valid semver and hence a lot of people using pre-3.5.2 Helm could have been generating these invalids versions for years without knowing.

schollii:
Yeah that makes sense. I no longer use registry I install chart straight from git repo. So I'll have to pass on that part of the PR, and sounds like you have a good way to confirm it works at command line.

schollii:
Good job @rcousineau and thanks for taking the time. 

So now we just need official approval.

pierluigilenoci:
@mattfarina are there any chances of seeing this PR merged?

justinmchase (approved):

justinmchase:
Please, someone from the team drive this forward! Its a daily source of annoyance for me and trying to script around it requires so much added complexity. There is no reason at all this should be taking this much time or brain power. Just stop emitting the dang useless warnings please.

sp-ludovic-ivain:
would love to have this PR merged too 🥺 🙏

pierluigilenoci:
@adamreese
  @bacongobbler
  @hickeyma
  @jdolitsky
  @marckhouzam
  @mattfarina
  @sabre1041
  @scottrigby
  @SlickNik
  @technosophos

Can any of you help out?

justinmchase:
merge

pierluigilenoci:
@schollii could you please take a look?

schollii:
Indeed. I wonder what it takes to be a maintainer. I would love to but unfortunately I don't have the time.

justinmchase:
I'd be willing to be a maintainer and help drive this forward, if anyone here had that power.

pierluigilenoci:
@adamreese @bacongobbler @hickeyma @jdolitsky @marckhouzam @mattfarina @sabre1041 @scottrigby @SlickNik @technosophos are there any chances to see this PR merged?

jdmaguire:
Looking at the [`contributing.md` ](https://github.com/helm/helm/blob/main/CONTRIBUTING.md) file, the PR does need a `awaiting review` label @rcousineau-xandr. Maybe that will get some attention :)

rcousineau-xandr:
It already was reviewed though...?

jdolitsky (approved):

cmd/helm/repo_update.go
```diff
@@ -74,10 +75,12 @@ func newRepoUpdateCmd(out io.Writer) *cobra.Command {
 	// This should be deprecated in Helm 4 by update to the behaviour of `helm repo update` command.
 	f.BoolVar(&o.failOnRepoUpdateFail, "fail-on-repo-update-fail", false, "update fails if any of the repository updates fail")
 
+	f.BoolVar(&o.showValidationErrors, "show-repo-validation-errors", false, "show every invalid chart while indexing repository")
+
 	return cmd
 }
 
-func (o *repoUpdateOptions) run(out io.Writer) error {
```
hickeyma, line 80: Why have you changed the function signature?

cmd/helm/repo_update.go
```diff
@@ -41,7 +41,8 @@ To update all the repositories, use 'helm repo update'.
 var errNoRepositories = errors.New("no repositories found. You must add one before updating")
 
 type repoUpdateOptions struct {
-	update               func([]*repo.ChartRepository, io.Writer, bool) error
+	showValidationErrors bool
```
hickeyma, line 44: Add below `failOnRepoUpdateFail`

cmd/helm/repo_update.go
```diff
@@ -64,7 +65,7 @@ func newRepoUpdateCmd(out io.Writer) *cobra.Command {
 			o.repoFile = settings.RepositoryConfig
 			o.repoCache = settings.RepositoryCache
 			o.names = args
-			return o.run(out)
+			return o.run(out, args)
```
hickeyma, line 68: No need to add args to run

cmd/helm/repo_update_test.go
```diff
@@ -216,7 +216,50 @@ func TestUpdateChartsFailWithError(t *testing.T) {
...
+	r, err := repo.NewChartRepository(&repo.Entry{
+		Name: "charts",
+		URL:  invalidURL,
+	}, getter.All(settings))
+	if err != nil {
+		t.Error(err)
+	}
+
+	b := bytes.NewBuffer(nil)
+	err = updateCharts([]*repo.ChartRepository{r}, b, true, true)
```
hickeyma, line 262: Would it be good idea to check the error here to confirm its a consolidated error message?

pkg/repo/chartrepo.go
```diff
@@ -116,6 +116,32 @@ func (r *ChartRepository) Load() error {
...
 func (r *ChartRepository) DownloadIndexFile() (string, error) {
+	return r.downloadIndexFile(true)
+}
+
+// ValidateAndDownloadIndexFile fetches the index without consolidating validation warnings.
+func (r *ChartRepository) ValidateAndDownloadIndexFile() (string, error) {
+	return r.downloadIndexFile(false)
+}
+
+// Index generates an index for the chart repository and writes an index.yaml file.
```
hickeyma, line 127: Reverse the logic for backwards compatibility

pkg/repo/index_test.go
```diff
@@ -147,13 +147,13 @@ func TestLoadIndex(t *testing.T) {
 
 // TestLoadIndex_Duplicates is a regression to make sure that we don't non-deterministically allow duplicate packages.
 func TestLoadIndex_Duplicates(t *testing.T) {
-	if _, err := loadIndex([]byte(indexWithDuplicates), "indexWithDuplicates"); err == nil {
+	if _, err := loadIndex([]byte(indexWithDuplicates), "indexWithDuplicates", false); err == nil {
```
hickeyma, line 150: Test for consolidated message as well?

hickeyma (changes requested):
Thanks for pushing this feature @rcousineau-xandr. Apologies for the delay in reviewing.

The main thing the stands out is that the PR breaks backwards compatibility - in other words it changes the expected output. Consolidating the error messages maybe the best approach going forward but this change can only occur in a major release i.e. Helm 4. Therefore, the logic should be reversed - a flag should be aded to consolidate the messages and be false by default. It is looking good apart from that. 

I have also added some comments inline.

justinmchase:
> ... in other words it changes the expected output.

Does it though? I think we're saying here, this output is not expected (or wanted).

hickeyma:
> @hickeyma Does it though? I think we're saying here, this output is not expected (or wanted).

The point is that this has been the expected output for Helm 3.x. If we want to change it it needs to be in a major release i.e. Helm 4. The PR can be pushed to Helm 4 if you like or otherwise, you can change the flag to consolidate the messages.

justinmchase:
Makes sense, making the flag backward compatible and the change opt-in for v3 makes sense to me.

rcousineau-xandr:
@justinmchase do you have any desire to make the requested changes? I'm a little... fatigued... by all of this. If not, I'll try to get around to it, but I have no guarantees when that'll happen.

schollii:
@hickeyma I can't imagine why anyone in the past 22 months would be relying, in their scripts or processes, on the presence of this flood of warnings. 

But if we have to abide by a hardline that no output shall be removed on minor/patch release, then so be it, I think the change to the PR will be minimal. 

I will try to get around to making it

hickeyma:
@jdmaguire @schollii The backwards compatibility rules are documented [here](https://github.com/helm/community/blob/main/hips/hip-0004.md) if you would like to take a look. Before I joined the community, backwards compatibility was always adhered to and taken seriously by maintainers. In rare cases, it may have required to break it because of no other option or in the case of a bug which would be rolled back when found.

PRs need to be reviewed properly adhering to the guidelines of the community.

schollii:
Ok I'll try to get this past finish

justinmchase:
I will try to help if you don't get to it sooner, I've just been slammed the last couple weeks coincidentally but it would definitely be a welcome improvement for many in my team.

pierluigilenoci:
It's a shame that after all this work, this PR has stalled and ended up forgotten.
@rcousineau-xandr, is this still relevant to you? Can you try and make it work?

jdmaguire:
It is sad that in the name of backwards compatibility, we can't restore some backwards compatibility.

pierluigilenoci:
giphy

Issue #9618, opened 2021-04-18
Data races in the chartutil package


devzx:
Hello there!

Firstly let me preface this by saying a massive thank you to everyone who works on this fantastic project.

To add a bit of background context on this particular issue:
At work we utilize `helm` for all our kubernetes deployments, as such we have bit a fair amount of tooling around analyzing our releases before carrying out the actual deployment. One of those tools makes use of various helm packages to essentially run a `helm template` to generate the would be release manifest we use that data to parse out various bits we are interested in such as the image to perform further checks e.g the image passes vulnerability scans etc. This tool runs on every PR raised to our deployment repository as such its very likely that the code is executed concurrently.

I've created a repository https://github.com/devzx/helm-data-race which provides code to recreate the data race as well as the code I have written in an accompanying PR (modified in the fix/vendor folder for convenience) that aims to resolve the issue.

The actual issue seems to occur when https://github.com/helm/helm/blob/main/pkg/chartutil/capabilities.go#L37-L40 these two defaults are called the same time (my assumption on why this hasn't been flagged before is because helm is usually used to execute a single command and then exits). Because they are package level variables that are pointers subsequent calls either retain data from the previous call (I noticed this in the tests, api version data from a previous test was persisted for the following tests leading to incorrect golden files) or data races occur when `APIVersions` are modified concurrently here https://github.com/helm/helm/blob/a499b4b179307c267bdf3ec49b880e3dbd2a5591/pkg/action/install.go#L200

```==================
WARNING: DATA RACE
Read at 0x000004123f90 by goroutine 20:
  helm.sh/helm/v3/pkg/action.(*Install).Run()
      /Users/eli/go/pkg/mod/helm.sh/helm/v3@v3.5.3/pkg/action/install.go:200 +0x249
  main.template()
      /Users/eli/git/helm-data-race/main.go:84 +0x67b

Previous write at 0x000004123f90 by goroutine 22:
  helm.sh/helm/v3/pkg/action.(*Install).Run()
      /Users/eli/go/pkg/mod/helm.sh/helm/v3@v3.5.3/pkg/action/install.go:200 +0x337
  main.template()
      /Users/eli/git/helm-data-race/main.go:84 +0x67b

Goroutine 20 (running) created at:
  main.main()
      /Users/eli/git/helm-data-race/main.go:25 +0xed

Goroutine 22 (running) created at:
  main.main()
      /Users/eli/git/helm-data-race/main.go:25 +0xed
==================
```

[...]

PR #9619, opened 2021-04-18
Fix potential data races in chartutil

devzx:
Hello!

This PR attempts to fix the data race issues raised in https://github.com/helm/helm/issues/9618.
The code in this PR is also demonstrated here https://github.com/devzx/helm-data-race under the fix/vendor folder

The main changes are:
- Replaced package level variables used as defaults in the chartutil package that were susceptible to data races with functions and replaced all references with the new functions
- Updated incorrect golden files that had been generated as a result of using the aforementioned package level variables which contained `kubernetes-api-version` data that was persisted from previous tests.
-  Fixed more data races within the `allKnownVersions()` function which was using a package level variable from kubernetes scheme pacakge, now we create and manage our own instance of it per call.

### DIFF
--- a/pkg/chartutil/capabilities.go
+++ b/pkg/chartutil/capabilities.go
@@ -19,6 +19,9 @@ import (
 	"fmt"
 	"strconv"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	runtime "k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/kubernetes/scheme"
 
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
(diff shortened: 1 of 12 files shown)


pkg/chartutil/capabilities.go
```diff
@@ -32,22 +35,6 @@ const (
 	k8sVersionMinor = 20
 )
 
-var (
-	// DefaultVersionSet is the default version set, which includes only Core V1 ("v1").
-	DefaultVersionSet = allKnownVersions()
```
bacongobbler, line 37: so unfortunately we need to keep these in for the time being. As per our contributing policy, we try not to make any breaking changes to the API until the next major release. Many users may not be as concerned about race conditions, but they would be concerned about their apps breaking in a 3.x release!

I would suggest adding these back, but then adding a comment above informing the reader that this has been deprecated in favour of `DefaultVersionSet()`. Same for `DefaultCapabilities`.

pkg/chartutil/capabilities.go
```diff
@@ -92,13 +79,56 @@ func allKnownVersions() VersionSet {
 	// We should register the built in extension APIs as well so CRDs are
 	// supported in the default version set. This has caused problems with `helm
 	// template` in the past, so let's be safe
-	apiextensionsv1beta1.AddToScheme(scheme.Scheme)
-	apiextensionsv1.AddToScheme(scheme.Scheme)
+	s := runtime.NewScheme()
```
bacongobbler, line 82: So one of the main reasons why we were using the upstream variable was to ensure we picked up new schemas as they became available in future Kubernetes versions.

This now looks like we're attempting to reconstruct our own schema, but that comes with the risk that should a new schema comes up, it will not be added.

Perhaps a better approach might be to propose a new function upstream that will recreate `scheme.Scheme` exactly as-is so that we don't need to maintain all of this cruft to rebuild the same schema from scratch.

PR #9621, opened 2021-04-19
Ingress v1 template

### LINKED ISSUE #9485: Ingress syntax mismatch, helm 3.5.3 and kubernetes 19.x

sagikazarmark:

Improve ingress template to add support for new ingress features added in Kubernetes 1.18 and 1.19:

- `networking.k8s.io/v1` support
- `ingressClassName` field support

### DIFF
--- a/pkg/chartutil/create.go
+++ b/pkg/chartutil/create.go
@@ -147,16 +147,15 @@ service:
 
 ingress:
   enabled: false
+  className: ""
   annotations: {}
     # kubernetes.io/ingress.class: nginx
     # kubernetes.io/tls-acme: "true"
   hosts:
     - host: chart-example.local
       paths:
-      - path: /
-        backend:
-          serviceName: chart-example.local
-          servicePort: 80
+        - path: /
+          pathType: ImplementationSpecific
   tls: []
   #  - secretName: chart-example-tls
   #    hosts:
@@ -216,7 +215,14 @@ const defaultIgnore = `# Patterns to ignore when building packages.
 const defaultIngress = `{{- if .Values.ingress.enabled -}}
 {{- $fullName := include "<CHARTNAME>.fullname" . -}}
 {{- $svcPort := .Values.service.port -}}
-{{- if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion -}}
+{{- if and .Values.ingress.className (not (semverCompare ">=1.18-0" .Capabilities.KubeVersion.GitVersion)) }}
+  {{- if not (hasKey .Values.ingress.annotations "kubernetes.io/ingress.class") }}
+  {{- $_ := set .Values.ingress.annotations "kubernetes.io/ingress.class" .Values.ingress.className}}
+  {{- end }}
+{{- end }}
+{{- if semverCompare ">=1.19-0" .Capabilities.KubeVersion.GitVersion -}}
+apiVersion: networking.k8s.io/v1
+{{- else if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion -}}
 apiVersion: networking.k8s.io/v1beta1
 {{- else -}}
 apiVersion: extensions/v1beta1
@@ -231,6 +237,9 @@ metadata:
     {{- toYaml . | nindent 4 }}
   {{- end }}
 spec:
+  {{- if and .Values.ingress.className (semverCompare ">=1.18-0" .Capabilities.KubeVersion.GitVersion) }}
+  ingressClassName: {{ .Values.ingress.className }}
+  {{- end }}
   {{- if .Values.ingress.tls }}
   tls:
     {{- range .Values.ingress.tls }}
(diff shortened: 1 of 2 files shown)


pkg/chartutil/create.go
```diff
@@ -147,16 +147,15 @@ service:
 
 ingress:
   enabled: false
+  className:
```
adrianliechti, line 150: would it be an idea to # this?
it's optional (because of ingress default class)

do you prefer to keep the annotation hint for kubernetes.io/ingress.class anyway?

sagikazarmark: > would it be an idea to # this?

It has the same effect, but I think it's better this way.

> do you prefer to keep the annotation hint for kubernetes.io/ingress.class anyway?

That's an interesting question. I don't mind keeping it.

adrianliechti: when comparing to serviceaccount -> name, this might be another approach for the sake of consistency:  
same effect - but includes note for the customer.

```
# If not set, the default ingress class is used
className: ""
```

pkg/chartutil/create.go
```diff
@@ -147,16 +147,15 @@ service:
...
     # kubernetes.io/ingress.class: nginx
     # kubernetes.io/tls-acme: "true"
   hosts:
     - host: chart-example.local
       paths:
-      - path: /
-        backend:
-          serviceName: chart-example.local
-          servicePort: 80
+        - path: /
```
mab, line 157: indentation seems to be incorrect.

sagikazarmark: Why? What's wrong with it?

adrianliechti: in most of the case node's children are intended by two characters (see hosts -> host).
on paths -> path it was not; márk fixed that (also on other parts of the file). it looks good for me

mab (approved):

pkg/chartutil/create.go
sagikazarmark: Updated the default to string. Thanks!

sagikazarmark:
@bacongobbler do you think this fits into the 3.6 milestone?

bacongobbler (approved):
LGTM. Needs another maintainer to sign off on the changes as per the contributing policy.

dunglas (approved):

sagikazarmark:
Today is the tentative date for RC1. Since this is a feature change, it should be merged before RC1 is tagged.

Also, the workflow run needs approval so that tests can be run by the CI.

bacongobbler:
I've gone ahead and approved the workflow. You'll have to reach out to the other core maintainers and find someone with enough time on their hands to review this.

sagikazarmark:
Thanks @bacongobbler !

technosophos (approved):
I am not a fan of the trend of making create's logic increasingly complicated. It was never my intent that the default chart would be anything other than a bare-bones scaffold. I am worried that with version-specific code, we are going to start seeing templates here that are so complicated that nobody will be able to maintain them or understand what they do. As a default experience, that is bad.

That said... we can't remove the ingress code during Helm 3's lifecycle. So apparently we _need_ to add this kind of logic. So I am approving this.

PR #9640, opened 2021-04-22
Add error check to avoid panic

wangchenglong01:
Add error check to avoid panic

If `err` is not nil, the `watcher` will be nil. So the following code calling `watcher.ResultChan()` will cause panic.

**If applicable**:


wangchenglong01:
@helm-bot Nobody reviews?

bacongobbler (approved):

PR #9644, opened 2021-04-23
Support linting generateName

### LINKED ISSUE #9488: Error when linting 'Job' that uses `generateName` instead of `name` in metadata
### Environment
Output of `helm version`:
```
```
Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.0", GitCommit:"02c110006bfef4ba53fa5042bb9eae170dd3dc1c", GitTreeState:"clean", BuildDate:"2021-01-09T01:48:22Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.0+e405995", GitCommit:"e40599583c035332dc295d9ea2c52e7635d48a6e", GitTreeState:"clean", BuildDate:"2021-02-05T01:23:59Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"linux/amd64"}
```

Cloud Provider/Platform (AKS, GKE, Minikube etc.):  OpenShift 4.6.17

### Problem
When linting a Helm chart that defines a `Job` resource (`batch/v1`) which defines `metadata.generateName` instead of `metadata.name` linting fails with an error:
```
[ERROR] templates/job.yaml: object name must be between 0 and 253 characters: ""
```
The resource definition looks like this:
```
apiVersion: batch/v1
kind: Job
metadata:
  generateName: test-
spec:
  template:
    spec:
      containers:
        - command:
          - "/bin/sh"
          - "-c"
          - |
            echo Hello World
          image: "alpine:latest"
          name: "echo"
      restartPolicy: "OnFailure"
```
According to the documentation `generateName` can be used instead of `name`:
> generateName <string>
>   GenerateName is an optional prefix, used by the server, to generate a
>   unique name ONLY IF the Name field has not been provided. If this field is
>   used, the name returned to the client will be different than the name
>   passed. This value will also be combined with a unique suffix. The provided
>   value has the same validation rules as the Name field, and may be truncated
>   by the length of the suffix required to make the value unique on the
>   server.
>
>   If this field is specified and the generated name exists, the server will
>   NOT return a 409 - instead, it will either return 201 Created or 500 with
>   Reason ServerTimeout indicating a unique name could not be found in the
>   time allotted, and the client should retry (optionally after the time
>   indicated in the Retry-After header).
>
>   Applied only if Name is not specified. More info:
>   https://git.k8s.io/community/contributors/devel/sig-architecture/api-conventions.md#idempotency

Source: [ObjectMeta documentation](https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/object-meta/#ObjectMeta)

In our case we are using this in combination with ArgoCD sync hooks.

According to the documentation it seems that `generateName` might be valid for other resources as well. Is there a reason why this is not supported in Helm? If so, would it maybe be possible to change this to a warning instead of an error?

### Work-around
The current work-around is to use an older helm version (e.g. 3.2.4) where this resource doesn't fail linting.

cgetzen:
Signed-off-by: Charlie Getzen <charliegetzenlc@gmail.com>

Closes https://github.com/helm/helm/issues/9488

Ensures that linting works when `generateName` is used instead of `name`.

**If applicable**:

### DIFF
--- a/pkg/lint/rules/template.go
+++ b/pkg/lint/rules/template.go
@@ -32,6 +32,7 @@ import (
 	apipath "k8s.io/apimachinery/pkg/api/validation/path"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"k8s.io/apimachinery/pkg/util/yaml"
+	"k8s.io/apiserver/pkg/storage/names"
 
 	"helm.sh/helm/v3/pkg/chart/loader"
 	"helm.sh/helm/v3/pkg/chartutil"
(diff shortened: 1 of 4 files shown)


mattfarina:
First, let me say that I want to support `generateName`. It would be nice to have it work well in Helm. But, there are some checks in helm that fail when I try to use it. In my case I did this by using `helm create` and then modifying the deployment.

Could you provide an example of a working chart that uses `generateName`?

cgetzen:
Hey, thanks for taking a look. After a brief search, I can't find any public charts that use generateName.
We've been using generateName with helm successfully for internal charts. The only portion that fails for is linting.
Are there other portions of helm that don't work with generateName?

Issue #9387, opened 2021-02-19
Local dependency charts breaks "helm dep update"


Ziphone:
Command _helm dep update_ "broke" when I added a local dependency to my chart. The chart works fine with the new dependency (e.g template, install, upgrade), but "helm dep update \<path of my chart\>" returns exit code 1 and this message:
```
Update Complete. ⎈Happy Helming!⎈
Error: Chart.yaml file is missing
helm.go:81: [debug] Chart.yaml file is missing
```
  
This might be a feature, but then I'd like to know the rationale behind. I suggest that Helm omits local dependencies rather than failing.
  
The dependencies section of my Chart.yaml.
```yaml
dependencies:
  - name: <some external chart>
    version: 1.2.x
    repository: <some external repo>
  - name: <my local chart>
    version: 0.1.8
    repository: file://charts/<my local chart>-0.0.1.tgz
```

Output of `helm version`:
version.BuildInfo{Version:"v3.4.2", GitCommit:"23dd3af5e19a02d4f4baa5b2f242645a1a3af629", GitTreeState:"dirty", GoVersion:"go1.15.5"}

Output of `kubectl version`:
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.8", GitCommit:"9f2892aab98fe339f3bd70e3c470144299398ace", GitTreeState:"clean", BuildDate:"2020-08-13T16:12:48Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"darwin/amd64"}

Cloud Provider/Platform (AKS, GKE, Minikube etc.): No clusters involved.

mattfarina:
The error looks like one you would get if `<path of my chart>` were wrong.

I notice the dependencies section isn't quire right for a local dependency. It should be:

```yaml
dependencies:
  - name: <some external chart>
    version: 1.2.x
    repository: <some external repo>
  - name: <my local chart>
    version: 0.1.8
```

Notice the local chart does not have a repository. Does changing that make a difference?

Ziphone:
Hi, Matt. Thanks for looking at my issue.

First I have to point out a mistake in my first post. My dependency section initially looked like this (version 0.1.8 of \<my local chart\> was a mistake):
``` yaml
dependencies:
  - name: <some external chart>
    version: 1.2.x
    repository: <some external repo>
  - name: <my local chart>
    version: 0.0.1
    repository: file://charts/<my local chart>-0.0.1.tgz
```

[...]

PR #9659, opened 2021-04-28
Feature/upgrade dep up

### LINKED ISSUE #7162: Add --dependency-update flag to the helm upgrade command

mladedav:

closes #7162

Requested multiple times.

This somewhat duplicates https://github.com/helm/helm/pull/9399 but this does not want to fix the behaviour of the flag, just copy it from install.
I don't think it's possible to deduplicate the code in install and upgrade since it is just a single call and error handling with a lot of parameters, so even if there was a common function to call, it would just have the same arguments and similar error handling on both calling sites. If you believe it would be still worthwhile, it can be changed.

I do not know how documentation is handled, but I have changed the description of --dependency-update both in upgrade and install to reflect the actual behaviour. Is a change needed elsewhere or does it propagate? Does the new flag need to be documented somewhere or does that also propagate?

**If applicable**:

### DIFF
--- a/cmd/helm/install.go
+++ b/cmd/helm/install.go
@@ -145,7 +145,7 @@ func addInstallFlags(cmd *cobra.Command, f *pflag.FlagSet, client *action.Instal
 	f.StringVar(&client.NameTemplate, "name-template", "", "specify template used to name the release")
 	f.StringVar(&client.Description, "description", "", "add a custom description")
 	f.BoolVar(&client.Devel, "devel", false, "use development versions, too. Equivalent to version '>0.0.0-0'. If --version is set, this is ignored")
-	f.BoolVar(&client.DependencyUpdate, "dependency-update", false, "run helm dependency update before installing the chart")
+	f.BoolVar(&client.DependencyUpdate, "dependency-update", false, "update dependencies if they are missing before installing the chart")
 	f.BoolVar(&client.DisableOpenAPIValidation, "disable-openapi-validation", false, "if set, the installation process will not validate rendered templates against the Kubernetes OpenAPI Schema")
 	f.BoolVar(&client.Atomic, "atomic", false, "if set, the installation process deletes the installation on failure. The --wait flag will be set automatically if --atomic is used")
 	f.BoolVar(&client.SkipCRDs, "skip-crds", false, "if set, no CRDs will be installed. By default, CRDs are installed if not already present")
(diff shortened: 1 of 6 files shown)


bacongobbler:
Hi @mladedav. Generally speaking we ask that tests are provided along with any changes to ensure the feature works as expected. Would you mind going through and writing tests for this?

Even if it's just copying code from elsewhere it's probably good to have a few test cases to ensure it works as expected.

bacongobbler:
> Is a change needed elsewhere or does it propagate? Does the new flag need to be documented somewhere or does that also propagate?

CLI flags are propagated elsewhere via doc updates. We update the CLI documentation every release using the hidden `helm doc` command.

If you'd like to expand further on the flag's usage and how to use it, feel free to send a PR over to https://github.com/helm/helm-www and link it back to this PR so we know not to merge it until this PR has been merged.

Thanks!

mladedav:
From what I could find there are no tests for this in the install_test.go (or dependency_test.go) file so there is nothing to copy. I could write something simple later today or tomorrow.

I am ok with the flag descriptions as they are.

mladedav:
@bacongobbler I have copied the basic test that is in install test for `--dependency-update`. However that test is mostly useless, if you delete the flag, the test runs the same way, because the dependency is already present in `charts` directory (both the original in install and the new test in upgrade).

I have tried to make it work with local repository, but it seems that the dependency update command does not work well with local repositories which means this is not easily tested. It is somewhat described in #9387 so I will add what I have found out there.

mladedav:
I have found out the tests run fine when I run them individually with `go test ./cmd/helm/ -run TestUpgradeCmd` but fail when all are run with `go test ./cmd/helm/` (thus they fail in CI). This is because for some reason, the `HELM_REPOSITORY_CONFIG` env var is set to `testdata/helmhome/helm/repositories.yaml` where there is one dummy repository, which alters the output.

I cannot figure out how or why that happens.

Options I see now are:
a) revert the test in install back to the branch where the update is not actually run, because no dependency is missing. The state of the tests is the same as it was before I started working on this but it isn't as good as other options. This may be fine since this is marked for deprecation, but it should be a conscious decision to cut corners here.
b) write the expected output of the tests so that running all tests works. Tests don't work individually for some reason though.
c) overwrite the repository config environment variable - since I have no idea where it came from I also have no idea what it can possibly break. I would like someone else to explain the inner workings here

mladedav:
@bacongobbler Hi, I didn't get an answer so I went with a) because I don't like option b) and I am not going to do c) on my own. Also this flag is to be deprecated before v4.

I hope this can be merged now.

bacongobbler:
Thanks. I'm not sure I'm going to have enough time in the near future to review this.

mattfarina (approved):

hickeyma (approved):
LGTM, thanks @mladedav

PR #7272, opened 2019-12-19
Port PR #4161 to Helmv3 - Fix incorrect timestamp when helm package

Hiruma31:
Source: https://github.com/helm/helm/pull/4161/commits/96a85a06ed48e618c53441cdc988680eb00f8e0c

PR #9674, opened 2021-05-07
Generate reproducible tarball

tamalsaha:
Signed-off-by: Tamal Saha <tamal@appscode.com>

Currently `helm package` does not produce reproducible tarballs. This change sets the timestamps to `0` to fix this. We store the tarballs in git repos. This helps us avoid unnecessary changes everytime we run `helm package`.

See also: https://unix.stackexchange.com/a/438330/42136

**If applicable**:

### DIFF
--- a/pkg/chartutil/save.go
+++ b/pkg/chartutil/save.go
@@ -23,7 +23,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"time"
+	"sort"
 
 	"github.com/pkg/errors"
 	"sigs.k8s.io/yaml"
@@ -203,6 +203,9 @@ func writeTarContents(out *tar.Writer, c *chart.Chart, prefix string) error {
 	}
 
 	// Save templates
+	sort.Slice(c.Templates, func(i, j int) bool {
+		return c.Templates[i].Name < c.Templates[j].Name
+	})
 	for _, f := range c.Templates {
 		n := filepath.Join(base, f.Name)
 		if err := writeToTar(out, n, f.Data); err != nil {
@@ -211,6 +214,9 @@ func writeTarContents(out *tar.Writer, c *chart.Chart, prefix string) error {
 	}
 
 	// Save files
+	sort.Slice(c.Files, func(i, j int) bool {
+		return c.Files[i].Name < c.Files[j].Name
+	})
 	for _, f := range c.Files {
 		n := filepath.Join(base, f.Name)
 		if err := writeToTar(out, n, f.Data); err != nil {
@@ -231,10 +237,9 @@ func writeTarContents(out *tar.Writer, c *chart.Chart, prefix string) error {
 func writeToTar(out *tar.Writer, name string, body []byte) error {
 	// TODO: Do we need to create dummy parent directory names if none exist?
 	h := &tar.Header{
-		Name:    filepath.ToSlash(name),
-		Mode:    0644,
-		Size:    int64(len(body)),
-		ModTime: time.Now(),
+		Name: filepath.ToSlash(name),
+		Mode: 0644,
+		Size: int64(len(body)),
 	}
 	if err := out.WriteHeader(h); err != nil {
 		return err
--- a/pkg/chartutil/save_test.go
+++ b/pkg/chartutil/save_test.go
@@ -123,12 +123,7 @@ func Indent(n int, text string) string {
 	return startOfLine.ReplaceAllLiteralString(text, indentation)
 }
 
-func TestSavePreservesTimestamps(t *testing.T) {
-	// Test executes so quickly that if we don't subtract a second, the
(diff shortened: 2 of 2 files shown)


mattfarina:
I think #3612 is the bug being fixed by this.

mattfarina:
I have an open question on this PR with the other maintainers. That is, is the removal of the timestamp a breaking change? We have been discussing these nuances lately. I added this issue to the agenda for the next meeting.

pkg/chartutil/save.go
```diff
@@ -231,10 +237,9 @@ func writeTarContents(out *tar.Writer, c *chart.Chart, prefix string) error {
 func writeToTar(out *tar.Writer, name string, body []byte) error {
 	// TODO: Do we need to create dummy parent directory names if none exist?
 	h := &tar.Header{
-		Name:    filepath.ToSlash(name),
-		Mode:    0644,
-		Size:    int64(len(body)),
-		ModTime: time.Now(),
```
baijum, line 237: Removing `ModTime` will reintroduce behavior described in #4158, which was fixed in #4161 and ported to V3 in #7272

tamalsaha: IMHO, the current implementation of modTime is also wrong because it just sets the time when helm package was run. If it wants to be correct, it should set the original template files timestamp.

Also, zero-ing out timestamps is a common pattern for reproducible builds https://github.com/GoogleContainerTools/distroless/issues/242

bacongobbler: Regardless, @baijum is right. We tried this, and someone reported it as a bug that was addressed with #4161. To go back and re-implement this again would be repeating past mistakes.

tamalsaha: May be this can be added as a flag to helm package command for those of us want to have reproducible charts.

bacongobbler: https://github.com/helm/helm/issues/3612#issuecomment-525340295 suggests the issue is not due to timestamps, but instead the sorted order of the files within the tar file.

Can you please look further into the suggestions made in that thread before coming up with a solution. Even if you strip mtime fields from the archive, you may still not produce a reproducible build. We've looked into this issue multiple times and discovered it's not the mtime of the original files causing the issue here. Or at least, there are multiple layers causing this issue, and to solve it we need to address them all.

We tested this theory in another area and discovered stripping mtime from the archive does not resolve the issue. See https://github.com/helm/helm/pull/8216#pullrequestreview-437935656.

scottrigby: Can we add a backwards-compatible flag, like `helm package --timestamp` where `timestamp` defaults to `true`, but users can set it to false for reproducibility on that level.

bacongobbler: The first step would be to write a test to identify the symptom. Then write a fix to alleviate the symptom.

It is possible that [we only need to sort the files in the archive](https://github.com/helm/helm/pull/8216#pullrequestreview-437935656) to fix the issue, which means there would not be any need for stripping file mtimes (and therefore no need for a `--timestamp` flag).

We did test the theory of stripping file mtimes before and [that did not allay the symptom](https://github.com/helm/helm/pull/8216#pullrequestreview-437935656). @jdolitsky can also attest to this as he and his coworker tried to fix this for `helm chart save`. So I'm not confident that stripping file mtimes will fix the problem.


Let's get some tests in first before we start providing solutions.

mattfarina (changes requested):
Thank you for putting this pull request together.

One thing this PR needs is tests to make sure the tarball build is reproducible. We need to build a package at least twice and compare the hash of them to verify the fix works.

This is important for Helm so that we can catch a regression in the future.

pkg/chartutil/save.go
mattfarina: This should be tried without the timestamp going to 0. But, I suspect any timestamp difference will cause a different hash. A different time stored in the tar when everything else is the same is still different content that would produce a different hash.

In another part of the PR there is sorting on the content. The order of the files matters and I hope the change here solves the issue. I have not checked, though.

If the timestamp needs to be 0 then that would need to be hidden behind a flag. We can't break backwards compatibility. I would suggest a name like `--reproducible-build`.

baijum: Please be aware that [Git doesn't preserve modification time](https://git.wiki.kernel.org/index.php/Git_FAQ#Why_isn.27t_Git_preserving_modification_time_on_files.3F) on files. So an Application Distributor cannot create a reproducible build by preserving the original file's modification time because the modified time will be different whenever the chart source checks out. I think using a flag (`--reproducible-build`) and setting timestamp to 0 looks acceptable to me—some tests to avoid regression would be great.

tamalsaha:
>  If not, someone else can pick it up.

If someone else can pick up that will be really great! Thanks.

PR #9675, opened 2021-05-07
Migrate to oras-go library

ndeloof:
see https://github.com/deislabs/oras/pull/265

https://github.com/deislabs/oras has been rearchitectured to introduce oras-go library
This PR adopts this library

**If applicable**:


jdolitsky (approved):

go.mod
```diff
@@ -34,7 +34,7 @@ require (
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.7.0
 	github.com/xeipuuv/gojsonschema v1.2.0
-	golang.org/x/crypto v0.0.0-20210220033148-5ea612d1eb83
+	golang.org/x/crypto v0.0.0-20210322153248-0c34fe9e7dc2
```
bacongobbler, line 37: Is there a way to prevent this update in this PR? I'm trying to figure out what's necessary for the name change vs. what should be pulled out into a separate PR. Changes to /x/crypto might be best as a separate review item if at all possible.

ndeloof: I can't tell, this has been introduced by `go mod tidy`, so I assume required by transitive dependencies...

hickeyma: @ndeloof I agree with @bacongobbler. Lets not change modules that are not part of the PR in question. Do you mind rolling this back?

ndeloof: As I said, this update is forced by `go mod`, if I revert this it will be added again any time you run `go mod xx` commands.

hickeyma: @ndeloof Looking further https://github.com/oras-project/oras-go/blob/main/go.mod is added in https://github.com/oras-project/oras-go/blob/main/go.mod, so probably picking it up from there.

hickeyma (approved):
I am ok with the change to `golang.org/x/crypto` module as stated in https://github.com/helm/helm/pull/9675#discussion_r643802199. LGTM. I will hold on merging as would like @bacongobbler to give his opinion on it.

thaJeztah:
ping @bacongobbler, ptal 🙏

ndeloof:
just rebased to fix go.sum conflict
@bacongobbler?

bacongobbler (approved):

howardjohn:
A heads up - this change causes an indirect import of k8s.io/kubernetes library which is generally pretty heavy. Opened https://github.com/containerd/containerd/issues/5781 to track this

PR #9677, opened 2021-05-07
Add JSON/YAML output support for `helm test`

lindhe:
Today when using `helm test` in CI, there is no robust way to extract the logs from the test pods. This commit makes it possible to get the logs in a machine readable way using the `--output` flag.

We created this PR as two commits to make the refactoring of `GetPodLogs` optional. (We recommend `git show --ignore-space-change` for that commit.)

It is unclear to us if it makes sense to let the user specify `--logs` when using JSON or YAML format, since in that case the logs are always included anyway. Another aspect to consider is whether the logs should be included in the first place unless `--logs` is given. Currently, `--logs` has no effect when combined with `-o json` or `-o yaml`.

We should also add tests and update the documentation accordingly.

**Testing for backwards compatibility**:
We have tested the changes by installing [metatest-0.1.0.tar.gz](https://github.com/helm/helm/files/6442599/metatest-0.1.0.tar.gz) (GitHub didn't allow `.tgz` hence `.tar.gz`).

To test that the PR is indeed backwards compatible, we ran the following commands and noted that the output was identical compared to upstream `helm`:
* `helm test metatest`
* `helm test metatest --logs`

Since the `-o` flag didn't exist before, introducing it should break nothing.

**If applicable**:


pkg/action/hooks.go
```diff
@@ -83,6 +86,20 @@ func (cfg *Configuration) execHook(rl *release.Release, hook release.HookEvent,
 		err = cfg.KubeClient.WatchUntilReady(resources, timeout)
 		// Note the time of success/failure
 		h.LastRun.CompletedAt = helmtime.Now()
+
+		if isTestHook(h) {
+			// "Why create a new client for each hook?" Because `cfg` can be `nil` outside the `for` loop.
+			client, err := cfg.KubernetesClientSet()
```
bacongobbler, line 92: can you provide a test case for this? I have no context how this issue can occur. A link to existing kubernetes client documentation or an issue in client-go describing this behaviour would also suffice.

Also can't this be an issue where a not-insignificant amount of test hooks create a large amount of kubernetes clients? That seems... Sub-optimal. I'm sure there must be a better way to handle this.

Have you looked for existing kubernetes client implementations of the log endpoint for better ideas?

bacongobbler:
Hi @lindhe.

Generally speaking we ask contributors to please either link to an existing issue or open a new ticket before submitting a PR. That way we have a clear test plan in place ("something is broken; here is how to reproduce the issue. Cannot reproduce the issue with PR X") and we can plan out how to schedule a fix into the next release. It makes things much easier on our end to prioritize work items.

Looking at this PR, it appears to be a feature. Because we already have a lot of features to review for Helm 3.7.0, I'm going to go ahead and schedule a review after Helm 3.7.0 has been released (September 8th).

In the meantime, we do ask for contributions to provide test cases. That way the code being introduced can be tested for regressions in future updates. I did not see any tests for the new `--output` flags being introduced here. Would you please provide a few?

Thanks.

pkg/action/hooks.go
```diff
@@ -107,6 +124,17 @@ func (cfg *Configuration) execHook(rl *release.Release, hook release.HookEvent,
 	return nil
 }
 
+// getHookLog gets the log from the pod associated with the given hook, which is expected to be a test hook.
+func getHookLog(client kubernetes.Interface, rel *release.Release, hook *release.Hook) (release.HookLog, error) {
+	req := client.CoreV1().Pods(rel.Namespace).GetLogs(hook.Name, &v1.PodLogOptions{})
+	responseBody, err := req.DoRaw(context.Background())
+	if err != nil {
+		var nothing release.HookLog
+		return nothing, errors.Wrapf(err, "unable to get pod logs for %s", hook.Name)
```
bacongobbler, line 133: Instead of returning an empty structure... Why not return the response body? That way we can check for `nil, err` rather than `release.HookLog{}, err`.

An empty struct generally means the object can still be handled. Returning a `nil` object is usually better because it provides context to the caller that the should not perform any business logic with this object.

pkg/action/hooks.go
```diff
@@ -107,6 +124,17 @@ func (cfg *Configuration) execHook(rl *release.Release, hook release.HookEvent,
 	return nil
 }
 
+// getHookLog gets the log from the pod associated with the given hook, which is expected to be a test hook.
+func getHookLog(client kubernetes.Interface, rel *release.Release, hook *release.Hook) (release.HookLog, error) {
+	req := client.CoreV1().Pods(rel.Namespace).GetLogs(hook.Name, &v1.PodLogOptions{})
+	responseBody, err := req.DoRaw(context.Background())
```
bacongobbler, line 130: Why the change from `req.Stream` to `req.DoRaw`?

bacongobbler (changes requested):
Marking as "do not merge" until a basic set of unit tests are provided.

lindhe:
> Generally speaking we ask contributors to please either link to an existing issue or open a new ticket before submitting a PR. That way we have a clear test plan in place ("something is broken; here is how to reproduce the issue. Cannot reproduce the issue with PR X") and we can plan out how to schedule a fix into the next release. It makes things much easier on our end to prioritize work items.

Yeah, sorry for that. In order to know what we wanted to ask for specifically, we began looking into the code. That led us to try out a few things and one thing led to another and then we had a PR instead of an issue. We'll aim to follow the official approach in the future.

> In the meantime, we do ask for contributions to provide test cases. That way the code being introduced can be tested for regressions in future updates. I did not see any tests for the new `--output` flags being introduced here. Would you please provide a few?

We'll absolutely provide test cases! Thanks a lot for the comments, they are really helpful. And we will try to answer your other questions too, as soon as we get time to look into this code again and remind ourselves what the reasoning was. :)

pkg/action/hooks.go
SimonAlling: > Instead of returning an empty structure... Why not return the response body?

I'm not sure I'm following here, but if we've promised (in the name and type signature of `getHookLog`) to return a `HookLog`, it wouldn't feel honest to return something that's definitely not a hook log.

Note that we're not returning an empty structure, but rather the empty string.

> That way we can check for `nil, err` rather than `release.HookLog{}, err`
>
> An empty struct generally means the object can still be handled. Returning a `nil` object is usually better because it provides context to the caller that the should not perform any further business logic with this object.

I must admit that I may not fully understand the point and its implications here, but I'd expect `getHookLog` to be idiomatically called like this:

```go
hookLog, err := getHookLog(client, rl, h)
if err != nil {
	// Handle error …
}
```

That is, if the second element of `getHookLog`'s return value (`err`) is not `nil`, it shouldn't matter what the first element (`hookLog`) is. Furthermore, the first element _can't_ be `nil`, because `nil` is not a member of the `HookLog` (`string`) type.

Finally, I'd like to mention that I find Go's standard error handling idiom – which uses a product type (_T_ × `error`) to represent what is, for all intents and purposes, actually a sum type (_T_ + `error`) – pretty frustrating to work with. I don't know of any "perfect" way to use it; the `nothing` idiom is basically the least bad one AFAIK.

mattfarina:
@lindhe We are looking at this for the upcoming 3.8 release in January. Would you be able to update this PR for that?

lindhe:
Hi! Thanks for the heads-up, we're glad to hear that you are considering this for 3.8! Yes, we will bump it up in our backlog then and start working on it during the new sprint we have starting today! I think it should not be too much work for us, it's just that we haven't gotten around to it before. Having it complete by January seems doable.

pkg/action/hooks.go
SimonAlling: For the record, the tests already fail if `client` is defined once outside the `for` loop.

<details>

```diff
diff --git a/pkg/action/hooks.go b/pkg/action/hooks.go
index 3d864e65..3d0a0871 100644
--- a/pkg/action/hooks.go
+++ b/pkg/action/hooks.go
@@ -44,6 +44,11 @@ func (cfg *Configuration) execHook(rl *release.Release, hook release.HookEvent,
        // hooke are pre-ordered by kind, so keep order stable
        sort.Stable(hookByWeight(executingHooks))
 
+       client, err := cfg.KubernetesClientSet()
+       if err != nil {
+               return errors.Wrapf(err, "unable to create Kubernetes client set to fetch pod logs")
+       }
+
        for _, h := range executingHooks {
                // Set default delete policy to before-hook-creation
                if h.DeletePolicies == nil || len(h.DeletePolicies) == 0 {
@@ -88,11 +93,6 @@ func (cfg *Configuration) execHook(rl *release.Release, hook release.HookEvent,
                h.LastRun.CompletedAt = helmtime.Now()
 
                if isTestHook(h) {
-                       // "Why create a new client for each hook?" Because `cfg` can be `nil` outside the `for` loop.
-                       client, err := cfg.KubernetesClientSet()
-                       if err != nil {
-                               return errors.Wrapf(err, "unable to create Kubernetes client set to fetch pod logs")
-                       }
                        hookLog, err := getHookLog(client, rl, h)
                        if err != nil {
                                return err
```

```
=== RUN   TestInstall/basic_install
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x2e6db5f]

goroutine 461 [running]:
helm.sh/helm/v3/pkg/action.(*Configuration).KubernetesClientSet(0xc0006e7d40)
	/…/pkg/action/action.go:281 +0x3f
helm.sh/helm/v3/pkg/action.(*Configuration).execHook(0xc0006e7d40, 0xc00039ed90, {0x362c20a, 0x1}, 0xc000da2000)
	/…/pkg/action/hooks.go:47 +0x1a7
helm.sh/helm/v3/pkg/action.(*Install).performInstall(0xc000d1e9a0, 0xe05785, 0xc00039ed90, {0x0, 0x3a02b40, 0x0}, {0x565e208, 0xc00135ad40, 0x0})
	/…/pkg/action/install.go:358 +0xff
created by helm.sh/helm/v3/pkg/action.(*Install).RunWithContext
	/…/pkg/action/install.go:347 +0x1fa8
FAIL	helm.sh/helm/v3/cmd/helm	26.000s
```

```
=== RUN   TestInstallRelease
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x2ea855f]

goroutine 37 [running]:
helm.sh/helm/v3/pkg/action.(*Configuration).KubernetesClientSet(0xc000340340)
	/…/pkg/action/action.go:281 +0x3f
helm.sh/helm/v3/pkg/action.(*Configuration).execHook(0xc000340340, 0xc000036f50, {0x33694ab, 0x100000000ebaf7b}, 0x0)
	/…/pkg/action/hooks.go:47 +0x1a7
helm.sh/helm/v3/pkg/action.(*Install).performInstall(0xc000350420, 0xc0000601f0, 0xc000036f50, {0x0, 0x0, 0x0}, {0x5238368, 0xc000060238, 0x0})
	/…/pkg/action/install.go:358 +0xff
created by helm.sh/helm/v3/pkg/action.(*Install).RunWithContext
	/…/pkg/action/install.go:347 +0x1fa8
FAIL	helm.sh/helm/v3/pkg/action	0.181s
```

</details>

SimonAlling: However, seems like, contrary what our comment claims, it is not `cfg` that is `nil` – it is in fact e.g. `&{<nil> 0xc000475240 0xc0004b4820 0xc000180870 0x47eb1a0 0x2eccec0}`.

SimonAlling: Currently investigating why the `TestInstallRelease` test fails with the above change (calling `cfg.KubernetesClientSet` before the `for` loop). Turns out `i.cfg.RESTClientGetter` is `nil` [when `i.cfg.execHook` is called in `performInstall` in `pkg/action/install.go`](https://github.com/helm/helm/blob/5d47255a3d6c886850d6c6cbead1b00b204d07b1/pkg/action/install.go#L358).

pkg/action/hooks.go
SimonAlling: WDYT, @bacongobbler?

pkg/action/hooks.go
SimonAlling: In the current implementation, [the logs are acquired](https://github.com/helm/helm/blob/03d9e7ca62e428b9263dcf19542691b0531c0ccc/pkg/action/release_testing.go#L114) and then immediately [written to an `io.Writer`](https://github.com/helm/helm/blob/03d9e7ca62e428b9263dcf19542691b0531c0ccc/pkg/action/release_testing.go#L120). With our suggested changes, the acquisition and the writing are decoupled such that no `io.Writer` is in scope [when the logs are acquired](https://github.com/helm/helm/blob/d478b411b953f69e2f5d8a5505f49ba2dedc8524/pkg/action/hooks.go#L129). Instead, [`getHookLog` _returns_ the hook log](https://github.com/helm/helm/blob/d478b411b953f69e2f5d8a5505f49ba2dedc8524/pkg/action/hooks.go#L127), which can then be [written to an `io.Writer`](https://github.com/helm/helm/blob/d478b411b953f69e2f5d8a5505f49ba2dedc8524/pkg/action/release_testing.go#L115) or handled in any other suitable way.

Note that, unlike the current implementation, `getHookLog`

  * can _acquire_ logs without the side effect of _writing_ to an `io.Writer`, and
  * is used at multiple locations ([in `hooks.go`](https://github.com/helm/helm/blob/d478b411b953f69e2f5d8a5505f49ba2dedc8524/pkg/action/hooks.go#L95) and [in `release_testing.go`](https://github.com/helm/helm/blob/d478b411b953f69e2f5d8a5505f49ba2dedc8524/pkg/action/release_testing.go#L110)).

It might be possible to throw out `req.DoRaw` and shoehorn `req.Stream` into its place, but I don't know how, and it seems to me that doing so would just make `getHookLog` unnecessarily complex.

That being said, I can't assess e.g. performance implications of using `req.DoRaw` over `req.Stream` – I usually think in types and values rather than side effects and low-level implementation details.

SimonAlling:
> Marking as "do not merge" until a basic set of unit tests are provided.

I have dug into the code base again to try to find existing tests that could serve as a base for tests for this PR, but it seems like the behavior of `helm test` isn't covered by tests at all. These changes at least don't cause any test failures:

```diff
diff --git a/pkg/action/release_testing.go b/pkg/action/release_testing.go
index 5ae909bd..28a184c2 100644
--- a/pkg/action/release_testing.go
+++ b/pkg/action/release_testing.go
@@ -50,2 +50,5 @@ func NewReleaseTesting(cfg *Configuration) *ReleaseTesting {
 func (r *ReleaseTesting) Run(name string) (*release.Release, error) {
+       if true {
+               panic("Panic in Run")
+       }
        if err := r.cfg.KubeClient.IsReachable(); err != nil {
@@ -102,2 +105,5 @@ func (r *ReleaseTesting) Run(name string) (*release.Release, error) {
 func (r *ReleaseTesting) GetPodLogs(out io.Writer, rel *release.Release) error {
+       if true {
+               panic("Panic in GetPodLogs")
+       }
        client, err := r.cfg.KubernetesClientSet()
```

pkg/action/hooks.go
bacongobbler: The pattern of returning `nil, err` when an error occurs is the standard idiom across Go's standard library. The ONLY exception is when the user can actually do something useful with the data, like when the developer calls `ReadLine` and it hits EOF. In that case it should return `line, io.EOF` because the user can still so something with the rest of the line and catch `io.EOF`.

In this case there isn't anything the user can do with the hook log, so we should return a `nil` object.

Hope this helps.

pkg/action/hooks.go
```diff
@@ -83,6 +86,19 @@ func (cfg *Configuration) execHook(rl *release.Release, hook release.HookEvent,
 		err = cfg.KubeClient.WatchUntilReady(resources, timeout)
 		// Note the time of success/failure
 		h.LastRun.CompletedAt = helmtime.Now()
+
+		if isTestHook(h) {
+			client, err := cfg.KubernetesClientSet() // "Why create a new client for each hook?" Because `cfg.RESTClientGetter` can be `nil` before the `for` loop, as detected by the `TestInstall/basic_install` and `TestInstallRelease` tests.
```
bacongobbler, line 91: I'm really confused. Why are we creating a new client here? Why is there no check for this earlier? On line 86 we call `cfg.KubeClient` to perform a WatchUntilReady. In other words the argument being made here seems illogical because we've already used the client prior to this line, so it shouldn't be nil.

pkg/action/hooks.go
bacongobbler: https://pkg.go.dev/k8s.io/client-go@v0.22.4/rest#Request.Stream

https://pkg.go.dev/k8s.io/client-go@v0.22.4/rest#Request.DoRaw

Seems like `req.Stream` performs some higher-level operations like handling non-2xx response codes whereas `req.DoRaw` does nothing of the sort. As a result, I'm very concerned that this change will not capture the same error cases Kubernetes might throw at us as before.

I'd highly suggest looking into the differences between these two API calls before swapping them out.

bacongobbler:
> but it seems like the behavior of helm test isn't covered by tests at all.

We shoot for good test coverage wherever we can. A lack of tests in place shouldn't mean we should continue that practice. Especially when we (the maintainers) are the ones who will be maintaining this code long after the pull request has been merged.

pkg/action/hooks.go
SimonAlling: > Why are we creating a new client _here_ [and not somewhere else]?

If we move `client, err := cfg.KubernetesClientSet()` to _before_ the `for _, h := range executingHooks` loop, then the `TestInstall/basic_install` and `TestInstallRelease` tests fail because `cfg.RESTClientGetter` is `nil` when `cfg.KubernetesClientSet` is called.

> Why are we creating a new client [at all]?

`getHookLog` must be passed a `kubernetes.Interface`, and there is none in scope.

> On line 86 we call `cfg.KubeClient` to perform a WatchUntilReady. In other words the argument being made here seems illogical because we've already used the client prior to this line, so it shouldn't be nil.

I'm also really confused. :brain: We _create_ the client on that line (`client, err := cfg.KubernetesClientSet()`), so we cannot possibly have used it prior to it, can we? On line 86, `cfg.KubeClient.WatchUntilReady` is called, but `cfg.KubeClient` is a `kube.Interface`, whereas `client` is a `kubernetes.Interface`. There seems to be some misunderstanding going on here.

One might argue that `getHookLog` should be rewritten so that it doesn't require a `kubernetes.Interface`, but I have no idea how to do so. We probably just copied that code from somewhere else in the code base and modified it for our purpose.

pkg/action/hooks.go
SimonAlling: But we literally cannot return `nil`, because the return type is `(string, error)` and `nil` is not a member of `string`.

Surely the standard idiom in general is not returning `nil, err`, but rather returning `<the zero value>, err` (e.g. `"", err`, `0, err` etc)? Some examples of functions in the standard library that do so are [`os.MkdirTemp`](https://pkg.go.dev/os@go1.17.3#MkdirTemp), [`os.ReadLink`](https://pkg.go.dev/os@go1.17.3#Readlink) and [`bytes.Reader.Read`](https://pkg.go.dev/bytes@go1.17.3#Reader.Read). As far as I can tell, `getHookLog` is idiomatically compliant in that aspect.

pkg/action/hooks.go
SimonAlling: Yes. I'll see if I can revert to `req.Stream`.

pkg/action/hooks.go
bacongobbler: I don't understand why you're returning an untyped raw response body in the first place. This might have more to do with `req.DoRaw` than the actual implementation details but in general passing raw response bodies back to the user is asking for trouble.

Most functions provide some type of higher-level abstraction so that the caller can actually do something useful with the response body...

pkg/action/release_testing.go
```diff
@@ -108,20 +106,16 @@ func (r *ReleaseTesting) GetPodLogs(out io.Writer, rel *release.Release) error {
...
-					return errors.Wrapf(err, "unable to write pod logs for %s", h.Name)
-				}
+		if isTestHook(h) {
+			hookLog, err := getHookLog(client, rel, h)
+			if err != nil {
+				return err
+			}
+			fmt.Fprintf(out, "POD LOGS: %s\n", h.Name)
+			_, err = fmt.Fprint(out, string(hookLog))
+			fmt.Fprintln(out)
```
bacongobbler, line 116: This will `panic` if `out` cannot be written to. `err` needs to be handled before you attempt to call `Fprint` a second time.

I don't understand why you have three different calls to `fmt.Fprintf`, `fmt.Fprint`, and `fmt.Fprintln` all in quick succession when a simple `fmt.Fprinf(out, "POD LOGS" %s\n%s\n", h.Name, string(hookLog))` would've accomplished the same thing.

bacongobbler (dismissed):

pkg/action/release_testing.go
SimonAlling: The reason is because that's how the code was already written, as can be seen in the diff. I've only changed `io.Copy(out, logReader)` to `fmt.Fprint(out, string(hookLog))`.

Just to be clear, do you mean that `err` needs to be handled before `Fprintln` is called on line 116? If so, why was that not the case before (when the logs were written using `io.Copy` instead of `fmt.Fprint`)?

pkg/action/hooks.go
SimonAlling: I have now pushed a commit that reverts from `req.DoRaw` to `req.Stream`.

pkg/action/hooks.go
SimonAlling: Done.

pkg/action/hooks.go
SimonAlling: I'm finally starting to understand the problem at hand. The `RESTClientGetter` field in the config is assigned a non-`nil` value when actions are run as part of actual commands, but not when they are tested. So any code using `KubernetesClientSet` will probably panic during testing, since [it assumes the `RESTClientGetter` field to have a non-`nil` value](https://github.com/helm/helm/blob/8ca401398d8b6dfd41b46f446abcd5285e219829/pkg/action/action.go#L281).

SimonAlling: This probably implies that we cannot test [`GetPodLogs`](https://github.com/helm/helm/blob/8ca401398d8b6dfd41b46f446abcd5285e219829/pkg/action/release_testing.go#L104) (which calls the `KubernetesClientSet` method) without refactoring it substantially.

SimonAlling: We managed to populate the `RESTClientGetter` field in the affected test cases with `cli.New().RESTClientGetter()`. However, that `RESTClientGetter` is only useful to communicate with an actual cluster, not when working with fake releases in a testing scenario. All in all, it seems like the tests requested for this PR are inherently complicated to write.

SimonAlling: I'd say the most problematic line from the perspective of writing tests is [this one](https://github.com/helm/helm/blob/8ca401398d8b6dfd41b46f446abcd5285e219829/pkg/action/release_testing.go#L113):

```go
req := client.CoreV1().Pods(r.Namespace).GetLogs(h.Name, &v1.PodLogOptions{})
```

I find it non-trivial to see how this could be mocked in the test suite.

lindhe: It would be wonderful if @bacongobbler or someone else could help us out with this part. What has the approach been in other cases when Helm have wanted to write tests for something that more-or-less requires connectivity to a cluster?

pkg/action/hooks.go
SimonAlling: > Also can't this be an issue where a not-insignificant amount of test hooks create a large amount of kubernetes clients? That seems... Sub-optimal.

On the other hand, a client is only ever needed when hook logs are to be acquired, i.e. only for test hooks. So while creating the client outside the `for` loop means we only create one, no matter the number of test hooks, it also means we _always_ create one, even when it's not needed at all.

pkg/action/hooks.go
SimonAlling: We finally managed to piece together a unit test by adding `GetHookLog` as a field in `action.Configuration`, so that we could mock it; see e0b9ef060198e7836bde9408699254196cce9c8f or `git show e0b9ef060198e7836bde9408699254196cce9c8f --color-moved`.

Couldn't really avoid creating a new `kubernetes.Interface` for each hook though, but, as [previously stated](https://github.com/helm/helm/pull/9677#discussion_r765903140), that may not be a major issue.

SimonAlling: Working on a unit test for the command (in contrast to the action).

SimonAlling: The problem right now is that [`execHook`](https://github.com/helm/helm/blob/8ca401398d8b6dfd41b46f446abcd5285e219829/pkg/action/hooks.go#L30) overwrites the `StartedAt` and `CompletedAt` fields with `helmtime.Now()`, whose value can't be predicted and included in a golden-output text file.

SimonAlling: 👆 Almost done with a solution to that problem.

SimonAlling: I have added unit tests for the command now!

SimonAlling:
_I could not reply to [the existing thread](https://github.com/helm/helm/pull/9677/#discussion_r762115450), probably because the original commit wasn't signed so I had to force-push or something. I don't know._

Having thought a bit more about this, I just want to mention that the reason I didn't change to a single `fmt.Fprintf` call was because I always try to minimize unnecessary diff noise. When I review PRs or examine old commits myself, fly-by cleanup changes and the like often leave me confused, wondering whether they are part of the change being made or not.

What I didn't realize was that it actually makes perfect sense to make said change in this PR, since the printing couldn't be done in a single call before, but now it can. So, would you like us to make the change below?

```diff
--- a/pkg/action/release_testing.go
+++ b/pkg/action/release_testing.go
@@ -106,9 +106,7 @@ func (r *ReleaseTesting) GetPodLogs(out io.Writer, rel *release.Release) error {
                        if err != nil {
                                return err
                        }
-                       fmt.Fprintf(out, "POD LOGS: %s\n", h.Name)
-                       _, err = fmt.Fprint(out, string(hookLog))
-                       fmt.Fprintln(out)
+                       _, err = fmt.Fprintf(out, "POD LOGS: %s\n%s\n", h.Name, string(hookLog))
                        if err != nil {
                                return errors.Wrapf(err, "unable to write pod logs for %s", h.Name)
                        }
```

pkg/action/release_testing.go
SimonAlling: Suggested change made in 73000df8dd16911783fd1e860a262cf6d0221c9b.

SimonAlling:
Checks failed, but not our fault.

```
Download action repository 'github/codeql-action@v1' (SHA:5f532563584d71fdef14ee64d17bafb34f751ce5)
Warning: Failed to download action 'https://api.github.com/repos/github/codeql-action/tarball/5f532563584d71fdef14ee64d17bafb34f751ce5'. Error: The operation was canceled.
Warning: Back off 19.821 seconds before retry.
Warning: Failed to download action 'https://api.github.com/repos/github/codeql-action/tarball/5f532563584d71fdef14ee64d17bafb34f751ce5'. Error: The operation was canceled.
Warning: Back off 25.718 seconds before retry.
Error: The operation was canceled.
```

pkg/action/action.go
```diff
@@ -362,7 +395,7 @@ func (cfg *Configuration) recordRelease(r *release.Release) {
 }
 
 // Init initializes the action configuration
-func (cfg *Configuration) Init(getter genericclioptions.RESTClientGetter, namespace, helmDriver string, log DebugLog) error {
+func (cfg *Configuration) Init(getter genericclioptions.RESTClientGetter, namespace, helmDriver string, log DebugLog, hookLogGetter HookLogGetter) error {
```
SimonAlling, line 398: Is the addition of a parameter here a breaking change?

SimonAlling:
Ah, ye olde missing sign-off error …

pkg/action/action.go
SimonAlling: Can I still comment here after force-pushing?

SimonAlling:
This PR has become so large that we might want to consider breaking it apart into one PR adding `helm test` test cases along with the necessary refactors and one follow-up PR adding `--output` support to `helm test`.

pkg/action/action.go
bacongobbler: Yes, it is a breaking change. Programs using this function using the old function signature would no longer compile.

lindhe: We tried to find a way so that the logs aren't included in the output unless `--logs` is set. This is apparently not they way, since it breaks functionality, and I am pretty sure there is no other way to prohibit fetching the logs at all. What we could do is to either always include the logs (as before this recent change) or to remove the `logs` element from the object before printing it... Do you think any of that would be desirable?

SimonAlling: I think it's possible to have the cake and eat it in this case.

SimonAlling: There. I have also updated [the commit I created](https://github.com/simonalling/helm/commit/72ca6c485afcc4a6988506a672234b241770eae4) that only includes the `helm test` test suite, which I think should be merged as its own PR first. This PR can then be rebased/updated accordingly.

PR #9681, opened 2021-05-08
support max-col-width parameters

### LINKED ISSUE #9614: helm chart list could not show all messages
Output of `helm version`:
helm version

Output of `kubectl version`:
Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.4", GitCommit:"e87da0bd6e03ec3fea7933c4b5263d151aafd07c", GitTreeState:"clean", BuildDate:"2021-02-18T16:12:00Z", GoVersion:"go1.15.8", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"19+", GitVersion:"v1.19.6-eks-49a6c0", GitCommit:"49a6c0bf091506e7bafcdb1b142351b69363355a", GitTreeState:"clean", BuildDate:"2020-12-23T22:10:21Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"linux/amd64"}

EKS

Because of [here](https://docs.aws.amazon.com/AmazonECR/latest/userguide/ECR_on_EKS.html#using-helm-charts-eks), i need to create a long name for a chart cache, but "helm chart list" couldn't show all names and i couldn't use "helm chart remove <cache-name>".
image

### LINKED ISSUE #9678: Add option to `helm chart list` to avoid truncating refs with ellipsis
# Problem

Currently, if you call `helm chart list` and you have saved charts with long refs, the refs will be truncated with ellipsis. This is a problem because it cuts off the more changeable portions of the chart ref including the end of the registry and the tags. This makes it hard to determine what is stored in the local cache.

It would be helpful if a flag could be added to `helm chart list` that would force it to print the entire ref, no matter how long it is. Something like `helm chart list -l/--long`.

# Question

In the mean time, is there a way to work around this problem? I currently have several charts cached on my local machine whose refs always get truncated. I don't remember what tags they used and am unable to delete them.

Is there a way to find the full refs for existing saved charts?

Output of `helm version`: 
```
version.BuildInfo{Version:"v3.5.4", GitCommit:"1b5edb69df3d3a08df77c9902dc17af864ff05d1", GitTreeState:"clean", GoVersion:"go1.15.11"}
```

Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.5", GitCommit:"e6503f8d8f769ace2f338794c914a96fc335df0f", GitTreeState:"clean", BuildDate:"2020-07-04T15:01:15Z", GoVersion:"go1.14.4", Compiler:"gc", Platform:"darwin/amd64"}
```
*I should probably upgrade...*

On-prem

cndoit18:

closes #9678 
closes #9614 

**If applicable**:

### DIFF
--- a/internal/experimental/registry/client.go
+++ b/internal/experimental/registry/client.go
@@ -51,6 +51,7 @@ type (
 		authorizer      *Authorizer
 		resolver        *Resolver
 		cache           *Cache
+		columnWidth     uint
 	}
 )
 
@@ -94,6 +95,10 @@ func NewClient(opts ...ClientOption) (*Client, error) {
 			return nil, err
 		}
 		client.cache = cache
+	}
+
+	if client.columnWidth == 0 {
+		client.columnWidth = 60
 	}
 	return client, nil
 }
@@ -279,7 +284,7 @@ func (c *Client) RemoveChart(ref *Reference) error {
 // PrintChartTable prints a list of locally stored charts
 func (c *Client) PrintChartTable() error {
 	table := uitable.New()
-	table.MaxColWidth = 60
+	table.MaxColWidth = c.columnWidth
 	table.AddRow("REF", "NAME", "VERSION", "DIGEST", "SIZE", "CREATED")
 	rows, err := c.getChartTableRows()
 	if err != nil {
--- a/pkg/action/chart_list.go
+++ b/pkg/action/chart_list.go
@@ -18,11 +18,14 @@ package action
 
 import (
 	"io"
+
+	"helm.sh/helm/v3/internal/experimental/registry"
 )
 
 // ChartList performs a chart list operation.
 type ChartList struct {
-	cfg *Configuration
+	cfg         *Configuration
+	ColumnWidth uint
 }
 
 // NewChartList creates a new ChartList object with the given configuration.
@@ -34,5 +37,8 @@ func NewChartList(cfg *Configuration) *ChartList {
 
 // Run executes the chart list operation
 func (a *ChartList) Run(out io.Writer) error {
-	return a.cfg.RegistryClient.PrintChartTable()
+	client := a.cfg.RegistryClient
+	opt := registry.ClientOptColumnWidth(a.ColumnWidth)
+	opt(client)
+	return client.PrintChartTable()
 }
--- a/pkg/action/dependency.go
+++ b/pkg/action/dependency.go
@@ -37,11 +37,14 @@ type Dependency struct {
 	Verify      bool
 	Keyring     string
 	SkipRefresh bool
+	ColumnWidth uint
 }
 
 // NewDependency creates a new Dependency object with the given configuration.
 func NewDependency() *Dependency {
-	return &Dependency{}
+	return &Dependency{
+		ColumnWidth: 80,
+	}
 }
 
 // List executes 'helm dependency list'.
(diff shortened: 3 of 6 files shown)


cndoit18:
@bacongobbler I have modified the hard-coded place, help me review him, please

jsolbrig:
@cndoit18 Thank you for this fix. I tested this out and it definitely addresses the issue that I brought up in #9678.

pkg/action/dependency.go
```diff
@@ -45,7 +45,7 @@ func NewDependency() *Dependency {
 }
 
 // List executes 'helm dependency list'.
-func (d *Dependency) List(chartpath string, out io.Writer) error {
+func (d *Dependency) List(chartpath string, out io.Writer, columnWidth uint) error {
```
bacongobbler, line 48: changing the function signature would be construed as a backwards-incompatible change, so this will have to be changed. For context: https://github.com/helm/helm/blob/main/CONTRIBUTING.md#semantic-versioning

In other actions, we add new fields to the action object. Those are then read during `Run`. That way we can provide these features without breaking compatibility.

For example, take a look at install.go: https://github.com/helm/helm/blob/43853ea772e779b617b011a8d4e69205636fa4f9/pkg/action/install.go#L68-L104

pkg/action/chart_list.go
```diff
@@ -33,6 +33,6 @@ func NewChartList(cfg *Configuration) *ChartList {
 }
 
 // Run executes the chart list operation
-func (a *ChartList) Run(out io.Writer) error {
-	return a.cfg.RegistryClient.PrintChartTable()
+func (a *ChartList) Run(out io.Writer, columnWidth uint) error {
```
bacongobbler, line 36: same here; we will have to change this so we won't break compatibility.

internal/experimental/registry/client.go
```diff
@@ -277,9 +277,9 @@ func (c *Client) RemoveChart(ref *Reference) error {
 }
 
 // PrintChartTable prints a list of locally stored charts
-func (c *Client) PrintChartTable() error {
+func (c *Client) PrintChartTable(columnWidth uint) error {
```
bacongobbler, line 280: since this is in `experimental`, this is fine. However it does seem like an awkward API design to be passing optional fields as mandatory.

Why not add this as a field to the Client object?

bacongobbler (changes requested):

internal/experimental/registry/client.go
```diff
@@ -95,6 +96,10 @@ func NewClient(opts ...ClientOption) (*Client, error) {
 		}
 		client.cache = cache
 	}
+
+	if client.columnWidth == 0 {
+		client.columnWidth = 80
```
bacongobbler, line 101: the old default was 60. Can you change this to match?

pkg/cli/environment.go
```diff
@@ -115,6 +117,7 @@ func (s *EnvSettings) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&s.RegistryConfig, "registry-config", s.RegistryConfig, "path to the registry config file")
 	fs.StringVar(&s.RepositoryConfig, "repository-config", s.RepositoryConfig, "path to the file containing repository names and URLs")
 	fs.StringVar(&s.RepositoryCache, "repository-cache", s.RepositoryCache, "path to the file containing cached repository indexes")
+	fs.UintVar(&s.MaxColWidth, "max-col-width", 80, "maximum column width for output table")
```
bacongobbler, line 120: This seems slightly confusing... `func (s *EnvSettings) AddFlags` adds these flags to the global flagset. It should be added where needed for now so it won't clobber the localized `--max-col-width` for `helm search repo` as well as `helm depedendency list`.

This also adds this flag to every single command, even if it doesn't produce table output.

bacongobbler (changes requested):

cndoit18:
@bacongobbler I have done some way of backward compatibility.

bacongobbler (approved):
This looks great. Thank you!

Needs another maintainer to sign off on the PR as per the [contributing policy](https://github.com/helm/helm/blob/main/CONTRIBUTING.md#size-labels)

hickeyma (approved):
LGTM, thanks @cndoit18  for adding this capability.

PR #9690, opened 2021-05-11
error eturn instead of panic for pkg/action.Configuration#Init() method

wuhuizuo:
Signed-off-by: wuhuizuo <wuhuizuo@126.com>

**What this PR does / why we need it**: fix issue #9698 

**What I did**: error eturn instead of panic for pkg/action.Configuration#Init() method
**Why**: since the method have a error type return value, it should not panic in method body.

**If applicable**:

### DIFF
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -408,7 +408,7 @@ func (c *Configuration) Init(getter genericclioptions.RESTClientGetter, namespac
 		store = storage.Init(d)
 	default:
 		// Not sure what to do here.
-		errors.Errorf("Unknown driver in HELM_DRIVER: %s", helmDriver)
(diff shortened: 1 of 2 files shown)


cndoit18 (commented):
I think these are necessary storage drivers, if there has a problem, then all the follow-up all the operations will be invalid, so there should throw panic.

wuhuizuo:
> PRs must be linked to an existing issue. If none exists, may you please open one explaining the issue and expected outcome?
> 
> Please also write tests to test and verify this PR fixes your issue. Thanks.

I created issue #9698.

wuhuizuo:
unit tests added.

wuhuizuo:
close it becauseof the discussion did not reach agreement.

PR #9713, opened 2021-05-18
fix(helm): Retry Conflict error for createResource, deleteResource

### LINKED ISSUE #9710: Conflict error because resource quota version is old
This is a direct consequence of https://github.com/kubernetes/kubernetes/issues/67761

When helm gets a 409 Conflict, it will not retry:
https://github.com/helm/helm/blob/51097a55f14f208e5382d730d85662580887976a/pkg/kube/client.go#L371

However, as a sonsequence of https://github.com/kubernetes/kubernetes/issues/67761 and until that is fixed, helm should retry when it receives a `409` that's a result of a manipulation of kind `resourcequota` 

I am using an older version of OpenShift (4.6.17) with the latest helm built from this repository.

It does not happen a lot, but it does, and the following script helped me catch it. In https://github.com/kubernetes/kubernetes/issues/67761, they talked about how this would fail when installing gitlab, so I used that helm chart.

~~~
cat << 'FOE' > test_gitlab.sh
#!/bin/bash

function create_project() {
    oc new-project gitlab || oc project gitlab
    cat <<'EOF' | oc apply -f -
apiVersion: v1
kind: ResourceQuota
metadata:
  name: quota
spec:
  hard:
    configmaps: "62"
    limits.cpu: "170"
    limits.memory: 260Gi
    openshift.io/imagestreams: "0"
    persistentvolumeclaims: "50"
    pods: "100"
    replicationcontrollers: "50"
    requests.cpu: "71"
    requests.memory: 205Gi
    requests.storage: 350Gi
    limits.cpu: "71"
    limits.memory: 205Gi
    requests.storage: 350Gi
    resourcequotas: "1"
    secrets: "512"
    services: "70"
    services.nodeports: "5"
EOF
cat <<'EOF' | oc apply -f -
apiVersion: v1
kind: LimitRange
metadata:
  name: cpu-limit-range
spec:
  limits:
  - default:
      memory: 256Mi
      cpu: 125m
    defaultRequest:
      cpu: 125m
      memory: 128Mi
    type: Container
EOF
    oc adm policy add-scc-to-user privileged -z default
    oc adm policy add-scc-to-user privileged -z gitlab-shared-secrets
}

function delete_project() {
    oc project default
    oc delete project gitlab
    if [ "$?" != "1" ]; then
        echo "Sleeping for 60 seconds to give the project time to be deleted"
        sleep 60
    fi
}

function install_gitlab() {
    helm upgrade --debug --install gitlab gitlab/gitlab   --timeout 600s   --set global.hosts.domain=example.com   --set global.hosts.externalIP=10.10.10.10   --set certmanager-issuer.email=me@example.com
    if [ "$?" == "0" ]; then
        GREEN='\033[0;32m'
        NC='\033[0m' # No Color
        printf "${GREEN}SUCCEED${NC}\n"
    else
        RED='\033[0;31m'
        NC='\033[0m' # No Color
        printf "${RED}FAIL${NC}\n"
	exit 1
    fi
}

function uninstall_gitlab() {
    helm uninstall gitlab
}

delete_project
while true; do 
    echo "=== $(date) ==="
    echo "=== Create project ==="
    create_project
    echo "=== Install gitlab ==="
    install_gitlab    
    sleep 5
    echo "=== Uninstall gitlab ==="
    uninstall_gitlab
    echo "=== Delete project ==="
    delete_project
done
FOE
bash -x test_gitlab.sh 2>&1 | tee -a test_gitlab.log
~~~

I created a debug build of helm:
~~~
[akaris@linux helm]$ git diff |cat
diff --git a/pkg/kube/client.go b/pkg/kube/client.go
index 34079e7a..90388235 100644
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -120,7 +120,40 @@ func (c *Client) IsReachable() error {
 // Create creates Kubernetes resources specified in the resource list.
 func (c *Client) Create(resources ResourceList) (*Result, error) {
 	c.Log("creating %d resource(s)", len(resources))
-	if err := perform(resources, createResource); err != nil {
+
+	createResourceLogger := func(info *resource.Info) error {
+		c.Log("Creating resource: {" +
+			"Time UnixNano: " + fmt.Sprintf("%v", time.Now().UnixNano()) + ", " +
+			"Name: " + info.Name + ", " +
+			"Namespace: " + info.Namespace + ", " +
+			"Source: " + info.Source + ", " +
+			"ResourceVersion: " + info.ResourceVersion +
+			"}")
+		j, err := json.Marshal(info.Object)
+		if err == nil {
+			c.Log("==> Object: " + string(j))
+		} else {
+			c.Log("Could not marshal Name " + info.Name + " Namespace " + info.Namespace)
+		}
+		returnErr := createResource(info)
+		if returnErr != nil {
+			c.Log(
+				"FAILED Time UnixNano: " + fmt.Sprintf("%v", time.Now().UnixNano()) +
+					" Name: " + info.Name +
+					", Namespace: " + info.Namespace +
+					", ERROR: " + returnErr.Error(),
+			)
+			errJson, errErr := json.Marshal(returnErr)
+			if errErr == nil {
+				c.Log("==> ERROR JSON: " + string(errJson))
+			} else {
+				c.Log("Could not marshal ERROR")
+			}
+		}
+		return returnErr
+	}
+
+	if err := perform(resources, createResourceLogger); err != nil {
 		return nil, err
 	}
 	return &Result{Created: resources}, nil
~~~

Please find the output for a failed run attached. In helm, one should be able to work around this easily by retrying a `perform` if it returns a 409. Notwithstanding, the actual kubernetes issues here https://github.com/kubernetes/kubernetes/issues/67761 should be fixed instead of asking every client to fix this client side, but as far as  I can see, pending a solution on the kubernetes side, clients implement a retry to work around this.

andreaskaris:
kubernetes might at any time throw 409 Conflict Error codes. Clients
are supposed to retry when this happens. As an example, see
kubernetes/issues/67761 where such an issues might happen when the
cluster manipulates a projects's ResourceQuotas.
    
Catch such Conflict Errors on createResource and deleteResource and
retry before giving up. Due to the more complex logic and focus on
kubernetes/issues/67761, this patch purposefully omits possibly
needed changes to updateResource and instead defers them to another
patch if required in the future.
    
Closes issue #9710
    
Signed-off-by: Andreas Karis <ak.karis@gmail.com>

**If applicable**:

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -55,6 +55,7 @@ import (
 	"k8s.io/client-go/rest"
 	cachetools "k8s.io/client-go/tools/cache"
 	watchtools "k8s.io/client-go/tools/watch"
+	"k8s.io/client-go/util/retry"
 	cmdutil "k8s.io/kubectl/pkg/cmd/util"
 )
 
@@ -596,17 +597,25 @@ func batchPerform(infos ResourceList, fn func(*resource.Info) error, errs chan<-
 }
 
 func createResource(info *resource.Info) error {
-	obj, err := resource.NewHelper(info.Client, info.Mapping).WithFieldManager(getManagedFieldsManager()).Create(info.Namespace, true, info.Object)
-	if err != nil {
-		return err
-	}
-	return info.Refresh(obj, true)
+	return retry.RetryOnConflict(
+		retry.DefaultRetry,
+		func() error {
+			obj, err := resource.NewHelper(info.Client, info.Mapping).WithFieldManager(getManagedFieldsManager()).Create(info.Namespace, true, info.Object)
+			if err != nil {
+				return err
+			}
+			return info.Refresh(obj, true)
+		})
 }
 
 func deleteResource(info *resource.Info, policy metav1.DeletionPropagation) error {
-	opts := &metav1.DeleteOptions{PropagationPolicy: &policy}
-	_, err := resource.NewHelper(info.Client, info.Mapping).WithFieldManager(getManagedFieldsManager()).DeleteWithOptions(info.Namespace, info.Name, opts)
-	return err
+	return retry.RetryOnConflict(
+		retry.DefaultRetry,
+		func() error {
+			opts := &metav1.DeleteOptions{PropagationPolicy: &policy}
+			_, err := resource.NewHelper(info.Client, info.Mapping).WithFieldManager(getManagedFieldsManager()).DeleteWithOptions(info.Namespace, info.Name, opts)
+			return err
+		})
 }
 
 func createPatch(target *resource.Info, current runtime.Object) ([]byte, types.PatchType, error) {
--- a/pkg/kube/client_test.go
+++ b/pkg/kube/client_test.go
@@ -90,6 +90,13 @@ func newResponse(code int, obj runtime.Object) (*http.Response, error) {
 	return &http.Response{StatusCode: code, Header: header, Body: body}, nil
 }
 
+func newResponseJSON(code int, json []byte) (*http.Response, error) {
+	header := http.Header{}
+	header.Set("Content-Type", runtime.ContentTypeJSON)
+	body := io.NopCloser(bytes.NewReader(json))
+	return &http.Response{StatusCode: code, Header: header, Body: body}, nil
+}
+
 func newTestClient(t *testing.T) *Client {
 	testFactory := cmdtesting.NewTestFactory()
 	t.Cleanup(testFactory.Cleanup)
@@ -100,6 +107,103 @@ func newTestClient(t *testing.T) *Client {
 	}
 }
 
+func TestCreate(t *testing.T) {
+	// Note: c.Create with the fake client can currently only test creation of a single pod in the same list. When testing
+	// with more than one pod, c.Create will run into a data race as it calls perform->batchPerform which performs creation
+	// in batches. The first data race is on accessing var actions and can be fixed easily with a mutex lock in the Client
+	// function. The second data race though is something in the fake client itself in  func (c *RESTClient) do(...)
+	// when it stores the req: c.Req = req and cannot (?) be fixed easily.
+	listA := newPodList("starfish")
+	listB := newPodList("dolphin")
+
+	var actions []string
+	var iterationCounter int
+
+	c := newTestClient(t)
+	c.Factory.(*cmdtesting.TestFactory).UnstructuredClient = &fake.RESTClient{
+		NegotiatedSerializer: unstructuredSerializer,
+		Client: fake.CreateHTTPClient(func(req *http.Request) (*http.Response, error) {
+			path, method := req.URL.Path, req.Method
+			bodyReader := new(strings.Builder)
+			_, _ = io.Copy(bodyReader, req.Body)
+			body := bodyReader.String()
+			actions = append(actions, path+":"+method)
+			t.Logf("got request %s %s", path, method)
+			switch {
+			case path == "/namespaces/default/pods" && method == "POST":
+				if strings.Contains(body, "starfish") {
+					if iterationCounter < 2 {
+						iterationCounter++
+						return newResponseJSON(409, resourceQuotaConflict)
+					}
+					return newResponse(200, &listA.Items[0])
+				}
+				return newResponseJSON(409, resourceQuotaConflict)
+			default:
+				t.Fatalf("unexpected request: %s %s", method, path)
+				return nil, nil
+			}
+		}),
+	}
+
+	t.Run("Create success", func(t *testing.T) {
+		list, err := c.Build(objBody(&listA), false)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		result, err := c.Create(list)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		if len(result.Created) != 1 {
+			t.Errorf("expected 1 resource created, got %d", len(result.Created))
+		}
+
+		expectedActions := []string{
+			"/namespaces/default/pods:POST",
+			"/namespaces/default/pods:POST",
+			"/namespaces/default/pods:POST",
+		}
+		if len(expectedActions) != len(actions) {
+			t.Fatalf("unexpected number of requests, expected %d, got %d", len(expectedActions), len(actions))
+		}
+		for k, v := range expectedActions {
+			if actions[k] != v {
+				t.Errorf("expected %s request got %s", v, actions[k])
+			}
+		}
+	})
+
+	t.Run("Create failure", func(t *testing.T) {
+		list, err := c.Build(objBody(&listB), false)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		_, err = c.Create(list)
+		if err == nil {
+			t.Errorf("expected error")
+		}
+
+		expectedString := "Operation cannot be fulfilled on resourcequotas \"quota\": the object has been modified; " +
+			"please apply your changes to the latest version and try again"
+		if !strings.Contains(err.Error(), expectedString) {
+			t.Errorf("Unexpected error message: %q", err)
+		}
+
+		expectedActions := []string{
+			"/namespaces/default/pods:POST",
+		}
+		for k, v := range actions {
+			if expectedActions[0] != v {
+				t.Errorf("expected %s request got %s", v, actions[k])
+			}
+		}
+	})
+}
+
 func TestUpdate(t *testing.T) {
 	listA := newPodList("starfish", "otter", "squid")
 	listB := newPodList("starfish", "otter", "dolphin")
@@ -108,6 +212,7 @@ func TestUpdate(t *testing.T) {
 	listC.Items[0].Spec.Containers[0].Ports = []v1.ContainerPort{{Name: "https", ContainerPort: 443}}
 
 	var actions []string
+	var iterationCounter int
 
 	c := newTestClient(t)
 	c.Factory.(*cmdtesting.TestFactory).UnstructuredClient = &fake.RESTClient{
@@ -146,6 +251,10 @@ func TestUpdate(t *testing.T) {
 				}
 				return newResponse(200, &listB.Items[0])
 			case p == "/namespaces/default/pods" && m == "POST":
+				if iterationCounter < 2 {
+					iterationCounter++
+					return newResponseJSON(409, resourceQuotaConflict)
+				}
 				return newResponse(200, &listB.Items[1])
 			case p == "/namespaces/default/pods/squid" && m == "DELETE":
 				return newResponse(200, &listB.Items[1])
@@ -199,7 +308,9 @@ func TestUpdate(t *testing.T) {
 		"/namespaces/default/pods/otter:GET",
 		"/namespaces/default/pods/otter:GET",
 		"/namespaces/default/pods/dolphin:GET",
-		"/namespaces/default/pods:POST",
+		"/namespaces/default/pods:POST", // create dolphin
+		"/namespaces/default/pods:POST", // retry due to 409
+		"/namespaces/default/pods:POST", // retry due to 409
 		"/namespaces/default/pods/squid:GET",
 		"/namespaces/default/pods/squid:DELETE",
 	}
@@ -558,3 +669,6 @@ spec:
         ports:
         - containerPort: 80
 `
+
+var resourceQuotaConflict = []byte(`
+{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Operation cannot be fulfilled on resourcequotas \"quota\": the object has been modified; please apply your changes to the latest version and try again","reason":"Conflict","details":{"name":"quota","kind":"resourcequotas"},"code":409}`)


pkg/kube/client.go
```diff
@@ -369,13 +370,55 @@ func batchPerform(infos ResourceList, fn func(*resource.Info) error, errs chan<-
...
 }
 
+// Helper function which checks if error e is a ResourceQuota conflict
+//
+// A ResourceQuota conflict is hit when the following 3 conditions are met:
+// a) Is this a StatusError?
+// b) Is this a Conflict?
+// c) Is status.Details.Kind == "resourcequotas"
+func isResourceQuotaConflict(e error) bool {
+	// Is this a StatusError?
```
cndoit18, line 408:
```golang
var (
  err *apierrors.StatusError
  ok = true
)
err, ok = e.(*apierrors.StatusError)
if !ok {
  return false
}
ok = apierrors.IsConflict(err)
if !ok {
  return false
}
status := err.Status()
return status.Details.Kind == "resourcequotas"
```
I think this is too much nested.
You can change like this

pkg/kube/client.go
```diff
@@ -369,13 +370,55 @@ func batchPerform(infos ResourceList, fn func(*resource.Info) error, errs chan<-
 			kind = currentKind
 		}
 		wg.Add(1)
-		go func(i *resource.Info) {
-			errs <- fn(i)
+		go func(info *resource.Info) {
+			// try 3 times total
+			tries := 3
```
cndoit18, line 375: can it be changed from the outside?

pkg/kube/client.go
```diff
@@ -369,13 +370,55 @@ func batchPerform(infos ResourceList, fn func(*resource.Info) error, errs chan<-
...
+				// is a resourcequota
+				if fnErr == nil ||
+					kind == "resourcequotas" ||
+					!isResourceQuotaConflict(fnErr) {
+					break
+				}
+
+				// Increasing wait: 10-19ms, 40-59ms
+				waitTime := time.Duration(counter*counter*10+rand.Intn(counter*10)) * time.Millisecond
+				time.Sleep(waitTime)
```
cndoit18, line 390: You can use `wait.Pull` from `k8s.io/apimachinery/pkg/util/wait`, not `time.Sleep`

cndoit18:
You can use `fixes #9710` to associate  your issue

pkg/kube/client.go
andreaskaris: you mean move t`ries := 3` 2 lines up to right after wg.Add(1)?

cndoit18: I am just thinking about, for example, it is changed to `func batchPerform(infos ResourceList, fn func(context.Context, *resource.Info) error, errs chan<- error)`

pkg/kube/client.go
andreaskaris: I adjusted this

andreaskaris:
Does the repository have some documentation for how to write and run unit tests? I'd like to cover this in a test :-)

pkg/kube/client.go
andreaskaris: I hope I understood you right. Now, it looks like this:
~~~
func batchPerform(infos ResourceList, fn func(*resource.Info) error, errs chan<- error) {
        var kind string
        var wg sync.WaitGroup
        for _, info := range infos {
                currentKind := info.Object.GetObjectKind().GroupVersionKind().Kind
                if kind != currentKind {
                        wg.Wait()
                        kind = currentKind
                }
                wg.Add(1)

                quotaConflictRetries := 3
                go func(info *resource.Info) {
~~~

pkg/kube/client.go
andreaskaris: wait.Poll polls at regular intervals though and waits for a condition. I'll think about this, but I'd have to create another (nested) function with all the logic and pass that as a function variable?

cndoit18: > wait.Poll polls at regular intervals though and waits for a condition. I'll think about this, but I'd have to create another (nested) function with all the logic and pass that as a function variable?

Sorry, this is my problem, I have a mistake.
You can use `retry.RetryOnConflict` from `k8s.io/client-go/util/retry`

pkg/kube/client.go
```diff
@@ -369,13 +370,60 @@ func batchPerform(infos ResourceList, fn func(*resource.Info) error, errs chan<-
...
+// a) Is this a StatusError?
+// b) Is this a Conflict?
+// c) Is status.Details.Kind == "resourcequotas"
+func isResourceQuotaConflict(e error) bool {
+	var (
+		err *apierrors.StatusError
+		ok  bool
+	)
+	// Is this a StatusError?
+	err, ok = e.(*apierrors.StatusError)
```
cndoit18, line 413: You can use `errors.As`

bacongobbler:
> Does the repository have some documentation for how to write and run unit tests?

https://helm.sh/docs/community/developers/

https://golang.org/pkg/testing/

bacongobbler:
Hey there! I noticed that you mentioned to several other projects in the issue that came to the same conclusion. Because of that I’m curious if this is code you wrote yourself or if it’s been ripped from another program. If the latter we may need to ensure we do some legal due diligence.

andreaskaris (commented):
> Hey there! I noticed that you mentioned several other projects that came to the same conclusion. Because of that I’m curious if this is code you wrote yourself or if it’s been ripped from another program. If the latter we may need to ensure we do some legal due diligence.

Hey!

For better or for worse ( ;-) ) it's code that I came up with myself.

- Andreas

andreaskaris:
> Hey there! I noticed that you mentioned several other projects that came to the same conclusion. Because of that I’m curious if this is code you wrote yourself or if it’s been ripped from another program. If the latter we may need to ensure we do some legal due diligence.

Hey!

For better or for worse ( ;-) ) it's code that I came up with myself.

    Andreas

pkg/kube/client.go
```diff
@@ -369,13 +371,64 @@ func batchPerform(infos ResourceList, fn func(*resource.Info) error, errs chan<-
...
+			Duration: time.Duration(10) * time.Millisecond,
+			Factor:   5,
+			Jitter:   5,
+			Steps:    3,
+		}
+		go func(info *resource.Info) {
+			var fnErr error
+
+			// Retry according to backoff if this is a ResourceQuota conflict
+			_ = retry.RetryOnConflict(
```
cndoit18, line 385: The error should be returned here

pkg/kube/client.go
```diff
@@ -369,13 +371,64 @@ func batchPerform(infos ResourceList, fn func(*resource.Info) error, errs chan<-
...
+				backoff,
```
cndoit18, line 386: you can use `retry.DefaultRetry`

pkg/kube/client.go
```diff
@@ -369,13 +371,64 @@ func batchPerform(infos ResourceList, fn func(*resource.Info) error, errs chan<-
...
-			errs <- fn(i)
+
+		backoff := wait.Backoff{
+			Duration: time.Duration(10) * time.Millisecond,
+			Factor:   5,
+			Jitter:   5,
+			Steps:    3,
+		}
+		go func(info *resource.Info) {
+			var fnErr error
```
cndoit18, line 382: no need to declaration here

pkg/kube/client.go
```diff
@@ -369,13 +371,64 @@ func batchPerform(infos ResourceList, fn func(*resource.Info) error, errs chan<-
...
+
+			// Retry according to backoff if this is a ResourceQuota conflict
+			_ = retry.RetryOnConflict(
+				backoff,
+				func() (err error) {
+					fnErr = fn(info)
```
cndoit18, line 388: `err = fn(info)`

pkg/kube/client.go
andreaskaris: That's the issue which I see with https://pkg.go.dev/k8s.io/client-go/util/retry#RetryOnConflict, though:
~~~
RetryOnConflict is used to make an update to a resource when you have to worry about conflicts caused by other code making unrelated updates to the resource at the same time. fn should fetch the resource to be modified, make appropriate changes to it, try to update it, and return (unmodified) the error from the update function. On a successful update, RetryOnConflict will return nil. If the update function returns a "Conflict" error, RetryOnConflict will wait some amount of time as described by backoff, and then try again. On a non-"Conflict" error, or if it retries too many times and gives up, RetryOnConflict will return an error to the caller. 
~~~

If the code is to catch all Conflicts, then it would also have to run a Get to refresh the resource in question (is this even possible at this point here in the code?). Without a refresh / get of the resource, any other conflict would just loop around and would repeat right away. 
With the resourcequota issue, though, it is sufficient to simply retry as the resource itself is not in Conflict, but the resourcequota.

That's why I have the anonymous function return nil in the case of other errors because I do not want RetryOnConflict to take action if it's not specifically a conflict on resourcequota.

Regardless of that, the anonymous function sets `err` in line:
~~~
err = fn(info)
~~~
 
And that's then appended to the channel: 
~~~
errs <- err
~~~

I'm open to other suggestions, I can't wrap my head around that one otherwise.

andreaskaris: Alternatively, if we can guarantee that the functions which are passed to fn() do a refresh, this could be greatly simplified to:
~~~
                go func(info *resource.Info) {
                        // Retry according to backoff if this is a ResourceQuota conflict
                        errs <- retry.RetryOnConflict(
                                retry.DefaultRetry,
                                func() error {
                                        return fn(info)
                                })
                        wg.Done()
                }(info)
~~~

And the helper function would not be needed, either. Do I see this right?

pkg/kube/client.go
```diff
@@ -369,8 +370,14 @@ func batchPerform(infos ResourceList, fn func(*resource.Info) error, errs chan<-
 			kind = currentKind
 		}
 		wg.Add(1)
-		go func(i *resource.Info) {
-			errs <- fn(i)
+
+		go func(info *resource.Info) {
```
cndoit18, line 374: I re-read the Issue, I don't think this will solve the problem. Because it seems to be used only when it is created.

andreaskaris: batchPerform is used by perform and that's used by Create, Delete, WatchUntilReady

It might not impact the Update action, but my issue is not with an update to the resource. My problem is that on create, kubernetes modifies the resourcequota and throws a 409 conflict. That conflict can be worked around by simply retrying the same create again.

cndoit18: > batchPerform is used by perform and that's used by Create, Delete, WatchUntilReady
> 
> Update uses `Create` when a resource does not yet exist.
> 
> It might not impact an actual update action, but my issue is not with an update to the resource (at least not in my test case ...). My problem is that inside a namespace with a resourcequota, on create of a resource that's covered by the resourcequota, kubernetes modifies the resourcequota and throws a 409 conflict for the resourcequota. That conflict can be worked around by simply retrying the same create again.

I probably understand, can you add some unit tests to prove its effectiveness? please

andreaskaris: Yeah, that's up next but that's where I'm stuck (due to my unfamiliarity with the project). Where would be the best place to simulate a 409 Conflict message after any resource create?

From my lab tests, I add the following debug output:
~~~
// Create creates Kubernetes resources specified in the resource list.
func (c *Client) Create(resources ResourceList) (*Result, error) {
        c.Log("creating %d resource(s)", len(resources))
        createResourceLogger := func(info *resource.Info) error {
                returnErr := createResource(info)
                if returnErr != nil {
                        c.Log(
                                "FAILED Time UnixNano: " + fmt.Sprintf("%v", time.Now().UnixNano()) +
                                        " Name: " + info.Name +
                                        ", Namespace: " + info.Namespace +
                                        ", ERROR: " + returnErr.Error(),
                        )
                        errJson, errErr := json.Marshal(returnErr)
                        if errErr == nil {
                                c.Log("==> ERROR JSON: " + string(errJson))
                        } else {
                                c.Log("Could not marshal ERROR")
                        }
                }
                return returnErr
        }

        if err := perform(resources, createResourceLogger); err != nil {
                return nil, err
        }
        return &Result{Created: resources}, nil
}

/*
// Create creates Kubernetes resources specified in the resource list.
func (c *Client) Create(resources ResourceList) (*Result, error) {
        c.Log("creating %d resource(s)", len(resources))
        if err := perform(resources, createResource); err != nil {
                return nil, err
        }
        return &Result{Created: resources}, nil
}
*/
~~~

And then when I hit the issue (it is super intermittend) I get:
~~~
$ tail -f test_gitlab.log | grep 'ERROR' -B2
(...)
client.go:315: [debug] Starting delete for "gitlab-shared-secrets-1-dzw" Job
client.go:123: [debug] creating 115 resource(s)
client.go:127: [debug] FAILED Time UnixNano: 1621355710894172045 Name: gitlab-prometheus-server, Namespace: gitlab, ERROR: Operation cannot be fulfilled on resourcequotas "quota": the object has been modified; please apply your changes to the latest version and try again
client.go:135: [debug] ==> ERROR JSON: {"ErrStatus":{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Operation cannot be fulfilled on resourcequotas \"quota\": the object has been modified; please apply your changes to the latest version and try again","reason":"Conflict","details":{"name":"quota","kind":"resourcequotas"},"code":409}}
(...)
~~~

Normally, with the above, it would have failed, but with the retry code, helm recovers.

pkg/kube/client_test.go
```diff
@@ -254,12 +256,17 @@ func TestBuild(t *testing.T) {
 }
 
 func TestPerform(t *testing.T) {
+	var fn func(info *resource.Info) error
```
andreaskaris, line 259: I know that what I did here is incredibly ugly; I quickly added this as a proof of concept for a valid unit test.

I tested this with and without my modification and the test passes / fails accordingly:
~~~
[akaris@linux helm]$ grep -i perform /tmp/test
=== RUN   TestPerform
=== RUN   TestPerform/Valid_input
=== RUN   TestPerform/Empty_manifests
=== RUN   TestPerform/Conflict_409
--- PASS: TestPerform (0.04s)
    --- PASS: TestPerform/Valid_input (0.01s)
    --- PASS: TestPerform/Empty_manifests (0.01s)
    --- PASS: TestPerform/Conflict_409 (0.02s)
~~~

The test also reflects what I see  in real life: the exact same call might return a Conflict once, and then the next call to e.g. Create does not return a Conflict for ResourceQuota. The JSON for the StatusError directly comes from my debugging, so it's "real life" output.

I'm looking forward to reading your suggestions, otherwise I'll push some prettier code tomorrow :-)

dperaza4dustbit (commented):
I was able to recreate this issue on OCP 4.6 and 4.7. On 4.7 it was reproduced more often. I have been testing since this morning with this fix and it seems to do the job.

pkg/kube/client_test.go
```diff
@@ -254,32 +256,79 @@ func TestBuild(t *testing.T) {
...
```
cndoit18, line 259: I don't think it should be added to this place.
You should add a fake response in TestClient

https://github.com/helm/helm/blob/b1e24764325135b364a7f81fb0aab683d190c546/pkg/kube/client_test.go#L121

andreaskaris: Hi,

Do you mean TestBuild?
~~~
func TestBuild(t *testing.T) {
        tests := []struct {
                name      string
                namespace string
                reader    io.Reader
                count     int
                err       bool
        }{
                {
                        name:      "Valid input",
                        namespace: "test",
                        reader:    strings.NewReader(guestbookManifest),
                        count:     6,
                }, {
                        name:      "Valid input, deploying resources into different namespaces",
                        namespace: "test",
                        reader:    strings.NewReader(namespacedGuestbookManifest),
                        count:     1,
                },
        }

        c := newTestClient(t)
        for _, tt := range tests {
                t.Run(tt.name, func(t *testing.T) {
                        // Test for an invalid manifest
                        infos, err := c.Build(tt.reader, false)
                        if err != nil && !tt.err {
                                t.Errorf("Got error message when no error should have occurred: %v", err)
                        } else if err != nil && strings.Contains(err.Error(), "--validate=false") {
                                t.Error("error message was not scrubbed")
                        }

                        if len(infos) != tt.count {
                                t.Errorf("expected %d result objects, got %d", tt.count, len(infos))
                        }
                })
        }
}
~~~

There is no function `TestClient`.

andreaskaris: I understand what you mean, now. perform only catches batch creation / deletion. While create / delete call perform, update does not call it. So in the UpdateTest, I cannot create my unit test (because the logic which I added to perform would not catch this issue).  Update calls createResource(info) directly, whereas Create calls perform(infolist, createResource), the same for delete. But this would mean that I have to create a wrapper function for the 3 atomic actions: create, update, delete, correct?

andreaskaris: alternatively, update could call "perform" ;-)

andreaskaris: I pushed a new suggestion. That should cover what you meant (so, retry create, update, delete on conflict) by creating wrapper functions for each of the original methods. I also adjusted the unit test. I yet have to test this in my environment to see if this really fixes the issue that I'm hitting, but it should. Any suggestions to make this prettier are more than welcome. Also, are there any other methods which would need to be wrapped?

cndoit18: Sorry, please forgive me for bad English, so that there is no expression clear

I just mean it to recommend that we build an error response in fakehttpClient, used to do unit testing.

like this
```
Client: fake.CreateHTTPClient(func(req *http.Request) (*http.Response, error) {
			if req.Method == http.MethodPost {}
}
```

cndoit18: Sorry, please forgive me for bad English, so that there is no expression clear

I just mean it to recommend that we build an error response in fakehttpClient, used to do unit testing.

like this
```
Client: fake.CreateHTTPClient(func(req *http.Request) (*http.Response, error) {
}
```

andreaskaris: Hi,

Please see now. I moved the retry logic from "perform" to `createResource` and `deleteResource`. So, when `Update` calls these methods, it will benefit from the retry, too.

`updateResource` is more complex and I did not dare change it. Particularly because the issue which I found is on `createResource`. Therefore, I did not modify the `updateResource` method and it does not retry at the moment. 

I added a 409 Conflict in `TestUpdate` and I created a new test method `TestCreate` which runs 2 different tests. 

I also found DATA RACEs while testing when running `TestCreate` with a list that contains more than 1 pod. That might be an issue with tests or with the existing code (`perform`/`batchPerform` call 2 levels of nested go functions), it's not part of anything new that I added.

Please let me know what you think.

- Andreas

pkg/kube/client.go
```diff
@@ -377,18 +378,26 @@ func batchPerform(infos ResourceList, fn func(*resource.Info) error, errs chan<-
 }
 
 func createResource(info *resource.Info) error {
-	obj, err := resource.NewHelper(info.Client, info.Mapping).Create(info.Namespace, true, info.Object)
-	if err != nil {
-		return err
-	}
-	return info.Refresh(obj, true)
+	return retry.RetryOnConflict(
```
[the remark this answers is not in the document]
andreaskaris: Hi,

Do you mean:
~~~
func createResource(info *resource.Info) error {
	return retry.RetryOnConflict(
		retry.DefaultRetry,
		func() error {
			return createResourceNoRetry(info)
		})
}

func createResourceNoRetry(info *resource.Info) error {
	obj, err := resource.NewHelper(info.Client, info.Mapping).Create(info.Namespace, true, info.Object)
	if err != nil {
		fmt.Println("here")
		e, _ := json.Marshal(err)
		fmt.Println(string(e))
		return err
	}
	return info.Refresh(obj, true)
}
~~~

Or:
~~~
func batchPerform(infos ResourceList, fn func(*resource.Info) error, errs chan<- error) {
	var kind string
	var wg sync.WaitGroup
	for _, info := range infos {
		currentKind := info.Object.GetObjectKind().GroupVersionKind().Kind
		if kind != currentKind {
			wg.Wait()
			kind = currentKind
		}
		wg.Add(1)

		go func(info *resource.Info) {
			// Retry if this is a 409 Conflict
			errs <- retry.RetryOnConflict(
				retry.DefaultRetry,
				func() error {
					return fn(info)
				})
			wg.Done()
		}(info)
	}
}
~~~

If I add the change to batchPerform, there is a problem: `updateResource` calls `createResource` and `deleteResource` directly. It does not call `perform`. So, changing `createResource` and `deleteResource` would catch more issues than just chaning `batchPerform`.

pkg/kube/client_test.go
```diff
@@ -101,6 +108,102 @@ func newTestClient(t *testing.T) *Client {
...
+			_, _ = io.Copy(bodyReader, req.Body)
+			body := bodyReader.String()
+			actions = append(actions, path+":"+method)
+			t.Logf("got request %s %s", path, method)
+			switch {
+			case path == "/namespaces/default/pods/squid" && method == "GET":
+				return newResponse(404, notFoundBody())
+			case path == "/namespaces/default/pods" && method == "POST":
+				if strings.Contains(body, "starfish") {
+					if iterationCounter < 2 {
```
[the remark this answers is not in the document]
andreaskaris: https://golang.org/pkg/sync/#Once

From the example:
~~~
package main

import (
	"fmt"
	"sync"
)

func main() {
	var once sync.Once
	onceBody := func() {
		fmt.Println("Only once")
	}
	done := make(chan bool)
	for i := 0; i < 10; i++ {
		go func() {
			once.Do(onceBody)
			done <- true
		}()
	}
	for i := 0; i < 10; i++ {
		<-done
	}
}
~~~

That looks more complex than checking if the counter is < x. And the function in onceBody does not return a value.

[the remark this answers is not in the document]
cndoit18: You are right(ignore my comment)

pkg/kube/client_test.go
```diff
@@ -101,6 +108,102 @@ func newTestClient(t *testing.T) *Client {
...
+		NegotiatedSerializer: unstructuredSerializer,
+		Client: fake.CreateHTTPClient(func(req *http.Request) (*http.Response, error) {
+			path, method := req.URL.Path, req.Method
+			bodyReader := new(strings.Builder)
+			_, _ = io.Copy(bodyReader, req.Body)
+			body := bodyReader.String()
+			actions = append(actions, path+":"+method)
+			t.Logf("got request %s %s", path, method)
+			switch {
+			case path == "/namespaces/default/pods/squid" && method == "GET":
```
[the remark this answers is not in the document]
andreaskaris: Fixed

pkg/kube/client.go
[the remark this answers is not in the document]
andreaskaris: Personally, I'd prefer to update `createResource` and `deleteResource` directly. After some thought, `batchPerform` is not the correct place to address this. Not all methods (e.g. `Update`) call `batchPerform`. The only way to add this to `batchPerform` would be that all methods use `perform` / `batchPerform`, but that would require more changes than I intended. I hope that makes sense.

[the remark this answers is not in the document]
andreaskaris: Do you prefer this:
~~~
func createResource(info *resource.Info) error {
	return retry.RetryOnConflict(
		retry.DefaultRetry,
		func() error {
			return createResourceNoRetry(info)
		})
}

func createResourceNoRetry(info *resource.Info) error {
	obj, err := resource.NewHelper(info.Client, info.Mapping).Create(info.Namespace, true, info.Object)
	if err != nil {
		fmt.Println("here")
		e, _ := json.Marshal(err)
		fmt.Println(string(e))
		return err
	}
	return info.Refresh(obj, true)
}
~~~

Over the current suggestion with the nested anonymous function? If so, I will amend the commit and revert to this.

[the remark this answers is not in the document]
andreaskaris: Hi,

Just to reiterate - the simpler implementation here [0] IMO is not the correct place to implement a retry. I think that a retry should be implemented in the individual createResource/deleteResource/updateResource methods because createResource/deleteResource are also individually called from `Update` (`Update` does not call `perform`, thus fixing this in `batchPerform` would not fix this for `Update`.

The current suggestion [1] addresses the issue at the source, in `createResource` and `deleteResource`. It also is IMO not complex. It's just a wrapper around the current logic. I could also imagine a scenario with a function `func retryResource(info *resource.Info, fn func (*resource.Info) error) which could be called everywhere like this:
`retryResource(info, createResource)`, `retryResource(updateResource)`. But that's actually more complex than [1] at this point in time.
In order to avoid further complexity, I also left out `updateResource`, as I dread that changing this one might introduce bugs or other complications and the issue with `resourcequota` shows up only when the actual quota changes.

The current unit tests show that [1] fixes what needed to be fixed. [0] Does not pass the new `TestUpdate` variation.
If you prefer [0], then I can recommit this (including a revert of `TestUpdate` and only using `TestCreate` to test this).

- Andreas 

[0]
~~~
func batchPerform(infos ResourceList, fn func(*resource.Info) error, errs chan<- error) {
	var kind string
	var wg sync.WaitGroup
	for _, info := range infos {
		currentKind := info.Object.GetObjectKind().GroupVersionKind().Kind
		if kind != currentKind {
			wg.Wait()
			kind = currentKind
		}
		wg.Add(1)

		go func(info *resource.Info) {
			// Retry if this is a 409 Conflict
			errs <- retry.RetryOnConflict(
				retry.DefaultRetry,
				func() error {
					return fn(info)
				})
			wg.Done()
		}(info)
	}
}
~~~

[1]
~~~
func createResource(info *resource.Info) error {
	return retry.RetryOnConflict(
		retry.DefaultRetry,
		func() error {
			obj, err := resource.NewHelper(info.Client, info.Mapping).Create(info.Namespace, true, info.Object)
			if err != nil {
				return err
			}
			return info.Refresh(obj, true)
		})
}

func deleteResource(info *resource.Info) error {
	return retry.RetryOnConflict(
		retry.DefaultRetry,
		func() error {
			policy := metav1.DeletePropagationBackground
			opts := &metav1.DeleteOptions{PropagationPolicy: &policy}
			_, err := resource.NewHelper(info.Client, info.Mapping).DeleteWithOptions(info.Namespace, info.Name, opts)
			return err
		})
}
~~~

gufik3333 (approved):

andreaskaris:
Hi,

Is there anything else that needs to be done to move this forward? Our customers are being affected by the lack of retries in helm (https://bugzilla.redhat.com/show_bug.cgi?id=1934604) and I'd like to make progress on this. I am happy to change / improve this pull request.

@cndoit18 

Thanks!

- Andreas

cndoit18:
> Hi,
> 
> Is there anything else that needs to be done to move this forward? Our customers are being affected by the lack of retries in helm (https://bugzilla.redhat.com/show_bug.cgi?id=1934604) and I'd like to make progress on this. I am happy to change / improve this pull request.
> 
> @cndoit18
> 
> Thanks!
> 
> * Andreas

I do not have the authority, you can ask the community

andreaskaris:
Sorry, my bad, I thought you were a member of the project :-)  Thanks for reviewing this, then.

@bacongobbler What are the next steps? :-)

cndoit18:
> Sorry, my bad, I thought you were a member of the project :-) Thanks for reviewing this, then.
> 
> @bacongobbler What are the next steps? :-)

You will need to get approvals from two members, which is expected to be a long wait.

dperaza4dustbit:
I will bring it up on the community call today to bring some attention

bacongobbler:
As discussed in the dev call, it would be great to understand

1. Who is impacted by this bug
2. How to test/reproduce the issue
3. What is the severity of this issue (blocker, workaround available, totally breaks Helm usage, etc)

As far as I can tell, this particular Kubernetes issue has been [open since 2018](https://github.com/kubernetes/kubernetes/issues/67761). This is the first time it's been brought to the attention of the maintainers since it was opened, so it's unclear how many people are impacted by this issue and whether it needs to be expedited to Helm 3.6.1, or if it can be triaged for Helm 3.7.0 (to be released Sept 9th [as per HIP 2](https://github.com/helm/community/blob/main/hips/hip-0002.md)).

In addition, if you can provide a way where reviewers can reproduce the issue and test the solution against the bug, that would be helpful as well.

Thanks!

dperaza4dustbit:
This is the bugzilla issue to provide some context on this: https://bugzilla.redhat.com/show_bug.cgi?id=1934604. The customer reporting will like to stay anonymous but basically is a global semiconductor manufacturer. I was easily able to recreate using gitlab chart (gitlab repo: https://charts.gitlab.io/ ) in OpenShift 4.6 and 4.7 with the following script:

`cat test_gitlab.sh
#!/usr/bin/env bash

set -e

while true; do 
    helm install --debug gitlab gitlab/gitlab --timeout 600s --set global.hosts.domain=example.com --set global.hosts.externalIP=10.10.10.10 --set certmanager-issuer.email=me@example.com
    sleep 5
    helm uninstall gitlab
done` 

To set the test I run this:
`#!/usr/bin/env bash

set -e

oc new-project gitlab
oc apply -f quota.yaml
oc apply -f cpu_limit.yaml
oc adm policy add-scc-to-user privileged -z default
oc adm policy add-scc-to-user privileged -z gitlab-shared-secrets`

Here is the quota resource:

`bash-5.1$ cat quota.yaml 
apiVersion: v1
kind: ResourceQuota
metadata:
  name: quota
spec:
  hard:
    configmaps: "62"
    limits.cpu: "170"
    limits.memory: 260Gi
    openshift.io/imagestreams: "0"
    persistentvolumeclaims: "50"
    pods: "100"
    replicationcontrollers: "50"
    requests.cpu: "71"
    requests.memory: 205Gi
    requests.storage: 350Gi
    limits.cpu: "71"
    limits.memory: 205Gi
    requests.storage: 350Gi
    resourcequotas: "1"
    secrets: "512"
    services: "70"
    services.nodeports: "5"`

Here is the limits resource:

`bash-5.1$ cat cpu_limit.yaml 
apiVersion: v1
kind: LimitRange
metadata:
  name: cpu-limit-range
spec:
  limits:
  - default:
      memory: 256Mi
      cpu: 125m
    defaultRequest:
      cpu: 125m
      memory: 128Mi
    type: Container`

andreaskaris:
Hi,

Apart from this and the master branch having diverged, how does the helm project deal with upstream contributions?

Because this pull request has been open for 3 months now. I'm open to discussing the fix, I'm open to modifying it, I'm even open to dropping it (though I would have a lot of valid arguments against dropping it). 

Instead, this pull request has been kind of ignored for the last 3 months (the same goes for any other contribution which I made around that time). In other Open Source projects, I am used to seeing some fruitful discussion around any outside contribution and some welcoming attitude towards people who are willing to contribute to the project.

My hurt pride aside (I'm kidding, really), I am just wondering about the workflow and how we'd ever fix https://bugzilla.redhat.com/show_bug.cgi?id=1934604 or any other issue  or part that might be improved that one found in helm.

@bacongobbler Is the (current version) of the helm project on the backburner?

- Andreas

geoberle:
I would also love to see some sort of reaction/discussion on this PR since I'm heavily affected by the issue it addresses. Thanks for your time. 

- Gerd

bakito:
Is there any change this PR gets merged soon?

patrickmx:
Any chance this will be merged soon?

amitsingla:
We are facing the similar issue when we upgraded the kube to 1.20.15 and moved to Kube services to Binary. It would be great if get some update on merging this PR soon?

pkg/kube/client.go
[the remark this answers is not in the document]
aslafy-z: @cndoit18 Any inputs?

nikita698:
Is there any chance this will be merged? We are using OpenShift 4.10 and we are greatly impacted by this, a lot of our services are deployed by using helm and we experience this issue a lot. 
Is there at least some sensible work around for the time being?

mrmadira:
We are yet another team that is looking forward to this PR being merged!

ptrnull:
Hey @mattfarina , Milestone 3.11.0 was closed and this PR from 2021 is still open, is it going to be merged for Milestone 3.12.0 by May 10, 2023? 
Thanks

nightguide:
Hi, I got the same issue.

```
Error: release cm-consumer failed, and has been uninstalled due to atomic being set: Operation cannot be fulfilled on resourcequotas "object-counts": the object has been modified; please apply your changes to the latest version and try again
helm.go:84: [debug] Operation cannot be fulfilled on resourcequotas "object-counts": the object has been modified; please apply your changes to the latest version and try again
```

Is there a client side fix planned?

antonroman:
We see this problem very often in many different deployments. Is it planned to be merged? Thanks

darshan-karia:
@mattfarina any reason why this keeps getting pushed to next release and isn't merging? This is an issue we're facing since at least 3 years, and we are using forked helm-idk with this PR as patch applied to it since almost 2-3 years and it is working perfectly fine for us. But, due to nature of maintaining fork, it is becoming very cumbersome.

In fact, due to delay in merge, this PR is actually outdated and needs conflict resolution, which I am sure we can get it done, as it is simple enough, provided there is no issue with actual logic or test.

andreaskaris:
I rebased it again and I also sent an email to Matt. With that said, I'm totally open to changes (especially to the unit tests), so please let me know what you think

Vovanys:
Hi. Please merge this PR

sabre1041 (approved):
LGTM

mattfarina:
Note, the force push was used to kick-off CI and did not change any code (as can be seen in the diff provided by GitHub).

mattfarina (approved):

PR #9623, opened 2021-04-20
Fix install scripts, hardcode latest/last v2 release

link108:
**What this PR does / why we need it**:
Helm's install script, `get`, is currently broken due to no recent v2 release. When attempting to get the latest tag, no tag is found as there has not been a recent v2 release. I recognize that v2 has EOL'ed, I figured I'd make this PR anyways since its more readable/maintainable (not relying on http parsing).  

**If applicable**:
- [ ] this PR contains documentation
- [ ] this PR contains unit tests
- [x] this PR has been tested for backwards compatibility

PR #9715, opened 2021-05-18
Fix: Tweak scripts/get to fetch latest v2 tag from releases

### LINKED ISSUE #9714: Bug: scripts/get not installing latest helm 2.x version
I'm using the script [helm/scripts/get](https://github.com/helm/helm/blob/main/scripts/get) to install latest version of helm2 but it is erroring out as below.

```bash
$ wget -O - https://raw.githubusercontent.com/kubernetes/helm/master/scripts/get 2>/dev/null | bash

Downloading https://get.helm.sh/helm--linux-amd64.tar.gz
SHA sum of /tmp/helm-installer-sWGDPd/helm--linux-amd64.tar.gz does not match. Aborting.
Failed to install helm
```

Looks like the script is checking the GitHub releases page for a version 2.x but running into error as no release with v2.0 is available at the first page itself. v2.x was pushed out to second page after recent releases.

ashithwilson:
## Motivation

Fixes #9714

The script [helm/scripts/get](https://github.com/helm/helm/blob/main/scripts/get) to install latest version of helm2 errors out like so.

```bash
<SNIPPED>
Downloading https://get.helm.sh/helm--linux-amd64.tar.gz
SHA sum of /tmp/helm-installer-sWGDPd/helm--linux-amd64.tar.gz does not match. Aborting.
<SNIPPED>
```

The script finds the latest release of `v2.x` by checking the GitHub releases page. However, as new releases are added,  v2.x got pushed to second page which fails the script. 

## Fix

Instead of using the GitHub release url, use GitHub API with `per_page=100`. This gives a JSON output of the last 100 releases.

### DIFF
--- a/scripts/get
+++ b/scripts/get
@@ -77,12 +77,12 @@ verifySupported() {
 checkDesiredVersion() {
   if [ "x$DESIRED_VERSION" == "x" ]; then
     # Get tag from release URL
-    local release_url="https://github.com/helm/helm/releases"
+    local release_url="https://api.github.com/repos/helm/helm/tags?per_page=100"
     if type "curl" > /dev/null; then
 
-      TAG=$(curl -Ls $release_url | grep 'href="/helm/helm/releases/tag/v2.[0-9]*.[0-9]*\"' | grep -v no-underline | head -n 1 | cut -d '"' -f 2 | awk '{n=split($NF,a,"/");print a[n]}' | awk 'a !~ $0{print}; {a=$0}')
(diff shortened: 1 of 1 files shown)


bacongobbler:
Duplicate of #9623. Earlier iterations of #9623 also proposed switching to api.github.com. I listed the reasons why we cannot do that in https://github.com/helm/helm/pull/9623#issuecomment-822997291.

for that reason, I am closing this in favour of that PR.

PR #9718, opened 2021-05-19
Improve markdown rendering of commands help for shell completion

zigarn:
**What this PR does / why we need it**: Allow to have a readable doc on the onine Helm doc website

**Special notes for your reviewer**: Cf. helm/helm-www#1100

**If applicable**:

### DIFF
--- a/cmd/helm/completion.go
+++ b/cmd/helm/completion.go
@@ -33,33 +33,41 @@ const bashCompDesc = `
 Generate the autocompletion script for Helm for the bash shell.
 
 To load completions in your current shell session:
-$ source <(helm completion bash)
+
+    source <(helm completion bash)
 
 To load completions for every new session, execute once:
-Linux:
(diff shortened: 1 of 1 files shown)


marckhouzam (approved):
LGTM

This will make the helm.sh docs look much better for the completion commands.  Thanks @zigarn!

As the patch is officially bigger than XS, it will need another maintainer to approve.

I suggest we pick this into the 3.6 RC so that the docs look better.

bacongobbler (approved):
Also LGTM. I'll leave it to the release engineer in charge to determine if this should go into 3.6 or not. If not we'll merge this and it'll be available in Helm 3.7. Thanks!

marckhouzam:
I've asked and the policy is that only regressions can be added to an RC.  So this will need to wait until the 3.7 release.
With some luck, the RC will have no problems and it will be the final 3.6 release, so it makes sense not to add anything unless it is absolutely critical.
I will merge into main.

PR #9739, opened 2021-05-25
fix(helm): Add WARNING when --repo parameter is ignored

### LINKED ISSUE #9731: `helm render` with `--repo` fails to find `Chart.yaml` with a values file in subdirectory with the same name as the chart
Output of `helm version`: `version.BuildInfo{Version:"v3.5.4", GitCommit:"1b5edb69df3d3a08df77c9902dc17af864ff05d1", GitTreeState:"clean", GoVersion:"go1.15.11"}`

Output of `kubectl version`: `Server Version: version.Info{Major:"1", Minor:"18+", GitVersion:"v1.18.9-eks-d1db3c", GitCommit:"d1db3c46e55f95d6a7d3e5578689371318f95ff9", GitTreeState:"clean", BuildDate:"2020-10-20T22:18:07Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}`

Y'all, this is a really bizarre one.

`helm render` fails when you use `--repo`, *and* provide a values file that is in a subdirectory with the same name as the chart name (not release name), that is exactly one layer deeper than the current directory. Consider a directory tree:

```
% tree helm-bug 
helm-bug
└── cert-manager
    └── values.yaml

1 directory, 1 file
```

`values.yaml` is empty here; its content doesn't matter.

This works fine:

```
helm template cert-manager cert-manager --repo https://charts.jetstack.io -f helm-bug/cert-manager/values.yaml
```

If I change directory into `helm-bug/cert-manager/`, this works fine:

```
helm template cert-manager cert-manager --repo https://charts.jetstack.io -f values.yaml
```

If I change directory into `helm-bug/`, this works fine:

```
helm repo add jetstack https://charts.jetstack.io
helm repo update
helm template cert-manager jetstack/cert-manager -f cert-manager/values.yaml
```

However, if I change directory into `helm-bug/`, *none* of the following work:

```
helm template cert-manager cert-manager --repo https://charts.jetstack.io -f cert-manager/values.yaml
helm template cert-manager cert-manager --repo https://charts.jetstack.io -f ./cert-manager/values.yaml
helm template cert-manager cert-manager --repo https://charts.jetstack.io -f "$(readlink -f cert-manager/values.yaml)"
```

Moreover, if I rename the `cert-manager` directory to `certmanager`, it works fine. Similarly, if I omit the values file, this works fine:

```
helm template cert-manager cert-manager --repo https://charts.jetstack.io
```

So something about passing that values file in makes helm look for the chart locally instead of in the repo that I've told it to use.

Debug output is pretty terse:

```
% helm --debug template cert-manager cert-manager --repo https://charts.jetstack.io -f cert-manager/values.yaml
install.go:173: [debug] Original chart version: ""
install.go:190: [debug] CHART PATH: /home/stpierre/helm-bug/cert-manager

Error: Chart.yaml file is missing
helm.go:81: [debug] Chart.yaml file is missing
helm.sh/helm/v3/pkg/chart/loader.LoadFiles
        /home/circleci/helm.sh/helm/pkg/chart/loader/load.go:158
helm.sh/helm/v3/pkg/chart/loader.LoadDir
        /home/circleci/helm.sh/helm/pkg/chart/loader/directory.go:119
helm.sh/helm/v3/pkg/chart/loader.DirLoader.Load
        /home/circleci/helm.sh/helm/pkg/chart/loader/directory.go:41
helm.sh/helm/v3/pkg/chart/loader.Load
        /home/circleci/helm.sh/helm/pkg/chart/loader/load.go:62
main.runInstall
        /home/circleci/helm.sh/helm/cmd/helm/install.go:199
main.newTemplateCmd.func2
        /home/circleci/helm.sh/helm/cmd/helm/template.go:73
        /go/pkg/mod/github.com/spf13/cobra@v1.1.1/command.go:850
        /go/pkg/mod/github.com/spf13/cobra@v1.1.1/command.go:958
        /go/pkg/mod/github.com/spf13/cobra@v1.1.1/command.go:895
main.main
        /home/circleci/helm.sh/helm/cmd/helm/helm.go:80
runtime.main
        /usr/local/go/src/runtime/proc.go:204
runtime.goexit
        /usr/local/go/src/runtime/asm_amd64.s:1374
```

That seems to indicate that it's looking for the `Chart.yaml` in the same directory as `values.yaml` instead of in the remote chart.

There are workarounds available, but the command I've given is run by `skaffold` so unfortunately the least disruptive one -- adding the repo and templating the chart from there -- isn't an option.

andreaskaris:
When helm finds a chart under a local (relative) path, it will
always use that chart. Thus, helm will ignore the --repo parameter.
Issue a warning message to make users aware of this behavior.

Closes #9731

Signed-off-by: Andreas Karis <ak.karis@gmail.com>

**If applicable**:

### DIFF
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -418,3 +418,8 @@ func (c *Configuration) Init(getter genericclioptions.RESTClientGetter, namespac
 
 	return nil
 }
+
+func warning(format string, v ...interface{}) {
+	format = fmt.Sprintf("WARNING: %s\n", format)
+	fmt.Fprintf(os.Stderr, format, v...)
+}
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -643,6 +643,11 @@ func (c *ChartPathOptions) LocateChart(name string, settings *cli.EnvSettings) (
 				return "", err
 			}
 		}
+		if c.RepoURL != "" {
+			warning("Chart found at local path '%v'. Ignoring Repository URL '%v.'",
+				abs,
+				c.RepoURL)
+		}
 		return abs, nil
 	}
 	if filepath.IsAbs(name) || strings.HasPrefix(name, ".") {


pkg/action/action.go
```diff
@@ -418,3 +418,8 @@ func (c *Configuration) Init(getter genericclioptions.RESTClientGetter, namespac
 
 	return nil
 }
+
+func warning(format string, v ...interface{}) {
+	format = fmt.Sprintf("WARNING: %s\n", format)
+	fmt.Fprintf(os.Stderr, format, v...)
```
cndoit18, line 424: If I refer to the action package and want to change the error output stream, what do I need to do?

pkg/action/install.go
```diff
@@ -643,6 +643,11 @@ func (c *ChartPathOptions) LocateChart(name string, settings *cli.EnvSettings) (
 				return "", err
 			}
 		}
+		if c.RepoURL != "" {
+			warning("Chart found at local path '%v'. Ignoring Repository URL '%v.'",
```
cndoit18, line 647: In `action.Configuration`, we have the log field. It might be possible to use it to output. WDYT

cndoit18 (changes requested):

pkg/action/install.go
andreaskaris: Thanks.

There's a log field in:
~~~
 69 type Install struct {
 70         cfg *Configuration
~~~

That's a private attribute. I could access that with a Getter which I'd have to create (`action.Install.GetLog()`) and then do:
~~~
172 func runInstall(args []string, client *action.Install, valueOpts *values.Options, out io.Writer) (*release.Releas    e, error) {
(...)
185         cp, err := client.ChartPathOptions.LocateChart(chart, settings, client.GetLogger())
(...)
~~~

And I'd have to change that everywhere where I encounter `LocateChart`.

The problem then is that I'd have to change that for action.Show, as well, but action.show does not have a Configuration object:
~~~
cmd/helm/show.go:179:74: client.GetLog undefined (type *action.Show has no field or method GetLog)
~~~

pkg/action/action.go
cndoit18: > I simply copied that from cmd/helm/helm.go
> 
> That's used in:
> 
> cmd/helm/helm.go
> 
> ```
> func main() {
>         actionConfig := new(action.Configuration)
>         cmd, err := newRootCmd(actionConfig, os.Stdout, os.Args[1:])
>         if err != nil {
>                 warning("%+v", err)
>                 os.Exit(1)
>         }
> ```
> 
> And several other places in package `main` and the implementation is exactly the same:
> 
> ```
>  56 func warning(format string, v ...interface{}) {
>  57         format = fmt.Sprintf("WARNING: %s\n", format)
>  58         fmt.Fprintf(os.Stderr, format, v...)
>  59 }
> ```
> 
> I'd really prefer to get and pass func Log from somewhere, but I don't know how to pass a Log method into `func (c *ChartPathOptions) LocateChart(name string, settings *cli.EnvSettings) (string, error) {` without making this a really complex change.

The action package will be called by people, it is not the same as the cmd package. So I don't think it can be copied directly over

pkg/action/install.go
cndoit18: > Thanks.
> 
> There's a log field in:
> 
> ```
>  69 type Install struct {
>  70         cfg *Configuration
> ```
> 
> That's a private attribute. I could access that with a Getter which I'd have to create (`action.Install.GetLog()`) and then do:
> 
> ```
> 172 func runInstall(args []string, client *action.Install, valueOpts *values.Options, out io.Writer) (*release.Releas    e, error) {
> (...)
> 185         cp, err := client.ChartPathOptions.LocateChart(chart, settings, client.GetLogger())
> (...)
> ```
> 
> And I'd have to change that everywhere where I encounter `LocateChart`.
> 
> The problem then is that I'd have to change that for action.Show, as well, but action.show does not have a Configuration object:
> 
> ```
> cmd/helm/show.go:179:74: client.GetLog undefined (type *action.Show has no field or method GetLog)
> ```

Yes, so I think the warning should be thrown to the structure that references it to handle
It should not be handled in LocateChart

PR #9775, opened 2021-06-07
Fix helm install --create-namespace flag RBAC issue

distorhead:
Allow case when namespace already exists in the cluster and user does not have a permission to create namespaces. `helm install --create-namespace` would not try to create namespace (and receive an unnecessary failure) in such case.

This change makes `--create-namespace` flag behaviour more idempotent.

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -309,7 +309,7 @@ func (i *Install) Run(chrt *chart.Chart, vals map[string]interface{}) (*release.
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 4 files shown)


cndoit18 (commented):
Hello

bacongobbler:
> `helm install --create-namespace` would not try to create namespace (and receive an unnecessary failure)

I would argue that the user would want to know that namespace creation failed due to invalid RBAC permissions if they provided the `--create-namespace` flag.

joejulian:
I agree with bacongobbler. Closing.

Issue #9738, opened 2021-05-25
Unable to get an update from the chart repository: tag explicitly required (OCI ACR chart dependency)


chuanm:
Output of `helm version`: 

`version.BuildInfo{Version:"v3.6.0-rc.1", GitCommit:"7f2df6467771a75f5646b7f12afb408590ed1755", GitTreeState:"clean", GoVersion:"go1.16.3"}`

Output of `kubectl version`: 
```

Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.0", GitCommit:"e8462b5b5dc2584fdcd18e6bcfe9f1e4d970a529", GitTreeState:"clean", BuildDate:"2019-06-19T16:40:16Z", GoVersion:"go1.12.5", Compiler:"gc", Platform:"windows/amd64"}
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.9", GitCommit:"6c90dbd9d6bb1ae8a4c0b0778752be06873e7c55", GitTreeState:"clean", BuildDate:"2021-03-22T23:02:49Z", GoVersion:"go1.15.8", Compiler:"gc", Platform:"linux/amd64"}
```

Cloud Provider/Platform (AKS, GKE, Minikube etc.):  AKS using OCI for repositories

I'm having issues including an OCI ACR as a chart dependency as follows:

```
apiVersion: v2
name: test-operator
description: test-operator
version: 0.0.1

dependencies:
- name: dependency-chart
  version: "0.5.4"
  repository: "oci://someaddresshere.azurecr.io/helm"

```

I'm getting back the following error when I run `helm dep up` on the chart to download the dependency:

```

Getting updates for unmanaged Helm repositories...
...Unable to get an update from the "oci://someaddresshere.azurecr.io/helm/" chart repository:
        tag explicitly required
```

Not quite sure what is going on here, since doing a `helm chart pull` for the corresponding repo as follows works just fine:

`helm chart pull someaddresshere.azurecr.io/helm/dependency-chart:0.5.4`

I'm adhering to the following piece of documentation: [https://helm.sh/docs/topics/registries/#specifying-dependencies](https://helm.sh/docs/topics/registries/#specifying-dependencies), so I'm not sure why this doesn't work

EDIT: So my inability to download the dependencies ended up being the result of something else, but I still get the "tag explicitly required" error, is this some sort of bug?

jdolitsky:
Potentially fixed by https://github.com/helm/helm/pull/9539

[...]

Issue #9694, opened 2021-05-13
OCI: no version range support for dependencies/`helm install`


TZdybel:
I'm trying to make my helm chart with dependencies to get newest version I pass to it.
I see that "latest" tag is not possible so I've tried with semantic versions like "~1.x" which should work according to documentation.
It works for things like mongodb chart from bitnami repo, but not for my Azure Container Registry.
My `Chart.yaml`:
```yaml
apiVersion: v2
name: somename
description: A Helm chart for Kubernetes
version: 0.1.0
appVersion: "1.0"
dependencies:
  - name: common
    version: 1.0.0
    repository: "oci://myazurecr.azurecr.io/helm"
```
Output from `helm dependency update`:
```
Getting updates for unmanaged Helm repositories...
...Unable to get an update from the "oci://myazurecr.azurecr.io/helm" chart repository:
        tag explicitly required
Hang tight while we grab the latest from your chart repositories...
...Successfully got an update from the "kured" chart repository
...Successfully got an update from the "ingress-nginx" chart repository
...Successfully got an update from the "gitlab" chart repository
...Successfully got an update from the "stable" chart repository
Update Complete. ⎈Happy Helming!⎈
Saving 1 charts
Downloading common from repo oci://myazurecr.azurecr.io/helm
~1.x: Pulling from myazurecr.azurecr.io/helm/common
Save error occurred:  could not download oci://myazurecr.azurecr.io/helm/common: myazurecr.azurecr.io/helm/common:~1.x: not found
Deleting newly downloaded charts, restoring pre-update state
Error: could not download oci://myazurecr.azurecr.io/helm/common: myazurecr.azurecr.io/helm/common:~1.x: not found
```
Is this a bug related to recent introduction of OCI references as repository?

bacongobbler:
As far as I can tell, version range support has not yet been implemented with the OCI feature. It looks like it can only lookup the literal version number.

https://github.com/helm/helm/blob/d4cc130fa90027e8c1c69ad2a06c6dee4c31d29d/pkg/getter/ocigetter.go#L43-L45

cc @jdolitsky

cndoit18:
I will try to fix it this weekend.

bacongobbler:
This is not something that can be fixed over the weekend, @cndoit18. The API lacks certain features preventing this functionality. More specifically, the API does not currently have a way to determine which tags are available when given a particular repository.

We need to plan and work through the design criteria before we implementing something.

cndoit18:
[...]

PR #9539, opened 2021-03-25
fix: nil reference when running `helm install --dependency-update`

tchupp:
There are multiple helm commands that under the hood use

solves #9535 
solves #9515

**What this PR does / why we need it**:
This PR will remove the panic that is observed when running `helm template --dependency-update` when using `oci://` dependencies. It also attempts to remove the same issue when using `helm package`, although I'm less familiar with that command.

**Special notes for your reviewer**:
- I'm somewhat familiar with golang, but not super confident I've followed the best practices.  
- There didn't seem to be any OCI specific tests at the "action" level, and I didn't add or modify any existing tests. I'm happy to add them, but I will need guidance.
- I added a "constructor" for `downloader.Manager` in order to have an explicit way to construct this struct, and to help avoid this kind of issue in the future.

**If applicable**:
- [ ] this PR contains documentation
- [ ] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

PR #9782, opened 2021-06-08
Implement changes proposed in HIP 6: OCI Support

### LINKED ISSUE #6982: Install chart from registry
Hi dream team community!

This is a little feature request: **Install chart from registry**

## Before

```
helm registry login
helm chart pull ...
helm chart export ... ./charts
helm install ... /charts/my-chart
```

## After

```
helm install gcr.io/my-project/my-chart@sha256:xxx --from-registry
```

— or —

```
helm chart install my-chart@sha256:xxx
```

## Now

I need your comments. And I want to know if you want help to create this?

/label feature

### LINKED ISSUE #8059: OCI registry cache becomes corrupted
Related to conversation in https://github.com/helm/helm/issues/7981

In some scenarios (which are yet to be determined), users get the following error:
```
> helm chart pull private-reg.comp.name/project/mychart:0.1.0
Error: blob sha256:355a6550ed5ec3bca56caac650ef0229f274ea7f4b47e29425b39d0f8cfe254e expected at /home/ant/.cache/helm/registry/cache/blobs/sha256/355a6550ed5ec3bca56caac650ef0229f274ea7f4b47e29425b39d0f8cfe254e: not found
```

Currently the only fix is to wipe the cache:
```
rm -rf /home/ant/.cache/helm/registry/cache
```

Need to figure out what scenario causes this, and perhaps Helm can check for this and auto-purge problematic blobs from the cache.

### LINKED ISSUE #9112: Issue with `helm chart save` and fully numeric tags
Hi, 

I am not sure whether it is a bug or an expected behavior, but I am facing some weird behaviors with the experimental support for OCI (`HELM_EXPERIMENTAL_OCI=1`) and `helm chart save`.

I noticed that when I perform a `helm chart save` using a numeric only tag (being a git short sha for my particular automation use-case), I end up with an error:

```shell
# simplified stdi/o from my CI pipeline
$ helm chart save "./helm/charts/my-chart" myregistry.com/mymonorepo/helm-branches/microservice-1-develop:64875383
Error: ref may only contain a single colon character (:) unless specifying a port number
```

# Current Configuration

Output of `helm version`:

```shell
$ helm version
```

Output of `kubectl version`: Not Applicable

Cloud Provider/Platform (AKS, GKE, Minikube etc.): Not Applicable

# How to reproduce

- Fetch any chart or use any local chart
- Enable experimental support for OCI
```shell
$ export HELM_EXPERIMENTAL_OCI=1
```
- Save the chart into the cache
```shell
$ helm chart save <THE_CHART> myregistry.com/the-chart:64875383
```
- Exit code `1` with the following stderr `Error: ref may only contain a single colon character (:) unless specifying a port number`

Please let me know if you need any additional information.

### LINKED ISSUE #9496: Helm 3: OCI dependencies not found?
Output of `helm version`:
```
```
Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.4", GitCommit:"e87da0bd6e03ec3fea7933c4b5263d151aafd07c", GitTreeState:"clean", BuildDate:"2021-02-18T16:12:00Z", GoVersion:"go1.15.8", Compiler:"gc", Platform:"linux/amd64"}
```
AWS. EKS for the cluster (probably not relevant) and ECR for the registry

I'm trying out the experimental OCI support and I stumbled upon something that looks like a bug to me (though I don't want to rule out that I might be using it wrong):

I have pushed a custom chart called `my-chart` to an amazon ECR registry as described [here](https://docs.aws.amazon.com/AmazonECR/latest/userguide/push-oci-artifact.html)

Then, in an "aggregation chart", I'm depending on this artifact using the following dependencies:

```
dependencies:
        - name: my-chart
          version: 0.1.0
          repository: oci://<awsaccount>.dkr.ecr.<region>.amazonaws.com
        - name: kube-prometheus-stack
          version: 14.0.1
          repository: https://prometheus-community.github.io/helm-charts
```

(The prometheus stack is there to have the comparison with a traditional repository)

When I then run `helm dependency update`, both charts are downloaded to the `charts/` directory as expected:

```
$ ls charts/
my-chart-0.1.0.tgz  kube-prometheus-stack-14.0.1.tgz
```

When I then use `helm install` passing each of these charts individually, they are installed on kubernetes as expected (so the charts themselves seem to be okay).
However, when I run `helm install . --generate-name` in the "aggregation chart", I get the following output
```
install.go:173: [debug] Original chart version: ""
install.go:190: [debug] CHART PATH: <...>

Error: found in Chart.yaml, but missing in charts/ directory: my-chart
helm.go:81: [debug] found in Chart.yaml, but missing in charts/ directory: my-chart
helm.sh/helm/v3/pkg/action.CheckDependencies
	/home/circleci/helm.sh/helm/pkg/action/install.go:612
main.runInstall
	/home/circleci/helm.sh/helm/cmd/helm/install.go:216
main.newInstallCmd.func2
	/home/circleci/helm.sh/helm/cmd/helm/install.go:120
	/go/pkg/mod/github.com/spf13/cobra@v1.1.1/command.go:850
	/go/pkg/mod/github.com/spf13/cobra@v1.1.1/command.go:958
	/go/pkg/mod/github.com/spf13/cobra@v1.1.1/command.go:895
main.main
	/home/circleci/helm.sh/helm/cmd/helm/helm.go:80
runtime.main
	/usr/local/go/src/runtime/proc.go:204
runtime.goexit
	/usr/local/go/src/runtime/asm_amd64.s:1374
```

Maybe helm is expecting a different filename convention in case of a chart from an OCI registry?
If I comment out the `my-chart` block and only keep prometheus, the same install command works as expected.

### LINKED ISSUE #9515: OCI: Segfault on helm show chart
```
$ helm version
version.BuildInfo{Version:"v3.5.3", GitCommit:"041ce5a2c17a58be0fcd5f5e16fb3e7e95fea622", GitTreeState:"dirty", GoVersion:"go1.16"}
```

```
$ helm show chart oci://europe-west1-docker.pkg.dev/foo/external-dns:0.1.0
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x21fd8a3]

goroutine 1 [running]:
helm.sh/helm/v3/internal/experimental/registry.(*Client).PullChart(0x0, 0xc000395860, 0xc000395860, 0x0, 0x0)
	/private/tmp/helm-20210310-51670-30olmy/internal/experimental/registry/client.go:156 +0x183
helm.sh/helm/v3/pkg/getter.(*OCIGetter).get(0xc000a6f360, 0xc000a76a20, 0x51, 0xc0005e3940, 0xc000a76a20, 0x51)
	/private/tmp/helm-20210310-51670-30olmy/pkg/getter/ocigetter.go:52 +0xc7
helm.sh/helm/v3/pkg/getter.(*OCIGetter).Get(0xc000a6f360, 0xc000a76a20, 0x51, 0xc000621bf0, 0x4, 0x6, 0xc000a6f360, 0x0, 0x0)
	/private/tmp/helm-20210310-51670-30olmy/pkg/getter/ocigetter.go:36 +0x77
helm.sh/helm/v3/pkg/downloader.(*ChartDownloader).DownloadTo(0xc0005e3bd8, 0x7ffeefbff750, 0x51, 0x0, 0x0, 0xc0002f8120, 0x28, 0x203000, 0x203000, 0x5, ...)
	/private/tmp/helm-20210310-51670-30olmy/pkg/downloader/chart_downloader.go:99 +0x142
helm.sh/helm/v3/pkg/action.(*ChartPathOptions).LocateChart(0xc0002e0f00, 0x7ffeefbff750, 0x51, 0xc0005cc000, 0x1, 0x24, 0xc0005c8780, 0xc0008add28)
	/private/tmp/helm-20210310-51670-30olmy/pkg/action/install.go:675 +0x412
main.runShow(0xc0006c68a0, 0x1, 0x1, 0xc0002e0f00, 0xc0000f4000, 0x27d74ce, 0x4, 0x11d0ce0)
	/private/tmp/helm-20210310-51670-30olmy/cmd/helm/show.go:179 +0xea
main.newShowCmd.func4(0xc0003c0000, 0xc0006c68a0, 0x1, 0x1, 0x0, 0x0)
	/private/tmp/helm-20210310-51670-30olmy/cmd/helm/show.go:116 +0x8a
github.com/spf13/cobra.(*Command).execute(0xc0003c0000, 0xc0006c6880, 0x1, 0x1, 0xc0003c0000, 0xc0006c6880)
	/Users/brew/Library/Caches/Homebrew/go_mod_cache/pkg/mod/github.com/spf13/cobra@v1.1.1/command.go:850 +0x472
github.com/spf13/cobra.(*Command).ExecuteC(0xc0005c02c0, 0xc000010f18, 0x1, 0xc0008adf60)
	/Users/brew/Library/Caches/Homebrew/go_mod_cache/pkg/mod/github.com/spf13/cobra@v1.1.1/command.go:958 +0x375
	/Users/brew/Library/Caches/Homebrew/go_mod_cache/pkg/mod/github.com/spf13/cobra@v1.1.1/command.go:895
main.main()
	/private/tmp/helm-20210310-51670-30olmy/cmd/helm/helm.go:80 +0x25b
```

Actually I'm not sure whether I'm invoking show chart with the correct syntax. There's almost no OCI-related docs for that on the net. :D

### LINKED ISSUE #9649: Helm stripping comments from chart
Version 3.5.4 of helm on linux.

Let me start by saying I've seen several other issues from a year ago referencing a similar issue, but this seems different, and I can't tell where the problem is coming from.

My charts are stored in the Azure Container Registry as an OCI repo.  I have 2 charts, we'll call it "Chart1" and "Chart2", and "Chart2" depends on "Chart1".  In Chart2, I don't package Chart1, I just reference it in the dependencies of the Chart.yaml.

Here is the basic sequence to get to the problem...

```
#save and push chart1
helm chart save Chart1 xxx.azurecr.io/helm/Chart1:1.0.0
helm chart push xxx.azurecr.io/helm/Chart1:1.0.0

#save and push chart2
helm chart save Chart2 xxx.azurecr.io/helm/Chart2:1.0.0
helm chart push xxx.azurecr.io/helm/Chart2:1.0.0

#on another system where I want to deploy the chart

#get chart2
helm chart pull xxx.azurecr.io/helm/Chart2:1.0.0 
helm chart export . xxx.azurecr.io/helm/Chart2:1.0.0 

#pull dependency
helm dependency update Chart2

```

Now here is what I see, in Chart2/values.yaml, it was stripped of comments and sorted alphabetically.  I expected it to be the values.yaml that I submitted that had many comments and was organized logically instead of alphabetically.

In Chart2/charts/Chart1-1.0.0.tgz, the values.yaml file is in the original format (including comments and my ordering), so that one is preserved.  I only note pulling the dependencies and comparing here because the behavior is different than the values.yaml in the base chart.

Any idea why the values.yaml got re-written?

EDIT 4/28: Fixed helm commands from "helm save" and "helm push" to "helm chart save" and "helm chart push"

### LINKED ISSUE #9810: Helm Chart.lock hash not same for same chart on diff. repos
With latest helm version - if I do helm chart save "some upstream chart".. and then push this to my OCI repo (to use for cache as my k8s cluster does not have internet access directly - for security reasons) - then the Chart.lock hash for the SAME package is different.

If you do the same with apt or yum (creating a mirror of packages) - the hashes and gpg signatures are the same - no matter which repo its in.

This means I cannot easily confirm whats in OCI repo (cache) is the same as what upstream has.. 
I would really prefer that the Chart HASH - actually was a HASH of the package you get when you do: helm dep up and it puts a .tgz in your charts/ folder.. That way, I can commit the .lock file to my git repo and trust Helm ensures that no matter which repo it pulls from (OCI or upstream) - it verifes the hash matches whats in .lock file..
Currently I'd have to recalculate all .lock files - if I want them to fetch from a different OCI repo f.ex. :(

jdolitsky:
For more information, please see the following URL: https://github.com/helm/community/blob/main/hips/hip-0006.md

Note: OCI support remains experimental, and you are still required to set HELM_EXPERIMENTAL_OCI=1 in your environment.

This PR implements the changes outlined in HIP 6. This will provide a more stable foundation for Helm's OCI registry support.

This adds a new `helm push` subcommand and `Pusher` interface for uploading charts, which can be expanded to other non-OCI protocols and various uploader plugins in the future (equivalent of `helm pull` subcommand and `Getter` interface).

This removes the experimental `helm chart` subcommand, and any notion of an OCI cache has been removed for the time being. The `helm registry` subcommand remains as is.

Throughout the rest of the code, OCI-based charts are now supported. For example, in `helm install`:
```
helm install oci://example.com/charts/nginx --version 0.1.0
```

Support for provenance files has been added, which are uploaded as a second layer on the OCI image manifest. These can be uploaded ~~using the `--with-prov` flag on `helm push`~~ if an associated `.prov` file is present in the same directory.

Chart references are strictly enforced (for now) to follow the following pattern:
```
<host>/<namespace>/<chart_name>:<chart_version>
```

Tags such as "latest" or "stable" are not yet supported.

Lastly, the `internal/experimental/registry` package has been stabilized. The public-facing `Login`, `Logout`, `Pull`, and `Push` methods all follow a common pattern, taking options and returning a result object.

Once the Go API is made public, for users wishing to upload charts using arbitrary tags or basenames, the `PushOptStrictMode(false)` option is available to enable this behavior.

All of the changes under `pkg/pusher/` have been based on `pkg/getter/`, and all of the changes under `pkg/uploader/` have been based on `pkg/downloader/`. An attempt was made to follow existing patterns wherever possible.

An initial version of this changeset included support for "uploader" plugins, but it has since been removed due to the fact that this change is already very large. The framework is in place to add support for this later.

Since these changes operate against remote registries, a new test suite in helm/acceptance-testing has been created (not yet merged). This test suite has been used to test these changes throughout the development process.

The test suite verifies that a compiled version of Helm performs as expected when operating against a live OCI registry. The test suite source can be found here: https://github.com/bloodorangeio/acceptance-testing/blob/hip-6-push/testsuites/registries.robot

Documentation has been added on Go methods, structs, etc., but not yet on helm/helm-www. Once these changes are merged, documentation will be addressed further.

**If applicable**:


internal/experimental/registry/registry_op_push.go
```diff
@@ -0,0 +1,93 @@
...
+		return nil, err
+	}
+	if operation.strictMode {
+		if !strings.HasSuffix(ref, fmt.Sprintf("/%s:%s", meta.Name, meta.Version)) {
+			return nil, errors.New(
+				"strict mode enabled, ref basename and tag must match the chart name and version")
+		}
+	}
+	store := content.NewMemoryStore()
+	chartDescriptor := store.Add("", ChartLayerMediaType, data)
```
jdolitsky, line 49: This empty string argument indicates a descriptor/layer without an OCI name annotation. We may consider adding a name based on the chart filename (e.g. `mychart-0.1.0.tgz`):

```
...
    {
      "mediaType": "application/vnd.cncf.helm.chart.content.v1.tar+gzip",
      "digest": "sha256:1b251d38cfe948dfc0a5745b7af5ca574ecb61e52aed10b19039db39af6e1617",
      "size": 2487,
      "annotations": {
         "org.opencontainers.image.title": "mychart-0.1.0.tgz"  // <----------------
      }
    }
...
```

This would allow tools such as ORAS to pull Helm charts directly to disk with the proper filename without any changes needed.

SteveLasker:
Thanks @jdolitsky 
> Support for provenance files has been added, which are uploaded as a second layer on the OCI image manifest.

👍 This is a great example of how multiple layers can be used for different artifact types.

joejulian:
A number of files under `/pkg` are being removed. I believe that this directory is considered to be part of the public API as opposed to the packages under `/internal`. This is still a matter of discussion according to [HIP-004](https://github.com/helm/community/pull/145/files) so I may be wrong. If it is, don't we need to keep them?

jdolitsky:
@joejulian good call. I had to look again because I did not intend to.

For reviewers: these are source files in `pkg/action/` tied to the removed `helm chart` subcommands. These were always experimental, but the action package is part of public API.

Should I just inform the user that these are deprecated? Something like:
```go
func (a *ChartRemove) Run(out io.Writer, ref string) error {
    fmt.Fprintln(&out, "this method is now deprecated")
    return nil
}
```

? Or just let it go?

joejulian:
@jdolitsky I'm not ignoring your question. I was hoping a maintainer would chime in. I did raise this at the last helm meeting and they noted it was worth more thought.

bacongobbler:
Good point.

It's a bit of an interesting situation... We exceedingly marked it as experimental, though it somehow went through code review and ended up in a public package.

[Our compatibility contract](https://github.com/helm/helm/blob/main/CONTRIBUTING.md#semantic-versioning) states that changes SHOULD remain backward compatible. The keyword SHOULD was carefully considered for cases like this. In this particular case (an experimental flag), I think we can make the exception to remove it entirely. Had the experiment failed and we had to remove the experiment from the Helm project, we would've had to do the same thing anyways. /shrug

I'd be comfortable with removing it entirely this time. But we should be more strict about putting experimental APIs in `internal` moving forward.

I'll let other maintainers chime in here and see what they think as well - I'm sure community members relying on these APIs may have opinions as well.

bacongobbler:
As far as [HIP 4](https://github.com/helm/community/pull/145) is concerned:

> * Experimental features are not required to ensure backward compatibility for their feature set. (They cannot, however, break backward compatibility for other parts of Helm.) Thus, a release new release of an existing experimental feature may break APIs, change its chart representations, or modify its command-line flags as long as it does not break the compatibility of non-experimental features.

Following that guideline, new releases of experimental features may break backwards compatibility.

hickeyma:
Looking at the files added to public package (`/pkg`), they are experimental and can be removed cleanly without affecting the other supported APIs. It is unfortunate that they ended up in the public package but it doesn't break backwards compatibility per se. I would have no objection removing them.

adamreese:
Running `helm dep update` gives an error but still succeeds
```
helm dep update
...Unable to get an update from the "oci://localhost:5000/myuser" chart repository:
	object required
Saving 1 charts
Downloading mychart from repo oci://localhost:5000/myuser
Pulled: localhost:5000/myuser/mychart:0.1.0
Digest: sha256:6cb1fa62c9df1f962093a528b34a0f197e68b46a2656764db62b068cd6c56890
Deleting outdated charts
```

jdolitsky:
@adamreese - was able to recreate, thanks for catching that.

pkg/action/install.go
```diff
@@ -716,5 +724,10 @@ func (c *ChartPathOptions) LocateChart(name string, settings *cli.EnvSettings) (
 	if version != "" {
 		atVersion = fmt.Sprintf(" at version %q", version)
 	}
-	return filename, errors.Errorf("failed to download %q%s (hint: running `helm repo update` may help)", name, atVersion)
+
+	hint := "running `helm repo update` may help"
```
technosophos, line 728: pkg/action should not assume a CLI

pkg/action/install.go
```diff
@@ -663,6 +663,14 @@ func (c *ChartPathOptions) LocateChart(name string, settings *cli.EnvSettings) (
 		RepositoryConfig: settings.RepositoryConfig,
 		RepositoryCache:  settings.RepositoryCache,
 	}
+
+	if strings.HasPrefix(name, "oci://") {
+		if version == "" {
+			return "", errors.Errorf("--version flag is explicitly required for OCI registries")
```
technosophos, line 669: Should not assume CLI

pkg/action/install.go
```diff
@@ -716,5 +724,10 @@ func (c *ChartPathOptions) LocateChart(name string, settings *cli.EnvSettings) (
 	if version != "" {
 		atVersion = fmt.Sprintf(" at version %q", version)
 	}
-	return filename, errors.Errorf("failed to download %q%s (hint: running `helm repo update` may help)", name, atVersion)
+
+	hint := "running `helm repo update` may help"
+	if strings.HasPrefix(name, "oci://") {
+		hint = "running `helm registry login` may help)"
```
technosophos, line 730: Same... no comments about CLI in pkg/action

bacongobbler:
I noticed there are several significant packages being introduced into the public API. Most notably being `pkg/pusher` and `pkg/uploader`.

As these are being introduced through the experimental API, can we most these to the `internal` package?

cmd/helm/push.go
```diff
@@ -0,0 +1,63 @@
...
+			if err != nil {
+				return err
+			}
+			fmt.Fprint(out, output)
+			return nil
+		},
+	}
+
+	f := cmd.Flags()
+	f.BoolVar(&client.WithProv, "with-prov", false, "upload .prov file")
```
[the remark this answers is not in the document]
jdolitsky: Do you mean, if the .prov file is present on disk next to .tgz, upload it everytime?

pkg/action/install.go
jdolitsky: Note the existing error message assuming CLI. Should we just remove the hint entirely?

jdolitsky:
@bacongobbler - sure, can do.

cmd/helm/push.go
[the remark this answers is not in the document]
bacongobbler: yes.

internal/experimental/registry/registry_op_pull.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	var meta *chart.Metadata
+	err = json.Unmarshal(configData, &meta)
+	if err != nil {
+		return nil, err
+	}
+	var chartData []byte
+	if operation.withChart {
+		var ok bool
+		_, chartData, ok = store.Get(*chartDescriptor)
```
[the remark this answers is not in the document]
jdolitsky: Since `chartData` is used later in the method (see line 150), is this irrelevant?

jdolitsky:
FYI - this file contains all the commands that this change has been tested against:
https://github.com/bloodorangeio/acceptance-testing/blob/hip-6-push/testsuites/registries.robot
cc @adamreese

pkg/pusher/ocipusher.go
```diff
@@ -0,0 +1,95 @@
...
+	meta, err := loader.Load(chartRef)
+	if err != nil {
+		return err
+	}
+
+	client := pusher.opts.registryClient
+
+	chartBytes, err := ioutil.ReadFile(chartRef)
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
jdolitsky: In the same vein as @technosophos 's comments in `pkg/action/`, should we actually avoid any mention of the CLI in error messages?

Should it be something like `chart ref must be the path to a chart archive (.tgz)` ?

internal/experimental/registry/registry_op_push.go
jdolitsky: will let this go for now. this is a future feature-add if needed

internal/experimental/registry/registry_op_pull_result.go
```diff
@@ -0,0 +1,43 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package registry // import "helm.sh/helm/v3/internal/experimental/registry"
+
+import (
```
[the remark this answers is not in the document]
jdolitsky: resolved in 817ac8ff9f67cb2d334a02b358560586d7e36c8c

internal/experimental/registry/registry_op_push_result.go
```diff
@@ -0,0 +1,42 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package registry // import "helm.sh/helm/v3/internal/experimental/registry"
+
+import (
```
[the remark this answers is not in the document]
jdolitsky: resolved in 817ac8ff9f67cb2d334a02b358560586d7e36c8c

internal/experimental/registry/registry_client.go
```diff
@@ -0,0 +1,78 @@
...
+
+	"github.com/containerd/containerd/remotes"
+	"oras.land/oras-go/pkg/auth"
+	dockerauth "oras.land/oras-go/pkg/auth/docker"
+
+	"helm.sh/helm/v3/internal/version"
+	"helm.sh/helm/v3/pkg/helmpath"
+)
+
+var (
```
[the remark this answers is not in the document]
jdolitsky: resolved in 817ac8ff9f67cb2d334a02b358560586d7e36c8c

internal/experimental/registry/registry_op_pull.go
```diff
@@ -0,0 +1,164 @@
...
+	_, manifestData, ok := store.Get(manifest)
+	if !ok {
+		return nil, errors.Errorf("Unable to retrieve blob with digest %s", manifest.Digest)
+	}
+	_, configData, ok := store.Get(*configDescriptor)
+	if !ok {
+		return nil, errors.Errorf("Unable to retrieve blob with digest %s", configDescriptor.Digest)
+	}
+	var meta *chart.Metadata
+	err = json.Unmarshal(configData, &meta)
```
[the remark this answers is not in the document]
jdolitsky: resolved in dd7bc40bc42b44de732595acc0c8e6e24729754e

internal/experimental/registry/registry_op_pull.go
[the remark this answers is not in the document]
jdolitsky: Never mind, I see what you meant. Resolved in 4073e4d1181851862ee91029a20756f8590326b2

jdolitsky:
@bacongobbler regarding https://github.com/helm/helm/pull/9782#issuecomment-862743238, I've moved `pusher` and `uploader` into `internal/experimental/` (see 24a216d3b27e1262f184d2606f74e77a0ae3c4ca).

I also created a `internal/experimental/action` package to store the experimental actions so we don't run into future issues with public API (see eba63beab1a76a13c482b574a5824372c63f956b). If you're opposed to this, I'm happy to revert. It actually removes the Login/Logout from the public API, breaking the current public API further, a concern expressed earlier in https://github.com/helm/helm/pull/9782#issuecomment-858787691. All things considered, I'm of the opinion this is probably the better way to go about it.

cmd/helm/push.go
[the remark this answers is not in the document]
jdolitsky: resolved in 35fdc688cd4ab03f93fe451b157d4021df2cd224

pkg/action/install.go
jdolitsky: resolved in 3e17c1a7efe7396b6c9cda5dbdb1a46a759e4b91

pkg/action/install.go
jdolitsky: resolved in 3e17c1a7efe7396b6c9cda5dbdb1a46a759e4b91

pkg/action/install.go
jdolitsky: resolved in 3e17c1a7efe7396b6c9cda5dbdb1a46a759e4b91

pkg/pusher/ocipusher.go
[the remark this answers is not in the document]
jdolitsky: resolved in e856df8a54cac3eba27f94a3d0fe555ff051e556

jdolitsky:
@adamreese regarding https://github.com/helm/helm/pull/9782#issuecomment-862702445, this has been resolved by e5b03c73f086a7fb874a1933ef0164961706f74f and 4ed8e770260cd9f6800b86347c1cbe8bf56eacbe.

I've re-run the robot tests with the latest change and the strange "object required" error no longer appears.

Screenshot of test report from https://github.com/bloodorangeio/acceptance-testing/runs/2862018269:

Screen Shot 2021-06-18 at 5 48 37 PM

bacongobbler:
I found a number of issues that will be fixed once this is merged. There are likely a few more.

@jdolitsky would you mind going through the issue queue and looking for issues that will be fixed once this is merged?

lindhe:
I think including the provenance file as a layer in the OCI image is an elegant solution for backwards comparability. Is it supposed to stay as the long-term solution, or do you envision moving to another system like Notary in order to not have to keep a custom signing solution just for Helm?

cmd/helm/push.go
```diff
@@ -0,0 +1,61 @@
...
+
+	"github.com/spf13/cobra"
+
+	"helm.sh/helm/v3/cmd/helm/require"
+	experimental "helm.sh/helm/v3/internal/experimental/action"
+	"helm.sh/helm/v3/pkg/action"
+)
+
+const pushDesc = `
+Upload a package to a registry.
```
lindhe, line 31: Do we call them packages or charts?

cmd/helm/push.go
```diff
@@ -0,0 +1,61 @@
...
+
+If the --with-prov flag is specified, the chart MUST have an associated
+provenance file, which will also be uploaded.
+`
+
+func newPushCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
+	client := experimental.NewPushWithOpts(experimental.WithPushConfig(cfg))
+
+	cmd := &cobra.Command{
+		Use:               "push [chart] [remote]",
```
lindhe, line 41: Since `chart` and `remote` are required arguments, I think [the convention](https://en.wikipedia.org/wiki/Usage_message#Pattern) would be to put them in angled brackets:

```
push <chart> <remote>
```

jdolitsky:
@bacongobbler - Here is a complete list of related PRs and issues that should be revisited once this is merged:

### Pull Requests:
- https://github.com/helm/helm/pull/7844
- https://github.com/helm/helm/pull/9539
- https://github.com/helm/helm/pull/9564

### Issues:
- https://github.com/helm/helm/issues/5861
- https://github.com/helm/helm/issues/6214
- https://github.com/helm/helm/issues/6324
- https://github.com/helm/helm/issues/6982
- https://github.com/helm/helm/issues/7623
- https://github.com/helm/helm/issues/8059
- https://github.com/helm/helm/issues/9112
- https://github.com/helm/helm/issues/9188
- https://github.com/helm/helm/issues/9292
- https://github.com/helm/helm/issues/9441
- https://github.com/helm/helm/issues/9496
- https://github.com/helm/helm/issues/9515
- https://github.com/helm/helm/issues/9535
- https://github.com/helm/helm/issues/9649
- https://github.com/helm/helm/issues/9694
- https://github.com/helm/helm/issues/9738
- https://github.com/helm/helm/issues/9810

cmd/helm/push.go
jdolitsky: In this case, I am following the convention in other files (such as `cmd/helm/pull.go`). I think it's better to be consistent. Maybe in a follow up PR we can modify this everywhere.

cmd/helm/push.go
jdolitsky: Thanks for catching this. Below we had bad docs on removed `--with-prov` option too. Resolved in 56028e8d0dc9ea6e879a1156f93e5b6e16f3aa5d

jdolitsky:
@lindhe - regarding https://github.com/helm/helm/pull/9782#issuecomment-866599992, perhaps support for something like Notary will come later on. If you have opinions on this, I encourage you to propose a [HIP](https://github.com/helm/community/blob/main/hips/hip-0001.md).

The good news is that these layer mediatypes are versioned (i.e. `v1.prov`), so we can provide some level of backwards compatibility if a new provenance solution is introduced.

technosophos (changes requested):
This review was with @adamreese and @bacongobbler . We were reading for stylistic and backward compatibility issues. Thanks for the last round of changes.

cmd/helm/dependency_build_test.go
```diff
@@ -45,7 +45,8 @@ func TestDependencyBuildCmd(t *testing.T) {
 
 	ociChartName := "oci-depending-chart"
 	c := createTestingMetadataForOCI(ociChartName, ociSrv.RegistryURL)
-	if err := chartutil.SaveDir(c, ociSrv.Dir); err != nil {
+	_, err = chartutil.Save(c, ociSrv.Dir)
```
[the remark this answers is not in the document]
jdolitsky: resolved in 31588e7908b5b19b413c6e039ada8e09763e97e2

internal/experimental/registry/registry_client.go
```diff
@@ -0,0 +1,76 @@
...
+
+	"github.com/containerd/containerd/remotes"
+	"oras.land/oras-go/pkg/auth"
+	dockerauth "oras.land/oras-go/pkg/auth/docker"
+
+	"helm.sh/helm/v3/internal/version"
+	"helm.sh/helm/v3/pkg/helmpath"
+)
+
+var userAgent = fmt.Sprintf("helm/%s", version.GetVersion())
```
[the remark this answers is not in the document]
jdolitsky: resolved in faf7fc5efd13ea6d1da9cb27768bfe9d837bb45c

internal/experimental/uploader/chart_uploader.go
```diff
@@ -0,0 +1,62 @@
...
+	if u.Scheme == "" {
+		return errors.New("scheme prefix missing from remote (e.g. \"oci://\")")
+	}
+
+	pusher, err := c.Pushers.ByScheme(u.Scheme)
+	if err != nil {
+		return err
+	}
+
+	err = pusher.Push(ref, u.String(), c.Options...)
```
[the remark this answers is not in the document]
jdolitsky: resolved in 4d948b4a060ba2f7cf72d56de2b05ee4cfee1890.

variable also renamed to `p`, since it was colliding with the pusher import path

internal/experimental/registry/registry_constants.go
```diff
@@ -0,0 +1,31 @@
+/*
```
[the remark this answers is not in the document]
jdolitsky: resolved in bfa5036a1a5f86129825a1aeb3bed2cd30150531

internal/experimental/registry/registry_op_logout.go
```diff
@@ -16,18 +16,23 @@ limitations under the License.
...
+	"fmt"
 )
 
-// KnownMediaTypes returns a list of layer mediaTypes that the Helm client knows about
-func KnownMediaTypes() []string {
-	return []string{
-		HelmChartConfigMediaType,
-		HelmChartContentLayerMediaType,
+// Logout logs out of a registry
+func (c *Client) Logout(host string, opts ...LogoutOption) (*LogoutResult, error) {
```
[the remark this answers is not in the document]
jdolitsky: resolved in bfa5036a1a5f86129825a1aeb3bed2cd30150531

internal/experimental/registry/registry_op_logout_options.go
```diff
@@ -14,16 +14,11 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package registry
+package registry // import "helm.sh/helm/v3/internal/experimental/registry"
```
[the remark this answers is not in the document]
jdolitsky: resolved in bfa5036a1a5f86129825a1aeb3bed2cd30150531

internal/experimental/registry/registry_op_logout_result.go
```diff
@@ -16,13 +16,9 @@ limitations under the License.
...
 
-import (
-	"github.com/containerd/containerd/remotes"
-)
-
 type (
-	// Resolver provides remotes based on a locator
-	Resolver struct {
-		remotes.Resolver
+	// LogoutResult is the result returned upon successful logout.
```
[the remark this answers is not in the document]
jdolitsky: resolved in bfa5036a1a5f86129825a1aeb3bed2cd30150531

internal/experimental/registry/registry_op_pull.go
```diff
@@ -0,0 +1,168 @@
+/*
```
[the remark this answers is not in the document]
jdolitsky: resolved in bfa5036a1a5f86129825a1aeb3bed2cd30150531

pkg/downloader/manager.go
```diff
@@ -258,6 +258,9 @@ func (m *Manager) downloadAll(deps []*chart.Dependency) error {
 		return errors.Errorf("%q is not a directory", destPath)
 	}
 
+	if err := os.RemoveAll(tmpPath); err != nil {
```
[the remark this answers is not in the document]
jdolitsky: This appeared to be necessary in order to resolve the issue described in https://github.com/helm/helm/pull/9782#issuecomment-862702445.

If you look at the line below, a new directory is being renamed to tmpPath anyway. What harm would this cause? Additionally, unit tests still appear to pass after making this change.

cmd/helm/install.go
```diff
@@ -182,6 +183,12 @@ func runInstall(args []string, client *action.Install, valueOpts *values.Options
 	}
 	client.ReleaseName = name
 
+	if strings.HasPrefix(chart, "oci://") {
```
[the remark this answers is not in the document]
jdolitsky: Resolved in 230e37ca522829b6bc8001e2cc0644d5efb6b3c5

I've added a `OCIScheme` constant to eliminate repeated hardcoded "oci" strings everywhere, and a utlity function `IsOCI(url string) bool` to replace all of the `strings.HasPrefix(url, "oci://")` calls everywhere

pkg/downloader/manager.go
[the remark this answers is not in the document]
bacongobbler: There are other PRs out there proposing similar changes. @mattfarina pointed out some of the issues with this approach here: https://github.com/helm/helm/pull/8846#pullrequestreview-557239952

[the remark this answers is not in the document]
bacongobbler: Using a temporary directory can also lead to other failure cases, so we need to come up with an alternative solution. https://github.com/helm/helm/issues/7987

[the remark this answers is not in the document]
jdolitsky: I don't want this to hold up this PR. Should I remove this for now, despite the issue described in https://github.com/helm/helm/pull/9782#issuecomment-862702445?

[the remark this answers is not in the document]
bacongobbler: #9889 should address this. Once that's merged this should be safe to remove.

internal/experimental/action/push.go
```diff
@@ -0,0 +1,72 @@
...
+		Pushers: pusher.All(p.Settings),
+		Options: []pusher.Option{},
+	}
+
+	if registry.IsOCI(remote) {
+		c.Options = append(c.Options, pusher.WithRegistryClient(p.cfg.RegistryClient))
+	}
+
+	err := c.UploadTo(chartRef, remote)
+	return out.String(), err
```
adamreese, line 71: This can be compressed into a single line.

`return out.String(), c.UploadTo(chartRef, remote)`

internal/experimental/registry/client.go
```diff
@@ -17,325 +17,476 @@ limitations under the License.
...
+}
 
-	if ref.Tag == "" {
-		return buf, errors.New("tag explicitly required")
+type (
+	// LogoutOption allows specifying various settings on logout
+	LogoutOption func(*logoutOperation)
+
+	// LogoutResult is the result returned upon successful logout.
+	LogoutResult struct {
```
adamreese, line 171: What are the LogoutResult and LoginResult types used for?

internal/experimental/registry/client.go
```diff
@@ -17,325 +17,476 @@ limitations under the License.
...
-		return buf, err
+		return nil, err
 	}
-
-	numLayers := len(layerDescriptors)
-	if numLayers < 1 {
-		return buf, errors.New(
-			fmt.Sprintf("manifest does not contain at least 1 layer (total: %d)", numLayers))
+	result := &LogoutResult{
+		Host: host,
```
adamreese, line 189: It looks like the host is just passed through the function.  The host is already known if it is a param.

internal/experimental/pusher/ocipusher.go
```diff
@@ -0,0 +1,108 @@
...
+	ref := fmt.Sprintf("%s:%s",
+		path.Join(strings.TrimPrefix(href, fmt.Sprintf("%s://", registry.OCIScheme)), meta.Metadata.Name),
+		meta.Metadata.Version)
+
+	_, err = client.Push(chartBytes, ref, pushOpts...)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
adamreese, line 87: lines 83-87 can be replaced with `return err`

internal/experimental/pusher/ocipusher.go
```diff
@@ -0,0 +1,108 @@
...
+
+	chartBytes, err := ioutil.ReadFile(chartRef)
+	if err != nil {
+		return err
+	}
+
+	var pushOpts []registry.PushOption
+	provRef := fmt.Sprintf("%s.prov", chartRef)
+	if _, err := os.Stat(provRef); err == nil {
+		provBytes, err := ioutil.ReadFile(provRef)
```
adamreese, line 71: Rather than call `os.Stat()` before `ioutil.ReadFile()` could an error type be checked from `ioutil.ReadFile()?

internal/experimental/registry/client.go
```diff
@@ -17,325 +17,476 @@ limitations under the License.
...
-func (c *Client) getChartTableRows() ([][]interface{}, error) {
-	rr, err := c.cache.ListReferences()
+	pushOperation struct {
+		provData   []byte
+		strictMode bool
+	}
+)
+
+// Push uploads a chart to a registry.
+func (c *Client) Push(data []byte, ref string, options ...PushOption) (*PushResult, error) {
```
adamreese, line 416: It looks like PushResult is only used in tests.  Should this information be displayed to the user?  If not it should probably be removed from the public API.

pkg/getter/ocigetter.go
```diff
@@ -39,22 +39,30 @@ func (g *OCIGetter) Get(href string, options ...Option) (*bytes.Buffer, error) {
...
 
-	ref := strings.TrimPrefix(href, "oci://")
+	ref := strings.TrimPrefix(href, fmt.Sprintf("%s://", registry.OCIScheme))
+
+	var pullOpts []registry.PullOption
+	requestingProv := strings.HasSuffix(ref, ".prov")
+	if requestingProv {
+		ref = strings.TrimSuffix(ref, ".prov")
+		pullOpts = append(pullOpts,
+			registry.PullOptWithChart(false),
```
adamreese, line 49: What happens if PullOptWithChart is true or omitted?

adamreese (changes requested):

internal/experimental/registry/client.go
lindhe: Maybe when using `--debug`?

internal/experimental/action/push.go
jdolitsky: resolved in ec84339d44b1766f140be7a8147fd3cd59fb4847

internal/experimental/pusher/ocipusher.go
jdolitsky: resolved in ec84339d44b1766f140be7a8147fd3cd59fb4847

internal/experimental/pusher/ocipusher.go
jdolitsky: attempted this in ec84339d44b1766f140be7a8147fd3cd59fb4847, but it turns out there is not a reliable way to determine if the error is "not exists" from `ioutil.ReadFile`. If you have a better way, please let me know. Marking resolved for now

pkg/getter/ocigetter.go
jdolitsky: If omitted, defaults to true. If true, then true. I had considered having an option such as `PullOptWithoutChart` for the prov-only case, but it seemed strange if one option was "With" and the other "Without". Happy to go that route though

internal/experimental/registry/client.go
jdolitsky: I'm considering the usage of this package outside the scope of Helm. Users may want to operate on the digests etc. when creating their own clients.

Due to the way the Pusher is implemented (based on Getter), there is no easy way to write to terminal from the `Push` method. So instead, I'm writing to `c.out`, which some people may set to /dev/null when using their own client

internal/experimental/registry/client.go
jdolitsky: Yes this is true. I simply wanted to populate the result with some information

internal/experimental/registry/client.go
jdolitsky: While they are not useful today, consider a future registry login which returns some sort of token etc. back to the user.

In general, I tried to follow the same pattern for all registry actions (`Push`, `Pull`, `Login`, `Logout`), where they each return a custom "Result" type. I'm attempting to future-proof this design with options and results.

I'm anticipating preventing the need to add methods such as `LoginWithToken` and deprecating `Login` when/if that time comes (which I've seen done a number of times in this codebase).

jdolitsky: Similarly, you'll notice  the signature of `Logout` is:
```go
func (c *Client) Logout(host string, opts ...LogoutOption) (*LogoutResult, error)
```

even though there are no such `LogoutOption`'s defined yet.

Never want to have to break this API!

pkg/getter/ocigetter.go
bacongobbler: To clarify, @adamreese and I were trying to determine what would happen if someone were to provide both `registry.PullOptWithChart(true)` and `registry.PullOptWithProv(true)` in the same request. As we were reading through the test cases, we did not see any scenario testing for that case.

We couldn't tell if the `result` on line 57 would contain both the chart and the provenance file. [The PullResult](https://github.com/helm/helm/blob/593eac8876cb16e95a2fd8bd6886dbdb3ee36a68/internal/experimental/registry/client.go#L200-L206) does appear to have both fields, and looking at the code more closely [it does appear both fields are populated](https://github.com/helm/helm/blob/593eac8876cb16e95a2fd8bd6886dbdb3ee36a68/internal/experimental/registry/client.go#L334-L359), assuming both options are set.

internal/experimental/registry/client.go
bacongobbler: Right. but why would the caller need a return type with the same host name they passed as a function parameter?

internal/experimental/registry/client.go
bacongobbler: If the idea is to return a "result" based on a generic login/logout implementation, why not return an interface instead? That way you don't have a never-ending struct with multiple fields handling different protocols like `Username/Password` and `AuthToken`. Instead you could return a `Result` type with function definitions like `Identity() string` that returns relevant information to an identity provider.

https://golang.org/doc/effective_go#interfaces_and_types

jdolitsky: Resolved in 8202ccd6998e6bf8aa34c8d9d7527e08d326397b

internal/experimental/registry/client.go
jdolitsky: Resolved in 8202ccd6998e6bf8aa34c8d9d7527e08d326397b

internal/experimental/pusher/ocipusher_test.go
```diff
@@ -14,15 +13,18 @@ See the License for the specific language governing permissions and
...
-		remotes.Resolver
+func TestNewOCIPusher(t *testing.T) {
+	testfn := func(ops *options) {
+		if ops.registryClient == nil {
+			t.Fatalf("the OCIPusher's registryClient should not be null")
+		}
 	}
-)
+
+	NewOCIPusher(testfn)
```
kensipe, line 29: what is expected here?  the `NewOCIPusher` returns an error.. do we want to assert nil?  or that the client != nil?  is this call with the expectation of a panic?   This test looks incomplete.

pkg/repo/repotest/server.go
```diff
@@ -185,31 +169,39 @@ func (srv *OCIServer) Run(t *testing.T, opts ...OCIServerOpt) {
...
 	}
 
-	err = registryClient.PushChart(ref)
+	_, err = registryClient.Push(contentBytes, ref)
 	if err != nil {
-		t.Fatal("error pushing chart")
+		t.Fatalf("error pushing dependent chart: %s", err)
 	}
 
 	if cfg.DependingChart != nil {
```
kensipe, line 189: the nested `if`s has me wondering if it is possible/reasonable to reverse the logic and return if `DependingChart` is nil.  it would likely require the assignment L208 to come before this check / return.  Not a bigger... this is readable.

pkg/repo/repotest/server.go
```diff
@@ -185,31 +169,39 @@ func (srv *OCIServer) Run(t *testing.T, opts ...OCIServerOpt) {
...
+		absPath := filepath.Join(srv.Dir,
+			fmt.Sprintf("%s-%s.tgz", c.Metadata.Name, c.Metadata.Version))
+		contentBytes, err := ioutil.ReadFile(absPath)
 		if err != nil {
-			t.Fatal("error saving depending chart")
+			t.Fatal("could not load chart into memory")
 		}
 
-		err = registryClient.PushChart(dependingRef)
+		_, err = registryClient.Push(contentBytes, dependingRef)
```
kensipe, line 202: do we want to throw away the `PushResult`?   Perhaps a debug log?

kensipe (commented):
This is really clean code!  nice work!  Left a few comments to consider.

kensipe:
I'm curious if the HIP should move past "draft"

bacongobbler:
@kensipe good question. The feature (and therefore, the HIP) is still under active development, as shown here in this PR. Until the requirements described in the [Specification](https://github.com/helm/community/blob/main/hips/hip-0006.md#specification) have been implemented and we feel it the feature is stable, then we can mark it as "final", signalling that the feature is ready for general use.

The process for marking HIPs as "final" is described in greater detail in [HIP 1](https://github.com/helm/community/blob/main/hips/hip-0001.md).

kensipe:
thanks for details... I'm used to the KEP process (which is a point of reference for me and perhaps incorrectly).  I was expecting that implementation details or acceptance criteria in the HIP would move to "provisional" or "implementable" (which I don't in HIP-1) would come before or concurrent to the feature.   Certainly the expectation of "final" would be after features landing.  I'm new to HIP process... certainly don't want to add process if not desire or needed.  Thanks for the awareness!

bacongobbler:
> I was expecting that implementation details or acceptance criteria in the HIP would move to "provisional" or "implementable" (which I don't see in HIP-1) would come before or concurrent to the feature

I think I understand where the confusion comes from. A HIP can either be marked as a "draft" or as "provisional" prior to being marked as "final", so the usage of the work "provisional" is different between KEPs and HIPs.

From https://github.com/helm/community/blob/main/hips/hip-0001.md#proposal-review--resolution

> To allow gathering of additional design and interface feedback before committing to long term stability for a feature or client library API, a HIP may also be marked as "provisional". This is short for "provisionally accepted", and indicates that the proposal has been accepted for inclusion in the reference implementation, but additional user feedback is needed before the full design can be considered "final". Unlike regular accepted proposals, provisionally accepted proposals may still be rejected even after the related changes have been included in a project's release.

The HIP process is more closely aligned to the Python PEP process, which also utilizes the "provisional" status in a similar fashion. See https://www.python.org/dev/peps/pep-0001/ for more details.

cmd/helm/registry_login.go
```diff
@@ -67,7 +68,7 @@ func newRegistryLoginCmd(cfg *action.Configuration, out io.Writer) *cobra.Comman
 	return cmd
 }
 
-// Adapted from https://github.com/deislabs/oras
+// Adapted from the ORAS project
```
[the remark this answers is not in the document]
jdolitsky: The project moved around a bit, but I re-added link directly to the source. Resolved in 241f23d72a53ee578bfaef84ae26ed86b4b5d6d2

internal/experimental/pusher/ocipusher_test.go
jdolitsky: Resolved in 241f23d72a53ee578bfaef84ae26ed86b4b5d6d2

pkg/repo/repotest/server.go
jdolitsky: Resolved in 30a8ce59861a4d9fc935d8e1dd53664208876621

pkg/repo/repotest/server.go
jdolitsky: Resolved in 30a8ce59861a4d9fc935d8e1dd53664208876621

jdolitsky:
@kensipe - I appreciate the review! I Just addressed your comments

jdolitsky:
Hoping this PR gets merged soon, but for others who may benefit, I've published a Docker image containing these changes: https://github.com/bloodorangeio/helm-hip6-docker
```
ghcr.io/bloodorangeio/helm:hip-6
ghcr.io/bloodorangeio/helm:hip-6@sha256:7dbbef4c686187af76ce6f485b4670cef1df1a115a8b5fa054a533e304d5ac83
```

kensipe:
@jdolitsky looks like recent activities resulted in merge conflicts.   I have the same interest as you in seeing this merged soonish.

bacongobbler:
>  I have the same interest as you in seeing this merged soonish.

Just to clarify with expectations. Helm 3.7.0 is scheduled for September 8th - approximately 6 weeks from now.

During the first round of reviews, @technosophos, @adamreese and I have spent a several hours reviewing this PR and only managed to scratch the surface, mostly commenting on code styling, checking backwards compatibility, and manual testing. We still haven't managed to review the actual changes like the `Pusher` interface or changes to other packages like `pkg/downloader`.

Given the amount of changes involved in this PR (currently at 3,500 lines of code changed), it is very possible we may have to extend the review out to Helm 3.8.0.

jdolitsky:
resolved merge conflicts @bacongobbler @kensipe

technosophos:
At this point, I am leaning toward merging this since it is under the experimental flag. I think the stuff that touches non-experimental code is relatively well understood. Any objections @bacongobbler or @adamreese ?

bacongobbler:
I'm okay with merging this once we resolve https://github.com/helm/helm/pull/9782#discussion_r660913292, as that does have a chance at impacting users. #9889 should be merged in first to address the underlying issue, then we can remove those lines of code. I'll address comments in that PR today.

Once that's done and the PR has been rebased, I'd be okay to merge this in.

bacongobbler:
Ping

technosophos (approved):
I am good with merging this to ship in the next feature release

internal/experimental/action/registry_login.go
```diff
@@ -18,21 +18,27 @@ package action
...
 	"io"
+
+	"helm.sh/helm/v3/internal/experimental/registry"
+	"helm.sh/helm/v3/pkg/action"
 )
 
 // RegistryLogin performs a registry login operation.
 type RegistryLogin struct {
-	cfg *Configuration
+	cfg *action.Configuration
```
mattfarina, line 28: This whole file is moved out of action and into internal/experimental. I'm just putting this here because we not call into the `action` package.

The functions like `NewRegistryLogin` have been in the action package and in use by others through the SDK. Some examples include [this](https://github.com/clcchai/codefresh-onprem-kcfi) and [this](https://github.com/fybrik/fybrik) (there are more). By moving the code from action to internal we are breaking the usage and not providing a functional path forward.

mattfarina (changes requested):
Is there a reason we are breaking the action SDK API with this change? It's going to break SDK users in the wild.

internal/experimental/action/registry_login.go
jdolitsky: Resolved in 77750962d5c803857e19032bcb5e1bec0885ba25

jdolitsky:
[unit tests](https://github.com/helm/helm/actions/runs/1182730571) and [acceptance tests](https://github.com/bloodorangeio/acceptance-testing/runs/3463420458) passing

internal/experimental/action/registry_login.go
hickeyma: FYI, this was covered in https://github.com/helm/helm/pull/9782#issuecomment-858787691, https://github.com/helm/helm/pull/9782#issuecomment-858908117 and https://github.com/helm/helm/pull/9782#issuecomment-861903870 where it was agreed that the file should never have ended up in `pkg/` and because there are experimental they should be moved there.

technosophos: Seriously, we discussed this AT LENGTH. We're trying to get as much experimental code out of `action` as we can. But whatever... just merge this thing and get it done.

mattfarina: @technosophos let's have a quick chat about this.

jdolitsky:
@mattfarina - reverted earlier changes as discussed

"All checks have passed"

technosophos (approved):
At @mattfarina 's request, I am re-LGTMing this.

mattfarina (approved):

nbraun-wolf:
Hi, this doesnt seem to work for me.

```
$ helm version
version.BuildInfo{Version:"v3.7.0-rc.1", GitCommit:"548fae169aa64cbb9efaf1fa4dd62e9adb65701b", GitTreeState:"clean", GoVersion:"go1.16.7"}
$ echo "$ACR_TOKEN" | helm registry login "$ACR" -u "$ACR_USER" --password-stdin 
Login Succeeded
$  helm install test "oci://$ACR/helm/emqx" --version 0.1.1 --dry-run --debug
Error: INSTALLATION FAILED: failed to authorize: failed to fetch anonymous token: unexpected status: 401 Unauthorized
helm.go:88: [debug] unexpected status: 401 Unauthorized
```

Before version 7 we also had the command `helm chart` available. This is now gone. So I can't inspect the charts in the repo anymore, via `helm chart list`.

I really like the idea, though, and hope it makes it into terraform soon.

lindhe:
Perhaps try to remove the cache and try again?

bacongobbler:
Use `helm package` and `helm push`. It is likely that backwards-incompatible changes have been made - this is experimental work so breaking changes are to be expected.

jdolitsky:
@nbraun-wolf - here is a gist that may help migrate from old OCI charts: https://gist.github.com/jdolitsky/66e0768ceedb6eab7395422e86e16a53

sathieu:
FYI this is breaking `helm push` plugin (https://github.com/chartmuseum/helm-push/issues/103)

zychina:
I think this is a great feature from this issue.
Is it included in helm 3.7 ? I do not see it in 3.7 official document.

TBBle:
It's part of Helm 3.7, and the [release notes](https://github.com/helm/helm/releases/tag/v3.7.0) list it under "Notable changes".

maati55:
Hi 

The helm install with oci doesn’t work for me, but the helm pull command works fine

helm version v3.7.0

```
> helm pull oci://***.azurecr.io/helm/my-chart --version 0.1.0
Pulled: ***.azurecr.io/helm/my-chart:0.1.0
Digest: sha256:14cc4c044b514c34ae6d815298adf8e1f5ee5116f2bba273f6538e5fac35a371
```

```
> helm install oci://***.azurecr.io/helm/my-chart --version 0.1.0 my-chart --dry-run
Error: INSTALLATION FAILED: failed to download "my-chart" at version "0.1.0"
```

PR #9794, opened 2021-06-11
Add hint to missing dependency message

j6s:

This PR adds a hint on how to resolve the situation to the 'missing dependency' error message.
Since switching to a shared library, this is the single biggest cause for confusion for developers in my team which could be solved by adding the  _"You may need to run `helm dependency build` to fetch missing dependencies."_ message at the end.

n/a

**If applicable**:

### DIFF
--- a/cmd/helm/install.go
+++ b/cmd/helm/install.go
@@ -214,6 +214,7 @@ func runInstall(args []string, client *action.Install, valueOpts *values.Options
 		// As of Helm 2.4.0, this is treated as a stopping condition:
 		// https://github.com/helm/helm/issues/2209
 		if err := action.CheckDependencies(chartRequested, req); err != nil {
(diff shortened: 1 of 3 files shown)


cndoit18 (commented):

j6s:
I have created #9823 for this PR. The PR also includes an update to the test-case for this error message.

bacongobbler:
With a few exceptions, we try to avoid avoid referring to Helm CLI commands in `pkg/action`. This is because `pkg/action` can be used by other programs as a library, so asking the user to run `helm repo update` may be completely irrelevant.

Can you please look into alternative ways to provide this hint to the user? One way might be to capture the error from `pkg/action` in `cmd/helm`, then add the appropriate context around the error message.

Hope this helps.

bacongobbler (changes requested):
cannot use CLI error messages in `pkg/action`

bacongobbler (approved):
thank you!

PR #9426, opened 2021-02-28
feat(helm): add ability for --dry-run to do lookup functions

tapaskapadia:
When a helm command is run with the --dry-run flag, it will try to connect to the cluster
to be able to render lookup functions.
Closes #8137

Signed-off-by: Tapas Kapadia <tapaskapadia10@gmail.com>

**What this PR does / why we need it**: It is hard to debug the lookup function and currently there is not a good way to test it with any flags. #8137 Stated that the `--dry-run` was fair game to try to implement this logic as long as the `helm template` logic stays the same.

**Special notes for your reviewer**: This is my first PR for the Helm; please let me know if I need to add or change anything. 

**If applicable**:
- [ ] this PR contains documentation
- [ ] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

PR #9797, opened 2021-06-11
Lookup on upgrade/install dry run

### LINKED ISSUE #1879: apply does not execute lookup in charts with a resource lookup
helmfile apply doesn't end up executing the helm lookup template function in charts.  my particular case:
```yaml
{{- if .Values.monitoring.enabled -}}
{{- $prometheus := (lookup "monitoring.coreos.com/v1" "Prometheus" "" "").items | mustFirst }}
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  name: {{ template "webapplication2.fullname" . }}
  namespace: {{ $prometheus.metadata.namespace }}
  labels:
    app: {{ template "webapplication2.name" . }}
    chart: {{ template "webapplication2.chart" . }}
    release: {{ .Release.Name }}
    heritage: {{ .Release.Service }}
spec:
  endpoints:
  {{- range .Values.monitoring.sources}}
  - interval: {{ .interval | default "10s" }}
    port: {{ .port }}
  {{- end }}
  namespaceSelector:
    matchNames:
    - {{ .Release.Namespace }}
  selector:
    matchLabels:
      app: {{ template "webapplication2.name" . }}
      release: {{ .Release.Name }}
{{- end }}
```
my helmfile uses `helmfiles:` to ensure the kube-prometheus-stack is installed before installing the charts for webapplication2 (the name is fantastic) which include this service monitor.  My workaround, for now, is that I know what namespace I am deploying prometheus to so I am not _required_ to look it up or jump through any other hoops to share the namespace (or I could change the prometheus configuration to watch for service monitors in any namespace).

I _think_ the issue is actually in helm; I opened this PR: https://github.com/helm/helm/pull/9797

apply uses the helm-diff plugin, helm-diff runs `helm upgrade --dry-run` which should (I think) be expected to make cluster contact... there was talk of a security issue with letting cluster contact be made in ways that might produce output in the related issue but it sounded like maintainers were willing to accept PRs for install and upgrade --dry-run to make cluster contact. The linked issue is for debugging the output of lookup with --dry-run, the PR only tries to change upgrade since template and install were pretty closely coupled.

### LINKED ISSUE #8137: Unable to debug "lookup" function, as its disabled with `helm template`
Output of `helm version`:
```
```

Output of `kubectl version`:
```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"13+", GitVersion:"v1.13.4+7bd2e5b", GitCommit:"7bd2e5b", GitTreeState:"clean", BuildDate:"2019-05-19T23:52:43Z", GoVersion:"go1.11.5", Compiler:"gc", Platform:"linux/amd64"}
```

all

I am trying to use the "lookup" function, which is resulting in problems with my yaml formatting:

```
$ helm install --namespace myapp api ./charts/deploy/
Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: ValidationError(Secret.metadata): unknown field "type" in io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta
```

The template file from which I believe the error is coming from looks like this:

```
---
apiVersion: v1
data:
  auth: {{ (lookup "v1" "Secret" "app2" "api-k8s-htpasswd").data.auth }}
kind: Secret
metadata:
  labels:
    k8s-app: myap-api
  name: htpasswd-secret
type: Opaque
```

From what I understand, the lookup function was disabled when you run `helm template`, as trying to process this chart results in a nil pointer error:
```
Error: template: deploy/templates/htpasswd-secret.yaml:4:56: executing "deploy/templates/htpasswd-secret.yaml" at <"api-k8s-htpasswd">: nil pointer evaluating interface {}.auth
```

I would like to have a discussion as to why I can't use `helm template` to debug my chart, as that, in my opinion, should be the primary purpose of having a `template` command.

hypnopotamus:
`upgrade` and `install` are expected to be able to contact a cluster, even on a dry run.  `template` should not render with current cluster state (no cluster contact) even if `--validate` is specified to prevent accidentally sharing cluster content

~~refs #8137 - does not deal with the entire thing, only upgrade --dry-run; turned out template and install were pretty closely tangled, probably better to deal with them in their own PR~~
fixes #8137 - install and upgrade will render connected to a cluster (dry run or otherwise) , template will not (with validate or otherwise).  Based on what I see in the discussion on this issue that's the expected behavior.
fixes roboll/helmfile#1879

My specific case that led to this:
- [helmfile](https://github.com/roboll/helmfile) apply uses the helm-diff plugin. I expected it to make cluster contact to determine what parts of the grouped charts were new and should be applied
- [helm-diff](https://github.com/databus23/helm-diff) uses `helm upgrade --dry-run` to produce an output of what would change if a chart were installed or upgraded
- `helm upgrade --dry-run` was not making cluster contact => the lookup function returned nil, as expected for operations not expected to be allowed to make cluster contact

**If applicable**:

### DIFF
--- a/cmd/helm/install.go
+++ b/cmd/helm/install.go
@@ -117,6 +117,7 @@ func newInstallCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 			return compInstall(args, toComplete, client)
 		},
 		RunE: func(_ *cobra.Command, args []string) error {
+			client.RenderConnected = true
 			rel, err := runInstall(args, client, valueOpts, out)
 			if err != nil {
 				return err
--- a/cmd/helm/template.go
+++ b/cmd/helm/template.go
@@ -74,6 +74,7 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 			}
 
 			client.DryRun = true
+			client.RenderConnected = false
 			client.ReleaseName = "RELEASE-NAME"
 			client.Replace = true // Skip the name check
 			client.ClientOnly = !validate
--- a/cmd/helm/upgrade.go
+++ b/cmd/helm/upgrade.go
@@ -100,6 +100,7 @@ func newUpgradeCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 					instClient.CreateNamespace = createNamespace
 					instClient.ChartPathOptions = client.ChartPathOptions
 					instClient.DryRun = client.DryRun
+					instClient.RenderConnected = true
 					instClient.DisableHooks = client.DisableHooks
 					instClient.SkipCRDs = client.SkipCRDs
 					instClient.Timeout = client.Timeout
--- a/go.mod
+++ b/go.mod
@@ -36,6 +36,7 @@ require (
 	github.com/xeipuuv/gojsonschema v1.2.0
 	golang.org/x/crypto v0.0.0-20210220033148-5ea612d1eb83
 	golang.org/x/term v0.0.0-20210220032956-6a3ed077a48d
+	golang.org/x/tools v0.1.3 // indirect
 	k8s.io/api v0.21.0
 	k8s.io/apiextensions-apiserver v0.21.0
 	k8s.io/apimachinery v0.21.0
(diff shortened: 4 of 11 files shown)


stavros-k:
I think this can be closed? https://github.com/helm/helm/pull/9426

PR #9818, opened 2021-06-17
import all values from a dependency chart

### LINKED ISSUE #9817: Allow import-values to use all values from a sub chart
Currently it is not allowed to define import values with a dot `.` in order to use all values from a sub chart:
```
# parent's Chart.yaml file
dependencies:
  - name: subchart1
    version: 0.1.0
    ...
    import-values:
      - child: .
        parent: myimports
```

This is especially useful for using values, which are at the root level of a sub chart:
```
# sub chart''s values.yaml file
myValue1: true
myValue2: "foo"
myValueN:
  foo: "1"
  bar: "false"
```
Since only table values can be used it is currently just possible to import the values `myValueN` as described in the [docs](https://helm.sh/docs/topics/charts/#importing-child-values-via-dependencies). When trying to import `myValue1` or `myValue2` a warning is shown: `dependencies.go:247: Warning: ImportValues missing table from chart subchart: "" is not a table`.

This issue relates to #2746.

aemaem:
Signed-off-by: Michael Mair <michael.mair@leftshift.one>

Allows for importing root level values from sub charts.
closes #9817

**Special notes for your reviewer**: -

**If applicable**:


digihunch:
Is this feature on-schedule to be merged?

aemaem:
Apparently there is no need for this functionality...

hacktron95:
Hi @aemaem, please consider reopening this, I think it helps in a lot of the workflows

aabouzaid:
This feature is really important and the change looks small compared to the value out of it!

aemaem:
I reopended and updated the PR. We'll see when it will be merged.

aabouzaid:
@aemaem thank you :raised_hands: 

This PR actually fixes other issues.

Beside importing the whole vars from the deps also fixes the issue where it's not possible to import non-map values (i.e. it's not possible to import top-level string keys).

hwo411:
Would be great to have this merged!

mysiki:
+1

jacek-jablonski:
Could we have this super useful funcitonality?

Makeshift:
This would be an incredibly useful addition to `import-values`. Currently, the only way to achieve something similar is to nest all values in a subchart under a key, then `import-values` the key into the parent.

For example:

```yaml
# charts/subchart/values.yaml
all:
  foo: bar
```

```yaml
# Chart.yaml
dependencies:
  - name: foo
    version: 0.0.1
    import-values:
      - child: all
        parent: .
```

Thankfully, merging into `.` in the parent works, but this `values.yaml` layout means that a lot of other tooling (such as helm-docs, schema generation, etc) doesn't work properly.

aemaem:
Would be happy to update this PR accordingly. Maybe someone can tell me what I can do, that the Helm team would consider a review on this?

banjoh:
Works like a charm!

banjoh (approved):

aemaem:
@banjoh is there anything I can do to get this merged?

TerryHowe (approved):
/lgtm 

I don't know if this behavior is  desired behavior though.

banjoh:
> /lgtm
> 
> I don't know if this behavior is desired behavior though.

I've raised this for our next dev meeting https://docs.google.com/document/d/1d-6xJEx0C78csIYSPKJzRPeWaHG_8W1Hjl72OJggwdc/edit?tab=t.0

aemaem:
Just for information if anyone is waiting for this. These are the comments from the developer meeting where @banjoh brought it up:

> Scott: Concern of how to handle special/reserved key that might be pulled in and break things
Webb: Collisions concern. Can the order of import be controlled?
George: Concern that this couples parent and subchart value format in a way that may not be ideal
Existing docs: https://helm.sh/docs/topics/charts/#importing-child-values-via-dependencies
Scott: for charts v3, and rethinking import/export values, we should also look at https://github.com/helm/community/blob/main/hips/hip-0016.md. Rethinking for v3 because it overcomplicates our values coalescing logic, and don't we already have these via the Subcharts built-in? (see https://helm.sh/docs/chart_template_guide/builtin_objects/)

Find the meeting notes here: https://docs.google.com/document/d/1d-6xJEx0C78csIYSPKJzRPeWaHG_8W1Hjl72OJggwdc/edit?pli=1&tab=t.0#heading=h.5spcjj44t85m

banjoh:
Given the concerns raised https://github.com/helm/helm/pull/9818#issuecomment-4457559346, I think we should punt on this one for now in favour of simplifying the code. Such changes would target Chart v3.

We've seen a few regressions to values handling (e.g https://github.com/helm/helm/issues/31919, https://github.com/helm/helm/issues/12756, https://github.com/helm/helm/issues/31446) caused by making changes in this are of code.

PR #9828, opened 2021-06-18
use TLS client information from repo config when downloading a chart

### LINKED ISSUE #9826: Helm dep build/update doesn't inherit --ca-file --cert-file --key-file from helm repo add
After adding a repo with `helm repo add --ca-file pathToCA.pem https://example.com/helm`, helm dependency update on an umbrella chart will fail with `x509: certificate signed by unknown authority`

It looks like the same problem as for https://github.com/helm/helm/issues/8868. Fixing this one could also fix https://github.com/helm/helm/issues/9568

Output of `helm version`: 3.6.0

Output of `kubectl version`: 1.20.7

Cloud Provider/Platform (AKS, GKE, Minikube etc.): minikube

cvila84:
Signed-off-by: Christophe VILA <christophe.vila@thalesgroup.com>

**What this PR does / why we need it**: closes #9826

**Special notes for your reviewer**: there seems to be some TLS client config logic already in place in https://github.com/helm/helm/blob/179f90151d5ecb4aa3d35ada35e82b5c1e791752/pkg/downloader/chart_downloader.go#L156 but I think it is useless (either the downloader options configuration is already set by the caller, either this logic is not reached because of a possible bug in https://github.com/helm/helm/blob/179f90151d5ecb4aa3d35ada35e82b5c1e791752/pkg/downloader/chart_downloader.go#L344 around URL comparison which fails because one is relative and the other is absolute)

**If applicable**:

### DIFF
--- a/pkg/downloader/manager.go
+++ b/pkg/downloader/manager.go
@@ -310,7 +310,7 @@ func (m *Manager) downloadAll(deps []*chart.Dependency) error {
 
 		// Any failure to resolve/download a chart should fail:
 		// https://github.com/helm/helm/issues/1439
-		churl, username, password, insecureskiptlsverify, passcredentialsall, err := m.findChartURL(dep.Name, dep.Version, dep.Repository, repos)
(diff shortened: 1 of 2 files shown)


adamreese (approved):
Thanks, This was a good catch!

Issue #9820, opened 2021-06-17
Helm 3.6.1 breaks "helm pull" with repo/chartname


derkoe:
Helm 3.6.1 breaks `helm pull` with repo/chartname (after adding the repo with auth).

How to reproduce:

```
$ helm repo add myrepo https://example.com/myrepo --username *** --password ***
"myrepo" has been added to your repositories
$ helm repo update
Hang tight while we grab the latest from your chart repositories...
...Successfully got an update from the "myrepo" chart repository
$ helm pull myrepo/mychart
Error: failed to fetch https://example.com/myrepo/mychart-1.0.0.tgz : 401 Unauthorized
```

Output of `helm version`:  
```
version.BuildInfo{Version:"v3.6.1", GitCommit:"61d8e8c4a6f95540c15c6a65f36a6dd0a45e7a2f", GitTreeState:"clean", GoVersion:"go1.16.5"}
```

derkoe:
I've debugged the issue into `httpgetter.go` - when running `helm pull myrepo/mychart` the variable `u1` contains ` myrepo/mychart` - that's why this if does not work https://github.com/helm/helm/blob/release-3.6.1/pkg/getter/httpgetter.go#L74

dol:
The same happens with any other command that's using httpgetter. Like `helm show values some/chart` . As @derkoe pointed out correctly, the `u1` doesn't contains a full url to compare with. Only the `some/chart`.  I did a similar debugging as Christian and came to the same conclusion.
Thank you @derkoe for raising the issue and root cause so fast.

dol:
Not sure if this helps but it looks like that the url, once discovered from the repositories is not set as url to the options.

I made a small adjustment that resolves the issue: https://github.com/helm/helm/compare/v3.6.1...dol:patch/v3.6.1?expand=1

derkoe:
@mattfarina please have a look at this issue (usually i wouldn't ping but this is a major issue).

derkoe:
@dol tried your fix and it works - could your create a PR?

iss0:
Could this be related to https://github.com/helm/helm/issues/9816#issuecomment-863868716? :thinking:

iss0:
@dol I tried your fix for my problem mentioned in https://github.com/helm/helm/issues/9816#issuecomment-863868716 and it works for me as well!

bacongobbler:
Thanks @dol. If you want to submit a patch along with a test case to help confirm the issue, that would be appreciated.

[...]

PR #9844, opened 2021-06-22
Fix regression of v3.6.1 with remote repository and authentiation

dol:
As discusses in #9820. Possible fix to the regression.


dol:
Closed in favor of #9871.

PR #5182, opened 2019-01-17
feat(helm): adding --name to update single repo

thesattiraju:
**What this PR does / why we need it**:
fixes https://github.com/helm/helm/issues/5127

**Special notes for your reviewer**:

**If applicable**:
- [x] this PR contains documentation
- [x] this PR contains unit tests
- [x] this PR has been tested for backwards compatibility

PR #9845, opened 2021-06-22
feat(cmd): Allow to specify which repos to update

### LINKED ISSUE #8071: 'helm repo update {REPO} ' feature
Some `helm repo` commands accept the repository name as parameter, that is not the case of `helm repo update`, this command doesn't receive any extra parameter, updating all the installed repos.

There are some cases where it would be useful to just update one specific repository, for example, if I am working with one specific repository, I don't need to update all of them, just the one I want to obtain the latest changes.

With the "new" approach of distributed repositories, it will be more habitual to have a larger list of repositories. Do you think something like this feature is useful? Is it a good option for a first-contribution PR?

```console
$ helm repo list
NAME             	URL
bitnami          	https://charts.bitnami.com/bitnami
stable           	https://kubernetes-charts.storage.googleapis.com
azure-marketplace	https://marketplace.azurecr.io/helm/v1/repo

$ helm repo update bitnami
Hang tight while we grab the latest from your chart repository...
```
instead of
```console
$ helm repo list
NAME             	URL
bitnami          	https://charts.bitnami.com/bitnami
stable           	https://kubernetes-charts.storage.googleapis.com
azure-marketplace	https://marketplace.azurecr.io/helm/v1/repo

$ helm repo update
...Successfully got an update from the "azure-marketplace" chart repository
```

marckhouzam:
Fixed #8071 

As discussed in the Helm dev call, this is a port to helm v3 of #5182, which allows to specify which repositories to update.
 
 A little more flexible than the v2 version, this PR allows to specify a list of repositories that should be updated instead of a single one.  This is in line with v3's support for `helm repo remove` which also allows a list of repos to be specified.

**If applicable**:

### DIFF
--- a/cmd/helm/testdata/repositories.yaml
+++ b/cmd/helm/testdata/repositories.yaml
@@ -2,3 +2,8 @@ apiVersion: v1
 repositories:
   - name: charts
     url: "https://charts.helm.sh/stable"
+  - name: firstexample
+    url: "http://firstexample.com"
+  - name: secondexample
+    url: "http://secondexample.com"
+
--- a/cmd/helm/repo_update.go
+++ b/cmd/helm/repo_update.go
@@ -32,6 +32,10 @@ import (
 const updateDesc = `
 Update gets the latest information about charts from the respective chart repositories.
 Information is cached locally, where it is used by commands like 'helm search'.
+
+You can optionally specify a list of repositories you want to update.
+	$ helm repo update <repo_name> ...
+To update all the repositories, use 'helm repo update'.
 `
 
 var errNoRepositories = errors.New("no repositories found. You must add one before updating")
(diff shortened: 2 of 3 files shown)


cmd/helm/testdata/repositories.yaml
```diff
@@ -2,3 +2,8 @@ apiVersion: v1
 repositories:
   - name: charts
     url: "https://charts.helm.sh/stable"
+  - name: example
+    url: "http://example.com"
+  - name: other
+    url: "http://example.com"
```
bridgetkromhout, line 8: To prevent confusion, would it make sense to name this `otherexample` and `otherexample.com`?

marckhouzam: Good point.  I've pushed an update which uses `firstexample` and `secondexample`.
I could not use `example` and `otherexample` because the test checks if the name of the repo is part of the output, so we can't have one name (`example`) as part of another name (`otherexample`) or else the test could get confused.

dol:
I watched your weekly dev call and found out about this missing functionality from `v2`. This PR also resolved #8071. I did some tinkering (https://github.com/dol/helm/commit/34eb045c44d1727809b8761e1c0e2829b4f27f40) to understand more about the inner workings of helm. I did not spend much time on the issue and the commit lacks any test cases. I'll will delete the experimental branch in favor of this PR.

bacongobbler (approved):

marckhouzam:
In https://github.com/helm/helm/issues/8071#issuecomment-627649796, @technosophos  mentioned "the case where a chart in Repo A references a chart in Repo B, and Repo A gets locally updated but Repo B does not".

Is it ok to leave it up to the user to handle such a case?  I personally feel it would be ok, especially since helm v2 allowed to only update a single repo.

technosophos:
We discussed this in the dev call, and we do not think there are any significant risks in going this route. So I am good with this PR.

sbose78:
Agreed, makes sense to support specification of helm repos - helps from a resource perspective 👍

PR #9847, opened 2021-06-23
Add a destination option to `helm create`

quantumsheep:
Related to #10048.

This PR adds a new option to `helm create`: `--directory` (shorthand `-d`). Here's an example:
```bash
$ helm create coolapp -d coolapp-helm
Creating coolapp

$ ls ./coolapp-helm
Chart.yaml  charts      templates   values.yaml
```

**If applicable**:

### DIFF
--- a/cmd/helm/create.go
+++ b/cmd/helm/create.go
@@ -52,6 +52,7 @@ will be overwritten, but other files will be left alone.
 
 type createOptions struct {
 	starter    string // --starter
+	directory  string // --directory
 	name       string
 	starterDir string
 }
@@ -81,6 +82,7 @@ func newCreateCmd(out io.Writer) *cobra.Command {
 	}
 
 	cmd.Flags().StringVarP(&o.starter, "starter", "p", "", "the name or absolute path to Helm starter scaffold")
+	cmd.Flags().StringVarP(&o.directory, "directory", "d", "", "the location of the created chart")
 	return cmd
 }
 
@@ -97,6 +99,11 @@ func (o *createOptions) run(out io.Writer) error {
 		APIVersion:  chart.APIVersionV2,
 	}
 
+	destination := o.directory
+	if destination == "" {
+		destination = o.name
+	}
+
 	if o.starter != "" {
 		// Create from the starter
 		lstarter := filepath.Join(o.starterDir, o.starter)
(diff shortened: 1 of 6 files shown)


technosophos:
I hate to be a stickler for things like this, but a feature PR must be linked to an issue that describes the feature.

quantumsheep:
@technosophos no worries :) Here you go: #10048.

cmd/helm/create.go
```diff
@@ -97,17 +99,22 @@ func (o *createOptions) run(out io.Writer) error {
 		APIVersion:  chart.APIVersionV2,
 	}
 
+	destination := o.directory
+	if destination == "" {
+		destination = o.name
```
bacongobbler, line 104: Don't we want the destination to be `filepath.Dir(o.name)`? That would match the old behaviour.

pkg/chartutil/create.go
```diff
@@ -570,20 +570,16 @@ func Create(name, dir string) (string, error) {
...
-	path, err := filepath.Abs(dir)
+	cdir, err := filepath.Abs(dir)
 	if err != nil {
-		return path, err
+		return cdir, err
 	}
 
-	if fi, err := os.Stat(path); err != nil {
-		return path, err
+	parentPath := filepath.Dir(cdir)
```
bacongobbler, line 578: By doing this, you change the current behaviour of `Create()`, which is a [publicly available API endpoint](https://pkg.go.dev/helm.sh/helm/v3@v3.6.3/pkg/chartutil#Create). Changing its internal behaviour will silently break users consuming the public API.

Notice how you had to change every invocation of `chartutil.Create` in the tests in order to match its new calling signature.

In other words, if I previously expected

```
chartutil.Create("/foo")
```

To create a new chart in directory `/foo`, then the same should hold true until Helm 4.

bacongobbler (changes requested):
Looks good! just need to fix a breaking change in the API's behaviour before we can merge.

bacongobbler:
I had to double check with `helm lint` and `helm package` to see if this convention was valid, but after the removal of the "chart name must match the directory name" constraint in Helm 3 this appears to work.

```
bacongobbler@ultralisk /tmp ><> helm create coolapp
Creating coolapp
bacongobbler@ultralisk /tmp ><> mv coolapp coolapp-helm
bacongobbler@ultralisk /tmp ><> helm lint ./coolapp-helm
==> Linting ./coolapp-helm

bacongobbler@ultralisk /tmp ><> helm package ./coolapp-helm
Successfully packaged chart and saved it to: /tmp/coolapp-0.1.0.tgz
```

bacongobbler:
Note that `helm create` accepts a path as an argument.

From the help text:

> 'helm create' takes a path for an argument. If directories in the given path
do not exist, Helm will attempt to create them as it goes. If the given
destination exists and there are files in that directory, conflicting files
will be overwritten, but other files will be left alone.

Please make sure you accommodate this in your tests. Thanks!

PR #9854, opened 2021-06-24
Fix env HELM_PLUGINS behavior

### LINKED ISSUE #9814: Confusing behavior between HELM_PLUGIN and HELM_DATA_HOME
Output of `helm version`: 3.5.2

A recent project required me to run Helm in a container with a persistent volume. My plan was to put `HELM_*_HOME` on the persistent volume and put `HELM_PLUGIN` to a file system in the image. I thought this would allow me create a Docker image with all the plugins preinstalled. Then, anything related to "transient" data like the added repos and their charts would be kept in `HELM_*_HOME` thus on the persistent volume. 

This did not work. `helm plugin install` writes plugins to `HELM_DATA_HOME/plugins`. This is unexpected because:

> $HELM_PLUGINS | set the path to the plugins directory

I assumed that `helm install` would write files to `HELM_PLUGINS`. That is not the case.

I dug around one of the plugins I'm using and found a [reference](https://github.com/jkroepke/helm-secrets/blob/main/plugin.yaml#L7) to `HELM_PLUGIN_DIR`. That led me to the [plugin docs](https://helm.sh/docs/topics/plugins/) which state:

> Plugins are installed using the $ helm plugin install <path|url> command. You can pass in a path to a plugin on your local file system or a url of a remote VCS repo. The helm plugin install command clones or copies the plugin at the path/url given into $HELM_PLUGINS

So my question is: what is the expected behavior here? If `HELM_PLUGINS` is set, then should all plugin installations go into that directory? If so, then why would files be written to `HELM_DATA_HOME`. 

I was able to reproduce this problem using a simple `Dockerfile`:

```Dockerfile
FROM alpine/helm:3.5.2

# uncomment this to fix the build
ENV HELM_PLUGINS=/var/lib/helm/plugins

RUN apk add git ncurses bash curl

RUN helm plugin install https://github.com/jkroepke/helm-secrets

# Fails with command not found
RUN helm secrets --version
```

The `docker build` fails with:

```
 => ERROR [4/4] RUN helm secrets --version                                                                                                                                                                      0.3s
------
 > [4/4] RUN helm secrets --version:
#7 0.257 Error: unknown command "secrets" for "helm"
#7 0.257 Run 'helm --help' for usage.
```

yxxhero:

fix: https://github.com/helm/helm/issues/9814 https://github.com/helm/helm/issues/9942

**If applicable**:

### DIFF
--- a/pkg/plugin/installer/base.go
+++ b/pkg/plugin/installer/base.go
@@ -16,6 +16,7 @@ limitations under the License.
 package installer // import "helm.sh/helm/v3/pkg/plugin/installer"
 
 import (
+	"os"
 	"path/filepath"
 
 	"helm.sh/helm/v3/pkg/helmpath"
@@ -34,6 +35,10 @@ func newBase(source string) base {
 func (b *base) Path() string {
 	if b.Source == "" {
 		return ""
+	}
+	helmPluginsDir := os.Getenv("HELM_PLUGINS")
+	if helmPluginsDir != "" {
+		return filepath.Join(helmPluginsDir, filepath.Base(b.Source))
 	}
 	return helmpath.DataPath("plugins", filepath.Base(b.Source))
 }
--- a/pkg/plugin/installer/base.go
+++ b/pkg/plugin/installer/base.go
@@ -16,19 +16,21 @@ limitations under the License.
 package installer // import "helm.sh/helm/v3/pkg/plugin/installer"
 
 import (
-	"os"
 	"path/filepath"
-
-	"helm.sh/helm/v3/pkg/helmpath"
 )
 
 type base struct {
 	// Source is the reference to a plugin
 	Source string
+	// PluginsDirectory is the directory where plugins are installed
+	PluginsDirectory string
 }
 
-func newBase(source string) base {
-	return base{source}
+func newBase(source, pluginsDirectory string) base {
+	return base{
+		Source:           source,
+		PluginsDirectory: pluginsDirectory,
+	}
 }
 
 // Path is where the plugin will be installed.
@@ -36,9 +38,5 @@ func (b *base) Path() string {
 	if b.Source == "" {
 		return ""
 	}
-	helmPluginsDir := os.Getenv("HELM_PLUGINS")
-	if helmPluginsDir != "" {
-		return filepath.Join(helmPluginsDir, filepath.Base(b.Source))
-	}
-	return helmpath.DataPath("plugins", filepath.Base(b.Source))
+	return filepath.Join(b.PluginsDirectory, filepath.Base(b.Source))
 }
--- a/pkg/plugin/installer/installer.go
+++ b/pkg/plugin/installer/installer.go
@@ -64,19 +64,19 @@ func Update(i Installer) error {
 }
 
 // NewForSource determines the correct Installer for the given source.
-func NewForSource(source, version string) (Installer, error) {
+func NewForSource(source, version, pluginsDirectory string) (Installer, error) {
 	// Check if source is a local directory
 	if isLocalReference(source) {
-		return NewLocalInstaller(source)
+		return NewLocalInstaller(source, pluginsDirectory)
 	} else if isRemoteHTTPArchive(source) {
-		return NewHTTPInstaller(source)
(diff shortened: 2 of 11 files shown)


cndoit18 (approved):
LGTM

pkg/plugin/installer/base.go
```diff
@@ -35,5 +36,9 @@ func (b *base) Path() string {
 	if b.Source == "" {
 		return ""
 	}
+	helmPluginsDir := os.Getenv("HELM_PLUGINS")
```
mattfarina, line 39: Two things....

1. The plugins directory should be passed in rather than directly pulled from the environment variable. The could be a public property on `base` and set as part of the install command.
2. There are other places this needs to be modified beyond base. I didn't look entirely but I did see http installer has a `Path` function that overrides this one. It needs to be modified, too.

mattfarina (changes requested):
Thanks for trying to fix this. I did find a couple things that need to be changed on a quick pass.

pkg/plugin/installer/base.go
yxxhero: I will fix it.

pkg/plugin/installer/installer.go
```diff
@@ -64,19 +64,19 @@ func Update(i Installer) error {
 }
 
 // NewForSource determines the correct Installer for the given source.
-func NewForSource(source, version string) (Installer, error) {
+func NewForSource(source, version, pluginsDirectory string) (Installer, error) {
```
bacongobbler, line 67: So unfortunately, this would constitute a breaking change as `installer.NewForSource` is a publicly exposed function call. Is there another way we can fix this?

For reference: https://pkg.go.dev/helm.sh/helm/v3@v3.6.3/pkg/plugin/installer#NewForSource

pkg/plugin/installer/installer.go
```diff
@@ -64,19 +64,19 @@ func Update(i Installer) error {
...
+		return NewHTTPInstaller(source, pluginsDirectory)
 	}
-	return NewVCSInstaller(source, version)
+	return NewVCSInstaller(source, version, pluginsDirectory)
 }
 
 // FindSource determines the correct Installer for the given source.
-func FindSource(location string) (Installer, error) {
-	installer, err := existingVCSRepo(location)
+func FindSource(location, pluginsDirectory string) (Installer, error) {
```
bacongobbler, line 78: same here: https://pkg.go.dev/helm.sh/helm/v3@v3.6.3/pkg/plugin/installer#FindSource

pkg/plugin/installer/installer.go
```diff
@@ -64,19 +64,19 @@ func Update(i Installer) error {
...
 	// Check if source is a local directory
 	if isLocalReference(source) {
-		return NewLocalInstaller(source)
+		return NewLocalInstaller(source, pluginsDirectory)
 	} else if isRemoteHTTPArchive(source) {
-		return NewHTTPInstaller(source)
+		return NewHTTPInstaller(source, pluginsDirectory)
 	}
-	return NewVCSInstaller(source, version)
+	return NewVCSInstaller(source, version, pluginsDirectory)
```
bacongobbler, line 74: Same here: https://pkg.go.dev/helm.sh/helm/v3@v3.6.3/pkg/plugin/installer#NewVCSInstaller

pkg/plugin/installer/local_installer.go
```diff
@@ -28,13 +28,13 @@ type LocalInstaller struct {
 }
 
 // NewLocalInstaller creates a new LocalInstaller.
-func NewLocalInstaller(source string) (*LocalInstaller, error) {
+func NewLocalInstaller(source, pluginsDirectory string) (*LocalInstaller, error) {
```
bacongobbler, line 31: https://pkg.go.dev/helm.sh/helm/v3@v3.6.3/pkg/plugin/installer#NewLocalInstaller

bacongobbler (changes requested):

pkg/plugin/installer/installer.go
yxxhero: Wait for me to think. Do you have any suggestions?

PR #9870, opened 2021-06-25
Set Helm as manager for managedFields

mattfarina:
managedFields were a changed that landed in 1.18. This is an array
under metadata with managedFields. The kubernetes client pkgs that
Helm uses automatically add them.

This change added a manager for the managedFields. The flow for
deciding on the name to use is:

1. An explicit name if one is chosen
2. The base name of the first os.Arg (the binary name) if no name
   explicitly set.
3. unknown if no name set and name cannot be detected

The name is at the package level as there is no other place to easily
set it for Helm v3. Since the name is for the binary or app it should
be ok to set app wide.

Signed-off-by: Matt Farina <matt.farina@suse.com>

**If applicable**:

### DIFF
--- a/cmd/helm/helm.go
+++ b/cmd/helm/helm.go
@@ -32,6 +32,7 @@ import (
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/cli"
 	"helm.sh/helm/v3/pkg/gates"
+	"helm.sh/helm/v3/pkg/kube"
 	kubefake "helm.sh/helm/v3/pkg/kube/fake"
 	"helm.sh/helm/v3/pkg/release"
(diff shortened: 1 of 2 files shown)


technosophos (approved):
I debated for a while whether `unknown` was the right default, but in the end I agreed with your PR: `unknown` is the right value to set if the ManagedFieldsManager is manually set to ""

mattfarina:
To add extra detail for anyone who comes along to this in the future, `unknown` is what Kubernetes sets it to when it can't figure it out on its own. I discovered `unknown` as the phrasing during my testing. `unknown` is a common phrase for k8s when a state isn't known. For example, when Kubernetes can't figure out the User Agent for a request it's internal functions return `unknown`. This may be way `unknown` is set in my testing.

PR #9873, opened 2021-06-27
Add the ability to output the computed values using the template sub-command

jamoflaw:

This adds the ability to template the resolved values files (all the values including the chart defaults). This is equivalent to the output seen in the `--dry-run` `COMPUTED VALUES:` section.

The default behaviour of template is untouched and only changed when you use the `--computed-values` flag, which will output 
 the computed values _instead_ of the fully rendered manifest.

**example:**
```
# Display the computed values of the chart in its default state
helm template ./chart/name --computed-values
# Display the computed values of the chart after overrides have been applied
helm template ./chart/name --computed-values -f overrides.yaml
```

will output the values _after_ they have been resolved by the helm parsers.

This is useful in simple CI situations where a particular value is expected especially when traversing multiple environments, it is also useful as an input for other parsers or templating tools to provide customised output based on the values passed into the chart.

### DIFF
--- a/cmd/helm/template.go
+++ b/cmd/helm/template.go
@@ -34,6 +34,7 @@ import (
 	"helm.sh/helm/v3/cmd/helm/require"
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/chartutil"
+	"helm.sh/helm/v3/pkg/cli/output"
 	"helm.sh/helm/v3/pkg/cli/values"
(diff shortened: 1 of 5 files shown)


cndoit18:

cndoit18 (commented):
Please add the unit test.

jamoflaw:
Ref issue #9883

jamoflaw:
Have added some unit tests to cover the output - in all cases the output is the same as the output from `install` with `--debug --dry-run`

Run the tests locally and passes without issue

PR #9886, opened 2021-06-29
Adds urlutil.CanonicalAddr and use it when comparing chart repository and chart url

dol:
If a chart repository was added with https://myhost/chart and the
corresponding index.yaml file points to
https://myhost:443/helm/somechart-1.0.0.tgz the newly introduced
security check in v3.6.1 will mark the two hostname and scheme
comparison as different.
Canonicalize the hostname by always adding the port, if not defined, based
upon the used scheme.
http  maps to '80'
https maps to '443'

### DIFF
--- a/internal/urlutil/urlutil.go
+++ b/internal/urlutil/urlutil.go
@@ -17,6 +17,7 @@ limitations under the License.
 package urlutil
 
 import (
+	"net"
 	"net/url"
 	"path"
 	"path/filepath"
@@ -70,4 +71,21 @@ func ExtractHostname(addr string) (string, error) {
 		return "", err
 	}
 	return u.Hostname(), nil
+}
+
+// From: https://golang.org/src/net/http/transport.go:2712
+var portMap = map[string]string{
+	"http":  "80",
+	"https": "443",
+}
+
+// CanonicalAddr returns url.Host but always with a ":port" suffix
+// From: https://golang.org/src/net/http/transport.go:2719 without idna to ascii conversion
+func CanonicalAddr(url *url.URL) string {
+	addr := url.Hostname()
+	port := url.Port()
+	if port == "" {
+		port = portMap[url.Scheme]
+	}
+	return net.JoinHostPort(addr, port)
 }
--- a/internal/urlutil/urlutil_test.go
+++ b/internal/urlutil/urlutil_test.go
@@ -16,7 +16,10 @@ limitations under the License.
 
 package urlutil
 
-import "testing"
+import (
+	"net/url"
+	"testing"
+)
 
 func TestURLJoin(t *testing.T) {
 	tests := []struct {
@@ -75,6 +78,36 @@ func TestExtractHostname(t *testing.T) {
 	}
 	for start, expect := range tests {
 		if got, _ := ExtractHostname(start); got != expect {
+			t.Errorf("Got %q, expected %q", got, expect)
+		}
+	}
+}
+
+func TestCanonicalAddr(t *testing.T) {
+	tests := map[*url.URL]string{
+		{
+			Scheme: "http",
+			Host:   "example.com",
+			Path:   "/",
+		}: "example.com:80",
+		{
+			Scheme: "https",
+			Host:   "example.com",
+			Path:   "/",
+		}: "example.com:443",
+		{
+			Scheme: "https",
+			Host:   "example.com",
+			Path:   "/foo",
+		}: "example.com:443",
+		{
+			Scheme: "https",
+			Host:   "example.com:31337",
+			Path:   "/not/with/a/bang/but/a/whimper",
+		}: "example.com:31337",
+	}
+	for urlInput, expect := range tests {
+		if got := CanonicalAddr(urlInput); got != expect {
 			t.Errorf("Got %q, expected %q", got, expect)
 		}
 	}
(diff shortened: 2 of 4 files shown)


internal/urlutil/urlutil.go
```diff
@@ -71,3 +72,20 @@ func ExtractHostname(addr string) (string, error) {
...
+	"https": "443",
+}
+
+// CanonicalAddr returns url.Host but always with a ":port" suffix
+// From: https://golang.org/src/net/http/transport.go:2719 without idna to ascii conversion
+func CanonicalAddr(url *url.URL) string {
+	addr := url.Hostname()
+	port := url.Port()
+	if port == "" {
+		port = portMap[url.Scheme]
```
bacongobbler, line 88: how does this interact with [downloader plugins](https://helm.sh/docs/topics/plugins/#downloader-plugins) that can specify an alternative URL scheme, like `git://` or `oci://`? Looks like this only handles more "traditional" URL schemes but cannot account for other getter schemes registered with the helm client.

Can you add a test for those?

erpel: I've tried my hand at this, adding two test cases:
```
		{
			Scheme: "myprotocol",
			Host:   "example.com:1234",
			Path:   "/",
		}: "example.com:1234",
```
This test fails, because the existing code would append a `:` even if looking up a port number from the map comes up empty.

I also have a fix for that (very little experience with go, so probably not ideal). To share this code, should I open a PR against the branch in the fork by @dol or what would be proper Github etiquette for this?

dol: @bacongobbler Thank you for the hint. I wasn't aware of this plugin capability of helm. I need to do some digging to understand the topic in more depth.
At the moment I don't see any reason to add other protocols and their default port mapping. I currently considering to change the function name to `HttpCanonicalAddr` to give a clear hint that the function is only meant to be used for http schemes.

technosophos: I am unclear: Is there more to do on this PR, or is this for later/future work?

erpel: If the CanonicalAddr function is used in cases of downloader plugins, IMHO the PR must ensure that URL schemes that are not covered by the mapping are not changed. Currently, `myprotocol://example.com` would get "canonicalized" to `myprotocol://example.com:` which is probably not good.

I've pushed a rudimentary "fix" here: https://github.com/dol/helm/pull/52

dol: IMHO there is nothing additional to do.
The mentioned [Downloader Plugins](https://helm.sh/docs/topics/plugins/#downloader-plugins) allows for other protocols. The custom protocol handler is an external command that's outside of helm. For this reason the focus on http as protocol is IMHO enough.

bacongobbler: Okay, that partially makes sense.

However what I was referring to was the use of CanonicalAddr in the `action` package on install.go, line 693. It is entirely possible for that line of code to receive a URL like the test case @erpel provided above, as u1 could be a plugin getter.

In other words, if you intend to use this ONLY in the context of HTTP/HTTPS protocols, then it should be used ONLY in httpgetter.go.

Hope that makes sense.

cobexer:
The suggested interface here invites callers to misuse the normalized URL.
And is also in conflict with the recommendation of RFC3986: https://www.rfc-editor.org/rfc/rfc3986#section-3.2.3

```
... URI producers and
normalizers should omit the port component and its ":" delimiter if
port is empty or if its value would be the same as that of the
scheme's default.
```
Ignoring this brought us here... let us be better and do it right here

erpel:
I think modifying the normalization to conform with the RFC is a good addition. As long as we normalize the URL from the repository configuration and the chart url the same way, the underlying issue can still be addressed.
It is unfortunate that Jfrog caused this because they chose to produce URLs with explicit default ports but if repeating this here can be avoided let’s do it.

bacongobbler:
It may be worth asking JFrog if they're willing to fix this issue upstream. That way no fix has to be applied in Helm to work around an issue with a vendor who is not conforming to the RFC.

Thanks for the link @cobexer.

erpel:
After beginning to write up the issue for JFrog, I saw they did not have the problem on their official registry. They are aware of some issue introduced with helm 3.6.1: https://www.jfrog.com/jira/browse/RTFACT-26063 is marked as resolved and mentioned in the release notes for 7.24.1/7.24.3.
I'm upgrading our instance now to verify this.

tof1a:
As now jfrog is adding the port in index.yaml urls, this issue is very problematic.
In order to avoid to use the --pass-credentials (which is somehow a security hole), we have to configure all our configuration with helm repo defined with port.
Even if now jfrog remove the port (which I doubt they will do anyway) this will break peope who have defined helm repo with port !
So only valid solution for me, is the implementation of this PR, and honestly I even see this as a bug added in 3.6.1 as having credentials shared for url should not be dependent if standard http/https ports are provided, so it could even be part of bug fix release in 3.6/3.7, but if we already have it in 3.8 it would be great.
Any chances to have it part of 3.8.0 as it has been removed already from 3.7.0 milestone ?

bacongobbler:
Closing as this pull request targets the `release-3.6.2` branch. In the future, please target your fixes against `main`. Please feel free to re-open this pull request against `main`. Thanks.

PR #9680, opened 2021-05-07
fix: dep up when exists tmpcharts

jorgevillaverde:
Closes: #5567

Signed-off-by: jorgevillaverde <jorge.villaverde@toucansoft.io>

**What this PR does / why we need it**: If the process is killed when running `helm dep up` and the folder `tmpcharts` was created and not delete before the process was kill, the next time the user ties to run `helm dep up` will result on the error

```
Error: Unable to move current charts to tmp dir: rename /mychart/charts /mychart/tmpcharts: file exists
```

This PR adds a check to see if `tmpcharts` exists, and delete it before [RenameWithFallback](https://github.com/helm/helm/blob/main/pkg/downloader/manager.go#L261). This prevents the error from happening.

**Special notes for your reviewer**:

**If applicable**:
- [ ] this PR contains documentation
- [x] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

PR #9889, opened 2021-06-29
fix: refactor downloadAll

### LINKED ISSUE #5567: when killing `helm dep up` the next `dep up` fails

bacongobbler:
This refactor cleans up downloadAll's validation, download, and save logic:

1. A temporary directory is created, and removed after all references to
   the struct have been dropped via `defer`
2. Any local dependencies in the `charts` directory are kept intact and validated
3. Charts that have been updated are moved to the `charts` directory
4. Outdated charts in `charts` are removed (barring local dependencies)

This refactor has a number of improvements, including:

- tmpCharts is removed after execution
- no remote charts are downloaded to destPath: they are all pulled into tmpPath, validated, then moved to destPath
- lots of code cleanup/improvements, like the `if` block checking whether the `charts` directory was actually not a directory.
- the cyclomatic complexity of the code was simplified
- extra (and in some cases, dangerous) calls to `os.RemoveAll(destPath)` have been removed, cleaning the code and preventing certain failure cases.

A test has been provided to demonstrate the tmpCharts removal issue has been fixed. I also left a comment that further improvements can be made to this piece of code should someone feel the need to make improvements there.

closes #5567
closes #8846 
closes #9680

### DIFF
--- a/pkg/downloader/manager.go
+++ b/pkg/downloader/manager.go
@@ -249,22 +249,24 @@ func (m *Manager) downloadAll(deps []*chart.Dependency) error {
 	destPath := filepath.Join(m.ChartPath, "charts")
 	tmpPath := filepath.Join(m.ChartPath, "tmpcharts")
 
-	// Create 'charts' directory if it doesn't already exist.
-	if fi, err := os.Stat(destPath); err != nil {
+	// Check if 'charts' directory is not actally a directory. If it does not exist, create it.
+	if fi, err := os.Stat(destPath); err == nil {
+		if !fi.IsDir() {
+			return errors.Errorf("%q is not a directory", destPath)
+		}
+	} else if os.IsNotExist(err) {
 		if err := os.MkdirAll(destPath, 0755); err != nil {
 			return err
 		}
-	} else if !fi.IsDir() {
-		return errors.Errorf("%q is not a directory", destPath)
+	} else {
+		return fmt.Errorf("unable to retrieve file info for '%s': %v", destPath, err)
 	}
 
-	if err := fs.RenameWithFallback(destPath, tmpPath); err != nil {
-		return errors.Wrap(err, "unable to move current charts to tmp dir")
-	}
-
-	if err := os.MkdirAll(destPath, 0755); err != nil {
+	// Prepare tmpPath
+	if err := os.MkdirAll(tmpPath, 0755); err != nil {
 		return err
 	}
+	defer os.RemoveAll(tmpPath)
 
 	fmt.Fprintf(m.Out, "Saving %d charts\n", len(deps))
 	var saveError error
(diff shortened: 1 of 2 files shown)


jdolitsky:
@bacongobbler - checked the failing tests. I believe I experienced the same "no such file or directory" w/ the OCI changes... May need to make a change similar to https://github.com/helm/helm/pull/9782/files#diff-a689ad99783d272ab2c91f3372579ffa3e0e96a64ead1b2872e05ba8e86d953b. Just a hunch

bacongobbler:
Good call. Thanks! I'll look into that.

bacongobbler:
Hmm... I think the issue was due to a simple bug in my code. I think I caught it and it should work as expected now.

bacongobbler:
I'm going to change some of the behaviour around `safeDeleteDep`. It looks like there can be a case where a fail to "safely" delete a dependency can cause data loss. That doesn't seem optimal. Code should remain the same, but updates should be more atomic as a result.

jdolitsky (approved):

bacongobbler:
I think I failed to account for one last use case, which is when a "local" dependency (one where the "repository" field in your dependencies is empty) is packaged as a tarball instead of a directory. I've seen some use cases utilizing that.

bacongobbler:
> That may call for some deeper introspection of the chart dependency tree than I was hoping for, but /shrug

done

bacongobbler:
As it turns out, local dependencies cannot be anything other than directories according to https://github.com/helm/helm/blob/21251378b432fb629cb1a49af3892e20c1ec9624/pkg/downloader/manager.go#L277, so I just added some extra safeguards in case we decide to change that assumption. :)

hickeyma (changes requested):
@bacongobbler Do you mind rebasing to resolve the conflict? I will finish me review then.

bacongobbler:
rebased and squashed.

technosophos (approved):

Issue #9829, opened 2021-06-18
Helm annotations nested in List objects are not respected


arnouthoebreckx:
Output of `helm version`: 
version.BuildInfo{Version:"v3.1.1", GitCommit:"afe70585407b420d0097d07b21c47dc511525ac8", GitTreeState:"clean", GoVersion:"go1.13.8"}

Output of `kubectl version`: 
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.0", GitCommit:"2bd9643cee5b3b3a5ecbd3af49d09018f0773c77", GitTreeState:"clean", BuildDate:"2019-09-18T14:36:53Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"windows/amd64"}
Server Version: version.Info{Major:"1", Minor:"11+", GitVersion:"v1.11.0+d4cacc0", GitCommit:"d4cacc0", GitTreeState:"clean", BuildDate:"2021-02-21T19:36:48Z", GoVersion:"go1.10.8", Compiler:"gc", Platform:"linux/amd64"}

Cloud Provider/Platform (AKS, GKE, Minikube etc.): Openshift 3.11

-----------------------------------------------------------------------------------------
Helm does not respect annotations held by Objects within a List. This might be a serious design flaw as a List object does not exist within a cluster and you would expect that the contained Objects would be validated for annotations.

When a List is created containing Objects with the annotation "helm.sh/resource-policy: keep", you would expect when an object is removed from the List that the Object would still be kept, but this is not the case.

Let's say we have the following namespace.yaml template which creates a List of Namespaces.

```
---
{{- if .Values.namespaces -}}
apiVersion: v1
kind: List
items:
{{- range $namespace := (.Values.namespaces | uniq) }}
  - apiVersion: v1
    kind: Namespace
    metadata:
      name: {{ .name }}
      annotations:
        helm.sh/resource-policy: keep
      labels:
        name: {{ .name }}
{{- end }}
{{- end }}
```
With the following values file
```
namespaces:
- test-ns-1
- test-ns-2
- test-ns-3
```

When a namespace is removed and the helm release is updated you would expect the namespace to be kept because of the annotation. Sadly this is not the case and the namespace is deleted even though the annotation is present on it.

A workaround would be to update the YAML file to look this with the annotation set on the List. But as mentioned above a List is not a real-time Object in the cluster and thus is not ideal.
```
---
{{- if .Values.namespaces -}}
apiVersion: v1
kind: List
[...]

PR #9914, opened 2021-07-06
feat(linter): check for nested resource-policy annotations

adamreese:
The 'helm.sh/resource-policy' annotation is only supported on top level
objects.  The annotation is ignored if given on a nested object within a
list.

Ref #9829


technosophos (approved):
LGTM

PR #9918, opened 2021-07-07
fix(install): fix 9915

### LINKED ISSUE #9915: `helm install` installs crds from disabled child charts
https://github.com/helm/helm/blob/main/pkg/chart/chart.go#L153-L168

We have a parent chart that wants to conditionally enable/disable child charts. We don't want crds from child charts to be installed in the cluster, that child chart is not enabled. From Helm code, it seems that conditions are not considered when calculating `CRDObjects()`. This feels like a bug.

Output of `helm version`: v3.6.1

Output of `kubectl version`:

cndoit18:

fixes: #9915 

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -198,6 +198,10 @@ func (i *Install) RunWithContext(ctx context.Context, chrt *chart.Chart, vals ma
 		return nil, err
 	}
 
+	if err := chartutil.ProcessDependencies(chrt, vals); err != nil {
+		return nil, err
+	}
+
 	// Pre-install anything in the crd/ directory. We do this before Helm
 	// contacts the upstream server and builds the capabilities object.
 	if crds := chrt.CRDObjects(); !i.ClientOnly && !i.SkipCRDs && len(crds) > 0 {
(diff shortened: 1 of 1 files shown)


yxxhero:
@cndoit18 from the issue, it  conditionally enable/disable child charts. example: 
```yaml
dependencies: # A list of the chart requirements (optional)
  - name: The name of the chart (nginx)
    version: The version of the chart ("1.2.3")
    repository: (optional) The repository URL ("https://example.com/charts") or alias ("@repo-name")
    condition: (optional) A yaml path that resolves to a boolean, used for enabling/disabling charts (e.g. subchart1.enabled )
```
I think you should decide whether to load chart's crds by **condition** filed. do you think of?

cndoit18:
> @cndoit18 from the issue, it conditionally enable/disable child charts. example:
> 
> ```yaml
> dependencies: # A list of the chart requirements (optional)
>   - name: The name of the chart (nginx)
>     version: The version of the chart ("1.2.3")
>     repository: (optional) The repository URL ("https://example.com/charts") or alias ("@repo-name")
>     condition: (optional) A yaml path that resolves to a boolean, used for enabling/disabling charts (e.g. subchart1.enabled )
> ```
> 
> I think you should decide whether to load chart's crds by **condition** filed. do you think of?

Very good, I gave missed this part. I'll rethink it.

pkg/action/action_test.go
```diff
@@ -289,3 +290,12 @@ func TestGetVersionSet(t *testing.T) {
 		t.Error("Non-existent version is reported found.")
 	}
 }
+
+func loadChart(t *testing.T, path string) *chart.Chart {
```
yxxhero, line 294: this is a utils func, why has testing.T arg? you can return (*chart.Chart, error), WDYT?

yxxhero (changes requested):

yxxhero (approved):
LGTM
@bacongobbler what do  you think of?

hickeyma (approved):
LGTM, thanks @cndoit18 for the fix. Verified by testing manually with and without the fix.  I cannot think of any way we can test this with CRDs, so unit test is not required.

This requires another core maintainer to merge.

bacongobbler (approved):

PR #9933, opened 2021-07-11
Update Cobra to version 1.2.1

marckhouzam:

Update to Cobra 1.2.1.
I don't know why Dependabot was not triggered for this.

The changes part of Cobra 1.2.1 are safe and are mostly focused on bug fixes for the completion support.

A bash completion V2 is also available. I will eventually post a PR proposing to use this new version.

A completion test was removed in this PR as it was failing with the new version of Cobra.
The test was actually testing Cobra's behaviour (which changed in the new release), which is not something Helm tests should focus on, so I simply removed it.

I don't expect Cobra's change of behaviour to have much of an impact on Helm, and I was being over-zealous when I wrote the test.

The completion tests of the acceptance-testing repo continue to pass.

**_Should I be checking the changes to go.sum in any way?_**


marckhouzam:
Here are the release notes for the [1.2.0](https://github.com/spf13/cobra/releases/tag/v1.2.0) and [1.2.1](https://github.com/spf13/cobra/releases/tag/v1.2.1) release of Cobra, if anyone is curious.

cmd/helm/plugin_test.go
```diff
@@ -277,11 +277,6 @@ func TestPluginDynamicCompletion(t *testing.T) {
 		cmd:    "__complete echo -n mynamespace ''",
 		golden: "output/plugin_echo_no_directive.txt",
 		rels:   []*release.Release{},
-	}, {
-		name:   "completion for plugin bad directive",
-		cmd:    "__complete echo ''",
-		golden: "output/plugin_echo_bad_directive.txt",
-		rels:   []*release.Release{},
```
mattfarina, line 284: Out of curiosity, why was this test dropped?

marckhouzam: (From the description of the PR 😄 ):

A completion test was removed in this PR as it was failing with the new version of Cobra.
The test was actually testing Cobra's behaviour (which changed in the new release), which is not something Helm tests should focus on, so I simply removed it.

I don't expect Cobra's change of behaviour to have much of an impact on Helm, and I was being over-zealous when I wrote the test.

hickeyma (approved):
LGTM, thanks @marckhouzam.

PR #9938, opened 2021-07-13
Use jsonmergepatch.CreateThreeWayJSONMergePatch for unstructured objects

### LINKED ISSUE #9937: Custom resources should update with three way merge patch
Builtin resources like Deployment update with three way merge as expected. However custom resources like istio VirtualService update with two way merge patch which I think isn't expected.

Looking at the code it seems that custom resources are parsed as unstructured objects, an a two way merge patch is used for these objects.

https://github.com/helm/helm/blob/433b90c4b6010415524bfd98b77efca0e6ec7205/pkg/kube/client.go#L457

kubectl apply uses `jsonmergepatch.CreateThreeWayJSONMergePatch` to create three way merge patch for unstructured objects as well:

https://github.com/kubernetes/kubectl/blob/b83b2ec7d15f286720bccf7872b5c72372cb8e80/pkg/cmd/apply/patcher.go#L129

I'd like to file a PR for this if the current behavior isn't intended.

wxdao:

closes #9937

Use jsonmergepatch.CreateThreeWayJSONMergePatch to create three way merge patch for unstructured objects like kubectl apply does.

**If applicable**:

### DIFF
--- a/cmd/helm/root.go
+++ b/cmd/helm/root.go
@@ -163,6 +163,8 @@ func newRootCmd(actionConfig *action.Configuration, out io.Writer, args []string
 	}
 	actionConfig.RegistryClient = registryClient
 
+	actionConfig.UseThreeWayMergePatchForUnstructured = settings.UseThreeWayMergePatchForUnstructured
+
 	// Add subcommands
 	cmd.AddCommand(
 		// chart commands
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,6 @@ require (
 	github.com/docker/distribution v2.7.1+incompatible
 	github.com/docker/docker v17.12.0-ce-rc1.0.20200618181300-9dc6525e6118+incompatible
 	github.com/docker/go-units v0.4.0
-	github.com/evanphx/json-patch v4.9.0+incompatible
 	github.com/gobwas/glob v0.2.3
 	github.com/gofrs/flock v0.8.0
 	github.com/gosuri/uitable v0.0.4
--- a/go.mod
+++ b/go.mod
@@ -16,6 +16,7 @@ require (
 	github.com/docker/distribution v2.7.1+incompatible
 	github.com/docker/docker v17.12.0-ce-rc1.0.20200618181300-9dc6525e6118+incompatible
 	github.com/docker/go-units v0.4.0
+	github.com/evanphx/json-patch v4.9.0+incompatible
 	github.com/gobwas/glob v0.2.3
 	github.com/gofrs/flock v0.8.0
 	github.com/gosuri/uitable v0.0.4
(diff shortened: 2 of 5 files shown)


cndoit18 (approved):
LGTM

mattfarina (changes requested):
This needs to sit behind an opt-in flag. See https://github.com/helm/helm/issues/9937#issuecomment-880853027 for some more detail.

wxdao:
@mattfarina

Since `kube.Client.Update` is called at multiple places besides upgrade action, I've added a `--use-three-way-merge-patch-for-unstructured` global flag to control the behavior.

Yes, the name of the flag is a bit ridiculously long. I'd like some idea about how to name it better if any...

bacongobbler:
Say someone was upgrading from a version of Helm without this flag, then started using it after upgrading to Helm 3.7. Would Kubernetes be able to resolve the upgrade or would it return an error? Or worse, is the behaviour underfined when switching from a generic merge patch to a generic three-way JSON merge patch?

Conversely, if someone calls `helm --use-three-way-merge-patch-for-unstructured install mychart .` then `helm upgrade mychart .` (generic three-way JSON merge patch to generic merge patch), what happens?

I'm trying to determine if users can use these merge patch strategies interchangeably or not. Because I can imaging the scenario where someone forgets to add this flag during a `helm upgrade` and suddenly puts their Helm release into an irrecoverable state.

wxdao:
> Say someone was upgrading from a version of Helm without this flag, then started using it after upgrading to Helm 3.7. Would Kubernetes be able to resolve the upgrade or would it return an error? Or worse, is the behavior undefined when switching from a generic merge patch to a generic three-way JSON merge patch?
> 
> Conversely, if someone calls `helm --use-three-way-merge-patch-for-unstructured install mychart .` then `helm upgrade mychart .` (generic three-way JSON merge patch to generic merge patch), what happens?
> 
> I'm trying to determine if users can use these merge patch strategies interchangeably or not. Because I can imagine the scenario where someone forgets to add this flag during a `helm upgrade` and suddenly puts their Helm release into an irrecoverable state.

To summarize any behavior change if the flag is set:
- Only affects custom resources, like VirtualService, and CustomResourceDefinition resources.
- Only affects changes to those resources, i.e. mostly when running `helm upgrade` and `helm rollback`.

It's ok to interchangeably use any strategies. Because:
- There will be no difference for the stored release meta, since it's the rendered manifests that are stored. Patch strategies only affect how the live state of resources will become after an update.
- The behavior of either patch strategies is determined, and has nothing to do with what was used previously. Actually the two way merge patch will only miss out some changes and never do more. See below for the only possible difference of these two patch strategies:

```
# original / last applied
a: 1

# target / to update to
a: 1

# live state
a: 2

# result live state
a: 2 # if it's two way
a: 1 # if it's three way
```

Sorry for the late reply.

zonggen:
Could you provide unit tests for this feature? would love to keep the review cycles going :)

chris-str-cst (approved):

Issue #5550, opened 2019-04-04
cannot unmarshal string into Go value of type releaseutil.SimpleHead


Saurabh-16:
I am getting unmarshal error on simple validation that I am performing from my .tpl file.

I have a function defined in _helpers.tpl like:-

```
{{- define "actionValidate" -}}
  {{ $action := .Values.actions }}
  {{- if or (eq $action "action1") (eq $action "action2") (eq $action "action3") -}}
    true
  {{- end -}}
{{- end -}}

```
I 'm  invoking this function from one of the .tpl file inside templates folder .

`{{ include "actionValidate" .  | required "Action value is incorrect. The valid values are 'action1', 'action2', 'action3' " }}`

On running helm install command , I'm getting following issue:-
'**_error unmarshaling JSON: json: cannot unmarshal string into Go value of type releaseutil.SimpleHead_**'

However helm dry run command does not show any error and is executed successfully.

Output of `helm version`:
 Client: &version.Version{SemVer:"v2.13.1", GitCommit:"618447cbf203d147601b4b9bd7f8c37a5d39fbb4", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.13.1", GitCommit:"618447cbf203d147601b4b9bd7f8c37a5d39fbb4", GitTreeState:"clean"}

Output of `kubectl version`:
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.0", GitCommit:"641856db18352033a0d96dbc99153fa3b27298e5", GitTreeState:"clean", BuildDate:"2019-03-25T15:53:57Z", GoVersion:"go1.12.1", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.0", GitCommit:"641856db18352033a0d96dbc99153fa3b27298e5", GitTreeState:"clean", BuildDate:"2019-03-25T15:45:25Z", GoVersion:"go1.12.1", Compiler:"gc", Platform:"linux/amd64"}

Saurabh-16:
Is this an issue or there is something wrong with the code syntax .

hickeyma:
@Saurabh-16 I am unable to reproduce this issue. Here is the output of my investigation using a scaffold chart (`helm create`):

```console
$ cat chrt-5550/templates/_helpers.tpl 
 [....]
{{/*
Create chart name and version as used by the chart label.
*/}}
{{- define "chrt-5550.chart" -}}
[...]

PR #9941, opened 2021-07-14
A better error message for issue #5500

wizbit:
closes #5500 cannot unmarshal string into Go value of type releaseutil.SimpleHead

Signed-off-by: Marcus Speight <marcus.speight@hotmail.co.uk>

This PR provides a better error message to explain the error described in #5500

Happy for the error message to be alter if better wording is required.
Test is based on @caleb15 replication steps in the issue.

**If applicable**:

### DIFF
--- a/cmd/helm/template_test.go
+++ b/cmd/helm/template_test.go
@@ -131,6 +131,12 @@ func TestTemplateCmd(t *testing.T) {
 			cmd:    fmt.Sprintf(`template '%s' --skip-tests`, chartPath),
 			golden: "output/template-skip-tests.txt",
 		},
+		{
+			name:      "template with required statement",
+			cmd:       fmt.Sprintf("template '%s'", "testdata/testcharts/issue-5550"),
+			wantError: true,
+			golden:    "output/issue-5550.txt",
+		},
 	}
 	runTestCmd(t, tests)
 }
--- a/pkg/releaseutil/manifest_sorter.go
+++ b/pkg/releaseutil/manifest_sorter.go
@@ -143,6 +143,10 @@ func (file *manifestFile) sort(result *result) error {
 
 		var entry SimpleHead
 		if err := yaml.Unmarshal([]byte(m), &entry); err != nil {
+			// Return a more understandable error message when the yaml fails to parse
+			if strings.Contains(err.Error(), "releaseutil.SimpleHead") {
+				return errors.Errorf("A non Kubernetes resource was generated on %s", file.path)
+			}
 			return errors.Wrapf(err, "YAML parse error on %s", file.path)
 		}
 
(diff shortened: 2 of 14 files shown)


pkg/releaseutil/manifest_sorter.go
```diff
@@ -143,6 +143,10 @@ func (file *manifestFile) sort(result *result) error {
 
 		var entry SimpleHead
 		if err := yaml.Unmarshal([]byte(m), &entry); err != nil {
+			// Return a more understandable error message when the yaml fails to parse
+			if strings.Contains(err.Error(), "releaseutil.SimpleHead") {
```
cndoit18, line 147: May you use something like `errors.As` or `errors.Is` instead of it?

technosophos: @wizbit just checking in on whether you had an answer for @cndoit18 's question.

wizbit: I don't know if it is possible. The error comes from the following stack

encoding/json/decode.go#938 `d.saveError(&UnmarshalTypeError{Value: "string", Type: v.Type(), Offset: int64(d.readIndex())})`
yaml.go#76 `return fmt.Errorf("while decoding JSON: %v", err)`
yaml.go#60 `return fmt.Errorf("error unmarshaling JSON: %v", err)`

My understanding of this is the `UnmarshalTypeError` is removed from the error in yaml.go such that only a string error exists.

cndoit18: I checked this error chain and they don't support the wrapper.

It looks like the only way to compare is by strings.

cndoit18 (approved):
LGTM

zonggen:
If I'm not mistaken, this should be linked to https://github.com/helm/helm/issues/5550 :)

cmd/helm/template_test.go
```diff
@@ -131,6 +131,12 @@ func TestTemplateCmd(t *testing.T) {
 			cmd:    fmt.Sprintf(`template '%s' --skip-tests`, chartPath),
 			golden: "output/template-skip-tests.txt",
 		},
+		{
+			name:      "template with required statement",
```
zonggen, line 135: Maybe rename this test to reflect the root cause as you stated here: https://github.com/helm/helm/issues/5550#issuecomment-879130623.

zonggen:
Local testing:

`--- PASS: TestTemplateCmd/template_with_required_statement (0.01s)`.

```
$ helm template foo2
Error: A non Kubernetes resource was generated on foo2/templates/deployment.yaml

```

MarcusSorealheis:
I miss the automation infra of this repo

PR #9945, opened 2021-07-15
fix(resolver): fix issue 9944

### LINKED ISSUE #9944: bug: version is NOT optional anymore for file://.. subcharts
Output of `helm version`: v3.6.3

Output of `kubectl version`:

PR: https://github.com/helm/helm/pull/9736/files has introduced a change that is breaking `helm dep up` command. In prior release of helm, `version` is optional for sub charts loaded from local filesystem `file://`. So, when the version is not set, https://github.com/helm/helm/pull/9736/files#diff-ff3593e6a6cb00f927118c829760e054ccca7686eae492f9f4f18b9b851a9a20R60-R63 line is throwing error `"dependency %q has an invalid version/constraint format"`. 

I think a possible solution will be to set the constraint to `*` is it was not set and dependency was loaded from `file://` .

cndoit18:

fixes: #9944

**If applicable**:

### DIFF
--- a/internal/resolver/resolver.go
+++ b/internal/resolver/resolver.go
@@ -18,6 +18,7 @@ package resolver
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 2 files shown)


internal/resolver/resolver.go
```diff
@@ -57,7 +57,11 @@ func (r *Resolver) Resolve(reqs []*chart.Dependency, repoNames map[string]string
 	locked := make([]*chart.Dependency, len(reqs))
 	missing := []string{}
 	for i, d := range reqs {
-		constraint, err := semver.NewConstraint(d.Version)
+		v := d.Version
+		if len(strings.TrimSpace(v)) == 0 {
```
tamalsaha, line 61: `if len(strings.TrimSpace(v)) == 0 && strings.HasPrefix(d.Repository, "file://") {`

cndoit18: Is just `file://` optional?

tamalsaha: That is my understanding. I can't confirm. https://helm.sh/docs/helm/helm_dependency/

PR #9971, opened 2021-07-23
WIP: Render templates in parallel

ls-jad-elkik:

This PR makes the template rendering process run in parallel to optimize the `helm template` and `helm install` operations when the chart contains lots of files to render.

We've been using this for a few months and haven't noticed any difference in the final output of the `template` step.

**If applicable**:

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -22,8 +22,11 @@ import (
 	"path"
 	"path/filepath"
 	"regexp"
+	"runtime"
 	"sort"
+	"strconv"
 	"strings"
+	"sync"
 	"text/template"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


yxxhero:
@ls-jad-elkik Would you mind adding unittests. Thanks you very much.

pkg/engine/engine.go
```diff
@@ -126,6 +147,10 @@ func (e Engine) initFunMap(t *template.Template, referenceTpls map[string]render
 
 	// Add the 'tpl' function here
 	funcMap["tpl"] = func(tpl string, vals chartutil.Values) (string, error) {
+		// if tpl is not a template just return the string.
+		if !strings.Contains(tpl, "{") {
```
bacongobbler, line 151: what does this change accomplish with regards to parallel processing?

pkg/engine/engine.go
```diff
@@ -255,19 +291,55 @@ func (e Engine) renderWithReferences(tpls, referenceTpls map[string]renderable)
...
+			if renderResult.error != nil {
+				return map[string]string{}, cleanupExecError(renderResult.filename, renderResult.error)
+			}
+			rendered[renderResult.filename] = renderResult.rendered
+		}
+	}
+	return rendered, nil
+}
+
+type RenderResults struct {
```
bacongobbler, line 322: Unless this is something that is returned in a function signature, this should be marked private to the package using `renderResults` instead of `RenderResults` as the name.

pkg/engine/engine.go
```diff
@@ -102,14 +105,32 @@ var warnRegex = regexp.MustCompile(warnStartDelim + `(.*)` + warnEndDelim)
 func warnWrap(warn string) string {
 	return warnStartDelim + warn + warnEndDelim
 }
+func goid() int {
+	var buf [64]byte
```
bacongobbler, line 109: so if I'm reading this right... If the stack exceeds 64 goroutines we will return a panic stack trace. That seems like a major regression. Can you clarify the behaviour here?

There are several larger charts that will very easily hit this limit with their usage of the `tpl` function.

pkg/engine/engine.go
```diff
@@ -102,14 +105,32 @@ var warnRegex = regexp.MustCompile(warnStartDelim + `(.*)` + warnEndDelim)
...
+	n := runtime.Stack(buf[:], false)
+	idField := strings.Fields(strings.TrimPrefix(string(buf[:n]), "goroutine "))[0]
+	id, err := strconv.Atoi(idField)
+	if err != nil {
+		panic(fmt.Sprintf("cannot get goroutine id: %v", err))
```
bacongobbler, line 114: there must be a cleaner way to handle the conversion than to return a `panic`...

bacongobbler (changes requested):
I will echo the same comment as @yxxhero here and mention that we do ask for test coverage to accompany changes. The goal is to cover all possible code paths the code introduces, as well as test out a few common scenarios like "render an example chart" as well as "render a more complex chart". In this case, we probably want a chart that will test the limits of running multiple template renders in parallel.

ls-jad-elkik:
I will close this. The patch has been working for us but there's less need for it now. And there were some edge cases where it did break stuff, probably related to the comments in this PR.

PR #9980, opened 2021-07-28
feat(helm): add --password-stdin to `helm repo add`

### LINKED ISSUE #10032: `helm repo add` is missing --password-stdin
**What**

You currently can't pipe the password into `helm repo add`:

```
$ echo password | helm repo add repo-name https://repo-url --username username
Password:
Error: inappropriate ioctl for device
```

I would expect the option `--password-stdin` to be available:

```
$ echo password | helm repo add repo-name https://repo-url --username username --password-stdin
"repo-name" has been added to your repositories
```

**Why**

There are two reasons I see for adding this:

* I personally would expect that it's possible to pipe the password into
  `helm repo add` but that's currently not the case. If I understand it
  correctly, you currently either need to pass the password via a cli
  parameter (`--password`) or use a expect/send mechanism. The
  cli parameter mechanism also isn't secure as it might expose the
  password via the process name.

* Subcommands like `helm registry login` already support
  `--password-stdin`. The cli interfaces should be consistent regarding
  which options they support.

schra:
**What**

Without this commit, you can't pipe the password into `helm repo add`:

```
$ echo password | helm repo add repo-name https://repo-url --username username
Password:
Error: inappropriate ioctl for device
```

This commit introduces `--password-stdin`:

```
$ echo password | helm repo add repo-name https://repo-url --username username --password-stdin
"repo-name" has been added to your repositories
```

Fixes #10032

**Why**

There are two reasons I see for adding this:

* I personally would expect that it's possible to pipe the password into
  `helm repo add` but that's currently not the case. If I understand it
  correctly, you currently either need to pass the password via a cli
  parameter (`--password`) or use a expect/send mechanism.

* Subcommands like `helm registry login` already support
  `--password-stdin`. The cli interfaces should be consistent regarding
  which options they support.

**Notes**

* I basically just copy-pasted code from `cmd/helm/registry_login.go`.

* I don't know if this is the best approach to enable piping into the command. Maybe it would be sufficient to replace `term.ReadPassword(fd)` with something else that also supports piping/ doesn't yield `inappropriate ioctl for device`.

**If applicable**:

### DIFF
--- a/cmd/helm/repo_add.go
+++ b/cmd/helm/repo_add.go
@@ -146,7 +146,7 @@ func (o *repoAddOptions) run(out io.Writer) error {
 
 	if o.username != "" && o.password == "" {
 		if o.passwordFromStdinOpt {
-			passwordFromStdin, err := ioutil.ReadAll(os.Stdin)
+			passwordFromStdin, err := io.ReadAll(os.Stdin)
 			if err != nil {
 				return err
 			}
--- a/cmd/helm/repo_add.go
+++ b/cmd/helm/repo_add.go
@@ -48,6 +48,7 @@ type repoAddOptions struct {
 	url                  string
 	username             string
 	password             string
+	passwordFromStdinOpt bool
 	passCredentialsAll   bool
 	forceUpdate          bool
 	allowDeprecatedRepos bool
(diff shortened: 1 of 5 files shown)


cndoit18:

cmd/helm/repo_add.go
```diff
@@ -143,14 +145,24 @@ func (o *repoAddOptions) run(out io.Writer) error {
...
 
 	if o.username != "" && o.password == "" {
-		fd := int(os.Stdin.Fd())
-		fmt.Fprint(out, "Password: ")
-		password, err := term.ReadPassword(fd)
-		fmt.Fprintln(out)
-		if err != nil {
-			return err
+		if o.passwordFromStdinOpt {
+			passwordFromStdin, err := ioutil.ReadAll(os.Stdin)
```
cndoit18, line 149: Is it possible to use `fmt.Scanln` here instead, or use the `io.ReadAll` instead?

cndoit18 (commented):
Could you add some unit tests?

cmd/helm/repo_add.go
schra: Done

technosophos (approved):
Code LGTM

yxxhero (approved):
LGTM

bacongobbler (approved):

Issue #9921, opened 2021-07-07
Move to GitHub Actions


mattfarina:
The current CI setup has gotten more flaky and slower over time. For example, there have been numerous times where a PR has been submitted and the CI webhook gets a 400 (Bad Request) response from CircleCI. No details on why its bad are provided and resubmitting it also returns a 400.

In addition to that, test run times have gotten a lot longer. For example, we have had to increase the lint test timeout once and may need to again due to recent timeouts. All of this while the codebase is not significantly different in size. The time to build binaries for releases has also gone up a lot.

Moving to GitHub Actions will remove the 400 errors and should perform at a faster speed. In addition to that, we can add better windows testing.

cndoit18:
This is great!

bacongobbler:
Actions still need to be implemented:

- [ ] make build-pr trigger on merges to the main branch
- [ ] disable CircleCI builds
- [ ] create a "publish" action

tylerauerbeck:
@bacongobbler Happy to help out here on 1 & 3 if this is still available.

bacongobbler:
Feel free!

PR #9989, opened 2021-07-31
Add deploy github actions workflows

tylerauerbeck:
Signed-off-by: Tyler Auerbeck <tauerbec@redhat.com>

refs #9921 

This PR adds the remaining required github actions workflows in order to complete the transition from CircleCI. Specifically the following will now happen:
- build-pr will now run when a PR is merged to the main branch
- deploy will now happen when:
  - a tag is pushed to the repository
  - the workflow is manually run from GitHub Actions with a provided version name (some questions here if this is desired)
  - on a push/merge to the main branch, a deploy happens with VERSION=canary

@bacongobbler This is about 99% complete. Just wanted to put this out as a draft for you to take a look at while I finish some final testing. One question I had is really around the release process. As I'm not familiar with the release process, I just copied what I saw in the `deploy.sh` script. So things like do something when a tag is specified, etc. However, after doing some work in other helm repositories (chart-release, chart-testing), I noticed those repo's also had a `workflow-dispatch` action that allow them to trigger the release manually via GHA. Is this something that would be desired here? I've stubbed out a majority of it in case you would be interested. But otherwise, let me know and I can pull it out.

This will require two secrets to be created inside of the repo in GitHub:
- AZURE_STORAGE_CONNECTION_STRING
- AZURE_STORAGE_CONTAINER_NAME

To take a look at the output of this, you can look at a successful run of the workflow here:
https://github.com/tylerauerbeck/helm/actions/runs/1099917788

**If applicable**:

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -3,6 +3,10 @@ on:
   pull_request:
     branches:
       - main
+  push:
+    branches:
+      - main
+
 jobs:
   build:
     runs-on: ubuntu-latest
@@ -27,3 +31,31 @@ jobs:
         run: make test-style
       - name: Run unit tests
         run: make test-coverage
+
+  canary-release:
+    runs-on: ubuntu-latest
+    needs: build
+    env:
+      AZURE_STORAGE_CONNECTION_STRING: "${{ secrets.AZURE_STORAGE_CONNECTION_STRING }}"
+      AZURE_STORAGE_CONTAINER_NAME: "${{ secrets.AZURE_STORAGE_CONTAINER_NAME }}"
+      VERSION: canary
+    if: ${{ github.event }} == "push" && github.ref == 'refs/heads/main'
+    steps:
+      - name: Checkout source code
+        uses: actions/checkout@v2
+
+      - name: Setup Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: '1.16'
+
+      - name: Build Helm Binaries
+        run: |
+          make build-cross
+          make dist checksum VERSION="${VERSION}"
+
+      - name: Azure CLI Action
+        uses: Azure/cli@1.0.4
+        with:
+          inlineScript: |
+            az storage blob upload-batch -s _dist/ -d "$AZURE_STORAGE_CONTAINER_NAME" --pattern 'helm-*' --connection-string "$AZURE_STORAGE_CONNECTION_STRING"
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -55,7 +55,7 @@ jobs:
           make dist checksum VERSION="${VERSION}"
 
       - name: Azure CLI Action
-        uses: Azure/cli@1.0.4
+        uses: Azure/cli@v1
         with:
           inlineScript: |
             az storage blob upload-batch -s _dist/ -d "$AZURE_STORAGE_CONTAINER_NAME" --pattern 'helm-*' --connection-string "$AZURE_STORAGE_CONNECTION_STRING"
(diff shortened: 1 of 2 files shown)


tylerauerbeck:
Should the CircleCI config be removed as part of this? Or should this be done in a follow up after we see these actions working ?

bacongobbler:
Good question. We probably want to disable CircleCI as soon as this is merged. Otherwise both Github Actions as well as CircleCI will be uploading competing uploads.

We can fix these up in follow-up PRs if need be.

.github/workflows/build-pr.yml
```diff
@@ -27,3 +31,31 @@ jobs:
...
+        with:
+          go-version: '1.16'
+
+      - name: Build Helm Binaries
+        run: |
+          make build-cross
+          make dist checksum VERSION="${VERSION}"
+
+      - name: Azure CLI Action
+        uses: Azure/cli@1.0.4
```
bacongobbler, line 58: probably want to be using the rolling `v1` branch instead of a specific version.

.github/workflows/release.yml
```diff
@@ -0,0 +1,47 @@
...
+          echo "::set-output name=VERSION::${{ github.event.inputs.version }}"
+        if: github.event == "workflow_dispatch"
+
+      - name: Build Helm Binaries
+        run: |
+          make build-cross
+          make dist checksum VERSION="${VERSION}"
+
+      - name: Azure CLI Action
+        uses: Azure/cli@1.0.4
```
bacongobbler, line 44: same here; probably want to use the rolling `v1` branch instead.

.github/workflows/release.yml
```diff
@@ -0,0 +1,47 @@
+name: release
+on:
+  create:
+    tags:
+      - v*
+
+  workflow_dispatch:
```
bacongobbler, line 7: why would we want to manually dispatch official releases?

.github/workflows/release.yml
```diff
@@ -0,0 +1,47 @@
...
+        uses: actions/setup-go@v2
+        with:
+          go-version: '1.16'
+
+      - name: Set Version (tag event)
+        run: |
+          echo "::set-output name=VERSION::${GITHUB_REF##*/}"
+        if: github.event == "create"
+
+      - name: Set Version (dispatch event)
```
bacongobbler, line 33: delete?

bacongobbler (changes requested):

.github/workflows/release.yml
tylerauerbeck: Removed.

.github/workflows/release.yml
tylerauerbeck: Removed.

.github/workflows/release.yml
tylerauerbeck: Fixed.

.github/workflows/build-pr.yml
tylerauerbeck: Fixed.

tylerauerbeck:
@bacongobbler This should be in good shape. Let me know if there is anything else you would like to see.

bacongobbler (approved):
This looks great. Code LGTM.

Note that this will require some coordination before we merge, as we will need to migrate our existing credentials over from CircleCI to Github Actions. I'll see if I can find some time to do this a few weeks prior to Helm 3.8's scheduled release so we have a bit of a buffer before the next release is cut.

bacongobbler:
I scheduled myself to merge this PR today. Now that I realize it's Black Friday... Probably best to avoid merging code that could affect the release assets... Especially on the busiest sales day of the year. ;)

I'll reschedule reviewing and merging this for Monday.

Curious why CircleCI is failing to report its status here too. No matter - it will soon be gone :)

mattfarina:
This has already been completed in another PR so I'm closing this one. Thank you for the work and to try to help us along with this.

PR #9993, opened 2021-08-02
Make Chart's IsRoot available to templates

### LINKED ISSUE #9991: Check if Chart is top level Chart or included as dependency
Hi,

is there a way to tell if a Chart is being installed/templated as the top level Chart or included as a dependency of another Chart?

I would like to have everything of the Chart in a named template so that it can be included n-times by a parent Chart dynamically and in case it is installed on its own I would like it to be rendered directly one time so that it can still be installed via helm install.

Thanks in advance.

Ka0o0:
This PR closes #9991 as it adds the information if the currently processed (templated) Chart is the root or not. This allows to have different behavior in case a Chart is installed on its own or included as part of another Chart. 

Templates can check this flag like so:

```
IsRoot: {{ .Chart.IsRoot }}
```

This is my first PR to Helm so please let me know if I'm missing something

**If applicable**:

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -345,8 +345,13 @@ func allTemplates(c *chart.Chart, vals chartutil.Values) map[string]renderable {
 // As it recurses, it also sets the values to be appropriate for the template
 // scope.
 func recAllTpls(c *chart.Chart, templates map[string]renderable, vals chartutil.Values) {
+	chartMetaData := struct {
+		chart.Metadata
+		IsRoot bool
+	}{*c.Metadata, c.IsRoot()}
+
 	next := map[string]interface{}{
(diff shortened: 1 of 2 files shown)


technosophos (changes requested):
I went back and forth on whether we should add this as metadata, and I think the answer is yes, this is the best place to put this.

So all this is missing is a test. Probably, you can just add a few lines in a test like this one:

https://github.com/helm/helm/blob/main/pkg/engine/engine_test.go#L318

Ka0o0:
Hi @technosophos ,
I've added a test case to cover both cases - IsRoot false and true.
Let me know what you think.

technosophos (approved):

mattfarina (approved):

PR #10007, opened 2021-08-06
Hooks with the same Weight but different Kind should not be compared by Name

### LINKED ISSUE #10006: Hook resource creation precedence not honored when weights are equal
Output of `helm version`:
```
version.BuildInfo{Version:"v3.6.0", GitCommit:"7f2df6467771a75f5646b7f12afb408590ed1755", GitTreeState:"dirty", GoVersion:"go1.16.4"}
```

Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.1", GitCommit:"5e58841cce77d4bc13713ad2b91fa0d961e69192", GitTreeState:"clean", BuildDate:"2021-05-12T14:11:29Z", GoVersion:"go1.16.3", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.1", GitCommit:"5e58841cce77d4bc13713ad2b91fa0d961e69192", GitTreeState:"clean", BuildDate:"2021-05-21T23:01:33Z", GoVersion:"go1.16.4", Compiler:"gc", Platform:"linux/amd64"}
```

Cloud Provider/Platform (AKS, GKE, Minikube etc.): KinD (but applicable to all)

This bug was observed when running `helm test` on a release with two test resources:
1) Pod
2) ConfigMap which the Pod depends on

The test would only create the Pod which in turn would fail due to the missing ConfigMap. `helm get hooks` on the release shows both the Pod and the ConfigMap.

The [delve](https://github.com/go-delve/delve) debugger helped identify [this line in *Configuration.execHook()](https://github.com/helm/helm/blob/efe2638f87b597403f34009e1029a6f0b44db8f2/pkg/action/hooks.go#L42):
```golang
// hooke are pre-ordered by kind, so keep order stable
sort.Stable(hookByWeight(executingHooks))
```

which in turn calls [hookByWeight.Less()](https://github.com/helm/helm/blob/efe2638f87b597403f34009e1029a6f0b44db8f2/pkg/action/hooks.go#L115-L120):
```golang
func (x hookByWeight) Less(i, j int) bool {
	if x[i].Weight == x[j].Weight {
		return x[i].Name < x[j].Name
	}
	return x[i].Weight < x[j].Weight
}
```

It looks like #7416 helpfully pre-sorts the hooks based on resource precedence. However the sort function compares and swaps the two different `Kind`s of resources in the `executingHooks` array based on their `Name` since the Pod and the ConfigMap have the same `Weight`. The `Name` compare and swap should probably only happen if the resources have the same `Weight` *and `Kind`*.

imheresamir:
closes #10006 

When hooks are being sorted, hooks resources with the same `Weight` but different `Kind`s should not be compared based on their `Name`s since that could result the loss of resource creation precedence.

**If applicable**:

### DIFF
--- a/pkg/action/hooks.go
+++ b/pkg/action/hooks.go
@@ -38,7 +38,7 @@ func (cfg *Configuration) execHook(rl *release.Release, hook release.HookEvent, 
 		}
 	}
 
-	// hooke are pre-ordered by kind, so keep order stable
+	// hooks are pre-ordered by kind, so keep order stable
 	sort.Stable(hookByWeight(executingHooks))
 
 	for _, h := range executingHooks {
@@ -113,7 +113,7 @@ type hookByWeight []*release.Hook
 func (x hookByWeight) Len() int      { return len(x) }
 func (x hookByWeight) Swap(i, j int) { x[i], x[j] = x[j], x[i] }
 func (x hookByWeight) Less(i, j int) bool {
-	if x[i].Weight == x[j].Weight {
+	if x[i].Weight == x[j].Weight && x[i].Kind == x[j].Kind {
 		return x[i].Name < x[j].Name
 	}
 	return x[i].Weight < x[j].Weight


cndoit18 (approved):
LGTM

bacongobbler (changes requested):
While I agree with this in principle, in practice many users have adjusted to the current behaviour. That is, changing the sorting behaviour can affect a current chart’s execution, and therefore introduce a breaking change.

Because of that, we will have to wait until Helm 4 before we can introduce this.

imheresamir:
Darn. Understandable, closing.

Issue #9942, opened 2021-07-14
Incorrect location for helm plugins installation.


zodiac182:
I have set the environment variables as following.
declare -x HELM_CACHE_HOME="/opt/storage/helm_home/helm3/.cache/"
declare -x HELM_CONFIG_HOME="/opt/storage/helm_home/helm3/.config/"
declare -x HELM_DATA_HOME="/opt/storage/helm_home/helm3/.local/share/"
declare -x HELM_PLUGINS="/opt/storage/helm_home/helm3/.local/share/**helm**/plugins/"

But when doing helm plugin install, new plugins will be installed into /opt/storage/helm_home/helm3/.local/share/plugins/.
And after new plugin have been installed, it could not be listed via "helm plugin list".
"helm plugin list" could show the right plugin list under $HELM_PLUGINS.
"helm plugin uninstall" could remove the right plugin under $HELM_PLUGINS.
But "helm plugin install" could not install plugins to the right location. Seems it would be installed into $HELM_DATA_HOME/plugin/, but not $HELM_PLUGINS.

Output of `helm version`:
v3.6.2
Output of `kubectl version`:
Version: v1.21.0
Platform: linux/amd64

yxxhero:
It's similar to https://github.com/helm/helm/issues/9814, It already had a PR. @zodiac182

mattfarina:
This is a duplicate of #9814. It is a bug and one I think we should fix in the near future.

PR #10008, opened 2021-08-07
Fix helm plugins behavior another way

### LINKED ISSUE #9814: Confusing behavior between HELM_PLUGIN and HELM_DATA_HOME

yxxhero:

fix: https://github.com/helm/helm/issues/9814 https://github.com/helm/helm/issues/9942

**If applicable**:

### DIFF
--- a/pkg/plugin/installer/base.go
+++ b/pkg/plugin/installer/base.go
@@ -18,16 +18,22 @@ package installer // import "helm.sh/helm/v3/pkg/plugin/installer"
 import (
 	"path/filepath"
 
-	"helm.sh/helm/v3/pkg/helmpath"
+	"helm.sh/helm/v3/pkg/cli"
 )
 
 type base struct {
 	// Source is the reference to a plugin
 	Source string
+	// PluginsDirectory is the directory where plugins are installed
+	PluginsDirectory string
 }
 
 func newBase(source string) base {
-	return base{source}
+	settings := cli.New()
+	return base{
+		Source:           source,
+		PluginsDirectory: settings.PluginsDirectory,
+	}
 }
 
 // Path is where the plugin will be installed.
@@ -35,5 +41,5 @@ func (b *base) Path() string {
 	if b.Source == "" {
 		return ""
 	}
-	return helmpath.DataPath("plugins", filepath.Base(b.Source))
+	return filepath.Join(b.PluginsDirectory, filepath.Base(b.Source))
 }
--- a/pkg/plugin/installer/base_test.go
+++ b/pkg/plugin/installer/base_test.go
@@ -1,0 +1,48 @@
+/*
+Copyright The Helm Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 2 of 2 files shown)


technosophos (approved):
This looks correct to me.

mattfarina (commented):
By using `cli.New()` is means other applications cannot use this package. It will them be coupled to helm environment variable names. The locations cannot be programmatically passed in.

I don't like this but I think it's a design issue we can't fix prior to Helm v4. I'll look a little more at it.

yxxhero:
@mattfarina thanks very much. if you need me do some thing. please tell me.

mattfarina:
I've added the topic of how we got here to the agenda for Thursdays meeting. If there are no other ideas in the short term, I think we can merge this.

yxxhero:
ok. Thanks very much.

mattfarina:
We talked about this is this past Helm developer call. We could not come up with a better way to fix this issue.

We also think we need to use a context or dependency injection in Helm v4 as a way to handle passing around some of the details in the future.

mattfarina (approved):

PR #10055, opened 2021-08-24
Respect original Chart.yaml when 'helm create -p'

### LINKED ISSUE #10057: Make `helm create -p` respect the Chart.yaml of the template chart
Hey!

When creating a helm chart from a template with `helm create -p` we saw that the Chart.yaml created does not include the dependencies key. We propose to create helm charts respecting the information in the Chart.yaml of the template.

ls-robert-holosynskyi:
As a helm user I would like for charts created with `helm create -p` to contain the dependencies described in Chart.yaml of the template.

This PR changes the way `helm create` works by not overriding the Chart.yaml read from the template directory.

Should we be concerned here with backward compatibility? 
If merged the charts that are created from a template will not look the same they used to.
Should this be hidden under a flag (--respect-chart-yaml)?

Closes #10057 

**If applicable**:

### DIFF
--- a/cmd/helm/create.go
+++ b/cmd/helm/create.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"fmt"
+	"helm.sh/helm/v3/pkg/chart"
 	"io"
 	"path/filepath"
 
@@ -50,9 +51,10 @@ will be overwritten, but other files will be left alone.
 `
 
 type createOptions struct {
-	starter    string // --starter
-	name       string
-	starterDir string
+	starter      string // --starter
+	name         string
+	starterDir   string
+	originalMeta bool
 }
 
 func newCreateCmd(out io.Writer) *cobra.Command {
@@ -80,6 +82,7 @@ func newCreateCmd(out io.Writer) *cobra.Command {
 	}
 
 	cmd.Flags().StringVarP(&o.starter, "starter", "p", "", "the name or absolute path to Helm starter scaffold")
+	cmd.Flags().BoolVar(&o.originalMeta, "original-meta", false, "use Chart.yaml from the starter template")
 	return cmd
 }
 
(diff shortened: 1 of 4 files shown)


bacongobbler:
PRs must be linked to an existing issue. If none exists, can you please open one? Thanks.

ls-robert-holosynskyi:
Added the issue =)

pkg/chartutil/create.go
```diff
@@ -511,13 +511,12 @@ spec:
 var Stderr io.Writer = os.Stderr
 
 // CreateFrom creates a new chart, but scaffolds it from the src chart.
-func CreateFrom(chartfile *chart.Metadata, dest, src string) error {
+func CreateFrom(chartname, dest, src string) error {
```
yxxhero, line 514: So unfortunately, this would constitute a breaking change as chartutil.CreateFrom is a publicly exposed function call.  
see this: https://pkg.go.dev/helm.sh/helm/v3@v3.6.3/pkg/chartutil#CreateFrom

yxxhero (commented):
Missing unittests

mattfarina (changes requested):

ls-robert-holosynskyi:
Hey! 
Finally got some time to work on this one.

Moved the implementation to the cmd folder, hid it under a flag, added a test.
WDYT?

ls-robert-holosynskyi:
Hey) @mattfarina and @yxxhero, is there a chance to have you take a second look at this PR?)

cmd/helm/create.go
```diff
@@ -96,18 +112,15 @@ func (o *createOptions) run(out io.Writer) error {
...
-			lstarter = o.starter
+	if o.originalMeta {
+		chartYamlPath := filepath.Join(lstarter, "Chart.yaml")
+		originalChartfile, err := chartutil.LoadChartfile(chartYamlPath)
+		if err != nil {
+			return err
 		}
-		return chartutil.CreateFrom(cfile, filepath.Dir(o.name), lstarter)
+		originalChartfile.Name = chartname
+		cfile = originalChartfile
```
zonggen, line 122: Question: wouldn't this also preserves other information other than dependencies from `Chart.yaml`? In the PR description the main goal seems like to preserve dependencies, but I could be wrong.

ls-robert-holosynskyi: yes indeed, all the other information will be preserved)

ls-robert-holosynskyi:
Hey, how do we get this merged?)

ls-robert-holosynskyi:
Closing because of no updates) Our workaround is do instruct users to copy over the dependencies to the chart created from a template

PR #7477, opened 2020-01-28
Added export-values handling to dependencies

withnale:
NB: This is a reworked implementation of the functionality found in PR #3243 

---

See issue #3242 

There is a need for an `export-values` method when working with requirements.

The attached PR approaches this by mimicking the `import-values` functionality but in the other direction - allowing you to specify the tree in the parent chart you wish to export into the values of the child chart. It also allows you to _ringfence_ settings to a duplicate (aliased) subchart, as can be seen in the example below:

```
dependencies:
      - name: subchart1
        repository: http://localhost:10191
        version: 0.1.0
        tags:
          - front-end
          - subchart1
        export-values:
          - child: release
            parent: release.old

      - name: subchart1
        repository: http://localhost:10191
        version: 0.1.0
        export-values:
          - child: release
            parent: release.new
        alias: subchart1a
```
Obviously, this simple use case can be achieved through other methods - it's when there are many requirements using a mix & match of various parent values that this change become effective.

NB: We should really rename these methods but not sure what would be a good plan. Should it go from `processDependencyImportValues` to `processDependencyImportExportValues`, something else, or should we just leave them with their existing names?

PR #3243, opened 2017-12-09
Added support for export-values in requirements

withnale:
See issue #3242 

There is a need for an `export-values` method when working with requirements.

The attached PR approaches this by mimicking the `import-values` functionality but in the other direction - allowing you to specify the tree in the parent chart you wish to export into the values of the child chart.It also allows you to _ringfence_ settings to a duplicate (aliased) subchart, as can be seen in the example below:

```
dependencies:
      - name: subchart1
        repository: http://localhost:10191
        version: 0.1.0
        tags:
          - front-end
          - subchart1
        export-values:
          - child: release
            parent: release.old

      - name: subchart1
        repository: http://localhost:10191
        version: 0.1.0
        export-values:
          - child: release
            parent: release.new
        alias: subchart1a
```
Obviously, this simple use case can be achieved through other methods - it's when there are many requirements using a mix & match of various parent values that this change become effective.

Issue #3242, opened 2017-12-09
Add 'export-values' to requirements


withnale:
There is a need for an `export-values` method when working with requirements.

At present there is already a mechanism for _pulling_ values from subcharts into the parent chart using the 'import-values' method defined in [docs/chart.md](https://github.com/kubernetes/helm/blob/master/docs/charts.md#importing-child-values-via-requirementsyaml). However, there is no equivalent method for encapsulating parts of the values tree when wishing to provide specific trees to subcharts.

We use parent charts often with many subcharts, but often these subcharts are used by many parent charts (M:N). In this scenario there is often a set of values we wish to pass to various sets of charts.

The methods we have right now allow:

1. We can put the value explicitly in the subcharts tree i.e.  `subchart1.myvalue` - however, this doesn't scale too well, since we want to define this value once in the parent chart but if multiple subcharts require this value, we need to explicitly set for each subchart in the parents values. i.e. `subchart1.myvalue`, `subchart2.myvalue`, etc. When we spin up a release, we want to specify the minimal set of value overrides (using `-f overrides.yaml`) as possible and not have to specify this for each subchart - so this isn't really viable.
1. Alternatively, we could use the `globals` namespace within the values tree. However, this isn't ideal either, since this tree is duplicated in all subcharts the values file quickly gets massive, and there are benefits in having the trees in use by each chart encapsulated specifically in the requirements. Managing a global namespace isn't ideal and can easily generate unforeseen side effects.

The attached PR approaches this by mimicking the `import-values` functionality but in the other direction - allowing you to specify the tree in the parent chart you wish to export into the values of the child chart.It also allows you to _ringfence_ settings to a duplicate (aliased) subchart, as can be seen in the example below:

```
dependencies:
      - name: subchart1
        repository: http://localhost:10191
        version: 0.1.0
        export-values:
          - child: release
            parent: release.old

      - name: subchart1
        repository: http://localhost:10191
        version: 0.1.0
        export-values:
          - child: release
            parent: release.new
        alias: subchart1a
```
Obviously, this simple use case can be achieved through other methods - it's when there are many requirements using a mix & match of various parent values that this change become effective.

vedit:
I believe this list should be created automatically during a dependency installation with cli

withnale:
Maintainers - any update on the following? We have some complex scenarios we are trying to map out regarding parent charts and subcharts and this would really help simplify/encapsulate the 'interfaces' between charts

jascott1:
[...]

PR #10059, opened 2021-08-24
feat: implement export-values

ilya-lesikov:
This PR implements `export-values` directive which is similar to the [import-values](https://helm.sh/docs/topics/charts/#importing-child-values-via-dependencies) directive, but instead of importing the values of the subchart into the current chart it exports the values of the current chart into subchart.

Initially the change was discussed here #3242. There were several attempts to implement this: #7477, #3243.

## Usage:
In the full form of `export-values` you specify which value tree to export (with `parent:`) and where it should be exported in the subchart (with `child:`).

In the short form of `export-values` you only specify a relative path inside of the `exports:` key of the parent chart values: specified values will be exported to the root of the subchart values (same as with the `import-values`). This allows to export/import values defined in `exports:` both to the parent chart or to the child chart depending on the directive used (`import-values` or `export-values`).

Unlike `import-values`, `export-values` can export not only the trees of values (maps), but other types of values too (strings, ints, etc).

Example:
```yaml
# values.yaml
parent-map:
  key1: parent-value1
  key2: parent-value2
  
exports:
  parent-map:
    key3: parent-value3
  
subchart1:
  subchart-map:
    key1: subchart-value1
```

```yaml
# Chart.yaml:
dependencies:
  - name: subchart1
    version: 1.0.0
    export-values:
    - parent: "parent-map"
      child: "subchart-map"
    - "parent-map.key3"
```

```yaml      
# charts/subchart1/dump-values-in-subchart.yaml
dump-values-in-subchart: {{ .Values | toYaml | nindent 2 }}
```

Result of `helm template`:
```yaml
# charts/subchart1/dump-values-in-subchart.yaml
dump-values-in-subchart:
  subchart-map:
    key1: parent-value1
    key2: parent-value2
    key3: parent-value3
```

**If applicable**:

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -198,7 +198,7 @@ func (i *Install) RunWithContext(ctx context.Context, chrt *chart.Chart, vals ma
 		return nil, err
 	}
 
-	if err := chartutil.ProcessDependencies(chrt, vals); err != nil {
+	if err := chartutil.ProcessDependencies(chrt, &vals); err != nil {
 		return nil, err
 	}
 
@@ -349,7 +349,7 @@ func (i *Install) RunWithContext(ctx context.Context, chrt *chart.Chart, vals ma
 	go i.performInstall(rChan, rel, toBeAdopted, resources)
 	go i.handleContext(ctx, rChan, doneChan, rel)
 	result := <-rChan
-	//start preformInstall go routine
+	// start preformInstall go routine
 	return result.r, result.e
 }
 
--- a/pkg/action/upgrade.go
+++ b/pkg/action/upgrade.go
@@ -207,7 +207,7 @@ func (u *Upgrade) prepareUpgrade(name string, chart *chart.Chart, vals map[strin
 		return nil, nil, err
 	}
 
-	if err := chartutil.ProcessDependencies(chart, vals); err != nil {
+	if err := chartutil.ProcessDependencies(chart, &vals); err != nil {
 		return nil, nil, err
 	}
 
--- a/pkg/chart/dependency.go
+++ b/pkg/chart/dependency.go
@@ -45,6 +45,9 @@ type Dependency struct {
 	// ImportValues holds the mapping of source values to parent key to be imported. Each item can be a
 	// string or pair of child/parent sublist items.
 	ImportValues []interface{} `json:"import-values,omitempty"`
+	// ExportValues holds the mapping of parent values to child key to be exported. Each item can be a
+	// string or pair of parent/child sublist items.
+	ExportValues []interface{} `json:"export-values,omitempty"`
 	// Alias usable alias to be used for the chart
 	Alias string `json:"alias,omitempty"`
 }
--- a/pkg/chartutil/dependencies.go
+++ b/pkg/chartutil/dependencies.go
@@ -217,27 +217,57 @@ func set(path []string, data map[string]interface{}) map[string]interface{} {
 	return cur
 }
 
-// processImportExportValues merges values between child and parent based on the chart's dependencies' ImportValues
-// and ExportValues fields.
-func processImportExportValues(c *chart.Chart) error {
+// processImportValues merges Values from dependent charts to the current chart based on ImportValues field.
+func processImportValues(c *chart.Chart) error {
 	if c.Metadata.Dependencies == nil {
 		return nil
 	}
-	// combine chart values and empty config to get Values
+
 	cvals, err := CoalesceValues(c, nil)
 	if err != nil {
 		return err
 	}
-	b := make(map[string]interface{})
-	// For each dependency export values from parent or import values from child if export-values or import-values
-	// specified.
+
+	importedValuesMap := make(map[string]interface{})
 	for _, r := range c.Metadata.Dependencies {
-		b = CoalesceTables(b, getExportedValues(c, r, cvals))
-		b = CoalesceTables(b, getImportedValues(r, cvals))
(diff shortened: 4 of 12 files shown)


bacongobbler:
Hi @ilya-lesikov. I'm going to re-express the same concerns I had with the last PR (#7477) that there was not enough supporting documentation provided to better understand

- how it works
- why it's useful
- how it addresses backwards compatibility
- why existing features can't handle the same use case
- how are we supposed to teach this to new users
- etc.

With that said, I believe this proposal should go through [the HIP process](https://github.com/helm/helm/blob/main/CONTRIBUTING.md#proposing-an-idea) BEFORE any further code review. The process helps answer some basic questions like "how does this address backwards compatibility?" and "how is this feature expected to be useful?" which I haven't seen answered in either #3242, #7477, or this PR. And I think that would help alleviate a lot of the concerns I expressed earlier.

Let me know if you have any questions about the process. Thanks!

ilya-lesikov:
I'll do the docs, although I thought I might get some hints about the overall design before writing them. The usage could've been found in tests and the first post.
I'll do the HIP/docs soon then.

bacongobbler:
> The usage could've been found in tests and the first post.

While I understand the sentiment, we (the maintainers) have a very limited amount of time we can spend each week to review code. Documenting your feature means we can spend more time reviewing your code. If I have to read your unit tests and attempt to make assumptions about how your feature is supposed to work, I'm going to spend my time looking at the other 150+ PRs in the queue.

> although I thought I might get some hints about the overall design before writing them

Honestly, that's up to you. This is open source - you the author have the authority to choose how it works. Though that also comes with the burden to document and teach others how it works. Especially if you want to get it merged :)

MajorBreakfast:
Use case: forwarding of `imagePullSecret` name value (string) to various subcharts.
Many public helm charts have expose setting the image pull secret differently in their values.yaml, e.g. `imagePullSecret: string`, `imagePullSecrets: string[]` or `imagePullSecrets: { name: string}[]` (typescript type syntax). If this feature was implemented I think it would let my write a parent chart that accepts the pull secret value top level and forwards it to the child charts

loewenstein:
May I politely ask about the state of this PR? Has a HIP been created already? Any idea as to when this might be available in a released version of Helm (like are we talking weeks, months, years or even "we don't know if this gets merged at all")?

If I've followed the breadcrumbs correctly, this and similar though more generic proposals have been asked for and discusses for years.

This approach seems to follow a compatibility oriented approach while fixing a from my point of view significant deficiency with regards to compositionality of Helm charts / abstraction around parent and child charts.

It would be really great to get `export-mapping` and if there is anything that would help to drive this forward I'd be interested to hear.

ilya-lesikov:
Hello @loewenstein
Frankly, looking at the HIP repo I was not sure how much additional time will all of this take and will it have any chance of landing in upstream. Also, it was not immediately obvious how much value this feature adds on top of passing these same values via `--set` cli flags (if values come from an external system) or using yaml aliases in values.yaml (if values are already present in values.yaml). Surely, this might look better than passing a handful of `--set` flags on each invocation, and yaml aliases have their fair share of limitations (e.g. inability to reference anchors from another values file). Is this enough to get it merged is hard to tell. Is this what you need? Maybe you can come up with some additional uses of this feature?

In werf this feature was really useful since we inject in Helm some values at runtime and the user does not have direct control of these values. Because of this, we needed something truly dynamic to be able to actually _map_ these values from one place to another. For us it worked well.

loewenstein:
FYI we've created a HIP for export-values. https://github.com/helm/community/pull/242

loewenstein:
I think we did figure out that this PR would not yet readily provide an implementation though.

jdolitsky (dismissed):
Lets resolve the issue with old PR but otherwise lgtm

loewenstein:
> Lets resolve the issue with old PR but otherwise lgtm

I understood from @c0d1ngm0nk3y and others that this PR would not be a full implementation of the HIP we provided. I’d rather put this on hold (at least until they had a chance to comment). We might just provide a separate PR with a proper implementation - potentially based of of this proposal.

loewenstein:
cc @phil9909

jdolitsky:
@loewenstein - is this a partial implementation? Would you be able to push your changes for HIP 16 directly onto this PR?

cc @ilya-lesikov

loewenstein:
We'll discuss how to continue tomorrow 9 CET. We are not related to @ilya-lesikov at all, just need the feature, found the existing approaches and decided to pick up the request for a HIP as prerequisite for am implementation.

jdolitsky (commented):
revoking approve until the HIP is fully implemented

bajoben:
What is the state of play here?

joejulian:
@ilya-lesikov Can you rebase this?

dtomasi:
Is there any progress? This would be sooo useful and I was looking forward to use it since 2021 ...

sparecycles:
Hi community, I don't know if this is a good idea or how compatible it is across helm versions, but the `import-values` mechanism can be used to implement `export-values` in limited cases.

For example if you want to export some values from your project's `config.info` into a `configuration` subchart at `configuration.info`, just create an empty chart and add it as a dependency.

```yaml
dependencies:
  - name: my-empty-chart
    alias: config
    import-values:
      - child: info
        parent: configuration.info
```

This will copy `config.info` to `configuration.info` ... and since the parent chart's values should overwrite values in the child before `import-values`, this should produce a reasonable result in most cases? 🤞 

Test test test!

PR #10064, opened 2021-08-25
Fixing issue where helm show all fails to show crds when no readme

### LINKED ISSUE #10058: helm show all fails when there is no readme
If there is no readme in a chart then `helm show all` will fail to display CRDs

This is because a lack of a readme causes the function to return. See...

https://github.com/helm/helm/blob/9443a7ead4513e50518b2c12cb81534143ef8090/pkg/action/show.go#L115-L118

Output of `helm version`:
tip of main on date issue filed

Output of `kubectl version`:
N/A

N/A

mattfarina:
Closes #10058

When a chart has no README.md file but has CRDs and you try to use `helm show all` on it.... the CRDs will not be displayed.

More detail can be found on #10058

**If applicable**:

### DIFF
--- a/pkg/action/show.go
+++ b/pkg/action/show.go
@@ -109,14 +109,13 @@ func (s *Show) Run(chartpath string) (string, error) {
 	}
 
 	if s.OutputFormat == ShowReadme || s.OutputFormat == ShowAll {
-		if s.OutputFormat == ShowAll {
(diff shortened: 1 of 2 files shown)


hickeyma (approved):
LGTM, thanks for fixing this @mattfarina

Issue #1892, opened 2017-01-31
Feature request: Allow inheriting files from parent templates


andrewhowdencom:
At the moment, children templates can inherit values in the parent `Values.yml` file if that file includes the value, prefixed by the chart name. For example, 

```
mysql:
    mysqlRootPassword: herapderpa # This overrides the mysqlRootPassword on the top level of its Values.yml
```

There is no such inheritance allowed for files. It appears that values are being explicitly passed down the stack in the following snippet: 

https://github.com/kubernetes/helm/blob/c8af660b181daf4000aca87e19e583fd1abec0b3/pkg/engine/engine.go#L217-L222

It would be good for files to be passed down the stack also, though there is no convention for doing as such. One suggestion might be to create a folder with the name of the child chart, and inherit the files supplied in that folder. For example, 

```
mysql/
    mysql.conf
```

In the case above, the mysql.conf file would override the mysql.conf file in the root of the mysql chart, stored in charts/ (note: don't go looking, I made that up)

Thaanks!

technosophos:
I think we could do this, provided we have a good convention. Maybe along side `charts/` we have something like an `overrides/` directory that could contain information like this. It might get tricky. For example, how do I override `mychart/charts/wordpress/charts/mysql/my.cnf` but not `mychart/charts/drupal/charts/mysql/my.cnf`?

andrewhowdencom:
If I was to choose an API, I would try and model it after the inheritance behaviour of .Values.yml. So, instead of 

```
overrides/my.conf
```

You would have

```
overrides/drupal/mysql/${my.conf
overrides/wordpress/mysql/my.conf
```

So, the format would be something like 

```
overrides/${CHART_A}/${CHART...N}/${PATH_TO_FILE}
```

If I understand correctly, .Values.yml inherits along a simliar syntax

```
---
# child
foo: bar
```

```
# parent
child:
  foo: baz
```

jascott1:
@andrewhowdencom Closing this issue due lack of activity for 6 months. Please feel free to re-open  or create a new proposal.

andrewhowdencom:
Hola @jascott1; I think this is still valid -- A pattern of embedding files in the yaml has started to emerge, which is not so elegant: 

https://github.com/kubernetes/charts/blob/master/stable/prometheus/values.yaml#L545

I just don't have the skills to implement it (:

Cannot reopen.

jascott1:
Re-opening based on OP request.

johnny:
[...]

PR #10077, opened 2021-08-28
Include external files

### LINKED ISSUE #3276: Allow helm to add external files to chart deployment

itaispiegel:
## **What this PR does / why we need it**:
This PR adds the flag `--include-path` which allows you to include external files in the install, upgrade and template commands.

### Example
Given the configmap:
```yaml
---
apiVersion: v1
kind: ConfigMap
metadata:
    name: {{ .Release.Name }}-test
data: {{ (.Files.Glob "/tmp/externals/*").AsConfig | nindent 2 }}
```
And the directory structure:
```
├── /tmp/externals
│   ├── external1.txt
│   └── external2.txt
```
Running `helm template test_chart --include-path /tmp/externals` will generate the manifest:
```yaml
---
# Source: test_chart/templates/test.configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
    name: RELEASE-NAME-test
data:
  external1.txt: |
    External file 1
  external2.txt: |
    External file 2
```

If the `--include-path` is not passed, the output will be:
```
Error: template: test_chart/templates/test.configmap.yaml:6:41: executing "test_chart/templates/test.configmap.yaml" at <(.Files.Glob "/tmp/externals/*").AsConfig>: error calling AsConfig: must pass files

```

This PR continues the work that @hunsche started in his PR: https://github.com/helm/helm/pull/8841.
I tried contacting him to discuss his work, but unfortunately, I couldn't reach him so I implemented this myself after discussing this at the Helm dev call which took place on August 19th.

This is my first contribution to Helm, and my first time working with Go. I'd be glad to get feedback for this regarding code style, best practices, and design.

close #3276 
close #8227

Related documentation PR: https://github.com/helm/helm-www/pull/1180

Thanks in advance! :)


mattfarina:
There's feedback on #8841 that needs to be taken into account here.

itaispiegel:
@mattfarina Can you please address what feedback?

pkg/engine/files.go
```diff
@@ -46,11 +47,11 @@ func newFiles(from []*chart.File) files {
 //
 // This is intended to be accessed from within a template, so a missed key returns
 // an empty []byte.
-func (f files) GetBytes(name string) []byte {
+func (f files) GetBytes(name string) ([]byte, error) {
```
mattfarina, line 50: `GetBytes` is a public function in the SDK. It's signature cannot be changed until we get to Helm v4. We only do backwards compatible feature additions. The same applied for all the functions in Helm.

Can you please revert any sig changes.

mattfarina (changes requested):

pkg/engine/files.go
itaispiegel: I applied this change because the current behaviour is that if you try to access a file that isn't included - the engine will treat it as empty. This behaviour is a bit unexpected and can make it hard to understand why the chart isn't rendered as expected.
It's more important now when this behaviour can happen because the user didn't include the file he wanted to include.

Is there perhaps a better way we can deal with this without breaking the API?

itaispiegel:
@mattfarina Btw, I thought about it more and I think I should revert this commit: https://github.com/helm/helm/pull/10077/commits/d1f9bcb0fbf00bfbb76c0c1e844f87689d284d41

My initial thought was that I want to be able to access the file in the chart by its path and not by an alias.
On the other hand, this doesn't allow us to treat relative paths, because the user can pass `--include-path ../external.txt`, but then if he references `/external.txt` in the chart, it won't recognize this file.
If I revert this commit I can do: `--include-path ../external.txt:externals/external.txt` and then reference `externals/external.txt` in the chart.

What do you think about this?

pkg/engine/files.go
itaispiegel: Maybe it would be enough to just log a warning message for now?
Do you think though that we can apply this change in Helm 4.0?

mattfarina:
@itaispiegel Using a colon as the separator won't work on Windows. Helm is used on Windows, too.

As for the idea of having a key/value pair, that requires the Helm user (chart user) to know about the internal workings of the chart. For many that is not the case. The consumer of a package (just like someone using apt to install mariadb) doesn't usually know the app that's being installed or how it works. I'm not sure the UX is a great experience for that setup.

Am I missing something?

itaispiegel:
> @itaispiegel Using a colon as the separator won't work on Windows. Helm is used on Windows, too.

I'm not sure this is an issue. For example, you can see that in Docker they use a colon as a delimiter for the `-v` flag: 
https://docs.docker.com/engine/reference/commandline/run/#mount-volume--v---read-only

> As for the idea of having a key/value pair, that requires the Helm user (chart user) to know about the internal workings of the chart. For many that is not the case. The consumer of a package (just like someone using apt to install mariadb) doesn't usually know the app that's being installed or how it works. I'm not sure the UX is a great experience for that setup.
> 
> Am I missing something?

How is this different than passing values to a chart during the installation? For example, if I want to install the MariaDB chart, I'd need to go over the README file, understand the internal values, and set them accordingly.
I don't think there's really a difference with external files. If a chart requires them, then its documentation should explain how to pass them correctly.

That's also why I think that changing the behavior to fail if the file isn't included (instead of considering it empty) is a necessary behavior. Because then if you include the wrong file, the chart will fail to install, instead of it being installed in a weird state. I don't think though that it should block this PR.

itaispiegel:
Hey guys, I'm going on vacation so I won't be able to continue working on this until I come back in October.
I'm very motivated to get the work done, so I'll continue this when I come back.
If anyone wants to contribute to this PR by then, feel free to contact me by email: itai.spiegel@gmail.com :)

pkg/engine/files.go
bacongobbler: You can introduce a new function. But you cannot change existing function signatures.

bacongobbler:
> @mattfarina Can you please address what feedback?

I believe @mattfarina is referring to these comments:

https://github.com/helm/helm/pull/8841#pullrequestreview-661474703
https://github.com/helm/helm/pull/8841#issuecomment-748243016

We discussed them in a developer call, but I don't believe there are any concrete test cases covering this scenario with the `--include-file` flag.

bacongobbler:
Why was `--include-file` removed?

pkg/chart/loader/local.go
```diff
@@ -0,0 +1,75 @@
...
+
+import (
+	"errors"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+// ExpandLocalPath expands a local file, dir or glob path to a list of files
+func ExpandLocalPath(path string) ([]string, error) {
```
bacongobbler, line 24: `Local` should be `File` - "local" is a little too generic in this context. A local URL path like `http://localhost/my-chart.tgz` is still considered by many to be a "local" path.

pkg/engine/files.go
```diff
@@ -59,8 +60,13 @@ func (f files) GetBytes(name string) []byte {
 // template.
 //
 //	{{.Files.Get "foo"}}
-func (f files) Get(name string) string {
-	return string(f.GetBytes(name))
+func (f files) Get(name string) (string, error) {
```
bacongobbler, line 63: same issue here RE: breaking compatibility.

pkg/engine/files.go
```diff
@@ -101,9 +107,9 @@ func (f files) Glob(pattern string) files {
 //
 //   data:
 // {{ .Files.Glob("config/**").AsConfig() | indent 4 }}
-func (f files) AsConfig() string {
-	if f == nil {
-		return ""
+func (f files) AsConfig() (string, error) {
```
bacongobbler, line 110: same issue here RE: breaking compatibility.

pkg/engine/files.go
```diff
@@ -130,9 +136,9 @@ func (f files) AsConfig() string {
 //
 //   data:
 // {{ .Files.Glob("secrets/*").AsSecrets() }}
-func (f files) AsSecrets() string {
-	if f == nil {
-		return ""
+func (f files) AsSecrets() (string, error) {
```
bacongobbler, line 139: same issue here RE: breaking compatibility.

pkg/engine/files.go
```diff
@@ -151,10 +157,10 @@ func (f files) AsSecrets() string {
 //
 // {{ range .Files.Lines "foo/bar.html" }}
 // {{ . }}{{ end }}
-func (f files) Lines(path string) []string {
+func (f files) Lines(path string) ([]string, error) {
```
bacongobbler, line 160: same issue here RE: breaking compatibility.

bacongobbler (dismissed):

R-omk:
You need not forget about `helm get all`.

R-omk:
What about named aliases for each individual directory?  With this approach, the chart does not have to rely on absolute paths.  And when some existing release is updated, it will be possible to individually replace the named directory.  It's kind of like a docker  mount named volumes only in the opposite direction `vol_name:/path/to/dir`.
Also, my proposal may be related to https://github.com/helm/helm/issues/1892#issuecomment-939436851

anamarija:
Hi everyone, I would like to contribute to this feature. 
Can I pick up the work and continue the implementation :) 

My plan would be to:
 - rebase the current work to the current master
 - extend the ```loadExternalPaths``` implementation to support named and default aliases like 

``` 
--include-path vol_name1:/path/to/dir1 --include-path vol_name2:/path/to/dir2
```
``` 
 ├── /path/to/dir1
│   ├── external1.txt
│   └── external2.txt
```
```
├── /path/to/dir2
│   ├── external1.txt
│   └── external2.txt 
```
```
---
apiVersion: v1
kind: ConfigMap
metadata:
    name: {{ .Release.Name }}-test
data: {{ (.Files.Get "vol_name1/external1.txt") | nindent 2 }}
```
```
---
apiVersion: v1
kind: ConfigMap
metadata:
    name: {{ .Release.Name }}-test
data: {{ (.Files.Glob "vol_name1/*").AsConfig | nindent 2 }}
```

example with multiple ```--include-path``` without named alias. Helm would define a default alias like 0 and 1.

```
--include-path /path/to/dir1 --include-path /path/to/dir2
```
```
---
apiVersion: v1
kind: ConfigMap
metadata:
    name: {{ .Release.Name }}-test
data: {{ (.Files.Get "0/external1.txt") | nindent 2 }}
```

Also I thought about not only including files from file directories but also maybe github repostiories. Does it make sense to incorporate such a functionality as part of this PR, to be able to include files from github repositories?

itaispiegel:
Be my guest :)
Regarding the option to include files from Github repositories - why would you want that?
Remember you can always clone the repository locally and use a symlink. I'm not sure about this addition, so we should wait and see the maintainers' opinion on this.

anamarija:
To give more context :)

From my experience working on different enterprise projects, it is common to have different gh repositories for application and ops. 
A common use case in a DevOps automated pipeline could be that the helm chart is part of an ops repository that wants to get the application configuration present in a different application gh repository.
It would make sense to have a capability for helm to automatically get the application configuration and make it available to the chart. Additionally, it would play well with tools like ArgoCD and in an environment where there is a huge amount of services that have different application configurations. 

But its just a proposition :)

dbgjerez:
Hello, 
Somebody has any update about this?
Thanks a lot!

pkg/chart/loader/local.go
itaispiegel: > Hello, Somebody has any update about this? Like @anamarija said, we're using Argo CD too, and it's a common approach getting the configuration from another repository. Thanks a lot!

You're right. I also really want this feature released!
I'm sorry for delaying this for so long. I went on a trip after working on this, and then when I came back I couldn't find the time to continue this. I'll try to put some effort on this in the next days so we can get this done!
You're welcome to join me :)

itaispiegel:
> Hello, Somebody has any update about this? Like @anamarija said, we're using Argo CD too, and it's a common approach getting the configuration from another repository. Thanks a lot!

You're right. I also really want this feature released!
I'm sorry for delaying this for so long. I went on a trip after working on this, and then when I came back I couldn't find the time to continue this. I'll try to put some effort on this in the next days so we can get this done!
You're welcome to join me :)

itaispiegel:
@bacongobbler @mattfarina Hey, I finally fixed the comments on the PR!
I'd be glad if you could review this. To be honest, it's been a while since I've been in contact with you about this.
If you think that there are major gaps in this PR, please make contact with me, and I'd be glad to further discuss this :)

dbgjerez:
> > Hello, Somebody has any update about this? Like @anamarija said, we're using Argo CD too, and it's a common approach getting the configuration from another repository. Thanks a lot!
> 
> You're right. I also really want this feature released! I'm sorry for delaying this for so long. I went on a trip after working on this, and then when I came back I couldn't find the time to continue this. I'll try to put some effort on this in the next days so we can get this done! You're welcome to join me :)

Hi, It's alright dude, thanks a lot for your time. 
I can help you with whatever you need, I have revised all the changes and I think it's correct. It is just what we need :)

itaispiegel:
> > > Hello, Somebody has any update about this? Like @anamarija said, we're using Argo CD too, and it's a common approach getting the configuration from another repository. Thanks a lot!
> > 
> > 
> > You're right. I also really want this feature released! I'm sorry for delaying this for so long. I went on a trip after working on this, and then when I came back I couldn't find the time to continue this. I'll try to put some effort on this in the next days so we can get this done! You're welcome to join me :)
> 
> Hi, It's alright dude, thanks a lot for your time. I can help you with whatever you need, I have revised all the changes and I think it's correct. It is just what we need :)

Great. So let's wait until one of the maintainers updates us how to proceed.

itaispiegel:
> any news on this here?

Hey, unfortunately I couldn't get enough attention on this so I didn't get a CR...

mktbec:
that is very sad :) - we also need this very much. I will keep my fingers crossed :)

itaispiegel:
> that is very sad :) - we also need this very much. I will keep my fingers crossed :)

I agree. Great opportunity for one of the maintainers to try help us here :)

PauloFerreira25:
@joejulian 

Please, it is not possible that for 3 consecutive times, the same code will be passed from pull request to pull request.
The basics have already been coded, making the feature work.

This issue has been opened here:
https://github.com/helm/helm/issues/3276 - 19 Dec 2017

Blocked and played to closed, here:
https://github.com/helm/helm/pull/8227 - 29 May 2020

Is here:
https://github.com/helm/helm/pull/8841 - 2 Oct 2020
https://github.com/helm/helm/pull/8840 - 2 Oct 2020

I live in Brazil, not even here bureaucracy takes 2 years for a stamp.

Or is it project policy to take what you don't want to do and keep closing pull requests?

joejulian:
Bureaucrats get paid to do their job. I'm doing this as a kindness to the community. I'd be happy to have help. Care to volunteer?

PauloFerreira25:
I'm already helping before you even ask.

Now, do you want help or do you want an employee?

Example is this thread. It's ready, it's tested, it's functional.

What else do you need?

If you tell me, what's missing in this thread for it to be coded, I'll do what's missing.

Or do you want an employee who doesn't have permission to run the project and just cleans up Issues and Pulls?

If it's something coding, programming, I'll gladly help.

joejulian:
What do I need? I need a couple core maintainers to review it. I only have permissions to do triage.

yxxhero:
> 

@joejulian we are the same.

PauloFerreira25:
> 
Unfortunately I don't have that power either.

joejulian:
It frustrates me, too, Paulo. That's why I closed the dead one and added a milestone to this one. I'm hoping to get some eyes on it.

itaispiegel:
Hey, any news from the maintainers?

joejulian:
@itaispiegel I've added it to the January milestone. It's either going to get reviewed or bumped to the next milestone. Either way, you'll get some sort of feedback from them by then. :crossed_fingers:

MaesterZ:
In some cases, its great to create helm charts to have a whole software stack setup the right way and have the end user bring his own configuration (BYOC maybe? 😅), a bit like the BYOL model. Would allow to bring a "flat" file, potentially templated out from another software. Kubernetes is already complex enough for most people, if you add Helm, the step to get into the ecosystem is quite high IMO.

ikegentz:
> To give more context :)
> 
> From my experience working on different enterprise projects, it is common to have different gh repositories for application and ops. A common use case in a DevOps automated pipeline could be that the helm chart is part of an ops repository that wants to get the application configuration present in a different application gh repository. It would make sense to have a capability for helm to automatically get the application configuration and make it available to the chart. Additionally, it would play well with tools like ArgoCD and in an environment where there is a huge amount of services that have different application configurations.
> 
> But its just a proposition :)

Sounds like a neat idea, especially for GitOps workflows with config stored in other directories. To play devil's advocate however, it seems like this could introduce an expectation by Helm users to maintain other integrations, like GitLab repositories for example. Not the worst thing, but it seems beyond the scope of this particular PR. Maybe a `--from-repo` flag, separate from this, would serve better?

bricss:
LFTI 👀 this thing is gonna help a lot with DevOps & GitOps especially within `monorepos`, as well as with multi-culti aka regional deployments 🚢 
Coz, atm it require plenty of shenanigans 🕹️to achieve those goals 🎯

rediphone:
@bacongobbler @mattfarina 
is this something that can be progressed? it looks like everything is ready here? this would be super useful and it seems to just be waiting for something to happen...... thanks!

invertigo:
> LFTI 👀 this thing is gonna help a lot with DevOps & GitOps. Especially within `monorepos`, as well as with multi-culti aka regional deployments 🚢 Coz, atm it require plenty of shenanigans 🕹️to achieve those goals 🎯

For what it's worth, my stopgap solution has been to dynamically symlink the external files directory into the helm chart inside a small shell wrapper around the `helm` binary.

mattfarina (changes requested):
I see that a lot of work has been put into this and I appreciate the care and dedication. I'm sorry I didn't get back to this sooner. GitHub notifications have become mostly useless to be as I'm attached to too many things. So, I missed this for awhile.

I have two major issues with this and I think both can be overcome.

First, Helm is both a client and an SDK. There are many applications, such as controllers and operators, that use the SDK. To support both we need to keep the separation of concerns in mind. For example, the Helm client loads a chart and passes the loaded chart to the action. See the cmd/helm/install.go file for an example. This allows the application to load the chart as it needs to and pass it to the install/upgrade/etc logic to operate on. In the same way an operator could load a chart through a different means and pass it to the logic. _This means we can't assume the SDK should be accessing the local file system._

The local files need to be loaded by the application (i.e. `cmd/helm`) and then passed into the business logic. This way something like a controller could load files through a different means and pass them to the business logic.

The idea is that we need to keep applications in mind that are not the Helm CLI.

Second, the original request appeared to ask for specific files to be passed in. This enables providing a directory and the complete directory tree is passed in. Three things come to mind with that....

1. Why do a directly instead of the specific files? I'm asking for a use case for this as it's not been shared before. Please think of the application installer first.
2. Helm needs to internally store all of these files to deal with things like rollback. These are stored, by default, on a secret. This means there are size implications. Being lazy with paths and loading things one does not need will cause size issues that need support.
3. There is a security implication here. One could create a copy and paste prompt and point to the external files of `/etc`. Many people may blindly copy/paste and then leak everything in that directory. Even without the directory support someone could leak something in a protected directory due to copy/paste. How do you see handling something like this?

If you already covered this somewhere and I missed it, which is entirely possible, please point me to it and I'll read it.

cmd/helm/helm.go
```diff
@@ -84,6 +84,11 @@ func main() {
 		}
 	})
 
+	if err := actionConfig.Init(settings.RESTClientGetter(), settings.Namespace(), os.Getenv("HELM_DRIVER"), debug); err != nil {
+		debug("%+v", err)
+		os.Exit(1)
+	}
```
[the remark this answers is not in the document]
itaispiegel: I'm not sure. This was committed by @hunsche and to me it seems irrelevant to this PR.
@mattfarina Should I revert this change?

pkg/action/action_test.go
```diff
@@ -33,6 +34,8 @@ import (
 	"helm.sh/helm/v3/pkg/time"
 )
 
+const WithExternalPathsTemplatePath = "templates/with-external-paths"
```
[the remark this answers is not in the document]
itaispiegel: Seems like it doesn't. This is actually my first code in Go, so sorry for the small mistakes :)

pkg/chart/loader/local.go
```diff
@@ -0,0 +1,75 @@
...
+
+import (
+	"errors"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+// ExpandFilePath expands a local file, dir or glob path to a list of files
+func ExpandFilePath(path string) ([]string, error) {
```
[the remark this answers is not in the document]
itaispiegel: `action` does seem like a more suitable place :)
I don't remember what was the reason it was placed in the `chart/loader` package, but I `ExpandFilePath` was public only because it's used in the `action` package.
It indeed seems like it doesn't need to be public.

itaispiegel:
@mattfarina Thanks for your reply!

> First, Helm is both a client and an SDK. There are many applications, such as controllers and operators, that use the SDK. To support both we need to keep the separation of concerns in mind. For example, the Helm client loads a chart and passes the loaded chart to the action. See the cmd/helm/install.go file for an example. This allows the application to load the chart as it needs to and pass it to the install/upgrade/etc logic to operate on. In the same way an operator could load a chart through a different means and pass it to the logic. This means we can't assume the SDK should be accessing the local file system.

The local files need to be loaded by the application (i.e. cmd/helm) and then passed into the business logic. This way something like a controller could load files through a different means and pass them to the business logic.

The idea is that we need to keep applications in mind that are not the Helm CLI.

If I understand correctly - the problem here is that applications using the Helm SDK might not be able to access the local filesystem (for example they might be running from within a container/pod). I think this can be solved well by handling the file loading only in the `cmd` package. The `action` package shouldn't be aware of this logic, because the person using it can simply pass all external files directly to the Chart's `Files` attribute. Did I understand this correctly?

> Why do a directly instead of the specific files? I'm asking for a use case for this as it's not been shared before. Please think of the application installer first.
The use of directories is needed, as it simplifies the use of this feature. If the user wants to include many files, he can pass only a single directory instead of passing all of the files separately.

> Helm needs to internally store all of these files to deal with things like rollback. These are stored, by default, on a secret. This means there are size implications. Being lazy with paths and loading things one does not need will cause size issues that need support.

Correct me if I'm mistaken - but I think this should be true about everything that the user does. If one decides to include many files, he should be aware that it might have size implications on the cluster.
Maybe a good idea though should be to have a limit on the file size, and if the user wants to include large files, he can pass a flag like `--no-file-size-limit`. What do you think about this?

> There is a security implication here. One could create a copy and paste prompt and point to the external files of /etc. Many people may blindly copy/paste and then leak everything in that directory. Even without the directory support someone could leak something in a protected directory due to copy/paste. How do you see handling something like this?

This has been discussed and I think that the user should be aware that he might be exposing secret files with this feature.
I think first of all this should be documented on Helm's website (I already opened a PR, but it was closed because it was inactive). In addition, we can try to think of some heuristics we can run to decide if this feature is used unsafely. For example, we can fail the installation if Helm detects a symlink in the included directory, or disallow certain file types.

I'll be glad to hear your opinion on this :)

rediphone:
Hope we all had a great break - its amazing to come back and see this moving forward, does this needs a final push to get there?

itaispiegel:
@mattfarina @bacongobbler Hey guys, can we please merge this? 🙏

joejulian:
The 3.12 release is scheduled for May 10th. I don't really know why they do it this way, but they usually don't merge things until preparing the release.

Reviews, however, are still needed. I haven't seen @bacongobbler around in a while so hopefully another @helm/helm-core-maintainers can review his change request and dismiss his review if he's unable to do so himself.

I know @mattfarina has been really busy lately but he always comes back in time for some release reviews.

itaispiegel:
> The 3.12 release is scheduled for May 10th. I don't really know why they do it this way, but they usually don't merge things until preparing the release.
> 
> Reviews, however, are still needed. I haven't seen @bacongobbler around in a while so hopefully another @helm/helm-core-maintainers can review his change request and dismiss his review if he's unable to do so himself.
> 
> I know @mattfarina has been really busy lately but he always comes back in time for some release reviews.

I hope we can at least merge this soon so I can consider my job here done...

itaispiegel:
@mattfarina @bacongobbler Any update? I'm still waiting for my PR to be reviewed.

itaispiegel:
Hey @jan-lucansky, I want this to be merged too, and I know there are many other users who want this feature and need it.
Unfortunately the project maintainers keep ignoring the existence of this PR, although I tried contacting them.

jan-lucansky:
@itaispiegel I see that 2 changes are requested from reviewers, isn't the PR stuck on you making those changes?

itaispiegel:
> @itaispiegel I see that 2 changes are requested from reviewers, isn't the PR stuck on you making those changes?

 I already updated the PR according to the requested changes about 5 months ago, but never got a reply from the reviewers...

rvantunes:
@mattfarina @joejulian @bacongobbler any updates on reviewing this?

joejulian (changes requested):
I agree with @mattfarina. By using globs, the risk to a novice user of exposing sensitive data without knowing it is very high. Most users use helm blindly as an install tool (think apt for kubernetes). Many don't have the experience to understand the implications of exposing an entire directory and can easily disclose secret data.

z4ce:
IMHO one the best things about helm is that it's almost all yaml-in-yaml-out. Exceptions being for crypto and lookup functions. This adds another exception to that. It seems like to me it would be better to add option to turn the files into values that get processed like anything else.  Could the use case here be solved by just allowing `-set-file` to take a directory like `kubectl create secret generic` does?

ostlerc:
Given there is some concern with this approach, it might be prudent to look at the root problem to see if there is another solution.

The problem (IMO):
users frequently want to override configuration files in a helm template. The only way to override large configuration files wholesale is through a yaml key with a multiline file as the value. Editing yaml files as sub-keys requires a certain level of indentation for the file and makes it harder to copy paste files around. The conversion from raw config file to yaml mostly is done manually adding an element of human error to making sure those details are exactly correct. It is also harder to remember how to do, requiring looking at examples and hoping you've done it right. 

The other alternative takes a lot more effort from a chart development. Requiring identifying the entire configuration file and adding template-able sections based on individual keys in the values.yaml. The problem with this approach is that adding or changing configuration template sections now requires a chart change instead of just a values.yaml change. This is a big enough QoL issue I would imagine a robust solution would go a long way making helm better.

Z4ce mentioned you can automate this with a tool or a flag to remove the human element. I think that's somewhat of a bandaid. Would it be possible to have core support in helm for this feature? i know it's very difficult, but how hard would it be in a values.yaml to specify a file syntax so a key can be populated from a local file that way?

Regardless of solution, I hope we can all agree this is a core issue of helm and should be prioritized.

joejulian:
@ostlerc I think that last line of agreement might be the hangup. If you think of helm as a package installer, like apt or yum, they don't offer ways to replace part of the package during the installation or upgrade process.

side story: back a decade or so ago, if you went into most distro's IRC channels and told them you were replacing packaged files, they'd call you names and make you vow to never try to get community help again.

Now clearly we have no desire to ever see that behavior in our community but their point, that the packaging that the chart provider is offering is done with intent is what makes helm charts supportable by their community. As a packaging tool, I don't agree that it's the responsibility of helm to facilitate overriding pieces of the package. We can leave that to (and to provide the hooks to do so) plugins or post-render tools.

ostlerc:
@joejulian  that's insightful thank you!

But I think fundamentally helm would be useless without the ability to template. Also most package installs do allow several flags to alter the way it is used. And lastly, linux package managers are fundamentally different than helm install. Once those are installed, users are intended to change user configuration files, but with kubernetes and helm that should be done at the time of install, not at a later step.

I could be wrong, but that's how I see it anyways. My team uses helm extensively for the ability to deploy to different environments the same packages. If we took out the ability to change the configuration per environment, it's just a bunch of yaml files. So I think at the core, this is what helm was meant to solve.

z4ce:
I think it's most important to think about the core interface with Helm. Right now it's values.yaml and chart in, and on the other side you get a set of Kubernetes resources. To change the contract to be values.yaml and another set of files, I think would be a mistake.  Especially when you start to think about things like Argo and Flux.

I would like to understand the use cases more and figure out how we could still keep the contract of helm as it is, while still accommodating whatever use cases are applicable here.

Why would tooling to help people put files and directories into the values structure not be sufficient?

zfalen-deloitte:
Honestly the comparison here to `apt` or `yum` etc falls short to me. It's 2023 - we are delivering software in portable containers configured to use the 12 Factor philosophy. The entire _point_ is they run via injected configuration. We aren't _(often)_ SSH'ing into VMs and running `apt-get install` anymore. With Kubernetes and Helm - in my mind the point is to pass _in_ configuration, and get a functional software stack out of it using the supplied configuration. 

_Not_ being able to pass in the _most_ common method of configuration _(files..... pg_config, myTemplate.pug, nginx.conf)_ in a coherent way from local / repository filesystems is crazy to me

yann-soubeyrand:
@zfalen-deloitte if your chart supports passing configuration files via values, I guess the `--set-file` option is what you need :wink:

benjie91:
So... it's 2024, is there any chance that this can be merged? I am still doing the workaround of copying the configs into the chart folder.

gilesknap:
I also am still using the same ugly workaround.

nooperpudd:
well, well. 4 years, my child is 4 years old. but I'm still waiting for this feature.

kalyanidani:
Well, it's mid-2024! :(
Eagerly waiting for this feature.

girdevstack:
It sure feels like every time I come across a problem with modern open source the problem was either debated until everyone gave up and moved on (from the problem or some from open source at all), or some brave soul was tenacious to work a solution and then it never gets merged because ?????

This managed to be both :sob: :sob: :sob: 

I understand maintainers and sponsors and so forth have their visions and perspectives and disagreements. But this sure feels disrespectful to everyone involved - is there some way to find consensus for these type of things or is the solution to just not become attached to *a thing*? It's so confusing. 

Back to tilting at windmills.

ostlerc:
I also wish the dialogue from maintainers would be more open minded. But given that I've accepted they are / will not, we had to design a solution around this.

What we did was create a new schema and have a helm library function template out our common use cases. It was written entirely in text/template, ouch! This is a sample of how a simple deployment looks:

ostlerc:
I also wish the dialogue from maintainers was more open-minded. But given that I’ve accepted they are not (and likely won’t be), we had to design a solution around that.

What we did was create a new schema and a Helm library function to template out our common use cases. We call it Helmix, our own mixture of Helm. It’s written entirely in text/template (ouch!) The main idea is that all configuration files you want to be easily override-able must now be in YAML format. This isn’t ideal, but it makes it possible to automatically modify configuration without reinventing the wheel each time.

Here’s an example that templates a simple configurable configmap from files on disk:


secrets/foobar.yaml
```
fooval: {{ include "some.function" . }}
status: foobar
```

helmix.yaml
```
foobar:
  configmap:
    folder: secrets
    data:
      foobar.yaml:
        status: overridden-status
```

We have a lot of defaults defined, but the entire ConfigMap is fully configurable via our schema. The goal is to keep only the most important information visible in the YAML. This is specifically designed for Kubernetes objects.

The project isn’t open source, but the idea should be straightforward for others to re-implement in their own way. If you find this approach useful, let me know—I’d love to hear if others have alternative or better solutions.

PauloFerreira25:
Vai fechar não bot, pode deixar aberto isso ai.

PauloFerreira25:
Vai ficar aberto até resolver

PauloFerreira25:
Vai ficar aberto até resolver

PauloFerreira25:
Vai ficar aberto até resolver, não vou deixar fechar

PauloFerreira25:
Fecha não, bota os agents de IA pre resolver...

joejulian:
This has been open a long time, and I want to resolve it honestly rather than leave it sitting open indefinitely.

First, the demand is real and I am not waving it off. A lot of people have asked for this over the years, @itaispiegel put real work in here, and the origin request in #3276 is still open. This is the second PR to try it (#8841 was the first, closed in 2022), and both stalled on the same thing, which tells me the blocker is the shape, not the effort.

The blocker is Helm's core contract. Helm takes a chart plus values in and produces Kubernetes resources out, and that chart is a single unit that gets loaded and passed around. External files alongside the chart add a third kind of input to that contract, and it breaks the client-and-SDK split we have to preserve: the many controllers and operators that use the Helm SDK load a chart and hand it to Helm, and they have no way to also hand over files sitting on someone's filesystem. Matt raised that here, and z4ce put the general version of it well: Helm is almost entirely yaml-in, yaml-out (with lookup and the crypto functions as the rare exceptions), and this adds another exception to the thing that makes charts portable.

The supported ways to do what people are reaching for already exist, without changing that contract:

- Turn the files into values and let them flow through templating like everything else (z4ce's suggestion in the thread).
- For replacing rendered output wholesale, a post-renderer.

If external-file input should become a real, first-class part of what a chart is, that is worth doing properly as a Helm Improvement Proposal anchored on #3276, where the contract change can be designed on purpose rather than added through an install/template flag.

I have kept this open across a few milestones hoping a shape would surface that fit, and I left change requests here myself back in 2023. After four years, and with Helm 4 not changing the chart contract, that shape has not emerged, so I am going to close this. I am leaving #3276 open as the home for the need.

Thank you, @itaispiegel, and everyone who engaged. The want was legitimate the whole way through. This is a close on shape, not on worth.

PauloFerreira25:
@joejulian igual a todas as essas https://github.com/helm/helm/pull/10077#issuecomment-1283044522

PR #10096, opened 2021-09-03
Make `version` optional for local chart dependencies

### LINKED ISSUE #10095: `helm dependency update` complains about lack of `version` even when `repository` is not set
Let's say I have:

```yaml
# Chart.yaml
apiVersion: v2
appVersion: "0.1"
name: chart-b
version: 0.0.1

dependencies:
  - name: chart-a
    condition: chart-a.enabled
```

And my `chart-a` is under `chart-b/charts/chart-a`.

The following happens:

```console
$ cd chart-a
$ helm dependency update
Error: dependency "chart-a" has an invalid version/constraint format: improper constraint:
```

Nonetheless, if I put the version, the following happens:

```
$ helm dependency update
Saving 1 charts
Dependency chart-a did not declare a repository. Assuming it exists in the chart directory
Deleting outdated charts
```

As it is stated in the output, as `repository` is not set, `helm` correctly assumes that we are managing the `charts/chart-a` by ourselves. Thus, it is redundant for checking the presence of `version` when `repository` is not set.

It is even a problem because when we bump the version of the `chart-a`, now it needs to be bumped in `chart-b` as well, even though the dependency is directly linked on the file system.

felipecrs:

Fixes #10095

References https://github.com/helm/helm-www/pull/1183/files
References https://github.com/SchemaStore/schemastore/pull/1815
References https://github.com/SchemaStore/schemastore/pull/1816

**If applicable**:
  The `--help` text was updated to reflect that `version` is optional for local charts. Also, the referenced PR changes the same in the website.
  Everything should remain functional, this only prevents an uneeded failure. The fact that none of the unit tests were touched is already a good signal of that.


internal/resolver/resolver.go
```diff
@@ -58,43 +58,54 @@ func (r *Resolver) Resolve(reqs []*chart.Dependency, repoNames map[string]string
...
-		if err != nil {
-			return nil, errors.Wrapf(err, "dependency %q has an invalid version/constraint format", d.Name)
-		}
-
-		if d.Repository == "" {
-			// Local chart subfolder
-			if _, err := GetLocalPath(filepath.Join("charts", d.Name), r.chartpath); err != nil {
-				return nil, err
-			}
+		var constraint *semver.Constraints
```
yxxhero, line 61: var chartpath string
var err error

internal/resolver/resolver.go
```diff
@@ -58,43 +58,54 @@ func (r *Resolver) Resolve(reqs []*chart.Dependency, repoNames map[string]string
...
 
-			locked[i] = &chart.Dependency{
-				Name:       d.Name,
-				Repository: "",
-				Version:    d.Version,
+		// If version is defined
+		if d.Version != "" {
+			var err error
```
yxxhero, line 65: remove `var err error`

internal/resolver/resolver.go
```diff
@@ -58,43 +58,54 @@ func (r *Resolver) Resolve(reqs []*chart.Dependency, repoNames map[string]string
...
-			continue
 		}
-		if strings.HasPrefix(d.Repository, "file://") {
 
-			chartpath, err := GetLocalPath(d.Repository, r.chartpath)
-			if err != nil {
-				return nil, err
+		// Local chart
+		if d.Repository == "" || strings.HasPrefix(d.Repository, "file://") {
+			var chartpath string
```
yxxhero, line 74: remove `var chartpath string`

internal/resolver/resolver.go
```diff
@@ -58,43 +58,54 @@ func (r *Resolver) Resolve(reqs []*chart.Dependency, repoNames map[string]string
...
+					return nil, err
+				}
+			} else {
+				// From file:// repository
+				var err error
+				chartpath, err = GetLocalPath(d.Repository, r.chartpath)
+				if err != nil {
+					return nil, err
+				}
 			}
```
yxxhero, line 91: remove `else`

```yaml
chartpath, err = GetLocalPath(d.Repository, r.chartpath)
if err != nil {
    return nil, err
}
```

yxxhero (changes requested):

internal/resolver/resolver.go
felipecrs: I don't understand. Why would I remove the `else`?

felipecrs: The else is needed so I run the block of commands only when the repository is set with `file://`.

yxxhero: @felipecrs When there are break, continue, goto or return statements in the if structure, the common way to write go code is to omit the else part.

felipecrs: @yxxhero I agree. But the `return` which is inside of the `if` statement

https://github.com/helm/helm/blob/3f7a288ed2eb94b95727214d372dc8b3bb4f4504/internal/resolver/resolver.go#L76-L82

is only called when an error happens, and not always.

If I remove the `else`, `chartpath` would get assigned twice if no repository is set, with the wrong value btw.

I am really a first-time contributor here, and in Golang at all, so I am really sorry if I am missing something.

yxxhero: @felipecrs you are right, it is good.

felipecrs: That's great then :D

crabique:
Are there any blockers to get this merged? This is a really silly problem and the PR has been cooking for years now.

felipecrs:
@crabique, I rebased it. It's ready again for review.

internal/resolver/resolver.go
```diff
@@ -58,42 +58,51 @@ func (r *Resolver) Resolve(reqs []*chart.Dependency, repoNames map[string]string
...
+			// Load chart to validate the version
 			ch, err := loader.LoadDir(chartpath)
 			if err != nil {
 				return nil, err
 			}
 
 			v, err := semver.NewVersion(ch.Metadata.Version)
 			if err != nil {
-				// Not a legit entry.
+				// If version is not set and the version from the local chart is invalid
```
gjenkins8, line 97: This comment is just parroting the logic,  I think is unnecessary

internal/resolver/resolver.go
```diff
@@ -58,42 +58,51 @@ func (r *Resolver) Resolve(reqs []*chart.Dependency, repoNames map[string]string
...
+				if d.Version == "" {
+					missing = append(missing, d.Name)
```
gjenkins8, line 99: Whats the rationale for a child chart to be errored as an invalid dependency version, just because the parent chart didn't specify a version?

(ie. I'm not quite understanding the reason for this)

gjenkins8:
```suggestion
					missing = append(missing, fmt.Sprintf(... 
```

similar to line 106

internal/resolver/resolver.go
```diff
@@ -58,42 +58,51 @@ func (r *Resolver) Resolve(reqs []*chart.Dependency, repoNames map[string]string
...
-				return nil, err
-			}
-
-			locked[i] = &chart.Dependency{
-				Name:       d.Name,
-				Repository: "",
-				Version:    d.Version,
+		// If version is defined
+		if d.Version != "" {
+			constraint, err = semver.NewConstraint(d.Version)
```
gjenkins8, line 67: `constraint` isn't used until line 105, why move the version passing logic down there?

internal/resolver/resolver.go
```diff
@@ -58,42 +58,51 @@ func (r *Resolver) Resolve(reqs []*chart.Dependency, repoNames map[string]string
...
+				}
 			}
 
+			// Load chart to validate the version
 			ch, err := loader.LoadDir(chartpath)
 			if err != nil {
 				return nil, err
 			}
 
 			v, err := semver.NewVersion(ch.Metadata.Version)
```
gjenkins8, line 95: this check and others below was not performed when `d.Repository == ""`. We need to retain the prior behavior to avoid breaking existing users

internal/resolver/resolver.go
```diff
@@ -58,42 +58,51 @@ func (r *Resolver) Resolve(reqs []*chart.Dependency, repoNames map[string]string
...
-			continue
 		}
-		if strings.HasPrefix(d.Repository, "file://") {
-			chartpath, err := GetLocalPath(d.Repository, r.chartpath)
-			if err != nil {
-				return nil, err
+
+		// Local chart
+		if d.Repository == "" || strings.HasPrefix(d.Repository, "file://") {
+			if d.Repository == "" {
```
gjenkins8, line 75: this conditional IMHO has made the code more complex. The empty repository and `file://` loading logic has been combined into a single branch. With the intent of utilizing the same version processing logic.

But... (while this sounds like a good thing) the subtle differences in these code paths is leading to lots of conditional checks. It would be great to find a way to simplify the conditional logic here I think.

gjenkins8 (changes requested):
There is a behavioral difference I think. And overall (it is hard to quantify), it feels like this change is significantly increasing the complexity of the logic here. Reducing the complexity would be preferred :)

felipecrs:
@gjenkins8 thanks a lot for reviewing this PR too.

It has been so much since I wrote this that I'll need to spend more time trying to understand everything I did to address your feedback.

I will probably not be able to finish this before next weekend, thus I will mark it as _Draft_ for now.

internal/resolver/resolver.go
felipecrs: I understand, but I don't think it's as bad as it sounds. We know that in real life helm will never have to iterate on these conditions long enough to make a difference in performance. It's not like there will be a chart with thousands of dependencies (which is probably still a low number of iterations).

Nevertheless, I'm still thinking about it.

crabique:
@felipecrs hi, do you plan to resume this effort in the foreseeable future?

internal/resolver/resolver.go
felipecrs: To validate and fail as early as possible for invalid versions. I can move this section down if you want.

felipecrs: I included a code comment to make this clearer. I will mark this as resolved, but feel free to unmark it.

internal/resolver/resolver.go
felipecrs: @gjenkins8 I guess I became dumber over time. I had to ask Claude Sonnet 4 to help me.

But I reviewed the outcome, and it looks fine.

Can you please review this again?

internal/resolver/resolver.go
felipecrs: I believe it's satisfactory now. I will mark as resolved, but feel free to unmark it.

felipecrs:
> @felipecrs hi, do you plan to resume this effort in the foreseeable future?

Yes, I just rebased and handled @gjenkins8 comments. Hopefully in a satisfactory manner.

felipecrs:
@gjenkins8 maybe you can have a new look into this?

crabique:
@gjenkins8 @yxxhero please have a look, if released who knows, it may be a good enough reason for people to switch to v4 😄

felipecrs:
Not stale.

felipecrs:
Not stale.

PR #10106, opened 2021-09-06
feat(*): add --quiet flag to 'helm lint'

### LINKED ISSUE #9994: helm lint is too noisy
When running `helm lint` against a large number of charts, I see a lot of output even in the case when no problems are detected.

Each invocation prints `1 chart(s) linted, 0 chart(s) failed` and `Using values from 'foo/bar/baz.yaml'`. This makes it hard to tell when there actually is a problem.

It would be nice if `helm lint` had a "quiet" option or some other way to reduce the verbosity, such that the only text that is written to the console is lint warnings/errors, and a clean run with no issues presents no output.

presztak:

Quiet flag reduce the verbosity for `helm lint` command, such that the only text
that is written to the console is lint warnings/errors

Closes #9994

**If applicable**:


zonggen (commented):
Code looks sane to me, waiting for maintainers to confirm the functionality.

Can confirm the test case passes on my machine:
```bash
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.0", GitCommit:"c2b5237ccd9c0f1d600d3072634ca66cefdf272f", GitTreeState:"clean", BuildDate:"2021-08-04T18:03:20Z", GoVersion:"go1.16.6", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.1", GitCommit:"632ed300f2c34f6d6d15ca4cef3d3c7073412212", GitTreeState:"clean", BuildDate:"2021-08-19T15:39:34Z", GoVersion:"go1.16.7", Compiler:"gc", Platform:"linux/amd64"}
```

cmd/helm/lint.go
```diff
@@ -112,14 +128,17 @@ func newLintCmd(out io.Writer) *cobra.Command {
...
+				fmt.Fprintln(out, summary)
+			}
 			return nil
 		},
 	}
 
 	f := cmd.Flags()
 	f.BoolVar(&client.Strict, "strict", false, "fail on lint warnings")
 	f.BoolVar(&client.WithSubcharts, "with-subcharts", false, "lint dependent charts")
+	f.BoolVar(&client.Quiet, "quiet", false, "print only warning and errors")
```
bacongobbler, line 141: typo: warnings*

cmd/helm/lint.go
```diff
@@ -93,10 +101,18 @@ func newLintCmd(out io.Writer) *cobra.Command {
 				}
 
 				for _, msg := range result.Messages {
-					fmt.Fprintf(&message, "%s\n", msg)
+					if client.Quiet {
```
bacongobbler, line 104: You can significantly reduce the cyclomatic complexity of this `if/else` block by rewriting it like this:

```
if !client.Quiet || msg.Severity > support.InfoSev {
    fmt.Fprintf(&message, "%s\n", msg)
}
```

truth table:

```
quiet | severity | output |
0     | 0         | 1     |
0     | 1         | 1     |
1     | 0         | 0     |
1     | 1         | 1     |
```

bacongobbler (dismissed):

mattfarina:
@bacongobbler can you take another look at this with the latest changes?

hickeyma (approved):
LGTM, thanks @presztak

cmd/helm/lint.go
```diff
@@ -112,14 +122,17 @@ func newLintCmd(out io.Writer) *cobra.Command {
...
+				fmt.Fprintln(out, summary)
+			}
 			return nil
 		},
 	}
 
 	f := cmd.Flags()
 	f.BoolVar(&client.Strict, "strict", false, "fail on lint warnings")
 	f.BoolVar(&client.WithSubcharts, "with-subcharts", false, "lint dependent charts")
+	f.BoolVar(&client.Quiet, "quiet", false, "print only warnings and errors")
```
mattfarina, line 135: The message here is not correct. `--quiet` will only show errors by default. If you want warnings to display you need to also have `--strict` set. Was the intent to always show warnings when quiet? If so, there is a large code change needed. If the current functionality is ok then the flag doc string needs to be updated.

mattfarina (changes requested):

cmd/helm/lint.go
hickeyma: Discussed this with @mattfarina last week and a good catch by him. Sorry only getting to respond now.

You can reproduce this by creating a scaffold chart (`helm create`) and remove the `templates/` directory. You can see the difference when you test with, without the flag and with `strict` flag:

```console
$ helm lint lint-warns
==> Linting lint-warns
[WARNING] templates/: directory not found

$ helm lint lint-warns --quiet

$ helm lint lint-warns --quiet --strict 
==> Linting lint-warns
[WARNING] templates/: directory not found

```

@presztak How do you think it should work with the `--quiet` flag?

presztak: @hickeyma

I believe that intention of this issue was displaying errors as well warnings.
I fixed the code and it should be better right now. What do you think about this solution?
I also added new test which checks is quiet flag works well with chart which has only warnings.

If this solution is not ok we can turn back to the previous one and change quiet flag description.

hickeyma (approved):
LGTM, thanks @presztak

hickeyma:
@mattfarina When you get a chance, can you take a look again?

yxxhero (approved):
LGTM

mattfarina (approved):

PR #10124, opened 2021-09-13
feat(values): support YAML streams in values files (closes #10125)

### LINKED ISSUE #10125: Support streams in values files
It would be useful if single values files could contain multiple documents. This way, a single file with multiple hashes could be provided.

This would be useful for example when passing ad-hoc values to ArgoCD Applications, as they are passed as a single string.

When implemented, passing a single values files containing:

```
---
image: "foo"
---
image: "overridden-foo"
```

should result in ` .Values.image` with value `overridden-foo`.

raphink:
Signed-off-by: Raphaël Pinson <raphael.pinson@camptocamp.com>

This simple PR allows to consider values files as streams of multiple documents. This is useful to pass multiple value documents as one file, as required for example by ArgoCD applications (when passing verbatim values), without having to merge the values beforehand.

As a result, a single values files such as:

```
---
image: "foo"
---
image: "overridden-foo"
```

will now set `.Values.image` to `overridden-foo`.

Note: this does not apply to the default values file, only the additional files passed to the CLI.

**If applicable**:

### DIFF
--- a/pkg/cli/values/options.go
+++ b/pkg/cli/values/options.go
@@ -17,13 +17,14 @@ limitations under the License.
 package values
 
 import (
+	"io"
 	"io/ioutil"
 	"net/url"
 	"os"
 	"strings"
 
 	"github.com/pkg/errors"
-	"sigs.k8s.io/yaml"
+	"k8s.io/apimachinery/pkg/util/yaml"
 
 	"helm.sh/helm/v3/pkg/getter"
 	"helm.sh/helm/v3/pkg/strvals"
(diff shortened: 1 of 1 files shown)


zonggen:
Could you link/create the corresponding issue to this PR, thank you :)

Reference: [CONTRIBUTING.md#how-to-contribute-a-patch](https://github.com/helm/helm/blob/main/CONTRIBUTING.md#how-to-contribute-a-patch)

raphink:
@zonggen Do I need to create a HIP for this, or just open an issue?

zonggen:
Just an issue should be sufficient, then it would be triaged by one of the maintainers.

raphink:
Opened #10125 for that

yxxhero:
@raphink You should [warrants a HIP](https://github.com/helm/helm/blob/main/CONTRIBUTING.md#proposing-an-idea). And add unittest. Thanks very much.

PR #10126, opened 2021-09-14
feat: plugin.yaml support platformHooks (closes #7117)

### LINKED ISSUE #7117: Windows support for Helm plugin installers
Helm's plugin install hook system relies on `/bin/sh`, regardless of the operating system present. Windows users can work around this by using Helm under WSL, but it would be great to have true Windows support out of the box.

Example use case as demonstrated in https://github.com/helm/helm-2to3/issues/55:

```
$ helm3 plugin install https://github.com/helm/helm-2to3 --debug
[debug] updating https://github.com/helm/helm-2to3
[debug] symlinking C:\Users\nlowe\AppData\Local\Temp\helm\plugins\https-github.com-helm-helm-2to3 to C:\Users\nlowe\AppData\Roaming\helm\plugins\helm-2to3
plugin_install.go:75: [debug] loading plugin from C:\Users\nlowe\AppData\Roaming\helm\plugins\helm-2to3
plugin.go:60: [debug] running install hook: &{sh [sh -c cd $HELM_PLUGIN_DIR; scripts/install_plugin.sh] []  <nil> <nil> <nil> [] %!s(*syscall.SysProcAttr=<nil>) %!s(*os.Process=<nil>) <nil> <nil> %!s(*exec.Error=&{sh 0xc00008d5c0}) %!s(bool=false) [] [] [] [] %!s(chan error=<nil>) %!s(chan struct {}=<nil>)}
Error: exec: "sh": executable file not found in %PATH%
helm.go:81: [debug] exec: "sh": executable file not found in %PATH%
```

dimabru:
This pr adds `platformHook` behavior for plugin installations. This is mainly to support plugin installation for windows, since at this point, helm plugins are not able to run on windows.
Usage is very similar to current `platformCommand`. The user can specify specific os to run hooks on, and if missing, default to root `hooks` key in plugin yaml

closes #7117

For env vars, there was an issue with using built-in `os.ExpandEnv` function on windows. 
Hence, there's no support for `$var` in windows scripts. What is working though is windows env var syntax `%VAR%` - mainly used for `%HELM_PLUGIN_DIR%`

**If applicable**:

### DIFF
--- a/cmd/helm/plugin.go
+++ b/cmd/helm/plugin.go
@@ -19,6 +19,8 @@ import (
 	"io"
 	"os"
 	"os/exec"
+	"runtime"
+	"strings"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
@@ -47,15 +49,22 @@ func newPluginCmd(out io.Writer) *cobra.Command {
 
 // runHook will execute a plugin hook.
 func runHook(p *plugin.Plugin, event string) error {
-	hook := p.Metadata.Hooks[event]
-	if hook == "" {
+	hookParts, hookErr := p.PrepareHook(event)
+	if len(hookParts) == 0 {
 		return nil
 	}
+	if hookErr != nil {
+		os.Stderr.WriteString(hookErr.Error())
+		return errors.Errorf("plugin hook %s exited with error", event)
+	}
+	hook := strings.Join(hookParts, " ")
 
-	prog := exec.Command("sh", "-c", hook)
-	// TODO make this work on windows
-	// I think its ... ¯\_(ツ)_/¯
-	// prog := exec.Command("cmd", "/C", p.Metadata.Hooks.Install())
+	var prog *exec.Cmd
+	if runtime.GOOS == "windows" {
+		prog = exec.Command("cmd", "/C", hook)
+	} else {
+		prog = exec.Command("sh", "-c", hook)
+	}
 
 	debug("running %s hook: %s", event, prog)
 
--- a/pkg/plugin/plugin.go
+++ b/pkg/plugin/plugin.go
@@ -50,6 +50,12 @@ type PlatformCommand struct {
 	Command         string `json:"command"`
 }
 
+type PlatformHook struct {
+	OperatingSystem string `json:"os"`
+	Architecture    string `json:"arch"`
+	Hooks           Hooks  `json:"hooks"`
+}
+
 // Metadata describes a plugin.
 //
 // This is the plugin equivalent of a chart.Metadata.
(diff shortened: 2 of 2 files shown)


dimabru:
closes #7117

dimabru:
Hi @bacongobbler, any chance you can have a look at the code? 
Thanks a lot :)

TBBle:
With a quick look, this also changes the behaviour of `hooks`, which I don't think is desirable, as it will break existing working setups, e.g. running `helm plugin install` under MSYS2 which is the work-around I've used for #7117.

> since at this point, helm plugins are not able to run on windows.

Helm plugins _run_ on Windows, as long as their installer bash script recognises Windows and installs a Windows binary as, e.g., https://github.com/helm/helm-2to3/blob/main/scripts/install_plugin.sh, does.

Also, it continues to hard-code a per-platform shell. At least, [by my suggestion in #7117](https://github.com/helm/helm/issues/7117#issuecomment-657342139), `platformHooks` would be up to the platform hook itself to specify `sh -c` or `cmd /c` or whatever the appropriate shell is, e.g. on Windows, PowerShell would be a reasonable choice for hook implementation. This would mean that the existing hooks are equivalent to a "default" platformHook of `sh -c <existing content>` which is reasonably non-surprising.

I'm not sure if I'd group the platform hooks by OS as done here, or by event as the current hooks are. That's assuming I'm reading this correctly: lack of tests means I have to imagine what the config file looks like with these changes.

It might even be handy to prepare a draft PR against a plugin or two (that already support WIndows) to see how your changes look in the real world, or ask the maintainers of those plugins to do so. That would be important ecosystem feedback anyway.

dimabru:
Hi @TBBle 
Thanks a lot for the detailed comment.

In your suggestion in https://github.com/helm/helm/issues/7117#issuecomment-657342139 you suggest to deprecate `hooks` and keep only the new `platformHooks` behavior. Why is that? Why not keep `hooks` and use `sh -c` as fallback for backwards compatibility? Also, I think its not far-fetched to assume that helm defaults to shell executable in case `platformHook` is missing.

I think I understand what you mean when you say this change will break existing behavior. I will change the script to not assume exec script by checking os, but rather add an option for the user to explicitly declare the preffered way of execution.

As you suggested, I will add unit-tests and a draft pr in another plugin to give a clear example on how this behavior might look.

TBBle:
Deprecating `hooks` would in-effect be keeping it as backwards-compat behaviour: Existing plugins would install exactly as they do now, including assuming `sh -c` and hence no working on Windows without further user interaction. That's also the reason to deprecate it, because it assumes something that isn't a great assumption across platforms.

Basically, I was mirroring the behaviour of `platformCommand`, and that [does not prepend a shell in the command](https://github.com/helm/helm/blob/9fafb4ad6811afb017cc464b630be2ff8390ac63/cmd/helm/load_plugins.go#L124-L132). I think having `platformHooks` act differently to `platformCommand` in this regard would be a surprising mistake.

mindw:
Please re-open this, helm plugins on windows is a thing.

PR #10148, opened 2021-09-17
Export namespace in cli.EnvSettings

ndhanushkodi:
Changes `EnvSettings.namespace` to `EnvSettings.KubeNamespace` so it's exported.

This is useful for applications that use the Go SDK and need to set the
namespace without relying on the HELM_NAMESPACE environment variable. Example here: https://github.com/hashicorp/consul-k8s/blob/main/cli/cmd/install/install.go#L191-L198

The action.Configuration Init function initializes a kube client using the namespace in cli.EnvSettings, and the only way we'd be able to currently set it is by using the HELM_NAMESPACE environment variable. For the helm CLI itself, it can rely on the flags to update this value, but other applications using the Go SDK can't use this. 

I changed `namespace` to `KubeNamespace` rather than `Namespace` to not conflict with the function name.

**If applicable**:

### DIFF
--- a/pkg/cli/environment.go
+++ b/pkg/cli/environment.go
@@ -39,9 +39,10 @@ const defaultMaxHistory = 10
 
 // EnvSettings describes all of the environment settings.
 type EnvSettings struct {
-	namespace string
-	config    *genericclioptions.ConfigFlags
+	config *genericclioptions.ConfigFlags
 
+	// KubeNamespace is the namespace to use for the Kubernetes client.
+	KubeNamespace string
 	// KubeConfig is the path to the kubeconfig file
 	KubeConfig string
 	// KubeContext is the name of the kubeconfig context.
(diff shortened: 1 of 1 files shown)


bacongobbler:
How do you reconcile the difference between `EnvSettings{}.Namespace()` and `EnvSettings{}.KubeNamespace`? Because both are technically supposed perform the same task (get the environment's current namespace), but now we have two separate ways of retrieving the namespace. And both can return differing results.

While I agree that exporting this in some manner would be ideal to package consumers... Exposing this parameter leaves the package in a bit of an awkward state.

ndhanushkodi:
Thanks for the fast response @bacongobbler! I see what you are saying.

Another option could be to leave the namespace field unexported, and add an `EnvSettings{}.SetNamespace(ns string)` function that allows you to set the unexported field `EnvSettings{}.namespace`. That way, we would just keep the 1 existing getter function `EnvSettings{}.Namespace()`. What do you think?

Akkpon75 (approved):

bacongobbler:
closed via https://github.com/helm/helm/pull/10230

PR #10174, opened 2021-09-22
Fix subcharts dependencies processing

### LINKED ISSUE #10160: [Bug] 3.7.0 - Helm generates an incorrect chart tree in some cases
**Edit**: I initially thought this had to do with the new `.Subcharts` feature, which I was using, but it seems the problem is unrelated. I updated the title to reflect that.

---

Let me start by thanking @vflaux and others for this long awaited feature. The sort of `call-nested` workaround was a pain to use and broken in many scenarios. Thank you all 👍.

That being said, I'm either doing something really wrong here or there is a bug. It seems the `.Subcharts` are not correctly propagated when using *multiple* levels of subcharts _if_ importing the same subchart (using different aliases, of course).

You can find my attempt at reproducing the issue [here](https://github.com/tavlima/helm-370-subchart-issue). It may be possible to reproduce it with a smaller example, but that is what I could put together in the time I had available.

The dependencies look like this:

```
composition <- part (as part1) <- subpart (as subpart1) <- common
                               <- subpart (as subpart2) <- common
            <- part (as part2) <- subpart (as subpart1) <- common
                               <- subpart (as subpart2) <- common
```

The `common` chart includes a slightly modified version of the default `foo.fullname` named template, to highlight the issue (the `not_a_string` else clause):

```tpl
{{/*
Create a default fully qualified app name.
We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
If release name contains chart name it will be used as a full name.
*/}}
{{- define "common.fullname" -}}
  {{- if .Values.fullnameOverride }}
    {{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" }}
  {{- else }}
    {{- $name := default .Chart.Name .Values.nameOverride }}
    {{- if typeIs "string" $name }}
      {{- if contains $name .Release.Name }}
        {{- .Release.Name | trunc 63 | trimSuffix "-" }}
      {{- else }}
        {{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" }}
      {{- end }}
    {{- else }}
      {{- printf "not_a_string" }}
    {{- end }}
  {{- end }}
{{- end }}
```

And this is the (unexpected) output of `helm template composition` (after using `helm dep up` at all levels, of course):

```yaml
---
# Source: composition/charts/part1/charts/subpart1/templates/subpart.yaml
subpart:
  title: This is from subpart
  fullname: RELEASE-NAME-subpart1
---
# Source: composition/charts/part1/charts/subpart2/templates/subpart.yaml
subpart:
  title: This is from subpart
  fullname: RELEASE-NAME-subpart2
---
# Source: composition/charts/part1/templates/part.yaml
part:
  title: This is from part
  fullname:
    subpart1: RELEASE-NAME-subpart1
    subpart2: RELEASE-NAME-subpart2
---
# Source: composition/charts/part2/charts/subpart/templates/subpart.yaml
subpart:
  title: This is from subpart
  fullname: RELEASE-NAME-subpart
---
# Source: composition/charts/part2/templates/part.yaml
part:
  title: This is from part
  fullname:
    subpart1: not_a_string
    subpart2: not_a_string
---
# Source: composition/templates/composition.yaml
composition:
  title: This is from the composition
  fullname:
    part1: RELEASE-NAME-part1
    part2: RELEASE-NAME-part2
```

The fun part: if I change `composition`'s `Chart.yaml` to change the dependencies order, by aliasing the first `part` dependency as `part2` and the last as `part1`, it is now the `composition/charts/part1/templates/part.yaml` that reports the `not_a_string` values. 🤷‍♂️

Thoughts?

---

Output of `helm version`: `version.BuildInfo{Version:"v3.7.0", GitCommit:"eeac83883cb4014fe60267ec6373570374ce770b", GitTreeState:"clean", GoVersion:"go1.16.8"}`
Output of `kubectl version`: not deemed relevant
Cloud Provider/Platform (AKS, GKE, Minikube etc.): not deemed relevant

vflaux:

There is an issue when you use a subchart more than once with alias and that this subchart also have a subchart.

In pkg/chartutil/dependencies.go, getAliasDependency() return a partial copy of the passed chart keeping the same slice for metadata.Dependencies. This results in metadata.Dependencies been shared by multiple Charts objects.
At some point processDependencyEnabled() changes the name of each dependencies by their alias. This impact others dependencies generated by getAliasDependency() and leads to processDependencyEnabled() to not correctly handles their dependencies (as it use the name to do so and the name is now the alias).

This patch make a copy of all the dependencies in a new slice to avoid this issue.

fix #10160

**If applicable**:

### DIFF
--- a/pkg/chartutil/dependencies.go
+++ b/pkg/chartutil/dependencies.go
@@ -106,6 +106,15 @@ func getAliasDependency(charts []*chart.Chart, dep *chart.Dependency) *chart.Cha
 		md := *c.Metadata
 		out.Metadata = &md
 
+		if md.Dependencies != nil {
+			dependencies := make([]*chart.Dependency, len(md.Dependencies))
+			for i := range md.Dependencies {
+				d := *md.Dependencies[i]
+				dependencies[i] = &d
+			}
+			md.Dependencies = dependencies
+		}
+
 		if dep.Alias != "" {
 			md.Name = dep.Alias
 		}
--- a/pkg/chartutil/dependencies_test.go
+++ b/pkg/chartutil/dependencies_test.go
@@ -312,20 +312,34 @@ func TestDependentChartAliases(t *testing.T) {
 	c := loadChart(t, "testdata/dependent-chart-alias")
 	req := c.Metadata.Dependencies
 
-	if len(c.Dependencies()) != 2 {
-		t.Fatalf("expected 2 dependencies for this chart, but got %d", len(c.Dependencies()))
+	if len(c.Dependencies()) != 3 {
+		t.Fatalf("expected 3 dependencies for this chart, but got %d", len(c.Dependencies()))
 	}
 
 	if err := processDependencyEnabled(c, c.Values, ""); err != nil {
 		t.Fatalf("expected no errors but got %q", err)
 	}
 
-	if len(c.Dependencies()) != 3 {
+	if len(c.Dependencies()) != 5 {
 		t.Fatal("expected alias dependencies to be added")
 	}
 
 	if len(c.Dependencies()) != len(c.Metadata.Dependencies) {
 		t.Fatalf("expected number of chart dependencies %d, but got %d", len(c.Metadata.Dependencies), len(c.Dependencies()))
+	}
+
+	for _, dep := range c.Dependencies() {
+		if len(dep.Metadata.Dependencies) == 0 {
+			continue
+		}
+		if len(dep.Dependencies()) != len(dep.Metadata.Dependencies) {
+			t.Fatalf("subchart %s: expected number of dependencies %d, but got %d", dep.Name(), len(dep.Metadata.Dependencies), len(dep.Dependencies()))
+		}
+		for i := range dep.Dependencies() {
+			if dep.Metadata.Dependencies[i].Name != dep.Dependencies()[i].Name() {
+				t.Fatalf("subchart %s: expected name %s, but got %s", dep.Name(), dep.Metadata.Dependencies[i].Name, dep.Dependencies()[i].Name())
+			}
+		}
 	}
 
 	aliasChart := getAliasDependency(c.Dependencies(), req[2])
--- a/pkg/chartutil/testdata/dependent-chart-alias/Chart.yaml
+++ b/pkg/chartutil/testdata/dependent-chart-alias/Chart.yaml
@@ -27,3 +27,11 @@ dependencies:
     version: "4.3.2"
     repository: https://example.com/charts
     alias: mariners1
+  - name: gizmo
+    version: "0.1.0"
+    repository: https://example.com/charts
+    alias: gizmo1
+  - name: gizmo
+    version: "0.1.0"
+    repository: https://example.com/charts
+    alias: gizmo2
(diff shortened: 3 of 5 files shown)


bacongobbler:
Please provide unit tests to ensure your change fixes the issue. Thanks.

pkg/chartutil/dependencies.go
```diff
@@ -102,8 +102,15 @@ func getAliasDependency(charts []*chart.Chart, dep *chart.Dependency) *chart.Cha
 			continue
 		}
 
-		out := *c
 		md := *c.Metadata
+		deps := make([]*chart.Dependency, 0, len(md.Dependencies))
+		for _, dep := range md.Dependencies {
+			dep := *dep
```
bacongobbler, line 108: Why is this de-referencing the address, then re-assigning a new pointer? Wouldn't a `copy` be safer?

What happens when you provide a list of `nil` dependencies? Wouldn't that result in a stack trace when you attempt to de-reference `nil`? How should this code handle nil dependencies?

vflaux: A copy would not give us new Dependency structs.
We need new ones because their name are modified here which leaks on other Charts:  https://github.com/helm/helm/blob/9fafb4ad6811afb017cc464b630be2ff8390ac63/pkg/chartutil/dependencies.go#L139-L146

When you call `getAliasDependency(c.Dependencies(), req)` multiple time with the same `req.Name`, you get new Charts structs each time but their metadata.Dependencies are the same slice which come from the Chart in `c.Dependencies()` that match the Dependency (`req`).
Then we iterate over these dependencies (*chart.Chart) and call `processDependencyEnabled()`
https://github.com/helm/helm/blob/9fafb4ad6811afb017cc464b630be2ff8390ac63/pkg/chartutil/dependencies.go#L186-L191
At this point the dependencies generated by `getAliasDependency` are now the pointer `c`.
Then we iterate over `c.Metadata.Dependencies` (and call again `getAliasDependency`) then replace `req.Name` with `req.Alias`. But `c.Metadata.Dependencies` is the slice return by `getAliasDependency` in the parent, which may be shared with another Chart as seen before. 
When the `processDependencyEnabled` is called with such a Chart (same name but different alias), Dependency structs in `c.Metadata.Dependencies` have already their name changed to the alias and the following match do not work as expected:
https://github.com/helm/helm/blob/9fafb4ad6811afb017cc464b630be2ff8390ac63/pkg/chartutil/dependencies.go#L130-L133
The only Chart dependencies that are kept are the one in `c.Dependencies`, not in `c.Metadata.Dependencies`.

With a new slice and new structs for `.metadata.Dependencies`, the `req.Name = req.Alias` do not leaks on other Charts.

Not sure if I'm clear enough here, sorry :confused:.

--- 
Charts.Dependencies are apparently not supposed to contains nil pointers as this line would panic on the `req.Name` in this case: https://github.com/helm/helm/blob/9fafb4ad6811afb017cc464b630be2ff8390ac63/pkg/chartutil/dependencies.go#L130-L132
But I can add checks for nil if you prefer.

vflaux:
Just find that there is another PR for the same issue and with the same fix: https://github.com/helm/helm/pull/9175

PR #10182, opened 2021-09-24
changes to support any mediatype for layers on download

### LINKED ISSUE #10176: OCI: Allow any mediatype for layers on download
As of Helm 3.7, old charts published with `helm chart push` or `oras` are no longer able to be used with `helm pull` or `helm install` etc.

This is because there is strict validation of the chart layer mediatype to match that which is used on push:
```
application/vnd.cncf.helm.chart.content.v1.tar+gzip
```

The old mediatype used was `application/tar+gz` (or in some cases `application/octet-stream`).

In order to prevent breaking people who have published old charts, this strict validation should be lifted, making the assumption that anything with 1 or 2 layers is valid, and that they are assumed to be the chart and prov layer respectively.

Labelling as "bug" since this is causing some issues for people.

LochanRn:
OCI: Allow any mediatype for layers on download
fixes #10176 

**If applicable**:

### DIFF
--- a/internal/experimental/registry/client.go
+++ b/internal/experimental/registry/client.go
@@ -224,7 +224,7 @@ func (c *Client) Pull(ref string, options ...PullOption) (*PullResult, error) {
 	minNumDescriptors := 1 // 1 for the config
 	if operation.withChart {
 		minNumDescriptors++
-		allowedMediaTypes = append(allowedMediaTypes, ChartLayerMediaType)
+		allowedMediaTypes = append(allowedMediaTypes, ChartLayerMediaType, LegacyChartLayerMediaType)
 	}
 	if operation.withProv {
 		if !operation.ignoreMissingProv {
@@ -256,6 +256,9 @@ func (c *Client) Pull(ref string, options ...PullOption) (*PullResult, error) {
 			chartDescriptor = &d
 		case ProvLayerMediaType:
 			provDescriptor = &d
+		case LegacyChartLayerMediaType:
+			chartDescriptor = &d
+			fmt.Fprintf(c.out, "Warning: chart media type %s is deprecated\n", LegacyChartLayerMediaType)
 		}
 	}
 	if configDescriptor == nil {
--- a/internal/experimental/registry/constants.go
+++ b/internal/experimental/registry/constants.go
@@ -31,4 +31,7 @@ const (
 
 	// ProvLayerMediaType is the reserved media type for Helm chart provenance files
 	ProvLayerMediaType = "application/vnd.cncf.helm.chart.provenance.v1.prov"
+
+	// LegacyChartLayerMediaType is the legacy reserved media type for Helm chart package content.
+	LegacyChartLayerMediaType = "application/tar+gzip"
 )


LochanRn:
@jdolitsky @bacongobbler  can you please review this PR :)

internal/experimental/registry/constants.go
```diff
@@ -31,4 +31,7 @@ const (
 
 	// ProvLayerMediaType is the reserved media type for Helm chart provenance files
 	ProvLayerMediaType = "application/vnd.cncf.helm.chart.provenance.v1.prov"
+
+	// LegacyChartContentLayerMediaType is the legacy reserved media type for Helm chart package content.
+	LegacyChartContentLayerMediaType = "application/tar+gzip"
```
jdolitsky, line 36: Can we name this `LegacyChartLayerMediaType` to match `ChartLayerMediaType`?

internal/experimental/registry/client.go
```diff
@@ -220,6 +220,7 @@ func (c *Client) Pull(ref string, options ...PullOption) (*PullResult, error) {
 	store := content.NewMemoryStore()
 	allowedMediaTypes := []string{
 		ConfigMediaType,
+		LegacyChartContentLayerMediaType,
```
jdolitsky, line 223: This should be appended under `if operation.withChart {` below
```
allowedMediaTypes = append(allowedMediaTypes, ChartLayerMediaType, LegacyChartLayerMediaType)
```

internal/experimental/registry/client.go
```diff
@@ -252,7 +253,7 @@ func (c *Client) Pull(ref string, options ...PullOption) (*PullResult, error) {
 		switch d.MediaType {
 		case ConfigMediaType:
 			configDescriptor = &d
-		case ChartLayerMediaType:
+		case ChartLayerMediaType, LegacyChartContentLayerMediaType:
```
jdolitsky, line 256: It might be better to put this into its own `case`, so that we can add some log warning that the chart is using a legacy type:
```
fmt.Fprintf(c.out, "Warning: chart media type %s is deprecated\n", LegacyChartLayerMediaType)

**Review by jdolitsky — CHANGES_REQUESTED** [review]
Thanks so much for looking at this! Left a few comments

File: internal/experimental/registry/constants.go (hunk 1 above)
Comment (LochanRn) [reply internal/experimental/registry/constants.go:36]: ack

File: internal/experimental/registry/client.go (hunk 2 above)
Comment (LochanRn) [reply internal/experimental/registry/client.go:223]: ack

File: internal/experimental/registry/client.go (hunk 3 above)
Comment (LochanRn) [reply internal/experimental/registry/client.go:256]: ack

**Comment by SteveLasker** [comment]
Thanks @LochanRn, @jdolitsky this is great to see customers that took a bet on the `helm registry` experimental feature have their investments supported with the current release. Now, users will no longer have to re-pull/push their charts.

**Review by jdolitsky — APPROVED** [review]
Tested this manually and works as expected:
```
$ bin/helm pull oci://ghcr.io/mattfarina/test-chart --version 0.1.0
Warning: chart media type application/tar+gzip is deprecated
Pulled: ghcr.io/mattfarina/test-chart:0.1.0
Digest: sha256:14dc2026d12272575758984e807405e6ac303c627b10a17817ae56bb811c314b
```

cc @mattfarina

cndoit18 (approved):
LGTM

bacongobbler:
I thought one of the goals of a "media type allow list" was to prevent malicious clients from injecting unknown binary blobs in the metadata. Why is Helm changing the media type, then going right back to accepting the old media type?

With this change, an attacker performing a MITM attack could inject additional metadata into the response body, which then the client will silently accept and download onto the user's machine. This could lead to further security issues, or at worst a RCE attack on the user's machine.

If the goal is to support "older" OCI charts, shouldn't they be versioned? That way we can determine whether we're attempting to download a "v1" OCI chart or a "v2" OCI chart and apply the appropriate media type allow list?

On a related note... I don't understand why we need to provide support to prior versions of charts for an experimental flag. It is well understood that an experimental feature is meant to break backwards compatibility and users should understand what that entails. Because at best, this provides "legacy" support for a small number of users running experimental features in a production scenario (which should NOT be happening), and at worst it could lead to future security issues.

Thoughts?

mattfarina:
@bacongobbler ...

> Why is Helm changing the media type, then going right back to accepting tar+gzip media?

We discussed it on the dev call just over a week ago.

>  I don't understand why we need to provide support to prior versions of charts for an experimental flag.

In theory we shouldn't have to provide support. It's an experiment, right? If only...

>  It is well understood that an experimental feature is meant to break backwards compatibility and users should understand what that entails.

This is not well understood. People have been using the experiment as if it were production. _In some cases companies have told their customers to use it._ I some cases open source projects have used it and told their users to install this way. Without understanding it's an experiment that may make breaking changes, people have been just using it.

That means we can make one of two choices:

1. Break the API compatibility as we said might happen and a bunch of people will be mad at Helm. Note, there are some who will be mad at Helm rather than the people who told them to use an experiment.
2. Keep the compatibility which means a little more to deal with and go back on the possibility of breaking the API there. This will make Helm users happy.

The way this whole situation went down makes me consider experiments a failure. People who surround Helm treated it as if it were production ready. They misused our good will. I'm now in the camp who thinks we should do no more experiments and that all of them should happen through plugins. If there isn't a pluggable location for something then the feature to add is to make plugins more capable. If an experiment via a plugin goes mainstream then we can look to fold it into Helm proper.

> With this change, an attacker performing a MITM attack could inject additional metadata into the response body...

I'd love to talk more about this. I don't think we should take a media type as the source of truth. An attacker could use the Helm media type. We need to make sure the package we download really is a chart as we expect it to be before working with it. The media type is suggested by the server and shouldn't be regarded as trusted.

> Thoughts?

By making a breaking change we just broke the experience for a bunch of people. The breaking change is something we communicated we'd do. People aren't paying attention to that. Others are telling people to just use it. I can't figure out a way we (the Helm project) don't look bad for Helm users by making this breaking change. Even if we do exactly what we said. It's frustrating.

jdolitsky:
Regarding MITM attacks, I'm having trouble understanding how the mediatype affects the actual content being unpacked... the attack vector seems to be identical if the mediatype is the new one vs. old one.

As far as experiments, I think this is a whole other discussion. If this was done as a plugin, I'm feeling strongly the adoption would not be where it is today. But I do agree it created a misunderstanding of "experimental".

---

So do we want to support this? Depending on how you look at it, #10176 could be considered a bug (or not).

mattfarina (approved):
Connected with @bacongobbler on any possible issues. After chatting with him it looks like we can merge this.

PR #10201, opened 2021-09-30
Add more error checks during the signing process

SgtCoDFish:
**What this PR does / why we need it**: Before this change, several of the potential errors during the process of signing a package were skipped.

Crucially, `Close()`ing the ReadCloser from the gpg clearsigner is the call which actually does the signing, and so has several points of failure which are ignored; for example, if there's a problem with the format of the key.

Also changes the error from `messageBlock` to be propagated rather than being swallowed, and adds a unit test which fails when run against the `main` branch currently. 

**Special notes for your reviewer**: `io.Copy` is documented to not return EOF, so there's no need to check for that case when checking its error.

I guess returning these errors is, in a way, backwards-incompatible - a workflow today could be silently failing to sign and returning `"", nil` which wouldn't be reported as an error. I'm not sure how useful such a workflow would be though since the generated "signature" would be an empty file. In such a case I'd want `helm` to start loudly failing to sign.

### DIFF
--- a/pkg/provenance/sign.go
+++ b/pkg/provenance/sign.go
@@ -216,7 +216,7 @@ func (s *Signatory) ClearSign(chartpath string) (string, error) {
 
 	b, err := messageBlock(chartpath)
 	if err != nil {
-		return "", nil
+		return "", err
 	}
 
 	// Sign the buffer
@@ -224,9 +224,24 @@ func (s *Signatory) ClearSign(chartpath string) (string, error) {
 	if err != nil {
 		return "", err
 	}
+
 	_, err = io.Copy(w, b)
-	w.Close()
-	return out.String(), err
+
+	if err != nil {
+		// NB: We intentionally don't call `w.Close()` here! `w.Close()` is the method which
+		// actually does the PGP signing, and therefore is the part which uses the private key.
+		// In other words, if we call Close here, there's a risk that there's an attempt to use the
+		// private key to sign garbage data (since we know that io.Copy failed, `w` won't contain
+		// anything useful).
+		return "", errors.Wrap(err, "failed to write to clearsign encoder")
+	}
+
+	err = w.Close()
+	if err != nil {
+		return "", errors.Wrap(err, "failed to either sign or armor message block")
+	}
+
+	return out.String(), nil
 }
 
 // Verify checks a signature and verifies that it is legit for a chart.
--- a/pkg/provenance/sign_test.go
+++ b/pkg/provenance/sign_test.go
@@ -16,6 +16,9 @@ limitations under the License.
 package provenance
 
 import (
+	"crypto"
+	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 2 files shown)


bacongobbler (approved):

pkg/provenance/sign.go
```diff
@@ -216,17 +216,26 @@ func (s *Signatory) ClearSign(chartpath string) (string, error) {
...
 	if err != nil {
 		return "", err
 	}
+
 	_, err = io.Copy(w, b)
-	w.Close()
-	return out.String(), err
+	if err != nil {
+		return "", errors.Wrap(err, "failed to write to clearsign encoder")
+	}
```
hickeyma, line 231: `Encode` returns a WriteCloser which means if there is an error here, it looks like `Close()` may not be called?

Just wondering if it would be better to `defer w.Close()` and not check error below or is there value from the error code on the `Close()`?

SgtCoDFish: I think there are two things here:

First, the error code from `Close()` is _absolutely critical_, and it's the core reason I raised the PR. [Encode](https://cs.opensource.google/go/x/crypto/+/ae814b36:openpgp/clearsign/clearsign.go;l=331-402) can fail in multiple ways but it doesn't actually do any "real" work. The WriteCloser it returns actually does the signing in [Close](https://cs.opensource.google/go/x/crypto/+/ae814b36:openpgp/clearsign/clearsign.go;l=293-329). Specifically, the call to Sign is [here](https://cs.opensource.google/go/x/crypto/+/ae814b36:openpgp/clearsign/clearsign.go;l=314).

We could handle the error inside a defer but that'll get messy because our `Signatory.ClearSign` function should fail if the signing fails.

That said, you're absolutely right that this introduces a case where we'll fail to call `Close()` if `io.Copy` fails. My reading of the [Close](https://cs.opensource.google/go/x/crypto/+/ae814b36:openpgp/clearsign/clearsign.go;l=293-329) method is that failing to `Close` isn't an issue in this case.

Usually, my concern with failing to `Close` would be that it could cause a resource leak but I couldn't see anything in this _specific_ Close method which would leak. Calling Close could actually be harmful - because Close tries to do the actual signing, it could try to sign whatever garbage data is left in `w` after io.Copy failed, and ideally we don't want to attempt to generate signatures on anything in this instance since the buffer will only contain garbage.

Ultimately this overloading of Close to mean both "Close" and "Sign" is I think a failing of the design of the go openpgp library (which is probably one of the several reasons it's deprecated :grin:).

I think there are two possible solutions here:

1. Write a comment explaining why we don't `Close` here, because it's totally counterintuitive to _intentionally_ not Close
2. If `io.Copy` returns `err != nil`, then call `_ = w.Close()` before we return, ignoring any errors from it.

I don't really love option 2 because - as mentioned above - it introduces the risk of touching the private key needlessly, but I'd happily do it if you disagree! I'll do option 1 for now and push that.

SgtCoDFish:
Is there anything else needed on this PR? Let me know if so! :grin:

pkg/provenance/sign.go
hickeyma: Sounds good.

hickeyma (approved):

Issue #10163, opened 2021-09-21
Helm does not wait for Statefulset to be ready after upgrade


libsamek:
Helm does not wait for Statefulset to be ready after upgrade to 3.6.3 from 3.3.1. I have tried using versions:
* 3.7.0 (does not work),
* 3.6.3 (does not work),
* 3.5.4 (works!),
* 3.3.1 (works!).

It seems that this behavior was introduced with: https://github.com/helm/helm/commit/98d98369ed9abcaec9b75352e07dd6f759a8804f

Helm 3.7.0 behavior - unexpected:

```
helm upgrade --install --reset-values XXX chart --version X.Y.Z --wait --timeout 900s --create-namespace --namespace dummy --values /tmp/helmfile351904154/XXX-values-6fd57cf55 ... --history-max 10 --debug
helm:whTHc> history.go:56: [debug] getting history for release XXX
helm:whTHc> upgrade.go:139: [debug] preparing upgrade for XXX
helm:whTHc> upgrade.go:488: [debug] resetting values to the chart's original version
helm:whTHc> upgrade.go:147: [debug] performing update for XXX
helm:whTHc> upgrade.go:319: [debug] creating upgraded release for XXX
helm:whTHc> client.go:218: [debug] checking 4 resources for changes
helm:whTHc> client.go:501: [debug] Looks like there are no changes for ConfigMap "XXX"
helm:whTHc> client.go:501: [debug] Looks like there are no changes for Service "XXX"
helm:whTHc> wait.go:48: [debug] beginning wait for 4 resources with timeout of 15m0s
helm:whTHc> upgrade.go:154: [debug] updating status for upgraded release for XXX
helm:whTHc> Release "XXX" has been upgraded. Happy Helming!
NAME: XXX
...
```

Helm 3.5.4 behavior - expected:
```
exec: helm upgrade --install --reset-values XXX chart --version X.Y.Z --wait --timeout 900s --create-namespace --namespace dummy --values /tmp/helmfile549477625/XXX-values-68c7579fb5 ... --history-max 10 --debug
helm:whTHc> history.go:56: [debug] getting history for release XXX
helm:whTHc> upgrade.go:123: [debug] preparing upgrade for XXX
helm:whTHc> upgrade.go:437: [debug] resetting values to the chart's original version
helm:whTHc> upgrade.go:131: [debug] performing update for XXX
helm:whTHc> upgrade.go:303: [debug] creating upgraded release for XXX
[...]

PR #10225, opened 2021-10-07
fix: improve logging & safety of statefulSetReady

dnwe:

This isn't necessarily a fix, but the additional logging may help us in
determining what the underlying issue is and it seems sensible to
confirm that the current and updated revision numbers also match as part
of the readiness check.

Contributes-to: #10163

### DIFF
--- a/pkg/action/upgrade.go
+++ b/pkg/action/upgrade.go
@@ -380,6 +380,9 @@ func (u *Upgrade) releasingUpgrade(c chan<- resultMessage, upgradedRelease *rele
 	}
 
 	if u.Wait {
+		u.cfg.Log(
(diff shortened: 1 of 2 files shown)


dnwe:
@mattfarina / @bacongobbler any thoughts on this? It appears to have solved the issue for the OP and also ourselves

jonny08152 (commented):
This fix works perfectly. Waiting for StatefulSet rollout to finish now works as it should.

dnwe:
FY II closed and re-opened this as [#10622](/helm/helm/pull/10622) in January to try and get the PR at the top of the incoming queue, but still haven't had a review sadly

PR #10230, opened 2021-10-08
add SetNamespace method for EnvSettings, in order to set namespace in helm sdk

### LINKED ISSUE #10215: Recreate flag does not work for rollback in the SDK
This is about the Helm Go SDK. I'm attempting a rollback programmatically. 

Sample code 
```
	rollback := action.NewRollback(actionConfig)
	rollback.Version = targetV
	rollback.Recreate = true
	rollback.Wait = true
	rollback.WaitForJobs = true
	rollback.Timeout = time.Duration(5) * time.Minute
```
The Recreate flag is set to true, the chart gets rolled back to the previous version (targetV is defined previously) but the pods do not get recreated. 

However, the flag --recreate-pods flag works as expected in the helm rollback CLI command. I thought the CLI tool also uses the same codebase, so the behaviour should be identical.

What am I doing wrong?

yxxhero:
add SetNamespace method for EnvSettings, in order to set namespace in helm sdk

fix https://github.com/helm/helm/issues/10215
add SetNamespace method for EnvSettings, in order to set namespace in helm sdk
**If applicable**:

### DIFF
--- a/pkg/cli/environment.go
+++ b/pkg/cli/environment.go
@@ -180,6 +180,11 @@ func (s *EnvSettings) Namespace() string {
 	return "default"
 }
 
+// SetNamespace sets the namespace in the configuration
+func (s *EnvSettings) SetNamespace(namespace string) {
+	s.namespace = namespace
+}
+
 // RESTClientGetter gets the kubeconfig from EnvSettings
 func (s *EnvSettings) RESTClientGetter() genericclioptions.RESTClientGetter {
 	return s.config
--- a/pkg/cli/environment_test.go
+++ b/pkg/cli/environment_test.go
@@ -25,6 +25,20 @@ import (
 	"github.com/spf13/pflag"
 )
 
+func TestSetNamespace(t *testing.T) {
+	settings := New()
+
+	if settings.namespace != "" {
+		t.Errorf("Expected empty namespace, got %s", settings.namespace)
+	}
+
+	settings.SetNamespace("testns")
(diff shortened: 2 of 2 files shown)


zonggen:
Tried change with linked issue, still does not seem to work without explicitly calling `os.Setenv("HELM_NAMESPACE", "testchart")`. Does this require a call to `EnvSettings.SetNamespace()`?

The `action.Configuration.Init()` call initiliazes the secrets but I don't see a namespace is set:
https://github.com/helm/helm/blob/932d2a98a2a01c069aaefb25b12c1ece1f1879d5/pkg/action/action.go#L376

To fix this issue do we need to set the namespace in "secret" case of the `action.Configuration.Init()`, like in the "memory"?
https://github.com/helm/helm/blob/932d2a98a2a01c069aaefb25b12c1ece1f1879d5/pkg/action/action.go#L397

yxxhero:
@zonggen Thanks for your verification.

cndoit18 (approved):

bacongobbler:
Why not make the namespace field public? Why do we need to introduce getters (`.Namespace()`) and setters (`.GetNamespace()`)?

See the conversation in https://github.com/helm/helm/pull/10148.

bacongobbler (approved):
Meh. I don't particularly appreciate the package design, but this is probably the best we can do until Helm 4, and this gives users the option to set the namespace as the tests suggest.

LGTM

PR #10244, opened 2021-10-18
fix: rollback failure after update error

### LINKED ISSUE #10219: Unable to roll back as release left in invalid state
Hello, 
In our pipelines, we auto rollback on the next release if the previous deployment failed.  We do not use `--atomic` because we want to allow people time to investigate the failure by looking at the releases.

In some circumstances, we've seen helm fail to rollback with an error like: `Error: no Example with the name "example2" found`.  I've struggled to recreate it, but now believe I have the scenario down - it seems to happen when you have a failed release which meets two conditions:

- A resource in the release was `DELETED`
- Another resource in the release was `MODIFIED`, however it's modification was rejected by Kubernetes validation

I've created https://github.com/Stono/helm-bug which contains the example required to replicate the bug - the simplest example was using a crd but it'd be the same for any resource, but the steps are:

- kubectl apply -f crd.yaml
- helm upgrade example --install . 
```
❯ helm upgrade example --install . 
Release "example" does not exist. Installing it now.
NAME: example
LAST DEPLOYED: Wed Oct  6 10:43:23 2021
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
```

- edit [templates/instance1.yaml](templates/instance1.yaml) to be `spec.name=invalid`, and remove [templates/instance2.yaml](templates/instance2.yaml) from the chart
- helm upgrade example --install . 

```
❯ helm upgrade example --install . 
Error: UPGRADE FAILED: cannot patch "example1" with kind Example: Example.stono.io "example1" is invalid: spec.name: Invalid value: "": spec.name in body should match '^valid$'
```

- Attempt to (but unable to) roll back to previous release:
```
❯ helm rollback example 1
Error: no Example with the name "example2" found
```

Which seems to come from here: https://github.com/helm/helm/blob/2062a59566056f4900c139b6425ceb66df893177/pkg/kube/client.go#L246

As you see, you end up in a state that you cannot rollback.

Helm version: v3.6.2
Kubernetes version: v1.19.13

mmulholla:

closes #10219

Helm update stops after processing changes if there was an error, this as the effect the deletes are skipped. Now when rollback is requested it can fails because the delete was not done.

The fix moves the logic to process errors from after processing changes to after processing changes and deletes.

I ran the go test against the fix and everything passed. Was not sure how to add a test for this scenario - if a test os required please provide some guidance on how best to do this.

**If applicable**:

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -256,13 +256,6 @@ func (c *Client) Update(original, target ResourceList, force bool) (*Result, err
 		return nil
 	})
 
-	switch {
-	case err != nil:
-		return res, err
-	case len(updateErrors) != 0:
-		return res, errors.Errorf(strings.Join(updateErrors, " && "))
-	}
-
 	for _, info := range original.Difference(target) {
 		c.Log("Deleting %q in %s...", info.Name, info.Namespace)
 
@@ -283,6 +276,13 @@ func (c *Client) Update(original, target ResourceList, force bool) (*Result, err
 			continue
 		}
 		res.Deleted = append(res.Deleted, info)
+	}
+
+	switch {
+	case err != nil:
+		return res, err
+	case len(updateErrors) != 0:
+		return res, errors.Errorf(strings.Join(updateErrors, " && "))
 	}
 	return res, nil
 }


bacongobbler:
I think I understand how this resolves the issue. However it does concern me that in order to fix this, Helm will now attempt to delete certain objects before reporting back an error that occurred during the update/create stage.

We will need to be very careful with this PR. If this doesn't work as intended in all cases (specifically thinking about `helm upgrade` error cases), this has the potential to delete resources before the user has had the chance to resolve the failed update.

For example, looking at @Stono's error case provided in https://github.com/helm/helm/issues/10219:

> it seems to happen when you have a failed release which meets two conditions:

>    A resource in the release was DELETED
>    Another resource in the release was MODIFIED, however it's modification was rejected by Kubernetes validation

Prior to this change, a `helm upgrade` would not delete `templates/instance2.yaml`. But now it will.

I think this has significant impact to those cases and more testing from the community is necessary to better understand the implications of this change.

bacongobbler:
@mmulholla have you had the chance to look further into my last comment?

yxxhero:
@mmulholla ping

PR #10248, opened 2021-10-19
Resolve #10212 Change the error message for `helm show` to include --devel option in the error

### LINKED ISSUE #10212: Feature request: improve error message when no official releases are published and --devel is not set
Output of `helm version`: v3.7.0"

Currently, if you try to interact with a helm chart that does not have a official release set, you will get an error message that is a bit misleading:
```
$ helm show readme foo/bar
Error: failed to download "foo/bar"
```

Adding `--devel` fixes it, but its not very discoverable.

It would be great if it was changed to something like
```
$ helm show readme foo/bar
Error: failed to find a suitable version of "foo/bar", but a development version was available. Use --devel to use it.
```

anshulvermapatel:
Added a new error message which tell to use `--devel` option.

**If applicable**:

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -766,5 +766,5 @@ func (c *ChartPathOptions) LocateChart(name string, settings *cli.EnvSettings) (
 		atVersion = fmt.Sprintf(" at version %q", version)
 	}
 
-	return filename, errors.Errorf("failed to download %q%s", name, atVersion)
+	return filename, errors.Errorf("failed to download %q%s. Use --devel for a development version", name, atVersion)
 }
--- a/cmd/helm/show_test.go
+++ b/cmd/helm/show_test.go
@@ -19,7 +19,6 @@ package main
 import (
 	"fmt"
 	"path/filepath"
-	"strings"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


bacongobbler:
ping.

pkg/action/install.go
```diff
@@ -766,5 +766,5 @@ func (c *ChartPathOptions) LocateChart(name string, settings *cli.EnvSettings) (
 		atVersion = fmt.Sprintf(" at version %q", version)
 	}
 
-	return filename, errors.Errorf("failed to download %q%s", name, atVersion)
+	return filename, errors.Errorf("failed to download %q%s. Use --devel for a development version", name, atVersion)
```
bacongobbler, line 769: With a few exceptions, we try to avoid avoid referring to Helm CLI commands in `pkg/action`. This is because `pkg/action` can be used by other programs as a library, so asking the user to run certain commands or feature flags may be completely irrelevant to the program.

bacongobbler (changes requested):

pkg/action/install.go
anshulvermapatel: Hi, Okay .. I will do that and update the PR

PR #10255, opened 2021-10-20
Fix ResourceList Difference with different GroupVersionKind objects

yalosev:
ResourceList `Difference(x)` function compares only Kind but not GroupVersionKind (
although it is said in the comment to the function).  It works for an upgrade of resources with different Kinds. But it leads to a wrong result while upgrading a release with resources that have different Group/Version but the same Kind.

The real case of this bug is when you are upgrading `certmanager.k8s.io` Certificates to the new `cert-manager.io` version

**If applicable**:


zonggen:
Thank you for submitting the patch! Can you link the specific issue to this PR, if one does not already exist can you create one? Please refer to https://github.com/helm/helm/blob/main/CONTRIBUTING.md#how-to-contribute-a-patch for contributing a patch. This way the issue will get triaged and raise more attention.

PR #10269, opened 2021-10-23
Add security features to test template file

duhow:

This PR adds more Pod settings to the default template `test-connection` Pod. Fixes suggested changes by [KubeLinter](https://github.com/stackrox/kube-linter):

```
templates/tests/test-connection.yaml: (object: <no namespace>/test-release-example-test-connection /v1, Kind=Pod) container "wget" does not have a read-only root file system (check: no-read-only-root-fs, remediation: Set readOnlyRootFilesystem to true in the container securityContext.)
templates/tests/test-connection.yaml: (object: <no namespace>/test-release-example-test-connection /v1, Kind=Pod) container "wget" is not set to runAsNonRoot (check: run-as-non-root, remediation: Set runAsUser to a non-zero number and runAsNonRoot to true in your pod or container securityContext. Refer to https://kubernetes.io/docs/tasks/configure-pod-container/security-context/ for details.)
templates/tests/test-connection.yaml: (object: <no namespace>/test-release-example-test-connection /v1, Kind=Pod) container "wget" has cpu request 0 (check: unset-cpu-requirements, remediation: Set CPU requests and limits for your container based on its requirements. Refer to https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits for details.)
templates/tests/test-connection.yaml: (object: <no namespace>/test-release-example-test-connection /v1, Kind=Pod) container "wget" has cpu limit 0 (check: unset-cpu-requirements, remediation: Set CPU requests and limits for your container based on its requirements. Refer to https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits for details.)
templates/tests/test-connection.yaml: (object: <no namespace>/test-release-example-test-connection /v1, Kind=Pod) container "wget" has memory request 0 (check: unset-memory-requirements, remediation: Set memory requests and limits for your container based on its requirements. Refer to https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits for details.)
templates/tests/test-connection.yaml: (object: <no namespace>/test-release-example-test-connection /v1, Kind=Pod) container "wget" has memory limit 0 (check: unset-memory-requirements, remediation: Set memory requests and limits for your container based on its requirements. Refer to https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits for details.)
```

N/A

**If applicable**:

### DIFF
--- a/pkg/chartutil/create.go
+++ b/pkg/chartutil/create.go
@@ -500,7 +500,18 @@ spec:
     - name: wget
       image: busybox
       command: ['wget']
-      args: ['{{ include "<CHARTNAME>.fullname" . }}:{{ .Values.service.port }}']
+      args: ['-O', '/dev/null', '{{ include "<CHARTNAME>.fullname" . }}:{{ .Values.service.port }}']
+      securityContext:
+        runAsUser: 1000
+        runAsNonRoot: true
+        readOnlyRootFilesystem: true
+      resources:
+        limits:
+          memory: 60Mi
+          cpu: 200m
+        requests:
+          memory: 10Mi
+          cpu: 10m
   restartPolicy: Never
 `
 


pkg/chartutil/create.go
```diff
@@ -500,7 +500,18 @@ spec:
...
+        runAsUser: 1000
+        runAsNonRoot: true
+        readOnlyRootFilesystem: true
+      resources:
+        limits:
+          memory: 60Mi
+          cpu: 200m
+        requests:
+          memory: 10Mi
+          cpu: 10m
```
mattfarina, line 514: We usually recommend setting the resources to be opt-in. The reason for this is that they can vary widely. For example, take a PostgreSQL instance. What it needs depends on how much data or connections are being made to it. This is difficult for the chart author to determine.

Why set the resources here and if it's needed, is there a better way to handle this for the application operator (chart installer)?

PR #10293, opened 2021-10-29
[helmignore] Support for '!' as a special leading sequence

### LINKED ISSUE #3622: Whitelisting in .helmignore with '/*' returns "chart metadata (Chart.yaml) missing"
Hi guys,

I am trying a whitelist approach with the following .helmignore

```
/*
!.helmignore
!Chart.yaml
!templates
!values.yaml
```

Unfortunately this gives me the following error:
```
Error: chart metadata (Chart.yaml) missing
```

Is this the desired behavior or am I missing something?

helm/tiller version:
```
$ helm version
Client: &version.Version{SemVer:"v2.8.1", GitCommit:"6af75a8fd72e2aa18a2b278cfe5c7a1c5feca7f2", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.8.1", GitCommit:"6af75a8fd72e2aa18a2b278cfe5c7a1c5feca7f2", GitTreeState:"clean"}
```

taliastocks:
.helmignore has "supported" negative matches uselessly for a long time, reported as an open ticket (https://github.com/helm/helm/issues/3622) as early as 2018. This revision implements negative matching correctly and usefully for the first time.

Closes #3622.

**If applicable**:

### DIFF
--- a/internal/ignore/doc.go
+++ b/internal/ignore/doc.go
@@ -62,6 +62,5 @@ Notable differences from .gitignore:
 	- The globbing library is Go's 'filepath.Match', not fnmatch(3)
 	- Trailing spaces are always ignored (there is no supported escape sequence)
 	- The evaluation of escape sequences has not been tested for compatibility
-	- There is no support for '\!' as a special leading sequence.
 */
 package ignore // import "helm.sh/helm/v3/internal/ignore"
--- a/internal/ignore/rules.go
+++ b/internal/ignore/rules.go
@@ -100,22 +100,30 @@ func (r *Rules) Ignore(path string, fi os.FileInfo) bool {
 	if path == "." || path == "./" {
 		return false
 	}
+
+	// Negative rules should only override positive ignore rules, rather
+	// than exclude everything aside from the pattern (e.g. Chart.yaml).
+	// See issue:
+	// 3622 Whitelisting in .helmignore with '/*' returns "chart metadata (Chart.yaml) missing"
+	positiveMatch := false
+	negativeMatch := false
+
 	for _, p := range r.patterns {
 		if p.match == nil {
 			log.Printf("ignore: no matcher supplied for %q", p.raw)
 			return false
 		}
 
-		// For negative rules, we need to capture and return non-matches,
-		// and continue for matches.
+		// For negative rules, we need to continue for non-matches,
+		// and record matches.
 		if p.negate {
 			if p.mustDir && !fi.IsDir() {
-				return true
+				continue
 			}
 			if !p.match(path, fi) {
-				return true
+				continue
 			}
-			continue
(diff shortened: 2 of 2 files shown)


bacongobbler:
Hi @taliastocks! Sorry for not seeing this sooner.

Would you mind writing unit tests covering these changes? This seems like something we can a test suite to check against expected input/outputs. That way we can also check against existing .hemlignores to ensure we're not accidentally breaking some charts.

taliastocks:
Hi @bacongobbler,

Thanks for taking a look at this. As it stands, I unfortunately can't dedicate more time to this, as it's no longer a blocker for my team (not to mention, I'm about to be on leave for three months...). We've instead opted to use the `.Files` API to solve our original problem.

That said, if there's anyone in the community who would like to take this work from here, please feel free to commandeer. A couple notes that might come in handy, copied from my post to the bug report thread:

> I think we can avoid breaking _most_ compatibility due to the fact that this feature, as written, is nearly useless:
> 
> * As written, all rules must match `/Chart.yaml`, otherwise the helm chart stops working entirely.
> * This means, more or less, that the only useful negative rule is `!*.yaml`.
> 
> To maintain _most_ backwards compatibility, we could treat a set of rules that, aside from directory rules, only contain negatives as implicitly including `*` as a positive rule:
> 
> * `!*.yaml` would continue to exclude all non-yaml files when included on its own in a `.helmignore`.
> * Directory rules would continue to work as-is due to the above exception, and files in these directories would not be included because excluded directories are not traversed.
> 
> This isn't perfect, however, because some users may currently include redundant positive rules alongside `!*.yaml`, for example `README.md`. The presence of these rules doesn't currently add anything to the `!*.yaml` negative rule, but nonetheless they may appear in the wild.

Note that not all of the above backwards compatibility is implemented yet in this branch -- only the standard `.gitignore` behavior.

Regards,
Talia

evankanderson:
Hey, I just hit this as well.  I'm looking to use https://ko.build/ to build images in our helm chart by running something like the following:

```bash
rm templates/combined.yaml; ko resolve --base-import-paths -f templates/ > templates/combined.yaml
```

And then have helm ignore all the files _except_ `combined.yaml`.  This sorta works (you need to finesse `ko`'s YAML parser for helm templates), but the simple `.helmignore` you'd want to write ends up excluding `Chart.yaml`:

```
!templates/combined.yaml
templates/*.yaml
```

Even worse, it appears that combining two negative patterns limits the included files to only those that match _both_ patterns.

joejulian:
@taliastocks Could you please rebase?

evankanderson:
This is 2 years old, but https://github.com/helm/helm/pull/12265 might be a suitable replacement if you'd like to look at it

PR #10297, opened 2021-10-30
WIP Fix dependency condition is always true when it's checking a value that doesn't exist

### LINKED ISSUE #10296: Helm dependency condition is always true when it's checking a value that doesn't exist
For each dependency in `Chart.yml`, the condition will be considered as `true` if it is checking a value that doesn't exist.

For this example:
```
...

dependencies:
  - name: subchart
    version: 0.1.0
    condition: subchart.enabled

...
```
If `subchart.enabled` is missing in `values.yml`, the `subchart` will be enabled by default which is considered as an unexpected behavior.

sincejune:
Signed-off-by: Chao <oleooo@126.com>

For each dependency in `Chart.yml`, the condition will be considered as `true` if it is checking a value that doesn't exist.

For this example:
```
...

dependencies:
  - name: subchart
    version: 0.1.0
    condition: subchart.enabled

...
```
If `subchart.enabled` is missing in `values.yml`, the `subchart` will be enabled by default which is considered as an unexpected behavior.

This PR fixes the issue, closes #10296

**If applicable**:

### DIFF
--- a/cmd/helm/testdata/testcharts/subchart/values.yaml
+++ b/cmd/helm/testdata/testcharts/subchart/values.yaml
@@ -53,3 +53,9 @@ exports:
       SC1exported2:
         all:
           SC1exported3: "SC1expstr"
+
+subcharta:
+  enabled: true
+
+subchartb:
+  enabled: true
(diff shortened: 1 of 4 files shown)


sincejune:
Seems that some unit tests failed in CircleCI, but those are running fine at my local. Does anyone know the reason?

Issue #4789, opened 2018-10-16
How to avoid translate some `{{` of the charts?


cheyang:
Output of `helm version`:
v2.9.1
Output of `kubectl version`:
v1.11.2
Cloud Provider/Platform (AKS, GKE, Minikube etc.): 
Baremetal

I want to put the following CRD into helm chart, but it contains go raw template. How to make helm not translate `{{` and `}}` inside `rawTemplate`. Thanks for your response.
https://github.com/kubeflow/katib/blob/master/examples/random-example.yaml

```
apiVersion: "kubeflow.org/v1alpha1"
kind: StudyJob
metadata:
  namespace: katib
  labels:
    controller-tools.k8s.io: "1.0"
  name: random-example
spec:
  studyName: random-example
  owner: crd
  optimizationtype: maximize
  objectivevaluename: Validation-accuracy
  optimizationgoal: 0.99
  requestcount: 4
  metricsnames:
    - accuracy
  parameterconfigs:
    - name: --lr
      parametertype: double
      feasible:
        min: "0.01"
        max: "0.03"
    - name: --num-layers
      parametertype: int
      feasible:
        min: "2"
        max: "5"
    - name: --optimizer
      parametertype: categorical
      feasible:
        list:
        - sgd
        - adam
        - ftrl
  workerSpec:
    goTemplate:
        rawTemplate: |-
          apiVersion: batch/v1
          kind: Job
          metadata:
            name: {{.WorkerId}}
            namespace: katib
          spec:
            template:
              spec:
                containers:
                - name: {{.WorkerId}}
                  image: katib/mxnet-mnist-example
                  command:
                  - "python"
                  - "/mxnet/example/image-classification/train_mnist.py"
                  - "--batch-size=64"
                  {{- with .HyperParameters}}
                  {{- range .}}
                  - "{{.Name}}={{.Value}}"
                  {{- end}}
                  {{- end}}
                restartPolicy: Never
  suggestionSpec:
    suggestionAlgorithm: "random"
    requestNumber: 3
```

mortent:
I can think of two ways to work around this:

Update the controller that handles the `StudyJob` resource to use a different delimiter with the templating engine: [https://golang.org/pkg/html/template/#Template.Delims](https://golang.org/pkg/html/template/#Template.Delims). This should work pretty well, but require that you can either update the StudyJob Controller or that it already allows you to override the template delimiters.

[...]

Issue #2931, opened 2017-09-14
Proposal: Add template action delimiters parameter


dhoeld:
# Problem
We want to template some files e.x. Prometheus Alertmanager alert rules with helm. When trying to do do this, there is an issue:
The templating language used by the Prometheus Alertmanager is go-templating. Therefore there is a clash in the use of the standard template delimiters {{ }} when templating alert rules.

# Questions
From my knowledge level there are two possibilities:

- escaping the template delimiters within the alert rules with ``{{ `{{ ALERT_RULE_VARIABLE `}} }}`` which seems quite hacky
- extending the render package of helm to support using other delimiters than the default ones. E.x. `[[ ALERT_RULE_VARIABLE ]]` instead of `{{ ALERT_RULE_VARIABLE }}`

Is there any (maybe undocumented) way of using another set of delimiters?

If not and there are no arguments against extending the rendering package, I maybe would give it a try and add a PR.

bacongobbler:
pkg text/template in Go does allow a way to [change the delimiters](https://golang.org/pkg/text/template/#Template.Delims). It's not baked into helm, but one could in theory come up with a PR to change the delimiters.

The concern would come down to chart compatibility, though. If one chart uses one set of delimiters and another uses a different delimiter, both charts are incompatible with each other once you start to add each together as dependencies or otherwise.

bacongobbler:
related discussion: #856

Issue #9302, opened 2021-01-30
Pluggable template engine support for helm 3?


vroad:
https://github.com/helm/helm/issues/8290#issuecomment-642395892

Is this still true that helm provides a way for adding templating engines?

I'm not familiar with internal structure of helm, but looks like current implementation has no way to replace only go template renderer with something else.

https://github.com/helm/helm/blob/46d80f68c11d8107b688b9c8299fa02bccba312f/pkg/engine/engine.go#L250

helm 3 already already has a way to install plugins. Can we extend its functionality so that users can develop their own renderer implementation?

That would prevent other tools from using helm as a library though.
AFAIK terraform helm provider is one of such example. The provider doesn't need helm CLI to be installed because it incorporates helm as a library.

Developing many renderers in-tree might make it difficult to update dependencies and make CLI binary bloated.

Or... could plugin renderers run on server side? Maybe too much complicated to do so.

bacongobbler:
> Is this still true that helm provides a way for adding templating engines?

This functionality was removed in Helm 3. Helm 2 was under active development for nearly 5 years, and no other template engine was ever introduced. Several community members advocated for new template engines. We pointed them towards [the `engine` mechanism](https://github.com/helm/helm/issues/2577#issuecomment-341548308), but none was ever written.

> helm 3 already already has a way to install plugins. Can we extend its functionality so that users can develop their own renderer implementation?

Plugins currently cannot inject themselves into another command like `helm install` or `helm upgrade`. They act more like git plugins in which they're standalone "helper" commands. `helm env` and `helm diff` are a few examples.

bacongobbler:
Closing this issue as answered. Please feel free to comment here and let us know if you have further questions. Thanks!

Issue #8290, opened 2020-06-11
I don't want to use go templates


chinazj:
go templates likes this.
```
{{/*
Create a default fully qualified app name.
We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
If release name contains chart name it will be used as a full name.
*/}}
{{- define "manager.fullname" -}}
{{- if .Values.fullnameOverride -}}
{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" -}}
{{- else -}}
{{- $name := default .Chart.Name .Values.nameOverride -}}
{{- if contains $name .Release.Name -}}
{{- .Release.Name | trunc 63 | trimSuffix "-" -}}
{{- else -}}
{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
{{- end -}}
{{- end -}}
{{- end -}}
```
Disadvantages
No 

- indentation
- have to modify not only `values.yaml`, but also `_helper.tpl`
- I think this is two syntax, `yaml` and `template`

bacongobbler:
So, some context here.

We have always said we are open to alternative template engine implementations. There even was support in helm’s rendering engine to support multiple template engines for Helm 2. This has been discussed across multiple threads, the largest being https://github.com/helm/helm/issues/2577. See also https://github.com/helm/helm/issues/2577#issuecomment-341548308 for more information.

Nobody from the community has been willing to put the time into writing one. So, that’s where we stand today.

PR #10299, opened 2021-10-31
Add alternative template delimiter

neerfri:

Some popular kubernetes components (prometheus,  argo-workflows, ...) use templating with `{{...}}` syntax.
This makes it hard for helm users to template the resources for these workloads (see links below).
The current workarounds render the templates hard to read and hard to maintain.

This PR addresses this by allowing a specific template to change its template delimiters.
The delimiter change is set using "magic comments" which is a wide spread trick used to carry more metadata inlined in a file (see examples below).

To change the template delimiter, a helm user would add a `# helm: delim=[,]` comment to the head of the template file:

```yaml
# helm: delimiters=[[,]]
apiVersion: v1
kind: ConfigMap
metadata:
  name: [[ template "chart-with-alt-delim.fullname" . ]]-alt
data:
  myvalue: "Hello {{world}}"
```

This will only effect the specific template and so compatibility with other charts and templates is not impacted.

Examples of magic comments in our ecosystems:
- [vim](https://vim.fandom.com/wiki/Modeline_magic)
- [go-lang](https://github.com/xaionaro-go/hackery)
- [Ruby](https://docs.ruby-lang.org/en/3.0.0/doc/syntax/comments_rdoc.html#label-Magic+Comments)
- [Webpack](https://webpack.js.org/api/module-methods/#magic-comments)

refs #4789
refs #2931
https://stackoverflow.com/questions/64802290/how-can-i-use-argo-workflows-templates-in-helm
https://stackoverflow.com/questions/56341558/how-to-escape-and-in-argo-workflow
https://tempered.works/posts/2020-05-21-helm-argo/

Added a unit test to verify rendering of a normal template as well as a template with alternative delimiter side by side.
I'd be happy to add documentation if this is merged.

**If applicable**:

**TODO**:
- [x] Open HIP
- [x] Scan lines in template only up to the first line that does not start with `#`

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -17,7 +17,6 @@ limitations under the License.
 package engine
 
 import (
-	"bufio"
 	"fmt"
 	"log"
 	"path"
@@ -84,7 +83,14 @@ func RenderWithClient(chrt *chart.Chart, values chartutil.Values, config *rest.C
 	}.Render(chrt, values)
 }
 
-var magicCommentsRegexp = regexp.MustCompile(`^#\s*helm:\s*(?:(\w+)=([^\s]*))*`)
+// A regexp to match against first lines in a template that start with #
+var templateHeaderRegexp = regexp.MustCompile(`(?sm)\A((?:^#\s[^\n]*\n)*)`)
+
+// A regexp to capture a magic comment line in the template header
+var magicCommentsRegexp = regexp.MustCompile(`(?m-s)^#\s+helm:\s*(.*)$`)
+
+// The magic comment directive to set the template delimiters
+const setDelimitersDirective = "delimiters"
 
 type templateOpts struct {
 	// Left template delimiter
@@ -97,6 +103,8 @@ type templateOpts struct {
 type renderable struct {
 	// tpl is the current template.
 	tpl string
+	// header is the header extracted from the template (first lines starting with #)
+	header []byte
 	// vals are the values to be supplied to the template.
 	vals chartutil.Values
 	// namespace prefix to the templates of the current chart
@@ -269,6 +277,12 @@ func (e Engine) renderWithReferences(tpls, referenceTpls map[string]renderable) 
 		vals := tpls[filename].vals
 		vals["Template"] = chartutil.Values{"Name": filename, "BasePath": tpls[filename].basePath}
 		var buf strings.Builder
+
+		// Add template header to output without rendering it through the template engine
+		if _, err := buf.Write([]byte(tpls[filename].header)); err != nil {
+			return map[string]string{}, cleanupExecError(filename, err)
+		}
+
 		if err := t.ExecuteTemplate(&buf, filename, vals); err != nil {
 			return map[string]string{}, cleanupExecError(filename, err)
 		}
@@ -389,10 +403,12 @@ func recAllTpls(c *chart.Chart, templates map[string]renderable, vals chartutil.
 		if !isTemplateValid(c, t.Name) {
 			continue
 		}
+		templateHeader, templateBody := extractTemplateHeaderAndBody(t.Data)
 		templates[path.Join(newParentID, t.Name)] = renderable{
-			tpl:      string(t.Data),
+			tpl:      string(templateBody),
+			header:   templateHeader,
 			vals:     next,
-			opts:     readTemplateMagicComments(string(t.Data)),
+			opts:     readTemplateMagicComments(templateHeader),
 			basePath: path.Join(newParentID, "templates"),
 		}
 	}
@@ -413,33 +429,31 @@ func isLibraryChart(c *chart.Chart) bool {
 	return strings.EqualFold(c.Metadata.Type, "library")
 }
 
-func readTemplateMagicComments(templateBody string) templateOpts {
-	templateOpts := templateOpts{}
-	templateHeader := templateHeader(templateBody)
-	matches := magicCommentsRegexp.FindAllSubmatch([]byte(templateHeader), -1)
-	for _, match := range matches {
-		if strings.EqualFold(string(match[1]), "delim") {
-			delim := strings.SplitN(string(match[2]), ",", 2)
-			templateOpts.delimL = strings.Repeat(delim[0], 2)
-			templateOpts.delimR = strings.Repeat(delim[1], 2)
-		}
-	}
-
-	return templateOpts
+// extractTemplateHeaderAndBody splits the template to the header and body components of the template
+//
+// the header is considered all the lines from the template until the first line that does not start with "#"
+// This is used to avoid scanning whole templates with a regular expression when searching for magic comments
+// and avoid rendering the header while still keeping it part of the output
+func extractTemplateHeaderAndBody(template []byte) ([]byte, []byte) {
+	headerBytes := templateHeaderRegexp.Find(template)
+	return headerBytes, template[len(headerBytes):]
 }
 
-// templateHeader returns the lines from the template until the first line that does not start with "#"
-//
-// This is used to avoid scanning whole templates with a regular expression when searching for magic comments.
-func templateHeader(templateBody string) string {
-	var headerBuffer strings.Builder
-	scanner := bufio.NewScanner(strings.NewReader(templateBody))
-	for scanner.Scan() {
-		if strings.HasPrefix(scanner.Text(), "#") {
-			headerBuffer.WriteString(scanner.Text())
-		} else {
-			break
+// readTemplateMagicComments reads magic comments from the template header and returns a `templateOpts` struct
+func readTemplateMagicComments(templateHeader []byte) templateOpts {
+	templateOpts := templateOpts{}
+	matches := magicCommentsRegexp.FindAllSubmatch(templateHeader, -1)
+	for _, match := range matches {
+		magicCommentBody := strings.SplitN(string(match[1]), "=", 2)
+		if len(magicCommentBody) == 2 && strings.EqualFold(magicCommentBody[0], setDelimitersDirective) {
+			delimiters := strings.SplitN(magicCommentBody[1], ",", 2)
+			if len(delimiters) != 2 {
+				log.Printf("Warning: invalid magic comment: %s", match[1])
+				continue
+			}
+			templateOpts.delimL = delimiters[0]
+			templateOpts.delimR = delimiters[1]
 		}
 	}
-	return headerBuffer.String()
+	return templateOpts
 }
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -391,7 +391,7 @@ func recAllTpls(c *chart.Chart, templates map[string]renderable, vals chartutil.
 		templates[path.Join(newParentID, t.Name)] = renderable{
 			tpl:      string(t.Data),
 			vals:     next,
-			opts: 	  readTemplateMagicComments(string(t.Data)),
+			opts:     readTemplateMagicComments(string(t.Data)),
 			basePath: path.Join(newParentID, "templates"),
 		}
 	}
@@ -416,7 +416,7 @@ func readTemplateMagicComments(templateBody string) templateOpts {
 	templateOpts := templateOpts{}
 	matches := magicCommentsRegexp.FindAllSubmatch([]byte(templateBody), -1)
 	for _, match := range matches {
-		if (strings.EqualFold(string(match[1]), "delim")) {
+		if strings.EqualFold(string(match[1]), "delim") {
 			delim := strings.SplitN(string(match[2]), ",", 2)
 			templateOpts.delimL = strings.Repeat(delim[0], 2)
 			templateOpts.delimR = strings.Repeat(delim[1], 2)
@@ -424,4 +424,4 @@ func readTemplateMagicComments(templateBody string) templateOpts {
 	}
 
 	return templateOpts
- }
+}
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -17,6 +17,7 @@ limitations under the License.
 package engine
 
 import (
+	"bufio"
 	"fmt"
 	"log"
 	"path"
@@ -414,7 +415,8 @@ func isLibraryChart(c *chart.Chart) bool {
 
 func readTemplateMagicComments(templateBody string) templateOpts {
 	templateOpts := templateOpts{}
-	matches := magicCommentsRegexp.FindAllSubmatch([]byte(templateBody), -1)
+	templateHeader := templateHeader(templateBody)
+	matches := magicCommentsRegexp.FindAllSubmatch([]byte(templateHeader), -1)
 	for _, match := range matches {
 		if strings.EqualFold(string(match[1]), "delim") {
 			delim := strings.SplitN(string(match[2]), ",", 2)
@@ -424,4 +426,20 @@ func readTemplateMagicComments(templateBody string) templateOpts {
 	}
 
 	return templateOpts
+}
+
+// templateHeader returns the lines from the template until the first line that does not start with "#"
+//
+// This is used to avoid scanning whole templates with a regular expression when searching for magic comments.
+func templateHeader(templateBody string) string {
+	var headerBuffer strings.Builder
+	scanner := bufio.NewScanner(strings.NewReader(templateBody))
+	for scanner.Scan() {
+		if strings.HasPrefix(scanner.Text(), "#") {
+			headerBuffer.WriteString(scanner.Text())
+		} else {
+			break
+		}
+	}
+	return headerBuffer.String()
 }
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -83,6 +83,15 @@ func RenderWithClient(chrt *chart.Chart, values chartutil.Values, config *rest.C
 	}.Render(chrt, values)
 }
 
+var magicCommentsRegexp = regexp.MustCompile(`^#\s*helm:\s*(?:(\w+)=([^\s]*))*`)
+
+type templateOpts struct {
+	// Left template delimiter
+	delimL string
+	// Right template delimiter
+	delimR string
+}
+
 // renderable is an object that can be rendered.
 type renderable struct {
 	// tpl is the current template.
@@ -91,6 +100,8 @@ type renderable struct {
 	vals chartutil.Values
 	// namespace prefix to the templates of the current chart
 	basePath string
+	// template engine options
+	opts templateOpts
 }
 
 const warnStartDelim = "HELM_ERR_START"
@@ -230,7 +241,7 @@ func (e Engine) renderWithReferences(tpls, referenceTpls map[string]renderable) 
 
 	for _, filename := range keys {
 		r := tpls[filename]
-		if _, err := t.New(filename).Parse(r.tpl); err != nil {
+		if _, err := t.New(filename).Delims(r.opts.delimL, r.opts.delimR).Parse(r.tpl); err != nil {
 			return map[string]string{}, cleanupParseError(filename, err)
 		}
 	}
@@ -380,6 +391,7 @@ func recAllTpls(c *chart.Chart, templates map[string]renderable, vals chartutil.
 		templates[path.Join(newParentID, t.Name)] = renderable{
 			tpl:      string(t.Data),
 			vals:     next,
+			opts: 	  readTemplateMagicComments(string(t.Data)),
 			basePath: path.Join(newParentID, "templates"),
 		}
 	}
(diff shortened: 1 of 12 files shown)


neerfri:
@jdolitsky @mattfarina your review will be appreciated, thanks!

neerfri:
👋  Friendly bump here...
Would love to know the team's opinion on this change and see that I'm passing all CI checks
Thanks!

/cc @hickeyma @scottrigby @yxxhero

mattfarina:
@neerfri thanks for the pull request and clear description of the need. I've added this to the agenda to discuss in our next meeting.

pkg/engine/engine.go
```diff
@@ -399,3 +411,17 @@ func isTemplateValid(ch *chart.Chart, templateName string) bool {
...
 }
+
+func readTemplateMagicComments(templateBody string) templateOpts {
+	templateOpts := templateOpts{}
+	matches := magicCommentsRegexp.FindAllSubmatch([]byte(templateBody), -1)
+	for _, match := range matches {
+		if strings.EqualFold(string(match[1]), "delim") {
+			delim := strings.SplitN(string(match[2]), ",", 2)
+			templateOpts.delimL = strings.Repeat(delim[0], 2)
+			templateOpts.delimR = strings.Repeat(delim[1], 2)
```
bacongobbler, line 422: Should there be a length check in place? What happens if I write the following?

```
# helm: delim=[
```

Wouldn't that result in a `panic()`?

pkg/engine/engine.go
```diff
@@ -399,3 +411,17 @@ func isTemplateValid(ch *chart.Chart, templateName string) bool {
...
 	return strings.EqualFold(c.Metadata.Type, "library")
 }
+
+func readTemplateMagicComments(templateBody string) templateOpts {
+	templateOpts := templateOpts{}
+	matches := magicCommentsRegexp.FindAllSubmatch([]byte(templateBody), -1)
+	for _, match := range matches {
+		if strings.EqualFold(string(match[1]), "delim") {
+			delim := strings.SplitN(string(match[2]), ",", 2)
+			templateOpts.delimL = strings.Repeat(delim[0], 2)
```
bacongobbler, line 421: Why can't I specify my own delimiter syntax? For example,

```
# helm: delim={%,%}
```

Why does it have to be a single character repeated twice? Why not allow single character delimiters?

```
# helm: delim=$(,)
```

What happens if I wanted commas as a delimiter?

```
# helm: delim=,,,
```

bacongobbler: What happens if I wrap the delimiters in single or double quotes?

```
# helm: delim="{,}"
```

or

```
# helm: "delim={,}"
```

bacongobbler (changes requested):
I had some concerns around the stability of the parser. I left a few comments surrounding those concerns.

I'm curious why you chose to do this per-template instead of a field in the Chart.yaml. It'd be less computationally expensive because you wouldn't have to run a regular expression against each template, and the syntax would be much simpler to understand (it's just another field in a YAML file). With regards to backwards compability concerns... You could attempt to load each dependency's chart metadata so their templates are all rendered based on the specific chart's delimiter. That way users can write charts with their own "custom" delimiter syntax and include charts that use another delimiter.

I do have some concerns that this may make it VERY difficult to port to another template engine in the future. Assuming we do introduce multiple template engine support in the future, this feature implies every template engine should be able to switch the delimiters in use. text/template is the rare exception - many template engines do not provide this ability. See https://github.com/helm/helm/issues/2577 for some context.

In some ways this makes #2577 more compelling - users could override the default template engine through inheritance, change the default delimiter, then select their custom template engine.

In any case this may be worth going through the HIP process. There are a number of critical design decisions in this PR to consider which may warrant a broader discussion with the community. See https://github.com/helm/helm/blob/main/CONTRIBUTING.md#proposing-an-idea for more details.

Happy to hear your thoughts on this one.

neerfri:
@bacongobbler Thanks for your reply. You're bringing up important points.

> I'm curious why you chose to do this per-template instead of a field in the Chart.yaml. It'd be less computationally expensive because you wouldn't have to run a regular expression against each template, and the syntax would be much simpler to understand (it's just another field in a YAML file). 

Using the `Chart.yaml` is a good alternative, here's a small comparison

Specifying the delimiter in the `Chart.yaml` file will mean:

* more simple to document/understand
* no need to "read-ahead" the template with a regex
* will solve the "delimiter in the template" problem (where it's specified as `delim=[,]` and used as `[[ ... ]]`)

Specifying the delimiter in a magic comment will mean:

* the delimiter is explicitly defined in a template (so a "copy-paste" preserves the template semantics)
* ability to only use the alternative delimiter to certain CRDs where it is useful and keep other templates "community-compatible"
* charts can be refactored to new delimiter gradually
In regards to the performance, the regex scan should only be done on the first few lines of each template, specifically I should stop the scan on the first line that does not start with `#`. This should be changed in the current implementation so I'm adding it as a TODO to this pr.

> I do have some concerns that this may make it VERY difficult to port to another template engine in the future. Assuming we do introduce multiple template engine support in the future, this feature implies every template engine should be able to switch the delimiters in use. text/template is the rare exception - many template engines do not provide this ability. See #2577 for some context.

Consider this feature a feature of helm's `text/template` engine rather than helm. Some template engines don't even have the notion of a delimiter. Honestly I still hope the community (or myself) will implement a "templateless" template engine using yaml's rich type tags (maybe a bit similar to what AWS did in CludFormations's YAML representation), but that's a discussion for another HIP :-D

> In any case this may be worth going through the HIP process. There are a number of critical design decisions in this PR to consider which may warrant a broader discussion with the community. See https://github.com/helm/helm/blob/main/CONTRIBUTING.md#proposing-an-idea for more details.

Sounds good. Now that I see the core team is interested in the idea in general I'd be happy to invest in turning this to a HIP.

Thanks again for the thorough review ❤️ 

✌️

pkg/engine/engine.go
neerfri: Yes, good catch!
This makes me wonder if it may be better to use separate modifiers for each delimiter like the text/template` does:

```
# helm: delimL=[
# helm: delimR=]
```

I'll give it another thought and play with the code to see the pros/cons

pkg/engine/engine.go
neerfri: > Why can't I specify my own delimiter syntax? For example,
> 
> ```
> # helm: delim={%,%}
> ```
> 
> Why does it have to be a single character repeated twice? Why not allow single character delimiters?
> 
> ```
> # helm: delim=$(,)
> ```

I mentioned this in the PR description.
The problem is that when the template is rendered the line `# helm: delim={%,%}` is parsed by the engine and it will try to evaluate `{%,%}` and panic since `,` is not a valid `go-template` expression.

I could remove the magic comment from the template, but I think it adds value to keep it there since you can see the magic comments that were part of the rendering process in the final output which contributes to debugging, not to mention performance (altering the string templates, that are kept in memory)

I could also try to escape it somehow, but just the thought of it gives me a headache considering we are talking about a dynamic delimiter :-D 
```
# helm: delim={%,%}
```
should become
```
# helm: delim={%`{%`%},{%`%}`%}
```

> What happens if I wanted commas as a delimiter?
> 
> ```
> # helm: delim=,,,
> ```

😱

Maybe we can even help downstream users and limit the possible delimiters to a set of characters that we know will not collide with other things in the templates.
Imagine if users would choose "`" as the delimiter or even `#`...

What do you think about the set: `[ { ( %`?

This paves the way to another good option which is to have a closed set of "delimiting style":

```
# helm: delimiterStyle=curlyBrackets
# helm: delimiterStyle=squareBrackets
# helm: delimiterStyle=roundBrackets
# helm: delimiterStyle=percentageSigns
```

This will give users less freedom but in return will have a much more simple, stable and stupid-proof UX, as an added bonus this solves the problem of having the explicit delimiter in the template (the one explained in the top of this comment) because it's not using the delimiter itself to describe the delimiter.

Thoughts?

bacongobbler: I still think this would be better solved by implementing something similar to what was asked in https://github.com/helm/helm/issues/9302. We've heard more users ask about whether they can use alternative template engines like jinja2 rather than extending the existing `text/template` engine. See also https://github.com/helm/helm/issues/8290.

If we're going to allow users to extend the existing template engine, IMO it would be m,ore preferable to implement engine swap support rather than just switching config flags through a "magic comments" system. Then you can write your own template engine that extends Helm's current `pkg/engine`, swapping out the delimiter with your own choosing. If we find that system to be too clumsy then we can revisit this idea in the future. But we should be able to accomodate both your proposal here as well as what others have asked for in one fell swoop.

mattfarina:
@neerfri thanks for bringing this up. We discussed it in the Helm developer call.

One thing that came up in discussion is backwards compatibility. If an older version of Helm picks up a chart with this change it will silently ignore the annotation and render the chart incorrectly. It it could be silently broken. This is different from the situation where a new template function is introduced. An older version of Helm would throw an error in that situation.

How would you handle this?

I think we need two HIPs. One to introduce annotations that specify behavior. We should think that through so we can make sure it works for other cases that will come up. The second one is for the delimeter functionality.

I've labeled this as v4 because I think it would require us to increment the chart version to v3 for compatibility. We might want to create a separate label for that.

neerfri:
/reply @mattfarina 

> One thing that came up in discussion is backwards compatibility. If an older version of Helm picks up a chart with this change it will silently ignore the annotation and render the chart incorrectly. It it could be silently broken. This is different from the situation where a new template function is introduced. An older version of Helm would throw an error in that situation.
> 
> How would you handle this?
> 

I guess the answer is embedded in the question :-D
Assuming this feature makes it to `3.8`:

in `NOTES.txt` add:

```
{{- $minimumHelmVersion := "3.8" }}
{{- if semver $minimumHelmVersion | (semver .Capabilities.HelmVersion.Version).Compare | gt 0 }}
{{- printf "Helm version must be greater than %s" $minimumHelmVersion | fail }}
{{- end }}
```

But in this particular case users want to use this feature so they can use `{{ ... }}` in their templates, so the chart will most probably fail anyway:

**current Helm without the feature:**
```
➜  helm git:(add-alternative-template-delimiter) ✗ helm template cmd/helm/testdata/testcharts/chart-with-alt-delim
Error: parse error at (chart-with-alt-delim/templates/alt-configmap.yaml:8): function "world" not defined

```

<details><summary>**Helm with the feature:** (click to expand)</summary>

```
➜  helm git:(add-alternative-template-delimiter) ✗ ./bin/helm template cmd/helm/testdata/testcharts/chart-with-alt-delim
---
# Source: chart-with-alt-delim/templates/alt-configmap.yaml
# helm: delim=[,]
apiVersion: v1
kind: ConfigMap
metadata:
  name: RELEASE-NAME-chart-with-alt-delim-alt
  name: RELEASE-NAME-chart-with-alt-delim-alt
data:
  myvalue: "Hello {{world}}"
---
# Source: chart-with-alt-delim/templates/normal-configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: RELEASE-NAME-chart-with-alt-delim-normal
data:
  myvalue: "Hello world"
```
</details>

> I think we need two HIPs. One to introduce annotations that specify behavior. We should think that through so we can make sure it works for other cases that will come up. The second one is for the delimeter functionality.
> 
I created the HIP for the alternative delimiters https://github.com/helm/community/pull/222

I don't know how to create the other HIP because I don't have much to say about it that is not directly about the alternative delimiter.

> I've labeled this as v4 because I think it would require us to increment the chart version to v3 for compatibility. We might want to create a separate label for that.

Not sure we need to increment the chart version for this, while this might look "big" it's not because it's not breaking any chart that does not use this feature.

Labeling this `v4` is really unfavorable for this change since it will anyway take a long time for community charts to adopt this (because it requires deprecating the supported helm version by those charts).

Given the fact that it's not a breaking change, I would be happy if the helm team can consider putting this into `3.8` after we resolve the discussions in this PR. In case we the team thinks this should be explicitly experimental we can put this under a `HELM_EXPERIMENTAL_DELIM` flag (like `HELM_EXPERIMENTAL_OCI`)

pkg/engine/engine.go
neerfri: I agree. I'd be extremely happy if Helm would have a pluggable template engine feature.
I don't see why this is a reason not to accept a tiny feature in the current built-in `text/template` engine.

pkg/engine/engine.go
neerfri: fixed in new code

pkg/engine/engine.go
neerfri: Fixed.
The new code will now parse `# helm: delimiter={%,%}` as valid input and will not "double" the delimiters.
It should also not panic for bad input

mattfarina:
To add some context....

Helm has [a list of user profiles (roles) in a priority order](https://github.com/helm/community/blob/main/user-profiles.md). The top role is the "Application Operator". This is someone who takes a chart and installs a workload with it. They may not know much about Kubernetes or the application they are trying to run. For example, take PostgreSQL. They may not know much about operating it or StatefulSets. This is conceptually similar to a Linux package manager where someone can install an application without knowing how do work with systemd, where config files need to live, or where to put the binaries. They can install an app and it just works.

With that in mind...

> in NOTES.txt add:
> 
>  ```
> {{- $minimumHelmVersion := "3.8" }}
> {{- if semver $minimumHelmVersion | (semver .Capabilities.HelmVersion.Version).Compare | gt 0 }}
> {{- printf "Helm version must be greater than %s" $minimumHelmVersion | fail }}
> {{- end }}
> ```

This requires the chart maintainer to put something in their chart. There isn't a way to Helm itself to protect the "Application Operator" in older versions that are already released.

As we've seen, chart maintainers often fail to include things like this. They often don't realize their impact because it's outside of their personal use or the environments they test it.

In Helm we want to protect those end users. Many of whom don't read the contents of a chart or completely understand them. There is a level of trust that Helm and the chart authors will protect them. In Helm, we try to put some level of basic protections in.

If a chart author fails to put this sort of statement in and an end user installs something with an older version of Helm it could break their workload. This is the type of situation we try to avoid if possible. We are conservative in how we handle things prioritizing the Application Operator over the Application Distributor.

How could we have this level of protection for those top priority users?

>  In case we the team thinks this should be explicitly experimental we can put this under a HELM_EXPERIMENTAL_DELIM flag (like HELM_EXPERIMENTAL_OCI)

In my opinion, the experiment system has failed. This is because some people outside the Helm maintainers told their users that the experiments were production ready and people used it. Without checking on its status. This says a lot about human nature and how it's not logical or researched. Using the experiment the way they did ended up causing problems with turned into support issues for Helm. People didn't treat the experiment as an experiment. So, we are hesitant to perform other experiments.

In the case of this feature, an experiment would not protect users of older Helm versions. If they tried to install the chart and didn't have the experiment it would cause a broken situation. The very thing we are trying to avoid.

While not ideal, there are other ways to solve this problem. I've helped others on their charts. Have you tried any of the alternative methods to embed one go template in another and keep it as a template?

neerfri:
@mattfarina Thanks for the explanation, I agree with all of it and I think it's awesome that you guys are thinking about this.

As I mentioned any reasonable use-case for this feature will want to use `{{ ... }}` expressions without the need to escape them. If the **Application Operator** will install that chart with a helm version that does not support the feature it will most likely fail because the value in the `{{ ... }}` expression will fail to resolve to a proper helm value.

Is this reasonable protection for the **Application Operator**?

neerfri:
@mattfarina here's an example of running the test chart this PR adds with a version of helm that does not have the feature:

```
➜  helm git:(add-alternative-template-delimiter) ✗ helm template cmd/helm/testdata/testcharts/chart-with-alt-delim
Error: parse error at (chart-with-alt-delim/templates/alt-configmap.yaml:8): function "world" not defined

```

bacongobbler:
Here's an example that does not result in your error case.

```
# helm: delimiters=[[,]]
apiVersion: v1
kind: ConfigMap
metadata:
  name: foo
data:
  myvalue: "Hello [[ printf "world" ]]"
```

If we use Helm 3.0.0, it will render the following:

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: foo
data:
  myvalue: "Hello [[ printf "world" ]]"
```

If we use this PR, it will render the following:

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: foo
data:
  myvalue: "Hello world"
```

Many users would argue this could break certain workloads. Both are valid manifests according to Kubernetes, but one of them was rendered incorrectly because they were using an older version of Helm.

This reinforces the point that there needs to be some level of protection put in place in Helm to prevent this chart from rendering. We cannot rely on the chart author to do the right thing and put these safeguards in place. In the past we have done that by increasing the Chart.yaml's `apiVersion` field from `v1` to `v2` to note that there have been changes made to how helm renders a chart.

neerfri:
@bacongobbler 👍

neerfri:
@bacongobbler Given this goes to helm v4 are there still any code changes needed? (your review marked "change requested")

@bacongobbler @mattfarina How do you suggest to proceed here?

neerfri:
> Here's an example that does not result in your error case.

@bacongobbler actually the example you posted does fail, which shows how unlikely it is to use the feature in a way that will not fail on an older helm version.

Here's an example of a template that will be rendered wrong:

```
# helm: delimiters=[[,]]
apiVersion: v1
kind: ConfigMap
metadata:
  name: foo
data:
  myvalue: "Hello [[ printf `world` ]]"
```


Here's how the previously posted example fails (due to invalid YAML):

```
➜  helm git:(add-alternative-template-delimiter) ✗ helm version
version.BuildInfo{Version:"v3.6.3", GitCommit:"d506314abfb5d21419df8c7e7e68012379db2354", GitTreeState:"clean", GoVersion:"go1.16.5"}

➜  helm git:(add-alternative-template-delimiter) ✗ cat cmd/helm/testdata/testcharts/chart-with-alt-delim/templates/alt-configmap.yaml
# helm: delimiters=[[,]]
apiVersion: v1
kind: ConfigMap
metadata:
  name: foo
data:
  myvalue: "Hello [[ printf "world" ]]"

➜  helm git:(add-alternative-template-delimiter) helm template cmd/helm/testdata/testcharts/chart-with-alt-delim
Error: YAML parse error on chart-with-alt-delim/templates/alt-configmap.yaml: error converting YAML to JSON: yaml: line 6: did not find expected key

```

ebuildy:
This is a fantastic feature we'd love to have! espacially to deal with argo-workflow template & prometheus alerting.

Any chance to have it merged?

rhgreene:
Is this change still being considered? This would fix all my woes when deploying applications with inbuilt go-templating in config - like the aforementioned prometheus and alertmanager.

TheMeier:
Any updates here?

neerfri:
@ebuildy @rhgreene @TheMeier 

As you can see from the discussions above the Helm maintainers still have reservations regarding introducing that feature into Helm.

I personally believe the feature is safe enough and improves the UX and safety of charts that need to render resources with go templates in them but I assume I'm biased and respect the maintainers' say regarding this.

if a Helm maintainer is interested I am willing to keep pushing changes to make this happen

nicktimko:
I'm also in this boat, where using Argo + Helm is really painful. I'd like *some* fix, and here's maybe more of the possible solution space, if anything is maybe easier or more agreeable.

* Custom delimiters: changing the literal text in the source file to something else. This allows
  * user-specified vs. fixed (developer) selection. Simplest would be some toggle that switches to a fixed alternative, ideally one that is sufficiently 'rare'.
  * specified via 
    * file/document comment directive (unsure if comments are attached to documents or files in YAML)
    * command line switch (global or file-by-file)
    * options file
* Directive to disable template parsing (limited use if you need to actually template the templates, which is quite plausible)
  * Start/end delimited
    * Example: [Jinja2](https://jinja.palletsprojects.com/en/3.0.x/templates/#escaping) uses `{% raw %}unparsed text{% endraw %}` 
    * Example: [Black](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html) uses comments like `# fmt: off`, `lines...`, `# fmt: on`.
  * Disable by line
    * Example: [Black](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html) uses comments like `unescaped  # fmt: skip`
* Simpler escaping (this is kinda like custom delimiters, but maybe could be more restrictive) like `{{{` or `{{{{` or whatever. Excel uses double double quotes `""` to show one quote...

Given the presence of the delimiter changing `template.Delims` option, that's probably the easiest path...

pbenefice:
Just referencing here what personnally helped me on this topic, the "raw string" :
```shell
# Excerpt from a Prometheus Alertmanager yaml
receivers:
- name: slack-receiver
  slack_configs:
  - text: |-
      {{`{{ range .Alerts }}
        *Alert:* {{ .Annotations.summary }}
      {{ end }}`}}
```
Notice the backtick on starting and ending curly braces

Found thanks to : https://github.com/helm/helm/issues/2798#issuecomment-467319526

kahirokunn:
I need this feature :cry:

alexander-rabenstein:
Same here, Templates are getting really ugly if you run helm + argo + externalsecrets. 
And I think it will get worse: Go is the go to language for Kubernetes Implementations, go-template is the go to template library. 
Image now you want to create and maintain a helm chart which deploys an external-secret which contains some weird application template logic which again is using {{ as a delimiter.

I get all the implications, but still this is a big pain point

PR #10302, opened 2021-11-01
Fail if target-directory does not exist

### LINKED ISSUE #10298: Installation via `scripts/get` fails if directory /usr/local/bin does not exist
When installing current version via `scripts/get`, installation fails if the target directory for the binaries does not exist.

At https://github.com/helm/helm/blob/v3.7.1/scripts/get#L141 a simple copy is done without checking if the directory exists. This results in the binary being installed as "/usr/local/bin", i.e. an executable file instead of a directory.

Creating the directory before starting the script makes it work fine.

Output of `helm version`: not installed yet, tried latest master, v3.7.1

FYI, `get-helm-3` uses a slightly different copy-operation which will fail if the bin-directory does not exist.

centic9:
By specify the binary-name in the copy-operation the
script will fail if the directory does not exist instead
of silently installing the binary with the name of
the directory (e.g. /usr/local/bin)

Closes #10298

Signed-off-by: Dominik Stadler <dominik.stadler@gmx.at>

### DIFF
--- a/scripts/get
+++ b/scripts/get
@@ -138,10 +138,10 @@ installFile() {
   HELM_TMP_BIN="$HELM_TMP/$OS-$ARCH/$PROJECT_NAME"
   TILLER_TMP_BIN="$HELM_TMP/$OS-$ARCH/$TILLER_NAME"
   echo "Preparing to install $PROJECT_NAME and $TILLER_NAME into ${HELM_INSTALL_DIR}"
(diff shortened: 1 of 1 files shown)


bacongobbler:
Are there applicable changes required for the get-helm-3 script as well?

centic9:
No, get-helm-3 already does it this way.

bacongobbler (approved):

Issue #3481, opened 2018-02-09
Watch hooks logs and log release-checking status in helm install/upgrade


distorhead:
There are 2 issues related to logs streaming, that we are experiencing using helm.

## Watch hooks logs

Generally we have a project, that uses helm for deploy (https://github.com/flant/dapp). And in our scheme deploy run as follows:

* run helm install/upgrade in main thread (blocking);
* parse hooks-jobs templates and show logs for jobs with annotation `dapp/watch-logs = true` in separate thread.

It works ok, but we have a strong belief, that it is a job of the helm to show logs.
We have done some work on that in ruby (https://github.com/flant/dapp/blob/master/lib/dapp/kube/dapp/command/deploy.rb#L119) and now we can use that experience to improve helm.

So, there is an idea to use hook-job annotation like `watch-logs = true|false`.
Say, by default none of the hooks is logged, so it is supposed that `watch-logs = false` for all hooks. In that case tiller will stream only logs related to the hooks where `watch-logs = true`.
And as the next step add option to helm client & tiller to log all hooks by default. With that option tiller will suppose `watch-logs = true` for all hooks by default.

And that logs streaming could occur directly in install / upgrade release grpc-action using server-side streaming.

## Check release state and show status during checking

There is related to the logging issue about checking release state (`--wait` in helm). 

Release checking may also log some status info about "what is being checked" and "what we are waiting for". It is not necessary pod's logs, it could be container status, information about replicas count, that needed to be reached and current replicas count for deployment.

Again, in our project we have done some work on Deployment status watching (https://github.com/flant/dapp/blob/master/lib/dapp/kube/kubernetes/manager/deployment.rb#L33). After helm install / upgrade is done the thread that watches hooks logs is also done, we run a deployments-watching procedure.

It is similar to what `--wait` is doing, but it gives interactive info about what is being checked and shows, among other things, pod's logs. That logging is not "ideal", but there is an idea, that release readiness check could work that way: making release checks polling and at the same time give some info to the user.

The main purpose of that logging is to show enough info for the end user during install / upgrade operation, so that user can diagnose the problem and make changes to templates without running kubectl to dig what is wrong.

And again we have a strong belief, that helm in `--wait` mode could do something like that to become a trully end-user kubernetes deployment tool.

---

Also there is a related issue https://github.com/kubernetes/helm/issues/2298, but separate helm command to show logs will not make helm more user-friendly deployment tool, that works out-of-the box.

distorhead:
[...]

Issue #2298, opened 2017-04-19
Feature: Stream hook logs back to helm client


johnw188:
When performing helm release operations such as `install`, `upgrade`, `delete` and `rollback`, the helm client will block without feedback while hooks execute. If chart owners aren't careful with their error handling, kubernetes can put the hook job into an infinite loop of failure. Conversely, long running hooks such as database migrations can be executing successfully while the user is unaware that code is executing.

To mitigate these issues helm should stream hook logs back to the helm client.

## UI

Successful install:
```
› helm install internal/jenkins
Executing pre-install hooks: ldap-update-job, credential-update-job
[credential-update-job] Updating Jenkins credentials
[ldap-update-job] Updating LDAP configuration
[credential-update-job] Validating credentials
[credential-update-job] Successfully updated credentials!
[ldap-update-job] Successfully updated LDAP!

Executing post-install hooks: migrate-db
[migrate-db] Starting database migration
[migrate-db] Database migration completed

Install was a success! Happy Helming!

```

Install with issues:
```
› helm install internal/jenkins
[credential-update-job] Updating Jenkins credentials
[ldap-update-job] Updating LDAP configuration
[credential-update-job] Validating credentials
[credential-update-job] Successfully updated credentials!
[ldap-update-job] ERROR: Invalid password/token for user: john.welsh
[ldap-update-job] Updating LDAP failed
[credential-update-job] Updating Jenkins credentials
[ldap-update-job] ERROR: Invalid password/token for user: john.welsh
[ldap-update-job] Updating LDAP failed
...
```

Open questions

- Timestamps?
- Should pod name be surfaced somewhere?
  - Note that the pod name can change if the hook fails or gets evicted before it completes
  - Pod names are super long (`john-ci-jenkins-update-credentials-xzz5f-1z166`)

Future enhancements

- Retry counter for jobs?

## Implementation

In the tiller proto definition:

```
// GetReleaseLogs gets a stream of log events for a given release.
    rpc GetReleaseLogs(GetReleaseLogsRequest) returns (stream GetReleaseLogsResponse) {
}

message GetReleaseLogsRequest {
    // Name is the name of the release
    string name = 1;
    // Version is the version of the release
    int32 version = 2;
}

message GetReleaseLogsResponse {
    // Source is the name of the hook that generated the log
    string source = 1;
    // Log is a single log line
    string log = 2;
}
```

The helm client will then make two gRPC calls - one to install, the other to grab logs.

Questions

[...]

PR #10305, opened 2021-11-02
WIP: Output logs of failed hooks pods and jobs

mishamo:
Addresses https://github.com/helm/helm/issues/3481 and https://github.com/helm/helm/issues/2298 in outputting the logs of hook jobs and pods on failure.

Happy to add test coverage and documentation, just wanted early feedback of whether this approach would be acceptable.

Considerations:
 - Using the existing `c.Log(...)` to leverage the `--debug` flag can't really work here. We could read the container logs through a buffer, but `c.Log` adds a newline, which would split the log lines.
 - Considered adding a feature flag to enable this behaviour, but it feels quite invasive to pass flags from all 4 commands (Install, Upgrade, Rollback, Uninstall) down to this layer. Happy to do that if there's pushback.
 - There doesn't seem to be a clear place to add tests for this, but we'll find somewhere if the general approach looks ok.

**If applicable**:

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"log"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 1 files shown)


mishamo:
Closed in favour of https://github.com/helm/helm/pull/10309

PR #10325, opened 2021-11-09
allow ldflags to overwrite k8s version

### LINKED ISSUE #10324: Helm lint does not report api deprecations in k8s 1.21+
Output of `helm version`: 3.6 / 3.7

helm 3.7 compiled against client-go 1.22
helm 3.6 compiled against client-go 1.21

Linting a chart with a policy/v1 PodDisruptionBudget or a PodSecurityPolicy (both deprecated in 1.21) should trigger a deprecation warning. But it doesn't...

After some digging through the code I believe the issue is the following:
The k8s version used to run `k8s.io/apiserver/pkg/endpoints/deprecation` is supplied as [const](https://github.com/helm/helm/blob/release-3.6/pkg/lint/rules/deprecations.go#L28). Overriding const values with `ldflags` is not possible and silently fails / does nothing (golang [issue](https://github.com/golang/go/issues/47072)). Hence the deprecation check is always performed against k8s 1.20, and not the expected client-go version defined with [`ldflags`](https://github.com/helm/helm/blob/main/Makefile#L65).

sboschman:
Signed-off-by: Sverre Boschman <1142569+sboschman@users.noreply.github.com>

Allow ldflags to overwrite the k8s version as intented. Hopefully fixes #10324 

**If applicable**:

### DIFF
--- a/pkg/chartutil/capabilities.go
+++ b/pkg/chartutil/capabilities.go
@@ -28,12 +28,10 @@ import (
 	helmversion "helm.sh/helm/v3/internal/version"
 )
 
-const (
+var (
 	k8sVersionMajor = 1
 	k8sVersionMinor = 20
-)
 
-var (
 	// DefaultVersionSet is the default version set, which includes only Core V1 ("v1").
 	DefaultVersionSet = allKnownVersions()
 
--- a/pkg/lint/rules/deprecations.go
+++ b/pkg/lint/rules/deprecations.go
@@ -25,7 +25,7 @@ import (
 	kscheme "k8s.io/client-go/kubernetes/scheme"
 )
 
-const (
+var (
 	// This should be set in the Makefile based on the version of client-go being imported.
 	// These constants will be overwritten with LDFLAGS
 	k8sVersionMajor = 1


pkg/chartutil/capabilities.go
```diff
@@ -28,12 +28,10 @@ import (
 	helmversion "helm.sh/helm/v3/internal/version"
 )
 
-const (
+var (
 	k8sVersionMajor = 1
 	k8sVersionMinor = 20
-)
 
-var (
```
bacongobbler, line 36: Since this starts with a capital letter, this makes it have global visibility. Someone using the SDK could modify this variable, whereas before this wasn’t possible due to it being declared a `const`. I don’t think mutability is a desirable design choice.

bacongobbler: Meh… prior to this change it was declared a `var` so this doesn’t change its existing functionality. I don’t agree with this design choice, but I guess it is what it is /shrug

bacongobbler (approved):

mattfarina (approved):

bacongobbler:
A community member mentioned they see the following when trying to compile Helm:

image

I think we need to revert this; clearly it isn't working as intended, and we've gone from silently ignoring the flag to a compiler error. CircleCI confirms this with the latest builds too: https://app.circleci.com/jobs/github/helm/helm/14796?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

PR #10206, opened 2021-09-30
test(linter): switch from golint to revive and fix lint errors

mmulholla:
Signed-off-by: Martin Mulholland <mmulholl@redhat.com>

**What this PR does / why we need it**: 
It does 2 things:
- changes from the golint linter, which is deprecated, to the replacment revive
- fixes all of the linting errors including existing ones from gofmt, and govet, and new ones from revive

without these changes ```make test``` fails and does not run any tests. 

**Special notes for your reviewer**:

I fixed all of the gofmt errors by running ```gofmt -s -w <file>```, for example:
```gofmt -s -w cmd/helm/root_unix.go```

**If applicable**:
- [ ] this PR contains documentation
- [ ] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

closes #10205

PR #10349, opened 2021-11-16
WIP: Migrate from golangci-lint to revive

jdolitsky:
Upon running `make test-style`, `golangci-lint` recommends deprecating itself for [revive](https://github.com/mgechev/revive):
```
Run make test-style
level=warning msg="[runner] The linter 'golint' is deprecated (since v1.41.0) due to: The repository of the linter has been archived by the owner.  Replaced by revive."
```


bacongobbler:
> GO111MODULE=on revive ./...
> bash: line 1: revive: command not found

Once that's resolved looks like this is good to go

kke:
golangci-lint itself is not deprecated nor archived, but golint, which is used as one of the sublinters in golangci-lint is. They have replaced their golint dependency to revive, which is a drop-in replacement.

I would think the only necessary action in helm is to bump the used golangci-lint version.

The deprecation notices are discussed here: https://github.com/golangci/golangci-lint/issues/1987

kke:
Also, there's already #10206

jdolitsky:
@kke - thank you for clarifying. Closing this in favor of #10206

Issue #10363, opened 2021-11-18
Broken 3 way merge cases


distorhead:
Hi!

There is 2 cases of failed release upgrades with helm, which require manual intervention to fix release integrity, instead of automatically resolving the situation by the 3 way merge.

## Release with a new resource in the chart

1. Chart was successfully deployed.
2. Add new resource into the chart, say sts/mysts. And introduce some error in this custom branch for some already existing resource (not mysts, but some already existing resource).
3. Deploy of the updated chart will fail with an expected error. New resource mysts will be created at this step.
4. Try to upgrade release using previous version of the chart, where mysts resource is not exists and there is no errors in the chart templates. After upgrade introduced error will be repaired as expected, but newly created resource mysts will not be deleted.

## Release with a new field of existing resource in the chart

1. Chart was successfully deployed.
2. Add new field into the existing resource, say field `spec.containers.command` into the deployment/mydeploy, which is intentionally broken command (`command: [ "no-such-cmd" ]` for example).
3. Deploy of the updated chart will fail with an expected error (use `--wait` flag).
4. Next try to upgrade release using previous version of the chart, where there is no broken command field. After upgrade resource in the cluster will still have the broken command field.

Both of these errors occur because helm creates 3 way merge patches not between 1) previous version of the release, 2) current chart templates and 3) current kubernetes resources, but between 1) previous version of the release in the **deployed**  state, 2) current chart templates and 3) current kubernetes resources.

As a result in the first case `sts/mysts` does not treated as a managed resource by the helm, and this resource is not deleted.

In the second case new field `spec.containers.command` does not treated as a managed field by the helm as well, and this field is not deleted.

Solution: create 3wm patches between previous version of the release in whatever state.

bacongobbler:
I believe this was intentional behaviour when we first wrote this up. Following this code path, your chart was successfully deployed to Kubernetes (it was a valid manifest), but you applied a change that caused the app to fail when starting up.

In this case it is expected for the user to issue a `helm rollback` to return back to a known good state before upgrading.

can you please provide a test case to reproduce this issue?

bacongobbler:
Hmm... Perhaps the `--wait` flag is what's tripping the release status to enter a FAILED state. That may cause differing behaviour than what's expected.

bacongobbler:
Marking as a question/support issue until we can determine whether this is a bug or not.

distorhead:
[...]

PR #10364, opened 2021-11-18
fix: fixed broken 3 way merge when previous release was failed

distorhead:
Create a patch from previous release revision no matter succeeded it was or failed, rather than always creating a patch from the previous succeeded release revision.

This change solves cases proveded in the issue: https://github.com/helm/helm/issues/10363.

### DIFF
--- a/pkg/action/upgrade.go
+++ b/pkg/action/upgrade.go
@@ -181,20 +181,15 @@ func (u *Upgrade) prepareUpgrade(name string, chart *chart.Chart, vals map[strin
 		return nil, nil, errPending
 	}
 
-	var currentRelease *release.Release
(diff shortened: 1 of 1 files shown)


bacongobbler:
Hi @distorhead. A couple of things:

- please sign off on the DCO. More information can be found in the failing github checkL https://github.com/helm/helm/pull/10364/checks?check_run_id=4249826379
- generally speaking we ask for unit tests to coincide with any changes. There are several significant changes here to the main code flow for `helm install` and `helm upgrade` here so we want to make sure they're well tested before we roll these changes out to users.

Let me know when you've addressed those and I can take a closer look.

PR #10369, opened 2021-11-18
Populate repo url in chart meta

### LINKED ISSUE #4256: Show Installed Chart Source Location or Source Repository URL
Use case: Cluster Admin needs to determine where an installed chart came from purely from querying a running cluster.  i.e Which file location or chart repository or whatever each chart came from.  The given admin may or may not have access to the original chart yaml.  

 It appears that none of the helm commands are able to return this information, and that said information is probably not recorded in the first place.  

It would be useful as an option to the helm list/ls command and the helm get command.

Output of `helm version`:
Client: &version.Version{SemVer:"v2.7.2", GitCommit:"8478fb4fc723885b155c924d1c8c410b7a9444e6", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.7.2", GitCommit:"8478fb4fc723885b155c924d1c8c410b7a9444e6", GitTreeState:"clean"}

Output of `kubectl version`:
Client Version: version.Info{Major:"1", Minor:"8", GitVersion:"v1.8.6", GitCommit:"6260bb08c46c31eea6cb538b34a9ceb3e406689c", GitTreeState:"clean", BuildDate:"2017-12-21T06:34:11Z", GoVersion:"go1.8.3", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"9", GitVersion:"v1.9.0", GitCommit:"925c127ec6b946659ad0fd596fa959be43f0cc05", GitTreeState:"clean", BuildDate:"2017-12-15T20:55:30Z", GoVersion:"go1.9.2", Compiler:"gc", Platform:"linux/amd64"}

AWS

lucasreed:
This PR closes #4256
This will add a new field to the release.chart.metadata stored in the release object (regardless of storage driver) called `repoURL` so that a given release will note what repository this chart was installed from. This solves a problem of not knowing if a forked version of a chart was used initially.

This field will only be populated if the chart is installed with a remote repository. It will be omitted if the chart is installed from a local filesystem, which I think is ok?

Here's the output of the metadata object stored in the cluster with the secrets storage driver (default) after making my changes:

```json
❯ k get secret sh.helm.release.v1.goldilocks.v1 -o json |jq -r .data.release | base64 -D|base64 -D|gunzip|jq .chart.metadata
{
  "name": "goldilocks",
  "sources": [
    "https://github.com/FairwindsOps/goldilocks"
  ],
  "version": "4.0.1",
  "description": "A Helm chart for running Fairwinds Goldilocks. See https://github.com/FairwindsOps/goldilocks ",
  "keywords": [
    "goldilocks",
    "resources",
    "kubernetes"
  ],
  "maintainers": [
    {
      "name": "sudermanjr"
    },
    {
      "name": "lucasreed"
    }
  ],
  "icon": "https://raw.githubusercontent.com/FairwindsOps/charts/master/stable/goldilocks/icon.png",
  "apiVersion": "v2",
  "appVersion": "v4.0.0",
  "repoURL": "https://charts.fairwinds.com/stable"
}
```

This is my first time contributing to the helm project, so I hope I'm not missing something that this would affect elsewhere, but tests seem to pass and the tool still functions 😅 . Happy to make adjustments where necessary, of course!

**If applicable**:

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -750,8 +750,8 @@ func (c *ChartPathOptions) LocateChart(name string, settings *cli.EnvSettings) (
 		return "", err
 	}
 
-	filename, rURL, _, err := dl.DownloadTo(name, version, settings.RepositoryCache)
-	c.RepoURL = rURL
+	filename, _, err := dl.DownloadTo(name, version, settings.RepositoryCache)
+	c.RepoURL = dl.RepositoryURL
 	if err == nil {
 		lname, err := filepath.Abs(filename)
 		if err != nil {
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -750,7 +750,8 @@ func (c *ChartPathOptions) LocateChart(name string, settings *cli.EnvSettings) (
 		return "", err
 	}
 
-	filename, _, err := dl.DownloadTo(name, version, settings.RepositoryCache)
+	filename, rURL, _, err := dl.DownloadTo(name, version, settings.RepositoryCache)
+	c.RepoURL = rURL
 	if err == nil {
 		lname, err := filepath.Abs(filename)
 		if err != nil {
--- a/pkg/downloader/chart_downloader.go
+++ b/pkg/downloader/chart_downloader.go
@@ -160,6 +160,10 @@ func (c *ChartDownloader) ResolveChartVersion(ref, version string) (*url.URL, er
 		return nil, errors.Errorf("invalid chart URL format: %s", ref)
 	}
 
+	if u.Scheme == "oci" {
+		c.RepositoryURL = ref
+	}
+
 	rf, err := loadRepoConfig(c.RepositoryConfig)
 	if err != nil {
 		return u, err
--- a/pkg/downloader/chart_downloader.go
+++ b/pkg/downloader/chart_downloader.go
@@ -72,6 +72,7 @@ type ChartDownloader struct {
 	RegistryClient   *registry.Client
 	RepositoryConfig string
 	RepositoryCache  string
+	RepositoryURL    string
 }
 
 // DownloadTo retrieves a chart. Depending on the settings, it may also download a provenance file.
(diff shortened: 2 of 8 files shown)


lucasreed:
Updated so that OCI registries will also work. Tested locally with a temporary public OCI helm repo/registry hosted in ECR. That changes the repoURL field to look like this:

```yaml
"repoURL": "oci://${ACCOUNT_NUMBER}.dkr.ecr.us-east-1.amazonaws.com/goldilocks"
```

pkg/downloader/chart_downloader.go
```diff
@@ -83,22 +84,22 @@ type ChartDownloader struct {
 //
 // For VerifyNever and VerifyIfPossible, the Verification may be empty.
 //
-// Returns a string path to the location where the file was downloaded and a verification
-// (if provenance was verified), or an error if something bad happened.
-func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *provenance.Verification, error) {
```
mattfarina, line 88: `DownloadTo` is a function that's part of the public API to the package. This can't change until Helm has a major version increment. Helm follows semantic versioning (like Go packages are supposed to for Go modules). This signature can't change.

What we typically do is add a new function and document the old one as deprecated. One can even use the other.

pkg/action/install.go
```diff
@@ -750,7 +750,8 @@ func (c *ChartPathOptions) LocateChart(name string, settings *cli.EnvSettings) (
 		return "", err
 	}
 
-	filename, _, err := dl.DownloadTo(name, version, settings.RepositoryCache)
+	filename, rURL, _, err := dl.DownloadTo(name, version, settings.RepositoryCache)
+	c.RepoURL = rURL
```
mattfarina, line 754: The RepoURL should be reset at upgrade time each time. This is because the chart location can change. For example, let's say I install wordpress from the bitnami repo. Then I decide to vendor it into my own repo. From my own repo I perform an upgrade. The upgraded release should show my vendor repo location instead of the original install location.

mattfarina (changes requested):
I like the idea of recording this information. I wonder if this should be a Helm Improvement Proposal (HIP) to make sure we do it well and think through everything.

I found a couple basic issues with the PR on a quick review.

pkg/action/install.go
lucasreed: I think I cover that functionality in the `upgrade.go` file here: https://github.com/helm/helm/pull/10369/files#diff-8f7c1d7e2cfeb70c465f36198e54a053fb517420d8647ffaf72a15e5525eb596R155-R157

That being said, if I'm adding a new function and not updating the `DownloadTo` function I think maybe I can do this in a better way that applies to both install/upgrade commands.

lucasreed:
Thanks for the review @mattfarina! I'm happy to write-up/submit a HIP to the community repo. Should we keep this PR around to help track? I can make adjustments to my implementation based on the feedback you gave if we do keep this PR around.

pkg/action/install.go
lucasreed: Apparently that link doesn't do what I hoped it did. Here's the snippet from `cmd/upgrade.go`. Looking at it again I think you're right. My intention was to update the `repoURL` during an upgrade as well, but I think I may have missed that. I'll fix it up.

pkg/downloader/chart_downloader.go
lucasreed: Reverted this and achieved a similar result with a much simpler implementation. Thank you for helping guide me in that direction!

bitfisher:
unstale please

lucasreed:
This PR has been superseded by #11378

PR #10370, opened 2021-11-18
Fix specifying of Kubernetes version from build scripts

### LINKED ISSUE #10367: Builds Failing due to use of -X to set non-string during build
Output of `helm version`: tip of main

CI builds are failing because of the use of -X in the build scripts to set versions which are not strings.

This happened because of three things:

- The normal tests run on a PR do not test building Helm. The build scripts were not tested until the most recent change landed.
- #8608 made changes around linting. LDFLAGS -X were used to set a constants which are silently ignored.
- #10325 moved those constants to vars so they could be set.

This was not caught by CI on the PR but was caught when the second PR was merged into main.

mattfarina:
This bug came about because of three issues that this change
corrects:
- The CI scripts run on a pull request did not test building
  Helm. This means that a failure to set a variable using LDFLAGS
  had no opportunity to be caught.
- #8608 provided a means to match the k8s version used in linting
  and chartutil with the version of the package we pull in. With
  one problem. It attempts to set a const as if it were a string.
  This is ignored and everyone missed it.
- #10325 moved those constants to vars so it could be set. This
  looked good and passed tests but missed that you can't set an
  int as if it were a string. See first bullet.

This change fixes this by moved the internal representation to
be a string. These are internal variables not exposed in the public
API which makes this change non-breaking to the API.

Closes #10367

**What this PR does / why we need it**: It fixes Helm on main being unable to be built right now.

**If applicable**:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,6 +26,9 @@ jobs:
       - run:
           name: test
           command: make test-coverage
+      - run:
+          name: test build
+          command: make
       - deploy:
           name: deploy
           command: .circleci/deploy.sh
--- a/pkg/chartutil/capabilities.go
+++ b/pkg/chartutil/capabilities.go
@@ -29,8 +29,10 @@ import (
 )
 
 var (
-	k8sVersionMajor = 1
-	k8sVersionMinor = 20
+	// The Kubernetes version can be set by LDFLAGS. In order to do that the value
+	// must be a string.
+	k8sVersionMajor = "1"
+	k8sVersionMinor = "20"
 
 	// DefaultVersionSet is the default version set, which includes only Core V1 ("v1").
 	DefaultVersionSet = allKnownVersions()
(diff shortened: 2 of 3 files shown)


bridgetkromhout:
I pulled down the branch with this PR and tested `make` locally. It built without errors and produced a plausible artifact:

$ bin/helm version
version.BuildInfo{Version:"v3.7+unreleased", GitCommit:"7838fb769d00f53c858d994a873f42c2960a7d25", GitTreeState:"clean", GoVersion:"go1.16.3"}

jdolitsky (approved):

bacongobbler (approved):
LGTM

mattfarina:
It looks like CodeQL ran but something stopped the status from updating. See https://github.com/helm/helm/actions/runs/1477452626

bacongobbler:
Odd... I'm okay with merging this. Feel free to merge when ready.

mattfarina:
@bacongobbler thanks. I opened an issue with GitHub because rerunning CI didn't cause it to run and it never timed out. Giving them a short bit of time to look at this.

PR #10371, opened 2021-11-18
allow ldflags to overwrite k8s version

### LINKED ISSUE #10324: Helm lint does not report api deprecations in k8s 1.21+

sboschman:
ldflags must be vars of type string

fixes #10324, #10367

### DIFF
--- a/pkg/chartutil/capabilities.go
+++ b/pkg/chartutil/capabilities.go
@@ -29,8 +29,10 @@ import (
 )
 
 var (
-	k8sVersionMajor = 1
-	k8sVersionMinor = 20
+	// This should be set in the Makefile based on the version of client-go being imported.
+	// These constants will be overwritten with LDFLAGS (ldflags should be var and type string)
+	k8sVersionMajor = "1"
+	k8sVersionMinor = "20"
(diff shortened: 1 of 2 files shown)


sboschman:
in favor of #10370

PR #10378, opened 2021-11-19
Add --output flag to `helm dependency list`

### LINKED ISSUE #10345: Allow different output formats in `helm dependency list`
The current `helm dependency list` subcommand only allows configuring the standard's output column width.

However, for automated processes it'd be nice to have a way to configure the output format, so a user could run `helm dep list -o json` (or `yaml`, as in other commands). It doesn't seem complicated to show a JSON or YAML view of the standard table:

```json
[
  {
    "name": "mariadb",
    "version": "9.x.x",
    "repository": "https://charts.bitnami.com/bitnami",
    "status": "missing"
  }
]
```

```yaml
---
- name: mariadb
  version: 9.x.x
  repository: https://charts.bitnami.com/bitnami
  status: missing
```

I believe this minor proposal doesn't require a HIP.

Output of `helm version`:

> version.BuildInfo{Version:"v3.7.1", GitCommit:"1d11fcb5d3f3bf00dbe6fe31b8412839a96b3dc4", GitTreeState:"clean", GoVersion:"go1.17.2"}

Output of `kubectl version`:

Doesn't apply.

Doesn't apply.

paleloser:
Closes #10345

This adds the `--output` flag to `helm dependency list` command:

**JSON**

```sh
$ helm dep list bitnami/kubeapps -o json
[{"name":"common","version":"1.x.x","repository":"https://charts.bitnami.com/bitnami","status":"missing"},{"name":"postgresql","version":"10.x.x","repository":"https://charts.bitnami.com/bitnami","status":"missing"},{"name":"redis","version":"15.x.x","repository":"https://charts.bitnami.com/bitnami","status":"missing"}]
```
**YAML**

```sh
- name: common
  repository: https://charts.bitnami.com/bitnami
  status: missing
  version: 1.x.x
- name: postgresql
  repository: https://charts.bitnami.com/bitnami
  status: missing
  version: 10.x.x
- name: redis
  repository: https://charts.bitnami.com/bitnami
  status: missing
  version: 15.x.x
```

* Unlike other command implementations, which dedicate a file for the `list` subcommand, here its logic was bundled in the file of the main command (`cmd/helm/dependency.go`). I moved the `list` command to a dedicated `cmd/helm/dependency_list.go`.
* Some cmd-related logic lived inside `pkg/action/dependency.go`. I moved out of that file the logic dedicated to print the list of dependencies to the new `cmd/helm/dependency_list.go`. This way, the `action` file only contains the _internal_ logic to list the dependencies.
* Some `WARNING` logs taking place when there were missing deps are only logged in `table` mode. This is to not interfere the JSON/YAML output, which could cause trouble to machines parsing the output of the CLI.
* There were 2 extra empty lines printed at the end of this command instead of 1. I removed the extra `print` so it only adds one.

**If applicable**:
- [x] this PR contains unit tests (kind of, it refactors some of them)


pkg/action/dependency.go
```diff
@@ -48,25 +55,28 @@ func NewDependency() *Dependency {
 }
 
 // List executes 'helm dependency list'.
-func (d *Dependency) List(chartpath string, out io.Writer) error {
```
mattfarina, line 51: Functions that are part of the public interface can't have their name or signature changed in a minor version. Helm follows semantic versioning so breaking changes can only happen in major versions. What we typically do is add a new function with the new signature. If the old function is superseded we mark it as deprecated in a comment.

mattfarina (changes requested):
Thanks for submitting the change. This feature is one that would be nice to have.

While I have not had a chance to give this a thorough review, I did find one thing that must be changed.

pkg/action/dependency.go
paleloser: Thanks for pointing that out! I recovered back the original signatures for the public functions in 3c47f3b3. Also, I've added a deprecation note in this one. Please if it should have a different structure let me know.

PR #10381, opened 2021-11-21
Not possible to pass values to second instance of sub sub chart with alias

### LINKED ISSUE #10407: Values not passed to second instance of aliased sub sub chart
Output of `helm version`:
```
version.BuildInfo{Version:"v3.7.1", GitCommit:"1d11fcb5d3f3bf00dbe6fe31b8412839a96b3dc4", GitTreeState:"clean", GoVersion:"go1.17.2"}
```
Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.13", GitCommit:"2444b3347a2c45eb965b182fb836e1f51dc61b70", GitTreeState:"clean", BuildDate:"2021-11-17T13:05:33Z", GoVersion:"go1.15.15", Compiler:"gc", Platform:"darwin/amd64"}
```

## Problem

I came across this bug when trying to create multiple instances of a of dependency that in turn had aliased dependencies. Can be reproduced with the following chart.

```yml
apiVersion: v2
name: parent
description: parent
type: application
version: 0.0.1
dependencies:
- name: sub-chart
  version: 0.0.1
  alias: sub1
- name: sub-chart
  version: 0.0.1
  alias: sub2
```

```yml
apiVersion: v2
name: sub-chart
description: sub-chart
type: application
version: 0.0.1
dependencies:
- name: sub-sub-chart
  version: 0.0.1
  alias: sub-sub1
```

In sub-sub-chart we have a template like this.
```yml
name: {{ required "name is required" .Values.name }}
```

Using this values file for the parent chart.
```yml
sub1:
  sub-sub1:
    name: foo
sub2:
  sub-sub1:
    name: baz
```

Testing it out gives this error
```
helm template test ./subsubalias
Error: execution error at (parent/charts/sub2/charts/sub-sub-chart/templates/test.yaml:1:9): name is required
```

Values are not getting passed to `sub2.sub-sub1`.

From my limited understanding of the code base, it seems the bug comes from the fact that a requirement with an alias has its name overwritten to the alias value at some point. And if that requirement is used again it cannot be matched against an existing chart by name. See linked PR with chart example and proposed code fix: https://github.com/helm/helm/pull/10381

thrawny:
Resolves https://github.com/helm/helm/issues/10407

I came across this bug when trying to create multiple instances of a chart that in turn had multiple instances of a sub chart and was using aliases. Don't think this is a very common use case, but I thought it would be good/educational to try and fix it.

From my limited understanding of the code base, it seems the bug comes from the fact that a requirement with an alias has its name overwritten to the alias value at some point. And if that requirement is used again it cannot be matched against an existing chart by name.

This was reproducible with a test case and a minimal chart example.

I did also fix the bug by introducing a new field called `orignalName` and a method `equalName` on `Dependency` that should be used to match requirements and charts.

Since I am not super well versed in the code base I don't know if the proposed solution is good enough and if there are any issues with backwards compat and so on. So happy change/drop it according to any feedback given.

**If applicable**:

### DIFF
--- a/pkg/chart/dependency.go
+++ b/pkg/chart/dependency.go
@@ -47,6 +47,23 @@ type Dependency struct {
 	ImportValues []interface{} `json:"import-values,omitempty"`
 	// Alias usable alias to be used for the chart
 	Alias string `json:"alias,omitempty"`
+
+	originalName string
+}
+
+// SetNameFromAlias sets originalName if empty, so can be called multiple times.
+// This assumes that the first time SetNameFromAlias is called, d.Name is the name as
+// specified in the dependency's Chart.yaml
+func (d *Dependency) SetNameFromAlias() {
+	if d.originalName == "" {
+		d.originalName = d.Name
+	}
+	d.Name = d.Alias
+}
+
+// EqualName returns true if the dependency's name or originalName is equal to the given name.
+func (d *Dependency) EqualName(name string) bool {
+	return d.Name == name || d.originalName == name
 }
 
 // Validate checks for common problems with the dependency datastructure in
--- a/pkg/chartutil/dependencies.go
+++ b/pkg/chartutil/dependencies.go
@@ -95,7 +95,7 @@ func getAliasDependency(charts []*chart.Chart, dep *chart.Dependency) *chart.Cha
 		if c == nil {
 			continue
 		}
-		if c.Name() != dep.Name {
+		if !dep.EqualName(c.Name()) {
 			continue
 		}
 		if !IsCompatibleRange(dep.Version, c.Metadata.Version) {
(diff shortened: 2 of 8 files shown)


zonggen:
Thanks for submitting a patch. Could you create an issue for the bug and provide a simple example to reproduce it? (ref: https://github.com/helm/helm/blob/main/CONTRIBUTING.md#how-to-contribute-a-patch). This will help raise more visibility and move your patch forward.

thrawny:
Thx for the input! Kind of missed the issue creation part. Created one here: https://github.com/helm/helm/issues/10407

PR #10391, opened 2021-11-23
Add suggestion to error message when dependencies missing

howardjohn:
This provides an actionable next step to users when they fail to render due to missing dependencies.

**If applicable**:


bacongobbler:
With a few exceptions, we try to avoid avoid referring to Helm CLI commands in `pkg/action`. This is because `pkg/action` can be used by other programs as a library, so asking the user to run `helm repo update` may be completely irrelevant.

In any case, this has been resolved in https://github.com/helm/helm/pull/9794 which I just merged today.

PR #10397, opened 2021-11-24
Adjust error message wrongly claiming that there is a resource conflict

mtesseract:

This PR fixes a misleading error message. Within our CI we stumbled upon the following error message as a result of an attempted `helm install`:

```
Error: rendered manifests contain a resource that already exists. Unable to continue with install: could not get information about the resource: an error on the server ("Internal Server Error: [...]: the server is currently unable to handle the request") has prevented the request from succeeding [...]
```

As can be seen from this error message, an internal server error is -- at the end -- interpreted as a resource conflict error.
By looking at the Helm source (current master) I noticed the following:

The error message about the resource conflict is added in `pkg/action/install.go`, line 275:
```go
	if !i.ClientOnly && !isUpgrade && len(resources) > 0 {
		toBeAdopted, err = existingResourceConflict(resources, rel.Name, rel.Namespace)
		if err != nil {
			return nil, errors.Wrap(err, "rendered manifests contain a resource that already exists. Unable to continue with install")
		}
	}
```

as a result of `existingResourceConflict()` returning a non-`nil` `err`. But looking at that function, we can see that the `could not get information about the resource` error is returned (l.54) in case it was not possible to look up the resource in question, while the case of the resource being already present is handled later on (l.58-59), including a specific error wrapping. Here is the function:

```go
func existingResourceConflict(resources kube.ResourceList, releaseName, releaseNamespace string) (kube.ResourceList, error) {
	var requireUpdate kube.ResourceList

	err := resources.Visit(func(info *resource.Info, err error) error {
		if err != nil {
			return err
		}

		helper := resource.NewHelper(info.Client, info.Mapping)
		existing, err := helper.Get(info.Namespace, info.Name)
		if err != nil {
				return nil
			}
		}

		// Allow adoption of the resource if it is managed by Helm and is annotated with correct release name and namespace.
		if err := checkOwnership(existing, releaseName, releaseNamespace); err != nil {
			return fmt.Errorf("%s exists and cannot be imported into the current release: %s", resourceString(info), err)
		}

		requireUpdate.Append(info)
		return nil
	})

	return requireUpdate, err
}
```

The same issue exists in one other place: `pkg/action/upgrade.go`, l.278.

**If applicable**:

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -272,7 +272,7 @@ func (i *Install) Run(chrt *chart.Chart, vals map[string]interface{}) (*release.
 	if !i.ClientOnly && !isUpgrade && len(resources) > 0 {
 		toBeAdopted, err = existingResourceConflict(resources, rel.Name, rel.Namespace)
 		if err != nil {
-			return nil, errors.Wrap(err, "rendered manifests contain a resource that already exists. Unable to continue with install")
+			return nil, errors.Wrap(err, "Unable to continue with install")
 		}
 	}
 
--- a/pkg/action/upgrade.go
+++ b/pkg/action/upgrade.go
@@ -277,7 +277,7 @@ func (u *Upgrade) performUpgrade(originalRelease, upgradedRelease *release.Relea
 
 	toBeUpdated, err := existingResourceConflict(toBeCreated, upgradedRelease.Name, upgradedRelease.Namespace)
 	if err != nil {
-		return nil, errors.Wrap(err, "rendered manifests contain a resource that already exists. Unable to continue with update")
+		return nil, errors.Wrap(err, "Unable to continue with update")
 	}
 
 	toBeUpdated.Visit(func(r *resource.Info, err error) error {


bacongobbler (dismissed):
Can you please add a conditional `if err == ...` instead of removing the error message entirely? This seems like more of an edge case. An overwhelming majority of error cases is due to resource conflicts, which is why more information on the error was introduced.

mtesseract:
Hi, I am a bit confused now. My PR (removing the part in the error message about existing resources completely) is based on the following observation:

In the function `existingResourceConflict` we have two error exits after attempting to fetch the resource object:

The first one is here:
```go
existing, err := helper.Get(info.Namespace, info.Name)
if err != nil {
		return nil
	}
}
```

In case an error occured which is not `NotFound`, we return with error. Here it would be wrong to enrich this error with a message about resource-conflicts since that case is handled separately (below)?

Second is here:

```go
// Allow adoption of the resource if it is managed by Helm and is annotated with correct release name and namespace.
if err := checkOwnership(existing, releaseName, releaseNamespace); err != nil {
	return fmt.Errorf("%s exists and cannot be imported into the current release: %s", resourceString(info), err)
}
```

Here we check explicitly if the already existing resource could be adopted. If yes, no error. If no, then we have a resource conflict error and we already produce the right error message saying that resource in question already exists. Therefore, also in this case there would be no point in further enriching this message with "resource already exists", as it is already in the message.

Am I overlooking something?
Thanks.

mtesseract:
Hi @bacongobbler, just curious, is there anything else you need from me regarding this?

mtesseract:
@bacongobbler Can we get this through?

mtesseract:
Ping? Can we get this through?
@bacongobbler

joejulian (approved):

joejulian:
closing and re-opening to (hopefully) get the tests to run

PR #9760, opened 2021-06-01
feat: helm pull --repo to use repo creds from helm repos file

andreaskaris:
Up to this point, helm pull --repo required --username and --password
even if the repository had already been added via helm repo add.
    
From now on, check repositories in the repositories file and if the URL
matches, pull the chart with the username and password from the entry.

Fixes #9599

Signed-off-by: Andreas Karis <ak.karis@gmail.com>

**What this PR does / why we need it**:

Set up the following reproducer - set up the following directory structure for httpd:
```
yum install httpd -y
mkdir cd /var/www/html/charts
cd /var/www/html/charts
helm repo add brigade https://brigadecore.github.io/charts
helm pull brigade/brigade
helm pull brigade/kashti
helm repo index . --url http://127.0.0.1/charts
htpasswd -c .htpasswd testuser   # --> "password"
cat <<'EOF' > .htaccess
AuthType Basic
AuthName "Restricted Access"
AuthUserFile /var/www/html/charts/.htpasswd
Require valid-user
EOF
restorecon -R .

vi /etc/httpd/conf/httpd.conf
# <Directory "/var/www/html/charts"> 
#    AllowOverride AuthConfig
# </Directory>
```

Now, the repo should be password protected:
```
curl -u testuser:password  http://127.0.0.1/charts/index.yaml
```

```
# helm repo add localhost http://127.0.0.1/charts/ --username testuser --password password
# helm search repo localhost
NAME             	CHART VERSION	APP VERSION	DESCRIPTION                                       
localhost/brigade	1.10.0       	v1.5.0     	Brigade provides event-driven scripting of Kube...
localhost/kashti 	0.7.0        	v0.4.0     	A Helm chart for Kubernetes
# helm pull localhost/brigade
# ll | grep brigade
-rw-r--r--. 1 root root 14278 Feb 10 15:01 brigade-1.10.0.tgz
```

However, with --repo, this won't work:
```
# helm pull --repo http://127.0.0.1/charts/ brigade
Error: failed to fetch http://127.0.0.1/charts/brigade-1.10.0.tgz : 401 Unauthorized
```

The username and password must be provided:
```
# helm pull --repo http://127.0.0.1/charts/ brigade
Error: looks like "http://127.0.0.1/charts/" is not a valid chart repository or cannot be reached: failed to fetch http://127.0.0.1/charts/index.yaml : 401 Unauthorized
```

After applying the fix:
```
[...]

PR #10406, opened 2021-11-29
fix(helm): helm pull --repo to use repo creds from helm repos file

### LINKED ISSUE #9599: `helm pull --repo [url]` does not use the credentials set with `helm repo add`
When calling `helm dependency update`, or `helm dependency build`, `helm` will look for credentials in any repository set previously with `helm repo add`, no matter the alias given to the repository itself.

Such behavior is not retained when using `helm pull --repo [url]`.

The problem is that I'm using `helm pull` to download charts in a script, but some of these repos needs credentials to be set. While `helm dep up` works fine, the `helm pull --repo` does not as it does not take the credentials I have set before with `helm repo add`. I can't use `helm pull --repo [url] --user [user] --password [pass]` because I don't always know beforehand which repo I'm going to download from in the script, and thus different credentials would need be used for each different repo.

To work around it, I switched from `helm pull` to `curl` with `--netrc-optional`, and instead of `helm repo add`, I'm setting the credentials in the `~/.netrc`. The there are multiple caveats:

1. I have to "login" twice: `helm repo add` (for `helm dep up`) and `echo [creds] > ~/.netrc` (for downloading the charts with `curl`).
2. I have to build my own URL to fetch the chart, and while it's working for me, there is no rule for the chart location in a repository. `helm pull` would first look for the `index.yaml` as it was supposed to be, but my `curl` call will just use a hardcoded location for the `.tgz`, currently:  `<repo>/<chart-name>/<chart-name>-<chart-version>.tgz`.
3. I can only set credentials per host, and not per repo, as it would be possible with `helm repo add`.

Output of `helm version`:
```sh-session
❯ helm version
version.BuildInfo{Version:"v3.5.3", GitCommit:"041ce5a2c17a58be0fcd5f5e16fb3e7e95fea622", GitTreeState:"dirty", GoVersion:"go1.16.2"}
```

felipecrs:
This PR is a continuation of https://github.com/helm/helm/pull/9760. All the credits goes to @andreaskaris, I just fixed the comments provided by @yxxhero.

Fixes #9599

### DIFF
--- a/pkg/action/pull.go
+++ b/pkg/action/pull.go
@@ -122,7 +122,19 @@ func (p *Pull) Run(chartRef string) (string, error) {
 	}
 
 	if p.RepoURL != "" {
-		chartURL, err := repo.FindChartInAuthAndTLSAndPassRepoURL(p.RepoURL, p.Username, p.Password, chartRef, p.Version, p.CertFile, p.KeyFile, p.CaFile, p.InsecureSkipTLSverify, p.PassCredentialsAll, getter.All(p.Settings))
+		username := p.Username
+		password := p.Password
+		if username == "" && password == "" {
+			repoEntry, err := repo.FindRepoEntry(p.RepoURL, p.Settings.RepositoryConfig)
+			if err != nil {
+				return out.String(), err
+			}
+			if repoEntry != nil {
+				username = repoEntry.Username
+				password = repoEntry.Password
+			}
+		}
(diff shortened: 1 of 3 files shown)


scottrigby:
See https://github.com/helm/helm/pull/9760#issuecomment-1012213294

felipecrs:
@andreaskaris revived https://github.com/helm/helm/pull/9760, thus this PR is no longer necessary.

PR #10412, opened 2021-11-30
Expose NewRootCmd

kke:

Exposes the `newRootCmd` in `cmd/helm` as `NewRootCmd` so that it can be imported as a subcommand into other tools (mostly this would probably be kubernetes distributions that embed a bunch of tools). The functionality that was in `func init()` has been moved to `func main()` to avoid side-effects when importing the `github.com/helm/helm/cmd/helm`. 

There's still a lot happening in `main()` that need to be duplicated by the importer, but changing that will require a more significant refactoring of the command initialization.

### DIFF
--- a/cmd/helm/helm.go
+++ b/cmd/helm/helm.go
@@ -43,10 +43,6 @@ const FeatureGateOCI = gates.Gate("HELM_EXPERIMENTAL_OCI")
 
 var settings = cli.New()
 
-func init() {
-	log.SetFlags(log.Lshortfile)
-}
-
 func debug(format string, v ...interface{}) {
 	if settings.Debug {
 		format = fmt.Sprintf("[debug] %s\n", format)
@@ -60,6 +56,8 @@ func warning(format string, v ...interface{}) {
 }
 
 func main() {
+	log.SetFlags(log.Lshortfile)
+
 	// Setting the name of the app for managedFields in the Kubernetes client.
 	// It is set here to the full name of "helm" so that renaming of helm to
 	// another name (e.g., helm2 or helm3) does not change the name of the
@@ -67,7 +65,7 @@ func main() {
 	kube.ManagedFieldsManager = "helm"
 
 	actionConfig := new(action.Configuration)
-	cmd, err := newRootCmd(actionConfig, os.Stdout, os.Args[1:])
+	cmd, err := NewRootCmd(actionConfig, os.Stdout, os.Args[1:])
 	if err != nil {
 		warning("%+v", err)
 		os.Exit(1)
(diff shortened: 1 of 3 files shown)


bacongobbler:
The intent of the `cmd` is to intentionally not expose these calls to the SDK, as the implementation details of `cmd` can change at any point. For example, we may swap out cobra with another CLI framework, which could change the entire structure of `cmd` including `NewRootCmd` because its function signature currently returns `(*cobra.Command, error)`.

Users should instead be relying on packages exposed in `pkg` like `pkg/cli`.

What parts of the command-line interface are you trying to expose? Perhaps we can start there to figure out what parts need to be moved over to `pkg`.

bacongobbler:
From https://github.com/helm/helm/blob/main/CONTRIBUTING.md#semantic-versioning:

> - Go libraries inside of pkg/ SHOULD remain backward compatible, though code inside of cmd/ and internal/ may be changed from release to release without notice.

kke:
> What parts of the command-line interface are you trying to expose? Perhaps we can start there to figure out what parts need to be moved over to `pkg`.

The goal is to have the whole helm CLI embedded as a subcommand like we have already done for kubectl and containerd. We are already using parts of helm for some operations, but including the whole helm CLI as a bundled subcommand would be quite handy for the users and require very little code if the root command just was importable. Doing it presently would require duplicating the entire contents of `cmd/helm` or creating some kind of patcher / generator that would expose the newRootCmd for importing. It's doable of course if you don't want the burden of having an exposed function in `cmd/`.

The nice thing in the kube ecosystem is that all of the essential tools are written in go and mostly use cobra (which I myself am not a big fan of) so something like this can be very simple to do.

bacongobbler:
> The goal is to have the whole helm CLI embedded as a subcommand

I don’t believe this is something we plan to support. The behaviour of the CLI can and does change often enough that we cannot guarantee code safety from one version to the next. For example, what if our code registers a global flag? That could override other CLI’s feature flags embedding Helm. That happened to us with kubectl’s `--verbose` flag in Helm when we embedded `klog`. It is very possible that other flags we provide may clobber with the CLI it is attempting to be embedded within.

In other words, I’m not comfortable making the same guarantees with `cmd` as we do with the rest of our codebase (and realistically, I may never feel different about that), so I’m 👎🏼 on this proposal. However if there are parts of `cmd` that may be useful to other CLIs (read: things like flag parsers or reading confit from multiple sources… not embedding Helm in another CLI), then I’m okay with that.

kke:
Yep, kubectl also adds some global flags on import that need to be hidden when embedding. I'm ok with this decision.

PR #10440, opened 2021-12-06
Fix memory leak in upgrade action

### LINKED ISSUE #10439: Memory leak in the upgrade action
During the upgrade action we noticed a memory leak which only occurs when helm is used as a "library" and not as a cli tool.

Actually, there are two places:

1. The [`performUpgrade`](https://github.com/helm/helm/blob/1d11fcb5d3f3bf00dbe6fe31b8412839a96b3dc4/pkg/action/upgrade.go#L260) method waits [at this point](https://github.com/helm/helm/blob/1d11fcb5d3f3bf00dbe6fe31b8412839a96b3dc4/pkg/action/upgrade.go#L326) for the `rChan`, which is either filled by the `releasingUpgrade` or `handleContext` method. If the result comes from `releasingUpgrade()`, then [`handleContext()`](https://github.com/helm/helm/blob/1d11fcb5d3f3bf00dbe6fe31b8412839a96b3dc4/pkg/action/upgrade.go#L344) will wait until the context ([`ctx`](https://github.com/helm/helm/blob/1d11fcb5d3f3bf00dbe6fe31b8412839a96b3dc4/pkg/action/upgrade.go#L347)) is cancelled, which in most cases will never happen and keeps the routine running.

2. If [`ctx`](https://github.com/helm/helm/blob/1d11fcb5d3f3bf00dbe6fe31b8412839a96b3dc4/pkg/action/upgrade.go#L347) is indeed cancelled, `handleContext()` calls `reportToPerformUpgrade()` which fills the already [consumed `rChan`](https://github.com/helm/helm/blob/1d11fcb5d3f3bf00dbe6fe31b8412839a96b3dc4/pkg/action/upgrade.go#L339). However, this is never read because the only consumer (`performUpgrade()`) has already returned; The routine remains running forever.

---

Output of `helm version`:
```shell
version.BuildInfo{Version:"v3.7.1", GitCommit:"1d11fcb5d3f3bf00dbe6fe31b8412839a96b3dc4", GitTreeState:"clean", GoVersion:"go1.16.9"}
```

jkmw:
This PR fixes the memory leak in the upgrade action when helm is used as library.

fixes helm/helm#10439

### DIFF
--- a/pkg/action/upgrade.go
+++ b/pkg/action/upgrade.go
@@ -325,15 +325,13 @@ func (u *Upgrade) performUpgrade(ctx context.Context, originalRelease, upgradedR
 	doneChan := make(chan interface{})
 	go u.releasingUpgrade(rChan, upgradedRelease, current, target, originalRelease)
 	go u.handleContext(ctx, doneChan, ctxChan, upgradedRelease)
-	var result resultMessage
 	select {
-	case result = <-rChan:
+	case result := <-rChan:
 		doneChan <- true
-	case result = <-ctxChan:
-		close(rChan)
+		return result.r, result.e
+	case result := <-ctxChan:
+		return result.r, result.e
 	}
-
-	return result.r, result.e
 }
 
 // Function used to lock the Mutex, this is important for the case when the atomic flag is set.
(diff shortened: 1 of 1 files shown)


bacongobbler:
Hi @jkmw. Can you please look at the failing tests? It does appear there are cases where your change may be breaking something:

> panic: send on closed channel

jkmw:
done

mattfarina (approved):

technosophos (approved):
LGTM.

PR #10453, opened 2021-12-10
Remove runTestActionCmd and use runTestCmd instead

SimonAlling:
`runTestCmd` is identical to `runTestActionCmd` except that it

  * runs the command multiple times if the `repeat` property is greater
    than zero,
  * calls its storage `storage` instead of `store`,
  * fails if the `Create` method returns an error, and
  * logs a "running cmd" message.

This is most easily demonstrated by copying the functions to their own
files and diffing the files with this command:

    git diff --no-index --ignore-space-change --color-words=. runTestActionCmd.go runTestCmd.go

Also, `runTestCmd` is used 32 times; `runTestActionCmd` only 1.

### DIFF
--- a/cmd/helm/helm_test.go
+++ b/cmd/helm/helm_test.go
@@ -71,27 +71,6 @@ func runTestCmd(t *testing.T, tests []cmdTestCase) {
 	}
 }
 
-func runTestActionCmd(t *testing.T, tests []cmdTestCase) {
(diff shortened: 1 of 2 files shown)


SimonAlling:
Status on this?

hickeyma (approved):
LGTM, thanks @SimonAlling

Issue #10432, opened 2021-12-04
Feature request: overall timeout option 


caleb15:
Output of `helm version`: version.BuildInfo{Version:"v3.5.2", GitCommit:"167aac70832d3a384f65f9745335e9fb40169dc2", GitTreeState:"dirty", GoVersion:"go1.15.7"}

Output of `kubectl version`: Client Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.4", GitCommit:"b695d79d4f967c403a96986f1750a35eb75e75f1", GitTreeState:"clean", BuildDate:"2021-11-17T15:48:33Z", GoVersion:"go1.16.10", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"21+", GitVersion:"v1.21.2-eks-0389ca3", GitCommit:"8a4e27b9d88142bbdd21b997b532eb6d493df6d2", GitTreeState:"clean", BuildDate:"2021-07-31T01:34:46Z", GoVersion:"go1.16.5", Compiler:"gc", Platform:"linux/amd64"}

Cloud Provider/Platform (AKS, GKE, Minikube etc.): EKS

Helm has a timeout option for individual kubernetes operations, like jobs for hooks. However, it does not have a overall timeout option for the entire upgrade/install. A overall timeout would be useful so helm knows to cancel the deloyment and clean up if it takes too long. Otherwise you run into issues where a supervising process/system (ex: CircleCI) hard-cancels the depoyment after the supervisor's timeout, leaving the release in a bad state, like pending-rollback.

caleb15:
Here's an example of another person who ran into this issue: [Kubernetes slack link](https://kubernetes.slack.com/archives/C0NH30761/p1638264516075700)

llu94:
If it's ok with everyone here I'd like to try my hand at this. Expect something by the end of the weekend out of me. Unless someone has already started working on this.

caleb15:
I haven't started working on it yet, go ahead :)

llu94:
Quick update: I have run into some data races which is always very fun. Will work on this a bit more and maybe try another track if I can't untangle the mess. Stay tuned.

llu94:
@caleb15 @mattfarina I have a functioning implementation of an overall timeout for rollback operations in Helm. Please take a look whenever convenient. If it looks good, I will cut another PR for upgrades and installs.

llu94:
Does the community still want this feature/does anyone want to review my code?

PR #10457, opened 2021-12-13
Timeout For Rollback Operation [Issue: 10432]

llu94:
This pull request implements an overall timeout command line option for the rollback operation. 

Signed-off-by: Lucas Lu <lucasludev@gmail.com>

This PR implements the timeout for rollbacks requested in the following issue.
 'refs #10432'
https://github.com/helm/helm/issues/10432

This PR is a proof of concept. If the reviewers are happy with it, similar work will be done for install and upgrade timeouts as per the referenced issue. Timeouts were requested for upgrade and install functions too.

**If applicable**:
- [ X ] this PR contains unit tests

### DIFF
--- a/pkg/action/upgrade.go
+++ b/pkg/action/upgrade.go
@@ -63,8 +63,6 @@ type Upgrade struct {
 	SkipCRDs bool
 	// Timeout is the timeout for individual hooks/kubernetes operations
 	Timeout time.Duration
-	// UpgradeTimeout is the timeout for this entire upgrade operation
-	UpgradeTimeout time.Duration
 	// Wait determines whether the wait operation should be performed after the upgrade is requested.
 	Wait bool
 	// WaitForJobs determines whether the wait operation for the Jobs should be performed after the upgrade is requested.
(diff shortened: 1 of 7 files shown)


pkg/action/upgrade.go
```diff
@@ -475,6 +477,7 @@ func (u *Upgrade) failRelease(rel *release.Release, created kube.ResourceList, e
 		rollin.Recreate = u.Recreate
 		rollin.Force = u.Force
 		rollin.Timeout = u.Timeout
+		rollin.RollbackTimeout = u.UpgradeTimeout / 2
```
caleb15, line 480: UpgradeTimeout is never assigned a value, so wouldn't this division operation fail?

(I have very little go experience so I might be missing something here)

caleb15: Wait a second - https://stackoverflow.com/a/4223332

> because by default all variables (including dynamically allocated ones) are set to “zero values” according to type (numbers zero, references nil)

weeeiirrrrddddd

So this would be 0/2, which is 0. Would this cause the rollback to immediately timeout?

llu94: @caleb15 I've added some defaults to deal with the upgrade timeout dangling issue. Please review,

caleb15 (approved):
Looks fine to me, but I should mention I'm not a go programmer, so this would still need to be reviewed by someone who knows go.

PR #10463, opened 2021-12-15
feat(lint): Allow customization of APIVersions when using helm lint

### LINKED ISSUE #10190: Add ability to pass Capabilities to lint
With #5392 support was added to helm template to provide capabilities. It would be great if that would also be possible for helm lint. 

e.g. currently I'm getting warnings that the produced ingress is deprecated, but this is only because the chart supports newer and older versions.

This was originally requested as part of #3377 but I think it got lost

amuniz:
Usage: `helm lint --api-versions []string`

closes #10190

**What this PR does / why we need it**: Adds `--api-versions` support for `helm lint`

**If applicable**:


amuniz:
I don't have the time for following up here. Feel free to take over ownership.

PR #10513, opened 2021-12-31
Allow fuzzy matching during shell completion for zsh and fish

marckhouzam:

This PR improves shell completion for `zsh` and `fish` by allowing for fuzzy matching.

We had made the assumption that when doing shell completion, we only needed choices that had for *prefix* what the user has typed. However, the `zsh` and `fish` shells have a more advanced matching system which first matches on prefix, but if no match is found, then does more advanced matching attempts, such as sub-strings; `fish` even matches
on descriptions of completions.

For example, with both `zsh` and `fish`:
```
$ bin/helm -n test status <TAB>
test-api          test-api-jwt      test-jaeger       test-node-api     test-nodelts-api
$ bin/helm  -n test status jaeger<TAB>
$ bin/helm  -n test status test-jaeger
```
As can be seen in the above example, the fuzzy matching allows the user to avoid typing (or completing) the prefix `test-` before reaching the differentiating part of the release name.  Instead they can simply type the differentiating string (`jaeger` in the example above) and the shell will figure out the rest.

Another example I expect to be common will be for namespace names.  In my clusters, the naming scheme uses prefixes to group namespaces, which means there are a lot of common prefixes.  For example:
```
$ bin/helm -n infra-<TAB>
infra-diagnostics     infra-test-prometheus    infra-test-auth   infra-test-prometheus-query-exporter
$ bin/helm -n infra-test-<TAB>
infra-test-prometheus    infra-test-auth   infra-test-prometheus-query-exporter
$ bin/helm -n infra-test-prometheus<TAB>
infra-test-prometheus    infra-test-prometheus-query-exporter
$ bin/helm -n infra-test-prometheus-q<TAB>
$ bin/helm -n infra-test-prometheus-query-exporter

# Instead, with fuzzy matching:
$ bin/helm -n quer<TAB>
$ bin/helm -n infra-test-prometheus-query-exporter
```
Such fuzzy matching can make completion even more efficient in cases where identical prefixes are common.

The PR allows fuzzy matching to work for the different helm completion situations, except for chart names which is a more complicated case and which I will leave for later.

The way to enable such fuzzy matching is simply to return all completion choices no matter what the user has already typed, or put another way, not to filter on `toComplete`.  After that, it is the shell that deals with proper filtering.

The completion scripts provided by Cobra are prepared for such an approach.  For `bash` and `powershell` which don't support fuzzy matching, the completion script itself does the filtering on prefix.  There is therefore no need for helm to pre-filter as we are currently doing.

### DIFF
--- a/cmd/helm/docs.go
+++ b/cmd/helm/docs.go
@@ -69,14 +69,7 @@ func newDocsCmd(out io.Writer) *cobra.Command {
 	f.BoolVar(&o.generateHeaders, "generate-headers", false, "generate standard headers for markdown files")
 
 	cmd.RegisterFlagCompletionFunc("type", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
-		types := []string{"bash", "man", "markdown"}
-		var comps []string
-		for _, t := range types {
-			if strings.HasPrefix(t, toComplete) {
-				comps = append(comps, t)
-			}
-		}
-		return comps, cobra.ShellCompDirectiveNoFileComp
+		return []string{"bash", "man", "markdown"}, cobra.ShellCompDirectiveNoFileComp
 	})
 
 	return cmd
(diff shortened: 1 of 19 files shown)


marckhouzam:
I added some new tests to the PR, to make sure we don't regress this feature in the future.

mattfarina (approved):
This looks great. As a zsh user, this is going to make my life easier.

marckhouzam:
> This looks great. As a zsh user, this is going to make my life easier.

Thanks for the quick review!

PR #10514, opened 2022-01-01
fea(#10511): add --post-renderer-args to support assign args to PostR…

### LINKED ISSUE #10511: can  `--post-renderer`  support  assign  the cmd args into  `executable` ?
now helm  `--post-renderer` only accept  the `executable` in `PATH`;  

but it is very  necessary that  the `post-renderer-executable`   worked with some args ,for example change the docker.io registry into private docker registry when deploy the chart without internet access.

can helm can add `args` to assign the args in to  `post-renderer-executable` such as ` --post-renderer-args` 
```
helm install mychart stable/wordpress --post-renderer ./path/to/render_executable --post-renderer-args "--args_to_render_executable  "
````
if this proposal be ok, i d like  to contribute a PR.

### PR DESCRIPTION
**Pull request by tanguofu**
…enderer exec

closes #10511 

Now helm  `--post-renderer` only accept  the `executable` in `PATH`;  

but it is very  necessary that  the `post-renderer-executable`   worked with some args ,for example change the docker.io registry into private docker registry when deploy the chart without internet access.

can helm can add `args` to assign the args in to  `post-renderer-executable` such as ` --post-renderer-args` 
```
helm install mychart stable/wordpress --post-renderer ./path/to/render_executable --post-renderer-args "--args_1_to_render_executable" --post-renderer-args "--args_2_to_render_executable"
````

this implement to make the less change of the helm code.  it is  more compatibility than  #9047.

**If applicable**:

### DIFF
--- a/cmd/helm/flags.go
+++ b/cmd/helm/flags.go
@@ -117,7 +117,7 @@ func (o *outputValue) Set(s string) error {
 func bindPostRenderFlag(cmd *cobra.Command, varRef *postrender.PostRenderer) {
 	p := &postRendererOptions{varRef, "", []string{}}
 	cmd.Flags().Var(&postRendererString{p}, postRenderFlag, "the path to an executable to be used for post rendering. If it exists in $PATH, the binary will be used, otherwise it will try to look for the executable at the given path")
-	cmd.Flags().Var(&postRendererArgsSlice{p}, postRenderArgsFlag, "the args to an executable to be used for post rendering.  (can specify multiple)")
+	cmd.Flags().Var(&postRendererArgsSlice{p}, postRenderArgsFlag, "an argument to the post-renderer (can specify multiple)")
 }
 
 type postRendererOptions struct {
@@ -164,10 +164,15 @@ func (p *postRendererArgsSlice) Type() string {
 }
 
 func (p *postRendererArgsSlice) Set(val string) error {
-	if val == "" || p.options.binaryPath == "" {
+	if val == "" {
 		return nil
 	}
+
 	p.options.args = append(p.options.args, val)
+
+	if p.options.binaryPath == "" {
+		return nil
+	}
 	// overwrite if already create PostRenderer by `post-renderer` flags
 	pr, err := postrender.NewExec(p.options.binaryPath, p.options.args...)
 	if err != nil {
--- a/cmd/helm/flags.go
+++ b/cmd/helm/flags.go
@@ -164,10 +164,8 @@ func (p *postRendererArgsSlice) Type() string {
 }
 
 func (p *postRendererArgsSlice) Set(val string) error {
-	if val == "" {
-		return nil
-	}
 
+	// a post-renderer defined by a user may accept empty arguments
 	p.options.args = append(p.options.args, val)
 
 	if p.options.binaryPath == "" {
--- a/cmd/helm/flags.go
+++ b/cmd/helm/flags.go
@@ -36,8 +36,11 @@ import (
 	"helm.sh/helm/v3/pkg/repo"
 )
 
-const outputFlag = "output"
-const postRenderFlag = "post-renderer"
+const (
+	outputFlag         = "output"
+	postRenderFlag     = "post-renderer"
+	postRenderArgsFlag = "post-renderer-args"
+)
 
 func addValueOptionsFlags(f *pflag.FlagSet, v *values.Options) {
 	f.StringSliceVarP(&v.ValueFiles, "values", "f", []string{}, "specify values in a YAML file or a URL (can specify multiple)")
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: pkg/postrender/exec.go [hunk 1]
```diff
@@ -27,23 +27,24 @@ import (
...
 	binaryPath string
+	args       []string
 }
 
 // NewExec returns a PostRenderer implementation that calls the provided binary.
 // It returns an error if the binary cannot be found. If the path does not
 // contain any separators, it will search in $PATH, otherwise it will resolve
 // any relative paths to a fully qualified path
-func NewExec(binaryPath string) (PostRenderer, error) {
+func NewExec(binaryPath string, args []string) (PostRenderer, error) {
```
marckhouzam, line 37: This is a public API so we cannot change the signature of the function.
You can either create a new function that will satisfy your requirement, or you will need to use the current one without changing its signature.

We need to do that to avoid breaking other programs using helm's API.

marckhouzam (changes requested):
Thanks for the PR.  I've haven't looked at it all but let's first address the requirement of backwards-compatibility.  Please see the comment within the review.

pkg/postrender/exec.go
tanguofu: i got. i had add a new function `NewExecWithArgs` and keep back the `func NewExec(binaryPath string) (PostRenderer, error) `
thanks.

cmd/helm/flags.go
```diff
@@ -112,30 +115,70 @@ func (o *outputValue) Set(s string) error {
...
+
+type postRendererArgsFlag struct {
+	options *postRendererOptions
+}
+
+func (p postRendererArgsFlag) String() string {
+	return ""
+}
+
+func (p postRendererArgsFlag) Type() string {
```
marckhouzam, line 162: Since this flag can be repeated, it would be nice to let the shell completion logic know that it can suggest the flag more than once.  To achieve this you must include the string "Array" or "Slice" in the `Type()`.  You can see where the Cobra underlying library checks this: https://github.com/spf13/cobra/blob/cb9d7b1cec87c2bb005c6e2790553bcd629bc542/completions.go#L329

So can you `return "postrenderer-argsSlice"` here?

If you want to test the difference you can build your new helm and do the following:
```
$ helm __complete install --post-renderer echo --post-renderer-args arg1 --p
```
This command returns the possible completions for the specified input.  You will want to see that the `--post-renderer-args` keeps being suggested as a completion (but not the `--post-renderer` flag)

pkg/postrender/exec.go
```diff
@@ -38,12 +39,20 @@ func NewExec(binaryPath string) (PostRenderer, error) {
 	if err != nil {
 		return nil, err
 	}
-	return &execRender{fullPath}, nil
+	return &execRender{fullPath, []string{}}, nil
+}
+
+func NewExecWithArgs(binaryPath string, args []string) (PostRenderer, error) {
```
marckhouzam, line 45: I realized there is a better way to be backwards-compatible.  Instead of adding a new function `NewExecWithArgs()`, you can instead change the existing one, from `func NewExec(binaryPath string) (PostRenderer, error)` to `func NewExec(binaryPath string, args ...string) (PostRenderer, error)`.

This approach allows to call `NewExec(path, args...)` to specify arguments, but also allows calling `NewExec(path)` like before.

Sorry about not having thought about it before.

marckhouzam (changes requested):
Nice job with this.  A couple of comments and we should be good.

pkg/postrender/exec.go
tanguofu: i got it,  this is best way! i did not found yet, very thanks for u suggestion, i will fix those later.

tanguofu:
@marckhouzam  i am  sorry for update take a few days,  may you review again? thanks  for your instruction.

pkg/postrender/exec_test.go
```diff
@@ -124,6 +128,40 @@ func TestExecRun(t *testing.T) {
...
+func TestNewExecWithOneArgsRun(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		// the actual Run test uses a basic sed example, so skip this test on windows
+		t.Skip("skipping on windows")
+	}
+	is := assert.New(t)
+	testpath, cleanup := setupTestingScript(t)
+	defer cleanup()
+
+	renderer, err := NewExec(testpath, "FOOTEST")
```
marckhouzam, line 140: Can you call the argument `arg` instead?  I find reusing `FOOTEST` to be confusing.

pkg/postrender/exec_test.go
```diff
@@ -124,6 +128,40 @@ func TestExecRun(t *testing.T) {
...
+func TestNewExecWithTwoArgsRun(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		// the actual Run test uses a basic sed example, so skip this test on windows
+		t.Skip("skipping on windows")
+	}
+	is := assert.New(t)
+	testpath, cleanup := setupTestingScript(t)
+	defer cleanup()
+
+	renderer, err := NewExec(testpath, "FOOTEST", "FOOTEST")
```
marckhouzam, line 157: Can you call the arguments `arg1` and `arg2` instead?  I find reusing `FOOTEST` to be confusing.

cmd/helm/flags.go
```diff
@@ -112,33 +115,82 @@ func (o *outputValue) Set(s string) error {
 }
 
 func bindPostRenderFlag(cmd *cobra.Command, varRef *postrender.PostRenderer) {
-	cmd.Flags().Var(&postRenderer{varRef}, postRenderFlag, "the path to an executable to be used for post rendering. If it exists in $PATH, the binary will be used, otherwise it will try to look for the executable at the given path")
+	p := &postRendererOptions{varRef, "", []string{}}
+	cmd.Flags().Var(&postRendererString{p}, postRenderFlag, "the path to an executable to be used for post rendering. If it exists in $PATH, the binary will be used, otherwise it will try to look for the executable at the given path")
+	cmd.Flags().Var(&postRendererArgsSlice{p}, postRenderArgsFlag, "the args to an executable to be used for post rendering.  (can specify multiple)")
```
marckhouzam, line 120: Can you change this string to be:
"an argument to the post-renderer  (can specify multiple)"

cmd/helm/flags.go
```diff
@@ -112,33 +115,82 @@ func (o *outputValue) Set(s string) error {
...
+func (p *postRendererArgsSlice) String() string {
+	return "[" + strings.Join(p.options.args, ",") + "]"
+}
+
+func (p *postRendererArgsSlice) Type() string {
+	return "postRendererArgsSlice"
+}
+
+func (p *postRendererArgsSlice) Set(val string) error {
+	if val == "" || p.options.binaryPath == "" {
```
marckhouzam, line 167: This is a bit of a corner case, but you had it better in your previous version.
Even when `p.options.binaryPath == ""` we should still call `p.options.args = append(p.options.args, val)`.  This allows to use the `--post-renderer-args` before the `--post-renderer` flag on the command-line.
For example, if I run:
```
$ bin/helm template test bitnami/nginx --post-renderer-args arg1  --post-renderer echo --post-renderer-args arg2
arg2
```
Notice that arg1 is ignored because it was specified before `--post-renderer`

marckhouzam (changes requested):
Nicely done.  One corner case to fix and I think we are done.

cmd/helm/flags.go
tanguofu: got! .  i will fix all of these.

cmd/helm/flags.go
```diff
@@ -112,33 +115,87 @@ func (o *outputValue) Set(s string) error {
...
+
+func (p *postRendererArgsSlice) Type() string {
+	return "postRendererArgsSlice"
+}
+
+func (p *postRendererArgsSlice) Set(val string) error {
+	if val == "" {
+		return nil
+	}
+
```
marckhouzam, line 170: I think that for the arguments to the post-renderer, an empty argument could be valid.
I don't have an example, but a post-renderer defined by a user may accept empty arguments specified like this:
```
$ bin/helm template test bitnami/nginx --post-renderer myrenderer --post-renderer-args ''

```
Notice that is this different than forgetting to add an argument, in which case an error will be shown:
```
$ bin/helm template test bitnami/nginx --post-renderer echo --post-renderer-args
Error: flag needs an argument: --post-renderer-args
```
So, what do you think of removing the check for `val == ""` here?

tanguofu: i got !
very thanks your  detailed and comprehensive advice, that help me  learn much more code skills.
many thanks!  and `helm` is awesome golang project !

marckhouzam (approved):
LGTM Thanks @tanguofu !

This will need approval from another maintainer.
/cc @hickeyma @bacongobbler

hickeyma (approved):
LGTM, thanks @tanguofu. 

Tested manually using https://github.com/thomastaylor312/advanced-helm-demos/tree/master/post-render where first tested without passing the args (as per the script) and then using the args instead of the script.

PR #10515, opened 2022-01-01
Add support for importing single values from dependencies

### LINKED ISSUE #2746: import-values expects a table and not an individual value
Or alternately, you can't import all the values of a child chart.

Example (altered from docs)
```
# parent's requirements.yaml file
dependencies:
  - name: subchart1
    version: 0.1.0
    ...
    import-values:
      - child: myint
        parent: myimports.myint
     - child: ''
        parent: childvals

# parent's values.yaml file

myimports:
  myint: 0
  mybool: false
  mystring: "helm rocks!"
  
# subchart1's values.yaml file

myint: 999
mybool: true

# parent's expected final values

myimports:
  myint: 999
  mybool: false
  mystring: "helm rocks!"
childvals:
  myint: 999
  mybool: true
```
What actually happens is you get errors.

mateuszjenek:
Add support for importing single values ​​from dependencies. This keeps the value file clean, especially when using subcharts with different clean code practices. It closes #2746

**If applicable**:

### DIFF
--- a/pkg/chartutil/dependencies.go
+++ b/pkg/chartutil/dependencies.go
@@ -198,19 +198,23 @@ Loop:
 }
 
 // pathToMap creates a nested map given a YAML path in dot notation.
-func pathToMap(path string, data map[string]interface{}) map[string]interface{} {
+func pathToMap(path string, data interface{}) map[string]interface{} {
 	if path == "." {
-		return data
+		if istable(data) {
+			return data.(map[string]interface{})
+		}
+		return nil
 	}
 	return set(parsePath(path), data)
 }
 
-func set(path []string, data map[string]interface{}) map[string]interface{} {
+func set(path []string, data interface{}) map[string]interface{} {
 	if len(path) == 0 {
 		return nil
 	}
-	cur := data
-	for i := len(path) - 1; i >= 0; i-- {
+	cur := make(map[string]interface{})
+	cur[path[len(path)-1]] = data
+	for i := len(path) - 2; i >= 0; i-- {
 		cur = map[string]interface{}{path[i]: cur}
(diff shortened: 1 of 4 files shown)


websi96:
Hi I am currently also struggeling with importing values from subcharts, but i tried to import arrays and not just single values.

Here is what I did to share values between subcharts:

Assume this mongodb dependencies in my Parent Chart.yaml
```yaml
#Chart.yaml in parent chart

dependencies:
  - name: mongodb
    version: "13.4.1"
    repository: "https://charts.bitnami.com/bitnami"
    condition: mongodb.enabled
    import-values:
    - child: auth
      parent: global.mongodb.auth
  ```
 
  Here I pass the auth values from the mongodb subchart to the global values of my parent chart.
  Then I can set the auth settings in values.yaml of the parent chart like so:
  
 ```yaml
#values.yaml in parent chart
 
mongodb:
  enabled: true
  auth:
    usernames:
    - user_1
    - user_2
    passwords:
    - user_1_pw
    - user_2_pw
    databases:
    - user_1_db
    - user_2_db
  ```
  
  In any other Sub Chart I can then access the users like os:
  
  ```yaml
 #any template.yaml in subchart
 
{{- with .Values.global.mongodb.auth -}}

  {{- $pws := .passwords -}}
  {{- $dbs := .databases -}}
  {{- $users := .usernames }}

{{- end -}}
  ```
  
My only problem is, that this only works with objects like auth and not the users array directly.
If I would reference an array I get this:
  ```
dependencies.go:247: Warning: ImportValues missing table from chart mongodb: "usernames" is not a table
  ```
  
How would we also allow arrays here?

PR #10091, opened 2021-09-02
feat(helm): add --allow-empty flag to 'helm repo list'

presztak:
**What this PR does / why we need it**:
`--allow-empty` flag signal that it is ok (return exit code 0) to not have any repositories to show.

Closes #10028

Signed-off-by: Piotr Resztak <piotr.resztak@gmail.com>

**Special notes for your reviewer**:

**If applicable**:
- [ ] this PR contains documentation
- [x] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

PR #10519, opened 2022-01-04
fix(helm): ignore file-not-found error for `helm repo list -o json`

nightkr:

Currently, `helm repo list -o json` returns (either `[]` or an english "no repositories" error message, depending on whether `$HELM_CONFIG_HOME/repositories.yaml` exists). This PR aligns the two cases so that it always returns `[]`.

This makes Helm more predictable to interact with for external scripts.

This is extracted from https://github.com/helm/helm/pull/10091#discussion_r778070678.

**If applicable**:

### DIFF
--- a/cmd/helm/repo_list.go
+++ b/cmd/helm/repo_list.go
@@ -38,8 +38,8 @@ func newRepoListCmd(out io.Writer) *cobra.Command {
 		Args:              require.NoArgs,
 		ValidArgsFunction: noCompletions,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			f, err := repo.LoadFile(settings.RepositoryConfig)
-			if isNotExist(err) || (len(f.Repositories) == 0 && !(outfmt == output.JSON || outfmt == output.YAML)) {
+			f, _ := repo.LoadFile(settings.RepositoryConfig)
+			if len(f.Repositories) == 0 && !(outfmt == output.JSON || outfmt == output.YAML) {
 				return errors.New("no repositories to show")
 			}
 


teozkr:
```bash
# foo doesn't exist
$ export HELM_CONFIG_HOME=foo
$ bin/helm repo ls -o json
[]
$ echo $?
0
# helm 3.7.2
$ helm repo ls -o json
Error: no repositories to show
$ echo $?
1

# this was already the behaviour if $HELM_CONFIG_HOME had previously contained any repositories...
$ helm repo add stable https://charts.helm.sh/stable
"stable" has been added to your repositories
$ helm repo rm stable
"stable" has been removed from your repositories
$ helm repo ls -o json
[]
```

FrenchBen:
In this case, it seems valuable to know if a repo file exists - Adding a repo, and removing it, creates the file and keeps it. 
Your PR seems to remove that distinction.

For example, if I was to supply a `repositories.yaml` file, that was empty or invalid, I would want to know that subtle difference:
```
$ touch repositories.yaml
$ helm repo list --repository-config repositories.yaml
Error: no repositories to show
```
Your PR would remove this and simply tell me you don't have any repo!

Which means that I would then have to catch the error further, if I was to try to add a new repo, to a file that couldn't be parsed in the first place:
```
$ echo "garbage" > repositories.yaml
$ helm repo add stable https://charts.helm.sh/stable --repository-config repositories.yaml
Error: error unmarshaling JSON: while decoding JSON: json: cannot unmarshal string into Go value of type repo.File
```

teozkr:
> In this case, it seems valuable to know if a repo file exists - Adding a repo, and removing it, creates the file and keeps it.
Your PR seems to remove that distinction.

Because they are functionally equivalent states everywhere else in Helm.

> For example, if I was to supply a repositories.yaml file, that was empty or invalid, I would want to know that subtle difference:
> 
> ```bash
> $ touch repositories.yaml
> $ helm repo list --repository-config repositories.yaml
> Error: no repositories to show
> ```
> 
> Your PR would remove this and simply tell me you don't have any repo!

Agreed, but this is already the case with the released Helm v3.7.2:

```bash
$ helm version
version.BuildInfo{Version:"v3.7.2", GitCommit:"663a896f4a815053445eec4153677ddc24a0a361", GitTreeState:"", GoVersion:"go1.16.10"}
$ echo garbage > repositories.yaml
$ helm repo list --repository-config repositories.yaml -o yaml
[]
# Even a completely unreadable file just causes it to return []!
$ chmod a= repositories.yaml 
$ helm repo list --repository-config repositories.yaml -o yaml
[]
```

The only error that is not simply thrown away and ignored is "file not found", and even _that_ just returns the message "Error: no repositories to show", which is entirely useless for figuring out what the actual problem is.

bacongobbler:
Can you please test https://github.com/helm/helm/pull/10745 and see if that fixes your issue?

bacongobbler (approved):
LGTM.

lfrancke:
Thank you @bacongobbler!

Issue #10393, opened 2021-11-23
Moving OCI integration from experimental to full feature


scottrigby:
Most of the work outlined in [hip-0006](https://github.com/helm/community/blob/main/hips/hip-0006.md) has been finished.

The upstream [ORAS Go](https://github.com/oras-project/oras-go) project that Helm imports now has a stable `1.0.0` release, with a backward compatibility guarantee. This was the last major step in unblocking releasing Helm OCI integration as a full feature.

## Remaining steps

- [x] Documentation only: https://github.com/helm/helm/issues/10169#issuecomment-976973309
    - #1249
- [x] #10534 
    - #10535
- [x] #10166
    - https://github.com/helm/helm/pull/10521
- [x] #10156
    - https://github.com/oras-project/oras-go/pull/92
    - #10536 
- [x] #9694
    - https://github.com/oras-project/oras-go/pull/89
    - https://github.com/helm/helm/pull/10527
- [x] #9188
    - #10546 to be merged last, after all other work above is complete

## Not identified as blockers

However, if someone has time to test these oci-related PRs, they may still get into 3.8.0.

- [ ] #6324
    - PR ready for review 👀: #10408 
- [ ] #10175 
    - PR ready for review 👀  #10430

## Follow-up

This issue is only concerned with comprehensively listing any blockers for moving OCI support to a full feature. New features can be added after it is moved from experimental. After that point, only backward-compatible changes will be accepted until Helm 4.

- Join the discussion about backward-compatible features here: https://github.com/helm/helm/issues/10312

_Update (by mattfarina) 12/3: Connecting with technosophos and jdolitsky we reviewed the open issues from bacongobbler and added two additional tasks to the remaining steps._

bacongobbler:
> Anything else?

Here's my laundry list... I started by searching the issue queue. Here's the search terms I used to look for open tickets:

https://github.com/helm/helm/issues?q=is%3Aissue+is%3Aopen+oci
https://github.com/helm/helm/issues?page=1&q=is%3Aissue+is%3Aopen+docker
https://github.com/helm/helm/issues?q=is%3Aissue+is%3Aopen+registry

This was just a casual search... It might be worth triaging the entire queue just to make sure no stone is left unturned.

Anyways, I saw a few tickets that might be worth evaluating (some have been discussed at length during previous developer calls):

- https://github.com/helm/helm/issues/10013
- https://github.com/helm/helm/issues/9694
- https://github.com/helm/helm/issues/6324
[...]

PR #10521, opened 2022-01-05
For OCI, convert + to _ before pull, and + to _ before push

### LINKED ISSUE #10166: Adopt Helm best practice for altering build affix in versions when pushing to OCI
Output of `helm version`: 3.7.0

According to https://helm.sh/docs/chart_best_practices/conventions/#version-numbers, chart versions should follow `SemVer 2`. OCI labels choke on `+`. 

> When SemVer versions are stored in Kubernetes labels, we conventionally alter the + character to an _ character, as labels do not allow the + sign as a value.

It would be great to see that convention adopted when publishing Helm charts to an OCI registry.

Currently, when I include a build affix in the version, I can package the chart but not push it.

```
apiVersion: v2
name: foo
description: The Foo Chart

type: application
version: 0.1.0+3d53eea
appVersion: "1.0.0"
```

```
[foo] helm package .
Successfully packaged chart and saved it to: /some/path/to/foo/foo-0.1.0+3d53eea.tgz
[foo] helm push foo-0.1.0+3d53eea.tgz oci://12345.dkr.ecr.us-west-1.amazonaws.com/some-namespace
Error: unexpected status: 400 Bad Request
```

But if I use an `_` instead of an `+`:

```
helm package .
Error: validation: chart.metadata.version "0.1.0_3d53eea" is invalid
```

Ideally, I would use the `+` separator for the build affix (per SemVer 2), and Helm would apply the documented convention and push an image label with `_` substituted for `+`.

scottrigby:

- needed for #10393
- closes #10166 

- Will additionally need to change underscore  to plus in `(*Resolver).Resolve`, but that needs to happen after code for OCI tag listing API is added

**If applicable**:


jdolitsky (approved):
lgtm. does it work?

mattfarina (changes requested):
I tested it and this change does not work. Here is what I found. In https://github.com/helm/helm/blob/39ca699ca790e02ba36753dec6ba4177cc68d417/internal/experimental/registry/client.go#L424-L429

The `HasSuffix` has a ref with the `_` while the `meta.Version` has the +. So this doesn't work. The `meta.Version` should also use the `_` instead of the `+`.

Interesting side note (that I've not tracked down yet) ... when running `helm dep build` there's an error trying to treat the registry like a repository...
```
❯ helm dep build .
object required
...Unable to get an update from the "oci://ghcr.io/mattfarina" chart repository:
	object required
```

In the output of running `helm dep` commands I noticed it has the _. Like so...
```
Pulled: ghcr.io/mattfarina/i10521:0.1.0_build.1
```
This is from...
https://github.com/helm/helm/blob/39ca699ca790e02ba36753dec6ba4177cc68d417/internal/experimental/registry/client.go#L357-L358

Seeing this back and forth between `_` and `+` is going to confuse people. I wonder if this should detect that a `_` is present and have a note next to the url such as `_ substituted for + in version`. I also wonder if the push and pull commands should detect the `+` and leave a message in the output to tell them about the switch so it's not a surprise. Thoughts? Just don't want to catch people off guard.

scottrigby:
@mattfarina Thanks!

Good points about `HasSuffix` and various output with an underscore instead of the `+` users would be expecting. We should probably move this string manipulation inside `(*Client).Push` and `(*Client).Pull`, keep an original and transformed vars there for whenever we need them, so we don't have to convert back and forth.

For user transparency, maybe we just output one message whenever we do a conversion during Push/Pull.

> Interesting side note (that I've not tracked down yet) ... when running helm dep build there's an error trying to treat the registry like a repository...

I saw this too. It doesn't seem to hurt anything AFAICT, but we should find and DESTROY this lol. Prob a separate (minor?) PR.

scottrigby:
@mattfarina Ok, changes made.

One note: this means SDK users will be responsible to adopt this same convention before push and pull. Alternatively I could move the string change back to ocigetter and ocipusher, and just double convert back to original for comparison within the client. I can see benefits of either way here - do you have a strong preference?

mattfarina (commented):
This looks much better and I only had nits on the language in the message. I ran a number of manual test that all work, too.

internal/experimental/registry/client.go
```diff
@@ -37,6 +37,13 @@ import (
 	"helm.sh/helm/v3/pkg/helmpath"
 )
 
+// See https://github.com/helm/helm/issues/10166
+const registryUnderscoreMessage = `
+OCI artifact references do not support the plus sign(+). To solve this Helm
+adopts the convention of changing plus (+) to underscore (_) in chart version
+references during registry push - and OCI artifact reference from underscore
+back to plus on registry pull.`
```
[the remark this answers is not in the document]
scottrigby: Hey @mattfarina Yes I think noting semver in the message makes sense
- Updated that and added your co-authorship in the commit trailer
- Added newlines to message output instead of the message const itself

internal/experimental/registry/client.go
```diff
@@ -207,6 +214,11 @@ type (
 
 // Pull downloads a chart from a registry
 func (c *Client) Pull(ref string, options ...PullOption) (*PullResult, error) {
+	// Convert plus (+) to underscore (_) in request string before pull
+	// This must remain right at the top of client.Pull()
+	// See https://github.com/helm/helm/issues/10166
+	ref = strings.ReplaceAll(ref, "+", "_")
```
mattfarina, line 220: I'm sorry I didn't catch this earlier. The `+` should only be changed in the tag and not the whole `ref`. A `ref` is a URI and other parts of the URI (e.g., the path) can contain a `+` character. It's not as easy as replacing all the + with _.

mattfarina (changes requested):

internal/experimental/registry/client.go
mattfarina: Note, this applies to all places there the swap happens.

scottrigby: Ah yes, good catch. Myself and several other people are divvying up tasks for wrapping this up, so either I or someone else will push a fix to this ASAP ✨

joshrwolf: leaving an update on the approach (pushed in fb0eeeb) here after talking with @scottrigby:

to scope to a tag we first need to identify a tag, and to identify a tag we _should_ first validate the reference.  so fb0eeeb first identifies a tag and does the `'+' -> '_'` replacement only on the tag if it exists, then returns a validated reference (using oras' `registry.ParseReference`) or error.  the remainder of `Push`/`Pull` then uses the validated `oras.Reference` struct for all ops.

while this means adding the updated oras dependency to this PR, it will eventually be consolidated to the updated oras release alongside #10527

internal/experimental/registry/util.go
```diff
@@ -54,3 +55,25 @@ func ctx(out io.Writer, debug bool) context.Context {
...
 }
+
+// parseReference will parse and validate the reference, and clean tags when applicable
+// 	tags are only cleaned when plus (+) signs are present, and are converted to underscores (_) before pushing
+// 	See https://github.com/helm/helm/issues/10166
+func parseReference(raw string) (registry.Reference, error) {
+	// The sole possible reference modification is replacing plus (+) signs present in tags with underscores (_)
+	// 	To do this properly, we first need to identify a tag, and then pass it on to the reference parser
+	// 	NOTE: Passing immediately to the reference parser will fail since (+) signs are an invalid tag character, and
+	// 		  simply replacing all plus (+) occurances could invalidate other portions of the URI
```
scottrigby, line 66:
```suggestion
	// 		  simply replacing all plus (+) occurences could invalidate other portions of the URI
```

internal/experimental/registry/client.go
```diff
@@ -495,6 +518,11 @@ func (c *Client) Push(data []byte, ref string, options ...PushOption) (*PushResu
 	}
 	fmt.Fprintf(c.out, "Pushed: %s\n", result.Ref)
 	fmt.Fprintf(c.out, "Digest: %s\n", result.Manifest.Digest)
+	if strings.Contains(result.Ref, "_") {
```
mattfarina, line 521: The `result.Ref` is the full URI to the chart in a registry. The `_` character is a valid path character. This will detect cases outside of the tag. Shouldn't it be scoped to the tag?

internal/experimental/registry/client.go
```diff
@@ -206,7 +213,12 @@ type (
 )
 
 // Pull downloads a chart from a registry
-func (c *Client) Pull(ref string, options ...PullOption) (*PullResult, error) {
+func (c *Client) Pull(rawRef string, options ...PullOption) (*PullResult, error) {
```
mattfarina, line 216: In both `Pull` and `Push` the variable in the function signature is being changed to `rawRef` and `ref` is the parsed version. Would it be possible to keep the function sig the same and instead call the internal one `parsedRef`? Since the type is changing the use of this internally had to change.

internal/experimental/registry/util.go
```diff
@@ -54,3 +55,28 @@ func ctx(out io.Writer, debug bool) context.Context {
 	orascontext.GetLogger(ctx).Logger.SetLevel(logrus.DebugLevel)
 	return ctx
 }
+
+// parseReference will parse and validate the reference, and clean tags when
+// applicable tags are only cleaned when plus (+) signs are present, and are
+// converted to underscores (_) before pushing
+// See https://github.com/helm/helm/issues/10166
+func parseReference(raw string) (registry.Reference, error) {
```
mattfarina, line 63: This function makes me nervous without tests.

internal/experimental/registry/util.go
```diff
@@ -54,3 +55,28 @@ func ctx(out io.Writer, debug bool) context.Context {
...
+	// signs are an invalid tag character, and simply replacing all plus (+)
+	// occurrences could invalidate other portions of the URI
+	parts := strings.Split(raw, ":")
+	if len(parts) > 1 && !strings.Contains(parts[len(parts)-1], "/") {
+		tag := parts[len(parts)-1]
+
+		if tag != "" {
+			// Replace any plus (+) signs with known underscore (_) conversion
+			newTag := strings.ReplaceAll(tag, "+", "_")
+			raw = strings.ReplaceAll(raw, tag, newTag)
```
mattfarina, line 77: If the path happens to contain the same string as the old tag it would be replaced and broken. I will admit this is an incredibly rare situation (if it were to ever happen). I'm not even sure a registry would handle a `+` in their namespace. Just making the observation.

mattfarina (commented):
I have one issue, one nit, one nice to have, and one comment about what's going on....

internal/experimental/registry/util.go
joshrwolf: you're right, but if we're assuming this only works for registries adhering with the oci distribution spec, then we should be covered by the limitations on registry namespaces ([here](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pulling-manifests)).

buuut that also doesn't excuse the fact that there isn't a check for the scenario you outlined

internal/experimental/registry/client.go
joshrwolf: absolutely! I'm not married to the initial proposed naming, it's been updated in fed0d31

internal/experimental/registry/client.go
joshrwolf: good catch, updated in fed0d31

internal/experimental/registry/util.go
mattfarina: Thanks for filling in details on the namespace spec.

mattfarina (approved):

scottrigby:
Hmm, the circleCI job did not appear to run for the last commit. I found the payload in gh settings webhook recent deliveries (by pr/sha), and re-sent the payload, but I don't see a corresponding delivery in circle 🤔 I'll see if an empty commit kicks it off.

PR #10527, opened 2022-01-06
OCI version range support

### LINKED ISSUE #9694: OCI: no version range support for dependencies/`helm install`

scottrigby:

closes #9694 

- Thank you to @mattfarina and @jdolitsky for co-navigating this initially
- Collaborating on this PR branch of my fork with @sabre1041

**If applicable**:
- ~[ ] this PR contains documentation~

### DIFF
--- a/internal/experimental/registry/client.go
+++ b/internal/experimental/registry/client.go
@@ -604,7 +604,9 @@ func (c *Client) Tags(ref string) ([]string, error) {
 
 	var tagVersions []*semver.Version
 	for _, tag := range registryTags {
-		tagVersion, err := semver.StrictNewVersion(tag)
+		// Change underscore (_) back to plus (+) for Helm
+		// See https://github.com/helm/helm/issues/10166
+		tagVersion, err := semver.StrictNewVersion(strings.ReplaceAll(tag, "_", "+"))
 		if err == nil {
 			tagVersions = append(tagVersions, tagVersion)
 		}
--- a/internal/experimental/registry/client.go
+++ b/internal/experimental/registry/client.go
@@ -31,6 +31,9 @@ import (
 	dockerauth "oras.land/oras-go/pkg/auth/docker"
 	"oras.land/oras-go/pkg/content"
 	"oras.land/oras-go/pkg/oras"
+	"oras.land/oras-go/pkg/registry"
+	registrremote "oras.land/oras-go/pkg/registry/remote"
+	registryauth "oras.land/oras-go/pkg/registry/remote/auth"
 
 	"helm.sh/helm/v3/internal/version"
 	"helm.sh/helm/v3/pkg/chart"
@@ -49,10 +52,11 @@ type (
 	Client struct {
 		debug bool
 		// path to repository config file e.g. ~/.docker/config.json
-		credentialsFile string
-		out             io.Writer
-		authorizer      auth.Client
-		resolver        remotes.Resolver
+		credentialsFile    string
+		out                io.Writer
+		authorizer         auth.Client
+		registryAuthorizer *registryauth.Client
+		resolver           remotes.Resolver
 	}
 
 	// ClientOption allows specifying various settings configurable by the user for overriding the defaults
@@ -87,6 +91,15 @@ func NewClient(options ...ClientOption) (*Client, error) {
 			return nil, err
 		}
 		client.resolver = resolver
+	}
+	if client.registryAuthorizer == nil {
+		client.registryAuthorizer = &registryauth.Client{
+			Header: http.Header{
+				"User-Agent": {version.GetUserAgent()},
+			},
+			Cache: registryauth.DefaultCache,
+		}
+
 	}
 	return client, nil
 }
@@ -538,4 +551,19 @@ func PushOptStrictMode(strictMode bool) PushOption {
 	return func(operation *pushOperation) {
 		operation.strictMode = strictMode
 	}
+}
+
+// Tags lists all tags for a given repository
+func (c *Client) Tags(ref string) ([]string, error) {
+	parsedReference, err := registry.ParseReference(ref)
+	if err != nil {
+		return nil, err
+	}
+
+	repository := registrremote.Repository{
+		Reference: parsedReference,
+		Client:    c.registryAuthorizer,
+	}
+
+	return registry.Tags(ctx(c.out, c.debug), &repository)
 }
--- a/internal/experimental/registry/client.go
+++ b/internal/experimental/registry/client.go
@@ -567,20 +567,21 @@ func (c *Client) Tags(ref string) ([]string, error) {
 		Client:    c.registryAuthorizer,
 	}
 
-	registrtyTags, err := registry.Tags(ctx(c.out, c.debug), &repository)
+	registryTags, err := registry.Tags(ctx(c.out, c.debug), &repository)
 	if err != nil {
 		return nil, err
 	}
 
 	var tagVersions []*semver.Version
-	for _, tag := range registrtyTags {
+	for _, tag := range registryTags {
 		tagVersion, err := semver.StrictNewVersion(tag)
-		if err != nil {
+		if err == nil {
 			tagVersions = append(tagVersions, tagVersion)
 		}
 	}
 
-	sort.Sort(semver.Collection(tagVersions))
+	// Sort the collection
+	sort.Sort(sort.Reverse(semver.Collection(tagVersions)))
 
 	tags := make([]string, len(tagVersions))
 
--- a/internal/experimental/registry/client.go
+++ b/internal/experimental/registry/client.go
@@ -22,8 +22,10 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	"sort"
 	"strings"
 
+	"github.com/Masterminds/semver/v3"
 	"github.com/containerd/containerd/remotes"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
@@ -553,7 +555,7 @@ func PushOptStrictMode(strictMode bool) PushOption {
 	}
 }
 
-// Tags lists all tags for a given repository
+// Tags provides an all semver compliant tags for a given repository
 func (c *Client) Tags(ref string) ([]string, error) {
 	parsedReference, err := registry.ParseReference(ref)
 	if err != nil {
@@ -565,5 +567,27 @@ func (c *Client) Tags(ref string) ([]string, error) {
 		Client:    c.registryAuthorizer,
 	}
 
-	return registry.Tags(ctx(c.out, c.debug), &repository)
+	registrtyTags, err := registry.Tags(ctx(c.out, c.debug), &repository)
+	if err != nil {
+		return nil, err
+	}
+
+	var tagVersions []*semver.Version
+	for _, tag := range registrtyTags {
+		tagVersion, err := semver.StrictNewVersion(tag)
+		if err != nil {
+			tagVersions = append(tagVersions, tagVersion)
+		}
+	}
+
+	sort.Sort(semver.Collection(tagVersions))
+
+	tags := make([]string, len(tagVersions))
+
+	for iTv, tv := range tagVersions {
+		tags[iTv] = tv.String()
+	}
+
+	return tags, nil
+
 }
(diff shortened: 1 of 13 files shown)


pkg/downloader/manager.go
```diff
@@ -233,6 +233,26 @@ func (m *Manager) loadChartDir() (*chart.Chart, error) {
...
+			// TODO(scottrigby): fix HTTP error
+			// > Error: Get "https://localhost:5000/v2/helm-charts/tags/list": http: server gave HTTP response to HTTPS client
+			ref := strings.TrimPrefix(d.Repository, "oci://")
+			tags, err := m.RegistryClient.Tags(ref)
+			if err != nil {
+				return nil, err
+			}
+			req[i].OCITagVersions = tags
+		}
+	}
```
mattfarina, line 255: Why was this additional loop added here? This could be done in `internal/resolver/resolver.go` at the same place where `OCITagVersions` is used. Then the `Dependency` struct doesn't need to have any new property put on it. Localizing the work to the resolver would make the change simpler with fewer possible complications to consider.

internal/experimental/registry/client.go
```diff
@@ -511,3 +524,18 @@ func PushOptStrictMode(strictMode bool) PushOption {
 		operation.strictMode = strictMode
 	}
 }
+
+// Tags lists all tags for a given repository
```
mattfarina, line 528: It might be useful to note that the response of tags isn't sorted.

internal/resolver/resolver.go
```diff
@@ -139,6 +139,20 @@ func (r *Resolver) Resolve(reqs []*chart.Dependency, repoNames map[string]string
 				return nil, errors.Wrapf(FeatureGateOCI.Error(),
 					"repository %s is an OCI registry", d.Repository)
 			}
+
+			// This works, fake it to test mocking the version
+			// tags := []string{"0.1.0", "0.1.1", "0.2.0"}
+			tags := d.OCITagVersions
```
mattfarina, line 145: The tags here aren't sorted. I didn't test it but reading through the oras tag code it doesn't appear to be. This list is turned into `ChartVersion` which are still unsorted by version. But, the version constraint handling expects it to be sorted. So, these tags need to be sorted and non-semver tags pulled out as part of the process. This will be especially important because there is work to support sigstore which uses tags.

You can find a sorting example at https://github.com/Masterminds/semver/blob/master/collection_test.go (tests as an example)

mattfarina (changes requested):
Found some issues with the current setup. I added suggestions on how to simplify the structure and fix the issues.

pkg/downloader/manager.go
scottrigby: The problem is I don't see a backward compatible way to be able to get `RegistryClient` from within (*Resolver).Resolve, which is needed to call the API. However the calling function `(*Manager).resolve` has this already. So making the tag API call from there, then attaching the result to `Dependency` object was a solution for this that is backward compatible. Do you see another way?

internal/resolver/resolver.go
scottrigby: Thanks this is helpful. Will do soon.

pkg/downloader/manager.go
mattfarina: The resolver is an [`internal`](https://go.dev/doc/go1.4#internalpackages) package and not part of the public API. Take a look at its location. This is not a general purpose resolver meant to be imported and used by others. Its only access is through the `resolve` method here. It's meant just to support Helms specific use.

You already added a 3rd argument to `resolver.New` to pass in the registry client and you extended `Resolver` to hold on to it. Now you just need to use it within the `Resolve` method.

sabre1041: @mattfarina thats exactly the approach that I am working on today

scottrigby: Thanks @mattfarina you're right – and yes that makes it easier! @sabre1041 & I talked about this approach last night. Yep I see Andy added that now. Much better

scottrigby: In touch w Andy on this, so we will get it done tonight one way or another 💯

internal/resolver/resolver.go
scottrigby: Been collabing w Andy, who just pushed a fix for this - added directly to registry `(*Client).tags` 💯

internal/experimental/registry/client.go
scottrigby: No longer true since https://github.com/helm/helm/pull/10527/commits/348942dfc79b193a08c700d0c69eebc6f5eef3b7. Yay. See https://github.com/helm/helm/pull/10527#discussion_r780831755

scottrigby:
Merging this PR is on hold until the HTTP HTTPS auth issue is addressed. We are planning to connect with @deitch tomorrow to get direction on the best way to manage auth and http configuration on the registry client with those defined on an oras client. Slack convo [here](https://cloud-native.slack.com/archives/CJ1KHJM5Z/p1641842277021100).

Apart from that though, this pr is working fine with HTTPS and no auth. Both for chart OCI dependencies using semver range, and helm install semver logic. See the manual test output below testing with an example HTTPS publicly accessible repo (https://github.com/sigstore/helm-charts/pkgs/container/helm-charts%2Frekor/versions ). Looks correct to me. Feel free to test for yourself with different semver ranges ([see supported semver constraint definitions here](https://github.com/Masterminds/semver)).

```sh
$ make

# Ranges
$ ./bin/helm install --dry-run oci://ghcr.io/sigstore/helm-charts/rekor --generate-name --set server.ingress.enabled=false --version "<0.2.10" | grep helm.sh/chart | head -n1
    helm.sh/chart: rekor-0.2.

$ ./bin/helm install --dry-run oci://ghcr.io/sigstore/helm-charts/rekor --generate-name --set server.ingress.enabled=false --version ">0.2.9"| grep helm.sh/chart | head -n1
    helm.sh/chart: rekor-0.2.10

# PATCH wildcard
$ ./bin/helm install --dry-run oci://ghcr.io/sigstore/helm-charts/rekor --generate-name --set server.ingress.enabled=false --version "0.2.*" | grep helm.sh/chart | head -n1
    helm.sh/chart: rekor-0.2.10

# Missing tags
$ ./bin/helm install --dry-run oci://ghcr.io/sigstore/helm-charts/rekor --generate-name --set server.ingress.enabled=false --version "artifacthub.io" | grep helm.sh/chart | head -n1
Error: INSTALLATION FAILED: failed to download "oci://ghcr.io/sigstore/helm-charts/rekor" at version "artifacthub.io"

$ ./bin/helm install --dry-run oci://ghcr.io/sigstore/helm-charts/rekor --generate-name --set server.ingress.enabled=false --version "0.1.*" | grep helm.sh/chart | head -n1
Error: INSTALLATION FAILED: failed to download "oci://ghcr.io/sigstore/helm-charts/rekor" at version "0.1.*"

# No version passed
$ ./bin/helm install --dry-run oci://ghcr.io/sigstore/helm-charts/rekor --generate-name --set server.ingress.enabled=false | grep helm.sh/chart | head -n1
    helm.sh/chart: rekor-0.2.10
```

pkg/downloader/manager.go
```diff
@@ -578,7 +578,8 @@ func (m *Manager) resolveRepoNames(deps []*chart.Dependency) (map[string]string,
 	missing := []string{}
 	for _, dd := range deps {
 		// Don't map the repository, we don't need to download chart from charts directory
-		if dd.Repository == "" {
+		// When OCI is used there is no Helm repository
+		if dd.Repository == "" || registry.IsOCI(dd.Repository) {
```
mattfarina, line 582: You might want to rebase off `main` or otherwise undo this change. Adding this back causes an error.

pkg/action/pull.go
```diff
@@ -92,13 +92,13 @@ func (p *Pull) Run(chartRef string) (string, error) {
...
-
 		c.Options = append(c.Options,
-			getter.WithRegistryClient(p.cfg.RegistryClient),
-			getter.WithTagName(p.Version))
+			getter.WithRegistryClient(p.cfg.RegistryClient))
+
+		if p.Version != "" {
+			c.Options = append(c.Options,
+				getter.WithTagName(p.Version))
+		}
```
mattfarina, line 101: The version here isn't a version but can be a specific version or range. In either case, it's not a tag. So, using `WithTagName` don't convey what it does. As an external API to API users this is confusing.

Also, the other built-in getters don't turn a version range to a version in this part of the code. Doing it here makes it different from the others. I would pull the version part from here. See my other comments on where to put the version handling.

pkg/getter/ocigetter.go
```diff
@@ -50,10 +52,28 @@ func (g *OCIGetter) get(href string) (*bytes.Buffer, error) {
...
+	// If exact version, try to find it
+	// If semver constraint string, try to find a match
+	providedVersion := g.opts.version
+
+	tag, err := registry.GetTagMatchingVersionOrConstraint(tags, providedVersion)
+	if err != nil {
+		return nil, err
 	}
 
+	ref = fmt.Sprintf("%s:%s", ref, tag)
```
mattfarina, line 75: Sorting out the version for the other downloaders happens in another place. In `DownloadTo` a version or range is passed. The first thing `DownloadTo` does is call `c.ResolveChartVersion` which turns a version into a URI to call. This is then passed off to a getter which does the thing it was designed to do... get the thing.

Shouldn't this getter function using a similar code flow so the logic is in the same places?

mattfarina (changes requested):

pkg/downloader/manager.go
scottrigby: Yes thank you, will do

scottrigby: cherry-picked ee73a0263c02a6c53261fdaf2ad97542b18a844c from https://github.com/helm/helm/pull/10535 back into this branch 👍

mattfarina (changes requested):
:(

Resolving and pulling images with build metadata fails. For example, if I try to pull `0.3.0+build.1` and another one of `0.3.0` exists it will pull the one without the build metadata.

internal/experimental/registry/client.go
```diff
@@ -539,3 +572,53 @@ func PushOptStrictMode(strictMode bool) PushOption {
...
+			return nil, err
+		}
+
+		break
+
+	}
+
+	var tagVersions []*semver.Version
+	for _, tag := range registryTags {
+		tagVersion, err := semver.StrictNewVersion(tag)
```
mattfarina, line 607:
```
tagVersion, err := semver.StrictNewVersion(strings.ReplaceAll(tag, "_", "+"))
```

Need to make sure the _ has been turned back into a + before parsing the version

mattfarina (changes requested):

internal/experimental/registry/client.go
scottrigby: Ah, thanks. I even left a note under "special notes" https://github.com/helm/helm/pull/10521 so we didn’t forget to make sure to add back to this PR. Guess what, I forgot 😛

mattfarina (approved):

pkg/action/pull.go
scottrigby: done

pkg/getter/ocigetter.go
scottrigby: done

scottrigby:
Can not emphasize enough how much work went into this and how good the team collaboration was

# 👏 🥇 🎉

PR #10535, opened 2022-01-10
Fixing issue where OCI handling early causes a bad message

### LINKED ISSUE #10534: OCI registry being treated as Helm registry with failed update
When running `helm dep up` on a dependency with a chart stored in an OCI registry I'm seeing the message:

```
object required
...Unable to get an update from the "oci://ghcr.io/mattfarina" chart repository:
	object required
```

This is showing a bad message (i.e. `object required`) and it's trying to treat OCI as if it's a Helm repository.

OCI registries should not be attempted to be updated.

Output of `helm version`: tip of main (also in 3.7)

mattfarina:
Note, there is OCI handling later in the funtion that should
handle the situation instead.

Closes #10534

See the issue. A bad output message.

Look no further for the blame of this bug than the author of the fix.

**If applicable**:

### DIFF
--- a/pkg/downloader/manager.go
+++ b/pkg/downloader/manager.go
@@ -578,8 +578,7 @@ func (m *Manager) resolveRepoNames(deps []*chart.Dependency) (map[string]string,
 	missing := []string{}
 	for _, dd := range deps {
 		// Don't map the repository, we don't need to download chart from charts directory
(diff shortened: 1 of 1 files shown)


scottrigby (approved):
Thank you @mattfarina 💯

PR #10538, opened 2022-01-11
Add --jsonpath to helm show chart

gferon:
In some CI pipelines, we want to get some fields of `Chart.yaml` and JSONpath templates are the best way to achieve this. Since `helm` right now prints the chart metadata as YAML, it's not really easy.

- `--jsonpath` is already a flag in `helm show values`, and it uses the same approach.
- I couldn't find any unit tests of `--jsonpath` for `helm show values` and have decided not to add any


yxxhero:
@gferon Thanks for this PRs. Can you add some unittest for this pr.

PR #10554, opened 2022-01-13
feat(comp): Shell completion for OCI

### LINKED ISSUE #10553: Add shell completion for OCI feature
The new OCI commands and charts should have shell completion when possible.

marckhouzam:

Adds shell completion for the OCI feature

Fixes #10553 

There are a couple of things I would like to clarify in my understanding of the current OCI support.  I will post review comments to make it easier to discuss.

### DIFF
--- a/cmd/helm/push.go
+++ b/cmd/helm/push.go
@@ -24,6 +24,7 @@ import (
 
 	"helm.sh/helm/v3/cmd/helm/require"
 	"helm.sh/helm/v3/pkg/action"
+	"helm.sh/helm/v3/pkg/pusher"
 )
 
 const pushDesc = `
@@ -41,6 +42,23 @@ func newPushCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 		Short: "push a chart to remote",
 		Long:  pushDesc,
 		Args:  require.MinimumNArgs(2),
+		ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+			if len(args) == 0 {
+				// Do file completion for the chart file to push
+				return nil, cobra.ShellCompDirectiveDefault
+			}
+			if len(args) == 1 {
+				providers := []pusher.Provider(pusher.All(settings))
+				var comps []string
+				for _, p := range providers {
+					for _, scheme := range p.Schemes {
+						comps = append(comps, fmt.Sprintf("%s://", scheme))
+					}
+				}
+				return comps, cobra.ShellCompDirectiveNoFileComp | cobra.ShellCompDirectiveNoSpace
+			}
+			return nil, cobra.ShellCompDirectiveNoFileComp
+		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			chartRef := args[0]
 			remote := args[1]
--- a/cmd/helm/registry_login.go
+++ b/cmd/helm/registry_login.go
@@ -41,10 +41,11 @@ func newRegistryLoginCmd(cfg *action.Configuration, out io.Writer) *cobra.Comman
 	var passwordFromStdinOpt, insecureOpt bool
 
 	cmd := &cobra.Command{
-		Use:   "login [host]",
-		Short: "login to a registry",
-		Long:  registryLoginDesc,
-		Args:  require.MinimumNArgs(1),
+		Use:               "login [host]",
+		Short:             "login to a registry",
+		Long:              registryLoginDesc,
+		Args:              require.MinimumNArgs(1),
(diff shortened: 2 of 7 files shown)


marckhouzam:
This is a small change but the bot labelled it "L" because of the copyright headers for three new tests files.

cmd/helm/push.go
```diff
@@ -41,6 +41,16 @@ func newPushCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 		Short: "push a chart to remote",
 		Long:  pushDesc,
 		Args:  require.MinimumNArgs(2),
+		ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+			if len(args) == 0 {
+				// Do file completion for the chart file to push
+				return nil, cobra.ShellCompDirectiveDefault
```
marckhouzam, line 47: The documentation here https://helm.sh/docs/topics/registries/ mentions that "The push subcommand can only be used against .tgz files".  But I when I renamed a package created with helm to any other name, I was still able to push it.

Therefore, I did not limit the file completion to files ending with `.tgz`.

cmd/helm/registry_login.go
```diff
@@ -41,10 +41,11 @@ func newRegistryLoginCmd(cfg *action.Configuration, out io.Writer) *cobra.Comman
...
 	cmd := &cobra.Command{
-		Use:   "login [host]",
-		Short: "login to a registry",
-		Long:  registryLoginDesc,
-		Args:  require.MinimumNArgs(1),
+		Use:               "login [host]",
+		Short:             "login to a registry",
+		Long:              registryLoginDesc,
+		Args:              require.MinimumNArgs(1),
+		ValidArgsFunction: noCompletions,
```
marckhouzam, line 48: Must a registry start with "oci://" or could it be something else? If it MUST be "oci://" we could have the shell completion suggest it.

cmd/helm/registry_logout.go
```diff
@@ -31,10 +31,11 @@ Remove credentials stored for a remote registry.
...
 	return &cobra.Command{
-		Use:   "logout [host]",
-		Short: "logout from a registry",
-		Long:  registryLogoutDesc,
-		Args:  require.MinimumNArgs(1),
+		Use:               "logout [host]",
+		Short:             "logout from a registry",
+		Long:              registryLogoutDesc,
+		Args:              require.MinimumNArgs(1),
+		ValidArgsFunction: noCompletions,
```
marckhouzam, line 38: Is it safe to obtain a list of registries for which the user is logged in by looking at the `$HELM_REGISTRY_CONFIG` file?  Or could registries be stored elsewhere (e.g., `~/.docker/config.json`).

It would be nice to complete the registry addresses that the user can actually logout from, if we can find an exhaustive list.

cmd/helm/registry_login.go
marckhouzam: I see that in https://helm.sh/docs/topics/registries/ there is an example that does `helm registry login -u myuser localhost:5000`.  So the `oci://` prefix is not necessary, so we shouldn't suggest it.

cmd/helm/registry_logout.go
marckhouzam: To do this, I think we should use an ORAS API to obtain the list.  I haven't had time to look if such an API exists yet.

cmd/helm/push.go
scottrigby: Oh interesting. Perhaps we should limit this in `push` as a follow-up. Makes sense for now 👌

cmd/helm/registry_logout.go
jdolitsky: Its not so straightforward, as certain registries may skip `login` altogether and use credential helpers

jdolitsky (approved):

cmd/helm/registry_logout.go
marckhouzam: Ok, so let's not provide any completions right now

marckhouzam:
To be clear, making shell completion work better with OCI could be seen as a feature.  But if we consider that helm provides full support for shell completion for every command, then the lack of such support for OCI can be seen as a "bug", which then justifies it being in 3.8.1 instead of 3.9.  

I'm bias when it comes to shell completion so I abstain from such a decision (I probably value shell completion way more than a reasonable person 😉 ).  But since the milestone has been set to 3.8.1 by another maintainer (👍 ), I have labelled the PR as "needs-pick".  However, we need to wait until 3.8.0 is cut before doing the cherry-pick for 3.8.1.

PR #10568, opened 2022-01-17
HTTPGetter: Reuse http transport

matthiasfehr:
The `HTTPGetter` makes a new instance of the `http.Transport` on each HTTP request, which according to the docs should be avoided.

From the go docs (https://github.com/golang/go/blob/master/src/net/http/client.go#L32 and https://github.com/golang/go/blob/master/src/net/http/transport.go#L68)
```
The Client's Transport typically has internal state (cached TCP connections)
...
Transports should be reused instead of created as needed.
```

In polling clients which make use of this `HTTPGetter` like https://github.com/fluxcd/source-controller this can cause problems as for every request a new TCP connection is created and never closed, which in our case went to thousands of open TCP connections from this polling client.

This PR makes a single transport per `HTTPGetter` client and reuses it for future requests.

**If applicable**:

### DIFF
--- a/pkg/getter/getter.go
+++ b/pkg/getter/getter.go
@@ -121,6 +121,7 @@ func WithUntar() Option {
 	}
 }
 
+// WithTransport sets the http.Transport to allow overwriting the HTTPGetter default.
 func WithTransport(transport *http.Transport) Option {
 	return func(opts *options) {
 		opts.transport = transport
--- a/pkg/getter/getter.go
+++ b/pkg/getter/getter.go
@@ -18,6 +18,7 @@ package getter
 
 import (
 	"bytes"
+	"net/http"
 	"time"
 
 	"github.com/pkg/errors"
@@ -43,6 +44,7 @@ type options struct {
 	version               string
 	registryClient        *registry.Client
 	timeout               time.Duration
+	transport             *http.Transport
 }
 
 // Option allows specifying various settings configurable by the user for overriding the defaults
@@ -116,6 +118,12 @@ func WithRegistryClient(client *registry.Client) Option {
 func WithUntar() Option {
 	return func(opts *options) {
 		opts.unTar = true
+	}
+}
+
+func WithTransport(transport *http.Transport) Option {
+	return func(opts *options) {
+		opts.transport = transport
 	}
 }
 
--- a/pkg/getter/httpgetter.go
+++ b/pkg/getter/httpgetter.go
@@ -109,20 +109,19 @@ func NewHTTPGetter(options ...Option) (Getter, error) {
 }
 
 func (g *HTTPGetter) httpClient() (*http.Client, error) {
-	var transport *http.Transport
-
 	if g.opts.transport != nil {
-		transport = g.opts.transport
-	} else {
-		g.once.Do(func() {
-			g.transport = &http.Transport{
-				DisableCompression: true,
-				Proxy:              http.ProxyFromEnvironment,
-			}
-		})
-
-		transport = g.transport
+		return &http.Client{
+			Transport: g.opts.transport,
+			Timeout:   g.opts.timeout,
+		}, nil
 	}
+
+	g.once.Do(func() {
+		g.transport = &http.Transport{
+			DisableCompression: true,
+			Proxy:              http.ProxyFromEnvironment,
+		}
+	})
 
 	if (g.opts.certFile != "" && g.opts.keyFile != "") || g.opts.caFile != "" {
 		tlsConf, err := tlsutil.NewClientTLS(g.opts.certFile, g.opts.keyFile, g.opts.caFile)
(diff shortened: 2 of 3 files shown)


pkg/getter/httpgetter.go
```diff
@@ -106,10 +107,13 @@ func NewHTTPGetter(options ...Option) (Getter, error) {
 }
 
 func (g *HTTPGetter) httpClient() (*http.Client, error) {
-	transport := &http.Transport{
-		DisableCompression: true,
-		Proxy:              http.ProxyFromEnvironment,
+	if g.transport == nil {
+		g.transport = &http.Transport{
+			DisableCompression: true,
+			Proxy:              http.ProxyFromEnvironment,
```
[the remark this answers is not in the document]
matthiasfehr: That's the same configuration as before, so I didn't wanted to change these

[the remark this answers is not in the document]
yxxhero: @matthiasfehr Oh. that's ok. Thanks for your PR.

yxxhero:
@matthiasfehr Can you try add unittest and a issue for this pr.

pkg/getter/httpgetter.go
```diff
@@ -123,21 +127,21 @@ func (g *HTTPGetter) httpClient() (*http.Client, error) {
 		}
 		tlsConf.ServerName = sni
 
-		transport.TLSClientConfig = tlsConf
+		g.transport.TLSClientConfig = tlsConf
```
pjbgf, line 130: For users providing a non-ephemeral transport would it make sense for the `httpgetter` to amend its internals? I had the same debate whilst looking at this problem, and feels problematic the possibility of passing a transport to this func and have side effects in other places that share the same transport instance.

For example, depending on the order of the calls, something else completely unrelated to helm can be called after this func and from that point onwards have `InsecureSkipVerify` set to `true`, leaving callers in a insecure position.

To avoid such side effects, it may be worth making the caller fully responsible for configuring the transport if they so wish. So they either rely on `httpgetter` to manage the transport, or they take that onto themselves. Using an in-between approach may lead to unexpected and potentially less secure outcomes.

pkg/getter/httpgetter.go
```diff
@@ -106,10 +107,13 @@ func NewHTTPGetter(options ...Option) (Getter, error) {
 }
 
 func (g *HTTPGetter) httpClient() (*http.Client, error) {
-	transport := &http.Transport{
-		DisableCompression: true,
-		Proxy:              http.ProxyFromEnvironment,
+	if g.transport == nil {
+		g.transport = &http.Transport{
```
pjbgf, line 111: It may be worth creating the transport on an [once.Do() block](https://pkg.go.dev/sync#Once.Do) to avoid race conditions.

hiddeco (commented):
For flexibility reasons (e.g. providing a different non-environmental non-global proxy configuration), I would prefer a solution which also allows injection of a specific `http.Transport` as an addition to lazy-loading (as done in #10668).

matthiasfehr:
I also like the idea of having an option to set the transport. I updated the code to have an option like in https://github.com/helm/helm/pull/10668, with the addition if none is set that only the first time a transport is created and then reused.

pkg/getter/getter.go
```diff
@@ -119,6 +121,12 @@ func WithUntar() Option {
 	}
 }
 
+func WithTransport(transport *http.Transport) Option {
```
hiddeco, line 124:
```suggestion
// WithTransport sets the http.Transport to allow overwriting the HTTPGetter default.
func WithTransport(transport *http.Transport) Option {
```

hiddeco (approved):
Besides the missing doc block, this looks good to me. Thanks @matthiasfehr :bow:

pjbgf (approved):
@matthiasfehr really good stuff, thanks for working on this. :bow:

hiddeco:
Given we currently maintain a fork to allow configuring this, I am wondering what the chances are to get this merged @yxxhero. It would be of immense help to the [source-controller](https://github.com/fluxcd/source-controller) where we "mirror" charts while allowing per-object credential configurations, and do not want things to leak connections and/or credentials.

pkg/getter/httpgetter_test.go
```diff
@@ -443,3 +449,84 @@ func verifyInsecureSkipVerify(t *testing.T, g HTTPGetter, caseName string, expec
...
+		t.Fatal(err)
+	}
+
+	if httpClient2 == nil {
+		t.Fatalf("Expected non nil value for http client")
+	}
+
+	transport2 := (httpClient2.Transport).(*http.Transport)
+
+	if transport1 != transport2 {
```
yxxhero, line 518:
```go
if transport1 != transport
```
is better? What do you think?

jdolitsky (approved):
👍 looks good, thank you

pkg/getter/httpgetter_test.go
matthiasfehr: I test this already on line 502

scottrigby (approved):
Discussed in Helm dev meeting today. See follow-up PR note. Thanks everyone 👍  Glad to be prioritizing SDK users!

yxxhero:
> Discussed in Helm dev meeting today. See follow-up PR note. Thanks everyone 👍 Glad to be prioritizing SDK users!

ok Thanks very much.

PR #10574, opened 2022-01-18
refactor: use `os.ReadDir` for lightweight directory reading

Juneezee:
`os.ReadDir` is a more efficient than `ioutil.ReadDir` as stated here https://pkg.go.dev/io/ioutil#ReadDir. Since we are only using `IsDir()` and `Name()` in `safeMoveDeps`, we don't need to read the directory with full stat information using `ioutil.ReadDir`.

The full proposal can be read here: https://github.com/golang/go/issues/41467

**If applicable**:


bacongobbler (approved):

hickeyma (approved):
LGTM, thanks for updating this @Juneezee

PR #10575, opened 2022-01-18
chore (lint): Replace golint with revive

hickeyma:
`golint` which is used as one of the sub-linters in [golangci-lint](https://github.com/golangci/golangci-lint) is deprecated.
It is replaced with `revive` which is a drop-in replacement.

Signed-off-by: Martin Hickey <martin.hickey@ie.ibm.com>

Co-authored-by: Martin Mulholland <mmulholl@redhat.com>

**What this PR does / why we need it**:
[golangci-lint](https://github.com/golangci/golangci-lint) [deprecated the `golint` sublinter](https://github.com/golangci/golangci-lint/issues/1892). It needs to be replaced with `revive`.

This PR supersedes #10206 

**Special notes for your reviewer**:
How to handle the deprecation notices in the future is discussed here: golangci/golangci-lint#1987

Lint errors when `revive` aded as a sub-linter:

```console
$ make test-style

GO111MODULE=on golangci-lint run
pkg/registry/client.go:299:15: errorf: should replace errors.New(fmt.Sprintf(...)) with fmt.Errorf(...) (revive)
		return nil, errors.New(
			fmt.Sprintf("manifest does not contain minimum number of descriptors (%d), descriptors found: %d",
				minNumDescriptors, numDescriptors))
pkg/registry/client.go:321:15: errorf: should replace errors.New(fmt.Sprintf(...)) with fmt.Errorf(...) (revive)
		return nil, errors.New(
			fmt.Sprintf("could not load config with mediatype %s", ConfigMediaType))
pkg/registry/client.go:325:15: errorf: should replace errors.New(fmt.Sprintf(...)) with fmt.Errorf(...) (revive)
		return nil, errors.New(
			fmt.Sprintf("manifest does not contain a layer with mediatype %s",
				ChartLayerMediaType))
pkg/cli/environment_test.go:54:3: var-naming: don't use leading k in Go names; struct field kAsUser should be asUser (revive)
		kAsUser      string
		^
pkg/cli/environment_test.go:55:3: var-naming: don't use leading k in Go names; struct field kAsGroups should be asGroups (revive)
		kAsGroups    []string
		^
pkg/cli/environment_test.go:56:3: var-naming: don't use leading k in Go names; struct field kCaFile should be caFile (revive)
		kCaFile      string
		^
cmd/helm/completion_test.go:50:4: errorf: should replace t.Error(fmt.Sprintf(...)) with t.Errorf(...) (revive)
[...]

PR #10581, opened 2022-01-19
Set go version to 1.17 to match CI

### LINKED ISSUE #10580: Build fails with go1.16 but works with 1.17, Go.mod points to go1.16
The commit 3841af9 updates helm to use go 1.17 on the CI but we have this on master:

https://github.com/helm/helm/blob/86a94f2f53b0519a3b22648a11f332c85cd8da1a/go.mod#L3

When building I encountered this error:

```
/data/go/pkg/mod/sigs.k8s.io/json@v0.0.0-20211020170558-c049b76a60c6/internal/golang/encoding/json/encode.go:1255:18: sf.IsExported undefined (type reflect.StructField has no field or method IsExported)
```

And found [here](https://github.com/kubernetes-sigs/controller-tools/issues/643#issuecomment-1004459362)  that updating to go1.17 fixes the issue.

I think, we need to update the go version in go.mod to 1.17.

smaftoul:
Closes #10580 

Bump go version to 1.17 to match CI. It's needed because building with 1.16 is broken and because the CI uses 1.17.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module helm.sh/helm/v3
 
-go 1.16
+go 1.17
 
 require (
 	github.com/BurntSushi/toml v0.4.1


smaftoul:
CI Has this error:
```
ERRO Running error: context loading failed: no go files to analyze
```

The PR doesn't contain any modification to go files, I guess this is normal (but probably should not fail ?)

bacongobbler:
likely fixed once https://github.com/helm/helm/pull/10575 is merged - golint is deprecated.

bacongobbler:
For reference, the version number listed in go.mod dictates the minimum version of Go you can use to compile the project. It does not have to match the same version in CI - in fact it can be several versions behind. However in the case described in #10580, it sounds like Helm cannot be compiled against Go 1.16 due to an issue with a transitive dependency, so bumping go.mod to a version that does compile seems fine to me.

bacongobbler (approved):
LGTM once we merge #10575 and this PR is rebased to ensure everything still works

bacongobbler:
> I think, the CI should use the minimum version of Go you can use to compile the project, so this situation doesn't happen.

I agree that we should run the test suite against all supported versions of Go (and we'd appreciate a PR to do so next time!). But I don't think we should release new versions of Helm compiled against the minimum version. Newer versions of Go contain performance improvements and bug fixes, many of which will benefit Helm users, which is why we compile Helm against the latest stable version of Go.

bacongobbler:
Can you please rebase? Thanks

smaftoul:
@bacongobbler Done !

bacongobbler (approved):

Issue #9892, opened 2021-06-30
helm test --logs fails on tests with more than one container


dwaynebailey:
Output of `helm version`: 
version.BuildInfo{Version:"v3.6.1", GitCommit:"61d8e8c4a6f95540c15c6a65f36a6dd0a45e7a2f", GitTreeState:"dirty", GoVersion:"go1.16.5"}

Output of `kubectl version`:
Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.2", GitCommit:"092fbfbf53427de67cac1e9fa54aaa09a28371d7", GitTreeState:"clean", BuildDate:"2021-06-16T12:52:14Z", GoVersion:"go1.16.5", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"20+", GitVersion:"v1.20.7-gke.1800", GitCommit:"cd6f9121893d6cdb30e3c48d57830ed44411e4ba", GitTreeState:"clean", BuildDate:"2021-06-07T09:17:22Z", GoVersion:"go1.15.12b5", Compiler:"gc", Platform:"linux/amd64"}

Cloud Provider/Platform (AKS, GKE, Minikube etc.): 
GKE

I have a test to check that we can login to Google Cloud SQL. To access Cloud SQL I'm using the recommended approach of a sidecar proxy.  I've got his working and the test passes.

I'm using this in a pipeline where I run `helm test --log` to retrieve the logs so that they're available in the case of a failure.

I get this failure in a test run:
```Error: unable to get pod logs for test-db-credentials: a container name must be specified for pod test-db-credentials, choose one of: [credentials-test cloud-sql-proxy]```

**Comment by bacongobbler** [comment]
https://github.com/helm/helm/issues/6902 was opened asking for similar functionality, but was closed due to inactivity. Feel free to look into this one.

**Comment by dwaynebailey** [comment]
@bacongobbler thanks, looks very similar to the solution I crafted:

```bash
helm test -n $NAMESPACE $RELEASE --timeout=15m
for pod in $(kubectl get pods -n $NAMESPACE -l helm.sh/hook=test -o name)
do
  kubectl logs -n $NAMESPACE --all-containers $pod
done
```

Issue #6902, opened 2019-11-07
helm test --logs for all containers in a pod


tanenbaum:
Output of `helm version`:
```
Client: &version.Version{SemVer:"v2.16.0", GitCommit:"e13bc94621d4ef666270cfbe734aaabf342a49bb", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.16.0", GitCommit:"e13bc94621d4ef666270cfbe734aaabf342a49bb", GitTreeState:"clean"}
```

Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-17T17:16:09Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-15T19:09:08Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}
```

Cloud Provider/Platform (AKS, GKE, Minikube etc.): Minikube

It would be great if `helm test --cleanup --logs` printed all container logs in each test pod. Currently this works fine for 1 container in a pod.

Rather than run test images that contain all the tools I need, I often create pods with a couple of containers that orchestrate the tests (e.g. redis-cli + python script + shared volume).

`helm test [release] --cleanup --logs` will error with:
RUNNING: my-test
FAILED: my-test, run `kubectl logs my-test --namespace default` for more info
ERROR: error in opening log stream, got: a container name must be specified for pod my-test, choose one of: [container-0 container-1]

I currently use a shell script as a workaround (can't find source of this anymore) like so:

```
helm_release=$1

helm test $helm_release
status=$?

test_pods=$(helm status $helm_release -o json | jq -r .info.status.last_test_suite_run.results[].name)
namespace=$(helm status $helm_release -o json | jq -r .namespace)

for test_pod in $test_pods; do
  echo "Test Pod: $test_pod"
  echo "============"
  echo ""
  kubectl -n $namespace logs --all-containers $test_pod
  kubectl -n $namespace delete pod $test_pod
  echo ""
  echo "============"
done

exit $status
```

[...]

PR #10603, opened 2022-01-22
Trigger hook-delete-policy after log retrieval for helm test

### LINKED ISSUE #9098: Helm test --logs fails with hook-delete-policy "hook-failed" or "hook-succeed"
When I execute:

`helm test <release> --logs`

I get the following error:

`unable to get pod logs for <release>: pods "<release>" not found`

This only happens, if I set `helm.sh/hook-delete-policy` to `hook-failed` or `hook-succeeded`

While in general, it makes sense that no logs can be retrieved, if the test pod was deleted, in Helm2 this used to work. The logs were retrieved before deletion of the pods.

Output of `helm version`: version.BuildInfo{Version:"v3.2.0", GitCommit:"e11b7ce3b12db2941e90399e874513fbd24bcb71", GitTreeState:"clean", GoVersion:"go1.13.10"}

Output of `kubectl version`: Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.3", GitCommit:"1e11e4a2108024935ecfcb2912226cedeafd99df", GitTreeState:"clean", BuildDate:"2020-10-14T12:50:19Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"linux/amd64"}

mfbieber:

Closes #9098 **Helm test --logs fails with hook-delete-policy "hook-failed" or "hook-succeed"**

If we want to run `helm test --logs`, this results in an error like:
````
$ helm --namespace test validation --logs
NAME: validation
LAST DEPLOYED: Sat Jan 22 22:43:19 2022
NAMESPACE: test
STATUS: deployed
REVISION: 2
TEST SUITE:     test-secrets-store
Last Started:   Sat Jan 22 22:43:25 2022
Last Completed: Sat Jan 22 22:43:27 2022
Phase:          Succeeded

Error: unable to get pod logs for test-secrets-store: pods "test-secrets-store" not found
````

When calling `helm test`, the following functions are called in the code:
1. `helm test` results in a call of `newReleaseTestCmd()` in `helm\release_testing.go`
2. `newReleaseTestCmd()` results in a call of `Run(args[0])`  in `action\release_testing.go`
3. `Run(args[0])` results in a call of `execHook(rel, release.HookTest, r.Timeout)` in `hooks.go`
4.  `execHook(rel, release.HookTest, r.Timeout)` results in a deletion of Pods, if e.g. `"helm.sh/hook-delete-policy": "before-hook-creation,hook-succeeded,hook-failed"` has been specified by e.g. calling `deleteHookByPolicy(h, release.HookSucceeded)`
5. when all of this returns to `newReleaseTestCmd()` in `helm\release_testing.go`, where it tries to retrieve the logs by calling `GetPodLogs(out, rel)`, the Pods have already been deleted, resulting in the error we see.

I've added a shutdown hook, so that the `newReleaseTestCmd()` in `helm\release_testing.go` function can trigger the deletion of Pods as it would have done before, but using a `defer` in `newReleaseTestCmd()` that calls that hook function.

I also made sure not to change the signature of `execHook()` in `hooks.go`. The functions that have modified signatures are only called in the context described above and are thus ok to modify.

I'll add some comments to this PR in the code since I am not completely sure if the error handling is as we want it to be. Please have a look at it and advise if you think it is ok, or if we should refactor it.

### CODE REVIEW

File: cmd/helm/release_testing.go [hunk 1]
```diff
@@ -64,7 +64,16 @@ func newReleaseTestCmd(cfg *action.Configuration, out io.Writer) *cobra.Command
...
-			rel, runErr := client.Run(args[0])
+
+			rel, shutdown, runErr := client.Run(args[0])
+			defer func() {
+				if shutdownErr := shutdown(); shutdownErr != nil {
+					if returnError == nil {
+						returnError = shutdownErr
+					}
+				}
+			}()
```
mfbieber, line 75: Before, the error returned was the last error that was encountered in this function `newReleaseTestCmd()`. That could have been e.g. an error while retrieving the logs.

By using `defer shutdown()` we would have changed the behavior since now a possible error of the deletion of the pods would be the error to be returned. Before, this error possibly was ignored if `rel` was nil.

pkg/action/hooks.go
```diff
@@ -28,6 +28,24 @@ import (
 
 // execHook executes all of the hooks for the given hook event.
 func (cfg *Configuration) execHook(rl *release.Release, hook release.HookEvent, timeout time.Duration) error {
+	shutdown, err := cfg.execHookWithDelayedShutdown(rl, hook, timeout)
+	if err != nil {
+		if err := shutdown(); err != nil {
```
mfbieber, line 33: My IDE (GoLand)  warns for `shutdown()` in this line:
>  'shutdown' might have 'nil' or other unexpected value as its corresponding error variable might be not 'nil' 

From my implementation, I don't see how that could be `nil`. Since I am quite new to go: is there a possibility to specify that that function cannot be `nil`? Something like a `@Nonnull` in Java? Or would you ignore this warning or write these parts differently?

fcervantes2: for the type of the `shutdown` returned, the default value if none is assigned would be `nil`; uninitialized func variables are `nil`. Even though looking at the implementation for `execHookWithDelayedShutdown`, it does appear to "never" return a `nil` function, it's still possible from the perspective of the client, since the actual value of it cannot be inferred at compile time. To be safe (and to appease your IDE), you should also check for a `nil` shutdown function and use the no-op function if nil. 

That might mean that you can change your implementation to return a `nil` shutdown, and let users of your function determine what to do in such cases.

jnodorp-jaconi:
As one of our customers is affected by this, I'd love to see this PR getting in. From the contributing guidelines it looks like the next step is a triage of the PR. @joejulian, @yxxhero, or @zonggen, you are listed for triage in the OWNERS file. Does any of you have the capacity to look at this?

yxxhero:
@jnodorp-jaconi I will see this later.

sethgupton-mastery:
Thank you @mfbieber ! Looking forwards to this getting merged!

pkg/action/hooks.go
mfbieber: Thanks @fcervantes2! I added 
````
        if shutdown == nil {
		return nil
	}
````
to the function. I checked for the usages of `execHook()` and they all do a nil check - so that should be indeed be handled already.

**Comment by mfbieber** [comment]
Thanks! Actually, I had compiled and tested the implementation of course. But please verify!

**Comment by jnodorp-jaconi** [comment]
> @jnodorp-jaconi I will see this later.

Any update on this?

**Comment by micxer** [comment]
Please someone review and merge it if it fixes the issue! This is really annoying and not really helpful.

**Comment by joejulian** [comment]
I just noticed I was pinged on this back in April. Sorry, the triage maintainers just triage issues. We have no :heavy_check_mark:'s to give.

**Comment by srbhklkrn** [comment]
Any update on this?

**Comment by mfbieber** [comment]
@joejulian and @yxxhero: obviously I won't review my own PR, but for the future: it seems like Helm needs more reviewers/maintainers. How can I help?

**Comment by joejulian** [comment]
Yes, the project does need more maintainers. The first step (: heavy_checkmark:), contribute code.

Come join the developer meeting, if you can: https://github.com/helm/community/blob/main/communication.md#meetings.

Post your interest in becoming a maintainer on the [mailing list](https://github.com/helm/community/blob/main/communication.md#mailing-lists).

I also recommend familiarizing yourself with the [existing hips](https://github.com/helm/community/tree/main/hips#existing-hips)

File: pkg/action/hooks.go [hunk 3]
```diff
@@ -28,6 +28,27 @@ import (
...
+	return shutdown()
+}
+
+type ExecuteShutdownHooks = func() error
+
+func ShutdownNoOp() error {
+	return nil
+}
+
+// execHook executes all of the hooks for the given hook event and returns a shutdownHook function to trigger deletions after doing other things like e.g. retrieving logs.
```
fty4, line 50:
```suggestion
// execHookWithDelayedShutdown executes all of the hooks for the given hook event and returns a shutdownHook function to trigger deletions after doing other things like e.g. retrieving logs.
```

scottrigby:
@mfbieber can you rebase this against either `main` (if it still applies to Helm 4, and then backport to `dev-v3` in a separate PR), or if it only applies to Helm 3 now rebase this against `dev-v3` branch?

svalbuena:
this bug is very annoying, it gets impossible to debug tests that fail without seeing the pod logs, when will this be merged and released? seems like the coding part is already done

mueslo:
unfortunately due to a related bug the very straightforward use case of "i want to see the test output" still fails if there are multiple containers in a test pod due to a related issue that has been present for a long time:  https://github.com/helm/helm/issues/6902 https://github.com/helm/helm/issues/9892

PR #10616, opened 2022-01-27
Improve URL credential handling

devfbe:
Currently, if the index.yaml of a helm repo contains entries, pointing to the same repo but with explicit port declaration of port 80 for http or 443 for https, the helm client needs the flag `--pass-credentials` to work, because it thinks that http://foo.com:80 is not the same location as http://foo.com. In fact, for https and http we can always say, that the default port is 443/80.

There are big helm repository managers like jfrog artifactory 6, which show that behavior (adding the port to the index.yaml for https / http). There it's then necessary to add the `--pass-credentials` flag which is really dangerous, because it can lead to accidentially forwarded secrets to external registries - as it happened in the company I am working for.

This pull request mitigates this issue and implements that the http(s) urls with and without default port declaration matches, so that in the case just described, the helm client doesn't nee the `--pass-credentials` anymore.

This merge request consists of 3 commits. First ist just a tiny cleanup of a unit test (eliminate dead code), just a warmup.

Because integration testing with port 80 / 443 is not simply possible, because it would require root access / special permissions on the dev machine, I decided to extract the check logic the urlutil in the second commit, keeping the exact same logic to make sure that I don't break anything and guard the logic with a unit test.

The third commit then introduces the change which contains the special handling for http(s) schemes when no port is declared explicitly.

**If applicable**:

### DIFF
--- a/internal/urlutil/urlutil.go
+++ b/internal/urlutil/urlutil.go
@@ -71,3 +71,11 @@ func ExtractHostname(addr string) (string, error) {
 	}
 	return u.Hostname(), nil
 }
+
+// SchemeHostAndPortMatches returns if the scheme, port and hostname of the given url matches
+func SchemeHostAndPortMatches(u1, u2 *url.URL) bool {
+	// Host on URL (returned from url.Parse) contains the port if present.
+	// This check ensures credentials are not passed between different
+	// services on different ports.
+	return u1.Scheme == u2.Scheme && u1.Host == u2.Host
+}
--- a/internal/urlutil/urlutil.go
+++ b/internal/urlutil/urlutil.go
@@ -77,5 +77,19 @@ func SchemeHostAndPortMatches(u1, u2 *url.URL) bool {
 	// Host on URL (returned from url.Parse) contains the port if present.
 	// This check ensures credentials are not passed between different
 	// services on different ports.
-	return u1.Scheme == u2.Scheme && u1.Host == u2.Host
+	getPort := func(url *url.URL) string {
+		if url.Port() == "" {
+			if url.Scheme == "https" { // is always lower case
+				return "443"
+			}
+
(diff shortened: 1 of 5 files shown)


devfbe:
Hi @bacongobbler,

the RFC says:
>
> URI producers and
> normalizers should omit the port component and its ":" delimiter if
> port is empty or if its value would be the same as that of the
> scheme's default.
>
It's a "should", not a "must".  I think it's okay when Helm normalizes the url to avoid problems when the repository manager suddenly removes the :443.  Helm does not "normalize" repo URLs when adding with `helm repo add .. https://...:443/..`, too. So I think this would be consistent behavior.

matthyx:
+1 here
@bacongobbler can you reply to the latest comment from @devfbe ?
I wouldn't want to run a forked version of Helm just because of JFrog...

coxalass:
This bug was already raised to JFrog support, see https://www.jfrog.com/jira/browse/RTFACT-26063.

This issue is flagged as resolved in version 7.24, however, we are running 7.38.10 and Artifactory is still adding https default port 443 to the url of virtual repositories...

I cloned the original issue, please vote on it : https://www.jfrog.com/jira/browse/RTFACT-27496

devfbe:
Hey @coxalass, try to remove the $server_port from the "X-JFrog-Override-Base-Url" header passed, when using the generated nginx configuration for Artifactory.

Bad
```
proxy_set_header    X-JFrog-Override-Base-Url $http_x_forwarded_proto://$host:$server_port;
```

Good:
```
proxy_set_header    X-JFrog-Override-Base-Url $http_x_forwarded_proto://$host;
```

Thats what the JFrog support told us (they don't consider fixing their shitty nginx config generation wizard 🤦).

At least that works for us.

Issue #10624, opened 2022-01-28
Azure acr - Could not retrieve list of tags for repository oci...


michal-kania:
Command from azure dev ops windows agent:
`helm dependency update ./filePath`

The chart in ./filePath contains dependencies like 
```
dependencies:
- name: yyyy
  version: ...
  repository: "oci://redacted.azurecr.io/helm/xxx"
```
Helm 3.8.0
**Error: could not retrieve list of tags for repository oci://redacted.azurecr.io/helm/xxx: GET "https://redacted.azurecr.io/v2/helm/xxx/yyyy/tags/list": GET "https://redacted.azurecr.io/oauth2/token?scope=repository%3Ahelm%2Fxxx%2Fyyyy%3Ametadata_read%2Cpull&service=redacted.azurecr.io": unexpected status code 401: unauthorized: authentication required, visit https://aka.ms/acr/authorization for more information.**

It worked with helm 3.7.0

I tested it with service principal and acr admin user, but both of them are probably missing posibility for metadata/read permission described here:
https://docs.microsoft.com/en-us/azure/container-registry/container-registry-repository-scoped-permissions#concepts
Helm 3.8.0 works only with ACR Tokens (preview)
https://docs.microsoft.com/en-us/azure/container-registry/container-registry-repository-scoped-permissions#create-token---portal
with metadata/read permission scope.
This is marked as Preview and allowed only with azure container registry Premium SKU.

Is it intended behavior?
Is there any posibility to skip GET .../tags/list which might not be necessary and i am already specifiying dependency version?

Thanks.

jdolitsky:
Hi. Would you be able to share what you have for the `version` value? It is my understanding that if this is an exact version, tag listing will not/should not occur

bacongobbler:
Possibly related to #10608?

michal-kania:
Hi, version is (in my case) replaced by previous pipeline step and it is specified exactly.
During the day i was abel to force previous helm 3.7.0 by changing version of helm installer to HelmInstaller@1 (previous @0 helm intaller is always using latest installed version from cache even if lower version number is specified).
And with helm 3.7.0 everything is working fine again.
**Log:**
...Unable to get an update from the "oci://redacted.azurecr.io/helm/xxx" chart repository:
	object required
Downloading accesspoint-ms from repo oci://redacted.azurecr.io/helm/xxx
Pulled: redacted.azurecr.io/helm/xxx/yyyy:0.1.34
[...]

PR #10628, opened 2022-01-30
Avoid querying for OCI tags can explicit version provided in chart dependencies

sabre1041:
Some OCI registries do not support listing tags as an available endpoint. When specifying a chart located in one of these registries, failures occur.

This enhancement aligns the functionality where tag listing is only queried whenever either no version is specified or a semver range is provided. 

Related to #10624

### DIFF
--- a/internal/resolver/resolver.go
+++ b/internal/resolver/resolver.go
@@ -135,22 +135,36 @@ func (r *Resolver) Resolve(reqs []*chart.Dependency, repoNames map[string]string
 			found = false
 		} else {
 			version = d.Version
-			// Retrieve list of tags for repository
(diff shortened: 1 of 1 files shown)


jdolitsky (approved):
👍

scottrigby (approved):
Good catch – this looks correct. For anyone else reviewing, it's the same logic we added to `(*ChartDownloader).getOciURI`:

https://github.com/helm/helm/blob/12f1bc0acdeb675a8c50a78462ed3917fb7b2e37/pkg/downloader/chart_downloader.go#L148-L153

Can not test at this moment, but approving based on code review only. Would someone please help test this? I pinged the OP and commenter of https://github.com/helm/helm/issues/10624 in Slack https://kubernetes.slack.com/archives/C51E88VDG/p1644516855721679 in case they have bandwidth to test before someone else does.

sennerholm:
What kind of tests do you need to be runned?

PR #10630, opened 2022-01-31
Make ParseReference visible outside of registry package.

### LINKED ISSUE #10629: Make parseReference function visible outside of the registry package
Output of `helm version`: version.BuildInfo{Version:"v3.8.0", GitCommit:"d14138609b01886f544b2025f5000351c9eb092e", GitTreeState:"clean", GoVersion:"go1.17.5"}

When using helm as a library it's useful to call [parseReference](https://github.com/helm/helm/blob/main/pkg/registry/util.go#L112) function and thus make it visible outside of the registry package.

revit13:
Closes #10629

### DIFF
--- a/pkg/registry/client.go
+++ b/pkg/registry/client.go
@@ -247,7 +247,7 @@ type (
 
 // Pull downloads a chart from a registry
 func (c *Client) Pull(ref string, options ...PullOption) (*PullResult, error) {
-	parsedRef, err := parseReference(ref)
+	parsedRef, err := ParseReference(ref)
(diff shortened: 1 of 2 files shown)


anthonyloupos:
Do we need this still now that I see Pull in client file? Can we close this out?

revit13:
Please see https://github.com/fybrik/fybrik/blob/master/pkg/helm/helm.go#L333 for use case.

joejulian:
closing as the issue has gone stale and the reason for this PR seems to be in error.

PR #10656, opened 2022-02-07
feat(plugin): Allow plugins to receive empty arg

### LINKED ISSUE #10654: Cannot pass empty argument to plugin
Empty arguments are mistakenly swallowed by the plugin parsing logic.

This is not a big problem as empty arguments are probably quite rare. However, the shell completion logic does have a use case for an empty argument, and there may be other situations where a plugin needs to be passed empty arguments.

Plugins using Cobra have a `__complete` command that can be called to obtain the valid completion choices.  That command takes as arguments what the user has typed.  When the user has not typed anything the command needs to be passed an explicit empty argument.  For example:
```
  helm 2to3 __complete ''
```
However, because empty arguments are being swallowed when helm parses them, helm ends up calling the `2to3` plugin without the empty argument.

To compare the behaviour one can do:
```
# Call the 2to3 plugin directly with an empty argument
> $(helm env HELM_PLUGINS)/helm-2to3/bin/2to3 __complete ''
cleanup	cleanup Helm v2 configuration, release data and Tiller deployment
convert	migrate Helm v2 release in-place to Helm v3
help	Help about any command
move	migrate Helm v2 configuration in-place to Helm v3
:4
Completion ended with directive: ShellCompDirectiveNoFileComp

# Make a similar call through helm and notice the missing argument error
> helm 2to3 __complete ''
Error: requires at least 1 arg(s), only received 0
Error: plugin "2to3" exited with error
```
which causes a missing argument error because the `__complete` command requires at least one argument.

Output of `helm version`:

This currently happens on the main branch.

I'll be posting a PR soon.

marckhouzam:

Fixes #10654

This small fix allows empty arguments to be properly passed to plugins.

To compare before and after:
```
> helm 2to3 __complete ''
Error: requires at least 1 arg(s), only received 0
Error: plugin "2to3" exited with error

> bin/helm 2to3 __complete '' 
cleanup	cleanup Helm v2 configuration, release data and Tiller deployment
convert	migrate Helm v2 release in-place to Helm v3
help	Help about any command
move	migrate Helm v2 configuration in-place to Helm v3
:4
Completion ended with directive: ShellCompDirectiveNoFileComp
```

**If applicable**:

### DIFF
--- a/cmd/helm/load_plugins.go
+++ b/cmd/helm/load_plugins.go
@@ -170,7 +170,9 @@ func manuallyProcessArgs(args []string) ([]string, []string) {
 				return v
 			}
 		}
-		return ""
+		// Return something different than what we received by adding a random prefix.
+		// We cannot simply return an empty string as v could be an empty string itself.
+		return fmt.Sprintf("x%s", v)
 	}
 
 	for i := 0; i < len(args); i++ {
--- a/cmd/helm/plugin_test.go
+++ b/cmd/helm/plugin_test.go
@@ -116,6 +116,8 @@ func TestLoadPlugins(t *testing.T) {
 		{"env", "env stuff", "show the env", "env\n", []string{}, 0},
 		{"exitwith", "exitwith code", "This exits with the specified exit code", "", []string{"2"}, 2},
 		{"fullenv", "show env vars", "show all env vars", envs + "\n", []string{}, 0},
+		// Test that empty arguments are passed to the plugin
+		{"numargs", "echo num args", "echo number of arguments received", "3\n", []string{"-a", "", "-b"}, 0},
 	}
 
 	plugins := cmd.Commands()
@@ -196,6 +198,7 @@ func TestLoadPluginsForCompletion(t *testing.T) {
 				{"more", []string{"one", "two"}, []string{"b", "ball"}, []staticCompletionDetails{}},
 			}},
 		}},
+		{"numargs", []string{}, []string{}, []staticCompletionDetails{}},
 	}
 	checkCommand(t, cmd.Commands(), tests)
 }
--- a/cmd/helm/testdata/helmhome/helm/plugins/numargs/numargs.sh
+++ b/cmd/helm/testdata/helmhome/helm/plugins/numargs/numargs.sh
@@ -1,0 +1,2 @@
+#!/bin/bash
+echo $#
--- a/cmd/helm/testdata/helmhome/helm/plugins/numargs/plugin.yaml
+++ b/cmd/helm/testdata/helmhome/helm/plugins/numargs/plugin.yaml
@@ -1,0 +1,4 @@
+name: numargs
+usage: "echo num args"
+description: "echo number of arguments received"
+command: "$HELM_PLUGIN_DIR/numargs.sh"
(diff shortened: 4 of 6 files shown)


marckhouzam:
Anyone have time to look at this fix?  It is a one-line change + tests.
It is a subtle bug but I hope the description of the PR and the corresponding issue #10654 explain things clearly.

17sushmita:
Hi @marckhouzam, I would like to work on this

lucasrod16 (approved):
Note: I am a community reviewer, not a maintainer. I was able to reproduce this error/bug by installing the helm `2to3` plugin and running the `helm 2to3 __complete ''` command as noted in the [issue](https://github.com/helm/helm/issues/10654).

After checking out the changes locally on this forked branch, I built the `helm` binary by running `make build`, which output the `helm` binary to `bin/helm`.

I then executed `bin/helm 2to3 __complete ''` to test the changes in this PR.

The output:

```
./bin/helm 2to3 __complete ''
cleanup cleanup Helm v2 configuration, release data and Tiller deployment
completion      Generate the autocompletion script for the specified shell
convert migrate Helm v2 release in-place to Helm v3
help    Help about any command
move    migrate Helm v2 configuration in-place to Helm v3
:4
Completion ended with directive: ShellCompDirectiveNoFileComp
```
There are unit tests and all workflows are passing. LGTM

cmd/helm/load_plugins.go
```diff
@@ -170,7 +170,9 @@ func manuallyProcessArgs(args []string) ([]string, []string) {
 				return v
 			}
 		}
-		return ""
+		// Return something different than what we received by adding a random prefix.
+		// We cannot simply return an empty string as v could be an empty string itself.
+		return fmt.Sprintf("x%s", v)
```
mattfarina, line 175: I don't think this is the right place to make the change. I think there are some unintended consequences though I've not worked out a better way to solve this problem.

`isKnown` is used to see if something is part of the above `kargs`. In the switch statement below this triggers some logic. With this change to `isKnown` the default section of the `switch` statement below will never be executed. All arguments will be labeled as known and unknown will always be empty.

mattfarina (changes requested):

cmd/helm/load_plugins.go
marckhouzam: Thanks @mattfarina for reviewing this.

It's been a long time but looking at the switch statement, I read the second `case` as equivalent to `if a == isKnown(a)`.  Or am I wrong?

If that is the case, then the requirement is that `isKnown()` return anything different than `a` if `a` is not part of `kvargs`, so returning `a` with a prefix of `x` satisfies this requirement, no?

But I might be misreading something.

tleed5: Pre-empting this comment with I am very new to this code base and this can be treated as me thinking out loud. 

I did some testing by debugging the `TestManuallyProcessArgs` test and I believe this assumption `if a == isKnown(a)` is correct, strings like `""` and `" "` or even `"x"` end up in the unknown arg list. 

That said I do think the switch statement is limiting in this scenario since it causes you to handle edge cases by modifying `IsKnown` rather than allowing you to add an explicit check for this scenario. Converting the switch statement into if statements  would allow you to avoid the need to add a prefix to the arg just so it ends up in the unknown arg list.

PR #10682, opened 2022-02-17
feat: supports complex dependency condition like go tags build expr

### LINKED ISSUE #10272: Conditional `OR` dependency logic possible ?
Team, 
we wish to control which sub-charts need to be installed among the many sub-charts in the umbrella charts, without any major changes in the whole chart structure. For instance, in this example, we have frontend, backend and database sub-charts. Let's say we want to only update backend and database, without disturbing frontend. We can achieve this in helm using conditions  as below Chart.yaml
dependencies:
  - name: frontend
    version: ~1.1.0
    repository: https://myhelmrepo
  - name: backend
    version: ^1.1.0
    repository: https://myanotherhelmrepo
    condition: backend.enabled, backend.my_condition

This condition  is AND logic , is there any way to use logical OR without using tags ?
cc @rimusz

Output of `helm version`: v3.7.1

Output of `kubectl version`: v1.18.20-eks-8c579e

anurnomeru:
Supports complex dependency condition like go tags build expr

```
' ' for OR, ',' for AND, '!' prefix for neg
```

closes #10272 

Adapted from git@github.com:golang/go.git src/go/build/constraint/expr.go:394

In order to be compatible with the previous design, this is a bit different from the tag resolve logic of go.

' ' for OR, ',' for AND, '!' prefix for neg

If an expr is not specified, it is not false, but null, that is, it will be ignored

```
!cond1,cond2 cond3
cond1 = true, cond2 = true, cond3 = true
Usually means (false && true || true) -> true
```

but if
```
cond1 = false, cond2 = null, cond3 = null
the result is (true && null || null) -> true
but not (true && false || false) -> false
```

**Special notes for backward compatibility**

Because two new keywords '!' ' ' are added, compatibility problems will occur if you use a condition starting with `!`, or a condition key containing ` ` spaces.

**If applicable**:


anurnomeru:
Would anyone be willing to help review this PR or offer any sugguestions?

anurnomeru:
> Would anyone be willing to help review this PR or offer any sugguestions?

For exmaple, enclosing the entire statement with special symbols to avoid bronken change.

PR #10693, opened 2022-02-22
feat: add --set-json flag to set json values.

### LINKED ISSUE #10428: Passing json objects/arrays as values through --set
I would like to set  json objects/arrays as values through --set.
According to https://helm.sh/docs/intro/using_helm/#the-format-and-limitations-of---set
currently that is not possible (with the exception of arrays of scalar types) 

As example, let's  take the following excerpt from my values.yaml file:

```
spec:
  ....
  sNssaiUpfInfoList:
  - sNssai:
      sd: 0002f0
      sst: 1
    dnnUpfInfoList:
    - dnn: intranet
      ipv4AddressRanges:
      - end: 9.9.9.94
        start: 9.9.9.65
  ....
```

In the template there is all the "range" based logic for slices and maps, to correctly render sNssaiUpfInfoList from .Values.spec.sNssaiUpfInfoList, so further elements could be added to each of the above arrays and they would appear in the final manifest generated by helm.

In order to set sNssaiUpfInfoList above, currently I have to set the individual leaf attributes of the data structure, by providing the following with --set  on command line:

`'spec.sNssaiUpfInfoList[0].sNssai.sd=0002f0,spec.sNssaiUpfInfoList[0].sNssai.sst=1,spec.sNssaiUpfInfoList[0].dnnUpfInfoList[0].dnn=intranet, spec.sNssaiUpfInfoList[0].dnnUpfInfoList[0].ipv4AddressRanges[0].end=9.9.9.94, spec.sNssaiUpfInfoList[0].dnnUpfInfoList[0].ipv4AddressRanges[0].start=9.9.9.65'
`

This is not very flexible because extra logic is required to translate the data structure into individual key=val for each of its leaf attributes. Moreover, the arrays are dynamic, and quite a long list of key=val could be generated.

I think it would be much simpler if we had an option to set a complete json array/object on the command line.  
For example, to set sNssaiUpfInfoList above, we could use something like:

`--set-json  'spec.sNssaiUpfInfoList=[{"sNssai":{"sd":"0002f0","sst":1},"dnnUpfInfoList":[{"dnn":"intranet","ipv4AddressRanges":[{"end":"9.9.9.94","start":"9.9.9.65"}]}]}]'
`

Is it possible to add this option to helm?

lucadirocco:
When used with helm install, helm template, helm upgrade, it enables
to set json values (scalars/objects/arrays) from the command line.

Closes #10428

Signed-off-by: Luca Di Rocco <lucadirocco@gmail.com>

This PR allows to set any key of the exposed values (scalars, objects, arrays) of a chart, by using a json value on the command line.
It become easier and more concise to update entire portions of the exposed values, as json data structures on command line, rather than setting the value of each individual leaf attribute of the data structure (#10428 provides an example) 

The implementation has tried to reuse as much as possible the existing code and the standard libraries, and is quite straightforward.
There are some obvious changes in cmd/helm/flags.go and pkg/cli/values/options.go in order to include the new option flag (--set-json)
The parser logic from pkg/strvals/parser.go is reused, with an extra field in the parser struct (isjsonval bool, true only for --set-json) and some extra code in the following two functions, to parse a json value when a "=" char is met:
func (t *parser) key(data map[string]interface{}) (reterr error)
func (t *parser) listItem(list []interface{}, i int) ([]interface{}, error) 

**If applicable**:
        Little updates in the description constants of cmd/helm/install.go and cmd/helm/upgrade.go to.
        Together with the info from cmd/helm/flags.go, this should update the documentation when "helm docs --type markdown --generate-headers" is executed.
        Some extra unit tests with function TestParseJSON in pkg/strvals/parser_test.go
        All unit test run, I don't know if something else is required. Being this a new feature, there are no changes in existing functionalities.
        Please note that I had also to remove the ending new line '\n' from an errors.Errorf statement in the  
func compare(actual []byte, filename string) error, in file ./internal/test/test.go, because go lint was failing and also tests were not running.

### DIFF
--- a/pkg/strvals/parser_test.go
+++ b/pkg/strvals/parser_test.go
@@ -567,6 +567,107 @@ func TestParseIntoString(t *testing.T) {
 	}
 }
 
+func TestParseJSON(t *testing.T) {
+	tests := []struct {
+		input  string
+		got    map[string]interface{}
+		expect map[string]interface{}
+		err    bool
+	}{
+		{ // set json scalars values, and replace one existing key
+			input: "outer.inner1=\"1\",outer.inner3=3,outer.inner4=true,outer.inner5=\"true\"",
+			got: map[string]interface{}{
+				"outer": map[string]interface{}{
+					"inner1": "overwrite",
+					"inner2": "value2",
+				},
+			},
+			expect: map[string]interface{}{
+				"outer": map[string]interface{}{
+					"inner1": "1",
+					"inner2": "value2",
+					"inner3": 3,
+					"inner4": true,
+					"inner5": "true",
+				},
+			},
+			err: false,
+		},
+		{ // set json objects and arrays, and replace one existing key
+			input: "outer.inner1={\"a\":\"1\",\"b\":2,\"c\":[1,2,3]},outer.inner3=[\"new value 1\",\"new value 2\"],outer.inner4={\"aa\":\"1\",\"bb\":2,\"cc\":[1,2,3]},outer.inner5=[{\"A\":\"1\",\"B\":2,\"C\":[1,2,3]}]",
+			got: map[string]interface{}{
+				"outer": map[string]interface{}{
+					"inner1": map[string]interface{}{
+						"x": "overwrite",
+					},
+					"inner2": "value2",
+					"inner3": []interface{}{
+						"overwrite",
+					},
+				},
+			},
+			expect: map[string]interface{}{
+				"outer": map[string]interface{}{
+					"inner1": map[string]interface{}{"a": "1", "b": 2, "c": []interface{}{1, 2, 3}},
+					"inner2": "value2",
+					"inner3": []interface{}{"new value 1", "new value 2"},
+					"inner4": map[string]interface{}{"aa": "1", "bb": 2, "cc": []interface{}{1, 2, 3}},
+					"inner5": []interface{}{map[string]interface{}{"A": "1", "B": 2, "C": []interface{}{1, 2, 3}}},
+				},
+			},
+			err: false,
+		},
+		{ // null assigment, and no value assigned (equivalent to null)
+			input: "outer.inner1=,outer.inner3={\"aa\":\"1\",\"bb\":2,\"cc\":[1,2,3]},outer.inner3.cc[1]=null",
+			got: map[string]interface{}{
+				"outer": map[string]interface{}{
+					"inner1": map[string]interface{}{
+						"x": "overwrite",
+					},
+					"inner2": "value2",
+				},
+			},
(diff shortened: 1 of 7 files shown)


pkg/strvals/parser_test.go
```diff
@@ -567,6 +567,107 @@ func TestParseIntoString(t *testing.T) {
...
 }
 
+func TestParseJSON(t *testing.T) {
+	tests := []struct {
+		input  string
+		got    map[string]interface{}
+		expect map[string]interface{}
+		err    bool
+	}{
+		{ // set json scalars values, and replace one existing key
```
joejulian, line 577: What happens if the flag is specified twice with conflicting values or subsets of values?
```
--setjson='foo={"key1":"value1","key2":"value2"}' --setjson='foo.key2="bar"'
```
or 
```
--setjson='foo=["one", "two", "three"]' --setjson='foo=["four"]'
```

lucadirocco: Sure. I can add some more info/examples to install.go and upgrade.go.

lucadirocco: @joejulian 
I have checked that  in cmd/helm/install.go  (const installDesc) it is already explained the behavior of the --set flag when it is specified multiple times.
I have added the above examples in order to highlight a similar behavior with --set-json:

```
You can specify the ‘–set’ flag multiple times. The priority will be given to the last (right-most) set specified. For example, if both ‘bar’ and ‘newbar’ values are set for a key called ‘foo’, the ‘newbar’ value would take precedence:

$ helm install --set foo=bar --set foo=newbar  myredis ./redis

Similarly, in the following example ‘foo’ is set to ‘[“four”]’:

$ helm install --set-json='foo=["one", "two", "three"]' --set-json='foo=["four"]' myredis ./redis

And in the following example, ‘foo’ is set to ‘{“key1”:“value1”,“key2”:“bar”}’:

$ helm install --set-json='foo={"key1":"value1","key2":"value2"}' --set-json='foo.key2="bar"' myredis ./redis

```

joejulian (approved):
LGTM, but I'm not an approver.

lucadirocco:
> LGTM, but I'm not an approver.

@joejulian 
Am I required to take any action (e.g. look for an approver)?
It's unclear to me what next a steps should be.
Can you please clarify?

hickeyma (approved):
LGTM, thanks @lucadirocco

lucadirocco:
> LGTM, thanks @lucadirocco

Thank you @hickeyma

hickeyma:
This need approval from another maintainer to merge.

sabre1041 (approved):
LGTM

PR #10695, opened 2022-02-22
Print downloaded chart path.

### LINKED ISSUE #10707: No visual feedback when downloading Chart
`helm pull` has no console output when it is run successfully which can be confusing.

danthelion:
Closes https://github.com/helm/helm/issues/10707

### DIFF
--- a/pkg/downloader/chart_downloader.go
+++ b/pkg/downloader/chart_downloader.go
@@ -111,7 +111,7 @@ func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *proven
 
 	destfile := filepath.Join(dest, name)
 	if err := fileutil.AtomicWriteFile(destfile, data, 0644); err != nil {
-		fmt.Fprintf(&out, "Chart downloaded to: %s\n", destfile)
+		fmt.Fprintf(c.Out, "Chart downloaded to: %s\n", destfile)
 		return destfile, nil, err
 	}
 
--- a/pkg/downloader/chart_downloader.go
+++ b/pkg/downloader/chart_downloader.go
@@ -86,6 +86,8 @@ type ChartDownloader struct {
 // Returns a string path to the location where the file was downloaded and a verification
 // (if provenance was verified), or an error if something bad happened.
 func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *provenance.Verification, error) {
+	var out strings.Builder
+
 	u, err := c.ResolveChartVersion(ref, version)
 	if err != nil {
 		return "", nil, err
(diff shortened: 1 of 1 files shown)


yxxhero:
@danthelion 
1.Please create a issue for this PR. Thanks very much.
2.fix the DCO in the ci.

pkg/downloader/chart_downloader.go
```diff
@@ -86,6 +86,8 @@ type ChartDownloader struct {
 // Returns a string path to the location where the file was downloaded and a verification
 // (if provenance was verified), or an error if something bad happened.
 func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *provenance.Verification, error) {
+	var out strings.Builder
+
```
joejulian, line 90:
```suggestion
```

pkg/downloader/chart_downloader.go
```diff
@@ -109,6 +111,7 @@ func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *proven
 
 	destfile := filepath.Join(dest, name)
 	if err := fileutil.AtomicWriteFile(destfile, data, 0644); err != nil {
+		fmt.Fprintf(&out, "Chart downloaded to: %s\n", destfile)
```
joejulian, line 114: Shouldn't this use ChartLoader.Out to be consistent with the rest of the code?

```suggestion
		fmt.Fprintf(c.Out, "Chart downloaded to: %s\n", destfile)
```

joejulian (changes requested):

danthelion:
Huh I'll guess accepting suggestions from the iOS GitHub app won't suffice, I'll fix the commita when I'm back near my computer @joejulian @yxxhero

joejulian (approved):
lgtm, but I'm not an approver

bacongobbler:
I am concerned about this being put in DownloadTo. Perhaps the calling function requested the package to be downloaded to a temporary path before performing additional operations, in which case the output path is totally useless to the user.

In general we reserve `c.Out` to be used more as a diagnosis/debugging logger than standard output IIRC.

`helm pull` is [supposed to display the output path](https://github.com/helm/helm/blob/65d8e72504652e624948f74acbba71c51ac2e342/cmd/helm/pull.go#L68-L72). Perhaps there's an underlying bug preventing that from happening? Hvae you looked closer at the existing infrastructure to confirm whether something else is going on?

bacongobbler:
ping.

danthelion:
Hmm you are right, it should print the output but it doesn't. I'll look into it! Sorry for being slow about it, hectic times.

joejulian:
Ah, sorry, I just noticed that everything else used c.Out so I thought that was the "right way".

yxxhero:
ping

joejulian:
closing this as abandoned

PR #10715, opened 2022-03-02
Add pagination support while listing Kubernetes Secrets

### LINKED ISSUE #7997: Errors on "helm list" AND "helm install"
**Output of `helm version`:**

**Output of `kubectl version`:**
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.2"}
Server Version: version.Info{Major:"1", Minor:"15+", GitVersion:"v1.15.11-gke.9"}

**Cloud Provider/Platform (AKS, GKE, Minikube etc.):** 
GKE

-----

On Helm list I get

**Error: list: failed to list: stream error when reading response body, may be caused by closed connection. Please retry. Original error: stream error: stream ID 3; INTERNAL_ERROR**

On Helm install chart I get:

**request.go:924] Unexpected error when reading response body: net/http: request canceled (Client.Timeout exceeded while reading body)
Error: unable to build kubernetes objects from release manifest: unexpected error when reading response body. Please retry. Original error: net/http: request canceled (Client.Timeout exceeded while reading body**

"Helm delete" is working. Was able to uninstall a release

---

Additional notes:

- there is network activity on "helm list" over one minute or so. (maybe timeout trigered?)

- The setup was running for months without any problems

- I did the update to Helm v3.1.2 in the this current debugging process for this issue

- There was a Nodeupdate on the Kubernetes side recently - (maybe relevant)

- Created also a new Cluster on GKE for testing and there "Helm list" .."Helm install" are working.

krish7919:
This commit adds support for pagination while listing Kubernetes
Secrets.
This is useful in clusters with a large number of releases or ones with
a large number of release versions. Since each version of a Helm release
is stored in Kubernetes Secret, the Secrets increase drastically,
causing timeouts and server errors while listing them.
The fix uses pagination to `300` releases at a time, causing the server
to never hit the timeout limit or other errors.

Closes https://github.com/helm/helm/issues/7997

**If applicable**:

EDIT:

More details and before/after tests are at https://github.com/helm/helm/issues/7997#issuecomment-1057294371

### DIFF
--- a/pkg/storage/driver/secrets.go
+++ b/pkg/storage/driver/secrets.go
@@ -35,8 +35,12 @@ import (
 
 var _ Driver = (*Secrets)(nil)
 
-// SecretsDriverName is the string name of the driver.
-const SecretsDriverName = "Secret"
+const (
+	// SecretsDriverName is the string name of the driver.
+	SecretsDriverName = "Secret"
+	// ListPaginationLimit is the number of Secrets we fetch in a single API call.
+	ListPaginationLimit = int64(300)
+)
 
 // Secrets is a wrapper around an implementation of a kubernetes
 // SecretsInterface.
@@ -78,14 +82,35 @@ func (secrets *Secrets) Get(key string) (*rspb.Release, error) {
 // List fetches all releases and returns the list releases such
 // that filter(release) == true. An error is returned if the
 // secret fails to retrieve the releases.
+// We read `ListPaginationLimit` Secrets at a time so as not to overwhelm the
+// `api-server` in a cluster with many releases; fixes
+// https://github.com/helm/helm/issues/7997
 func (secrets *Secrets) List(filter func(*rspb.Release) bool) ([]*rspb.Release, error) {
 	lsel := kblabels.Set{"owner": "helm"}.AsSelector()
-	opts := metav1.ListOptions{LabelSelector: lsel.String()}
+	opts := metav1.ListOptions{LabelSelector: lsel.String(), Limit: ListPaginationLimit}
 
+	// Perform an initial list
 	list, err := secrets.impl.List(context.Background(), opts)
 	if err != nil {
 		return nil, errors.Wrap(err, "list: failed to list")
 	}
+
+	// Fetch more results from the server by making recursive paginated calls
+	isContinue := list.Continue
+	for isContinue != "" {
+		secrets.Log("list: fetched %d secrets, more to fetch..\n", ListPaginationLimit)
+		opts = metav1.ListOptions{LabelSelector: lsel.String(), Limit: ListPaginationLimit, Continue: isContinue}
+		batch, err := secrets.impl.List(context.Background(), opts)
+		if err != nil {
+			return nil, errors.Wrap(err, "list: failed to perform paginated listing")
+		}
+
+		// Append the results to the initial list
+		list.Items = append(list.Items, batch.Items...)
+
+		isContinue = batch.Continue
+	}
+	secrets.Log("list: fetched %d releases\n", len(list.Items))
 
 	var results []*rspb.Release
 


krish7919:
Bump? Who can get this reviewed and merged?

krish7919:
Bump??

joejulian:
I'm still not sure that this will have any effect. The problem still comes from requesting _any_ secret from an api-server that does not have secrets cached. When the api-server fetches the secrets from etcd, it fetches all of them. If any addional requests for secrets come in while that query is still processing, they will also be treated as cache misses and the api-server will fetch _all_ the secrets from etcd. This compounding of requests causes a pile-up and will result in timeouts. Those timeouts will cause retries which will hit cache misses, etc., compounding the problem.

Pagenation only happens on the client side of the api-server, not the storage side.

krish7919:
> The problem still comes from requesting any secret from an api-server that does not have secrets cached. 
> ...
>  If any addional requests for secrets come in while that query is still processing, they will also be treated as cache misses and the api-server will fetch all the secrets from etcd. 

The api-server is not the problem here, especially in managed kubernetes environments - they scale automatically based on number of requests.

> This compounding of requests causes a pile-up and will result in timeouts. T

Even if the situaltion you describe happens, the timeout **will** be reduced by specifying a smaller number in the list option.

> Pagenation only happens on the client side of the api-server, not the storage side.

And this still solves the problem; have you had a chance to review the code and the tests?

joejulian:
Ah, found part of my quote from an old issue. Trying to find where we discussed it in sig-api-machinery:
> We've had users that, through automation, have created a couple hundred 1k release secrets. A couple applications in the cluster poll secrets (grumble) and when those polls happened to stack up, and the api server couldn't serve the request from cache, the list queries are individually sent to etcd which cannot filter the lookup. Etcd can only respond with all the secrets in the namespace (or all secrets if the query is not namespaced). These queries can take some time to read and confirm consensus. When the etcd responses start timing out, this perfect storm starts to happen and the control-plane ends up in a deadlock.

krish7919:
@joejulian I do not see the relevance here. We are talking 2 different problems here.

1. Pagination: Helm doesn't use one, and the MR adds it. We are at ~4k secrets and the solutions works in real-world, production environment.
2. After pagination, we might still fail some API calls due to other scale issues, design flaws, etc. This is not what is solved here.

joejulian:
I'm just trying to help. We traced the root cause of the same issue you're reporting in my $dayjob a little over a year ago. We even filed a cve report (it was rejected as a cve). I can bring down a control plane with only 200 1MB resources whether they're Secrets, ConfigMaps, I've even accomplished it with Pods, and the right sequence of api requests.

I'm not against this PR at all. I'm just trying to help you with our RCA to (hopefully) save you some strife.

krish7919:
I did not mean to sound confrontational. Just pointing out that this is a very different problem.

The way I think of it, the repeated API requests issue can only be solved by listing and caching the Helm releases in a Tiller-like service (which will effectively be a Operator keeping a watch on Helm release Secrets). And since we are on Tiller-less Helm 3, this is the best way it would work. Just trying to work out all the scenarios.

Your comments are very helpful either way as it helps me think and prempt any more problems in our production 👍

pkg/storage/driver/secrets.go
```diff
@@ -35,8 +35,12 @@ import (
 
 var _ Driver = (*Secrets)(nil)
 
-// SecretsDriverName is the string name of the driver.
-const SecretsDriverName = "Secret"
+const (
+	// SecretsDriverName is the string name of the driver.
+	SecretsDriverName = "Secret"
+	// ListPaginationLimit is the number of Secrets we fetch in a single API call.
+	ListPaginationLimit = int64(300)
```
bacongobbler, line 42: Is this something we should make configurable for commands using this API, such as `helm list`? Alternatively, should this be a field under the secrets driver, defaulting to this value if none is provided in the constructor?

krish7919: Wow, I just checked my emails to see that this is still open! Its a pity that a PR opened on March 22 was only reviewed on May 6, and I had moved on from this work since then. _And the related issue has been open since 2020!_

What's the guarantee that if I/anyone carve out time for this then the reviewers/maintainers would be willing to move faster here?

joejulian: There are no guarantees. Folks contribute their time as it becomes available. Nobody's getting paid to work on Helm full time. We all just do our best to do what we can in life.

miaoxiaoy:
Helm's arrogant attitude is truly speechless. Helm is doomed.

PR #10740, opened 2022-03-08
feat: add pdb to chart create util

naseemkullah:

Add a `pdb.yaml` template  in `helm create` to ease the use of [pod disruption budgets](https://kubernetes.io/docs/concepts/workloads/pods/disruptions/#pod-disruption-budgets) when bootstrapping a newchart.

**If applicable**:


naseemkullah:
PTAL @unguiculus @scottrigby @rimusz @hickeyma

naseemkullah:
@mattfarina @hickeyma PTAL

pkg/action/show.go
```diff
@@ -113,7 +113,7 @@ func (s *Show) Run(chartpath string) (string, error) {
 			printer.Execute(&out, s.chart.Values)
 		} else {
 			for _, f := range s.chart.Raw {
-				if f.Name == chartutil.ValuesfileName {
+				if f.Name == chartutil.ValuesFilename {
```
mattfarina, line 116: Why are there change of things like `ValuesfileName` to `ValuesFilename` and 'ChartfileName` to `ChartFilename`? These seem unrelated and get into the semantics of how we talk about things internally. Might I suggest reverting those changes.

pkg/chartutil/create.go
```diff
@@ -39,36 +39,22 @@ import (
...
+	IgnoreFilename                  = ".helmignore"
+	IngressFilename                 = TemplatesDir + sep + "ingress.yaml"
+	DeploymentFilename              = TemplatesDir + sep + "deployment.yaml"
+	ServiceFilename                 = TemplatesDir + sep + "service.yaml"
+	ServiceAccountFilename          = TemplatesDir + sep + "serviceaccount.yaml"
+	HorizontalPodAutoscalerFilename = TemplatesDir + sep + "hpa.yaml"
+	PodDisruptionBudgetFilename     = TemplatesDir + sep + "pdb.yaml"
+	NotesFilename                   = TemplatesDir + sep + "NOTES.txt"
+	HelpersFilename                 = TemplatesDir + sep + "_helpers.tpl"
+	TestConnectionFilename          = TemplatesTestsDir + sep + "test-connection.yaml"
```
mattfarina, line 57: All of these had comments on them describing what they do. Some of that content matters. For example, `TemplateDir` is the _relative_ directory name. The relative nature is important. Can you revert the removing those.

mattfarina (changes requested):

pkg/chartutil/create.go
naseemkullah: ok, but then shoudnt we just make the name self descriptive?

`TemplatesRelativeDir` ? Or is that a go anti pattern of sorts

pkg/action/show.go
naseemkullah: Because filename is one word 

ref.: https://en.wikipedia.org/wiki/Filename

It was done in a separate commit

naseemkullah: i guess can be either, ill revert 
image

pkg/chartutil/create.go
naseemkullah: reverted, lets keep comments everywhere, dont tell uncle bob

naseemkullah:
flaky ci
image

Issue #10734, opened 2022-03-07
Inconsistent helm repo list behavior 


luinnar:
Command `helm repo list` act differently when repository config file don't exists:

```bash
$ helm repo list
Error: no repositories to show
$ helm repo list -o json
[]
$ echo $?
0

$ rm ~/.config/helm/repositories.*

$ helm repo list -o json
Error: no repositories to show
$ echo $?
1
```

In both cases helm should return return empty list of repositories without exiting with error code.

Output of `helm version`:
* version.BuildInfo{Version:"v3.7.2", GitCommit:"663a896f4a815053445eec4153677ddc24a0a361", GitTreeState:"clean", GoVersion:"go1.16.10"}
* version.BuildInfo{Version:"v3.8.0", GitCommit:"d14138609b01886f544b2025f5000351c9eb092e", GitTreeState:"clean", GoVersion:"go1.17.5"}

bacongobbler:
fixed in https://github.com/helm/helm/pull/10519.

PR #10745, opened 2022-03-09
Fix Inconsistent helm repo list behavior

yankay:

FIX The issue "Inconsistent helm repo list behavior" https://github.com/helm/helm/issues/10734, 

Signed-off-by: Kay Yan <kay.yan@daocloud.io>

### DIFF
--- a/cmd/helm/repo_list.go
+++ b/cmd/helm/repo_list.go
@@ -39,7 +39,7 @@ func newRepoListCmd(out io.Writer) *cobra.Command {
 		ValidArgsFunction: noCompletions,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			f, err := repo.LoadFile(settings.RepositoryConfig)
-			if isNotExist(err) || (len(f.Repositories) == 0 && !(outfmt == output.JSON || outfmt == output.YAML)) {
+			if (isNotExist(err) || len(f.Repositories) == 0) && !(outfmt == output.JSON || outfmt == output.YAML) {
 				return errors.New("no repositories to show")
 			}
 


cmd/helm/repo_list.go
```diff
@@ -39,7 +39,7 @@ func newRepoListCmd(out io.Writer) *cobra.Command {
 		ValidArgsFunction: noCompletions,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			f, err := repo.LoadFile(settings.RepositoryConfig)
-			if isNotExist(err) || (len(f.Repositories) == 0 && !(outfmt == output.JSON || outfmt == output.YAML)) {
+			if (isNotExist(err) || len(f.Repositories) == 0) && !(outfmt == output.JSON || outfmt == output.YAML) {
```
soenkeliebau, line 42: I think this code will have the desired effect, but I also think it is a bit complicated to understand.

The entire `err` variable is more or less redundant here, as `len(f.Repositories) == 0` will always also be true if `isNotExist(err)` is true, so this could be simplified by removing it from the condition. 

Probably comes down to preference, but I find the solution in https://github.com/helm/helm/pull/10519 easier to understand.

yankay: That's great. There are two solutions for one issue.The difference between the solution is not very big.
This PR keep the error, less code change. And https://github.com/helm/helm/pull/10519 is simpler.

lfrancke: For what it's worth: I'm with Sönke and I like the other solution better.

soenkeliebau: I'm not sure I completely follow your comment, but I'd personally be fine merging either of these PRs, I think they'll both solve the issue :)

yankay: Should we merge it.

soenkeliebau: fine for me

bacongobbler:
fixed in https://github.com/helm/helm/pull/10519.

PR #10787, opened 2022-03-18
 releaseutil: fix sort ordering

### LINKED ISSUE #10768: Hook manifests are applied ordered by name rather than by Kind
Hey there,

While recently working on one of our charts, our team (@alvarocabanas) discovered a weird behavior on how Helm manages the order in which hooks are applied into the cluster. We noticed this after seeing that a `Job` annotated as a `pre-install` hook was getting stuck due to the `ServiceAccount` it required (and was also annotated) not being there. After some trial and error, we figured that this condition only happened when the name of the `ServiceAccount` was such that it would appear later on an alphabetically sorted list that also contained the `Job` name.

Our understanding is that, as defined here, this should not happen: https://github.com/helm/helm/blob/8ab18f7567cedffdfa5ba4d7f6abfb58efc313f8/pkg/releaseutil/kind_sorter.go#L31

Moreover the docs [say](https://helm.sh/docs/topics/charts_hooks/):
> The library sorts hooks by weight (assigning a weight of 0 by default), by resource kind and finally by name in ascending order.

Which is definitely not the behavior we were seeing.

After some troubleshooting and messing with the debugger, we found out this to be the problematic function:

https://github.com/helm/helm/blob/8ab18f7567cedffdfa5ba4d7f6abfb58efc313f8/pkg/action/hooks.go#L41-L42

Which eventually calls:

https://github.com/helm/helm/blob/8ab18f7567cedffdfa5ba4d7f6abfb58efc313f8/pkg/action/hooks.go#L115-L120

Our understanding is that the code is not doing what the `// hooke are pre-ordered by kind, so keep order stable` comment hints it is supposed to be doing. The Less function defined above will actually _override_ the Kind order if the weight is equal, resulting in a list ordered by weight, name, and finally by kind.

Our team has prepared a very naive fix for this, which essentially gets rid of the name-based sort: https://github.com/helm/helm/compare/main...newrelic-forks:fix-hook-order

As I said, this fix is very naive and maybe not ready to be merged, but we will be happy to pursue this and contribute a proper fix with this, if you can confirm this issue and guide us in the right direction.

---

Output of `helm version`:

```
version.BuildInfo{Version:"v3.8+unreleased", GitCommit:"8ab18f7567cedffdfa5ba4d7f6abfb58efc313f8", GitTreeState:"dirty", GoVersion:"go1.17.8"}
```

Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.4", GitCommit:"e6c093d87ea4cbb530a7b2ae91e54c0842d8308a", GitTreeState:"archive", BuildDate:"2022-02-18T22:31:12Z", GoVersion:"go1.17.7", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.4", GitCommit:"e6c093d87ea4cbb530a7b2ae91e54c0842d8308a", GitTreeState:"clean", BuildDate:"2022-02-16T12:32:02Z", GoVersion:"go1.17.7", Compiler:"gc", Platform:"linux/amd64"}
```

```
Kubeadm
```

nadiamoe:

As discussed in https://github.com/helm/helm/issues/10768, prior to this PR when hooks were sorted by weight and name, kind order was overwritten. This PR modifies the render-time hook sorting function to do all of the sorting: By weight, kind, and finally name. With this change, the second sorting pass can be entirely removed and the sorting should be consistent.

Additionally, while working on this I identified that the `lessByKind` function had unused arguments and, more importantly, was converting the SortingOrder slice to a map on every call to `Less`, which I think is substantially inefficient. I've made a minor change to this function in a separate commit (https://github.com/helm/helm/commit/9d8e9f6b8da63078c97e0c50be1a9df8e77e1625).

This PR should make the sorting of hooks adhere to what the documentation says. However, there might be a slight change of behavior in `helm template`, which previously output hooks without weight ordering, and how it will respect it as well. We have done this to avoid sorting in two different places, but I'm open to changing the approach if this behavior is not wanted.

**If applicable**:

---

Fixes https://github.com/helm/helm/issues/10768

### DIFF
--- a/pkg/action/hooks.go
+++ b/pkg/action/hooks.go
@@ -17,11 +17,9 @@ package action
 
 import (
 	"bytes"
-	"sort"
 	"time"
 
 	"github.com/pkg/errors"
-
 	"helm.sh/helm/v3/pkg/release"
 	helmtime "helm.sh/helm/v3/pkg/time"
 )
@@ -37,9 +35,6 @@ func (cfg *Configuration) execHook(rl *release.Release, hook release.HookEvent, 
 			}
 		}
 	}
-
-	// hooke are pre-ordered by kind, so keep order stable
-	sort.Stable(hookByWeight(executingHooks))
 
 	for _, h := range executingHooks {
 		// Set default delete policy to before-hook-creation
@@ -105,18 +100,6 @@ func (cfg *Configuration) execHook(rl *release.Release, hook release.HookEvent, 
 	}
 
 	return nil
-}
-
-// hookByWeight is a sorter for hooks
-type hookByWeight []*release.Hook
-
-func (x hookByWeight) Len() int      { return len(x) }
-func (x hookByWeight) Swap(i, j int) { x[i], x[j] = x[j], x[i] }
-func (x hookByWeight) Less(i, j int) bool {
-	if x[i].Weight == x[j].Weight {
-		return x[i].Name < x[j].Name
-	}
-	return x[i].Weight < x[j].Weight
 }
 
 // deleteHookByPolicy deletes a hook if the hook policy instructs it to
--- a/pkg/action/hooks.go
+++ b/pkg/action/hooks.go
@@ -20,6 +20,7 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+
 	"helm.sh/helm/v3/pkg/release"
 	helmtime "helm.sh/helm/v3/pkg/time"
 )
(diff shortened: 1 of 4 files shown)


bacongobbler:
cc @jlegrone as I know he worked on hook ordering for a while.

kang-makes:
Any news about this?

Xitric:
@bacongobbler @jlegrone Any chance someone wants to take a look at this? We just spent the better part of a day trying to figure out why Helm wasn't behaving according to its own documentation when it comes to ordering of hook resources, and eventually we found our way here.

Issue #10312, opened 2021-11-04
What do you want to see in OCI Support?


jdolitsky:
Add comments here with what you would like to see in Helm's OCI support.

Please use the :+1: emoji to indicate a upvote for a particular feature, and :-1: to indicate a downvote.

jdolitsky:
Arbitrary tags (e.g. `latest`)

jdolitsky:
Specifying tag/version using Docker-like colon notation (`:1.2.3` vs. `--version 1.2.3`)

jdolitsky:
Digest verification via `--digest=sha256:...` or `:1.2.3@sha256:...` (depending on comment above)

jdolitsky:
Semantic version range support / automatically determining latest semantic version (see also https://github.com/helm/helm/issues/9694)

jdolitsky:
Support in `helm search` (not yet defined in OCI distribution-spec)

SteveLasker:
> 1. Support in `helm search` (not yet defined in OCI distribution-spec)

There's a few options here. The current [_tag listing API](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#content-discovery) can likely be used to discover the versions for a specific chart. 

The [_catalog API](https://github.com/opencontainers/distribution-spec/pull/69) was moved to reserved, as some, but not all registries were able to implement it.

If this API, conceptually works, we can explore a new API using the [extensions API](https://github.com/opencontainers/distribution-spec/pull/111)

Adding here in case someone would like to experiment how well these apis could enable server side search, avoiding the challenges with pulling an index.yaml file that could include 10's of thousands of charts from automated builds.

almson:
> Arbitrary tags (e.g. latest)

IMHO, Helm's insistence on semantic versioning is inappropriate for internal projects. Semantic versioning is most useful for public projects with many users. We cheat by using the pre-release tag (`1.0.0-<pre-release>`) as the version. It's constructed as git tag and hash (via git describe). The pre-release tag offers the flexibility we need, but being forced to prepend `1.0.0-` is redundant.

That said, it would be wrong for the repo tag to be different from the version. So I think either relax the version requirements or don't allow arbitrary tags.

jdolitsky:
@almson -

> That said, it would be wrong for the repo tag to be different from the version. So I think either relax the version requirements or don't allow arbitrary tags.

100% agree. This is something that needs to be addressed unrelated to OCI support.

mot256:
[...]

PR #10799, opened 2022-03-22
feat: get charts by specifying the SHA256 checksum as the version.

ewollesen:

This worked in version 3.7, but seems to have been lost or removed in
the upgrade to 3.8.

It is useful because not all repositories have immutable tags.

This is tangentially related to this comment:

https://github.com/helm/helm/issues/10312#issuecomment-961396434

**If applicable**:
- [X ] this PR contains unit tests

### DIFF
--- a/pkg/downloader/chart_downloader.go
+++ b/pkg/downloader/chart_downloader.go
@@ -21,6 +21,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/Masterminds/semver/v3"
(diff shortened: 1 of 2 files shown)


TerryHowe:
@mattfarina we count on installing helm charts by shasum for security reasons

jdolitsky (changes requested):
In the past we have had discussions forcing version to be semver.

IMO the `digest` field should be additive/separate from `version`, and image refs should end up in the form:
```
registry.example.com/myuser/mychart:0.1.0@sha256:...
```

pkg/downloader/chart_downloader.go
```diff
@@ -141,36 +142,41 @@ func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *proven
...
 func (c *ChartDownloader) getOciURI(ref, version string, u *url.URL) (*url.URL, error) {
-	var tag string
 	var err error
 
 	// Evaluate whether an explicit version has been provided. Otherwise, determine version to use
-	_, errSemVer := semver.NewVersion(version)
-	if errSemVer == nil {
-		tag = version
+	if sha256re.MatchString(version) {
+		u.Path = fmt.Sprintf("%s@%s", u.Path, version)
```
scottrigby, line 152: This lgtm in general. My only question is why all of the existing code needs to move around for this. Can we just move this check to the top and return if true, leaving the rest of the code as-is?

lewisdiamond: The rest of the code is effectively untouched. The large diff seems to be generated by the move of `var tag string` to the `else` scope as this is the only place where it is used. The diff matches the wrong `} else {`.

This wraps the existing logic with a condition `if sha256re...` that sets `u.Path` otherwise the previous logic is run unmodified and sets `u.Path` as it previously did.

lewisdiamond: Github seems to support it too. Click the cog at the top of the `Files Changed` tab and click `Hide whitespaces`

ewollesen:
> In the past we have had discussions forcing version to be semver.
> 
> IMO the `digest` field should be additive/separate from `version`, and image refs should end up in the form:
> 
> ```
> registry.example.com/myuser/mychart:0.1.0@sha256:...
> ```

I'll have to look into this more. The way I understand the code right now, the version is required when pulling from an OCI registry. It seems redundant to require a version when there's a checksum specified already. Assuming I can relax the version requirement when a checksum is present, this should probably be doable (though it will increase the contact points of the change).

tamalsaha:
Hello, I would appreciate if this feature/pr could be merged.

sabre1041:
@ewollesen Wanted to circle back and see whether you might have some availability coming up to look into this issue. No pressure whatsoever, just checking from a planning perspective

TerryHowe:
> @ewollesen Wanted to circle back and see whether you might have some availability coming up to look into this issue. No pressure whatsoever, just checking from a planning perspective

I'll revive this unless Eric is really wanting to work on this

sabre1041:
@TerryHowe is this something you can pick up? Like yourself, would be good to get this capability in a supported fashion

yxxhero:
@TerryHowe ping

TerryHowe:
Yeh, I got put on a project that was rather consuming, but I should specifically be freed up to work on projects like this very soon. Sorry, it has been rocky.

yxxhero:
@ewollesen good luck to you.

sabre1041:
@TerryHowe Any chance that you might be able to pick this up any time soon. 

#12558 looks to add support for specifying a tag and ideally we would want to be able to target either a sha256 or tag

PR #10804, opened 2022-03-25
feat: implement export-values

c0d1ngm0nk3y:
This PR implements export-values directive which is similar to the [import-values](https://helm.sh/docs/topics/charts/#importing-child-values-via-dependencies) directive, but instead of importing the values of the subchart into the current chart it exports the values of the current chart into subchart.

This implements the HIP https://github.com/helm/community/pull/242.

Initially the change was discussed here https://github.com/helm/helm/issues/3242. There were several attempts to implement this: https://github.com/helm/helm/pull/7477, https://github.com/helm/helm/pull/3243. This pr is based on #10059 and did some minor modification.
 
## Usage:
In the full form of export-values you specify which value tree to export (with parent:) and where it should be exported in the subchart (with child:).

In the short form of export-values you only specify a relative path inside of the exports: key of the parent chart values: specified values will be exported to the root of the subchart values (same as with the import-values). This allows to export/import values defined in exports: both to the parent chart or to the child chart depending on the directive used (import-values or export-values).

Unlike import-values, export-values can export not only the trees of values (maps), but other types of values too (strings, ints, etc).

Example:
```yaml
# values.yaml
parent-map:
  key1: parent-value1
  key2: parent-value2
  
exports:
  parent-map:
    key3: parent-value3
  
subchart1:
  subchart-map:
    key1: subchart-value1
```
```yaml
# Chart.yaml:
dependencies:
  - name: subchart1
    version: 1.0.0
    export-values:
    - parent: "parent-map"
      child: "subchart-map"
    - "parent-map.key3"
```
``` yaml
# charts/subchart1/dump-values-in-subchart.yaml
dump-values-in-subchart: {{ .Values | toYaml | nindent 2 }}
```

Result of helm template:

```yaml
# charts/subchart1/dump-values-in-subchart.yaml
dump-values-in-subchart:
  subchart-map:
    key1: parent-value1
    key2: parent-value2
    key3: parent-value3
```

## Open Points / Next Steps

* The combination of `import-values` and `export-values` does not work yet. This is not explicitly mentioned in the HIP, but could be assumed by the user.

Example: We want to import a value `port` from `subchartA` and export the very same value to `subchartB`.

This would probably require a change in the way the values are processed and would also affect the current `import-values` behavior. So we would see this independent of this PR. But it is a valid next step and would meet users' expectations better.

If applicable:

* [ ] this PR contains documentation
* [ ] this PR has been tested for backwards compatibility

### DIFF
--- a/pkg/chartutil/dependencies_test.go
+++ b/pkg/chartutil/dependencies_test.go
@@ -218,7 +218,7 @@ func TestProcessDependencyImportValues(t *testing.T) {
 	e["SCBexported2A"] = "blaster"
 	e["global.SC1exported2.all.SC1exported3"] = "SC1expstr"
 
-	if err := processDependencyImportValues(c, false); err != nil {
+	if err := processDependencyImportExportValues(c, false); err != nil {
 		t.Fatalf("processing import values dependencies %v", err)
 	}
 	cc := Values(c.Values)
@@ -258,7 +258,7 @@ func TestProcessDependencyImportValues(t *testing.T) {
 	}
 
 	c = loadChart(t, "testdata/subpop")
-	if err := processDependencyImportValues(c, true); err != nil {
+	if err := processDependencyImportExportValues(c, true); err != nil {
 		t.Fatalf("processing import values dependencies %v", err)
 	}
 	cc = Values(c.Values)
@@ -294,7 +294,7 @@ func TestProcessDependencyImportValuesMultiLevelPrecedence(t *testing.T) {
 	e["app2.service.port"] = "9090"
 	e["app3.service.port"] = "9090"
 	e["app4.service.port"] = "1234"
-	if err := processDependencyImportValues(c, true); err != nil {
+	if err := processDependencyImportExportValues(c, true); err != nil {
 		t.Fatalf("processing import values dependencies %v", err)
 	}
 	cc := Values(c.Values)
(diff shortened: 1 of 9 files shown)


pkg/chartutil/dependencies_test.go
```diff
@@ -239,11 +239,66 @@ func TestProcessDependencyImportValues(t *testing.T) {
...
+	expectation["subchart1.merged-values.SPNested1.SPExtra19"] = "exported-from-parent-6"
+
+	// `exports` style
+	expectation["subchart1.merged-values.SPExtra9"] = "exported-from-parent-4"
+	expectation["subchart1.merged-values.SPExtra12"] = "should-be-unchanged"
+
+	// passed from child to grandchild with overrides
+	expectation["subchart1.subcharta.exported-overridden-chart1.SCAbool"] = true
+	expectation["subchart1.subcharta.exported-overridden-chart1.SCAfloat"] = 33.3
+	expectation["subchart1.subcharta.exported-overridden-chart1.SCAint"] = float64(333)
```
c0d1ngm0nk3y, line 267: Why are all number values in helm `float64` and not `int`?

loewenstein:
@jdolitsky any chance to get a timely review on this?

joejulian:
@c0d1ngm0nk3y Can you rebase and squash this?

chenlujjj:
Any updates ?

loewenstein:
Not from me or @c0d1ngm0nk3y - this is ready for review and merge already for more than a while.

@mattfarina any updates?

scottrigby:
New features now need to be made for Helm 4, so this will need to be rebased against `main`.

emmeowzing:
Shouldn't this get added to the appropriate release?

https://github.com/helm/helm/milestone/166

benoittgt:
If this is still wanted I think a new PR needs to be opened and code rebased on top of main.

loewenstein-sap:
I still think this is a useful feature, nicely complementing the support for import-values and introducing some rudimentary "data orchestration" between Helm chart hierarchies.

Looking at the initial date of this PR though, I think it would be up to Helm maintainers to pick it up if they see this as a useful addition to Helm. This PR is not the best sign for project health if you ask me.

joejulian:
Closing this. The home for export-values is HIP-0016 and the discussion around it, not a four-year-old implementation PR, and this one has sat through repeated stale cycles without the rebase Scott flagged it needs.

To be clear about what is and is not being closed: export-values as a feature is not rejected. But HIP-0016 is still in `draft` (open since 2022), so the design is not settled, and this PR is built on the pre-Helm-4 `pkg/chartutil` layout, so it would need a rewrite against `main` regardless. Carrying an unmaintained implementation ahead of an unsettled HIP does not move either one forward, it just keeps a stale PR on the board.

If the feature is still wanted, the path is to take HIP-0016 to a decision and then implement it fresh against Helm 4 `main`. I am glad to help review the HIP if the authors want to drive it.

Thank you @c0d1ngm0nk3y for the work, and the SAP team for the HIP. The want was legitimate. This is housekeeping on a stalled PR, not a no on export-values.

Issue #10597, opened 2022-01-21
Bug - Client certificate not reaching for tls handshake (helm oci pull) 


suryatech27-cloud:
Helm Version - v3.7.2
While doing **helm oci:// pull for artifact repository** its causes to handshake failure issue. 

So I debug helm CLI code, I just added a few print statements in go native library - net/HTTP - transport.go 
addTLS function does not get tlsconfig value for a handshake so handshake failure is happening.

`transport.go : addTLS () =================> TLSClientConfig :  &{<nil> <nil> [] map[] <nil> <nil> <nil> <nil> <nil> <nil> [h2 http/1.1]  NoClientCert <nil> false [] false false [0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0] <nil> 0 0 [] false 0 <nil> {{0 0} 0 0 0 0} [] []}`

`Error: failed to do request: Head "https://myprivatedns.com/v2/reponame/context/demotestchart/manifests/1.0": remote error: tls: handshake failure
`

so I manually added few line of code for client certificate in **transport.go** , now the tls handshake failure got resolved. but now i am getting **401 unauthorized.**

```
func (pconn *persistConn) addTLS(name string, trace *httptrace.ClientTrace) error {
	fmt.Println("transport.go : addTLS () =================> TLSClientConfig : ", pconn.t.TLSClientConfig)
	// Initiate TLS and check remote host name against certificate.
	if tlsConf := pconn.t.TLSClientConfig; tlsConf.Certificates == nil && tlsConf.ClientCAs == nil && tlsConf.CipherSuites == nil {
		cert, err := tls.LoadX509KeyPair("/etc/docker/certs.d/myprivatedns.com/client.pem", "/etc/docker/certs.d/myprivatedns.com/client.key")
		if err != nil {
			fmt.Println(err, "Failed in LoadX509KeyPair")
		}
		capool := x509.NewCertPool()
		cacert, err := ioutil.ReadFile("/etc/docker/certs.d/myprivatedns.com/cacerts.crt")
		capool.AppendCertsFromPEM(cacert)
		if err != nil {
			fmt.Println(err, "readCacerts from pool")
		}

		tlsConfig := &tls.Config{
			Certificates: []tls.Certificate{cert},
			ClientCAs:    capool,
			ClientAuth:   tls.RequireAndVerifyClientCert,
[...]

PR #10808, opened 2022-03-28
Fix - TLS hnadshake failure for oci pull (artifact repo).

suryatech27-cloud:
Fix for https://github.com/helm/helm/issues/10597
- Added 2-way TLS Support for oci pull for artifact repository which causes TLS handshake failure error.

- For now, this fix will be going to support Linux only.
**If applicable**:

### DIFF
--- a/internal/tlsutil/cfg.go
+++ b/internal/tlsutil/cfg.go
@@ -19,7 +19,13 @@ package tlsutil
 import (
 	"crypto/tls"
 	"crypto/x509"
+	"fmt"
+	"io/ioutil"
 	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"strings"
 
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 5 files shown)


bacongobbler:
Hi @suryatech27-cloud, generally speaking we ask that pull requests come with a set of unit tests to ensure the feature works as expected. Can you please have a look at existing test cases and see if you can write one for this?

Additionally, is there a way maintainers should test this PR?

Thank you.

pkg/registry/client.go
```diff
@@ -131,6 +135,94 @@ func NewClient(options ...ClientOption) (*Client, error) {
 	return client, nil
 }
 
+func NewCrosClient(chartref string, options ...ClientOption) (*Client, error) {
```
bacongobbler, line 138: what is a "Cros" client?

pkg/action/pull.go
```diff
@@ -125,7 +125,26 @@ func (p *Pull) Run(chartRef string) (string, error) {
 
 	saved, v, err := c.DownloadTo(chartRef, p.Version, dest)
 	if err != nil {
-		return out.String(), err
+		if registry.IsOCI(chartRef) && strings.Contains(fmt.Sprint(err), "remote error: tls: handshake failure") {
+			registryClient, err := registry.NewCrosClient(chartRef,
```
bacongobbler, line 129: Why are we re-creating a one-off registry client instead of altering `c.RegistryClient`? Shouldn't this setup code be part of [NewOciGetter](https://github.com/helm/helm/blob/c137bfb68b4f8b7a10687f043b0dbccb09574969/pkg/getter/ocigetter.go#L65)?

bacongobbler (changes requested):

pkg/registry/client.go
suryatech27-cloud: **Cros  (Connection Roundtripper Over Ssl)** 
This **NewCrosclient** only uses two-way TLS support. Due to less experience in golang development, I thought maybe my code affect existing functionality so I created a new function named as newCrosClient.

If that naming convention does not match helm expectations we can change that.

pkg/action/pull.go
suryatech27-cloud: Yes, you are right but there is one problem if we add that setup code in [NewOciGetter](https://github.com/helm/helm/blob/c137bfb68b4f8b7a10687f043b0dbccb09574969/pkg/getter/ocigetter.go#L65) then we need to distinguish between one-way TLS and two-way TLS. 
if we validate then a specific client setup will be done because we don't need a two-way tls client setup for one-way and vice-versa.

`if registry.IsOCI(chartRef) && strings.Contains(fmt.Sprint(err), "remote error: tls: handshake failure")`
so above code is validating when client couldn't establish connection(required client cert and key) then we setting up the client with roundtripper/tlsconfig data.

suryatech27-cloud: **(When the Registry server is a one-way TLS that time it works fine  but when the registry server is a two-way TLS , the server looking for client certificate/key (two-way TLS), then the helm client does not provide the client info (TLSconfig) that causes a handshake failure. so my code is to resolve those issue)**

suryatech27-cloud:
> Hi @suryatech27-cloud, generally speaking we ask that pull requests come with a set of unit tests to ensure the feature works as expected. Can you please have a look at existing test cases and see if you can write one for this?
> 
> Additionally, is there a way maintainers should test this PR?
> 
> Thank you.

@bacongobbler, Yes there is a way to test this PR, just need to enable two-way SSL/TLS in the registry server and validate.

internal/tlsutil/cfg.go
```diff
@@ -56,3 +62,77 @@ func ClientConfig(opts Options) (cfg *tls.Config, err error) {
 	cfg = &tls.Config{InsecureSkipVerify: opts.InsecureSkipVerify, Certificates: []tls.Certificate{*cert}, RootCAs: pool}
 	return cfg, nil
 }
+
+func ReadCertFromSecDir(host string) (opts Options, err error) {
+	if runtime.GOOS == "windows" || runtime.GOOS == "unix" {
+		fmt.Printf("%v OS not supported for this oci pull. Contact your administrator for more information !!!", runtime.GOOS)
```
bacongobbler, line 68: Why should the user have to contact their administrator for Helm support?

internal/tlsutil/cfg.go
```diff
@@ -56,3 +62,77 @@ func ClientConfig(opts Options) (cfg *tls.Config, err error) {
...
+		os.Exit(1)
```
bacongobbler, line 69: https://pkg.go.dev/log#Fatalf

internal/tlsutil/cfg.go
```diff
@@ -56,3 +62,77 @@ func ClientConfig(opts Options) (cfg *tls.Config, err error) {
...
+	} else {
+		cmd, err := exec.Command("helm", "env", "HELM_CLIENT_TLS_CERT_DIR").Output()
+		if err != nil {
+			fmt.Printf("Error : %s", err)
```
bacongobbler, line 73: https://pkg.go.dev/log#Fatalf

internal/tlsutil/cfg.go
```diff
@@ -56,3 +62,77 @@ func ClientConfig(opts Options) (cfg *tls.Config, err error) {
...
+
+		if clientCertDir[len(clientCertDir)-1] != '/' {
+			clientCertDir = fmt.Sprintf("%s/%s", clientCertDir, host)
+		} else {
+			clientCertDir = fmt.Sprintf("%s%s", clientCertDir, host)
+		}
+		if _, err := os.Stat(clientCertDir); err != nil {
+			if os.IsNotExist(err) {
+				os.MkdirAll(clientCertDir, os.ModePerm)
+				return opts, errors.Wrapf(err, clientCertDir, "%v\n%v Directory created.")
```
bacongobbler, line 90: why are we creating the directory, then throwing back the same error?

internal/tlsutil/cfg.go
```diff
@@ -56,3 +62,77 @@ func ClientConfig(opts Options) (cfg *tls.Config, err error) {
...
+			}
+		} else {
+
+			if files, err := ioutil.ReadDir(clientCertDir); err == nil {
+				for _, file := range files {
+					if filepath.Ext(file.Name()) == ".crt" {
```
bacongobbler, line 96: why is this necessary?

internal/tlsutil/cfg.go
```diff
@@ -56,3 +62,77 @@ func ClientConfig(opts Options) (cfg *tls.Config, err error) {
...
+						opts.CaCertFile = fmt.Sprintf("%s/%s", clientCertDir, file.Name())
+					} else if filepath.Ext(file.Name()) == ".pem" {
+						opts.CertFile = fmt.Sprintf("%s/%s", clientCertDir, file.Name())
+					} else if filepath.Ext(file.Name()) == ".key" {
+						opts.KeyFile = fmt.Sprintf("%s/%s", clientCertDir, file.Name())
+					}
+				}
+			} else {
+				fmt.Printf(" Certificate not found in current directory - %v\n ", err)
```
bacongobbler, line 105: https://pkg.go.dev/log#Fatalf

bacongobbler (changes requested):

bacongobbler:
> Yes there is a way to test this PR, just need to enable two-way SSL/TLS in the registry server and validate.

I do not know how to set that up. Please provide a set of instructions so developers can test this PR. It is your responsibility to teach others how to test this.

pkg/action/pull.go
bacongobbler: Just make two-way TLS an `Option` like we do for the rest of the getter feature flags.

for example:

https://github.com/helm/helm/blob/c137bfb68b4f8b7a10687f043b0dbccb09574969/pkg/getter/getter.go#L54

bacongobbler:
Alternatively provide support for two-way TLS with the test server and use that against these changes.

https://github.com/helm/helm/blob/main/pkg/repo/repotest/server.go#L75

internal/tlsutil/cfg.go
suryatech27-cloud: Yeah, you are right user don't need to contact their administrator because our user message is sufficient.
I will remove that message.
Thanks.

internal/tlsutil/cfg.go
suryatech27-cloud: Correct, that has no meaning if we are creating the directory and throwing back the same error.
I will remove the directory creation code and will update the error message.

Thanks.

bacongobbler:
I just realized this pull request is targeting the release-3.8 branch. In the future, please target your pull requests against the `main` branch.

Closing as we are not going to merge PRs against a release branch. Please re-open against `main`. Thanks.

yxxhero:
@suryatech27-cloud ping

PR #10810, opened 2022-03-28
Apply provenance file ending to URL path

### LINKED ISSUE #10770: Helm cannot download provenance files when the Chart URL has a query string.
Output of `helm version`: v3.7.1

Output of `kubectl version`: N/A - This fails before communicating with Kubernetes

Cloud Provider/Platform (AKS, GKE, Minikube etc.): N/A - This fails before communicating with Kubernetes

Steps to reproduce:

Create a helm repository `index.yaml` where a Chart has the following URL: `https://www.apache.org/dyn/closer.lua/solr/solr-operator/v0.5.0/helm-charts/solr-0.5.0.tgz?action=download`

All commands will pass and this will work until you try to use `--verify`, because Helm tries to add `.prov` to the end of the URL instead of adding it to the end of the path, like it does throughout the rest of the codebase.

```
failed to decode signature
helm.sh/helm/v3/pkg/provenance.(*Signatory).Verify
	helm.sh/helm/v3/pkg/provenance/sign.go:246
helm.sh/helm/v3/pkg/downloader.VerifyChart
	helm.sh/helm/v3/pkg/downloader/chart_downloader.go:308
helm.sh/helm/v3/pkg/downloader.(*ChartDownloader).DownloadTo
	helm.sh/helm/v3/pkg/downloader/chart_downloader.go:131
helm.sh/helm/v3/pkg/action.(*ChartPathOptions).LocateChart
	helm.sh/helm/v3/pkg/action/install.go:753
main.runShow
	helm.sh/helm/v3/cmd/helm/show.go:205
main.newShowCmd.func4
	helm.sh/helm/v3/cmd/helm/show.go:121
	github.com/spf13/cobra@v1.2.1/command.go:856
	github.com/spf13/cobra@v1.2.1/command.go:974
main.main
	helm.sh/helm/v3/cmd/helm/helm.go:87
runtime.main
	runtime/proc.go:255
runtime.goexit
	runtime/asm_amd64.s:1581
```

Note: This is how official apache release artifacts are meant to be downloaded, so a fix here would very much help the many Apache projects that provide official Helm charts.

HoustonPutman:
closes #10770

This fixes where the `.prov` ending is placed on URLs (the end of the path rather than the end of the URL).

However if the new implementation is not successful, it will attempt to download the provenance file using the old strategy. This allows for backwards compatibility.

The only issue is if the new method downloads the wrong thing (not the provenance file), but the server does not send back an error response. We will save the bad information as the provenance file, then fail at a later point instead of failing immediately and trying the old method. This is the only way backwards-compatibility would be broken.

**If applicable**:

### DIFF
--- a/pkg/downloader/chart_downloader.go
+++ b/pkg/downloader/chart_downloader.go
@@ -116,13 +116,21 @@ func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *proven
 	// If provenance is requested, verify it.
 	ver := &provenance.Verification{}
 	if c.Verify > VerifyNever {
-		body, err := g.Get(u.String() + ".prov")
+		initialURL := u.String()
+		u.Path += ".prov"
+		body, err := g.Get(u.String())
 		if err != nil {
-			if c.Verify == VerifyAlways {
-				return destfile, ver, errors.Errorf("failed to fetch provenance %q", u.String()+".prov")
+			// Try again adding the ".prov" to the end of the URL (not the path).
+			// This maintains backwards-compatibility with chart Repos built to work with older versions of Helm
+			var retryErr error
+			body, retryErr = g.Get(initialURL + ".prov")
+			if retryErr != nil {
+				if c.Verify == VerifyAlways {
+					return destfile, ver, errors.Errorf("failed to fetch provenance %q", u.String())
+				}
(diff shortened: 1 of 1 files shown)


bacongobbler:
Can you please write a unit test for this?

https://github.com/helm/helm/blob/ee3f270e1eff0d462312635ad91cecd6f1fce620/pkg/downloader/chart_downloader_test.go#L172

Thanks.

PR #10811, opened 2022-03-28
On hook failure, the cleanup mechanism should also run on all previous hooks

### LINKED ISSUE #10279: Hook delete policy of "hook-failed" not honored for preceding resource on a failure
Output of `helm version`: `version.BuildInfo{Version:"v3.6.0", GitCommit:"7f2df6467771a75f5646b7f12afb408590ed1755", GitTreeState:"clean", GoVersion:"go1.16.3"}`

Output of `kubectl version`: `Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.3", GitCommit:"1e11e4a2108024935ecfcb2912226cedeafd99df", GitTreeState:"clean", BuildDate:"2020-10-14T12:50:19Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"17+", GitVersion:"v1.17.17-eks-087e67", GitCommit:"087e67e479962798594218dc6d99923f410c145e", GitTreeState:"clean", BuildDate:"2021-07-31T01:39:55Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}`

Cloud Provider/Platform (AKS, GKE, Minikube etc.): AWS / EKS

This looks to be the same issue as [this](https://github.com/helm/helm/issues/3714) which was closed due to inactivity. That issue is for on helm 2, but I believe the same issue persists on helm 3. 

Let's say I have 2 Job resources that i want to run during the `helm test` hook. I set both of them with `"helm.sh/hook-delete-policy": "hook-succeeded,hook-failed"`.

Now if both succeed, they are both deleted. Perfect.

But, if the first one succeeds and the second one fails, then the first one is never deleted. 

From looking at the [code](https://github.com/helm/helm/blob/a499b4b179307c267bdf3ec49b880e3dbd2a5591/pkg/action/hooks.go#L30-L108), it looks like if all hook resources succeed, we delete all of the ones with `hook-succeeded` at the very end of the process. But, if one of them fails, we only delete that 1 resource and immediately return with the error. 

This isn't what I would expect, and believe it's probably not what most people would want. 

In the case of `Job` resources specifically, this isn't much of an issue cause we can just set the `ttlSecondsAfterFinished`. But for things like `Deployment` resources, it causes more problems cause you end up with running pods sticking around when you don't want them to. 

Here's a relatively simple demonstration. 

```yaml
apiVersion: batch/v1
kind: Job
metadata:
  name: "successful-job"
  labels:
    "app": "hook-testing"
  annotations:
    "helm.sh/hook": "test"
    "helm.sh/hook-weight": "1"
    "helm.sh/hook-delete-policy": "hook-succeeded,hook-failed"
spec:
  template:
    spec:
      restartPolicy: Never
      containers:
      - name: "successful-job"
        command:
          - bash
          - -c 
          - exit 0
        image: bash:latest
---
apiVersion: batch/v1
kind: Job
metadata:
  name: "failure-job"
  labels:
    "app": "hook-testing"
  annotations:
    "helm.sh/hook": "test"
    "helm.sh/hook-weight": "2"
    "helm.sh/hook-delete-policy": "hook-succeeded,hook-failed"
spec:
  backoffLimit: 0
  template:
    spec:
      restartPolicy: Never
      containers:
      - name: "failure-job"
        command:
          - bash
          - -c 
          - exit 1
        image: bash:latest
---
apiVersion: v1
kind: Pod
metadata:
  name: "non-hook-pod"
  labels:
    "app": "hook-testing"
spec:
  containers:
  - name: "sleeping-pod"
    command:
      - bash
      - -c 
      -  while :; do sleep 1; done
    image: bash:latest
```

prafull01:
closes #10279 

Currently, In case of multiple hooks if any of the hook fail, only that hook is deleted (if deletion policy is set). The other hooks are not deleted because code immediately returns error after cleaning up failed hook.

This PR adds mechanism to not only clean the failed hook, but clean the successful hook as well based on their deletion policy

**If applicable**:

### DIFF
--- a/pkg/action/hooks.go
+++ b/pkg/action/hooks.go
@@ -41,7 +41,7 @@ func (cfg *Configuration) execHook(rl *release.Release, hook release.HookEvent, 
 	// hooke are pre-ordered by kind, so keep order stable
 	sort.Stable(hookByWeight(executingHooks))
 
-	for _, h := range executingHooks {
+	for i, h := range executingHooks {
 		// Set default delete policy to before-hook-creation
 		if h.DeletePolicies == nil || len(h.DeletePolicies) == 0 {
 			// TODO(jlegrone): Only apply before-hook-creation delete policy to run to completion
@@ -91,6 +91,12 @@ func (cfg *Configuration) execHook(rl *release.Release, hook release.HookEvent, 
 			if err := cfg.deleteHookByPolicy(h, release.HookFailed); err != nil {
 				return err
 			}
+
+			// If a hook is failed, check the annotation of the previous successful hooks to determine whether the hook
+			// should be deleted under succeeded condition.
+			if err := cfg.deleteHooksByPolicy(executingHooks[0:i], release.HookSucceeded); err != nil {
+				return err
+			}
 			return err
 		}
 		h.LastRun.Phase = release.HookPhaseSucceeded
@@ -98,10 +104,8 @@ func (cfg *Configuration) execHook(rl *release.Release, hook release.HookEvent, 
 
 	// If all hooks are successful, check the annotation of each hook to determine whether the hook should be deleted
 	// under succeeded condition. If so, then clear the corresponding resource object in each hook
-	for _, h := range executingHooks {
-		if err := cfg.deleteHookByPolicy(h, release.HookSucceeded); err != nil {
-			return err
-		}
+	if err := cfg.deleteHooksByPolicy(executingHooks, release.HookSucceeded); err != nil {
+		return err
 	}
 
 	return nil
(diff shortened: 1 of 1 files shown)


bacongobbler:
Can you please write a unit test for `execHook` to catch and check for this new behavior?

prafull01:
IMHO execHook function can not be properly covered with unit tests. The execHook function creates hooks resources reading from the release and deletes the successful/failed hooks resources based on the delete policy.  Unless I can check that a particular resources is created or deleted (as per delete policy) after running the `execHook` function. This can be only achieved if we are running `kube-apiserver` and `etcd` (using envtest) where this objects are persisted.

Do you have any pointer, how can I write unit tests for this function?

prafull01:
@bacongobbler Can you please have a look at above comment. Thanks :)

uvegla:
@bacongobbler @prafull01 Hi! 👋🏻 I have a pull request with this change and tests added: https://github.com/helm/helm/pull/11334 Could you please check it?

joejulian:
As this PR has had no responses and there's another PR that has the requested changes, I'm closing this in favor of #11334.

PR #10830, opened 2022-04-04
Updating vcs to latest version

mattfarina:
vcs had a release due to a CVE. This updates to the latest version
that mitigates the CVE.

Note, the Makefile was updated so that `make build` would build
without cgo just like gox does in the CI pipeline. They should
both build without cgo so we can catch issues before merging to
master where a canary build would pick up the problem.

It updates a dependency to the latest.

**If applicable**:


hickeyma (approved):
LGTM, thanks @mattfarina

PR #10831, opened 2022-04-04

### LINKED ISSUE #8660: Wait logic doesn't wait for all deployment and daemonset pods to be updated/ready

cbf123:
The wait logic for Deployments and Daemonsets subtracts their
maxUnavailable value from the amount of replicas they should
have ready. This can lead to waiting for 0 pods to be ready.
This change ensures at least 1 replica is ready, unless there
are 0 desired replicas.

Fixes #8660.

This is based on https://github.com/helm/helm/pull/8671 which
was never merged.

Signed-off-by: Chris Friesen <chris.friesen@windriver.com>

**If applicable**:

### DIFF
--- a/pkg/kube/ready.go
+++ b/pkg/kube/ready.go
@@ -272,7 +272,17 @@ func (c *ReadyChecker) volumeReady(v *corev1.PersistentVolumeClaim) bool {
 }
 
 func (c *ReadyChecker) deploymentReady(rs *appsv1.ReplicaSet, dep *appsv1.Deployment) bool {
+	// If there are no desired replicas, there will be nothing to wait for.
+	if *dep.Spec.Replicas == 0 {
+		return true
+	}
+
+	// Ensure the expected available replicas are ready.
 	expectedReady := *dep.Spec.Replicas - deploymentutil.MaxUnavailable(*dep)
+	// Also ensure at least 1 replica is ready.
+	if expectedReady < 1 {
+		expectedReady = 1
+	}
 	if !(rs.Status.ReadyReplicas >= expectedReady) {
 		c.log("Deployment is not ready: %s/%s. %d out of %d expected pods are ready", dep.Namespace, dep.Name, rs.Status.ReadyReplicas, expectedReady)
 		return false
@@ -286,11 +296,18 @@ func (c *ReadyChecker) daemonSetReady(ds *appsv1.DaemonSet) bool {
 		return true
 	}
 
+	// If there are no desired replicas, there will be nothing to wait for.
+	if ds.Status.DesiredNumberScheduled == 0 {
+		return true
+	}
+
 	// Make sure all the updated pods have been scheduled
 	if ds.Status.UpdatedNumberScheduled != ds.Status.DesiredNumberScheduled {
 		c.log("DaemonSet is not ready: %s/%s. %d out of %d expected pods have been scheduled", ds.Namespace, ds.Name, ds.Status.UpdatedNumberScheduled, ds.Status.DesiredNumberScheduled)
 		return false
 	}
+
+	// Ensure the expected available replicas are ready.
 	maxUnavailable, err := intstr.GetValueFromIntOrPercent(ds.Spec.UpdateStrategy.RollingUpdate.MaxUnavailable, int(ds.Status.DesiredNumberScheduled), true)
 	if err != nil {
 		// If for some reason the value is invalid, set max unavailable to the
@@ -298,8 +315,11 @@ func (c *ReadyChecker) daemonSetReady(ds *appsv1.DaemonSet) bool {
 		// `MaxUnavailable` function in deploymentutil
 		maxUnavailable = int(ds.Status.DesiredNumberScheduled)
 	}
-
 	expectedReady := int(ds.Status.DesiredNumberScheduled) - maxUnavailable
+	// Also ensure at least 1 replica is ready.
+	if expectedReady < 1 {
+		expectedReady = 1
+	}
 	if !(int(ds.Status.NumberReady) >= expectedReady) {
 		c.log("DaemonSet is not ready: %s/%s. %d out of %d expected pods are ready", ds.Namespace, ds.Name, ds.Status.NumberReady, expectedReady)
 		return false


cbf123:
See https://github.com/helm/helm/pull/8671 for the original context, but basically this is a rebase of the original fix from that pull request.  I contacted Sean Eagan privately and he said that he had given up on getting it merged.

cbf123:
This issue is causing problems for FluxCD currently: https://github.com/fluxcd/helm-controller/issues/81

cbf123:
Just wondering if any of the maintainers have had a chance to look at this yet...

mattfarina (commented):
Thanks for working on this. I would love to see this bug fixed.

Can you add a test. There are other ready tests that could serve as a bases for writing a new one.

mattfarina (commented):
Thanks for working on the rebase and getting this up.

The commit was originally by @seaneagan but the current commit doesn't have anything about him in it. Can you please amend the commit to add him. Using something like `Co-authored-by: author-name <name@xxx.com>` and then add a second `Signed-off-by:` with his info.

mattfarina (approved):

pkg/kube/ready.go
```diff
@@ -272,7 +272,17 @@ func (c *ReadyChecker) volumeReady(v *corev1.PersistentVolumeClaim) bool {
 }
 
 func (c *ReadyChecker) deploymentReady(rs *appsv1.ReplicaSet, dep *appsv1.Deployment) bool {
+	// If there are no desired replicas, there will be nothing to wait for.
+	if *dep.Spec.Replicas == 0 {
+		return true
+	}
```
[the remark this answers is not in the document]
cbf123: I've got no objections to adding a test, but it'll be some time before I can get to it.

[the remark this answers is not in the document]
scottrigby: @cbf123 Thinking we may only need something like this below. Looking at the function, I’m thinking the other params don't matter as long as `newDeployment()` passes 0 replicas. Does this make sense to you?

```diff
diff --git a/pkg/kube/ready_test.go b/pkg/kube/ready_test.go
index 931b8fa1..8400b118 100644
--- a/pkg/kube/ready_test.go
+++ b/pkg/kube/ready_test.go
@@ -57,6 +57,14 @@ func Test_ReadyChecker_deploymentReady(t *testing.T) {
                        },
                        want: false,
                },
+               {
+                       name: "deployment is ready when there are no desired replicas",
+                       args: args{
+                               rs:  newReplicaSet("foo", 0, 0),
+                               dep: newDeployment("foo", 0, 1, 0),
+                       },
+                       want: true,
+               },
                {
                        name: "deployment is ready when maxUnavailable is set",
                        args: args{
```

And similar with DaemonSet?

```diff
@@ -106,6 +114,13 @@ func Test_ReadyChecker_daemonSetReady(t *testing.T) {
                        },
                        want: false,
                },
+               {
+                       name: "daemonset is ready when there are no desired number scheduled",
+                       args: args{
+                               ds: newDaemonSet("foo", 0, 0, 0, 0),
+                       },
+                       want: true,
+               },
                {
                        name: "daemonset is ready when maxUnavailable is set",
                        args: args{
```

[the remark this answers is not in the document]
cbf123: I am not an expert in this code but that looks reasonable for the case where the desired number of replicas is zero.

I think we'd also need new tests for the cases in lines 284 and 321 below where expectedReady had been smaller than 1 and we're forcing it to 1.

pkg/kube/ready.go
```diff
@@ -272,7 +272,17 @@ func (c *ReadyChecker) volumeReady(v *corev1.PersistentVolumeClaim) bool {
...
+
+	// Ensure the expected available replicas are ready.
 	expectedReady := *dep.Spec.Replicas - deploymentutil.MaxUnavailable(*dep)
+	// Also ensure at least 1 replica is ready.
+	if expectedReady < 1 {
+		expectedReady = 1
+	}
```
mattfarina, line 285: I would actually like to revisit this. Sorry. As I was writing tests for this is occurred to me that if someone explicitly sets their max unavailable higher than their desired replicas than Helm would be going against the request if it didn't honor that.

So, if a user says they want 2 replicas and their max unavailable is 2 than they are saying they are OK with none being available and the app being ready. While I would not suggest doing that and an end user may not want that, the issue isn't in Helm but in what they declared if they don't want that.

Why should Helm do something other than what is explicitly declared?

mattfarina (changes requested):

pkg/kube/ready.go
mattfarina: Sorry @cbf123, I found a base reasoning question.

mattfarina: Who determines when something is "ready"? In the Kubernetes sense, this is codified in the resources which is what Helm reads. I don't think Helm should try to be smarter than that and override what users have specified. If a user didn't specify what they really want is a different problem.

If Helm decides to do something different from what is put in the manifests by the users we will get other bugs because of this. We would need a really good reason.

If we're talking about taking PDB into account, that's interesting. How do you see that working in to this?

felipecrs: I think we should follow whatever `kubectl` does. Currently `kubectl` reports that the deployment isn't ready if not at least 1 pod is ready, even with a `maxUnavailable` of 1.

felipecrs:
For those who need this PR, I think the below is the easiest way to build a custom helm binary with the version of your choice with this PR applied.

```
$ cat <<'EOF'
FROM golang:1.19.3 AS helm

WORKDIR /go/src

ARG HELM_VERSION=3.9.4
RUN curl -fsSL https://github.com/helm/helm/archive/refs/tags/v${HELM_VERSION}.tar.gz | tar -xz --strip-components=1

# Apply patch for PR https://github.com/helm/helm/pull/10831
RUN curl -fsSL https://github.com/helm/helm/commit/4e21723a0c1f4fcb92df40c89b9c7b36e47a15e1.patch | git apply

# Build helm
RUN make build

FROM scratch

COPY --from=helm /go/src/bin/helm /helm
EOF

$ docker buildx build . --output=.
```

Which after its execution will place the `helm` binary in your current directory.

cbf123:
I don't think there's any appetite for pulling this in, but I've rebased the code change and made it available at https://github.com/cbf123/helm/tree/wait-1-replica-rebased

joejulian:
I'm going to go ahead and close this based on [the last response](https://github.com/helm/helm/pull/10831#issuecomment-1460491038).

Issue #10285, opened 2021-10-28
Error: INSTALLATION FAILED: failed to download


l1z4a:
Output of `helm version`:
```
$ helm version
version.BuildInfo{Version:"v3.7.1", GitCommit:"1d11fcb5d3f3bf00dbe6fe31b8412839a96b3dc4", GitTreeState:"clean", GoVersion:"go1.16.9"}
```

Output of `kubectl version`:
```
$ sudo k3s kubectl version
Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.5+k3s2", GitCommit:"724ef700bab896ff252a75e2be996d5f4ff1b842", GitTreeState:"clean", BuildDate:"2021-10-05T19:59:14Z", GoVersion:"go1.16.8", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.5+k3s2", GitCommit:"724ef700bab896ff252a75e2be996d5f4ff1b842", GitTreeState:"clean", BuildDate:"2021-10-05T19:59:14Z", GoVersion:"go1.16.8", Compiler:"gc", Platform:"linux/amd64"}
```

Cloud Provider/Platform (AKS, GKE, Minikube etc.): 
```
$ sudo k3s --version
k3s version v1.21.5+k3s2 (724ef700)
go version go1.16.8
```

Steps to reproduce failure:
* Use a fresh Ubuntu 20.04 server edition installation with packages up-to-date
* `$ curl -sfL https://get.k3s.io | sh -`
* `$ alias k="sudo k3s kubectl"`
* `$ curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3`
* `$ chmod 700 get_helm.sh`
* `$ ./get_helm.sh`

```
Downloading https://get.helm.sh/helm-v3.7.1-linux-amd64.tar.gz
Verifying checksum... Done.
Preparing to install helm into /usr/local/bin
helm installed into /usr/local/bin/helm
```

* `$ helm repo add rancher-stable https://releases.rancher.com/server-charts/stable`
* `k create namespace cattle-system`
* `$ helm repo add jetstack https://charts.jetstack.io`
* `$ export KUBECONFIG=/etc/rancher/k3s/k3s.yaml`
* `$ helm repo update`

```
Hang tight while we grab the latest from your chart repositories...
...Successfully got an update from the "jetstack" chart repository
[...]

PR #10834, opened 2022-04-05
Log error message on failed download

stan-sz:
Add logging of the actual cause of the failure to download the helm chart. Related to #10285

### DIFF
--- a/cmd/helm/show_test.go
+++ b/cmd/helm/show_test.go
@@ -47,14 +47,14 @@ func TestShowPreReleaseChart(t *testing.T) {
 			name:        "show pre-release chart",
 			args:        "test/pre-release-chart",
 			fail:        true,
-			expectedErr: "failed to download \"test/pre-release-chart\"",
+			expectedErr: "chart \"pre-release-chart\" matching  not found in test index. (try 'helm repo update'): no chart version found for pre-release-chart-",
 		},
 		{
 			name:        "show pre-release chart",
 			args:        "test/pre-release-chart",
 			fail:        true,
 			flags:       "--version 1.0.0",
-			expectedErr: "failed to download \"test/pre-release-chart\" at version \"1.0.0\"",
+			expectedErr: "chart \"pre-release-chart\" matching 1.0.0 not found in test index. (try 'helm repo update'): no chart version found for pre-release-chart-1.0.0",
 		},
 		{
 			name:  "show pre-release chart with 'devel' flag",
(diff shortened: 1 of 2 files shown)


stan-sz:
@bacongobbler - can this PR be merged for the upcoming `3.8.2` release?

hickeyma (changes requested):
@stan-sz Thanks for looking to provide better error handling in this method.

I would suggest looking again at this block: https://github.com/helm/helm/blob/main/pkg/action/install.go#L753-L769

I think we need to refactor this block where the error returned from `dl.DownloadTo(name, version, settings.RepositoryCache)` is handled straight away and not this rather peculiar block with debug.  What do you think?

@bacongobbler @mattfarina If the error message is changed, does this break backwards compatibility?

stan-sz:
Thanks for feedback. Indeed the code can be simplified to just `return filename, err`. Looking forward to hearing for concerns about back-compat.

stan-sz:
@mattfarina It would be great if this change could be part of the upcoming release. Thanks!

stan-sz:
@hickeyma - I've addressed your feedback in the latest iteration. Thanks!

stan-sz:
@mattfarina is this change (and other listed under the 3.9.0 milestone) slated for the 3.9.0 release?

hickeyma (approved):
Thanks for pushing the fix @stan-sz and also the refactor to the code.

I think the error messages are more informative now. Examples shown:

```console
$ helm install cert-manager jetstack/cert-manager --namespace cert-manager --create-namespace --version v1.6.0

Error: INSTALLATION FAILED: repo jetstack not found
```

```console
$ helm install cert-manager jetstack/cert-manager --namespace cert-manager --create-namespace --version v0.0.0 --dry-run

Error: INSTALLATION FAILED: chart "cert-manager" matching v0.0.0 not found in jetstack index. (try 'helm repo update'): no chart version found for cert-manager-v0.0.0
```

hickeyma:
@mattfarina Change to the error message but they bring better clarity to the user on what the error is.

scottrigby (approved):
Agreed these errors are helpful to pass through to the end user.

Verified locally and all looks good. Nice work!

Backwards compatibility rules documentation [hip-0004](https://github.com/helm/community/blob/main/hips/hip-0004.md#specification) says this is acceptable for CLI output:
> - errors can change if doing so increases their usability

PR #10854, opened 2022-04-12
Fix confusing test failure message

### LINKED ISSUE #10466: Confusing test error message when `wantError` is not true
If a test case has `wantError: false` (explicitly or not; `false` is the default) and the command being tested does throw an error, the error message can be pretty confusing. For example:

```diff
diff --git a/cmd/helm/lint_test.go b/cmd/helm/lint_test.go
index 3501ccf8..b03d9682 100644
--- a/cmd/helm/lint_test.go
+++ b/cmd/helm/lint_test.go
@@ -27,7 +27,7 @@ func TestLintCmdWithSubchartsFlag(t *testing.T) {
                name:      "lint good chart with bad subcharts",
                cmd:       fmt.Sprintf("lint %s", testChart),
                golden:    "output/lint-chart-with-bad-subcharts.txt",
-               wantError: true,
+               wantError: false,
        }, {
                name:      "lint good chart with bad subcharts using --with-subcharts flag",
                cmd:       fmt.Sprintf("lint --with-subcharts %s", testChart),
```

```console
$ GO111MODULE=on go test  -run TestLint ./... -race
--- FAIL: TestLintCmdWithSubchartsFlag (0.15s)
    --- FAIL: TestLintCmdWithSubchartsFlag/lint_good_chart_with_bad_subcharts (0.07s)
        helm_test.go:61: running cmd (attempt 1): lint testdata/testcharts/chart-with-bad-subcharts
        helm_test.go:64: expected error, got '1 chart(s) linted, 1 chart(s) failed'
FAIL
FAIL	helm.sh/helm/v3/cmd/helm	0.315s
```

The reason seems to be the semantical mismatch between [this `if` statement](https://github.com/helm/helm/blob/47c2fd78a2b793f904ead0a9364ebec14814c271/cmd/helm/helm_test.go#L63:L65)'s antecedent (which is essentially an XOR expression) and its consequent (which doesn't acknowledge that):

```go
if (err != nil) != tt.wantError {
	t.Errorf("expected error, got '%v'", err)
}
```

(The exact same problem is present [here](https://github.com/helm/helm/blob/47c2fd78a2b793f904ead0a9364ebec14814c271/cmd/helm/helm_test.go#L85:L87), but I think `runTestActionCmd` should be removed anyway, as suggested in #10453.)

SimonAlling:
Resolves #10466.

### DIFF
--- a/cmd/helm/helm_test.go
+++ b/cmd/helm/helm_test.go
@@ -60,8 +60,11 @@ func runTestCmd(t *testing.T, tests []cmdTestCase) {
 				}
 				t.Logf("running cmd (attempt %d): %s", i+1, tt.cmd)
 				_, out, err := executeActionCommandC(storage, tt.cmd)
-				if (err != nil) != tt.wantError {
-					t.Errorf("expected error, got '%v'", err)
+				if tt.wantError && err == nil {
+					t.Errorf("expected error, got success with the following output:\n%s", out)
+				}
+				if !tt.wantError && err != nil {
+					t.Errorf("expected no error, got: '%v'", err)
 				}
 				if tt.golden != "" {
 					test.AssertGoldenString(t, out, tt.golden)


yxxhero:
@SimonAlling Thanks for your PR. I think all is ok.

hickeyma (approved):

PR #10861, opened 2022-04-14
replace +build with go:build

bacongobbler:
go:build is the new conditional compilation directive used to specify build constraints. It was introduced in Go 1.17. It is meant to replace the old +build directives.

Now that go.mod points to Go 1.17 we no longer need to support both
build flags.

https://pkg.go.dev/cmd/go#hdr-Build_constraints

Signed-off-by: Matthew Fisher <matt.fisher@fermyon.com>

### DIFF
--- a/cmd/helm/root_unix.go
+++ b/cmd/helm/root_unix.go
@@ -1,5 +1,4 @@
 //go:build !windows
-// +build !windows
 
 /*
 Copyright The Helm Authors.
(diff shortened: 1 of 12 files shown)


hickeyma (approved):
As mentioned in the [transition doc](https://go.googlesource.com/proposal/+/master/design/draft-gobuild.md#transition): "Builds will no longer fail if a Go file contains `//go:build` without `// +build.`". Also, `// +build.` will be removed in 1.18.

Thanks for picking this up @bacongobbler

Issue #6040, opened 2019-07-18
helm install bash script with "--no-sudo" raises an error


mozai:
Helm version: n/a

Steps to reproduce:
- `[ $(id -u) != 0 ] && echo "I Am Not Root"`  #  I Am Not Root
- `curl -LO https://raw.githubusercontent.com/kubernetes/helm/master/scripts/get`
- `chmod 0700 get`
- `./get -h`  # "[--no-sudo]  ->> install without sudo
- `./get --no-sudo`

Expected outcome:
Gets the appropriate "helm" binary for my OS and arch, puts it either in cwd or in the first writable dir in $PATH.  Or reminds me I must provide a destination path if I'm not root.

Seen outcome:
```
Downloading https://get.helm.sh/helm-vX.XX.X-linux-amd64.tar.gz
Preparing to install helm and tiller into /usr/local/bin
cp: cannot create regular file '/usr/local/bin/helm': Permission denied
Failed to install helm with the arguments provided: --no-sudo
```

If I can't use this without being root, why give the option not to use sudo?

Workaround:
- Get the url from the error message
- `cd $(mktemp -d)`
- `curl -L https://get.helm.sh/helm-vX.XX.X-linux-amd64.tar.gz |tar xz`
- copy ./\*/helm and ./\*/tiller to $HOME/.local/bin or wherever is appropriate for your workstation.

Something like this could work (please don't copy-paste this blindly)
```
OS=linux; ARCH=amd64; IFS=':' read -ra paths <<<"$PATH"; for PRG in helm tiller; do
dest=""; for p in ${paths[@]}; do cp "$OS-$ARCH/helm" $p/ && dest="$p/$PRG" && break ;
done; if [ "$dest" ]; then echo "$PRG copied to $dest"; else 
echo "$PRG could not be installed, find it in $(pwd)/$OS-$ARCH/$PRG"; fi; done
```

bacongobbler:
https://github.com/helm/helm/pull/4649 was provided by a community member. If you can figure out the proper way to fix this, that'd be great.

You could also try [setting $USE_SUDO to "false"](https://github.com/helm/helm/blob/88bb48bc6f573e21909147792a16daad7665e45f/scripts/get#L23).

bacongobbler:
I'm going to close this ticket due to inactivity, but please re-open if this still needs to be addressed. Thanks!

CDSRV:
this version is slightly different:
[...]

PR #10879, opened 2022-04-20
fix: Installation error when using `--no-sudo` flag due to lack of privileges

birdx0810:
When using the `--no-sudo` flag, the installation will fail as non-root
user have no privileges for `/usr/local/bin` directory. Changed the
default `$HELM_INSTALL_DIR` to `$HOME/.local/bin` when using `--no-sudo` flag, and allow user to specify install path using `--install-dir` or `-i` flag.

This issue was opened in (refs #6040) but was not resolved and closed due to inactivity (there was a proposed fix later in the issue but not commited).

CHANGES
- Added if-else statement to check if using root/non-root user, and run respective `cp` command. (scripts/get-helm-3#L172-176)
- Added `$HOME/.local/bin` as default installation path if using `--no-sudo` flag.
- Added `--install-dir` path for user to specify install path.

Signed-off-by: Daniel Tan <birdx0810@gmail.com>

### DIFF
--- a/scripts/get-helm-3
+++ b/scripts/get-helm-3
@@ -22,7 +22,8 @@
 : ${DEBUG:="false"}
 : ${VERIFY_CHECKSUM:="true"}
 : ${VERIFY_SIGNATURES:="false"}
-: ${HELM_INSTALL_DIR:="/usr/local/bin"}
+: ${DEFAULT_ROOT_INSTALL_DIR:="/usr/local/bin"}
+: ${DEFAULT_USER_INSTALL_DIR:="${HOME}/.local/bin"}
 : ${GPG_PUBRING:="pubring.kbx"}
 
 HAS_CURL="$(type "curl" &> /dev/null && echo true || echo false)"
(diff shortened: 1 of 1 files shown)


mattfarina:
Why the use of `~/.local/bin` as the location? Is that consistently the location people will use? I ask because there are situations, such as a mac homebrew user, who may have access to /usr/local/bin without being root. While I didn't write the original code, I would not be surprised if it was tested there.

There is an environment variable where you can set the location for the binary.

birdx0810:
> Why the use of `~/.local/bin` as the location? Is that consistently the location people will use? 
I'm not a Mac user, but for other Linux-based OS, the user might not have access to `/usr/local/bin` if they lack sudo privileges. I'm not sure if most developers would prefer the path `~/.local/bin`. That being said, the installation guide for `kubectl` suggests `~/.local/bin` for non-root users ([ref](https://kubernetes.io/docs/tasks/tools/install-kubectl-linux/). Therefore I would assume that `~/.local/bin` would work well as a default path for non-sudoers.

> There is an environment variable where you can set the location for the binary.
Sorry for my ignorance, but is the environment variable `$HELM_INSTALL_DIR`? Following the installation instructions from https://helm.sh/docs/intro/install/, I fail to install the binary to the desired path even when I've set the variable to `~/.local/bin`.

image

I'm new to helm, so I would be happy to update the installation manual (documentation) if that's a better solution for helping new developers like me to have a better installation experience.
Cheers.

birdx0810:
@mattfarina Any thoughts?

PR #10906, opened 2022-04-27
fix(helm): update golang.org/x/crypto to latest

cmurphy:
Update the crypto dependency to address CVE-2022-27191.

Helm does not use the ssh module of crypto, so is not vulnerable to this
CVE, but the dependency gets flagged by vulnerability scanners

Signed-off-by: Colleen Murphy <colleen.murphy@suse.com>


hickeyma (approved):
LGTM, thanks @cmurphy

PR #7728, opened 2020-03-04
Supporting helm3 to show up resource names that were deployed as part…

sarvaniy:
… of release in helm status command

PR #10912, opened 2022-04-28
feat(helm): Supporting helm3 to show up resource names in helm status

SoujanyaMangipudi:
feat(helm): Supporting helm3 to show up resource names that were deployed as part of release in helm status command

Creating a new PR based on this existing stale PR https://github.com/helm/helm/pull/7728

Sample output for helm status command with new --show-resources flag:
image


SoujanyaMangipudi:
@bacongobbler, @mattfarina I picked this previous PR https://github.com/helm/helm/pull/7728 for updating helm status command as this is a feature that will be useful in our internal implementations. This code also fixes some of the issues the previous PR has i.e. duplicate pod display and not all resources getting displayed. Can you please review and provide your feedback on the same? Is possible our target is to merge this feature into helm 3.9.0

yxxhero:
@SoujanyaMangipudi Please squash your commits into one commit.

SoujanyaMangipudi:
> @SoujanyaMangipudi Please squash your commits into one commit.

@yxxhero squashed all my commits as mentioned

SoujanyaMangipudi:
re opening

pkg/kube/interface.go
```diff
@@ -68,6 +68,8 @@ type Interface interface {
 
 	// IsReachable checks whether the client is able to connect to the cluster.
 	IsReachable() error
+
+	Get(resources ResourceList, reader io.Reader) (string, error)
```
mattfarina, line 72: Unfortunately, this Interface can't be changed. It would be compatibility breaking so we can't do it until Helm v4. The [HIP that describes backwards compatibility](https://github.com/helm/community/blob/main/hips/hip-0004.md) points to the [Go projects stance and method for handling this](https://go.dev/blog/module-compatibility#working-with-interfaces).

If you look in this file you'll see we're done this before with `InterfaceExt`. We'll need another new one like `InterfaceGet` or something similar. Could you please add the comment about merging this with the main interface for Helm v4, like `InterfaceExt` does.

Also, could you please document the function.

cmd/helm/status.go
```diff
@@ -124,6 +124,10 @@ func (s statusPrinter) WriteTable(out io.Writer) error {
 		fmt.Fprintf(out, "DESCRIPTION: %s\n", s.release.Info.Description)
 	}
 
+	if len(s.release.Info.Resources) > 0 {
+		fmt.Fprintf(out, "RESOURCES:\n%s\n", s.release.Info.Resources)
+	}
```
mattfarina, line 129: The Helm [HIP on backwards compatibility](https://github.com/helm/community/blob/main/hips/hip-0004.md) covers CLI output and notes that structure can't change. Since Helm v3 has not showed the resources up to this point, displaying them needs to happen behind a flag. Could you please add a flag for this with a note to remove the flag in Helm v4.

mattfarina, line 129: The Helm [HIP on backwards compatibility](https://github.com/helm/community/blob/main/hips/hip-0004.md) covers CLI output and notes that structure can't change. Since Helm v3 has not showed the resources up to this point, displaying them needs to happen behind a flag. Could you please add a flag for this with a note to remove the flag in Helm v4.

mattfarina (changes requested):
Thank you for getting this going again.

While I have not had a chance to do a full review, a couple things jumped right out to me. I was looking for these as we had discussed them with regard to the previous work on this.

cmd/helm/status.go
SoujanyaMangipudi: Updated implementation to move the new status functionality behind `--show-resources` flag as recommended

SoujanyaMangipudi: Updated implementation to move the new status functionality behind `--show-resources` flag as recommended

SoujanyaMangipudi:
@mattfarina a gentle reminder on this PR. Can you please review?

mattfarina:
@SoujanyaMangipudi I was coming to review this (it's on my todo list... for real) and I noticed there is a merge conflict with the `go.sum` file. Can you fix this and then I'll come back and review it. I'm really looking forward to getting this in.

SoujanyaMangipudi:
> @SoujanyaMangipudi I was coming to review this (it's on my todo list... for real) and I noticed there is a merge conflict with the `go.sum` file. Can you fix this and then I'll come back and review it. I'm really looking forward to getting this in.

@mattfarina I merged helm main branch into my branch now and the PR shows no conflicts anymore. Please let me know in case anything else is needed.

cmd/helm/status.go
```diff
@@ -124,6 +127,10 @@ func (s statusPrinter) WriteTable(out io.Writer) error {
 		fmt.Fprintf(out, "DESCRIPTION: %s\n", s.release.Info.Description)
 	}
 
+	if s.showResources && len(s.release.Info.Resources) > 0 {
+		fmt.Fprintf(out, "RESOURCES:\n%s\n", s.release.Info.Resources)
+	}
```
mattfarina, line 132: This is sending a text blob with the status rather than a JSON structure of the status. kubectl will provide a JSON object with the status when outputting json (or YAML when outputting YAML). Would it be possible to have a JSON output for the status instead of the text block reported for table output?

This would, I think, mean moving the printer usage out of `Get` and to the place it's printed.

mattfarina (commented):
Thanks for continuing to work on this PR. It's almost there. When testing with JSON and YAML output formats I noticed an issue and was wondering if you could fix it.

cmd/helm/status.go
SoujanyaMangipudi: @mattfarina While testing for this comment I realized helm status command output with this change is same as what was before for json and yaml flags. json and yaml output already prints manifest which is a blob. This PR did make any changes there. Do you recommend updating the existing command?

SoujanyaMangipudi: @mattfarina while checking further I realized json and yaml outputs for status command already print manifest which is the is the list of release resources. Do you think it will be a better implementation to remove resources from these output formats and keep it only for normal output mode? What is mean is Resources will be shown only when helm status command is executed without any output formats. json or yaml format will not show helm status and continue to behave the way if it is today

Jont828:
@SoujanyaMangipudi Thanks for working on this! I'm working on a project where I import Helm as a library and I'd like to be able to check to see if all the pods/deployments in the manifest are ready. If I understand correctly, is the idea of this PR to query the manifest and build a kube client to query each resource?

cmd/helm/status.go
SoujanyaMangipudi: @mattfarina I really want to merge this PR in release 3.10 and need a little input and help for this comment. At this point I wonder if we can leave this as is as this does not break the json functionality. Json and Yaml commands already print manifest so resources will just be something additional. What are your thoughts on this? @bridgetkromhout  FYI on the resources printing feature for helm 3.10

pkg/kube/client.go
```diff
@@ -132,6 +138,134 @@ func (c *Client) Create(resources ResourceList) (*Result, error) {
...
+	if err != nil {
+		return "", err
+	}
+
+	var keys []string
+	for key := range objs {
+		keys = append(keys, key)
+	}
+
+	for _, t := range keys {
```
mattfarina, line 191: When I try to output using yaml or json the `resources` section is one long string. This is because `Get` returns a string rather than a data structure that can be transformed into yaml, json, or a string output. See...
Screenshot from 2022-09-08 13-39-37

`Get` should return a data structure for the resources. Then when they are printed they can be turned into a string, json, or yaml.

The conversion to a string should happen elsewhere and should not be in the client API.

mattfarina (changes requested):
While I like the idea, the place where the resources are turned into a string is the wrong place. It should not happen within the client API. That should be a data structure that can be turned into the display at the display time. This is the one major change that I think needs to happen. This way yaml and json outputs can be supported in addition to Helm SDK users... so they don't need to parse strings.

pkg/kube/client.go
SoujanyaMangipudi: @mattfarina thank you for your guidance on this! I believe I was able to make the required changes and successfully test on local with all 3 formats
image
Yaml
image
Json
image

Can you please take a look to see if it works?

SoujanyaMangipudi:
> @SoujanyaMangipudi Thanks for working on this! I'm working on a project where I import Helm as a library and I'd like to be able to check to see if all the pods/deployments in the manifest are ready. I'm not too familiar with Helm's codebase, but if I understand correctly, is the idea of this PR to query the manifest and build a kube client to query each resource?

Yes @Jont828 like you said the idea of this PR to query the manifest and build a kube client to query each resource

pkg/kube/client.go
SoujanyaMangipudi: @mattfarina sorry for repeated pings like I mentioned before we wanted this feature to get merged into helm 3.10 release. Hence was wondering if you take a look at latest changes?

mattfarina (approved):
Note, there is cleanup that would need to be done after this to restore k8s 1.25 support. I'm willing to make the change for that prior to cutting the release.

SoujanyaMangipudi:
@mattfarina thank you so much for approving my PR. Regarding next steps as I don't have permissions to merge the PR will you be able to help me with that?

SoujanyaMangipudi:
> Hi @SoujanyaMangipudi Looks like we now have merge conflicts that need to be resolved on this PR. Thanks.

@bridgetkromhout I resolved all the merge conflicts now

mattfarina:
@SoujanyaMangipudi I came to review this again only to find the go.sum merge conflict again.

SoujanyaMangipudi:
> @SoujanyaMangipudi I came to review this again only to find the go.sum merge conflict again.

@mattfarina I resolved the conflicts again. Can you please re-trigger the workflows?

scottrigby (approved):
This all looks fantastic, great work @SoujanyaMangipudi 👍 

As a side note, there was an issue reported in the PR that this one was based on (#7728) related to helm rollback. It wasn't clear how that PR would affect helm rollback, and more info was requested by the reporter but never given. For good measure here, I've tried but haven't been able to reproduce that issue here.

SoujanyaMangipudi:
> This all looks fantastic, great work @SoujanyaMangipudi 👍
> 
> As a side note, there was an issue reported in the PR that this one was based on (#7728) related to helm rollback. It wasn't clear how that PR would affect helm rollback, and more info was requested by the reporter but never given. For good measure here, I've tried but haven't been able to reproduce that issue here.

Thank you @scottrigby! We now moved the new functionality behind a --show-resources flag for helm status command and also created a new InterfaceResources extension to avoid breaking changes. I believe these two changes handled the original issue reported.

mattfarina (approved):

PR #10914, opened 2022-04-29
feat(helm): add chart name and version in json/yaml output

yong-jie-gong:
…yaml

currently “helm list” merge the char name and version as single field chart. but sometimes, we need know the chart name/version one  installed release is referring  to

**If applicable**:
- [ Y] this PR has been tested for backwards compatibility


yxxhero:
@yong-jie-gong can you add some unittest? thanks very much.

yong-jie-gong:
> @yong-jie-gong can you add some unittest? thanks very much.

@yxxhero  unitests have been submitted

cmd/helm/list.go
```diff
@@ -153,12 +155,14 @@ func newReleaseListWriter(releases []*release.Release, timeFormat string) *relea
 	elements := make([]releaseElement, 0, len(releases))
 	for _, r := range releases {
 		element := releaseElement{
-			Name:       r.Name,
-			Namespace:  r.Namespace,
-			Revision:   strconv.Itoa(r.Version),
-			Status:     r.Info.Status.String(),
-			Chart:      formatChartname(r.Chart),
-			AppVersion: formatAppVersion(r.Chart),
```
yxxhero, line 161: why not use formatAppVersion?

yxxhero:
@yong-jie-gong 
The chart field already contains all the information. Can you get the information you want through program processing? In addition, you need to see formatapversion and formatchartname.

cmd/helm/list.go
yong-jie-gong: what i want to know is the chart version instead of appVersion. but currenlty it is not convenient to know. 
for example, for helm list command output are below
#helm list -oyaml
- app_version: 1.16.0
  chart: abc-hello-0.1.0+rc-1.2.0-19
  name: test1
  namespace: default
  revision: "1"
  status: deployed
  updated: 2022-06-15 02:44:41.606385155 -0700 PDT

with output above, it is hard to know what chart versions. i have write speical logic to parse from chart name

alexmg:
I found this issue after running into the same problem when working with the CLI and would love to see this addressed. When using `helm upgrade` to update values you need to provide the current version to prevent an upgrade to the latest version of the chart. The fact that client-side parsing is required to extract the current version makes the interaction clumsy. You provide a version in the initial command and then it is no longer available in the format required for subsequent commands.

cmd/helm/list.go
```diff
@@ -153,12 +155,14 @@ func newReleaseListWriter(releases []*release.Release, timeFormat string) *relea
...
+			Name:         r.Name,
+			Namespace:    r.Namespace,
+			Revision:     strconv.Itoa(r.Version),
+			Status:       r.Info.Status.String(),
+			Chart:        fmt.Sprintf("%s-%s", r.Chart.Metadata.Name, r.Chart.Metadata.Version),
```
yxxhero, line 162: Chart: formatChartname(r.Chart),

cmd/helm/list.go
```diff
@@ -153,12 +155,14 @@ func newReleaseListWriter(releases []*release.Release, timeFormat string) *relea
...
+			ChartName:    r.Chart.Metadata.Name,
+			ChartVersion: r.Chart.Metadata.Version,
+			AppVersion:   r.Chart.Metadata.AppVersion,
```
yxxhero, line 165: AppVersion: formatAppVersion(r.Chart),

yxxhero:
1. DCO.
2. sqush your commits.
3. go.mod don't need change.

cmd/helm/list.go
yxxhero: need to fix.

PR #10916, opened 2022-04-29
Add `WithIfModifiedSince` to enable use of HTTP caching

pjbgf:

The HTTP Getter can leverage the use of [RFC7234 4.3.2](https://datatracker.ietf.org/doc/html/rfc7234#section-4.3.2) by setting the
HTTP Header 'If-Modified-Since'. This will enable callers to avoid
downloading index files when they haven't been modified since the
last get operation.

**If applicable**:

### DIFF
--- a/pkg/getter/getter.go
+++ b/pkg/getter/getter.go
@@ -45,6 +45,7 @@ type options struct {
 	registryClient        *registry.Client
 	timeout               time.Duration
 	transport             *http.Transport
+	ifModifiedSince       *time.Time
 }
 
 // Option allows specifying various settings configurable by the user for overriding the defaults
@@ -118,6 +119,20 @@ func WithRegistryClient(client *registry.Client) Option {
 func WithUntar() Option {
 	return func(opts *options) {
 		opts.unTar = true
+	}
+}
+
+// WithIfModifiedSince leverages the HTTP caching mechanism
+// defined on RFC7234 4.3.2. It prevents the HTTP getter from
+// downloading index that haven't been modified based on `since`.
+// If the file was not modified, getter will return a
+// `304 Not Modified`, as per RFC.
+//
+// The caller is responsible for caching previously downloaded
+// index files and handling the returning error.
+func WithIfModifiedSince(since time.Time) Option {
+	return func(opts *options) {
+		opts.ifModifiedSince = &since
 	}
 }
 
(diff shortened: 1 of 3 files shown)


yxxhero:
@pjbgf I don't see any function calls related to withifmodifiedsince. Thanks very much.

bacongobbler:
Agreed. This is nice, but I don’t see how this improves the experience for helm users as this code isn’t being used by the client. Is there a way we can integrate this with vanilla chart repositories?

yxxhero:
@pjbgf ping

pjbgf:
@yxxhero @bacongobbler thank you very much for reviewing this.

This change is mostly targeted at folks that rely on the helm SDK, rather than its binary. Here's an example of its use on an [open Flux PR](https://github.com/fluxcd/source-controller/pull/700):
https://github.com/fluxcd/source-controller/pull/700/commits/d67f15edb3793737738372961d916d0856b92852

At present, the cost of downloading and loading a large repository index (i.e. [bitnami](https://charts.bitnami.com/bitnami/index.yaml)) can translate to spikes of hundreds of MBs in memory usage. The main bottleneck being the YAML unmarshalling. To be able to run this process in parallel requires a pod to be running on a cgroup with very generous Memory limits (north of 1Gb), otherwise it will out of memory and therefore be killed.

We are looking into ways of optimising this process, and one of the most obvious quick wins is to short-circuit it before even downloading the chart index, saving on memory, CPU and network bandwidth usage. 

To be honest, I am not sure this would be a great fit to integrate with the helm binary. 
Would this be a blocker for this PR?

yxxhero:
@pjbgf oh. I known. Thanks very much. 
What do you think of? @bacongobbler

bacongobbler:
IMO I'm against this change because the only beneficiaries are downstream packages. No users of the Helm client benefit from this change, and we are left on the hook to maintain support for a third-party feature. If `WithIfModifiedSince` support was provided in [chartmuseum](https://github.com/helm/chartmuseum), I'd be more in favour of this change.

At a higher level, most of the project's recent efforts have been towards [OCI integration](https://helm.sh/blog/storing-charts-in-oci/). OCI registry support went GA in Helm 3.8. This feature goes against that initiative as it provides support for chart repositories (more specifically, support for a specific vendor's Helm client) but not OCI registies. This begs the question: why are we continuing to provide support for APIs we eventually plan to deprecate?

Curious on other core maintainer's thoghts here.

PR #10926, opened 2022-05-03
Fix --untardir reference in --destination description

SimonAlling:


SimonAlling:
The reference fixed in this PR has survived ever since it was added in 593718d74910c6d180b05b52e97cc20c8bb18d66. :ninja:

mattfarina (approved):

PR #10965, opened 2022-05-16
Adding details on the Factory interface

mattfarina:
This covers both the property and the minimal copy of the Factory
interface. It also notes that this interface is not covered by the
Helm backwards compatibility and why.

**What this PR does / why we need it**: There is confusion around what is covered by the Helm backwards compatibility. This helps clear it up with regard to this interface.

**If applicable**:

### DIFF
--- a/pkg/kube/factory.go
+++ b/pkg/kube/factory.go
@@ -26,6 +26,12 @@ import (
 
 // Factory provides abstractions that allow the Kubectl command to be extended across multiple types
 // of resources and different API sets.
+// This interface is a minimal copy of the kubectl Factory interface containing only the functions
+// needed by Helm. Since Kubernetes Go APIs, including interfaces, can change in any minor release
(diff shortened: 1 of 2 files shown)


pkg/kube/factory.go
```diff
@@ -27,6 +27,12 @@ import (
 
 // Factory provides abstractions that allow the Kubectl command to be extended across multiple types
 // of resources and different API sets.
+// This interface is a minimal copy of the kubectl Factory interface containering only the functions
```
hickeyma, line 30: containering --> containing

hickeyma (dismissed):
Small nit

pkg/kube/factory.go
joejulian: @hickeyma I pushed that typo fix

joejulian (approved):

PR #11003, opened 2022-05-30
feat: updateable hooks

### LINKED ISSUE #10987: Proposal: infra hooks
With projects like https://crossplane.io/ or https://kubernetes.io/docs/concepts/extend-kubernetes/service-catalog/ and any other CR that defines a resource that needs to exist before the helm release is installed or upgraded, it would be nice to have a way to ensure thhese CR are created or updated before the helm release, this would allow us to deploy app external dependencies, such as DBs, for example, as hooks so we can, for example, ensure that a DB is provisioned before running the DB migration pre-install/pre-upgrade hook.

This would act as a regular hook but on upgrades, instead of replacing the resouce or complaining that the resource already exists helm would just update it in place.

With the available hooks we have  to maintain the app's infra dependencies on a separate helm release and ensure that one is installed/uppgraded before the app's release.

Using the default `"helm.sh/hook-delete-policy": before-hook-creation` would lead to the DB being deleted and re-created on each upgrade, definetely not what we want and using `"helm.sh/hook-delete-policy": hook-faild` blocks any upgrade because the hook resoruce already exists.

another way of solving this could also be something like what's proposed in https://github.com/helm/helm/issues/8439 but a hooks would probably be better, like in the following scenario:

> The chart has a hook for provisioning a DB in the cloud provier, this hook is weighted so it's the very first to run, the chart also has a DB migration hook that runs a k8s Job to preform the DB migration, this job can have an init container to wait for the DB to be ready and ensure the script in the main container can connect to the DB.

I co0uld open a PR to implement this if you think the proposal makes sense (suggestions rgarding how to name the hook would be welcomed).

luisdavim:

This should fix: #10987, instead of just deleting hooks before creation or just failing if the resource already exists allow the hook to be updated if the right labels are present.

- This implements a previously existing `TODO` comment: https://github.com/helm/helm/blob/a499b4b179307c267bdf3ec49b880e3dbd2a5591/pkg/action/hooks.go#L47:L50
- This can be a breaking change for users that don't explicitly set the hook deletion policies and/or don't set the helm release ownership labels, the new code will add the labels but for upgrades it can be a problem that we have to document, adding the right labels to the existing resources or explicitly setting a deletion policy will fix the issue, so it's not a blocker but it might require manual intervention. 
- An alternative approach could be to create an `UpdateInPlace` function that just takes one `ResourceList` and creates or updates the resources regardless of the ownership labels, I've considered that option, we could use the object returned by the call that is made to check for the existence of the resource to create the patch, but that could be potentially more dangerous than just failing the release and require manual intervention to move forward
- ~~I've added an additional commit with some linter fixes, I guess I'm running a newer version of the `golangci-lint` tool and it complained about some issues unrelated to my changes.~~ update: no longer needed after rebase
- There might be other resource types that should have the delete before creation policies by default, but at the moment I can only think of `Jobs` and `Pods` being the ones that should have it for sure, mainly because of immutability, if you think of other resources that might not work well with updates, let me know and I can update the PR.

**If applicable**:

### DIFF
--- a/pkg/action/action_test.go
+++ b/pkg/action/action_test.go
@@ -74,6 +74,7 @@ metadata:
   name: test-cm
   annotations:
     "helm.sh/hook": post-install,pre-delete,post-upgrade
+    "helm.sh/hook-delete-policy": before-hook-creation
 data:
   name: value`
 
@@ -810,6 +811,7 @@ metadata:
   name: test-cm-postrendered
   annotations:
     "helm.sh/hook": post-install,pre-delete,post-upgrade
+    "helm.sh/hook-delete-policy": before-hook-creation
 data:
   name: value`
 
--- a/pkg/action/hooks.go
+++ b/pkg/action/hooks.go
@@ -44,12 +44,13 @@ func (cfg *Configuration) execHook(rl *release.Release, hook release.HookEvent, 
 		}
 	}
 
-	// hooke are pre-ordered by kind, so keep order stable
+	// hooks are pre-ordered by kind, so keep order stable
 	sort.Stable(hookByWeight(executingHooks))
 
 	for i, h := range executingHooks {
 		// Set default delete policy to before-hook-creation
 		cfg.hookSetDeletePolicy(h)
+		setDefaultDeletePolicy(h)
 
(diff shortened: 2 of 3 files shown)


luisdavim:
Not stale, I'll rebase it.

PR #11022, opened 2022-06-06
Add proper support for CRD 'crds/' templating & support for CRDs in 'templates/' folder

inteon:
Currently CRDs are too much of an afterthought. Keeping in mind <https://github.com/helm/community/blob/main/hips/hip-0011.md>, there are still quite a few improvements that can be made to Helm with regards to installing CRDs.

- adds support for templating crds in the 'crds/' folder
- filters crds in the 'templates/' folder and installs these resources first, also makes sure that the "discoveryClient.Invalidate" function is called after installing these CRDs

**Context**:
A request that we get a lot from cert-manager Helm Chart users is to move our CRDs to the "crds/" folder. However, we cannot do this because there are some templated fields in the CRDs (like labels & conversion webhook configs). Instead, we place the CRDs in the "templates/" folder, which makes it impossible to also add CRs in the Helm chart or in charts that use cert-manager as a dependency.
This PR proposes two fixes for the problem: 1) make resources in "crds/" folder templatable and 2) support CRDs in the "templates/" folder

Feel free to post some comments/ thoughts on these improvements. 

**If applicable**:
- [ ] this PR contains documentation // TODO

### DIFF
--- a/cmd/helm/testdata/output/template-with-crds.txt
+++ b/cmd/helm/testdata/output/template-with-crds.txt
@@ -1,5 +1,5 @@
 ---
-# Source: crds/crdA.yaml
+# Source: subchart/crds/crdA.yaml
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
@@ -14,7 +14,6 @@ spec:
     shortNames:
       - tc
     singular: authconfig
-
 ---
 # Source: subchart/templates/subdir/serviceaccount.yaml
 apiVersion: v1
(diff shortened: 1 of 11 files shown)


rufreakde:
TBH having the ability to only use a subset for templating like in the `metadata.annotations` and `metadata.labels` would be already a very good improvement. Even if it would not be by real templating but maybe just something like `--crd-annotation x:y` or `--crd-label x:y` in the CLI parameters.

The full templating support might cause a lot of issues to think about. But Annotations and Labels would improve a lot of Declarative Pipeline Tools Usages with less risk for issues.

joejulian:
This would need a PR to hip-0011 to be approved first.

PR #11031, opened 2022-06-08
Fail if could not retrieve latest release tag from github

shoce:
Happened to me when trying to install helm on a server located in China. Github was blocked so `curl` could not get a page from github, so the $TAG was empty and after that the URL for installing helm was missing version but the error was not clear at all as it was difficult to notice that the URL is missing the version. Had to look into the script code to figure out what went wrong. I propose to fail the script early if the version info could not be retrieved from github. 

Signed-off-by: Sho Ce <261885+shoce@users.noreply.github.com>

**If applicable**:

### DIFF
--- a/scripts/get-helm-3
+++ b/scripts/get-helm-3
@@ -108,11 +108,17 @@ verifySupported() {
 checkDesiredVersion() {
   if [ "x$DESIRED_VERSION" == "x" ]; then
     # Get tag from release URL
-    local latest_release_url="https://github.com/helm/helm/releases"
+    local latest_release_url="https://api.github.com/repos/helm/helm/releases/latest"
+    local latest_release_response=""
     if [ "${HAS_CURL}" == "true" ]; then
-      TAG=$(curl -Ls $latest_release_url | grep 'href="/helm/helm/releases/tag/v3.[0-9]*.[0-9]*\"' | sed -E 's/.*\/helm\/helm\/releases\/tag\/(v[0-9\.]+)".*/\1/g' | head -1)
+      latest_release_response=$( curl -L --silent --show-error --fail "$latest_release_url" 2>&1 || true )
     elif [ "${HAS_WGET}" == "true" ]; then
-      TAG=$(wget $latest_release_url -O - 2>&1 | grep 'href="/helm/helm/releases/tag/v3.[0-9]*.[0-9]*\"' | sed -E 's/.*\/helm\/helm\/releases\/tag\/(v[0-9\.]+)".*/\1/g' | head -1)
+      latest_release_response=$( wget "$latest_release_url" -O - 2>&1 || true )
+    fi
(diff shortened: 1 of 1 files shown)


joejulian (approved):

scripts/get-helm-3
```diff
@@ -114,6 +114,10 @@ checkDesiredVersion() {
     elif [ "${HAS_WGET}" == "true" ]; then
       TAG=$(wget $latest_release_url -O - 2>&1 | grep 'href="/helm/helm/releases/tag/v3.[0-9]*.[0-9]*\"' | sed -E 's/.*\/helm\/helm\/releases\/tag\/(v[0-9\.]+)".*/\1/g' | head -1)
     fi
+    if [ "x$TAG" == "x" ]; then
```
joejulian, line 117: I tested this out and got the error, but in doing so I realized that this error doesn't tell me what went wrong. There's no way for me to guess how to fix this.

What if we capture the error and display it if there's a failure:
```bash
    if [ "${HAS_CURL}" == "true" ]; then
        {
            IFS=$'\n' read -r -d '' ERR;
            IFS=$'\n' read -r -d '' TAG;
        } < <((printf '\0%s\0' "$(curl -Ls --fail --show-error $latest_release_url | grep 'href="/helm/helm/releases/tag/v3.[0-9]*.[0-9]*"' | sed -E 's/.*\/helm\/helm\/releases\/tag\/(v[0-9\.]+)".*/\1/g' | head -1)" 1>&2) 2>&1)
    elif [ "${HAS_WGET}" == "true" ]; then
        {
            IFS=$'\n' read -r -d '' ERR;
            IFS=$'\n' read -r -d '' TAG;
        } < <((printf '\0%s\0' "$(wget $latest_release_url -O - | grep 'href="/helm/helm/releases/tag/v3.[0-9]*.[0-9]*"' | sed -E 's/.*\/helm\/helm\/releases\/tag\/(v[0-9\.]+)".*/\1/g' | head -1)" 1>&2) 2>&1)
    fi
    if [ "$TAG" == "" ]; then
      printf "Could not retrieve the latest release tag information from %s: %s.\n" "${latest_release_url}" "${ERR}"
      exit 1
    fi
```

joejulian (changes requested):
Let's try to make the error message give at least a hint to what the problem is.

scripts/get-helm-3
shoce: Thanks! Your implementation seemed too complicated to me so I have implemented it simpler, hope you like it.

shoce: I also switched to using Github JSON API to avoid parsing html with grep+sed, json response would be probably more stable.

shoce:
Sorry i had troubles with merging origin main so had to rebase.

shoce:
I am not sure if i need to press any button to signal this one is ready for another review, there is a badge saying "1 change requested", i could not find how to mark it done. Just saying there is a change ready for review. Thanks!

joejulian (approved):

PR #11071, opened 2022-06-22
Added support for Endpoints to be sorted properly

drivera-armedia:

In some specific use cases one must manually create endpoints in order to support externally-provided services that lack a hostname (i.e. only an IP address). This can be done by manually creating the endpoint that the service will eventually bind to, configured in a special manner. In order for this to work, the Endpoints object(s) must be installed prior to the Service objects that consume them, and then eventually un-installed in reverse order.

This addresses [issue # 11145](https://github.com/helm/helm/issues/11145).

**If applicable**:


drivera-armedia:
Please help: the CI tests fail with this error, which have nothing to do (as far as I can tell) with the code being tested but rather seems to be an environmental issue

```
2022/11/24 16:01:39 http: TLS handshake error from 127.0.0.1:43576: read tcp 127.0.0.1:33057->127.0.0.1:43576: use of closed network connection
--- FAIL: TestDownloadTo_TLS (0.01s)
    chart_downloader_test.go:253: Get "https://127.0.0.1:33057/signtest-0.1.0.tgz": x509: certificate has expired or is not yet valid: current time 2022-11-24T16:01:39Z is after 2022-08-21T22:36:36Z
FAIL
```

yxxhero:
@drivera-armedia need to rebase

PR #11086, opened 2022-06-27
Make token caching an opt in feature in OCI registry client

souleb:
Signed-off-by: Soule BA <bah.soule@gmail.com>

If implemented, caching credential token is an opt in feature in oci registry client. This client is part of the Helm SDK, and consumers sometime do not wish to cache credentials.

**If applicable**:


scottrigby (approved):
This fix makes sense to me for SDK users. 

No functionality is changed for the Helm CLI – PR passes tests, and I also did manual testing from this PR branch with `helm push` and `helm pull`. There is no way I know of to manually break this caching option using the CLI, only through using the SDK.

IMO from an SDK user perspective, this should be considered a bug fix. SDK users likely don't know the current code automatically caches the token. This is not desirable behavior without an opt-in, considering the linked issue (for example when using auto-rotation best practice in ERC and other registries).

mattfarina (commented):
/lgtm

scottrigby:
@mattfarina update on that question – the SDK change was tested using Flux to pull in this PR branch. What happens is that the cache is not set and the token is read from credentials file every time 👍 

For future ref for others – size/S PRs need 2 maintainer approvals, so Matt looked over this retroactively. Only size/XS needs one approval. I knew this but mixed it up. Thanks Matt

PR #11087, opened 2022-06-27
WIP: Initial code for HIP Support Helm release data stored in multiple K8s…

ralfv:
… Secrets PR https://github.com/helm/community/pull/256

This is an initial unpolished implementation for HIP https://github.com/helm/community/pull/256 "HIP Support Helm release data stored in multiple K8s Secrets".

It allows Helm releases to be stored in multiple K8s Secrets if the size is > (1024*1000) bytes. For more details see the HIP.

**If applicable**:

Signed-off-by: Ralf Van Dorsselaer <ralfvandorsselaer@gmail.com>

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -39,7 +39,7 @@ jobs:
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@1813ca74c3faaa3a2da2070b9b8a0b3e7373a0d8 # pinv2.21.0
+      uses: github/codeql-action/init@0ba4244466797eb048eb91a6cd43d5c03ca8bd05 # pinv2.21.2
       with:
         languages: ${{ matrix.language }}
         # If you wish to specify custom queries, you can do so here or in a config file.
@@ -50,7 +50,7 @@ jobs:
     # Autobuild attempts to build any compiled languages  (C/C++, C#, or Java).
     # If this step fails, then you should remove it and run the build manually (see below)
     - name: Autobuild
-      uses: github/codeql-action/autobuild@1813ca74c3faaa3a2da2070b9b8a0b3e7373a0d8 # pinv2.21.0
+      uses: github/codeql-action/autobuild@0ba4244466797eb048eb91a6cd43d5c03ca8bd05 # pinv2.21.2
 
     # ℹ️ Command-line programs to run using the OS shell.
     # 📚 https://git.io/JvXDl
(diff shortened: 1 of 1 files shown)


pkg/storage/driver/secrets.go
```diff
@@ -14,10 +14,23 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+// Supports both single and multiple Secrets to hold a Helm release.
+
+// TODO
+// Arch
+//   - Consider implementing chunk support as v2 driver?
```
liorkesos, line 21: +1

sciornei-dox:
@ralfv as pre last week meeting, (notes https://docs.google.com/document/d/1d-6xJEx0C78csIYSPKJzRPeWaHG_8W1Hjl72OJggwdc ), should you be still interested in this PR, what would be acceptable is a new HELM_DRIVER for storage (e.g. multisecrets ).

ralfv:
> @ralfv as pre last week meeting, (notes https://docs.google.com/document/d/1d-6xJEx0C78csIYSPKJzRPeWaHG_8W1Hjl72OJggwdc ), should you be still interested in this PR, what would be acceptable is a new HELM_DRIVER for storage (e.g. multisecrets ).

OK that's great news. We'll work on the PR over the next month or two (summer holidays permitting), as my POC code needs cleanup and tests added plus separation into new storage driver (I had it implemented initially like that).

I think we would still prefer that the code be merged into the existing secrets driver, but we are already very happy that you considered this PR for merging, as some of our charts are close the size limit.

Thanks!

sciornei-dox:
Thank you @ralfv for the updates.
Would it be possible to have the implementation BWC, so one will require to use the new HELM_DRIVER only when interacting with deployments which had to split the secrets?
Meaning: if a deployment was created with the native storage driver, would we be able to manage it with the new HELM_DRIVER, while still keeping it bwc with the native storage driver (as long as the size was small and did not require split) ?

ralfv:
> Thank you @ralfv for the updates.
> Would it be possible to have the implementation BWC, so one will require to use the new HELM_DRIVER only when interacting with deployments which had to split the secrets?
> Meaning: if a deployment was created with the native storage driver, would we be able to manage it with the new HELM_DRIVER, while still keeping it bwc with the native storage driver (as long as the size was small and did not require split) ?

Sure. The current POC code in this PR already does this if the compressed data size is <= 1MB. It currently always adds some chunk metadata into the K8s Secret that the regular secrets driver just ignores, but we can just as well remove that and thus generate a K8s Secret that is 100% the same as they are today if splitting is not required.

It still seems odd to have the splitting only be active explicitly when using a new value like multisecrets for the HELM_DRIVER environment variable. With code that generates BWC Helm secrets if not splitting is required, it seems to me that an additional driver is not needed. What was the main reason for suggesting a separate driver? It is not evident what the rationale was when reading through the developer meeting notes.

Also, one other option (although currently unexplored) is if a PR for K8s Secrets splitting is better directed towards the K8s project, so that for Helm it essentially becomes transparent.

Thanks,
Ralf

ralfv:
Oops - did not mean to close this PR now - working on new one.

ralfv:
NEW PR for split Helm secrets. Still POC code, but this time implemented as HELM_DRIVER=multisecret | multisecrets.

https://github.com/helm/helm/pull/12277

PR #11090, opened 2022-06-28
feat: expose repo add as action

johanneswuerbach:

We would like to add helm repositories from code and currently need to copy large parts of `cmd/helm/repo_add.go` as the functionality isn't exposed. 

This PR moves the respective adding logic into an action (like `action.Install`) so it can be used by 3rd parties more easily.

**If applicable**:

### DIFF
--- a/cmd/helm/repo_add.go
+++ b/cmd/helm/repo_add.go
@@ -17,56 +17,16 @@ limitations under the License.
 package main
 
 import (
-	"context"
-	"fmt"
 	"io"
-	"io/ioutil"
-	"os"
-	"path/filepath"
-	"strings"
-	"time"
 
-	"github.com/gofrs/flock"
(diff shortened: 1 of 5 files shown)


mattfarina (commented):
Thanks for the PR.

Could you please add tests for the action. A `repo_add_test.go` file with some tests.

There are repo list, update, and remove commands. What are your thoughts on those being actoins, too? Will you need those?

johanneswuerbach:
Thanks for the review @mattfarina.

I've move the tests that only test the action to `pkg/action/repo_add_test.go` and left the tests testing action + cmd in `cmd/helm/repo_add_test.go`. Do you think this is sufficient to ensure coverage?

For our use case we only need repo add, so for us exposing those actions isn't really necessary.

PR #11102, opened 2022-06-30
bugfix: helm install --wait doesn't work

yudong2015:
Signed-off-by: jackyu <jackyu@yunify.com>

**If applicable**:


pkg/kube/wait.go
```diff
@@ -50,7 +50,7 @@ func (w *waiter) waitForResources(created ResourceList) error {
 	ctx, cancel := context.WithTimeout(context.Background(), w.timeout)
 	defer cancel()
 
-	return wait.PollImmediateUntil(2*time.Second, func() (bool, error) {
```
phil9909, line 53: I guess this does not really fix the problem, it only makes it less likely to occur.
It delays the initial check by 2 seconds, which might work, but might also fail, if the kubernetes api server is under load.

A proper solution would probably require checking the [`Status.ObservedGeneration`](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.24/#daemonsetstatus-v1-apps) field of the DaemonSet. Probably somewhere in the [`daemonSetReady` function](https://github.com/helm/helm/blob/79cd4dda8124d905ce95710609b35d050b11e4af/pkg/kube/ready.go#L283)

bacongobbler:
duplicate of https://github.com/helm/helm/pull/10831?

pkg/kube/wait.go
yudong2015: ok. Update: check Status.ObservedGeneration in daemonSetReady function,  ds.Status.ObservedGeneration==0 means the resource not generated or not loaded, return false;

yudong2015:
> duplicate of #10831?

#10831 doesn't fix bug of waiting DaemonSet ready, if apiserver doesn't load, ds.Status.DesiredNumberScheduled is 0 and daemonSetReady func return true.

njuptlzf:
It's still an issue and I don't understand why the maintainers aren't interested.

TerryHowe:
This code has changed a lot with v4 kstatus support. If this is still relevant for v4, please create a new pr, thanks.

PR #11110, opened 2022-07-03
bump Go 1.18

yxxhero:
Signed-off-by: yxxhero <aiopsclub@163.com>

**What this PR does / why we need it**:
bump Go 1.18 https://github.com/helm/helm/issues/11106
**Special notes for your reviewer**:
bump Go 1.18
**If applicable**:
- [ ] this PR contains documentation
- [ ] this PR contains unit tests
- [x] this PR has been tested for backwards compatibility

PR #11119, opened 2022-07-06
fix: use `go install` instead of `go get`

bacongobbler:
There were a few other locations where `go get` was being used, which was missed in #11110. This fixes the current issues in CI.


yxxhero (approved):
/lgtm

hickeyma (approved):

mattfarina (approved):

PR #11125, opened 2022-07-08
bugfix: daemonSetReady doesn't work

jmc-9304:
Signed-off-by: mincheol_jeon fdop104@gmail.com

If daemonSetReady() is called immediately after daemonset is created, 
`ds.Status.DesiredNumberScheduled is set to 0 -> expectedReady is set to negative`. 

So **true is unconditionally returned.**

I could be wrong, but here's the reason this is happening
 - daemonSetReady()  is called before ds.Status.DesiredNumberScheduled is calculated by scheduler

**If applicable**:

### DIFF
--- a/pkg/kube/ready.go
+++ b/pkg/kube/ready.go
@@ -300,7 +300,7 @@ func (c *ReadyChecker) daemonSetReady(ds *appsv1.DaemonSet) bool {
 	}
 
 	expectedReady := int(ds.Status.DesiredNumberScheduled) - maxUnavailable
-	if !(int(ds.Status.NumberReady) >= expectedReady) {
+	if !(int(ds.Status.NumberReady) >= expectedReady) || expectedReady < 0 {
 		c.log("DaemonSet is not ready: %s/%s. %d out of %d expected pods are ready", ds.Namespace, ds.Name, ds.Status.NumberReady, expectedReady)
 		return false
 	}


mattfarina:
@jmc-9304 could you please file an issue with details about the problem

pkg/kube/ready.go
```diff
@@ -300,7 +300,7 @@ func (c *ReadyChecker) daemonSetReady(ds *appsv1.DaemonSet) bool {
 	}
 
 	expectedReady := int(ds.Status.DesiredNumberScheduled) - maxUnavailable
-	if !(int(ds.Status.NumberReady) >= expectedReady) {
+	if !(int(ds.Status.NumberReady) >= expectedReady) || expectedReady < 0 {
```
mattfarina, line 303: `DesiredNumberScheduled` is not calculated until it's taken off the queue and processed. There is a potential for a race condition here. If the `maxUnavailable > 0` and `DesiredNumberScheduled = 0` then `NumberReady` of 0 would be > `expectedReady` (which would be negative as 0 - maxUnavailable). This case would fall in to the `if` statement.

But, if `maxUnavailable` is set to 0 you have 0 desired number scheduled - 0 max unavailable equaling 0 or more as the number ready.

I don't think this completely fixes the bug because it shows up when `expectedReady` is 0. As in both `DesiredNumberScheduled` and `maxUnavailable` are 0.

mattfarina (changes requested):
Thanks for finding this issue. I see the problem but, I think, the logic needs to be changed to handle all of the cases.

joejulian:
closing this as abandoned. Please re-open it if you can address the change request.

PR #11160, opened 2022-07-18
Fixing x509 test on darwin

### LINKED ISSUE #11159: Go tests failing on Mac due to inconsistent Go error message
When running the test suite on mac I get the following error:

```
2022/07/18 15:15:15 http: TLS handshake error from 127.0.0.1:55752: remote error: tls: bad certificate
--- FAIL: TestFindChartInAuthAndTLSAndPassRepoURL (0.02s)
    chartrepo_test.go:314: Expected TLS error for function  FindChartInAuthAndTLSAndPassRepoURL not found, but got a different error (looks like "https://127.0.0.1:55750" is not a valid chart repository or cannot be reached: Get "https://127.0.0.1:55750/index.yaml": x509: “Acme Co” certificate is not trusted)
```

This is due to the mac platform provider having a different response from other platforms. In the golang source they even test for different messages from [mac](https://github.com/golang/go/blob/go1.18.4/src/crypto/x509/root_darwin_test.go#L56) and [windows](https://github.com/golang/go/blob/go1.18.4/src/crypto/x509/root_windows_test.go#L68).

Output of `helm version`:
tip of main (developing for 3.10.0). Commit 937798868518b7c5dbcbefac5bfe0b2e045e3bae

Cloud Provider/Platform (AKS, GKE, Minikube etc.):  N/A

mattfarina:
Go passes x509 verification off to the platform and different
platforms provide different responses. The Go tests for x509
even have different test files for different platform providers
that check for different messages.

This update haldes darwins difference for x509 authority handling

Closes #11159

Signed-off-by: Matt Farina <matt@mattfarina.com>

**If applicable**:

### DIFF
--- a/pkg/repo/chartrepo_test.go
+++ b/pkg/repo/chartrepo_test.go
@@ -24,6 +24,7 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
+	"runtime"
 	"strings"
 	"testing"
 	"time"
@@ -309,8 +310,15 @@ func TestFindChartInAuthAndTLSAndPassRepoURL(t *testing.T) {
 
 	// If the insecureSkipTLsverify is false, it will return an error that contains "x509: certificate signed by unknown authority".
 	_, err = FindChartInAuthAndTLSAndPassRepoURL(srv.URL, "", "", "nginx", "0.1.0", "", "", "", false, false, getter.All(&cli.EnvSettings{}))
-
-	if !strings.Contains(err.Error(), "x509: certificate signed by unknown authority") {
+	// Go communicates with the platform and different platforms return different messages. Go itself tests darwin
+	// differently for its message. On newer versions of Darwin the message includes the "Acme Co" portion while older
+	// versions of Darwin do not. As there are people developing Helm using both old and new versions of Darwin we test
+	// for both messages.
+	if runtime.GOOS == "darwin" {
+		if !strings.Contains(err.Error(), "x509: “Acme Co” certificate is not trusted") && !strings.Contains(err.Error(), "x509: certificate signed by unknown authority") {
+			t.Errorf("Expected TLS error for function  FindChartInAuthAndTLSAndPassRepoURL not found, but got a different error (%v)", err)
+		}
+	} else if !strings.Contains(err.Error(), "x509: certificate signed by unknown authority") {
 		t.Errorf("Expected TLS error for function  FindChartInAuthAndTLSAndPassRepoURL not found, but got a different error (%v)", err)
 	}
 }


pkg/repo/chartrepo_test.go
```diff
@@ -309,8 +310,9 @@ func TestFindChartInAuthAndTLSAndPassRepoURL(t *testing.T) {
 
 	// If the insecureSkipTLsverify is false, it will return an error that contains "x509: certificate signed by unknown authority".
 	_, err = FindChartInAuthAndTLSAndPassRepoURL(srv.URL, "", "", "nginx", "0.1.0", "", "", "", false, false, getter.All(&cli.EnvSettings{}))
-
-	if !strings.Contains(err.Error(), "x509: certificate signed by unknown authority") {
+	if runtime.GOOS == "darwin" && !strings.Contains(err.Error(), "x509: “Acme Co” certificate is not trusted") {
+		t.Errorf("Expected TLS error for function  FindChartInAuthAndTLSAndPassRepoURL not found, but got a different error (%v)", err)
+	} else if runtime.GOOS != "darwin" && !strings.Contains(err.Error(), "x509: certificate signed by unknown authority") {
 		t.Errorf("Expected TLS error for function  FindChartInAuthAndTLSAndPassRepoURL not found, but got a different error (%v)", err)
 	}
```
phil9909, line 317: How about?

```suggestion
	expectedError := "x509: certificate signed by unknown authority"
	if runtime.GOOS == "darwin" {
		expectedError = "x509: “Acme Co” certificate is not trusted"
	}
	if !strings.Contains(err.Error(), expectedError) {
		t.Errorf("Expected TLS error for function  FindChartInAuthAndTLSAndPassRepoURL not found, but got a different error (%v)", err)
	}
```

mattfarina: done

sabre1041:
On OSX Catalina (10.15.7)

Existing codebase passes

Proposed change fails:

```
=== RUN   TestFindChartInAuthAndTLSAndPassRepoURL
2022/07/21 19:30:19 http: TLS handshake error from 127.0.0.1:52902: remote error: tls: bad certificate
    chartrepo_test.go:320: Expected TLS error for function  FindChartInAuthAndTLSAndPassRepoURL not found, but got a different error (looks like "https://127.0.0.1:52900" is not a valid chart repository or cannot be reached: Get "https://127.0.0.1:52900/index.yaml": x509: certificate signed by unknown authority)
```
It appears that the change appears only in more recent versions of OSX. Should both be supported or only the logic found in recent versions of OSX?

scottrigby (commented):
> @sabre1041 It appears that the change appears only in more recent versions of OSX. Should both be supported or only the logic found in recent versions of OSX?

My gut leans toward supporting a reasonable number of previous MacOS versions, if it's easy to do and maintain. We have specific k8s supported versions, but AFAIK we don't have a matrix of supported OS versions where the Helm client runs 🤔  …or do we?

mattfarina:
@scottrigby this is for those who are developing Helm itself rather than anyone at all. But, we have people developing on a variety of versions so we should support that, IMHO.

mattfarina:
@scottrigby @sabre1041 can you take a look at the way this is handled now. It should work for both older and newer darwin.

scottrigby (approved):

PR #10110, opened 2021-09-07
Fix value precedence

zegerius:
Fix value precedence when importing values from child, as described in issue https://github.com/helm/helm/issues/9940

**What this PR does / why we need it**:

 Closes https://github.com/helm/helm/issues/9940

**If applicable**:
- [ ] this PR contains documentation
- [x] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

PR #11162, opened 2022-07-18
Introduce import-values strategy + revert default strategy for backwards compat

### LINKED ISSUE #9940: Value files precedence behaviour changes when custom values file is used
Edit: The issue can be reproduced and outputs can be seen here - https://github.com/ankitabhopatkar13/helm-values-issue

Output of `helm version`: `version.BuildInfo{Version:"v3.6.1", GitCommit:"61d8e8c4a6f95540c15c6a65f36a6dd0a45e7a2f", GitTreeState:"dirty", GoVersion:"go1.16.5"}`

I have 3 kind of charts -

- Library Chart
- App Chart (Uses Library Chart as dependecy)
- Umbrella Chart (Has all the app charts as dependencies)

Library chart has value structure like -
----- templates/
----- values.yaml
```
- exports:
     service:
       type: ClusterIP
       port: 9090
```

App Chart has value structure like -
----- templates/
----- values.yaml
```
service:
  type: ClusterIP
  port: 1234
```

Umbrella Chart has value structure like -
----- values/
------- aws/
--------- test.yaml
--------- accp.yaml
              
`helm upgrade --install release release --values aws/test.yaml`, the service port is still used as 9090 and not 1234.

But if I don't pass the custom values file and just use values.yaml in the umbrella chart, the port is getting set to 1234 which is what should be the case even when custom values file(test.yaml) is used.

### LINKED ISSUE #10899: Values in library subcharts are ignored
Output of `helm version`:
```
```

Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.1", GitCommit:"86ec240af8cbd1b60bcc4c03c20da9b98005b92e", GitTreeState:"clean", BuildDate:"2021-12-16T11:41:01Z", GoVersion:"go1.17.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.7+k3s1", GitCommit:"8432d7f239676dfe8f748c0c2a3fabf8cf40a826", GitTreeState:"clean", BuildDate:"2022-02-24T23:03:47Z", GoVersion:"go1.16.10", Compiler:"gc", Platform:"linux/amd64"}
```

## Description 

Starting from helm v3.8.2 a NBC change has been included, changing the precedence of imported values from library charts via *imported-values* in Chart.yaml. The following commit introduced that change: https://github.com/helm/helm/commit/5d017e11f1f47345a3559bf70f63d81a7edc981a.

An illustration for this behavior change can be seen in the following repository https://github.com/albgp22/helm-values-issue but can be stated simply as follows: when we import a value from a given library via *imported-values* in Chart.yaml file in our application chart, the value gets ignored if it is already present in the values.yaml file.

As this precedence change potentially breaks compatibility of existing charts with newer helm versions, I wonder if it will be reverted in future helm versions. As I see it, imported values from libraries overriding the ones in the application *values.yaml* file was the expected and logical behavior.

Cybermaxke:
With the introduction of a bug fix https://github.com/helm/helm/issues/9940 https://github.com/helm/helm/commit/5d017e11f1f47345a3559bf70f63d81a7edc981a there was a breaking introduced that makes it no longer possible to import values from a sub-chart that override values that are already defined in the importing chart. This issue is described in https://github.com/helm/helm/issues/10899

Between the original behavior and the one that "fixed" there are two possible scenarios for a string value:
1. override: import values from the sub-chart and override everything in the parent chart (old behavior)
2. ifNotPresent: import values from the sub-chart and only import the value if it's not defined in the chart chart (current behavior since the "fix")

Something else I noticed is that the ifNotPresent behavior was already applied to an import-values entry with child and parent paths before, so it was not consistently implemented even before the "fix". Even though the helm docs mention "override" behavior.

| Type | Original behavior | Since "fix" | With PR | [Docs](https://helm.sh/docs/topics/charts/) |
|---|---|---|---|---|
| string | override | ifNotPresent | override | override  (not explicitly mentioned, but I assume the same as child + parent path) |
| child + parent path | ifNotPresent | ifNotPresent | ifNotPresent | override |

So the aim of this PR is to revert the breaking changes, which should happen at major versions. And additionally introduce a "strategy" to explictly define the behavior (naming can still be discussed). Adding the strategy looks at follows:
```
# alternative for string with a different strategy
dependencies:
  - name: child
    import-values:
      - key: data
        strategy: override
```
```
# child + parent path with a strategy
dependencies:
  - name: child
    import-values:
      - child: default.data
        parent: myimports
        strategy: ifNotPresent
```

A few todos/questions:
- [ ] does the json spec of Chart.yaml need to updated somewhere? I wasn't able to find it in the repo.
- [ ] if desired for short term, we could just revert to the "fix PR" and you can use child + parent path instead
- [ ] docs will need to be adjusted, unless we want to change the default behavior of child + parent path from ifNotPresent to override to reflect the docs?
- [ ] expand unit tests?

Closes https://github.com/helm/helm/issues/10899

**If applicable**:

### DIFF
--- a/pkg/chartutil/dependencies.go
+++ b/pkg/chartutil/dependencies.go
@@ -226,49 +226,72 @@ func processImportValues(c *chart.Chart) error {
 	if err != nil {
 		return err
 	}
-	b := make(map[string]interface{})
+	c.Values = copyMap(cvals)
+	overrides := make(map[string]interface{})
 	// import values from each dependency if specified in import-values
 	for _, r := range c.Metadata.Dependencies {
 		var outiv []interface{}
 		for _, riv := range r.ImportValues {
+			var strategy string
+			var values map[string]interface{}
+			var child string
+			parent := "."
 			switch iv := riv.(type) {
 			case map[string]interface{}:
-				child := iv["child"].(string)
-				parent := iv["parent"].(string)
-
-				outiv = append(outiv, map[string]string{
-					"child":  child,
-					"parent": parent,
-				})
-
-				// get child table
-				vv, err := cvals.Table(r.Name + "." + child)
-				if err != nil {
-					log.Printf("Warning: ImportValues missing table from chart %s: %v", r.Name, err)
-					continue
+				var strategyOk bool
+				strategy, strategyOk = iv["strategy"].(string)
+				if !strategyOk {
+					strategy = "ifNotPresent"
 				}
-				// create value map from child to be merged into parent
-				b = CoalesceTables(cvals, pathToMap(parent, vv.AsMap()))
+				if key, ok := iv["key"].(string); ok {
+					child = "exports." + key
+					vm, err := cvals.Table(r.Name + "." + child)
+					if err != nil {
+						log.Printf("Warning: ImportValues missing table: %v", err)
+						continue
+					}
+					values = vm.AsMap()
+				} else {
+					child = iv["child"].(string)
+					parent = iv["parent"].(string)
+					vv, err := cvals.Table(r.Name + "." + child)
+					if err != nil {
+						log.Printf("Warning: ImportValues missing table from chart %s: %v", r.Name, err)
+						continue
+					}
+					values = pathToMap(parent, vv.AsMap())
+				}
 			case string:
+				// backwards compat to use the override strategy when using a string
+				strategy = "override"
 				child := "exports." + iv
-				outiv = append(outiv, map[string]string{
-					"child":  child,
-					"parent": ".",
-				})
 				vm, err := cvals.Table(r.Name + "." + child)
 				if err != nil {
 					log.Printf("Warning: ImportValues missing table: %v", err)
 					continue
 				}
-				b = CoalesceTables(b, vm.AsMap())
+				values = vm.AsMap()
 			}
+			// merge the values from the child into the parent values according to the import strategy
+			if strategy == "override" {
+				overrides = CoalesceTables(overrides, values)
+			} else if strategy == "ifNotPresent" {
+				c.Values = CoalesceTables(c.Values, values)
+			} else {
+				log.Printf("Warning: ImportValues with an invalid import strategy %v", strategy)
+				continue
+			}
+			outiv = append(outiv, map[string]string{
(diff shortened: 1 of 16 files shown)


antaloala:
This PR looks to be waiting for approval for some long time.
Is there anything preventing it to happen? I understand there could be a discussion on the new proposed yamls fields to introduce in the Chart.yaml schema ... but nothing is "happening" around that (?) 
We are not able to upgrade Helm from 3.8.1 to any newer release due to the NBC introduced by 3.8.2 on imported-values behavior; it looks we are not the only ones "suffering" this problem.

Asking here in order to understand if not accepting (yet) this PR is only because Helm guys allowed to accept it are overloaded ... or is it because they see some "issues/problems" on accepting it (?).

Thanks in advance.

joejulian:
Yes, the project does need more maintainers. If you're interested, the first step is contributing code.

Come join the developer meeting, if you can: https://github.com/helm/community/blob/main/communication.md#meetings.

Post your interest in becoming a maintainer on the [mailing list](https://github.com/helm/community/blob/main/communication.md#mailing-lists).

I also recommend familiarizing yourself with the [existing hips](https://github.com/helm/community/tree/main/hips#existing-hips)

Even if you're not interested in becoming a maintainer, the developer meeting is a great place to meet the maintainers and shed light on issues or PRs that need attention.

gusyag:
Hello!
Would it be possible to have this PR merged? Is there anything missing? 
We cannot use Helm >= 3.8.2 since https://github.com/helm/helm/pull/10110 was introduced. 
Thanks!

gusyag:
Hello @joejulian ,
I see you included this in 3.11.0 milestone, does it mean it will be implemented in that release?
Thanks!

joejulian:
I'm just a triage maintainer. No promises 😕.

That's the first opportunity for it to be released. Actual inclusion depends on the core maintainers.

scottrigby:
Needs rebase onto `main` — Helm 3 is not accepting new features.

PR #11165, opened 2022-07-20
add release tag for install and history

45cali:
A lot of companies use a single helm chart to deploy different services within their org. This pr adds the ability to have a release tag that can be used to help identify a version of a service that was deployed. 

A great example of this is if a helm rollback is needed. Currently helm history does not really show what release you want to rollback to only the revision.

An optional flag `--release-tag` has been added to the install and upgrade command.

```
go run ./ history demo-nginx
REVISION	UPDATED                 	STATUS    	CHART       	APP VERSION	DESCRIPTION 	RELEASE TAG
1       	Tue Jul 19 15:31:37 2022	superseded	nginx-13.1.2	1.23.0     	Release Mock	921103db
2       	Tue Jul 19 15:31:52 2022	superseded	nginx-13.1.2	1.23.0     	Release Mock	1a6f3966
3       	Tue Jul 19 15:32:05 2022	superseded	nginx-13.1.2	1.23.0     	Release Mock	211bd2b6    
4       	Tue Jul 19 15:32:19 2022	deployed  	nginx-13.1.2	1.23.0     	Release Mock	886f1e55
```

**If applicable**:


yxxhero:
@45cali thanks very much. This PR so large. Please write a HIP for this idea. Thanks very much.

45cali:
@yxxhero will do

45cali:
@yxxhero  Here is the HIP https://github.com/helm/community/pull/265

pkg/action/install.go
```diff
@@ -405,6 +407,7 @@ func (i *Install) performInstall(c chan<- resultMessage, rel *release.Release, t
 		rel.SetStatus(release.StatusDeployed, "Install complete")
 	}
 
+	rel.SetReleaseTag(i.ReleaseTag)
```
mattfarina, line 410: Is there a reason this needs to be a function? Can the public property just be set to be equal? `SetStatus` exists because 2 things are set at the same time. It's the only property updated with a function.

cmd/helm/history.go
```diff
@@ -104,9 +105,13 @@ func (r releaseHistory) WriteYAML(out io.Writer) error {
 
 func (r releaseHistory) WriteTable(out io.Writer) error {
 	tbl := uitable.New()
-	tbl.AddRow("REVISION", "UPDATED", "STATUS", "CHART", "APP VERSION", "DESCRIPTION")
+	tbl.AddRow("REVISION", "UPDATED", "STATUS", "CHART", "APP VERSION", "DESCRIPTION", "RELEASE TAG")
```
mattfarina, line 108: According to the [HIP on backwards compatibility](https://github.com/helm/community/blob/main/hips/hip-0004.md), "the format of structured output (tables, JSON, YAML) must not change". Adding a column changes the structure of the output. We have scripts that parse the output. This means the display change needs to be behind a flag.

Helm search of the hub already has an example of this... https://github.com/helm/helm/blob/d9f776bb3561647378b9ff588bb5042fbd37987c/cmd/helm/search_hub.go#L143-L147

mattfarina (changes requested):
Thanks for the PR. I've not had a chance to look through the HIP yet but I did find a couple of design issues in the PR I wanted to quickly point out.

cmd/helm/history.go
45cali: No problem ill add that in.

pkg/action/install.go
45cali: no problem i can switch it to that i was just following the pattern for SetStatus

cmd/helm/history.go
45cali: @mattfarina are thinking i should create a like searchhub does?
```
type hubSearchWriter struct {
	elements    []hubChartElement
	columnWidth uint
	listRepoURL bool
}
```
so it would be like this
```
type releaseHistoryWriter struct {
  elements         []releaseInfo
  listReleaseTag bool
}
```

45cali: @mattfarina just following up for clarity ^

45cali: @mattfarina Happy New Year!

Just wanted to follow up on this to see if i can get a little clarification.

45cali:
@mattfarina  

first time caller long time listener...

is this feature beneficial or should i close the pr? more than happy to refactor it with a little feedback.

Issue #11180, opened 2022-07-27
[feature]  On helm push command add --insecure param


lengrongfu:
Cotent:
   When I use helm to push a chart to the registry of the http environment locally, it will automatically access port 443. I hope that a `--insecure` parameter can be added to the push command to skip mandatory access to https.

Output of `helm version`:

```shell
$ helm version
version.BuildInfo{Version:"v3.9.2", GitCommit:"1addefbfe665c350f4daf868a9adc5600cc064fd", GitTreeState:"clean", GoVersion:"go1.17.12"}
```

Output of `kubectl version`:

```shell
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"24", GitVersion:"v1.24.3", GitCommit:"aef86a93758dc3cb2c658dd9657ab4ad4afc21cb", GitTreeState:"clean", BuildDate:"2022-07-13T14:30:46Z", GoVersion:"go1.18.3", Compiler:"gc", Platform:"linux/amd64"}
```

Cloud Provider/Platform (AKS, GKE, Minikube etc.): Minikube

lengrongfu:
@yxxhero Hello， can you place review of this issue?

PR #10408, opened 2021-11-29
Push to OCI insecure registry

pytimer:
**What this PR does / why we need it**:

This PR implement push chart to OCI registry, or pull chart from OCI registry.

Fixes: https://github.com/helm/helm/issues/6324

Usage:

```bash
# push
$ ./bin/helm push consul-3.9.5.tgz oci://docker.registry:30001/library --insecure-skip-tls-verify 

# pull
$ ./bin/helm pull oci://docker.registry:30001/library/consul --version 3.9.5 --insecure-skip-tls-verify

# install
$ ./bin/helm install consul oci://docker.registry:30001/library/consul --version 3.9.5 --insecure-skip-tls-verify

# upgrade
$ ./bin/helm upgrade --install consul oci://docker.registry:30001/library/consul --version 3.9.6 --insecure-skip-tls-verify
```

**Special notes for your reviewer**:

**If applicable**:
- [ ] this PR contains documentation
- [ ] this PR contains unit tests
- [x] this PR has been tested for backwards compatibility

PR #11183, opened 2022-07-27
feat: add --registry-insecure param

lengrongfu:
Signed-off-by: lengrongfu <1275177125@qq.com>

refs: [https://github.com/helm/helm/issues/11180](https://github.com/helm/helm/issues/11180)

- This PR mainly provides an option to use HTTP for data transfer instead of HTTPS when pushing/pulling charts to non-secure repositories.
- We often don't care about security when verifying certain functions individually or as a team. At this time, we need to quickly perform functional verification.

This `PR` has no changes to the previous function, but after adding this parameter, I can quickly verify whether the `registry` I deploy can store `helm chart` artifacts, instead of spending time generating certificates and verifying them.

**If applicable**:


yxxhero:
@lengrongfu chat in dingding.

yxxhero:
@lengrongfu Can you add some unittest for your param and funcs.

cmd/helm/push_test.go
```diff
@@ -17,11 +17,138 @@ limitations under the License.
...
+}
+
+func getClientIP() (string, error) {
+	addrs, err := net.InterfaceAddrs()
+
+	if err != nil {
+		return "", err
+	}
+	for _, address := range addrs {
+		// 检查ip地址判断是否回环地址
```
yxxhero, line 66: not use chinese.

cmd/helm/push_test.go
```diff
@@ -17,11 +17,138 @@ limitations under the License.
...
+		}
+	}
+
+	return "", errors.New("can not find the client ip address")
+
+}
+
+func (suite *HelmPushTestSuite) SetupSuite() {
+	suite.WorkspaceDir = testWorkspaceDir
+	os.RemoveAll(suite.WorkspaceDir)
```
yxxhero, line 80: why not use defer?

cmd/helm/push_test.go
```diff
@@ -17,11 +17,138 @@ limitations under the License.
...
+	os.Mkdir(suite.WorkspaceDir, 0700)
+
+	var out bytes.Buffer
+	suite.Out = &out
+	//credentialsFile := filepath.Join(suite.WorkspaceDir, registry.CredentialsFileBasename)
```
yxxhero, line 85: if the code is useless. please remove it.

cmd/helm/push_test.go
```diff
@@ -17,11 +17,138 @@ limitations under the License.
...
+	dockerRegistry, err := registry.NewRegistry(context.Background(), config)
+	suite.Nil(err, "no error creating test registry")
+
+	suite.CompromisedRegistryHost = fmt.Sprintf("%s:%d", ip, port)
+
+	// Start Docker registry
+	go dockerRegistry.ListenAndServe()
+}
+
+func (suite *HelmPushTestSuite) Test_1_Push() {
```
yxxhero, line 108: try to write test like:
```golang
tests:=[]struct{}
for _,tt := range tests {}
```
Thanks very much

cmd/helm/push_test.go
lengrongfu: Because the `registry` service is to be started here, SetupSuite is used to manage this logic. If using defer, there will be problems starting the `registry` when running tests in parallel.

cmd/helm/push_test.go
lengrongfu: This variable is initialized in `SetupSuite` func and used in `Test_1_Push` and `Test_2_Push` test func

yxxhero: //credentialsFile := filepath.Join(suite.WorkspaceDir, registry.CredentialsFileBasename)

cmd/helm/push_test.go
lengrongfu: I think I can make the test more intuitive by changing the function name. For example, `Test_1_Push` is changed to `TestSecretPush`, and `Test_2_Push` is changed to `TestInsecretPush`. Currently, each test function has initial logic, and it is not easy to initialize using this format.

yxxhero: ok

cmd/helm/push_test.go
lengrongfu: yes，You are right

cmd/helm/push_test.go
```diff
@@ -17,11 +17,137 @@ limitations under the License.
...
+	suite.Suite
+	Out                     io.Writer
+	DockerRegistryHost      string
+	CompromisedRegistryHost string
+	WorkspaceDir            string
+	RegistryClient          *helmRegistry.Client
+	PushClient              *action.Push
+}
+
+func getClientIP() (string, error) {
```
yxxhero, line 59: need unittest for this func.

pkg/cli/environment.go
```diff
@@ -139,6 +142,8 @@ func (s *EnvSettings) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&s.RepositoryConfig, "repository-config", s.RepositoryConfig, "path to the file containing repository names and URLs")
 	fs.StringVar(&s.RepositoryCache, "repository-cache", s.RepositoryCache, "path to the file containing cached repository indexes")
 	fs.IntVar(&s.BurstLimit, "burst-limit", s.BurstLimit, "client-side default throttling limit")
+	fs.BoolVar(&s.RegistryInsecure, "registry-insecure", s.RegistryInsecure, "enable registry is insecure mode, use http transport data")
```
yxxhero, line 145: enable registry is insecure mode?

cmd/helm/push_test.go
```diff
@@ -17,11 +17,145 @@ limitations under the License.
...
+	"github.com/stretchr/testify/suite"
 )
 
 func TestPushFileCompletion(t *testing.T) {
 	checkFileCompletion(t, "push", true)
 	checkFileCompletion(t, "push package.tgz", false)
 	checkFileCompletion(t, "push package.tgz oci://localhost:5000", false)
 }
+
+var (
```
yxxhero, line 45: use const is better?

cmd/helm/push_test.go
```diff
@@ -17,11 +17,145 @@ limitations under the License.
...
+	config.HTTP.DrainTimeout = time.Duration(10) * time.Second
+	config.Storage = map[string]configuration.Parameters{"inmemory": map[string]interface{}{}}
+
+	dockerRegistry, err := registry.NewRegistry(context.Background(), config)
+	suite.Nil(err, "no error creating test registry")
+
+	suite.CompromisedRegistryHost = fmt.Sprintf("%s:%d", ip, port)
+
+	// Start Docker registry
+	go dockerRegistry.ListenAndServe()
```
yxxhero, line 112: shall we should stop it after test?

yxxhero: resolved?

willzhang:
when it will release?

sabre1041:
@lengrongfu thanks for your contribution. The concept of insecure as it relates to connecting to OCI registries is actually covered by two concepts:

1. Ignoring the verification of SSL certificates
2. Connecting over non-secure transport mechanisms (HTTP)

This PR appears to cover #2 but does not address #1. 

#10408 appears to be a more comprehensive implementation for which is being tested currently

PR #11201, opened 2022-08-04
Recursively call SaveDir, so that dependencies appear as directory

JensAc:
Signed-off-by: Jens Schneider <jens.schneider.ac@posteo.de>

It makes sure that dependencies are included as directories, when charutil.SaveDir() is called. IMHO, this makes sense when you want to store charts and their dependencies locally for development purposes.
I don't know whether there was a specific reason for saving tar.gz-charts, instead. If so, just let me know :slightly_smiling_face: and reject the PR.

**If applicable**:
- [x] This PR contains minimal changes :wink:

### DIFF
--- a/pkg/chartutil/save.go
+++ b/pkg/chartutil/save.go
@@ -83,8 +83,8 @@ func SaveDir(c *chart.Chart, dest string) error {
 	// Save dependencies
 	base := filepath.Join(outdir, ChartsDir)
 	for _, dep := range c.Dependencies() {
-		// Here, we write each dependency as a tar file.
-		if _, err := Save(dep, base); err != nil {
+		// Here, we call the SaveDir function again for each dependency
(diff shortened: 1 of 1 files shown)


mattfarina (changes requested):
The handling for dependent charts as archives is different from those that are directories. When one is an archive it can be updated using Helm commands like `helm dependency update`. When a dependent chart is a directory it's expected to be managed separately from that. That the chart author will maintain that.

This would also be the type of change that's a breaking change. Helm follows semantic versioning and this would be a breaking functionality change.

I hope this context help.

mattfarina:
I'm going to close this PR as it is a change we can't accept right now.

JensAc:
Ok, I am totally fine with that. Then we will use a fork instead, as we need this functionality for development purposes.

PR #11220, opened 2022-08-09
[FIX] Validate chart name before packaging

### LINKED ISSUE #11200: Invalid reference format with charts containing space in name
Steps to reproduce:
- In Chart.yaml, specify a name with a space, e.g., `sample chart`
- Package the chart
- Push to a registry, e.g., `sample chart-1.1.3.tgz oci://localhost:5001/sample-charts`

Expected result: The chart is now available in the registry. Or a descriptive error message is given back.
Actual result: I get a cryptic `invalid reference format` error.

Output of `helm version`:
```
version.BuildInfo{Version:"v3.9.2", GitCommit:"1addefbfe665c350f4daf868a9adc5600cc064fd", GitTreeState:"clean", GoVersion:"go1.18.4"}
```

Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"24", GitVersion:"v1.24.1", GitCommit:"3ddd0f45aa91e2f30c70734b175631bec5b5825a", GitTreeState:"clean", BuildDate:"2022-05-24T12:26:19Z", GoVersion:"go1.18.2", Compiler:"gc", Platform:"darwin/amd64"}
Kustomize Version: v4.5.4
Server Version: version.Info{Major:"1", Minor:"24", GitVersion:"v1.24.0", GitCommit:"4ce5a8954017644c5420bae81d72b09b735c21f0", GitTreeState:"clean", BuildDate:"2022-05-19T15:39:43Z", GoVersion:"go1.18.1", Compiler:"gc", Platform:"linux/amd64"}
```

debdutdeb:
Signed-off-by: Debdut Chakraborty <debdutdeb@outlook.com>

Closes #11200 

**If applicable**:

### DIFF
--- a/pkg/action/package.go
+++ b/pkg/action/package.go
@@ -62,8 +62,8 @@ func (p *Package) Run(path string, vals map[string]interface{}) (string, error) 
 		return "", err
 	}
 
-	if matched, _ := regexp.MatchString("^[[:lower:][:digit:]-]+$", ch.Name()); !matched {
-		return "", errors.New(fmt.Sprintf("Invalid chart name \"%s\". Chart names must be lower case letters and numbers with optionally dashes (-) for separator\n", ch.Name()))
+	if err := validateName(ch.Name()); err != nil {
+		return "", err
 	}
 
 	// If version is set, modify the version.
@@ -107,6 +107,14 @@ func (p *Package) Run(path string, vals map[string]interface{}) (string, error) 
 	}
 
 	return name, err
+}
+
+// validateName verifies if the chart name has any illegal characters or not
+func validateName(name string) error {
+	if matched, _ := regexp.MatchString("^[[:lower:][:digit:]-]+$", name); !matched {
+		return errors.New(fmt.Sprintf("Invalid chart name \"%s\". Chart names must be lower case letters and numbers with optionally dashes (-) for separator\n", name))
+	}
+	return nil
 }
 
 // validateVersion Verify that version is a Version, and error out if it is not.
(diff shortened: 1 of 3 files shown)


pkg/action/package.go
```diff
@@ -104,6 +109,14 @@ func (p *Package) Run(path string, vals map[string]interface{}) (string, error)
 	return name, err
 }
 
+// validateName verifies if the chart name has any illegal characters or not
+func validateName(name string) error {
+	if matched, _ := regexp.MatchString("^[[:lower:][:digit:]-]+$", name); !matched {
+		return errors.New(fmt.Sprintf("Invalid chart name \"%s\". Chart names must be lower case letters and numbers with optionally dashes (-) for separator\n", name))
```
mattfarina, line 115: Linking is noting... `should replace errors.New(fmt.Sprintf(...)) with fmt.Errorf(...)`. Can you please make this change?

pkg/action/package.go
```diff
@@ -104,6 +109,14 @@ func (p *Package) Run(path string, vals map[string]interface{}) (string, error)
 	return name, err
 }
 
+// validateName verifies if the chart name has any illegal characters or not
+func validateName(name string) error {
```
mattfarina, line 113: This function doesn't validate the same as we do elsewhere. See https://github.com/helm/helm/blob/26ce96464f73ca72d0855cf49af31bac47f38bb7/pkg/chartutil/create.go#L679-L687

I think we need to be consistent throughout Helm. Can you please update this?

I also wonder if this should become a public function (with tests).

debdutdeb: > I also wonder if this should become a public function (with tests).

I agree.

But, the chart name regex here does not match the requirements mentioned in the doc; 
https://github.com/helm/helm/blob/26ce96464f73ca72d0855cf49af31bac47f38bb7/pkg/chartutil/create.go#L39
https://helm.sh/docs/chart_best_practices/conventions/#chart-names

debdutdeb: @mattfarina Not sure whether to change regex to what's mentioned in the doc or not.

PR #10207, opened 2021-10-01
Fix User-Agent header in requests made by Helm

programmer04:
Current Helm v3.8.1 installed from package manager or build from main branch when makes requests to K8s API (with use of K8s SDK), e.g. executing the command

```
helm history test
```
does not have the proper value set in `User-Agent` header. It looks like that on macOS Monterey with Apple Silicon.
```yaml
User-Agent: Go-http-client/1.1
```
This PR fixes it and makes it consistent with `User-Agent` set when requests are made to chart repositories (it works as expected). So now it always looks like that
```yaml
User-Agent: Helm/3.8+unreleased
```
or
```yaml
User-Agent: Helm/3.8
```

Hence I removed those comments too from the code
```
// TODO add user-agent
```
because it's been already done and works as expected for interaction with chart repos and etc. Here is the line which provides it https://github.com/helm/helm/blob/0d9ebc7885e427373fbdaa95f461b349fb0b1ac1/pkg/getter/httpgetter.go#L53

PR #11226, opened 2022-08-12
Fix User-Agent header in requests made by Helm 

programmer04:
> This is the recreation of PR #10207 which has been closed due to problems kicking off CI for it.

Current Helm v3.8.1 installed from package manager or build from main branch when makes requests to K8s API (with use of K8s SDK), e.g. executing the command

```
helm history test
```
does not have the proper value set in `User-Agent` header. It looks like that on macOS Monterey with Apple Silicon.
```yaml
User-Agent: Go-http-client/1.1
```
This PR fixes it and makes it consistent with `User-Agent` set when requests are made to chart repositories (it works as expected). So now it always looks like that
```yaml
User-Agent: Helm/3.8+unreleased
```
or
```yaml
User-Agent: Helm/3.8
```

Hence I removed those comments too from the code
```
// TODO add user-agent
```
because it's been already done and works as expected for interaction with chart repos and etc. Here is the line which provides it https://github.com/helm/helm/blob/0d9ebc7885e427373fbdaa95f461b349fb0b1ac1/pkg/getter/httpgetter.go#L53

### DIFF
--- a/pkg/cli/environment.go
+++ b/pkg/cli/environment.go
@@ -14,7 +14,8 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-/*Package cli describes the operating environment for the Helm CLI.
+/*
+Package cli describes the operating environment for the Helm CLI.
 
 Helm's environment encapsulates all of the service dependencies Helm has.
 These dependencies are expressed as interfaces so that alternate implementations
@@ -121,6 +122,7 @@ func New() *EnvSettings {
 			config.Wrap(func(rt http.RoundTripper) http.RoundTripper {
 				return &retryingRoundTripper{wrapped: rt}
 			})
+			config.UserAgent = version.GetUserAgent()
 			return config
 		},
 	}
@@ -231,9 +233,5 @@ func (s *EnvSettings) SetNamespace(namespace string) {
 
 // RESTClientGetter gets the kubeconfig from EnvSettings
 func (s *EnvSettings) RESTClientGetter() genericclioptions.RESTClientGetter {
-	s.config.WrapConfigFn = func(c *rest.Config) *rest.Config {
-		c.UserAgent = version.GetUserAgent()
-		return c
-	}
 	return s.config
 }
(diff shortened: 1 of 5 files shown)


pkg/cli/environment_test.go
```diff
@@ -231,6 +233,21 @@ func TestEnvOrBool(t *testing.T) {
 	}
 }
 
+func TestUserAgentHEaderInK8sRESTClientConfig(t *testing.T) {
```
phil9909, line 236:
```suggestion
func TestUserAgentHeaderInK8sRESTClientConfig(t *testing.T) {
```

phil9909 (changes requested):

davidkarlsen (approved):
Lgtm

phil9909 (approved):

pkg/cli/environment.go
```diff
@@ -230,5 +231,9 @@ func (s *EnvSettings) SetNamespace(namespace string) {
 
 // RESTClientGetter gets the kubeconfig from EnvSettings
 func (s *EnvSettings) RESTClientGetter() genericclioptions.RESTClientGetter {
+	s.config.WrapConfigFn = func(c *rest.Config) *rest.Config {
+		c.UserAgent = version.GetUserAgent()
+		return c
+	}
```
mattfarina, line 237: My concern is that a `WrapConfigFn` is already setup. Won't this overwrite that? Should we combine them? See...

https://github.com/helm/helm/blob/5a8e91a010a923c3505c19954cfda4f01432bbc9/pkg/cli/environment.go#L118-L124

programmer04: Yeah, the lines mentioned by you were introduced after my original PR. Good catch! I am moving my change here

mattfarina (approved):
Note, this needs a second maintainer to approve before it can be merged.

mattfarina:
@programmer04 could you please rebase this PR from main.

programmer04:
Thanks for reaching me, it's done @mattfarina

hickeyma (approved):
LGTM, thanks @programmer04 for pushing new PR to fix CI issue.

PR #11237, opened 2022-08-18
feat(engine): add jq filter template func

### LINKED ISSUE #11212: Support jq filter and query values
Currently the values is basically a json structure(a go `map[string]any`), we can easily specify which values we want to populate in the template(like `a.b.c`).

However, when it comes to list type, apart from `first` or `last`, there is no template func to filter. We have to use `{{ range }} {{ if }}` to filter what we want. Consider a simple case, when to want to filter a k8s container port:

```yaml
ports:
- containerPort: 80
  name: http
  protocol: TCP
- containerPort: 443
  name: https
  protocol: TCP
- containerPort: 8080
  name: grpc
  protocol: TCP
- containerPort: 9090
  name: http
  protocol: TCP
```

In the template we would write:

```
# other parts of yaml...
port: {{ range $i, $val := .Values.ports }}
  {{- if eq $val.name "http" -}}
    {{ $val.containerPort }}
  {{- end -}}
{{- end }}
# other parts of yaml...
```

For a simple port value which is verbose and sometimes the new line and space is error prone.

Consider a more complex example, say we have terraform state file, which have lots of resource then we want to filter and inject some value into our template:

```json
{
  "version": 4,
  "terraform_version": "1.2.6",
  "serial": 359,
  "lineage": "ba7d9cd8-8885-2336-4748-560d2caa712c",
  "outputs": {},
  "resources": [
    {...},
    {...},
    {
      "mode": "managed",
      "type": "tencentcloud_vpc",
      "name": "vpc",
      "provider": "provider[\"registry.terraform.io/tencentcloudstack/tencentcloud\"]",
      "instances": [
        {
          "schema_version": 0,
          "attributes": {
            "assistant_cidrs": [],
            "cidr_block": "10.2.0.0/16",
            "create_time": "2022-08-01 16:05:35",
            "default_route_table_id": "rtb-321",
            "dns_servers": [
              "8.8.8.8"
            ],
            "id": "vpc-123",
            "is_default": false,
            "is_multicast": true,
            "name": "tf-eks-vpc",
            "tags": {}
          },
          "sensitive_attributes": [],
          "private": "bnVsbA=="
        }
      ]
    }
  ]
}
```

There are is two nested level array, which is hard to use go template `range` and `if`.

What if we could use the popular [jq](https://stedolan.github.io/jq/) query syntax?

The result could be simple and clear online.

```yaml
# ...
vpcId: "{{ . | jq `.resources[] | select(.name == "vpc" ) | .instances[0].attributes.id` }}"
# ...
```

I believe providing a `jq` template function is really helpful and powerful for the filter situation.

longkai:
resolve #11212

see issue #11212 

No.

**If applicable**:

### DIFF
--- a/pkg/engine/funcs.go
+++ b/pkg/engine/funcs.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/BurntSushi/toml"
 	"github.com/Masterminds/sprig/v3"
+	"github.com/itchyny/gojq"
 	"sigs.k8s.io/yaml"
 )
 
@@ -35,12 +36,11 @@ import (
 //
 // Known late-bound functions:
 //
-//	- "include"
-//	- "tpl"
+//   - "include"
+//   - "tpl"
 //
 // These are late-bound in Engine.Render().  The
 // version included in the FuncMap is a placeholder.
-//
 func funcMap() template.FuncMap {
 	f := sprig.TxtFuncMap()
 	delete(f, "env")
@@ -55,6 +55,7 @@ func funcMap() template.FuncMap {
 		"toJson":        toJSON,
 		"fromJson":      fromJSON,
 		"fromJsonArray": fromJSONArray,
+		"jq":            jq,
 
 		// This is a placeholder for the "include" function, which is
 		// late-bound to a template. By declaring it here, we preserve the
@@ -174,4 +175,37 @@ func fromJSONArray(str string) []interface{} {
 		a = []interface{}{err.Error()}
 	}
 	return a
+}
+
+// jq is like sed for JSON/YAML data - you can use it to slice and filter
+// and map and transform structured data with the same ease that sed, awk, grep
+// and friends let you play with text.
+//
+// See the [jq Tutorial](https://stedolan.github.io/jq/tutorial/) for more.
+//
+// This is a pure go implementation of jq. The type of v should be
+// []interface{} for an array or map[string]interface{} for a map.
+// Since the jq filter returns a iterator, it will be collected as a slice
+// even there is only one item.
+// It will insert the returned error message string as
+// the first and only item in the returned array.
+func jq(exp string, v interface{}) []interface{} {
+	query, err := gojq.Parse(exp)
+	if err != nil {
+		return []interface{}{err.Error()}
+	}
+
+	iter := query.Run(v)
+	var list []interface{}
+	for {
+		it, ok := iter.Next()
+		if !ok {
+			break
+		}
+		if err, ok := it.(error); ok {
+			return []interface{}{err.Error()}
+		}
+		list = append(list, it)
+	}
+	return list
 }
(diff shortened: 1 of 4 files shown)


pkg/engine/funcs.go
```diff
@@ -175,3 +177,36 @@ func fromJSONArray(str string) []interface{} {
...
+//
+// See the [jq Tutorial](https://stedolan.github.io/jq/tutorial/) for more.
+//
+// This is a pure go implementation of jq. The type of v should be
+// []interface{} for an array or map[string]interface{} for a map.
+// Since the jq filter returns a iterator, it will be collected as a slice
+// even there is only one item.
+// It will insert the returned error message string as
+// the first and only item in the returned array.
+func jq(exp string, v any) []any {
```
phil9909, line 193: I guess you should change this function to return `([]any, error)`, see https://pkg.go.dev/text/template#FuncMap

pkg/engine/funcs_test.go
```diff
@@ -99,6 +99,28 @@ func TestFuncs(t *testing.T) {
...
+					"name":          "https",
+					"protocol":      "TCP",
+				},
+				map[string]any{
+					"containerPort": 8080,
+					"name":          "grpc",
+					"protocol":      "TCP",
+				},
+			},
+		},
```
sudermanjr, line 123: Would it be possible to introduce a test case or two that cover the error conditions here?

sudermanjr (changes requested):
Seems like it works really well, and is a straightforward implementation. Just a small request for a little bit more coverage on the tests.

phil9909 (changes requested):
I have some mixed feelings about this:
I definitely see how this could be useful, but on the other hand it introduces a second language to helm, which could cause confusion.

pkg/engine/funcs_test.go
longkai: You are right. I will add more test cases later.

pkg/engine/funcs.go
longkai: Returning `([]any, error)` is better. However I just saw the functions(like `fromYaml, toJson`) in `funcs.go` all inserting the error message into the normal returned value, besides the `TestFuncs()` use the error message to check the expectation. If we change the return values, how about introducing another test func?

joejulian:
Please address CI failures.

avo-sepp:
Hi, I'm checking if this work has gone stale? I'd like to continue on this PR as I am in need of some additional `jq` functionality.

longkai:
Hello, thanks for the reply. I've been busy at work recently and I'll fix the CI failures this week.

longkai:
@joejulian hello, the conflicts have been resolved but the ci failure below has nothing to do with my pr. Should I fix it via another PR? Also, please take a look at this PR. It's almost half a year passed.

image

jcchavezs:
Any movement on this?

Issue #11169, opened 2022-07-21
Possible data race in kube.New because of AddToScheme


sivanov-nuodb:
**Issue**

I'm using helm pkg in combination with controller-runtime to programmatically install Helm charts by Kubernetes operator.
Enabling the race detector prints several reports for unguarded concurrent access to the global scheme object.
It looks like https://github.com/helm/helm/issues/6566 fixes the problem between several `kube.New` concurrent invocations , however, it doesn't take into account a global scheme object is used.

```
WARNING: DATA RACE
Write at 0x00c0003284e0 by goroutine 72:
  runtime.mapassign()
      /usr/local/Cellar/go/1.18.1/libexec/src/runtime/map.go:578 +0x0
  k8s.io/apimachinery/pkg/runtime.(*Scheme).AddKnownTypeWithName()
      /go/pkg/mod/k8s.io/apimachinery@v0.23.5/pkg/runtime/scheme.go:174 +0x3a4
  k8s.io/apimachinery/pkg/runtime.(*Scheme).AddKnownTypes()
      /go/pkg/mod/k8s.io/apimachinery@v0.23.5/pkg/runtime/scheme.go:148 +0x2e8
  k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1.addKnownTypes()
      /go/pkg/mod/k8s.io/apiextensions-apiserver@v0.23.4/pkg/apis/apiextensions/v1/register.go:48 +0x1d8
  k8s.io/apimachinery/pkg/runtime.(*SchemeBuilder).AddToScheme()
      /go/pkg/mod/k8s.io/apimachinery@v0.23.5/pkg/runtime/scheme_builder.go:29 +0x95
  k8s.io/apimachinery/pkg/runtime.(*SchemeBuilder).AddToScheme-fm()
      <autogenerated>:1 +0x44
  helm.sh/helm/v3/pkg/kube.New.func1()
      /go/pkg/mod/helm.sh/helm/v3@v3.8.1/pkg/kube/client.go:83 +0x59
  sync.(*Once).doSlow()
      /usr/local/Cellar/go/1.18.1/libexec/src/sync/once.go:68 +0x101
  sync.(*Once).Do()
      /usr/local/Cellar/go/1.18.1/libexec/src/sync/once.go:59 +0x46
  helm.sh/helm/v3/pkg/kube.New()
      /go/pkg/mod/helm.sh/helm/v3@v3.8.1/pkg/kube/client.go:82 +0x6a
[...]

PR #11249, opened 2022-08-19
Fix data race in Client

ash2k:

Makes `Client` safe for concurrent use. Relates to https://github.com/helm/helm/issues/11169.

Ideally `Client` would just take `*kubernetes.Clientset` as a direct dependency, but that'd break existing users.

**If applicable**:

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -68,6 +68,7 @@ type Client struct {
 	// Namespace allows to bypass the kubeconfig file for the choice of the namespace
 	Namespace string
 
+	mu         sync.Mutex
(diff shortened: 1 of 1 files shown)


ash2k:
Actually, after looking at more of Helm code, I'm not sure what guarantees or conventions are in place. Things seem to be mutated all over the place. So this PR doesn't make sense - a more holistic review may be needed.

As an SDK user, this is confusing. So much mutable state everywhere, I'm not sure what is safe for concurrent use and what is not.

Fixing any of that is likely impossible without breaking changes due to v3 compatibility guarantees.

PR #9482, opened 2021-03-14
Continue to add support for git urls as dependency repositories

yxxhero:
Continue the PR https://github.com/helm/helm/pull/6734 work.
related issue: https://github.com/helm/helm/issues/9461

**If applicable**:
- [x] this PR contains documentation
- [x] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

Issue #9461, opened 2021-03-11
Working with unpackaged charts as dependency


BlueCog:
Hello Helm Community! 

Out team is working with Helm for some time now and we like to use it to do our K8 deployments for internal usage. 

Out team has (**unpackaged**) Helm umbrella charts & dependend Helm charts in a seperate GIT repositories. With GitOps in mind we'd like to work only with unpackaged charts. We also like to minimize any build steps, so we would like to skip the packaging & pushing to internal Helm repo steps. 

So we would like to point a helm dependency in `chart.yaml` to a GIT repository. 

```
# Chart.yaml
dependencies:
- name: nginx
  version: "1.2.3"
  repository: "ssh://git@git.host:7999/helm-dep/helm-dep.git"
- name: memcached
  version: "3.2.1"
  repository: "https://another.example.git.host/helm-dep"
``` 

A different solution would be to pull the dependend charts by hand (or scripted). But with GitOps in mind out team like to have the some kind overview of the deployed dependend charts (versions) in that moment of time. And we like to minimize any custom scripting and files. We also could use the `file:// ` option in `chart.yaml` but it seems working with the  `file:// ` option doesn't work togehter with other Helm repo's. So a ```helm dependency update``` wont work....

I've searched trough blogs/github issues/stackoverflow but it seems this is not supported at the moment. Is this something wat would be on the Helm roadmap? Or is this not the way to go / working as intended.

bacongobbler:
Helm does not currently support git repositories in its dependency resolver. https://github.com/helm/helm/pull/6734 is one of the longest standing PRs in the PR queue, and it appears that the author has not been active recently. If you're interested in this feature I'd suggest reaching out and seeing if you can carry on the work.

yxxhero:
@bacongobbler   I think  I can carry on the work. Can it assign to me?  Thanks very much!

BlueCog:
Hello @bacongobbler and @yxxhero 

Thanks for the reactions and pointer to the PR, i didn't find this one. It seems that this is exactly what I wanted to accomplish. 

It seems the PR is allmost ready, awesome that yxxhero can carry on the work! I will monitor this PR closely, if you need to have any active testing on this let me know.  

Greets

yxxhero:
@BlueCog  https://github.com/helm/helm/pull/9482 You can use it for the tests.

caleb15:
I still think this would be a cool feature, @bacongobbler can you unstale this and add a ignore-stale lifecycle please?

[...]

PR #6734, opened 2019-10-22
Add support for git urls as dependency repositories

rally25rs:
**What this PR does / why we need it**:

This PR adds support for using a git URL as a `repository` in `requirements.yaml`.

I thought this would be a nice-to-have feature for organizations that house their helm charts in a git repository and would like to maintain private access to those charts.

The format supported is:

```
dependencies:
- name: "{dependency name}"
  repository: "git:{git repo url}"
  version: "{git ref name}"
```
where:
* `{git repo url}` is any url type that a `git clone ...` command would accept (an ssh location, an https url, etc.).
* `{git ref name}` is an existing tag or branch name on the repo (a commit sha cannot be used at this time)

For example:

```
dependencies:
- name: "a"
  repository: "git:https://github.com/rally25rs/helm-test-chart.git"
  version: "master"
```

(the above `rally25rs/helm-test-chart` does exist. feel free to use it for testing if you want.)

**Special notes for your reviewer**:

This is my first time coding in Go, so please feel free to comment on the code and provide any refactoring suggestions or ways to do things better.

Since other package managers like `npm/yarn`, `ruby bundler`, etc. support github URLs, I thought this might be a nice to have feature. If this isn't the direction that the `helm` community wants to take for dependency management, we can reject this PR. I won't be offended :) It was still a fun learning experience to do something with Go, and this only took a day and a half to whip up. I actually made this PR instead of trying to build my own private custom registry in the traditional helm way, because that seemed more difficult and would likely require some process to monitor our github organization for changes, check them out to a central location, run the helm command to build the index.yaml, and push it all to a central server that only our employees have access to. Git already has SSH to secure it, so including our private charts in this way seemed a lot easier.

Thanks for your consideration and for taking the time to review this!

**If applicable**:
- [X] this PR contains documentation (I added to ` cmd/helm/dependency.go` but it looks like the rest of the generated docs are no longer in the v3 codebase, and `make docs` doesn't seem to be a command any more? If there are additional docs to edit please let me know)
- [X] this PR contains unit tests (I added a few where I could. Also see Manual Verification notes below.)
- [X] this PR has been tested for backwards compatibility (since no existing registry urls would have started with `git:` there shouldn't be any problem with existing registries)

**Implementation Details**

The implementation is a simplified take on what `yarn` does for git dependencies https://github.com/yarnpkg/yarn/blob/master/src/util/git.js (I'm also a `yarn` contributor so was familiar with their code base) 

What it basically does is, if it find a `registry` that starts with `git:` then:
[...]

PR #11258, opened 2022-08-23
Support git url repository

yxxhero:
Continue the PR https://github.com/helm/helm/pull/6734 work.
related issue: https://github.com/helm/helm/issues/9461

**If applicable**:

see: https://github.com/helm/helm/pull/9482


pkg/getter/gitgetter.go
```diff
@@ -0,0 +1,114 @@
...
+		return nil, err
+	}
+	if err := repo.UpdateVersion(version); err != nil {
+		return nil, err
+	}
+
+	// A .helmignore that includes an ignore for .git/ should be included in the git repo itself,
+	// but a lot of people will probably not think about that.
+	// To prevent the git history from bleeding into the charts archive, append/create .helmignore.
+	g.ensureGitDirIgnored(chartTmpDir)
```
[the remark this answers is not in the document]
TBBle: Also, does this even work? https://github.com/helm/helm/blob/v3.9.3/pkg/chart/loader/load.go#L55-L57 suggests it will be ignored when this archive is handled.

Would it be better to honour .helmignore `CompressDirToTgz` perhaps?

internal/resolver/resolver.go
```diff
@@ -107,6 +110,27 @@ func (r *Resolver) Resolve(reqs []*chart.Dependency, repoNames map[string]string
 			continue
 		}
 
+		if strings.HasPrefix(d.Repository, "git://") {
```
TBBle, line 113: Perhaps all the open-coded `strings.HasPrefix(d.Repository, "git://")` calls could be moved into a function, e.g., `gitutils.IsGit`, analog to `registry.IsOCI`.

internal/resolver/resolver.go
```diff
@@ -107,6 +110,27 @@ func (r *Resolver) Resolve(reqs []*chart.Dependency, repoNames map[string]string
...
+
+			found, err := hasGitReference(strings.TrimPrefix(d.Repository, "git://"), d.Version, d.Name)
```
TBBle, line 115: It'd make this sort of code a bit nicer if the APIs in gitutils took an untrimmed URL, so that we don't have the explicit `"git://"` peppering the non-git-specific files more than necessary. (I see also a use of external `giturls.Parse` later, perhaps wrap that in a gitutils function that trims first and passes it through).

pkg/getter/gitgetter.go
```diff
@@ -0,0 +1,114 @@
...
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/Masterminds/vcs"
+
+	"helm.sh/helm/v3/internal/fileutil"
+)
+
+// GitGetter is the default HTTP(/S) backend handler
```
TBBle, line 31: Is this comment correct?

pkg/gitutils/gitutils.go
```diff
@@ -0,0 +1,47 @@
...
+	if err != nil {
+		return false, err
+	}
+	repo, err := vcs.NewRepo(gitRepo, local)
+
+	if err != nil {
+		return false, err
+	}
+
+	if err := repo.Get(); err != nil {
```
TBBle, line 42: Is this doing a full clone just for this check? Could this be minimised into a shallow clone of the given ref, and failure means it doesn't exist? For that matter, if we only care about branches or tags (as is currently the case), [the docs imply we can get that without cloning at all](https://pkg.go.dev/github.com/Masterminds/vcs#GitRepo).

pkg/gitutils/gitutils.go
```diff
@@ -0,0 +1,47 @@
...
+package gitutils
+
+import (
+	"os"
+
+	"github.com/Masterminds/vcs"
+)
+
+func HasGitReference(gitRepo, ref, repoName string) (bool, error) {
+	local, err := os.MkdirTemp("", repoName)
```
TBBle, line 32: Is there a point in passing/using `repoName` just for this? The temp directory name doesn't really matter, I'd think?

pkg/downloader/chart_downloader.go
```diff
@@ -103,10 +113,17 @@ func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *proven
 	}
 
 	name := filepath.Base(u.Path)
-	if u.Scheme == registry.OCIScheme {
+	if u.Scheme == registry.OCIScheme || u.Scheme == "git" {
```
TBBle, line 116: This new condition doesn't seem to do anything, because if it's true, then `scheme` would also be `"git"`, and so the next block will immediately overwrite `name` anyway. I also don't see how this block would apply to git sources anyway, they don't separate the chart name and version with a `colon` like OCI sources.

pkg/downloader/manager.go
```diff
@@ -594,6 +611,16 @@ func (m *Manager) resolveRepoNames(deps []*chart.Dependency) (map[string]string,
 			reposMap[dd.Name] = dd.Repository
 			continue
 		}
+		// if dep chart is from a git url, assume it is valid for now.
+		// if the repo does not exist then it will later error when we try to fetch branches and tags.
+		// we could check for the repo existence here, but trying to avoid anotehr git request.
```
TBBle, line 616: nit: Typo "anotehr".

rblaine95:
Any specific reasons why this PR was closed?

Issue #10733, opened 2022-03-07
Creating a CronJob with PVC and a WaitForFirstConsumer SC times out


threesquared:
Output of `helm version`: `version.BuildInfo{Version:"v3.5.4", GitCommit:"1b5edb69df3d3a08df77c9902dc17af864ff05d1", GitTreeState:"clean", GoVersion:"go1.15.11"}`

Output of `kubectl version`: `Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.5", GitCommit:"aea7bbadd2fc0cd689de94a54e5b7b758869d691", GitTreeState:"clean", BuildDate:"2021-09-15T21:10:45Z", GoVersion:"go1.16.8", Compiler:"gc", Platform:"darwin/amd64"}`

Cloud Provider/Platform (AKS, GKE, Minikube etc.): `EKS`

When specifying a PVC on a CronJob using a StorageClass that uses the WaitForFirstConsumer [volume binding mode](https://kubernetes.io/docs/concepts/storage/storage-classes/#volume-binding-mode) as is recommended when using a topology-constrained storage backend like EBS. Helm upgrade will time out as the PVC Status is "Pending"

```
|   Type    Reason                Age                   From                         Message                                                                                                                   │
│   ----    ------                ----                  ----                         -------                                                                                                                   │
│   Normal  WaitForFirstConsumer  4m28s (x62 over 19m)  persistentvolume-controller  waiting for first consumer to be created before binding
```

However this is the expected success condition as the volume should not be bound until the cronjob runs.

```
apiVersion: batch/v1
kind: CronJob
metadata:
  name:cronjob
spec:
  jobTemplate:
    spec:
      template:
        metadata:
          name: cronjob
        spec:
          containers:
          - command:
            - /app/scripts/test.sh
            volumeMounts:
            - mountPath: /tmp/test
              name: backup
          volumes:
          - name: backup
            persistentVolumeClaim:
              claimName: pvc
----
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: ebs-sc
parameters:
  encrypted: "true"
  type: gp3
provisioner: ebs.csi.aws.com
reclaimPolicy: Delete
volumeBindingMode: WaitForFirstConsumer
----
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: pvc
spec:
  accessModes:
    - ReadWriteOnce
  storageClassName: ebs-sc
  resources:
    requests:
[...]

PR #11335, opened 2022-09-12
Consider PVC with a WaitForFirstConsumer SC ready

adrien-f:
As discussed in https://github.com/helm/helm/issues/10733.

Charts that have PVCs associated with resources not starting immediately after install (like a CronJob) will wait forever if the associated StorageClass has a volume binding mode of `WaitForFirstConsumer`.

This PR attempts to solve this by querying the associated StorageClass (or the default one) and checking its `volumeBindingMode` field.

Signed-off-by: Adrien Fillon <adrien.fillon@manomano.com>

* Consider PVCs with a SC with a volumeBindingMode of `WaitForFirstConsumer` to be ready

I'm not 100% sure of the UX Helm wish for this kind of behavior, I could also go for a flag similar to `WaitForJobs` to have a `SkipWaitForPvc` or something 🤔 

**If applicable**:

### DIFF
--- a/pkg/kube/ready.go
+++ b/pkg/kube/ready.go
@@ -18,6 +18,7 @@ package kube // import "helm.sh/helm/v3/pkg/kube"
 
 import (
 	"context"
+	"fmt"
 
 	appsv1 "k8s.io/api/apps/v1"
 	appsv1beta1 "k8s.io/api/apps/v1beta1"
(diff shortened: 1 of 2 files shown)


raxod502-plaid:
We encountered this bug in production. Is there any plan to review the PR solving it?

pkg/kube/ready.go
```diff
@@ -131,7 +135,13 @@ func (c *ReadyChecker) IsReady(ctx context.Context, v *resource.Info) (bool, err
 		if err != nil {
 			return false, err
 		}
-		if !c.volumeReady(claim) {
+
+		storageClass, err := getStorageClassFromPvc(ctx, c.client, claim)
```
gjenkins8, line 139: If I were to nitpick, perhaps `volumeReady` should (continue) to have all "volume ready" logic. And should call `getStorageClassFromPvc `.

(perhaps this would make `volumeReady` a candidate for refactoring itself, since adding this call would increase complexity there; but I don't think the right approach is to put some of the functionality into the caller here)

gjenkins8 (commented):
LGTMe (non-maintainer). Couple of minor comments around structure, but logic looks good.

ctron:
This is pretty annoying, and it looks like there's a fix pending for close to year now. Can this be merged please?

mananpreetsingh:
Can someone please look into it, Pending ro 2 years now.

Issue #8002, opened 2020-04-27
`tpl` function not performant


himmakam:
I have a scenario of having one umbrella chart with many sub-charts as dependencies. When I have the templates folder in umbrella chart, helm lint takes more time like almost 30 to 40 minutes. When this folder is not there, helm lint returns very fast. Can I know the reason.running in debug mode not giving any logs. It is helm v3.

hickeyma:
@himmakam Thanks for raising the issue.

How many sub-charts in the umbrella chart?

I have tested with 30 sub-charts and having a templates folder in the umbrella chart. `helm lint` is running and returning as expected. Would you be able to test with different number of sub-charts to see when the delay starts occurring? Have you nested sub-charts within sub-charts?

technosophos:
Providing a reproducible case would be great. I am currently working this sprint on `helm lint` issues, so I can take a look if someone can point me to a reproducible case.

technosophos:
I did recently uncover an issue where if you use `tpl` in your templates (a notoriously poorly performing function that should be avoided whenever possible), you can drive lint times quite high. Could that be the problem here?

himmakam:
> 
> 
> @himmakam Thanks for raising the issue.
> 
> How many sub-charts in the umbrella chart?
> 
> I have tested with 30 sub-charts and having a templates folder in the umbrella chart. `helm lint` is running and returning as expected. Would you be able to test with different number of sub-charts to see when the delay starts occurring? Have you nested sub-charts within sub-charts?

@hickeyma  I have a total of 120 sub-charts and do not have any nested sub-charts.

himmakam:
> 
> 
> I did recently uncover an issue where if you use `tpl` in your templates (a notoriously poorly performing function that should be avoided whenever possible), you can drive lint times quite high. Could that be the problem here?

Might be because I use tpl many times in these charts. Also helm install also takes almost 30 to 45 minutes to parse all the 120 sub-charts I have.

technosophos:
I did a test with a massive chart with hundreds of subcharts. I changed the `tpl` function to return an empty string. It cut that chart's `helm lint` down from 2.5 hours to 2.08 seconds. If you want to test reproducing this, all you have to do is comment out this: https://github.com/helm/helm/blob/master/pkg/engine/engine.go#L127-L152 and then recompile

himmakam:
[...]

PR #11351, opened 2022-09-16
Speed up `tpl`

greed42:

This dramatically improves the performance of the `tpl` function, resolving the issues described in https://github.com/helm/helm/issues/8002. Our typical charts go from 2-3 minute renders to 1-2 seconds. A particularly pathological one went from 37 minutes to 23 seconds.

It uses a clone of the invoking Go `Template` instance, rather that re-building everything from scratch, to isolate the effects of parsing the `tpl` text from the calling context. To allow `include`s of `define`s from the `tpl` text to continue to work, it also re-defines `include` within the clone so it can close over the clone itself.

The only user-noticeable change should be to performance: actual chart output is unaffected by this change. All tests are unaffected.

**If applicable**:


yxxhero:
@greed42 good job.

StephanDollberg:
@yxxhero can this be triaged/merged/looked at?

Seems like a massive benefit for a little change.

yxxhero:
@bacongobbler @hickeyma @joejulian @scottrigby  could you have free time to review this feature? this is a good change. Thanks very much.

joejulian:
How does this differ from #8371?

yxxhero:
@joejulian main logic is the same. I think we should discuss the PR in the meeting. WDYT?

joejulian:
Sounds good

yxxhero:
@joejulian my English is not good. So looking forward to your discussion. Thanks very much.

joejulian:
Actually, no meetings the rest of October.

If one isn't any better than the other, we should just add the older one to the next minor release milestone and close the newer.

yxxhero:
@joejulian I'm not familiar with this core logic of `tpl`. So I can't pick out which is better.

joejulian:
@greed42 @diafour any comment about the difference of these two PRs?

greed42:
#8371 permutes the existing code less, so is trivially better on that measure. It's also better for DRY-ness.

Mine also has a bug with `define` inside `tpl` that then calls `tpl` and `include`s that `define`, so #8371 is also better there.

greed42:
Though #8371 has an infinite recursion bug with a case that involves `.Files.Get` which my approach (just not re-using the filename:string render interface) doesn't suffer. I haven't gotten it isolated yet.

greed42:
`.Files.Get` is a possible red-herring for #8371; it's an empty string to `tpl` that's the problem, basically:

```
{{ tpl "" . }}
```

Except that PR looks to do `.New(name).Parse(tpl)` which I would expect to not suffer from that problem... except the sign you haven't done `.New(name)` is that:

```
 {{ tpl `{{ "" }` . }}
```

 also recurses and it doesn't.

greed42:
OK I've worked out that the difference between `t.Execute()` and `t.ExecuteTemplate()` is critical to how empty behaves. The latter will let "earlier" definitions leak through, as `t.Parse()` will always grab the previous value for a given name for the named association... while the "plain" template `t.Execute()` uses will see the empty text.

I think that means it is impractical to share Execute code between `tpl` and regular manifest expansion. `tpl` needs to be a special case, like `include` is a special case.

I will see if I can get some tests for the various edge cases I've observed with my chart library regression suite, and set them up as a separate PR so that this PR does not have additions for tests that cover existing behaviour.

joejulian:
Based on that analysis, I'll add #8371 to the v3.11.0 milestone and hope that the core maintainers have time to get it reviewed and merged.

joejulian:
I believe you're saying that #11456 is the better solution now? I'm happy to switch which PR's in the milestone. I can't duplicate the issue so I have to count on you folks that see the problem to guide me here.

greed42:
> I believe you're saying that #11456 is the better solution now? I'm happy to switch which PR's in the milestone. I can't duplicate the issue so I have to count on you folks that see the problem to guide me here.

No; #11456 is just tests for existing behaviour. It doesn't fix anything. #11351 (this one) is the fix I would now recommend.

The fault in #8371 is with this manifest or anything containing:

```
---
kind: Test
data: {{ tpl "" . }}
```

Or other define-only things like:

```
---
kind: Test
data: {{ tpl `{{define "foo"}}test{{end}}`  . }}
```

You get a stack overflow, as the empty template never becomes the template `ExecuteTemplate` sees, and so:

```
% bin/helm template test
runtime: goroutine stack exceeds 1000000000-byte limit
runtime: sp=0xc0eb46a380 stack=[0xc0eb46a000, 0xc10b46a000]
fatal error: stack overflow
...
```

mattfarina:
I like the speedup. We'll need to take a very careful look at this change as there is potential to break existing charts.

greed42:
> I like the speedup. We'll need to take a very careful look at this change as there is potential to break existing charts.

I definitely recommend getting https://github.com/helm/helm/pull/11456 merged, as then there will be tests which will show "expected" changes from this PR. And make it much easier to cover cases not already covered in that (test-only) PR.

davesmits:
I don't know how hard it would be with the infrastructure but maybe offer a preview build with this PR in it so people can help to test?

g-schaeffer:
+1 for at least a preview build. It's a mustT

FlamingFlowair:
This could drastically reduce our build time as we massively use tpl and umbrella chart to deploy an app with multiples components and micro services.

fan-tad:
is this feature still in review? Would like to have this in the next release as it would help up reduce the build time

tduverge:
I tried this change on my project with a huge chart umbrella (about 50 subcharts), it works and saves a lot of time.
We also use helm to generate yaml templates for non-kubernetes environment and this feature works as well.
This is a game changer as we were thinking of leaving helm.

yxxhero:
@joejulian @mattfarina need your help. thanks very much.

xeonx:
As mentioned above by original author @greed42, this PR and #11456 goes together.
#11456 add some tests for the `tpl` function, especially for scenario with `define` mixed together with `tpl`.

This PR changes some of those (unspecified) behaviors.
You can see the impact of this PR versus those new test cases here: https://github.com/xeonx/helm/commit/1ea011f3625f860943d627b0993392427afe5633#diff-b16f55bb6289c9df88deaedc99a45db09154ba13ee3a186b7c5636d67f5b345d

So I definitely recommend merging first #11456, then this PR will have to be updated to show the impact on the current behavior 
(mainly when there is a `define` override from within a `tpl`, of course impacting only inside `tpl` whereas in current implementation the globally defined `define` is used).

Also, for those interested I delivered an unofficial beta version of helm that include this change, so feel free to test with your charts and report any unexpected behavior.
https://github.com/xeonx/helm/releases/tag/v3.12.0-tpl-beta.1

TheBritz:
Geeking out about this, @greed42 - you're my hero! This took our 100+ umbrella chart render time from 4.5 min down to under 7 seconds.

Can I somehow send you a beer? Haha!

xeonx:
For those interested, as the milestone for this is now moved to 3.13 (September 2023), I will keep updating my fork and create releases allowing to test this change based on official latest helm release. 
https://github.com/xeonx/helm/releases/tag/v3.12.0-rc.1-fast-tpl
Feel free to ping me in the helm-dev slack for any question.

xeonx:
FYI, while testing this, I found an issue in the text/template package related to clone
https://github.com/golang/go/issues/60100
Thus until this bug is fixed (or a workaround is implemented in this PR) this PR will introduce a regression.

joejulian (dismissed):
Thanks for that @xeonx. I'm going to block this on upstream https://github.com/golang/go/issues/43022.

greed42:
> FYI, while testing this, I found an issue in the text/template package related to clone
> [golang/go#60100](https://github.com/golang/go/issues/60100)
> Thus until this bug is fixed (or a workaround is implemented in this PR) this PR will introduce a regression.

I've added a check to my test backfill PR for this case: https://github.com/helm/helm/pull/11456/commits/f235f0f28564b4391ef8b0b5f06b2d754bc13873

ashleydavies:
Just to add a voice to the chorus, this dramatically sped up Helm for us. We were using a home-grown function that avoids invoking `tpl` as much as possible by hacking at common cases (e.g. straightforward variable lookups were done without `tpl`), but it's impossible to avoid sometimes and also our chart dependencies use it very frequently 😄

For us, this took our chart that contains third-party charts from ~15 seconds per Helm operation to ~1s(!), and 

@greed42 Am I right in understanding that with your workaround, nothing is blocking this PR from being reviewed? If there's anything blocking that can be helped with some investigation and work I would be happy to try to help if I can.

greed42:
> @greed42 Am I right in understanding that with your workaround, nothing is blocking this PR from being reviewed? If there's anything blocking that can be helped with some investigation and work I would be happy to try to help if I can.

As far as I can tell, there shouldn't be anything blocking. If we got the test backfill PR (https://github.com/helm/helm/pull/11456) merged, then any possible issues could be raised as new checks which would cause this PR to fail testing.

ashleydavies:
🥳 

@joejulian as the assignee, are you able to take a look at the associated PR and this one? It seems like the author's got them to a great place and there's a reasonable amount of demand here for what is a really significant optimisation.

I have a feeling it'll have a much bigger impact than what's obvious from the reactions to this PR - a lot of people probably just assume Helm is just slow for big charts, unless they've spent a lot of time messing with their chart to discover the issue is `tpl` invocations. It took me quite a while to get to the bottom of it (luckily we added a lot of `tpl` in one go that was obvious in a `git bisect`, but in another world we never figured it out 😄)

[Sorry to @ you, I hope it's not considered rude! Wasn't sure what the workflow is here]

joejulian (approved):
It works for me. I don't have anything that takes more than 2 seconds, anyway, so I can't see the improvement, but there's enough feedback to give me confidence that it works for others.

greed42:
> It works for me. I don't have anything that takes more than 2 seconds, anyway, so I can't see the improvement, but there's enough feedback to give me confidence that it works for others.

Thanks. The new tests will fail after merging in main, so I've pushed a merge and the appropriate test changes to this branch.

stavros-k:
Here is some numbers from someone what has something big :)

```shell
/tmp/helm/bin/helm version   
version.BuildInfo{Version:"v3.12+unreleased", GitCommit:"b261a1b1bee93343cf9fe92335d3f1ccf3e24558", GitTreeState:"clean", GoVersion:"go1.20.6"}

time /tmp/helm/bin/helm template . > /dev/null
/tmp/helm/bin/helm template . > /dev/null  0,51s user 0,04s system 159% cpu 0,344 total
```

```shell
helm version
version.BuildInfo{Version:"v3.12.2", GitCommit:"1e210a2c8cc5117d1055bfaa5d40f51bbc2e345e", GitTreeState:"clean", GoVersion:"go1.20.5"}

time helm template . > /dev/null 
helm template . > /dev/null  22,30s user 0,45s system 147% cpu 15,452 total
```

stavros-k (approved):

yxxhero (approved):

yxxhero:
speed up!!!!

yxxhero:
@joejulian would you mind to talk the PR in the dev meeting?

starwarsfan:
What's the current state here? Will this be fixed with 3.13?

starwarsfan:
As 3.13.0 is released, the fix did not make it into it, right?

Kuzmenko-Pavel:
We look forward to speeding up!!!

HrechykhinVlad:
Please speed up, we need those changes. Thank you!

Lokker29:
We really need this improvement. Please, merge it!

Mixxxxerr:
Need this asap. Merge pls

DaniilOmelianenko:
Hey! This really cool! Speed up it, please.

oleksandrhrekov:
Looks cool! Speed it up, please.

FlamingFlowair:
No news on this, is it still under review ? we're almost 3 months after the first approval.
Hope this gets merged soon

Ornias1993:
@scottrigby and @mattfarina Can we get a final review from you two as well?

It seems most issues including previous upstream blockers have been dealth with and this PR would give a significant boost to making Helm more usefull for projects with more complex requirements relying on heavy tpl usage.

swythan:
Any update on getting this merged?

sabre1041 (approved):
LGTM

yxxhero:
Cool.

PR #11362, opened 2022-09-19
New feature to enable updating individual chart dependencies

### LINKED ISSUE #11308: helm dependency update should have an option to skip upgrades of existing dependencies
Output of `helm version`:
```
version.BuildInfo{Version:"v3.9.4", GitCommit:"dbc6d8e20fe1d58d50e6ed30f09a04a77e4c68db", GitTreeState:"clean", GoVersion:"go1.19"}
```

I have a couple of use cases that helm doesn't seem to handle very well right now.
1. I want to add a new dependency to Chart.yaml without touching existing dependencies.  Currently it seems that if I use version ranges such as 1.x, then I have to do an upgrade of this dependency when I add a new dependency.
2. I want to switch from using a specific dependency version to using a version range without upgrading.

One possibility for implementation would be to add a new sub-command `build dependency upgrade` which uses the current behaviour of `build dependency update`, and then change `update` so that it will only add/delete dependencies and update the checksum, but it will not upgrade existing dependencies.

pgier:
Closes #11308

It's currently not possible to update a single Chart dependency if ranges are used.  The `helm dependency update` command will update all dependencies.  This PR adds a new option called `--dependency` which allows one or more chart dependencies to be selected for update.  The remaining unselected dependencies in Chart.yaml will remain at the version currently set by Chart.lock.

**If applicable**:
- [ X ] this PR contains documentation
- [ X ] this PR contains unit tests
- [ X ] this PR has been tested for backwards compatibility

### DIFF
--- a/pkg/downloader/manager.go
+++ b/pkg/downloader/manager.go
@@ -71,6 +71,8 @@ type Manager struct {
 	Keyring string
 	// SkipUpdate indicates that the repository should not be updated first.
 	SkipUpdate bool
+	// Dependencies is a list of specific dependencies to update instead of updating all.
+	Dependencies []string
 	// Getter collection for the operation
 	Getters          []getter.Provider
 	RegistryClient   *registry.Client
@@ -124,7 +126,7 @@ func (m *Manager) Build() error {
 			if v2Sum != lock.Digest {
 				return errors.New("the lock file (requirements.lock) is out of sync with the dependencies file (requirements.yaml). Please update the dependencies")
 			}
-		} else {
+		} else if len(m.Dependencies) == 0 {
 			return errors.New("the lock file (Chart.lock) is out of sync with the dependencies file (Chart.yaml). Please update the dependencies")
 		}
 	}
@@ -151,6 +153,7 @@ func (m *Manager) Build() error {
 // negotiate versions based on that. It will download the versions
 // from remote chart repositories unless SkipUpdate is true.
 func (m *Manager) Update() error {
+
 	c, err := m.loadChartDir()
 	if err != nil {
 		return err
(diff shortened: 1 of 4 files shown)


joejulian (approved):
This looks good to me, but core maintainers must approve.

pkg/downloader/manager.go
```diff
@@ -896,3 +906,54 @@ func key(name string) (string, error) {
...
+func dependenciesToResolve(req []*chart.Dependency, lock *chart.Lock, selected []string) ([]*chart.Dependency, error) {
+	if len(selected) == 0 {
+		return req, nil
+	}
+	if lock == nil {
+		return nil, fmt.Errorf("file Chart.lock not found, the '-d/--dependency' requires an existing Chart.lock file")
+	}
+	depsToUpdate := []*chart.Dependency{}
+	for _, dep := range req {
+		if stringSliceContains(selected, dep.Name) {
```
gjenkins8, line 926: A chart can have multiple dependencies of the same name, but with different [aliases](https://github.com/helm/helm/blob/42ef0fe2ff18935ee9600de700b1ac0a6f54850e/pkg/chart/dependency.go#L49).

Shouldn't this logic should account for aliases? I think in that if an alias is present, that should be matched, rather than the name.

gjenkins8 (commented):
This I think is a nice change. 

Comment on matching aliases. And having some test cases to show this functionality works as expected I think would be great.

dklimpel:
I find the feature very interesting.
Where does the PR stand or who still has to do what so that it can continue?

MrNocTV:
Please help to add this feature, it's a very nice to have.
Updating a specific dependencies means a lot if you have a legacy project with a huge helm, where update to a specific sub-chart or component is done manually and you want to move all of the sub-chart out and use as dependencies.

mattfarina:
@pgier could you please rebase and fix the conflicts.

hmhuan:
Please add this feature because we really need it for our project. Thank you so much

violuke:
Not stale

scottrigby:
Needs rebase. Note Helm 3 is not accepting new features, so this is for Helm 4.

PR #11363, opened 2022-09-20
Enable custom certificates option for OCI interaction

souleb:
Signed-off-by: Soule BA <bah.soule@gmail.com>

If implemented, users will be able to use custom certificates and CA when interacting with OCI registries.

This is convenient for user interacting with registries that hold custom signed certificates. By providing the `CA` the server certificate can be validate during handshakes.

I believed this is also useful for mutual `tls` validation.

For the `oras` client, I use an `option` to provide a custom `httpClient` that can then be passed to `authorizer` and `resolver`. I believe that it is more flexible this way, as sdk user can construct their client with any needed custom transport.

I tested this by setting up a docker registry in an ec2 server with a certificate and CA generated with openSSL.

Then on my local machine I tested the followinig helm cmd with the `--ca-file` set with the custom `CA`:
- helm push
- helm pull
- helm show
- helm template
- helm install
- helm upgrade

cc @sabre1041 @scottrigby


souleb:
the test fails on circle-ci because:

`Expected nil, but got: &url.Error{Op:"Get", URL:"https://0x7f000001:35715/v2/", Err:(*net.OpError)(0xc000550e60)}``

`0x7f000001` is not valid for `127.0.0.1` in their runner...

sabre1041 (approved):
Merging into feature branch. Roadmap for CircleCI is to hopefully migrate to GitHub actions soon enough, but can revisit issue prior to merging to `main` branch

PR #11396, opened 2022-09-29
bugfix: unexpected behavior when using "dependencies.condition"

### LINKED ISSUE #11338: Unexpected behavior when using "dependencies.condition"
Hi, this is supposed to be a bug report (or at least I am looking for an explanation for my observed behavior).

**Current state**

Given I have this in my `Chart.yaml` file. Two dependencies that have same alias, but with different conditions:

```
dependencies:
  - name: hello-world
    version: 0.1.0
    alias: alias-a
    repository: 'https://helm.sikalabs.io'
    condition: conditionA

  - name: hello-world
    version: 0.2.0
    alias: alias-a
    repository: 'https://helm.sikalabs.io'
    condition: conditionB
```

When I then run `helm install --set conditionA=false --set conditionB=true demo helm --namespace demo`

Then **no deployment** is created at all.

**My expected state**
Then **one deployment** with version `0.2.0` is created, because `conditionB` is `true`.

**Comment**
Maybe helm stops after `conditionA` for same alias `alias-a` was `false`? But I don't understand it. I want to make use of this constellation to deploy my alias in different variants via one `Chart.yaml` file.

----

Output of `helm version`:
version.BuildInfo{Version:"v3.9.4", GitCommit:"dbc6d8e20fe1d58d50e6ed30f09a04a77e4c68db", GitTreeState:"clean", GoVersion:"go1.19"}

Output of `kubectl version`:
Client Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.5", GitCommit:"c285e781331a3785a7f436042c65c5641ce8a9e9", GitTreeState:"clean", BuildDate:"2022-03-16T15:51:05Z", GoVersion:"go1.17.8", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.4", GitCommit:"b695d79d4f967c403a96986f1750a35eb75e75f1", GitTreeState:"clean", BuildDate:"2021-11-17T15:42:41Z", GoVersion:"go1.16.10", Compiler:"gc", Platform:"linux/amd64"}

Docker for Desktop

wujunwei:
Signed-off-by: wujunwei <wjw3323@live.com>

fixes #11338

**If applicable**:


pkg/chartutil/dependencies.go
```diff
@@ -158,30 +158,43 @@ Loop:
 	processDependencyTags(c.Metadata.Dependencies, cvals)
 	processDependencyConditions(c.Metadata.Dependencies, cvals, path)
 	// make a map of charts to remove
-	rm := map[string]struct{}{}
+	rm := map[string][]string{}
 	for _, r := range c.Metadata.Dependencies {
 		if !r.Enabled {
 			// remove disabled chart
-			rm[r.Name] = struct{}{}
+			rm[r.Name] = append(rm[r.Name], r.Version)
```
kszpakowski, line 165: Hi @wujunwei, would you please take into consideration using chart alias from the dependencies as well? When I have a helm chart with two dependencies for chart X, version 0.0.1, but with different aliases, it will still break conditionals.

Chart.yaml
```yaml
apiVersion: v2
name: my-app-configuration
version: 0.0.1
dependencies:
- name: my-service-chart
  alias: serviceOne
  version: 0.3.0
  repository: "@stable"
- name: my-service-chart
  alias: serviceTwo
  version: 0.3.0
  repository: "@stable"
  
```

wujunwei: In fact, I have considered this situation.
As the code below shows, when deal with package's name ,it have considered the alias name.
[](https://github.com/helm/helm/blob/623f4c6248c2823b9b18ba1956df6bc4d8f8d22e/pkg/chartutil/dependencies.go#L110)

wujunwei: @kszpakowski  I added some test cases To prove my point. PTAL :)

kszpakowski: Thanks, your change looks good to me. Unfortunately it looks like I'm facing another issue and it still does not work on your version. Basically, I have a chart with two dependencies to another chart, wich then have its own conditional dependencies disabled by default in its values.yaml. When I render it, all resources are created ignoring conditions.  I'll create example repo and open new issue.

kszpakowski: I've just found the solution in the docs: "Tags and conditions values must be set in the top parent's values." :)

SaeidAshian: > top parent's values

Could you send and example

wujunwei: >  the first time getAliasDependency is called and it modifies dep name to the alias (md.Name = dep.Alias). When it is called again, dep name is already set to an alias, and it doesn't match c.Name (Line 98)

why does it not match c.Name? when called again, it's another `dep`.

>  the metadata name is not being changed to an alias. Then there is a mismatch between rm content and it cannot get the right value from rm in line 173

`n.Metadata.Name` in line 173 comes from c.Dependencies(0 which reset by line 147, it has already changed to alias  from the code point of view.

Have you test it? could you provide a test case that can reproduct it ? thanks a lot !

kszpakowski: I made this sample to reproduce the error: https://github.com/kszpakowski/helm-alias-condition-error

wujunwei: > I made this sample to reproduce the error: https://github.com/kszpakowski/helm-alias-condition-error

Thanks for this demo, It can be produced at the latest `main branch`
phenomenon : when I run `helm template chartC` output contained sa.yaml which is disabled at chartC's `values.yaml` . 
```
> helm  template ./helm-alias-condition-error/chartC/
---
# Source: chartC/charts/firstApp/templates/ns.yaml
apiVersion: v1
kind: Namespace
metadata:
  name: test
---
# Source: chartC/charts/secondApp/templates/ns.yaml
apiVersion: v1
kind: Namespace
metadata:
  name: test
---
# Source: chartC/charts/secondApp/charts/chartA/templates/sa.yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: test

```
After Looking into the code ,I found , It's cause by code below. it tried to copy a chart, but the `medata.dependencies`  is a array of chart point,it will copy reference not truly struct. as a result they will share same address, when change the name of `firstApp.chartA` to `firstApp.chartAalias` `secondApp.chartA` will be set to `secondApp.chartAalias` at the same time.
https://github.com/helm/helm/blob/488606613e97d5622043b7d40ec08df6d9beb449/pkg/chartutil/dependencies.go#L105-L112

bug

It's a new bug(although the phenomenon is same as this issue), you can submit a new issue to trace it .

wujunwei:
@mattfarina @joejulian  could anyone take a look ?  Another new user has encountered the same bug again!

PR #11397, opened 2022-09-29
bugfix: helm lint infinite loop

### LINKED ISSUE #11391: helm lint stuck with malformed template object
Hi

Output of `helm version`:

> version.BuildInfo{Version:"v3.10.0", GitCommit:"ce66412a723e4d89555dc67217607c6579ffcb21", GitTreeState:"clean", GoVersion:"go1.19.1"}

With this template file (notice the space between the first and the second curly brackets): 
`{ {- $relname := .Release.Name -}}`

helm lint does not respond whereas helm template throws an error (which is the normal behaviour) :
> Error: YAML parse error on onboarding-ms/templates/secret.yaml: error converting YAML to JSON: yaml: did not find expected node content
Use --debug flag to render out invalid YAML`

wujunwei:
Signed-off-by: wujunwei <wjw3323@live.com>

Fixes #11391

**If applicable**:

### DIFF
--- a/pkg/lint/lint_test.go
+++ b/pkg/lint/lint_test.go
@@ -19,6 +19,7 @@ package lint
 import (
 	"strings"
 	"testing"
+	"time"
 
 	"helm.sh/helm/v3/pkg/chartutil"
 	"helm.sh/helm/v3/pkg/lint/support"
@@ -34,6 +35,7 @@ const badValuesFileDir = "rules/testdata/badvaluesfile"
 const badYamlFileDir = "rules/testdata/albatross"
 const goodChartDir = "rules/testdata/goodone"
 const subChartValuesDir = "rules/testdata/withsubchart"
+const malformedTemplate = "rules/testdata/malformed-template"
 
 func TestBadChart(t *testing.T) {
 	m := All(badChartDir, values, namespace, strict).Messages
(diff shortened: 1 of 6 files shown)


pkg/lint/rules/template.go
```diff
@@ -141,10 +141,11 @@ func Templates(linter *support.Linter, values map[string]interface{}, namespace
 					break
 				}
 
-				// If YAML linting fails, we sill progress. So we don't capture the returned state
-				// on this linter run.
-				linter.RunLinterRule(support.ErrorSev, fpath, validateYamlContent(err))
-
+				// If YAML linting fails, it will always be fails next block. so we should return here.
```
yardenshoham, line 144:
```suggestion
				// If YAML linting fails here, it will always fail in the next block as well, so we should return here.
```

yardenshoham (changes requested):

pkg/lint/rules/template.go
wujunwei: thanks for this good suggestion :) !

wujunwei:
This PR may fix a serious bug, could anyone have a look?

wujunwei:
> @wujunwei sure. could you add some tests for this case? Thanks very much.

I have add some test ,  PTAL :)

pkg/lint/lint_test.go
```diff
@@ -151,3 +153,26 @@ func TestSubChartValuesChart(t *testing.T) {
 		}
 	}
 }
+
+// lint stuck with malformed template object
+// See https://github.com/helm/helm/issues/11391
+func TestMalformedTemplate(t *testing.T) {
+	c := time.After(3 * time.Second)
```
yxxhero, line 160: why 3s?

wujunwei: > why 3s?

Before repair in the code, it will cause infinite loop when lint the malformed template . So  we wait it for 3s to test wheather a  infinite loop happen or not.  

It doesn't have to be 3s. Do you have any better suggestions?

yxxhero: @wujunwei ok. got it. Thanks very much.

wujunwei:
I think this is a critical bug which should be fix as soon as possible.

mattfarina (approved):

sabre1041 (approved):
LGTM

Issue #11385, opened 2022-09-26
Malformed JSON output


stan-sz:
Output of `helm version`: 3.10.0

>helm upgrade foo-bar oci://foo.azurecr.io/helm/foo-bar --install --namespace foo --version 1.0.0 --output json

yields

```json
Pulled: foo.azurecr.io/helm/foo-bar:1.0.0
Digest: sha256:f4bcf923e259199f5ef3506e3edea83a70f3474be5d94ead6bae4654ef1b91e7
{
    "name": "foo-bar",
    "info": {/*content*/},
    "chart": {/*content*/}
}
```

So there are extra two lines before the regular json response. This is due to:

https://github.com/helm/helm/blob/c52cd40acef183d29f6def5026e66616b9993c88/pkg/registry/client.go#L419-L420

nilsalex:
I have a similar issue. We use output from `helm show values oci:///foo.io/bar --version 1.0.0 --jsonpath '{.foo.bar}'`, which with helm version 3.9.4 used to be just the value of `.foo.bar`.

With 3.10.0, we get
```
Pulled: ...
Digest: ...
<value of .foo.bar>
```

I did a bisect and the issue starts to appear with this commit: https://github.com/helm/helm/commit/9f199b6517c21394bca555983c70fc232d65014c

yxxhero:
@nilsalex Thanks very much. I will test this issue.

yxxhero:
[root@devops ~]# helm show values oci://localhost:5000/aaa/abc --version 0.1.0 --jsonpath '{.autoscaling.enabled}'
Pulled: localhost:5000/aaa/abc:0.1.0
Digest: sha256:fc93deb645f85bd9ecb46f570927702f9440a9331cb0aa0a3cb746ad27656eab
false[root@devops ~]#

false[root@devops ~]# helm version
version.BuildInfo{Version:"v3.10.0", GitCommit:"ce66412a723e4d89555dc67217607c6579ffcb21", GitTreeState:"clean", GoVersion:"go1.18.6"}

@nilsalex I can't reproduce your issue.

d-d-up:
@yxxhero I want to fix it.

stan-sz:
Please note that there are more cases of such message logging in that file.

wernerb:
@d-d-up This is a larger issue. Since helm 3.10.0 it makes any output from `helm template` invalid kubernetes as well.
```bash
[...]

PR #11423, opened 2022-10-08
fix the problem of  json output

d-d-up:
Signed-off-by: d-d-up <qhr6113@163.com>

Fixes: issue https://github.com/helm/helm/issues/11385

**If applicable**:

### DIFF
--- a/pkg/registry/client.go
+++ b/pkg/registry/client.go
@@ -23,6 +23,7 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	"os"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


pkg/registry/client.go
```diff
@@ -416,8 +417,8 @@ func (c *Client) Pull(ref string, options ...PullOption) (*PullResult, error) {
 		}
 	}
 
-	fmt.Fprintf(c.out, "Pulled: %s\n", result.Ref)
-	fmt.Fprintf(c.out, "Digest: %s\n", result.Manifest.Digest)
+	fmt.Fprintf(os.Stderr, "Pulled: %s\n", result.Ref)
```
a-palchikov, line 420: This kind of defeats the purpose of the `out` attribute and it does not solve the issue completely as below you can see another write to `c.out` (in this case it is a warning/error).
I would suggest to configure os.Stderr at the [root](https://github.com/helm/helm/blob/ce66412a723e4d89555dc67217607c6579ffcb21/cmd/helm/root.go#L158):

```go
		registry.ClientOptWriter(os.Stderr),
```

yxxhero:
@d-d-up ping

jouve:
as a side effect, this breaks [helmfile/helmfile](https://github.com/helmfile/helmfile/) template

yxxhero:
close this PR, see: https://github.com/helm/helm/pull/11480

PR #11436, opened 2022-10-12
Updating the deb location for azure cli

mattfarina:
This update moves to ubuntu jammy. The name moves off of debian version names because jammy is based on bookworm and azure cli does not have debian package support with bookworm

This is an update to the previous update for this because the CircleCI environment is ubuntu 20.04 but the container this is running is has 22.04 as the ubuntu version. Updating the in container to 22.04 (jammy).


sabre1041 (approved):
LGTM

PR #11456, opened 2022-10-20
Backfill tests for the `tpl` function

greed42:

This PR contains test-only changes for the `tpl` function to cover all the sharp edges I encountered creating PR <http://github.com/helm/helm/pull/11351> and trying to use <http://github.com/helm/helm/pull/8371> instead.

It verifies current behaviour, especially the aspects affected by those PRs. (Or which I found mattered to our charts before composing <http://github.com/helm/helm/pull/11351>.)

I have run these tests against both v3.9.4 and v3.10.1 tags as well as the HEAD of main.

Both proposed `tpl` performance improvement PRs will fail after this PR is merged. I have a second set of cherries to pick onto <http://github.com/helm/helm/pull/11351>; I believe the changes are reasonable. (They involve access to `define` in the manifest invoking `tpl` and being able to override `define`s from partial files.)

 However, due to the nature of the `(t) ExecuteTemplate(...)` golang function, I now think that <http://github.com/helm/helm/pull/8371> cannot be used without effectively turning it into something very similar to <http://github.com/helm/helm/pull/11351>: it must use `(t) Execute(...)` instead.

**If applicable**:

### DIFF
--- a/pkg/engine/engine_test.go
+++ b/pkg/engine/engine_test.go
@@ -947,3 +947,54 @@ func TestRenderTplTemplateNames(t *testing.T) {
 		}
 	}
 }
+
+func TestRenderTplRedefines(t *testing.T) {
+	// Redefining a template inside 'tpl' does not affect the outer definition
+	c := &chart.Chart{
+		Metadata: &chart.Metadata{Name: "TplRedefines"},
+		Templates: []*chart.File{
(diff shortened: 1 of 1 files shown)


xeonx:
In order to progress on the `tpl` performance issue (cf. https://github.com/helm/helm/issues/8002 and https://github.com/helm/helm/pull/11351) would it be possible to review and merge this PR?

joejulian (approved):
This is great!

sabre1041 (approved):
LGTM

PR #11459, opened 2022-10-21
Validates against OpenAPI schema while rolling back to previous release

nitinmalik00:
Validates against OpenAPI schema while rolling back to the previous release.
Also, i think this validation should be part of dry-run itself, so in dry-run only we can catch this behaviour.
There's always a possibility that user can corrupt the Helm revisions secrets, stored in users namespace only and to avoid that we should validate it before applying/rolling back it. 

Any advice is greatly appreciated.

**If applicable**:

### DIFF
--- a/pkg/action/rollback.go
+++ b/pkg/action/rollback.go
@@ -146,11 +146,12 @@ func (r *Rollback) performRollback(currentRelease, targetRelease *release.Releas
 		return targetRelease, nil
 	}
 
-	current, err := r.cfg.KubeClient.Build(bytes.NewBufferString(currentRelease.Manifest), false)
+	current, err := r.cfg.KubeClient.Build(bytes.NewBufferString(currentRelease.Manifest), true)
 	if err != nil {
 		return targetRelease, errors.Wrap(err, "unable to build kubernetes objects from current release manifest")
 	}
-	target, err := r.cfg.KubeClient.Build(bytes.NewBufferString(targetRelease.Manifest), false)
+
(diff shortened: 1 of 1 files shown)


bacongobbler:
Prior versions of Helm allowed objects that were invalid against the OpenAPI schema, but Kubernetes still accepted them as valid resources.

To enable this now would break every Helm user who deployed charts with invalid resources.

nitinmalik00:
Got it.

1. Since there is always the possibility that changes can happen in k8s apis and this situation (in `rollback`) may occur at any time.
Eg., let's say there is support for _k1_ attribute in any one of the api in k8s (at time of helm install/upgrade) but in next release of k8s they take it out this attribute (_k1_) and now when we rollback with openAPI validation enabled helm will not allow to rollback to version which contains that attribute (_k1_). right ?

So, that's why we are not enabling the openAPI validation in rollback action and will keep it disabled always ?

2. Also, shouldn't the _Build_ (creates a resource list) must be included in the dry-run path ? 
Dry-run should catch every possible error while installing/upgrading/rolling back, which gives more confidence to developer for their releases.

Please correct me if i'm wrong anywhere ;)

nitinmalik00:
Regarding openAPI validation in rollback action.
What I think is, shouldn't we give this as a flag or parameter to user, so it gives more capability to user, to choose from what they want to do ?
Just like we do in install and upgrade action... there we have _DisableOpenAPIValidation_ parameter.
What do you think ?

nitinmalik00:
@bacongobbler Any thoughts on this?

PR #11460, opened 2022-10-21
feat(package): add value options to 'helm package'

### LINKED ISSUE #3141: add --values and --set flag to helm package

cndoit18:

Closes #3141

The parsing of kyaml itself will save the comments. The problem of missing comments can be solved by merging them with kyaml's merge.

**If applicable**:

### DIFF
--- a/cmd/helm/package_test.go
+++ b/cmd/helm/package_test.go
@@ -25,6 +25,7 @@ import (
 
 	"helm.sh/helm/v4/pkg/chart"
 	"helm.sh/helm/v4/pkg/chart/loader"
+	"helm.sh/helm/v4/pkg/chartutil"
 )
 
 func TestPackage(t *testing.T) {
@@ -188,6 +189,173 @@ func TestSetAppVersion(t *testing.T) {
 	}
 	if ch.Metadata.AppVersion != expectedAppVersion {
 		t.Errorf("expected app-version %q, found %q", expectedAppVersion, ch.Metadata.AppVersion)
+	}
+}
+
+func TestAddValueOptions(t *testing.T) {
+	tests := []struct {
+		name   string
+		args   []string
+		expect string
+		chart  string
+		err    bool
+	}{
+		{
+			name: "package without set value",
+			args: []string{},
+			expect: `# The pod name
+Name: my-alpine
+List: []
+List1:
+- a
+`,
+			chart: "issue-3141",
+			err:   false,
+		},
+		{
+			name: "package set values",
+			args: []string{
+				"--set Name=other",
+			},
+			chart: "issue-3141",
+			expect: `# The pod name
+Name: other
+List: []
+List1:
+- a
+`,
+			err: false,
+		},
+		{
+			name: "package append values",
+			args: []string{
+				"--set Key=other",
+			},
+			chart: "issue-3141",
+			expect: `# The pod name
+Name: my-alpine
+List: []
+List1:
+- a
+Key: other
+`,
+			err: false,
+		},
+		{
+			name: "package multiple values",
+			args: []string{
+				"--set a=b,c=d",
+			},
+			chart: "issue-3141",
+			expect: `# The pod name
+Name: my-alpine
+List: []
+List1:
+- a
+a: b
+c: d
+`,
+			err: false,
+		},
+		{
+			name: "package more complex expressions",
+			args: []string{
+				"--set outer.inner=value",
+			},
+			chart: "issue-3141",
+			expect: `# The pod name
+Name: my-alpine
+List: []
+List1:
+- a
+outer:
+  inner: value
+`,
+			err: false,
+		},
+		{
+			name: "package lists values",
+			args: []string{
+				"--set List={a,b,c}",
+			},
+			chart: "issue-3141",
+			expect: `# The pod name
+Name: my-alpine
+List:
+- a
+- b
+- c
+List1:
+- a
+`,
+			err: false,
+		},
+		{
+			name: "package replace values",
+			args: []string{
+				"--set List1={a,b,c}",
+			},
+			chart: "issue-3141",
+			expect: `# The pod name
+Name: my-alpine
+List: []
+List1:
+- a
+- b
+- c
+`,
+			err: false,
+		},
+		{
+			name: "package no values file",
+			args: []string{
+				"--set List={a,b,c}",
+			},
+			chart: "issue-3141-without-values",
+			expect: `List:
+- a
+- b
+- c
+`,
+			err: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			dir := t.TempDir()
+			t.Cleanup(func() {
+				os.RemoveAll(dir)
+			})
+
+			chartToPackage := "testdata/testcharts/" + tt.chart
+			cmd := fmt.Sprintf("package %s --destination=%s %s", chartToPackage, dir, strings.Join(tt.args, " "))
+			_, _, err := executeActionCommand(cmd)
+			if err != nil {
+				if tt.err {
+					return
+				}
+				t.Fatal(err)
+			}
+			if tt.err {
+				t.Fatalf("Expected error in test %q", tt.name)
+			}
+			chartPath := filepath.Join(dir, tt.chart+"-0.1.0.tgz")
+			if fi, err := os.Stat(chartPath); err != nil {
+				t.Errorf("expected file %q, got err %q", chartPath, err)
+			} else if fi.Size() == 0 {
+				t.Errorf("file %q has zero bytes.", chartPath)
+			}
+			ch, err := loader.Load(chartPath)
+			if err != nil {
+				t.Fatalf("unexpected error loading packaged chart: %v", err)
+			}
+			for _, f := range ch.Raw {
+				if f.Name == chartutil.ValuesfileName && string(f.Data) != tt.expect {
+					t.Errorf("expected values %q, found %q", tt.expect, string(f.Data))
+				}
+			}
+		})
 	}
 }
 
--- a/pkg/action/package.go
+++ b/pkg/action/package.go
@@ -26,6 +26,10 @@ import (
 	"github.com/pkg/errors"
 	"golang.org/x/term"
 
+	kyaml "sigs.k8s.io/kustomize/kyaml/yaml"
+	"sigs.k8s.io/kustomize/kyaml/yaml/merge2"
+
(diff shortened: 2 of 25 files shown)


cndoit18:
This PR will solve this comment
https://github.com/helm/helm/pull/3471#issuecomment-381779783

yxxhero:
@cndoit18 good job. I think we should add more descriptions for this PR.  Not necessarily familiar with this library sigs.k8s.io/kustomize/kyaml.

cmd/helm/package_test.go
```diff
@@ -192,6 +193,120 @@ func TestSetAppVersion(t *testing.T) {
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			dir := ensure.TempDir(t)
+			t.Cleanup(func() {
+				os.RemoveAll(dir)
+			})
+
+			chartToPackage := "testdata/testcharts/issue1979"
```
yxxhero, line 279: @cndoit18 why use issue1979 chart?

cndoit18: updated

pkg/action/package.go
```diff
@@ -92,6 +96,50 @@ func (p *Package) Run(path string, vals map[string]interface{}) (string, error)
 		dest = p.Destination
 	}
 
+	// If vals is not empty and the values file does not exist, then we need to generate a values file.
```
yxxhero, line 99:
```suggestion
	// If vals is not empty and the values.yaml file does not exist, then we need to generate a values.yaml file.
```

cndoit18: done

ritsuki1227:
@yxxhero Thank you for reviewing and contributing to Helm, it has been a great help to me! I would appreciate it if this feature could be merged in for the 3.12 release, as it would be very helpful for my project. Thank you again!

dromadaire54:
I just want to say that this feature was available in the previous helm version I don't remember the version exactly. I don't know why it has been deleted. This is an important feature for the CI to build helmchart and set the image tag values.

ritsuki1227:
Is there any plan to merge this PR?
#3141 is one of the most commented open issues in helm. I think many users are awaiting this feature!

dhirschfeld:
@cndoit18 has been keeping this branch up-to-date for going on 2 years now. I imagine it's pretty disheartening to see it has conflicts again because other changes have been merged in the interim.

It would be great if a maintainer could comment on what more needs to be done to get this PR merged?

sabre1041 (dismissed):
As others have commented on previously, this feature will be very useful when needing to update values dynamically as part of a build/CI system. 

Not sure if its a big issue, but one issue with this feature is since values are being added dynamically at packaging time, there will be a loss of clarity into the values that are published to the chart repository/registry versus what is defined in SCM.

The benefits of this feature outweighs the detriments

yxxhero (approved):

wyardley:
Hi @sabre1041 @yxxhero -- seems like this has been approved in the past, and seems like this is the most viable solution -- is there anything else pending / outstanding that @cndoit18 needs to do, or are there any additional approvals needed, other than for someone to resolve the conflicts? As @dhirschfeld mentioned, this PR has been open for over 2 years, and the problem that it fixes has been unresolved for over 7.

It seems like it missed the 3.16 milestone - is it on track to hit the 3.17 one still?

sabre1041:
> Hi @sabre1041 @yxxhero -- seems like this has been approved in the past, and seems like this is the most viable solution -- is there anything else pending / outstanding that @cndoit18 needs to do, or are there any additional approvals needed, other than for someone to resolve the conflicts? As @dhirschfeld mentioned, this PR has been open for over 2 years, and the problem that it fixes has been unresolved for over 7.
> 
> It seems like it missed the 3.16 milestone - is it on track to hit the 3.17 one still?

unfortunately, pushing new changes dismisses prior reviews which has caused for some of the delay. when the latest conflicts have been resolved, we can have the maintainers (myself included), review again.  apologies for the delay

cndoit18:
> > Hi @sabre1041 @yxxhero -- seems like this has been approved in the past, and seems like this is the most viable solution -- is there anything else pending / outstanding that @cndoit18 needs to do, or are there any additional approvals needed, other than for someone to resolve the conflicts? As @dhirschfeld mentioned, this PR has been open for over 2 years, and the problem that it fixes has been unresolved for over 7.
> > It seems like it missed the 3.16 milestone - is it on track to hit the 3.17 one still?
> 
> unfortunately, pushing new changes dismisses prior reviews which has caused for some of the delay. when the latest conflicts have been resolved, we can have the maintainers (myself included), review again. apologies for the delay

updated

cndoit18:
> > Hi @sabre1041 @yxxhero -- seems like this has been approved in the past, and seems like this is the most viable solution -- is there anything else pending / outstanding that @cndoit18 needs to do, or are there any additional approvals needed, other than for someone to resolve the conflicts? As @dhirschfeld mentioned, this PR has been open for over 2 years, and the problem that it fixes has been unresolved for over 7.
> > It seems like it missed the 3.16 milestone - is it on track to hit the 3.17 one still?
> 
> unfortunately, pushing new changes dismisses prior reviews which has caused for some of the delay. when the latest conflicts have been resolved, we can have the maintainers (myself included), review again. apologies for the delay

Updated

sabre1041 (approved):
LGTM

wyardley:
Amazing. @yxxhero would you be able to reapprove also?

yxxhero (approved):
LGTM

pkg/action/package.go
```diff
@@ -98,6 +102,51 @@ func (p *Package) Run(path string, _ map[string]interface{}) (string, error) {
...
+	for _, f := range ch.Raw {
+		// Always run to ensure that the values.yaml file is formatted.
+		if f.Name == chartutil.ValuesfileName {
+			dest, err := kyaml.Parse(string(f.Data))
+			if err != nil {
+				return "", err
+			}
+
+			// In the case of saving yaml comments, merges fields from src into dest.
+			rnode, err := merge2.Merge(kyaml.NewRNode(src), dest, kyaml.MergeOptions{})
```
gjenkins8, line 122: Helm has specific logic for "merging" values. Using two different library/methology will I think ultimately have some behavioral differences between the two. Which will confuse / cause issues for users.

An alternative here I think would be to simply append a new yaml document onto the chart's existing values.yaml. Then the values merging logic could merge that when releasing the chart

gjenkins8: (I'm not sure / can't remember directly whether we support multiple yaml documents in a values.yaml, but that would be a worthy additional regardless. I vaguely remember a recent'ish issue asking for support)

wyardley: > Helm has specific logic for "merging" values. Using two different library/methology will I think ultimately have some behavioral differences between the two. Which will confuse / cause issues for users.

If that merging logic / behavior is critical, maybe there should be some test coverage added by the helm maintainers (separately from this PR) that covers these cases, since it seems nothing is failing from these changes? Is this a theoretical concern, or are there specific things you're worried about happening here?

> An alternative here I think would be to simply append a new yaml document onto the chart's existing values.yaml.

Just an outside observer here. But I would just say that often, saying "just" or "simply" doing something can gloss over implementation details that may end up making something less than "simple". In particular, if the logic to merge two documents isn't already there, that would seem to make this much less "simple", especially if there are also testing gaps.

The author of this PR has jumped through a lot of hoops, and gone through a lot of cycles of resolving conflicts, requests for changes, etc. Previous authors of PRs trying to fix this issue have seemingly gotten mired in a lot of red tape (and inflexibility from maintainers) as well, not to mention arguments and resistance around some of the third party libraries in use to handle all of this yaml parsing / merging. I just wonder if it may be worth considering letting the perfect not be the enemy of the good here, especially since no one seems to have brought up these concerns previously?

If this multiple document vaules file would actually be inspected by users of the final packed chart, I think that would also be confusing to end-users, for example, in the scenario where a single value is set near the top of a large `values.yaml`, then overridden in a second document included at the bottom?

Just a hypothetical, but if it were implemented in a narrower way where `--set` worked but `--values` didn't, would that alleviate some of these concerns?

> I'm not sure / can't remember directly whether we support multiple yaml documents in a values.yaml, but that would be a worthy additional regardless

Without being an expert on helm's behavior, to me, it would seem a bit unexpected to me that multiple documents in a single `values.yaml` would be merged or combined sequentially against the same chart? And, in either event, even if that were a desirable feature, probably it would represent a lot of scope creep to support this, unless it were first implemented / tested separately, before (or after) this change.

Maybe also a dumb question, but a broader question... I guess the reason this can't have the same behavior as `helm template` (which does already have `--set` implemented) can't be used for `helm package` is because that use case doesn't require preserving the comments, which was the main issue with previous attempts at implementing this functionality? It seems to me like the best behavior would be for the two to behave as identically as possible.

gjenkins8: My explicit concern is that these will behave differently:

```
helm package  -f value-overrides.yaml --set foo=bar mychart/
helm install -g mychart.tar.gz
```
vs.
```
helm package mychart/
helm install -g mychart.tar.gz  -f value-overrides.yaml --set foo=bar
```

By using different logic to merge/updates values for package, there is a risk that a behavioral difference will occur.

scottrigby: added a discussion item to this week's helm dev meeting

scottrigby:
Follow up to https://github.com/helm/helm/pull/11460#discussion_r2054607168

The results of the April 24 Helm dev meeting are that supporting a separate YAML library for merging—with either different or attempting to emulate the same merge logic as we currently do with other Helm commands—is too much complexity to maintain in Helm, especially for something that can be accomplished by other existing in the ecosystem.

My understanding of the use case for this feature request is to modify specific values in YAML before `helm package` — main example is the tag of default container images installed by the chart—so that you can bump versions as an application developer using CI-only rather than making these changes in Git. Am I correct?

There are existing tools to change values in YAML such as `yq`, or the more recent YamlScript (written by the author of YAML), which you can use in CI to do this before the `helm package` command.

For now, closing as "won't fix" for Helm 4. However, if at any point the future we adopt a new YAML merge tool for all values coalescing that allows retaining YAML comments, this feature request could be reevaluated because at that point it may not be as much added complexity to maintain.

pkg/action/package.go
scottrigby: See https://github.com/helm/helm/pull/11460#issuecomment-2836832666

PR #11480, opened 2022-10-27
redirect registry client output to stderr

### LINKED ISSUE #11385: Malformed JSON output

jouve:
**What this PR does / why we need it**: Fix #11385

before: `helm template oci://registry/charts/name --version 1.0.5 | head` yields on stdout :
```
Pulled: registry/charts/name:1.0.5
Digest: sha256:117bf9af073f070d41af119dc3bf31241b7817ba9ab9692e336e1cb39df1e354
---
# Source: xxx.yaml
apiVersion: v1
...
```

after: same output, but Pulled/Digest are on stderr
stderr:
```
Pulled: registry/charts/name:1.0.5
Digest: sha256:117bf9af073f070d41af119dc3bf31241b7817ba9ab9692e336e1cb39df1e354
```
stdout
```
---
# Source: xxx.yaml
apiVersion: v1
...
```

this impacts the output of *all* the commands interacting with the registry.

`make test`  ok locally

my reading of https://github.com/helm/helm/blob/main/CONTRIBUTING.md#semantic-versioning says it is backward compatible :)

**If applicable**:

### DIFF
--- a/cmd/helm/root.go
+++ b/cmd/helm/root.go
@@ -155,7 +155,7 @@ func newRootCmd(actionConfig *action.Configuration, out io.Writer, args []string
 	registryClient, err := registry.NewClient(
 		registry.ClientOptDebug(settings.Debug),
 		registry.ClientOptEnableCache(true),
-		registry.ClientOptWriter(out),
+		registry.ClientOptWriter(os.Stderr),
 		registry.ClientOptCredentialsFile(settings.RegistryConfig),
 	)
 	if err != nil {


sabre1041 (approved):
LGTM

dirien:
What is the reason behind this change?

mot256:
> What is the reason behind this change?

Yes I must ask too?! This breaks allot of scripts that happy paths to not output to stderr

dogukanarkan:
@jouve Do you have any explanation for the questions above? This is the breaking changes for mostly operations/processes, and it doesn't make any sense. I'm trying to parse helm command's errors, but I can't because of this.

OneCricketeer:
From what I can tell, this is meant to address the following 

```
error validating "./to-apply.yaml": error validating data: [apiVersion not set, kind not set]; if you choose to ignore these errors, turn validation off with --validate=false
```

After doing 

```sh
helm template ... --version ... --set ... > to-apply.yaml
kubectl apply -f ./to-apply.yaml
```

> trying to parse helm command's errors

I don't see why you cannot do 

```sh
helm ... 2>errors.log
grep error errors.log
```

PR #11487, opened 2022-10-29
fix template --output-dir issue

### LINKED ISSUE #11481: Missing RBAC resources when templating using --output-dir option
RBAC resources are not rendered in a very specific case when they reside in the same template file that another resource with hook annotations and we use `--output-dir` option.

Data to reproduce:

**templates/dummy.yaml**
```
apiVersion: v1
kind: ServiceAccount
metadata:
  name: dummy-sa
---
kind: Role
metadata:
  name: dummy-role
rules:
- apiGroups:
  - ""
  resources:
  - secrets
  verbs:
  - create
  - get
---
kind: RoleBinding
metadata:
  name: dummy-rb
roleRef:
  kind: Role
  name: dummy-role
subjects:
- kind: ServiceAccount
  name: dummy-sa
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: dummy-cm
  annotations:
    "helm.sh/hook": post-install
    "helm.sh/hook-weight": "-10"
    "helm.sh/hook-delete-policy": hook-succeeded
data:
  dummy_data: dummy_data
```

Run helm template command:
```
helm template helm-dummy-chart --output-dir ./tmp
```

Rendered file:
```
---
# Source: helm/templates/cm.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: dummy-cm
  annotations:
    "helm.sh/hook": post-install
    "helm.sh/hook-weight": "-10"
    "helm.sh/hook-delete-policy": hook-succeeded
data:
  dummy_data: dummy_data
```

If I remove the `"helm.sh/hook": post-install` from annotations, the file is rendered correctly and all RBAC resources are present.

Output of `helm version`:
```
➜ helm version
version.BuildInfo{Version:"v3.10.1", GitCommit:"9f88ccb6aee40b9a0535fcc7efea6055e1ef72c9", GitTreeState:"clean", GoVersion:"go1.19.2"}
```

yxxhero:

fix: https://github.com/helm/helm/issues/11481

**If applicable**:

### DIFF
--- a/cmd/helm/template.go
+++ b/cmd/helm/template.go
@@ -106,11 +106,15 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 							if client.UseReleaseName {
 								newDir = filepath.Join(client.OutputDir, client.ReleaseName)
 							}
+							_, err := os.Stat(filepath.Join(newDir, m.Path))
+							if err == nil {
+								fileWritten[m.Path] = true
+							}
+
 							err = writeToFile(newDir, m.Path, m.Manifest, fileWritten[m.Path])
 							if err != nil {
 								return err
 							}
-							fileWritten[m.Path] = true
 						}
 
 					}


cmd/helm/template.go
```diff
@@ -106,11 +106,17 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 							if client.UseReleaseName {
 								newDir = filepath.Join(client.OutputDir, client.ReleaseName)
 							}
+							// fix https://github.com/helm/helm/issues/11481
+							// more info https://github.com/helm/helm/blob/7ed9d16dc764a5b94b378a7e217865efaa0d9ac8/cmd/helm/template.go#L200-L204
+							_, err := os.Stat(strings.Join([]string{newDir, m.Path}, string(filepath.Separator)))
```
mattfarina, line 111: Why is `strings.Join` being used instead of `filepath.Join`? This second form is used just a few lines above. It has some filepath checking and handles the correct Separator.

cmd/helm/template.go
```diff
@@ -106,11 +106,17 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 							if client.UseReleaseName {
 								newDir = filepath.Join(client.OutputDir, client.ReleaseName)
 							}
+							// fix https://github.com/helm/helm/issues/11481
+							// more info https://github.com/helm/helm/blob/7ed9d16dc764a5b94b378a7e217865efaa0d9ac8/cmd/helm/template.go#L200-L204
```
mattfarina, line 110: The code comment isn't necessary. It confused me, at first.

mattfarina (changes requested):
This bug makes sense and I understand the fix here. I just have one question about the choice of code to accomplish it. I didn't understand why this method was used to solve the problem when we typically use another one.

yxxhero:
@mattfarina done

mattfarina (approved):

mattfarina:
Merging as this is XS and only needs 1 reviewer.

PR #11490, opened 2022-10-31
Add support for CSVs in template --api-versions arg

### LINKED ISSUE #11485: helm template --api-versions does not support comma-separated values
## TL;DR

Hello! I noticed that the `--api-versions` flag for `helm template` uses a [StringArrayVar](https://pkg.go.dev/github.com/spf13/pflag#StringArrayVar) flag. `StringArrayVar` doesn't support separating given values by commas, treating any value with one or more commas as-is. Support for using CSVs for `--api-versions` in a backwards-compatible way can be added by switching `StringArrayVar` to `StringSliceVar`

```patch
diff --git cmd/helm/template.go cmd/helm/template.go
index f9c51542..ce2be55b 100644
--- cmd/helm/template.go
+++ cmd/helm/template.go
@@ -181,7 +181,7 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
        f.BoolVar(&skipTests, "skip-tests", false, "skip tests from templated output")
        f.BoolVar(&client.IsUpgrade, "is-upgrade", false, "set .Release.IsUpgrade instead of .Release.IsInstall")
        f.StringVar(&kubeVersion, "kube-version", "", "Kubernetes version used for Capabilities.KubeVersion")
-       f.StringArrayVarP(&extraAPIs, "api-versions", "a", []string{}, "Kubernetes api versions used for Capabilities.APIVersions")
+       f.StringSliceVarP(&extraAPIs, "api-versions", "a", []string{}, "Kubernetes api versions used for Capabilities.APIVersions")
        f.BoolVar(&client.UseReleaseName, "release-name", false, "use release name in the output-dir path.")
        bindPostRenderFlag(cmd, &client.PostRenderer)
```

## Explanation

This feels like a bug because if a user accidentally passes their extra API versions separated by commas to a single `--api-versions` flag, then all of their individual API versions that they would like to add to `.Capabilities.APIVersions` for chart verification will instead be treated as a single API version. Since `.Capabilities.APIVersions.Has` performs checks [based on equality](https://github.com/helm/helm/blob/main/pkg/chartutil/capabilities.go#L106), this could cause checks on the passed-in API versions to fail in an unexpected way.

Additionally, using `StringArrayVar` requires users who want to specify more than one API version to use an extended syntax like this:

```
helm template --api-versions <one> --api-versions <two> --api-versions <three>
```

Rather than a shortened form like this:

```
helm template --api-versions <one>,<two>,<three>
```

Which, depending on the need, can quickly increase the size of the needed `helm template`. 

These two "issues" are pretty minor, but by switching `StringArrayVar` to `StringSliceVar` this issue can be avoided in a backwards-compatible way, since `StringSliceVar` supports using same syntax as `StringArrayVar` while also allowing CSVs. (And from https://github.com/helm/helm/issues/10427 it looks like helm does string splitting in-house via [strvals](https://github.com/helm/helm/tree/main/pkg/strvals) for some flags that use `StringArrayVar`, however I think just changing `StringArrayVar` to `StringSliceVar` should suffice for this.)

## Example

Given the following templates:

```yaml
# validation.yaml
# Check for "test1" in API Versions
{{- if not (.Capabilities.APIVersions.Has "test1") }}
{{ fail "no test1" }}
{{- end }}
```
```yaml
# capabilities.yaml
# Print values in .Capabilities.APIVersions that start with "test"
# Truncates output so we don't have to look past all the default versions
capabilities:
  {{- range $a := .Capabilities.APIVersions }}
    {{- if hasPrefix "test" $a }}
  - {{ $a }}
    {{- end }}
  {{- end }}
```

Before the patch:

```
$ helm template . --api-versions test1 --api-versions test2,test3
---
# Source: apiversions-test/templates/capabilities.yaml
capabilities:
  - test1
  - test2,test3
$ helm template . --api-versions test1,test2,test3
Error: execution error at (apiversions-test/templates/validate.yaml:2:3): no test1

```

After the patch:

```
$ ../helm/helm template . --api-versions test1 --api-versions test2,test3
---
# Source: apiversions-test/templates/capabilities.yaml
capabilities:
  - test1
  - test2
  - test3
$ ../helm/helm template . --api-versions test1,test2,test3
---
# Source: apiversions-test/templates/capabilities.yaml
capabilities:
  - test1
  - test2
  - test3
```

## Versions

Output of `helm version`:
`version.BuildInfo{Version:"v3.10.1", GitCommit:"v3.10.1", GitTreeState:"", GoVersion:"go1.18.6"}`

Output of `kubectl version`:
```
clientVersion:
  buildDate: "1980-01-01T00:00:00Z"
  compiler: gc
  gitCommit: 434bfd82814af038ad94d62ebe59b133fcb50506
  gitTreeState: archive
  gitVersion: v1.25.3
  goVersion: go1.19.2
  major: "1"
  minor: "25"
  platform: linux/amd64
kustomizeVersion: v4.5.7
```

n/a

Thanks!

learnitall:

This commit changes the template sub-command argument `--api-versions` from `StringArrayVarP` to `StringSliceVarP`, to allow support for comma separated variables. This will allow users to use a syntax such as `--api-versions one,two,three` over `--api-versions one --api-versions two --api-versions three`.

Closes #11485

Signed-off-by: Ryan Drew <ryan.drew@isovalent.com>

Please see #11485 for more details.

### DIFF
--- a/cmd/helm/template.go
+++ b/cmd/helm/template.go
@@ -181,7 +181,7 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 	f.BoolVar(&skipTests, "skip-tests", false, "skip tests from templated output")
 	f.BoolVar(&client.IsUpgrade, "is-upgrade", false, "set .Release.IsUpgrade instead of .Release.IsInstall")
 	f.StringVar(&kubeVersion, "kube-version", "", "Kubernetes version used for Capabilities.KubeVersion")
-	f.StringArrayVarP(&extraAPIs, "api-versions", "a", []string{}, "Kubernetes api versions used for Capabilities.APIVersions")
+	f.StringSliceVarP(&extraAPIs, "api-versions", "a", []string{}, "Kubernetes api versions used for Capabilities.APIVersions")
 	f.BoolVar(&client.UseReleaseName, "release-name", false, "use release name in the output-dir path.")
 	bindPostRenderFlag(cmd, &client.PostRenderer)
 


yxxhero (approved):
LGTM

phil9909 (approved):

learnitall:
Hey! Just checking in. Is there anything left to do for this before it can be merged?

mattfarina:
This is backwards compatible. From the flag handling docs...

```
// For example:
//   --ss="v1,v2" --ss="v3"
// will result in
//   []string{"v1", "v2", "v3"}
```

And, a comma is not an allowed character in a group/version that would impact the input.

mattfarina (approved):

mattfarina:
XS PR which can be merged on one maintainer approval. So, merging.

PR #11494, opened 2022-11-02
Add a flag var to check git is installed or not

### LINKED ISSUE #11492: plugin install command needs git
It looks that `helm plugin install` command needs git to be installed.
When I execute command in machine where git is not installed, i found error message like link below
https://github.com/snapcrafters/helm/issues/15

because of this issue, `snapcratfers` developers updated their required packages configuration.
https://github.com/snapcrafters/helm/pull/51/files#diff-56759910381a014fecfd7556dd72ddd68c747d922a5b7df2044b9ce7c552f5f5R27

So, how about changing `get-helm-3` script like below to notify users? 
here is an example I think.
https://github.com/Wonong/helm/commit/9e21460515d836cb7169acb3e7615800778fd7e1

Wonong:

This changes update `scripts/get-helm-3` to notify users to pre-install `git` CLI before installing HELM

close #11492 

**If applicable**:

### DIFF
--- a/scripts/get-helm-3
+++ b/scripts/get-helm-3
@@ -29,6 +29,7 @@ HAS_CURL="$(type "curl" &> /dev/null && echo true || echo false)"
 HAS_WGET="$(type "wget" &> /dev/null && echo true || echo false)"
 HAS_OPENSSL="$(type "openssl" &> /dev/null && echo true || echo false)"
 HAS_GPG="$(type "gpg" &> /dev/null && echo true || echo false)"
+HAS_GIT="$(type "git" &> /dev/null && echo true || echo false)"
 
 # initArch discovers the architecture for this system.
 initArch() {
@@ -96,6 +97,11 @@ verifySupported() {
       echo "Please set VERIFY_SIGNATURES=false or verify the signatures manually."
       exit 1
     fi
+  fi
+
+  if [ "${HAS_GIT}" != "true" ]; then
+    echo "Git is required for plugin install"
+    exit 1
   fi
 }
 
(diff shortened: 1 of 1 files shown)


yxxhero:
@Wonong git is a common tool. check git is valuable.

phil9909 (changes requested):
I think this is of questionable value.

It is clear why the installation fails if `wget` and `curl` are missing, they are required, but `git`?
This is only needed when installing plugins (not even for all plugins, only for "vcs" plugins) so why prevent the installation?

I'd suggest to either turn this into a warning, e.g.:

```bash
  if [ "${HAS_GIT}" != "true" ]; then
    echo "[WARNING] Could not find git. It is required for plugin installation."
  fi
```

or at least give users a way to ignore the error, e.g.:

```bash
: ${VERIFY_GIT:="true"}

# ...

  if [ "${VERIFY_GIT}" == "true" ] && [ "${HAS_GIT}" != "true" ]; then
    echo "Git is required for plugin install. Set VERIFY_GIT=false to ignore this"
    exit 1
  fi
```

Wonong:
@phil9909 thank your comment :)
I think your first suggestion is good. It is enough to print warning log.
I will change codes to apply your opinion.

phil9909 (approved):

jadahbytes (approved):
Can confirm that the script prints the warning as expected.
(P.S. this is my 1st PR review T_T)

scottrigby (approved):
Straightforward user experience improvement 👍 Thanks @Wonong 
Thanks @phil9909 @yxxhero and @jadahbytes for reviews 🙌 

PS @jadahbytes welcome!

hickeyma (approved):

PR #11497, opened 2022-11-03
Fix namespace mismatch during installation

openqt:
Using the SDK and pass namespace to installation does not take effect, helm show the ns is different to the real deployed ns.
namespace shoud be passed to kube.Client anyway.

**If applicable**:

### DIFF
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -366,6 +366,7 @@ func (cfg *Configuration) recordRelease(r *release.Release) {
 func (cfg *Configuration) Init(getter genericclioptions.RESTClientGetter, namespace, helmDriver string, log DebugLog) error {
 	kc := kube.New(getter)
 	kc.Log = log
+	kc.Namespace = namespace
 
 	lazyClient := &lazyClient{
 		namespace: namespace,


pkg/action/action.go
```diff
@@ -366,6 +366,7 @@ func (cfg *Configuration) recordRelease(r *release.Release) {
 func (cfg *Configuration) Init(getter genericclioptions.RESTClientGetter, namespace, helmDriver string, log DebugLog) error {
 	kc := kube.New(getter)
 	kc.Log = log
+	kc.Namespace = namespace
```
mattfarina, line 369: This code is a bit confusing and not ideal. I understand the desire to add the namespace here.

Without the `kc.Namespace` addition, the namespace is calculated inside the `kube.Client` here:

https://github.com/helm/helm/blob/b4b629c3c2a0e122d9332f8c436ddf9cf2d06968/pkg/kube/client.go#L315-L323

The factory that's used is setup here:

https://github.com/helm/helm/blob/b4b629c3c2a0e122d9332f8c436ddf9cf2d06968/pkg/kube/client.go#L97-L100

And all of that is from the [`cli.EnvSettings`](https://github.com/helm/helm/blob/b4b629c3c2a0e122d9332f8c436ddf9cf2d06968/pkg/cli/environment.go#L48).

It's a hard code path to follow.

All of this is because there are two (at least) namespaces in play. First, there is the namespace where the Helm Secret/ConfigMap, etc is stored. Second, there is the namespace the resources are installed into. Those can technically be separate.

In the original code the `getter` should have the namespace set on it. When `kc` communicates with the Kubernetes API it will use the namespace figured out from `getter` if it's not set by `kc.Namespace`.

In Helm this is figured out from the `settings` object...

https://github.com/helm/helm/blob/b4b629c3c2a0e122d9332f8c436ddf9cf2d06968/cmd/helm/helm.go#L75-L77

`settings` has a `SetNamespace` on it. So you could so something like...

```go
settings.SetNamespace("foo")
if err := actionConfig.Init(settings.RESTClientGetter(), settings.Namespace(), helmDriver, debug); err != nil {
...
```

This would setup the KubeClient to use the namespace you want to set programatically.

None of the current code is idea and I have to reorient myself to it each time. But, I think it needs a refactor rather than this small change. This small change would need some good QA testing before we could approve it. I can think of some use cases where it would break some SDK users and that's my concern.

mattfarina (changes requested):
You touched on a piece of difficult to reason code. I think this section of code would need a new initialization function and to be reworked to do this. While your change might work for your use case and for Helm itself, there are others doing work using the Helm SDK that might break from it.

If you want to pursue such a change, I'm happy to chat about what would need to happen.

Otherwise, I shared a snipped that you can use to easily set the Namespace with no change at all to Helm.

joejulian:
Closing this as abandoned. If you can come back and address the change request, please reopen the PR.

Issue #11272, opened 2022-08-25
How to create and install plugin tgz


mblaschke:
Is the plugin installation with tgz still supported/working? It's mentioned in documentation at https://helm.sh/docs/topics/plugins/#installing-a-plugin

But when trying to download a .tgz from github releases this happens:

```
$ helm plugin install https://github.com/xxxx/xxxx/releases/download/xx-xx-xx/helm-plugin.tgz
Error: Unable to get repository: Cloning into '/Users/mblaschke/Library/Caches/helm/plugins/https-github.com-xxxx-helm-xxxxx-releases-download-xx-xx-xx-helm-plugin.tgz'...
remote: Not Found
fatal: repository 'https://github.com/xxxx/xxxx/releases/download/xx-xx-xx/helm-plugin.tgz/' not found
: exit status 128
```

when fetching the tgz and trying it locally it also fails:

```
$ helm --debug plugin install release-assets/helm-plugin.tgz
Error: plugin metadata (plugin.yaml) missing
helm.go:84: [debug] plugin metadata (plugin.yaml) missing
helm.sh/helm/v3/pkg/plugin/installer.init
	helm.sh/helm/v3/pkg/plugin/installer/installer.go:32
runtime.doInit
	runtime/proc.go:6222
runtime.doInit
	runtime/proc.go:6199
runtime.main
	runtime/proc.go:233
runtime.goexit
	runtime/asm_arm64.s:1263


$ tar tzvf release-assets/helm-plugin.tgz | grep plugin
-rw-r--r--  0 mblaschke staff      693 Aug 25 15:39 plugin.yaml


$ file release-assets/helm-plugin.tgz
release-assets/helm-plugin.tgz: gzip compressed data, last modified: Thu Aug 25 13:39:09 2022, from Unix, original size modulo 2^32 73625600
```

might be also related to #10402

Output of `helm version`:
version.BuildInfo{Version:"v3.9.2", GitCommit:"1addefbfe665c350f4daf868a9adc5600cc064fd", GitTreeState:"clean", GoVersion:"go1.18.4"

Output of `kubectl version`:
n/A

Cloud Provider/Platform (AKS, GKE, Minikube etc.): 
n/A

phil9909:
Hi @mblaschke,

- `helm plugin install https://github.com/xxxx/xxxx/releases/download/xx-xx-xx/helm-plugin.tgz` seems to be a problem in `isRemoteHTTPArchive` not following redirects.
[...]

PR #11508, opened 2022-11-04
improve error message on plugin install

phil9909:

Plugins are commonly downloadable as tar files from github releases.
If a users downloads the archive and tries to install it without extracting helm fails with the following error `Error: plugin metadata (plugin.yaml) missing`. This is not very helpful, as there is a `plugin.yaml` in the archive, helm simply expects a folder instead of an archive. Which led to the issue https://github.com/helm/helm/issues/11272

This PR changes the error message to detect this case and print a (hopefully) more helpful error message: `Error: expected plugin to be a folder`

Additional, helm could just extract the archive. Most of the code for this is already there (plugins installed via http support tar files already). I can contribute this, if you think this would be helpful.

**If applicable**:

### DIFF
--- a/pkg/plugin/installer/local_installer.go
+++ b/pkg/plugin/installer/local_installer.go
@@ -22,6 +22,9 @@ import (
 	"github.com/pkg/errors"
 )
 
+// ErrPluginNotAFolder indicates that the plugin path is not a folder.
+var ErrPluginNotAFolder = errors.New("expected plugin to be a folder")
+
 // LocalInstaller installs plugins from the filesystem.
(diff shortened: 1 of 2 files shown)


yxxhero:
@phil9909 please fix the ci issue. Thanks very much.

phil9909:
@yxxhero done

mattfarina (approved):
Note, this needs approval from a second Helm maintainer

sabre1041 (approved):
LGTM

hickeyma (approved):

PR #11510, opened 2022-11-04
Add skip-schema-validation flag

### LINKED ISSUE #10398: Proposal: disable values schema validation for charts including a values.schema.json file
Even when a chart designer/maintainer has decided to include a ``values.schema.json`` file in the Helm chart he/she releases it could be that the user of that chart (using it directly or as a dependency in an integration/umbrella Helm chart) cannot assure to fulfill the defined schema for the deployment specific values passed to that subchart (e.g. he/she passes a deployment specific values.yaml file to the ``helm install/upgrade`` command applied on that chart, where some "old" Helm parameter removed from the ``values.schema.json`` of a newer version of that (sub)chart still remains).

I think it would be useful to support a new option in ``helm install``, ``helm upgrade``, ``helm lint`` and ``helm template`` allowing to disable the values schema validation (in case a ``values.schema.json`` file is packaged in the target Helm chart or some listed subchart)

JvD-Ericsson:
Adds a "--skip-schema-validation" flag option for helm install, upgrade, lint, and template that skips validation of the values.schema.json file.

closes https://github.com/helm/helm/issues/10398

This is my first time contributing, so please let me know if there is something I am doing wrong : )

**If applicable**:

Signed-off-by: Jeff van Dam <jeff.van.dam@est.tech>


thisisrachelramos (approved):

thisisrachelramos:
This is valuable for scenarios where one creates a chart and uses other mechanisms to validate the values.yaml file but wants to ensure their chart still produces valid YAML. 

Since a lot of templates use Go template syntax, YAML validation is valuable.

lucacome:
any chance this will get in `v3.12.0`? 🙏

xrow:
Why is this open so long? Looks like ready and complete. Ty @JvD-Ericsson

sumo-drosiek:
cc: @hickeyma

Cerebus:
Desperately need this, I have offline environments to support and forking charts just to remove validation is a royal PITA.

jouve:
@JvD-Ericsson can you resolve the conflicts, maybe it will helm advance this PR ?

JvD-Ericsson:
Resolved now. Maybe @mattfarina or @joejulian can review now

pkg/action/__debug_bin
jouve, line 1: I think you added this file by mistake

pkg/lint/rules/template.go
```diff
@@ -45,7 +45,7 @@ var (
 )
 
 // Templates lints the templates in the Linter.
-func Templates(linter *support.Linter, values map[string]interface{}, namespace string, strict bool) {
+func Templates(linter *support.Linter, values map[string]interface{}, namespace string, strict bool, skipSchemaValidation bool) {
```
jouve, line 48: this probably needs a new method like (ToRenderValuesSkipSchemaValidation) to keep api backward compatibilty ?

anessi:
Tried a new approach in https://github.com/helm/helm/pull/12743

pkg/lint/rules/template.go
scottrigby: Correct 👍 we can not change the signature of public functions in Helm 3

vijays-rafay:
Any updated on this. I verified latest helm binary however couldn't find the flag --skip-schema-validation

Issue #11518, opened 2022-11-07
Import-values do not override default values


pbusko:
Following the import-values example described [on the official website](https://helm.sh/docs/topics/charts/#importing-child-values-via-dependencies), we can't get the result shown on the page. If the parent chart has default values and the same value gets imported from the child chart, it's not overridden. Attached a packaged chart with the example from the documentation.
[test-import-0.1.0.tgz](https://github.com/helm/helm/files/9952674/test-import-0.1.0.tgz)

Expected output:

```console
$ helm template test test-import-0.1.0.tgz 
---
# Source: test-import/templates/test.yaml
myimports:
  mybool: true
  myint: 999
  mystring: helm rocks!
subchart:
  default:
    data:
      mybool: true
      myint: 999
  global: {}
```

Actual output (`myimports.mybool` and `myimports.myint` are kept as defaults):

```console
$ helm template test test-import-0.1.0.tgz 
---
# Source: test-import/templates/test.yaml
myimports:
  mybool: false
  myint: 0
  mystring: helm rocks!
subchart:
  default:
    data:
      mybool: true
      myint: 999
  global: {}
```

Output of `helm version`:
```console
$ helm version
version.BuildInfo{Version:"v3.10.1", GitCommit:"9f88ccb6aee40b9a0535fcc7efea6055e1ef72c9", GitTreeState:"clean", GoVersion:"go1.19.2"}
```
Output of `kubectl version`:
```console
$ kubectl version
WARNING: This version information is deprecated and will be replaced with the output from kubectl version --short.  Use --output=yaml|json to get the full version.
Client Version: version.Info{Major:"1", Minor:"25", GitVersion:"v1.25.3", GitCommit:"434bfd82814af038ad94d62ebe59b133fcb50506", GitTreeState:"clean", BuildDate:"2022-10-12T10:47:25Z", GoVersion:"go1.19.2", Compiler:"gc", Platform:"darwin/amd64"}
Kustomize Version: v4.5.7
Server Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.12", GitCommit:"c6939792865ef0f70f92006081690d77411c8ed5", GitTreeState:"clean", BuildDate:"2022-09-21T12:13:07Z", GoVersion:"go1.17.13", Compiler:"gc", Platform:"linux/amd64"}
WARNING: version difference between client (1.25) and server (1.23) exceeds the supported minor version skew of +/-1
```

joejulian:
[...]

PR #11527, opened 2022-11-10
Import values override defaults

pbusko:
Address the issue https://github.com/helm/helm/issues/11518

**If applicable**:

### DIFF
--- a/pkg/chartutil/dependencies.go
+++ b/pkg/chartutil/dependencies.go
@@ -248,7 +248,7 @@ func processImportValues(c *chart.Chart) error {
 					continue
 				}
 				// create value map from child to be merged into parent
-				b = CoalesceTables(cvals, pathToMap(parent, vv.AsMap()))
+				b = CoalesceTables(b, pathToMap(parent, vv.AsMap()))
 			case string:
 				child := "exports." + iv
 				outiv = append(outiv, map[string]string{
@@ -268,7 +268,7 @@ func processImportValues(c *chart.Chart) error {
 	}
 
 	// set the new values
-	c.Values = CoalesceTables(cvals, b)
+	c.Values = CoalesceTables(b, cvals)
 
 	return nil
 }
--- a/pkg/chartutil/dependencies_test.go
+++ b/pkg/chartutil/dependencies_test.go
@@ -181,10 +181,10 @@ func TestProcessDependencyImportValues(t *testing.T) {
 	e["imported-chartA-B.SPextra5"] = "k8s"
 	e["imported-chartA-B.SC1extra5"] = "tiller"
 
-	e["overridden-chart1.SC1bool"] = "false"
-	e["overridden-chart1.SC1float"] = "3.141592"
-	e["overridden-chart1.SC1int"] = "99"
-	e["overridden-chart1.SC1string"] = "pollywog"
+	e["overridden-chart1.SC1bool"] = "true"
+	e["overridden-chart1.SC1float"] = "3.14"
+	e["overridden-chart1.SC1int"] = "100"
+	e["overridden-chart1.SC1string"] = "dollywood"
 	e["overridden-chart1.SPextra2"] = "42"
 
 	e["overridden-chartA.SCAbool"] = "true"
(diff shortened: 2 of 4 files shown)


technosophos:
I think this is the PR that introduced the current behavior: https://github.com/helm/helm/commit/7ea4d8c7c460eae18bf74c682832acfb009445ce

digitalstudium:
I prefer current behavior. The use case is when I import default values from a library chart and then redefine some of them if needed in a parent chart.
Example which works for me today:
library chart `values.yaml`:
```
defaults:
  image:
    registry: quay.io
    name: ""
```
`values.yaml` of a chart which imports library chart:
```
image:
  name: nginx
```

Resulting `values.yaml`:
```
  image:
    registry: quay.io
    name: nginx
 ```
 
So I prefer if you change documentation, not the code. Because I want my library chart to have default values and I want to have a method to redefine some of them in parent's `values.yaml`. This is very convenient way.
Such way library chart's `values.yaml` is like a class which I can inherit and then extend or modify.

Side note: the way I import library chart is such:
```
dependencies:
  - name: base
    repository: file://../../libraries/base
    version: "*"
    import-values:
      - child: defaults
        parent: .
```

digitalstudium:
And BTW, if you merge this PR it will be breaking change. Some people already wrote charts which work with current behavior and which will not work after merging this code.

hedayat:
> And BTW, if you merge this PR it will be breaking change. Some people already wrote charts which work with current behavior and which will not work after merging this code.

And there are some people who already wrote charts which work with the old behavior, and are not currently working with the new one. And it doesn't follow the documented behavior. 

Having a very inconsistent behavior based on different versions of helm doesn't look good. Anyway, #11162 seems to be a good solution.

scottrigby (changes requested):
We've been attempting to clarify how this should work, so that both documentation and tests for code follow. Here is the google doc WIP for documenting this: https://docs.google.com/document/d/1ZjmIc5VYzNbaJ-SQ6nT02Oo1t9ctlvomufP8ulR6q70/edit

As I understand it, @digitalstudium is correct that the current behavior is as intended. Documentation improvement in progress here: [What is the order of precedence logic?](https://docs.google.com/document/d/1ZjmIc5VYzNbaJ-SQ6nT02Oo1t9ctlvomufP8ulR6q70/edit#heading=h.brer0zqgu3qn):

> Logical specification/ordering (from [12480](https://github.com/helm/helm/pull/12480)) should be:
> 1. User specified values (e.g CLI --set, -f)
> 2. Parent chart default values
> 3. Imported values get merged from child chart(s) into parent chart’s value
> 4. Parent chart’s values get propagated to sub-chart(s), and merged with sub-chart values

Based on this, I believe the changes should be an update to the docs.

hedayat:
@scottrigby The doc is still WIP, and inconsistent (also mentioned in the comments). And the priority doesn't cover all the use cases mentioned there. Unless decisions has been made somewhere else, I think it's still not ready to be concluded from!

intetinte:
The current ordering is:
1. User specified values (e.g CLI --set, -f)
2. Parent chart default values
3. Parent chart’s values get propagated to sub-chart(s), and merged with sub-chart values
4. Imported values get merged from child chart(s) into parent chart’s value

With this pull request implemented it will instead be:
1. User specified values (e.g CLI --set, -f)
2. Imported values get merged from child chart(s) into parent chart’s value
3. Parent chart default values
4. Parent chart’s values get propagated to sub-chart(s), and merged with sub-chart values

And this doc https://docs.google.com/document/d/1ZjmIc5VYzNbaJ-SQ6nT02Oo1t9ctlvomufP8ulR6q70/edit#heading=h.brer0zqgu3qn explains another ordering (aligned in this PR https://github.com/helm/helm/pull/12590) :
1. User specified values (e.g CLI --set, -f)
2. Parent chart default values
3. Imported values get merged from child chart(s) into parent chart’s value
4. Parent chart’s values get propagated to sub-chart(s), and merged with sub-chart values

Someone need to state what the ordering should be an align to that.

PR #11591, opened 2022-12-01
Use wrapped error so that ErrNoObjectsVisited can be compared

ruanxin:
In the current client, when encounter  `ErrNoObjectsVisited`, it will be converted to a new error, I understand the purpose of the original thought is trying to provide an error message that is more readable. But the problem with this approach is, it's lost the possibility to compare the error `ErrNoObjectsVisited` directly when needed.

In this PR, it just use `fmt.Errorf` with wrapped error `%w` verb so that this error can be used by `errors.Is` and `errors.As`.

**If applicable**:

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -325,8 +325,8 @@ func (c *Client) Delete(resources ResourceList) (*Result, []error) {
 		return nil
 	})
 	if err != nil {
-		if err == ErrNoObjectsVisited {
-			err = fmt.Errorf("object not found, skipping delete: %w", ErrNoObjectsVisited)
+		if errors.Is(err, ErrNoObjectsVisited) {
+			err = fmt.Errorf("object not found, skipping delete: %w", err)
 		}
 		errs = append(errs, err)
 	}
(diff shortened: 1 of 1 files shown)


pkg/kube/client.go
```diff
@@ -325,10 +325,8 @@ func (c *Client) Delete(resources ResourceList) (*Result, []error) {
 		return nil
 	})
 	if err != nil {
-		// Rewrite the message from "no objects visited" if that is what we got
-		// back
 		if err == ErrNoObjectsVisited {
-			err = errors.New("object not found, skipping delete")
+			err = fmt.Errorf("object not found, skipping delete: %w", ErrNoObjectsVisited)
```
joejulian, line 329: Shouldn't this be:

```suggestion
			err = fmt.Errorf("object not found, skipping delete: %w", err)
```
in case err has any additional context?

ruanxin: Hi @joejulian ,
Thanks for the comment! This ErrNoObjectsVisited is customized one, and at the moment the usage for this error has no additional context appended, but I agreed with you, by directly return this err itself make it more generic. Please check my update.

joejulian (approved):

Issue #11352, opened 2022-09-16
[Umbrella Issue] Stabilize OCI support


makkes:
This is an umbrella issue to track any work on stabilizing OCI support in Helm. There are multiple PRs in the works that are related and some even overlap. This issue is there to help us coordinate and align the work better.

- [ ] #11129
- [ ] #10408
- [ ] #10798
- [ ] #9564

@scottrigby @sabre1041

joejulian:
Should https://github.com/helm/helm/issues/11286 be part of this?

makkes:
> Should https://github.com/helm/helm/issues/11286 be part of this?

Added it in. Thanks @joejulian!

sabre1041:
Lets also add #10452 which I reopened earlier today as well

joejulian:
https://github.com/helm/helm/issues/10881

scottrigby:
Also 
- #11363

joejulian:
Also [Support pull/install helm chart from OCI repository over index.yaml](https://github.com/helm/helm/issues/11361) #11361

joejulian:
I'm starting to wonder if we should just have an OCI label. :)

makkes:
> I'm starting to wonder if we should just have an OCI label. :)

For the longer term that sounds totally reasonable. Our goal with this umbrella issue for now is to coordinate the various PRs that already exist and even partly overlap in terms of the functionality they provide.

scottrigby:
> > I'm starting to wonder if we should just have an OCI label. :)
> 
> For the longer term that sounds totally reasonable. Our goal with this umbrella issue for now is to coordinate the various PRs that already exist and even partly overlap in terms of the functionality they provide.

I made and applied the label. Also thinking umbrella issues seem like a good idea.

![image](https://user-images.githubusercontent.com/407675/192400240-a3004cea-084c-4ec7-bde6-5b27443584bf.png)

I'm thinking we could set milestones on individual issues if we want to organize a strategy, but only if it helps us instead of making more busywork.

This umbrella issue also feels pretty useful for now. Maybe we could make smaller umbrella issues if this one becomes too broad. For example, a number of these issues/PRs relate to OCI authentication. Those in particular I think are useful to group to talk about all together, because some handle the same issues in different ways and it would be good to have a place to discuss it and give guidance.

dvcanton:
@makkes Any update on thoses issues?

makkes:
> @makkes Any update on those issues?

[...]

PR #11599, opened 2022-12-04
Added support for insecure OCI registries

sabre1041:
Signed-off-by: Andrew Block <andy.block@gmail.com>

Supports connectivity for insecure OCI registries. 

Relates to #11352

### DIFF
--- a/pkg/getter/httpgetter.go
+++ b/pkg/getter/httpgetter.go
@@ -123,8 +123,8 @@ func (g *HTTPGetter) httpClient() (*http.Client, error) {
 		}
 	})
 
-	if (g.opts.certFile != "" && g.opts.keyFile != "") || g.opts.caFile != "" {
-		tlsConf, err := tlsutil.NewClientTLS(g.opts.certFile, g.opts.keyFile, g.opts.caFile)
+	if (g.opts.certFile != "" && g.opts.keyFile != "") || g.opts.caFile != "" || g.opts.insecureSkipVerifyTLS {
+		tlsConf, err := tlsutil.NewClientTLS(g.opts.certFile, g.opts.keyFile, g.opts.caFile, g.opts.insecureSkipVerifyTLS)
 		if err != nil {
 			return nil, errors.Wrap(err, "can't create TLS config for client")
 		}
--- a/pkg/getter/httpgetter.go
+++ b/pkg/getter/httpgetter.go
@@ -123,7 +123,7 @@ func (g *HTTPGetter) httpClient() (*http.Client, error) {
 		}
 	})
 
-	if (g.opts.certFile != "" && g.opts.keyFile != "") || g.opts.caFile != "" || g.opts.insecureSkipVerifyTLS {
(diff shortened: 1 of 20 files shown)


pkg/getter/httpgetter.go
```diff
@@ -123,8 +123,8 @@ func (g *HTTPGetter) httpClient() (*http.Client, error) {
 		}
 	})
 
-	if (g.opts.certFile != "" && g.opts.keyFile != "") || g.opts.caFile != "" {
-		tlsConf, err := tlsutil.NewClientTLS(g.opts.certFile, g.opts.keyFile, g.opts.caFile)
+	if (g.opts.certFile != "" && g.opts.keyFile != "") || g.opts.caFile != "" || g.opts.insecureSkipVerifyTLS {
```
souleb, line 126: this should be another PR imo.

pkg/pusher/ocipusher_test.go
```diff
@@ -35,10 +35,12 @@ func TestNewOCIPusher(t *testing.T) {
 	cd := "../../testdata"
 	join := filepath.Join
 	ca, pub, priv := join(cd, "rootca.crt"), join(cd, "crt.pem"), join(cd, "key.pem")
+	insecureSkipTLSverify := false
 
 	// Test with options
 	p, err = NewOCIPusher(
 		WithTLSClientConfig(pub, priv, ca),
+		WithInsecureSkipTLSVerify(insecureSkipTLSverify),
```
souleb, line 43: can you add this in `ocigetter_test.go` as well?

sabre1041:
@souleb for the comment related to the `registry_login.go` file, did you mean for it to reference the `action` package instead of `cmd`?

souleb:
> ```go
> 		registry.LoginOptInsecure(insecure),
> ```

I mean that we should use the `insecureSkipTLSVerify` flag there too, and use the option `registry.LoginOptInsecure(insecure)` when login.

sabre1041:
@souleb got it. There is no actual functionality in the login process aside from the flag option, so I will work on integrating that in as part of this PR

souleb:
`pkg/getter/httpgetter.go` should be tested I think and have its own PR.

sabre1041:
@souleb is it just the enhancements to `pkg/getter/httpgetter.go` that should be in a separate PR? Just trying to determine where to split it up

souleb:
> @souleb is it just the enhancements to `pkg/getter/httpgetter.go` that should be in a separate PR? Just trying to determine where to split it up

Yes. To me that addresses `insecure` for `https` repositories.

souleb:
To me here we adresses OCI registries hosting helm artifacts. An `httpgetter` is used to access a helm repository (with a index file). So I would say it's out of scope.

sabre1041:
@souleb made the requested changes

souleb (approved):
LGTM!

keskad:
Tested it, seems to be working.

https://github.com/helm/helm/issues/11683#issuecomment-1378482533

PR #11616, opened 2022-12-08
docs: add docs for cli/values.Options

zalsader:
Add documentation for cli/values.Options

Without reading the code, it is hard to know which flag corresponds to which field in `values.Options`. This PR adds the simplest possible documentation.

**If applicable**:

### DIFF
--- a/pkg/cli/values/options.go
+++ b/pkg/cli/values/options.go
@@ -29,16 +29,17 @@ import (
 	"helm.sh/helm/v3/pkg/strvals"
 )
 
+// Options captures the different ways to specify values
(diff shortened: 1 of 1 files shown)


mattfarina (approved):
Due to separation of concerns, I normally wouldn't want a change like this here. This is in a package that's part of the SDK and the flags are part of the client application. But, this file is already commented with the flags. So, it should either be right and up to date or removed. So, approving.

Note, this needs to be approved by a second maintainer before it can be merged.

technosophos (approved):

PR #11628, opened 2022-12-11
change linting error messages for null values in arrays

### LINKED ISSUE #11627: Linting error on null in dependencies is misleading
When doing `helm lint` on a chart with the following dependencies: 

```
dependencies:
  - name: foo
    version: 1.2.3
  - null
```

it fails with 
```
==> Linting .
[ERROR] templates/: validation: dependency cannot be an empty list
```

This is quite misleading, as a chart with an empty list (`dependencies: []`) does not fail linting.

This was caused by #11044 and I would suggest to simply change the error message to a more meaningful one. 

Output of `helm version`: 
```
version.BuildInfo{Version:"v3.10.2", GitCommit:"50f003e5ee8704ec937a756c646870227d7c8b58", GitTreeState:"clean", GoVersion:"go1.19.3"}
```

dastrobu:
Closes #11627

**If applicable**:

### DIFF
--- a/pkg/chart/metadata_test.go
+++ b/pkg/chart/metadata_test.go
@@ -82,7 +82,7 @@ func TestValidate(t *testing.T) {
 					nil,
 				},
 			},
-			ValidationError("dependency cannot be an empty list"),
+			ValidationError("dependencies must not contain empty or null nodes"),
 		},
 		{
 			&Metadata{
@@ -94,7 +94,7 @@ func TestValidate(t *testing.T) {
 					nil,
 				},
 			},
-			ValidationError("maintainer cannot be an empty list"),
+			ValidationError("maintainers must not contain empty or null nodes"),
 		},
 		{
 			&Metadata{APIVersion: "v2", Name: "test", Version: "1.2.3.4"},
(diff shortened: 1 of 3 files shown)


pkg/chart/metadata_test.go
```diff
@@ -5,7 +5,7 @@ Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-    http://www.apache.org/licenses/LICENSE-2.0
```
hickeyma, line 8: remove indentation

hickeyma (changes requested):
Looks good bar a small nit

pkg/chart/metadata_test.go
dastrobu: @hickeyma I fixed this for this PR. 

It is a bit strange, though, to have part of the indentation with spaces and part with tabs, which is why my IDE changed this automatically: 

image

I noticed that this seems to be something that is not consistent over the codebase. Not sure, if there is some discussion ongoing about this issue somewhere....

hickeyma (approved):

pkg/chart/metadata_test.go
hickeyma: Not sure if this was ever raised. If you want, you could raise a issue and some examples in it?

PR #11633, opened 2022-12-12
Reimplementation support for Insecure TLS for OCI branch

sabre1041:
Reimplementation of httpgetter conditional for Insecure TLS for OCI branch


souleb (approved):
LGTM!

PR #11645, opened 2022-12-15
Add easy way to override default container registry

jcpunk:

Folks may have local/alternate container registries they are using to avoid pulling across expensive links.  This adds an easy way for the default templates to specify an explicit container registry.

**If applicable**:

### DIFF
--- a/pkg/chartutil/create.go
+++ b/pkg/chartutil/create.go
@@ -110,6 +110,7 @@ const defaultValues = `# Default values for %s.
 replicaCount: 1
 
 image:
+  registry: ""
   repository: nginx
   pullPolicy: IfNotPresent
   # Overrides the image tag whose default is the chart appVersion.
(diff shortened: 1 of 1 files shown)


pkg/chartutil/create.go
```diff
@@ -110,6 +110,7 @@ const defaultValues = `# Default values for %s.
 replicaCount: 1
 
 image:
+  registry: ""
   repository: nginx
```
joejulian, line 114: Are there a lot of charts that do this? I've seen relatively few. Most just have you adjust the respository, ie:
```
repository: myproxy/nginx
```

I don't believe that this will produce a valid reference if the registry is empty:
```
$ docker pull /mariadb:10.7
invalid reference format
```

joejulian (changes requested):
This will produce an invalid image reference

pkg/chartutil/create.go
joejulian: If you want to check for explicit registry usage, you could always just check to see that the repository has at least two `/`. There is a more complex regex that I saw somewhere, too.

If you really wanted to have a separate key for the registry, you could check to see if its empty and not add the slash.

I think that's too complex for the default generated chart though. It's important to keep it simple.

jouve:
for reference, how it's implemented in [bitnami/common](https://github.com/bitnami/charts/blob/main/bitnami/common/templates/_images.tpl#L9)

``` 
image:
  registry: docker.io
  repository: bitnami/haproxy
  tag: 2.8.2-debian-11-r10
  digest: ""
```

PR #11649, opened 2022-12-16
Use common labels instead of selectorLabels for pod labels

tim-chaplin-dd:
Signed-off-by: Tim Chaplin <tim.chaplin@datadoghq.com>

By default, a newly generated Helm chart provides a `labels` macro and a `selectorLabels` macro. `selectorLabels` is currently used for the `selector.matchLabels` section of a Deployment, as well as `template.metadata.labels`. As a result, pods only get a subset of the common labels applied. I couldn't find any documented rationale for why you would only want to use `selectorLabels` for pods, so my guess is that it's a bug.

**If applicable**:

### DIFF
--- a/pkg/chartutil/create.go
+++ b/pkg/chartutil/create.go
@@ -310,7 +310,7 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       labels:
-        {{- include "<CHARTNAME>.selectorLabels" . | nindent 8 }}
+        {{- include "<CHARTNAME>.labels" . | nindent 8 }}
 	{{- with .Values.podLabels }}
         {{- toYaml . | nindent 8 }}
         {{- end }}


mattfarina:
This is an interesting one. The labels include that it's managed by Helm and more. Looking back at the history it has long been what we now use as the selector labels. This is a bit philosophical on what should be there.

Any thoughts @helm/charts-maintainers ?

joejulian:
I completely agree with this. The Deployment's selector:
```
  selector:
    matchLabels:
      {{- include "<CHARTNAME>.selectorLabels" . | nindent 6 }}
```
so the pod's labels should also come from selectorLabels in order for the Deployment to select them.

joejulian (approved):

joejulian:
@tim-chaplin-dd can you rebase this please?

tim-chaplin-dd:
> @tim-chaplin-dd can you rebase this please?

Done, sorry for the delay!

joejulian (approved):
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: release-name-foo
  labels:
    helm.sh/chart: foo-0.1.0
    app.kubernetes.io/name: foo
    app.kubernetes.io/instance: release-name
spec:
  replicas: 1
  selector:
    matchLabels:
      app.kubernetes.io/name: foo
      app.kubernetes.io/instance: release-name
  template:
    metadata:
      labels:
        helm.sh/chart: foo-0.1.0
        app.kubernetes.io/name: foo
        app.kubernetes.io/instance: release-name
```

This is good. If you had the full list of labels, when some of those dynamic labels change, updates would fail since selectors are immutable.

PR #11650, opened 2022-12-16
Use 3-way diff when producing JSON merge patches for unstructured and CRD resources

### LINKED ISSUE #11638: Modifications not reflecting on adopted IstioOperator resource
It's tricky to tell whether this is an issue specifically with Helm or perhaps with the IstioOperator CRD, however I figure this is a good place to start.  Here's a scenario that demonstrates a strange behavior that I'm seeing:

* Starting with a fresh minikube cluster
* Run some initial setup to install the Istio operator into the cluster with `istioctl operator init` and `kubectl create namespace istio-system`
* Deploy an IstioOperator CRD into the cluster with `kubectl apply -f ./pilot-raw.yaml`
```yaml
# pilot-raw.yaml
apiVersion: install.istio.io/v1alpha1
kind: IstioOperator
metadata:
  namespace: istio-system
  name: istio-operator
spec:
  profile: default
  components:
    ingressGateways:
      - name: istio-ingressgateway
        enabled: false
  values:
    global:
      proxy:
        resources:
          requests:
            cpu: 50m
```

The goal is "adopt" this resource into a new Helm chart for managing Istio.
* First we must patch the resource to include the annotations expected by Helm
```bash
patchBody='{
  "metadata": {
    "annotations": {
      "meta.helm.sh/release-name": "my-istio",
      "meta.helm.sh/release-namespace": "istio-system"
    },
    "labels": {
      "app.kubernetes.io/managed-by": "Helm"
    }
  }
}'
kubectl patch -n istio-system istiooperator "istio-operator" --type=merge -p "$patchBody"
```
* Now we can go ahead an install the Helm chart with `helm upgrade my-istio ./helm --install --namespace istio-system`
* Here the `./helm` directory is a chart which contains just a single template of the same IstioOperator CRD, but modified to add some new values under `.spec.values.global`
```yaml
# pilot-helm.yaml
apiVersion: install.istio.io/v1alpha1
kind: IstioOperator
metadata:
  namespace: istio-system
  name: istio-operator
spec:
  profile: default
  components:
    ingressGateways:
      - name: istio-ingressgateway
        enabled: false
  values:
    global:
      proxy:
        resources:
          requests:
            cpu: 50m
      # These below should be added by the helm chart install
      meshID: my-mesh
      multiCluster:
        clusterName: my-cluster
      network: my-network
```
* The chart installs "successfully", however the expected modifications to the resource are not applied!

Now for the strange findings that are leaving me confused whether this is a Helm-problem or an Istio-problem...

* I attempted adding `--force` to the `helm upgrade` command above but the same behavior is observed.  _However_, if I repeat the upgrade a 2nd time with `--force` then the new values get added.
* If I do this same experiment on some other "normal" resource, like an arbitrary Opaque secret, it seems to work just fine on the first try, and no `--force` necessary. 

Are CRDs handled as a special case by Helm?
Could the IstioOperator CRDs be rejecting the modifications but Helm isn't able to see that?  If so, why does it magically work after the 2nd upgrade with --force?

---

Output of `helm version`:
```
version.BuildInfo{Version:"v3.10.1", GitCommit:"9f88ccb6aee40b9a0535fcc7efea6055e1ef72c9", GitTreeState:"clean", GoVersion:"go1.18.7"}
```

Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.14", GitCommit:"3321ffc07d2f046afdf613796f9032f4460de093", GitTreeState:"clean", BuildDate:"2022-11-09T13:40:19Z", GoVersion:"go1.17.13", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.4", GitCommit:"e6c093d87ea4cbb530a7b2ae91e54c0842d8308a", GitTreeState:"clean", BuildDate:"2022-02-16T12:32:02Z", GoVersion:"go1.17.7", Compiler:"gc", Platform:"linux/amd64"}
```
```
Minikube
```
Istio version
```
client version: 1.13.7
control plane version: 1.13.7
data plane version: none
```

cotej:

- This change will produce more reliable patches when modifying these resource types.
- It especially helps when _adopting_ an existing one of these resources into a chart. (closes #11638)

- I guess the biggest concern here would be maintaining existing behavior.  I imagine that this 3-way diff should be the more correct solution overall, but it's tough to say whether it may introduce a behavior change or not in certain cases.  How much confidence does a successful CI pipeline give us in this regard?

**If applicable**:

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -46,6 +46,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/jsonmergepatch"
 	"k8s.io/apimachinery/pkg/util/strategicpatch"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
@@ -385,7 +386,7 @@ func (c *Client) BuildTable(reader io.Reader, validate bool) (ResourceList, erro
 // occurs, a Result will still be returned with the error, containing all
 // resource updates, creations, and deletions that were attempted. These can be
 // used for cleanup or other logging purposes.
-func (c *Client) Update(original, target ResourceList, force bool) (*Result, error) {
+func (c *Client) Update(original, target ResourceList, force, force3WayMergePatch bool) (*Result, error) {
 	updateErrors := []string{}
 	res := &Result{}
 
@@ -420,7 +421,7 @@ func (c *Client) Update(original, target ResourceList, force bool) (*Result, err
 			return errors.Errorf("no %s with the name %q found", kind, info.Name)
 		}
 
-		if err := updateResource(c, info, originalInfo.Object, force); err != nil {
+		if err := updateResource(c, info, originalInfo.Object, force, force3WayMergePatch); err != nil {
 			c.Log("error updating the resource %q:\n\t %v", info.Name, err)
 			updateErrors = append(updateErrors, err.Error())
 		}
@@ -608,7 +609,7 @@ func deleteResource(info *resource.Info, policy metav1.DeletionPropagation) erro
 	return err
 }
 
-func createPatch(target *resource.Info, current runtime.Object) ([]byte, types.PatchType, error) {
+func createPatch(target *resource.Info, current runtime.Object, force3WayMergePatch bool) ([]byte, types.PatchType, error) {
 	oldData, err := json.Marshal(current)
 	if err != nil {
 		return nil, types.StrategicMergePatchType, errors.Wrap(err, "serializing current configuration")
@@ -636,7 +637,7 @@ func createPatch(target *resource.Info, current runtime.Object) ([]byte, types.P
 
 	// Unstructured objects, such as CRDs, may not have an not registered error
 	// returned from ConvertToVersion. Anything that's unstructured should
-	// use the jsonpatch.CreateMergePatch. Strategic Merge Patch is not supported
+	// use the jsonmergepatch.CreateThreeWayJSONMergePatch. Strategic Merge Patch is not supported
 	// on objects like CRDs.
 	_, isUnstructured := versionedObject.(runtime.Unstructured)
 
@@ -645,6 +646,15 @@ func createPatch(target *resource.Info, current runtime.Object) ([]byte, types.P
 
 	if isUnstructured || isCRD {
 		// fall back to generic JSON merge patch
+
+		// TODO Helm 4: the logic controlled by this flag should become the default.
+		//      The flag can then be removed from all the commands where it is accepted: install, upgrade, & rollback.
+		if force3WayMergePatch {
+			patch, err := jsonmergepatch.CreateThreeWayJSONMergePatch(oldData, newData, currentData)
+			return patch, types.MergePatchType, err
+		}
+
+		// 2-way merge patch logic that is kept to maintain backwards-compatibility for Helm 3
 		patch, err := jsonpatch.CreateMergePatch(oldData, newData)
 		return patch, types.MergePatchType, err
 	}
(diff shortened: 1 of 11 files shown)


cotej:
Rebased + force-pushed to resolve a conflict that was in `go.mod`.

flavioelawi:
@cotej you may have just fixed an issue I was facing for some time, just FYI maybe also update the comments referring to `jsonpatch.CreateMergePatch`

cotej:
Thanks @flavioelawi, I hadn't noticed the stale comment - fixed.

joejulian:
@cotej I really don't know anything about the new library, what side-affects it might bring, nor why it will "produce more reliable patches". Could you update the description to justify this change more, please?

cotej:
Hi @joejulian 

The crux of the issue is that this branch of code that's responsible for generating patches against unstructured & CRDs resources was missed when the original changes went in to start using 3-way merge patching.
See: 18ca0dd6c4c9bd4c0c95f741c92376ba8cc3bc57

So this effectively changed the `createPatch` functionality from:
```
if unstructured or crd:
  perform 2-way merge patch using: oldData + newData
else
  perform 2-way strategic merge patch using: oldData + newData
```

to:
```
if unstructured or crd:
  perform 2-way merge patch using: oldData + newData
else
  perform 3-way strategic merge patch using: oldData + newData + currentData
```

This means that patching performed on unstructured & CRD resources completely ignores the live state of the resource (`currentData`), and is subject to the pitfalls outlined in the Helm docs around 2-way vs. 3-way patching:
https://helm.sh/docs/faq/changes_since_helm2/#improved-upgrade-strategy-3-way-strategic-merge-patches

joejulian (dismissed):
While I can't see a good reason for modifying a CRD and having a 3-way merge and think it shouldn't be done, if someone agrees with me and doesn't modify them this doesn't break anything. If they did modify them, hopefully they understand why they did and want this behavior.

I think it's fine but I still want a second review.

pkg/kube/client.go
```diff
@@ -640,7 +640,7 @@ func createPatch(target *resource.Info, current runtime.Object) ([]byte, types.P
 
 	if isUnstructured || isCRD {
 		// fall back to generic JSON merge patch
-		patch, err := jsonpatch.CreateMergePatch(oldData, newData)
+		patch, err := jsonmergepatch.CreateThreeWayJSONMergePatch(oldData, newData, currentData)
```
mattfarina, line 643: Due to our backwards compatibility guarantees, this change needs to be opt-in. We need to capture a command line bool flag (that can also have an environment variable). Based on that we need to use the old or new path. We need to use the old path by default.

In the Helm developer call we discussed when we could make it the default (i.e. Helm v4) and when we can start to talk about that. A todo to make it the default or only option in Helm v4 would be great.

mattfarina (changes requested):
Thanks for this pull request. This is a change I would like to see. There is one change we need to deal with Helm's backwards compatibility guarantees.

joejulian:
@cotej can you address Matt's comments?

pkg/kube/client.go
cotej: CLI flag and TODO have been added.

antoinedeschenes:
Any blockers left?

lemonup:
hi, any blockers left?

el4v:
any blockers left?

Shashi5041:
Is this solution merged and validated ?

bruksnys:
We are interested in this feature, is there anything that can be done to help this get merged?

cotej:
I've rebased to resolve conflicts.  @mattfarina is there any further changes you're looking for here?  CLI flag and TODO were added some time ago.

Phil1602:
@cotej @mattfarina any ideas how to proceed here?

This feature seems to be non-breaking (because of the implemented switch) and would improve the handling of CRs within helm a lot!

cotej:
I've tried many times to get some attention back to this from the maintainers, both on here and on the community Slack, but it seems like this issue's fallen into a blackhole.  There's a number of conflicts that have popped up again, but I won't attempt to resolve them unless we see a pulse.

I can see there's another newer PR now that fixes the same issue here: https://github.com/helm/helm/pull/13411

It had some brief attention, but also looks to be lingering now.  Hopefully at least one of them will get merged before the faulty behavior becomes the default in Helm v4.

frittentheke:
> I can see there's another newer PR now that fixes the same issue here: #13411

Which itself is an improved resubmit of https://github.com/helm/helm/pull/9938

frittentheke:
@joejulian ?

frittentheke:
@joejulian
@mattfarina

could you kindly give a update on this one here?

PR #11651, opened 2022-12-16
Avoid confusing error when passing in '--version X.Y.Z' to get-helm

### LINKED ISSUE #11578: Cannot install helm, invalid checksum
```
$ /tmp/get_helm.sh --version 3.10.2
Helm 3.10.2 is available. Changing from version v3.6.3.
Downloading https://get.helm.sh/helm-3.10.2-linux-amd64.tar.gz
Verifying checksum... SHA sum of /tmp/helm-installer-qyyrng/helm-3.10.2-linux-amd64.tar.gz does not match. Aborting.
Failed to install helm with the arguments provided: --version 3.10.2
```

It does seem that the file I'm getting from the releases page does not match the checksum displayed therein...

```
$ sha256sum ~/Downloads/helm-v3.10.2-linux-amd64.tar.gz 
2315941a13291c277dac9f65e75ead56386440d3907e0540bf157ae70f188347  /home/callek/Downloads/helm-v3.10.2-linux-amd64.tar.gz
```

Callek:
closes #11578

This magically adds "v" to the start of the desired version string if it was not passed in.

**If applicable**:

CC: @joejulian

### DIFF
--- a/scripts/get
+++ b/scripts/get
@@ -207,6 +207,10 @@ while [[ $# -gt 0 ]]; do
        shift
        if [[ $# -ne 0 ]]; then
            export DESIRED_VERSION="${1}"
+           if [[ "$1" != "v"* ]]; then
+               echo "Expected version arg ('${DESIRED_VERSION}') to begin with 'v', fixing..."
+               export DESIRED_VERSION="v${1}"
+           fi
        else
            echo -e "Please provide the desired version. e.g. --version v2.4.0 or -v latest"
            exit 0
--- a/scripts/get-helm-3
+++ b/scripts/get-helm-3
@@ -299,6 +299,10 @@ while [[ $# -gt 0 ]]; do
        shift
        if [[ $# -ne 0 ]]; then
            export DESIRED_VERSION="${1}"
+           if [[ "$1" != "v"* ]]; then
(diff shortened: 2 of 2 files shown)


scripts/get
```diff
@@ -207,6 +207,10 @@ while [[ $# -gt 0 ]]; do
        shift
        if [[ $# -ne 0 ]]; then
            export DESIRED_VERSION="${1}"
+           if [[ "$1" != "v"* ]]; then
```
joejulian, line 210: This will fail if the version is `latest`

mattfarina (dismissed):
Can you please address Joe's feedback.

scripts/get
lucasrod16: Wanted to make note that the current install script on `main` will fail if you pass `--version latest` to it as well, so this doesn't appear to be a change in behavior. The default behavior of the script to fetch the `latest` version by default when no `--version` flag is passed in is still functioning as expected. These changes appear to fix the issue that was raised.

lucasrod16: Adding a check like: `&& [[ "$1" != "latest" ]]` to the if statement hushes the output that notifies the user that it's fixing the version input when it fails. It may be less confusing to hush the output when it does fail

joejulian (approved):

PR #162, opened 2015-12-10
Clean up the printing of the types, show the short url 

vaikas:
- Display collection which was omitted previously
- show the github name you would use in templates to refer to the type.

PR #11674, opened 2022-12-25
Implementing AllNamespaces in action.NewList

### LINKED ISSUE #11656: action.NewList does not implement AllNamespaces properly.
I'm using helm as a go library where I'm trying to get all deployments across all namespaces.

```go
listClient := action.NewList(x.ActionConfig)
listClient.AllNamespaces = true
listClient.All = true
releases, _ := listClient.Run()
```

The above case returns only deployments that are in the namespace of the first initial ActionConfiguration.

I looked at the helm CLI tool to understand the implementation and found the following:
https://github.com/helm/helm/blob/538f6aecca19f0a4288168661cdde8f9a2e76f08/cmd/helm/list.go#L61-L62
https://github.com/helm/helm/blob/538f6aecca19f0a4288168661cdde8f9a2e76f08/cmd/helm/list.go#L72-L74

If I understand it correctly, it's mutating the configuration in order to be able to get the deployments in all the namespaces. I might be wrong here but I assume that an action should be responsible for something like AllNamespaces just like it does with Limit? Am I wrong if I think AllNamespaces should be fulfilled by NewList.Run rather than List Command?

Also, the test around AllNamespaces have release cases all with the same namespace. 
https://github.com/helm/helm/blob/a499b4b179307c267bdf3ec49b880e3dbd2a5591/pkg/action/list_test.go#L79-L82
https://github.com/helm/helm/blob/a499b4b179307c267bdf3ec49b880e3dbd2a5591/pkg/action/list_test.go#L263-L276

VaibhavSharma-47:
Signed-off-by: Vaibhav Sharma <17532va@gmail.com>

The goal of this MR is to move the implementation of --all-namespace from cmd to action
I am not too sure about the unit test. I would suggestions on how to do that better. 
**If applicable**:
closes #11656

### DIFF
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -19,7 +19,6 @@ package action
 import (
 	"bytes"
 	"fmt"
-	"log"
 	"os"
 	"path"
 	"path/filepath"
@@ -45,17 +44,6 @@ import (
 	"helm.sh/helm/v3/pkg/storage/driver"
 	"helm.sh/helm/v3/pkg/time"
 )
-
-func init() {
-	log.SetFlags(log.Lshortfile)
-}
-
-func debug(format string, v ...interface{}) {
-	if settings.Debug {
-		format = fmt.Sprintf("[debug] %s\n", format)
-		log.Output(2, fmt.Sprintf(format, v...))
-	}
-}
 
 // Timestamper is a function capable of producing a timestamp.Timestamper.
 //
@@ -113,8 +101,7 @@ type Configuration struct {
 //
 // TODO: This function is badly in need of a refactor.
 // TODO: As part of the refactor the duplicate code in cmd/helm/template.go should be removed
-//
-//	This code has to do with writing files to disk.
+//       This code has to do with writing files to disk.
 func (cfg *Configuration) renderResources(ch *chart.Chart, values chartutil.Values, releaseName, outputDir string, subNotes, useReleaseName, includeCrds bool, pr postrender.PostRenderer, dryRun bool) ([]*release.Hook, *bytes.Buffer, string, error) {
 	hs := []*release.Hook{}
 	b := bytes.NewBuffer(nil)
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -19,6 +19,7 @@ package action
 import (
 	"bytes"
 	"fmt"
+	"log"
 	"os"
 	"path"
 	"path/filepath"
@@ -44,6 +45,17 @@ import (
 	"helm.sh/helm/v3/pkg/storage/driver"
 	"helm.sh/helm/v3/pkg/time"
 )
+
+func init() {
+	log.SetFlags(log.Lshortfile)
+}
+
+func debug(format string, v ...interface{}) {
+	if settings.Debug {
+		format = fmt.Sprintf("[debug] %s\n", format)
+		log.Output(2, fmt.Sprintf(format, v...))
+	}
+}
 
 // Timestamper is a function capable of producing a timestamp.Timestamper.
 //
@@ -101,7 +113,8 @@ type Configuration struct {
 //
 // TODO: This function is badly in need of a refactor.
 // TODO: As part of the refactor the duplicate code in cmd/helm/template.go should be removed
-//       This code has to do with writing files to disk.
+//
+//	This code has to do with writing files to disk.
 func (cfg *Configuration) renderResources(ch *chart.Chart, values chartutil.Values, releaseName, outputDir string, subNotes, useReleaseName, includeCrds bool, pr postrender.PostRenderer, dryRun bool) ([]*release.Hook, *bytes.Buffer, string, error) {
 	hs := []*release.Hook{}
 	b := bytes.NewBuffer(nil)
--- a/pkg/action/list.go
+++ b/pkg/action/list.go
@@ -152,13 +152,14 @@ func (l *List) Run() ([]*release.Release, error) {
 		return nil, err
 	}
 
+	cfg := l.cfg
 	if l.AllNamespaces {
 		// For Empty Namespace the default behavior is to get all namespaces
 		clonedCfg, err := l.cfg.CloneWithNewNamespace("")
 		if err != nil {
 			return nil, err
 		}
-		l.cfg = clonedCfg
+		cfg = clonedCfg
 	}
 
 	var filter *regexp.Regexp
@@ -170,7 +171,7 @@ func (l *List) Run() ([]*release.Release, error) {
 		}
 	}
 
-	results, err := l.cfg.Releases.List(func(rel *release.Release) bool {
+	results, err := cfg.Releases.List(func(rel *release.Release) bool {
 		// Skip anything that doesn't match the filter.
 		if filter != nil && !filter.MatchString(rel.Name) {
 			return false
--- a/pkg/action/list.go
+++ b/pkg/action/list.go
@@ -17,14 +17,18 @@ limitations under the License.
 package action
 
 import (
+	"os"
 	"path"
 	"regexp"
 
 	"k8s.io/apimachinery/pkg/labels"
 
+	"helm.sh/helm/v3/pkg/cli"
 	"helm.sh/helm/v3/pkg/release"
 	"helm.sh/helm/v3/pkg/releaseutil"
 )
+
+var settings = cli.New()
 
 // ListStates represents zero or more status codes that a list item may have set
 //
@@ -148,6 +152,12 @@ func NewList(cfg *Configuration) *List {
 func (l *List) Run() ([]*release.Release, error) {
 	if err := l.cfg.KubeClient.IsReachable(); err != nil {
 		return nil, err
+	}
+
+	if l.AllNamespaces {
+		if err := l.cfg.Init(settings.RESTClientGetter(), "", os.Getenv("HELM_DRIVER"), debug); err != nil {
+			return nil, err
+		}
 	}
 
 	var filter *regexp.Regexp
--- a/pkg/action/list.go
+++ b/pkg/action/list.go
@@ -22,13 +22,11 @@ import (
 	"strings"
 
 	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/cli-runtime/pkg/genericclioptions"
 
-	"helm.sh/helm/v3/pkg/cli"
 	"helm.sh/helm/v3/pkg/release"
 	"helm.sh/helm/v3/pkg/releaseutil"
 )
-
-var settings = cli.New()
 
 // ListStates represents zero or more status codes that a list item may have set
 //
@@ -340,7 +338,7 @@ func (l *List) SetStateMask() {
 func (cfg *Configuration) CloneWithNewNamespace(namespace string) (*Configuration, error) {
 	newConf := &Configuration{}
 	newConf.Releases = cfg.Releases
-	getter := settings.RESTClientGetter()
+	getter := cfg.RESTClientGetter.(genericclioptions.RESTClientGetter)
 	helmDriver := strings.ToLower(cfg.Releases.Driver.Name()) //note: Driver.Name() returns capitalized driver names.
 
 	err := newConf.Init(getter, namespace, helmDriver, cfg.Log)
--- a/pkg/action/list.go
+++ b/pkg/action/list.go
@@ -17,9 +17,9 @@ limitations under the License.
 package action
 
 import (
-	"os"
 	"path"
 	"regexp"
+	"strings"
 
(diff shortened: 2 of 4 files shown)


pkg/action/action.go
```diff
@@ -45,6 +46,17 @@ import (
 	"helm.sh/helm/v3/pkg/time"
 )
 
+func init() {
+	log.SetFlags(log.Lshortfile)
+}
+
+func debug(format string, v ...interface{}) {
```
AleksandarFaraj, line 53: I believe this is your own debug code, this should not be here.

pkg/action/list.go
```diff
@@ -150,6 +154,12 @@ func (l *List) Run() ([]*release.Release, error) {
 		return nil, err
 	}
 
+	if l.AllNamespaces {
+		if err := l.cfg.Init(settings.RESTClientGetter(), "", os.Getenv("HELM_DRIVER"), debug); err != nil {
```
AleksandarFaraj, line 158: Regarding #11656 : 
There are two problems the issue raises. 
The first is that the namespace logic is in the helm/cli/list.go instead of pkg/action/list.go.
The second is that the solution depends on a mutating a configuration state.

The context:
Generally speaking; the code is used by CLI and the code could also be used as a package (hint: https://helm.sh/docs/topics/advanced/)

Both the master branch and this PR, do this by mutating the configuration.

For CLI: The code works because configuration is reset before each command.
https://github.com/helm/helm/blob/c137bfb68b4f8b7a10687f043b0dbccb09574969/cmd/helm/helm.go#L72-L75

The issue for package users:
The current implementation, puts requirements on the package users implementations. They _must_ reset the configuration each run.
If not: 
1. There is a risk of exposing unwanted namespaces due to either concurrent or subsequent uses of the client.
2. The code will break for subsequent use after AllNamespaces is utilized.

I don't think this is safe as it creates a trap with dangerous default behavior that gets triggered only when AllNamespaces is used, all other code with work without resetting the configuration.

Suggestion: 
I think the code should clone/create a new configuration that then is mutated as line 157.

AleksandarFaraj: `if err := l.cfg.Init(settings.RESTClientGetter(), "", os.Getenv("HELM_DRIVER"), debug); err != nil {`
should be:
`if err := l.cfg.Init(cfg.RESTClientGetter(), "", os.Getenv("HELM_DRIVER"), debug); err != nil {`

pkg/action/list.go
```diff
@@ -17,15 +17,19 @@ limitations under the License.
...
 	"regexp"
 
 	"k8s.io/apimachinery/pkg/labels"
 
+	"helm.sh/helm/v3/pkg/cli"
 	"helm.sh/helm/v3/pkg/release"
 	"helm.sh/helm/v3/pkg/releaseutil"
 )
 
+var settings = cli.New()
```
AleksandarFaraj, line 31: I think this line can be removed in favor of:

`if err := l.cfg.Init(l.cfg.RESTClientGetter, "", os.Getenv("HELM_DRIVER"), debug); err != nil {`

AleksandarFaraj:
@VaibhavSharma-47 Added comments on the PR.

Summary:
The issue is two-fold. First, of course the logic is in the wrong place. Second, the mutation of configuration creates a "contract" or "need" for package users to implement configuration resetting.

Suggestion:
Add a clone/copy function to the configuration.

In action.go it could look like
```
func (cfg *Configuration) CloneWithNewNamespace(namespace string) (*Configuration, error) {
	newConf := &Configuration{}
	getter := cfg.RESTClientGetter.(genericclioptions.RESTClientGetter)
	helmDriver := strings.ToLower(newConf.Releases.Driver.Name()) //note: Driver.Name() returns capitalized driver names. 

	err := newConf.Init(getter, namespace, helmDriver, cfg.Log)
	return newConf, err
}
```

List.go would still mutate the List.Configuration and it could be avoided by doing using a local configuration that gets set if AllNamespaces=true. Could be as follows:
```
 func (l *List) Run() ([]*release.Release, error) {
	configuration := l.cfg 

	if l.AllNamespaces {
		clonedCfg, err := configuration.CloneWithNewNamespace("")
		if  err != nil {
			return nil, err
		}
		configuration = clonedCfg
	}
	if err := configuration.KubeClient.IsReachable(); err != nil {
		return nil, err
	}
```

pkg/action/action.go
```diff
@@ -101,7 +113,8 @@ type Configuration struct {
 //
 // TODO: This function is badly in need of a refactor.
 // TODO: As part of the refactor the duplicate code in cmd/helm/template.go should be removed
-//       This code has to do with writing files to disk.
+//
```
AleksandarFaraj, line 116: not relevant.

pkg/action/list_test.go
```diff
@@ -286,6 +288,32 @@ func makeMeSomeReleases(store *storage.Storage, t *testing.T) {
...
+		if err := store.Create(rel); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	_, err := store.ListDeployed()
+	assert.NoError(t, err)
+	//assert.Len(t, all, 3, "sanity test: three items added")
+}
+
```
AleksandarFaraj, line 316: Regarding testing, I'm not that familiar with go to know what's possible, but some ideas are:
Test if AllNamespaces utilizes a cloned cfg.
Test if original cfg is used/accessed after the cloning has happened (make sure devs don't accidently use l.cfg instead of suggested configuration := l.cfg).
Test if original cfg is still the same after Run().
Test if all the driver Name() strings map to Configuration.Init(helmDriver string)

VaibhavSharma-47:
@AleksandarFaraj Working on it, I will update once done

VaibhavSharma-47:
@AleksandarFaraj I have done the requested changes, please let me know if any more comments

pkg/action/list_test.go
VaibhavSharma-47: I have added a test to check if we get all the releases in all namespaces and testing if the config is cloned.

pkg/action/list.go
```diff
@@ -322,3 +335,15 @@ func (l *List) SetStateMask() {
 
 	l.StateMask = state
 }
+
+// Create a clone of the config with a new namespace
+func (cfg *Configuration) CloneWithNewNamespace(namespace string) (*Configuration, error) {
+	newConf := &Configuration{}
+	newConf.Releases = cfg.Releases
+	getter := settings.RESTClientGetter()
```
AleksandarFaraj, line 343: getter := cfg.RESTClientGetter.(genericclioptions.RESTClientGetter) does this type cast not work?

AleksandarFaraj, line 343:
```go
getter := cfg.RESTClientGetter.(genericclioptions.RESTClientGetter)
```

Does this type cast not work? It's better to reuse if possible.

pkg/action/list.go
```diff
@@ -150,6 +154,15 @@ func (l *List) Run() ([]*release.Release, error) {
 		return nil, err
 	}
 
+	if l.AllNamespaces {
+		// For Empty Namespace the default behavior is to get all namespaces
+		clonedCfg, err := l.cfg.CloneWithNewNamespace("")
+		if err != nil {
+			return nil, err
+		}
+		l.cfg = clonedCfg
```
AleksandarFaraj, line 163:
```go
l.cfg = clonedCfg
```
Looks good! I don't know if this matters but this approach potentially breaks reusing an list action. Do you think that's a problem?

pkg/action/list.go
```diff
@@ -322,3 +335,15 @@ func (l *List) SetStateMask() {
 
 	l.StateMask = state
 }
+
+// Create a clone of the config with a new namespace
+func (cfg *Configuration) CloneWithNewNamespace(namespace string) (*Configuration, error) {
```
AleksandarFaraj, line 340: I didn't think of the fact that action.go and list.go are in the same package. CloneWithNewNamespace could be changed to private cloneWithNewNamespace

pkg/action/list.go
gjenkins8: I agree with @AleksandarFaraj -- since `l.cfg` is only used [#L175](https://github.com/helm/helm/pull/11674/files#diff-957d45c5563ffa9a681a8aa0bb489b1297b8e8afed5858d7bfd10fce32c94ad2R175), it should be simple to do something like:

```
cfg := l.cfg
if l.AllNamespaces {
    cfg, err := l.cfg.CloneWithNewNamespace("")
    ...
```

And below do:
```
results, err := cfg.Releases.List(func(rel *release.Release) bool {
```

gjenkins8 (commented):
Thanks greatly for the PR @VaibhavSharma-47 !. And for the nice review(s) @AleksandarFaraj

I've left a couple of additional (small) comments. And I've added this PR to Thursday's developer call agenda.

pkg/action/list.go
```diff
@@ -19,13 +19,17 @@ package action
...
+	"strings"
 
 	"k8s.io/apimachinery/pkg/labels"
 
+	"helm.sh/helm/v3/pkg/cli"
 	"helm.sh/helm/v3/pkg/release"
 	"helm.sh/helm/v3/pkg/releaseutil"
 )
 
+var settings = cli.New()
```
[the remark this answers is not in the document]
AleksandarFaraj: It's related to: https://github.com/helm/helm/pull/11674#discussion_r1060874323

If getter takes form cfg instead of settings then this can be removed. 

```go
getter := settings.RESTClientGetter()

---->

getter := cfg.RESTClientGetter.(genericclioptions.RESTClientGetter)
```

pkg/action/list.go
AleksandarFaraj: https://github.com/helm/helm/blob/55f35060cbc662df056801fafc5c33ce1e4a5ad4/pkg/action/list.go#L152-L157
Would it be better to make this declaration before the reachable check such that it checks the configuration it cloned over the previous one?

pkg/action/list.go
VaibhavSharma-47: My bad, changed it now

pkg/action/list.go
VaibhavSharma-47: I may be wrong but aren't you declaring cfg twice this way ?

```
	cfg := l.cfg
	if l.AllNamespaces {
		cfg, err := l.cfg.CloneWithNewNamespace("")
		// For Empty Namespace the default behavior is to get all namespaces
		if err != nil {
			return nil, err
		}
	}
```
So below:
```
results, err := cfg.Releases.List(func(rel *release.Release) bool {
```
Will this work ?

AleksandarFaraj: You are right, I wrote this app to test it. Just use the clonedCfg and then cfg = clonedCfg

```go
package main

func main() {
	a := true
	b := 5
	if a {
		b, e := r() // b declared but not used (Seems to be local to the if-statement).
		if e != nil {
			print(e)
		}
	}
	println(b)
}
func r() (int, error) {
	return 5, nil
}
```

gjenkins8: > I may be wrong but aren't you declaring cfg twice this way ?

Yep, I was trying to keep my example concise (perhaps too concise :) ). I've updated it to be more correct.

pkg/action/list.go
```diff
@@ -150,6 +152,15 @@ func (l *List) Run() ([]*release.Release, error) {
...
 	}
 
+	if l.AllNamespaces {
+		// For Empty Namespace the default behavior is to get all namespaces
+		clonedCfg, err := l.cfg.CloneWithNewNamespace("")
+		if err != nil {
+			return nil, err
+		}
+		l.cfg = clonedCfg
+	}
```
gjenkins8, line 162: I think to get this PR over the finish line, just this small change is needed (and corresponding change line #162, but github won't let me make a suggestion there via the UI). Thoughts?

```suggestion
       cfg := l.cfg
	if l.AllNamespaces {
		// For Empty Namespace the default behavior is to get all namespaces
		clonedCfg, err := l.cfg.CloneWithNewNamespace("")
		if err != nil {
			return nil, err
		}
		cfg = clonedCfg
	}
```

gjenkins8 (changes requested):

pkg/action/list.go
VaibhavSharma-47: Correct me if I am wrong but this change is required as this will fix [this](https://github.com/helm/helm/issues/11656#issuecomment-1370160073) comment ?

AleksandarFaraj:
Yes that’s correct.

tors 12 jan. 2023 kl. 20:15 skrev Vaibhav Sharma ***@***.***>:

VaibhavSharma-47:
@AleksandarFaraj @gjenkins8 What is stopping this MR from getting merged ?

PR #11675, opened 2022-12-25
fix(pkg/plugin/installer): unix-style paths in HELM_PLUGINS

### LINKED ISSUE #11310: Helm supports unix style path in HELM_PLUGINS for loading, but fails when installing new plugins
Helm can parse a `:` separated list of paths in HELM_PLUGINS.
https://github.com/helm/helm/blob/8199db309ae73d745b983fb7ea2bfdd944a834ae/pkg/plugin/plugin.go#L261-L262

which works for loading plugins from multiple directories.

For example, installing two versions of the helm diff plugin in two paths under /tmp:
```
$ HELM_PLUGINS=/tmp/xyz helm plugin install https://github.com/databus23/helm-diff --version v3.2.0
$ HELM_PLUGINS=/tmp/abc helm plugin install https://github.com/databus23/helm-diff --version v3.5.0
```
Both versions show up when listing plugins
```
$ HELM_PLUGINS=/tmp/abc:/tmp/xyz helm plugin list
NAME    VERSION DESCRIPTION                           
diff    3.5.0   Preview helm upgrade changes as a diff
diff    3.2.0   Preview helm upgrade changes as a diff
```
and the plugin in the leftmost directory in HELM_PLUGINS takes precedence.
```
$ HELM_PLUGINS=/tmp/abc:/tmp/xyz helm diff version
3.5.0
```
But, when installing a plugin with this path set:
```
$ HELM_PLUGINS=/tmp/abc:/tmp/xyz helm plugin install https://github.com/jkroepke/helm-secrets --version v3.12.0
Installed plugin: secrets
```
it "succeeds" in the install and doesn't give any error, but the plugin does not show up in the installed list of plugins.
```
$ HELM_PLUGINS=/tmp/abc:/tmp/xyz helm plugin list
NAME    VERSION DESCRIPTION                           
diff    3.5.0   Preview helm upgrade changes as a diff
diff    3.2.0   Preview helm upgrade changes as a diff

$ HELM_PLUGINS=/tmp/abc:/tmp/xyz helm secrets version
Error: unknown command "secrets" for "helm"
```
The new plugin has been installed in the literal path `/tmp/abc:/tmp/xyz/`, which is not searched when loading plugins because the paths are split.
```
$ ls '/tmp/abc:/tmp/xyz/'
helm-secrets
```

My intuition would be that in this case, helm would take the leftmost path from HELM_PLUGINS and install the new plugin there.

My use case for this is: I provide helm binaries over NFS for users, and I could also provide a base set of plugins by setting up their environment with the default location in their home directory, plus a plugins directory over NFS e.g `/home/USER/.local/share/helm/plugins:/nfs/apps/kubernetes/helm/3.9.2/plugins`. They would get the base set of plugins from NFS, but not be prevented from installing new plugins for themselves in their home directory.

---

Output of `helm version`:
```
$ helm version
version.BuildInfo{Version:"v3.9.2", GitCommit:"1addefbfe665c350f4daf868a9adc5600cc064fd", GitTreeState:"clean", GoVersion:"go1.17.12"}
```

xodmqjw:
Supports unix-style paths (with a colon) in the HELM_PLUGINS env for installing new plugins.

Closes #11310

Signed-off-by: Sergei Karpov <xodmqjw@gmail.com>

**If applicable**:

### DIFF
--- a/pkg/plugin/installer/base.go
+++ b/pkg/plugin/installer/base.go
@@ -30,9 +30,10 @@ type base struct {
 
 func newBase(source string) base {
 	settings := cli.New()
+	pluginsDirs := filepath.SplitList(settings.PluginsDirectory)
 	return base{
 		Source:           source,
-		PluginsDirectory: settings.PluginsDirectory,
+		PluginsDirectory: pluginsDirs[0],
 	}
 }
 
--- a/pkg/plugin/installer/base.go
+++ b/pkg/plugin/installer/base.go
@@ -29,12 +29,22 @@ type base struct {
 }
 
 func newBase(source string) base {
-	settings := cli.New()
-	pluginsDirs := filepath.SplitList(settings.PluginsDirectory)
+	pluginsDir := getPluginsDir()
 	return base{
 		Source:           source,
-		PluginsDirectory: pluginsDirs[0],
+		PluginsDirectory: pluginsDir,
 	}
+}
+
+func getPluginsDir() string {
+	settings := cli.New()
+
+	if settings.PluginsDirectory == "" {
+		return ""
+	}
+
+	pluginsDirs := filepath.SplitList(settings.PluginsDirectory)
+	return pluginsDirs[0]
 }
 
 // Path is where the plugin will be installed.
--- a/pkg/plugin/installer/base_test.go
+++ b/pkg/plugin/installer/base_test.go
@@ -14,6 +14,7 @@ limitations under the License.
 package installer // import "helm.sh/helm/v3/pkg/plugin/installer"
 
 import (
+	"fmt"
 	"os"
 	"testing"
 )
@@ -34,7 +35,7 @@ func TestPath(t *testing.T) {
 			expectPath:     "/helm/data/plugins/helm-secrets",
 		}, {
 			source:         "https://github.com/jkroepke/helm-secrets",
-			helmPluginsDir: "/helm/data/plugins:/remote/helm/data/plugins",
+			helmPluginsDir: fmt.Sprintf("/helm/data/plugins%c/remote/helm/data/plugins", os.PathListSeparator),
 			expectPath:     "/helm/data/plugins/helm-secrets",
 		},
 	}
--- a/pkg/plugin/installer/base_test.go
+++ b/pkg/plugin/installer/base_test.go
@@ -32,6 +32,10 @@ func TestPath(t *testing.T) {
 			source:         "https://github.com/jkroepke/helm-secrets",
 			helmPluginsDir: "/helm/data/plugins",
 			expectPath:     "/helm/data/plugins/helm-secrets",
+		}, {
+			source:         "https://github.com/jkroepke/helm-secrets",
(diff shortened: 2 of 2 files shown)


yxxhero (approved):

pkg/plugin/installer/base_test.go
```diff
@@ -32,6 +32,10 @@ func TestPath(t *testing.T) {
 			source:         "https://github.com/jkroepke/helm-secrets",
 			helmPluginsDir: "/helm/data/plugins",
 			expectPath:     "/helm/data/plugins/helm-secrets",
+		}, {
+			source:         "https://github.com/jkroepke/helm-secrets",
+			helmPluginsDir: "/helm/data/plugins:/remote/helm/data/plugins",
```
mattfarina, line 37: This test is going to fail on Windows which is one of our supported environments. We need to use the correct path list separator for each environment. See https://pkg.go.dev/path/filepath#pkg-constants

mattfarina (changes requested):
Thanks for a PR to fix this issue. I have some feedback.

This isn't just for Unix path lists. It applies to Windows as well. You can have path lists there.

I think the docs need to be updated as well. We want to be explicit in telling people where something will be installed if there are multiple locations.

pkg/plugin/installer/base_test.go
xodmqjw: I'm not sure where I have to add docs. How should I add it in the current MR? Could you please give me a link where I can find how to do this?

tghartland:
@gjenkins8 

> I didn't understand the rational of the change here. IIUC, the change is to allow specifying multiple `HELM_PLUGINS` paths, splitting those plugins directory path(s), but then immediately discarding all/any additional paths.

Helm already supports HELM_PLUGINS path with multiple entries, for _reading_ plugins.
https://github.com/helm/helm/blob/8199db309ae73d745b983fb7ea2bfdd944a834ae/pkg/plugin/plugin.go#L261-L262

But if you set up the path like that, it breaks _writing_ (installing) plugins, as the entire env variable is treated as a single directory path.
Helm creates and installs the plugin in that directory, but it is then impossible for it to read the plugin from that directory, as it gets split.

Full example in #11310.

pkg/plugin/installer/base.go
```diff
@@ -30,9 +30,10 @@ type base struct {
 
 func newBase(source string) base {
 	settings := cli.New()
+	pluginsDirs := filepath.SplitList(settings.PluginsDirectory)
 	return base{
 		Source:           source,
-		PluginsDirectory: settings.PluginsDirectory,
+		PluginsDirectory: pluginsDirs[0],
```
mattfarina, line 36: This is almost there. If `settings.PluginsDirectory` is an empty string than calling `pluginsDir[0]` will result in a panic. Can we check the length of the slice and if it's 0 use an empty string for the value.

mattfarina (changes requested):
This is almost there. There is an edge case that can cause a panic. Can you please fix that?

PR #11676, opened 2022-12-27
CircleCI Migration

sabre1041:
Migrates existing workflow on CircleCI job to GitHub actions native build

* Creation of release workflow
* Updates [build-pr job](https://github.com/helm/helm/blob/main/.github/workflows/build-pr.yml) to incorporate building and uploading binaries
* Updates [code-analysis job](https://github.com/helm/helm/blob/main/.github/workflows/codeql-analysis.yml) to pin dependencies

### Tasks:

- [ ] Add `AZURE_STORAGE_CONNECTION_STRING` and `AZURE_STORAGE_CONTAINER_NAME` as GitHub Secrets
- [ ] Ensure all CircleCI assets removed

Builds upon #9989


.github/workflows/release.yml
```diff
@@ -0,0 +1,38 @@
...
+
+      - name: Setup Go
+        uses: actions/setup-go@6edd4406fa81c3da01a34fa6f6343087c207a568 # pin@3.5.0
+        with:
+          go-version: '1.18'
+
+      - name: Set Version
+        id: set-version
+        run: |
+          echo "VERSION=$(echo ${GITHUB_REF/refs\/tags\//})" >> $GITHUB_OUTPUT
```
mattfarina, line 23: Since this is already filtered for tags, is there a reason we aren't using the `GITHUB_REF_NAME` variable?

.github/workflows/release.yml
```diff
@@ -0,0 +1,38 @@
+name: release
```
mattfarina, line 1: Should this PR remove the deploy from CircleCI? So that the next release does not have two systems pushing it.

.github/workflows/build-pr.yml
```diff
@@ -27,3 +30,32 @@ jobs:
         run: make test-style
       - name: Run unit tests
         run: make test-coverage
+      - name: Test build
+        run: make
+
+  canary-release:
```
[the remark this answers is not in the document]
sabre1041: Everything CircleCI should be removed. I can certainly add the removal of the circleci script if we would like in this PR.

.github/workflows/release.yml
sabre1041: Everything CircleCI should be removed. I can certainly add the removal of the circleci script if we would like in this PR.

.github/workflows/release.yml
sabre1041: Fixed

.github/workflows/build-pr.yml
```diff
@@ -27,3 +30,32 @@ jobs:
...
+        run: |
+          make build-cross
+          make dist checksum VERSION="canary"
+
+      - name: Upload Binaries
+        uses: bacongobbler/azure-blob-storage-upload@50f7d898b7697e864130ea04c303ca38b5751c50 # pin@3.0.0
+        with:
+          source_dir: _dist
+          container_name: ${{ secrets.AZURE_STORAGE_CONTAINER_NAME }}
+          connection_string: ${{ secrets.AZURE_STORAGE_CONNECTION_STRING }}
```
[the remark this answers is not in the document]
sabre1041: done

scottrigby (commented):
Hey @sabre1041 👋 I closed #11820 in favor of this one. I created that PR only because I thought you said this work had yet to be finished and I got frustrated with Circle again so just busted it out. I see now that you meant your PR just needed to be finished. Kind of good to see though that the solution is roughly the same in both PRs.

As I said in https://github.com/helm/helm/pull/11820#issuecomment-1433346734:

> I think there are some things in my PR that would improve yours, but your PR has good things like pinning by sha, and it was first. So let's work on it together from yours and get this in 😄 

Here's what 'd like to see merged from the two:

- Pinning by sha:
    Nice improvement overall (for everyone else, this is rec'd by GitHub [actions hardening doc](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions))
- adding `make test build` from Circle config:
    Both PRs add this because it was missing, and this is clearly the logical place 👍 
- adding logic from deploy.sh to Circle config:
    - I added them both to a single release workflow, just as they were together in the script
    - You added the canary ("main" branch) release to build-pr.yml, and tagged releases to a new release.yml. But I don't think canary releases belong in the PR workflow (honestly that should just be renamed "test", but either way it is just for PRs). I also think there's value in the release to AZ logic in the same file, just so that practically when improvements are made they're done in the same place. The only difference between the two in the script is a single `--overwrite` flag, which I noted in my PR
- Very minor. It's not necessary to export `VERSION` from a previous build step

.github/workflows/build-pr.yml
```diff
@@ -3,14 +3,17 @@ on:
...
     runs-on: ubuntu-latest
     steps:
       - name: Checkout source code
-        uses: actions/checkout@v2
+        uses: actions/checkout@755da8c3cf115ac066823e79a1e1788f8940201b # pin@v3.2.0
       - name: Setup Go
-        uses: actions/setup-go@v2
+        uses: actions/setup-go@6edd4406fa81c3da01a34fa6f6343087c207a568 # pin@3.5.0
         with:
           go-version: '1.18'
```
lucasrod16, line 18: This is more of a question/suggestion than a "we should do this". Curious if there would be any value in using the `go-version-file` input with the setup-go action. It takes the path to the `go.mod` file, and uses the version specified there, so when the go version is updated, it would automatically be reflected in the workflows. [Link](https://github.com/actions/setup-go#getting-go-version-from-the-gomod-file) to the docs. My thinking is that it would be one less place to worry about updates/maintenance, but I'm curious to hear others' thoughts.

sabre1041: thats a new(er) enhancement and I like that it will reduce the number of changes that need to be made when adopting new versions of go.

Thanks for calling it out!

scottrigby: @lucasrod16 I didn't know about this feature. Very nice!

scottrigby: @lucasrod16 for now, we have to revert back to specifying the version explicitly. Because it appears there was a regression (mismatch between what's actually in go.mod and the circle tests), that was introduced here 

https://github.com/helm/helm/blob/017785a2f1104eb4299f8d31feb550c213350f13/go.mod#L3

in this commit https://github.com/helm/helm/pull/10912/commits/ffa19a4b5d836283a91a4c16f8b81e734a973afc (see line https://github.com/helm/helm/pull/10912/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6) from this PR #10912. Note the commit comment:

> fix circleci tests

We revert go.mod back to 1.18 here but that would be scope creep for this PR. Let's do that in a follow-up PR, at which point we could use the `go-version-file` option.

.github/workflows/build-pr.yml
```diff
@@ -3,16 +3,19 @@ on:
...
     steps:
       - name: Checkout source code
-        uses: actions/checkout@v2
+        uses: actions/checkout@755da8c3cf115ac066823e79a1e1788f8940201b # pin@v3.2.0
       - name: Setup Go
-        uses: actions/setup-go@v2
+        uses: actions/setup-go@6edd4406fa81c3da01a34fa6f6343087c207a568 # pin@3.5.0
         with:
-          go-version: '1.18'
+          go-version-file: go.mod
```
scottrigby, line 18: See https://github.com/helm/helm/pull/11676#discussion_r1112595497

```suggestion
          go-version: '1.18'
```

.github/workflows/build-pr.yml
```diff
@@ -27,3 +30,32 @@ jobs:
...
+    needs: build
+    if: ${{ github.event == 'push' && github.ref == 'refs/heads/main' && github.repository_owner == 'helm' }}
+    steps:
+      - name: Checkout source code
+        uses: actions/checkout@755da8c3cf115ac066823e79a1e1788f8940201b # pin@v3.2.0
+
+      - name: Setup Go
+        uses: actions/setup-go@6edd4406fa81c3da01a34fa6f6343087c207a568 # pin@3.5.0
+        with:
+          go-version-file: go.mod
```
scottrigby, line 47: See above

```suggestion
          go-version: '1.18'
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,38 @@
...
+    if: ${{ github.repository_owner == 'helm'}}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout source code
+        uses: actions/checkout@755da8c3cf115ac066823e79a1e1788f8940201b # pin@v3.2.0
+
+      - name: Setup Go
+        uses: actions/setup-go@6edd4406fa81c3da01a34fa6f6343087c207a568 # pin@3.5.0
+        with:
+          go-version-file: go.mod
```
scottrigby, line 18: See above

```suggestion
          go-version: '1.18'
```

.github/workflows/build-pr.yml
scottrigby: Made follow-up issue https://github.com/helm/helm/issues/11838

scottrigby (changes requested):
Change back to explicitly setting Go version. See https://github.com/helm/helm/pull/11676#discussion_r1112595497 and follow-up issue https://github.com/helm/helm/issues/11838.

.github/workflows/build-pr.yml
lucasrod16: @scottrigby ahh I see, thanks for the context! That sounds good, I can write-up an issue and link this discussion for more context.

@sabre1041 apologies for the side-track here

lucasrod16: @scottrigby nevermind I see that you've created a follow-up [issue](https://github.com/helm/helm/issues/11838) already.

scottrigby (approved):
Actions (on PRs) pass, and all actions lgtm! Our plan is to test the build/upload by both merging this PR into main and then cutting a dev tag to test the process for canary.

@sabre1041 I made one more code suggestion (wrapping up my comments above), as a minor speed/simplification improvement. Feel free to merge that suggestion, but either way adding my approval here now. Let's unblock the things 💖

README.md
```diff
@@ -1,6 +1,6 @@
 # Helm
 
-[![CircleCI](https://circleci.com/gh/helm/helm.svg?style=shield)](https://circleci.com/gh/helm/helm)
+[![Build Status](https://github.com/helm/helm/workflows/push/badge.svg?branch=main)](https://github.com/helm/helm/actions?workflow=build-pr)
```
scottrigby, line 3: I think we want the release badge for the Readme, right? According to GH UI, I believe this is the right syntax (if no branch query is specified, the default branch - main - is assumed)

```suggestion
[![Build Status](https://github.com/helm/helm/workflows/release.yml/badge.svg)](https://github.com/helm/helm/actions/workflows/release.yml)
```

scottrigby (commented):
Also this one. Please check using another workflow name (as this wf does not yet exist to test)

scottrigby:
Excellent! I think we should wait for one other maintainer to approve since you and I both have commits on this

sabre1041:
@mattfarina any chance you might be able to review once again?

mattfarina (changes requested):

mattfarina (approved):

scottrigby:
🎉 🙌

PR #11749, opened 2023-01-19
Helm dependency build --debug should return output even if there are no dependencies

### LINKED ISSUE #11686: Helm dependency build --debug should return output even if there are no dependencies
Output of `helm version`: version.BuildInfo{Version:"v3.10.2", GitCommit:"50f003e5ee8704ec937a756c646870227d7c8b58", GitTreeState:"clean", GoVersion:"go1.19.3"}

If there are no deps `dependency build` returns no output which I think is unhelpful. I think it should at least return "no dependencies found" or something of the like.

Regardless of the default behavior `--debug` should _definitely_ output something. Its weird to have no output even with debug logging on. Made me think I had an issue.

VaibhavSharma-47:
closes #11686 The aim of this MR is that helm dependency build --debug should return something  even if there are no dependencies.

**If applicable**:

### DIFF
--- a/pkg/downloader/manager.go
+++ b/pkg/downloader/manager.go
@@ -85,6 +85,10 @@ type Manager struct {
 // If SkipUpdate is set, this will not update the repository.
 func (m *Manager) Build() error {
 	c, err := m.loadChartDir()
+	if c.Metadata.Dependencies == nil && m.Debug {
+		fmt.Print("No dependencies found to build")
+		return nil
+	}
 	if err != nil {
 		return err
 	}
--- a/pkg/downloader/manager.go
+++ b/pkg/downloader/manager.go
@@ -85,14 +85,14 @@ type Manager struct {
 // If SkipUpdate is set, this will not update the repository.
 func (m *Manager) Build() error {
 	c, err := m.loadChartDir()
-	if c.Metadata.Dependencies == nil && m.Debug {
-		fmt.Print("No dependencies found to build")
-		return nil
-	}
 	if err != nil {
 		return err
 	}
 
+	if c.Metadata.Dependencies == nil && m.Debug {
+		fmt.Print("No dependencies found to build")
+		return nil
+	}
 	// If a lock file is found, run a build from that. Otherwise, just do
 	// an update.
 	lock := c.Lock
--- a/pkg/downloader/manager.go
+++ b/pkg/downloader/manager.go
@@ -90,7 +90,7 @@ func (m *Manager) Build() error {
 	}
 
 	if c.Metadata.Dependencies == nil && m.Debug {
-		fmt.Print("No dependencies found to build")
+		fmt.Fprintln(m.Out, "No dependencies found to build")
 		return nil
 	}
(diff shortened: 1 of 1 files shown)


pkg/downloader/manager.go
```diff
@@ -85,6 +85,10 @@ type Manager struct {
 // If SkipUpdate is set, this will not update the repository.
 func (m *Manager) Build() error {
 	c, err := m.loadChartDir()
+	if c.Metadata.Dependencies == nil && m.Debug {
+		fmt.Print("No dependencies found to build")
+		return nil
```
gjenkins8, line 90: Should the early `return` here be done? It isn't clear to me the logging means code should also early return.

pkg/downloader/manager.go
```diff
@@ -85,6 +85,10 @@ type Manager struct {
 // If SkipUpdate is set, this will not update the repository.
 func (m *Manager) Build() error {
 	c, err := m.loadChartDir()
+	if c.Metadata.Dependencies == nil && m.Debug {
```
gjenkins8, line 88: I think this should be done around line 101 (after `req := c.Metadata.Dependencies`). And certainly after the error check.

Right now, the change here would ignore `err` and probably simply print `"No dependencies found to build"`, which would be confusing at best.

gjenkins8 (changes requested):
Thanks for the PR. Left a couple of comments. I think some changes are needed.

pkg/downloader/manager.go
VaibhavSharma-47: My assumption was that if there are no dependencies to build we can simply return. Is there something we do even if there are no dependencies? I didn't see a chart.lock file being created.

pkg/downloader/manager.go
VaibhavSharma-47: I have moved to after  the error check. But if I move it to after req := `c.Metadata.Dependencies`. 
```
	lock := c.Lock
	if lock == nil {
		return m.Update()
	}
```
We may end up returning here and still not print anything

pkg/downloader/manager.go
```diff
@@ -89,6 +89,10 @@ func (m *Manager) Build() error {
 		return err
 	}
 
+	if c.Metadata.Dependencies == nil && m.Debug {
+		fmt.Print("No dependencies found to build")
```
mattfarina, line 93: This call is inside the Helm SDK and is used other applications. We can't assume it's a CLI here.

In these cases we use...

```go
fmt.Fprintln(m.Out, "No dependencies found to build")
```

This will send the output to the right location whether it's the CLI or some other location an SDK user needs it.

mattfarina (changes requested):
One little thing needs to be resolved around how the message is written out.

joejulian (dismissed):

joejulian:
If there are no dependencies but there is a lock file, do we need a warning? Should we fail and suggest running update?

scottrigby:
Needs rebase. Note Helm 3 is not accepting new features, so this now goes into Helm 4.

Issue #3617, opened 2018-03-06
Extra newline being added when using "range .Files.Lines <filename>"


bobclarke:
Hello,

I have the following template:
```
kind: ClusterRoleBinding
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: cluster-admin-binding
subjects:
  {{- range .Files.Lines "admins.data" }}
- kind: User
  name: {{ . }}
  apiGroup: rbac.authorization.k8s.io
  {{- end }}
roleRef:
  kind: ClusterRole
  name: cluster-admin
  apiGroup: rbac.authorization.k8s.io
```
This is iterating over the file **admins.data**  which has the contents:
```
Alice
Bob
```

When helm install (dry-run) is executed the following output is produced:
```
---
# Source: devops-k8s-access-control/templates/cluster-admin-role-binding.yaml
kind: ClusterRoleBinding
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: cluster-admin-binding
subjects:
- kind: User
  name: Alice
  apiGroup: rbac.authorization.k8s.io
- kind: User
  name: Bob
  apiGroup: rbac.authorization.k8s.io
- kind: User
  name:
  apiGroup: rbac.authorization.k8s.io
roleRef:
  kind: ClusterRole
  name: cluster-admin
  apiGroup: rbac.authorization.k8s.io
```

As can be seen an extra iteration is taking place which is resulting in a blank name entry as follows:
```
- kind: User
  name:
  apiGroup: rbac.authorization.k8s.io
```
This causes an exception when I try to apply to my cluster because of the blank name field.

I have tried pretty much eveything I can think of and have checked the file with cat -v and vi (:set list), I can see nothing wrong. 

Can you tell me how to step this extra iteration for occurring?

Thanks,
Bob

bacongobbler:
Closing due to inactivity. Please open a new ticket against Helm 3 if this issue persists. Thanks!

PR #11758, opened 2023-01-20
strip trailing newline from Files.Lines

### LINKED ISSUE #10561: Extra newline being added when using "range .Files.Lines <filename>"
Exactly as in #3617. Templates generated with `range .Files.Lines <filename>` pretend like there is an empty line at the end of the file, but there isn't. For example:

paths.list:
```
/api/v1
/api/v2
```
with template
```
{{- range .Files.Lines "paths.list" }}
- path: {{ . }}
  pathType: ImplementationSpecific
{{- end }}
```
produces
```
- path: /api/v1
  pathType: ImplementationSpecific
- path: /api/v2
  pathType: ImplementationSpecific
- path: 
  pathType: ImplementationSpecific
```

Output of `helm version`: `version.BuildInfo{Version:"v3.7.2", GitCommit:"663a896f4a815053445eec4153677ddc24a0a361", GitTreeState:"clean", GoVersion:"go1.16.10"}`

joejulian:
A trailing newline in a file would produce an extra empty list entry when using Files.Lines

**If applicable**:

Fixes #10561

### DIFF
--- a/pkg/engine/files.go
+++ b/pkg/engine/files.go
@@ -131,7 +131,7 @@ func (f files) AsConfig() string {
 //
 //	data:
 //
-// {{ .Files.Glob("secrets/*").AsSecrets() }}
+// {{ .Files.Glob("secrets/*").AsSecrets() | indent 4 }}
 func (f files) AsSecrets() string {
 	if f == nil {
 		return ""
@@ -157,6 +157,9 @@ func (f files) Lines(path string) []string {
 	if f == nil || f[path] == nil {
 		return []string{}
 	}
-
-	return strings.Split(string(f[path]), "\n")
+	s := string(f[path])
+	if s[len(s)-1] == '\n' {
+		s = s[:len(s)-1]
+	}
+	return strings.Split(s, "\n")
 }
--- a/pkg/engine/files_test.go
+++ b/pkg/engine/files_test.go
@@ -28,7 +28,8 @@ var cases = []struct {
 	{"ship/stowaway.txt", "Legatt"},
 	{"story/name.txt", "The Secret Sharer"},
 	{"story/author.txt", "Joseph Conrad"},
-	{"multiline/test.txt", "bar\nfoo"},
+	{"multiline/test.txt", "bar\nfoo\n"},
+	{"multiline/test_with_blank_lines.txt", "bar\nfoo\n\n\n"},
 }
 
 func getTestFiles() files {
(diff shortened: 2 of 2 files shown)


jdolitsky (approved):

SoulPancake (approved):

lucasrod16 (approved):
Was able to reproduce the bug described in https://github.com/helm/helm/issues/10561 and https://github.com/helm/helm/issues/3617. This PR trims the newline away and looks good to me.

pkg/engine/files.go
```diff
@@ -156,5 +156,5 @@ func (f files) Lines(path string) []string {
 		return []string{}
 	}
 
-	return strings.Split(string(f[path]), "\n")
+	return strings.Split(strings.TrimRight(string(f[path]), "\n"), "\n")
```
gjenkins8, line 159: Apologies for nitpicking, but this will remove all empty lines at the end. e.g.

```
foo\n
bar\n
\n
\n
\n
```

Would now result in `["foo", "bar"]`. Previously, this would have been `["foo", "bar", "", "", "", ""]`.

OTOH, I agree with the setiment. A user with a file:
```
foo\n
bar\n
```

Almost certainly wants `["foo", "bar"]`.

A more correct, and perhaps simple fix. Where the user is in control, might be:
```
{{- range .Files.Lines "paths.list" | without "" }}
- path: {{ . }}
  pathType: ImplementationSpecific
{{- end }}
```

I'm not necessary against this PR. I do think we should consider not stripping multiple new lines. But also as above, letting the user choose also might be best.

gjenkins8, line 159: On another note, does this work on Windows? (I didn't read any details about how we load the files here)

joejulian: If it doesn't, it never has. CRLF files would leave the CR at the end. Changing that behavior is beyond the scope of this PR.

sabre1041:
@joejulian thanks for the PR. After looking at the various options, there is a loss of control for the end user by implementing this change. There may be situations where they would like there to be a empty line (or lines) at the end of the file.

However, if the presence of a new line is not desired, they could look to implement such suggestion shared by @gjenkins8 where the use of the `without` sprig filter would remove any empty strings from the list of lines. 

Additional controls for further refinement could also be used, but ultimately, it is given to the end user

pkg/engine/files.go
gjenkins8: > I think even in your example, the expectation would be to have only three blank entries, not four. There are only three blank lines.

heh, yes. Off by one error from me.

joejulian: I have made this change.

joejulian:
@sabre1041 @gjenkins8 Are either of you strongly opposed to this PR? It only removes the one trailing newline so if extra empty items are desired, each empty line will provide it. This now just deletes the unexpected empty entry.

PR #11763, opened 2023-01-23
[FIX]: adds warning message to default for k8s version

### LINKED ISSUE #11751: helm template: "Error: chart requires kubeVersion: ${version} which is incompatible with Kubernetes v1.18.0
Output of `helm version`:

```
» helm version
version.BuildInfo{Version:"v3.4.1", GitCommit:"c4e74854886b2efe3321e185578e6db9be0a6e29", GitTreeState:"dirty", GoVersion:"go1.15.4"}
```

Output of `kubectl version`: (N/A)

<details>
<summary>It doesn't matter for the context of this bug.</summary>

```
» kubectl version
WARNING: This version information is deprecated and will be replaced with the output from kubectl version --short.  Use --output=yaml|json to get the full version.
Client Version: version.Info{Major:"1", Minor:"25", GitVersion:"v1.25.2", GitCommit:"5835544ca568b757a8ecae5c153f317e5736700e", GitTreeState:"clean", BuildDate:"2022-09-21T14:33:49Z", GoVersion:"go1.19.1", Compiler:"gc", Platform:"darwin/amd64"}
Kustomize Version: v4.5.7
Server Version: version.Info{Major:"1", Minor:"25", GitVersion:"v1.25.2", GitCommit:"5835544ca568b757a8ecae5c153f317e5736700e", GitTreeState:"clean", BuildDate:"2022-09-21T14:27:13Z", GoVersion:"go1.19.1", Compiler:"gc", Platform:"linux/amd64"}
```

</details>

-----

We run `helm template` in our CI pipeline to see if charts have a shot at rendering prior to production. Today, we encountered this error from `helm template`:

```
Error: chart requires kubeVersion: >=1.19.0-0 which is incompatible with Kubernetes v1.18.0
```

This diagnostic leaves the user bewildered and confused: yes, it is true that the chart requires the version stated. And yes, that's incompatible with Kubernetes v1.18.0.

… _but where is "Kubernetes v1.18.0" coming from?_

Note that unlike a `helm install`, where there is a cluster in play, there _isn't_ a cluster during a `helm template`. I don't even _have_ a cluster at that version.

AFAICT, `helm template` internally, literally just [pulls that version out of thin air](https://github.com/helm/helm/blob/main/pkg/chartutil/capabilities.go#L34-L35), and uses it.

1. I don't think `helm template` should make utterly random assumptions like this. This is extremely surprising behavior.
2. The [docs page](https://helm.sh/docs/helm/helm_template/) doesn't mention this.

At the very least, as long as `helm template` wishes to behave this way, the diagnostic should inform the user that, oh, by the way, I just chose that number at random, and if you want to change that, you should specify it manually with `--kube-version`. (Worse, _that flag doesn't exist in this version_, so I can't actually use that.)

But these are the types of error messages that waste hours of devs' time: the behavior is baffling, the error is useless, and the error is un-Googleable (the results are all people hitting it rightfully during a `helm install`…).

Yes, this is a somewhat old version: I'll try to see if the error message has changed in later versions; AFAICT the random-k8s-version thing still exists in the latest `helm`. (We've been unable to upgrade as newer `helm` versions were too buggy to successfully move to…)

mritunjaysharma394:
Signed-off-by: Mritunjay Sharma <mritunjaysharma394@gmail.com>

This PR aims to fix #11751 by supporting the error message with warning message to avoid confusion and better UX

**If applicable**:

### DIFF
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -113,6 +113,7 @@ func (cfg *Configuration) renderResources(ch *chart.Chart, values chartutil.Valu
 
 	if ch.Metadata.KubeVersion != "" {
 		if !chartutil.IsCompatibleRange(ch.Metadata.KubeVersion, caps.KubeVersion.String()) {
+			fmt.Fprintf(b, "WARNING: no --kube-version parameter passed, defaulting to %s\n", caps.KubeVersion.String())
 			return hs, b, "", errors.Errorf("chart requires kubeVersion: %s which is incompatible with Kubernetes %s", ch.Metadata.KubeVersion, caps.KubeVersion.String())
 		}
 	}


pkg/action/action.go
```diff
@@ -113,6 +113,7 @@ func (cfg *Configuration) renderResources(ch *chart.Chart, values chartutil.Valu
 
 	if ch.Metadata.KubeVersion != "" {
 		if !chartutil.IsCompatibleRange(ch.Metadata.KubeVersion, caps.KubeVersion.String()) {
+			fmt.Fprintf(b, "WARNING: no --kube-version parameter passed, defaulting to %s\n", caps.KubeVersion.String())
```
gjenkins8, line 116: We can't print directly to stdout here I think. This package is used both by the helm cli and directly as an API.

The Action's Configuration's Log functionality should be used: https://github.com/helm/helm/blob/main/pkg/action/action.go#L97

gjenkins8 (changes requested):
Thanks for the PR! A small adjustment please.

pkg/action/action.go
mritunjaysharma394: Got it, thanks a lot for the review and help, will definitely try to fix it soon!

mritunjaysharma394:
Sure @gjenkins8, I think as much as I understood, I have to use first Action's Configuration's Log functionality and then see if it gets printed to stderr or should I check with the existing code change? Also, this `helm template | kubectl apply --dry-run=server -f -
` command should be enough to test it right?

pkg/action/action.go
joejulian: I think this should also only be output with the debug flag.

joejulian:
@mritunjaysharma394 Are you still wanting to work on this PR?

joejulian:
closing this as abandoned

PR #11777, opened 2023-01-31
Ensure `helm test --logs` only gets logs from pods

### LINKED ISSUE #11236: `helm test` using `Jobs` cannot show logs
We're unable to get the logs of pods when using `helm test` with [Kubernetes Jobs](https://kubernetes.io/docs/concepts/workloads/controllers/job/). Looks like documentation is only describing the use of Pods.

This is a continuation to #9776

`example-chart/templates/tests/job.yaml`

```yaml
apiVersion: batch/v1
kind: Job
metadata:
  name: example
  annotations:
    helm.sh/hook: test
spec:
  backoffLimit: 1
  parallelism: 1
  completions: 1
  template:
    spec:
      containers:
        - name: example
          image: {{ .Values.image | quote }}
          resources:
            limits:
              cpu: 500m
              memory: 1Gi
            requests:
              cpu: 500m
              memory: 1Gi
      restartPolicy: Never
```

Running the test command.

```bash
$ helm test --logs example

Error: unable to get pod logs for example: pods "example" not found
```

It looks like it's trying to get the pod name, but since there's a Kubernetes Job in charge of producing the pod, there's a unique ID at the end of the pod name e.g. `example-6qh95` instead of `example`.

Output of `helm version`:

version.BuildInfo{Version:"v3.9.0", GitCommit:"7ceeda6c585217a19a1131663d8cd1f7d641b2a7", GitTreeState:"clean", GoVersion:"go1.18.2"}

Output of `kubectl version`:

```
Client Version: version.Info{Major:"1", Minor:"24", GitVersion:"v1.24.1", GitCommit:"3ddd0f45aa91e2f30c70734b175631bec5b5825a", GitTreeState:"clean", BuildDate:"2022-05-24T12:17:11Z", GoVersion:"go1.18.2", Compiler:"gc", Platform:"darwin/arm64"}

Server Version: version.Info{Major:"1", Minor:"21+", GitVersion:"v1.21.13-eks-84b4fe6", GitCommit:"e1318dce57b3e319a2e3fecf343677d1c4d4aa75", GitTreeState:"clean", BuildDate:"2022-06-09T18:22:07Z", GoVersion:"go1.16.15", Compiler:"gc", Platform:"linux/amd64"}
```

EKS

akshaymankar:

Partially fixes https://github.com/helm/helm/issues/11236#issuecomment-1332554000 by ensuring that `helm test --logs` only tries to get logs of a hook of kind `Pod`. 
Of course, a more comprehensive fix would be nice but perhaps someone is already working on it as the issue is labelled `in-progress`.

**If applicable**:

### DIFF
--- a/pkg/action/release_testing.go
+++ b/pkg/action/release_testing.go
@@ -108,19 +108,21 @@ func (r *ReleaseTesting) GetPodLogs(out io.Writer, rel *release.Release) error {
 	}
 
 	for _, h := range rel.Hooks {
-		for _, e := range h.Events {
-			if e == release.HookTest {
-				req := client.CoreV1().Pods(r.Namespace).GetLogs(h.Name, &v1.PodLogOptions{})
-				logReader, err := req.Stream(context.Background())
-				if err != nil {
-					return errors.Wrapf(err, "unable to get pod logs for %s", h.Name)
-				}
+		if h.Kind == "Pod" {
+			for _, e := range h.Events {
+				if e == release.HookTest {
+					req := client.CoreV1().Pods(r.Namespace).GetLogs(h.Name, &v1.PodLogOptions{})
+					logReader, err := req.Stream(context.Background())
+					if err != nil {
+						return errors.Wrapf(err, "unable to get pod logs for %s", h.Name)
+					}
 
-				fmt.Fprintf(out, "POD LOGS: %s\n", h.Name)
-				_, err = io.Copy(out, logReader)
-				fmt.Fprintln(out)
-				if err != nil {
-					return errors.Wrapf(err, "unable to write pod logs for %s", h.Name)
+					fmt.Fprintf(out, "POD LOGS: %s\n", h.Name)
+					_, err = io.Copy(out, logReader)
+					fmt.Fprintln(out)
+					if err != nil {
+						return errors.Wrapf(err, "unable to write pod logs for %s", h.Name)
+					}
 				}
 			}
 		}


MichaelMorrisEst:
The PR only seems to prevent the error being thrown rather than outputting the logs from the pods. Perhaps an improvement might be, in the case of a job, to fetch the pods associated with the job and then get the logs of those pods

I have found the following to work to get the pods from the jobs, but perhaps there is a better way (Im relatievly new to helm so others may have better ideas)
					ctx := context.Background()
					job, _ := client.BatchV1().Jobs(r.Namespace).Get(ctx, h.Name, metav1.GetOptions{})
					listOptions := metav1.ListOptions{
						LabelSelector: "job-name=" + job.Name,
					}

					pods, _ := client.CoreV1().Pods("").List(ctx, listOptions)

pkg/action/release_testing.go
```diff
@@ -108,19 +108,21 @@ func (r *ReleaseTesting) GetPodLogs(out io.Writer, rel *release.Release) error {
...
-				if err != nil {
-					return errors.Wrapf(err, "unable to get pod logs for %s", h.Name)
-				}
-
-				fmt.Fprintf(out, "POD LOGS: %s\n", h.Name)
-				_, err = io.Copy(out, logReader)
-				fmt.Fprintln(out)
-				if err != nil {
-					return errors.Wrapf(err, "unable to write pod logs for %s", h.Name)
+		if h.Kind == "Pod" {
```
joejulian, line 111: shortcut the logic here and outdent the remainder of the for loop
```suggestion
		if h.Kind != "Pod" {
			continue
		}
```

joejulian:
It does seem like we should be getting logs from any of the application types, but that's out of scope for this PR. Since we're only getting logs from pods, it seems reasonable to gate the fetch on it being an actual pod.

@akshaymankar if you could shortcut the for loop as suggested and rebase this PR, it's good to merge.

akshaymankar:
Sorry, I don't need this anymore. Please feel free to create a separate PR with the branch in this PR.

PR #11780, opened 2023-02-01
fix(create): add podLabels to chart create

genofire:


joejulian (approved):
lgtm

pkg/chartutil/create.go
```diff
@@ -308,6 +309,9 @@ spec:
       {{- end }}
       labels:
         {{- include "<CHARTNAME>.selectorLabels" . | nindent 8 }}
+	{{- with .Values.podLabels }}
```
l2dy, line 312: Line 312 is tab-indented, while others are space-indented. Should they be consistent?

genofire: oh sorry

genofire: https://github.com/helm/helm/pull/12457

PR #11788, opened 2023-02-03
wait for release before post hooks

### LINKED ISSUE #11778: post-install/post-upgrade hooks work have inconsistent behavior based on the --wait flag
If a post-install or post-upgrade hook need to operate on a running deployment/statefulset/etc, those resources will only be available if helm is run with the `--wait` flag. If it's not, the hook is run before those resources are ready.

This is inconsistent and leads to surprises.

Relevant code: https://github.com/helm/helm/blob/76157c6d06f981a42bfbb9ec636bae5443cb2c8f/pkg/action/install.go#L382-L402

I think this should be changed such that if there are any post-* hooks, wait is automatically enabled at line 382.

Output of `helm version`:
version.BuildInfo{Version:"v3.11.0", GitCommit:"472c5736ab01133de504a826bd9ee12cbe4e7904", GitTreeState:"clean", GoVersion:"go1.19.5"}

Output of `kubectl version`:
N/A

N/A

joejulian:

If a release has post-install, post-delete, or post-rollback hooks, always wait for the main resources before doing the post - otherwise it's not really "post". This ensures that the behavior is correct if a chart user doesn't apply the `--wait` flag when a chart has "post" operations.

This was seen in the wild when one of our (Redpanda) chart users installed without the `--wait` flag and the post-install configuration failed because the statefulset wasn't ready to accept connections from the post-install,post-upgrade job. We never caught this in CI because ct always sets the `--wait` flag.

Fixes #11778

https://github.com/helm/chart-testing always tests with the `--wait` flag so this is a no-op for charts that are tested. In fact, this ensures the user has the same experience as the test if the user fails to set wait.

**If applicable**:

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -432,7 +432,9 @@ func (i *Install) performInstall(c chan<- resultMessage, rel *release.Release, t
 		}
 	}
 
-	if i.Wait {
+	waitBeforePostHooks := !i.DisableHooks && i.cfg.hasPostInstallHooks(rel) || i.cfg.hasPostUpgradeHooks(rel)
+
+	if i.Wait || waitBeforePostHooks {
 		if i.WaitForJobs {
 			if err := i.cfg.KubeClient.WaitWithJobs(resources, i.Timeout); err != nil {
 				i.reportToRun(c, rel, err)
--- a/pkg/action/hooks.go
+++ b/pkg/action/hooks.go
@@ -38,7 +38,7 @@ func (cfg *Configuration) execHook(rl *release.Release, hook release.HookEvent, 
 		}
 	}
 
-	// hooke are pre-ordered by kind, so keep order stable
+	// hooks are pre-ordered by kind, so keep order stable
 	sort.Stable(hookByWeight(executingHooks))
 
 	for _, h := range executingHooks {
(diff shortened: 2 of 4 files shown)


phil9909 (approved):

phil9909:
While technically this can be breaking (let's say a Chart contains a `Deployment` and a `Job`, the `Job` is marked as `post-install` and the `Deployment` mounts e.g. some secret generated by the `Job`) I'm still in favor of this change.

It only breaks Charts that were already more-or-less broken, as they can't be installed with `--wait` enabled either.

pkg/action/install.go
```diff
@@ -379,7 +379,9 @@ func (i *Install) performInstall(c chan<- resultMessage, rel *release.Release, t
 		}
 	}
 
-	if i.Wait {
+	waitBeforePostHooks := !i.DisableHooks && i.cfg.hasPostInstallHooks(rel) || i.cfg.hasPostUpgradeHooks(rel)
```
gjenkins8, line 382: Do we need to consider an option to allow `post-install|upgrade` (and `rollback` below) hooks run before waiting? To avoid a breaking users who might depend on this behavior here?

joejulian: I don't see how this could break users. The behavior was always inconsistent. If they test their charts, they would have discovered the failure.

All the CD systems either use wait or they template everything and use their own applicator that applies the post-* jobs after the main body is ready.

I don't think polluting the cli with yet-another-flag has any value here.

scottrigby:
@joejulian could we add a unit test for this?

scottrigby (changes requested):
@joejulian this looks great! 👏 

one note—we need to add the Helm license header to `pkg/action/hooks_test.go` https://github.com/helm/helm/actions/runs/12820083645/job/35749023847?pr=11788

scottrigby (approved):
License headers have been fixed. This now just needs a rebase. Moving to next patch release and approving in advance of rebase.

joejulian:
Closing this as superseded. Current `main` already contains the core behavior this PR proposed: Helm now waits before post-install and post-rollback hooks when relevant post hooks exist, so this old branch is no longer the path to land that change.

PR #11801, opened 2023-02-08
Update to func handling

crenshaw-dev:
Argo CD v2.5.x uses Helm v3.10.3. I'd like to avoid bumping to 3.11.x in 2.5.x. Can we cherry-pick the fix to release-3.10 and get a 3.10.4?


mattfarina:
@crenshaw-dev We don't normally do releases like that. Is there a specific reason for this release? Or a pattern you're looking for?

crenshaw-dev:
Argo CD users are vulnerable to CVE-2023-25165.

To keep minor release series relatively stable, we try to avoid bumping Helm minor versions in an Argo CD patch release.

A 3.10.4 release with the CVE fix would let us resolve the vulnerability for our users without doing a potentially-disruptive minor version Helm upgrade in an Argo CD patch release.

PR #9130, opened 2020-12-15
Add --hide-secrets flag

Szymongib:
Fixes: https://github.com/helm/helm/issues/7275

**What this PR does / why we need it**:
The PR contains my proposition for resolving https://github.com/helm/helm/issues/7275 issue. Instead of implementing the new flag just for the `--dry-run` option I decided to go with more general approach applicable for other actions like `install`.

The PR introduces new flag `--hide-secrets` which replaces all values in Secrets data fields with `[HIDDEN]` text.
For example:
Secret:
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: secret-sample
data:
  test: YmFyCg==
  password: bXktcGFzc3dvcmQ=
```
would be displayed as:
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: secret-sample
data:
  test: '[HIDDEN]'
  password: '[HIDDEN]'
```

The replacing is done after applying manifest and before outputting it.

**Special notes for your reviewer**:
I am curious to hear your thoughts/criticism/other ideas.

**If applicable**:
- [ ] this PR contains documentation
- [x] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

Issue #7275, opened 2019-12-19
helm 3 shows secrets with --dry-run option in clear text


darkstarmv:
Output of `helm version`:

`helm version
version.BuildInfo{Version:"v3.0.2", GitCommit:"19e47ee3283ae98139d98460de796c1be1e3975f", GitTreeState:"clean", GoVersion:"go1.13.5"}`
Output of `kubectl version`:

`kubectl version
Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.3", GitCommit:"2d3c76f9091b6bec110a5e63777c332469e0cba2", GitTreeState:"clean", BuildDate:"2019-08-19T11:13:54Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"darwin/amd64"}`

Cloud Provider/Platform (AKS, GKE, Minikube etc.): 
### Issue

`helm3`  `--dry-run` command prints content of the secrets where helm2 just the fact that secret has been created. 
**This creates issue running helm `--dry-run` in CI/CD tools as it exposes secrets.**

```
 helm upgrade --install myrelease  mychart  --set thanos.sayamlbase64=TESTSECRET  --dry-run
Release "myrelease" has been upgraded. Happy Helming!
NAME: myrelease
LAST DEPLOYED: Thu Dec 19 15:11:43 2019
NAMESPACE: kube-system
STATUS: pending-upgrade
REVISION: 400
TEST SUITE: None
HOOKS:
MANIFEST:
---
# Source: mychart/templates/thanos-objstore-config-secret.yaml
apiVersion: v1
data:
  thanos.yaml: TESTSECRET
kind: Secret
metadata:
  name: thanos-objstore-config
  namespace: kube-system
type: Opaque
---
```

```
helm2 upgrade --install release-name chart  --set thanos.sayamlbase64=TESTSECRET  --dry-run
Release "release-name" has been upgraded. Happy Helming!
LAST DEPLOYED: Thu Dec 19 14:51:59 2019
NAMESPACE: kube-system
STATUS: DEPLOYED

RESOURCES:
==> v1/Secret
NAME                    TYPE    DATA  AGE
thanos-objstore-config  Opaque  1     80d
 ```

bacongobbler:
Hi @darkstarmv. `helm install --dry-run` intentionally shows the same output as `helm install`. It even looks as though this is intentional behaviour:

[...]

PR #11802, opened 2023-02-09
Add --hide-values flag

codersasha:

There is a need to run `helm install`/`helm upgrade` commands in Debug mode without leaking secrets (see #7275). This PR fixes that in a more general way by adding a `--hide-values` flag, which, when combined with `--debug`, prints relevant debugging information (such as why a helm upgrade command failed) but does not display the values or manifest file, removing the risk of leaking secret config and sensitive manifest contents.

This PR was partially inspired by the feedback on #9130.

I haven't signed my commit, as per the contribution guidelines -- still figuring out how to do that 😅 -- but very happy for someone to take over this PR and rewrite it as they see fit. This is a very important PR for our organisation and, without it, we have no way of easily debugging our `helm install` deploy commands on CI. We cannot run with `--debug` and risk leaking secrets and other sensitive manifest values, so we need a flag like this to skip displaying the values entirely.

**If applicable**:


sashamor:
Yeah, that makes sense. I had similar concerns when writing this PR.

What do you think of, `--debug=VERBOSE` (the default `--debug`) and `--debug=QUIET` (the equivalent of `--hide-values`).

sashamor:
Closing as we found another solution - this is a `sed` command that deletes everything from `USER-SUPPLIED VALUES:` down:

```bash
helm install --debug ... | sed '/USER-SUPPLIED VALUES:/,$d'
```

We now use this in our deploys, so we can debug when a deploy fails, but not print any secrets/configuration that may reveal sensitive data.

wirwolf:
But its сrutches...

wirwolf:
@codersasha But its сrutches...

PR #11806, opened 2023-02-09
don't mark issues as stale where a PR is in progress

joejulian:

PR reviews take much longer than 90 days. Let's not nag issues that are labeled "in progress"


marckhouzam:
I like the idea but what's the workflow difference between the "in progress" label compared to "keep open".

joejulian:
It's mostly for convenience when I'm doing triage and support. There are things that you folks mark "keep open" as things you've found important for the project. I mark things "in progress" when there's a PR open so I can manage the help-wanted/good-first-issue labels as appropriate.

marckhouzam (approved):
Sounds good to me.

hickeyma (approved):
Nice idea @joejulian

PR #11833, opened 2023-02-19
feature: extra check for linting, chart name

### LINKED ISSUE #10537: helm lint misses bad chart names
`helm lint` only catches if a chart name is missing. You can see that in the following snippet...

https://github.com/helm/helm/blob/292368a64d6c9794b4573afa5a8dc6e940a21045/pkg/lint/rules/chartfile.go#L106-L111

The `chartutil` package has the ability to validate names in some forms. We could provide better hints (like no capital letters) via lint.

Output of `helm version`: 3.7

lexhuismans:
This PR adds an extra check for the linter. It checks if the chart-name only contains lower-case letters and (-). According to the [guidelines](https://helm.sh/docs/chart_best_practices/conventions/#chart-names).

It closes #10537 

**If applicable**:

### DIFF
--- a/pkg/lint/rules/chartfile.go
+++ b/pkg/lint/rules/chartfile.go
@@ -111,7 +111,7 @@ func validateChartName(cf *chart.Metadata) error {
 		return errors.New("name is required")
 	}
 	if !validNameRegex.MatchString(cf.Name) {
-		return fmt.Errorf("invalid release name, must match regex %s", validNameRegex.String())
+		return fmt.Errorf("invalid chart name, must match regex %s", validNameRegex.String())
 	}
 	return nil
 }
--- a/pkg/lint/rules/chartfile.go
+++ b/pkg/lint/rules/chartfile.go
@@ -111,7 +111,7 @@ func validateChartName(cf *chart.Metadata) error {
 		return errors.New("name is required")
 	}
 	if !validNameRegex.MatchString(cf.Name) {
-		return errors.New(fmt.Sprintf("invalid release name, must match regex %s", validNameRegex.String()))
+		return fmt.Errorf("invalid release name, must match regex %s", validNameRegex.String())
 	}
 	return nil
 }
--- a/pkg/lint/rules/chartfile.go
+++ b/pkg/lint/rules/chartfile.go
@@ -21,6 +21,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"regexp"
 
 	"github.com/Masterminds/semver/v3"
 	"github.com/asaskevich/govalidator"
@@ -104,8 +105,13 @@ func validateChartYamlFormat(chartFileError error) error {
 }
 
 func validateChartName(cf *chart.Metadata) error {
+	var validNameRegex = regexp.MustCompile(`^[a-z0-9](([-a-z0-9]*)?)$`)
+
 	if cf.Name == "" {
 		return errors.New("name is required")
+	}
+	if !validNameRegex.MatchString(cf.Name) {
+		return errors.New(fmt.Sprintf("invalid release name, must match regex %s", validNameRegex.String()))
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


pkg/lint/rules/chartfile_test.go
```diff
@@ -65,9 +67,15 @@ func TestValidateChartYamlFormat(t *testing.T) {
 }
 
 func TestValidateChartName(t *testing.T) {
-	err := validateChartName(badChart)
-	if err == nil {
-		t.Errorf("validateChartName to return a linter error, got no error")
+	badNameChart, _ := chartutil.LoadChartfile(goodChartFilePath)
```
gjenkins8, line 70: looks to be a mismatch here: `badNameChart := f(goodChartFilePath)`

pkg/lint/rules/chartfile.go
```diff
@@ -104,9 +105,14 @@ func validateChartYamlFormat(chartFileError error) error {
 }
 
 func validateChartName(cf *chart.Metadata) error {
+	var validNameRegex = regexp.MustCompile(`^[a-z0-9](([-a-z0-9]*)?)$`)
```
gjenkins8, line 108: minor: this could be a package scoped? (no need to instantiate every time the function is called)

gjenkins8, line 108: Where did we get this regex from? In part, because we need to ensure it is correct.

pkg/lint/rules/chartfile.go
```diff
@@ -104,9 +105,14 @@ func validateChartYamlFormat(chartFileError error) error {
...
+
 	if cf.Name == "" {
 		return errors.New("name is required")
 	}
+	if !validNameRegex.MatchString(cf.Name) {
+		return fmt.Errorf("invalid release name, must match regex %s", validNameRegex.String())
```
gjenkins8, line 114: Did you mean "invalid chart name" here?

```suggestion
		return fmt.Errorf("invalid chart name, must match regex %s", validNameRegex.String())
```

pkg/lint/rules/chartfile.go
```diff
@@ -104,9 +105,14 @@ func validateChartYamlFormat(chartFileError error) error {
 }
 
 func validateChartName(cf *chart.Metadata) error {
+	var validNameRegex = regexp.MustCompile(`^[a-z0-9](([-a-z0-9]*)?)$`)
+
 	if cf.Name == "" {
```
gjenkins8, line 110: strictly, this check is redundant. But I think the explicit check for an empty name is worthwhile in terms of the messages presented to the user (ie. keep this)

pkg/lint/rules/chartfile.go
```diff
@@ -104,9 +105,14 @@ func validateChartYamlFormat(chartFileError error) error {
 }
 
 func validateChartName(cf *chart.Metadata) error {
```
gjenkins8, line 107: Perhaps a comment linking to chart name requirements?

gjenkins8 (commented):
Thanks for the PR! Added some (mostly minor) comments (helm community member, not maintainer here).

From experience, helm tries very hard to ensure backwards compatibility. I think ensuring the regex is correct is the most pertinent part of this change (ie.the regex won't reject otherwise valid chart names)

pkg/lint/rules/chartfile_test.go
lexhuismans: Thank you for the comment. In the next lines I change the name of the good chart to one with a "bad" name. I used the good chart as a reference so that the rest is correct, only the name is "bad". Hope this makes sense. 

The test catches the bad names so is working as it should.

pkg/lint/rules/chartfile.go
lexhuismans: Only case I think it might miss is when a chart name starts with a "-". But according to the guidelines it should not start (or end) with a "-". Words can be separated by a "-".

lexhuismans:
Thanks for your comments. I replied and resolved a few. Some we might leave for judgement of the maintainers.

PR #11843, opened 2023-02-22

### LINKED ISSUE #8660: Wait logic doesn't wait for all deployment and daemonset pods to be updated/ready

felipecrs:
This PR rebases #10831, which was based on #8671, and fixes #8660.


felipecrs:
Ok, we all know this will never be merged. :D

PR #11853, opened 2023-02-24
Addressing multi agent test race condition

Freydal:
This addresses the helm test command overwriting any change in state to a release during its execution. While I'm not certain which behavior is more desired, when other actors execute a helm test while the installation is pending an install with --atomic of --wait a data race occurs on the status.  If the test execution begins while the release is in pending-upgrade and install/upgrade completes setting the status to deployed  before test completes the end status will be pending-upgrade.

fluxcd/flagger#816

While attempting to add unit tests to validate my data race I uncovered a data race detected that failed the tests.  I included a commit to address this.  This work was done a few months ago, but sat on it due to uncertainty of it was a worthwhile contribution.

**If applicable**:

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -422,7 +422,10 @@ func (i *Install) performInstall(c chan<- resultMessage, rel *release.Release, t
 	}
 
 	if !i.DisableHooks {
-		if err := i.cfg.execHook(rel, release.HookPostInstall, i.Timeout); err != nil {
+		i.Lock.Lock()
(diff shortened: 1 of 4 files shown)


Freydal:
Correct.  If you have some CICD agent deploying a change and something like Flagger(which supports Helm webhooks) in cluster that  triggers a test prior to the release agent completing a task you may see the failure in this described scenario.

Freydal:
Rebased to address the circleci config failure.

Freydal:
Rebased again to keep up to date.

PR #11878, opened 2023-03-06
feat: pass basic auth to env-vars when running download plugins

### LINKED ISSUE #11877: feat: support passing credentials to downloader plugins
Currently `helm` [supports passing `--username` and `--password`](https://github.com/helm/helm/blob/e007c900ce5f2233c8583565d5ba1b0433b1a213/pkg/getter/httpgetter.go#L76-L78) for its built-in HTTP getter. This allows us to add private HTTP repos which require basic username/password authentication.  
The same is not true for downloader plugins - the [`Get` function](https://github.com/helm/helm/blob/e007c900ce5f2233c8583565d5ba1b0433b1a213/pkg/getter/plugingetter.go#L66-L86) which executes downloader plugins never writes the username/password to env variables, so they aren't available to the downloader plugin.

This causes issues when you want to e.g. download on the git protocol. Although this is supported by [the `helm-git` plugin](https://github.com/aslafy-z/helm-git), the plugin is never passed the username/password given to helm, and therefore can't authenticate with them even if it implemented support for basic auth.

Exchizz:
This PR passes username and password from "helm repo add --username X --password Y" to download plugins. See #11877 for motivation.

**If applicable**:

closes #11877

I created this PR with help from @birjj

### DIFF
--- a/pkg/getter/plugingetter.go
+++ b/pkg/getter/plugingetter.go
@@ -66,6 +66,7 @@ type pluginGetter struct {
 func (p *pluginGetter) setupOptionsEnv(env []string) []string {
 	env = append(env, fmt.Sprintf("HELM_PLUGIN_USERNAME=%s", p.opts.username))
 	env = append(env, fmt.Sprintf("HELM_PLUGIN_PASSWORD=%s", p.opts.password))
+	env = append(env, fmt.Sprintf("HELM_PLUGIN_PASS_CREDENTIALS_ALL=%t", p.opts.passCredentialsAll))
 	return env
 }
 
--- a/pkg/getter/plugingetter.go
+++ b/pkg/getter/plugingetter.go
@@ -17,6 +17,7 @@ package getter
 
 import (
 	"bytes"
+	"fmt"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


Exchizz:
Hi,

What's the status of this PR ? Currently an [issue ](https://github.com/aslafy-z/helm-git/issues/255) in helm-git is waiting for this this PR to be merged.

\Exchizz

pkg/getter/plugingetter.go
```diff
@@ -62,6 +63,12 @@ type pluginGetter struct {
 	opts     options
 }
 
+func (p *pluginGetter) setupOptionsEnv(env []string) []string {
+	env = append(env, fmt.Sprintf("HELM_PLUGIN_USERNAME=%s", p.opts.username))
+	env = append(env, fmt.Sprintf("HELM_PLUGIN_PASSWORD=%s", p.opts.password))
```
mattfarina, line 68: There is one more `opt` that should be passed along for this. It is `passCredentialsAll`.

This is for security. The username and password are for the specific "thing" that's being downloaded. This could be an individual chart of an index file for a chart repo. When getting the index file, the charts could be on a different domain. Do you pass the username/password on to this other domain? If so, you could leak the credentials.

Helm stores the intent if the username/password should be passed on to other domains in this option. This is due to [CVE-2021-32690](https://github.com/helm/helm/security/advisories/GHSA-56hp-xqp3-w2jf). The details of this situation can be found there.

mattfarina (changes requested):
There is one change that, I think, needs to happen. It's due to a CVE that came to Helm a couple years ago and has to do with where the username/password are used.

pkg/getter/plugingetter.go
Exchizz: I've added HELM_PLUGIN_PASS_CREDENTIALS_ALL as env-var :)

Exchizz:
Hi @mattfarina,

I can see this PR is set to be included in 3.13.0 which is set to be released in a couple of days.
Do you think this PR will make it into the next release ? 

Thanks in advance

joejulian (approved):

joejulian:
Since @mattfarina's concern was addressed, I'm going to go ahead and merge this.

PR #11881, opened 2023-03-06
Apply label selector at server side

### LINKED ISSUE #11875: Helm list releases uses high amount of memory when there are many releases deployed
We have an application which periodically lists deployed helm releases to perform certain action in a Kubernetes cluster owned by the customer. We recently discovered that some of the customers have many (300+ or even more, 6000+) versions deployed, which can cost a significant amount of memory, because every filtering is done in the client side: for secret and configmap storage, every Helm release secret/configmap is listed first, and filters area applied at the client side. Unfortunately using `--selector status=deployed` parameter didn't help, since label selector are also applied at client side, limiting the version history isn't an option either.

This can be improved if there's a possibility to do the label selector based filtering in the storage layer (at k8s side) which could improve both performance and reduce memory footprint.
I did a POC for pushing label selectors to server side and there's a significant improvement in a cluster with ~400 releases:
Helm v3.10.3:
```
/usr/bin/time -l -h -p helm list --selector status=deployed
real 16.56
user 2.37
sys 0.54
           233897984  maximum resident set size
                   0  average shared memory size
                   0  average unshared data size
                   0  average unshared stack size
               13310  page reclaims
                1380  page faults
                   0  swaps
                   0  block input operations
                   0  block output operations
                2956  messages sent
               10773  messages received
                1389  signals received
                 163  voluntary context switches
               43158  involuntary context switches
         20076385267  instructions retired
          6049319295  cycles elapsed
           215238016  peak memory footprint
```
Server side selectors:
```
/usr/bin/time -l -h -p helm list --selector status=deployed
real 2.35
user 0.16
sys 0.06
            42467328  maximum resident set size
                   0  average shared memory size
                   0  average unshared data size
                   0  average unshared stack size
                1599  page reclaims
                1407  page faults
                   0  swaps
                   0  block input operations
                   0  block output operations
                 102  messages sent
                 230  messages received
                 130  signals received
                  57  voluntary context switches
                2446  involuntary context switches
           828154262  instructions retired
           369166677  cycles elapsed
            23331904  peak memory footprint
```

Output of `helm version`:
version.BuildInfo{Version:"v3.10.3", GitCommit:"835b7334cfe2e5e27870ab3ed4135f136eecc704", GitTreeState:"clean", GoVersion:"go1.19.4"}

Output of `kubectl version`:
Client Version: version.Info{Major:"1", Minor:"25", GitVersion:"v1.25.0", GitCommit:"a866cbe2e5bbaa01cfd5e969aa3e033f3282a8a2", GitTreeState:"clean", BuildDate:"2022-08-23T17:36:43Z", GoVersion:"go1.19", Compiler:"gc", Platform:"darwin/arm64"}
Kustomize Version: v4.5.7
Server Version: version.Info{Major:"1", Minor:"24", GitVersion:"v1.24.9", GitCommit:"57fbbcc2804848b95cad5519f5ec9d6355430db9", GitTreeState:"clean", BuildDate:"2023-02-08T17:22:38Z", GoVersion:"go1.18.9", Compiler:"gc", Platform:"linux/amd64"}

nandorKollar:
This change introduces an option to push down label selectors to storage layer, if supported. When Kubernetes secrets or configmaps are used as backing storage for releases, then pushing down label selectors to API server can result in significant performance improvement and reduced memory footprint when there are hundreds or thousands of versions in the cluster. Currently this change does pushdown only for `--selector` parameter, but probably something similar would make sense to consider for other Helm commands, like list by release status. This closes #11875 

**If applicable**:

### DIFF
--- a/pkg/action/list.go
+++ b/pkg/action/list.go
@@ -159,14 +159,21 @@ func (l *List) Run() ([]*release.Release, error) {
 		}
 	}
 
-	results, err := l.cfg.Releases.List(func(rel *release.Release) bool {
+	filterFunction := func(rel *release.Release) bool {
 		// Skip anything that doesn't match the filter.
 		if filter != nil && !filter.MatchString(rel.Name) {
 			return false
 		}
 
 		return true
-	})
+	}
+	var results []*release.Release
+	var err error
+	if l.cfg.Releases.CanPushDownLabelSelector() {
+		results, err = l.cfg.Releases.ListWithSelector(filterFunction, l.Selector)
+	} else {
+		results, err = l.cfg.Releases.List(filterFunction)
+	}
 
 	if err != nil {
 		return nil, err
@@ -187,12 +194,16 @@ func (l *List) Run() ([]*release.Release, error) {
 	// latest releases, otherwise outdated entries can be returned
 	results = l.filterStateMask(results)
 
-	// Skip anything that doesn't match the selector
-	selectorObj, err := labels.Parse(l.Selector)
-	if err != nil {
-		return nil, err
+	// If storage layer can't perform label selector filtering,
+	// then fall back to client side filtering
+	if !l.cfg.Releases.CanPushDownLabelSelector() {
+		// Skip anything that doesn't match the selector
+		selectorObj, err := labels.Parse(l.Selector)
+		if err != nil {
+			return nil, err
+		}
+		results = l.filterSelector(results, selectorObj)
 	}
-	results = l.filterSelector(results, selectorObj)
 
 	// Unfortunately, we have to sort before truncating, which can incur substantial overhead
 	l.sort(results)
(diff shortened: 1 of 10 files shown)


pkg/action/list.go
```diff
@@ -159,14 +159,21 @@ func (l *List) Run() ([]*release.Release, error) {
 		}
 	}
 
-	results, err := l.cfg.Releases.List(func(rel *release.Release) bool {
+	filterFunction := func(rel *release.Release) bool {
```
gjenkins8, line 162: Now (since we are targeting Helm v4, and do not need to maintain API backwards compatibility), we can just pass the label sector to `List(...)`, and avoid the extra complexity of  `CanPushDownLabelSelector(...)` and `ListWithSelector(...)`.

ie., just do the below and fold the selector logic into `List(..)` itself (where applicable, based on the driver):
```
	results, err := l.cfg.Releases.List(l.Selector, func(rel *release.Release) bool {
```

PR #11886, opened 2023-03-07
Properly invalidate client after CRD install

hiddeco:
As the CRDs are installed before the capabilities are gathered, the current call to invalidate the discovery client is premature and expensive.

What actually is required is an invalidation of the REST mapper, as otherwise the Helm install action may later on fail with a `resource mapping not found` error. More specifically when the caller of the action is making use of a persisting[1] `RESTClientGetter`.

Which is not something done by the Helm CLI (albeit it could, and this would potentially save quite some resources?), but is a default configuration offered by the Helm SDK via `kube.New` when a nil value is provided as the `getter`.

[1]: https://github.com/kubernetes/cli-runtime/blob/v0.26.2/pkg/genericclioptions/config_flags.go#L118

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -34,6 +34,7 @@ import (
 	"github.com/pkg/errors"
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/cli-runtime/pkg/resource"
 	"sigs.k8s.io/yaml"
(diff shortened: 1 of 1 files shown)


pkg/action/install.go
```diff
@@ -159,22 +160,38 @@ func (i *Install) installCRDs(crds []chart.CRD) error {
...
+		if i.cfg.Capabilities != nil {
+			discoveryClient, err := i.cfg.RESTClientGetter.ToDiscoveryClient()
+			if err != nil {
+				return err
+			}
+
+			i.cfg.Log("Clearing discovery cache")
+			discoveryClient.Invalidate()
+
+			_, _ = discoveryClient.ServerGroups()
```
hiddeco, line 182: Quite certain invalidation itself should be sufficient, but keeping this in place given the history of #6332.

sabre1041 (approved):
LGTM tested and validated approach

hiddeco:
@sabre1041 would you be open to a follow up which further tries to see if we can remove the expensive `ServerGroups` call in full? As in a fork of the CRD logic we are using in helm-controller, this does not seem to be required at all (we do not even invalidate, but that's allowed mainly because `getCapabilities` later does this).

In addition, I could fill [d7d624d](https://github.com/hiddeco/helm/commit/d7d624d53c9646590dfbd3f19dd2f440820605ef) as a separate PR to reduce resource usage of the CLI (re-using the instantiated clients really is cheaper, and saves quite a bit of RAM).

sabre1041:
@hiddeco lets track any further enhancements in a separate PR. Given that the PR contained here does address the challenge of invalidating the client after CRD install, lets work on getting this merged and follow up with subsequent enhancements

mattfarina (approved):
Ran a number of local tests and performs as expected.

PR #11896, opened 2023-03-09
Bump the Go version

mattfarina:
Use a supported version of Go.

**If applicable**:


.github/workflows/build-test.yml
```diff
@@ -26,8 +26,8 @@ jobs:
           sudo mv golangci-lint-$GOLANGCI_LINT_VERSION-linux-amd64/golangci-lint /usr/local/bin/golangci-lint
           rm -rf golangci-lint-$GOLANGCI_LINT_VERSION-linux-amd64*
         env:
-          GOLANGCI_LINT_VERSION: '1.46.2'
-          GOLANGCI_LINT_SHA256: '242cd4f2d6ac0556e315192e8555784d13da5d1874e51304711570769c4f2b9b'
+          GOLANGCI_LINT_VERSION: '1.51.2'
```
TerryHowe, line 29: This version works on mac! 1.50 with go 1.20 was borked.

TerryHowe (approved):

hickeyma (changes requested):
Looks good @mattfarina except should the `go.mod` be bumped to 1.20?

scottrigby:
@mattfarina @souleb  @sabre1041 @jdolitsky hmm Matt I see your test details in Keybase, fun change in Go 🧐 If we don't have a solution for this now I agree we should revert in the meantime, work to resolve this and then bring back into main after resolving (pun intended!). To help us with this - and for future, any objections to setting up our GitHub action on a matrix of Helm's supported Go versions, architectures, etc? https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs

scottrigby:
@souleb is working on a potential fix. @mattfarina any objection to holding off on reverting for a bit? Say until Monday?

mattfarina:
@hickeyma can you please re-review? It's working now.

The large size of the change is because of the deprecation of ioutil and the linter making it an error now. The change has been around for that long.

Also, there is a mock DNS resolver because some systems have trouble with the system resolvers which are inconsistent.

hickeyma (approved):
Great work on this @mattfarina especially tracking and fixing the dns issue.

PR #11898, opened 2023-03-10
Fix goroutine leak in perform

### LINKED ISSUE #11871: why are  many goroutines got stuck  at batchPerform
Output of `helm version`:
helm v3.9.0 

I use helm sdk(v3.9.0) in my operator to CRUD charts,  I find the operator's memory is growing .  I use pprof to detact this problem. Finally, I find many goroutines are got stuck at batchPerform()

image

image

Can any one help me? thanks

willzgli:
 the  function perform()  shoud wait all goroutines  created in  function batchPeform finish  to avoid goroutine leak, 

closes #11871
@gjenkins8 @joejulian please give some comments asap, thanks  a lot

**If applicable**:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,6 +17,7 @@ require (
 	github.com/gobwas/glob v0.2.3
 	github.com/gofrs/flock v0.8.1
 	github.com/gosuri/uitable v0.0.4
+	github.com/hashicorp/go-multierror v1.1.1
 	github.com/jmoiron/sqlx v1.3.5
 	github.com/lib/pq v1.10.7
 	github.com/mattn/go-shellwords v1.0.12
@@ -92,6 +93,7 @@ require (
 	github.com/gorilla/handlers v1.5.1 // indirect
 	github.com/gorilla/mux v1.8.0 // indirect
 	github.com/gregjones/httpcache v0.0.0-20180305231024-9cad4c3443a7 // indirect
+	github.com/hashicorp/errwrap v1.1.0 // indirect
 	github.com/hashicorp/golang-lru v0.5.4 // indirect
 	github.com/huandu/xstrings v1.4.0 // indirect
 	github.com/imdario/mergo v0.3.13 // indirect
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -37,6 +37,7 @@ import (
 	apiextv1beta1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 
+	multierror "github.com/hashicorp/go-multierror"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -532,6 +533,8 @@ func (c *Client) WatchUntilReady(resources ResourceList, timeout time.Duration) 
 }
 
 func perform(infos ResourceList, fn func(*resource.Info) error) error {
+	var result error
+
 	if len(infos) == 0 {
 		return ErrNoObjectsVisited
 	}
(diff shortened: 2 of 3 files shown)


gjenkins8 (approved):
LGTM, looks to be a (minimal) fix to the bug/issue (not a maintainer, will need maintainer review).

Overall, the code here could be improved. It isn't ideal how `perform`/`batchPerform` implement scatter/gather ad-hoc, rather than relying on a library primitive (which gets the details like error handling right). But a minimal fix to the existing code is first is fine.

pkg/kube/client.go
```diff
@@ -542,9 +543,12 @@ func perform(infos ResourceList, fn func(*resource.Info) error) error {
 	for range infos {
 		err := <-errs
 		if err != nil {
-			return err
+			errList = append(errList, err)
 		}
 	}
+	if errList != nil {
+		return errList[0]
```
TerryHowe, line 550: Why would you save all the errors and just use the first?

willzgli: HI, Howe,  I don't find a proper way to keep the return type not be changed and  return all the errors. I think the import thing is avoid gorotine leak here. If you have other ideas, please tell me,thanks.

TerryHowe: `var lastError error` and just return it instead of `return nil`

willzgli: yes, use lastError looks good, thanks.

TerryHowe (approved):

willzgli:
@mattfarina  why the test case TestInstallRelease fails, I run the test case locally, it passed.

willzgli:
@mattfarina  why the test case TestInstallRelease fails( ref Test build  log line 1962), I run the test case locally, it passed.

pkg/kube/client.go
```diff
@@ -532,6 +532,7 @@ func (c *Client) WatchUntilReady(resources ResourceList, timeout time.Duration)
 }
 
 func perform(infos ResourceList, fn func(*resource.Info) error) error {
+	var lastError error
```
mattfarina, line 535: Did you think about using a multi-error to capture all of them and return all of them?

mattfarina:
@willzgli I think it was a CI flake. Problem seems to be gone after I re-ran the job.

pkg/kube/client.go
willzgli: yeah , I searched two packages 
1. https://github.com/uber-go/multierr
2. https://github.com/hashicorp/go-multierror
But both of them are not active recently, and I am not sure which is better.   Do you have any suggestion to use which package ? or is there a new famous package ?

willzgli: @mattfarina  please review again

mattfarina (approved):

mattfarina:
Note, this needs a second maintainer to approve before it can be merged.

go.mod
```diff
@@ -17,6 +17,7 @@ require (
 	github.com/gobwas/glob v0.2.3
 	github.com/gofrs/flock v0.8.1
 	github.com/gosuri/uitable v0.0.4
+	github.com/hashicorp/go-multierror v1.1.1
```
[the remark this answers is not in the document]
hiddeco: With the built-in multi-error support in Go 1.20, it might be an idea to put a `// TODO` somewhere once the version in `go.mod` allows to do so. Which won't happen with #11896, but support for `1.19` should end in [~5 months](https://endoflife.date/go).

hiddeco (approved):
Not a maintainer, but quite familiar with the internals and this looks OK to me.

go.mod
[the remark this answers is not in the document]
wujunwei: I don't think it's necessary to introduce a new library, we can use`k8s.io/apimachinery/pkg/util/errors.NewAggregate([]error)` to aggregate errors, be consistent with the behavior (incorrect format) of Kubernetes incidentally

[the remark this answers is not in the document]
mattfarina: What's happening here is fine for now.

sabre1041 (approved):
LGTM

PR #11903, opened 2023-03-12
Fix random fails on TestPerform

gaiaz-iusipov:
Fix random fails on TestPerform caused by missing sync.Mutex

Also some small changes by static analysis

**If applicable**:

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -500,7 +500,7 @@ func delete(c *Client, resources ResourceList, propagation metav1.DeletionPropag
 		}
 		errs = append(errs, err)
 	}
-	if errs != nil {
+	if len(errs) != 0 {
 		return nil, errs
 	}
 	return res, nil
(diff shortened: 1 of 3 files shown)


pkg/kube/client.go
```diff
@@ -579,8 +579,8 @@ func batchPerform(infos ResourceList, fn func(*resource.Info) error, errs chan<-
 		}
 		wg.Add(1)
 		go func(i *resource.Info) {
+			defer wg.Done()
 			errs <- fn(i)
-			wg.Done()
```
willzgli, line 583: why you make this change here ? It's seems the same effect.

gaiaz-iusipov: This change is meaningless, but in general it's a good practice to do `wg.Done()` by `defer` so never forget to do that in any condition

gaiaz-iusipov: I reverted this change

gaiaz-iusipov:
Hello, any reviewers?

gaiaz-iusipov:
@mattfarina can you please review this?

PR #11908, opened 2023-03-14
WIP: Backup Kube Config to Helm Config Directory

### LINKED ISSUE #11419: Loading KUBECONFIG from stdin
Output of `helm version`: version.BuildInfo{Version:"v3.10.0", GitCommit:"ce66412a723e4d89555dc67217607c6579ffcb21", GitTreeState:"clean", GoVersion:"go1.18.6"}

Output of `kubectl version`: Client Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.1", GitCommit:"86ec240af8cbd1b60bcc4c03c20da9b98005b92e", GitTreeState:"clean", BuildDate:"2021-12-16T11:41:01Z", GoVersion:"go1.17.5", Compiler:"gc", Platform:"darwin/amd64"}

In kubectl I'm able to cat the content of a KUBECONFIG file into a variable and get connection to a cluster

```bash
$ export KUBECONFIG=$(cat ~/my-cluster.yaml)
$ kubectl --kubeconfig <(echo $KUBECONFIG) get services           
NAME         TYPE        CLUSTER-IP   EXTERNAL-IP   PORT(S)   AGE
kubernetes   ClusterIP   10.43.0.1    <none>        443/TCP   589d
```

With helm I would have the same behavior, but it doesn't work:

```bash
$ helm --kubeconfig <(echo $KUBECONFIG) list  
Error: Kubernetes cluster unreachable: Get "http://localhost:8080/version": dial tcp [::1]:8080: connect: connection refused
```
The idea is to provide credentials in a CI pipeline without storing the data on the disc. Workaround would be `helm --kube-apiserver xxx --kube-token xxxx --kube-as-user xxxx`, but this requires a lot of variables, what I can have in one. thx.

Bhargav-InfraCloud:

Backup kubeconfig to `<helm-config-path>/kubeconfig` during initialization of sub-commands (`env`, `install`, etc.).
The backup file will have `rw` permissions for `user`. All sub-commands will then use the backup kubeconfig file.

This change will help in reading the kubeconfig from a file descriptor. Eg., `/proc/self/fd/16`
```
helm --kubeconfig <(echo $KUBECONFIG) list 
```

Closes #11419 

Since the backup of kubeconfig runs on all command initialization (nothing to add explicitly by the user), as per my understanding, this should be backup compatible. But please review and let me know.

Also, since this backup happens with all sub-command initializations, no need to delete the backup file explicitly. It will get replaced every time.

I believe no documentation is required for this change as nothing much changes for the user.

**If applicable**:

@eumel8 @joejulian @programmer04 Please validate whether these changes achieve the expected result. Thanks!

### DIFF
--- a/cmd/helm/helm.go
+++ b/cmd/helm/helm.go
@@ -80,7 +80,7 @@ func main() {
 				log.Fatal(err)
 			}
 		} else {
-			debug("kube config backup aborted due to empty KubeConfig path in settings")
+			debug("kubeconfig backup aborted due to the empty KubeConfig path in the settings")
 		}
 
 		if err := actionConfig.Init(settings.RESTClientGetter(), settings.Namespace(), helmDriver, debug); err != nil {
--- a/cmd/helm/helm.go
+++ b/cmd/helm/helm.go
@@ -74,7 +74,7 @@ func main() {
 		helmDriver := os.Getenv("HELM_DRIVER")
 
 		if settings.KubeConfig != "" {
-			// If KubeConfig path is not empty, backup kube config to: <helm-config-path>/kubeconfig
+			// If KubeConfig path is not empty, backup kubeconfig to: <helm-config-path>/kubeconfig
 			// When the backup is successful, the settings.KubeConfig path is updated to backup file's path.
 			if err := settings.BackupKubeConfig(); err != nil {
 				log.Fatal(err)
--- a/cmd/helm/helm.go
+++ b/cmd/helm/helm.go
@@ -72,6 +72,18 @@ func main() {
 	// run when each command's execute method is called
 	cobra.OnInitialize(func() {
 		helmDriver := os.Getenv("HELM_DRIVER")
+
+		if settings.KubeConfig != "" {
+			// If KubeConfig path is not empty, backup kube config to: <helm-config-path>/kubeconfig
+			// When the backup is successful, the settings.KubeConfig path is updated to backup file's path.
+			err = settings.BackupKubeConfig()
+			if err != nil {
+				log.Fatal(err)
+			}
+		} else {
+			debug("kube config backup aborted due to empty KubeConfig path in settings")
+		}
+
 		if err := actionConfig.Init(settings.RESTClientGetter(), settings.Namespace(), helmDriver, debug); err != nil {
 			log.Fatal(err)
 		}
--- a/cmd/helm/helm.go
+++ b/cmd/helm/helm.go
@@ -76,8 +76,7 @@ func main() {
 		if settings.KubeConfig != "" {
 			// If KubeConfig path is not empty, backup kube config to: <helm-config-path>/kubeconfig
 			// When the backup is successful, the settings.KubeConfig path is updated to backup file's path.
-			err = settings.BackupKubeConfig()
-			if err != nil {
+			if err := settings.BackupKubeConfig(); err != nil {
 				log.Fatal(err)
 			}
 		} else {
--- a/pkg/cli/environment.go
+++ b/pkg/cli/environment.go
@@ -270,7 +270,7 @@ func (s *EnvSettings) BackupKubeConfig() error {
 			s.KubeConfig, kubeConfigBackupFilename, err)
 	}
 
-	// Update the kube config in EnvSettings to backup file path
+	// Update the kubeconfig in EnvSettings to a backup file path.
 	s.KubeConfig = kubeConfigBackupFilename
 
 	return nil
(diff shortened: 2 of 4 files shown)


Bhargav-InfraCloud:
@mattfarina My bad, I missed the [style check][1]. Please trigger CICD on the latest forch-pushed commit.

[1]: https://github.com/helm/helm/actions/runs/4416513060

MichaelMorrisEst (commented):
Tested (on linux) and looks good.
One possible small efficiency could be to only create the backup when needs be, i.e. if a file is provided proceed as today, if process substitution is in use proceed to backup. 
This would depend on there being an easy and reliable way to tell the difference, which I dont know if there is

programmer04 (commented):
Thanks for your contribution @Bhargav-InfraCloud 🎉 It solves the problem. For the code review, I've found just some nitpicks.

But still, I'm thinking about the overall approach. The drawbacks of that solution are
- It won't work when `helm` is invoked in an environment with a read-only filesystem.
- It introduces a security risk, someone wants to use process substitution to avoid storing config on the machine, especially in some CI environments, but now it will be saved. Next run of CI - different pipeline can read it and reveal secrets. Cleaning up the config on exit is risky too because `helm` may fail before doing it and the config won't be removed.

eumel8:
fantastic work, @Bhargav-InfraCloud 

now it's the same behaviour:


```
$ export KUBECONFIG=$(echo ~/.config/quickstart.yaml)
$ kubectl --kubeconfig <(cat $KUBECONFIG) get services
NAME         TYPE        CLUSTER-IP   EXTERNAL-IP   PORT(S)   AGE
kubernetes   ClusterIP   10.43.0.1    <none>        443/TCP   27d

$ ./helm --kubeconfig <(cat $KUBECONFIG) -n kube-system list
NAME            NAMESPACE       REVISION        UPDATED                                 STATUS          CHART                APP VERSION
traefik         kube-system     21              2023-03-17 08:38:45.260444092 +0000 UTC deployed        traefik-19.0.400     2.9.4
traefik-crd     kube-system     1               2023-02-17 20:24:48.089607491 +0000 UTC deployed        traefik-crd-19.0.400
```

Bhargav-InfraCloud:
Thanks for confirming @eumel8 😊

Bhargav-InfraCloud:
@MichaelMorrisEst Thanks for reviewing.
> One possible small efficiency could be to only create the backup when needs be, i.e. if a file is provided proceed as today, if process substitution is in use proceed to backup.
This would depend on there being an easy and reliable way to tell the difference, which I dont know if there is

Agreed. Currently, with these changes, `kubeconfig` will be backed-up when the user specifies a path using `--kubeconfig`. i.e., not when using the default path. However, I think further limiting the scenarios to backup will only be better. 

Open to ideas on this from others. Thanks!

Bhargav-InfraCloud:
@programmer04 Thanks for the review and picks. I've made the changes accordingly. About the observations:
> It won't work when helm is invoked in an environment with a read-only filesystem.

I doubt this is a valid argument as helm does write files like `repositories.yaml` and `repositories.lock` in the default config path. Despite that, if a user is using it with read-only fs, I feel that's an odd usage.

> It introduces a security risk, someone wants to use process substitution to avoid storing config on the machine, especially in some CI environments, but now it will be saved. Next run of CI - different pipeline can read it and reveal secrets. Cleaning up the config on exit is risky too because helm may fail before doing it and the config won't be removed.

➕  Yep. I agree with this. Want to hear from others about alternative approaches to this.

programmer04:
yup @Bhargav-InfraCloud, I've checked and Helm doesn't work with a read-only file system currently. It has to write those files as you mentioned. So we shouldn't bother about it

Why have I thought about this use case in the first place? Because in K8s you can run containers with [SecurityContext](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) configured in the way that the root file system is read-only and I thought (wrongly) that Helm is able to work in such environment. It's quite popular to have such a policy in heavily regulated environments

joejulian (commented):
The reason the issue exists is that there are organizations that do not want the credentials written to disk. Writing it to disk goes against security policies.

Bhargav-InfraCloud:
@programmer04 Agreed. K8s and helm have their baseline differences in design. However, [security risks][1] that you mentioned still stands.

@joejulian Got it. Will see if there's an alternative approach with in-memory.

[1]: https://github.com/helm/helm/pull/11908#pullrequestreview-1342162759

programmer04:
Making it in memory without weird hacks requires a bigger refactor that may be a little bit tricky, because `"helm.sh/helm/v3/pkg/kube"` is imported by other projects thus API can't be broken. I described it in more details in the https://github.com/helm/helm/issues/11419#issuecomment-1376129937

PR #11930, opened 2023-03-22
Use persistent `RESTClientGetter` in CLI

hiddeco:
This changes the `RESTClientGetter` to one which caches the underlying
clients after first initialization.

This reduces memory usage for all commands, with observed reductions
from ~4MiB for `helm get all`, to ~10MiB for `helm install` (using
https://github.com/stefanprodan/podinfo).

For charts which make use of Custom Resource Definitions, this is
likely even higher as the client will not be initialized twice (one
time for the installation process of the CRDs, the second time for the
actual installation of the chart).

Follow up on #11886

### DIFF
--- a/pkg/cli/environment.go
+++ b/pkg/cli/environment.go
@@ -106,27 +106,36 @@ func New() *EnvSettings {
 	env.Debug, _ = strconv.ParseBool(os.Getenv("HELM_DEBUG"))
 
 	// bind to kubernetes config flags
-	env.config = &genericclioptions.ConfigFlags{
-		Namespace:        &env.namespace,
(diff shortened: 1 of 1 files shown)


hiddeco:
Converting this to a draft as it will cause issues when charts are for example creating Custom Resource Definitions on demand during install (for example [OPA Gatekeeper](https://artifacthub.io/packages/helm/gatekeeper/gatekeeper/3.1.1)) while later attempting to make use of them in a post-install hook.

As Helm does not clear the REST mapper between individual action steps (e.g. after pre-install and before post-install hooks), it does not "see" the new Custom Resource Definitions and the installation will fail.

PR #11711, opened 2023-01-10
Oci transport refinement

souleb:
**What this PR does / why we need it**:
If implemented, users will be able to use custom certificates and CA when interacting with OCI registries.

This is convenient for user interacting with registries that hold custom signed certificates. By providing the CA the server certificate can be validate during handshakes.

This also enable interacting with insecure registries.

The following flags can be used when using the helm cli with OCI Registries:
- `cert-file`, identify registry client using this SSL certificate file
- `key-file`, identify registry client using this SSL key file
- `ca-file`, verify certificates of HTTPS-enabled servers using this CA bundle
- `insecure-skip-tls-verify`, skip tls certificate checks

**Special notes for your reviewer**:
For the oras client, we use a functional option to provide a custom `httpClient` that can then be passed to `authorizer` and `resolver`. This is flexible, and sdk users can construct their client with any needed custom transport.

We tested this by setting up a docker registry in an ec2 server with a certificate and CA generated with openSSL.

Regression test have been manually performed for the following helm cmds:
- helm push
- helm pull
- helm show
- helm template
- helm install
- helm upgrade

**If applicable**:
- [ ] this PR contains documentation
- [x] this PR contains unit tests
- [x] this PR has been tested for backwards compatibility

cc @sabre1041

PR #11938, opened 2023-03-23
rely on mockdns to resolve 0x7f000001 instead of default resolver

souleb:

Rely on mockdns to resolve 0x7f000001 instead of default resolver. Related to #11711 

**If applicable**:

### DIFF
--- a/pkg/registry/utils_test.go
+++ b/pkg/registry/utils_test.go
@@ -39,6 +39,8 @@ import (
 	"github.com/stretchr/testify/suite"
 	"golang.org/x/crypto/bcrypt"
 
+	"github.com/foxcpp/go-mockdns"
+
 	"helm.sh/helm/v3/internal/tlsutil"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/registry/utils_test.go
```diff
@@ -82,9 +84,19 @@ func setup(suite *TestSuite, tlsEnabled bool, insecure bool) *registry.Registry
...
 		var tlsConf *tls.Config
 		tlsConf, err = tlsutil.NewClientTLS(tlsCert, tlsKey, tlsCA, insecure)
+
+		r := mockdns.Resolver{
+			Zones: map[string]mockdns.Zone{
+				"0x7f000001.": {
+					A: []string{"127.0.0.1"},
+				},
+			},
+		}
```
mattfarina, line 94: Thanks for your work on this.

I tried this change in a variety of environments and it didn't always work. Along the way the default resolver is still sometimes used. :facepalm: 

I was able to make a bigger change to #11896 based on a similar idea (and same packages). After seeing it here and discussing it on the Helm dev call. See [this file](https://github.com/helm/helm/pull/11896/files#diff-6f6a553e3139587305901b6320ef20e65ea23b47a760f84e700b065e7a0239a1).

PR #11948, opened 2023-03-28
fix: failed testcase on windows

### LINKED ISSUE #11947: Run testcase  failed on windows
I found some test case will fail on windows.
https://github.com/helm/helm/blob/2e3e22a0033cc6d3833d6d3e84353c1aecb06dc6/pkg/chartutil/save_test.go#L37-L45
https://github.com/helm/helm/blob/2e3e22a0033cc6d3833d6d3e84353c1aecb06dc6/pkg/chartutil/save_test.go#L200-L204
output like below:
```
=== RUN   TestSaveDir
    save_test.go:230: Templates data did not match
--- FAIL: TestSaveDir (25.29s)
```

One of the reason why first testcase failed is that `path.Join(tmp, "newdir")` should be `filepath.Join` in line 41,the second is that variable `c2` should be compared to `c` and its Templates's name is always combined with slashes, not applicable to `filepath.Join`

wujunwei:
fix #11947

**If applicable**:

### DIFF
--- a/pkg/chartutil/save_test.go
+++ b/pkg/chartutil/save_test.go
@@ -38,7 +38,7 @@ func TestSave(t *testing.T) {
 	tmp := ensure.TempDir(t)
 	defer os.RemoveAll(tmp)
 
-	for _, dest := range []string{tmp, path.Join(tmp, "newdir")} {
+	for _, dest := range []string{tmp, filepath.Join(tmp, "newdir")} {
 		t.Run("outDir="+dest, func(t *testing.T) {
 			c := &chart.Chart{
 				Metadata: &chart.Metadata{
@@ -210,7 +210,7 @@ func TestSaveDir(t *testing.T) {
 			{Name: "scheherazade/shahryar.txt", Data: []byte("1,001 Nights")},
 		},
 		Templates: []*chart.File{
-			{Name: filepath.Join(TemplatesDir, "nested", "dir", "thing.yaml"), Data: []byte("abc: {{ .Values.abc }}")},
+			{Name: path.Join(TemplatesDir, "nested", "dir", "thing.yaml"), Data: []byte("abc: {{ .Values.abc }}")},
 		},
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/chartutil/save_test.go
```diff
@@ -210,7 +208,7 @@ func TestSaveDir(t *testing.T) {
 			{Name: "scheherazade/shahryar.txt", Data: []byte("1,001 Nights")},
 		},
 		Templates: []*chart.File{
-			{Name: filepath.Join(TemplatesDir, "nested", "dir", "thing.yaml"), Data: []byte("abc: {{ .Values.abc }}")},
+			{Name: TemplatesDir + "/nested/dir/thing.yaml", Data: []byte("abc: {{ .Values.abc }}")},
```
joejulian, line 211: Why is this changed?

wujunwei: > Why is this changed?

https://github.com/helm/helm/blob/2e3e22a0033cc6d3833d6d3e84353c1aecb06dc6/pkg/chartutil/save_test.go#L230-L232

`c2.Templates[0].name` is always joined with slashes ("templates/nested/dir/thing.yaml"), not equal to `filepath.Join(TemplatesDir, "nested", "dir", "thing.yaml")` whose value is "templates\nested\dir\thing.yaml".

It will fail on windows with message below:
```
=== RUN   TestSaveDir
    save_test.go:231: Templates data did not match
--- FAIL: TestSaveDir (0.13s)
```

joejulian: > c2.Templates[0].name is always joined with slashes

should it be on windows?

wujunwei: > > c2.Templates[0].name is always joined with slashes
> 
> should it be on windows?

Yes, Templates[0].name comes from `tar.Header.name` ,even the tar archive file created on widnows ( or created by `helm create`),  I have try  it in go and python, the file name  is always joined with slashes.

Also you can see from the go's source code, it has hard coded that join with a '/':
https://github.com/golang/go/blob/f80e270babcddd28e29a52c1cb834ffed427bd8a/src/archive/tar/reader.go#L462-L464

And I found [standard introduction for `tar`](https://www.gnu.org/software/tar/manual/html_node/Standard.html) ,it shows "The name field is the file name of the file, with directory names (if any) preceding the file name, separated by slashes."

BTW,In the end, the following code is useless and may have misleading effects.
https://github.com/helm/helm/blob/27ebfa8c561e758b60872b9bb081253036e559ff/pkg/chart/loader/archive.go#L135-L145

mattfarina (approved):
Verified the fix on windows

mattfarina:
Merging with single review as this is size/XS which is allowed.

PR #11952, opened 2023-03-29
feat: helm test --logs, support outputing logs only from failed pods, succeeded pods, or all pods.

### LINKED ISSUE #11819: Helm test --logs to optionally output only from failed tests
It would be great if aside from [respecting --filter when using the --logs flag](https://github.com/helm/helm/issues/10566) we were also able to specify that `helm test --logs` outputs only logs from failed tests.

An example use-case is for CI/CD pipelines, so that logs are output only if there is an error, for investigation, whereas successful tests simply yield some short success message and exit.

This is especially useful for cases where we want to delete the test pod after execution so as to not take up space on the cluster, especially since investigating test failures can take place a while after the test execution. Furthermore, looking at test job results is faster and more convenient than searching through Pod logs.

### Suggested solution 1 - an additional 'logs-failed' flag

`helm test mychart --logs-failed`

### Suggested solution 2 - parameters for the existing 'logs' flag

`helm test mychart --logs all` (default)
`helm test mychart --logs failed`
`helm test mychart --logs succeeded`

rhzx3519:

1. helm test --logs all (default) // output logs of all pods 
2. helm test --logs failed  // only output logs of failed pods 
3. helm test --logs succeeded // only output logs of succeeded pods

In some CI/CD scenarios, operators only care about failed test logs; too many logs of succeeded tests would be a distraction for them.

Have been tested manually.
**Test case**: a helm with two test pods. One is doomed to succeed, and the other is doomed to fail.

**If applicable**:

closes #11819

### DIFF
--- a/cmd/helm/release_testing.go
+++ b/cmd/helm/release_testing.go
@@ -40,7 +40,7 @@ The tests to be run are defined in the chart that was installed.
 func newReleaseTestCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 	client := action.NewReleaseTesting(cfg)
 	var outfmt = output.Table
-	var outputLogs bool
+	var outputLogs string
 	var filter []string
 
 	cmd := &cobra.Command{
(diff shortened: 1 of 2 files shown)


rhzx3519:
### Test Case

A chart with two test pods

Screenshot 2023-03-29 at 9 54 22 pm


### `helm test demo --logs succeeded`

Screenshot 2023-03-29 at 9 52 18 pm

### `helm test demo --logs failed`
Screenshot 2023-03-29 at 9 51 21 pm


### `helm test demo --logs all`
Screenshot 2023-03-29 at 9 53 21 pm

PR #11963, opened 2023-04-03
fix: do not error for remote chart if there's a missing local repo cache

### LINKED ISSUE #11961: Helm fails to retrieve a remote chart if there are local repos that are not cached
This is a continuation of #9533 with better reproduction steps/description of the bug.

Helm fails to retrieve a remote chart if some other local repo is missing a local cache. Here are reproduction steps:

1. Remove all local repos `helm repo remove ...`
2. Add some remote repo: `helm repo add ingress-nginx https://kubernetes.github.io/ingress-nginx`
3. Remove helm cache (on Windows, this was deleting all files in `C:\Users\<user>\AppData\Local\Temp\helm\respository\*`)
3. Try to run template for a different remote chart: `helm template traefik --repo https://helm.traefik.io/traefik`
4. Notice the output says:
```
Error: no cached repo found. (try 'helm repo update'): open C:\Users\<user>\AppData\Local\Temp\helm\repository\ingress-nginx-index.yaml: The system cannot find the file specified.
```

Specifically see that the output is pointing to `ingress-nginx` instead of `traefik`. Having a valid cache for ingress-nginx shouldn't impact fetching an unrelated remote call, especially since the cache is periodically cleared. You can see that once you do `helm repo update`, the `template` command succeeds, despite there's still nothing cached for `traefik`

---

Output of `helm version`:

```
version.BuildInfo{Version:"v3.10.2", GitCommit:"50f003e5ee8704ec937a756c646870227d7c8b58", GitTreeState:"clean", GoVersion:"go1.18.8"}
```

Output of `kubectl version`:

```
Client Version: version.Info{Major:"1", Minor:"26", GitVersion:"v1.26.0", GitCommit:"b46a3f887ca979b1a5d14fd39cb1af43e7e5d12d", GitTreeState:"clean", BuildDate:"2022-12-08T19:58:30Z", GoVersion:"go1.19.4", Compiler:"gc", Platform:"windows/amd64"}
Kustomize Version: v4.5.7
```

Cloud Provider/Platform (AKS, GKE, Minikube etc.):  N/A

mikeseese:
This PR closes #11961 by continuing to `scanReposForURL` if one of the local repos doesn't have a cache rather than erroring. This PR also adds the hint to `helm repo update` to `ErrNoOwnerRepo` to catch cases that are expecting to find it in the local cache (i.e. `u := localrepo/cache`).

Ultimately, this makes it so that a full HTTP URL for the chart will never fail because in `chart_downloader.go` it will proceed to downloading it if `scanReposForUrl` returns `ErrNoOwnerRepo` (line 216):
> If there is no special config, return the default HTTP client and swallow the error.

This PR also adds a test that tests for the case where there's a non-empty repository config but an empty repository cache.

I verified this actually fixes #11961 by using the reproduction steps listed there compared with the released/installed version of `helm` and the manually built version located in `bin/helm` (after running `make build`). Here was the final output of that:

```
/mnt/d/work/helm$ helm template traefik --repo https://helm.traefik.io/traefik
Error: no cached repo found. (try 'helm repo update'): open /home/seese/.cache/helm/repository/ingress-nginx-index.yaml: no such file or directory
/mnt/d/work/helm$ bin/helm template traefik --repo https://helm.traefik.io/traefik
---
# Source: traefik/templates/rbac/serviceaccount.yaml
kind: ServiceAccount
apiVersion: v1
metadata
<...>
```

**If applicable**:

### DIFF
--- a/pkg/downloader/chart_downloader.go
+++ b/pkg/downloader/chart_downloader.go
@@ -54,7 +54,7 @@ const (
 )
 
 // ErrNoOwnerRepo indicates that a given chart URL can't be found in any repos.
-var ErrNoOwnerRepo = errors.New("could not find a repo containing the given URL")
+var ErrNoOwnerRepo = errors.New("could not find a repo containing the given URL (try 'helm repo update')")
 
 // ChartDownloader handles downloading a chart.
 //
@@ -380,7 +380,8 @@ func (c *ChartDownloader) scanReposForURL(u string, rf *repo.File) (*repo.Entry,
 		idxFile := filepath.Join(c.RepositoryCache, helmpath.CacheIndexFile(r.Config.Name))
 		i, err := repo.LoadIndexFile(idxFile)
 		if err != nil {
-			return nil, errors.Wrap(err, "no cached repo found. (try 'helm repo update')")
+			// No cache for this repository; let's keep looking rather than erroring
+			continue
 		}
 
 		for _, entry := range i.Entries {
--- a/pkg/downloader/chart_downloader_test.go
+++ b/pkg/downloader/chart_downloader_test.go
@@ -28,8 +28,9 @@ import (
 )
 
 const (
-	repoConfig = "testdata/repositories.yaml"
-	repoCache  = "testdata/repository"
+	repoConfig     = "testdata/repositories.yaml"
+	repoCache      = "testdata/repository"
+	repoCacheEmpty = "testdata/repository-empty"
 )
 
 func TestResolveChartRef(t *testing.T) {
@@ -340,6 +341,33 @@ func TestScanReposForURL(t *testing.T) {
 
 	// A lookup failure should produce an ErrNoOwnerRepo
 	u = "https://no.such.repo/foo/bar-1.23.4.tgz"
+	if _, err = c.scanReposForURL(u, rf); err != ErrNoOwnerRepo {
+		t.Fatalf("expected ErrNoOwnerRepo, got %v", err)
+	}
+}
+
+// In an scenario where a full URL is provided, chart_downloader will swallow the
+// ErrNoOwnerRepo error and return an HTTP client to download the chart. This specific
+// test shows that the chart will get downloaded in that scenario if the cache is empty
+// (hence `repoCacheEmpty`).
+func TestScanReposForURLWithEmptyCache(t *testing.T) {
+	c := ChartDownloader{
+		Out:              os.Stderr,
+		Verify:           VerifyLater,
+		RepositoryConfig: repoConfig,
+		RepositoryCache:  repoCacheEmpty,
+		Getters: getter.All(&cli.EnvSettings{
+			RepositoryConfig: repoConfig,
+			RepositoryCache:  repoCacheEmpty,
+		}),
+	}
+
+	u := "http://example.com/alpine-0.2.0.tgz"
+	rf, err := repo.LoadFile(repoConfig)
+	if err != nil {
+		t.Fatal(err)
+	}
+
 	if _, err = c.scanReposForURL(u, rf); err != ErrNoOwnerRepo {
 		t.Fatalf("expected ErrNoOwnerRepo, got %v", err)
 	}


mikeseese:
Looking at the comments in `chart_downloader.go`, I think the original intention of `scanReposForURL` is to handle more than just finding the first repo in `repositories.yaml` that has the same URL. I don't have time to ensure the changes are not introducing a regression, so I'm leaving the PR as is. This ensures the least amount of regressions, is easier to grok/approve, requires less tests, solves my specific use cases, etc. in favor of leaving "making scanReposForURL" more performant for another, more determined contributor.

MissedTheMark (approved):
Change looks good, as does the new test.

Manually testing the change locally shows that it works correctly.
I have also verified that the new test fails on the old code, and passes with this change.

### New test
#### Test failing on old code:
SCR-20240510-maac

#### Test passing on new code:
SCR-20240510-lzib

### Manually testing locally
#### Before this change:
After adding a repo (ingress-nginx) then clearing the Helm cache (note the unexpected reference to `ingress-nginx-index.yaml`):
``` console
$ helm template traefik --repo https://helm.traefik.io/traefik
Error: no cached repo found. (try 'helm repo update'): open /Users/mark/Library/Caches/helm/repository/ingress-nginx-index.yaml: no such file or directory
```

#### After this change:
After adding a repo (ingress-nginx) then clearing the Helm cache:
``` console
$ helm template traefik --repo https://helm.traefik.io/traefik
---
# Source: traefik/templates/rbac/serviceaccount.yaml
kind: ServiceAccount
apiVersion: v1
metadata:
  name: release-name-traefik
 ... etc
```

MissedTheMark:
@mikeseese Looks like there's some CI issue, might be worth a merge from main?

mikeseese:
Thanks @MissedTheMark for the review! I just rebased off of `main`. Let me know if there are any other issues.

pkg/downloader/chart_downloader.go
```diff
@@ -54,7 +54,7 @@ const (
 )
 
 // ErrNoOwnerRepo indicates that a given chart URL can't be found in any repos.
-var ErrNoOwnerRepo = errors.New("could not find a repo containing the given URL")
+var ErrNoOwnerRepo = errors.New("could not find a repo containing the given URL (try 'helm repo update')")
```
mattfarina, line 57: This error is part of a package that can be used by part of the Helm CLI or part of other tools that import the library.  Suggesting to try `helm repo update` isn't going to work here. Something higher up should catch this error and add the helm command to help to it.

mattfarina (changes requested):
The idea makes sense. Due to the separation between the Helm Client (CLI) and the SDK there is something that needs to be tweaked in the messaging.

pkg/downloader/chart_downloader.go
mikeseese: @mattfarina There are 2 other occurrences of `(try 'helm repo update')` in this file (lines 284 and 289), which is why this messaging was used (in favor of consistency above all). I'd also argue that providing this messaging isn't misleading as it still cues the user to run the CLI command (even if they're not in the CLI). As a user, I wouldn't mind seeing this message in libraries/UIs/etc. If you still want a change, I'd rather remove the hint altogether if it gets this PR accepted faster. If that's also not an option, then I'll try to fit in some time to implement the requested change; let me know how you'd like to proceed.

mikeseese: @mattfarina Just wanted to follow up on how you'd like to proceed ☝️ thanks!

mikeseese:
@mattfarina @gjenkins8 (or someone else more appropriate) Is there any desire to fix this ongoing issue that [impacts Pulumi](https://github.com/pulumi/pulumi-kubernetes/issues/1518)? I'm happy to refresh this PR with upstream changes since the original submission in 2023 if yes; but if there isn't any desire to fix #11961 (or fix it this way), I don't think there's a reason to keep this PR open.

mikeseese:
Due to the staleness from reviewers/maintainers, I'm closing this PR as it seems unwanted. I can reopen and update if there's renewed interest in the future.

PR #11973, opened 2023-04-07
fail quiet linting when an error occurs

### LINKED ISSUE #11411: Using `helm lint <path> --quiet` produces no output, even when there's an error.
### What happened?

When running `helm lint <path> --quiet`, for some reason, `helm doesn't show any errors. However, if `--quiet` is removed, then errors are shown.

### Steps to reproduce

1. run `helm lint nonexistent/ --quiet`
2. No error is displayed despite that `nonexistent/` doesn't exist. The command exit code is `0` and not `1` to indicate an error.
3. run `helm lint nonexistent/`
4. the output is
```
==> Linting nonexistent/
Error unable to check Chart.yaml file in chart: stat nonexistent/Chart.yaml: no such file or directory

```

1. run `mkdir charts/`
2. run `helm lint charts/ --quiet`
3. no error is displayed despite the missing `Chart.yaml`. The command exit code is `0` and not `1` to indicate an error.
4. run `helm lint charts/`
```
==> Linting charts/
Error unable to check Chart.yaml file in chart: stat charts/Chart.yaml: no such file or directory

```

### Expected Behavior

As stated in the command usage `--quiet print only warnings and errors` which is not honored in this case. On error, the command should return `1`.

### Actual Behavior

No outputs at all, including errors. Even with a clear error, the command exit code is `0`.

### Additional context

Output of `helm version`: `version.BuildInfo{Version:"v3.10.0", GitCommit:"ce66412a723e4d89555dc67217607c6579ffcb21", GitTreeState:"clean", GoVersion:"go1.18.6"}`

Output of `kubectl version`: `Client Version: version.Info{Major:"1", Minor:"25", GitVersion:"v1.25.2", GitCommit:"5835544ca568b757a8ecae5c153f317e5736700e", GitTreeState:"clean", BuildDate:"2022-09-21T14:33:49Z", GoVersion:"go1.19.1", Compiler:"gc", Platform:"linux/amd64"}`

Cloud Provider/Platform (AKS, GKE, Minikube etc.): `k3s` and bare-metal

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

joejulian:

When linting with the `--quiet` flag, non-linting errors, such as `no such file or directory` would not fail and would exit with a 0 result. This now checks for errors as well as linting messages and fails correctly.

Fixes #11411 

**If applicable**:

### DIFF
--- a/cmd/helm/lint_test.go
+++ b/cmd/helm/lint_test.go
@@ -53,6 +53,11 @@ func TestLintCmdWithQuietFlag(t *testing.T) {
 		name:   "lint chart with warning using --quiet flag",
 		cmd:    "lint --quiet testdata/testcharts/chart-with-only-crds",
 		golden: "output/lint-quiet-with-warning.txt",
+	}, {
+		name:      "lint non-existent chart using --quiet flag",
+		cmd:       "lint --quiet thischartdoesntexist/",
+		golden:    "",
+		wantError: true,
 	}}
 	runTestCmd(t, tests)
 
--- a/pkg/action/lint.go
+++ b/pkg/action/lint.go
@@ -82,7 +82,7 @@ func HasWarningsOrErrors(result *LintResult) bool {
 			return true
 		}
 	}
-	return false
+	return len(result.Errors) > 0
 }
 
 func lintChart(path string, vals map[string]interface{}, namespace string, strict bool) (support.Linter, error) {


MichaelMorrisEst (commented):
The comment in cmd/helm/lint.go func newLintCmd lines 97-100 could probably also be clarified - only the lint errors are included in results.Messages, there may be other errors in results.Errors that are not in result.Messages
The logic of the if statements following the comment I think  is still ok though as it prevents errors which are in both result.Messages and result.Errors from being output twice but doesn't prevent any of the errors only in result.Errors being output

mattfarina (approved):
This makes sense. Thanks for helping lint catch non-lint errors that should be reported.

mattfarina:
Merged as size/XS can be merged with a single maintainers review.

Issue #11805, opened 2023-02-09
SDK 3.9.4: Goroutine leak in reportToRun 


Segflow:
In helm.sh/helm/v3 v3.9.4 the method `reportToRun` is implemented as follow:
```
func (i *Install) reportToRun(c chan<- resultMessage, rel *release.Release, err error) {
	i.Lock.Lock()
	if err != nil {
		rel, err = i.failRelease(rel, err)
	}
	c <- resultMessage{r: rel, e: err}
	i.Lock.Unlock()
}
```

I have confirmed that after performing an install action (using `RunWithContext`) leads to goroutine leak. 

![image](https://user-images.githubusercontent.com/2049730/217885060-010137c2-b77b-4e0a-a361-5f88c9af9902.png)

The above screenshot shows that goroutines are stuck in `reportToRun`. Here a sample stack trace of one of those routines:

```
goroutine 15463 [chan send, 32 minutes]:
helm.sh/helm/v3/pkg/action.(*Install).reportToRun(0xc0026086e0, 0xc0026804c0?, 0x3?, {0x22dd9c0?, 0xc006a7bf20?})
	/go/pkg/mod/helm.sh/helm/v3@v3.9.4/pkg/action/install.go:435 +0xd6
helm.sh/helm/v3/pkg/action.(*Install).performInstall(0xc0026086e0, 0xc005269f60?, 0xc005216310, {0x0?, 0x0?, 0x0?}, {0xc0026804c0, 0x3, 0x4})
	/go/pkg/mod/helm.sh/helm/v3@v3.9.4/pkg/action/install.go:389 +0x2ea
created by helm.sh/helm/v3/pkg/action.(*Install).RunWithContext
	/go/pkg/mod/helm.sh/helm/v3@v3.9.4/pkg/action/install.go:349 +0x128a
```

The block seems to be in the `chan send` call. 

On another note, I believe if we fall into the error path, the  `i.Lock` will stay locked forever.

Segflow:
According to this code:
```

// Run executes the installation with Context
func (i *Install) RunWithContext(ctx context.Context, chrt *chart.Chart, vals map[string]interface{}) (*release.Release, error) {
        ...
        rChan := make(chan resultMessage)
	doneChan := make(chan struct{})
	defer close(doneChan)
	go i.performInstall(rChan, rel, toBeAdopted, resources)
	go i.handleContext(ctx, rChan, doneChan, rel)
	result := <-rChan
	//start preformInstall go routine
[...]

PR #11978, opened 2023-04-10
Fix goroutine leak in action install

mattfarina:
During the install process there was a place where an install process could be stuck trying to write to a channel. This would happen when a context had completed prior to performInstall finishing. In a short running Helm Client this was not a problem. But, for long running applications that use Helm as an SDK there are problems where a memory leak ends up happening due to goroutines never being able to complete.

This fix provides a means for performInstall to write to its channel using the method already used to fix the upgrade issue of the same kind.

**What this PR does / why we need it**: Fixes a memory leak described in #11805

**If applicable**:


Segflow:
Thanks!

sabre1041 (approved):
LGTM

PR #11985, opened 2023-04-12
Only run tests once

mattfarina:
Tests are being run twice in CI. Once with code coverage being report and once without. Tests only need to be run once.

When Helm previously had testing in CircleCI it was the code coverage tests so that was retained here.

**What this PR does / why we need it**: CI doesn't need to spend the time running tests twice.

**If applicable**:


sabre1041 (approved):
LGTM

PR #11991, opened 2023-04-13
Make test scripts run without /bin/bash

twz123:

The test scripts hardcoded `#!/bin/bash` while they are not really requiring bash. Use the more portable `#!/usr/bin/env sh` instead, so that they use the default shell. This makes things work on systems without a bash in the /bin folder.

**If applicable**:


benoittgt:
Based on "Some distros don't ship bash by default, most notably Alpine. Since those scripts are by all means POSIX compatible, it seems unnecessary to require bash"

I think we should validate this PR.

TerryHowe (approved):
/lgtm

banjoh (approved):

gjenkins8 (approved):

PR #12009, opened 2023-04-20
feat(env): add output flag to env command

### LINKED ISSUE #11909: Can the --output flag be added to the "helm env" command?
Output of `helm version`: version.BuildInfo{Version:"v3.10.2", GitCommit:"50f003e5ee8704ec937a756c646870227d7c8b58", GitTreeState:"clean", GoVersion:"go1.19.3"}

Small request to have `helm env` support `--output`.

I'm writing some cross platform tooling and calling `helm env` to get all the helm config paths on the current system.

It would be really great to be able to do `helm env --output json`!

kkocha:

Add `--output` flag to `helm env` command.
Supported formats are env(key-value), python, and yaml.

Also add test cases for the flag.

closes #11909 

This PR adds help of `--output` flag to `helm help env`.
Though I have not checked below, does it mean containing documentation in this PR?

**If applicable**:


joejulian:
> does it mean containing documentation in this PR?

The answer in the maintainers' doc: "If a PR contains a new feature, we tend to require accompanying documentation. This may come in the form of additional help text in the `helm --help` section, or it may require additional edits or updates in the docs."

This should probably go into the contributor doc as well.

pkg/cli/environment.go
```diff
@@ -189,7 +189,7 @@ func envCSV(name string) (ls []string) {
 
 func (s *EnvSettings) EnvVars() map[string]string {
 	envvars := map[string]string{
-		"HELM_BIN":               os.Args[0],
+		"HELM_BIN":               envOr("HELM_BIN", os.Args[0]),
```
gjenkins8, line 192:
```suggestion
		"HELM_BIN":               os.Args[0],
```

gjenkins8 (changes requested):
review comment

pkg/cli/environment.go
gjenkins8: > Would it be all right if I revert the line and remove those test cases?

I think the test cases should be kept to demostrate the functionality works (and ensure it continues to work into the future)

> Unfortunately, I have been able to find the way to solve this FAILs.

It seems like the code could be structured differently? To e.g. pass the value of the helm binary path into the logic. Or, the output could be templated/formatted to avoid requiring a static string?

scottrigby:
Needs rebase. Note Helm 3 is not accepting new features.

PR #12029, opened 2023-04-26
Support for multiple yaml documents in stdin/file

### LINKED ISSUE #10866: Support multiple YAML documents through stdin
Hi,

as maintainer of https://github.com/jkroepke/helm-secrets, I'm looking into a way to pass decrypted value secured files to helm.

Currently, encrypted value files needs to be decrypted first and the decrypted value file needs to be passed 2 a second helm call.

Helm allow values from stdin (https://github.com/helm/helm/pull/2809), but it's not possible to pass multiple YAML documents through stdin.

jkroepke:
Support for multiple yaml documents in stdin/file

Fixes #10866 

Tested stdin yaml:

```yaml
---
a:
  b: c
  c: d
  e: f
---
e: value
a:
  b: f
  e:
  
test: >-
  hello
  ---
  world
---
a:
  c: 4
```

Before (main branch):
```
% cat values.yaml | ./bin/helm template --repo https://jkroepke.github.io/helm-charts/ values -f - 
---
# Source: values/templates/values.yaml
a:
  b: c
  c: d
  e: f
```

After (this PR):
```
% cat values.yaml | ./bin/helm template --repo https://jkroepke.github.io/helm-charts/ values -f - 
---
# Source: values/templates/values.yaml
a:
  b: f
  c: 4
  e: null
e: value
test: hello --- world
```

**If applicable**:

### DIFF
--- a/pkg/cli/values/options.go
+++ b/pkg/cli/values/options.go
@@ -17,13 +17,14 @@ limitations under the License.
 package values
 
 import (
+	"bytes"
 	"io"
 	"net/url"
 	"os"
 	"strings"
 
 	"github.com/pkg/errors"
-	"sigs.k8s.io/yaml"
+	"k8s.io/apimachinery/pkg/util/yaml"
 
 	"helm.sh/helm/v3/pkg/getter"
 	"helm.sh/helm/v3/pkg/strvals"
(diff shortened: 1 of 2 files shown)


pkg/cli/values/options.go
```diff
@@ -97,6 +95,28 @@ func (opts *Options) MergeValues(p getter.Providers) (map[string]interface{}, er
 	return base, nil
 }
 
+// mergeYaml merges data into an existing Go map. data must be a YAML format.
+// If the file contains multiple documents, each document will be merged separately.
+func mergeYaml(base map[string]interface{}, data []byte) (map[string]interface{}, error) {
```
gjenkins8, line 100: Helm already has a well defined process for merging yaml files. For example, maps are merged, arrays are replaced, `null` values cause deletion. See: https://github.com/helm/helm/blob/main/pkg/chartutil/coalesce.go I think it would be best to follow existing behavior, and effectively treat a multiple-doc yaml values file as multiple values files.

Also see this (upcoming, presumably) PR, which looks to modify this process some:
https://github.com/helm/helm/pull/11440/files

gjenkins8: Or, at least, I think the business logic of how to handle multiple yaml documents doesn't belong in options parsing code (given the above business logic around values handling)

gjenkins8 (changes requested):
I think `mergeYaml` should follow existing helm value processes

gjenkins8 (changes requested):
I think mergeYaml should follow existing helm value processes

pkg/cli/values/options.go
jkroepke: @gjenkins8 the existing behavior is not changed. The `mergeYaml` function will use the current `mergeMaps` function. The relevant function `mergeMaps` which holds the logic has been moved from line 59 to line 114.

jkroepke:
> I think mergeYaml should follow existing helm value processes

This is the current design in this PR.

pkg/cli/values/options.go
gjenkins8: ugh, oops. You're right. I missed that the existing logic ie. passing multiple files uses `mergeMaps` already.

jkroepke:
@technosophos I would appreciate a review here. Thanks!

jkroepke:
I would appreciate some feedback here, thanks!

jkroepke:
I would appreciate some feedback here, thanks!

PR #12034, opened 2023-04-27
Updating the Helm maintainers

mattfarina:
These maintainers have made tremendous contributions but are currently inactive. I have connected with each of them and they are focused on other work, at present. They all confirmed moving to an emeritus status. If they ever want to return to active development/maintainer-ship of Helm, I would be delighted.

The governance has a process for inactive folks and, as there are items that need to be voted on, inactive maintainer make decision making difficult.

These maintainers are and will continue to be missed.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,19 +1,18 @@
 maintainers:
-  - adamreese
-  - bacongobbler
   - hickeyma
   - jdolitsky
   - marckhouzam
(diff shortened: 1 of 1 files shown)


sabre1041 (approved):
LGTM

technosophos (approved):
LGTM

PR #12045, opened 2023-05-02
add govulncheck to CI

joejulian:
We get a lot of false-positives from users running security scanners that are incorrectly identifying upstream packages as being relevant. This would give us a reference to show that those issues are not valid.

**If applicable**:

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -30,6 +30,11 @@ jobs:
           GOLANGCI_LINT_SHA256: '4de479eb9d9bc29da51aec1834e7c255b333723d38dbd56781c68e5dddc6a90b'
       - name: Test style
         run: make test-style
+      - name: Scan for Vulnerabilities
+        uses: Templum/govulncheck-action@6bb063b41d78c53c2fc7f5589828e30ad7c697ee # pin@v1.0.0
+        with:
+          skip-upload: true
+          go-version: '1.20'
       - name: Run unit tests
         run: make test-coverage
       - name: Test build


yxxhero (approved):

.github/workflows/build-test.yml
```diff
@@ -30,6 +30,11 @@ jobs:
           GOLANGCI_LINT_SHA256: '4de479eb9d9bc29da51aec1834e7c255b333723d38dbd56781c68e5dddc6a90b'
       - name: Test style
         run: make test-style
+      - name: Scan for Vulnerabilities
+        uses: Templum/govulncheck-action@v0.10.1
```
mattfarina, line 34: Our actions are pinned to a revision. For example,

```
actions/setup-go@6edd4406fa81c3da01a34fa6f6343087c207a568 # pin@3.5.0
```

Can you pin this one?

mattfarina (changes requested):

scottrigby (approved):
This lgtm. I havne't used govulncheck until today. Thanks @gjenkins8 for pointing it out.

Speaking of:
> One question -- does this introduce any blocking CI checks? One pitfall with CVE checks is blocking PRs unrelated to an existing, but then discovered CVE. IMHO only PRs that introduce a CVE should be blocked. CVEs which already exist in the code should be reported (of course).

@gjenkins8 we can dismiss checks if the code found is unrelated to the PR, yes?

PS @mattfarina's requested change was made (version is now pinned) https://github.com/helm/helm/pull/12045#pullrequestreview-1444273920

scottrigby (commented):
Backing out of my approval for the moment because - while this seems useful to me - a bit more discussion may help. I checked in with @gjenkins8 to better understand the concern. I'll let him describe his point of view.

One additional note though, we may want to check in with the Minio folks to ask their experience. They've been using govulncheck in CI for 2 years. While infrequent, it seems from a search that this has come in handy. Here is an example comment that I think exemplifies @joejulian 's initial motivation for this PR: https://github.com/minio/minio/issues/17730#issuecomment-1652599721

PR #12056, opened 2023-05-05
update autoscaling/v2beta1 to autoscaling/v2 in skeleton chart

kupnu4x:
Update hpa.yaml in templates to support autoscaling/v2 instead of deprecated autoscaling/v2beta1


joejulian (approved):

mattfarina:
Noting the bug because the current resource, before the change, is no longer served by k8s versions this version of Helm supports.

mattfarina (approved):

Issue #11994, opened 2023-04-13
Deprecate `github.com/pkg/errors` in favor of stdlib


komish:
The https://github.com/pkg/errors project has been archived for some time in favor of the proposals that landed in the stdlib. Just taking a quick look, seems like `errors` and `fmt` should be able to handle most of the use cases that Helm uses. I'm actively doing this for a tool that depends on the helm libraries which drew my attention to its use here.

Happy to help contribute swapping those out, but I want to make sure the project would be interested.

yxxhero:
@komish PR is welcome.

jgsheppa:
@komish @yxxhero Do you mind if I create a PR for this?

jgsheppa:
Hm, now that I've looked into this a bit more, it seems like the main benefit of using `github.com/pkg/errors` are the stack traces which are returned by `Errorf`, `Wrap`, and `Wrapf`. For instance, it's possible to replace `errors.Errorf` with `fmt.Errorf` without trouble from the type system, but you lose the stack trace. I'm guessing these stack traces are pretty important @yxxhero? If so, then the PR I drafted might not be of much use, but if it's fine to lose the stack traces from some of these calls, then we could move forward.

komish:
@jgsheppa apologies, I totally missed notification for this issue. Thanks for taking care of it!

SD007CSE:
Anyone working on this issue?

SD007CSE:
@jgsheppa Are you working on this issue?

jgsheppa:
> @jgsheppa Are you working on this issue?
> 

As far as I know, it is still being discussed by the team.

mitar:
Shameless plug: you can also switch it to drop-in replacement [`gitlab.com/tozd/go/errors`](https://gitlab.com/tozd/go/errors). It fixes [many issues](https://gitlab.com/tozd/go/errors#why-a-new-go-errors-package), is maintained, and supports modern Go's error patterns (sentinel errors, %w formatting, joined errors, etc.). It also provides some nice utility functions and structured details so that it is easy to extract dynamic data out of errors (instead of trying to get them out of formatted strings). Has improved error formatting and JSON marshaling of errors. It is interoperable with other errors packages and does not require a dependency on itself to extract data (e.g., stack trace) from errors.

HosseinMirjalali:
This is a duplicate of #13460 and is fixed, right? (I'm going through good-first-issue issues)  @scottrigby

komish:
Yup. Closing.

PR #12063, opened 2023-05-07
Use standard lib `errors` to replace most uses of `github.com/pkg/errors`

jgsheppa:
This PR refs #11994 and removes the package `github.com/pkg/errors` from many parts of the code base, aside from the instances of `errors.Wrap`, `errors.Wrapf`, and `errors.Cause` from `github.com/pkg/errors`. I'm unsure how the Helm team might want to solve the replacement of `Wrap(f)`, so maybe we can discuss that in another issue. 

1. Replaced `githubErrors.Errorf` with `fmt.Errorf`
2. Replaced `githubErrors.New` with `errors.New`
3. Replaced `githubErrors.Is` with `errors.Is`

**If applicable**:

### DIFF
--- a/cmd/helm/docs.go
+++ b/cmd/helm/docs.go
@@ -22,7 +22,6 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/spf13/cobra/doc"
 	"golang.org/x/text/cases"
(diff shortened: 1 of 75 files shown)


jgsheppa:
@gjenkins Thanks for the update! After creating this PR I set it to draft because the problem seemed more complex than I originally anticipated. I look forward to hearing what the team has to say!

Issue #12086, opened 2023-05-18
performInstall go routine continues running after context is cancelled


anr-tonic:
[Small reproduction and copy of this report here](https://github.com/anr-tonic/helm-repro/tree/main)

# Background

We have an installer that uses helm as an SDK to install several charts needed
for our product to work. We noticed that after we implemented functionality to
abort an installation that we would still receive output to stdout from helm
after the context provided to `RunWithContext` is canceled. In our case, it
appears that it is because use the helm provided wait functionality.

In this report, I'm picking on `action.Install` as its where I've spent the
most time, however `action.Upgrade` also exposes a `RunWithContext` method that
appears to have similar characteristics to Install's.

While it is possible for our installer to work around this issue by
implementing our own wait logic, there are other avenues to trigger this
behavior:

* Requiring post-install hooks to run after chart resources have become
  available
* pre-install hook that does not become ready before the context is cancelled
  by the caller

Theoretically, a pre-install hook could run for timeout - 1, the context would
cancel, but the creation of k8s resources and wait could still happen until the
timeout expires. I haven't attempted to contrive a scenario to examine what
would actually happen in this case.

# Cause

Even though `RunWithContext` will return control to the caller after its
context has been canceled, the `performInstall` go routine is still running in
the background because it does not participate in using the context passed to
`RunWithContext`.

Methods on `kube.Interface` also do not participate in using a context. When
calling `kube.Client` methods that invoke a method on
`k8s.io/client-go/kubernetes/ClientSet` that requires a context, it uses
`context.Background()`, often combined with `context.WithTimeout` from a
`time.Duration` that has been passed down.

# Possible solution

Similar to other extensions to `kube.Interface` that seek to avoid breaking
backwards compatibility, an additional interface could be created that exposes
methods that accept `context.Context` instead of `time.Duration`. The
implementations of the existing non context aware methods then delegate to
implementations of this interface. This would additionally require methods on
types like `kube.waiter` to accept this context instead of constructing one
from the background context and a timeout value. Roughly:

```go
type InterfaceWithContext interface {
    WaitWithContext(context.Context, ResourceList) error
    WatchUntilReadyWithContext(context.Context, ResourceList) error
    WaitAndGetCompletedPodPhaseWithContext(context.Context, string) (v1.PodPhase, error)
    WaitForDeleteWithContext(context.Context, ResourceList) error
}

func (c *Client) Wait(resources ResourceList, timeout time.Duration) {
    return c.WaitWithContext(
[...]

PR #12088, opened 2023-05-18
Fix leaking goroutines in Install

### LINKED ISSUE #11971: Install:RunWithContext function has memory/goroutine leak when context is canceled
Output of `helm version`:

Output of `kubectl version`: 
Client Version: v1.24.0
Kustomize Version: v4.5.4
Server Version: v1.24.10

In "RunWithContext" method of "*Install" structure, there is the following code snippet (line 349-356):

	rChan := make(chan resultMessage)
	doneChan := make(chan struct{})
	defer close(doneChan)
	go i.performInstall(rChan, rel, toBeAdopted, resources)
	go i.handleContext(ctx, rChan, doneChan, rel)
	result := <-rChan
	//start preformInstall go routine
	return result.r, result.e

The problem with this is that when context is canceled, the goroutine handleContext will send the result on rChan and the result will be received and returned.
But after finishing its tasks, the goroutine performInstall, will also try to send result on rChan channel.
The reader of that channel has already received the result and returned from the function. This will cause the goroutine performInstall to be blocked on sending result to the channel rChan forever, as there is no reader.
Thus, entering a deadlock and ultimately leaking memory.

The channel should be a buffered one to fix this memory leak.

Evidence to support my claim:

 1. Below is the sample code that I stubbed out to do memory profiling of underlying helm library code. It mimicks the how the channels are being used.
```
 // basic is a program used to define a goroutine leak.
package main

import (
	"context"
	"fmt"
	"os"
	"runtime"
	"runtime/pprof"
	"time"
)

func main() {
	f1, _ := os.Create("/tmp/profile3.pb.gz")
	defer f1.Close()
	pprof.WriteHeapProfile(f1)
	startingGs := runtime.NumGoroutine()
	ctx, _ := context.WithTimeout(context.Background(), 5*time.Second)
	rChan := make(chan int)
	doneChan := make(chan int)
	defer close(doneChan)
	go performInstall(rChan)
	go handleContext(ctx, rChan, doneChan)
	result := <-rChan
	fmt.Println(result)

	endingGs := runtime.NumGoroutine()
	fmt.Println("========================================")
	fmt.Println("Number of goroutines before:", startingGs)
	fmt.Println("Number of goroutines after :", endingGs)
	pprof.Lookup("goroutine").WriteTo(os.Stdout, 1)
	f2, _ := os.Create("/tmp/profile4.pb.gz")
	defer f2.Close()
	runtime.GC()
	//pprof.Lookup("heap").WriteTo(os.Stdout, 1)
	pprof.WriteHeapProfile(f2)
	fmt.Println("Number of goroutines leaked:", endingGs-startingGs)
}

func performInstall(c chan<- int) {
	fmt.Println("hello from performInstall")
	time.Sleep(25 * time.Second)
	reportToRun(c)
	fmt.Println("bye from performInstall")
}

func handleContext(ctx context.Context, c chan<- int, done chan int) {
	fmt.Println("hello from handleContext")
	select {
	case <-ctx.Done():
		fmt.Println("calling reportToRun from handleContext")
		reportToRun(c)
	case <-done:
		fmt.Println("bye from handleContext")
		return
	}
}

func reportToRun(c chan<- int) {
	fmt.Println("hello from reportToRun")
	c <- 5
	fmt.Println("bye from reportToRun")
}
```

 
2. Attaching corresponding memory profile(profile3.pd.gz corresponds to heap before and profile4.pd.gz corresponds to heap after)

[profile3.pb.gz](https://github.com/helm/helm/files/11175178/profile3.pb.gz)
[profile4.pb.gz](https://github.com/helm/helm/files/11175179/profile4.pb.gz)

3. Attaching snippets corresponding to profile3.pd.gz and profile4.pd.gz

profile003
profile004

mslapek:

Closes  #11971

No public API was changed.

Earlier `failRelease` could be called twice (when `ctx` was cancelled and the background job had an error).
Now `failRelease` can be called at most once - synchronized with a channel (instead of a mutex lock).

Added comment with true behavior - running the job in the background after cancel is a bit surprising...

Done small refactoring. All fan-in logic is in a small function.

**If applicable**:

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -219,6 +219,9 @@ func (i *Install) Run(chrt *chart.Chart, vals map[string]interface{}) (*release.
 }
 
 // Run executes the installation with Context
+//
+// When the task is cancelled through ctx, the function returns and the install
+// proceeds in the background.
 func (i *Install) RunWithContext(ctx context.Context, chrt *chart.Chart, vals map[string]interface{}) (*release.Release, error) {
 	// Check reachability of cluster unless in client-only mode (e.g. `helm template` without `--validate`)
 	if !i.ClientOnly {
@@ -385,17 +388,31 @@ func (i *Install) RunWithContext(ctx context.Context, chrt *chart.Chart, vals ma
 		// not working.
 		return rel, err
 	}
-	rChan := make(chan resultMessage)
-	ctxChan := make(chan resultMessage)
-	doneChan := make(chan struct{})
-	defer close(doneChan)
-	go i.performInstall(rChan, rel, toBeAdopted, resources)
-	go i.handleContext(ctx, ctxChan, doneChan, rel)
+
+	rel, err = i.performInstallCtx(ctx, rel, toBeAdopted, resources)
+	if err != nil {
+		rel, err = i.failRelease(rel, err)
+	}
+	return rel, err
+}
+
+func (i *Install) performInstallCtx(ctx context.Context, rel *release.Release, toBeAdopted kube.ResourceList, resources kube.ResourceList) (*release.Release, error) {
+	type Msg struct {
+		r *release.Release
+		e error
+	}
+	resultChan := make(chan Msg, 1)
+
+	go func() {
+		rel, err := i.performInstall(rel, toBeAdopted, resources)
+		resultChan <- Msg{rel, err}
+	}()
 	select {
-	case result := <-rChan:
-		return result.r, result.e
-	case result := <-ctxChan:
-		return result.r, result.e
+	case <-ctx.Done():
+		err := ctx.Err()
+		return rel, err
+	case msg := <-resultChan:
+		return msg.r, msg.e
 	}
 }
 
@@ -407,13 +424,12 @@ func (i *Install) isDryRun() bool {
 	return false
 }
 
-func (i *Install) performInstall(c chan<- resultMessage, rel *release.Release, toBeAdopted kube.ResourceList, resources kube.ResourceList) {
-
+func (i *Install) performInstall(rel *release.Release, toBeAdopted kube.ResourceList, resources kube.ResourceList) (*release.Release, error) {
+	var err error
 	// pre-install hooks
 	if !i.DisableHooks {
 		if err := i.cfg.execHook(rel, release.HookPreInstall, i.Timeout); err != nil {
-			i.reportToRun(c, rel, fmt.Errorf("failed pre-install: %s", err))
-			return
+			return rel, fmt.Errorf("failed pre-install: %s", err)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


gjenkins8 (commented):
To me (community member; not a helm maintainer), this looks like a fairly nice change generally. It refactors and improves significant amount of messy go-routine and error handling/synchronization logic. Plus looks to fix the mentioned go-routine leak (bug).

I do think this PR somewhat exposes a larger issue, cancelling the context doesn't really 'cancel' the install (or upgrade). It does "fail" the install/upgrade  (calls the respective `failRelease`), but perhaps worse, likely the install/upgrade is failed whilst the install/update is still (likely) running. As my comment, I think this is a seperate issue from this PR.

I will raise both this PR, and the async/cancel issue in the helm weekly developer call.

pkg/action/install.go
```diff
@@ -211,6 +211,9 @@ func (i *Install) Run(chrt *chart.Chart, vals map[string]interface{}) (*release.
 }
 
 // Run executes the installation with Context
+//
+// When the task is cancelled through ctx, the function returns and the install
```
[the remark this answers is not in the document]
mslapek: > Completely agree, this is surprising.

I was myself surprised with that - so added a comment for other devs. 😉

> And I think worth discussing further with helm maintainers.

Looks like the discussion started in issue https://github.com/helm/helm/issues/12086

pkg/action/install.go
```diff
@@ -368,63 +371,76 @@ func (i *Install) RunWithContext(ctx context.Context, chrt *chart.Chart, vals ma
...
+	}()
+	go func() {
+		select {
+		case <-ctx.Done():
+			err := ctx.Err()
+			resultChan <- Msg{rel, err}
+		case <-doneChan:
+			return
+		}
+	}()
```
mattfarina, line 403: When the context is canceled the second goroutine will exit but it looks like the first will still run until `performInstall` reaches completion. Am I missing something?

I wonder if `performInstall` needs to be modified to accept a context and regularly check if it's done to exit.

mslapek: > When the context is canceled the second goroutine will exit but it looks like the first will still run until performInstall reaches completion.

This is right - it was the original behavior even before this PR. This was captured by the added comment to the function. I think it's a technical debt.

> I wonder if performInstall needs to be modified to accept a context and regularly check if it's done to exit.

Yes. The API communicating with K8s (like `KubeClient.WaitWithJobs`) would need to take context instead of timeout, some people consider this in https://github.com/helm/helm/issues/12086

I think that this PR might be a good starting point to fix the mentioned issue (separetes the code into `releasingUpgradeCtx` and `performInstallCtx`).

scottrigby:
I pinged @hiddeco to see if he can review this as a heavy Helm SDK user. I recall goroutines and client workarounds were needed in Flux's use of Helm SDK.

pkg/action/install.go
hiddeco, line 403: Think this whole bit (and below) is an obsolete routine? AFAIK, the following should work without any issues while yielding the same behavior and getting rid of a `chan`:

```go
select {
case <-ctx.Done():
	return rel, ctx.Err()
case msg := <-resultChan:
	return msg.r, msg.e
}

File: pkg/action/install.go (hunk 1 above)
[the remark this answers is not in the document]
Comment (hiddeco) [reply pkg/action/install.go:215]: :scream: we missed this as our whole container gets killed in scenarios where a context is cancelled, but I would definitely like to see this addressed as well.

File: pkg/action/install.go (hunk 2 above)
Comment (gjenkins8) [reply pkg/action/install.go:403]: I think this PR (as is) fixes lost goroutines in performInstall/Upgrade.

https://github.com/helm/helm/issues/12086 (and associate PR: https://github.com/helm/helm/pull/12109) might be a better place to tackle synchronizing install.upgrades to not continue to run in the background?

Comment (gjenkins8) [reply pkg/action/install.go:403]: I think this PR (as is) fixes lost goroutines in performInstall/Upgrade.

https://github.com/helm/helm/issues/12086 (and associate PR: https://github.com/helm/helm/pull/12109) might be a better place to tackle synchronizing install.upgrades to not continue to run in the background?

**Comment by joejulian** [comment]
Is there any way to add regression tests for this?

**Comment by tsuna** [comment]
Is this holding up the 3.12.2 release?

**Comment by mslapek** [comment]
## Recent changes

* limited the scope of this PR to Install action only (was Install and Upgrade previously)
* added unit-tests checking for leaking goroutines @joejulian
* applied a hint from @hiddeco (thanks!)

## Tests

Added to a few tests inspection of [`runtime.NumGoroutine()`](https://pkg.go.dev/runtime#NumGoroutine).

They nicely fail on main:

```
$ go clean -testcache && go test ./pkg/action

--- FAIL: TestInstallRelease_Wait (0.00s)
    install_test.go:380: 
                Error Trace:    /home/awesome/repo/helm/pkg/action/install_test.go:380
                Error:          Not equal: 
                                expected: 2
                                actual  : 3
                Test:           TestInstallRelease_Wait
--- FAIL: TestInstallRelease_Wait_Interrupted (11.01s)
    install_test.go:405: 
                Error Trace:    /home/awesome/repo/helm/pkg/action/install_test.go:405
                Error:          Not equal: 
                                expected: 2
                                actual  : 3
                Test:           TestInstallRelease_Wait_Interrupted
```

(the tests show one additional running goroutine - the leaking one!)

Of course, they should succeed on this branch. 😉

joejulian (approved):
Nice! Thanks

hiddeco (approved):

mattfarina (approved):

mattfarina:
@mslapek can you rebase this. When I went to merge it there was a conflict.

mslapek:
> @mslapek can you rebase this. When I went to merge it there was a conflict.

Rebased. The conflict was a trivial new-line clash of added isDryRun from main.

joejulian (approved):

mattfarina (approved):

mslapek:
@joejulian @mattfarina @hiddeco Thanks for review! 🍭

Issue #9761, opened 2021-06-01
helm.sh/helm/v3 3.6.0 context deadline exceeded


manuelmazzuola:
Output of `helm version`: not using

Output of `kubectl version`: not using

Cloud Provider/Platform (AKS, GKE, Minikube etc.):  AKS

Hello, I'm using the golang library to programmatically deploy helm charts on azure.  
I get an error using the new `3.6.0` release, while installing a chart the outcome is: `context deadline exceeded`.  
Just that, I can't figure out why.  
It happens with any chart.  

The version I used before, `3.5.4`, works fine, just the `3.6.0` version gives me that error.

```go
client := action.NewInstall(actionConfig)
client.Namespace = namespace
client.ReleaseName = chartName
client.CreateNamespace = true
client.Atomic = true

_, err := client.Run(chartReq, chartValues);

```
![image](https://user-images.githubusercontent.com/1757492/120338197-4af04080-c2f4-11eb-885a-aff4468c7820.png)

invidian:
This seems to be caused by https://github.com/helm/helm/commit/98d98369ed9abcaec9b75352e07dd6f759a8804f#diff-0620aa4a73421adf60e6357e7543130a651180fa27bf53472280d5095eec313dR49-R50.

Timeout must be now specified explicitly, otherwise the operation timeouts immediately. Seems like a bug to me, as zero time for timeout should be validated and rejected.

bacongobbler:
@joelanford, do you have time to look into this?

joelanford:
Yep! Will do.

joelanford:
Did a little digging. It looks like the culprit is the use of the passed context in ReadyChecker's new `IsReady` method. Before, we called the underlying `kubernetes.Interface` methods with `context.Background()`. Now, we use the passed context, which will always cause a deadline exceeded error if the timeout is 0.

Just for completeness sake, I also double-checked the change from `wait.Poll()` to `wait.PollImmediateUntil()`. It looks like in either case, the passed function is called at least once, so that should have no impact on the bug.

Another piece of context is that command line users will only experience this bug if they explicitly set the timeout to 0 (the default is 5m0s).

Not sure what folks think the most appropriate fix is here. One option would be to revert to `context.TODO()` and add a comment/note with a link to this issue explaining why its set that way and that it should be changed for Helm v4.

> Seems like a bug to me, as zero time for timeout should be validated and rejected.

I'm not sure we can do this until v4 because that sounds like a breaking change.

[...]

PR #12118, opened 2023-06-02
Increase timeout for operations per second

LucasSaldanha21:

**If applicable**:

### DIFF
--- a/pkg/kube/wait.go
+++ b/pkg/kube/wait.go
@@ -50,7 +50,7 @@ func (w *waiter) waitForResources(created ResourceList) error {
 	ctx, cancel := context.WithTimeout(context.Background(), w.timeout)
 	defer cancel()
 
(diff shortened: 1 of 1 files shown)


yxxhero:
@BlazeF35 maybe we can make it configurable. CLI flag or other methods.

BlazeF35:
yes, it would be very helpful

yxxhero:
@BlazeF35 I think so. I don't like fixed value.

felipecrs:
BTW I did some tests recently with this PR (which is based on https://github.com/felipecrs/helm/pull/2), and I could no longer find a situation in which it helped.

In fact, I don't even remember why I did it (but probably it was an attempt to solve https://github.com/helm/helm/issues/9761), and I don't even trust myself that it provided any help back then.

My advice would be to close it indeed, unless we find a way to reproduce a situation in which increasing these timeouts actually improves something.

PR #12126, opened 2023-06-08
Allow setting namespace in `helm template` without a kubeconfig

### LINKED ISSUE #12124: helm template inconsistently respects the --namespace option in different environments
## Expected behavior:
When running `helm template` with the `--namespace` argument, it should set the `.Release.Namespace` variable, and any templates using that variable should be set to the value of the `--namespace` argument.

## Actual behavior:
It is inconsistent about setting the `.Release.Namespace` variable. On my and my coworker's laptops, it works as expected. In our CI, it doesn't set it to the value of the `--namespace` argument, but instead just defaults to `default`.

It works on our laptops with or without docker, but in CI it doesn't. The same versions of `helm` and `kubectl` are used on my laptop and in CI. The inconsistency persists even when running locally in the same Docker image used by CI (so the python versions are also the same).

We've looked very hard for differences in our environments, and can find none, which leaves me to suspect helm itself is inconsistent.

## Additional context:
Specifically, we're attempting to template the aws-ebs-csi-driver chart, version 2.19.0, from https://kubernetes-sigs.github.io/aws-ebs-csi-driver .

We're calling `helm template` in a python script pointing at a locally vendored copy of the chart. Values are passed to stdin.
```python3
values = {
    "controller": {
        "serviceAccount": {
            "create": True,
        },
        "nodeSelector": {
            "workload": "mz-system",
        },
    },
    "node": {
        "affinity": {
            "nodeAffinity": {
                "requiredDuringSchedulingIgnoredDuringExecution": {
                    "nodeSelectorTerms": [
                        {
                            "matchExpressions": [
                                {
                                    "key": "eks.amazonaws.com/compute-type",
                                    "operator": "NotIn",
                                    "values": ["fargate"],
                                },
                                {
                                    "key": "workload",
                                    "operator": "NotIn",
                                    "values": ["materialize-egress"],
                                },
                            ],
                        },
                    ],
                },
            },
        },
    },
}

args = [
    "helm",
    "template",
    "aws-ebs-csi-driver",
    "infra/helm/vendor/charts/aws-ebs-csi-driver/2.19.0",
    "-f",
    "-",
    "--namespace",
    "kube-system",
]

rendered_text = subprocess.run(
    args,
    input=yaml.safe_dump(values),
    check=True,
    text=True,
    capture_output=True,
).stdout
```

An example template in the chart which is rendered differently in CI vs locally (the namespace is `kube-system` locally, but in CI is rendered to `default`):
```yaml
{{- if .Values.controller.serviceAccount.create -}}
apiVersion: v1
kind: ServiceAccount
metadata:
  name: {{ .Values.controller.serviceAccount.name }}
  namespace: {{ .Release.Namespace }}
  labels:
    {{- include "aws-ebs-csi-driver.labels" . | nindent 4 }}
  {{- with .Values.controller.serviceAccount.annotations }}
  annotations:
    {{- toYaml . | nindent 4 }}
  {{- end }}
  {{- if eq .Release.Name "kustomize" }}
  #Enable if EKS IAM roles for service accounts (IRSA) is used. See https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html for details.
  #annotations:
  #  eks.amazonaws.com/role-arn: arn:<partition>:iam::<account>:role/ebs-csi-role
  {{- end }}
{{- end -}}
```

### Output of `helm version`:
```
version.BuildInfo{Version:"v3.12.0", GitCommit:"c9f554d75773799f72ceef38c51210f1842a1dea", GitTreeState:"clean", GoVersion:"go1.20.3"}
```

### Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"25", GitVersion:"v1.25.9", GitCommit:"a1a87a0a2bcd605820920c6b0e618a8ab7d117d4", GitTreeState:"clean", BuildDate:"2023-04-12T12:16:51Z", GoVersion:"go1.19.8", Compiler:"gc", Platform:"linux/amd64"}
```
Server version is not relevant to `helm template`.

### Cloud Provider/Platform (AKS, GKE, Minikube etc.): 
N/A, but our CI is running in Github Actions and we're running `helm template` in a Docker container.

alex-hunt-materialize:

If you don't have a working kubeconfig, `helm template` does not respect the `--namespace` option, instead silently ignoring it and using "default". It gets set on the settings object, but does not get returned by the `Namespace()` method. `helm template` should not require a working kubeconfig or any communication with a kubernetes cluster, as it is only generating templates.

This PR adds a fallback to return the already set namespace field, rather than ignoring the user's intentions and returning "default".

Closes https://github.com/helm/helm/issues/12124

**If applicable**:


joejulian (changes requested):
After giving this some thought, this seems correct. There is another flag in this struct that overrides the environment variable. https://github.com/helm/helm/blob/f8da9276d8ed30480b2ffc3ce92037e8b3651098/pkg/cli/environment.go#L106

We can't do `namespace` the same way, though, because the kubeconfig flags will need parsed to get the default namespace from the kubeconfig (if set and env is not). This seems like the appropriate place to process this override.

I would like to see this added to the unit test though.

CalvinKrist:
@joejulian I copied their fix and added a unit test here https://github.com/helm/helm/pull/12979. I confirmed that without the fix being present the test fails.

PR #12129, opened 2023-06-09
Allow configuring custom ResourceVisitor in Install and Upgrade actions

luksa:
Adds the ability to set a custom resource visitor function to the Install and Upgrade actions when Helm is used as a library. 

This feature can be used to modify the resources (e.g. add an `ownerReference`). Although this could be done using a PostRenderer, this approach is not ideal, because it requires all the manifests to be parsed and then reserialized. Given that Helm already parses the manifests to process the resources with the `setMetadataVisitor` function, it makes sense to allow clients to register their own additional visitor. 

No unit test, because the fake KubeClient used in the tests never produces any resource objects. For the same reason, there is no existing test that would check if the `setMetadataVisitor` is applied to the resources.

**If applicable**:

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -104,6 +104,12 @@ type Install struct {
 	// OutputDir/<ReleaseName>
 	UseReleaseName bool
 	PostRenderer   postrender.PostRenderer
+	// ResourceVisitor is an optional VisitorFunc applied to each rendered resource
+	//
+	// This is an alternative to PostRenderer. Unike PostRenderer, it allows each
+	// individual resource to be modified without having to parse and re-serialize the
+	// manifests.
+	ResourceVisitor resource.VisitorFunc
 	// Lock to control raceconditions when the process receives a SIGTERM
 	Lock sync.Mutex
 }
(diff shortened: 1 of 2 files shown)


jwendell:
Hi. Can we have a review on this one?

amannijhawan:
Can you add some tests and maybe some code  examples that show how this works ? That we can use to test on our own?

luksa:
@amannijhawan This is actually hard to test, because it requires `Install.cfg.KubeClient` to return an array of `resource.Info`s, which the fake client does not. I tried implementing such a client, but then ran into other issues because the `Install.Run()` function then runs a bunch of other actions on this client. Note that the function that calls this external resource visitor already calls the `setMetadataVisitor` that also has no tests.

The purpose of this change is to add allow us to make additional changes to each rendered resource. This is very similar to how the `setMetadataVisitor` is used on the rendered resources [here](https://github.com/helm/helm/pull/12129/files#diff-4a0aea3959569dfbc7af305435031b78827ed0878e9174e019b85916cd153861R302). 

For example, we use this to add an ownerReference to resources. We register the visitor [here](https://github.com/Maistra/istio-operator/blob/602fdcded70f5af0a68d2eafd67cd2ace2196a37/pkg/helm/charts.go#L118). And [here](https://github.com/maistra/istio-operator/blob/602fdcded70f5af0a68d2eafd67cd2ace2196a37/pkg/helm/ownerreferences.go#L20-L43) is the visitor implementation.

jwendell:
ping

PR #12135, opened 2023-06-11
Support for using custom template functions when helm is used as a library

alexwo:

#### Motivation

Helm is widely used as a package manager for Kubernetes, but it can also be leveraged as a library for programmatic use. This PR aims to enhance Helm's flexibility by enabling users to define and configure their own custom template functions when using Helm programmatically.

#### Changes Made
- Added new engine config param `WithCustomFuncs` to the Engine package.

#### Usage Example

```go
e := Engine{
        WithCustomFuncs: func(funcs template.FuncMap) template.FuncMap {
            funcs["myFunc"] =  func(name string) string {return fmt.Sprintf("%s-from-custom-func", name)} 
            // Add other custom functions as needed
            return funcs
        },
    }
```

Link to issue:
https://github.com/helm/helm/issues/12138

**If applicable**:

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -44,6 +44,11 @@ type Engine struct {
 	config *rest.Config
 	// EnableDNS tells the engine to allow DNS lookups when rendering templates
 	EnableDNS bool
+	// WithCustomFuncs allows you to provide a template.FuncMap containing custom template functions to be added or
+	// overridden. It is typically used as a configuration option to extend the default set of template functions with
+	// custom ones. The provided template.FuncMap can extend the default template functions, and the resulting
+	// FuncMap is used by the template engine.
+	WithCustomFuncs func(funcs template.FuncMap) template.FuncMap
 }
 
 // New creates a new instance of Engine using the passed in rest config.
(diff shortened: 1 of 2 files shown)


bennesp:
Maybe @mattfarina or @joejulian can have a look at this PR? This looks very useful for any tooling around helm
Thank you very much 🙏

bennesp:
Apologies for the notification, I would just kindly know if there is something we can do to help having this PR reviewed. If so, I am willing to help. Thank you very much, I really appreciate your work @mattfarina @scottrigby

MikaelElkiaer:
This would be really useful for what I describe here: https://github.com/helm/community/issues/340

And, in any case where you are "forced" to use an `include` when piping functions.
E.g. a bunch of the named templates from https://github.com/bitnami/charts/tree/main/bitnami/common could be turned into custom functions.

gjenkins8 (dismissed):
I still have the concern I listed here: https://github.com/helm/helm/issues/12138#issuecomment-1682438450

> Any chart which depends on a specialized function would only be usable from a (custom) SDK operator. And not from any standard Helm CLI/SDK (worth considering the downsides here when it comes to development and usage)

(The author never responded there; and the linked PR never progressed)

But if we accept that this will only be used for "internal" charts, for an internal purpose. And given the minimal intrusiveness of this PR. I think it is okay to proceed.

PR #12151, opened 2023-06-16
Add json support for index.yaml file #10542

### LINKED ISSUE #10542: index.json - necessary for large projects
There are multiple issues reported regarding performance problems when the index.yaml size starts to increase:
https://github.com/helm/helm/issues/6155
https://github.com/helm/helm/issues/9931
https://github.com/helm/helm/issues/9865 (not related to the index file being .yaml, but still related)

@mattfarina has created a benchmark showing that JSON parsing is an order of magnitude faster: https://github.com/mattfarina/yamlbench
@mattfarina also created https://github.com/mattfarina/helm-index-json to show some JSON split testing.

As more and more indices are growing in size, this problem is bound to happen to more and more users.

Would it be due time to look into placing a .json index file next to the .yaml file, so that Helm opts to use the .json file but falls back to .yaml if it does not exist?

Just the switch from YAML -> JSON should help significantly in this area. JSON is also more natural for computer-written data as opposed to YAML which might be more suited for data where users also at times edit the files manually.

smmhatami:
This is a fix to solve Issue 10542. Resolves https://github.com/helm/helm/issues/10542. With these changes one can use json formatted index file instead of yaml. the code now looks for index.json file as default and switches to index.yaml if not found.

### DIFF
--- a/pkg/repo/chartrepo.go
+++ b/pkg/repo/chartrepo.go
@@ -98,7 +98,7 @@ func (r *ChartRepository) Load() error {
 	// what repos to use?
 	filepath.Walk(r.Config.Name, func(path string, f os.FileInfo, err error) error {
 		if !f.IsDir() {
-			if strings.Contains(f.Name(), "-index.yaml") {
+			if strings.Contains(f.Name(), "-index.json") || strings.Contains(f.Name(), "-index.yaml") {
 				i, err := LoadIndexFile(path)
 				if err != nil {
 					return err
@@ -115,12 +115,16 @@ func (r *ChartRepository) Load() error {
 
 // DownloadIndexFile fetches the index from a repository.
 func (r *ChartRepository) DownloadIndexFile() (string, error) {
-	indexURL, err := ResolveReferenceURL(r.Config.URL, "index.yaml")
+	jsonIndexURL, err := ResolveReferenceURL(r.Config.URL, "index.json")
+	if err != nil {
+		return "", err
+	}
+	yamlIndexURL, err := ResolveReferenceURL(r.Config.URL, "index.yaml")
 	if err != nil {
 		return "", err
 	}
 
-	resp, err := r.Client.Get(indexURL,
+	resp, err := r.Client.Get(jsonIndexURL,
 		getter.WithURL(r.Config.URL),
 		getter.WithInsecureSkipVerifyTLS(r.Config.InsecureSkipTLSverify),
 		getter.WithTLSClientConfig(r.Config.CertFile, r.Config.KeyFile, r.Config.CAFile),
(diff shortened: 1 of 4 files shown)


yxxhero:
@smmhatami please fix the DCO issue. maybe it's a breaking change?

gausk:
Changes LGTM

pkg/repo/index.go
```diff
@@ -331,13 +334,25 @@ func IndexDirectory(dir, baseURL string) (*IndexFile, error) {
 // This will fail if API Version is not set (ErrNoAPIVersion) or if the unmarshal fails.
 func loadIndex(data []byte, source string) (*IndexFile, error) {
 	i := &IndexFile{}
+	isjson := strings.HasSuffix(source, ".json")
```
hiddeco, line 337:
```suggestion
	var isJSON bool
	if fileExt := filepath.Ext(source); fileExt == ".json" {
		isJSON = true
	}
```

pkg/repo/chartrepo.go
```diff
@@ -115,12 +115,16 @@ func (r *ChartRepository) Load() error {
...
+	if err != nil {
+		return "", err
+	}
+	yamlIndexURL, err := ResolveReferenceURL(r.Config.URL, "index.yaml")
 	if err != nil {
 		return "", err
 	}
 
-	resp, err := r.Client.Get(indexURL,
+	resp, err := r.Client.Get(jsonIndexURL,
```
hiddeco, line 127: If the file does not exist (HTTP 404), this would return an error and it would not fallback to loading the `index.yaml`. See also https://github.com/helm/helm/blob/4e447d87cd3222aa5e5ea2b38170e9d80b99c0b4/pkg/getter/httpgetter.go#L91-L93

In addition, on a successful retrieval any parsing error would be swallowed by an attempt to fetch and load the `index.yaml` (which may not exist). Yielding absolutely no information to the user to figure out what went wrong.

Given this, I think it would be better to not perform a fallback based on loading errors. But to rather fall back based on the error from `r.Client.Get`, which may require introducing a typed error in `httpgetter.go` to distinguish a HTTP 404 from transient errors.

hiddeco:
In addition, this lacks any configuration options to generate an `index.json`. For this to happen, and as requested in https://github.com/helm/helm/issues/10542#issuecomment-1433878292. I would advise to add the following things:

- Add a flag here: https://github.com/helm/helm/blob/d6708667da5c79ab4a093ea7473e96b93be5f1f7/cmd/helm/repo_index.go#L70-L72
- Add a `WriteJSONFile` here (to not break backwards compatibility): https://github.com/helm/helm/blob/4e7e939f19719e056c0f6e47b02068bdca6f420c/pkg/repo/index.go#L233
- Conditionally use `WriteFile` or `WriteJSONFile` here: https://github.com/helm/helm/blob/d6708667da5c79ab4a093ea7473e96b93be5f1f7/cmd/helm/repo_index.go#L108

hiddeco:
@smmhatami do you think you would be able to continue with this attempt based on my comments? Otherwise I would be happy to take it over, as I think it would be beneficial to Helm given the growing number of (bigger) indexes.

PR #12161, opened 2023-06-22
Do not validate metadata.name for List kind

visit1985:
Fix false-positive linter warning about missing metadata.name for List kind.

https://github.com/kubernetes/kubectl/issues/837#issuecomment-1342986901
https://github.com/helm/helm/issues/8439#issuecomment-1068979423

**If applicable**:

### DIFF
--- a/pkg/lint/rules/template.go
+++ b/pkg/lint/rules/template.go
@@ -261,7 +261,7 @@ func validateMetadataNameFunc(obj *K8sYamlStruct) validation.ValidateNameFunc {
 		return validation.ValidateNamespaceName
 	case "serviceaccount":
 		return validation.ValidateServiceAccountName
-	case "certificatesigningrequest":
+	case "certificatesigningrequest", "list":
 		// No validation.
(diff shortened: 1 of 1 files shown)


joejulian:
Following the links you shared, one links to the official api docs which says:

> Note that `kubectl` and other tools sometimes output collections of resources as kind: List. Keep in mind that kind: List is not part of the Kubernetes API; it is exposing an implementation detail from client-side code in those tools, used to handle groups of mixed resources.
> https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#types-kinds

This suggests to me that having a List kind in a helm chart isn't valid. I think lists should fail validation, in that case, but probably just because they're lists, not because they don't have a metadata.name.

visit1985:
To me, it is just cosmetics, to get rid of the warning. If you follow the second link I shared, you see, that there are people using this kind to workaround some ordering issues in Helm. So, making it fail would break some really suffisticated charts.

joejulian:
Yes. I see that hack but it's just that, a hack. Let's get a hip contribution to document a solution and go with a real solution. Lists are not valid and I'm really sorry to say this but I can't approve a PR to try to make them so when the word from Kubernetes is so definitive.

PR #12162, opened 2023-06-22
Fix multiple bugs in values handling

### LINKED ISSUE #9027: Could not delete default keys from subcharts if subchart is defined in dependencies
Hi,

looking at docs https://helm.sh/docs/chart_template_guide/values_files/#deleting-a-default-key I'm able to remove keys from default values located in subcharts:

values.yaml:
```yaml
# demo/values.yaml:
global:
  hash:
    key1: null
    key2: null
    key3: 13

subchart:
  hash:
    key1: null
    key2: null
    key3: 13
---
# demo/charts/subchart/values.yaml:
global:
  hash:
    key1: 1
    key2: 2
    key3: 3
    key4: 4
    key5: 5
    key6: 6

hash:
  key1: 1
  key2: 2
  key3: 3
  key4: 4
  key5: 5
  key6: 6
```

Output:

```
jkr@joe-nb demo % helm template .
---
global:
  hash:
    key3: 13
    key4: 4
    key5: 5
    key6: 6
hash:
  key3: 13
  key4: 4
  key5: 5
  key6: 6

```

But If I add:

```yaml
dependencies:
  - name: subchart
    version: 0.1.0
```

to the main Chart.yaml the behavior changes:

```
jkr@joe-nb demo % helm template .
---
global:
  hash:
    key1: 1
    key2: 2
    key3: 13
    key4: 4
    key5: 5
    key6: 6
hash:
  key1: 1
  key2: 2
  key3: 13
  key4: 4
  key5: 5
  key6: 6

```

The merge of values is still working while removing key isn't working.
This bug exists on helm2 w/ requirements.yaml, too. Just for notice.

Output of `helm version`:

```
jkr@joe-nb demo % helm version  
version.BuildInfo{Version:"v3.4.0", GitCommit:"7090a89efc8a18f3d8178bf47d2462450349a004", GitTreeState:"dirty", GoVersion:"go1.15.3"}
```

Demo: [chart.tar.gz](https://github.com/helm/helm/files/5549239/chart.tar.gz) (wrong file extension, its an uncompressed tar)

### LINKED ISSUE #10052: import-values in child-parent format doesn't work
Hi, Can't get example from here to work:
https://helm.sh/docs/topics/charts/#using-the-child-parent-format

To reproduce:
```bash
cat >Chart.yaml <<EOF
apiVersion: v2
name: mychart
version: 0.1.0
dependencies:
- name: subchart1
  version: 0.1.0
  import-values:
  - child: default.data
    parent: myimports
EOF

cat >values.yaml <<EOF
myimports:
  myint: 0
  mybool: false
  mystring: "helm rocks!"
EOF

mkdir -p templates
cat >templates/test.yaml <<EOF
out: | {{ $.Values | toYaml | nindent 2 }}
EOF

mkdir -p charts/subchart1/
cat >charts/subchart1/Chart.yaml <<EOF
apiVersion: v2
name: subchart1
version: 0.1.0
EOF

cat >charts/subchart1/values.yaml <<EOF
default:
  data:
    myint: 999
    mybool: true
EOF

helm template .
```

Result:
```bash
---
# Source: mychart/templates/test.yaml
out: | 
  myimports:
    mybool: false
    myint: 0
    mystring: helm rocks!
  subchart1:
    default:
      data:
        mybool: true
        myint: 999
    global: {}
```

Expected:
```bash
---
# Source: mychart/templates/test.yaml
out: | 
  myimports:
    mybool: true
    myint: 999
    mystring: helm rocks!
  subchart1:
    default:
      data:
        mybool: true
        myint: 999
    global: {}
```

Output of `helm version`:
```
version.BuildInfo{Version:"v3.6.3", GitCommit:"d506314abfb5d21419df8c7e7e68012379db2354", GitTreeState:"clean", GoVersion:"go1.16.5"}
```

Strange thing is that these two differ: https://github.com/helm/helm/blob/4a0b3d8c3d9fbbf3a87a347b215ee9f9cdea6ad9/pkg/chartutil/dependencies.go#L251
https://github.com/helm/helm/blob/4a0b3d8c3d9fbbf3a87a347b215ee9f9cdea6ad9/pkg/chartutil/dependencies.go#L263
Can't quite get my head around it. The short form of import-values works differently and looks like it's fine.

### LINKED ISSUE #10899: Values in library subcharts are ignored

mattfarina:
First, some notes about priority and how some code flow works.

For Helm handling values, the expected order of precedence is:
1. User specified values (e.g CLI)
2. Imported values
3. Parent chart values
4. Subchart values

Helm handles dependency values slightly differently. If there are dependencies in the charts folder that are not marked as dependencies all of the values, including nil values, are pulled in. If those charts are listed as a dependency in the Chart.yaml file than they are processed for import handling. Prior to the changes here, it caused nil values at the top level to NOT remove values specified.

The changes:

1. The order of priority was changed from the list above. Parent chart values would override specifically imported values. This is due to a change from just over a year ago that introduced a bug. That was undone by changing the precedence when maps were merged.

2. To handle merging while retaining the nil values, which was causing inconsistent behavior, a new set of Merge functions were introduced. These functions are just like coalesce except that they DO NOT remove nil/null values. The new functions are used in a backward compatible manner meaning some new functions were introduced that called them.

Specific issues fixed (that are known):

Closes #9027

Can now delete subkeys from charts when specified in the parent. This behavior was previously inconsistent. Sometimes they could be deleted and other times it did not work. Now it is consistent.

Closes #10899

Imported values (from library or other subcharts) are now used following the order above.

The previous behavior was inconsistent. import-values using just a string would import them. When named with a child/parent it did not work if the parent already had a value. If string and named were mixed the imports worked if the string happened first but just for the string not the named. If the named parent/child went first then none of them worked for cases where the parent already had a value. It was inconsistent and the tests sometimes mirrored the functionality rather than expected behavior.

Tests for this fall into the sub-packages and are in the template tests to verify it's happening in the output. Including having values passed at the CLI as the ultimate highest priority to be used.

This relates to a fix that went in for #9940. The expected values there don't fit the precedence above where the parent value would override the imported value. That fix/change introduced more bugs.

Closes #10052

This is the case where imported values using the parent/child designation just didn't work right. That has been fixed and there are tests. The underlying issue had to do with the precedence order handling.

Note, a lot of tests were added. Hope we got it more right this time.

**What this PR does / why we need it**: import-values is broken which is causing a number of bugs (see above).

There have been changes around this code for years which has introduced bugs. Also, some of the old tests check for the wrong values. For example, expecting a value to be overridden but making sure it's the original value instead of the override. The system around import-values has long had issues and those issues have changed over time leading to long term inconsistency.

This change fixes some long standing issues. Some of which come up just under edge conditions (e.g., you use import-values but you use parent/child before using a string value and the parent has an existing value for the import).

**If applicable**:

### DIFF
--- a/pkg/chartutil/coalesce.go
+++ b/pkg/chartutil/coalesce.go
@@ -53,7 +53,35 @@ func CoalesceValues(chrt *chart.Chart, vals map[string]interface{}) (Values, err
 	if valsCopy == nil {
 		valsCopy = make(map[string]interface{})
 	}
-	return coalesce(log.Printf, chrt, valsCopy, "")
+	return coalesce(log.Printf, chrt, valsCopy, "", false)
+}
+
+// MergeValues is used to merge the values in a chart and its subcharts. This
+// is different from Coalescing as nil/null values are preserved.
+//
+// Values are coalesced together using the following rules:
+//
+//   - Values in a higher level chart always override values in a lower-level
+//     dependency chart
+//   - Scalar values and arrays are replaced, maps are merged
+//   - A chart has access to all of the variables for it, as well as all of
+//     the values destined for its dependencies.
+//
+// Retaining Nils is useful when processes early in a Helm action or business
+// logic need to retain them for when Coalescing will happen again later in the
+// business logic.
+func MergeValues(chrt *chart.Chart, vals map[string]interface{}) (Values, error) {
+	v, err := copystructure.Copy(vals)
+	if err != nil {
+		return vals, err
+	}
+
+	valsCopy := v.(map[string]interface{})
+	// if we have an empty map, make sure it is initialized
+	if valsCopy == nil {
+		valsCopy = make(map[string]interface{})
+	}
+	return coalesce(log.Printf, chrt, valsCopy, "", true)
 }
 
 type printFn func(format string, v ...interface{})
@@ -61,13 +89,13 @@ type printFn func(format string, v ...interface{})
 // coalesce coalesces the dest values and the chart values, giving priority to the dest values.
 //
 // This is a helper function for CoalesceValues.
-func coalesce(printf printFn, ch *chart.Chart, dest map[string]interface{}, prefix string) (map[string]interface{}, error) {
-	coalesceValues(printf, ch, dest, prefix)
-	return coalesceDeps(printf, ch, dest, prefix)
+func coalesce(printf printFn, ch *chart.Chart, dest map[string]interface{}, prefix string, merge bool) (map[string]interface{}, error) {
+	coalesceValues(printf, ch, dest, prefix, merge)
+	return coalesceDeps(printf, ch, dest, prefix, merge)
 }
 
 // coalesceDeps coalesces the dependencies of the given chart.
-func coalesceDeps(printf printFn, chrt *chart.Chart, dest map[string]interface{}, prefix string) (map[string]interface{}, error) {
+func coalesceDeps(printf printFn, chrt *chart.Chart, dest map[string]interface{}, prefix string, merge bool) (map[string]interface{}, error) {
 	for _, subchart := range chrt.Dependencies() {
 		if c, ok := dest[subchart.Name()]; !ok {
 			// If dest doesn't already have the key, create it.
@@ -78,13 +106,11 @@ func coalesceDeps(printf printFn, chrt *chart.Chart, dest map[string]interface{}
 		if dv, ok := dest[subchart.Name()]; ok {
 			dvmap := dv.(map[string]interface{})
 			subPrefix := concatPrefix(prefix, chrt.Metadata.Name)
-
 			// Get globals out of dest and merge them into dvmap.
-			coalesceGlobals(printf, dvmap, dest, subPrefix)
-
+			coalesceGlobals(printf, dvmap, dest, subPrefix, merge)
 			// Now coalesce the rest of the values.
 			var err error
-			dest[subchart.Name()], err = coalesce(printf, subchart, dvmap, subPrefix)
+			dest[subchart.Name()], err = coalesce(printf, subchart, dvmap, subPrefix, merge)
 			if err != nil {
 				return dest, err
 			}
@@ -96,7 +122,7 @@ func coalesceDeps(printf printFn, chrt *chart.Chart, dest map[string]interface{}
 // coalesceGlobals copies the globals out of src and merges them into dest.
 //
 // For convenience, returns dest.
-func coalesceGlobals(printf printFn, dest, src map[string]interface{}, prefix string) {
+func coalesceGlobals(printf printFn, dest, src map[string]interface{}, prefix string, merge bool) {
 	var dg, sg map[string]interface{}
 
 	if destglob, ok := dest[GlobalKey]; !ok {
@@ -130,7 +156,10 @@ func coalesceGlobals(printf printFn, dest, src map[string]interface{}, prefix st
 					// Basically, we reverse order of coalesce here to merge
 					// top-down.
 					subPrefix := concatPrefix(prefix, key)
-					coalesceTablesFullKey(printf, vv, destvmap, subPrefix)
+					// In this location coalesceTablesFullKey should always have
+					// merge set to true. The output of coalesceGlobals is run
+					// through coalesce where any nils will be removed.
+					coalesceTablesFullKey(printf, vv, destvmap, subPrefix, true)
 					dg[key] = vv
 				}
 			}
@@ -156,12 +185,38 @@ func copyMap(src map[string]interface{}) map[string]interface{} {
 // coalesceValues builds up a values map for a particular chart.
 //
 // Values in v will override the values in the chart.
-func coalesceValues(printf printFn, c *chart.Chart, v map[string]interface{}, prefix string) {
+func coalesceValues(printf printFn, c *chart.Chart, v map[string]interface{}, prefix string, merge bool) {
 	subPrefix := concatPrefix(prefix, c.Metadata.Name)
-	for key, val := range c.Values {
+
+	// Using c.Values directly when coalescing a table can cause problems where
+	// the original c.Values is altered. Creating a deep copy stops the problem.
+	// This section is fault-tolerant as there is no ability to return an error.
+	valuesCopy, err := copystructure.Copy(c.Values)
+	var vc map[string]interface{}
+	var ok bool
+	if err != nil {
+		// If there is an error something is wrong with copying c.Values it
+		// means there is a problem in the deep copying package or something
+		// wrong with c.Values. In this case we will use c.Values and report
+		// an error.
+		printf("warning: unable to copy values, err: %s", err)
+		vc = c.Values
+	} else {
+		vc, ok = valuesCopy.(map[string]interface{})
+		if !ok {
+			// c.Values has a map[string]interface{} structure. If the copy of
+			// it cannot be treated as map[string]interface{} there is something
+			// strangely wrong. Log it and use c.Values
+			printf("warning: unable to convert values copy to values type")
+			vc = c.Values
+		}
+	}
+
+	for key, val := range vc {
 		if value, ok := v[key]; ok {
-			if value == nil {
-				// When the YAML value is null, we remove the value's key.
+			if value == nil && !merge {
+				// When the YAML value is null and we are coalescing instead of
+				// merging, we remove the value's key.
 				// This allows Helm's various sources of values (value files or --set) to
 				// remove incompatible keys from any previous chart, file, or set values.
 				delete(v, key)
@@ -177,7 +232,7 @@ func coalesceValues(printf printFn, c *chart.Chart, v map[string]interface{}, pr
 				} else {
 					// Because v has higher precedence than nv, dest values override src
 					// values.
-					coalesceTablesFullKey(printf, dest, src, concatPrefix(subPrefix, key))
+					coalesceTablesFullKey(printf, dest, src, concatPrefix(subPrefix, key), merge)
 				}
 			}
 		} else {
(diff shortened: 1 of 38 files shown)


sabre1041 (changes requested):
@mattfarina this is really good!

One small removal of debugging

pkg/chartutil/coalesce.go
```diff
@@ -156,12 +185,39 @@ func copyMap(src map[string]interface{}) map[string]interface{} {
...
 // Values in v will override the values in the chart.
-func coalesceValues(printf printFn, c *chart.Chart, v map[string]interface{}, prefix string) {
+func coalesceValues(printf printFn, c *chart.Chart, v map[string]interface{}, prefix string, merge bool) {
 	subPrefix := concatPrefix(prefix, c.Metadata.Name)
-	for key, val := range c.Values {
+
+	// Using c.Values directly when coalescing a table can cause problems where
+	// the original c.Values is altered. Creating a deep copy stops the problem.
+	// This section is fault-tolerant as there is no ability to return an error.
+	//log.Printf(">>>>>>>>>>>>>>>> %#v", c.Values)
```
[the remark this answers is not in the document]
mattfarina: ugh. Good catch. thanks.

mattfarina:
@sabre1041 can you please re-review it?

jamoflaw:
Thanks for jumping on this @mattfarina @sabre1041! 🚀

jamoflaw:
Interestingly with this one I can get a situation where a null appears in the resulting template
image

This situation has:

Child values.yaml
```
resources:
  limits:
    cpu: 1000m
    memory: 2Gi
  requests:
    cpu: 1000m
    memory: 2Gi
```

Parent values.yaml
```
resources:
  requests:
    cpu: 1000m
    memory: 2Gi
  limits:
    memory: 2Gi
    cpu: null
```

CLI passed in values:
```
  resources:
    requests:
      cpu: 100m
      memory: 6Gi
    limits:
      memory: 32Gi
```

So in this case the null is being persisted rather than trimmed from the final result - is that desired?

jamoflaw:
Minimal repro is:

`Chart.yaml` with
```
dependencies:
  - name: elasticsearch
    version: 7.16.3
    repository: https://helm.elastic.co
```

chart values.yaml as
```
elasticsearch:
  resources:
    limits:
      cpu: null
```

The null will persist through into the template

mattfarina:
@jamoflaw You bring up a good point. Thanks for providing a good example. Here are my thoughts...

1. "is that desired?" I don't know that it is desired. But, I was not trying to build the fully desired system for values handling. That system is built and in wide use. We won't want to break anything. Helm places a high priority on backwards compatibility. This PR makes enough changes I'm almost not comfortable and I will widely share this change, if it lands, to get people testing it.
2. Older versions of Helm (I tested multiple) preserve the `null` you see in your examples.
3. The goal of this PR was to make null delete a key everywhere consistently and to fix the import-values bugs. Those 2 problems were in the same code. Removing all `null` values was not part of this (and I didn't see a bug on it).

Before removing all `null` values we would need a lot of backwards compatibility testing. I think that should be a separate change.

pkg/chartutil/coalesce.go
```diff
@@ -53,21 +53,49 @@ func CoalesceValues(chrt *chart.Chart, vals map[string]interface{}) (Values, err
...
+		valsCopy = make(map[string]interface{})
+	}
+	return coalesce(log.Printf, chrt, valsCopy, "", true)
 }
 
 type printFn func(format string, v ...interface{})
 
 // coalesce coalesces the dest values and the chart values, giving priority to the dest values.
 //
 // This is a helper function for CoalesceValues.
```
gjenkins8, line 91: nit:
```suggestion
// This is a helper function for MergeValues/CoalesceValues.
```

pkg/chartutil/dependencies_test.go
```diff
@@ -274,7 +321,7 @@ func TestProcessDependencyImportValuesForEnabledCharts(t *testing.T) {
 	c := loadChart(t, "testdata/import-values-from-enabled-subchart/parent-chart")
 	nameOverride := "parent-chart-prod"
 
-	if err := processDependencyImportValues(c); err != nil {
+	if err := processDependencyImportValues(c, false); err != nil {
```
gjenkins8, line 324: Why would this be `false`? (Or, at least, I don't understand why we would expect the old merge behavior for this test)

pkg/chartutil/coalesce.go
mattfarina: Did that while cleaning up a previous comment. Good catch.

pkg/chartutil/dependencies_test.go
mattfarina: I think I missed it before. This particular test produces the same results whether the values is true or false. Updating to test the new behavior.

pkg/chartutil/coalesce.go
```diff
@@ -53,21 +53,49 @@ func CoalesceValues(chrt *chart.Chart, vals map[string]interface{}) (Values, err
...
+	v, err := copystructure.Copy(vals)
+	if err != nil {
+		return vals, err
+	}
+
+	valsCopy := v.(map[string]interface{})
+	// if we have an empty map, make sure it is initialized
+	if valsCopy == nil {
+		valsCopy = make(map[string]interface{})
+	}
```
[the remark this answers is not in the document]
mattfarina: Done

mattfarina:
@gjenkins8 @sabre1041 can you please re-review

gjenkins8 (approved):
Nice! LGTM from me

yxxhero (approved):
LGTM

sabre1041 (approved):

benjaminjb:
@mattfarina Have you tested unsetting subchart keys with a `parent->child->grandchild` setup? I've only started to experiment, so maybe I'm doing something wrong, but I'm not seeing keys being unset:

Parent-to-child unset works: `helm template . --set parent.child.key=null` --> child.key is null
Parent-to-grandchild unset doesn't work: `helm template . --set parent.child.grandchild.key=null` --> grandchild.key is the non-null default.

alice-sawatzky:
@benjaminjb I can reproduce this, but there's an additional nuance that's even more problematic. This doesn't just not fix the issue, it also introduces a new bug. I've attached a test script below, it:
- sets up a minimal parent, child, and grandchild chart
- configures the child chart `values.yaml` to override a key in the grandchild chart `values.yaml`
- configures the parent chart `values.yaml` to override a key in the child chart `values.yaml`

for clarity, in this comment I'll be using the words "parent", "child", and "grandchild" in the absolute sense, as the names of the charts, not as relative terms to a prior reference.

- `./test.sh helm template .`:
  - :heavy_check_mark:  grandchild chart value is correctly overridden by value from child chart's `values.yaml`
  - :heavy_check_mark:  child chart value is correctly overridden by the parent chart's `values.yaml`
- `./test.sh helm template --set child.key=null .`:
  - :heavy_check_mark:  grandchild chart value is correctly overridden by the child chart's `values.yaml`
  - :x: child chart value is not unset (as @benjaminjb observed)
  - :exclamation: child chart value is reverted to its own `values.yaml` file's value, despite being overridden by the parent chart's `values.yaml`
- `./test.sh helm template --set child.grandchild.key=null`:
  - :heavy_check_mark: child chart value is correctly overridden by value from parent chart's `values,yaml`
  - :x: grandchild chart value is not unset (as @benjaminjb observed)
  - :exclamation: grandchild chart value is reverted to its own `values.yaml` file's value, despite being overridden by the parent chart's `values.yaml`

```
#!/bin/bash
# Usage: ./test.sh <helm command>
set -eou pipefail

HELM="$1"
shift

TESTDIR="${TESTDIR:-"/tmp/helmtest"}"

function chartYaml {
    echo -e "apiVersion: v2\nname: $1\nversion: 0.1.0"
}

function makeChart {
    NAME="$(basename "$1")"
    mkdir -p "$1/templates" "$1/charts"
    touch "$1/templates/placeholder.txt"
    touch "$1/values.yaml"
    chartYaml "$NAME" > "$1/Chart.yaml"
}

PARENT_DIR="$TESTDIR/parent"
CHILD_DIR="$PARENT_DIR/charts/child"
GRANDCHILD_DIR="$CHILD_DIR/charts/grandchild"
rm -rf "$PARENT_DIR"
# Parent Chart Stetup
makeChart "$PARENT_DIR"
echo -e "# Parent Chart's view of values:\n{{ toYaml .Values }}" > "$PARENT_DIR/templates/testcase.yaml"
cat <<EOF > "$PARENT_DIR/values.yaml"
key: "Value for parent chart, from parent chart values.yaml"
child:
  key: "Value for child chart, overridden from parent chart values.yaml"
EOF

# Child Chart Setup
makeChart "$CHILD_DIR"
echo -e "# Child Chart's view of values:\n{{ toYaml .Values }}" > "$CHILD_DIR/templates/testcase.yaml"
cat <<EOF > "$CHILD_DIR/values.yaml"
key: "Value for childChart, from childChart values.yaml"
grandchild:
  key: "Value for grandchild chart, overridden from child chart values.yaml"
EOF

# Granchild Chart Setup
makeChart "$GRANDCHILD_DIR"
echo -e "# Grandchild Chart's view of values:\n{{ toYaml .Values }}" > "$GRANDCHILD_DIR/templates/testcase.yaml"
cat <<EOF > "$GRANDCHILD_DIR/values.yaml"
key: "Value for grandchild chart, from grandchild values.yaml"
EOF

set -x; cd "$PARENT_DIR" && "$HELM" template "$@"
```

alice-sawatzky:
@mattfarina should I move my comment to a new issue? For my use case this is a serious regression

Laurens-W:
Adding to what @alice-sawatzky wrote:
A work around that was known to me was to explicitly pass a values file: `helm template . -f values.yaml`
This no longer works as of 3.13.0 and now seems to behave the same as `helm template .` did pre 3.13

So from my tests:
3.12.3
`helm template .` : doesn't handle null values correctly
`helm template . -f values.yaml` : does handle null values correctly

3.13.0
`helm template .` : does handle null values correctly
`helm template . -f values.yaml` : doesn't handle null values correctly

zegerius:
Echoing @Laurens-W and others.

The fix I introduced together with my colleague @ankitabhopatkar13 with the referenced https://github.com/helm/helm/issues/9940 is now reverted and broke all of our charts. I don't think you can willy-nilly change these things and add it as an (unreferenced!) footnote in a release.

For very clear reference, this is now (again) broken: https://github.com/ankitabhopatkar13/helm-values-issue 

@mattfarina

PR #11726, opened 2023-01-13
Improve helm dependency update performance

JvD-Ericsson:
What this PR does / why we need it:
This PR was created to improve performance of the dependency update command by skipping unnecessary downloading and loading of index files that have already been downloaded and loaded

I believe this would close refs #9865

Signed-off-by: Jeff van Dam <jeff.van.dam@est.tech>

PR #12167, opened 2023-06-25
perf: add an in-memory cache for index files

yardenshoham:
This saves a lot of unnecessary yaml parsing

This made `helm dependency update` run on my 63 dependencies chart in 60 seconds instead of 800 seconds


stavros-k:
That would be nice. I've had to write a custom script to fetch dependencies as it was taking over **3 hours** for ~800 Charts.
This script dropped it down to **~2 minutes**
https://github.com/truecharts/charts/blob/master/.github/scripts/fetch_helm_deps.sh

sabre1041 (commented):
@yardenshoham This is a good feature. Is there a chance that you might be able to add a test?

yardenshoham:
I'm not sure how to test a performance improvement, I didn't change the logic

mattfarina (changes requested):
This is part of the Helm SDK. The SDK is used in long running processes. It looks like the cache will just grow and grow. This is fine for the Helm CLI but not the long running services that will use this function.

`FindChartInAuthAndTLSAndPassRepoURL`, which calls this function, will delete the files from the filesystem which the content here remains.

The lack of cache invalidation is going to be a problem. If Helm is going to cache here, how can it be reworked to not have stale data and to handle long running apps (where the indexes will changes over time)?

yardenshoham:
I can drop the first cache, the one with the key as path and just use the cache by the content of the file. We can also make the cache opt-in by passing a boolean parameter (long-running applications can just not use the cache, as before).

yardenshoham:
@mattfarina do you think the boolean parameter to opt-in for the cache is a good approach?

yardenshoham:
Another option: use https://github.com/dgraph-io/ristretto and only cache by file content, we could make sure the cost is limited to 1GB of memory

yardenshoham:
@mattfarina what do you think of the options I suggested?

mattfarina:
@yardenshoham caching is hard. For example, if you did a catch based on the index file itself it would catch changes to that. But, as a long running process continued to run the cache would continue to get bigger and bigger as files changed. There would be no chance to remove things from the cache.

I was thinking about how I would write it. I would do the following:
- I would create an interface for a cache that let you add something to the cache and get something from the cache if it exists. This way a user could write their own implementation
- I would create an implementation of that cache that works for a short lived application like the Helm client
- I would create a no-op version of the cache
- I would create a new function with a name like `LoadIndexFileWithCache` and pass in the cache as a pointer (so it can be maintained outside of the function). It would do the existing work of `LoadIndexFile`
- I would update `LoadIndexFile` to call `LoadIndexFileWithCache` and pass in the no-op cache. This way the existing functionality is preserved for the function and anyone who uses it from the SDK.
- For the current uses of `LoadIndexFile` in the `cmd/` directory (where it is the Helm client application) I would switch to `LoadIndexFileWithCache` with the functional cache.
- For the other things that call it internally, I would find a way to add the cache as a reference that's passed around. I would also handle the case where the property on the struct is `nil` and use the no-op cache to preserve existing behavior.

The goal would be to preserve existing behavior for API callers while giving them an opportunity to inject their own cache. For the Helm client application we would use a cache designed for short life where invalidation isn't a problem.

Does that make sense?

joejulian:
There are two hard problems in computer science:
1. Naming things
1. Cache Invalidation
1. Off-by-one errors

Have you thought about how to handle cache invalidation when used as a long-running service?

yardenshoham:
@mattfarina I like the idea. I wonder if I should try it or if we should take https://github.com/helm/helm/pull/11726. What would you prefer?

yardenshoham:
Seeing as https://github.com/helm/helm/pull/11726 was reverted, I am proceeding with this PR. I implemented @mattfarina's requested changes.

PR #12172, opened 2023-06-30
fix: add HELM_INSTALL_DIR to testVersion

### LINKED ISSUE #12158: get-helm-3 script does not allow helm to be downloaded into directory not on system path
## Observed behavior:

When using `get-helm-3` to download `helm` into a directory not on the system path, the script fails with the following output:

```
curl -s "https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3" | DESIRED_VERSION=v3.5.2 HELM_INSTALL_DIR=/home/circleci/project/bin USE_SUDO=false bash
Downloading https://get.helm.sh/helm-v3.5.2-linux-amd64.tar.gz
Verifying checksum... Done.
Preparing to install helm into /home/circleci/project/bin
helm installed into /home/circleci/project/bin/helm
helm not found. Is /home/circleci/project/bin on your $PATH?
Failed to install helm
```

## Expected behavior:

I would expect `get-helm-3` to allow me to download `helm` into whatever directory I choose. In fact, it takes `HELM_INSTALL_DIR` as an environment variable.

But then it performs the following check in `testVersion` that enforces the constraint that the `HELM_INSTALL_DIR` is on the system path:

https://github.com/helm/helm/blob/4e447d87cd3222aa5e5ea2b38170e9d80b99c0b4/scripts/get-helm-3#L258-L262

> **NOTE**: Actually, if `HELM_INSTALL_DIR` is not on the system path, but there is some other `helm` executable already on the system path, then the `testVersion` check will pass.

I think it is a perfectly valid use case to use `get-helm-3` to download `helm` into a directory not on the system path so that it can be invoked explicitly.

As an example, Kubebuilder scaffolds a [Makefile](https://book.kubebuilder.io/reference/makefile-helpers.html) with phony targets that download tools into a local `bin` directory where they can be invoked directly, and it is useful to integrate Helm with this scaffolding.

wujunwei:
fixes #12158 

**If applicable**:

### DIFF
--- a/scripts/get-helm-3
+++ b/scripts/get-helm-3
@@ -261,7 +261,7 @@ fail_trap() {
 # testVersion tests the installed client to make sure it is working.
 testVersion() {
   set +e
-  HELM="$(command -v $BINARY_NAME)"
+  HELM="$(command -v $HELM_INSTALL_DIR/$BINARY_NAME)"
   if [ "$?" = "1" ]; then
     echo "$BINARY_NAME not found. Is $HELM_INSTALL_DIR on your "'$PATH?'
     exit 1
--- a/scripts/get-helm-3
+++ b/scripts/get-helm-3
@@ -263,7 +263,7 @@ testVersion() {
   set +e
   HELM="$(command -v $HELM_INSTALL_DIR/$BINARY_NAME)"
   if [ "$?" = "1" ]; then
-    echo "$BINARY_NAME not found. Is $HELM_INSTALL_DIR on your "'$PATH?'
+    echo "$BINARY_NAME not found or can not be executed"
     exit 1
   fi
   set -e


harsimranmaan (approved):
Lgtm. Needed this for a workflow. Although this changes existing behaviour. But someone setting the install path likely knows what they are doing.

scripts/get-helm-3
```diff
@@ -255,7 +255,7 @@ fail_trap() {
 # testVersion tests the installed client to make sure it is working.
 testVersion() {
   set +e
-  HELM="$(command -v $BINARY_NAME)"
+  HELM="$(command -v $HELM_INSTALL_DIR/$BINARY_NAME)"
```
gjenkins8, line 258: I think this is (unfortunately) too simple. I think we need to conditionally check:

- Iff the user sets `HELM_INSTALL_DIR`, validate the installed binary is executable at `$HELM_INSTALL_DIR/$BINARY_NAME` ie. `command -v $HELM_INSTALL_DIR/$BINARY_NAME`
- Else, if the user does not set `HELM_INSTALL_DIR`, check `$BINARY_NAME`, which would be expected to be in the user's path. ie. the current test: `command -v $BINARY_NAME`

I think the current point of `testVersion` is to verify that Helm is accessible to the user in their PATH (see the error message). This change allows this check to succeed, even if that isn't true.

wujunwei: You can see from the script, if the user does not set `HELM_INSTALL_DIR`,   `HELM_INSTALL_DIR` will be default value "/usr/local/bin" , so `command -v $HELM_INSTALL_DIR/$BINARY_NAME` is fine.
```
: ${HELM_INSTALL_DIR:="/usr/local/bin"}
```
I think the verification method for 'testVersion' should also be consistent with the following methods `checkHelmInstalledVersion`.
```
checkHelmInstalledVersion() {
  if [[ -f "${HELM_INSTALL_DIR}/${BINARY_NAME}" ]]; then
    local version=$("${HELM_INSTALL_DIR}/${BINARY_NAME}" version --template="{{ .Version }}")
    if [[ "$version" == "$TAG" ]]; then
      echo "Helm ${version} is already ${DESIRED_VERSION:-latest}"
      return 0
    else
      echo "Helm ${TAG} is available. Changing from version ${version}."
      return 1
    fi
  else
    return 1
  fi
}
```

joejulian (changes requested):
@wujunwei Please work with @gjenkins8 to address his comments.

wujunwei:
@joejulian  
From the perspective of the user, it should just check whether the executable file can be downloaded to the directory specified by the user, not that it must be in the System Path.

I have change the error message,PTAL.

PR #12173, opened 2023-06-30
feat: add support for S3 Compatible Blob storages as StorageDriver

DaspawnW:

```
In case you run a large installation of Kubernetes with a lot of releases the amount of secrets and the size get a potential issue. To overcome this there are other storage drivers like SQL, but if you are in a restricted environment usually Databases are the worth protecting resource. Due to this S3 is IMHO a valid alternative and as the protocol is widely adapted also outside of the AWS area it's not only cloud provider specific.
```

**If applicable**:
- [x] this PR contains documentation (https://github.com/helm/helm-www/pull/1481)


TerryHowe (commented):
I don't think bringing in all these dependencies is a good idea. I am a small voice, but I don't think helm should integrate directly with storage solutions unless it was through some plugin type architecture.

I think the real solution here is make some small changes to the OCI client to read files instead of registries in OCI layout format.

TerryHowe:
For example https://oras.land/blog/lightweight-cloud-registry-oras/

PR #12202, opened 2023-07-13
chore: remove unuse untar option

Abirdcfly:
[`getter.options`](https://github.com/helm/helm/blob/14d363669f8cc257f0413995d42237e673422d81/pkg/getter/getter.go#L33) has parameters [`untar`](https://github.com/helm/helm/blob/14d363669f8cc257f0413995d42237e673422d81/pkg/getter/getter.go#L38) and its setting function [`WithUntar()`]( https://github.com/helm/helm/blob/14d363669f8cc257f0413995d42237e673422d81/pkg/getter/getter.go#L118C6-L118C15).
But it's not used anywhere, not even in the tests.

**If applicable**:

### DIFF
--- a/pkg/getter/getter.go
+++ b/pkg/getter/getter.go
@@ -35,7 +35,6 @@ type options struct {
 	certFile              string
 	keyFile               string
 	caFile                string
-	unTar                 bool
 	insecureSkipVerifyTLS bool
 	username              string
 	password              string
@@ -112,12 +111,6 @@ func WithTagName(tagname string) Option {
 func WithRegistryClient(client *registry.Client) Option {
 	return func(opts *options) {
 		opts.registryClient = client
-	}
-}
-
-func WithUntar() Option {
-	return func(opts *options) {
-		opts.unTar = true
 	}
 }
 


mattfarina (changes requested):
Helm follows semantic versioning. We cannot remove this until Helm v4 or the Go API will be broken.

A useful change would be to have a code comment noting that it's deprecated.

Abirdcfly:
> We cannot remove this until Helm v4 or the Go API will be broken.

But `options.untar` is an **Non-capitalized** identifier. it can not be exported and can only be accessed from within the same package, and we do not use this identifier in `options` package...

So, neither `options.untar` nor `WithUntar()` will have any effect.

mattfarina:
@Abirdcfly The issue is with `WithUntar`. That is a public method that could be used in some calling code for someone using the Helm SDK. Even if it does nothing. Removing it will break their calling code. This is part of the public Helm Go SDK API so we can't remove it until Helm v4. To remove it would violate Semantic Versioning.

joejulian:
Thanks for noticing this and making an effort, @Abirdcfly . I'm going to go ahead and close this, though, since it's not something that can be merged.

PR #12203, opened 2023-07-13
chore: HTTPGetter add default timeout

0xff-dev:
The timeout for http requests should be added when adding or updating a repo. By default there is no timeout limit. 

Or Maybe need to add a timeout parameter to `repo add` and `repo update`?

**If applicable**:

### DIFF
--- a/pkg/getter/getter.go
+++ b/pkg/getter/getter.go
@@ -172,9 +172,21 @@ func (p Providers) ByScheme(scheme string) (Getter, error) {
 	return nil, errors.Errorf("scheme %q not supported", scheme)
 }
 
+const (
+	// The cost timeout references curl's default connection timeout.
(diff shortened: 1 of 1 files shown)


pkg/getter/getter.go
```diff
@@ -172,9 +172,14 @@ func (p Providers) ByScheme(scheme string) (Getter, error) {
 	return nil, errors.Errorf("scheme %q not supported", scheme)
 }
 
+var defaultOptions = []Option{WithTimeout(time.Second * 120)}
```
joejulian, line 175: Can we make the duration a const with a comment explaining how the value was chosen, please?

joejulian (changes requested):
Thanks for this! I have one small change requested.

pkg/getter/getter.go
0xff-dev: Done.

joejulian (approved):

PR #12241, opened 2023-07-20
fix: output warning message when pass-credentials are overwritten

### LINKED ISSUE #12117: --pass-credentials does not take effect
Output of `helm version`: 3.12.0

`helm pull` 的 `--pass-credentials` 会被 `repositories.yaml` 的 `pass_credentials_all` 设置覆盖
这会导致 `pass_credentials_all` 为 `false` 时,  `helm pull xxx`即便使用了`--pass-credentials`也是无效的, 认证信息都不会被传递

我不确定这是helm的设计还是一个bug, 我的感觉是bug, 因为这个逻辑太奇怪了

google translate:
`--pass-credentials` for `helm pull` will be overridden by `pass_credentials_all` setting in `repositories.yaml`

This will cause `helm pull xxx` to be invalid even if `--pass-credentials` is used when `pass_credentials_all` is `false`, and the authentication information will not be passed

I'm not sure if this is the design of helm or a bug, I feel like it's a bug, because the logic is so weird

https://github.com/helm/helm/blob/c9f554d75773799f72ceef38c51210f1842a1dea/pkg/action/pull.go#L90
https://github.com/helm/helm/blob/c9f554d75773799f72ceef38c51210f1842a1dea/pkg/downloader/chart_downloader.go#L239

0xff-dev:
Fix:  https://github.com/helm/helm/issues/12117

The current implementation is to put the parameters of repositories.yaml after the list, which will cause the user and password to be overwritten even if we set them.
So, we put the contents of repositories.yaml at the front of the list.

@joejulian 

**If applicable**:

### DIFF
--- a/pkg/downloader/chart_downloader.go
+++ b/pkg/downloader/chart_downloader.go
@@ -233,6 +233,7 @@ func (c *ChartDownloader) ResolveChartVersion(ref, version string) (*url.URL, er
 			c.Options = append(c.Options, getter.WithTLSClientConfig(rc.CertFile, rc.KeyFile, rc.CAFile))
 		}
 		if rc.Username != "" && rc.Password != "" {
+			fmt.Fprintf(c.Out, "WARNING: find the same URL repo with a configured password will override the pass-credentials parameter with PassCredentialsAll: %v.", rc.PassCredentialsAll)
 			c.Options = append(
 				c.Options,
 				getter.WithBasicAuth(rc.Username, rc.Password),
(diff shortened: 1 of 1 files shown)


joejulian (changes requested):
This PR is contrary to the discussion that was had on the PR. The final outcome was expected to be a PR that would warn the user if `--pass-credentials` is used when `pass_credentials_all` is `false`.

https://github.com/helm/helm/issues/12117#issuecomment-1577945270

0xff-dev:
Yes, the discussion said to add some warnings, so this PR should not fix that issue.

The problem with this PR is that when the user sets the username and password during pull and install, the information in the configuration is still used.

Should my PR be unlinked from that issue? And submit another PR to add the warning?

joejulian:
@0xff-dev Yes. This PR changes the behavior which goes against the guiding principal in [hip-0004](https://github.com/helm/community/blob/main/hips/hip-0004.md).

0xff-dev:
> @0xff-dev Yes. This PR changes the behavior which goes against the guiding principal in [hip-0004](https://github.com/helm/community/blob/main/hips/hip-0004.md).

Thanks for the review, the code has been modified.

PR #12245, opened 2023-07-20
Add support for creating repository indexes in JSON format

### LINKED ISSUE #10542: index.json - necessary for large projects

hiddeco:

This adds support for creating an `index.yaml` in JSON format using `helm repo index [...] --json`.

Fixes #10542
Supersedes #12151

Given that it does not create unnecessary HTTP requests to discover `index.json` and/or `index.yaml` files hosted by a repository server, nor requires rigorous changes to the caching logic, and is fully backwards compatible due to YAML being a superset of JSON. This appears to be the least invasive way to add support for the JSON format.

Benchmark using Bitnami's infamous ["full index" archive](https://raw.githubusercontent.com/bitnami/charts/archive-full-index/bitnami/index.yaml):

```console
$  benchstat old.txt new.txt                                  
goos: linux
goarch: amd64
pkg: helm.sh/helm/v3/pkg/repo
cpu: Intel(R) Core(TM) i7-10750H CPU @ 2.60GHz
                      │   old.txt    │               new.txt               │
                      │    sec/op    │   sec/op     vs base                │
LoadIndexFile/YAML-12     2.692 ± 1%    2.770 ± 3%        ~ (p=0.242 n=20)
LoadIndexFile/JSON-12   2318.3m ± 0%   582.6m ± 1%  -74.87% (p=0.000 n=20)
geomean                   2.498         1.270       -49.15%

                      │    old.txt    │               new.txt                │
                      │     B/op      │     B/op      vs base                │
LoadIndexFile/YAML-12    836.0Mi ± 0%   836.0Mi ± 0%        ~ (p=0.355 n=20)
LoadIndexFile/JSON-12   671.99Mi ± 0%   93.84Mi ± 0%  -86.03% (p=0.000 n=20)
geomean                  749.5Mi        280.1Mi       -62.63%

                      │   old.txt    │               new.txt               │
                      │  allocs/op   │  allocs/op   vs base                │
LoadIndexFile/YAML-12    12.03M ± 0%   12.03M ± 0%        ~ (p=0.683 n=20)
LoadIndexFile/JSON-12   10.015M ± 0%   1.078M ± 0%  -89.23% (p=0.000 n=20)
geomean                  10.98M        3.601M       -67.19%
```

**If applicable**:

### DIFF
--- a/cmd/helm/repo_index.go
+++ b/cmd/helm/repo_index.go
@@ -43,6 +43,7 @@ type repoIndexOptions struct {
 	dir   string
 	url   string
 	merge string
+	json  bool
 }
 
 func newRepoIndexCmd(out io.Writer) *cobra.Command {
@@ -70,6 +71,7 @@ func newRepoIndexCmd(out io.Writer) *cobra.Command {
 	f := cmd.Flags()
 	f.StringVar(&o.url, "url", "", "url of chart repository")
 	f.StringVar(&o.merge, "merge", "", "merge the generated index into the given index")
+	f.BoolVar(&o.json, "json", false, "output in JSON format")
 
 	return cmd
 }
@@ -80,10 +82,10 @@ func (i *repoIndexOptions) run(out io.Writer) error {
 		return err
 	}
 
-	return index(path, i.url, i.merge)
+	return index(path, i.url, i.merge, i.json)
 }
 
-func index(dir, url, mergeTo string) error {
+func index(dir, url, mergeTo string, json bool) error {
 	out := filepath.Join(dir, "index.yaml")
 
 	i, err := repo.IndexDirectory(dir, url)
@@ -95,7 +97,7 @@ func index(dir, url, mergeTo string) error {
 		var i2 *repo.IndexFile
 		if _, err := os.Stat(mergeTo); os.IsNotExist(err) {
 			i2 = repo.NewIndexFile()
-			i2.WriteFile(mergeTo, 0644)
+			writeIndexFile(i2, mergeTo, json)
 		} else {
 			i2, err = repo.LoadIndexFile(mergeTo)
 			if err != nil {
@@ -105,5 +107,12 @@ func index(dir, url, mergeTo string) error {
 		i.Merge(i2)
 	}
 	i.SortEntries()
+	return writeIndexFile(i, out, json)
+}
+
+func writeIndexFile(i *repo.IndexFile, out string, json bool) error {
+	if json {
+		return i.WriteJSONFile(out, 0644)
+	}
 	return i.WriteFile(out, 0644)
 }
--- a/cmd/helm/repo_index_test.go
+++ b/cmd/helm/repo_index_test.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"bytes"
+	"encoding/json"
 	"io"
 	"os"
 	"path/filepath"
@@ -66,6 +67,28 @@ func TestRepoIndexCmd(t *testing.T) {
 	expectedVersion := "0.2.0"
 	if vs[0].Version != expectedVersion {
 		t.Errorf("expected %q, got %q", expectedVersion, vs[0].Version)
+	}
+
+	b, err := os.ReadFile(destIndex)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if json.Valid(b) {
+		t.Error("did not expect index file to be valid json")
+	}
+
+	// Test with `--json`
+
+	c.ParseFlags([]string{"--json", "true"})
+	if err := c.RunE(c, []string{dir}); err != nil {
+		t.Error(err)
+	}
+
+	if b, err = os.ReadFile(destIndex); err != nil {
+		t.Fatal(err)
+	}
+	if !json.Valid(b) {
+		t.Error("index file is not valid json")
 	}
 
 	// Test with `--merge`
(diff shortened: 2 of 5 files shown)


pjbgf (approved):
This looks pretty good, it will make a massive difference processing large helm indexes.

sabre1041:
This is a really good enhancement and verified that the enhancement does function correctly. However, the use of `index.yaml` for both JSON and YAML content is confusing as to the actual format of content included in this file.

It might be interesting to consider two potential files that could contain repository metadata: `index.yaml` and `index.json` (where `index.yaml` would take precedence if both files were found)

hiddeco:
> It might be interesting to consider two potential files that could contain repository metadata: `index.yaml` and `index.json` (where `index.yaml` would take precedence if both files were found)

In an iteration before creating this PR, I actually had an approach like this where [`ChartRepository#DownloadIndexFile`](https://github.com/helm/helm/blob/v3.12.2/pkg/repo/chartrepo.go#L117) would loop over possible URLs (first JSON, then YAML), while breaking the loop on the first non-error response (or returning a collection of errors if none was found).

Besides this approach causing potentially loads of HTTP 404s for the server when a YAML isn't served, after looking closer at the caching logic (more precisely the various [`CacheIndexFile`](https://github.com/helm/helm/blob/f099c2e883f43f853cdffd974ad2b8f1792cee1c/pkg/helmpath/home.go#L30-L35) calls throughout the code base, and required changes to certain [error types](https://github.com/helm/helm/blob/4e7e939f19719e056c0f6e47b02068bdca6f420c/pkg/repo/index.go#L52-L53) to be factually correct). Getting this done right appeared more painful than leveraging YAML being a superset of JSON.

Another upside of the approach taken is that it's backwards compatible without having to serve (and thus store) two different file types. A Helm repository administrator can simply decide to start serving JSON, and any version of Helm will happily be able to download and work with the new index format (albeit only the Hem version which would include this change with memory consumption improvements).

joejulian (approved):
> ... the use of index.yaml for both JSON and YAML content is confusing ...

I disagree. Since json is a subset of yaml, it seems perfectly valid to me.

The rest of hiddeco's rational is also very compelling.

I tested this on my use cases and it works for me.

sabre1041:
@hiddeco @joejulian My concern with mixing JSON and YAML content is centered around two primary areas:

1. Understanding content types within `index.yaml` file. Yes YAML is a subset of JSON, but if there was an assumption that the contents included in these files were one format or the other, unexpected parsing errors could occur
2. Backwards compatibility. Do we break it by implementing such a change?

joejulian:
Other way around. Yaml is a superset of json. Anything in json is 100% compatible with yaml. Yaml has features that are not in json.

So any json content in a yaml file is guaranteed compatible.

hiddeco:
> YAML is a subset of JSON, but if there was an assumption that the contents included in these files were one format or the other, unexpected parsing errors could occur

The actual parser Helm uses [converts YAML to JSON](https://github.com/kubernetes-sigs/yaml/blob/b02c3522cfd003a3879afea2f58356bd02fede0f/yaml.go#L72-L81). In addition, from the [YAML 1.2 spec (2009, revised in 2021)](https://yaml.org/spec/1.2.2/#12-yaml-history):

> [...] Its primary focus was **making YAML a strict superset of JSON**. It also removed [...]

---

> Backwards compatibility. Do we break it by implementing such a change?

As noted in my [previous comment](https://github.com/helm/helm/pull/12245#issuecomment-1647442138), the exact opposite is true. As introducing an `index.json` would mean that you either:

1. Have to continue to publish both an `index.yaml` and `index.json` file as a Helm repository administrator, as otherwise older versions of Helm are unable to work with your repository. It's also worth to note this makes publishing itself more complicated, as you now have to ensure both files are in sync.
2.  Stop publishing an `index.yaml`, which would yield your Helm repository useless for older versions of Helm.

sabre1041 (approved):
@hiddeco @joejulian Reevaluated code changes, comments and LGTM

thernstig:
@hiddeco @sabre1041 this is amazing! Should it not be mentioned in the docs?

hiddeco:
They are partly automatically documented: https://github.com/helm/helm-www#updating-the-helm-cli-reference-docs

mattfarina:
Note, while writing the release notes I discovered this doesn't fully work. I filed a bug at https://github.com/helm/helm/issues/12422.

hiddeco:
It will not help if you are not in control of the published format of the repository index (which appears to be the case based on what you describe).

PR #12265, opened 2023-08-01
More gitignore-like implementation of negative ignore rules

### LINKED ISSUE #3622: Whitelisting in .helmignore with '/*' returns "chart metadata (Chart.yaml) missing"

evankanderson:

Fixes #3622

closes #3622

I'm willing to add a special case for a single negated rule that basically translates:

```
!*.yaml
```
to
```
!*.yaml
*
```

If that's useful.  I'm also willing to add more negative tests if needed.

I didn't reach for any of the go `git ignore` libraries because I assumed the goal was to minimize dependencies.

https://github.com/helm/helm/issues/3622#issuecomment-1247153893 says:

> > There is no support for '!' as a special leading sequence.
>
> is misleading - what does it mean "special leading sequence"? In contrast to the earlier statement, it is either a contradiction or it can also be mistakenly understood that (given the earlier point about whitespace in filenames) these are files starting with an exclamation mark (!). Simply put: That sentence is semantically wrong.

This seems to be because the comment is written as `\!`, which is consumed by markdown to `!`.  I added an extra `\` to compensate.

**If applicable**:

Per discussion in https://github.com/helm/helm/issues/3622#issuecomment-956257458 and https://github.com/helm/helm/issues/3622#issuecomment-370947322, it's not clear that we _want_ to be backwards-compatible with the current implementation, which doesn't really work for multiple negative patterns.  (It allows only names which *union* those patterns, rather than names that match any one of the negative patterns.)


evankanderson:
@joejulian -- what do I need to do to get this reviewed?

joejulian:
> @joejulian -- what do I need to do to get this reviewed?

@evankanderson Can I share my wife's honey-do list so I can have more free time? 🤣

evankanderson:
Hearing that it is in the queue works for me.

I didn't know whether there was a process for setting the reviewer field, for example.  😁

evankanderson:
I rebased -- tests were previously passing and `go test ./pkg/ignore` continues to pass.

I'm not going to sink a lot of time into keeping this up to date until I get a reviewer ping, though I think this is a genuine improvement in the usability of combined negative and positive rules.

pkg/ignore/doc.go
```diff
@@ -63,6 +63,6 @@ Notable differences from .gitignore:
   - The globbing library is Go's 'filepath.Match', not fnmatch(3)
   - Trailing spaces are always ignored (there is no supported escape sequence)
   - The evaluation of escape sequences has not been tested for compatibility
-  - There is no support for '\!' as a special leading sequence.
+  - There is no support for '\\!' as a special leading sequence for files that begin with `!`
 */
```
eddymoulton, line 67:
```suggestion
  - The first filename match will determine if it is included or excluded, subsequent lines are ignored
  
Example:

	# This is valid .gitignore syntax to only include Chart.yaml and values.yaml, but invalid for .helmignore
	*
	!Chart.yaml
	!values.yaml
	
	# Instead, this syntax will work as .helmignore takes the first match per file
	!Chart.yaml
	!values.yaml
	*
*/
```

eddymoulton (commented):
@joejulian (or another maintainer :slightly_smiling_face:) I'm leaving these comments here to hopefully assist in making a decision for this PR simpler.

While the the change is potentially breaking in some edge cases, I agree with Evan that it is a geniune improvement and is worth merging into v3.

The linked issue contains a lot of people expecting it to work more like .gitignore, which this PR _will not_ solve. I've added a suggestion for some docs changes to make it clearer how you would achieve one of the common examples in the issue.
Based on that fact, we should still have an issue targetting v4 to more closely align .helmignore with .gitignore.


> I'm willing to add a special case for a single negated rule that basically translates...

The potentially breaking nature of this change can be almost entirely mitigated by adding the special case mentioned for a single negated rule - especially considering how unlikely it is that anyone got this working with multiple negated rules the way it currently functions. Bearing in mind my suggestion to revisit how this works in v4, I think adding in the special handling is a reasonable idea.


> This seems to be because the comment is written as \!, which is consumed by markdown to !. I added an extra \ to compensate.

I've read through this and I agree that this change matches the original intention :+1:

evankanderson:
Is there interest in this fix?

evankanderson:
@scottrigby -- this seems like it just needs review; it seems like this is being bumped along milestones, but is there anything I can do to get this landed?

jamesbradlee:
@evankanderson seems to be merge conflicts.

@mattfarina @scottrigby Any response on this?

> this seems like it just needs review; it seems like this is being bumped along milestones, but is there anything I can do to get this landed?
> — @evankanderson

evankanderson:
I'm happy to update this for the latest state if it can get landed (though I don't personally need it anymore).

jamesbradlee (approved):

evankanderson:
Just checking in -- any interest in fixing negative expressions to allow multiple to work correctly?

scottrigby:
@evankanderson this is not a bad idea IMO. The main issue is there have been a lot of other higher priority things to review. I'm moving milestone again, this time because Helm 3 is not accepting new features so this would be for Helm 4, which is scheduled to release in November this year. Would you be interested in joining one of the weekly helm dev meetings to discuss this feature?

evankanderson:
Sure!

TerryHowe (commented):
I think the copilot comments have some merit. I ran the tests and minor improvement on coverage, everything passes.
```
out.main:helm.sh/helm/v4/pkg/ignore/rules.go:90:								Ignore						84.2%
out.more-negativity:helm.sh/helm/v4/pkg/ignore/rules.go:90:								Ignore					84.6%

```

evankanderson:
Sorry for the delay on getting to these comments; since I worked around this a couple years ago, it hasn't been my top priority, but it seems like it could be handy for the next person.

While I was fixing the copilot comments, I looked at the semantics of `.gitignore` (since I'm assuming that's the pattern people will assume for `.helmignore`).  Unfortunately, the semantics of `.gitignore` is that the last matching rule wins, rather than the first.  Given that rules with multiple negative patterns never previously worked (they would always exclude everything, because one of the negative patterns would be treated as ignore), I switched the implementation to be closer to `.gitignore` and have a last-rule-wins evaluation.

This also simplified the documentation.

TerryHowe:
Last commit is not signed off. It might be easier for you to squash those commits.

evankanderson:
Oops, fixed the DCO (half my projects have CLAs, the other half don't).  I can squash if you want -- not sure what the preferred style here is.

TerryHowe (approved):
/lgtm

sabre1041 (approved):
LGTM

evankanderson:
Should this be merged?  It looks to have the necessary approvals.

scottrigby (changes requested):
@evankanderson I think we can not do this until Charts v3, because even if perhaps not intuitive, functional changes like this could break existing chart .helmignore behavior. But I do think changing .helmignore file targeting syntax and functionality to have parity with .gitignore would be a valuable change for Charts v3.

I thought about your PR a lot the night before last, and wrote up a HIP with what I think would be necessary, including the Chart apiVersion, and other requirements. Would appreciate your feedback on this, given that you've invested energy into this PR and clearly thought a lot about this topic too: https://github.com/helm/community/pull/426

The HIP does not include implementation details, but IMO we would do well to add – not only incremental, but – full .gitignore targeting parity starting with Charts v3. I did a little investigation into existing Go libraries and their limitations, and ultimately one of the best examples I could find is how Flux source-controller handled .gitignore parity for .sourceignore. Can share that anytime.

evankanderson:
As described in #3622, the _current_ behavior of `!` (retain this file) is not very useful with multiple negative patterns (because of the short-circuit -- any retained file would need to match _all_ of the negative patterns).

My initial solution was to implement something _different_ than gitignore patterns (effectively, "retain / `!` wins"), because of the concern about backwards compatibility.  I'm not sure why I switched to the gitignore pattern on the most recent round of feedback -- I think it was suggested by copilot and seemed encouraged, but it _does_ change the semantics of retain for the single negative pattern case -- you'd need to have the retain as the first line of the `.helmignore` file.

I'm happy to either revert to the "retain wins" behavior, or queue this for some sort of Helm 5 release -- sorry that this didn't get into Helm 4 when that was released earlier this year.

scottrigby:
> the current behavior of ! (retain this file) is not very useful with multiple negative patterns

While this may be true, changing behevior for current charts (apiVersion v2) could cause unintended effects for users of certain charts. However, Chart apiVersion v3 (Charts v3) can make behavior changes like this. IMO this would be a very big improvement and perfect for this.

> I'm happy to either revert to the "retain wins" behavior, or queue this for some sort of Helm 5 release -- sorry that this didn't get into Helm 4 when that was released earlier this year.

To be clear, Charts v3 will be coming out during the Helm 4 release lifecycle. This feature would not need to wait until Helm 5. Changing the .helmignore file behavior can not happen for Charts v2.

I'm removing the milestone only because we don't yet know what minor release Charts v3 will be part of yet.

Also, would still like your input on https://github.com/helm/community/pull/426

PR #11631, opened 2022-12-12
Add `helm get metadata` command

mih-kopylov:
Closes #11625

**What this PR does / why we need it**:
The PR introduces a new command `helm get metadata`

**Special notes for your reviewer**:

**If applicable**:
- [x] this PR contains unit tests
- [x] this PR has been tested for backwards compatibility

PR #12272, opened 2023-08-03
Revert "Add `helm get metadata` command"

joejulian:
Reverts helm/helm#11631

This fails linting, and after fixing the linting, it causes a segfault.

Will need to re-submit this PR.


scottrigby (approved):
👍

PR #12278, opened 2023-08-07
Fix helm may identify achieve of the application/x-gzip as application/vnd.ms-fontobject

### LINKED ISSUE #12261: Helm identifies the compressed package of the application/x-gzip type as application/vnd.ms-fontobject
Output of `helm version`: v3.10.3

Output of `kubectl version`: v1.25.3

### Problem Description：
Helm uses the native Go method to verify the package type, but occasionally identifies the `application/x-gzip` package as `application/vnd.ms-fontobject`,
For example, the following package matches both `application/x-gzip` and `application/vnd.ms-fontobject`. However, because the `application/vnd.ms-fontobject` type matches before `application/x-gzip`, 

**error：**
failed to load chart: file '/var/paas/cam-registry/charts/redis-clusterserver-service2404094491/rediscluster-24.1.110000008.tgz' does not appear to be a gzipped archive; got 'application/vnd.ms-fontobject' 

**Helm code：**
https://github.com/helm/helm/blob/main/pkg/chart/loader/archive.go#L79

```
// ensureArchive's job is to return an informative error if the file does not appear to be a gzipped archive.
//
// Sometimes users will provide a values.yaml for an argument where a chart is expected. One common occurrence
// of this is invoking `helm template values.yaml mychart` which would otherwise produce a confusing error
// if we didn't check for this.
func ensureArchive(name string, raw *os.File) error {
	defer raw.Seek(0, 0) // reset read offset to allow archive loading to proceed.

	// Check the file format to give us a chance to provide the user with more actionable feedback.
	buffer := make([]byte, 512)
	_, err := raw.Read(buffer)
	if err != nil && err != io.EOF {
		return fmt.Errorf("file '%s' cannot be read: %s", name, err)
	}
	if contentType := http.DetectContentType(buffer); contentType != "application/x-gzip" {
		// TODO: Is there a way to reliably test if a file content is YAML? ghodss/yaml accepts a wide
		//       variety of content (Makefile, .zshrc) as valid YAML without errors.

		// Wrong content type. Let's check if it's yaml and give an extra hint?
		if strings.HasSuffix(name, ".yml") || strings.HasSuffix(name, ".yaml") {
			return fmt.Errorf("file '%s' seems to be a YAML file, but expected a gzipped archive", name)
		}
		return fmt.Errorf("file '%s' does not appear to be a gzipped archive; got '%s'", name, contentType)
	}
	return nil
}
```

**the go source code：**

```
// Font types
	&maskedSig{
		// 34 NULL bytes followed by the string "LP"
		pat: []byte("\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00LP"),
		// 34 NULL bytes followed by \xF\xF
		mask: []byte("\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xFF\xFF"),
		ct:   "application/vnd.ms-fontobject",
	},
	&exactSig{[]byte("\x00\x01\x00\x00"), "font/ttf"},
	&exactSig{[]byte("OTTO"), "font/otf"},
	&exactSig{[]byte("ttcf"), "font/collection"},
	&exactSig{[]byte("wOFF"), "font/woff"},
	&exactSig{[]byte("wOF2"), "font/woff2"},

	// Archive types
	&exactSig{[]byte("\x1F\x8B\x08"), "application/x-gzip"},
	&exactSig{[]byte("PK\x03\x04"), "application/zip"},
```

heijian123:
Fixes:  #12261

I think the `ensureArchive` method should be hardened and the judgment of whether the achieve is `application/x-gzip` should be added. `contentType` can be used as an error message for the user.

**If applicable**:

### DIFF
--- a/pkg/chart/loader/archive.go
+++ b/pkg/chart/loader/archive.go
@@ -85,7 +85,10 @@ func ensureArchive(name string, raw *os.File) error {
 	if err != nil && err != io.EOF {
 		return fmt.Errorf("file '%s' cannot be read: %s", name, err)
 	}
-	if contentType := http.DetectContentType(buffer); contentType != "application/x-gzip" {
+
+	// Helm may identify achieve of the application/x-gzip as application/vnd.ms-fontobject.
+	// Fix for: https://github.com/helm/helm/issues/12261
+	if contentType := http.DetectContentType(buffer); contentType != "application/x-gzip" && !isGZipApplication(buffer) {
 		// TODO: Is there a way to reliably test if a file content is YAML? ghodss/yaml accepts a wide
 		//       variety of content (Makefile, .zshrc) as valid YAML without errors.
 
@@ -96,6 +99,12 @@ func ensureArchive(name string, raw *os.File) error {
 		return fmt.Errorf("file '%s' does not appear to be a gzipped archive; got '%s'", name, contentType)
 	}
 	return nil
+}
+
+// isGZipApplication checks whether the achieve is of the application/x-gzip type.
+func isGZipApplication(data []byte) bool {
+	sig := []byte("\x1F\x8B\x08")
+	return bytes.HasPrefix(data, sig)
 }
 
 // LoadArchiveFiles reads in files out of an archive into memory. This function


gjenkins8 (commented):
Personally, I would switch the order of `ensureArchive` and `LoadArchive`. Something like:

```
	c, err := LoadArchive(raw)
	if err != nil {
            err = ensureArchive(name, raw)
	    if err != nil {
	 	return nil, err
	    }
	
            if err == gzip.ErrHeader {
			return nil, fmt.Errorf("file '%s' does not appear to be a valid chart file (details: %s)", name, err)
	    }
	}
```

And then rename `ensureArchive` to something more appropriate e.g. `reportContentTypeError`.

This means we only try to sniff the file content if and only if the actual loading fails. Which is what I see `ensureArchive` trying to do: report a more informative error iff the file is not an archive.

heijian123:
> Personally, I would switch the order of `ensureArchive` and `LoadArchive`. Something like:
> 
> ```
> 	c, err := LoadArchive(raw)
> 	if err != nil {
>             err = ensureArchive(name, raw)
> 	    if err != nil {
> 	 	return nil, err
> 	    }
> 	
>             if err == gzip.ErrHeader {
> 			return nil, fmt.Errorf("file '%s' does not appear to be a valid chart file (details: %s)", name, err)
> 	    }
> 	}
> ```
> 
> And then rename `ensureArchive` to something more appropriate e.g. `reportContentTypeError`.
> 
> This means we only try to sniff the file content if and only if the actual loading fails. Which is what I see `ensureArchive` trying to do: report a more informative error iff the file is not an archive.

Thanks for the review. But I think verifying the file type before `LoadArchive` might be a more reasonable step. And `LoadArchive` may return an error for other reasons than the wrong file type.

pkg/chart/loader/archive.go
```diff
@@ -85,7 +85,10 @@ func ensureArchive(name string, raw *os.File) error {
 	if err != nil && err != io.EOF {
 		return fmt.Errorf("file '%s' cannot be read: %s", name, err)
 	}
-	if contentType := http.DetectContentType(buffer); contentType != "application/x-gzip" {
+
+	// Helm may identify achieve of the application/x-gzip as application/vnd.ms-fontobject.
+	// Fix for: https://github.com/helm/helm/issues/12261
+	if contentType := http.DetectContentType(buffer); contentType != "application/x-gzip" && !isGZipApplication(buffer) {
```
joejulian, line 91: Have you looked to see if this is an upstream bug? Why is [DetectContentType](https://pkg.go.dev/net/http#DetectContentType) returning vnd.ms-fontobject?

heijian123: This is not an upstream bug. I occasionally encounter this bug. The achieve I use meets both the `application/x-gzip` and `application/vnd.ms-fontobject` types. However, because `application/vnd.ms-fontobject` matches first, so `DetectContentType` returning vnd.ms-fontobject

Here's the questionable achieve I used, which can be decompressed and tested.
[rediscluster-24.1.110000008.tgz.zip](https://github.com/helm/helm/files/12299544/rediscluster-24.1.110000008.tgz.zip)

gjenkins8: @joejulian -- `DetectContentType` has a bunch of heuristics:

https://github.com/helm/helm/issues/12261#issuecomment-1661247878

joejulian: Thanks!

joejulian (approved):

pkg/chart/loader/archive.go
jouve: zip != gzip, now the .tgz inside the zip file : rediscluster-24.1.110000008.tgz would be detected correctly as application/x-gzip

PR #12298, opened 2023-08-11
replace hashicorp multierror with core errors

joejulian:
This just moves from using the hashicorp multierrors pkg to the core errors pkg. One thing this does is reduce the licensing burden.

**If applicable**:


yxxhero (approved):
LGTM

scottrigby:
we should have merged your PR a long time ago, but since then the dep has been removed (https://github.com/helm/helm/pull/30781). We also have a depguard rule for this now too, which we should do for packages that are unmaintained or we want to explicitly avoid (see https://github.com/helm/helm/pull/13460#pullrequestreview-2784779178). But yay, retroactive thanks!

PR #12300, opened 2023-08-13
Adding separate timeout for hooks

### LINKED ISSUE #11983: Separate Hook Timeout
I am using pre install hooks to set an environment for the chart and post-delete hooks to clean up, it is working really well but one question I had was is there a possibility of having a separate timeout for hooks as to the chart in general? Or even the concept of an install timeout and an upgrade timeout?

I could just set the timeout really high but I like having them at a sensible amount for the upgrades to find issues quickly but that means that the pre install hooks take too long and the install times out. Just wondered if this had discussed before or there's a better way for me to do this.

Thank you for you time

BenHesketh21:
closes #11983 

This PR adds another timeout specifically for hooks, it allows for long-running hooks to complete with an installation/upgrade/deletion to succeed without having to set a large general timeout when the normal K8s objects in the chart do not need the large timeout and there would lose the ability to find issues quickly if they do timeout.

I have added the timeout option to every operation that had hooks involved, the only thing I wasn't sure on was the release testing operation, the timeout seemed to only be used for hooks and therefore I can possibly change that timeout to be a hook time out to be more consistent or I can leave that operation with the standard time out or I can leave it with both incase the standard timeout is needed in future for something else.

**If applicable**:


gjenkins8 (changes requested):
Thanks for the PR. At a minimum, Helm will need to remain compatible with existing usage.

I think the way to do this, is to continue to utilize `--timeout` for hooks, and only utilize `--hook-timeout` for hooks if it is specified. Otherwise, a user today who is expecting to see `--timeout` used for their hooks, will be "suprised" when `--hook-timeout` (which they may not know about) now takes effect.

pkg/action/upgrade.go
```diff
@@ -64,6 +64,8 @@ type Upgrade struct {
 	SkipCRDs bool
 	// Timeout is the timeout for this operation
 	Timeout time.Duration
+	// HookTimeout is the timeout for any hooks in this operation
+	HookTimeout time.Duration
```
gjenkins8, line 68: Additionally to my note on flag behavioral compatibility, Helm SDK users would get a default value of zero (due to zero initialization) for `HookTimeout` (here and Install/Uninstall/Rollback too).

BenHesketh21:
@gjenkins8 Thank you for your time, I'll get on with the suggestions

BenHesketh21:
@gjenkins8 I have changed the default value to be 0 seconds and then with each operation before using the `--hook-timeout` value checking to see if it set and if not then set it to the same value as the `--timeout` value. This should cover already working instances where only `--timeout` is set and the behaviour should remain the same but gives the option to set the other kind

pkg/action/upgrade.go
dyrone: Good explanation.
Thanks.

BenHesketh21: This should be fine since with the CLI the default is now zero and it is only used if it is set, and Helm uses the standard timeout if it is not set.

cmd/helm/install.go
```diff
@@ -177,7 +177,8 @@ func addInstallFlags(cmd *cobra.Command, f *pflag.FlagSet, client *action.Instal
 	f.BoolVar(&client.Force, "force", false, "force resource updates through a replacement strategy")
 	f.BoolVar(&client.DisableHooks, "no-hooks", false, "prevent hooks from running during install")
 	f.BoolVar(&client.Replace, "replace", false, "re-use the given name, only if that name is a deleted release which remains in the history. This is unsafe in production")
-	f.DurationVar(&client.Timeout, "timeout", 300*time.Second, "time to wait for any individual Kubernetes operation (like Jobs for hooks)")
+	f.DurationVar(&client.Timeout, "timeout", 300*time.Second, "time to wait for any individual Kubernetes operation")
+	f.DurationVar(&client.HookTimeout, "hook-timeout", 0*time.Second, "time to wait for any individual hook")
```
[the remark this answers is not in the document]
gjenkins8: (and others)

gjenkins8 (commented):
One minor comment from myself on the flag description. Maintainers will need to review as well. Thanks for updating the PR!

cmd/helm/install.go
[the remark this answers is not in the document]
BenHesketh21: @gjenkins8 Thank you again for taking a look, I've made this change in all relevant areas

BenHesketh21:
@gjenkins8 Thank you again for looking at my PR. Is there anything I'm missing? Just wondering why a maintainer hasn't commented anything

Issue #11592, opened 2022-12-01
Disable warning when Kube Config is group/world-readable


theAkito:
Pretty much https://github.com/helm/helm/issues/11083 and https://github.com/helm/helm/issues/9115.

My configuration file is a symlink, which does not have permissions, but according to `ls` & whatever Helm uses internally, it looks like it has `777` permissions, which is ultimately not true, though.

Changing permissions of a symlink only works on BSD and whatever operating systems use the same tools. Linux is not part of that group.

I would like to be able to disable this permanent warning, because it is wrong. The file itself is safe. It's just the symbolic link, which appears as if it were accessible by everyone.

et304383:
Wow.  Just came here from Google.  This warning message should be based on the real file path, not the symlink.  It should be easy to check if it's a symlink first.

At the very least, provide a way to disable this annoying warning.

theAkito:
> At the very least, provide a way to disable this annoying warning.

There was some talk about "our audit team recommended to never hide this warning" or something, but in this case, it is utterly unnecessary, because it is simply *wrong*. So, this warning should just be removed, if a symbolic link leads to a properly permissioned file.

xodmqjw:
Maybe I'm missing some point but in [this code](https://github.com/helm/helm/blob/3636f6824757ff734cb265b8770efe48c1fb3737/cmd/helm/root_unix.go#L45) permissions already are checking using `stat` syscall. Thus there should be no warnings because this syscall gets status of a file, not a symlink (according to the [docs](https://man7.org/linux/man-pages/man2/lstat.2.html) )

Here is a test to check:
```go
func TestCheckSymlinkPerms(t *testing.T) {
	tdir := t.TempDir()
	tfile := filepath.Join(tdir, "testconfig")
	_, err := os.OpenFile(tfile, os.O_CREATE|os.O_APPEND|os.O_RDWR, 0400)
	if err != nil {
		t.Errorf("Failed to create temp file: %s", err)
	}

	linkfile := filepath.Join(tdir, "testconfig.symlink")
	_ = os.Symlink(tfile, linkfile)

	tconfig := settings.KubeConfig
	settings.KubeConfig = linkfile
	defer func() { settings.KubeConfig = tconfig }()

	text, err := checkPermsStderr()
	if err != nil {
		t.Fatalf("could not read from stderr: %s", err)
	}
	if text != "" {
		t.Errorf("Expected no warning messages, got %q", text)
	}
}
```

[...]

Issue #11083, opened 2022-06-27
don't warn about group/world-readable KUBECONFIG if it contains no secrets


flokli:
helm currently warns on every command:

```
WARNING: Kubernetes configuration file is group-readable. This is insecure. Location: …
WARNING: Kubernetes configuration file is world-readable. This is insecure. Location: …
```

The warning isn't very useful.

In my case, the kubeconfig uses OIDC for logins, so it doesn't contain any secrets.
Instead of embedding client key data certs (via `client-key-data`), `kubeconfig` might also contain paths to files containing secrets (via `client-key`). Permissions for these are not checked either.

This check should probably simply be removed. It produces both false positives and false negatives, and understanding `kubeconfig` shouldn't be a concern of `helm` either, but of the Kubernetes SDK.

related: https://github.com/helm/helm/issues/9115

yxxhero:
@flokli This is only a reminder. you can change the permission of KUBECONFIG to remove it.

jetersen:
@yxxhero Yes you can change the permissions of kubeconfig if you follow the extra steps: https://github.com/helm/helm/issues/9115#issuecomment-1137431441

Might be easy on some OSes but Windows and WSL requires extra step also when tools like rancher-desktop suggest to symlink the kubeconfig between Windows and WSL so you do not have to maintain two configs.

By default in WSL it cannot see file permission metadata on Windows files you have to enable that.

sudermanjr:
Personally I think that an environment variable to suppress these warnings is a cleaner approach than simply removing them. Something along the lines of `SUPPRESS_KUBECONFIG_PERMS_WARNING` or something explicit like that.

flokli:
The main case here is that the `kubeconfig` files doesn't always contain secrets:

 - Secrets can either live in other files, that are referred to from `kubeconfig` (`client-key` instead of `client-key-data`)
 - Secrets can not exist at all (OIDC)

It shouldn't be a concern of helm to try to understand all the various formats.

jetersen:
if it is not a concern for helm than it should remove the readable check.
Or at the very least provide a way to suppress the warning.

colin-arnott-xero:
as the description implies, this check is quite flaky, thus if you have the flexibility to use the `KUBECONFIG` environment variable, providing the "insecure" one second[0] will work around this check without making any permission changes.

0] you can even use an empty first file `KUBECONFIG=":path/to/insecure"`

flokli:
[...]

Issue #9115, opened 2020-12-10
WARNING: Kubernetes configuration file is group/world-readable


ydekel6:
* Output of `helm version`:
version.BuildInfo{Version:"v3.4.1", GitCommit:"c4e74854886b2efe3321e185578e6db9be0a6e29", GitTreeState:"clean", GoVersion:"go1.14.11"}
Output of `kubectl version`:
* Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.3", GitCommit:"1e11e4a2108024935ecfcb2912226cedeafd99df", GitTreeState:"clean", BuildDate:"2020-10-14T12:50:19Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.3", GitCommit:"1e11e4a2108024935ecfcb2912226cedeafd99df", GitTreeState:"clean", BuildDate:"2020-10-14T12:41:49Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"linux/amd64"}
* Platform: ubuntu 20.4 (latest through microsoft store as to 10/12/2020) on windows 10 pro (clean install of build 19042.662) through wsl2.

when doing any `helm` command you always these lines:
WARNING: Kubernetes configuration file is group-readable. This is insecure. Location: /home/username/.kube/config
WARNING: Kubernetes configuration file is world-readable. This is insecure. Location: /home/username/.kube/config

pretty new to the wsl2/helm/k8s but I couldn't find any way to remove those  warnings. probably a duplicate of #8776

bacongobbler:
Your `~/.kube/config` should only be readable by you. Use `chmod` to change the file's read/write permissions.

```
><> ls -al ~/.kube/
total 20
drwxr-xr-x  3 bacongobbler bacongobbler 4096 Dec  1 09:43 .
drwxr-xr-x 23 bacongobbler bacongobbler 4096 Dec  9 13:18 ..
drwxr-x---  4 bacongobbler bacongobbler 4096 Nov 10 10:32 cache
-rw-------  1 bacongobbler bacongobbler 5514 Dec  1 09:43 config
```

To demonstrate, I'll make the file group-readable, then world-writeable. Then show how to remove those permissions.

```
><> chmod g+r ~/.kube/config
><> helm version
WARNING: Kubernetes configuration file is group-readable. This is insecure. Location: /home/bacongobbler/.kube/config
[...]

Issue #8776, opened 2020-09-21
Warning after upgrading to 3.3.3


pieterv-icloud-com:
Output of `helm version`:

version.BuildInfo{Version:"v3.3.3", GitCommit:"55e3ca022e40fe200fbc855938995f40b2a68ce0", GitTreeState:"dirty", GoVersion:"go1.15.2"}

Output of `kubectl version`:

Client Version: version.Info{Major:"1", Minor:"10", GitVersion:"v1.10.3", GitCommit:"2bba0127d85d5a46ab4b778548be28623b32d0b0", GitTreeState:"clean", BuildDate:"2018-05-21T09:17:39Z", GoVersion:"go1.9.3", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.6", GitCommit:"96fac5cd13a5dc064f7d9f4f23030a6aeface6cc", GitTreeState:"clean", BuildDate:"2019-08-19T11:05:16Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"linux/amd64"}

Cloud Provider/Platform (AKS, GKE, Minikube etc.): Rancher

After upgrading to helm 3.3.3 we are getting the following warning when doing the following helm upgrade:

helm upgrade --namespace cib-k8-magic-dev --install --values /workspace/3/s/dotnet-core/deployment/charts/hello-world-api-values.yaml --set replicaCount=3,support.contact=Support Contact,support.url=Support Email,snowCorrelationId=SNOWCORRELATIONID,smoketest.repository=harbor.intra.absa.co.za/base/appropriate/curl,image.repository=harbor.intra.absa.co.za/k8-magic/hello-world-api,image.tag=0.0.513,ingress.enabled=true,ingress.hosts[0].host=hello-world-api.cib-k8-magic-dev.cib-fx.sdc-nonprod.caas.absa.co.za,connectionstring=*** --wait --version 0.0.513 hello-world-api hello-world-api/hello-world-api --version 0.0.513

The upgrade works, but there appears to be a YAML Exception:

2020-09-21T09:09:27.4439528Z WARNING: Kubernetes configuration file is group-readable. This is insecure. Location: /workspace/_temp/helmTask/1600679351732/config
2020-09-21T09:09:27.4441316Z WARNING: Kubernetes configuration file is world-readable. This is insecure. Location: /workspace/_temp/helmTask/1600679351732/config
[...]

PR #12316, opened 2023-08-18
Disable warning when Kube Config is group-readable.

fungiboletus:

`helm` warns when the kube config is group readable as it may be insecure. This is not necessarily the case and an annoying warning as discussed in #8776, #9115, #11083, and #11592 

My proposed change is to ignore the warning for group readable. Ignoring the warning for word readable would require more work to detect whether it's secure or not. I would guess it's very often insecure so I left it.

Alternatively, the whole warning feature could be removed. I understood that you want the security warning to stay because [one member said](https://github.com/helm/helm/issues/8776#issuecomment-812146980
):

> To be clear, will not accept PRs to suppress the security warning. This issue was pointed out by a security audit sponsored by the CNCF, and the warning was the security team’s recommendation.

But I had already opened the pushed my commit when I read this comment, and I kinda disagree with the idea. Sometimes security audits are not correct and I think it's fair to dismiss one of their recommendations.

Feel free to reject the pull request. I would understand. And thanks for your work. 

**If applicable**:

### DIFF
--- a/cmd/helm/root_unix.go
+++ b/cmd/helm/root_unix.go
@@ -49,9 +49,6 @@ func checkPerms() {
 	}
 
 	perm := fi.Mode().Perm()
-	if perm&0040 > 0 {
-		warning("Kubernetes configuration file is group-readable. This is insecure. Location: %s", kc)
-	}
 	if perm&0004 > 0 {
 		warning("Kubernetes configuration file is world-readable. This is insecure. Location: %s", kc)
 	}
--- a/cmd/helm/root_unix_test.go
+++ b/cmd/helm/root_unix_test.go
@@ -50,7 +50,7 @@ func checkPermsStderr() (string, error) {
 func TestCheckPerms(t *testing.T) {
 	tdir := t.TempDir()
 	tfile := filepath.Join(tdir, "testconfig")
-	fh, err := os.OpenFile(tfile, os.O_CREATE|os.O_APPEND|os.O_RDWR, 0440)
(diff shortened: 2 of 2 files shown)


joejulian:
This has been discussed by the maintainers and there is a consensus that the security audit's findings will not be overruled.

johanot:
> Sometimes security audits are not correct and I think it's fair to dismiss one of their recommendations.

Totally agree. That audit recommendation and subsequent decision is wrong. I already posted my kubeconfig for everyone to see here: https://github.com/helm/helm/issues/11592#issuecomment-1484888567 (i.e. world-readable), and it's _not_ a security risk. The helm error message claims it is.

Some people say that helm is a beginner tool, and as such, it should help beginners being more secure. My response to that: Please be so kind and don't tell lies to our beginners, tell them how security _really_ works. We will get better software in the end that way. At my day job, I seriously have juniors at my desk asking me to take a look at their screens, because Helm is telling them "something disturbing" and they are unsure whether they're doing something wrong. I have to tell them: "Never mind that, Helm is just being weird".

On #11592, others discuss how to follow symlinks and misc. methods to drill down an tell whether the kubeconfig permissions really are insecure. Would I accept the error if helm was _actually_ capable of introspecting my auth method and revealing that I expose secrets in an insecure manner? Maybe yes. Would I think it's worth the effort to extend the current check code with hundreds, maybe thousands, of lines of code to support every possible permutation of kubernetes auth methods and secret references? Probably not, which likely is why kubectl and other tools doesn't try to be smart here.

PR #12321, opened 2023-08-22
fix(helm): fix GetPodLogs, the hooks should be sorted before get the logs of each hook

### LINKED ISSUE #12320: Bug: `helm test --logs` fails to show logs when one of multiple tests fails
## What happened
I defined multiple tests in a helm chart. And when I use `helm test --logs` to execute test, the first one failed, but there are no logs shown

Here is a sample for reproducing:
```sh
helm create foo
rm foo/templates/*
rm foo/templates/tests/*

cat <<EOF > foo/templates/tests/test-cases.yaml
apiVersion: v1
kind: Pod
metadata:
  name: "bb-test"
  annotations:
    "helm.sh/hook": test
spec:
  containers:
    - name: success
      image: ubuntu:22.04
      command:
      - sh
      - -c
      - |
        echo hello from success
        exit 0
  restartPolicy: Never
---
apiVersion: v1
kind: Pod
metadata:
  name: "aa-test"
  annotations:
    "helm.sh/hook": test
spec:
  containers:
    - name: failure
      image: ubuntu:22.04
      command:
      - sh
      - -c
      - |
        echo hello from failure
        exit 1
  restartPolicy: Never
EOF

helm install foo ./foo
helm test foo --logs
```

And you can see the output is:
```
NAME: foo
LAST DEPLOYED: Tue Aug 22 01:40:23 2023
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE:     aa-test
Last Started:   Tue Aug 22 01:40:30 2023
Last Completed: Tue Aug 22 01:40:32 2023
Phase:          Failed

Error: unable to get pod logs for bb-test: pods "bb-test" not found
```

But actually the pod `aa-test` has finished running:
```sh
# k logs aa-test

hello from failure
```

## Possible solution
I think here is the root cause of this issue.

The hooks are sorted by name and weight at `execHook`:
https://github.com/helm/helm/blob/b2b2145388ad324683ff51efca08ad6c90fdbeb6/pkg/action/hooks.go#L41-L44

But they are not sorted at `GetPodLogs`
https://github.com/helm/helm/blob/b2b2145388ad324683ff51efca08ad6c90fdbeb6/pkg/action/release_testing.go#L109-L115

LuBingtan:
closes #12320 

### DIFF
--- a/pkg/action/release_testing.go
+++ b/pkg/action/release_testing.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"sort"
 	"time"
 
 	"github.com/pkg/errors"
@@ -112,7 +113,9 @@ func (r *ReleaseTesting) GetPodLogs(out io.Writer, rel *release.Release) error {
 		return errors.Wrap(err, "unable to get kubernetes client to fetch pod logs")
 	}
 
-	for _, h := range rel.Hooks {
+	hooksByWight := append([]*release.Hook{}, rel.Hooks...)
+	sort.Stable(hookByWeight(hooksByWight))
+	for _, h := range hooksByWight {
 		for _, e := range h.Events {
 			if e == release.HookTest {
 				if contains(r.Filters[ExcludeNameFilter], h.Name) {


joejulian (approved):
Elegant. I was having the same problem and have been putting off trying to find a fix. Thank you!

PR #12341, opened 2023-08-28
Add support for encryption of Helm release body

fiscafusca:
This PR adds the possibility to encrypt and decrypt release bodies. The encryption algorithm and key can be respectively set via the environment variables `HELM_RELEASE_ENCRYPTION_ALGO` and `HELM_RELEASE_ENCRYPTION_KEY`. 

Currently, it only supports AES encryption in Galois/Counter Mode, but the code leaves room for supporting other algorithms in the future. 

The rationale behind these changes lies in the need to secure Helm release information. Releases can contain sensitive data, such as passwords and other keys, which must be protected not only at rest, but also during transport. Encrypting the release on client side would add an additional layer of security to send and store sensitive data in a secure way - given that the encryption key is managed correctly.  

When the environment variables are unset, the behavior of the `encodeRelease` and `decodeRelease` functions is still the same as before. 

Feedback is more than welcome, and if you're interested in moving forward I'm open to any suggestion to improve the proposal ☺️ 

**If applicable**:

### DIFF
--- a/pkg/storage/driver/util.go
+++ b/pkg/storage/driver/util.go
@@ -19,9 +19,14 @@ package driver // import "helm.sh/helm/v3/pkg/storage/driver"
 import (
 	"bytes"
 	"compress/gzip"
+	"crypto/aes"
+	"crypto/cipher"
+	"crypto/rand"
 	"encoding/base64"
 	"encoding/json"
+	"fmt"
 	"io"
+	"os"
 
 	rspb "helm.sh/helm/v3/pkg/release"
 )
@@ -39,12 +44,25 @@ func encodeRelease(rls *rspb.Release) (string, error) {
 	if err != nil {
 		return "", err
 	}
+
+	encryptionAlgorithm := os.Getenv("HELM_RELEASE_ENCRYPTION_ALGO")
+	encryptionKey := os.Getenv("HELM_RELEASE_ENCRYPTION_KEY")
+
+	var data []byte
+
+	switch encryptionAlgorithm {
+	case "aes":
+		data, err = encryptDataAES(b, []byte(encryptionKey))
+	default:
+		data = b
+	}
+
 	var buf bytes.Buffer
 	w, err := gzip.NewWriterLevel(&buf, gzip.BestCompression)
 	if err != nil {
 		return "", err
 	}
-	if _, err = w.Write(b); err != nil {
+	if _, err = w.Write(data); err != nil {
 		return "", err
 	}
 	w.Close()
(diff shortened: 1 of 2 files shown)


joejulian:
Thanks for the PR, I have a few questions to start with:

This may sound rude, but I don't mean it to be. Why do we need this? You can already store releases in secrets. You can use a number of secret stores with secrets and if your kubernetes cluster is configured properly, you have encryption at rest. Access control should prevent unauthorized access to the secret (or IAM, etc for external secret stores). What does this bring that can't already be done (arguably better) using existing tools?

How do you transition from not-encrypted to encrypted (or vice versa) on upgrade?
What happens if a user tries to use an older version of helm with the encrypted release, what's their experience?

I do wonder if this should go through the [hip process](https://github.com/helm/community/blob/main/hips/hip-0001.md) to get some feedback before you go too much farther.

fiscafusca:
Hi @joejulian thanks for the feedback! And no, you didn't sound rude. It's my fault for not providing enough background.

The scope of the changes introduced by this PR is primarily targeted at scenarios where the SQL backend is enabled. While you're absolutely correct that releases can be securely stored in secrets, there are certain scenarios where this might not be the most optimal solution:
- **Release Information Size:** In some instances, the size of Helm releases' information might exceed 1MB, making it challenging to store effectively within secrets.
- **Cluster Secrets Management:** Storing Helm release data within secrets could potentially contribute to the cumulative size of cluster secrets, which in turn could lead to disruptions within the cluster due to limitations or performance considerations.
- **Compliance and Flexibility:** There could be cases where compliance requirements necessitate moving release information to a different location.

Leveraging SQL storage backends enables us to conserve space within etcd, which can become crucial as the history of deployments grows. This approach also allows for a longer history of deployments. 

By offering the possibility of client-side encryption, users gain the ability to safeguard their data during transport and while at rest when using SQL storage backends. This added layer of security can be invaluable in scenarios where other methods might not align with stringent compliance guidelines.

For starters, we could reduce the scope of this contribution to provide encryption only for SQL backends. 

Regarding transitioning and compatibility: we could employ a metadata table, which could store crucial information like the encryption key used and the version of Helm responsible for the encryption, ensuring a smooth transition process. For older Helm versions, recognizing encrypted data and displaying relevant errors could be implemented, along with storing the Helm version for future migrations. But we'd need to backport this to the previous point release.

Let me know what your thoughts are, in light of the additional context I've provided. 
If you believe it's necessary, I'm open to initiating the HIP process for further input and insights.

PR #12348, opened 2023-08-30
use dependabot's group support for k8s.io dependencies

joejulian:

It's safer to upgrade k8s.io dependencies as a group. Dependabot has added the ability to create dependency groups. This PR enables that feature for k8s.io dependencies.

**If applicable**:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,6 +5,14 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+    groups:
+      k8s.io:
+        patterns:
+          - "k8s.io/*"
+        exclude-patterns:
+          - "k8s.io/utils"
+          - "k8s.io/klog/v2"
+          - "k8s.io/kube-openapi"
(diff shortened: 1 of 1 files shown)


mattfarina (approved):

.github/dependabot.yml
```diff
@@ -5,6 +5,14 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+    group:
```
mattfarina, line 8: Isn't it `groups` instead of `group`? I've not used this. Just looking at [the docs](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups).

mattfarina (changes requested):

.github/dependabot.yml
joejulian: :facepalm:

mattfarina (approved):

PR #12358, opened 2023-09-01
improve get-helm-3 GH api interaction

bewing:
This commit solves two problems:  The Github API has extremely low rate-limits on requests that could cause issues if you have a large CI/CD farm behind a NAT gateway or similar that causes a large amount of requests to appear to come from the same IP.  By utilizing the environment variable GITHUB_TOKEN if present, CI/CD can pass in a GH personal access token to bypass severe rate limiting.

Additionally, the Github API is not determinisitic on the return format of its JSON -- it may or may not be compressed and/or single line.  We should try to more accurately extract the tag name value instead of assuming multi-line output.

### DIFF
--- a/scripts/get-helm-3
+++ b/scripts/get-helm-3
@@ -111,11 +111,19 @@ checkDesiredVersion() {
     local latest_release_url="https://api.github.com/repos/helm/helm/releases/latest"
     local latest_release_response=""
     if [ "${HAS_CURL}" == "true" ]; then
-      latest_release_response=$( curl -L --silent --show-error --fail "$latest_release_url" 2>&1 || true )
+      if [ "${GITHUB_TOKEN}" != ""]; then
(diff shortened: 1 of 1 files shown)


bewing:
Tested with cURL:
```
└$ ./get-helm-3
Could not retrieve the latest release tag information from https://api.github.com/repos/helm/helm/releases/latest: curl: (22) The requested URL returned error: 403
Failed to install helm
└$ export GITHUB_TOKEN=<redacted>
└$ ./get-helm-3
Downloading https://get.helm.sh/helm-v3.12.3-linux-amd64.tar.gz
Verifying checksum... Done.
Preparing to install helm into /usr/local/bin
[sudo] password for bewing:
```

bewing:
This PR becomes moot if #12396 and #12397 are merged, I believe.

mattfarina:
@bewing thanks for figuring out how to improve this.

mattfarina:
I think we're likely to go with this method to improve the script instead. https://github.com/helm/helm/pull/12397

It will circumvent the GitHub API all together. Next wednesday is when we hope to move forward with that.

PR #12362, opened 2023-09-04
Install new CRDs during upgrade

### LINKED ISSUE #11969: Helm Upgrade with Install Flag enabled doesn't install new crds.
When using `helm upgrade -i <>`, my crds don't seem to be installing when the release has already been installed. I'm guessing this is because we go through the upgrade flow instead of the install flow due to the release being installed. However, according to the docs here: https://helm.sh/docs/helm/helm_upgrade/, it should be default be installing crds? Would be nice to have a way of adding new crds on upgrade.

Output of `helm version`:
`version.BuildInfo{Version:"v3.11.2", GitCommit:"912ebc1cd10d38d340f048efaf0abda047c3468e", GitTreeState:"clean", GoVersion:"go1.18.10"}`

EKS

richardwhiuk:
`helm upgrade --install` currently skips installing all CRDs during upgrade, despite having a SkipCRDs option.

This PR updates the code so that we install new CRDs during upgrade, which allows an updated chart to provide a new CRD.

This is done in the same order as it would take place on an install

This closes #11969 and helps in ref #6581

Currently, if you install a chart, Helm will install new CRDs. If you upgrade to a chart which has new CRDs, Helm won't install them.

This is despite:
- https://helm.sh/docs/chart_best_practices/custom_resource_definitions/ implying that new CRDs will be installed on upgrade
- `helm upgrade` taking a `SkipCRDs` flag, which currently doesn't do anything (and claims: "if set, no CRDs will be installed when an upgrade is performed with install flag enabled. By default, CRDs are installed if not already present, when an upgrade is performed with install flag enabled"). This change makes that text accurate.

1) Note, in theory this might not be regarded as backwards compatible, because `helm upgrade <foo>` will now install CRDs. However, if the user didn't want CRDs installed, they could have passed `--skip-crds`.

2) Although this is flagged as a large change, the bulk of the change is just moving stuff from `install.go` to `actions.go`, so that it can be used in `upgrade.go`.

**If applicable**:

### DIFF
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -24,10 +24,13 @@ import (
 	"path/filepath"
 	"regexp"
 	"strings"
+	"time"
 
 	"github.com/pkg/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
+	"k8s.io/cli-runtime/pkg/resource"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
@@ -42,14 +45,14 @@ import (
 	"helm.sh/helm/v3/pkg/releaseutil"
 	"helm.sh/helm/v3/pkg/storage"
 	"helm.sh/helm/v3/pkg/storage/driver"
-	"helm.sh/helm/v3/pkg/time"
+	helmtime "helm.sh/helm/v3/pkg/time"
 )
 
 // Timestamper is a function capable of producing a timestamp.Timestamper.
 //
-// By default, this is a time.Time function from the Helm time package. This can
+// By default, this is a helmtime.Time function from the Helm time package. This can
 // be overridden for testing though, so that timestamps are predictable.
-var Timestamper = time.Now
+var Timestamper = helmtime.Now
 
 var (
 	// errMissingChart indicates that a chart was not provided.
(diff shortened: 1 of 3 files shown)


AntoineJac:
Hi all, it would be really useful to have this PR approved after testing as "Install new CRDs during upgrade" is frequent request from our client using our Helm Chart. Thanks

jouve:
The link you provide explicitly says that it does *not* supports upgrades in [Some caveats (and explanations)](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#some-caveats-and-explanations)

> There is no support at this time for upgrading or deleting CRDs using Helm.

I think this PR is interesting but it should be guarded by as switch (--force-upgrade-cdrs ?) and disabled by default for backward compatibility.

richardwhiuk:
@jouve 

Thanks for the feedback

> The link you provide explicitly says that it does _not_ supports upgrades in [Some caveats (and explanations)](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#some-caveats-and-explanations)
> 
> > There is no support at this time for upgrading or deleting CRDs using Helm.
> 
> I think this PR is interesting but it should be guarded by as switch (--force-upgrade-cdrs ?) and disabled by default for backward compatibility.

This PR **doesn't change that** - it doesn't **upgrade** or **delete** CRDs, it merely adds new ones.

The existing flags on helm already suggests this will be done, but prior to this change, this is not the case.

From `helm upgrade --help`.

> --skip-crds                                  if set, no CRDs will be installed when an upgrade is performed with install flag enabled. By default, CRDs are installed if not already present, when an upgrade is performed with install flag enabled

As such, I don't think a new command line option is sensible.

If a user wants to preserve the existing behaviour, they can run `helm upgrade --install --skip-crds`. If they want the new behaviour they can run `helm upgrade --install`.

We could add `--crds`, as a no-op (or worse, `--no-skip-crds` 🤮 ), and leave the current, bugged, behaviour, and fix the options text, but I'm not convinced this is helpful to anyone.

lheringer-bt:
Any chance this gets merged anytime soon?

alsyia:
Hi,

The documented behaviour for --skip-crds currently contradicts the actual behaviour as Helm does _not_ install new CRDs. Could the doc be fixed, or (better), this PR merged? :D

arthurdarcet:
really a shame to have closed PRs that fix years old invalid behavior :/ - the problem still exist in latest helm

benoittgt:
I can have a look a re-open a new PR if I am able to reproduce.

PR #12382, opened 2023-09-12
fix(pkg/lint): unmarshals Chart.yaml strictly

### LINKED ISSUE #12381: "helm lint" does not complain on duplicate keys in chartfile YAML
Currently, `helm lint` accepts chartfiles which have duplicate mapping keys (this constitutes invalid YAML), e.g. it accepts the following:

```yaml
apiVersion: v2
name: test
type: application
version: 0.1.0
appVersion: "1.16.0"
appVersion: "1.2.3"      <------ DUPLICATE KEY
dependencies:
  - name: trivy-operator
    repository: https://aquasecurity.github.io/helm-charts/
    version: 0.18.0
    version: 0.17.1      <------ DUPLICATE KEY
 ```
 
I think to fix this we can simply:
 - change [this function call](https://github.com/helm/helm/blob/main/pkg/chartutil/chartfile.go#L36) in pkg/chartutil to the [strict version](https://github.com/kubernetes-sigs/yaml/blob/master/yaml.go#L63) (`yaml.UnmarshalStrict`)
  - ~~check the resulting Metadata struct has a non-empty Name, a required field [according to this comment](https://github.com/helm/helm/blob/main/pkg/chart/metadata.go#L48)~~ this shouldn't be necessary, we can just use the error return value of the unmarshal function as we currently do

Happy to submit a PR with this change!

Minimal steps to reproduce:
  - `helm create testy-mctest`
  - update `Chart.yaml` with example above (or any key duplicated)
   - `helm lint`
   
Output of `helm version`:

```
version.BuildInfo{Version:"v3.10.2", GitCommit:"50f003e5ee8704ec937a756c646870227d7c8b58", GitTreeState:"clean", GoVersion:"go1.18.8"}
```

### PR DESCRIPTION
**Pull request by edbmiller**
When "helm lint" is run, it now errors on invalid Chart.yaml files, e.g. those with duplicate keys

Closes #12381

**What this PR does / why we need it**: Fixes a bug with the current `helm lint` command, where it does not error on Chart.yaml with duplicate mapping keys

**If applicable**:

### DIFF
--- a/pkg/lint/lint_test.go
+++ b/pkg/lint/lint_test.go
@@ -35,6 +35,7 @@ const badYamlFileDir = "rules/testdata/albatross"
 const goodChartDir = "rules/testdata/goodone"
 const subChartValuesDir = "rules/testdata/withsubchart"
 const malformedTemplate = "rules/testdata/malformed-template"
+const invalidChartFileDir = "rules/testdata/invalidchartfile"
 
 func TestBadChart(t *testing.T) {
 	m := RunAll(badChartDir, values, namespace).Messages
@@ -87,6 +88,16 @@ func TestInvalidYaml(t *testing.T) {
 	}
 	if !strings.Contains(m[0].Err.Error(), "deliberateSyntaxError") {
 		t.Errorf("All didn't have the error for deliberateSyntaxError")
+	}
+}
+
+func TestInvalidChartYaml(t *testing.T) {
+	m := All(invalidChartFileDir, values, namespace, strict).Messages
+	if len(m) != 1 {
+		t.Fatalf("All didn't fail with expected errors, got %#v", m)
+	}
+	if !strings.Contains(m[0].Err.Error(), "unable to parse YAML") {
+		t.Errorf("All didn't have the error for duplicate YAML keys")
 	}
 }
 
--- a/pkg/lint/lint_test.go
+++ b/pkg/lint/lint_test.go
@@ -96,7 +96,7 @@ func TestInvalidChartYaml(t *testing.T) {
 	if len(m) != 1 {
 		t.Fatalf("All didn't fail with expected errors, got %#v", m)
 	}
-	if !strings.Contains(m[0].Err.Error(), "failed to strictly parse chartfile") {
+	if !strings.Contains(m[0].Err.Error(), "failed to strictly parse chart metadata file") {
 		t.Errorf("All didn't have the error for duplicate YAML keys")
 	}
 }
--- a/pkg/lint/lint_test.go
+++ b/pkg/lint/lint_test.go
@@ -96,7 +96,7 @@ func TestInvalidChartYaml(t *testing.T) {
 	if len(m) != 1 {
 		t.Fatalf("All didn't fail with expected errors, got %#v", m)
 	}
-	if !strings.Contains(m[0].Err.Error(), "unable to parse YAML") {
+	if !strings.Contains(m[0].Err.Error(), "failed to strictly parse chartfile") {
 		t.Errorf("All didn't have the error for duplicate YAML keys")
 	}
 }
--- a/pkg/lint/lint_test.go
+++ b/pkg/lint/lint_test.go
@@ -92,7 +92,7 @@ func TestInvalidYaml(t *testing.T) {
 }
 
 func TestInvalidChartYaml(t *testing.T) {
-	m := All(invalidChartFileDir, values, namespace, strict).Messages
+	m := RunAll(invalidChartFileDir, values, namespace).Messages
 	if len(m) != 1 {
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by ed-dot-miller** [comment]
Hi @gjenkins8, could I get a review for this?

File: pkg/chartutil/chartfile.go [hunk 1]
```diff
@@ -33,7 +33,7 @@ func LoadChartfile(filename string) (*chart.Metadata, error) {
 		return nil, err
 	}
 	y := new(chart.Metadata)
-	err = yaml.Unmarshal(b, y)
+	err = yaml.UnmarshalStrict(b, y)
```
gjenkins8, line 36: Unfortunately, this will break non-lint usages of `LoadChartFile` which might have non-strict yaml (e.g. installing a chart). Helm maintains strong backwards compatibility guarantees in order to avoid breaking existing users.

(And to note `LoadChartfile` itself will follow API compatibility guarantees, and can't have its signature changed)

ed-dot-miller: Thanks for the update! I think the following preserves backward compatibility (of the other helm commands, as well as `helm lint` without the --strict option):
 - add a separate function `StrictLoadChartfile` to pkg/chartutil
 - use this in a WARNING-level linter rule in pkg/lint
 
 Would that be acceptable?

Brightside56:
Today I almost end up with deleted deployments in production because someone from engineers changed
```
dependencies:
  - xxx
  - yyy
```
to 
```
2dependencies:
  - xxx
  - yyy
```

pkg/chartutil/chartfile.go
ed-dot-miller: Have updated this PR according to the above comment - this catches duplicate keys (and @Brightside56's error) at WARNING-level. Could I have another review @gjenkins8?

gjenkins8 (approved):
Nice, thanks for the update. Changes LGTMe. Will have maintainers take a look

gjenkins8 (dismissed):
<dup>

edbmiller:
Hey @gjenkins8, any chance this could go into a minor 3.16.x release?

edbmiller:
Have rebased :+1:

pkg/lint/lint_test.go
```diff
@@ -91,6 +92,16 @@ func TestInvalidYaml(t *testing.T) {
 	}
 }
 
+func TestInvalidChartYaml(t *testing.T) {
+	m := All(invalidChartFileDir, values, namespace, strict).Messages
+	if len(m) != 1 {
+		t.Fatalf("All didn't fail with expected errors, got %#v", m)
+	}
+	if !strings.Contains(m[0].Err.Error(), "failed to strictly parse chartfile") {
```
sabre1041, line 100:
```suggestion
	if !strings.Contains(m[0].Err.Error(), "failed to strictly parse chart metadata file") {
```

sabre1041 (dismissed):
Everything works as expected. Small request to update the content of the error message

sabre1041:
@edbmiller Would you be able to resolve the conflicts so that this can be re-reviewed?

edbmiller:
Conflicts resolved 👍

gjenkins8 (dismissed):

edbmiller:
Bungled that, sorry - tests should pass now 👍

gjenkins8 (dismissed):

scottrigby:
@edbmiller it looks like @sabre1041 's request here https://github.com/helm/helm/pull/12382#discussion_r1950359433 was reverted in 0d1f9dad3e48dbeac4ba68dd05f38327feccd98b. Can you please change implement that change request again? I think you would just need to make the error message and expected error string match to pass CI. Otherwise this looks great!

scottrigby (commented):
> @edbmiller it looks like @sabre1041 's request here [#12382 (comment)](https://github.com/helm/helm/pull/12382#discussion_r1950359433) was reverted in [0d1f9da](https://github.com/helm/helm/commit/0d1f9dad3e48dbeac4ba68dd05f38327feccd98b). Can you please change implement that change request again? I think you would just need to make the error message and expected error string match to pass CI. Otherwise this looks great!

I believe you will have to not only change the string in `lint.TestInvalidChartYaml()` but this string should also match in `rules.validateChartYamlStrictFormat()`.

But also, tests are failing for a reason not related to this PR, which I believe have been fixed in `main`—can you please rebase onto `main` again?

edbmiller:
Thanks @scottrigby - fixed and rebased onto main 👍

gjenkins8 (approved):

scottrigby (approved):
Feedback addressed, automated tests pass, manual tests pass
LGTM 👍 Thanks @edbmiller

PR #12396, opened 2023-09-15
Write latest version to get.helm.sh bucket

z4ce:
The purpose of this change is leverage this output for the get-helm script.

It will provide the following benefits:
- Use true semver so a future release of a patch version so a future release of a patch version does not cause an unexpected minor/major version downgrade

- Avoid reliance on the github api which has rate limiting. This will also increase availability by reducing the number of systems the get-helm script depends on.

- Simplify the script so parsing json is not necessary

- Increase uptime by only requiring the Azure CDN to be available

https://github.com/helm/helm/issues/12351

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,6 +19,7 @@ jobs:
     steps:
       - name: Checkout source code
         uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # pin@v3.6.0
+        fetch-depth: 0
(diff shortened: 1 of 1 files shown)


z4ce:
@mattfarina this is the change I was talking about to start writing the latest version to the bucket

z4ce:
Script change once the file is available in the bucket: https://github.com/helm/helm/pull/12397

mattfarina (approved):

PR #12397, opened 2023-09-15
Update get-helm-3 to get version through get.helm.sh

z4ce:
Updates the script to use the new method of getting the latest version that avoids the github API and the associated rate limits. See the matching PR at https://github.com/helm/helm/pull/12396 for the server side change.


z4ce:
Marking as a draft so it doesn't get merged before we have the version info available on get.helm.sh

z4ce:
Now that we have `https://get.helm.sh/helm-latest-version` I tested this with wget and curl on fresh ubuntu containers. Seems to work as expected.

joejulian (approved):

PR #12437, opened 2023-09-29
Add missing with clause to release gh action

z4ce:
Fix missing with statement. Not sure why this missing passed an `act release -n`.


yxxhero (approved):

FrenchBen (commented):
LGTM - By default this is set to 1, do we actually need to fetch all? 
>0 indicates all history for all branches and tags

IMHO we can probably remove it in favor of the default

z4ce:
We need this for finding all the tags to find the latest release

mattfarina (approved):

PR #10533, opened 2022-01-08
Adds labels support for install and upgrade commands

dm3ch:
/closes https://github.com/helm/helm/issues/7007

**What this PR does / why we need it**:
Adding support of providing custom labels during install/update of release that could be used for selector filtering introduced in https://github.com/helm/helm/pull/8532.

**Special notes for your reviewer**:
This PR focused on adding custom label support during install/upgrade actions.
There was similar PRs before that, unfortunately, haven't been finished: https://github.com/helm/helm/pull/8546 https://github.com/helm/helm/pull/8099 https://github.com/helm/helm/pull/9087

In this PR I'm trying to make changes as small as possible, to make development and review easier.
My goal is to make selector functionality in the list command useable for install/upgrade actions with secret and configmap storage drivers.

I don't want to add SQL support in this particular PR, cause it seems it would at least double the number of changes.

TODO:
Functionality (support labels argument during following actions):
- [x] install
- [x] upgrade --install
- [x] upgrade (basic)
- [x] upgrade (keep labels on superseded releases)
- [x] tests for new functionality (don't know how to implement it better)
- [x] Secret storage driver
- [x] Configmap storage driver

Not planned in this PR:
- [x] SQL - support is not currently planned in this PR. But there are some related developments in https://github.com/helm/helm/pull/8546 that could be prepared as separate PR later.
- [x] Merge original release labels with upgrade labels

**UPD**:
Now I have implemented even features I haven't planned for this PR. So right now this PR is completely ready for review.

If you're interested why I've created a separate DB table (in SQL driver) for custom releases - my idea was that some day we may want to make release filtering on storage backend side (k8s and SQL support requests with filtering by labels/joined table). so I decided that it would be better to split it to a separate table instead of having it encoded somewhere in json, because data migration is always annoying.

**If applicable**:
- [ ] this PR contains documentation
- [x] this PR contains unit tests
- [x] this PR has been tested for backwards compatibility

PR #12447, opened 2023-10-03
Allow using label selectors for system labels for secrets and configmap

### LINKED ISSUE #12435: helm list --selector no longer working in version 3.13.0 cli
Using a `--selector` flag on `helm list` no longer works after updating to version 3.13.0 from 3.12.1

```
$ helm version
version.BuildInfo{Version:"v3.13.0", GitCommit:"825e86f6a7a38cef1112bfa606e4127a706749b1", GitTreeState:"clean", GoVersion:"go1.20.8"}

$ helm list --deployed -n kube-system --selector name=metrics-server
NAME    NAMESPACE       REVISION        UPDATED STATUS  CHART   APP VERSION
```

when using version 3.12.3 =
```
$ helm version

$ helm list --deployed -n kube-system --selector name=metrics-server
NAME            NAMESPACE       REVISION        UPDATED                                 STATUS          CHART                   APP VERSION
metrics-server  kube-system     3               2023-04-10 17:04:54.325699796 -0700 PDT deployed        metrics-server-3.9.0    0.6.3
```

`helm ls` _does_ work without the selector.

Output of `helm version`: version.BuildInfo{Version:"v3.13.0", GitCommit:"825e86f6a7a38cef1112bfa606e4127a706749b1", GitTreeState:"clean", GoVersion:"go1.20.8"}

Output of `kubectl version`:
Client Version: version.Info{Major:"1", Minor:"24+", GitVersion:"v1.24.9-eks-49d8fe8", GitCommit:"53e5004c71669ab3835f7b9c30f79dee82c30c17", GitTreeState:"clean", BuildDate:"2023-01-09T23:48:25Z", GoVersion:"go1.18.9", Compiler:"gc", Platform:"linux/amd64"}

### LINKED ISSUE #12517: Label selector in list command no longer working
Hello,

In Helm versions prior to 3.13.1 (for example 3.12.3) it was possible to list Helm releases by label selector, for example by Helm release name.

If I install a simple chart like incubator/raw like this:
```
$ helm repo add incubator https://charts.helm.sh/incubator
"incubator" has been added to your repositories
$ helm upgrade -i raw incubator/raw
Release "raw" does not exist. Installing it now.
WARNING: This chart is deprecated
NAME: raw
LAST DEPLOYED: Sun Oct 22 15:44:47 2023
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
```

If I then use label selector _name=raw_ using Helm 3.13.1 it does not find any the previously installed raw release:
```
$ helm version
version.BuildInfo{Version:"v3.13.1", GitCommit:"3547a4b5bf5edb5478ce352e18858d8a552a4110", GitTreeState:"clean", GoVersion:"go1.20.8"}
$ helm list -l name=raw
$
```

However if I downgrade Helm to 3.12.3 then the raw release is found:

```
$ helm version
version.BuildInfo{Version:"v3.12.3", GitCommit:"3a31588ad33fe3b89af5a2a54ee1d25bfe6eaa5e", GitTreeState:"clean", GoVersion:"go1.20.7"
$ helm list -l name=raw
NAME	NAMESPACE	REVISION	UPDATED                                 	STATUS  	CHART    	APP VERSION
raw 	default  	1       	2023-10-22 15:44:47.089882209 +0200 CEST	deployed	raw-0.2.5	0.2.3      
$
```

Is this an intentional change or a bug?

dm3ch:
backends.

Fixes #12435 that have been introduced by #10533

**If applicable**:

/fix #12517 /fix #12435

### DIFF
--- a/pkg/storage/driver/cfgmaps.go
+++ b/pkg/storage/driver/cfgmaps.go
@@ -107,7 +107,7 @@ func (cfgmaps *ConfigMaps) List(filter func(*rspb.Release) bool) ([]*rspb.Releas
 			continue
 		}
 
-		rls.Labels = filterSystemLabels(item.ObjectMeta.Labels)
+		rls.Labels = item.ObjectMeta.Labels
 
 		if filter(rls) {
 			results = append(results, rls)
--- a/pkg/storage/driver/cfgmaps.go
+++ b/pkg/storage/driver/cfgmaps.go
@@ -146,7 +146,7 @@ func (cfgmaps *ConfigMaps) Query(labels map[string]string) ([]*rspb.Release, err
 			cfgmaps.Log("query: failed to decode release: %s", err)
 			continue
 		}
-		rls.Labels = filterSystemLabels(item.ObjectMeta.Labels)
+		rls.Labels = item.ObjectMeta.Labels
 		results = append(results, rls)
 	}
 	return results, nil
--- a/pkg/storage/driver/cfgmaps_test.go
+++ b/pkg/storage/driver/cfgmaps_test.go
@@ -128,6 +128,16 @@ func TestConfigMapList(t *testing.T) {
 	if len(ssd) != 2 {
 		t.Errorf("Expected 2 superseded, got %d", len(ssd))
 	}
+	// Check if release having both system and custom labels, this is needed to ensure that selector filtering would work.
+	rls := ssd[0]
+	_, ok := rls.Labels["name"]
+	if !ok {
+		t.Fatalf("Expected 'name' label in results, actual %v", rls.Labels)
+	}
+	_, ok = rls.Labels["key1"]
+	if !ok {
+		t.Fatalf("Expected 'key-1' label in results, actual %v", rls.Labels)
+	}
 }
 
 func TestConfigMapQuery(t *testing.T) {
--- a/pkg/storage/driver/cfgmaps_test.go
+++ b/pkg/storage/driver/cfgmaps_test.go
@@ -136,7 +136,7 @@ func TestConfigMapList(t *testing.T) {
 	}
 	_, ok = rls.Labels["key1"]
 	if !ok {
-		t.Fatalf("Expected 'key-1' label in results, actual %v", rls.Labels)
+		t.Fatalf("Expected 'key1' label in results, actual %v", rls.Labels)
 	}
 }
 
--- a/pkg/storage/driver/secrets_test.go
+++ b/pkg/storage/driver/secrets_test.go
@@ -128,6 +128,16 @@ func TestSecretList(t *testing.T) {
 	if len(ssd) != 2 {
 		t.Errorf("Expected 2 superseded, got %d", len(ssd))
 	}
+	// Check if release having both system and custom labels, this is needed to ensure that selector filtering would work.
+	rls := ssd[0]
+	_, ok := rls.Labels["name"]
+	if !ok {
+		t.Fatalf("Expected 'name' label in results, actual %v", rls.Labels)
+	}
+	_, ok = rls.Labels["key1"]
+	if !ok {
+		t.Fatalf("Expected 'key-1' label in results, actual %v", rls.Labels)
+	}
 }
 
 func TestSecretQuery(t *testing.T) {
--- a/pkg/storage/driver/secrets_test.go
+++ b/pkg/storage/driver/secrets_test.go
@@ -136,7 +136,7 @@ func TestSecretList(t *testing.T) {
 	}
 	_, ok = rls.Labels["key1"]
 	if !ok {
-		t.Fatalf("Expected 'key-1' label in results, actual %v", rls.Labels)
+		t.Fatalf("Expected 'key1' label in results, actual %v", rls.Labels)
 	}
 }
 
--- a/pkg/storage/driver/sql.go
+++ b/pkg/storage/driver/sql.go
@@ -19,6 +19,7 @@ package driver // import "helm.sh/helm/v3/pkg/storage/driver"
 import (
 	"fmt"
 	"sort"
+	"strconv"
 	"time"
 
 	"github.com/jmoiron/sqlx"
@@ -368,6 +369,9 @@ func (s *SQL) List(filter func(*rspb.Release) bool) ([]*rspb.Release, error) {
 			s.Log("failed to get release %s/%s custom labels: %v", record.Namespace, record.Key, err)
 			return nil, err
 		}
+		for k, v := range getReleaseSystemLabels(release) {
+			release.Labels[k] = v
+		}
 
 		if filter(release) {
 			releases = append(releases, release)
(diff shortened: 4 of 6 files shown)


dm3ch:
Fixed the issue described in #12435 for Secret and Configmap backends and added tests to ensure it wouldn't be broken again.

This PR is still missing the same fix for SQL storage right now, it seems I need to finish it before review.

P.S. Ideally I believe it would be to use Query backend method for filtering (instead of filtering after List) and make filtering work on backend side, but it requires discussions and a lot of effort, so I don't want to change it now.

dm3ch:
Added fix for the SQL storage driver. I'm not happy with the way used to rebuild system labels (cause createdAt and modifiedAt are missing). It's possible to improve it if I rewrite the select statement but I would prefer not to do this.

pkg/storage/driver/cfgmaps_test.go
```diff
@@ -128,6 +128,16 @@ func TestConfigMapList(t *testing.T) {
...
 	}
+	// Check if release having both system and custom labels, this is needed to ensure that selector filtering would work.
+	rls := ssd[0]
+	_, ok := rls.Labels["name"]
+	if !ok {
+		t.Fatalf("Expected 'name' label in results, actual %v", rls.Labels)
+	}
+	_, ok = rls.Labels["key1"]
+	if !ok {
+		t.Fatalf("Expected 'key-1' label in results, actual %v", rls.Labels)
```
FrenchBen, line 139:
```suggestion
		t.Fatalf("Expected 'key1' label in results, actual %v", rls.Labels)
```

pkg/storage/driver/secrets_test.go
```diff
@@ -128,6 +128,16 @@ func TestSecretList(t *testing.T) {
...
 	}
+	// Check if release having both system and custom labels, this is needed to ensure that selector filtering would work.
+	rls := ssd[0]
+	_, ok := rls.Labels["name"]
+	if !ok {
+		t.Fatalf("Expected 'name' label in results, actual %v", rls.Labels)
+	}
+	_, ok = rls.Labels["key1"]
+	if !ok {
+		t.Fatalf("Expected 'key-1' label in results, actual %v", rls.Labels)
```
FrenchBen, line 139:
```suggestion
		t.Fatalf("Expected 'key1' label in results, actual %v", rls.Labels)
```

pkg/storage/driver/sql_test.go
```diff
@@ -157,6 +157,17 @@ func TestSQLList(t *testing.T) {
...
+
+	// Check if release having both system and custom labels, this is needed to ensure that selector filtering would work.
+	rls := ssd[0]
+	_, ok := rls.Labels["name"]
+	if !ok {
+		t.Fatalf("Expected 'name' label in results, actual %v", rls.Labels)
+	}
+	_, ok = rls.Labels["key1"]
+	if !ok {
+		t.Fatalf("Expected 'key-1' label in results, actual %v", rls.Labels)
```
FrenchBen, line 169:
```suggestion
		t.Fatalf("Expected 'key1' label in results, actual %v", rls.Labels)
```

pkg/storage/driver/cfgmaps.go
```diff
@@ -107,7 +107,7 @@ func (cfgmaps *ConfigMaps) List(filter func(*rspb.Release) bool) ([]*rspb.Releas
 			continue
 		}
 
-		rls.Labels = filterSystemLabels(item.ObjectMeta.Labels)
+		rls.Labels = item.ObjectMeta.Labels
```
FrenchBen, line 110: should the same change be made for `func (cfgmaps *ConfigMaps) Query(labels map[string]string) ([]*rspb.Release, error) {` ?

pkg/storage/driver/sql.go
```diff
@@ -681,3 +685,13 @@ func (s *SQL) getReleaseCustomLabels(key string, namespace string) (map[string]s
 
 	return filterSystemLabels(labelsMap), nil
 }
+
+// Rebuild system labels from release object
+func getReleaseSystemLabels(rls *rspb.Release) map[string]string {
```
FrenchBen, line 690: IMHO, this should be part of the `util.go` as all other functions are shared there.

FrenchBen (changes requested):
Good start to solving labels, but would like to see more consistency. 
Additionally is there any value to keeping `func filterSystemLabels` if it's not doing the intended work?

pkg/storage/driver/cfgmaps.go
dm3ch: When I started this PR, I didn't make this change, cause `Query` is used only for internal purposes as far as I understood (used during upgrade, delete and etc). So it doesn't affect user, cause not involved in helm list.

Right now I checked all code places where `filterSystemLabels()` is used and dropped it if it doesn't brake unit tests.
So far there're only two use cases where this function is used:
- Get operation -  in Get operation tests, expect the release after creation to be deepEquals. Potentially it's possible do drop it from there, but I don't know which side effects it can cause.
- SQL driver - in SQL driver system labels are stored separately in corresponding table columns, so we really need such filtering in SQL driver code.

pkg/storage/driver/sql.go
dm3ch: The reason why I've placed this code in SQL driver is that as far as I understand it's only useful for SQL driver.
To be honest, it's actually just a workaround to rebuild metadata (which in SQL driver is stored separately).
The reason why I use such an approach to rebuild metadata from the release is that I'm not sure that rewriting SQL queries is worth it, cause all useful fields (all metadata except revision, createdAt, modifiedAt) can be already extracted from the release body, so it seems it doesn't worth it to extract it from DB two times.

@FrenchBen what do you think:
- Is it worth it to rewrite SQL driver code to receive full metadata from corresponding columns instead of rebuilding only the useful part of metadata from the release body?
- Is it worth it to move this helper to util if it shouldn't be used in any other drivers (cause they already read this metadata from storage)?

dm3ch:
@FrenchBen Friendly reminder, that I'm awaiting answers in questions I asked in discussions to finish the work on this PR.

pkg/storage/driver/sql.go
z4ce: @joejulian @mattfarina just pinging a couple other maintainers here so this can be approved as it's fixing a pretty critical bug

pkg/storage/driver/cfgmaps.go
joejulian: Query is part of the public API and could be used by downstream users. Will the change, or lack thereof, cause confusion if this public function is used?

pkg/storage/driver/sql.go
joejulian: I think this is a reasonable reason to leave it put.

z4ce:
This maybe related as well https://github.com/helm/helm/issues/12521

joejulian (approved):
I was able to successfully repro the problem and this PR fixes it.

mattfarina (approved):

PR #12495, opened 2023-10-14
Fix hook install order

### LINKED ISSUE #12414: ServiceAccount is not created before Jobs/Pods
According to the [`InstallOrder`](https://github.com/helm/helm/blob/3a31588ad33fe3b89af5a2a54ee1d25bfe6eaa5e/pkg/releaseutil/kind_sorter.go#L31C1-L32C1), `ServiceAccount` should be installed before both `Pod` and `Job`. However, I've observed occasions where the `ServiceAccount` is not created before Jobs (or even standalone Pods).

# Steps to reproduce
1. Run `helm create ...` to create a fresh new Helm chart.

2. Copy the following resource definitions under `templates/`.
```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: my-sa
  annotations:
    helm.sh/hook: pre-install
    helm.sh/hook-weight: "-3"
    helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded,hook-failed
---
kind: Job
apiVersion: batch/v1
metadata:
  name: my-job
  annotations:
    helm.sh/hook: pre-install
    helm.sh/hook-weight: "-3"
    helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded,hook-failed
spec:
  template:
    spec:
      serviceAccountName: my-sa
      containers:
      - name: my-pod
        image: mcr.microsoft.com/cbl-mariner/base/python:3.9
        command: ["/bin/bash", "-c", "sleep 600"]
      restartPolicy: Never
```
Note that both the `ServiceAccount` and `Job` have the same hook weight of `-3`.

3. `kubectl get serviceaccounts` and `kubectl get jobs -n $namespace -w` to watch the resource creation.

4. Install the chart: `helm install -n $namespace ...`.

The output of the watched resources shows that the `Job` is created but no `ServiceAccount` is created.
image

The `Job` also fails to create `Pods`.
image

---
If I weight the `ServiceAccount` lower so it gets prioritized, it will get created before the `Job`, so the hook itself seems to work as expected.
image

image

image

---

Output of `helm version`: `version.BuildInfo{Version:"v3.12.0", GitCommit:"c9f554d75773799f72ceef38c51210f1842a1dea", GitTreeState:"clean", GoVersion:"go1.20.3"}`

Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"27", GitVersion:"v1.27.2", GitCommit:"7f6f68fdabc4df88cfea2dcf9a19b2b830f1e647", GitTreeState:"clean", BuildDate:"2023-05-17T14:20:07Z", GoVersion:"go1.20.4", Compiler:"gc", Platform:"linux/amd64"}
Kustomize Version: v5.0.1
Server Version: version.Info{Major:"1", Minor:"26", GitVersion:"v1.26.3", GitCommit:"f18584a06fc476806da2e340e2eed960659871e8", GitTreeState:"clean", BuildDate:"2023-06-12T18:45:20Z", GoVersion:"go1.19.7", Compiler:"gc", Platform:"linux/amd64"}
```

lynshi:
**What this PR does / why we need it**: closes #12414

According to [Hooks and the Release Lifecycle](https://helm.sh/docs/topics/charts_hooks/#hooks-and-the-release-lifecycle), hooks are sorted "by weight (assigning a weight of 0 by default), by resource kind and finally by name in ascending order". In practice, the current behavior is that hooks are sorted by `(Weight, Name)`.

The code currently uses `sort.Stable` under the assumption that the hooks are already sorted by Kind. However, stability only applies to elements with the same key, so since `Less` on the `hookByWeight` comparator orders based on the name of the hook too, the existing ordering by Kind is never respected.

This PR fixes the comparator so that hooks are sorted by `(Weight, Kind, Name)` as expected and adds tests for this behavior.

**If applicable**:

### DIFF
--- a/pkg/action/hooks.go
+++ b/pkg/action/hooks.go
@@ -125,15 +125,13 @@ func (x hookByWeight) Less(i, j int) bool {
 		first, iok := ordering[x[i].Kind]
 		second, jok := ordering[x[j].Kind]
 
-		// As in https://github.com/helm/helm/blob/fe595b69d78b213ab181d98ce24dde2454a56f9d/pkg/releaseutil/kind_sorter.go#L145C15-L145C15
 		if !iok && !jok {
-			// If both are unknown then sort alphabetically by kind.
-			if x[i].Kind != x[j].Kind {
-				return x[i].Kind < x[j].Kind
+			// Sort unknown kinds alphabetically, with name as the tiebreaker.
+			if x[i].Kind == x[j].Kind {
+				return x[i].Name < x[j].Name
 			}
 
-			// Otherwise, let Stable() preserve the original order.
-			return false
+			return x[i].Kind < x[j].Kind
 		}
 
 		// Unknown kind is last.
--- a/pkg/action/hooks.go
+++ b/pkg/action/hooks.go
@@ -115,13 +115,6 @@ func (x hookByWeight) Len() int      { return len(x) }
 func (x hookByWeight) Swap(i, j int) { x[i], x[j] = x[j], x[i] }
 func (x hookByWeight) Less(i, j int) bool {
 	if x[i].Weight == x[j].Weight {
-		return x[i].Name < x[j].Name
-	}
-	return x[i].Weight < x[j].Weight
-}
-
-func (x hookByWeight) LessFixed(i, j int) bool {
-	if x[i].Weight == x[j].Weight {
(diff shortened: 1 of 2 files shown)


lynshi:
Only one of this and https://github.com/helm/helm-www/pull/1520 should be merged.

PR #12502, opened 2023-10-18
fix: --disable-openapi-validation option was never applied

### LINKED ISSUE #12470: What does --disable-openapi-validation do exactly?
Hello,
I try to understand the exact role of the option `--disable-openapi-validation`.
I created a simple chart with just this pod resource:
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: testdenouche
spec:
  containers:
  - name: nginx
    image: nginx:1.14.2
    ports:
    - containerPort: 80
```
Then I tested the following modifications:
- add a spec.toto = tutu 
- add a metadata.labels.tutu = test@denouche
- change the data format of spec.containers from an array to a map (just by removing the - before name: nginx)

For each of these changes I tested the helm ugrade with and without the option --disable-openapi-validation and the behavior is exactly the same each time : I have an error and a new secret is created, and I call see the failed release in `helm history ...` output list.

The only things I found on this option is the doc `if set, the upgrade process will not validate rendered templates against the Kubernetes OpenAPI Schema` and some issues and PR when this option have been introduced #6819 of @bacongobbler but I cannot find a case where I can deploy something using the option --disable-openapi-validation which is not allowed without this flag.

Thanks for your help.

Output of `helm version`:

Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"26", GitVersion:"v1.26.5", GitCommit:"890a139214b4de1f01543d15003b5bda71aae9c7", GitTreeState:"clean", BuildDate:"2023-05-17T14:14:46Z", GoVersion:"go1.19.9", Compiler:"gc", Platform:"linux/amd64"}
Kustomize Version: v4.5.7
Server Version: version.Info{Major:"1", Minor:"26", GitVersion:"v1.26.7-gke.500", GitCommit:"178d88870d1ad59817801635c1da4b49cb46a016", GitTreeState:"clean", BuildDate:"2023-07-26T09:33:20Z", GoVersion:"go1.20.6 X:boringcrypto", Compiler:"gc", Platform:"linux/amd64"}
```
GKE

denouche:
**Why we need it**:

Currently there is an issue with openapi validation in Helm.
Here is the problem description.
Let's take a pod definition with a problem, the spec.toto field does not exist:
```yaml
apiVersion: v1
kind: Pod
metadata:
  namespace: mynamespace
  name: mypod
spec:
  toto: tutu
  containers:
  - name: nginx
    image: nginx:1.14.2
    ports:
    - containerPort: 80
```

First I try to create it using kubectl, with no specific option:
```bash
$ kubectl apply -f pod.yaml 
Error from server (BadRequest): error when creating "pod.yaml": Pod in version "v1" cannot be handled as a Pod: strict decoding error: unknown field "spec.toto"
```
Then I try different options of --validate:
```bash
$ kubectl apply -f pod.yaml --validate=strict
Error from server (BadRequest): error when creating "pod.yaml": Pod in version "v1" cannot be handled as a Pod: strict decoding error: unknown field "spec.toto"

$ kubectl apply -f pod.yaml --validate=ignore
pod/mypod created
```

This is okay and this is the kubectl behavior.

Now the same thing with Helm.
First test with no option:
```bash
$ helm install testchart .
W1018 10:43:09.834267  122282 warnings.go:70] unknown field "spec.toto"
NAME: testchart
LAST DEPLOYED: Wed Oct 18 10:43:09 2023
NAMESPACE: mynamespace
STATUS: deployed
REVISION: 1
TEST SUITE: None
```
Here I just have a warning but the resource is created. This is not normal.

Then with different values of the option:
```bash
$ helm upgrade testchart . --disable-openapi-validation
W1018 10:44:48.986008  122562 warnings.go:70] unknown field "spec.toto"
Release "testchart" has been upgraded. Happy Helming!
NAME: testchart
LAST DEPLOYED: Wed Oct 18 10:44:48 2023
NAMESPACE: mynamespace
STATUS: deployed
REVISION: 2
TEST SUITE: None
```
and:
```bash
$ helm upgrade testchart . --disable-openapi-validation=false
W1018 10:45:05.128872  122606 warnings.go:70] unknown field "spec.toto"
Release "testchart" has been upgraded. Happy Helming!
NAME: testchart
LAST DEPLOYED: Wed Oct 18 10:45:04 2023
NAMESPACE: mynamespace
STATUS: deployed
REVISION: 3
TEST SUITE: None
```

So currently the option `--disable-openapi-validation` of Helm does not change anything.

The source of the problem is:
Currently we validate the manifest in the prepareUpgrade here:
https://github.com/helm/helm/blob/main/pkg/action/upgrade.go#L290
Then the validation direct is calculated here:
https://github.com/helm/helm/blob/main/pkg/kube/client.go#L344-L347
and then given to the kube client.
And in the kube client the validation apply client-side only if the ServerSideFieldValidation is disabled on the api-server, as described here: https://kubernetes.io/blog/2023/04/24/openapi-v3-field-validation-ga/#server-side-field-validation
```kubectl will skip client side validation and will automatically use server side field validation in Strict mode```
But in the `prepareUpgrade` step, no creation is done, so no server-side validation is applied neither client-side validation.

Then after the `prepareUpgrade` step we go to the `performUpgrade` step and here we really do the Update, here:
https://github.com/helm/helm/blob/main/pkg/action/upgrade.go#L408
if we take a look deeper into the Update func of our client, the real update is done here:
https://github.com/helm/helm/blob/main/pkg/kube/client.go#L424
and here is the way we instanciate the Helper that will be used for the patch:
https://github.com/helm/helm/blob/main/pkg/kube/client.go#L665

If we compare this to the way kubectl instanciate the Helper, there is a big difference:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kubectl/pkg/cmd/apply/apply.go#L559-L562
In kubectl there are adding the line:
```go
WithFieldValidation(o.ValidationDirective)
```
and this line configure the client to add the validate field with strict or ignore value.
So in our case we never give the validationDirective value to the API when we do the update, and so the defaut value applies:
https://kubernetes.io/blog/2023/04/24/openapi-v3-field-validation-ga/#server-side-field-validation
```
If the parameter is not passed, server side field validation is in Warn mode by default.
```
That's why in the different tests you always see a Warning but the resources are always created, even when there is an error in the manifest.

**What this PR does**:

So my PR forward the validation boolean in client and add the validationDirective in the the create/update.
I also chose to set the default value to "Warn" instead of "Strict", because it's the current behavior and I didn't want to break it.

Let me know what you think about all this! 

**Result:**

After my changes I have the following behavior with differents values of the --disable-openapi-validation parameter:
when enabling the validation I still have the same warning as previously:
```bash
$ /home/denouche/workspace/src/github.com/denouche/helm/bin/helm upgrade testchart . --disable-openapi-validation=false
W1018 11:15:15.140002  126116 warnings.go:70] unknown field "spec.toto"
Release "testchart" has been upgraded. Happy Helming!
NAME: testchart
LAST DEPLOYED: Wed Oct 18 11:15:14 2023
NAMESPACE: mynamespace
STATUS: deployed
REVISION: 2
TEST SUITE: None
```

and when disabling, no warning anymore:
```bash
$ /home/denouche/workspace/src/github.com/denouche/helm/bin/helm upgrade testchart . --disable-openapi-validation=true
Release "testchart" has been upgraded. Happy Helming!
NAME: testchart
LAST DEPLOYED: Wed Oct 18 11:15:20 2023
NAMESPACE: mynamespace
STATUS: deployed
REVISION: 3
TEST SUITE: None
```

Closes #12470

### DIFF
--- a/pkg/action/hooks.go
+++ b/pkg/action/hooks.go
@@ -73,7 +73,7 @@ func (cfg *Configuration) execHook(rl *release.Release, hook release.HookEvent, 
 		h.LastRun.Phase = release.HookPhaseUnknown
 
 		// Create hook resources
-		if _, err := cfg.KubeClient.Create(resources); err != nil {
+		if _, err := cfg.KubeClient.Create(resources, false); err != nil {
 			h.LastRun.CompletedAt = helmtime.Now()
 			h.LastRun.Phase = release.HookPhaseFailed
 			return errors.Wrapf(err, "warning: Hook %s %s failed", hook, h.Path)
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -161,7 +161,7 @@ func (i *Install) installCRDs(crds []chart.CRD) error {
 		}
 
 		// Send them to Kube
-		if _, err := i.cfg.KubeClient.Create(res); err != nil {
+		if _, err := i.cfg.KubeClient.Create(res, !i.DisableOpenAPIValidation); err != nil {
 			// If the error is CRD already exists, continue.
 			if apierrors.IsAlreadyExists(err) {
 				crdName := res[0].Name
@@ -368,7 +368,7 @@ func (i *Install) RunWithContext(ctx context.Context, chrt *chart.Chart, vals ma
 		if err != nil {
 			return nil, err
 		}
-		if _, err := i.cfg.KubeClient.Create(resourceList); err != nil && !apierrors.IsAlreadyExists(err) {
+		if _, err := i.cfg.KubeClient.Create(resourceList, !i.DisableOpenAPIValidation); err != nil && !apierrors.IsAlreadyExists(err) {
 			return nil, err
 		}
 	}
@@ -437,9 +437,9 @@ func (i *Install) performInstall(rel *release.Release, toBeAdopted kube.Resource
 	// do an update, but it's not clear whether we WANT to do an update if the re-use is set
 	// to true, since that is basically an upgrade operation.
 	if len(toBeAdopted) == 0 && len(resources) > 0 {
-		_, err = i.cfg.KubeClient.Create(resources)
+		_, err = i.cfg.KubeClient.Create(resources, !i.DisableOpenAPIValidation)
 	} else if len(resources) > 0 {
-		_, err = i.cfg.KubeClient.Update(toBeAdopted, resources, i.Force)
+		_, err = i.cfg.KubeClient.Update(toBeAdopted, resources, i.Force, !i.DisableOpenAPIValidation)
 	}
 	if err != nil {
 		return rel, err
--- a/pkg/action/rollback.go
+++ b/pkg/action/rollback.go
@@ -187,7 +187,7 @@ func (r *Rollback) performRollback(currentRelease, targetRelease *release.Releas
 	if err != nil {
 		return targetRelease, errors.Wrap(err, "unable to set metadata visitor from target release")
 	}
-	results, err := r.cfg.KubeClient.Update(current, target, r.Force)
+	results, err := r.cfg.KubeClient.Update(current, target, r.Force, false)
 
 	if err != nil {
 		msg := fmt.Sprintf("Rollback %q failed: %s", targetRelease.Name, err)
--- a/pkg/action/upgrade.go
+++ b/pkg/action/upgrade.go
@@ -405,7 +405,7 @@ func (u *Upgrade) releasingUpgrade(c chan<- resultMessage, upgradedRelease *rele
 		u.cfg.Log("upgrade hooks disabled for %s", upgradedRelease.Name)
 	}
 
-	results, err := u.cfg.KubeClient.Update(current, target, u.Force)
+	results, err := u.cfg.KubeClient.Update(current, target, u.Force, !u.DisableOpenAPIValidation)
 	if err != nil {
 		u.cfg.recordRelease(originalRelease)
 		u.reportToPerformUpgrade(c, upgradedRelease, results.Created, err)
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -138,9 +138,9 @@ func (c *Client) IsReachable() error {
 }
 
 // Create creates Kubernetes resources specified in the resource list.
-func (c *Client) Create(resources ResourceList) (*Result, error) {
+func (c *Client) Create(resources ResourceList, validate bool) (*Result, error) {
 	c.Log("creating %d resource(s)", len(resources))
-	if err := perform(resources, createResource); err != nil {
+	if err := perform(resources, getCreateResource(getValidationDirective(validate))); err != nil {
 		return nil, err
 	}
 	return &Result{Created: resources}, nil
@@ -386,9 +386,11 @@ func (c *Client) BuildTable(reader io.Reader, validate bool) (ResourceList, erro
 // occurs, a Result will still be returned with the error, containing all
 // resource updates, creations, and deletions that were attempted. These can be
 // used for cleanup or other logging purposes.
-func (c *Client) Update(original, target ResourceList, force bool) (*Result, error) {
+func (c *Client) Update(original, target ResourceList, force bool, validate bool) (*Result, error) {
 	updateErrors := []string{}
 	res := &Result{}
+
+	validationDirective := getValidationDirective(validate)
 
 	c.Log("checking %d resources for changes", len(target))
 	err := target.Visit(func(info *resource.Info, err error) error {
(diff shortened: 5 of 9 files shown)


tanvp112:
Hi, has this being released already?

MissedTheMark (approved):
This change looks good, and I ran the relevant commands locally and got the expected responses (see below). 

My one concern is that it's a breaking change to the signatures of a couple of functions in `pkg/kube/client.go` - but I'll leave it up to the maintainers to determine if that's acceptable or not.  

Local testing setup was done by running `helm create simple-test`, and adding the following `pod.yaml` template, as per the PR description
``` yaml
apiVersion: v1
kind: Pod
metadata:
  namespace: default
  name: mypod
spec:
  toto: tutu
  containers:
  - name: nginx
    image: nginx:1.14.2
    ports:
    - containerPort: 80
```

### Before this change
**With validation**
``` console
$ helm install simple-test simple-test
W0514 15:32:24.231150   12049 warnings.go:70] unknown field "spec.toto"
NAME: simple-test
LAST DEPLOYED: Tue May 14 15:32:24 2024
[etc]
```

**With validation disabled**
Note that a validation warning is output:
``` console
$ helm install simple-test simple-test --disable-openapi-validation
W0514 15:32:46.040075   12359 warnings.go:70] unknown field "spec.toto"
NAME: simple-test
LAST DEPLOYED: Tue May 14 15:32:45 2024
[etc]
```

### After this change
**With validation**
Note that a validation warning is output as expected:
``` console
$ helm install simple-test simple-test
W0514 15:32:24.231150   12049 warnings.go:70] unknown field "spec.toto"
NAME: simple-test
LAST DEPLOYED: Tue May 14 15:32:24 2024
[etc]
```

**With validation disabled**
Note that no validation warning is output:
``` console
$ helm install simple-test simple-test --disable-openapi-validation
NAME: simple-test
LAST DEPLOYED: Fri May 17 12:04:47 2024
[etc]
```

I have also tested `upgrade` and `rollback` commands which work as expected.

mjnagel:
I think this does solve one half of the issue (actually allowing for disabling schema validation), but I'm not sure that openapi-validation is behaving as expected assuming `kubectl apply --validate='ignore'|'strict'` is the target equivalent. A normal `kubectl apply` (strict validation being the default) would "be blocked" on schema validation failure, but the same is not true with `helm`. With the changes here schema validation is skipped as expected when the flag is present, but an install will still complete regardless of the flag.

Happy to converse further on the issue if that was not the intent of this PR, or open a separate issue - just came across this today and noticed https://github.com/helm/helm/issues/12470#issuecomment-1753794744 referring to the `kubectl` parallel.

pkg/action/hooks.go
```diff
@@ -73,7 +73,7 @@ func (cfg *Configuration) execHook(rl *release.Release, hook release.HookEvent,
 		h.LastRun.Phase = release.HookPhaseUnknown
 
 		// Create hook resources
-		if _, err := cfg.KubeClient.Create(resources); err != nil {
+		if _, err := cfg.KubeClient.Create(resources, false); err != nil {
```
gjenkins8, line 76: It seems like (for consistency) the operations `DisableOpenAPIValidation` value should be passed in here.

Otherwise, doing e.g. `helm install --disable-openapi-validation` could still fail schema validation for a malformed hook

(that said, this PR is fixing a regression. it doesn't need to be perfect to be an improvement)

chiragjn:
Can we please have this fix merged ?
This blocks argocd from using this
which in creates a lot of pain for charts that are trying to use helm hooks to upgrade crds

benoittgt:
@denouche Do you want to tackle this and the comment? Otherwise, I can take care of it today.

denouche:
> @denouche Do you want to tackle this and the comment? Otherwise, I can take care of it today.

Hello,
Yes if you can take care of it, it would be great. 
I have no time for this these days, sorry.

benoittgt:
There is is little more work with the new Server Side Apply introduced in 4 but I will try to handle this case too. :)

benoittgt:
Ok, I dug deeper into the PR but I see little interest of doing it since [we'll now default to Server-Side Apply in Helm V4](https://github.com/helm/community/blob/main/hips/hip-0023.md). Or maybe core-maintainers want to support `--disable-openapi-validation=true` for Client-Side Apply (cc @gjenkins8) ?

From the HIP
> Client-side API object validation isn't applicable when SSA is enabled (helm install|upgrade --disable-openapi-validation)

For the current behavior on k8s:

> If you submit a request that specifies an unrecognized field, and that is also invalid for a different reason (for example, the request provides a string value where the API expects an integer for a known field), then the API server responds with a 400 Bad Request error, but will not provide any information on unknown or duplicate fields (only which fatal error it encountered first).
>
> You always receive an error response in this case, no matter what field validation level you requested. [[source](https://kubernetes.io/docs/reference/using-api/api-concepts/#setting-the-field-validation-level)]

Regarding the [memory issue](https://github.com/helm/helm/issues/12470#issuecomment-1755422967) from OP,it was raised two years ago. Is it still a problem?

From what I gathered in the dev meeting, we seem to be moving toward stricter validation instead.

PR #12518, opened 2023-10-23
Recreate immutable resources on upgrade

### LINKED ISSUE #7082: proposal: handle immutable resources on update

Vegoo89:
Closes #7082 

**What this PR does / why we need it**: This PR tries to solve the recreation resources problem in minimalistic way by adding new `--recreate-immutable-resources` argument to command line for `helm upgrade` only.
It should be mentioned that I read the whole discussion about the issues and I understand why maintainers are not eager to merge it, but I thought I will give it a try.
In my solution I would like to mark this feature as experimental (which I was not sure how to do) or just explicitly mention that if it would be used, bad things may happen (like example with deleting PVC).
However - as a helm user - I would like to use it myself in non-critical workflows where for example I change labels for my deployment.

**Special notes for your reviewer**: I am pretty sure reviewers are familiar with this problem. I didn't want to introduce the breaking change and I thought additional function parameter in `Update` is not breaking, but I may be wrong. If you think MR is worth to work on, I am happy to spend some more time on it, however if you think merging this would be very difficult or this is not really the feature you are looking for, information would also be appreciated.

**If applicable**:

Signed-off-by: Jakub Domanski <vegetto_sys@yahoo.com>

### DIFF
--- a/cmd/helm/rollback.go
+++ b/cmd/helm/rollback.go
@@ -85,6 +85,7 @@ func newRollbackCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 	f.BoolVar(&client.WaitForJobs, "wait-for-jobs", false, "if set and --wait enabled, will wait until all Jobs have been completed before marking the release as successful. It will wait for as long as --timeout")
 	f.BoolVar(&client.CleanupOnFail, "cleanup-on-fail", false, "allow deletion of new resources created in this rollback when rollback fails")
 	f.IntVar(&client.MaxHistory, "history-max", settings.MaxHistory, "limit the maximum number of revisions saved per release. Use 0 for no limit")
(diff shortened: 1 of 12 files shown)


joejulian:
What happens if an upgrade fails but `--atomic` is set?

Vegoo89:
> What happens if an upgrade fails but `--atomic` is set?

Added handling for this. Totally missed it as I didn't really want to add this functionality neither to `install`, nor to `rollback` to keep the MR simple.
With pushed changes `--atomic` should reverse the release (so delete existing resource and create old one).

dastrobu:
remove stale

PR #12519, opened 2023-10-23
fix version retrieval of get-helm-3

MarHum26:
add quiet switch to prevent output  like progress information in downloaded data

I wasn't able to install helm using the script when I had only wget installed. The reason was that exactly on the line with the version information from the download some output from wget was inserted.

The documentation of wget states the danger of using -O to write information because it just behaves like a shell redirection.
Using -q fixed the problem for me because no output from wget occurs and therefor the version retrieval was working.

None

**If applicable**:


joejulian (approved):
That makes sense to me. If curl is silent, wget should be too.

mattfarina (approved):

Issue #12491, opened 2023-10-13
unexpected status from HEAD request on push (3.13.0 & 3.13.1)


vmercierfr:
Output of `helm version`:

```
version.BuildInfo{Version:"v3.13.1", GitCommit:"3547a4b5bf5edb5478ce352e18858d8a552a4110", GitTreeState:"clean", GoVersion:"go1.20.8"}
````

Output of `kubectl version`: n/a

Cloud Provider/Platform (AKS, GKE, Minikube etc.): AWS public ECR

---

Since Helm 3.13.0 (including 3.13.1), we receive HTTP 400 when pushing the Helm chart on the OCI AWS ECR public repository that prevent us to release new Helm chart version.

The HTTP 400 error is not reported in AWS Cloudtrail, which suggests that the query was rejected early.

Full logs:

```
helm registry logout public.ecr.aws # optional cleanup

aws ecr-public get-login-password --region us-east-1 |helm registry login --username AWS --password-stdin public.ecr.aws

helm package configs/helm --app-version 0.2.3-rc3 --version 0.2.3-rc3

helm push --debug <our-repository>-0.2.3-rc2.tgz oci://public.ecr.aws/<repository>

WARN[0000] reference for unknown type: application/vnd.cncf.helm.chart.content.v1.tar+gzip 
WARN[0000] reference for unknown type: application/vnd.cncf.helm.config.v1+json 
DEBU[0000] checking and pushing to                       url="https://public.ecr.aws/v2/<our-organization>/<our-repository>/blobs/sha256:0d2ec2c0f851156ef2efe939fa35ed5cb739e868525818e20bafebfeee119368"
DEBU[0000] checking and pushing to                       url="https://public.ecr.aws/v2/<our-organization>/<our-repository>/blobs/sha256:7af1085958bd4086125c09f442d0c98c004f880794ed0fd3a7f508d4e45eadc0"
DEBU[0000] do request                                    request.header.accept="application/vnd.cncf.helm.config.v1+json, */*" request.header.user-agent=Helm/3.13.1 request.method=HEAD url="https://public.ecr.aws/v2/<our-organization>/<our-repository>/blobs/sha256:7af1085958bd4086125c09f442d0c98c004f880794ed0fd3a7f508d4e45eadc0"
[...]

PR #12524, opened 2023-10-25
Revert "fix(registry): address anonymous pull issue"

mattfarina:
This is one of 2 PRs that's causing a regression in #12491. Both need to be reverted and this is the first one.

Reverts helm/helm#12424


sabre1041 (approved):
LGTM

PR #12543, opened 2023-11-05
Publish the release to Winget on Git tag creation

ste93cry:

Now that `Helm` is available on [`Winget`](https://github.com/microsoft/winget-pkgs/tree/master/manifests/h/Helm/Helm), each release should be published regularly on it. Until now, some people volunteered to keep the manifests up-to-date, but it would be better to have an automated process in place so that we can be sure that no release is left behind.

The `WINGET_TOKEN` must be made available in the workflow with enough permissions to be able to fork the `microsoft/winget-pkgs` repository and open a PR to it.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -51,6 +51,20 @@ jobs:
           connection_string: ${{ secrets.AZURE_STORAGE_CONNECTION_STRING }}
           extra_args: '--pattern helm-*'
 
+      - name: Setup Komac
+        run: |
+          curl -fSsL -o komac.jar https://github.com/russellbanks/Komac/releases/download/v1.11.0/Komac-1.11.0-all.jar
+          echo '1c26bb7bee6228ad095fe67d10ed254363099b3395f809bd6d95a3f72b2740fc komac.jar' | sha256sum --check --status -
+
+      - name: Publish to Winget
+        run: |
+          $JAVA_HOME_17_X64/bin/java -jar komac.jar update \
+            --id Helm.Helm \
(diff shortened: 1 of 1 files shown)


mattfarina (commented):
Thanks for your interest in getting this published over there. But, the Helm project has not been the one to maintain and push to these types of locations. For example, the Helm project does not do this work for Homebrew, which is used on Mac and Linux. There are other examples, this is not the only one.

It might be better to setup automation that discovers new Helm releases and then creates the pull request separate from the Helm release process.

ste93cry:
I understand, although I don't completely agree with the fact that the Helm project shouldn't be responsible for publishing releases to package managers. One of the reasons this makes sense to me is that it would be a strong signal that the installation method is officially supported. Nonetheless, I understand that there may not be sufficient resources to handle this.

Anyway, since the documentation mentions the other package managers like Scoop and Homebrew, so I wonder whether at least we can add `Winget` to the list. I prepared a [PR](https://github.com/helm/helm-www/pull/1514) some time ago for this exact purpose.

mattfarina:
@ste93cry Thanks for the request. At the moment we don't have any maintainers who have expressed interest in maintaining this. The Helm project has never maintained all of the builds and packages providing Helm. Homebrew on Mac is a common location for people to install Helm and it is not maintained by the Helm project.

You can find more on my personal view at https://codeengineered.com/blog/2024/open-source-not-builds/

Since there are no maintainers interested in managing this, I'm going to close the issue.

If someone does take this on we are open to documenting the location alongside others who build Helm.

Issue #10481, opened 2021-12-17
Relative filesystem schema references in `values.schema.json` broken


jedcunningham:
Output of `helm version`:
```
version.BuildInfo{Version:"v3.7.2", GitCommit:"663a896f4a815053445eec4153677ddc24a0a361", GitTreeState:"clean", GoVersion:"go1.17.3"}
```

## Issue

When using a relative filesystem reference in `values.schema.json`, it looks for the file in your CWD, not in the chart directory/tarball.

In practice, this means one cannot use relative references in charts when they may want to (e.g. so you don't need outbound routing to an external/remote schema).

## How to reproduce

Create a new chart:

```
$ helm create demo
Creating demo
$ cd demo
```

Create `values.schema.json`:

```yaml
{
    "$schema": "http://json-schema.org/schema#",
    "properties": {
        "replicaCount": {
            "$ref": "file://./another_schema.json"
        }
    },
    "title": "Values",
    "type": "object"
}
```

And create `another_schema.json`:

```yaml
{
    "$schema": "http://json-schema.org/schema#",
    "type": "integer"
}
```

Now, from `demo`, lint the chart:

```
$ helm lint .
==> Linting .
[INFO] Chart.yaml: icon is recommended

1 chart(s) linted, 0 chart(s) failed
```

Now, `cd` up a directory and lint again:

```
$ cd ..
$ helm lint demo
==> Linting demo
[INFO] Chart.yaml: icon is recommended
[ERROR] values.yaml: open ./another_schema.json: no such file or directory
[ERROR] templates/: values don't meet the specifications of the schema(s) in the following chart(s):
demo:
open ./another_schema.json: no such file or directory

Error: 1 chart(s) linted, 1 chart(s) failed
```

The same behavior happens with a tarball:

```
$ cd demo
$ helm package .
Successfully packaged chart and saved it to: ...demo/demo-0.1.0.tgz
$ helm lint demo-0.1.0.tgz 
==> Linting demo-0.1.0.tgz
[INFO] Chart.yaml: icon is recommended

1 chart(s) linted, 0 chart(s) failed
$ cd ..
$ helm lint demo/demo-0.1.0.tgz 
==> Linting demo/demo-0.1.0.tgz
[INFO] Chart.yaml: icon is recommended
[ERROR] values.yaml: open ./another_schema.json: no such file or directory
[ERROR] templates/: values don't meet the specifications of the schema(s) in the following chart(s):
demo:
open ./another_schema.json: no such file or directory

[...]

PR #12553, opened 2023-11-08
Support loading subschemas from the schemas directory in a chart

bossmc:

Related to #10481.

It's often nice to be able to write JSON schemas as separate files and to reference them with the `$ref` key into the root schema (this is handy when the subschemas are not your own, or when other tools wish to integrate with part of your data model, or if you have a large schema with duplication and want to compartmentalize the various sections of configuration.

JSON schema allows `$ref` markers to point to a "canonical location" for a subschema, with the intention that the schema validator will go and fetch the subschema during compilation of the schema for a validation operation.  This is "fine" (with the usual caveats) if the canonical location is an `http(s)://` URL since the validator can go and fetch that URL when it needs to, but is less fine for a `file://` URL since that file might not be available if the Chart is published and subsequently fetched onto another machine.  Fortunately the JSONschema module that helm uses allows "pre-loading" of subschemas which can then be used to satisfy `$ref` markers without needing to fetch from the canonical location (the `$id` element of the subschema names the canonical location for cross-corelation).

Using the above, this MR allows chart authors to place subschemas into the `schemas` folder of their chart, which will be pre-loaded when processing the `values.schema.json` file.  This means that local subschemas can be named by the Chart and still be accessible when the chart is use on a second device (since the `schemas` folder will be packaged with the chart).

Schemas can still use `http(s)` references as before.

I've not written any documentation for this change yet, since I'm not sure having a special `schemas` folder is the right model, I considered having a list in the Chart.toml of extra schemas to load, or loading all files called `*.schema.json` in the Chart root but thought a folder would be easiest for the user.

**If applicable**:

### DIFF
--- a/pkg/chartutil/testdata/test-values-with-extra.schema.json
+++ b/pkg/chartutil/testdata/test-values-with-extra.schema.json
@@ -1,0 +1,52 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "properties": {
+    "addresses": {
+      "$ref": "https://helm.sh/schemas/addresses"
+    },
+    "age": {
+      "description": "Age",
+      "minimum": 0,
+      "type": "integer"
+    },
+    "employmentInfo": {
+      "properties": {
+        "salary": {
+          "minimum": 0,
+          "type": "number"
+        },
+        "title": {
+          "type": "string"
+        }
+      },
+      "required": [
+        "salary"
+      ],
+      "type": "object"
+    },
+    "firstname": {
+      "description": "First name",
+      "type": "string"
+    },
+    "lastname": {
+      "type": "string"
+    },
+    "likesCoffee": {
+      "type": "boolean"
+    },
+    "phoneNumbers": {
+      "items": {
+        "type": "string"
+      },
+      "type": "array"
+    }
+  },
+  "required": [
+    "firstname",
+    "lastname",
+    "addresses",
+    "employmentInfo"
+  ],
+  "title": "Values",
+  "type": "object"
+}
--- a/pkg/lint/rules/values.go
+++ b/pkg/lint/rules/values.go
@@ -18,6 +18,7 @@ package rules
 
 import (
 	"os"
+	"io/fs"
 	"path/filepath"
 
 	"github.com/pkg/errors"
@@ -76,11 +77,46 @@ func validateValuesFile(valuesPath string, overrides map[string]interface{}) err
 	ext := filepath.Ext(valuesPath)
 	schemaPath := valuesPath[:len(valuesPath)-len(ext)] + ".schema.json"
 	schema, err := os.ReadFile(schemaPath)
+
+	// Check for zero-length _before_ checking for errors, since we want schema files to be optional
 	if len(schema) == 0 {
 		return nil
 	}
 	if err != nil {
 		return err
 	}
-	return chartutil.ValidateAgainstSingleSchema(coalescedValues, schema)
+
+	extraSchemas := make([][]byte, 0)
+	schemasPath := filepath.Dir(valuesPath) + "/schemas"
+	err = filepath.WalkDir(schemasPath, func(path string, dentry fs.DirEntry, err error) error {
+		if path == schemasPath && dentry == nil && err != nil {
+			// The "schemas" folder could not be opened if it doesn't exist, treat it as empty
+			if os.IsNotExist(err) {
+				return fs.SkipAll
+			} else {
+				return err
+			}
+		}
+
+		if err != nil {
+			return err
+		}
+
+		if !dentry.IsDir() {
+			schema, err = os.ReadFile(path)
+			if err != nil {
+				return err
+			}
+
+			extraSchemas = append(extraSchemas, schema)
+		}
+
+		return nil
+	})
+
+	if err != nil {
+		return err
+	}
+
+	return chartutil.ValidateAgainstSingleSchema(coalescedValues, schema, extraSchemas)
 }
--- a/pkg/lint/rules/values.go
+++ b/pkg/lint/rules/values.go
@@ -103,12 +103,12 @@ func validateValuesFile(valuesPath string, overrides map[string]interface{}) err
 		}
 
 		if !dentry.IsDir() {
-			schema, err = os.ReadFile(path)
+			extraSchema, err := os.ReadFile(path)
 			if err != nil {
 				return err
 			}
 
-			extraSchemas = append(extraSchemas, schema)
+			extraSchemas = append(extraSchemas, extraSchema)
 		}
 
 		return nil
--- a/pkg/chart/chart.go
+++ b/pkg/chart/chart.go
@@ -46,8 +46,10 @@ type Chart struct {
 	Templates []*File `json:"templates"`
 	// Values are default config for this chart.
 	Values map[string]interface{} `json:"values"`
-	// Schema is an optional JSON schema for imposing structure on Values
+	// Schema is an optional JSON schema imposing structure on the chart values
 	Schema []byte `json:"schema"`
+	// Additional schemas are optional JSON schemas that can be referenced from the root schema
+	ExtraSchemas [][]byte `json:"extra_schemas"`
 	// Files are miscellaneous files in a chart archive,
 	// e.g. README, LICENSE, etc.
 	Files []*File `json:"files"`
--- a/pkg/chart/loader/load.go
+++ b/pkg/chart/loader/load.go
@@ -137,7 +137,8 @@ func LoadFiles(files []*BufferedFile) (*chart.Chart, error) {
 			if c.Metadata.APIVersion == chart.APIVersionV1 {
 				c.Files = append(c.Files, &chart.File{Name: f.Name, Data: f.Data})
 			}
-
+		case strings.HasPrefix(f.Name, "schemas/"):
+			c.ExtraSchemas = append(c.ExtraSchemas, f.Data)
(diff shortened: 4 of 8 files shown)


Ornias1993:
This is yet another great PR that needlessly does not get attention by the maintainers here.
This is not acceptable from a CNCF project.

pkg/chartutil/testdata/test-values-with-extra.schema.json
```diff
@@ -0,0 +1,52 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "properties": {
+    "addresses": {
+      "$ref": "https://helm.sh/schemas/addresses"
```
martincomputershare, line 5: Does the change add support for referencing JSON Schemas from local files as well? For instance:

```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "properties": {
    "addresses": {
      "$ref": "file://./test-values-extra.schema.json"
```

martincomputershare:
Huge chagrin to see that Helm's been lacking this feature for so long. Big hopes for the PR!

td-macompta:
Great hope that this PR will succeed.

pkg/chartutil/testdata/test-values-with-extra.schema.json
martincomputershare: What about `$ref`-loading a schema that sits in a parent directory? For example: using the schema of a chart that sits in the same repo?

```
🐙 repo
  📁 charts
    📁 apps
  ┌─► 📦 app-1
  │     📜 Chart.yaml
  │     📄 schema.json ◄──────────────────────────────────────────────┐
  ┌─► 📦 app-2                                                        │
  │     📜 Chart.yaml                                                 │
  │     📄 schema.json ◄──────────────────────────────────────────────┐
  │   📦 app-3                                                        │
  │     📜 Chart.yaml                                                 │
  │       > dependencies:                                             │
  │       >   - name: app-1                                           │
  │       >     version: 0.0.0                                        │
  └────── >     repository: file://../app-1                           │
  │       >   - name: app-2                                           │
  │       >     version: 0.0.0                                        │
  └────── >     repository: file://../app-2                           │
        📄 schema.json                                                │
          > {                                                         │
          >   "type": "object",                                       │
          >   "properties": {                                         │
          >     "app-1": { "$ref": "file://../app-1/schema.jsson" }, ─┘
          >     "app-2": { "$ref": "file://../app-2/schema.jsson" }  ─┘
          >   }
          > }
```

td-macompta: If I understood correctly, it's not desirable, because, like with the .Files object, Helm is not intended to access a file outside of the current chart. 
This is to avoid malicious behavior

martincomputershare: I understand. Then, what would be the best way to implement my use case?

scme0 (changes requested):
In the interest of progressing this PR, as I think it's valuable, I've given it a review and test.

Generally I think the approach is good, but there are a couple of things that need to be fixed and some more tests need to be added.

During testing I found that `helm lint` was failing with this error:
```
[ERROR] values.yaml: unable to compile schema: Reference already exists: "https://helm.sh/schemas/addresses"
```
This is fixed with my [change suggestion](https://github.com/helm/helm/pull/12553#discussion_r1602637885) in `values.go`

Other than that, I have also tested `helm template` and `help upgrade` and `helm install` and they seem to work as expected in respect to schema validation.

martincomputershare:
Spamming here so that it doesn't get closed by a bot.

pkg/lint/rules/values.go
```diff
@@ -76,11 +77,46 @@ func validateValuesFile(valuesPath string, overrides map[string]interface{}) err
...
+		}
+
+		if !dentry.IsDir() {
+			schema, err = os.ReadFile(path)
+			if err != nil {
+				return err
+			}
+
+			extraSchemas = append(extraSchemas, schema)
+		}
```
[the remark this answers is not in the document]
bossmc: Thanks - good spot!

pkg/lint/rules/values.go
```diff
@@ -76,11 +77,46 @@ func validateValuesFile(valuesPath string, overrides map[string]interface{}) err
...
 	if len(schema) == 0 {
 		return nil
 	}
 	if err != nil {
 		return err
 	}
-	return chartutil.ValidateAgainstSingleSchema(coalescedValues, schema)
+
+	extraSchemas := make([][]byte, 0)
+	schemasPath := filepath.Dir(valuesPath) + "/schemas"
```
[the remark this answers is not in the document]
bossmc: Ugh, naming is hard 😞, from a JSONSchema perspective, these are all schemas (there's no such thing as a "subschema", all schemas are schemas, and a schema can refer to another).  The key difference is that the `values.schema.json` is the thing that the `values.yaml` file is checked against - whereas the others are only there to be referred to.

pkg/chartutil/testdata/test-values-with-extra.schema.json
bossmc: From my understanding of helm's linting rules, I think your use case is already implemented?  When chart A depends on chart B, chart A's value structure is automatically augmented with a "B" key that contains the value structure for the child B chart (and is linted against B's schema if present).

A more complex case would be that chart B exposes a utility schema (e.g. a schema for a type that's shared between multiple charts' schemas) - this would be similar to a library chart, exposing templates to multiple parent charts.  In theory in this case helm could be extended to allow a schema in a parent chart to import a schema from a child chart (then all the parents can include the same child to share that schema).  That would probably be a fairly simple extension of the groundwork laid in this PR, but I'd prefer to get the basic function in first before getting too complicated.

ehoalcraft:
This is exactly what I need.  Hoping this PR will succeed!

Xitric:
Considering the (albeit old) HIP helm/community#200 and the recent feature request #30726, would it make sense to combine all schemas of a chart into a single, self-contained file during publish, rather than to store multiple schema files in the resulting chart?

If a chart consumer runs a command such as `helm show schema`, I think it would be better if the command returned a fully-functional schema with all local references embedded, rather than to return the contents of `values.schema.json` with references that cannot be resolved by the end-user.

bossmc:
I don't really see the advantage to modifying the chart on publish, that would mean that the chart is different when used locally and remotely?  It would also get in the way of the suggestion earlier that a parent chart might be able to refer to a schema in a child (possibly library) chart in order to ensure structural compatibility between the parent's schema and the child's.

A hypothetical `helm show schema` probably should show all the schema parts, but that's as simple as concatenating every file with `---` between them - would you expect `helm show schema` to resolve `http(s)` references too?

In case you were suggesting this, inlining the references is not possible, since one key use of references is breaking infinite loops (e.g. "this object is a foo with an optional child foo" would be of infinite size if the `foo` reference was inlined).

Xitric:
> I don't really see the advantage to modifying the chart on publish, that would mean that the chart is different when used locally and remotely?

Not in function, since the schemas work the same regardless of whether you reference a local schema or place the sub-schema definition in-line.

> It would also get in the way of the suggestion earlier that a parent chart might be able to refer to a schema in a child (possibly library) chart in order to ensure structural compatibility between the parent's schema and the child's.

No, this would still work, since the effort of packing together the JSON schemas happens at each level.

Consider this scenario: an umbrella chart references a sub-umbrella chart, that references a subchart. Each umbrella chart references the schema of its subchart. How would the output of `helm show schema` look when executed against he top-level umbrella chart? Do we wxtract all the schemas of subcharts transitively? Or do we return a single, self-contained JSON schema that is structurally equivalent to all subcharts, transitively?

>  would you expect helm show schema to resolve http(s) references too?

There would be no need for this, since they should be possible to resolve by anyone.

> In case you were suggesting this, inlining the references is not possible, since one key use of references is breaking infinite loops (e.g. "this object is a foo with an optional child foo" would be of infinite size if the foo reference was inlined).

There are more thorough approaches to this than simply in-lining the schema definition as plain-text. For instance, you can place the subschema in a `$defs` section of the resulting, self-contained JSON schema, and use local references within the schema instead of reaching out to the file system :)

Xitric:
> A hypothetical helm show schema probably should show all the schema parts, but that's as simple as concatenating every file with --- between them

That would make it harder for users of the chart to store these schemas locally in a way that enables them to reference schemas of subcharts.

To clarify my use case, we want to make it easy for users of a chart (both standalone charts and umbrella charts) to obtain a reference to the schema of a Helm chart for multiple purposes:

1. Referencing schema of subcharts
2. Referencing schemas of charts to integrate with IntelliSense of your IDE

We have built a plugin for Helm to handle this today, but would like to see it supported natively.

> In case you were suggesting this, inlining the references is not possible, since one key use of references is breaking infinite loops (e.g. "this object is a foo with an optional child foo" would be of infinite size if the foo reference was inlined).

I just tested this with our internal plugin. Given the following schemas:

`values.schema.json`
```json
{
  "properties": {
    "prop1": {
      "type": "integer"
    },
    "prop2": {
      "$ref": "subschema.json"
    }
  }
}
```

`subschema.json`
```json
{
  "properties": {
    "prop": {
      "$ref": "subschema.json"
    }
  }
}
```

We generate the following `values.schema.json` on publish:

```json
{
  "properties": {
    "prop1": {
      "type": "integer"
    },
    "prop2": {
      "properties": {
        "prop": {
          "$ref": "#/properties/prop2"
        }
      }
    }
  }
}
```

They are functionally equivalent, we eliminate all local file references to subschemas, and recursive references are still supported :)

PR #12557, opened 2023-11-10
Feat/repo import command

### LINKED ISSUE #12542: Import list of repositories from a file
We are able to export list of added repositories to a file:
`helm repo ls -o json > helm_repo_ls.json`

It would be nice to have ability to import list of repositories from the same json/yaml file directly using `helm` binary, instead of using hand-made scripts.
For example:
```sh
#!/usr/bin/env bash

if ! command -v jq &> /dev/null; then
  echo "jq is not installed. Please install jq before running this script."
  exit 1
fi

if ! command -v helm &> /dev/null; then
  echo "helm is not installed. Please install helm before running this script."
  exit 1
fi

json_file="helm_repo_ls.json"

# Check if the JSON file exists
if [ ! -f "$json_file" ]; then
  echo "JSON file not found: $json_file"
  exit 1
fi

echo "Repositories to add: $(jq length helm_repo_ls.json)"

jq -c '.[]' "$json_file" | while read -r json; do
  name=$(echo "$json" | jq -r '.name')
  url=$(echo "$json" | jq -r '.url')
  if [[ -n "$name" && -n "$url" ]]; then
    echo "Adding Helm repo: $name - $url"
    helm repo add "$name" "$url" --force-update
  else
    echo "Skipping invalid entry in JSON: $json"
  fi
done

echo "Finished adding Helm repos."
```

itaispiegel:
closes #12542.
This PR adds an option to import repositories from an external file. To use this, run:
`helm repo import my-repos.yaml`

The sub-command I implemented uses internally `helm repo add` which already has validations for the added repositories.

### DIFF
--- a/cmd/helm/repo.go
+++ b/cmd/helm/repo.go
@@ -45,6 +45,7 @@ func newRepoCmd(out io.Writer) *cobra.Command {
 	cmd.AddCommand(newRepoRemoveCmd(out))
 	cmd.AddCommand(newRepoIndexCmd(out))
 	cmd.AddCommand(newRepoUpdateCmd(out))
+	cmd.AddCommand(newRepoImportCmd(out))
 
 	return cmd
 }
--- a/cmd/helm/repo_import.go
+++ b/cmd/helm/repo_import.go
@@ -1,0 +1,55 @@
+/*
+Copyright The Helm Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"io"
+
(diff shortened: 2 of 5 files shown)


den-is:
Any specific reason on why `YAML` was chosen as a primary and only source to import from?

itaispiegel:
Yes. I thought YAML would be the perfect choice, since it's widely used,
and in particular it's used here, in Helm. Also, YAML is a superset of
JSON, which means that the import command works perfectly with JSON files
as well.

den-is:
Ah right. This is a """helm's feature""" to be able parse json files too. (saying "helm's feature", coz when I was developing something small in Go, my standard YAML file argument and parser were working only with YAML files. I did not start digging too deep to solve this)
I'm using it for passing JSON values files instead of YAML files.
I think that things should be documented, so users know that file argument accepts both YAML or JSON.

itaispiegel:
> Ah right. This is a """helm's feature""" to be able parse json files too. (saying "helm's feature", coz when I was developing something small in Go, my standard YAML file argument and parser were working only with YAML files. I did not start digging too deep to solve this) I'm using it for passing JSON values files instead of YAML files. I think that things should be documented, so users know that file argument accepts both YAML or JSON.

I think it's expected from people who are using YAML to know that it's a superset of JSON, but I do agree the command could be documented better.

scottrigby:
Interesting, but did you know you can already pass in a custom file using `--repository-config` flag (path to the file containing repository names and URLs)?

Issue #12466, opened 2023-10-06
Feature Request: bundle value sets with chart


howardjohn:
tl;dr: add a mechanism to bundle files containing values with the chart itself, allowing users to do things like `helm install --values builtin-settings`.

Chart authors today can provide fine grained and/or coarse grained settings. For example, to enable an "HA" mode for a chart, they may have users...

Configure a lot of things:
```yaml
hpa:
  enabled: true
  min: 3
pdb:
  enabled: true
database:
  mode: postgres
  ...
```

Or they may abstract thing:
```yaml
ha: true
```

The former gives maximum configuration, but is hard to customize and complex to reason about (and implement as a chart author). For example, what if they set `ha=true;hpa.enabled=false`?

A nice way to handle this is to instead provide a file `hpa.yaml` somewhere with the "fine grained" contents. This allows fairly simple usage: `helm install -f hpa.yaml`.

However, the issue is: where are these files hosted? A completely different distribution mechanism is needed for 'example values' than the rest of usage. Even if its a URL used, its still a bit awkward to host the chart in OCI and the value samples over HTTP (for example).

---

I propose that we add some concept of "bundled value sets" with charts.  Not tied to any names or UX, but this would look something like:

* In the chart, a new folder `values/` could be optionally added. Within this, users place `<name>.yaml` files.
* `values/` is bundled as part of the chart
* Users can use them with `--values <name>` (note no trailing .yaml suffix) (alternatives: new flag, builtin://name, etc)

jacob-delgado:
I would love this as well as being able to move all of the "globals" to a single section as well.

jacob-delgado:
Something similar to https://github.com/nico-ulbricht/helm-multivalues/tree/master looks like something that can be explored.

howardjohn:
I implemented a lightweight alternative in https://github.com/helm/helm/pull/12565

howardjohn:
Some other areas this can help (with some other tricks in https://github.com/istio/istio/issues/47838)

Allowing projects to evolve on defaults without breaking existing users:
```
$ helm install # v1.0.0
somevalue=legacy
# ... v1.1.0 releases, we want to change value to "new"
$ helm upgrade # v1.1.0. We preserve old value
somevalue=legacy
$ helm upgrades --profile=v1.1.0 # opt into modern defaults, using profiles!
somevalue=new
$ helm uninstall; helm install # fresh install gets new values.
somevlaue=new
[...]

PR #12565, opened 2023-11-13
Support overwritting values from `import-values` dependencies

howardjohn:
Alternative to https://github.com/helm/helm/issues/12466

This allows the above requirement to be implemented in terms of existing Helm functionality. Instead of built in support for a "bundled value set", a chart can expose a dependency on a chart which provides some sets of values. These can be conditionally enabled with `import-values` and `conditions` on the dependency.

The one gap in this approach is the parent values take precedence. This allows a user to *opt in* to making children high precedence.

Additionally, a unit test and additional validation for the field is added.

Note: overall, this is a worse experience than #12466 could provide for this use case. So I would prefer to still do that if feasible. However, this is a low-cost approach

**If applicable**:
- [ ] this PR contains documentation - NO, I can send a PR to helm-www if this merges though


howardjohn:
Actually don't think we need this, will pause until further notice

Issue #12563, opened 2023-11-13
timeout for creating CRDs is hardcoded to 1 minute.


scottrix:
Output of `helm version`: version.BuildInfo{Version:"v3.13.1", GitCommit:"3547a4b5bf5edb5478ce352e18858d8a552a4110", GitTreeState:"clean", GoVersion:"go1.20.8"}

Output of `kubectl version`: Client Version: v1.28.2-eks-a5df82a
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.28.2-eks-f8587cb

Cloud Provider/Platform (AKS, GKE, Minikube etc.):  EKS

In pkg/actions/install.go there is a hardcoded timeout for installing CRDs "i.cfg.KubeClient.Wait(totalItems, 60*time.Second)" it would be useful to be able to control this on the command line.  Some charts have a lot of CRDs to install and 1 minute isn't always enough.

mattfarina:
Thanks for the suggestion. This makes sense. A pull request to add the feature is welcome.

opencmit2:
Hi@mattfarina 
Let me try it, please assign it to me.

opencmit2:
There is a Timeout in the parameter that acts on time to wait for any individual Kubernetes operation (like Jobs for hooks).
If this parameter is also used for CRD, is it reasonable?

opencmit2:
There is a Timeout in the parameter that acts on time to wait for any individual Kubernetes operation (like Jobs for hooks).
If this parameter is also used for CRD, is it reasonable?

PR #12570, opened 2023-11-15
Fix ISSUE #12563 by reusing the command line Timeout parameter to res…

opencmit2:
…olve hardcoded timeouts for CRDs.

This PR fixes ISSUE #12563 by utilizing the existing command line Timeout parameter instead of hardcoding timeouts for CRDs. This allows for more flexibility and easier configuration of timeouts.

**If applicable**:


pkg/action/install.go
```diff
@@ -174,7 +173,7 @@ func (i *Install) installCRDs(crds []chart.CRD) error {
 	}
 	if len(totalItems) > 0 {
 		// Give time for the CRD to be recognized.
-		if err := i.cfg.KubeClient.Wait(totalItems, 60*time.Second); err != nil {
+		if err := i.cfg.KubeClient.Wait(totalItems, i.Timeout); err != nil {
```
gjenkins8, line 176: Strictly, this is a behavior change. A user with a very fast to deploy change, might have e.g. `--timeout=5s`. With the understanding that CRDs would wait for a minute.

I think we should continue the existing 60s (as minimum), and only allow the timeout to be extended: `max(60*time.Second, i.Timeout)`

gjenkins8: The alternative, would be to introduce a new command line flag / API option. Personally, I feel like this is unnecessary here (but that is up for discussion)

opencmit2: Hi @gjenkins8  
I feel that what you said about a minimum of 60 seconds is indeed correct. It does make more sense logically.

opencmit2: This is indeed logically reasonable, but I found that many reused timeouts do not have a fixed minimum value.
image

image

cmd/helm/install.go
```diff
@@ -177,7 +177,7 @@ func addInstallFlags(cmd *cobra.Command, f *pflag.FlagSet, client *action.Instal
 	f.BoolVar(&client.Force, "force", false, "force resource updates through a replacement strategy")
 	f.BoolVar(&client.DisableHooks, "no-hooks", false, "prevent hooks from running during install")
 	f.BoolVar(&client.Replace, "replace", false, "re-use the given name, only if that name is a deleted release which remains in the history. This is unsafe in production")
-	f.DurationVar(&client.Timeout, "timeout", 300*time.Second, "time to wait for any individual Kubernetes operation (like Jobs for hooks)")
+	f.DurationVar(&client.Timeout, "timeout", 300*time.Second, "time to wait for any individual Kubernetes operation (like Jobs or CRDs for hooks)")
```
joejulian, line 180: nit: Did the original text have a typo? Should it have been "Jobs _or_ hooks"?

opencmit2: There is no typo in the original text, but this parameter has also been added to the crds reuse

opencmit2: like Jobs and CRDs for hooks ？

PR #2976, opened 2017-09-27
Ignore Apiversion, use Kind and Name only to search for original resource on update/delete

redbaron:
Kind can change it's APIVersion with time, such as "Deployment" which was
"extensions/v1beta1" and then migrated to "apps/v1beta1" in Kube 1.6.

Name, Kind and optinally Namespace are used to find relevant Info
object for a original resource which is being upgraded, it safe enough,
as it is uniquely identifies an object.

Fixes: #2885

PR #12581, opened 2023-11-20
Consider full GroupVersionKind when matching resources

### LINKED ISSUE #12578: Helm does not remove old resource when switching to a new apiVersion
I have the same issue as this one [issue](https://github.com/helm/helm/issues/10748), but the issue is closed and not resolved

When migrating resource with `kind: CronJob` from `apiVersion: batch/v1` to self-written `apiVersion: kubecronic.<company-name>/v1` in existing release, Helm not deleted the resource which contains an old api version, and created another one.
 
For deployment it uses werf v1.2.248, but under the hood it uses Helm
 
Output of `werf helm version`: `version.BuildInfo{Version:"v3.11", GitCommit:"", GitTreeState:"", GoVersion:"go1.20.3"}`

Output of `kubectl version`:  `v1.25.4`

Cloud Provider/Platform (AKS, GKE, Minikube etc.): self-hosted K8s cluster `v1.27.6`

MichaelMorrisEst:
This change shall take Group, Version and Kind from GroupVersionKind into consideration instead of the current behavior of only considering the Kind when matching resources.
This is needed as resources with different apiVersion are being considered the same.

Closes: #12578

**If applicable**:

### DIFF
--- a/pkg/kube/resource.go
+++ b/pkg/kube/resource.go
@@ -81,5 +81,5 @@ func (r ResourceList) Intersect(rs ResourceList) ResourceList {
 
 // isMatchingInfo returns true if infos match on Name and GroupVersionKind.
 func isMatchingInfo(a, b *resource.Info) bool {
-	return a.Name == b.Name && a.Namespace == b.Namespace && a.Mapping.GroupVersionKind.Kind == b.Mapping.GroupVersionKind.Kind && a.Mapping.GroupVersionKind.Group == b.Mapping.GroupVersionKind.Group
+	return a.Name == b.Name && a.Namespace == b.Namespace && a.Mapping.GroupVersionKind == b.Mapping.GroupVersionKind
 }
--- a/pkg/kube/resource_test.go
+++ b/pkg/kube/resource_test.go
@@ -59,3 +59,42 @@ func TestResourceList(t *testing.T) {
 		t.Error("expected intersect to return bar")
 	}
 }
+
+func TestIsMatchingInfo(t *testing.T) {
+	gvk := schema.GroupVersionKind{Group: "group1", Version: "version1", Kind: "pod"}
+	resourceInfo := resource.Info{Name: "name1", Namespace: "namespace1", Mapping: &meta.RESTMapping{GroupVersionKind: gvk}}
+
+	gvkDiffGroup := schema.GroupVersionKind{Group: "diff", Version: "version1", Kind: "pod"}
+	resourceInfoDiffGroup := resource.Info{Name: "name1", Namespace: "namespace1", Mapping: &meta.RESTMapping{GroupVersionKind: gvkDiffGroup}}
+	if isMatchingInfo(&resourceInfo, &resourceInfoDiffGroup) {
+		t.Error("expected resources not equal")
+	}
+
+	gvkDiffVersion := schema.GroupVersionKind{Group: "group1", Version: "diff", Kind: "pod"}
+	resourceInfoDiffVersion := resource.Info{Name: "name1", Namespace: "namespace1", Mapping: &meta.RESTMapping{GroupVersionKind: gvkDiffVersion}}
+	if isMatchingInfo(&resourceInfo, &resourceInfoDiffVersion) {
+		t.Error("expected resources not equal")
+	}
+
+	gvkDiffKind := schema.GroupVersionKind{Group: "group1", Version: "version1", Kind: "deployment"}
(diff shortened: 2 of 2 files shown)


gjenkins8 (commented):
Funnily enough, back in 2017, it was determined that Helm should ignore the API Group/Version when comparing objects: https://github.com/helm/helm/pull/2976

From a through read of that PR (and related issues), I didn't find any reason that made sense to me to make that change (in fact https://github.com/helm/helm/pull/2976#issuecomment-334206483 IMHO was correct, then and today)

I think this PR makes the logic more correct. But given Helm has been like this for a (very) long time, it is probably worth proceeding with (a lot of) caution (my 2c).

jnonino:
Hi, the comment [#2976 (comment](https://github.com/helm/helm/pull/2976#issuecomment-334206483) makes sense, the issue is that when you upgrade a chart that have resources with API group and/or version changes the old resource is not removed but it is not managed by Helm anymore. 

The old resource is left orphaned in Kubernetes, we found that when we upgraded Traefik chart and they changed the API group from `traefik.containo.us` to `traefik.io`, at some point we started seeing weird errors and they were caused by orphaned IngressRoutes that were still active with the old API group.

So, I get that if the API group/version is different Helm mus recognise them as different resources, but if a resource was being managed by Helm and in an upgrade operation a resource is not there anymore because it was replaced with a new resource with new API group/version, Helm should do something with the old resource it was managing, right now it just leave it orphaned as it was never part of the chart.

MichaelMorrisEst:
I'm starting to think my solution is wrong. 
Using 'kubectl apply' with a changed value for apiVersion considers it the same resource as the already deployed resource regardless of the fact the apiVersions are different.
Helm should probably be consistent with this behaviour.

Currently helm will check for an existing resource on the cluster with the same name, namespace and apiVersion as given in the template, and not finding it considers the resource as a new resource and creates it. This, I believe, is where the logic is wrong - it should instead look for a resource on the cluster using the original apiVersion rather than the new apiVersion. This way it will find the original resource and can upgrade then upgrade it.

I will look into the code changes that would be needed to achieve this

MichaelMorrisEst:
I believe the solution of considering resources to be different if the apiVersion is different is consistent with the behaviour of 'kubectl apply', 'kubectl diff'
The change proposed results in the old resource that was being left orphaned, now being deleted

jnonino:
For our use case that would be the expected behaviour, the old resource being deleted rather that left orphaned. You can always know if something like is going to happen by running the command with dry run and analyse the result.

scottrigby (commented):
> @MichaelMorrisEst -- could you please e.g. rebase to kick CI off (again). We should merge this for Helm 4

Overall I agree comparing `GroupVersionKind` is a good idea now. https://github.com/helm/helm/pull/13583 added group since this PR, so you can remove that too after the rebase.

Can you also update/write a test to check these conditions? This test function hasn't been updated in many years: https://github.com/helm/helm/blob/b2456202aedb10b0d6be5d6d320d8428bb932400/pkg/kube/resource_test.go#L27

MichaelMorrisEst:
Hi @gjenkins8 , @scottrigby 
Rebase done and test case added

gjenkins8 (approved):

scottrigby (approved):
LGTM

matheuscscp:
> Funnily enough, back in 2017, it was determined that Helm should ignore the API Group/Version when comparing objects: #2976
> 
> From a through read of that PR (and related issues), I didn't find any reason that made sense to me to make that change (in fact [#2976 (comment)](https://github.com/helm/helm/pull/2976#issuecomment-334206483) IMHO was correct, then and today)
> 
> I think this PR makes the logic more correct. But given Helm has been like this for a (very) long time, it is probably worth proceeding with (a lot of) caution (my 2c).

Well, this has now grave consequences in Helm 4. The right choice was indeed not to take the API version in consideration all along. Because of this PR, Helm now deletes the "old" resource (old API version) and keeps waiting forever on the "new" resource (new API version), which never happens, because the object was the same in etcd all along. We need to revert this PR ASAP

This PR is actually the root cause for this issue: #31768

@MichaelMorrisEst @gjenkins8 @scottrigby @stefanprodan

scottrigby:
@matheuscscp I was wrong earlier in my review of this PR. Glad we discussed in today's dev meeting and you got it sorted in https://github.com/helm/helm/pull/31772. Thanks!

PR #12583, opened 2023-11-20
Added `toYamlPretty` template function

### LINKED ISSUE #12577: `toYaml` does not properly indent list items
Calling `toYaml` always places array items at the same indentation depth as the parent item. Here's an example of this behavior:

Template:
```yaml
---
{{- $valuesInObjForm := mustDeepCopy .Values -}}

# Via pipe:
{{ $valuesInObjForm | toYaml }}

# Via arg:
{{ toYaml $valuesInObjForm }}
```

Values:
```yaml
---
array:
  - val 1
  - val 2

key:
  nestedArray:
    - val 1
    - val 2
```

Output of `helm template`:
```yaml
---
# Source: bugtest/templates/test-template.yaml
# Via pipe:
array:
- val 1
- val 2
key:
  nestedArray:
  - val 1
  - val 2

# Via arg:
array:
- val 1
- val 2
key:
  nestedArray:
  - val 1
  - val 2
```

I would expect the following output instead:
```yaml
---
# Source: bugtest/templates/test-template.yaml
# Via pipe:
array:
  - val 1
  - val 2
key:
  nestedArray:
    - val 1
    - val 2

# Via arg:
array:
  - val 1
  - val 2
key:
  nestedArray:
    - val 1
    - val 2
```

Am I missing something here? If this is an actual bug then I figured somebody else would have found it by now.

Output of `helm version`:
`version.BuildInfo{Version:"v3.13.2", GitCommit:"2a2fb3b98829f1e0be6fb18af2f6599e0f4e8243", GitTreeState:"clean", GoVersion:"go1.21.4"}`
Also tested this on v3.10 with identical results

fheinecke:
Currently the `toYaml` implementation does not indent list items, which causes issues with other tools in the YAML ecosystem such as linters and formatters. Rather than introduce a potentially breaking change to `toYaml` to support this, I've added a `toYamlPretty` function as suggested by somebody in the [related issue](https://github.com/helm/helm/issues/12577).

Given a Golang map described as:
```go
m := map[string]interface{} {
    "key": []int {1, 2, 3}
}
```

The `toYaml` function will produce:
```yaml
key:
- 1
- 2
- 3
```

The `toYamlPretty` function will produce:
```yaml
key:
  - 1
  - 2
  - 3
```

These are both syntactically correct, but other tools use an indented version which is where this function is useful.

I'm guessing this needs docs somewhere, but I'm not sure where to add them. Could somebody please point me in the right direction?

**If applicable**:

closes #12577


pkg/engine/funcs.go
```diff
@@ -88,6 +90,19 @@ func toYAML(v interface{}) string {
 	return strings.TrimSuffix(string(data), "\n")
 }
 
+func toYAMLPretty(v interface{}) string {
+	var data bytes.Buffer
+	encoder := goYaml.NewEncoder(&data)
+	encoder.SetIndent(2)
```
gjenkins8, line 96: I think this should be configurable (but I would agree that the default should be 2)

pkg/engine/funcs.go
```diff
@@ -88,6 +90,19 @@ func toYAML(v interface{}) string {
 	return strings.TrimSuffix(string(data), "\n")
 }
 
+func toYAMLPretty(v interface{}) string {
+	var data bytes.Buffer
+	encoder := goYaml.NewEncoder(&data)
```
gjenkins8, line 95: Is it possible to use the existing `sigs.k8s.io/yaml` ? I know `sigs.k8s.io/yaml` is a wrapper around `gopkg.in/yaml.v3`, and probably for encoding yaml simply a passthrough to `gopkg.in/yaml.v3`. But to keep things "consistent" if possible.

fheinecke:
@gjenkins8 is there anything I can do myself to move this forward?

pkg/engine/funcs.go
fheinecke: For what it's worth this approach is consistent with `toPrettyJson`, [documented here](https://helm.sh/docs/chart_template_guide/function_list/#toprettyjson-musttoprettyjson). This is a Sprig function call that takes no arguments as [documented here](https://masterminds.github.io/sprig/defaults.html#:~:text=.Item.-,toPrettyJson,-%2C%20mustToPrettyJson), and just prints what is conventionally considered "pretty JSON".

gjenkins8: oops, I meant to give an example, that any consistent indent value is allowed. e.g. this is valid yaml:

```
foo:
           - bar1
           - bar2
           - bar3
```

But also, the sprig function could be extended at some point (I think) to have a default ident of '2' if someone (not sure who :) ) would ever want a different default.

gjenkins8 (commented):
I'll ask another maintainer to review. I forget if spring functions can be extended later. This LGTMe

I think before merging, we would want to add some [docs](https://helm.sh/docs/chart_template_guide/function_list/)

fheinecke:
Thanks!

Docs PR here: https://github.com/helm/helm-www/pull/1540

fheinecke:
Is there a maintainer that can merge this? cc @mattfarina and @joejulian as it looks like you two have merged most of the recently closed PRs

sabre1041 (approved):
LGTM

jonathino2590:
Hello how are you? There is a merge date for this PR, so we need it.

fheinecke:
Not that I'm aware of. I've bugged maintainers on Slack two or three times now and gotten nowhere. It looks like maintainers have merged basically no third party PRs in several months now.

robertsirc:
It's a breaking change what are you looking to do with this?

fheinecke:
> It's a breaking change

@robertsirc What part of this is breaking? This adds a new feature, not affecting any existing functionality.

robertsirc:
@fheinecke my bad i commented on the wrong issue. sorry.

markcheret:
I strongly support this PR.

yxxhero (approved):
LGTM

scottrigby (approved):
lgtm 👍 thanks @fheinecke and @gjenkins8

jonathino2590:
???

fheinecke:
@scottrigby Why has the milestone for this been moved?

scottrigby:
@fheinecke Because v3.16.0 was released before this was merged. I'll take responsibility for that one.

This is an example of me being too busy the day I approved to circle back and merge. And other maintainers having so many PRs that no one noticed this one was ready to merge but had not been merged yet. 

I just checked the merge commit since, and all looks good. I'll merge now, so this will be in the next scheduled feature (semver MINOR) release, 3.17.0.

PR #12588, opened 2023-11-21
Make the authorizer and registry authorizer configurable

### LINKED ISSUE #12584: Set custom registryAuthorizer for Go client
## Summary

We're encountering issues doing a programmatic pull from `ghcr.io`. Specifically, the user might have cached or stale credentials in their docker credential store. The default Helm client will send these credentials even if they are known to be expired.

Our chart is public, and we would be unblocked if we could force the Helm client to use anonymous access.

## Solutions

I found this issue: https://github.com/helm/helm/issues/12144 which seems similar. 

If we could override the authorizer using options [here](https://github.com/helm/helm/blob/4e447d87cd3222aa5e5ea2b38170e9d80b99c0b4/pkg/registry/client.go#L82C1-L82C1) then I think this would do what we wanted.

Are you open to a PR that adds an option for this?

I'm thinking it will look something like this:

```go
// ClientOptRegistryAuthorizer returns a function that sets the authorizer used to access to the registry.
func ClientOptRegistryAuthorizer(authorizer auth.Client) ClientOption {
	return func(client *Client) {
		client.authorizer = authorizer
	}
}
```

rynowak:
Fixes: #12584

This change makes the authorizer and registryAuthorizer of the registry client configurable via options. This allows Go SDK users to override the authentication behavior of the client.

In my use-case we're doing programmatic pulls of a public chart stored on ghcr.io. We've found that many users logged into ghcr.io a long time ago and never logged out. This results in our pull getting rejected (on our public chart) because the user has cached/stored credentials. This is a common problem for ghcr.io because they don't use subdomains (`ghcr.io/my-project` not `myproject.ghcr.io`) so the scope of a login is all of ghcr.io.

This PR makes both the authorizer and registryAuthorizer configurable because depending on the exact scenario that may be needed. The default registryAuthorizer only supports a specific implementation of the authorizer.

I wasn't able to find any unit tests for the these features. Please advise on what you think would be helpful to cover and I can add more tests.

**If applicable**:

### DIFF
--- a/pkg/registry/client.go
+++ b/pkg/registry/client.go
@@ -175,6 +175,26 @@ func ClientOptWriter(out io.Writer) ClientOption {
 	}
 }
 
+// ClientOptAuthorizer returns a function that sets the authorizer setting on a client options set. This
+// can be used to override the default authorization mechanism.
+//
+// Depending on the use-case you may need to set both ClientOptAuthorizer and ClientOptRegistryAuthorizer.
+func ClientOptAuthorizer(authorizer auth.Client) ClientOption {
+	return func(client *Client) {
+		client.authorizer = authorizer
+	}
+}
+
+// ClientOptRegistryAuthorizer returns a function that sets the registry authorizer setting on a client options set. This
+// can be used to override the default authorization mechanism.
+//
+// Depending on the use-case you may need to set both ClientOptAuthorizer and ClientOptRegistryAuthorizer.
+func ClientOptRegistryAuthorizer(registryAuthorizer *registryauth.Client) ClientOption {
+	return func(client *Client) {
+		client.registryAuthorizer = registryAuthorizer
+	}
+}
+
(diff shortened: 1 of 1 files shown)


rynowak:
Hi @sabre1041 - I saw you assigned yourself :smile: Please let me know if there's anything I can do to help 👍

sabre1041 (approved):
LGTM

scottrigby (approved):
👌 moved this feature to next MINOR release milestone as we missed the last

hiddeco:
This would be extremely welcome to be able to isolate pulling from registries fully.

blampe:
@scottrigby @gjenkins8 this has been approved for several months, is there anything we can do to help get it merged?

rynowak:
> @rynowak would you be willing to do that from this same branch on your fork?

Sure, I will do that.

rynowak:
Done: https://github.com/helm/helm/pull/13480

Issue #12511, opened 2023-10-19
Values precedence in helm 3.13.1


intetinte:
With this commit (https://github.com/helm/helm/pull/12480) the precedence was changed to:
    1. User specified values (e.g CLI)
    2. Sub-chart values
    3. Imported values
    4. Parent chart values

This makes it only possible to add new values to sub-charts from parent or imported values. Modify existing values is no longer possible. In helm versions 3.8.2-3.12.3 it was possible to fill in empty or null values in the sub-charts from imports or parent chart.  This is no longer possible in helm 3.13.1
In helm 3.13.0, it was possible to change any value in sub-charts from the imported or parent charts.

Is the precedence in helm 3.13.1 the final word on this?

z4ce:
Can you give an example of what you're trying to achieve with expected input and output?

intetinte:
I need a way to set values in the sub-charts from the parent chart. 
Prior to 3.8.2 and in 3.13.0 it is possible to import values in the parent and apply those to sub-charts.
In 3.8.2-3.12 it is possible to set values from an import in the sub-charts if those had null or empty definitions in the sub-charts.
In 3.13.1, it is no longer possible to either update values in sub-charts from the parent or from imports in the parent. Default value, null or empty remains. The only option is when the sub-chart does not have the parameter names defined.

z4ce:
If you give an example of what you're trying and what's not working as you expect, that would help. The functionality here should still work: https://helm.sh/docs/chart_template_guide/subcharts_and_globals/#overriding-values-from-a-parent-chart

intetinte:
Here an import is done with three helm versions. With three different outcome:
Charts:
--------------------------------------
$ cat mychart/values.yaml
favorite:
  drink: coffee
  food: pizza
pizzaToppings:
  - mushrooms
  - cheese
  - peppers
  - onions

profile:
  myimport: true

mysubchart:
  dessert: ice cream

----------------------------------------------
$ cat mychart/charts/mysubchart/values.yaml
dessert: cake
topping:
----------------------------------------------
$ cat mychart/Chart.yaml
apiVersion: v2
appVersion: "1.0"
name: test
version: 1.0.0
description: test
dependencies:
- condition: profile.myimport
  import-values:
  - myimport
  name: myimport
  repository: file://../charts/myimport
  version: 1.2.3
- name: mysubchart
  repository: file://../charts/mysubchart
  version: 0.1.0
-----------------------------------------------------------------------------
[...]

PR #12480, opened 2023-10-11
Fixing precedence issue with the import of values.

mattfarina:
The ordering should be:
1. User specified values (e.g CLI)
2. Parent chart values
3. Imported values
4. Sub-chart values

This enables parent charts to import large set of values from a child and then override select values.

This change is needed for backwards compatibility.

Fixes #12460

**What this PR does / why we need it**: The values fixes in 3.13.0 had the wrong prioritization for values which broke backwards compatibility and did not make sense. Imported values should be able to be used (e.g., with a library chart) while being able override select values.

**Special notes for your reviewer**:
- Tests that change here are restored to 3.12.x values for the tests.
- This restores the previous ordering. [See 3.12.3 here](https://github.com/helm/helm/blob/v3.12.3/pkg/chartutil/dependencies.go#L271).

**If applicable**:
- [ ] this PR contains documentation
- [x] this PR contains unit tests
- [x] this PR has been tested for backwards compatibility

PR #12590, opened 2023-11-22
Fix issue with precedence of imports. Imports in parent did not overr…

intetinte:
…ide values in subcharts

This pull request fixed the issue described in #12511

The value precedence for import from the parent is set higher than the sub chart values. This enables importing sets of data in the parent that is applied in the sub-charts. This has been possible in older helm versions than 3.8.2 and in 3.13.0. This fix doe not interfere with the other fixes in the value precedence - i.e backwards compatible.

Fix for these:
https://github.com/helm/helm/issues/12730
https://github.com/helm/helm/issues/12511

As an addition to these:
https://github.com/helm/helm/pull/12162
https://github.com/helm/helm/pull/12480

### DIFF
--- a/pkg/chartutil/dependencies.go
+++ b/pkg/chartutil/dependencies.go
@@ -304,8 +304,8 @@ func processImportValues(c *chart.Chart, merge bool) error {
 		// deep copying the cvals as there are cases where pointers can end
 		// up in the cvals when they are copied onto b in ways that break things.
 		// Merge dependencies in the chart values-file first
-	    b = MergeTables(deepCopyMap(c.Values), b)
-	    // next merge sub-chart values in the chart-values merged with the dependencies
+		b = MergeTables(deepCopyMap(c.Values), b)
+		// next merge sub-chart values in the chart-values merged with the dependencies
 		c.Values = MergeTables(deepCopyMap(b), cvals)
 	} else {
 		// Trimming the nil values from cvals is needed for backwards compatibility.
--- a/pkg/chartutil/dependencies.go
+++ b/pkg/chartutil/dependencies.go
@@ -303,8 +303,10 @@ func processImportValues(c *chart.Chart, merge bool) error {
 	if merge {
 		// deep copying the cvals as there are cases where pointers can end
 		// up in the cvals when they are copied onto b in ways that break things.
-		cvals = deepCopyMap(cvals)
-		c.Values = MergeTables(cvals, b)
+		// Merge dependencies in the chart values-file first
+	    b = MergeTables(deepCopyMap(c.Values), b)
+	    // next merge sub-chart values in the chart-values merged with the dependencies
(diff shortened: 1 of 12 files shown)


intetinte:
This commit makes it possible to import one or more libraries with values in the parent chart and apply those values on the child charts. The parent library values will take precedence of the child values.
Precedence will be
1. User specified values
2. Parent values
3. parent imports
4. child values
5. child imports

Previously it was:
1. User specified values 
2. Parent values
3. child values
4. parent import
5. child import

intetinte:
Rebased and did some manual comparison with the main branch and the commit on the fork.

Running with helm from main-branch:
$ helm template --set profile.myimport=false /home/epktojo/helmtest/mychart **# disabled the import of library**
\---
\# Source: test/charts/mysubchart/templates/configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: release-name-cfgmap2
data:
  dessert: ice cream   **<--- This value is set from the umbrella chart values.yaml**
  topping: nil    **<--- This value is set in mysubchart values.yaml**
  servedon:      **<--- This parameter is not defined in mysubchart values.yaml**

Running helm from main-branch:
$ helm template --set profile.myimport=true mychart **# enabled the import of library**
\---
\# Source: test/charts/mysubchart/templates/configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: release-name-cfgmap2
data:
  dessert: ice cream **<--- This value is set from the umbrella chart values.yaml**
  topping: nil          **<--- This value is set in mysubchart values.yaml. However the imported library does also set this to another value. _Subchart value takes precedence over library value_**
  servedon: plate    **<--- This value is set in imported library values.yaml**
  
Running helm from this pull request:
$ helm template --set profile.myimport=false mychart **# disabled the import of library**
\---
\# Source: test/charts/mysubchart/templates/configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: release-name-cfgmap2
data:
  dessert: ice cream **<--- This value is set from the umbrella chart values.yaml**
  topping:           **<--- This value is cleared compared with the main branch (was nil) which is correct?**
  servedon:    **<--- This parameter is not defined in mysubchart values.yaml**

Running helm from this pull request:
$ ./git/fork/helm/bin/helm template --set profile.myimport=true /home/epktojo/helmtest/mychart
\---
\# Source: test/charts/mysubchart/templates/configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: release-name-cfgmap2
data:
  dessert: ice cream **<--- This value is set from the umbrella chart values.yaml**
  topping: sugar      **<--- This value is set in imported library values.yaml. Subchart does have value nil. Imported value takes precedence over subchart value**
  servedon: plate    **<--- This value is set in imported library values.yaml**

PR #12593, opened 2023-11-24
fix ensureMissingRepos issue when has same chart name

### LINKED ISSUE #12666: An exception occurs when different dependencies have the same name
```
name: helmfile
version: 1.0.0
apiVersion: v2
dependencies:
- name: operator
  repository: git+https://github.com/minio/operator@helm/operator?ref=master
  version: '*'
- name: operator
  repository: https://charts.redpanda.com
  version: '*'
```
An exception occurs when different dependencies have the same name.

yxxhero:

fix: https://github.com/helm/helm/issues/12666

**If applicable**:

### DIFF
--- a/pkg/downloader/manager.go
+++ b/pkg/downloader/manager.go
@@ -509,6 +509,10 @@ func (m *Manager) ensureMissingRepos(repoNames map[string]string, deps []*chart.
 			continue
 		}
 
+		if _, ok := repoNames[dd.Repository]; ok {
+			continue
+		}
+
 		// When the repoName for a dependency is known we can skip ensuring
 		if _, ok := repoNames[dd.Name]; ok {
 			continue


yxxhero:
cases:
```
name: helmfile
version: 1.0.0
apiVersion: v2
dependencies:
- name: operator
  repository: git+https://github.com/minio/operator@helm/operator?ref=master
  version: '*'
- name: operator
  repository: https://charts.redpanda.com
  version: '*'
```

pkg/downloader/manager.go
```diff
@@ -509,6 +509,10 @@ func (m *Manager) ensureMissingRepos(repoNames map[string]string, deps []*chart.
 			continue
 		}
 
+		if _, ok := repoNames[dd.Repository]; ok {
+			continue
+		}
+
```
mattfarina, line 515: `repoNames` has a structure of `[dd.Name] = dd.Repository`. So, this change isn't going to work. This section of code does need to be refactored to fix this bug. I see the bug.

mattfarina (changes requested):
There is a bug here. Thanks for catching it.

In addition to fixing it, can we get a test to ensure it's fixed and so we don't have a regression in the future.

pkg/downloader/manager.go
MichaelMorrisEst: The repoNames is keyed on the chart name of the dependency, however that will not work in this case as there is more than one dependency with the same chart name.
I would suggest changing the key to the repo url, i.e. [dd.Repository] == repo.Name and refactoring methods 'resolveRepoNames', 'ensureMissingRepos' and 'resolve' accordingly

yxxhero: @MichaelMorrisEst thanks so much.

PR #12617, opened 2023-12-07
feature(pkg/engine): introduce RenderWithClientProvider

porridge:

In order to fully test charts that use the `lookup` function in environments that do not provide a kube API server (e.g. with [helmtest](https://github.com/stackrox/helmtest)), we need to inject a (fake) kubernetes client to the engine.

This change adds a `RenderWithClientProvider` function parallel to the already existing (somewhat unfortunately named) `RenderWithClient`, to make this possible.

Existing functionality is untouched.

**If applicable**:
- [x] this PR has been tested for backwards compatibility: I made sure that the `lookup` template function still works, by running a locally built `helm` binary against a chart which uses this functionality, on a `kind` cluster

Statement coverage with tests:

| file | before | after |
| --- | --- | --- |
| `engine.go` | 89.2% | 88.7% |
| `files.go`  |   85.3% | 85.3% |
| `funcs.go`   | 85.4% | 85.4% |
| `lookup_func.go` | 0% | 33.3% |

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -40,16 +40,17 @@ type Engine struct {
 	Strict bool
 	// In LintMode, some 'required' template values may be missing, so don't fail
 	LintMode bool
-	// the rest config to connect to the kubernetes api
-	config *rest.Config
+	// optional provider of clients to talk to the Kubernetes API
+	clientProvider *ClientProvider
 	// EnableDNS tells the engine to allow DNS lookups when rendering templates
 	EnableDNS bool
 }
 
 // New creates a new instance of Engine using the passed in rest config.
 func New(config *rest.Config) Engine {
+	var clientProvider ClientProvider = clientProviderFromConfig{config}
 	return Engine{
-		config: config,
+		clientProvider: &clientProvider,
 	}
 }
 
@@ -85,10 +86,21 @@ func Render(chrt *chart.Chart, values chartutil.Values) (map[string]string, erro
 
 // RenderWithClient takes a chart, optional values, and value overrides, and attempts to
 // render the Go templates using the default options. This engine is client aware and so can have template
-// functions that interact with the client
+// functions that interact with the client.
 func RenderWithClient(chrt *chart.Chart, values chartutil.Values, config *rest.Config) (map[string]string, error) {
+	var clientProvider ClientProvider = clientProviderFromConfig{config}
 	return Engine{
-		config: config,
+		clientProvider: &clientProvider,
+	}.Render(chrt, values)
+}
+
+// RenderWithClientProvider takes a chart, optional values, and value overrides, and attempts to
+// render the Go templates using the default options. This engine is client aware and so can have template
+// functions that interact with the client.
+// This function differs from RenderWithClient in that it lets you customize the way a dynamic client is constructed.
+func RenderWithClientProvider(chrt *chart.Chart, values chartutil.Values, clientProvider ClientProvider) (map[string]string, error) {
+	return Engine{
+		clientProvider: &clientProvider,
 	}.Render(chrt, values)
 }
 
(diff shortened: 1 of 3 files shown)


pkg/engine/lookup_func.go
```diff
@@ -39,9 +40,30 @@ type lookupFunc = func(apiversion string, resource string, namespace string, nam
...
 func NewLookupFunction(config *rest.Config) lookupFunc {
-	return func(apiversion string, resource string, namespace string, name string) (map[string]interface{}, error) {
+	return newLookupFunctionForConfig(config)
+}
+
+func newLookupFunctionForClient(client dynamic.Interface) lookupFunc {
+	return newLookupFunction(func(apiVersion, kind string) (dynamic.NamespaceableResourceInterface, bool, error) {
+		// This is suboptimal but the best kind->resource translation we can do without a discovery client.
+		gvr, _ := meta.UnsafeGuessKindToResource(schema.FromAPIVersionAndKind(apiVersion, kind))
+		namespaced := true // we infer this from the namespace argument to the lookup function
```
gjenkins8, line 50: can we detail what is happening here please. Any limitations of this functionality? And how come we can't use the resources gvk when invoked from the `lookup` function.

(the three levels of function closures are bending my brain a little; I think I might be missing something)

gjenkins8: Thinking about this more, if a non-Helmtest user started to use this `RenderWithDynamicClient` interface, they might (I think likely) be surprised by the `meta.UnsafeGuessKindToResource` usage here. Given my (admittedly superficial) understanding, that this function is trying to "guess" the resource name from the Version, Kind info ([meta.UnsafeGuessKindToResource source](https://github.com/kubernetes/apimachinery/blob/e2f405af78de67e27b5dd5ceb1a3eca76331e4d5/pkg/api/meta/restmapper.go#L126C8-L126C8)).

It seems to me, that it would be better to hoist the `clientProviderFunc` functionality to external Render interface? Rather than passing in a dynamic client, which gets only some of the way. Then needing to build in a workaround to use it.

Something like:
```
func RenderWithClientLookupFunc(chrt *chart.Chart, values chartutil.Values, clientLookupFunc ClientLookupFunc) (map[string]string, error) {
	return Engine{
		clientLookupFunc: ClientLookupFunc,
	}.Render(chrt, values)
}
```

porridge: @gjenkins8 thank you for the prompt review! You formulated clearly the concerns I had myself floating somewhere on the edge of my perception :sweat_smile:

I updated the PR to follow your advice. This also has the advantages of:
- not introducing any new function closures, at the expense of adding a new `struct`-based type, which I think is an acceptable overhead in this place,
- eradicating the somewhat dubious `UnsafeGuessKindToResource` call to the helmtest codebase, where it's appropriate.

pkg/engine/engine.go
```diff
@@ -40,16 +40,17 @@ type Engine struct {
...
 	// EnableDNS tells the engine to allow DNS lookups when rendering templates
 	EnableDNS bool
 }
 
 // New creates a new instance of Engine using the passed in rest config.
 func New(config *rest.Config) Engine {
+	var clientProvider ClientProvider = clientProviderFromConfig{config}
 	return Engine{
-		config: config,
+		clientProvider: &clientProvider,
```
gjenkins8, line 53: Nice, IMHO it is great to be able to re-write the existing implementation in terms of this new, more generic, interface

gjenkins8 (approved):
Changes LGTMe, will ask another maintainer to review. Thanks for updating the PR

joejulian:
There are no tests with this. Are the existing tests sufficient? How can I manually validate this?

joejulian (approved):
Thanks!

mattfarina (approved):

Issue #12623, opened 2023-12-09
Output of `helm show crds` is not a valid yaml file


papdaniel:
Output of `helm version`:
```
version.BuildInfo{Version:"v3.13.2", GitCommit:"2a2fb3b98829f1e0be6fb18af2f6599e0f4e8243", GitTreeState:"clean", GoVersion:"go1.20.10"}
```

Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.4", GitCommit:"e87da0bd6e03ec3fea7933c4b5263d151aafd07c", GitTreeState:"clean", BuildDate:"2021-02-18T16:12:00Z", GoVersion:"go1.15.8", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"24", GitVersion:"v1.24.0", GitCommit:"4ce5a8954017644c5420bae81d72b09b735c21f0", GitTreeState:"clean", BuildDate:"2022-05-19T15:39:43Z", GoVersion:"go1.18.1", Compiler:"gc", Platform:"linux/amd64"}
```
Cloud Provider/Platform (AKS, GKE, Minikube etc.): 
`kind v0.14.0 go1.18.2 linux/amd64`

It would be nice if the output of `helm show crds` was a valid yaml file. Currently it is not, as the output does not contain any document separators (`---`).
Let's say I want:
```
helm repo add apisix https://charts.apiseven.com
helm repo update
helm show crds apisix/apisix-ingress-controller --version 0.12.2 | kubectl apply -f -
```
The result is:
```
customresourcedefinition.apiextensions.k8s.io/apisixupstreams.apisix.apache.org created
```
Only the last CRD has been created, out of: 
```
ApisixClusterConfig.yaml
ApisixConsumer.yaml
ApisixGlobalRule.yaml
ApisixPluginConfig.yaml
ApisixRoute.yaml
ApisixTls.yaml
ApisixUpstream.yaml
```

gjenkins8:
Duplicate of: https://github.com/helm/helm/issues/11261

(but the PR associated with that issue seems to be abandoned; happy to accept a new PR)

Issue #11261, opened 2022-08-24
Inject document separators when rendering CRDs


mkjpryor:
I have some code that runs `helm show crds <chart args>` to get the CRDs for a chart, then parses the result as YAML to get the list of CRD objects.

This works fine if the CRD files in the chart include a leading `---`. However in the case where the CRD files do **not** include a leading `---`, `helm show crds` outputs a single YAML document where the attributes for one CRD splat over the attributes of the previous one and my code only picks up one CRD.

I propose that Helm should inject the document separators when they are not present.

Output of `helm version`:

```
version.BuildInfo{Version:"v3.9.3", GitCommit:"414ff28d4029ae8c8b05d62aa06c7fe3dee2bc58", GitTreeState:"clean", GoVersion:"go1.19"}
```

vaibhavkoshti7:
hello @joejulian . I want work on this issue .Can you assign it to me ?

joejulian:
GitHub doesn't let me assign. Your comment should be sufficient to let folks know you're working on it, I think.

mkjpryor:
This is still a problem!

PR #12624, opened 2023-12-09
show crds command output separated by document separator

papdaniel:

**If applicable**:

~~issue: https://github.com/helm/helm/issues/12623~~
issue: https://github.com/helm/helm/issues/11261


pkg/action/show_test.go
```diff
@@ -30,8 +30,8 @@ func TestShow(t *testing.T) {
 		Files: []*chart.File{
 			{Name: "README.md", Data: []byte("README\n")},
 			{Name: "crds/ignoreme.txt", Data: []byte("error")},
-			{Name: "crds/foo.yaml", Data: []byte("---\nfoo\n")},
```
gjenkins8, line 33: These tests are showing that a CRD yaml which already has `---` renders correctly

papdaniel: True, fixed

gjenkins8 (approved):
The final question here, is if Helm can consider this a "backwards compatible" change (unfortunately, people do things like parse CLI output, rather than using e.g. structured output. And even changing simple things like this can break peoples "scripts")

Apart from the above (someone else will need to opine), this change looks good to me

papdaniel:
Could someone review this?

shashankram:
@mattfarina could you review this fix?

joejulian (approved):
This tests out fine. WRT @gjenkins8 concern about a breaking change, this is for v4 now so a breaking change is allowed.

LGTM

Issue #12607, opened 2023-12-02
Helm template will cache the chart but does not use the cache


Ithrael:
I noticed that although helm template can cache, if the repositoryCache already has a chart, running helm template again will still re-download the chart and cache it in the repositoryCache directory. 

Following is my testing process.
```
rm -rf /Users/w/Library/Caches/helm/repository/grafana-7.0.11.tgz

helm template grafana/grafana
```

<img width="1125" alt="image" src="https://github.com/helm/helm/assets/38208715/a7f9134b-b232-4551-bb1d-5714f042f0e2">

run again
`helm template grafana/grafana`
<img width="1059" alt="image" src="https://github.com/helm/helm/assets/38208715/d3008854-7362-4a6e-951f-fbe8247d9e28">

The caching for the helm template seems to be pointless, is this normal?

Output of `helm version`:
v3.13+unreleased

Output of `kubectl version`:

v1.27.4
Cloud Provider/Platform (AKS, GKE, Minikube etc.):

gjenkins8:
I think if this issue is solved for pulling (https://github.com/helm/helm/issues/12602), then it will be solved for templating (since template is just pulling under the hood')

rassie:
Can this be reopened? It's rather annoying when used with frameworks like `cdk8s`, since it's fetch charts at every single generation cycle which takes time and bandwidth.

PR #12628, opened 2023-12-11
feat: If the required chart is already cached, there is no need to do…

Ithrael:
…wnload it again

https://github.com/helm/helm/issues/12607

**If applicable**:

### DIFF
--- a/pkg/downloader/chart_downloader.go
+++ b/pkg/downloader/chart_downloader.go
@@ -97,11 +97,6 @@ func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *proven
 		return "", nil, err
 	}
 
-	data, err := g.Get(u.String(), c.Options...)
-	if err != nil {
-		return "", nil, err
-	}
-
 	name := filepath.Base(u.Path)
 	if u.Scheme == registry.OCIScheme {
 		idx := strings.LastIndexByte(name, ':')
@@ -109,8 +104,22 @@ func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *proven
 	}
 
 	destfile := filepath.Join(dest, name)
-	if err := fileutil.AtomicWriteFile(destfile, data, 0644); err != nil {
-		return destfile, nil, err
+	_, err = os.Stat(destfile)
+
+	// If the required chart is already cached, there is no need to download it again
+	if err != nil {
+		if os.IsNotExist(err) {
+			data, err := g.Get(u.String(), c.Options...)
+			if err != nil {
+				return "", nil, err
+			}
+
+			if err := fileutil.AtomicWriteFile(destfile, data, 0644); err != nil {
+				return destfile, nil, err
+			}
+		} else {
+			return destfile, nil, err
+		}
 	}
 
 	// If provenance is requested, verify it.
--- a/pkg/downloader/chart_downloader_test.go
+++ b/pkg/downloader/chart_downloader_test.go
@@ -197,6 +197,7 @@ func TestDownloadTo(t *testing.T) {
 		},
 	}
 	cname := "/signtest-0.1.0.tgz"
+	// Distinguish the server's file directory from the downloaded directory.
 	dest := srv.Root() + "/dest"
 	os.Mkdir(dest, 0755)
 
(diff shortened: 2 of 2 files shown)


pkg/downloader/chart_downloader.go
```diff
@@ -97,20 +97,29 @@ func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *proven
...
 	name := filepath.Base(u.Path)
 	if u.Scheme == registry.OCIScheme {
 		idx := strings.LastIndexByte(name, ':')
 		name = fmt.Sprintf("%s-%s.tgz", name[:idx], name[idx+1:])
 	}
 
 	destfile := filepath.Join(dest, name)
-	if err := fileutil.AtomicWriteFile(destfile, data, 0644); err != nil {
-		return destfile, nil, err
+	_, err = os.Stat(destfile)
```
gjenkins8, line 107: There is a change in behavior here IMHO. If the chart has changed in the upstream registry, then the cache won't be invalidated.

It could be argued that released charts should never change (and I would tend to agree). The operative word is _should_, of course. Someone out there is probably updating a chart with re-versioning, and expecting it to work. I don't think we can break them here.

And prerelease version can be republished (under semver rules) I believe.

pkg/downloader/chart_downloader.go
```diff
@@ -97,20 +97,29 @@ func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *proven
 		return "", nil, err
 	}
 
-	data, err := g.Get(u.String(), c.Options...)
```
gjenkins8, line 100: Can we fetch the chart metadata only? And compare if the upstream content has changed?

That still incurs a network access, but avoids downloading the content if unchanged.

pkg/downloader/chart_downloader.go
jouve: the [digest](https://github.com/helm/helm/blob/8b4be22a81cb61e30affe3108abc84efd34b7f54/pkg/repo/index.go#L273) from the repo could be used to check wiether the chart needs to be downloaded again

rassie:
Is there a good reason why this has been closed? Could we revive this patch? I'm coming from https://github.com/helm/helm/issues/12607 and really miss the cache (pun intended).

pkg/downloader/chart_downloader.go
Ithrael: > There is a change in behavior here IMHO. If the chart has changed in the upstream registry, then the cache won't be invalidated.
> 
> It could be argued that released charts should never change (and I would tend to agree). The operative word is _should_, of course. Someone out there is probably updating a chart with re-versioning, and expecting it to work. I don't think we can break them here.
> 
> And prerelease version can be republished (under semver rules) I believe.

Yes, that is a possibility. Then fetching the chart metadata only also requires a network request. Perhaps we can use the --cache flag to specify whether to download or not?

Ithrael: > There is a change in behavior here IMHO. If the chart has changed in the upstream registry, then the cache won't be invalidated.
> 
> It could be argued that released charts should never change (and I would tend to agree). The operative word is _should_, of course. Someone out there is probably updating a chart with re-versioning, and expecting it to work. I don't think we can break them here.
> 
> And prerelease version can be republished (under semver rules) I believe.

Yes, that is a possibility. Then fetching the chart metadata only also requires a network request. Perhaps we can use the --cache flag to specify whether to download or not?

rassie: > Then fetching the chart metadata only also requires a network request. Perhaps we can use the --cache flag to specify whether to download or not?

In my world, the whole idea of local caching is to avoid network requests, i.e. if an application request can be satisfied locally it should do so (my particular use-case is doing a lot of `helm template --repo ...` locally, so I'm obviously biased). Re-released tarballs should not be the norm and there are ways to accomodate them without inducing network requests for well-behaving charts, i.e. re-fetch metadata only on pre-release charts, add a `--no-cache` option, etc. 

But this is still a (somewhat) breaking change. My personal preference would be to mention this in the release notes and leave it at that.

PR #12632, opened 2023-12-12
Update kubernetes default minor version to current supported version

yonahd:
This PR Updates the default Kubernetes minor version to the oldest supported version.
This allows helm template to by default use the newer api versions
The older versions can be supported by passing the `--kube-version ` flag


yonahd:
> helm template -g foo/ | grep kubeversion:

Your right. I was using helm 3.5.4 and embedded in it was version 1.20

PR #12649, opened 2023-12-18
fix: wrong cli description

yyzxw:

**If applicable**:

### DIFF
--- a/cmd/helm/plugin_install.go
+++ b/cmd/helm/plugin_install.go
@@ -39,8 +39,8 @@ This command allows you to install a plugin from a url to a VCS repo or a local 
 func newPluginInstallCmd(out io.Writer) *cobra.Command {
 	o := &pluginInstallOptions{}
 	cmd := &cobra.Command{
-		Use:     "install [options] <path|url>...",
-		Short:   "install one or more Helm plugins",
+		Use:     "install [options] <path|url>",
+		Short:   "install a Helm plugin",
 		Long:    pluginInstallDesc,
 		Aliases: []string{"add"},
 		Args:    require.ExactArgs(1),


cmd/helm/plugin_install.go
```diff
@@ -40,7 +40,7 @@ func newPluginInstallCmd(out io.Writer) *cobra.Command {
 	o := &pluginInstallOptions{}
 	cmd := &cobra.Command{
 		Use:     "install [options] <path|url>...",
-		Short:   "install one or more Helm plugins",
+		Short:   "install one Helm plugins",
```
sabre1041, line 43:
```suggestion
		Short:   "install a Helm plugin",
```

sabre1041 (changes requested):
Good catch. Proposed a grammatical change

cmd/helm/plugin_install.go
```diff
@@ -40,7 +40,7 @@ func newPluginInstallCmd(out io.Writer) *cobra.Command {
 	o := &pluginInstallOptions{}
 	cmd := &cobra.Command{
 		Use:     "install [options] <path|url>...",
```
joejulian, line 42: If this is true, then this should remove the ellipsis.

joejulian (changes requested):

cmd/helm/plugin_install.go
yyzxw: done

mattfarina (approved):

mattfarina:
Merging as the requested changes have been addressed. Thanks for the fix.

mattfarina:
Note, merging with one maintainer approval since this is XS and that's allowed at this size of a change.

PR #12662, opened 2023-12-22
Make the `ignore` pkg public again

ialidzhikov:
In helm@v2, the `ignore` pkg was public and were are using it.
In helm@v3, the `ignore` pkg was made internal with https://github.com/helm/helm/pull/6176.
Having a look into https://github.com/helm/helm/blob/v3.10.3/pkg/chart/loader/directory.go I think every `ChartLoader` implementation (that is different from the default one that reads from the filesystem) has to use the ignore pkg to make `.helmignore` work.
For example, in gardener/gardener we have a logic to load a chart from embedded fiilesystem `embed.FS`: https://github.com/gardener/gardener/blob/cee120166b7b8a857dea90fc0217d86c2e41850f/pkg/chartrenderer/default.go#L193-L260

N/A

**If applicable**:


ialidzhikov:
/cc @mattfarina
ref https://github.com/helm/helm/pull/6176#pullrequestreview-273320867

joejulian (approved):

joejulian:
This is XS and I'm going to take https://github.com/helm/helm/pull/6176#pullrequestreview-273320867 as tacit approval from @mattfarina.

PR #12670, opened 2023-12-27
fix: Enable release action to update helm-latest-version

### LINKED ISSUE #12669: get-helm-3 script fails to get latest version
The `get-helm-3` script installs `v3.13.1` as latest version, not `v3.13.3` as it should when running the following commands.

```
$ curl -fsSL -o get_helm.sh https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3
$ chmod 700 get_helm.sh
$ ./get_helm.sh
```

It seems that `https://get.helm.sh/helm-latest-version` hasn't been updated.

gjenkins8:
Currently, Helm release action (`release.yml`) does not allow updating https://get.helm.sh/helm-latest-version , which causes `get-helm-3` script to download stale versions.

Set the [overwrite](https://github.com/bacongobbler/azure-blob-storage-upload/blob/main/action.yml#L29-L31) input on the Azure upload action to enable updating the `helm-latest-version` object.

fixes: #12669

I also created https://github.com/helm/helm/pull/12672 with some general fixups, and to avoid passing the `overwrite` option to the binary upload 

**If applicable**:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -46,6 +46,7 @@ jobs:
           AZURE_STORAGE_CONNECTION_STRING: "${{ secrets.AZURE_STORAGE_CONNECTION_STRING }}"
           AZURE_STORAGE_CONTAINER_NAME: "${{ secrets.AZURE_STORAGE_CONTAINER_NAME }}"
         with:
+          overwrite: 'true'
           source_dir: _dist
           container_name: ${{ secrets.AZURE_STORAGE_CONTAINER_NAME }}
           connection_string: ${{ secrets.AZURE_STORAGE_CONNECTION_STRING }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -45,6 +45,7 @@ jobs:
         env:
           AZURE_STORAGE_CONNECTION_STRING: "${{ secrets.AZURE_STORAGE_CONNECTION_STRING }}"
           AZURE_STORAGE_CONTAINER_NAME: "${{ secrets.AZURE_STORAGE_CONTAINER_NAME }}"
+        overwrite: 'true'
```
z4ce, line 48: This should be in the 'with' statement

z4ce (changes requested):

z4ce (approved):

yxxhero (approved):

PR #12690, opened 2024-01-07
feat: OCI install by digest

### LINKED ISSUE #10678: OCI Specify Digest in Version Parameter
Hi,

I was wondering what your thoughts towards specifying the digest for an OCI Chart is. It would be useful in cases where you want to be sure that a very specific Chart is used, e.g. in cases where you want to perform a rollback.

A suggestion has already been made [here](https://github.com/helm/helm/issues/10312#issuecomment-961396434). Currently, there is a strong need to have the Chart version match the OCI tag and I think that this can still stay valid. My suggestion would be to add an optional `--digest` parameter to all commands that download the remote OCI manifest. The digest will be added to the final download tag, e.g. `helm install ... oci://registry.com/chart --version 1.0.0 --digest sha256:1f2d257aacc56f95650049a2c077790e37365ddff9ad2370ab153a310baa90a1` yields a full reference of `registry.com/test:1.0.0@sha256:1f2d257aacc56f95650049a2c077790e37365ddff9ad2370ab153a310baa90a1`.

TerryHowe:
Ths PR Allows helm OCI install of charts using digest. Supported formats include:

```
helm install  billy oci://localhost:5000/demo@sha256:d1c2884a2ac2d2f80fb1bf384e45b4cc72669498ccd237843dcc63bfcac810a3 --version 0.1.0
helm install  billy oci://localhost:5000/demo:0.1.0 --version 0.1.0
```
Without version option
```
helm install  billy oci://localhost:5000/demo@sha256:d1c2884a2ac2d2f80fb1bf384e45b4cc72669498ccd237843dcc63bfcac810a3
helm install  billy oci://localhost:5000/demo:0.1.0
```

**If applicable**:

Closes: https://github.com/helm/helm/pull/10799
Partial: https://github.com/helm/helm/issues/10312
Closes: https://github.com/helm/helm/issues/10678
Closes: https://github.com/helm/helm/pull/12558

Related docs PR https://github.com/helm/helm-www/pull/1629


sabre1041 (commented):
@TerryHowe Thanks for bringing this forward. Being able to obtain a chart via digests is a feature that is certainly desired. 

It looks really good. One area that I would like to discuss is when both a tag/digest is provided along with a version. When a tag differs from the version, an error is correctly emitted. However, when a digest and a version is provided, only the digest used and the version is silently ignored.

This brings up the question. Should a version flag be used at all when specifying a tag/digest?

honnix:
This is a great feature! Is there a plan to move it forward? Thanks.

TerryHowe:
> This is a great feature! Is there a plan to move it forward? Thanks.

This is still on my list, but haven't had time to address @sabre1041 comment

emosbaugh:
Any plans to pick this back up soon? This would be a great feature to add.

banjoh:
@TerryHowe if you need additional help, let me know. I'm interested in the feature and I might have cycles to spare

TerryHowe:
Unit tests are passing for me and functionally seems to work fine.

origxm:
This is a great improvement!
Hopefully, it will be merged and will be part of the next release.

pkg/downloader/chart_downloader.go
```diff
@@ -143,7 +143,43 @@ func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *proven
...
+	} else {
+		if registryReference.Tag != "" && registryReference.Tag != version {
+			return nil, errors.Errorf("chart reference and version mismatch: %s is not %s", version, registryReference.Tag)
+		}
+	}
+
+	if registryReference.Digest == "" {
+		if version == "" {
+			return u, errors.Errorf("a digest or version must be specified for an OCI chart")
+		}
```
mattfarina, line 164: This is a breaking change from today. When no version is passed in Helm will look at the tags and use semver to identify the latest version. This behavior must be preserved to follow the backwards compatibility rules.

Can you update the code to preserve the current behavior when there is no tag, digest, or version present?

mattfarina (changes requested):
Thanks for working on this. I look forward to the change landing.

We need to preserve backwards compatibility within minor versions of Helm. So, there is one thing that needs to change. See below...

pkg/downloader/chart_downloader.go
TerryHowe: The error will only be reached if neither digest, `--version` or a tag is not specified.

https://github.com/helm/helm/pull/12690/files#diff-2c9eb8ecd19796ab58f91f47766276c8d2a0c6732276fe1712d47db6d776cfe6R152

TerryHowe: The only resolution I can think of here is to use `latest` as the tag, but I'm not thrilled about that.

sabre1041: The [current functionality](https://github.com/helm/helm/blob/main/pkg/downloader/chart_downloader.go#L153) queries the list of available tags and then selects the most recent semver ordered version. The same logic should be able to be applied in this instance

TerryHowe: Done.

origxm:
Hey, any update on that issue?

@mattfarina

sabre1041:
@TerryHowe when attempting to install by digest, the following error is emitted

```
Error: object required
```

```
helm template helm-digest oci://<charrt>@sha256:<SHA>
```

TerryHowe:
> @TerryHowe when attempting to install by digest, the following error is emitted
> 
> ```
> Error: object required
> ```
> 
> Command used
> 
> ```
> helm template helm-digest oci://<charrt>@sha256:<SHA>
> ```
> 
> Other uses, such as including a tag, including a tag + version, version only and no tag or version work without issue

Fixed

TerryHowe:
```
Error: object required
```
Is just an ambiguous error message from oras-go v1 meaning the tag was required and not provided. The latest change uses the digest if the tag was not provided.

sabre1041:
> ```
> Error: object required
> ```
> 
> Is just an ambiguous error message from oras-go v1 meaning the tag was required and not provided. The latest change uses the digest if the tag was not provided.

Issue has been resolved with the latest update. Thanks!

Still one outstanding question and to circle back from my question some time ago. When a digest and version is supplied, what should occur. If both are included, digest always win

jimmyjones2:
Docker, buildah and CRI-O all ignore the tag if a digest is specified. Some discussion here: https://github.com/containers/buildah/issues/1407

TerryHowe:
> one outstanding question and to circle back from my question some time ago. When a digest and version is supplied, what should occur. If both are included, digest always win

I can't say how all registries will handle it, but...

In this PR, helm is going to validate that the tag returns the same digest here https://github.com/helm/helm/pull/12690/files#diff-2c9eb8ecd19796ab58f91f47766276c8d2a0c6732276fe1712d47db6d776cfe6R169-R176

sabre1041 (dismissed):
LGTM. Looking forward to seeing this feature becoming part of the project

TerryHowe:
Maybe tag "HasOneApproval"

pkg/downloader/chart_downloader.go
```diff
@@ -143,7 +143,39 @@ func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *proven
 
 func (c *ChartDownloader) getOciURI(ref, version string, u *url.URL) (*url.URL, error) {
 	var tag string
-	var err error
+
+	registryReference, err := registry.NewReference(u.Path)
```
gjenkins8, line 147: IMHO, we will need to document the below logic

TerryHowe: When you say document, do you mean comments or installation docs on https://helm.sh/ ?

gjenkins8: https://helm.sh/docs/topics/registries/ I think

The thought was to add to the examples e.g. installing by digest. And perhaps a "version specification" section which documents the `--version` and `:tag` and `@digest` options and precedence with respect to `helm pull`, `helm install`, etc.

(This PR can be merged without docs to clarify, just docs would make the contribution much more valuable IMHO)

TerryHowe: Sounds good. I'll put together a PR for the docs

TerryHowe: https://github.com/helm/helm-www/pull/1629

gjenkins8: Nice, thanks!

pkg/registry/reference.go
```diff
@@ -0,0 +1,78 @@
...
+package registry
+
+import (
+	"strings"
+
+	orasregistry "oras.land/oras-go/pkg/registry"
+)
+
+type Reference struct {
+	OrasReference orasregistry.Reference
```
gjenkins8, line 26: Apologies in advance for the slow drip of comments

Do you think it is worthwhile not exposing this Oras V1 type directly? Especially given the experience in https://github.com/helm/helm/pull/13382 ?

gjenkins8: (the suggestion is to make `NewReference` and internal type (not exposed to the public Helm API). And so it won't be subject to API compatibility guarantees once merged.

gjenkins8 (commented):
PR / changes here nearly LGTM for me. Just one comment/concern about exposing the Oras V1 type

And on the subject of Oras V1, do you think it would be better to aim to merge this PR or  https://github.com/helm/helm/pull/13382 first? I assume merging https://github.com/helm/helm/pull/13382 will require changes here, and vice versa. Asking, because I think worthy to push (both) through. But perhaps better to focus on one first.

(that said, I didn't review https://github.com/helm/helm/pull/13382 fully yet)

TerryHowe:
> And on the subject of Oras V1, do you think it would be better to aim to merge this PR or https://github.com/helm/helm/pull/13382 first?

I don't think the order will matter much. The digest change just added one mehtod to the ORAS wrapper and v2 has a method that would simplify that call.

pkg/registry/reference.go
TerryHowe: I made all that private to registry

gjenkins8 (approved):

TerryHowe:
Changes since last review are one commit https://github.com/helm/helm/pull/12690/commits/d2b94f62004e79864ec530989109cf0effd4aaae

sabre1041:
> Should we aim to backport this to helm v3? (add the "needs v3 fix" label and cherrypick to dev-v3 branch)

Yes, I believe that we should as the benefits can be realized prior to the release of Helm 4

pkg/registry/reference.go
```diff
@@ -0,0 +1,78 @@
...
+	// present in tags with underscores (_). To do this properly, we first
+	// need to identify a tag, and then pass it on to the reference parser
+	// NOTE: Passing immediately to the reference parser will fail since (+)
+	// signs are an invalid tag character, and simply replacing all plus (+)
+	// occurrences could invalidate other portions of the URI
+	lastIndex := strings.LastIndex(raw, "@")
+	if lastIndex >= 0 {
+		result.Digest = raw[(lastIndex + 1):]
+		raw = raw[:lastIndex]
+	}
```
mattfarina, line 51: There's a bug here, technically. You can have a URI with user/pass in it like "user:password@example.com/". For this case when there is no digest, it will be improperly parsed. This is a rare situation and I think it can be fixed in a follow-up.

mattfarina (approved):

PR #12691, opened 2024-01-08
Removes the hardcoded timeout (60 seconds) and replaces it with dynam…

opencmit2:
…ically calculated maximum timeout

This PR fixes ISSUE 'refs #12563 ' by utilizing the existing command line Timeout parameter instead of hardcoding timeouts for CRDs. This allows for more flexibility and easier configuration of timeouts.

Removes the hardcoded timeout (60 seconds) and replaces it with dynamically calculated maximum timeout.

**If applicable**:


pkg/action/install.go
```diff
@@ -63,6 +63,9 @@ const notesFileSuffix = "NOTES.txt"
 
 const defaultDirectoryPermission = 0755
 
+// Default timeout for CRD
+const defaultTimeout = 60 * time.Second
```
gjenkins8, line 67:
```suggestion
// Minimum timeout for CRD installation (retained for compatibility reasons)
const minimumCRDTimeout = 60 * time.Second
```

nit: it isn't really the "default", it is the minimum. And only for CRDs. So the name should reflect that

opencmit2:
@gjenkins8 
Corrected

opencmit2:
@gjenkins8 Could you please approve the CI

PR #10920, opened 2022-05-02
Verify generation in readiness checks

rohnyn:
Signed-off-by: James Oden <james.oden.dev@gmail.com>

**What this PR does / why we need it**: This patches a bug where helm preemptively marks resources as ready.  This is due to a race condition between the helm readiness check and the associated k8s controller updating the resource.  I've added generational validation to readiness checks for all applicable resources.

**Special notes for your reviewer**:

**If applicable**:
- [ ] this PR contains documentation
- [x] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

closes #10859
Associated k8s [issue](https://github.com/kubernetes/kubernetes/issues/109490#issuecomment-1101704320)
Thanks for the help!

PR #12698, opened 2024-01-09
Fix issues when verify generation readiness was merged

mattfarina:
CI, tests, and building failed after #10920 was merged. This change fixes the issues that were introduced.

**What this PR does / why we need it**: CI, building, testing, and linting are failing on master due to issues introduced in a PR where CI passed. This change fixes the codebase.

**Special notes for your reviewer**: The issue here is blocking the release of the RC for Helm 3.14.

**If applicable**:

### DIFF
--- a/pkg/kube/ready.go
+++ b/pkg/kube/ready.go
@@ -90,13 +90,6 @@ type ReadyChecker struct {
 // IsReady will fetch the latest state of the object from the server prior to
 // performing readiness checks, and it will return any error encountered.
 func (c *ReadyChecker) IsReady(ctx context.Context, v *resource.Info) (bool, error) {
-	var (
-		// This defaults to true, otherwise we get to a point where
-		// things will always return false unless one of the objects
-		// that manages pods has been hit
-		ok  = true
-		err error
-	)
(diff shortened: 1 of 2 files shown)


pkg/kube/ready.go
```diff
@@ -90,13 +90,6 @@ type ReadyChecker struct {
 // IsReady will fetch the latest state of the object from the server prior to
 // performing readiness checks, and it will return any error encountered.
 func (c *ReadyChecker) IsReady(ctx context.Context, v *resource.Info) (bool, error) {
-	var (
-		// This defaults to true, otherwise we get to a point where
-		// things will always return false unless one of the objects
-		// that manages pods has been hit
-		ok  = true
-		err error
-	)
```
scottrigby, line 99: Looking over 10920 I get why we need the rest of the fixes, but can you help clarify this part? Tempted to rubber stamp but just checking on this.

mattfarina: Those variables were used as part of the code that was removed. They are now unused variables so the Go compiler gives an area. The code that uses them is gone.

scottrigby (approved):
Lgtm. Thanks!

PR #12700, opened 2024-01-09
Include annotations in helm get metadata

### LINKED ISSUE #12686: Helm get metadata command - info omitted
Output of `helm version`: 3.13.3
Output of `kubectl version`: 1.27.4

Issue: 

We are using the annotations on the Chart.yaml structure and wanted to be able to retrieve those informations on the deployed chart with helm get metadata command. But it seems that the annotations field is not retrievable with that command. 

Is this normal behaviour and/or is there a way to retrieve chart annotation once deployed ? 

Thank you.

MichaelMorrisEst:
Closes #12686 

**If applicable**:

### DIFF
--- a/pkg/action/get_metadata.go
+++ b/pkg/action/get_metadata.go
@@ -28,14 +28,15 @@ type GetMetadata struct {
 }
 
 type Metadata struct {
-	Name       string `json:"name" yaml:"name"`
-	Chart      string `json:"chart" yaml:"chart"`
-	Version    string `json:"version" yaml:"version"`
-	AppVersion string `json:"appVersion" yaml:"appVersion"`
-	Namespace  string `json:"namespace" yaml:"namespace"`
-	Revision   int    `json:"revision" yaml:"revision"`
-	Status     string `json:"status" yaml:"status"`
-	DeployedAt string `json:"deployedAt" yaml:"deployedAt"`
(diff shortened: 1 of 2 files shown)


pkg/action/get_metadata.go
```diff
@@ -28,14 +28,15 @@ type GetMetadata struct {
...
-	DeployedAt string `json:"deployedAt" yaml:"deployedAt"`
+	Name        string            `json:"name" yaml:"name"`
+	Chart       string            `json:"chart" yaml:"chart"`
+	Version     string            `json:"version" yaml:"version"`
+	AppVersion  string            `json:"appVersion" yaml:"appVersion"`
+	Namespace   string            `json:"namespace" yaml:"namespace"`
+	Revision    int               `json:"revision" yaml:"revision"`
+	Status      string            `json:"status" yaml:"status"`
+	DeployedAt  string            `json:"deployedAt" yaml:"deployedAt"`
+	Annotations map[string]string `json:"annotations,omitempty" yaml:"annotations,omitempty"`
```
MissedTheMark, line 39: This outputs as `ANNOTATIONS: map[key0:val0 key1:val1 ...etc]`
Perhaps taking the same approach as `get_values.go` which outputs the map as yaml might be appropriate?

PR #12704, opened 2024-01-11
Update architecture detection method

weidongkl:
```
In some OS, obtaining the processor type from `uname - p` may fail.
For Golang programs, we should use `go env GOARCH` to obtain the architecture

=====================
$ go env GOARCH
amd64
$ uname -m
x86_64
$ uname -p
unknown
```

**If applicable**:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ GOBIN         = $(shell go env GOPATH)/bin
 endif
 GOX           = $(GOBIN)/gox
 GOIMPORTS     = $(GOBIN)/goimports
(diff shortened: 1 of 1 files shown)


mattfarina:
It looks like this fixes a possible bug. I just found that using the current uname approach returns a different value than expected in an environment. Not our CI system but outside it. Labeling as a bug and putting into the 3.14.1 release.

mattfarina (approved):
As this is an XS sized pull request, a single approval is all that's needed.

PR #12737, opened 2024-01-21
Correct incorrect error message of invalid tag

### LINKED ISSUE #12728: Correct incorrect error message of invalid tag.
when you run a helm push and provide a full OCI path like 
`helm push app-0.1.2-066a0ae.tgz oci://428571310775.dkr.ecr.us-east-1.amazonaws.com/cbdc-backend-helm:0.1.2-066a0ae
Error: invalid_reference: invalid tag`

Output of `helm version`:
Output of `kubectl version`:
N/A
AWS ECR.

I would like to see a more helpful output of something like version not needed in path to server. Better error reporting would have saved me a few hours of debuging only to find out I don't need to pass in a version to the [remote]

Rashmi-K-A:
Closes #12728 by adding a clearer error log that says `Version tag <VERSION> need not be passed for remote"`

See local run result:
Screenshot 2024-01-21 at 2 26 37 PM

Note for local development: The strings.SplitN(...) command works only with go version >=1.18, not sure if a backwards compatible strings.Split(...) is required instead.

**If applicable**:

### DIFF
--- a/pkg/uploader/chart_uploader.go
+++ b/pkg/uploader/chart_uploader.go
@@ -24,6 +24,7 @@ import (
 
 	"helm.sh/helm/v3/pkg/pusher"
 	"helm.sh/helm/v3/pkg/registry"
+	"strings"
 )
 
 // ChartUploader handles uploading a chart.
@@ -48,6 +49,12 @@ func (c *ChartUploader) UploadTo(ref, remote string) error {
 	if u.Scheme == "" {
 		return fmt.Errorf("scheme prefix missing from remote (e.g. \"%s://\")", registry.OCIScheme)
 	}
+
+	// See: https://github.com/helm/helm/issues/12728
+	path := strings.SplitN(u.Path, ":", 2)
+	if len(path) > 1 {
+		return fmt.Errorf("Version tag \"%s\" need not be passed for remote", path[1])
+	} 
 
 	p, err := c.Pushers.ByScheme(u.Scheme)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/uploader/chart_uploader.go
```diff
@@ -49,6 +50,12 @@ func (c *ChartUploader) UploadTo(ref, remote string) error {
 		return fmt.Errorf("scheme prefix missing from remote (e.g. \"%s://\")", registry.OCIScheme)
 	}
 
+	// See: https://github.com/helm/helm/issues/12728
+	path := strings.SplitN(u.Path, ":", 2)
+	if len(path) > 1 {
+		return fmt.Errorf("Version tag \"%s\" need not be passed for remote", path[1])
+	}
```
mattfarina, line 57: This check should be placed inside _ocipusher.go_. `UploadTo` has basic error handling for the URI to make sure it's valid and we have the ability to route it to a pusher. This logic is specific to OCI style things so it should go in that pusher.

mattfarina (changes requested):
Thanks for the pull request. The change needs to go in a different file. Where you put it is in the routing logic and this needs to go in a more specific place.

Rashmi-K-A:
Sure, I will make some changes this week

Rashmi-K-A:
Moved it, didnt catch error for parsing since a bad url Parse would have been caught at the chart_uploader level. Tested it out locally and it seems to work

Rashmi-K-A:
@mattfarina this one has been open for a while, can you take a look

PR #12751, opened 2024-01-26
Prevent error message involving CLI arguments in case there's a failure in verifySupported

krichter722:

"need" is a strong word. You might want it in order to improve feedback when there's a missing binary like `openssl` and users ought to be able to focus on the issue rather than be confused by the CLI parameter error message when the failure is compeletely unrelated to the CLI parameter.

Missing openssl is the case in most small Docker images, like Alpine.

./.

**If applicable**:

### DIFF
--- a/scripts/get-helm-3
+++ b/scripts/get-helm-3
@@ -25,6 +25,8 @@
 : ${HELM_INSTALL_DIR:="/usr/local/bin"}
 : ${GPG_PUBRING:="pubring.kbx"}
 
+EXIT_VERIFY_FAILED=2
+
 HAS_CURL="$(type "curl" &> /dev/null && echo true || echo false)"
 HAS_WGET="$(type "wget" &> /dev/null && echo true || echo false)"
 HAS_OPENSSL="$(type "openssl" &> /dev/null && echo true || echo false)"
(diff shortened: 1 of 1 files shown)


banjoh:
@krichter722 thanks for your contribution. My comments are based on my test below

```sh
$ docker run -it --rm alpine
/ # wget -q -O - https://raw.githubusercontent.com/helm/helm/main/scripts/get-helm-3 | sh; echo $?
In order to verify checksum, openssl must first be installed.
Please install openssl or set VERIFY_CHECKSUM=false in your environment.
Failed to install helm
1
/ # wget -q -O - https://raw.githubusercontent.com/krichter722/helm/avoid-confusing-helm-output-on-verify-supported-fail/scripts/get-helm-3 | sh; echo $?
In order to verify checksum, openssl must first be installed.
Please install openssl or set VERIFY_CHECKSUM=false in your environment.
The script failed the installation parameter validation
2
/ #
```

- This is more of my opinion here on the UX. I think the new message would mislead me more than the older one. `The script failed the installation parameter validation` message prompts me to find which parameter I may have provided that is invalid. As for the `openssl` warning, both scripts inform seem to say the same thing.
- Exit code 2
  - I believe this falls in the category of breaking an API? If some automation has logic tied to this error code, it will not catch this error when `2` is set.
  - `bash` uses this exit code to report errors in misuse of builtins. Whether people follow that or not, I can't tell :)

PR #12762, opened 2024-01-31
feat(helm): add --use-source-hooks flag to helm 'rollback'

### LINKED ISSUE #12757: Proposal: add --use-source-hooks flag to helm 'rollback'
When rolling back to a previous release, for example B -> A, helm only executes rollback hooks from the target release (A) and ignores any rollback hooks from current release (B). This makes it difficult to undo changes in the context of the current release before rolling back to the previous release. 

While the behavior makes sense for helm 'upgrade' (A->B), the behavior is less intuitive for helm 'rollback' (B->A: use hooks in A to undo changes made in B?). 

For backwards compatibility, I propose a new flag should be added to helm 'rollback' that lets the user specify from which release hooks should be run for rollback. By default, helm 'rollback' should run hooks from target release, but if the flag (--use-source-hooks) is present, helm 'rollback' should run hooks from the current release.  

Code-wise I think main change would only be in [rollback.go](https://github.com/helm/helm/blob/main/pkg/action/rollback.go).
A condition could be added that swaps out which release is used in execHook call, based on whether --use-source-hooks flag was used. Please correct me if I'm wrong! 

https://github.com/helm/helm/blob/e81f6140ddb22bc99a08f7409522a8dbe5338ee3/pkg/action/rollback.go#L177-L184

https://github.com/helm/helm/blob/e81f6140ddb22bc99a08f7409522a8dbe5338ee3/pkg/action/rollback.go#L244-L249

This could also serve as a solution for these issues:
- https://github.com/helm/helm/issues/10233
- https://github.com/helm/helm/issues/5825
- https://github.com/helm/helm/issues/3549

I look forward to your feedback :)

JPula:
Provides option to use rollback hooks from current release instead of target release when running helm 'rollback'

Added the flag to remain backwards compatible.
closes #12757

**If applicable**:

### DIFF
--- a/cmd/helm/rollback.go
+++ b/cmd/helm/rollback.go
@@ -80,6 +80,7 @@ func newRollbackCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 	f.BoolVar(&client.Recreate, "recreate-pods", false, "performs pods restart for the resource if applicable")
 	f.BoolVar(&client.Force, "force", false, "force resource update through delete/recreate if needed")
 	f.BoolVar(&client.DisableHooks, "no-hooks", false, "prevent hooks from running during rollback")
+	f.BoolVar(&client.UseSourceHooks, "use-source-hooks", false, "run hooks from current release during rollback")
 	f.DurationVar(&client.Timeout, "timeout", 300*time.Second, "time to wait for any individual Kubernetes operation (like Jobs for hooks)")
 	f.BoolVar(&client.Wait, "wait", false, "if set, will wait until all Pods, PVCs, Services, and minimum number of Pods of a Deployment, StatefulSet, or ReplicaSet are in a ready state before marking the release as successful. It will wait for as long as --timeout")
 	f.BoolVar(&client.WaitForJobs, "wait-for-jobs", false, "if set and --wait enabled, will wait until all Jobs have been completed before marking the release as successful. It will wait for as long as --timeout")
--- a/cmd/helm/rollback_test.go
+++ b/cmd/helm/rollback_test.go
@@ -62,6 +62,11 @@ func TestRollbackCmd(t *testing.T) {
 		golden: "output/rollback-wait-for-jobs.txt",
 		rels:   rels,
 	}, {
+		name:   "rollback a release with source hooks",
+		cmd:    "rollback funny-honey 1 --use-source-hooks",
+		golden: "output/rollback-use-source-hooks.txt",
+		rels:   rels,
+	}, {
 		name:   "rollback a release without revision",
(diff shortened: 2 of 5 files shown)


z4ce:
I'd like to see a test that would fail if the wrong hooks were run. Otherwise, looks good to me

JPula:
@z4ce Thanks for the feedback! 
I have added 2 tests to check whether source or target hooks were run during rollback. Cheers!

(I forgot to signoff my last commit 🤦‍ so I rebased my commits, apologies!)

z4ce (approved):

mattfarina (dismissed):

mattfarina:
@JPula thanks for the PR. Can you please fix the linting issue.

JPula:
@mattfarina I have fixed the linting issue

Gin-sen:
Hi, I was wondering how would that work with the "helm upgrade --atomic" command ?

Can we add the "--use-source-hooks" to it so the rollback triggered by a failed upgrade still trigger the sources hooks ?

Very glad to see this feature been addressed, thx a lot !

JPula:
> Hi, I was wondering how would that work with the "helm upgrade --atomic" command ?
> 
> Can we add the "--use-source-hooks" to it so the rollback triggered by a failed upgrade still trigger the sources hooks ?
> 
> Very glad to see this feature been addressed, thx a lot !
  
Adding the behavior for atomic upgrades wouldn't be difficult. Though the flag name should be more explicit if it is added to the 'helm upgrade' command (maybe "--use-source-hooks-on-rollback" ?).

That said, I think it would be better if you create a new ticket for this.
For now this PR only aims to add the flag to the helm 'rollback' command.

mluckam:
@JPula thanks for the PR, great feature add!  @z4ce and @mattfarina what are next steps to ensure this gets merged?  Looks like it is close.  Glad to help out however possible.

z4ce:
@gjenkins8 @scottrigby might be a fresh set of eyes for this one

mluckam:
@z4ce thanks for the quick response.   @mattfarina @gjenkins8 @scottrigby mind taking a look at this PR?  Linting issues were addressed as requested Feb 10, 2024.

balmeida-nokia:
@mattfarina @gjenkins8 @scottrigby, any news?

balmeida-nokia:
@mattfarina Is this on track for 3.19?
@JPula Seems to need a rebase.

balmeida-nokia:
@JPula Seems to need a rebase.

vfedorishchev:
Hi 👋 is there any help needed to make this PR ready for merge? I think it's a pretty good addition, allowing for auto-rollback execution for DB migrations.

PR #12769, opened 2024-02-01
fix(helm): pass down username/password CLI parameters to OCI registry clients

### LINKED ISSUE #12767: Helm template doesnt work with credentials for oci registries
Following command doesn't work without login to the registry 
`helm template <release-name> oci://ghcr.io/org-name/charts/chart-name --version 0.0.15 --username <user> --password <password>`

But if i do
`helm registry login 
oci://ghcr.io/org-name`
and then try to template the chart from the source, it works as expected

Output of `helm version`:

Output of `kubectl version`:
Client Version: v1.29.1
Server Version: v1.27.3

Cloud Provider/Platform (AKS, GKE, Minikube etc.): Kind, OpenShift

banjoh:

Ensures basic auth credentials passed in via helm CLI are passed down to clients connecting to the OCI registry

closes #12767, #13404

**If applicable**:
- [x] this PR contains unit tests (existing tests cover this)

**Test cases**

| Command | Pass/Fail | Comment |
| ---------- |:---------:| -----:|
| `helm install` | ✅ | |
| `helm upgrade` | ✅  | |
| `helm template` | ✅ | |
| `helm show chart/all/readme/values` | ✅  | |
| `helm pull` | ✅ | |
| `helm push` | ✅ | |
| `helm dependency build\|update` | ✅ | |
| `helm show all\|chart\|crds\|readme\|values` | ✅ | |
| `helm package` | ✅ | |


cmd/helm/root.go
```diff
@@ -292,10 +297,26 @@ func newDefaultRegistryClient(plainHTTP bool) (*registry.Client, error) {
...
+func newRegistryClientWithTLS(
+	certFile, keyFile, caFile string, insecureSkipTLSverify bool, username, password string,
+) (*registry.Client, error) {
+	tlsConf, err := tlsutil.NewClientTLS(certFile, keyFile, caFile, insecureSkipTLSverify)
+	if err != nil {
+		return nil, fmt.Errorf("can't create TLS config for client: %w", err)
+	}
+
 	// Create a new registry client
-	registryClient, err := registry.NewRegistryClientWithTLS(os.Stderr, certFile, keyFile, caFile, insecureSkipTLSverify,
```
banjoh, line 297: I found it unnecessary to expose a new public function to pass in the username/password. The `ClientOpt...` registry builder APIs are sufficient

banjoh:
Test cases

| Command | Pass/Fail | Comment |
| ---------- |:---------:| -----:|
| `helm install` | ✅ | |
| `helm upgrade` | ✅  | |
| `helm template` | ✅ | |
| `helm show chart/all/readme/values` | ✅  | |
| `helm pull` | ✅ | |
| `helm push` | ❌ | username and password options are absent |

banjoh:
This PR is ready for review

sabre1041:
@banjoh is the goal for `--username` and `--password` to be CLI options? If so, they are not currently exposed

banjoh:
> @banjoh is the goal for `--username` and `--password` to be CLI options? If so, they are not currently exposed

Yes, the goal is to allow a user to run one command which authenticates and does the requested action (install, pull, push...) instead of needing to first login then perform the needed action in 2 steps. Here is an install use case

#### Before this PR
```sh
helm registry login registry.com --username john@mail.com --password <SECRET>
helm install my-chart oci://registry.com/chart:0.0.1
```

#### After this PR
```sh
helm install my-chart oci://registry.com/chart:0.0.1 --username john@mail.com --password <SECRET>
```

Also note that, with http/https registry, running `helm install my-chart https://registry.com/chart:0.0.1 --username john@mail.com --password <SECRET>` works. This only fails when we have an OCI registry

> If so, they are not currently exposed

Do you mean that `--username` and `--password` options are not exposed entirely, or are there specific cases they aren't? The intention of this PR is to ensure the already exposed options that work with http/https also behave the same way as with OCI registries

sabre1041:
> > @banjoh is the goal for `--username` and `--password` to be CLI options?
> 
> Yes, the goal is to allow a user to run one command which authenticates and does the requested action (install, pull, push...) instead of needing to first login then perform the needed action in 2 steps. Here is an install use case
> 
> #### Before this PR
> ```shell
> helm registry login registry.com --username john@mail.com --password <SECRET>
> helm install my-chart oci://registry.com/chart:0.0.1
> ```
> 
> #### After this PR
> ```shell
> helm install my-chart oci://registry.com/chart:0.0.1 --username john@mail.com --password <SECRET>
> ```
> 
> Also note that, with http/https registry, running `helm install my-chart https://registry.com/chart:0.0.1 --username john@mail.com --password <SECRET>` works. This only fails when we have an OCI registry
> 
> > If so, they are not currently exposed
> 
> Do you mean that `--username` and `--password` options are not exposed entirely, or are there specific cases they aren't? The intention of this PR is to ensure the already exposed options that work with http/https also behave the same way as with OCI registries

The issue is that not all commands with OCI interaction have the fields exposed. For example, `helm push` do not expose a `--username` or `--password` option. In addition, did you have any thoughts as they relate to `helm package` and `helm dependency` subcommands? Would we assume that associated registries require valid authentication to be stored?

sabre1041 (dismissed):
This looks really good! This PR also supersedes #13096

pkg/action/install.go
```diff
@@ -759,9 +759,10 @@ func (c *ChartPathOptions) LocateChart(name string, settings *cli.EnvSettings) (
 		Getters: getter.All(settings),
 		Options: []getter.Option{
 			getter.WithPassCredentialsAll(c.PassCredentialsAll),
-			getter.WithTLSClientConfig(c.CertFile, c.KeyFile, c.CaFile),
+			getter.WithTLSClientConfig(c.CertFile, c.KeyFile, c.CAFile),
```
scottrigby, line 762: can't change this until Helm 4 (breaking change as this is a public package)

cmd/helm/dependency.go
```diff
@@ -120,3 +121,16 @@ func newDependencyListCmd(out io.Writer) *cobra.Command {
 	f.UintVar(&client.ColumnWidth, "max-col-width", 80, "maximum column width for output table")
 	return cmd
 }
+
+func addDependencySubcommandFlags(f *pflag.FlagSet, client *action.Dependency) {
+	f.BoolVar(&client.Verify, "verify", false, "verify the packages against signatures")
+	f.StringVar(&client.Keyring, "keyring", defaultKeyring(), "keyring containing public keys")
+	f.BoolVar(&client.SkipRefresh, "skip-refresh", false, "do not refresh the local repository cache")
+	f.StringVar(&client.Username, "username", "", "chart repository username where to locate the requested chart")
+	f.StringVar(&client.Password, "password", "", "chart repository password where to locate the requested chart")
```
scottrigby, line 130: what do you think about adding `--password-stdin` like docker does, for the same reason (setting password in a CLI command is insecure, especially in CI where many folks use Helm commands).

I know password opt is used already, so this may be a bigger issue outside the scope of your PR. May be good to approach all these consistently - eg, we have `passwordFromStdinOpt` only for repo add and registry login, etc.

cmd/helm/root.go
```diff
@@ -292,10 +297,26 @@ func newDefaultRegistryClient(plainHTTP bool) (*registry.Client, error) {
 	return registryClient, nil
 }
 
-func newRegistryClientWithTLS(certFile, keyFile, caFile string, insecureSkipTLSverify bool) (*registry.Client, error) {
+func newRegistryClientWithTLS(
```
scottrigby, line 300: I'm looking at the `NewRegistryClientWithTLS` function. It's only ever called from here. But since it's public we can't remove it until Helm 4. I wonder though, could we consolidate some of the logic in that function, in `newRegistryClientWithTLS`, and the logic wrapping the few places where these are called?

scottrigby (changes requested):
Thanks for this @banjoh 
One change request and a few comments

cmd/helm/root.go
banjoh: I think this can be done. We’d leave `NewRegistryClientWithTLS` as just a wrapper calling `newRegistryClientWithTLS` like you suggested

cmd/helm/dependency.go
banjoh: I think I can follow-up with another PR to limit the number of changes in this PR

cmd/helm/root.go
banjoh: @scottrigby now I remember why I could not DRY this code previously. I'd need to expose a new public function (i.e `NewRegistryClientWithTLSWithCreds`) in `pkg/registry/util` for it to be called in `cmd/helm` since we cannot add new parameters to `NewRegistryClientWithTLS`. Copying code over to `cmd/helm` seemed like the lesser evil

scottrigby: That seems reasonable to me 👍

cmd/helm/dependency_build.go
```diff
@@ -41,7 +41,7 @@ If no lock file is found, 'helm dependency build' will mirror the behavior
 of 'helm dependency update'.
 `
 
-func newDependencyBuildCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
+func newDependencyBuildCmd(_ *action.Configuration, out io.Writer) *cobra.Command {
```
joejulian, line 44: This is a private function, we can just remove cfg completely.

cmd/helm/dependency_update.go
```diff
@@ -43,7 +44,7 @@ in the Chart.yaml file, but (b) at the wrong version.
 `
 
 // newDependencyUpdateCmd creates a new dependency update command.
-func newDependencyUpdateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
+func newDependencyUpdateCmd(_ *action.Configuration, out io.Writer) *cobra.Command {
```
joejulian, line 47: Same here, private function
```suggestion
func newDependencyUpdateCmd(out io.Writer) *cobra.Command {
```

cmd/helm/package.go
```diff
@@ -47,7 +47,7 @@ If '--keyring' is not specified, Helm usually defaults to the public keyring
 unless your environment is otherwise configured.
 `
 
-func newPackageCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
+func newPackageCmd(_ *action.Configuration, out io.Writer) *cobra.Command {
```
joejulian, line 50: again, private function

joejulian (changes requested):

joejulian:
I'm concerned about this PR, and when I talked with @sabre1041 he agreed with my concern. There are so many variations on the implementation of the OCI spec, it's impossible for us to test that it works with all of them.

Can you think of some way to satisfy this concern? Even if you can demonstrate a manual proof that this doesn't break the most common cloud-company oci registries that would go a long way to satisfying my worry.

We've had this problem before so I'm a bit touchy.

banjoh:
> I'm concerned about this PR, and when I talked with @sabre1041 he agreed with my concern. There are so many variations on the implementation of the OCI spec, it's impossible for us to test that it works with all of them.
> 
> Can you think of some way to satisfy this concern? Even if you can demonstrate a manual proof that this doesn't break the most common cloud-company oci registries that would go a long way to satisfying my worry.
> 
> We've had this problem before so I'm a bit touchy.

That's a valid concern. I'll work on it.

sfc-gh-jlangefeld:
We just ran into this and are interested in this fix as well!

banjoh:
> We just ran into this and are interested in this fix as well!

I'll address the requested comments next week.

banjoh:
- docker registry - https://github.com/distribution/distribution (Tested already)
- AWS ECR - https://docs.aws.amazon.com/AmazonECR/latest/userguide/push-oci-artifact.html
- GCP artifact registry - https://cloud.google.com/artifact-registry/docs/helm/manage-charts
- Gitlab package registry - https://docs.gitlab.com/ee/user/packages/helm_repository/
- ArtifactHub - https://artifacthub.io/docs/topics/repositories/helm-charts/
- DockerHub - https://docs.docker.com/docker-hub/oci-artifacts/

@joejulian this is the list of OCI registries I'll test with. Feel free to add more to the list. I'm not sure I'll be able to cover all, especially if costs are involved

I also noticed https://helm.sh/docs/topics/registries/#use-hosted-registries list. Is this to say that we need to test against all these registries?

banjoh:
Using the same tests in the PR description, here are test runs from different registries

| Command | Pass/Fail | Comment |
| ---------- |:---------:| -----:|
| `docker registry` | ✅ | |

joejulian:
> Using the same tests in the PR description, here are test runs from different OCI registries
> Command 	Pass/Fail
> Docker registry 	✅
> AWS ECR 	✅
> GCP artifact registry 	✅
> Gitlab package registry 	✅
> DockerHub 	✅

This works for me. Thank you.

joejulian (dismissed):

sabre1041 (dismissed):
LGTM from my side as well

Thank you for working through not only the iterative development process, but testing the various configurations and various registries

cmd/helm/dependency.go
```diff
@@ -120,3 +121,16 @@ func newDependencyListCmd(out io.Writer) *cobra.Command {
...
+	f.StringVar(&client.CertFile, "cert-file", "", "identify HTTPS client using this SSL certificate file")
+	f.StringVar(&client.KeyFile, "key-file", "", "identify HTTPS client using this SSL key file")
+	f.BoolVar(&client.InsecureSkipTLSverify, "insecure-skip-tls-verify", false, "skip tls certificate checks for the chart download")
+	f.BoolVar(&client.PlainHTTP, "plain-http", false, "use insecure HTTP connections for the chart download")
+	f.StringVar(&client.CAFile, "ca-file", "", "verify certificates of HTTPS-enabled servers using this CA bundle")
```
scottrigby, line 135: Following up on https://github.com/helm/helm/pull/12769#pullrequestreview-2215880469 I suggest that we make all newly added fields consistent case, matching the existing field we can't change. So `CaFile` not `CAFile`
```suggestion
	f.StringVar(&client.CaFile, "ca-file", "", "verify certificates of HTTPS-enabled servers using this CA bundle")
```

cmd/helm/dependency_build.go
```diff
@@ -54,21 +54,27 @@ func newDependencyBuildCmd(cfg *action.Configuration, out io.Writer) *cobra.Comm
 			if len(args) > 0 {
 				chartpath = filepath.Clean(args[0])
 			}
+			registryClient, err := newRegistryClient(client.CertFile, client.KeyFile, client.CAFile,
```
scottrigby, line 57: see above:
```suggestion
			registryClient, err := newRegistryClient(client.CertFile, client.KeyFile, client.CaFile,
```

cmd/helm/dependency_update.go
```diff
@@ -57,13 +58,19 @@ func newDependencyUpdateCmd(cfg *action.Configuration, out io.Writer) *cobra.Com
 			if len(args) > 0 {
 				chartpath = filepath.Clean(args[0])
 			}
+			registryClient, err := newRegistryClient(client.CertFile, client.KeyFile, client.CAFile,
```
scottrigby, line 61: see above:
```suggestion
			registryClient, err := newRegistryClient(client.CertFile, client.KeyFile, client.CaFile,
```

cmd/helm/package.go
```diff
@@ -75,6 +75,12 @@ func newPackageCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 				return err
 			}
 
+			registryClient, err := newRegistryClient(client.CertFile, client.KeyFile, client.CAFile,
```
scottrigby, line 78: see above:
```suggestion
			registryClient, err := newRegistryClient(client.CertFile, client.KeyFile, client.CaFile,
```

cmd/helm/package.go
```diff
@@ -119,6 +125,13 @@ func newPackageCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 	f.StringVar(&client.AppVersion, "app-version", "", "set the appVersion on the chart to this version")
 	f.StringVarP(&client.Destination, "destination", "d", ".", "location to write the chart.")
 	f.BoolVarP(&client.DependencyUpdate, "dependency-update", "u", false, `update dependencies from "Chart.yaml" to dir "charts/" before packaging`)
+	f.StringVar(&client.Username, "username", "", "chart repository username where to locate the requested chart")
+	f.StringVar(&client.Password, "password", "", "chart repository password where to locate the requested chart")
+	f.StringVar(&client.CertFile, "cert-file", "", "identify HTTPS client using this SSL certificate file")
+	f.StringVar(&client.KeyFile, "key-file", "", "identify HTTPS client using this SSL key file")
+	f.BoolVar(&client.InsecureSkipTLSverify, "insecure-skip-tls-verify", false, "skip tls certificate checks for the chart download")
+	f.BoolVar(&client.PlainHTTP, "plain-http", false, "use insecure HTTP connections for the chart download")
+	f.StringVar(&client.CAFile, "ca-file", "", "verify certificates of HTTPS-enabled servers using this CA bundle")
```
scottrigby, line 134: see above:
```suggestion
	f.StringVar(&client.CaFile, "ca-file", "", "verify certificates of HTTPS-enabled servers using this CA bundle")
```

pkg/action/dependency.go
```diff
@@ -34,10 +34,17 @@ import (
...
-	ColumnWidth uint
+	Verify                bool
+	Keyring               string
+	SkipRefresh           bool
+	ColumnWidth           uint
+	Username              string
+	Password              string
+	CertFile              string
+	KeyFile               string
+	CAFile                string
```
scottrigby, line 45: see above:
```suggestion
	CaFile                string
```

pkg/action/package.go
```diff
@@ -44,8 +44,15 @@ type Package struct {
...
-	RepositoryConfig string
-	RepositoryCache  string
+	RepositoryConfig      string
+	RepositoryCache       string
+	PlainHTTP             bool
+	Username              string
+	Password              string
+	CertFile              string
+	KeyFile               string
+	CAFile                string
```
scottrigby, line 54: see above:
```suggestion
	CaFile                string
```

scottrigby (changes requested):
one minor change request for consistency. Following up on https://github.com/helm/helm/pull/12769#pullrequestreview-2215880469 I suggest that we make all newly added fields consistent case, matching the existing field we can't change. So `CaFile` not `CAFile`

joejulian (approved):

scottrigby (approved):
7a22dd28d13b02c6cd15bea19c35825f4f420caf commit squash looks good! All lgtm. Agree with sentiments above, really nice work. Especially manual testing across cloud providers 👏  Tests pass, and we have had 2 other approvers 😄

cmd/helm/root.go
EronWright: One reason is to encapsulate the call to `tlsutil.NewClientTLS` which is internal.

scottrigby:
We need to also bring this into the `dev-v3` branch. "needs v3 port" might be a better label.

cmd/helm/root.go
banjoh: I'll plan to add this in a follow-up PR I'm working on

PR #12775, opened 2024-02-03
Add support for optionally post-rendering hooks

### LINKED ISSUE #7891: Post-renderer not receiving chart hook manifests?

### LINKED ISSUE #10961: Support Run PostRenderer with Hooks Manifest.
Support  Run PostRenderer with Hooks Manifest. 

There is user case is when install helm chart into private k8s cluster which has no internet access,  after we push the images to local docker registry,  we need rewrite the repo of images within charts to local docker registry domain.
```
helm install chart  ./chart_path  -n namespace  --post-renderer=sed --post-renderer-args='s|docker.io\([^\"]*\)|192.168.1.1\1-aarch64|g'
```
but at current versions the hooks Manifest did not  run with PostRenderer.

So i push a  PR #10873  https://github.com/helm/helm/pull/10873 to fix this.  And open this issues for reviews  

many thanks.

Michael0x2a:
Fixes #7891; fixes #10961.

This PR adds support for optionally post-rendering hooks.

I opted to implement this by adding two new CLI flags: `--post-renderer-hooks` and `--post-renderer-hooks-args`. Why?

1. It retains full backwards compatibility: we default to not post-rendering hooks.

2. It gives users the ability to customize their main manifests and hooks in separate ways. One example of where this might be useful is if somebody is using kubectl-kustomize as their post-renderer to add a new resource to their chart. We would duplicate the new resource in the hooks manifest if we reused that post-renderer as-is, which does not seem like a good idea.

Two remaining possible issues with this implementation are:

1. It repeatedly re-execs the hooks post-renderer against each hook manifest, which may incur a perf penalty.
2. Some post-renderers may want to apply different transforms for different hooks and may find parsing the comment added to the hook manifests to be challenging.

But I think these issues are fine to ignore for now: users can use the Go library or write a custom plugin if they want finer-grained control.

**If applicable**:


busser:
I don't think the double run is a good idea. In one run the post-renderer doesn't know about hooks and in the second run it knows _only_ about hooks. There are likely use-cases where the post-renderer wants to know about all resources at once.

I think there is no need to duplicate the `--post-renderer` and `post-renderer-args` flags. Rather, it would be simpler to add a `--post-render-hooks` flag that adds hooks to the data sent to the post-renderer. If the post-renderer wants to adapt its behaviour depending on whether a given resource is a hook or not, it can check the resource for Helm hook annotations to make a decision.

PR #12783, opened 2024-02-07
Fix: uninstall action resource namespace

PaulBarrie:
closes#12766

At init configuration the kubeclient namespace is not set. When the uninstall action is triggered, if the release has been installed in a non-default namespace, the resource without any namespace provided (which are installed in the release's namespace) are not deleted since it tries to delete resources in the namespace "".

### DIFF
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -368,6 +368,7 @@ func (cfg *Configuration) recordRelease(r *release.Release) {
 // Init initializes the action configuration
 func (cfg *Configuration) Init(getter genericclioptions.RESTClientGetter, namespace, helmDriver string, log DebugLog) error {
 	kc := kube.New(getter)
+	kc.Namespace = namespace
 	kc.Log = log
 
 	lazyClient := &lazyClient{


rain-on:
I've run the provided fix and confirmed:
* the issue is able to be replicated prior to the fix being applied, and
* provided fix resolves the issue.

Effectively 2 namespace values are passed into the cfg.init:
1. Via the ClientGetter (aka environment/context defined)
2. Via the namespace parameter

The namespace parameter is used by the driver, while the KubeClient gets its namespace from the ClientGetter - so this change, to align the values seems reasonable.

All commands (bar "List") force the namespace parameter to be the same as the context/environment provided value - so _generally_ these two values _are_ the same.

There is some code in the Client which manages "use override namespace or value from environment", which maybe able to be removed if an "merged" config was created upfront (rather than managing 2 namespace variables through the codebase).

rain-on (approved):
LGTM

robertsirc:
This is done else where in the code this should be rejected.

mattfarina (changes requested):
This is a common ask but it's not so simple to set it here. The Initialized configuration can be used in multiple actions in a log running process. Such as a daemon that uses this as an SDK.

This initialization is for where the record of the release is stored instead of the manifests that are generated and uploaded. They can technically be in different places and the SDK allows for that.

You can see an example of how the actions put the namespace in over at https://github.com/helm/helm/blob/d1595e701f737f1fb367fb6b528c865e43a759ae/pkg/action/install.go#L334

I don't think this setup is ideal but we can't break the intent until Helm v4 where we can make some broader changes.

PR #12788, opened 2024-02-08
changed Itoa to Sprintf to allow int64 value preventing 2038 overflow

### LINKED ISSUE #12786: Helm getting into year 2038 issue
lbs.set("modifiedAt", strconv.Itoa(int(time.Now().Unix()))) this line is getting to the unix 2038 issue resulting failure of secret creation with below error.
Error: INSTALLATION FAILED: create: failed to create: Secret "sh.helm.release.v1.operator.v1" is invalid: metadata.labels: Invalid value: "-2147255893": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')
helm.go:84: [debug] Secret "sh.helm.release.v1.operator.v1" is invalid: metadata.labels: Invalid value: "-2147255893": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')

https://github.com/helm/helm/blob/e81f6140ddb22bc99a08f7409522a8dbe5338ee3/pkg/storage/driver/secrets.go#L180C2-L180C61

vtrenton:
Prevents an overflow when getting Unix time past the year 2038

I'm very open to recommendations on how to better fix this problem. We should probably replace Itoa() used for time anywhere to avoid this problem in the future :). I tested everything here: https://play.golang.com/p/y4HMAegdQ_r

closes #12786


ganesh-karthick:
Another option https://play.golang.com/p/rniO4YNgkxg   

```
// Format it as a string
	str := fmt.Sprintf("%v", t.Unix())
```

tleed5 (commented):
I'm not a contributor or maintainer on this project so take all of this with a grain of salt. 

Since there is quite a few places where this time -> string conversion occurs it might be better to encapsulate this into a method in `pkg/time` something like 
```
func (t Time) FormatAsString() string
```
That way all of the time formatting stuff is kept together and you won't have to import the `fmt` package everywhere.

Extra benefit is that you would be easily be able to add unit tests to cover the issue this is fixing 😄

gjenkins8 (commented):
can we include a test to show the issue is fixed, and prevent a regression in the future please

mattfarina (approved):
There is still one place that is no 2038 ready and this is the SQL driver that needs to be moved to a `BIGINT`. I think this can happen in a separate PR.

mattfarina:
Note, the force-push contained no changes. It was just used to trigger CI.

robertsirc (approved):
LGTM

PR #12802, opened 2024-02-16
Use canonical json-patch v5 import

skitt:

The canonical import for json-patch v5 is github.com/evanphx/json-patch/v5, as indicated in its go.mod and confirmed by its author in https://github.com/helm/helm/pull/12802#discussion_r1511339303

Using the v5-specific path also allows the v4 dependencies to be tracked separately, which is desirable for k/k (see https://github.com/kubernetes/kubernetes/pull/120327 for context).

**If applicable**:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,7 +13,7 @@ require (
 	github.com/containerd/containerd v1.7.12
 	github.com/cyphar/filepath-securejoin v0.2.4
 	github.com/distribution/distribution/v3 v3.0.0-20221208165359-362910506bc2
-	github.com/evanphx/json-patch v5.7.0+incompatible
+	github.com/evanphx/json-patch/v5 v5.9.0
 	github.com/foxcpp/go-mockdns v1.0.0
 	github.com/gobwas/glob v0.2.3
 	github.com/gofrs/flock v0.8.1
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -13,7 +13,6 @@ require (
 	github.com/containerd/containerd v1.7.12
 	github.com/cyphar/filepath-securejoin v0.2.4
 	github.com/distribution/distribution/v3 v3.0.0-20221208165359-362910506bc2
-	github.com/evanphx/json-patch v5.7.0+incompatible
```
mmorel-35, line 16: Why using gopkg.in URL instead of `github.com/evanphx/json-patch/v5` ?

Does it make any difference ?

skitt: I’m using the gopkg.in URL because that’s what’s documented... @evanphx would you care to weigh in?

evanphx: I forgot all about that part. I should remove it, use the direct GitHub url.

skitt: I’ve updated this PR to do that and opened https://github.com/evanphx/json-patch/pull/203 to fix the documentation.

evanphx: @skitt Thank you!

PR #12806, opened 2024-02-20
feat(helm): return all errors that occur during release uninstall

### LINKED ISSUE #12782: Helm uninstall does not return real error when it failed
We use helm client but not helm cli to do helm uninstall.

https://github.com/helm/helm/blob/main/pkg/action/uninstall.go#L125

This line will not return concrete error when helm uninstall failed.  Only helm release name is not enough

Could it return the ""errors" which return by "deleteRelease"

tleed5:
During the uninstall process all errors that occur will be bundled up into a new error struct `uninstallError` and be returned. This change only really affects the SDK as the CLI will still only surface the top level error as it currently does.

Closes https://github.com/helm/helm/issues/12782

This PR makes it so that errors that occur during uninstall aren't lost to the void when they method returns. In it's current state any errors that occur from the `deleteRelease` method are lost as they get replaced with a new error which gets returned to the command handler.

- I'm new to Golang and this codebase so any advice to improve this code would be awesome

**If applicable**:

### DIFF
--- a/pkg/action/uninstall.go
+++ b/pkg/action/uninstall.go
@@ -47,6 +47,11 @@ type Uninstall struct {
 	Description         string
 }
 
+type uninstallError struct {
+	error
+	uninstallErrors []error
+}
+
 // NewUninstall creates a new Uninstall object with the given configuration.
 func NewUninstall(cfg *Configuration) *Uninstall {
 	return &Uninstall{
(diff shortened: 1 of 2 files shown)


pkg/action/uninstall.go
```diff
@@ -122,7 +127,7 @@ func (u *Uninstall) Run(name string) (*release.UninstallReleaseResponse, error)
 	deletedResources, kept, errs := u.deleteRelease(rel)
 	if errs != nil {
 		u.cfg.Log("uninstall: Failed to delete release: %s", errs)
-		return nil, errors.Errorf("failed to delete release: %s", name)
+		return nil, uninstallError{error: errors.Errorf("failed to delete release: %s", name), uninstallErrors: errs}
```
tleed5, line 130: I wasn't 100% sure if changing the error message would be considered a breaking change which is why I left it like it was. If it isn't I think this should be updated to join the errors like the other places in this method do, that way all of the errors are surfaced to the CLI

Issue #11118, opened 2022-07-06
Unclear error about "table" destination


Woovie:
In multiple places, Helm uses the term "table" when describing the destination for a value. Example:

https://github.com/helm/helm/blob/db2485b20c8f53d802f4fc98c2f95e1fc0460d15/pkg/chartutil/coalesce.go#L222-L224

Error seen:

jbanasik@akira:~/k8s/loki$ helm install cassandra bitnami/cassandra --namespace loki -f cassandra.values.yaml
coalesce.go:223: warning: destination for cassandra.dbUser.existingSecret is a table. Ignoring non-table value ()
coalesce.go:223: warning: destination for cassandra.dbUser.existingSecret is a table. Ignoring non-table value ()

My code where this error is coming from:

```yaml
dbUser:
  user: cassandra
  forcePassword: false
  existingSecret:
    name: loki-cassandra-secret
    keyMapping:
      cassandra-password: cassandra-password
```

How do I, a consumer and user of Helm, know what this error means? What's a table? What's a non-table? None of this is defined, and even looking at the code I couldn't explain to you what a table is. I'd like to see something a bit more verbose to explain exactly what this means.

The container picks the value up fine ultimately, but I still didn't find this error to be anything that helped me diagnose my issue and only lead to more confusion.

bacongobbler:
> How do I, a consumer and user of Helm, know what this error means? What's a table? What's a non-table? 

From the doc comment:

> istable is a special-purpose function to see if the present thing matches the definition of a YAML table.

https://github.com/helm/helm/blob/fd9f4ccbedec345f75c0a9b58598ba87211e4721/pkg/chartutil/values.go#L168-L172

In technical terms, it checks to see if the type maps to a `map[string]interface{}`.

bacongobbler:
I understand where you're coming from. No offense taken. I'm just explaining the background context behind the code. `istable` was introduced to Helm 6 years ago in https://github.com/helm/helm/commit/e8109048a9e59b7b3617a58c42d55f8df8d2fb76.

If you would like to improve the error message, we welcome contributions via pull request. [The Contributing Guidelines](https://github.com/helm/helm/blob/main/CONTRIBUTING.md) will help you get started.

Woovie:
> > For example, calling it a map or mapping, which are YAML structures.
> 
> I like the idea of calling out the YAML name. It makes it easier for someone looking at the spec or reading some other docs on it.

[...]

PR #12812, opened 2024-02-22
fix: added extra conditional to warning about non-table values for expected tables

jessesimpson36:
Related to https://github.com/helm/helm/issues/11118

Related to https://github.com/bitnami/charts/issues/17465

Hello all, I came across this issue while working on a helm chart that I maintain.  Our helm chart makes use of the bitnami postgresql helm chart, which has some variables named `customRules` with default value `{}`.  Even though my helm chart does not override this value,  our helm installs get littered with these warnings:

```
coalesce.go:289: warning: destination for identity.postgresql.networkPolicy.ingressRules.readReplicasAccessOnlyFrom.customRules is a table. Ignoring non-table value ([])
...
``` 

And after some testing, I believe there is a faulty null check `val != nil`  which does not properly account for certain types.  I included a unit test which will demonstrate the warning getting placed improperly on a mock chart / mock values that is very similar to the postgresql helm chart that I use.

I also have a more practical integration test more specific to my use case  to demonstrate the error, which I will also provide, but in a separate branch since that testdata isn't suitable for this PR.

Branch with extra test:  https://github.com/jessesimpson36/helm/tree/jesse-extra-debugging

**If applicable**:


jessesimpson36:
I think there might be one more situation unaccounted for.  When `val` is equal to an empty string instead of an `interface{}`. If I get time, I'll add another commit.

tleed5 (commented):
I'm not a helm maintainer, i'm just reviewing PRs to get a better understanding of the codebase. 

To me the code looks good and I also gave it a test locally against a couple of charts and didn't have any issues with it.

jessesimpson36:
Hey all, I added a test to cover the empty-string scenario, and  I also broke out the boolean expression into functions since one of my colleagues was concerned about the casting causing a panic.

@tleed5 , although you're not a maintainer, I would appreciate your feedback on the latest commits.

tleed5 (commented):
I think that was a good call encapsulating that logic into separate helper functions like you did. Looks good to me

zzorica:
We also hit this when using bitnami/keycloak chart that has dependency on postgresql chart as well...

```
coalesce.go:237: warning: skipped value for keycloak.configuration: Not a table.
coalesce.go:289: warning: destination for keycloak.postgresql.networkPolicy.ingressRules.readReplicasAccessOnlyFrom.customRules is a table. Ignoring non-table value ([])
coalesce.go:289: warning: destination for keycloak.postgresql.networkPolicy.ingressRules.primaryAccessOnlyFrom.customRules is a table. Ignoring non-table value ([])
coalesce.go:289: warning: destination for keycloak.postgresql.networkPolicy.egressRules.customRules is a table. Ignoring non-table value ([])
```

jessesimpson36:
Some time has passed on this one, so I rebased it, and fixed up an edge-case.

jessesimpson36:
I also cleaned up the `jesse-extra-debugging` branch from the description.

pkg/chart/v2/util/values.go
```diff
@@ -183,6 +183,20 @@ func istable(v interface{}) bool {
...
+	table, ok := val.(map[string]interface{})
+	if !ok {
+		return false
+	}
+
+	return len(table) > 0
+}
+
+func isNonEmptyString(val interface{}) bool {
+	valueString := fmt.Sprintf("%s", val)
```
TerryHowe, line 196:
```suggestion
	valueString := fmt.Sprintf("%v", val)
```
I missed earlier discussion on this PR, seems like the format is wrong though.

jessesimpson36: I included this change. thanks for spotting!

jessesimpson36:
@TerryHowe , this is the PR I was referring to when I said in the dev meeting that enough time has passed that I don't feel strongly about it anymore. I'm thinking leaving this PR as open is just unnecessary noise.

I'm thinking I want to close this one, and if anyone wanted to push harder for this change to be included, they can copy the diff or re-implement it for helm v4. If no objections, I'll close it in 2 days.

TerryHowe:
I wasn't talking about any PR specifically.

pkg/chart/v2/util/coalesce.go
```diff
@@ -301,7 +301,7 @@ func coalesceTablesFullKey(printf printFn, dst, src map[string]interface{}, pref
 			} else {
 				printf("warning: cannot overwrite table with non table for %s (%v)", fullkey, val)
 			}
-		} else if istable(dv) && val != nil {
+		} else if istable(dv) && val != nil && isNonEmptyString(val) && isNonEmptyTable(dv) {
 			printf("warning: destination for %s is a table. Ignoring non-table value (%v)", fullkey, val)
```
TerryHowe: These prints should be slog calls.

Should we have a warning more specific to the condition?

jessesimpson36: > Should we have a warning more specific to the condition?

I think you're asking if the additional conditionals means we should change the text of the warning. I would say no. I think the wording of the warning matches the conditionals now, and actually did not match the conditional  prior to this PR, since the warning would be printed unnecessarily.

If we're adjusting the wording though, I'd suggest removing the use of the word "table" as helm end-users don't know what a table is.

```suggestion
			printf("warning: overridden setting %s is a table. Ignoring non-matching data type (%v)", fullkey, val)
```

jessesimpson36: > These prints should be slog calls.

I agree. It's a little strange how log.Printf is passed into this function. I'm also finding that there is now a `internal/chart/v3/util/coalesce.go` AND a `pkg/chart/v2/util/coalesce.go` (wasn't here the last time I updated this PR). Perhaps changing the logging + conditional should only be done in the v3 version.

TerryHowe: Matt made a copy of v2 in preparation for releasing charts v3 in helm v4. 

This change might best target charts v3 because changing to slog sounds like a breaking change.

jessesimpson36: I rebased on latest main and ported the change to charts v3 instead of v2. haven't gotten around to the slog change yet.

jessesimpson36: I changed the print statements to slog.Warn, but I preserved the ability to pass in your own printf functions. Which also means I am keeping the "warning" part of the string because some printf functions you pass in may not print the "warning" part as  `slog.Warn` does.

PR #12849, opened 2024-03-08
Added toYamlPretty template function to avoid errors with the yamllint tool in pipelines

efranco-technisys:
Currently the toYaml implementation does not indent list items, which causes issues with other tools in the YAML ecosystem such as linters and formatters. Rather than introduce a potentially breaking change to toYaml to support this, I've added a toYamlPretty function as suggested by somebody in the https://github.com/helm/helm/issues/12577.

Given a Golang map described as:

```
m := map[string]interface{} {
    "key": []int {1, 2, 3}
}
```

The toYaml function will produce:

```
key:
- 1
- 2
- 3
```
The toYamlPretty function will produce:

```
key:
  - 1
  - 2
  - 3
```
These are both syntactically correct, but other tools use an indented version which is where this function is useful.

Special notes for your reviewer:
I'm guessing this needs docs somewhere, but I'm not sure where to add them. Could somebody please point me in the right direction?

If applicable:

 

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -45,7 +45,7 @@ require (
 	k8s.io/klog/v2 v2.110.1
 	k8s.io/kubectl v0.29.0
 	oras.land/oras-go v1.2.4
-	sigs.k8s.io/yaml v1.3.0
+	sigs.k8s.io/yaml v1.4.0
 )
 
 require (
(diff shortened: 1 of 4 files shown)


scme0:
This is a duplicate of another, earlier PR: https://github.com/helm/helm/pull/12583

PR #12871, opened 2024-03-12
Enabling hide secrets on install and upgrade dry run

mattfarina:
This change adds a new flag to the install and upgrade commands in the Helm client and properties to the install and upgrade action. The new flag is --hide-secret and can only be used with the --dry-run flag.

The --dry-run flag is designed to send all chart rendered manifests to stdout so that they can be inspected.

When the --hide-secret flag is used they are absent from the output. This is because the output is checked for openapi validation and to ensure it is valid yaml. While openapi validation can be skipped, yaml validation is not easily skipped. So, secrets are removed from the output.

**What this PR does / why we need it**: #7275

**If applicable**:

### DIFF
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -103,7 +103,7 @@ type Configuration struct {
 // TODO: As part of the refactor the duplicate code in cmd/helm/template.go should be removed
 //
 //	This code has to do with writing files to disk.
-func (cfg *Configuration) renderResources(ch *chart.Chart, values chartutil.Values, releaseName, outputDir string, subNotes, useReleaseName, includeCrds bool, pr postrender.PostRenderer, interactWithRemote, enableDNS bool) ([]*release.Hook, *bytes.Buffer, string, error) {
+func (cfg *Configuration) renderResources(ch *chart.Chart, values chartutil.Values, releaseName, outputDir string, subNotes, useReleaseName, includeCrds bool, pr postrender.PostRenderer, interactWithRemote, enableDNS, hideSecret bool) ([]*release.Hook, *bytes.Buffer, string, error) {
 	hs := []*release.Hook{}
 	b := bytes.NewBuffer(nil)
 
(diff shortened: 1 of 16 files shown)


pkg/action/action.go
```diff
@@ -200,7 +200,11 @@ func (cfg *Configuration) renderResources(ch *chart.Chart, values chartutil.Valu
 
 	for _, m := range manifests {
 		if outputDir == "" {
-			fmt.Fprintf(b, "---\n# Source: %s\n%s\n", m.Name, m.Content)
+			if hideSecret && m.Head.Kind == "Secret" && m.Head.Version == "v1" {
```
joejulian, line 203: Should we be specific about the group, here? Would it be a surprise to have a Secret CRD and have its output suppressed?

pkg/action/action.go
```diff
@@ -200,7 +200,11 @@ func (cfg *Configuration) renderResources(ch *chart.Chart, values chartutil.Valu
...
+				cfg.Log("Kubernetes Secret %q hidden from output", m.Name)
```
joejulian, line 204: Should this be in the form of a comment to keep the yaml valid?

pkg/action/action.go
mattfarina: The `Version` property holds the group and version. For example, for deployments it's `apps/v1`. The `Head` is a `SimpleHead` type that has some information in a simplified form. This should only catch the core Secret.

The output is completely suppressed because it's validated for yaml and openapi schema. It's hard to change the output to something like `[HIDDEN]` which would make it fail both of those. Someone had to opt-in for it to be hidden.

joejulian (approved):
works on my machine

peterwwillis:
Is it possible to add the same option for the non-dry-run modes? IIRC, the secret values are still shown when you install/upgrade (which I expect, but it would be convenient to hide that information in output; currently if you want to watch your helm upgrade in CI, but don't want secrets in logs, you basically have to invent a weird program that will read a file or pipe unbuffered and filter out the secret values while still outputting all the other lines in real time)

aviadshimoni:
hi, updating docs:
https://github.com/helm/helm-www/pull/1608

PR #12876, opened 2024-03-13
Allow install, update action to adopt existing resources (sdk only)

manno:
We use the Helm SDK in a GitOps operator and need more control over existing resources. We considered patching them with the Helm labels and annotations, but that would mean another update request per resource. Furthermore it seems wrong to apply Helm labels to resources, before the chart has actually been applied to them.

The PR introduces a new option `TakeOwnership`. If `TakeOwnership` is set, existing resources will be adopted, overwritten, etc. when installing or updating. Helm will not check for labels.

If it is not set, Helm behaves like before: adoption is only possible if they existing resources have the right labels
(`managed-by`) and annotations (`release-name`, ...).

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -105,7 +105,9 @@ type Install struct {
 	// Used by helm template to add the release as part of OutputDir path
 	// OutputDir/<ReleaseName>
 	UseReleaseName bool
-	PostRenderer   postrender.PostRenderer
+	// TakeOwnership will ignore the check for helm annotations and take ownership of the resources.
+	TakeOwnership bool
+	PostRenderer  postrender.PostRenderer
 	// Lock to control raceconditions when the process receives a SIGTERM
 	Lock sync.Mutex
 }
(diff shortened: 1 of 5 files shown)


mattfarina (approved):

scottrigby (approved):
Nice work @manno 👍 

This could be a useful feature to bring into the CLI install and upgrade CLI commands too at some point.

atsu85:
@manno, thanks for implementing this feature!

PR description:
> If `TakeOwnership` is set, existing resources will be adopted, **overwritten**, etc. when installing or updating.

Based on this i would assume that the existing resource should get overwritten/updated based on values and chart templates fused for the Helm relase, but it seems that it is not the case? As i understand with `TakeOwnership` helm install/upgrade won't fail if helm release name and namespace don't match what's present on existing resource annotations, but the existing resource **won't get updated by helm install/upgrade? That doesn't seem to match what you wrote in the PR description!?**

If i understand correctly then `TakeOwnership` works basically the same way as manually adding release name/namespace annotations (or changing them in case of taking ownership of k8s resources from other helm release) before running helm install/upgrade with new release name?

I'm wondering if there is a way to make helm actually update the existing resource (without removing and re-creating it) with new content instead of silently ignoring changes that should be applied based on values and chart templates?

atsu85:
hi @manno, could you share your ideas/thoughts/insights about [my question/comment](https://github.com/helm/helm/pull/12876#issuecomment-2619259876)?

manno:
> If i understand correctly then TakeOwnership works basically the same way as manually adding release name/namespace annotations (or changing them in case of taking ownership of k8s resources from other helm release) before running helm install/upgrade with new release name?

Yes, TakeOwnership only ignores the existing labels. Updating the resource is still handled by Helm as usual. This PR didn't change how Helm updates, it just allowed Helm to update existing resources.

We found out that updating (three way merge) doesn't work for custom resources, see https://github.com/helm/helm/pull/13439

Issue #12741, opened 2024-01-23
Deleting a value is not supported in dependencies


Thorndike:
Output of `helm version`:
v3.14.0
Output of `kubectl version`:
v1.20.5
Cloud Provider/Platform (AKS, GKE, Minikube etc.): 
vanilla Kubernetes

I found multiple descriptions on how to delete a value in a helm configuration by overriding it with null. This works fine if I use a helm chart directly. In the case that I use a helm chart as a dependency this no longer gives the desired result. I can override the value with different values but no longer delete it by overriding it with null.

gjenkins8:
yes, unfortunately this is a bug. I have been trying to find time for the last few weeks to track it down.

rptaylor:
This is yet another duplicate of https://github.com/helm/helm/issues/12488

mbaldessari:
I am hitting some variant of this but not exactly like #12488 (which is slightly different). The problem I am seeing is that keys in subcharts are correctly removed, but not in subsubcharts (my use case is using the gitea helm chart as a dependency. gitea brings in a few other subcharts with it). My full repro with docs is here: https://github.com/mbaldessari/helm-issue/blob/subsubchart/

zsolt-marta-bitrise:
Seeing something similar here https://github.com/helm/helm/issues/12991. I added a minimal example.

Works if we set the dependent object to empty string `""` instead of `null`

pierluigilenoci:
Can #12879 or #13654 solve this?

Issue #12637, opened 2023-12-13
Nested `null` values do not remove keys from sub-charts


rodrigoechaide:
Nested `null` values do not remove keys from sub-charts.

I have the following setting in a sub-chart:

```
fluentd:
  livenessProbe:
    httpGet: null
    tcpSocket:
      port: 24224
    initialDelaySeconds: 20
  readinessProbe:
    httpGet: null
    tcpSocket:
      port: 24224
    initialDelaySeconds: 20
```

After running the helm template command, I still have the `httpGet` default values for the readiness and liveness probes. I want to remove them because otherwise, Kubernetes will reject the installation of the chart. It is not possible to configure more than one readiness and liveness probes kind simultaneously.

 
Output of `helm version`:

```
version.BuildInfo{Version:"v3.13.2", GitCommit:"2a2fb3b98829f1e0be6fb18af2f6599e0f4e8243", GitTreeState:"clean", GoVersion:"go1.21.4"}
```

Output of `kubectl version`:

```
Client Version: v1.25.4
Kustomize Version: v4.5.7
```

The following issues are related:

* https://github.com/helm/helm/issues/9136
* https://github.com/helm/helm/issues/5184

I don't know why they are closed if the problem still persists.

francostellari:
I observed the same issue in `v3.13.2`.

My `Chart.yaml` contains:

```yaml
dependencies:
- name: postgresql
  version: "13.1.5"
  repository: "oci://registry-1.docker.io/bitnamicharts"
```

My parent `values.yaml` contains:

```yaml
postgresql:
  primary:
    extendedConfiguration: "max_connections=1000"
    priorityClassName: system-node-critical
    podSecurityContext:
      fsGroup: null
      seccompProfile:
        type: RuntimeDefault
    containerSecurityContext:
      runAsUser: null
      allowPrivilegeEscalation: false
      runAsNonRoot: true
      seccompProfile:
        type: RuntimeDefault
      capabilities:
        drop: [ "ALL" ]
  volumePermissions:
    enabled: false
  shmVolume:
    enabled: false
```

Yet the yaml generated with a `--dry-run` still contains the `fsGroup: 1001`:

```yaml
      securityContext:
        fsGroup: 1001 # <<<<<<<<<<
        seccompProfile:
          type: RuntimeDefault
```

and

```yaml
            runAsGroup: 0
            runAsNonRoot: true
            runAsUser: 1001 # <<<<<<<<
            seccompProfile:
[...]

Issue #12594, opened 2023-11-24
Deleting a default key does not work as expected


pmeier:
We are experiencing an issue with [deleting a default key](https://helm.sh/docs/chart_template_guide/values_files/#deleting-a-default-key):

```yaml
# Chart.yaml
apiVersion: v2
name: seccompProfile-debug
version: "0.0.1"

dependencies:
  - name: postgresql
    version: 13.0.0
    repository: https://charts.bitnami.com/bitnami
```

```yaml
# values-env.yaml
postgresql:
  primary:
    containerSecurityContext:
      # only used for diagnostics in this issue
      runAsUser: "values-env.yaml"
      seccompProfile: null
```

The expected result is that the `seccompProfile` key is not present in the rendered template. However this is not the case in this setting:

### `values.yaml` file does not exist

```bash
$ helm template . -f values-env.yaml  | grep -A20 -e '- name: postgresql' | grep -A4 'runAsUser:'
            runAsUser: values-env.yaml
            seccompProfile:
              type: RuntimeDefault
          env:
            - name: BITNAMI_DEBUG
```

- `seccompProfile: null` is ignored and filled with the default instead

### `values.yaml` exists with `seccompProfile: null`

```yaml
# values.yaml
postgresql:
  primary:
    containerSecurityContext:
      runAsUser: "values.yaml"
      seccompProfile: null
```

```bash
helm template . -f values-env.yaml  | grep -A20 -e '- name: postgresql' | grep -A4 'runAsUser:'
            runAsUser: values-env.yaml
          env:
            - name: BITNAMI_DEBUG
              value: "false"
            - name: POSTGRESQL_PORT_NUMBER
```

- Overwrite for `runAsUser` worked since we get `values-env.yaml`
- `seccompProfile: null` was respected, since the key is no longer present in the output
- This is our current workaround

### `values.yaml` exists with no `seccompProfile` set

```yaml
# values.yaml
postgresql:
  primary:
    containerSecurityContext:
      runAsUser: "values.yaml"
```

```bash
$ helm template . -f values-env.yaml  | grep -A20 -e '- name: postgresql' | grep -A4 'runAsUser:'
            runAsUser: values-env.yaml
            seccompProfile:
              type: RuntimeDefault
          env:
            - name: BITNAMI_DEBUG
```

- `seccompProfile: null` is ignored and filled with the default instead

### `seccompProfile` key removed from `values-env.yaml` and `values.yaml` exists with `seccompProfile: null` set

```yaml
[...]

Issue #12522, opened 2023-10-25
Can't override value in subchart


o4rz3l:
Output of `helm version`: 3.13.1

Output of `kubectl version`: v1.25.2

Hi, based on the changelog and on some issues (e.g. https://github.com/helm/helm/pull/11440), it should be possible, to remove a value from a subchart by using `value: null`. Unfortunately, this does not work in my case.

Here are some some easy steps to reproduce my issue:
1 `helm create umbrella-chart`
2. remove all templates and the clear values.yaml
3. add the prometheus dependency to the Chart.yaml
  ```
dependencies:
  - name: prometheus
    version: 25.2.0
    repository: https://prometheus-community.github.io/helm-charts
  ```
4. `helm dep update`
5. Add this to values.yaml
  ```
  prometheus:
    server:
      securityContext: null
  ```
6. Run `helm template foo -s charts/prometheus/templates/deploy.yaml . -f values.yaml`

**Expected Result**
The block 
```      
      securityContext:
        fsGroup: 65534
        runAsGroup: 65534
        runAsNonRoot: true
        runAsUser: 65534
``` 
should not be in the deploy.yaml

**Actual Result**
The block `securityContext` with it's subkeys is in the deploy.yaml. No way to remove it.

bt909:
I experience the same issue, but if I put my values to override in the values.yaml of my umbrella chart directly and then don't use -f "file" I got my values of my sub chart overwritten. The problem seems to be related to use additional files via -f.

rptaylor:
This is another duplicate of https://github.com/helm/helm/issues/12488

PR #12879, opened 2024-03-14
bugfix: Override subcharts with null values

### LINKED ISSUE #12469: subchart override with --set not taken into account
Output of `helm version`:
```
version.BuildInfo{Version:"v3.13.0", GitCommit:"825e86f6a7a38cef1112bfa606e4127a706749b1", GitTreeState:"clean", GoVersion:"go1.20.8"}
```

Output of `kubectl version`:
```
Client Version: v1.24.13
Kustomize Version: v4.5.4
Server Version: v1.24.13
```

Cloud Provider/Platform (AKS, GKE, Minikube etc.): GKE / On Premise

###########################################
###########################################
can't override subchart values with "--set"
It seems it only works with values override from file directly

let say I have a chart called subchart with these values

```
strategy:
  rollingUpdate:
    maxSurge: 25%
    maxUnavailable: 10%
  type: RollingUpdate
```
In one of my other chart, I want to use this subchart as a dependency, but I want to have a Recreate strategy instead of the RollingUpdate one:

so I'm doing this:
```
subchart:
  strategy:
    rollingUpdate: null
    type: Recreate
```

```
helm template --release-name demo path/to/chart --output-dir .
```

My Deployment has 
```
spec:
  strategy:
    type: Recreate
```

But if I try with --set flag instead of null in the values file
```
helm template --release-name demo path/to/chart --set subchart.strategy.rollingUpdate=null --output-dir .
```
It doesn't work In the end I have 

```
spec:
  strategy:
    rollingUpdate:
      maxSurge: 25%
      maxUnavailable: 10%
    type: Recreate
```

### LINKED ISSUE #12488: subchart keys can not be removed by parent chart in v3.13.* when --values (-f) flag is used
Output of `helm version`:  `version.BuildInfo{Version:"v3.13.1", GitCommit:"3547a4b5bf5edb5478ce352e18858d8a552a4110", GitTreeState:"clean", GoVersion:"go1.20.8"}`

Output of `kubectl version`:   N/A, reproducible without k8s

Cloud Provider/Platform (AKS, GKE, Minikube etc.):   N/A

First off I am very excited to have a fix for https://github.com/helm/helm/issues/9027 , thank you for working on this!!
I seem to have found a peculiar bug though.
Following closely the reproduction steps from https://github.com/helm/helm/issues/9027 , I set up an environment using the same tarball:
```
$ mkdir demo
$ cd demo
$ wget https://github.com/helm/helm/files/5549239/chart.tar.gz
$ tar xf chart.tar.gz
$ cat Chart.yaml 
apiVersion: v2
name: demo
version: 0.1.0
dependencies:
  - name: subchart
    version: 0.1.0
$ file charts/subchart/templates/._values.yaml           #  ???? 
charts/subchart/templates/._values.yaml: AppleDouble encoded Macintosh file
$ rm charts/subchart/templates/._values.yaml
```

Confirm that the original issue is fixed - subchart.hash.key1 and key2 are removed since they are set to null in values.yaml
```
$ ~/helm-v3.13.1/helm template .
---
global:
  hash:
    key3: 13
    key4: 4
    key5: 5
    key6: 6
hash:
  key3: 13
  key4: 4
  key5: 5
  key6: 6
```

But after hours of banging my head against the wall I eventually found that the strange issue I was seeing is apparently triggered when you use the `--values` command line option.  Even though ./values.yaml is the real source of values being used by the above command, if you specify it with `--values values.yaml` the behaviour changes!
```
$ ~/helm-v3.13.1/helm template . --values values.yaml 
---
global:
  hash:
    key3: 13
    key4: 4
    key5: 5
    key6: 6
hash:
  key1: 1
  key2: 2
  key3: 13
  key4: 4
  key5: 5
  key6: 6
```

Same thing if you specify an absolute path to the values.yaml file:
```
$ ~/helm-v3.13.1/helm template .   --values /home/fedora/demo/values.yaml
```
 or copy it somewhere else:
```
$ cp values.yaml  /tmp/whatever.yaml 
$ ~/helm-v3.13.1/helm template .  --values /tmp/whatever.yaml
```

I found the same result with 3.13.0.

In v3.12.3 the behaviour is different but if anything more confusing:
```
$ ~/helm-v3.12.3/helm  template .
---
global:
  hash:
    key3: 13
    key4: 4
    key5: 5
    key6: 6
hash:
  key1: 1
  key2: 2
  key3: 13
  key4: 4
  key5: 5
  key6: 6
$ ~/helm-v3.12.3/helm template . --values values.yaml
---
global:
  hash:
    key1: 1
    key2: 2
    key3: 13
    key4: 4
    key5: 5
    key6: 6
hash:
  key3: 13
  key4: 4
  key5: 5
  key6: 6
```

Anyway surely the location of the values file and the fact that I have specified it with --values should not make any difference - only the content of that file should affect the outcome of the templating.  I thought values (and values files) specified on the command line should still be the highest priority?

ryanhockstad:
This PR closes #12469  and closes #12488

Helm should allow users to not only override default values, but also completely remove any default values by setting a config to `null`. 

This works fine for regular charts, but default values within sub-charts cannot be `null`-ed.  The linked issue has a good example of this created by user "naemono." 

The reason this issue is happening is because the `coalesce` function goes over sub-chart values that are defined in a values file or with a `--set` flag twice due to the logic [here](https://github.com/helm/helm/blob/d37e2e9097f9715d4d184b2e3cc313588f106030/pkg/chartutil/coalesce.go#L100-L101).  `merge` is always set to `false` in this context, and the first time `coalesce` gets called, the null value gets removed during the `coalesceTablesFullKey` function [here.](https://github.com/helm/helm/blob/d37e2e9097f9715d4d184b2e3cc313588f106030/pkg/chartutil/coalesce.go#L279) This is fine for regular chart values, but for sub-chart values, the `coalesceDeps` function runs the logic again for every value in the subChart, so we end right back to the spot in the previous link.  But the config option that was explicitly set to null was already removed, so we end up at [this point](https://github.com/helm/helm/blob/d37e2e9097f9715d4d184b2e3cc313588f106030/pkg/chartutil/coalesce.go#L281) two lines down, which overwrites the user defined `null` config the sub-chart default value.

This problem only occurs because when the `coalesce` function is run while `merge` is `false`.  

To prevent this bug from removing sub-chart config values that are explicitly set to `null` and therefore getting overwritten when `coalesceDeps` runs through the sub-chart values a second time, I added logic to check if the key was the name of a sub-chart, and if so setting `merge` to `false` so the null config option is not deleted.
 

Follow-up if this PR is accepted may be to add a unit-test to confirm sub-chart nulls aren't overwritten.

### DIFF
--- a/pkg/chartutil/coalesce.go
+++ b/pkg/chartutil/coalesce.go
@@ -237,6 +237,9 @@ func coalesceValues(printf printFn, c *chart.Chart, v map[string]interface{}, pr
 						printf("warning: skipped value for %s.%s: Not a table.", subPrefix, key)
 					}
 				} else {
+					// If the key is a child chart, coalesce tables with Merge set to true
+					merge := childChartMergeTrue(c, key, merge)
+
 					// Because v has higher precedence than nv, dest values override src
 					// values.
 					coalesceTablesFullKey(printf, dest, src, concatPrefix(subPrefix, key), merge)
@@ -247,6 +250,15 @@ func coalesceValues(printf printFn, c *chart.Chart, v map[string]interface{}, pr
 			v[key] = val
 		}
 	}
+}
+
+func childChartMergeTrue(chrt *chart.Chart, key string, merge bool) bool {
+	for _, subchart := range chrt.Dependencies() {
+		if subchart.Name() == key {
+			return true
+		}
+	}
+	return merge
 }
 
 // CoalesceTables merges a source map into a destination map.
--- a/pkg/chartutil/coalesce_test.go
+++ b/pkg/chartutil/coalesce_test.go
@@ -44,6 +44,7 @@ global:
     boat: true
 
 pequod:
+  boat: null
   global:
     name: Stinky
     harpooner: Tashtego
@@ -55,7 +56,7 @@ pequod:
     boat: null
     nested:
       foo: true
-      bar: null
+      boat: null
 `)
 
 func withDeps(c *chart.Chart, deps ...*chart.Chart) *chart.Chart {
@@ -82,6 +83,13 @@ func TestCoalesceValues(t *testing.T) {
 			"global": map[string]interface{}{
 				"nested2": map[string]interface{}{"l0": "moby"},
 			},
+			"pequod": map[string]interface{}{
+				"boat": "maybe",
+				"ahab": map[string]interface{}{
+					"boat":   "maybe",
+					"nested": map[string]interface{}{"boat": "maybe"},
+				},
+			},
 		},
 	},
 		withDeps(&chart.Chart{
@@ -91,6 +99,11 @@ func TestCoalesceValues(t *testing.T) {
 				"scope": "pequod",
 				"global": map[string]interface{}{
 					"nested2": map[string]interface{}{"l1": "pequod"},
+				},
+				"boat": false,
+				"ahab": map[string]interface{}{
+					"boat":   false,
+					"nested": map[string]interface{}{"boat": false},
 				},
 			},
 		},
@@ -104,7 +117,7 @@ func TestCoalesceValues(t *testing.T) {
 					"scope":  "ahab",
 					"name":   "ahab",
 					"boat":   true,
-					"nested": map[string]interface{}{"foo": false, "bar": true},
+					"nested": map[string]interface{}{"foo": false, "boat": true},
 				},
 			},
 		),
@@ -200,13 +213,18 @@ func TestCoalesceValues(t *testing.T) {
 		t.Error("Expected nested boat key to be removed, still present")
 	}
 
-	subchart := v["pequod"].(map[string]interface{})["ahab"].(map[string]interface{})
+	subchart := v["pequod"].(map[string]interface{})
 	if _, ok := subchart["boat"]; ok {
 		t.Error("Expected subchart boat key to be removed, still present")
 	}
 
-	if _, ok := subchart["nested"].(map[string]interface{})["bar"]; ok {
-		t.Error("Expected subchart nested bar key to be removed, still present")
+	subsubchart := subchart["ahab"].(map[string]interface{})
+	if _, ok := subsubchart["boat"]; ok {
+		t.Error("Expected sub-subchart ahab boat key to be removed, still present")
+	}
+
+	if _, ok := subsubchart["nested"].(map[string]interface{})["boat"]; ok {
+		t.Error("Expected sub-subchart nested boat key to be removed, still present")
 	}
 
 	// CoalesceValues should not mutate the passed arguments
--- a/pkg/chartutil/coalesce_test.go
+++ b/pkg/chartutil/coalesce_test.go
@@ -57,6 +57,7 @@ pequod:
     nested:
       foo: true
       boat: null
+    object: null
 `)
 
 func withDeps(c *chart.Chart, deps ...*chart.Chart) *chart.Chart {
@@ -118,6 +119,7 @@ func TestCoalesceValues(t *testing.T) {
 					"name":   "ahab",
 					"boat":   true,
 					"nested": map[string]interface{}{"foo": false, "boat": true},
+					"object": map[string]interface{}{"foo": "bar"},
 				},
 			},
 		),
@@ -225,6 +227,10 @@ func TestCoalesceValues(t *testing.T) {
 
 	if _, ok := subsubchart["nested"].(map[string]interface{})["boat"]; ok {
 		t.Error("Expected sub-subchart nested boat key to be removed, still present")
+	}
+
+	if _, ok := subsubchart["object"]; ok {
+		t.Error("Expected sub-subchart object map to be removed, still present")
 	}
 
 	// CoalesceValues should not mutate the passed arguments
--- a/pkg/chartutil/coalesce_test.go
+++ b/pkg/chartutil/coalesce_test.go
@@ -51,6 +51,7 @@ pequod:
     nested:
       boat: false
       sail: true
+      foo2: null
   ahab:
     scope: whale
     boat: null
@@ -112,7 +113,7 @@ func TestCoalesceValues(t *testing.T) {
 				Metadata: &chart.Metadata{Name: "ahab"},
 				Values: map[string]interface{}{
 					"global": map[string]interface{}{
-						"nested":  map[string]interface{}{"foo": "bar"},
+						"nested":  map[string]interface{}{"foo": "bar", "foo2": "bar2"},
 						"nested2": map[string]interface{}{"l2": "ahab"},
 					},
 					"scope":  "ahab",
@@ -170,6 +171,7 @@ func TestCoalesceValues(t *testing.T) {
 		{"{{.pequod.ahab.nested.foo}}", "true"},
 		{"{{.pequod.ahab.global.name}}", "Ishmael"},
 		{"{{.pequod.ahab.global.nested.foo}}", "bar"},
+		{"{{.pequod.ahab.global.nested.foo2}}", "<no value>"},
 		{"{{.pequod.ahab.global.subject}}", "Queequeg"},
 		{"{{.pequod.ahab.global.harpooner}}", "Tashtego"},
 		{"{{.pequod.global.name}}", "Ishmael"},


messiahUA:
When this could be merged?

AndrewFarley:
Please merge/release ASAP.  All builds from 3.13 onwards are "effectively" broken for many of my clients because of this.

ssttehrani:
We are also blocked and we can not upgrade our ArgoCD instances above 2.9.1

aabouzaid:
I hope to see this in the next Helm release, there are many workarounds used to overcome this bug.

duplabe:
I think this is also related https://github.com/helm/helm/issues/12991

AndrewFarley:
Beep boop we need this, boop beep

AndrewFarley:
Ahh, alas, another week goes by, and yet more DevOps engineers are fraught with this unnecessary pain in having to continue to use such an old version of Helm without this bug.  Beep boop, someone merge and release this, boop beep.

AndrewFarley:
Another week nudge on this.  Sorry to be the annoying jerk, but this is important.

Tawmu:
Sorry to go against etiquette here with comment spam but we're also affected by this bug, and with this PR waiting for review for almost 3 months now I think some of us might be losing hope :-)

pisto:
Just collecting all the duplicate bugs about this over the years, for fun I guess...

#12991 #12741 #12637 #12594 #12522 #12511 #12490 #12488 #12469 #12441 #12417 #11567 #9806 #9804 #9696 #9136 #9027 #6277 #5184

I'm so happy that the helm devs keep this bug alive so that I can charge my clients *hours* in working around it. Thanks! Tonight I eat steak thanks to you!

On a more serious note, who can we ping about this? I see @mattfarina worked on #12480 , sorry for the rude ping but maybe you were at least aware of the issue at some point in time.

MrBlaise:
I'd like to join the ping party, even though I know it won't help and normally I'd consider it rude as well, but this is getting ridiculous. Even if this is merged it will be some time until tools like argo-cd picks it up so that things can finally go back to the way they were.

Anyone with merge rights on this repo please take a look, lots of people would appreciate it ❤️

AndrewFarley:
Hmm, can we just start aggressively @'d owners and/or higher tier contributors on this repo?  I feel like maybe this thread has been muted

sylwit:
This PR fixes a really inconvenient bug and is highly expected to be merge. 
@joejulian do you know who we should ping to have a look at it or if this PR isn't ready can we have some feedback please ?
Thank you ❤️

joejulian:
I really want @mattfarina to look at this first. He recently dove deep into this and has the most context.

dannyg28:
Is there any update on this? Like others we're blocked on Argo version updates until this defect is resolved

rptaylor:
There are Helm developer meetings every Thursday I believe: https://docs.google.com/document/d/1d-6xJEx0C78csIYSPKJzRPeWaHG_8W1Hjl72OJggwdc/edit?pli=1#heading=h.cm5b1kdljh4m
Maybe an interested contributor could join to discuss the next steps for this PR (be nice!)

arya-harness:
We would really like this change as well. Is it possible to know which older version of helm didnt have this issue?

c0deaddict:
Also waiting for this. I want to override some default cpu limits of a subchart in ArgoCD, but can't..

swagnerfaw:
Is this still broken? Whats the reason this is not fixed for such a long time?

shastaxc:
Please?

AndrewFarley:
This issue going completely ignored so long makes me want to write a cronjob which just auto-reposts in here every day for infinity until this gets merged.  There's no reason for this, this is just causing a bunch of us needless pain and suffering.  Arguably, work, but, painful work.  

Please someone review/merge/fix this.  I'm going to create a new issue right now to reference this issue in this repo, in case this issue is on a ignore/unsubscribe list.

TylerHelmuth:
The [OpenTelemetry Helm Charts](https://github.com/open-telemetry/opentelemetry-helm-charts/issues/1160) would also love this feature.  I could've sworn this was fixed in `3.13`, but it appears not.

pisto:
Alright I'll bite and join the weekly meeting, if I am not busy. I am no longer affected by this bug (all charts I manage are custom), nor I am confident with the PR code itself, so I'd highly appreciate the presence of @ryanhockstad . Otherwise, I'll just point to this PR and explain the situation.

pisto:
I have a reproducible testcase btw https://github.com/pisto/helm-subchart-null-bug

ChieloNewctle:
I encountered the same problem when attempting to unset `seccompProfile` in the subchart `kong` of `kubernetes-dashboard`.
Sorry for anothing weekly ping. But this fix is really needed. Pelase. @mattfarina

MrBlaise:
> Alright I'll bite and join the weekly meeting, if I am not busy. I am no longer affected by this bug (all charts I manage are custom), nor I am confident with the PR code itself, so I'd highly appreciate the presence of @ryanhockstad . Otherwise, I'll just point to this PR and explain the situation.

Did you have a chance to join the meeting?

Plus another PING for me. Please for the love of god, should we fork helm is this project abandoned? I am obviously joking but this PR sure makes it look like a dead / amateur project.

messiahUA:
just a reminder that we can celebrate 5 months since this bug was fixed in the PR

gionn:
@mattfarina can we get a feedback on this from a maintainer, please?

pierluigilenoci:
Let's see if anything moves... 💣 

@joejulian
@jdolitsky
@marckhouzam
@mattfarina
@sabre1041 
@scottrigby 
@technosophos 

Please take a look.

Sorry for the bad manners. 🙏🏻

shawnhwei:
Bumping as I've also encountered this issue many times over the years.

TylerHelmuth:
@mattfarina, @technosophos  as discussed in the SIG meeting call today, I spent some time working on tests for this PR. I first started investigating why the [existing test](https://github.com/helm/helm/blob/b307c08c2be41f68424484b4700a61358cfcceb0/pkg/chartutil/coalesce_test.go#L66) passes. That test has several `null` merging scenarios with global and subcharts so I wanted to understand why they are properly deleted/merged and how the test differs from an actual `install` or `template` scenario.

Debugging via a `template` command and test scenario, I found that in an actual install scenario the call to `chartutil.ProcessDependenciesWithMerge` [here](https://github.com/helm/helm/blob/b307c08c2be41f68424484b4700a61358cfcceb0/pkg/action/install.go#L249) ends up modifying the chart's `Values` (I believe on purpose, as part of `Imports`) to include any default values from subcharts.

The subchart's default values being present in the parent's values ends up causing problems during the second call to `CoalesceValues` in the [Render function](https://github.com/helm/helm/blob/b307c08c2be41f68424484b4700a61358cfcceb0/pkg/chartutil/values.go#L161). When the subchart's default values are present in the parent chart's values.yaml, they end up being used, as described in the PR description, during the `coalesceDeps` step of `coalesce`, after the key had already been deleted.

The current unit tests does not have this problem because it does not call `ProcessDependenciesWithMerge` and the setup does not include any subchart values in the parent's values. As a result, when `coalesceDeps` is called there is no key to add back.

I've updated the test to show this failure scenario [here](https://github.com/TylerHelmuth/helm/commit/c34cd0c254102b35139e157e485581e72a0e36ce) by manually adding a subchart's default value to a parent chart's values. I have also (not rigorously) confirmed [here](https://github.com/helm/helm/compare/main...TylerHelmuth:helm:subchart-null-bug) that the proposed changes in this PR fixes this scenario.

I am going to work on more test cases/scenarios and try to determine if the fix should be made in coalesce, as proposed by this PR, or somewhere in `ProcessDependenciesWithMerge`. My goal will be to create enough test scenarios that you feel confident that this PR does not create a Whack-a-Mole scenario and the change and its new tests can be merged (or some equivalent change in `ProcessDependenciesWithMerge` if that ends up being more appropriate).

ivankatliarchuk:
Is there is a way to add `--experimental or --beta` flag or similar, so that we could release it and community to test it? Another option is to use https://github.com/helm-unittest/helm-unittest on top of unit tests itself

jkroepke:
Welcome @gjenkins8 - the community invites you to solve this issue. 

Thank you.

xrow:
I need this bugfix to override security contexts in OCP and Gitlab. My current workaround is similar to this :-(.

find ./cp-gitlab -type f -name "values.yaml" -exec sed -r -i 's/runAsUser:(\s+\w*|)/runAsUser: null/g' {} +
find ./cp-gitlab -type f -name "values.yaml" -exec sed -r -i 's/fsGroup:(\s+\w*|)/fsGroup: null/g' {} +
find ./cp-gitlab -type f -name "values.yaml" -exec sed -r -i 's/runAsGroup:(\s+\w*|)/runAsGroup: null/g' {} +

rehevkor5:
@TylerHelmuth I think you should make a PR against `ryanhockstad:subchart-null` which adds your unit tests, and get them to merge it? The addition of a meaningful unit test will make this PR more mature.

TylerHelmuth:
I have gotten pulled into other work and have not been able to get to the next step of creating tests. I believe to really prove this out we'd need some end-to-end tests that render some files to compare against expected renders. This ensures all the parts of the process are considered (not doing this is why the current unit test pass). The problem is there isn't any existing end-to-end tests like this to use, so it would mean creating a new process.

If someone wants to help move this forward the best thing to do next, in my opinion, is make a bunch of tests that go through the whole render process, as if you were doing a `helm template`, and compare the results to expected results.  I don't have time for this right now, but I hope someone else does.

ExNG:
When templating a chart with nested dependencies i've encountered a case of untyped subchart values.
Tested with patched helm `3.15.4` and `3.16.2`, also with unpatched `3.14.4`, `3.15.4` and `3.16.2`.

My test case was done using the [camunda/camunda-platform](https://github.com/camunda/camunda-platform-helm/blob/main/charts/camunda-platform-8.6/Chart.yaml) at v11.0.1, but interesstingly not with v10.4.2 and [bitnami/wordpress](https://github.com/bitnami/charts/blob/main/bitnami/wordpress/Chart.yaml) at v23.1.23 which have sub-sub-dependencies too.

The debug error printed is:
```
Error: template: test/charts/camunda/charts/identityKeycloak/templates/statefulset.yaml:163:27: executing "test/charts/camunda/charts/identityKeycloak/templates/statefulset.yaml" at <include "keycloak.databaseSecretName" .>: error calling include: template: test/charts/camunda/charts/identityKeycloak/templates/_helpers.tpl:168:75: executing "keycloak.databaseSecretName" at <.Values.postgresql.auth.existingSecret>: wrong type for value; expected string; got interface {}
helm.go:86: 2024-10-24 08:31:48.294951444 +0000 UTC m=+0.172370966 [debug] template: test/charts/camunda/charts/identityKeycloak/templates/statefulset.yaml:163:27: executing "test/charts/camunda/charts/identityKeycloak/templates/statefulset.yaml" at <include "keycloak.databaseSecretName" .>: error calling include: template: test/charts/camunda/charts/identityKeycloak/templates/_helpers.tpl:168:75: executing "keycloak.databaseSecretName" at <.Values.postgresql.auth.existingSecret>: wrong type for value; expected string; got interface {}
```

I did test other revisions of the camunda/camunda-platform v11 chart but didn't find a working state, this might hint on a verry rare case in which this fix might not work.
Luckily this specific chart includes a `global.compatibility.openshift.adaptSecurityContext` option as fallback.
I hope this insight is helpful, though i am sorry i cannot provide a fix.

ryanhockstad:
@ExNG I'm not sure what values you used in your test chart, but I think I know the problem.  If you are using the default `values.yaml`, it [explicitly sets](https://github.com/camunda/camunda-platform-helm/blob/10d6034c0cc900ecab0be9cbc0d98cb372d31051/charts/camunda-platform-8.6/values.yaml#L2597) the `identityKeycloak.postgresql.auth.existingSecret` (which is in the postgresql subchart to the keycloak subchart) to a null value. The whole premise of this PR is to respect that null value rather than have it get overwritten.  As you can see [here](https://github.com/bitnami/charts/blob/754f9e2365ab6d0786a997e4d582e2b6ce92c237/bitnami/keycloak/values.yaml#L1316), the Keycloak subchart does explicitly set this value to an empty string. 

So this raises a pretty existential question for this PR: Should this bug be patched if people are accidentally depending on it?

On one hand, I think it's definitely safe to call this a bug.  In the Camunda chart case, it may be very helpful that values they are explicitly setting end up getting overwritten, but that still seems like a bug.

On the other hand... this could be a disruptive patch for some existing helm charts.  I think the error you saw made it fairly clear what the problem was and should make it quick for chart maintainers to update their code.  It should also be easy enough for an end user to patch with their custom values file.  Having said that, Helm wants to be very backwards compatible, and in some cases, this will be a breaking change. I would understand if the maintainers want to move this to Helm 4 because of that.

I would obviously like it if it got merged before Helm V4, and I know that is the prevailing feeling here.  But I think we need to hear from a Helm maintainer on this.

MrBlaise:
> > On the other hand... this could be a disruptive patch for some existing helm charts
> 
> While I agree that this will now be disruptive for some, I'm not sure you're considering that this change has already been made and has already broken a bunch of people's Helm charts, and continues to do so as people upgrade their version of Helm to a "fixed/broken" version. Additionally, I'd bet that far less people are depending on this new logic compared to the entire history of Helm utilizing this. Finally, and more importantly, we, in Helm, NEED a way to unset a value in a subchart. Currently, this is impossible with the latest version of Helm. I personally stay out of date because of this. Subcharting in Helm is one of the more critical features if you ask me, it allows you to fully templatize and craft an entire (simple) application stack in a single helm chart with multiple sub-charts. Without a way to override a subchart, what you end up with is then needing to fork/clone the upstream chart and maintain it yourself to be able to change some "dumb" default that has no easy way to disable besides unsetting that value.
> 
> Overall, I think the good that undoing the bad that was done by "fixing" this outweighs the bad. By far. Unless you can come up with a way to unset/override subcharts with some other internal keyword meaning the same thing?

Exactly! We haven’t updated argocd since this issue occured, because we are refusing to rewrite our helm charts to the obviously wrong version. People who already changed their helm charts, will change it again and things will go back to normal. When this bug was introduced it caused a lot of headaches, and the longer it is not fixed the bigger headache it will cause again when it finally fixed, but that doesn’t mean it shouldn’t.

Just because we can workaround a bad design doesnt justify the bad design.

xrow:
Idea 1: If you are concerned about BC, maybe it is time for helm 2.X.X.
Idea 2: Have a similar field like kubeVersion called helmVersion in Chart.yaml

AndrewFarley:
Can someone just merge and release this already?  What's the blocker?  Whose door do I need to bust down?

fcrespofastly:
I have to say this issue is making charts dependency completely broken :(

MrBlaise:
Any updates here, what is the latest blocker?

xmj:
Reading through this thread I am reminded of https://xkcd.com/1172/.

What's the holdup in getting it in by now? One does lose track...

xrow:
@scottrigby @karenhchu @mattfarina Trying to poke a relevant maintainer. Whoever the proper person is. To get direction, guidance or management to stop the ongoing loop of ideas and comments. Please take a moment care. Ty.

pisto (approved):
As far as I can tell, this is ok.

karenhchu:
Hi all! Thank you for the contribution. The maintainers need to take a deeper look since there have been several attempts on this issue. That being said, as a next step to help this move along – can we start with getting some tests?

scottrigby:
Thanks everyone for your feedback and contributions on this.

I can understand some frustration here, but—as Karen said https://github.com/helm/helm/pull/12879#issuecomment-2551982738—PRs need tests in order to get merged. Especially when the PR affects code that could have larger consequences. And that's exactly the case here, since this is part of a larger issue with Helm values coalescing. Helm has because a critical piece of infrastructure tooling for many large organizations, and we need to make sure a fix for one issue doesn't break things elsewhere.

Note that was an earlier fix to the coalesce issue merged, but was later rolled back because it broke something else. This is because the coalesce logic in Helm is very confusing—the precedence in code was different in some cases to what was documented, and we didn't have tests that clearly defined this precedence order in the coalesce logic. Because of this, we held weekly meetings on this topic for a time with a goal of coming to a final resolution on the precedence question, and we have done that. The results are in this Google doc, which also links to a number of issues and pull requests. Please take a look: https://docs.google.com/document/d/1ZjmIc5VYzNbaJ-SQ6nT02Oo1t9ctlvomufP8ulR6q70/edit?tab=t.0#heading=h.stcvhj9ydagu

I would like to help address these connected issues properly now that we have resolved the precedence discrepancy. Several other Helm maintainers would too. Will add a link here to new issue/PR for that.

scottrigby:
Happy holidays everyone ❄️🌲

I added some tests as a linked git patch, and based on that believe this PR works as expected without breaking other things—at least what we have test coverage for and I could think to test manually.

This patch can be git applied and pushed to this PR branch by @ryanhockstad, or a Helm maintainer can do it. 

I am happy to do that, but I would like some community feedback first. I believe the use cases for null values are now covered in these tests via this patch, but please take a look and give feedback about whether I missed any relevant use cases here.

Note that this patch also:
- Adds consistency for null test in given values, parent chart, subchart, and    
  sub-sub-chart                                                                 
- Removes bar null test to keep consistent with boat=null at top level

If you want to test locally:

1. checkout this PR branch locally
  _for folks unfamiliar with how to do this, run `gh pr checkout 12879` in your local clone, using the [GitHub CLI](https://cli.github.com/)_
2. download this patch file [0001-Tests-for-bugfix-Override-subcharts-with-null-values.patch](https://github.com/user-attachments/files/18246595/0001-Tests-for-bugfix-Override-subcharts-with-null-values.patch)
3. run `git apply /PATH/TO/PATCHFILE`
4. run `go test ./pkg/chartutil -run TestCoalesceValues`. You should get `ok  	helm.sh/helm/v3/pkg/chartutil`

scottrigby:
Fixed milestone. This will also need a v3 backport (at this point, it should be the exact same commits)

ryanhockstad:
Thanks for adding these tests, @scottrigby! I can confirm they are failing in the current `main` branch, but passing in this PR, so I added your commit.

As far as coverage goes, I think we should add a test that 'nullifies' an entire map.  I've added that change in [this commit](https://github.com/helm/helm/pull/12879/commits/ef2eb552837446fe8a5a91bc68abf4d9eab685a8). What do you think?

scottrigby:
> Thanks for adding these tests, @scottrigby! I can confirm they are failing in the current `main` branch, but passing in this PR, so I added your commit.

👍

> As far as coverage goes, I think we should add a test that 'nullifies' an entire map.  I've added that change in [this commit](https://github.com/helm/helm/pull/12879/commits/ef2eb552837446fe8a5a91bc68abf4d9eab685a8). What do you think?  

I think it's a good addition 👍

ryanhockstad:
@TylerHelmuth  I think adding a test for global is a good idea to address how that section is handled differently.  I added a test in the latest commit [here](https://github.com/helm/helm/pull/12879/commits/8da0374896fabdec12b0d9ab3ef6f523c001448f).  Setting the new global `foo2` to null under the `pequod` section nullifies its `ahab` subchart global value as expected.

I think the reason this works without requiring the code change you suggested is because the bug that my change fixes prevents `coalesce` from forgetting about explicit `null`s set in a subchart.  The logic that handles dealing with `global` values is totally separate.

ryanhockstad:
I was just testing this out a little more locally, and there is a weird case where one chart might set a default for a global value and a subchart could set a default for the same value. 

In this case, if a user were to explicitly set the top level global to `null`, the subchart's global value would not be nullified. This is because the top level global would disappear, so nothing at the top level would exist to nullify the default subchart global value.

@scottrigby if that is the situation you were thinking about, I think the best solution would probably be for chart users to explicitly nullify the global value at the top level and global value within the subchart, which works.  I could update the test I just added to reflect that.

V02460:
First, thank you for working on this PR! I have some concerns with it, though: While it fixes important issues, it is another bandaid on top of the more fundamental issue of coalescing: Overloading the meaning of `null` will always bite someone in some way.

A better solution, to my understanding, would be to completely replace coalescing with simple merging. This will make it easy to understand how final values are computed and eliminates all edge cases that come with the current inconsistent mix of both. Remove coalescing and there is no more double meaning of what `null` means.

For the rare cases where a key with a `null` value is interpreted differently to a non-existing key (a bad anti-pattern in my opinion), the considered `--unset` command line flag could be realized. Or, more broadly, a patch mechanism like, e.g., JSON Patch could support key deletion in a clean way besides offering additional handy operations. Whatever the tool to perform this deletion, it must be available, but be separated from the basic merging of Helm values.

This PR is a move in the right direction by removing some uses of coalescing, but I think the proper solution is abolishing it completely. Thanks for considering!

scottrigby:
@V02460 I agree this is something we should reconsider for Helm 4. Now `main` is for Helm 4, while Helm 3 backports should be made against `dev-v3` branch after they are fixed in `main`.  That said, this PR is _very_ close to being ready to merge. At the moment a back port would be nearly as-is—because most of the code in both branches is the same, although I expect this to change very soon as we begin to make major architectural changes for Helm 4. Because of that, I would like to see this fixed in `main` with the current functionality—if possible—before making a change to the method we use to delete keys. Note that I initially introduced the functionality in Helm to delete keys, and because of increasing complexity of coalescing over time and problems caused, I suggested the `--unset` alternative and still think that would be a better solution in Helm 4. That would be a separate PR, and should start as a [Helm Improvement Proposal (HIP)](https://github.com/helm/community/tree/main/hips).

scottrigby:
> I was just testing this out a little more locally, and there is a weird case where one chart might set a default for a global value and a subchart could set a default for the same value.
> 
> In this case, if a user were to only explicitly set the top level global to `null`, the subchart's global value would not be nullified. This is because the top level global would disappear, so nothing at the top level would exist to nullify the default subchart global value. Likewise, if a user only nullified the subchart global value, it would end up still getting overwritten by the top level default global value.
> 
> @scottrigby if that is the situation you were thinking about, I think the best solution would probably be for chart users to explicitly nullify the global value at the top level and global value within the subchart, which works. I could update the test I just added to reflect that.

@ryanhockstad I have an idea. How about waiting until after coalescing is finished to delete any keys that have a remaining (non-overridden) `null` value? That way, existing coalescing precedence would be respected.

ryanhockstad:
@scottrigby I'm not sure the added complexity of adding logic to nullify certain keys after the merge/coalesce recursion has occurred once is worth dealing with edge cases that users are able to handle themselves anyway.

Also, the way that this should be handled is a bit ambiguous anyway.  A user may want to actually nullify a parent global key without nullifying all of the children globals.

scottrigby:
OK so I think we should keep the scope of this PR to a fix for known issues users are having. It addresses them very well with many new tests, and as such is a valuable improvement. Let's move consideration for fixing possible issues with globals and null values to a separate issue. I'm running the tests now—let's see where they land.

scottrigby (approved):
lgtm 👍

joejulian (approved):
lgtm

scottrigby:
Thanks @joejulian and everyone 👏 Merging

@ryanhockstad do you want to open a v3 backport PR from your same branch while the code applies as-is, and link to this one?

AndrewFarley:
10 months later, amazing, thank you team.  Now just waiting for the backport and release of Helm 3 with this patch

ryanhockstad:
Here is the backport PR: https://github.com/helm/helm/pull/13654

pierluigilenoci:
@ryanhockstad thank you!

scottrigby:
📣 Backport PR https://github.com/helm/helm/pull/13654 merged 👏 This fix is scheduled to be in Helm 3.17.1

PR #12894, opened 2024-03-18
added correct usage of helm install command

### LINKED ISSUE #12848: Document the default value for `.Release.Name`
Output of `helm version`: N/A

Can the documentation at <https://helm.sh/docs/chart_template_guide/builtin_objects/> document what value `.Release.Name` will take, in particular, in the case that it isn't specified?

NeerajNagure:
**What this PR does / why we need it**:  Documented the correct usage of helm install command by specifying the necessary arguments which were considered as options before

closes #12848 
**If applicable**:

### DIFF
--- a/cmd/helm/install.go
+++ b/cmd/helm/install.go
@@ -132,7 +132,7 @@ func newInstallCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 	var outfmt output.Format
 
 	cmd := &cobra.Command{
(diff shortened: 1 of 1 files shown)


Bhargav-InfraCloud (changes requested):
Unit test(s) should be fixed. Tests under `TestInstall` are failing.

Also, it should be updated in `install-no-args.txt`.

PR #12906, opened 2024-03-24
Fixed lint.Run issue so that it wont modify input param

### LINKED ISSUE #12412: lint.Run should not modify input param
Output of `helm version`:
v3@3.12.3
Output of `kubectl version`:

////////////////////Problem description//////////////////////////////////////////////
Function 
func (l *Lint) Run(paths []string, **vals** map[string]interface{}) *LintResult, 
after calling it, the second input param **vals** got changed,  this is unexpected, i think this is a bug because lint is not supposed to change the input param!

**I've done some experiment, when input map "vals" has some keys same as the values.yaml( exists in paths), after calling this lint.Run, the vals got changed.**

////////////////////////my test code:///////////////////////////////////////////////////////

```go

package main

import (
	"encoding/json"
	"fmt"
	"helm.sh/helm/v3/pkg/action"
)

func main() {
	chartPath := "hello-world"
	jsonString := {"key1":{"key2":"false","key3":"false"},"key4":{"exitcode":0},"key5":null}

	var data map[string]interface{}
	err := json.Unmarshal([]byte(jsonString), &data)
	if err != nil {
		fmt.Println("Unmarshal Error:", err)
		return
	}
	fmt.Println("before lint: vals=", string(jsonString))
	lint := action.NewLint()
	lint.WithSubcharts = true
	result := lint.Run([]string{chartPath}, data)
	if len(result.Errors) > 0 {
		fmt.Println("error!", result.Errors)
	}
	bs, _ := json.Marshal(data)
	fmt.Println("after lint: vals=", string(bs))
	return
}

```
//////////////////////////////////////////test result////////////////////////////////////
before lint: vals= {"key1":{"key2":"false","key3":"false"},"key4":{"exitcode":0},"key5":null}
after lint: vals= {"key1":{"key2":"false","key3":"false","product":"ABC"},"key4":{"exitcode":0,"testexitcode":0},"key5":null}

values.yaml has following content:
key1:
  product: "ABC"
  key2: "true"
  key3: "false"

key4:
  exitcode: 0
  key5: 0

////////////////////////////////////////////////////////////////////////////////

NeerajNagure:
This PR fixes the lint.Run issue that was modifying the input params provided to it.Used a copy of the params to avoid that modification
closes #12412 

**If applicable**:

### DIFF
--- a/pkg/lint/rules/values.go
+++ b/pkg/lint/rules/values.go
@@ -70,7 +70,11 @@ func validateValuesFile(valuesPath string, overrides map[string]interface{}) err
 	// We could change that. For now, though, we retain that strategy, and thus can
 	// coalesce tables (like reuse-values does) instead of doing the full chart
 	// CoalesceValues
-	coalescedValues := chartutil.CoalesceTables(make(map[string]interface{}, len(overrides)), overrides)
+	overridesCopy := make(map[string]interface{})
+	for key, value := range overrides {
+		overridesCopy[key] = value
+	}
+	coalescedValues := chartutil.CoalesceTables(make(map[string]interface{}, len(overridesCopy)), overridesCopy)
 	coalescedValues = chartutil.CoalesceTables(coalescedValues, values)
 
 	ext := filepath.Ext(valuesPath)


NeerajNagure:
Can someone review this PR and provide suggestions?

NeerajNagure:
Can any member please review this PR?  Its been pending since more than a month

PR #12910, opened 2024-03-26
Added suit tests and updated the package versions

anirudh-hegde:
Updated the package versions and added piece of code for test suites


anirudh-hegde:
Please somebody review my code

robertsirc (changes requested):
So, fortio.org/assert doesn't have any dependencies and you are changing dependencies that should not need to be updated.

pkg/registry/client_http_test.go
```diff
@@ -63,6 +65,19 @@ func (suite *HTTPRegistryClientTestSuite) Test_4_ManInTheMiddle() {
 	suite.True(errdefs.IsFailedPrecondition(err))
 }
 
+func (suite *HTTPRegistryClientTestSuite) SetupTest() {
+	suite.VariableThatShouldStartAtFive = 5
+}
+
+func (suite *HTTPRegistryClientTestSuite) TestExample() {
+	if suite.Equal(5, suite.VariableThatShouldStartAtFive) {
+		assert.Equal(suite.T(), 5, suite.VariableThatShouldStartAtFive)
```
mattfarina, line 74: Why is this assert package being added when testify can already do what's being used?

mattfarina (changes requested):
The new package being added doesn't have any dependencies. So, the dependency tree should not be updated like this. Can you please back out and only make the Go dependencies update for this single change?

Or, maybe use the existing assert package with testify, unless I'm missing something.

PR #12912, opened 2024-03-26
feat: add httproute from gateway-api to create chart template

### LINKED ISSUE #12603: gateway api template
Now that the gateway api has hit v1.0, it would be nice if:
```
helm create foo
```

Included generating support for a gateway api HTTPRoute (https://gateway-api.sigs.k8s.io/api-types/httproute/) out of the box.

hegerdes:
This closes #12603

This PR adds the HTTPRoute from [gateway-api](https://gateway-api.sigs.k8s.io/reference/spec/) to the example getting started chart. 
It is disabled by default, just like ingress.  
Adding the resource will make helm charts future-proof for the gateway-api and may also increase its adoption rate.

If enabled, the via `httpRoute.enabled: true`, it will produce a valid `HTTPRoute` but it will not route real traffic since a `Gateway` must be installed and configured which is in the responsibility of the cluster admin, just like ingress-controller.

The gateway api is still quite young and the gateway-api CRDs are not yet part of a default k8s installation. They currently have to be installed separably with `kubectl apply -f https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.0.0/standard-install.yaml`.  
~~The template checks if the `v1` or the `v1alpha2` version is installed and sets the api version accordingly.  
If the CRDs are missing the installation fails.~~

**If applicable**:

### DIFF
--- a/pkg/chart/v2/util/create.go
+++ b/pkg/chart/v2/util/create.go
@@ -54,8 +54,8 @@ const (
 	IgnorefileName = ".helmignore"
 	// IngressFileName is the name of the example ingress file.
 	IngressFileName = TemplatesDir + sep + "ingress.yaml"
-	// HttpRouteFileName is the name of the example HTTPRoute file.
-	HttpRouteFileName = TemplatesDir + sep + "httproute.yaml"
+	// HTTPRouteFileName is the name of the example HTTPRoute file.
+	HTTPRouteFileName = TemplatesDir + sep + "httproute.yaml"
 	// DeploymentName is the name of the example deployment file.
 	DeploymentName = TemplatesDir + sep + "deployment.yaml"
 	// ServiceName is the name of the example service file.
@@ -335,7 +335,7 @@ spec:
 {{- end }}
 `
 
-const defaultHttpRoute = `{{- if .Values.httpRoute.enabled -}}
+const defaultHTTPRoute = `{{- if .Values.httpRoute.enabled -}}
 {{- $fullName := include "<CHARTNAME>.fullname" . -}}
 {{- $svcPort := .Values.service.port -}}
 {{- if .Capabilities.APIVersions.Has "gateway.networking.k8s.io/v1" -}}
@@ -755,8 +755,8 @@ func Create(name, dir string) (string, error) {
 		},
 		{
 			// httproute.yaml
-			path:    filepath.Join(cdir, HttpRouteFileName),
-			content: transform(defaultHttpRoute, name),
+			path:    filepath.Join(cdir, HTTPRouteFileName),
+			content: transform(defaultHTTPRoute, name),
 		},
 		{
 			// deployment.yaml
--- a/pkg/chart/v2/util/create.go
+++ b/pkg/chart/v2/util/create.go
@@ -189,6 +189,7 @@ httpRoute:
   parentRefs:
   - name: gateway
     sectionName: http
+    # namespace: default
   # -- Hostnames matching HTTP header.
   hostnames:
   - "example.com"
@@ -525,7 +526,20 @@ spec:
 `
 
 const defaultNotes = `1. Get the application URL by running these commands:
-{{- if .Values.ingress.enabled }}
+{{- if .Values.httpRoute.enabled }}
+{{- if .Values.httpRoute.hostnames }}
+    export APP_HOSTNAME={{ .Values.httpRoute.hostnames | first }}
+{{- else }}
+    export APP_HOSTNAME=$(kubectl get --namespace {{(first .Values.httpRoute.parentRefs).namespace | default .Release.Namespace }} gateway/{{ (first .Values.httpRoute.parentRefs).name }} -o jsonpath="{.spec.listeners[0].hostname}")
+  {{- end }}
+{{- if and .Values.httpRoute.rules (first .Values.httpRoute.rules).matches (first (first .Values.httpRoute.rules).matches).path.value }}
+    echo "Visit http://$APP_HOSTNAME{{ (first (first .Values.httpRoute.rules).matches).path.value }} to use your application"
+
+    NOTE: Your HTTPRoute depends on the listener configuration of your gateway and your HTTPRoute rules.
+    The rules can be set for path, method, header and query parameters.
+    You can check the gateway configuration with 'kubectl get --namespace {{(first .Values.httpRoute.parentRefs).namespace | default .Release.Namespace }} gateway/{{ (first .Values.httpRoute.parentRefs).name }} -o yaml'
+{{- end }}
+{{- else if .Values.ingress.enabled }}
 {{- range $host := .Values.ingress.hosts }}
   {{- range .paths }}
   http{{ if $.Values.ingress.tls }}s{{ end }}://{{ $host.host }}{{ .path }}
--- a/pkg/chart/v2/util/create.go
+++ b/pkg/chart/v2/util/create.go
@@ -179,7 +179,9 @@ ingress:
   #    hosts:
   #      - chart-example.local
 
-# -- How the service is exposed via gateway-apis HTTPRoute.
+# -- Expose the service via gateway-api HTTPRoute
+# Requires Gateway API resources and suitable controller installed incluster
+# (see: https://gateway-api.sigs.k8s.io/guides/)
 httpRoute:
   # -- HTTPRoute enabled.
   enabled: false
--- a/pkg/chart/v2/util/create.go
+++ b/pkg/chart/v2/util/create.go
@@ -338,11 +338,7 @@ spec:
 const defaultHTTPRoute = `{{- if .Values.httpRoute.enabled -}}
 {{- $fullName := include "<CHARTNAME>.fullname" . -}}
 {{- $svcPort := .Values.service.port -}}
-{{- if .Capabilities.APIVersions.Has "gateway.networking.k8s.io/v1" -}}
 apiVersion: gateway.networking.k8s.io/v1
-{{- else -}}
-apiVersion: gateway.networking.k8s.io/v1alpha2
-{{- end }}
 kind: HTTPRoute
 metadata:
   name: {{ $fullName }}
--- a/pkg/chart/v2/util/create.go
+++ b/pkg/chart/v2/util/create.go
@@ -54,6 +54,8 @@ const (
 	IgnorefileName = ".helmignore"
 	// IngressFileName is the name of the example ingress file.
 	IngressFileName = TemplatesDir + sep + "ingress.yaml"
+	// HttpRouteFileName is the name of the example HTTPRoute file.
+	HttpRouteFileName = TemplatesDir + sep + "httproute.yaml"
 	// DeploymentName is the name of the example deployment file.
 	DeploymentName = TemplatesDir + sep + "deployment.yaml"
 	// ServiceName is the name of the example service file.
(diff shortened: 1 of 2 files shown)


gjenkins8 (commented):
nice, thanks for the PR.

A suggestion: perhaps add an example on how to access the installed deployment into NOTES.txt (https://github.com/helm/helm/blob/main/pkg/chartutil/create.go#L438) ?

(I'm not familiar enough with gateway API to know if this is entirely reasonable; but if so, might be helpful to chart consumers/authors?)

pkg/chartutil/create.go
```diff
@@ -299,6 +337,50 @@ spec:
 {{- end }}
 `
 
+const defaultHttpRoute = `{{- if .Values.httpRoute.enabled -}}
+{{- $fullName := include "<CHARTNAME>.fullname" . -}}
+{{- $svcPort := .Values.service.port -}}
+{{- if .Capabilities.APIVersions.Has "gateway.networking.k8s.io/v1" -}}
+apiVersion: gateway.networking.k8s.io/v1
+{{- else -}}
+apiVersion: gateway.networking.k8s.io/v1alpha2
```
kfox1111, line 346: Is v1 common enough now to just drop v1alpha2 from the template?

gjenkins8: nice spot @kfox111 -- this code is going to be in Helm and created charts for a "long time". I think if v1 is already available, then don't include support for the alpha version. A user who really needs this can manually edit.

sabre1041 (commented):
Reviewed the content of the PR and here are some comments:

1. Let's align the default hostname among `Ingress` and `HTTPRoute` so that there is pariety. Yes, you could potentially install both, but in most cases, you would pick one or the other. So set the default hostname of the `HTTPRoute` to be `chart-example.local`
2. Support for `v1beta1`. While the Gateway API is now GA, most who are currently using it is making use of the beta version. Similar to `Ingress`, there should be a version check to determine which API Group/Version to apply
3. Overall thoughts: I am hesitant about adding this to the default scaffolded chart. This is mainly from a user experience perspective as we would want the getting started experience to be as smooth as possible. Most users will not have the Gateway API installed and there is a chance that they would face errors when attempting to leverage some of the features that are available in the default scaffolded chart.

hegerdes:
Thanks for the feedback @sabre1041 

> 1. Let's align the default hostname among Ingress and HTTPRoute so that there is pariety. Yes, you could potentially install both, but in most cases, you would pick one or the other. So set the default hostname of the HTTPRoute to be chart-example.local

Sure will do!

> 2. Support for v1beta1. While the Gateway API is now GA, most who are currently using it is making use of the beta version. Similar to Ingress, there should be a version check to determine which API Group/Version to apply

I thought so too at first but then the question was raised if it is used that much by people, so I removed it. I can revert that without a problem but you also mentioned user experience. I will make the assumption that the people using `v1beta1` are technical enough and on the cutting edge side so they will know to change the api version to `v1beta1`. And if not and they use `v1beta1` functions the deployment will just fail. Let the community/maintainers discuss and decide. I have no real preference here.

> 3. Overall thoughts: I am hesitant about adding this to the default scaffolded chart. This is mainly from a user experience perspective as we would want the getting started experience to be as smooth as possible. Most users will not have the Gateway API installed and there is a chance that they would face errors when attempting to leverage some of the features that are available in the default scaffolded chart.

I understand that concern. My goal was to make using the GatewayAPI easy because at fist I struggled transforming something usable into helm. To not disturb the user experience the gateway stuff is disabled by default, you have to enable it explicitly and then it _should_ provide a good starting point for your first routes. I see it similar to Ingress and HorizontalPodAutoscaler. A lot of people don't use these and just delete the template but if you need them it is nice to have a good starting point.

sabre1041:
> Thanks for the feedback @sabre1041
> 
> > 1. Let's align the default hostname among Ingress and HTTPRoute so that there is pariety. Yes, you could potentially install both, but in most cases, you would pick one or the other. So set the default hostname of the HTTPRoute to be chart-example.local
> 
> Sure will do!
> 
> > 2. Support for v1beta1. While the Gateway API is now GA, most who are currently using it is making use of the beta version. Similar to Ingress, there should be a version check to determine which API Group/Version to apply
> 
> I thought so too at first but then the question was raised if it is used that much by people, so I removed it. I can revert that without a problem but you also mentioned user experience. I will make the assumption that the people using `v1beta1` are technical enough and on the cutting edge side so they will know to change the api version to `v1beta1`. And if not and they use `v1beta1` functions the deployment will just fail. Let the community/maintainers discuss and decide. I have no real preference here.
> 
> > 3. Overall thoughts: I am hesitant about adding this to the default scaffolded chart. This is mainly from a user experience perspective as we would want the getting started experience to be as smooth as possible. Most users will not have the Gateway API installed and there is a chance that they would face errors when attempting to leverage some of the features that are available in the default scaffolded chart.
> 
> I understand that concern. My goal was to make using the GatewayAPI easy because at fist I struggled transforming something usable into helm. To not disturb the user experience the gateway stuff is disabled by default, you have to enable it explicitly and then it _should_ provide a good starting point for your first routes. I see it similar to Ingress and HorizontalPodAutoscaler. A lot of people don't use these and just delete the template but if you need them it is nice to have a good starting point.

The biggest concern with including an `HTTPRoute` with the default set of scaffolded resources during Helm chart creation is that unlike `Ingress` or `HorizontalPodAutoscaler`, an `HTTPRoute` is not an included API within the cluster. So, if they attempt to enable the feature as part of the installation of a chart to the cluster, an error will be returned, not as a result of an error within the contents of the chart, but due to component(s) not being available within the cluster. Its all about the user experience.

kfox1111:
The Kubernetes folks intend for the gateway api to replace the ingress api though. There is a chicken and the egg problem though. Enough software needs to adopt the ability to use the gateway api before enough clusters deploy it. Clusters are reluctant to deploy it unless software uses it. Software developers are less likely to support it if they don't have a good example of how to do it. Hence this pr's goal to help break the deadlock.

hegerdes:
Changed the default `HTTPRoute` hostname in values cording to `chart-example.local` as  suggested by @sabre1041

Rebased from main

sabre1041:
> Changed the default `HTTPRoute` hostname in values cording to `chart-example.local` as suggested by @sabre1041
> 
> Rebased from main

Any thoughts about supporting older versions of the Gateway API pre GA?

hegerdes:
Seems like there a no real opinions on that. Also asked in the k8s gateway-api slack channel.

Given that `v1.1` is already released I don`t see the real need to support beta versions as I dont think they will be needed once it hits a higher adoption rate.

youngnick:
So, a few comments from the Gateway API maintainer POV (that's me!):

* Gateway API is a set of extension CRDs that are _an official part of Kubernetes_. They are subject to the same API review and promotion controls that included APIs are in the core Kubernetes binaries. To be honest, Gateway API is the first of many here - this pattern is already adopted by the AdminNetworkPolicy folks, and more will follow. It makes sense for Helm to have a way to handle this "definitely a Kubernetes API but not included by default" issue. (There's a larger issue about ways to ensure that the required version of Gateway API is installed as well, but that's a _way_ larger discussion). As @kfox1111 says, Gateway API and these other APIs are most likely _never_ going to be installed by default though.
* For Helm charts, I think the way to go is to end up with three settings - `hostname`, `gatewayName`, `gatewayNamespace`, which can be then populated into a HTTPRoute resource and link it up with already-existing infrastructure. Setting one of these should require all three, and the HTTPRoute files should only be included if the values are not empty. (Presumably, this goes for `helm create` as well).

I'm happy to come to a Helm community call or something at some point to discuss this in a higher-bandwidth way, or to meet up at Kubecon NA and talk there too.

There are two Helm-related problems that have been on my mind for some time:
* How to help chart authors include HTTPRoutes more easily (that's this one)
* How to help chart authors say "Gateway API being installed is a dependency" and "Gateway API must be at least this version", both of which are surprisingly hard problems that will take a while to work out. But having more communication between Gateway API folks and Helm folks here will help a lot, I think.

gjenkins8 (dismissed):
I approve, but other maintainers may still have differing opinions

pkg/chartutil/create.go
```diff
@@ -163,6 +165,42 @@ ingress:
   #    hosts:
   #      - chart-example.local
 
+# -- How the service is exposed via gateway-apis HTTPRoute.
```
gjenkins8, line 168:
```suggestion
# -- Expose the service via gateway-apis HTTPRoute
# Requires Gateway API resources and suitable controller installed incluster
# (see: https://gateway-api.sigs.k8s.io/guides/)
```

hegerdes:
Thanks for approving and your feedback @gjenkins8. I will update the comment tomorrow and will also rebase.

Just out of interest: Did the topic of any none core kubernetes-apis come up at kubecon? I was not able to attend, but am quite interested if there where opinions exchanged. Regarding the role of helm and gateway-api

gjenkins8 (dismissed):

kfox1111:
Anything left to do on this?

kfox1111:
Whats left to get this in?

pkg/chartutil/create.go
```diff
@@ -177,6 +179,44 @@ ingress:
   #    hosts:
   #      - chart-example.local
 
+# -- Expose the service via gateway-api HTTPRoute
+# Requires Gateway API resources and suitable controller installed incluster
```
sabre1041, line 183:
```suggestion
# Requires Gateway API resources and suitable controller installed within the cluster
```

pkg/chartutil/create.go
```diff
@@ -177,6 +179,44 @@ ingress:
...
+# (see: https://gateway-api.sigs.k8s.io/guides/)
+httpRoute:
+  # -- HTTPRoute enabled.
```
sabre1041, line 186:
```suggestion
  # HTTPRoute enabled.
```

pkg/chartutil/create.go
```diff
@@ -177,6 +179,44 @@ ingress:
...
+  enabled: false
+  # -- HTTPRoute annotations.
```
sabre1041, line 188:
```suggestion
  # HTTPRoute annotations.
```

pkg/chartutil/create.go
```diff
@@ -177,6 +179,44 @@ ingress:
...
+  annotations: {}
+  # -- Which Gateways this Route is attached to
```
sabre1041, line 190:
```suggestion
  # Which Gateways this Route is attached to
```

pkg/chartutil/create.go
```diff
@@ -177,6 +179,44 @@ ingress:
...
+  parentRefs:
+  - name: gateway
+    sectionName: http
+    # namespace: default
+  # -- Hostnames matching HTTP header.
```
sabre1041, line 195:
```suggestion
  # Hostnames matching HTTP header.
```

pkg/chartutil/create.go
```diff
@@ -177,6 +179,44 @@ ingress:
...
+  hostnames:
+  - chart-example.local
+  # -- List of rules and filters applied.
```
sabre1041, line 198:
```suggestion
  # List of rules and filters applied.
```

sabre1041 (changes requested):
Logic wide, it looks good. Just a couple cosmetic changes since they are applied by default, to all newly created charts

hegerdes:
Implemented the requested changes to values.yaml template

sabre1041 (dismissed):
LGTM

hegerdes:
@gjenkins8 No problem!

No code changes required, git handled the modified repo structure just fine in a simple rebase

gjenkins8 (approved):

sabre1041 (approved):
LGTM

Issue #12896, opened 2024-03-19
Kind Sorter missing list based resources


jmajoor:
The kind sorter is missing many of the list based resources.
For example:
Deployment, StatefulSet, ReplicaSet, DaemonSet, Ingress, Service, ConfigMap are defined in the kind sorter, but DeploymentList, StatefulSetList, ReplicaSetList, DaemonSetList, IngressList, ServiceList, ConfigMapList are missing.

There are a few list based resources such as RoleList and SecretList.

Would it be possible for the install/uninstall order to systematically add the corresponding list resource?
Or otherwise make sure that resources are always added in pairs.

z4ce:
Do you have an example of current behavior and desired behavior?

jmajoor:
The current behavior can result in failures due to dependencies across resources, due to missing dependent resources.

For example if you define a ConfigMapList and have a Deployment that refers to config maps created in that list, then the deployment will fail because the ConfigMapList will be created after the deployment instead of before.

Or with a ServiceList and then having an Ingress definition refering to the service. The ServiceList needs to be created before the Ingress.

In other words the current behavior is that all resources not listed in the kind sorter will be created in no particular order at the end, and this can cause failures with dependencies.

Desired behavior is for all List variants of resources to be defined at the same level in the kind sorter as their non-list equivalent, to avoid deployment failures due to cross dependencies.

jmajoor:
Can someone review PR #12915, that addresses this issue?

PR #12915, opened 2024-03-28
Support List based resources in the KindSortOrder. 

jmajoor:
Fix for #12896 

The KindSortOrder maintains 2 lists for InstallOrder and UninstallOrder, which are important to make sure that resources that have dependencies are (un)installed in the correct order without missing their dependencies.

Some List based resources are present, such as SecretList, RoleList, but many are missing such as StatefulList or IngressList.

The PR is needed so you can use List based resources without dependency problems.

The solution here is to automatically create a List entry for each entry of a resource in the InstallOrder.
This will make sure that List equivalents will automatically be supported in the future if new resources are to be added.
Existing List resources have been removed as those are now automatically injected.

The solution also removed the separate Uninstall list and this is now automatically the reverse order of the install list.

**If applicable**:
- This PR contains unit tests
   The existing KindSortOrder tests are sufficient. Those are testing several List based resources and those test still pass even though the InstallOrder doesn't register List resources explicitly anymore.
   The Uninstall test needed a modification as the Service resource was uninstalled before a Statefulset, but Statefulset have a dependency on a headless service and should have been deleted first.

### DIFF
--- a/pkg/releaseutil/kind_sorter.go
+++ b/pkg/releaseutil/kind_sorter.go
@@ -64,7 +64,7 @@ var ResourceDependencyOrder KindSortOrder = []string{
 
 // addListKind adds the List equivalent for each resource kind in the list.
 func addListKind(resources KindSortOrder) KindSortOrder {
-	output := make(KindSortOrder, 2*len(resources))
+	output := make(KindSortOrder, 0, 2*len(resources))
 	for _, resource := range resources {
 		output = append(output, resource)
 		output = append(output, resource+"List")
--- a/pkg/releaseutil/kind_sorter.go
+++ b/pkg/releaseutil/kind_sorter.go
@@ -25,10 +25,10 @@ import (
 // KindSortOrder is an ordering of Kinds.
 type KindSortOrder []string
 
-// InstallOrder is the order in which manifests should be installed (by Kind).
+// ResourceDependencyOrder is the order in which manifests should be installed (by Kind).
 //
 // Those occurring earlier in the list get installed before those occurring later in the list.
-var InstallOrder KindSortOrder = []string{
+var ResourceDependencyOrder KindSortOrder = []string{
 	"PriorityClass",
 	"Namespace",
 	"NetworkPolicy",
@@ -38,20 +38,15 @@ var InstallOrder KindSortOrder = []string{
 	"PodDisruptionBudget",
 	"ServiceAccount",
 	"Secret",
-	"SecretList",
 	"ConfigMap",
 	"StorageClass",
 	"PersistentVolume",
 	"PersistentVolumeClaim",
 	"CustomResourceDefinition",
 	"ClusterRole",
-	"ClusterRoleList",
 	"ClusterRoleBinding",
-	"ClusterRoleBindingList",
 	"Role",
-	"RoleList",
 	"RoleBinding",
-	"RoleBindingList",
 	"Service",
 	"DaemonSet",
 	"Pod",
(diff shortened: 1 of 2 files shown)


pkg/releaseutil/kind_sorter.go
```diff
@@ -67,47 +62,34 @@ var InstallOrder KindSortOrder = []string{
 	"APIService",
 }
 
+// addListKind adds the List equivalent for each resource kind in the list.
+func addListKind(resources KindSortOrder) KindSortOrder {
+	output := make(KindSortOrder, 2*len(resources))
```
luxurine, line 67: try `output := make(KindSortOrder, 0, 2*len(resources))` if use `append`

jmajoor:
Changed the usage of "make" to properly specify the length.

pkg/releaseutil/kind_sorter.go
```diff
@@ -67,47 +62,34 @@ var InstallOrder KindSortOrder = []string{
...
+	output := make(KindSortOrder, 0, 2*len(resources))
+	for _, resource := range resources {
+		output = append(output, resource)
+		output = append(output, resource+"List")
+	}
+	return output
+}
+
+// reverseKindOrder reverses the order of the KindSortOrder.
+func reverseKindOrder(resources KindSortOrder) KindSortOrder {
```
gjenkins8, line 76: fwiw, this would be `slices.Reverse(resources)` now (https://pkg.go.dev/slices#Reverse)

pkg/releaseutil/kind_sorter.go
```diff
@@ -67,47 +62,34 @@ var InstallOrder KindSortOrder = []string{
...
-	"ServiceAccount",
-	"PodDisruptionBudget",
-	"PodSecurityPolicy",
-	"LimitRange",
-	"ResourceQuota",
-	"NetworkPolicy",
-	"Namespace",
-	"PriorityClass",
-}
+var UninstallOrder = reverseKindOrder(addListKind(ResourceDependencyOrder))
```
gjenkins8, line 92: `UninstallOrder` isn't the strict reverse of `InstallOrder`

pkg/releaseutil/kind_sorter.go
```diff
@@ -25,10 +25,10 @@ import (
 // KindSortOrder is an ordering of Kinds.
 type KindSortOrder []string
 
-// InstallOrder is the order in which manifests should be installed (by Kind).
+// ResourceDependencyOrder is the order in which manifests should be installed (by Kind).
 //
 // Those occurring earlier in the list get installed before those occurring later in the list.
```
gjenkins8, line 30:
```suggestion
// Those occurring earlier, or their corresponding "*List" types (e.g.  "PriorityClassList"), get installed before those occurring later in the list.
```

pkg/releaseutil/kind_sorter.go
```diff
@@ -67,47 +62,34 @@ var InstallOrder KindSortOrder = []string{
 	"APIService",
 }
 
+// addListKind adds the List equivalent for each resource kind in the list.
+func addListKind(resources KindSortOrder) KindSortOrder {
```
gjenkins8, line 66: Is it guaranteed that all Kubernetes tyes have a corresponding "List" type? Offhand, I would guess it is. But there may be exceptions?

And does it actually matter if we generate an "invalid" *List type here?

gjenkins8 (changes requested):
Thanks for the PR! But I have a major concern: This PR does two things:

1. Generates "*List" types for existing `InstallOrder` types
2. Makes `UninstallOrder` the exact reverse of `InstallOrder` (but this is not true today)

The second is IMHO unwarranted/undesirable. Can we separate these two changes into separate PRs please (and/or drop the second)

pkg/releaseutil/kind_sorter.go
jmajoor: Different versions of Kubernetes support different versions of the K8S resources. Older versions may not necessarily support list types for all of them.
However, this doesn't matter. Whether a list type is used or not is up to the author of a Helm template. 

All that is done in this part of the code is to make sure that templates are ordered correctly. To make sure that templates / resources that have dependencies are applied in the correct order for Kubernetes to resolve those dependencies.
Without this part Kubernetes may fail to apply the templates.

The core part of this code has been around since the beginning of Helm. Every time someone runs into an ordering problem this has been updated. Over time some list types were added as needed. Since I needed a StatefulsetList, I decided to just automatically add list types to all registered types, so we don't have to do this one by one whenever someone runs into a problem.

pkg/releaseutil/kind_sorter.go
jmajoor: In its current implementation the uninstall order is different in one aspect. The order of the uninstall of the Service resource. However, that seems to be incorrect. A statefulset has a dependency on a headless service.
The install order therefore creates the Service resource before the statefulset. It should however remove the statefulset before the dependent Service resource upon uninstall and doesn't do that.  This issue is now corrected as well by this PR.

PR #12917, opened 2024-03-29
Implement helm.sh/resource-deletion-policy annotation to override deletion cascade per resource

Danil-Grigorev:
Helm uninstall command allows to provider an optional `--cascade` flag to specify a deletion policy to use for all resources. In some cases this policy has to be `foreground`, to ensure the ownership resources are deleted in order, and the deletion is finished until the next hook phase is executed.

To achieve this, a new annotation is provided, `helm.sh/resource-deletion-policy`. When it is set on the resource, the policy from the value is applied on delete request, regardless of the `--cascade` flag value.

**If applicable**:
(backwards compatibility is ensured as it is just adding an annotation. Previous releases would ignore it)

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -484,7 +484,15 @@ func rdelete(c *Client, resources ResourceList, propagation metav1.DeletionPropa
 	mtx := sync.Mutex{}
 	err := perform(resources, func(info *resource.Info) error {
 		c.Log("Starting delete for %q %s", info.Name, info.Mapping.GroupVersionKind.Kind)
-		err := deleteResource(info, propagation)
+		propagationPolicy := propagation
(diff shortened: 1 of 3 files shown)


robertsirc:
Not to be difficult but shouldn't this have some testing?

Danil-Grigorev:
Hi @robertsirc yeah, tests sound good. Just added some, can this be `ok-to-test` now? It is passing for me locally.

robertsirc:
Hey @Danil-Grigorev this is a great question if you could join us for the weekly developer meeting on Thursday and ask one of their maintainers. I was just helping read through these and as i looked at all the code you added I wanted to question if it made sense for us to write up some test as we could ensure we are getting the functionality you added.

PR #12924, opened 2024-04-04
Drop unused field.

porridge:

it seems that `manifestFile.apis` is a write-only field, and can safely be removed.

The `Capabilities.APIVersions` field, while no longer used on its own by the Helm code directly, remains exposed to helm charts as part of the `.Capabilities` struct, which may well depend on it, so it must stay untouched.

Introducing new versions of functions with one less argument seems like an overkill, so I just replaced the argument names with an underscore.

**If applicable**:


robertsirc (approved):
Not sure if the maintainers plan on using this in the future or not. Comments do denote it isn't used. But this looks good to me.

porridge:
Thanks for taking a look @robertsirc !

Removed one questionable comment line that I had added, and explained the situation a bit more in the PR description.

mattfarina (approved):

joejulian (approved):

PR #12971, opened 2024-04-21
Update testdata PKI with keys that have validity until 3393 (closes #12880)

### LINKED ISSUE #12880: Tests fail after 2024-06-07
Output of `helm version`: 3.14.2

While working on [reproducible builds](https://reproducible-builds.org/) for [openSUSE](https://en.opensuse.org/openSUSE:Reproducible_Builds), I found that our `helm-3.14.2` package stops building soon, because tests fail after 2024-06-07.

Background:
As part of my work on reproducible builds for openSUSE, I check that software still gives identical build results in the future.
The usual offset is +16 years, because that is how long I expect some software will be used in some places.
This showed up failing tests in our package build.
See https://reproducible-builds.org/ for why this matters.

In this case, https://github.com/helm/helm/tree/main/pkg/registry/testdata/tls contains expiring .crt files.
There is also `testdata/crt.pem` that will expire in 2032.

There are at least two possible approaches to solve this:
* generate test .crt files at time of test, so they are always fresh
* use static test .crt files that expire in the far future, e.g. in 2999.

dirkmueller:
The ca.crt had to be regenerated because there was no ca.key. Added new ca.key so that going forward only the certs need to be updated.

replaces the testing PKI with one that has extended validity (50000 days, or until 3393). 

**If applicable**:


mattfarina (approved):

sabre1041 (approved):
LGTM

PR #12310, opened 2023-08-16
chore(oci): migrate to ORAS Golang library v2

zregvart:
**What this PR does / why we need it**:

Updates from `oras.land/oras-go` to `oras.land/oras-go/v2`. The main user facing change is that the `--plain-http` parameter is now required even when accessing OCI registries running on `localhost`.

Fixes #11821

**Special notes for your reviewer**:

I'm new to this codebase and some tests needed to be adjusted to match the new dependency, please look over the test changes to see if there are any issues with those.

**If applicable**:
- [ ] this PR contains documentation
- [x] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

PR #12998, opened 2024-05-02
bump oras minor version

AustinAbro321:
This is a minor update to the oras version while we wait for the major oras v2 upgrade. This update matters because the current auth mechanism in oras breaks when logging into an insecure repo unless an IP address is used. Oras v1.2.5 updates to use docker v25.0.1 for auth mechanisms which avoids this issue 

I can give the whole stack trace is wanted, but ultimately it boils down ORAS v1.2.4 calling functions which eventually use this faulty regex to determine if the domain is correct https://github.com/distribution/distribution/blob/v2.8.3/reference/regexp_deprecated.go#L12-L23

**If applicable**:


sabre1041 (approved):
@AustinAbro321 can you resolve the conflict?

sabre1041:
@AustinAbro321 can you resolve the conflicting files?

AustinAbro321:
@sabre1041 done

sabre1041 (approved):
LGTM

scottrigby (approved):
lgtm 👍 straightforward bump to [v1.2.5](https://github.com/oras-project/oras-go/releases/tag/v1.2.5). Although I'm very much looking forward to upgrading to oras v2 when/if we can (#12310).

AustinAbro321:
@scottrigby @sabre1041 Do you think you could get this into v3.15.2? It would solve a relatively big bug for my team and I

PR #13015, opened 2024-05-08
chore: Remove log message about resolving symlink.

connyay:
This log message is very spammy when building many helm charts.

This pr removes a log message about resolving symlinks.

**If applicable**:

### DIFF
--- a/internal/sympath/walk.go
+++ b/internal/sympath/walk.go
@@ -21,7 +21,6 @@ limitations under the License.
 package sympath
 
 import (
-	"log"
 	"os"
 	"path/filepath"
 	"sort"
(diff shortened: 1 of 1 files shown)


connyay:
I have a pretty atypical build setup (bazel + helm templates), but in our mono repo that produces ~45 helm charts this log message prints out thousands of times.

image

mattfarina:
The reason for the logging is due to accidental disclosure. Imaging someone creates a chart with a symlink to something in `/etc` and then the user of that chart sends the contents into a malicious container that sends the data somewhere. All without the person running helm knowing about the symlink. symlinks can be dangerous.

Symlink handling has been part of the discussion in our security audits because of the danger of misuse.

I understand the atypical setup you have, @connyay. Any idea how to inform Helm users (who are often not chart creators) without filling your logs?

mattfarina (changes requested):
We discussed this in the Helm developer call for May 23, 2024. The consensus was that we would be interested in a general log filtering mechanism instead of this specific one. If we have a flag to filter this one log message than others will ask to filter others with flags. We would want a general filtering capability.

It would essentially be like `helm <commands and arguments> | grep -v "<log message to filter>"`

Given that this is notification is there for security and possible abuses, we have concerns around a general flag that could easily be picked up in copy/paste commands for use with Helm.

Helm has [documented audiences](https://github.com/helm/community/blob/main/user-profiles.md) and the application operator (i.e. chart user) is the highest priority user. Most often, these users are getting their charts from someone else.

PR #13030, opened 2024-05-12
Updated validate-license.sh

Saahiti402:
The modified code has following features:
1.Readability
2.Syntax simplicity
3.Flexibility

The modified code has proper documentation and has no bugs according to my knowledge

### DIFF
--- a/scripts/validate-license.sh
+++ b/scripts/validate-license.sh
@@ -15,18 +15,13 @@
 # limitations under the License.
 set -euo pipefail
 IFS=$'\n\t'
-
+# removed -not to reduce complexity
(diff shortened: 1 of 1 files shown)


scripts/validate-license.sh
```diff
@@ -15,18 +15,13 @@
 # limitations under the License.
 set -euo pipefail
 IFS=$'\n\t'
-
+# removed -not to reduce complexity
```
sabre1041, line 18: This line can probably be removed since it is only applicable right now since `-not` is part of the existing `validate-license.sh` script.

sabre1041 (changes requested):
This looks good. One small request to remove one of the comment lines

PR #13049, opened 2024-05-20
Adjust name validation in pkg/chart/metadata to use filepath.IsLocal

### LINKED ISSUE #12868: Helm 3.14.1+ flags charts as invalid due to validation error on chart.metadata.name
## Description

When using helm `3.14.1` or later, running `helm repo update` gives this error / warning message:

```
index.go:370: skipping loading invalid entry for chart "bitnami/kibana" "10.13.0" from https://artifact.corp.org/artifactory/helm: validation: chart.metadata.name "bitnami/kibana" is invalid
```

We construct our own `index.yaml` and the entry for `bitnami/kibana` for the requested version is:

```yaml
apiVersion: v1
entries:
  ...
  bitnami/kibana:
  - description: "Kibana is an open source, browser based analytics and search dashboard for Elasticsearch. Kibana strives to be easy to get started with, while also being flexible and powerful."
    digest: 4cc5731d05a0a38f1bbc7d8f89cfac8a511534240eaf06c374dfdba7662dc787
    name: bitnami/kibana
    urls:
    - https://artifact.corp.org/artifactory/api/helm/helm/bitnami/kibana-10.13.0.tgz
    appVersion: "8.12.2"
    version: "10.13.0"
  ...
```

`helm search repo` and `helm show values` for the chart continue to work, returning the correct search result for the chart as well as displaying values respectively.

When testing with helm `3.14.0`, there is no error / warning message.

## Additional information

Output of `helm version`: Tested against `3.14.0`, `3.14.1` and `3.14.2`

Output of `kubectl version`: NA

Cloud Provider/Platform (AKS, GKE, Minikube etc.): NA

astraldawn:
**What this PR does / why we need it**: Closes #12868 

**If applicable**:


robertsirc (approved):
LGTM: pulled the branch built it and ran all the unit test and my smoke tests worked fine. But since this handles namespace I want to ask a core maintainer if this is the correct place for it.

pkg/chart/metadata.go
```diff
@@ -112,7 +112,7 @@ func (md *Metadata) Validate() error {
 		return ValidationError("chart.metadata.name is required")
 	}
 
-	if md.Name != filepath.Base(md.Name) {
+	if !filepath.IsLocal(md.Name) {
```
mattfarina, line 115: The [documentation has some details on what is allowed and not allowed in a name](https://helm.sh/docs/chart_best_practices/conventions/#chart-names). For example, a `.` is not allowed in a name. But, `filepath.IsLocal` returns true when a `.` is in the name. So, it's not good to use for this kind of validation.

mattfarina (changes requested):
`/` in the chart name is something we need to talk about. There are docs on the name at https://helm.sh/docs/chart_best_practices/conventions/#chart-names. `/` is not covered but the allowed characters are covered and it's not included. I understand the use case and what artifactory did. I added it to the agenda for this weeks developer call.

Even if this is allowed, the current PR code mishandles `.` which is not allowed.

mattfarina (changes requested):
I've been looking into this one. Some things I've learned...

* Running `helm create foo/bar` (i.e. a name with a `/` in it) will return an error when it tries to create a directory.
* The chart helpers created by default (and the same style help functions widely used) will generate manifest names with `/` in them which Kubernetes will reject.
* In the `charts` directory within a Chart, where archives are stored, we run into a couple problems...
  * If the chart is nested in one or more subdirectories then Helm will return an error and not use the nested chart.
  * If you have two charts with the same base name (e.g., `foo/bar` and `baz/bar`) with them both being stored in the root of the `charts` directory there will be a problem with only one of them being present.

I think this worked for artifactory because the listing in `index.yaml` was namespaced but not the name in the chart itself. Also, people are only using one chart with the same basename.

If we want to just do the name in the `index.yaml` file than we should do a custom validation function when loading that file and not affect what's in the `Chart.yaml` file. We should also document that the base names cannot be the same.

astraldawn:
> I think this worked for artifactory because the listing in index.yaml was namespaced but not the name in the chart itself. Also, people are only using one chart with the same basename. If we want to just do the name in the index.yaml file than we should do a custom validation function when loading that file and not affect what's in the Chart.yaml file.

Agree, will adjust the MR accordingly.

> Even if this is allowed, the current PR code mishandles . which is not allowed.

The convention page does not explicitly disallow dots ("should not be used"), so the docs would need an update as well.

astraldawn:
@mattfarina Based on https://helm.sh/docs/chart_best_practices/conventions/#chart-names

```
1. Chart names must be lower case letters and numbers.
2. Words may be separated with dashes (-)
3. Neither uppercase letters nor underscores can be used in chart names.
4. Dots should not be used in chart names.
```

If 4. (should) is not allowed, then chart.metadata.name validation must enforce 1. (must) and 3. (can).

This leads to multiple failures when running `make test-unit` due to chart names containing upper case letters, underscores and dots, suggesting a need to get consensus on whether best practices for naming should be enforced.

PR #13076, opened 2024-05-29
bug(lint): Ensure quiet flag supresses INFO messages

### LINKED ISSUE #13000: Helm lint "--quiet" flag fix
The issue is pretty straightforward, I want to use the `helm lint --quiet .` command to check for errors during CI/CD.
But I got a bunch of [INFO] Errors which are not of my interest.
```
engine.go:206: [INFO] Missing required value: .Values.deployment.stage is required.
engine.go:206: [INFO] Missing required value: .Values.license_id is required.
```
To remove them, the `helm lint --help` output states:
> --quiet                     print only warnings and errors

Which is not happening when I pass the `--quiet` flag. (I assume INFO is not ERROR nor WARNING).
I can bring a PR for this little change, either fix the documentation for the --help or (ideally) supress the INFO messages when the `--quiet` flag is set.

I am willing to provide any help or more information of any kind! Thanks in advance :) 

Output of `helm version`:
```sh
version.BuildInfo{Version:"v3.14.2", GitCommit:"c309b6f0ff63856811846ce18f3bdc93d2b4d54b", GitTreeState:"clean", GoVersion:"go1.21.7"}
```
Output of `kubectl version`:
```sh
Client Version: v1.28.2
```
Doesn't apply

blueprismo:
While I was using `helm lint --quiet mychart`, I saw that there were some INFO messages that shouldn't appear as the `--quiet` flag states: 
> --quiet                      print only warnings and errors

This PR makes the quiet flag to be respected.
The whole rendering of templates needs a refactor IMO, but I made the changes as minimal as possible to avoid any disrupting change :) 

If there's no rush, I'd like to make a common effort over the templating refactor. I might take some while to build all the context, but helm is a great project and quite widely used! 

**If applicable**:
- [n/a] this PR contains documentation
- [n/a] this PR contains unit tests
- [n/a] this PR has been tested for backwards compatibility

closes #13000

### DIFF
--- a/pkg/lint/lint.go
+++ b/pkg/lint/lint.go
@@ -25,12 +25,17 @@ import (
 )
 
 // All runs all of the available linters on the given base directory.
-func All(basedir string, values map[string]interface{}, namespace string, quiet bool) support.Linter {
-	return AllWithKubeVersion(basedir, values, namespace, nil, quiet)
+func All(basedir string, values map[string]interface{}, namespace string, _ bool) support.Linter {
+	return AllWithKubeVersion(basedir, values, namespace, nil)
 }
 
 // AllWithKubeVersion runs all the available linters on the given base directory, allowing to specify the kubernetes version.
-func AllWithKubeVersion(basedir string, values map[string]interface{}, namespace string, kubeVersion *chartutil.KubeVersion, quiet bool) support.Linter {
+func AllWithKubeVersion(basedir string, values map[string]interface{}, namespace string, kubeVersion *chartutil.KubeVersion) support.Linter {
+	return AllWithKubeVersionAndQuiet(basedir, values, namespace, kubeVersion, false)
+}
+
+// AllWithKubeVersionAndQuiet allows the quiet flag to be passed to the linter
+func AllWithKubeVersionAndQuiet(basedir string, values map[string]interface{}, namespace string, kubeVersion *chartutil.KubeVersion, quiet bool) support.Linter {
 	// Using abs path to get directory context
 	chartDir, _ := filepath.Abs(basedir)
 
--- a/pkg/lint/lint.go
+++ b/pkg/lint/lint.go
@@ -25,19 +25,19 @@ import (
 )
 
 // All runs all of the available linters on the given base directory.
-func All(basedir string, values map[string]interface{}, namespace string, _ bool) support.Linter {
-	return AllWithKubeVersion(basedir, values, namespace, nil)
+func All(basedir string, values map[string]interface{}, namespace string, quiet bool) support.Linter {
+	return AllWithKubeVersion(basedir, values, namespace, nil, quiet)
 }
 
 // AllWithKubeVersion runs all the available linters on the given base directory, allowing to specify the kubernetes version.
-func AllWithKubeVersion(basedir string, values map[string]interface{}, namespace string, kubeVersion *chartutil.KubeVersion) support.Linter {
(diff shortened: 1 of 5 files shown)


robertsirc (changes requested):
This PR is lacking testing, I think you need to fix the breaking test (see below). Also you have a typo 
```
pkg/engine/engine.go:189:6: `Supress` is a misspelling of `Suppress` (misspell)
		// Supress INFO messages if quiet mode is enabled
```.
Test error:
`pkg/action/action.go:1: : # helm.sh/helm/v3/pkg/action [helm.sh/helm/v3/pkg/action.test]
pkg/action/lint_test.go:80:75: not enough arguments in call to lintChart
	have (string, map[string]interface{}, string, nil)
	want (string, map[string]interface{}, string, *chartutil.KubeVersion, bool) (typecheck)`

File: pkg/lint/lint.go [hunk 1]
```diff
@@ -25,19 +25,19 @@ import (
...
 // All runs all of the available linters on the given base directory.
-func All(basedir string, values map[string]interface{}, namespace string, _ bool) support.Linter {
-	return AllWithKubeVersion(basedir, values, namespace, nil)
+func All(basedir string, values map[string]interface{}, namespace string, quiet bool) support.Linter {
+	return AllWithKubeVersion(basedir, values, namespace, nil, quiet)
 }
 
 // AllWithKubeVersion runs all the available linters on the given base directory, allowing to specify the kubernetes version.
-func AllWithKubeVersion(basedir string, values map[string]interface{}, namespace string, kubeVersion *chartutil.KubeVersion) support.Linter {
+func AllWithKubeVersion(basedir string, values map[string]interface{}, namespace string, kubeVersion *chartutil.KubeVersion, quiet bool) support.Linter {
```
mattfarina, line 33: This is a public function and the arguments are changing. This can only happen when the major version of Helm is incremented for [backwards compatability](https://github.com/helm/community/blob/main/hips/hip-0004.md). Those who use Helm as an SDK would have a broken experience, otherwise.

We normally handle this by having two functions. One with the old signature and behavior and one with the new name and the different signature. Often, the old one will call the new one.

Can you please ensure backwards compatibility.

mattfarina (changes requested):

pkg/lint/lint.go
blueprismo: @mattfarina  Thanks for reaching out,
I've created a new public function and updated it's reference in 8bdcd9d.
I also add a simple test `TestLint_ChartWithInfo` that checks there are no `INFO` messages.

I forgot to sign-off some commits, therefore the force-push, sorry about it!

blueprismo:
> This PR is lacking testing, I think you need to fix the breaking test (see below). Also you have a typo
> 
> ```
> pkg/engine/engine.go:189:6: `Supress` is a misspelling of `Suppress` (misspell)
> 		// Supress INFO messages if quiet mode is enabled
> ```.
> Test error:
> `pkg/action/action.go:1: : # helm.sh/helm/v3/pkg/action [helm.sh/helm/v3/pkg/action.test]
> pkg/action/lint_test.go:80:75: not enough arguments in call to lintChart
> 	have (string, map[string]interface{}, string, nil)
> 	want (string, map[string]interface{}, string, *chartutil.KubeVersion, bool) (typecheck)`
> 
> ```

Thanks @robertsirc, typo fixed and also added the `TestLint_ChartWithInfo` test!

PR #13085, opened 2024-06-01
Fix race condition in TestInstallRelease_Wait_Interrupted test

### LINKED ISSUE #12961: UnitTest Failed in case TestInstallRelease_Wait_Interrupted
Output of `helm version`: 3.14.4

Output of `kubectl version`: no

Cloud Provider/Platform (AKS, GKE, Minikube etc.): local

When I'm run project testcase with race flags, it occurs an error.

cmdline: `go test -race helm.sh/helm/v3/pkg/action`

image

alex-kattathra-johnson:
Fixes #12961

### DIFF
--- a/pkg/action/install_test.go
+++ b/pkg/action/install_test.go
@@ -429,16 +429,14 @@ func TestInstallRelease_Wait_Interrupted(t *testing.T) {
 	instAction.Wait = true
 	vals := map[string]interface{}{}
 
-	ctx := context.Background()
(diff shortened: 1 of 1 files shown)


mattfarina (approved):
This makes sense due to the race condition of getting a response to be checked and timing out.

mattfarina (approved):

joejulian (approved):

PR #13086, opened 2024-06-03
Allow index value override for values file

### LINKED ISSUE #13075: Enhancement: Allow deep overriding of Values in a list using `-f values.yaml` the way it works with `--set`
There's a strange asymmetry since Helm 2.5.0: one can override individual values that are part of `Values` inside lists with the syntax `--set servers[0].port=8080` but one cannot do the equivalent override using `-f override_values.yaml` overrides, i.e. the following does not work:

```
# values.yaml
servers:
- protocol: tcp
  port: 80
```

```
# override_values.yaml
servers[0]:
  port: 8080
```

It's unintuitive that one can do deep overrides in lists using indexes via `--set` syntax but not with `-f` overrides.

In fact, based on this [StackOverflow Q&A with 5k views](https://stackoverflow.com/questions/63833877/override-value-in-list-with-helm-override-file), indexed `-f` override seemed to work at some point in the past, maybe with v2, but it doesn't seem to work anymore.

My questions are:
- Is this intended behaviour, that one cannot use `-f` for deep overrides, the way one can do it with `--set`?
- When was the behaviour changed? Per StackOverflow Q&A it used to be possible (albeit with +/- 1 differences in index)
- Would it be possible to symmetrise `--set` and `-f`, by adding the deep override capability to `-f` overrides?

I looked carefully through previous issues, but couldn't find a recent one that was quite as explicit - there may well have been bugs/changes over time so I'm not sure old issues apply one-to-one.

Here are a few I looked at:
- https://github.com/helm/helm/issues/5711
- https://github.com/helm/helm/issues/11734

miles-w-3:
Closes #13075 

This feature allows for elements of list values to be overridden by new syntax. For example:

```yml
# incoming
header:
   foo[0]:
     newKey: value
```

```yml
# base valyes yaml
header:
  foo:
  - original: sample
  - next: item
```

```yml
# resulting values
header:
  foo:
  - newKey: value
  - next: item
```

This is *separate* functionality from the way `--set` allows overrides. That function will override an index with the provided value, but set the rest of the items in the list to null. This functionality will leave the other items in the list untouched.

Restrictions:
- Nested list merges are not possible, meaning that while the rest of the list is untouched, a map at index 0 being overridden will not be merged. As shown above, the `original:sample` key/valye disappears.
- The provided index must be valid within the underlying list (not out of bounds, not invalid spec like `[]`)
- the base key must also be a list
- if both an index override key and the standard key, ie `foo[0]` and `foo` are provided within the same layer,  the index override key `foo[0]` will be ignored

There are two main sites for implementation. The value merge functionality used to collpase values files will error out if any of the restrictions are hit. The coalesce functionality used to do the final merge of passed in values to the chart's base `values.yaml` does not currently contain functionality to error, and therefore will skip any indexes when the restrictions are hit, logging an error through the `printf` function.

**If applicable**:
- [x] this PR contains documentation: https://github.com/helm/helm-www/pull/1625

### DIFF
--- a/cmd/helm/flags.go
+++ b/cmd/helm/flags.go
@@ -49,6 +49,7 @@ func addValueOptionsFlags(f *pflag.FlagSet, v *values.Options) {
 	f.StringArrayVar(&v.FileValues, "set-file", []string{}, "set values from respective files specified via the command line (can specify multiple or separate values with commas: key1=path1,key2=path2)")
 	f.StringArrayVar(&v.JSONValues, "set-json", []string{}, "set JSON values on the command line (can specify multiple or separate values with commas: key1=jsonval1,key2=jsonval2)")
 	f.StringArrayVar(&v.LiteralValues, "set-literal", []string{}, "set a literal STRING value on the command line")
+	f.BoolVar(&v.IndexOverride, "index-override", false, "if set, allow keys in value files to override list indexes similar to --set syntax")
 }
 
 func addChartPathOptionsFlags(f *pflag.FlagSet, c *action.ChartPathOptions) {
--- a/cmd/helm/install.go
+++ b/cmd/helm/install.go
@@ -318,6 +318,7 @@ func runInstall(args []string, client *action.Install, valueOpts *values.Options
 		cancel()
 	}()
 
+	chartRequested.IndexOverride = valueOpts.IndexOverride
 	return client.RunWithContext(ctx, chartRequested, vals)
 }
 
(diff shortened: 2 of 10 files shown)


robertsirc:
Please fix your DCO, so we can review please.

miles-w-3:
@robertsirc Apologies for the delay, I have pushed up new changes that gate this feature behind an `--index-override` flag based on feedback from @gjenkins8 . I am working on the helm-www documentation PR as well and hope to have that up by the end of the weekend

PR #13098, opened 2024-06-11
Fix: Helm chart is removed even if pre-delete hook is failing

### LINKED ISSUE #9700: Helm pre-delete hooks Remove Chart Metadata on Failure
Output of `helm version`: 

```
version.BuildInfo{Version:"v3.5.4", GitCommit:"1b5edb69df3d3a08df77c9902dc17af864ff05d1", GitTreeState:"clean", GoVersion:"go1.15.11"}
```

Output of `kubectl version`:

```
Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.1", GitCommit:"5e58841cce77d4bc13713ad2b91fa0d961e69192", GitTreeState:"clean", BuildDate:"2021-05-12T14:18:45Z", GoVersion:"go1.16.4", Compiler:"gc", Platform:"w
indows/amd64"}
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.1", GitCommit:"206bcadf021e76c27513500ca24182692aabd17e", GitTreeState:"clean", BuildDate:"2020-09-14T07:30:52Z", GoVersion:"go1.15", Compiler:"gc", Platform:"lin
ux/amd64"}
```

Cloud Provider/Platform (AKS, GKE, Minikube etc.): kind Cluster

In order to ensure that the cluster is in a good state to delete the helm release, I am onboarding a pre-delete hook to check the state before actually deleting the release. This works fine and times out for the resources themselves, but the chart release metadata is gone when running `helm ls -A`.

image

Because of this behavior, if the initial delete fails, I am no longer able to remove the resource through a `helm uninstall` because the chart release itself has disappeared.

My pre-delete hook is below:

```
---
apiVersion: batch/v1
kind: Job
metadata:
  name: {{ .Release.Name }}-pre-delete
  namespace: {{ .Release.Namespace }}
  labels:
  annotations:
    "helm.sh/hook": pre-delete
    "helm.sh/hook-weight": "0"
    "helm.sh/hook-delete-policy": hook-succeeded,hook-failed
spec:
  template:
    metadata:
      name: {{ .Release.Name }}
      namespace: {{ .Release.Namespace }}
      labels:
    spec:
      serviceAccountName: flux-operatorsa
      restartPolicy: Never
      containers:
        - name: pre-delete-job
          image: mcr.microsoft.com/oss/kubernetes/kubectl:v1.21.0
          command:
            - 'sh'
            - '-c'
            - '/scripts/pre-delete.sh'
          volumeMounts:
            - name: pre-install-script
              mountPath: "/scripts"
      volumes:
        - name: pre-install-script
          configMap:
            defaultMode: 0777
            name: flux-extension-hooks
---
apiVersion: v1
kind: ConfigMap
metadata:
  annotations:
    "helm.sh/hook": pre-delete
    "helm.sh/hook-weight": "-1"
    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded,hook-failed
  name: flux-extension-hooks
  labels:
    chart: {{ .Chart.Name }}-{{ .Chart.Version }}
    heritage: {{ .Release.Service }}
    release: {{ .Release.Name }}
data:
  pre-delete.sh: |
    #!/bin/bash

    export HAS_FLUX_CONFIG_CRD=$(kubectl get crd fluxconfigs.clusterconfig.azure.com -o json)
    if [ ! -z "$HAS_FLUX_CONFIG_CRD" ]
    then
      echo "Found the flux config CRD on the system"
      export FLUX_CONFIGS=$(kubectl get fluxconfigs -A -o json)
      if [ ! -z "$FLUX_CONFIGS" ]
      then
        echo "Flux configurations still remain on the cluster, cleanup of these instances must be done before deletion of extension..."
        exit 1
      fi
    fi
    echo "No more flux configuration instances found on the cluster, deletion is proceeding..."
```

BarrMan:
closes #9700

Moved release status, deleted and description update statements to after the pre-delete hooks have succeeded. According to this comment in the code:
```
// From here on out, the release is currently considered to be in StatusUninstalling
// state.
```
It looks pretty promising that the uninstalling status should be changed only after the hooks have succeeded.

**If applicable**:

### DIFF
--- a/pkg/action/hooks.go
+++ b/pkg/action/hooks.go
@@ -97,8 +97,6 @@ func (cfg *Configuration) execHook(rl *release.Release, hook release.HookEvent, 
 		h.LastRun.Phase = release.HookPhaseSucceeded
 	}
 
-	cfg.recordRelease(rl)
-
 	// If all hooks are successful, check the annotation of each hook to determine whether the hook should be deleted
 	// under succeeded condition. If so, then clear the corresponding resource object in each hook
 	for _, h := range executingHooks {
--- a/pkg/action/hooks.go
+++ b/pkg/action/hooks.go
@@ -66,7 +66,6 @@ func (cfg *Configuration) execHook(rl *release.Release, hook release.HookEvent, 
 			StartedAt: helmtime.Now(),
 			Phase:     release.HookPhaseRunning,
 		}
-		cfg.recordRelease(rl)
 
 		// As long as the implementation of WatchUntilReady does not panic, HookPhaseFailed or HookPhaseSucceeded
 		// should always be set by this function. If we fail to do that for any reason, then HookPhaseUnknown is
@@ -96,6 +95,8 @@ func (cfg *Configuration) execHook(rl *release.Release, hook release.HookEvent, 
 		}
 		h.LastRun.Phase = release.HookPhaseSucceeded
 	}
+
+	cfg.recordRelease(rl)
 
 	// If all hooks are successful, check the annotation of each hook to determine whether the hook should be deleted
 	// under succeeded condition. If so, then clear the corresponding resource object in each hook
--- a/pkg/action/hooks.go
+++ b/pkg/action/hooks.go
@@ -66,6 +66,7 @@ func (cfg *Configuration) execHook(rl *release.Release, hook release.HookEvent, 
 			StartedAt: helmtime.Now(),
 			Phase:     release.HookPhaseRunning,
 		}
+		cfg.recordRelease(rl)
 
 		// As long as the implementation of WatchUntilReady does not panic, HookPhaseFailed or HookPhaseSucceeded
 		// should always be set by this function. If we fail to do that for any reason, then HookPhaseUnknown is
@@ -95,6 +96,8 @@ func (cfg *Configuration) execHook(rl *release.Release, hook release.HookEvent, 
 		}
 		h.LastRun.Phase = release.HookPhaseSucceeded
 	}
+
+	cfg.recordRelease(rl)
 
 	// If all hooks are successful, check the annotation of each hook to determine whether the hook should be deleted
 	// under succeeded condition. If so, then clear the corresponding resource object in each hook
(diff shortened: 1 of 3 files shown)


mattfarina (commented):
What happens if a pre-delete hook partially succeeds and partially fails? The installation could be in a bad state. It's no longer running normally. Shouldn't that be recorded somewhere? If we are going to fix this, and we should, is there a way to handle this situation?

There is a mix around saving releases that is causing this. If I understand it right...
1. The release update is created
2. pre-delete hooks are fired. This causes a save to happen.
3. The save with the release update happens

The code doesn't account for the release saving happening inside of hooks being fired.

I can think of two things that would be useful if an uninstall gets into a bad state:

1. re-run the uninstall process again
2. upgrade the release with changes that will allow for it to be uninstalled

It would be nice to have both of these situations handled. Still poking at this but wanted to give my initial feedback.

BarrMan:
> What happens if a pre-delete hook partially succeeds and partially fails? The installation could be in a bad state. It's no longer running normally

Well, currently if the pre-delete hook fails, the release is already in a bad state. The status of the release is Uninstalling instead of Deployed which prevents users from using helm ls and seeing the release. Only helm ls -a which reveal the true reason or when looking farther more into the secret of the release.
> Shouldn't that be recorded somewhere?

How would you like this to be recorded? Will adding a note in the docs suffice?
> The code doesn't account for the release saving happening inside of hooks being fired.

IMO, saving the state permanently inside the hook is wrong. Maybe I don't really understand the reasoning for this but the execHook function doesn't change the release's state at all which makes it be more logical to handle the persistency layer on the caller's side.
>I can think of two things that would be useful if an uninstall gets into a bad state:
> 
> re-run the uninstall process again

When the uninstallation gets into a bad state, if the helm.sh/hook-delete-policy is set to the correct value, then it can be clear for the user what the needed actions are: before-hook-creation,hook-succeeded, in the pod that was created by the pre-delete hook I verified some stuff before deleting, if the verification failed, I logged the error and kept the pod alive in order for the user to keep investigating what the issue is by looking into the logs. Re-running the uninstall process again seems like something the user should be deciding on and not an automatic process. If something does indeed change between the uninstall executions, then the user might wish to run the uninstall again. But this is really a product's decision.
> upgrade the release with changes that will allow for it to be uninstalled

I'm not really sure what you meant by that.

BarrMan:
Bump

Issue #13101, opened 2024-06-12
Get more debug informations about why and upgrade failed with "release: already exists"


benoittgt:
Hello

We are having random error on `helm upgrade`. The error is simply `UPGRADE FAILED: release: already exists`. While looking at logs it doesn't give a lot's of clues about why exactly it is failing. We are not doing concurrent upgrade when the error is happening (like [here](https://github.com/helm/helm/issues/11058)). I am wondering if `--debug` could provide even more informations about what is happening, for example if the configmap already exist, or a secret.

While looking at the code we can return this error in two places also:
- https://github.com/helm/helm/blob/d1595e701f737f1fb367fb6b528c865e43a759ae/pkg/storage/driver/cfgmaps.go#L156
- https://github.com/helm/helm/blob/d1595e701f737f1fb367fb6b528c865e43a759ae/pkg/storage/driver/secrets.go#L147

At the moment the last information is here https://github.com/helm/helm/blob/66a969e7cc08af2377d055f4e6283c33ee84be33/pkg/action/upgrade.go#L324

Debug logs:
```
history.go:56: [debug] getting history for release app-staging
upgrade.go:144: [debug] preparing upgrade for app-staging
upgrade.go:152: [debug] performing update for app-staging
upgrade.go:324: [debug] creating upgraded release for app-staging
Error: UPGRADE FAILED: release: already exists
helm.go:84: [debug] release: already exists
```

Output of `helm version`: 3.11.1
Output of `kubectl version`:1.25.6

I can work on a pull request if it could be relevant.

idsulik:
@benoittgt hi! thank you for the issue.  
Pushed [PR](https://github.com/helm/helm/pull/13104) with updated logs, it should be clear now.  

In your case, if you don't specify `HELM_DRIVE` env, you should use "secrets" to store release data, so it fails because the secret already exists

benoittgt:
Thanks @idsulik for the answer and the pull request.

benoittgt:
What is weird in this case is... I do not see this log message

https://github.com/helm/helm/blob/v3.11.1/pkg/storage/storage.go#L61

benoittgt:
We probably found the root cause on our side. The Api Server was under provisioned and it seems that api calls where randomly partially failing. We reduce the scope a https://github.com/stakater/Reloader to stop pulling too big payload from Api Server. It reduced the load and we stopped seeing this error. But still I think we should improve the debug output.

PR #13104, opened 2024-06-13
chore(logs): update log messages for better clarity

idsulik:
Fixes: #https://github.com/helm/helm/issues/13101
Updated logs for better clarity

### DIFF
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -385,9 +385,11 @@ func (cfg *Configuration) Init(getter genericclioptions.RESTClientGetter, namesp
 
 	if helmDriver == "" {
 		helmDriver = "secrets"
+		log("init: using default storage driver %s", helmDriver)
+	} else {
+		log("init: using %s storage driver", helmDriver)
 	}
 
-	log("init: using %s storage driver", helmDriver)
 	var store *storage.Storage
 	switch helmDriver {
 	case "secret", "secrets":
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -383,9 +383,14 @@ func (cfg *Configuration) Init(getter genericclioptions.RESTClientGetter, namesp
 		clientFn:  kc.Factory.KubernetesClientSet,
 	}
 
+	if helmDriver == "" {
+		helmDriver = "secrets"
+	}
+
+	log("init: using %s storage driver", helmDriver)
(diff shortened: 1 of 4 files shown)


pkg/action/action.go
```diff
@@ -379,9 +379,14 @@ func (cfg *Configuration) Init(getter genericclioptions.RESTClientGetter, namesp
 		clientFn:  kc.Factory.KubernetesClientSet,
 	}
 
+	if helmDriver == "" {
+		helmDriver = "secrets"
+	}
+
+	log("init: using %s storage driver", helmDriver)
```
benoittgt, line 386: Don't we want instead to say that is was not set so we default at `secrets` ?

pkg/storage/driver/secrets.go
```diff
@@ -161,6 +161,7 @@ func (secrets *Secrets) Create(key string, rls *rspb.Release) error {
 	// push the secret object out into the kubiverse
 	if _, err := secrets.impl.Create(context.Background(), obj, metav1.CreateOptions{}); err != nil {
 		if apierrors.IsAlreadyExists(err) {
+			secrets.Log("create: release %q already exists in secrets", rls.Name)
 			return ErrReleaseExists
 		}
```
marcmillien, line 167: Should we have an else here to log a potential unknown other cases?

pkg/action/action.go
idsulik: good idea, pushed changes

pkg/storage/driver/secrets.go
idsulik: otherwise it returns an error so I think there is no need to log it https://github.com/helm/helm/pull/13104/files#diff-1007a5ca11ecfdc4a0ffaa5b178ad1c03d1a8e1cc625288656b1e5ab25765382R168

benoittgt (approved):

marcmillien (approved):

idsulik:
Hi @mattfarina    ,

I noticed that this pull request has been pending for a while. Is there anything more needed from my side? Please let me know if there are any updates or additional changes required.

Thanks!

idsulik:
Hi @mattfarina ,

I noticed that this pull request has been pending for a while. Is there anything more needed from my side? Please let me know if there are any updates or additional changes required.

Thanks!

PR #13114, opened 2024-06-15
Introduce --disable-fetching-api-versions for install and upgrade

eolatham:
When enabled, during the rendering process, this feature flag will populate the built-in `Capabilities.APIVersions` template object with a default version set instead of fetching the full set from the Kubernetes API server.

The motivation for this is to reduce unnecessary Kubernetes API calls when installing or upgrading charts that don't rely heavily on `Capabilities.APIVersions`. In my experience, actual usage of `Capabilities.APIVersions` is rare, yet fetching the version set on clusters supporting many CRDs is very costly.

This change is backwards-compatible.

Signed-off-by: Eric Latham ericoliverlatham@gmail.com

### DIFF
--- a/pkg/action/uninstall.go
+++ b/pkg/action/uninstall.go
@@ -196,7 +196,7 @@ func joinErrors(errs []error) string {
 // deleteRelease deletes the release and returns list of delete resources and manifests that were kept in the deletion process
 func (u *Uninstall) deleteRelease(rel *release.Release) (kube.ResourceList, string, []error) {
 	var errs []error
-	caps, err := u.cfg.getCapabilities()
+	caps, err := u.cfg.getCapabilities(true)
 	if err != nil {
 		return nil, rel.Manifest, []error{errors.Wrap(err, "could not get apiVersions from Kubernetes")}
 	}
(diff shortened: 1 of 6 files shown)


pkg/action/uninstall.go
```diff
@@ -196,7 +196,7 @@ func joinErrors(errs []error) string {
 // deleteRelease deletes the release and returns list of delete resources and manifests that were kept in the deletion process
 func (u *Uninstall) deleteRelease(rel *release.Release) (kube.ResourceList, string, []error) {
 	var errs []error
-	caps, err := u.cfg.getCapabilities()
+	caps, err := u.cfg.getCapabilities(true)
```
eolatham, line 199: I have opened a separate PR to optimize the uninstall action (it doesn't need capabilities at all):
https://github.com/helm/helm/pull/13113

gjenkins8 (commented):
My personal take, is that it would be better to avoid introducing this flag/functionality if possible. A quick search on github shows this functionality is "often" used in charts: https://github.com/search?q=%22%7B%7B%22+Capabilities.APIVersions&type=code

Can you please quantify "fetching the version set on clusters supporting many CRDs is very costly."? Can we profile the operation (against a cluster with many CRDs)? And is there a way to optimize? And what are the concerns with the cost of the fetch?

The specific concerns I have are:
- If a user is wanting to install a chart which uses `.Capabilities.APIVersions` on a cluster with a large number of CRDs. Then what?
- If a user has installed a chart which doesn't use `.Capabilities.APIVersions`, but then wants to upgrade to a version which does. Then what?
- A new flag here introduces a coupling between the chart user, and the chart author. And an additional complexity for the chart user to be concerned about when using Helm.  Minimizing this complexity (not having this functionality to concern the user) is better.

There are also perhaps alternatives: Could Helm fetch version sets lazily? Or check specific API versions on demand?

Overall, not strictly again this flag (ie. the implementation in this PR). But strongly think the need should be quantified, and alternatives should be considered first.

eolatham:
@gjenkins8 After more consideration of your feedback, I've opened a separate PR as an alternative to this: https://github.com/helm/helm/pull/13146

It implements fetching the version sets lazily, without the need for users to be aware of a chart's usage of `.Capabilities.APIVersions`.

Thanks again for taking a look at this.

eolatham:
Closing in favor of https://github.com/helm/helm/pull/13146

PR #13120, opened 2024-06-16
fix(dependencyBuild): prevent race condition in concurrent helm dependency

### LINKED ISSUE #13110: Concurrent `helm dependency build` fail due to race condition
We're building and templating a lot of Helm charts in parallel, when some of them share a subchart using symlinks it's possible the build fails due to a race condition: Both build try to download to the same location, one of them finishes first and deletes the file and then the second build crashes as the file is missing:

```
Error: Error rendering ops/k8s/ingress-nginx-public/Chart.yaml: Error running 'helm dependency build --repository-config /tmp/tmpnb8ufin2.yaml ops/k8s/ingress-nginx-public/charts/ingress-nginx':
...Successfully got an update from the "f6d59f60" chart repository
Saving 1 charts
Downloading ingress-nginx from repo https://kubernetes.github.io/ingress-nginx
Save error occurred:  could not download https://github.com/kubernetes/ingress-nginx/releases/download/helm-chart-4.10.1/ingress-nginx-4.10.1.tgz: open ops/k8s/ingress-nginx-public/charts/ingress-nginx/tmpcharts/ingress-nginx-4.10.1.tgz3661314739: no such file or directory
Error: could not download https://github.com/kubernetes/ingress-nginx/releases/download/helm-chart-4.10.1/ingress-nginx-4.10.1.tgz: open ops/k8s/ingress-nginx-public/charts/ingress-nginx/tmpcharts/ingress-nginx-4.10.1.tgz3661314739: no such file or directory
```

Output of `helm version`:
>version.BuildInfo{Version:"v3.14.3", GitCommit:"f03cc04caaa8f6d7c3e67cf918929150cf6f3f12", GitTreeState:"clean", GoVersion:"go1.22.1"}

idsulik:
closes #13110

it fixes this issue by adding a suffix to the `tmpcharts` directory so each instance can have its own directory
> building and templating a lot of Helm charts in parallel, when some of them share a subchart using symlinks it's possible the build fails due to a race condition

### DIFF
--- a/pkg/downloader/manager.go
+++ b/pkg/downloader/manager.go
@@ -246,7 +246,7 @@ func (m *Manager) downloadAll(deps []*chart.Dependency) error {
 	}
 
 	destPath := filepath.Join(m.ChartPath, "charts")
-	tmpPath := filepath.Join(m.ChartPath, "tmpcharts")
+	tmpPath := filepath.Join(m.ChartPath, fmt.Sprintf("tmpcharts-%d", os.Getpid()))
 
 	// Check if 'charts' directory is not actually a directory. If it does not exist, create it.
 	if fi, err := os.Stat(destPath); err == nil {
--- a/cmd/helm/dependency_update_test.go
+++ b/cmd/helm/dependency_update_test.go
@@ -200,8 +200,9 @@ func TestDependencyUpdateCmd_DoNotDeleteOldChartsOnError(t *testing.T) {
 		}
 	}
 
-	// Make sure tmpcharts is deleted
-	if _, err := os.Stat(filepath.Join(dir(chartname), "tmpcharts")); !os.IsNotExist(err) {
+	// Make sure tmpcharts-x is deleted
+	tmpPath := filepath.Join(dir(chartname), fmt.Sprintf("tmpcharts-%d", os.Getpid()))
+	if _, err := os.Stat(tmpPath); !os.IsNotExist(err) {
 		t.Fatalf("tmpcharts dir still exists")
 	}
 }


pkg/downloader/manager.go
```diff
@@ -246,7 +246,7 @@ func (m *Manager) downloadAll(deps []*chart.Dependency) error {
 	}
 
 	destPath := filepath.Join(m.ChartPath, "charts")
-	tmpPath := filepath.Join(m.ChartPath, "tmpcharts")
+	tmpPath := filepath.Join(m.ChartPath, fmt.Sprintf("tmpcharts-%d", os.Getpid()))
```
mattfarina, line 249: While this generally works, there are some tests to make sure we don't leave the _tmpcharts_ directory behind. The tests need to be updated to ensure the directory under the new name isn't left behind. You can find these by searching for _tmpcharts_ in the codebase. One example is in the test `TestDependencyUpdateCmd_DoNotDeleteOldChartsOnError`.

Can you please make sure the tests are updated for this change.

mattfarina (changes requested):

pkg/downloader/manager.go
idsulik: Thanks for pointing that out, fixed it

mattfarina (approved):
I tested that the PID across both is consistent, even with `-race` and that this fails when things aren't handled properly. This looks good.

mattfarina:
Since this is XS in size, a single maintainer approval is all that's required to merge.

hanikesn:
Thank you @idsulik!

PR #13135, opened 2024-06-20
Expand kubeclient interface to avoid leaking goroutines in Upgrade/Install RunWithContext when context is cancelled

### LINKED ISSUE #13126: Bug: Using RunWithContext leaks go routines in Wait/WaitWithJobs
Output of `helm version`: v3.15.2 (should be independent of the version)

Output of `kubectl version`:
Independent of kubectl version

Independent of cloud provider.

When using the RunWithContext method for Upgrade/Install while having wait/atomic/waitwithjobs set, a context cancellation does not lead to polling on the ready checks in Wait/WaitWithJobs.

This will not be an issue while using the helm binary, but surfaces when using the packages.

Below is an example of how I extended the kubeclient to handle the issue.
Link -> https://github.com/saisrikark/helm/pull/1
I am propagating the context to have PollUntilContextCancel cancel it on the next poll.

saisrikark:
closes #13126

When using RunWithContext for Upgrade/Install with a Wait/WaitWithJobs/Atomic set it is possible for goroutines to leak/remain dangling when the context is cancelled prematurely.

This is due to these methods not waiting for the helm kubeclient's Wait and WaitWithJobs to complete.

With this PR I am introducing new methods in the kubeclient interface.
WaitWithContext and WaitWithJobsWithContext
This way a context is propagated and goroutines are exited solving dangling resources.

Snippet to try this out

```
package main

import (
	"context"
	"log"
	"time"

	"helm.sh/helm/v3/pkg/action"
)


func main() {

	installAction := action.NewInstall(&action.Configuration{
		// fill this up
		// ensure to use a logger to display the log in the dangling go routine
	})

	installAction.Wait = true
	installAction.Timeout = time.Hour

	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
	defer cancel()

	// fill up the chart and values
	go func() {
		_, err := installAction.RunWithContext(ctx, "chart", "values")
		if err != nil {
			log.Printf("encountered an error %s", err.Error())
		}
	}()

	waitChan := make(chan struct{})
	<-waitChan
}
```

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -413,9 +413,10 @@ func (i *Install) performInstallCtx(ctx context.Context, rel *release.Release, t
 	resultChan := make(chan Msg, 1)
 
 	go func() {
-		rel, err := i.performInstall(rel, toBeAdopted, resources)
+		rel, err := i.performInstall(ctx, rel, toBeAdopted, resources)
 		resultChan <- Msg{rel, err}
 	}()
+
 	select {
 	case <-ctx.Done():
 		err := ctx.Err()
@@ -433,7 +434,7 @@ func (i *Install) isDryRun() bool {
 	return false
 }
 
-func (i *Install) performInstall(rel *release.Release, toBeAdopted kube.ResourceList, resources kube.ResourceList) (*release.Release, error) {
+func (i *Install) performInstall(ctx context.Context, rel *release.Release, toBeAdopted kube.ResourceList, resources kube.ResourceList) (*release.Release, error) {
 	var err error
 	// pre-install hooks
 	if !i.DisableHooks {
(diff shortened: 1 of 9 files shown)


robertsirc:
I am still checking to see if this is a breaking change since the contract has changed. Is there an opportunity of having testing for this change?

saisrikark:
Testcases are a good idea.
Let me try it out.

gjenkins8 (commented):
Wouldn't we want implement e.g. `Wait` in terms of `WaitWithContext` ?

```
func (c *Client) Wait(resources ResourceList, timeout time.Duration) error {
   return WaitWithContext(context.Background(), resources, timeout)
}
```

(and others of course)

gjenkins8 (commented):
Can we please add a comment to the replaced ie. non-context methods, saying to the effect that they are deprecated and due for removal in Helm v4

saisrikark:
Added testcases and addressed comments.
handleContext looks to be used to handle sigterm and sigint.

I wonder what is the expected behaviour when we see these kind of signals.

I feel we should store information about the release (which elucidates the failure reason) and stop other operations.
Will update this PR once I have a better understanding.

PR #13138, opened 2024-06-24
fix: helm push to OCI registry failed when the registry integrate with keycloak

### LINKED ISSUE #12121: helm push to OCI registry failed when the registry integrate with keycloak
Output of `helm version`: v3.12.0

Output of `kubectl version`: v1.25.3

OCI registry version: https://github.com/distribution/distribution 2.8.1

I am using helm3.12.0 to pull push helm chart to OCI registry, the registry integrated with keycloak which means will use token to auth.
I could login and pull chart from OCI registry by `helm registry login` and `helm pull`, but when I try to push chart by `helm push`, it reports an error 
`Error: push access denied, repository does not exist or may require authorization: server message: insufficient_scope: authorization failed`

From the packet capture by using tcpdum, I see the request to get token as below
`GET /auth/realms/master/protocol/docker-v2/auth?scope=repository:mychart:pull&scope=repository:mychart:pull,push&service=registry`
There are two scope to request permissions, one is `scope=repository:mychart:pull`, another is `scope=repository:mychart:pull,push`, this is redundant, so that keycloak will only return pull permission in the token.
So I think it's a bug, only one scope is enough to request permissions

abinch:
fixes #12121

bump github.com/containerd/containerd from 1.7.12 to 1.7.18
Containerd 1.7.16 fixed a issue https://github.com/containerd/containerd/pull/10065, which could fix the issue #12121 on helm


guigomcha:
I can confirm that this fixes #12121 which is extremely inconvenient
image

@CheneyCookie thanks!

guigomcha:
I am not sure if this applies still but it should be considered to merge

benoittgt:
containerd is no longer used in https://github.com/helm/helm/blob/main/go.mod

PR #13146, opened 2024-06-28
Fetch Capabilities.APIVersions lazily

eolatham:
(Alternative to https://github.com/helm/helm/pull/13114)

With this change, `Capabilities.APIVersions` will only be fetched if `APIVersions` is referenced in a chart's templates.

This will reduce unnecessary Kubernetes API calls when installing and upgrading charts that don't use `Capabilities.APIVersions`.

As far as I can tell, `crds/` do not need to be checked because they are not rendered with support for `Capabilities.APIVersions` like `templates/` are.

Note: Since there are various ways to use `Capabilities.APIVersions` in a template, checking only for the substring `Capabilities.APIVersions` would not be safe. `APIVersions` is the most specific substring we can use. This obviously allows some false positives, but still provides an optimization for many charts.

Signed-off-by: Eric Latham [ericoliverlatham@gmail.com](mailto:ericoliverlatham@gmail.com)


merusso:
Nice idea!

This would really help us out. We're having issues with slow Helm chart installs in clusters that have hundreds of CRDs.

merusso (approved):

pkg/action/action_test.go
```diff
@@ -273,6 +274,19 @@ func namedReleaseStub(name string, status release.Status) *release.Release {
 	}
 }
 
+func Test_chartNeedsAPIVersions(t *testing.T) {
```
astellingwerf, line 277: I can think of several false positives and false negatives, like:
```
{{ (.Capabilities).APIVersions }}
```
```
# Warning: .Capabilities.APIVersions should not be used under any circumstances
```

eolatham: @astellingwerf Ah, thanks for pointing out these edge cases.

The function definitely needs to avoid false negatives, but I think some false positives are tolerable.

Maybe it should be updated to:
* Check for just `.APIVersions` instead of `.Capabilities.APIVersions` to avoid false negatives
* Ignore `.APIVersions` if it occurs in a comment (a line beginning with `#`)

That should be possible with a regex.

What do you think?

astellingwerf: The first bullet makes sense, @eolatham, but as you can see in line 2 of my second example, lines starting with `#` don't necessarily imply no actual use of capabilities.

I must admit that I've hardly seen any of the Helm code, but I'm afraid that, unless you have access to the AST of the Chart, it'll be quite a challenge to get a the test just right. I agree with your analysis that false negatives should be prevented.

eolatham: Yeah, I'd want to keep the implementation and testing as simple as possible since this is meant to be a straightforward optimization.

But should references to `.APIVersions` in comments be considered "usage" from this perspective?
I wouldn't think so because comments have no effect on the rendered output, right?

astellingwerf: Depends on the kind of comment. A YAML comment (though quite possibly not impacting the functioning of the rendered app) does impact the output:
```
# {{ .Capabilities.APIVersions | toJson }}
```
whereas this, with the "Helm code" commented out, doesn't:
```
{{/*{{ .Capabilities.APIVersions | toJson }}*/}}
```

eolatham: Right.

So I guess, if we want to consider "usage of `.APIVersions`" as any reference that alters the rendered output of the templates (including references that have no functional app impact, as you pointed out), we would have to include references to `.APIVersions` in comments as usage.

In that case, maybe the best solution is to simply check if templates contain `.APIVersions`, and avoid adding any other complexity to the logic. I think we can tolerate false positives from comments and other non-functional references since those are probably rare, and this change would still be a nice optimization for many charts.

If you agree, I'll go ahead and update the PR.

astellingwerf: I'm not an expert in any way (I just stumbled upon this PR because I was facing different issues with `helm template --api-versions` myself), but that makes sense.

astellingwerf: ...or do we wanna support an edge case like:
```
{{ get (.Capabilities | toJson | fromJson) "APIVersions" | has "batch/v1" }}
```

eolatham: Nice catch, I think we have to support that too. I just updated the PR.

pkg/action/action.go
```diff
@@ -97,6 +97,19 @@ type Configuration struct {
 	Log func(string, ...interface{})
 }
 
+// chartNeedsAPIVersions returns true if any of the Chart's templates reference "APIVersions".
+// If it returns false, it is safe to omit Capabilities.APIVersions from rendering values.
+func chartNeedsAPIVersions(chart *chart.Chart) bool {
```
gjenkins8, line 102: The implementation here: string search of the templates is a little brittle IMHO (at least the `tpl` function exists, which can dynamically execute templates

Let me try and see if there is an alternative.

eolatham: Thanks @gjenkins8. I would definitely prefer something more precise, but don't yet know a better way.
One of my colleagues had an idea: could the implementation of `APIVersions` change to be a function that when referenced, fetches the list or returns a cached copy from previous calls?

gjenkins8: https://kubernetes.io/docs/concepts/overview/kubernetes-api/#aggregated-discovery perhaps (I didn't check the details, but from the tin seems to be a more efficient way of fetching API versions)

eolatham: Following up on the suggested alternatives...

> Could the implementation of APIVersions change to be a function that when referenced, fetches the list or returns a cached copy from previous calls?

After taking a closer look, even if viable, this would require a breaking change to the `APIVersions` field within the public `Capabilities` struct.

> https://kubernetes.io/docs/concepts/overview/kubernetes-api/#aggregated-discovery

It looks like [ServerGroupsAndResources](https://github.com/kubernetes/client-go/blob/v0.31.0/discovery/discovery_client.go#L471) (used in [GetVersionSet](https://github.com/helm/helm/blob/v3.16.1/pkg/action/action.go#L319)) uses aggregated discovery by default if the provided client supports it. I'll try to evaluate the amount of effort it would take to use a compatible client.

eolatham: I'm almost positive the Helm library already supports using aggregated discovery clients. It just depends on the implementation of RESTClientGetter.ToDiscoveryClient. If your ToDiscoveryClient returns a client using [k8s.io/client-go/discovery](https://pkg.go.dev/k8s.io/client-go/discovery) v0.30 and up, it should use aggregated discovery OOTB.

What I'm not sure about is how the Helm CLI implements RESTClientGetter. I haven't been able to find that anywhere in the code so far.

eolatham: I confirmed that the discovery client used by the latest version of the Helm CLI supports aggregated discovery.
So I think this PR is not necessary! Thanks @gjenkins8 again for your feedback and suggestions.

eolatham:
The latest version of Helm supports [aggregated discovery](https://kubernetes.io/docs/concepts/overview/kubernetes-api/#aggregated-discovery), which should address (in a much preferred way) the issues this PR is meant to solve.

PR #13147, opened 2024-06-29
fix(dryRun): Remove (install|upgrade).DryRun and start using DryRunOption

### LINKED ISSUE #13144: Helm install in --dry-run still attempts kubernetes connection
According to the `helm install` help text:
> If --dry-run is set with no option being specified ... it will **not** attempt cluster connections.

Running the following command:
`helm install --dry-run --output yaml test ./mychart`

Should allow me to generate yaml of the install without contacting the kubernetes API. Instead I receive the following error:
`Error: INSTALLATION FAILED: Kubernetes cluster unreachable: Get "http://localhost:8080/version": dial tcp [::1]:8080: connect: connection refused`

It has the same issue with `helm install --dry-run=client --output yaml test ./mychart`

Output of `helm version`:
`version.BuildInfo{Version:"v3.15.1", GitCommit:"e211f2aa62992bd72586b395de50979e31231829", GitTreeState:"clean", GoVersion:"go1.22.3"}`

Output of `kubectl version`:
`Irrelevant`

`Irrelevant`

idsulik:
Fixes https://github.com/helm/helm/issues/13144

There are two fields DryRun and DryRunOption, DryRun is redundant because of DryRunOption. It confuses when you see similar fields

### DIFF
--- a/cmd/helm/template.go
+++ b/cmd/helm/template.go
@@ -86,12 +86,10 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 			if client.DryRunOption == "" {
 				client.DryRunOption = "true"
 			}
+			client.DryRun = true
 			client.ReleaseName = "release-name"
 			client.Replace = true // Skip the name check
-			client.DryRunOption = "client"
-			if validate {
-				client.DryRunOption = "server"
-			}
+			client.ClientOnly = !validate
 			client.APIVersions = chartutil.VersionSet(extraAPIs)
 			client.IncludeCRDs = includeCrds
 			rel, err := runInstall(args, client, valueOpts, out)
--- a/cmd/helm/template.go
+++ b/cmd/helm/template.go
@@ -86,10 +86,12 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 			if client.DryRunOption == "" {
 				client.DryRunOption = "true"
 			}
-			client.DryRun = true
 			client.ReleaseName = "release-name"
 			client.Replace = true // Skip the name check
-			client.ClientOnly = !validate
+			client.DryRunOption = "client"
+			if validate {
+				client.DryRunOption = "server"
+			}
 			client.APIVersions = chartutil.VersionSet(extraAPIs)
 			client.IncludeCRDs = includeCrds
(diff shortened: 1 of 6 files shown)


cmd/helm/template.go
```diff
@@ -86,10 +86,12 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 			if client.DryRunOption == "" {
 				client.DryRunOption = "true"
 			}
-			client.DryRun = true
 			client.ReleaseName = "release-name"
 			client.Replace = true // Skip the name check
-			client.ClientOnly = !validate
+			client.DryRunOption = "client"
```
idsulik, line 91: I'm not sure about this part

pkg/action/install.go
```diff
@@ -69,10 +69,8 @@ type Install struct {
 
 	ChartPathOptions
 
-	ClientOnly      bool
```
mattfarina, line 72: We can't remove public properties on structs because it would be a breaking change to the Go clients API. We can deprecate things but not remove them until there is a major version change for Helm.

mattfarina (changes requested):
I didn't give this a thorough review. I noticed that properties are being removed from public structs. This is API breaking to code that consumes the Helm SDK. We can't do that until there is a major version change for Go. Things need to be handled in a backwards compatible manner.

pkg/action/install.go
idsulik: Yes, totally agree, missed that moment, reverted my changes, and fixed in an easy way.  But I think we should start deprecating some of the fields because they duplicate the same logic so you can set ClientOnly=true, or DryRun="client"(or "true"), some code checks ClientOnly, some code checks DryRun

idsulik:
Hi @mattfarina ,

I noticed that this pull request has been pending for a while. Is there anything more needed from my side? Please let me know if there are any updates or additional changes required.

Thanks!

idsulik:
Hi @mattfarina 

I noticed that this pull request has been pending for a while. Is there anything more needed from my side? Please let me know if there are any updates or additional changes required.

Thanks!

idsulik:
Hi @mattfarina

I noticed that this pull request has been pending for a while. Is there anything more needed from my side? Please let me know if there are any updates or additional changes required.

Thanks!

PR #13154, opened 2024-07-02
Allow post-renderer to process hooks

### LINKED ISSUE #7891: Post-renderer not receiving chart hook manifests?

carloslima:
closes #7891

Allow post-renderer to receive and process manifests with hook annotations.

In order to avoid dealing with YAML comments or reconciling file contents back to their filenames, this MR calls the post-renderer right after the templates are rendered and before they get split into hooks, manifests and partials lists.

**If applicable**:

### DIFF
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -241,8 +241,8 @@ func (cfg *Configuration) renderResources(ch *chart.Chart, values chartutil.Valu
 		// them back into a map of filename -> content.
 
 		// Merge files as stream of documents for sending to post renderer
-		var merged string
-		if merged, err = annotateAndMerge(files); err != nil {
+		merged, err := annotateAndMerge(files)
+		if err != nil {
 			return hs, b, notes, fmt.Errorf("error merging manifests: %w", err)
 		}
 
@@ -253,7 +253,8 @@ func (cfg *Configuration) renderResources(ch *chart.Chart, values chartutil.Valu
 		}
 
 		// Use the file list and contents received from the post renderer
-		if files, err = splitAndDeannotate(postRendered.String()); err != nil {
+		files, err = splitAndDeannotate(postRendered.String())
+		if err != nil {
 			return hs, b, notes, fmt.Errorf("error while parsing post rendered output: %w", err)
 		}
 	}
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -128,9 +128,9 @@ func annotateAndMerge(files map[string]string) (string, error) {
 	return merged, nil
 }
 
-// SplitAndDeannotate reconstructs individual files from a merged YAML stream,
+// splitAndDeannotate reconstructs individual files from a merged YAML stream,
 // removing filename annotations and grouping documents by their original filenames.
-func SplitAndDeannotate(postrendered string) (map[string]string, error) {
+func splitAndDeannotate(postrendered string) (map[string]string, error) {
 	manifests, err := kio.ParseAll(postrendered)
 	if err != nil {
 		return nil, fmt.Errorf("re-parsing merged buffer: %w", err)
@@ -253,7 +253,7 @@ func (cfg *Configuration) renderResources(ch *chart.Chart, values chartutil.Valu
 		}
 
 		// Use the file list and contents received from the post renderer
-		if files, err = SplitAndDeannotate(postRendered.String()); err != nil {
+		if files, err = splitAndDeannotate(postRendered.String()); err != nil {
 			return hs, b, notes, fmt.Errorf("error while parsing post rendered files: %w", err)
 		}
 	}
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -99,9 +99,9 @@ const (
 	filenameAnnotation = "helm-postrender-filename"
 )
 
-// AnnotateAndMerge combines multiple YAML files into a single stream of documents,
+// annotateAndMerge combines multiple YAML files into a single stream of documents,
 // adding filename annotations to each document for later reconstruction.
-func AnnotateAndMerge(files map[string]string) (string, error) {
+func annotateAndMerge(files map[string]string) (string, error) {
 	var combinedManifests []*kyaml.RNode
 	for fname, content := range files {
 		// Skip partials and empty files.
@@ -242,7 +242,7 @@ func (cfg *Configuration) renderResources(ch *chart.Chart, values chartutil.Valu
 
 		// Merge files as stream of documents for sending to post renderer
 		var merged string
-		if merged, err = AnnotateAndMerge(files); err != nil {
+		if merged, err = annotateAndMerge(files); err != nil {
 			return hs, b, notes, fmt.Errorf("error merging manifests: %w", err)
 		}
 
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -96,7 +96,7 @@ type Configuration struct {
 const (
 	// filenameAnnotation is the annotation key used to store the original filename
 	// information in manifest annotations for post-rendering reconstruction.
-	filenameAnnotation = "helm-postrender-filename"
+	filenameAnnotation = "postrenderer.helm.sh/postrender-filename"
 )
 
 // annotateAndMerge combines multiple YAML files into a single stream of documents,
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -22,9 +22,11 @@ import (
 	"fmt"
 	"io"
 	"log/slog"
+	"maps"
 	"os"
 	"path"
 	"path/filepath"
+	"slices"
 	"strings"
 	"sync"
 	"text/template"
@@ -103,7 +105,12 @@ const (
 // adding filename annotations to each document for later reconstruction.
 func annotateAndMerge(files map[string]string) (string, error) {
 	var combinedManifests []*kyaml.RNode
-	for fname, content := range files {
+
+	// Get sorted filenames to ensure result is deterministic
+	fnames := slices.Sorted(maps.Keys(files))
+
+	for _, fname := range fnames {
+		content := files[fname]
 		// Skip partials and empty files.
 		if strings.HasPrefix(path.Base(fname), "_") || strings.TrimSpace(content) == "" {
 			continue
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -94,9 +94,9 @@ type Configuration struct {
 }
 
 const (
-	// FilenameAnnotation is the annotation key used to store the original filename
+	// filenameAnnotation is the annotation key used to store the original filename
 	// information in manifest annotations for post-rendering reconstruction.
-	FilenameAnnotation = "helm-postrender-filename"
+	filenameAnnotation = "helm-postrender-filename"
 )
 
 // AnnotateAndMerge combines multiple YAML files into a single stream of documents,
@@ -114,7 +114,7 @@ func AnnotateAndMerge(files map[string]string) (string, error) {
 			return "", fmt.Errorf("parsing %s: %w", fname, err)
 		}
 		for _, manifest := range manifests {
-			if err := manifest.PipeE(kyaml.SetAnnotation(FilenameAnnotation, fname)); err != nil {
+			if err := manifest.PipeE(kyaml.SetAnnotation(filenameAnnotation, fname)); err != nil {
 				return "", fmt.Errorf("annotating %s: %w", fname, err)
 			}
 			combinedManifests = append(combinedManifests, manifest)
@@ -142,11 +142,11 @@ func SplitAndDeannotate(postrendered string) (map[string]string, error) {
 		if err != nil {
 			return nil, fmt.Errorf("getting metadata: %w", err)
 		}
-		fname := meta.Annotations[FilenameAnnotation]
+		fname := meta.Annotations[filenameAnnotation]
 		if fname == "" {
 			fname = fmt.Sprintf("generated-by-postrender-%d.yaml", i)
 		}
-		if err := manifest.PipeE(kyaml.ClearAnnotation(FilenameAnnotation)); err != nil {
+		if err := manifest.PipeE(kyaml.ClearAnnotation(filenameAnnotation)); err != nil {
 			return nil, fmt.Errorf("clearing filename annotation: %w", err)
 		}
 		manifestsByFilename[fname] = append(manifestsByFilename[fname], manifest)
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -133,7 +133,7 @@ func annotateAndMerge(files map[string]string) (string, error) {
 func splitAndDeannotate(postrendered string) (map[string]string, error) {
 	manifests, err := kio.ParseAll(postrendered)
 	if err != nil {
-		return nil, fmt.Errorf("re-parsing merged buffer: %w", err)
+		return nil, fmt.Errorf("error parsing YAML: %w", err)
 	}
 
 	manifestsByFilename := make(map[string][]*kyaml.RNode)
@@ -254,7 +254,7 @@ func (cfg *Configuration) renderResources(ch *chart.Chart, values chartutil.Valu
 
 		// Use the file list and contents received from the post renderer
 		if files, err = splitAndDeannotate(postRendered.String()); err != nil {
-			return hs, b, notes, fmt.Errorf("error while parsing post rendered files: %w", err)
+			return hs, b, notes, fmt.Errorf("error while parsing post rendered output: %w", err)
 		}
 	}
 
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -34,6 +34,8 @@ import (
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
+	"sigs.k8s.io/kustomize/kyaml/kio"
+	kyaml "sigs.k8s.io/kustomize/kyaml/yaml"
 
 	chart "helm.sh/helm/v4/pkg/chart/v2"
 	chartutil "helm.sh/helm/v4/pkg/chart/v2/util"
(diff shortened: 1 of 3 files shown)


robertsirc:
Hello, this looks to be a breaking change, is there any consideration for backwards compatibility?

carloslima:
@robertsirc It may have escaped my attention. Can you elaborate on what does it break? The new behavior is behind an extra flag `--post-renderer-hooks`. Unless enabled by the flag the previous behavior should continue unchanged. Did I forget to consider a particular case?

carloslima:
@robertsirc I think you meant that I changed the `NewExec` signature. I have pushed a fix for that. Does it solve the problem you raised? Anything else I can add/change to improve the PR?

rossbachp:
Can you merge this feature, please!

rossbachp:
When is the release of Charts 2.x planned? I need the feature :)

viclim:
any chance this PR could get a review?

robertsirc:
Great PR. We are taking a deeper look at it now. One thing we have a style error if you could fix:

```bash 
golangci-lint run ./...
pkg/postrender/exec.go:41: File is not `gofmt`-ed with `-s` (gofmt)
  return NewExecHooks(binaryPath, false, args...)
``

**Comment by carloslima** [comment]
>  we have a style error if you could fix:

Ooops. Whitespace mistake, fixed :-)

**Comment by carloslima** [comment]
@robertsirc Is there anything I can do to help move this forward?

**Comment by spideyfusion** [comment]
This feature would really help me in my current deployment workflows. 🙏

**Comment by AbyPhantom** [comment]
I'd really love to see this implemented. Would be really helpful to me.

**Review by spideyfusion — APPROVED** [review]

File: pkg/postrender/postrender.go [hunk 1]
```diff
@@ -26,4 +26,12 @@ type PostRenderer interface {
...
 	// error if there was an issue or failure while running the post render step
 	Run(renderedManifests *bytes.Buffer) (modifiedManifests *bytes.Buffer, err error)
+	// RunIncHooks expects and returns a map of file names and their rendered contents
+	// Example:
+	// > map[string]string{
+	// >   "templates/foo.yaml": "Kind: Pod..",
+	// >   "templates/baz.yaml": "Kind: ConfigMap...",
+	// > }
+	RunIncHooks(renderedFiles map[string]string) (modifiedFiles map[string]string, err error)
+	IncHooks() bool
```
joejulian, line 36: This will break any existing 3rd party software that implements this interface, which might be ok since we're on the helm 4 path.

carloslima: Would [this change](https://github.com/helm/helm/pull/13154/commits/7f3e88d038dbd5d489a907649880793e8b54887a) be appropriate to keep the interface backwards compatible?

I was going to rebase the MR but noticed that main is moving to helm v4. Should I rebase on top of dev-v3 instead?

carloslima: Hi @joejulian 
Can I get a review on this MR? Anything further I can do to help move it along?

joejulian: > Would [this change](https://github.com/helm/helm/pull/13154/commits/7f3e88d038dbd5d489a907649880793e8b54887a) be appropriate to keep the interface backwards compatible?
> 
> I was going to rebase the MR but noticed that main is moving to helm v4. Should I rebase on top of dev-v3 instead?

No, this is fine. I wanted to confirm with some of the other maintainers.

carloslima: @joejulian Let me know what the next steps are to move this forward.

benoittgt (approved):
I've reviewed this patch several times and support this well-executed addition to meet the request. ;)

cmd/helm/flags.go
```diff
@@ -116,7 +117,9 @@ func (o *outputValue) Set(s string) error {
 }
 
 func bindPostRenderFlag(cmd *cobra.Command, varRef *postrender.PostRenderer) {
-	p := &postRendererOptions{varRef, "", []string{}}
+	p := &postRendererOptions{varRef, "", []string{}, false}
+	cmd.Flags().Var(&postRendererBool{p}, postRenderHooksFlag, "include hooks when running the post-renderer")
+	cmd.Flags().Lookup(postRenderHooksFlag).NoOptDefVal = "true"
```
gjenkins8, line 122: I'm wondering if we want this flag. For compatibility purposes, certainly, we would need a flag to prevent the behavior from unexpectedly changing.

But for Helm 4, we don't have that strong compatibility requirement. We do want to maintain workflow compatiblity, especially for common workflows.

I'm wondering if there is any real use case for a post renderer to not process hooks? And even if so, whether the post renderer should exclude hook manifests based on its own internal logic (it could/would not post-render manifests with the hook annotation).

If this did sound agreeable, we could simply always pass hook manifets to the post-renderer. And avoid the additional flag, which presents itself as a thing a user has to discover and understand (and to a much lesser degree of concern, the implementation complexity).

cmd/helm/flags.go
```diff
@@ -125,6 +128,29 @@ type postRendererOptions struct {
...
+func (p *postRendererBool) Type() string {
+	return "postRendererBool"
+}
+
+func (p *postRendererBool) String() string {
+	return fmt.Sprintf("%t", p.options.incHooks)
+}
+
+func (p *postRendererBool) Set(val string) error {
+	p.options.incHooks = (val != "false" && val != "0")
```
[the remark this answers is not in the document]
gjenkins8: (but see note above on whether we should have flag at all)

gjenkins8 (commented):
Lets decide if we want a flag, then can look at the details on the implementation

cmd/helm/flags.go
carloslima: Based on the discussions in #7891 and the fact that it's marked as a bug, it doesn't seem like excluding hooks from the post-renderer was ever intended.

I agree, it makes sense for Helm 4 to have a single, consistent post-rendering behavior, rather than supporting two modes behind a flag.

If we still want to provide backward compatibility with Helm 3, we could offer an external wrapper that processes the new Helm 4 format, removes the hooks, and pass the manifests in the legacy Helm 3 format to the legacy post renderer.

Something like this:
```
#!/usr/bin/env ruby

# Usage: helm3-adapter <legacy-post-renderer> <arg1> <arg2> ..

require 'yaml'
require 'open3'

# Transform Helm 4 output into legacy Helm 3 format
yaml_output = StringIO.new
YAML.load(STDIN.read).reject { |filename, manifest| manifest.dig(*%w(metadata annotations helm.sh/hook)) }.each do |f, c|
  yaml_output.puts "---"
  yaml_output.puts "# Source: #{f}"
  yaml_output.puts c.to_yaml.sub(/^---\n/, '')
end

# Start the external process with the provided arguments
if ARGV.length > 0
  stdout, stderr, status = Open3.capture3(*ARGV, stdin_data: yaml_output.string)
  
  # Print the output from the process
  puts stdout unless stdout.empty?
  STDERR.puts stderr unless stderr.empty?
  exit status.exitstatus
end
```
to be invoked as:
```
helm install ... --post-renderer helm3-adapter --post-renderer-args 'legacy-post-renderer' --post-renderer-args 'legacy pr arguments'
```

That reduces the complexity inside Helm while still offering a workable backwards compatible solution.

I can make the changes to remove the flag soon.

cmd/helm/flags.go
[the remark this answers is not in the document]
carloslima: Thanks for pointing this out. Yes, if we have a boolean flag it should only accept `false, False, FALSE, f, F, 0`, `true, True, TRUE, t, T, 1` :+1:

cmd/helm/flags.go
carloslima: @gjenkins8 I've pushed the changes to drop Helm 3 compatibility. Can you review again?

pkg/postrender/exec.go
```diff
@@ -41,8 +43,31 @@ func NewExec(binaryPath string, args ...string) (PostRenderer, error) {
 	return &execRender{fullPath, args}, nil
 }
 
+func (p *execRender) Run(renderedFiles map[string]string) (modifiedFiles map[string]string, err error) {
```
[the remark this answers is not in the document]
carloslima: No strong reason. I initially thought the naming made things clearer for documentation purposes, but I don't think it actually does.

It's fixed now.

pkg/postrender/exec.go
```diff
@@ -41,8 +43,31 @@ func NewExec(binaryPath string, args ...string) (PostRenderer, error) {
 	return &execRender{fullPath, args}, nil
 }
 
+func (p *execRender) Run(renderedFiles map[string]string) (map[string]string, error) {
+	// Serialize rendered files into a YAML buffer
+	originalManifests, err := yaml.Marshal(renderedFiles)
```
gjenkins8, line 48: (Apologies for the drip-fed review comments)

I don't think we can simply change the post-renderer input format like this. The canonical use-case for the post renderer is a tool which accepts a stream of yaml documents, modifies them, and emits the stream back to Helm e.g. [kustomize](https://kustomize.io/).

Explicitly, this simple stdin/stdout "pipe" + yaml doc interface is how the majority of tools out there expect to consume a yaml stream. Changing the format breaks any/all tools which expect this simple format, and would force an unexpected change on users.

We also do have an explicit goal for Helm 4 that we will be compatible with the majority of user workflows. It is a judgement call to whether e.g. passing hook yaml in post-renderer vs changing the format is "compatible with the majority of user workflows". But for this example, not passing hook yaml is fairly unexpected from a user perspective. And easy'ish for users to mitigate against, if they actually needed to (judged to be unexpected), when they upgrade to Helm 4.

But forcing probably all users to adapt post-renderer format is a significant breaking change. Certainly, e.g. yamlscript could be used to "intercept" the map, and invoke a sub-command per-document. But this is still an "unexpected" (IMHO) change for the majority of post-renderer users.

For this PR, IMHO we should (must) keep the scope: to passing hook yaml to the post-renderer only. If we want to change the post-renderer format, we would need further discussion. And make a separate change (that does account for varying user scenarios).



```
file1.yaml: |
   <content>
_helpers.tpl:
  <content>
```

gjenkins8 (changes requested):
https://github.com/helm/helm/pull/13154/files#r2096150476

pkg/postrender/exec.go
carloslima: I understand this is a significant change and that's why I called it out clearly in the PR description, added comments in the code, and initially gated it behind a flag. 🙂

Let me try a different approach. I need a way to recover the filenames after they've been processed by the post-render step, but without relying on YAML comments, since most YAML parsers don't preserve them.

gjenkins8: > I understand this is a significant change and that's why I called it out clearly in the PR description, added comments in the code, and initially gated it behind a flag.

Apologies. I missed the format change originally (too much multi-tasking sorry)

gjenkins8: >I understand this is a significant change and that's why I called it out clearly in the PR description, added comments in the code, and initially gated it behind a flag. 🙂

Apologies, too much multi-tasking here. My review was very round-a-bout.

But I do still think a flag is not warranted -- too many tools will expect the document-based yaml stream. *As well as* being able to process hook yaml. So IMHO we are still on the right path.

> Let me try a different approach. I need a way to recover the filenames after they've been processed by the post-render step, but without relying on YAML comments, since most YAML parsers don't preserve them.

This is the behavior for non-hook Kubernetes YAML today? The post renderer will drop the filename comments.. Is it necessary for Helm to recover these?

If it is necessary, I think we can assume Kubernetes' API object type and name (the tuple of: `(GroupVersion, Kind, Namespace, Name)`) is unique. A map of this tuple to the filename, would then allow mapping back to the filename post-rendering I think. New or mutated objects that don't match, wouldn't map back to a "file". But that is expected I think.

carloslima: I think I have a workable solution. I will publish it soon :hourglass_flowing_sand:

carloslima: I gotta run out now. I'll fix this PR later but the changes are here: https://github.com/carloslima/helm/pull/1/files

Tests are long but the actual changes are relatively small. Can you take a look?

carloslima: @gjenkins8 I've updated the MR. Please check when you have time :)

robertsirc:
@gjenkins8 you are requesting changes on this PR is this still active or have your changes been met?

dwalters:
I clearly missed a lot of changes here since the original PR!   I rebased a similar change in #30778 against v3 earlier today, but it's basically the same approach that's now being taken here, and this looks fine to me.  I'd suggest this be the default behavior in helm v4.

For v3, I do prefer opting in to the new behavior via an annotation instead of having to provide a CLI flag.

carloslima:
@dwalters I've been waiting on a review for this PR for about a year now. It’s clear this isn’t a top priority :-)

pkg/action/action.go
```diff
@@ -91,6 +93,76 @@ type Configuration struct {
 	mutex sync.Mutex
 }
 
+const (
+	// FilenameAnnotation is the annotation key used to store the original filename
+	// information in manifest annotations for post-rendering reconstruction.
+	FilenameAnnotation = "helm-postrender-filename"
```
gjenkins8: Can we make this private? Unless there is a good reason to make it public, and so part of Helm's SDK public interface, private is preferred. As that allows it to remain an implementation detail, and able to be changed in the future. If public, it becomes part of Helm's public SDK interface, and non-amendable to changes/non-backwards compatible updates in the future.

pkg/action/action.go
```diff
@@ -91,6 +93,76 @@ type Configuration struct {
...
+)
+
+// AnnotateAndMerge combines multiple YAML files into a single stream of documents,
+// adding filename annotations to each document for later reconstruction.
+func AnnotateAndMerge(files map[string]string) (string, error) {
```
gjenkins8: Similar to above, can we make this private(?)

pkg/action/action.go
```diff
@@ -91,6 +93,76 @@ type Configuration struct {
...
+	merged, err := kio.StringAll(combinedManifests)
+	if err != nil {
+		return "", fmt.Errorf("writing merged docs: %w", err)
+	}
+	return merged, nil
+}
+
+// SplitAndDeannotate reconstructs individual files from a merged YAML stream,
+// removing filename annotations and grouping documents by their original filenames.
+func SplitAndDeannotate(postrendered string) (map[string]string, error) {
```
gjenkins8: Similar to above, can we make this private(?)

pkg/action/action.go
```diff
@@ -91,6 +93,76 @@ type Configuration struct {
...
+	manifests, err := kio.ParseAll(postrendered)
+	if err != nil {
+		return nil, fmt.Errorf("re-parsing merged buffer: %w", err)
```
gjenkins8: nit: If the post renderer produces invalid yaml, the user will get a somewhat (IMHO) confusing error like:
```
error while parsing post rendered files: re-parsing merged buffer: MalformedYAMLError: yaml: line 3: could not find expected ':'
```

(<https://go.dev/play/p/gT-Z-yLmDvQ>) 

The user does not (and should not) know what the "merged buffer" is. I think we should align this error chain message closer to what will be helpful to the user e.g. something like:
```
error while parsing post renderer output: error parsing YAML: MalformedYAMLError: yaml: line 3: could not find expected ':'
```

pkg/action/action.go
```diff
@@ -160,6 +232,32 @@ func (cfg *Configuration) renderResources(ch *chart.Chart, values chartutil.Valu
...
+		// We need to send files to the post-renderer before sorting and splitting
+		// hooks from manifests. The post-renderer interface expects a stream of
+		// manifests (similar to what tools like Kustomize and kubectl expect), whereas
+		// the sorter uses filenames.
+		// Here, we merge the documents into a stream, post-render them, and then split
+		// them back into a map of filename -> content.
+
+		// Merge files as stream of documents for sending to post renderer
+		var merged string
+		if merged, err = AnnotateAndMerge(files); err != nil {
```
gjenkins8: style nit:
```suggestion
		merged, err := AnnotateAndMerge(files)
		if err != nil {
```

pkg/action/action.go
```diff
@@ -160,6 +232,32 @@ func (cfg *Configuration) renderResources(ch *chart.Chart, values chartutil.Valu
...
+		}
+
+		// Run the post renderer
+		postRendered, err := pr.Run(bytes.NewBufferString(merged))
+		if err != nil {
+			return hs, b, notes, fmt.Errorf("error while running post render on files: %w", err)
+		}
+
+		// Use the file list and contents received from the post renderer
+		if files, err = SplitAndDeannotate(postRendered.String()); err != nil {
```
gjenkins8: style nit:
```suggestion
		manifests, err := SplitAndDeannotate(postRendered.String())
		if err != nil {
```

pkg/action/action.go
```diff
@@ -160,6 +232,32 @@ func (cfg *Configuration) renderResources(ch *chart.Chart, values chartutil.Valu
...
+			return hs, b, notes, fmt.Errorf("error while parsing post rendered files: %w", err)
```
gjenkins8: nit: the post renderer doesn't handle "files". It streams an output:
```suggestion
			return hs, b, notes, fmt.Errorf("error while parsing post renderer output: %w", err)
```

pkg/action/action.go
gjenkins8: We should probably name this annotation more canonically. I think something like:
```
postrenderer.helm.sh/postrender-filename
```

pkg/action/action.go
gjenkins8: This function does two distinct things:
1. Parses the yaml string
2. Reconstructs the filename -> yaml string map

I wonder if it would be better to split these disinct things into seperate functions? In that it would be easier to test and reason about?

pkg/action/action.go
```diff
@@ -91,6 +93,76 @@ type Configuration struct {
...
+			return nil, fmt.Errorf("getting metadata: %w", err)
+		}
+		fname := meta.Annotations[FilenameAnnotation]
+		if fname == "" {
+			fname = fmt.Sprintf("generated-by-postrender-%d.yaml", i)
+		}
+		if err := manifest.PipeE(kyaml.ClearAnnotation(FilenameAnnotation)); err != nil {
+			return nil, fmt.Errorf("clearing filename annotation: %w", err)
+		}
+		manifestsByFilename[fname] = append(manifestsByFilename[fname], manifest)
```
gjenkins8, line 159: If the post renderer contains documents with duplicate annotations, we will overwrite the previous documents here?

pkg/action/action.go
gjenkins8: This function does three distinct things:

1. Parses the yaml in filename -> yaml doc map into kyaml types
2. Loops each manifest in the kyaml type and adds the annotation
3. finally, writes the array of manifests into a yaml doc

I wonder if it would be better to split these distinct things into separate functions/functionality? For ease of testing and readability?

gjenkins8 (changes requested):
https://github.com/helm/helm/pull/13154/files#r2147057212 is a bug we need to fix I think

otherwise, most of my comments are style/refactoring type concerns. logic here I think is the way to go. thanks for the updates.

pkg/action/action.go
carloslima: Done

pkg/action/action.go
carloslima: done

pkg/action/action.go
carloslima: done

pkg/action/action.go
carloslima: done

pkg/action/action.go
carloslima: done

pkg/action/action.go
carloslima: done

pkg/action/action.go
carloslima: done

pkg/action/action.go
carloslima: If I understood your question correctly, no, they will be combined. See test case here: https://github.com/helm/helm/pull/13154/files#diff-52b738d4936c2effb6743132c20ba5d13a6be01903130a3fe78f6a7754f8c17dR617-R650

pkg/action/action.go
carloslima: The YAML parsing is already handled by a single function call followed by an if err check. I don’t see a way to make it any simpler. If you have any suggestions that could improve readability or testability, let me know :-)

pkg/action/action.go
carloslima: Same here: YAML parsing and dumping are already handled by a single function call followed by an `if err` check. Anything I can think of would only make it noisier. If you have a suggestion, I’d be happy to hear it :)

pkg/action/action.go
carloslima: done

carloslima:
> https://github.com/helm/helm/pull/13154/files#r2147057212 is a bug we need to fix I think
> 
> otherwise, most of my comments are style/refactoring type concerns. logic here I think is the way to go. thanks for the updates.

Thanks for the review. I made all the changes except for the three cases I mentioned above. They are in separate commits to make review easier, but I’ll squash them if you're happy with the changes.

gjenkins8 (dismissed):
nice, thanks for the updates. Agree https://github.com/helm/helm/pull/13154/files#r2147057212 is actually handled. 

LGTM, just nee to resolve the go.mod conflict and I can reapprove (will require one more maintainer approval)

carloslima:
> nice, thanks for the updates. Agree https://github.com/helm/helm/pull/13154/files#r2147057212 is actually handled.
> 
> LGTM, just nee to resolve the go.mod conflict and I can reapprove (will require one more maintainer approval)

Done. I've rebased it on top of latest main :)

carloslima:
> We should also probably document the `postrenderer.helm.sh/postrender-filename` annotation. Something along the lines like: setting this annotation in the post-renderer will cause Helm to associate that "filename" with the YAML doc. https://helm.sh/docs/topics/advanced/#post-rendering
> 
> I've added the `docs-needed` label to track.

Is this enough?

https://github.com/helm/helm-www/pull/1734

gjenkins8 (approved):

scottrigby (approved):
nice work 👏 Tested manually as well. LGTM

Anany-k:
Will this be included in a release for helm v3 ?

benoittgt:
@Anany-k I don’t think so, based on https://github.com/helm/helm/pull/13154#issuecomment-2972932742

PR #13155, opened 2024-07-04
Fix #12798 | Lint sub-charts with properly scoped values

### LINKED ISSUE #12798: helm lint with-subcharts applies wrong values to subcharts
Output of `helm version`:

Output of `kubectl version`:

ziollek:

This PR closes #12798. The issue is related to the problem of passing the copy of all values to the linted sub-charts instead of passing only appropriate ones - what values should be passed can be found in the [docs](https://helm.sh/docs/chart_template_guide/subcharts_and_globals/).

To simplify changes related to the scoping values I decided to get rid of undocumented behavior that allowed passing multiple paths to `helm lint` command  (take a look at `TestLintCmdWithMultipleChartsIsProhibited`).

The original approach of linting sub-charts is based on inspecting `charts` subdirectory instead of analyzing dependencies of the main chart. It could be counterintuitive, but I didn't want to revolutionize too much so there are only slight changes in such logic:
- parent chart values are loaded and merged with passed values
- parent chart metadata is used to determine what scope should be used for a particular sub-chart. It is important to mention that a single sub-chart can be used in many contexts by using [aliases](https://helm.sh/docs/helm/helm_dependency/). To handle each sub-chart can be validated many times (for found scope)
- to be consistent with the original approach - sub-charts that were found in `charts` directory but are not found in the main chart dependencies are still linted but with empty values

It is also worth mentioning that lint output is slightly extended by adding scope, for example:
```
==> Linting testdata/testcharts/chart-with-deprecated-api (scope: root)
``` 
The part `(scope: root)` was added and can have one of three values:
- root - main chart
- empty - the sub-chart that was found in `charts` subdirectory is absent in the main chart dependencies
- <name|alias> - name or alias of linted dependency

**If applicable**:


cmd/helm/lint.go
```diff
@@ -153,3 +178,39 @@ func newLintCmd(out io.Writer) *cobra.Command {
...
+		return map[string]chartutil.Values{"root": vals}
+	}
+	result := map[string]chartutil.Values{}
+	if parent.metadata != nil {
+		// try to load dependent chart and match it with parent definition of dependencies
+		if subChart, err := loader.Load(single.path); err == nil {
+			for _, dependency := range parent.metadata.Dependencies {
+				// it is worth remembering that there could be the same sub-chart used in many dependencies
+				// to handle this we should validate such sub-chart as many times (with the appropriate set of values)
+				// aa it is defined in the parent chart
```
awmackowiak, line 193: 'aa' is not necessary I think.

johnjeffers:
Is this PR intended to be merged eventually? I'm on version 3.16.4 and I still can't lint subcharts properly because values do not get passed to the subcharts.

PR #13177, opened 2024-07-12
Improves handling of Helm index with broken helm chart versions #13176

### LINKED ISSUE #13176: improve handling of Helm index with broken helm chart versions
While trying to figure out the issue https://github.com/fluxcd/source-controller/issues/1515 in the flux source controller of filtering chart versions from a Helm index with broken dependencies we stumbled over this issue https://github.com/helm/helm/issues/12748 where it has already been fixed in Helm.

On further investigation it appeared that the exclusion of invalid versions in https://github.com/helm/helm/blob/1a500d5625419a524fdae4b33de351cc4f58ec35/pkg/repo/index.go#L345 was not ideal.

even though the copied slice `cvs` was adjusted to the correct size, the original slice `i.Entries` still remained the original size. It just contained copies of references of valid versions.

example of `i.Entries` before filtering:
```
[0xc000130f30 0xc0001315f0  0xc000131a70 0xc00021f710]
```

assuming chart versions at location `0xc0001315f0` and `0xc000131a70` are invalid. The content of cvs after validating the versions looked like this (which is correct):
```
[0xc000130f30 0xc00021f710]
```

But `i.Entries` still remains the original size:
```
[0xc000130f30 0xc00021f710 0xc00021f710 0xc00021f710]
```

Technically it does not contain any invalid version. But later on the index gets further sorted and filtered which unnecessarily contains duplicate versions.

a simple reassignment of `cvs` to `i.Entries[name]` solves this issue.

Additionally:

**If the last chart in the index is malformed, it will not be removed**

bb-Ricardo:
See issue #13176

**If applicable**:

closes #13176

### DIFF
--- a/pkg/repo/index.go
+++ b/pkg/repo/index.go
@@ -353,6 +353,10 @@ func loadIndex(data []byte, source string) (*IndexFile, error) {
 		return i, err
 	}
 
+	if i.APIVersion == "" {
+		return i, ErrNoAPIVersion
+	}
+
 	for name, cvs := range i.Entries {
 		for idx := len(cvs) - 1; idx >= 0; idx-- {
 			if cvs[idx] == nil {
@@ -371,11 +375,10 @@ func loadIndex(data []byte, source string) (*IndexFile, error) {
 				cvs = append(cvs[:idx], cvs[idx+1:]...)
 			}
 		}
+		// adjust slice to only contain a set of valid versions
+		i.Entries[name] = cvs
 	}
 	i.SortEntries()
-	if i.APIVersion == "" {
-		return i, ErrNoAPIVersion
-	}
 	return i, nil
 }
 
--- a/pkg/repo/index.go
+++ b/pkg/repo/index.go
@@ -353,10 +353,6 @@ func loadIndex(data []byte, source string) (*IndexFile, error) {
 		return i, err
 	}
 
-	if i.APIVersion == "" {
-		return i, ErrNoAPIVersion
-	}
-
 	for name, cvs := range i.Entries {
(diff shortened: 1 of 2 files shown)


bb-Ricardo:
Added the missing tests.

pkg/repo/index.go
```diff
@@ -371,11 +375,10 @@ func loadIndex(data []byte, source string) (*IndexFile, error) {
 				cvs = append(cvs[:idx], cvs[idx+1:]...)
 			}
 		}
+		// adjust slice to only contain a set of valid versions
+		i.Entries[name] = cvs
 	}
 	i.SortEntries()
-	if i.APIVersion == "" {
```
gjenkins8, line 376: I recommend not combining unrelated change into a single PR. Agree, I see no reason not to "optimize" the behavior here. But, IMHO coupling unrelated changes (the bugfix for i.Entries filtering, and this optimization) is bad practice.

bb-Ricardo: Hi, I agree. Will be able to change it in a couple of days as I currently out of town.

gjenkins8 (changes requested):
https://github.com/helm/helm/pull/13177/files#r1693820690

bb-Ricardo:
Still out of town, back next week.

pkg/repo/index.go
bb-Ricardo: reverted the change regarding the check if APIVersion is empty

gjenkins8 (approved):

mattfarina (approved):

mattfarina:
Note, the force push was to make CI run so the checks passed. No code was changed by the maintainers.

Issue #13174, opened 2024-07-11
Different behavior between `helm` from `brew` and `helm` from the `get.helm.sh`


davinkevin:
I have a strange behavior when I use `helm` template, and this is different based on the `helm` distribution I use (from `brew` or from `get.helm.sh`). 

The evaluation of `.Capabilities.KubeVersion.Version` is different based on the `helm` distribution I use. 

With the simplest chart possible: 

```yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: foo
  annotations:
    version: {{ .Capabilities.KubeVersion.Version }}
```

If I render this with `helm` from `get.helm.sh`, I have: 

```shell
$ ./.bin/helm template foo
---
# Source: foo/templates/serviceaccount.yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: foo
  annotations:
    version: v1.30.0
```

If I do this with the version installed with `brew`:

```shell
$ /opt/homebrew/Cellar/helm/3.15.3/bin/helm template foo
---
# Source: foo/templates/serviceaccount.yaml
apiVersion: v1
kind: ServiceAccount
metadata:
  name: foo
  annotations:
    version: v1.20.0
```

You can see the version is `1.20.0` for the `brew` distribution and `1.30.0` for `helm` standard.

You [can find a project with reproducers](https://gitlab.com/davinkevin.fr/issues/helm) and `Taskfile` to run commands

Output of `helm version`:

The one from `brew`: 

```
version.BuildInfo{Version:"v3.15.3", GitCommit:"Homebrew", GitTreeState:"clean", GoVersion:"go1.22.5"}
```

The one from `get.helm.sh`:

```
version.BuildInfo{Version:"v3.15.3", GitCommit:"3bb50bbbdd9c946ba9989fbe4fb4104766302a64", GitTreeState:"clean", GoVersion:"go1.22.5"}
```

I expect the value to be the same, and eventually for me the 1.30 which is the highest value.

Output of `kubectl version`: N/A, I'm not connected to a cluster and use only `helm template`.

Cloud Provider/Platform (AKS, GKE, Minikube etc.): None

chenrui333:
should be fixed with https://github.com/Homebrew/homebrew-core/pull/177094

gjenkins8:
Thanks @chenrui333 !

gjenkins8:
[...]

PR #13180, opened 2024-07-16
build: set kube version via `debug.BuildInfo`

branchv:

This is inspired by #13174, where Homebrew built `helm` outside of the Makefile and accidentally did not set the correct ldflags for the `k8sVersion` variables (which were defaulted to `1.20` instead of `1.30`)

Instead, this uses `runtime/debug`'s [`ReadBuildInfo`](https://pkg.go.dev/runtime/debug#ReadBuildInfo) to automatically extract the correct `k8s.io/client-go` version, regardless if the Makefile was used (for example, using `go install helm.sh/helm/v3/cmd/helm@latest`). It also adds a similar fallback for the git metadata ldflags using 1.18's [`BuildSetting`](https://pkg.go.dev/runtime/debug#BuildSetting) (but respects the variables if they are already set):

```console
$ go build ./cmd/helm && ./helm version
version.BuildInfo{Version:"v3.15", GitCommit:"7af094d606aff055c1eb938775b6fd260e637158", GitTreeState:"clean", GoVersion:"go1.22.5"}
```

This does introduce a lot of test golden changes since tests now also run with the correct `k8s.io/client-go` version

**If applicable**:

### DIFF
--- a/internal/version/version.go
+++ b/internal/version/version.go
@@ -19,6 +19,7 @@ package version // import "helm.sh/helm/v3/internal/version"
 import (
 	"flag"
 	"runtime"
+	"runtime/debug"
 	"strings"
 )
 
@@ -78,4 +79,23 @@ func Get() BuildInfo {
 		v.GoVersion = ""
 	}
 	return v
+}
+
+func init() {
+	if gitCommit != "" || gitTreeState != "" {
+		return
+	}
+	info, ok := debug.ReadBuildInfo()
+	if !ok {
+		return
+	}
+	for _, setting := range info.Settings {
+		if setting.Key == "vcs.revision" && setting.Value != "" {
+			gitCommit = setting.Value
+		} else if setting.Key == "vcs.modified" && setting.Value == "true" {
+			gitTreeState = "dirty"
+		} else if setting.Key == "vcs.modified" && setting.Value == "false" {
+			gitTreeState = "clean"
+		}
+	}
 }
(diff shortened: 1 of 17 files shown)


robertsirc:
This PR is a bit in depth could you attend the developer call on Thursdays at 9:30-10:00 Pacific ([meeting details](https://github.com/helm/community/blob/master/communication.md#meetings))??

gjenkins8 (commented):
Thanks for the PR. I do think this is a worthy change. TIL about https://pkg.go.dev/runtime/debug#ReadBuildInfo . Removing these ldflags and Makefile logic from Helm's certainly seems an improvement.

And in particular, Helm is often used as an SDK. And SDK users (presumably) find out the hard way they need to set these ldflags in their build system.

The only concerns I can think of are a) accuracy. I assume `debug.ReadBuildInfo()` can be accurate here. But probably just need to do some diligence to confirm, and b) performance. As long as reading this info does have some significant latency (I can't imagine it does), I can't imagine Helm wouldn't want to read this info at runtime.

internal/version/version.go
```diff
@@ -79,3 +80,22 @@ func Get() BuildInfo {
 	}
 	return v
 }
+
+func init() {
+	if gitCommit != "" || gitTreeState != "" {
```
gjenkins8, line 85: Can we make a seperate PR for chaning the git state please? The Kubernetes changes look good to merge for me.

Here, I think we should take `gitCommit` from `vcs.revision`. But to keep the seperate changes independant, think worth creating separate PRs.

gjenkins8 (changes requested):
Git commit changes I think need more thought. If we move them to a seperate PR, can approve the Kubernetes client-go version changes

internal/version/version.go
branchv: No problem, backed this out

branchv:
Thanks for pushing this forward @gjenkins8

TerryHowe:
Whoops closed

branchv:
Yes this was superseded by #31296

Issue #12952, opened 2024-04-16
Command "repository add" raises "Client.Timeout or context cancellation while reading body"


jesuspv:
Dear helm developers,

we are suffering timeouts on `helm repository add` commands of the form:
```
Error: looks like "https://<some-repository-url-here>" is not a valid chart repository or cannot be reached: context deadline exceeded (Client.Timeout or context cancellation while reading body)
```
I have seen that the code in [#pkg/getter](https://github.com/helm/helm/blob/14d0c13e9eefff5b4a1b511cf50643529692ec94/pkg/getter/getter.go#L186) has a 120s timeout for such task, which seems not enough when facing slow networks or large repositories (ours have more than 200MB in metadata).

I was wondering whether or not there is a way to configure such timeout or any hope to have it implemented.

Please, let me know if you need any additional info.

Thank you very much in advance
/J

kim0:
This change is causing us some pain too. It originated in https://github.com/helm/helm/pull/12203
cc @0xff-dev would be better if this can be configurable somehow (env var?)

0xff-dev:
> This change is causing us some pain too. It originated in #12203 cc @0xff-dev would be better if this can be configurable somehow (env var?)

I looked at the code and the helm repo add command doesn't have a `timeout` parameter, so that's causing a timeout here.
I'm very sorry for the impact @jesuspv  , I think it would be possible to add a timeout parameter to the add command? That way the timeout is configurable.

PR #13185, opened 2024-07-17
chore: DefaultHTTPTimeout can be configured via the environment variable

0xff-dev:
The previous PR https://github.com/helm/helm/pull/12203, which added a default timeout for httpProvider, was not configurable, and commands such as `helm repo add`, `helm repo update`, etc. did not have a timeout parameter, and there was a problem with Issue https://github.com/helm/helm/issues/12952.

To solve this problem, there are two possible options

1. add a timeout parameter to the helm repo subcommands.
2. make DefaultHTTPTimeout configurable via environment variables.

Considering the minimum changes and impact, use option 2.

**If applicable**:

### DIFF
--- a/pkg/getter/getter.go
+++ b/pkg/getter/getter.go
@@ -19,6 +19,8 @@ package getter
 import (
 	"bytes"
 	"net/http"
+	"os"
+	"strconv"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


pkg/getter/getter.go
```diff
@@ -179,20 +181,25 @@ func (p Providers) ByScheme(scheme string) (Getter, error) {
...
+var (
 	// The cost timeout references curl's default connection timeout.
 	// https://github.com/curl/curl/blob/master/lib/connect.h#L40C21-L40C21
 	// The helm commands are usually executed manually. Considering the acceptable waiting time, we reduced the entire request time to 120s.
 	DefaultHTTPTimeout = 120
 )
 
-var defaultOptions = []Option{WithTimeout(time.Second * DefaultHTTPTimeout)}
+func init() {
+	if v, err := strconv.Atoi(os.Getenv("DEFAULT_HTTP_TIMEOUT")); err == nil {
```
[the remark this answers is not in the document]
0xff-dev: Good idea, I'll update it.

pkg/getter/getter.go
```diff
@@ -183,16 +185,26 @@ const (
 	// The cost timeout references curl's default connection timeout.
 	// https://github.com/curl/curl/blob/master/lib/connect.h#L40C21-L40C21
 	// The helm commands are usually executed manually. Considering the acceptable waiting time, we reduced the entire request time to 120s.
-	DefaultHTTPTimeout = 120
+	defaultHTTPTimeout = 120
```
TerryHowe, line 188:
```suggestion
	DefaultHTTPTimeout = 120
```

Since this is not an internal package, this probably needs to stay public or this would be a breaking change.

pkg/getter/getter.go
```diff
@@ -183,16 +185,26 @@ const (
...
 )
 
-var defaultOptions = []Option{WithTimeout(time.Second * DefaultHTTPTimeout)}
```
TerryHowe, line 189: I'm not sure you want to make`GetDfaultHTTPTimeout` and `DefaultOptions` public unless there is a need.

0xff-dev: My previous idea was that if a user's project depends on helm and wants to get the default http timeout, helm should provide a method to get it, so I set the `GetDefaultHTTPTimeout` method to public.
And there is no need to set the `DefaultOptions` method to public, I'll Update it.

PR #13196, opened 2024-07-23
Added targetPort to set container port number instead of using the service port.

RitwikGopi:
- Added targetPort to set container port number instead of using the service port.
- Since the `targetProt` represents the port in container it would be more apt to use the same to configure it.

**If applicable**:


RitwikGopi:
I understood why the sample was designed like it is. This change is not needed

PR #13200, opened 2024-07-25
feat(helm): add --namespace-labels flag to install and upgrade

caruccio:
It creates the new namespace with given values to flag '--namespace-labels'. If not if provider, the default 'name=<namespace-name>' is used.

New namespaces may need to be created with a custom set of labels. This PR adds this capability.

**If applicable**:

### DIFF
--- a/cmd/helm/install.go
+++ b/cmd/helm/install.go
@@ -175,6 +175,7 @@ func newInstallCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 
 func addInstallFlags(cmd *cobra.Command, f *pflag.FlagSet, client *action.Install, valueOpts *values.Options) {
 	f.BoolVar(&client.CreateNamespace, "create-namespace", false, "create the release namespace if not present")
+	f.StringToStringVarP(&client.NamespaceLabels, "namespace-labels", "L", nil, "Labels that would be added to created namespace. Should be divided by comma. If --create-namespace is not set, this is ignored. Default is 'name=<namespace-name>'")
 	// --dry-run options with expected outcome:
 	// - Not set means no dry run and server is contacted.
 	// - Set with no value, a value of client, or a value of true and the server is not contacted
--- a/cmd/helm/upgrade.go
+++ b/cmd/helm/upgrade.go
@@ -254,7 +254,7 @@ func newUpgradeCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 
 	f := cmd.Flags()
 	f.BoolVar(&createNamespace, "create-namespace", false, "if --install is set, create the release namespace if not present")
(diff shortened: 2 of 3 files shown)


caruccio:
I was just wondering if it's a good idea to also add the flag `--namespace-annotations`?

robertsirc:
You are going to need to fix your DCO. And shouldn't there be test for this change?

caruccio:
Just added my signature. Sorry for that.

I didn't included tests since `--create-namespace` also doesn't have it.

robertsirc:
I understand your stance on the tests, I am sure a maintainer can weigh in on this. My suggestion is to have both inline documentation and documentation in the docs site on what this enhancement means. If you look at `cmd/helm/upgrade.go` there is a variable called `upgradeDesc` it makes sense to add that there.

cmd/helm/upgrade.go
```diff
@@ -252,6 +254,7 @@ func newUpgradeCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 
 	f := cmd.Flags()
 	f.BoolVar(&createNamespace, "create-namespace", false, "if --install is set, create the release namespace if not present")
+	f.StringToStringVarP(&namespaceLabels, "namespace-labels", "L", nil, "Labels that would be added to created namespace. Should be divided by comma. If --create-namespace is not set, this is ignored. Default is 'name=<namespace-name>'")
```
gjenkins8, line 257:
```suggestion
	f.StringToStringVar(&namespaceLabels, "namespace-labels", nil, "labels to add to the created namespace. Comma separated key=value pairs. Ignored if --create-namespace is not set (default 'name=<namespace-name>'")
```

I would propose no "shorthand" flag here. IMHO, shorthand flags are useful to save typing for commonly used flags. I don't think a `--namespace-labels` flag meets that criteria.

Also some suggestions for the verbiage.

cmd/helm/install.go
```diff
@@ -175,6 +175,7 @@ func newInstallCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 
 func addInstallFlags(cmd *cobra.Command, f *pflag.FlagSet, client *action.Install, valueOpts *values.Options) {
 	f.BoolVar(&client.CreateNamespace, "create-namespace", false, "create the release namespace if not present")
+	f.StringToStringVarP(&client.NamespaceLabels, "namespace-labels", "L", nil, "Labels that would be added to created namespace. Should be divided by comma. If --create-namespace is not set, this is ignored. Default is 'name=<namespace-name>'")
```
gjenkins8, line 178: ditto the suggestion on `upgrade.go`

gjenkins8 (changes requested):
I echo @robertsirc comment that we would want tests to ensure this feature remain unbroken in the future. It is unfortunate there are not tests for the `--create-namespace` currently. That is a past mistake that shouldn't be replicated.

PR #13208, opened 2024-07-26
Gathers up lint ignore functions as an Ignorer object

dpritchett:
- Moves ignoring out of `pkg/lint/rules` and up to `pkg/lint` because it doesn't quite feel like a lint rule
- Gathers up the main ignore parser and should-we-ignore-this-message functions under a new `Ignorer` object


dpritchett:
meant to target our working fork again, sorry

Issue #13243, opened 2024-08-10
Improvement of OpenSSF Scorecard Score


harshitasao:
Hi, I'm Harshita. I’m working with [CNCF and the Google Open Source Security Team for the GSoC 2024 term](https://github.com/cncf/mentoring/issues/1196). We are collaborating to enhance security practices across various CNCF projects. The goal is to improve security for all CNCF projects by both using OpenSSF Scorecards and implementing its security improvements.

As this project already has scorecard action, I'm here to increase the final score by going over each check. I've listed all of the checks where work needs to be done, in order of its criticality. I plan to submit each PR for each fix. Please let me know what you think and for which ones a PR is welcome that I will submit it ASAP.

**Current Score:** 8.5

**Scorecard report:** https://scorecard.dev/viewer/?uri=github.com/helm/helm

Here's a few checks we can work on to improve the project's security posture:

- [ ] [Token-Permissions](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions): Score = 0
    - The issue here is that workflows doesn’t have a top-level read-only permissions block. Scorecard is quite severe in this check: a single workflow without top-level permissions gets a 0/10 for the check.

- [ ] [Signed-Releases](https://github.com/ossf/scorecard/blob/main/docs/checks.md#signed-releases): Score = 8
    -  [SLSA provenance](https://slsa.dev/spec/v0.1/index) needs to be added, which can be done using [this](https://github.com/slsa-framework/slsa-github-generator), increasing the overall score to 10.

- [ ] [Pinned-Dependencies](https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies): Score = 9
    - [Github action](https://github.com/helm/helm/blob/main/.github/workflows/stale-issue-bot.yaml) is not pinned by hash, which is resulting in a low score.
    - Pinning dependencies to a specific hash rather than allowing mutable versions or ranges of versions improves supply chain security.

/cc @joycebrum @diogoteles08 @pnacht @nate-double-u

ivankatliarchuk:
Hey @harshitasao . I can't see any scorecards github actions here. Can you explain a bit how its configured for helm project?

harshitasao:
Hi @ivankatliarchuk Sorry, I didn't clarify that the scorecard action addition [PR](https://github.com/helm/helm/pull/13233) was opened. I've modified the issue content accordingly.

ivankatliarchuk:
Great stuff. Thank you ;-)

PR #13259, opened 2024-08-18
fix: fixed the token-permission and pinned-dependencies issue

harshitasao:
fixed the token-permission and pinned-dependencies issue reported by [scorecard](https://scorecard.dev/viewer/?uri=github.com/helm/helm).

Part of #13243 

None

**If applicable**:

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -20,6 +20,9 @@ on:
   schedule:
     - cron: '29 6 * * 6'
 
+permissions:  
+  contents: read
+
 jobs:
   analyze:
     name: Analyze
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -8,6 +8,9 @@ on:
     branches:
       - main
 
+permissions:  
+  contents: read
+
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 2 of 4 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -20,6 +20,9 @@ on:
   schedule:
     - cron: '29 6 * * 6'
 
+permissions:  
+  contents: read
```
gjenkins8, line 24: (for other reviewers) see: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token

gjenkins8 (approved):
This is actually a very good improvement, and so a nice catch/win from the OpenSSF Scorecard

From reading https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token, a leaked token would actually be fairly catastrophic. It would have had a wide range of write privileges.

Limiting the tokens scope to just read is a very good improvement here. LGTMe. Thanks!

mattfarina (approved):

robertsirc (approved):
This looks good to me, but I do question does stale-issue-bot.yaml need this as well?

joejulian (approved):

Issue #13258, opened 2024-08-17
Add ActiveHelp to Helm's shell completion support


marckhouzam:
Active Help is a framework provided by Cobra which allows a program to define messages (hints, warnings, etc) that will be printed during program usage. It aims to make it easier for your users to learn how to use your program. If configured by the program, Active Help is printed when the user triggers shell completion.

I would like to add ActiveHelp to the `helm` binary along the lines of this: https://asciinema.org/a/421868
I will be posting some PRs for this.

For more details on ActiveHelp please refer to Cobra's documentation: https://github.com/spf13/cobra/blob/main/site/content/active_help.md#active-help

marckhouzam:
More active help can be added to guide users if someone is interested

benoittgt:
Do we want to go further than https://github.com/helm/helm/pull/13261 ?

marckhouzam:
My  general approach is that ActiveHelp is valuable whenever there are no completions to suggest. #13261 covers the cases where no more arguments are accepted but you can imagine other situations. 

Here is a short video I made about ideas I had. That was a long time ago, so they may not apply anymore, but it can possibly give you ideas: https://asciinema.org/a/421868

If you decide to post PRs, you can tag me for a review.

PR #13261, opened 2024-08-19
ActiveHelp for cmds that don't take any more args

marckhouzam:

This PR starts adding some ActiveHelp to helm for specific situations, partially addressing #13258
ActiveHelp is a framework provided by Cobra which allows a program to define messages (hints, warnings, etc) that will be printed during program usage.  Active Help is printed when the user triggers shell completion.

For more details on ActiveHelp please refer to Cobra's documentation: https://github.com/spf13/cobra/blob/main/site/content/active_help.md#active-help

**Note: ActiveHelp is only shown when doing shell completion with `bash` or `zsh` (it is simply ignored for `fish` or `powershell`)**

This PR starts by adding the ActiveHelp message: `This command does not take any more arguments (but may accept flags).` to situations where a command no longer takes any more arguments.

The result is that when doing shell completion, instead of getting no suggestions for commands that take no more arguments, the user will instead be shown an informative message:

```
$ helm list <TAB>
This command does not take any more arguments (but may accept flags).
```
The situations where this ActiveHelp message will be triggered are the following:
```
helm completion bash|zsh|fish|powershell <TAB>
helm create <name> <TAB>
helm docs <TAB>
helm env <var> <TAB>
helm get all|hooks|manifest|metadata|notes|values <release> <TAB>
helm history <release> <TAB>
helm list <TAB>
helm plugin install <path|url> <TAB>
helm plugin list <TAB>
helm push <chart> <uri> <TAB>
helm test <release> <TAB>
helm repo add <repo> <url> <TAB>
helm repo index <dir> <TAB>
helm repo list <TAB>
helm rollback <release> <revision> <TAB>
helm show all|values|chart|readme|crds <chart> <TAB>
helm status <release> <TAB>
helm upgrade <release> <chart> <TAB>
helm verify <path>
helm version <TAB>
```

It is possible to turn off all ActiveHelp messages by doing: `export HELM_ACTIVE_HELP=0`.  This environment variable is automatically provided by Cobra.

**If applicable**:

### DIFF
--- a/cmd/helm/registry_login.go
+++ b/cmd/helm/registry_login.go
@@ -53,7 +53,7 @@ func newRegistryLoginCmd(cfg *action.Configuration, out io.Writer) *cobra.Comman
 		Short:             "login to a registry",
 		Long:              registryLoginDesc,
 		Args:              require.MinimumNArgs(1),
-		ValidArgsFunction: noCompletions,
+		ValidArgsFunction: cobra.NoFileCompletions,
 		RunE: func(_ *cobra.Command, args []string) error {
 			hostname := args[0]
 
--- a/cmd/helm/registry_logout.go
+++ b/cmd/helm/registry_logout.go
@@ -35,7 +35,7 @@ func newRegistryLogoutCmd(cfg *action.Configuration, out io.Writer) *cobra.Comma
 		Short:             "logout from a registry",
 		Long:              registryLogoutDesc,
 		Args:              require.MinimumNArgs(1),
-		ValidArgsFunction: noCompletions,
+		ValidArgsFunction: cobra.NoFileCompletions,
 		RunE: func(_ *cobra.Command, args []string) error {
 			hostname := args[0]
 			return action.NewRegistryLogout(cfg).Run(out, hostname)
--- a/cmd/helm/show.go
+++ b/cmd/helm/show.go
@@ -60,18 +60,17 @@ func newShowCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 	client := action.NewShowWithConfig(action.ShowAll, cfg)
 
 	showCommand := &cobra.Command{
-		Use:               "show",
-		Short:             "show information of a chart",
-		Aliases:           []string{"inspect"},
-		Long:              showDesc,
-		Args:              require.NoArgs,
-		ValidArgsFunction: noCompletions, // Disable file completion
+		Use:     "show",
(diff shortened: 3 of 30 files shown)


cmd/helm/show.go
```diff
@@ -60,18 +60,17 @@ func newShowCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 	client := action.NewShowWithConfig(action.ShowAll, cfg)
 
 	showCommand := &cobra.Command{
-		Use:               "show",
-		Short:             "show information of a chart",
-		Aliases:           []string{"inspect"},
-		Long:              showDesc,
-		Args:              require.NoArgs,
-		ValidArgsFunction: noCompletions, // Disable file completion
```
marckhouzam, line 68: Because this is a command with sub-commands, we don't need to define a `ValidArgsFunction` for it as Cobra will automatically complete the sub-commands and turn off file completion.

This wasn't the case when shell completion was first implemented for this command which is why we had to use `ValidArgsFunction`.

cmd/helm/registry_login.go
```diff
@@ -53,7 +53,7 @@ func newRegistryLoginCmd(cfg *action.Configuration, out io.Writer) *cobra.Comman
 		Short:             "login to a registry",
 		Long:              registryLoginDesc,
 		Args:              require.MinimumNArgs(1),
-		ValidArgsFunction: noCompletions,
+		ValidArgsFunction: cobra.NoFileCompletions,
```
marckhouzam, line 56: This turns off file completion but does not print the ActiveHelp message since an argument is expected, it is just that it should not be a file.  I'm plan to improve on this with a specific ActiveHelp message in a future PR.

cmd/helm/registry_logout.go
```diff
@@ -35,7 +35,7 @@ func newRegistryLogoutCmd(cfg *action.Configuration, out io.Writer) *cobra.Comma
 		Short:             "logout from a registry",
 		Long:              registryLogoutDesc,
 		Args:              require.MinimumNArgs(1),
-		ValidArgsFunction: noCompletions,
+		ValidArgsFunction: cobra.NoFileCompletions,
```
marckhouzam, line 38: This turns off file completion but does not print the ActiveHelp message since an argument is expected, it is just that it should not be a file. I'm plan to improve on this with a specific ActiveHelp message in a future PR.

mattfarina (approved):

Issue #12602, opened 2023-11-29
Helm pull does not use cache


olivierboudet:
Helm pull does not use local cache as other commands do. It fetch the chart and saves it in current directory (or directory specified with --destination) but does not keep it in cache. So it is downloaded each time we try to fetch it.

I think we should keep it in cache and just copy it to the wanted destination.

How to reproduce : 

* check that chart does not exist in cache :
```
ls -l ~/.cache/helm/repository/mongodb-14.3.0.tgz
"/home/olivier/.cache/helm/repository/mongodb-14.3.0.tgz": No such file or directory (os error 2)
```
* run helm pull : `helm pull bitnami/mongodb --version 14.3.0`
* check that chart does not exist in cache : 
```
ls -l ~/.cache/helm/repository/mongodb-14.3.0.tgz
"/home/olivier/.cache/helm/repository/mongodb-14.3.0.tgz": No such file or directory (os error 2)
```
* run helm template : `helm template bitnami/mongodb`
* check that chart exists in cache :
```
ls -l ~/.cache/helm/repository/mongodb-14.3.0.tgz
.rw-r--r-- 83k olivier 29 nov.  17:40 /home/olivier/.cache/helm/repository/mongodb-14.3.0.tgz
```

Output of `helm version`:

version.BuildInfo{Version:"v3.13.2", GitCommit:"2a2fb3b98829f1e0be6fb18af2f6599e0f4e8243", GitTreeState:"clean", GoVersion:"go1.20.10"}

Ithrael:
I can reproduce this issue. 
I believe the 'helm template' command is more focused on local operation, thus there is a need for caching. On the other hand, the original intent of 'helm pull' should be to fetch charts from the remote repository; considering this aspect, I think it is reasonable for 'helm pull' not to cache. Perhaps we could add an option to 'helm pull'? For example: 'helm pull xxx --cache'.

olivierboudet:
Yes, one of option `--cache` or `--no-cache` depending on which behavior should be the default seems reasonable to me.

Ithrael:
I originally wanted to add cache functionality for helm pull. If I specify --cache, helm pull will cache the chart in the repositoryCache directory and download the chart to a specified directory. Moreover, if there is already a chart in the repositoryCache directory, it will not download it, just copy it to the specified directory. Later I noticed that although helm template can cache, if the repositoryCache already has a chart, running helm template again will still re-download the chart and cache it in the repositoryCache directory. 

[...]

PR #13270, opened 2024-08-25
Re-use cache when downloading charts

### LINKED ISSUE #12607: Helm template will cache the chart but does not use the cache

rassie:

Previously, when a chart has been downloaded, it had been put in the cache. However, this has happened unconditionally, so that the cache hasn't actually been used for anything. This patch attempts to fix this and checks if the target file exists in the cache before downloading.

Fixes: #12607, #12602

NB: I'm not really familiar with Helm codebase and for that matter with Go language, I tried fixing a nuisance for myself. Any input is welcome, and I fully intend this PR to be a conversation starter (or an invocation of Cunnigham's law, if you will).

### DIFF
--- a/pkg/downloader/chart_downloader.go
+++ b/pkg/downloader/chart_downloader.go
@@ -92,25 +92,28 @@ func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *proven
 		return "", nil, err
 	}
 
-	g, err := c.Getters.ByScheme(u.Scheme)
-	if err != nil {
-		return "", nil, err
-	}
-
-	data, err := g.Get(u.String(), c.Options...)
-	if err != nil {
-		return "", nil, err
-	}
-
 	name := filepath.Base(u.Path)
 	if u.Scheme == registry.OCIScheme {
 		idx := strings.LastIndexByte(name, ':')
 		name = fmt.Sprintf("%s-%s.tgz", name[:idx], name[idx+1:])
 	}
+	g, err := c.Getters.ByScheme(u.Scheme)
+	if err != nil {
+		return "", nil, err
+	}
 
 	destfile := filepath.Join(dest, name)
-	if err := fileutil.AtomicWriteFile(destfile, data, 0644); err != nil {
-		return destfile, nil, err
+
+	_, err = os.Stat(destfile)
+	if err != nil {
+		data, err := g.Get(u.String(), c.Options...)
+		if err != nil {
+			return "", nil, err
+		}
+
+		if err := fileutil.AtomicWriteFile(destfile, data, 0644); err != nil {
+			return destfile, nil, err
+		}
 	}
 
 	// If provenance is requested, verify it.


rassie:
Oops, completely forgot about #12628, as luck would have it, got a notification about a new comment on that one. Will ponder whether this is obsolete if that PR gets revived.

rassie:
I'd like to leave this open until maybe #12628 is re-opened. It's slightly different in the execution and also adds a test, so this PR should be obsolete.

pkg/downloader/chart_downloader.go
```diff
@@ -92,25 +92,28 @@ func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *proven
...
+		return "", nil, err
+	}
 
 	destfile := filepath.Join(dest, name)
-	if err := fileutil.AtomicWriteFile(destfile, data, 0644); err != nil {
-		return destfile, nil, err
+
+	_, err = os.Stat(destfile)
+	if err != nil {
+		data, err := g.Get(u.String(), c.Options...)
```
mattfarina, line 109: What if you have two repos with a chart of the same name and version but their content is different?

The existing functionality handles that by downloading the latest. The new functionality would cause an issue but not report anything.

For a cache, we might need something that's content based (e.g., hash based). Name version pairs is not a safe cache key.

mattfarina (changes requested):
Thanks for the PR. Unfortunately, I found a problem where incorrect content could be used in a caching situation.

pkg/downloader/chart_downloader.go
rassie: > What if you have two repos with a chart of the same name and version but their content is different?

Isn't this a design problem with Helm's cache? I'm not calculating cache locations myself, and the way they are calculated allows for this scenario, i.e. if you happened to use multiple repos with the same chart name and version, they would use the same cache location. Right now the system works correctly only because Helm doesn't use the cache at all. If it did, this scenario would fail immediately and especially provenance checks would fail badly.

> For a cache, we might need something that's content based (e.g., hash based). Name version pairs is not a safe cache key.

Certainly, but this patch (and also #12628) builds upon existing cache functionality, reworking how the cache is set up would be a much larger undertaking. If that's desired, we could go that way, which would probably among others entail using ETags and similar techniques to verify the download.

mattfarina: > Isn't this a design problem with Helm's cache?

Yes. You are correct. Before the cache can be used the design problem needs to be fixed.

We are open to bigger changes to correct the design.

mkulke: my understanding is that the existing behaviour and the user expectations that are based on it are prohibitive for a cache. users would publish different variations of charts using the same version and expect them to be picked up in subsequent helm invocations. Maybe an opt-in toggle would be an option `--use-cache`? 

Here would be a broader discussion around this topic: https://github.com/helm/community/pull/185

antholeole: is what we're asking for even possible here? fundamentally you cannot know when to invalidate a cache from just the cache itself. Many a whitepaper has been written about this. The same argument could broadly be applied to any cache ever.

This is blocking using `helm template` in airgapped environments where serving should be done totally from cache

PR #13272, opened 2024-08-26
Adds HELM_OCI_DATE_EPOCH wich can be use to reproducibly push a Helm chart to an OCI registry

inteon:
Currently, when using `helm push`, the creation time of the Helm tar file is used. This makes it impossible to push the same chart twice reproducibly. This PR introduces HELM_OCI_DATE_EPOCH to overwrite the timestamp used.

This environment variable allows us to reproducibly create and push a Helm chart:
```bash
$ helm pull jetstack/cert-manager
$ HELM_OCI_DATE_EPOCH=0 helm push ./cert-manager-v1.15.3.tgz oci://localhost:5000/test123
Pushed: localhost:5000/test123/cert-manager:v1.15.3
Digest: sha256:ce96368a62b8c8a68518df374e853677210ffd7f34ca8eb57112eb6ac28b61c8
$ touch cert-manager-v1.15.3.tgz 
$ HELM_OCI_DATE_EPOCH=0 helm push ./cert-manager-v1.15.3.tgz oci://localhost:5000/test123
Pushed: localhost:5000/test123/cert-manager:v1.15.3
Digest: sha256:ce96368a62b8c8a68518df374e853677210ffd7f34ca8eb57112eb6ac28b61c8
```

### DIFF
--- a/pkg/pusher/ocipusher.go
+++ b/pkg/pusher/ocipusher.go
@@ -21,6 +21,7 @@ import (
 	"net/http"
 	"os"
 	"path"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


TerryHowe:
Wouldn't it be easier to just change the create time of the file?

TerryHowe:
I thought Ctim was ctime, so I may be off

jkroepke:
As I know, ctime is rarely used at Linux and mtime (mod time) is more common.

sabre1041:
> @sabre1041 -- thoughts on the above?

#12903 was introduced to achieve a constant creation time based on the time the package was created. I agree that the best approach forward is to correct the existing behavior instead of introducing a workaround/new functionality

JaanJah:
Status update on this?

inteon:
I'll close the PR since https://github.com/helm/helm/pull/13376 should have fixed the issue.

PR #13282, opened 2024-08-29
Helm should be able to wait when the number of replicas is being decreased

### LINKED ISSUE #13269: Helm upgrade for charts that contain statefulsets and set replicas to 0 and wait is true，It seems like helm not wait statefulset related pods deleted.
Output of `helm version`:
latest master version

Output of `kubectl version`:
None

openshift conatiner platform 4.14

I use helm upgrade a chart contains statefulset,  set the statefulset replicas to 0 and wait is ture，I found statefulset related pods not deleted when helm upgrade command execute finish，Then I found helm only check readyReplicas in ready.go，So Is there consider add a check currentReplicas condition bellow in ready.go to check statefulset that satisfy target status?

```go
	if int(sts.Status.CurrentReplicas) != replicas {
		c.log("StatefulSet is not ready: %s/%s. %d out of %d expected pods are ready", sts.Namespace, sts.Name, sts.Status.CurrentReplicas, replicas)
		return false
	}
```

rajhawaldar:
Helm should be able to wait when the number of replicas is being decreased
Fixes: #13269 

**If applicable**:

### DIFF
--- a/pkg/kube/ready.go
+++ b/pkg/kube/ready.go
@@ -425,6 +425,11 @@ func (c *ReadyChecker) statefulSetReady(sts *appsv1.StatefulSet) bool {
 		c.log("StatefulSet is not ready: %s/%s. %d out of %d expected pods are ready", sts.Namespace, sts.Name, sts.Status.ReadyReplicas, replicas)
 		return false
 	}
+
+	if int(sts.Status.CurrentReplicas) != replicas {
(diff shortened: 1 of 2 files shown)


gjenkins8 (commented):
I wasn't able to replicate the original issue: https://github.com/helm/helm/issues/13269#issuecomment-2614059488

An example to replicate, to manually confirm the problem/fix I think is needed before we make changes. Thanks!

Issue #13069, opened 2024-05-27
JSON Schema 2019 & 2020 Support


forestsword:
The current module helm uses for JSON schema validation only supports up to Draft-7. It's not possible to use more advanced funtionality like `if`/`else`/`then` (I think it might be in v7 but may not be implemented in the module).

https://github.com/qri-io/jsonschema provides the 2019 implemenation where at least if/else is possible (my use-case) but it too sadly has not updated since about 3 years. It does provide a fix for https://github.com/helm/helm/issues/12994 and supports my use-case. The api is also very similiar to the current module. It's almost a drop in replacement.

There is another project actively maintained and supports up to the latest draft version. It is also widely used by other projects: https://github.com/santhosh-tekuri/jsonschema

I've done an implementation of it with helm and it works as it should but the output is significantly different without modifying it on the helm side. But in both cases I find the errors less obvious unless you have some experience with jsonschema. Here's what it looks like unmodified for a missing required property and an if then statement with subschemas (there's really only 2 distinct failures to validate):

```
[ERROR] values.yaml: - :  doesn't validate with file:///home/me/chart/values.schema.json#
- : /required missing properties: 'serviceName'
- /serviceAccount: /properties/serviceAccount/$ref doesn't validate with '/$defs/ServiceAccount'
- /serviceAccount/secret: /properties/serviceAccount/$ref/properties/secret/$ref doesn't validate with '/$defs/Secret'
- /serviceAccount/secret: /properties/serviceAccount/$ref/properties/secret/$ref/allOf/1 allOf failed
- /serviceAccount/secret: /properties/serviceAccount/$ref/properties/secret/$ref/allOf/1/then if-then failed
- /serviceAccount/secret: /properties/serviceAccount/$ref/properties/secret/$ref/allOf/1/then/dependentRequired/type/0 property 'value' is required, if 'type' property exists
```

Condensing it to be more like previous behaviour I can get it to this:

```
- (root): missing properties: 'serviceName'
- /serviceAccount/secret: property 'value' is required, if 'type' property exists
```

#### Proposal

I propose that first the helm project replaces the current module to use the `qri-io` module. And then add support for https://github.com/santhosh-tekuri/jsonschema but put it behind a feature flag and/or check the schema version (if 2020 validate with this module else the other one). Then people can test and transition and then eventually maybe in a major version switch to the https://github.com/santhosh-tekuri/jsonschema as the default and only jsonschema validation module.

WDYT?

Output of `helm version`:

[...]

PR #13283, opened 2024-08-29
adding support for JSON Schema 2020

win-t:

It adds support for json schema ~~2019 and~~ 2020 as requested in https://github.com/helm/helm/issues/13069

the new json schema validator is only used when the chart explicitly wants to use it in `$schema` field, this means it will be backward-compatible with the existing chart that doesn't specify `$schema` field.

IMO, we cannot completely remove `github.com/xeipuuv/gojsonschema`, because some users might depend on its behavior when no `$schema` is specified, https://xkcd.com/1172/
image

**If applicable**:


win-t:
The focus of this PR is just to introduce schema 2019 and 2020 only. By doing this, we will unblock the users. 

We can improve it later if we want to remove `github.com/xeipuuv/gojsonschema` and fully depend on `github.com/santhosh-tekuri/jsonschema`, I just don't want this issue to become a blocker for introducing schema 2019/2020.

mattfarina:
@win-t can you rebase?

win-t:
> @win-t can you rebase?

sure

forestsword (approved):
LGTM I hope the maintainers will take a look

pkg/chartutil/jsonschema.go
```diff
@@ -91,3 +99,51 @@ func ValidateAgainstSingleSchema(values Values, schemaJSON []byte) (reterr error
...
+	if err != nil {
+		return false, nil
+	}
+	if url.Host == "json-schema.org" {
+		switch url.EscapedPath() {
+		case
+			"/draft-04/schema",
+			"/draft-06/schema",
+			"/draft-07/schema":
+			return false, nil
```
gjenkins8, line 124: Do we want to support `2019-09`? Given that it has been superseded by `2020-12`. And since Helm never supported it, there should be no reason someone would prefer it over `2020-12`?

pkg/chartutil/jsonschema.go
```diff
@@ -91,3 +99,51 @@ func ValidateAgainstSingleSchema(values Values, schemaJSON []byte) (reterr error
...
+	if err != nil {
+		return true, err
+	}
+
+	validator, err := compiler.Compile("file:///values.schema.json")
+	if err != nil {
+		return true, err
+	}
+
+	return true, validator.Validate(values)
```
gjenkins8, line 148: Can you confirm that `validator.Validate` returns human friendly errors please. We need to make sure that any schema validation errors propagate to the user.

pkg/chartutil/jsonschema.go
```diff
@@ -91,3 +99,51 @@ func ValidateAgainstSingleSchema(values Values, schemaJSON []byte) (reterr error
 
 	return nil
 }
+
+// keep the old behaviour for empty $schema or the one that defined in
+// https://github.com/xeipuuv/gojsonschema/blob/v1.2.0/draft.go#L46-L62
+func validateUsingNewValidator(valuesJSON, schemaJSON []byte) (bool, error) {
```
gjenkins8, line 105: This function does four things:
1. determines the jsonschema version via partial JSON load (exiting early if no `$schema` key present)
2. Determine/filter the jsonschema version to use 
3. Load the chart's schema and values as JSON
4. Validates the values against the schema

It might be worth factoring out some of these logical functionalities into separate functions?

gjenkins8 (commented):
I very much agree with the approach here in the PR:
1. Leaving support for the existing jsonschema version and library
2. Utilizing jsonschema's inbuilt [`$schema`](https://json-schema.org/understanding-json-schema/reference/schema#schema) keyword to determine the version

I have one comment on the support for 2019-09 (this seems unnecessary). And one comment on the implementation.

gjenkins8 (commented):
One more comment, is we will need to update the docs (helm-www) to detail support for the new versions

And I forgot to say, thanks for the PR!

pkg/chartutil/jsonschema.go
win-t: I don't see the necessity to split up this function. IMO, factoring out is only worth it if:
- reusability, in this case, no other code reuse it, we can split it later if in the future turns out that some of the code can be used by other code
- readability, IMO, in this case, splitting up the function harms readability, you need to jump between functions to understand it, it is only worth it if we have complex branching in the function, in this case, we don't have branching

pkg/chartutil/jsonschema.go
win-t: I have no preference about this, I only follow https://github.com/xeipuuv/gojsonschema/blob/v1.2.0/draft.go#L46-L62

pkg/chartutil/jsonschema.go
win-t: yes, see the test code

pkg/chartutil/jsonschema.go
gjenkins8: My vote would be to not support `2019-09`. I'm not seeing any benefit for doing so. If a reason occurs in the future, support could be included then.

The downsides to supporting included increased support burden, greater potential for bugs, etc.

pkg/chartutil/jsonschema.go
gjenkins8: nice, thanks

pkg/chartutil/jsonschema.go
TerryHowe: Another good reason would be to make the code more testable, but that doesn't seem to be the case here.

My take on @gjenkins8 comment is this is doing more than validating

win-t: > My take on @gjenkins8 comment is this is doing more than validating

I agree with this, but looking at the function `ValidateAgainstSingleSchema` before this PR, it does everything in a single function.

At first, I even put all the code inside that `ValidateAgainstSingleSchema` function, but then I tried my best not to touch existing code as much as possible, by separating this function, I only added 3 line to the original function

pkg/chartutil/jsonschema.go
win-t: I'm not sure how I can remove support for `2019-09` here, this switch case only chooses if the code should use the old library or the new library.

are you suggesting that `2019-09` should be validated using old library `github.com/xeipuuv/gojsonschema`, which clearly does not support at all

gjenkins8: > are you suggesting that 2019-09 should be validated using old library github.com/xeipuuv/gojsonschema, which clearly does not support at all

effectively "yes". In that by not supporting `2019-09` here and returning to the old library: the old behavior is retained. ie. the same behavior that `"$schema": "http://json-schema.org/2019-09/schema#"` would invoke prior to this PR (Which presumably will be to error)

pkg/chartutil/jsonschema_test.go
```diff
@@ -165,3 +180,68 @@ func TestValidateAgainstSchemaNegative(t *testing.T) {
...
+	subchartJSON := []byte(subchartSchema2020)
+	subchart := &chart.Chart{
+		Metadata: &chart.Metadata{
+			Name: "subchart",
+		},
+		Schema: subchartJSON,
+	}
+	chrt := &chart.Chart{
+		Metadata: &chart.Metadata{
+			Name: "chrt",
```
gjenkins8, line 194:
```suggestion
			Name: "chart",
```

typo? (and below)

pkg/chartutil/jsonschema.go
win-t: support to 2019 schema is removed, it now explicitly only support 2020

win-t:
> One more comment, is we will need to update the docs (helm-www) to detail support for the new versions

here the PR for it https://github.com/helm/helm-www/pull/1642

pkg/chartutil/jsonschema_test.go
win-t: I just copied it from other tests

pkg/chartutil/jsonschema.go
win-t: also, the function is only around 50 lines, not sure how to break it.

pkg/chartutil/jsonschema_test.go
gjenkins8: if "intentional" then no need

pkg/chartutil/jsonschema.go
```diff
@@ -91,3 +99,36 @@ func ValidateAgainstSingleSchema(values Values, schemaJSON []byte) (reterr error
...
+
+	return validator.Validate(values)
+}
+
+func schemaIs2020(schemaJSON []byte) bool {
+	var partialSchema struct {
+		Schema string `json:"$schema"`
+	}
+	_ = json.Unmarshal(schemaJSON, &partialSchema)
+	url, _ := url.Parse(partialSchema.Schema)
```
gjenkins8, line 132: We should not elide the error here.

gjenkins8: that said, iiuc, the only correct value for `$schema` is `https://json-schema.org/draft/2020-12/schema`

So IMHO a simple is correct / enough:
```suggestion
	return "https://json-schema.org/draft/2020-12/schema" == partialSchema.Schema
```

this also means if any other other portions of the URL are specified e.g. query params, the schema version string will (rightly and pedantically) be rejected

gjenkins8 (commented):
minor'ish comment on the new `schemaIs2020` function. getting close. thanks for the updates!

pkg/chartutil/jsonschema.go
win-t: TBH, I don't want to be pedantic here because I've seen people using `http://json-schema.org/draft/2020-12/schema#` in the wild

win-t: wdyt?, I don't see any harm in relaxing this thing

gjenkins8: > TBH, I don't want to be pedantic here because I've seen people using https://json-schema.org/draft/2020-12/schema#/ in the wild

They should IMHO use the correct value :) We shouldn't be in the business of trying to "fix" invalid values for the user. 

I think in Helm's case, they would introduce a wrong value into their `values.schema.json`, get an error, then fix the wrong value. Ideally the error will show the allowed values (which would be an improvement for this PR).

We should also validate that the path, query, fragment & user are empty. And scheme is `https`.

gjenkins8 (changes requested):
I think we need to tighten up the validation of the JSON schema version value

https://github.com/helm/helm/pull/13283/files#r1898613891

pkg/chartutil/jsonschema.go
win-t: updated, following your suggestion

win-t:
@gjenkins8 any follow up to this, I already update it

gjenkins8 (dismissed):
Thanks for the updates. I verified indeed helm built with this branch does support 2020-12, and continues to support draft-07

Can we also make a PR to the docs (<https://github.com/helm/helm-www>) please. The existing JSON Schema docs should be updated:
https://helm.sh/docs/topics/charts/#schema-files

And before merging this PR we should have a docs PR ready to go.

pkg/chartutil/jsonschema_test.go
```diff
@@ -165,3 +180,68 @@ func TestValidateAgainstSchemaNegative(t *testing.T) {
 		t.Errorf("Error string :\n`%s`\ndoes not match expected\n`%s`", errString, expectedErrString)
 	}
 }
+
+func TestValidateAgainstSchema2020(t *testing.T) {
```
scottrigby, line 184: 👍

pkg/chartutil/jsonschema_test.go
```diff
@@ -165,3 +180,68 @@ func TestValidateAgainstSchemaNegative(t *testing.T) {
...
+			"data": []any{"hello", 12},
+		},
+	}
+
+	if err := ValidateAgainstSchema(chrt, vals); err != nil {
+		t.Errorf("Error validating Values against Schema: %s", err)
+	}
+}
+
+func TestValidateAgainstSchema2020Negative(t *testing.T) {
```
scottrigby, line 211: 👍

scottrigby (dismissed):
lgtm too, thanks @win-t for the PR and tests as well 💯 

Does need to fix conflicts now as well before we can merge.

win-t:
I've merged back from upstream to this branch

sabre1041 (approved):
LGTM

mattfarina (approved):

PR #13287, opened 2024-08-30
Adding golang govulncheck action

robertsirc:
I added the golang govulncheck action.

**If applicable**:


mattfarina (approved):

.github/workflows/build-test.yml
```diff
@@ -24,3 +24,15 @@ jobs:
...
+  govulncheck:
+    name: govulncheck
+    runs-on: ubuntu-latest
+    steps:
+      - name: Setup Go
+        uses: actions/setup-go@0a12ed9d6a96ab950c8f026ed9f722fe0da7ef32 # pin@5.0.2
+        with:
+          go-version: '1.22'
+      - name: govulncheck
+        uses: golang/govulncheck-action@dd0578b371c987f96d1185abb54344b44352bd58 # pin@1.0.3
```
gjenkins8, line 36: To put it on the record, I think there is a limitation in the way this action is implemented. That is, the plugin will alert on any CVE present in a change/PR. Whether the change introduced the CVE or not. It will also _not_ alert if a CVE is discovered.

So what I think is actually desirable:
1. Changes (ie. PRs) should blocked, if and only if they introduce a vulnerability. And not because they carry the existing dependency set with a pre-existing vulnerability. This in effect requires subtracting existing vulnerabilities from the target branch from the change branch.
2. If a CVE/vulnerability is discovered. There should be a proactive alert and remediation, without waiting for an unrelated PR. Dependabot serves most of this purpose today. But to my knowledge it doesn't integrate with govulncheck to e.g. perhaps more aggressively promote PRs which contain actual fixes (it uses the standard dependency graph mechanism to find updates). And to this point, it waits until there is an upstream package fix before it does anything. Where perhaps there could e.g. be a workaround.

So, even with the above, we might still want to utilize govulncheck here. Since it might be better than nothing. We (maintainers) probably want to be able to override/ignore any findings it reports (since they may be unrelated to the change on hand)

(and to note, this is based all on my understanding of the way the govulncheck action works. Which could be wrong of course)

I could/should also at least raise an issue on the govulncheck action. To see if some of these could be idea couldbe implemented (if I'm not mistaken about the workings here)

scottrigby:
Also see https://github.com/helm/helm/pull/12045

robertsirc:
@gjenkins8 and @mattfarina I updated this corrected this PR. I removed govulncheck from buid-test, added two more action. First a daily action that runs once a day to check for vulnerabilities. The second will check for vulnerabilities only if `go.sum` has a change to it.

mattfarina (approved):

gjenkins8 (approved):
Nice, thanks for updating!

joejulian (approved):

Issue #13284, opened 2024-08-29
.helmignore does not ignore symlinks


graywolf-s1:
Output of `helm version`:

```
version.BuildInfo{Version:"v3.15.4", GitCommit:"fa9efb07d9d8debbb4306d72af76a383895aa8c4", GitTreeState:"clean", GoVersion:"go1.22.6"}
```

Output of `kubectl version`:

```
Client Version: version.Info{Major:"1", Minor:"24", GitVersion:"v1.24.17", GitCommit:"22a9682c8fe855c321be75c5faacde343f909b04", GitTreeState:"archive", BuildDate:"2024-01-11T20:07:07Z", GoVersion:"go1.20.7", Compiler:"gc", Platform:"darwin/arm64"}
Kustomize Version: v4.5.4
```

Cloud Provider/Platform (AKS, GKE, Minikube etc.): 

```
AWS, self-managed cluster
```

When I put a symlink (broken) into the .helmignore file, `helm template` still fails.  Reproduction:

```
$ helm create test
Creating test
$ cd test
$ cd templates
$ ln -s foo bar
$ cd ..
$ echo bar >>.helmignore
$ 
$ helm template .
Error: error evaluating symlink /tmp/repro/test/templates/bar: lstat /private/tmp/repro/test/templates/foo: no such file or directory
```

ronantakizawa:
@graywolf-s1 @yxxhero I was able to replicate the error mentioned by @graywolf-s1. I implemented a fix that ignores broken symlinks, where it points to a non-existent file (like whats causing the issue in the example).

sielaq:
Moreover, even existing (not broken) symlinks, which are not related to helm
are not fully ignored by `.helmignore` always generating in stderr:
```
walk.go:74: found symbolic link in path: xxx resolves to xxx. Contents of linked file included and used
```

graywolf:
Further activity.

graywolf:
Further activity.

graywolf:
Further activity.

graywolf:
Stupid bot.

497a:
This is still an issue.

scottrigby:
Can you reproduce this issue with Helm 4?

graywolf:
> Can you reproduce this issue with Helm 4?

Yes.

graywolf:
Further activity.

graywolf:
Still stupid bot.

vimalyad:
[...]

PR #13293, opened 2024-09-01
fix: .helmignore ignores broken symlinks

ronantakizawa:
.helmignore now ignores broken symlinks, which is a symlink points to a file or directory that doesn't exist. Closes issue #13284.


internal/sympath/walk.go
```diff
@@ -69,7 +67,8 @@ func symwalk(path string, info os.FileInfo, walkFn filepath.WalkFunc) error {
 	if IsSymlink(info) {
 		resolved, err := filepath.EvalSymlinks(path)
 		if err != nil {
-			return errors.Wrapf(err, "error evaluating symlink %s", path)
+			log.Printf("Skipping broken symlink: %s", path) // Log broken symlink
+			return nil
```
mattfarina, line 71: I have three thoughts about this change...

1. This doesn't just change the behavior for JUST .helmignore files. It changes it for everything. Even those that are not ignored.
2. There is a situation where a broken symlink would have stopped rendering because of bad output and this change will allow for bad output because the broken symlink will be ignored and movement will continue. This could cause cases for an accidental bad deployment.
3. This brings in the ability to easily package broken symlinks. There are security implications to using symlinks as they can be easily abused and this makes including many broken ones (on the packagers system) possible.

I think there needs to be a better route where it's more directly linked to the helmignore feature rather than general walking.

mattfarina (changes requested):
I'm having trouble replicating the initial problem in the issue. I will ask some questions over there to get more detail.

internal/sympath/walk.go
mattfarina: I did some more digging and am able to reproduce this. Will need to do some more testing to make sure there are no security issues with it AND to make sure nothing breaks.

Issue #13318, opened 2024-09-11
High Severity Vulnerabilities [CVE-2024-34156,CVE-2024-34155,CVE-2024-34158] detected.


Rchanger:
```
 Library │ Vulnerability  │ Severity │ Status │ Installed Version │ Fixed Version  │                            Title                            │
├─────────┼────────────────┼──────────┼────────┼───────────────────┼────────────────┼─────────────────────────────────────────────────────────────┤
│ stdlib  │ CVE-2024-34156 │ HIGH     │ fixed  │ 1.22.6            │ 1.22.7, 1.23.1 │ encoding/gob: golang: Calling Decoder.Decode on a message   │
│         │                │          │        │                   │                │ which contains deeply nested structures...                  │
│         │                │          │        │                   │                │ https://avd.aquasec.com/nvd/cve-2024-34156                  │
│         ├────────────────┼──────────┤        │                   │                ├─────────────────────────────────────────────────────────────┤
│         │ CVE-2024-34155 │ MEDIUM   │        │                   │                │ go/parser: golang: Calling any of the Parse functions       │
│         │                │          │        │                   │                │ containing deeply nested literals...                        │
│         │                │          │        │                   │                │ https://avd.aquasec.com/nvd/cve-2024-34155                  │
│         ├────────────────┤          │        │                   │                ├─────────────────────────────────────────────────────────────┤
│         │ CVE-2024-34158 │          │        │                   │                │ go/build/constraint: golang: Calling Parse on a "// +build" │
│         │                │          │        │                   │                │ build tag line with...                                      │
│         │                │          │        │                   │                │ https://avd.aquasec.com/nvd/cve-2024-34158   

```
Output of `helm version`: v3.15.4

Output of `kubectl version`: NA

Cloud Provider/Platform (AKS, GKE, Minikube etc.):  NA

Rchanger:
upgrading to latest patch version of go will resolve the issue 
https://go.dev/doc/devel/release#go1.22.minor

gjenkins8:
fixed by: https://github.com/helm/helm/pull/13323

(labeled question/support as I believe these don't affect Helm)

PR #13319, opened 2024-09-11
Update go version to 1.22.7 to fix [CVE-2024-34156],[CVE-2024-34155]

Rchanger:
update go version

https://github.com/helm/helm/issues/13318

**If applicable**:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module helm.sh/helm/v3
 
-go 1.22.0
+go 1.22.7
 
 require (
 	github.com/BurntSushi/toml v1.4.0


kanchwala-yusuf:
@mattfarina, @robertsirc Can you please approve this PR? This PR fixes some of the CVEs https://nvd.nist.gov/vuln/detail/CVE-2024-34156

go.mod
```diff
@@ -1,6 +1,6 @@
 module helm.sh/helm/v3
 
-go 1.22.0
+go 1.22.7
```
mattfarina, line 3: This changes the _minimum_ version used to build Helm from a compatibility standpoint. That's what the `go` directive.

We don't want to change the minimum version of compatibility. We want to build with the newer version.

mattfarina (changes requested):
The build system should use the latest patch version of Go when we do a release. A release was created today and _should_ have used Go 1.22.7. For some reason, that did not happen. We are investigating the issue.

Updates through the build system is what we use instead of putting specific patch versions as the minimum version in the go.mod file.

go.mod
Rchanger: got it will close this PR, thanks @mattfarina

PR #13323, opened 2024-09-11
bumping version to 1.22.7

robertsirc:

**If applicable**:


mattfarina (approved):

mattfarina:
Note, the reason for this change is that the [GitHub runners currently cache 1.22.6](https://github.com/actions/runner-images/blob/b60b7ab30f16e8d7048a8f3d437ddefec573aae9/images/ubuntu/Ubuntu2404-Readme.md#go). When a patch release is not specified the cached version is used. In order to use 1.22.7 (with the standard library CVE fix) you need to specify the patch version in builds.

technosophos (approved):

technosophos:
Totally makes sense. So going forward we will pin to a specific patch version?

PR #13326, opened 2024-09-12
Revert "Improve helm dependency update performance"

mattfarina:
This is a draft to test if this PR caused a regression.

Reverts helm/helm#11726


mattfarina:
Confirmed that the original PR caused the issue.

PR #13328, opened 2024-09-12
adding check-latest:true

robertsirc:
I am fixing the go setup to check for latest. Based on the docs [here](https://github.com/actions/setup-go).

tested it with my actions and it resolves correctly:
`Run actions/setup-go@0a12ed9d6a96ab950c8f026ed9f722fe0da7ef32
Setup go version spec 1.22
Attempting to resolve the latest version from the manifest...
matching 1.22...
Resolved as '1.22.[7](https://github.com/robertsirc/helm/actions/runs/10835235498/job/30066292590#step:3:8)'`

**If applicable**:

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -20,7 +20,8 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@0a12ed9d6a96ab950c8f026ed9f722fe0da7ef32 # pin@5.0.2
         with:
-          go-version: '1.22.7'
+          go-version: '1.22'
+          check-latest: true
       - name: Test source headers are present
         run: make test-source-headers
       - name: Run unit tests
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -18,7 +18,8 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@0a12ed9d6a96ab950c8f026ed9f722fe0da7ef32 # pin@5.0.2
         with:
-          go-version: "1.22.7"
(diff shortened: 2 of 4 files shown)


gjenkins8 (approved):

mattfarina (approved):

.github/workflows/build-test.yml
```diff
@@ -20,7 +20,8 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@0a12ed9d6a96ab950c8f026ed9f722fe0da7ef32 # pin@5.0.2
         with:
-          go-version: '1.22.7'
+          go-version: '1.22'
+          check-latest: true
```
chenrui333, line 24: why not just use `go-version-file: go.mod`

robertsirc: This is a good point. I can address that in another PR.

TerryHowe: https://github.com/helm/helm/pull/13347

mattfarina: The `go` directive in the `go.mod` file sets the minimum version of Go that _can_ be used to compile the code. When we build Helm, we want to use the latest patch version on the minor version we are using. This way we get all of the bug and security fixes.

If Helm does use the `go.mod` file (either the `go` or `toolchain` directives) than we will need to update this file with every new Go release to ensure we build against the latest patch version of Go.

I'm not in favor of this extra work. If we went this way we would need to automate the updates to the `go.mod` file.

TerryHowe: I think with the change I made, it selects the latest version of whatever is specified in the go.mod file.

PR #13338, opened 2024-09-15
Remove the Kubernetes configuration file permissions check

### LINKED ISSUE #13320: Kubernetes configuration file is group-readable
Normally when you try to execute `helm` commands as a normal user you will get `permission denied`, because the config file is owned by root and not readable:
```
$ helm list
Error: Kubernetes cluster unreachable: error loading config file "/etc/rancher/k3s/k3s.yaml": open /etc/rancher/k3s/k3s.yaml: permission denied
```

The only solution to have a config file that all users can use that I see:
```
sudo chmod 644 /etc/rancher/k3s/k3s.yaml
```

But then you get these warnings:
```
WARNING: Kubernetes configuration file is group-readable. This is insecure. Location: /etc/rancher/k3s/k3s.yaml
WARNING: Kubernetes configuration file is world-readable. This is insecure. Location: /etc/rancher/k3s/k3s.yaml
```

I do not understand the logic here.\
How can I use a `k3s` config file for all users and be able to use `helm` without warnings?

yardenshoham:
See https://github.com/helm/helm/issues/13320#issuecomment-2346724311

We remove an out-of-scope check.

- Resolves https://github.com/helm/helm/issues/13320


gjenkins8 (approved):

sundaram2021 (commented):
This PR correctly addresses an outdated security check that no longer aligns with modern Helm use cases. The reasoning for removal is solid, especially considering that other Kubernetes tools no longer enforce this check. By removing it, Helm avoids generating unnecessary warnings in valid deployment scenarios (e.g., mounted kubeconfig secrets).

Suggested Improvements:
Ensure documentation is updated, informing users that permissions checks are no longer enforced.
If possible, provide a note about alternative ways users can enforce security checks on kubeconfig files outside Helm.

TerryHowe:
i would think release notes would be enough documentation for this change.

TerryHowe (approved):
/lgtm

but IANAM

robertsirc (approved):
LGTM

mattfarina (approved):
There are a couple reasons to remove this check:

1. When Helm is used in a container running in Kubernetes and the kubeconfig is mounted as a file via a secret you cannot control the permissions and these messages show up. This is a common enough occurance.
2. Helm is not the creator or maintainer of this configuration file. The tools that "own" this file are not providing these messages so why should Helm?

Note, this check was added as part of a security review and was debatable at the time.

PR #13340, opened 2024-09-18
fix: remove download logs in templates

bjorand:

`helm template --dependency-update` prints the manifest and logs on stdout making the yaml output unusable:
```
Saving 2 charts
Downloading datadog-operator from repo https://helm.datadoghq.com
Deleting outdated charts
---
# Source: datadog-operator/charts/datadog-operator/templates/service_account.yaml
apiVersion: v1
kind: ServiceAccount
...
```

### DIFF
--- a/cmd/helm/install.go
+++ b/cmd/helm/install.go
@@ -268,7 +268,7 @@ func runInstall(args []string, client *action.Install, valueOpts *values.Options
 			err = errors.Wrap(err, "An error occurred while checking for chart dependencies. You may need to run `helm dependency build` to fetch missing dependencies")
 			if client.DependencyUpdate {
 				man := &downloader.Manager{
-					Out:              out,
+					Out:              os.Stderr,
 					ChartPath:        cp,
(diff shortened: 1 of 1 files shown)


TerryHowe (approved):
/lgtm

but IANAM

robertsirc:
Hello thank you for your PR. Looking at this change you are trying to change basic output `out` to `os.Stdrr`. As this isn't an error, warning or diagnostics, so I am not sure this makes sense. This type of change will break for anyone who parses the output.

TerryHowe:
A flag to quiet the output instead?

bjorand:
The expected behavior of `helm template` is that it outputs valid yaml to stdout, then you can pipe it to something or save the manifests.
It is not currently the expected behavior as it produces invalid yaml with logs of the dependency manager in the output.
I think it will break nothing: if people are filtering already on stdout these unexpected lines to fix and get a clean yaml, then with this PR they will have nothing to filter so it should change nothing for them.

PR #13343, opened 2024-09-19
Add annotations and dependencies to get metadata output

### LINKED ISSUE #12686: Helm get metadata command - info omitted

niladrih:
Closes https://github.com/helm/helm/issues/12686

Sample outputs for empty annotations/dependencies..
TABLE:
```yaml
CHART: foo
VERSION: 0.1.0-beta.1
APP_VERSION: 1.0
ANNOTATIONS: 
DEPENDENCIES: 
NAMESPACE: default
REVISION: 1
STATUS: deployed
```
YAML:
```yaml
appVersion: 1.0
chart: foo
deployedAt: "2024-10-04T10:55:24+05:30"
name: foo
namespace: default
revision: 1
status: deployed
version: 0.1.0-beta.1
```
JSON:
```json
{"name":"foo","chart":"foo","version":"0.1.0-beta.1","appVersion":"1.0","namespace":"default","revision":1,"status":"deployed","deployedAt":"2024-10-04T11:25:10+05:30"}
```
**N.B.:** The JSON and YAML outputs have omitempty

**What this PR does / why we need it:**
The output of helm get metadata includes a subset of the fields contained in the chart.Metadata struct. This change adds the values of the annotations field and the dependencies field to the output.

**If applicable:**


cmd/helm/testdata/output/get-metadata.txt
```diff
@@ -2,6 +2,8 @@ NAME: thomas-guide
 CHART: foo
 VERSION: 0.1.0-beta.1
 APP_VERSION: 1.0
+ANNOTATIONS: category=web-apps,supported=true
```
TerryHowe, line 5: is this order deterministic?

niladrih: Yes it is. The slice is sorted in the [FormatLabels](https://github.com/kubernetes/apimachinery/blob/v0.31.0/pkg/labels/labels.go#L87) function from k8s.io/apimachinery.

Ref: https://github.com/kubernetes/apimachinery/blob/v0.31.0/pkg/labels/labels.go#L46

The DEPENDENCIES list is sorted in this code change, in the FormattedDepNames method.

pkg/action/get_metadata.go
```diff
@@ -57,13 +65,30 @@ func (g *GetMetadata) Run(name string) (*Metadata, error) {
...
+		Revision:     rel.Version,
+		Status:       rel.Info.Status.String(),
+		DeployedAt:   rel.Info.LastDeployed.Format(time.RFC3339),
 	}, nil
 }
+
+// FormattedDepNames formats metadata.dependencies names into a comma-separated list.
+func (m *Metadata) FormattedDepNames() string {
+	if len(m.Dependencies) == 0 {
+		return "<none>"
```
sabre1041, line 84: To align with other fields that do not produce a result (a good example is App Version), recommend following the same pattern and omitting any output

sabre1041 (commented):
One suggested action

pkg/action/get_metadata.go
niladrih: Done. I've left the omitempty in for the json and yaml outputs. Would you rather that I let them displace `null`?

TerryHowe: I think he meant that actually omit annotations if there are none.

niladrih: In table view, for an empty values, appVersion shows `APP_VERSION: `. I've gone with the same output here for ANNOTATIONS and DEPENDENCIES.

In YAML/JSON view appVersion displays the empty output explicitly `appVersion: ""`. I've gone with omitempty in these cases, viz. the annotationa and the dependencies fields would be missing if they are null.

sabre1041 (approved):
LGTM

gjenkins8 (approved):

PR #13358, opened 2024-10-01
feat: Add --log functionality for additional commands and resources 

cgetzen:

- Jobs with the `"helm.sh/hook": "test"` annotation will now work with `helm test --logs`. This previously caused helm to error due to searching for a pod with the same name as the job.
- Upgrades that have Pods or Jobs that contain an upgrade annotation (e.g. `"helm.sh/hook": pre-upgrade`) can now run with `--logs` to print the logs once the upgrade completes.

This is important because the cause of the hook related upgrade failures is not possible to determine with the helm output, and requires investigating kubernetes job logs.

I'm unsure if this code is possible to unit test because it relies on the output of annotated pods. If there is an integration test framework for helm, I will add tests to that. I was able to manually test with the following helm single-file chart:

```yaml
apiVersion: batch/v1
kind: Job
metadata:
  name: post-install
  annotations:
    "helm.sh/hook": post-install
spec:
  template:
    spec:
      containers:
      - name: test
        image: ubuntu
        command: ["echo",  "hello world"]
      restartPolicy: Never
  backoffLimit: 1
---
apiVersion: batch/v1
kind: Job
metadata:
  name: test
  annotations:
    "helm.sh/hook": test
spec:
  template:
    spec:
      containers:
      - name: test
        image: ubuntu
        command: ["echo",  "I'm a test!"]
      restartPolicy: Never
  backoffLimit: 1
---
apiVersion: v1
kind: Pod
metadata:
  name: backwards-compatible
  annotations:
    "helm.sh/hook": test
spec:
  containers:
  - name: test
    image: ubuntu
    command: ["echo",  "I'm a test pod!"]
  restartPolicy: Never
```

**If applicable**:

### DIFF
--- a/cmd/helm/release_testing.go
+++ b/cmd/helm/release_testing.go
@@ -90,7 +90,7 @@ func newReleaseTestCmd(cfg *action.Configuration, out io.Writer) *cobra.Command 
 
 	f := cmd.Flags()
 	f.DurationVar(&client.Timeout, "timeout", 300*time.Second, "time to wait for any individual Kubernetes operation (like Jobs for hooks)")
-	f.BoolVar(&outputLogs, "logs", false, "dump the logs from test pods (this runs after all tests are complete, but before any cleanup)")
+	f.BoolVar(&outputLogs, "logs", false, "dump the logs from test jobs or pods (this runs after all tests are complete, but before any cleanup)")
 	f.StringSliceVar(&filter, "filter", []string{}, "specify tests by attribute (currently \"name\") using attribute=value syntax or '!attribute=value' to exclude a test (can specify multiple or separate values with commas: name=test1,name=test2)")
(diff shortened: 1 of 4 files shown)


robertsirc (changes requested):
The build-test and golinter actions are failing.

robertsirc:
Hello thank you for your PR. There are some issues with the GH actions could you please address that so we can review?

cgetzen:
Hi @robertsirc, the latest commit looks successful on my end. If there's any links to failures I can take a look!

robertsirc:
We appreciate your PR. Your account is coming up `Unverified` could you correct that?

cgetzen:
@robertsirc is there anything else that is needed to get a review? This PR does a few things, and I'd be happy to strip it down to do a single thing if that helps the review process -- I'm mostly interested in getting log output for helm upgrades.

gjenkins8 (changes requested):

Issue #13363, opened 2024-10-02
Helm hook (pre-install) is not able to delete k8s-resouces through hook-delete-policy - stopped working after upgrade to 3.14.0/3.16.1 from 3.13.3


radekcz:
We use Helm-Hooks, especially a "pre-install" hook, as a part of our deployments that works well with Helm version **3.13.3**. The goal of this pre-install hook/job is to mount the filesystem, do some work on this filesystem there and that's it. It means there are PersistentVolume (PV), PersistentVolumeClaim (PVC) k8s-resources as parts of the pre-install hook.

After upgrading to Helm **3.14.0 and above** (also tested on 3.16.1), the Helm upgrade command fails:

```client.go:142:  [debug] creating 1 resource(s)
client.go:142: [debug] creating 1 resource(s)  
client.go:142:  [debug] creating 1 resource(s)  
client.go:712: [debug] Watching for changes to Job myapp-init-pre-install-job with timeout of 5m0s
client.go:740:  [debug] Add/Modify event for myapp-init-pre-install-job: ADDED
client.go:779:  [debug] myapp-init-pre-install-job: Jobs active: 0, jobs failed: 0, jobs succeeded: 0
client.go:740:  [debug] Add/Modify event for myapp-init-pre-install-job: MODIFIED
client.go:779:  [debug] myapp-init-pre-install-job: Jobs active: 1, jobs failed: 0, jobs succeeded: 0
client.go:740:  [debug] Add/Modify event for myapp-init-pre-install-job: MODIFIED
client.go:779:  [debug] myapp-init-pre-install-job: Jobs active: 0, jobs failed: 0, jobs succeeded: 0
client.go:740:  [debug] Add/Modify event for myapp-init-pre-install-job: MODIFIED
client.go:486:  [debug] Starting delete for "myapp-init-job-pv" PersistentVolume
wait.go:104:  [debug] beginning wait for 1 resources to be deleted with timeout of 5m0s
Error: failed pre-install: context deadline exceeded
helm.go:86:  [debug] failed pre-install: context deadline exceeded
```

The reason for the failure is that Helm is trying to delete PV first but there is still PVC existing, so it's clear that is not possible to delete PV. As I mentioned, this deployment (with hooks) works nicely with 3.13.3 Helm version, but it started failing on 3.14.0 and later.

We also use "helm.sh/hook-weight" to determine the order of creating resources like:

```apiVersion: batch/v1
kind: Job
metadata:
[...]

PR #13365, opened 2024-10-02
fix(hooks): correct hooks delete order

idsulik:
Fixed deletion hooks order, deletion hooks should execution in reverse order. 
Issue https://github.com/helm/helm/issues/13363

**If applicable**:

### DIFF
--- a/pkg/action/hooks.go
+++ b/pkg/action/hooks.go
@@ -99,7 +99,8 @@ func (cfg *Configuration) execHook(rl *release.Release, hook release.HookEvent, 
 
 	// If all hooks are successful, check the annotation of each hook to determine whether the hook should be deleted
 	// under succeeded condition. If so, then clear the corresponding resource object in each hook
-	for _, h := range executingHooks {
+	for i := len(executingHooks) - 1; i >= 0; i-- {
+		h := executingHooks[i]
 		if err := cfg.deleteHookByPolicy(h, release.HookSucceeded, timeout); err != nil {
 			return err
 		}


pkg/action/hooks.go
```diff
@@ -21,6 +21,7 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+	"helm.sh/helm/v3/pkg/releaseutil"
```
robertsirc, line 24: The linter doesn't like this not having a line between these two lines.

robertsirc (changes requested):
Hello, thank you for your PR, the linter is erroring out you can see these errors by running `make test`.

idsulik:
> Hello, thank you for your PR, the linter is erroring out you can see these errors by running `make test`.

Hi @robertsirc , thank you. Pushed commit with style fix and unit test

robertsirc (approved):
LGTM: I pulled this down ran all the test then tested this with the hook-weight annotation.

robertsirc (dismissed):

TerryHowe (approved):
/lgtm

pkg/action/hooks.go
```diff
@@ -97,9 +98,11 @@ func (cfg *Configuration) execHook(rl *release.Release, hook release.HookEvent,
 		h.LastRun.Phase = release.HookPhaseSucceeded
 	}
 
+	hooksToDelete := releaseutil.SortHooksToDelete(executingHooks)
```
gjenkins8, line 101: why are we "re-sorting" the hooks here? If the intention is to delete in reverse order, why not just iterate `executingHooks` in reverse order?

idsulik: they should be sorted by kind and by weight, here it sorts only by weight

mattfarina: Sorting by type happens when resources are rendered. For example, when they are rendered for install like here... https://github.com/helm/helm/blob/a2cffa6e3f9f921d8083d187d55471615e5b7515/pkg/action/action.go#L168

Those sorted resources are passed into this function (`execHook`). Then, they are sorted again by weight and name. See the implementation at https://github.com/helm/helm/blob/a2cffa6e3f9f921d8083d187d55471615e5b7515/pkg/action/hooks.go#L112-L121.

Any sorting in here again started with resources sorted by type. The difference between sorting the original list again and using the existing list is the part by name.

It should be possible to save on work by iterating the existing sort in reverse order. It would be reverse weight / alpha. It would also save on exposing more of a public API.

mattfarina (changes requested):

mattfarina:
Note, I'm happy to answer any questions around this and I like the general direction.

pkg/action/hooks.go
idsulik: Maybe I'm missing something, I don't understand why @mattfarina and @gjenkins8 wrote about the initial sorting by type if we rewrite the inital sorting by `hookByWeight`, it doesn't keep the original sorting, it rewrites the sorting order:  

- before `hookByWeight` - the items sorted by type(btw, they will be sorted by install order while we should sort by uninstall order)
- after `hookByWeight` - the items sorted by weight and name

mattfarina: I was showing how current sorting works so that people could follow along. Especially some newer maintainers who are on this PR.

What's wrong with doing something like...

```go
for i, := len(executingHooks)-1; i >= 0; i-- {
	if err := cfg.deleteHookByPolicy(executingHooks[i], release.HookSucceeded, timeout); err != nil {
		return err
	}
}
```

This takes the existing `executingHooks` and deals with them in reverse order. Note, I did not test this.

This avoids resorting and just iterates the existing list backwards.

idsulik: @mattfarina we can't just iterate in reverse order, because the hooks are **sorted by weight and name** only, while the hooks should be sorted **by weight and type**

idsulik: @mattfarina hi! check please the comment above

idsulik: @mattfarina because the `executingHooks` are sorted **only**(because it overwrites the sort by kind) by `weight` and if the `weight` is zero(or just the same), then you can't guarantee what order the resources will have.

idsulik: @mattfarina I apologize. I just tried to reproduce the issue that led me to add my custom sorting instead of reversing the order for deletion, but I couldn’t. Either I did something wrong back then, or I don’t remember how to reproduce it. I’ve pushed changes where I simply reverse the order for deletion.

mattfarina (approved):

Issue #12973, opened 2024-04-22
Support reproducible builds to an OCI registry


stevehipwell:
I'd like to be able to support creating Helm charts targeting an OCI registry as a reproducible build, which means being able to pass in the value for the `org.opencontainers.image.created` OCI annotation to `helm push`. This would build on the changes made in #12903 but work for cases where the package was being recreated.

An example use case for this would be as follows.

```bash
chart="my-chart"
chart_path="./charts/${chart}"
version="0.1.0"

date_epoch="$(git log -1 --format='%ct' "${chart_path}")"
created_date="$(date -u --date="@${date_epoch}" "+%Y-%m-%dT%H:%M:%SZ" 2>/dev/null)"

helm package "${chart_path}" --version "${version}"
helm push --created-date "${created_date}" "${chart}-${version}.tgz" "oci://my-registry.io/my-repo/${chart}"
```

CC @sabre1041

sabre1041:
@stevehipwell been thinking about the different options at hand and comparing existing container/artifact packaging tools.

Docker does not natively support adding creation timestamps via CLI options, but Podman does. For Docker, labels (associated with the `org.opencontainers.image.created`) can be add during build time. ORAS does provide the ability to specify annotations that are applied to the artifact either from inline parameters or from a file.

There are really two options that could be applied at artifact push time:
* The change proposed here where the build time is provided inline
* Providing a set of annotations (either inline or file) that could be applied to the artifact that combines the annotations in the `Chart.yaml` and the provided values. There would need to be appropriate validation logic to ensure it complies with the appropriate timestamp format

stevehipwell:
@sabre1041 my preference for _Helm_ v3 would be to support either the proposed `--created-date` or _ORAS_ mirroring `--annotation` (I'm not sure why this is annotation and not `--label`?) flag. Logically this is aligned well with the current implementation which is either `now` or the created date of a file.

For _Helm_ v4 I think there are some really good use cases for driving the annotations from another location to _Chart.yaml_; personally I'd like to see the chart source be (more) static with common changes such as image tag/digest, CHANGELOG, etc being merged to create the package. But I guess that's another discussion; I missed the Helm 4 highway KubeCon session and can't find the recording so I don't know if this was covered.

sabre1041:
[...]

PR #13376, opened 2024-10-05
fix: Use chart archive modifed time for OCI push

gjenkins8:
Helps with: https://github.com/helm/helm/issues/12973

- Strictly, this is changing the behavior of OCI push. But from https://github.com/helm/helm/pull/13272#issuecomment-2381977860, the current behavior incorrect/unexpectedly implemented anyway
- I changed the implementation of the `ctime.Created` function. However, this function never returned a "created" time. On Linux, it returned the "inode last updated time". And on non-linux, it returns the files "ModTime". (IMHO, we should just remove `ctime.Created`. But want/need to retain public API compatibility.

**If applicable**:


kalvarahul (approved):

robertsirc (approved):
LGTM

sabre1041 (approved):
LGTM

PR #13382, opened 2024-10-08
chore(oci): migrate to ORAS Golang library v2

### LINKED ISSUE #11821: Switch to `oras.land/oras-go/v2`
Hey folks,

I'm wondering if there are any plans to switch to the new `oras.land/oras-go/v2` dependency rather than using v1.2.2. It looks like that the library has been completely redesigned, would it still match helm's requirements?

TerryHowe:

I could not create a PR on Zoran's https://github.com/helm/helm/pull/12310 so this combines the two with:
* rebased
* fixed broken tests
* diffs from Zoran's https://github.com/helm/helm/pull/13382/commits/e88e515453acf28916327a6bc409b29311c70e24

Updates from oras.land/oras-go to oras.land/oras-go/v2. The main user facing change is that the --plain-http parameter is now required even when accessing OCI registries running on localhost.

Fixes https://github.com/helm/helm/issues/11821
Closes https://github.com/helm/helm/pull/12310

I'm new to this codebase and some tests needed to be adjusted to match the new dependency, please look over the test changes to see if there are any issues with those.

**If applicable**:


TerryHowe:
Tests failed in CI, but pass on the mac. I assume the problem is with the credential backend.

TerryHowe:
> Two of the commits are `Unverified`.

Hmm, since I cant do a PR on Zoran's repo, I'll probably have to merge  main instead of rebase.

TerryHowe:
> > Two of the commits are `Unverified`.
> 
> Hmm, since I cant do a PR on Zoran's repo, I'll probably have to merge main instead of rebase.

Fixed

Happylinzy:
Hi. Any update in merging this pr?

Happylinzy (approved):
LGTM

pkg/registry/client.go
```diff
@@ -195,12 +191,9 @@ func ClientOptPlainHTTP() ClientOption {
...
 
-// ClientOptResolver returns a function that sets the resolver setting on a client options set
-func ClientOptResolver(resolver remotes.Resolver) ClientOption {
-	return func(client *Client) {
-		client.resolver = func(_ registry.Reference) (remotes.Resolver, error) {
-			return resolver, nil
-		}
+func ClientOptResolver(_ remotes.Resolver) ClientOption {
+	return func(c *Client) {
+		c.err = errDeprecatedRemote
```
gjenkins8, line 196: I honestly think, converting the previous build time breakage to a runtime error (in the name of maintaining API compatibility) is a much worse idea.

That said, aside from deferring this PR to Helm v4, I'm not sure how to solve this others. The choices are API or functional breakage.

gjenkins8 (commented):
I have managed to get to a partial review. I would like to finish full review next few days.

The sticking point for this PR is still going to be the breaking change. ie. this comment: https://github.com/helm/helm/pull/13382#discussion_r1836977600

We might just be better targeting Helm v4 and moving forward?

pkg/registry/client_insecure_tls_test.go
```diff
@@ -66,7 +66,10 @@ func (suite *InsecureTLSRegistryClientTestSuite) Test_3_Tags() {
 
 func (suite *InsecureTLSRegistryClientTestSuite) Test_4_Logout() {
 	err := suite.RegistryClient.Logout("this-host-aint-real:5000")
-	suite.NotNil(err, "error logging out of registry that has no entry")
+	if err != nil {
+		// credential backend for mac generates an error
+		suite.NotNil(err, "failed to delete the credential for this-host-aint-real:5000")
+	}
```
TerryHowe, line 72:
```suggestion
	suite.NotNil(err, "failed to delete the credential for this-host-aint-real:5000")
```

pkg/registry/client_insecure_tls_test.go
```diff
@@ -66,7 +66,10 @@ func (suite *InsecureTLSRegistryClientTestSuite) Test_3_Tags() {
 
 func (suite *InsecureTLSRegistryClientTestSuite) Test_4_Logout() {
 	err := suite.RegistryClient.Logout("this-host-aint-real:5000")
-	suite.NotNil(err, "error logging out of registry that has no entry")
+	if err != nil {
```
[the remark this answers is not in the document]
TerryHowe: Alright, I was confused too, but the comment is actually correct.

On the mac the docker credential backend produces and error for credential not found, but it is silently ignored by other credential providers.

pkg/registry/client_tls_test.go
```diff
@@ -66,7 +66,10 @@ func (suite *TLSRegistryClientTestSuite) Test_3_Tags() {
 
 func (suite *TLSRegistryClientTestSuite) Test_4_Logout() {
 	err := suite.RegistryClient.Logout("this-host-aint-real:5000")
-	suite.NotNil(err, "error logging out of registry that has no entry")
+	if err != nil {
```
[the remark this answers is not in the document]
TerryHowe: On the mac the docker credential backend produces and error for credential not found, but it is silently ignored by other credential providers. I don't think v1 oras-go supported using the native key store, so this change needs to happen. I'm not sure of a more graceful way to handle this.

pkg/registry/client_test.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/containerd/containerd/remotes"
+)
+
+func TestNewClientResolverNotSupported(t *testing.T) {
+	var r remotes.Resolver
+
+	client, err := NewClient(ClientOptResolver(r))
+	if client != nil || err != errDeprecatedRemote {
+		t.Errorf("expected nil, %v, got: %v, %v", errDeprecatedRemote, client, err)
+	}
```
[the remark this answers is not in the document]
TerryHowe: Done. I prefer testify as well.

pkg/action/registry_login.go
```diff
@@ -84,5 +92,6 @@ func (a *RegistryLogin) Run(_ io.Writer, hostname string, username string, passw
 		hostname,
 		registry.LoginOptBasicAuth(username, password),
 		registry.LoginOptInsecure(a.insecure),
-		registry.LoginOptTLSClientConfig(a.certFile, a.keyFile, a.caFile))
+		registry.LoginOptTLSClientConfig(a.certFile, a.keyFile, a.caFile),
```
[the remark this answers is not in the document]
TerryHowe: I don't see anything in the underlying code that would validate those options. It looks like to me that if you specified plainHTTP, it is going to ignore any TLS options which makes sense.

I don't think there is any mutually exclusive check on insecure/TLS as well.

If PlainHTTP is an issue, I'd almost sooner remove the flag from this PR and throw it in another PR. It is a new feature in this PR.

pkg/registry/client.go
TerryHowe: More discussion on this from Zoran's PR https://github.com/helm/helm/pull/12310/files#r1499833508 I don't love it either, but I'm not seeing a better solution

TerryHowe: Bad for a few, but great for many.

gjenkins8: For Helm 4, we should just remove `ClientOptResolver` I think

For a Helm 3 backport, what would be the alternative for users? Can we document in a function doc comment?

gjenkins8 (dismissed):

gjenkins8 (dismissed):

TerryHowe:
Rebased the changes from the digest PR.

gjenkins8 (dismissed):

pkg/registry/client.go
```diff
@@ -68,11 +75,13 @@ type (
 		username           string
 		password           string
 		out                io.Writer
-		authorizer         auth.Client
+		authorizer         *auth.Client
 		registryAuthorizer RemoteClient
+		credentialsStore   credentials.Store
 		resolver           func(ref registry.Reference) (remotes.Resolver, error)
```
twz123, line 81: If I'm not mistaken, this never gets assigned and will probably cause a nil pointer dereference in `Client.Resolve` on line 837:

https://github.com/helm/helm/blob/6505d416f618be17a481beb19b344369be141daa/pkg/registry/client.go#L827-L844

scottrigby: @twz123  It appears to be set in `NewClient()` on lines 147 or 150, depending on dockerStore status.

scottrigby (commented):
@TerryHowe this is all looking good IMO. Just a few things:

1. needs a rebase again
2. can you please verify my reply to this question https://github.com/helm/helm/pull/13382#discussion_r1936313286
3. Since the starting point from this was #12310—and you're a good community player by noting that and linking to a diff— would you mind also please adding @zregvart's [co-author commit trailer](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors) to your commit message? Like this:
```
Co-authored-by: Zoran Regvart <zoran@regvart.com>
```

pkg/registry/client.go
twz123: @scottrigby I wasn't talking about `credentialsStore` in line 80, but about `resolver` in line 81. This is not directly obvious from how GitHub displays this comment. Sorry, I should have been clearer.

Still pretty sure that `resolver` never gets assigned and will produce a nil pointer dereference in the aforementioned code block.

scottrigby: @twz123 that makes more sense. Not needed for Helm 4, but to continue attempting a reasonable backwards compatibility for Helm 3 (re this thread https://github.com/helm/helm/pull/13382#discussion_r1836977600), we'd need to keep `resolver` in the struct. For that, what would you recommend?

TerryHowe:
@scottrigby 

1. rebase done
2. I'll respond to question https://github.com/helm/helm/pull/13382#discussion_r1936313286
3. Thanks for this suggestion, Zoran added as co-author

pkg/registry/client.go
TerryHowe: I removed `resolver` in the latest changes. It is no longer needed.

TerryHowe: `resolver` was private, so I don't think we need to keep it in the struct, but the latest changes keeps ClientOptResolver with a deprecated message.

TerryHowe:
One more thing with this change, `--plain-http` is *now required* to communicate with a registry that only supports http. From a security perspective, this is the way it should work.

If/When this gets merged to Helm 3, we should probably fallback to http with a warning message.

gjenkins8 (approved):

scottrigby (approved):
lgtm!

scottrigby:
kudos to your perseverance @TerryHowe 

let's see if we can find a way to get a backport PR backwards-compatible for Helm 3. This PR is very close already.

PR #13383, opened 2024-10-08
chore: Make retryingRoundTripper type public

### LINKED ISSUE #13052: Make retryingRoundTripper type public
Hello team,

My team encountered an issue in Fluxcd (https://github.com/fluxcd/flux2/issues/4804) that seems to be related to an issue that was already solved here (https://github.com/helm/helm/pull/11426). When trying to replicate the solution there, I saw on the code that the `retryingRoundTripper` is private to the helm cli package (https://github.com/helm/helm/blob/main/pkg/cli/roundtripper.go#L27). 

Can we make it public? Do you guys see any issues on doing this?

luisdavim:
Make `retryingRoundTripper` type public to allow it being used in other projects

closes #13052

Some projects use helm as a library but need to create the client outside, making this object public allows using the same functionality without having to maintain a copy.

none.

**If applicable**:

### DIFF
--- a/pkg/cli/environment.go
+++ b/pkg/cli/environment.go
@@ -36,6 +36,7 @@ import (
 
 	"helm.sh/helm/v3/internal/version"
 	"helm.sh/helm/v3/pkg/helmpath"
+	"helm.sh/helm/v3/pkg/kube"
 )
 
 // defaultMaxHistory sets the maximum number of releases to 0: unlimited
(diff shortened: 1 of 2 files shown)


luisdavim:
Opned this as a followup of:
* https://github.com/fluxcd/flux2/issues/4804
* https://github.com/fluxcd/helm-controller/pull/1084/files

luisdavim:
I can also move the RetryingRoundTripper to its own package if that's preferable when making it public.

pkg/cli/roundtripper.go
```diff
@@ -24,15 +24,15 @@ import (
 	"strings"
 )
 
-type retryingRoundTripper struct {
+type RetryingRoundTripper struct {
```
hiddeco, line 27: The better place for this would probably be `pkg/kube`, as this contains other Kubernetes client related things.

luisdavim: Sure, I can move it. My initial approach was to make the bare minimum changes....

luisdavim: Updated, thanks for having a look.

robertsirc (approved):
I pulled this down, ran my test against it and this didn't break anything.

mattfarina (approved):

PR #13402, opened 2024-10-22
adding more unit test

robertsirc:
Adding more unit test, covering mostly the negative. 

**If applicable**:

### DIFF
--- a/pkg/kube/ready_test.go
+++ b/pkg/kube/ready_test.go
@@ -36,8 +36,7 @@ import (
 
 const defaultNamespace = metav1.NamespaceDefault
 
-// TODO: correct test cases to cover the positive and not just the negative.
-func Test_ReadyChecker_IsReady(t *testing.T) {
+func Test_ReadyChecker_IsReady_Pod(t *testing.T) {
 	type fields struct {
 		client        kubernetes.Interface
 		log           func(string, ...interface{})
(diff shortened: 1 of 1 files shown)


pkg/kube/ready_test.go
```diff
@@ -20,16 +20,212 @@ import (
...
+			},
+			want:    false,
+			wantErr: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			c := &ReadyChecker{
+				client:        tt.fields.client,
+				log:           tt.fields.log,
```
mattfarina, line 213: If `log` is always the same, you don't need a variable for it. You can use set it here.

pkg/kube/ready_test.go
```diff
@@ -521,6 +717,144 @@ func Test_ReadyChecker_serviceReady(t *testing.T) {
 	}
 }
 
+func Test_ReadyChecker_crdBetaReady(t *testing.T) {
```
mattfarina, line 720: I see there are different test functions for different types against the same `ReadyChecker`. The different functions are testing different internal functions. Should these all be run through `IsReady` (the public function that calls these internal ones)? Thoughts?

robertsirc: You are correct, I don't like how this is written currently but I wanted to cover with test first then start refactoring. As I don't want to break anything. IsReady is huge and needs to be paired down.

pkg/kube/ready_test.go
robertsirc: This is a stub out I am sure I will refactor this out later.

mattfarina (approved):

PR #13411, opened 2024-10-25
Use three way merge for unstructured objects

### LINKED ISSUE #9937: Custom resources should update with three way merge patch

p-se:
Based on work of wxdao <waxiadao@gmail.com> in the same issue, supersedes #9938. Basically just adds tests.

closes #9937

**If applicable**:

### DIFF
--- a/cmd/helm/install.go
+++ b/cmd/helm/install.go
@@ -207,6 +207,7 @@ func addInstallFlags(cmd *cobra.Command, f *pflag.FlagSet, client *action.Instal
 	f.BoolVar(&client.EnableDNS, "enable-dns", false, "enable DNS lookups when rendering templates")
 	f.BoolVar(&client.HideNotes, "hide-notes", false, "if set, do not show notes in install output. Does not affect presence in chart metadata")
 	f.BoolVar(&client.TakeOwnership, "take-ownership", false, "if set, install will ignore the check for helm annotations and take ownership of the existing resources")
+	f.BoolVar(&client.ThreeWayMergeForCustomResources, "three-way-merge-for-custom-resources", false, "also considers the state of custom resources and custom resource definitions in the cluster when adopting resources")
 	addValueOptionsFlags(f, valueOpts)
 	addChartPathOptionsFlags(f, &client.ChartPathOptions)
 
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -112,7 +112,9 @@ type Install struct {
 	UseReleaseName bool
 	// TakeOwnership will ignore the check for helm annotations and take ownership of the resources.
 	TakeOwnership bool
-	PostRenderer  postrender.PostRenderer
+	// Also considers the state of custom resources and custom resource definitions in the cluster when adopting resources.
+	ThreeWayMergeForCustomResources bool
+	PostRenderer                    postrender.PostRenderer
 	// Lock to control raceconditions when the process receives a SIGTERM
 	Lock sync.Mutex
 }
@@ -455,7 +457,7 @@ func (i *Install) performInstall(rel *release.Release, toBeAdopted kube.Resource
 	if len(toBeAdopted) == 0 && len(resources) > 0 {
 		_, err = i.cfg.KubeClient.Create(resources)
 	} else if len(resources) > 0 {
-		_, err = i.cfg.KubeClient.Update(toBeAdopted, resources, i.Force)
+		_, err = i.cfg.KubeClient.Update(toBeAdopted, resources, i.Force, i.ThreeWayMergeForCustomResources)
 	}
 	if err != nil {
 		return rel, err
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -113,6 +113,7 @@ type Install struct {
 	// TakeOwnership will ignore the check for helm annotations and take ownership of the resources.
 	TakeOwnership bool
 	// Also considers the state of custom resources and custom resource definitions in the cluster when adopting resources.
+	// TODO Helm 4: Remove this config and always merge custom resources
 	ThreeWayMergeForCustomResources bool
 	PostRenderer                    postrender.PostRenderer
 	// Lock to control raceconditions when the process receives a SIGTERM
--- a/pkg/action/rollback.go
+++ b/pkg/action/rollback.go
@@ -45,6 +45,8 @@ type Rollback struct {
 	Force         bool // will (if true) force resource upgrade through uninstall/recreate if needed
 	CleanupOnFail bool
 	MaxHistory    int // MaxHistory limits the maximum number of revisions saved per release
+	// Also considers the state of custom resources and custom resource definitions in the cluster when upgrading (or adopting) resources.
+	ThreeWayMergeForCustomResources bool
 }
 
 // NewRollback creates a new Rollback object with the given configuration.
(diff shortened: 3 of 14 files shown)


robertsirc:
Hello @p-se your PR has the DCO missing please fix this as we can then take a look at it.

mattfarina (changes requested):
Thanks for working on this. I've wanted to see three way merge expanded here.

To share, when Helm v3 was released, it wasn't an option to have 3-way merge on unstructured resources. It was added after v3. That's why Helm doesn't do it.

There are some changes I would like to see to fit in with the rest of Helm.

pkg/cli/environment.go
```diff
@@ -89,6 +90,10 @@ type EnvSettings struct {
 	BurstLimit int
 	// QPS is queries per second which may be used to avoid throttling.
 	QPS float32
+	// ThreeWayMergeForUnstructured controls whether to use three way merge
+	// patch for unstructured objects (custom resource, custom definitions,
+	// etc).
+	ThreeWayMergeForUnstructured bool
```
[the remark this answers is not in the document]
p-se: About the actions, it currently is implemented as a global option. The alternative to that is to add it to the upgrade and install action commands individually. The environment is just an additional option. But I have no issue removing it from there.

Did I understand correctly that there should not be an environment variable to configure this flag and that that the option should not be global but implemented for the install and upgrade actions individually instead?

pkg/cli/environment.go
```diff
@@ -111,6 +116,7 @@ func New() *EnvSettings {
 		QPS:                       envFloat32Or("HELM_QPS", defaultQPS),
 	}
 	env.Debug, _ = strconv.ParseBool(os.Getenv("HELM_DEBUG"))
+	env.ThreeWayMergeForUnstructured, _ = strconv.ParseBool(os.Getenv("HELM_THREE_WAY_MERGE_FOR_UNSTRUCTURED"))
```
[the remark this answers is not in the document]
p-se: This will affect custom resources and custom resource definitions. What do you think about `ThreeWayMergeForCustomResources`, or if shorter is preferred but still properly understood, `ThreeWayMergeForCRs`?

pkg/cli/environment.go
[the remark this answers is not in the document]
p-se: Well, technically adding the option to the install action should have no effect unless `TakeOwnership` is also part of the CLI and not only of the SDK. But since the SDK needs it anyway for all resources of a chart that are already present and are supposed to be adopted by having the  `TakeOwnership` option set, it might make sense to add it, too. Or even go as far as to enable `TakeOwnership` for install and upgrade commands in the CLI, too?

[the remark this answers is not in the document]
p-se: Adoption of resources owned by helm is a feature of helm that is independent of `TakeOwnership`, so it will make sense to have it in the CLI in any case.

cmd/helm/install.go
```diff
@@ -201,6 +201,7 @@ func addInstallFlags(cmd *cobra.Command, f *pflag.FlagSet, client *action.Instal
 	f.StringToStringVarP(&client.Labels, "labels", "l", nil, "Labels that would be added to release metadata. Should be divided by comma.")
 	f.BoolVar(&client.EnableDNS, "enable-dns", false, "enable DNS lookups when rendering templates")
 	f.BoolVar(&client.HideNotes, "hide-notes", false, "if set, do not show notes in install output. Does not affect presence in chart metadata")
+	f.BoolVar(&client.ThreeWayMergeForCustomResources, "three-way-merge-for-custom-resources", false, "also considers the state of custom resources and custom resource definitions in the cluster when adopting resources")
```
gjenkins8, line 204: Isn't this redundant for install? Even when adopting resources, there isn't going to be a previous version (only the new and real versions). So no three-way merge will be possible (only two-way)

Unless I'm mistaken?

pkg/action/install.go
```diff
@@ -112,7 +113,9 @@ type Install struct {
 	UseReleaseName bool
 	// TakeOwnership will ignore the check for helm annotations and take ownership of the resources.
 	TakeOwnership bool
-	PostRenderer  postrender.PostRenderer
+	// Also considers the state of custom resources and custom resource definitions in the cluster when adopting resources.
+	ThreeWayMergeForCustomResources bool
```
gjenkins8, line 117: Can we add a `TODO Helm 4: Remove ..` here please
```suggestion
	// Also considers the state of custom resources and custom resource definitions in the cluster when adopting resources.
	// TODO Helm 4: Remove this config and always merge custom resources
```

For Helm 4, we should always look to three-way merge CRs IMHO.

(and same below for upgrade/rollback below)

pkg/action/rollback.go
```diff
@@ -45,6 +45,8 @@ type Rollback struct {
 	Force         bool // will (if true) force resource upgrade through uninstall/recreate if needed
 	CleanupOnFail bool
 	MaxHistory    int // MaxHistory limits the maximum number of revisions saved per release
+	// Also considers the state of custom resources and custom resource definitions in the cluster when upgrading (or adopting) resources.
```
gjenkins8, line 48: nit:
```suggestion
	// Also considers the state of custom resources and custom resource definitions in the cluster when downgrading resources.
```

gjenkins8 (commented):
Some comments/questions please

cmd/helm/install.go
p-se: No, it is not redundant.

Adoption of resources end up being performed in Client.Update (helm/pkg/kube/client.go), even if the install command was used. For that to work, it effectively sets the original and target versions of a resource to the same content, regardless of the merge strategy used. Later, when the patch is to be created, it checks for the types to differentiate between native and unstructured resources and fetches the current version of the resources from the cluster.

If a two-way merge is used for patch creation, only the original and target versions are compared, never the current version of the cluster. This leads the patch creation code to return an empty patch, so that the unstructured resource is not changed.

If a three-way-merge is used, the current version is also considered and that enables updates of unstructured resources. 

Additionally but potentially less relevant, if `TakeOwnership` (only in SDK, not a CLI flag) is used, which is supposed to work regardless of the ownership of resources, the result looks more like a failure (although Helm doesn't currently see it as such), since the resources also aren't touched, but that means they end up in a state not owned by helm. Having that three-way-merge flag in the CLI for adoption with TakeOwnership is not necessary, as TakeOwnership cannot currently be enabled from the CLI, but it could prevent developers missing to add it if TakeOwnership is made available to the CLI eventually.

gjenkins8: fwiw, the CLI should gain `--take-ownership` support: https://github.com/helm/helm/pull/13439

pkg/action/rollback.go
p-se: Done.

p-se:
@gjenkins8 I think I have addressed all the comments. Is there anything else I can do?

frittentheke:
@gjenkins8 @mattfarina any chance this PR could move forward? Better support for the lifecycle of CR is really highly anticipated.

please also see https://github.com/helm/helm/pull/11650#issuecomment-2815986578

Nuckal777:
Agree with @frittentheke. Is there a chance to move this PR forward?

frittentheke:
@gjenkins8 @mattfarina with Helm 4.x released I suppose this issue is not in focus anymore?

frittentheke:
@gjenkins8 @mattfarina ?

PR #7096, opened 2019-11-27
Allow tests to run on s390x

vivkong:
In addition to https://github.com/helm/helm/pull/6988, this will allow tests to run on s390x.

PR #13412, opened 2024-10-28
Allow tests to run on loong64

xtlsoft:

When building Helm on LoongArch-based systems, unit tests fail on `TestPlatformPrepareCommand` because it does not account for `loong64`. By applying this patch, the unit tests should run properly on the aforementioned architecture.

Several similar PRs have been merged, such as one for s390x (https://github.com/helm/helm/pull/7096) and another for RISC-V.


robertsirc (approved):
LGTM. I am not as familiar to the loong64 architecture. As this is not something for me here. The change is minimal, it adds to test (which I personally love) and it shouldn't impact at all. Thank you for the PR.

PR #13416, opened 2024-11-01
Increasing the size of the runner used for releases.

mattfarina:
Note, this runner is provided as part of the CNCF access to GitHub Enterprise runners.

Builds, used for releases, can take considerable time. This moves releases to use a larger runner which should reduce build times.

**If applicable**:


robertsirc (approved):
LGTM this is to increase the size of the runners.

PR #13423, opened 2024-11-06
Updating subchart load error to be more descriptive

tjasko:
In the event some malformed folders/files make its way into the `charts/` directory meant for subcharts, it is possible one may see an error message like the below:
```
Error: error unpacking foo in app: Chart.yaml file is missing
```

This error is not very descriptive, as it doesn't reference where on disk `Chart.yaml` is missing. In the event this happens, this error is being updated to include that `Chart.yaml` is missing for the `foo` subchart.

Without this updated error messaging, one would need to look into Helm's codebase to troubleshoot what exactly is wrong.

N/A

**If applicable**:
  - No applicable user facing changes.
  - Existing unit tests were updated to assert for the new error messaging.
  - Error messaging will be changed, otherwise backwards compatible.


TerryHowe (approved):
/lgtm

robertsirc (approved):
LGTM, this change is small and is only changing the output of the error to be more descriptive to the user. I pulled this down all code builds, all test pass and coverage stays the same.

PR #13439, opened 2024-11-12
Add --take-ownership flag for install and upgrade commands

### LINKED ISSUE #2730: Proposal: helm commandeer

### LINKED ISSUE #10304: helm overwrite existing resources deployed by non-helm
Hi:
I have an EKS cluster created with Terraform and working to implement AWS Console RBAC in the cluster discussed [here](https://docs.aws.amazon.com/eks/latest/userguide/add-user-role.html).  As part of this process, I am able to deploy the Role, Binding using Helm Chart and patch aws-auth configmap using kubectl.  

Deploying configmap using Helm complains the resource already exists and requires the configmap metadata fields (ownership,etc) to be set to helm values.  Ideally would like helm to overwrite the resource, so we can start developing custom charts and use helm for cluster updates/deployments.  

Can helm overwrite existing resources deployed by non-helm, or is there a workaround that should make this possible ?

Error:
Error: rendered manifests contain a resource that already exists. Unable to continue with install: ConfigMap "aws-auth" in namespace "kube-system" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: key "app.kubernetes.io/managed-by" must equal "Helm": current value is "Terraform"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "console-rbac"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "kube-system"

Output of `helm version`:
v3.5.3+g041ce5a

Output of `kubectl version`: 
Client Version: v1.19.0
Server Version: v1.20.7-eks-d88609

AWS EKS

mayankshah1607:
closes #10304, closes #2730

`TakeOwnership` option was added to the Install and Upgrade actions in https://github.com/helm/helm/pull/12876. This PR allows setting this option on the CLI commands using a new `--take-ownership` flag

xref: https://github.com/helm/helm/issues/10304#issuecomment-2468532665

**If applicable**:

### DIFF
--- a/cmd/helm/install.go
+++ b/cmd/helm/install.go
@@ -201,6 +201,7 @@ func addInstallFlags(cmd *cobra.Command, f *pflag.FlagSet, client *action.Instal
 	f.StringToStringVarP(&client.Labels, "labels", "l", nil, "Labels that would be added to release metadata. Should be divided by comma.")
 	f.BoolVar(&client.EnableDNS, "enable-dns", false, "enable DNS lookups when rendering templates")
 	f.BoolVar(&client.HideNotes, "hide-notes", false, "if set, do not show notes in install output. Does not affect presence in chart metadata")
+	f.BoolVar(&client.TakeOwnership, "take-ownership", false, "if set, install will ignore the check for helm annotations and take ownership of the existing resources.")
 	addValueOptionsFlags(f, valueOpts)
 	addChartPathOptionsFlags(f, &client.ChartPathOptions)
 
--- a/cmd/helm/install.go
+++ b/cmd/helm/install.go
@@ -201,7 +201,7 @@ func addInstallFlags(cmd *cobra.Command, f *pflag.FlagSet, client *action.Instal
 	f.StringToStringVarP(&client.Labels, "labels", "l", nil, "Labels that would be added to release metadata. Should be divided by comma.")
 	f.BoolVar(&client.EnableDNS, "enable-dns", false, "enable DNS lookups when rendering templates")
 	f.BoolVar(&client.HideNotes, "hide-notes", false, "if set, do not show notes in install output. Does not affect presence in chart metadata")
-	f.BoolVar(&client.TakeOwnership, "take-ownership", false, "if set, install will ignore the check for helm annotations and take ownership of the existing resources.")
+	f.BoolVar(&client.TakeOwnership, "take-ownership", false, "if set, install will ignore the check for helm annotations and take ownership of the existing resources")
 	addValueOptionsFlags(f, valueOpts)
 	addChartPathOptionsFlags(f, &client.ChartPathOptions)
 
--- a/cmd/helm/upgrade.go
+++ b/cmd/helm/upgrade.go
@@ -151,6 +151,7 @@ func newUpgradeCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 					instClient.Labels = client.Labels
 					instClient.EnableDNS = client.EnableDNS
 					instClient.HideSecret = client.HideSecret
+					instClient.TakeOwnership = client.TakeOwnership
 
 					if isReleaseUninstalled(versions) {
 						instClient.Replace = true
--- a/cmd/helm/upgrade.go
+++ b/cmd/helm/upgrade.go
@@ -280,6 +280,7 @@ func newUpgradeCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 	f.StringVar(&client.Description, "description", "", "add a custom description")
 	f.BoolVar(&client.DependencyUpdate, "dependency-update", false, "update dependencies if they are missing before installing the chart")
 	f.BoolVar(&client.EnableDNS, "enable-dns", false, "enable DNS lookups when rendering templates")
+	f.BoolVar(&client.TakeOwnership, "take-ownership", false, "if set, upgrade will ignore the check for helm annotations and take ownership of the existing resources.")
 	addChartPathOptionsFlags(f, &client.ChartPathOptions)
 	addValueOptionsFlags(f, valueOpts)
 	bindOutputFlag(cmd, &outfmt)
--- a/cmd/helm/upgrade.go
+++ b/cmd/helm/upgrade.go
@@ -280,7 +280,7 @@ func newUpgradeCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 	f.StringVar(&client.Description, "description", "", "add a custom description")
 	f.BoolVar(&client.DependencyUpdate, "dependency-update", false, "update dependencies if they are missing before installing the chart")
 	f.BoolVar(&client.EnableDNS, "enable-dns", false, "enable DNS lookups when rendering templates")
-	f.BoolVar(&client.TakeOwnership, "take-ownership", false, "if set, upgrade will ignore the check for helm annotations and take ownership of the existing resources.")
+	f.BoolVar(&client.TakeOwnership, "take-ownership", false, "if set, upgrade will ignore the check for helm annotations and take ownership of the existing resources")
 	addChartPathOptionsFlags(f, &client.ChartPathOptions)
 	addValueOptionsFlags(f, valueOpts)
 	bindOutputFlag(cmd, &outfmt)
--- a/cmd/helm/install_test.go
+++ b/cmd/helm/install_test.go
@@ -102,6 +102,12 @@ func TestInstall(t *testing.T) {
 			cmd:    "install aeneas testdata/testcharts/empty --replace",
 			golden: "output/install-and-replace.txt",
 		},
+		// Install, take ownership
+		{
+			name:   "install and replace release",
+			cmd:    "install aeneas-take-ownership testdata/testcharts/empty --take-ownership",
+			golden: "output/install-and-take-ownership.txt",
+		},
 		// Install, with timeout
 		{
 			name:   "install with a timeout",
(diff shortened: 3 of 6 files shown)


banjoh (approved):

unclok (approved):

cmd/helm/install.go
```diff
@@ -201,6 +201,7 @@ func addInstallFlags(cmd *cobra.Command, f *pflag.FlagSet, client *action.Instal
 	f.StringToStringVarP(&client.Labels, "labels", "l", nil, "Labels that would be added to release metadata. Should be divided by comma.")
 	f.BoolVar(&client.EnableDNS, "enable-dns", false, "enable DNS lookups when rendering templates")
 	f.BoolVar(&client.HideNotes, "hide-notes", false, "if set, do not show notes in install output. Does not affect presence in chart metadata")
+	f.BoolVar(&client.TakeOwnership, "take-ownership", false, "if set, install will ignore the check for helm annotations and take ownership of the existing resources.")
```
gjenkins8, line 204:
```suggestion
	f.BoolVar(&client.TakeOwnership, "take-ownership", false, "if set, install will ignore the check for helm annotations and take ownership of the existing resources")
```

minor consistency thing pls

cmd/helm/upgrade.go
```diff
@@ -280,6 +280,7 @@ func newUpgradeCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 	f.StringVar(&client.Description, "description", "", "add a custom description")
 	f.BoolVar(&client.DependencyUpdate, "dependency-update", false, "update dependencies if they are missing before installing the chart")
 	f.BoolVar(&client.EnableDNS, "enable-dns", false, "enable DNS lookups when rendering templates")
+	f.BoolVar(&client.TakeOwnership, "take-ownership", false, "if set, upgrade will ignore the check for helm annotations and take ownership of the existing resources.")
```
gjenkins8, line 283:
```suggestion
	f.BoolVar(&client.TakeOwnership, "take-ownership", false, "if set, upgrade will ignore the check for helm annotations and take ownership of the existing resources")
```

gjenkins8 (commented):
very minor fix up of the help text please

cmd/helm/install.go
mayankshah1607: updated, thank you

cmd/helm/upgrade.go
mayankshah1607: updated, thank you

gjenkins8 (dismissed):

banjoh:
@scottrigby / @joejulian are you able to take a look at this one?

banjoh:
@mayankshah1607 since this PR needs to merged into v3, can you please create a new PR to [dev-v3 branch](https://github.com/helm/helm/tree/dev-v3)?

scottrigby (commented):
This PR looks good, however with manual testing this doesn't seem to be working for upgrade, only install. I'm not sure why this would be the case, since it should work according to #12876

Can someone else reproduce this?

```sh
# create example chart
helm create foo

# create matching plain template files
helm template foo foo --output-dir foo-template
<remove all helm labels>
<remove tests dir>

# install plan manifests without helm labels or metadata
kubectl apply -f foo-template/foo/templates/

# it makes sense that this wouldn't work
$ helm upgrade foo foo/ --take-ownership
Error: UPGRADE FAILED: "foo" has no deployed releases

# from this PR it seems this should work, but doesn't
$ helm upgrade --install foo foo/ --take-ownership
Error: Unable to continue with install: ServiceAccount "foo" in namespace "default" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: missing key "app.kubernetes.io/managed-by": must be set to "Helm"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "foo"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "default"

# this works
$ helm install foo foo/ --take-ownership
<expected output>
$ helm ls  
NAME	NAMESPACE	REVISION	UPDATED                             	STATUS  	CHART    	APP VERSION
foo 	default  	1       	2024-12-11 00:21:59.727628 -0500 EST	deployed	foo-0.1.0	1.16.0  
```

PS I'm also updating the description to point to the correct PR (#12876 not #13130), and adding closing keywords to related issues (#10304 and #2730).

scottrigby:
For `upgrade --install` to work, I believe we also need to add this:
```diff
diff --git a/cmd/helm/upgrade.go b/cmd/helm/upgrade.go
index 1fe793a92..75788459b 100644
--- a/cmd/helm/upgrade.go
+++ b/cmd/helm/upgrade.go
@@ -151,6 +151,7 @@ func newUpgradeCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
                                        instClient.Labels = client.Labels
                                        instClient.EnableDNS = client.EnableDNS
                                        instClient.HideSecret = client.HideSecret
+                                       instClient.TakeOwnership = client.TakeOwnership
 
                                        if isReleaseUninstalled(versions) {
                                                instClient.Replace = true
```

scottrigby (changes requested):
For `upgrade --install` to work, we also need to add this:
```diff
diff --git a/cmd/helm/upgrade.go b/cmd/helm/upgrade.go
index 1fe793a92..75788459b 100644
--- a/cmd/helm/upgrade.go
+++ b/cmd/helm/upgrade.go
@@ -151,6 +151,7 @@ func newUpgradeCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
                                        instClient.Labels = client.Labels
                                        instClient.EnableDNS = client.EnableDNS
                                        instClient.HideSecret = client.HideSecret
+                                       instClient.TakeOwnership = client.TakeOwnership
 
                                        if isReleaseUninstalled(versions) {
                                                instClient.Replace = true
```

mayankshah1607:
Thanks @scottrigby , I've updated the PR

mayankshah1607:
> @mayankshah1607 since this PR needs to merged into v3, can you please create a new PR to [dev-v3 branch](https://github.com/helm/helm/tree/dev-v3) with the same changes here?

@banjoh can I just change the base of this PR? Or does it need to be created against both `main` and `dev-v3`?

scottrigby (dismissed):
lgtm! great work @mayankshah1607

We also need better test coverage for this feature, as evidenced by my finding a bug only through manual testing. I've opened a separate issue for that which someone can pick up in a separate PR: https://github.com/helm/helm/issues/13519

gjenkins8 (dismissed):

scottrigby:
A failing test now needs to be fixed

banjoh:
@mayankshah1607 have you had any luck looking into the failing test?

mayankshah1607:
Running the tests locally, everything seems to pass. I'm not sure what the issue is, but I pushed a fix anyway.. not sure if that'll do.

mayankshah1607:
Ok, tests are fixed. PR against `dev-v3` branch is ready: https://github.com/helm/helm/pull/13531

banjoh (approved):

gjenkins8 (approved):

scottrigby (approved):
@mayankshah1607 test fix looks good. Nice work 👏

lucasfcnunes:
@yxxhero take a look at this. Helmfile would greatly benefit from it.

atsu85:
if i understand correctly, then [official documentation](https://helm.sh/docs/helm/helm_upgrade/#options) isn't currently updated with this new flag?

Seems just the feature i was looking for - very good timing regarding the release of this feature :)

PR #13441, opened 2024-11-12
Replace the splitManifests regex

borissmidt:
The regex is slow, we use helm in a gitops repo and template a lot of helm in our CI steps and it is getting quite slow because i've profiled helm itself and found true bottlenecks:

1. Splitting the yaml output (this part)
2. Sorting the yaml output because it has to parse the source
3. Reading the chart index in the chart downloader

**If applicable**:

### DIFF
--- a/pkg/releaseutil/manifest.go
+++ b/pkg/releaseutil/manifest.go
@@ -17,8 +17,6 @@ limitations under the License.
 package releaseutil
 
 import (
-	"fmt"
-	"regexp"
 	"strconv"
 	"strings"
 )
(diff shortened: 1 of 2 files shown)


mattfarina:
> Reading the chart index in the chart downloader

FYI, you can get a massive performance boost by using JSON in the index.yaml. A JSON document is valid YAML. More recent versions of Helm detect JSON and use a JSON parser. JSON parsers are faster and use fewer resources. YAML is more difficult to parse due to things like anchors.

pkg/releaseutil/manifest_test.go
```diff
@@ -49,13 +50,99 @@ spec:
     image: fake-image
     cmd: fake-command`
```
TerryHowe, line 52: I'd think there would be more test cases

pkg/releaseutil/manifest.go
```diff
@@ -45,20 +43,115 @@ func SplitManifests(bigFile string) map[string]string {
...
+	docEndIdx := 0
+	var char rune
+	var width int
+	// using a 'raw' for loop to avoid creating a complicated statemachine
+	i := 0
+	//if strings.HasPrefix(bigFile, "---") {
+	//	i = 3
+	//}
+
+	for i < len(bigFile) {
```
TerryHowe, line 67: If I was writing this, I'd be tempted to look at creating a scanner and/or state machine to make this more readable. I think something could be done that wouldn't hurt performance.

borissmidt:
Yes it is still a draft pr, i will clean it up later but i wanted to know
how the reaction would be.

I'm not sure if a state machine would be simpler but i could try.

There where not many tests to start with but i can add a few more.

pkg/releaseutil/manifest.go
borissmidt: Today i tried two implementations, one with a 'scanner/iterator' but this was clearer to read.

But then looking at the profiler output i saw that 'strings.Prefix' is highly optimized so i tried a 'naive' implementation which was actually twice as fast.

I.e. we now check at very subslice that it start with "\n---".

borissmidt:
I've made an updated implementation which is simpler and more maintainable.

pkg/releaseutil/manifest.go
TerryHowe: It is a lot easier to read now, thanks!

TerryHowe:
You'll probably get more eyes on this if the DCO was fixed https://github.com/helm/helm/pull/13441/checks?check_run_id=33081873991

borissmidt:
@TerryHowe thank you i've added the singing  which fixes the DCO,  I also rebased on master and improved the commti message.  So i hope it is ready now for review.

borissmidt:
Is there anything holding us back here?

PR #13445, opened 2024-11-14
Remove the ShowDescription flag

LalatenduMohanty:
It is one of the `TODO Helm 4:`

**If applicable**:

### DIFF
--- a/cmd/helm/get_all.go
+++ b/cmd/helm/get_all.go
@@ -58,7 +58,7 @@ func newGetAllCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 				}
 				return tpl(template, data, out)
 			}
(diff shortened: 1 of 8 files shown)


robertsirc:
Hello, thank you for your PR, does this close an issue?

LalatenduMohanty:
@robertsirc This fixes the following TODO in the code 

```
// If true, display description to output format,
// only affect print type table.
// TODO Helm 4: Remove this flag and output the description by default.
```

gjenkins8 (approved):
nice! (agree this should be cleaned up / removed for helm 4)

sabre1041:
@LalatenduMohanty According to the TODO, the goal is to remove the flag, but display description by default. The flag is being removed, but the description was also removed

PR #13444, opened 2024-11-14
refactor(status): remove --show-desc and --show-resources flags

justenstall:
**What this PR does / why we need it**:

Resolves two `TODO Helm 4` items.

**Special notes for your reviewer**:

**If applicable**:
- [x] this PR contains user facing changes (the `docs needed` label should be applied if so)
- [ ] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

PR #13456, opened 2024-11-16
fix: replace "github.com/pkg/errors" with stdlib "errors" package

justenstall:

Replaces the archived "github.com/pkg/errors" package with the Go standard library equivalents.

**If applicable**:

### DIFF
--- a/cmd/helm/status.go
+++ b/cmd/helm/status.go
@@ -140,7 +140,7 @@ func (s statusPrinter) WriteTable(out io.Writer) error {
 	}
 	_, _ = fmt.Fprintf(out, "DESCRIPTION: %s\n", s.release.Info.Description)
 
-	if s.release.Info.Resources != nil && len(s.release.Info.Resources) > 0 {
+	if len(s.release.Info.Resources) > 0 {
 		buf := new(bytes.Buffer)
 		printFlags := get.NewHumanPrintFlags()
 		typePrinter, _ := printFlags.ToPrinter("")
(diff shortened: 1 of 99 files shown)


cmd/helm/status.go
```diff
@@ -140,11 +138,9 @@ func (s statusPrinter) WriteTable(out io.Writer) error {
 		_, _ = fmt.Fprintf(out, "VERSION: %s\n", s.release.Chart.Metadata.Version)
 		_, _ = fmt.Fprintf(out, "APP_VERSION: %s\n", s.release.Chart.Metadata.AppVersion)
 	}
-	if s.showDescription {
-		_, _ = fmt.Fprintf(out, "DESCRIPTION: %s\n", s.release.Info.Description)
-	}
+	_, _ = fmt.Fprintf(out, "DESCRIPTION: %s\n", s.release.Info.Description)
```
mattfarina, line 141: This change is breaking some tests and they don't expect the description to be present.

mattfarina (commented):
I have not given this much of a review. I did find one change that's causing numerous tests to fail. The expected output needs to be updated so the tests pass.

cmd/helm/status.go
justenstall: I accidentally started this branch from #13444, I'll cherry-pick the commits to a new branch and re-open the PR

justenstall: Reopened as #13460

PR #13467, opened 2024-11-19
Don't concatenate u.Path and tag if tag is already in OCI URI

### LINKED ISSUE #13466: Error: invalid_reference: invalid tag when chart exposed as OCI url in index.yaml
Running an helm command with `--repo` fails with `invalid tag` error when the selected chart is exposed as OCI url in the repository index.yaml .

```
$ helm template --repo https://charts.bitnami.com/bitnami external-dns --version 8.5.1 | head
---
# Source: external-dns/templates/networkpolicy.yaml
kind: NetworkPolicy
metadata:
  name: release-name-external-dns
  namespace: "default"
  labels:
    app.kubernetes.io/instance: release-name

$ helm template --repo https://charts.bitnami.com/bitnami external-dns --version 8.6.0
Error: invalid_reference: invalid tag
```

These are the relevant [index.yaml](https://charts.bitnami.com/bitnami/index.yaml) bits:

8.6.0

```yaml
  external-dns:
  - name: external-dns
    urls:
    - oci://registry-1.docker.io/bitnamicharts/external-dns:8.6.0
    version: 8.6.0
    ...
```

8.5.1

```yaml
  - name: external-dns
    urls:
    - https://charts.bitnami.com/bitnami/external-dns-8.5.1.tgz
    version: 8.5.1
    ...
```

Output of `helm version`:

```
version.BuildInfo{Version:"v3.16.3", GitCommit:"cfd07493f46efc9debd9cc1b02a0961186df7fdf", GitTreeState:"dirty", GoVersion:"go1.23.3"}
```

termdew:
Closes #13466.

If `helm template` is run for an OCI-based Helm chart, which has the tag already in the OCI URI, the download of the chart fails because of an invalid tag.

### DIFF
--- a/pkg/downloader/chart_downloader.go
+++ b/pkg/downloader/chart_downloader.go
@@ -171,7 +171,11 @@ func (c *ChartDownloader) getOciURI(ref, version string, u *url.URL) (*url.URL, 
 		}
 	}
 
-	u.Path = fmt.Sprintf("%s:%s", u.Path, tag)
+	// If the tag is already inside the OCI URI, don't concatenate u.Path with tag.
+	// Otherwise the tag gets invalid.
+	if !(strings.Contains(u.Path, ":"+tag)) {
+		u.Path = fmt.Sprintf("%s:%s", u.Path, tag)
+	}
 
 	return u, err
 }


chintai-paul:
Consider the edge case when the version in the OCI URI and the provided version don't match.
What should the outcome be?
Overriding the OCI URI tag would avoid another `invalid tag` issue.

pkg/downloader/chart_downloader.go
```diff
@@ -171,7 +171,11 @@ func (c *ChartDownloader) getOciURI(ref, version string, u *url.URL) (*url.URL,
 		}
 	}
 
-	u.Path = fmt.Sprintf("%s:%s", u.Path, tag)
+	// If the tag is already inside the OCI URI, don't concatenate u.Path with tag.
+	// Otherwise the tag gets invalid.
+	if !(strings.Contains(u.Path, ":"+tag)) {
```
gjenkins8, line 176: I think we should use proper OCI URI parsing, rather than string manipulation. For example (as an extreme edge case), if the first characters of a URI which contains a password match the tag, this will break.

Overall this solution feels like it is working around a larger problem too

PR #13471, opened 2024-11-21
Use a more direct and less error-prone return value

wangjingcun:

Since err != nil has been judged before, nil is returned directly here, which is more obvious, readable and less error-prone.

**If applicable**:

### DIFF
--- a/pkg/kube/roundtripper.go
+++ b/pkg/kube/roundtripper.go
@@ -49,7 +49,7 @@ func (rt *RetryingRoundTripper) roundTrip(req *http.Request, retry int, prevResp
 	b, err := io.ReadAll(resp.Body)
 	resp.Body.Close()
 	if err != nil {
-		return resp, rtErr
+		return resp, err
 	}
 
 	var ke kubernetesError
@@ -58,10 +58,10 @@ func (rt *RetryingRoundTripper) roundTrip(req *http.Request, retry int, prevResp
 	r.Seek(0, io.SeekStart)
 	resp.Body = io.NopCloser(r)
 	if err != nil {
-		return resp, rtErr
+		return resp, nil
 	}
 	if ke.Code < 500 {
-		return resp, rtErr
+		return resp, nil
 	}
 	// Matches messages like "etcdserver: leader changed"
 	if strings.HasSuffix(ke.Message, "etcdserver: leader changed") {
@@ -71,7 +71,7 @@ func (rt *RetryingRoundTripper) roundTrip(req *http.Request, retry int, prevResp
 	if strings.HasSuffix(ke.Message, "raft proposal dropped") {
 		return rt.roundTrip(req, retry-1, resp)
 	}
-	return resp, rtErr
+	return resp, nil
 }
 
 type kubernetesError struct {
--- a/pkg/plugin/installer/vcs_installer.go
+++ b/pkg/plugin/installer/vcs_installer.go
@@ -63,7 +63,7 @@ func NewVCSInstaller(source, version string) (*VCSInstaller, error) {
 		Version: version,
 		base:    newBase(source),
 	}
-	return i, err
+	return i, nil
 }
 
 // Install clones a remote repository and installs into the plugin directory.
--- a/pkg/provenance/sign_test.go
+++ b/pkg/provenance/sign_test.go
@@ -34,7 +34,7 @@ const (
 	// phrase. Use `gpg --export-secret-keys helm-test` to export the secret.
 	testKeyfile = "testdata/helm-test-key.secret"
 
-	// testPasswordKeyFile is a keyfile with a password.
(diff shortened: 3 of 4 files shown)


pkg/plugin/installer/vcs_installer.go
```diff
@@ -63,7 +63,7 @@ func NewVCSInstaller(source, version string) (*VCSInstaller, error) {
 		Version: version,
 		base:    newBase(source),
 	}
-	return i, err
```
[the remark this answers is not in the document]
wangjingcun: I think this check has been done on line 58 above, so in fact, the err on line 66 must be nil.

Directly returning can avoid many problems

image

godhanipayal (approved):

pkg/kube/roundtripper.go
```diff
@@ -71,7 +71,7 @@ func (rt *RetryingRoundTripper) roundTrip(req *http.Request, retry int, prevResp
 	if strings.HasSuffix(ke.Message, "raft proposal dropped") {
 		return rt.roundTrip(req, retry-1, resp)
```
TerryHowe, line 72: Looks like possible errors on line 61 and 52 above. The whole `rtErr` thing isn't great.

TerryHowe (changes requested):
I think `rtErr` issues should be fixed

pkg/kube/roundtripper.go
wangjingcun: Maybe this is a small bug.

It has been modified, thanks for your review.

pkg/kube/roundtripper.go
```diff
@@ -58,10 +58,10 @@ func (rt *RetryingRoundTripper) roundTrip(req *http.Request, retry int, prevResp
 	r.Seek(0, io.SeekStart)
 	resp.Body = io.NopCloser(r)
 	if err != nil {
-		return resp, rtErr
+		return resp, err
```
TerryHowe, line 61: I think this is waiting for more data, so it should return nil

pkg/kube/roundtripper.go
```diff
@@ -58,10 +58,10 @@ func (rt *RetryingRoundTripper) roundTrip(req *http.Request, retry int, prevResp
...
 	}
 	if ke.Code < 500 {
-		return resp, rtErr
+		return resp, err
```
TerryHowe, line 64: This was ignoring errors less than 500, so nil probably

pkg/kube/roundtripper.go
wangjingcun: Modified. Please review again.

pkg/kube/roundtripper.go
TerryHowe: I haven't looked at the tests, but if this change is causing a failure, I would guess that the tests are wrong or line 52 should be nil too. Line 52 looks correct to me though, but maybe I'm missing something.

TerryHowe (approved):
/lgtm

pkg/kube/roundtripper.go
wangjingcun: Thanks for your reply, sorry I just saw it. The previous test failed because the `LoadFile`  method in `pkg/chart/loader/archive.go` should not be modified, that is

```go
c, err := LoadArchive(raw)
if err != nil {
if err == gzip.ErrHeader {
return nil, fmt.Errorf("file '%s' does not appear to be a valid chart file (details: %s)", name, err)
}
}
return c, err
```

I have rolled back, and all tests have passed.

wangjingcun:
Anything need I to do ? I am happy to do it.

pkg/provenance/sign_test.go
```diff
@@ -34,7 +34,7 @@ const (
 	// phrase. Use `gpg --export-secret-keys helm-test` to export the secret.
 	testKeyfile = "testdata/helm-test-key.secret"
 
-	// testPasswordKeyFile is a keyfile with a password.
```
gjenkins8, line 37: this is of course inconsequential. but please prefer distinct PRs for distinct changes (this is fixing a spelling mistake, other parts of the PR are fixing error return values)

gjenkins8 (dismissed):

pkg/provenance/sign_test.go
wangjingcun: At first I thought that opening a new PR would increase the burden of the maintainer

Thank you for reminding, I will pay attention next time. Do I need to return to the revision here?

wangjingcun:
friendly ping~

godhanipayal (approved):

wangjingcun:
Thanks for your approved.

robertsirc (approved):
LGTM

mattfarina (approved):
rtErr is nil in all these places so this is better documented. It's going to help us better see what's happening here.

PR #13480, opened 2024-11-22
Make the authorizer and registry authorizer configurable

rynowak:
Fixes: #12584

This change makes the authorizer and registryAuthorizer of the registry client configurable via options. This allows Go SDK users to override the authentication behavior of the client.

This PR makes both the authorizer and registryAuthorizer configurable because depending on the exact scenario that may be needed. The default registryAuthorizer only supports a specific implementation of the authorizer.

This PR ports #12588 to the dev-v3 branch. I'm opening this per-request after it was merged to v4/main https://github.com/helm/helm/pull/12588#issuecomment-2492867470

**If applicable**:

### DIFF
--- a/pkg/registry/client.go
+++ b/pkg/registry/client.go
@@ -175,6 +175,26 @@ func ClientOptWriter(out io.Writer) ClientOption {
 	}
 }
 
+// ClientOptAuthorizer returns a function that sets the authorizer setting on a client options set. This
+// can be used to override the default authorization mechanism.
+//
+// Depending on the use-case you may need to set both ClientOptAuthorizer and ClientOptRegistryAuthorizer.
+func ClientOptAuthorizer(authorizer auth.Client) ClientOption {
+	return func(client *Client) {
+		client.authorizer = authorizer
+	}
+}
+
+// ClientOptRegistryAuthorizer returns a function that sets the registry authorizer setting on a client options set. This
+// can be used to override the default authorization mechanism.
+//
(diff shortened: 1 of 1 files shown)


TerryHowe (commented):
At this point I think we should revisit this after oras v2 merges

scottrigby:
> At this point I think we should revisit this after oras v2 merges

This is just a backport for v3 of the PR that was already merged into main (now that it's for v4). At this point there is no significant divergence.

scottrigby (approved):
@rynowak thanks for the backport. Approving with same reasoning as #12588

gjenkins8 (approved):

pkg/registry/client.go
```diff
@@ -175,6 +175,26 @@ func ClientOptWriter(out io.Writer) ClientOption {
...
+	return func(client *Client) {
+		client.authorizer = authorizer
+	}
+}
+
+// ClientOptRegistryAuthorizer returns a function that sets the registry authorizer setting on a client options set. This
+// can be used to override the default authorization mechanism.
+//
+// Depending on the use-case you may need to set both ClientOptAuthorizer and ClientOptRegistryAuthorizer.
+func ClientOptRegistryAuthorizer(registryAuthorizer *registryauth.Client) ClientOption {
```
TerryHowe, line 192: No separation of concerns here since `registryauth.Client` is an ORAS object and a v1 object too.

TerryHowe (changes requested):

PR #13481, opened 2024-11-25
feat: Enable CPU and memory profiling

banjoh:

Add capability to profile cli commands to allow users to troubleshoot memory and CPU usage. This is based on  https://go.dev/blog/pprof. Users can use https://github.com/google/pprof/blob/main/doc/README.md to analyse the collected profiling data

Example run
```sh
$ helm show all bitnami/nginx --memprofile mem.prof --cpuprofile cpu.prof

# Visualize graphs. You need to have installed graphviz package in your system
$ go tool pprof -http=":8000" cpu.prof
Serving web UI on http://localhost:8000
```

You will endup with a view similar to below
image

**If applicable**:


CONTRIBUTING.md
```diff
@@ -276,12 +276,24 @@ Like any good open source project, we use Pull Requests (PRs) to track code chan
...
     - If the owner of a PR is _not_ listed in `OWNERS`, any core maintainer may merge the PR.
 
-#### Documentation PRs
+### Documentation PRs
 
 Documentation PRs should be made on the docs repo: <https://github.com/helm/helm-www>. Keeping Helm's documentation up to date is highly desirable, and is recommended for all user facing changes. Accurate and helpful documentation is critical for effectively communicating Helm's behavior to a wide audience.
 
 Small, ad-hoc changes/PRs to Helm which introduce user facing changes, which would benefit from documentation changes, should apply the `docs needed` label. Larger changes associated with a HIP should track docs via that HIP. The `docs needed` label doesn't block PRs, and maintainers/PR reviewers should apply discretion judging in whether the `docs needed` label should be applied.
 
+### Testing PRs
```
TerryHowe, line 285:
```suggestion
### Profiling PRs
```
I feel like you could focus this section on profiling and drop the first two sentences from the paragraph.

cmd/helm/root.go
```diff
@@ -95,6 +95,26 @@ func newRootCmd(actionConfig *action.Configuration, out io.Writer, args []string
 		Short:        "The Helm package manager for Kubernetes.",
 		Long:         globalUsage,
 		SilenceUsage: true,
+		PersistentPreRun: func(cmd *cobra.Command, _ []string) {
+			cpuprof, err := cmd.Flags().GetString("cpuprofile")
```
gjenkins8, line 99: Would it be better to expose profiling as env vars? e.g. 

```
HELM_PPROF=cpu=/path/to/cpu.prof,mem=/path/to/mem.prof
```

The rationale is this seems a little less intrusive than cli flags. Thoughts?

gjenkins8: Is there prior art to CLIs adding pprof support too?

banjoh: Using env vars makes sense. I'll make the change

> Is there prior art to CLIs adding pprof support too?

Besides https://github.com/replicatedhq/troubleshoot, which is a project I contribute to, I have not come across any other. I know web server implementations such as Gorilla Mux and Negroni expose debug info using https://pkg.go.dev/net/http/pprof.

banjoh: For what its worth, golang recommends this usage https://pkg.go.dev/runtime/pprof

gjenkins8: > Using env vars makes sense. I'll make the change

Thanks, I do think env var(s) would be better. The way I think about it, is pprof is an advanced usage. And (newbie) users who run `helm cmd --help` will have the pprof flags presented, and this just adds to the things they need to parse and understand.

(so that said, flags I guess would be okay if they were hidden to the casual user)

cmd/helm/root.go
```diff
@@ -206,6 +216,9 @@ func newRootCmd(actionConfig *action.Configuration, out io.Writer, args []string
 	// Check for expired repositories
 	checkForExpiredRepos(settings.RepositoryConfig)
 
+	// CPU and memory profiling flags that are available to all commands
+	addProfilingFlags(cmd)
```
gjenkins8, line 220: (mostly duplicated from other comment)

Do we really want to have these flags, which are only for certain, advanced use-cases. Exposed to every user who runs `helm cmd --help`?

The way I think about it, is pprof is an advanced usage. And (newbie) users who run `helm cmd --help` will have the pprof flags presented, and this just adds to the things they need to parse and understand.

(so that said, flags I guess would be okay if they were hidden to the casual user)

cmd/helm/profiling.go
```diff
@@ -0,0 +1,124 @@
...
+		return fmt.Errorf("errors while stopping profiling: [%s]", strings.Join(errs, ", "))
+	}
+
+	return nil
+}
+
+// addProfilingFlags adds the --cpuprofile and --memprofile flags to the given command.
+func addProfilingFlags(cmd *cobra.Command) {
+	// Persistent flags to make available to subcommands
+	cmd.PersistentFlags().String("cpuprofile", "", "File path to write cpu profiling data")
```
gjenkins8, line 97: nit: if we kleep the flags, we should call them:
```
--pprof-cpuprofile
--pprof-memprofile
```

As this groups them together

gjenkins8 (commented):
this is pretty nice. just need to figure out the cli flags vs env var thing, then LGTM to me

banjoh:
> this is pretty nice. just need to figure out the cli flags vs env var thing, then LGTM to me

Thanks

I have made changes to prefer environment variables over CLI flags. Users can invoke this functionality using

```
HELM_PPROF_CPU_PROFILE=cpu.prof HELM_PPROF_MEM_PROFILE=mem.prof helm create foo
```

banjoh:
> this is pretty nice. just need to figure out the cli flags vs env var thing, then LGTM to me

Thanks @gjenkins8 

I have made changes to prefer environment variables over CLI flags. Users can invoke this functionality using

```sh
HELM_PPROF_CPU_PROFILE=cpu.prof HELM_PPROF_MEM_PROFILE=mem.prof helm create foo
```

gjenkins8 (dismissed):
nice

TerryHowe (approved):

cmd/helm/profiling.go
```diff
@@ -0,0 +1,91 @@
...
+		if err != nil {
+			return fmt.Errorf("could not create CPU profile: %w", err)
+		}
+		if err := pprof.StartCPUProfile(cpuProfileFile); err != nil {
+			cpuProfileFile.Close()
+			cpuProfileFile = nil
+			return fmt.Errorf("could not start CPU profile: %w", err)
+		}
+	}
+	return nil
```
gjenkins8, line 54: `if memProfilePath != "" {` seems to be missing here (bad merge?)

TerryHowe: I thought a snapshot was just taken at the end.

banjoh: > I thought a snapshot was just taken at the end.

Correct. Unlike CPU, there is no step where collecting memory profiling data is started. The snapshot can be taken at any time.

banjoh:
@gjenkins8 I made a tiny documentation update to the PR. Your approval was dismissed.

TerryHowe (approved):
/lgtm

cmd/helm/profiling.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+// stopProfiling stops profiling CPU and memory usage.
+// It writes memory profile to the file path specified in HELM_PPROF_MEM_PROFILE
+// environment variable.
+func stopProfiling() error {
+	errs := []string{}
```
gjenkins8, line 61:
```
errs := []error{}
```

cmd/helm/profiling.go
```diff
@@ -0,0 +1,91 @@
...
+
+		runtime.GC() // get up-to-date statistics
+		if err := pprof.WriteHeapProfile(f); err != nil {
+			errs = append(errs, err.Error())
+		}
+	}
+
+	if len(errs) > 0 {
+		return fmt.Errorf("errors while stopping profiling: [%s]", strings.Join(errs, ", "))
+	}
```
gjenkins8, line 88:
```suggestion
	if err := errors.Join(errs...); err != nil {
		return fmt.Errorf("error(s) while stopping profiling: %w", err)
	}
```

gjenkins8 (dismissed):
Spotted one more thing to potentially tidy up. LGTM either way

cmd/helm/profiling.go
gjenkins8: oops, my bad

cmd/helm/profiling.go
banjoh: Good suggestion

banjoh:
Here is an example showing warnings being printed. The operation succeeds

```sh
[evans] $ HELM_PPROF_CPU_PROFILE=/my/non-existent/dir/cpu.prof HELM_PPROF_MEM_PROFILE=/my/non-existent/dir/mem.prof helm create foo
root.go:100: Warning: Failed to start profiling: could not create CPU profile: open /my/non-existent/dir/cpu.prof: no such file or directory
Creating foo
root.go:105: Warning: Failed to stop profiling: error(s) while stopping profiling: open /my/non-existent/dir/mem.prof: no such file or directory
[evans] $ ll foo/
total 16
drwxr-xr-x  7 evans  224 Feb 17 18:57 ./
drwxr-xr-x  8 evans  256 Feb 17 18:57 ../
-rw-r--r--  1 evans  349 Feb 17 18:57 .helmignore
-rw-r--r--  1 evans 1139 Feb 17 18:57 Chart.yaml
drwxr-xr-x  2 evans   64 Feb 17 18:57 charts/
drwxr-xr-x 10 evans  320 Feb 17 18:57 templates/
-rw-r--r--  1 evans 4289 Feb 17 18:57 values.yaml
```

gjenkins8 (approved):

scottrigby (approved):
Tested all scenarios 👍 and code looks good. Nice work

PR #13483, opened 2024-11-25
fix(helm): pass down username/password CLI parameters to OCI registry clients (v3 backport)

banjoh:

Backport https://github.com/helm/helm/pull/12769 to v3 branch

**If applicable**:


joejulian (approved):

cmd/helm/push.go
```diff
@@ -40,6 +40,8 @@ type registryPushOptions struct {
 	caFile                string
 	insecureSkipTLSverify bool
 	plainHTTP             bool
+	password              string
+	username              string
```
TerryHowe, line 44: Too late now, but I thought the idea was to authenticate before using push

sabre1041 (approved):
LGTM

cmd/helm/push.go
banjoh: > authenticate before using push

Do you mean `helm registry login` then `helm push` or are you referring to a logic in code?

TerryHowe: Yeh, I thought these commands were set up to authenticate with `helm login` and all the other commands need to deal with TLS etc, but not authn Sounds like my assumption was incorrect.

sabre1041: > Yeh, I thought these commands were set up to authenticate with `helm login` and all the other commands need to deal with TLS etc, but not authn Sounds like my assumption was incorrect.

You can do both. You can login to allow multiple invocations without having the explicitly specifying each option, or one can integrate ad-hoc, as enabled through this enhancement

EronWright:
Should this be assigned to the [3.17.0](https://github.com/helm/helm/milestone/147) milestone?

mardonner:
Is this v3 backport currently part of any release?
With Helm 3.18.4 I get 401s when using --username and --password.
Only helm registry login works for me.

banjoh:
> Is this v3 backport currently part of any release?

Yes, this is part of v3

> With Helm 3.18.4 I get 401s when using --username and --password

It was released in [v3.17.0](https://github.com/helm/helm/releases/tag/v3.17.0) and it works with that version. It doesn't in v3.18.4. I'll investigate

PR #13497, opened 2024-12-02
Unify behavior of includeFun in regard to nil value output

### LINKED ISSUE #13487: <no value> should be considered empty
Returning `nil` from a template function yields a `<no value>`, which is unintuitively not considered empty.

```helm
# _helpers.tpl
---
{{- define "noval" -}}
  {{- and nil }}
{{- end -}}
```
```helm
{{- $noval := include "noval" . }}
{{- fail (print
  "empty " nil ": " (empty nil) ", "
  "empty " $noval ": " (empty $noval)
)}}
```

Current output
```
empty <nil>: true, empty <no value>: false
````
Expected output
```
empty <nil>: true, empty <no value>: true
```

---

```console
$ helm version
version.BuildInfo{Version:"3.16.2", GitCommit:"5a5449dc42be07001fd5771d56429132984ab3ab", GitTreeState:"clean", GoVersion:"go1.23.1"}
```

V02460:
Unifies the behavior of `includeFun` with `tplFun` and `render` regarding rendered nil values ("\<no value>"). This is done by applying the same replaceAll function to the output as the other two functions, replacing any occurrence of "\<no value>" in the string output of `includeFun` with the empty string.

Closes #13487

Note that this is still a hack as not only an actual nil value will be rendered as the empty string, but also any actual string output of "\<no value>". Sadly, I’m not deep enough into Go Templates to come up with a proper solution, so I propose to make Helm’s behavior consistent at least.

**If applicable**:

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -139,7 +139,7 @@ func includeFun(t *template.Template, includedNames map[string]int) func(string,
 		}
 		err := t.ExecuteTemplate(&buf, name, data)
 		includedNames[name]--
-		return buf.String(), err
+		return strings.ReplaceAll(buf.String(), "<no value>", ""), err
 	}
 }
 
--- a/pkg/engine/engine_test.go
+++ b/pkg/engine/engine_test.go
@@ -1300,3 +1300,34 @@ func TestRenderTplMissingKeyString(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+func TestRenderNilValue(t *testing.T) {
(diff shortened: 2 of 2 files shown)


V02460:
I added two additional commits with unittests:

1. Positive checks that there is no stray <no value> somewhere.
2. Negative checks that valid <no value> strings are not lost on us. These currently fail.

In the spirit of being test-driven, it would be great to fix the failing tests, but I don’t know how. In case nobody steps up, we can revert (2) and be done with it, too.

PR #13514, opened 2024-12-07
feat: add latest version and outdated flag to plugin list command

idsulik:
Added `LATEST` column to plugin list output and `--outdated` flag to show only plugins with available updates, helping users identify plugins that need updating.

example:
image

**If applicable**:


idsulik:
@gjenkins8 hi! What do you think about the feature? It was inspired by Homebrew's outdated command, which makes it easy to see which plugins have updates available.

idsulik:
@robertsirc hi! Could you please check the PR? Thank you

PR #13525, opened 2024-12-12
Rewrite Chart.yaml dependencies' paths to a fully self-contained tgz package

killerwhile:
Rewrite Chart.yaml dependencies' paths to local file://charts/...chart... to package a fully self-contained tgz package

This PR enables rewriting Chart.yaml dependencies' paths to point all dependenciese to `file://charts/...dependency chart...`, so the resulting tgz package in fully self-contained.
This is usefull when dealing with a mono-repo, where charts dependencies contains relative path such as file://../another-dependent-chart, or when transfering a chart with public dependencies to an air-gap environment. The resulting chart is self-contained and doesn't have any other external dependencies.

**If applicable**:


robertsirc:
Hello Friend. You need to correct a few things your DCO is broken and the branch has some conflicts.

TerryHowe:
https://github.com/helm/helm/pull/13525/checks?check_run_id=45827386490

Issue #12987, opened 2024-04-26
 toToml filter converts int to float


dud225:
Given the following data structure in values.yaml:
```
config:
  bar: 9
```

And this in my configmap template:

```
data:
  telegraf-local.conf: |-
{{ .Values.config | toToml | indent 4 }}
```

I get this:
```
data:
  telegraf-local.conf: |-
    foo = 9.0
```
Even casting the variable via bar: `!!int 9` doesn't change the behavior, can't reproduce using the toJson filter either.

I'm running Helm version.BuildInfo{Version:"v3.14.4", GitCommit:"81c902a123462fd4052bc5e9aa9c513c4c8fc142", GitTreeState:"clean", GoVersion:"go1.21.9"}

gjenkins8:
`toToml` uses github.com/BurntSushi/toml to encode:

```
	e := toml.NewEncoder(b)
	err := e.Encode(v)
```

https://github.com/helm/helm/blob/c6beb169d26751efd8131a5d65abe75c81a334fb/pkg/engine/funcs.go#L125-L133

Either Helm is representing the value internally as float, and toToml is rendering that. Or the design decision/issue with rendering ints as floats lies upstream with github.com/BurntSushi/toml. The former should be determined first I think.

johanfleury:
IIRC, the internal representation is float. #6010 was made to use [`json.Number`](https://pkg.go.dev/encoding/json#Number) but was then reverted.

It seems like the toml module supports `json.Number`: https://github.com/BurntSushi/toml/blob/77ce8589413d1fc280be524c8992599c42027c61/encode.go#L259-L272

CalvinKrist:
I made a draft PR to demonstrate this, but adding unit tests for the `toToml` function that test the int and float conversion makes it appear that `BurntSushi` is handling this correctly and the issue is likely elsewhere https://github.com/helm/helm/pull/13012

CalvinKrist:
That said, I did repro this issue. I did `helm create` to make a simple helm chart, added the following config map:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: game-demo
data:
  telegraf-local.conf: |-
{{ .Values.config | toToml | indent 4 }}
```

copied their values:
```
config:
  bar: 9
```

and then `helm template .` returns:
```
---
# Source: mychart/templates/cm.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: game-demo
data:
  telegraf-local.conf: |-
    bar = 9.0
[...]

Issue #6539, opened 2019-09-30
toToml filter converts integers and floats to quoted-strings


johanfleury:
# Bug Report

When using the `toToml` filter in a template, integers and floats are converted to quoted-strings (which is not the case with `toYaml`).

This lead to issues with other TOML decoders (`telegraf` will crash with: `E! [telegraf] Error running agent: Error parsing /etc/telegraf/telegraf.conf, line 41: (config.AgentConfig.MetricBatchSize) cannot unmarshal TOML string into int`).

# Example

The following “Chart”:

```
$ tree
.
├── Chart.yaml
├── templates
│   └── test.yaml
└── values.yaml

1 directory, 3 files

$ cat Chart.yaml
---
name: test
version: 0.1.0

$ cat values.yaml
---
config:
  test1: 1

  test2:
    test2: 1

  test3:
    test3:
      - test3: 1

  test4: 1.1

  test5:
    test5: 1.1

  test6:
    test6:
      - test6: 1.1

$ cat templates/test.yaml
{{ .Values.config | toToml }}
{{ .Values.config | toYaml }}
```

Produces:

```
$ helm template .
---
# Source: test/templates/test.yaml
test1 = "1"
test4 = "1.1"

[test2]
  test2 = "1"

[test3]

  [[test3.test3]]
    test3 = "1"

[test5]
  test5 = "1.1"

[test6]

  [[test6.test6]]
    test6 = "1.1"

test1: 1
test2:
  test2: 1
test3:
  test3:
  - test3: 1
test4: 1.1
test5:
  test5: 1.1
test6:
  test6:
  - test6: 1.1
```

# Related issues / pull requests

This issue was introduced in PR #6010 which was made to fix #1707 and #2474.

# Installation details

Output of `helm version`:
```
Client: &version.Version{SemVer:"v2.14+unreleased", GitCommit:"3b51dfbdbcf34c9ece3e6840dad0ab70cc13c1c2", GitTreeState:"clean"}                                                                                    
Server: &version.Version{SemVer:"v2.14.3", GitCommit:"0e7f3b6637f7af8fcfddb3d2941fcc7cbebb0085", GitTreeState:"clean"}   
```

Output of `kubectl version`: **not relevant**

Cloud Provider/Platform (AKS, GKE, Minikube etc.): **not relevant**

bacongobbler:
[...]

Issue #2474, opened 2017-05-22
toToml filter converts int to float


mattsch:
Given the following data structure in values.yaml:
```yaml
config:
  bar: 9
```
And this in my configmap template:
```yaml
data:
  telegraf-local.conf: |-
{{ .Values.config | toToml | indent 4 }}
```

I get this:
```
data:
  telegraf-local.conf: |-
    foo = 9.0
```

Even casting the variable via `bar: !!int 9` doesn't change the behavior, can't reproduce using the toJson filter either.

technosophos:
Which version of Helm/Tiller?

mattsch:
Sorry, it was with 2.4.2.

mlieberman85:
I'm seeing the same issue with 2.5.0

copumpkin:
From poking around the code a bit, it seems like the YAML parser Helm is using converts YAML to JSON and then parses that. And although the YAML spec distinguishes between integers and floats, the JSON spec doesn't (as far as I can tell), so the conversion process is lossy.

For reference, the conversion to JSON is happening here: https://github.com/ghodss/yaml/blob/73d445a93680fa1a78ae23a5839bad48f32ba1ee/yaml.go#L98

bacongobbler:
related issue: https://github.com/kubernetes/helm/issues/1707

bacongobbler:
Hey @johanfleury, if you could please open a new issue with the version of Helm and Kubernetes you are using as well as a demonstration/reproduction of the issue, that would be great. This issue was opened over two years ago and was likely fixed in https://github.com/helm/helm/pull/6010, so if you could try and reproduce using the canaries, that'd be great. Thanks!

dud225:
reopened as https://github.com/helm/helm/issues/12987

PR #13533, opened 2024-12-13
fix: (toToml) renders int as float

althmoha:
This commit fixes the issue where the `yaml.Unmarshaller` converts all int values into float64; this passes an option to the decoder, enabling int conversion into.

closes #12987 #6539 #2474 

Signed-off-by: Althaf M <althafm@outlook.com>

**What this PR does / why we need it**:

This is required so that helm function toToml prints out the integer values as integers rather than floats. Without this, it prints it as float strings, appending a .0 to the end.

Currently, 

Assuming values.yaml 
```
global:
  key: 13
```

if we have a template that renders using `toToml` function, like 

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: issue-totoml
data: |
  {{ .Values.global | toToml }}
```

Will get rendered as 

```
# Source: issue-totoml/templates/configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: issue-totoml
data: |
  key = 13.0
```

With this PR, it will be rendered as integer as expected : 

```
# Source: issue-totoml/templates/configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: issue-totoml
data: |
  key = 13
```

**Special notes for your reviewer**:

Why does this happen only in the `toToml` function?

The [library](https://github.com/BurntSushi/toml) we use to convert to the TOML uses the reflection to identify the type of elements or values and prints them out based on the type. Other libraries, like YAML and JSON, ignore the element type. This causes the output to be printed out as a float rather than as an integer. The root cause is that when using the `yaml.Unmarshal` library, we are parsing all the integers as float64 values.

**If applicable**:
- [ ] this PR contains user facing changes (the `docs needed` label should be applied if so)
- [x] this PR contains unit tests
- [x] this PR has been tested for backwards compatibility

PR #13534, opened 2024-12-14
fix (helm) : toToml` renders int as float [ backport to v3 ]

althmoha:
This commit fixes the issue where the `yaml.Unmarshaller` converts all int values read from yaml docs into float64; this passes an option to the decoder, enabling int conversion into.

closes https://github.com/helm/helm/issues/12987 https://github.com/helm/helm/issues/6539 https://github.com/helm/helm/issues/2474

Signed-off-by: Althaf M [althafm@outlook.com](mailto:althafm@outlook.com)

This is required so that helm function `toToml` prints out the integer values as integers rather than floats. Without this, it prints it as float strings, appending a `.0` to the end.

Currently, 
Assuming values.yaml 
```
global:
  key: 13
```

if we have a template that renders using `toToml` function, like 

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: issue-totoml
data: |
  {{ .Values.global | toToml }}
```

Will get rendered as 

```
# Source: issue-totoml/templates/configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: issue-totoml
data: |
  key = 13.0
```

With this PR, it will be rendered as integer as expected : 

```
# Source: issue-totoml/templates/configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: issue-totoml
data: |
  key = 13
```

Why does this happen only in the `toToml` function?

I have ensured all tests pass with the `make test-unit`  command and fixed the failing tests due to the type change.

The [library](https://github.com/BurntSushi/toml) we use to convert to the TOML uses the reflection to identify the type of elements or values and prints them out based on the type. Other libraries, like YAML and JSON, ignore the element type. This causes the output to be printed out as a float rather than as an integer. The root cause is that when using the `yaml.Unmarshal` library, we are parsing all the integers as float64 values.

**If applicable**:

### DIFF
--- a/cmd/helm/template_test.go
+++ b/cmd/helm/template_test.go
@@ -161,6 +161,11 @@ func TestTemplateCmd(t *testing.T) {
 			cmd:    fmt.Sprintf("template '%s' -f %s/extra_values.yaml", chartPath, chartPath),
 			golden: "output/template-subchart-cm-set-file.txt",
 		},
+		{
+			name:   "check toToml function rendering",
+			cmd:    fmt.Sprintf("template '%s'", "testdata/testcharts/issue-totoml"),
+			golden: "output/issue-totoml.txt",
+		},
 	}
 	runTestCmd(t, tests)
 }
--- a/cmd/helm/testdata/output/issue-totoml.txt
+++ b/cmd/helm/testdata/output/issue-totoml.txt
@@ -1,0 +1,8 @@
+---
+# Source: issue-totoml/templates/configmap.yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: issue-totoml
+data: |
+  key = 13
--- a/cmd/helm/testdata/testcharts/issue-totoml/Chart.yaml
+++ b/cmd/helm/testdata/testcharts/issue-totoml/Chart.yaml
@@ -1,0 +1,3 @@
+apiVersion: v2
+name: issue-totoml
+version: 0.1.0
--- a/cmd/helm/testdata/testcharts/issue-totoml/templates/configmap.yaml
+++ b/cmd/helm/testdata/testcharts/issue-totoml/templates/configmap.yaml
@@ -1,0 +1,6 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: issue-totoml
+data: |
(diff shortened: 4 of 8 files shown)


TerryHowe (commented):
dup of https://github.com/helm/helm/pull/13533 ?

althmoha:
> dup of #13533 ?

Yes, backport to dev-v3

sabre1041 (approved):
LGTM

althmoha:
Will this be merged any time soon ?

robertsirc (approved):
LGTM

althmoha:
Will this be added to the 3.17.1  milestone : https://github.com/helm/helm/milestone/154 ?

althmoha:
Any update on this ?

guru-disco:
Any updates here? Would like to see this merged as soon as possible. Thanks :)

nushemto:
> Any update on this ?

+1

althmoha:
@robertsirc  We look forward to merging this to update our helm executables. Could you please help us here?

mattfarina (approved):
I did analysis to see what the impact would be to charts. This had been marked as a bug in the past. There have been numerous issues over time about it. Looking over the charts on GitHub there are numerous places `toToml` is used with a flexible structure. For these cases it's hard to know the input that would be used there without learning those applications.

I also tested various in/out combinations to see what would happen with this change and a chart.

The impact of this fix should be minimal and maybe none at all, except for people who want this fix.

I could be wrong but have been conservative in handling this one to try and ensure the impact is low to none.

WieslerAA:
Is this included in 3.17.1? Because I just tested it with this version and I still get `key = 13.0`.

althmoha:
> Is this included in 3.17.1? Because I just tested it with this version and I still get `key = 13.0`.

It would have been great if this was included in 3.17.1, or at least 3.17.2 , but from the above comment it looks like it was added to 3.18.0,  which means we may have to wait till may to get this fix :( 

```
[mattfarina](https://github.com/mattfarina) added this to the [3.18.0](https://github.com/helm/helm/milestone/153) milestone [2 weeks ago](https://github.com/helm/helm/pull/13534#event-16200472785)
```

PR #13537, opened 2024-12-16
implement post-renderer for dependencies

### LINKED ISSUE #13536: implement post-renderer for dependencies
### Feature Proposal (with implementation)

Add support for a post-renderer operation on dependency such that the operation can be "packaged" with the root chart avoiding the need for the dependency to be forked.

e.g. _(YTT just used as an example.)_
```
dependencies:
- name: nginx
  version: "1.2.3"
  repository: "https://example.com/charts"
  postRenderer:
    command: ytt
    args:
    - "--file"
    - "overlays/nginx/overlay.yaml"
    - "--file"
    - "-"
```

### The Appeal

First. Today, Helm's dependency management is handcuffed by the fact that it does not have the ability to mutate the dependency's manifests as part of the deliverables of the root chart. This often leads to the need to fork the dependency.

This means that I, as a chart creator, cannot depend on a third party chart using `dependencies` if said chart overlooks even a single concern that is important to me. This belies composition of charts.

Second. Kubernetes is an evolving landscape. Helm's current design does not allow me as chart developer to adopt new transversal features until all the dependencies of my chart have adopted said features, leading again to the need for forking.

As an example, suppose I want to add [Pod Scheduling Readiness](https://kubernetes.io/docs/concepts/scheduling-eviction/pod-scheduling-readiness/) to all the pods in my chart and it's dependencies. Again, I would be forced to fork every chart until such a time as they adopt this feature.

### The Upside

Helm already has a internal mechanism that could be leveraged to address this use case with very little change. Please find a PR (incoming) for review that demonstrates a working implementation of this.

rotty3000:
resolves #13536

e.g.
```yaml
dependencies:
- name: nginx
  version: "1.2.3"
  repository: "https://example.com/charts
  postRenderer:
    command: ytt
    args:
      - "--file"
      - "overlays/nginx/overlay.yaml"
      - "--file"
      - "-"
```

See https://github.com/helm/helm/issues/13536

The change is fairly simple

**If applicable**:

### DIFF
--- a/pkg/chart/dependency.go
+++ b/pkg/chart/dependency.go
@@ -17,6 +17,11 @@ package chart
 
 import "time"
 
+type PostRendererOptions struct {
+	Command string   `json:"command" yaml:"command"`
+	Args    []string `json:"args,omitempty" yaml:"args,omitempty"`
+}
+
 // Dependency describes a chart upon which another chart depends.
 //
 // Dependencies can be used to express developer intent, or to capture the state
@@ -47,6 +52,8 @@ type Dependency struct {
 	ImportValues []interface{} `json:"import-values,omitempty" yaml:"import-values,omitempty"`
 	// Alias usable alias to be used for the chart
 	Alias string `json:"alias,omitempty" yaml:"alias,omitempty"`
+	// A post rendering operation that will be applied to the rendered outputs of this dependency
+	PostRenderer *PostRendererOptions `json:"postRenderer,omitempty" yaml:"postRenderer,omitempty"`
 }
 
 // Validate checks for common problems with the dependency datastructure in
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -17,6 +17,7 @@ limitations under the License.
 package engine
 
 import (
+	"bytes"
 	"fmt"
 	"log"
 	"path"
@@ -31,6 +32,7 @@ import (
 
 	"helm.sh/helm/v3/pkg/chart"
 	"helm.sh/helm/v3/pkg/chartutil"
+	"helm.sh/helm/v3/pkg/postrender"
 )
 
 // Engine is an implementation of the Helm rendering implementation for templates.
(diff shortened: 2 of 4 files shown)


gjenkins8 (commented):
Hi, thanks for the PR. How to improve "post-rendering" has been discussed a lot 

iiuc (I wasn't part of the discussions), there were two main reasons the post-renderer wasn't written like this originally, and rather implemented as an "add on" CLI flag:
1. the security concerns of a chart being able to control and execute malicious/arbitrary code
2. the chart now depends on the configuration of the users system ie. requires `ytt` to be installed (and be the right version, etc)

For both these reasons, I understand it was decided to implement the post-renderer as a CLI flag. Where the user has more control of these things. 

We would like to solve these problem with "plugins". WASM in particular would meet the needed sandboxing and distribution requirements.

rotty3000:
thanks for the feedback @gjenkins8 :)

> the security concerns of a chart being able to control and execute malicious/arbitrary code
> the chart now depends on the configuration of the users system ie. requires ytt to be installed (and be the right version, etc)

I anticipated these concerns and wondered if the execution were container based (i.e. similar to how hooks execute a job) then perhaps this concern would be addressed?

> implement the post-renderer as a CLI flag

This does not address the use case in question, which is **adjusting/mutating/refining the outputs of a dependency**. I want this to be a result of my parent Chart and NOT require the user's input. The goal here is to solve the old forking-third-party-chart issue. I don't want to fork a dependency chart, I just want to mutate it just-in-time and transparently to the user.

If it was possible to have a hook to which all rendered resources were passed before deployment, this would amount to an equally useful solution.

Do you think it would be interesting for me to pursue moving this into a container based solution (more akin to hooks and perhaps following a similar design)?

rotty3000:
> Do you think it would be interesting for me to pursue moving this into a container based solution (more akin to hooks and perhaps following a similar design)?

It certainly would be interesting to me.. But what I mean to ask is, would there be interest from the helm maintainers for me to pursue the suggested approach?

PR #13541, opened 2024-12-17
WIP: Add tests to test TakeOwnership

### LINKED ISSUE #13519: Improve test coverage for TakeOwnership option
While reviewing #13439 to add flags for the SDK function added in #12876, I found a bug only through manual testing. I recommended the fix in this comment https://github.com/helm/helm/pull/13439#pullrequestreview-2494390079 and it was added to the PR. However, this means our test coverage for this feature needs to improve. A test needs to be added for `upgrade --install` with `--take-ownership`, as I outlined in this comment https://github.com/helm/helm/pull/13439#pullrequestreview-2494359393.

alex-kattathra-johnson:
Tests the `TakeOwnership` flow in `RunWithContext`
closes #13519 

**If applicable**:

### DIFF
--- a/pkg/action/install_test.go
+++ b/pkg/action/install_test.go
@@ -798,3 +798,46 @@ func TestInstallWithSystemLabels(t *testing.T) {
 
 	is.Equal(fmt.Errorf("user supplied labels contains system reserved label name. System labels: %+v", driver.GetSystemLabels()), err)
 }
+
+func TestInstallWithTakeOwnershipOption(t *testing.T) {
+	is := assert.New(t)
+	req := require.New(t)
+
+	instAction := installAction(t)
+	instAction.ClientOnly = false
+	instAction.IsUpgrade = false
+	instAction.TakeOwnership = true
+	failer := instAction.cfg.KubeClient.(*kubefake.FailingKubeClient)
+	failer.BuildDummy = true
+	instAction.cfg.KubeClient = failer
+
+	vals := map[string]interface{}{}
+	ctx, done := context.WithCancel(context.Background())
+	res, err := instAction.RunWithContext(ctx, buildChart(), vals)
+	if err != nil {
+		t.Fatalf("Failed install: %s", err)
+	}
+	is.Equal(res.Name, "test-install-release", "Expected release name.")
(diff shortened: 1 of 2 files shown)


alex-kattathra-johnson:
@banjoh: I'm stuck at this error where a dummy object isn't able to be converted to a `meta.Object`. Have any ideas on the best way to proceed?

pkg/action/install_test.go
```diff
@@ -798,3 +798,46 @@ func TestInstallWithSystemLabels(t *testing.T) {
 
 	is.Equal(fmt.Errorf("user supplied labels contains system reserved label name. System labels: %+v", driver.GetSystemLabels()), err)
 }
+
+func TestInstallWithTakeOwnershipOption(t *testing.T) {
+	is := assert.New(t)
+	req := require.New(t)
```
ivelichkovich, line 804: The rest of the code seems to only use is and not require in this test file, maybe worth just removing req and using is.NoError on line 841

pkg/action/install_test.go
```diff
@@ -798,3 +798,46 @@ func TestInstallWithSystemLabels(t *testing.T) {
...
+	vals := map[string]interface{}{}
+	ctx, done := context.WithCancel(context.Background())
+	res, err := instAction.RunWithContext(ctx, buildChart(), vals)
+	if err != nil {
+		t.Fatalf("Failed install: %s", err)
+	}
+	is.Equal(res.Name, "test-install-release", "Expected release name.")
+	is.Equal(res.Namespace, "spaced")
+
+	rel, err := instAction.cfg.Releases.Get(res.Name, res.Version)
```
ivelichkovich, line 823: Overall, IIUC the take-ownership flag is used to update existing k8s resources to be owned by helm using the labels/annotations. So the test should create some resources and test that when take-ownership is set helm is able to install and the resources now have the helm mgmt metadata.

Additionally, it might be good to have a negative test where we test without take-ownership and verify that the install does not work in that case

ivelichkovich: This function shows what that metadata is: https://github.com/helm/helm/blob/main/pkg/action/validate.go#L94

PR #13546, opened 2024-12-19
Update to Go 1.23

mattfarina:
Multiple changes were made to pass linting. Some Go built-in names are being used for variables (e.g., min). This happens in the Go source itself including the Go standard library and is not always a bad practice.

To handle allowing some built-in names to be used the linter config is updated to allow (via opt-in) some names to pass. This allows us to still check for re-use of Go built-in names and opt-in to any new uses.

There were also several cases where a value was checked for nil before checking its length when this is already handled by len() or the types default value. These were cleaned up.

The license validation was updated because it was checking everything in the .git directory including all remote content that was local. The previous vendor directory was from a time prior to Go modules when Helm handled dependencies differently. It was no longer needed.

**What this PR does / why we need it**: This is needed to update to the latest k8s packages as they require Go 1.23

**Special notes for your reviewer**: Most of the changes are to handling new linting messages when Go was updated.

**If applicable**:


pkg/kube/wait.go
```diff
@@ -153,7 +153,7 @@ func SelectorsForObject(object runtime.Object) (selector labels.Selector, err er
 	case *batchv1.Job:
 		selector, err = metav1.LabelSelectorAsSelector(t.Spec.Selector)
 	case *corev1.Service:
-		if t.Spec.Selector == nil || len(t.Spec.Selector) == 0 {
```
robertsirc, line 156: Are we ok with removing the check if spec selector is nil? will this break functionality?

robertsirc (approved):
The vast majority of this is to upgrade us to Go 1.23, I do have a question about changes to some conditional if statements.

PR #13531, opened 2024-12-13
v3 backport: Add --take-ownership flag for install and upgrade commands

mayankshah1607:
v3 backport of #13439

See: https://github.com/helm/helm/pull/13439#issue-2652013164

PR #13566, opened 2024-12-21
Fix dev-v3 from take ownership changes

mattfarina:
When #13531 was merged in the tests were not passing. This is because it took into account #13444 but that had not been backported to v3. Tests had not been automatically run on dev-v3 but that was fixed via #13563.

**What this PR does / why we need it**: dev-v3 branch development is failing tests

**If applicable**:


robertsirc (approved):
LGTM

PR #13578, opened 2024-12-28
refactor: Remove ChartRepository `[]ChartPaths`

gjenkins8:
The ChartRepository `[]ChartPaths` member is unused in Helm, and is a holdover from Helm v2. Remove this member and associated functionality

This was desired at one point according to the "Deprecated: remove in Helm 4." comment

**If applicable**:


TerryHowe (approved):
/lgtm

mattfarina (commented):
Looking into this one, I think there is a lot more that can be removed around it. FOr example, this populates the `ChartPaths` property which is only ever used by `generateIndex`. That is only used by `.Index()` which isn't used by anything else.

I think `generateIndex`, `.Index`, and `saveIndexFile` can all be safely removed as the index functions in `index.go` should be used instead. Also, the `ChartPaths` property can be removed... I think.

pkg/repo/chartrepo_test.go
```diff
@@ -40,37 +40,22 @@ const (
 	testURL        = "http://example-charts.com"
 )
 
-func TestLoadChartRepository(t *testing.T) {
```
gjenkins8, line 43: This test I removed, as its purpose appears to be to test the `chart.Load()` function

pkg/repo/chartrepo_test.go
```diff
@@ -40,37 +40,22 @@ const (
...
-	}
-
-	if r.Config.URL != testURL {
-		t.Errorf("Expected url for chart repository to be %s but got %s", testURL, r.Config.URL)
-	}
+	r.ChartPaths = append(r.ChartPaths, globArchives("*.tgz")...)
+	r.ChartPaths = append(r.ChartPaths, globArchives("**/*.tgz")...)
 }
 
 func TestIndex(t *testing.T) {
```
gjenkins8, line 61: `TestIndex` which originally used the `Load(0` function I think needs to remain, as it is testing the `Index()` function (and no other test does)
So it now uses `loadFromDir()` as a test helper

pkg/repo/chartrepo_test.go
```diff
@@ -40,37 +40,22 @@ const (
...
-	r, err := NewChartRepository(&Entry{
-		Name: testRepository,
-		URL:  testURL,
-	}, getter.All(&cli.EnvSettings{}))
-	if err != nil {
-		t.Errorf("Problem creating chart repository from %s: %v", testRepository, err)
-	}
+// loadFromDir a directory of charts archives (including sub-directories),
+// appending to the repositores ChartPath
+func loadFromDir(t *testing.T, r *ChartRepository, dir string) {
```
gjenkins8, line 45: I rewrote and simplified the original `Load()` function to this test helper, as it is still needed for `TestIndex`

TerryHowe (approved):
/lgtm

pkg/repo/chartrepo.go
```diff
@@ -156,46 +118,6 @@ func (r *ChartRepository) DownloadIndexFile() (string, error) {
 	return fname, os.WriteFile(fname, index, 0644)
 }
 
-// Index generates an index for the chart repository and writes an index.yaml file.
-func (r *ChartRepository) Index() error {
```
scottrigby, line 160: My only concern is removing this non-deprecated public function, even if we don't use it internally. Do we have an alternative recommendation for people who may be using it out in the wild?

scottrigby (commented):
one question about `[repo.Index]`

pkg/repo/chartrepo.go
gjenkins8: `(r *ChartRepository) Load() error {` is deprecated, and is the only thing that writes to `ChartPaths()`. So in effect, removing that (deprecated) function, but not this would be quite awkward for users.

mattfarina: `repo.IndexDirectory()` will take a directory and generate the index. This is what the `helm repo index` command uses. The code path being used here isn't used because of the alternative path.

The code path this PR is removing is problematic. For example, right after this `r.generateIndex()` is called. That function uses `loader.Load` which can load a chart as a directory. This is something that should not happen.

I can understand having an internal representation that can generate an index file from a known set of charts. I would want OCI charts included in that. This cannot handle those situations.

I did some searching, using GitHub search, and I did not find anyone using this section of code.

mattfarina (approved):
lgtm

pkg/repo/chartrepo.go
scottrigby: i think this makes sense. we are able to make breaking changes in v4 either way, i just wanted to think it through together. thanks!

PR #13579, opened 2024-12-28
refactor: Remove duplicate `FindChartIn*RepoURL` functions

gjenkins8:
Refactor/cleanup of duplicated `FindChartIn*RepoURL` functions. Replace with option functions.

**If applicable**:

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -789,8 +789,18 @@ func (c *ChartPathOptions) LocateChart(name string, settings *cli.EnvSettings) (
 		dl.Verify = downloader.VerifyAlways
 	}
 	if c.RepoURL != "" {
-		chartURL, err := repo.FindChartInAuthAndTLSAndPassRepoURL(c.RepoURL, c.Username, c.Password, name, version,
-			c.CertFile, c.KeyFile, c.CaFile, c.InsecureSkipTLSverify, c.PassCredentialsAll, getter.All(settings))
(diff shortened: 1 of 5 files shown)


godhanipayal (approved):

pkg/action/install.go
```diff
@@ -785,8 +785,18 @@ func (c *ChartPathOptions) LocateChart(name string, settings *cli.EnvSettings) (
...
+			c.RepoURL,
+			name,
+			version,
+			c.CertFile,
+			c.KeyFile,
+			c.CaFile,
+			getter.All(settings),
+			repo.WithUsernamePassword(c.Username, c.Password),
+			repo.WithInsecureSkipTLSverify(c.InsecureSkipTLSverify),
+			repo.WithPassCredentialsAll(c.PassCredentialsAll),
```
mattfarina, line 798: If options with `.With*` functions are being used for some things, why are they not used for others? I'm thinking about consistency. I like the use of options here.

See https://github.com/helm/helm/blob/0d66425d9a745d8a289b1a5ebb6ccc744436da95/pkg/getter/getter.go#L58 for why I'm thinking about this.

gjenkins8: Loosely, required arguments IMHO should be named. Optional things should be options. 

That said, good spot here. I initially just converted the optional arguments from the function "overloads". but digging further, the client TLS and chart version are also optional.

mattfarina (approved):

PR #13581, opened 2024-12-29
Upgrade golang.org/x/net to v0.33.0 to address CVE-2024-45338

### LINKED ISSUE #13551: CVE-2024-45338
Description:
[CVE-2024-45338](https://github.com/advisories/GHSA-w32m-9786-jp63) in golang.org/x/net which needs an Upgrade package golang.org/x/net to version 0.33.0 or above.

Please help to remediate this CVE.

Output of `helm version`: v3.16.4

ldlb9527:
fixed: #13551


nikvin15:
Thanks for the work. Is there any timeframe we are looking to release this PR ?

mattfarina:
You should see this in the upcoming v3.17.0 release in mid January.

Helm does not directly use the package with the issue. But, it is a transitive dependency of the spdy transport layer for client-go (which Helm uses). If I understand it right, for this to be exploited Helm would need to connect to a compromised Kubernetes API server and the result would be Helm operating more slowly. The exploit, in general, enables DDOS attacks which could impact software built on the Helm SDK. Though, if your API server is compromised then Helm being slow is likely a lower priority issue.

mattfarina (approved):

PR #13598, opened 2025-01-05
refactor: Move chartutil release code to releaseutil

gjenkins8:
A significant amount of code in `chartutil` package is actually related to releases (ie operating on the release of a chart)

To reflect this, move this code from `chartutil` -> `releaseutil`

(I think there would be merit in further refactoring the code in `releaseutil`, but move the code to start)

No functional changes, code move only

**If applicable**:


godhanipayal (approved):
Looks good to me

robertsirc (approved):
LGTM

mattfarina (changes requested):
`chartutil` as a package was designed to deal with a generated chart package from protobuf. This goes back to Helm v2. `releaseutil` as a package was the same thing for a release package that was generated from protobuf. These should be folded into the current `chart` and `release` packages.

This was part of the HIP for refactoring for v2 & v3 charts.

Issue #12122, opened 2023-06-06
Allow calling helm from go


arikmaor:
I would like to embed `helm` inside a cli program that I'm writing
many cli programs, specially ones written using `cobra`, allow me to do that like this:
```go
package tools

import (
	k3d "github.com/k3d-io/k3d/v5/cmd"
	"github.com/spf13/cobra"
	kubectl "k8s.io/kubectl/pkg/cmd"
)

var ToolsCmd = &cobra.Command{
	Use:   "tools",
}

func init() {
	ToolsCmd.AddCommand(k3d.NewCmdK3d())
	ToolsCmd.AddCommand(kubectl.NewDefaultKubectlCommand())
	RootCommand.AddCommand(ToolsCmd)
}
```

When I try to do the same for `helm` I get the flowing error:
```
import "helm.sh/helm/v3/cmd/helm" is a program, not an importable package
```

In order to support this for `helm`, the main file (named `helm.go` in this case) needs to be separated from the rest of the code.

joejulian:
I don't really know the legacy behind this structure, but per [hip-0004](https://github.com/helm/community/blob/main/hips/hip-0004.md) it cannot be changed before Helm 4.

PR #13617, opened 2025-01-09
 Refactor cmd/helm to allow library usage

AustinAbro321:
This is reopening #12725

The [Zarf project](https://github.com/defenseunicorns/zarf) currently ships additional vendored [tools](https://github.com/defenseunicorns/zarf/tree/main/src/cmd/tools) that may be needed in an airgapped environment. Helm doesn't expose the root command and makes use of internal APIs ([version for example](https://github.com/helm/helm/blob/main/cmd/helm/version.go#L27)) so we are a bit stuck at the moment with attempting to vendor a [quasi-fork in tree](https://github.com/defenseunicorns/zarf/tree/main/src/cmd/tools/helm).

It looks like someone else was attempting to do something similar https://github.com/helm/helm/issues/12122.

I do not believe this to be a breaking change as anything that was in cmd/helm was part of a main package and not exported. I think the bigger question is what surface (if any) do you want to expose.

Here's an [example](https://github.com/defenseunicorns/zarf/commit/8cc3aaf26ce5ac07ccfee991e2de60ddfa893b16#diff-5723c1a6985047b35cb44e779cd12650d08f9042a7f942f6a5d97fa2fcebd4b4R25) of what this looks like to be used.

My initial goal here is to find out if y'all are open to a change like this - doesn't have to be in this shape. My approach was to publicly expose the bare minimum to make it possible to embed the root Helm command. Aside from the NewRootCmd this included the debug/warning log functions and the PluginError. Would love a discussion and I'm happy to modify/resubmit as needed.

**If applicable**:
- [x] this PR has been tested for backwards compatibility - refactor, no behavior behavior changes


TerryHowe:
Are you looking for a user experience like https://oras.land/docs/1.3.0-beta.1/how_to_guides/go_script

The associated PR https://github.com/oras-project/oras/pull/825

AustinAbro321:
@TerryHowe Somewhat. The implementation is essentially the same, expose the root command. However, I plan to use the root command so I can embed it into a tool I maintain called Zarf, which simplifies air-gap Kubernetes. We do a similar embedding of Kubectl, see - https://github.com/zarf-dev/zarf/blob/fc183abb20353c8d4ce3ba2ecc16c6190e5dbecf/src/cmd/tools/kubectl.go#L23, which allows us to make kubectl callable using the zarf binary with `zarf tools kubectl`

pkg/cmd/root.go
```diff
@@ -89,7 +90,21 @@ By default, the default directories depend on the Operating System. The defaults
...
+func Debug(format string, v ...interface{}) {
+	if settings.Debug {
+		format = fmt.Sprintf("[debug] %s\n", format)
+		log.Output(2, fmt.Sprintf(format, v...))
+	}
+}
+
+func Warning(format string, v ...interface{}) {
+	format = fmt.Sprintf("WARNING: %s\n", format)
+	fmt.Fprintf(os.Stderr, format, v...)
```
TerryHowe, line 104: Concatenation would be faster

```suggestion
	fmt.Fprintf(os.Stderr, "WARNING: " + format + "\n", v...)
```

pkg/cmd/root.go
```diff
@@ -89,7 +90,21 @@ By default, the default directories depend on the Operating System. The defaults
 | Windows          | %TEMP%\helm               | %APPDATA%\helm                 | %APPDATA%\helm          |
 `
 
-func newRootCmd(actionConfig *action.Configuration, out io.Writer, args []string) (*cobra.Command, error) {
+var settings = cli.New()
+
+func Debug(format string, v ...interface{}) {
+	if settings.Debug {
+		format = fmt.Sprintf("[debug] %s\n", format)
+		log.Output(2, fmt.Sprintf(format, v...))
```
TerryHowe, line 98: I realize this was just copied code, but...

```suggestion
		log.Output(os.Stderr.Fd(), fmt.Sprintf("[debug] " + format + "\n",, v...))
```

TerryHowe (approved):
/lgtm

SECURITY.md
joejulian, line 1: This file should be restored

joejulian (changes requested):
Just need to restore the SECURITY.md. The rest looks good to me.

joejulian (dismissed):
Thanks!

joejulian:
@AustinAbro321 looks like it needs a rebase already.

AustinAbro321:
@joejulian Looks like it's failing from a flake, I'm able to repeat the flake on main #30573

AustinAbro321:
@joejulian @TerryHowe Looks like the flake was fixed, should be good now

joejulian:
Can you squash this, please. That tar file will remain in the data that would get merged.

AustinAbro321:
Done

gjenkins8 (approved):
(no changes to PR since last approval)

joejulian (approved):

TerryHowe:
A quick example in the docs would bring awareness to this feature.

PR #13629, opened 2025-01-13
Rename 'atomic' -> 'rollback-on-failure'

gjenkins8:
Helm implements functionality to revert a failed upgrade/install. Uninstalling the release for installs, rolling back on upgrades. This PR renames this functionality to "rollback-on-failure", rather than today's "atomic".

The reason for the rename, is that "atomic" is misleading (the intermediate states of the operation are very visible to users, and there is no guarantee that the reversion to the original state won't fail) . And using more "plain language" terminology makes the feature more accessible and visible to users.

The `--atomic` CLI flag is retained, but marked as deprecated.

**If applicable**:


pkg/action/upgrade.go
```diff
@@ -91,8 +91,8 @@ type Upgrade struct {
 	Recreate bool
 	// MaxHistory limits the maximum number of revisions saved per release
 	MaxHistory int
-	// Atomic, if true, will roll back on failure.
-	Atomic bool
+	// RollbackOnFailure enables rollingback the upgraded release on failure
```
miledxz, line 94: @gjenkins8 let's update `rollingback` to `rolling back`, if that's fine for you

pkg/action/install.go
```diff
@@ -76,21 +76,22 @@ type Install struct {
...
+	Devel            bool
+	DependencyUpdate bool
+	Timeout          time.Duration
+	Namespace        string
+	ReleaseName      string
+	GenerateName     bool
+	NameTemplate     string
+	Description      string
+	OutputDir        string
+	// RollbackOnFailure enables rollingback (uninstalling) the release on failure if set
```
miledxz, line 93: @gjenkins8 let's update `rollingback` to `rolling back`, if that's fine for you

miledxz (changes requested):

cmd/helm/install.go
```diff
@@ -199,7 +199,9 @@ func addInstallFlags(cmd *cobra.Command, f *pflag.FlagSet, client *action.Instal
 	f.BoolVar(&client.Devel, "devel", false, "use development versions, too. Equivalent to version '>0.0.0-0'. If --version is set, this is ignored")
 	f.BoolVar(&client.DependencyUpdate, "dependency-update", false, "update dependencies if they are missing before installing the chart")
 	f.BoolVar(&client.DisableOpenAPIValidation, "disable-openapi-validation", false, "if set, the installation process will not validate rendered templates against the Kubernetes OpenAPI Schema")
-	f.BoolVar(&client.Atomic, "atomic", false, "if set, the installation process deletes the installation on failure. The --wait flag will be set automatically if --atomic is used")
+	f.BoolVar(&client.RollbackOnFailure, "rollback-on-failure", false, "if set, Helm will rollback (uninstall) the installation on failure. The --wait flag will be set automatically if --rollback-on-failure is used")
```
payalgodhanioracle, line 202: Maybe clarify that Helm will rollback to previous revision (uninstall). Lot of ppl using helm often tends to believe that rollbackOnFailure will rollback to a previous stable revision which is not the case

godhanipayal, line 202: Maybe clarify that Helm will rollback to previous revision (uninstall). Lot of ppl using helm often tends to believe that rollbackOnFailure will rollback to a previous stable revision which is not the case

robertsirc (changes requested):
I think you need to update the `const installDesc` with some clarity there for end users this might address @godhanipayal concerns as well. Other than that the code builds and passes all unit test.

cmd/helm/install.go
gjenkins8: not quite following sorry

I have:

> if set, Helm will rollback (uninstall) the installation on failure.

godhanipayal: My bad, I meant to comment on the upgrade.go change . Rollback on an upgrade with --atomic flag will revert to previous revision as per my understanding. It will do an uninstall if its an install or first revision.

miledxz:
> > I think you need to update the `const installDesc` with some clarity there for end users this might address @godhanipayal concerns as well. Other than that the code builds and passes all unit test.
> 
> I could add some details for `--rolback-on-failure` to `installDesc`. I would prefer to have the description for the `--rollback-on-failure` good enough (keep the description in one place, rather than have them sprawl into multiple places that a user has to find/read.
> 
> (also, `--atomic` doesn't have a mention in `installDesc`. If we think docs are necessary for either `--rollback-on-failure` or `--atomic` (however the flag is spelt), we should perhaps add those regardless ie. a separate PR IMHO)

sounds good to me, @gjenkins8 have you tried to build binary with your changes and test how all works ? 
I will be looking more closely to help out there

TerryHowe (commented):
Just a nit on

pkg/action/install.go
```diff
@@ -76,21 +76,22 @@ type Install struct {
...
+	Devel            bool
+	DependencyUpdate bool
+	Timeout          time.Duration
+	Namespace        string
+	ReleaseName      string
+	GenerateName     bool
+	NameTemplate     string
+	Description      string
+	OutputDir        string
+	// RollbackOnFailure enables rolling back (uninstalling) the release on failure if set
```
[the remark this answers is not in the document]
gjenkins8: yeah, this is copy/pasta from upgrade.go (which does have comments). I figured I could be the first in install.go

TerryHowe (approved):
/lgtm

robertsirc (approved):
LGTM

PR #13655, opened 2025-01-24
feat: support multi-document values files

### LINKED ISSUE #13097: RFE: Support for multi-document values files
Helm support providing multiple values files. Unfortunately, only first YAML document of each file is taken into consideration.

As many values files (i.e. document) are supported, there's no functional reason to not support one values file with many documents. Documents in a single values file should be considered if they were split into many files and just passed in order.

To be more precise giving:

* `A.yaml`

```yaml
---
# A-1.yaml
---
# A-2.yaml
````

* `B.yaml`

```yaml
---
# B-1.yaml
---
# B-2.yaml
````

implies these two equivalent:

* `helm template my-release my-chart -f A.yaml -f B.yaml`
* `helm template my-release my-chart -f A-1.yaml -f A-2.yaml -f B-1.yaml -f B-2.yaml`

### PR DESCRIPTION
**Pull request by LuBingtan**
close #13097 

### DIFF
--- a/pkg/chart/loader/load.go
+++ b/pkg/chart/loader/load.go
@@ -17,14 +17,17 @@ limitations under the License.
 package loader
 
 import (
+	"bufio"
 	"bytes"
 	"encoding/json"
+	"io"
 	"log"
 	"os"
 	"path/filepath"
 	"strings"
 
 	"github.com/pkg/errors"
+	utilyaml "k8s.io/apimachinery/pkg/util/yaml"
 	"sigs.k8s.io/yaml"
 
 	"helm.sh/helm/v4/pkg/chart"
@@ -104,13 +107,11 @@ func LoadFiles(files []*BufferedFile) (*chart.Chart, error) {
 				return c, errors.Wrap(err, "cannot load Chart.lock")
 			}
 		case f.Name == "values.yaml":
-			c.Values = make(map[string]interface{})
-			if err := yaml.Unmarshal(f.Data, &c.Values, func(d *json.Decoder) *json.Decoder {
-				d.UseNumber()
-				return d
-			}); err != nil {
+			values, err := LoadValues(f.Data)
+			if err != nil {
 				return c, errors.Wrap(err, "cannot load values.yaml")
 			}
+			c.Values = values
 		case f.Name == "values.schema.json":
 			c.Schema = f.Data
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by robertsirc** [comment]
Hello, thank you for your pull request. This function doesn't have any associated unit test for this. Is it possible to have that added to ensure we don't break functionality.

**Review by gjenkins8 — CHANGES_REQUESTED** [review]
Along with @robertsirc comment about tests, this only addresses values files specified on the command line. I think for consistency, we should support multi-document values files for both cli overrides, and default chart values.

This also somewhat implies we should unify the value file load process between chart's and the cli.

(it could be argued we do one first, but at least my opinion is we should do both at thg same time for consistency)

**Comment by LuBingtan** [comment]
Thanks for your comments! Let me update

**Comment by LuBingtan** [comment]
Hi @robertsirc @gjenkins8 ,
I've updated accordingly, and let me exaplain briefly what I did here:

1. Added a public function `LoadValues` along with its UT under `pkg/chart/loader/load.go`. This function can be used by both default chart values and the CLI.
2. To facilitate the use of the function `mergeMaps`, I also relocated it and its UT to `pkg/chart/loader/load.go`.

Not sure if this change is optimal. Please take a look, would appreciate your feedback.

File: pkg/chart/loader/load.go [hunk 1]
```diff
@@ -205,3 +206,45 @@ func LoadFiles(files []*BufferedFile) (*chart.Chart, error) {
 
 	return c, nil
 }
+
```
gjenkins8, line 209: Can we add a doc string please

pkg/chart/loader/load.go
```diff
@@ -205,3 +206,45 @@ func LoadFiles(files []*BufferedFile) (*chart.Chart, error) {
...
+func LoadValues(data []byte) (map[string]interface{}, error) {
```
gjenkins8, line 210:
```suggestion
func LoadValues(data io.Reader) (map[string]interface{}, error) {
```

It is preferred to take an interface that can support the necessary functionality, rather than a concrete type

pkg/chart/loader/load_test.go
```diff
@@ -488,6 +489,118 @@ func TestLoadInvalidArchive(t *testing.T) {
 	}
 }
 
+func TestLoadValues(t *testing.T) {
+	testDatas := []struct {
```
gjenkins8, line 493: nit: `testCases` might be better (and more consistent) naming/terminology here

```suggestion
	testCases := []struct {
```

gjenkins8: Personally, I prefer:
```
testCases := map[string]struct {
```

Where the map key is the name of the test. It does make the test ordering non-deterministic (which IMHO is a good things). Up to you if you would also prefer a map here.

gjenkins8 (commented):
Nice, thanks for the update. Close to LGTM, few minor comments

LuBingtan:
Hi @gjenkins8  thanks for your comments, I've updated accordingly, please take a look again

pkg/chart/loader/load.go
```diff
@@ -205,3 +206,46 @@ func LoadFiles(files []*BufferedFile) (*chart.Chart, error) {
...
+			d.UseNumber()
+			return d
+		}); err != nil {
+			return nil, errors.Wrap(err, "cannot unmarshal yaml document")
+		}
+		values = MergeMaps(values, currentMap)
+	}
+	return values, nil
+}
+
```
TerryHowe, line 233: This should technically have a doc string too since it is exported.

LuBingtan: thanks, also updated here

TerryHowe (approved):
/lgtm

pkg/chart/loader/load.go
```diff
@@ -205,3 +206,48 @@ func LoadFiles(files []*BufferedFile) (*chart.Chart, error) {
 
 	return c, nil
 }
+
+// LoadValues loads chat values from a reader.
```
gjenkins8, line 210: typo

And values are not necessary from a chart. They can be either be a chart's default values. Or user supplied values

LuBingtan: thanks, updated here

LuBingtan:
Hi @gjenkins8 would you please take a look again?

gjenkins8 (dismissed):
LGTM

robertsirc (dismissed):
LGTM

robertsirc:
Hi @LuBingtan your branch has some conflicts could you please take a look

gjenkins8 (approved):

sabre1041:
LGTM

PR #13656, opened 2025-01-25
fix: Bind repotest server to `localhost`

gjenkins8:
Currently, `repotest.Server` is binding to all interfaces (empty "host" in the servers "Addr"). This causes the server to also bind to the hosts `0.0.0.0` interface, and as such enables the server to be accessed over the network.

This is perhaps a (very minor) security problem. But annoyingly, on my MacBook I get security popups when running "unit" tests, asking if I want to allow `helm.test` and `registry.test` to "Accept incoming connections?" (see screenshot below).

Instead, bind the server to local interface. Avoiding these popups, as the server is for testing and only needs to accept local connections.

Binding to `127.0.0.1` is fine, see `httptest.NewServer` implementation:
https://cs.opensource.google/go/go/+/refs/tags/go1.23.5:src/net/http/httptest/server.go;l=68

**If applicable**:

image

### DIFF
--- a/pkg/repo/repotest/server.go
+++ b/pkg/repo/repotest/server.go
@@ -113,7 +113,7 @@ func NewOCIServer(t *testing.T, dir string) (*OCIServer, error) {
 		t.Fatalf("error finding free port for test registry")
 	}
 
-	config.HTTP.Addr = fmt.Sprintf(":%d", port)
+	config.HTTP.Addr = fmt.Sprintf("127.0.0.1:%d", port)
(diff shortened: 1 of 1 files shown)


TerryHowe (approved):
/lgtm

godhanipayal (approved):
LGTM

miledxz:
LGTM

robertsirc (approved):
LGTM

yxxhero (approved):
LGTM

PR #13657, opened 2025-01-26
chore: remove unused pkg/gates

gjenkins8:
`pkg/gates` is not used within Helm codebase. Remove it.

**If applicable**:


TerryHowe:
Is there another mechanism for this?

TerryHowe:
> I figure if it is unused, it can be removed. Alternatively, we could move it to be an `internal` package if we anticipated future use? (or the functionality could just be resurrected at that point)

Sounds good.

TerryHowe (approved):
/lgtm

mattfarina (changes requested):
This package is used for gating experimental features. There currently no experimental features which is why you don't see users.

When we do the next APIVersion of charts I would like to use this until that's GA (after 4.0.0 is out in my estimation). Can we keep this around but maybe instead update the docs on it?

PR #19078, opened 2025-02-02
Iterate over dependencies in a chart tarball during lint action

### LINKED ISSUE #12832: helm lint does not lint subcharts if helm package is .tgz or tar.gz
A helm chart with subscharts can lint subcharts by using the command switch --with-subcharts. 
This does not work if the helm-chart with subscharts are in gzipped tar format.

$ helm lint --with-subcharts mychart
==> Linting mychart

==> Linting mychart/charts/myimport

==> Linting mychart/charts/mysubchart
[ERROR] Chart.yaml: apiVersion 'v0' is not valid. The value must be either "v1" or "v2"
[ERROR] Chart.yaml: chart type is not valid in apiVersion 'v0'. It is valid in apiVersion 'v2'

Error: 3 chart(s) linted, 1 chart(s) failed

$ helm lint --with-subcharts mychart.tgz
==> Linting mychart.tgz

Output of `helm version`: version.BuildInfo{Version:"v3.14.2", GitCommit:"c309b6f0ff63856811846ce18f3bdc93d2b4d54b", GitTreeState:"clean", GoVersion:"go1.21.7"}

manslaughter03:

Load chart when lint a chart tarball, save into a temp directory and add it to path to lint.

closes #12832 

**If applicable**:

### DIFF
--- a/cmd/helm/lint.go
+++ b/cmd/helm/lint.go
@@ -27,6 +27,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"helm.sh/helm/v4/pkg/action"
+	"helm.sh/helm/v4/pkg/chart/loader"
 	"helm.sh/helm/v4/pkg/chartutil"
 	"helm.sh/helm/v4/pkg/cli/values"
 	"helm.sh/helm/v4/pkg/getter"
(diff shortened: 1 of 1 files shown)


cmd/helm/lint.go
```diff
@@ -67,16 +68,35 @@ func newLintCmd(out io.Writer) *cobra.Command {
 
 			if client.WithSubcharts {
 				for _, p := range paths {
-					filepath.Walk(filepath.Join(p, "charts"), func(path string, info os.FileInfo, _ error) error {
-						if info != nil {
-							if info.Name() == "Chart.yaml" {
-								paths = append(paths, filepath.Dir(path))
-							} else if strings.HasSuffix(path, ".tgz") || strings.HasSuffix(path, ".tar.gz") {
-								paths = append(paths, path)
+					if strings.HasSuffix(p, ".tgz") || strings.HasSuffix(p, ".tar.gz") {
```
godhanipayal, line 71: It would be better if this is optional. If there are more dependencies, it will increase the total time of linting helm chart. Also, could you add a supporting test case ?

manslaughter03: Hi, thanks for your feedback.

As this iteration is already done when `--with-subcharts` option is activate i didn't think that would be necessary because user already ask to lint subcharts ?

For the test case i was stuck on strict equal done in `cmd/helm/lint_test.go::TestLintCmdWithSubchartsFlag` as in output path will be dynamic cause of  `os.MkdirTemp("", "helm-lint")`.

I'm going to re-think how to test it.

PR #30296, opened 2025-02-07
[v3 backport] Oras v2 v3 backport with http fallback

TerryHowe:

Update oras v2 which removes many unwanted dependencies.

This PR has two commits. The first is the commit from the [helm v4 PR ](https://github.com/helm/helm/pull/13382) and the second is some code to automatically fallback to http which was added to not break current v3 behavior.

**If applicable**:


godhanipayal (approved):
/ lgtm

connyay:
We (@tanium) are hitting some pains where our internal registry is returning 500s randomly when fetching oauth tokens and this PR would dramatically improve our lives with `retry.NewTransport(transport)`. Would love to see this land 👍 

Thank you @TerryHowe 👏

sabre1041:
@TerryHowe would you be able to resolve the conflicts?

TerryHowe:
Conflicts resolved

sabre1041 (approved):
LGTM

pkg/registry/fallback.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// RoundTrip wraps base round trip with conditional insecure retry.
+func (t *fallbackTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	if ok := t.forceHTTP.Load(); ok {
+		req.URL.Scheme = "http"
+		return t.Base.RoundTrip(req)
+	}
+	resp, err := t.Base.RoundTrip(req)
+	if err != nil && req.URL.Scheme == "https" {
```
gjenkins8, line 61: Can we reference why downgrading to http is done here?

(I can speculate ORAS v1 did this, but since it is quite unusual, I think worth linking to why Helm would do this / where in ORAS v1 this behavior existed)

Just a reference in a comment here will be fine. Can PR in to source comment in a followup (try and get this merged without multiple approval cycles)

TerryHowe: Sure thing. 

Yes, this was only added for compatibility.

TerryHowe:
```suggestion
	// We are falling back to http here for backward compatibility with Helm v3.
	// ORAS v1 provided fallback automatically, but ORAS v2 does not.
	if err != nil && req.URL.Scheme == "https" {
```

TerryHowe:
Closes: https://github.com/helm/helm/issues/30667

scottrigby (approved):
Thanks for putting in the work for this backport. Important for v3 maintainability. Looks great!

gjenkins8 (approved):

TerryHowe:
conflicts resolved again

Issue #30507, opened 2025-02-13
Problem with public key verification on latest 3.17.1 release


eimhin-rover:
Hello! We are currently having issues installing helm via the [kubectl-helm-minikube devcontainer feature](https://github.com/devcontainers/features/tree/main/src/kubectl-helm-minikube) as reported in [this issue](https://github.com/devcontainers/features/issues/1272) ever since the latest release. 

```
Downloading Helm...
HELM_VERSION=3.17.1
(*) Keyserver hkp://keyserver.pgp.com is not reachable.
Verification failed!
gpg: assuming signed data in '/tmp/helm/helm-v3.17.1-linux-arm64.tar.gz'
gpg: Signature made Wed Feb 12 20:57:29 2025 UTC
gpg:                using RSA key 748CE9B6B2EDF4D2033318CA07A0310EB5B9076F
gpg: Can't check signature: No public key
```

Would anyone be able to provide some insight into what might be wrong here? Perhaps the wrong pubilc key was pushed with the latest release? Thanks!

TerryHowe:
It could be @robertsirc needs to add his key https://keyserver.pgp.com/vkd/GetUploadKeyScreen.event

robertsirc:
This could be the case. Let me and @mattfarina take a look.

kewne:
Yeah, the script that the devcontainers feature uses tries to verify the signature from the 3.17.1 release and fails because of that missing key.

I also checked the [`get-helm3.sh` script](https://helm.sh/docs/intro/install/#from-script), which goes through mostly the same steps but succeeds because it gets 3.17.0 as the latest instead.

eimhin-rover:
I'm also not sure where the `get-helm` script is getting that `3.17.0` When I look at the response of the URL being used to grab the latest version https://get.helm.sh/helm-latest-version it returns `v3.17.1`.

Either way, initially I thought there was an issue in how the devcontainer feature was configured as when I run the `get-helm` script with the version explicitly set to `3.17.1` it works. However as it turns out the reason it works is because by default, this `get-helm` script [does not actually verify signatures by default](https://github.com/helm/helm/blob/3a94215585b91d5ac41ebb258e376aa11980b564/scripts/get-helm-3#L24). You have to explicitly set an environment variable to turn it on `VERIFY_SIGNATURES=true` and doing so allows you to reproduce the same issue as is seen in the devcontainer feature:
```
[...]

PR #30508, opened 2025-02-13
Update version option description with more accurate info

eimhin-rover:

Follow up from https://github.com/helm/helm/issues/30507. This PR updates the version option description to make it clearer that the latest version tag is fetched from the Helm CDN as opposed from Github releases.

This should ideally make it clearer if in future, there are issues where the latest version does not seem to be updating and is more accurate for the end user.

**If applicable**:


miledxz:
Hi @eimhin-rover, thank you for the PR, from my side this looks good and useful, let's see what @scottrigby thinks ?

robertsirc (approved):
LGTM

PR #30512, opened 2025-02-14
actually enforce semantic versioning in chart Metadata

canardleteer:

`helm`, currently does not enforce Semantic Versioning, although it claims to. In the current `helm`, you can set at least 2 styles of invalid versions, explicitly [forbidden under SemVer 2.0.0 §2](https://semver.org/#spec-item-2).

`Chart.yaml` can include the following `version:`

- `version: 1`
- `version: 1.0`

What was happening, is that the `semver.NewVersion()` library was "trying to be helpful", and upgrading the Version being passed into it (example: `1` was being returned as `1.0.0`), but this mutation never triggered an `Error` and was silently dropped by `isValidSemver` to return a boolean result. `semver.StrictNewVersion()` actually enforces what is needed. The "fixed" Version was never passed forward into the `Metadata`, so charts with `1` can be `lint`ed, `package`'d, and deployed without error.

The tests themselves, had incorrect Semantic Versioning as well, so needed to be fixed up.

I expect this will actually introduce some issues for people who have not been using Semantic Versioning, as explicitly outlined in the comments for Chart Metadata YAML for them to do. If it's "backwards compatible" or not, is a bit like asking if we should preserve the bug or not to avoid breaking people's workarounds.

An alternative is to mutate the Metadata during validation, but then we're just passing forward the dangerously "helpful behavior" of allowing people to sidestep Semantic Versioning. I don't think that's the right behavior to encourage.

The documentation was saying to do the right thing, the code was just not enforcing it correctly, , so no documentation changes needed.

It's been awhile since I've put on my Golang hat, not going to be surprised if I did some of this wrong :)

**If applicable**:


canardleteer:
Found some more failing tests, because they were authored without valid Semantic Versioning. Still truing these up.

canardleteer:
Fixed the various issues for testing this 1 change.

Continuing along in another branch for other cases where Semantic Versioning is not enforced, since that commit will be much larger.

TerryHowe (commented):
Was this a desired breaking change for Helm 4?

mattfarina (changes requested):
I don't have complete background on why Helm does the "loose" handling for versions. While I'm the author of the semver package, I wasn't the person who brought it into Helm. When I refer to "loose", I'm referring to the term as nodejs semver uses it.

I have some good news and some bad news. I'll start with the bad.

With Helm, we work hard for backwards compatibility. If we severely break users, they will fork Helm, revolt, give us lots of support work, and more. We're currently working on Helm v4 and I've had numerous people talk to me about not breaking their charts. So, we don't want to make this change in the existing charts.

There are tens of thousands of charts, many of which are private. There are millions of installations of charts every month (this can be easily reasoned and we used to have some statistics). We don't want to break this ecosystem.

The good new is, we are looking to create a new version of charts. [The Helm Improvement Proposal is here](https://github.com/helm/community/pull/381). I would really like to see this kind of change come into Helm for this next version of the charts. We're just not quire ready to do this yet.

canardleteer:
@TerryHowe - Personally, I'd like to see the bug fixed in 3+, my change targets 4, but I'd be happy to port it to a 3 branch.

@mattfarina - The backwards compatibility concern is a real one that will be disruptive (I have no doubt about that, all those concerns are legitimate), but I don't know how effective the HIP process will be at fixing it as a bug. The currently documented requirement is that users use SemVer 2, with a link to the spec, both in [the template](https://github.com/helm/helm/blob/main/pkg/chartutil/create.go#L93-L96) provided by `helm create` as well as the [online documentation](https://helm.sh/docs/chart_best_practices/conventions/#version-numbers). Anyone interfacing with charts, is lead to believe that these charts are versioned within the SemVer 2 rules. All that would be said in an HIP, with today's context, is what is already said in the documentation & code today.

An alternative approach, would be to document the "loose semver dialect" currently in place, across all relevant public locations, and then decide to require SemVer going forward as an improvement, or just maintaining a "loose semver dialect" specification as an unfortunate exception.

At the moment, people don't know what to expect from chart versions, since what is documented, is inaccurate.

TerryHowe:
You got a hard no on 4, so I don't think this will fly for Helm 3.

banjoh:
> An alternative approach, would be to document the "loose semver dialect" currently in place, across all relevant public locations, and then decide to require SemVer going forward as an improvement (via HIP), or just maintaining a "loose semver dialect" specification as an unfortunate exception.

This will most likely be the case as it stands for Helm v3. For charts v3, this could be made stricter. Migration to/creation of a chart v3 would require having a correct SemVer 2 version string.

canardleteer:
@gjenkins8 / @mattfarina - Just to make sure I'm capturing what you two are suggesting (which I think is the happy path forward):

- chart schema v2 (in helm v3 & v4): Lint Warning + Potential Documentation Updates/Warnings.
- chart schema v3: Strict Compliance.
  - Would need to be introduced as a HIP.

mattfarina:
The HIP just landed to enable a new chart apiVersion. As soon as that's in place, we can do this work on that new apiVersion for charts.

To break a bunch of end-users isn't something we can really do without consequences we don't want to take on, as maintainers.

canardleteer:
@gjenkins8, added a comment over in https://github.com/helm/helm/issues/30513, and will close this PR.

PR #30552, opened 2025-02-19
v3 backport: feat: Enable CPU and memory profiling

scottrigby:
Helm v3 backport of #13481 

Cherry-picked commits, code still applies cleanly to `dev-v3`, and works as expected.

**If applicable**:


banjoh (approved):

cmd/helm/profiling.go
```diff
@@ -0,0 +1,76 @@
...
+			errs = append(errs, err.Error())
+		}
+		cpuProfileFile = nil
+	}
+
+	if memprofile != "" {
+		f, err := os.Create(memprofile)
+		if err != nil {
+			errs = append(errs, err.Error())
+		}
```
godhanipayal, line 77: Maybe add a return stmt here :

f, err := os.Create(memprofile)
if err != nil {
    return fmt.Errorf("could not create memory profile file: %w", err)
}
defer f.Close()

If there is an issue in creating a file, f will be nil and calling f.Close() might result in a runtime error

banjoh: Good catch. Luckily, golang thought of people like me who at times forget to handle such cases, and got us covered. [f.Close() will check if f == nil before attempting to close the file](https://github.com/golang/go/blob/64d82cd72c222caa93b2f71c5970a00ec4e7929a/src/os/file_posix.go#L19)

banjoh:
> Is there another v3 release coming that we would want to merge this into?

According to [Feb 6th developer call meeting notes](https://docs.google.com/document/d/1d-6xJEx0C78csIYSPKJzRPeWaHG_8W1Hjl72OJggwdc/edit?tab=t.0#heading=h.ladoikd5kxw7) we should expect one last 3.18 release

godhanipayal (approved):
lgtm

gjenkins8 (approved):

scottrigby:
> > Is there another v3 release coming that we would want to merge this into?
> 
> According to [Feb 6th developer call meeting notes](https://docs.google.com/document/d/1d-6xJEx0C78csIYSPKJzRPeWaHG_8W1Hjl72OJggwdc/edit?tab=t.0#heading=h.ladoikd5kxw7) we should expect one last 3.18 release

Just letting folks know to help manage expectations, while there will be more minor releases for helm 3 (for k8s package updates), 3.18.0 is planned to be the last one that includes new features for Helm 3. This is at the discretion of the Helm 4 release engineer per the ["Maintaining Helm 3" section of the Helm 4 development process HIP](https://github.com/helm/community/blob/main/hips/hip-0012.md#maintaining-helm-3).

I spoke with Matt Farina about this (Helm v4 release engineer), and he is HESITANTLY ok with us doing v3 backports into the upcoming `3.18.0` release, only if:

1. it doesn't take too much maintainer time away from Helm 4 work and PR review
2. related to # 1, the code in question hasn't diverged much between `main` and `dev-v3` yet (the cherry-picked commits apply cleanly or with minimal conflict resolution)
3. we ensure breaking changes aren't introduced in a v3 backport (because of course breaking changes that may have been OK for a Helm 4 PR are not OK for Helm 3)

If these conditions aren't met for a potential backport for a PR, we will need to say no. We should also not encourage folks to do a lot of this, mainly so that we don't give ourselves more that we need to manage with all the Helm 4 work that needs to be done, and also to not encourage disappointment when we have to say no.

For this PR, it took me only a few mins to backport, cherry-picked commits still applied cleanly, test still pass, PR still works as expected, and the reviewing maintainers had just reviewed the same commits in the main PR so was a low cognitive load. There may be a few other backportable features like this too, but as Helm 4 progresses between now and the upcoming `3.18.0` release, there will be more and more divergence in the codebase making backports more difficult and time involved to review, and therefore less likely to get in.

PR #30580, opened 2025-02-24
Move pkg/releaseutil to pkg/release/util

mattfarina:
The releaseutil package was originally designed to work against a generated codebase from a protobuf in Helm v2. This is when Helm used gRPC to communicate to a server side component named Tiller. When Helm moved everything client side, this package remained and it supported the release package.

This change moves releaseutil to be a sub-packge of release. This is part of the change to support apiVersion v3 charts which is documented in HIP 20

**If applicable**:


robertsirc (approved):
LGTM

PR #30589, opened 2025-02-26
Move pkg/release to pkg/release/v1 to support v3 charts

mattfarina:
This is part of HIP 20 which provides a means to have v3 charts that live alongside v2 charts while having breaking changes.

The plan is to have a different release object for v3 chart instances for at least a couple reasons:
1. So that the chart object on the release can be fundamentally different.
2. So that Helm v3 does not detect or try to work with instances of charts whose apiVersion it does not know about.

Note: it is expected that Helm v3 usage will be used long after the Helm project no longer supports it. 5 years after Helm v2 had reached end-of-life there was still usage of it.

Note: The release util package is separate from the versioned elements as it is planned to use generics to handle multiple release object versions.

**If applicable**:


robertsirc (approved):
LGTM

PR #30592, opened 2025-02-26
changing from log to slog

robertsirc:
This is my first pass at replacing `log` with `slog`.

### DIFF
--- a/pkg/repo/index.go
+++ b/pkg/repo/index.go
@@ -19,7 +19,7 @@ package repo
 import (
 	"bytes"
 	"encoding/json"
-	"log"
+	"log/slog"
 	"os"
 	"path"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


pkg/repo/index.go
```diff
@@ -356,7 +356,7 @@ func loadIndex(data []byte, source string) (*IndexFile, error) {
 	for name, cvs := range i.Entries {
 		for idx := len(cvs) - 1; idx >= 0; idx-- {
 			if cvs[idx] == nil {
-				log.Printf("skipping loading invalid entry for chart %q from %s: empty entry", name, source)
+				slog.Info("skipping loading invalid entry for chart %q from %s: empty entry", name, source)
```
TerryHowe, line 359: Why was this one an info? "invalid entry" sounds like an error

robertsirc: This was a 1 for 1 replacement we can bring this up to an error if it is warranted. I plan on discussing this in our developer meeting.

robertsirc: fixed this.

TerryHowe (approved):
/lgtm

mattfarina (approved):

PR #30593, opened 2025-02-26
EXAMPLE: Updating checkIfInstallable to handle multiple chart apiVersions

mattfarina:
This pull request is an example of the types of changes needed to be made to handle multiple apiVersions for charts. This is opened as a draft for discussion purposes as this is a small change that would be replicated many places throughout the code.


TerryHowe:
I prefer this approach

PR #30608, opened 2025-02-28
fix: update repo_list.go (Closes 30606!)

### LINKED ISSUE #30606: No repository is not an error,use the helm repo list command ,if there is no repository,it should not be an error
helm:v3.17.1
1.No repository is in my local computer. After using the helm repo list command, it prompts "Error: no repositories to show"

Image
2.Expected Results  After using the helm repo list command, it prompts "no repositories to show"

pipi1111111:
我的本地电脑没有仓库，
使用helm repo list命令后提示“Error: no repositories to show”使用helm repo list命令后提示“no repositories to show”

fixes: #30606

### DIFF
--- a/pkg/cmd/repo_list.go
+++ b/pkg/cmd/repo_list.go
@@ -21,7 +21,6 @@ import (
 	"io"
 
 	"github.com/gosuri/uitable"
-	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 
 	"helm.sh/helm/v4/pkg/cli/output"
@@ -39,10 +38,6 @@ func newRepoListCmd(out io.Writer) *cobra.Command {
 		ValidArgsFunction: noMoreArgsCompFunc,
 		RunE: func(_ *cobra.Command, _ []string) error {
 			f, _ := repo.LoadFile(settings.RepositoryConfig)
-			if len(f.Repositories) == 0 && !(outfmt == output.JSON || outfmt == output.YAML) {
-				return errors.New("no repositories to show")
-			}
-
 			return outfmt.Write(out, &repoListWriter{f.Repositories})
 		},
 	}
@@ -62,6 +57,10 @@ type repoListWriter struct {
 }
 
 func (r *repoListWriter) WriteTable(out io.Writer) error {
+	if len(r.repos) == 0 {
+		_, _ = fmt.Fprintln(out, "no repositories to show")
+		return nil
+	}
 	table := uitable.New()
 	table.AddRow("NAME", "URL")
 	for _, re := range r.repos {
--- a/pkg/cmd/repo_list.go
+++ b/pkg/cmd/repo_list.go
@@ -37,6 +37,7 @@ func newRepoListCmd(out io.Writer) *cobra.Command {
 		Args:              require.NoArgs,
 		ValidArgsFunction: noMoreArgsCompFunc,
 		RunE: func(_ *cobra.Command, _ []string) error {
+			f, _ := repo.LoadFile(settings.RepositoryConfig)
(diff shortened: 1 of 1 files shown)


TerryHowe (commented):
I am not a maintainer, but if there is going to be a breaking change like this, I'd vote for silently ignore or `Debug("no repositories to show")`

mattfarina:
For some historical context, the reason we did not change this sooner is because we did not want to break compatibility in Helm v3 when it came to return codes. This one has long bugged me. No repositories should not be an error.

mattfarina (changes requested):
Thanks for coming after this bit of functionality. I have wanted to see this changed for a long time. With Helm v4 we can now do this. I have one minor request.

pipi1111111 (commented):
I have adopted your suggestion and moved the logic for handling empty repositories from repoListWriter to the RunE function

TerryHowe (approved):
/lgtm

pkg/cmd/repo_list.go
```diff
@@ -39,16 +38,19 @@ func newRepoListCmd(out io.Writer) *cobra.Command {
...
-			if len(f.Repositories) == 0 && !(outfmt == output.JSON || outfmt == output.YAML) {
-				return errors.New("no repositories to show")
+			if len(f.Repositories) == 0 {
+			    switch outfmt {
+			    case output.Table:
+				_, _ = fmt.Fprintln(out, "no repositories to show")
+				return nil
+			     case output.JSON, output.YAML:
+				return outfmt.Write(out, &repoListWriter{repos: []*repo.Entry{}})
+			     }
```
TerryHowe, line 48:
```suggestion
			if len(f.Repositories) == 0 {
				switch outfmt {
				case output.Table:
					_, _ = fmt.Fprintln(out, "no repositories to show")
					return nil
				case output.JSON, output.YAML:
					return outfmt.Write(out, &repoListWriter{repos: []*repo.Entry{}})
				}
```

pkg/cmd/repo_list.go
```diff
@@ -39,16 +38,19 @@ func newRepoListCmd(out io.Writer) *cobra.Command {
...
-			if len(f.Repositories) == 0 && !(outfmt == output.JSON || outfmt == output.YAML) {
-				return errors.New("no repositories to show")
+			if len(f.Repositories) == 0 {
+				switch outfmt {
+				case output.Table:
+					_, _ = fmt.Fprintln(out, "no repositories to show")
+					return nil
+				case output.JSON, output.YAML:
+					return outfmt.Write(out, &repoListWriter{repos: []*repo.Entry{}})
+				}
```
gjenkins8, line 48: The empty repositories + JSON or YAML case is already handled by line 50. There is no need to handle it specially here I think. I think all that is needed for this PR is:

```suggestion
			if len(f.Repositories) == 0 && !(outfmt == output.JSON || outfmt == output.YAML) {
				_, _ = fmt.Fprintln(out, "no repositories to show")
					return nil
```

pkg/cmd/repo_list.go
```diff
@@ -39,16 +38,19 @@ func newRepoListCmd(out io.Writer) *cobra.Command {
...
 			}
-
 			return outfmt.Write(out, &repoListWriter{f.Repositories})
 		},
 	}
-
```
gjenkins8, line 49: Can we revert these unrelated (and unintended?) whitespace changes please.

(it is bad practice to include multiple unrelated changes into a single PR: it often makes the PR harder to review/reviewer needs to juggle multiple concerns, and commingles the intent of the different changes)

mattfarina:
Closing as an alternative fix was merged.

Issue #30573, opened 2025-02-22
Test flake on TestDedupeRepos


AustinAbro321:
Reproducible steps - 
1. clone main
2. cd `pkg/downloads` 
3. `go test -run TestDedupeRepos -count=1` <- this should pass
4. `go test -run TestDedupeRepos -count=100` <- this fails pass

Seeing this flake on PR #13617, I have not looked into the cause at all

```go
--- FAIL: TestDedupeRepos (0.00s)
    --- FAIL: TestDedupeRepos/no_duplicates (0.00s)
        manager_test.go:661: received:
            [{"name":"","url":"https://example.com/charts2","username":"","password":"","certFile":"","keyFile":"","caFile":"","insecure_skip_tls_verify":false,"pass_credentials_all":false} {"name":"","url":"https://example.com/charts","username":"","password":"","certFile":"","keyFile":"","caFile":"","insecure_skip_tls_verify":false,"pass_credentials_all":false}]
            want:
            [{"name":"","url":"https://example.com/charts","username":"","password":"","certFile":"","keyFile":"","caFile":"","insecure_skip_tls_verify":false,"pass_credentials_all":false} {"name":"","url":"https://example.com/charts2","username":"","password":"","certFile":"","keyFile":"","caFile":"","insecure_skip_tls_verify":false,"pass_credentials_all":false}]
---... happens more...---
--- FAIL: TestDedupeRepos (0.00s)
    --- FAIL: TestDedupeRepos/no_duplicates (0.00s)
        manager_test.go:661: received:
           <same-output-again>
FAIL
exit status 1
```

yxxhero:
@AustinAbro321 is still fail when set count=2?

gjenkins8:
This test was introduced (recently): https://github.com/helm/helm/pull/11112/files

gjenkins8:
The test is assuming the results from the `dedupeRepos` function are deterministically ordered from this map iteration: https://github.com/helm/helm/pull/11112/files#diff-8157c4e80f3a8abb1dd88a83411502f7d02b91e112df1a6a7c25b176d3d18d71R671

@felipecrs

felipecrs:
@gjenkins8, alright. I'll push a PR today.

felipecrs:
@gjenkins8 please take a look at https://github.com/helm/helm/pull/30576.

PR #30610, opened 2025-03-02
feat: Enable shuffle for tests

gjenkins8:
Add `-shuffle=on` to test execution. This causes tests to be executed in a non-deterministic order, and with "unique" randomness (seeded from system time).

The should help prevent issues like: https://github.com/helm/helm/issues/30573, where likely the author's testing and/or PR checks will manage to execute and fail tests that have some sort of reliance on deterministic test execution.

Ideally we might add `-count=3` (or higher). Deferring that to another PR.

**If applicable**:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ ACCEPTANCE_RUN_TESTS=.
 PKG         := ./...
 TAGS        :=
 TESTS       := .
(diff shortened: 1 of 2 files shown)


godhanipayal:
TestInstallRelease_Wait_Interrupted  seems to be failing with this PR

PR #8499, opened 2020-07-22
support `--untar` flag for dep update

Cweiping:
**What this PR does / why we need it**:
support `--untar` flag for dep update,and use it  what untar `*.tgz` to `./charts/`. At the same time, the local tgz package will be deleted

fix issue:https://github.com/helm/helm/issues/8479

**Special notes for your reviewer**:

**If applicable**:
- [ ] this PR contains documentation
- [x] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

PR #30616, opened 2025-03-03
Add untar option to dependency update

### LINKED ISSUE #8479: Feature request: `helm dependency update --untar`
This is about adding the `--untar` flag to `helm dependency update`, the same way it exists for `helm pull`, so it would:

1. Download the dependencies just as before
2. Unpack them
3. Delete the .tgz files

## References

https://github.com/helm/helm/issues/2247#issuecomment-293991937

Shaps:
Add an option to untar the dependencies after downloading. 

It's mostly a copy and resurrection of #8499 and as that, it also fixes #8479

I am more than happy to make more changes where required. ( I am thinking the first check box should be selected, but I guess it can be done at a later stage. ) 

**If applicable**:

### DIFF
--- a/pkg/cmd/dependency_update_test.go
+++ b/pkg/cmd/dependency_update_test.go
@@ -254,6 +254,7 @@ func TestDependencyUpdateCmd_WithRepoThatWasNotAdded(t *testing.T) {
 		fmt.Sprintf("dependency update '%s' --repository-config %s --repository-cache %s --content-cache %s", dir(chartname),
 			dir("repositories.yaml"), dir(), contentCache),
 	)
+
 	if err != nil {
 		t.Logf("Output: %s", out)
 		t.Fatal(err)
--- a/pkg/cmd/dependency_update_test.go
+++ b/pkg/cmd/dependency_update_test.go
@@ -149,6 +149,19 @@ func TestDependencyUpdateCmd(t *testing.T) {
 	if _, err := os.Stat(expect); err != nil {
 		t.Fatal(err)
 	}
+
+	// When using `--untar`, ./charts/*.tgz should not exist
+	_, _, err = executeActionCommand(
+		fmt.Sprintf("dependency update '%s' --repository-config %s --repository-cache %s --untar", dir(chartname), dir("repositories.yaml"), dir()),
+	)
+	if err != nil {
+		t.Fatal(err)
+	}
+	// Make sure the actual file is downloaded and untar
+	expect = dir(chartname, "charts/reqtest")
+	if _, err := os.Stat(expect); err != nil {
+		t.Fatal(err)
+	}
 }
 
 func TestDependencyUpdateCmd_DoNotDeleteOldChartsOnError(t *testing.T) {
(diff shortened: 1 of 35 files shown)


pkg/cmd/dependency_update_test.go
```diff
@@ -234,7 +247,6 @@ func TestDependencyUpdateCmd_WithRepoThatWasNotAdded(t *testing.T) {
 		fmt.Sprintf("dependency update '%s' --repository-config %s --repository-cache %s", dir(chartname),
 			dir("repositories.yaml"), dir()),
 	)
-
```
felipecrs, line 237: I'd recommend undoing these cosmetic changes

Shaps: yes, sorry, editor autoformat, undoing now

OzBenSimhonTraceloop (commented):
Any updates here guys? 🙏

pkg/downloader/manager.go
```diff
@@ -357,6 +358,10 @@ func (m *Manager) downloadAll(deps []*chart.Dependency) error {
 			break
 		}
 
+		if m.Untar {
+			chartutil.ExpandFile(m.ChartPath+"/charts/", tmpPath)
```
benoittgt: Should we check for errors? 
Something like
```go
    if err := chartutil.ExpandFile(m.ChartPath+"/charts/", tmpPath); err != nil {
        return fmt.Errorf("failed to expand chart %s: %w", tmpPath, err)
    }
```

Also what about cleaning after as suggested in https://github.com/helm/helm/issues/8479#issue-659655334.

benoittgt (commented):
I’m wondering how closely we should aim to match the behavior of helm pull.

helm pull:
	•	Downloads to a temp directory when --untar is used
	•	Extracts to the final destination only after verification
	•	Automatically cleans up temp files
	•	Performs conflict detection before extraction
	•	Supports a custom untar directory via --untardir

helm dependendy update (patched):
	•	Downloads directly to the final location
	•	Leaves both the .tgz and the extracted directory (potential duplication)
	•	No conflict detection
	•	No cleanup of the original .tgz file
	•	Fixed extraction location (charts/ directory) — which I think is fine

pkg/downloader/manager.go
Shaps: wow. Completely lost track of this PR :laughing:.
I'm not sure what you mean when you say cleaning after as suggested in the comment. I can definitely add/implement with some guidance. 
I am guessing you mean create a new tmpdir and then adding the `os.RemoveAll`, but just want to be 100% sure.

Thanks!

Shaps:
@benoittgt I think having it as close as possible to `helm pull` probably makes sense.

I only really needed to add `untar` and as it seemed a popular enough ask I started working on that without looking too much into feature parity. 

I am more than happy to add the missing bits, assuming I can use `pull` as a reference for the missing features

felipecrs:
What would happen if a chart depends on two different versions of the same chart, what would the extracted folder name be like?

felipecrs:
> What would happen if a chart depends on two different versions of the same chart, what would the extracted folder name be like?

I believe the correct answer is to save the charts in a directory named after the aliases (if they have aliases set in Chart.yaml).

Shaps:
I"m going to close this PR and reopen a new one with a fresh branch from main since I see the branches have diverged ( + I messed up the rebase/conflicts :laughing: ).

PR #30618, opened 2025-03-03
Fix namespace flag not registering

### LINKED ISSUE #30619: Helm 4 namespace flag does not work
The namespace flag is not being properly passed through as of [13617](https://github.com/helm/helm/pull/13617). 

Reproducible steps:
1. clone helm and cd into the directory
2. `make`
3. `helm create foo `
4. `/bin/helm install my-chart foo -n bar --create-namespace`

This will install the chart to default rather than the given namespace. This can be verified with `kubectl get pods`

AustinAbro321:
In #13617 a bug was introduced that caused the `--namespace` flag to not be passed through the CLI correctly. This fixes that bug, by moving more into the pkg/cmd layer. 

Fixes: #30619

**If applicable**:

### DIFF
--- a/cmd/helm/helm.go
+++ b/cmd/helm/helm.go
@@ -23,7 +23,6 @@ import (
 	// Import to initialize client auth plugins.
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 
-	"helm.sh/helm/v4/pkg/action"
 	helmcmd "helm.sh/helm/v4/pkg/cmd"
 	"helm.sh/helm/v4/pkg/kube"
 )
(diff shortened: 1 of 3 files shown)


scottrigby (approved):
Fixes the bug, which is reproducible manually (not with our tests).

This is in line with refactor in https://github.com/helm/helm/pull/13617 and was just missed. LGTM

mattfarina (approved):
I tested and this fixes the problem.

Looking over the change, this makes sense. I see what's going on.

mattfarina:
@scottrigby non-trivial changes often require some manual testing to verify.

Issue #3486, opened 2018-02-09
Helm should preform deep merge on multiple values files


connormckelvey:
It looks like #1620 was supposed to provide this functionality but I am seeing that a list in the last specified values file completely override the list from the previous file

I have 2 additional values files foo.yaml, and bar.yaml as follows:

foo.yaml
```yaml
cronjob:
  fileName: foo_job
  schedule: "0 11 * * *"
  env:
    - name: FOO
      value: bar
```

bar.yaml
```yaml
cronjob:
  env:
  - name: BAR
    value: foo
```
cronjob.yaml template
```
apiVersion: batch/v1beta1
kind: CronJob
metadata:
  name: {{ template "cronjob.fullname" . }}
  labels:
    app: {{ template "cronjob.name" . }}
    chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
    release: {{ .Release.Name }}
    heritage: {{ .Release.Service }}
spec:
  schedule: {{ .Values.cronjob.schedule | quote }}
  startingDeadlineSeconds: {{ .Values.cronjob.startingDeadlineSeconds }}
  jobTemplate:
    spec:
      template:
        metadata:
          name: {{ template "cronjob.fullname" . }}
        spec:
          restartPolicy: Never
          imagePullSecrets:
          - name: quay-sts
          containers:
          - name: {{ template "cronjob.fullname" . }}
            imagePullPolicy: Always
            image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
            args:
            - npm
            - start
            - {{ .Values.cronjob.fileName }}
{{ if .Values.cronjob.env }}
            env:
{{ toYaml .Values.cronjob.env | indent 12 }}
{{ end }}
```

When running `helm install --dry-run --debug ./cronjob -f values/foo.yaml,values/bar.yaml` or when running `helm install --dry-run --debug ./cronjob -f values/foo.yaml -f values/bar.yaml`

It outputs the following:

**actual output**
```yaml
apiVersion: batch/v1beta1
kind: CronJob
metadata:
  name: intended-jaguar-cronjob
  labels:
    app: cronjob
    chart: cronjob-2.0.8
    release: intended-jaguar
    heritage: Tiller
spec:
  schedule: "0 11 * * *"
  startingDeadlineSeconds: 60
  jobTemplate:
    spec:
      template:
        metadata:
          name: intended-jaguar-cronjob
        spec:
          restartPolicy: Never
[...]

PR #30632, opened 2025-03-07
feat: Merge lists when passing multiple values files, allowing to sig…

rsafonseca:
…nal a unique key for merging lists of maps

This PR allows merging lists inside values files, which is a very useful feature when using overlays to simplify/merge deployment files, e.g. merging kubernetes environment variables from several values files.

It has the following behaviour:
- Behaviour is unchanged for any lists whose key name does not start with *

- Lists of plain values are concatenated when the key name is prefixed by *
- Lists of maps are concatenated when the key name is prefixed by *, allowing to specify a unique key (by prefixing the key name with *) for merging specific entries instead of adding. 
- The * prefix will be stripped from key names of list properties in the final rendering 

Addresses: 
[#3486](https://github.com/helm/helm/issues/3486)
https://github.com/fluxcd/helm-controller/issues/300

This enables users to opt-in to list merging with some degree of control over which lists are merged, and how they are merged.

Applicability example:

values1.yaml
```
env:
  - name: MYVAR1
    value: value1
  - name: MYVAR2
    value: value2
```
values2.yaml
```
*env:
  - *name: MYVAR1
    value: override_this_value
  - name: MYVAR3
    value: add_this
```
result:
```
env:
  - name: MYVAR1
    value: override_this_value
  - name: MYVAR2
    value: value2
  - name: MYVAR3
    value: add_this
```

**If applicable**:

### DIFF
--- a/pkg/chart/v2/loader/load.go
+++ b/pkg/chart/v2/loader/load.go
@@ -24,6 +24,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"reflect"
 	"strings"
 
 	"github.com/pkg/errors"
@@ -236,21 +237,79 @@ func LoadValues(data io.Reader) (map[string]interface{}, error) {
 
 // MergeMaps merges two maps. If a key exists in both maps, the value from b will be used.
 // If the value is a map, the maps will be merged recursively.
+// If the value is a list, the lists will be merged
 func MergeMaps(a, b map[string]interface{}) map[string]interface{} {
 	out := make(map[string]interface{}, len(a))
 	for k, v := range a {
 		out[k] = v
 	}
 	for k, v := range b {
-		if v, ok := v.(map[string]interface{}); ok {
+
+		if val, ok := v.(map[string]interface{}); ok {
 			if bv, ok := out[k]; ok {
 				if bv, ok := bv.(map[string]interface{}); ok {
-					out[k] = MergeMaps(bv, v)
+					out[k] = MergeMaps(bv, val)
 					continue
 				}
 			}
+		} else if reflect.TypeOf(v).Kind() == reflect.Slice {
+			if sourceList, ok := out[k].([]map[string]interface{}); ok {
+
+				val, ok := v.([]map[string]interface{})
+				if !ok {
+					log.Println("Property mismatch during merge")
+					out[k] = v
+					continue
+				}
+
+				out[k] = MergeMapLists(sourceList, val)
+				continue
+			} else if sourceList, ok := out[k].([]interface{}); ok {
+				if val, ok := v.([]interface{}); ok {
+					out[k] = append(sourceList, val...)
+				} else {
+					out[k] = v
+				}
+				continue
+			}
+
 		}
 		out[k] = v
 	}
 	return out
+}
+
+// MergeMapLists merges two lists of maps. If a prefix of * is set on a map key,
+// that key will be used to de-duplicate/merge with the source map
+func MergeMapLists(a, b []map[string]interface{}) []map[string]interface{} {
+	out := a
+	for j, mapEntry := range b {
+		var mergeKey string
+		var mergeValue interface{}
+		for k, v := range mapEntry {
+			if strings.HasPrefix(k, "*") {
+				mergeKey = k
+				mergeValue = v
+				b[j][strings.TrimPrefix(mergeKey, "*")] = v
+				delete(b[j], mergeKey)
+				break
+			}
+		}
+		if len(mergeKey) > 0 {
+			strippedMergeKey := strings.TrimPrefix(mergeKey, "*")
+			for i, sourceMapEntry := range out {
+				for k, v := range sourceMapEntry {
+					if (k == strippedMergeKey || k == mergeKey) && v == mergeValue {
+						mergedMapEntry := MergeMaps(sourceMapEntry, mapEntry)
+						out[i] = mergedMapEntry
+						break
+					}
+				}
+			}
+		} else {
+			out = append(out, mapEntry)
+		}
+	}
+	return out
+
 }
--- a/pkg/chart/v2/loader/load.go
+++ b/pkg/chart/v2/loader/load.go
@@ -303,23 +303,23 @@ func MergeMapLists(a, b []any) []any {
 			continue
 		}
 
-		var mergeKey string
-		var mergeValue interface{}
+		var uniqueKey string
+		var dedupValue interface{}
 		for k, v := range mapEntry {
 			if strings.HasPrefix(k, mergePrefix) {
-				mergeKey = k
-				mergeValue = v
+				uniqueKey = k
+				dedupValue = v
 				bj, ok := b[j].(map[string]any)
 				if !ok {
 					continue
 				}
-				bj[strings.TrimPrefix(mergeKey, mergePrefix)] = v
-				delete(bj, mergeKey)
+				bj[strings.TrimPrefix(uniqueKey, mergePrefix)] = v
+				delete(bj, uniqueKey)
 				break
 			}
 		}
-		if len(mergeKey) > 0 {
-			strippedMergeKey := strings.TrimPrefix(mergeKey, mergePrefix)
+		if len(uniqueKey) > 0 {
+			strippedMergeKey := strings.TrimPrefix(uniqueKey, mergePrefix)
 
 			for i, sourceMapEntry := range out {
 				sourceMapEntry, ok := sourceMapEntry.(map[string]any)
@@ -327,7 +327,7 @@ func MergeMapLists(a, b []any) []any {
 					continue
 				}
 				for k, v := range sourceMapEntry {
-					if (k == strippedMergeKey || k == mergeKey) && v == mergeValue {
+					if (k == strippedMergeKey || k == uniqueKey) && v == dedupValue {
 						mergedMapEntry := MergeMaps(sourceMapEntry, mapEntry)
 						out[i] = mergedMapEntry
 						break
--- a/pkg/chart/v2/loader/load.go
+++ b/pkg/chart/v2/loader/load.go
@@ -34,6 +34,8 @@ import (
 	chart "helm.sh/helm/v4/pkg/chart/v2"
 )
 
+const mergePrefix = "*"
+
 // ChartLoader loads a chart.
 type ChartLoader interface {
 	Load() (*chart.Chart, error)
@@ -252,23 +254,25 @@ func MergeMaps(a, b map[string]interface{}) map[string]interface{} {
 					continue
 				}
 			}
-		} else if reflect.TypeOf(v).Kind() == reflect.Slice {
-			if sourceList, ok := out[k].([]map[string]interface{}); ok {
+		} else if reflect.TypeOf(v).Kind() == reflect.Slice && strings.HasPrefix(k, mergePrefix) {
+			strippedKey := strings.TrimPrefix(k, mergePrefix)
+			out[strippedKey] = out[k]
+			delete(out, k)
+			if sourceList, ok := out[strippedKey].([]map[string]interface{}); ok {
 
 				val, ok := v.([]map[string]interface{})
 				if !ok {
 					log.Println("Property mismatch during merge")
-					out[k] = v
 					continue
 				}
 
-				out[k] = MergeMapLists(sourceList, val)
+				out[strippedKey] = MergeMapLists(sourceList, val)
 				continue
-			} else if sourceList, ok := out[k].([]interface{}); ok {
+			} else if sourceList, ok := out[strippedKey].([]interface{}); ok {
 				if val, ok := v.([]interface{}); ok {
-					out[k] = append(sourceList, val...)
+					out[strippedKey] = append(sourceList, val...)
 				} else {
-					out[k] = v
+					out[strippedKey] = v
 				}
 				continue
 			}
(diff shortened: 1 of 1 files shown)


rsafonseca:
I can write up the docs for this, just wanted to discuss the approach first and get the take of the reviewer(s)

pkg/chart/v2/loader/load.go
```diff
@@ -236,21 +239,81 @@ func LoadValues(data io.Reader) (map[string]interface{}, error) {
...
 	}
 	return out
 }
+
+// MergeMapLists merges two lists of maps. If a prefix of * is set on a map key,
+// that key will be used to de-duplicate/merge with the source map
+func MergeMapLists(a, b []map[string]interface{}) []map[string]interface{} {
+	out := a
+	for j, mapEntry := range b {
+		var mergeKey string
```
miledxz, line 291: let's update mergeKey to something more readable, what you think about `uniqueKey` ?

pkg/chart/v2/loader/load.go
```diff
@@ -236,21 +239,81 @@ func LoadValues(data io.Reader) (map[string]interface{}, error) {
...
+		var mergeValue interface{}
```
miledxz, line 292: same as above, what you think about naming this as `dedupValue` or `uniqueValue` ?

pkg/chart/v2/loader/load.go
```diff
@@ -236,21 +239,81 @@ func LoadValues(data io.Reader) (map[string]interface{}, error) {
...
+
 		}
 		out[k] = v
 	}
 	return out
 }
+
+// MergeMapLists merges two lists of maps. If a prefix of * is set on a map key,
+// that key will be used to de-duplicate/merge with the source map
+func MergeMapLists(a, b []map[string]interface{}) []map[string]interface{} {
```
miledxz, line 288: we should handle edge cases if two maps have multiple prefixed keys? Only the first one is used. Should this be extended ? 

also, if a or b contains nil maps? A nil check should be added.

pkg/chart/v2/loader/load.go
```diff
@@ -236,21 +239,81 @@ func LoadValues(data io.Reader) (map[string]interface{}, error) {
...
+
+		if val, ok := v.(map[string]interface{}); ok {
 			if bv, ok := out[k]; ok {
 				if bv, ok := bv.(map[string]interface{}); ok {
-					out[k] = MergeMaps(bv, v)
+					out[k] = MergeMaps(bv, val)
 					continue
 				}
 			}
+		} else if reflect.TypeOf(v).Kind() == reflect.Slice && strings.HasPrefix(k, mergePrefix) {
```
miledxz, line 257: using reflect for type checking can be inefficient sometimes, instead we can use type assertions like `val, ok := v.([]map[string]interface{})`

what is your opinion on that ?

pkg/chart/v2/loader/load.go
```diff
@@ -236,21 +239,81 @@ func LoadValues(data io.Reader) (map[string]interface{}, error) {
...
+			strippedKey := strings.TrimPrefix(k, mergePrefix)
+			out[strippedKey] = out[k]
+			delete(out, k)
+			if sourceList, ok := out[strippedKey].([]map[string]interface{}); ok {
+
+				val, ok := v.([]map[string]interface{})
+				if !ok {
+					log.Println("Property mismatch during merge")
```
miledxz, line 265: would be good to extend logging and include which key caused the issue

miledxz (commented):
Hi @rsafonseca thank you for this amazing PR 👍  I'm taking a look and will build binaries locally and test how this works, I added few comments for improvements but overall everything looks good.

Once again thank you for being part of community and I'm looking forward to future work on this PR and many others.

pkg/chart/v2/loader/load.go
rsafonseca: I would tend to agree, reflect is not the most efficient lib and would also increase the binary size, i just wrote it this way to make it a bit more readable, so i'm fine with changing this. For the plain lists we could do the same `v.([]interface{})`

pkg/chart/v2/loader/load.go
rsafonseca: easy enough, i'll push the change in the morning :)

pkg/chart/v2/loader/load.go
miledxz: @rsafonseca sure, let's try, thank you for collaboration

ingydotnet:
@rsafonseca I assume you are aware that `*foo` is YAML syntax for an [alias](https://yaml.org/spec/1.2.2/#3222-anchors-and-aliases) referencing a node with an anchor `&foo`. 

YAML plain (unquoted) scalars can't begin with a `*`, so calling it a prefix is wrong. The key `*env` is not a scalar (string) it's a YAML alias (pointer). If the current loader happens to be reporting it as a string, that's technically a bug or at least incorrect behavior.

Also the example files are invalid YAML (and cause issue with YAML tools) because you can't define an alias before the anchor it references:
```
$ yq values2.yaml 
Error: bad file 'values2.yaml': yaml: unknown anchor 'env' referenced
$ yamllint values2.yaml 
values2.yaml
  1:1       warning  missing document start "---"  (document-start)
  1:1       error    found undeclared alias "env"  (anchors)
  2:5       error    found undeclared alias "name"  (anchors)
```

rsafonseca:
Those are all great points @ingydotnet I hadn't considered the impact on yaml tooling and those definitely can prevent this from working seamlessly in some environments. 
The idea of using * as a prefix was precisely to avoid any potential clashing with other key names, and avoid having to opt in to enable merging via additional cli arguments or lib options.
Given the caveats you noted, I think we can come up with a better prefix that doesn't clash with any existing uses of helm out there, if you have any suggestions they're welcome, if not I'll give it a better think soon :)

rsafonseca:
```
bash-5.2$ cat values2.yaml 
\*env:
  - \*name: MYVAR1
    value: override_this_value
  - name: MYVAR3
    value: add_this
bash-5.2$ yq values2.yaml 
\*env:
  - \*name: MYVAR1
    value: override_this_value
  - name: MYVAR3
    value: add_this
bash-5.2$ yamllint values2.yaml 
values2.yaml
  1:1       warning  missing document start "---"  (document-start)
```

rsafonseca:
hum... this needs more work, this approach causes schema validation to fail, i hadn't tested that 🤔 

Error: values don't meet the specifications of the schema(s) in the following chart(s):
test-chart:
- deployment: Additional property \*env is not allowed

ingydotnet:
> maybe something like this, WDYT @ingydotnet ?
> 
> ```
> bash-5.2$ cat values2.yaml 
> \*env:
>   - \*name: MYVAR1
> ```

From a YAML perspective that _does_ make the keys be valid strings...
You could also use any number of extremely-unlikely-to-exist-plain-yaml-keys including keys with whitespace in them like `\ env`.

However (since you asked) this feels like a too generic of a DSL to address a very specific merge strategy for a very specific problem.
As I grok it so far you want to make the general helm merge strategy for merging extra values files into values.yaml.
This strategy looks for specially marked nodes in the overrides files to be "merged" (but a very special kind of merge that works on sequences of mappings).

Changing the YAML loader to do this with a new DSL seems like the wrong choice.
Especially since there's countless other useful transforms that people might want.

The problem to solve is https://github.com/helm/helm/issues/3486, right?
The first thing that comes to mind is why use:
```
env:
- name: FOO
  value: 12345
 ```
 instead of:
 ```
 env:
   FOO: 12345
 ```
where the merge already works the way I think you want?

Regardless, the need for custom transformations certainly does exist!

If you give me a day or so to whip up a demo, I have another approach I'd like to run by you.
It should handle this use cases with elegance, and also allow for any other type of transformation that any Helm user thinks up...

**Comment by rsafonseca** [comment]
> The first thing that comes to mind is why use:
> 
> env:
> - name: FOO
>   value: 12345
> instead of:
> 
> env:
>   FOO: 12345
> where the merge already works the way I think you want?

File: pkg/chart/v2/loader/load.go (hunk 2 above)
Comment (rsafonseca) [reply pkg/chart/v2/loader/load.go]: done

File: pkg/chart/v2/loader/load.go (hunk 4 above)
Comment (rsafonseca) [reply pkg/chart/v2/loader/load.go]: done

File: pkg/chart/v2/loader/load.go (hunk 5 above)
Comment (rsafonseca) [reply pkg/chart/v2/loader/load.go]: done

File: pkg/chart/v2/loader/load.go (hunk 3 above)
Comment (rsafonseca) [reply pkg/chart/v2/loader/load.go]: regarding the multiple prefixed keys, i don't think it's worth adding unless a specific need comes up for some reason, i'd say it's highly unlikely that anyone would need to match on multiple keys for uniqueness

as for the nil checks, the changes i've done to type assertions should prevent that issue, not sure if i missed any potential nil case 🤔

**Comment by ingydotnet** [comment]
Hi @rsafonseca 

I'll give my alternative solution at the end of this, but I have to say that I think this PR overall is not a good idea. Essentially you are changing (at least from a user perspective) how YAML works in Helm, and for one very specific use case.
There are likely many ways to do this, but I wouldn't go this route.

I have a bunch of thoughts and questions around this. Full disclosure I'm not a Kubernetes or Helm expert (but I guess I would be a YAML expert. :)

I'm not sure what the prefix is needed for in your solution. Could you explain that? 

I wonder if you based the idea for it after YAML's `<<` merge key.
The `<<` merge key was a bit of an accident in YAML and was removed in YAML 1.2 (2009) but many frameworks still implement it (although almost all slightly differently). YAML is meant to be a data only language, like JSON...

Mostly. Tags are used as metadata that can trigger functions. It this effort were to go forward, you'd really want to use tags to indicate your intent, not unique plain scalars (prefixes). That said, I don't think there's prior art for tags within Helm, (nor with using prefixes) so probably not worth pursuing.

I don't think you've written out what the exact behavior of your list-merge is. There are so many little details that would be questioned later on. Those need to be questioned up front, esp if you want to change Helm core for it.

I _think_ that Helm's YAML loader comes upstream from Kubernetes. Changing the downstream to differ might not be desirable.

I know that https://github.com/helm/helm/issues/3486 is an issue that many people want a fix for, but a deep merge has dozens of details that should be configurable. Addressing just one in a bespoke manner isn't a good idea.

### A YS Detour

I've spent the last year+ creating a new functional programming language called [YS](https://yamlscript.org) ("wise" aka YAMLScript) that is 100% valid YAML syntax, embeds cleanly in YAML to extend it, and is itself a YAML loader (available in many programming languages).

I gave a talk about it at the last KubeCon in SLC, and presented [HelmYS](https://yamlscript.org/helmys/) for using YS in chart templates.

One thing you can do easily in YS is make new functions. Here's one that implements a list merge!

```
$ cat example.ys 
--- &list1
- name: FOO
  value: foo
- name: BAR
  value: bar

--- &list2
- name: FOO
  value: new foo
- name: BAZ
  value: new baz

---
!YS-v0

defn list-merge(list1 list2): !:vals
  reduce \(_.assoc(%2.name %2)) {}:
    list1 + list2

say: !:list-merge*:yaml/dump:
- *list1
- *list2
$ ys example.ys 
- name: FOO
  value: new foo
- name: BAR
  value: bar
- name: BAZ
  value: new baz
```

This merge result seems the same as what you had, but the nice thing here is that you can tweak it to your exact needs over time.

YS works great in Helm templates and other aspects of Helm.
I'm currently working with the Helm 4 team to have YS work better there for more situations,.


## A Possible Solution for YAML List Merges

First off sorry if you thought I was going to make a modification of your direction.

Here's what I came https://github.com/ingydotnet/helm-post-renderer-merge-list-example

This is done as a post-renderer that anyone can use today.

I'm keen to hear your thoughts on it.

rsafonseca:
Hi @ingydotnet, thanks for your extensive analysis from a YAML perspective.

> I'm not sure what the prefix is needed for in your solution. Could you explain that?
The prefix is needed in order to be able to the optional behaviour of merging lists, with the additional benefit of avoiding a breaking change.
Currently, the Helm Values files loader merges maps, but replaces lists, the prefix allows the user to signal which lists they would need to merge, or not include the prefix to keep the existing behaviour (replace it entirely)

>  There are likely many ways to do this, but I wouldn't go this route.
Yes, there are likely many ways to do it, but this is the more elegant and hands-off solution, which also avoids introducing additional complexity for the user. Adding a prefix to a key to signal a behaviour is a simple user-centric solution to solve this problem, whenever the user needs it.

> I think that Helm's YAML loader comes upstream from Kubernetes. Changing the downstream to differ might not be desirable.
The implementation of how merges are handled are Helm's and not coming from upstream. This PR targets behaviour that is specific to helm (how to merge some data types when passing multiple values files).

Sadly, i don't think your YamlScript solution would work here, without needing to fundamentally change how helm is handling multiple Yaml files for values inputs, and on top of that making it harder for the end-user to get to the intended behaviour (adding a simple prefix to a list that needs to be merged when passing multiple values files is much easier than to ask a simple helm user to write yaml functions/transforms for this)

The solution in this PR is the cleanest solution that doesn't require extensive changes to how helm loader works, provides a per-list configurable opt-in behaviour and while there's potentially dozens of other things that you can do with YamlScript, there's no other clear problem that needs solving (at least that I know of) in how helm merges separate values files, though this does seem potentially very handy for folks writing helm charts (which again, is not what this PR is touching).

The post-render solution also makes things much more complex for the user, and uses a single values file, which in turn makes it unusable when you have multiple ( > 2 layers ), and requires potentially writing a number of custom patches instead of simply feeding a number of ordered yaml values files to helm.

robertsirc:
Why don't we have test to cover this new functionality?

jmleddy:
@miledxz does this actually have a chance of getting merged?

miledxz:
> @miledxz does this actually have a chance of getting merged?

sorry for late reply,

yes, I would say it's WIP and after writing unit test / testing, we can proceed

jmleddy:
Okay, I'm not sure if @raffaelespazzoli is still interested, but I could definitely write some unit tests and documentation for this. Worst case open another PR

pkg/chart/v2/loader/load.go
jmleddy: This kind of type coercion isn't actually possible sadly https://go.dev/doc/faq#convert_slice_of_interface

jmleddy:
Couldn't open a pr for @rsafonseca's branch, opened pr #31314  instead.

crabique:
For some additional perspective, one approach to this I really like is what rke2 does: https://github.com/rancher/rke2/issues/3108#issuecomment-1167812384

Limiting this to be an "explicit append" instead of a flexible merge reduces the complexity by a lot, and the `+` suffix in the key name is verbose enough to understand what it does at a glance while not breaking the YAML spec.

rm3l:
Hi team,
Any chance we can get this and/or #31314 merged? This would significantly reduce user confusion around how Helm handles lists.
Thanks.

rm3l:
not stale

PR #30677, opened 2025-03-17
chore: Update Golang to v1.24

dongjiang1989:

This is needed to update to the latest k8s packages(v1.33.beta.0) as they require Go 1.24

**If applicable**:


mattfarina (changes requested):
`.github/env` files are not part of the GitHub Actions features. So, the change in this PR didn't actually work.

In this case `go-version` is an empty string. This is leading to the CI log message:

```console
Warning: go-version input was not specified. The action will try to use pre-installed version.
```

Can you please create a PR where the go version is working.

dongjiang1989:
> `.github/env` files are not part of the GitHub Actions features. So, the change in this PR didn't actually work.
> 
> In this case `go-version` is an empty string. This is leading to the CI log message:
> 
> ```
> Warning: go-version input was not specified. The action will try to use pre-installed version.
> ```
> 
> Can you please create a PR where the go version is working.

Thanks @mattfarina
Fixed. 
PTAL recheck it.

.github/env
```diff
@@ -0,0 +1,2 @@
+golang-version=1.24
+golangci-lint-version=v1.64
```
TerryHowe, line 2: I don't think it makes a lot of sense to set environment variables with a `-` in the name even if this happens to work in this context. A `_` makes a lot mor sense and generally environment varialbe names are uppercase, so GOLANG_VERSION and GOLANGCI_LINT_VERSION.

I am not a manainter though.

dongjiang1989: Thanks @TerryHowe  
I got it.
Fixed

.github/workflows/release.yml
```diff
@@ -24,14 +24,15 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Import environment variables from file
```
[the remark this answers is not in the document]
dongjiang1989: `add variables to environment from file` ?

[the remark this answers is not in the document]
TerryHowe: `GITHUB_ENV` is the name of the environment file in my view and you are adding to it.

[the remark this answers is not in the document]
dongjiang1989: Thanks @TerryHowe 
Got it

TerryHowe (approved):
/lgtm

.github/workflows/build-test.yml
```diff
@@ -19,6 +19,8 @@ jobs:
     steps:
       - name: Checkout source code
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # pin@v4.2.2
+      - name: Import environment variables from file
+        run: cat ".github/env" >> "$GITHUB_ENV"
```
scottrigby: I think this makes sense. GitHub recommends exporting env vars to  `GITHUB_ENV` here: https://docs.github.com/en/actions/sharing-automations/creating-actions/metadata-syntax-for-github-actions#runsstepsenv. We also are not exporting anything sensitive—noting because these are printed to console, which is fine for all of this.

.github/workflows/release.yml
```diff
@@ -24,14 +24,15 @@ jobs:
...
+        run: cat ".github/env" >> "$GITHUB_ENV"
+
       - name: Setup Go
         uses: actions/setup-go@f111f3307d8850f501ac008e886eec1fd1932a34 # pin@5.3.0
         with:
           go-version: '${{ env.golang-version }}'
-
       - name: Run unit tests
         run: make test-coverage
-
```
scottrigby: I wouldn't remove the blank lines around "Run unit tests" step, since the rest have blank lines let's keep it consistent.

scottrigby (changes requested):
This all works now. Tested both with CI and locally. Nice job.

Matt's change request here has been fixed https://github.com/helm/helm/pull/30677#pullrequestreview-2691827577

I'm asking for one small style change request. Revert the one change where you removed extra lines. Let's keep extra lines consistent in that file.

Once that's done, this should be good to go 👍

.github/env
```diff
@@ -0,0 +1,2 @@
+GOLANG_VERSION=1.24
+GOLANGCI_LINT_VERSION=v1.64
```
dongjiang1989, line 2:
```suggestion
GOLANGCI_LINT_VERSION=v2.0.2
```

dongjiang1989:
> This all works now. Tested both with CI and locally. Nice job.
> 
> Matt's change request here has been fixed [#30677 (review)](https://github.com/helm/helm/pull/30677#pullrequestreview-2691827577)
> 
> I'm asking for one small style change request. Revert the one change where you removed extra lines. Let's keep extra lines consistent in that file.
> 
> Once that's done, this should be good to go 👍

Thanks @scottrigby 
Fixed and rebase done

scottrigby (approved):
LGTM 👍 Thanks @dongjiang1989 for your persistence.

Note for other reviewers: never mind my nit about line breaks around steps. Our GH Actions workflows are already inconsistent from file to file and sometimes within the same file. If we care to format these consistently we can do that in a follow up.

robertsirc (approved):
LGTM

mattfarina (approved):

py7778 (approved):

PR #30686, opened 2025-03-19
Remove deprecated code

mattfarina:
This change removes deprecated code (marked for removal in Helm v4).

This also exposes the Helm template engine function map for SDK users. The removed public function constructor was used by SDK users who were trying to reconstruct some of Helms template functions so their custom code can render render templates. With the removal of this public function that no longer works. By exposing this publicly, those use cases can continue.

**What this PR does / why we need it**: Removing deprecated code while keeping a solution for community user use cases.

**Special notes for your reviewer**: There are three ways to handle this situation.

1. Remove `NewLookupFunction` and no longer expose this for SDK users. Some of the SDK users are using it outside of Helm. Some of them are partially providing Helms functions. I did not find a public app on GitHub using all of Helms template functions. Removing it means they will need to vendor a copy of our code or write their own.
2. Do not deprecate the `NewLookupFunction` function and and instead make it `newLookupFunction`. This will help some community users who use it.
3. Expose all the Helm template functions for the community.

My initial reaction was to go with (1) to keep this code Helm focused. 

If we are going to expose our template functions, we could do so with those in mind who are working with Helm charts and need all our functions. This is to power SDK users who are doing Helm work. That's why I went with (3) in this PR.

**If applicable**:

### DIFF
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -23,7 +23,6 @@ import (
 	"os"
 	"path"
 	"path/filepath"
-	"regexp"
 	"strings"
 
 	"github.com/pkg/errors"
@@ -62,21 +61,6 @@ var (
 	// errPending indicates that another instance of Helm is already applying an operation on a release.
 	errPending = errors.New("another operation (install/upgrade/rollback) is in progress")
 )
-
-// ValidName is a regular expression for resource names.
-//
-// DEPRECATED: This will be removed in Helm 4, and is no longer used here. See
-// pkg/lint/rules.validateMetadataNameFunc for the replacement.
-//
-// According to the Kubernetes help text, the regular expression it uses is:
-//
-//	[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*
-//
-// This follows the above regular expression (but requires a full string match, not partial).
-//
-// The Kubernetes documentation is here, though it is not entirely correct:
-// https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
-var ValidName = regexp.MustCompile(`^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`)
 
 // Configuration injects the dependencies that all actions share.
 type Configuration struct {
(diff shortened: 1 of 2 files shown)


robertsirc (dismissed):
LGTM

banjoh (changes requested):
I attempted invoking `FuncMap(...)` but got blocked constructing `*ClientProvider`. For me to get the same behaviour as the SDK, I'd need to reimplement what's in `clientProviderFromConfig`

Two options come to mind to address this
- Have `FuncMap` accept `*rest.Config` instead of `*ClientProvider` just like the deprecated function had
- Expose a constructor for `*ClientProvider` to use when invoking `FuncMap(...)` i.e
```go
func NewClientProvider(config *rest.Config) ClientProvider {
	return clientProviderFromConfig{config: config}
}
```

**Nice to have**: Can we make `lookupFunc` function prototype public? This way, a client does not have to redeclare it

Here is my sample code

```go
type LookupFunc = func(apiversion string, resource string, namespace string, name string) (map[string]interface{}, error)

func lookup(apiversion string, resource string, namespace string, name string) (map[string]interface{}, error) {
	var cfg *rest.Config
	cp := engine.NewClientProvider(cfg)
	lkFunc := engine.FuncMap(false, false, &cp)["lookup"]
	// OR: lkFunc := engine.FuncMap(false, false, cfg)["lookup"]
	lookupFunc, ok := lkFunc.(LookupFunc)
	if !ok {
		return nil, fmt.Errorf("Failed to cast lookup function")
	}

	return lookupFunc(apiversion, resource, namespace, name)
}
```

scottrigby (changes requested):
Thanks @mattfarina for reaching out about existing use of this function.

To continue allowing end users to make use of this functionality, we would need to either add an additional function as @banjoh said in https://github.com/helm/helm/pull/30686#pullrequestreview-2702482532, or reconsider the earlier (unexplained) comment to deprecate this function and just leave as-is in Helm 4.

mattfarina:
@scottrigby @banjoh  can you please re-review this. I removed the deprecated notice on the public function in the SDK that y'all are using. This is the simpler path.

banjoh (approved):
LGTM

scottrigby (approved):
Lgtm 👍

PR #13610, opened 2025-01-07
Improve Test Coverage for --take-ownership Flag in Helm Commands

AnanyaDevGo:
This PR enhances the test coverage for the --take-ownership flag introduced in [#13439](https://github.com/helm/helm/pull/13439). It addresses gaps in testing for the helm upgrade --install --take-ownership functionality, ensuring both success and failure scenarios are adequately covered.

closes #13519 

Key Changes:
Setup Test Environment:

Added a SetupTestEnvironment utility to create a fake Kubernetes namespace for testing.
Ensures namespace cleanup after test execution.
Integration Tests for --take-ownership:

Validates ownership transfer in scenarios involving Helm-managed resources.
Tests edge cases with both successful and failing updates when --take-ownership is toggled.
Enhanced Client Behavior:

Ensures accurate simulation of Kubernetes client operations using FailingKubeClient.
Bug Reproduction & Fix Validation:

Verifies the resolution of a prior issue where helm upgrade --install --take-ownership failed to handle existing unmanaged resources correctly.
Tested locally  and passed all the testcases
Checklist:
- [ ] This PR contains user-facing changes
- [x] This PR contains unit tests
- [ ] This PR has been tested for backward compatibility
Signed-off-by: Ananya E P areduvail@gmail.com

PR #30697, opened 2025-03-21
Fix --take-ownership for custom resources - closes #30622

### LINKED ISSUE #30622: --take-ownership does not adopt custom resources
If a resource exists in the cluster and is to be adopted by `helm install --take-ownership`, it is left unchanged while helm reports the installation to have succeeded.

# Reproduce

0. Helm version v3.17.1+g980d8ac

1. Install any CRD

```yaml
apiVersion: apiextensions.k8s.io/v1
metadata:
  name: datas.crd.com
spec:
  group: crd.com
  scope: Namespaced
  names:
    plural: datas
    singular: data
    kind: Data
  versions:
  - name: v1
    served: true
    storage: true
    schema:
      openAPIV3Schema:
        type: object
        x-kubernetes-preserve-unknown-fields: true
```

2. Apply a custom resource onto the cluster.

```yaml
kind: Data
apiVersion: crd.com/v1
metadata:
  name: crd-test
  namespace: kube-system
  labels:
    test: adopt
spec:
  color: blue
  size: large
```

3. `helm install --take-ownership` a chart that contains a different version of the custom resource.

```yaml
kind: Data
apiVersion: crd.com/v1
metadata:
  name: crd-test
  namespace: kube-system
  labels:
    test: adopt
spec:
  weight: heavy
```

4. Notice that the custom resource has not changed at all (still at initial state) but helm reported the operation to have succeeded, including the creation of a helm release.

```
helm install foo ./foo --take-ownership 
NAME: foo
LAST DEPLOYED: Wed Mar  5 17:10:16 2025
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
```

# Suggested Solution

Transparently use three-way merge for `--take-ownership`.

p-se:
If a resource exists in the cluster and is to be adopted by helm install --take-ownership, it is left unchanged while helm reports the installation to have succeeded.

This is due to CRs and CRDs being merged without three-way-merge, which results in an empty patch.

By using a three-way-merge transparently when --take-ownership is used, helm behaves as expected without breaking previous behavior.

Fixes #30622

**If applicable**:
  --take-ownership does not appear to be documented at all, so perhaps yes from that perspective, but no from the perspective of this PR.
  The changes in this PR are meant to not change any existing behavior, except for how helm behaves when called with `--take-ownership`.


p-se:
The YAML after a successful `--take-ownership` with the changes of this PR:

```yaml
apiVersion: crd.com/v1
kind: Data
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"crd.com/v1","kind":"Data","metadata":{"annotations":{},"labels":{"test":"adopt"},"name":"crd-test","namespace":"kube-system"},"spec":{"color":"blue","size":"large"}}
    meta.helm.sh/release-name: foo
    meta.helm.sh/release-namespace: default
  creationTimestamp: "2025-03-21T12:37:34Z"
  generation: 2
  labels:
    test: adopt
  name: crd-test
  namespace: kube-system
  resourceVersion: "374259"
  uid: 4e87ddf2-2012-4ee7-a804-a8a7406c72b7
spec:
  color: blue
  size: large
  weight: heavy
```

banjoh:
@p-se the PR needs to be merged to `main` first, then another PR cherry picking commits into `dev-v3` needs to be created. Please have the PR point to `main` branch

pkg/kube/interface.go
```diff
@@ -54,7 +54,7 @@ type Interface interface {
 
 	// Update updates one or more resources or creates the resource
 	// if it doesn't exist.
-	Update(original, target ResourceList, force bool) (*Result, error)
+	Update(original, target ResourceList, force bool, threeWayMergeForUnstructured bool) (*Result, error)
```
banjoh, line 57: Since we are targeting to fix this in v3, this API change will break backward compatibility

banjoh (changes requested):
I tested a happy path test using the example in https://github.com/helm/helm/issues/30622 and it works. Testing a real use case would be needed.

robertsirc (changes requested):
Hello, thank you for your PR could you please update and ensure you have test.

pkg/kube/client.go
```diff
@@ -645,14 +647,27 @@ func createPatch(target *resource.Info, current runtime.Object) ([]byte, types.P
...
-	// use the jsonpatch.CreateMergePatch. Strategic Merge Patch is not supported
+	// use generic JSON merge patch. Strategic Merge Patch is not supported
 	// on objects like CRDs.
 	_, isUnstructured := versionedObject.(runtime.Unstructured)
 
 	// On newer K8s versions, CRDs aren't unstructured but has this dedicated type
 	_, isCRD := versionedObject.(*apiextv1beta1.CustomResourceDefinition)
 
 	if isUnstructured || isCRD {
+		if threeWayMergeForUnstructured {
```
banjoh, line 658: @mattfarina in light of https://github.com/helm/helm/pull/13411#pullrequestreview-2406232091, is there any objection defaulting to 3-way merge for CRDs and unstructured resources in Helm v3, or should we have a CLI flag to opt into the merge strategy?

For v4 I'm leaning towards having 3-way merge as default.

pkg/kube/interface.go
p-se: Would like me to add a method to the interface and use that new method instead of adapting the existing method to preserve backward compatibility?

pkg/action/install.go
```diff
@@ -455,7 +455,7 @@ func (i *Install) performInstall(rel *release.Release, toBeAdopted kube.Resource
 	if len(toBeAdopted) == 0 && len(resources) > 0 {
 		_, err = i.cfg.KubeClient.Create(resources)
 	} else if len(resources) > 0 {
-		_, err = i.cfg.KubeClient.Update(toBeAdopted, resources, i.Force)
+		_, err = i.cfg.KubeClient.Update(toBeAdopted, resources, i.Force, i.TakeOwnership)
```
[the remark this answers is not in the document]
p-se: I'm sorry, I'm not sure what to test here for. The fake client used in the existing tests for KubeClient discards any real processing. It had ignored two of three arguments before this implementation and with this implementation it will ignore three of four arguments.

```go
// from pkg/kube/fake/printer.go

// Update implements KubeClient Update.
func (p *PrintingKubeClient) Update(_, modified kube.ResourceList, _, _ bool) (*kube.Result, error) {
	_, err := io.Copy(p.Out, bufferize(modified))
	if err != nil {
		return nil, err
	}
	// TODO: This doesn't completely mock out have some that get created,
	// updated, and deleted. I don't think these are used in any unit tests, but
	// we may want to refactor a way to handle future tests
	return &kube.Result{Updated: modified}, nil
}
```

However, I do have provided tests for the added functionality in `pkg/kube/client_test.go` which also ensure that previous functionality isn't altered for when `--take-ownership` isn't used.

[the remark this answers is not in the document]
banjoh: @robertsirc there are no tests for `--take-ownership` functionality at all today. We do have 3 PRs adding tests to cover this

- https://github.com/helm/helm/pull/30657
- https://github.com/helm/helm/pull/13610
- https://github.com/helm/helm/pull/13541

Would it be sufficient to ensure one of these PRs is reviewed and contains tests for this functionality, or should we incorporate changes in those PRs here?

pkg/kube/interface.go
banjoh: I'll bring this PR up in the developer call tomorrow. Depending on the outcome, you might not need to make any change to this function.

Notes: https://docs.google.com/document/d/1d-6xJEx0C78csIYSPKJzRPeWaHG_8W1Hjl72OJggwdc/edit?tab=t.0#heading=h.cafxi2yfgqg1

If we can default to 3-way merging, the additional `threeWayMergeForUnstructured` parameter may not be needed. If we need the parameter, the way to extend the API is to create a new interface which kube Client implements. Take a look at [here](https://github.com/helm/helm/blob/dev-v3/pkg/kube/interface.go#L78) and [here](https://github.com/helm/helm/blob/dev-v3/pkg/kube/interface.go#L125). In golang, extending an interface is a breaking change because any type implementing that interface will break

banjoh: @p-se remember to add `// TODO Helm 4: Remove this...` in the new interface, similar to the other extensions

banjoh: @p-se from today's developer call, it was agreed not to have to 3-way merging for CRDs and unstructured resources as the default behaviour so as not to surprise users with a change in behaviour.

We will keep the `threeWayMergeForUnstructured` boolean that will be `true` whenever `--take-ownership` flag is set.

banjoh: Please create a new interface like below (or similar) that `kube.Client` implements, then revert the previous change you made. Once that's done this PR should be good to go

```go
// TODO Helm 4: Remove InterfaceThreeWayMerge and integrate its method(s) into the Interface.
type InterfaceThreeWayMerge interface {
	UpdateThreeWayMerge(original, target ResourceList, force bool, threeWayMergeForUnstructured bool) (*Result, error)
}
```

pkg/action/install.go
```diff
@@ -459,7 +460,13 @@ func (i *Install) performInstall(rel *release.Release, toBeAdopted kube.Resource
 	if len(toBeAdopted) == 0 && len(resources) > 0 {
 		_, err = i.cfg.KubeClient.Create(resources)
 	} else if len(resources) > 0 {
-		_, err = i.cfg.KubeClient.Update(toBeAdopted, resources, i.Force)
+		if i.TakeOwnership {
+			if kubeClient, ok := i.cfg.KubeClient.(kube.InterfaceThreeWayMerge); ok {
```
banjoh, line 464: nit: You could remove this check since there is `var _ InterfaceThreeWayMerge = (*Client)(nil)` check that ensures the client implements this API. If you want to keep this check, then return an error in an `else` block

banjoh (approved):
LGTM

scottrigby (dismissed):
@p-se nice work 👍 LGTM

Re @robertsirc https://github.com/helm/helm/pull/30697#discussion_r2008245754 about tests. I agree we need to add tests to PRs whenever possible. Currently this PR fixes a glaring bug for a feature that did not have tests for the feature, but as @banjoh noted there are several open PRs for that already: https://github.com/helm/helm/pull/30697#discussion_r2012511407. However, this PR does add tests for code that is changed—specifically the merge strategies—which I think is important and covers what we need for this PR.

scottrigby:
Needs to remove the empty test commit (will also fix missing DCO) https://github.com/helm/helm/pull/30697/commits/250c69b03e147eaf7550d16851dcaae0a05b8b18 and now needs to resolve conflicts in upstream `pkg/kube/client.go`.

ivelichkovich:
Thanks for adding this, lgtm

robertsirc (dismissed):
LGTM

scottrigby (approved):
Nice work. Approving again to allow merge

PR #30685, opened 2025-03-19
Add similar ready debug logs as statefulSetReady for job and deployment

benoittgt:
**What this PR does / why we need it**:

⚠️  Please review first: https://github.com/helm/helm/pull/30698

This PR is build on top of https://github.com/helm/helm/pull/30698.

We are consuming debug logs on our project based on the helm SDK. We would like to add even more logs when resources like deployments and jobs are ready. At the moment there is a difference between statefullsets and deployments for example.

```
Deployment is not ready: l-app-staging/g-station-staging-lfapp-web. 1 out of 2 expected pods are ready
Deployment is not ready: l-app-staging/g-station-staging-lfapp-web. 1 out of 2 expected pods are ready
StatefulSet is ready: l-app-staging/g-station-staging-rediscache-master. 1 out of 1 expected pods are ready
StatefulSet is ready: l-app-staging/g-station-staging-redis-master. 1 out of 1 expected pods are ready
```

Thanks in advance for the review.

If approved, I can backport it to v3.

PR #30603, opened 2025-02-27
converting inline log to slog

robertsirc:
I converted inline log to slog. This is my first pass there will be more when i jump into other sections.

PR #30698, opened 2025-03-21
Migrate ready checks to slog

benoittgt:

While working on https://github.com/helm/helm/pull/30685 [it was mentioned that we should migrate to slog](https://github.com/helm/helm/pull/30685#discussion_r2006477698) following this work https://github.com/helm/helm/pull/30603. To simplify the initial contribution, I think it's a good idea to first migrate `ready.go` to slog, then do the contribution. 
This is a first attempt to migrate. I am not sure what do you want to do with log level. Everything as debug log ?

**If applicable**:

### DIFF
--- a/pkg/kube/ready.go
+++ b/pkg/kube/ready.go
@@ -287,17 +287,17 @@ func (c *ReadyChecker) deploymentReady(rs *appsv1.ReplicaSet, dep *appsv1.Deploy
 	}
 	// Verify the generation observed by the deployment controller matches the spec generation
 	if dep.Status.ObservedGeneration != dep.ObjectMeta.Generation {
-		slog.Debug("Deployment is not ready: observedGeneration does not match spec generation",
+		slog.Debug("Deployment generation does not match spec",
 			"namespace", dep.Namespace,
 			"name", dep.Name,
-			"observedGeneration", dep.Status.ObservedGeneration,
-			"specGeneration", dep.ObjectMeta.Generation)
+			"actualGeneration", dep.Status.ObservedGeneration,
+			"expectedGeneration", dep.ObjectMeta.Generation)
 		return false
 	}
 
 	expectedReady := *dep.Spec.Replicas - deploymentutil.MaxUnavailable(*dep)
 	if !(rs.Status.ReadyReplicas >= expectedReady) {
-		slog.Debug("Deployment is not ready: not enough pods ready",
+		slog.Debug("Deployment does not have enough pods ready",
 			"namespace", dep.Namespace,
 			"name", dep.Name,
 			"readyReplicas", rs.Status.ReadyReplicas,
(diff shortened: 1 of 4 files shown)


pkg/kube/ready.go
```diff
@@ -230,18 +226,18 @@ func (c *ReadyChecker) isPodReady(pod *corev1.Pod) bool {
 			return true
 		}
 	}
-	c.log("Pod is not ready: %s/%s", pod.GetNamespace(), pod.GetName())
```
benoittgt, line 233: I choose a capital letter for the first letter because it represent the resource.

robertsirc:
first thank you for the PR, I do have a question it isn't directed at you but to everyone else reviewing if it makes sense to change "name" to "Job" or the respective object. I am sure @mattfarina and @TerryHowe have opinions. For future reference please tag me in any PR when it comes to logging, it is my goal to get this in for Helm 4.

benoittgt:
>  it makes sense to change "name" to "Job" or the respective object

I tink it's a good idea.

> For future reference please tag me in any PR when it comes to logging, it is my goal to get this in for Helm 4.

Yes ! For sure. 🙏

robertsirc:
Like if someone was searching for the structured log for a job named "blah" it might make sense to call this for what it is not just a name.

benoittgt:
Hello @robertsirc 

We are using the helm-sdk and we are consuming debug log by having a custom DebugLog [here](https://github.com/helm/helm/blob/cc0bbbd6d6276b83880042c1ecb34087e84d41eb/pkg/action/action.go#L373). I'm wondering how this will work with slog ?
DebugLog was interesting because we could pipe logs of a specific upgrade with dedicated informations, like the original request uuid.

Thanks 😊

TerryHowe (commented):
Generally looks good, just wondered what people thought about updating the log message content a bit.

robertsirc:
> Hello @robertsirc
> 
> We are using the helm-sdk and we are consuming debug log by having a custom DebugLog [here](https://github.com/helm/helm/blob/cc0bbbd6d6276b83880042c1ecb34087e84d41eb/pkg/action/action.go#L373). I'm wondering how this will work with slog ? DebugLog was interesting because we could pipe logs of a specific upgrade with dedicated informations, like the original request uuid.
> 
> We should probably keep the log setup but instead having a custom interface that accept slog. So we could pass a custom one if needed to keep the same behavior. I think my proposal is not correct.
> 
> Thanks 😊

So, we would like to remove how we are doing this logging with `DebugLog` and just use `slog` statically.

benoittgt:
Thanks @robertsirc for your answer.

I will adapt my work to follow your choice in https://github.com/helm/helm/pull/30708 without havind the ability to pass a custom logger.

benoittgt:
In my attempt to make changes suggested in https://github.com/helm/helm/pull/30698#issuecomment-2777941574, to have a fully functionning code I had to go much much further. So I've updated the migrated pull request. See: https://github.com/helm/helm/pull/30708

cc @robertsirc

PR #30704, opened 2025-03-24
Handle missing resources during upgrade and rollback

### LINKED ISSUE #13437: Failure in pre-upgrade when a manifest is removed makes it impossible to roll back
If a chart used to contain a manifest like a ConfigMap that is no longer present in the next version of the chart and the upgrade fails during pre-upgrade, `helm rollback` will fail spuriously.

A manifest may no longer be present in a new version of the chart for a number of reasons: Merging RoleBindings, combining two Deployments of one container per pod into one Deployment of two containers per pod, etc.

To reproduce create a chart which installs a ConfigMap upon installation, contains nothing during upgrade except a pre-upgrade hook that will fail like so:

```yaml
{{- if .Release.IsInstall -}}
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: cfg
data:
  foo: bar
{{ end -}}
{{- if .Release.IsUpgrade -}}
---
apiVersion: batch/v1
kind: Job
metadata:
  name: "failurecauser"
  annotations:
    helm.sh/hook: pre-upgrade
    helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded,hook-failed
spec:
  template:
    spec:
      terminationGracePeriodSeconds: 1
      containers:
        - name: never
          image: "busybox:latest"
          args:
            - "/bin/sh"
            - "-c"
            - "false"
          imagePullPolicy: Always
      restartPolicy: Never
  backoffLimit: 0
{{ end -}}
```

Upon installation, the ConfigMap is created as expected. Upon upgrade, the ConfigMap would _eventually_ be removed, but as the pre-upgrade hook fails, Helm never gets a chance to remove the ConfigMap.

When it comes time to rollback (either via `helm rollback` or via `--atomic` used during the failing `helm upgrade`) Helm walks the list of manifests in the target (the initial deployment which contained only the ConfigMap), finds that it exists in the cluster *and then checks its state in the last deployed version* where it is absent and fails with:

```
rollback.go:65: 2024-11-12 10:01:59.033172463 +0100 CET m=+0.253517049 [debug] preparing rollback of test
rollback.go:131: 2024-11-12 10:01:59.180766777 +0100 CET m=+0.401111391 [debug] rolling back test (current: v2, target: v1)
rollback.go:72: 2024-11-12 10:01:59.244344963 +0100 CET m=+0.464689544 [debug] creating rolled back release for test
rollback.go:78: 2024-11-12 10:01:59.391610987 +0100 CET m=+0.611955586 [debug] performing rollback of test
client.go:393: 2024-11-12 10:01:59.538573362 +0100 CET m=+0.758917920 [debug] checking 1 resources for changes
rollback.go:195: 2024-11-12 10:01:59.60209464 +0100 CET m=+0.822439231 [debug] warning: Rollback "test" failed: no ConfigMap with the name "cfg" found
Error: no ConfigMap with the name "cfg" found
helm.go:86: 2024-11-12 10:01:59.753177208 +0100 CET m=+0.973521792 [debug] no ConfigMap with the name "cfg" found
```

The error message is somewhat misleading as it suggests that the ConfigMap is not found _on the cluster_, where it is clearly there. Where it is not found is the Secret that represents that last deployment where it really was absent (or would have been, had Helm ever made it that far). The log message is produced here: 

https://github.com/helm/helm/blob/5a5449dc42be07001fd5771d56429132984ab3ab/pkg/kube/client.go#L418-L422

Has the manifest been successfully removed in the preceding upgrade, this wouldn't fail due to 

https://github.com/helm/helm/blob/5a5449dc42be07001fd5771d56429132984ab3ab/pkg/kube/client.go#L399-L411

Line 401 shows care given to a Manifest being absent on the cluster, but no similar logic for a Manifest being absent from the release deemed to be the original. Of course, if `helm upgrade` is used instead of `helm rollback`, this specific problem doesn't occur, but only because the original release is determined to be the last _successful_ one instead of the live one. Had the failure occurred at post-upgrade instead and we were rolling back more than one revision, the same issue would present itself.

Just like how Helm does not fail on a Manifest having been deleted on the cluster and instead quickly recreates it, if a Manifest is not found in the original release but does find it on the cluster, I think Helm should handle that gracefully.

Otherwise, `--atomic` and `helm rollback` are broken in the presence of removed Manifests and potentially failing pre-upgrade hooks.

Possibly duplicates #12388, but I think this way to reproduce the issue is clearer.

fmuyassarov:
Updated logic to check if a resource exists in the cluster.

Attempt to fix: https://github.com/helm/helm/issues/13437 

**If applicable**:

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -427,17 +427,26 @@ func (c *Client) update(original, target ResourceList, force, threeWayMerge bool
 		}
 
 		originalInfo := original.Get(info)
-		if originalInfo == nil {
-			kind := info.Mapping.GroupVersionKind.Kind
-			return errors.Errorf("no %s with the name %q found", kind, info.Name)
-		}
-
-		if err := updateResource(c, info, originalInfo.Object, force, threeWayMerge); err != nil {
-			slog.Debug("error updating the resource", "namespace", info.Namespace, "name", info.Name, "kind", info.Mapping.GroupVersionKind.Kind, slog.Any("error", err))
-			updateErrors = append(updateErrors, err.Error())
+		if originalInfo != nil {
+			// Proceed to update the resource
+			if err := updateResource(c, info, originalInfo.Object, force, threeWayMerge); err != nil {
+				slog.Debug("Error updating the resource %q: %v", info.Name, err)
+				updateErrors = append(updateErrors, fmt.Sprintf("failed to update resource %q: %v", info.Name, err))
+			}
+		} else {
+			// If the original resource is not found, force update the resource
+			if err := updateResource(c, info, nil, true, threeWayMerge); err != nil {
+				slog.Debug("Error force updating the resource %q: %v", info.Name, err)
+				updateErrors = append(updateErrors, fmt.Sprintf("failed to force update resource %q: %v", info.Name, err))
+			}
 		}
 		// Because we check for errors later, append the info regardless
 		res.Updated = append(res.Updated, info)
+
+		kind := info.Mapping.GroupVersionKind.Kind
+		if _, err := helper.Get(info.Namespace, info.Name); err != nil && !apierrors.IsNotFound(err) {
+			return errors.Errorf("no %s with the name %q found in namespace %q", kind, info.Name, info.Namespace)
+		}
 
 		return nil
 	})
@@ -672,23 +681,31 @@ func createPatch(target *resource.Info, current runtime.Object, threeWayMergeFor
 
 func updateResource(_ *Client, target *resource.Info, currentObj runtime.Object, force, threeWayMergeForUnstructured bool) error {
 	var (
-		obj    runtime.Object
-		helper = resource.NewHelper(target.Client, target.Mapping).WithFieldManager(getManagedFieldsManager())
-		kind   = target.Mapping.GroupVersionKind.Kind
+		obj       runtime.Object
+		helper    = resource.NewHelper(target.Client, target.Mapping).WithFieldManager(getManagedFieldsManager())
+		kind      = target.Mapping.GroupVersionKind.Kind
+		patch     []byte
+		patchType types.PatchType
+		err       error
 	)
 
 	// if --force is applied, attempt to replace the existing resource with the new object.
 	if force {
-		var err error
 		obj, err = helper.Replace(target.Namespace, target.Name, true, target.Object)
 		if err != nil {
 			return errors.Wrap(err, "failed to replace object")
 		}
 		slog.Debug("replace succeeded", "name", target.Name, "initialKind", currentObj.GetObjectKind().GroupVersionKind().Kind, "kind", kind)
 	} else {
-		patch, patchType, err := createPatch(target, currentObj, threeWayMergeForUnstructured)
-		if err != nil {
-			return errors.Wrap(err, "failed to create patch")
+		if currentObj != nil {
+			patch, patchType, err = createPatch(target, currentObj, threeWayMergeForUnstructured)
+			if err != nil {
+				return errors.Wrap(err, "failed to create patch")
+			}
+		} else {
+			// If currentObj is nil, treat it as a full replacement
+			patch = nil
+			patchType = types.MergePatchType
 		}
 
 		if patch == nil || string(patch) == "{}" {
--- a/pkg/kube/client_test.go
+++ b/pkg/kube/client_test.go
@@ -317,6 +317,8 @@ func testUpdate(t *testing.T, threeWayMerge bool) {
 		"/namespaces/default/pods/starfish:GET",
 		"/namespaces/default/pods/starfish:GET",
 		"/namespaces/default/pods/starfish:PATCH",
+		"/namespaces/default/pods/starfish:GET",
+		"/namespaces/default/pods/otter:GET",
 		"/namespaces/default/pods/otter:GET",
 		"/namespaces/default/pods/otter:GET",
 		"/namespaces/default/pods/otter:GET",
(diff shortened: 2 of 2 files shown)


fmuyassarov:
@mattfarina Do you see any potential consequences of this change? I tested it with `--atomic` upgrade and rollback, and it seems to help prevent false failures (e.g., when a resource does exist in the cluster).

matthiasvegh:
@fmuyassarov I've not yet been able to do testing on this yet, I'll be able to later this week. At first glance you're skipping the update of any manifests that are present in the target but not in the source if they also exist in the cluster.
While this is okay in the trivial case where an upgrade fails before removing a manifest not present in the target and then rolling back to where it is present, it fails for more involved situations.

As an example, if a Helm chart contains a TLS certificate as a Secret in revision 1, but in revision 2 has a cert-manager.io/v1 Certificate with the same name to indirectly create that Secret. The Secret is of course no longer part of the release but is still conceptually part of it as we expect the Cert Manager controller to create/update it. While the chart author may expect that Secret to work just as well in that chart, it's likely not exactly the same (if nothing else, helm.sh annotations are likely to be absent).

Skipping the update not only does not take you back to the previous state, any deviations will then be carried forward as out-of-band changes via 3-way merge.

robertsirc (changes requested):
Hello, first thank you for your pull request we do appreciate it. If we could ask can we please get some test around this change? I know there might not be current test but we are trying to ensure we don't break current functionality and up our test coverage.

fmuyassarov:
> Hello, first thank you for your pull request we do appreciate it. If we could ask can we please get some test around this change? I know there might not be current test but we are trying to ensure we don't break current functionality and up our test coverage.

Hi @robertsirc . Sure on it, will add a separate commit with tests.

fmuyassarov:
@robertsirc PTAL. Updated the PR as per review from @matthiasvegh and realized that there is unit test coverage already in place for the Update function which I'm touching on this patch and I just had to adjust that test function little bit. I also checked the coverage percentage to make sure that it doesn't go below what it is currently (63%).

matthiasvegh:
Hi @fmuyassarov, I've done some testing, and while the spurious failure is gone, there is still no update made to the resource absent in the source.

Given the chart in the original issue #13437, I did:

  1. `helm install test .`
  2. `helm upgrade test .` (which fails in pre-, as expected)
  3. `kubectl edit configmap/cfg` and changed `data.foo` to `baz` (was `bar`)
  4. `helm rollback test 1` this did not fail, but left `data.foo` as `baz`.

For testing purposes, I'd suggest tweaking the ConfigMap to update `data.foo` itself in revision 2 before being deleted in revision 3, then rollback directly to revision 1 instead of just to 2. As things stand, this change only works for `--rollback` (and when nobody else in the release takes over the manifest). For convenience, I've wrapped this up in a Chart below.
<details>
<summary>Updated Chart to show lack of an update</summary>

```yaml
{{- if lt .Release.Revision 3 -}}
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: cfg
data:
  foo: bar-from-{{ .Release.Revision | toString }}
{{ end -}}

{{- if ge .Release.Revision 3 -}}
---
apiVersion: batch/v1
kind: Job
metadata:
  name: "failurecauser"
  annotations:
    helm.sh/hook: pre-upgrade
    helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded,hook-failed
spec:
  template:
    spec:
      terminationGracePeriodSeconds: 1
      containers:
        - name: never
          image: "busybox:latest"
          args:
            - "/bin/sh"
            - "-c"
            - "false"
          imagePullPolicy: Always
      restartPolicy: Never
  backoffLimit: 0
{{ end -}}
```
</details>

With this chart, `helm install` results in `.data.foo` being `bar-from-r1`, `helm upgrade` results in `bar-from-r2`, the third `helm upgrade` fails, and a `helm rollback 1` results in `bar-from-r2` instead of `bar-from-r1`.

I'm uncertain on which direction to take for this fix. I can see two strategies being viable.
Either behave like `--force` by passing force=true and changing the log call to to not log the original `kind` in
https://github.com/helm/helm/blob/e3dffbd9ec4a54e98dc7517bc5f9145ba80ef07c/pkg/kube/client.go#L687 as I can't see a case where the kind changes, but maybe it depends on the `apiVersion`.
This results in the manifest being completely overwritten when reintroduced during a rollback. This seems like sane behavior, but gives Helm an edge-case where 3-way merge behaves differently to normal.

Alternatively, Helm could keep better track of which manifests were patched prior to the upgrade failing e.g., by adding a new annotation so you can compute a different source per manifest. This would result in Helm realizing that the manifest had not made it to the higher revision and use the older revision as the source, with 3-way merge preserving out-of-band changes as expected. I imagine this working by `Update` taking in _two_ previous releases, zipping them before iterating and determining the "original" based on whether they were ever actually patched. This seems like significant rework, and I don't want to recommend it as a mere user, I hope @robertsirc or @mattfarina can chime in.

fmuyassarov:
Hi @robertsirc,
I've reworked the patch to ensure that resources are updated correctly during a rollback. I verified this behavior using the example that @matthiasvegh pointed out earlier. One notable change I made after @matthiasvegh's tests and feedback is that I've enforced resource replacement during the `Update()` call by setting `force` flag to true.
Could you please take a look ?

robertsirc:
Hello, got time to look at this but your branch has some conflicts in it.

kashifest:
@robertsirc can we have some review this one? This is a valid use case it seems.

ruben-gp:
@fmuyassarov, could you please rebase and solve the conflicts, so @robertsirc can review it? As @kashifest stated, it's a valid UC, enabling several rollback/downgrade scenarios.
Thanks in advance.

fmuyassarov:
Let's continue this fix on a new PR: https://github.com/helm/helm/pull/31412

/close

PR #30712, opened 2025-03-31
cleanup: Remove extra lint/rules.Template functions

gjenkins8:
Simple cleanup to remove/collapse the three Template lint functions, into a single function with options:

```
func Templates(linter *support.Linter, values map[string]interface{}, namespace string, _ bool) {
func TemplatesWithKubeVersion(linter *support.Linter, values map[string]interface{}, namespace string, kubeVersion 
func TemplatesWithSkipSchemaValidation(linter *support.Linter, values map[string]interface{}, namespace string, kubeVersion *chartutil.KubeVersion, skipSchemaValidation bool)
```

Replaces with a single:
```
func Templates(linter *support.Linter, namespace string, values map[string]any, options ...TemplateLinterOption)
```
 
No functional changes, refactor/cleanup only

**If applicable**:

### DIFF
--- a/pkg/chart/v2/lint/rules/template.go
+++ b/pkg/chart/v2/lint/rules/template.go
@@ -42,35 +42,66 @@ import (
 )
 
 // Templates lints the templates in the Linter.
-func Templates(linter *support.Linter, values map[string]interface{}, namespace string, _ bool) {
-	TemplatesWithKubeVersion(linter, values, namespace, nil)
+func Templates(linter *support.Linter, namespace string, values map[string]any, options ...TemplateLinterOption) {
+	templateLinter := newTemplateLinter(linter, namespace, values, options...)
+	templateLinter.Lint()
 }
 
-// TemplatesWithKubeVersion lints the templates in the Linter, allowing to specify the kubernetes version.
-func TemplatesWithKubeVersion(linter *support.Linter, values map[string]interface{}, namespace string, kubeVersion *common.KubeVersion) {
-	TemplatesWithSkipSchemaValidation(linter, values, namespace, kubeVersion, false)
+type TemplateLinterOption func(*templateLinter)
+
+func TemplateLinterKubeVersion(kubeVersion *common.KubeVersion) TemplateLinterOption {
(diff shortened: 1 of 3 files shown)


benoittgt (commented):
Thanks. I find it clearer this way. Should we document the new public functions ?

pkg/chart/v2/lint/rules/template.go
```diff
@@ -121,9 +152,8 @@ func TemplatesWithSkipSchemaValidation(linter *support.Linter, values map[string
 	*/
 	for _, template := range chart.Templates {
 		fileName := template.Name
-		fpath = fileName
 
-		linter.RunLinterRule(support.ErrorSev, fpath, validateAllowedExtension(fileName))
+		t.linter.RunLinterRule(support.ErrorSev, fileName, validateAllowedExtension(fileName))
 
 		// We only apply the following lint rules to yaml files
 		if filepath.Ext(fileName) != ".yaml" || filepath.Ext(fileName) == ".yml" {
```
benoittgt, line 159: Also I'm curious this is not an issue. 🤔 For .yaml, left side is false, right side is also false, so it won’t continue.

It should be ?
```go
ext := strings.ToLower(filepath.Ext(fileName))
if ext != ".yaml" && ext != ".yml" {
    continue
}
```

gjenkins8: Yes, I think the AI is correct here (good job bot). The way this is written is strange and wrong.

Personally, I would write:
```
func isYamlFileExtension(fileName string) {
    ext := strings.ToLower(filepath.Ext(fileName))
    return ext == ".yaml" ||  ext == ".yml"
}

..

// We only apply the following lint rules to yaml files
if isYamlFileExtension(fileName) {
    ...
}
```

But IMHO, this is a new "bug fix" PR. This PR is a refactor of the existing code (and PRs IMHO should be single purpose)

benoittgt: https://github.com/helm/helm/pull/31376

mattfarina (approved):

Issue #12138, opened 2023-06-12
Enhanced Flexibility and Extensibility: Support for Customizing Provided Template Functions in Helm


alexwo:
### Motivation

The motivation for supporting the customization of provided template functions in Helm stems from the need to enable greater flexibility and extensibility in managing Kubernetes deployments. Customizing template functions allows users to tailor the behavior of Helm's templating engine to their specific requirements, enhancing the capabilities and adaptability of Helm charts.

Adapting to Specific Use Cases: Different applications and deployment scenarios may require custom logic or transformations during the template rendering process. By allowing users to customize the provided template functions, Helm enables them to adapt the behavior of these functions to fit their specific use cases. This flexibility is crucial in accommodating unique requirements and business logic associated with complex deployments.

Extending Helm's Functionality: Helm's built-in template functions cover a wide range of common use cases, but every application has its own unique requirements. Allowing users to customize these functions empowers them to extend Helm's functionality to address specific needs. Users can implement custom functions that perform advanced calculations, apply complex filters, integrate with external APIs, or introduce new behavior to meet their specific deployment requirements.

### Helm as library
Helm is widely used as a package manager for Kubernetes, but it can also be leveraged as a library for programmatic use. This PR aims to enhance Helm's flexibility by enabling users to define and configure their own custom template functions when using Helm programmatically.

### PR Propsal:
https://github.com/helm/helm/pull/12135

Thanks,
Alex

gjenkins8:
Apologies, I missed responding here. I think bring this up in the #helm-dev. The effort seems small (and non-intrusive). I think Helm maintainers would agree. 

My only comments would be:
- a post-renderer might also be suitable, and not require any in-built functionality
- Any chart which depends on a specialized function would only be usable from a (custom) SDK operator. And not from any standard Helm cli/SDK (worth considering the downsides here when it comes to development and usage)

gjenkins8:
(and thanks, I did see the PR. I didnt get a chance to look at it)

bennesp:
This is still valid and pretty useful at least to me. Can we have it look at by someone, please? Thank you very much

PR #30734, opened 2025-04-08
feat(pkg/engine): add support for custom template funcs

### LINKED ISSUE #30733: Add support for custom template funcs when using Helm as a library
Current Behavior:

When using Helm as a library, users have no mechanism to inject their own custom template functions into the templating engine. This limitation confines users to only the default built-in functions provided by Helm, which may not cover all advanced or specialized use cases.

Desired Behavior:

The opportunity to inject custom template function via an action config.

ipaqsa:
This PR extends the Helm templating engine and action configuration to allow users to inject custom template functions when using Helm as a library. It adds a new field to both the action configuration and the engine structures so that custom functions can be provided via an action config and merged with Helm’s default function set.

How to Use:

Users can configure their custom template functions in the action configuration as shown below:
```
conf := &action.Configuration{
	Capabilities: chartutil.DefaultCapabilities,
	CustomTemplateFuncs: template.FuncMap{
		"exclaim": func(input string) string {
			return input + "!!!"
		},
	},
}
```

With this configuration, the Helm templating engine will have access to the custom exclaim function during rendering.

The problem is described in the issue.

Closes #30733

There is an old [PR](https://github.com/helm/helm/pull/12135) about this.
And an [old issue](https://github.com/helm/helm/issues/12138)

It can also help to solve [this](https://github.com/helm/community/issues/340)

**If applicable**:

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -196,6 +196,11 @@ func (e Engine) initFunMap(t *template.Template) {
 	funcMap := funcMap()
 	includedNames := make(map[string]int)
 
+	// Set custom template funcs
+	for k, v := range e.CustomTemplateFuncs {
+		funcMap[k] = v
+	}
+
 	// Add the template-rendering functions here so we can close over t.
 	funcMap["include"] = includeFun(t, includedNames)
 	funcMap["tpl"] = tplFun(t, includedNames, e.Strict)
@@ -244,11 +249,6 @@ func (e Engine) initFunMap(t *template.Template) {
 		funcMap["getHostByName"] = func(_ string) string {
 			return ""
 		}
-	}
-
-	// Set custom template func, do it here to give opportunity to override standard functions
-	for k, v := range e.CustomTemplateFuncs {
-		funcMap[k] = v
 	}
 
 	t.Funcs(funcMap)
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -44,6 +44,8 @@ type Engine struct {
 	clientProvider *ClientProvider
 	// EnableDNS tells the engine to allow DNS lookups when rendering templates
 	EnableDNS bool
+	// CustomTemplateFuncs is defined by users to provide custom template funcs
+	CustomTemplateFuncs template.FuncMap
 }
 
 // New creates a new instance of Engine using the passed in rest config.
@@ -242,6 +244,11 @@ func (e Engine) initFunMap(t *template.Template) {
 		funcMap["getHostByName"] = func(_ string) string {
 			return ""
 		}
+	}
+
+	// Set custom template func, do it here to give opportunity to override standard functions
+	for k, v := range e.CustomTemplateFuncs {
+		funcMap[k] = v
 	}
 
 	t.Funcs(funcMap)
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -196,11 +196,6 @@ func (e Engine) initFunMap(t *template.Template) {
 	funcMap := funcMap()
 	includedNames := make(map[string]int)
 
-	// Set custom template funcs
-	for k, v := range e.CustomTemplateFuncs {
-		funcMap[k] = v
-	}
-
 	// Add the template-rendering functions here so we can close over t.
 	funcMap["include"] = includeFun(t, includedNames)
 	funcMap["tpl"] = tplFun(t, includedNames, e.Strict)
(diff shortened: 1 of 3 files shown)


joejulian (changes requested):

pkg/engine/engine_test.go
```diff
@@ -1300,3 +1300,49 @@ func TestRenderTplMissingKeyString(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	// Expected output should be "hello!!!".
+	expected := "hello!!!"
+	key := "CustomFunc/templates/manifest"
+	if rendered, ok := out[key]; !ok || rendered != expected {
+		t.Errorf("Expected %q, got %q", expected, rendered)
+	}
+}
```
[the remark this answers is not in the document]
ipaqsa: Done!

juev:
➕ 🔟 these changes are simply 🔥

TerryHowe (approved):
/lgtm

scottrigby:
hi @ipaqsa thanks for updating this. I'm curious, what did you find lacking in this PR #12135?

ipaqsa:
hi! That PR adds custom funcs only for the engine, it cannot be used the action config(in our project we dont use separate engine instance) and I thought it was stale(the PR was created 2 years ago)

joejulian (dismissed):
LGTM, Thanks!

scottrigby (dismissed):
Valid use case for SDK users 👍 LGTM

pkg/engine/engine.go
```diff
@@ -194,6 +196,11 @@ func (e Engine) initFunMap(t *template.Template) {
 	funcMap := funcMap()
 	includedNames := make(map[string]int)
 
+	// Set custom template funcs
+	for k, v := range e.CustomTemplateFuncs {
+		funcMap[k] = v
```
gjenkins8, line 201: This will not allow overriding existing functions (a potential use case), only adding new functions. Do we want to take the approach #12135 takes?:

https://github.com/helm/helm/pull/12135/files#diff-9a4b0f64963e54937214c78c91c3d404eaef74b571e89f370ed7de3c9ff86ca1R215

ipaqsa:
one moment, i missed 1 detail

pkg/engine/engine.go
ipaqsa: It allows to override functions created by the "funcMap" function(check test case, it overrides the "upper" function), overriding others embedded funcs like tmpl, include and etc seems wrong

gjenkins8: (and on the OTOH, if we don't want to override these functions, right now we have a awkward interface where the user will attempt, and Helm will silently fail. User would literally have to come read the source code to understand the problem)

gjenkins8: ie. if we don't want to allow overriding  tmpl, include, etc. We should add an error if the user attempts to do so (rather than silently failing)

ipaqsa: Ok, I agree. I will move below to allow override all functions

gjenkins8 (approved):

ipaqsa:
why has it failed ? Should I rebase ?

sabre1041 (approved):
LGTM

scottrigby (approved):

DarrenBishop:
I see that this feature was raised in issue #30733, but was not sure whether to ask here (on the issue) or there (on the PR), so asking on both...

My question is, what is the reason this was not introduced as a general 'plugin', that is, why the limited use-case(s) of using Helm as a library?

The [current (v4) release overview](https://helm.sh/docs/overview/#custom-template-functions) is misleading when it states this capability is available via plugins - I only see CLI, Getter and PostRenderer plugins currently, so what gives?
Is this factual - and I have just missed the _how-to_, yet found this issue/PR - or is it indeed incorrect?

The hope is/was that as a plugin, we still have access to the `subprocess` runtime, which the lay-user can be immediately productive with, _while they learn Go for `extism` (Wasm) in the background_.

Irrespective of what is possible now vs in the future (i.e. as if there were a simple update to turn this into an additional `FuncMap` plugin type, say), what mitigations are there for function (name) shadowing? As in:
 - Is it really obvious (to the user) when multiple custom functions (maps) are loaded(?) introducing functions with the same name?
 - Is this an error i.e. panic
 - Is the last-one to be loaded the winner (i.e. the one that gets used during templating)... or the first-one?

PR #30743, opened 2025-04-11
Allow chart and file size limits to be configurable through environment variables and cmd flags

### LINKED ISSUE #30738: MaxDecompressedFileSize in v3.17.3 breaks existing charts
```
```

Produces the following error:
```
```

benoittgt:

Following security fix e4da497 (size limitations), implement env var configuration or command flags (for pull, install, upgrade commands) to allow increasing max size thresholds. This maintains security by default while providing an override for legitimate use cases. Documentation should be updated accordingly. No absolute maximum ceiling appears necessary.

- Would it make sense to manage this via a cmd flag instead ? ([source](https://github.com/helm/helm/issues/30738#issuecomment-2796489756))

Fix: [helm/helm#30738](https://github.com/helm/helm/issues/30738)

Special notes for your reviewer:

It has been tested with commands like :
```sh
# No unit,  upgrade command
$ bin/helm upgrade my-release oci://ghcr.io/stefanprodan/charts/podinfo --max-file-size=1 
Pulled: ghcr.io/stefanprodan/charts/podinfo:6.9.2
Digest: sha256:971fef0d04d5b3d03d035701dad59411ea0f60e28d16190f02469ddfe5587588
Error: unable to load chart archive: decompressed chart file "podinfo/Chart.yaml" is larger than the maximum file size 1

# No unit upgrade command on chart file size
$ bin/helm upgrade my-release oci://ghcr.io/stefanprodan/charts/podinfo --max-chart-size=1
Pulled: ghcr.io/stefanprodan/charts/podinfo:6.9.2
Digest: sha256:971fef0d04d5b3d03d035701dad59411ea0f60e28d16190f02469ddfe5587588
Error: unable to load chart archive: decompressed chart is larger than the maximum size 1

# Install with Kubernetes Quantity
$ bin/helm install my-release oci://ghcr.io/stefanprodan/charts/podinfo --max-chart-size=1Ki
Pulled: ghcr.io/stefanprodan/charts/podinfo:6.9.2
Digest: sha256:971fef0d04d5b3d03d035701dad59411ea0f60e28d16190f02469ddfe5587588
Error: INSTALLATION FAILED: unable to load chart archive: decompressed chart is larger than the maximum size 1Ki

# Env var
$ HELM_MAX_FILE_SIZE=1 bin/helm pull oci://registry-1.docker.io/bitnamicharts/mysql --untar
Pulled: registry-1.docker.io/bitnamicharts/mysql:14.0.3
Digest: sha256:2bbaf66e43dad7edce1eb563e19bcabd83eb8c8f79549505ad1c48e3a0cbfbba
Error: decompressed chart file "mysql/Chart.yaml" is larger than the maximum file size 1

# Invalid Kubernetes Quantity
$ HELM_MAX_CHART_SIZE=1ki bin/helm pull oci://registry-1.docker.io/bitnamicharts/mysql --untar
2025/11/06 14:58:11 WARN invalid value "1ki" (expected int or k8s Quantity like 512Mi): using default 100Mi
Pulled: registry-1.docker.io/bitnamicharts/mysql:14.0.3
Digest: sha256:2bbaf66e43dad7edce1eb563e19bcabd83eb8c8f79549505ad1c48e3a0cbfbba

# Prevent using "m"
HELM_MAX_CHART_SIZE=1m bin/helm pull oci://registry-1.docker.io/bitnamicharts/mysql --untar

2025/11/10 19:54:42 WARN quantity "1m" uses 'm' (milli) suffix which represents 0.001; please use IEC values like Ki, Mi, Gi: using default 100Mi
Pulled: registry-1.docker.io/bitnamicharts/mysql:14.0.3
Digest: sha256:2bbaf66e43dad7edce1eb563e19bcabd83eb8c8f79549505ad1c48e3a0cbfbba
```

I've cherry picked this PR [helm/helm#31464](https://github.com/helm/helm/pull/31464)

If applicable:

 - [x] this PR contains user facing changes (the docs needed label should be applied if so)


pkg/chart/loader/archive.go
```diff
@@ -233,3 +241,18 @@ func LoadArchive(in io.Reader) (*chart.Chart, error) {
 
 	return LoadFiles(files)
 }
+
+// ParseEnvSettings reads environment variables and updates the size limits
+func parseEnvSettings() {
+	if chartSizeStr := os.Getenv("HELM_MAX_DECOMPRESSED_CHART_SIZE"); chartSizeStr != "" {
```
benoittgt, line 247: I'm not sure if switching to cmd flag is a good idea. I see those settings as something exceptionnal and a global setup of helm commands.
Also loader package is pretty exempt from any cmd flags and any configuration values. What do you think ?

birkland: Maybe outside the scope of concern, but I believe argo uses the helm executable to template charts:
https://github.com/argoproj/argo-cd/blob/cde5b4b7d40683d84df98994e67c4cbd9b247255/util/helm/helm.go#L69-L75

With a cli arg, it should be possible to set this on a per-chart basis if needed, I think.  With an env var, it's kind of all or nothing unless argo enhances its API to expose env vars to the helm process when templating

robertsirc:
I think this is on the wrong branch, you are targeting `dev-v3` and if we were going to do this type of change we would want this on part of `main`.

scottrigby:
> I think this is on the wrong branch, you are targeting `dev-v3` and if we were going to do this type of change we would want this on part of `main`.

Exactly. @benoittgt please 
1. change the base branch in this PR to `main` https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-base-branch-of-a-pull-request
2. after the PR is accepted and merged into `main`, you may backport a fix to `dev-v3` (can reopen a new PR against this branch using the same branch from your fork). But please don't do this until a PR has been approved for main

benoittgt:
Oups sorry. It should be ok now @scottrigby @robertsirc

pkg/chart/v2/loader/archive.go
```diff
@@ -233,3 +241,18 @@ func LoadArchive(in io.Reader) (*chart.Chart, error) {
...
+			MaxDecompressedChartSize = size
+		}
+	}
+
+	if fileSizeStr := os.Getenv("HELM_MAX_DECOMPRESSED_FILE_SIZE"); fileSizeStr != "" {
+		if size, err := strconv.ParseInt(fileSizeStr, 10, 64); err == nil {
+			MaxDecompressedFileSize = size
+		}
+	}
+}
```
mattfarina, line 258: Helm is divided into an SDK and a CLI/Client. The SDK is used by a wide array of applications (e.g., Flux - a sister project in the CNCF). SDK users have a method to change this value via public package level variables.

To expose this in the CLI/Client, the code for parsing and using these variables should be there and then set the package level variables. This is to have a clean separation.

The Environment Variables should be set in _pkg/cmd/root.go_. This file also includes the right place to document them alongside the other environment variables.

mattfarina (dismissed):

pkg/chart/v2/loader/archive.go
benoittgt: Thanks @mattfarina , I will do the change.

benoittgt:
I have some reservations about the current approach in this PR, @mattfarina. Using package-level variables in the loader package creates limitations for advanced SDK users who need to configure size limits on a per-operation basis.

I'm also questioning whether exposing these settings as command flags is the right approach. While I initially considered adding shared command flags, these maximum size settings are quite specialized and only relevant to a small subset of commands.

Curious about what you think about this proposal. Thanks in advance for your feedbacks.

pkg/cli/environment.go
```diff
@@ -109,6 +113,8 @@ func New() *EnvSettings {
 		RepositoryCache:           envOr("HELM_REPOSITORY_CACHE", helmpath.CachePath("repository")),
 		BurstLimit:                envIntOr("HELM_BURST_LIMIT", defaultBurstLimit),
 		QPS:                       envFloat32Or("HELM_QPS", defaultQPS),
+		MaxChartSize:              envInt64Or("HELM_MAX_CHART_SIZE", 0),
+		MaxFileSize:               envInt64Or("HELM_MAX_FILE_SIZE", 0),
```
gjenkins8, line 117:
```suggestion
		MaxChartFileSize:               envInt64Or("HELM_MAX_CHART_FILE_SIZE", 0),
```

IMHO, terminology (here and other places) should use the more specific "max chart file size". Which is what this functionality applies to. "max file size" would seem to apply to any file.

pkg/cmd/install.go
```diff
@@ -255,6 +257,12 @@ func runInstall(args []string, client *action.Install, valueOpts *values.Options
 		return nil, err
 	}
 
+	if client.MaxChartSize > 0 {
+		loader.MaxDecompressedChartSize = client.MaxChartSize
+	}
+	if client.MaxFileSize > 0 {
+		loader.MaxDecompressedFileSize = client.MaxFileSize
```
gjenkins8, line 264: SDK users can run multiple Helm releases concurrently, this will introduce a data race.

pkg/cli/environment.go
gjenkins8: That said, for Helm 4 (the `main` branch). We can probably remove this "max chart file size" functionality all together. It was originally introduced to prevent https://github.com/helm/helm/security/advisories/GHSA-5xqw-8hwv-wg92 . But https://github.com/helm/helm/pull/30754 removed the usage of (the vulnerable) `github.com/xeipuuv/gojsonschema` library for JSON schema parsing.

I would propose continuing with the flag in this PR, so we can easily backport (via cherry-pick style PR to `dev-v3` branch). Then we can look to remove its usage for Helm 4.

pkg/cli/environment.go
```diff
@@ -109,6 +113,8 @@ func New() *EnvSettings {
 		RepositoryCache:           envOr("HELM_REPOSITORY_CACHE", helmpath.CachePath("repository")),
 		BurstLimit:                envIntOr("HELM_BURST_LIMIT", defaultBurstLimit),
 		QPS:                       envFloat32Or("HELM_QPS", defaultQPS),
+		MaxChartSize:              envInt64Or("HELM_MAX_CHART_SIZE", 0),
```
gjenkins8, line 116:
```suggestion
		MaxChartSize:              envInt64Or("HELM_MAX_CHART_SIZE", 100 * 1024 * 1024), // Default 100 MiB
```

Lets put the defaults into the defaults :) 

Then `helm env` can print them, and users don't have to go hunt for the "actual default".

gjenkins8 (dismissed):
Few comments/concerns to tidy up I think. In particular, the usage of the package level globals won't work for concurrent SDK users.

pkg/action/install.go
```diff
@@ -116,6 +116,10 @@ type Install struct {
 	PostRenderer  postrender.PostRenderer
 	// Lock to control raceconditions when the process receives a SIGTERM
 	Lock sync.Mutex
+	// MaxChartSize is the maximum size of a decompressed chart in bytes
+	MaxChartSize int64
+	// MaxFileSize is the maximum size of a single file in a chart in bytes
```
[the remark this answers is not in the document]
benoittgt: But if I read correctly the code, it means only env vars then? 🤔 Because I use this in for comand flags
```go
f.Int64Var(&client.MaxChartSize, "max-chart-size", settings.MaxChartSize, "maximum size in bytes for a decompressed chart (default is 100mb)")
```

pkg/cli/environment.go
```diff
@@ -205,6 +211,19 @@ func envFloat32Or(name string, def float32) float32 {
 	return float32(ret)
 }
 
+// We want to handle int64 like returned by https://pkg.go.dev/io/fs#FileInfo
+func envInt64Or(name string, def int64) int64 {
+	if name == "" {
+		return def
+	}
+	envVal := envOr(name, strconv.FormatInt(def, 10))
+	ret, err := strconv.ParseInt(envVal, 10, 64)
```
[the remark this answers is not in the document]
benoittgt: I would be interested to add this in another PR, except if you prefer to do all in one and add this two. I think it could be a great addition.

benoittgt:
@gjenkins8 I took a different approach:
- Preserved the existing package functionality, only affecting references to max sizes
- Added support for custom max chart and file sizes via command flags and environment variables
- Eliminated global variable definitions for max sizes in the loader package

I've probably missed a few things. The Helm codebase and Go in general are fairly new to me. Please feel free to provide any feedback - I appreciate your reviews.

benoittgt:
Hello

This fix will probably lands into Helm v4 (?) In the meantime I can still work on it if you have time for a review @gjenkins8. Thanks

benoittgt:
I can do one last rebase if there’s still interest in this. Just let me know!

nkawli:
@benoittgt @mattfarina  @birkland @gjenkins8 We are stuck with 3.17.2 release because of this issue as we are exceeding the limits. Would like to upgrade to new helm release but cannot go beyond 3.17.2 due to this issue. Can this be prioritised ?

acastong:
As security vulnerabilities accumulate in dependent libraries used to build 3.17.2, it will become more and more critical to have a better alternative to staying on that old version.  As of October 28th 2025, Trivy is reporting the following High vulnerabilities (and 6 other lower ones): 
* CVE-2025-22868 - golang.org/x/oauth2/jws: Unexpected memory consumption during token parsing in golang.org/x/oauth2/jws
* CVE-2025-47907 - database/sql: Postgres Scan Race Condition

benoittgt:
The rebase was almost a complete rewrite. Dealing with multiple loaders is not the easiest.

I’d appreciate a review @sabre1041 and @gjenkins8.

pkg/cli/environment.go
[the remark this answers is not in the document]
benoittgt: I’ve finally added it.

benoittgt:
I’m really sorry for the size of the PR. With the duplication of size checks and the handling of Kubernetes Quantity, it adds a lot of lines.

sabre1041 (commented):
A couple of notes:

1. This feature should also be added to `helm template` as the same type of logic applies and would be needed by tools, such as Argo CD
2. Thoughts about enabling an opt out enforcement of Chart size? Especially as a consumer of charts and not having the ability to easily know whether file or chart size may violate the limits. Keeping in mind the potential security concerns that precipitated this enhancement around enforcing max sizes

benoittgt:
Thanks @sabre1041, and good catch on the missing part. I’ll take a look at it.

Regarding your question about enabling an opt-out enforcement of chart size: isn’t it a good idea to always set a limit so that chart sizes aren’t unlimited like before? For example, a project might decide it feels safe to run helm commands with a chart size limit of 100Mi.

sabre1041:
> Thanks @sabre1041, and good catch on the missing part. I’ll take a look at it.
> 
> Regarding your question about enabling an opt-out enforcement of chart size: isn’t it a good idea to always set a limit so that chart sizes aren’t unlimited like before? For example, a project might decide it feels safe to run helm commands with a chart size limit of 100Mi.

I agree that there should always be a default value as it employs secure practices. I fear that some end users will set some arbitrary high value just to avoid hitting the max size limit. This comment was intended to spur conversation. I am fine with leaving the functionality as it currently stands. If there was a desire to disable enforcement of size limits, it could be employed either through an environment variable (such as `HELM_DISABLE_MAX_SIZE_LIMITS` or by setting `-1` to either of the options).

benoittgt:
@sabre1041 I've implemented the checks on template command.

sabre1041 (dismissed):
LGTM

benoittgt:
@GeorgeJ mentioned last week during the dev meeting that the original security issue is probably no longer present, since we changed the JSON library. I don't know the attack vector of the previous. Version feel free to send it to me.

But we have 2 solutions to resolve #30738
1. Accept this PR
2. Revert https://github.com/helm/helm/commit/e4da49785aa6e6ee2b86efd5dd9e43400318262b

Diliz:
This issue was opened a long time ago, I can see a lgtm in november, what's missing for it to be finally merged?

Since this is a breaking change (it is breaking previously packaged helm charts which had this issue), I was thinking this to be resolved faster than this instead of waiting nearly a year for this to be adressed... (And having the possibibility to change the this kind of values via environment variables should be standard when implementing this kind of troublesome hard coded limitation).

Right now I'm stuck in helm v3.17.1 because I need to be able to render older third party helm charts, and I can't decide if it's ok for me or not...

Thanks to @benoittgt for the pull request, hope it will be merged at some point

benoittgt:
@Diliz do you think you could answer to the small request I have on the issue in https://github.com/helm/helm/issues/30738#issuecomment-3671467888 ? I'm curious about your situation.

joejulian:
As mentioned in in the thread at https://github.com/helm/helm/pull/31748#issuecomment-3892847061, I do not believe we should remove the limit. Making it configurable is the more desirable behavior.

cc @gjenkins8

joejulian:
> > Thanks @sabre1041, and good catch on the missing part. I’ll take a look at it.
> > Regarding your question about enabling an opt-out enforcement of chart size: isn’t it a good idea to always set a limit so that chart sizes aren’t unlimited like before? For example, a project might decide it feels safe to run helm commands with a chart size limit of 100Mi.
> 
> I agree that there should always be a default value as it employs secure practices. I fear that some end users will set some arbitrary high value just to avoid hitting the max size limit. This comment was intended to spur conversation. I am fine with leaving the functionality as it currently stands. If there was a desire to disable enforcement of size limits, it could be employed either through an environment variable (such as `HELM_DISABLE_MAX_SIZE_LIMITS` or by setting `-1` to either of the options).

You can't always keep users from defeating tools that are there to protect them. They could also just remove the check from the source and build their own binary.

We should have reasonable defaults as determined by this project's understanding of the use of objects in Kubernetes and user feedback. If a user doesn't feel that is appropriate for their own use case and, rather than picking a value that is reasonable for them, chooses to do something unreasonable - there's not much we can or should do to stop them.

joejulian:
I'm going to dismiss the really stale change-request reviews so we can move forward fresh from this rewrite.

PR #30749, opened 2025-04-11
Removing the alpine test chart

mattfarina:
A .gitignore was previously setup to ignore this file. When pkg/cmd was setup the .gitignore was not updated. The change adds the new location to continue to ignore this file.

Note, the previous location is still included in the .gitignore because developers will have a file there and we do not want that accidently included in a commit.

**If applicable**:


robertsirc (approved):
LGTM

PR #30754, opened 2025-04-14
Simplify the JSON Schema checking

mattfarina:
A new library was introduced that provides JSON Schema checking for newer versions of the schema. In Helm v4, there is no need to have two packages doing the JSON schema validation. The message output can have breaking changes.

This change moves everything to the newer library. It also uses a wrapper error to enable a clean Helm only interface for the public Go API validation functions. This would enable the replacement of the Schema validation library, if needed, without breaking the Go API contract.

**If applicable**:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -156,7 +156,7 @@ format: $(GOIMPORTS)
 # Generate golden files used in unit tests
 .PHONY: gen-test-golden
 gen-test-golden:
-gen-test-golden: PKG = ./cmd/helm ./pkg/action
+gen-test-golden: PKG = ./pkg/cmd ./pkg/action
 gen-test-golden: TESTFLAGS = -update
 gen-test-golden: test-unit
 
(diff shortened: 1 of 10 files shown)


Makefile
```diff
@@ -156,7 +156,7 @@ format: $(GOIMPORTS)
 # Generate golden files used in unit tests
 .PHONY: gen-test-golden
 gen-test-golden:
-gen-test-golden: PKG = ./cmd/helm ./pkg/action
+gen-test-golden: PKG = ./pkg/cmd ./pkg/action
```
mattfarina, line 159: Note, this change should have happened when the cmd functionality was moved to pkg so that it could be reused. This is part of a make script that can update golden files en masse.

robertsirc (approved):
LGTM

It isn't for this PR but we should look to add some debug logging here as well.

PR #30760, opened 2025-04-16
adding slog debug to a few points

robertsirc:
Adding in some debug calls for jsonschema.

**If applicable**:

### DIFF
--- a/pkg/chart/v2/util/jsonschema.go
+++ b/pkg/chart/v2/util/jsonschema.go
@@ -19,6 +19,7 @@ package util
 import (
 	"bytes"
 	"fmt"
+	"log/slog"
 	"strings"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


TerryHowe (approved):
/lgtm 

I do wonder what the validator output looks like though

pkg/chart/v2/util/jsonschema.go
```diff
@@ -31,13 +32,14 @@ import (
 func ValidateAgainstSchema(chrt *chart.Chart, values map[string]interface{}) error {
 	var sb strings.Builder
 	if chrt.Schema != nil {
+		slog.Debug("chart name", "chart-name", chrt.Name)
```
mattfarina, line 35: I did a test of this and `chrt.Name` came out as `0x1031fed60`. This was not the chart name. This is a reference to it, I think. In any case, this needs to be fixed.

pkg/chart/v2/util/jsonschema.go
```diff
@@ -67,6 +69,7 @@ func ValidateAgainstSingleSchema(values Values, schemaJSON []byte) (reterr error
 	if err != nil {
 		return err
 	}
+	slog.Debug("unmarshalled JSON schema", "schema", schema)
```
mattfarina, line 72: Did you want to output the JSON or the parsed Go object (struct)? This is the struct being printed. If you want the JSON itself, that would be `schemaJSON`.

pkg/chart/v2/util/jsonschema.go
```diff
@@ -78,6 +81,7 @@ func ValidateAgainstSingleSchema(values Values, schemaJSON []byte) (reterr error
 	if err != nil {
 		return err
 	}
+	slog.Debug("validated JSON schema", "validator", validator)
```
mattfarina, line 84: `validator` is not JSON but a Go object. What do you want to display here?

mattfarina (changes requested):

pkg/chart/v2/util/jsonschema.go
robertsirc: Saw what I did here fixing it now.

mattfarina (approved):

PR #30766, opened 2025-04-17
Fix main branch by defining wait strategy parameter on hooks

benoittgt:
**What this PR does / why we need it**:

https://github.com/helm/helm/pull/30673 's merge broke `main` because of changes on wait strategy, `Log` parameter and different testing strategy. This PR try to fix linting and testing.

**Special notes for your reviewer**:

**If applicable**:
- [ ] this PR contains user facing changes (the `docs needed` label should be applied if so)
- [ ] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

PR #30767, opened 2025-04-17
Allow using a different value for the app.kubernetes.io/managed-by label

luksa:
Allows custom operators using the Helm SDK to override the default `app.kubernetes.io/managed-by: Helm` label value. If an operator uses Helm to deploy resources, the resources are not managed by Helm but by the operator, so it would be great if this fact could be represented in the `managed-by` label. Without this feature, the operator would have to introduce another `managed-by` label, which could be confusing.

**If applicable**:

### DIFF
--- a/pkg/action/validate.go
+++ b/pkg/action/validate.go
@@ -32,10 +32,13 @@ var accessor = meta.NewAccessor()
 
 const (
 	appManagedByLabel              = "app.kubernetes.io/managed-by"
-	appManagedByHelm               = "Helm"
 	helmReleaseNameAnnotation      = "meta.helm.sh/release-name"
 	helmReleaseNamespaceAnnotation = "meta.helm.sh/release-namespace"
 )
+
+// AppManagedByLabel is the value Helm should use in the app.kubernetes.io/managed-by label to
+// identify resources it manages. Defaults to "Helm", but can be overridden via this variable.
+var AppManagedByValue = "Helm"
 
 // requireAdoption returns the subset of resources that already exist in the cluster.
 func requireAdoption(resources kube.ResourceList) (kube.ResourceList, error) {
@@ -102,7 +105,7 @@ func checkOwnership(obj runtime.Object, releaseName, releaseNamespace string) er
 	}
 
 	var errs []error
-	if err := requireValue(lbls, appManagedByLabel, appManagedByHelm); err != nil {
+	if err := requireValue(lbls, appManagedByLabel, AppManagedByValue); err != nil {
 		errs = append(errs, fmt.Errorf("label validation error: %s", err))
 	}
 	if err := requireValue(annos, helmReleaseNameAnnotation, releaseName); err != nil {
@@ -150,7 +153,7 @@ func setMetadataVisitor(releaseName, releaseNamespace string, force bool) resour
 		}
 
 		if err := mergeLabels(info.Object, map[string]string{
-			appManagedByLabel: appManagedByHelm,
+			appManagedByLabel: AppManagedByValue,
 		}); err != nil {
 			return fmt.Errorf(
 				"%s labels could not be updated: %s",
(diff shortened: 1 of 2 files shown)


robertsirc:
Hello @luksa first thank you for your PR. You have some build issues if you can please address those we can then review and approve.

luksa:
@robertsirc Thanks, but those failures are not related to this PR. Please check https://github.com/helm/helm/pull/30766.

robertsirc:
@luksa thanks for the catch we will check that out today.

scottrigby:
@luksa thanks for getting this PR review ready.

My main concern with this idea is that helm releases installed by an operator that sets a different value for `app.kubernetes.io/managed-by` would then lock out users from managing those releases with the Helm CLI during a break glass scenario (for incident management etc). This is a big tradeoff that SDK users would need to willingly take on, and the current code—if we were to accept such a change to have Helm allow this option—would need a code comment making very clear this trade off. 

However, I'm wondering, is your goal to allow your operator to manage the helm release exclusively? I'm wondering OTOH if your goal is perhaps to just have a way to denote which Helm releases your operator installed, and not to prevent other tools from also being able to manage those releases (the above tradeoff). Can you clarify which of these it is?

Note some technical details about the tradeoff I mentioned— Helm actions currently inconsistently rely on this label today: 
- actions like install, upgrade, and rollback do
- actions like list and uninstall do not

Summary:
- if your goal is to denote that your operator has managed the Helm release, you may wish to choose a different label and leave the common `app.kubernetes.io/managed-by: Helm` label as-is
- if your goal is to take exclusive ownership of certain helm releases (ie, prevent other tools such as the Helm CLI or others that leverage the Helm SDK from interacting with them) then this would require additional changes than what this PR currently does, and may be best to write a HIP to make sure this ideas is fully thought through, all bases covered, this option is well-docmented, doesn't raise other issues when not used, etc

scottrigby (commented):
@luksa thanks for getting this PR review ready.

My main concern with this idea is that helm releases installed by an operator that sets a different value for `app.kubernetes.io/managed-by` would then lock out users from managing those releases with the Helm CLI during a break glass scenario (for incident management etc). This is a big tradeoff that SDK users would need to willingly take on, and the current code—if we were to accept such a change to have Helm allow this option—would need a code comment making very clear this trade off. 

However, I'm wondering, is your goal to allow your operator to manage the helm release exclusively? I'm wondering OTOH if your goal is perhaps to just have a way to denote which Helm releases your operator installed, and not to prevent other tools from also being able to manage those releases (the above tradeoff). Can you clarify which of these it is?

Note some technical details about the tradeoff I mentioned— Helm actions currently inconsistently rely on this label today: 
- actions like install, upgrade, and rollback do
- actions like list and uninstall do not

Summary:
- if your goal is to denote that your operator has managed the Helm release, you may wish to choose a different label and leave the common `app.kubernetes.io/managed-by: Helm` label as-is
- if your goal is to take exclusive ownership of certain helm releases (ie, prevent other tools such as the Helm CLI or others that leverage the Helm SDK from interacting with them) then this would require additional changes than what this PR currently does, and may be best to write a HIP to make sure this ideas is fully thought through, all bases covered, this option is well-docmented, doesn't raise other issues when not used, etc

luksa:
@scottrigby Yeah, I see your point. If an operator uses Helm in such a way that the Helm release shows up in commands like `helm ls`, then other `helm` CLI commands should continue to work against that release. If an operator wants to have full control of the applied resources, without allowing users to interfere with the `helm` CLI, then the operator should also ensure that `helm ls` doesn't see any releases. 

We've actually been planning to use the Helm SDK to only render the manifest and then have the operator apply them (for reasons other than the `app.kubernetes.io/managed-by` label). However, when we do this, we'll also be able to set this label to whatever we want. 

Thanks for your input. I'm closing this.

PR #30778, opened 2025-04-20
feat: send hooks through the post-renderer

### LINKED ISSUE #7891: Post-renderer not receiving chart hook manifests?

dwalters:
Includes hooks in the values passed to the post renderer.  Approach is as suggested by an early
comment on the issue, moving the post renderer earlier in the function before manifests are "sorted"
into hooks and other.

Backwards compatibility is maintained by only passing through hooks that opt-in to this behavior
via a `helm.sh/hook-post-render=true` annotation.  Would be worth considering making this the
default in helm4, and removing the associated code.

closes #7891 

**If applicable**:

### DIFF
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -32,6 +32,7 @@ import (
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
+	"sigs.k8s.io/yaml"
 
 	"helm.sh/helm/v3/pkg/chart"
 	"helm.sh/helm/v3/pkg/chartutil"
@@ -62,6 +63,8 @@ var (
 	// errPending indicates that another instance of Helm is already applying an operation on a release.
 	errPending = errors.New("another operation (install/upgrade/rollback) is in progress")
 )
+
+const HookPostRenderAnnotation = "helm.sh/hook-post-render"
 
 // ValidName is a regular expression for resource names.
 //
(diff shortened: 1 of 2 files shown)


dwalters:
If the approach seems acceptable I'm happy to make a similar change on main (but maybe remove the need for the annotation!)

scottrigby:
A a rule, PRs need to be against `main` before being backported to `dev-v3`, for both bug fixes and new features.

This has some timeline caveats, from [Maintaining Helm 3](https://github.com/helm/community/blob/main/hips/hip-0012.md#maintaining-helm-3) section of the Helm 4 Development Process HIP:
> - During the "Feature Development" phase of Helm 4, Helm 3 will continue to receive security, bug, and minor feature patches.
> - During the Alpha, Beta, and RC phases of Helm 4, no feature patches will be applied to Helm 3 to avoid introducing Helm 4 features prematurely while Helm 4's features are locked.
> - Helm 3 will receive bug fixes for 6-8 months following the release of Helm 4 and security fixes for one year from the day Helm 4 is made available.

To set expectations, maintainers are currently very focused on work for the first pre-release of Helm 4 (see [Timelines](https://github.com/helm/community/blob/main/hips/hip-0012.md#the-process--timelines) from that HIP), so are very sparing about accepting new features backported to Helm 3, so backports are a lower priority.

dwalters:
Sure, I'm happy to send a PR against main if there are no concerns with the overall approach.

scottrigby:
> Sure, I'm happy to send a PR against main if there are no concerns with the overall approach.

@dwalters I've added an agenda item to tomorrow's Helm dev meeting. You're welcome to join that (see [meeting info](https://github.com/helm/community/blob/main/communication.md#meetings)), otherwise will update here after.

scottrigby:
Linting errors also need to be fixed.

dwalters:
Linting errors are fixed.  

Made a separate PR for main over in #30968 .  I'm sure this will require some discussion, but I think it'd be better to to have helm v4 run hooks through the post renderer by default?  If the extra annotation check for backwards compatibility isn't necessary, this becomes a much simpler patch.

PR #13460, opened 2024-11-18
fix: replace "github.com/pkg/errors" with stdlib "errors" package

justenstall:
**What this PR does / why we need it**:

Replaces the archived "github.com/pkg/errors" package with the Go standard library equivalents, and updates error handling to use more recent standard library features.

**Special notes for your reviewer**:

Updates are generally:

- Replace `errors.Wrap(err, "msg")` and `errors.Wrapf(err, "msg")` with `fmt.Errorf("msg: %w", err)`
- Replace `errors.Errorf("msg", args)` with `fmt.Errorf("msg", args)`
- Replace uses of `[]string` and `strings.Join` for error messages with `[]error` and `errors.Join`
  - Used a simple wrapper type ([example](https://github.com/helm/helm/pull/13460/files#diff-4d76035a2539b2daa4c488f8494ef0a48903cc5e83558d3e2fffcafa553d00b8R187)) if the error messages were being joined with any string other than `"\n"` so they will render the same as before
- Replace `os.IsNotExist(err)` with `errors.Is(err, fs.ErrNotExist)`

**If applicable**:
- [ ] this PR contains user facing changes (the `docs needed` label should be applied if so)
- [ ] this PR contains unit tests
- [x] this PR has been tested for backwards compatibility

PR #30781, opened 2025-04-21
chore: remove `github.com/hashicorp/go-multierror` dependency

mmorel-35:

remove `github.com/hashicorp/go-multierror` dependency

`github.com/hashicorp/go-multierror` is under MPL-2.0 License which is not approved in CNCF Allowlist

**If applicable**:

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -4,6 +4,7 @@ run:
 linters:
   default: none
   enable:
+    - depguard
     - dupl
     - govet
     - ineffassign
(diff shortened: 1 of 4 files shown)


mattfarina (approved):
lgtm

godhanipayal (approved):

scottrigby (approved):
@mmorel-35 Thanks for this PR. Nice work.

Also noting that while this is also addressed in https://github.com/helm/helm/pull/13460, that doesn't include the depguard rule to `.golangci.yml` like your PR does. Because of this I think we should merge this, and then the other PR can rebase against main again to focus only on "github.com/pkg/errors". I also think that PR should add a similar depguard rule for that package after this merges. Will note there.

PR #30785, opened 2025-04-22
fix: govulncheck workflow

mmorel-35:

fixes govulncheck workflow

the checkout step is missing which make .github/env inaccessible when the first actual step is called

**If applicable**:


benoittgt (approved):

robertsirc (approved):
LGTM

PR #30788, opened 2025-04-22
ref(helm): Export Chart Not Found error

### LINKED ISSUE #30746: No Custom Exported Error types in `pkg/repo`
I've noticed that when pulling a helm chart, there is no custom exported error types for common failure modes such as a "not found" error.

For example: 
https://github.com/helm/helm/blob/d8ca55fc669645c10c0681d49723f4bb8c0b1ce7/pkg/repo/chartrepo.go#L300

In this case, the builtin `errors.errorString` default implementation is returned, with a little decoration. 

This means there's no way to identify this common error mode using `errors.Is()` or `errors.As()`. 

I'm happy to open a PR for this. 

If there are any other common error modes that we can identify in this same package, I'd be happy to add a custom error type for these too (for example with authentication issues).

stephenpmurray:
Closes #30746

- Export an error type for "Chart Not Found" type errors for error type checking

**If applicable**:

### DIFF
--- a/pkg/repo/chartrepo.go
+++ b/pkg/repo/chartrepo.go
@@ -219,7 +219,10 @@ func FindChartInRepoURL(repoURL string, chartName string, getters getter.Provide
 	}
 	cv, err := repoIndex.Get(chartName, opts.ChartVersion)
 	if err != nil {
-		return "", errors.Errorf("%s not found in %s repository", errMsg, repoURL)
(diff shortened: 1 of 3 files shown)


robertsirc:
@joejulian you commented in #30746 wanting to review this if there was a PR. Could you take a look with me on this?

joejulian (approved):
Excellent! Looks good, thanks!

Issue #12740, opened 2024-01-22
Inconsistencies in `helm install` and `helm temlplate` with `--dry-run`


lindhe:
Sorry for posting such a broad issue as this one, but I need to discuss this before I understand exactly what is wrong. It's likely to be a documentation issue, but may also be a novel approach to UX that I'm not getting, or even a bug. Let's figure it out!

For a TL;DR, skip the **Backstory** section and go directly to **Summary**.

## Backstory

I have a Helm chart that makes use of `.Capabilities.APIVersions.Has` for logic in its templates:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: foo
data:
  {{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress" }}
  wasTrue: yes
  {{- else }}
  wasTrue: no
  {{- end }}
```

Then I tried rendering it using `helm template` to see the results. I (naively) expected to render `wasTrue: yes` but I got `wasTrue: no`:

```console
$ helm template foo .
---
# Source: foo/templates/foo.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: foo
data:
  wasTrue: no
```

Well duh. 🤦 Obviously `helm template` cannot determine what capabilities my cluster has. Reading `helm template --help` to see if there's a way to let it communicate with the cluster, I found `--dry-run=server` which "allows attempting cluster connections". Bingo!

```console
$ helm template foo . --dry-run=server
---
# Source: foo/templates/foo.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: foo
data:
  wasTrue: no
```

Nope, still the same result. 🤔 This got me looking for related issues here on GitHub and I found [a comment](https://github.com/helm/helm/issues/12510#issuecomment-1773414759) that informed me that `--dry-run=server` only works for the `lookup` function. That's not what I gathered from the help text, so maybe it's just insufficiently documented?

Any way, I thought I'd try `helm install` too, just to see what happens:

```console
$ helm install foo . --dry-run=server
NAME: foo
LAST DEPLOYED: Mon Jan 22 20:10:40 2024
NAMESPACE: invenio-dev
STATUS: pending-install
REVISION: 1
TEST SUITE: None
HOOKS:
MANIFEST:
---
# Source: foo/templates/foo.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: foo
data:
  wasTrue: yes
```

[...]

PR #30796, opened 2025-04-23
feat: add kube version option for `helm install`

blakepettersson:

This is to make `helm install --dry-run` consistent with `helm template`. `helm template` can set `kube-version`, `helm install` up until now could not. In order for this to work as intended, we also need to set `client.clientOnly`. This could not be set before due to backwards compatibility concerns but for v4 this should be fine.

Related to #12740.

**If applicable**:

### DIFF
--- a/pkg/cmd/install.go
+++ b/pkg/cmd/install.go
@@ -31,6 +31,8 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 
+	chartutil "helm.sh/helm/v4/pkg/chart/v2/util"
+
 	"helm.sh/helm/v4/pkg/action"
 	chart "helm.sh/helm/v4/pkg/chart/v2"
 	"helm.sh/helm/v4/pkg/chart/v2/loader"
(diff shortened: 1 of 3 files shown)


gjenkins8 (dismissed):

pkg/cmd/install.go
```diff
@@ -144,6 +147,14 @@ func newInstallCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 			return compInstall(args, toComplete, client)
 		},
 		RunE: func(_ *cobra.Command, args []string) error {
+			if kubeVersion != "" {
+				parsedKubeVersion, err := chartutil.ParseKubeVersion(kubeVersion)
+				if err != nil {
+					return fmt.Errorf("invalid kube version '%s': %s", kubeVersion, err)
+				}
```
godhanipayal, line 154: Maybe , improve the fmt message for better readability ? 

return fmt.Errorf("failed to parse --kube-version '%s': %s", kubeVersion, err) 
this makes the error message more clearer that it's about the --kube-version flag

blakepettersson: I can do that, it's just that I copied the exact same code from `helm template` and wanted to be consistent with how it does it. Should I then do the same for `helm template` or should I leave `helm template` as-is? cc @gjenkins8

blakepettersson:
Not stale

sabre1041:
@blakepettersson would you foresee the need to also include this functionality in `helm upgrade` as well?

PR #30800, opened 2025-04-24
fix: dep fs errors

mmorel-35:

Fixes dep fs errors related to https://github.com/helm/helm/actions/runs/14629559870/job/41048978239
align the error generation with the original version, see https://github.com/golang/dep/blob/master/internal/fs/rename_windows.go and https://github.com/golang/dep/blob/master/internal/fs/rename.go

**If applicable**:


TerryHowe (approved):
/lgtm

gjenkins8 (approved):
Original was:
```
return errors.Wrapf(terr, "link error: cannot rename %s to %s", src, dst)
```

ie. `terr` is correct.

Thanks for the fix!

PR #30807, opened 2025-04-27
test: add chart_uploader tests for pkg/uploader

jessesimpson36:

Based on our coverage report, the chart_uploader has 0% coverage. This PR tests against it to assert error pathways and a basic happy-path test.

I was a little unsure about the part where I mock an OCI registry. I'm a little surprised I couldn't find a dedicated mocking library for OCI registries, and wasn't sure how it would go if I did something like TestContainers and spun up an oci registry before the test.

**If applicable**:


scottrigby:
> I was a little unsure about the part where I mock an OCI registry. I'm a little surprised I couldn't find a dedicated mocking library for OCI registries, and wasn't sure how it would go if I did something like TestContainers and spun up an oci registry before the test.

What do you think about how this is handled in [registry.TestSuite]? Example in this test func: https://github.com/helm/helm/blob/31e22b9866af91e1a0ea2ad381798f6c5eec7f4f/pkg/registry/utils_test.go#L219

jessesimpson36:
I think the implementation in utils_test.go is pretty good, although I can't import the module TestSuite or the setup function. it would be pretty repetitive if I copied/pasted most of the logic.  perhaps I should have some sort of common utils package that isn't in a test module, that way multiple testsuites could call it.

jessesimpson36:
maybe the happy path isn't a meaningful test in this case, since this functionality is already tested in the registry package.

jessesimpson36:
I changed the happy-path test to assert that the function that does that actual uploading is called, but doesn't test further than that like it did 1 commit prior. reason being is the code in the registry package is actually pretty well tested already.

go.mod
```diff
@@ -129,6 +129,7 @@ require (
 	github.com/shopspring/decimal v1.4.0 // indirect
 	github.com/sirupsen/logrus v1.9.3 // indirect
 	github.com/spf13/cast v1.7.0 // indirect
+	github.com/stretchr/objx v0.5.2 // indirect
```
TerryHowe, line 139: What triggered this?

pkg/uploader/chart_uploader_test.go
```diff
@@ -0,0 +1,121 @@
...
+		Pushers: pushers,
+	}
+
+	err := uploader.UploadTo("main", "invalid.com")
+	const expectedError = "scheme prefix missing from remote"
+
+	fmt.Println(err.Error())
+	if !strings.Contains(err.Error(), expectedError) {
+		t.Errorf("Expected error to contain '" + expectedError + "'")
+	}
```
TerryHowe: Something like:
```suggestion
	assert.ErrorContains(err, expectedError)
```

pkg/uploader/chart_uploader_test.go
```diff
@@ -0,0 +1,121 @@
...
+	uploader := ChartUploader{
+		Pushers: pushers,
+	}
+
+	err := uploader.UploadTo("main", "://invalid.com")
+	const expectedError = "invalid chart URL format"
+	fmt.Println(err.Error())
+	if !strings.Contains(err.Error(), expectedError) {
+		t.Errorf("Expected error to contain '" + expectedError + "'")
+	}
```
TerryHowe: Something like:
```suggestion
	assert.ErrorContains(err, expectedError)
```

pkg/uploader/chart_uploader_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+	mockedPusher.On("Push", "testdata/test-0.1.0.tgz", "oci://test").Return(nil)
+	err := uploader.UploadTo("testdata/test-0.1.0.tgz", "oci://test")
+	mockedPusher.AssertCalled(t, "Push", "testdata/test-0.1.0.tgz", "oci://test")
+
+	if err != nil {
+		fmt.Println(err)
+		t.Errorf("Expected push to succeed but got error")
+	}
```
TerryHowe:
```suggestion
	assert.NoError(err)
```

pkg/uploader/chart_uploader.go
```diff
@@ -29,7 +29,7 @@ type ChartUploader struct {
 	// Out is the location to write warning and info messages.
 	Out io.Writer
 	// Pusher collection for the operation
-	Pushers pusher.Providers
+	Pushers pusher.ProvidersResolver
```
TerryHowe, line 32: Can't you add your test provider to Providers to avoid the interface change?

jessesimpson36: I don't think I understand this comment. Provider is not an interface, it's a struct. I can't mock out a function on a struct, so that's why I have an interface where Provider is one implementation of that interface.

go.mod
jessesimpson36: I'm not sure. I tried removing it, and running `make` caused it to come back again.

```
> go mod why github.com/stretchr/objx

# github.com/stretchr/objx
helm.sh/helm/v4/pkg/uploader
helm.sh/helm/v4/pkg/uploader.test
github.com/stretchr/testify/mock
github.com/stretchr/objx
```

looks like the testify/mock library is pulling it in.

PR #30826, opened 2025-05-01
Improve test coverage of pkg/registry/util.go

jamoflaw:

Improves the test coverage of `pkg/registry/util.go`

Improves coverage of the package overall from 70.8% to 75.7%

**If applicable**:

### DIFF
--- a/pkg/registry/util.go
+++ b/pkg/registry/util.go
@@ -53,7 +53,8 @@ func ContainsTag(tags []string, tag string) bool {
 	return false
 }
 
-// GetTagMatchingVersionOrConstraint expects a sorted slice of semver tags (highest version first)
+// GetTagMatchingVersionOrConstraint gets the latest tag matching a given semver constraint.
+// Expects tags to be a sorted slice of semver tags (highest version first)
 func GetTagMatchingVersionOrConstraint(tags []string, versionString string) (string, error) {
 	var constraint *semver.Constraints
 	if versionString == "" {
--- a/pkg/registry/util.go
+++ b/pkg/registry/util.go
@@ -53,6 +53,7 @@ func ContainsTag(tags []string, tag string) bool {
 	return false
 }
 
+// GetTagMatchingVersionOrConstraint expects a sorted slice of semver tags (highest version first)
 func GetTagMatchingVersionOrConstraint(tags []string, versionString string) (string, error) {
 	var constraint *semver.Constraints
 	if versionString == "" {
--- a/pkg/registry/util_test.go
+++ b/pkg/registry/util_test.go
@@ -280,28 +280,28 @@ func TestIsOCI(t *testing.T) {
 
 	tests := []struct {
 		name    string
-		url     string
+		uri     string
 		isValid bool
 	}{
 		{
-			name:    "Valid url",
-			url:     "oci://example.com/myregistry:1.2.3",
+			name:    "Valid URI",
+			uri:     "oci://example.com/myregistry:1.2.3",
 			isValid: true,
 		},
 		{
-			name:    "Invalid URL prefix (boundary test 1)",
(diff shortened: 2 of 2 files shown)


pkg/registry/util.go
```diff
@@ -53,6 +53,7 @@ func ContainsTag(tags []string, tag string) bool {
 	return false
 }
 
+// GetTagMatchingVersionOrConstraint expects a sorted slice of semver tags (highest version first)
```
jamoflaw, line 56: I thought it is worth adding a comment here as to what the method expects as its not clear that it expects this input to be pre-sorted

pkg/registry/util_test.go
```diff
@@ -273,3 +274,185 @@ func TestGenerateOCICreatedAnnotations(t *testing.T) {
...
+}
+
+func TestContainsTag(t *testing.T) {
+
+	tagList := []string{"1.0.0", "1.0.1", "2.0.0", "2.1.0"}
+	assert.True(t, ContainsTag(tagList, "1.0.1"), "The tag 1.0.1 is in the list")
+	assert.False(t, ContainsTag(tagList, "1.0.2"), "the tag 1.0.2 is not in the list")
+}
+
+func TestGetTagMatchingVersionOrConstraint(t *testing.T) {
```
jamoflaw, line 292: This should test the scenarios outlined in the masterminds package https://pkg.go.dev/github.com/Masterminds/semver/v3@v3.3.0#section-readme as well as the explicit variants which are helm specific

mattfarina (changes requested):
Thank you for working on tests.

When we repeat the same sequence with different data there is a pattern for that. Could you please follow this pattern. There is an example [here](https://github.com/helm/helm/blob/31e22b9866af91e1a0ea2ad381798f6c5eec7f4f/pkg/cmd/pull_test.go#L52). If you search the code for `tests := []struct {` you'll find more examples.

jamoflaw:
> Thank you for working on tests.
> 
> When we repeat the same sequence with different data there is a pattern for that. Could you please follow this pattern. There is an example [here](https://github.com/helm/helm/blob/31e22b9866af91e1a0ea2ad381798f6c5eec7f4f/pkg/cmd/pull_test.go#L52). If you search the code for `tests := []struct {` you'll find more examples.

Updated as per review - I wasn't sure if there was a minimum for that style so I converted all of them - as it makes for a better reading experience anyway!

TerryHowe (approved):
/lgtm

Although TestGetTagMatchingVersionOrConstraint you could argue is testing NewConstraint in many cases.

pkg/registry/util_test.go
```diff
@@ -273,3 +275,282 @@ func TestGenerateOCICreatedAnnotations(t *testing.T) {
...
+
+func TestIsOCI(t *testing.T) {
+
+	tests := []struct {
+		name    string
+		url     string
+		isValid bool
+	}{
+		{
+			name:    "Valid url",
```
gjenkins8: nit:
```suggestion
			name:    "Valid URI",
```

(A URL is a location you might send e.g. HTTP requests; a URI is just an identifier, which is true for OCI references)

gjenkins8: (and other places where we refer to OCI URIs as URLs)

jamoflaw: Thanks @gjenkins8 - Hopefully thats covered now - I've updated all wording to be consistent I believe now!

TerryHowe (approved):
/lgtm

TerryHowe:
```
< 							IsOCI						100.0%
< 							ContainsTag					100.0%
< 							GetTagMatchingVersionOrConstraint		94.1%
---
> 							IsOCI						0.0%
> 							ContainsTag					0.0%
> 							GetTagMatchingVersionOrConstraint		0.0%
```

Issue #30814, opened 2025-04-28
Helm 4 - Testing coverage is low for certain packages


razvansodoleanu:
Testing coverage for the following helm 4 packages is low. Target is to reach 70-80% coverage per package, for the packages below:

- actions - 56.9%
- kube - 59.7%
- pusher - 32.9%
- time - 38.5%

TerryHowe:
I'll look at time

razvansodoleanu:
@jessesimpson36 is handling chart_uploader tests

AshmitBhardwaj:
I'm working on actions package. Test coverage has reached 60.5%. Will raise a PR soon

mahmou991:
تاداهشلا ضراع
صتﻻلا
DigiCert Global Root G2
ردصأ GeoTr
() ةسسؤملا حتدقل
Digicert inc ةداهش
)۞) ةيميظنتلا ةدحولا امولعم
www.digicert.com
دت
متي
يلسلستلا مقرلا ةثيدح
03:3A:F1:E6:A7:1:A9:A0:BB:28:64:51:1
D:O9:FA:ES اصتالا
زيمرت ةعو
مدختسي
رادصإلا ةهج مرت مت
(هN عئاشلا مسالا P5519
Digicert Global Root G2 بت اذا.
(O) ةسسؤملا
Digicert inc
ال) ةيميظنتلا ةدحولا
www.digicert.com
ةيحالصلا ةرتف
رادصإلا خيرات راطب مهفا
٣٠٣١١٠١١ ١موشلنويأةيراطب
دﺣأب؟ﺔﺎﺔxy فتاوه زيمت
ملا فرعادمألا لوطو تنولا
اتياطإلءاذأومقاﻖفس
ةيحالصلا ءاهتنا خيراتا يـ
٤٣٣٣٩٢١٧
؟HA-2ة6 ةيمزراوخل ةيعجرملا تافلملا
ةداهش
CB3CCB B76031E5 E0 13 8F8D D3
9A23F9 DE47FFC3 5E43 C1144C
EAZ7D46A 5A B1CB5F
اهتنايصو
ماع حاتفم
8B B5 93 A93BE1DO E8A822BB 88
ﻄانرالا فيرعت تافموجنوسماس
7C547890C3ETOBAAD2DAB762
انلآبg ،طابك الا ر
تاﺸاﻓينﻮمكلااجوملاانه مدخﺘ
[...]

PR #30829, opened 2025-05-03
Increase pkg/time test coverage

TerryHowe:
Add test coverage for `pkg/time` avoiding locale dependent methods.

Partial: https://github.com/helm/helm/issues/30814

**If applicable**:

### DIFF
--- a/pkg/time/time.go
+++ b/pkg/time/time.go
@@ -65,6 +65,7 @@ func Parse(layout, value string) (Time, error) {
 	t, err := time.Parse(layout, value)
 	return Time{Time: t}, err
 }
+
 func ParseInLocation(layout, value string, loc *time.Location) (Time, error) {
 	t, err := time.ParseInLocation(layout, value, loc)
 	return Time{Time: t}, err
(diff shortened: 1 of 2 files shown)


gjenkins8 (commented):
Not against this PR, but perhaps we could also remove this package?

It exists because of because of https://github.com/golang/go/issues/11939, according to: https://github.com/helm/helm/blob/da579a7aa6199ea1fbc325ece5e7fa49a4eb83fa/pkg/time/time.go#L20, 

Which is now (finally) solved by `time.Time` supporting `omitzero`:
https://tip.golang.org/doc/go1.24#encodingjsonpkgencodingjson

robertsirc (approved):
LGTM

godhanipayal (approved):

TerryHowe:
> Not against this PR, but perhaps we could also remove this package?
> 
> It exists because of because of [golang/go#11939](https://github.com/golang/go/issues/11939), according to:

I created an issue for this in case we want to do the pkg/time as a follow-up

gjenkins8 (approved):
nice, thanks!

Issue #30514, opened 2025-02-14
v3.17.0 helm dry-run with option none or client attempts cluster connections


barani129:
Hi,

Issue:

I'm relatively new to helm chart development. While running dry-run option to validate a simple helm chart (without cluster connection), helm tries to access API server.

Expectation:

When dry-run option set to none or client mode, helm should not attempt cluster connection.

Info:

% helm version
version.BuildInfo{Version:"v3.17.0", GitCommit:"301108edc7ac2a8ba79e4ebf5701b0b6ce6a31e4", GitTreeState:"clean", GoVersion:"go1.23.4"}

Output:

% helm install --dry-run confo ../nginx-chart
Error: INSTALLATION FAILED: Kubernetes cluster unreachable: Get "https://127.0.0.1:52243/version": dial tcp 127.0.0.1:52243: connect: connection refused

% helm install --dry-run confo ../nginx-chart
Error: INSTALLATION FAILED: Kubernetes cluster unreachable: Get "https://127.0.0.1:52243/version": dial tcp 127.0.0.1:52243: connect: connection refused

Problematic line in the code:

https://github.com/helm/helm/blob/3a94215585b91d5ac41ebb258e376aa11980b564/pkg/action/install.go#L258

if !i.isDryRun() || i.DryRunOption == "server" || i.DryRunOption == "none" || i.DryRunOption == "false" {
		interactWithRemote = true
	}

Regards,
Barani

gjenkins8:
> When dry-run option set to none or client mode, helm should not attempt cluster connection.

This is not correct. `--dry-run=client` is the only option which should not contact a cluster. `--dry-run=none` means no dry run, and so should contact a cluster

barani129:
Hi @gjenkins8 , thanks for looking into this issue. Appreciate it.

PR #30833, opened 2025-05-05
refactor/cleanup: Replace action 'DryRun' string with DryRunStrategy type + deprecations

gjenkins8:

This PR is a cleanup/refactor of Helm's "dry run" code:
- To remove the legacy `DryRun` bool from Install/Update/Rollback actions
- Replace the `DryRunOption string` member in Install/Upgrade/Rollback actions with an enum `DryRunStrategy` (named after the like in [kubectl](https://github.com/kubernetes/kubectl/blob/bb3c0d9f3ad622036ce70ca4e82deb7af1eaaabc/pkg/cmd/util/helpers.go#L673)).
- Remove Install action's `ClientOnly` flag (equivalent to `DryRunClient`)

The change also:
- updates rollback to include the new `--dry-run=none|client|server` option
- introduces logging/warnings for usage of the deprecated boolean `--dry-run=false|true` and no-option `--dry-run` syntaxes
- remove the `action.Install`, `action.Update` member functions: `i.DryRun()` and `i.InteractWithRemote`. And replaces them with standalone helpers (with tests)
- Deprecate `helm template --validate` flag (replaced by `helm template --dry-run=server`)

https://github.com/helm/helm/pull/9426/files#diff-5d4849bad97d2e075ce5d0ae5af3e93dbfde235ce1345fa96148279cf706af2fR76 added the `DryRunOption` member to the action structs (install and upgrade, it missed rollback) to support client-side vs server-side dry runs. And left the legacy `DryRun bool` member for backwards compatibility (which this PR now cleans up)

**If applicable**:

### DIFF
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -70,6 +70,21 @@ var (
 	errPending = errors.New("another operation (install/upgrade/rollback) is in progress")
 )
 
+type DryRunStrategy string
+
+const (
+	// DryRunNone indicates the client will make all mutating calls
+	DryRunNone DryRunStrategy = "none"
+
+	// DryRunClient, or client-side dry-run, indicates the client will avoid
+	// making calls to the server
+	DryRunClient DryRunStrategy = "client"
+
+	// DryRunServer, or server-side dry-run, indicates the client will send
+	// calls to the APIServer with the dry-run parameter to prevent persisting changes
+	DryRunServer DryRunStrategy = "server"
+)
+
 // Configuration injects the dependencies that all actions share.
 type Configuration struct {
 	// RESTClientGetter is an interface that loads Kubernetes clients.
@@ -527,4 +542,14 @@ func determineReleaseSSApplyMethod(serverSideApply bool) release.ApplyMethod {
 		return release.ApplyMethodServerSideApply
 	}
 	return release.ApplyMethodClientSideApply
+}
+
+// isDryRun returns true if the strategy is set to run as a DryRun
+func isDryRun(strategy DryRunStrategy) bool {
+	return strategy == DryRunClient || strategy == DryRunServer
+}
+
+// interactWithServer determine whether or not to interact with a remote Kubernetes server
+func interactWithServer(strategy DryRunStrategy) bool {
+	return strategy == DryRunNone || strategy == DryRunServer
 }
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -75,7 +75,6 @@ type Install struct {
 
 	ChartPathOptions
 
-	ClientOnly bool
 	// ForceReplace will, if set to `true`, ignore certain warnings and perform the install anyway.
 	//
 	// This should be used with caution.
@@ -87,8 +86,8 @@ type Install struct {
 	// see: https://kubernetes.io/docs/reference/using-api/server-side-apply/
 	ServerSideApply bool
 	CreateNamespace bool
-	DryRun          bool
-	DryRunOption    string
+	// DryRunStrategy can be set to prepare, but not execute the operation and whether or not to interact with the remote cluster
+	DryRunStrategy DryRunStrategy
 	// HideSecret can be set to true when DryRun is enabled in order to hide
 	// Kubernetes Secrets in the output. It cannot be used outside of DryRun.
 	HideSecret       bool
@@ -159,6 +158,7 @@ func NewInstall(cfg *Configuration) *Install {
 	in := &Install{
 		cfg:             cfg,
 		ServerSideApply: true,
+		DryRunStrategy:  DryRunNone,
 	}
 	in.registryClient = cfg.RegistryClient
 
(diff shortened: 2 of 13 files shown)


godhanipayal (approved):
Left minor comments.

pkg/action/action.go
```diff
@@ -64,6 +64,21 @@ var (
 	errPending = errors.New("another operation (install/upgrade/rollback) is in progress")
 )
 
+type DryRunStrategy string
+
+const (
+	// DryRunNone indicates the client will make all mutating calls
+	DryRunNone DryRunStrategy = "none"
```
TerryHowe, line 71: Do we have to have "none" as a valid value. It is a bit confusing that `--dry-run=none` is not a dry run.

godhanipayal: Is dry-run = none is functionally equivalent to not setting flag at all ? Maybe, its need to maintain parity with kubectl dryrun strategy

TerryHowe: Yeh, good point. I never really thought about it, but `none` doesn't seem great.

gjenkins8: Keep in mind this is modeled after (copied from) `kubectl` (aiming to have flags with the same options for consistency):

```
$ kubectl apply --help
...
    --dry-run='none':
        Must be "none", "server", or "client". If client strategy, only print the object that would be sent, without sending it. If server strategy, submit server-side request without persisting the resource.
```

pkg/cmd/template.go
```diff
@@ -203,6 +201,11 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 	f.StringVar(&kubeVersion, "kube-version", "", "Kubernetes version used for Capabilities.KubeVersion")
 	f.StringSliceVarP(&extraAPIs, "api-versions", "a", []string{}, "Kubernetes api versions used for Capabilities.APIVersions")
 	f.BoolVar(&client.UseReleaseName, "release-name", false, "use release name in the output-dir path.")
+	f.String(
+		"dry-run",
+		"server",
+		`simulate the operation. must be either "client" or "server" (default). If '--dry-run=client', operations will be performed client-side only, and cluster connections avoid. '--dry-run=server' will enable server connectivity`)
+	f.Lookup("dry-run").NoOptDefVal = "unset"
```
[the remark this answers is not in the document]
gjenkins8: minor note: template is always a "dry-run" so the 'none' option isn't available/described.

pkg/cmd/helpers.go
```diff
@@ -0,0 +1,83 @@
...
+	// - Not set means no dry run and server is contacted.
+	// - Set with no value, a value of client, or a value of true and the server is not contacted
+	// - Set with a value of false, none, or false and the server is contacted
+	// The true/false part is meant to reflect some legacy behavior while none is equal to "".
+	f := cmd.Flags()
+	f.String(
+		"dry-run",
+		"none",
+		`simulate the operation. must be either "none" (default), "client" or "server". If '--dry-run=client', operations will be performed client-side only, and cluster connections avoid. '--dry-run=server' will enable server connectivity`)
+	f.Lookup("dry-run").NoOptDefVal = "unset"
```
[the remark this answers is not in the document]
gjenkins8: nice, thanks. this is better (I did also make some minor amendments, let me know what you think)

pkg/action/action.go
```diff
@@ -418,3 +433,13 @@ func (cfg *Configuration) Init(getter genericclioptions.RESTClientGetter, namesp
 func (cfg *Configuration) SetHookOutputFunc(hookOutputFunc func(_, _, _ string) io.Writer) {
 	cfg.HookOutputFunc = hookOutputFunc
 }
+
+// isDryRun returns true if the strategy is set to run as a DryRun
+func isDryRun(strategy DryRunStrategy) bool {
```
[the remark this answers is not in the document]
gjenkins8: overall, I think I prefer decoupling the business logic and the type

pkg/action/install.go
```diff
@@ -74,8 +74,8 @@ type Install struct {
 	ClientOnly      bool
```
kilianpaquier: **suggestion:** `ClientOnly` should be removed in favor of `DryRunStrategy` with `client`.

Would fix #30514 (see `pkg/action/install.go:240` and `pkg/action/install.go:274`).

gjenkins8: thanks! nice spot. I do agree it makes sense for this PR to also remove `ClientOnly`.

kilianpaquier (approved):

TerryHowe (approved):
/lgtm

pkg/action/action.go
```diff
@@ -418,3 +433,13 @@ func (cfg *Configuration) Init(getter genericclioptions.RESTClientGetter, namesp
...
+	return strategy == DryRunClient || strategy == DryRunServer
+}
+
+// interactWithServer determine whether or not to interact with a remote Kubernetes server
+func interactWithServer(strategy DryRunStrategy) bool {
```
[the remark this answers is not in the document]
gjenkins8: (same as above) -- overall, I think I prefer decoupling the business logic and the type

pkg/action/action.go
scottrigby: I agree it's a bit weird, but also agree that mirroring this option from kubectl makes sense :+1:

scottrigby (approved):
lgtm. tested and works as described.

I agree the `kubectl` flag --dry=run=none seems odd. also odd that it's default is `none` (no dry run) but that the flag's default _value_ is `client`. but I do think we should mirror the same options and behavior as what's upstream in `kubectl`, and this PR does.

also note helm template does not support `none`, becuase it is implicitly dry run only. `client` and `server` options for this flag are only for lookup purposes. this PR handles all of this well.

PR #30852, opened 2025-05-10
fix: add warning when ignore repo flag

shfc:

While there is an argument about treating it as a bug or a UX improvement for this issue, we could add a warning for this unexpected behaviour instead of doing nothing. This makes it easier for others to identify the issue in the logs when they encounter installation problems, saving time and effort. It also gives a heads-up to anyone who might be unknowingly relying on this unexpected behaviour, giving them time to adapt before future upgrades.

refs #7862 

**If applicable**:

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"log/slog"
 	"net/url"
 	"os"
 	"path"
@@ -762,7 +763,7 @@ func (c *ChartPathOptions) LocateChart(name string, settings *cli.EnvSettings) (
 		// Issue #7862: Helm prioritizes local charts over --repo flag.
 		// This behavior is maintained for backwards compatibility but with a warning.
 		if c.RepoURL != "" {
-			fmt.Printf("WARNING: local chart %s found in current working directory. --repo flag will be ignored\n", name)
+			slog.Warn("local chart found in current working directory. --repo flag ignored", "chart", name)
 		}
 		return abs, nil
 	}
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -759,6 +759,11 @@ func (c *ChartPathOptions) LocateChart(name string, settings *cli.EnvSettings) (
 				return "", err
 			}
 		}
+		// Issue #7862: Helm prioritizes local charts over --repo flag.
+		// This behavior is maintained for backwards compatibility but with a warning.
+		if c.RepoURL != "" {
+			fmt.Printf("WARNING: local chart %s found in current working directory. --repo flag will be ignored\n", name)
+		}
 		return abs, nil
 	}
 	if filepath.IsAbs(name) || strings.HasPrefix(name, ".") {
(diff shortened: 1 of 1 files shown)


pkg/action/install.go
```diff
@@ -759,6 +759,11 @@ func (c *ChartPathOptions) LocateChart(name string, settings *cli.EnvSettings) (
 				return "", err
 			}
 		}
+		// Issue #7862: Helm prioritizes local charts over --repo flag.
+		// This behavior is maintained for backwards compatibility but with a warning.
+		if c.RepoURL != "" {
+			fmt.Printf("WARNING: local chart %s found in current working directory. --repo flag will be ignored\n", name)
```
TerryHowe, line 765: Maybe something along the lines of
```suggestion
			slog.Warn("local chart found in current working directory. --repo flag ignored", "chart", name)
```

shfc: updated and tested

TerryHowe (commented):
/lgtm

Maybe @robertsirc has some opinions on the text logged since he's been working on logging

pkg/action/install.go
```diff
@@ -759,6 +760,11 @@ func (c *ChartPathOptions) LocateChart(name string, settings *cli.EnvSettings) (
 				return "", err
 			}
 		}
+		// Issue #7862: Helm prioritizes local charts over --repo flag.
+		// This behavior is maintained for backwards compatibility but with a warning.
+		if c.RepoURL != "" {
+			slog.Warn("local chart found in current working directory. --repo flag ignored", "chart", name)
+		}
 		return abs, nil
```
godhanipayal, line 768: Repo flag sometimes adds to the confusion. The above warning looks good. Perhaps, we could add a more actionable warning for clarity
slog.Warn("Local chart found in current working directory; --repo flag ignored. To use the remote chart, remove or rename the local directory.", "chart", name)

godhanipayal (approved):

pkg/action/install.go
shfc: Thanks for the thoughtful suggestion! I reviewed the existing warning log practices in this repo and noticed two general styles: some logs simply state what’s happening, while others provide actionable guidance. It’s not entirely clear to me when one style is preferred over the other.

@robertsirc, do you have any thoughts on which approach aligns best with our current logging conventions?

shfc:
I have made a small change to move the warning to the top of the block, so it’s always logged.  This will make it clear if any subsequent errors are because a local chart is being used instead of a remote repository.

mattfarina (changes requested):

pkg/action/install.go
```diff
@@ -750,6 +751,12 @@ func (c *ChartPathOptions) LocateChart(name string, settings *cli.EnvSettings) (
 	version := strings.TrimSpace(c.Version)
 
 	if _, err := os.Stat(name); err == nil {
+		// Issue #7862: Helm prioritizes local charts over --repo flag.
+		// This behavior is maintained for backwards compatibility but with a warning.
+		if c.RepoURL != "" {
+			slog.Warn("local chart found in current working directory. --repo flag ignored", "chart", name)
```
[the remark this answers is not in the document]
shfc: Thanks for your suggestion. I have updated it to be more general.

TerryHowe (approved):
/lgtm

robertsirc (approved):
LGTM

mattfarina (approved):
lgtm

PR #30771, opened 2025-04-18
[v3 backport] chore: Update Golang to v1.24

dongjiang1989:
**What this PR does / why we need it**:

Backport #30677  to dev v3

**Special notes for your reviewer**:

**If applicable**:
- [ ] this PR contains user facing changes (the `docs needed` label should be applied if so)
- [ ] this PR contains unit tests
- [x] this PR has been tested for backwards compatibility

PR #30752, opened 2025-04-14
Bump golangci lint to last major version and fix static-check errors

benoittgt:
**What this PR does / why we need it**:

Trying to close dependabot auto-update https://github.com/helm/helm/pull/30706

The update t[riggered few static-check errors](https://github.com/helm/helm/actions/runs/14401559126/job/40388281241) that I fixed. It's grouped by commit for easier review 

**Special notes for your reviewer**:

**If applicable**:
- [ ] this PR contains user facing changes (the `docs needed` label should be applied if so)
- [ ] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

Closes: https://github.com/helm/helm/pull/30706

PR #30853, opened 2025-05-11
chore: Include `.github/env` on `dev-v3` branch

gjenkins8:
Dependabot bumps (at least) are currently failing as they are trying to read `.github/env` file e.g.

https://github.com/helm/helm/pull/30840

File is copy from `main` branch

**If applicable**:


.github/env
```diff
@@ -1,2 +1,2 @@
 GOLANG_VERSION=1.24
-GOLANGCI_LINT_VERSION=v1.64
```
benoittgt, line 2: I think if we do this we have to "backport" what we have on main branch. Like I did in https://github.com/helm/helm/pull/30752 but there is a little more after in golang ci lint config file.

benoittgt: Also related https://github.com/helm/helm/pull/30771/files#r2050843093

robertsirc: I believe @benoittgt is correct on this one...

gjenkins8: I was confused by outdated local branch. https://github.com/helm/helm/pull/30785 is the fix which needs to be backported

PR #11766, opened 2023-01-26
Recursive dependencies

migueleliasweb:
**What this PR does / why we need it**:

This PR attempts to address the ~6 year long issue https://github.com/helm/helm/issues/2247.

The idea behind this PR is to allow users to recursively deal with chart dependencies. Currently, the behaviour of the `--dependency-update` flag is constrainted to the current chart we're interfacing with. If the chart has downstream dependencies, they won't be updated.

This PR adds a new flags. The new flag `--dependency-update-recursive` is set by default to false and don't interfere with the current behaviour of the existing `--dependency-update` flag. For the `helm dependency *` commands, the flag `--recursive` is also set to `false` by default and won't interfere with the current behaviour of helm unless manually set.

The final commands look like:

```
helm install --dependency-update-recursive my-chart
helm template --dependency-update-recursive my-chart
helm dependency build --recursive
helm dependency update --recursive
```

**Special notes for your reviewer**:

**If applicable**:
- [ ] this PR contains documentation
- [x] this PR contains unit tests
- [x] this PR has been tested for backwards compatibility

PR #30855, opened 2025-05-12
Recursive dependencies

### LINKED ISSUE #2247: Recursive dependency handling for charts.
### Use case:
Complex applications may have nested dependencies between the different charts. As the dependencies are visible by recursively scanning the charts, it can be naturally desired to assemble top-level charts with single ```helm``` operation.
### Problem:
Currently ```helm dependency build``` handles only first-level dependencies. These first level dependencies are put into the ```charts/``` folder in tgz format and treated as complete (assuming second level dependencies built in the first level deps tgz).
### Workaround:
Currently user has to create tool which either knows the dependencies in-built or able to scan ```requirements.yaml``` while calling or faking ```helm dependency build``` steps.
### Solution:
Would be good to have a ```--recursive``` flag to the ```dependency build``` command in ```helm```. This would scan the dependencies recursively and fixing any missing charts immediately. This would also store the dependencies under ```/charts``` in an untarred manner in order to avoid inception :smile:.
The ultimate solution would be to have this implicit chart procurement option in case of ```helm install```.

[Problem statement in slack archive.](https://kubernetes.slack.com/conversation/C0NH30761/p1491579986296921)

alikhil:

This PR is a rebase of #11766 
Closes #2247 
Note from original PR:
> The idea behind this PR is to allow users to recursively deal with chart dependencies. Currently, the behaviour of the --dependency-update flag is constrainted to the current chart we're interfacing with. If the chart has downstream dependencies, they won't be updated.
>
>This PR adds a new flags. The new flag --dependency-update-recursive is set by default to false and don't interfere with the current behaviour of the existing --dependency-update flag. For the helm dependency * commands, the flag --recursive is also set to false by default and won't interfere with the current behaviour of helm unless manually set.

>The final commands look like:

```shell
helm install --dependency-update-recursive my-chart
helm template --dependency-update-recursive my-chart
helm dependency build --recursive
helm dependency update --recursive
```

**If applicable**:


banjoh (changes requested):
Thanks a lot for re-raising this PR. It certainly is a helpful feature to have.

I tested it I got an error where a subchart has a dependency from a remote repository

```
helm dep update --recursive
Error: could not find https://kedacore.github.io/charts: stat https://kedacore.github.io/charts: no such file or directory
```

Some additional logging would also be very helpful. I'll try to add some suggestions in the PR

pkg/downloader/manager.go
```diff
@@ -218,13 +240,13 @@ func (m *Manager) Update() error {
 	return writeLock(m.ChartPath, lock, c.Metadata.APIVersion == chart.APIVersionV1)
 }
 
-func (m *Manager) loadChartDir() (*chart.Chart, error) {
-	if fi, err := os.Stat(m.ChartPath); err != nil {
-		return nil, fmt.Errorf("could not find %s: %w", m.ChartPath, err)
+func (m *Manager) loadChartDir(chartPath string) (*chart.Chart, error) {
+	if fi, err := os.Stat(chartPath); err != nil {
```
banjoh, line 287:
```suggestion
	slog.Debug("loading chart directory", "chartPath", chartPath)
	if fi, err := os.Stat(chartPath); err != nil {
```

pkg/downloader/manager.go
```diff
@@ -767,6 +789,61 @@ func (m *Manager) findChartURL(name, version, repoURL string, repos map[string]*
 	return url, username, password, false, false, "", "", "", err
 }
 
+// locateLocalDependencies locates local dependencies for the given chart (optionally recursively)
+//
+// The returned list of Chart paths is ordered from "leaf to root" so we can issue updates in
+// the right order when iterating over this list.
+func (m *Manager) locateLocalDependencies(baseChartPath string, resursive bool) ([]string, error) {
+	reversedDeps := []string{}
```
banjoh, line 842:
```suggestion
	slog.Debug("locating local dependencies", "baseChartPath", baseChartPath, "resursive", resursive)
	reversedDeps := []string{}
```

pkg/downloader/manager.go
```diff
@@ -149,8 +149,30 @@ func (m *Manager) Build() error {
...
+			}
+		}
+	}
+
+	// for last, update the root chart
+	return m.doUpdate(m.ChartPath)
+}
+
+func (m *Manager) doUpdate(chartPath string) error {
+	c, err := m.loadChartDir(chartPath)
```
banjoh, line 217:
```suggestion
	slog.Debug("update chart dependencies", "chartPath", chartPath)
	c, err := m.loadChartDir(chartPath)
```

pkg/downloader/manager.go
banjoh, line 287: The error where a remote repository is assumed to be a directory is coming from here. You could check the `file://` protocol or absence of it to assume its a path

When updating your tests, you can setup a local remote repo to serve your charts like its done in this test https://github.com/helm/helm/blob/d2236e95ed05e3c5b2598ceb677baa76011d9597/pkg/cmd/pull_test.go#L31

banjoh: https://github.com/helm/helm/blob/d2236e95ed05e3c5b2598ceb677baa76011d9597/pkg/downloader/manager.go#L633 is where a valid repo url is checked. You could extract that code to a helper function

banjoh:
@alikhil remember to `git commit -C` to have the DCO check passing

alikhil:
@banjoh thank you for your feedback!
I am on it!

banjoh:
> @banjoh thank you for your feedback! I am on it!

Let me know once done. I was just running tests and realised that the functionality isn't working quite right. I'll wait for your nudge.

alikhil:
@banjoh I have fixed the issue and updated tests. Could you give a look please?

banjoh:
@alikhil thanks for being persistent in continuously adding changes as I ask for them. Apologies I've not gotten back sooner.

I ran som tests using [foo.tgz](https://github.com/user-attachments/files/23106317/foo.tgz) simple chart which has recursive subcharts. The tree structure is like below

```sh
foo
├── baz
│   ├── bar
│   │   ├── Chart.yaml
│   │   ├── charts
│   │   └── values.yaml
│   ├── Chart.lock
│   ├── Chart.yaml
│   ├── charts
│   └── values.yaml
├── Chart.lock
├── Chart.yaml
├── charts
└── values.yaml
```

After running `helm package foo --dependency-update-recursive` I expect to see `foo/baz/bar/charts/nginx-ingress-controller.tgz` present in the unarchived. This is not the case. If anything, the command is failing

`helm dependency build --recursive foo` should pull `foo/baz/bar/charts/nginx-ingress-controller.tgz` and recursively package up all dependencies up the directory tree.

Could you please check why this isn't working? Also, once you have a resolved this, please verify that the commands below work as expected

```
helm install --dependency-update-recursive my-chart
helm template --dependency-update-recursive my-chart
helm dependency build --recursive
helm dependency update --recursive
helm package --dependency-update-recursive
```

alikhil:
@banjoh Thanks a lot for catching that! 🙌
I've fixed the related bugs and verified that all the commands now work as expected.

TerryHowe:
Last commit is not signed off, DCO failed.

alikhil:
@TerryHowe sorry for that. It's signed now

alikhil:
@banjoh could you look at this PR again, please?

alikhil:
Checking in to see if there are any thoughts or updates. Thanks!

alikhil:
The [adjacent pull request](https://github.com/helm/helm/pull/31524) has received more attention and will most likely be merged. It's a pity we couldn't revive and merge this original pull request. I'm closing it.

PR #30859, opened 2025-05-15
feat(engine): add checkHostByName to resolve hostnames with DNS error…

asmabahri:
### What this MR does

This introduces a function `checkHostByName` which resolves a hostname to its first IP address, and **fails rendering** if DNS resolution fails due to network issues or DNS errors like SERVFAIL or timeout.

---

### Why this is needed

We rely on conditional logic in templates, and Sprig's existing `getHostByName` does not handle DNS errors — it silently returns an empty string. This can lead to incorrect or invalid rendering when DNS is misconfigured or unreachable.

We created this patch internally to ensure rendering fails clearly when DNS infrastructure is broken.

However, we are not looking to maintain a long-term patch. We’d like to see Helm offer a safer alternative to `getHostByName`, like this function, or provide guidance on the correct Helm-native way to safely detect and handle DNS failures in templates.

This patch is intended to be upstreamable. If an alternative solution is preferred, we’re open to feedback and collaboration.

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -259,6 +259,9 @@ func (e Engine) initFunMap(t *template.Template) {
 		funcMap["getHostByName"] = func(_ string) string {
 			return ""
 		}
+		funcMap["checkHostByName"] = func(_ string) string {
+			return ""
+		}
 	}
 
 	// Set custom template funcs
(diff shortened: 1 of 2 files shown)


mattfarina (changes requested):
There is a certain class of attacks that can be performed through DNS. This is why there is a flag to opt-in to DNS checks. The other DNS function fails to return a response when it's not enabled. Any change, like this, also needs to follow that.

See https://github.com/helm/helm/blob/9aefd2b32809960203d0178c89522f6bf3d4af9f/pkg/engine/engine.go#L246

Issue #30880, opened 2025-05-20
typeIs/typeOf behavior has changed since helm 3.18.0


marcmillien:
With helm 3.18.0, there is a change in `typeIs`/`typeOf` gotemplate behavior.

For example in:
https://github.com/oauth2-proxy/manifests/blob/4001c2385dc72990add71437b02e18d7ec7f8ffd/helm/oauth2-proxy/templates/_ingress.tpl#L40-L44

While debugging, we saw that `typeOf .servicePort` has changed from `float64` to `json.Number`.

This could be a bad practice from the chart (that should probably use `KindIs`), but this is widely used and could break many charts.

Step to reproduce:

Helm 3.17.3:
```sh
$ helm repo add oauth2-proxy https://oauth2-proxy.github.io/manifests
"oauth2-proxy" has been added to your repositories

$ helm version
version.BuildInfo{Version:"v3.17.3", GitCommit:"e4da49785aa6e6ee2b86efd5dd9e43400318262b", GitTreeState:"clean", GoVersion:"go1.23.7"}

$ cat <<EOF | helm template demo oauth2-proxy/oauth2-proxy -f -
{
  "ingress": {
    "enabled": true,
    "hosts": ["demo.fr"]
  }
}
EOF
[...]
# Source: oauth2-proxy/templates/ingress.yaml
[...]
spec:
  rules:
    - host: "demo.fr"
      http:
        paths:
          - path: /
            pathType: ImplementationSpecific
            backend:
              service:
                name: demo-oauth2-proxy
                port:
                  number: 80                                         <==== OK
```

Helm 3.18.0:
```sh
$ helm version
version.BuildInfo{Version:"v3.18.0", GitCommit:"cc58e3f5a3aa615c6a86275e6f4444b5fdd3cc4e", GitTreeState:"clean", GoVersion:"go1.24.3"}

$ cat <<EOF | helm template demo oauth2-proxy/oauth2-proxy -f -
{
  "ingress": {
    "enabled": true,
    "hosts": ["demo.fr"]
  }
}
EOF
[...]
# Source: oauth2-proxy/templates/ingress.yaml
[...]
spec:
  rules:
    - host: "demo.fr"
      http:
        paths:
          - path: /
            pathType: ImplementationSpecific
            backend:
              service:
                name: demo-oauth2-proxy
                port:                                            <======= missing
```

came88:
I think this is an effect of #13534

mattfarina:
[...]

PR #30888, opened 2025-05-21
ci: Add test case for issue 30880

marcmillien:
Proposing a testcase for https://github.com/helm/helm/issues/30880

Here we add a test on `typeOf` that can be used in helm templates, so we validate there is no change in the behavior.

## In details (issue)

With helm 3.18.0, we had a change in `typeIs`/`typeOf` gotemplate behavior.

For example in:
https://github.com/oauth2-proxy/manifests/blob/4001c2385dc72990add71437b02e18d7ec7f8ffd/helm/oauth2-proxy/templates/_ingress.tpl#L40-L44

While debugging, we saw that `typeOf .servicePort` has changed from `float64` to `json.Number`.

This could be a bad practice from the chart (that should probably use `KindIs`), but this is widely used and could break many charts.

Step to reproduce:

Helm 3.17.3:
```sh
$ helm repo add oauth2-proxy https://oauth2-proxy.github.io/manifests
"oauth2-proxy" has been added to your repositories

$ helm version
version.BuildInfo{Version:"v3.17.3", GitCommit:"e4da49785aa6e6ee2b86efd5dd9e43400318262b", GitTreeState:"clean", GoVersion:"go1.23.7"}

$ cat <<EOF | helm template demo oauth2-proxy/oauth2-proxy -f -
{
  "ingress": {
    "enabled": true,
    "hosts": ["demo.fr"]
  }
}
EOF
[...]
# Source: oauth2-proxy/templates/ingress.yaml
[...]
spec:
  rules:
    - host: "demo.fr"
      http:
        paths:
          - path: /
            pathType: ImplementationSpecific
            backend:
              service:
                name: demo-oauth2-proxy
                port:
                  number: 80                                         <==== OK
```

Helm 3.18.0:
```sh
$ helm version
version.BuildInfo{Version:"v3.18.0", GitCommit:"cc58e3f5a3aa615c6a86275e6f4444b5fdd3cc4e", GitTreeState:"clean", GoVersion:"go1.24.3"}

$ cat <<EOF | helm template demo oauth2-proxy/oauth2-proxy -f -
{
  "ingress": {
    "enabled": true,
    "hosts": ["demo.fr"]
  }
}
EOF
[...]
# Source: oauth2-proxy/templates/ingress.yaml
[...]
spec:
  rules:
    - host: "demo.fr"
      http:
        paths:
          - path: /
            pathType: ImplementationSpecific
            backend:
              service:
                name: demo-oauth2-proxy
                port:                                            <======= missing
```

**If applicable**:

### DIFF
--- a/pkg/engine/funcs_test.go
+++ b/pkg/engine/funcs_test.go
@@ -127,6 +127,14 @@ keyInElement1 = "valueInElement1"`,
 		tpl:    `{{ lookup "v1" "Namespace" "" "unlikelynamespace99999999" }}`,
 		expect: `map[]`,
 		vars:   `["one", "two"]`,
+	}, {
+		tpl:    `{{ typeOf . }}`,
+		expect: `string`,
+		vars:   `"WEB"`,
+	}, {
+		tpl:    `{{ typeOf . }}`,
+		expect: `int`,
+		vars:   `80`
 	}}
 
 	for _, tt := range tests {


TerryHowe:
While mentioning the issue is useful, so actual prose about what this PR is about would be useful as well.

marcmillien:
> While mentioning the issue is useful, so actual prose about what this PR is about would be useful as well.

Thanks :), I did with also a better commit message content.
I am asking myself, should I move the target branch to `dev-v3` ?

PR #30886, opened 2025-05-20
Log warning when platformCommand|Hook and Command|Hook are set

benoittgt:
same for `PlatformHooks` and `hooks`.

**What this PR does / why we need it**:

When [backporting a feature for Helm v4](https://github.com/helm/helm/pull/13592) we forgot to mark as deprecation validations what will be enforced in v4. It was instand error even if [documentation say something different](https://helm.sh/docs/topics/plugins/#the-pluginyaml-file).

See the PR (search for "Remove in Helm 4."):
- https://github.com/helm/helm/pull/13592/files#diff-fc137d62147eb6a9c89c0566479ff219320ca691f2505631ebacbe513bcc8108R268-R270

**I am wondering how much enforcing like this in the PR was a good idea. Maybe we wanted only to inform, that soon helm will no longer support both attributes or `command` or `hooks`**.

My patch, **is probably not what we want** but it fixes #30876. Do we want to enforce this change in v3.18.0? If yes we can close this PR.

Before:

```sh
make build && bin/helm version && bin/helm plugin install https://github.com/jkroepke/helm-secrets --version v4.6.4 --debug                                                     1 ↵
make: Nothing to be done for `build'.
failed to load plugins: both platformCommand and command are set in "/Users/benoit.tigeot/Library/helm/plugins/helm-secrets/plugin.yaml"
version.BuildInfo{Version:"v3.18+unreleased", GitCommit:"e6122aba27fde30306dac79865df3189c5ff3be7", GitTreeState:"dirty", GoVersion:"go1.24.0"}
failed to load plugins: both platformCommand and command are set in "/Users/benoit.tigeot/Library/helm/plugins/helm-secrets/plugin.yaml"
Error: plugin already exists
helm.go:92: 2025-05-20 22:13:45.12691 +0200 CEST m=+0.138222667 [debug] plugin already exists
helm.sh/helm/v3/pkg/plugin/installer.Install
        helm.sh/helm/v3/pkg/plugin/installer/installer.go:53
main.(*pluginInstallOptions).run
        helm.sh/helm/v3/cmd/helm/plugin_install.go:78
main.newPluginInstallCmd.func3
        helm.sh/helm/v3/cmd/helm/plugin_install.go:59
github.com/spf13/cobra.(*Command).execute
        github.com/spf13/cobra@v1.9.1/command.go:1015
github.com/spf13/cobra.(*Command).ExecuteC
        github.com/spf13/cobra@v1.9.1/command.go:1148
[...]

PR #30891, opened 2025-05-21
fix: Port pluginCommand & command warning

gjenkins8:

Port the fix from Helm 3 to `main` to not error if `platformCommand` and `command` (and `platformHook` and `hook`) are set:
https://github.com/helm/helm/pull/30886

**If applicable**:

### DIFF
--- a/pkg/plugin/plugin.go
+++ b/pkg/plugin/plugin.go
@@ -83,7 +83,7 @@ type Metadata struct {
 	PlatformCommand []PlatformCommand `json:"platformCommand"`
 
 	// Command is the plugin command, as a single string.
-	// Providing a command will result in an deprecation warning if PlatformCommand is also set.
+	// Providing Command and PlatformCommand will result in a warning being emitted (PlatformCommand takes precedence).
 	//
 	// The command will be passed through environment expansion, so env vars can
 	// be present in this command. Unless IgnoreFlags is set, this will
@@ -119,7 +119,7 @@ type Metadata struct {
 	PlatformHooks PlatformHooks `json:"platformHooks"`
 
 	// Hooks are commands that will run on plugin events, as a single string.
-	// Providing a command will result in an deprecation warning if PlatformHooks is also set.
+	// Providing Hook and PlatformHooks will result in a warning being emitted (PlatformHooks takes precedence).
 	//
 	// The command will be passed through environment expansion, so env vars can
 	// be present in this command.
--- a/pkg/plugin/plugin.go
+++ b/pkg/plugin/plugin.go
@@ -83,7 +83,7 @@ type Metadata struct {
 	PlatformCommand []PlatformCommand `json:"platformCommand"`
 
 	// Command is the plugin command, as a single string.
-	// Providing a command will result in an error if PlatformCommand is also set.
+	// Providing a command will result in an deprecation warning if PlatformCommand is also set.
 	//
 	// The command will be passed through environment expansion, so env vars can
 	// be present in this command. Unless IgnoreFlags is set, this will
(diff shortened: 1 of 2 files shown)


scottrigby (dismissed):
Thanks for porting this v3 regression fix forward to Helm 4 👍

benoittgt (approved):
LGTM

robertsirc (dismissed):
LGTM

TerryHowe (approved):
/lgtm

scottrigby (dismissed):

scottrigby:
Not merging yet until PR is out of draft

pkg/plugin/plugin.go
```diff
@@ -92,7 +92,7 @@ type Metadata struct {
 	// Note that command is not executed in a shell. To do so, we suggest
 	// pointing the command to a shell script.
 	//
-	// DEPRECATED: Use PlatformCommand instead. Remove in Helm 4.
+	// DEPRECATED: Use PlatformCommand instead
```
TerryHowe, line 96: Is this going to be removed in Helm 4 or Helm 5?

gjenkins8: I think removing is more likely to be done in a future plugin version (rather than Helm version).

pkg/plugin/plugin.go
```diff
@@ -119,14 +119,14 @@ type Metadata struct {
 	PlatformHooks PlatformHooks `json:"platformHooks"`
 
 	// Hooks are commands that will run on plugin events, as a single string.
-	// Providing a hooks will result in an error if PlatformHooks is also set.
+	// Providing a command will result in an deprecation warning if PlatformHooks is also set.
```
TerryHowe: Should this be something along these lines?
```suggestion
	// Providing a platform and plugin hook will result in an deprecation warning.
```

pkg/plugin/plugin.go
```diff
@@ -83,7 +83,7 @@ type Metadata struct {
 	PlatformCommand []PlatformCommand `json:"platformCommand"`
 
 	// Command is the plugin command, as a single string.
-	// Providing a command will result in an error if PlatformCommand is also set.
+	// Providing a command will result in an deprecation warning if PlatformCommand is also set.
```
TerryHowe:
```suggestion
	// Providing a platform and plugin command will result in an deprecation warning.
```

pkg/plugin/plugin.go
```diff
@@ -265,11 +265,11 @@ func validatePluginData(plug *Plugin, filepath string) error {
 	plug.Metadata.Usage = sanitizeString(plug.Metadata.Usage)
 
 	if len(plug.Metadata.PlatformCommand) > 0 && len(plug.Metadata.Command) > 0 {
-		return fmt.Errorf("both platformCommand and command are set in %q", filepath)
+		fmt.Printf("WARNING: both 'platformCommand' and 'command' are set in %q (this will become an error in a future Helm version)\n", filepath)
```
TerryHowe: slog for these for v4?

TerryHowe (changes requested):

pkg/plugin/plugin.go
gjenkins8: https://github.com/helm/helm/pull/30891#issuecomment-2974726358

I'm not sure. If we have a convention for warnings with slog, we can utilize it. But at this point I don't think we do? (And I was not wanting to create one during this PR; these can easily be converted in the future)

pkg/plugin/plugin.go
gjenkins8: ah. yeah. this was just lifted from the `dev-v3` PR. It isn't very well worded

gjenkins8, line 122:
```suggestion
	// Providing the 'hooks' and 'platformHooks' field will result in a deprecation warning. `platformHooks` will take precedence.
```

pkg/plugin/plugin.go
TerryHowe: @robertsirc was working on logging and some of his PRs merged with just `slog.Warn(...)` I think that is a fair approach and you don't have to pass around the logger or create a singleton.

gjenkins8: sure, switched to `slog.Warn(...)`

TerryHowe (approved):
/lgtm

scottrigby (approved):
still lgtm

PR #30889, opened 2025-05-21
[v3 backport] Prevent push cmd failure in 3.18 by handling version tag resolution in ORAS memory store

benoittgt:
**What this PR does / why we need it**:

Fix: https://github.com/helm/helm/issues/30881

Backport of https://github.com/helm/helm/pull/30894

**Special notes for your reviewer**:

**If applicable**:
- [ ] this PR contains user facing changes (the `docs needed` label should be applied if so)
- [ ] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

PR #30894, opened 2025-05-21
Prevent push cmd failure in 3.18 by handling version tag resolution in ORAS memory store

### LINKED ISSUE #30881: helm push error 3.18.0
Since version 3.18.0, we observe an error during the deployment process which didn't exist in previous versions. `helm push` ends with an error:
```
Error: not found
```
There were no changes in our deployment process between Helm versions.

benoittgt:

Fix: https://github.com/helm/helm/issues/30881

Initial work for `dev-v3` in https://github.com/helm/helm/pull/30889#issuecomment-2898909440 but as requested by @gjenkins8, we should backport after on v3. :)

The issue is:
- The `newReference()` function transforms version tags by replacing + with _ for OCI compatibility (as we can see in the output message)
- But the code was using the original ref (with +) for `TagBytes()`
- Then it tries to find the tagged reference using `parsedRef.String()` (with _)
- This mismatch causes the `Resolve` method to fail with "not found"
- By using `parsedRef.String()` consistently in both places, the references will match and the lookup will succeed.

I extracted the `TagBytes` function to improve testability. `Push()` includes several external calls that are hard to mock, so isolating this logic makes testing more manageable.

The patch could be summarized as:
```diff
- oras.TagBytes(ctx, memoryStore, ocispec.MediaTypeImageManifest, manifestData, ref)
+ oras.TagBytes(ctx, memoryStore, ocispec.MediaTypeImageManifest, manifestData, parsedRef.String())
```

The initial regression is something that was missed in https://github.com/helm/helm/commit/5a7046b9bfaf1d7e97c555e33fa8744bba758004#diff-acbc1e889c3e91003a88f18e21c3073b64937801b20f663fcf2601540b9e3cdaR709

Before :
```sh
» ../../helm/bin/helm push ./test-0.0.0-snapshot+dirty.tgz oci://ghcr.io/benoittgt/helm-charts/test --debug
Error: failed to perform "Resolve" on source: ghcr.io/benoittgt/helm-charts/test/test:0.0.0-snapshot_dirty: not found
helm.go:92: 2025-05-21 16:41:49.581662 +0200 CEST m=+0.041508960 [debug] failed to perform "Resolve" on source: ghcr.io/benoittgt/helm-charts/test/test:0.0.0-snapshot_dirty: not found
```

After :

```sh
» ../../helm/bin/helm push ./test-0.0.0-snapshot+dirty.tgz oci://ghcr.io/benoittgt/helm-charts/test --debug
Pushed: ghcr.io/benoittgt/helm-charts/test/test:0.0.0-snapshot_dirty
Digest: sha256:680f23c71cb6727ae8845daf01f019dc11bdac4b652289c9a970b270c56ba79d
ghcr.io/benoittgt/helm-charts/test/test:0.0.0-snapshot_dirty contains an underscore.

OCI artifact references (e.g. tags) do not support the plus sign (+). To support
storing semantic versions, Helm adopts the convention of changing plus (+) to
an underscore (_) in chart version tags when pushing to a registry and back to
a plus (+) when pulling from a registry.
```

**If applicable**:
- [ ] this PR has been tested for backwards compatibility (it will be backported)

### DIFF
--- a/pkg/registry/client.go
+++ b/pkg/registry/client.go
@@ -686,8 +686,8 @@ func (c *Client) Push(data []byte, ref string, options ...PushOption) (*PushResu
 
 	ociAnnotations := generateOCIAnnotations(meta, operation.creationTime)
 
-	manifestDescriptor, err := c.tagManifest(ctx, memoryStore, ref, configDescriptor,
-		layers, ociAnnotations)
+	manifestDescriptor, err := c.tagManifest(ctx, memoryStore, configDescriptor,
+		layers, ociAnnotations, parsedRef)
 	if err != nil {
 		return nil, err
 	}
@@ -891,8 +891,8 @@ func (c *Client) ValidateReference(ref, version string, u *url.URL) (*url.URL, e
 
 // tagManifest prepares and tags a manifest in memory storage
 func (c *Client) tagManifest(ctx context.Context, memoryStore *memory.Store,
-	ref string, configDescriptor ocispec.Descriptor, layers []ocispec.Descriptor,
-	ociAnnotations map[string]string) (ocispec.Descriptor, error) {
+	configDescriptor ocispec.Descriptor, layers []ocispec.Descriptor,
+	ociAnnotations map[string]string, parsedRef reference) (ocispec.Descriptor, error) {
 
 	manifest := ocispec.Manifest{
 		Versioned:   specs.Versioned{SchemaVersion: 2},
@@ -902,11 +902,6 @@ func (c *Client) tagManifest(ctx context.Context, memoryStore *memory.Store,
 	}
 
 	manifestData, err := json.Marshal(manifest)
-	if err != nil {
-		return ocispec.Descriptor{}, err
-	}
-
-	parsedRef, err := newReference(ref)
 	if err != nil {
 		return ocispec.Descriptor{}, err
 	}
--- a/pkg/registry/client.go
+++ b/pkg/registry/client.go
@@ -685,19 +685,9 @@ func (c *Client) Push(data []byte, ref string, options ...PushOption) (*PushResu
 	})
 
 	ociAnnotations := generateOCIAnnotations(meta, operation.creationTime)
-	manifest := ocispec.Manifest{
-		Versioned:   specs.Versioned{SchemaVersion: 2},
-		Config:      configDescriptor,
-		Layers:      layers,
-		Annotations: ociAnnotations,
-	}
 
-	manifestData, err := json.Marshal(manifest)
-	if err != nil {
-		return nil, err
-	}
-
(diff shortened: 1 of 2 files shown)


pkg/registry/client.go
```diff
@@ -685,19 +685,9 @@ func (c *Client) Push(data []byte, ref string, options ...PushOption) (*PushResu
 	})
 
 	ociAnnotations := generateOCIAnnotations(meta, operation.creationTime)
```
TerryHowe, line 687: I think the `parsedRef, err := newReference(ref)` above should be removed

TerryHowe (changes requested):
Need to remove the unused parsedRef

pkg/registry/client.go
```diff
@@ -898,3 +888,29 @@ func (c *Client) ValidateReference(ref, version string, u *url.URL) (*url.URL, e
...
+		return ocispec.Descriptor{}, err
+	}
+
+	parsedRef, err := newReference(ref)
+	if err != nil {
+		return ocispec.Descriptor{}, err
+	}
+
+	return oras.TagBytes(ctx, memoryStore, ocispec.MediaTypeImageManifest,
+		manifestData, parsedRef.String())
```
TerryHowe, line 915: This is the only actual change here I think, using `parsedRef.String()` instead of `ref` although the tests are nice.

benoittgt: Yes, exactly. I should have specified it.

benoittgt: Simplified it's this.

```diff
- oras.TagBytes(ctx, memoryStore, ocispec.MediaTypeImageManifest, manifestData, ref)
+ oras.TagBytes(ctx, memoryStore, ocispec.MediaTypeImageManifest, manifestData, parsedRef.String())
```

pkg/registry/client.go
benoittgt: Not sure I get it. The `parsedRef` is used here ? 🤔

TerryHowe: Line 611. It won't let me comment up there

pkg/registry/client.go
TerryHowe: Yeh, this PR could in theory be just that one line change the way I read it.

pkg/registry/client.go
benoittgt: Is that what you had in mind? https://github.com/helm/helm/pull/30894/commits/2c01f66d8b7d455ba9962018517d2bc87da15d2b

TerryHowe: Ah, yeh, that works

TerryHowe (approved):
/lgtm

mattfarina (approved):
lgtm

Also manually tested.

robertsirc (approved):
LGTM

Issue #30873, opened 2025-05-19
helm registry login failing when updated cli to v3.18.0


athornton2012:
In our CI we auth to a GCP artifact registry by running 

`gcloud auth print-access-token | helm registry login -u oauth2accesstoken --password-stdin https://us-docker.pkg.dev`

When we bumped to the new version of the helm cli we started to see the error message

`Error: invalid reference: invalid registry "https://us-docker.pkg.dev/"`

DmytroDerkach:
I had the same issue with helm, just remove `https://`

`gcloud auth print-access-token | helm registry login -u oauth2accesstoken --password-stdin us-docker.pkg.dev`

benoittgt:
Hello

It seems related to oras ref requirement in [oras-go](https://github.com/oras-project/oras-go/blob/05a2b09cbf2eab1df691411884dc4df741ec56ab/registry/reference.go#L195-L196), and OCI [repository requirement](https://github.com/oras-project/oras-go/blob/05a2b09cbf2eab1df691411884dc4df741ec56ab/registry/reference.go#L38C41-L38C127). 

But I think the error could be clearer.

gershonkoral:
> I had the same issue with helm, just remove `https://`

Unfortunately this did not work. Rolled back to v17

gjenkins8:
ACR also seems to be affected, from https://github.com/helm/helm/issues/30883

--- 

since 3.18.0 our pipeline stopped working on helm repo authentication. I can reproduce that locally from the 3.18.0 tag

helm 3.18.0:

az acr login -n REDACTED.azurecr.io --expose-token -o tsv --query accessToken 2>/dev/null | go run ./cmd/helm registry login REDACTED.azurecr.io  --password-stdin 
Error: authenticating to "REDACTED.azurecr.io": GET "https://REDACTED.azurecr.io/v2/": GET "https://REDACTED.azurecr.io/oauth2/token?service=REDACTED.azurecr.io": response status code 401: unauthorized: authentication required, visit https://aka.ms/acr/authorization for more information. CorrelationId: b65a8fc5-6038-4ca3-aa92-46b2b6976849
exit status 1
helm 3.17.3:

az acr login -n REDACTED.azurecr.io --expose-token -o tsv --query accessToken 2>/dev/null | helm registry login REDACTED.azurecr.io  --password-stdin
Login Succeeded

gjenkins8:
[...]

PR #30897, opened 2025-05-22
fix: strip repository for login for backwards compatibility

TerryHowe:
In order to maintain backward compatibility for oras-go v2 for 3.18, we need to strip the repository off the url provided on the login command line. The protocol also needs to be stripped off which is handled in https://github.com/helm/helm/pull/30887

Partial https://github.com/helm/helm/issues/30873

**If applicable**:

### DIFF
--- a/pkg/registry/client.go
+++ b/pkg/registry/client.go
@@ -229,8 +229,23 @@ type (
 	}
 )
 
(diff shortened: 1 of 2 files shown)


scottrigby:
incorporated into https://github.com/helm/helm/pull/30887 ✅

PR #30900, opened 2025-05-22
Add timeout flag to repo add and update commands

### LINKED ISSUE #12952: Command "repository add" raises "Client.Timeout or context cancellation while reading body"

unguiculus:

Adds `--timeout` flags to `repo add` and `repo up` commands.
Fixes: #12952

This PR supersedes #13185. Adding an environment variable is not ideal IMO. Using CLI flag is more appropriate and consistent with what other commands do.

**If applicable**:

### DIFF
--- a/pkg/getter/getter.go
+++ b/pkg/getter/getter.go
@@ -191,24 +191,32 @@ const (
 
 var defaultOptions = []Option{WithTimeout(time.Second * DefaultHTTPTimeout)}
 
-var httpProvider = Provider{
-	Schemes: []string{"http", "https"},
-	New: func(options ...Option) (Getter, error) {
-		options = append(options, defaultOptions...)
-		return NewHTTPGetter(options...)
-	},
-}
-
-var ociProvider = Provider{
-	Schemes: []string{registry.OCIScheme},
-	New:     NewOCIGetter,
+func Getters(extraOpts ...Option) Providers {
+	return Providers{
+		Provider{
+			Schemes: []string{"http", "https"},
+			New: func(options ...Option) (Getter, error) {
+				options = append(options, defaultOptions...)
+				options = append(options, extraOpts...)
+				return NewHTTPGetter(options...)
+			},
+		},
+		Provider{
+			Schemes: []string{registry.OCIScheme},
+			New: func(options ...Option) (Getter, error) {
+				options = append(options, defaultOptions...)
+				options = append(options, extraOpts...)
+				return NewOCIGetter(options...)
+			},
+		},
+	}
 }
 
 // All finds all of the registered getters as a list of Provider instances.
 // Currently, the built-in getters and the discovered plugins with downloader
 // notations are collected.
-func All(settings *cli.EnvSettings) Providers {
-	result := Providers{httpProvider, ociProvider}
+func All(settings *cli.EnvSettings, opts ...Option) Providers {
+	result := Getters(opts...)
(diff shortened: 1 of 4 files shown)


robertsirc:
I am not sure this is a good PR for v3.x rather it should go to v4.x.

unguiculus:
Why would you want to push this out to Helm 4? IMO this is nothing that needs to wait. It's a non-breaking change. We are really suffering from this.

pkg/getter/getter.go
```diff
@@ -196,24 +196,32 @@ const (
...
+		},
+	}
 }
 
 // All finds all of the registered getters as a list of Provider instances.
 // Currently, the built-in getters and the discovered plugins with downloader
 // notations are collected.
-func All(settings *cli.EnvSettings) Providers {
-	result := Providers{httpProvider, ociProvider}
+func All(settings *cli.EnvSettings, opts ...Option) Providers {
```
scottrigby, line 223: ### Note in advance of a v3 backport

On the v3 backport PR, we can't change the function signature for [getter.All] in Helm 3, because it would be a breaking change. It would be fine for Helm 4 though, since it's a major version change and there will be an SDK migration guide.

To not break backwards compatibility in Helm 3 you could add a wrapper function, like [getter.AllWithOpts] or something, and change the action to call that instead of [getter.All]. Then change getter.All to wrap [getter.AllWithOpts] with empty opts so it won't break SDK users existing calls to All.

scottrigby (commented):
https://github.com/helm/helm/issues/12952 is a `bug` that affects users when Helm HTTP repos are sufficiently large. To address the error, users should be able to specify a timeout in some way. This PR is a proposal for how to fix that bug.

## Process note

> Why would you want to push this out to Helm 4? IMO this is nothing that needs to wait. It's a non-breaking change. We are really suffering from this.

Robert is right that this should go into Helm 4 first. That doesn't necessarily mean it should not be backported to Helm 3 though. We are currently in a feature freeze for Helm 3, but we are supporting bug and security fixes according to this plan https://github.com/helm/community/blob/main/hips/hip-0012.md#maintaining-helm-3.

So the process is you should [change the base branch](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-base-branch-of-a-pull-request) of this PR to `main` (where Helm 4 is being developed). Once merged into main, big fixes can be backported to Helm 3 by opening a new PR against the `dev-v3` branch.

unguiculus:
Thanks @scottrigby, I'll rebase the PR.

unguiculus:
I rebased and changed the base branch.

scottrigby (approved):
LGTM 👍

gjenkins8 (approved):

PR #30902, opened 2025-05-22
fix(client): allow layer selection permissive 

### LINKED ISSUE #30890: Helm v3.18.0: OCI Image pull error
After upgrading to Helm v3.18.0 a simple oci Helm Chart pull results in:

```
helm pull oci://public.ecr.aws/dynatrace/dynatrace-operator --version 1.5.1
 Error: media type "application/vnd.cncf.helm.chart.provenance.v1.prov" is not allowed, found in descriptor with digest: "sha256:812020c07e5e758fbc5cf2cec2640aed415c08a70b9801451a5607e4c5528ede"
``` 
After downgrading again to v3.17.x the pull works without any problems.

brandtkeller:

As noted in the issue below - the switch to ORAS V2 required implementation of the `PreCopy` function which checks that a descriptors `mediatype` is in the allowed list and otherwise returns an error. I believe the ORAS v1 logic was more permissive and rather did not include layers if they were not in the allowedMediaTypes list. 

In order to meet parity with `v3.17.3` with regards to OCI copy operations - I believe that this should instead be returning nil. 

additionally the manifest is already included in the layers and does not need to be appended to descriptors. 

closes #30890

**If applicable**:

### DIFF
--- a/pkg/registry/client.go
+++ b/pkg/registry/client.go
@@ -459,7 +459,7 @@ func (c *Client) Pull(ref string, options ...PullOption) (*PullResult, error) {
 			PreCopy: func(_ context.Context, desc ocispec.Descriptor) error {
 				mediaType := desc.MediaType
 				if i := sort.SearchStrings(allowedMediaTypes, mediaType); i >= len(allowedMediaTypes) || allowedMediaTypes[i] != mediaType {
-					return nil
+					return oras.SkipNode
 				}
 
 				mu.Lock()
--- a/pkg/registry/client.go
+++ b/pkg/registry/client.go
@@ -473,7 +473,6 @@ func (c *Client) Pull(ref string, options ...PullOption) (*PullResult, error) {
 		return nil, err
 	}
 
-	descriptors = append(descriptors, manifest)
 	descriptors = append(descriptors, layers...)
 
 	numDescriptors := len(descriptors)
--- a/pkg/registry/client.go
+++ b/pkg/registry/client.go
@@ -459,7 +459,7 @@ func (c *Client) Pull(ref string, options ...PullOption) (*PullResult, error) {
 			PreCopy: func(_ context.Context, desc ocispec.Descriptor) error {
 				mediaType := desc.MediaType
 				if i := sort.SearchStrings(allowedMediaTypes, mediaType); i >= len(allowedMediaTypes) || allowedMediaTypes[i] != mediaType {
-					return errors.Errorf("media type %q is not allowed, found in descriptor with digest: %q", mediaType, desc.Digest)
+					return nil
 				}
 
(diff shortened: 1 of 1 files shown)


pkg/registry/client.go
```diff
@@ -459,7 +459,7 @@ func (c *Client) Pull(ref string, options ...PullOption) (*PullResult, error) {
 			PreCopy: func(_ context.Context, desc ocispec.Descriptor) error {
 				mediaType := desc.MediaType
 				if i := sort.SearchStrings(allowedMediaTypes, mediaType); i >= len(allowedMediaTypes) || allowedMediaTypes[i] != mediaType {
-					return errors.Errorf("media type %q is not allowed, found in descriptor with digest: %q", mediaType, desc.Digest)
+					return nil
```
gjenkins8, line 462: From ORAS code/docs:
```
        // PreCopy handles the current descriptor before it is copied. PreCopy can
	// return a SkipNode to signal that desc should be skipped when it already
	// exists in the target.
	PreCopy func(ctx context.Context, desc ocispec.Descriptor) error
```

Returning nil will cause (iiuc) ORAS to copy the layer. Which Helm is not expecting, nor wants/needs to occur. I think we want to return `SkipNode` instead, to avoid the unneeded (e.g. provenance layer being copied)

@TerryHowe  -- that sound correct?

```suggestion
					return oras.SkipNode
```

sabre1041: @gjenkins8 you are correct. we do want `oras.SkipNode` as the resolution here as this type of functionality was how it was handled prior to the incorporation of ORAS v2. the ORAS library took care of it when `oras.WithAllowedMediaTypes` was provided

pkg/registry/client.go
```diff
@@ -473,7 +473,6 @@ func (c *Client) Pull(ref string, options ...PullOption) (*PullResult, error) {
 		return nil, err
 	}
 
-	descriptors = append(descriptors, manifest)
```
gjenkins8, line 476: We should not be removing this?

brandtkeller: The manifest descriptor exists within 'layers' to my knowledge. I can re-add this if that layer descriptor duplication is intentional

robertsirc (approved):
LGTM

robertsirc:
I can forward port this if needed.

mattfarina (approved):
lgtm

Manually tested and it works

PrivatePuffin:
@mattfarina This bug  *completely!* broke OCI+Provenance installs for everyone on production software (Helm), why is the hot fix not out yet?!

mattfarina:
@PrivatePuffin we still have a couple known regressions. A partial fix release is coming that will include this.

PR #30904, opened 2025-05-23
[Doc] Help users avoid specifying URL scheme and path with `helm registry`

benoittgt:

We’ve noticed that some users [still include the URL scheme](https://github.com/helm/helm/issues/30873) and full path when logging into an OCI registry, for example:

```sh
helm registry login -u $OCI_REGISTRY_USER --password-stdin oci://ghcr.io/org/repo
```

This is no longer necessary and will not be supported in Helm v4.

To guide users toward the correct usage, we should show an example of the ideal command, as[ it was suggested by @scottrigby](https://github.com/helm/helm/issues/30873#issuecomment-2903202958).

Do we want to backport? I think it could be helpful for user to migrate to Helm v4 ?

**If applicable**:

### DIFF
--- a/pkg/cmd/registry_login.go
+++ b/pkg/cmd/registry_login.go
@@ -33,6 +33,10 @@ import (
 
 const registryLoginDesc = `
 Authenticate to a remote registry.
+
+For example for Github Container Registry:
+
+    echo "$GITHUB_TOKEN" | helm registry login ghcr.io -u $GITHUB_USER --password-stdin
 `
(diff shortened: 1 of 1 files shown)


robertsirc (approved):
LGTM

robertsirc:
I don' think this is a backport as we wouldn't want this change back in v3.

TerryHowe (commented):
Good addition, thanks.

PR #30910, opened 2025-05-24
fix: ignore empty credentials files for v3

TerryHowe:
Related: https://github.com/fluxcd/source-controller/pull/1805

**If applicable**:

### DIFF
--- a/pkg/registry/client.go
+++ b/pkg/registry/client.go
@@ -115,13 +115,29 @@ func NewClient(options ...ClientOption) (*Client, error) {
 	}
 	store, err := credentials.NewStore(client.credentialsFile, storeOptions)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


sabre1041:
> Related to [fluxcd/source-controller#1805](https://github.com/fluxcd/source-controller/pull/1805) this works, but I'm not thrilled with this approach. Going to make this a draft.

This is a bit of a "hack" and I feel like taking a step back and seeing one options are available is the recommended path

scottrigby:
closing in favor of https://github.com/helm/helm/pull/30932

mattfarina:
FYI https://github.com/helm/helm/pull/30932 built on the ideas in this PR. Just tried to do less work and handle the case or writes.

PR #30913, opened 2025-05-26
Added goreleaser instead of gox #30896

### LINKED ISSUE #30896: Replace `github.com/mitchellh/gox`
Helm uses https://github.com/mitchellh/gox for cross-platform builds:

https://github.com/helm/helm/blob/0dffc580b0aec957b686fc08be681d3f74707749/Makefile#L170-L171
https://github.com/helm/helm/blob/0dffc580b0aec957b686fc08be681d3f74707749/Makefile#L181-L182

However, `github.com/mitchellh/gox` is archived and no longer supported. And should be replaced.

Bhargavkonidena:
Resolves #30896 

**If applicable**:


.github/workflows/release.yml
```diff
@@ -31,14 +31,26 @@ jobs:
...
         run: make test-coverage
+
       - name: Build Helm Binaries
         run: |
           set -eu -o pipefail
 
-          make build-cross VERSION="${{ github.ref_name }}"
-          make dist checksum VERSION="${{ github.ref_name }}"
+      - name: Run Go Release
+        uses: goreleaser/goreleaser-action@v2
```
gjenkins8, line 43: Can we use the latest version (6.3.0) please, and pin to that specific release via hash (see an example `uses:` above)

.github/workflows/release.yml
```diff
@@ -93,10 +105,18 @@ jobs:
       - name: Run unit tests
         run: make test-coverage
 
-      - name: Build Helm Binaries
+      - name: Run Go Release
+        uses: goreleaser/goreleaser-action@v2
```
gjenkins8, line 109: (same as above)

gjenkins8 (commented):
Thanks for the PR! I took an initial pass, I still need to review some of the specifics. But some initial comments:

- Please don't remove `.github/issue_template.md`. That should be done in a separate and specific PR.
- can we update the `build-cross` target in `Makefile` to also use `goreleaser`, and any related/then unused functionality e.g. `TARGETS`, `TARGET_OBJS`, `GOX` (and download), etc
  - the `test-acceptance` target will need to be updated somehow. currently this only requires a build of `linux/amd64`. ideally we could emulate this logic

plus the comments inline on the PR. Thanks!

Bhargavkonidena:
> Thanks for the PR! I took an initial pass, I still need to review some of the specifics. But some initial comments:
> 
> * Please don't remove `.github/issue_template.md`. That should be done in a separate and specific PR.
> * can we update the `build-cross` target in `Makefile` to also use `goreleaser`, and any related/then unused functionality e.g. `TARGETS`, `TARGET_OBJS`, `GOX` (and download), etc
>   
>   * the `test-acceptance` target will need to be updated somehow. currently this only requires a build of `linux/amd64`. ideally we could emulate this logic
> 
> plus the comments inline on the PR. Thanks!

I am done with incorporating these review comments. Please check.

Bhargavkonidena:
@gjenkins8 can you please review this PR

robertsirc:
@mattfarina and I can review this on the next Helmternoon.

robertsirc (dismissed):
LGTM

Makefile
```diff
@@ -167,19 +163,20 @@ gen-test-golden: test-unit
...
 # ------------------------------------------------------------------------------
 #  release
 
 .PHONY: build-cross
-build-cross: LDFLAGS += -extldflags "-static"
-build-cross: $(GOX)
-	GOFLAGS="-trimpath" CGO_ENABLED=0 $(GOX) -parallel=3 -output="_dist/{{.OS}}-{{.Arch}}/$(BINNAME)" -osarch='$(TARGETS)' $(GOFLAGS) -tags '$(TAGS)' -ldflags '$(LDFLAGS)' ./cmd/helm
+build-cross:
+	@echo "==> Building cross-platform binaries using GoReleaser"
+	goreleaser build --clean --skip-validate --single-target
```
gjenkins8, line 175: `--skip-validate` is deprecated -- see: https://goreleaser.com/deprecations/

don't you want/need to refer to the config file? And why is `--single-target` being passed here?

Makefile
```diff
@@ -167,19 +163,20 @@ gen-test-golden: test-unit
...
+
+.PHONY: build-linux-amd64
+build-linux-amd64:
+	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build $(GOFLAGS) -trimpath -tags '$(TAGS)' -ldflags '$(LDFLAGS)' -o '$(BINDIR)/$(BINNAME)' ./cmd/helm
```
gjenkins8, line 179: I think it would be better to use (for consistency) `goreleaser` w/ `--single-target`?

.github/workflows/.goreleaser.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+builds:
+  - id: helm
+    main: ./cmd/helm
+    binary: helm
+    goos:
+      - linux
+      - darwin
+      - windows
+    goarch:
```
gjenkins8, line 11: for `GOARCH` we previously had:
```
darwin/amd64 darwin/arm64 linux/amd64 linux/386 linux/arm linux/arm64 linux/ppc64le linux/s390x linux/riscv64 windows/amd64 windows/arm64
```

ie. `linux/386`, `linux/arm`, `linux/ppc64le`, `linux/s390x linux/riscv64` need to continue to be built, but are missing here.

.github/workflows/release.yml
```diff
@@ -31,14 +31,26 @@ jobs:
...
         run: make test-coverage
+
       - name: Build Helm Binaries
         run: |
           set -eu -o pipefail
 
-          make build-cross VERSION="${{ github.ref_name }}"
-          make dist checksum VERSION="${{ github.ref_name }}"
+      - name: Run Go Release
+        uses: goreleaser/goreleaser-action@v6.3.0
```
gjenkins8, line 43: we need to pin to a specific release via hash (see other `uses:`). Please also include the semver version as a comment.

.github/workflows/release.yml
```diff
@@ -93,10 +105,18 @@ jobs:
       - name: Run unit tests
         run: make test-coverage
 
-      - name: Build Helm Binaries
+      - name: Run Go Release
+        uses: goreleaser/goreleaser-action@v6.3.0
```
gjenkins8, line 109: as above (we need to pin to a specific release via hash (see other uses:). Please also include the semver version as a comment.)

Makefile
```diff
@@ -131,8 +128,7 @@ test-source-headers:
 	@scripts/validate-license.sh
 
 .PHONY: test-acceptance
-test-acceptance: TARGETS = linux/amd64
-test-acceptance: build build-cross
+test-acceptance: build build-linux-amd64
```
gjenkins8, line 131: `test-acceptance` previously built current arch + all architectures. Now it is building current + linux/amd64. 

If I'm not mistaken, we previously did this to:
- build all architectures to ensure all architectures build
- build current arch, to run acceptable tests

We need to continue the above behavior here (rather than building current + linux/amd64 only; runners are most likely to be linux/amd64, which makes this redundant too)

.github/workflows/release.yml
```diff
@@ -31,14 +31,26 @@ jobs:
...
+
       - name: Build Helm Binaries
         run: |
           set -eu -o pipefail
 
-          make build-cross VERSION="${{ github.ref_name }}"
-          make dist checksum VERSION="${{ github.ref_name }}"
+      - name: Run Go Release
+        uses: goreleaser/goreleaser-action@v6.3.0
+        if: success() && startsWith(github.ref, 'refs/tags/')
```
gjenkins8, line 44: I this necessary? given https://github.com/helm/helm/pull/30913/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34L19

.github/workflows/.goreleaser.yaml
```diff
@@ -0,0 +1,38 @@
+project_name: helm
```
gjenkins8, line 1: How are you testing this change? A superficial run fails with:
```
$ goreleaser build --clean --skip validate -f .github/workflows/.goreleaser.yaml 
  • only  version: 2  configuration files are supported, yours is  version: 0 , please update your configuration
  ⨯ build failed after 0s                    error=only  version: 2  configuration files are supported, yours is  version: 0 , please update your configuration
```

.github/workflows/release.yml
```diff
@@ -31,14 +31,26 @@ jobs:
...
+        with:
+          version: latest
+          args: release --rm-dist
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_PAT }}
+
+      - name: Verify checksum
+        run: |
+            make dist checksum VERSION="${{ github.ref_name }}"
```
gjenkins8, line 53: unless I'm mistaken, `make dist` is recreating the release archives. Which `.goreleaser.yaml` is also configured to do:
https://github.com/helm/helm/pull/30913/files#diff-413d66267886290506ba7e2df00c65cb65341ec7347cd34568117ff435f63623R25

please use one or the other (preference would be to use goreleaser, rather than makefile logic I think)

.github/workflows/.goreleaser.yaml
```diff
@@ -0,0 +1,38 @@
...
+      - arm64
+    goarm:
+      - 6
+    env:
+      - CGO_ENABLED=0
+      - GOFLAGS=-trimpath
+    ldflags:
+      - -s -w -extldflags "-static"
+    flags:
+      - -tags=netgo
```
gjenkins8, line 22: where did this "netgo" tag come from?

gjenkins8 (changes requested):
there are quite a few things to fix up here. once these are fixed, it is needed to test this change (as much as possible) locally. Then we can consider how to validate in github actions.

PR #30924, opened 2025-05-27
Oras v1 credentials file shim

scottrigby:

Fix for regression in v3.18.0
ORAS v1 was tolerant of an empty config file
ORAS v2 is not
to workaround this regression, ensure configfile is valid JSON

heavily inspired by https://github.com/helm/helm/pull/30910. Changed some handling logic based on investigation with other helm maintainers. Also isolated to a separate func to easily mark deprecated and for clearer diffs with `main`.

**If applicable**:

### DIFF
--- a/pkg/registry/client.go
+++ b/pkg/registry/client.go
@@ -107,7 +107,7 @@ func handleEmptyConfgFile(credentialsFile string, client *Client) {
 	if err == nil {
 		var configData map[string]json.RawMessage
 		// handle only if credentials file is empty
-		if err := json.NewDecoder(f).Decode(&configData); err != nil && !errors.Is(err, io.EOF) {
+		if err := json.NewDecoder(f).Decode(&configData); err != nil && errors.Is(err, io.EOF) {
 			// Attempt to write empty JSON map to config file
 			client.credentialsFile = ""
 			encoder := json.NewEncoder(f)
--- a/pkg/registry/client.go
+++ b/pkg/registry/client.go
@@ -90,10 +90,11 @@ type (
 
 // TODO remove in Helm 4
 // Deprecated
-// Helm 3 shim for ORAS v2 not tolerant of an empty config file
-// to workaround this, set configFile path to a non-existing path
-// so that ORAS v2 [config.Load] will handle this for us
-func orasV1CredentialsFileShim(credentialsFile string, client *Client) string {
+// Fix for regression in v3.18.0
+// ORAS v1 was tolerant of an empty config file
+// ORAS v2 is not
+// to workaround this regression, ensure configfile is valid JSON
+func handleEmptyConfgFile(credentialsFile string, client *Client) string {
 	f, err := os.Open(credentialsFile)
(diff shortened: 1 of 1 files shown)


benoittgt:
> Gonna ask the question: how is this to be tested?

Is it working with Flux CI  ? 😄

mattfarina (changes requested):
This doesn't fix the empty file issue. I still get the error `invalid config format: EOF`.

scottrigby:
> This doesn't fix the empty file issue. I still get the error `invalid config format: EOF`.

Fixed. Ready for manual re-test. Working on unit test now

mattfarina (changes requested):
`helm registry login` does not work with this change. Error is:

```
Error: failed to save config file: rename ./oras_credstore_temp_637423049 : no such file or directory
```

pkg/registry/client.go
```diff
@@ -88,6 +88,39 @@ type (
 	ClientOption func(*Client)
 )
 
+// TODO remove in Helm 4
+// Deprecated
+// Fix for regression in v3.18.0
+// ORAS v1 was tolerant of an empty config file
+// ORAS v2 is not
+// to workaround this regression, ensure configfile is valid JSON
+func handleEmptyConfgFile(credentialsFile string, client *Client) {
```
robertsirc, line 97: There is no test coverage for this.

robertsirc (changes requested):
We need to have test for this change.

pkg/registry/client.go
```diff
@@ -88,6 +88,45 @@ type (
...
+		}
+		// handle only if credentials file is empty
+		if fileInfo.Size() == 0 {
+			emptyMap := make(map[string]interface{})
+			jsonData, err := json.Marshal(emptyMap)
+			if err != nil {
+				client.err = err
+			}
+			// Attempt to write empty JSON map to config file
+			err = os.WriteFile(credentialsFile, jsonData, 0644)
```
mattfarina, line 120: Two issues with this...

1. It makes a configuration file for credentials readable to those other than the user. This should be 0600.
2. If the file is mounted from a read only file system, this will have an error. That would not happen with v3.17. This can happen if Helm is run in a container.

mattfarina (changes requested):

pkg/registry/client.go
scottrigby: it's almost as if that was already acknowledged with a plan if we go forward with this approach https://github.com/helm/helm/pull/30924#issuecomment-2916584926

pkg/registry/client.go
scottrigby: 1. agree on mode, that's an easy fix.
2. passing a non-writable, empty credentials file feels like a far edge case, but technically agree it would have worked in 3.17.3 so fair enough. I like your bool approach in https://github.com/helm/helm/pull/30932 better. Could have applied that here but you already have it so closing this PR in favor of yours 👍 nice job there

PR #30927, opened 2025-05-28
fix: downloader loadChartRepositories dynamic return

TerryHowe:
This causes occasional test failures and it currently blocking https://github.com/helm/helm/pull/30917

**If applicable**:


benoittgt:
I may have found the issue. It was not easy. ahah https://github.com/helm/helm/pull/30930

PR #30932, opened 2025-05-29
Handle an empty registry config file.

mattfarina:
In the move to oras v2, an existing but empty registry config file became an uncaught error. A missing file caused no error. This change catches the error and works around it so that Helm can continue to be fault tolerant to this issue.

**What this PR does / why we need it**: When Helm moved to oras v2, the handling of empty registry configuration files changed. It previously loaded default values and continued on. In oras v2, an error is produced.

This change restores the fault tolerant behavior so that the helm regression is fixed.

**If applicable**:

### DIFF
--- a/cmd/helm/registry_login_test.go
+++ b/cmd/helm/registry_login_test.go
@@ -17,9 +17,73 @@ limitations under the License.
 package main
 
 import (
+	"fmt"
+	"os"
+	"path/filepath"
 	"testing"
+
+	"helm.sh/helm/v3/pkg/repo/repotest"
 )
 
 func TestRegistryLoginFileCompletion(t *testing.T) {
 	checkFileCompletion(t, "registry login", false)
+}
+
+func TestRegistryLogin(t *testing.T) {
+	srv, err := repotest.NewTempServerWithCleanup(t, "testdata/testcharts/*.tgz*")
+	if err != nil {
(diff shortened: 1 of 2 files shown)


mattfarina:
Note, handling the empty file happens as part of error handling. It's done this way so that the happy path sees no slowdown or extra work. Only if there is an en error is extra work performed.

Also, the test flexes both opening the extra file and writing to it.

sabre1041 (approved):
LGTM

scottrigby (approved):
I agree this fixes one case that the approach in https://github.com/helm/helm/pull/30924 does not: the case where a user has an empty credentials file that is ALSO unwritable. While an edge case, it was technically possible with 3.17.3, so I agree we should fix it if we can.

I like this PR's solution to that additional edge case:
- setting a bool in [registry.NewClient] (instead of attempting a write there, since the write attempt in 3.17.3 only happens in [registry.Login])
- attempting the `{}` write in [registry.Login] only if the bool was set on the client

I do think since this fix is at the registry package level, the test should be there (eg in `pkg/registry/client_test.go` etc) rather than at the cmd level in this PR. But I agree we should get this fix into another v3 regression fix patch release ASAP, so don't want to hold it up and a command level test doesn't hurt. We can add the pkg/registry level test as a follow-up.

PR #30916, opened 2025-05-26
test: add functional tests for "Fix 3.18.0 regression: registry login with scheme"

jessesimpson36:
**What this PR does / why we need it**:

This PR introduces extra tests for filtering out the registry protocol during the helm registry login commands.
This should augment the recently merged https://github.com/helm/helm/pull/30887 change which fixes this issue https://github.com/helm/helm/issues/30873.

I'm targeting v3 with this change since that's where the regression was introduced. 

**Special notes for your reviewer**:

**If applicable**:
- [ ] this PR contains user facing changes (the `docs needed` label should be applied if so)
- [ ] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

PR #30936, opened 2025-05-29
fix: port registry login stripping to v4

jessesimpson36:

I made a v3 PR that introduced some integration tests that tested the functionality of stripping the url protocol prior to a `helm login` command. https://github.com/helm/helm/pull/30916 .  When I was figuring out whether the tests makes sense for v4, I noticed that the tests failed, and realized that the patch https://github.com/helm/helm/pull/30887 was not applied to v4 yet.

I'm still working out whether that is by-design or not, but for now I'm going to open up this PR pre-emptively until I have an answer.

**If applicable**:

### DIFF
--- a/pkg/registry/client.go
+++ b/pkg/registry/client.go
@@ -238,8 +238,39 @@ type (
 	}
 )
 
+// Deprecated: will be removed in Helm 4
+// Added for backwards compatibility for Helm < 3.18.0 after moving to ORAS v2
+// ref: https://github.com/helm/helm/issues/30873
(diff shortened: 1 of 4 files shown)


jessesimpson36:
In this github issue: https://github.com/helm/helm/issues/30873

it was consciously decided that the patch for the url stripping would not be ported to v4.  My mistake.
As for the tests that I added to augment this url stripping, those are also irrelevant as for v4 we will require more strict url parameters be used.

Closing this PR.

PR #30937, opened 2025-05-29
fix: legacy docker support broken for login

### LINKED ISSUE #30929: helm v3.18.0 registry login creates non-functional config for docker hub
On helm v3.18.0, after successfully logging in to a docker hub repository with `helm registry login`, OCI operations such as `helm pull` and `helm show all` do not work without a manual edit to the `config.json`.

The problem seems to be the auth key name in the json. On helm `v3.17.3` the config generated by `helm registry login` uses a different auth key than  `v3.18.0` (presumably due to the oras-go v2 upgrade).

In `v3.17.3`
```json
{
    "auths": {
        "https://index.docker.io/v1/": {
            "auth": "<redacted>"
        }
    }
}
```

In `v3.18.0`
```json
{
	"auths": {
		"registry-1.docker.io": {
			"auth": "<redacted>"
		}
	}
}
```

**Steps to repro:**
```bash
echo -n $DOCKERHUB_PAT | helm registry login registry-1.docker.io --username $DOCKERHUB_USERNAME --password-stdin
# Login Succeeded

helm show all oci://registry-1.docker.io/$ORG/$CHART
# Error: GET "https://registry-1.docker.io/v2/<ORG>/<CHART>/tags/list": response status code 401: unauthorized: authentication required: [map[Action:pull Class: Name:addgene/addgene-core Type:repository]]
```

**Reverting to the old auth key fixes it**

```bash
sed -i 's|registry-1.docker.io|https://index.docker.io/v1/|g' ~/.config/helm/registry/config.json

helm show all oci://registry-1.docker.io/$ORG/$CHART
# Successful response displayed
```

I have temporarily added a python workaround in our codebase (which more or less achieves the same thing as the `sed -i` command above).

```python
def _apply_registry_config_workaround():
    """
    This is a workaround for helm v3.18.0.

    The `helm registry login` command creates an auth entry for `registry-1.docker.io`
    but docker hub expects `https://index.docker.io/v1/`, so we copy it over.
    """
    config_path = Path.home() / ".config" / "helm" / "registry" / "config.json"

    # Load existing config
    with config_path.open() as f:
        config = json.load(f)

    auth_data = config["auths"].get("registry-1.docker.io")
    if auth_data:
        # Copy the auth from `registry-1.docker.io` to `https://index.docker.io/v1/`
        # which seems to be the legacy key for Docker Hub.
        config["auths"]["https://index.docker.io/v1/"] = auth_data

        # Write it back
        with config_path.open("w") as f:
            json.dump(config, f, indent=4)
```

TerryHowe:
When logging in with `registry-1.docker.io` it does not map it to `https://index.docker.io/v1/`

Fixes: https://github.com/helm/helm/issues/30929

**If applicable**:

### DIFF
--- a/pkg/registry/client.go
+++ b/pkg/registry/client.go
@@ -262,6 +262,7 @@ func (c *Client) Login(host string, options ...LoginOption) error {
 	}
 
 	key := credentials.ServerAddressFromRegistry(host)
+	key = credentials.ServerAddressFromHostname(key)
 	if err := c.credentialsStore.Put(ctx, key, cred); err != nil {
 		return err
 	}


mattfarina (approved):
I reviewed this one. It's intermittent to experience. I looked at the configuration as it's stored using `docker`, Helm 3.17, and Helm 3.18. There was a change in 3.18 when logging in to `registry-1.docker.io`. Previously, the credentials were stored at `index.docker.io`.

This PR restores the behavior as it was in Helm 3.17 when credentials are stored.

Note, `docker` is a bit strange in what it stores. It stores both `index.docker.io` and `registry-1.docker.io`.

This change works fix the storage behavior change.

PR #13375, opened 2024-10-05
Allow same deps with same name but different repo. closes #7413

Kerat:
Signed-off-by: Tarek Marcé <tarek.marce@free.fr>

**What this PR does / why we need it**:
This PR fix issue #7413 by do not using repository cache if there is two dependencies with same name but different repository URL

**Special notes for your reviewer**:

**If applicable**:
- [ ] this PR contains user facing changes (the `docs needed` label should be applied if so)
- [x] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

PR #30946, opened 2025-06-01
fix: resolve dependencies with same name from different repositories

### LINKED ISSUE #30875: Helm dependency resolution failure for same chart names but from different repositories
### Problem Description
When a Helm chart declares multiple dependencies with the same name but from different repositories, dependency resolution fails. For example:

```yaml
dependencies:
- name: common-chart
  alias: common-chart
  repository: '@repo1'
  version: 1.0.0
- name: common-chart
  alias: common-chart-alt
  repository: '@repo2'  
  version: 2.0.0
```

In this case, Helm incorrectly tries to resolve both dependencies using only one of the repositories, leading to errors like:

```
Error: can't get a valid version for 1 subchart(s): "common-chart" (repository "https://repo2-url", version "1.0.0"). Make sure a matching chart version exists in the repo, or change the version constraint in Chart.yaml
```

### Impact
This limitation prevents legitimate use cases such as:
- Using the same chart from different repositories with aliases
- Developing against both production and development versions of a dependency
- Using forks of the same chart for different environments

### Analyse
I think that the core problem lies in how Helm identifies and indexes dependencies during the resolution process. 
In the resolveRepoNames function, a map (reposMap) is created that maps dependency names to repository names. This map uses only the dependency name as a key (`reposMap[dd.Name] = repo.Name`).
When multiple dependencies have the same name but use different repositories, each subsequent dependency overwrites the previous one in this map. Later in the Resolve function, `repoName := repoNames[d.Name]` is used to look up the correct repository, but because there is only one entry per name, some dependencies get assigned to the wrong repository.

I think this explains why multiple dependencies with the same name cause problems: the last dependency with that name overwrites all previous ones in the map.

### Proposed Solution
Create composite keys that uniquely identify each dependency using name, repository, and index.
Also maintain backward compatibility by implementing fallback logic in the resolver that tries composite keys first, then falls back to simple name lookup.

rickbrouwer:

Fix dependency resolution for charts with same name from different repositories. This PR resolves the issue where Helm fails to handle multiple dependencies with the same chart name from different repositories, even when using aliases.

When a Helm chart declares multiple dependencies with the same name but from different repositories, dependency resolution fails. For example:

```yaml
dependencies:
- name: common-chart
  alias: common-chart
  repository: '@repo1'
  version: 1.0.0
- name: common-chart
  alias: common-chart-alt
  repository: '@repo2'  
  version: 2.0.0
```

In this case, Helm incorrectly tries to resolve both dependencies using only one of the repositories, leading to errors like:

```
Error: can't get a valid version for 1 subchart(s): "common-chart" (repository "https://repo2-url", version "1.0.0"). Make sure a matching chart version exists in the repo, or change the version constraint in Chart.yaml
```

More information in the issue.
This PR is an update to a [previous ](https://github.com/helm/helm/pull/30947)PR that was made from dev-v3
Fixes: #30875

**If applicable**:

### DIFF
--- a/internal/resolver/resolver.go
+++ b/internal/resolver/resolver.go
@@ -17,6 +17,8 @@ package resolver
 
 import (
 	"bytes"
+	"crypto"
+	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -50,6 +52,19 @@ func New(chartpath, cachepath string, registryClient *registry.Client) *Resolver
 		cachepath:      cachepath,
 		registryClient: registryClient,
 	}
+}
+
+// dependencyKey creates a unique key for a dependency that includes name, repository, and index
+// to handle cases where multiple dependencies have the same name but different repositories
+func dependencyKey(name, repository string, index int) string {
+	// Use a combination of name, repository hash, and index to ensure uniqueness
+	repoHash := ""
+	if repository != "" {
+		hasher := crypto.SHA256.New()
+		hasher.Write([]byte(repository))
+		repoHash = hex.EncodeToString(hasher.Sum(nil))[:8] // Use first 8 chars for brevity
+	}
+	return fmt.Sprintf("%s-%s-%d", name, repoHash, index)
 }
 
 // Resolve resolves dependencies and returns a lock file with the resolution.
@@ -107,7 +122,15 @@ func (r *Resolver) Resolve(reqs []*chart.Dependency, repoNames map[string]string
 			continue
 		}
 
-		repoName := repoNames[d.Name]
+		// Try to get repository name using composite key first, then fall back to simple name
+		key := dependencyKey(d.Name, d.Repository, i)
+		repoName := repoNames[key]
+		
+		// Fall back to simple name lookup for backward compatibility
+		if repoName == "" {
+			repoName = repoNames[d.Name]
+		}
+
 		// if the repository was not defined, but the dependency defines a repository url, bypass the cache
 		if repoName == "" && d.Repository != "" {
 			locked[i] = &chart.Dependency{
--- a/internal/resolver/resolver.go
+++ b/internal/resolver/resolver.go
@@ -17,7 +17,7 @@ package resolver
 
 import (
 	"bytes"
-	"crypto"
+	"crypto/sha256"
 	"encoding/hex"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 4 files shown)


rickbrouwer:
From @benoittgt in the other PR:

> If we merge this, I think this could also close this one PR #13375

benoittgt:
Tested and it is working as expected.

```yaml
apiVersion: v2
name: test-chart
description: Test
type: application
version: 0.1.0
appVersion: "1.16.0"

dependencies:
- name: grafana
  alias: grafana-bitnami
  repository: 'https://charts.bitnami.com/bitnami'
  version: '11.0.0'
- name: grafana
  alias: grafana-official
  repository: 'https://grafana.github.io/helm-charts'
  version: '8.5.2'
```

```sh
$ /bin/helm dependency update

...Successfully got an update from the "https://grafana.github.io/helm-charts" chart repository
...Successfully got an update from the "rancher" chart repository
Saving 2 charts
Downloading grafana from repo https://charts.bitnami.com/bitnami
Downloading grafana from repo https://grafana.github.io/helm-charts
Deleting outdated charts

$ helm dependency update  # 3.19

...Successfully got an update from the "rancher" chart repository
Error: can't get a valid version for 1 subchart(s): "grafana" (repository "https://grafana.github.io/helm-charts", version "8.5.2"). Make sure a matching chart version exists in the repo, or change the version constraint in Chart.yaml
```

internal/resolver/resolver.go
```diff
@@ -52,6 +54,19 @@ func New(chartpath, cachepath string, registryClient *registry.Client) *Resolver
...
 
+// DependencyKey creates a unique key for a dependency that includes name, repository, and index
+// to handle cases where multiple dependencies have the same name but different repositories
+func DependencyKey(name, repository string, index int) string {
+	// Use a combination of name, repository hash, and index to ensure uniqueness
+	repoHash := ""
+	if repository != "" {
+		hasher := sha256.New()
+		hasher.Write([]byte(repository))
+		repoHash = hex.EncodeToString(hasher.Sum(nil))[:8] // Use first 8 chars for brevity
```
benoittgt, line 65: I think 8 is good enough with the amount of chance and usage of the same name in `Chart.yaml`.

But I'm not sure we need the dependency index from the key. name+normalized-repository should be sufficient for map uniqueness. If the same name+repo appears twice, both entries map to the same repo, so overwriting is harmless at this mapping layer and removes index-coupling. What do you think ?

benoittgt:
Tiny comment but the overall looks good to me. Thanks @rickbrouwer

rickbrouwer:
> Tiny comment but the overall looks good to me. Thanks @rickbrouwer

Hi @benoittgt 

Thanks for the review! And you're right!

But before I make some changes, I see there's another PR (https://github.com/helm/helm/pull/30976) that also fixes the problem. I think that PR is better and cleaner than this one. What do you think? If you agree with me, feel free to close this PR so that the other PR can continue. If I'm missing something that makes this PR better, please let me know.

PR #30947, opened 2025-06-01
fix: resolve dependencies with same name from different repositories

rickbrouwer:

Fix dependency resolution for charts with same name from different repositories. This PR resolves the issue where Helm fails to handle multiple dependencies with the same chart name from different repositories, even when using aliases.

When a Helm chart declares multiple dependencies with the same name but from different repositories, dependency resolution fails. For example:

```yaml
dependencies:
- name: common-chart
  alias: common-chart
  repository: '@repo1'
  version: 1.0.0
- name: common-chart
  alias: common-chart-alt
  repository: '@repo2'  
  version: 2.0.0
```

In this case, Helm incorrectly tries to resolve both dependencies using only one of the repositories, leading to errors like:

```
Error: can't get a valid version for 1 subchart(s): "common-chart" (repository "https://repo2-url", version "1.0.0"). Make sure a matching chart version exists in the repo, or change the version constraint in Chart.yaml
```

More information in the issue.
Fixes: #30875

**If applicable**:

### DIFF
--- a/internal/resolver/resolver.go
+++ b/internal/resolver/resolver.go
@@ -17,6 +17,8 @@ package resolver
 
 import (
 	"bytes"
+	"crypto"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"os"
@@ -49,6 +51,19 @@ func New(chartpath, cachepath string, registryClient *registry.Client) *Resolver
 		cachepath:      cachepath,
 		registryClient: registryClient,
 	}
+}
+
+// dependencyKey creates a unique key for a dependency that includes name, repository, and index
+// to handle cases where multiple dependencies have the same name but different repositories
+func dependencyKey(name, repository string, index int) string {
+	// Use a combination of name, repository hash, and index to ensure uniqueness
+	repoHash := ""
+	if repository != "" {
+		hasher := crypto.SHA256.New()
+		hasher.Write([]byte(repository))
+		repoHash = hex.EncodeToString(hasher.Sum(nil))[:8] // Use first 8 chars for brevity
+	}
+	return fmt.Sprintf("%s-%s-%d", name, repoHash, index)
 }
 
 // Resolve resolves dependencies and returns a lock file with the resolution.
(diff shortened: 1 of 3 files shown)


benoittgt:
If we merge this, I think this could also close this one PR https://github.com/helm/helm/pull/13375

benoittgt:
@rickbrouwer, your PR should target the main branch. If necessary, it can be backported, but I believe it will likely only be included in Helm v4.

pkg/downloader/manager.go
```diff
@@ -641,6 +648,19 @@ repository, use "https://charts.example.com/" or "@example" instead of
 	return reposMap, nil
 }
 
+// dependencyKey creates a unique key for a dependency that includes name, repository, and index
```
benoittgt, line 651: Should we have this in only one location ?

internal/resolver/resolver.go
```diff
@@ -51,6 +53,19 @@ func New(chartpath, cachepath string, registryClient *registry.Client) *Resolver
 	}
 }
 
+// dependencyKey creates a unique key for a dependency that includes name, repository, and index
+// to handle cases where multiple dependencies have the same name but different repositories
+func dependencyKey(name, repository string, index int) string {
+	// Use a combination of name, repository hash, and index to ensure uniqueness
+	repoHash := ""
+	if repository != "" {
+		hasher := crypto.SHA256.New()
```
benoittgt, line 62: What about importing `crypto/sha256` and use directly `sha256.New()` ?

pkg/downloader/manager.go
rickbrouwer: Yes, absolutely true! I made an update from the main branch https://github.com/helm/helm/pull/30946.
I will close this PR.

internal/resolver/resolver.go
rickbrouwer: Yes! Great feedback, thanks! I made an update from the main branch https://github.com/helm/helm/pull/30946.
I will close this PR.

PR #30950, opened 2025-06-02
fix: Revert Helm v3 to OAuth2 workflow for OCI

scottrigby:

ORAS v1 used POST for auth as part of OAuth workflow
ORAS v2 changed to GET for this

While this does not affect most users, the change has caused a regression in at least one custom registry that relied on checking the POST auth request. 

This one-line PR reverts the OCI OAuth workflow behavior back how 3.17.3 worked.

We can use the ForceAttemptOAuth2 setting to revert to prior behavior

This was not well-documented in the docker implementation used by ORAS 1

This is well-documented in the ORAS 2 OCI auth implementation:
https://github.com/opencontainers/wg-auth/blob/main/docs/implementations/ORAS.md?plain=1

Unit tests: this PR does not add any additional unit tests because the auth workflow is tested extensively in ORAS 2 unit tests, and all of our tests in Helm apply equally to them. This PR however has been tested extensively for backwards compatibility, fixing one registry edge case while all other behavior remains the same. See:
- https://pkg.go.dev/oras.land/oras-go/v2/registry/remote/auth@v2.6.0#Client.ForceAttemptOAuth2
- https://github.com/oras-project/oras-go/blob/v2.6.0/registry/remote/auth/client_test.go

See 

**If applicable**:

### DIFF
--- a/pkg/registry/client.go
+++ b/pkg/registry/client.go
@@ -156,6 +156,8 @@ func NewClient(options ...ClientOption) (*Client, error) {
 			authorizer.Cache = auth.NewCache()
 		}
 
+		authorizer.ForceAttemptOAuth2 = true
+
 		client.authorizer = &authorizer
 	}
 


benoittgt (approved):
Should we add a comment to explain this line, even though it’ll be dropped soon?

sabre1041:
@scottrigby Verification of the enhancement did not prove successful 

```
./bin/helm registry login xx.com                
Username: xxxxxx
Password: 
Error: authenticating to "xx.com": GET "https://xx.com/v2/": POST "https://xx.com/v2/auth": response status code 405: Method Not Allowed
````

scottrigby:
> @scottrigby Verification of the enhancement did not prove successful
> Error: authenticating to "xx.com": GET "https://xx.com/v2/": POST "https://xx.com/v2/auth": response status code 405: Method Not Allowed

This works with Dockerhub and some other registries I've tested. But your'e right—I can confirm this fails with both quay and google artifact registry with the same error.

The direction of this PR is sound—many of the successful tests in https://github.com/oras-project/oras-go/blob/v2.6.0/registry/remote/auth/client_test.go set `ForceAttemptOAuth2` successfully. Moving this PR to draft while we investigate.

mattfarina:
@scottrigby Can you verify with ghcr.io & Quay in addition to Docker Hub. This will flex against some different implementations.

scottrigby:
> @scottrigby Can you verify with ghcr.io & Quay in addition to Docker Hub. This will flex against some different implementations.

I will definitely do before moving this PR back out of draft.

As an aside, we should really set up accounts for Helm testing in all the major OCI registries—I realize this is known, but noting this one detail here since it's relevant.

mattfarina:
@scottrigby we don't do outside integration testing. All the testing is self contained so that anyone can run it in a development environment without needing credentials. Previously, some work had been done at https://github.com/helm/acceptance-testing. I think integration testing would be great but it needs to be carefully.

mattfarina:
Note, the underlying issue here appears to be another regression from the ORAS v1 -> v2 transition.

scottrigby:
> @scottrigby we don't do outside integration testing. All the testing is self contained so that anyone can run it in a development environment without needing credentials. Previously, some work had been done at https://github.com/helm/acceptance-testing. I think integration testing would be great but it needs to be carefully.

Yes, I meant integration not unit testing with this note 👍

scottrigby:
Closing this PR but will open an issue to track per last dev meeting

PR #30966, opened 2025-06-07
Display OCI annotations in 'helm show chart'

### LINKED ISSUE #30787: Display oci annotations or created date if annotation exists in helm show chart command
It would be cool either with a new sub command under `helm show` or in `helm show chart` that displays all OCI annotations and/or displays standard ones like the created timestamp as a special field.

If this functionality already exists under a different command please let me know. Otherwise, here is an example:
```
oras manifest fetch helm.foo.com/bar-app:3.45.6 | jq '.annotations."org.opencontainers.image.created"'
"2025-04-11T20:12:25Z"
```

Victorthedev:
When showing chart information via `helm show chart`, the command now
includes any OCI-specific annotations (prefixed with 'org.opencontainers.image.')
that exist in the chart metadata.

Fixes #30787

**If applicable**:


pkg/action/show.go
```diff
@@ -95,6 +95,17 @@ func (s *Show) Run(chartpath string) (string, error) {
...
+		// Display OCI annotations if they exist
+		if s.chart.Metadata != nil && len(s.chart.Metadata.Annotations) > 0 {
+			fmt.Fprintln(&out, "---")
+			fmt.Fprintln(&out, "Annotations:")
+			for k, v := range s.chart.Metadata.Annotations {
+				if strings.HasPrefix(k, "org.opencontainers.image.") {
+					fmt.Fprintf(&out, "  %s: %s\n", k, v)
+				}
+			}
+		}
```
TerryHowe, line 115: I need to look back at the issue, but I'd expect this wouldn't be the default behavior and some flag would enable this.

TerryHowe (commented):
An output same would be nice.

pkg/action/show.go
Victorthedev: I've made it optional via a flag

benoittgt: But it's cli flags, it seems we are missing some code here.

Aamir017:
Hi, I’d like to work on this issue if it’s available. Could you please assign it to me?

benoittgt:
@Aamir017 feel free to take it as it turns stale.

PR #30972, opened 2025-06-10
fix: kube client create mutex

TerryHowe:
This code will core with a data race every once in a while.

**If applicable**:


mattfarina (approved):
lgtm. I did some performance testing and was not able to see any issues.

Issue #12977, opened 2024-04-23
--dependency-update is not working as expected


JoniJnm:
`helm upgrade --dependency-update` is not updating the downloaded chart tgz files when the version doesn't match with the chart.lock

I created a repository to reproduce the bug: https://github.com/JoniJnm/helm-bug

* The main branch needs the dependency `mysql:9.12.3`
* The v2 branch needs the dependency `mysql:9.16.3`

The steps:

```shell
$ cd /tmp
$ git clone https://github.com/JoniJnm/helm-bug
$ ls
Chart.lock  Chart.yaml  README.md
$ helm upgrade --install --dependency-update helm-bug .
Release "helm-bug" does not exist. Installing it now.
Hang tight while we grab the latest from your chart repositories...
...Successfully got an update from the "bitnami" chart repository
Update Complete. ⎈Happy Helming!⎈
Saving 1 charts
Downloading mysql from repo https://charts.bitnami.com/bitnami
Deleting outdated charts
NAME: helm-bug
LAST DEPLOYED: Tue Apr 23 18:42:47 2024
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
```

This downloads the mysql-9.12.3.tgz file:
```shell
$ ls charts/
mysql-9.12.3.tgz
```

But know, when the chart lock changes, it doesn't update the chart:
```shell
$ git switch v2
$ helm upgrade --install --dependency-update helm-bug .
Release "helm-bug" has been upgraded. Happy Helming!
NAME: helm-bug
LAST DEPLOYED: Tue Apr 23 18:45:15 2024
NAMESPACE: default
STATUS: deployed
REVISION: 2
TEST SUITE: None
```
The chart is not updated:

```shell
$ cat Chart.lock | grep version
  version: 9.16.3
$ ls charts/
mysql-9.12.3.tgz
```

```
$ helm version
version.BuildInfo{Version:"v3.14.4", GitCommit:"81c902a123462fd4052bc5e9aa9c513c4c8fc142", GitTreeState:"clean", GoVersion:"go1.21.9"}
```

PR #30974, opened 2025-06-11
fix: raised error from downloadAll in case of dependency conflict

### LINKED ISSUE #30710: Helm allows conflicting subcharts and dependencies without warning
Given a helm chart with a subchart in the `charts` directory, say `nginx`, and a conflicting dependency is added to `Chart.yaml` Helm does not detect this when downloading dependencies. Instead, a duplicate chart is downloaded.

**Chart.yaml**
```yaml
apiVersion: v2
name: foo
type: application
version: 0.1.0
appVersion: "1.16.0"
dependencies:
  - name: nginx
    version: 19.0.3
    repository: oci://registry-1.docker.io/bitnamicharts
```

**Contents of the chart**
```
.
├── Chart.lock
├── Chart.yaml
├── charts
│   ├── nginx
│   │   ├── Chart.lock
│   │   ├── Chart.yaml
...
├── templates
│   ├── NOTES.txt
...
└── values.yaml
```

**The problem**
`charts/nginx` chart was downloaded using `helm pull oci://registry-1.docker.io/bitnamicharts/nginx --untar -d charts/`. Running `helm dependency update` will download nginx dependency to `charts/nginx-19.0.3.tgz`. Helm will template seems to prefer the `charts/nginx` subchart. This behaviour can be confusing because it would not be clear which dependency would be preferred. 

**Proposal**
Helm ought to better detect that a duplicate dependency in `Chart.yaml` conflicts with another subchart in `charts` directory and error accordingly with something like below

```sh
Error: Dependency conflict detected: A subchart named 'nginx' already exists in charts/ directory.  
Please remove the manually added subchart or exclude it from Chart.yaml dependencies.
```

manslaughter03:
fix #30710

Check if subchart already exist before downloading it, then in case of conflict raised an error.

I'm currently raising the error directly from downloadAll, wouldn't it be more appropriate to simply log a warning instead? In order to continue to iterate on deps to download ?

**If applicable**:

### DIFF
--- a/pkg/downloader/manager.go
+++ b/pkg/downloader/manager.go
@@ -271,11 +271,11 @@ func (m *Manager) downloadAll(deps []*chart.Dependency) error {
 	var saveError error
 	churls := make(map[string]struct{})
 	for _, dep := range deps {
+		chartPath := filepath.Join(destPath, dep.Name)
 		// No repository means the chart is in charts directory
 		if dep.Repository == "" {
 			fmt.Fprintf(m.Out, "Dependency %s did not declare a repository. Assuming it exists in the charts directory\n", dep.Name)
 			// NOTE: we are only validating the local dependency conforms to the constraints. No copying to tmpPath is necessary.
-			chartPath := filepath.Join(destPath, dep.Name)
 			ch, err := loader.LoadDir(chartPath)
 			if err != nil {
 				return fmt.Errorf("unable to load chart '%s': %v", chartPath, err)
@@ -297,6 +297,13 @@ func (m *Manager) downloadAll(deps []*chart.Dependency) error {
 			}
 			continue
 		}
+
+		// Check if subchart already exist before downloading tarball.
+		// https://github.com/helm/helm/issues/30710
+		if _, err := os.Stat(chartPath); !os.IsNotExist(err) {
+			return fmt.Errorf("dependency conflict detected: A subchart named '%s' already exists in charts/ directory", dep.Name)
+		}
+
 		if strings.HasPrefix(dep.Repository, "file://") {
 			if m.Debug {
 				fmt.Fprintf(m.Out, "Archiving %s from repo %s\n", dep.Name, dep.Repository)
--- a/pkg/downloader/manager.go
+++ b/pkg/downloader/manager.go
@@ -162,6 +162,13 @@ func (m *Manager) Update() error {
 		return nil
 	}
 
+	// Check if there is already subchart in charts directory before downloading tarball.
+	// https://github.com/helm/helm/issues/30710
+	err = m.checkSubchartConflict(req)
+	if err != nil {
+		return fmt.Errorf("%s. Please remove the manually added subchart or exclude it from Chart.yaml dependencies", err.Error())
+	}
+
 	// Get the names of the repositories the dependencies need that Helm is
 	// configured to know about.
 	repoNames, err := m.resolveRepoNames(req)
@@ -296,12 +303,6 @@ func (m *Manager) downloadAll(deps []*chart.Dependency) error {
 				break
 			}
 			continue
-		}
-
-		// Check if subchart already exist before downloading tarball.
-		// https://github.com/helm/helm/issues/30710
-		if _, err := os.Stat(chartPath); !os.IsNotExist(err) {
-			return fmt.Errorf("dependency conflict detected: A subchart named '%s' already exists in charts/ directory", dep.Name)
 		}
 
 		if strings.HasPrefix(dep.Repository, "file://") {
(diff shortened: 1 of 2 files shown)


benoittgt:
Hello @manslaughter03 

I just tested your patch

```yaml
apiVersion: v2
name: foo
type: application
version: 0.1.0
appVersion: "1.16.0"
dependencies:
  - name: nginx
    version: 20.0.7
    repository: oci://registry-1.docker.io/bitnamicharts
```

First
```sh
helm pull oci://registry-1.docker.io/bitnamicharts/nginx --version '20.0.7' --untar -d charts/
```
Then
```sh
helm dependency update

Saving 1 charts
Error: dependency conflict detected: A subchart named 'nginx' already exists in charts/ directory
```

Should this error happen earlier?
Not sure how feasible that is technically, but it might improve UX.
Also, maybe we could provide a clearer suggestion about what action should be taken in this case?

### Other behavior that look a little bit surprising

Probably unreleated
<details>
 <summary>Unexpected behavior with <code>--dependency-update</code></summary>

```sh
$ rm -rf charts/nginx* Chart.lock
$ cat Chart.yaml | grep -A2 nginx
  - name: nginx
    version: 20.0.6
    repository: oci://registry-1.docker.io/bitnamicharts
$ helm upgrade --install test-release . --dependency-update

Release "test-release" has been upgraded. Happy Helming!
NAME: test-release
LAST DEPLOYED: Wed Jun 11 22:48:03 2025
NAMESPACE: default
STATUS: deployed
REVISION: 5
DESCRIPTION: Upgrade complete
TEST SUITE: None
$ ls -la charts
total 104
drwxr-xr-x  3 benoit.tigeot     96 Jun 11 22:46 .
drwxr-xr-x  7 benoit.tigeot   224 Jun 11 22:47 ..
-rw-r--r--  1 benoit.tigeot  52940 Jun 11 22:46 nginx-20.0.5.tgz # dependency not updated.
```

I think it could be related to https://github.com/helm/helm/issues/12977
</details>

manslaughter03:
Hi @benoittgt 

Thanks for your feedback.

* We might check potential conflict earlier, it can be done in `manager.Update` after loading chart directory, what do you think ?
* I can add a suggestion like it was propose in the related issue #30710: `remove the manually added subchart or exclude it from Chart.yaml dependencies.`

Concerning the other surprising behavior, i can reproduce #12977 but not when i clean `Chart.lock` and `charts/nginx*` before running upgrade with dependency update option.

It seems that `CheckDependencies` function called by upgrade/install action only check dependency name and not constraint version: https://github.com/helm/helm/blob/0a8194c9a3cdb33e4907d80088aa791147af994c/pkg/action/install.go#L736.

In addition i can't fetch nginx version `20.0.6`:

```bash
$ helm pull oci://registry-1.docker.io/bitnamicharts/nginx --version 20.0.6 --destination /tmp
Error: registry-1.docker.io/bitnamicharts/nginx:20.0.6: not found
```

Can you check with another nginx version ?

benoittgt:
> * We might check potential conflict earlier, it can be done in `manager.Update` after loading chart directory, what do you think ?

I think that could be a better approach, but I might be wrong.  
@‌gjenkins8 — maybe you’d like to share your thoughts on this?

> * I can add a suggestion like it was propose in the related issue [Helm allows conflicting subcharts and dependencies without warning #30710](https://github.com/helm/helm/issues/30710): `remove the manually added subchart or exclude it from Chart.yaml dependencies.`

Yes, good idea! 😉

As for the other issue — it seems unrelated to your patch, so I think we should continue that discussion in a separate issue or PR.

manslaughter03:
@benoittgt i made some changes:
* move conflict check logic into a specific function and call it in `manager.Update`
* Update error message raised to provide a clearer suggestion to the user.

benoittgt:
Thanks @manslaughter03. I think it's much better like that. Maybe the message could be on two lines but it's ok. I think we need a review from someone else. Maybe @TerryHowe ?

Current output:
```sh
» helm pull oci://registry-1.docker.io/bitnamicharts/nginx --version '20.0.7' --untar -d charts/

Pulled: registry-1.docker.io/bitnamicharts/nginx:20.0.7
Digest: sha256:1449692e778bf269197cd98f73c19ebd2172c756e0f0459a5e0076fa95497528

» helm dependency update

Error: dependency conflict detected: A subchart named 'nginx' already exists in charts/ directory. Please remove the manually added subchart or exclude it from Chart.yaml dependencies
```

pkg/downloader/manager.go
```diff
@@ -271,11 +278,11 @@ func (m *Manager) downloadAll(deps []*chart.Dependency) error {
 	var saveError error
 	churls := make(map[string]struct{})
 	for _, dep := range deps {
+		chartPath := filepath.Join(destPath, dep.Name)
```
TerryHowe, line 281: This change doesn't seem to add any value anymore.

pkg/downloader/manager.go
```diff
@@ -840,6 +848,22 @@ func (m *Manager) loadChartRepositories() (map[string]*repo.ChartRepository, err
...
+// checkSubchartConflict Check if subchart already exist before downloading tarball.
+func (m *Manager) checkSubchartConflict(req []*chart.Dependency) error {
+	for _, dep := range req {
+		// No repository means the chart is in charts directory
+		if dep.Repository == "" {
+			continue
+		}
+		chartPath := filepath.Join(m.ChartPath, "charts", dep.Name)
+		if _, err := os.Stat(chartPath); !os.IsNotExist(err) {
+			return fmt.Errorf("dependency conflict detected: A subchart named '%s' already exists in charts/ directory", dep.Name)
```
TerryHowe, line 860: Maybe generate a slice of strings here and report them all at the end if there is anything on the slice.

TerryHowe (changes requested):

TerryHowe:
Reopen if there are updates

PR #30812, opened 2025-04-28
HIP-0023: Helm support server-side apply

gjenkins8:
**What this PR does / why we need it**:
Implement Helm client and SDK changes to support Server Side Apply.

Notably, this PR does:
1. Creates the `--service-side` and `--force-conflicts` flags (and corresponding API objects), and passes these to Kubernetes client
2. Creates an `ApplyMethod` field in release, which tracks the apply method (client or server side) used to deploy the chart
    - Upon upgrade, the next release will default to that apply method (unless override by e.g. CLI flag)
    - Prints this field in `helm get metadata`
3. When upgrade a previous chart release that utilized client-side apply, Kubernetes field management objects are converted/upgrade to server-side management.

**Special notes for your reviewer**:
Specification documented in: [HIP-0023](https://github.com/helm/community/blob/main/hips/hip-0023.md)

The Kubernetes client changes were done in: https://github.com/helm/helm/pull/31030

**If applicable**:
- [x] this PR contains user facing changes (the `docs needed` label should be applied if so)
- [x] this PR contains unit tests
- [x] this PR has been tested for backwards compatibility

PR #30982, opened 2025-06-15
Rename 'force' to 'force-replace'

gjenkins8:
https://github.com/helm/helm/pull/30812 introduces a `--force-conflicts` flag (and corresponding `ForceConflict` SDK variable).

In order to avoid user confusion with the current `--force` flag / `Force` SDK variable, this PR renames these to `--force-replace` / `ForceReplace` to accurately describe the functionality (`--force` today causes Helm to use the Kubernetes 'replace' API (only))

Note: in https://github.com/helm/helm/pull/30812, it is not possible for `ForceReplace` and `ForceConflicts` to be used simultaneously. There must be separate flags to describe which "force" behavior the user desires.

Note: I also tidied up the naming of `forceOwnership` in `setMetadataVisitor`

The `--force` flag still exists (and behaves the same), but is marked deprecated, so will print a warning if used.

**If applicable**:

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -71,8 +71,11 @@ type Install struct {
 
 	ChartPathOptions
 
-	ClientOnly      bool
-	Force           bool
+	ClientOnly bool
+	// ForceReplace will, if set to `true`, ignore certain warnings and perform the install anyway.
+	//
+	// This should be used with caution.
+	ForceReplace    bool
 	CreateNamespace bool
 	DryRun          bool
 	DryRunOption    string
@@ -346,7 +349,7 @@ func (i *Install) RunWithContext(ctx context.Context, chrt *chart.Chart, vals ma
 		return nil, fmt.Errorf("unable to build kubernetes objects from release manifest: %w", err)
 	}
 
-	// It is safe to use "force" here because these are resources currently rendered by the chart.
+	// It is safe to use "forceOwnership" here because these are resources currently rendered by the chart.
 	err = resources.Visit(setMetadataVisitor(rel.Name, rel.Namespace, true))
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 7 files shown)


pkg/action/install.go
```diff
@@ -71,8 +71,10 @@ type Install struct {
 
 	ChartPathOptions
 
-	ClientOnly      bool
-	Force           bool
+	ClientOnly   bool
+	ForceReplace bool
+	// ForceConflicts causes server-side apply to force conflicts ("Overwrite value, become sole manager")
+	// see: https://kubernetes.io/docs/reference/using-api/server-side-apply/#conflicts
 	CreateNamespace bool
```
TerryHowe: Are you adding `ForceConflicts` in a follow-up PR?

TerryHowe (changes requested):

mattfarina (commented):
Why do we want to change this? Who does it benefit?

`--force` is the flag used in _kubectl_. `--force-conflicts` is used for the conflicts option.

I'm tempted to suggest we keep things they way they are which pairs with the flags from kubectl. For existing users and cloud native users that would be easier, right?

pkg/action/install.go
gjenkins8: oops, I split this from that PR. fixing

pkg/action/install.go
```diff
@@ -71,8 +71,11 @@ type Install struct {
 
 	ChartPathOptions
 
-	ClientOnly      bool
-	Force           bool
+	ClientOnly bool
```
TerryHowe, line 74:
```suggestion
	ClientOnly      bool
```

TerryHowe (commented):
looks good with nit

pkg/action/install.go
gjenkins8: `go fmt` would disagree :)

TerryHowe: Weird, but okay

TerryHowe (approved):

joejulian:
I like this change. I've always thought that "force" was ambiguous. This makes it much more explicit.

joejulian (approved):

godhanipayal (approved):

PR #30988, opened 2025-06-18
Use https://github.com/otiai10/copy/ for copying directories

zachburg:

This is the first step for removing the partial copy of https://github.com/golang/dep in
https://github.com/helm/helm/blob/main/internal/third_party/dep/fs/fs.go, which is a deprecated and archived repo.

**If applicable**:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,6 +25,7 @@ require (
 	github.com/mitchellh/copystructure v1.2.0
 	github.com/moby/term v0.5.2
 	github.com/opencontainers/image-spec v1.1.1
+	github.com/otiai10/copy v1.14.0
 	github.com/phayes/freeport v0.0.0-20220201140144-74d24b5ae9f5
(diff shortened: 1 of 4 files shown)


robertsirc:
So, I am curious on why this is needed?

benoittgt:
I think it's because we're using [this file](https://github.com/helm/helm/blob/main/internal/third_party/dep/fs/fs.go), which is a copy of a deprecated and archived project: [golang/dep's fs.go](https://github.com/golang/dep/blob/master/internal/fs/fs.go).  

However, I'm wondering — why add a new external dependency? Are we facing any issues with the currently copied `fs.go`? @zachburg  

I tried to find issues mentioning this "third party", but didn’t come across any.

joejulian:
If we were to replace this, I would prefer we do it with a package that has more than one maintainer and a consistent commit cadence and release cycle.

PR #30995, opened 2025-06-20
Added Env for max decompressed chart size and chart file size limits

### LINKED ISSUE #30738: MaxDecompressedFileSize in v3.17.3 breaks existing charts

eunanio:
closes #30738

### Summary
Allows the user to set the max decompressed chart and file size limits introduced in v3.17.3 via environment variables, This can be used to resolve existing charts exceeding the size limits raised in #30738 

### New Environment Variables 
`HELM_MAX_DECOMPRESSED_CHART_SIZE` - Default: 100 MiB
`HELM_MAX_DECOMPRESSED_FILE_SIZE` - Default: 5 MiB

### DIFF
--- a/pkg/chart/v2/loader/archive.go
+++ b/pkg/chart/v2/loader/archive.go
@@ -27,10 +27,14 @@ import (
 	"os"
 	"path"
 	"regexp"
+	"strconv"
 	"strings"
 
 	chart "helm.sh/helm/v4/pkg/chart/v2"
 )
+
+const MaxChartFileSizeEnv = "HELM_MAX_DECOMPRESSED_FILE_SIZE"
+const MaxChartSizeEnv = "HELM_MAX_DECOMPRESSED_CHART_SIZE"
 
 // MaxDecompressedChartSize is the maximum size of a chart archive that will be
 // decompressed. This is the decompressed size of all the files.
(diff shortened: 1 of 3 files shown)


pkg/chart/v2/loader/archive.go
```diff
@@ -27,11 +27,15 @@ import (
...
 	"regexp"
+	"strconv"
 	"strings"
 
 	chart "helm.sh/helm/v4/pkg/chart/v2"
 )
 
+const MaxChartFileSizeEnv = "HELM_MAX_DECOMPRESSED_FILE_SIZE"
+const MaxChartSizeEnv = "HELM_MAX_DECOMPRESSED_CHART_SIZE"
+
```
TerryHowe, line 38: Why are these duplicated from pkg/cli/environment.go ?

eunanio: Good point I will keep everything defined within `archive.go` for simplicity and remove the duplication form `environment.go`

mattfarina (changes requested):
Thanks for your interest in this.

Think about helm this way, helm has an SDK that numerous applications use and a CLI/client. Much of that is in pkg/cmd. What users input needs to come through the client and the way we typically do things there.

benoittgt:
I made another proposal here https://github.com/helm/helm/pull/30743 that follows @mattfarina recommendations.

PR #31011, opened 2025-06-27
feature: add labels to metadata

### LINKED ISSUE #31008: Return release labels in the GetMetadata() function

Release object has the [Labels field](https://github.com/helm/helm/blob/main/pkg/release/v1/release.go#L44), which is saved as part of metadata of the release.
But [metadata object](https://github.com/helm/helm/blob/main/pkg/action/get_metadata.go#L36) does not have this field, so doesn't pass it on [fetch](https://github.com/helm/helm/blob/main/pkg/action/get_metadata.go#L67) action.

Could wee add Labels to the Metadata struct and return them?

### Why is this needed?

If we store something in metadata, it would be logical to have a method to get this info without fetching the whole release object. It could be very useful in some cases, like storing the checksum of the manifests in a label.

yalosev:
Closes #31008

Release object has the [Labels field](https://github.com/helm/helm/blob/main/pkg/release/v1/release.go#L44), which is saved as part of metadata of the release.
But fetching this metadata doesn't return stored labels.

**If applicable**:


pkg/action/get_metadata.go
```diff
@@ -34,11 +34,14 @@ type GetMetadata struct {
...
-	Chart        string              `json:"chart" yaml:"chart"`
-	Version      string              `json:"version" yaml:"version"`
-	AppVersion   string              `json:"appVersion" yaml:"appVersion"`
-	Annotations  map[string]string   `json:"annotations,omitempty" yaml:"annotations,omitempty"`
+	Name       string `json:"name" yaml:"name"`
+	Chart      string `json:"chart" yaml:"chart"`
+	Version    string `json:"version" yaml:"version"`
+	AppVersion string `json:"appVersion" yaml:"appVersion"`
+	// Annotations are fetched from the Chart.yaml file
+	Annotations map[string]string `json:"annotations,omitempty" yaml:"annotations,omitempty"`
```
TerryHowe, line 42: These lines probably shouldn't change

yalosev: I rolled back the changes, they were caused by `go fmt`

yalosev: Only left comments for annotations. Because `Labels` for the metadata are taken from the secret (release) labels but `Annotations` are taken from the `Chart.yaml` file.

benoittgt:
Just tested and it's working well.

LGTM :)

robertsirc:
Hi @yalosev, thank you for PR looks like some of the checks need attention. If you could please clear these up and we can review your PR.

pkg/action/get_metadata.go
yalosev: Changed again, because otherwise it couldn't pass the golangci-lint

yalosev:
Hi @robertsirc, sorry, was confused by [this](https://github.com/helm/helm/pull/31011#discussion_r2172463276) comment. Fixed

robertsirc (approved):
LGTM

joejulian (approved):
LGTM

pkg/action/get_metadata.go
TerryHowe: Can't fight lint/go fmt

TerryHowe (approved):
/lgtm

Issue #30892, opened 2025-05-21
Run Helm unit-tests with `-shuffle=on` and `-count=n` (n > 1)


gjenkins8:
`go test` by default runs unit tests once in a deterministic order.  Leading to:
1. deterministic order: tests relying on their order of execution to pass
2. running once: higher probability of flakey tests passing

Adding `-shuffle=on` and `-count=n` (for `n > 1`; perhaps 2 or 3, TBD) will improve these behaviors:

> -count n
	    Run each test, benchmark, and fuzz seed n times (default 1).
	    If -cpu is set, run n times for each GOMAXPROCS value.
	    Examples are always run once. -count does not apply to
	    fuzz tests matched by -fuzz.

> -shuffle off,on,N
	    Randomize the execution order of tests and benchmarks.
	    It is off by default. If -shuffle is set to on, then it will seed
	    the randomizer using the system clock. If -shuffle is set to an
	    integer N, then N will be used as the seed value. In both cases,
	    the seed will be reported for reproducibility.

https://pkg.go.dev/cmd/go/internal/test

7h3-3mp7y-m4n:
can I work on this?

benoittgt:
@7h3-3mp7y-m4n if you want I started working on this. But feel free to continue, it's still flaky. :) 

https://github.com/helm/helm/compare/main...benoittgt:helm:shuffle-test?expand=1

7h3-3mp7y-m4n:
It's alright @benoittgt , it's all yours ;)
All the best :)

benoittgt:
Careful if you work on this. It appears it breaks the Mac OS keychain. I had 3 settings wipe since yesterday, and it appears that the issue is happening to a test close to `TestTLSRegistryClientTestSuite/Test_4_Logout`. It calls [this in oras](https://github.com/oras-project/oras-go/blob/05a2b09cbf2eab1df691411884dc4df741ec56ab/registry/remote/credentials/registry.go#L63-L65) behind. I've lost some data settings in the battle.

robertsirc:
For anyone who has a PR for this please take me and gjenkins8 we will be happy to review.

adharsh277:
can i work on this??

mattfarina:
@gjenkins8 Moving to `-count=2` adds about 2.5 minutes to the test run time in CI. It adds a bunch of time to the local test runs, too. This impacts the inner loop on development. What do we gain by doing this? It is worth it?

[...]

PR #30952, opened 2025-06-03
fix(tests): add -shuffle=on -count=2 to test-unit in Makefile, Updated .golangci.yml to support test stability improvements.

Victorthedev:
reference Issue #30892 and PR #30951

**What this PR does / why we need it**:

**Special notes for your reviewer**:

**If applicable**:
- [ ] this PR contains user facing changes (the `docs needed` label should be applied if so)
- [ ] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

PR #31013, opened 2025-06-27
`-shuffle=on` and `-count=2` option is added to TESTFLAGS

AlfonsoUceda:
This PR addresses https://github.com/helm/helm/issues/30892

**If applicable**:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ ACCEPTANCE_RUN_TESTS=.
 PKG         := ./...
 TAGS        :=
 TESTS       := .
-TESTFLAGS   :=
+TESTFLAGS   := -shuffle=on -count=2
 LDFLAGS     := -w -s
 GOFLAGS     :=
 CGO_ENABLED ?= 0
@@ -119,7 +119,7 @@ test-unit:
 test-coverage:
 	@echo
 	@echo "==> Running unit tests with coverage <=="
-	@ ./scripts/coverage.sh
+	@ TESTFLAGS="${TESTFLAGS}" ./scripts/coverage.sh
 
 .PHONY: test-style
 test-style:
--- a/pkg/lint/support/message_test.go
+++ b/pkg/lint/support/message_test.go
@@ -21,7 +21,6 @@ import (
 	"testing"
 )
 
-var linter = Linter{}
 var errLint = errors.New("lint failed")
 
 func TestRunLinterRule(t *testing.T) {
@@ -45,6 +44,7 @@ func TestRunLinterRule(t *testing.T) {
 		{-1, errLint, 4, false, ErrorSev},
 	}
 
+	var linter = Linter{}
 	for _, test := range tests {
 		isValid := linter.RunLinterRule(test.Severity, "chart", test.LintError)
 		if len(linter.Messages) != test.ExpectedMessages {
--- a/scripts/coverage.sh
+++ b/scripts/coverage.sh
@@ -16,6 +16,8 @@
 
 set -euo pipefail
 
+TESTFLAGS=${TESTFLAGS:-}
+
 covermode=${COVERMODE:-atomic}
 coverdir=$(mktemp -d /tmp/coverage.XXXXXXXXXX)
 profile="${coverdir}/cover.out"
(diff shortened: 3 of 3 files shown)


AlfonsoUceda:
@gjenkins8 @robertsirc I've created this tiny PR to address that issue, I hope everything is ok and according to the requirements of that issue.

AlfonsoUceda:
I saw this PR https://github.com/helm/helm/pull/30952 but after 3 weeks without any change, I considered to do this PR.

robertsirc:
Hi @AlfonsoUceda, looks like `build-test` failed could you take a look at it?

AlfonsoUceda:
@robertsirc it seems `TestRunLinterRule` was a flaky spec

robertsirc:
Hi @AlfonsoUceda it looks like the linter is breaking as well, it should be an easy fix.

AlfonsoUceda:
@robertsirc will take care of it.

have a question, should I reduce count to 2? The flaky spec I fixed was failing with count 2, not sure if worth running all tests 3 times

robertsirc:
I think reducing it to 2 is fine. I am aware of flaky test from other package maintainers I am going to open an issue for them.

AlfonsoUceda:
@robertsirc should be fine now ;)

robertsirc (approved):
LGTM

robertsirc:
This looks good to me @gjenkins8 i added you to this to review as your issue #30892 is mentioned in this PR.

TerryHowe (approved):
/lgtm

It isn't clear from the automated tests that the shuffle is used, so I downloaded and ran to verify

Makefile
```diff
@@ -21,7 +21,7 @@ ACCEPTANCE_RUN_TESTS=.
 PKG         := ./...
 TAGS        :=
 TESTS       := .
-TESTFLAGS   :=
+TESTFLAGS   := -shuffle=on -count=2
```
mattfarina, line 24: `-count=2` will add approximately 2.5 minutes to the test run time in CI. Running the tests twice currently takes about that much longer.

Is this something we want to do? What's the benefit?

pkg/lint/support/message_test.go
```diff
@@ -45,6 +44,7 @@ func TestRunLinterRule(t *testing.T) {
 		{-1, errLint, 4, false, ErrorSev},
 	}
 
+	var linter = Linter{}
```
gjenkins8, line 47: I created https://github.com/helm/helm/pull/31024 to focus this PR on changing the test execution only

Makefile
gjenkins8: Given it does practically cause unit tests to take twice as long, and this can be a few minutes on low core machines.  think we can/should keep local dev to the default count (`-count=1`). That means a user who runs `make unit-test` locally, will retain today's behavior.

Perhaps this means we should pass the test count in from the GitHub action. But before making changes, it is likely worth waiting for the discussion on https://github.com/helm/helm/issues/30892#issuecomment-3030678499 to happen.

---

To pass in from the GitHub acton ( https://github.com/helm/helm/blob/main/.github/workflows/build-test.yml#L32 ), I would recommend something like:

```
TESTFLAGS   := -shuffle=on
ifneq ($(TESTCOUNT),)
TESTFLAGS += -count=$(TESTCOUNT)
endif
```

Then the action can do:
```
      - name: Run unit tests
        run: make test-coverage TESTCOUNT=2
```

(but again, before making changes, worth waiting for discussion)

AlfonsoUceda: Will wait until agreed.

TerryHowe: There was a discussion item brought up by Benoit https://docs.google.com/document/d/1d-6xJEx0C78csIYSPKJzRPeWaHG_8W1Hjl72OJggwdc/edit?tab=t.0#heading=h.cafxi2yfgqg1

The notes aren't entirely clear, but as I recall : make it configurable, default count=1, CI test overrides with test count=2

benoittgt: > CI test overrides with test count=2
For this it was suggested that it could be daily task.

AlfonsoUceda:
Sorry folks I have to close this because I don't have time to work on this.

PR #31015, opened 2025-06-27
Add linter support for the `crds/` directory

zachburg:

Check that the `crds/` directory only contains YAML files that define K8s resources with `kind: CustomResourceDefinition`.

Checking that the YAML files are not templates will be done in a separate commit. (https://helm.sh/docs/topics/charts/#custom-resource-definitions-crds)

Refs #31014 

**If applicable**:


pkg/lint/rules/crds.go
```diff
@@ -0,0 +1,97 @@
...
+	"helm.sh/helm/v4/pkg/chart/v2/loader"
+	"helm.sh/helm/v4/pkg/lint/support"
+)
+
+// Crds lints the CRDs in the Linter.
+func Crds(linter *support.Linter) {
+	fpath := "crds/"
+	crdsPath := filepath.Join(linter.ChartDir, fpath)
+
+	crdsDirExist := linter.RunLinterRule(support.WarningSev, fpath, validateCrdsDir(crdsPath))
```
benoittgt: But when the folder does no exist we do not return error, so `!crdsDirExist` will be false?

zachburg: This is copied from https://github.com/helm/helm/blob/main/pkg/lint/rules/template.go, which silently discards the error from `os.Stat`.

`crds/` is an optional directory in a chart, so it OK if it does not exist. But if does exist, we should make sure it is a directory.

zachburg: Oh, I see where the issue is, which I also means the template linter should also be fixed too.

pkg/lint/rules/crds.go
```diff
@@ -0,0 +1,100 @@
...
+
+	chartLoaded := linter.RunLinterRule(support.ErrorSev, fpath, err)
+
+	if !chartLoaded {
+		return
+	}
+
+	/* Iterate over all the CRDs to check:
+	- It is a YAML file and not a template
+	- The kind is CustomResourceDefinition
```
gjenkins8: very minor nit, a numbered list would read better here:
```suggestion
	1. It is a YAML file and not a template
	2. The kind is CustomResourceDefinition
```

pkg/lint/rules/crds.go
```diff
@@ -0,0 +1,100 @@
...
+func Crds(linter *support.Linter) {
+	fpath := "crds/"
+	crdsPath := filepath.Join(linter.ChartDir, fpath)
+
+	// crds directory is optional
+	if _, err := os.Stat(crdsPath); errors.Is(err, fs.ErrNotExist) {
+		return
+	}
+
+	linter.RunLinterRule(support.WarningSev, fpath, validateCrdsDir(crdsPath))
```
gjenkins8: If the CRDs dir is not valid, then should we return early?

pkg/lint/rules/crds.go
```diff
@@ -0,0 +1,100 @@
...
+
+	/* Iterate over all the CRDs to check:
+	- It is a YAML file and not a template
+	- The kind is CustomResourceDefinition
+	*/
+	for _, crd := range chart.CRDObjects() {
+		fileName := crd.Name
+		fpath = fileName
+
+		decoder := yaml.NewYAMLOrJSONDecoder(bytes.NewReader(crd.File.Data), 4096)
```
gjenkins8, line 68: Is there a reason you chose `NewYAMLOrJSONDecoder` rather than e.g. `NewYAMLReader`?

Using `NewYAMLOrJSONDecoder` doesn't "hurt", but given JSON is subset of YAML, we don't need the extra complexity(?)

pkg/lint/rules/crds.go
```diff
@@ -0,0 +1,100 @@
...
+		return err
+	}
+	if !fi.IsDir() {
+		return errors.New("not a directory")
+	}
+	return nil
+}
+
+func validateCrdKind(obj *K8sYamlStruct) error {
+	if obj.Kind != "CustomResourceDefinition" {
```
gjenkins8, line 109: Should we also validate that the API Group is `apiextensions.k8s.io` ? It seems a little (very) far fetched, but in theory a user could have their own custom `CustomResourceDefinition` kind.

pkg/lint/rules/crds_test.go
```diff
@@ -0,0 +1,51 @@
...
+package rules
+
+import (
+	"strings"
+	"testing"
+
+	"helm.sh/helm/v4/pkg/lint/support"
+)
+
+const crdsTestBaseDir = "./testdata/withcrd"
```
gjenkins8: can we add a (valid) `crds/` dir to an existing chart in `testdata/`? (to avoid having yet another test data set; given this is for the good/happy test case)

pkg/lint/rules/crds_test.go
```diff
@@ -0,0 +1,51 @@
...
+const invalidCrdsDir = "./testdata/invalidcrdsdir"
+
+func TestCrdsDir(t *testing.T) {
+	linter := support.Linter{ChartDir: crdsTestBaseDir}
+	Crds(&linter)
+	res := linter.Messages
+
+	if len(res) > 0 {
+		t.Fatalf("Expected no errors, got %d, %v", len(res), res)
+	}
```
gjenkins8: we make heavy usage of https://github.com/stretchr/testify in Helm already, this could be written as:
```
assert.Emptyf(t, res, "Expected no errors, got %v", res)
```

pkg/lint/rules/crds_test.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func TestInvalidCrdsDir(t *testing.T) {
+	linter := support.Linter{ChartDir: invalidCrdsDir}
+	Crds(&linter)
+	res := linter.Messages
+
+	if len(res) != 1 {
```
gjenkins8: similar to above:

```
assert.Len(t, res, 1)
```

pkg/lint/rules/crds_test.go
```diff
@@ -0,0 +1,51 @@
...
+		t.Fatalf("Expected one error, got %d, %v", len(res), res)
+	}
+
+	if !strings.Contains(res[0].Err.Error(), "not a directory") {
+		t.Errorf("Unexpected error: %s", res[0])
+	}
```
gjenkins8: similar:
```
assert.ErrorContains(t, err, "not a directory")
```

pkg/lint/lint_test.go
```diff
@@ -111,6 +112,16 @@ func TestBadValues(t *testing.T) {
 	}
 }
 
+func TestBadCrdFile(t *testing.T) {
+	m := RunAll(badCrdFileDir, values, namespace).Messages
+	if len(m) < 1 {
```
gjenkins8: similar to other comments, use https://github.com/stretchr/testify ?

pkg/lint/lint_test.go
```diff
@@ -111,6 +112,16 @@ func TestBadValues(t *testing.T) {
...
+		t.Fatalf("All didn't fail with expected errors, got %#v", m)
+	}
+	if !strings.Contains(m[0].Err.Error(), "object kind is not 'CustomResourceDefinition'") {
```
gjenkins8: similar to other comments, use https://github.com/stretchr/testify ?

gjenkins8 (commented):
Thanks for the PR! Some (mostly minor) comments

pkg/lint/rules/crds_test.go
zachburg: Done.

pkg/lint/lint_test.go
zachburg: Is there an open issue to switch Helm to https://github.com/stretchr/testify? It feels odd to mix test styles in the same file.

pkg/lint/rules/crds.go
zachburg: Done.

pkg/lint/rules/crds.go
zachburg: `NewYAMLOrJSONDecoder` accepts a `bytes.Reader`, `NewYAMLReader` requires creating a `bytes.Reader` and passing it to `bufio.NewReader`. I also saw it was already used in https://github.com/helm/helm/blob/main/pkg/lint/rules/template.go#L138.

It also handles decoding into the YAML struct so we don't have to do it.

Unfortunately, there is no `NewYAMLDecoder`, which would be best.

pkg/lint/rules/crds.go
zachburg: Done.

gjenkins8 (commented):
just this question remains: https://github.com/helm/helm/pull/31015/files#r2187431837

(I think it was auto resolved when you pushed your last updates)

gjenkins8 (approved):
nice, thanks for the updates!

TerryHowe (approved):
/lgtm

joejulian (approved):
LGTM

PR #31018, opened 2025-07-01
Move logging setup to be configurable

mattfarina:
**What this PR does / why we need it**: If another application uses NewRootCmd it will step any logger they have setup. This change allows NewRootCmd to be called and it tied to the logger they are using in their application as long as it works through the slog interface.

**Special notes for your reviewer**: Parsing flags was used further up to make sure debug is available to pass around.

**If applicable**:


robertsirc (approved):
LGTM

PR #31019, opened 2025-07-01
Return early when linting if the `templates/` directory does not exist

zachburg:

The comments in pkg/lint/rules/template.go say that it will return early if the `templates/` directory does not exist, but `validateTemplatesDir` silently discards the error from `os.Stat` and still passes if `templates/` does not exist. So check first if file path exists before running any lint checks.

Discovered in https://github.com/helm/helm/pull/31015, which adds linter support for the `crds/` directory.

**If applicable**:

### DIFF
--- a/pkg/lint/rules/template.go
+++ b/pkg/lint/rules/template.go
@@ -54,12 +54,12 @@ func TemplatesWithSkipSchemaValidation(linter *support.Linter, values map[string
 	fpath := "templates/"
 	templatesPath := filepath.Join(linter.ChartDir, fpath)
 
-	templatesDirExist := linter.RunLinterRule(support.WarningSev, fpath, validateTemplatesDir(templatesPath))
-
 	// Templates directory is optional for now
-	if !templatesDirExist {
+	if _, err := os.Stat(templatesPath); errors.Is(err, os.ErrNotExist) {
 		return
 	}
+
+	linter.RunLinterRule(support.WarningSev, fpath, validateTemplatesDir(templatesPath))
 
 	// Load chart and parse templates
 	chart, err := loader.Load(linter.ChartDir)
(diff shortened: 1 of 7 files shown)


pkg/lint/rules/template.go
```diff
@@ -54,13 +54,13 @@ func TemplatesWithSkipSchemaValidation(linter *support.Linter, values map[string
 	fpath := "templates/"
 	templatesPath := filepath.Join(linter.ChartDir, fpath)
 
-	templatesDirExist := linter.RunLinterRule(support.WarningSev, fpath, validateTemplatesDir(templatesPath))
-
 	// Templates directory is optional for now
-	if !templatesDirExist {
+	if _, err := os.Stat(templatesPath); errors.Is(err, os.ErrNotExist) {
```
gjenkins8: The `templates/` directory is required for Helm charts. The linter should report if it is missing I think.

I think this `errors.Is(err, os.ErrNotExist)` check should be moved into `validateTemplatesDir`.

And we need to retain the `if !templatesDirExist { return }` logic

gjenkins8 (changes requested):
https://github.com/helm/helm/pull/31019/files#r2187994497

pkg/lint/rules/template.go
zachburg: Making the `templates/` directory required might be a breaking change. Is it OK to use the warning severity if it does not exist?

zachburg: Updated to raise a warning if `templates/` does not exist, and an error if it is not a directory.

gjenkins8 (approved):
thanks!

pkg/lint/rules/template.go
gjenkins8: > Making the templates/ directory required might be a breaking change. Is it OK to use the warning severity if it does not exist?

Agree with the warning (to answer this question and thinking more). You could e.g. have an umbrella chart with no templates (only child charts). In this case, silencing the linter is simply `mkdir templates`. And also while the docs do indicate `templates/` dir is non-optional: https://helm.sh/docs/topics/charts/#the-chart-file-structure , in practice we just consume templates from that dir if they exist.

joejulian (approved):
lgtm

PR #31022, opened 2025-07-02
helm search: support for oci:// repos

### LINKED ISSUE #11000: Allow listing Versions with Helm on OCI
As OCI doesn’t provide a way to index / list chart versions, it could still be implemented Helm Client Side:
- it’s always possible to list the latest versions 
- By specifying SemVer level e.g. patch, it would be possible to iterate over the versions as there should be no gaps on a SemVer Level
- e.g. 0.0.17 is latest, Start version could be 0.0.0 helm could try to check all versions in between
- It’s a bit hacky and won’t support all features of SemVer but definitely helpful for CI/CD where you want to deploy automatically in a version Range

jdolitsky:

This PR resolves https://github.com/helm/helm/issues/11000, adding `helm search` capabilities for OCI-based repos.

This should be completely backwards-compatible. The one limitation is that search on `oci://` repos is more limited, since OCI does not have a good way to list repos (besides the unofficial catalog endpoint which maybe we can add later).

This lists tags on the repo provided, and fetches the config layer to get the metadata about the chart like app version and description.

The result is the following:
```
$ helm search repo oci://ghcr.io/stefanprodan/charts/podinfo
NAME                                     	CHART VERSION	APP VERSION	DESCRIPTION                      
oci://ghcr.io/stefanprodan/charts/podinfo	6.9.0        	6.9.0      	Podinfo Helm chart for Kubernetes
```

This also supports multi-version listing using `--versions`:

```
$ helm search repo oci://ghcr.io/stefanprodan/charts/podinfo --versions
NAME                                     	CHART VERSION	APP VERSION	DESCRIPTION                      
oci://ghcr.io/stefanprodan/charts/podinfo	6.9.0        	6.9.0      	Podinfo Helm chart for Kubernetes
oci://ghcr.io/stefanprodan/charts/podinfo	6.8.0        	6.8.0      	Podinfo Helm chart for Kubernetes
oci://ghcr.io/stefanprodan/charts/podinfo	6.7.1        	6.7.1      	Podinfo Helm chart for Kubernetes
oci://ghcr.io/stefanprodan/charts/podinfo	6.7.0        	6.7.0      	Podinfo Helm chart for Kubernetes
oci://ghcr.io/stefanprodan/charts/podinfo	6.6.3        	6.6.3      	Podinfo Helm chart for Kubernetes
```

Since this makes several requests to registry, it limits itself to 5 versions to prevent rate limiting etc. The can also be overridden using the `--max-versions=<num>` flag:

```
$ helm search repo oci://ghcr.io/stefanprodan/charts/podinfo --versions --max-versions=10
NAME                                     	CHART VERSION	APP VERSION	DESCRIPTION                      
oci://ghcr.io/stefanprodan/charts/podinfo	6.9.0        	6.9.0      	Podinfo Helm chart for Kubernetes
oci://ghcr.io/stefanprodan/charts/podinfo	6.8.0        	6.8.0      	Podinfo Helm chart for Kubernetes
oci://ghcr.io/stefanprodan/charts/podinfo	6.7.1        	6.7.1      	Podinfo Helm chart for Kubernetes
oci://ghcr.io/stefanprodan/charts/podinfo	6.7.0        	6.7.0      	Podinfo Helm chart for Kubernetes
oci://ghcr.io/stefanprodan/charts/podinfo	6.6.3        	6.6.3      	Podinfo Helm chart for Kubernetes
oci://ghcr.io/stefanprodan/charts/podinfo	6.6.2        	6.6.2      	Podinfo Helm chart for Kubernetes
oci://ghcr.io/stefanprodan/charts/podinfo	6.6.1        	6.6.1      	Podinfo Helm chart for Kubernetes
oci://ghcr.io/stefanprodan/charts/podinfo	6.6.0        	6.6.0      	Podinfo Helm chart for Kubernetes
oci://ghcr.io/stefanprodan/charts/podinfo	6.5.4        	6.5.4      	Podinfo Helm chart for Kubernetes
oci://ghcr.io/stefanprodan/charts/podinfo	6.5.3        	6.5.3      	Podinfo Helm chart for Kubernetes
```

In cases where the chart is tagged multiple times, with no difference besides the chart version, semver sorting is implemented to prefer the major.minor.bugfix version. For example, for a chart tagged `20`, `20.0`, `20.0.1-r1`, and `20.0.1`, the `20.0.1` tag is preferred in the search output.

Lastly, tags prefixed with `sha256-` (sigstore/OCI referrers) are ignored completely.

**If applicable**:

### DIFF
--- a/pkg/cmd/search_repo.go
+++ b/pkg/cmd/search_repo.go
@@ -31,6 +31,7 @@ import (
 	"github.com/gosuri/uitable"
 	"github.com/spf13/cobra"
 
+	chart "helm.sh/helm/v4/pkg/chart/v2"
 	"helm.sh/helm/v4/pkg/cli/output"
 	"helm.sh/helm/v4/pkg/cmd/search"
 	"helm.sh/helm/v4/pkg/helmpath"
@@ -64,19 +65,27 @@ Repositories are managed with 'helm repo' commands.
 const searchMaxScore = 25
 
 type searchRepoOptions struct {
-	versions       bool
-	regexp         bool
-	devel          bool
-	version        string
-	maxColWidth    uint
-	repoFile       string
-	repoCacheDir   string
-	outputFormat   output.Format
-	failOnNoResult bool
+	versions              bool
+	regexp                bool
+	devel                 bool
+	version               string
+	maxColWidth           uint
+	repoFile              string
+	repoCacheDir          string
+	outputFormat          output.Format
+	failOnNoResult        bool
+	certFile              string
+	keyFile               string
+	caFile                string
+	insecureSkipTLSverify bool
+	plainHTTP             bool
+	username              string
+	password              string
+	maxVersions           int
 }
 
 func newSearchRepoCmd(out io.Writer) *cobra.Command {
-	o := &searchRepoOptions{}
+	o := &searchRepoOptions{maxVersions: 5}
 
 	cmd := &cobra.Command{
 		Use:   "repo [keyword]",
@@ -96,6 +105,11 @@ func newSearchRepoCmd(out io.Writer) *cobra.Command {
 	f.StringVar(&o.version, "version", "", "search using semantic versioning constraints on repositories you have added")
 	f.UintVar(&o.maxColWidth, "max-col-width", 50, "maximum column width for output table")
 	f.BoolVar(&o.failOnNoResult, "fail-on-no-result", false, "search fails if no results are found")
+	f.BoolVar(&o.plainHTTP, "plain-http", false, "use insecure HTTP connections for the registry")
+	f.BoolVar(&o.insecureSkipTLSverify, "insecure-skip-tls-verify", false, "skip TLS certificate checks for the registry")
+	f.StringVar(&o.username, "username", "", "registry username")
+	f.StringVar(&o.password, "password", "", "registry password")
+	f.IntVar(&o.maxVersions, "max-versions", 5, "maximum number of versions to fetch when searching OCI registries (to prevent rate limiting)")
 
 	bindOutputFlag(cmd, &o.outputFormat)
 
@@ -104,6 +118,11 @@ func newSearchRepoCmd(out io.Writer) *cobra.Command {
 
 func (o *searchRepoOptions) run(out io.Writer, args []string) error {
 	o.setupSearchedVersion()
+
+	// Check if searching for OCI registry
+	if len(args) > 0 && strings.HasPrefix(args[0], "oci://") {
+		return o.searchOCI(out, args[0])
+	}
 
 	index, err := o.buildIndex()
 	if err != nil {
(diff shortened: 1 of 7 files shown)


pkg/registry/client.go
```diff
@@ -790,6 +791,282 @@ func (c *Client) Tags(ref string) ([]string, error) {
...
+
+// Search lists all versions of a chart in an OCI registry and returns metadata
+func (c *Client) Search(ref string, maxVersions int) ([]SearchResult, error) {
+	// Remove oci:// prefix
+	ref = strings.TrimPrefix(ref, OCIScheme+"://")
+
+	ctx := context.Background()
+	repository, err := remote.NewRepository(ref)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create repository for %s: %w", ref, err)
```
benoittgt, line 882: I'm wondering if it should be different message. From the source code it seems it's mostly an error from [`ParseReference`](https://github.com/oras-project/oras-go/blob/05a2b09cbf2eab1df691411884dc4df741ec56ab/registry/remote/repository.go#L354)

Maybe `invalid repository reference` ?

benoittgt (commented):
That's great to see this PR landing, especially given the amount of interest in the issue.  
I'm wondering if we might want to further split the logic into functions — for example, separating business logic from concurrency ?

pkg/registry/client.go
```diff
@@ -790,6 +791,282 @@ func (c *Client) Tags(ref string) ([]string, error) {
...
+	}
+	repository.PlainHTTP = c.plainHTTP
+	repository.Client = c.authorizer
+
+	var results []SearchResult
+	digestToMetadata := make(map[string]*chart.Metadata)
+	var digestMutex sync.Mutex
+
+	// Add timeout to prevent hanging on large repositories
+	ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
```
benoittgt, line 892: We may want to make this configurable?

pkg/registry/client.go
```diff
@@ -790,6 +791,282 @@ func (c *Client) Tags(ref string) ([]string, error) {
...
+	// Sort tags by semver (newest first) before processing
+	sort.Slice(validTags, func(i, j int) bool {
+		v1, _ := semver.NewVersion(strings.ReplaceAll(validTags[i], "_", "+"))
+		v2, _ := semver.NewVersion(strings.ReplaceAll(validTags[j], "_", "+"))
+		return v1.GreaterThan(v2)
+	})
+
+	// Process only the specified number of tags to avoid overwhelming large repositories
+	// Users can use --max-versions to increase the limit
+	if maxVersions > 0 && len(validTags) > maxVersions {
```
benoittgt, line 930: I'm wondering why we are not using [`TagListPageSize`](https://pkg.go.dev/oras.land/oras-go/v2/registry/remote#Repository.TagListPageSize) ?

pkg/cmd/search_repo.go
```diff
@@ -199,6 +218,59 @@ func (o *searchRepoOptions) buildIndex() (*search.Index, error) {
 	return i, nil
 }
 
+func (o *searchRepoOptions) searchOCI(out io.Writer, ref string) error {
```
mattfarina, line 221: OCI registries aren't really Helm repositories. Helm repositories have an `index.yaml` file. Helm repos can be searched offline. There are differences.

Helm currently has `helm search hub` and `helm search repo`. I think it would make sense to have `helm search oci` as a 3rd option and for this functionality to exist there. I could even see plugins as an option to plugin with the various providers from harbor to Docker Hub. The OCI distribution spec doesn't have a search API and there are differences out there. Plugins would be a way to deal with that variation.

What do you think?

mattfarina (changes requested):

pkg/cmd/search_repo.go
jdolitsky: Fair points- one thing that came up in conversations on this (although I cannot seem to find an existing issue) is that `helm upgrade`  does not work on releases sources from an OCI repo. This "Search" functionality is really just an effective way to get the latest version of a chart. Curious if this can also be leveraged to install charts from OCI repos without necessarily specifying the --version/tag... Any thoughts on that @mattfarina ?

jdolitsky: To clarify- if this is simply to satisfy search, then a plugin may make more sense, but it seems that it unlocks a lot of other things

pkg/registry/client.go
joejulian: Agree. As a user, "failed to create repository" sounds like it's trying to create some hosting service.

pkg/cmd/search_repo.go
scottrigby: #i was thinking the same thing re `helm search oci`. 

good point about the different services out there requiring implementations. in helm 4 we now have structured plugin types, so we could now create a "search" plugin type (for example).

josh, re getting the latest ref - is this implementation agnostic? (sorry I can't easily check through the code right now, in transit). if so, we could perhaps have both this in helm core and plugins for per-implementation/advanced search functionality?

promptless-for-oss:
Promptless prepared a documentation update related to this change.

Triggered by [helm/helm#31022](https://github.com/helm/helm/pull/31022)

Added documentation for the new OCI registry search capability, including the `--max-versions` flag that limits version fetching to prevent rate limiting.

Review at https://github.com/helm/helm-www/pull/2010

PR #31034, opened 2025-07-07
Feat: Add color output functionality and tests for release statuses

Mazafard:
closes #31031

This PR introduces ANSI color support to Helm CLI output to improve human readability and align Helm with modern CLI UX practices. The implementation addresses the feature request in #31031 for colorful CLI output, which is now standard in many popular DevOps tools (e.g., kubectl, docker, git, terraform).

**Note**: This is a clean resubmission of PR #31032 which was closed due to merge conflicts with the main branch. This PR contains the same functionality but is properly rebased on the latest upstream/main.

### Key Features:
- **Automatic color support**: Colors are enabled by default for interactive terminals
- **Smart TTY detection**: Automatically disables colors when output is piped to files or non-interactive contexts using `term.IsTerminal()`
- **Opt-out mechanism**: Users can disable colors using the `--no-color` flag or `NO_COLOR` environment variable
- **Targeted implementation**: Adds color support to release status displays across multiple commands

### Implementation Details:
- Added a new color output module (`pkg/cli/output/color.go`) with functions for colorizing status, headers, and namespaces
- Integrated `--no-color` flag support in the CLI environment settings
- Added color formatting to commands: `helm status`, `helm list`, `helm install`, `helm upgrade`, `helm get all`, and `helm test`
- Created unit tests (219 lines) to ensure color formatting works correctly
- Uses the `fatih/color` library for ANSI color support
- Removed unused `ColorEnabled()` function to clean up the codebase

### Benefits:
- **Improved readability**: Visual distinction for release statuses (deployed in green, failed in red, pending in yellow)
- **Better debugging**: Easier to spot issues during CI/CD or release debugging
- **Enhanced developer experience**: Bold headers and colored namespaces help users quickly parse output
- **Modern CLI UX**: Aligns Helm with other popular DevOps tools while maintaining backward compatibility

The current implementation provides a foundation for color support with the following color scheme:
- **Green**: Deployed/successful releases
- **Red**: Failed releases or unknown status
- **Yellow**: Pending operations (install, upgrade, rollback, uninstalling)
- **Bold**: Headers for better readability
- **Cyan**: Namespaces to distinguish them from other text
- **Default**: Uninstalled, superseded, and other statuses remain uncolored

The implementation respects terminal capabilities through TTY detection and honors the `NO_COLOR` environment variable standard. All color output can be disabled using the `--no-color` flag or by setting the `NO_COLOR` environment variable, ensuring full backward compatibility for automation and scripting use cases.

Note: While issue #31031 mentioned opt-in colors (disabled by default), this implementation enables colors by default for interactive terminals, following the pattern of modern CLI tools. The opt-out mechanism via `--no-color` flag provides the necessary control for users who prefer no colors.

**Updated based on review feedback**: Changed the default color mode from "never" to "auto" to enable colors by default for interactive terminals, following team consensus for the v4 release.

### New Files:
- `pkg/cli/output/color.go` - Color formatting functions
- `pkg/cli/output/color_test.go` - Comprehensive unit tests

### Modified Files:
- `pkg/cli/environment.go` - Added `ColorMode` field, `--color` flag support, and `envColorMode()` function
- **Updated**: Changed default color mode from "never" to "auto"
- **Updated**: Removed unused `ColorEnabled()` function
- `cmd/helm/status.go` - Colorized status output
- `cmd/helm/list.go` - Colorized list output
- `cmd/helm/install.go` - Colorized install status
- `cmd/helm/upgrade.go` - Colorized upgrade status
- `cmd/helm/get_all.go` - Colorized get all output
- `cmd/helm/release_testing.go` - Colorized test results

### Additional Fixes:
- Replaced `os.Setenv()` with `t.Setenv()` in test files to comply with golangci-lint recommendations
- Fixed `usetesting` linter warnings for better test isolation

**If applicable**:

## Testing
- All unit tests pass
- Tested with `NO_COLOR` environment variable
- Tested with `HELM_COLOR` environment variable (never/auto/always)
- Verified color output in different terminals (iTerm2, Terminal.app, VS Code terminal)
- Tested piping output to files (colors correctly disabled)
- Verified with golangci-lint v2.1.0
- Tested `--color` flag with all values (never/auto/always)


robertsirc:
So, first thing thank you for the PR. I have a questions to the other maintainers on `--no-color` and if that should be on by default or off by default. Besides that this looks good from my side. I was able to test this and it works as expected.

robertsirc (dismissed):
LGTM

pkg/cli/output/color.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/fatih/color"
+	"golang.org/x/term"
+
+	release "helm.sh/helm/v4/pkg/release/v1"
+)
+
+// ColorizeStatus returns a colorized version of the status string based on the status value
+func ColorizeStatus(status release.Status, noColor bool) string {
+	// Disable color if requested or if not in a terminal
+	if noColor || os.Getenv("NO_COLOR") != "" || !term.IsTerminal(int(os.Stdout.Fd())) {
```
gjenkins8: If I am reading https://github.com/fatih/color?tab=readme-ov-file#disableenable-color correctly, it is unnecessary to re-check terminal info, env var, etc here.

pkg/cli/environment.go
```diff
@@ -160,6 +163,7 @@ func (s *EnvSettings) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&s.RepositoryCache, "repository-cache", s.RepositoryCache, "path to the directory containing cached repository indexes")
 	fs.IntVar(&s.BurstLimit, "burst-limit", s.BurstLimit, "client-side default throttling limit")
 	fs.Float32Var(&s.QPS, "qps", s.QPS, "queries per second used when communicating with the Kubernetes API, not including bursting")
+	fs.BoolVar(&s.NoColor, "no-color", s.NoColor, "disable colorized output")
```
gjenkins8: I think we should have `--color=[never|auto|always]`  like e.g. [grep](https://linux.die.net/man/1/grep) (default: `auto` or `never`*)

> --color[=WHEN], --colour[=WHEN]
              Surround the matched (non-empty) strings, matching lines,
              context lines, file names, line numbers, byte offsets, and
              separators (for fields and groups of context lines) with
              escape sequences to display them in color on the terminal.
              WHEN is never, always, or auto.

(we could also support `--colour` for British English spelling, not sure if either is correct for internationalized tooling).

---

*For the default value of either: `auto` or `never` 

Since we are only supporting a subset of Helm commands for colourized output initially, I'm not sure if it makes sense to enable colourized output by default. Since some sub command output will use colour, others will not!

The alternative would be off by default, until we thought enough sub commands had color output enabled.

(Not sure of the best option, this is for discussion)

pkg/cli/output/color.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	// Use bold for headers
+	return color.New(color.Bold).Sprint(header)
+}
+
+// ColorizeNamespace returns a colorized version of a namespace string
+func ColorizeNamespace(namespace string, noColor bool) string {
+	// Disable color if requested or if not in a terminal
+	if noColor || os.Getenv("NO_COLOR") != "" || !term.IsTerminal(int(os.Stdout.Fd())) {
```
gjenkins8: same as `ColorizeStatus`

pkg/cli/output/color.go
```diff
@@ -0,0 +1,70 @@
...
+	default:
+		// For uninstalled, superseded, and any other status
+		return status.String()
+	}
+}
+
+// ColorizeHeader returns a colorized version of a header string
+func ColorizeHeader(header string, noColor bool) string {
+	// Disable color if requested or if not in a terminal
+	if noColor || os.Getenv("NO_COLOR") != "" || !term.IsTerminal(int(os.Stdout.Fd())) {
```
gjenkins8: same as `ColorizeStatus`

pkg/cli/output/color.go
```diff
@@ -0,0 +1,70 @@
+/*
```
gjenkins8: We should make this package internal (public functions exposed become part of Helm's public SDK, and making e.g. incompatible API changes becomes forbidden). As this is an internal helper for Helm internal purposes only, we should make internal.

gjenkins8 (changes requested):
Some initial comments. 

Requesting change to make the `output` package internal (unless there is a very good reason we think it should be public?)

pkg/cli/output/color.go
Mazafard: Thank you for the review @gjenkins8! 👍 
You're absolutely right. The fatih/color library already handles these checks internally via its NoColor global variable. 
I'll remove the redundant checks and rely on the library's built-in functionality.

Mazafard:
> Some initial comments.
> 
> Requesting change to make the `output` package internal (unless there is a very good reason we think it should be public?)

Thank you for the review @gjenkins8 , Could you clarify what you mean by making the package public? The output package was already public. I just added functions to it since it seemed like the most relevant place for output-related code.
I don't think moving the output package would be wise as it would require extensive changes throughout the codebase.

pkg/cli/environment.go
Mazafard: Thanks for the great suggestion @gjenkins8!

I like the idea of `--color=[never|auto|always]`.  definitely more standard than what I had. Here's what I'm thinking:

1. **Default to "auto"** - seems like the most sensible choice:
   - Color on when it's a terminal
   - Off when piped/redirected  
   - Pretty much what everyone expects from grep, ls, etc.

2. **Implementation**:
   - Global `--color` flag with [never|auto|always]
   - Maybe also support `--colour` for our British friends :)
   - Can leverage fatih/color's terminal detection for the auto mode

3. **About partial implementation**: Yeah, only some commands will have color at first, but I still think "auto" makes sense as the default. Commands without color just show normal output anyway, and the ones that do support it will look better right away. Plus it'll encourage us to add color to more commands over time. 

What do you think?

pkg/cli/output/color.go
Mazafard: Done! I've applied the same fix to `ColorizeNamespace`

pkg/cli/output/color.go
Mazafard: Done! I've applied the same fix to `ColorizeNamespace`

pkg/cli/output/color.go
Mazafard: Good point! I'll move this to `internal/cli/output/color` to keep it internal. Makes sense that color functionality shouldn't be part of Helm's public SDK.👍

pkg/cli/environment.go
```diff
@@ -213,6 +218,23 @@ func envCSV(name string) (ls []string) {
...
+	// Check HELM_COLOR environment variable
+	if v, ok := os.LookupEnv("HELM_COLOR"); ok {
+		v = strings.ToLower(v)
+		switch v {
+		case "never", "auto", "always":
+			return v
+		}
+	}
+	// Default to never (disabled) until more commands support color
+	return "never"
```
joejulian: I differ in opinion from @gjenkins8. I don't see any harm in using auto by default, especially as an initial v4 release of this feature. If we default to `never` prior to locking v4, we'll be stuck with that default until v5. Let's do it now and ensure that users don't shoot themselves in the foot by coding around a default we'd like to later change.

robertsirc: I agree with this approach. Could we discuss further on this @gjenkins8 ?

gjenkins8: sure, lets default to "auto" then

pkg/cli/environment.go
```diff
@@ -265,3 +287,25 @@ func (s *EnvSettings) SetNamespace(namespace string) {
 func (s *EnvSettings) RESTClientGetter() genericclioptions.RESTClientGetter {
 	return s.config
 }
+
+// ColorEnabled returns true if color output should be enabled based on the ColorMode setting
+func (s *EnvSettings) ColorEnabled() bool {
```
gjenkins8: I think this is unused now?

Mazafard: You're right!

Mazafard:
@gjenkins8 @joejulian @robertsirc Would you mind taking a look when you get a chance?

robertsirc (approved):
LGTM

gjenkins8 (approved):

PR #31042, opened 2025-07-09
fix: test teardown dns data race

TerryHowe:

Fixes data race in mock dns teardown


robertsirc (approved):
LGTM

robertsirc:
I pulled this down, built it, ran all the test, they all pass. Compared the test run times locally and there was no visible impact in run times.

PR #31057, opened 2025-07-11
Pass credentials when either chart repo or repo dont specify a port but it matches the default port of that scheme

danilobuerger:

As a user I would normally specify my repository without a port if its just on the default one, like: 

```
https://example.com/foo
```

The `index.yaml` might however contain urls with the port (as is the case with Artifactory).

So then you end up with a comparison like

```
u1.Host = "example.com"
u2.Host = "example.com:443"
```

which currently fails. Checking for the default port would allow the credentials to still pass.

**If applicable**:

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -746,6 +746,21 @@ OUTER:
 	return nil
 }
 
+func portOrDefault(u *url.URL) string {
+	if p := u.Port(); p != "" {
+		return p
+	}
+
+	switch u.Scheme {
+	case "http":
+		return "80"
+	case "https":
+		return "443"
+	default:
+		return ""
+	}
+}
+
 // LocateChart looks for a chart directory in known places, and returns either the full path or an error.
 //
 // This does not ensure that the chart is well-formed; only that the requested filename exists.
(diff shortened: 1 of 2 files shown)


pkg/action/install.go
```diff
@@ -833,7 +848,7 @@ func (c *ChartPathOptions) LocateChart(name string, settings *cli.EnvSettings) (
 		// Host on URL (returned from url.Parse) contains the port if present.
 		// This check ensures credentials are not passed between different
 		// services on different ports.
-		if c.PassCredentialsAll || (u1.Scheme == u2.Scheme && u1.Host == u2.Host) {
+		if c.PassCredentialsAll || (u1.Scheme == u2.Scheme && u1.Hostname() == u2.Hostname() && portOrDefault(u1) == portOrDefault(u2)) {
```
TerryHowe: I like the idea of creating a urlEquals with some tests to cover the cases for all these conditions `(u1.Scheme == u2.Scheme && u1.Hostname() == u2.Hostname() && portOrDefault(u1) == portOrDefault(u2))`

danilobuerger: done

TerryHowe (approved):
/lgtm

joejulian (approved):
lgtm and I don't see any negative security implications

robertsirc (approved):
LGTM

PR #31062, opened 2025-07-15
Updated chart name check for create and lint

### LINKED ISSUE #10537: helm lint misses bad chart names

JanitSri:
- closes #10537 
- validates chart names as outlined by [docs](https://helm.sh/docs/chart_best_practices/conventions/#chart-names)

**If applicable**:


pkg/lint/rules/chartfile.go
```diff
@@ -113,12 +121,11 @@ func validateChartYamlStrictFormat(chartFileError error) error {
...
-	if cf.Name == "" {
-		return errors.New("name is required")
+	if cf.Name == "" || len(cf.Name) > maxChartNameLength {
+		return fmt.Errorf("chart name must be between 1 and %d characters", maxChartNameLength)
 	}
-	name := filepath.Base(cf.Name)
-	if name != cf.Name {
-		return fmt.Errorf("chart name %q is invalid", cf.Name)
+	if !chartName.MatchString(cf.Name) {
+		return fmt.Errorf("chart name must use only lowercase letters, digits, and dashes (no underscores, dots, or uppercase letters)")
```
TerryHowe:
```suggestion
		return fmt.Errorf("chart name must use only lowercase letters or digits optionally with dash separators")
```
I think a list of the allowed characters would be better

pkg/lint/rules/chartfile_test.go
```diff
@@ -76,6 +76,46 @@ func TestValidateChartName(t *testing.T) {
...
+		{Name: "chart_name"},             // underscore
+		{Name: "chart.name"},             // dot
+		{Name: "chart--name"},            // double dash (optional)
+		{Name: "-chartname"},             // starts with dash
+		{Name: "chartname-"},             // ends with dash
+		{Name: "chart$name"},             // special character
+		{Name: "chart/name"},             // forward slash
+		{Name: "chart\\name"},            // backslash
+		{Name: "chart name"},             // space
+		{Name: strings.Repeat("a", 251)}, // 241 chars — too long
```
TerryHowe:
```suggestion
		{Name: strings.Repeat("a", 251)}, // 251 chars — too long
```

TerryHowe (changes requested):

pkg/lint/rules/chartfile.go
robertsirc: I agree list of allowed characters would be better as not to make the user guess on what is allowed.

JanitSri: Sounds good! I updated to include allowed characters.

pkg/lint/rules/chartfile_test.go
JanitSri: good catch! updated.

TerryHowe (dismissed):
/lgtm

PR #31074, opened 2025-07-17
add missing template directory to badcrdfile testdata

joejulian:
#31019 was merged after #31015 which caused the test in 31019 to fail as it didn't have a templates directory, creating an unexpected warning.

This ensures that the templates directory exists which allows the result to be output as expected. 

**If applicable**:


TerryHowe (approved):
/lgtm

mattfarina (approved):
lgtm

PR #31081, opened 2025-07-22
Initial addition of v3 charts

mattfarina:
This change adds v3 charts. The code for the chart, including a loader, is present. It is based on v2 charts as a starting point.

Note, this change does not make the charts available for use with Helm CLI commands or the action package. That will be in follow-up changes.

**If applicable**:


TerryHowe (approved):
/lgtm

I just downloaded this PR and diffed with pkg/chart/v2 and looks reasonable. The changes look as expected.

scottrigby (dismissed):
lgtm :+1:

The diff is quite minimal apart from the move to internal, and all make sense.  For anyone else reviewing this PR, check it out with `diff -u pkg/chart/v2 internal/chart/v3`.

Double-checked and only applicable `v2` -> `v3` replacements missing are in some of the test files, but don't want to hold this initial large move PR up for those. Attaching a patch in case it's remotely helpful. Approving with or without those.

[31081-test-v2-replace.patch](https://github.com/user-attachments/files/21434115/31081-test-v2-replace.patch)

mattfarina:
@scottrigby thanks for catching those issues in the tests. The link to your patch isn't working. It's a 404. I did search and fixed up the ones I found. Hope that covers everything.

scottrigby (approved):
weird yeah i guess gh attachment didn't work. But i just re-reviewed and you got em all :+1:

Issue #31028, opened 2025-07-04
Helm 3.18.3: Go SDK ignores cache folder for certain procedures


douglascamata:
### What happened?

I am using the Go SDK to manually pull a chart from a non-OCI repository and for some reason the pull client is still trying to write to the default XDG cache folder path instead of the one I tell it to use. 

I create my settings like this:

```go
settings := cli.New()
cacheFolder, err := os.MkdirTemp(os.TempDir(), "helm-cache")
if err != nil {
	return nil, fmt.Errorf("failed to create cache folder: %w", err)
}
settings.RepositoryCache = cacheFolder
```

Then the pull client is set up like so:

```go
pullClient := action.NewPullWithOpts(action.WithConfig(actionConfig))
pullClient.Settings = h.settings
pullClient.DestDir = h.settings.RepositoryCache
pullClient.Version = chartVersion
```

And finally I pull my chart like so:

```go
pullClient.RepoURL = repoURL
repoEntry := &repo.Entry{
	Name: "temp-repo",
	URL:  repoURL,
}
helmRepo, err := repo.NewChartRepository(repoEntry, getter.All(h.settings))
helmRepo.CachePath = h.settings.RepositoryCache
if err != nil {
	return "", "", nil, fmt.Errorf("failed to create chart repository: %w", err)
}
if _, err := helmRepo.DownloadIndexFile(); err != nil {
	return "", "", nil, fmt.Errorf("failed to download repository index: %w", err)
}
h.logger.Debug("pull client dest dir", zap.String("path", pullClient.DestDir))
result, err := pullClient.Run(chartDownloadRef)
```

### What did you expect to happen?

For the chart to be pulled correctly into `pullClient.DestDir` and any cache to be written to `settings.RepositoryCache`.

Instead of this, it fails with the following error: `looks like \"https://cgx.jfrog.io/artifactory/coralogix-charts-virtual\" is not a valid chart repository or cannot be reached: open /.cache/helm/repository/uHi1Binjiq+3SUKLYEHanfXfRQc=-index.yaml: no such file or directory`.

For some reason it is ignoring my settings for the repository cache. My logs just before the error confirm that my dest dir and repository cache are what I expect them to be:

```
{"level":"debug","ts":1751627480.1150873,"caller":"helm/chart_downloader.go:83","msg":"pull client dest dir","path":"/tmp/helm-cache296943029"}
[...]

PR #31092, opened 2025-07-25
fix(chartrepository): cache path customization in `FindChartInRepoURL` funcs 

douglascamata:

This PR addresses the issue reported in #31028 for the v3 branch. It doesn't necessarily close it yet because the problem still needs fixing in the v4 branch. I am planning to open a similar PR to fix the v4 code after this.

This pull request adds a `cachePath` parameter to all the functions in the `FindChartInRepoURL` family.

Following the previous change, there were updates to three places where that functionw is being used:

- In the `action.ChartPathOptions.LocateChart()` method: to adopt the cache path set from its `settings` parameter.
- In the `action.Pull.Run()` method: to adopt the cache path from its settings field.
- in the `downloader.Manager`: to adopt the cache path set in the manager itself.

**If applicable**:

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -796,7 +796,7 @@ func (c *ChartPathOptions) LocateChart(name string, settings *cli.EnvSettings) (
 		dl.Verify = downloader.VerifyAlways
 	}
 	if c.RepoURL != "" {
-		chartURL, err := repo.FindChartInAuthAndTLSAndPassRepoURL(c.RepoURL, c.Username, c.Password, name, version,
+		chartURL, err := repo.FindChartInAuthAndTLSAndPassAndCacheRepoURL(c.RepoURL, c.Username, c.Password, name, version,
 			c.CertFile, c.KeyFile, c.CaFile, c.InsecureSkipTLSverify, c.PassCredentialsAll, getter.All(settings), settings.RepositoryCache)
 		if err != nil {
 			return "", err
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -797,7 +797,7 @@ func (c *ChartPathOptions) LocateChart(name string, settings *cli.EnvSettings) (
 	}
 	if c.RepoURL != "" {
 		chartURL, err := repo.FindChartInAuthAndTLSAndPassRepoURL(c.RepoURL, c.Username, c.Password, name, version,
-			c.CertFile, c.KeyFile, c.CaFile, c.InsecureSkipTLSverify, c.PassCredentialsAll, getter.All(settings))
+			c.CertFile, c.KeyFile, c.CaFile, c.InsecureSkipTLSverify, c.PassCredentialsAll, getter.All(settings), settings.RepositoryCache)
 		if err != nil {
 			return "", err
 		}
--- a/pkg/action/pull.go
+++ b/pkg/action/pull.go
@@ -122,7 +122,7 @@ func (p *Pull) Run(chartRef string) (string, error) {
 	}
 
 	if p.RepoURL != "" {
-		chartURL, err := repo.FindChartInAuthAndTLSAndPassRepoURL(p.RepoURL, p.Username, p.Password, chartRef, p.Version, p.CertFile, p.KeyFile, p.CaFile, p.InsecureSkipTLSverify, p.PassCredentialsAll, getter.All(p.Settings), p.Settings.RepositoryCache)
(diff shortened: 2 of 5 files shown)


robertsirc:
This looks to be a breaking change to our `dev-v3` branch are you sure this is where you wanted this change to go?

douglascamata:
@robertsirc maybe I'm missing something, but why would it be a breaking change? Given what I understand, which could be wrong or incomplete, it's more like a bugfix/enhancement because today I configure everything I can to use a certain cache dir, so I expect that dir to be used, and yet there are functions somewhere that ignore my configuration (as seen in the issue #31028).

One thing to note too is that the files are immediately deleted after the execution of the `FindChartInAuthAndTLSAndPassRepoURL` function, so in the end there should be leftovers.

douglascamata:
This is particularly annoying when running Helm or using it via the Go SDK in an environment that has a read-only filesystem and I'm manually mounting writeable folders for Helm to use. Any user customizing their cache folder would expect everything that needs caching to use that folder and not anything else.

robertsirc:
Hi @douglascamata,  I guess I should ask, if we change this now and an SDK user is calling this function `FindChartInRepoURL` wouldn't it break for them? We are adding parameter. I would think if there is a way that we could do this without changing the contract?

douglascamata:
@robertsirc aaaah, I see what you mean. There's only one way to fix this without changing the contract: to create a new function in the `FindChartInRepoURL` family, something like `FindChartInRepoURLWithCustomCacheFolder` and then use that in the appropriate places. I'll do this change.

Ideally, in the v4 version of this function (if it exists) we should use a custom type to carry the parameters of the function (it has waaaayy too many). This approach makes it much easier to add parameters without breaking the contract, because than you just have a zero-value to deal with.

douglascamata:
@robertsirc fixed it in [`b05e1c6` (#31092)](https://github.com/helm/helm/pull/31092/commits/b05e1c6d965f47a9b98c888ea15673f6ee485717). All the previously exported functions retain their original signatures and I created a new function called `FindChartInAuthAndTLSAndPassAndCacheRepoURL` that is used where applicable.

robertsirc:
Thanks @douglascamata for the PR, I didn't want to leave you hanging. I am on holiday this week. Let's see if we can get some other maintainers into looks at this in the mean time, thank you ever so much.

douglascamata:
No worries and thanks for the heads up, @robertsirc. Enjoy your holiday!

mattfarina (changes requested):
The cache location isn't a good cache to rely on. I'll try to illustrate with an example. Let's say you have a chart with the same name in 2 different repos. And have pulled them so they are in the cache. For example, _mariadb_ and the chart version is _1.2.3_. And, the content of these two is different.

If you use the chart in the cache, which will be picked? It will be the last one downloaded into the cache. The cache isn't content addressable. Names are not globally unique. It's possible to use the wrong chart/versions content.

The caching system really needs to be rewritten.

While I can't give out the name, I was looking at charts for an application with this problem earlier today.

Any ideas how to solve this when the cache works this way?

douglascamata:
> The cache location isn't a good cache to rely on. I'll try to illustrate with an example. Let's say you have a chart with the same name in 2 different repos. And have pulled them so they are in the cache. For example, _mariadb_ and the chart version is _1.2.3_. And, the content of these two is different.
> 
> If you use the chart in the cache, which will be picked? It will be the last one downloaded into the cache. The cache isn't content addressable. Names are not globally unique. It's possible to use the wrong chart/versions content.
> 
> The caching system really needs to be rewritten.
> 
> While I can't give out the name, I was looking at charts for an application with this problem earlier today.
> 
> Any ideas how to solve this when the cache works this way?

Great points, @mattfarina. Indeed I think that the cache system could use a rewrite. It might be too late for a rewrite on v3 (not sure), but there's an opportunity to improve this for v4.

Now, back to our issue here in v3-world: I think we can take advantage of the fact that the logic in the `FindChartIn(...)` functions is downloading the index and chart but cleaning them up right after (see snippet below from `dev-v3` branch at commit `ad28cfd058d267d47c77f91dbb723359123452a9`) to use a subfolder of the `cachePath` dir dedicated (can even be temporary or fixed, it doesn't matter because the files will be removed anyway) for this. This approach keeps it isolated and still handles my usecase of the read-only filesystem with cache folders mounted as writeable.

https://github.com/helm/helm/blob/ad28cfd058d267d47c77f91dbb723359123452a9/pkg/repo/chartrepo.go#L243-L254

What do you think?

PR #31094, opened 2025-07-25
chore(deps): remove phayes/freeport module

mikelolasagasti:

Replaces the `phayes/freeport` module with the standard library's `net.Listen("tcp", "127.0.0.1:0")` idiom. This removes an unnecessary dependency and simplifies the codebase.


TerryHowe (approved):
/lgtm

Tests passed for me

mattfarina (approved):
lgtm

The code here is doing the same thing the external library does.

robertsirc (approved):
LGTM

PR #31096, opened 2025-07-27
feat(template): Add `--notes` flag

### LINKED ISSUE #6901: v3: helm template does not render NOTES.txt
Helm version:
```
$ h3 version
version.BuildInfo{Version:"v3.0.0-rc.3", GitCommit:"2ed206799b451830c68bff30af2a52879b8b937a", GitTreeState:"clean", GoVersion:"go1.13.4"}
```

Steps to recreate:
```
$ h3 create foo
Creating foo
```

```
$ h3 template foo -s templates/NOTES.txt --notes
Error: unknown flag: --notes
```

```
$ tree foo
foo
├── charts
├── Chart.yaml
├── templates
│   ├── deployment.yaml
│   ├── _helpers.tpl
│   ├── ingress.yaml
│   ├── NOTES.txt
│   ├── serviceaccount.yaml
│   ├── service.yaml
│   └── tests
│       └── test-connection.yaml
└── values.yaml
```

How feature works in Helm 2.x 
 ```
$ helm template bar -x templates/NOTES.txt --notes
---
# Source: bar/templates/NOTES.txt
  export POD_NAME=$(kubectl get pods --namespace default -l "app.kubernetes.io/name=bar,app.kubernetes.io/instance=release-name" -o jsonpath="{.items[0].metadata.name}")
```

### PR DESCRIPTION
**Pull request by Bhargav-InfraCloud**

Added a new flag `--notes` to the `helm template` command, with the default set to `false`.  
If enabled, it will print the rendered `NOTES.txt` at the end of the manifests.  
If `--render-subchart-notes` (this flag already exists on the `template` command, as it is inherited from the `install` command) is also enabled, the output will include the subcharts’ notes as well.

Updated the `install` command’s `--render-subchart-notes` flag description to clarify that, in the case of the `template` command, it is only valid when the `--notes` flag is enabled.  
Note: the `template` command uses most of the `install` command’s flags, including `--render-subchart-notes`.

Updated `action.renderResources()` to format the root chart’s notes, and the subcharts’ notes if the `--render-subchart-notes` flag is enabled, and return them as a string. The output depends on whether it is:

- **Dry run**  
  - Notes are prefixed with the source filename and separated with `---`.  
  - For `helm install` / `helm upgrade`, the `--dry-run` flag enables this behavior.  
  - The `helm template` command is considered a dry run by default.  
  - Example:
    ```
    foo NOTES HERE

    bar NOTES HERE

    baz NOTES HERE

    qux NOTES HERE
    ```

- **Not a dry run**  
  - Only the root chart’s notes are rendered by default.  
  - If the `--render-subchart-notes` flag is enabled, the subcharts’ notes are also included.  
  - Each note is prefixed with the source filename and separated with `---`.  
  - Example:
    ```
    ---
    # Source: foo/templates/NOTES.txt
    foo NOTES HERE
    ---
    # Source: foo/charts/bar/templates/NOTES.txt
    bar NOTES HERE
    ---
    # Source: foo/charts/baz/templates/NOTES.txt
    baz NOTES HERE
    ---
    # Source: foo/charts/bar/charts/qux/templates/NOTES.txt
    qux NOTES HERE
    ```

In both cases, the notes are sorted based on the following conditions:  
1. Depth  
2. Alphabetical order (since the notes file name is always the same, the sorting applies to the chart/subchart name).

Added tests for the following scenarios:
 - helm template:
   1. On the chart without notes or subcharts
   2. `--notes` on chart without notes or subcharts
   3. `--render-subchart-notes` on chart without notes or subcharts
   4. `--notes --render-subchart-notes` on chart without notes or subcharts
   5. On the chart with notes without subcharts
   6. `--notes` on chart with notes without subcharts
   7. `--render-subchart-notes` on chart with notes without subcharts
   8. `--notes --render-subchart-notes` on chart with notes without subcharts
   9. On the chart with notes and subchart
   10. `--notes` on chart with notes and subchart
   11. `--render-subchart-notes` on the chart with  notes and subchart
   12. `--notes --render-subchart-notes` on chart  with notes and subchart
   13. `--help`
 - helm install:
   1. On the chart without notes or subcharts
   2. `--dry-run` on chart without notes or subcharts
   3. `--render-subchart-notes` on chart without notes or subcharts
   4. `--dry-run --render-subchart-notes` on chart without notes or subcharts
   5. On the chart with notes without subcharts
   6. `--dry-run` on chart with notes without subcharts
   7. `--render-subchart-notes` on chart with notes without subcharts
   8. `--dry-run --render-subchart-notes` on chart with notes without subcharts
   9. On a chart with notes and subcharts
   10. `--dry-run` on chart with notes and  subcharts
   11. `--render-subchart-notes` on chart with  notes and subcharts
   12. `--dry-run --render-subchart-notes` on chart  with notes and subcharts
   13. `--help`
 - helm upgrade:
   1. On the chart without notes or subcharts
   2. `--dry-run` on chart without notes or subcharts
   3. `--render-subchart-notes` on chart without notes or subcharts
   4. `--dry-run --render-subchart-notes` on chart without notes or subcharts
   5. On the chart with notes without subcharts
   6. `--dry-run` on chart with notes without subcharts
   7. `--render-subchart-notes` on chart with notes without subcharts
   8. `--dry-run --render-subchart-notes` on chart with notes without subcharts
   9. On the chart with notes and subcharts
   10. `--dry-run` on chart with notes and subcharts
   11. `--render-subchart-notes` on chart with  notes and subcharts
   12. `--dry-run --render-subchart-notes` on  chart with notes and subcharts
   13. `--help`

Fixes #6901

- The method `(*Configuration).renderResources()` from `action` package was returning the notes string even when there is an error. This doesn't make any sense as upper layers though assigning it to `rel.Info.Notes`, but never use it as the run itself exits with an error. So, I replaced it with empty strings. No tests are failing. My guess is some old and stale code. But if you think it is needed, let me know.

**If applicable**:

---

## Testing

Follow these steps to test the new and existing notes flags:

### 1. Clone and Build

- Clone the forked Helm repository:
    ```
    git clone https://github.com/Bhargav-InfraCloud/helm
    ```
- Switch to the feature branch:
    ```
    git checkout bhargav-ravuri/6901/add_notes_flag_to_helm_template_command
    ```
- Build the Helm binary:
    ```
    make build
    ```
- Copy the built binary to a directory in your `$PATH`:
    ```
    cp ./bin/helm /usr/local/bin/
    ```

### 2. Create Nested Helm Charts

- Create a temporary directory and nested charts with custom `NOTES.txt` files:
    ```sh
    cd $(mktemp -d)
    helm create foo && echo "foo TEXT HERE" > foo/templates/NOTES.txt
    cd foo/charts
    helm create bar && echo "bar TEXT HERE" > bar/templates/NOTES.txt
    helm create baz && echo "baz TEXT HERE" > baz/templates/NOTES.txt
    cd bar/charts
    helm create qux && echo "qux TEXT HERE" > qux/templates/NOTES.txt
    cd ../../../..
    ```

### 3. Test Commands

Run the following commands on the nested Helm charts you created in the previous step to verify the behavior of the `--notes` and `--render-subchart-notes` flags.

#### helm template

1. `helm template luffy foo`
    - Expected: Helm template output without any notes (from the current chart or subcharts).
    - Output: [1.1.helm-template-luffy-foo.txt](https://github.com/user-attachments/files/22661545/1.1.helm-template-luffy-foo.txt)

2. `helm template luffy foo --notes`
    - Expected: Helm template output with only the current chart's notes.
    - Output: [1.2.helm-template-luffy-foo-with-flag-notes.txt](https://github.com/user-attachments/files/22661547/1.2.helm-template-luffy-foo-with-flag-notes.txt)

3. `helm template luffy foo --render-subchart-notes`
    - Expected: Helm template output without any notes (from the current chart or subcharts), since subchart notes are only rendered when `--notes` is enabled.
    - Output: [1.3.helm-template-luffy-foo-with-flag-render-subchart-notes.txt](https://github.com/user-attachments/files/22661548/1.3.helm-template-luffy-foo-with-flag-render-subchart-notes.txt)

4. `helm template luffy foo --notes --render-subchart-notes`
    - Expected: Helm template output with both the current chart's and subcharts' notes, in order (first by depth, then alphabetically).
    - Output: [1.4.helm-template-luffy-foo-with-flag-notes-and-render-subchart-notes.txt](https://github.com/user-attachments/files/22661550/1.4.helm-template-luffy-foo-with-flag-notes-and-render-subchart-notes.txt)

#### helm install

1. `helm install luffy foo`
    - Expected: Helm install output with the current chart's notes, as `--notes` is enabled by default in `helm install`.
    - Output: [2.1.helm-install-luffy-foo.txt](https://github.com/user-attachments/files/22661551/2.1.helm-install-luffy-foo.txt)

2. `helm install luffy foo --dry-run`
    - Expected: Helm install dry-run output with the current chart's notes, as `--notes` is enabled by default in `helm install`.
    - Output: [2.2.helm-install-luffy-foo-with-flag-dry-run.txt](https://github.com/user-attachments/files/22661552/2.2.helm-install-luffy-foo-with-flag-dry-run.txt)

3. `helm install luffy foo --render-subchart-notes`
    - Expected: Helm install output with the current chart's and subcharts' notes, as `--notes` is enabled by default in `helm install`, in order (first by depth, then alphabetically).
    - Output: [2.3.helm-install-luffy-foo-with-flag-render-subchart-notes.txt](https://github.com/user-attachments/files/22661553/2.3.helm-install-luffy-foo-with-flag-render-subchart-notes.txt)

4. `helm install luffy foo --dry-run --render-subchart-notes`
    - Expected: Helm install dry-run output with the current chart's and subcharts' notes, as `--notes` is enabled by default in `helm install`, in order (first by depth, then alphabetically).
    - Output: [2.4.helm-install-luffy-foo-with-flag-dry-run-and-render-subchart-notes.txt](https://github.com/user-attachments/files/22661555/2.4.helm-install-luffy-foo-with-flag-dry-run-and-render-subchart-notes.txt)

#### helm upgrade

1. `helm upgrade luffy foo`
    - Expected: Helm upgrade output with the current chart's notes, as `--notes` is enabled by default in `helm upgrade`.
    - Output: [3.1.helm-upgrade-luffy-foo.txt](https://github.com/user-attachments/files/22661559/3.1.helm-upgrade-luffy-foo.txt)

2. `helm upgrade luffy foo --dry-run`
    - Expected: Helm upgrade dry-run output with the current chart's notes, as `--notes` is enabled by default in `helm upgrade`.
    - Output: [3.2.helm-upgrade-luffy-foo-with-flag-dry-run.txt](https://github.com/user-attachments/files/22661562/3.2.helm-upgrade-luffy-foo-with-flag-dry-run.txt)

3. `helm upgrade luffy foo --render-subchart-notes`
    - Expected: Helm upgrade output with the current chart's and subcharts' notes, as `--notes` is enabled by default in `helm upgrade`, in order (first by depth, then alphabetically).
    - Output: [3.3.helm-upgrade-luffy-foo-with-flag-render-subchart-notes.txt](https://github.com/user-attachments/files/22661567/3.3.helm-upgrade-luffy-foo-with-flag-render-subchart-notes.txt)

4. `helm upgrade luffy foo --dry-run --render-subchart-notes`
    - Expected: Helm upgrade dry-run output with the current chart's and subcharts' notes, as `--notes` is enabled by default in `helm upgrade`, in order (first by depth, then alphabetically).
    - Output: [3.4.helm-upgrade-luffy-foo-with-flag-dry-run-and-render-subchart-notes.txt](https://github.com/user-attachments/files/22661570/3.4.helm-upgrade-luffy-foo-with-flag-dry-run-and-render-subchart-notes.txt)

### DIFF
--- a/pkg/cmd/testdata/output/template-help.txt
+++ b/pkg/cmd/testdata/output/template-help.txt
@@ -1,0 +1,86 @@
+
+Render chart templates locally and display the output.
+
+Any values that would normally be looked up or retrieved in-cluster will be
+faked locally. Additionally, none of the server-side testing of chart validity
+(e.g. whether an API is supported) is done.
+
+Usage:
+  helm template [NAME] [CHART] [flags]
+
+Flags:
+  -a, --api-versions strings                       Kubernetes api versions used for Capabilities.APIVersions (multiple can be specified)
+      --ca-file string                             verify certificates of HTTPS-enabled servers using this CA bundle
+      --cert-file string                           identify HTTPS client using this SSL certificate file
+      --create-namespace                           create the release namespace if not present
+      --dependency-update                          update dependencies if they are missing before installing the chart
+      --description string                         add a custom description
+      --devel                                      use development versions, too. Equivalent to version '>0.0.0-0'. If --version is set, this is ignored
+      --disable-openapi-validation                 if set, the installation process will not validate rendered templates against the Kubernetes OpenAPI Schema
+      --dry-run string[="unset"]                   simulates the operation either client-side or server-side. Must be either: "client", or "server". '--dry-run=client simulates the operation client-side only and avoids cluster connections. '--dry-run=server' simulates/validates the operation on the server, requiring cluster connectivity. (default "client")
+      --enable-dns                                 enable DNS lookups when rendering templates
+      --force-conflicts                            if set server-side apply will force changes against conflicts
+      --force-replace                              force resource updates by replacement
+  -g, --generate-name                              generate the name (and omit the NAME parameter)
+  -h, --help                                       help for template
+      --hide-notes                                 if set, do not show notes in install output. Does not affect presence in chart metadata
+      --include-crds                               include CRDs in the templated output
+      --insecure-skip-tls-verify                   skip tls certificate checks for the chart download
+      --is-upgrade                                 set .Release.IsUpgrade instead of .Release.IsInstall
+      --key-file string                            identify HTTPS client using this SSL key file
+      --keyring string                             location of public keys used for verification (default "test-gpg/pubring.gpg")
+      --kube-version string                        Kubernetes version used for Capabilities.KubeVersion
+  -l, --labels stringToString                      Labels that would be added to release metadata. Should be divided by comma. (default [])
+      --name-template string                       specify template used to name the release
+      --no-hooks                                   prevent hooks from running during install
+      --notes                                      if set, render the current chart's notes file
+      --output-dir string                          writes the executed templates to files in output-dir instead of stdout
+      --pass-credentials                           pass credentials to all domains
+      --password string                            chart repository password where to locate the requested chart
+      --plain-http                                 use insecure HTTP connections for the chart download
+      --post-renderer postRendererString           the name of a postrenderer type plugin to be used for post rendering. If it exists, the plugin will be used
+      --post-renderer-args postRendererArgsSlice   an argument to the post-renderer (can specify multiple) (default [])
+      --release-name                               use release name in the output-dir path.
+      --render-subchart-notes                      if set, render subchart notes along with the parent. Note: This will only work if --notes flag is enabled.
+      --replace                                    reuse the given name, only if that name is a deleted release which remains in the history. This is unsafe in production
+      --repo string                                chart repository url where to locate the requested chart
+      --rollback-on-failure                        if set, Helm will rollback (uninstall) the installation upon failure. The --wait flag will be default to "watcher" if --rollback-on-failure is set
+      --server-side                                object updates run in the server instead of the client (default true)
+      --set stringArray                            set values on the command line (can specify multiple or separate values with commas: key1=val1,key2=val2)
+      --set-file stringArray                       set values from respective files specified via the command line (can specify multiple or separate values with commas: key1=path1,key2=path2)
+      --set-json stringArray                       set JSON values on the command line (can specify multiple or separate values with commas: key1=jsonval1,key2=jsonval2 or using json format: {"key1": jsonval1, "key2": "jsonval2"})
+      --set-literal stringArray                    set a literal STRING value on the command line
+      --set-string stringArray                     set STRING values on the command line (can specify multiple or separate values with commas: key1=val1,key2=val2)
+  -s, --show-only stringArray                      only show manifests rendered from the given templates
+      --skip-crds                                  if set, no CRDs will be installed. By default, CRDs are installed if not already present
+      --skip-schema-validation                     if set, disables JSON schema validation
+      --skip-tests                                 skip tests from templated output
+      --take-ownership                             if set, install will ignore the check for helm annotations and take ownership of the existing resources
+      --timeout duration                           time to wait for any individual Kubernetes operation (like Jobs for hooks) (default 5m0s)
+      --username string                            chart repository username where to locate the requested chart
+  -f, --values strings                             specify values in a YAML file or a URL (can specify multiple)
+      --verify                                     verify the package before using it
+      --version string                             specify a version constraint for the chart version to use. This constraint can be a specific tag (e.g. 1.1.1) or it may reference a valid range (e.g. ^2.0.0). If this is not specified, the latest version is used
+      --wait WaitStrategy[=watcher]                if specified, will wait until all resources are in the expected state before marking the operation as successful. It will wait for as long as --timeout. Valid inputs are 'watcher' and 'legacy' (default hookOnly)
+      --wait-for-jobs                              if set and --wait enabled, will wait until all Jobs have been completed before marking the release as successful. It will wait for as long as --timeout
+
+Global Flags:
+      --burst-limit int                 client-side default throttling limit (default 100)
+      --color string                    use colored output (never, auto, always) (default "auto")
+      --colour string                   use colored output (never, auto, always) (default "auto")
+      --content-cache string            path to the directory containing cached content (e.g. charts) (default "test-content-cache")
+      --debug                           enable verbose output
+      --kube-apiserver string           the address and the port for the Kubernetes API server
+      --kube-as-group stringArray       group to impersonate for the operation, this flag can be repeated to specify multiple groups.
+      --kube-as-user string             username to impersonate for the operation
+      --kube-ca-file string             the certificate authority file for the Kubernetes API server connection
+      --kube-context string             name of the kubeconfig context to use
+      --kube-insecure-skip-tls-verify   if true, the Kubernetes API server's certificate will not be checked for validity. This will make your HTTPS connections insecure
+      --kube-tls-server-name string     server name to use for Kubernetes API server certificate validation. If it is not provided, the hostname used to contact the server is used
+      --kube-token string               bearer token used for authentication
+      --kubeconfig string               path to the kubeconfig file
+  -n, --namespace string                namespace scope for this request (default "test-namespace")
+      --qps float32                     queries per second used when communicating with the Kubernetes API, not including bursting
+      --registry-config string          path to the registry config file (default "test-registry-config.json")
+      --repository-cache string         path to the directory containing cached repository indexes (default "test-repository-cache-dir")
+      --repository-config string        path to the file containing repository names and URLs (default "test-repository-config.yaml")
--- a/pkg/cmd/testdata/testcharts/chart-with-notes/templates/rbac/role.yaml
+++ b/pkg/cmd/testdata/testcharts/chart-with-notes/templates/rbac/role.yaml
@@ -1,0 +1,8 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{ .Chart.Name }}-role
+rules:
+- apiGroups: [""]
+  resources: ["pods"]
+  verbs: ["get","list","watch"]
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -27,6 +27,7 @@ import (
 	"path"
 	"path/filepath"
 	"slices"
+	"sort"
 	"strings"
 	"sync"
 	"text/template"
@@ -219,7 +220,9 @@ func splitAndDeannotate(postrendered string) (map[string]string, error) {
 // TODO: As part of the refactor the duplicate code in cmd/helm/template.go should be removed
 //
 //	This code has to do with writing files to disk.
-func (cfg *Configuration) renderResources(ch *chart.Chart, values common.Values, releaseName, outputDir string, subNotes, useReleaseName, includeCrds bool, pr postrenderer.PostRenderer, interactWithRemote, enableDNS, hideSecret bool) ([]*release.Hook, *bytes.Buffer, string, error) {
+func (cfg *Configuration) renderResources(ch *chart.Chart, values common.Values, releaseName, outputDir string,
+	renderSubchartNotes, useReleaseName, includeCrds bool, pr postrenderer.PostRenderer, interactWithRemote, enableDNS,
+	hideSecret, includeNotesSource bool) ([]*release.Hook, *bytes.Buffer, string, error) {
 	var hs []*release.Hook
 	b := bytes.NewBuffer(nil)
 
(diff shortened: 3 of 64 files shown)

### CODE REVIEW

**Comment by benoittgt** [comment]
Hello @Bhargav-InfraCloud,  

Thanks for the PR! I tested it locally and it works as expected.  
Here are the commands I used for testing:  

```sh
./bin/helm template myrelease test-chart
./bin/helm template myrelease test-chart --notes
./bin/helm template myrelease test-chart --notes --render-subchart-notes
./bin/helm template myrelease test-chart --render-subchart-notes
./bin/helm install myrelease test-chart --dry-run
./bin/helm install myrelease test-chart --dry-run --render-subchart-notes
```

I'm just not sure this behavior https://github.com/helm/helm/issues/6901#issuecomment-2652091864 will work as _notes_  are not valid yaml. Piping the output directly into `kubectl apply` can cause issues.  

Example:  

```sh
./bin/helm template myrelease test-chart --notes > manifests.yaml
kubectl apply --dry-run=client -f manifests.yaml
```

Results:  

```text
serviceaccount/myrelease-subchart created (dry run)
serviceaccount/myrelease-test-chart created (dry run)
service/myrelease-subchart created (dry run)
service/myrelease-test-chart created (dry run)
deployment.apps/myrelease-subchart created (dry run)
deployment.apps/myrelease-test-chart created (dry run)
pod/myrelease-subchart-test-connection created (dry run)
pod/myrelease-test-chart-test-connection created (dry run)
error: error validating "manifests.yaml": error validating data: invalid object to validate; if you choose to ignore these errors, turn validation off with --validate=false
```

So to me, this is more of a developer responsibility :) :  
- either make their notes compatible,  
- or exclude them when generating manifests for `kubectl apply`, no `--*notes`  

Overall, the patch looks solid and thanks for the tests. Could you please fix the conflicts?

Bhargav-InfraCloud:
@benoittgt Great! Thanks for reviewing it! I've fixed the conflicts.

> I'm just not sure this behavior https://github.com/helm/helm/issues/6901#issuecomment-2652091864 will work as notes are not valid yaml. Piping the output directly into kubectl apply can cause issues.
> So to me, this is more of a developer responsibility :) :
> 
> - either make their notes compatible,
> - or exclude them when generating manifests for kubectl apply, no --*notes

Agreed. Without the `--notes` flag, the `helm template` output should be the same as before. Users can use it for piping output to `kubectl`.

Bhargav-InfraCloud:
FYI, rebased on the latest `main`.

jcogilvie:
Off the wall idea, but maybe a way to include notes.txt safely so downstream things don't get tripped up?

`--safe-notes` : prepend `#` before lines of rendered notes.txt?

Bhargav-InfraCloud:
> Off the wall idea, but maybe a way to include notes.txt safely so downstream things don't get tripped up?

Since the new behavior is gated behind a new flag `--notes`, it won’t affect existing users.

> `--safe-notes` : prepend `#` before lines of rendered notes.txt?

That's one option, and it would be a small change. But IMO, adding another layer isn't great for UX. Instead, I think users should follow this approach:

```bash
# Render all templates (including NOTES). Output is plain text, not YAML-parseable.
helm template luffy foo --notes
helm template luffy foo --notes --render-subchart-notes

# If YAML parsing is needed:
helm template luffy foo | yq

# To view only the NOTES output:
helm template luffy foo --notes --show-only templates/NOTES.txt
helm template luffy foo --notes --render-subchart-notes --show-only charts/bar/templates/NOTES.txt
```

WDYT?

Bhargav-InfraCloud:
@ all, FYI.
- Added testing steps and outputs for common use cases in the description above (unit tests already cover these).
- Rebased on the latest main.

outofrange:
Thanks for working on this - I'm mostly using Helm through kustomize or ArgoCD, so I often forget that NOTES.txt are actually a thing :)

I want to suggest writing notes to _stderr_ instead of _stdout_, or making it configurable;
I'd even argue that it's what users could expect as a default, since notes can be seen as _diagnostic output_ for which [POSIX suggests using _stderr_](https://pubs.opengroup.org/onlinepubs/009695399/).

Benefits, at least from my point of view:
- can still be piped to other tools expecting YAML
- `helm template ... --notes | kubectl apply -f -` would even display the notes in the terminal like `helm install` would (although I'm not sure why anyone would directly pipe to apply instead of doing `helm upgrade --install`)
- tools like kustomize could add support for setting `--notes`, also surfacing notes when doing something like `kustomize build --enable-helm . | kubectl apply -f -` (which is quite common, in contrast to directly applying helm templated manifests)
- tools like ArgoCD could simply capture output from _stderr_ for notes, instead of splitting _stdout_, or processing templates twice (which would introduce additional load during the often already quite expensive reconciliation loop)
- since _stdout_ would remain untouched, `--notes=true` could even be the default for `helm template` - either set as default by Helm itself, or configured by users through aliases, functions, env vars, etc... at least I wouldn't see a reason to **not** enable notes when they don't interfere with my other tooling, but might provide valuable information I might miss otherwise

Not sure how I would treat `helm install`; writing to _stderr_ would be justified in this case as well imho, but probably not worth a breaking change. If the output stream for notes would be configurable, `helm template` and `helm install` could use different defaults.

To summarize, my idea for `helm template` is
- write notes to _stderr_
  - if anyone can think of reasons why users would need to have it on stdout: make the output stream configurable through flag, `~/.config/helm` and/or env var
- make it possible to **always** show notes by configuring it through `~/.config/helm` and/or env var

Bhargav-InfraCloud:
@outofrange Nice summary! The point about [POSIX suggests using stderr](https://pubs.opengroup.org/onlinepubs/009695399/) is interesting, but I’m not sure how that helps in this case. I’m trying to look at it from a user’s perspective — why would someone include notes, redirect them to `stderr`, and then parse only the `stdout` through a pipe? Wouldn’t it be simpler to do something like this instead:

* **When you want to parse the output as YAML** or pipe it into `kubectl` commands, skip the notes flags.

  > **Note:** `--notes` is `false` by default in `helm template`.

  ```bash
  helm template luffy foo | yq
  helm template luffy foo | kubectl get -f -
  ```

* **When you only need the notes** (without manifests or hooks):

  ```bash
  helm template luffy foo --notes --show-only templates/NOTES.txt
  helm template luffy foo --notes --render-subchart-notes \
    --show-only charts/bar/templates/NOTES.txt \
    --show-only charts/baz/templates/NOTES.txt
  ```

* **When you want all templates** and don’t need YAML parsing or `kubectl` piping:

  ```bash
  helm template luffy foo --notes
  helm template luffy foo --notes --render-subchart-notes
  ```

---

**PS:** You can test the above commands by following the steps outlined in the **#testing** section of the description.

Bhargav-InfraCloud:
FYI, rebased on the latest `main`.

davinkevin:
We would love to see this PR merged as we struggle to correctly test the `NOTES.txt`.

pkg/cmd/testdata/output/template-help.txt
```diff
@@ -0,0 +1,86 @@
+
+Render chart templates locally and display the output.
```
scottrigby: same concern. Remove full help output from testdata

pkg/cmd/testdata/testcharts/chart-with-notes/templates/rbac/role.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
scottrigby: we can simplify all of these new test charts dramatically, to only include what we're actually testing, NOTES.txt in the parent and subcharts. Keep as minimal as possible to test out this specific functionality.

scottrigby:
@Bhargav-InfraCloud I'm closing this PR as NOTES output should not be part of helm template command output. I agree with what @benoittgt initially said.

I explain in detain on this comment https://github.com/helm/helm/issues/6901#issuecomment-3786612928

@davinkevin also see that comment. You can test `NOTES.txt` output using `helm install` or `helm upgrade` with `--dry-run` options, and optionallly the same `--render-subchart-notes` flag that works today.

davinkevin:
@scottrigby Thank you for this answer. 

Unfortunately, the `upgrade --dry-run` has a major issue, it requires a cluster to be executed and that isn't convenient when you want to run that in a simple CI with just `helm` binary. 

If we could disable the server connection, that would be perfect 😇

benoittgt:
@davinkevin there is [another issue](https://github.com/helm/helm/issues/13144) related to the `--dry-run`. I can look at the upgrade command. It has been improved for others in v4. See: https://github.com/helm/helm/issues/13144#issuecomment-3785562013

scottrigby:
> @scottrigby Thank you for this answer.
> 
> Unfortunately, the `upgrade --dry-run` has a major issue, it requires a cluster to be executed and that isn't convenient when you want to run that in a simple CI with just `helm` binary.
> 
> If we could disable the server connection, that would be perfect 😇

hi @davinkevin, @benoittgt is correct - this is exactly what `--dry-run=client` now allows 😄

Issue #31103, opened 2025-07-31
cli.New() forget to initialize  Kubeconfig field


LinPr:
### What happened?

In pkg/cli/environment.go
https://github.com/helm/helm/blob/99d42acf6090b2e585987de441db555f7a5d2f7c/pkg/cli/environment.go#L122

full codes here
```go
func New() *EnvSettings {
	env := &EnvSettings{
		namespace:                 os.Getenv("HELM_NAMESPACE"),
		MaxHistory:                envIntOr("HELM_MAX_HISTORY", defaultMaxHistory),
		KubeContext:               os.Getenv("HELM_KUBECONTEXT"),
		KubeToken:                 os.Getenv("HELM_KUBETOKEN"),
		KubeAsUser:                os.Getenv("HELM_KUBEASUSER"),
		KubeAsGroups:              envCSV("HELM_KUBEASGROUPS"),
		KubeAPIServer:             os.Getenv("HELM_KUBEAPISERVER"),
		KubeCaFile:                os.Getenv("HELM_KUBECAFILE"),
		KubeTLSServerName:         os.Getenv("HELM_KUBETLS_SERVER_NAME"),
		KubeInsecureSkipTLSVerify: envBoolOr("HELM_KUBEINSECURE_SKIP_TLS_VERIFY", false),
		PluginsDirectory:          envOr("HELM_PLUGINS", helmpath.DataPath("plugins")),
		RegistryConfig:            envOr("HELM_REGISTRY_CONFIG", helmpath.ConfigPath("registry/config.json")),
		RepositoryConfig:          envOr("HELM_REPOSITORY_CONFIG", helmpath.ConfigPath("repositories.yaml")),
		RepositoryCache:           envOr("HELM_REPOSITORY_CACHE", helmpath.CachePath("repository")),
		BurstLimit:                envIntOr("HELM_BURST_LIMIT", defaultBurstLimit),
		QPS:                       envFloat32Or("HELM_QPS", defaultQPS),
	}
	env.Debug, _ = strconv.ParseBool(os.Getenv("HELM_DEBUG"))

	// bind to kubernetes config flags
	config := &genericclioptions.ConfigFlags{
		Namespace:        &env.namespace,
		Context:          &env.KubeContext,
		BearerToken:      &env.KubeToken,
		APIServer:        &env.KubeAPIServer,
		CAFile:           &env.KubeCaFile,
		KubeConfig:       &env.KubeConfig,
		Impersonate:      &env.KubeAsUser,
[...]

PR #31104, opened 2025-07-31
fix: assign KUBECONFIG environment variable value to env.Kubeconfig

LinPr:
when I imported this pacakge, I found it can't get the kubeconfig from the KUBECONFIG env  automatically，so I read the source code. The field supposed to be assigned

To get the kubeconfig from KUBECONFIG environment variable automatically
Refer to this issue: #31103 

**If applicable**:

### DIFF
--- a/pkg/cli/environment.go
+++ b/pkg/cli/environment.go
@@ -95,6 +95,7 @@ func New() *EnvSettings {
 	env := &EnvSettings{
 		namespace:                 os.Getenv("HELM_NAMESPACE"),
 		MaxHistory:                envIntOr("HELM_MAX_HISTORY", defaultMaxHistory),
+		KubeConfig:                os.Getenv("KUBECONFIG"),
 		KubeContext:               os.Getenv("HELM_KUBECONTEXT"),
 		KubeToken:                 os.Getenv("HELM_KUBETOKEN"),
 		KubeAsUser:                os.Getenv("HELM_KUBEASUSER"),


pkg/cli/environment.go
```diff
@@ -109,6 +109,7 @@ func New() *EnvSettings {
 		RepositoryCache:           envOr("HELM_REPOSITORY_CACHE", helmpath.CachePath("repository")),
 		BurstLimit:                envIntOr("HELM_BURST_LIMIT", defaultBurstLimit),
 		QPS:                       envFloat32Or("HELM_QPS", defaultQPS),
+		KubeConfig:                os.Getenv("KUBECONFIG"),
```
TerryHowe: I think it would make more sense to place this above KubeContext.

TerryHowe (changes requested):

pkg/cli/environment.go
LinPr: ok I will move it to a better place

TerryHowe: I looked at the code and I see no reason this isn't done, but there may be a reason I don't understand. Need some feedback from others.

LinPr: yes, thanks for your  meticulous work, and also looing  forward to feedback from others.

LinPr:
Any other maintainer review this?🤔

pkg/cli/environment.go
LinPr: > I looked at the code and I see no reason this isn't done, but there may be a reason I don't understand. Need some feedback from others.

No other maintainers reviewing this since then🤥🤥 what about just merge this PR, I think it will be great

gjenkins8 (approved):

TerryHowe (approved):
/lgtm

PR #31108, opened 2025-08-02
Add debug logging for helm pull command

S0umith29:
Hi :)

This pull request closes #31098, where the --debug flag was being ignored when running helm pull. The changes ensure that HTTP-level debug logs (e.g., URLs called, redirects, proxy usage) are printed when the --debug flag is enabled.

### Changes Made

- **pkg/action/pull.go:** Added ```Debug``` to the ```HTTPGetter``` configuration to pass environment settings for debugging.
- **pkg/getter/httpgetter.go:** Integrated ```Debug``` to configure the HTTP client with a debug transport when enabled.
- **pkg/getter/getter.go:** Updated the ```options``` struct to include the Debug field.
- **pkg/downloader/chart_downloader.go:** Added support for debug transport to log HTTP requests and responses during chart downloads.

### Testing

**MP4 Video attached**

- I have tested this with the strimzi repository after adding the repo.
- Added the Strimzi repository: `helm repo add strimzi https://strimzi.io/charts`
- Ran `helm pull strimzi/strimzi-kafka-operator --version 0.46.1 --debug`
- Verified that HTTP-level debug logs are now printed:
  
  https://github.com/user-attachments/assets/ca1a9e95-37cc-4647-99d1-93e313b6f2d9

### DIFF
--- a/pkg/downloader/chart_downloader.go
+++ b/pkg/downloader/chart_downloader.go
@@ -21,6 +21,8 @@ import (
 	"io"
 	"io/fs"
 	"net/url"
+	"net/http"
+	"net/http/httputil"
 	"os"
 	"path/filepath"
 	"strings"
@@ -69,9 +71,42 @@ type ChartDownloader struct {
 	Getters getter.Providers
 	// Options provide parameters to be passed along to the Getter being initialized.
 	Options          []getter.Option
+        Debug bool	//Added to capture the --debug flag
 	RegistryClient   *registry.Client
 	RepositoryConfig string
 	RepositoryCache  string
+}
+
+type debugTransport struct {
+	*http.Transport
+	out       io.Writer
+}
+
+func (t *debugTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+
+	fmt.Fprintf(t.out, "DEBUG: HTTP request to %s\n", req.URL.String())
+	// Log the request
+	reqDump, err := httputil.DumpRequestOut(req, false)
+	if err == nil {
+		fmt.Fprintf(t.out, "%s\n", reqDump)
+	}
+	// Perform the request
+	resp, err := t.Transport.RoundTrip(req)
+	if err != nil {
+		fmt.Fprintf(t.out, "HTTP request failed: %v\n", err)
+		return nil, err
+	}
+	// Log the response
+	respDump, err := httputil.DumpResponse(resp, false)
+	if err == nil {
+		fmt.Fprintf(t.out, "HTTP Response: \n%s\n", respDump)
+	}
+	if resp.StatusCode >= 300 && resp.StatusCode < 400 {
+            	location, _ := resp.Header["Location"]
+          	fmt.Fprintf(t.out, "DEBUG: Redirect to: %v\n", location)
+        }
+
+	return resp, err
 }
 
 // DownloadTo retrieves a chart. Depending on the settings, it may also download a provenance file.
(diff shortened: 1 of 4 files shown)


robertsirc:
Before we review this could you address the broken workflows?

S0umith29:
> Before we review this could you address the broken workflows?

Hi Robert :)
Can you please check now?

pkg/downloader/chart_downloader.go
```diff
@@ -88,18 +128,49 @@ type ChartDownloader struct {
 func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *provenance.Verification, error) {
 	u, err := c.ResolveChartVersion(ref, version)
 	if err != nil {
+		if c.Debug {
+			fmt.Fprintf(c.Out, "DEBUG: Failed to resolve chart version: %v\n", err)
+		}
```
mattfarina: For logging we are using `slog`. In place of if calls with checks for Debug, `slog.Debug()` should be used instead. See https://pkg.go.dev/log/slog#Debug

mattfarina (changes requested):
Thanks for your interest in adding logging here. This is something that would be useful to have. There are two changes I'm hoping you can make.

1. Please use `slog` for the logs
2. This would work for some getters but not all of them. We need a solution that works for all of them, including plugins. Can you pass a debug flag to the getters. For plugins it can be an environment variable. Then have the getter handle the logging so we can have it everywhere.

pkg/downloader/chart_downloader.go
S0umith29: Used ```slog``` in place of the conditional ```if``` statements.

S0umith29:
Hi Matt,

Thanks for the feedback!

- I've updated the implementation to use ```slog``` for logging, as suggested.
- The ```debug``` flag is now passed to all getters via the existing options mechanism. For plugin-based getters, the ```HELM_DEBUG``` environment variable is used to enable consistent debug behavior.
- All logging is handled within the getter itself, ensuring uniform logging across built-in and plugin-based getters.

Let me know if there's anything else you'd like adjusted!

pkg/downloader/chart_downloader.go
```diff
@@ -68,12 +71,51 @@ type ChartDownloader struct {
...
+	Options []getter.Option
+	// Debug            bool //Added to capture the --debug flag
 	RegistryClient   *registry.Client
 	RepositoryConfig string
 	RepositoryCache  string
 }
 
+type debugTransport struct {
+	*http.Transport
+	out io.Writer
```
TerryHowe, line 83:
```suggestion
```

pkg/downloader/chart_downloader.go
```diff
@@ -86,20 +128,43 @@ type ChartDownloader struct {
 // Returns a string path to the location where the file was downloaded and a verification
 // (if provenance was verified), or an error if something bad happened.
 func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *provenance.Verification, error) {
+	debug := os.Getenv("HELM_DEBUG") == "true"
+
+	// If debug is enabled, wrap the getter's HTTP client with a debug transport
+	if debug {
+		dt := &debugTransport{
+			Transport: http.DefaultTransport.(*http.Transport).Clone(),
+			out:       c.Out,
```
TerryHowe, line 137:
```suggestion
```

pkg/downloader/chart_downloader.go
```diff
@@ -117,6 +182,21 @@ func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *proven
 	// If provenance is requested, verify it.
 	ver := &provenance.Verification{}
 	if c.Verify > VerifyNever {
+		// If debug is enabled, use the same debug transport for provenance file
+		if debug {
+			dt := &debugTransport{
+				Transport: http.DefaultTransport.(*http.Transport).Clone(),
+				out:       c.Out,
```
TerryHowe, line 189:
```suggestion
```

pkg/downloader/chart_downloader.go
```diff
@@ -117,6 +182,21 @@ func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *proven
...
+			}
+			dt.DisableKeepAlives = true
+			c.Options = append(c.Options, getter.WithClient(&http.Client{
```
TerryHowe, line 192: I'm not super familiar with this code, can the user provide a client here? Do you need to possibly work with a client specified by the user?

pkg/getter/getter.go
```diff
@@ -191,6 +194,18 @@ const (
 
 var defaultOptions = []Option{WithTimeout(time.Second * DefaultHTTPTimeout)}
 
+func init() {
```
TerryHowe, line 197: I think this could be handled by pkg/cmd/root.go or an SDK user

pkg/getter/getter.go
```diff
@@ -47,6 +49,7 @@ type options struct {
 	registryClient        *registry.Client
 	timeout               time.Duration
 	transport             *http.Transport
+	client                *http.Client
```
TerryHowe, line 52: This name should be distinguished from the registry client

TerryHowe:
Conflicts. 

Also, it would be convenient if the repository pull used the same round tripper as the registry pull if possible.

PR #31116, opened 2025-08-07
chore: check if go modules are tidy before build

banjoh:

Fail PR build if `go mod tidy` has not been run

The build will fail like https://github.com/helm/helm/actions/runs/16803496134/job/47590088636?pr=31116

**If applicable**:


Makefile
```diff
@@ -246,3 +246,7 @@ info:
 	 @echo "Git Tag:           ${GIT_TAG}"
 	 @echo "Git Commit:        ${GIT_COMMIT}"
 	 @echo "Git Tree State:    ${GIT_DIRTY}"
+
```
TerryHowe, line 246: Do we want to add this as a dependency of the `build` target?

banjoh: I guess we could.

Makefile
```diff
@@ -75,7 +75,7 @@ all: build
 #  build
 
 .PHONY: build
-build: $(BINDIR)/$(BINNAME)
+build: $(BINDIR)/$(BINNAME) info
```
TerryHowe:
```suggestion
build: $(BINDIR)/$(BINNAME) tidy
```

TerryHowe (changes requested):

Makefile
banjoh: Lol. This is what you meant :D

TerryHowe (approved):
/lgtm

.github/workflows/build-test.yml
```diff
@@ -28,6 +28,8 @@ jobs:
           check-latest: true
       - name: Test source headers are present
         run: make test-source-headers
+      - name: Check if go modules need to be tidied
+        run: go mod tidy -diff
```
scottrigby, line 32: since you also added `tidy` on the `build` target, it will automatically run before build when we call the `Test build` step below. so we don't need this here now, right?
```
build: $(BINDIR)/$(BINNAME) tidy
```

scottrigby (commented):
the note below, and also needs rebase again

.github/workflows/build-test.yml
banjoh: `go mod tidy -diff` checks if the user ran `go mod tidy` before creating the PR.

`build: $(BINDIR)/$(BINNAME) tidy` is a shortcut to ensure users don't forget to tidy up.

scottrigby: oh thank you, i read that too fast. this lgtm :+1: just needs rebase

scottrigby (dismissed):
lgtm once merge conflict is resolved

TerryHowe (approved):
/lgtm

scottrigby (approved):

PR #31126, opened 2025-08-08
fix(transport): leverage same tls config

### LINKED ISSUE #31111: Helm repo add is not leveraging rootStore
### What happened?

When adding a repo leveraging a proxy:
```bash
HTTPS_PROXY=https://localhost:8080 helm repo add test https://newrelic.github.io/helm-charts
```
I get:
```
Error: looks like "https://newrelic.github.io/helm-charts" is not a valid chart repository or cannot be reached: Get "https://newrelic.github.io/helm-charts/index.yaml": malformed HTTP status code "protocol"
```

However I am able to connect via proxy with a simple curl since certs are loaded in the system:
```
HTTPS_PROXY=https://localhost:8080  curl  "https://newrelic.github.io/helm-charts/index.yaml"
```

To fix that I have to pass manually the ca
```
HTTPS_PROXY=https://localhost:8080 helm repo add test https://newrelic.github.io/helm-charts  --ca-file ~/.mitmproxy/mitmproxy-ca.pem
```

 - it should work leveraging system certs. 
 - the log should provide some info regarding what is going on

run a proxy
```
docker run --rm -v  ~/.mitmproxy:/home/mitmproxy/.mitmproxy -it -p 8081:8081 -p 8080:8080 mitmproxy/mitmproxy:12 mitmweb --web-host 0.0.0.0
```
add certs to your local env (on mac)
```
sudo security add-trusted-cert -d -p ssl -p basic -k /Library/Keychains/System.keychain ~/.mitmproxy/mitmproxy-ca-cert.pem
```

### Helm version

version.BuildInfo{Version:"v3.18.4", GitCommit:"d80839cf37d860c8aa9a0503fe463278f26cd5e2", GitTreeState:"clean", GoVersion:"go1.24.5"}

### Kubernetes version

does not apply

paologallinaharbur:
it fixes https://github.com/helm/helm/issues/31111, but the reason is not obvious. 

Closes: https://github.com/helm/helm/issues/31111
---

Before:
 - if `NewTLSConfig` was called: TLSClientConfig was created empty and then modified with the needed values
 - if `NewTLSConfig` was not called: TLSClientConfig was `nil`. 
 
Keep in mind that:
```
	// TLSClientConfig specifies the TLS configuration to use with
	// tls.Client.
	// If nil, the default configuration is used.
	// If non-nil, HTTP/2 support may not be enabled by default.
```
Therefore, we were using de facto a different `TLSClientConfig` depending whenever those inputs were passed and NewTLSConfig was called or not. In one case the default, in the other an empty one.

---

Now:
 - if `NewTLSConfig` is not called tlsConf = &tls.Config{}

This is the behaviour already implemented [elsewhere in the code!](https://github.com/helm/helm/blob/main/pkg/registry/client.go#L296-L296)
 
This change fixes my issue. I noticed a different behaviour whenever I was passing one of those flags.

```
HTTPS_PROXY=https://localhost:8080 go run ./... repo add test4 https://newrelic.github.io/helm-charts 
"test4" has been added to your repositories
```

---

**If applicable**:

### DIFF
--- a/pkg/getter/httpgetter.go
+++ b/pkg/getter/httpgetter.go
@@ -122,6 +122,9 @@ func (g *HTTPGetter) httpClient() (*http.Client, error) {
 		g.transport = &http.Transport{
 			DisableCompression: true,
 			Proxy:              http.ProxyFromEnvironment,
+			// Being nil would cause the tls.Config default to be used
+			// "NewTLSConfig" modifies an empty TLS config, not the default one
+			TLSClientConfig: &tls.Config{},
 		}
 	})
 
--- a/pkg/getter/ocigetter.go
+++ b/pkg/getter/ocigetter.go
@@ -17,6 +17,7 @@ package getter
 
 import (
 	"bytes"
+	"crypto/tls"
 	"fmt"
 	"net"
 	"net/http"
@@ -124,6 +125,9 @@ func (g *OCIGetter) newRegistryClient() (*registry.Client, error) {
 			TLSHandshakeTimeout:   10 * time.Second,
 			ExpectContinueTimeout: 1 * time.Second,
 			Proxy:                 http.ProxyFromEnvironment,
+			// Being nil would cause the tls.Config default to be used
+			// "NewTLSConfig" modifies an empty TLS config, not the default one
+			TLSClientConfig: &tls.Config{},
 		}
 	})
 


paologallinaharbur:
This is to avoid having this weird behaviour depending on args passed

Leveraging a proxy:
```
$ export HTTPS_PROXY=https://localhost:8080

$ go run ./... repo add test https://newrelic.github.io/helm-charts                
Error: looks like "https://newrelic.github.io/helm-charts" is not a valid chart repository or cannot be reached: Get "https://newrelic.github.io/helm-charts/index.yaml": malformed HTTP status code "protocol"
exit status 1

$ touch test     # that obviously is completely empty
$ go run ./... repo add test https://newrelic.github.io/helm-charts  --ca-file=test                           
"test" has been added to your repositories
```

joejulian:
Doesn't this _just_ disable h2 which, if your proxy only supports http/1.1 would explain why this fixes it. The nil setting goes with the defaults which only supports h2.

Have you tried this?
```
&tls.Config{
	NextProtos: []string{"h2", "http/1.1"},
}
```
with
```
_ = http2.ConfigureTransport(tr)
```

TerryHowe:
Please include

Closes: https://github.com/helm/helm/issues/31111

in your description so the issues is closed when/if this merges. I don't think that is picked up from a comment like this one.

paologallinaharbur:
> Doesn't this just disable h2 which, if your proxy only supports http/1.1 would explain why this fixes it.

Yes, having `NextProtos` set as `[]string{"h2", "http/1.1"}` does not work, but it works if changed to `[]string{"http/1.1"}` While this is possibly related to https://github.com/golang/go/issues/26479, I did not want just to "patch" my use-case. 

---

The goal of this PR was to ensure consistency and clarity in how TLSClientConfig is created. The core issue of the code before the change, in my opinion, is that the behavior changes in a non-transparent way: while `TLSClientConfig` defaults to `nil`, calling `NewTLSConfig` results in an "empty" `TLSClientConfig` with a few flags modified (see [here](https://github.com/helm/helm/blob/main/internal/tlsutil/tls.go#L99-L99)). **Sadly having it "nil" or "empty" is not the same and the behaviour is slightly different as the docs mention**.

In the example pasted before, passing an empty `ca-file` changes in an expected way the `TLSClientConfig`, causing as a side-effect a different protocol to be used:
```
$ export HTTPS_PROXY=https://localhost:8080

$ go run ./... repo add test https://newrelic.github.io/helm-charts                
Error: [...]: malformed HTTP status code "protocol"
exit status 1

$ touch test     # that obviously is completely empty
$ go run ./... repo add test https://newrelic.github.io/helm-charts  --ca-file=test                           
"test" has been added to your repositories
```

That is why I believe that in order to have no discrepancy it is better to always define it with the desired config. This will help to avoid having a different behaviour depending on the side-effects whenever some args are passed.

mattfarina (approved):
To recap, Helm has 2 paths for http. If you specify certs, h2 is not used because a custom `tls.Config` is created. If certs are not specified, a default tlsconfig is used that specifies h2. For reference, you can find this behavior [in the Go code here.](https://github.com/golang/go/blob/084c0f849467d5e45b7242cda3dd957352e86b8f/src/net/http/transport.go#L456-L481)

This change uses a custom `tls.Config` through the whole process so things are consistent. This has the affect of disabling http2. This allows some proxies to work where go, the proxy, and h2 don't work properly together.

Note, [the default Transport specifies to attempt http2](https://github.com/golang/go/blob/1718828c81f2bad44b51b13fb4b34540a9c2c096/src/net/http/transport.go#L51).

When dealing with a single file at a time, h2 doesn't provide a major performance benefit. It can sometimes be slower or roughly the same as h1.1.
When dealing with multiple files, like fetching index files, there is usually a performance improvement to use h1. In my testing with and without this change, updating the index files for 12 repos was consistently faster with the patch.

I'm ok with this change. It fixes a proxy issue, makes things more consistent process wise, and in my testing was a little faster.

PR #31128, opened 2025-08-11
fix(downloader): prevent concurrent file access errors on Windows

### LINKED ISSUE #11096: Concurrency issues pulling charts in Helm 3.7.2
Output of `helm version`: 
```
version.BuildInfo{Version:"v3.7.2", GitCommit:"663a896f4a815053445eec4153677ddc24a0a361", GitTreeState:"clean", GoVersion:"go1.16.10"}
```

Output of `kubectl version`: 
```
Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.5", GitCommit:"6b1d87acf3c8253c123756b9e61dac642678305f", GitTreeState:"clean", BuildDate:"2021-03-18T01:10:43Z", GoVersion:"go1.15.8", Compiler:"gc", Platform:"windows/amd64"}
Server Version: version.Info{Major:"1", Minor:"21+", GitVersion:"v1.21.12-eks-a64ea69", GitCommit:"d4336843ba36120e9ed1491fddff5f2fec33eb77", GitTreeState:"clean", BuildDate:"2022-05-12T18:29:27Z", GoVersion:"go1.16.15", Compiler:"gc", Platform:"linux/amd64"}
```

Cloud Provider/Platform (AKS, GKE, Minikube etc.):  AWS EKS

---

I have encountered an issue, superficially similar to to #10735, although I think the exact cause is different. In my case, the issue occurs when a command such as `helm template` attempts to pull the `.tgz` of a chart not already in the cache; if multiple such commands are being run concurrently, which involve the same chart and chart version, then, at least on Windows, they may attempt to place the `.tgz` file into the cache simultaneously, leading to errors.

Given the concurrency-related nature of this issue, it's not reproducible _reliably_, but I'm able to trigger it fairly often using a Powershell script like this:

```powershell
rm -r $env:LocalAppData\Temp\helm\repository\fluent-bit-0.20.3.tgz
1,2,3,4,5,6 | ForEach-Object -Parallel { helm --kube-context test template fluent-bit fluent/fluent-bit --values values.yaml --version 0.20.3 --debug --output-dir ./tmp/$_ } -ThrottleLimit 6
rm -r ./tmp
```

This will sometimes lead to failures, with output like

```
install.go:178: [debug] Original chart version: "0.20.3"
install.go:178: [debug] Original chart version: "0.20.3"
install.go:178: [debug] Original chart version: "0.20.3"
install.go:178: [debug] Original chart version: "0.20.3"
install.go:178: [debug] Original chart version: "0.20.3"
install.go:178: [debug] Original chart version: "0.20.3"
install.go:199: [debug] CHART PATH: C:\Users\me\AppData\Local\Temp\helm\repository\fluent-bit-0.20.3.tgz
install.go:199: [debug] CHART PATH: C:\Users\me\AppData\Local\Temp\helm\repository\fluent-bit-0.20.3.tgz
install.go:199: [debug] CHART PATH: C:\Users\me\AppData\Local\Temp\helm\repository\fluent-bit-0.20.3.tgz
Error: link error: cannot rename C:\Users\me\AppData\Local\Temp\helm\repository\fluent-bit-0.20.3.tgz456375323 to C:\Users\me\AppData\Local\Temp\helm\repository\fluent-bit-0.20.3.tgz: rename C:\Users\me\AppData\Local\Temp\helm\repository\fluent-bit-0.20.3.tgz456375323 C:\Users\me\AppData\Local\Temp\helm\repository\fluent-bit-0.20.3.tgz: Access is denied.
helm.go:88: [debug] rename C:\Users\me\AppData\Local\Temp\helm\repository\fluent-bit-0.20.3.tgz456375323 C:\Users\me\AppData\Local\Temp\helm\repository\fluent-bit-0.20.3.tgz: Access is denied.
link error: cannot rename C:\Users\me\AppData\Local\Temp\helm\repository\fluent-bit-0.20.3.tgz456375323 to C:\Users\me\AppData\Local\Temp\helm\repository\fluent-bit-0.20.3.tgz
helm.sh/helm/v3/internal/third_party/dep/fs.renameFallback
        helm.sh/helm/v3/internal/third_party/dep/fs/rename_windows.go:64
helm.sh/helm/v3/internal/third_party/dep/fs.RenameWithFallback
        helm.sh/helm/v3/internal/third_party/dep/fs/fs.go:62
helm.sh/helm/v3/internal/fileutil.AtomicWriteFile
        helm.sh/helm/v3/internal/fileutil/fileutil.go:50
helm.sh/helm/v3/pkg/downloader.(*ChartDownloader).DownloadTo
        helm.sh/helm/v3/pkg/downloader/chart_downloader.go:110
helm.sh/helm/v3/pkg/action.(*ChartPathOptions).LocateChart
        helm.sh/helm/v3/pkg/action/install.go:753
main.runInstall
        helm.sh/helm/v3/cmd/helm/install.go:194
main.newTemplateCmd.func2
        helm.sh/helm/v3/cmd/helm/template.go:82
        github.com/spf13/cobra@v1.2.1/command.go:856
        github.com/spf13/cobra@v1.2.1/command.go:974
main.main
        helm.sh/helm/v3/cmd/helm/helm.go:87
runtime.main
        runtime/proc.go:225
runtime.goexit
        runtime/asm_amd64.s:1371
install.go:199: [debug] CHART PATH: C:\Users\me\AppData\Local\Temp\helm\repository\fluent-bit-0.20.3.tgz
install.go:199: [debug] CHART PATH: C:\Users\me\AppData\Local\Temp\helm\repository\fluent-bit-0.20.3.tgz
```

orgads:
When multiple processes try to download the same chart version concurrently (e.g., via Terraform), they can race to write the destination file. On Windows, this results in 'Access Denied' errors because the file cannot be renamed while another process has a handle to the destination.

This commit introduces 'PlatformAtomicWriteFile' to the fileutil package. On Unix-like systems, it simply delegates to AtomicWriteFile, maintaining existing behavior. On Windows, it coordinates writes using a lock file (.lock). It acquires the lock, then performs the atomic write.

Crucially, this implementation ensures that existing files are overwritten (rather than skipped). This ensures that if a chart is republished with the same version, the local cache is correctly updated, preventing stale data issues.

Fixes #31633

Signed-off-by: Orgad Shaneh <orgad.shaneh@audiocodes.com>

**If applicable**:

### DIFF
--- a/internal/fileutil/fileutil_unix.go
+++ b/internal/fileutil/fileutil_unix.go
@@ -1,0 +1,32 @@
+//go:build !windows
+
+/*
+Copyright The Helm Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package fileutil
+
+import (
+	"io"
+	"os"
+)
+
+// PlatformAtomicWriteFile atomically writes a file to disk.
+//
+// On non-Windows platforms we don't need extra coordination, so this simply
+// delegates to AtomicWriteFile to preserve the existing overwrite behaviour.
+func PlatformAtomicWriteFile(filename string, reader io.Reader, mode os.FileMode) error {
+	return AtomicWriteFile(filename, reader, mode)
+}
--- a/pkg/downloader/chart_downloader.go
+++ b/pkg/downloader/chart_downloader.go
@@ -156,7 +156,11 @@ func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *proven
 	}
 
 	destfile := filepath.Join(dest, name)
-	if err := fileutil.AtomicWriteFile(destfile, data, 0644); err != nil {
+
+	// Use PlatformAtomicWriteFile to handle platform-specific concurrency concerns
+	// (Windows requires locking to avoid "Access Denied" errors when multiple
+	// processes write the same file)
+	if err := fileutil.PlatformAtomicWriteFile(destfile, data, 0644); err != nil {
 		return destfile, nil, err
 	}
 
@@ -186,7 +190,9 @@ func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *proven
 			}
 		}
 		provfile := destfile + ".prov"
-		if err := fileutil.AtomicWriteFile(provfile, body, 0644); err != nil {
+
+		// Use PlatformAtomicWriteFile for the provenance file as well
+		if err := fileutil.PlatformAtomicWriteFile(provfile, body, 0644); err != nil {
 			return destfile, nil, err
 		}
 
--- a/internal/fileutil/fileutil_test.go
+++ b/internal/fileutil/fileutil_test.go
@@ -119,3 +119,29 @@ func TestAtomicWriteFile_LargeContent(t *testing.T) {
 		t.Fatalf("expected large content to match, got different length: %d vs %d", len(largeContent), len(got))
 	}
 }
+
+// TestPlatformAtomicWriteFile_OverwritesExisting verifies that the platform
+// helper replaces existing files instead of silently skipping them.
+func TestPlatformAtomicWriteFile_OverwritesExisting(t *testing.T) {
+	dir := t.TempDir()
+	path := filepath.Join(dir, "overwrite_test")
+
+	first := bytes.NewReader([]byte("first"))
+	if err := PlatformAtomicWriteFile(path, first, 0644); err != nil {
+		t.Fatalf("first write failed: %v", err)
+	}
+
+	second := bytes.NewReader([]byte("second"))
+	if err := PlatformAtomicWriteFile(path, second, 0644); err != nil {
+		t.Fatalf("second write failed: %v", err)
+	}
+
+	contents, err := os.ReadFile(path)
+	if err != nil {
+		t.Fatalf("failed reading result: %v", err)
+	}
+
+	if string(contents) != "second" {
+		t.Fatalf("expected file to be overwritten, got %q", string(contents))
+	}
+}
--- a/internal/fileutil/fileutil_windows.go
+++ b/internal/fileutil/fileutil_windows.go
@@ -1,0 +1,54 @@
+//go:build windows
+
+/*
+Copyright The Helm Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package fileutil
+
+import (
+	"io"
+	"os"
+
+	"github.com/gofrs/flock"
+)
+
+// PlatformAtomicWriteFile atomically writes a file to disk with file locking to
+// prevent concurrent writes. This is particularly useful on Windows where
+// concurrent writes to the same file can cause "Access Denied" errors.
+//
+// The function acquires a lock on the target file and performs an atomic write,
+// preserving the existing behaviour of overwriting any previous content once
+// the lock is obtained.
+func PlatformAtomicWriteFile(filename string, reader io.Reader, mode os.FileMode) error {
+	// Use a separate lock file to coordinate access between processes
+	// We cannot lock the target file directly as it would prevent the atomic rename
+	lockFileName := filename + ".lock"
+	fileLock := flock.New(lockFileName)
+
+	// Lock() ensures serialized access - if another process is writing, this will wait
+	if err := fileLock.Lock(); err != nil {
+		return err
+	}
+	defer func() {
+		fileLock.Unlock()
+		// Clean up the lock file
+		// Ignore errors as the file might not exist or be in use by another process
+		os.Remove(lockFileName)
+	}()
+
+	// Perform the atomic write while holding the lock
+	return AtomicWriteFile(filename, reader, mode)
+}
--- a/pkg/downloader/chart_downloader_windows_test.go
+++ b/pkg/downloader/chart_downloader_windows_test.go
@@ -1,0 +1,131 @@
+//go:build windows
+
+/*
+Copyright The Helm Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package downloader
+
+import (
+	"os"
+	"path/filepath"
+	"sync"
+	"testing"
+
+	"helm.sh/helm/v4/pkg/cli"
+	"helm.sh/helm/v4/pkg/getter"
+	"helm.sh/helm/v4/pkg/repo/v1/repotest"
+)
+
+// TestParallelDownloadTo tests that parallel downloads to the same file
+// don't cause "Access Denied" errors on Windows. This test is Windows-specific
+// because the file locking behavior is only needed on Windows.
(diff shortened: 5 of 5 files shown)


TerryHowe (commented):
You might want to wait for feedback on other maintainers, but maybe make a LockedAtomicWirteFile to fileutils instead of this and definitely don't use a goto in that.

orgads:
> You might want to wait for feedback on other maintainers, but maybe make a LockedAtomicWirteFile to fileutils instead of this and definitely don't use a goto in that.

Thank you! Done.

orgads:
ping?

internal/fileutil/fileutil.go
```diff
@@ -48,3 +50,35 @@ func AtomicWriteFile(filename string, reader io.Reader, mode os.FileMode) error
...
+// LockedAtomicWriteFile atomically writes a file to disk with file locking to prevent
+// concurrent writes. This is particularly useful on Windows where concurrent writes
+// to the same file can cause "Access Denied" errors.
+//
+// The function acquires a lock on the target file, checks if it already exists,
+// and only writes if it doesn't exist. This prevents duplicate work when multiple
+// processes try to write the same file simultaneously.
+//
+// Returns true if the file was written, false if it already existed.
+func LockedAtomicWriteFile(filename string, data []byte, mode os.FileMode) (bool, error) {
```
TerryHowe: Why are you passing bytes here instead of the Reader?

orgads: Good point. Done.

orgads:
ping

orgads:
Please review this. It hits me all the time when working on Windows. See hashicorp/terraform-provider-helm#1683.

internal/fileutil/fileutil.go
```diff
@@ -48,3 +50,34 @@ func AtomicWriteFile(filename string, reader io.Reader, mode os.FileMode) error
...
+//
+// The function acquires a lock on the target file, checks if it already exists,
+// and only writes if it doesn't exist. This prevents duplicate work when multiple
+// processes try to write the same file simultaneously.
+//
+// Returns true if the file was written, false if it already existed.
+func LockedAtomicWriteFile(filename string, reader io.Reader, mode os.FileMode) (bool, error) {
+	// Acquire a file lock for this specific file path to prevent concurrent writes
+	fileLock := flock.New(filename + ".lock")
+	if err := fileLock.Lock(); err != nil {
```
TerryHowe: Should this to a `fileLock.TryLock()` instead? If that fails, it is probably being written by another thread/process would be my understanding.

I'm not entirely sure how flock works, but couldn't you just use the file as the lock file?

Messing around with flock on mac and `fileLock` file is not removed so this will leave junk around.

internal/fileutil/fileutil.go
```diff
@@ -48,3 +50,34 @@ func AtomicWriteFile(filename string, reader io.Reader, mode os.FileMode) error
...
+		return false, err
+	}
+	defer fileLock.Unlock()
+
+	// Check if the file already exists (another process might have already written it)
```
TerryHowe: If you use TryLock, you'd probably want to change this to look the returned `FileInfo.Size()` here.

TerryHowe (changes requested):

internal/fileutil/fileutil.go
orgads: @TerryHowe Thank you for the thorough review! You were absolutely right about the lock file cleanup issue.

## Changes made:

1. **Lock file cleanup**: You're correct that `.lock` files were being left behind as "junk" on Mac (and other systems). The `flock` library only releases the lock but doesn't remove the file. I've added explicit cleanup:
```go
defer func() {
    fileLock.Unlock()
    // Clean up the lock file to avoid leaving "junk" around
    os.Remove(lockFileName)
}()
```
This ensures the .lock file is removed after use, addressing your concern.

2. Why we can't use the target file as lock: After testing, I discovered that on Windows, flock.Lock() creates an empty file if the target doesn't exist. This breaks atomic writes:
- Lock creates empty target file
- Code sees file exists and skips writing content
- Result: empty files instead of actual chart data
3. Lock() vs TryLock(): Keeping Lock() for serialized access without retry complexity.
4. Enhanced test: Added comprehensive verification:
- File is not empty
- File has correct size (matches source)
- File has valid gzip magic bytes (0x1f 0x8b)
- No lock file is left behind

Testing:

- ✅ Lock files are properly cleaned up (test verifies this)
- ✅ Concurrent writes work correctly (only one writes, others skip)
- ✅ Test verifies actual content is written, not empty files
- ✅ Test catches both empty file bug and lock file cleanup issues
- ✅ All downloader tests pass

The implementation now handles concurrent writes correctly while cleaning up lock files to avoid leaving junk around, with tests ensuring both behaviors work properly.

TerryHowe (dismissed):
/lgtm

orgads:
ping?

orgads:
Anybody? @gjenkins8 @robertsirc @scottrigby?

SeanKilleen:
Especially with the Helm 4.0 announcement, it would be great to get this in so that the basics of Helm continue to work solidly for everyone. I hit this issue multiple times per day.

I noticed it's linked to an issue that appears to be auto-closed. I wonder if a new issue being opened and linked to this would get it the attention it deserves?

TerryHowe:
Closes: https://github.com/helm/helm/issues/31633

orgads:
> Closes: #31633

Updated the commit message and the PR description.

banjoh (approved):
LGTM

banjoh (dismissed):

SeanKilleen:
What is the process for this PR to be reviewed and approved by one additional person?

Asking because there was a comment on the issue that the PR is stale, but it's already received one approval and is just pending (presumably) 1 more approval.

Asking as the non-author, because I have hit this issue 4 times now during attempted Helm deployments this morning and am very much looking forward to the symptom being resolved. :)

orgads:
I just gave up and switched to WSL. It works much better.

SeanKilleen:
> I just gave up and switched to WSL. It works much better.

I'd consider that except I'm working with a team whose understanding and familiarity with WSL varies. Hopefully this PR can get the one additional approval it needs and get things resolved. 🤞

pkg/downloader/chart_downloader.go
```diff
@@ -156,7 +156,11 @@ func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *proven
 	}
 
 	destfile := filepath.Join(dest, name)
-	if err := fileutil.AtomicWriteFile(destfile, data, 0644); err != nil {
+
+	// Use LockedAtomicWriteFile to handle concurrent writes safely
+	// This prevents "Access Denied" errors on Windows when multiple processes
+	// try to write to the same file simultaneously
+	if _, err := fileutil.LockedAtomicWriteFile(destfile, data, 0644); err != nil {
```
gjenkins8: I don't think we should change the behavior on non-windows. Linux (for example) doesn't have the problem mention in https://github.com/helm/helm/issues/31633 I believe.

To do this, build target specific tags would/should be used. ref: https://go.dev/wiki/TargetSpecific

orgads: Even if the file is not locked, having multiple concurrent writers to the same file is bad and can lead to unexpected behavior.

SeanKilleen:
Hi all -- I see @gjenkins8 reviewed this and added a comment and @orgads responded in a few hours. That was 3 weeks ago. Coming back to this thread because I was once again blocked on a deployment and had to repeat it 5 times due to the locking issue.

Is there a way we can come to agreement on this? The point @orgads raises makes sense to be in general, and it seems that all the environments would benefit from the change. However, I'm not an expert in this specific stack.

pkg/downloader/chart_downloader.go
orgads: Done.

orgads:
Rebased and separated Windows/non-Windows.

internal/fileutil/fileutil_unix.go
```diff
@@ -0,0 +1,45 @@
...
+		// File already exists, skip writing
+		return false, nil
+	}
+
+	// File doesn't exist, write it atomically
+	if err := AtomicWriteFile(filename, reader, mode); err != nil {
+		return false, err
+	}
+
+	return true, nil
```
TerryHowe: I'd be curious what others think, but I'd be inclined to go this direction to preserve current behavior.
```suggestion
	return true, AtomicWriteFile(filename, reader, mode)
```

Another thing is this is not locked at all. Since we are ignoring the written bool, maybe call this something like `PlatformAtomicWriteFile`  and it just calls `AtomicWriteFile` here and `LockedAtomicWriteFile` for windows.

joejulian: I agree. The current PR changes the behavior and will break re-publishing charts with the same version/name.

joejulian (changes requested):
This breaks existing behavior. It will leave stale chart/provenance data if a chart is republished with the same version/name

internal/fileutil/fileutil_unix.go
orgads: Done

joejulian (approved):
lgtm, thanks!

TerryHowe (approved):
/lgtm

I don't have a windows system atm to test this, but looks good

SeanKilleen:
Thank you @joejulian and @TerryHowe for giving this the reviews it needs. Thank you to @orgads for continuing to work on this and get it across the finish line 🎉 

It seems like there's nothing blocking this at this point from being merged and getting into the release cycle. I and other Helm users on Windows will be grateful for that!

PR #31140, opened 2025-08-15
docs: Add installation method with gah

marverix:

Hello!
This PR adds to the documentation installation method using [gah](https://github.com/marverix/gah):
> gah is an GitHub Releases app installer, that DOES NOT REQUIRE SUDO! It is a simple bash script that downloads the latest release of an app from GitHub and installs it in ~/.local/bin. It is designed to be used with apps that are distributed as a single binary file.

N/A

**If applicable**:


TerryHowe (approved):
/lgtm

Issue #31144, opened 2025-08-17
hip-0026: implementation tracking issue (Helm 4 plugin system)


scottrigby:
Tracking issue for Helm 4 plugin system HIP implementation: https://github.com/helm/community/blob/main/hips/hip-0026.md#implementation-plan

### Status
- [x] Implementation complete and fully functional
- [ ] Maintainer review and approval (@gjenkins8 and @scottrigby collaborated on the work, so needs another maintainer review)
- [ ] Merge into upstream `main` before the Helm 4 pre-release (scheduled for Aug 31, 2025)

### Reviewing
- Complete work is reviewable as a single, squashed commit: https://github.com/scottrigby/helm/tree/_plugin-system_squash_rebase_main
- Commit history (for the curious): https://github.com/scottrigby/helm/tree/_plugin-system_original

### PRs

As smaller PRs are encouraged for Helm generally, we are quickly breaking out the complete work branch into smaller, more easily reviewable PRs, some of which must be merged sequentially:
- [ ] https://github.com/helm/helm/pull/31142
- [ ] Runtime abstraction to encapsulate subprocess code and enable future runtimes
- [ ] Plugin type abstraction to structure existing CLI and Getter plugins and enable future types
- [ ] Move Postrenderer to a plugin type
- [ ] Implement Wasm runtime
- [ ] Wasm runtime support for each plugin type (Wasm host functions)

Additional PRs (can be after prerelease):
- [ ] Update pkg/getter and pkg/registry to download plugins as well as charts (no public-facing changes needed in other packages)
- [ ] Refactor internal/plugin/installer package (fix tarball support for HTTP and Local installers, and add OCI installer)
- [ ] Add plugin signing and verification

scottrigby:
Removing milestone. The work is complete, the presentation was just complete at KubeCon NA 2025 for ContribFest, the docs are well underway and completing that task has an issue in the helm-www repo. The blog post still needs to be done and so does collaborating with existing plugin authors to update their plugins. Will keep this open until we either make sub-issues for those open tasks or complete them.

PR #31142, opened 2025-08-16
[HIP-0026] Move pkg/plugin -> internal/plugin

gjenkins8:
Part of implementing hip-0026. For full context, see:
- https://github.com/helm/helm/issues/31144

**What this PR does / why we need it**:
This PR moves the existing plugin package to `internal/`, as a precursor for merging in changes from https://github.com/scottrigby/helm/tree/plugin-system (the reworked plugin system to support greater plugin extensibility and Wasm based plugins)

(As part of those changes, public interfaces will be returned to `pkg/plugin`)

**Special notes for your reviewer**:

**If applicable**:
- [ ] this PR contains user facing changes (the `docs needed` label should be applied if so)
- [ ] this PR contains unit tests
- [x] this PR has been tested for backwards compatibility

PR #31145, opened 2025-08-17
[HIP-0026] Plugin runtime interface

scottrigby:
Adds plugin Runtime abstraction to encapsulate subprocess code and enable future runtimes.

Depends on:
- #31142

X This PR does NOT contain user facing changes


internal/plugin/runtime_subprocess_getter.go
```diff
@@ -0,0 +1,98 @@
...
+		msg.Options.KeyFile,
+		msg.Options.CAFile,
+		msg.Href)
+
+	// TODO should we append to input.Env too?
+	//env := append(os.Environ(), input.Env...)
+	//env = append(env,
+	//	fmt.Sprintf("HELM_PLUGIN_USERNAME=%s", msg.Options.Username),
+	//	fmt.Sprintf("HELM_PLUGIN_PASSWORD=%s", msg.Options.Password),
+	//	fmt.Sprintf("HELM_PLUGIN_PASS_CREDENTIALS_ALL=%t", msg.Options.PassCredentialsAll))
```
jessesimpson36: seems like this comment doesn't need to be here since it's uncommented 2 lines down.

internal/plugin/loader.go
```diff
@@ -0,0 +1,225 @@
...
+// filterFunc is a function that filters plugins
+type filterFunc func(Plugin) bool
+
+// FindPlugins returns a list of plugins that match the descriptor
+func FindPlugins(pluginsDirs []string, descriptor Descriptor) ([]Plugin, error) {
+	return findPlugins(pluginsDirs, LoadAll, makeDescriptorFilter(descriptor))
+}
+
+// findPlugins is the internal implementation that uses the find and filter functions
+func findPlugins(pluginsDirs []string, findFunc findFunc, filterFunc filterFunc) ([]Plugin, error) {
```
jessesimpson36: I'd suggest the name findFunc not be of type findFunc, otherwise you're shadowing over the type declaration with a variable name, and I don't think you could create another of type findFunc in this body.

internal/plugin/metadata.go
```diff
@@ -0,0 +1,142 @@
...
+	SourceURL string
+
+	// Config contains the type-specific configuration for this plugin
+	Config Config
+
+	// RuntimeConfig contains the runtime-specific configuration
+	RuntimeConfig RuntimeConfig
+}
+
+func (m Metadata) Validate() error {
```
jessesimpson36, line 51: This Validate function returns at the first error it sees. I could imagine a more helpful Validate function returning all the validation errors all at once, sortof how the way linters in editors work. Perhaps this should return a list of errors rather than just an error? or could wrap a list of errors.

jessesimpson36 (approved):
I've left comments, but I still approve. 

Overall, I've been testing this out manually, and everything seems to work fine. The unit tests appear to cover the happy-paths well, and could easily be copy/pasted later to test any missing error edge-cases, or used as helper documentation for a user to make your own plugin (that's how I've been making one of my test plugins). 

I reviewed the HIP once more, and these changes appear to abide by what was declared in the HIP, no surprising changes.

jessesimpson36:
Test downloader plugin I wrote from this PR
https://github.com/jessesimpson36/helm-serial-getter

internal/plugin/runtime_subprocess_getter.go
scottrigby: I agree. Will trim it. 

BTW, @gjenkins8 found a novel way to address this TODO in a PR on my fork. We're keeping the upstream PRs as tidy as possible since several of them have sequential depednencies. If you're curious, it's that fix is here: https://github.com/scottrigby/helm/pull/24 We'll likely wait until after the pre-release freeze to make an upstream PR for those changes.

internal/plugin/loader.go
scottrigby: good point. will fix this variable shadowing issue.

internal/plugin/metadata.go
scottrigby: good idea. we use errors.Join() elsewhere too (eg, chart validation). done

joejulian: Did we ever do this?
https://github.com/helm/helm/pull/12298

internal/plugin/loader.go
```diff
@@ -0,0 +1,225 @@
...
+func newPrototypePluginManager() *prototypePluginManager {
+	return &prototypePluginManager{
+		runtimes: map[string]Runtime{
+			"subprocess": &RuntimeSubprocess{},
+		},
+	}
+}
+
+func (pm *prototypePluginManager) RegisterRuntime(runtimeName string, runtime Runtime) {
+	//if _, exists := pm.runtimes[runtimeName]; exists {
```
joejulian:
```suggestion
```

internal/plugin/metadata_test.go
```diff
@@ -0,0 +1,137 @@
...
+		{true, mockSubprocessCLIPlugin(t, "abcdefghijklmnopqrstuvwxyz0123456789_-ABC")},
+		{true, mockSubprocessCLIPlugin(t, "foo-bar-FOO-BAR_1234")},
+		{false, mockSubprocessCLIPlugin(t, "foo -bar")},
+		{false, mockSubprocessCLIPlugin(t, "$foo -bar")}, // Test leading chars
+		{false, mockSubprocessCLIPlugin(t, "foo -bar ")}, // Test trailing chars
+		{false, mockSubprocessCLIPlugin(t, "foo\nbar")},  // Test newline
+		{true, mockNoCommand},     // Test no command metadata works
+		{true, mockLegacyCommand}, // Test legacy command metadata works
+		{false, mockWithCommand},  // Test platformCommand and command both set fails
+		{false, mockWithHooks},    // Test platformHooks and hooks both set fails
```
joejulian: You've done the work of describing the test, why not make that string part of the test data so the result can say which test failed (or succeeded)?

joejulian (dismissed):

internal/plugin/metadata.go
scottrigby: > Did we ever do this? #12298

Yes, earlier this year. I think your PR should have been merged 2 yrs ago, I think it just got missed. Added some history here: https://github.com/helm/helm/pull/12298#issuecomment-3212770756 🙂

internal/plugin/loader.go
scottrigby: fixed

internal/plugin/metadata_test.go
```diff
@@ -0,0 +1,137 @@
...
+	} {
+		err := item.plug.Metadata().Validate()
+		if item.pass && err != nil {
+			t.Errorf("failed to validate case %d: %s", i, err)
+		} else if !item.pass && err == nil {
+			t.Errorf("expected case %d to fail", i)
+		}
+	}
+}
```
jessesimpson36:
```suggestion
	for i, item := range []struct {
		pass      bool
		plug      Plugin
		errString string
	}{
		{true, mockSubprocessCLIPlugin(t, "abcdefghijklmnopqrstuvwxyz0123456789_-ABC"), ""},
		{true, mockSubprocessCLIPlugin(t, "foo-bar-FOO-BAR_1234"), ""},
		{false, mockSubprocessCLIPlugin(t, "foo -bar"), "invalid name"},
		{false, mockSubprocessCLIPlugin(t, "$foo -bar"), "invalid name"}, // Test leading chars
		{false, mockSubprocessCLIPlugin(t, "foo -bar "), "invalid name"}, // Test trailing chars
		{false, mockSubprocessCLIPlugin(t, "foo\nbar"), "invalid name"},  // Test newline
		{true, mockNoCommand, ""},     // Test no command metadata works
		{true, mockLegacyCommand, ""}, // Test legacy command metadata works
		{false, mockWithCommand, "runtime config validation failed: both platformCommand and command are set"}, // Test platformCommand and command both set fails
		{false, mockWithHooks, "runtime config validation failed: both platformHooks and hooks are set"},       // Test platformHooks and hooks both set fails
	} {
		err := item.plug.Metadata().Validate()
		if item.pass && err != nil {
			t.Errorf("failed to validate case %d: %s", i, err)
		} else if !item.pass && err == nil {
			t.Errorf("expected case %d to fail", i)
		}
		if !item.pass && err.Error() != item.errString {
			t.Errorf("index [%d]: expected the following error: %s, but got: %s", i, item.errString, err.Error())
		}
	}
}
```

internal/plugin/metadata_test.go
scottrigby: I committed @jessesimpson36 's code suggestion addressing this 1238a78300d15691e9b5122a7db22dfbc5173e96. Is this what you meant?

internal/plugin/loader.go
scottrigby: committed

internal/plugin/metadata_test.go
scottrigby: committed

joejulian (approved):

scottrigby:
Thanks y'all, merging so we can tee up the next PR in the list on #31144

PR #31156, opened 2025-08-18
fix: set repo authorizer in registry.Client.Resolve()

estroz:

The `registry.Client.Resolve` method does not set the `Client` field on the repository attempting to be resolved.

**If applicable**:

### DIFF
--- a/pkg/registry/client.go
+++ b/pkg/registry/client.go
@@ -818,6 +818,7 @@ func (c *Client) Resolve(ref string) (desc ocispec.Descriptor, err error) {
 		return desc, err
 	}
 	remoteRepository.PlainHTTP = c.plainHTTP
+	remoteRepository.Client = c.authorizer
 
 	parsedReference, err := newReference(ref)
 	if err != nil {


TerryHowe:
Could you make an issue for this explaining how to reproduce?

estroz:
@TerryHowe is that necessary? All other methods on this type set `Repository.Client = c.authorizer`. Are you asking for tracking's sake?

estroz:
#31158

estroz:
@TerryHowe is it possible to backport this to 3.18? If not that is ok

TerryHowe:
Thanks. I just wanted to understand your use case where this turned up. It doesn't happen in the CLI that I know of.

TerryHowe:
I would think a backport to 3.18 would be possible, but I'm just one person.

estroz:
>Thanks. I just wanted to understand your use case where this turned up. It doesn't happen in the CLI that I know of.

Yes I am not able to see where this would matter in the CLI.

>I would think a backport to 3.18 would be possible, but I'm just one person.

I can make the cherry pick once this merges.

scottrigby (approved):
thanks @estroz

> All other methods on this type set `Repository.Client = c.authorizer`

agreed, we just missed this 👍

scottrigby:
> I would think a backport to 3.18 would be possible, but I'm just one person.

@TerryHowe is right, because this is a legitimate bug fix. see the timeline policy on this: https://github.com/helm/community/blob/main/hips/hip-0012.md#maintaining-helm-3

TerryHowe:
Thanks @estroz please open a PR to backport this commit to `dev-v3`

PR #31169, opened 2025-08-20
fix(helm-lint): do not validate metadata.name for List resources

### LINKED ISSUE #13192: Helm Lint: ConfigMapList metadata.name false positive
## Summary

Whenever you have a plain kubernetes resource of type `ConfigMapList`, `helm lint` warns about `metadata.name`:

```
% helm lint strict path/to/our/chart
[WARNING] templates/dashboards-configmap.yaml: object name does not conform to Kubernetes naming requirements: "": metadata.name: Invalid value: "": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')

```

Sample manifest file to reproduce:

```
% cat configmap.yaml
apiVersion: v1
kind: ConfigMapList
items:
  - apiVersion: v1
    kind: ConfigMap
    metadata:
      name: foo
    data: null
```

However, this is incorrect. `ConfigMapList`s do not support `metadata.name`. Only individual `ConfigMap` entries in it do.

This seems to be loosely related to https://github.com/kubernetes-sigs/kustomize/issues/769 in `kustomize`.

Follow-up-of: https://github.com/helm/helm/issues/8615, as per https://github.com/helm/helm/issues/8615#issuecomment-2024771964

## Output of `helm version`:

```
% helm version
version.BuildInfo{Version:"v3.15.1", GitCommit:"e211f2aa62992bd72586b395de50979e31231829", GitTreeState:"clean", GoVersion:"go1.22.3"}
```

## Output of `kubectl version`:

```
% kubectl version
Client Version: v1.28.2
Server Version: v1.25.15
WARNING: version difference between client (1.28) and server (1.25) exceeds the supported minor version skew of +/-1
```

## Cloud Provider/Platform (AKS, GKE, Minikube etc.):

AKS

thiagowfx:
Closes #13192

- Issue: helm lint was incorrectly validating `metadata.name` for `List`-type resources (like `ConfigMapList`, `SecretList`, etc.)
- Root Cause: The validation logic in `pkg/lint/rules/template.go` was calling `validateMetadataName()` for ALL resources without checking if they were `List` types
- Impact: `List` resources don't require meaningful names at the top level, only the items within the list need names

Enhance the list detection logic to be defensive by checking for both the "List" suffix and the presence of an `items` field.

**Special notes for your reviewer**: c.f. linked ticket

**If applicable**:

### DIFF
--- a/pkg/lint/rules/template.go
+++ b/pkg/lint/rules/template.go
@@ -162,7 +162,7 @@ func TemplatesWithSkipSchemaValidation(linter *support.Linter, values map[string
 					// Refs https://github.com/helm/helm/issues/8596
 					// Skip metadata name validation for List resources as they don't require meaningful names
 					// Refs https://github.com/helm/helm/issues/13192
-					if !isListResource(yamlStruct) {
+					if !isListResource(yamlStruct, renderedContent) {
 						linter.RunLinterRule(support.WarningSev, fpath, validateMetadataName(yamlStruct))
 					}
 					linter.RunLinterRule(support.WarningSev, fpath, validateNoDeprecations(yamlStruct, kubeVersion))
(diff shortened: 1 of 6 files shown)


pkg/lint/rules/template.go
```diff
@@ -235,6 +239,12 @@ func validateYamlContent(err error) error {
 	return nil
 }
 
+// isListResource returns true if the resource is a Kubernetes List type
+// (e.g. ConfigMapList, SecretList).
+func isListResource(obj *k8sYamlStruct) bool {
+	return strings.HasSuffix(obj.Kind, "List")
```
benoittgt: I'm wondering if we should be more defensive here. 
For example, treat as a list only if:
- kind == "List", or
- kind ends with "List" AND the top-level object has an items field that is an array.

This would avoid incorrectly skipping `metadata.name` validation for custom resources  that happen to end with "List" but aren't actually list containers. While uncommon, a CRD could define a Kind like "AccessList" that represents a single resource, not a collection.

The items field check should be reliable since all Kubernetes list types  (ConfigMapList, SecretList, etc.) have this structure, but non-list resources  typically wouldn't. 🤷

benoittgt (approved):

pkg/lint/rules/template.go
thiagowfx: Noted. Will re-request your review once a new revision is ready. Thanks!

thiagowfx: Done.

PR #31146, opened 2025-08-18
[HIP-0026] Plugin types and plugin apiVersion v1

scottrigby:
Adds plugin types and plugin apiVersion v1.

V1 plugins structure existing CLI and Getter plugins, and enable future plugin types.

Part of implementing hip-0026. For full context, see:
- https://github.com/helm/helm/issues/31144

Depends on:
- #31145
Please review that PR first. After that PR is merged, will change this PR base branch to `main` and move out of draft.

Checklist:
- [x] This PR contains user facing changes
- [x] this PR contains unit tests
- [x] this PR has been tested for backwards compatibility

PR #31172, opened 2025-08-21
[HIP-0026] Plugin OCI installer

scottrigby:
Adds OCI installer to Helm 4 plugins. Also updates pkg/getter and pkg/registry for better re-use. These packages now allow downloading plugins as well as charts. No public SDK changes for these packages were needed in this implementation.

Depends on:
- #31146

Checklist:

### Manual testing

Unit tests are comprehensive. You can also manually test with:

```console
% go run ./cmd/helm registry login ghcr.io
Login Succeeded

% go run ./cmd/helm plugin install oci://ghcr.io/scottrigby/example-cli:v0.2.0
Installed plugin: example-cli

% go run ./cmd/helm plugin list                                               
NAME			VERSION		TYPE		APIVERSION		SOURCE                                              
example-cli		0.1.0  		cli/v1		v1				https://github.com/scottrigby/h4-simple-plugin-types
```

Source code for example plugin: https://github.com/scottrigby/h4-example-plugin-types.

### DIFF
--- a/internal/plugin/installer/oci_installer.go
+++ b/internal/plugin/installer/oci_installer.go
@@ -1,0 +1,229 @@
+/*
+Copyright The Helm Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package installer
+
+import (
+	"archive/tar"
+	"bytes"
+	"compress/gzip"
+	"fmt"
+	"io"
+	"log/slog"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"helm.sh/helm/v4/internal/plugin/cache"
+	"helm.sh/helm/v4/internal/third_party/dep/fs"
+	"helm.sh/helm/v4/pkg/cli"
+	"helm.sh/helm/v4/pkg/getter"
+	"helm.sh/helm/v4/pkg/helmpath"
+	"helm.sh/helm/v4/pkg/registry"
+)
+
+// OCIInstaller installs plugins from OCI registries
+type OCIInstaller struct {
+	CacheDir   string
+	PluginName string
+	base
+	settings *cli.EnvSettings
+	getter   getter.Getter
+}
+
+// NewOCIInstaller creates a new OCIInstaller with optional getter options
+func NewOCIInstaller(source string, options ...getter.Option) (*OCIInstaller, error) {
+	ref := strings.TrimPrefix(source, fmt.Sprintf("%s://", registry.OCIScheme))
+
+	// Extract plugin name from OCI reference
+	// e.g., "ghcr.io/user/plugin-name:v1.0.0" -> "plugin-name"
+	parts := strings.Split(ref, "/")
+	if len(parts) < 2 {
+		return nil, fmt.Errorf("invalid OCI reference: %s", source)
+	}
+	lastPart := parts[len(parts)-1]
+	pluginName := lastPart
+	if idx := strings.LastIndex(lastPart, ":"); idx > 0 {
+		pluginName = lastPart[:idx]
+	}
+	if idx := strings.LastIndex(lastPart, "@"); idx > 0 {
+		pluginName = lastPart[:idx]
+	}
+
+	key, err := cache.Key(source)
+	if err != nil {
+		return nil, err
+	}
+
+	settings := cli.New()
+
+	// Always add plugin artifact type and any provided options
+	pluginOptions := append([]getter.Option{getter.WithArtifactType("plugin")}, options...)
+	getterProvider, err := getter.NewOCIGetter(pluginOptions...)
+	if err != nil {
+		return nil, err
+	}
+
+	i := &OCIInstaller{
+		CacheDir:   helmpath.CachePath("plugins", key),
+		PluginName: pluginName,
+		base:       newBase(source),
+		settings:   settings,
+		getter:     getterProvider,
+	}
+	return i, nil
+}
+
+// Install downloads and installs a plugin from OCI registry
+// Implements Installer.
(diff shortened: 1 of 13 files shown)


internal/plugin/installer/oci_installer.go
```diff
@@ -0,0 +1,229 @@
...
+	CacheDir   string
+	PluginName string
+	base
+	settings *cli.EnvSettings
+	getter   getter.Getter
+}
+
+// NewOCIInstaller creates a new OCIInstaller with optional getter options
+func NewOCIInstaller(source string, options ...getter.Option) (*OCIInstaller, error) {
+	ref := strings.TrimPrefix(source, fmt.Sprintf("%s://", registry.OCIScheme))
```
TerryHowe: There is a `newReference` although I wish I could push that code to oras-go. oras-go has something, but as I recall they don't support tag and digest at the same time properly. Maybe in `pkg/regsitry/plugin.go` some function called `GetPluginName(source string) (string, error)` that does all this?

pkg/registry/plugin.go
```diff
@@ -0,0 +1,176 @@
...
+
+// PluginPullResult contains the result of a plugin pull operation
+type PluginPullResult struct {
+	Manifest       ocispec.Descriptor
+	PluginData     []byte
+	ProvenanceData []byte // Optional provenance data
+	Ref            string
+	PluginName     string
+}
+
```
TerryHowe, line 47: Given some greenfield project I would of tried to provide some generic content generator to pass to a generic oci client rather than the content having a client, but given the current code base, I think this make more sense.

pkg/registry/oci.go
```diff
@@ -0,0 +1,162 @@
...
+
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+	"oras.land/oras-go/v2"
+	"oras.land/oras-go/v2/content"
+	"oras.land/oras-go/v2/content/memory"
+	"oras.land/oras-go/v2/registry/remote"
+	"oras.land/oras-go/v2/registry/remote/auth"
+	"oras.land/oras-go/v2/registry/remote/credentials"
+)
+
```
TerryHowe: why isn't this file named `generic_client.go`?

pkg/cmd/plugin_install.go
```diff
@@ -68,10 +88,39 @@ func (o *pluginInstallOptions) complete(args []string) error {
...
+		if o.insecureSkipTLSverify {
+			options = append(options, getter.WithInsecureSkipVerifyTLS(o.insecureSkipTLSverify))
+		}
+
+		if o.plainHTTP {
+			options = append(options, getter.WithPlainHTTP(o.plainHTTP))
+		}
+
+		if o.username != "" || o.password != "" {
+			options = append(options, getter.WithBasicAuth(o.username, o.password))
```
jessesimpson36: Not that I have a hard opinion on this, but I'm wondering if these `getter.WithBasicAuth()` or other `With` function calls could contain the logic for `if o.username != "" || if o.password != ""`

Then this `newInstallerForSource` function would just be a registration of options, like

```
options = append(options, getter.WithTLSClientConfig(o.certFile, o.keyFile, o.caFile))
options = append(options, getter.WithInsecureSkipVerifyTLS(o.insecureSkipTLSverify))
options = append(options, getter.WithPlainHTTP(o.plainHTTP))
options = append(options, getter.WithBasicAuth(o.username, o.password))

```

internal/plugin/installer/oci_installer.go
scottrigby: `newReference` is nice. looks like you made it private about 6 months ago https://github.com/helm/helm/pull/12690/commits/d2b94f62004e79864ec530989109cf0effd4aaae. Was it for a reason that's still relevant now?

If there's no objection to making reference struct and newReference func public again I'd update this to use that. Otherwise I'm making a fairly minimal first pass with this here.

pkg/registry/oci.go
scottrigby: good idea. done

pkg/cmd/plugin_install.go
scottrigby: you're right, there's no need to wrap in these conditionals :+1: I simplified

```go
options := []getter.Option{
	getter.WithTLSClientConfig(o.certFile, o.keyFile, o.caFile),
	getter.WithInsecureSkipVerifyTLS(o.insecureSkipTLSverify),
	getter.WithPlainHTTP(o.plainHTTP),
	getter.WithBasicAuth(o.username, o.password),
}
```

internal/plugin/installer/oci_installer.go
TerryHowe: Rather than make newReference public and being stuck with it, I was thinking in pkg/regsitry/plugin.go some function called GetPluginName(source string) (string, error) that does all this and uses newReference.

scottrigby: i see, this is what you meant. not a bad idea. will do 👍

scottrigby: WDYT? 4c83dd7

TerryHowe (approved):
/lgtm

jessesimpson36 (approved):

internal/plugin/installer/oci_installer.go
```diff
@@ -0,0 +1,217 @@
...
+			if err := os.MkdirAll(path, 0755); err != nil {
+				return err
+			}
+		case tar.TypeReg:
+			dir := filepath.Dir(path)
+			if err := os.MkdirAll(dir, 0755); err != nil {
+				return err
+			}
+
+			outFile, err := os.OpenFile(path, os.O_CREATE|os.O_RDWR, os.FileMode(header.Mode))
```
sabre1041, line 199: Should we add a `defer outfile.Close()` to avoid multiple calls to close?

sabre1041 (dismissed):
Looks really good. Only one comment, but its not a blocking comment. Feel free to address your thoughts. Otherwise, LGTM

internal/plugin/installer/oci_installer.go
scottrigby: good catch. fixing

scottrigby: fixed in both http and oci plugin installers. added you as co-author https://github.com/helm/helm/pull/31172/commits/73f058c27fc7cb9745766ac37d0850517f3052e2

sabre1041 (approved):
LGTM

scottrigby:
thanks everyone. bringing this one in

PR #31174, opened 2025-08-21
[HIP-0026] Plugin tarball support for HTTP and local installers

scottrigby:
Allows installing from a tarball both over HTTP and from local path.

Depends on:
- #31146

Checklist:

Manually test with:
```sh
% go run ./cmd/helm plugin install https://github.com/scottrigby/h4-example-plugin-types/raw/refs/heads/main/tarballs/example-cli.tgz
Installed plugin: example-cli

# example tarball attached to a GitHub release
% go run ./cmd/helm plugin install https://github.com/scottrigby/h4-example-plugin-types/releases/download/tarball-test-0.2.0/example-legacy-downloader.tgz

# or download a tarball, then install from a local path
% go run ./cmd/helm plugin install PATH/TO/example-cli.tgz
```

### DIFF
--- a/internal/plugin/installer/http_installer_test.go
+++ b/internal/plugin/installer/http_installer_test.go
@@ -348,3 +348,255 @@ func TestMediaTypeToExtension(t *testing.T) {
 		}
 	}
 }
+
+func TestExtractWithNestedDirectories(t *testing.T) {
+	source := "https://repo.localdomain/plugins/nested-plugin-0.0.1.tar.gz"
+	tempDir := t.TempDir()
+
+	// Set the umask to default open permissions so we can actually test
+	oldmask := syscall.Umask(0000)
+	defer func() {
+		syscall.Umask(oldmask)
+	}()
+
+	// Write a tarball with nested directory structure
+	var tarbuf bytes.Buffer
+	tw := tar.NewWriter(&tarbuf)
+	var files = []struct {
+		Name     string
+		Body     string
+		Mode     int64
+		TypeFlag byte
+	}{
+		{"plugin.yaml", "plugin metadata", 0600, tar.TypeReg},
+		{"bin/", "", 0755, tar.TypeDir},
+		{"bin/plugin", "#!/bin/bash\necho plugin", 0755, tar.TypeReg},
+		{"docs/", "", 0755, tar.TypeDir},
+		{"docs/README.md", "readme content", 0644, tar.TypeReg},
+		{"docs/examples/", "", 0755, tar.TypeDir},
+		{"docs/examples/example1.yaml", "example content", 0644, tar.TypeReg},
+	}
+
+	for _, file := range files {
+		hdr := &tar.Header{
+			Name:     file.Name,
+			Typeflag: file.TypeFlag,
+			Mode:     file.Mode,
+			Size:     int64(len(file.Body)),
+		}
+		if err := tw.WriteHeader(hdr); err != nil {
+			t.Fatal(err)
+		}
+		if file.TypeFlag == tar.TypeReg {
+			if _, err := tw.Write([]byte(file.Body)); err != nil {
+				t.Fatal(err)
+			}
+		}
+	}
+
+	if err := tw.Close(); err != nil {
+		t.Fatal(err)
+	}
+
+	var buf bytes.Buffer
+	gz := gzip.NewWriter(&buf)
+	if _, err := gz.Write(tarbuf.Bytes()); err != nil {
+		t.Fatal(err)
+	}
+	gz.Close()
+
+	extractor, err := NewExtractor(source)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// First extraction
+	if err = extractor.Extract(&buf, tempDir); err != nil {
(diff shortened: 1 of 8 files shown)


internal/plugin/installer/plugin_structure.go
```diff
@@ -0,0 +1,80 @@
...
+	"path/filepath"
+	"strings"
+
+	"helm.sh/helm/v4/internal/plugin"
+)
+
+// detectPluginRoot searches for plugin.yaml in the extracted directory
+// and returns the path to the directory containing it.
+// This handles cases where the tarball contains the plugin in a subdirectory.
+func detectPluginRoot(extractDir string) (string, error) {
```
jessesimpson36, line 30: I'm a little confused by the use-case for `detectPluginRoot` checking one directory deep for a `plugin.yaml` file. Might just be a convenience function.

What it makes me think of is how bitnami charts do something like have a `charts/` directory, maybe someone would have plugins in a `plugins/` directory. But if I ran `helm plugin install`, on `plugins/` then it would be the first directory with a `plugin.yaml` file in it.

Another kind of user might expect a `helm plugin install` to install ALL plugins in that directory.

internal/plugin/installer/plugin_structure_test.go
```diff
@@ -0,0 +1,165 @@
...
+			setup: func(dir string) error {
+				subdir := filepath.Join(dir, "my-plugin")
+				if err := os.MkdirAll(subdir, 0755); err != nil {
+					return err
+				}
+				return os.WriteFile(filepath.Join(subdir, "plugin.yaml"), []byte("name: test"), 0644)
+			},
+			expectRoot:  "my-plugin",
+			expectError: false,
+		},
```
jessesimpson36, line 50:
```suggestion
		},
		{
			name: "multiple plugin.yamls in subdirectories",
			setup: func(dir string) error {
				subdir := filepath.Join(dir, "plugins", "plugin1")
				if err := os.MkdirAll(subdir, 0755); err != nil {
					return err
				}
				os.WriteFile(filepath.Join(subdir, "plugin.yaml"), []byte("name: test"), 0644)
				subdir = filepath.Join(dir, "plugins", "plugin2")
				if err := os.MkdirAll(subdir, 0755); err != nil {
					return err
				}
				return os.WriteFile(filepath.Join(subdir, "plugin.yaml"), []byte("name: test2"), 0644)
			},
			expectRoot:  "my-plugin",
			expectError: false,
		},
```

In case you wanted to be more explicit about supporting having multiple folders each with a plugin.yaml, but only getting the first one.

internal/plugin/installer/http_installer_test.go
```diff
@@ -348,3 +348,255 @@ func TestMediaTypeToExtension(t *testing.T) {
 		}
 	}
 }
+
+func TestExtractWithNestedDirectories(t *testing.T) {
+	source := "https://repo.localdomain/plugins/nested-plugin-0.0.1.tar.gz"
+	tempDir := t.TempDir()
+
+	// Set the umask to default open permissions so we can actually test
+	oldmask := syscall.Umask(0000)
```
jessesimpson36: I'm inclined to say that this should explicitly open files with the permissions it needs during Open calls.  If you call `syscall.Umask` , I'm wondering if you hit Ctrl-C Ctrl-C in an unfortunate timing, your shell might get stuck creating new files with 777 permissions unless you explicitly close the terminal and restart it. Or for subprocesses of this test case to not know that the umask has changed.

that might be overly paranoid though.

jessesimpson36 (changes requested):
Ack or respond to my other comments and I'll switch this to approve.

internal/plugin/installer/plugin_structure.go
scottrigby: i can see what you mean. But in this case, this is only called on ziptar archive. 

We're even more strict with about validating the location of Chart.yaml file within a chart archive. Helm enforces a Chart.yaml file within a top-level folder (of any name) at the root of the archive. All helm commands that accept a chart archive as an argument fail otherwise. I think this is acceptable since we have a `helm package` command to ensure it's packaged in the correct way.

In the plugin packaging/signing/verifying PR, I updated this logic to be equally as strict as charts. that PR adds a `plugin package` command so there isn't room for ambiguity about the archive structure, where the plugin.yaml file lives, etc (plugin.yaml is the comparable metadata file to Chart.yaml).

internal/plugin/installer/plugin_structure_test.go
scottrigby: I just think we should align plugin archives with how chart archives are handled.

internal/plugin/installer/http_installer_test.go
scottrigby: oof. this test has been here for 6 years. and you have a good point. I pushed a fix. How does this look? d3e13229a4bc7c5303bfcb030dd3641e123cf8d2

jessesimpson36 (approved):

jessesimpson36 (approved):

scottrigby:
pushing an amended commit to trigger build gh workflow

sabre1041 (approved):
LGTM

PR #31176, opened 2025-08-21
[HIP-0026] Plugin packaging, signing, and verification

scottrigby:
Adds plugin packaging, signing, and verification commands.

Depends on:
- #31146 
- #31172 
- #31174 (for now this PR includes the commit from there cherry-picked)

Please review these PRs first. After these PRs are merged, will change this PR base branch to `main` and move out of draft.

Checklist:

## Manual testing

you should be able to verify testing the tarballs in https://github.com/scottrigby/h4-example-plugins/tree/main/tarballs using my public key. You can either use the keybase method as described here https://helm.sh/docs/topics/provenance/#verifying-packages, or you can also just download my public key from here https://keybase.io/r6by/pgp_keys.asc and import with gpg.

You can of course also use your own key to test packging with signing and verifying the packages you made.

## Changes in this PR

1. adds `plugin package` and `plugin verify` commands

    attempts to sign the package by default. This plugin command has the same CLI options as `helm package`. You may use `--sign=false` to skip signing (not recommended for distribution, and emits a warning)

1. adds `PROVENANCE` status to `plugin list` output

    ```
    % go run ./cmd/helm plugin list
    NAME                     	VERSION	TYPE     	APIVERSION	PROVENANCE	SOURCE                                          
    example-cli              	0.1.0  	cli/v1   	v1        	signed    	https://github.com/scottrigby/h4-example-plugins
    example-getter           	0.1.0  	getter/v1	v1        	signed    	https://github.com/scottrigby/h4-example-plugins
    example-legacy-cli       	0.1.0  	cli/v1   	legacy    	signed    	unknown                                         
    example-legacy-downloader	0.1.0  	getter/v1	legacy    	signed    	unknown                                         
    example-unsigned         	0.1.0  	cli/v1   	v1        	unsigned  	https://github.com/scottrigby/h4-example-plugins
    ```
    provenance status can be:
    - "local dev": Plugin is a symlink (development mode)
    - "unsigned": No provenance file found
    - "invalid provenance": Provenance file is malformed
    - "mismatched provenance": Provenance file does not exactly match plugin
    - "signed": Valid signature exists for this exact plugin
1. `plugin install` now attempts to verify provenance by default:
    ```
    plugin install path/to/example-cli-0.1.0.tgz
    Verifying plugin signature...
    Signed by: Scott Rigby <scott@r6by.com>
    Using Key With Fingerprint: 208DD36ED5BB3745A16743A4C7C6FBB5B91C1155
    Plugin Hash Verified: sha256:4ef0c4b1cd402d6a170fadb14d68733226e37bff9e2df67ee34754f4a6d0d1c6
    Installed plugin: example-cli
    ```
    it also downloads the tarball along with the prov file alongside the unpacked plugin, inside the HELM_PLUGINS directory.

    if a correctly named provenance file exists next to the plugin — whether local directory, a local or HTTP tarball, or in the same OCI artifact, helm install will attempt verification. If the chart passes verification, the unpacked plugin dir and provenance file will be downloaded to the HELM_PLUGINS directory. Example:

    ```
    % ls -lah $(go run ./cmd/helm env HELM_PLUGINS)                               
    drwxr-xr-x@  4 r6by  staff   128B Aug 25 05:42 example-cli
    -rw-r--r--@  1 r6by  staff   511B Aug 25 05:42 example-cli.tgz
    -rw-r--r--@  1 r6by  staff   1.4K Aug 25 05:42 example-cli.tgz.prov
    ```
    If there is no provenance file, or if verification fails, the plugin will not be downloaded. You may bypass this with `--verify=false`, but the user is warned this is insecure if they do this. It is secure by default.
1. `plugin uninstall` now also removes the plugin's accompanying tarball and provenance file from HELM_PLUGIN directory if they exist locally
1. added `plugin verify`

    this accepts a path to a plugin tarball, which we also save for you in HELM_PLUGINS dir on `plugin install`

### DIFF
--- a/internal/plugin/installer/http_installer.go
+++ b/internal/plugin/installer/http_installer.go
@@ -16,22 +16,14 @@ limitations under the License.
 package installer // import "helm.sh/helm/v4/internal/plugin/installer"
 
 import (
-	"archive/tar"
 	"bytes"
-	"compress/gzip"
-	"errors"
 	"fmt"
-	"io"
 	"log/slog"
 	"os"
-	"path"
 	"path/filepath"
-	"regexp"
-	"slices"
 	"strings"
 
-	securejoin "github.com/cyphar/filepath-securejoin"
-
+	"helm.sh/helm/v4/internal/plugin"
 	"helm.sh/helm/v4/internal/plugin/cache"
 	"helm.sh/helm/v4/internal/third_party/dep/fs"
 	"helm.sh/helm/v4/pkg/cli"
@@ -46,45 +38,8 @@ type HTTPInstaller struct {
 	base
 	extractor Extractor
 	getter    getter.Getter
-}
-
-// TarGzExtractor extracts gzip compressed tar archives
-type TarGzExtractor struct{}
-
-// Extractor provides an interface for extracting archives
-type Extractor interface {
-	Extract(buffer *bytes.Buffer, targetDir string) error
-}
-
-// Extractors contains a map of suffixes and matching implementations of extractor to return
-var Extractors = map[string]Extractor{
-	".tar.gz": &TarGzExtractor{},
-	".tgz":    &TarGzExtractor{},
-}
-
-// Convert a media type to an extractor extension.
-//
-// This should be refactored in Helm 4, combined with the extension-based mechanism.
-func mediaTypeToExtension(mt string) (string, bool) {
-	switch strings.ToLower(mt) {
-	case "application/gzip", "application/x-gzip", "application/x-tgz", "application/x-gtar":
-		return ".tgz", true
-	case "application/octet-stream":
-		// Generic binary type - we'll need to check the URL suffix
-		return "", false
-	default:
-		return "", false
-	}
-}
-
-// NewExtractor creates a new extractor matching the source file name
-func NewExtractor(source string) (Extractor, error) {
-	for suffix, extractor := range Extractors {
-		if strings.HasSuffix(source, suffix) {
-			return extractor, nil
-		}
-	}
-	return nil, fmt.Errorf("no extractor implemented yet for %s", source)
+	// Provenance data to save after installation
+	provData []byte
 }
 
 // NewHTTPInstaller creates a new HttpInstaller.
@@ -114,19 +69,6 @@ func NewHTTPInstaller(source string) (*HTTPInstaller, error) {
 	return i, nil
 }
 
-// helper that relies on some sort of convention for plugin name (plugin-name-<version>)
-func stripPluginName(name string) string {
-	var strippedName string
-	for suffix := range Extractors {
-		if strings.HasSuffix(name, suffix) {
-			strippedName = strings.TrimSuffix(name, suffix)
-			break
-		}
-	}
-	re := regexp.MustCompile(`(.*)-[0-9]+\..*`)
-	return re.ReplaceAllString(strippedName, `$1`)
-}
-
 // Install downloads and extracts the tarball into the cache directory
 // and installs into the plugin directory.
 //
@@ -135,6 +77,31 @@ func (i *HTTPInstaller) Install() error {
 	pluginData, err := i.getter.Get(i.Source)
 	if err != nil {
 		return err
+	}
+
+	// Save the original tarball to plugins directory for verification
+	// Extract metadata to get the actual plugin name and version
+	pluginBytes := pluginData.Bytes()
+	metadata, err := plugin.ExtractPluginMetadataFromReader(bytes.NewReader(pluginBytes))
+	if err != nil {
+		return fmt.Errorf("failed to extract plugin metadata from tarball: %w", err)
+	}
+	filename := fmt.Sprintf("%s-%s.tgz", metadata.Name, metadata.Version)
+	tarballPath := helmpath.DataPath("plugins", filename)
+	if err := os.MkdirAll(filepath.Dir(tarballPath), 0755); err != nil {
+		return fmt.Errorf("failed to create plugins directory: %w", err)
+	}
+	if err := os.WriteFile(tarballPath, pluginBytes, 0644); err != nil {
+		return fmt.Errorf("failed to save tarball: %w", err)
+	}
+
+	// Try to download .prov file if it exists
+	provURL := i.Source + ".prov"
+	if provData, err := i.getter.Get(provURL); err == nil {
+		provPath := tarballPath + ".prov"
+		if err := os.WriteFile(provPath, provData.Bytes(), 0644); err != nil {
+			slog.Debug("failed to save provenance file", "error", err)
+		}
 	}
 
 	if err := i.extractor.Extract(pluginData, i.CacheDir); err != nil {
@@ -175,111 +142,62 @@ func (i HTTPInstaller) Path() string {
 	return helmpath.DataPath("plugins", i.PluginName)
 }
 
-// cleanJoin resolves dest as a subpath of root.
-//
-// This function runs several security checks on the path, generating an error if
-// the supplied `dest` looks suspicious or would result in dubious behavior on the
-// filesystem.
-//
-// cleanJoin assumes that any attempt by `dest` to break out of the CWD is an attempt
-// to be malicious. (If you don't care about this, use the securejoin-filepath library.)
-// It will emit an error if it detects paths that _look_ malicious, operating on the
-// assumption that we don't actually want to do anything with files that already
-// appear to be nefarious.
-//
-//   - The character `:` is considered illegal because it is a separator on UNIX and a
-//     drive designator on Windows.
-//   - The path component `..` is considered suspicions, and therefore illegal
-//   - The character \ (backslash) is treated as a path separator and is converted to /.
-//   - Beginning a path with a path separator is illegal
-//   - Rudimentary symlink protects are offered by SecureJoin.
-func cleanJoin(root, dest string) (string, error) {
-
-	// On Windows, this is a drive separator. On UNIX-like, this is the path list separator.
-	// In neither case do we want to trust a TAR that contains these.
-	if strings.Contains(dest, ":") {
-		return "", errors.New("path contains ':', which is illegal")
-	}
-
-	// The Go tar library does not convert separators for us.
-	// We assume here, as we do elsewhere, that `\\` means a Windows path.
-	dest = strings.ReplaceAll(dest, "\\", "/")
-
-	// We want to alert the user that something bad was attempted. Cleaning it
-	// is not a good practice.
-	if slices.Contains(strings.Split(dest, "/"), "..") {
-		return "", errors.New("path contains '..', which is illegal")
-	}
-
-	// If a path is absolute, the creator of the TAR is doing something shady.
-	if path.IsAbs(dest) {
-		return "", errors.New("path is absolute, which is illegal")
-	}
-
-	// SecureJoin will do some cleaning, as well as some rudimentary checking of symlinks.
-	// The directory needs to be cleaned prior to passing to SecureJoin or the location may end up
-	// being wrong or returning an error. This was introduced in v0.4.0.
-	root = filepath.Clean(root)
-	newpath, err := securejoin.SecureJoin(root, dest)
-	if err != nil {
-		return "", err
-	}
-
-	return filepath.ToSlash(newpath), nil
+// SupportsVerification returns true if the HTTP installer can verify plugins
+func (i *HTTPInstaller) SupportsVerification() bool {
+	// Only support verification for tarball URLs
+	return strings.HasSuffix(i.Source, ".tgz") || strings.HasSuffix(i.Source, ".tar.gz")
 }
 
-// Extract extracts compressed archives
-//
-// Implements Extractor.
-func (g *TarGzExtractor) Extract(buffer *bytes.Buffer, targetDir string) error {
-	uncompressedStream, err := gzip.NewReader(buffer)
+// PrepareForVerification downloads the plugin and signature files for verification
+func (i *HTTPInstaller) PrepareForVerification() (string, func(), error) {
+	if !i.SupportsVerification() {
+		return "", nil, fmt.Errorf("verification not supported for this source")
+	}
+
+	// Create temporary directory for downloads
+	tempDir, err := os.MkdirTemp("", "helm-plugin-verify-*")
 	if err != nil {
-		return err
+		return "", nil, fmt.Errorf("failed to create temp directory: %w", err)
 	}
 
-	if err := os.MkdirAll(targetDir, 0755); err != nil {
-		return err
+	cleanup := func() {
+		os.RemoveAll(tempDir)
 	}
 
-	tarReader := tar.NewReader(uncompressedStream)
-	for {
-		header, err := tarReader.Next()
-		if err == io.EOF {
-			break
-		}
-		if err != nil {
-			return err
-		}
+	// Download plugin tarball
+	pluginFile := filepath.Join(tempDir, filepath.Base(i.Source))
 
-		path, err := cleanJoin(targetDir, header.Name)
-		if err != nil {
-			return err
-		}
-
-		switch header.Typeflag {
-		case tar.TypeDir:
-			if err := os.MkdirAll(path, 0755); err != nil {
-				return err
-			}
-		case tar.TypeReg:
-			// Ensure parent directory exists
-			if err := os.MkdirAll(filepath.Dir(path), 0755); err != nil {
-				return err
-			}
-			outFile, err := os.OpenFile(path, os.O_CREATE|os.O_RDWR, os.FileMode(header.Mode))
-			if err != nil {
-				return err
-			}
-			defer outFile.Close()
-			if _, err := io.Copy(outFile, tarReader); err != nil {
-				return err
-			}
-		// We don't want to process these extension header files.
-		case tar.TypeXGlobalHeader, tar.TypeXHeader:
-			continue
-		default:
-			return fmt.Errorf("unknown type: %b in %s", header.Typeflag, header.Name)
-		}
+	g, err := getter.All(new(cli.EnvSettings)).ByScheme("http")
+	if err != nil {
+		cleanup()
+		return "", nil, err
 	}
-	return nil
+
+	data, err := g.Get(i.Source, getter.WithURL(i.Source))
+	if err != nil {
+		cleanup()
+		return "", nil, fmt.Errorf("failed to download plugin: %w", err)
+	}
+
+	if err := os.WriteFile(pluginFile, data.Bytes(), 0644); err != nil {
+		cleanup()
+		return "", nil, fmt.Errorf("failed to write plugin file: %w", err)
+	}
+
+	// Download signature file
+	provData, err := g.Get(i.Source+".prov", getter.WithURL(i.Source+".prov"))
+	if err != nil {
+		cleanup()
+		return "", nil, fmt.Errorf("failed to download plugin signature: %w", err)
+	}
+
+	if err := os.WriteFile(pluginFile+".prov", provData.Bytes(), 0644); err != nil {
+		cleanup()
+		return "", nil, fmt.Errorf("failed to write signature file: %w", err)
+	}
+
+	// Store the provenance data so we can save it after installation
+	i.provData = provData.Bytes()
+
+	return pluginFile, cleanup, nil
 }
--- a/internal/plugin/installer/http_installer.go
+++ b/internal/plugin/installer/http_installer.go
@@ -184,20 +184,15 @@ func (i *HTTPInstaller) PrepareForVerification() (string, func(), error) {
 		return "", nil, fmt.Errorf("failed to write plugin file: %w", err)
 	}
 
-	// Download signature file
-	provData, err := g.Get(i.Source+".prov", getter.WithURL(i.Source+".prov"))
-	if err != nil {
-		cleanup()
-		return "", nil, fmt.Errorf("failed to download plugin signature: %w", err)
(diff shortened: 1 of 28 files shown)


pkg/provenance/sign.go
```diff
@@ -30,9 +30,6 @@ import (
 	"golang.org/x/crypto/openpgp/clearsign" //nolint
```
jessesimpson36, line 30: I'm a little concerned about using these libraries. According to the following pages:
https://pkg.go.dev/golang.org/x/crypto/openpgp
https://github.com/golang/go/issues/44226

> Deprecated: this package is unmaintained except for security fixes. New applications should consider a more focused, modern alternative to OpenPGP for their specific task

Since around March 2021. 

If the deadline for helm v4 marks the end of backwards incompatible updates for a while, do we want to tie ourselves strictly to this deprecated library?

pkg/cmd/plugin_package.go
```diff
@@ -0,0 +1,206 @@
...
+		return err
+	}
+
+	// If signing is requested, prepare the signer first
+	var signer *provenance.Signatory
+	if o.sign {
+		// Load the signing key
+		signer, err = provenance.NewFromKeyring(o.keyring, o.key)
+		if err != nil {
+			return err
```
jessesimpson36: In my investigation of why my keyring is throwing 
```
openpgp: unsupported feature: public key type: 22
```
I added an error wrap to see if it was happening during the loading of the keyring

```suggestion
			return fmt.Errorf("Error reading from keyring: %w", err)
```
I think it'd be a good idea to add more error wrapping in this package.

pkg/provenance/sign.go
jessesimpson36: Wrote up this issue in case we want to solve it separately from this PR
https://github.com/helm/helm/issues/31180

scottrigby: thank you. will follow up there separately. should not be addressed in this PR.

pkg/cmd/plugin_package.go
```diff
@@ -0,0 +1,206 @@
...
+			return o.run(out)
+		},
+	}
+
+	f := cmd.Flags()
+	f.BoolVar(&o.sign, "sign", true, "use a PGP private key to sign this plugin")
+	f.StringVar(&o.key, "key", "", "name of the key to use when signing. Used if --sign is true")
+	f.StringVar(&o.keyring, "keyring", defaultKeyring(), "location of a public keyring")
+	f.StringVar(&o.passphraseFile, "passphrase-file", "", "location of a file which contains the passphrase for the signing key. Use \"-\" to read from stdin.")
+	f.StringVarP(&o.destination, "destination", "d", ".", "location to write the plugin tarball.")
```
jessesimpson36, line 72: I think the default being `.` is problematic. on my system, when I run
```
/home/jesse/code/helm/bin/helm plugin package --key "Helm Signing Key" --keyring ./exported-keyring.gpg  ./
```

It saves the plugin package to `..tgz.prov` and `..tgz`

```
Successfully signed. Signature written to: ..tgz.prov
Successfully packaged plugin and saved it to: ..tgz
```

makes it hard to find the package.

pkg/cmd/plugin_package.go
```diff
@@ -0,0 +1,206 @@
...
+			return err
+		}
+	} else {
+		// User explicitly disabled signing
+		fmt.Fprintf(out, "WARNING: Skipping plugin signing. This is not recommended for plugins intended for distribution.\n")
+	}
+
+	// Now create the tarball (only after signing prerequisites are met)
+	pluginName := filepath.Base(o.pluginPath)
+	tarballPath := filepath.Join(o.destination, pluginName+".tgz")
```
jessesimpson36: In this comment: https://github.com/helm/helm/pull/31176/files#r2296746809

I think me using a plugin path of `./` is what causes the path to be `..tgz`. So if I was one directory higher, I would've been fine.

jessesimpson36: I think pluginName should actually be the name from the `plugin.yaml` configuration file.

pkg/cmd/plugin_package.go
```diff
@@ -0,0 +1,206 @@
...
+
+	tarFile, err := os.Create(tarballPath)
+	if err != nil {
+		return fmt.Errorf("failed to create tarball: %w", err)
+	}
+	defer tarFile.Close()
+
+	if err := plugin.CreatePluginTarball(o.pluginPath, tarFile); err != nil {
```
jessesimpson36: Perhaps as a feature request, I think we should add the ability to have a `.helmignore` inside the plugin directory so you don't accidentally package your secret keyring or something like that

jessesimpson36 (changes requested):
See above comments. Otherwise, I'm happy.

internal/plugin/installer/local_installer.go
```diff
@@ -62,8 +99,102 @@ func (i *LocalInstaller) Install() error {
...
+	}
+
+	pluginName := filepath.Base(i.Source)
+	if i.isArchive {
+		// Strip archive extension to get plugin name
+		pluginName = stripPluginName(pluginName)
+	}
+
+	return helmpath.DataPath("plugins", pluginName)
+}
```
jessesimpson36, line 179: I feel like `Path()` should be the name of the plugin, and not the directory name. Ideally this would be the same, but in my case, I was trying to install `./` , and then it created the file `..tgz`. Which when I then tried to install the plugin from `..tgz`, it failed because `..` exists already within the ~/.cache/helm/plugins dir

scottrigby: agree. I have changed this, and also added version to the packaged plugin archive, now similar to charts, eg, example-cli-0.1.0.tgz. This is also based on plugin name, no longer cares what the directory was named before being archived.

pkg/cmd/plugin_package.go
scottrigby: this should be fixed now

pkg/cmd/plugin_package.go
scottrigby: agreed, updated to do exactly that

pkg/cmd/plugin_package.go
scottrigby: interesting idea. our boilerlate chart .helmignore doesn't exclude secrets key patterns either. but i agree could be a good feature request. given our timeline i'd suggest an issue to follow up on

scottrigby:
I made changes to 
- https://github.com/helm/helm/pull/31172

in response to review feedback. hope to have that merged today. will rebase this PR after

pkg/cmd/plugin_package.go
jessesimpson36: https://github.com/helm/helm/issues/31184

pkg/cmd/plugin_package.go
jessesimpson36: Thank you!

jessesimpson36 (approved):

internal/plugin/installer/http_installer.go
```diff
@@ -175,111 +142,62 @@ func (i HTTPInstaller) Path() string {
...
+	tempDir, err := os.MkdirTemp("", "helm-plugin-verify-*")
+	if err != nil {
+		return "", nil, fmt.Errorf("failed to create temp directory: %w", err)
+	}
 
-	// We want to alert the user that something bad was attempted. Cleaning it
-	// is not a good practice.
-	if slices.Contains(strings.Split(dest, "/"), "..") {
-		return "", errors.New("path contains '..', which is illegal")
+	cleanup := func() {
```
banjoh, line 163: Why not just `defer os.RemoveAll(tempDir)`? It looks like the caller of this API just does the same. If the call fails, any returned file is useless anyway

internal/plugin/verify.go
```diff
@@ -0,0 +1,72 @@
...
+
+	// Only support tarball verification
+	if fi.IsDir() {
+		return nil, errors.New("directory verification not supported - only plugin tarballs can be verified")
+	}
+
+	// Verify it's a tarball
+	if !isTarball(pluginPath) {
+		return nil, errors.New("plugin file must be a gzipped tarball (.tar.gz or .tgz)")
+	}
```
banjoh, line 46: Is this check necessary? Signature verification just checks if the file, regardless of the format, was signed using the provided PGP key.

pkg/cmd/plugin_install.go
```diff
@@ -113,10 +125,51 @@ func (o *pluginInstallOptions) run(out io.Writer) error {
...
+	}
+
+	// Install the plugin with options
+	verifyResult, err := installer.InstallWithOptions(i, opts)
+	if err != nil {
 		return err
 	}
 
+	// If verification was successful, show the details
+	if shouldVerify && verifyResult != nil {
```
banjoh:
```suggestion
	if verifyResult != nil {
```

Check is not necessary. If `shouldVerify == false`, `verifyResult` will always be `nil`

banjoh (commented):
Changes look good. I added a few minor comments

pkg/cmd/plugin_install.go
scottrigby: you're right, it's unnecessary. *VerificationResult without initialization will return nil anyway. I've committed your suggestion 👍

internal/plugin/installer/http_installer.go
scottrigby: The cleanup logic is a bit convoluted. The complexity exists because:

1. InstallWithOptions() calls PrepareForVerification() which writes temporary files for HTTP/OCI installers (but not Local)
2. pkg/provenance functions currently require file paths, not data
3. Only some installers need cleanup, hence the optional cleanup function pattern

The root cause is that verification currently requires files on disk. I've created a follow-up PR (#31196) that refactors pkg/provenance to accept data directly instead of file paths, eliminating the need for temporary files and this cleanup complexity entirely.

Let's merge this PR as-is to get the immediate improvements, then address the I/O elimination in the follow-up PR.

internal/plugin/verify.go
scottrigby: I agree this is overly restrictive. This function assumes plugin signing is only for tarballs (similar to chart signing), but the logic could be more robust.

However, this logic is entirely removed in the follow-up PR (#31196) as part of eliminating the file-based verification approach. Since it's a minor issue that gets completely resolved in the follow-up, let's merge this PR and address it there.

banjoh (approved):
LGTM. Ship it!

scottrigby:
Just waiting for one other core maintainer to approve

scottrigby:
@sabre1041 f4ded70 fixes the issue you found

scottrigby:
Fixed the "mismatched provenance" status aef38ea.

Now it works as expected: f you change the hash in the provenance file, or the tarball is not exactly what was signed (tampered with, wrong tarball or version, etc), the mismatched status shows correctly.

scottrigby:
test statuses like this:

1. first, install my public key. see "Manual testing" steps in the PR description above.
2. then:
```console
# install signed plugin
% go run ./cmd/helm plugin install https://github.com/scottrigby/h4-example-plugins/raw/refs/heads/main/tarballs/example-cli-0.1.0.tgz
Verifying plugin signature...
Signed by: Scott Rigby <scott@r6by.com>
Using Key With Fingerprint: 208DD36ED5BB3745A16743A4C7C6FBB5B91C1155
Plugin Hash Verified: sha256:4ef0c4b1cd402d6a170fadb14d68733226e37bff9e2df67ee34754f4a6d0d1c6
Installed plugin: example-cli

# install unsigned plugin (must pass --verify=false)
% go run ./cmd/helm plugin install --verify=false https://github.com/scottrigby/h4-example-plugins/raw/refs/heads/main/tarballs/example-unsigned-0.1.0.tgz
WARNING: Skipping plugin signature verification
Installed plugin: example-unsigned

# manually install plugin with mismatched prov
% cd $(go run ./cmd/helm env HELM_PLUGINS)
% curl -SsLO https://github.com/scottrigby/h4-example-plugins/raw/refs/heads/main/tarballs/example-mismatched-provenance-0.1.0.tgz
% curl -SsLO https://github.com/scottrigby/h4-example-plugins/raw/refs/heads/main/tarballs/example-mismatched-provenance-0.1.0.tgz.prov
% tar -zxvf example-mismatched-provenance-0.1.0.tgz 
x example-mismatched-provenance
x example-mismatched-provenance/plugin.yaml

# see statuses
% go run ./cmd/helm plugin list
NAME                         	VERSION	TYPE  	APIVERSION	PROVENANCE           	SOURCE                                          
example-cli                  	0.1.0  	cli/v1	v1        	signed               	https://github.com/scottrigby/h4-example-plugins
example-mismatched-provenance	0.1.0  	cli/v1	v1        	mismatched provenance	https://github.com/scottrigby/h4-example-plugins
example-unsigned             	0.1.0  	cli/v1	v1        	unsigned             	https://github.com/scottrigby/h4-example-plugins
```

sabre1041:
@scottrigby Performed some additional testing and from an end user experience, defaulting verification to `true` could prove troublesome. The following is an alternate approach that we might be able to consider

1. `--verify` option false by default if no `.prov` file found. This aligns with the functionality of Helm charts
2. Attempt to verify if `.prov` file found. User can set `--verify` option which they can use to disable chart verification if desired

scottrigby:
> @scottrigby Performed some additional testing and from an end user experience, defaulting verification to `true` could prove troublesome. The following is an alternate approach that we might be able to consider
> 
> 1. `--verify` option false by default if no `.prov` file found. This aligns with the functionality of Helm charts

ok i can see this. if .prov is not found, do you think it's acceptable from a user experience for helm to emit a warning rather than a failure?

> 2. Attempt to verify if `.prov` file found. User can set `--verify` option which they can use to disable chart verification if desired

ok great, that's how this part works now. can update the above if you agree about warning

sabre1041:
> > @scottrigby Performed some additional testing and from an end user experience, defaulting verification to `true` could prove troublesome. The following is an alternate approach that we might be able to consider
> > 
> > 1. `--verify` option false by default if no `.prov` file found. This aligns with the functionality of Helm charts
> 
> ok i can see this. if .prov is not found, do you think it's acceptable from a user experience for helm to emit a warning rather than a failure?
> 
> > 2. Attempt to verify if `.prov` file found. User can set `--verify` option which they can use to disable chart verification if desired
> 
> ok great, that's how this part works now. can update the above if you agree about warning

@scottrigby agree with the approach as it provides a balance of ensuring that existing plugins are unaffected functionally, but illustrates the benefits that come with signing and encourages signing plugins in the future.

scottrigby:
@sabre1041 ok, pushed 0d54097 which updates the behavior for installing a plugin with missing .prov file to now emit a warning and continue, instead of failing with an error as it previously did.

```console
% go run ./cmd/helm plugin install https://github.com/scottrigby/h4-example-plugins/raw/refs/heads/main/tarballs/example-unsigned-0.1.0.tgz               
Verifying plugin signature...
WARNING: No provenance file found for plugin. Plugin is not signed and cannot be verified.
Installed plugin: example-unsigned
```

adding missing unit tests to check this, coming next.

scottrigby:
@andy - i just added comprehensive plugin verification tests in 2adb5a890:
- Test missing .prov files (warns but continues)
- Test invalid .prov file formats (fails verification)
- Test hash mismatches in .prov files (fails verification)
- Test .prov file access errors (fails appropriately)
- Test directory plugins don't support verification
- Test installation without verification enabled (succeeds)
- Test with valid .prov files (fails on empty keyring as expected)

scottrigby:
also added remaining missing test coverage for this PR (cmd/plugin_verify CLI tests) in 5746baef3

internal/plugin/installer/http_installer.go
```diff
@@ -137,6 +79,31 @@ func (i *HTTPInstaller) Install() error {
...
 
+	// Save the original tarball to plugins directory for verification
+	// Extract metadata to get the actual plugin name and version
+	pluginBytes := pluginData.Bytes()
+	metadata, err := plugin.ExtractPluginMetadataFromReader(bytes.NewReader(pluginBytes))
+	if err != nil {
+		return fmt.Errorf("failed to extract plugin metadata from tarball: %w", err)
+	}
+	filename := fmt.Sprintf("%s-%s.tgz", metadata.Name, metadata.Version)
+	tarballPath := helmpath.DataPath("plugins", filename)
```
gjenkins8, line 90: I think we should put this into the cache dir (at this is a temporary artifact)

internal/plugin/installer/http_installer_test.go
```diff
@@ -49,7 +49,7 @@ func (t *TestHTTPGetter) Get(_ string, _ ...getter.Option) (*bytes.Buffer, error
 }
 
 // Fake plugin tarball data
-var fakePluginB64 = "H4sIAKRj51kAA+3UX0vCUBgGcC9jn+Iwuk3Peza3GeyiUlJQkcogCOzgli7dJm4TvYk+a5+k479UqquUCJ/fLs549sLO2TnvWnJa9aXnjwujYdYLovxMhsPcfnHOLdNkOXthM/IVQQYjg2yyLLJ4kXGhLp5j0z3P41tZksqxmspL3B/O+j/XtZu1y8rdYzkOZRCxduKPk53ny6Wwz/GfIIf1As8lxzGJSmoHNLJZphKHG4YpTCE0wVk3DULfpSJ3DMMqkj3P5JfMYLdX1Vr9Ie/5E5cstcdC8K04iGLX5HaJuKpWL17F0TCIBi5pf/0pjtLhun5j3f9v6r7wfnI/H0eNp9d1/5P6Gez0vzo7wsoxfrAZbTny/o9k6J8z/VkO/LPlWdC1iVpbEEcq5nmeJ13LEtmbV0k2r2PrOs9PuuNglC5rL1Y5S/syXRQmutaNw1BGnnp8Wq3UG51WvX1da3bKtZtCN/R09DwAAAAAAAAAAAAAAAAAAADAb30AoMczDwAoAAA="
+var fakePluginB64 = "H4sIAAAAAAAAA+3SQUvDMBgG4Jz7K0LwapdvSxrwJig6mCKC5xHabBaXdDSt4L+3cQ56mV42ZPg+lw+SF5LwZmXf3OV206/rMGEnIgdG6zTJaDmee4y01FOlZpqGHJGZSsb1qS401sfOtpyz0FTup9xv+2dqNep/N/IP6zdHPSMVXCh1sH8yhtGMDBUFFTL1r4iIcXnUWxzwz/sP1rsrLkbfQGTvro11E4ZlmcucRNZHu04py1OO73OVi2Vbb7td9vp7nXevtvsKRpGVjfc2VMP2xf3t4mH5tHi5mz8ub+bPk9JXIvvr5wMAAAAAAAAAAAAAAAAAAAAAnLVPqwHcXQAoAAA="
```
gjenkins8, line 52: this is a legacy plugin.yaml. intentional?

internal/plugin/installer/http_installer_test.go
```diff
@@ -49,7 +49,7 @@ func (t *TestHTTPGetter) Get(_ string, _ ...getter.Option) (*bytes.Buffer, error
...
 
 func TestStripName(t *testing.T) {
```
gjenkins8, line 54: Should move this test to extrator_test.go (as stripPluginName has moved now)

internal/plugin/sign.go
```diff
@@ -0,0 +1,166 @@
...
+	if err != nil {
+		return "", fmt.Errorf("failed to marshal plugin metadata: %w", err)
+	}
+
+	// Use the generic provenance signing function
+	return signer.ClearSign(tarballPath, metadataBytes)
+}
+
+// ExtractPluginMetadata extracts plugin metadata from a tarball
+func ExtractPluginMetadata(tarballPath string) (*Metadata, error) {
```
gjenkins8: can we make this helper private?

internal/plugin/sign.go
```diff
@@ -0,0 +1,166 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+
+	return ExtractPluginMetadataFromReader(f)
+}
+
+// ExtractPluginMetadataFromReader extracts plugin metadata from a tarball reader
+func ExtractPluginMetadataFromReader(r io.Reader) (*Metadata, error) {
```
gjenkins8, line 66:
```suggestion
func ExtractTgzPluginMetadata(r io.Reader) (*Metadata, error) {
```

internal/plugin/installer/http_installer.go
scottrigby: it's needed for whenever someone wants to get signing details with `plugin verify` at any point after the plugin is initially installed. This file, the provenance, and the unpacked plugin dir are all removed when the plugin is uninstalled.

internal/plugin/installer/http_installer_test.go
scottrigby: not specifically, but we support legacy plugins so it worked. can we change this in future when needed?

internal/plugin/sign.go
scottrigby: done in c8086d08d

internal/plugin/installer/http_installer_test.go
scottrigby: as this is minor will do this in a follow-up PR

internal/plugin/sign.go
scottrigby: will do this PR:
- https://github.com/helm/helm/pull/31196

gjenkins8 (approved):

internal/plugin/sign.go
scottrigby: added to #31196 :+1: i agree it's a more explicit name

idsulik:
@scottrigby hi! Check please this comment https://github.com/helm/helm/pull/31405#issuecomment-3484012831

PR #31185, opened 2025-08-26
fix: Improve error reporting for `helm template --debug` with `--show-only`

### LINKED ISSUE #31183: Bug: `helm template --debug` fails with "could not find template" when `--show-only` targets a template with a rendering error.
### What happened?

When using the `helm template` command with the `--show-only` flag, the behavior differs unexpectedly depending on whether the `--debug` flag is used, specifically when the target template file contains a rendering error.
With the `--debug` flag added, the command fails with an incorrect "could not find template" error instead of the rendering error.

The same error should be printed.

1. Create an empty chart.
2. Add `templates/a.yaml` (with an intentional error)
    ```yaml
    apiVersion: v1
    kind: Secret
    metadata:
      name: my-secret
    data:
      # Reference a non-existent value to trigger a template error
      some-key: {{ .Values.nonExistentKey | b64enc | quote }}
    ```
3. Run `helm template . --show-only templates/a.yaml`, check the following error
    ```console
    Error: template: repro-chart/templates/a.yaml:7:40: executing "repro-chart/templates/a.yaml" at <b64enc>: invalid value; expected string

    ```
5. Run `helm template . --show-only templates/a.yaml --debug`, check the different errors occurring
    ```console
    install.go:222: [debug] Original chart version: ""
    install.go:239: [debug] CHART PATH: /Users/kyoku/repro

    Error: could not find template templates/a.yaml in chart
    helm.go:84: [debug] could not find template templates/a.yaml in chart
    ```

### Helm version

<details>

```console
$ helm version
version.BuildInfo{Version:"v3.15.4", GitCommit:"fa9efb07d9d8debbb4306d72af76a383895aa8c4", GitTreeState:"clean", GoVersion:"go1.22.6"}
```

</details>

### Kubernetes version

<details>

```console
$ kubectl version
Client Version: v1.33.0
Kustomize Version: v5.6.0
Server Version: v1.31.6+orb1
WARNING: version difference between client (1.33) and server (1.31) exceeds the supported minor version skew of +/-1
```

</details>

kyokuping:
closes #31183

This PR fixes a bug where `helm template --debug --show-only` would show a misleading "could not find template" error when the target template had a rendering error.
Currently, the `--debug` flag suppresses the initial template rendering error and then later reports the misleading error, which hides the true cause of the failure. This change ensures that the original rendering error is captured and prioritized, providing users with accurate and actionable feedback.

The core of the fix is in the error handling logic after the templates are rendered. I've introduced a change to preserve the error from the `runInstall()` function, even in debug mode. The final error report now checks for this rendering error first before checking if the file specified in `--show-only` exists in the manifest output.

**If applicable**:

### DIFF
--- a/pkg/cmd/template.go
+++ b/pkg/cmd/template.go
@@ -186,7 +186,12 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 							missing = false
 						}
 						if missing {
-							return fmt.Errorf("could not find template %s in chart", f)
+							if err != nil && settings.Debug {
+								// assume the manifest itself is too malformed to be rendered
+								return err
+							} else {
+								return fmt.Errorf("could not find template %s in chart", f)
+							}
 						}
 					}
 					for _, m := range manifestsToRender {
--- a/pkg/cmd/template.go
+++ b/pkg/cmd/template.go
@@ -113,7 +113,7 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 				}
 				return err
 			}
-			savedErr := err
+			installErr := err
 
 			// We ignore a potential error here because, when the --debug flag was specified,
 			// we always want to print the YAML, even if it is not valid. The error is still returned afterwards.
@@ -187,9 +187,9 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 							missing = false
 						}
 						if missing {
-							if savedErr != nil && settings.Debug {
+							if installErr != nil && settings.Debug {
 								// assume the manifest itself is too malformed to be rendered
-								return savedErr
+								return installErr
 							}
 							return fmt.Errorf("could not find template %s in chart", f)
 						}
@@ -202,7 +202,7 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 				}
 			}
 
-			return savedErr
+			return installErr
 		},
 	}
 
--- a/pkg/cmd/template.go
+++ b/pkg/cmd/template.go
@@ -113,6 +113,7 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 				}
 				return err
 			}
+			savedErr := err
 
 			// We ignore a potential error here because, when the --debug flag was specified,
 			// we always want to print the YAML, even if it is not valid. The error is still returned afterwards.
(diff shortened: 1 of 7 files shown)


pkg/cmd/template.go
```diff
@@ -177,7 +177,12 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 							missing = false
 						}
 						if missing {
-							return fmt.Errorf("could not find template %s in chart", f)
+							if err != nil && settings.Debug {
```
TerryHowe: This would probably clearer and more reliable if you saved the error above like `saveError := err` and then no need to check settings.Debug. I get this works now, but someone might come along and not realize that err gets used way down here later.

TerryHowe (changes requested):

pkg/cmd/template.go
kyokuping: Done!

TerryHowe (approved):
/lgtm

Thanks!

victoravt:
Hey, waiting for this 🫡

TerryHowe (dismissed):
Please fix the lint error.

pkg/cmd/template.go
```diff
@@ -187,6 +188,10 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 							missing = false
 						}
 						if missing {
+							if savedErr != nil && settings.Debug {
+								// assume the manifest itself is too malformed to be rendered
+								return savedErr
```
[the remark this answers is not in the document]
gjenkins8: agree with the above

pkg/cmd/template.go
```diff
@@ -187,6 +188,10 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 							missing = false
 						}
 						if missing {
+							if savedErr != nil && settings.Debug {
+								// assume the manifest itself is too malformed to be rendered
```
TerryHowe, line 191:
```suggestion
								// render error caused --show-only output to be empty; return the original error
```

TerryHowe (dismissed):

TerryHowe (dismissed):
/lgtm

kyokuping:
@TerryHowe Could you please approve the CI run so we can proceed with merging?

pkg/cmd/template.go
```diff
@@ -113,6 +113,7 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 				}
 				return err
 			}
+			savedErr := err
```
gjenkins8:
```suggestion
			installErr := err
```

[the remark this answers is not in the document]
gjenkins8: also agree with this

gjenkins8 (changes requested):
https://github.com/helm/helm/pull/31185/changes#r3176880472
https://github.com/helm/helm/pull/31185/changes#r3253045479

please

gjenkins8 (approved):

TerryHowe (approved):
/lgtm

PR #31194, opened 2025-08-27
[HIP-0026] Plugin extism/v1 runtime

gjenkins8:

Implements an "extism/v1" plugin runtime, enabling plugins to be written as ([extism](https://extism.org/docs/concepts/host-sdk) based) Wasm modules.

[internal/plugin/testdata/src/extismv1-test/main.go](https://github.com/helm/helm/compare/main...gjenkins8:gjenkins/plugin-integration/wasm_runtime?expand=1#diff-b4be36490eb9cff06be89b1fe5aa0d7f994b35e8229f183f1ff7947668e376d9R1) shows a basic Wasm plugin

(and https://github.com/helm/helm/compare/main...gjenkins8:gjenkins/plugin-integration/wasm_runtime?expand=1#diff-b550ca624f8f180334db80471e7f61b10a468c0b1633015447b5e5fa1ff0eceeR49 builds this at unit test time 😱 )

**If applicable**:

### DIFF
--- a/internal/plugin/runtime_extismv1.go
+++ b/internal/plugin/runtime_extismv1.go
@@ -1,0 +1,297 @@
+/*
+Copyright The Helm Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package plugin
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"log/slog"
+	"os"
+	"path/filepath"
+	"reflect"
+
+	extism "github.com/extism/go-sdk"
+	"github.com/tetratelabs/wazero"
+)
+
+const ExtistmV1WasmBinaryFilename = "plugin.wasm"
+
+type RuntimeConfigExtismV1Memory struct {
+	// The max amount of pages the plugin can allocate
+	// One page is 64Kib. e.g. 16 pages would require 1MiB.
+	// Default is 4 pages (256KiB)
+	MaxPages uint32 `yaml:"maxPages,omitempty"`
+
+	// The max size of an Extism HTTP response in bytes
+	// Default is 4096 bytes (4KiB)
+	MaxHTTPResponseBytes int64 `yaml:"maxHttpResponseBytes,omitempty"`
+
+	// The max size of all Extism vars in bytes
+	// Default is 4096 bytes (4KiB)
+	MaxVarBytes int64 `yaml:"maxVarBytes,omitempty"`
+}
+
+type RuntimeConfigExtismV1FileSystem struct {
+	// If specified, a temporary directory will be created and mapped to /tmp in the plugin's filesystem.
+	// Data written to the directory will be visible on the host filesystem.
+	// The directory will be removed when the plugin invocation completes.
+	CreateTempDir bool `yaml:"createTempDir,omitempty"`
+
+	// // An optional set of mappings between the host's filesystem and the paths a plugin can access.
+	// TODO: shuld Helm expose this?
+	//AllowedPaths map[string]string `yaml:"allowedPaths,omitempty"`
+}
+
+// RuntimeConfigExtismV1 defines the user-configurable options the plugin's Extism runtime
+// The format loosely follows the Extism Manifest format: https://extism.org/docs/concepts/manifest/
+type RuntimeConfigExtismV1 struct {
+	// Describes the limits on the memory the plugin may be allocated.
+	Memory RuntimeConfigExtismV1Memory `yaml:"memory"`
+
+	// The "config" key is a free-form map that can be passed to the plugin.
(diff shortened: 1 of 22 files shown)


internal/plugin/runtime_extismv1.go
```diff
@@ -0,0 +1,297 @@
...
+
+type RuntimeConfigExtismV1FileSystem struct {
+	// If specified, a temporary directory will be created and mapped to /tmp in the plugin's filesystem.
+	// Data written to the directory will be visible on the host filesystem.
+	// The directory will be removed when the plugin invocation completes.
+	CreateTempDir bool `yaml:"createTempDir,omitempty"`
+
+	// // An optional set of mappings between the host's filesystem and the paths a plugin can access.
+	// TODO: shuld Helm expose this?
+	//AllowedPaths map[string]string `yaml:"allowedPaths,omitempty"`
```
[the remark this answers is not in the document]
scottrigby: we can move the TODO with these same questions to a comment above the type

internal/plugin/runtime_extismv1.go
```diff
@@ -0,0 +1,297 @@
...
+
+	// The name of entry function name to call in the plugin
+	// Defaults to "helm_plugin_main".
+	EntryFuncName string `yaml:"entryFuncName,omitempty"`
+}
+
+var _ RuntimeConfig = (*RuntimeConfigExtismV1)(nil)
+
+func (r *RuntimeConfigExtismV1) Validate() error {
+	// TODO
```
scottrigby, line 90: I think this is fine to wait on. The underlying tooling will give errors that we'll surface when configuration options and combinations are invalid. We'll know after more use which of these we want to give friendlier/more contextual messages for better DX.

internal/plugin/runtime_extismv1.go
```diff
@@ -0,0 +1,297 @@
...
+			MaxVarBytes:          rc.Memory.MaxVarBytes,
+		},
+		Config:       rc.Config,
+		AllowedHosts: allowedHosts,
+		AllowedPaths: allowedPaths,
+		Timeout:      rc.Timeout,
+	}, nil
+}
+
+func buildPluginConfig(input *Input, r *RuntimeExtismV1) (extism.PluginConfig, error) {
```
scottrigby: should we remove `error` return from the func signature, since none of these specific wazero functions return errors?

scottrigby (commented):
just some comments for now. I'm aware of the structure and this is looking really good. am flying in the morning and it's late so need to do more testing, but wanted to start with the few comments I had for now.

internal/plugin/runtime_extismv1.go
scottrigby: done

internal/plugin/runtime_extismv1.go
```diff
@@ -0,0 +1,292 @@
...
+func (p *ExtismV1PluginRuntime) Dir() string {
+	return p.dir
+}
+
+func (p *ExtismV1PluginRuntime) Invoke(ctx context.Context, input *Input) (*Output, error) {
+
+	var tmpDir string
+	if p.rc.FileSystem.CreateTempDir {
+		tmpDirInner, err := os.MkdirTemp(os.TempDir(), "helm-plugin-*")
+		slog.Debug("created plugin temp dir", slog.String("dir", tmpDir), slog.String("plugin", p.metadata.Name))
```
scottrigby: quick note looks like `tmpDir` will be null at this point.
I feel the pain of this scoped defer cleanup stuff. Understand we do need filesystem write here because that's the point of the `CreateTempDir` option for Wasm targeted filesystem access. Anyway, this is the last issue I see… a standard test Wasm getter plugin is working perfectly!

scottrigby (dismissed):
This looks and works great. Only one issue noted about the tmpdir cleanup. Once that's resolved, this has my big LGTM 👍 (marking approved even though I know GitHub will dismiss my approval when you fix that. Doing so because I've thoroughly reviewed and tested this - we can always fix any bugs we find during pre-release hardening)

internal/plugin/runtime_extismv1.go
```diff
@@ -0,0 +1,292 @@
...
+	MaxVarBytes int64 `yaml:"maxVarBytes,omitempty"`
+}
+
+// RuntimeConfigExtismV1FileSystem exposes filesystem options for the configuration
+// TODO: should Helm expose AllowedPaths?
+type RuntimeConfigExtismV1FileSystem struct {
+	// If specified, a temporary directory will be created and mapped to /tmp in the plugin's filesystem.
+	// Data written to the directory will be visible on the host filesystem.
+	// The directory will be removed when the plugin invocation completes.
+	CreateTempDir bool `yaml:"createTempDir,omitempty"`
```
scottrigby, line 55: Note that SDK users that do not allow filesystem writes can, if they wish, check for this config and disallow a plugin that sets this because it requires a temp directory on the filesystem. They are both legitimate requirements, even if incompatible. This is much more well structured and detectable than the subprocess runtime (which can do nearly anything the running user can do). Nice 💯 

a controller can check for this if they wish and say "we don't support plugins that require filesystem access". which i think would be legit. We may even want to mention that explicitly in the docs.

internal/plugin/runtime_extismv1.go
gjenkins8: oops, right (restructured this code late, and missed this). Fixed

scottrigby: fix looks good!

scottrigby (approved):
tempdir cleanup bit has been fixed. Nice work :+1: 🚀

PR #31195, opened 2025-08-27
Include Jobs in Wait

mattfarina:
Feedback requested on this change. Should the waiting for jobs be merged into the wait for other things? There are 2 ways to look at this:

1. All waits should be together behind one flag
2. Jobs can run for a long time. Merging into --wait means the wait doesn't complete until the Job is done. So, they should be separate.

Thoughts?

When --wait-for-jobs was introduced it was added as a new flag so the behavior on --wait didn't change in a minor version. This happened in #8363. In a major version of Helm we can merge these into a single behavior behind the --wait flag.

**If applicable**:

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -85,7 +85,6 @@ type Install struct {
 	DisableHooks     bool
 	Replace          bool
 	WaitStrategy     kube.WaitStrategy
-	WaitForJobs      bool
 	Devel            bool
 	DependencyUpdate bool
 	Timeout          time.Duration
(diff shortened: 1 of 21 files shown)


scottrigby (commented):
feedback: --wait-for-jobs as a separate flag also enables a use case where you want to ensure the chart's resources installed successfully, but want to exclude jobs because your chart includes jobs that take a long time to complete. Here is a real world example https://buildkite.com/docs/agent/v3/agent-stack-k8s/long-running-jobs.

I don't know how widely used the flag is for that use case though, and could see the argument that we combine both into one --wait flag, and charts which mix loger-running tasks (that they don't want the CLI to wait on) with other resources (that they do want to wait on) could move those into separate charts and just not run --wait on chart with longer running tasks. Definitely a nice cleanup if we want to go that route.

godhanipayal:
FWIW, lot of OCI customers like to use --wait flag and --wait-for-jobs as they serve separate purposes

Issue #13447, opened 2024-11-14
multiline stacktraces for error messages in helper functions


jessesimpson36:
Hey all,

I work on a pretty large helm chart. Whenever I encounter an issue,  theres a long oneline error message that's pretty hard to read.  one such example is described by this repository (https://github.com/jessesimpson36/helm_error_format)

> Error: INSTALLATION FAILED: template: camunda-platform/templates/zeebe-gateway/deployment.yaml:90:24: executing "camunda-platform/templates/zeebe-gateway/deployment.yaml" at <include "camundaPlatform.authIssuerBackendUrl" .>: error calling include: template: camunda-platform/templates/camunda/_helpers.tpl:197:10: executing "camundaPlatform.authIssuerBackendUrl" at <include "identity.keycloak.url" .>: error calling include: template: camunda-platform/templates/identity/_helpers.tpl:194:9: executing "identity.keycloak.url" at <include "identity.keycloak.port" .>: error calling include: template: camunda-platform/templates/identity/_helpers.tpl:174:23: executing "identity.keycloak.port" at <.Values.identityKeycloak.service.ports>: wrong type for value; expected map[string]interface {}; got interface {}

I built an [awk script](https://github.com/jessesimpson36/helm_error_format/blob/main/format_error.awk) that would format this to be the following:

And this awk script would output the following:

```
 Error: INSTALLATION FAILED:
   template: camunda-platform/templates/zeebe-gateway/deployment.yaml:90:24:
     executing "camunda-platform/templates/zeebe-gateway/deployment.yaml"
      at <include "camundaPlatform.authIssuerBackendUrl" .>:
 error calling include:
   template: camunda-platform/templates/camunda/_helpers.tpl:197:10:
     executing "camundaPlatform.authIssuerBackendUrl"
      at <include "identity.keycloak.url" .>:
 error calling include:
   template: camunda-platform/templates/identity/_helpers.tpl:194:9:
     executing "identity.keycloak.url"
      at <include "identity.keycloak.port" .>:
 error calling include:
   template: camunda-platform/templates/identity/_helpers.tpl:174:23:
     executing "identity.keycloak.port"
      at <.Values.identityKeycloak.service.ports>:
 wrong type for value; expected map[string]interface {}; got interface {}
```

I would like this nicer form of error message to be supported directly in helm without my AWK script.

[...]

PR #31198, opened 2025-08-27
refactor: replace pkg/engine regular expressions with parser

### LINKED ISSUE #31192: TestAlterFuncMap_include times out after 10m
### What happened?

TestAlterFuncMap_include times out after 10m when you run `make test` 

```
helm.sh/helm/v4/pkg/engine.TestAlterFuncMap_include(0xc000414c40)
	/test/pkg/engine/engine_test.go:842 +0xa04
```

It to pass

`make test`

### Helm version

<details>
```console
$ helm version
version.BuildInfo{Version:"v4.0+unreleased", GitCommit:"e3124e488f5566751f0b4c4b224580b06569eb97", GitTreeState:"dirty", GoVersion:"go1.24.5"}
```
</details>

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

TerryHowe:
My junky MacBook does not have the memory or cpu to handle this regular expression parsing and the tests time out after 10m. With this change the entire pkg/engine tests run in less than 3 seconds.

Closes: https://github.com/helm/helm/issues/31192

**If applicable**:

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -34,18 +34,6 @@ import (
 	chartutil "helm.sh/helm/v4/pkg/chart/v2/util"
 )
 
-// taken from https://cs.opensource.google/go/go/+/refs/tags/go1.23.6:src/text/template/exec.go;l=141
-// > "template: %s: executing %q at <%s>: %s"
-var execErrFmt = regexp.MustCompile(`^template: (?P<templateName>(?U).+): executing (?P<functionName>(?U).+) at (?P<location>(?U).+): (?P<errMsg>(?U).+)(?P<nextErr>( template:.*)?)$`)
-
-// taken from https://cs.opensource.google/go/go/+/refs/tags/go1.23.6:src/text/template/exec.go;l=138
-// > "template: %s: %s"
-var execErrFmtWithoutTemplate = regexp.MustCompile(`^template: (?P<templateName>(?U).+): (?P<errMsg>.*)(?P<nextErr>( template:.*)?)$`)
-
-// taken from https://cs.opensource.google/go/go/+/refs/tags/go1.23.6:src/text/template/exec.go;l=191
-// > "template: no template %q associated with template %q"
-var execErrNoTemplateAssociated = regexp.MustCompile(`^template: no template (?P<location>.*) associated with template (?P<functionName>(.*)?)$`)
-
 // Engine is an implementation of the Helm rendering implementation for templates.
 type Engine struct {
 	// If strict is enabled, template rendering will fail if a template references
@@ -338,7 +326,7 @@ func cleanupParseError(filename string, err error) error {
 	location := tokens[1]
 	// The remaining tokens make up a stacktrace-like chain, ending with the relevant error
 	errMsg := tokens[len(tokens)-1]
-	return fmt.Errorf("parse error at (%s): %s", string(location), errMsg)
+	return fmt.Errorf("parse error at (%s): %s", location, errMsg)
 }
 
 type TraceableError struct {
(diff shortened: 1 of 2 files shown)


TerryHowe:
@ jessesimpson36 if you get a chance, please take a look. Thanks!

jessesimpson36:
I went through each of the charts in my helm_error_charts repo (https://github.com/jessesimpson36/helm_error_charts) on a build of this PR branch to confirm that each error that was formatted before  still gets formatted now, and it does. 

And the unit test that I added in #13447  still passes. So without looking too much at the code implementation, it looks functional. 

Overall, I think the idea of switching back to for-loops and string manipulations for the sake of performance given the issues you're seeing makes sense, however, I'm very surprised to hear that these tests would take 10 minutes over regex searches, even if it was doing 1000 of them.  I assume the test that is going slow is the `TestRenderRecursionLimit` test.

Would you be able to run this test in isolation to confirm that this is the one that is taking majority of the 10 minutes? And perhaps run strace or dtruss to see what's happening on a system level (like is it swapping memory in and out of disk?)

I'll be away for about 4-5 days, and after that I can look harder at the implementation.

jessesimpson36:
time before and after (my systems biased since my machine has more memory):

Before PR branch
```
time make test
...
make test  894.41s user 21.46s system 148% cpu 10:17.30 total
```

After PR branch
```
time make test
...
make test  564.37s user 58.97s system 349% cpu 2:58.59 total
```

Okay, I believe you. Definitely not just a "someones machines slow" thing.

update: whats odd is running the tests multiple times doesn't yield the same timings. even if I clear out cache or built binaries

jessesimpson36:
I'm starting to think a test somewhere is causing a deadlock. Otherwise, I think the timings would be more deterministic

jessesimpson36:
I ran the tests 10 times to get some data before and after this test to see if we run into the 10 minute testing timeout scenario, but I couldn't reproduce that aside from my first test run. 

Using this data, I do see noticeable performance improvements after this patch. roughly 1 minute saved.

```
> sqlite3 test_benchmark.db 'select commit_hash,real_time_seconds,user_time_seconds,sys_time_seconds,exit_code from test_runs;'
-- Loading resources from /home/jesse/.sqliterc
|               commit_hash                | real_time_seconds | user_time_seconds | sys_time_seconds | exit_code |
|------------------------------------------|-------------------|-------------------|------------------|-----------|
| 934f761e08988df730c786ba82831457ced11de6 | 48.73             | 115.35            | 13.4             | 0         |
| 934f761e08988df730c786ba82831457ced11de6 | 47.81             | 129.12            | 14.62            | 0         |
| 934f761e08988df730c786ba82831457ced11de6 | 45.72             | 120.15            | 13.73            | 2         |
| 934f761e08988df730c786ba82831457ced11de6 | 48.86             | 129.24            | 14.83            | 0         |
| 934f761e08988df730c786ba82831457ced11de6 | 47.87             | 120.84            | 13.85            | 0         |
| 934f761e08988df730c786ba82831457ced11de6 | 46.07             | 120.72            | 13.47            | 2         |
| 934f761e08988df730c786ba82831457ced11de6 | 48.06             | 120.76            | 13.82            | 0         |
| 934f761e08988df730c786ba82831457ced11de6 | 47.78             | 113.64            | 13.23            | 0         |
| 934f761e08988df730c786ba82831457ced11de6 | 51.33             | 125.11            | 13.99            | 0         |
| 934f761e08988df730c786ba82831457ced11de6 | 47.52             | 112.88            | 13.16            | 0         |

commit after patch

| 284bd980b67f2be59b989abcc30ed2549936f2a6 | 37.58             | 75.37             | 13.21            | 2         |
| 284bd980b67f2be59b989abcc30ed2549936f2a6 | 38.47             | 75.88             | 13.01            | 0         |
| 284bd980b67f2be59b989abcc30ed2549936f2a6 | 38.52             | 82.42             | 13.52            | 0         |
| 284bd980b67f2be59b989abcc30ed2549936f2a6 | 38.46             | 82.96             | 13.67            | 0         |
| 284bd980b67f2be59b989abcc30ed2549936f2a6 | 38.4              | 75.78             | 13.11            | 0         |
| 284bd980b67f2be59b989abcc30ed2549936f2a6 | 38.48             | 82.31             | 13.95            | 0         |
| 284bd980b67f2be59b989abcc30ed2549936f2a6 | 38.46             | 75.69             | 12.94            | 0         |
| 284bd980b67f2be59b989abcc30ed2549936f2a6 | 38.43             | 90.73             | 15.07            | 0         |
| 284bd980b67f2be59b989abcc30ed2549936f2a6 | 38.43             | 74.89             | 13.2             | 0         |
| 284bd980b67f2be59b989abcc30ed2549936f2a6 | 38.43             | 75.13             | 13.26            | 0         |
```

jessesimpson36:
I read the diff and based on what I read I do believe it's functionally equivalent to the regex logic.

I'm thinking of scenarios that might prompt someone to change these lines in the future, and potentially make errors as they do so.  

So : 
1. go's text/template library updates to have a new error line returned not covered by the string manipulations or regex's.  In the regex implementation, you'd add a new regex with a comment of the source code in text/template, and in this implementation, you'd be rocking array+string logic that might be 20 lines down from the comment representing the error line format.
2. An author on a completely irrelevant PR decides to do a drive-by refactor on this block because someone gets confused looking at the `[index:]` and thinks it should be `[index+1:]`. Silly things like that are still possible with the regex implementation, but I think it'd be more easy to spot during PR review.

On the flip side, it's much harder to use debuggers on regex libraries.

jessesimpson36:
Maybe the trick is to use function calls representing each error line type and the function itself serves to parse it.

```
	// Executing form: "<templateName>: executing \"<funcName>\" at <<location>>: <errMsg>[ template:...]"
	// Matches https://cs.opensource.google/go/go/+/refs/tags/go1.23.6:src/text/template/exec.go;l=141

	traceableErr := ParseExecutingFuncAtLocation()
```

and then the different error line formats would be 3 different function calls that way the reviewer always see what the string is trying to parse. does that make sense?

TerryHowe:
Yes, it is `TestRenderRecursionLimit` it isn't deadlocked. I ran it in the debugger and I can see it chugging along, but just 16G mac air running lots of things.

I'm not sure why fork PR didn't run build-test.

jessesimpson36 (approved):
In summary of my previous messages:

I'm not entirely confident that the really long test runtimes in pkg/engine are hardware related, or if they are, there could be 2 separate problems,
1. The low memory environments that Terry is running into
2. a test-shuffling-related deadlock that I may have ran into (which would be pretty hard to diagnose)

As a higher priority than tracing down a possible deadlock, we gotta unblock Terry so he can do more wonderful things for us :D 

Code looks good, and with my recent tests seem equivalent in functionality.
Might be less pretty than before, but optimization almost always has that tradeoff. (And we now have tests for error string parsing, which helps if we find edge-cases in error parsing in the future)

scottrigby (dismissed):
have reviewed, tested, and considered all this. I agree with Jesse it's an overall improvement. Trade off is we have to maintain the logic now, but it's a performance improvement over the regex approach.

hat tip to https://github.com/helm/helm/pull/13586 though. all good steps

pkg/engine/engine.go
```diff
@@ -350,25 +338,128 @@ type TraceableError struct {
...
+	if !strings.HasPrefix(s, prefix) {
+		return TraceableError{}, false
+	}
+	remainder := s[len(prefix):]
+
+	// Special case: "template: no template %q associated with template %q"
+	// Matches https://cs.opensource.google/go/go/+/refs/tags/go1.23.6:src/text/template/exec.go;l=191
+	traceableError, done := parseTemplateNoTemplateError(s, remainder)
+	if done {
+		return traceableError, done
```
gjenkins8: nitpick: it is an extra mental jump to infer that done must be true, so this is the equivalent:
```suggestion
		return traceableError, true
```

pkg/engine/engine.go
```diff
@@ -350,25 +338,128 @@ type TraceableError struct {
...
+	traceableError, done := parseTemplateNoTemplateError(s, remainder)
+	if done {
+		return traceableError, true
+	}
+
+	// Executing form: "<templateName>: executing \"<funcName>\" at <<location>>: <errMsg>[ template:...]"
+	// Matches https://cs.opensource.google/go/go/+/refs/tags/go1.23.6:src/text/template/exec.go;l=141
+	traceableError, done = parseTemplateExecutingAtErrorType(remainder)
+	if done {
+		return traceableError, done
```
gjenkins8:
```suggestion
		return traceableError, true
```

pkg/engine/engine.go
```diff
@@ -350,25 +338,128 @@ type TraceableError struct {
...
+	}
+
+	// Simple form: "<templateName>: <errMsg>"
+	// Use LastIndex to avoid splitting colons within line:col info.
+	// Matches https://cs.opensource.google/go/go/+/refs/tags/go1.23.6:src/text/template/exec.go;l=138
+	traceableError, done = parseTemplateSimpleErrorString(remainder)
+	if done {
+		return traceableError, done
```
gjenkins8:
```suggestion
		return traceableError, true
```

(I was going to approve an merge either way, but since we fixed the first, can we please fix these two as well)

gjenkins8 (approved):

mattfarina (approved):
This change ran the engine package tests in 1/38 the time on my machine.

lgtm

jessesimpson36:
> We could remove this logic, if we chased on/got this implemented: [golang/go#71801](https://github.com/golang/go/issues/71801) (good for someone who wants the brownie points for getting a commit into golang core :) ) (ref: [#13586 (comment)](https://github.com/helm/helm/pull/13586#issuecomment-2663613086) )

the implementation of that one is pretty easy. the trouble is moreso that golang core has their own governance process that this issue needs to be accepted or denied as a proposal first.

https://github.com/golang/proposal?tab=readme-ov-file#the-proposal-process

https://go.dev/doc/contribute#design

they might be willing to break the process and accept a PR  for trivial changes, but I'm not sure.

PR #31199, opened 2025-08-27
fix: flaky registry data race on mockdns close

TerryHowe:
This failure just shows up every once in a while where there is a DATA RACE on closing mockdns. I tried a couple things before, but I think this makes the most sense.  Passes for me after running multiple times.

**If applicable**:


robertsirc (approved):
LGTM

sabre1041 (approved):
LGTM

PR #31200, opened 2025-08-27
fix: installer action goroutine count

TerryHowe:
The number of go threads running varies a lot when shuffling, so it is better if the installer counts the subroutines.

**If applicable**:

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -30,6 +30,7 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"text/template"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pkg/action/install.go
```diff
@@ -458,6 +462,11 @@ func (i *Install) performInstallCtx(ctx context.Context, rel *release.Release, t
 	}
 }
 
+// GetGoroutineCount return the number of running routines
+func (i *Install) GetGoroutineCount() int32 {
```
mattfarina: Should this be a public function that's part of the SDK?

TerryHowe: No, fixed. Thanks

robertsirc (commented):
LGTM

sabre1041 (approved):
LGTM

scottrigby (approved):

PR #31201, opened 2025-08-27
fix: oci push test chart_read_error broken when run in container as root

TerryHowe:

The ocipusher chart_read_test fails when I run the tests in a container as root.

**If applicable**:

### DIFF
--- a/pkg/pusher/ocipusher_test.go
+++ b/pkg/pusher/ocipusher_test.go
@@ -355,6 +355,13 @@ func TestOCIPusher_Push_ChartOperations(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
+			// Permission-related behavior can differ when running as root.
+			// The "chart read error" case relies on chmod 0000 causing a read failure,
+			// which root may bypass in some environments (e.g., Ubuntu containers).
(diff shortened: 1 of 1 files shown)


pkg/pusher/ocipusher_test.go
```diff
@@ -256,7 +256,6 @@ func TestOCIPusher_newRegistryClient(t *testing.T) {
 
 func TestOCIPusher_Push_ChartOperations(t *testing.T) {
 	// Path to test charts
-	chartPath := "../../pkg/cmd/testdata/testcharts/compressedchart-0.1.0.tgz"
```
gjenkins8: > On a mac, the test temporary directory ignores file permissions, so this test will not work. 

Which temp directory? Would a `ensure.HelmHome(t)` fix?

gjenkins8: > On a mac, the test temporary directory ignores file permissions, so this test will not work. 

Which temp directory? Would a `ensure.HelmHome(t)` fix?

TerryHowe: I've taken another approach.

TerryHowe: I've taken another approach.

PR #31206, opened 2025-08-27
pull: log resolved version and digest for repo/non-OCI pulls

benoittgt:

Fixes: https://github.com/helm/helm/issues/31205

When users run `helm pull` without a version, OCI pulls already print “Pulled:” and “Digest:”. HTTP/repo pulls do not. Some providers (e.g. Bitnami) resolve --repo pulls to `oci://` but skip the registry client summary, so nothing is printed.

- Now we print a summary for `--repo` and direct HTTP pulls after download.
- Keep direct oci:// behavior unchanged (avoid duplicate output).
- Mirror OCI format: 
    ```
    Pulled: <chart>:<version>
    Digest: sha256:<digest>
    ```

For HTTP/repo, the digest is the archive SHA-256 (.tgz). For OCI, the digest is the manifest digest (unchanged).

I tested with command like:

```sh
helm pull ingress-nginx --repo=https://kubernetes.github.io/ingress-nginx --version 4.10.1
helm pull oci://registry-1.docker.io/bitnamicharts/nginx --version '20.0.7'
helm/bin/helm pull redis --repo=https://charts.bitnami.com/bitnami --version='22.0.4'
```

Overall, I think the solution a little bit _hacky_ but interested with your feedbacks.

See inline pull request comments.

**If applicable**:

### DIFF
--- a/pkg/action/pull.go
+++ b/pkg/action/pull.go
@@ -19,6 +19,7 @@ package action
 import (
 	"crypto/sha256"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"strings"
@@ -144,17 +145,25 @@ func (p *Pull) Run(chartRef string) (string, error) {
 	// path does not go through the registry client's summary printer, so we emit
 	// a consistent summary here. We mirror the OCI output format:
 	//
-	//   Pulled: <chart>:<version>
+	//   Pulled: <chartUrl>:<version>
 	//   Digest: sha256:<digest>
 	//
 	// For HTTP/repo pulls, the digest is the SHA-256 of the downloaded .tgz archive.
 	// For direct OCI pulls, the registry client already prints "Pulled:" and
 	// "Digest:" (manifest digest), so we do not print here to avoid duplicates.
 	if p.RepoURL != "" || !registry.IsOCI(downloadSourceRef) {
-		fmt.Fprintf(&out, "Pulled: %s\n", downloadSourceRef)
+		base := trimAnySuffix(downloadSourceRef, ".tar.gz", ".tgz")
+		chart, ver := splitChartNameVersion(base)
 
-		if f, err := os.ReadFile(saved); err == nil {
-			sum := sha256.Sum256(f)
+		tag := chart
+		if ver != "" {
+			tag += ":" + ver
+		} else if p.Version != "" {
+			tag += ":" + p.Version
+		}
+		fmt.Fprintf(&out, "Pulled: %s\n", tag)
+
+		if sum, err := sha256File(saved); err == nil {
 			fmt.Fprintf(&out, "Digest: sha256:%x\n", sum)
 		}
 	}
@@ -193,4 +202,34 @@ func (p *Pull) Run(chartRef string) (string, error) {
 		return out.String(), chartutil.ExpandFile(ud, saved)
 	}
 	return out.String(), nil
+}
+
+func trimAnySuffix(s string, suffixes ...string) string {
+	for _, suf := range suffixes {
+		if strings.HasSuffix(s, suf) {
+			return strings.TrimSuffix(s, suf)
+		}
+	}
+	return s
+}
+
+func splitChartNameVersion(s string) (name, version string) {
+	if i := strings.LastIndex(s, "-"); i > 0 {
+		return s[:i], s[i+1:]
+	}
+	return s, ""
+}
+
+func sha256File(path string) ([]byte, error) {
+	f, err := os.Open(path)
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+
+	h := sha256.New()
+	if _, err := io.Copy(h, f); err != nil {
+		return nil, err
+	}
+	return h.Sum(nil), nil
 }
--- a/pkg/action/pull.go
+++ b/pkg/action/pull.go
@@ -17,6 +17,7 @@ limitations under the License.
 package action
 
 import (
+	"crypto/sha256"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -136,6 +137,37 @@ func (p *Pull) Run(chartRef string) (string, error) {
 	saved, v, err := c.DownloadTo(downloadSourceRef, p.Version, dest)
 	if err != nil {
 		return out.String(), err
+	}
+
+	// Print a pull summary for repo mode and non-OCI downloads.
+	// Some repos (e.g., Bitnami) resolve --repo pulls to oci:// refs, but that
+	// path does not go through the registry client's summary printer, so we emit
+	// a consistent summary here. We mirror the OCI output format:
+	//
+	//   Pulled: <chart>:<version>
+	//   Digest: sha256:<digest>
+	//
+	// For HTTP/repo pulls, the digest is the SHA-256 of the downloaded .tgz archive.
+	// For direct OCI pulls, the registry client already prints "Pulled:" and
+	// "Digest:" (manifest digest), so we do not print here to avoid duplicates.
+	if p.RepoURL != "" || !registry.IsOCI(downloadSourceRef) {
+		base := strings.TrimSuffix(filepath.Base(saved), ".tgz")
+		chart := base
+		ver := ""
+		if i := strings.LastIndex(base, "-"); i > 0 {
+			chart = base[:i]
+			ver = base[i+1:]
+		}
+		if ver != "" {
+			fmt.Fprintf(&out, "Pulled: %s:%s\n", chart, ver)
+		} else {
+			fmt.Fprintf(&out, "Pulled: %s\n", chart)
+		}
+
+		if f, err := os.ReadFile(saved); err == nil {
+			sum := sha256.Sum256(f)
+			fmt.Fprintf(&out, "Digest: sha256:%x\n", sum)
+		}
 	}
 
 	if p.Verify {
--- a/pkg/action/pull.go
+++ b/pkg/action/pull.go
@@ -214,7 +214,7 @@ func trimAnySuffix(s string, suffixes ...string) string {
 }
 
 func splitChartNameVersion(s string) (name, version string) {
-	if i := strings.LastIndex(s, "-"); i > 0 {
+	if i := strings.LastIndex(s, "-"); i >= 0 {
 		return s[:i], s[i+1:]
 	}
 	return s, ""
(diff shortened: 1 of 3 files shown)


pkg/action/pull.go
```diff
@@ -138,6 +139,37 @@ func (p *Pull) Run(chartRef string) (string, error) {
...
+	// path does not go through the registry client's summary printer, so we emit
+	// a consistent summary here. We mirror the OCI output format:
+	//
+	//   Pulled: <chart>:<version>
+	//   Digest: sha256:<digest>
+	//
+	// For HTTP/repo pulls, the digest is the SHA-256 of the downloaded .tgz archive.
+	// For direct OCI pulls, the registry client already prints "Pulled:" and
+	// "Digest:" (manifest digest), so we do not print here to avoid duplicates.
+	if p.RepoURL != "" || !registry.IsOCI(downloadSourceRef) {
```
benoittgt, line 154: Not a big fan of this since it feels a bit hacky, but with OCI auto-resolving like Bitnami I was short on alternatives.

pkg/action/pull.go
```diff
@@ -138,6 +139,37 @@ func (p *Pull) Run(chartRef string) (string, error) {
...
+		base := strings.TrimSuffix(filepath.Base(saved), ".tgz")
+		chart := base
+		ver := ""
+		if i := strings.LastIndex(base, "-"); i > 0 {
+			chart = base[:i]
+			ver = base[i+1:]
+		}
+		if ver != "" {
+			fmt.Fprintf(&out, "Pulled: %s:%s\n", chart, ver)
```
benoittgt: Are we ok that it will differ a little bit as OCI classic output ?

zachburg: Could this print the URL that was pulled, to better match the OCI path?

Repos can specify multiple URLs (and sometimes these are `oci://` URLs): https://helm.sh/docs/topics/chart_repository/#the-index-file

benoittgt: Yes completly. I added https://github.com/helm/helm/pull/31206/commits/bf27e96d0eb7ba8edf4da37003323cba98ca7d11

```sh
❯ bin/helm pull ingress-nginx --repo=https://kubernetes.github.io/ingress-nginx --version 4.10.1
Pulled: https://github.com/kubernetes/ingress-nginx/releases/download/helm-chart-4.10.1/ingress-nginx-4.10.1.tgz
Digest: sha256:b3e4b14509f56080e09b022cde05dd02822fb054616ea212d93981b461732913

❯ bin/helm pull redis --repo=https://charts.bitnami.com/bitnami --version='22.0.5'
Pulled: oci://registry-1.docker.io/bitnamicharts/redis:22.0.5
Digest: sha256:59768807b09772b4b7d2b237a0053701811cda0ea2e58e1e12dbfdbda42d2580

File: pkg/action/pull.go [hunk 3]
```diff
@@ -138,6 +140,34 @@ func (p *Pull) Run(chartRef string) (string, error) {
...
+	// For direct OCI pulls, the registry client already prints "Pulled:" and
+	// "Digest:" (manifest digest), so we do not print here to avoid duplicates.
+	if p.RepoURL != "" || !registry.IsOCI(downloadSourceRef) {
+		base := trimAnySuffix(downloadSourceRef, ".tar.gz", ".tgz")
+		chart, ver := splitChartNameVersion(base)
+
+		tag := chart
+		if ver != "" {
+			tag += ":" + ver
+		} else if p.Version != "" {
```
benoittgt, line 161: I’m not sure we want this. It means we can select one version, but another might actually be downloaded during resolution… That feels even less safe than extracting the version from the URL.  

I’m fine with removing it or moving the logic elsewhere, but if we really want to read the exact version, we’ll need to rely on `chart/loader` and I don't think we want this.

pkg/action/pull.go
```diff
@@ -173,3 +203,33 @@ func (p *Pull) Run(chartRef string) (string, error) {
...
+		return nil, err
+	}
+	defer f.Close()
+
+	h := sha256.New()
+	if _, err := io.Copy(h, f); err != nil {
+		return nil, err
+	}
+	return h.Sum(nil), nil
+}
```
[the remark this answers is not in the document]
benoittgt: This is not true. When looking at [copyBuffer](https://cs.opensource.google/go/go/+/refs/tags/go1.25.0:src/io/io.go;l=407)'s code, we have defined 32kb and we reuse it.
```go
func Copy(dst Writer, src Reader) (written int64, err error) {
	return copyBuffer(dst, src, nil)
}
```

pkg/action/pull.go
```diff
@@ -173,3 +203,33 @@ func (p *Pull) Run(chartRef string) (string, error) {
 	}
 	return out.String(), nil
 }
+
```
TerryHowe, line 208: Since trimAnySuffix is only used for tgz type suffixes, maybe just make it trim those

pkg/action/pull.go
```diff
@@ -138,6 +140,34 @@ func (p *Pull) Run(chartRef string) (string, error) {
...
+		if ver != "" {
+			tag += ":" + ver
+		} else if p.Version != "" {
+			tag += ":" + p.Version
+		}
+		fmt.Fprintf(&out, "Pulled: %s\n", tag)
+
+		if sum, err := sha256File(saved); err == nil {
+			fmt.Fprintf(&out, "Digest: sha256:%x\n", sum)
+		}
```
TerryHowe, line 170: Maybe an else with a warning about the failure?

TerryHowe (commented):
I like this idea, but I generally stick to the oci side of things

pkg/action/pull.go
benoittgt: https://github.com/helm/helm/pull/31206/commits/5b0dc8f021859ed9bf32837e9f1db94c10f827e1

pkg/action/pull.go
benoittgt: https://github.com/helm/helm/pull/31206/commits/df391c0324861b59317a5c610ab9687354efae9a

TerryHowe (dismissed):
/lgtm

pkg/action/pull.go
sabre1041: @benoittgt Certainly interesting ideas here. however, before investigating the comments/source code, I questioned how the digest was being calculated. 

What about for http based pulls, we swap the output to say `Checksum` (which aligns to the actual value) instead of `Digest`?

benoittgt: Thanks @sabre1041 you are right. It is misleading. Because we use `sha256.New()` it is a [checksum](https://pkg.go.dev/crypto/sha256#New). 

Changed in https://github.com/helm/helm/pull/31206/commits/83789b112b43828e17bfbf6a9a1c964278a54f77

sabre1041 (approved):
LGTM

benoittgt:
@TerryHowe Can I have a "has one approval" ? Thanks

pkg/action/pull.go
```diff
@@ -138,6 +140,36 @@ func (p *Pull) Run(chartRef string) (string, error) {
...
+		} else if p.Version != "" {
+			tag += ":" + p.Version
+		}
+		fmt.Fprintf(&out, "Pulled: %s\n", tag)
+
+		if sum, err := sha256File(saved); err == nil {
+			fmt.Fprintf(&out, "Checksum: sha256:%x\n", sum)
+		} else {
+			fmt.Fprintf(&out, "Checksum failed: %v\n", err)
+		}
```
mattfarina, line 170: Internally, `DownloadTo` will sometimes have the hash of a file. The digest/hash is one of the things on a Repo index. I wonder if that should be exposed and checked against the calculated hash of the file. Then a message displayed if they are different.

Thoughts?

benoittgt: ~~That's a good idea but I am wondering if this could be in another PR and as an additionnal check that live into `DownloadTo` function ?~~

Ok so always print the checksum but add a warning before if there is a mismatch. I will try to implement this.

benoittgt: Ok I tried quickly with something like:

```diff
--- a/pkg/action/pull.go
+++ b/pkg/action/pull.go
@@ -164,7 +170,16 @@ func (p *Pull) Run(chartRef string) (string, error) {
                fmt.Fprintf(&out, "Pulled: %s\n", tag)
 
                if sum, err := sha256File(saved); err == nil {
-                       fmt.Fprintf(&out, "Checksum: sha256:%x\n", sum)
+                       calculatedDigest := fmt.Sprintf("%x", sum)
+                       fmt.Fprintf(&out, "Checksum: sha256:%s\n", calculatedDigest)
+                       if c.ExpectedDigest != "" {
+                               if calculatedDigest != c.ExpectedDigest {
+                                       fmt.Fprintf(&out, "WARNING: Checksum mismatch! Expected: %s, Got: %s\n", c.ExpectedDigest, calculatedDigest)
+                               }
+                       }
                } else {
                        fmt.Fprintf(&out, "Checksum failed: %v\n", err)
                }

--- a/pkg/downloader/chart_downloader.go
+++ b/pkg/downloader/chart_downloader.go
@@ -85,6 +85,12 @@ type ChartDownloader struct {
 
        // Cache specifies the cache implementation to use.
        Cache Cache
+
+       ExpectedDigest string
 }
 
 // DownloadTo retrieves a chart. Depending on the settings, it may also download a provenance file.
@@ -111,6 +117,9 @@ func (c *ChartDownloader) DownloadTo(ref, version, dest string) (string, *proven
                return "", nil, err
        }
 
+       // Store for later verification.
+       c.ExpectedDigest = hash
+
        g, err := c.Getters.ByScheme(u.Scheme)
        if err != nil {
                return "", nil, err
```

But of course it breaks some tests and I need to look at it.

benoittgt: Ok.. I will see how I can add a test for that but here is the output with a manually corrupted file.

```sh
$ HELM_TEST_CORRUPT=1  bin/helm pull stable/mysql

Pulled: stable/mysql
Checksum: sha256:3791aa11aed2d587c575be44fdd56895d0ecf62a84fe3a326e51ef0fc03208be
WARNING: Checksum mismatch!
  Expected: 35f232f0b4df50e85c6d65cbe7e5291b8ac8fab2c5a6afa8a0b816c7ce594390
  Got:      3791aa11aed2d587c575be44fdd56895d0ecf62a84fe3a326e51ef0fc03208be
```

benoittgt:
I will look into it

PR #31224, opened 2025-09-01
fix: Adjust PostRenderer plugin output to value

gjenkins8:

Plugin output type convention is to pass the message by value. Adjust post-renderer plugin code to be consistent with other places.

**If applicable**:


scottrigby (approved):
fair. lgtm

PR #31225, opened 2025-09-01
Move lint pkg to be part of each chart version

mattfarina:
Linting is specific to the chart versions. A v2 and v3 chart will
lint differently.

To accomplish this, packages like engine need to be able to handle different chart versions. This was accomplished by some changes:

1. The introduction of a Charter interface for charts
2. The ChartAccessor which is able to accept a chart and then provide access to its data via an interface. There is an interface, factory, and implementation for each version of chart.
3. Common packages were moved to a common and util packages. Due to some package loops, there are 2 packages which may get some consolidation in the future.

The new interfaces provide the foundation to move the actions and cmd packages to be able to handle multiple apiVersions of charts.

**What this PR does / why we need it**: It moves _lint_ to be specific to the chart versions.

1. The _actions_ have not been updated to deal with multiple versions. This would need to happen in a follow-up change.
2. For package names the linting has been removed for common names. It was triggered for existing _util_ packages that had minor changes.
3. The foundation setup with the Charter and Accessor can be expanded to other things to enable multiple API versions to be supported.

**If applicable**:


robertsirc:
@mattfarina we have some branch conflicts can we address and then we can get the checks to run?

scottrigby (approved):
lgtm. nice work

robertsirc (approved):
LGTM

PR #31227, opened 2025-09-03
Use filepath.Path when handling directory names

evankanderson:

On Windows, attempting to use `helm template --include-crds` on `oci://docker.io/envoyproxy/gateway-helm:v1.5.0` fails with an error:

```
$ helm template envoy-gateway-system . -n envoy-gateway-system -f .\values.yaml --include-crds --output-dir ..\..\argo\

Error: open ..\..\argo\\envoy-gateway-system\charts\gateway-helm\crds\gatewayapi-crds.yaml: The system cannot find the path specified.
```

This seems to be because the `ensureDirectoryForFile` function uses `path.Join` rather than `filepath.Join`, which produces the wrong type of slashes on Windows.

**If applicable**:

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -26,7 +26,6 @@ import (
 	"log/slog"
 	"net/url"
 	"os"
-	"path"
 	"path/filepath"
 	"strings"
 	"sync"
@@ -677,7 +676,7 @@ func createOrOpenFile(filename string, appendData bool) (*os.File, error) {
 
 // check if the directory exists to create file. creates if doesn't exist
 func ensureDirectoryForFile(file string) error {
-	baseDir := path.Dir(file)
+	baseDir := filepath.Dir(file)
 	_, err := os.Stat(baseDir)
 	if err != nil && !errors.Is(err, fs.ErrNotExist) {
 		return err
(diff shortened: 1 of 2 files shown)


evankanderson:
@TerryHowe -- this ended up blocking some of my other work; I'd love to get it into a Helm v3 release if possible (it looks like you're currently triage maintainer, apologies if I got that wrong).

If I understand the contribution process correctly, this goes into v4 first, then needs a backport?  Should I do the backport once it is merged, or is there automation for that?

TerryHowe:
Yes and the backport process is manual.

TerryHowe (approved):
/lgtm

pkg/action/install.go
```diff
@@ -677,7 +676,7 @@ func createOrOpenFile(filename string, appendData bool) (*os.File, error) {
 
 // check if the directory exists to create file. creates if doesn't exist
 func ensureDirectoryForFile(file string) error {
-	baseDir := path.Dir(file)
+	baseDir := filepath.Dir(file)
```
mattfarina, line 679: This same issue is in the `ensureDirectoryForFile` function in the _pkg/cmd_ directory for templates. Can you please fix it both places.

evankanderson: Done.  I hadn't encountered this as part of my error debugging, so I missed updating this function.

TerryHowe (approved):
/lgtm

evankanderson:
(I think this needs workflows approved again)

gjenkins8 (approved):

PR #31238, opened 2025-09-03
Use consistent slashes whether expanded on Windows or POSIX

evankanderson:
(This builds on #31227, but is separate because it actually changes output on Windows)

Currently, when expanding helm charts via `helm template` on Windows, Windows-style path separators are used.  This can generate unnecessary diffs like the following when checking in `helm template` results across different platforms (e.g. in the [rendered manifests pattern](https://akuity.io/blog/the-rendered-manifests-pattern)).

Example:

```
diff --git a/prod/argo/envoy-gateway-system/charts/gateway-helm/crds/generated/gateway.envoyproxy.io_securitypolicies.yaml b/prod/argo/envoy-gateway-system/charts/gateway-helm/crds/generated/gateway.envoyproxy.io_securitypolicies.yaml
index d3c1ba2..f831846 100644
--- a/prod/argo/envoy-gateway-system/charts/gateway-helm/crds/generated/gateway.envoyproxy.io_securitypolicies.yaml
+++ b/prod/argo/envoy-gateway-system/charts/gateway-helm/crds/generated/gateway.envoyproxy.io_securitypolicies.yaml
@@ -1,5 +1,5 @@
 ---
-# Source: envoy-gateway-system/charts/gateway-helm/crds/generated/gateway.envoyproxy.io_securitypolicies.yaml
+# Source: envoy-gateway-system\charts\gateway-helm\crds\generated\gateway.envoyproxy.io_securitypolicies.yaml
```

This could technically be breaking if people are relying on these elsewhere, but I'm not sure how to assess that.

If there's a standard way to add Windows-specific tests, I'd be happy to add some for this.  On the other hand, `go test ./...` has a lot of failures on Windows, so there may be more to clean up.

**If applicable**:

### DIFF
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -281,7 +281,7 @@ func (cfg *Configuration) renderResources(ch *chart.Chart, values common.Values,
 			if strings.TrimSpace(content) == "" {
 				continue
 			}
-			fmt.Fprintf(b, "---\n# Source: %s\n%s\n", name, content)
+			fmt.Fprintf(b, "---\n# Source: %s\n%s\n", filepath.ToSlash(name), content)
 		}
 		return hs, b, "", err
 	}
@@ -292,7 +292,7 @@ func (cfg *Configuration) renderResources(ch *chart.Chart, values common.Values,
 	if includeCrds {
 		for _, crd := range ch.CRDObjects() {
 			if outputDir == "" {
-				fmt.Fprintf(b, "---\n# Source: %s\n%s\n", crd.Filename, string(crd.File.Data[:]))
+				fmt.Fprintf(b, "---\n# Source: %s\n%s\n", filepath.ToSlash(crd.Filename), string(crd.File.Data[:]))
 			} else {
 				err = writeToFile(outputDir, crd.Filename, string(crd.File.Data[:]), fileWritten[crd.Filename])
 				if err != nil {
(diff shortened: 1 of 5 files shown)


TerryHowe (approved):
/lgtm

evankanderson:
(I think this needs the workflows approved)

godhanipayal (approved):
Fairly straightforward. /lgtm

evankanderson:
Following the [feedback on the other PR](https://github.com/helm/helm/pull/31227#discussion_r2337640253), I applied path fixes to all instances of `fmt\..?[Pp]rintf([^)]+# Source:`  (I actually scanned a bit wider than that, but that regex covers what I found.)

TerryHowe (dismissed):
/lgtm

robertsirc (dismissed):
LGTM

gjenkins8 (dismissed):

evankanderson:
Sorry for the late reply (and I'm not sure what happened with the merge base).

> While the "rendered manifest" pattern is valuable, this change also worsens the experience for another set of users.

What's the set of users where the experience is worse?  I think that set of users is a set of users using `helm` _only_ on Windows platforms with tools which don't know how to convert paths from POSIX-style to UNC-style.

> Furthermore, the problem of the rendered manifest difference across platforms could be solved with a simple helm template | sed -i 's/^# Source: .*$//' (or whatever the equivalent is on windows).

So -- this solution introduces the following compromises:

1. You lose the pointers back to the original files, which seems unfortunate.  A better pattern might be to post-replace all the `\` with `/` in the lines that start with `^# Source:`, but that's more-complex with `sed`, and I'm not sure what the equivalent is on Windows, either.
1. You need different tooling on Windows and Mac/Linux, unless you're requiring the client to do a lot of set-up.
1. I'm particularly using this with `helm template --output-dir`, so pipes don't work, and instead you need to crawl all the output files.

If this were easy to "fix in post", I probably would have, since I'm already wrapping the `helm` invocation in a script.

TerryHowe:
Conflicts [maybe with your other PR](https://github.com/helm/helm/pull/31227) need to be resolved

Issue #31170, opened 2025-08-20
Regression: Helm 3.18.6 fails during schema validation with `no URLLoader registered`


felipecrs:
### What happened?

```console
$ helm template ./mychart
02:33:45  Error: values don't meet the specifications of the schema(s) in the following chart(s):
02:33:45  dependency-a:
02:33:45  failing loading "urn:example:helm:schemas:v1:helm-schema-validation-conditions:v1/helmSchemaValidation-true": no URLLoader registered for "urn:example:helm:schemas:v1:helm-schema-validation-conditions:v1/helmSchemaValidation-true"
```

### What did you expect to happen?

To work as it did in 3.18.4. 3.18.5 did not work because of [this](https://github.com/helm/helm/issues/31136).

### How can we reproduce it (as minimally and precisely as possible)?

I haven't built a minimal reproduction yet, and the chart is private.

### Helm version

```console
$ helm version
v3.18.6+gb76a950
```

### Kubernetes version

N/A

egerkke:
Maybe caused by this change?
https://github.com/helm/helm/commit/b9180e674fccb57e6ea6934ed7deb4448a3c9ddb

safderali5:
Same error for me, seen in helm 3.18.6 and 3.18.5.

While the 3.18.4 is working fine.

benoittgt:
I dig into this subject and it was not easy to figure this out.

There is a [json library migration in 3.18.5](https://github.com/helm/helm/compare/v3.18.4...v3.18.5#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R32) that means that json schema is stricker now. What I understand is 3.18.4 didn’t check external URNs unless validation walked into them (and/or they were preloaded).
3.18.5’s loader wiring made external $refs compile-time strict, exposing the missing urn handler.

It seems that URN handling for external reference is not easily possible. We have to use [`addResource`](https://pkg.go.dev/github.com/santhosh-tekuri/jsonschema/v6#Compiler.AddResource) in Go code to dynamicaly loading other related json files. Same-document URNs (schema has $id: urn:… and refs use urn:…#anchor) work without a loader; only external URNs need AddResource/loader.
I played with the cli tool of the new jsonschema lib and I was not able to trick the schema to use an external json expect using `--map` (see: https://github.com/santhosh-tekuri/jsonschema/commit/ed65924244c7fad4784aebfafcad7aa1ece6d2c2)

[...]

PR #31240, opened 2025-09-04
jsonschema: warn and ignore unresolved URN $ref to match v3.18.4

benoittgt:
**What this PR does / why we need it**:

Fixes: https://github.com/helm/helm/issues/31170

- v3.18.5 switched jsonschema and began resolving external $ref at compile-time, exposing missing urn handling (“no URLLoader registered for urn:…”).
- Sor I added in this PR a urn scheme loader and pluggable URNResolver. If unresolved, log a warning and return a permissive true schema (back-compatibilty like in 3.18.4).

To properly deal with another json file, we could in another PR add the ability to import other schema. Similar function exist jsonschema cli as with `--map` (https://github.com/santhosh-tekuri/jsonschema/commit/ed65924244c7fad4784aebfafcad7aa1ece6d2c2).

Tested using a small chart [issue-31170-0.1.0.tgz](https://github.com/user-attachments/files/22139344/issue-31170-0.1.0.tgz)
With this basic `value.schema.json`

```js
{
  "$schema": "https://json-schema.org/draft-07/schema#",
  "$ref": "urn:example:helm:schemas:v1:helm-schema-validation-conditions:v1/helmSchemaValidation-true"
}
```

Before:
```sh
$ ./bin/helm lint ./testdata/charts/issue-31170
==> Linting ./testdata/charts/issue-31170
[INFO] Chart.yaml: icon is recommended
[ERROR] values.yaml: failing loading "urn:example:helm:schemas:v1:helm-schema-validation-conditions:v1/helmSchemaValidation-true": no URLLoader registered for "urn:example:helm:schemas:v1:helm-schema-validation-conditions:v1/helmSchemaValidation-true"
[ERROR] templates/: values don't meet the specifications of the schema(s) in the following chart(s):
issue-31170:
failing loading "urn:example:helm:schemas:v1:helm-schema-validation-conditions:v1/helmSchemaValidation-true": no URLLoader registered for "urn:example:helm:schemas:v1:helm-schema-validation-conditions:v1/helmSchemaValidation-true"

Error: 1 chart(s) linted, 1 chart(s) failed
```

After:
```
$ ./bin/helm lint ./testdata/charts/issue-31170                  
level=WARN msg="unresolved URN reference ignored; using permissive schema" urn=urn:example:helm:schemas:v1:helm-schema-validation-conditions:v1/helmSchemaValidation-true
[...]

PR #31247, opened 2025-09-05
[backport] jsonschema: warn and ignore unresolved URN $ref to match v3.18.4

benoittgt:

- v3.18.5 switched jsonschema and began resolving external $ref at compile-time, exposing missing urn handling (“no URLLoader registered for urn:…”).
- Add urn scheme loader and pluggable URNResolver. If unresolved, log a warning and return a permissive true schema (back-compat).
- Avoid having a duplicated warn logs when we load the schema twice on linting for example.

Close: #31170
Related: https://github.com/helm/helm/pull/31240

🤗

**If applicable**:

### DIFF
--- a/pkg/chartutil/jsonschema.go
+++ b/pkg/chartutil/jsonschema.go
@@ -82,7 +82,20 @@ func ValidateAgainstSchema(chrt *chart.Chart, values map[string]interface{}) err
 	}
 
 	for _, subchart := range chrt.Dependencies() {
-		subchartValues := values[subchart.Name()].(map[string]interface{})
+		raw, exists := values[subchart.Name()]
+		if !exists || raw == nil {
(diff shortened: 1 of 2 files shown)


pkg/chartutil/jsonschema.go
```diff
@@ -21,7 +21,9 @@ import (
 	"crypto/tls"
 	"errors"
 	"fmt"
+	"log"
```
benoittgt: Used log and not slog because this is what we use here. https://github.com/benoittgt/helm/blob/dev-v3/pkg/chartutil/dependencies.go#L67

benoittgt:
@scottrigby @mattfarina @gjenkins8 Could you review this? It affects Flux and other users. This approach was validated during the Helm dev meeting a couple of weeks ago.

PR #31250, opened 2025-09-05
fix: Ignore irregular files during copy

### LINKED ISSUE #12926: Unable to add plugins, helm uses fsmonitor regardless of git config
I am running into https://github.com/helm/helm/issues/12125, but I cannot seem to fix this issue even by entirely removing my .gitconfig

Running macOS 14.4.1

Trying to install helm-diff:

```
helm plugin install https://github.com/databus23/helm-diff
Error: copying directory failed: copying file failed: open /Users/j/Library/Caches/helm/plugins/https-github.com-databus23-helm-diff/.git/fsmonitor--daemon.ipc: operation not supported on socket
```

Git claims not to be running the fsmonitor--daemon:

```
git fsmonitor--daemon status
fsmonitor-daemon is not watching '/Users/j/Library/Caches/helm/plugins/https-github.com-databus23-helm-diff'
```

When cloning the repo manually, the `.git/fsmonitor--daemon` and `.git/fsmonitor--daemon.ipc` files do not exist, when trying to install via helm, they are created for some reason. I've rebooted after removing my entire .gitconfig and the issue persists.

Output of `git config -l`:
credential.helper=osxkeychain
init.defaultbranch=main

Output of `helm version`:
version.BuildInfo{Version:"v3.14.3", GitCommit:"f03cc04caaa8f6d7c3e67cf918929150cf6f3f12", GitTreeState:"clean", GoVersion:"go1.22.1"}

banjoh:

Ignore copying `.git` directory when installing plugins except when the source is from version control.

Fixes: https://github.com/helm/helm/issues/31139
Fixes: https://github.com/helm/helm/issues/12926

**Manual test**
```sh
# Enable Git FSMonitor. A domain socket will automatically be created when cloning helm-unittest.git repo
$ git config --global core.fsmonitor true

$ bin/helm plugin install https://github.com/helm-unittest/helm-unittest.git --debug --verify=false
WARNING: Skipping plugin signature verification
level=DEBUG msg=updating source=https://github.com/helm-unittest/helm-unittest.git destination=/Users/evans/Library/Caches/helm/plugins/https-github.com-helm-unittest-helm-unittest.git
level=DEBUG msg="copying files" source=/Users/evans/Library/Caches/helm/plugins/https-github.com-helm-unittest-helm-unittest.git destination=/Users/evans/Library/helm/plugins/helm-unittest.git
level=DEBUG msg="skip copying non-regular file" src=/Users/evans/Library/Caches/helm/plugins/https-github.com-helm-unittest-helm-unittest.git/.git/fsmonitor--daemon.ipc dst=/Users/evans/Library/helm/plugins/helm-unittest.git/.git/fsmonitor--daemon.ipc type=socket mode=Srwxr-xr-x
level=DEBUG msg="loading plugin" path=/Users/evans/Library/helm/plugins/helm-unittest.git
```

Note `"skip copying non-regular file"` log line

**If applicable**:

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -22,6 +22,6 @@ jobs:
           go-version: '${{ env.GOLANG_VERSION }}'
           check-latest: true
       - name: golangci-lint
-        uses: golangci/golangci-lint-action@4afd733a84b1f43292c63897423277bb7f4313a9 #pin@8.0.0
+        uses: golangci/golangci-lint-action@0a35821d5c230e903fcfe077583637dea1b27b47 #pin@9.0.0
         with:
           version: ${{ env.GOLANGCI_LINT_VERSION }}


scottrigby:
We should also add the git config fix you noted here https://github.com/helm/helm/issues/31139#issuecomment-3257699078 to the docs, for Helm versions prior to releasing this fix. Added `docs needed` label to remind us.

banjoh:
> We should also add the git config fix you noted here [#31139 (comment)](https://github.com/helm/helm/issues/31139#issuecomment-3257699078) to the docs, for Helm versions prior to releasing this fix. Added `docs needed` label to remind us.

@scottrigby https://github.com/helm/helm-www/pull/1763

banjoh:
> Would a more "sensible" alternative to be to not copy the `.git` dir when installing a plugin?
> 
> (IMHO "install from source(aka. VCS)" is wrong -- for reasons like this. Distribution should happen "built" artifacts only)

I agree, not copying `.git` dir would certainly address this particular problem. It does not address copying any problematic files not in `.git` dir if such were to be present. I also prefer narrowing down the solution to ignoring the specific files we want to ignore, rather than a whole directory containing other files that are not in scope

banjoh:
> > > Would a more "sensible" alternative to be to not copy the `.git` dir when installing a plugin?
> > > (IMHO "install from source(aka. VCS)" is wrong -- for reasons like this. Distribution should happen "built" artifacts only)
> > 
> > 
> > I agree, not copying `.git` dir would certainly address this particular problem. It does not address copying any problematic files not in `.git` dir if such were to be present. I also prefer narrowing down the solution to ignoring the specific files we want to ignore, rather than a whole directory containing other files that are not in scope
> 
> The thought, is that if we ignore any/all irregular files, we have to do a deep dive and/or risk breaking some valid use case. Whatever that might be ie. Setuid/Setuid files, or Sticky directories I think are counted as non-regular. And with this logic, Helm will ignore locked files, but I think this should be an error.
> 
> Whereas, we directly know (I think) that the `.git` dir is special, and uneeded by Helm.
> 
> Note: This isn't a super strongly held opinion, but more of a concern that there might be a bunch of gotchas in the land of POSIX non-regular files.

I see where you're coming from. I'll adjust to focus on addressing this specific known issue

banjoh:
@gjenkins8 I changed the implementation to ignore `.git` when installing plugins. This addresses the bug instead of attempting to implement a generic solution

banjoh:
@gjenkins8 I reopened the PR https://github.com/helm/helm/pull/31504

PR #31239, opened 2025-09-04
Add tests to action package to improve coverage

StephanieHhnbrg:
**What this PR does / why we need it**:
This PR is adding tests to the action package, improving the coverage ot the package from 61.1% to 70,7%

**Special notes for your reviewer**:
I added a new function to the Makefile and adapted coverage.sh to accept a directory as argument.
So that I could run the test coverage for a single package locally, rather than for the whole repository.
 
`make coverage-dir DIR=./pkg/action`

When no argument is given to `coverage.sh`, it is defaulting to test the whole repository.

closes #30814

**If applicable**:
- [ ] this PR contains user facing changes (the `docs needed` label should be applied if so)
- [x] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

PR #31259, opened 2025-09-07
Adapt test-coverage command to be able to run for a certain package

StephanieHhnbrg:
This PR adapts the test-coverage function in the Makefile and adapt coverage.sh to accept a directory as argument.
So that the test coverage can be run for a single package locally, rather than for the whole repository.

When no argument is given to coverage.sh, it is defaulting to test the whole repository.

Test locally: `make tet-coverage PKG=./pkg/action`

Originally part of the PR  #31239 

**If applicable**:


scottrigby (approved):
good speed up, fully backwards-compatible 💯

PR #31270, opened 2025-09-08
chore: registry utils clean up

TerryHowe:

Minor clean up on pkg/registry. Any file named utils is most likely a lazy place to throw junk. I divided it out to chart, tag and reference. Also `NewRegistryClientWithTLS` was not being used, so v4 is good time to remove junk. Minor test coverage improvement from 67.5% to 71.5%.

**If applicable**:


pkg/registry/reference.go
```diff
@@ -76,3 +77,8 @@ func (r *reference) String() string {
 	}
 	return r.orasReference.String()
 }
+
+// IsOCI determines whether a URL is to be treated as an OCI URL
+func IsOCI(url string) bool {
```
benoittgt, line 82: I'm curious why you put this here ?

benoittgt (approved):
Much better now.

pkg/registry/reference.go
TerryHowe: This file contains `newReference` which trims off `oci://` from references and parses out registry repository and tag.

Maybe you meant why did I put it at the bottom instead of the top of the file though.

benoittgt: Sorry I meant in `reference.go`. :) But no problem.

gjenkins8 (commented):
Changes look good (agree "utils" basically means "code author didn't think about how to structure" :) 

Can we fix conflicts and I can (re)approve pls

TerryHowe:
Rebased, not sure why the conflict was showing.

gjenkins8 (approved):

scottrigby (approved):

PR #31271, opened 2025-09-08
Add Detailed Logging and Error Handling for Chart Download

godhanipayal:
Add Detailed Logging and Error Handling for Chart Download

Signed-off-by: Payal Godhani <godhanipayal@gmail.com>

This PR adds detailed logging and error messages for chart download operations

- Errors are logged with context for easier troubleshooting.
- Defensive check for empty data.
- Early returns reduce nesting.
- Logging now uses slog.Error for actual errors, slog.Debug for process information

Enhances maintainability and debuggability of the chart download logic by providing clearer logs and handling edge cases gracefully.

**If applicable**:

### DIFF
--- a/pkg/downloader/chart_downloader.go
+++ b/pkg/downloader/chart_downloader.go
@@ -246,14 +246,14 @@ func (c *ChartDownloader) DownloadToCache(ref, version string) (string, *provena
 	if len(digest) == 0 || err != nil {
 		slog.Debug("attempting to download chart", "ref", ref, "version", version)
 		if err != nil && !os.IsNotExist(err) {
-			slog.Error("failed to access chart cache", "error", err, "ref", ref, "version", version)
+			slog.Debug("failed to access chart cache", "error", err, "ref", ref, "version", version)
 			return "", nil, err
 		}
 
 		// Get file not in the cache
 		data, gerr := g.Get(u.String(), c.Options...)
 		if gerr != nil {
-			slog.Error("failed to download chart from remote", "error", gerr, "url", u.String(), "ref", ref, "version", version)
+			slog.Debug("failed to download chart from remote", "error", gerr, "url", u.String(), "ref", ref, "version", version)
 			return "", nil, gerr
 		}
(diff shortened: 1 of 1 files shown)


pkg/downloader/chart_downloader.go
```diff
@@ -246,22 +246,31 @@ func (c *ChartDownloader) DownloadToCache(ref, version string) (string, *provena
...
+			slog.Error("failed to download chart from remote", "error", gerr, "url", u.String(), "ref", ref, "version", version)
 			return "", nil, gerr
 		}
 
 		// Generate the digest
 		if len(digest) == 0 {
-			digest32 = sha256.Sum256(data.Bytes())
+			// Defensive: check data.Bytes() is not nil/empty
+			bytes := data.Bytes()
+			if len(bytes) == 0 {
```
TerryHowe, line 264: Does something else check for empty chart data or is it supposed to be valid?

The sh256 should work fine for ""

TerryHowe (commented):
I'd like someone else to weigh in here that knows this code better, but I would think we'd want to slog.Debug these instead and let the higher levels do logging if any.

The other alternative would be wrap the errors instead to provide additional information.

pkg/downloader/chart_downloader.go
godhanipayal: Thanks @TerryHowe  for reviewing.  I checked the codebase and dint find any explicit checks for empty chart data right after downloading. There are other validations for chart contents like metadata and such.

You have a good point about sha256.Sum256 and it does work with empty data. I added this check as in this context, empty chart download likely means somehting went wrong like network issue or bad remote response. This check will make it fail fast and prevents unnecessary processing

godhanipayal:
> I'd like someone else to weigh in here that knows this code better, but I would think we'd want to slog.Debug these instead and let the higher levels do logging if any.
> 
> The other alternative would be wrap the errors instead to provide additional information.

- Changed the log level to debug

PR #31272, opened 2025-09-09
fix: idea gitignore entry

TerryHowe:

So the JetBrains file doesn't show up.


mattfarina (approved):
lgtm

PR #31076, opened 2025-07-18
pkg/registry: Login option for passing TLS config in memory

matheuscscp:
**What this PR does / why we need it**: Closes #31075

**If applicable**:
- [x] this PR contains unit tests
- [x] this PR has been tested for backwards compatibility

PR #31284, opened 2025-09-12
pkg/registry: Login option for passing TLS config in memory

matheuscscp:
Backporting #31076 to v3 dev branch

cc @scottrigby @mattfarina

### DIFF
--- a/pkg/registry/client_tls_test.go
+++ b/pkg/registry/client_tls_test.go
@@ -17,6 +17,8 @@ limitations under the License.
 package registry
 
 import (
+	"crypto/tls"
+	"crypto/x509"
 	"os"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


pkg/registry/client_tls_test.go
```diff
@@ -52,6 +54,30 @@ func (suite *TLSRegistryClientTestSuite) Test_0_Login() {
...
 
+func (suite *TLSRegistryClientTestSuite) Test_1_Login() {
+	err := suite.RegistryClient.Login(suite.DockerRegistryHost,
+		LoginOptBasicAuth("badverybad", "ohsobad"),
+		LoginOptTLSClientConfigFromConfig(&tls.Config{}))
+	suite.NotNil(err, "error logging into registry with bad credentials")
+
+	// Create a *tls.Config from tlsCert, tlsKey, and tlsCA.
+	cert, err := tls.LoadX509KeyPair(tlsCert, tlsKey)
+	suite.Nil(err, "error loading x509 key pair")
```
miledxz, line 65: small suggestion here, 
what you think about, that instead of using, `suite.Nil(err, "error loading x509 key pair")`, which records the failure but continues execution of the test, we use `suite.Require().NoError(err)`, which would record the failure and then aborts the test immediately (like t.Fatal).

This way, if setup fails, the test exits immediately, because there’s maybe no point testing login when the TLS config can’t even be built

What do you think ?

TerryHowe: A good suggestion, but I'd sooner this stay the same as the version that merged to main.

TerryHowe (approved):
/lgtm

sabre1041 (approved):
LGTM

gjenkins8 (approved):

PR #31288, opened 2025-09-12
chore: add get-helm script for latest release

TerryHowe:

I'm not sure how everyone wants to handle this, but I just copied the scripts/get-helm-3 and changed v3 to v4. Release doesn't matter with this script, but it looks a bit odd to call `get-helm-3` for v4.

### DIFF
--- a/scripts/get-helm
+++ b/scripts/get-helm
@@ -1,0 +1,347 @@
+#!/usr/bin/env bash
+
+# Copyright The Helm Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


robertsirc (approved):
LGTM

TerryHowe:
I changed the title of this PR to make it more clear. This script gets the latest helm.  We should probably have this script and advertise it instead of advertising `get-helm-3` and then change `get-helm-3` to get the latest v3.

TerryHowe:
> I created #31318 to fix the impending Helm v4 release issue

sounds good to me

PR #31294, opened 2025-09-15
Remove implicit support for helm lint current directory

### LINKED ISSUE #8051: proposal: Require a path for 'helm lint'
For a major Helm release, it would be nice to make the following breaking change to `helm lint`:

Right now, `helm lint` is the one Helm command that supports an implicit path for ARG1. If `helm lint foo` is called, Helm will load the chart at `foo`. But if `helm lint` is called, it will implicitly call `helm lint` on the current working directory.

No other Helm commands behave this way, and it can lead to unexpected behaviors in cases like `helm lint $mychart` (where $mychart is not set).

We should align `helm lint` with install/upgrade/package/etc. and remove implicit path support.

See #7155

TerryHowe:
This change removes the implicit path support for the helm lint command to align it with other Helm commands like install, upgrade, and package.

Closes: https://github.com/helm/helm/issues/8051

### DIFF
--- a/pkg/cmd/lint.go
+++ b/pkg/cmd/lint.go
@@ -30,6 +30,7 @@ import (
 	"helm.sh/helm/v4/pkg/chart/common"
 	"helm.sh/helm/v4/pkg/chart/v2/lint/support"
 	"helm.sh/helm/v4/pkg/cli/values"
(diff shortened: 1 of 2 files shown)


mattfarina (approved):
lgtm

robertsirc (approved):
LGTM

PR #31300, opened 2025-09-16
adding in error handling for fmt.Fprintln

robertsirc:
I added error handling for `fmt.Fprintln`. This clears up a bit of warning in our builds. Some areas we ignored as error handling was further down the code, or when it was just writing a string out.

**If applicable**:


pkg/action/dependency.go
```diff
@@ -62,12 +62,18 @@ func (d *Dependency) List(chartpath string, out io.Writer) error {
 	}
 
 	if c.Metadata.Dependencies == nil {
-		fmt.Fprintf(out, "WARNING: no dependencies at %s\n", filepath.Join(chartpath, "charts"))
+		_, err := fmt.Fprintf(out, "WARNING: no dependencies at %s\n", filepath.Join(chartpath, "charts"))
+		if err != nil {
+			return err
+		}
 		return nil
```
TerryHowe: Could just return err here

pkg/cmd/get_manifest.go
```diff
@@ -54,7 +54,10 @@ func newGetManifestCmd(cfg *action.Configuration, out io.Writer) *cobra.Command
 			if err != nil {
 				return err
 			}
-			fmt.Fprintln(out, res.Manifest)
+			_, err = fmt.Fprintln(out, res.Manifest)
+			if err != nil {
+				return err
+			}
```
TerryHowe: return err here

TerryHowe (commented):
Sorry the copilot review is a bit more than I thought and some junk in there, but a couple good suggestions I think.

TerryHowe:
Thanks for doing this by the way. I don't want to turn off these warnings in my IDE and this just creates noise.

robertsirc:
@TerryHowe no problem on copilot i can make some of these changes you requested.

TerryHowe (approved):
/lgtm

mattfarina (commented):
Why are we making this change for the _fmt_ functions?

I ask because Go itself does not do this. It ignores the errors. The same goes for Kubernetes. I wouldn't say this change was idiomatic.

TerryHowe:
> Why are we making this change for the _fmt_ functions?

If command cannot write output, the user did not receive the result and that is a failure.

mattfarina (commented):
Some areas, like buffers, never return an error. The error handling around `bytes.Buffer`, for example, will never be exercised. See [here](https://github.com/golang/go/blob/3db5979e8cc4cc86c4fefb4cecc5a2041b32404d/src/bytes/buffer.go#L181-L188) for the underlying source.

If we're looking at `os.Stdout` or `os.Stderr` it's different. Errors can happen there (see [here](https://github.com/golang/go/blob/bffe7ad9f11b874429d6c2561b0383e931502164/src/os/file.go#L211-L230)). I was looking at other projects that capture these errors and a number of them panic when stderr or stdout aren't available to write to.

With the Helm CLI, if there is a problem writing to stderr/stdout the error is bubbled up to be displayed by stderr. How would this work? Instead of stderr/stdout, an SDK user could pass something else in. Would that be a different kind of situation? I suspect it would be logged.

I think if the underlying thing being printed to, like a `bytes.Buffer`, doesn't ever return an error it's not worth having error handling.

Thoughts?

robertsirc:
So what @mattfarina is saying since we primarily use `bytes.Buffer` that returns nil for err. Any other implementation could return an err but not all do. The issue we have is that linters will not know the underlying implementation so they warn that we are not handling the error. Either we implement handling the errors as (admittedly) over kill or mute  the warnings.

TerryHowe:
If everything is bytes.Buffer, a `_, _ =` would do I guess

pkg/cmd/template.go
```diff
@@ -108,15 +108,21 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 			// we always want to print the YAML, even if it is not valid. The error is still returned afterwards.
 			if rel != nil {
 				var manifests bytes.Buffer
-				fmt.Fprintln(&manifests, strings.TrimSpace(rel.Manifest))
+				_, err := fmt.Fprintln(&manifests, strings.TrimSpace(rel.Manifest))
+				if err != nil {
+					return err
+				}
```
mattfarina, line 114: `bytes.Buffer` never returns an error. The `if err != nil` will never be run because the underlying writer to the buffer doesn't produce an error. It always returns nil to follow the interface.

This was something I learned about Go when I went digging into the source.

Can we remove the error handling around `bytes.Buffer` cases since it will never execute?

pkg/cmd/show.go
```diff
@@ -113,7 +116,10 @@ func newShowCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 			if err != nil {
 				return err
 			}
-			fmt.Fprint(out, output)
+			_, err = fmt.Fprint(out, output)
+			if err != nil {
+				return err
+			}
```
mattfarina, line 122: `out` is the output (e.g. stdout). Today the error is ignored. This means that the Helm CLI can be run detached from a terminal. This change would cause an error if Helm is run detached from a terminal.

This may cause a problem. Do we want to exclude that use case? I've not used it before but I can imagine someone might.

In any case, this is a change in behavior. Credit to the gophers slack. Discussing this topic there brought up this angle.

mattfarina (commented):
This is an interesting topic. There is https://github.com/golang/go/issues/20803 that discussing making handling all returns mandatory and discusses an ignore syntax to make it explicit.

Yet, in the Go source the returns are implicitly ignored rather than explicitly ignored. `fmt` seems to be an exception to the general rule on this topic. I looked through numerous popular Go projects including Go itself, k8s, prometheus, terraform, and vault. None of them do explicit ignored returns on `fmt` functions.

As I thought about it, if you can't write to sdterr it would produce an error (if we return an error for that) that bubbles up to be output to stderr (which still isn't likely to work).

I think this should be discussed at a Helm maintainers call.

Note, I did find a couple issues in the handling that are either always uncalled code or a behavior change in some edge cases.

mattfarina:
Note, I think this is a style change that goes against what's common in Go codebases (including Go itself) but follows the general guidance for Go (as in don't ignore any errors). It's a weird situation. I would like to see this discussed in the Helm maintainers call before moving forward on it so the discussion can be wider than a PR.

mattfarina:
Note, errcheck (which is what people use to check this with golangci-lint) skips these fmt packages by default. See https://github.com/kisielk/errcheck/blob/dacab891ef4a1c38ecf6c4d94fd66746bb1247d5/errcheck/excludes.go#L13-L58

robertsirc:
I think this is a bigger topic for this PR. I think we can table this for the time being.

PR #31302, opened 2025-09-17
fix: helm verify Run signature

TerryHowe:

I'm not vey familiar with the provenance and so I haven't been able to properly functionally test this, but seems simple enough and tests pass.

**If applicable**:


godhanipayal (approved):

gjenkins8 (approved):
It would be (much) better if the output was structured. But this is an improvement regardless. Thanks!

scottrigby (approved):
Nice :+1:

PR #31303, opened 2025-09-17
Update the action interfaces for chart apiversions

mattfarina:
This change is about handling the interfaces to public functions for different chart apiVersions. The internals are still focused on v2. This enables v3 to be layered in layer.

**Special notes for your reviewer**: This adds a loader that can handle v2 and v3 charts. Work still needs to be done to make v3 work and have it be an experiment. This is the next step by enabling the arguments to functions to support multiple versions.

**If applicable**:


pkg/action/install.go
```diff
@@ -252,7 +253,17 @@ func (i *Install) Run(chrt *chart.Chart, vals map[string]interface{}) (*release.
...
 // When the task is cancelled through ctx, the function returns and the install
 // proceeds in the background.
-func (i *Install) RunWithContext(ctx context.Context, chrt *chart.Chart, vals map[string]interface{}) (*release.Release, error) {
+func (i *Install) RunWithContext(ctx context.Context, ch ci.Charter, vals map[string]interface{}) (*release.Release, error) {
+	var chrt *chart.Chart
+	switch c := ch.(type) {
+	case *chart.Chart:
+		chrt = c
+	case chart.Chart:
+		chrt = &c
```
scottrigby, line 262: my main curiosity is what problem i solved by switching on both pointer and object. taking a guess at the problem, would it be better solved by choosing one or the other to pass from the caller instead of this func accepting both?

same question for the other switch statements that account for pointer and non-pointer.

scottrigby (approved):
since this is experimental to solve the charts v3 problem, rubber stamping for now. Just one question, but won't block.

pkg/action/install.go
mattfarina: The caller to these function can pass in a chart directly or by reference through the interface. Type switching will know the difference. To only do one or the other assumes the calling function of this one will know or do one way or the other. Since it's an interface, the tooling can't give a hint. Handling both ways makes the code more fault tolerant.

PR #31306, opened 2025-09-18
feat(lint): allow files starting with _ to be skipped

### LINKED ISSUE #31305: The lint rule for checking template file extensions should exclude files that start with an underscore.
While template files can only generate a limited set of types, this shouldn't be true for "underscore files", which don't generate anything directly, but only define named templates. I often use these to generate values for use in configmaps or envoy configuration, which may be ini files or json files or lua code. being able to use the appropriate file extension would enable editors to apply syntax highlighting.

Relevant code: https://github.com/helm/helm/blob/90a91e097ff7fb4845b8e754502d3aa1b1b983c2/internal/chart/v3/lint/rules/template.go#L222

Qu-Ack:
Currently, `helm lint` fails for files in the templates directory that have extensions other than `.yaml`, `.yml`, `.tpl`, or `.txt`.  

This PR adds support for files starting with `_`, allowing them to be skipped during linting.  
This makes it possible to include additional file types (e.g., `.json`, `.lua`) in the templates directory while still passing lint checks.  
Using `_` also makes it clear that these files are intended to be ignored by Helm.

**Which issue this PR fixes**:
closes #31305

N/A

**If applicable**:
- [ ] this PR contains user facing changes (apply the `docs needed` label if so)

### DIFF
--- a/internal/chart/v3/lint/rules/template.go
+++ b/internal/chart/v3/lint/rules/template.go
@@ -220,6 +220,13 @@ func validateTemplatesDir(templatesPath string) error {
 }
 
 func validateAllowedExtension(fileName string) error {
+	baseName := filepath.Base(fileName)
+
+	if strings.HasPrefix(baseName, "_") {
+		fmt.Println("skipping _ files")
+		return nil
+	}
+
 	ext := filepath.Ext(fileName)
 	validExtensions := []string{".yaml", ".yml", ".tpl", ".txt"}
 
--- a/internal/chart/v3/lint/rules/template.go
+++ b/internal/chart/v3/lint/rules/template.go
@@ -123,6 +123,11 @@ func TemplatesWithSkipSchemaValidation(linter *support.Linter, values map[string
 		fileName := template.Name
 		fpath = fileName
 
+		// skip the linting rules if the file starts with _ and is not a yml/yaml file
+		if isHelperFile(fileName) {
+			continue
+		}
+
 		linter.RunLinterRule(support.ErrorSev, fpath, validateAllowedExtension(fileName))
 
 		// We only apply the following lint rules to yaml files
@@ -219,14 +224,19 @@ func validateTemplatesDir(templatesPath string) error {
 	return nil
 }
 
-func validateAllowedExtension(fileName string) error {
+// checks if the file starts with '_'.
+func isHelperFile(fileName string) bool {
 	baseName := filepath.Base(fileName)
+	ext := filepath.Ext(fileName)
 
-	if strings.HasPrefix(baseName, "_") {
-		fmt.Println("skipping _ files")
-		return nil
+	if strings.HasPrefix(baseName, "_") && ext != ".yaml" && ext != ".yml" {
+		return true
 	}
 
+	return false
+}
+
+func validateAllowedExtension(fileName string) error {
 	ext := filepath.Ext(fileName)
 	validExtensions := []string{".yaml", ".yml", ".tpl", ".txt"}
 
--- a/internal/chart/v3/lint/rules/template.go
+++ b/internal/chart/v3/lint/rules/template.go
@@ -123,12 +123,9 @@ func TemplatesWithSkipSchemaValidation(linter *support.Linter, values map[string
 		fileName := template.Name
 		fpath = fileName
 
-		// skip the linting rules if the file starts with _ and is not a yml/yaml file
-		if isHelperFile(fileName) {
-			continue
+		if !isHelperFile(fileName) {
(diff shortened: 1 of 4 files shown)


internal/chart/v3/lint/rules/template.go
```diff
@@ -220,6 +220,13 @@ func validateTemplatesDir(templatesPath string) error {
 }
 
 func validateAllowedExtension(fileName string) error {
```
TerryHowe: I don't think the name `validateAllowedExtension` makes a lot of sense if it is used this way. It probably makes more sense to add a method named something like `isHelperFile` and return bool and call that instead.

TerryHowe (changes requested):
I haven't worked on lint a lot, so I am not sure if this is a desired feature even. Thanks for the pull request though!

internal/chart/v3/lint/rules/template.go
Qu-Ack: You're right. isHelperFile makes the intent a lot clearer. I will update the code accordingly

Qu-Ack:
I abstracted the functionality in a IsHelperFile function, any files with extension yaml/yml will not be considered helper files even if they start with '_'

pkg/chart/v2/lint/rules/template.go
```diff
@@ -123,6 +123,11 @@ func TemplatesWithSkipSchemaValidation(linter *support.Linter, values map[string
 		fileName := template.Name
 		fpath = fileName
 
+		// skip the linting rules if the file starts with _ and is not a yaml/yml file
+		if isHelperFile(fileName) {
+			continue
```
brightbyte: This will skip *all* linting for helper files. Linting of the file content should still be applied (if it's a yaml file). The only thing that should be skipped if it's a helper file is validateAllowedExtension().

internal/chart/v3/lint/rules/template.go
```diff
@@ -123,6 +123,11 @@ func TemplatesWithSkipSchemaValidation(linter *support.Linter, values map[string
 		fileName := template.Name
 		fpath = fileName
 
+		// skip the linting rules if the file starts with _ and is not a yml/yaml file
+		if isHelperFile(fileName) {
+			continue
```
brightbyte: I don't think this is correct, see my comment on the v2 version of the code.

pkg/chart/v2/lint/rules/template.go
Qu-Ack: I think there might be a misunderstanding. 
'_helpers.yml' -> would still go through normal linting
'_regular.json' -> would skip all linting
'_hello.lua' -> would skip all linting

my code produces the correct results for the above examples. 
Am I missing something really obvious here?

internal/chart/v3/lint/rules/template.go
```diff
@@ -123,6 +123,11 @@ func TemplatesWithSkipSchemaValidation(linter *support.Linter, values map[string
...
+		}
+
 		linter.RunLinterRule(support.ErrorSev, fpath, validateAllowedExtension(fileName))
```
TerryHowe:
```suggestion
		linter.RunLinterRule(support.ErrorSev, fpath, (!isHelperFile(fileName) && validateAllowedExtension(fileName)))
```

pkg/chart/v2/lint/rules/template.go
brightbyte: Ah right, I misunderstood the behavior of "is helper file"... I would expect it to return true for e.g. _helpers.yml - to me at least, that is also a "helper file" (because it is not a manifest file).

From what I can tell, you code works, but it's a little confusing to me. But then, I'm new to helm, and I have no experise or authority on this code base :)

What I would suggest is this: there should be two conditions which are independent of each other: one for "is a helper file" and one for "is a yaml file". If it is not a helper file, make sure it has one of the allowed file extensions. If it is a yaml file, do yaml lint checks.

This seems more clear to me: a certain type of check is conditional on a certain propery of the file.

brightbyte: Oh, right, yes - I missed the fact that isHelperFile returns false for yaml files. So I suppose you are correct, your code does the right thing.

I find it a bit confusing thoug. I think it would be more clear to have two conditions that control two behaviors: 
* "is helper file" controls whether the file extension should be checked.
* "is yaml file" controls whether the yaml lin rules should apply

Keeping these two things separate would avoid confusion I think. But then, I'm new to helm, and I have no experience or authority in this code base...

In any case, thank you for working on this! I would make my life easier if it was merged :)

PR #31307, opened 2025-09-18
fix: Ignore absolute path when RepoUrl is provided

### LINKED ISSUE #7862: Bug: `--repo` should takes precedence over local folder

TerryHowe:

Ignore absolute path when RepoUrl is provided

Closes: https://github.com/helm/helm/issues/7862

**If applicable**:

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -828,7 +828,7 @@ func urlEqual(u1, u2 *url.URL) bool {
 // This does not ensure that the chart is well-formed; only that the requested filename exists.
 //
 // Order of resolution:
-// - relative to current working directory
+// - relative to current working directory when --repo flag is not presented
 // - if path is absolute or begins with '.', error out here
 // - URL
 //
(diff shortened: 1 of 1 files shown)


pkg/action/install.go
```diff
@@ -817,20 +817,26 @@ func (c *ChartPathOptions) LocateChart(name string, settings *cli.EnvSettings) (
...
+			if err != nil {
+				return abs, err
+			}
+			if c.Verify {
+				if _, err := downloader.VerifyChart(abs, abs+".prov", c.Keyring); err != nil {
+					return "", err
+				}
+				return abs, nil
 			}
+			if filepath.IsAbs(name) || strings.HasPrefix(name, ".") {
```
gjenkins8: Why is it wanted/needed to check if name is an absolute or hidden file here?

```
helm create foo
./bin/helm template `realpath foo`
Error: path "/Users/gjenkins8/Source/helm/helm/foo" not found
```

This seems to be wrong behavior?

TerryHowe: Bad cherry pick I think

gjenkins8 (approved):

joejulian (approved):

PR #31311, opened 2025-09-21
fix: accept semver range in ValidateReference

### LINKED ISSUE #31304: `helm pull/show` doesn't work with `--version` range`
### What happened?

https://helm.sh/docs/helm/helm_pull/ says that the `--version` flag may use a valid range like `^2.0.0`, but it doesn't appear to be usable:

```
helm show chart redis --repo=https://charts.bitnami.com/bitnami --version="^22.0.0"
Error: chart reference and version mismatch: ^22.0.0 is not 22.0.7
```

The version constraint will resolve to a chart package and pull/show it

`helm show chart redis --repo=https://charts.bitnami.com/bitnami --version="^22.0.0"`

### Helm version

<details>
```console
$ helm version
version.BuildInfo{Version:"v3.19.0", GitCommit:"3d8990f0836691f0229297773f3524598f46bda6", GitTreeState:"clean", GoVersion:"go1.24.7"}
```
</details>

### Kubernetes version

N/A

kyokuping:

Closes #31304

This PR removes incorrect version validation that was blocking usage of semantic version ranges.

When a user provide a semantic version range for the  `--version` option, the validation is already performed at `GetTagMatchingVersionOrConstraint`. Therefore, addititonal validtion logic was not only unncessary but also prevented the semantic version range feature.

Support for semantic version range when a url includes a `digest` is still doesn't avaliable in this PR. this will be improve in future PRs. 

**If applicable**:

### DIFF
--- a/pkg/registry/client.go
+++ b/pkg/registry/client.go
@@ -835,10 +835,6 @@ func (c *Client) ValidateReference(ref, version string, u *url.URL) (string, *ur
 	if version == "" {
 		// Use OCI URI tag as default
 		version = registryReference.Tag
-	} else {
-		if registryReference.Tag != "" && registryReference.Tag != version {
-			return "", nil, fmt.Errorf("chart reference and version mismatch: %s is not %s", version, registryReference.Tag)
-		}
 	}
 
 	if registryReference.Digest != "" {


TerryHowe (changes requested):
This change disables all version checking and we need to verify thtat the version specified match in this case. I think you would replace this with some regular expression matching.

Better yet, just do not use `--version` when using oci references.

TerryHowe:
This pull request makes a change to OCI charts and the issue mentioned is for the older chart repositories.

kyokuping:
I'll check the details later and update/close as needed

TerryHowe:
Reopen if you have something new, thanks just the same.

PR #31312, opened 2025-09-21
Remove unused 'Settings' from plugin schema

gjenkins8:
Removes unused "Settings" from plugin messages. And in particular, dependency on the `cli` package causes a painful circular dependency if the "Settings" object is to maintain a reference to loaded plugins in the future.

**If applicable**:

### DIFF
--- a/internal/plugin/schema/cli.go
+++ b/internal/plugin/schema/cli.go
@@ -15,13 +15,10 @@ package schema
 
 import (
 	"bytes"
-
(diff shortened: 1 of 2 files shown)


TerryHowe (approved):
/lgtm

scottrigby (changes requested):
removing comment, crossed wires with another change.

scottrigby (approved):

PR #31314, opened 2025-09-22
feat: Allow deep merge of lists with an optional index

jmleddy:

This PR is unit tests on top of pr #30632.

This PR allows merging or composition of lists inside values files, which is a very useful feature when using overlays to simplify/merge deployment files, e.g. merging kubernetes environment variables from several values files.

It has the following behaviour:

* Behaviour is unchanged for any lists whose key name does not start with `\*`
* Lists of plain values are concatenated when the key name is prefixed by `\*`
* Lists of maps are concatenated when the key name is prefixed by `\*`, allowing to specify a unique key by prefixing the key with `\*` for merging specific entries where desired

If matched, the `\*` prefix will be stripped from key names of list properties in the final rendering

**Addresses**:
* https://github.com/helm/helm/issues/3486
* https://github.com/fluxcd/helm-controller/issues/300

I can add checks for empty maps or arrays, but as @rsafonseca mentioned in prior review, I don't think there's any way for the yaml library to demarshal such a type.

**If applicable**:
- [x] this PR contains user facing changes (the `docs needed` label should be applied, couldn't figure out how)


jmleddy:
@miledxz so you're aware. These were the tests you asked for for merging list functionality

TerryHowe:
It may be that this should be a pull request on `./internal/chart/v3` I'm not sure how we are handling that.

jmleddy:
@TerryHowe who would know how we're handling it? For context I just added unit tests on what @rsafonseca started

jmleddy:
activity occurs

rm3l:
Hi team,
Any chance we can get this and/or #30632 merged? This would significantly reduce user confusion around how Helm handles lists.
Thanks.

PR #31315, opened 2025-09-22
Remove unused golangci-lint rules that produce warning

benoittgt:
Avoid noisy output when running lint locally:

```sh
❯ golangci-lint run ./...
WARN [runner/exclusion_rules] Skipped 0 issues by rules: [Text: "var-naming: avoid meaningless package names", Linters: "revive"]
```

It looks like we don’t need this exclusion rule anymore.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -42,11 +42,6 @@ linters:
       - legacy
       - std-error-handling
 
-    rules:
-      - linters:
-          - revive
-        text: 'var-naming: avoid meaningless package names'
-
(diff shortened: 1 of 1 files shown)


gjenkins8 (approved):

benoittgt:
Maybe a version difference? I have the issue with v2.0.2 or v2.1.0. Thanks for the merge

mattfarina:
Please note, this rule is being used. See #31330 to restore it.

PR #31318, opened 2025-09-23
fix: Helm v4 release distribution & `get-helm-3` script

gjenkins8:
Fixes the impending situation where when Helm 4 is released, users of `get-helm-3` will actually get Helm v4 installed!

This PR:
1. Adds a check to ensure the major version matches the destination: ie. `v4` -> `helm4-latest-version`
2. Removes the write to `helm-latest-version` (see below)
3. Updates `get-helm-3` to use `helm3-latest-version`

As currently the latest version of Helm is written to `helm-latest-version` and `helm3-latest-version`. And awkwardly, the `get-helm-3` script reads `helm-latest-version` to determine the latest version to install.

I removed the write to `helm-latest-version` out of an abundance of caution. We should IMHO return this once we are confident consumers of `get-helm-3` have updated.

**If applicable**:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -49,8 +49,11 @@ jobs:
           # Push the latest semver tag, excluding prerelease tags
           LATEST_VERSION="$(git tag | sort -r --version-sort | grep '^v[0-9]' | grep -v '-' | head -n1)"
           echo "LATEST_VERSION=${LATEST_VERSION}"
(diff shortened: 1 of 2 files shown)


benoittgt (approved):

TerryHowe (approved):
/lgtm

mattfarina (approved):

PR #31320, opened 2025-09-23
provenance: allow RSA signing when ed25519 keys are present (switch to ProtonMail/go-crypto)

kosiew:

Helm previously failed to sign charts or plugins with RSA keys when an Ed25519 key was present in the same keyring. The underlying OpenPGP reader aborted with the error `unsupported feature: public key type: 22` when encountering Ed25519 material, preventing signing even though the RSA key was available.

This PR fixes that by:

* Replacing imports of `golang.org/x/crypto/openpgp` with the maintained fork `github.com/ProtonMail/go-crypto/openpgp` which has broader public key algorithm support and is compatible with Ed25519 material.
* Adding a small compatibility change when verifying signatures (`bytes.NewReader` in place of `bytes.NewBuffer` and passing a default PGP config) so verification uses the expected reader and config.
* Adding a unit test (`TestMixedKeyringRSASigningAndVerification`) and a mixed-key keyring (`testdata/helm-mixed-keyring.pub`) to reproduce and guard against regressions where a keyring contains both RSA and Ed25519 keys.
* Updating `go.mod` / `go.sum` to include `github.com/ProtonMail/go-crypto v1.3.0` and `github.com/cloudflare/circl v1.6.0` (indirect) required to build and parse modern OpenPGP material.

This addresses the issue where Helm fails to sign when an Ed25519 key is present in the keyring by ensuring the OpenPGP implementation used can read Ed25519 material and allowing RSA signing to continue.

**Change summary (files changed)**:

* `go.mod` / `go.sum`: add `github.com/ProtonMail/go-crypto` and updated indirect deps
* `internal/plugin/signing_info.go`: switch clearsign import to ProtonMail fork
* `pkg/provenance/sign.go`: switch openpgp imports to ProtonMail fork; use `bytes.NewReader` and pass `defaultPGPConfig` to signature verification
* `pkg/provenance/sign_test.go`: add `TestMixedKeyringRSASigningAndVerification` and import updates for test helpers
* `pkg/provenance/testdata/helm-mixed-keyring.pub`: new test keyring containing RSA + Ed25519 public keys

* Security/compatibility: Replacing the OpenPGP import is non-trivial — please review the change surface for potential behavioral differences (parsing, verification details, default config values). The ProtonMail fork is widely used as a drop-in replacement for `golang.org/x/crypto/openpgp`, but an audit of the dependency bump is recommended.

* Backwards compatibility: The intent is to preserve signing and verification behavior for existing RSA-only workflows. The added test verifies that an RSA signer can sign and verify even when Ed25519 public key material is present in the same keyring.

* Tests: A unit test was added to prevent regressions. Please run `go test ./pkg/provenance -v` and a repository-wide `go test ./...` as part of review.

* Reproducer: The new test `TestMixedKeyringRSASigningAndVerification` reproduces the original problem locally by loading a keyring with both RSA and Ed25519 public keys and asserting signing/verification succeeds using the RSA private key.

* Notable code points:

  * `pkg/provenance/sign.go:verifySignature` — switched to `bytes.NewReader` and passed `&defaultPGPConfig` to `openpgp.CheckDetachedSignature` to ensure the reader and config used match expectations.
  * `internal/plugin/signing_info.go` and other files updated to import the ProtonMail fork in place of `golang.org/x/crypto/openpgp`.

**Manual verification steps**:

1. `git checkout -b fix/provenance-mixed-keyring`
2. `go mod tidy`
3. `go test ./pkg/provenance -v` (verify new test passes)
4. Optionally reproduce the original error by attempting to sign a chart with a keyring that contains both an RSA private key and an Ed25519 public key — confirm signing succeeds.

**If applicable**:

* [ ] this PR contains user facing changes (the `docs needed` label should be applied if so)
* [x] this PR has been tested for backwards compatibility


mattfarina (approved):
lgtm

pkg/provenance/sign_test.go
```diff
@@ -266,6 +270,85 @@ func TestClearSign(t *testing.T) {
...
+			if subkey.PublicKey != nil && subkey.PublicKey.PubKeyAlgo == packet.PubKeyAlgoEdDSA {
+				hasEdDSA = true
+				break
+			}
+		}
+
+		if hasEdDSA {
+			break
+		}
+	}
```
gjenkins8, line 300: I think this could have been written something like (for future reference):
```
hasEdDSA := slices.ContainsFunc(signer.KeyRing, func(entity Entity) {
   if entity.PrimaryKey != nil && entity.PrimaryKey.PubKeyAlgo == packet.PubKeyAlgoEdDSA) {
     return true
   }
     
   return slices.Contains(entity.Subkeys, func(subkey SubKey) {
     return subkey.PublicKey != nil && subkey.PublicKey.PubKeyAlgo == packet.PubKeyAlgoEdDSA
   })
})
```

(a lot more compact and tidy IMHO)

pkg/provenance/sign_test.go
```diff
@@ -266,6 +270,85 @@ func TestClearSign(t *testing.T) {
...
+
+	if !hasEdDSA {
+		t.Fatalf("expected %s to include an Ed25519 public key", testMixedKeyring)
+	}
```
gjenkins8, line 304: testify/assert would clean these up:

```
assert.True(t, hasEdDSA)
```

and e.g. below
```
assert.NotNil(t, signer.Entity)
```

(messages become optional IMHO, as the single line assert typically conveys enough info)

gjenkins8 (approved):

pkg/provenance/sign_test.go
TerryHowe: I created https://github.com/helm/helm/pull/31326 as a follow up to address this

kosiew:
Thanks @mattfarina , @gjenkins8 for your review and feedback.

PR #31331, opened 2025-09-25
coalesce: respect empty-map overrides for sub-keys when merging values

### LINKED ISSUE #31291: Values file set to "{}" cannot overwrite
### What happened?

We have a helm chart with following values.yaml, like this
```
affinity:
  nodeAffinity:
    requiredDuringSchedulingIgnoredDuringExecution:
      nodeSelectorTerms:
        - matchExpressions:
            - key: kubernetes.io/arch
              operator: In
              values:
                - amd64
                - ppc64le
nodeSelector:
  kubernetes.io/arch: amd64
  type: worker
tolerations:
  - key: "dedicated"
    operator: "Equal"
    value: "backup"
    effect: "NoSchedule"
  - key: "special"
    operator: "Exists"
    effect: "NoExecute"
    tolerationSeconds: 3600
```

Helm chart template is

```
  {{- if and .Values.nodeSelector (kindIs "map" .Values.nodeSelector) (not (empty .Values.nodeSelector)) }}
  nodeSelector: {{- toYaml .Values.nodeSelector | nindent 4 }}
  {{- end }}
  {{- if and .Values.affinity (kindIs "map" .Values.affinity) (not (empty .Values.affinity)) }}
  affinity:
  {{- toYaml .Values.affinity | nindent 4 }}
  {{- end }}
  {{- if and .Values.tolerations (kindIs "slice" .Values.tolerations) (not (empty .Values.tolerations)) }}
  tolerations:
  {{- toYaml .Values.tolerations | nindent 4 }}
  {{- end }}
```

Now, when we pass  `affinity` or any of the above three fields using a values-override.yaml file
```
affinity: {}
```

we are still seeing the default values that are set in values.yaml file

If I have a `values-override.yaml` file and I'm passing it using the helm install command, where I just want to overwrite my default values using `{}`  the default values should not be set. 

```
affinity: {}
```

Although I have already conducted some research on the topic and found a couple of similar issues.
1. https://github.com/helm/helm/issues/5407 (this is the same issue that I'm facing  and it mentioned of using `null` which works as expcted.) 
2. https://github.com/minio/operator/issues/2255#issuecomment-2297333690 (though it quite similar to null before once yaml unmarshal it done it is converted to null)

Also while I was trying to use the `--set` flag to overwrite those default in that case it was working with the above template that I shared.

I check the code as well and I think the reason why this is happning is how [loader.MergeMaps()](https://github.com/helm/helm/blob/main/pkg/cli/values/options.go#L59) works

1. Below if what values.yaml file looks like

```
affinity:
  nodeAffinity:
    requiredDuringSchedulingIgnoredDuringExecution:
      nodeSelectorTerms:
        - matchExpressions:
            - key: kubernetes.io/arch
              operator: In
              values:
                - amd64
                - ppc64le
nodeSelector:
  kubernetes.io/arch: amd64
  type: worker
# Tolerations to allow scheduling on tainted nodes
tolerations:
  - key: "dedicated"
    operator: "Equal"
    value: "backup"
    effect: "NoSchedule"
  - key: "special"
    operator: "Exists"
    effect: "NoExecute"
    tolerationSeconds: 3600
```

2. Create a values-override.yaml file

```
affinity: {}
```

3. Run helm template/install command

```
helm template mychart . -f values-override.yaml > generated.yaml
```

4. After checking the `generated.yaml` file we see the affinity values are set though in values-override.yaml file we try setting it to `{}`

Though if I directly use `--set` flag I can see it works as expected, no default values are getting set for affinity which is what we want.

```
helm template mychart .  --set affinity={}
```
Also `null` works in both cases when I'm using `--set` flag or using it in values-overide.yaml file 

### Helm version

```
$ helm version                                       
version.BuildInfo{Version:"v3.18.3", GitCommit:"6838ebcf265a3842d1433956e8a622e3290cf324", GitTreeState:"clean", GoVersion:"go1.24.4"}
```

### Kubernetes version

```
kubectl version           
Client Version: v1.33.2
Kustomize Version: v5.6.0
Server Version: v1.30.4+k3s1
WARNING: version difference between client (1.33) and server (1.30) exceeds the supported minor version skew of +/-1
```

kosiew:
Closes #31291 

This PR fixes a bug in Helm's value coalescing where an explicitly empty map provided in an overrides file (for a *sub-key*) did not suppress the chart defaults. In other words, when a user specifies `someKey: {}` in a `values-override.yaml`, chart defaults under `someKey` could still be merged into the final values, which is surprising and inconsistent with the behavior of `--set someKey={}`.

This change makes the coalescer treat an *empty destination map* as an intentional override for non-top-level keys: if the destination map is empty and we are coalescing a sub-key (i.e. `prefix != ""`) and we are *not* in merge mode, we short-circuit and do not copy defaults from the source map. Top-level coalescing is unchanged (top-level `values` should still inherit chart defaults when appropriate).

Why this matters:

* Users commonly expect `key: {}` in an overrides YAML to remove chart defaults under `key`.
* Previously this only worked reliably when using `--set key={}`, causing an inconsistency between CLI and file-based overrides.
* Fixing this makes file-based overrides behave the same as `--set` and aligns with user expectations.

---

## Summary of changes

* **Code:** Added a short-circuit in `coalesceTablesFullKey` to return early when:

  * destination (`dst`) is empty, and
  * we're operating on a sub-key (`prefix != ""`), and
  * we are *not* merging (`!merge`).

  This prevents copying defaults from `src` into an intentionally-empty `dst` map.

* **Tests:** Added unit tests that assert the new behavior:

  * `TestCoalesceValuesEmptyMapOverride` — validates that an empty map in overrides removes default keys for that key, and `nil` removes top-level keys.
  * Subtest in `TestCoalesceTables` — validates `CoalesceTables(dst, src)` leaves an intentionally-empty `dst["config"]` empty when `src` contains defaults.

Relevant files changed:

* `pkg/chart/common/util/coalesce.go` (implementation)
* `pkg/chart/common/util/coalesce_test.go` (new unit tests)

Small illustrative diff (conceptual):

```diff
@@
-    if dst == nil {
-        return src
-    }
+    if dst == nil {
+        return src
+    }
+    // If dst is empty and we're operating on a sub-key (prefix != ""),
+    // bail out when not merging so that child charts don't inherit defaults
+    // for an explicitly empty table in the parent. We only do this for
+    // non-top-level coalescing (prefix != "") because top-level coalescing
+    // still needs to pull chart defaults across when dst is empty.
+    if len(dst) == 0 && prefix != "" && !merge {
+        return dst
+    }
```

---

* Behavior change: this modifies how empty maps are treated for sub-keys. It intentionally aligns file-based overrides behavior with `--set`-style overrides; please review for any places that intentionally relied on the previous semantics.
* The change is narrow and guarded by the `prefix != ""` and `!merge` checks to limit effect surface area.
* Unit tests have been added to guard the expected behavior. Please run the test suite and verify nothing else regresses.

---

## How to reproduce / test manually

1. Create a chart with `values.yaml` containing nested defaults, e.g.:

```yaml
config:
  enabled: true
  port: 8080
```

2. Create an override file `values-override.yaml` with an empty map under `config`:

```yaml
config: {}
```

3. Render the templates using the chart and the override file:

```bash
helm template ./mychart --values values-override.yaml
```

Expected result after this PR: no `config.enabled` or `config.port` defaults should be present in the rendered outputs for the override scope (the empty `config` should suppress defaults). Prior to this PR, those defaults might still be present.

Also test `toDelete: null` behaviour in overrides to ensure nil explicitly removes a key.

---

## Backwards compatibility

* This is a behavioral fix that makes Helm treat explicit empty maps in YAML overrides as an intentional removal of nested defaults for non-top-level keys. This could be considered a breaking change for charts or users that accidentally relied on the previous merging behavior. However, the change increases consistency with CLI `--set` behavior and is more consistent with user expectations, so it is implemented as a bug fix.

* Top-level merging behavior is intentionally preserved.

---

## If applicable

* [ ] this PR contains user facing changes (the `docs needed` label should be applied if so)
* [x] this PR has been tested for backwards compatibility


scottrigby (changes requested):
The results are not as they should be. This behavior with `--set key={}` is incorrect too.

In both Helm 3 (current release `3.19.0`) and Helm 4 dev (`main`), the template output for `--set key={}` is:
```yaml
key:
- ""
```

And this PR does replicate that behavior with a custom values file. Eg:
```yaml
# file: custom-values.yaml
key: {}
```

gives the same output as `--set key={}` now.

However, that output is not correct. We want an empty map, but instead we get a list with one empty item.

We get the same output when the custom values file contains:
```yaml
# file: custom-values.yaml
key: [""]
```

The output should be:
```yaml
key: {}
```

instead of:
```yaml
key:
- ""
```

kosiew:
@scottrigby 
I have fixed the above.

PR #31337, opened 2025-09-26
Fixing rollback and uninstall client WaitStrategy

### LINKED ISSUE #31317: Incorrect error message when GetWaiter fails during rollback
### What happened?

When rollback fails because the waiter is not set correctly, the following error message prints, even though the failure did not happen while setting the metadata visitor:

"Unable to set metadata visitor from target release: unknown wait strategy"

Expected to get an error message like
"Unable to get waiter for release: unknown wait strategy"

Here's where this happens:
https://github.com/helm/helm/blob/5b43b744b83adab08e8ebbf007452b97d49586fd/pkg/action/rollback.go#L236C2-L239C3

I imagine this error message was just copy-pasted from higher in the file where the metadata visitor was being set, and the error was never updated to fit within its context. 

This is where the metadata visitor is actually being set:
https://github.com/helm/helm/blob/5b43b744b83adab08e8ebbf007452b97d49586fd/pkg/action/rollback.go#L203

N/A

### Helm version

N/A. Using the api directly

### Kubernetes version

<details>

```console
$ kubectl version
Client Version: v1.34.0
Kustomize Version: v5.7.1
Server Version: v1.32.7
```

</details>

rachelvweber:
Correctly sets a WaitStrategy for the Uninstall client created when `helm install` rolls back and sets the WaitStrategy for the Rollback client created when `helm upgrade` rolls back. Also fixes an incorrect error message that occurs when the WaitStrategy is not set.

Rollback for `helm install` always fails with "unknown wait strategy".  Rollback for `helm upgrade` fails with the same error message unless the WaitStrategy is set to HookOnlyStrategy.

closes #31322
closes #31317

**If applicable**:

### DIFF
--- a/pkg/action/upgrade.go
+++ b/pkg/action/upgrade.go
@@ -563,9 +563,7 @@ func (u *Upgrade) failRelease(rel *release.Release, created kube.ResourceList, e
 
 		rollin := NewRollback(u.cfg)
 		rollin.Version = filteredHistory[0].Version
-		if u.WaitStrategy == kube.HookOnlyStrategy {
-			rollin.WaitStrategy = kube.StatusWatcherStrategy
-		}
+		rollin.WaitStrategy = u.WaitStrategy
 		rollin.WaitForJobs = u.WaitForJobs
 		rollin.DisableHooks = u.DisableHooks
 		rollin.ForceReplace = u.ForceReplace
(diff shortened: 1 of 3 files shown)


pkg/action/upgrade.go
```diff
@@ -563,9 +563,7 @@ func (u *Upgrade) failRelease(rel *release.Release, created kube.ResourceList, e
 
 		rollin := NewRollback(u.cfg)
 		rollin.Version = filteredHistory[0].Version
-		if u.WaitStrategy == kube.HookOnlyStrategy {
```
rachelvweber, line 566: This default should already be set for the upgrade client on line 178. As long as the rollback client is set to the same WaitStrategy as the upgrade client, this will be set correctly.

TerryHowe (approved):
/lgtm

robertsirc (approved):
LGTM

Issue #12707, opened 2024-01-11
Mitchellh Dependencies


mattfarina:
mitchellh has moved on from Go development and has archived his repos (list [here](https://gist.github.com/mitchellh/90029601268e59a29e64e55bab1c5bdc)). In addition to that gox has been archived. Helm uses gox and some dependencies including:

- github.com/mitchellh/gox
- github.com/mitchellh/go-wordwrap
- github.com/mitchellh/reflectwalk
- https://github.com/helm/helm/pull/31342

We need to handle these to ensure we are using projects that being maintained (e.g., there is no way a CVE could be reported to any of these).

TerryHowe:
go-wordwrap is an indirect dependency:
```
 % grep go-wordwrap go.mod
	github.com/mitchellh/go-wordwrap v1.0.1 // indirect
```

```
% go mod why github.com/mitchellh/go-wordwrap
# github.com/mitchellh/go-wordwrap
helm.sh/helm/v4/pkg/cmd
k8s.io/kubectl/pkg/cmd/get
k8s.io/kubectl/pkg/util/templates
github.com/mitchellh/go-wordwrap
```

I assume kubectl will fix that eventually.

TerryHowe:
go-reflectwalk is a copystructure dependency that will go way when we get rid of that dependency and sprig does the same:
```
% grep github.com/mitchellh/reflectwalk go.mod
	github.com/mitchellh/reflectwalk v1.0.2 // indirect
% go mod why github.com/mitchellh/reflectwalk
# github.com/mitchellh/reflectwalk
helm.sh/helm/v4/pkg/action
github.com/Masterminds/sprig/v3
github.com/mitchellh/copystructure
github.com/mitchellh/reflectwalk
```

TerryHowe:
Sprig issue https://github.com/Masterminds/sprig/issues/456

Shlomit-B:
Hey, I’d like to remove the remaining gox dependency by replacing the build-cross target with a go build loop across platforms. Does that sound okay?

TerryHowe:
Thanks @Shlomit-B but there is an outstanding PR for that

PR #31342, opened 2025-09-29
chore: replace github.com/mitchellh/copystructure

TerryHowe:

The github.com/mitchellh/copystructure repository has been archived and is no longer maintained.

Partial: https://github.com/helm/helm/issues/12707

I looked at several other alternatives to replace it and opted to implement an internal replacement because for example:

* encoding/json can be used for deep copy, but it is hacky and performance is terrible
* k8s.io/apimachinery has a deep copy, but it seemed a bit outside the scope of what that primarily does
* https://github.com/brunoga/deep is not very popular, but nice simple implementation
* The implementation of the functionality needed is pretty simple

The implementation in github.com/mitchellh/copystructure has many features which we were not using, so I expected and got better performance

```
go test -bench=. -benchmem ./internal/copystructure/
goos: darwin
goarch: arm64
pkg: helm.sh/helm/v4/internal/copystructure
cpu: Apple M1
BenchmarkInternalCopy-8              	  271351	      4429 ns/op	    5264 B/op	     111 allocs/op
BenchmarkMitchellhCopy-8             	   61238	     19645 ns/op	   15008 B/op	     433 allocs/op
BenchmarkInternalCopyLargeMap-8      	    1364	    889158 ns/op	  808347 B/op	   19022 allocs/op
BenchmarkMitchellhCopyLargeMap-8     	     315	   3784941 ns/op	 2130339 B/op	   75045 allocs/op
BenchmarkInternalCopyDeepNested-8    	   44197	     27205 ns/op	   27648 B/op	     651 allocs/op
BenchmarkMitchellhCopyDeepNested-8   	    5334	    226008 ns/op	   95231 B/op	    2696 allocs/op
PASS
```

Benchmark that is not included in PR https://gist.github.com/TerryHowe/f53f97016389c6d9cc849f573487d44c

**If applicable**:

### DIFF
--- a/internal/copystructure/copystructure.go
+++ b/internal/copystructure/copystructure.go
@@ -1,0 +1,120 @@
+/*
+Copyright The Helm Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package copystructure
+
+import (
+	"fmt"
+	"reflect"
+)
+
+// Copy performs a deep copy of the given interface{}.
+// This implementation handles the specific use cases needed by Helm.
+func Copy(src interface{}) (interface{}, error) {
+	if src == nil {
+		return make(map[string]interface{}), nil
+	}
+	return copyValue(reflect.ValueOf(src))
+}
+
+// copyValue handles copying using reflection for non-map types
+func copyValue(original reflect.Value) (interface{}, error) {
+	switch original.Kind() {
+	case reflect.Bool, reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32,
+		reflect.Int64, reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32,
+		reflect.Uint64, reflect.Uintptr, reflect.Float32, reflect.Float64,
+		reflect.Complex64, reflect.Complex128, reflect.String, reflect.Array:
+		return original.Interface(), nil
+
+	case reflect.Interface:
+		if original.IsNil() {
+			return original.Interface(), nil
+		}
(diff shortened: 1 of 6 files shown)


TerryHowe:
The reason for the indirect dependency:
```
% go mod why github.com/mitchellh/copystructure 
# github.com/mitchellh/copystructure
helm.sh/helm/v4/pkg/action
github.com/Masterminds/sprig/v3
github.com/mitchellh/copystructure
```

benoittgt:
That's a great idea ! Thanks for providing benchmarks too.

internal/copystructure/copystructure.go
```diff
@@ -0,0 +1,120 @@
...
+package copystructure
+
+import (
+	"fmt"
+	"reflect"
+)
+
+// Copy performs a deep copy of the given interface{}.
+// This implementation handles the specific use cases needed by Helm.
+func Copy(src interface{}) (interface{}, error) {
```
gjenkins8: Helm only needs this for deep copy of values ie. copying `map[string]any`. And in particular, 'any' here is only POD types from values.

Would it make sense to restrict our "deep copy" code to meet only these criteria. In particular if this would simplify the implementation and/or provide performance improvements?

gjenkins8: In particular, this reflect based approach was (very) non-performant for Wasm builds:
https://github.com/gjenkins8/copystructure_perf

(which may be of interest in the future)

gjenkins8: I copied the code here over to that repo:
https://github.com/gjenkins8/copystructure_perf/pull/1

The good news is that this new code is ~4-5 times faster (including for Wasm). The bad news is that Wasm is still 22x slower than native.

Overall, this is still a clear win. So also would be happy for further improvements to occur later.

gjenkins8 (dismissed):

internal/copystructure/copystructure.go
gjenkins8:
```suggestion
func Copy(src any) (any, error) {
```

(and others) :)

TerryHowe: Pushed another commit that converts `interface{}` to `any`

gjenkins8: @stefanprodan fyi

gjenkins8 (approved):

scottrigby (approved):
This is a good, incremental improvement, and addresses the package dependency issue :+1: We can follow-up with incremental improvements (for example, potential alternatives to reflect).

PR #31347, opened 2025-10-01
Support oci repo/chart:version and add --ensure-version check

benoittgt:

Closes: https://github.com/helm/helm/issues/11736

- Allow specifying chart and version in OCI push targets, e.g. oci://ghcr.io/org/repo/mychart:1.2.3.
- Same with `--ensure-version` flag as a safety check. If provided and it differs, the command fails.
- Reject digest references for push.

1. Aligns with prior work in #12690, but implements via a small centralized helper and a check-only flag.
2. ~~I am not sure we want to add `--version`, as it seems a little misleading to me. One could expect it to push a specific version. Curious to hear what you think. 🤷 Maybe `:version` is good enough.~~

Tested via:

```sh
$ bin/helm create myapp
Creating myapp

$ sed -i '' 's/^version:.*/version: 0.1.0+meta/' myapp/Chart.yaml

$ bin/helm package myapp
Successfully packaged chart and saved it to: /Users/benoit.tigeot/projects/lifen/helm/myapp-0.1.0+meta.tgz

$ bin/helm push myapp-0.1.0+meta.tgz oci://ghcr.io/benoittgt/my-app 
Pushed: ghcr.io/benoittgt/my-app/myapp:0.1.0_meta
Digest: sha256:e232ddea0dd3f93dfc978df1e3a9db586fce0ef292a7f3f502c494218a8670f3
ghcr.io/benoittgt/my-app/myapp:0.1.0_meta contains an underscore.

OCI artifact references (e.g. tags) do not support the plus sign (+). To support
storing semantic versions, Helm adopts the convention of changing plus (+) to
an underscore (_) in chart version tags when pushing to a registry and back to
a plus (+) when pulling from a registry.

$ bin/helm push myapp-0.1.0+meta.tgz oci://ghcr.io/benoittgt/my-app --ensure-version 0.1.0+meta
Pushed: [...redacted...]

$ bin/helm push myapp-0.1.0+meta.tgz oci://ghcr.io/benoittgt/my-app --ensure-version 0.2.0+meta
Error: specified --ensure-version "0.2.0+meta" does not match chart version "0.1.0+meta"

$ bin/helm push myapp-0.1.0+meta.tgz oci://ghcr.io/benoittgt/my-app:0.1.0+meta
Pushed: [...redacted...]

$ bin/helm push myapp-0.1.0+meta.tgz oci://ghcr.io/benoittgt/my-app:0.2.0+meta
Error: tag "0.2.0+meta" does not match provided chart version "0.1.0+meta"
```

**If applicable**:

### DIFF
--- a/pkg/pusher/ocipusher.go
+++ b/pkg/pusher/ocipusher.go
@@ -59,6 +59,11 @@ func (pusher *OCIPusher) push(chartRef, href string) error {
 		return err
 	}
 
+	// If an expected version is specified via --version, enforce it matches Chart.yaml
+	if pusher.opts.expectedVersion != "" && pusher.opts.expectedVersion != meta.Metadata.Version {
+		return fmt.Errorf("specified --version %q does not match chart version %q", pusher.opts.expectedVersion, meta.Metadata.Version)
+	}
+
 	ref, err := registry.BuildPushRef(href, meta.Metadata.Name, meta.Metadata.Version)
 	if err != nil {
 		return err
--- a/pkg/pusher/ocipusher.go
+++ b/pkg/pusher/ocipusher.go
@@ -61,7 +61,7 @@ func (pusher *OCIPusher) push(chartRef, href string) error {
 
 	// If an expected version is specified via --version, enforce it matches Chart.yaml
 	if pusher.opts.expectedVersion != "" && pusher.opts.expectedVersion != meta.Metadata.Version {
-		return fmt.Errorf("specified --version %q does not match chart version %q", pusher.opts.expectedVersion, meta.Metadata.Version)
+		return fmt.Errorf("specified --ensure-version %q does not match chart version %q", pusher.opts.expectedVersion, meta.Metadata.Version)
 	}
 
 	ref, err := registry.BuildPushRef(href, meta.Metadata.Name, meta.Metadata.Version)
--- a/pkg/pusher/ocipusher.go
+++ b/pkg/pusher/ocipusher.go
@@ -59,7 +59,7 @@ func (pusher *OCIPusher) push(chartRef, href string) error {
 		return err
 	}
 
-	// If an expected version is specified via --version, enforce it matches Chart.yaml
+	// If an expected version is specified via --ensure-version, enforce it matches Chart.yaml
 	if pusher.opts.expectedVersion != "" && pusher.opts.expectedVersion != meta.Metadata.Version {
 		return fmt.Errorf("specified --ensure-version %q does not match chart version %q", pusher.opts.expectedVersion, meta.Metadata.Version)
 	}
--- a/pkg/pusher/ocipusher.go
+++ b/pkg/pusher/ocipusher.go
@@ -22,8 +22,6 @@ import (
 	"net"
 	"net/http"
 	"os"
-	"path"
-	"strings"
 	"time"
 
 	"helm.sh/helm/v4/internal/tlsutil"
@@ -61,6 +59,11 @@ func (pusher *OCIPusher) push(chartRef, href string) error {
 		return err
 	}
 
+	ref, err := registry.BuildPushRef(href, meta.Metadata.Name, meta.Metadata.Version)
+	if err != nil {
+		return err
+	}
+
 	client := pusher.opts.registryClient
 	if client == nil {
 		c, err := pusher.newRegistryClient()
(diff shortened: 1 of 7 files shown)


stevehipwell:
> I am not sure we want to add --version, as it seems a little misleading to me. One could expect it to push a specific version. Curious to hear what you think. 🤷 Maybe :version is good enough.

I like the inclusion of the `--version` flag, I don't think it's confusing given the push command it targeting a specific artifact and it's consistent with install where the flag can be used as well as an OCI tag. From an automation perspective keeping the version as a separate input makes the whole flow clearer, e.g. `helm push "${artifact}" "${repository}" --version "${version}"`.

benoittgt:
We had a discussion about this new feature today during the [developer meeting](https://docs.google.com/document/d/1d-6xJEx0C78csIYSPKJzRPeWaHG_8W1Hjl72OJggwdc/edit?pli=1&tab=t.0). I received some good feedback, as well as some suggestions from @scottrigby about the flag name (https://github.com/helm/helm/pull/31350).

benoittgt:
I’m okay with letting it die; there doesn’t seem to be much interest.

stevehipwell:
@benoittgt is there a reason why this is no longer progressing?

benoittgt:
@stevehipwell it can happen if the PR has low interest or if maintainers have no time to review. ;)

stevehipwell:
> @benoittgt is there a reason why this is no longer progressing?

@benoittgt I meant specifically to this PR; I know PRs here can take time to progress.

benoittgt:
I will rework on this one.

benoittgt:
We should be good now.

stevehipwell (approved):
LGTM

TerryHowe:
Might need to rebase to trigger DCO

benoittgt:
Thanks @TerryHowe. DCO seems fine now.

pkg/registry/pushref.go
```diff
@@ -0,0 +1,62 @@
...
+package registry
+
+import (
+	"fmt"
+	"path"
+	"strings"
+)
+
+func BuildPushRef(href, chartName, chartVersion string) (string, error) {
+	ref, err := newReference(href)
```
TerryHowe, line 26: This needs to work for `helm push chart.tgz oci://localhost:5000` case or that might break people.

pkg/registry/pushref.go
```diff
@@ -0,0 +1,62 @@
...
+
+	// Normalize for comparison only; the registry client converts '+' to '_' for the actual tag
+	normalizedVersion := strings.ReplaceAll(chartVersion, "+", "_")
+	// if href tag present, it must match normalized chart version
+	if ref.Tag != "" && ref.Tag != normalizedVersion {
+		normalizedTag := strings.ReplaceAll(ref.Tag, "_", "+")
+		return "", fmt.Errorf("tag %q does not match provided chart version %q", normalizedTag, chartVersion)
+	}
+
+	// Ensure repository ends with the chart name once (avoid duplication)
```
TerryHowe, line 43: While you could argue this is an improvement, dedup is going to have to wait for a major release as that would be a breaking change.

pkg/pusher/ocipusher.go
```diff
@@ -61,6 +59,16 @@ func (pusher *OCIPusher) push(chartRef, href string) error {
 		return err
 	}
 
+	// If an expected version is specified via --ensure-version, enforce it matches Chart.yaml
+	if pusher.opts.expectedVersion != "" && pusher.opts.expectedVersion != meta.Metadata.Version {
+		return fmt.Errorf("specified --ensure-version %q does not match chart version %q", pusher.opts.expectedVersion, meta.Metadata.Version)
```
TerryHowe, line 64:
```suggestion
		return fmt.Errorf("expected chart version %q does not match chart version %q", pusher.opts.expectedVersion, meta.Metadata.Version)
```

pkg/pusher/ocipusher.go
```diff
@@ -61,6 +59,16 @@ func (pusher *OCIPusher) push(chartRef, href string) error {
...
+	}
+
+	ref, err := registry.BuildPushRef(href, meta.Metadata.Name, meta.Metadata.Version)
```
TerryHowe, line 67: On the dedup topic maybe you could trigger if expectedVersion is set, dedup. That would kind of not be a breaking change and we'd get that logic in there.

pkg/registry/pushref.go
```diff
@@ -0,0 +1,62 @@
...
+*/
+
+package registry
+
+import (
+	"fmt"
+	"path"
+	"strings"
+)
+
```
TerryHowe, line 24: Doc comment here

pkg/registry/pushref_test.go
```diff
@@ -0,0 +1,120 @@
...
+			want:     "",
+			wantErr:  true,
+		},
+	}
+
+	for _, tt := range tests {
+		if tt.wantErr {
+			t.Run(tt.name, func(t *testing.T) {
+				_, err := BuildPushRef(tt.registry, tt.chart, tt.version)
+				if err == nil {
```
TerryHowe, line 102: Maybe testify if it isn't too much of a hassle.

TerryHowe (changes requested):

PR #31296, opened 2025-09-15
build: set kube version via `debug.BuildInfo`

gjenkins8:
**What this PR does / why we need it**:
Read the Kubernetes client version (the `client-go` library version) directly from from Go's internal `runtime.DebugInfo`. Rather than setting the link time build variables.

The PR is an update of https://github.com/helm/helm/pull/13180 with some (needed) changes

Thanks to @branchvincent for the idea and initial implementation

**Special notes for your reviewer**:

**If applicable**:
- [ ] this PR contains user facing changes (the `docs needed` label should be applied if so)
- [ ] this PR contains unit tests
- [x] this PR has been tested for backwards compatibility

PR #31351, opened 2025-10-01
feat: `helm version` print Kubernetes (client-go) version

gjenkins8:
Enable the output of `helm version` to print the Kubernetes client library version (client-go version) that Helm was built against. This allows Helm users to easily discover this information, which can be useful for understanding Kubernetes compatibility.

This conflicts with https://github.com/helm/helm/pull/31296 (this PR/change would need to be adapted to be based on this method eventually, depending on which is merged first)

Also note, test coverage is limited. As these values change "dynamically". So is not easy to validate the actual expected version (see the modification to `version.txt`)

**If applicable**:


benoittgt:
Good idea !

Seeing tests failures, do we want to print only when we have a clear version or the issue is somewhere else?
```
            WANT:
            'version.BuildInfo{Version:"v4.0", GitCommit:"", GitTreeState:"", GoVersion:""}
            '
            
            GOT:
            'version.BuildInfo{Version:"v4.0", GitCommit:"", GitTreeState:"", GoVersion:"", KubeClientVersion:"v."}
            '
```

Edit: I was more thinking about printing the version at all of empty string, rather than "v" but I got my answer. :)

benoittgt:
Thanks, LGTM

TerryHowe (approved):
/lgtm

robertsirc (approved):
LGTM

mattfarina (approved):
lgtm

PR #31352, opened 2025-10-02
fix: Use PAX tar format for accurate timestamps

argggh:

Go's `archive/tar` package rounds archive member file timestamps to the closest second unless the POSIX.1-2001 "PAX" tar format is used (see https://pkg.go.dev/archive/tar#Format). This results in noise in processes where the helm charts are immediately extracted by (at least) GNU tar for repackaging/modification/inspection, where tar complains (if the original creation time ended up being rounded up) that `time stamp 2025-10-02 12:06:47 is 0.469095762 s in the future` or similar. This behavior is unlikely to be changed in the the `archive/tar` package; ref comments in golang/go#48275. Using the PAX header format resolves this.

An alternative fix is to truncate the file mtimes to the nearest second in the past using `ModTime: time.Now().Truncate(time.Second)`, as also suggested in the golang issue referenced above.

### DIFF
--- a/internal/chart/v3/util/save.go
+++ b/internal/chart/v3/util/save.go
@@ -233,6 +233,7 @@ func writeToTar(out *tar.Writer, name string, body []byte) error {
 		Mode:    0644,
 		Size:    int64(len(body)),
 		ModTime: time.Now(),
+		Format:  tar.FormatPAX,
 	}
 	if err := out.WriteHeader(h); err != nil {
 		return err
--- a/internal/chart/v3/util/save_test.go
+++ b/internal/chart/v3/util/save_test.go
@@ -137,10 +137,7 @@ func Indent(n int, text string) string {
 }
 
 func TestSavePreservesTimestamps(t *testing.T) {
-	// Test executes so quickly that if we don't subtract a second, the
-	// check will fail because `initialCreateTime` will be identical to the
-	// written timestamp for the files.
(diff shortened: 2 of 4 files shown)


benoittgt:
I think this is a good idea, and the fact that the tests are now cleaned up reinforces that this is the right path forward. :D

I dug into the history and never saw `FormatPAX` suggested.

TerryHowe (approved):
/lgtm

TerryHowe:
https://serverfault.com/questions/250511/which-tar-file-format-should-i-use

TerryHowe:
https://en.wikipedia.org/wiki/Tar_(computing) it looks like it has been supported in gnu tar since 2004 although I did some discussion about the BSD pax command not supporting pax format 🤷

argggh:
https://mgorny.pl/articles/portability-of-tar-features.html this is an analysis from 2018. If backwards compatibility is a concern, I can rework the patch to rather just always floor the creation time to the nearest second in the past.

banjoh (approved):

TerryHowe:
Any change this could be revived?

argggh:
> Any change this could be revived?

@TerryHowe I can rebase this; have we settled on whether we are happy with the PAX format or if we would rather go with a more defensive approach of just truncating timestamps?

TerryHowe:
I'm sure @mattfarina would have an opinion on this

benoittgt:
I think this is still an interesting feature.

PR #31357, opened 2025-10-05
fix: ClientOnly Configuration State issue in Install Action

### LINKED ISSUE #11463: Install action does not install anything if ran previously with ClientOnly flagged
With the current Action Go package APIs, I am having an issue with the [install action](https://github.com/helm/helm/blob/f08db5c899c4e312c72ae863acbb90ac8a4747ed/pkg/action/install.go#L487).

The issue I am facing is that if I run two install actions, the first with `ClientOnly` flag set to `true`, also the second one does not reach the cluster.

Here's the code of the 1st action that runs to produce template files in the `outputDir`:
```go
client := action.NewInstall(settings)
client.DryRun = true
client.ReleaseName = "my-test-release"
client.Replace = true
client.ClientOnly = true
client.OutputDir = outputDir
client.UseReleaseName = true
_, err = client.RunWithContext(ctx, chart, values)
```

Then, when the flow proceeds, I run another install action with the aim of actually installing the release:
```go
client := action.NewInstall(settings)
client.ReleaseName = releaseName
client.Namespace = "my-test-namespace"
client.DryRun = false
client.ClientOnly = false
client.Wait, client.WaitForJobs, client.Timeout = release.GetWaitConditions()
rel, err = client.RunWithContext(ctx, chart, values)
```

The second action does not produce any error but it actually does not deploy anything.
If, in exactly the same flow, the first install action runs with `ClientOnly = false`, the second install works as expected and actually installs the release into the cluster.

adityaraj178:
closes #11463 

Fixes a bug in Helm's Go API where setting ClientOnly = true on an install action would permanently modify the shared action.Configuration, causing subsequent actions using the same configuration to fail.

**Problem**
When using Helm's Go API with multiple install actions sharing the same `action.Configuration`, there was a state pollution issue:
1. **First action** sets `ClientOnly = true` (e.g., for template generation)
2. **Helm internally** replaces the real Kubernetes client with a fake client in the shared configuration
3. **Configuration remains modified** after the first action completes
4. **Subsequent actions** inherit the corrupted state and fail, even with `ClientOnly = false`

Changes made:
Implemented a configuration state preservation mechanism that:
  - Saves the original configuration state before ClientOnly modifications
  - Restores the original state when the function returns (using defer)
  - Ensures no permanent pollution of shared configuration

**If applicable**:

### DIFF
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -138,6 +138,55 @@ func NewConfiguration(options ...ConfigurationOption) *Configuration {
 	return c
 }
 
+// ConfigurationState represents a saved state of Configuration that can be restored
+type ConfigurationState interface {
+	// unexported method to ensure only our implementation can satisfy this interface
+	restore(*Configuration)
+}
+
+// configurationSnapshot represents a saved state of Configuration for restoration
+type configurationSnapshot struct {
+	originalKubeClient   kube.Interface
+	originalCapabilities *common.Capabilities
+	originalReleases     *storage.Storage
+}
+
+// restore implements ConfigurationState interface
+func (cs *configurationSnapshot) restore(cfg *Configuration) {
+	cfg.KubeClient = cs.originalKubeClient
+	cfg.Capabilities = cs.originalCapabilities
+	cfg.Releases = cs.originalReleases
+}
+
+// SaveState creates a snapshot of the current configuration state
+func (cfg *Configuration) SaveState() ConfigurationState {
+	cfg.mutex.Lock()
+	defer cfg.mutex.Unlock()
+
+	return cfg.saveStateUnsafe()
+}
+
+// saveStateUnsafe creates a snapshot without locking (internal use only)
+func (cfg *Configuration) saveStateUnsafe() ConfigurationState {
+	return &configurationSnapshot{
+		originalKubeClient:   cfg.KubeClient,
+		originalCapabilities: cfg.Capabilities,
+		originalReleases:     cfg.Releases,
+	}
+}
+
+// RestoreState restores the configuration to a previously saved state
+func (cfg *Configuration) RestoreState(snapshot ConfigurationState) {
+	if snapshot == nil {
+		return
+	}
+
+	cfg.mutex.Lock()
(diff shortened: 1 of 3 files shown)


pkg/action/action.go
```diff
@@ -96,6 +96,31 @@ type Configuration struct {
 	mutex sync.Mutex
 }
 
+// ConfigurationSnapshot represents a saved state of Configuration for restoration
+type ConfigurationSnapshot struct {
```
benoittgt: Do it need to be exported ?

benoittgt (commented):
I'm wondering about thread-safety. Right now, concurrent ClientOnly and non-ClientOnly operations on the same cfg could race ? 🤔

pkg/action/action.go
adityaraj178: > I'm wondering about thread-safety. Right now, concurrent ClientOnly and non-ClientOnly operations on the same cfg could race ? 🤔

I didn't thought about this that time, I'll use the existing cfg.mutex in the SaveState() and RestoreState() methods to prevent race conditions.

> Do it need to be exported ?

`ConfigurationSnapshot ` doesn't need to be exported  since it's only used internally by the Configuration methods. I'll change it to` configurationSnapshot`  to make it package-private.

Will fix both issues in the next commit. Thanks for the thorough review!

adityaraj178:
> #30833 cleaned this up somewhat (it standardized on a "DryRun" flag/option, rather than ClientOnly). Does this fix your issue? (I didn't get a chance to look into your issue)

Thanks for the heads up about #30833! I checked and while it nicely refactored `ClientOnly `→ `DryRunStrategy`, the core state pollution bug still exists.
The problem still persist in `install.go`, just an example

Race Condition Scenario:
Thread A: Runs install with `DryRunClient`→ replaces `cfg.KubeClient` with fake client
Thread B: Runs install with `DryRunNone`→ expects real Kubernetes client but finds fake client
Result: Thread B fails because shared config was polluted

The core issue just moved from `if i.ClientOnly` to `if !interactWithServer(i.DryRunStrategy)` - same bug, different condition.

PR #31362, opened 2025-10-07
Clarify the intent of the resource instructions

fabiocarneiro:
This PR refactors the boilerplate comment in values.yaml to clarify the intent behind the recommendation on setting resource requests and limits.

The current advice to omit default resources is a critical best practice for publicly distributed charts, as it ensures maximum portability across unknown user environments with varying resource policies and capacities. However, this guidance can be misleading for the large number of developers creating charts solely for internal use, where setting a baseline is a valid and often beneficial strategy.   

This change clarifies that the recommendation is primarily aimed at public charts, providing better context for all users and reducing ambiguity for those developing in a controlled, internal environment.

I have assumed that this was the original intent, but maybe it wasn't.

**None of the checks were applicable as this is documentation only**

### DIFF
--- a/internal/chart/v3/util/create.go
+++ b/internal/chart/v3/util/create.go
@@ -218,9 +218,10 @@ httpRoute:
   #       value: v2
 
 resources: {}
-  # We usually recommend not to specify default resources and to leave this as a conscious
-  # choice for the user. This also increases chances charts run on environments with little
-  # resources, such as Minikube. If you do want to specify resources, uncomment the following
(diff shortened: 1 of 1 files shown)


robertsirc (approved):
LGTM

scottrigby:
Added `docs needed` label because we may want to look at our docs to see if these improvements should make their way there too. Also, should we update `https://github.com/helm/helm/blob/main/pkg/chart/v2/util/create.go` with this same comment?

PR #31372, opened 2025-10-08
Enable Releases To Have Multiple Versions

mattfarina:
**What this PR does / why we need it**: Releases are tightly coupled to v1 charts. This change enables new versions of releases that can have different chart api versions. A new release version is not yet included. This change changes the API to enable that.

**If applicable**:

### DIFF
--- a/pkg/action/get_values.go
+++ b/pkg/action/get_values.go
@@ -16,7 +16,13 @@ limitations under the License.
 
 package action
 
-import "helm.sh/helm/v4/pkg/chart/common/util"
+import (
+	"fmt"
+
+	"helm.sh/helm/v4/pkg/chart/common/util"
+	release "helm.sh/helm/v4/pkg/release"
+	rspb "helm.sh/helm/v4/pkg/release/v1"
+)
 
 // GetValues is the action for checking a given release's values.
 //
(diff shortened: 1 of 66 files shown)


pkg/action/get_values.go
```diff
@@ -56,3 +67,18 @@ func (g *GetValues) Run(name string) (map[string]interface{}, error) {
...
+// releaserToV1Release is a helper function to convert a v1 release passed by interface
+// into the type object.
+func releaserToV1Release(rel release.Releaser) (*rspb.Release, error) {
+	switch r := rel.(type) {
+	case rspb.Release:
+		return &r, nil
+	case *rspb.Release:
+		return r, nil
+	case nil:
+		return nil, nil
```
banjoh, line 80: Since `nil` is valid, callers of this function now need safeguards.

pkg/cmd/root.go
```diff
@@ -465,3 +466,31 @@ type CommandError struct {
...
+// releaserToV1Release is a helper function to convert a v1 release passed by interface
+// into the type object.
+func releaserToV1Release(rel ri.Releaser) (*release.Release, error) {
+	switch r := rel.(type) {
+	case release.Release:
+		return &r, nil
+	case *release.Release:
+		return r, nil
+	case nil:
+		return nil, nil
```
banjoh, line 479: Since `nil` is valid, callers of this function now need safeguards.

banjoh (commented):
Minor comment on converters returning `nil` which is expected since its a valid input param. Callers would need to ensure the check is made. Maybe a godoc comment would be good to have

banjoh (approved):

pkg/action/get_values.go
mattfarina: What if someone passes `nil` to this function? Should it return an error or `nil`? This is why the case was added. There should be checks in the code for nil all around this.

Thoughts?

scottrigby (approved):
good job. enabling charts v3 👏 we can rework internals and add docs later

PR #31383, opened 2025-10-12
Avoid accessing .Items on nil object

dirkmueller:
When listing fails for whatever reason, the return value is nil, err. so handle err explicitly.


TerryHowe (changes requested):
DCO issue, other than that, looks good

benoittgt (approved):

mattfarina (approved):
lgtm

PR #31384, opened 2025-10-13
deepcopy original before upgradeClientSideFieldManager

wallyatgithub:

This PR made a deep copy of original before upgradeClientSideFieldManager, and use the deep copy to call upgradeClientSideFieldManager

why we need it?

original and target share the same object, this is done by requireAdoption

While upgradeClientSideFieldManager will call Get() on original, which will refresh original with live info from cluster, and the consequence is, target is refreshed as well, as it shares the same object with original

So next when applying target to cluster, the patch data will have the info loaded from cluster, such as managedfields (if SSA is already used in install), status, etc., and can cause unexpected consequences, such as patch rejection due to non-nil managedfields as "metadata.managedFields must be nil."

From k8s source code:

```
// Apply implements Manager.
func (f *structuredMergeManager) Apply(liveObj, patchObj runtime.Object, managed Managed, manager string, force bool) (runtime.Object, Managed, error) {
        // Check that the patch object has the same version as the live object
        if patchVersion := patchObj.GetObjectKind().GroupVersionKind().GroupVersion(); patchVersion != f.groupVersion {
                return nil, nil,
                        errors.NewBadRequest(
                                fmt.Sprintf("Incorrect version specified in apply patch. "+
                                        "Specified patch version: %s, expected: %s",
                                        patchVersion, f.groupVersion))
        }

        patchObjMeta, err := meta.Accessor(patchObj)
        if err != nil {
                return nil, nil, fmt.Errorf("couldn't get accessor: %v", err)
        }
        if patchObjMeta.GetManagedFields() != nil {
                return nil, nil, errors.NewBadRequest("metadata.managedFields must be nil")
        }

```
This is only applicable for helm4 where SSA is enabled, no impact to helm3 as helm3 does not have SSA

**If applicable**:

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -736,7 +736,10 @@ func (c *Client) Update(originals, targets ResourceList, options ...ClientUpdate
 					slog.String("gvk", target.Mapping.GroupVersionKind.String()))
 
 				if updateOptions.upgradeClientSideFieldManager {
-					patched, err := upgradeClientSideFieldManager(original, updateOptions.dryRun, updateOptions.fieldValidationDirective)
+					// target and original shares same Object, deep copy original to avoid changing target
+					copiedOriginal := *original
+					copiedOriginal.Object = original.Object.DeepCopyObject()
+					patched, err := upgradeClientSideFieldManager(&copiedOriginal, updateOptions.dryRun, updateOptions.fieldValidationDirective)
 					if err != nil {
 						slog.Debug("Error patching resource to replace CSA field management", slog.Any("error", err))
 						return err


wallyatgithub:
> hi (and thanks!). Can you describe how this issue manifests in greater detail please? And how to reproduce? (you mention `requireAdoption`?)

Hi, 

I am using helm4 module to install a helm chart like this, 

```
	install := action.NewInstall(actionConfig)
	install.ReleaseName = releaseName
	install.Namespace = nameSpace
	install.DryRun = false
	install.WaitStrategy = kube.LegacyStrategy
	install.WaitForJobs = true
	install.TakeOwnership = true
	install.Timeout = time.Duration(timeout) * time.Second
	if useSSA {
		install.ServerSideApply = true
		install.ForceConflicts = true
	} else {
		install.ForceReplace = true
	}
	install.DisableHooks = disableHooks

	chartRequested, err := loader.Load(chartPath)
	if err != nil {
		fmt.Println("failed to load chart: ", err)
		return nil, err
	}

	rel, err := install.Run(chartRequested, values)

```

but some of the resource to be installed already exists in cluster, so it will run into this:

```

		if i.TakeOwnership {
			toBeAdopted, err = requireAdoption(resources)
```

here toBeAdopted is the original and resources is the target, and they actually share same objects respectively as that is a interface by default will not deep copy when pushing element from resource to toBeAdopted

Then it finally runs into this because I enabled serverSideApply and somehow upgradeClientSideFieldManager is hardcoded to true today for install, so it will query from cluster to produce patch to upgrade from CSA to SSA, and this will cause Get() called on original which refreshed the object, and thus target is also refreshed, and making target object populated with live info from cluster like Managementfield, and this gets into the apply patch finally when target is applied to cluster and then rejected due to non-nil Managementfield
```

		} else if updateOptions.serverSideApply {
			slog.Debug(
				"using server-side apply for resource update",
				slog.Bool("forceConflicts", updateOptions.forceConflicts),
				slog.Bool("dryRun", updateOptions.dryRun),
				slog.String("fieldValidationDirective", string(updateOptions.fieldValidationDirective)),
				slog.Bool("upgradeClientSideFieldManager", updateOptions.upgradeClientSideFieldManager))
			return func(original, target *resource.Info) error {

				logger := slog.With(
					slog.String("namespace", target.Namespace),
					slog.String("name", target.Name),
					slog.String("gvk", target.Mapping.GroupVersionKind.String()))
				if updateOptions.upgradeClientSideFieldManager {
					patched, err := upgradeClientSideFieldManager(original, updateOptions.dryRun, updateOptions.fieldValidationDirective)
					if err != nil {
						slog.Debug("Error patching resource to replace CSA field management", slog.Any("error", err))
						return err
					}
```

wallyatgithub:
Although I did not test, but I think an install using CSA and an upgrade then using SSA will also trigger the same problem, as it would cause ClientUpdateOptionUpgradeClientSideFieldManager set to true in this case, and will finally run into the same branch around upgradeClientSideFieldManager
```

	upgradeClientSideFieldManager := isReleaseApplyMethodClientSideApply(originalRelease.ApplyMethod) && serverSideApply // Update client-side field manager if transitioning from client-side to server-side apply
	results, err := u.cfg.KubeClient.Update(
		current,
		target,
		kube.ClientUpdateOptionForceReplace(u.ForceReplace),
		kube.ClientUpdateOptionServerSideApply(serverSideApply, u.ForceConflicts),
		kube.ClientUpdateOptionUpgradeClientSideFieldManager(upgradeClientSideFieldManager))
```

PR #31155, opened 2025-08-18
feature: add nightly shuffle test

TerryHowe:
**What this PR does / why we need it**:

Worked on my fork https://github.com/TerryHowe/helm/actions/runs/17072588177

Should wait for:

- [x] https://github.com/helm/helm/pull/31199
- [x] https://github.com/helm/helm/pull/31198
- [x] https://github.com/helm/helm/pull/31200
- [x] https://github.com/helm/helm/pull/31349
- [ ] https://github.com/helm/helm/pull/31353
- [x] https://github.com/helm/helm/pull/31354
- [ ] https://github.com/helm/helm/pull/31377

Closes: https://github.com/helm/helm/issues/31016
Closes: https://github.com/helm/helm/issues/30892

**Special notes for your reviewer**:

**If applicable**:
- [ ] this PR contains user facing changes (the `docs needed` label should be applied if so)
- [ ] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

Issue #31016, opened 2025-06-29
Test with intermittent failures (Flaky Tests)


robertsirc:
### What happened?

Working with some package maintainers for helm they have identified a few test that will fail randomly. 

cmd/helm/dependency_build_test.go
cmd/helm/dependency_update_test.go
cmd/helm/pull_test.go
cmd/helm/registry_login_test.go 

### What did you expect to happen?

That tests are somewhat consistent in how and when they run.

### How can we reproduce it (as minimally and precisely as possible)?

As this is intermittent there is no reproducible steps to recreate this.

### Helm version

<details>
```console
$ helm version
# paste output here
```
</details>

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

TerryHowe:
Core dumps or just FAIL?

robertsirc:
What I am told they just fail, the system is here in [OBS](https://build.opensuse.org/package/show/Virtualization:containers/helm). I need to fork this and then change the spec to build these and get the results.

gjenkins8:
I can easily reproduce failures with `-shuffle=on` on the `dev-v3` branch:

```
$ git switch dev-v3 && git pull
$ go clean -testcache; go test  -run . -shuffle=on ./cmd/helm    # may require a couple of runs
...
--- FAIL: TestLoadPlugins (0.00s)
    plugin_test.go:124: Expected 5 plugins, got 0
--- FAIL: TestLoadPluginsForCompletion (0.00s)
    plugin_test.go:280: Expected commands [{args [] [] []} {echo [] [] []} {env [] [global] [{list [] [a all log] []} {remove [all one] [] []}]} {exitwith [] [] [{code [] [a b] []}]} {fullenv [] [q z] [{empty [] [] []} {full [] [] [{less [] [a all] []} {more [one two] [b ball] []}]}]}], got []
"foo" has been added to your repositories
"bar" has been added to your repositories
"baz" has been added to your repositories
time="2025-07-02T21:57:15.560657-07:00" level=error msg="traces export: Post \"https://localhost:4318/v1/traces\": dial tcp [::1]:4318: connect: connection refused" go.version=go1.24.4
FAIL
FAIL    helm.sh/helm/v3/cmd/helm        15.408s
FAIL
```

I can't reproduce on `main` (and keep in mind, majority of `cmd/helm` logic was moved to `pkg/cmd`).

TerryHowe:
[...]

PR #31396, opened 2025-10-15
Increase test count from 1 to 2 for improved reliability

### LINKED ISSUE #30892: Run Helm unit-tests with `-shuffle=on` and `-count=n` (n > 1)

jtaylortech:
## Summary
Increases the test count in TESTFLAGS from 1 to 2 to better detect flakey tests and order-dependent issues as recommended in issue #30892.

## Changes
- Updated  in Makefile to use  instead of 
- Tests will now run twice per execution, helping identify:
  - Order-dependent tests
  - Flakey/inconsistent test results
  - Race conditions

## Testing
The  flag was already in place, so combined with , the test suite will:
1. Randomize test execution order
2. Run each test twice
3. Surface any tests that fail inconsistently

Closes #30892

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ ACCEPTANCE_RUN_TESTS = .
 PKG         := ./...
 TAGS        :=
 TESTS       := .
-TESTFLAGS   := -shuffle=on -count=1
+TESTFLAGS   := -shuffle=on -count=2
 LDFLAGS     := -w -s
 GOFLAGS     :=
 CGO_ENABLED ?= 0


Makefile
```diff
@@ -21,7 +21,7 @@ ACCEPTANCE_RUN_TESTS = .
 PKG         := ./...
 TAGS        :=
 TESTS       := .
-TESTFLAGS   := -shuffle=on -count=1
```
gjenkins8, line 24: please see: https://github.com/helm/helm/issues/30892#issuecomment-3030678499

TerryHowe (changes requested):
This would cause a lot of problems right now.

jtaylortech:
> This would cause a lot of problems right now.

@TerryHowe Thanks for the feedback .. I've updated the PR to make TESTCOUNT configurable instead of hardcoded:
  - Local development now defaults to -count=1 (fast iteration, no changes to existing behavior)
  - CI explicitly passes TESTCOUNT=2 to catch flaky tests
  - Developers can still override locally if needed: `make test TESTCOUNT=2`

This addresses the concerns about slowing down local development while still providing the benefits in CI.

Let me know if this approach works or if you had anything else in mind, as your original comment was very vague.

Makefile
jtaylortech: @gjenkins8 I've implemented the suggested approach from the discussions mentioned and made TESTCOUNT configurable via environment variable, defaulting to `-count=1 `for local dev, with CI overriding to `TESTCOUNT=2`. 

What do you think about the change? Also, thanks for the detailed feedback!

.github/workflows/build-test.yml
```diff
@@ -31,6 +31,6 @@ jobs:
       - name: Check if go modules need to be tidied
         run: go mod tidy -diff
       - name: Run unit tests
-        run: make test-coverage
+        run: make test-coverage TESTCOUNT=2
```
TerryHowe: I don't think test-coverage would use TESTCOUNT.

There was a discussion about this in a dev meeting a while back and I think it is reflected in George's comment he highlight.

One of the outcomes of the discussion was creating a periodic shuffle job https://github.com/helm/helm/pull/31155/files which is on hold waiting for the flakes to shake out.

What we could use help on is fixes for the various flakes. There are to at least listed at the bottom of  https://github.com/helm/helm/issues/31016

jtaylortech: Thanks for clarifying, @TerryHowe! 
I'll close this PR & go take a look at some of the ones listed in #31016.

PR #31400, opened 2025-10-17
feature: make helm plugin install error more helpful.

### LINKED ISSUE #31398: Helm v4: Unhelpful error `Cannot detect VCS` when trying to `helm plugin install` with invalid file path

```
# Built from HEAD
$ helm version
version.BuildInfo{Version:"v3.19.0", GitCommit:"3d8990f0836691f0229297773f3524598f46bda6", GitTreeState:"clean", GoVersion:"go1.25.1"}

$ helm plugin install non-existing-file
Error: Cannot detect VCS
```

The error message should be actionable by saying what's wrong and what options the user has.

### Why is this needed?

Plugin installation supports multiple sources. As a fallback to installing from a plugin directory, other things are tried. The error should maybe include what was tried, since users don't expect errors for other source types.

haojingcn:
**Summary**:
Some issues(#31398) indicate that returning "Cannot detect CVS" during the helm plugin install process is very unfriendly. 
1. You don't know if it's a problem in your input process or in the access process. 
2. In addition, adding -debug in the command line will not display more information that can help you troubleshoot.

For example:
```
$ helm plugin install diff
Error: Cannot detect VCS
exit status 1

$ helm plugin install diff --debug      
Error: Cannot detect VCS
exit status 1
```

Changes:
- Add the checkVCSReference method to check the validity of the VCS source input before creating the installer object (OCI, Local, HTTP all have checks, but VCS does not).
   - Add some testCase for checkVCSReference methods, including support for git, bzr, SVN, hg, basically covering the github.com/Masterminds/vcs all supported type.
- Add a little bit of debug information during the process of new plugin installer.

Closes #31398

**If applicable**:

### DIFF
--- a/internal/chart/v3/lint/rules/crds.go
+++ b/internal/chart/v3/lint/rules/crds.go
@@ -80,8 +80,10 @@ func Crds(linter *support.Linter) {
 				return
 			}
 
-			linter.RunLinterRule(support.ErrorSev, fpath, validateCrdAPIVersion(yamlStruct))
-			linter.RunLinterRule(support.ErrorSev, fpath, validateCrdKind(yamlStruct))
+			if yamlStruct != nil {
+				linter.RunLinterRule(support.ErrorSev, fpath, validateCrdAPIVersion(yamlStruct))
+				linter.RunLinterRule(support.ErrorSev, fpath, validateCrdKind(yamlStruct))
+			}
 		}
 	}
 }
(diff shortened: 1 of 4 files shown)


AndiDog:
Sorry, didn't see this PR exists and I posted https://github.com/helm/helm/pull/31408 as another solution for the error message. The PRs do different things but could be combined depending on what maintainers want.

PR #13095, opened 2024-06-08
fix(LocateChart): Fixed install.LocateChart func to support remote chart while local folder with the same name exists

idsulik:
**What this PR does / why we need it**:
It fixes this issue https://github.com/GoogleContainerTools/skaffold/issues/9347 when you want to install remote chart, but you have a local folder with the chart name. 

**Steps to reproduce the behavior**
1. git clone https://github.com/google/turbinia.git
2. from the root of the repo execute `helm install dev-release turbinia --repo https://google.github.io/osdfir-infrastructure/`
3. Note: the git repo contains a folder named **turbinia** which is the same as the chart name

**It will output**

> Error: INSTALLATION FAILED: Chart.yaml file is missing

PR #31403, opened 2025-10-18
feat: add latest version and outdated flag to plugin list command

idsulik:
Added `LATEST` column to plugin list output and `--outdated` flag to show only plugins with available updates, helping users identify plugins that need updating.

before:
image

after:
image

### DIFF
--- a/internal/plugin/installer/http_installer.go
+++ b/internal/plugin/installer/http_installer.go
@@ -146,6 +146,11 @@ func (i *HTTPInstaller) Update() error {
 	return fmt.Errorf("method Update() not implemented for HttpInstaller")
 }
 
+// GetLatestVersion fetches the latest version of the plugin.Add a comment on lines R158 to R160Add diff commentMarkdown input:  edit mode selected.WritePreviewAdd a suggestionHeadingBoldItalicQuoteCodeLinkUnordered listNumbered listTask listMentionReferenceSaved repliesAdd FilesPaste, drop, or click to add filesCancelCommentStart a reviewReturn to code
+func (i *HTTPInstaller) GetLatestVersion() (string, error) {
+	return "", fmt.Errorf("not supported")
+}
+
 // Path is overridden because we want to join on the plugin name not the file name
 func (i HTTPInstaller) Path() string {
 	if i.Source == "" {
--- a/internal/plugin/installer/http_installer.go
+++ b/internal/plugin/installer/http_installer.go
@@ -146,7 +146,7 @@ func (i *HTTPInstaller) Update() error {
 	return fmt.Errorf("method Update() not implemented for HttpInstaller")
 }
 
-// GetLatestVersion fetches the latest version of the plugin.Add a comment on lines R158 to R160Add diff commentMarkdown input:  edit mode selected.WritePreviewAdd a suggestionHeadingBoldItalicQuoteCodeLinkUnordered listNumbered listTask listMentionReferenceSaved repliesAdd FilesPaste, drop, or click to add filesCancelCommentStart a reviewReturn to code
+// GetLatestVersion fetches the latest version of the plugin.
 func (i *HTTPInstaller) GetLatestVersion() (string, error) {
 	return "", fmt.Errorf("not supported")
 }
--- a/internal/plugin/installer/vcs_installer.go
+++ b/internal/plugin/installer/vcs_installer.go
@@ -110,6 +110,26 @@ func (i *VCSInstaller) Update() error {
 	return nil
 }
 
+func (i *VCSInstaller) GetLatestVersion() (string, error) {
+	_, err := i.Repo.RunFromDir("git", "fetch", "-q", "--tags")
+	if err != nil {
+		return "", fmt.Errorf("failed to fetch tags: %w", err)
+	}
+
+	refs, err := i.Repo.Tags()
+	if err != nil {
+		return "", fmt.Errorf("failed to get tags: %w", err)
+	}
+
+	semvers := getSemVers(refs)
+	if len(semvers) == 0 {
+		return "", fmt.Errorf("no valid semver tags found in repository")
+	}
+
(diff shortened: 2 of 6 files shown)


internal/plugin/installer/http_installer.go
```diff
@@ -146,6 +146,11 @@ func (i *HTTPInstaller) Update() error {
 	return fmt.Errorf("method Update() not implemented for HttpInstaller")
 }
 
+// GetLatestVersion fetches the latest version of the plugin.Add a comment on lines R158 to R160Add diff commentMarkdown input:  edit mode selected.WritePreviewAdd a suggestionHeadingBoldItalicQuoteCodeLinkUnordered listNumbered listTask listMentionReferenceSaved repliesAdd FilesPaste, drop, or click to add filesCancelCommentStart a reviewReturn to code
```
benoittgt: I don't understand this comment. 🤔

pkg/cmd/plugin_list.go
```diff
@@ -51,7 +58,7 @@ func newPluginListCmd(out io.Writer) *cobra.Command {
 			signingInfo := plugin.GetSigningInfoForPlugins(plugins)
 
 			table := uitable.New()
-			table.AddRow("NAME", "VERSION", "TYPE", "APIVERSION", "PROVENANCE", "SOURCE")
+			table.AddRow("NAME", "VERSION", "LATEST", "TYPE", "APIVERSION", "PROVENANCE", "SOURCE")
```
benoittgt: Is it a breaking change to add a new column? I'm thinking about script that parse the output? Maybe it should be scopped behind `o.showOutdated` ?

internal/plugin/installer/vcs_installer.go
```diff
@@ -110,6 +110,26 @@ func (i *VCSInstaller) Update() error {
 	return nil
 }
 
+func (i *VCSInstaller) GetLatestVersion() (string, error) {
```
benoittgt, line 113: The issue I see is it seems we are calling this even if not requested. It should be only when we have `--outdated` flag, no?
Also I think it can be super slow with many plugins to do the `git fetch -q --tags`, also are sure we have `git` always ?

pkg/cmd/plugin_list.go
```diff
@@ -63,7 +70,15 @@ func newPluginListCmd(out io.Writer) *cobra.Command {
 				if info, ok := signingInfo[m.Name]; ok {
 					signedStatus = info.Status
 				}
-				table.AddRow(m.Name, m.Version, m.Type, m.APIVersion, signedStatus, sourceURL)
+
+				latest, err := getLatestVersion(p)
+				if err != nil {
```
benoittgt: Maybe add `--debug` logs with the error ?

pkg/cmd/plugin_list.go
```diff
@@ -63,7 +70,15 @@ func newPluginListCmd(out io.Writer) *cobra.Command {
...
+					latest = "unknown"
+				}
+
+				if !o.showOutdated || (latest != "unknown" && latest != m.Version) {
```
benoittgt: Is the logic correct ? It should be?
```go
if o.showOutdated {
  if latest != "unknown" && latest != m.Version {
      table.AddRow(...)
  }
} else {
  table.AddRow(...)
}
```

benoittgt (commented):
Thanks for the PR but I think we will need some tests, also I'm wondering if it is a good idea to have a feature that is so scoped to VCS plugins? Last one but not least, should we compare real semver parsed and not string?

internal/plugin/installer/http_installer.go
idsulik: fixed

pkg/cmd/plugin_list.go
idsulik: agree, fixed

internal/plugin/installer/vcs_installer.go
idsulik: Started using it only with the `--outdated` flag

pkg/cmd/plugin_list.go
idsulik: fixed

pkg/cmd/plugin_list.go
idsulik: fixed

idsulik:
@gjenkins8 @benoittgt hi! Could you please check the PR?

benoittgt (commented):
LGTM, just a few last comments:
1. Should it be `--show-latest` instead?
2. Maybe it will be hard to understand the `unknown` without debug logs.
3. I’m wondering whether we should prefer stable releases over pre-releases (for example, skip -rc, -beta), or show the highest stable version if available, and only fall back to the highest pre-release when no stable tags exist. 🤷

The only thing that concerns me is the fetching of tags. I know the options are limited, but if you have many plugins and a poor network connection, it could be slow ... although still much faster than manually checking plugin versions. :)

idsulik:
@benoittgt I agree there are many things that we can improve, but we can do it step by step) I'm afraid that if I add more changes, then the PR won't be merged, because usually it takes ages to merge, even if I make small changes.

idsulik:
@gjenkins8 hi! Could you please check the PR?

idsulik:
It is sad when your PRs receive so little attention. You spend your time trying to contribute something useful, but in the end, it feels like that effort leads nowhere.

What is especially disappointing is how little respect there seems to be for other people’s time and work. I contributed in 2024, and those PRs were automatically closed a year later. In 2025, I decided to try again, hoping things might be different, but unfortunately, nothing changed. Once again, even approved PRs were left without any follow-up, and the bot marked them as stale again.

It would be easier to see this as a temporary issue, but in the Helm repository, it seems more like a pattern than an exception. PRs have been automatically closed after a year, before as well, simply because maintainers did not review them. One example is this: https://github.com/helm/helm/pull/13095

I probably will not be contributing here anymore, but I am still grateful for such a great tool.
@mattfarina

PR #31405, opened 2025-10-18
Bug: fix binary plugin url support

### LINKED ISSUE #13221: Install tarball plugin from remote URL
The [documentation for installing Helm plugins](https://helm.sh/docs/topics/plugins/#installing-a-plugin) says that plugins may be installed by specifying the URL to a tarball:
```
You can also install tarball plugins directly from url by issuing helm plugin install https://domain/path/to/plugin.tar.gz
```

However, this does not seem to be the case:
```
$ helm --debug plugin install https://gitlab.com/api/v4/projects/59977310/packages/generic/helm-bump/0.0.1/bump.tar.gz
Error: Cannot detect VCS
helm.go:84: [debug] Cannot detect VCS
```

It would be nice to install plugins from a URL to a tarball but what is the intended behavior? It seems that plugins cannot be installed from tarballs according to #12684.

Helm version: `version.BuildInfo{Version:"v3.14.4", GitCommit:"81c902a123462fd4052bc5e9aa9c513c4c8fc142", GitTreeState:"clean", GoVersion:"go1.21.9"}`

idsulik:
The old PR https://github.com/helm/helm/pull/13225 was closed, so I created a new one(but it'll be better to reopen the old PR to keep the original history)

closes [13221](https://github.com/helm/helm/issues/13221)

A `tar.gz` file can have the MIME type `application/octet-stream`, so I added support for this case.
Example plugin to test the case https://gitlab.com/api/v4/projects/75420382/packages/generic/helm-cel/0.0.1/helm-cel

**If applicable**:
- [N] this PR contains user facing changes (the `docs needed` label should be applied if so)
- [Y] this PR contains unit tests
- [N] this PR has been tested for backwards compatibility

### DIFF
--- a/internal/plugin/installer/extractor.go
+++ b/internal/plugin/installer/extractor.go
@@ -68,6 +68,10 @@ func NewExtractor(source string) (Extractor, error) {
 			return extractor, nil
 		}
 	}
+	if strings.HasPrefix(source, "http") && isGzipArchiveFromURL(source) {
+		return &TarGzExtractor{}, nil
+	}
+
 	return nil, fmt.Errorf("no extractor implemented yet for %s", source)
 }
 
--- a/internal/plugin/installer/extractor.go
+++ b/internal/plugin/installer/extractor.go
@@ -68,8 +68,15 @@ func NewExtractor(source string) (Extractor, error) {
 			return extractor, nil
 		}
 	}
-	if strings.HasPrefix(source, "http") && isGzipArchiveFromURL(source) {
-		return &TarGzExtractor{}, nil
+	if strings.HasPrefix(source, "http") {
+		isGzip, err := isGzipArchiveFromURL(source)
+		if err != nil {
+			return nil, err
+		}
+
+		if isGzip {
+			return &TarGzExtractor{}, nil
+		}
 	}
 
 	return nil, fmt.Errorf("no extractor implemented yet for %s", source)
--- a/internal/plugin/installer/gzip.go
+++ b/internal/plugin/installer/gzip.go
@@ -1,0 +1,50 @@
+/*
+Copyright The Helm Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package installer
+
+import "net/http"
+
+// isGzipArchive checks if data represents a gzip archive by checking the magic bytes
+func isGzipArchive(data []byte) bool {
+	return len(data) >= 2 && data[0] == 0x1f && data[1] == 0x8b
+}
+
+// isGzipArchiveFromURL checks if a URL points to a gzip archive by reading the magic bytes
+func isGzipArchiveFromURL(url string) bool {
+	// Make a GET request to read the first few bytes
+	req, err := http.NewRequest(http.MethodGet, url, nil)
+	if err != nil {
+		return false
+	}
+
+	// Request only the first 512 bytes to check magic bytes
+	req.Header.Set("Range", "bytes=0-511")
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return false
+	}
+	defer resp.Body.Close()
+
+	// Read the first few bytes
+	buf := make([]byte, 2)
+	n, err := resp.Body.Read(buf)
+	if err != nil || n < 2 {
+		return false
+	}
+
+	return isGzipArchive(buf)
+}
--- a/internal/plugin/installer/gzip.go
+++ b/internal/plugin/installer/gzip.go
@@ -15,7 +15,13 @@ limitations under the License.
 
 package installer
 
-import "net/http"
+import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"time"
+)
 
 // isGzipArchive checks if data represents a gzip archive by checking the magic bytes
 func isGzipArchive(data []byte) bool {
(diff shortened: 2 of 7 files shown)


internal/plugin/installer/gzip.go
```diff
@@ -0,0 +1,50 @@
...
+func isGzipArchive(data []byte) bool {
+	return len(data) >= 2 && data[0] == 0x1f && data[1] == 0x8b
+}
+
+// isGzipArchiveFromURL checks if a URL points to a gzip archive by reading the magic bytes
+func isGzipArchiveFromURL(url string) bool {
+	// Make a GET request to read the first few bytes
+	req, err := http.NewRequest(http.MethodGet, url, nil)
+	if err != nil {
+		return false
```
benoittgt: What about debug errors if there is underlying issues ? Ideally it should returns the error and not only a boolean but `isRemoteHTTPArchive` return a boolean too. With the code we don't know it is an network issue or "not a gzip".

internal/plugin/installer/gzip.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	// Request only the first 512 bytes to check magic bytes
+	req.Header.Set("Range", "bytes=0-511")
```
benoittgt, line 44: Does all server support range ? If range are not supported we will download all the file?

internal/plugin/installer/gzip.go
```diff
@@ -0,0 +1,50 @@
...
+
+	resp, err := http.DefaultClient.Do(req)
```
benoittgt, line 46: Maybe a default short timeout here ? With a context ? Also return early if it's not a good http status [206](https://developer.mozilla.org/fr/docs/Web/HTTP/Reference/Status/206) or 200 ?

internal/plugin/installer/extractor.go
```diff
@@ -68,6 +68,10 @@ func NewExtractor(source string) (Extractor, error) {
 			return extractor, nil
 		}
 	}
+	if strings.HasPrefix(source, "http") && isGzipArchiveFromURL(source) {
```
benoittgt: Maybe we should use the same [check](https://github.com/helm/helm/pull/31405/files#diff-3c81ef74b9a5a5d7ab90cc8dce634fb253f775552aad7055d9e2b22a6795d167R177) as in other places?

internal/plugin/installer/gzip.go
```diff
@@ -0,0 +1,50 @@
...
+	if err != nil {
+		return false
+	}
+	defer resp.Body.Close()
+
+	// Read the first few bytes
+	buf := make([]byte, 2)
+	n, err := resp.Body.Read(buf)
```
benoittgt: Maybe we want to use `ReadFull` or `ReadAtLeast` ? So we [will no accept less](https://pkg.go.dev/io#ReadAtLeast) ?

internal/plugin/installer/gzip.go
idsulik: good point, added err return

internal/plugin/installer/gzip.go
idsulik: yes, it'll download all the file if the server doesn't support range

internal/plugin/installer/gzip.go
idsulik: done

internal/plugin/installer/extractor.go
idsulik: I think it'll be redundant, it'll check if the source is gzip or not

internal/plugin/installer/gzip.go
idsulik: done

internal/plugin/installer/gzip.go
```diff
@@ -0,0 +1,66 @@
...
+	"time"
+)
+
+// isGzipArchive checks if data represents a gzip archive by checking the magic bytes
+func isGzipArchive(data []byte) bool {
+	return len(data) >= 2 && data[0] == 0x1f && data[1] == 0x8b
+}
+
+// isGzipArchiveFromURL checks if a URL points to a gzip archive by reading the magic bytes
+func isGzipArchiveFromURL(url string) (bool, error) {
```
gjenkins8, line 32: Before going into details -- it should be possible to download the full archive. Then peek to decide what type of install action to take?

(not suggesting the code is structured to do this today; but thinking about how we might want to fix the issue here without adding more complexity)

idsulik: if it's possible, we should download only the amount of bytes that we need to decide what type file is this, otherwise if can't download range of bytes, we should download the whole file.

idsulik:
@gjenkins8 @benoittgt hi! Any updates here?

benoittgt:
I tested and I had to change something to make it work with the url provided in PR body.

```sh
bin/helm --debug plugin install https://gitlab.com/api/v4/projects/75420382/packages/generic/helm-cel/0.0.1/helm-cel --verify=false
```

```diff
diff --git a/internal/plugin/installer/http_installer.go b/internal/plugin/installer/http_installer.go
index bb96314f4..7ff8fdcc7 100644
--- a/internal/plugin/installer/http_installer.go
+++ b/internal/plugin/installer/http_installer.go
@@ -90,6 +90,7 @@ func (i *HTTPInstaller) Install() error {
        if err != nil {
                return fmt.Errorf("failed to extract plugin metadata from tarball: %w", err)
        }
+       i.PluginName = metadata.Name
```

Otherwise it fails. I don't know if it should be fixed here.

idsulik:
@benoittgt it doesn't work properly because of the recent merge by @scottrigby https://github.com/helm/helm/pull/31176  

```
// internal/plugin/installer/http_installer.go:65
PluginName: stripPluginName(filepath.Base(source)),
```
it returns an empty plugin name, because the url has the plugin name without a version.

and, here it returns the error, because now it checks if the plugin exists, but because the plugin name is empty, it checks if the plugins dir exists, which is true
```
// pkg/cmd/plugin_install.go:159
verifyResult, err := installer.InstallWithOptions(i, opts)
```


We either need to change the `stripPluginName` and make the version optional or need to do the @scottrigby changes in a different way

idsulik:
@scottrigby hi! Could you please check this comment - https://github.com/helm/helm/pull/31405#issuecomment-3484012831 ?

idsulik:
It is sad when your PRs receive so little attention. You spend your time trying to contribute something useful, but in the end, it feels like that effort leads nowhere.

What is especially disappointing is how little respect there seems to be for other people’s time and work. I contributed in 2024, and those PRs were automatically closed a year later. In 2025, I decided to try again, hoping things might be different, but unfortunately, nothing changed. Once again, even approved PRs were left without any follow-up, and the bot marked them as stale again.

It would be easier to see this as a temporary issue, but in the Helm repository, it seems more like a pattern than an exception. PRs have been automatically closed after a year, before as well, simply because maintainers did not review them. One example is this: https://github.com/helm/helm/pull/13095

I probably will not be contributing here anymore, but I am still grateful for such a great tool.
@mattfarina

Issue #31402, opened 2025-10-18
nil pointer dereference on helm upgrade


jessesimpson36:
### What happened?

I was running the helm v4 beta on my companies helm chart repository against our unit tests, and I came across the following error when performing a `helm upgrade`.

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1fee912]

goroutine 16724 [running]:
helm.sh/helm/v4/pkg/action.(*Upgrade).releasingUpgrade(0xc00033d4a0, 0xc000060c40, 0xc001509480, {0xc000cbb688, 0x4d, 0x8f}, {0xc0016ee008, 0x4d, 0x8f}, 0xc000498000, ...)
	helm.sh/helm/v4/pkg/action/upgrade.go:472 +0x772
created by helm.sh/helm/v4/pkg/action.(*Upgrade).performUpgrade in goroutine 1
	helm.sh/helm/v4/pkg/action/upgrade.go:411 +0x13a8
exit status 2
```

One interesting bit that I saw was that addr2line seems to think there is an equality comparison on an action.Pull object. But I think this is misleading.

```
> go tool addr2line ~/code/helm/bin/helm           
0x1fee912
type:.eq.helm.sh/helm/v4/pkg/action.Pull
```

I'm also somewhat thrown off by the memory addresses from the error message. If there are memory addresses, does that not mean that the pointers are not nil? Are they simply memory references pointing to garbage?

What I think is the real source of this issue is the following line of code:

https://github.com/helm/helm/blob/490dffeb3458a1ad1a8e0140b33a1d1b43ce7a04/pkg/action/upgrade.go#L464-L474

`u.cfg.KubeClient.Update`  has the possibility of returning nil since it's a pointer, and when I look at the implementation of that function, I see that there are some scenarios where it does return nil.

https://github.com/helm/helm/blob/490dffeb3458a1ad1a8e0140b33a1d1b43ce7a04/pkg/kube/client.go#L691-L705

I don't know which of the codepaths my CI threw this error on though. Step 1 should probably be clearing up the nil dereference.

### What did you expect to happen?

Successful helm upgrade.

### How can we reproduce it (as minimally and precisely as possible)?

Not sure yet, but I'll fill in this later when I have something that doesn't require a lot of CI.

### Helm version

<details>

```console
$ helm version
[...]

PR #31406, opened 2025-10-18
fix: kube client should return empty results objects instead of nil

jessesimpson36:
Related to #31402 ,

This change changes returning `nil` on error pathways in kube client to returning an empty Result. We need this because downstream callers of this function expect to report back the status on Result.Created, but the result is `nil` so `.Created` panics.

Linked issue has more details

**If applicable**:


mattfarina (approved):
lgtm. In Helm v3 this function never returned `nil`.

mattfarina:
merging as this is an XS.

PR #31408, opened 2025-10-20
Improve error message when plugin source cannot be determined or a non-directory is passed

### LINKED ISSUE #31398: Helm v4: Unhelpful error `Cannot detect VCS` when trying to `helm plugin install` with invalid file path

AndiDog:

Closes #31398

**If applicable**:

### DIFF
--- a/internal/plugin/installer/installer.go
+++ b/internal/plugin/installer/installer.go
@@ -139,18 +139,24 @@ func Update(i Installer) error {
 }
 
 // NewForSource determines the correct Installer for the given source.
-func NewForSource(source, version string) (Installer, error) {
-	// Check if source is an OCI registry reference
+func NewForSource(source, version string) (installer Installer, err error) {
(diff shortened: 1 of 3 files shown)


robertsirc (approved):
LGTM

mattfarina (approved):
lgtm

PR #31411, opened 2025-10-21
fix: reinstate logger parameter to actions package

### LINKED ISSUE #31399: Helm v4 SDK no longer offers the ability to specify a logger per action
### What happened?

While testing Helm v4 SDK. One regression for us is the [inability to pass a custom logger ](https://pkg.go.dev/helm.sh/helm/v4@v4.0.0-alpha.1/pkg/action#Configuration.Init)to `action.Init` anymore. ([before](https://pkg.go.dev/helm.sh/helm/v3/pkg/action#Configuration.Init)). 
I understand the change, but **it removes the SDK’s ability to easily attach specific log contexts per action**.
At work, we enrich logs with request metadata before sending them via webhook or message queue, so each Helm action has properly scoped logs with a deployment id we provide. It works very well.
I had started [implementing](https://github.com/helm/helm/commit/b642bca8f61dfbe19e98ad75b73b2eaac2511405) something similar when migrating parts of Helm to slog, but it was later removed. Still the regression is [my fault](https://github.com/helm/helm/pull/30708#issuecomment-2793640713).
**Currently, the only way I see is through `slog.SetDefault(customLogger)`, which is global and can’t handle concurrent requests properly.**

It will be a regression for Flux too. See https://github.com/fluxcd/helm-controller/blob/de97aa9d843c2e9765de217470ba67195b5ce733/internal/action/config.go#L52

Be able to have a custom logger per action.

It is via [SDK](https://github.com/helm/helm-www/blob/d44a2b4cf281b3863077fe74f0df4900ec061b93/sdkexamples/main.go#L38) but code will look like
```diff
	if err := actionConfig.Init(
		settings.RESTClientGetter(),
		namespace,
		helmDriver,
-		logger.Printf); err != nil {
+		); err != nil {
		return nil, err
	}
```

### Helm version

<details>

```console
$ helm version
v4 beta 1
```
</details>

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

banjoh:

Add ability to override logger when creating new actions

Here is an example of how to use it

```go
package main

import (
	"log/slog"
	"os"

	"helm.sh/helm/v4/pkg/action"
	"helm.sh/helm/v4/pkg/cli"
)

func main() {
	settings := cli.New()

	mylogger := slog.New(slog.NewTextHandler(os.Stdout, nil))

	actionConfig := action.NewConfiguration()
	actionConfig.SetLogger(mylogger)              // <------ New method to set logger

	// You can pass an empty string instead of settings.Namespace() to list
	// all namespaces
	if err := actionConfig.Init(settings.RESTClientGetter(), settings.Namespace(), os.Getenv("HELM_DRIVER")); err != nil {
		mylogger.Error("failed to init action config", "error", err)
		os.Exit(1)
	}
}
```

Fixes: #31399

- As opposed to v3, the logger is set using a `SetLogger()` method instead of passing the logger in `action.Configuration.Init()` function, or setting it via `action.Configuration.Log`
- The logger is an `slog.Logger` instead of the older `log` logger library in v3

**If applicable**:

### DIFF
--- a/internal/logging/logging.go
+++ b/internal/logging/logging.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"log/slog"
 	"os"
+	"sync/atomic"
 )
 
 // DebugEnabledFunc is a function type that determines if debug logging is enabled
@@ -93,3 +94,28 @@ type LoggerSetterGetter interface {
 	// Logger returns the logger for the object
 	Logger() *slog.Logger
 }
+
+type LogHolder struct {
+	// logger is an slog.Logger pointer to use the driver
+	logger atomic.Pointer[slog.Logger]
+}
+
+// Logger returns the logger for the LogHolder. If nil, returns slog.Default().
+func (l *LogHolder) Logger() *slog.Logger {
+	if lg := l.logger.Load(); lg != nil {
+		return lg
+	}
+	return slog.Default() // We rarely get here, just being defensive
+}
+
+// SetLogger sets the logger for the LogHolder. If nil, sets the default logger.
+func (l *LogHolder) SetLogger(newLogger *slog.Logger) {
+	if newLogger == nil {
+		l.logger.Store(slog.New(slog.DiscardHandler)) // Assume nil as discarding logs
+		return
+	}
+	l.logger.Store(newLogger)
+}
+
+// Ensure LogHolder implements LoggerSetterGetter
+var _ LoggerSetterGetter = &LogHolder{}
--- a/internal/logging/logging.go
+++ b/internal/logging/logging.go
@@ -97,6 +97,7 @@ type LoggerSetterGetter interface {
 
 type LogHolder struct {
 	// logger is an atomic.Pointer[slog.Logger] to store the slog.Logger
+	// We use atomic.Pointer for thread safety
 	logger atomic.Pointer[slog.Logger]
 }
 
--- a/internal/logging/logging.go
+++ b/internal/logging/logging.go
@@ -89,14 +89,14 @@ func NewLogger(debugEnabled DebugEnabledFunc) *slog.Logger {
 
 // LoggerSetterGetter is an interface that can set and get a logger
 type LoggerSetterGetter interface {
-	// SetLogger sets the logger for the object
-	SetLogger(logger *slog.Logger)
-	// Logger returns the logger for the object
+	// SetLogger sets a new slog.Handler
+	SetLogger(newLogger slog.Handler)
+	// Logger returns the slog.Logger created from the slog.Handler
 	Logger() *slog.Logger
 }
 
 type LogHolder struct {
-	// logger is an slog.Logger pointer to use the driver
+	// logger is an atomic.Pointer[slog.Logger] to store the slog.Logger
 	logger atomic.Pointer[slog.Logger]
 }
 
@@ -109,12 +109,12 @@ func (l *LogHolder) Logger() *slog.Logger {
 }
 
 // SetLogger sets the logger for the LogHolder. If nil, sets the default logger.
-func (l *LogHolder) SetLogger(newLogger *slog.Logger) {
+func (l *LogHolder) SetLogger(newLogger slog.Handler) {
 	if newLogger == nil {
 		l.logger.Store(slog.New(slog.DiscardHandler)) // Assume nil as discarding logs
 		return
 	}
-	l.logger.Store(newLogger)
+	l.logger.Store(slog.New(newLogger))
 }
 
 // Ensure LogHolder implements LoggerSetterGetter
--- a/internal/logging/logging.go
+++ b/internal/logging/logging.go
@@ -90,7 +90,7 @@ func NewLogger(debugEnabled DebugEnabledFunc) *slog.Logger {
 // LoggerSetterGetter is an interface that can set and get a logger
 type LoggerSetterGetter interface {
 	// SetLogger sets a new slog.Handler
-	SetLogger(newLogger slog.Handler)
+	SetLogger(newHandler slog.Handler)
 	// Logger returns the slog.Logger created from the slog.Handler
 	Logger() *slog.Logger
 }
@@ -109,12 +109,12 @@ func (l *LogHolder) Logger() *slog.Logger {
 }
 
 // SetLogger sets the logger for the LogHolder. If nil, sets the default logger.
-func (l *LogHolder) SetLogger(newLogger slog.Handler) {
-	if newLogger == nil {
+func (l *LogHolder) SetLogger(newHandler slog.Handler) {
+	if newHandler == nil {
 		l.logger.Store(slog.New(slog.DiscardHandler)) // Assume nil as discarding logs
 		return
 	}
-	l.logger.Store(slog.New(newLogger))
+	l.logger.Store(slog.New(newHandler))
 }
 
 // Ensure LogHolder implements LoggerSetterGetter
--- a/internal/logging/logging.go
+++ b/internal/logging/logging.go
@@ -85,3 +85,11 @@ func NewLogger(debugEnabled DebugEnabledFunc) *slog.Logger {
 
 	return slog.New(dynamicHandler)
 }
+
+// LoggerSetterGetter is an interface that can set and get a logger
+type LoggerSetterGetter interface {
+	// SetLogger sets the logger for the object
+	SetLogger(logger *slog.Logger)
+	// Logger returns the logger for the object
+	Logger() *slog.Logger
+}
--- a/internal/logging/logging.go
+++ b/internal/logging/logging.go
@@ -106,7 +106,7 @@ func (l *LogHolder) Logger() *slog.Logger {
 	if lg := l.logger.Load(); lg != nil {
 		return lg
 	}
-	return slog.Default() // We rarely get here, just being defensive
+	return slog.New(slog.DiscardHandler) // Should never be reached
 }
 
 // SetLogger sets the logger for the LogHolder. If nil, sets the default logger.
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -535,6 +535,7 @@ func (cfg *Configuration) Init(getter genericclioptions.RESTClientGetter, namesp
 		if err != nil {
 			return fmt.Errorf("unable to instantiate SQL driver: %w", err)
 		}
+		d.SetLogger(cfg.Logger())
 		store = storage.Init(d)
 	default:
 		return fmt.Errorf("unknown driver %q", helmDriver)
@@ -543,7 +544,6 @@ func (cfg *Configuration) Init(getter genericclioptions.RESTClientGetter, namesp
 	cfg.RESTClientGetter = getter
 	cfg.KubeClient = kc
 	cfg.Releases = store
-	cfg.Releases.SetLogger(cfg.Logger())
 	cfg.HookOutputFunc = func(_, _, _ string) io.Writer { return io.Discard }
 
 	return nil
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -564,9 +564,8 @@ func (cfg *Configuration) Logger() *slog.Logger {
 
 // SetLogger sets the logger for the Configuration. If nil, sets the default logger.
 func (cfg *Configuration) SetLogger(newLogger *slog.Logger) {
-	// Only set logger if it's currently nil
 	if newLogger == nil {
-		cfg.logger.Store(slog.Default()) // We never want to set the logger to nil
+		cfg.logger.Store(slog.New(slog.DiscardHandler)) // Assume nil as discarding logs
 		return
 	}
 	cfg.logger.Store(newLogger)
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -530,7 +530,6 @@ func (cfg *Configuration) Init(getter genericclioptions.RESTClientGetter, namesp
 	case "sql":
 		d, err := driver.NewSQL(
 			os.Getenv("HELM_DRIVER_SQL_CONNECTION_STRING"),
-			cfg.Logger(),
 			namespace,
 		)
 		if err != nil {
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -29,6 +29,7 @@ import (
 	"slices"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"text/template"
 	"time"
 
@@ -109,7 +110,17 @@ type Configuration struct {
 	// HookOutputFunc called with container name and returns and expects writer that will receive the log output.
 	HookOutputFunc func(namespace, pod, container string) io.Writer
 
+	// logger is an slog.Logger pointer to use with the Configuration instance
+	logger atomic.Pointer[slog.Logger]
+
+	// Mutex is an exclusive lock for concurrent access to the action
 	mutex sync.Mutex
+}
+
+func NewConfiguration() *Configuration {
+	c := &Configuration{}
+	c.SetLogger(slog.Default())
+	return c
 }
 
 const (
@@ -376,8 +387,8 @@ func (cfg *Configuration) getCapabilities() (*common.Capabilities, error) {
 	apiVersions, err := GetVersionSet(dc)
 	if err != nil {
 		if discovery.IsGroupDiscoveryFailedError(err) {
-			slog.Warn("the kubernetes server has an orphaned API service", slog.Any("error", err))
-			slog.Warn("to fix this, kubectl delete apiservice <service-name>")
+			cfg.Logger().Warn("the kubernetes server has an orphaned API service", slog.Any("error", err))
+			cfg.Logger().Warn("to fix this, kubectl delete apiservice <service-name>")
 		} else {
 			return nil, fmt.Errorf("could not get apiVersions from Kubernetes: %w", err)
 		}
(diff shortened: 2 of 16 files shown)


pkg/action/action.go
```diff
@@ -109,9 +110,19 @@ type Configuration struct {
...
 	HookOutputFunc func(namespace, pod, container string) io.Writer
 
+	// logger is an slog.Logger pointer to use with the Configuration instance
+	logger atomic.Pointer[slog.Logger]
+
+	// Mutex is an exclusive lock for concurrent access to the action
 	mutex sync.Mutex
 }
 
+func NewConfiguration() *Configuration {
```
gjenkins8:
```suggestion
func NewConfiguration(options...ConfigurationOption) *Configuration {
```

Option parameters might be better than this `SetLogger()` pattern?

pkg/action/action.go
```diff
@@ -510,6 +524,7 @@ func (cfg *Configuration) Init(getter genericclioptions.RESTClientGetter, namesp
 		if d == nil {
 			d = driver.NewMemory()
 		}
+		d.SetLogger(cfg.Logger())
```
gjenkins8: should we only set when creating? ie.

```
if d == nil {
	d = driver.NewMemory()
	d.SetLogger(cfg.Logger())
}
```

pkg/action/action.go
```diff
@@ -538,6 +554,23 @@ func (cfg *Configuration) SetHookOutputFunc(hookOutputFunc func(_, _, _ string)
 	cfg.HookOutputFunc = hookOutputFunc
 }
 
+// Logger returns the logger for the Configuration. If nil, returns slog.Default().
+func (cfg *Configuration) Logger() *slog.Logger {
+	if lg := cfg.logger.Load(); lg != nil {
+		return lg
+	}
+	return slog.Default() // We rarely get here, just be defensive
```
gjenkins8: It seems not possible for this to be reached, given the logic in `SetLogger()`

That said, why not be consistent with the logic in `SetLogger()`? And return `slog.New(slog.DiscardHandler)` ?

pkg/action/action.go
```diff
@@ -510,6 +524,7 @@ func (cfg *Configuration) Init(getter genericclioptions.RESTClientGetter, namesp
 		if d == nil {
 			d = driver.NewMemory()
 		}
+		d.SetLogger(cfg.Logger())
 		d.SetNamespace(namespace)
 		store = storage.Init(d)
 	case "sql":
```
benoittgt, line 545: We should also set the logger for SQL?

pkg/kube/client.go
```diff
@@ -1194,3 +1198,19 @@ func (e *joinedErrors) Error() string {
 func (e *joinedErrors) Unwrap() []error {
 	return e.errs
 }
+
+// logger returns the logger for the Client. If nil, returns slog.Default().
+func (c *Client) Logger() *slog.Logger {
+	if lg := c.logger.Load(); lg != nil {
```
benoittgt: To reduce duplication what about having something like ?

```go
// internal/logging/logger.go
package logging

import "log/slog"

func GetLogger(logger *slog.Logger) *slog.Logger {
[...]
}

func SetLogger(newLogger *slog.Logger) *slog.Logger {
[...]
}
```

or maybe go even further if we accept this (also with the nil as `slog.Default()`).
```go
type Holder struct {
    logger *slog.Logger
}

func (h *Holder) Logger() *slog.Logger {
    if h.logger != nil {
        return h.logger
    }
    return slog.Default()
}

func (h *Holder) SetLogger(newLogger *slog.Logger) {
    if newLogger == nil {
        h.logger = slog.Default()
        return
    }
    h.logger = newLogger
}
```

pkg/kube/client.go
```diff
@@ -1194,3 +1198,19 @@ func (e *joinedErrors) Error() string {
...
+		return lg
+	}
+	return slog.Default() // We rarely get here, just being defensive
+}
+
+func (c *Client) SetLogger(newLogger *slog.Logger) {
+	if newLogger == nil {
+		c.logger.Store(slog.New(slog.DiscardHandler)) // Assume nil as discarding logs
```
benoittgt: Same comment as before. What about using `slog.Default()` when the logger is `nil`, and requiring users to explicitly pass something like `slog.New(slog.NewJSONHandler(io.Discard, nil))` if they want to discard logs?

pkg/action/action.go
```diff
@@ -109,9 +110,19 @@ type Configuration struct {
 	// HookOutputFunc called with container name and returns and expects writer that will receive the log output.
 	HookOutputFunc func(namespace, pod, container string) io.Writer
 
+	// logger is an slog.Logger pointer to use with the Configuration instance
+	logger atomic.Pointer[slog.Logger]
```
benoittgt: I'm curious, why are we not using a `*slog.Logger` directly here? If I use the [SDK example](https://github.com/helm/helm-www/blob/1f6c95e49102333845f7461c383a0fc98ae28d87/sdkexamples/upgrade.go#L15-L21) and adapt it with your code, I don't see the need to use and atomic pointer.

```go
func runUpgrade(....) error {
    upgradeLogger := slog.New(slog.NewJSONHandler(os.Stdout, nil)).With(
        slog.String("request_id", "abc-123"), 
    )
    actionConfig := action.NewConfiguration()
    actionConfig.SetLogger(upgradeLogger)
    // [...]
    upgradeClient := action.NewUpgrade(actionConfig) // then I have my logger per action 🤔 
```

Maybe I’m missing some scope detail or other considerations you had in mind?

benoittgt (commented):
Thanks a lot @banjoh for the work on this! I have few questions

pkg/kube/client.go
banjoh: I had that in the begin then [I changed my mind](https://github.com/helm/helm/pull/31411/commits/50e43f401719c4c5bcd1ddaf953726a460725539). When consuming an API like this, I find passing `nil` i.e the default value of a pointer is synonymous to "unset my value". The effect of "unsetting" a logger is not to log anything.

pkg/action/action.go
banjoh: Added

pkg/kube/client.go
banjoh: Good suggestion. I assume you are suggesting using `Holder` as a struct embedding, right? Like so

```go
type MyThing struct {
    Holder
}

func (m *MyThing) Do() {
    m.Logger().Info("Doing something")
}
```

pkg/action/action.go
banjoh: This is to make sure `SetLogger` and `Logger` are thread safe

benoittgt: I’m wondering in which scenario someone would actually modify the logger during execution. 🤔
If we typically set it once in Configuration (for example, at action initialization), I don’t see a case where concurrent reads and writes would happen:

```go
cfg := action.NewConfiguration()

// Goroutine 1: Constantly changing logger
go func() {
    for {
        cfg.SetLogger(logger1)  // Writing
    }
}()

// Goroutine 2: Reading logger
go func() {
    for {
        cfg.Logger().Info("test")  // Reading
    }
}()
```
Wouldn’t this kind of pattern be unusual in practice? It feels like we’d only set the logger once and reuse it across actions.

pkg/action/action.go
```diff
@@ -510,6 +524,7 @@ func (cfg *Configuration) Init(getter genericclioptions.RESTClientGetter, namesp
 		if d == nil {
 			d = driver.NewMemory()
 		}
+		d.SetLogger(cfg.Logger().Handler())
```
gjenkins8, line 542: should we only set the logger when creating? ie.

```
if d == nil {
	d = driver.NewMemory()
	d.SetLogger(cfg.Logger())
}
```

internal/logging/logging.go
```diff
@@ -85,3 +86,37 @@ func NewLogger(debugEnabled DebugEnabledFunc) *slog.Logger {
...
+	// We use atomic.Pointer for thread safety
+	logger atomic.Pointer[slog.Logger]
+}
+
+// Logger returns the logger for the LogHolder. If nil, returns slog.Default().
+func (l *LogHolder) Logger() *slog.Logger {
+	if lg := l.logger.Load(); lg != nil {
+		return lg
+	}
+	return slog.Default() // We rarely get here, just being defensive
```
gjenkins8:
```suggestion
	return slog.New(slog.DiscardHandler) // Should never be reached
```

It seems not possible for this to be reached, given the logic in `SetLogger()`

That said, why not be consistent with the logic in SetLogger()? And return `slog.New(slog.DiscardHandler)` ?

pkg/action/action.go
gjenkins8: ie. 
```
type ConfigurationOption func(c *Configuration)

// Override the default logging handler
// If unspecified, the default logger will be used
func ConfigurationSetLogger(h *slog.Handler) ConfigurationOption {
    return func(c *Configuration) {
       c.logger = slog.New(h)
    }
}
```

```
func NewConfiguration(options...ConfigurationOption) {
    c := &Configuration{}
    for _, o := options {
       o(c)
    }
    
     if o.logger == nil {
        o.logger = slog.Default()
     }

```

gjenkins8: this would assume that the user can set the logger when creating the action Configuration

(and doesn't want to change it later; but would be a somewhat "more idiomatic" or "cleaner" approach, perhaps)

pkg/action/action.go
banjoh: > Wouldn’t this kind of pattern be unusual in practice? It feels like we’d only set the logger once and reuse it across actions.

Its unusual to do this for sure, but given that the cost of implementing thread safety is low, why not ensure its there? If performance is a concern, using `atomic.Pointer` ensure we have lock-free thread safety so the cost is negligible.

pkg/action/action.go
banjoh: > It seems not possible for this to be reached, given the logic in SetLogger()

If `SetLogger()` is never called i.e "do not set a logger", the code can be reached.

```go
	actionConfig := new(Configuration)
	actionConfig.Logger().Info("I'm ready to go!")
```

pkg/action/action.go
banjoh: Done. It allows us to add more options in future

robertsirc (changes requested):
Not to be difficult, but shouldn't we put some test around this?

banjoh:
> Not to be difficult, but shouldn't we put some test around this?

We could. They should be trivial to add

banjoh:
> Not to be difficult, but shouldn't we put some test around this?

@robertsirc I've added some tests. Please take a look. Thanks

pkg/action/action.go
banjoh: We want to set the logger regardless of where the store instance is coming from. We want to avoid the code below

```go
var d *driver.Memory
if cfg.Releases != nil {
    if mem, ok := cfg.Releases.Driver.(*driver.Memory); ok {
        // This function can be called more than once (e.g., helm list --all-namespaces).
        // If a memory driver was already initialized, reuse it but set the possibly new namespace.
        // We reuse it in case some releases where already created in the existing memory driver.
        d = mem
        d.SetLogger(cfg.Logger().Handler())    <------ here
    }
}
if d == nil {
    d = driver.NewMemory()
    d.SetLogger(cfg.Logger().Handler())    <------ and here
}
```

pkg/action/action.go
```diff
@@ -109,7 +110,30 @@ type Configuration struct {
...
+		c.SetLogger(h)
+	}
+}
+
+func NewConfiguration(options ...ConfigurationOption) *Configuration {
+	c := &Configuration{}
+	for _, o := range options {
+		o(c)
+	}
+
```
gjenkins8, line 137: I think we need the `if c.logger == nil { c.SetLogger(slog.Default()) }` here?

Otherwise nothing will se tthe logger, and `Logger()` will default to the discard handler.

banjoh: Oh, yes. Good catch.

banjoh: Hmm. `logger` would never be `nil` since it's not a pointer type 

```go
type LogHolder struct {
	logger atomic.Pointer[slog.Logger]
}
```

gjenkins8: The atomic value/pointer is nil:
```
if c.logger.Load() == nil { c.SetLogger(slog.Default()) }
```

benoittgt:
Hello

TL;DR. Looks good on SDK and cli.

---

I did a full test. First in SDK scenario.

```go
// Inspired by https://github.com/helm/helm-www/blob/main/sdkexamples/upgrade.go
// main.go
package main

import (
	"context"
	"log/slog"
	"os"

	"helm.sh/helm/v4/pkg/action"
	"helm.sh/helm/v4/pkg/chart/loader"
	"helm.sh/helm/v4/pkg/cli"
	"helm.sh/helm/v4/pkg/kube"
	releasev1 "helm.sh/helm/v4/pkg/release/v1"
)

func main() {
	baseLogger := slog.New(slog.NewTextHandler(os.Stderr, &slog.HandlerOptions{
		Level: slog.LevelDebug,
	}))
	logger := baseLogger.With(
		slog.String("deployment-id", "12345"),
	)
	logger.Info("Starting Helm upgrade test with new slog integration")

	settings := cli.New()
	settings.Debug = true

	actionConfig := action.NewConfiguration()
	actionConfig.SetLogger(logger.Handler())

	if err := actionConfig.Init(
		settings.RESTClientGetter(),
		settings.Namespace(),
		os.Getenv("HELM_DRIVER"),
	); err != nil {
		logger.Error("Failed to initialize action config", "error", err)
		os.Exit(1)
	}

	upgradeClient := action.NewUpgrade(actionConfig)
	// release type will be *releasev1.Release"
	upgradeClient.WaitStrategy = kube.HookOnlyStrategy

	releaseName := "simple-test"
	chartPath := "../simple-test"

	logger.Info("Upgrading release", "release", releaseName, "chart", chartPath)

	chart, err := loader.Load(chartPath)
	if err != nil {
		logger.Error("Failed to load chart", "error", err)
		os.Exit(1)
	}

	rel, err := upgradeClient.RunWithContext(context.Background(), releaseName, chart, nil)
	if err != nil {
		logger.Error("Upgrade failed", "error", err)
		os.Exit(1)
	}

	r := rel.(*releasev1.Release)
	logger.Info("Upgrade successful", "release", r.Name, "version", r.Version, "status", r.Info.Status.String())
}
```

```go
go run main.go

[...]
time=2025-10-24T18:05:23.280+02:00 level=DEBUG msg="Patched resource" deployment-id=12345 namespace=default name=simple-test gvk="apps/v1, Kind=Deployment"
time=2025-10-24T18:05:23.297+02:00 level=DEBUG msg="updating status for upgraded release" deployment-id=12345 name=simple-test
time=2025-10-24T18:05:23.300+02:00 level=INFO msg="Upgrade successful" deployment-id=12345 release=simple-test version=21 status=deployed
```

WIth cli
```
./bin/helm upgrade simple-test simple-test --debug --wait
level=DEBUG msg="number of dependencies in the chart" dependencies=0
level=DEBUG msg="waiting for resources" count=3 timeout=5m0s
level=DEBUG msg="waiting for resource" name=simple-test kind=ServiceAccount expectedStatus=Current actualStatus=Unknown
level=DEBUG msg="waiting for resource" name=simple-test kind=Deployment expectedStatus=Current actualStatus=Unknown
Release "simple-test" has been upgraded. Happy Helming!
```

It looks good.

mattfarina (approved):
lgtm

I know there is another PR to expand on test, too.

robertsirc (approved):
LGTM

matheuscscp:
Hi, I'm gonna do a review on behalf of Flux later today

matheuscscp:
LGTM!

scottrigby (approved):
lgtm! Was also wanting to hear from Flux review - thanks @matheuscscp 👍

internal/logging/logging.go
```diff
@@ -85,3 +86,37 @@ func NewLogger(debugEnabled DebugEnabledFunc) *slog.Logger {
...
+
+type LogHolder struct {
+	// logger is an atomic.Pointer[slog.Logger] to store the slog.Logger
+	// We use atomic.Pointer for thread safety
+	logger atomic.Pointer[slog.Logger]
+}
+
+// Logger returns the logger for the LogHolder. If nil, returns slog.Default().
+func (l *LogHolder) Logger() *slog.Logger {
+	if lg := l.logger.Load(); lg != nil {
```
TerryHowe, line 106: It would be nice if this would default to `slog.Default().Handler()` so this doesn't have to call SetLogger all over the place. There could be a follow up to do this if anyone else thinks this is a good idea.

TerryHowe (approved):
/lgtm

Thanks for working on this, nice work!

internal/logging/logging.go
scottrigby: this does sound useful as a follow-up IMO

internal/logging/logging.go
```diff
@@ -85,3 +86,37 @@ func NewLogger(debugEnabled DebugEnabledFunc) *slog.Logger {
 
 	return slog.New(dynamicHandler)
 }
+
+// LoggerSetterGetter is an interface that can set and get a logger
+type LoggerSetterGetter interface {
```
matheuscscp, line 91: @banjoh There also used to be a method for passing a logger to the `pkg/storage.Storage`. Now the way you can do it is by passing a `driver.Driver` to `pkg/storage.Init()` that implements this `LoggerSetterGetter` interface. Can this interface be exposed, please? :pray:

pkg/storage/storage.go
```diff
@@ -331,7 +335,16 @@ func Init(d driver.Driver) *Storage {
...
 		Driver: d,
 	}
+
+	// Get logger from driver if it implements the LoggerSetterGetter interface
+	if ls, ok := d.(logging.LoggerSetterGetter); ok {
+		ls.SetLogger(s.Logger().Handler())
+	} else {
+		// If the driver does not implement the LoggerSetterGetter interface, set the default logger
+		s.SetLogger(slog.Default().Handler())
+	}
```
matheuscscp, line 348: Sorry, a bit too late, but there's a bug here. Line 344 should be `s.SetLogger()`

PR #31413, opened 2025-10-22
pkg/cmd: clarify install options, document them

stevekuznetsov:
Without a careful reading of the installation code, I made a number of mistakes configuring the install client with the `IncludeCRDs` and `SkipCRDs` options. The new names proposed in this commit aim to clarify the intent of these options and make such mistakes impossible in the future.

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -106,14 +106,23 @@ type Install struct {
 	Description      string
 	OutputDir        string
 	// RollbackOnFailure enables rolling back (uninstalling) the release on failure if set
-	RollbackOnFailure        bool
+	RollbackOnFailure bool
+	// SkipCRDs opts the chart installation out of installing or upgrading CRDs.
+	//  - If set to false, CRDs are installed (or upgraded) during a chart installation and
+	//    AlreadyExists errors are ignored.
+	//  - If set to true, no operations on CRDs are performed on the cluster during installation.
 	SkipCRDs                 bool
 	SubNotes                 bool
 	HideNotes                bool
 	SkipSchemaValidation     bool
 	DisableOpenAPIValidation bool
-	IncludeCRDs              bool
-	Labels                   map[string]string
+	// IncludeCRDs configures template rendering to include rendered CRD manifests in the overall list
+	// of manifests for the chart. This is useful for operations over helm templates, but should be
+	// used with great care during installs against live clusters - including the CRD content in the
+	// release's manifests during install will cause the CRDs to be deleted during upgrade, as future
+	// releases will not include CRDs in release manifest lists.
+	IncludeCRDs bool
+	Labels      map[string]string
 	// KubeVersion allows specifying a custom kubernetes version to use and
 	// APIVersions allows a manual set of supported API Versions to be passed
 	// (for things like templating). These are ignored if ClientOnly is false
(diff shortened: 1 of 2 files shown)


TerryHowe (changes requested):
While these names may be better, this is a breaking change and would have to wait until v5.  Do you want to make this change or update the comments?

stevekuznetsov:
@TerryHowe does your versioning scheme apply to the library as well as the flags for users? Also, isn't the point of alpha releases that you can still break APIs? :sweat_smile:

stevekuznetsov:
My preference would be to have a broader refactor, if I'm honest - the flow that I triggered by setting `installClient.IncludeCRDs` was terrifying - the next `helm upgrade` silently _deleted the CRDs_ from the cluster, causing data loss. While I can appreciate not making a breaking change, in my personal opinion the cost of this behavior being the way it is likely outweighs the (likely few) users of the V4 Helm library code that would be affected. Would be good to hear your thoughts on this.

TerryHowe:
The beta is out for v4 and there aren't supposed to be any breaking changes with the CLI or SDK at this point.

stevekuznetsov:
@mattfarina hi! I found a set of flags on the install SDK which cause Helm v4 to silently delete CRDs and cause catastrophic data loss on clusters. Do you think this warrants some breaking changes in the SDK (but not the CLI) to make it less likely that users hit this?

@TerryHowe I appreciate what you are saying but I don't think you are commenting on the important part of my question - should Helm silently delete CRDs on a cluster? Data loss on upgrade seems like the cardinal sin here and making that hard, if not impossible, seems like a good goal. (And a breaking change in the SDK, even if in a beta, seems like a much, much, much smaller issue.)

TerryHowe:
Documentation would help the poorly named booleans and would not break SDK users. Maybe put a TODO in there to update for v5. That would be my guess on where this is going to have to go.

stevekuznetsov:
Sure, that's the least we can do.

pkg/action/install.go
```diff
@@ -106,14 +106,23 @@ type Install struct {
 	Description      string
 	OutputDir        string
 	// RollbackOnFailure enables rolling back (uninstalling) the release on failure if set
-	RollbackOnFailure        bool
+	RollbackOnFailure bool
+	// SkipCRDs opts the chart installation out of installing or upgrading CRs on the cluster. By
+	// default, CRDs are installed (or upgraded) during a chart installation and AlreadyExists errors
+	// are ignored. When this flag is set, no operations over CRDs will be made against the cluster
+	// during installation.
```
TerryHowe:
```suggestion
	// SkipCRDs If set to false, CRDs are installed (or upgraded) during a chart installation and
	// AlreadyExists errors are ignored. If set to true, no operations on CRDs are performed on
	// the cluster during installation.
```

pkg/action/install.go
```diff
@@ -106,14 +106,23 @@ type Install struct {
...
 	SkipCRDs                 bool
```
gjenkins8, line 114: Can we add the same verbiage to Upgrade's `SkipCRDs` for consistency please

https://github.com/helm/helm/blob/main/pkg/action/upgrade.go#L70

pkg/action/install.go
stevekuznetsov: >SkipCRDs If set to false

Not really how GoDoc is normally written - adapted your comment to make it flow a little nicer.

PR #31425, opened 2025-10-27
chore: tests lazy client

StephanieHhnbrg:
Hi, this is a follow up PR of https://github.com/helm/helm/pull/31239 related to the test coverage issue of the action package https://github.com/helm/helm/issues/30814 

Thank you 

**If applicable**:

### DIFF
--- a/pkg/action/lazyclient_test.go
+++ b/pkg/action/lazyclient_test.go
@@ -38,6 +38,8 @@ import (
 	"helm.sh/helm/v4/pkg/kube"
 )
 
+const fakeErrorMessage = "fake helm error"
+
 func TestLazyClient_Init(t *testing.T) {
 	kc := kube.New(nil)
 	lazyClient := &lazyClient{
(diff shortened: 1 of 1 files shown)


pkg/action/lazyclient_test.go
```diff
@@ -0,0 +1,301 @@
...
+	assert.NoError(t, lazyClient.init())
+}
+
+func TestLazyClient_Init_Error(t *testing.T) {
+	lazyClient := &lazyClient{
+		namespace: namespace,
+		clientFn:  getErrorClient,
+	}
+
+	assert.Errorf(t, lazyClient.init(), "fake error")
```
mrlunchbox777: To preface i'm not suggesting we should test upstream, but rather that we should ensure that we don't read upstream errors as our own. I think where `assert.Errorf` is used in this file, it should probably be `assert.Error` or `assert.EqualError` with the required associated changes in addition to changing `"fake error"` to something more specific like `"fake helm error"`.

StephanieHhnbrg: Good point! I agree. We should just test that the error handling / forwarding is in place.

I have changed the assertions to assert.EqualError and updated the error message. Thanks for the review!

mrlunchbox777 (approved):
lgtm

PR #31427, opened 2025-10-28
Check plugin name is not used

tisonkun:

Resolve TODOs in code.

**If applicable**:


pkg/cmd/load_plugins.go
```diff
@@ -132,7 +132,13 @@ func loadCLIPlugins(baseCmd *cobra.Command, out io.Writer) {
 			DisableFlagParsing: true,
 		}
 
-		// TODO: Make sure a command with this name does not already exist.
+		for _, cmd := range baseCmd.Commands() {
+			if cmd.Name() == c.Name() {
+				fmt.Fprintf(os.Stderr, "failed to load plugins: name conflicts %s\n", c.Name())
```
TerryHowe: This should use `slog`

TerryHowe (changes requested):

pkg/cmd/load_plugins.go
tisonkun: This follows the pattern as in this file L66. I don't see `slog` used in this file. Or otherwise we'd refactor all `fmt.Fprintf` with `slog` in this codepath?

robertsirc: So, I agree with @TerryHowe that this should be `slog` but in the scope of this PR, it could remain the same as the rest of the file and let's have another PR to clean this up. YES, I agree 100% this is a little extra work but if we need to roll back it makes it a bit easier to handle.

robertsirc:
@tisonkun the build has some issues could you take a look.

tisonkun:
@robertsirc test failed because out current testdata have a plugin that conflicts with existing command name `env`.

Shall we rename the plugin `pkg/cmd/testdata/helmhome/helm/plugins/env` to something like `shortenv`?

tisonkun:
I'll change the test plugin name to `shortenv` later.

tisonkun:
cc @robertsirc may you trigger CI workflow again?

pkg/cmd/load_plugins.go
TerryHowe: https://github.com/helm/helm/issues/31454

TerryHowe (dismissed):
/lgtm

tisonkun:
@gjenkins8 may you take a look at this PR? It seems another candidate for long pending and finally stale :/

pkg/cmd/load_plugins.go
gjenkins8: #31454 is merged, can we update this PR to use `slog` please

tisonkun: Let me do the update.

I'm experiencing two PRs hanging on for months and I typically wait for an active reviewer commit to help in merging for resolving conflicts. Or else I'd do rebase again and again. Maybe it's not what helm looks like, but it's quite common :\

tisonkun: Updated.

tisonkun:
Any chance we move this PR forward?

tisonkun:
cc @banjoh

Sorry to mention more committers but this seems to be stalled without certain new requests.

banjoh (approved):
LGTM

PR #31431, opened 2025-10-28
Add mustToTOML template functions and let toTOML return empty string on error

### LINKED ISSUE #31430: Unalignment between toToml and toYaml/toJson
@technosophos why do we return `err.Error()` rather than `""`? This creates a confusing unalignment from `toYaml` and `toJson`.

_Originally posted by @tisonkun in https://github.com/helm/helm/pull/2354#discussion_r2468978448_

I'd suggest we revert `toToml`'s behavior. Not sure if it's considered as a breaking change or somehow a "bug fix".

With the addition of a `mustToToml` function (see https://github.com/helm/helm/pull/30553#issuecomment-3455769314), this revert should be fine. I can't imagine a user who want to render the error output.

tisonkun:

This closes https://github.com/helm/helm/issues/31430. See the issue for background. The changeset itself should be straightforward.

**If applicable**:


TerryHowe (changes requested):
Can you add some tests?

tisonkun:
> Can you add some tests?

Sure. Perhaps following https://github.com/helm/helm/pull/30553/. It's midnight for me now. I may handle it later.

TerryHowe:
Sounds good. Thanks for creating the pull request

pkg/engine/funcs_test.go
```diff
@@ -136,29 +135,35 @@ keyInElement1 = "valueInElement1"`,
 		assert.Equal(t, tt.expect, b.String(), tt.tpl)
 	}
 
-	loopMap := map[string]interface{}{
-		"foo": "bar",
+	nonSerializable := map[string]interface{}{
+		"foo": struct{ Fn func() }{},
```
tisonkun: Updated. I change the error case to an explicit non serializable type, since the toml lib would hang on cyclic reference.

pkg/engine/funcs_test.go
```diff
@@ -25,7 +25,6 @@ import (
 )
 
 func TestFuncs(t *testing.T) {
-	//TODO write tests for failure cases
```
tisonkun, line 28: The new cases contains failure ones.

benoittgt (approved):
LGTM Thanks. Also we should not forget doc update https://github.com/helm/helm-www/blob/d684820e2e902ce2361f5fbb6dd182fa4a75d967/content/en/docs/chart_template_guide/function_list.md?plain=1#L705

tisonkun:
@benoittgt Thanks for the cross ref. Will do once this patch gets merged. Not sure if you know some active helm maintainer that can help here.

TerryHowe (dismissed):
/lgtm

Tests pass for me

robertsirc (dismissed):
LGTM

tisonkun:
@robertsirc may we follow up for merging this patch?

tisonkun:
@scottrigby @robertsirc I think we have reached consensus on this patch. Hopefully we won't leave it long pending and even conflict later.

gjenkins8 (commented):
I do think we should make this consistent with other function e.g. `toYAML`/`mustToYAML`. So agree with this PR mostly.

The one consideration, is that now `toTOML` does not return a string representing the error - a behavior that a chart might be relying on (however unlikely, there are a lot of charts out there)

The one argument might be that because there was never a strong contract on the format of the returned string upon an error, we can modify this behavior. I'll discuss and revert.

Another option for this PR would be to simply introduce `mustToTOML`, and avoid modifying the behavior of `toTOML`. Which would result in the unfortunate situation where we have a duplicated function that awkwardly doesn't follow the same pattern as others. But this can be fixed/changed in a separate PR.

tisonkun:
@TerryHowe @robertsirc Shall we merge this patch now?

banjoh:
> @TerryHowe @robertsirc Shall we merge this patch now?

This has been added as a discussion item for next week's dev call as per @gjenkins8's comments.

Meeting notes: https://docs.google.com/document/d/1d-6xJEx0C78csIYSPKJzRPeWaHG_8W1Hjl72OJggwdc/edit?tab=t.0

banjoh (changes requested):
Block merging until after next weeks discussion

I'll also note that not returning an error is how its done in [sprig](https://github.com/Masterminds/sprig?tab=readme-ov-file#principles-driving-our-function-selection), the library implementing most template functions

tisonkun:
> Block merging until after next weeks discussion
> 
> I'll also note that not returning an error is how its done in [sprig](https://github.com/Masterminds/sprig?tab=readme-ov-file#principles-driving-our-function-selection), the library implementing most template functions

Thanks! The meeting happens at about 1AM in my timezone so I'll wait for the conclusion :D

tisonkun:
@TerryHowe  @banjoh any resolution on this PR?

TerryHowe:
@banjoh do you remember how this discussion went? We have any conclusion here? There are conflicts which doesn't help.

tisonkun:
I can do the rebase, but I'd like to see a signal of mergeable first.

tisonkun:
Conflict resolved.

I saw the community had wasted two more PRs for the same purpose:

* https://github.com/helm/helm/pull/31915
* https://github.com/helm/helm/pull/31942

Can we move forward with this one to avoid further conflicts (both code and collaborations)?

Request for review @robertsirc @TerryHowe @banjoh 

in cc @JiwaniZakir @realgagangupta

tisonkun:
@TerryHowe Is a manual commits squashing required? Or you guys use GitHub's "Squash and merge" to do it simply. I can forsee a squash and force pushing invalidate all reviews later if it's required 🤣

gjenkins8 (changes requested):
Thanks for the PR. The contention item here is changing the behavior of `toTOML` (which is unfortunate that it was written this way, but has been for a very long time), as Chart's may rely on this behavior while rendering.

I think we should split this PR into two. One to add `mustToTOML` (an entirely additive change). The other to change the behavior of `toTOML` (where the merits of a potentially incompatible change can be debated).

It will (continue to be) unfortunate that `toTOML` and `mustToTOML` will not follow the same behavior as e.g. `toYAML`/`mustToYAML` (until the second PR is merged). But this unblocks having `mustToTOML`, and gives a migration path for charts who might rely on `toTOML`'s current behavior) without a breaking change.

tisonkun:
Seems https://github.com/helm/helm/pull/31957/ resolves the issue.

Not sure what I can do here or just irrelevant further.

tisonkun:
> Thanks for the PR. The contention item here is changing the behavior of `toTOML` (which is unfortunate that it was written this way, but has been for a very long time), as Chart's may rely on this behavior while rendering.
> 
> I think we should split this PR into two. One to add `mustToTOML` (an entirely additive change). The other to change the behavior of `toTOML` (where the merits of a potentially incompatible change can be debated).
> 
> It will (continue to be) unfortunate that `toTOML` and `mustToTOML` will not follow the same behavior as e.g. `toYAML`/`mustToYAML` (until the second PR is merged). But this unblocks having `mustToTOML`, and gives a migration path for charts who might rely on `toTOML`'s current behavior) without a breaking change.

Weird I don't get any notifaction of this review comment.

TerryHowe:
Thanks just the same @tisonkun

PR #31441, opened 2025-10-29
Remove klog flags

gjenkins8:
Removes (hidden) `klog` flags, which are unneeded now we have switch to `slog`

**If applicable**:


mattfarina (changes requested):
The kubernetes packages use klog and those flags affect it.

Try running the following 3 commands and look at the difference...

```
helm --debug ls
```

```
helm --debug -v 5 ls
```

```
helm --debug -v 10 ls
```

For certain situations, I think we still want those flags to work.

PR #31514, opened 2025-11-15
Refactor environment variable expansion in PrepareCommands and update tests

yxxhero:
fix helm plugin cmd args env issue

**If applicable**:

### DIFF
--- a/internal/plugin/subprocess_commands_test.go
+++ b/internal/plugin/subprocess_commands_test.go
@@ -16,7 +16,6 @@ limitations under the License.
 package plugin
 
 import (
-	"os"
 	"reflect"
 	"runtime"
 	"testing"
@@ -225,7 +224,6 @@ func TestPrepareCommandsNoCommands(t *testing.T) {
 }
 
 func TestPrepareCommandsExpand(t *testing.T) {
-	t.Setenv("TESTX", "testx")
 	cmdMain := "sh"
 	cmdArgs := []string{"-c", "echo \"${TESTX}${TESTY}\""}
 	cmds := []PlatformCommand{
@@ -234,8 +232,10 @@ func TestPrepareCommandsExpand(t *testing.T) {
 
 	expectedArgs := []string{"-c", "echo \"testxtesty\""}
 
-	env := parseEnv(os.Environ())
-	env["TESTY"] = "testy"
+	env := map[string]string{
+		"TESTX": "testx",
+		"TESTY": "testy",
+	}
 
 	cmd, args, err := PrepareCommands(cmds, true, []string{}, env)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


yxxhero:
@gjenkins8

yxxhero:
@scottrigby could you take an review? thanks so much.

internal/plugin/subprocess_commands_test.go
```diff
@@ -224,16 +225,18 @@ func TestPrepareCommandsNoCommands(t *testing.T) {
 }
 
 func TestPrepareCommandsExpand(t *testing.T) {
-	t.Setenv("TEST", "test")
+	t.Setenv("TESTX", "testx")
```
gjenkins8: minor: Can we just move this logic to be included below?:

```
env["TESTX"] = "testx"
```

(as the usage of `parseEnv` is just causing this to round-trip through the process env vars)

internal/plugin/subprocess_commands_test.go
```diff
@@ -253,8 +256,8 @@ func TestPrepareCommandsNoExpand(t *testing.T) {
 	cmds := []PlatformCommand{
 		{OperatingSystem: "", Architecture: "", Command: cmdMain, Args: cmdArgs},
 	}
+	env := parseEnv(os.Environ())
```
gjenkins8: similar to above:

```suggestion
	env := map[string]string {
	    "TEST": "test",
	}
```

gjenkins8: (and remove the usage of `t.Setenv("TEST", "test")`)

internal/plugin/subprocess_commands_test.go
yxxhero: got it.

sabre1041 (approved):
LGTM

robertsirc (approved):
LGTM

robertsirc:
This is LGTM but @gjenkins8 had some comments giving him a chance to check on this before merging out of respect.

R

scottrigby (changes requested):
Nice work fixing the env map variable expansion! However, I'd suggest keeping test coverage for both OS-level and env map variables rather than replacing one with the other.

The current test changes remove `t.Setenv()` calls entirely in favor of only testing the env map. Ideally, `TestPrepareCommandsExpand()` should verify that expansion works correctly for:

1. OS-level environment variables (via `t.Setenv`)
2. Variables passed in the `env` parameter
3. Precedence when the same variable exists in both

I'm thinking we should keep one `t.Setenv()` call and add env map variables. Eg:

```go
func TestPrepareCommandsExpand(t *testing.T) {
    t.Setenv("OS_VAR", "os_value")  // Test OS-level
    
    env := map[string]string{
        "MAP_VAR": "map_value",     // Test env map
        "OS_VAR":  "override_value", // Test precedence (may also be good to test here)
    }
    
    cmdArgs := []string{"-c", "echo \"${OS_VAR} ${MAP_VAR}\""}
    // ... rest of test
}
```

This would ensure the expansion feature is thoroughly tested for scenarios where plugins encounter variables from both sources.

Basically, removing `t.Setenv` reduces coverage, we should add it back in addition to what you've added here. Checking for precedence is a nice to have, but if you don't mind ensuring this test covers that too in this same PR I think our bases will be covered.

`TestPrepareCommandsNoExpand()` should also ensure envs don't expand from either of the two sources (wouldn't need to test precedence).

scottrigby (approved):
@gjenkins8 you're right. @yxxhero never mind, my suggestion for coverage should be added elsewhere - eg, at the CLI level (ultimately testing the functionality of `plugin.SubprocessPluginRuntime.runCLI()` not in `TestPrepareCommandsExpand()`). Your test updates here are good for this function as-is :+1:

PR #31515, opened 2025-11-15
fix: Copy adopted resource info

### LINKED ISSUE #31510: Helm 4 --take-ownership doesn't work as expected with SSA
### What happened?

When running a helm install with `--take-ownership` that creates a pod with the same name as an existing pod I expected Helm to takeover the pod. Instead I receive the following error
`Error: INSTALLATION FAILED: metadata.managedFields must be nil`

Another thing that is unexpected here is that if you uninstall the failed chart, it destroys the existing pod that it fails to takeover.

Note: this only happens with SSA

When running a helm install that creates a pod with the same name as an existing pod I expect Helm to take the pod over and for the install to succeed. 

1. kubectl create namespace podinfo
2. kubectl apply -f deployment.yaml (below file)
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: podinfo
  namespace: podinfo
spec:
  replicas: 1
  strategy:
    type: RollingUpdate
    rollingUpdate:
      maxUnavailable: 1
  selector:
    matchLabels:
      app.kubernetes.io/name: podinfo
  template:
    metadata:
      labels:
        app.kubernetes.io/name: podinfo
    spec:
      containers:
        - name: podinfo
          image: "ghcr.io/stefanprodan/podinfo:6.4.0"
          command:
            - ./podinfo
            - --port=9898
```
3. helm pull oci://ghcr.io/stefanprodan/charts/podinfo --version=6.4.0
4. helm install podinfo podinfo-6.4.0.tgz -n podinfo --take-ownership

### Helm version

<details>

```console
$ helm version
version.BuildInfo{Version:"v4.0.0", GitCommit:"99cd1964357c793351be481d55abbe21c6b2f4ec", GitTreeState:"clean", GoVersion:"go1.25.4", KubeClientVersion:"v1.34"}
```
</details>

### Kubernetes version

<details>

```console
$ kubectl version
Client Version: v1.34.1
Kustomize Version: v5.7.1
Server Version: v1.31.5+k3s1
Warning: version difference between client (1.34) and server (1.31) exceeds the supported minor version skew of +/-1
```

</details>

gjenkins8:
`requireAdoption` (`--take-ownership`) or `existingResourceConflict` find resources that need updated by Helm. This PR adjusts these functions to return copies of these objects, rather than the pointer aliases returned today. So later operations on the original object don't affect the original.

fixes: https://github.com/helm/helm/issues/31510, https://github.com/helm/helm/pull/31384

**If applicable**:

### DIFF
--- a/pkg/action/validate.go
+++ b/pkg/action/validate.go
@@ -55,7 +55,8 @@ func requireAdoption(resources kube.ResourceList) (kube.ResourceList, error) {
 			return fmt.Errorf("could not get information about the resource %s: %w", resourceString(info), err)
 		}
 
-		requireUpdate.Append(info)
+		infoCopy := *info
+		requireUpdate.Append(&infoCopy)
 		return nil
 	})
 
@@ -84,7 +85,8 @@ func existingResourceConflict(resources kube.ResourceList, releaseName, releaseN
 			return fmt.Errorf("%s exists and cannot be imported into the current release: %s", resourceString(info), err)
 		}
 
-		requireUpdate.Append(info)
+		infoCopy := *info
+		requireUpdate.Append(&infoCopy)
 		return nil
 	})
 
--- a/pkg/action/validate_test.go
+++ b/pkg/action/validate_test.go
@@ -132,6 +132,7 @@ func TestRequireAdoption(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Len(t, found, 1)
 	assert.Equal(t, found[0], existing)
+	assert.NotSame(t, found[0], existing)
 }
 
 func TestExistingResourceConflict(t *testing.T) {
@@ -156,6 +157,7 @@ func TestExistingResourceConflict(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Len(t, found, 1)
 	assert.Equal(t, found[0], existing)
+	assert.NotSame(t, found[0], existing)
 
 	// Verify that an existing resource that lacks labels/annotations results in an error
 	resources = append(resources, conflict)


benoittgt:
LGTM

TerryHowe (approved):
/lgtm

pkg/action/validate.go
```diff
@@ -55,7 +55,8 @@ func requireAdoption(resources kube.ResourceList) (kube.ResourceList, error) {
 			return fmt.Errorf("could not get information about the resource %s: %w", resourceString(info), err)
 		}
 
-		requireUpdate.Append(info)
+		infoCopy := *info
```
mattfarina, line 58: Does this need to be a deep copy?

gjenkins8: for reference: https://pkg.go.dev/k8s.io/cli-runtime/pkg/resource#Info

The only field which *might* need deep coping is `Object`, as this is a pointer (Interface) to the underlying API object. So mutating this object would mutate on both copies. Other non-value fields (`Client`, `Mapping`) are find to be "shared".

I didn't deep copy, as we don't appear to need to do that -- we don't mutate the object in place. Just reload and replace. Not deep copying might be a potential footgun for someone later however. I can do so if we think this is desirable?

gjenkins8: (I should mention, I want to keep this PR minimal as it needs to be a Helm 4.0.1 bugfix. Additional hardening would/should be a new PR)

mattfarina (approved):
lgtm

PR #31517, opened 2025-11-16
fix: include version test for --dependency-update

### LINKED ISSUE #31455: helm template --dependency-update does not update dependencies
### What happened?

Create a Chart.yaml with a dependency:

```yaml
apiVersion: v2
name: mychart
description: A helm chart for deploying mychart
type: application
version: 1.0.0
dependencies:
  - name: mychart
    version: 0.0.6
    repository: oci://my-company.jfrog.io/helm
```

Run `helm template --dependency-update foo ./`

Update Chart.yaml so the dependency version is `0.0.7`

Run `helm template --dependency-update foo ./`

Observe that the dependency is not updated, despite the --dependency-update flag being passed

When using a flag called `--dependency-update`, I would expect the dependencies to be updated.

The current behavior seems to be "install dependencies if they are missing, ignore if any version of the dependency is present in the charts dir".

It seems that `helm dependency update` works as expected and does indeed update the dependencies. Another option is to delete the charts directory before running the helm template --dependency-update command

Described above

### Helm version

<details>

```console
$ helm version
version.BuildInfo{Version:"v3.19.0", GitCommit:"3d8990f0836691f0229297773f3524598f46bda6", GitTreeState:"clean", GoVersion:"go1.24.7"}
```
</details>

### Kubernetes version

Not applicable, this is during a local-only `helm template` command

Hathoute:
Closes #31455 

When using `--dependency-update`, helm does not update dependencies if an outdated version exists in `charts/`.
This PR includes a version check when attempting to match `Chart.yaml` deps to the ones in `charts/`

- Since this is my first PR in helm, I am not familiar with the testing env. I've modified internal test func to allow templating...
- Upon validating modifications for `install`, I'll add tests for `update` and `template`.

**If applicable**:

### DIFF
--- a/internal/test/test.go
+++ b/internal/test/test.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"text/template"
 )
 
 // UpdateGolden writes out the golden files with the latest values, rather than failing the test.
@@ -40,10 +41,10 @@ type HelperT interface {
 }
 
 // AssertGoldenString asserts that the given string matches the contents of the given file.
-func AssertGoldenString(t TestingT, actual, filename string) {
+func AssertGoldenString(t TestingT, actual, filename string, goldenArgs map[string]string) {
 	t.Helper()
 
-	if err := compare([]byte(actual), path(filename)); err != nil {
+	if err := compare([]byte(actual), path(filename), goldenArgs); err != nil {
 		t.Fatalf("%v\n", err)
 	}
 }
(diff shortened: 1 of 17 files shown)


robertsirc:
Hi @Hathoute could you please fix your DCO? Once that is good I will kick off the builds. 

R

Hathoute:
@robertsirc I've fixed the issue with the build. I've got some test failures but not sure whether it's because of me using windows.

baprx:
Hi, could this MR be reopened please?

TerryHowe:
I'm not sure the tests were passing. Do you want to take these commits and fixt them in a new PR?

PR #31547, opened 2025-11-20
Feature/add git based dependencies

mms-gianni:
### What it does

This enhancement adds support for using Git repositories as chart dependency sources. Charts can be fetched directly from Git repositories during `helm dep build` and `helm dep update` operations.

### Why / What the benefits are

**Zero-Configuration Local Development**
Unlike `file://...` which requires rigid directory structures and relative paths that break when moved, Git dependencies work consistently regardless of where developers clone repositories. New team members run helm dep update and get exactly the right dependencies without manual file copying, symlink creation, or environment-specific path configuration.

**Automatic Dependency Resolution**
Git dependencies automatically resolve nested dependencies using `helm package --dependency-update`. When a chart fetched from Git has its own dependencies (from Helm repos, OCI registries, or other Git repos), they're resolved transparently. This creates a unified dependency graph impossible to achieve with  dependencies which require manual management.

**Reproducible Builds with Commit SHAs**
Pin dependencies to exact commit SHAs for absolute build reproducibility. Unlike semantic versioning which can have tag moves or version overwrites, a Git commit SHA is immutable and cryptographically verifiable. This is critical for regulated industries requiring audit trails and exact artifact reproduction.

**Reduced Infrastructure Overhead**
Eliminates the need to maintain and secure separate Helm chart repositories (ChartMuseum, Harbor, Artifactory, etc.) for internal charts. Teams can leverage existing Git infrastructure (GitHub, GitLab, Bitbucket) they already know, reducing operational complexity, licensing costs, and the security surface area. (We are running ChartMuseum for a single chart)

### How it works

1. Clones the Git repository
2. Checks out the specified ref (branch, tag, or commit)
3. Automatically fetches any dependencies the chart itself requires
4. Packages the chart as a `.tgz` file 
5. Caches the packaged chart for future use

### Example:
```yaml
dependencies:
  - name: mychart
    version: "v1.2.3"  # ← This is the Git ref (tag/branch/commit or "*")
    repository: git+https://github.com/user/repo?path=charts/mychart
```

### Supported URL Formats:
- `git://`
- `git+https://`
- `git+http://`
- `git+ssh://`

### TODO
This PR might need some documentation.  

---------

**If applicable**:

### DIFF
--- a/pkg/getter/gitgetter.go
+++ b/pkg/getter/gitgetter.go
@@ -1,0 +1,277 @@
+/*
+Copyright The Helm Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package getter
+
+import (
+	"bytes"
+	"crypto/sha256"
+	"encoding/hex"
+	"fmt"
+	"log/slog"
+	"net/url"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+)
+
+// GitGetter handles fetching charts from Git repositories
+type GitGetter struct {
+	opts getterOptions
+}
+
+// Get performs a Get from a Git repository and returns the body.
+func (g *GitGetter) Get(href string, options ...Option) (*bytes.Buffer, error) {
+	for _, opt := range options {
+		opt(&g.opts)
+	}
+	return g.get(href)
+}
+
+// get clones a Git repository, packages the chart, and returns it as a buffer
+func (g *GitGetter) get(href string) (*bytes.Buffer, error) {
+	// Parse the Git URL
+	// Format: git://github.com/user/repo@ref?path=charts/mychart
+	// Or: git+https://github.com/user/repo@ref?path=charts/mychart
+
+	repoURL, ref, chartPath, err := parseGitURL(href)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse git URL: %w", err)
+	}
+
+	// Use version from options if provided (takes precedence over URL ref)
+	// This allows the dependency version field to specify the Git ref
+	if g.opts.version != "" && g.opts.version != "*" {
+		ref = g.opts.version
+	}
+
+	// Create a temporary directory for cloning
+	tmpDir, err := os.MkdirTemp("", "helm-git-")
+	if err != nil {
+		return nil, fmt.Errorf("failed to create temp directory: %w", err)
+	}
+	defer os.RemoveAll(tmpDir)
+
+	slog.Debug("cloning git repository", "url", repoURL, "ref", ref, "path", chartPath)
+
+	// Clone the repository
+	if err := g.cloneRepo(repoURL, ref, tmpDir); err != nil {
+		return nil, fmt.Errorf("failed to clone repository: %w", err)
+	}
+
+	// Determine the chart directory
+	chartDir := tmpDir
+	if chartPath != "" {
+		chartDir = filepath.Join(tmpDir, chartPath)
+	}
+
+	// Check if the chart directory exists
+	if _, err := os.Stat(chartDir); os.IsNotExist(err) {
+		return nil, fmt.Errorf("chart path %q does not exist in repository", chartPath)
+	}
+
+	// Package the chart into a tarball
+	tarData, err := g.packageChart(chartDir)
+	if err != nil {
+		return nil, fmt.Errorf("failed to package chart: %w", err)
+	}
+
+	return tarData, nil
+}
+
+// cloneRepo clones a Git repository to the specified directory
+func (g *GitGetter) cloneRepo(repoURL, ref, destDir string) error {
+	// Use shallow clone for better performance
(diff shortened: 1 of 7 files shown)


pkg/getter/gitgetter.go
```diff
@@ -0,0 +1,277 @@
...
+	args := []string{"clone", "--depth", "1"}
+
+	// If a specific ref is provided, clone that branch/tag
+	if ref != "" && ref != "HEAD" && ref != "master" && ref != "main" {
+		args = append(args, "--branch", ref)
+	}
+
+	args = append(args, repoURL, destDir)
+
+	cmd := exec.Command("git", args...)
```
banjoh: Cloning can take a long time due to network problems. Getters have `WithTimeout` option to allow stoping such operations on timeout. I'd use `exec.CommandContext` function to allow passing `context.WithTimeout(ctx, timeout)`

pkg/getter/gitgetter.go
```diff
@@ -0,0 +1,277 @@
...
+		if ref != "" && ref != "HEAD" && ref != "master" && ref != "main" {
+			slog.Debug("shallow clone failed, trying full clone", "error", stderr.String())
+			return g.fullCloneAndCheckout(repoURL, ref, destDir)
+		}
+		return fmt.Errorf("git clone failed: %s", stderr.String())
+	}
+
+	// If ref is specified but wasn't used in clone (HEAD, master, main), checkout now
+	if ref != "" && (ref == "HEAD" || ref == "master" || ref == "main") {
+		cmd := exec.Command("git", "-C", destDir, "checkout", ref)
```
banjoh: Same comment on using `exec.CommandContext`

pkg/getter/gitgetter.go
```diff
@@ -0,0 +1,277 @@
...
+
+// fullCloneAndCheckout performs a full clone and checks out a specific ref (commit SHA, tag, or branch)
+func (g *GitGetter) fullCloneAndCheckout(repoURL, ref, destDir string) error {
+	var stderr bytes.Buffer
+
+	// Remove the directory if it exists from failed shallow clone
+	os.RemoveAll(destDir)
+
+	// Full clone
+	cmd := exec.Command("git", "clone", repoURL, destDir)
```
banjoh: Use `exec.CommandContext` with timeout

pkg/getter/gitgetter.go
```diff
@@ -0,0 +1,277 @@
...
+	cmd.Stderr = &stderr
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("git clone failed: %s", stderr.String())
+	}
+
+	// Checkout the specific ref
+	cmd = exec.Command("git", "-C", destDir, "checkout", ref)
```
banjoh: Use `exec.CommandContext` with timeout

pkg/getter/gitgetter.go
```diff
@@ -0,0 +1,277 @@
...
+	cmd.Stderr = &stderr
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("git checkout failed for ref %q: %s", ref, stderr.String())
+	}
+
+	return nil
+}
+
+// packageChart packages a chart directory into a tarball
+func (g *GitGetter) packageChart(chartDir string) (*bytes.Buffer, error) {
```
banjoh: I do not think we should be calling the helm binary to package the chart. This command is implemented https://github.com/helm/helm/blob/main/pkg/action/package.go. I suggest utilising go implementing this in go using what we have in various packages in the project.

banjoh (changes requested):
Thanks for raising this PR. Its a valuable feature to have since git repositories are ubiquitous nowadays. Overall, your changes are OK and should work. 

My two concerns here are
- This implementation assumes that `git` will be available. SDK users cannot be expected to ship `git` as part of their applications. Are there git golang libraries we could rely on?
- How will authentication to git repositories be handled?

banjoh:
In addition to my comments, there is a long discussion https://github.com/helm/community/pull/321 on supporting git protocol for serving charts.

mms-gianni:
@banjoh Thanks for the review and for pointing out the HIP; I wasn't aware of it. I initially created this for myself but found it useful enough to share.

You're right about the timeouts, I'll add them.

While there are several good Go Git libraries (like [go-git/go-git](https://github.com/go-git/go-git)). I chose the exec approach to avoid complexity and keep the binary in its size. As you mentioned: I assumed someone willing to use this feature will have a installed and configured Git binary. But if preferred, i could also migrate to the library. 

Authentication is handled by the Git binary. SSH URL's use your local keys from your ssh-agent and for private HTTPS repositories, you can use Git credential helpers:

```bash
git config --global credential.helper store
```

banjoh (commented):
Code changes look good to me.

Git is a dependency that would not always be available. We'd need to communicate this to users and SDK clients. Such a chart cannot be installed since some dependencies would be missing. Should helm package / helm push refuse charts with git deps, or just warn via lint?

In addition to my comment above, the following items are still unresolved in https://github.com/helm/community/pull/321 HIP

  - Version semantics - version field won't be semver and may not match the version in the dependency's Chart.yaml. Will this be desired in Helm since users are accustomed to semver?
  - Sub-dependencies - How to handle git-sourced deps that have their own git-based sub-dependencies e.g submodules
  - Circular dependencies - Potential infinite loops if git deps reference each other
  - Recursive helm dep up - May need to run dependency updates inside cloned folders
  - Testing vs production workflow - Awkward UX when switching between git refs (for testing) and registry refs (for production)

mms-gianni:
Anything I can do to bring this topic forward?

banjoh:
> Anything I can do to bring this topic forward?

There is little to do in this PR. Any efforts would need to be directed to the HIP. Some questions there are fundamental to how Helm should work with releases.

PR #31555, opened 2025-11-21
Fix passing manifests to extism post-renderer plugins

Nitive:

Post-renderer plugins with the extism runtime were not working. The reason for that is that manifests were passed as buffer; therefore, they were marshaled as `"manifests": {}` for both plugin input and output. Type changed `*bytes.Buffer` -> `string` to avoid this issue. Also, tests were written to confirm that the code is working

**If applicable**:


Nitive:
@mattfarina @benoittgt Could you take a look, please?

PR #31563, opened 2025-11-23
fix: --dry-run=server now respect generateName

### LINKED ISSUE #31509: [Helm 4]  --dry-run=server with "generateName" creates error
### What happened?

Running `helm template . --dry-run=server` when a template contains `generateName`  generates an error. This used to work before helm 4.

```
Error: unable to continue with install: could not get information about the resource Job "" in namespace "hello-world": resource name may not be empty
```

No error.

Run `helm template . --dry-run=server` with the template below:

```
apiVersion: batch/v1
kind: Job
metadata:
  generateName: "hello-world-"
spec:
  template:
    spec:
      containers:
        - name: "hello-world"
          image: "hello-world"
      restartPolicy: Never

```

Error:

```
Error: unable to continue with install: could not get information about the resource Job "" in namespace "hello-world": resource name may not be empty
```

### Helm version

<details>

```console
$ helm version
version.BuildInfo{Version:"v4.0.0", GitCommit:"99cd1964357c793351be481d55abbe21c6b2f4ec", GitTreeState:"clean", GoVersion:"go1.25.3", KubeClientVersion:"v1.34"}
```
</details>

### Kubernetes version

<details>

```console
$ kubectl version
Client Version: v1.34.2
Kustomize Version: v5.7.1
Server Version: v1.34.1
```

</details>

Mujib-Ahasan:
Helm4 was facing problem with `--dry-run=server` as it was not respecting the `generateName`. The actual behavior is `generateName` is considered only in CREATE client method, not GET. A simple change is made, if `Name == ""`, this will be skipped to avoid error mentioned in issue #31509.

Other changes could also be possible, ex: we could `return` at the very first `Name == ""` instead of `continue`. This PR preferred continue because it is intended. 

**If applicable**:

closes #31509

### DIFF
--- a/pkg/action/validate.go
+++ b/pkg/action/validate.go
@@ -46,6 +46,10 @@ func requireAdoption(resources kube.ResourceList) (kube.ResourceList, error) {
 			return err
 		}
 
+		if info.Name == "" {
+			return nil
+		}
+
 		helper := resource.NewHelper(info.Client, info.Mapping)
 		_, err = helper.Get(info.Namespace, info.Name)
 		if err != nil {
@@ -68,6 +72,10 @@ func existingResourceConflict(resources kube.ResourceList, releaseName, releaseN
 	err := resources.Visit(func(info *resource.Info, err error) error {
 		if err != nil {
 			return err
+		}
+
+		if info.Name == "" {
+			return nil
 		}
 
 		helper := resource.NewHelper(info.Client, info.Mapping)
--- a/pkg/action/validate.go
+++ b/pkg/action/validate.go
@@ -47,7 +47,7 @@ func requireAdoption(resources kube.ResourceList) (kube.ResourceList, error) {
 		}
 
 		isGenerateName, err := validateNameAndgenerateName(info)
-		if isGenerateName {
+		if isGenerateName || err != nil {
 			return err
 		}
 
@@ -76,7 +76,7 @@ func existingResourceConflict(resources kube.ResourceList, releaseName, releaseN
 		}
 
 		isGenerateName, err := validateNameAndgenerateName(info)
-		if isGenerateName {
+		if isGenerateName || err != nil {
 			return err
 		}
 
@@ -216,7 +216,7 @@ func mergeStrStrMaps(current, desired map[string]string) map[string]string {
 func validateNameAndgenerateName(info *resource.Info) (bool, error) {
 	accessor, err := meta.Accessor(info.Object)
 	if err != nil {
-		return true, err
+		return false, err
 	}
 
 	if info.Name == "" && accessor.GetGenerateName() != "" {
--- a/pkg/action/validate.go
+++ b/pkg/action/validate.go
@@ -45,9 +45,11 @@ func requireAdoption(resources kube.ResourceList) (kube.ResourceList, error) {
 		if err != nil {
 			return err
 		}
-
-		if info.Name == "" {
+		accessor, _ := meta.Accessor(info.Object)
+		if info.Name == "" && accessor.GetGenerateName() != "" {
 			return nil
+		} else if info.Name != "" && accessor.GetGenerateName() != "" {
+			return fmt.Errorf("metadata.name and metadata.generateName cannot both be set")
 		}
 
 		helper := resource.NewHelper(info.Client, info.Mapping)
@@ -74,8 +76,11 @@ func existingResourceConflict(resources kube.ResourceList, releaseName, releaseN
 			return err
 		}
 
-		if info.Name == "" {
+		accessor, _ := meta.Accessor(info.Object)
+		if info.Name == "" && accessor.GetGenerateName() != "" {
 			return nil
+		} else if info.Name != "" && accessor.GetGenerateName() != "" {
+			return fmt.Errorf("metadata.name and metadata.generateName cannot both be set")
 		}
 
 		helper := resource.NewHelper(info.Client, info.Mapping)
--- a/pkg/action/validate.go
+++ b/pkg/action/validate.go
@@ -45,11 +45,10 @@ func requireAdoption(resources kube.ResourceList) (kube.ResourceList, error) {
 		if err != nil {
 			return err
 		}
-		accessor, _ := meta.Accessor(info.Object)
-		if info.Name == "" && accessor.GetGenerateName() != "" {
-			return nil
-		} else if info.Name != "" && accessor.GetGenerateName() != "" {
-			return fmt.Errorf("metadata.name and metadata.generateName cannot both be set")
+
+		skip, err := validateNameAndgenerateName(info)
+		if skip {
+			return err
 		}
 
 		helper := resource.NewHelper(info.Client, info.Mapping)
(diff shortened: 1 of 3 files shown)


pkg/kube/resource.go
```diff
@@ -29,6 +29,9 @@ func (r *ResourceList) Append(val *resource.Info) {
 // Visit implements resource.Visitor. The visitor stops if fn returns an error.
 func (r ResourceList) Visit(fn resource.VisitorFunc) error {
 	for _, i := range r {
+		if i.Name == "" {
```
gjenkins8: This function is a generic visitor for any/all scenarios. It should not implement business logic like skipping objects missing a name

gjenkins8 (changes requested):
https://github.com/helm/helm/pull/31563/files#r2554270362

pkg/kube/resource.go
Mujib-Ahasan: Thank you @gjenkins8 for pointing out. This now make sense to me that general visitor functions must not implement business logic. PTAL at new changes. Hope that would solve the problem.

gjenkins8 (changes requested):
Thanks for the updates!

The logic I think is complete. But can we (perhaps) create a helper function and include test coverage.

Mujib-Ahasan:
Thanks @gjenkins8 for the direction. As you mentioned, I created a helper function for that business logic and added test coverage(each separate commit).

pkg/action/validate.go
```diff
@@ -201,3 +209,14 @@ func mergeStrStrMaps(current, desired map[string]string) map[string]string {
...
+
+func validateNameAndgenerateName(info *resource.Info) (bool, error) {
+	accessor, _ := meta.Accessor(info.Object)
+	if info.Name == "" && accessor.GetGenerateName() != "" {
+		return true, nil
+	} else if info.Name != "" && accessor.GetGenerateName() != "" {
+		return true, fmt.Errorf("metadata.name and metadata.generateName cannot both be set")
+	}
+
+	return false, nil
```
gjenkins8: The return above make the else redundant here. Removing makes the logic easier to follow:

```suggestion
	accessor, _ := meta.Accessor(info.Object)
	if info.Name == "" && accessor.GetGenerateName() != "" {
		return true, nil
	}
	
	if info.Name != "" && accessor.GetGenerateName() != "" {
		return true, fmt.Errorf("metadata.name and metadata.generateName cannot both be set")
	}

	return false, nil
```

pkg/action/validate.go
```diff
@@ -46,6 +46,11 @@ func requireAdoption(resources kube.ResourceList) (kube.ResourceList, error) {
 			return err
 		}
 
+		skip, err := validateNameAndgenerateName(info)
+		if skip {
```
gjenkins8:
```suggestion
		isGenerateName, err := validateNameAndgenerateName(info)
		if isGenerateName {
```

and others below..

pkg/action/validate.go
```diff
@@ -201,3 +209,14 @@ func mergeStrStrMaps(current, desired map[string]string) map[string]string {
 	maps.Copy(result, desired)
 	return result
 }
+
+func validateNameAndgenerateName(info *resource.Info) (bool, error) {
```
gjenkins8:
```suggestion
// validateNameAndgenerateName validates that an object only has either `Name` or `GenerateName` set (and not both)
// If `GenerateName` is set, true is returned
// If an invalid combination of `Name` and `GenerateName` are set, an error is returned
func validateNameAndgenerateName(info *resource.Info) (bool, error) {
```

pkg/action/validate.go
```diff
@@ -201,3 +209,14 @@ func mergeStrStrMaps(current, desired map[string]string) map[string]string {
...
+	accessor, _ := meta.Accessor(info.Object)
```
gjenkins8: The error should be handled here I think? (any specific reason to not handle?)

```suggestion
	accessor, err := meta.Accessor(info.Object)
	if err != nil {
        ...
```

pkg/action/validate.go
Mujib-Ahasan: Yes, this sounds good!!

pkg/action/validate.go
Mujib-Ahasan: Well, I cannot remember the intension, but handling that would be a better solution. Thank you @gjenkins8

pkg/action/validate.go
```diff
@@ -71,6 +75,11 @@ func existingResourceConflict(resources kube.ResourceList, releaseName, releaseN
 			return err
 		}
 
+		isGenerateName, err := validateNameAndgenerateName(info)
+		if isGenerateName {
```
gjenkins8:
```suggestion
		if isGenerateName || err != nil {
```

Let's explicitly handle the error, rather than relying on `isGenerateName` being true

pkg/action/validate.go
```diff
@@ -46,6 +46,11 @@ func requireAdoption(resources kube.ResourceList) (kube.ResourceList, error) {
...
```
gjenkins8:
```suggestion
		if isGenerateName || err != nil {
```

Let's explicitly handle the error, rather than relying on `isGenerateName` being true

pkg/action/validate.go
```diff
@@ -201,3 +209,23 @@ func mergeStrStrMaps(current, desired map[string]string) map[string]string {
...
 	return result
 }
+
+// validateNameAndgenerateName validates that an object only has either `Name` or `GenerateName` set (and not both)
+// If `GenerateName` is set, true is returned
+// If an invalid combination of `Name` and `GenerateName` are set, an error is returned
+func validateNameAndgenerateName(info *resource.Info) (bool, error) {
+	accessor, err := meta.Accessor(info.Object)
+	if err != nil {
+		return true, err
```
gjenkins8:
```suggestion
		return false, err
```

`GenerateName` has not being determined to be true, we should not return true here

pkg/action/validate.go
```diff
@@ -201,3 +209,23 @@ func mergeStrStrMaps(current, desired map[string]string) map[string]string {
 	maps.Copy(result, desired)
 	return result
 }
+
+// validateNameAndgenerateName validates that an object only has either `Name` or `GenerateName` set (and not both)
+// If `GenerateName` is set, true is returned
+// If an invalid combination of `Name` and `GenerateName` are set, an error is returned
+func validateNameAndgenerateName(info *resource.Info) (bool, error) {
```
gjenkins8:
```suggestion
func validateNameAndGenerateName(info *resource.Info) (bool, error) {
```

Just noticed (on my final review) the casing isn't correctly "camelCase" here

pkg/action/validate.go
```diff
@@ -201,3 +209,23 @@ func mergeStrStrMaps(current, desired map[string]string) map[string]string {
 	maps.Copy(result, desired)
 	return result
 }
+
+// validateNameAndgenerateName validates that an object only has either `Name` or `GenerateName` set (and not both)
```
gjenkins8:
```suggestion
// validateNameAndGenerateName validates that an object only has either `Name` or `GenerateName` set (and not both)
```

As below

gjenkins8 (commented):
Thanks for the updates! All looks good, except I noticed (on my final pass though looking at the code) the mentioned casing issue

Mujib-Ahasan:
> Thanks for the updates! All looks good, except I noticed (on my final pass though looking at the code) the mentioned casing issue

Thank you @gjenkins8 for the reviews. Changes are made as suggested!

gjenkins8 (approved):

banjoh (approved):

gjenkins8 (approved):

PR #31564, opened 2025-11-23
Fix: generateName creates error

### LINKED ISSUE #31509: [Helm 4]  --dry-run=server with "generateName" creates error

saimanojk1:
**What this PR does / why we need it**: Closes #31509

**Special notes for your reviewer**: Skips checking for resource conflicts/require adoption if generateName is used for resource, i.e., Name is empty

**If applicable**:

### DIFF
--- a/pkg/action/validate.go
+++ b/pkg/action/validate.go
@@ -45,7 +45,10 @@ func requireAdoption(resources kube.ResourceList) (kube.ResourceList, error) {
 		if err != nil {
 			return err
 		}
-
+		// Using generatedName
+		if info.Name == "" {
+			return nil
+		}
 		helper := resource.NewHelper(info.Client, info.Mapping)
 		_, err = helper.Get(info.Namespace, info.Name)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/action/validate.go
```diff
@@ -45,7 +45,10 @@ func requireAdoption(resources kube.ResourceList) (kube.ResourceList, error) {
 		if err != nil {
 			return err
 		}
-
+		// Using generatedName
+		if info.Name == "" {
```
gjenkins8, line 49: a missing name does not explicitly mean `generateName` is present. IMHO we should only skip if `generateName` is present and `name` is omitted.

We should also follow the same error logic as `kubectl` if both `generateName` and `name` are present.

PR #31574, opened 2025-11-25
feat(helm): Add --output flag to version subcommand

vitorfloriano:
This PR introduces the `--output` flag to the `version` subcommand. This flag allows adding more formats to the `version` output without adding additional flags.

It also adds the `go` and `human` formats to the `--output` flag, for printing the version in a raw Go struct and a human-readable table format, respectively, as well as a proper error message:

```
$ helm version -o go
version.BuildInfo{Version:"v4.1+unreleased", GitCommit:"f05c21b6b6380d923696a7684c80a5e0847fbd7b", GitTreeState:"dirty", GoVersion:"go1.25.4", KubeClientVersion:"v1.34"}

$ helm version -o human
Version:             v4.1+unreleased
Git Commit:          f05c21b6b6380d923696a7684c80a5e0847fbd7b
Git Tree State:      dirty
Go Version:          go1.25.4
KubeClient Version:  v1.34

$ helm version -o asdf
Error: invalid output format: "asdf" (valid formats: "go", "human")
```

The current output of `version` has not been changed, for backwards compatibility:
```
$ helm version
version.BuildInfo{Version:"v4.1+unreleased", GitCommit:"f05c21b6b6380d923696a7684c80a5e0847fbd7b", GitTreeState:"dirty", GoVersion:"go1.25.4", KubeClientVersion:"v1.34"}
```
We may switch to the human-readable format as the default output for `version` in the next major release.

As follow-up, I suggest the following improvements:
- Update the help message for the version subcommand.
- Add shell completion for the available flags.
- Add other output formats, such as `yaml` and `json`.

**If applicable**:

### DIFF
--- a/pkg/cmd/version.go
+++ b/pkg/cmd/version.go
@@ -19,6 +19,7 @@ package cmd
 import (
 	"fmt"
 	"io"
+	"text/tabwriter"
 	"text/template"
 
 	"github.com/spf13/cobra"
@@ -41,6 +42,11 @@ version.BuildInfo{Version:"v3.2.1", GitCommit:"fe51cd1e31e6a202cba7dead9552a6d41
   built, and "dirty" if the binary was built from locally modified code.
 - GoVersion is the version of Go that was used to compile Helm.
 
+The --output flag allows you to change the representation with the following options:
+
+- "go": Prints the version in a raw Go struct.
+- "human": Prints the version in a human-readable format.
+
 When using the --template flag the following properties are available to use in
 the template:
 
@@ -55,6 +61,7 @@ For example, --template='Version: {{.Version}}' outputs 'Version: v3.2.1'.
 type versionOptions struct {
 	short    bool
 	template string
+	output   string
 }
 
 func newVersionCmd(out io.Writer) *cobra.Command {
@@ -73,6 +80,7 @@ func newVersionCmd(out io.Writer) *cobra.Command {
 	f := cmd.Flags()
 	f.BoolVar(&o.short, "short", false, "print the version number")
 	f.StringVar(&o.template, "template", "", "template for version string format")
+	f.StringVarP(&o.output, "output", "o", "", "output format (options: \"go\", \"human\")")
 
 	return cmd
 }
@@ -85,7 +93,18 @@ func (o *versionOptions) run(out io.Writer) error {
 		}
 		return tt.Execute(out, version.Get())
 	}
-	fmt.Fprintln(out, formatVersion(o.short))
+
+	switch o.output {
+	case "human":
+		return printVersionHuman(out)
+	case "go":
+		fmt.Fprintln(out, formatVersion(false))
+	case "":
+		fmt.Fprintln(out, formatVersion(o.short))
+		return nil
+	default:
+		return fmt.Errorf("invalid output format: %q (valid formats: \"go\", \"human\")", o.output)
+	}
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


benoittgt:
I am wondering how much this information is parsed by tools and if we should make it behind a `-o` flag or similar.

vitorfloriano:
> I am wondering how much this information is parsed by tools and if we should make it behind a `-o` flag or similar.

Hard to know that...but I think having the default output not human-readable is bad UX. The machine-readable output is the one that should be behind a `--json`  or `--verbose` flag or something like that.

OR, we could deprecate the `--short` flag and make it the new default, and make the current default output behind a `--verbose` flag, but that would be disruptive as well. 🤔

TerryHowe:
I was hoping @gjenkins8 would look at this because I thought he had a pull request still open around this. I agree there should be format support here. At least something like text or proper json.

banjoh:
> Hard to know that...but I think having the default output not human-readable is bad UX. The machine-readable output is the one that should be behind a `--json` or `--verbose` flag or something like that.

Changing the output format is considered a breaking change according to https://github.com/helm/community/blob/eba1acf2116793bcdc76bdfca8e5b06c13822890/hips/hip-0004.md#specification. It can be argued that since this getting version information of the helm binary, it falls under help text. I'm on the fence here, but I though I point this one out.

vitorfloriano:
@banjoh By reading the spec I understand that this would be a breaking change. Thanks for pointing that out. I could change the approach and add a flag for formatted print, but honestly I think this command has enough flags already 😅 What do y'all think?

mrlunchbox777:
I'd agree that it's a breaking change, there are almost certainly users relying on the output in the machine readable format. It feels like it'd be best to put it behind a `-o` to reduce the number of flags (vs adding `--human-readable`, `--json`, `--yaml`, etc). Then at the next major release switch the default to human readable, because that does feel like that should be the default behavior.

banjoh:
I agree with @mrlunchbox777 on adding a `-o` flag which follows how other cli tools used in the kuberneres ecosystem work.

It will allow adding more output formats. I suggest adding `short` flag there and commenting that in helm v5 it will be removed. It can be maeked as deprecated as well.

vitorfloriano:
@mrlunchbox777 @banjoh 
What would be the argument for the human-friendly format? `$ helm version -o human-readable`?

mrlunchbox777:
@vitorfloriano thank you for asking, for the `-o` arguments idea, in particular the `human-readable` option,  I was thinking that it would be a way to implement your improvement immediately as a flag, then later have it become the default. i.e.:

```yaml
helm 4: # after this MR
  go:
     - "helm version"
     - "helm version -o go"
  humanReadable:
     - "helm version -o human-readable"
  json:
     - "helm version -o json"
  yaml:
     - "helm version -o yaml"
  ...
helm 5:
  go:
     - "helm version -o go"
  humanReadable:
     - "helm version"
     - "helm version -o human-readable"
  json:
     - "helm version -o json"
  yaml:
     - "helm version -o yaml"
   ...
```

Edited to be clear, I just suggested some names that I thought might be good for the options for the arguments, but if there is something better I'd be in favor of that, e.g. `go-format` or `human`. I don't think this MR needs to add exhaustive, or even more than go and "human-readable", output conversions to this. Those should be additional issues.

mrlunchbox777:
FWIW, I'd lean towards `-o go` and `-o human` for these options, even though I suggested `-o human-readable` earlier

vitorfloriano:
@mrlunchbox777 @banjoh I have pivoted this PR to adding the `--ouput` flag with the `go` and `human` arguments instead. PTAL.

mrlunchbox777 (approved):

banjoh (approved):
LGTM

pkg/cmd/version.go
```diff
@@ -85,7 +93,18 @@ func (o *versionOptions) run(out io.Writer) error {
 		}
 		return tt.Execute(out, version.Get())
 	}
-	fmt.Fprintln(out, formatVersion(o.short))
+
+	switch o.output {
```
TerryHowe, line 97: Maybe I missed some discusssion, but I expected to see a more kubectl type set of outputs here

mrlunchbox777: that was me, [I suggested adding those as a follow on](https://github.com/helm/helm/pull/31574#issuecomment-3685115513), but I could also see it making more sense to put some more of those here.

PR #31583, opened 2025-11-26
feat(registry): support selecting artifacts from OCI Image Index

### LINKED ISSUE #31582: feat: Support selecting artifacts from OCI Image Index by artifactType
## What would you like to be added?

Support for selecting Helm chart artifacts from OCI Image Index manifests by `artifactType`.

When `helm pull` encounters an OCI Image Index (`application/vnd.oci.image.index.v1+json`):
1. Iterate through manifests in the index
2. Find the manifest with `artifactType: application/vnd.cncf.helm.config.v1+json` (or matching `config.mediaType`)
3. Pull that specific manifest

Example OCI Image Index containing both container image and Helm chart:

```json
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.index.v1+json",
  "manifests": [
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:abc...",
      "platform": { "architecture": "amd64", "os": "linux" }
    },
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:def...",
      "artifactType": "application/vnd.cncf.helm.config.v1+json"
    }
  ]
}
```

## Why is this needed?

Currently, `helm pull` fails when an OCI repository tag points to an Image Index:

```
Error: unable to retrieve blob with digest sha256:xxx: application/vnd.oci.image.index.v1+json: not found
```

Container runtimes (Docker, Podman) already handle Image Index by selecting manifests via `platform` field. Helm could use the same pattern with `artifactType`.

**Benefits:**
- Single source of truth: one tag for both the application image and its deployment chart
- Simplified CI/CD: publish once, consume differently based on tool
- Consistent versioning between image and chart
- Aligns with OCI 1.1 artifact model

**Current workarounds:**
- Use separate tags (`:0.0.5` for image, `:chart-0.0.5` for chart)
- Use separate repository paths (`/app` for image, `/app/chart` for chart)

**Related:**
- OCI Image Spec 1.1 Image Index: https://github.com/opencontainers/image-spec/blob/main/image-index.md
- OCI Distribution Spec 1.1 artifactType: https://github.com/opencontainers/distribution-spec/blob/main/spec.md
- Previous discussion: #8332

*Note: I understand significant features require a HIP. This issue is to gauge interest and gather feedback before potentially drafting one.*

lexfrei:

Implements OCI 1.1 artifact support for Helm charts (HIP: helm/community#424):

1. **Select a chart from an Image Index.** When a reference resolves to an Image Index with multiple manifests (a container image alongside a chart, or several charts), Helm selects the chart by descriptor `artifactType`, disambiguating multiple charts by the requested name then version, and falling back to the chart config for legacy indexes without annotations. An unresolvable choice is an error, never a silent guess.
2. **Set `artifactType` on push.** `helm push` sets the manifest `artifactType` and chart name/version annotations so charts are identifiable at the Index level without fetching each manifest.
3. **`--subject` for the Referrers API.** `helm push --subject <digest>` associates the chart with an existing artifact in the same repository; the digest is validated and resolved against the target repository to a full descriptor.

**Why this is needed:**

Since the index-pull regression was fixed (#31776), Helm traverses an Image Index but selects the chart by a hardcoded media-type scan, last-match-wins — order-dependent and able to return a mislabeled chart for multi-artifact/multi-chart indexes. This makes selection deterministic and lets a chart and its image be co-versioned under one reference.

**Usage:**

```bash
helm pull oci://registry/repo/chart --version 1.0.0           # selects the chart from an index
helm push chart.tgz oci://registry/repo --subject sha256:...  # referrers association
```

**Notes for reviewers:**

- Backward compatible: single-manifest charts are unaffected; a single chart in an index is selected regardless of name.
- `artifactType` is now set on every chart push (per OCI 1.1), which changes the manifest envelope digest versus older Helm for identical chart content; pre-1.1 registries should tolerate the optional field.
- `--subject` is per-repository; cross-repo subjects are out of scope (HIP Open Issues).

Closes #31582

**HIP**: helm/community#424

```release-note
`helm pull` now selects a chart from an OCI Image Index by artifactType (disambiguating multiple charts by name and version), and `helm push` sets the chart artifactType and adds a `--subject` flag to associate a chart with an existing same-repository artifact via the OCI Referrers API.
```

**If applicable**:

- [x] this PR contains user facing changes

### DIFF
--- a/pkg/registry/constants.go
+++ b/pkg/registry/constants.go
@@ -34,4 +34,8 @@ const (
 
 	// LegacyChartLayerMediaType is the legacy reserved media type for Helm chart package content.
 	LegacyChartLayerMediaType = "application/tar+gzip"
+
+	// ChartArtifactType is the artifact type for Helm charts in OCI v1.1+ Image Index.
+	// This is used to identify Helm chart manifests within a multi-artifact index.
+	ChartArtifactType = "application/vnd.cncf.helm.config.v1+json"
 )
--- a/pkg/registry/generic.go
+++ b/pkg/registry/generic.go
@@ -18,10 +18,13 @@ package registry
 
 import (
 	"context"
+	"encoding/json"
+	"fmt"
 	"io"
 	"net/http"
 	"slices"
 	"sort"
+	"strings"
 	"sync"
 
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
@@ -56,6 +59,15 @@ type GenericPullOptions struct {
 	SkipMediaTypes []string
 	// Custom PreCopy function for filtering
 	PreCopy func(context.Context, ocispec.Descriptor) error
+	// ArtifactType to select from OCI Image Index (empty means no filtering).
+	// When pulling from an Image Index containing multiple manifests,
+	// this field is used to select the manifest with matching artifactType.
+	ArtifactType string
+	// Selectors are descriptor annotations used to disambiguate when more than
+	// one manifest in an Image Index matches ArtifactType. A manifest matches
+	// only if its descriptor annotations contain every selector key with the
+	// same value. With a single matching manifest, selectors are not consulted.
+	Selectors map[string]string
 }
 
 // GenericPullResult contains the result of a generic pull operation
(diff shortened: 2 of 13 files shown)


pkg/registry/constants.go
```diff
@@ -34,4 +34,8 @@ const (
 
 	// LegacyChartLayerMediaType is the legacy reserved media type for Helm chart package content.
 	LegacyChartLayerMediaType = "application/tar+gzip"
+
+	// ChartArtifactType is the artifact type for Helm charts in OCI v1.1+ Image Index.
+	// This is used to identify Helm chart manifests within a multi-artifact index.
+	ChartArtifactType = "application/vnd.cncf.helm.config.v1+json"
```
atombrella: This is already defined in line 27.

pkg/registry/generic.go
```diff
@@ -83,6 +89,67 @@ func NewGenericClient(client *Client) *GenericClient {
...
+
+		// Check if config.mediaType matches our expected artifact type
+		if manifest.Config.MediaType == artifactType {
+			return candidate, nil
+		}
+	}
+
+	return ocispec.Descriptor{}, fmt.Errorf(
+		"no manifest with artifactType %q found in image index; available types: %v",
+		artifactType, availableTypes)
```
atombrella: Not sure if it's configurable with a linter, but I'd prefer `)` on a separate line. I think that's also more GoLang'ish style to end the last parameter with a  comma `,`

PR #31586, opened 2025-11-27
fix: use length check for MetaDependencies instead of nil comparison

calvinbui:

The MetaDependencies implementations for `v2Accessor` and `v3Accessor` always return a non-nil slice (using `make`), even when there are no dependencies. This made the `req != nil` condition always true and incorrectly treated empty dependency lists as present.

Update the logic to check `len(r) > 0 `instead of `r != nil` so that empty dependency slices are handled correctly.

Please close if this PR is incorrect.

**If applicable**:

### DIFF
--- a/pkg/action/package.go
+++ b/pkg/action/package.go
@@ -103,7 +103,7 @@ func (p *Package) Run(path string, _ map[string]interface{}) (string, error) {
 		ch.Metadata.AppVersion = p.AppVersion
 	}
 
-	if reqs := ac.MetaDependencies(); reqs != nil {
+	if reqs := ac.MetaDependencies(); len(reqs) > 0 {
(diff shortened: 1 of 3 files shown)


TerryHowe (approved):
/lgtm

banjoh (changes requested):
> v2Accessor and v3Accessor always return a non-nil slice (using make), even when there are no dependencies.

Can you add some tests that check this condition?

calvinbui:
> > v2Accessor and v3Accessor always return a non-nil slice (using make), even when there are no dependencies.
> 
> Can you add some tests that check this condition?

which file do i put them in?

banjoh:
> which file do i put them in?

Hmm. There isn't any value in the test now that I think more about it.

banjoh (approved):

PR #31589, opened 2025-11-28
Add new option to define api-versions from file

mozesn:
To implement compatibility with different K8S versions the Capabilities.APIVersions is used in the helm chart's template yaml or tpl files.

A helm version hardcodes a fixed set of APIVersions from a specific K8S release.

A new option is added to load the full list of APIVersions from files that "mocks" the API list of a given K8S version or distribution, so the default list can be overriden.

This way the compatibility of a helm chart can be tested offline using helm template with different set of K8S API Versions as input.

The output of helm template operation can be input to kubeval or kubeconform to test the schema compatibility.

Example: helm template mychart-1.0.0.tgz --api-versions-file k8s.v1.27.txt

where the k8s.v127.txt can contain the list of API versions. Example

```txt
v1
v1/Binding
v1/ComponentStatus
v1/ConfigMap
v1/Endpoints
v1/Event
v1/LimitRange
v1/Namespace
v1/Node
v1/PersistentVolume
v1/PersistentVolumeClaim
v1/Pod
v1/PodTemplate
v1/ReplicationController
v1/ResourceQuota
v1/Secret
v1/Service
v1/ServiceAccount
apps/v1
apps/v1/ControllerRevision
apps/v1/DaemonSet
apps/v1/Deployment
apps/v1/ReplicaSet
apps/v1/StatefulSet
etc.
```

**If applicable**:

### DIFF
--- a/pkg/cmd/template.go
+++ b/pkg/cmd/template.go
@@ -55,6 +55,7 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 	client := action.NewInstall(cfg)
 	valueOpts := &values.Options{}
 	var kubeVersion string
+	var apisFiles []string
 	var extraAPIs []string
 	var showFiles []string
 
@@ -94,6 +95,21 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 			client.ReleaseName = "release-name"
 			client.Replace = true // Skip the name check
 			client.APIVersions = common.VersionSet(extraAPIs)
+			if len(apisFiles) > 0 {
+				x := []string{}
+				for _, apisFile := range apisFiles {
+					data, err := os.ReadFile(apisFile)
+					if err != nil {
+						return err
+					}
+					for _, line := range strings.Split(string(data), "\n") {
+						if strings.TrimSpace(line) != "" {
+							x = append(x, strings.TrimSpace(line))
+						}
+					}
+				}
+				common.DefaultCapabilities.APIVersions = common.VersionSet(x)
+			}
 			client.IncludeCRDs = includeCrds
 			rel, err := runInstall(args, client, valueOpts, out)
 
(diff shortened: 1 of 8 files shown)


pkg/cmd/template.go
```diff
@@ -94,6 +95,17 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
 			client.ReleaseName = "release-name"
 			client.Replace = true // Skip the name check
 			client.APIVersions = common.VersionSet(extraAPIs)
+			if apisFiles != nil && len(apisFiles) > 0 {
```
banjoh: `len()` performs nil checks
```suggestion
			if len(apisFiles) > 0 {
```

pkg/cmd/template.go
```diff
@@ -94,6 +95,17 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
...
+				x := []string{}
+				for _, apisFile := range apisFiles {
+					data, err := os.ReadFile(apisFile)
+					if err != nil {
+						panic(err)
```
banjoh: Instead of `panic`king this ought to return an error.

pkg/cmd/template.go
```diff
@@ -94,6 +95,17 @@ func newTemplateCmd(cfg *action.Configuration, out io.Writer) *cobra.Command {
...
+					}
+					x = append(x, strings.Split(string(data), "\n")...)
+				}
+				common.DefaultCapabilities.APIVersions = common.VersionSet(x)
```
banjoh, line 111: Why not use `client.APIVersions` instead of replacing the default values?

With your changes (assuming use of `client.APIVersions`) `extraAPIs` input would be discarded. Is that intentional?

banjoh (changes requested):
Thanks for this change. The flag will be benefitial

Please add some tests to `TestTemplateCmd` to cover this new CLI flag

pkg/cmd/template.go
mozesn: thanks, will fix it.

mozesn:
thanks for the comment, I'll add tests

pkg/cmd/template.go
mozesn: the intention is to override the default capabilities, so we can input a full list of api versions a target k8s will have.

the extraApis are still added in case --api-versions option is used.

Maybe I can change the input parameter name to call it "--default-api-versions-file", instead?

banjoh: > Maybe I can change the input parameter name to call it "--default-api-versions-file", instead?

I that's the intention, this would be a better flag name choice

pkg/cmd/template.go
mozesn: thanks, fixed it

pkg/cmd/template.go
mozesn: I fixed the comments and added unit test

pkg/cmd/template.go
banjoh: Nit pick: would i5 be better to wrap the error in a message describing the error?

banjoh: Nit pick: would i5 be better to wrap the error in a message describing the error?

banjoh (approved):
LGTM

PR #31591, opened 2025-11-29
fix: escape non-ASCII characters in OCI annotation values; fixes #31507

bassg0navy:
Closes #31507

## Problem

Charts that include unicode characters (e.g., "Kröpke" in the maintainers field) fail to push to strict OCI registries like OCIR (Oracle Cloud Infrastructure Registry) due to manifest digest mismatches. This was first observed in [release 3.12.0 ](https://github.com/helm/helm/releases/tag/v3.12.0) after the addition of manifest annotations:

> When charts are pushed to OCI registries, annotations are attached using standard names that other tools can introspect (e.g. version)

## Solution

Added `escapeNonASCII()` function that converts non-ASCII characters to standard JSON Unicode escape sequences (e.g., `ö` → `\u00f6`). This allows manifest annotations to comply with the following OCI Spec recommendation:

> "Character set of the value SHOULD conform to alphanum of A-Za-z0-9 and separator set of -._:@/+"

Applied function to all annotation values in `generateChartOCIAnnotations()`.

**Before:** `"Jan-Otto Kröpke"` → OCIR corrupts to `"Jan-Otto Kr?pke"` → digest mismatch

**After:** `"Jan-Otto Kr\u00f6pke"` → Valid ASCII → push succeeds

## Testing

- Added unit tests for `escapeNonASCII()`
- Verified push to OCIR succeeds with prometheus-operator-crds chart
- All existing tests pass

**If applicable**:


pkg/registry/chart.go
```diff
@@ -65,14 +66,31 @@ annotations:
 	return ociAnnotations
 }
 
+func escapeNonASCII(s string) string {
+	var result strings.Builder
+	result.Grow(len(s)) // Pre-allocate for efficiency
+
+	for _, r := range s {
+		if r > 127 {
```
TerryHowe: Maybe unicode. MaxASCII here would be more clear

TerryHowe (commented):
Thanks for putting this together!

pkg/registry/chart.go
bassg0navy: Agreed; just updated.

TerryHowe (commented):
/lgtm

Thanks.

It probably doesn't matter, but a test with Chinese characters would be reassuring

TerryHowe (approved):

bassg0navy:
> /lgtm
> 
> Thanks.
> 
> It probably doesn't matter, but a test with Chinese characters would be reassuring

@TerryHowe Thanks for approval! I did add Chinese characters test case. This is my first time contributing. How should I go about getting approval for the 2 pending checks and final (write access) review?

benoittgt (commented):
This is good work ! Thanks

TerryHowe (changes requested):
```
golangci-lint run ./...
pkg/registry/chart.go:86:1: File is not properly formatted (gofmt)
 
^
pkg/registry/chart_test.go:414:1: File is not properly formatted (gofmt)
}
^
2 issues:
* gofmt: 2
make: *** [Makefile:132: test-style] Error 1
```

TerryHowe:
> @TerryHowe Thanks for approval! I did add Chinese characters test case. This is my first time contributing. How should I go about getting approval for the 2 pending checks and final (write access) review?

You'll need to be patient. Maintainers typically have $DAY_JOBS

bassg0navy:
> > @TerryHowe Thanks for approval! I did add Chinese characters test case. This is my first time contributing. How should I go about getting approval for the 2 pending checks and final (write access) review?
> 
> You'll need to be patient. Maintainers typically have $DAY_JOBS

Understood! Just added those newlines identified by `gofmt`; will be waiting patiently after that 😅

TerryHowe (approved):
/lgtm

```
before> ok  	helm.sh/helm/v4/pkg/registry	coverage: 71.5% of statements
after>  ok  	helm.sh/helm/v4/pkg/registry	coverage: 71.9% of statements
```

sabre1041 (commented):
Thanks for the contribution

While reviewing the guidance provided by the [OCI specification](https://github.com/opencontainers/image-spec/blob/main/annotations.md#pre-defined-annotation-keys), there does not appear to be any specific requirements that the fields that are being targeted in this PR must be in a specific format (other than a string).

This change could potentially introduce an undesirable/unexpected annotation value

TerryHowe:
> This change could potentially introduce an undesirable/unexpected annotation value

From the spec "Character set of the value SHOULD conform to alphanum of A-Za-z0-9 and separator set of -._:@/+" which is just a should and not a must. It also says nothing about unicode characters. The spec isn't great here, but it gives no support for unicode characters.

OCI CR does not support unicode characters in annotations.

bassg0navy:
Thanks for the response, guys. I acknowledge the OCI spec ambiguity and potential to break other Helm functions. 

What about an additional flag to allow folks w/ edge cases like mine? Example would be: "helm push --no-oci-annotations oci://ocir...<prometheus-repo>" to resort back to v3.11.3 behavior where no additional annotations are added to manifest.

TerryHowe (changes requested):
Yeh, Andrew is correct. That blurb I quoted is for ref.Name

TerryHowe:
I don't think this is going to happen. My understanding is this is an OCI CR

PR #31592, opened 2025-11-29
feat(create): add --chart-api-version flag (when HELM_EXPERIMENTAL_CHART_V3 env var is set)

banjoh:

Add `--chart-api-version` flag to helm create command to allow selecting chart API version (v2 or v3) when creating a new chart. It will allow adding Chart v3 specific features to created charts without breaking v2 charts.

- Default is v2 (existing behavior unchanged)
- v3 uses internal/chart/v3 scaffold generator
- Invalid versions return clear error message
- Works with `--starter` flag

**If applicable**:

### DIFF
--- a/internal/gates/gates.go
+++ b/internal/gates/gates.go
@@ -1,0 +1,21 @@
+/*
+Copyright The Helm Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package gates
+
+import "helm.sh/helm/v4/pkg/gates"
+
+// ChartV3 is the feature gate for chart API version v3.
+const ChartV3 gates.Gate = "HELM_EXPERIMENTAL_CHART_V3"
--- a/pkg/cmd/create.go
+++ b/pkg/cmd/create.go
@@ -23,6 +23,8 @@ import (
 
 	"github.com/spf13/cobra"
 
+	chartv3 "helm.sh/helm/v4/internal/chart/v3"
+	chartutilv3 "helm.sh/helm/v4/internal/chart/v3/util"
 	chart "helm.sh/helm/v4/pkg/chart/v2"
 	chartutil "helm.sh/helm/v4/pkg/chart/v2/util"
 	"helm.sh/helm/v4/pkg/cmd/require"
@@ -51,9 +53,10 @@ will be overwritten, but other files will be left alone.
 `
 
 type createOptions struct {
-	starter    string // --starter
-	name       string
-	starterDir string
+	starter         string // --starter
+	name            string
+	starterDir      string
+	chartAPIVersion string // --chart-api-version
 }
 
 func newCreateCmd(out io.Writer) *cobra.Command {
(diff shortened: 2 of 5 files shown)


TerryHowe (dismissed):

benoittgt:
It is working as excepted. Chart created is having one error because of https://github.com/SchemaStore/schemastore/blob/a43dd5e53d0f36f83376a0abd92fcd852ef4d409/src/schemas/json/chart.json#L5-L9 but this is a side subject. I've created https://github.com/helm/helm/issues/31593

Screenshot 2025-11-30 at 11 09 25

pkg/gates/gates.go
```diff
@@ -36,3 +36,6 @@ func (g Gate) IsEnabled() bool {
 func (g Gate) Error() error {
 	return fmt.Errorf("this feature has been marked as experimental and is not enabled by default. Please set %s=1 in your environment to use this feature", g.String())
 }
+
+// ChartV3 is the feature gate for chart API version v3.
+const ChartV3 Gate = "HELM_EXPERIMENTAL_CHART_V3"
```
mattfarina: This is a public constant. Following semver, it should not be removed until Helm v5. Even after the experiment is done it would still be present.

Can this be made an internal variable in the `cmd` package instead?

pkg/cmd/create.go
```diff
@@ -81,12 +85,28 @@ func newCreateCmd(out io.Writer) *cobra.Command {
 	}
 
 	cmd.Flags().StringVarP(&o.starter, "starter", "p", "", "the name or absolute path to Helm starter scaffold")
+	cmd.Flags().StringVar(&o.chartAPIVersion, "chart-api-version", chart.APIVersionV2, "chart API version to use (v2 or v3)")
```
mattfarina, line 88: Can this flag be hidden when the `HELM_EXPERIMENTAL_CHART_V3` experiment is not enabled.

pkg/cmd/create_test.go
```diff
@@ -60,27 +64,148 @@ func TestCreateCmd(t *testing.T) {
...
+				t.Fatalf("Could not write template: %s", err)
+			}
+
+			// Build the command
+			starterArg := "starterchart"
+			if tt.useAbsolutePath {
+				starterArg = filepath.Join(starterchart, "starterchart")
+			}
+			cmd := fmt.Sprintf("create --starter=%s", starterArg)
+			if tt.chartAPIVersion != "" {
```
mattfarina: Should this follow the convention used elsewhere with `if tt.chartAPIVersion == "v3"` for the same thing?

pkg/cmd/create_test.go
```diff
@@ -60,27 +64,148 @@ func TestCreateCmd(t *testing.T) {
...
+				t.Fatalf("no chart directory: %s", err)
+			} else if !fi.IsDir() {
+				t.Fatalf("chart is not directory")
+			}
+
+			// Load and verify the chart
+			var chartName, apiVersion string
+			var templates []string
+			if tt.chartAPIVersion == "v3" {
+				c, err := loaderv3.LoadDir(cname)
```
mattfarina: Did you consider using the common loader `pkg/chart/loader` and the accessor in `pkg/chart` to look at charts of both types?

mattfarina (dismissed):

pkg/cmd/create_test.go
banjoh: Change made

pkg/cmd/create_test.go
banjoh: Change made

pkg/cmd/create.go
banjoh: Change made

internal/gates/gates.go
```diff
@@ -0,0 +1,21 @@
+/*
```
banjoh, line 1: This was created as a separate internal package so it can be imported in various places without causing cyclic dependency issues.

sabre1041 (dismissed):
LGTM

TerryHowe (dismissed):
/lgtm

internal/gates/gates.go
scottrigby: good point. I initially thought maybe we should add to the internal/chart/v3 package, but i see now we can't because circular dep hell. then thought maybe we should add this to the public gates package, but then we'd need to keep it around even after charts v3 is out of experimental. so this makes sense to me.

small request - should we add this reasoning to a new `/internal/gates/doc.go` so this is clear to anyone who reads the code? Not a blocker for this PR.

pkg/cmd/create_test.go
```diff
@@ -46,41 +50,171 @@ func TestCreateCmd(t *testing.T) {
...
+		})
+	}
+}
+
+func TestCreateFileCompletion(t *testing.T) {
+	checkFileCompletion(t, "create", true)
+	checkFileCompletion(t, "create myname", false)
+}
+
+func TestCreateCmdChartAPIVersionV2(t *testing.T) {
```
scottrigby, line 210: like these 3 test funcs 👏

scottrigby (dismissed):
lgtm! my docs request is minor and could be a follow-up PR even.

pkg/gates/gates.go
scottrigby: the `cmd` package would cause circular dependencies too. This was made private though :+1:

scottrigby (approved):

promptless-for-oss:

New suggestion: **[Document experimental Charts v3 and --chart-api-version flag](https://app.gopromptless.ai/change-history/580e5bdf-32eb-4f43-a131-017c79611418)**

---
_Tip: Add or adjust Promptless's style guide in [Agent Knowledge Base](https://app.gopromptless.ai/configure/settings) ✍️_

PR #31595, opened 2025-11-30
fix: handle properly conditions with multilevel aliases; fixes 31187

### LINKED ISSUE #31187: Alias ignored in dependend subchart
### What happened?

I have a helmchart (top-level) which imports a subchart (second-level) twice. Same version with a different alias.
This subchart (second-level) has also a subchart (third-level), which could be disabled via condition.
Per default in the values of second-level disables the thrid-level-chart.

Now, when you render the top-level-chart the third-level-chart is rendered.

This does not happen when you import the second-level-chart only once.

Third-level-chart is NOT rendered and ignored

[chart-bug.zip](https://github.com/user-attachments/files/21983154/chart-bug.zip)

CD to top-level and do a:

```
❯ helm template .
---
# Source: top-level/charts/second-level-1/charts/third-level/templates/namespace.yaml
apiVersion: v1
kind: Namespace
metadata:
  name: release-name
  chart: third-level
  labels:
    name: release-name
```

### Helm version

<details>

```console
$ helm version
version.BuildInfo{Version:"v3.18.6", GitCommit:"b76a950f6835474e0906b96c9ec68a2eff3a6430", GitTreeState:"clean", GoVersion:"go1.24.6"}
```

</details>

### Kubernetes version

<details>

```console
$ kubectl version
Client Version: v1.33.4
Kustomize Version: v5.6.0
Server Version: v1.32.3
```

</details>

Antoinep77:

closes #31187

The issue was in the `processDependencyEnabled` function.
The chart dependencies are only copied with a shallow copy. When there are two versions of the same chart with aliases, then the metadata objects of their dependencies are the same.
This causes an issue if when a dependency has also an alias: its metadata name is mutated to the metadata alias (`line 161 of dependencies.go`) and as the copy is shallow, all metadatas of the copies of this chart are then changed and this cause an issue for the next copied dependencies processed in the `getAliasDependency (line 112)` function: there is a mismatch between the chart name and the dependency name, causing the condition to not be applied.

**If applicable**:

### DIFF
--- a/pkg/chartutil/dependencies.go
+++ b/pkg/chartutil/dependencies.go
@@ -187,7 +187,7 @@ Loop:
 	copy(cd, c.Dependencies()[:0])
 	for _, n := range c.Dependencies() {
 		if _, ok := rm[n.Metadata.Name]; !ok {
-			cd = append(cd, n)
+			cd = append(cd, deepCopyMetadataDependencies(n))
 		}
 	}
 	// don't keep disabled charts in metadata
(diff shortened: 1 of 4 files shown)


pkg/chartutil/dependencies.go
```diff
@@ -214,6 +214,18 @@ Loop:
 	return nil
 }
 
+func deepCopyMetadataDependencies(c *chart.Chart) *chart.Chart {
+	copy := *c
```
banjoh: nit: rename to `cCopy` or something else to avoid shadowing builtin `copy` function

pkg/chartutil/dependencies.go
```diff
@@ -214,6 +214,18 @@ Loop:
...
+	metadataCopy := *c.Metadata
```
banjoh, line 222: Check if `c.Metadata` is `nil`

Antoinep77:
Thank you for the feedback, I updated the PR

banjoh (approved):
LGTM

robertsirc:
Question: is this functionality reflective in Helm v4?

Antoinep77:
> Question: is this functionality reflective in Helm v4?

This bug is only in helm v3, helm v4 already handles this case properly

PR #31549, opened 2025-11-20
fix: prevent reporting fallback on version when none specified

benoittgt:
**What this PR does / why we need it**:

I inserted a regression via https://github.com/helm/helm/pull/31254. 

Fixes: https://github.com/helm/helm/issues/31548

**Special notes for your reviewer**:

Tested via

```sh
❯ helm pull rancher/longhorn-crd --version 106.2.0+up1.8.2 --destination /tmp/

level=WARN msg="unable to find exact version; falling back to closest available version" chart=longhorn-crd requested=106.2.0+up1.8.2 selected=106.2.0+up1.8.1

❯ bin/helm show chart brigade/brigade

apiVersion: v1
appVersion: v1.5.0
dependencies:
- condition: kashti.enabled
  name: kashti
  repository: https://brigadecore.github.io/charts
  version: 0.7.0
- condition: brigade-github-app.enabled
  name: brigade-github-app
  repository: https://brigadecore.github.io/charts
  version: 0.8.0
- alias: gw
  condition: gw.enabled
  name: brigade-github-oauth
  repository: https://brigadecore.github.io/charts
  version: 0.4.0
description: Brigade provides event-driven scripting of Kubernetes pipelines.
name: brigade
version: 1.10.0
```

**If applicable**:
- [ ] this PR contains user facing changes (the `docs needed` label should be applied if so)
- [ ] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

PR #31598, opened 2025-12-02
fix(repo/v1): suppress fallback warning when no version is specified #31548

maryamsul:
This PR fixes a warning in Helm v4 where the CLI emits  "unable to find exact version; falling back to closest available version" 
even when the user did not specify a chart version. The fix ensures that the warning is only logged when a  specific version is requested, preventing unnecessary warnings for users relying on the latest version of a chart.

###Changes Made:

Updated IndexFile.Get() in pkg/repo/v1/index.go to conditionally log warnings only when a specific version is requested.
Preserved the previous behavior for users who do specify a version.
Verified the fix locally with Helm commands.

##
run ./helm show chart brigade/brigade 

apiVersion: v1 
appVersion: v1.5.0 
dependencies: 
- condition: kashti.enabled 
  name: kashti 
  repository: https://brigadecore.github.io/charts 
  version: 0.7.0 
- condition: brigade-github-app.enabled 
  name: brigade-github-app 
  repository: https://brigadecore.github.io/charts 
  version: 0.8.0 
- alias: gw 
  condition: gw.enabled 
  name: brigade-github-oauth 
  repository: https://brigadecore.github.io/charts 
  version: 0.4.0 
description: Brigade provides event-driven scripting of Kubernetes pipelines. 

name: brigade 

version: 1.10.0

### DIFF
--- a/pkg/repo/v1/index.go
+++ b/pkg/repo/v1/index.go
@@ -200,29 +200,39 @@ func (i IndexFile) Get(name, version string) (*ChartVersion, error) {
 	}
 
 	// when customer inputs specific version, check whether there's an exact match first
-	if len(version) != 0 {
-		for _, ver := range vs {
-			if version == ver.Version {
-				return ver, nil
-			}
-		}
-	}
-
-	for _, ver := range vs {
-		test, err := semver.NewVersion(ver.Version)
-		if err != nil {
-			continue
-		}
-
-		if constraint.Check(test) {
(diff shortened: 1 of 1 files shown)


benoittgt:
Duplicate of https://github.com/helm/helm/pull/31549 ?

maryamsul:
> Duplicate of #31549 ?
@benoittgt 
Thanks for pointing it out! Yes, this PR overlaps with #31549. I opened mine
before noticing that a fix was already in review.

Since #31549 already has an approval and covers the same regression, I'm happy
to close this one to avoid duplication. Thanks!

Issue #31585, opened 2025-11-27
New watcher wait strategy lacking SDK logging


calvinbui:
### What would you like to be added?

I've updated our application to use v4 of the SDK.

With this change I switched to the new `watcher` wait strategy.

The log messages I get on his new strategy are severely lacking compared to the previous strategy.

New:

```
Pulled: 1234567890.dkr.ecr.us-east-2.amazonaws.com/helm-charts/app:1.2.3
Digest: sha256:3db06bdc755a136954f3142990b382745e2f487e32ee3d533aa7dd1f062c6952
preparing upgrade
determined release apply method
performing update
creating upgraded release
updating status for upgraded release
```

Legacy:

```
Pulled: 1234567890.dkr.ecr.us-east-2.amazonaws.com/helm-charts/app:1.2.3
Digest: sha256:3db06bdc755a136954f3142990b382745e2f487e32ee3d533aa7dd1f062c6952
preparing upgrade for default-myappv3
performing update for default-myappv3
creating upgraded release for default-myappv3
checking 6 resources for changes
Patch ServiceAccount "default-myappv3" in namespace default
Patch Service "default-myappv3-headless" in namespace default
I1117 04:57:36.937197   17214 warnings.go:110] "Warning: spec.SessionAffinity is ignored for headless services"
Patch Service "default-myappv3" in namespace default
Patch StatefulSet "default-myappv3" in namespace default
Patch Policy "default-myappv3" in namespace default
Patch TrustPolicy "default-myappv3" in namespace
waiting for release default-myappv3 resources (created: 0 updated: 6  deleted: 0)
beginning wait for 6 resources with timeout of 15m0s
StatefulSet is not ready: default/default-myappv3. 0 out of 2 expected pods have been scheduled
StatefulSet is not ready: default/default-myappv3. 0 out of 2 expected pods have been scheduled
StatefulSet is not ready: default/default-myappv3. 0 out of 2 expected pods have been scheduled
StatefulSet is not ready: default/default-myappv3. 0 out of 2 expected pods have been scheduled
StatefulSet is not ready: default/default-myappv3. 1 out of 2 expected pods have been scheduled
StatefulSet is not ready: default/default-myappv3. 1 out of 2 expected pods have been scheduled
StatefulSet is not ready: default/default-myappv3. 1 out of 2 expected pods are ready
StatefulSet is not ready: default/default-myappv3. 1 out of 2 expected pods are ready
[...]

PR #31605, opened 2025-12-04
Lack of logging solved for wait strategy for v4.

Mujib-Ahasan:
Previous legacy logging was much richer and nicer for developers, new logging lacks important information while having task. They all are enabled for the debug mode only. changes have been made so that users can get information as it runs.

I Think this simple change might not be the solution, but have open a PR and I would be happy to make changes according to what community suggests.

Related issue #31585 

**If applicable**:

### DIFF
--- a/pkg/kube/ready.go
+++ b/pkg/kube/ready.go
@@ -226,21 +226,21 @@ func (c *ReadyChecker) isPodReady(pod *corev1.Pod) bool {
 			return true
 		}
 	}
-	slog.Debug("Pod is not ready", "namespace", pod.GetNamespace(), "name", pod.GetName())
+	slog.Info("Pod is not ready", "namespace", pod.GetNamespace(), "name", pod.GetName())
 	return false
 }
 
 func (c *ReadyChecker) jobReady(job *batchv1.Job) (bool, error) {
 	if job.Status.Failed > *job.Spec.BackoffLimit {
-		slog.Debug("Job is failed", "namespace", job.GetNamespace(), "name", job.GetName())
+		slog.Info("Job is failed", "namespace", job.GetNamespace(), "name", job.GetName())
 		// If a job is failed, it can't recover, so throw an error
 		return false, fmt.Errorf("job is failed: %s/%s", job.GetNamespace(), job.GetName())
 	}
 	if job.Spec.Completions != nil && job.Status.Succeeded < *job.Spec.Completions {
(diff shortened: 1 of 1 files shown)


benoittgt:
Sorry @Mujib-Ahasan 
I missed that this was about the SDK. Since Helm v4, the logging library is `slog`, which is a bit different to hook into for the SDK. I’ve added an example in the original issue: https://github.com/helm/helm/issues/31585#issuecomment-3614145979

I don’t believe logs are missing. If there are gaps, we should add them as debug logs so they’re visible both in the SDK and in the CLI when using the `--debug` flag.

Mujib-Ahasan:
yes @benoittgt, so I think SDK consumers need to configure `--debug` by their own or they would not be able to catch important logs. And for CLI also( `--debug flag` ). Right?

benoittgt:
For the SDK, all logs emitted via slog are passed through, so we don’t handle the debug level ourselves—the SDK will manage log levels.
For the CLI, yes, you’re correct.

Mujib-Ahasan:
Ok so for SDK usage, the Helm code simply emits slog logs and it’s up to the embedding application to configure a slog handler if it wants to see debug output. right??

benoittgt:
> Ok so for SDK usage, the Helm code simply emits slog logs and it’s up to the embedding application to configure a slog handler if it wants to see debug output. right??

Yes. It was quite similar before in v3 but it was [`debugLog`](https://pkg.go.dev/helm.sh/helm/v3/pkg/action#DebugLog) via https://pkg.go.dev/helm.sh/helm/v3/pkg/action#Configuration.Init

Mujib-Ahasan:
I understand! I think for now we do not need any fix for this issue, do we??

benoittgt:
Yes. I don't think we can do more ATM. We may want to add more logs, like [I did few month ago](https://github.com/helm/helm/commit/f5aec508f5dd9e73ad668930743edb6742a87eb8), but it needs to be something we are missing.

Mujib-Ahasan:
yupp!! I understand, I think we should communicate to the issue creator about the natural behavior. Let's see the other opinions.

gjenkins8 (changes requested):
My understanding is that https://github.com/helm/helm/issues/31585 pertains to the Helm SDK, and the user needs to adjust (`slog`) verbosity with their usage. The change here will emit these traces for _all_ Helm CLI users (which isn't desirable)

benoittgt:
I think we can close this PR.

PR #31613, opened 2025-12-06
fix(getter): pass settings environment variables

aslafy-z:
This PR restores Helm settings environment variables for getter plugins.

**Commits:**
1. **Pass environment variables from CLI settings to getter plugins** - Calls `settings.EnvVars()` in `collectGetterPlugins()` to pass Helm environment to plugins
2. **Fix nil pointer dereference in `EnvSettings.Namespace()`** - Pre-existing bug triggered when `EnvVars()` is called on uninitialized `EnvSettings` (e.g., `&cli.EnvSettings{}` instead of `cli.New()`)

Fixes #31612 
Closes #31611 

The nil check fix addresses a pre-existing bug in `main` - calling `EnvVars()` on `&cli.EnvSettings{}` panics because `s.config` is nil. Existing tests using `&cli.EnvSettings{}` didn't trigger this because they never entered a code path that called `EnvVars()`.

**If applicable**:

### DIFF
--- a/pkg/getter/plugingetter.go
+++ b/pkg/getter/plugingetter.go
@@ -38,12 +38,14 @@ func collectGetterPlugins(settings *cli.EnvSettings) (Providers, error) {
 	if err != nil {
 		return nil, err
 	}
+	env := plugin.FormatEnv(settings.EnvVars())
 	pluginConstructorBuilder := func(plg plugin.Plugin) Constructor {
 		return func(option ...Option) (Getter, error) {
 
 			return &getterPlugin{
 				options: append([]Option{}, option...),
 				plg:     plg,
+				env:     env,
 			}, nil
 		}
 	}
@@ -91,6 +93,7 @@ func convertOptions(globalOptions, options []Option) schema.GetterOptionsV1 {
 type getterPlugin struct {
 	options []Option
 	plg     plugin.Plugin
+	env     []string
 }
 
 func (g *getterPlugin) Get(href string, options ...Option) (*bytes.Buffer, error) {
@@ -108,6 +111,7 @@ func (g *getterPlugin) Get(href string, options ...Option) (*bytes.Buffer, error
 			Options:  opts,
 			Protocol: u.Scheme,
 		},
+		Env: g.env,
 		// TODO should we pass Stdin, Stdout, and Stderr through Input here to getter plugins?
 		// Stdout: os.Stdout,
 	}
--- a/pkg/getter/plugingetter_test.go
+++ b/pkg/getter/plugingetter_test.go
@@ -144,3 +144,27 @@ func TestGetterPlugin(t *testing.T) {
 
 	assert.Equal(t, "fake-plugin output", buf.String())
 }
+
+func TestCollectGetterPluginsPassesEnv(t *testing.T) {
+	env := cli.New()
+	env.PluginsDirectory = pluginDir
+	env.Debug = true
+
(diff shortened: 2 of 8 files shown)


pkg/getter/plugingetter.go
```diff
@@ -38,12 +38,14 @@ func collectGetterPlugins(settings *cli.EnvSettings) (Providers, error) {
...
 		return nil, err
 	}
+	env := plugin.FormatEnv(settings.EnvVars())
 	pluginConstructorBuilder := func(plg plugin.Plugin) Constructor {
 		return func(option ...Option) (Getter, error) {
 
 			return &getterPlugin{
 				options: append([]Option{}, option...),
 				plg:     plg,
+				env:     env,
```
benoittgt, line 48: I think we should add small tests for this change and the next one in this file

benoittgt (commented):
Thanks for the clear issue. Maybe we could add tests.

pkg/getter/plugingetter.go
aslafy-z: I just added a test

pkg/getter/plugingetter_test.go
```diff
@@ -144,3 +144,27 @@ func TestGetterPlugin(t *testing.T) {
...
+	require.NotEmpty(t, providers, "expected at least one plugin provider")
+
+	getter, err := providers.ByScheme("test")
+	require.NoError(t, err)
+
+	gp, ok := getter.(*getterPlugin)
+	require.True(t, ok, "expected getter to be a *getterPlugin")
+
+	require.NotEmpty(t, gp.env, "expected env to be set on getterPlugin")
+	envMap := plugin.ParseEnv(gp.env)
```
TerryHowe, line 164: ParseEnv is exposed just for this test. It is an internal package though, so not as big a deal.

aslafy-z: If you want me to duplicate the logic instead, I can. I thought it would not be an issue to publish it as it is an internal package.

benoittgt:
I just tested on macOS, and it works as expected with the patch. I used the reproduction script from the issue: https://github.com/helm/helm/issues/31612#issue-3702117417

LGTM

benoittgt (approved):

aslafy-z:
Hi maintainers,

This PR fixes the downloader plugin env regression introduced by #31145 (missing HELM_* env vars), which is currently blocking some downloader plugins (example + repro in #31612).

Do you have an ETA for merge, and (if accepted) is this something you would consider backporting to the next 4.0.x patch release? The v4.0.4 release notes say 4.0.5 is planned for Jan 14, 2026 (and 4.1.0 for Jan 21, 2026). If a backport is needed, I'm happy to help prepare it.

@benoittgt would you agree to mention some other maintainer?

Thanks!

TerryHowe (approved):
/lgtm

aslafy-z:
/cc @banjoh @gjenkins8 @joejulian @mattfarina @robertsirc @sabre1041 @scottrigby @technosophos @TerryHowe 

Can you please give it a look?
Sorry for this massive ping, I'm not aware of another way to increase visibility before the release date hits.

robertsirc:
You can find us on slack you know.

banjoh (approved):
LGTM

aslafy-z:
Thank you @TerryHowe @banjoh 
Will it be automatically included into the next patch?

banjoh:
> Will it be automatically included into the next patch?

It should

PR #31616, opened 2025-12-07
fix(action): enable server-side validation for dry-run=server

### LINKED ISSUE #31505: Helm4 upgrade with dry-run and server-side enabled doesn't catch errors
Hi.

Let's assume we have the following chart:
1) Chart.yaml
```yaml
apiVersion: v2
name: helm4-test
version: 1.0.0
```
2) templates/deployment.yaml
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: bash
spec:
  replicas: 1
  selector:
    matchLabels:
      app.kubernetes.io/name: bash
      app.kubernetes.io/instance: bash
  template:
    metadata:
      labels:
        app.kubernetes.io/name: bash
        app.kubernetes.io/instance: bash
    spec:
      resources:
        requests:
          memory: "100Mi"
        limits:
          memory: "200Mi"
      containers:
        - image: bash
          imagePullPolicy: IfNotPresent
          name: bash
          command: ["bash", "-c", "echo 'Hello, World!'"]
      restartPolicy: Always
```

This is wrong spec - `resources` block is placed not under container.

Run helm dry-run with server-side:
```shell
helm upgrade test -n default --dry-run=server --server-side=true --install .
```

Result: No errors.

What is expected?

I expected that `Helm4` should return error that `resources` is unknown field for `spec`.

Because if we run actual `upgrade` without dry-run, we will get the error:
```shell
╰─ helm upgrade test -n default --server-side=true --install .
Release "test" does not exist. Installing it now.
Error: failed to create typed patch object (default/bash; apps/v1, Kind=Deployment): .spec.template.spec.resources: field not declared in schema
```

Comparing to `kubectl`, if we use just `deployment.yaml` in apply command, it will catch the error:
```shell
╰─ kubectl apply --dry-run=server --server-side=true -f manifests.yaml
Error from server: failed to create typed patch object (default/bash; apps/v1, Kind=Deployment): .spec.template.spec.resources: field not declared in schema
╰─ kubectl apply --dry-run=server -f manifests.yaml 
Error from server (BadRequest): error when creating "manifests.yaml": Deployment in version "v1" cannot be handled as a Deployment: strict decoding error: unknown field "spec.template.spec.resources"
```

Should `Helm4` with server-side work the same way? 

Otherwise, `dry-run` doesn't catch such errors and it can be a surprise in actual install/upgrade.

Tested on `1.28` k8s server
Helm `4.0.0`

biagiopietro:
When using `--dry-run=server` with `--server-side=true`, Helm now properly validates manifests against the Kubernetes API server. Previously, the `dry-run` would return early without calling the API, missing validation errors like unknown fields in the spec.

This fix ensures that DryRunServer mode calls KubeClient.Create/Update with the dry-run option, matching the behavior of kubectl apply --dry-run=server.

Fixes: helm/helm#31505

**If applicable**:


biagiopietro:
Tested also on `minikube` with `--kubernetes-version=v1.28.0`
```shell
➜  ~ minikube start --kubernetes-version=v1.28.0
😄  minikube v1.37.0 on Darwin 15.7.2 (arm64)
    ▪ MINIKUBE_CONFIF_FILE=/<redacted>/profiles/minikube/config.json
✨  Using the podman (experimental) driver based on user configuration
📌  Using Podman driver with root privileges
👍  Starting "minikube" primary control-plane node in "minikube" cluster
🚜  Pulling base image v0.0.48 ...
💾  Downloading Kubernetes v1.28.0 preload ...
    > preloaded-images-k8s-v18-v1...:  340.39 MiB / 340.39 MiB  100.00% 24.30 M
🔥  Creating podman container (CPUs=2, Memory=7865MB) ...
🐳  Preparing Kubernetes v1.28.0 on Docker 28.4.0 ...
🔗  Configuring bridge CNI (Container Networking Interface) ...
🔎  Verifying Kubernetes components...
    ▪ Using image gcr.io/k8s-minikube/storage-provisioner:v5
🌟  Enabled addons: storage-provisioner, default-storageclass

❗  /opt/homebrew/bin/kubectl is version 1.34.1, which may have incompatibilities with Kubernetes 1.28.0.
    ▪ Want kubectl v1.28.0? Try 'minikube kubectl -- get pods -A'
🏄  Done! kubectl is now configured to use "minikube" cluster and "default" namespace by default

➜  testHelm kubectl create namespace test
namespace/test created
➜  testHelm kubectl get all -n test
No resources found in test namespace.
➜  testHelm tree -l
.
├── Chart.yaml
└── templates
    └── deployment.yaml

2 directories, 2 files
➜  testHelm kubectl version
Client Version: v1.34.1
Kustomize Version: v5.7.1
Server Version: v1.28.0
Warning: version difference between client (1.34) and server (1.28) exceeds the supported minor version skew of +/-1
➜  testHelm /<redacted>/helm/bin/helm version
version.BuildInfo{Version:"v4.1+unreleased", GitCommit:"3214e803e2ada9dc6c1a8e8b1f258364685eb622", GitTreeState:"clean", GoVersion:"go1.25.1", KubeClientVersion:"v1.34"}
➜  testHelm /<redacted>/helm/bin/helm upgrade test -n test --dry-run=server --server-side=true --install .
Release "test" does not exist. Installing it now.
Error: failed to create typed patch object (test/bash; apps/v1, Kind=Deployment): .spec.template.spec.resources: field not declared in schema
```
switching to a newest `k8s` version `minikube`
```shell
➜  testHelm minikube delete
🔥  Deleting "minikube" in podman ...
🔥  Deleting container "minikube" ...
🔥  Removing /<redacted>/.minikube/machines/minikube ...
💀  Removed all traces of the "minikube" cluster.
➜  testHelm minikube start
😄  minikube v1.37.0 on Darwin 15.7.2 (arm64)
    ▪ MINIKUBE_CONFIF_FILE=/<redacted>/.minikube/profiles/minikube/config.json
✨  Using the podman (experimental) driver based on user configuration
📌  Using Podman driver with root privileges
👍  Starting "minikube" primary control-plane node in "minikube" cluster
🚜  Pulling base image v0.0.48 ...
🔥  Creating podman container (CPUs=2, Memory=7865MB) ...
🐳  Preparing Kubernetes v1.34.0 on Docker 28.4.0 ...
🔗  Configuring bridge CNI (Container Networking Interface) ...
🔎  Verifying Kubernetes components...
    ▪ Using image gcr.io/k8s-minikube/storage-provisioner:v5
🌟  Enabled addons: storage-provisioner, default-storageclass
🏄  Done! kubectl is now configured to use "minikube" cluster and "default" namespace by default
➜  testHelm kubectl create namespace test
namespace/test created
➜  testHelm kubectl version
Client Version: v1.34.1
Kustomize Version: v5.7.1
Server Version: v1.34.0
➜  testHelm /<redacted>/helm/bin/helm version
version.BuildInfo{Version:"v4.1+unreleased", GitCommit:"3214e803e2ada9dc6c1a8e8b1f258364685eb622", GitTreeState:"clean", GoVersion:"go1.25.1", KubeClientVersion:"v1.34"}
➜  testHelm /<redacted>/helm/bin/helm upgrade test -n test --dry-run=server --server-side=true --install .
Release "test" does not exist. Installing it now.
NAME: test
LAST DEPLOYED: Sun Dec  7 18:38:56 2025
NAMESPACE: test
STATUS: pending-install
REVISION: 1
DESCRIPTION: Dry run complete
TEST SUITE: None
HOOKS:
MANIFEST:
---
# Source: helm4-test/templates/deployment.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: bash
spec:
  replicas: 1
  selector:
    matchLabels:
      app.kubernetes.io/name: bash
      app.kubernetes.io/instance: bash
  template:
    metadata:
      labels:
        app.kubernetes.io/name: bash
        app.kubernetes.io/instance: bash
    spec:
      resources:
        requests:
          memory: "100Mi"
        limits:
          memory: "200Mi"
      containers:
        - image: bash
          imagePullPolicy: IfNotPresent
          name: bash
          command: ["bash", "-c", "echo 'Hello, World!'"]
      restartPolicy: Always
```

**Notes**

1. `Chart.yaml` and `templates/deployment.yml` contents, come from https://github.com/helm/helm/issues/31505#issue-3622717744
2. The changes add API call/s (otherwise how the spec will be validated?) that weren't happening before. For high-throughput CI/CD environments, this could impact API server load. (Although if user writes`dry-run=server` it's expected that _they should know that a call to the server happens_. 
3. Server-side `dry-run` requires the user to have permissions for the `dry-run` operation as now the API call is happening. Users who previously didn't need certain permissions might now encounter authorization errors. (Indeed before there was no call to the server now there are).

pkg/action/install.go
```diff
@@ -394,6 +394,26 @@ func (i *Install) RunWithContext(ctx context.Context, ch ci.Charter, vals map[st
...
+				_, err = i.cfg.KubeClient.Update(
+					toBeAdopted,
+					resources,
+					kube.ClientUpdateOptionServerSideApply(i.ServerSideApply, i.ForceConflicts),
+					kube.ClientUpdateOptionDryRun(true),
+				)
+			}
+			if err != nil {
+				return rel, err
+			}
```
gjenkins8: I think it is wanted/needed to split the adopted/created resources?

```suggestion
		if i.DryRunStrategy == DryRunServer {
			if len(toBeAdopted) > 0 {
				_, err := i.cfg.KubeClient.Update(
					toBeAdopted,
					resources,
					kube.ClientUpdateOptionServerSideApply(i.ServerSideApply, i.ForceConflicts),
					kube.ClientUpdateOptionDryRun(true),
				)
				if err != nil {
				    return rel, err
			    }
			}
			if len(resources) > 0 {
			    _, err := i.cfg.KubeClient.Create(
					resources,
					kube.ClientCreateOptionServerSideApply(i.ServerSideApply, false),
					kube.ClientCreateOptionDryRun(true),
				)
				if err != nil {
				    return rel, err
			    }
			}
```

I also suspect the error handling could be improved in the above perhaps. By not failing on the first error?

biagiopietro: @gjenkins8  thanks for finding the time for the review, I applied your feedback, could you please have a look when you have time? Thanks in advance.

pkg/action/install.go
```diff
@@ -401,6 +401,34 @@ func (i *Install) RunWithContext(ctx context.Context, ch ci.Charter, vals map[st
...
+			if len(resources) > 0 {
+				_, err := i.cfg.KubeClient.Create(
+					resources,
+					kube.ClientCreateOptionServerSideApply(i.ServerSideApply, false),
+					kube.ClientCreateOptionDryRun(true),
+				)
+				if err != nil {
+					errs = append(errs, err)
+				}
+			}
```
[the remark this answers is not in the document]
biagiopietro: I hope last commit makes you happy Copilot!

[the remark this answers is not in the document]
biagiopietro: Hello @gjenkins8 
Is there any pending action on my side to unblock the merge of this PR? thanks in advance 🙇

pkg/action/upgrade.go
```diff
@@ -394,6 +394,19 @@ func (u *Upgrade) performUpgrade(ctx context.Context, originalRelease, upgradedR
 
 	if isDryRun(u.DryRunStrategy) {
 		u.cfg.Logger().Debug("dry run for release", "name", upgradedRelease.Name)
+		// For server-side dry-run, validate resources against the API server
+		if u.DryRunStrategy == DryRunServer {
+			_, err := u.cfg.KubeClient.Update(
+				current,
+				target,
+				kube.ClientUpdateOptionForceReplace(u.ForceReplace),
+				kube.ClientUpdateOptionServerSideApply(serverSideApply, u.ForceConflicts),
```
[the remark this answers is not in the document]
biagiopietro: done

pkg/action/install.go
```diff
@@ -416,6 +416,27 @@ func (i *Install) RunWithContext(ctx context.Context, ch ci.Charter, vals map[st
...
+				_, err = i.cfg.KubeClient.Create(
+					resources,
+					kube.ClientCreateOptionServerSideApply(i.ServerSideApply, false),
+					kube.ClientCreateOptionDryRun(true),
+				)
+			} else if len(resources) > 0 {
+				_, err = i.cfg.KubeClient.Update(
+					toBeAdopted,
+					resources,
+					kube.ClientUpdateOptionServerSideApply(i.ServerSideApply, i.ForceConflicts),
```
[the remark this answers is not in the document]
biagiopietro: done

pkg/action/upgrade_test.go
```diff
@@ -637,6 +637,55 @@ func TestUpgradeRelease_DryRun(t *testing.T) {
...
+	req.NoError(upAction.cfg.Releases.Create(rel))
+
+	expectedErr := errors.New("validation error: unknown field in spec")
+	config.KubeClient.(*kubefake.FailingKubeClient).UpdateError = expectedErr
+	upAction.DryRunStrategy = DryRunServer
+
+	vals := map[string]interface{}{}
+	ctx, done := context.WithCancel(t.Context())
+	_, err := upAction.RunWithContext(ctx, rel.Name, buildChart(), vals)
+	done()
```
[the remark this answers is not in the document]
biagiopietro: it's like that in other part of tests. won't change it in scope of this PR

pkg/action/upgrade.go
```diff
@@ -397,6 +397,26 @@ func (u *Upgrade) performUpgrade(ctx context.Context, originalRelease, upgradedR
...
 	if isDryRun(u.DryRunStrategy) {
 		u.cfg.Logger().Debug("dry run for release", "name", upgradedRelease.Name)
+		// For server-side dry-run, validate resources against the API server
+		if u.DryRunStrategy == DryRunServer {
+			// Ensure this validation request is strictly non-mutating by always using
+			// server-side apply and disabling force-replace, regardless of the user's
+			// normal upgrade options.
+			dryRunServerSideApply := true
+			dryRunForceReplace := false
+			upgradeClientSideFieldManager := isReleaseApplyMethodClientSideApply(originalRelease.ApplyMethod) && dryRunServerSideApply
```
TerryHowe: Is it true the underlying code doesn't honor the dry run for delete?

biagiopietro: good catch, `helm upgrade --dry-run=server` on a release where the new manifest drops a resource will actually delete that resource from the cluster, unless it carries `helm.sh/resource-policy: keep` which is checked at 👇 

https://github.com/helm/helm/blob/c000a408b81266d328db64aea2ff9e2e70844790/pkg/kube/client.go#L680-L683

I will update the code.

biagiopietro: pushed new changes in last commit

pkg/action/install.go
```diff
@@ -421,6 +421,34 @@ func (i *Install) RunWithContext(ctx context.Context, ch ci.Charter, vals map[st
 
 	// Bail out here if it is a dry run
 	if isDryRun(i.DryRunStrategy) {
+		// For server-side dry-run, validate resources against the API server.
+		// Force server-side apply in this path because kube dry-run semantics are
+		// only honored by the server-side apply create/update code paths.
+		if i.DryRunStrategy == DryRunServer {
```
TerryHowe: `--create-namespace` regression: this block runs *before* the `CreateNamespace` block below (~L466), so a server-side apply into a not-yet-created namespace returns `namespaces "x" not found`. `helm install --dry-run=server --create-namespace` goes from succeeding to erroring. Same class of problem for any chart that templates a Namespace plus resources inside it — the dry-run Namespace apply doesn't persist, so everything in it 404s.

pkg/action/install.go
```diff
@@ -421,6 +421,34 @@ func (i *Install) RunWithContext(ctx context.Context, ch ci.Charter, vals map[st
...
+			serverSideDryRun := true
```
TerryHowe: Forcing SSA ignores `i.ServerSideApply`. On `--server-side=false` the manifests get validated under SSA semantics — including field-manager conflict detection — that the real apply will never perform, so the dry run can fail for an install that would succeed. The comment above is factually right about the client's limitation, but the consequence is that `--dry-run=server` stops previewing what will actually happen. Prefer using the real apply mode and explicitly skipping/erroring when it's client-side.

pkg/action/install.go
```diff
@@ -421,6 +421,34 @@ func (i *Install) RunWithContext(ctx context.Context, ch ci.Charter, vals map[st
...
+					resources,
+					kube.ClientCreateOptionServerSideApply(serverSideDryRun, false),
+					kube.ClientCreateOptionDryRun(true),
+				)
+			} else if len(resources) > 0 {
+				updateThreeWayMergeForUnstructured := i.TakeOwnership && !serverSideDryRun
+				_, err = i.cfg.KubeClient.Update(
+					toBeAdopted,
+					resources,
+					kube.ClientUpdateOptionForceReplace(i.ForceReplace),
```
TerryHowe: `Client.Update` rejects this combination outright (`invalid operation: cannot use server-side apply and force replace together`), so `helm install --dry-run=server --server-side=false --force-replace --take-ownership` fails on an option error before any validation happens. The upgrade dry-run path hardcodes `dryRunForceReplace := false` for exactly this reason — install should match.

pkg/action/install.go
```diff
@@ -421,6 +421,34 @@ func (i *Install) RunWithContext(ctx context.Context, ch ci.Charter, vals map[st
...
+					kube.ClientUpdateOptionServerSideApply(serverSideDryRun, i.ForceConflicts),
+					kube.ClientUpdateOptionDryRun(true),
+					kube.ClientUpdateOptionThreeWayMergeForUnstructured(updateThreeWayMergeForUnstructured),
+					kube.ClientUpdateOptionUpgradeClientSideFieldManager(true),
```
TerryHowe: This costs an extra GET+PATCH per resource on every dry run. It does honor `dryRun` (`upgradeClientSideFieldManager(original, updateOptions.dryRun, ...)`), so it's safe — but given the added API-server load you flagged in the PR description, worth gating rather than passing `true` unconditionally.

pkg/action/upgrade.go
```diff
@@ -397,6 +397,26 @@ func (u *Upgrade) performUpgrade(ctx context.Context, originalRelease, upgradedR
 
 	if isDryRun(u.DryRunStrategy) {
 		u.cfg.Logger().Debug("dry run for release", "name", upgradedRelease.Name)
+		// For server-side dry-run, validate resources against the API server
+		if u.DryRunStrategy == DryRunServer {
+			// Ensure this validation request is strictly non-mutating by always using
+			// server-side apply and disabling force-replace, regardless of the user's
+			// normal upgrade options.
+			dryRunServerSideApply := true
```
TerryHowe: Same concern as install: this forces SSA regardless of the `serverSideApply` value already computed by `getUpgradeServerSideValue` and passed into `performUpgrade` as a parameter. A `--server-side=false` upgrade gets validated under semantics its real apply won't use.

pkg/kube/client.go
```diff
@@ -902,7 +902,7 @@ func (c *Client) Update(originals, targets ResourceList, options ...ClientUpdate
 		}
 	}
 
-	return c.update(originals, targets, createApplyFunc, makeUpdateApplyFunc())
+	return c.update(originals, targets, createApplyFunc, makeUpdateApplyFunc(), updateOptions.dryRun)
```
TerryHowe, line 946: `dryRun` is still ignored on every client-side path: `makeCreateApplyFunc` returns bare `createResource` when `serverSideApply` is false, and `patchResourceClientSide`/`replaceResource` take no `dryRun` at all. After this change, `Update(..., ClientUpdateOptionDryRun(true), ClientUpdateOptionServerSideApply(false))` dry-runs the deletes but *really performs* the creates and patches — a worse failure mode than before, because dry-run now looks supported. Suggest rejecting `dryRun && !serverSideApply` in the option validation above, in the same style as the existing `forceConflicts && !serverSideApply` guard.

pkg/action/install_test.go
```diff
@@ -451,6 +451,41 @@ func TestInstallRelease_DryRunClient(t *testing.T) {
...
+	req := require.New(t)
+
+	config := actionConfigFixtureWithDummyResources(t, createDummyResourceList(false))
+
+	instAction := NewInstall(config)
+	instAction.Namespace = "spaced"
+	instAction.ReleaseName = "test-server-dry-run"
+
+	expectedErr := errors.New("validation error: unknown field in spec")
+	config.KubeClient.(*kubefake.FailingKubeClient).CreateError = expectedErr
```
TerryHowe: `PrintingKubeClient.Create`/`Update` discard the option varargs entirely, so this never asserts that `ClientCreateOptionDryRun(true)` was passed. Delete that option from `install.go` and this test still passes — while Helm would perform a real apply on `--dry-run=server`. Needs an option recorder on the fake; `TestUpdateDryRunDelete` in `pkg/kube` is the right shape.

pkg/action/upgrade_test.go
```diff
@@ -619,6 +619,55 @@ func TestUpgradeRelease_DryRun(t *testing.T) {
...
+	upAction := NewUpgrade(config)
+	upAction.Namespace = "spaced"
+
+	rel := releaseStub()
+	rel.Name = "test-server-dry-run"
+	rel.Info.Status = common.StatusDeployed
+	req.NoError(upAction.cfg.Releases.Create(rel))
+
+	expectedErr := errors.New("validation error: unknown field in spec")
+	config.KubeClient.(*kubefake.FailingKubeClient).UpdateError = expectedErr
```
TerryHowe: Same gap as the install test: removing `kube.ClientUpdateOptionDryRun(true)` from `upgrade.go` leaves this passing. The pairing against a client-side case that must *not* call the server is the right structure — it just can't see the option that makes the call non-mutating.

TerryHowe (changes requested):
Right root cause, and plumbing `dryRun` into `deleteResource` is a sharp catch — without it this change would have made `helm upgrade --dry-run=server` genuinely delete pruned resources. `TestUpdateDryRunDelete` is well built.

Blocking on three things inline: the `ForceReplace`+SSA combination in `install.go` (a straight port of upgrade's `dryRunForceReplace := false` fixes it), the unguarded `dryRun && !serverSideApply` hole in `pkg/kube/client.go`, and the `--create-namespace` ordering regression. The forced-SSA design question is worth a maintainer opinion before more code moves.

Two things not anchorable to a diff line:

- `pkg/action/rollback.go:216` has the same early return, so `helm rollback --dry-run=server` still validates nothing. Not required here, but worth calling out or filing as a follow-up.
- The "user facing changes" checkbox should be checked. This adds API calls where there were none, newly requires create/patch/**delete** RBAC verbs for a dry run, and will fail against admission webhooks that don't declare `sideEffects: None`/`NoneOnDryRun`. All of that is documented in your PR comment — it belongs in the description with the `docs needed` label.

pkg/action/upgrade.go
biagiopietro: Yes, that was the case before this PR. I've since added a guard in pkg/kube/client.go that rejects dryRun && !serverSideApply, and deleteResource already received the dryRun flag in the earlier commit, so deletes are now properly non-mutating too.

pkg/action/install.go
biagiopietro: Good catch. I've restructured the flow so the client-side dry-run bails out before CreateNamespace, and the server-side dry-run validation now runs after namespace creation. So --create-namespace --dry-run=server no longer fails with a namespace-not-found error.

pkg/action/install.go
biagiopietro: Agreed on both counts. I've dropped the forced serverSideDryRun := true and now return an explicit error when --dry-run=server is used with --server-side=false, rather than quietly validating under SSA semantics that the real install would never use. Same fix applied to upgrade.

pkg/action/install.go
biagiopietro: Fixed, ForceReplace is now hardcoded to false in the dry-run path, matching what upgrade already did.

pkg/action/install.go
biagiopietro: Gated it on i.TakeOwnership now. It only makes sense to upgrade field manager ownership when we are actually adopting existing resources, so passing true unconditionally was unnecessary load.

pkg/action/upgrade.go
biagiopietro: Same fix applied here. The code now uses the serverSideApply value already computed by getUpgradeServerSideValue, and returns an error if it is false rather than overriding it.

pkg/kube/client.go
biagiopietro: Added a guard in Update() that returns an error when dryRun && !serverSideApply, in the same style as the existing forceConflicts && !serverSideApply check. The client-side create and patch paths genuinely cannot honor dry-run, so failing loudly is safer than silently performing real writes.

pkg/action/install_test.go
biagiopietro: Fixed. Added ParseCreateOptions and ParseUpdateOptions helpers to pkg/kube that apply the option funcs and return the resolved values. FailingKubeClient now records the last call's parsed options in RecordedCreateOptions and RecordedUpdateOptions, and the test asserts RecordedUpdateOptions.DryRun is true for --dry-run=server. The test was also restructured to actually reach the dry-run validation path, since it was previously stopping at the ownership conflict check and accidentally passing because that error message also contains the string 'validation error'.

pkg/action/upgrade_test.go
biagiopietro: Same fix as the install test. ServerSideApply and the release ApplyMethod are now explicitly set so the real SSA path is exercised, and the assertion on RecordedUpdateOptions.DryRun ensures the option actually makes it to the client.

biagiopietro:
Any update on this team?

PR #31630, opened 2025-12-10
Fix: Data Race Bug detected via TestConcurrenyDownloadIndex

### LINKED ISSUE #31536: Data Race Bug detected via TestConcurrenyDownloadIndex
### What happened?

When the test _TestConcurrenyDownloadIndex_ is run a data race can occur and cause the test suite to fail. 

The test to always pass and no race to be detected.

Run the test with race detection on until you experience the error. For example...

```
go test -timeout 30s -race -run ^TestConcurrenyDownloadIndex$ helm.sh/helm/v4/pkg/repo/v1
```

### Helm version

<details>

```console
$ helm version --short
v4.1+unreleased+ga75fcc2
```
</details>

### Kubernetes version

N/A

haojingcn:
close #31536 

Add a RWMutex in struct ChartRepository, for avoid client get data race within concurrency.

**If applicable**:

### DIFF
--- a/pkg/repo/v1/chartrepo.go
+++ b/pkg/repo/v1/chartrepo.go
@@ -28,7 +28,6 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
-	"sync"
 
 	"helm.sh/helm/v4/internal/fileutil"
 	"helm.sh/helm/v4/pkg/getter"
@@ -54,7 +53,6 @@ type ChartRepository struct {
 	IndexFile *IndexFile
 	Client    getter.Getter
 	CachePath string
-	mutex     sync.RWMutex
 }
 
 // NewChartRepository constructs ChartRepository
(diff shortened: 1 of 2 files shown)


TerryHowe (commented):
/lgtm

gjenkins8 (changes requested):
The point of `TestConcurrenyDownloadIndex` (sic) as far as I can tell to show that concurrent actions on index files are correct.

>	// Simultaneously start multiple goroutines that:
	// 1) download index.yaml via DownloadIndexFile (write operation),
	// 2) read index.yaml via LoadIndexFile (read operation).
	// This checks for race conditions and ensures correct behavior under concurrent read/write access.

The change here serializes all calls `DownloadIndexFile` (for the given `ChartRepository` instance). Which is simply serializing the parallel calls to `DownloadIndexFile` from the test (may as well delete the test at this point)

Instead, IMHO the test should be amended to use a `ChartRepository` per go-routine, to avoid concurrent operation on the same `ChartRepository` instance. Which retains validating the concurrent index access, without concurrent `ChartRepository` access.

mrlunchbox777 (approved):
validated that these tests changes work and it keep concurrency by creating a `NewChartRepository` in each goroutine.

PR #31637, opened 2025-12-11
[v3 backport] fix: Use latest patch release of Go in releases

### LINKED ISSUE #31634: Helm releases with older Go minors
### What happened?

Helm releases went out with older patch versions of Go being used. If there are vulnerabilities in the older minors, Helm will be released with known vulnerabities.

I expect Helm releases to be built with the latest Go patch releases.

Check the latest releases of Helm and compare the version of Go used along with the release dates.

### Helm version

N/A

### Kubernetes version

N/A

mattfarina:
GitHub Actions has a cache of tools, like Go, and it does not update this cache when a new version comes out. It can take more than a week for a new version to be available.

This change forces the action to check if a newer version is available than is in the cache.

Closes #31634

(cherry picked from commit 8f636b5bd229c071ab5f0237a7b4553337392b44)

**If applicable**:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -31,6 +31,7 @@ jobs:
         uses: actions/setup-go@41dfa10bad2bb2ae585af6ee5bb4d7d973ad74ed # pin@5.1.0
         with:
           go-version: '${{ env.GOLANG_VERSION }}'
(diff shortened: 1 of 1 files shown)


robertsirc (approved):
LGTM

PR #31644, opened 2025-12-12
fix(values): preserve nil values when chart default is empty map

banjoh:

Only delete nil user values when overriding a non-nil chart default. When chart has empty map or no default for a key, preserve user's nil.

| Scenario | Result |
|----------|--------|
| User sets `baz: ~`, chart has `baz: "value"` | Key deleted | 
| User sets `baz: ~`, chart has empty map `{}` | Nil preserved |
| User sets `baz: ~`, chart has `baz: ~` | Nil preserved |

Fixes #31643

**If applicable**:

### DIFF
--- a/pkg/chart/common/util/coalesce_test.go
+++ b/pkg/chart/common/util/coalesce_test.go
@@ -25,6 +25,7 @@ import (
 	"text/template"
 
 	"github.com/stretchr/testify/assert"
+	req "github.com/stretchr/testify/require"
 
 	"helm.sh/helm/v4/pkg/chart/common"
 	chart "helm.sh/helm/v4/pkg/chart/v2"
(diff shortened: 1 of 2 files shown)


robertsirc (dismissed):
LGTM

pkg/chart/common/util/coalesce_test.go
```diff
@@ -25,6 +25,7 @@ import (
 	"text/template"
 
 	"github.com/stretchr/testify/assert"
+	req "github.com/stretchr/testify/require"
```
gjenkins8: nit: why not the standard import name ie. `require` ?
```suggestion
	"github.com/stretchr/testify/require"
```

pkg/chart/common/util/coalesce_test.go
```diff
@@ -401,131 +402,120 @@ func TestMergeValues(t *testing.T) {
...
+		}
 
-	if addr2["street"].(string) != "123 Spouter Inn Ct." {
-		t.Errorf("Unexpected address: %v", addr2["street"])
-	}
+		// What we expect is that anything in dst overrides anything in src, but that
+		// otherwise the values are coalesced.
+		CoalesceTables(dst, src)
+
+		is.Equal("Ishmael", dst["name"], "Unexpected name: %s", dst["name"])
```
gjenkins8: Would it be okay to split this refactoring of conditional logic to `assert` into a new PR. Why `assert` (IMHO) is much better for test readability, it is difficult here to review any logic changes (due to any changes in the code logic)

pkg/chart/common/util/coalesce_test.go
```diff
@@ -401,131 +402,120 @@ func TestMergeValues(t *testing.T) {
...
-
-	if dst2["name"] != "Ishmael" {
-		t.Errorf("Unexpected name: %s", dst2["name"])
-	}
-
-	addr2, ok := dst2["address"].(map[string]interface{})
-	if !ok {
-		t.Fatal("Address went away.")
-	}
+	t.Run("case 1", func(t *testing.T) {
```
gjenkins8: Perhaps this and the below `t.Run(...` separate test cases?

(but see my comment below about creating a new PR for these refactoring changes)

banjoh:
@gjenkins8 I removed all the test refactoring changes and left the unit test validating the reported bug

KrzysztofDziankowski:
I have verified the fix for my use case and it appears to work.
Could this pull request be included in the next Helm 3.20.x release?

TerryHowe (approved):
/lgtm

chojnack:
Could you please merge this into 3.20 branch too? We could build that ourselves then for our testing.

PR #31647, opened 2025-12-13
Enable the errname linter.

atombrella:

https://github.com/Antonboom/errname has some recommendations to suffix error names with `Error`.
I'm not sure if Helm is an API, and as such, if changing names should be considered harmful(!). Anyway, I think enabling more of the linters in `golangci-lint` where they make sense is a good idea.

I did toy around with `testifylint`, but changes are pretty substantial. `errname` was pretty simple to enable.

**If applicable**:


internal/chart/v3/lint/support/message.go
```diff
@@ -35,27 +35,27 @@ var sev = []string{"UNKNOWN", "INFO", "WARNING", "ERROR"}
...
-	Messages []Message
+	Messages []MessageError
 	// The highest severity of all the failing lint rules
 	HighestSeverity int
 	ChartDir        string
 }
 
 // Message describes an error encountered while linting.
-type Message struct {
+type MessageError struct {
```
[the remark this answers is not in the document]
atombrella: `LintMessage` or `LintMessageError` are both fine with me. It's possible to silence warnings about single instances locally with an `nolint` annotation.

pkg/chart/v2/lint/support/message.go
```diff
@@ -35,27 +35,27 @@ var sev = []string{"UNKNOWN", "INFO", "WARNING", "ERROR"}
 
 // Linter encapsulates a linting run of a particular chart.
 type Linter struct {
-	Messages []MessageError
+	Messages []LintMessageError
```
TerryHowe, line 38: Uggh, I realized these v2 changes are technically breaking changes for SDK users and it looks like they are generalized in `pkg/action/lint.go`. Unfortunately, we might not be able to make this change until v5 which would be a ways out.

TerryHowe (changes requested):

pkg/chart/v2/lint/support/message.go
atombrella: So resetting stuff under `v2` would be a good idea, then. It's possible to silence golangci-lint for certain directories.

gjenkins8 (changes requested):
Thanks for the PR.

As Terry mentioned, Helm is used as an API (SDK). So we must retain (public/non-internal) API compatibility.

Options I can think of:
1. Exclude non-internal directories (plus a TODO note/comment to remove for Helm v5 / next major version where breaking changes can happen)
2. Alias existing type names to the new type names (and use `//nolint:errname` if needed on those types e.g. `type Message = LintMessageError`

atombrella:
> Thanks for the PR.

You're welcome. I'm learning, and enjoy the rapid feedback process. It's motivating :smile: 
 
> As Terry mentioned, Helm is used as an API (SDK). So we must retain (public/non-internal) API compatibility.
> 
> Options I can think of:
>     2. Alias existing type names to the new type names (and use `//nolint:errname` if needed on those types e.g. `type Message = LintMessageError`

I went with the second option. `joinedListError` is internal, and thus hasn't been aliased. I think it'd be easy to remove the aliases when it is safe to do so.

atombrella:
One thing that strikes me when working on this is the absence of more custom error types. I think it'd make testing easier, rather than checking against the error message.

pkg/chart/v2/lint/support/message.go
atombrella: `Message` has been retained as an alias to `LintMessageError`.

PR #31649, opened 2025-12-14
Use errors.Is to check for io.EOF and gzip.ErrHeader

atombrella:

In GoLang, using the `==` operator to check for a certain error will not unwrap the error chain, and therefore may hide the problem.

These are recommendations from [errorlint](https://codeberg.org/polyfloyd/go-errorlint) It is somehow also how I have been taught to check for errors in GoLang. Note that changes to enable `errorlint` would be a little too overwhelming. Perhaps it's by design that `errors.Is` is not used to check for the affected instances.

**If applicable**:


TerryHowe (approved):
/lgtm

gjenkins8 (approved):

PR #31654, opened 2025-12-16
fix(doc): Update default wait strategy

deepakchethan:
Closes: #31652

Default wait strategy is hookOnly, so updated the docs to reflect the same

I reordered the values list, please let me know if you think otherwise.

**If applicable**:
- [X] this PR contains user facing changes (the `docs needed` label should be applied if so)

### DIFF
--- a/pkg/cmd/flags.go
+++ b/pkg/cmd/flags.go
@@ -59,7 +59,7 @@ func AddWaitFlag(cmd *cobra.Command, wait *kube.WaitStrategy) {
 	cmd.Flags().Var(
 		newWaitValue(kube.HookOnlyStrategy, wait),
 		"wait",
-		"if specified, wait until resources are ready (up to --timeout). Values: 'watcher' (default), 'hookOnly', and 'legacy'.",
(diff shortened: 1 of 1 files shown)


pkg/cmd/flags.go
```diff
@@ -59,7 +59,7 @@ func AddWaitFlag(cmd *cobra.Command, wait *kube.WaitStrategy) {
 	cmd.Flags().Var(
 		newWaitValue(kube.HookOnlyStrategy, wait),
 		"wait",
-		"if specified, wait until resources are ready (up to --timeout). Values: 'watcher' (default), 'hookOnly', and 'legacy'.",
+		"if specified, wait until resources are ready (up to --timeout). Values: 'hookOnly' (default), 'watcher', and 'legacy'.",
```
TerryHowe: That works fine, but It would be easier to do this I think:
```suggestion
		"if specified, wait until resources are ready (up to --timeout). Values: 'watcher', 'hookOnly', and 'legacy'.",
```

deepakchethan: Thank you for the review @TerryHowe, i have updated it as suggested

TerryHowe (approved):
/lgtm

TerryHowe:
Thanks!

PR #31669, opened 2025-12-21
fix: #31665 add default cases to switch statements

### LINKED ISSUE #31665: Switch statements missing enum-cases
I tried to experiment with [exhaustive](https://golangci-lint.run/docs/linters/configuration/#exhaustive) There are some instances where the switch-statement is missing one or more values from the `enum`. I don't yet know the code base particularly well, so I'm raising a question here whether this linter has highlighted some actual problems. The posted result is just a part of the findings.

```
internal/statusreaders/pod_status_reader.go:61:2: missing cases in switch of type v1.PodPhase: v1.PodPending, v1.PodRunning, v1.PodUnknown (exhaustive)
	switch corev1.PodPhase(phase) {
	^
pkg/chart/common.go:205:3: missing cases in switch of type reflect.Kind: reflect.Invalid, reflect.Bool, reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64, reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64, reflect.Uintptr, reflect.Float32, reflect.Float64, reflect.Complex64, reflect.Complex128, reflect.Array, reflect.Chan, reflect.Func, reflect.Interface, reflect.Map, reflect.String, reflect.UnsafePointer (exhaustive)
		switch value.Kind() {
		^
```

Note that some of the highlighted switch-case statements that are found when you enable the `exhaustive` linter in [.golangci.yml](https://github.com/helm/helm/blob/main/.golangci.yml) do not have the `default` case/branch.

brendenehlers:
closes: #31665 

The exhaustive linter will fail some switch statements because they don't have a `default` branch. This fix just moves some code around to include a default branch or adds a default branch with the comment `// intentionally left empty` for the cases where the intended behavior is to do nothing.

The default (no pun intended) behavior of the exhaustive linter requires covering all enum branches, without allowing a backup the default branch. I changed the [default-signifies-exhaustive](https://golangci-lint.run/docs/linters/configuration/#exhaustive) setting to true in `.golangci.yml` so it'll pass switch statements that have a default branch, since that's a more sensible default in my opinion.

**If applicable**:

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -73,6 +73,9 @@ linters:
               recommendations:
                 - github.com/evanphx/json-patch/v5
 
+    exhaustive:
+      default-signifies-exhaustive: true
+
 run:
   timeout: 10m
 
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -32,6 +32,7 @@ linters:
     - unused
     - usestdlibvars
     - usetesting
+    - exhaustive
 
   exclusions:
   
--- a/internal/statusreaders/pod_status_reader.go
+++ b/internal/statusreaders/pod_status_reader.go
@@ -86,19 +86,19 @@ func podConditions(u *unstructured.Unstructured) (*status.Result, error) {
 				},
 			},
 		}, nil
-	}
-
-	message := "Pod in progress"
-	return &status.Result{
-		Status:  status.InProgressStatus,
-		Message: message,
-		Conditions: []status.Condition{
-			{
-				Type:    status.ConditionReconciling,
-				Status:  corev1.ConditionTrue,
-				Reason:  "PodInProgress",
-				Message: message,
+	default:
+		message := "Pod in progress"
+		return &status.Result{
+			Status:  status.InProgressStatus,
+			Message: message,
+			Conditions: []status.Condition{
+				{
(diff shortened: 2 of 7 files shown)


.golangci.yml
```diff
@@ -73,6 +73,9 @@ linters:
               recommendations:
                 - github.com/evanphx/json-patch/v5
 
+    exhaustive:
```
atombrella, line 77: The linter is not enabled under the section at line 20.

brendenehlers: I went back and forth on this, because IMO the exhaustive linter is a good idea to have enabled by default, but on the other hand I didn't want to unilaterally make that decision if there had been previous discussions on it. What're your thoughts?

The default was left there because, if someone were to enable the exhaustive linter, it will _not_ flag switch statements that have default branch instead of covering every enum variant.

TerryHowe: If this is all that it takes to pass, enable it.

brendenehlers: Alright change has been pushed.

TerryHowe (approved):
/lgtm

gjenkins8 (approved):

PR #31675, opened 2025-12-24
fix(ignore): correct negation pattern handling in .helmignore

ma-ts:
## Summary
- Fix broken negation pattern (`!pattern`) handling in `.helmignore` that caused unrelated files like `Chart.yaml` to be incorrectly ignored
- Implement proper gitignore semantics where later rules override earlier ones and negation patterns re-include previously excluded files

## Problem
When a `.helmignore` file contained any negation pattern, files that did NOT match the negation pattern were incorrectly ignored. For example:

```
*.md
!README.md
```

This would cause `Chart.yaml` to be ignored, breaking `helm lint`, `helm template`, and `helm install` with the error: `Chart.yaml file is missing`

## Root Cause
The `Ignore()` function in `pkg/ignore/rules.go` had two bugs:
1. **Inverted logic**: When a file did NOT match a negation pattern, the function returned `true` (ignore)
2. **Early termination**: The function returned on the first matching rule, preventing negation rules from overriding earlier exclusions

## Solution
Rewrote `Ignore()` to follow proper gitignore semantics:
- Process all rules in order, tracking the ignore state
- The last matching rule wins
- Negation patterns (`!pattern`) set `ignored=false` when they match
- Regular patterns set `ignored=true` when they match

### DIFF
--- a/pkg/ignore/rules.go
+++ b/pkg/ignore/rules.go
@@ -85,8 +85,9 @@ func Parse(file io.Reader) (*Rules, error) {
 
 // Ignore evaluates the file at the given path, and returns true if it should be ignored.
 //
-// Ignore evaluates path against the rules in order. Evaluation stops when a match
-// is found. Matching a negative rule will stop evaluation.
+// Ignore evaluates path against all rules in order. Later rules can override earlier ones,
+// following gitignore semantics where negation patterns (starting with !) can re-include
+// previously excluded files.
 func (r *Rules) Ignore(path string, fi os.FileInfo) bool {
 	// Don't match on empty dirs.
 	if path == "" {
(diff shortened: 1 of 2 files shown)


TerryHowe:
This is probably related or a dup of https://github.com/helm/helm/pull/12265 which is under discussion https://github.com/helm/community/pull/426

ma-ts:
I don't know how I missed that one! Probably because it's a relatively old PR. @TerryHowe : what is the current status of discussing this? I think especially the negation patterns would really hit other developers (also in a very unknown way, it took me around an hour to investigate this).

TerryHowe:
Well, this project in general is a bit of a slow and careful project because of the blast radius of mistakes, but I would expect that HIP to move pretty (relatively) quickly.

PR #31678, opened 2025-12-26
fix(plugins): reset cache repo modifications before update

### LINKED ISSUE #31664: Helm plugin update fails
### What happened?

Helm fails to update some plugins.
```
$ helm plugin update unittest
Error: failed to update plugin unittest, got error (plugin repo was modified)
```

helm should update the plugin without error

```
$ helm plugin install https://github.com/helm-unittest/helm-unittest.git --verify=false
[...]

$ helm plugin update unittest
Error: failed to update plugin unittest, got error (plugin repo was modified)
```

This is because the install/update process update the plugin.yaml file in the `$HOME/.local/share/helm/plugins/helm-unittest.git` folder. The git repo is then considered dirty: https://github.com/helm/helm/blob/4d1150d2af2b5c3b922b816757b6772bf3b49dfc/internal/plugin/installer/vcs_installer.go#L98-L103

```
$ cd $HOME/.local/share/helm/plugins/helm-unittest.git
$ git git diff | cat
diff --git a/plugin.yaml b/plugin.yaml
index 7d15b83..47a2ea5 100644
--- a/plugin.yaml
+++ b/plugin.yaml
@@ -3,26 +3,7 @@ version: "1.0.3"
 usage: "unittest for helm charts"
 description: "Unit test for helm chart in YAML with ease to keep your chart functional and robust."
 ignoreFlags: false
-platformCommand:
-  - command: "$HELM_PLUGIN_DIR/untt"
-  - os: windows
-    command: "$HELM_PLUGIN_DIR\\untt.exe"
-platformHooks:
-  install:
-    - command: "$HELM_PLUGIN_DIR/install-binary.sh"
-    - os: windows
-      command: "powershell.exe"
-      args:
-        - -ExecutionPolicy
-        - Bypass
-        - -File
-        - "$HELM_PLUGIN_DIR\\install-binary.ps1"
-  update:
-    - command: "$HELM_PLUGIN_DIR/install-binary.sh"
-    - os: windows
-      command: "powershell.exe"
-      args:
-        - -ExecutionPolicy
-        - Bypass
-        - -File
-        - "$HELM_PLUGIN_DIR\\install-binary.ps1"
+command: "$HELM_PLUGIN_DIR/untt"
+hooks:
+  install: "cd $HELM_PLUGIN_DIR; ./install-binary.sh"
+  update: "cd $HELM_PLUGIN_DIR; HELM_PLUGIN_UPDATE=1 ./install-binary.sh"
```

Happens with helm 3 & 4.

### Helm version

<details>

```console
$ helm version
version.BuildInfo{Version:"v4.0.3", GitCommit:"9db13ee5c343196f642c568a03e58d3221b324d6", GitTreeState:"clean", GoVersion:"go1.25.5", KubeClientVersion:"v1.34"}
```

</details>

### Kubernetes version

<details>

```console
$ kubectl version
N/A
```

</details>

OlTrenin:
## Description

  This PR fixes the plugin update failure issue where Helm reports "plugin repo was modified" error even when users haven't made any manual modifications to the plugin repository.

  Closes #31664

  During plugin installation, Helm modifies the `plugin.yaml` file in the cached git repository (located at `$HOME/.local/share/helm/cache/plugins/`). When users attempt to update the plugin later, the `Update()` function checks if the repository is dirty and returns an error if any modifications are detected. This prevents plugin updates from succeeding.

  This PR automatically resets local modifications in the cached repository before attempting to update. Since the cached repository is managed entirely by Helm and should not contain user modifications, this is a safe operation that allows updates to proceed successfully.

  **Changes**:
  - Added `resetRepo()` function that discards local modifications using VCS-specific commands (git reset, hg revert, etc.)
  - Modified `Update()` to automatically reset the cache repository if it's dirty before pulling updates
  - Added comprehensive tests to verify the reset functionality works correctly

  - The fix only affects the **cached** repository (`~/.local/share/helm/cache/plugins/`), not the actual plugin directory (`~/.local/share/helm/plugins/`) where users might make local modifications
  - The `resetRepo()` function supports all VCS types that Helm supports: Git, Mercurial, Bazaar, and SVN
  - The original dirty check was added in commit 1e8ebae2 to provide better error messages, but it didn't account for Helm's own modifications during installation
  - This is a safe change because the cache directory is an internal implementation detail managed solely by Helm

  **If applicable**:
    - *This is a bug fix that makes updates work as expected - no documentation changes needed*
  - [ ]  this PR contains unit tests
    - *Added `TestResetRepo()` to test the reset functionality*
    - *Updated `TestVCSInstallerUpdate()` to verify automatic reset during updates*
  - [ ]  this PR has been tested for backwards compatibility
    - *The change only affects error cases that previously failed, making them succeed*
    - *No API changes, no breaking changes to existing functionality*

  **Testing**:

  The fix was tested with:
  1. Unit tests that simulate the exact scenario from issue #31664
  2. All existing plugin installer tests pass
  3. The new `TestResetRepo()` verifies git reset works correctly
  4. The updated `TestVCSInstallerUpdate()` confirms updates succeed after modifications

  **Reproducing the original issue** (before this PR):
  ```bash
  helm plugin install https://github.com/helm-unittest/helm-unittest.git
  helm plugin update unittest
  # Error: failed to update plugin unittest, got error (plugin repo was modified)

  After this PR:
  helm plugin install https://github.com/helm-unittest/helm-unittest.git
  helm plugin update unittest

### DIFF
--- a/internal/plugin/installer/vcs_installer.go
+++ b/internal/plugin/installer/vcs_installer.go
@@ -21,6 +21,7 @@ import (
 	stdfs "io/fs"
 	"log/slog"
 	"os"
+	"os/exec"
 	"sort"
 
 	"github.com/Masterminds/semver/v3"
@@ -95,12 +96,63 @@ func (i *VCSInstaller) Install() error {
 	return fs.CopyDir(i.Repo.LocalPath(), i.Path())
 }
 
+// resetRepo discards all local modifications in the repository.
+// This is used to clean the cached repository before updating.
+func resetRepo(repo vcs.Repo) error {
+	// Check the VCS type to determine the appropriate reset command
+	switch repo.Vcs() {
+	case vcs.Git:
+		// For Git, use 'git reset --hard' to discard all local changes
+		cmd := exec.Command("git", "reset", "--hard")
+		cmd.Dir = repo.LocalPath()
+		if output, err := cmd.CombinedOutput(); err != nil {
+			return fmt.Errorf("git reset failed: %w, output: %s", err, output)
+		}
+		return nil
+	case vcs.Hg:
+		// For Mercurial, use 'hg revert --all --no-backup'
+		cmd := exec.Command("hg", "revert", "--all", "--no-backup")
+		cmd.Dir = repo.LocalPath()
+		if output, err := cmd.CombinedOutput(); err != nil {
+			return fmt.Errorf("hg revert failed: %w, output: %s", err, output)
+		}
+		return nil
+	case vcs.Bzr:
+		// For Bazaar, use 'bzr revert'
+		cmd := exec.Command("bzr", "revert")
+		cmd.Dir = repo.LocalPath()
+		if output, err := cmd.CombinedOutput(); err != nil {
+			return fmt.Errorf("bzr revert failed: %w, output: %s", err, output)
+		}
+		return nil
+	case vcs.Svn:
+		// For SVN, use 'svn revert -R .'
+		cmd := exec.Command("svn", "revert", "-R", ".")
+		cmd.Dir = repo.LocalPath()
+		if output, err := cmd.CombinedOutput(); err != nil {
+			return fmt.Errorf("svn revert failed: %w, output: %s", err, output)
+		}
+		return nil
+	default:
+		return fmt.Errorf("unsupported VCS type: %v", repo.Vcs())
+	}
+}
+
 // Update updates a remote repository
 func (i *VCSInstaller) Update() error {
 	slog.Debug("updating", "source", i.Repo.Remote())
+
+	// Reset any local modifications in the cache directory before updating.
+	// The cached repository is managed by Helm and should not contain user modifications.
+	// Any modifications made by Helm itself (e.g., to plugin.yaml during installation)
+	// should be discarded before attempting to update.
 	if i.Repo.IsDirty() {
-		return errors.New("plugin repo was modified")
+		slog.Debug("resetting local modifications in cache", "path", i.Repo.LocalPath())
+		if err := resetRepo(i.Repo); err != nil {
+			return fmt.Errorf("failed to reset local modifications: %w", err)
+		}
 	}
+
 	if err := i.Repo.Update(); err != nil {
 		return err
 	}
--- a/internal/plugin/installer/vcs_installer.go
+++ b/internal/plugin/installer/vcs_installer.go
@@ -96,38 +96,41 @@ func (i *VCSInstaller) Install() error {
 	return fs.CopyDir(i.Repo.LocalPath(), i.Path())
 }
 
-// resetRepo discards all local modifications in the repository.
+// resetPluginYaml discards local modifications to plugin.yaml file.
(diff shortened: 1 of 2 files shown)


internal/plugin/installer/vcs_installer.go
```diff
@@ -95,12 +96,63 @@ func (i *VCSInstaller) Install() error {
 	return fs.CopyDir(i.Repo.LocalPath(), i.Path())
 }
 
+// resetRepo discards all local modifications in the repository.
+// This is used to clean the cached repository before updating.
+func resetRepo(repo vcs.Repo) error {
+	// Check the VCS type to determine the appropriate reset command
+	switch repo.Vcs() {
+	case vcs.Git:
```
TerryHowe, line 107: I can see how this will work, but ideally, vcs should add this I would think.

TerryHowe: Also, if plugin.yaml is the only problem, maybe just reset that one file.

OlTrenin: @TerryHowe hi, okey, I have improved by your comment's 😉

benoittgt:
I tested the change locally, and it does fix the failure. I’m not sure about resetting the repository during an update, but it seems like the simplest approach. I’m a bit worried we might miss something when resetting this cache folder.

I’m also wondering whether the reset logic should be extracted into a separate function so we can reuse it.

OlTrenin:
> I tested the change locally, and it does fix the failure. I’m not sure about resetting the repository during an update, but it seems like the simplest approach. I’m a bit worried we might miss something when resetting this cache folder.
> 
> I’m also wondering whether the reset logic should be extracted into a separate function so we can reuse it.

Thanks for the review! The concerns you raised have already been addressed in the second commit (98598c6):
  1. We now reset only plugin.yaml, not the entire repository
  2. The reset logic is already extracted into a separate resetPluginYaml() function
  3. plugin.yaml is the only file Helm modifies during installation, so there's nothing else to worry about missing

  The cache folder is managed by Helm internally, so resetting this specific file is safe

benoittgt:
My concerns stem from your last proposal. I’m wondering whether we still want to keep the change in 1e8ebae249ccf9cd1b9358c6b0c7780caf95f7b6. Another option could be to stop checking the dirty state altogether. 🤷

> The reset logic is already extracted into a separate resetPluginYaml() function

I was thinking about extracting the reset mechanism itself into a generic function that accepts a path and the VCS. So function does less and that way, we could easily reuse it to reset other files in other locations. That said, this might be premature optimization.

OlTrenin:
> I was thinking about extracting the reset mechanism itself into a generic function that accepts a path and the VCS. So function does less and that way, we could easily reuse it to reset other files in other locations. That said, this might be premature optimization.

Good point, but I agree it's premature optimization for now. We can always generalize it later if we need to reset other files. Keeping it simple for this PR or should I implement it?

benoittgt:
I'm curious about what @gjenkins8 is thinking about this PR.

PR #31679, opened 2025-12-27
Improve error message for invalid --set-json input

Vinaychinnu:
### What this PR does
Improves the error message shown when invalid JSON is passed to `--set-json`.

### Why this change is needed
The previous error messages were inconsistent and did not clearly indicate the cause of the failure.
This change:
- Safely quotes the invalid input
- Explicitly mentions that valid JSON is expected
- Keeps error messaging consistent across code paths

No functional behavior is changed.


banjoh (approved):
I'm not a fun of quoting logs, but its fine by me.

robertsirc:
I mean is this really needed?

Issue #31689, opened 2026-01-02
Lint Bash scripts with shellcheck


atombrella:
I experimented with [ShellCheck](https://www.shellcheck.net/)

There is an effort to fix various warnings and conform with the style guidelines in a local [feature branch](https://github.com/atombrella/helm/tree/feature/bash_lint_workflow)

The [workflow](https://github.com/atombrella/helm/actions/runs/20640906017/job/59272416176) runs smoothly, and the `ubuntu-latest` agent already has `shellcheck` installed.

The `sync-repo` script has been simplified a bit, and otherwise, I removed the legacy `[ "x$VAR" != "x" ]` trick (among other things) that is no longer needed. Is the `get` script really still needed?

I could open a PR if it is of interest. Unfortunately, the Slack channel requires an invite to join.

TerryHowe:
I don't know what all you found, but I imagine that would be worth a pull request. Thanks for looking into it.

sarfraj89:
@TerryHowe I plan to run ShellCheck on the existing bash scripts, fix the reported warnings (quoting, test expressions, deprecated patterns) without changing script logic, and ensure everything passes CI. I’ll open a PR once done.

moeritze:
I have created a PR to fix the shellcheck execution. Let me know what you think.

PR #31698, opened 2026-01-06
Lint sync-repo.sh with ShellCheck

sarfraj89:
This PR fixes ShellCheck warnings in scripts/sync-repo.sh by:
- Quoting variables to prevent word splitting
- Using read -r for safer input handling
- Correcting variable usage in case statement

No functional behavior is changed.

Related issue: #31689

### DIFF
--- a/scripts/sync-repo.sh
+++ b/scripts/sync-repo.sh
@@ -44,7 +44,8 @@ verify_prereqs() {
 }
 
 confirm() {
-  case $confirm in
+  # shellcheck disable=SC2154
+  case $response in
     [yY][eE][sS]|[yY])
       true
       ;;
--- a/scripts/sync-repo.sh
+++ b/scripts/sync-repo.sh
@@ -44,7 +44,7 @@ verify_prereqs() {
 }
 
 confirm() {
-  case $response in
+  case $confirm in
     [yY][eE][sS]|[yY])
       true
       ;;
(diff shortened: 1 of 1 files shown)


scripts/sync-repo.sh
```diff
@@ -44,7 +44,7 @@ verify_prereqs() {
 }
 
 confirm() {
-  case $response in
+  case $confirm in
```
banjoh: Why is this change needed?

banjoh:
Perhaps as a follow up PR, if you can add shellcheck linter GitHub action, that would be great

scripts/sync-repo.sh
sarfraj89: The variable response was never assigned in this script. User input is read into confirm (read -r ... confirm), so the case statement must evaluate confirm. This also resolves the ShellCheck SC2154 warning and fixes an actual logic bug rather than suppressing it.

sarfraj89:
> Perhaps as a follow up PR, if you can add shellcheck linter GitHub action, that would be great

Sounds good, thanks for the suggestion! I’m happy to follow up with a separate PR adding a ShellCheck GitHub Action once this one is merged, so we keep the changes focused.

scripts/sync-repo.sh
banjoh: If you are referring to `read -r -p...` in main, the `confirm` variable used is scoped within `main()` and is not the same one in `confirm()`. I don't think `confirm()` is used in this script, but might be used elsewhere in case this scripts is `source`d.

sarfraj89: You are right, thanks for pointing that out. The confirm() function does not receive input locally, and the confirm variable read in main() is scoped there. In that case, changing response to confirm inside confirm() is not correct.

To avoid altering behavior, I can revert this change and instead handle SC2154 by either initializing response locally in confirm() or suppressing the warning with a comment. Please let me know which approach you had prefer.

banjoh: Suppressing it with a comment would be my preference.

sarfraj89: Done, I have reverted the logic change and suppressed SC2154 with a targeted comment to avoid changing behavior.

banjoh (approved):

TerryHowe (approved):
/lgtm

sarfraj89:
> Perhaps as a follow up PR, if you can add shellcheck linter GitHub action, that would be great

follow-up pr opened to add the shellcheck GitHub action: #31704 . Thanks again for the suggestion!

PR #31701, opened 2026-01-07
fix: escape non-ASCII OCI annotation values

KyungHwanKim-devs:
Refs #31507.

This keeps OCI annotation values unchanged, but adds a targeted hint when a push fails with a manifest digest mismatch and the annotations contain non-ASCII. This should help users identify registries that rewrite annotation values without changing Helm behavior.

- Does **not** escape or modify annotation values.
- Adds a non-ASCII detection helper and only augments the error message for ErrMismatchedDigest cases.
- No behavior change on successful pushes or unrelated errors.

**If applicable**:

**Testing**:
- `go test ./pkg/registry -run TestHasNonASCIIAnnotationValues`

### DIFF
--- a/pkg/registry/chart.go
+++ b/pkg/registry/chart.go
@@ -18,8 +18,11 @@ package registry // import "helm.sh/helm/v4/pkg/registry"
 
 import (
 	"bytes"
+	"strconv"
 	"strings"
 	"time"
+	"unicode/utf16"
+	"unicode/utf8"
 
 	chart "helm.sh/helm/v4/pkg/chart/v2"
 	"helm.sh/helm/v4/pkg/chart/v2/loader"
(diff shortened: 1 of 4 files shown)


TerryHowe:
Dup of https://github.com/helm/helm/pull/31591 and I don't think that is going anywhere.

KyungHwanKim-devs:
Hi! I reviewed #31591 and the feedback about avoiding annotation mutation. Based on that, I minimized the changes here.

Summary of the update:
- Reverted the annotation escaping so values are left untouched.
- Added a small helper to detect non-ASCII keys/values and only when a push fails with ErrMismatchedDigest, provide a targeted hint that the registry may be rewriting annotations.
- No behavior changes on successful pushes or unrelated errors.

If you have time, could you please take one more look? If this direction is still not desired, I can close the PR right away.
Thank you!

PR #31704, opened 2026-01-08
Add ShellCheck Github Action for bash scripts

sarfraj89:
This PR adds a github actions workflow to run ShellCheck on Helm’s scripts directory.
It enforces shell linting in CI while allowing existing warnings to be addressed
incrementally in follow-up PRs.

### DIFF
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -18,7 +18,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout source code
-        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # pin@v4.0.1
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # pin@v6.0.1
         with:
           persist-credentials: false
 
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -17,8 +17,8 @@ jobs:
   shellcheck:
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout
-        uses: actions/checkout@v4
+      - name: Checkout source code
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # pin@v4.0.1
         with:
           persist-credentials: false
 
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -1,0 +1,28 @@
+name: ShellCheck
+
+on:
+  workflow_dispatch:
+  pull_request:
+  push:
+    branches:
+      - main
+    paths:
+      - '**.sh'
+      - '.github/workflows/shellcheck.yml'
+
+permissions:
+  contents: read
+
+jobs:
+  shellcheck:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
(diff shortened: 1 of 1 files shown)


sarfraj89:
The failing build-test looks unrelated to this PR and appears to be a flaky TestStatusWaitForDelete failure. Happy to re run if needed.

.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,28 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  shellcheck:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
```
banjoh: There is a newer v6 version

.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,28 @@
...
+
+      - name: Run ShellCheck
```
atombrella, line 25: Could you please just use the `shellcheck` command without this action? The action doesn't seem to updated regularly.

You can check the workflow file that I linked in the original issue :smile:

sarfraj89: Thanks for the suggestion! I have updated the workflow to run the shellcheck command directly, following the pattern you linked.

.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,26 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  shellcheck:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
```
gjenkins8:
```suggestion
      - name: Checkout source code
        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # pin@v6.0.1
```
we should pin (and use an updated version), same as other actions e.g.:

https://github.com/helm/helm/blob/1dffc817e5d5a70da7b98c2ebfaf990c74b87a73/.github/workflows/build-test.yml#L20

gjenkins8 (changes requested):
I'm not sure if shellcheck will be particularly useful here. Helm has limited scripts, and they don't change often.

OTOH, it doesn't hurt. And the action only runs on `scripts/**.sh` changes.

We should pin (as per comment). And also thanks for the contribution!

sarfraj89:
> I'm not sure if shellcheck will be particularly useful here. Helm has limited scripts, and they don't change often.
> 
> OTOH, it doesn't hurt. And the action only runs on `scripts/**.sh` changes.
> 
> We should pin (as per comment). And also thanks for the contribution!

Thank for the feedback, I have pinned actions/checkout to a specific commit SHA to match other workflows.

.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,26 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  shellcheck:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout source code
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # pin@v4.0.1
```
TerryHowe:
```suggestion
        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # pin@v6.0.1
```

TerryHowe (changes requested):

PR #31706, opened 2026-01-09
pkg/kube: introduce support for custom kstatus readers

matheuscscp:

Closes: #31705

Edit: See comment below

**If applicable**:

### DIFF
--- a/pkg/kube/client.go
+++ b/pkg/kube/client.go
@@ -87,6 +87,8 @@ type Client struct {
 	// WaitContext is an optional context to use for wait operations.
 	// If not set, a context will be created internally using the
 	// timeout provided to the wait functions.
+	//
+	// Deprecated: Use WithWaitContext wait option when getting a Waiter instead.
 	WaitContext context.Context
 
 	Waiter
@@ -139,7 +141,11 @@ func init() {
 	}
 }
 
-func (c *Client) newStatusWatcher() (*statusWaiter, error) {
+func (c *Client) newStatusWatcher(opts ...WaitOption) (*statusWaiter, error) {
+	var o waitOptions
+	for _, opt := range opts {
+		opt(&o)
+	}
 	cfg, err := c.Factory.ToRESTConfig()
 	if err != nil {
 		return nil, err
@@ -156,14 +162,23 @@ func (c *Client) newStatusWatcher() (*statusWaiter, error) {
 	if err != nil {
 		return nil, err
 	}
+	waitContext := o.ctx
+	if waitContext == nil {
+		waitContext = c.WaitContext
+	}
 	return &statusWaiter{
 		restMapper: restMapper,
 		client:     dynamicClient,
-		ctx:        c.WaitContext,
+		ctx:        waitContext,
+		readers:    o.statusReaders,
 	}, nil
 }
 
-func (c *Client) GetWaiter(strategy WaitStrategy) (Waiter, error) {
+func (c *Client) GetWaiter(ws WaitStrategy) (Waiter, error) {
+	return c.GetWaiterWithOptions(ws)
+}
+
+func (c *Client) GetWaiterWithOptions(strategy WaitStrategy, opts ...WaitOption) (Waiter, error) {
 	switch strategy {
 	case LegacyStrategy:
 		kc, err := c.Factory.KubernetesClientSet()
@@ -172,9 +187,9 @@ func (c *Client) GetWaiter(strategy WaitStrategy) (Waiter, error) {
 		}
 		return &legacyWaiter{kubeClient: kc, ctx: c.WaitContext}, nil
 	case StatusWatcherStrategy:
-		return c.newStatusWatcher()
+		return c.newStatusWatcher(opts...)
 	case HookOnlyStrategy:
-		sw, err := c.newStatusWatcher()
+		sw, err := c.newStatusWatcher(opts...)
 		if err != nil {
 			return nil, err
 		}
@@ -187,8 +202,12 @@ func (c *Client) GetWaiter(strategy WaitStrategy) (Waiter, error) {
 }
 
 func (c *Client) SetWaiter(ws WaitStrategy) error {
+	return c.SetWaiterWithOptions(ws)
+}
+
+func (c *Client) SetWaiterWithOptions(ws WaitStrategy, opts ...WaitOption) error {
 	var err error
-	c.Waiter, err = c.GetWaiter(ws)
+	c.Waiter, err = c.GetWaiterWithOptions(ws, opts...)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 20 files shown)


matheuscscp:
@AustinAbro321 PTAL :pray:

AustinAbro321 (approved):
LGTM, simple and well tested

pkg/kube/client.go
```diff
@@ -87,6 +87,8 @@ type Client struct {
 	// WaitContext is an optional context to use for wait operations.
 	// If not set, a context will be created internally using the
 	// timeout provided to the wait functions.
+	//
+	// Deprecated: Use WithWaitContext wait option when getting a Waiter instead.
```
matheuscscp, line 91: I introduced `WaitContext` recently just before v4.0.0 was cut. I have deprecated this field on behalf of `WaitOption`, which is more flexible as discussed above, and to unify wait options under a single interface.

Technically, removing `WaitContext` now would be a breaking change, but I really doubt any other SDK users are using it already. If you are OK with removing it, just let me know and I can update the PR!

matheuscscp (commented):
@scottrigby As discussed via Slack, I have updated the changes here to pass this new input from `pkg/action` via `GetWaiter`. This is more flexible, since an SDK user can create a single `pkg/kube.Client` and use it with different actions with different wait behaviors (the previous alternative would need one `pkg/kube.Client` to be created for each action to achieve the same).

With regards to `type StatusReader = engine.StatusReader` it turns out that it does not bring any benefit because the `engine.StatusReader` interface exposes other types defined in the `engine` package. That's a bummer, otherwise we could do this and expose only a type defined in the Helm SDK :/ The good news is that this type is now exposed in a minor interface of the SDK, the functional option `WithKStatusReaders()` (better than a field of the `pkg/kube.Client`).

0xsarawut:
> **What this PR does / why we need it**:
> 
> Closes: #31705
> 
> **Special notes for your reviewer**:
> 
> Edit: See comment below
> 
> **If applicable**:
> - [ ] this PR contains user facing changes (the `docs needed` label should be applied if so)
> - [x] this PR contains unit tests
> - [x] this PR has been tested for backwards compatibility
>

0xsarawut (approved):

pkg/kube/client.go
scottrigby: No, you did this perfectly. the correct way according to our strict backwards compatibility contract. Thank you!

pkg/kube/options.go
```diff
@@ -0,0 +1,45 @@
...
+
+// WithWaitContext sets the context for waiting on resources.
+func WithWaitContext(ctx context.Context) WaitOption {
+	return func(wo *waitOptions) {
+		wo.ctx = ctx
+	}
+}
+
+// WithKStatusReaders sets the status readers to be used while waiting on resources.
+func WithKStatusReaders(readers ...engine.StatusReader) WaitOption {
```
scottrigby, line 36: I was initially concerned about exposing these more generally, but this is the correct place for it. When choosing to call WithKStatusReaders the SDK user will always be choosing kstatus 💯 👏

scottrigby (dismissed):
Nice work 👏 

> @scottrigby As discussed via Slack, I have updated the changes here to pass this new input from `pkg/action` via `GetWaiter`. This is more flexible, since an SDK user can create a single `pkg/kube.Client` and use it with different actions with different wait behaviors (the previous alternative would need one `pkg/kube.Client` to be created for each action to achieve the same).
> 
> With regards to `type StatusReader = engine.StatusReader` it turns out that it does not bring any benefit because the `engine.StatusReader` interface exposes other types defined in the `engine` package. That's a bummer, otherwise we could do this and expose only a type defined in the Helm SDK :/ The good news is that this type is now exposed in a minor interface of the SDK, the functional option `WithKStatusReaders()` (better than a field of the `pkg/kube.Client`).

Thank you, the rework addresses my earlier concerns about adding kstatus object directly to the Client. Passing wait options (that include the reader) through the actions down to Client.GetWaiter as you are doing now is so much better 👏 

Note to other reviewers: this feature would need to be merged before we cut the RC (Monday Jan 12) if we want it in v4.1.0. Otherwise it would need to wait until the next minor release.

matheuscscp:
> Note to other reviewers: this feature would need to be merged before we cut the RC (Monday Jan 12) if we want it in v4.1.0. Otherwise it would need to wait until the next minor release.

I'd really love to see this in 4.1, Flux 2.8 schedule is based on Helm 4.1. Otherwise we can only get this in Flux half a year from now :cry:

AustinAbro321 (approved):

pkg/kube/interface.go
```diff
@@ -57,7 +57,7 @@ type Interface interface {
 	IsReachable() error
 
 	// Get Waiter gets the Kube.Waiter
-	GetWaiter(ws WaitStrategy) (Waiter, error)
+	GetWaiter(ws WaitStrategy, opts ...WaitOption) (Waiter, error)
```
banjoh: This is a breaking change. Create a new interface that defines this new method. See how this pattern is used in helm v3 https://github.com/helm/helm/blob/dev-v3/pkg/kube/interface.go#L78

banjoh (changes requested):

pkg/kube/statuswait_test.go
```diff
@@ -17,12 +17,18 @@ limitations under the License.
 package kube // import "helm.sh/helm/v3/pkg/kube"
```
banjoh:
```suggestion
package kube // import "helm.sh/helm/v4/pkg/kube"
```

pkg/kube/interface.go
matheuscscp: Fixed, thanks!

pkg/kube/statuswait.go
```diff
@@ -105,7 +105,8 @@ func (w *statusWaiter) WaitWithJobs(resourceList ResourceList, timeout time.Dura
 	slog.Debug("waiting for resources", "count", len(resourceList), "timeout", timeout)
 	sw := watcher.NewDefaultStatusWatcher(w.client, w.restMapper)
 	newCustomJobStatusReader := helmStatusReaders.NewCustomJobStatusReader(w.restMapper)
-	customSR := statusreaders.NewStatusReader(w.restMapper, newCustomJobStatusReader)
+	readers := append(w.readers, newCustomJobStatusReader)
```
banjoh: Nit pick: This is a rare one, but if `WaitWithJobs` is called multiple times, and the underlying array of `readers` slice has capacity, `newCustomJobStatusReader` object will replace the previous one

```suggestion
	readers := append([]engine.StatusReader(nil), w.readers...)
	readers = append(readers, newCustomJobStatusReader)
```

scottrigby (dismissed):

banjoh (dismissed):
Changes look good

banjoh (approved):

scottrigby (approved):

PR #31709, opened 2026-01-09
feat(release): add internal/release/v2 package for chart v3 support

banjoh:

Introduce `internal/release/v2` that mirrors `pkg/release/v1` but uses `*v3.Chart` from `internal/chart/v3`. The code is structurally identical to v1 with only import paths changed to reference `internal/chart/v3` instead of `pkg/chart/v2`.

- Add internal/release/v2 with Release, Info, Hook types
- Add internal/release/v2/util with filter, sorter, manifest utilities
- Update pkg/release/common.go with v2Accessor and v2HookAccessor
- Copy all test files from pkg/release/v1 and add a v2 test in common_test.go

**If applicable**:

### DIFF
--- a/internal/release/v2/doc.go
+++ b/internal/release/v2/doc.go
@@ -1,0 +1,17 @@
+/*
+Copyright The Helm Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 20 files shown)


scottrigby (approved):
> The code is structurally identical to v1 with only import paths changed to reference `internal/chart/v3` instead of `pkg/chart/v2`

confirmed with `diff -u --color pkg/release/v1 internal/release/v2`. 

also nice tests. LGTM :+1:

joejulian (approved):

TerryHowe (approved):
/lgtm

PR #31535, opened 2025-11-19
bump version to 4.1

mattfarina:
**What this PR does / why we need it**: In a recent Helm developer call we decided to bump the internal version on main to the next minor after the release branch was cut for a new minor. This is instead of bumping the minor on the release branch and then cherry-picking it to main.

This PR bumps main to 4.1 which is the next minor.

**Special notes for your reviewer**:

**If applicable**:
- [ ] this PR contains user facing changes (the `docs needed` label should be applied if so)
- [ ] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

PR #31718, opened 2026-01-12
bump version to v3.20

scottrigby:
Prep for v3.20. Not yet documented, we agreed in helm dev meeting this should happen in the major version branches (main and dev-v3) before creating the minor release branch, rather than the currently documented cherry-picking method.

See https://github.com/helm/helm/pull/31535 for 4.1 counterpart. Minor versions of Helm 4 and 3 are released together.


TerryHowe (approved):
/lgtm

PR #31721, opened 2026-01-13
feat(logging): re-introduce logging API

matheuscscp:
**What this PR does / why we need it**:

Exposes the interface SDK users can use to pass logger implementations to various parts of the SDK (only storage uses this at the moment).

**Special notes for your reviewer**:

@banjoh Follow up to https://github.com/helm/helm/pull/31411/files#r2683864962

**If applicable**:
- [ ] this PR contains user facing changes (the `docs needed` label should be applied if so)
- [x] this PR contains unit tests
- [x] this PR has been tested for backwards compatibility

PR #31723, opened 2026-01-13
bugfix(logging): fix storage not getting logger from driver

matheuscscp:

Fixes the storage not getting the logger from the driver.

**If applicable**:

### DIFF
--- a/pkg/storage/storage.go
+++ b/pkg/storage/storage.go
@@ -339,12 +339,14 @@ func Init(d driver.Driver) *Storage {
 		Driver: d,
 	}
 
+	var h slog.Handler
 	// Get logger from driver if it implements the LoggerSetterGetter interface
 	if ls, ok := d.(logging.LoggerSetterGetter); ok {
-		ls.SetLogger(s.Logger().Handler())
+		h = ls.Logger().Handler()
 	} else {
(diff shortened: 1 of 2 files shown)


matheuscscp:
Supersedes #31721

pkg/logging/logging.go
```diff
@@ -0,0 +1,27 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package logging
+
+import "log/slog"
+
+// LoggerSetterGetter is an interface that can set and get a logger
```
banjoh: Why are you exposing this interface?

From https://github.com/helm/helm/pull/31411/changes#r2683864962, you say "There also used to be a method for passing a logger to the pkg/storage.Storage.". This hasn't changed. You can do this

```go
package main

import (
	"helm.sh/helm/v4/pkg/storage"
)

func main() {
	s := storage.Init(nil)
	s.SetLogger(nil)
}
```

matheuscscp: You're right, thanks!

banjoh (commented):
You have some linter warnings

banjoh (approved):

gjenkins8 (approved):

PR #31726, opened 2026-01-14
chore: clarify --wait flag help text

banjoh:

Improve the description to explain the three usage scenarios:
- --wait alone defaults to 'watcher' strategy
- --wait=<value> uses the specified strategy
- flag omitted defaults to 'hookOnly'

**If applicable**:


TerryHowe (approved):
/lgtm

scottrigby (approved):
I think this help text makes sense and clarifies confusion between default when the flag is passed with no values vs default when the flag is not passed.

PR #31730, opened 2026-01-14
bugfix(kstatus): do not wait forever on failed resources

matheuscscp:

Fixes: #31729

In kstatus `Failed` is a terminal state. There's no point in making Helm wait on it forever, as it will never transition to `Current` without intervention.

**If applicable**:

### DIFF
--- a/pkg/kube/client_test.go
+++ b/pkg/kube/client_test.go
@@ -32,9 +32,11 @@ import (
 	"github.com/fluxcd/cli-utils/pkg/kstatus/polling/event"
 	"github.com/fluxcd/cli-utils/pkg/kstatus/status"
 	"github.com/fluxcd/cli-utils/pkg/object"
+	"github.com/fluxcd/cli-utils/pkg/testutil"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
+	batchv1 "k8s.io/api/batch/v1"
 	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
@@ -44,8 +46,10 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	jsonserializer "k8s.io/apimachinery/pkg/runtime/serializer/json"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/yaml"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
 	"k8s.io/cli-runtime/pkg/resource"
+	dynamicfake "k8s.io/client-go/dynamic/fake"
 	"k8s.io/client-go/kubernetes"
 	k8sfake "k8s.io/client-go/kubernetes/fake"
 	"k8s.io/client-go/kubernetes/scheme"
@@ -1241,7 +1245,7 @@ func (c createPatchTestCase) run(t *testing.T) {
 	}
 }
 
-func newTestCustomResourceData(metadata map[string]string, spec map[string]interface{}) *unstructured.Unstructured {
+func newTestCustomResourceData(metadata map[string]string, spec map[string]any) *unstructured.Unstructured {
 	if metadata == nil {
 		metadata = make(map[string]string)
 	}
@@ -1251,7 +1255,7 @@ func newTestCustomResourceData(metadata map[string]string, spec map[string]inter
 	if _, ok := metadata["namespace"]; !ok {
 		metadata["namespace"] = "default"
 	}
-	o := map[string]interface{}{
+	o := map[string]any{
 		"apiVersion": "crd.com/v1",
 		"kind":       "Data",
 		"metadata":   metadata,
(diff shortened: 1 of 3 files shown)


matheuscscp:
I have run the entire test suite of Flux's helm-controller with this fix and it passed :+1: 

https://github.com/fluxcd/helm-controller/actions/runs/21015460312/job/60419627508

AustinAbro321 (approved):
This is a nice improvement. Thanks

pkg/kube/statuswait.go
```diff
@@ -145,20 +145,26 @@ func (w *statusWaiter) waitForDelete(ctx context.Context, resourceList ResourceL
...
+		errs = append(errs, err)
+	}
+	switch len(errs) {
+	case 0:
+		return nil
+	case 1:
+		return errs[0]
+	default:
 		return errors.Join(errs...)
 	}
```
gjenkins8: shouldn't `return errors.Join(errs...)` suffice here? Do we really want/need to return a non-wrapped error if there is only one?

pkg/kube/statuswait.go
```diff
@@ -190,20 +196,26 @@ func (w *statusWaiter) wait(ctx context.Context, resourceList ResourceList, sw w
...
+			continue
 		}
-		errs = append(errs, ctx.Err())
+		errs = append(errs, fmt.Errorf("resource %s/%s/%s not ready. status: %s, message: %s",
+			rs.Identifier.GroupKind.Kind, rs.Identifier.Namespace, rs.Identifier.Name, rs.Status, rs.Message))
+	}
+	if err := ctx.Err(); err != nil {
+		errs = append(errs, err)
+	}
+	switch len(errs) {
```
gjenkins8: similar comment as above

pkg/kube/statuswait_test.go
```diff
@@ -279,19 +296,19 @@ func TestStatusWaitForDelete(t *testing.T) {
 		name              string
 		manifestsToCreate []string
 		manifestsToDelete []string
-		expectErrs        []error
+		expectErrStrs     []string
```
gjenkins8:
```suggestion
		expectErrs     []string
```

Perhaps just leave this name as is? (the type suffices to me). Either way works for me.

pkg/kube/client_test.go
```diff
@@ -2275,9 +2279,27 @@ metadata:
...
 	}
 
-	var err error
-	c.Waiter, err = c.GetWaiterWithOptions(StatusWatcherStrategy, WithKStatusReaders(statusReaders...))
+	// Create a fake dynamic client with the pod resource
+	fakeClient := dynamicfake.NewSimpleDynamicClient(scheme.Scheme)
+	fakeMapper := testutil.NewFakeRESTMapper(v1.SchemeGroupVersion.WithKind("Pod"))
+
+	// Create the pod in the fake client
+	m := make(map[string]interface{})
```
gjenkins8:
```suggestion
	m := make(map[string]any)
```

pkg/kube/client_test.go
```diff
@@ -2307,9 +2329,27 @@ metadata:
 		},
 	}
 
-	var err error
-	c.Waiter, err = c.GetWaiterWithOptions(StatusWatcherStrategy, WithKStatusReaders(statusReaders...))
+	// Create a fake dynamic client with the job resource
+	fakeClient := dynamicfake.NewSimpleDynamicClient(scheme.Scheme)
+	fakeMapper := testutil.NewFakeRESTMapper(batchv1.SchemeGroupVersion.WithKind("Job"))
+
+	// Create the job in the fake client
```
gjenkins8, line 2328: Perhaps this code could be factored into a helper? (it is nearly identical to the above)

gjenkins8 (commented):
Some comments, nothing major

pkg/kube/statuswait.go
matheuscscp: This usually cleans up garbage in the error message and makes it easier to read. But I don't feel strongly about this. Fixed, thanks!

pkg/kube/statuswait.go
matheuscscp: This usually cleans up garbage in the error message and makes it easier to read. But I don't feel strongly about this. Fixed, thanks!

pkg/kube/statuswait_test.go
matheuscscp: Fixed, thanks!

pkg/kube/client_test.go
matheuscscp: Fixed, thanks!

pkg/kube/client_test.go
matheuscscp: Fixed, thanks!

matheuscscp:
@gjenkins8 All fixed, pls take another look :pray:

TerryHowe (approved):
/lgtm

banjoh (approved):
LGTM

scottrigby (approved):
Really good catch. I'm a little surprised no one noticed this before now, since it's been here since v4.0.0.

Noksa:
@scottrigby can we have it in upcoming `4.1.0`?

PR #31732, opened 2026-01-15
fix(defaults): server-side apply SDK defaults should always match the CLI defaults

matheuscscp:

The SDK server-side apply defaults must always match the CLI defaults.

**If applicable**:


benoittgt (commented):
Thanks ! Good catch.

gjenkins8 (approved):

PR #31733, opened 2026-01-15
fix(engine): handle error when accessing dependency charts

Ekene-Chris:
This PR fixes a potential nil pointer panic in the template engine when processing chart dependencies.

In pkg/engine/engine.go, the recAllTpls function had a // TODO: Handle error comment where the error from ci.NewAccessor(child) was being silently ignored. If NewAccessor fails (e.g., for an unsupported chart type), it returns nil, and the subsequent call to sub.Name() would panic.

This change adds proper error handling consistent with the pattern already used earlier in the same function (lines 541-544), logging the error and skipping the problematic dependency instead of crashing.

This addresses an explicit TODO comment in the codebase
The error handling pattern matches the existing code style in the same function
No behavioral change for the happy path; only affects error cases

**If applicable**:

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -563,8 +563,11 @@ func recAllTpls(c ci.Charter, templates map[string]renderable, values common.Val
 	}
 
 	for _, child := range accessor.Dependencies() {
-		// TODO: Handle error
-		sub, _ := ci.NewAccessor(child)
+		sub, err := ci.NewAccessor(child)
+		if err != nil {
+			slog.Error("error accessing dependency chart", slog.Any("error", err))
+			continue
+		}
 		subCharts[sub.Name()] = recAllTpls(child, templates, next)
 	}
 


pkg/engine/engine.go
```diff
@@ -563,8 +563,11 @@ func recAllTpls(c ci.Charter, templates map[string]renderable, values common.Val
 	}
 
 	for _, child := range accessor.Dependencies() {
-		// TODO: Handle error
-		sub, _ := ci.NewAccessor(child)
+		sub, err := ci.NewAccessor(child)
+		if err != nil {
+			slog.Error("error accessing dependency chart", "error", err)
```
benoittgt: Do you think you can switch to `slog.Any` ? This is something we see often in [Helm codebase](https://github.com/search?q=repo%3Ahelm%2Fhelm%20slog.Any("error"%2C%20err)&type=code).

Ekene-Chris: Done, updated to use slog.Any for consistency with the rest of the codebase.

benoittgt (approved):
LGTM

robertsirc (changes requested):
First thank you for your PR. This is a solid change but this isn't covered by testing is there a way we can get this tested?

PR #31740, opened 2026-01-17
feat: New --values-tpl flag with limited template support

atorrescogollo:
Use --values-tpl to dynamically generate values based on previously merged configuration. Templates have access to .Values with basic template functions.

This is especially useful when you don't have control over upstream chart templates and want to avoid forking just to add tpl support.

## How it works

Templates are rendered after all other value sources (--values, --set) and have access to the merged .Values context:

```bash
# Basic usage
helm install myapp ./chart --values-tpl dynamic.yaml.tpl

# With values from previous sources
helm install myapp ./chart \
  --values base.yaml \
  --set environment=production \
  --values-tpl computed.yaml.tpl
```

```yaml
# base.yaml
region: us-east-1
environment: dev

# computed.yaml.tpl
replicas: {{ if eq .Values.environment "production" }}5{{ else }}1{{ end }}
resourceTier: {{ .Values.environment | upper }}
endpoint: "https://{{ .Values.region }}.example.com"
```

## Value merge order

1. Chart defaults (`values.yaml`)
2. All `--values` files
3. All `--set*` flags
4. All `--values-tpl` files (rendered with access to steps 1-3)

## Real-world use case: Dynamic ECR registry per cluster with ArgoCD

**Problem:** Each cluster is in a different AWS region/account, and you need different ECR registries. You want to:
- Avoid committing per-cluster files (values-prod-us.yaml, values-prod-eu.yaml, etc.)
- Not edit upstream chart templates to add tpl support
- Keep ArgoCD unaware of chart-specific paths (image.registry, etc.)

**Solution:** Single template file in git, cluster-specific values via ArgoCD:

```yaml
# values-custom.yaml.tpl (committed to repo, works for all clusters)
image:
  registry: "{{ .Values.accountId }}.dkr.ecr.{{ .Values.region }}.amazonaws.com"
  repository: myapp
  tag: "1.0.0"

service:
  extraEnvironment:  # Arbitrary path in values
    CLUSTER_NAME: "{{ .Values.clusterName }}"
```

```yaml
# ArgoCD Application for prod-us-east cluster
# Note: Requires ArgoCD support for valueTplFiles (pending PR)
spec:
  helm:
    parameters:
    - name: clusterName
      value: prod-us-east
    - name: region
      value: us-east-1
    - name: accountId
      value: "123456789"
    valueTplFiles:
    - values-custom.yaml.tpl
```

```yaml
# ArgoCD Application for prod-eu-west cluster (same template!)
spec:
  helm:
    parameters:
    - name: clusterName
      value: prod-eu-west
    - name: region
      value: eu-west-1
    - name: accountId
      value: "987654321"
    valueTplFiles:
    - values-custom.yaml.tpl
```

## Supported in templates

- Access to `.Values` from previous merge steps
- Sprig functions: `default`, `upper`, `lower`, `trim`, etc.
- Format functions: `toYaml`, `toJson`, `toToml`, `fromYaml`, `fromJson`, `fromToml`
- Conditionals: `if`/`else`, `range`
- Remote templates: `http://`, `https://`, etc.

## Not supported (chart-only features)

- `include` (requires chart template context)
- `tpl` (requires chart template context)
- `lookup` (requires Kubernetes client)

**If applicable**:

### DIFF
--- a/pkg/cli/values/options.go
+++ b/pkg/cli/values/options.go
@@ -26,18 +26,20 @@ import (
 	"strings"
 
 	"helm.sh/helm/v4/pkg/chart/v2/loader"
+	"helm.sh/helm/v4/pkg/engine"
 	"helm.sh/helm/v4/pkg/getter"
 	"helm.sh/helm/v4/pkg/strvals"
 )
 
 // Options captures the different ways to specify values
 type Options struct {
-	ValueFiles    []string // -f/--values
-	StringValues  []string // --set-string
-	Values        []string // --set
-	FileValues    []string // --set-file
-	JSONValues    []string // --set-json
-	LiteralValues []string // --set-literal
+	ValueFiles     []string // -f/--values
+	ValueTemplates []string // --values-tpl
+	StringValues   []string // --set-string
+	Values         []string // --set
+	FileValues     []string // --set-file
+	JSONValues     []string // --set-json
+	LiteralValues  []string // --set-literal
 }
 
 // MergeValues merges values from files specified via -f/--values and directly
@@ -112,6 +114,19 @@ func (opts *Options) MergeValues(p getter.Providers) (map[string]interface{}, er
 		}
 	}
 
+	// User specified a values template via --values-tpl
+	for _, filePath := range opts.ValueTemplates {
+		raw, err := readFile(filePath, p)
+		if err != nil {
+			return nil, err
+		}
+		rendered, err := LoadValuesTpl(filePath, string(raw), base)
+		if err != nil {
+			return nil, fmt.Errorf("failed to render values template %s: %w", filePath, err)
+		}
+		base = loader.MergeMaps(base, rendered)
+	}
+
 	return base, nil
 }
 
@@ -135,4 +150,32 @@ func readFile(filePath string, p getter.Providers) ([]byte, error) {
 		return nil, err
 	}
 	return data.Bytes(), nil
+}
+
+// LoadValuesTpl renders a values file as a Go template with .Values context
+func LoadValuesTpl(
+	templatePath string,
+	templateData string,
+	currentValues map[string]interface{},
+) (map[string]interface{}, error) {
+	// Create template context with .Values
+	context := map[string]interface{}{
+		"Values": currentValues,
+	}
+
+	// Render using the engine
+	eng := engine.Engine{}
+	rendered, err := eng.RenderString(templatePath, templateData, context)
+	if err != nil {
+		return nil, fmt.Errorf("failed to render template %s: %w", templatePath, err)
+	}
+
+	// Parse the rendered YAML output
+	buf := bytes.NewBufferString(rendered)
+	result, err := loader.LoadValues(buf)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse rendered template %s as YAML: %w", templatePath, err)
+	}
+
+	return result, nil
 }
--- a/pkg/cli/values/options_test.go
+++ b/pkg/cli/values/options_test.go
@@ -387,3 +387,284 @@ func TestMergeValuesCLI(t *testing.T) {
 		})
 	}
 }
+
+func TestLoadValuesTpl(t *testing.T) {
+	tests := []struct {
+		name          string
+		templateData  string
+		currentValues map[string]interface{}
+		expected      map[string]interface{}
+		wantErr       bool
+		errContains   string
+	}{
+		{
+			name: "basic template with .Values access",
+			templateData: `
+environment: {{ .Values.env }}
+replicas: {{ .Values.replicas }}
+`,
+			currentValues: map[string]interface{}{
+				"env":      "production",
+				"replicas": float64(5),
+			},
+			expected: map[string]interface{}{
+				"environment": "production",
+				"replicas":    float64(5),
+			},
+			wantErr: false,
+		},
+		{
+			name: "template with conditional logic",
+			templateData: `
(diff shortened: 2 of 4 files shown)


atorrescogollo:
Hi @gjenkins8 @TerryHowe , could I get some feedback about this? Thanks

atorrescogollo:
Still waiting for feedback

joejulian:
Thanks for the detailed writeup. The ECR-per-cluster case is a real pain and I get why you reached for this.

I'm going to close this one, and I want to be upfront that it isn't about the implementation. Templating `values.yaml` was proposed and decided against in #2492 ("Proposal: Allow templating in values.yaml"), and the reasoning there still holds. The short version, from that thread:

- Values must be valid YAML in their unparsed state, and they are merged in stages, some of which (`--set` resolution, `-f` merges, dependency constraints) run before the template subsystem is even available. Matt Butcher walked through exactly that here: https://github.com/helm/helm/issues/2492#issuecomment-304403347
- The standing decision was to not pre-process `values.yaml` as a template, and to point people at wrappers or plugins instead (gomplate was the example called out): https://github.com/helm/helm/issues/2492#issuecomment-413027879

Since then, values gained `values.schema.json` validation, which templated values would collide with (is the schema validating pre-render or post-render?). So the "keep values as data" position is, if anything, stronger now than it was in 2018.

The good news is your use case does not require forking the upstream chart. The supported ways to compute values dynamically all live just outside Helm:

- Generate the values in your pipeline (ArgoCD supports this via a config-management plugin, or a small pre-step) and hand the finished file to Helm.
- Pre-process with a wrapper or Helm plugin, or a tool like gomplate/envsubst, before the install. This is exactly what #2492 recommends.
- For shaping the rendered output, a post-renderer.

For the ECR example specifically, computing `image.registry` from the cluster's account and region in an ArgoCD config-management plugin or a small wrapper gets you the same result without adding a new value source to core.

If you think this should be revisited as a core capability despite #2492, the place to make that case is a Helm Improvement Proposal, since it would be a permanent change to how values work. But the standing position is to keep values as data and let this computation happen before Helm sees them.

Closing on that basis, and pointing at #2492 for the full history. Thanks again for the thorough PR and the real-world context.

atorrescogollo:
@joejulian Thanks for the detailed response. I think a few things here differ from the premises and the conclusion of #2492, and I'd like to know if they were taken into account:

### "Values must be valid YAML" and "not pre-process values.yaml as a template"

This PR doesn't change values or their behaviour, it adds a new concept: values-tpl. It's not YAML on purpose, but neither is `--set`. It's actually much closer to `--set` than to values: it's not a YAML document, it's something that produces values, and it always applies at the same point in the merge no matter where you put it in the args. It's a new optional layer, not a change to the existing one.

### "run before the template subsystem is even available"

That's why values-tpl doesn't aim to work as a chart template. It only has .Values available, no .Release, no .Chart, etc. The example in #2492 was `{{ .Release.Name }}.example.com` inside values.yaml, and that's where the ordering problem actually comes from: you'd need the release before you can compute the values. values-tpl never gives you that, so it can't get into that loop.

I think it's reasonable to have that as a built-in limitation. It also stops people from overengineering with values-tpl: it can only compute from values you already passed in, not from the release, the chart or the cluster, so there's much less room for the edge cases #2492 was worried about.

### "values gained values.schema.json validation, which templated values would collide with (is the schema validating pre-render or post-render?)"

It's post-render, always. A values-tpl file isn't data, it's a template, so there's nothing to validate before rendering. Whatever it renders gets merged into values before validation runs, same as `-f` and `--set`. So nothing changes for the schema. I don't see the collision honestly.

### "ArgoCD supports this via a config-management plugin"

Using a config-management plugin forces you to abandon the standard way of doing Helm with ArgoCD. You have to configure all the Helm commands from scratch and the Application manifest ends up completely different. That's fine when your tooling really is different, but I don't think that's the case here. For something like computing a registry URL, a CMP is probably overengineered.

### "Pre-process with a wrapper or Helm plugin, or a tool like gomplate/envsubst, before the install"

I agree that's the reasonable workaround today. But it's still a workaround IMO, and for a use case that shows up as soon as you have two clusters: one for dev and one for prod. My ECR example makes it look like a multi-region thing, but you hit this the moment the same chart needs different values per cluster.

I think Helm not providing a built-in way of doing this pushes people into bad patterns. For example:

**"Should the app-of-apps set the cluster-dependent values directly in the ArgoCD Application manifest?"**

That's a bad idea IMO, you end up with a values split brain. Especially painful on upgrades. And underneath it's a consequence of not having values-tpl: "I need templating in values.yaml, Helm won't give me that, so ArgoCD is the only place left to do it".

**"Committing the generated file is annoying, should we just run gomplate on the fly?"**

Committing is annoying, but running a custom templating tool on the fly is worse IMO. Not because of the tool, but because you're gluing two tools together with a script that probably isn't well tested. The day one of them makes a breaking change and you miss it, none of your apps will render. If you don't want that trouble, you're better off living with the commit.

All in all, I strongly think this is something Helm should give a solution to. I would also reject templating values.yaml itself, for the ordering and values-as-data reasons, but that's not what this PR does.

So my real question is whether you and the other maintainers would be open to discussing this again from a different angle: not templating the existing values, but adding a separate layer that leaves values untouched. That was actually [suggested in #2492](https://github.com/helm/helm/issues/2492#issuecomment-345532083) and never discussed on its own terms. Both the technical objections and the comment that closed the issue are about values.yaml itself. If it's worth revisiting, I'm happy to write a HIP so we can go through the corner cases properly.

Issue #30989, opened 2025-06-19
Add `gzip` / `gunzip` template functions


cjgibson:
### What would you like to be added?

Helm template functions for compressing string values via `gzip`, and optionally to uncompress them by the same.

### Why is this needed?

I was looking for a way to populate this field in `GrafanaDashboard` CRs:

```bash
$ kubectl explain GrafanaDashboard.spec.gzipJson
GROUP:      grafana.integreatly.org
KIND:       GrafanaDashboard
VERSION:    v1beta1

FIELD: gzipJson <string>


DESCRIPTION:
    GzipJson the model's JSON compressed with Gzip. Base64-encoded when in YAML.
```

It looks like while there isn't an existing function to compress arbitrary string templates today, there's a straightforward PR in the [upstream `sprig` repository to add them](https://github.com/Masterminds/sprig/pull/379) that's gone unmerged for a couple years now. Curious if there would be interest from the maintainers in adding these helper functions specifically to Helm if they're unwelcome as part of `sprig` proper? Would be happy to get a PR up instead of transforming all my `.Files.Glob` invocations into some shell-expansion like:

```bash
find dashboards/ -iname '*.json' -print0 \
| while read -r -d '' dashboard; do 
    printf \
      -- "--set 'dashboardContent[\"%s\"]=%s' " \
      "${dashboard}" \
      "$(gzip -9 -c <"${dashboard}" | base64)"
  done
```

adharsh277:
Can i work on this? @cjgibson

cjgibson:
> Can i work on this? [@cjgibson](https://github.com/cjgibson)

Oh, I'm not a contributor to this project; to be honest my understanding is that the venn diagram for Helm and Sprig authors is effectively a circle 😅 

I mostly created this issue to get the attention of the Helm contributors (@mattfarina in particular), pushing things along and offering to lend a hand if they'd be accepting of such a change. I'm not sure if the repository leads here would prefer it in Helm, in Sprig, or not at all.

gjenkins8:
Can I suggest `compress <type> <data>` / `decompress <type> <data>` functions, where `<type>` would be "gzip" and other  compression algorithm e.g. zstd, lz4, bzip2, xz, etc could be supported in the future. Instead of gzip specific functions.

kvaps:
Adding a concrete use-case from the Cluster API ecosystem:

[...]

PR #31746, opened 2026-01-20
feat(engine): add compression template functions

owanio1992:
This PR adds gzip and ungzip functions to the template engine, allowing users to manually compress and decompress strings within their Helm charts.
Related to  #30989

## result
```
$ ls -ahl
34K Jan 20 20:45 Node_Exporter_Detail.compressed.yaml
661K Jan 20 19:30 Node_Exporter_Detail.json

```

## docs
helm/helm-www#1989

### DIFF
--- a/pkg/engine/funcs.go
+++ b/pkg/engine/funcs.go
@@ -18,10 +18,15 @@ package engine
 
 import (
 	"bytes"
+	"compress/gzip"
+	"encoding/base64"
 	"encoding/json"
+	"fmt"
+	"io"
 	"maps"
 	"strings"
 	"text/template"
+	"unicode/utf8"
 
 	"github.com/BurntSushi/toml"
 	"github.com/Masterminds/sprig/v3"
@@ -60,6 +65,8 @@ func funcMap() template.FuncMap {
 		"mustToJson":    mustToJSON,
 		"fromJson":      fromJSON,
 		"fromJsonArray": fromJSONArray,
+		"gzip":          gzipFunc,
+		"ungzip":        ungzipFunc,
 
 		// This is a placeholder for the "include" function, which is
 		// late-bound to a template. By declaring it here, we preserve the
@@ -232,3 +239,74 @@ func fromJSONArray(str string) []any {
 	}
 	return a
 }
+
+// gzipFunc compresses a string using gzip and returns the base64 encoded result.
+//
+// It enforces a size limit (1MB) on the output to prevent creating objects too large for Kubernetes Secrets/ConfigMap.
+//
+// This is designed to be called from a template.
+func gzipFunc(str string) (string, error) {
+	var b bytes.Buffer
+	w := gzip.NewWriter(&b)
+	if _, err := w.Write([]byte(str)); err != nil {
+		return "", err
+	}
+	if err := w.Close(); err != nil {
+		return "", err
+	}
+
+	encoded := base64.StdEncoding.EncodeToString(b.Bytes())
+
+	// Kubernetes limit for Secret/ConfigMap is 1MB.
+	const maxLimit = 1048576
+	if len(encoded) > maxLimit {
+		return "", fmt.Errorf("gzip: output size %d exceeds limit of %d bytes", len(encoded), maxLimit)
+	}
+
+	return encoded, nil
+}
+
+// ungzipFunc decodes a base64 encoded and gzip-compressed string.
+//
+// It enforces a size limit (1MB) on the input and output to prevent abuse.
+//
+// This is designed to be called from a template.
+func ungzipFunc(str string) (string, error) {
+	// Kubernetes limit for Secret/ConfigMap is 1MB.
+	const maxLimit = 1048576
+
+	if len(str) > maxLimit {
+		return "", fmt.Errorf("ungzip: input size %d exceeds limit of %d bytes", len(str), maxLimit)
+	}
+
+	decoded, err := base64.StdEncoding.DecodeString(str)
+	if err != nil {
+		return "", fmt.Errorf("ungzip: base64 decode failed: %w", err)
+	}
+
+	r, err := gzip.NewReader(bytes.NewReader(decoded))
+	if err != nil {
+		return "", err
+	}
+	defer r.Close()
+
+	// Enforce a size limit on the decompressed content.
+	limitR := io.LimitReader(r, maxLimit+1)
+
+	b, err := io.ReadAll(limitR)
+	if err != nil {
+		return "", err
+	}
+
+	if len(b) > maxLimit {
+		return "", fmt.Errorf("ungzip: decompressed content exceeds size limit of %d bytes", maxLimit)
+	}
+
+	// Ensure the content is valid text (UTF-8) to prevent binary obfuscation.
+	if !utf8.Valid(b) {
+		return "", fmt.Errorf("ungzip: content is not valid UTF-8 text")
+	}
+
+	return string(b), nil
+}
+
--- a/pkg/engine/funcs_test.go
+++ b/pkg/engine/funcs_test.go
@@ -17,6 +17,9 @@ limitations under the License.
 package engine
 
 import (
+	"bytes"
+	"compress/gzip"
+	"encoding/base64"
 	"strings"
 	"testing"
 	"text/template"
@@ -127,6 +130,14 @@ keyInElement1 = "valueInElement1"`,
 		tpl:    `{{ lookup "v1" "Namespace" "" "unlikelynamespace99999999" }}`,
 		expect: `map[]`,
 		vars:   `["one", "two"]`,
+	}, {
+		tpl:    `{{ "hello world" | gzip }}`,
+		expect: `H4sIAAAAAAAA/8pIzcnJVyjPL8pJAQQAAP//hRFKDQsAAAA=`,
+		vars:   nil,
+	}, {
+		tpl:    `{{ "H4sIAAAAAAAA/8pIzcnJVyjPL8pJAQQAAP//hRFKDQsAAAA=" | ungzip }}`,
+		expect: `hello world`,
+		vars:   nil,
 	}}
 
 	for _, tt := range tests {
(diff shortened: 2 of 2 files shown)


benoittgt:
Hello

Thanks for your contribution @owanio1992.

Beyond decompression bombs, ungzip probably enables hiding/obfuscating rendered manifests inside charts/values, which can undermine review and increase supply-chain risk (especially in GitOps). What do you think about this?
 If we add it, we could (a) use base64 text, (b) enforce strict compressed+decompressed size limits (?), and (c) document security implications / discourage use on untrusted inputs.

But with all this in mind I am wondering if it still a good idea to add this. :/ 

Also you should look at DCO error + add tests.

owanio1992:
@benoittgt 
I have updated the implementation to address the security concerns and align with other Helm functions:

1.  **Size Limits (1MB)**:
    -   `gzip`: Enforces that the compressed output (Base64 encoded) does not exceed 1MB.
    -   `ungzip`: Enforces that the input string does not exceed 1MB, and the decompressed content does not exceed 1MB. This effectively prevents decompression bombs and ensures the result fits within Kubernetes Secret/ConfigMap limits.

2.  **Base64 Encoding**:
    -   `gzip` now automatically returns a Base64 encoded string.
    -   `ungzip` now expects a Base64 encoded string as input.
    -   This aligns with the existing `encryptAES`/`decryptAES` behavior and simplifies usage in Secrets.

3.  **Content Validation**:
    -   `ungzip` verifies that the decompressed content is valid UTF-8 text, valid UTF-8 text I think is more readable. This prevents the functions from being used to hide arbitrary binary executables or obfuscate content.

4.  **Tests**:
    -   Unit tests have been updated to reflect the Base64 changes and verify the functionality.

5.  **DCO**:
    -   Commits are now properly signed off.
    
Ready for review.

owanio1992:
@giuliocalzo 

I have Update: Extended support to include Zstd, LZ4, Snappy, and Brotli.
### Algorithm Selection Context
I evaluated a range of common compression algorithms to check for maintenance status and stability in the Go ecosystem.
- **Included**:
  - **Zstd & Snappy**: Using `github.com/klauspost/compress` (already an indirect dependency in Helm).
  - **LZ4**: Using `github.com/pierrec/lz4/v4` (widely used, strict memory safety).
  - **Brotli**: Using `github.com/andybalholm/brotli` (standard Go implementation).
- **Excluded (Originally Planned but Dropped)**:
  - **Bzip2**: Removed from the plan because the Go standard library (`compress/bzip2`) only supports decompression. The most common third-party writer (`github.com/dsnet/compress`) appears unmaintained/archived.
  - **XZ**: Removed due to concerns over the stability and maintenance of available pure Go implementations (e.g., `ulikuinitz/xz` is on v0.5.x), preferring to stick to specific, well-established libraries for now.

owanio1992:
> I've reopened #30989
> 
> Can you describe your use-case there for data (de)compression support in Helm templates.


I am looking for native data (de)compression support in Helm templates for the following use cases:

- Compression: To support Grafana gzipped dashboards (as seen in the [Grafana Operator documentation](https://grafana.github.io/grafana-operator/docs/examples/dashboard/gzip_json/readme/)), where dashboard JSON data needs to be compressed before being stored in a ConfigMap or Secret.

- Decompression: This would be used primarily to verify compression results within the template or to handle incoming compressed data.

Following the pattern of existing Helm functions like encryptAES and decryptAES, I believe it is consistent and helpful to have a corresponding decompress function for every compress function provided.

owanio1992:
@benoittgt @giuliocalzo

Hi, thanks for the insightful feedback regarding security and obfustication.

I have updated the PR to address these concerns:
Security: Added strict limits on the decompressed size to prevent decompression bombs.
Testing: Included comprehensive unit tests to ensure reliability.
DCO: Fixed the DCO sign-off issue.


Could you please take another look when you have a moment? Thanks!

joejulian (changes requested):
I have a few things:

1. Do we want independent functions, like this, or would it be better to have a 'compress/decompress' function that take an engine argument, ie:
```
{{ "hello world" | compress "gzip" }}
{{ "GwsAACRBQpBKkUVq+pw1CQ==" | decompress "brotli" }}
```

This would fit better with your initial framing and would scale better as algorithms evolve. It would allow DRY validation of input and output, and it would also simplify documentation and test burden.

2. I do wonder if this should still be behind a flag. Even with UTF-8 checking the contents can still be opaque and a malicious chart can still include binary blobs that cannot be audited.
```
{{ "<my-binary-virus-base64-compressed>| gunzip | b64dec" }}
```

A flag that ensures that you accept the trust issue to enable compression/decompression functions might be worth considering.

3. Can you add a decompression bomb test, please.

owanio1992:
Hi @joejulian, thanks for the thoughtful review! Here are my thoughts on each point:

3. Decompression bomb test

Absolutely, will add. I'll create a test that compresses >1MB of highly repetitive data (which compresses down to a very small payload) and verifies that decompression is correctly blocked by the io.LimitReader size limit.

1. Independent functions vs. unified compress/decompress

I've given this some thought, and I'd like to propose keeping independent functions and scaling back to gzip-only for now. Here's my reasoning:

Consistency with existing Helm patterns: Helm already follows the independent function convention — toJson/fromJson, toYaml/fromYaml, toToml/fromToml, encryptAES/decryptAES. None of these use a unified to "json" or encrypt "aes" pattern. I believe following the established convention is more user-friendly and predictable.
YAGNI: gzip remains the dominant compression format for the primary use case (e.g., Grafana Operator's gzipped dashboards). I originally expanded to 5 algorithms based on feedback from #30989, but on further reflection, I think it's more pragmatic to start with gzip only and add more algorithms if there's concrete demand. This also eliminates 2 new direct dependencies (brotli, lz4), reducing the supply-chain surface.
If we ever need to add more algorithms in the future, adding zstd/unzstd as independent functions would be straightforward and consistent.

2. Feature flag for compression functions

I understand the concern about opaque content. However, I'd note that existing Helm functions already handle opaque data without requiring a flag:

b64dec — can decode arbitrary opaque base64 content
decryptAES — produces/consumes opaque encrypted data
If we require a flag for gzip/ungzip, it would be inconsistent not to require one for these as well.

That said, the compression functions do include mitigations that b64dec doesn't have:

1MB input/output size limit (aligned with Kubernetes Secret/ConfigMap limits)
UTF-8 validation on decompressed output (prevents binary payloads)
io.LimitReader to cap decompression during the read process
I believe these mitigations adequately address the risk. But I'm open to adding a flag if the maintainers feel it's necessary for the project's security posture.

owanio1992:
Hi @joejulian @benoittgt @giuliocalzo, 

Since it's been a while without further feedback on my previous proposal, I have gone ahead and implemented the approach I outlined in my last comment to keep this moving forward:

1. **Scaled back to [gzip](cci:1://file:///home/owan/data/git/helm/pkg/engine/funcs.go:242:0-266:1) and [ungzip](cci:1://file:///home/owan/data/git/helm/pkg/engine/funcs.go:268:0-310:1) only**: Following the YAGNI principle and to minimize our third-party dependency surface, I have completely removed `zstd`, `lz4`, `snappy`, and `brotli` support along with their dependencies. This is now much closer to Helm's existing `encryptAES`/`toJson` functional patterns.
2. **Added a Decompression Bomb Test**: I've successfully added [TestUngzipDecompressionBomb](cci:1://file:///home/owan/data/git/helm/pkg/engine/funcs_test.go:256:0-280:1) to the test suite. It creates a highly repetitive 2MB zero-byte payload that shrinks to a tiny base64 string, and validates that [ungzipFunc](cci:1://file:///home/owan/data/git/helm/pkg/engine/funcs.go:268:0-310:1) strictly halts decompression via `io.LimitReader` and returns the expected `1048576 bytes` size limit error.
3. **No Feature Flag**: No feature flag was added, as the strict 1MB limits and UTF-8 enforcement (combined with the existing behavior matching `decryptAES` and `b64dec`) seem robust enough to mitigate risks out of the box.

The commits are signed-off properly and the codebase is much cleaner now. Could you please take another look when you have a moment? Thanks!

kvaps:
Adding a concrete use-case from the Cluster API ecosystem:

`cluster-api-operator` (`operator.cluster.x-k8s.io`) expects provider components (`bootstrap`/`controlplane`/`infrastructure`/`core` CRDs + RBAC + Deployment manifests) to ship as a **gzipped blob** inside `ConfigMap.binaryData.components`, marked with the `provider.cluster.x-k8s.io/compressed: "true"` annotation. The operator gunzips it at runtime to reconcile the bundle.

Today, anyone packaging a CAPI provider as a Helm chart has to keep **two** copies of the manifest in the repo: the plain `*-components.yaml` (for human review and `make update` automation) AND its gzipped `components.gz` companion (the actual artefact the chart embeds via `.Files.Get`). Every upstream provider bump regenerates both in lockstep and commits the binary diff.

Concrete numbers from one repo (cozystack — 4 CAPI-provider charts):

| Chart | Plain YAML | Committed `.gz` |
|---|---|---|
| `capi-providers-bootstrap` | 484 KB | 62 KB |
| `capi-providers-core` | 853 KB | 88 KB |
| `capi-providers-cpprovider` | 1.1 MB | 126 KB |
| `capi-providers-infraprovider` | 731 KB | 71 KB |

With `gzip`/`gunzip` template functions, all of this collapses to a single

```gotemplate
binaryData:
  components: {{ .Files.Get "files/control-plane-components.yaml" | gzip | b64enc | quote }}
```

— no binary blob in git, no lockstep regeneration burden, no broken `git apply` on fork PRs (which is exactly how it bit us: https://github.com/cozystack/cozystack/pull/2610). Air-gapped operators who fork community charts to vendor providers in-tree run into the same friction.

Strong +1.

UnasZole:
Can a maintainer please look at this PR and merge it ?

Several concrete use cases have been mentioned above, and more generally : helm templating can be used to generate configmaps, and configmaps are heavily limited in size.
So compression of configmap contents is a big technical enabler nowadays.

This PR seems simple enough, well reviewed and well validated : is there any remaining point of concern ?

PR #31748, opened 2026-01-21
refactor: remove per-file decompression size limit

### LINKED ISSUE #30738: MaxDecompressedFileSize in v3.17.3 breaks existing charts

benoittgt:

Fixes #30738

Remove the per-file `MaxDecompressedFileSize` (5 MiB) enforcement. The original limit was added to protect against vulnerabilities in an unmaintained JSON schema library that Helm no longer uses (migrated to `santhosh-tekuri/jsonschema/v6`). The variable itself is deprecated rather than removed for backward compatibility until v5.

Directory-based chart loading (`LoadDir`) previously had no aggregate size protection, unlike archive loading which already enforces `MaxDecompressedChartSize`. This PR adds a shared `ReadFileWithBudget` helper so both v2 and v3 directory loaders track the same 100 MiB total budget, addressing the concern raised in https://github.com/helm/helm/pull/31748#issuecomment-4138927643.

Charts with large individual files (like 8 MB `values.schema.json`) work fine as long as the total stays within `MaxDecompressedChartSize`.

Related:
  - https://github.com/helm/helm/pull/30743

The next step is probably to slightly increase the size limit for archives, as discussed in the Helm Dev Meeting to help users with very big charts to be able to upgrade Helm.

1. Do we need to document this?
2. Can you confirm whether we want to backport this change?

**If applicable**:

### DIFF
--- a/pkg/chart/loader/archive/archive.go
+++ b/pkg/chart/loader/archive/archive.go
@@ -37,10 +37,6 @@ import (
 // The default value is 100 MiB.
 var MaxDecompressedChartSize int64 = 100 * 1024 * 1024 // Default 100 MiB
 
-// MaxDecompressedFileSize is the size of the largest file that Helm will attempt to load.
-// The size of the file is the decompressed version of it when it is stored in an archive.
-var MaxDecompressedFileSize int64 = 5 * 1024 * 1024 // Default 5 MiB
-
 var drivePathPattern = regexp.MustCompile(`^[a-zA-Z]:/`)
 
 var utf8bom = []byte{0xEF, 0xBB, 0xBF}
@@ -126,10 +122,6 @@ func LoadArchiveFiles(in io.Reader) ([]*BufferedFile, error) {
 
 		if hd.Size > remainingSize {
 			return nil, fmt.Errorf("decompressed chart is larger than the maximum size %d", MaxDecompressedChartSize)
-		}
-
-		if hd.Size > MaxDecompressedFileSize {
-			return nil, fmt.Errorf("decompressed chart file %q is larger than the maximum file size %d", hd.Name, MaxDecompressedFileSize)
 		}
 
 		limitedReader := io.LimitReader(tr, remainingSize)
--- a/pkg/chart/loader/archive/archive.go
+++ b/pkg/chart/loader/archive/archive.go
@@ -37,6 +37,12 @@ import (
 // The default value is 100 MiB.
 var MaxDecompressedChartSize int64 = 100 * 1024 * 1024 // Default 100 MiB
 
+// MaxDecompressedFileSize is the size of the largest file that Helm will attempt to load.
+// The size of the file is the decompressed version of it when it is stored in an archive.
+//
+// Deprecated: This variable is no longer used internally and will be removed in Helm v5.
+var MaxDecompressedFileSize int64 = 5 * 1024 * 1024 // Default 5 MiB
+
 var drivePathPattern = regexp.MustCompile(`^[a-zA-Z]:/`)
 
 var utf8bom = []byte{0xEF, 0xBB, 0xBF}
--- a/pkg/chart/loader/archive/archive.go
+++ b/pkg/chart/loader/archive/archive.go
@@ -37,10 +37,10 @@ import (
 // The default value is 100 MiB.
 var MaxDecompressedChartSize int64 = 100 * 1024 * 1024 // Default 100 MiB
 
-// MaxDecompressedFileSize is the size of the largest file that Helm will attempt to load.
-// The size of the file is the decompressed version of it when it is stored in an archive.
+// MaxDecompressedFileSize was the per-file size limit enforced during chart loading.
+// It is no longer used internally; aggregate chart size is enforced via MaxDecompressedChartSize.
 //
-// Deprecated: This variable is no longer used internally and will be removed in Helm v5.
+// Deprecated: Retained for backward compatibility with external callers. Will be removed in Helm v5.
 var MaxDecompressedFileSize int64 = 5 * 1024 * 1024 // Default 5 MiB
 
 var drivePathPattern = regexp.MustCompile(`^[a-zA-Z]:/`)
--- a/pkg/chart/loader/archive/budget.go
+++ b/pkg/chart/loader/archive/budget.go
@@ -1,0 +1,43 @@
+/*
+Copyright The Helm Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package archive
+
+import (
+	"fmt"
+	"os"
+)
+
+// ReadFileWithBudget reads a file and decrements remaining by the bytes read.
+// It returns an error if the total would exceed MaxDecompressedChartSize.
+func ReadFileWithBudget(path string, size int64, remaining *int64) ([]byte, error) {
+	if size > *remaining {
+		return nil, fmt.Errorf("chart exceeds maximum decompressed size of %d bytes", MaxDecompressedChartSize)
+	}
+
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return nil, err
+	}
+
+	// Re-check with actual length: the file may have grown between stat and read.
+	if int64(len(data)) > *remaining {
+		return nil, fmt.Errorf("chart exceeds maximum decompressed size of %d bytes", MaxDecompressedChartSize)
+	}
+
+	*remaining -= int64(len(data))
+	return data, nil
+}
--- a/pkg/chart/loader/archive/budget.go
+++ b/pkg/chart/loader/archive/budget.go
@@ -23,26 +23,26 @@ import (
 	"os"
 )
 
-// budgetedReader tracks cumulative file reads against a size limit.
-type budgetedReader struct {
+// BudgetedReader tracks cumulative file reads against a size limit.
+type BudgetedReader struct {
 	max       int64
 	remaining int64
 }
 
-// newBudgetedReader creates a new budgetedReader with the given maximum decompressed chart size.
-// The remaining budget is initialized to the maximum size.
-func NewBudgetedReader(max int64) *budgetedReader {
-	return &budgetedReader{
+// NewBudgetedReader creates a BudgetedReader with the given maximum total size.
+// The remaining budget is initialized to the maximum.
+func NewBudgetedReader(max int64) *BudgetedReader {
+	return &BudgetedReader{
 		max:       max,
 		remaining: max,
 	}
 }
 
-// readFileWithBudget reads a file and decrements remaining by the bytes read.
-// It returns an error if the total would exceed the maximum decompressed chart size.
+// ReadFileWithBudget reads a file and decrements the remaining budget by the bytes read.
+// It returns an error if the total would exceed the configured maximum.
 // The read is capped via io.LimitReader so a file that grows between stat
 // and read cannot cause unbounded memory allocation.
-func (r *budgetedReader) ReadFileWithBudget(path string, size int64) ([]byte, error) {
+func (r *BudgetedReader) ReadFileWithBudget(path string, size int64) ([]byte, error) {
 	if size > r.remaining {
 		return nil, fmt.Errorf("chart exceeds maximum decompressed size of %d bytes", r.max)
 	}
--- a/pkg/chart/loader/archive/budget.go
+++ b/pkg/chart/loader/archive/budget.go
@@ -17,23 +17,37 @@ limitations under the License.
 package archive
 
 import (
+	"errors"
 	"fmt"
+	"io"
 	"os"
 )
 
 // ReadFileWithBudget reads a file and decrements remaining by the bytes read.
 // It returns an error if the total would exceed MaxDecompressedChartSize.
+// The read is capped via io.LimitReader so a file that grows between stat
+// and read cannot cause unbounded memory allocation.
 func ReadFileWithBudget(path string, size int64, remaining *int64) ([]byte, error) {
+	if remaining == nil {
+		return nil, errors.New("remaining budget must not be nil")
+	}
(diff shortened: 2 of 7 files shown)


pkg/chart/loader/archive/archive.go
```diff
@@ -128,7 +128,7 @@ func LoadArchiveFiles(in io.Reader) ([]*BufferedFile, error) {
 			return nil, fmt.Errorf("decompressed chart is larger than the maximum size %d", MaxDecompressedChartSize)
 		}
 
-		if hd.Size > MaxDecompressedFileSize {
+		if hd.Size > MaxDecompressedFileSize && n != "values.schema.json" {
```
gjenkins8: I think we should simply remove `MaxDecompressedFileSize` (and associated config).

The reason that `MaxDecompressedFileSize` was added was this advisory: https://github.com/helm/helm/security/advisories/GHSA-5xqw-8hwv-wg92, which is due to a vulnerability in (the unmaintained) <https://github.com/xeipuuv/gojsonschema>.

However, Helm no longer uses `github.com/xeipuuv/gojsonschema` (it now uses: `github.com/santhosh-tekuri/jsonschema/v6` ). So this vulnerability, and rationale for `MaxDecompressedFileSize` no longer exists IMHO.

---

I do think Helm should (must) have a total limit, for e.g. <https://github.com/helm/helm/security/advisories/GHSA-4hfp-h4cw-hj8p>. And so `MaxDecompressedChartSize` should (must) remain.

benoittgt: I've applied your suggestion.

gjenkins8 (dismissed):
LGTM

@mattfarina -- do you agree with my comment (recollection) here: https://github.com/helm/helm/pull/31748#discussion_r2713873626 ? (and then presumably the removal of `MaxDecompressedFileSize` ?)

pkg/chart/loader/archive/archive.go
```diff
@@ -37,10 +37,6 @@ import (
 // The default value is 100 MiB.
 var MaxDecompressedChartSize int64 = 100 * 1024 * 1024 // Default 100 MiB
 
-// MaxDecompressedFileSize is the size of the largest file that Helm will attempt to load.
-// The size of the file is the decompressed version of it when it is stored in an archive.
-var MaxDecompressedFileSize int64 = 5 * 1024 * 1024 // Default 5 MiB
```
mattfarina, line 42: We can't remove this variable. It's a public part of the API. If we stop using it, we need to leave it in until Helm v5. Can we get a removal notice for v5 instead.

mattfarina (dismissed):

pkg/chart/loader/archive/archive.go
benoittgt: You are right. I made the change via  https://github.com/helm/helm/pull/31748/commits/bd6fcf75e65fe2f5657958dce4861b046cb29f4d

gjenkins8 (dismissed):

TerryHowe (dismissed):
/lgtm

joejulian:
This does end up removing a safeguard for directory-based loading. A user (or automation) can point Helm at a local chart directory containing an extremely large file (e.g. a huge values.yaml or random blob). The loader calls os.ReadFile with no size checks so it will try to read the while file into memory. That can spike memory usage and trigger the OOM process killer.

joejulian (changes requested):
I believe without my comment being addressed, this will still be considered a vulnerability.

benoittgt:
Thanks @joejulian for your feedback. That's why I created https://github.com/helm/helm/pull/30743 first, so users could set their own limits for maximum file size or chart size. After further discussion, it was suggested we remove this limitation entirely.
I don't have additional arguments to make. The [initial PR](https://github.com/helm/helm/pull/30743) still seems like a reasonable approach to me.

wangboru:
> Thanks @joejulian for your feedback. That's why I created #30743 first, so users could set their own limits for maximum file size or chart size. After further discussion, it was suggested we remove this limitation entirely. I don't have additional arguments to make. The [initial PR](https://github.com/helm/helm/pull/30743) still seems like a reasonable approach to me.

The `MaxDecompressedFileSize` limit is meant to set secure by default. As @joejulian suggested, by removing `MaxDecompressedFileSize` limit, this allows Helm to attempt to read files of unlimited size into memory, leading to memory exhaustion and DoS attack. 
While https://github.com/helm/helm/pull/30743 is meant to allow users to set their own limits, I think secure by default and having some security guardrails by default is crucial. It will be better to make the limits configurable.

joejulian:
I asked @wangboru to give this a review. He has an MS in Computer and Information Systems Security and Information Assurance and has been working in software security since 2015. I trust his skills and from his review I believe we should not remove this restriction.

scottrigby:
> > Thanks @joejulian for your feedback. That's why I created #30743 first, so users could set their own limits for maximum file size or chart size. After further discussion, it was suggested we remove this limitation entirely. I don't have additional arguments to make. The [initial PR](https://github.com/helm/helm/pull/30743) still seems like a reasonable approach to me.
> 
> The `MaxDecompressedFileSize` limit is meant to set secure by default. As @joejulian suggested, by removing `MaxDecompressedFileSize` limit, this allows Helm to attempt to read files of unlimited size into memory, leading to memory exhaustion and DoS attack. While #30743 is meant to allow users to set their own limits, I think secure by default and having some security guardrails by default is crucial. It will be better to make the limits configurable.

@wangboru thanks for your review, and thanks Joe for requesting it. Agree with you that secure by default, with a configurable override, makes sense.

A point of detail: I want to make sure you saw this earlier comment by @gjenkins8 https://github.com/helm/helm/pull/31748#discussion_r2713873626. In short, [MaxDecompressedChartSize](https://github.com/helm/helm/blob/2b04c7804fdf743948f6f20a8abea58eb20abf0b/pkg/chart/loader/archive/archive.go#L35-L38) already addresses what you noted, and is not being changed by this PR (total size of all files in a chart). Whereas [MaxDecompressedFileSize](https://github.com/helm/helm/blob/2b04c7804fdf743948f6f20a8abea58eb20abf0b/pkg/chart/loader/archive/archive.go#L40-L42) was added to address an issue with a package that we no longer use, and this is the extra functionality this PR removes (and deprecates the variable, since we can not remove it for backwards compatibility as @mattfarina points out in https://github.com/helm/helm/pull/31748#discussion_r2738518696).

To clarify, are you saying that the functionality of `MaxDecompressedChartSize` (max total of all files in a chart) is not sufficient, and that we still need to additionally keep the functionality of `MaxDecompressedFileSize` (max size of any single file)? If so, why?

scottrigby:
> This does end up removing a safeguard for directory-based loading. A user (or automation) can point Helm at a local chart directory containing an extremely large file (e.g. a huge values.yaml or random blob). The loader calls os.ReadFile with no size checks so it will try to read the while file into memory. That can spike memory usage and trigger the OOM process killer.

actually I see what you mean. Dir-based loading does not have the safeguard that archive-based loading has. We could, however, use `MaxDecompressedChartSize` in the same way for dir/file loading as we do now with archive loading (tally total and check that next file size plus tally is < max size). As far as naming goes, strictly, a chart dir that has never been archived is "uncompressed", but "decompressed" seems clear that it refers to a chart dir that is not compressed as an archive.

benoittgt:
> To clarify, are you saying that the functionality of MaxDecompressedChartSize (max total of all files in a chart) is not sufficient, and that we still need to additionally keep the functionality of MaxDecompressedFileSize (max size of any single file)? If so, why?

Regarding this comment. I want to make sure I understand correctly. @joejulian @wangboru let me know if I'm wrong.

Attack scenario:
1. Attacker creates/contributes to a chart repo with a parent chart that has dependencies: `[{repository:
"file://../evil-subchart"}]`
2. The `evil-subchart/` directory contains a multi-G/MB `values.yaml`
3. Victim clones repo and runs `helm dependency update`
4. Without MaxDecompressedFileSize -> https://github.com/helm/helm/blob/2b04c7804fdf743948f6f20a8abea58eb20abf0b/pkg/chart/v2/loader/directory.go#L103-L107 tries to load the huge file into memory -> OOM kill

For me `MaxDecompressedChartSize` only protects
https://github.com/helm/helm/blob/main/pkg/chart/loader/archive/archive.go#L67. But `file://` dependencies trigger
https://github.com/helm/helm/blob/main/pkg/downloader/manager.go#L886 which calls `os.ReadFile` directly... no chart
size budget, only the per-file check.

I tested this locally with the check removed, a 100MB file gets loaded into memory with no validation. Happy to
share the test setup if helpful.

So +1 for https://github.com/helm/helm/pull/30743

joejulian:
I won't have time to give your question the attention it deserves before I leave on PTO for a week. I can look at this when I get back. If the other maintainers consider my opinion and find it to be invalid before then, feel free to dismiss my review and merge.

Otherwise, I'll see y'all when I return.

joejulian:
There's still the same issue of the only size guardrail on directory-based chart loading which turns local charts and file:// dependencies into an nbounded os.ReadFile path. 

I think the problem is that you and George are right. This doesn't need to be here any more. That said, there’s a better place to enforce this than the checks removed in this PR. 

The real issue here is not specifically “a single file over 5 MiB”, it’s unbounded memory use when Helm loads a chart directory into memory. Archive loading already protects against that with MaxDecompressedChartSize, but directory loading does not. After this change, LoadDir will os.ReadFile any regular file with no equivalent aggregate budget.

A better approach would be to make directory loading enforce a total chart-size budget the same way archive loading does. Concretely, pkg/chart/v2/loader/directory.go and internal/chart/v3/loader/directory.go could track a remaining byte budget and reject a chart once the total buffered size would exceed MaxDecompressedChartSize. That would preserve the legitimate goal of allowing large files such as values.schema.json, while still keeping Helm bounded and secure by default for local charts and file:// dependencies.

If you want to avoid duplicating the logic in both loaders, this seems like a good candidate for a shared helper that buffers files from a directory under a total-size budget.

I think that would address the regression concern much better than either:

1. keeping the old 5 MiB per-file limit as-is, or
2. removing the limit entirely without adding equivalent enforcement for directory loads.

benoittgt:
Thanks a lot @joejulian for your great proposal. I've implemented a version in https://github.com/helm/helm/pull/31748/commits/bb213c06afcae8492cb2eb3ae59badc8eb12ec9b.

Having a budget is a great idea.

Manually tested by temporarily setting zMaxDecompressedChartSizez to 1 byte and rebuilding. Both loading paths correctly reject:

```
$ bin/helm template test-release /tmp/podinfo/
Error: error reading .helmignore: chart exceeds maximum decompressed size of 1 bytes
```

```
$ bin/helm template test-release oci://ghcr.io/stefanprodan/charts/podinfo
Error: unable to load chart archive
```

joejulian:
Last tiny request: because this PR changes both the v2 and v3 directory loaders, could you add the same explicit LoadDir budget regression test for the v3 loader as well to prevent a future regression in v3?

benoittgt:
Oups missed it. Thanks @joejulian

joejulian (dismissed):
Thanks for seeing this through and putting up with my agonizing change requests.

Nice work! 🎉

pkg/chart/loader/archive/budget.go
```diff
@@ -0,0 +1,63 @@
...
+// ReadFileWithBudget reads a file and decrements remaining by the bytes read.
+// It returns an error if the total would exceed MaxDecompressedChartSize.
+// The read is capped via io.LimitReader so a file that grows between stat
+// and read cannot cause unbounded memory allocation.
+func ReadFileWithBudget(path string, size int64, remaining *int64) ([]byte, error) {
+	if remaining == nil {
+		return nil, errors.New("remaining budget must not be nil")
+	}
+	if size > *remaining {
+		return nil, fmt.Errorf("chart exceeds maximum decompressed size of %d bytes", MaxDecompressedChartSize)
```
gjenkins8: I don't think we should reference the global constant here (in theory, user could supply a different limit)

pkg/chart/loader/archive/budget.go
```diff
@@ -0,0 +1,63 @@
...
+	"io"
+	"math"
+	"os"
+)
+
+// ReadFileWithBudget reads a file and decrements remaining by the bytes read.
+// It returns an error if the total would exceed MaxDecompressedChartSize.
+// The read is capped via io.LimitReader so a file that grows between stat
+// and read cannot cause unbounded memory allocation.
+func ReadFileWithBudget(path string, size int64, remaining *int64) ([]byte, error) {
```
gjenkins8: Can we make this function private (no need to make public I think)

Also, rather than passing a pointer to the remaining, why not make this a type (struct) that can hold the remaining itself?

And similar, it could hold the max value (which would solve for the below comment on `MaxDecompressedChartSize`)

e.g. something like
```
type budgetedReader struct {
    max int64
    read int64
}

func (r *budgetedReader) ReadFile(path string, size int) {
   remaining := max - read
   if size > remaining {
		return nil, fmt.Errorf("chart exceeds maximum decompressed size of %d bytes", max)
   }
   ...
}
```

pkg/chart/loader/archive/budget.go
```diff
@@ -0,0 +1,63 @@
...
+	if limit < math.MaxInt64 {
+		limit++
+	}
+	data, err := io.ReadAll(io.LimitReader(f, limit))
+	if err != nil {
+		return nil, err
+	}
+
+	if int64(len(data)) > *remaining {
+		return nil, fmt.Errorf("chart exceeds maximum decompressed size of %d bytes", MaxDecompressedChartSize)
```
gjenkins8: Same as above

gjenkins8 (changes requested):
sorry, more review comments/changes requested. I think we should at least make `ReadFileWithBudget` non-public (so it can be evolved in the future)

benoittgt:
Thanks @gjenkins8. I will look at it.

pkg/chart/loader/archive/budget.go
benoittgt: The limitation is we use in other packages ? 🤔 Like https://github.com/helm/helm/pull/31748/changes#diff-1cc55dc7f8798b3b79251208f2eab2f7ea99fcdb46e7fdd2e9239256ed51c70bR108

I followed your good feedbacks in https://github.com/helm/helm/pull/31748/commits/41faef383f3418eedcd481ac12b3f8189fd08dbd

gjenkins8 (approved):

benoittgt:
@joejulian if you have time for a review ? ;)

jam-re:
Any update?

TerryHowe:
@scottrigby maybe have a look?

scottrigby (approved):

PR #31755, opened 2026-01-22
fix(template): deprecate unused --hide-notes and --render-subchart-notes flags

### LINKED ISSUE #31752: helm template command help output should not include --hide-notes or --render-subchart-notes flags
### What happened?

`helm template` should not render the install/upgrade instruction notes, and it fortunately does not. However there are two unused flags in the help output of this command that imply otherwise, leading to user confusion: `--hide-notes` and `--render-subchart-notes`. These were accidentally left over from Helm 2, and should be removed. See https://github.com/helm/helm/issues/6901#issuecomment-3786612928

Specifically, we should mark these flags deprecated and hide them when `helm template` command is used, then remove these for the `template` command in Helm 5.

```
% `helm template --help` | grep notes
      --hide-notes                                 if set, do not show notes in install output. Does not affect presence in chart metadata
      --render-subchart-notes                      if set, render subchart notes along with the parent
```

### Helm version

Helm major versions 3 and 4

### Kubernetes version

N/A

scottrigby:
The --hide-notes and --render-subchart-notes flags have no effect for `helm template` since template output never includes notes. Mark these flags as deprecated and hide them from help output.

The flags are preserved for backwards compatibility in Helm 4 and will be removed in Helm 5.

Fixes #31752

**If applicable**:

### DIFF
--- a/pkg/cmd/install.go
+++ b/pkg/cmd/install.go
@@ -209,6 +209,26 @@ func addInstallFlags(cmd *cobra.Command, f *pflag.FlagSet, client *action.Instal
 	f.BoolVar(&client.EnableDNS, "enable-dns", false, "enable DNS lookups when rendering templates")
 	f.BoolVar(&client.HideNotes, "hide-notes", false, "if set, do not show notes in install output. Does not affect presence in chart metadata")
 	f.BoolVar(&client.TakeOwnership, "take-ownership", false, "if set, install will ignore the check for helm annotations and take ownership of the existing resources")
+
+	// For `helm template`, these notes flags are legacy, unused, and should not show in help, but
+	// must remain accepted for backwards compatibility in Helm 4. Deprecate and hide them for now
+	// TODO remove these from template command in Helm 5
+	if cmd.Name() == "template" {
+		if err := cmd.Flags().MarkDeprecated("hide-notes", "this flag has no effect for 'helm template' and will be removed in Helm 5"); err != nil {
+			log.Fatal(err)
+		}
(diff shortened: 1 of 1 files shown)


benoittgt (commented):
Good catch!

Quick thought on the error handling here. Since `MarkDeprecated()` and `MarkHidden()` can only fail if the flag name is wrong (which should be caught in tests), maybe the `log.Fatal(err)` checks are a bit heavy-handed?

I was looking at usage in `kubectl` and they explicitly ignore it with `_`:
https://github.com/kubernetes/kubectl/blob/da94a05fbb652c3e2e97c88bd5add5da1f8740c0/pkg/cmd/autoscale/autoscale.go#L146-L147

Maybe follow that pattern here?

LGTM overall

TerryHowe (approved):
/lgtm

All our other deprecations ignore failure, but I kind of like this "blowed up real good" approach.

sabre1041 (approved):
LGTM

promptless-for-oss:
Promptless prepared a documentation update related to this change.

Triggered by [PR #31755](https://github.com/helm/helm/pull/31755)

Added documentation for the deprecated `--hide-notes` and `--render-subchart-notes` flags on `helm template`. These flags have no effect since template output never includes notes, and will be removed in Helm 5.

Review: [Document template deprecated flags](https://github.com/helm/helm-www/pull/2095)

TerryHowe:
I agree the log.Fatal is a bit heavy handed, but it will make a proble obvious

PR #31758, opened 2026-01-23
fix: prevent warning when using version range constraints

### LINKED ISSUE #31757: Helm 4: Emits "unable to find exact version" when a version range was specified
### What happened?

This is related to https://github.com/helm/helm/issues/31548 / https://github.com/helm/helm/pull/31549 / https://github.com/helm/helm/pull/31254.

When running commands like helm show chart or helm template using a version range, a warning is emitted that suggests that no exact version can be found:

```shell
helm show chart brigade/brigade --version '^1'
evel=WARN msg="unable to find exact version requested; falling back to closest available version" chart=brigade requested=^1 selected=1.10.0
apiVersion: v1
...
description: Brigade provides event-driven scripting of Kubernetes pipelines.
name: brigade
version: 1.10.0
```

In case a version range was specified (i.e. the `--version` parameter was set to `^1` or `~1.10`), no warning should be emitted and a version matching the criteria is used as described by the `--version` flag description:
>       --version string             specify a version constraint for the chart version to use. This constraint can be a specific tag (e.g. 1.1.1) or it may reference a valid range (e.g. ^2.0.0). If this is not specified, the latest version is used

1. `helm repo add brigade https://brigadecore.github.io/charts`
2. `helm repo update`
3. `helm show chart brigade/brigade --version ^1`

### Helm version

<details>

```console
$ helm version
version.BuildInfo{Version:"v4.0.5", GitCommit:"1b6053d48b51673c5581973f5ae7e104f627fcf5", GitTreeState:"clean", GoVersion:"go1.25.5", KubeClientVersion:"v1.34"}
```
</details>

### Kubernetes version

<details>

```console
$ kubectl version
Client Version: v1.32.7
Kustomize Version: v5.5.0
Server Version: v1.33.6
```

</details>

benoittgt:

When using version ranges like `^1` or `~1.10`, Helm incorrectly showed warnings about falling back to closest version 😬. Only show the warning when an exact version is requested but not found and move to debug for version range.

Fixes: https://github.com/helm/helm/issues/31757

I've also tested via the CLI that all unit test value are working properly. For example:

```sh
bin/helm show chart brigade/brigade --version '^1.8.0 || ^1.9.0' | grep -E '(level=|^version:)'
version: 1.10.0
```

Also other scenarios like

```sh
❯ bin/helm show chart jetstack/cert-manager --version 'v1' | head -1
level=WARN msg="unable to find exact version requested; falling back to closest available version" chart=cert-manager requested=v1 selected=v1.19.2
annotations:

❯ bin/helm show chart jetstack/cert-manager --version 'v1.x' | head -1
annotations:

❯ bin/helm show chart jetstack/cert-manager --version 'v1.x' --debug | head -1
level=DEBUG msg="original chart version" version=v1.x
level=DEBUG msg="set up default downloader cache"
level=DEBUG msg="selected version matching constraint" chart=cert-manager constraint=v1.x selected=v1.19.2
level=DEBUG msg="found chart in cache" id=87e2dafa946bd05c56be897b2fe2e171f2bbfad96c9d48409d4b1188d240af6f
annotations:

❯ bin/helm show chart jetstack/cert-manager --version '^v1' | head -1
annotations:

❯ bin/helm show chart brigade/brigade --version '^1' --debug | grep -E '(level=DEBUG.*selected version)'           
level=DEBUG msg="selected version matching constraint" chart=brigade constraint=^1 selected=1.10.0
```

**If applicable**:

### DIFF
--- a/pkg/repo/v1/index.go
+++ b/pkg/repo/v1/index.go
@@ -178,7 +178,7 @@ func (i IndexFile) SortEntries() {
 // isVersionRange checks if the version string is a range constraint (e.g., "^1", "~1.10")
 // rather than an exact version (e.g., "1.10.0").
 func isVersionRange(version string) bool {
-	return strings.ContainsAny(version, "^~<>=!*xX") || strings.Contains(version, " || ") || strings.Contains(version, " - ")
+	return strings.ContainsAny(version, "^~<>=!*xX") || strings.Contains(version, "||") || strings.Contains(version, " - ")
 }
 
 // Get returns the ChartVersion for the given name.
--- a/pkg/repo/v1/index.go
+++ b/pkg/repo/v1/index.go
@@ -223,6 +223,8 @@ func (i IndexFile) Get(name, version string) (*ChartVersion, error) {
 		if constraint.Check(test) {
 			if len(version) != 0 && !isVersionRange(version) {
 				slog.Warn("unable to find exact version requested; falling back to closest available version", "chart", name, "requested", version, "selected", ver.Version)
+			} else if len(version) != 0 && isVersionRange(version) {
+				slog.Debug("selected version matching constraint", "chart", name, "constraint", version, "selected", ver.Version)
 			}
 			return ver, nil
 		}
--- a/pkg/repo/v1/index.go
+++ b/pkg/repo/v1/index.go
@@ -175,6 +175,12 @@ func (i IndexFile) SortEntries() {
 	}
 }
 
+// isVersionRange checks if the version string is a range constraint (e.g., "^1", "~1.10")
+// rather than an exact version (e.g., "1.10.0").
+func isVersionRange(version string) bool {
+	return strings.ContainsAny(version, "^~<>=!*xX") || strings.Contains(version, " || ") || strings.Contains(version, " - ")
+}
+
 // Get returns the ChartVersion for the given name.
 //
 // If version is empty, this will return the chart with the latest stable version,
@@ -215,7 +221,7 @@ func (i IndexFile) Get(name, version string) (*ChartVersion, error) {
 		}
 
 		if constraint.Check(test) {
-			if len(version) != 0 {
+			if len(version) != 0 && !isVersionRange(version) {
 				slog.Warn("unable to find exact version requested; falling back to closest available version", "chart", name, "requested", version, "selected", ver.Version)
 			}
 			return ver, nil
--- a/pkg/repo/v1/index.go
+++ b/pkg/repo/v1/index.go
@@ -178,7 +178,14 @@ func (i IndexFile) SortEntries() {
 // isVersionRange checks if the version string is a range constraint (e.g., "^1", "~1.10")
 // rather than an exact version (e.g., "1.10.0").
 func isVersionRange(version string) bool {
-	return strings.ContainsAny(version, "^~<>=!*xX") || strings.Contains(version, "||") || strings.Contains(version, " - ")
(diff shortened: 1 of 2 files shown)


pkg/repo/v1/index.go
```diff
@@ -215,7 +221,7 @@ func (i IndexFile) Get(name, version string) (*ChartVersion, error) {
 		}
 
 		if constraint.Check(test) {
-			if len(version) != 0 {
+			if len(version) != 0 && !isVersionRange(version) {
```
Noksa, line 231: Can we also add a debug logging here when we use constraints and not exact version to see which version we got as closest which satisfy our requirement in `--version`?

Otherwise, we only see original (log before fix):
```shell
level=DEBUG msg="original chart version" version=1.9.1
level=DEBUG msg="set up default downloader cache"
level=WARN msg="unable to find exact version requested; falling back to closest available version" chart=cert-manager requested=1.9.1 selected=v1.9.1
level=DEBUG msg="found chart in cache" id=46273177b5ca6930af8047cd52bf0b784c8482ac042e7812d08643c2898692d5
```

It will show the version we got for `helm show` but I'm not sure about all commands :) 

Just as an idea.

benoittgt: I thought about this too as I think it is still interesting to know which version is selected when you specify a version range. I added the feature in https://github.com/helm/helm/pull/31758/commits/7749ebfe8c497f48643beb6fda5ab371d2717a59. Thanks for the good idea.

Noksa (approved):

pkg/repo/v1/index.go
```diff
@@ -175,6 +175,12 @@ func (i IndexFile) SortEntries() {
 	}
 }
 
+// isVersionRange checks if the version string is a range constraint (e.g., "^1", "~1.10")
+// rather than an exact version (e.g., "1.10.0").
+func isVersionRange(version string) bool {
+	return strings.ContainsAny(version, "^~<>=!*xX") || strings.Contains(version, " || ") || strings.Contains(version, " - ")
```
mattfarina: The check for " || " should remove the spaces and have "||". Spaces around the || aren't required.

mattfarina (dismissed):

pkg/repo/v1/index.go
benoittgt: You're right. https://github.com/helm/helm/pull/31758/commits/9e089fa80b6d458742c8b8e3c0748e024a48fcb3

benoittgt:
We should be good.

benoittgt:
Sorry @TerryHowe forgot about this PR. Legit feedback from copilot. Fixed in last commit.

benoittgt:
@mattfarina I think we should merge this to avoid unecessary noise.

TerryHowe (approved):
/lgtm

gjenkins8 (approved):

pkg/repo/v1/index.go
```diff
@@ -175,6 +175,19 @@ func (i IndexFile) SortEntries() {
 	}
 }
 
+// isVersionRange checks if the version string is a range constraint (e.g., "^1", "~1.10")
+// rather than an exact version (e.g., "1.10.0").
+func isVersionRange(version string) bool {
```
gjenkins8, line 180: It would be better to be able to get the constraint operations from `semver.Constraint` type, IMHO. Rather than trying to duplicate the parsing here.

But given this is simply for logging (though it is possible this could change in the future), this is fine I think.

PR #31780, opened 2026-01-31
feat(pull): enable debug logging for HTTP requests

akshat09867:
Closes #31098
Supersedes #31108 (which is stale and has conflicts)

### Description
This PR enables HTTP-level debug logging when the `--debug` flag is passed to `helm pull`.

**Changes:**
1. **pkg/getter**: Added `WithDebug` option and a `Debug` field to `getterOptions`.
2. **pkg/getter**: Implemented a `debugTransport` (RoundTripper) in `httpgetter` using `slog` to log request/response details.
3. **pkg/getter**: Updated `plugingetter` to inject `HELM_DEBUG=1` into the plugin environment if debug is enabled.
4. **pkg/action**: Passed the debug setting from `cli.EnvSettings` down to `ChartDownloader`.

### Testing
Ran `helm pull strimzi/strimzi-kafka-operator --version 0.46.1 --debug` and verified that `slog` debug messages appear for the HTTP request and redirects.

### DIFF
--- a/pkg/getter/httpgetter.go
+++ b/pkg/getter/httpgetter.go
@@ -133,7 +133,9 @@ func (g *HTTPGetter) httpClient() (*http.Client, error) {
 		g.transport = &http.Transport{
 			DisableCompression: true,
 			Proxy:              http.ProxyFromEnvironment,
-			TLSClientConfig:    &tls.Config{},
+			// Being nil would cause the tls.Config default to be used
+			// "NewTLSConfig" modifies an empty TLS config, not the default one
+			TLSClientConfig: &tls.Config{},
 		}
 	})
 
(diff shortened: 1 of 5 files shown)


pkg/getter/httpgetter.go
```diff
@@ -122,9 +133,7 @@ func (g *HTTPGetter) httpClient() (*http.Client, error) {
 		g.transport = &http.Transport{
 			DisableCompression: true,
 			Proxy:              http.ProxyFromEnvironment,
-			// Being nil would cause the tls.Config default to be used
```
benoittgt, line 125: Isn't something we want to keep?

akshat09867: Fixed! I restored the missing comments in httpgetter.go. I also realized I accidentally dropped some TODOs in plugingetter.go during the refactor, so I restored those as well. Thanks for catching this.

PR #31781, opened 2026-01-31
Ensure `--create-namespace` does not modify existing namespaces

Mujib-Ahasan:
When using `--create-namespace` with Helm, the chart tries to create the namespace if it doesn’t exist. With server-side apply (--server-side), Helm was previously trying to apply the namespace even if it already existed. This could overwrite or remove existing labels and annotations, which is not expected.

This PR fixes that by first checking if the namespace already exists. If it does, Helm skips creating it. If it doesn’t, Helm creates it (using server-side apply if requested). This makes `--create-namespace` idempotent and ensures existing namespace are left unchange.

closes #31767 
**If applicable**:

### DIFF
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -447,11 +447,22 @@ func (i *Install) RunWithContext(ctx context.Context, ch ci.Charter, vals map[st
 			return nil, err
 		}
 
-		if _, err := i.cfg.KubeClient.Create(
-			resourceList,
-			kube.ClientCreateOptionServerSideApply(i.ServerSideApply, false)); err != nil && !apierrors.IsAlreadyExists(err) {
+		namespaceExist := false
+		_, err = i.cfg.KubeClient.Get(resourceList, false)
+		if err == nil {
+			namespaceExist = true
+		} else if !apierrors.IsNotFound(err) {
 			return nil, err
 		}
+
+		if !namespaceExist {
+			if _, err := i.cfg.KubeClient.Create(
+				resourceList,
+				kube.ClientCreateOptionServerSideApply(i.ServerSideApply, false)); err != nil && !apierrors.IsAlreadyExists(err) {
+				return nil, err
+			}
+		}
+
 	}
 
 	// If Replace is true, we need to supersede the last release.
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -459,7 +459,6 @@ func (i *Install) RunWithContext(ctx context.Context, ch ci.Charter, vals map[st
 				return nil, err
 			}
 		}
-
 	}
 
 	// If Replace is true, we need to supersede the last release.
--- a/pkg/action/install.go
+++ b/pkg/action/install.go
@@ -447,15 +447,12 @@ func (i *Install) RunWithContext(ctx context.Context, ch ci.Charter, vals map[st
 			return nil, err
 		}
 
-		namespaceExist := false
-		_, err = i.cfg.KubeClient.Get(resourceList, false)
-		if err == nil {
-			namespaceExist = true
-		} else if !apierrors.IsNotFound(err) {
+		exists, err := namespaceExists(resourceList)
+		if err != nil {
 			return nil, err
 		}
 
-		if !namespaceExist {
+		if !exists {
 			if _, err := i.cfg.KubeClient.Create(
 				resourceList,
 				kube.ClientCreateOptionServerSideApply(i.ServerSideApply, false)); err != nil && !apierrors.IsAlreadyExists(err) {
@@ -485,6 +482,29 @@ func (i *Install) RunWithContext(ctx context.Context, ch ci.Charter, vals map[st
 		rel, err = i.failRelease(rel, err)
 	}
 	return rel, err
+}
+
+func namespaceExists(resourceList kube.ResourceList) (bool, error) {
+	exists := false
+
+	err := resourceList.Visit(func(info *resource.Info, err error) error {
+		if err != nil {
+			return err
+		}
+
+		_, err = resource.NewHelper(info.Client, info.Mapping).Get(info.Namespace, info.Name)
+		if apierrors.IsNotFound(err) {
+			return nil
+		}
+		if err != nil {
+			return err
+		}
+
+		exists = true
+		return nil
+	})
+
+	return exists, err
 }
 
 func (i *Install) performInstallCtx(ctx context.Context, rel *release.Release, toBeAdopted kube.ResourceList, resources kube.ResourceList) (*release.Release, error) {
--- a/pkg/action/install_test.go
+++ b/pkg/action/install_test.go
@@ -1320,7 +1320,7 @@ func fakeNamespaceResourceList(name string, statusCode int) kube.ResourceList {
 	restClient := &fake.RESTClient{
 		GroupVersion:         schema.GroupVersion{Version: "v1"},
 		NegotiatedSerializer: scheme.Codecs.WithoutConversion(),
-		Client: fake.CreateHTTPClient(func(req *http.Request) (*http.Response, error) {
+		Client: fake.CreateHTTPClient(func(_ *http.Request) (*http.Response, error) {
 			return newNamespaceResponse(statusCode, body), nil
 		}),
 	}
@@ -1369,7 +1369,7 @@ type namespaceInstallKubeClient struct {
 	namespaceCreateCalled bool
 }
 
-func (c *namespaceInstallKubeClient) Build(reader io.Reader, validate bool) (kube.ResourceList, error) {
+func (c *namespaceInstallKubeClient) Build(reader io.Reader, _ bool) (kube.ResourceList, error) {
 	b, err := io.ReadAll(reader)
 	if err != nil {
 		return nil, err
@@ -1382,7 +1382,7 @@ func (c *namespaceInstallKubeClient) Build(reader io.Reader, validate bool) (kub
 	return kube.ResourceList{}, nil
 }
 
-func (c *namespaceInstallKubeClient) Create(resources kube.ResourceList, options ...kube.ClientCreateOption) (*kube.Result, error) {
+func (c *namespaceInstallKubeClient) Create(resources kube.ResourceList, _ ...kube.ClientCreateOption) (*kube.Result, error) {
 	for _, info := range resources {
 		if info.Name == "spaced" && info.Mapping != nil && info.Mapping.GroupVersionKind.Kind == "Namespace" {
 			c.namespaceCreateCalled = true
--- a/pkg/action/install_test.go
+++ b/pkg/action/install_test.go
@@ -1294,3 +1294,41 @@ func TestInstallRelease_WaitOptionsPassedDownstream(t *testing.T) {
 	// Verify that WaitOptions were passed to GetWaiter
 	is.NotEmpty(failer.RecordedWaitOptions, "WaitOptions should be passed to GetWaiter")
 }
+
+func TestInstall_CreateNamespaceAlreadyExists(t *testing.T) {
+	is := assert.New(t)
+
+	// Namespace already exists in cluster
+	config := actionConfigFixtureWithDummyResources(t, createDummyResourceList(true))
+
+	instAction := installActionWithConfig(config)
+	instAction.CreateNamespace = true
+	instAction.Namespace = "spaced"
+
+	resi, err := instAction.Run(buildChart(), nil)
+	is.NoError(err)
+
+	res, err := releaserToV1Release(resi)
+	is.NoError(err)
+	is.Equal("spaced", res.Namespace)
+	is.Equal("Install complete", res.Info.Description)
+}
+
+func TestInstall_CreateNamespaceNotExists(t *testing.T) {
+	is := assert.New(t)
+
+	// Empty cluster state, thats mean namespace not present
+	config := actionConfigFixture(t)
+
+	instAction := installActionWithConfig(config)
+	instAction.CreateNamespace = true
+	instAction.Namespace = "spaced"
+
+	resi, err := instAction.Run(buildChart(), nil)
+	is.NoError(err)
+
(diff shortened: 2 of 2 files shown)


TerryHowe (dismissed):
This looks reasonable to me, but is it possible to add a test?

Mujib-Ahasan:
@TerryHowe Sure 🙂, I can add a test to cover this behavior. I will update the PR shortly.

pkg/action/install_test.go
```diff
@@ -1208,3 +1208,41 @@ func TestInstallRelease_WaitOptionsPassedDownstream(t *testing.T) {
...
+
+	instAction := installActionWithConfig(config)
+	instAction.CreateNamespace = true
+	instAction.Namespace = "spaced"
+
+	resi, err := instAction.Run(buildChart(), nil)
+	is.NoError(err)
+
+	res, err := releaserToV1Release(resi)
+	is.NoError(err)
```
wasd171, line 1416: Current code also does not error on an existing namespace, it silently replaces it and strips it off its labels / annotations. Could we check for this behaviour instead? So an existing namespace with a custom label should keep it after deploy

Mujib-Ahasan: Thank you for pointing it out @wasd171. there should be a test cover that explicitly verifies that existing namespace labesl/annotations remain unchanged after using `--create-namespace`. This actually aligns with your actual bug.

Mujib-Ahasan: Hey @TerryHowe `createDummyResourceList()` is only for deployment creation. Currently is there any thing that actually creates namespace resource in test.?

wasd171:
@TerryHowe could you please take a look at this PR? 🙏🏻

pkg/action/install_test.go
TerryHowe: Doesn't seem to be anything right now.

Mujib-Ahasan: > there should be a test cover that explicitly verifies that existing namespace labesl/annotations remain unchanged after using --create-namespace. This actually aligns with your actual bug.

is there a way I can achieve this?? @TerryHowe

TerryHowe: I don't know.

wasd171:
Thanks for working on this! I've looked a bit deeper into the original issue and the current test infra

So, my understanding of the bug is:

- `--create-namespace` promises to create the release namespace only if it is not present
- Helm builds a synthetic `Namespace` object and calls `KubeClient.Create(...)`
- With server-side apply enabled, this create path becomes an apply/PATCH operation
- If the namespace already exists, Kubernetes accepts that apply and may mutate the existing namespace metadata
- This is why an existing namespace with custom labels/annotations can be changed by `helm install/upgrade --create-namespace --server-side`

So I think the high-level approach in this PR is right: if the namespace already exists, Helm should skip creating/applying the synthetic namespace object entirely

However, I think there may be an implementation issue with the current existence check:

```go
_, err = i.cfg.KubeClient.Get(resourceList, false)
if err == nil {
    namespaceExist = true
}
```

`pkg/kube.Client.Get()` is probably not a good API for an existence check here. I think it swallows per-resource `NotFound` errors. Internally it calls `getResource(info)`, but if that returns an error, it writes the error to a buffer and continues. The method can still return an empty object map with `nil` error

That means a missing namespace may be interpreted as existing, so `--create-namespace` could skip creating a namespace that is not present

A safer fix may be to check the namespace resource directly, similar to the pattern used in `pkg/action/validate.go`, by visiting the built namespace resource and calling `resource.NewHelper(info.Client, info.Mapping).Get(...)`

Conceptually:

```go
namespaceExists := false

err = resourceList.Visit(func(info *resource.Info, err error) error {
    if err != nil {
        return err
    }

    _, err = resource.NewHelper(info.Client, info.Mapping).Get(info.Namespace, info.Name)
        return nil
    }
    if err != nil {
        return err
    }

    namespaceExists = true
    return nil
})
if err != nil {
    return nil, err
}

if !namespaceExists {
    if _, err := i.cfg.KubeClient.Create(
        resourceList,
        kube.ClientCreateOptionServerSideApply(i.ServerSideApply, false),
    ); err != nil && !apierrors.IsAlreadyExists(err) {
        return nil, err
    }
}
```

I think it is still worth keeping the `apierrors.IsAlreadyExists(err)` handling around `Create()`, because another actor could create the namespace between the GET and CREATE

About the tests: I think this is testable without a real Kubernetes cluster, but the current tests do not seem to verify the actual namespace behavior yet

A few reasons:

1. `createDummyResourceList(true)` creates a `Deployment`, not a `Namespace`
2. `FailingKubeClient.Build()` returns `DummyResources` for any manifest when `DummyResources != nil`, so the namespace manifest built by `--create-namespace` is not actually represented as a namespace in the test
3. `PrintingKubeClient.Get()` always returns `nil` error, so the `CreateNamespaceNotExists` test does not model a missing namespace
4. The tests only assert that the release install succeeds. They do not assert whether the namespace was created, skipped, or mutated

The important behavior to test is probably:

1. Existing namespace with custom labels/annotations:
   - arrange fake cluster state with namespace `spaced` containing custom labels/annotations
   - run install with `CreateNamespace = true` and `ServerSideApply = true`
   - assert that namespace create/apply was not called
   - assert the existing labels/annotations are unchanged

2. Missing namespace:
   - arrange fake cluster state with no namespace `spaced`
   - run install with `CreateNamespace = true`
   - assert that namespace creation was called for `spaced`

This could be done with a small spy/stateful fake kube client specifically for this test, or possibly using the existing REST fake test pattern from `pkg/kube/client_test.go`. The key is that the fake needs to distinguish `Get(existing namespace)` from `Get(missing namespace)` and record whether `Create()` was called for the namespace. The current `PrintingKubeClient` / `FailingKubeClient` does not keep that state, so it cannot catch this regression

Happy to help with a follow-up patch or a test sketch if useful! Please keep in mind that Go is not my primary programming language, so I might not be 100% right

pkg/action/install_test.go
```diff
@@ -1294,3 +1295,155 @@ func TestInstallRelease_WaitOptionsPassedDownstream(t *testing.T) {
...
+	)
+
+	return &http.Response{
+		StatusCode: code,
+		Header: map[string][]string{
+			"Content-Type": {"application/json"},
+		},
+		Body: io.NopCloser(strings.NewReader(body)),
+	}
+}
```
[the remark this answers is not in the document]
Mujib-Ahasan: kuberuntime.EncodeOrDie(...) returns string not []bytes.

Mujib-Ahasan:
@wasd171, sorry for the delayed response. 
Thank you for the direction. I tried to follow it. when the release namespace already exists, Helm no longer try to create the namespace, preventing modifications to existing namespace metadata when `--create-namespace` is used with server-side apply. Could you please give it a look.??

pkg/action/install.go
```diff
@@ -447,11 +447,18 @@ func (i *Install) RunWithContext(ctx context.Context, ch ci.Charter, vals map[st
...
-			kube.ClientCreateOptionServerSideApply(i.ServerSideApply, false)); err != nil && !apierrors.IsAlreadyExists(err) {
+		exists, err := namespaceExists(resourceList)
+		if err != nil {
 			return nil, err
 		}
+
+		if !exists {
+			if _, err := i.cfg.KubeClient.Create(
+				resourceList,
+				kube.ClientCreateOptionServerSideApply(i.ServerSideApply, false)); err != nil && !apierrors.IsAlreadyExists(err) {
```
wasd171, line 458: I think this still leaves the original bug class open because the `GET` and the subsequent create/apply are not atomic.

If the namespace is missing at line 450, but another actor creates it before line 456 runs, this path will still call `Create(...)` with `ClientCreateOptionServerSideApply(i.ServerSideApply, false)`. When `i.ServerSideApply` is true, that can become an apply / PATCH against an existing namespace, which is exactly the behavior this PR is trying to avoid.

I would suggest making the synthetic `--create-namespace` object always use ordinary create, independent of the install’s server-side-apply setting, and keep the `AlreadyExists` handling for the race case:

```go
if _, err := i.cfg.KubeClient.Create(
	resourceList,
	kube.ClientCreateOptionServerSideApply(false, false),
); err != nil && !apierrors.IsAlreadyExists(err) {
	return nil, err
}
```

With that shape, Helm does not need a preflight `GET`: if the namespace exists, Kubernetes returns `AlreadyExists`; if it does not, Helm creates it; in neither case does Helm server-side apply the synthetic namespace over an existing object.

pkg/action/install.go
```diff
@@ -476,6 +483,29 @@ func (i *Install) RunWithContext(ctx context.Context, ch ci.Charter, vals map[st
...
+		if err != nil {
+			return err
+		}
+
+		exists = true
+		return nil
+	})
+
+	return exists, err
+}
```
wasd171, line 507: If we switch the namespace bootstrap path to ordinary create + `AlreadyExists`, I think this helper can be removed entirely.

Keeping a preflight `GET` here has two downsides:

1. it introduces a new permission requirement for `get namespaces` before `--create-namespace` can proceed;
2. it still needs `AlreadyExists` handling afterward because `GET -> CREATE` is inherently racy.

The create-only path is both smaller and safer for this specific case because the synthetic namespace is not a release-managed object that needs server-side apply semantics. The invariant we need is simply: _create if absent; never mutate if present_.

pkg/action/install_test.go
```diff
@@ -1294,3 +1295,155 @@ func TestInstallRelease_WaitOptionsPassedDownstream(t *testing.T) {
...
+
+	resi, err := instAction.Run(buildChart(), nil)
+	is.NoError(err)
+
+	res, err := releaserToV1Release(resi)
+	is.NoError(err)
+	is.Equal("spaced", res.Namespace)
+	is.Equal("Install complete", res.Info.Description)
+
+	is.False(client.namespaceCreateCalled, "existing namespace should not be created/applied again")
```
wasd171, line 1420: This test is moving in the right direction, but I think it is asserting the implementation detail rather than the user-visible invariant.

The bug from #31767 is not specifically "Helm called `Create`"; it is "an existing namespace lost or changed its custom labels/annotations when `--create-namespace --server-side` was used."

As written, this test requires `Create` not to be called. That would fail a safer implementation that uses ordinary non-SSA create and receives `AlreadyExists`, even though that implementation would preserve the namespace correctly.

Could we instead model an existing namespace with custom labels/annotations and assert that running install with `CreateNamespace = true` and `ServerSideApply = true` does not modify that existing namespace metadata? That would test the external contract directly and leave room for the simpler atomic create + `AlreadyExists` implementation.

pkg/action/install_test.go
```diff
@@ -1294,3 +1295,155 @@ func TestInstallRelease_WaitOptionsPassedDownstream(t *testing.T) {
...
+}
+
+func (c *namespaceInstallKubeClient) Create(resources kube.ResourceList, _ ...kube.ClientCreateOption) (*kube.Result, error) {
+	for _, info := range resources {
+		if info.Name == "spaced" && info.Mapping != nil && info.Mapping.GroupVersionKind.Kind == "Namespace" {
+			c.namespaceCreateCalled = true
+		}
+	}
+
+	return &kube.Result{}, nil
```
wasd171, line 1392: This fake currently discards the create options, so the tests cannot detect the most important distinction for this bug: whether the synthetic namespace was created with ordinary create or with server-side apply.

Since the regression is specifically caused by server-side apply turning the namespace “create” path into an apply / PATCH against an existing namespace, I think the test fake should make that observable. For example, it could fail or record a distinct state if the namespace create path is invoked with server-side apply enabled.

That would let the test prove that `--server-side` still applies to chart resources, but not to the synthetic namespace bootstrap object.

pkg/action/install_test.go
```diff
@@ -1294,3 +1295,155 @@ func TestInstallRelease_WaitOptionsPassedDownstream(t *testing.T) {
...
+	resi, err := instAction.Run(buildChart(), nil)
+	is.NoError(err)
+
+	res, err := releaserToV1Release(resi)
+	is.NoError(err)
+	is.Equal("spaced", res.Namespace)
+	is.Equal("Install complete", res.Info.Description)
+
+	is.True(client.namespaceCreateCalled, "missing namespace should be created")
+}
```
wasd171, line 1449: Could we add a race-style case here as well?

The important edge case is:

1. Helm checks or attempts to create the namespace;
2. another actor has already created it by the time Helm reaches the create operation;
3. Helm receives `AlreadyExists`;
4. Helm treats that as success and does not server-side apply over the existing namespace.

That case is what makes the implementation robust against the non-atomic `exists -> create` window. If the implementation is changed to ordinary create + `AlreadyExists`, this becomes the key safety test.

wasd171 (changes requested):
Thanks for working on this! I agree with the goal of the PR: `--create-namespace` should not mutate an existing namespace, especially when server-side apply is enabled.

I think the current implementation is still not quite safe enough to merge, though. The important invariant is not just "skip create when a namespace is observed to exist"; it is "the synthetic namespace object must never be server-side applied to an already existing namespace."

The current `GET -> conditional Create(... ServerSideApply ...)` flow still has a race (albeit unlikely): if the namespace is created by another actor after the existence check but before Helm’s create call, Helm can still perform a server-side apply against an existing namespace and mutate its metadata.

I think the simpler and safer fix is to avoid the preflight existence check and always create the synthetic namespace with ordinary non-SSA create, while continuing to ignore `AlreadyExists`. That makes the operation atomic from Helm’s perspective: absent namespace is created; existing namespace is left unchanged.

wasd171:
TLDR; I think that SSA is a wrong approach for this synthetic namespace: it is hard to implement it atomically, it might try to take ownership of the resources that Helm did not create. I see `--create-namespace` as an escape hatch, so I think it's fine to use a non-SSA for it, while resources that were 100% intended for the Helm to manage can continue using SSA

wasd171:
There is a smaller PR https://github.com/helm/helm/pull/32587 that uses client-side apply for this namespace

PR #31782, opened 2026-01-31
docs: clarify error message for invalid --set syntax

loserfruit966:
This PR improves a vague parse error message in pkg/strvals/parser.go by making it clearer that the failure is due to invalid --set syntax and by explaining what was being parsed. This makes Helm errors more actionable for users.

### DIFF
--- a/pkg/strvals/parser.go
+++ b/pkg/strvals/parser.go
@@ -431,7 +431,7 @@ func (t *parser) listItem(list []interface{}, i, nestedNameLevel int) ([]interfa
 		}
 		return setIndex(list, i, inner)
 	default:
-		return nil, fmt.Errorf("parse error: unexpected token %v", last)
+		return nil, fmt.Errorf("invalid --set syntax: unexpected token %q while parsing list or nested key", last)
 	}
 }
 


pkg/strvals/parser.go
```diff
@@ -431,7 +431,7 @@ func (t *parser) listItem(list []interface{}, i, nestedNameLevel int) ([]interfa
 		}
 		return setIndex(list, i, inner)
 	default:
-		return nil, fmt.Errorf("parse error: unexpected token %v", last)
+		return nil, fmt.Errorf("invalid --set syntax: unexpected token %q while parsing list or nested key", last)
```
benoittgt, line 434: I don't think this is valid as this code can be called from SDK too. Library code shouldn't assume how it was invoked. The caller might be using the Helm CLI, or they might be using Helm as a Go library directly. 🤷                                                                          
I think we should :                                                                                                                                                                                                             
1. Keep the library error generic (for example: "parse error: unexpected token %q in list/key expression")                                                                                                                      
2. If better CLI context is wanted, use CLI layer ?

benoittgt (commented):
Thanks for the contribution! I have a concern about this change. See related comment.                                                                        
Also, the commit prefix should probably be `fix:` rather than `docs:`` since this modifies code behavior.

PR #31837, opened 2026-02-17
feat: add OCI registry search support for helm search repo

### LINKED ISSUE #11000: Allow listing Versions with Helm on OCI

veeceey:
When using OCI registries, `helm search repo` returns "No results found" since there's
no local index to search. This has been a gap since OCI support was added - there's no
way to discover available chart versions without external tools like `skopeo`, `crane`,
or `oras`.

This adds OCI registry support to `helm search repo` so that passing an `oci://` reference
queries the registry directly using the [OCI Distribution Spec tag listing endpoint](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#content-discovery)
(`GET /v2/{name}/tags/list`).

### How it works

For each semver-valid tag, only the config layer is fetched (not the full chart tarball),
which contains the Chart.yaml metadata (name, version, appVersion, description). This
keeps the operation lightweight.

### Usage

```bash
# List all versions of a chart in an OCI registry
helm search repo oci://ghcr.io/org/charts/mychart --versions

# Show only the latest stable version
helm search repo oci://ghcr.io/org/charts/mychart

# Filter by semver constraint
helm search repo oci://ghcr.io/org/charts/mychart --version '^1.0.0'

# Include pre-release versions
helm search repo oci://ghcr.io/org/charts/mychart --devel
```

### Before

```
$ helm search repo oci://ghcr.io/org/charts/mychart --versions
No results found
```

### After

```
$ helm search repo oci://ghcr.io/org/charts/mychart --versions
NAME                                    CHART VERSION   APP VERSION   DESCRIPTION
oci://ghcr.io/org/charts/mychart        1.2.0           2.0.0         My Helm chart
oci://ghcr.io/org/charts/mychart        1.1.0           1.9.0         My Helm chart
oci://ghcr.io/org/charts/mychart        1.0.0           1.0.0         My Helm chart
```

Also supports `--output json`, `--output yaml`, `--fail-on-no-result`, TLS flags,
and basic auth (`--username`, `--password`) for authenticated registries.

Fixes #11000
Closes https://github.com/helm/helm/pull/31022


promptless-for-oss:

New suggestion: **[Document OCI registry search support for helm search repo](https://github.com/helm/helm-www/pull/2010)**

---
_Tip: See how your feedback shapes Promptless in [Agent Knowledge Base](https://app.gopromptless.ai/configure/settings) 🧠_

veeceey:
hey, wanted to check if this is on anyone's radar for review? no rush, just making sure it didn't get lost

TerryHowe:
I wasn't involved in the reviews, but there was https://github.com/helm/helm/pull/31022 might take a quick look at that and see if any of that feedback applies here. If you think you have a better solution, maybe add a "Closes: https://github.com/helm/helm/pull/31022" to your description.

veeceey:
Thanks for the pointer @TerryHowe! I went through #31022 and the review feedback there. The main concern from mattfarina was whether this should be `helm search oci` instead of extending `helm search repo`, since OCI registries aren't technically Helm repositories (no `index.yaml`, can't be searched offline, etc.).

I think that's a fair architectural concern. My implementation here does keep the OCI path cleanly separated from the existing repo search logic — it only activates when an `oci://` prefix is passed, so it doesn't affect the existing repo search behavior at all. But I can see the argument for a dedicated subcommand too.

I also addressed the other feedback from that PR: using `TagListPageSize` for pagination, better error messages for invalid references, and the functionality works across different registries (GHCR, ECR, Docker Hub, Harbor, etc.) since it sticks to the OCI Distribution Spec's tag listing endpoint.

Happy to refactor this into `helm search oci` if the maintainers prefer that approach — the core logic would be the same either way.

Adding `Closes #31022` to the description as well since this covers the same use case.

veeceey:
hey, just checking in - I've addressed the feedback from #31022 as mentioned above. would appreciate a review when someone gets a chance\!

PR #31838, opened 2026-02-17
fix: clarify coalesce warnings about maps vs non-maps

allexistence:

Refs #11118

This PR improves the clarity of coalesce warning messages by replacing the ambiguous terms "table/non-table" with clearer and more accurate terminology "map/object" and "non-map value".

The previous wording could be confusing to users unfamiliar with Helm's internal terminology. Since Helm operates on YAML structures, using "map/object" better reflects the actual data type and improves the user experience when troubleshooting configuration issues.

This PR also updates the corresponding unit tests to match the revised warning messages.

No functional behavior has been changed. This is strictly a user-facing message clarity improvement.

- Only warning message strings and unit test expectations were updated.
- No functional or logic changes were made.
- All tests pass locally using `go test ./...`.

**If applicable**:

### DIFF
--- a/pkg/chart/common/util/coalesce.go
+++ b/pkg/chart/common/util/coalesce.go
@@ -245,7 +245,7 @@ func coalesceValues(printf printFn, c chart.Charter, v map[string]any, prefix st
 					// If the original value is nil, there is nothing to coalesce, so we don't print
 					// the warning
 					if val != nil {
-						printf("warning: skipped value for %s.%s: Not a table.", subPrefix, key)
+						printf("warning: skipped value for %s.%s: Not a map/object.", subPrefix, key)
 					}
 				} else {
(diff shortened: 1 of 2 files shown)


TerryHowe (commented):
Should these be slog warnings?

allexistence:
For this PR, I kept the existing logging approach and only updated the wording. If preferred, I can follow up with a separate PR to migrate these to `slog`.

TerryHowe:
Can you rebase this so the tests are triggered?

allexistence:
Rebased and addressed. PTAL

allexistence:
- Replace 'table/non-table' with 'map/object/non-map' for clarity
- Use %T instead of %v for destination value to avoid secret leakage
- Update tests to match new warning messages

Refs #11118

PR #31841, opened 2026-02-18
feat: helm performance improvement by deep copying coalesce globals

### LINKED ISSUE #31865: Major Performance & Memory Optimization Across Helm Commands (package, template, install, lint)

A Pull Request has been raised that introduces significant performance and memory optimizations across multiple Helm commands, including:

- helm package
- helm template
- helm install
- helm lint
- (and other rendering-related workflows)

## Related PR:
https://github.com/helm/helm/pull/31841

### Why is this needed?

While working with large and complex charts (heavy templating, deep values structures, repeated evaluations), Helm execution showed:

- Increased processing time
- High memory consumption
- Slower CI/CD execution in enterprise environments
- Scalability degradation with larger values files

This behavior becomes more visible in:
- Large-scale deployments
- Monorepos with multiple charts
- CI-heavy environments
- Enterprise workloads

AnubhavT007:
Resolves : https://github.com/helm/helm/issues/31865

This PR uses deep copy in coalesceGlobals instead of shallow copy.

Due to this change, helm command's performance got massive improvement (both in time & memory consumption)

This change is being used within my company since a very long time (was being used in helm v3 also), so I suggest to make this change in helm v4 which will be a huge performance improvement.

Moreover, I am below sharing some performance difference while executing helm commands with/without this perf change on big bundles of my company (can not share the bundle).

<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

  | helm-package | helm-template | helm-lint | helm-install:dry | helm-install
-- | -- | -- | -- | -- | --
Helm v3 | >35min (Needs very high memory)   | >40min (Needs very high memory)  |   |>40min (Needs very high memory) |  
Helm v3-perf |   | 4min 53sec | 3min 45sec | 5min 51sec | 5min 33sec
Helm v4 |   | 36min 21sec(Needs very high memory) | 31 sec | 34min 21sec |  
Helm v4-perf | 3min (around)  | 1min 44sec | 16sec | 2min 51sec |  

</body>

</html>

**If applicable**:

### DIFF
--- a/pkg/chart/common/util/coalesce.go
+++ b/pkg/chart/common/util/coalesce.go
@@ -19,7 +19,6 @@ package util
 import (
 	"fmt"
 	"log"
-	"maps"
 
 	"helm.sh/helm/v4/internal/copystructure"
 	chart "helm.sh/helm/v4/pkg/chart"
@@ -160,7 +159,8 @@ func coalesceGlobals(printf printFn, dest, src map[string]any, prefix string, _ 
 	// tables in globals.
 	for key, val := range sg {
 		if istable(val) {
-			vv := copyMap(val.(map[string]any))
+			valCopy, _ := copystructure.Copy(val)
+			vv := valCopy.(map[string]interface{})
 			if destv, ok := dg[key]; !ok {
 				// Here there is no merge. We're just adding.
 				dg[key] = vv
@@ -189,11 +189,6 @@ func coalesceGlobals(printf printFn, dest, src map[string]any, prefix string, _ 
 	dest[common.GlobalKey] = dg
 }
 
-func copyMap(src map[string]any) map[string]any {
-	m := make(map[string]any, len(src))
-	maps.Copy(m, src)
-	return m
-}
 
 // coalesceValues builds up a values map for a particular chart.
 //
--- a/pkg/chart/common/util/coalesce.go
+++ b/pkg/chart/common/util/coalesce.go
@@ -137,8 +137,6 @@ func coalesceDeps(printf printFn, chrt chart.Charter, dest map[string]any, prefi
 }
 
 // coalesceGlobals copies the globals out of src and merges them into dest.
-//
-// For convenience, returns dest.
 func coalesceGlobals(printf printFn, dest, src map[string]any, prefix string, _ bool) error {
 	var dg, sg map[string]any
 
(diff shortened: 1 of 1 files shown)


pkg/chart/common/util/coalesce.go
```diff
@@ -160,7 +159,8 @@ func coalesceGlobals(printf printFn, dest, src map[string]any, prefix string, _
 	// tables in globals.
 	for key, val := range sg {
 		if istable(val) {
-			vv := copyMap(val.(map[string]any))
+			valCopy, _ := copystructure.Copy(val)
```
banjoh: I think we should return this error to the caller (`coalesceDeps`)

pkg/chart/common/util/coalesce.go
```diff
@@ -160,7 +159,8 @@ func coalesceGlobals(printf printFn, dest, src map[string]any, prefix string, _
...
+			vv := valCopy.(map[string]interface{})
```
banjoh:
```suggestion
			vv := valCopy.(map[string]any)
```

banjoh: Casting this can fail. I would handle this like so

```go
vv, ok := valCopy.(map[string]any)
if !ok {
	printf("warning: unable to convert globals copy to Helm values type")
	continue
}
```

banjoh (changes requested):
Thank you for your contribution. Here is my first pass after review

banjoh:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

AnubhavT007:
Hello @banjoh 

I have made few changes as suggested by you, can you please have a look.

pkg/chart/common/util/coalesce.go
AnubhavT007: Incorporated suggested change

AnubhavT007: Incorporated suggested change

pkg/chart/common/util/coalesce.go
AnubhavT007: Incorporated suggested change

PR #31848, opened 2026-02-21
fix: skip waiting for Jobs with TTL in status watcher

mehrdadbn9:
## Summary

This PR fixes a regression where the kstatus watcher in Helm 4 would get stuck waiting for Jobs that have `ttlSecondsAfterFinished` set. These Jobs are automatically deleted by the TTL controller after completion, causing the watcher to see `NotFound` status and timeout.

## Changes

- Added a case in the `wait` function to skip waiting for Jobs that have `TTLSecondsAfterFinished` set
- Added unit tests for `Wait`, `WaitWithJobs`, and `WatchUntilReady` methods to verify Jobs with TTL are properly skipped

## Technical Details

Jobs with `.spec.ttlSecondsAfterFinished` get deleted by the TTL controller after they finish. The kstatus watcher would wait for these Jobs to reach "Current" status, but after the TTL controller deletes them, the status becomes "NotFound", causing the wait to timeout.

The fix follows the same pattern used for paused Deployments - simply skip adding these resources to the watch list. This matches the behavior from the legacy Helm 3 waiter which ignored deleted hooks.

## Testing

- Added test cases for Jobs with TTL in:
  - `TestStatusWait`
  - `TestWaitForJobComplete`
  - `TestWatchForReady`
- All existing tests pass

Fixes #31786

### DIFF
--- a/pkg/kube/statuswait.go
+++ b/pkg/kube/statuswait.go
@@ -33,6 +33,7 @@ import (
 	"github.com/fluxcd/cli-utils/pkg/kstatus/watcher"
 	"github.com/fluxcd/cli-utils/pkg/object"
 	appsv1 "k8s.io/api/apps/v1"
+	batchv1 "k8s.io/api/batch/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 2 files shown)


mehrdadbn9:
Superseded by a new PR with improved fix that waits for Jobs with TTL but treats NotFound as success.

PR #31850, opened 2026-02-21
fix: handle Jobs with TTL in status watcher

mehrdadbn9:
## Summary

This PR fixes a regression where the kstatus watcher in Helm 4 would get stuck waiting for Jobs that have `ttlSecondsAfterFinished` set. These Jobs are automatically deleted by the TTL controller after completion, causing the watcher to see `NotFound` status and timeout.

## Changes

- Collects Jobs with `TTLSecondsAfterFinished` set during wait setup
- Computes `obj` once per resource to avoid duplicated work
- Passes TTL Jobs context to the `statusObserver`
- Treats `NotFound` status for Jobs with TTL as `Current` status
- This allows the wait to complete successfully when the Job is deleted by the TTL controller

## Technical Details

Jobs with `.spec.ttlSecondsAfterFinished` get deleted by the TTL controller after they finish. The kstatus watcher would wait for these Jobs to reach "Current" status, but after the TTL controller deletes them, the status becomes "NotFound", causing the wait to timeout.

This fix keeps the Jobs in the watch list (so Helm still waits for them to complete), but treats `NotFound` status for Jobs with TTL as `Current` for aggregation purposes. This matches the behavior from the legacy Helm 3 waiter which ignored deleted hooks.

## Testing

- Added `jobWithTTLManifest` test fixture
- Added `TestJobWithTTLDeletedDuringWait` test function that:
  - Tests Jobs being deleted during wait (simulating TTL controller)
  - Tests Jobs that exist and are complete (not deleted)
  - Tests all wait methods: `WatchUntilReady`, `WaitWithJobs`, `Wait`
  - Uses proper synchronization with `sync.WaitGroup` to avoid goroutine test safety issues
- All existing tests pass

Fixes #31786

### DIFF
--- a/pkg/kube/statuswait.go
+++ b/pkg/kube/statuswait.go
@@ -33,6 +33,7 @@ import (
 	"github.com/fluxcd/cli-utils/pkg/kstatus/watcher"
 	"github.com/fluxcd/cli-utils/pkg/object"
 	appsv1 "k8s.io/api/apps/v1"
+	batchv1 "k8s.io/api/batch/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/client-go/dynamic"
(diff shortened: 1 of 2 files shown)


mehrdadbn9:
Superseded by a new PR addressing reviewer feedback.

PR #31852, opened 2026-02-21
fix(sdk): correct import comment in statuswait.go from v3 to v4

### LINKED ISSUE #31846: Incorrect Go pkg import comment causing downstream processing error (kythe)
### What happened?

https://github.com/helm/helm/blob/5caf0044d4ef3d62a955440272999e139aafbbed/pkg/kube/statuswait.go#L17 is in a v4 package but it says `// import "helm.sh/helm/v3/pkg/kube"`

https://github.com/helm/helm/blob/5caf0044d4ef3d62a955440272999e139aafbbed/pkg/kube/client.go#L17 has a correct one but it would be a mismatch with above.

Kythe error:

```
2026/02/19 18:31:04 ERROR: Error locating "github.com/cilium/cilium/vendor/...": louhi_ws/gopath/src/github.com/cilium/cilium/vendor/helm.sh/helm/v4/pkg/cli/environment.go:39:2: found import comments "helm.sh/helm/v4/pkg/kube" (client.go) and "helm.sh/helm/v3/pkg/kube" (statuswait.go) in /workspace/louhi_ws/gopath/src/github.com/cilium/cilium/vendor/helm.sh/helm/v4/pkg/kube
```

All import comments match, or no import comment at all.

Irrelevant

### Helm version

Irrelevant

### Kubernetes version

Irrelevant

rohansood10:

Fixes the Go import comment in `pkg/kube/statuswait.go` from `helm.sh/helm/v3/pkg/kube` to `helm.sh/helm/v4/pkg/kube`, matching every other file in the package. The stale v3 comment causes downstream tooling errors (e.g. kythe) when the module is vendored, since two conflicting import comments exist in the same package directory.

Fixes #31846

Single-line change — only the import path comment is updated.

**If applicable**:

### DIFF
--- a/pkg/kube/statuswait.go
+++ b/pkg/kube/statuswait.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package kube // import "helm.sh/helm/v3/pkg/kube"
+package kube // import "helm.sh/helm/v4/pkg/kube"
 
 import (
 	"context"


TerryHowe:
Still a draft? I think it would be better to mark this as a partial instead of fixes

rohansood10:
Good point @TerryHowe — updated! This is now marked ready for review. Changed the wording to indicate this is a partial fix rather than a full resolution.

gjenkins8 (approved):

TerryHowe (approved):
/lgtm

PR #31858, opened 2026-02-24
chore: `go fix` fixes

gjenkins8:

Utilize `go fix` to cleanup:

- `interface{}` -> `any` (99.9% of the PR changes)
- remove spurious `json:"omitempty"` for `time.Time`
- utilize `strings.Cut(...)` instead of `strings.Index(...)` where applicable

**If applicable**:

### DIFF
--- a/internal/chart/v3/chart.go
+++ b/internal/chart/v3/chart.go
@@ -49,7 +49,7 @@ type Chart struct {
 	// Schema is an optional JSON schema for imposing structure on Values
 	Schema []byte `json:"schema"`
 	// SchemaModTime the schema was last modified
-	SchemaModTime time.Time `json:"schemamodtime,omitempty"`
+	SchemaModTime time.Time `json:"schemamodtime"`
 	// Files are miscellaneous files in a chart archive,
 	// e.g. README, LICENSE, etc.
 	Files []*common.File `json:"files"`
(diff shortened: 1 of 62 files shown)


atombrella:
https://golangci-lint.run/docs/linters/configuration/#modernize Would it be an idea to enable this in `golangci-lint.yml` to guard against regressions of `interface{}`? I imagine that is the biggest culprit, because `any` doesn't seem to have caught on.

Essentially, this would run all of the rules, and fix whatever is found inline. `-rangeint` or `-slicescontains` would run only that particular rule.
```sh
go run golang.org/x/tools/gopls/internal/analysis/modernize/cmd/modernize@latest -fix -test ./...
```

internal/chart/v3/chart.go
```diff
@@ -49,7 +49,7 @@ type Chart struct {
 	// Schema is an optional JSON schema for imposing structure on Values
 	Schema []byte `json:"schema"`
 	// SchemaModTime the schema was last modified
-	SchemaModTime time.Time `json:"schemamodtime,omitempty"`
+	SchemaModTime time.Time `json:"schemamodtime"`
```
TerryHowe, line 52: Good time to get that change in. Seems like I got in trouble for approving a change to do that in charts v2, but this is the time for v3. Any more fields like that I wonder.

TerryHowe (approved):
/lgtm

PR #31860, opened 2026-02-25
chore: update golangci-lint to v2.10.1 and enable modernize linter

mmorel-35:
**What this PR does / why we need it**:

enable [modernize](https://golangci-lint.run/docs/linters/configuration/#modernize) linter to easily identify and fix the kind of issues that have recently been treated in #31858

**Special notes for your reviewer**:

**If applicable**:
- [ ] this PR contains user facing changes (the `docs needed` label should be applied if so)
- [ ] this PR contains unit tests
- [ ] this PR has been tested for backwards compatibility

PR #31863, opened 2026-02-25
chore: enable the modernize linter

atombrella:

[Follow-up to #31858](https://github.com/helm/helm/pull/31858#issuecomment-3955422278)
I bumped the version for golangci-lint to 2.8.0. I can be newer. 2.6.0 is the version that added the `modernize` linter.

I fixed a single case of `stringscut`.
There are three warnings from `revive` related to package names that overlap with the standard library.

I'm not sure if we need to add a configuration section for the `modernize` because new rules are added with new Go releases. 

**If applicable**:


TerryHowe:
Are you closing this in favor of the other PRs?

atombrella:
It's already in #31860 No need for two of us to work on the same thing :smile:

PR #31868, opened 2026-02-26
fix: insert newline after doc separators glued to content by template trimming

matheuscscp:
Fixes: #31867

### DIFF
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -144,6 +144,39 @@ const (
 	filenameAnnotation = "postrenderer.helm.sh/postrender-filename"
 )
 
+// fixDocSeparators ensures YAML document separators ("---") are always
+// followed by a newline in rendered template content. Go template whitespace
+// trimming ({{-) can remove the newline after "---", producing e.g.
+// "---apiVersion: v1" which is not a valid YAML document separator.
+// This function inserts a newline after any "---" at the start of a line
+// that is immediately followed by non-whitespace content.
+func fixDocSeparators(content string) string {
+	var b strings.Builder
+	remaining := content
+	for {
+		// Find "---" at the start of a line (or start of content).
+		idx := strings.Index(remaining, "---")
+		if idx == -1 {
+			b.WriteString(remaining)
+			break
+		}
+		// "---" must be at the start of a line: either idx==0 or preceded by '\n'.
+		if idx > 0 && remaining[idx-1] != '\n' {
+			b.WriteString(remaining[:idx+3])
+			remaining = remaining[idx+3:]
+			continue
+		}
+		b.WriteString(remaining[:idx+3])
+		remaining = remaining[idx+3:]
+		// If "---" is followed by non-whitespace (e.g. "---apiVersion"),
+		// insert a newline to make it a proper document separator.
+		if len(remaining) > 0 && remaining[0] != '\n' && remaining[0] != '\r' && remaining[0] != ' ' && remaining[0] != '\t' {
+			b.WriteByte('\n')
+		}
+	}
+	return b.String()
+}
+
 // annotateAndMerge combines multiple YAML files into a single stream of documents,
 // adding filename annotations to each document for later reconstruction.
 func annotateAndMerge(files map[string]string) (string, error) {
@@ -158,6 +191,13 @@ func annotateAndMerge(files map[string]string) (string, error) {
 		if strings.HasPrefix(path.Base(fname), "_") || strings.TrimSpace(content) == "" {
 			continue
 		}
+
+		// Fix document separators where Go template whitespace trimming
+		// ({{-) has removed the newline after "---", producing e.g.
+		// "---apiVersion: v1" which is not a valid YAML document
+		// separator. Insert the missing newline so kio.ParseAll can
+		// parse the content correctly.
+		content = fixDocSeparators(content)
 
 		manifests, err := kio.ParseAll(content)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


matheuscscp:
@banjoh @gjenkins8 Can you please review? 4.1.2 is coming soon

mrlunchbox777 (approved):
lgtm

pkg/action/action_test.go
```diff
@@ -403,6 +403,71 @@ func (m *mockPostRenderer) Run(renderedManifests *bytes.Buffer) (*bytes.Buffer,
...
+			name:     "multiple separators some glued",
+			input:    "---apiVersion: v1\nkind: ConfigMap\n---apiVersion: v1\nkind: Service\n",
+			expected: "---\napiVersion: v1\nkind: ConfigMap\n---\napiVersion: v1\nkind: Service\n",
+		},
+		{
+			name:     "empty string",
+			input:    "",
+			expected: "",
+		},
+		{
```
TerryHowe, line 447: Are all the test cases covered here to make sure someone doesn't break this in the future. I'm not sure cases like '---' followed by '\r', ' ', and '\t' are covered. That and maybe a test case for only '----\n' and '----more'

TerryHowe (dismissed):

pkg/action/action_test.go
banjoh: Ditto

matheuscscp: Added!

pkg/action/action.go
```diff
@@ -144,6 +144,40 @@ const (
...
 
+// fixDocSeparators ensures YAML document separators ("---") are always
+// followed by a newline in rendered template content. Go template whitespace
+// trimming ({{-) can remove the newline after "---", producing e.g.
+// "---apiVersion: v1" which is not a valid YAML document separator.
+// This function inserts a newline after any "---" at the start of a line
+// that is immediately followed by non-whitespace content.
+func fixDocSeparators(content string) string {
+	var b strings.Builder
+	b.Grow(len(content) + 16)
```
scottrigby: only question is this line

matheuscscp: Removed, thanks!

pkg/action/action.go
```diff
@@ -144,6 +144,39 @@ const (
...
+		if idx > 0 && remaining[idx-1] != '\n' {
+			b.WriteString(remaining[:idx+3])
+			remaining = remaining[idx+3:]
+			continue
+		}
+		b.WriteString(remaining[:idx+3])
+		remaining = remaining[idx+3:]
+		// If "---" is followed by non-whitespace (e.g. "---apiVersion"),
+		// insert a newline to make it a proper document separator.
+		if len(remaining) > 0 && remaining[0] != '\n' && remaining[0] != '\r' && remaining[0] != ' ' && remaining[0] != '\t' {
```
scottrigby, line 173: nice. note that this conforms to go's definition of whitespace trimming in templating context: https://pkg.go.dev/text/template#hdr-Text_and_spaces
>  For this trimming, the definition of white space characters is the same as in Go: space, horizontal tab, carriage return, and newline.

scottrigby (approved):
@matheuscscp thanks for the fix. Note that this is a Helm 4 only regression from Helm 3, so we should ship this in the next Helm 4 patch release.

Also addressed change requested by @TerryHowe and @banjoh in https://github.com/helm/helm/pull/31868#discussion_r2895855841.

As a follow-up, I would however like to propose that in Charts apiVersion v3 we drop support for special handling around gotemplate whitespace trimming for yaml separators. In general, when a `{{-` is used, all leading white space should be trimmed from the immediately following text - including "space, horizontal tab, carriage return, and newline" (see https://pkg.go.dev/text/template#hdr-Text_and_spaces). I do not yet know why we handle this special case as opposed to asking chart authors to fix this. We do have to support this for existing apiVersion v2 charts for backwards compatibility (both with Helm 3 and Helm 4), but Charts apiVersion v3 could simplify our code by not handling this in a special way for YAML separators. We would need to add this to a charts v2 -> v3 migration guide if we do.

pkg/action/action.go
```diff
@@ -144,6 +144,40 @@ const (
...
+			b.WriteByte('\n')
+		}
+	}
```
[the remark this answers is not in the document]
scottrigby: addressed

banjoh (approved):
LGTM

TerryHowe (commented):
/lgtm

matheuscscp:
> @matheuscscp thanks for the fix. Note that this is a Helm 4 only regression from Helm 3, so we should ship this in the next Helm 4 patch release.
> 
> Also addressed change requested by @TerryHowe and @banjoh in [#31868 (comment)](https://github.com/helm/helm/pull/31868#discussion_r2895855841).
> 
> As a follow-up, I would however like to propose that in Charts apiVersion v3 we drop support for special handling around gotemplate whitespace trimming for yaml separators. In general, when a `{{-` is used, all leading white space should be trimmed from the immediately following text - including "space, horizontal tab, carriage return, and newline" (see https://pkg.go.dev/text/template#hdr-Text_and_spaces). I do not yet know why we handle this special case as opposed to asking chart authors to fix this. We do have to support this for existing apiVersion v2 charts for backwards compatibility (both with Helm 3 and Helm 4), but Charts apiVersion v3 could simplify our code by not handling this in a special way for YAML separators. We would need to add this to a charts v2 -> v3 migration guide if we do.

@scottrigby We discussed in the Flux dev meeting today about this breaking change in Chart API v3. It was noted that this kind of behavior should not happen only when using postrenderers. The behavior fixed by this PR should be caught as an error for example on `helm template`, as the charts committing this abuse do not expect any form of postrendering at all, so the behavior with or without postrenderers must be the same.

scottrigby:
> he behavior with or without postrenderers must be the same

@matheuscscp yes, without question 👍 agree that any discrepancy between the two would be a bug, regardless of the expected behavior.

PR #31884, opened 2026-02-27
Properly format the extra field in gzipped packages

### LINKED ISSUE #31844: `helm package` produces malformed `.tgz`
### What happened?

When you run `helm package`, it produces a `.tgz` file that that works fine with GNU `tar` but is technically non-conforming to [RFC 1952 section 2.3.1.1](https://www.rfc-editor.org/rfc/rfc1952.html#page-8), and is rejected by some stickler Gzip implementations, such as the one that the Bazel downloader uses.

I tried to download [external-dns-1.20.0.tgz](https://github.com/kubernetes-sigs/external-dns/releases/tag/external-dns-helm-chart-1.20.0) using Bazel:

```bazel
# MODULE.bazel
http_archive(
    name = "external-dns-helm-chart",
    url = "https://github.com/kubernetes-sigs/external-dns/releases/download/external-dns-helm-chart-1.20.0/external-dns-1.20.0.tgz",
    integrity = "sha384-pfKzHGEwOh+i3LTLVS5BCg0kyIhxvqiMNmRMqVXiDB/sKp0++xb7F6onMuBlVjDN",
)
```

I got the following error:

```
Error in download_and_extract: java.io.IOException: Error extracting /home/will/.cache/bazel/_bazel_will/b9daf2201169cb22dce5a04a7c219ba1/external/+http_archive+external-dns-helm-chart/temp7900067191862921420/external-dns-1.20.0.tgz to /home/will/.cache/bazel/_bazel_will/b9daf2201169cb22dce5a04a7c219ba1/external/+http_archive+external-dns-helm-chart/temp7900067191862921420: Extra subfield lenght exceeds remaining bytes in extra: 21064 > 37
```

I expected it to extract the archive successfully, the way GNU `tar` does it.

Take any archive produced by `helm package`. You can use [this one](https://github.com/kubernetes-sigs/external-dns/releases/download/external-dns-helm-chart-1.20.0/external-dns-1.20.0.tgz) from kubernetes-sigs/external-dns, which originally motivated this issue, or you can build one with any version of Helm. I tried it with versions 3.11.3 - 4.1.1.

Given that archive, this Python script illustrates the problem by poking around the header:

```python
from struct import unpack
from urllib.request import urlopen

# Read an archive as raw bytes.
url = "https://github.com/kubernetes-sigs/external-dns/releases/download/external-dns-helm-chart-1.20.0/external-dns-1.20.0.tgz"
with urlopen(url) as r:
    archive = r.read()

# Gzip header (RFC 1952): ID1 ID2 CM FLG MTIME[4] XFL OS [XLEN[2] if FEXTRA]
flg = archive[3]
assert flg & 0x04, "FEXTRA flag not set"

xlen = unpack('<H', archive[10:12])[0]
extra = archive[12:12 + xlen]
print(f"XLEN = {xlen}")
print(f"Extra field bytes: {extra.hex()}")

offset = 0
while offset < len(extra):
    si1, si2 = extra[offset], extra[offset + 1]
    sublen = unpack('<H', extra[offset + 2:offset + 4])[0]
    remaining = len(extra) - offset - 4
    print(f"Subfield SI1=0x{si1:02x} SI2=0x{si2:02x} LEN={sublen}")
    print(f"Remaining bytes after subfield header: {remaining}")
    if sublen > remaining:
        print(f"ERROR: subfield length {sublen} exceeds remaining bytes {remaining}")
        break
    offset += 4 + sublen
```

You should see `ERROR: subfield length 21064 exceeds remaining bytes 37`, similar to the Bazel error.

It looks like this is due to [logic in `save.go`](https://github.com/helm/helm/blob/v4.1.1/pkg/chart/v2/util/save.go#L135) that saves the static blob [`+aHR0cHM6Ly95b3V0dS5iZS96OVV6MWljandyTQo=`](https://github.com/helm/helm/blob/v4.1.1/pkg/chart/v2/util/save.go#L36) in the extra field. That extra field does not conform to [RFC 1952 section 2.3.1.1](https://www.rfc-editor.org/rfc/rfc1952.html#page-8) which says that the extra field must consist of 4 subsections: `SI1`, `SI2`, `LEN`, and then the raw data. It looks like something Base64-encoded but it's undocumented. The second 2 bytes are `HR`, and those are interpreted as a 16-byte little-endian integer i.e. `0x5248` i.e. 21064. Then the rest of the blob after the 4-byte header is 37 bytes.

I'm not sure what that blob is, but it should have the proper format for the extra field. It needs those `SI1`, `SI2`, and `LEN` subsections.

### Helm version

<details>

```console
$ helm version
# paste output here
```
</details>

### Kubernetes version

<details>

```console
$ kubectl version
# paste output here
```

</details>

ouillie:
closes #31844

This retains the original intent "to add a string to an archive so that I could computationally check whether the archive had been generated by Helm".

### DIFF
--- a/pkg/chart/v2/util/save.go
+++ b/pkg/chart/v2/util/save.go
@@ -33,7 +33,7 @@ import (
 	chart "helm.sh/helm/v4/pkg/chart/v2"
 )
 
-var headerBytes = []byte("+aHR0cHM6Ly95b3V0dS5iZS96OVV6MWljandyTQo=")
+var headerBytes = []byte("rr(\x00aHR0cHM6Ly95b3V0dS5iZS96OVV6MWljandyTQo=")
 
 // SaveDir saves a chart as files in a directory.
 //
--- a/pkg/chart/v2/util/save.go
+++ b/pkg/chart/v2/util/save.go
@@ -33,7 +33,11 @@ import (
 	chart "helm.sh/helm/v4/pkg/chart/v2"
 )
 
-var headerBytes = []byte("rr(\x00aHR0cHM6Ly95b3V0dS5iZS96OVV6MWljandyTQo=")
+// RFC 1952 subfield header:
+// +---+---+---+---+==================================+
+// |SI1|SI2|  LEN  |... LEN bytes of subfield data ...|
+// +---+---+---+---+==================================+
+var headerBytes = []byte("rr\x28\x00aHR0cHM6Ly95b3V0dS5iZS96OVV6MWljandyTQo=")
 
 // SaveDir saves a chart as files in a directory.
 //
--- a/pkg/chart/v2/util/save_test.go
+++ b/pkg/chart/v2/util/save_test.go
@@ -130,6 +130,51 @@ func TestSave(t *testing.T) {
 	}
 }
 
+func TestSavedGzipExtraFieldIsValid(t *testing.T) {
+	tmp := t.TempDir()
+	c := &chart.Chart{
+		Metadata: &chart.Metadata{
+			APIVersion: chart.APIVersionV2,
+			Name:       "ahab",
+			Version:    "1.2.3",
+		},
+	}
+
+	where, err := Save(c, tmp)
+	if err != nil {
+		t.Fatalf("Failed to save: %s", err)
+	}
+
+	f, err := os.Open(where)
+	if err != nil {
+		t.Fatalf("Failed to open saved file: %s", err)
+	}
+	defer f.Close()
+
+	r, err := gzip.NewReader(f)
+	if err != nil {
+		t.Fatalf("Failed to create gzip reader: %s", err)
+	}
+	defer r.Close()
+
+	// RFC 1952 §2.3.1.1:
+	// Each subfield consists of SI1, SI2 (1 byte each),
+	// a 2-byte little-endian LEN, and LEN bytes of data.
+	// https://www.rfc-editor.org/rfc/rfc1952.html#page-8
+	// https://github.com/helm/helm/issues/31844
+	extra := r.Extra
+	if len(extra) > 0 {
+		if len(extra) < 4 {
+			t.Fatalf("gzip extra field too short to contain a valid subfield: %d byte(s)", len(extra))
+		}
+		dataLen := int(extra[2]) | int(extra[3])<<8
+		// Assume a single subfield
+		if len(extra) != 4+dataLen {
+			t.Errorf("gzip extra field has malformed subfield: LEN=%d but %d data byte(s) follow the subfield header", dataLen, len(extra)-4)
+		}
+	}
+}
+
 // Creates a copy with a different schema; does not modify anything.
 func withSchema(chart chart.Chart, schema []byte) chart.Chart {
 	chart.Schema = schema
@@ -299,7 +344,7 @@ func TestRepeatableSave(t *testing.T) {
 				Schema:        []byte("{\n  \"title\": \"Values\"\n}"),
 				SchemaModTime: modTime,
 			},
-			want: "fea2662522317b65c2788ff9e5fc446a9264830038dac618d4449493d99b3257",
+			want: "63358874b93ea095c857cd66bcf5d0a4464840cf84a07547db744d81d6c5af59",
 		},
 		{
 			name: "Package 2 files",
@@ -321,7 +366,7 @@ func TestRepeatableSave(t *testing.T) {
 				Schema:        []byte("{\n  \"title\": \"Values\"\n}"),
 				SchemaModTime: modTime,
 			},
-			want: "7ae92b2f274bb51ea3f1969e4187d78cc52b5f6f663b44b8fb3b40bcb8ee46f3",
+			want: "c2a43990053da788ad4e260d3b00d52a0b103ccc67ab9f48278a7b6dcfb2a4bd",
 		},
 	}
 	for _, test := range tests {
--- a/pkg/chart/v2/util/save_test.go
+++ b/pkg/chart/v2/util/save_test.go
@@ -21,6 +21,7 @@ import (
 	"bytes"
 	"compress/gzip"
 	"crypto/sha256"
+	"encoding/binary"
 	"encoding/hex"
 	"errors"
 	"io"
@@ -35,6 +36,8 @@ import (
 	"helm.sh/helm/v4/pkg/chart/common"
 	chart "helm.sh/helm/v4/pkg/chart/v2"
 	"helm.sh/helm/v4/pkg/chart/v2/loader"
+
+	"github.com/stretchr/testify/require"
 )
 
 func TestSave(t *testing.T) {
@@ -130,6 +133,7 @@ func TestSave(t *testing.T) {
 	}
 }
 
+// https://github.com/helm/helm/issues/31844
 func TestSavedGzipExtraFieldIsValid(t *testing.T) {
 	tmp := t.TempDir()
 	c := &chart.Chart{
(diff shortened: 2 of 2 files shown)


ouillie:
LMK if you like Copilot's suggestions, although my hunch is YAGNI.

TerryHowe:
I don't think copilots comments are useful

TerryHowe (commented):
Can we get this on charts v3 too?

TerryHowe:
@gjenkins8 any concerns with the somewhat breaking nature of this fix? I classify it is a bug, so I think it should happen.

pkg/chart/v2/util/save_test.go
```diff
@@ -130,6 +130,50 @@ func TestSave(t *testing.T) {
...
+	// a 2-byte little-endian LEN, and LEN bytes of data.
+	// https://www.rfc-editor.org/rfc/rfc1952.html#page-8
+	// https://github.com/helm/helm/issues/31844
+	extra := r.Extra
+	if len(extra) > 0 {
+		if len(extra) < 4 {
+			t.Fatalf("gzip extra field too short to contain a valid subfield: %d byte(s)", len(extra))
+		}
+		dataLen := int(extra[2]) | int(extra[3])<<8
+		if len(extra) != 4+dataLen {
```
gjenkins8:
```suggestion
        // Assume a single subfield
		if len(extra) != 4+dataLen {
```

gjenkins8: I think comment the assumption regarding the single subfield (to the point of co-pilots review). I agree we don't need to write full validation for multiple subfields

ouillie:
One person's bug is another person's feature, but I agree that this one is best thought of as a bug.

anicoletta:
Hi @ouillie,

I came across this PR while investigating the gzip packaging issue described in #31844.

First of all, thanks for the contribution! We’ve actually encountered the issue described in #31844 on our side, and it is currently blocking one of our workflows due to failures in stricter archive validation.

I noticed there hasn’t been any activity on this PR for a while, so I just wanted to kindly check whether there are any updates or if additional input is needed.

Thanks again for the work!

pkg/chart/v2/util/save.go
```diff
@@ -33,7 +33,7 @@ import (
 	chart "helm.sh/helm/v4/pkg/chart/v2"
 )
 
-var headerBytes = []byte("+aHR0cHM6Ly95b3V0dS5iZS96OVV6MWljandyTQo=")
+var headerBytes = []byte("rr(\x00aHR0cHM6Ly95b3V0dS5iZS96OVV6MWljandyTQo=")
```
gjenkins8:
```suggestion
// RFC 1952 subfield header: 
// +---+---+---+---+==================================+
// |SI1|SI2|  LEN  |... LEN bytes of subfield data ...|
// +---+---+---+---+==================================+
var headerBytes = []byte("rr\x29\x00aHR0cHM6Ly95b3V0dS5iZS96OVV6MWljandyTQo=")
```

gjenkins8: also note, I modified the LEN bytes to `\x29\x00`, to make the LEN a more obvious numerical value)

pkg/chart/v2/util/save_test.go
```diff
@@ -130,6 +130,51 @@ func TestSave(t *testing.T) {
 	}
 }
 
+func TestSavedGzipExtraFieldIsValid(t *testing.T) {
```
gjenkins8, line 137: I'm not really sure this test is proving anything - we don't need to show `gzip.NewWriter(f).Extra = headerBytes` "works"

gjenkins8: By this, I mean, I wonder if we should actually include this test?

ouillie: The test is useful to guard against potential regressions in the future if someone decides to do something else with extra fields. I suppose the ideal test would use Apache Commons Compress to try to read the generated gzip, since I'm pretty sure that's the library Bazel was using due to the [distinctive typo (lenght) in the error message](https://github.com/apache/commons-compress/commit/7ef0f5be9feded9075fbf04d27d1b1bf1d1734de#diff-e4bf9cdf8cf7c2aa9a5d5723f17134b7d29502189d18d623d9e27f0945f81814) but that would be a bit heavy-weight to introduce to a Go unit test, so I figured this much simpler native Go check was a good compromise.

Anyway, if you want to get rid of it, fine by me, but if you're asking for my opinion, I think we might as well include it.

pkg/chart/v2/util/save_test.go
```diff
@@ -130,6 +130,50 @@ func TestSave(t *testing.T) {
...
+	// https://github.com/helm/helm/issues/31844
+	extra := r.Extra
+	if len(extra) > 0 {
+		if len(extra) < 4 {
+			t.Fatalf("gzip extra field too short to contain a valid subfield: %d byte(s)", len(extra))
+		}
+		dataLen := int(extra[2]) | int(extra[3])<<8
+		if len(extra) != 4+dataLen {
+			t.Errorf("gzip extra field has malformed subfield: LEN=%d but %d data byte(s) follow the subfield header", dataLen, len(extra)-4)
+		}
```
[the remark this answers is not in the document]
gjenkins8: out of scope IMHO

pkg/chart/v2/util/save_test.go
gjenkins8: sure, we can keep (I guess I am trying to say wouldn't ask for it to be written, if I was reviewing this PR otherwise. But it has been written, and does potentially help)

gjenkins8 (commented):
can you fix the lint issue please (I'm not sure why it thinks the line is not correctly formatted?)

pkg/chart/v2/util/save.go
```diff
@@ -33,7 +33,11 @@ import (
 	chart "helm.sh/helm/v4/pkg/chart/v2"
 )
 
-var headerBytes = []byte("+aHR0cHM6Ly95b3V0dS5iZS96OVV6MWljandyTQo=")
+// RFC 1952 subfield header: 
+// +---+---+---+---+==================================+
+// |SI1|SI2|  LEN  |... LEN bytes of subfield data ...|
+// +---+---+---+---+==================================+
+var headerBytes = []byte("rr\x29\x00aHR0cHM6Ly95b3V0dS5iZS96OVV6MWljandyTQo=")
```
gjenkins8:
```suggestion
var headerBytes = []byte("rr\x28\x00aHR0cHM6Ly95b3V0dS5iZS96OVV6MWljandyTQo=")
```

It looks like I miscalculated the length:
```
--- FAIL: TestSavedGzipExtraFieldIsValid (0.00s)
    save_test.go:173: gzip extra field has malformed subfield: LEN=41 but 40 data byte(s) follow the subfield header
--- FAIL: TestRepeatableSave (0.00s)
```

pkg/chart/v2/util/save_test.go
```diff
@@ -130,6 +130,51 @@ func TestSave(t *testing.T) {
...
+	}
+	defer r.Close()
+
+	// RFC 1952 §2.3.1.1:
+	// Each subfield consists of SI1, SI2 (1 byte each),
+	// a 2-byte little-endian LEN, and LEN bytes of data.
+	// https://www.rfc-editor.org/rfc/rfc1952.html#page-8
+	// https://github.com/helm/helm/issues/31844
+	extra := r.Extra
+	if len(extra) > 0 {
```
gjenkins8: `testify/require` is preferred:

```suggestion
	require.NotEmpty(t, extra)
```

pkg/chart/v2/util/save_test.go
```diff
@@ -130,6 +130,51 @@ func TestSave(t *testing.T) {
...
+		if len(extra) < 4 {
```
gjenkins8:
```suggestion
		require.GreaterOrEqual(t, len(extra), 4)
```

pkg/chart/v2/util/save_test.go
```diff
@@ -130,6 +130,51 @@ func TestSave(t *testing.T) {
...
+			t.Fatalf("gzip extra field too short to contain a valid subfield: %d byte(s)", len(extra))
+		}
+		dataLen := int(extra[2]) | int(extra[3])<<8
```
gjenkins8:
```suggestion
		dataLen := binary.BigEndian.Uint16(data[2:3)
```

(I think I have the syntax correct here)

pkg/chart/v2/util/save_test.go
```diff
@@ -130,6 +130,51 @@ func TestSave(t *testing.T) {
...
+		// Assume a single subfield
+		if len(extra) != 4+dataLen {
```
gjenkins8:
```suggestion
		assert.Lenf(t, extra, 4+dataLen, "gzip extra field has malformed subfield: LEN=%d but %d data byte(s) follow the subfield header", dataLen, len(extra)-4)
```

pkg/chart/v2/util/save_test.go
gjenkins8: I do think if we keep the test, we should expect (require) that the tar has the Extra field

ouillie:
Sorry about the force-push. Had to do it because I'd accidentally push commits without the signed-off-by tag, so I had to edit the messages. Figured I'd condense all the accepted suggestions into a single commit while at it.

gjenkins8 (approved):

TerryHowe (approved):
/lgtm

TerryHowe (dismissed):
Wait a second, something I need to look into here.

PR #31828, opened 2026-02-11
`--all`/`-a` flag added to `helm list` command

Mujib-Ahasan:
**What this PR does / why we need it**:
This PR has re-added `--all`/`-a` flag to `helm list` command.  In v4, helm list was changed to show all releases by default, making the --all/-a CLI flag obsolete. But removal of the flag was a breaking change that was not called out in the release notes. 

**Special notes for your reviewer**:
Backwards compatibility has been taken care.

closes #31784 
**If applicable**:
- [x] this PR has been tested for backwards compatibility

PR #31886, opened 2026-02-27
fix(list): reintroduce deprecated --all/-a flag for backward compatibility

### LINKED ISSUE #31784: Breaking change to helm list cli command in v4

In v4, `helm list` was changed to show all releases by default, making the `--all/-a` CLI flag obsolete.
#31295
But removal of the flag was a breaking change that was not called out in the release notes.
 
Can the `-a/--all` flag be reintroduced with a proper deprecation warning? Behaviorally it could be a no-op.

### Why is this needed?

Existing scripts that use this helm cli option must be modified to upgrade to helm v4.

Care was taken to make other cli changes (renamed options etc.) non-breaking, continuing support with old options, so I don't think it was an intentional decision to make a breaking change here.

umut-polat:
Reintroduces --all/-a for helm list as a deprecated no-op with a warning. In v4, helm list shows all releases by default, making this flag unnecessary - but removing it broke existing scripts. Follows the same MarkDeprecated pattern used for --force and --atomic.

Fixes #31784


benoittgt:
Dup of https://github.com/helm/helm/pull/31828 ?

PR #31887, opened 2026-02-27
feat(list): warn on 'helm list' truncation with default --max limit

john-weathers:

This PR adds a warning to `helm list` when results are truncated due to the default `--max` limit (only when the user has not explicitly set `--max`).

Currently, `helm list` silently truncates results (default max=256), which can make it appear as if releases are missing. This change improves clarity by informing users that more results may exist.

The warning:
- is printed to stderr
- only applies to table (default) output
- does not affect JSON/YAML or other machine-readable formats
- is only shown when truncation is caused by the default limit

My intention in printing the warning to stderr and limiting it to table output was to preserve backward compatibility of structured and machine-readable output.

Closes #31617

- Implemented via a `Truncated` field in the action `List` struct
- No changes to existing CLI flags or output formats
- Manually tested with large (>256) release sets using `--max` and `--offset`
- Unit tests added to cover truncation scenarios

**If applicable**:

### DIFF
--- a/pkg/action/list.go
+++ b/pkg/action/list.go
@@ -153,6 +153,8 @@ func (l *List) Run() ([]ri.Releaser, error) {
 		return nil, err
 	}
 
+	l.Truncated = false
+
 	var filter *regexp.Regexp
 	if l.Filter != "" {
 		var err error
--- a/pkg/action/list.go
+++ b/pkg/action/list.go
@@ -135,6 +135,8 @@ type List struct {
 	Failed       bool
 	Pending      bool
 	Selector     string
+	// Truncated indicates results were omitted due to Limit
+	Truncated bool
 }
 
 // NewList constructs a new *List
@@ -213,7 +215,7 @@ func (l *List) Run() ([]ri.Releaser, error) {
 	}
 
 	// Calculate the limit and offset, and then truncate results if necessary.
-	limit := len(results)
+	limit := len(rresults)
 	if l.Limit > 0 && l.Limit < limit {
 		limit = l.Limit
 	}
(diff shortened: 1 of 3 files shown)


pkg/action/list.go
```diff
@@ -213,14 +215,16 @@ func (l *List) Run() ([]ri.Releaser, error) {
 	}
 
 	// Calculate the limit and offset, and then truncate results if necessary.
-	limit := len(results)
+	limit := len(rresults)
```
john-weathers, line 220: I updated this to use len(rresults) since the subsequent pagination operates on the filtered and sorted rresults. Please let me know if the original use of results was intentional.

promptless-for-oss:

New suggestion: **[Document helm list --max and --offset flags in CheatSheet](https://github.com/helm/helm-www/pull/2026)**

---
_Tip: Configure how Promptless handles changelogs in [Agent Settings](https://app.gopromptless.ai/configure/settings) 📋_

TerryHowe (approved):
/lgtm

john-weathers:
Hi @gjenkins8, just wanted to follow up here since you helped triage the issue associated with this. It looks like this PR needs an additional review. I wanted to check if you could take a look or if you could suggest someone for a second review. Thanks!

StantonMatt (commented):
I took a local validation pass on `1f77d24a`.

The focused limit/truncation tests pass:

```sh
go test ./pkg/action -run 'TestList_(Truncation|Limit|LimitOffset|LimitOffsetOutOfBounds)$' -count=1
go test ./pkg/cmd -run 'TestList(Cmd|OutputFormats|StatusMapping|CustomTimeFormat)?$' -count=1
```

I also ran the touched packages together:

```sh
go test ./pkg/action ./pkg/cmd -count=1
```

That passed as well. To check the command-level behavior directly, I ran a temporary local probe with 257 in-memory deployed releases: plain `list` emitted the new default-limit warning, while `list --max 256` and `list --output json` did not. I removed the probe afterward, and `git diff --check origin/main...HEAD` is clean.

The current shape looks right to me: the action layer tracks whether pagination actually omitted results, the command only warns for table output when the user did not explicitly set `--max`, and the warning goes through Cobra's stderr stream. I do not see a blocker from this pass.

PR #31895, opened 2026-03-03
feat: Allow empty output from PostRenderer when input is empty

simonklb:

Closes #31894

**If applicable**:

### DIFF
--- a/pkg/postrenderer/postrenderer.go
+++ b/pkg/postrenderer/postrenderer.go
@@ -61,6 +61,8 @@ type postRendererPlugin struct {
 
 // Run implements PostRenderer by using the plugin's Runtime
 func (r *postRendererPlugin) Run(renderedManifests *bytes.Buffer) (*bytes.Buffer, error) {
+	renderedManifestsLen := renderedManifests.Len()
+
 	input := &plugin.Input{
 		Message: schema.InputMessagePostRendererV1{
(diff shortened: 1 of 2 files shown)


pkg/postrenderer/postrenderer.go
```diff
@@ -74,9 +76,9 @@ func (r *postRendererPlugin) Run(renderedManifests *bytes.Buffer) (*bytes.Buffer
 
 	outputMessage := output.Message.(schema.OutputMessagePostRendererV1)
 
-	// If the binary returned almost nothing, it's likely that it didn't
-	// successfully render anything
-	if len(bytes.TrimSpace(outputMessage.Manifests.Bytes())) == 0 {
+	// If the binary returned almost nothing and the input isn't empty, it's
+	// likely that it didn't successfully render anything
+	if len(bytes.TrimSpace(outputMessage.Manifests.Bytes())) == 0 && renderedManifestsLen != 0 {
```
gjenkins8, line 81:
```suggestion
	if len(bytes.TrimSpace(outputMessage.Manifests.Bytes())) == 0 && renderedManifests.Len() != 0 {
```

no need to declare a variable specifically for this I think

simonklb: I think so because it's a buffer so when it's read it's going to be empty, right?

PR #31897, opened 2026-03-03
fix: bump fluxcd/cli-utils to v0.37.2-flux.1

TerryHowe:
## Summary

- Bumps `fluxcd/cli-utils` from `v0.37.1-flux.1` to `v0.37.2-flux.1`
- Includes upstream fix [fluxcd/cli-utils#22](https://github.com/fluxcd/cli-utils/pull/22) which excludes transient pod failures from controller status

## Problem

Since Helm v4.1.1 (specifically [PR #31730](https://github.com/helm/helm/pull/31730)), `FailedStatus` is treated as a terminal state in the watcher wait strategy. The kstatus library marks pods as `Failed` after just 15 seconds of being `Unschedulable`, then escalates that to the parent StatefulSet/Deployment. This causes `helm upgrade --install` to fail prematurely when:

- Cluster autoscalers (Karpenter, Cluster Autoscaler) need time to provision nodes
- Pods are being deleted during rolling updates

The `legacy` wait strategy is unaffected because it only checks replica counts, never pod-level status.

## Fix

The upstream `cli-utils` fix ([commit `28dac90`](https://github.com/fluxcd/cli-utils/commit/28dac9039c0e3adc374433513330f0adee504f94)) adds `isTransientPodFailure()` to the `podControllerStatusReader` which filters out:
- Pods that are `Unschedulable` (waiting for node provisioning)
- Pods that are being deleted (rolling update in progress)

These transient states no longer cause the parent controller to be marked as `Failed`, so Helm correctly continues waiting until timeout.

Fixes #31896

## Test plan

- [x] `make build` succeeds
- [x] `go test ./pkg/kube/...` passes
- [x] `go test ./pkg/action/...` passes
- [x] `go test ./internal/statusreaders/...` passes
- [ ] Manual verification: `helm upgrade --install` with a StatefulSet on a cluster using Karpenter where nodes need provisioning


matheuscscp (approved):
LGTM!

scottrigby (approved):

TerryHowe:
govuln failure is unrelated to this PR

scottrigby:
agreed:
```console
% go mod why -m go.opentelemetry.io/otel/sdk
# go.opentelemetry.io/otel/sdk
helm.sh/helm/v4/pkg/repo/v1/repotest
github.com/distribution/distribution/v3/registry
github.com/distribution/distribution/v3/tracing
go.opentelemetry.io/otel/sdk/resource
```

PR #31899, opened 2026-03-03
fix: race condition in FailingKubeClient.GetWaiterWithOptions

TerryHowe:
## Summary
- Fixes data race in `FailingKubeClient.GetWaiterWithOptions` where concurrent goroutines append to the `RecordedWaitOptions` slice without synchronization
- Adds a mutex to protect the slice and exposes a thread-safe `GetRecordedWaitOptions()` accessor
- Updates all test callers to use the new accessor

The race is triggered by `TestInstallRelease_RollbackOnFailure_Interrupted` and `TestUpgradeRelease_Interrupted_RollbackOnFailure` when run with `-race` (as `make test` does).

## Test plan
- [x] `make test` passes with zero failures (previously 2 race failures)
- [x] Targeted `-race` tests pass: `TestInstallRelease_RollbackOnFailure_Interrupted`, `TestUpgradeRelease_Interrupted_RollbackOnFailure`

### DIFF
--- a/pkg/action/hooks_test.go
+++ b/pkg/action/hooks_test.go
@@ -490,5 +490,5 @@ data:
 	is.NoError(err)
 
 	// Verify that WaitOptions were passed to GetWaiter
-	is.NotEmpty(failer.RecordedWaitOptions, "WaitOptions should be passed to GetWaiter")
+	is.NotEmpty(failer.GetRecordedWaitOptions(), "WaitOptions should be passed to GetWaiter")
 }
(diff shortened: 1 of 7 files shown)


benoittgt (approved):
LGTM

TerryHowe:
Closing as superseded by #31925, which landed the same race-condition fix on `main` (added a `sync.Mutex` guarding `RecordedWaitOptions`). The race this PR targeted is already resolved upstream, so this change is no longer needed.

PR #31901, opened 2026-03-04
fix: restore (deprecated) --atomic flag on install command for backwards compatibility

### LINKED ISSUE #31900: helm install with --atomic no longer works in Helm 4
### What happened?

In Helm 4 the atomic flag was renamed to `rollback-on-failure` in https://github.com/helm/helm/pull/13629 and the original `atomic` flag was marked as deprecated, this worked for helm upgrade because the original flag binding was left in place but in the helm install command the flag binding was removed so if you attempt to use `helm install --atomic` it returns the following error `Error: unknown flag: --atomic`

See the difference between the flag bindings in the commands 
[Helm Install](https://github.com/helm/helm/blob/df8e0301edf560a8c11c77ade830e38b2918c985/pkg/cmd/install.go#L203-L204) 
```
	f.BoolVar(&client.RollbackOnFailure, "rollback-on-failure", false, "if set, Helm will rollback (uninstall) the installation upon failure. The --wait flag will be default to \"watcher\" if --rollback-on-failure is set")
	f.MarkDeprecated("atomic", "use --rollback-on-failure instead")
```

[Helm Upgrade](https://github.com/helm/helm/blob/df8e0301edf560a8c11c77ade830e38b2918c985/pkg/cmd/upgrade.go#L290-L292)
```
	f.BoolVar(&client.RollbackOnFailure, "rollback-on-failure", false, "if set, Helm will rollback the upgrade to previous success release upon failure. The --wait flag will be defaulted to \"watcher\" if --rollback-on-failure is set")
	f.BoolVar(&client.RollbackOnFailure, "atomic", false, "deprecated")
	f.MarkDeprecated("atomic", "use --rollback-on-failure instead")
```

Helm install with the atomic flag should warn about the deprecated flag and still install the helm chart

1. Run `helm install --atomic sql-server bitnami/mysql`
2. Observe the error: `Error: unknown flag: --atomic`
3. Run `helm upgrade --install --atomic sql-server bitnami/mysql` 
4. Observe that the install continues

### Helm version

<details>

```console
$ helm version
version.BuildInfo{Version:"v4.1.1", GitCommit:"5caf0044d4ef3d62a955440272999e139aafbbed", GitTreeState:"clean", GoVersion:"go1.25.7", KubeClientVersion:"v1.35"}
```
</details>

### Kubernetes version

<details>

```console
$ kubectl version
Client Version: v1.35.2
Kustomize Version: v5.7.1
Server Version: v1.34.4-eks-3a10415
```

</details>

tleed5:
Fixes https://github.com/helm/helm/issues/31900 

In https://github.com/helm/helm/pull/13629 the --atomic flag was marked as deprecated for both the install and upgrade command however in the install command the flag binding was removed so the atomic flag no longer works and returns an error `Error: unknown flag: --atomic` when it should just alert that the flag is deprecated instead of failing the entire install

I didn't add any unit tests but if desired I can add some

**If applicable**:

### DIFF
--- a/pkg/cmd/install.go
+++ b/pkg/cmd/install.go
@@ -201,6 +201,7 @@ func addInstallFlags(cmd *cobra.Command, f *pflag.FlagSet, client *action.Instal
 	f.BoolVar(&client.DependencyUpdate, "dependency-update", false, "update dependencies if they are missing before installing the chart")
 	f.BoolVar(&client.DisableOpenAPIValidation, "disable-openapi-validation", false, "if set, the installation process will not validate rendered templates against the Kubernetes OpenAPI Schema")
 	f.BoolVar(&client.RollbackOnFailure, "rollback-on-failure", false, "if set, Helm will rollback (uninstall) the installation upon failure. The --wait flag will be default to \"watcher\" if --rollback-on-failure is set")
+	f.BoolVar(&client.RollbackOnFailure, "atomic", false, "deprecated")
 	f.MarkDeprecated("atomic", "use --rollback-on-failure instead")
 	f.BoolVar(&client.SkipCRDs, "skip-crds", false, "if set, no CRDs will be installed. By default, CRDs are installed if not already present")
 	f.BoolVar(&client.SubNotes, "render-subchart-notes", false, "if set, render subchart notes along with the parent")


gjenkins8 (approved):

TerryHowe (approved):
/lgtm

PR #31910, opened 2026-03-06
fix(action): normalize glued YAML doc separators before post-render parsing

LarytheLord:
## Summary
- normalize compact YAML document separators in `annotateAndMerge()` before calling `kio.ParseAll()`
- handle rendered content like `---apiVersion: ...` (produced by `---` followed by `{{- include ... }}` in templates)
- add regression coverage for both file-start and between-document glued separator cases

## Problem
In Helm v4 post-render flow, `annotateAndMerge()` parses rendered file content with `kio.ParseAll()`. When template whitespace trimming glues content to a document marker (for example `---apiVersion: v1`), YAML parsing treats it as a regular key and rewrites the manifest incorrectly.

## Fix
Before parsing, normalize line-start `---` markers that are immediately followed by non-whitespace into `---\n...`. This preserves intended document boundaries and avoids manifest corruption in post-render paths.

## Tests
Added `TestAnnotateAndMerge` cases:
- `normalizes glued document separator at file start`
- `normalizes glued document separator between manifests`

Fixes #31867

### DIFF
--- a/pkg/action/action.go
+++ b/pkg/action/action.go
@@ -26,6 +26,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"regexp"
 	"slices"
 	"strings"
 	"sync"
@@ -62,6 +63,7 @@ import (
 var Timestamper = time.Now
 
 var (
+	compactDocSeparatorPattern = regexp.MustCompile(`(?m)^---(?=\S)`)
 	// errMissingChart indicates that a chart was not provided.
 	errMissingChart = errors.New("no chart provided")
 	// errMissingRelease indicates that a release (name) was not provided.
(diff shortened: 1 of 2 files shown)


LarytheLord:
Scoped this PR down to issue #31867 only (post-renderer glued separator normalization). I rebased and dropped the unrelated REST client test change so this stays single-purpose and easier to review.

scottrigby:
the issue has been fixed in https://github.com/helm/helm/pull/31868

One note, this PR also does not work. Did you test it? Even without a postrenderer it breaks with `helm template`:

```
panic: regexp: Compile(`(?m)^---(?=\S)`): error parsing regexp: invalid or unsupported Perl syntax: `(?=`

goroutine 1 [running]:
regexp.MustCompile({0x103fe3b9c, 0xe})
        regexp/regexp.go:313 +0xb0
helm.sh/helm/v4/pkg/action.init()
        helm.sh/helm/v4/pkg/action/action.go:66 +0x28
```

PR #31912, opened 2026-03-06
fix: graceful fallback when debug.BuildInfo is unavailable

TerryHowe:
When Helm is built with toolchains other than `go build` (e.g. Bazel), `debug.ReadBuildInfo()` returns false, causing a panic during DefaultCapabilities initialization. Fall back to the default Kubernetes version (v1.20.0) instead of panicking.

Fixes #31904

**If applicable**:

### DIFF
--- a/pkg/chart/common/capabilities.go
+++ b/pkg/chart/common/capabilities.go
@@ -17,6 +17,7 @@ package common
 
 import (
 	"fmt"
+	"log/slog"
 	"slices"
 	"strconv"
 	"strings"
@@ -33,8 +34,8 @@ import (
 )
 
 const (
-	kubeVersionMajorTesting = 1
-	kubeVersionMinorTesting = 20
+	kubeVersionMajorDefault = 1
+	kubeVersionMinorDefault = 20
 )
 
 var (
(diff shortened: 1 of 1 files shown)


pkg/chart/common/capabilities.go
```diff
@@ -147,12 +148,15 @@ func makeDefaultCapabilities() (*Capabilities, error) {
...
-		return newCapabilities(kubeVersionMajorTesting, kubeVersionMinorTesting)
+		return newCapabilities(kubeVersionMajorDefault, kubeVersionMinorDefault)
 	}
 
 	vstr, err := helmversion.K8sIOClientGoModVersion()
 	if err != nil {
-		return nil, fmt.Errorf("failed to retrieve k8s.io/client-go version: %w", err)
+		// Build info may be unavailable when compiled with toolchains other
+		// than "go build" (e.g. Bazel). Fall back to a safe default.
+		slog.Warn("failed to retrieve k8s.io/client-go version, falling back to default Kubernetes version", slog.Any("error", err))
```
gjenkins8, line 158: `makeDefaultCapabilities()` is called from an `init()` function, will `slog` will work as expected here?

TerryHowe: Well, it would work, but maybe not as expected as it would hit the default slog handler. Do you have a different recommendation?

gjenkins8: I think best option would be to partially reintroduce the build variables removed by https://github.com/helm/helm/pull/31296 and enable a "non-standard build" to override the versions pulled from DebugInfo

That way, a non-standard build could set a build flag e.g. `-X helm.sh/helm/v4/internal/version.kubeClientVersionMinorOverride=34`, and if (and only if) this was set, `helmversion` would utilize this rather than the debuginfo version

TerryHowe:
I'm going to let someone else give this a shot.

PR #31915, opened 2026-03-08
fix: make toToml swallow errors consistent with toYaml and toJson

### LINKED ISSUE #31430: Unalignment between toToml and toYaml/toJson

realgagangupta:
## Problem

\`toToml\` returns \`err.Error()\` on failure, which renders the raw error message string directly into the template output. This is inconsistent with \`toYAML\` and \`toJSON\`, both of which swallow errors silently by returning \`""\`.

**Before:**
\`\`\`go
if err != nil {
    return err.Error()  // renders error text into template
}
\`\`\`

**After:**
\`\`\`go
if err != nil {
    // Swallow errors inside of a template.
    return ""  // consistent with toYAML and toJSON
}
\`\`\`

## Solution

Changed \`toTOML\` to return \`""\` on error, matching the behavior of \`toYAML\` and \`toJSON\`.

Fixes #31430

### DIFF
--- a/pkg/engine/funcs.go
+++ b/pkg/engine/funcs.go
@@ -150,6 +150,7 @@ func fromYAMLArray(str string) []any {
 
 // toTOML takes an interface, marshals it to toml, and returns a string. It will
 // always return a string, even on marshal error (empty string).
+// Errors are swallowed, consistent with toYAML and toJSON behavior.
 //
 // This is designed to be called from a template.
 func toTOML(v any) string {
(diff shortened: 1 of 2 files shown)


TerryHowe:
Description is still talking about the non existent mustToToml

realgagangupta:
Good catch\! Updated the description — `mustToToml` doesn't exist so that line was incorrect. Removed it.

PR #31925, opened 2026-03-11
fix: protect FailingKubeClient.RecordedWaitOptions from data race

TerryHowe:
## Summary

- Add `sync.Mutex` to `FailingKubeClient` to guard concurrent access to `RecordedWaitOptions` in `GetWaiterWithOptions`
- Fixes data race detected by `-race` flag when concurrent goroutines (upgrade + rollback, install + uninstall) both call `GetWaiterWithOptions` on the same `FailingKubeClient` instance

## Test plan

- [x] `go test -race -run TestUpgradeRelease_Interrupted_RollbackOnFailure ./pkg/action` passes
- [x] `go test -race -run TestInstallRelease_RollbackOnFailure_Interrupted ./pkg/action` passes
- [x] Full `make test-unit` passes

### DIFF
--- a/pkg/kube/fake/failing_kube_client.go
+++ b/pkg/kube/fake/failing_kube_client.go
@@ -19,6 +19,7 @@ package fake
 
 import (
 	"io"
+	"sync"
 	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -49,6 +50,7 @@ type FailingKubeClient struct {
 	WaitDuration           time.Duration
 	// RecordedWaitOptions stores the WaitOptions passed to GetWaiter for testing
 	RecordedWaitOptions []kube.WaitOption
+	mu                  sync.Mutex
 }
 
 var _ kube.Interface = &FailingKubeClient{}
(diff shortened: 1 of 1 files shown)


pkg/kube/fake/failing_kube_client.go
```diff
@@ -160,7 +162,9 @@ func (f *FailingKubeClient) GetWaiter(ws kube.WaitStrategy) (kube.Waiter, error)
 
 func (f *FailingKubeClient) GetWaiterWithOptions(ws kube.WaitStrategy, opts ...kube.WaitOption) (kube.Waiter, error) {
 	// Record the WaitOptions for testing
+	f.mu.Lock()
 	f.RecordedWaitOptions = append(f.RecordedWaitOptions, opts...)
+	f.mu.Unlock()
```
gjenkins8:
```suggestion
func (f *FailingKubeClient) appendRecordedWaitOptionsLocked(opts ...kube.WaitOption) {
	f.mu.Lock()
	defer f.mu.Unlock()
f.RecordedWaitOptions = append(f.RecordedWaitOptions, opts...)
}
```

Make this a helper function and utilize defer? Or overkill? (I dislike non-deferred cleanup, future footgun)

TerryHowe: Done.

scottrigby (approved):
Thanks for this catch 👍

